PR #4141, opened 2014-11-11
Default buffer encoding to core.fileEncoding config

### LINKED ISSUE #3: Correctly detect the encoding or be able to use a default value while opening.
It would be interesting that it would **detect** the encoding, instead of just allow us to select a new one. Or at least have a **default encoding** value.

It always opens as **UTF-8**, but most of the files that I use are **Windows 1252** / **ISO 8859-1**, so I have to manually change every file.

Thanks

kevinsawicki:
Adds `core.fileEncoding` which will be used as the editor/buffer's default encoding when reading/writing.

I went with `core` instead of the `editor` for the prefix since this default should be respected for any packages reading/writing files, regardless of whether they are using the editor/buffer API or not.

screen shot 2014-11-10 at 5 59 01 pm

Closes atom/encoding-selector#3

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -98,12 +98,13 @@ describe "TextEditor", ->
       expect(editor2.isFoldedAtBufferRow(4)).not.toBe editor.isFoldedAtBufferRow(4)
 
   describe "config defaults", ->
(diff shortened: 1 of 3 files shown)


demon386:
Is it possible to use "auto-detect" as value for core.fileEncoding?

PR #4153, opened 2014-11-11
Set theme settings variables from config

kevinsawicki:
### Overview
- [x] Inject all theme config settings as less variables when loading theme stylesheets
- [x] Don't load stylesheets until after the config has been activated for a theme
- [x] Make `activate` method in package main optional. Since themes will most likely only set a `config` object in their package main, requiring them all to have a no-op `activate: ->` seemed ugly. Also `deactivate` was already optional so this makes it consistent with that.

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -798,6 +798,12 @@ class Config
     @setDefaults(keyPath, @extractDefaultsFromSchema(schema))
     @setScopedDefaultsFromSchema(keyPath, schema)
 
+    if keyPath
+      value = @get(keyPath)
+      @set(keyPath, value) unless value is undefined
+
+    return
+
   load: ->
     @initializeConfigDirectory()
     @loadUserConfig()
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -800,7 +800,8 @@ class Config
 
     if keyPath
       value = @get(keyPath)
-      @set(keyPath, value) unless value is undefined
+      unless value is undefined
+        @unset(keyPath) unless @set(keyPath, value)
 
     return
 
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -801,7 +801,10 @@ class Config
     if keyPath
       value = @get(keyPath)
       unless value is undefined
-        @unset(keyPath) unless @set(keyPath, value)
+        try
+          @makeValueConformToSchema(keyPath, value)
+        catch error
+          @unset(keyPath)
 
     return
 
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -798,16 +798,6 @@ class Config
     @setDefaults(keyPath, @extractDefaultsFromSchema(schema))
     @setScopedDefaultsFromSchema(keyPath, schema)
 
-    if keyPath
-      value = @get(keyPath)
-      unless value is undefined
-        try
-          @makeValueConformToSchema(keyPath, value)
-        catch error
-          @unset(keyPath)
-
-    return
-
   load: ->
     @initializeConfigDirectory()
     @loadUserConfig()
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "jasmine-json": "~0.0",
     "jasmine-tagged": "^1.1.2",
     "jquery": "^2.1.1",
-    "less-cache": "0.20.0",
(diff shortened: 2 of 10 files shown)


src/config.coffee
```diff
@@ -798,6 +798,13 @@ class Config
     @setDefaults(keyPath, @extractDefaultsFromSchema(schema))
     @setScopedDefaultsFromSchema(keyPath, schema)
 
+    if keyPath
+      value = @get(keyPath)
+      unless value is undefined
+        @unset(keyPath) unless @set(keyPath, value)
```
kevinsawicki: @benogle this was the way I came up with to fix https://github.com/atom/atom/issues/5219, not sure if there is a better way. Basically when a schema is set, unset the value if it is no longer valid.

nevir:
Curious if there's any movement on this; I'm definitely very interested in using it! :)

kevinsawicki:
> Curious if there's any movement on this

No, the proof of concept went smoothly, but getting it production ready did not go as well.

The main issue was that it required drastically changing Atom's startup since package configs had to be loaded before loading any core stylesheets could begin since package config values could be pulled into computed core stylesheet variables.

It also required recompiling core, package, and theme stylesheets on config changes which is really expensive as well.  Correctly caching and busting the cache on config changes and stylesheet updates was also presenting issues.

Therefore I don't think this is something that will ship in Atom 1.0, perhaps using CSS variables once they ship in Chrome could be used for a similar feature.

nevir:
Fair enough!

wdanilo:
Hello! Is there any progress regarding this issue? Will be able to provide ui-theme options that affect less variables anytime soon? :)

braver:
@wdanilo I have been able to implement stylesheet options by [setting css variables in a style attribute](https://github.com/braver/fonts/commit/2072988be451bbc4c26896ce0314168d0808633a). Depending on what kind options you need this is also more elegant as the changes are applied without having to re-compile less, which in theory also lets you transition between values.

wbinnssmith:
@braver have you thought about contributing this to Atom. Likewise, Atom folks, do you think this has a place in core? I believe I do.

CSS custom properties derived from the user's config and declared in a high-level ancestor such as `atom-workspace` can allow the following:
* Avoids changing Atom's startup sequence and the need to recompile-LESS-on-change.
* Accessible at runtime, therefore accessible from both CSS and LESS user styles, themes, and packages
* The ability to modify this value at runtime (through settings etc) without an Atom reload
* The ability to read these values in **JavaScript** as well through `el.getPropertyValue('--my-var')`, though `atom.config` is still a much better place to read core config settings
* The ability to override the property value at a lower level in the tree

Current workarounds involve things like monitoring changes to `atom.config` and modifying a style node to reflect these changes, as is done in [the GitHub package](https://github.com/atom/github/blob/da2367d063fa834f21a2b0ca9eaf4bb776b126c3/lib/models/style-calculator.js). We'll likely have to implement something similar throughout Atom IDE, such as in its Outline pane.

I can even see a future where all UI and Theme constants are declared this way, thus allowing package authors and users to use whichever CSS preprocessor (or none at all!) they wish. Though this would require a transition plan from LESS variables 😬

cc @damieng @binarymuse

braver:
@wbinnssmith In my own professional projects where I can rely on a recent version of Chrome, I have moved to custom properties over Less variables (effectively using Less only for its nested selectors), and it’s amazing really. So much more powerful and flexible and everything updates immediately. 

Atom UI themes tend do a bunch of color calculations though, I’m not 100% sure how much of that is already implemented and available in Chrome. But color functions are in the CSS spec.

simurai:
@wbinnssmith Agreed. Would be awesome to use custom properties.

@braver has a valid concern about color calculations. Like `darken()`, `lighten()` or checking contrast. It will eventually be available in CSS, but doesn't seem to be listed here: https://www.chromestatus.com/features#color. So probably not that soon.

But might still be worth a try to see how far we get without them. Maybe add more variables so that package authors don't need to calculate other colors. And if they really do, they would have to go from CSS > JS > CSS. Instead of Less.

I played around with this https://github.com/simurai/tone-syntax theme, where you can change colors in the theme settings and then it spits out all the custom variables. Performance is pretty ok.

Tone settings

PR #4169, opened 2014-11-13
Add package upgrade guide

benogle:
This is an initial draft of docs for converting your packages. Initially i'm thinking we want to have 3 docs
### To Write
- [x] TL;DR
- [x] Viewing Deprecations
- [x] Updating Views
- [x] Updating Specs
- [x] Eventing changes + Disposables
- [x] Updating commands
- [x] A little bit about the stylesheets, linking back to the theme guides

My vision for them is written around specific patterns with short explanations and links to examples. Like "I have a SelectListView view, how do I do this with the new APIs?". And we can link to our [PRs for conversion](https://github.com/atom/command-palette/pull/19/files). Deprecations will get all the nits, but there are definite patterns we will find. 

The themes are handled in https://github.com/atom/atom/pull/4214

### DIFF
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -141,6 +141,8 @@ Sometimes it is as simple as converting the requires at the top of each view pag
 
 ### Upgrading classes extending any space-pen View
 
+#### `afterAttach` and `beforeRemove` updated
+
 The `afterAttach` and `beforeRemove` hooks have been replaced with
 `attached` and `detached` and the semantics have changed.
 
@@ -170,6 +172,10 @@ class MyView extends View
   detached: ->
     #...
 ```
+
+#### `subscribe` and `subscribeToCommand` methods removed
+
+Additionally, the `subscribe` and `subscribeToCommand` methods have been removed. See the Eventing and Disposables section for more info.
 
 ### Upgrading to the new TextEditorView
 
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -547,7 +547,12 @@ class Something
     @emitter.on 'did-change', callback
 
   methodThatFiresAChange: ->
-    @emitter.emit 'did-change'
+    @emitter.emit 'did-change', {data: 2}
+
+something = new Something
+something.onDidChange (eventObject) ->
+  console.log eventObject.data # => 2
+something.methodThatFiresAChange()
 ```
 
 ## Subscribing To Commands
@@ -568,7 +573,9 @@ atom.workspaceView.command 'core:close core:cancel', ->
   'core:close': ->
   'core:cancel': ->
 
-# When in a View class, you should have a `@element` object available. `@element` is a raw HTML element
+# You can register commands directly on individual DOM elements in addition to
+# using selectors. When in a View class, you should have a `@element` object
+# available. `@element` is a plain HTMLElement object
 @disposables.add atom.commands.add @element,
   'core:close': ->
   'core:cancel': ->
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -397,15 +397,15 @@ We're going to use `atom.views.getView()` again to get the editor element. As in
 ```coffee
 # New!
 describe 'Something', ->
-  [editor, editorView] = []
+  [editor, editorElement] = []
   beforeEach ->
     editor = atom.workspace.getActiveTextEditor()
-    editorView = atom.views.getView(editor)
+    editorElement = atom.views.getView(editor)
 ```
 
 ### Dispatching commands
 
-Since the `editorView` objects are no longer `jQuery` objects, they no longer support `trigger()`. Additionally, Atom has a new command dispatcher, `atom.commands`, that we use rather than commandeering jQuery's `trigger` method.
+Since the `editorElement` objects are no longer `jQuery` objects, they no longer support `trigger()`. Additionally, Atom has a new command dispatcher, `atom.commands`, that we use rather than commandeering jQuery's `trigger` method.
 
 From this:
 
@@ -420,14 +420,14 @@ To this:
 ```coffee
 # New!
 atom.commands.dispatch workspaceElement, 'a-package:toggle'
-atom.commands.dispatch editorView, 'find-and-replace:show'
+atom.commands.dispatch editorElement, 'find-and-replace:show'
 ```
 
 ## Eventing and Disposables
 
 A couple large things changed with respect to events:
 
-1. All model events are now methods that return a [`Disposable`][disposable] object
+1. All model events are now exposed as event subscription methods that return [`Disposable`][disposable] objects
 1. The `subscribe()` method is no longer available on `space-pen` `View` objects
 1. An Emitter is now provided from `require 'atom'`
 
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -105,6 +105,101 @@ Sometimes it is as simple as converting the requires at the top of each view pag
 {TextEditorView} = require 'atom-space-pen-views'
 ```
 
+If you are using the lifecycle hooks, you will need to update code as well.
+
+### Upgrading to space-pen's TextEditorView
+
+You should not need to change anything to use the new `TextEditorView`! See the [docs][TextEditorView] for more info.
+
+### Upgrading to space-pen's ScrollView
+
+See the [docs][ScrollView] for all the options.
+
+### Upgrading to space-pen's SelectListView
+
+Your SelectListView might look something like this:
+
+```coffee
+class CommandPaletteView extends SelectListView
+  initialize: ->
+    super
+    @addClass('command-palette overlay from-top')
+    atom.workspaceView.command 'command-palette:toggle', => @toggle()
+
+  confirmed: ({name, jQuery}) ->
+    @cancel()
+    # do something with the result
+
+  toggle: ->
+    if @hasParent()
+      @cancel()
+    else
+      @attach()
+
+  attach: ->
+    @storeFocusedElement()
+
+    items = # build items
+    @setItems(items)
+
+    atom.workspaceView.append(this)
+    @focusFilterEditor()
+
+  confirmed: ({name, jQuery}) ->
+    @cancel()
+```
+
+This attaches and detaches itself from the dom when toggled, canceling magically detaches it from the DOM, and it uses the classes `overlay` and `from-top`.
+
+Using the new APIs it should look like this:
+
+```coffee
+class CommandPaletteView extends SelectListView
+  initialize: ->
+    super
+    # no more need for the `overlay` and `from-top` classes
+    @addClass('command-palette')
+    atom.workspaceView.command 'command-palette:toggle', => @toggle()
+
+  # You need to implement the `cancelled` method and hide.
+  cancelled: ->
+    @hide()
+
+  confirmed: ({name, jQuery}) ->
+    @cancel()
+    # do something with the result
+
+  toggle: ->
+    # Toggling now checks panel visibility,
+    # and hides / shows rather than attaching to / detaching from the DOM.
+    if @panel?.isVisible()
+      @cancel()
+    else
+      @show()
+
+  show: ->
+    # Now you will add your select list as a modal panel to the workspace
+    @panel ?= atom.workspace.addModalPanel(item: this)
+    @panel.show()
+
+    @storeFocusedElement()
+
+    items = # build items
+    @setItems(items)
+
+    @focusFilterEditor()
+
+  hide: ->
+    @panel?.hide()
+```
+
+See the [SelectListView docs][SelectListView] for all the options.
+
 ## Specs
 
-TODO: come up with patterns for
+TODO: come up with patterns for converting away from using `workspaceView` and `editorView`s everywhere.
+
+
+[texteditorview]:https://github.com/atom/atom-space-pen-views#texteditorview
+[scrollview]:https://github.com/atom/atom-space-pen-views#scrollview
+[selectlistview]:https://github.com/atom/atom-space-pen-views#selectlistview
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -193,7 +193,7 @@ class CommandPaletteView extends SelectListView
     @panel?.hide()
 ```
 
-See the [SelectListView docs][SelectListView] for all the options.
+See the [SelectListView docs][SelectListView] for all the options. And check out the [conversion of CommandPaletteView][selectlistview-example] as a real-world example.
 
 ## Specs
 
@@ -203,3 +203,4 @@ TODO: come up with patterns for converting away from using `workspaceView` and `
 [texteditorview]:https://github.com/atom/atom-space-pen-views#texteditorview
 [scrollview]:https://github.com/atom/atom-space-pen-views#scrollview
 [selectlistview]:https://github.com/atom/atom-space-pen-views#selectlistview
+[selectlistview-example]:https://github.com/atom/command-palette/pull/19/files
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -22,9 +22,7 @@ TODO: image of deprecations in DepCop
 
 ## Views
 
-### The Old
-
-Previous to 1.0, views in packages were baked into Atom core. These views were based on jQuery and `space-pen`. They look something like this:
+Previous to 1.0, views in packages were baked into Atom core. These views were based on jQuery and `space-pen`. They looked something like this:
 
 ```coffee
 {$, TextEditorView, View} = require 'atom'
@@ -75,29 +73,32 @@ SelectListView
 
 `Workspace` and `WorkspaceView` are _no longer provided_ in any capacity. They should be unnecessary
 
-#### jQuery
+### Adding the module dependencies
 
-If you do not need `space-pen`, you can require jQuery directly. In your `package.json` add this to the `dependencies` object:
-
-```js
-"jquery": "^2"
-```
-
-#### NPM dependencies
+To use the new views, you need to specify a couple modules in your package dependencies in your `package.json` file:
 
 ```js
 {
   "dependencies": {
-    "jquery": "^2" // if you want to include jquery directly
     "space-pen": "^3"
     "atom-space-pen-views": "^0"
   }
 }
 ```
 
-#### Converting your views
+`space-pen` bundles jQuery. If you do not need `space-pen`, you can require jQuery directly.
 
-Sometimes it should be as simple as converting the requires at the top of each view page. In the case of our above example, you can just convert them to the following:
+```js
+{
+  "dependencies": {
+    "jquery": "^2"
+  }
+}
+```
+
+### Converting your views
+
+Sometimes it is as simple as converting the requires at the top of each view page. In the case of our above example, you can just convert them to the following:
 
 ```coffee
 {$, View} = require 'space-pen'
--- a/docs/upgrading/upgrading-your-package.md
+++ b/docs/upgrading/upgrading-your-package.md
@@ -1,0 +1,109 @@
+# Upgrading your package to 1.0 APIs
+
+Atom is rapidly approaching 1.0. Much of the effort leading up to the 1.0 has been cleaning up APIs in an attempt to future proof, and make a more pleasant experience developing packages.
+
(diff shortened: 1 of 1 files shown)


benogle:
This is ready for a review. It isnt the most exciting thing to read, I know.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
+# Upgrading your package to 1.0 APIs
+
+Atom is rapidly approaching 1.0. Much of the effort leading up to the 1.0 has been cleaning up APIs in an attempt to future proof, and make a more pleasant experience developing packages.
+
+This document will guide you through the large bits of upgrading your package to work with 1.0 APIs.
+
+## TL;DR
+
+We've set deprecation messages and errors in strategic places to help make sure you dont miss anything. You should be able to get 95% of the way to an updated package just by fixing errors and deprecations. There are a couple of things you need to do to enable all these errors and deprecations.
```
50Wliu: don't

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
...
+```
+
+So this:
+
+```coffee
+# Old way
+{$, TextEditorView, View, GitRepository} = require 'atom'
+```
+
+Would be replaced with this:
```
50Wliu: I feel like "by this" sounds better.  Might just be me though.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
...
+
+```coffee
+# New way
+{View} = require 'atom-space-pen-views'
+class MyView extends View
+  attached: ->
+    # Always called with the equivalent of @afterAttach(true)!
+    #...
+
+  removed: ->
```
lee-dohm: Should be `detached`?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
...
+```
+
+### Converting your views
+
+Sometimes it is as simple as converting the requires at the top of each view page. I assume you read the 'TL;DR' section and have updated all of your requires.
+
+### Upgrading classes extending any space-pen View
+
+The `afterAttach` and `beforeRemove` hooks have been replaced with
+`attached` and `detached` and their semantics have been altered. `attached` will only be called when all parents of the View are attached to the DOM.
```
lee-dohm: No mention of the difference in the `detached` hook's semantics?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
...
+
+```
+$
+$$
+$$$
+View
+TextEditorView
+ScrollView
+SelectListView
+```
```
50Wliu: This section seems to duplicate line 29

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,565 @@
...
+
+// Atom specific views
+TextEditorView
+ScrollView
+SelectListView
+```
+
+### Adding the module dependencies
+
+To use the new views, you need to specify the `atom-space-pen-views` module in your package's `package.json` file's dependencies:
```
lee-dohm: Make it clear that `atom-space-pen-views` is the dependency to include even if you only need `require 'space-pen'`?

docs/upgrading/upgrading-your-package.md
benogle: done

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,541 @@
...
+  toggle: ->
+    if @hasParent()
+      @cancel()
+    else
+      @attach()
+
+  attach: ->
+    @storeFocusedElement()
+
+    items = # build items
```
kevinsawicki: Sorry, forgot to leave a comment here as well about updating this one too.

benogle:
Ok, updated based on all the feedback. Would be good to get a quick read through from @nathansobo

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,541 @@
...
+
+  detached: ->
+    #...
+```
+
+### Upgrading to the new TextEditorView
+
+You should not need to change anything to use the new `TextEditorView`! See the [docs][TextEditorView] for more info.
+
+### Upgrading to classes extending ScrollView
```
kevinsawicki: Maybe drop the `to`?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,541 @@
...
+  initialize: (@model) ->
+    disposable = super()
+    # turn off default scrolling behavior from ScrollView
+    disposable.dispose()
+```
+
+* Check out [an example](https://github.com/atom/find-and-replace/pull/311/files#diff-9) from find-and-replace.
+* See the [docs][ScrollView] for all the options.
+
+### Upgrading to classes extending SelectListView
```
kevinsawicki: Also here, maybe drop the `to`?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
+# Upgrading your package to 1.0 APIs
+
+Atom is rapidly approaching 1.0. Much of the effort leading up to the 1.0 has been cleaning up APIs in an attempt to future proof, and make a more pleasant experience developing packages.
+
+This document will guide you through the large bits of upgrading your package to work with 1.0 APIs.
+
+## TL;DR
+
+We've set deprecation messages and errors in strategic places to help make sure you don't miss anything. You should be able to get 95% of the way to an updated package just by fixing errors and deprecations. There are a couple of things you need to do to enable all these errors and deprecations.
```
nathansobo: > There are a couple of things you need to do to enable all these errors and deprecations

This doesn't make sense to me. Aren't the deprecation warnings enabled by default? It seems like you would want to say: "There are a few basic changes most packages will require to avoid the majority of deprecation warnings." Or something like that.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+{
+  "dependencies": {
+    "atom-space-pen-views": "^2.0.2"
+  }
+}
+```
+
+Then run `apm install` in your package directory.
+
+### Require views from atom-space-pen-views
```
nathansobo: Maybe it would be worth mentioning the need to rename the `afterAttach` and `beforeRemove` hooks in the section about switching views. Also might be worth mentioning that the `::subscribe` method is no longer provided and link to the section about subscriptions?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+
+### Specs
+
+Just run your specs, and all the deprecations will be displayed in yellow.
+
+![spec-deps](https://cloud.githubusercontent.com/assets/69169/5637943/b85114ba-95b5-11e4-8681-b81ea8f556d7.png)
+
+### Deprecation Cop
+
+Run an atom window in dev mode (`atom -d`) with your package loaded, and open Deprecation Cop (search for `deprecation` in the command palette).
```
nathansobo: It won't have to be in dev mode once we go live with the deprecations in production. Not sure if that matters though.

nathansobo: Also maybe worth mentioning that deprecations won't appear until your package actually calls the methods. They might not be present immediately.

docs/upgrading/upgrading-your-package.md
nathansobo: I see. You expand on this below. Nevermind.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+```
+
+### Converting your views
+
+Sometimes it is as simple as converting the requires at the top of each view page. I assume you read the 'TL;DR' section and have updated all of your requires.
+
+### Upgrading classes extending any space-pen View
+
+The `afterAttach` and `beforeRemove` hooks have been replaced with
+`attached` and `detached`. The `attached` method semantics have changed: it will only be called when all parents of the View are attached to the DOM. The `detached` semantics have not changed.
```
nathansobo: The `detached` semantics actually *have* changed. The `beforeRemove` hook is only called when you call `$.fn.remove`, which is typically used when the node is completely removed from the DOM. The `detached` hook is called whenever the DOM node is detached, which could happen if the node is being removed. In short, if `beforeRemove` is called the node is never coming back. With `detached` it might be attached again later.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+    # Always called with the equivalent of @afterAttach(true)!
+    #...
+
+  detached: ->
+    #...
+```
+
+### Upgrading to the new TextEditorView
+
+You should not need to change anything to use the new `TextEditorView`! See the [docs][TextEditorView] for more info.
```
nathansobo: Actually, it doesn't have all the same methods as the original. They should direct everything through the model, available via `TextEditorView::getModel`.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+    @setItems(items)
+
+    atom.workspaceView.append(this)
+    @focusFilterEditor()
+
+  confirmed: ({name, jQuery}) ->
+    @cancel()
+```
+
+This attaches and detaches itself from the dom when toggled, canceling magically detaches it from the DOM, and it uses the classes `overlay` and `from-top`.
```
nathansobo: Might add clarity to explicitly state: "The new `SelectListView` no longer automatically detaches itself from the DOM when cancelled. It's up to you to implement this on your own if that's your desired behavior."

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+
+```coffee
+# Old!
+{WorkspaceView} = require 'atom'
+describe 'FindView', ->
+  beforeEach ->
+    atom.workspaceView = new WorkspaceView()
+```
+
+Instead, we will use the `atom.views.getView()` method. This will return a plain `HTMLElement`, not a `WorkspaceView` or jQuery object.
```
nathansobo: How about:

Many of the actions performed via `atom.workspaceView` can instead be managed via the workspace model layer, available at `atom.workspace`. For actions that still require the view, such as dispatching commands, you'll access the view via the `atom.views.getView()`  method. This will return a subclass of `HTMLElement` rather than a `WorkspaceView` instance, which was a subclass of `jQuery`.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+* And check out the [conversion of CommandPaletteView][selectlistview-example] as a real-world example.
+* See the [SelectListView docs][SelectListView] for all options.
+
+## Updating Specs
+
+`WorkspaceView` and `EditorView` have been deprecated. These two objects are used heavily throughout specs, mostly to dispatch events and commands. This section will explain how to remove them while still retaining the ability to dispatch events and commands.
+
+### Removing WorkspaceView references
+
+`WorkspaceView` has been deprecated. Everything you could do on the view, you can now do on the `Workspace` model.
```
nathansobo: I think we should say:

> The `WorkspaceView` class and `atom.workspaceView` global have been deprecated.

Maybe part of my suggestion below should be moved upward. I think it's important to say that *most* things can be can and should be performed through the model layer, but if you need the view, you can get it.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+Change it to:
+
+```coffee
+# New!
+jasmine.attachToDOM(workspaceElement)
+```
+
+### Removing EditorView references
+
+Like `WorkspaceView`, `EditorView` has been deprecated. Everything you needed to do on the view you are now able to do on the `Editor` model.
```
nathansobo: `Editor` => `TextEditor`

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+
+```coffee
+# Old!
+describe 'Something', ->
+  [editorView] = []
+  beforeEach ->
+    editorView = atom.workspaceView.getActiveView()
+```
+
+We're going to use `atom.views.getView()` again to get the editor element. As in the case of the `workspaceElement`, `getView` will return a plain `HTMLElement` rather than an `EditorView` or jQuery object.
```
nathansobo: Maybe:
> return a subclass of HTMLElement rather than a jQuery object

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+
+```coffee
+# New!
+describe 'Something', ->
+  [editor, editorView] = []
+  beforeEach ->
+    editor = atom.workspace.getActiveTextEditor()
+    editorView = atom.views.getView(editor)
```
nathansobo: Should we say `editorElement` here to drive the point home?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+workspaceView.trigger 'a-package:toggle'
+editorView.trigger 'find-and-replace:show'
+```
+
+To this:
+
+```coffee
+# New!
+atom.commands.dispatch workspaceElement, 'a-package:toggle'
+atom.commands.dispatch editorView, 'find-and-replace:show'
```
nathansobo: Again, I think `editorElement` would be better.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+```
+
+## Eventing and Disposables
+
+A couple large things changed with respect to events:
+
+1. All model events are now methods that return a [`Disposable`][disposable] object
```
nathansobo: How about:
> All model events are now exposed via event subscription methods that return a `Disposable` object.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+1. The `subscribe()` method is no longer available on `space-pen` `View` objects
```
nathansobo: Worth mentioning above and linking here?

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+{Emitter} = require 'atom'
+
+class Something
+  constructor: ->
+    @emitter = new Emitter
+
+  destroy: ->
+    @emitter.dispose()
+
+  onDidChange: (callback) ->
```
nathansobo: Worth showcasing with an argument?

docs/upgrading/upgrading-your-package.md
kevinsawicki: I suggested switching it from `editorElement` in a previous comment.

docs/upgrading/upgrading-your-package.md
```diff
@@ -0,0 +1,542 @@
...
+@subscribeToCommand 'core:close core:cancel', ->
+```
+
+```coffee
+# New!
+@disposables.add atom.commands.add 'atom-workspace',
+  'core:close': ->
+  'core:cancel': ->
+
+# When in a View class, you should have a `@element` object available. `@element` is a raw HTML element
```
nathansobo: I think it would be clearer if we prefaced this with

> You can register commands directly on individual DOM elements in addition to using selectors.

nathansobo:
:clap: This is incredibly well done. Thanks so much.

docs/upgrading/upgrading-your-package.md
benogle: Y'all can fight it out. I dont care, but lean toward `editorElement` as it is consistent with workspaceElement, and will be an element in this case. But I get the inconsistency with `getView`.

kevinsawicki: Yeah, I would say change it to `workspaceView` as well for consistency.

It seems weird to expect more from `getView` than a view, putting `element` in the name tells you it is more than a view, it goes into its (current) implementation.

It also makes it kind of weird in packages that don't use custom elements but use the `getView` API. In archive view, I would have `archiveView = atom.views.getView(archive)`.  So why not always be consistent, and assume the `getView` returns nothing more than a generic "view" concept.

nathansobo: Both are true, but emphasizing that it's an element seems worth it to me in this case.

kevinsawicki: Keeping it as the classic `editorView` and `workspaceView` can also reduce the amount of updates needed for this API upgrade as lots of code may already have local or instance variables named this way.

kevinsawicki: >  but emphasizing that it's an element seems worth it to me in this case.

Why, in this example, do you want to emphasize it is an HTML element?

nathansobo: I see your point about the `getView` naming. But in truth, this API always returns a DOM element. The implementation does matter in this case as the difference between a DOM node and a jQuery wrapper is far from encapsulated. They both have large public APIs that make the implementation matter quite a bit. You can't call `querySelector` on anything but an `HTMLElement` for example.

nathansobo: > Why, in this example, do you want to emphasize it is an HTML element?

Because it has a very different API from the previous jQuery subclass and methods used previously won't work.

benogle: > Why, in this example, do you want to emphasize it is an HTML element?

Because we're deprecating WorkspaceView and EditorView. I can see confusion calling these new objects `workspaceView` and `editorView`, because they are no longer instances of their namesake classes. The difference in naming here provides a clear separation. It's saying "yo, these are new objects! You can tell cause they arent named the same!".

docs/upgrading/upgrading-your-package.md
benogle: You dont get full effect of all the necessary errors and warnings until you require the views from space pen. It just tells you to require them from somewhere else, then when you do, you get all kinds of new errors.

I can probably be worded better.

docs/upgrading/upgrading-your-package.md
kevinsawicki: > But in truth, this API always returns a DOM element.

Yeah, I didn't realize that this was the case 100% of the time so I see your point better now about putting `element` in the variable name.

So I guess the `getView` name just feels weird then at this point.

nathansobo: > So I guess the getView name just feels weird then at this point.

Yeah, I've gone back and forth on that one. The element *is* a view. That's its role in the system on one conceptual level. But it's also true that we're pretty dependent on it being a DOM node because we need to attach the return value to the DOM, and that's probably not going to change. So the name could also be more specific. Do you think it's worth changing now? I'd be willing to consider it, but I also think the current name is reasonable.

docs/upgrading/upgrading-your-package.md
benogle: Ok, updated with modifications to your comments.

docs/upgrading/upgrading-your-package.md
benogle: Updated

docs/upgrading/upgrading-your-package.md
benogle: I added a whole new section on using the model layer instead of the view layer. It should provide some context, then I left the specs sections largely the same. But with some repetition so that people can skim, and jump to the section they need. 

I didnt put any motivations for using just the view layer in there. I think you should write a post about why just the model layer post API freeze, and we should link to it in that section.

docs/upgrading/upgrading-your-package.md
benogle: I dont think we can link internally in docs on atom.io yet

benogle:
Ok, updated. I'll make the final pass in the morning.

docs/upgrading/upgrading-your-package.md
kevinsawicki: >  Do you think it's worth changing now?

No, I think it the view <-> element relationship can be explained with docs and it will make complete sense.

So yeah, I'm good with element now, thanks for all the deets.

@benogle sorry for the hassle.

nathansobo:
@benogle What do you think about explicitly soliciting pull requests in the introduction to the guide for enhancement?

benogle:
@nathansobo Sure. I can put something at the bottom. Or do you want it more prominent?

nathansobo:
Whatever you think.

nathansobo:
@benogle One more thought. Should we encourage people to do the `,` separated version string in their engines field to indicate they're good to go for 1.0?

PR #4269, opened 2014-11-21
Better user experience for start atom

yongkangchen:
as discuss at https://discuss.atom.io/t/why-is-atom-so-slow/11376/47?u=yongkangchen
By delay load and active packages, user could see window and menu faster.
Cause I usually wait window shown and click menu to open.

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -553,12 +553,16 @@ class Atom extends Model
     @loadConfig()
     @keymaps.loadBundledKeymaps()
     @themes.loadBaseStylesheets()
-    @packages.loadPackages()
+    
+    setTimeout =>
+      @packages.loadPackages()
+      @packages.activate()
+    ,200
+    
     @deserializeEditorWindow()
 
     @watchProjectPath()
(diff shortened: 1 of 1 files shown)


m1ga:
:+1:

src/atom.coffee
```diff
@@ -553,12 +553,16 @@ class Atom extends Model
     @loadConfig()
     @keymaps.loadBundledKeymaps()
     @themes.loadBaseStylesheets()
-    @packages.loadPackages()
+    
+    setTimeout =>
+      @packages.loadPackages()
```
kevinsawicki: If packages are loaded after the editor window is deserialized (`@deserializeEditorWindow()`) then any previously open editors from packages (such as image view, archive view, etc.) won't reopen since their deserializers won't be registered until after deserialization occurs.

theblacklion:
Would be really nice if you could figure out how to solve that puzzle. Having Atom open windows fast is essential when using it for occasionally opening files from the terminal.

PR #4288, opened 2014-11-22
Preserve first line's leading whitespace on paste

maxbrunsfeld:
This affects the behavior of copy/paste when `Auto Indent on Paste` is turned _off_, i.e. in whitespace-sensitive languages. I think it will make things more intuitive for people. Selecting groups of lines and pasting them at column zero now works as expected.

Note that the relative indentation of the copied text is still preserved on paste, as before.

This change removes one 'Public' method - `Selection::normalizeLines`. I thought of deprecating it, but it implemented a behavior that this PR removes, and no package calls it.

Closes #4272.


ruby24:
Thanks very much guys - was driving me crazy before finding this

PR #4440, opened 2014-12-07
Add an apiVersion method to simplify providing APIs

### LINKED ISSUE #4438: Provide the ability to specify the API version of a package in the package.json?
For packages that provide a single API it could look something like this:

```json
{
  "name": "frobnicator",
  "version": "1.2.3",
  "apiVersion": "1.0.0",
}
```

And for multiple APIs it could look like this:

```json
{
  "name": "frobnicator",
  "version": "1.2.3",
  "apiVersion": {
    "frob": "1.1.0",
    "xyzzy": "2.2.0"
  }
}
```

I suppose that this can easily be done with a little bit of code, but making this a standard and having the code live in Atom Core would be helpful to package developers.

lee-dohm:
Fixes #4438

### DIFF
--- a/spec/service-hub-spec.coffee
+++ b/spec/service-hub-spec.coffee
@@ -1,0 +1,37 @@
+describe 'ServiceHub', ->
+  fdescribe 'apiVersion', ->
+    it 'gets the API version for the default API', ->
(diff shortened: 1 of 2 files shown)


lee-dohm:
A new approach to this has been implemented, closing.

PR #4567, opened 2014-12-12
Improve Config API with regard to scoped settings

nathansobo:
This may be able to supplant #4558.

/cc @maxbrunsfeld: I got the get/set API looking good.
### Problems with Config API

The Config API still has some awkwardness leftover from when scoped settings were merged in. It would be nice to make the API more coherent considering the expanded role of settings.
- We're using a private method, `::addScopedSettings` in packages. We need a public equivalent. The main difference is that `::addScopedSettings` takes a `source` argument. The `source` is used to distinguish between user settings and package settings internally, so it's important.
- Methods related to defaults don't make sense in the presence of scoped settings. The methods `::isDefault`, `::restoreDefault`, and `::getDefault` don't actually deal with true _default_ values from the schemas. Instead they consider a value "default" if its source isn't the user's `config.cson`
### Proposed Solution

We can eliminate `::addScopedSettings` and the methods that deal with defaults from the API. Instead, we replace them with the ability to talk about the _source_ of a given setting when getting and setting it. Then the settings view can make decisions based on the source being the user's `config.cson`, rather than exposing a murky conception of "default" in settings.

We propose to replace the following methods:
- getDefault
- isDefault
- restoreDefault
- addScopedSettings

With these new signatures for get and set
- set(keyPath, value, [{scopeSelector, source}])
- get(keyPath, [{scope, sources, excludeSources}])

Plus the new `unset` method... which can be used to remove a single key path or all key paths from a given scope.
- unset([keyPath], [{source}])

If we need to know if a value is assigned in the user's settings, we specify the desired _source_ in our getter. If we get no value back, we can assume we're using some other value (maybe a schema default, mabye from a package). If we want to determine the non-user-assigned value, we exclude it as a source when getting the setting.
### Steps
- [x] Store global settings in the scoped property store under `*` so they're associated with a source
- [ ] Refactor the scoped-property-store to allow mutating values without always adding new entries to the cascade (atom/scoped-property-store#8).
- [ ] Bugfix: Ensure the most recently assigned setting is honored when there is a specificity tie.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -2,6 +2,7 @@ path = require 'path'
 temp = require 'temp'
 CSON = require 'season'
 fs = require 'fs-plus'
+Grim = require 'grim'
 
 describe "Config", ->
   dotAtomPath = null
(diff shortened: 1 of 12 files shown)


nathansobo:
This was long ago addressed in other PRs.

Issue #4596, opened 2014-12-13
`getVisibleRowRange` reports wrong height (1 too big)


neojski:
I was trying to fix https://github.com/atom/vim-mode/issues/430 and it looks like the [visible row range calculation](https://github.com/atom/atom/blob/d9a5e3c7e1958269cbe92038483e1d4ae35fb935/src/display-buffer.coffee#L355) is wrong.

By the way, it is pretty unclear whether row and column in [scrollToScreenPosition](https://atom.io/docs/api/v0.158.0/TextEditor#instance-scrollToScreenPosition) are counted from 0 or 1.

PR #4597, opened 2014-12-13
Correct visible row range calculation.

neojski:
Fixes issue #4596.

There was already Math.ceil so no need for +1.

Example:
If
  height = 2.5, line height = 1
then
  heightInLines is 3 and not 4

### DIFF
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -352,7 +352,7 @@ class DisplayBuffer extends Model
   getVisibleRowRange: ->
     return [0, 0] unless @getLineHeightInPixels() > 0
 
-    heightInLines = Math.ceil(@getHeight() / @getLineHeightInPixels()) + 1
+    heightInLines = Math.ceil(@getHeight() / @getLineHeightInPixels())
     startRow = Math.floor(@getScrollTop() / @getLineHeightInPixels())
     endRow = Math.min(@getLineCount(), startRow + heightInLines)
 


izuzak:
Wondering what do you think about this pull request, @nathansobo? It doesn't break any specs, so I'm not really sure.

However, I just tried this in console:

screen shot 2015-03-03 at 17 40 52

The first visible row is 94 (half visible) and the last one is 97 (half visible). So, I was a bit surprised by `[93, 98]`. Is that right?

Also, is this method something other packages should be relying on?

nathansobo:
@izuzak Keep in mind that the line numbers in the gutter are 1-indexed and the row numbers are 0-indexed, so the return value is correct for the first row. As for the last row, I think there could be problems when the height is an exact multiple of the line height. Take this example...

height: 100
lineHeight: 10
scrollTop: 5

The height in lines is 10, but since the first line is only half-visible, we can actually see *11* lines, with the first and last lines both partially visible. Unless I'm making some error in my reasoning. One thing we could do is always add 1 to the height so we always round up...

```coffee
heightInLines = Math.ceil(@getHeight() + 1 / @getLineHeightInPixels())
```

That would avoid what is usually an extra line in cases where the height isn't an exact multiple, but avoid being a line short when it is. What do you think? Is my thinking seeming correct here?

neojski:
@nathansobo, there's already `Math.ceil` there so I'm not sure why you want to add that `1`. Also, your code snippet doesn't work because of the missing parens around addition but your reasoning seems fine to me and it's exactly what my change does.

izuzak:
@neojski I think your suggested change does not work correctly. Consider the following case: height: 100, lineHeight: 10, scrollTop: 0. In that case, the first 10 lines are visible and the returned result should be `[0, 9]`.

Following your suggestion, you will get this:

```coffee
heightInLines = Math.ceil(@getHeight() / @getLineHeightInPixels())  # heightInLines is 10
startRow = Math.floor(@getScrollTop() / @getLineHeightInPixels())   # startRow is 0
endRow = Math.min(@getLineCount(), startRow + heightInLines)        # endRow is 10, but that's not correct. It should be 9.
```

I think @nathansobo's suggestion isn't correct either, here's what you get with that for the same example:

```coffee
heightInLines = Math.ceil(@getHeight() + 1 / @getLineHeightInPixels()) # heightInLines is 11
startRow = Math.floor(@getScrollTop() / @getLineHeightInPixels())   # startRow is 0
endRow = Math.min(@getLineCount(), startRow + heightInLines)        # endRow is 11, but that's not correct. It should be 9.
```

I think I see the problem, and I think there's a more direct and simpler way to compute this:

```coffee
startRow = Math.floor(@getScrollTop() / @getLineHeightInPixels())
endRow = Math.ceil((@getScrollTop() + @getHeight()) / @getLineHeightInPixels()) - 1
endRow = Math.min(@getLineCount(), endRow)
```

The idea is not to compute the heightInLines, it's not necessary -- we just compute the start and end row directly using the offset from the top of the window. The start row computation didn't change. The end row computation is done in the same way, but does a ceil (not a floor) so that the screen row hit by the border is taken into account, and 1 is subtracted to get a 0-based index.

With this, the result for the same example is: startRow = floor(0/10) = 0, and endRow = ceil((0+100)/10) - 1 = 9.

And for the example which @nathansobo used (height: 100, lineHeight: 10, scrollTop: 5), I get startRow = floor(5/10) = 0, endRow = ceil((5+100)/10) - 1 = 10, which is correct since line 11 is half-visible as well (index 10).

What do you think?

neojski:
@izuzak, first thing you're right that we don't want heightInLines at all. It's not even not necessary. It breaks things because we want to report the same `startRow` and `endRow` for windows of sizes
```
[0+eps, 10-eps] ~> height = 10 - 2eps

[1-eps, 9+eps] ~> height = 8 + 2eps
```
So the difference of height may be bigger than `1` but we still want to report the same `[0,9]` value. That's a good point.

I also agree that the `endRow` value is pretty correct. The only question is about `-1`. I remember that this function might be supposed to return `endRow` as one past the last (most likely ranges are `[begin, end)` and not `[begin, end]`).

izuzak:
> I remember that this function might be supposed to return endRow as one past the last (most likely ranges are `[begin, end)` and not `[begin, end]`).

Ah, thanks for pointing that out -- totally missed that part. I'm not sure, to be honest (because there seem to be no tests for this method [`getVisibleRowRange`](https://github.com/atom/atom/blob/f15c029938d9929a3f6d6257ccc87450fac64495/src/display-buffer.coffee#L351-L358) directly). Perhaps @nathansobo can clarify that point -- should the returned range end in the last visible line or in last-visible-line + 1 ?

mark-hahn:
>  should the returned range end in the last visible line or in
last-visible-line + 1 ?

My understanding is that everything should be zero-based except when
displaying it.

izuzak:
> My understanding is that everything should be zero-based except when displaying it

Everything is 0-based, but we're not sure what the meaning of the end index is -- should it be the 0-based index of the last visible line (even if partially visible), or the 0-based index of the first line which is not visible (the line after the last visible line).

mark-hahn:
IMHO it should obviously be the normal line after the last visible line.

izuzak:
> IMHO it should obviously be the normal line after the last visible line.

I'm not 100% sure (due to the bug this pull request is trying to fix, the name of the method and the lack of description/tests). So I just want to make sure and that's why I pinged @nathansobo. Once we're sure, we can also extend this PR to also add a short description for that method as well. Does that sound good?

neojski:
Sounds good to me.

nathansobo:
This is a private API. I'm comfortable making it a public API but it should be available on the view (`TextEditorElement`) and not on the model. @as-cii would you be willing to shepherd this forward? I trust your judgment on merging changes to this area of the code. If you're too busy I can handle it.

izuzak:
> This is a private API. I'm comfortable making it a public API but it should be available on the view (`TextEditorElement`) and not on the model.

@nathansobo I don't think it's necessary to make this a public API for the sake of this issue.

As far as I can tell, the problem here is that vim-mode is using `getLastVisibleScreenRow`, [which is a public API](https://github.com/atom/atom/blob/abf368e9c02cf184dce1b978028ce157b672bc7c/src/text-editor-element.coffee#L204), but that method is calling `getVisibleRowRange` internally:

https://github.com/atom/atom/blob/abf368e9c02cf184dce1b978028ce157b672bc7c/src/text-editor.coffee#L2931

So, the method being fixed here (`getVisibleRowRange`) is not being called as a public API. However, the problem in that internal API is manifesting in the public API.

as-cii:
> I don't think it's necessary to make this a public API for the sake of this issue.

--------

> So, the method being fixed here (`getVisibleRowRange`) is not being called as a public API. However, the problem in that internal API is manifesting in the public API.

@izuzak @nathansobo :+1: 

> I'm not 100% sure (due to the bug this pull request is trying to fix, the name of the method and the lack of description/tests). So I just want to make sure and that's why I pinged @nathansobo. Once we're sure, we can also extend this PR to also add a short description for that method as well. Does that sound good?

[There's one method in DisplayBuffer](https://github.com/atom/atom/blob/master/src/display-buffer.coffee#L360-362) which uses `getVisibleRowRange`. In turn, `intersectsVisibleRowRange` is being used by a wrapper which exposes it to the public API and by `selectionIntersectsVisibleRowRange`. Here's an excerpt for you to consider:

```coffee
getVisibleRowRange: ->
  return [0, 0] unless @getLineHeightInPixels() > 0

  heightInLines = Math.ceil(@getHeight() / @getLineHeightInPixels()) + 1
  startRow = Math.floor(@getScrollTop() / @getLineHeightInPixels())
  endRow = Math.min(@getLineCount(), startRow + heightInLines)

  [startRow, endRow]

intersectsVisibleRowRange: (startRow, endRow) ->
  [visibleStart, visibleEnd] = @getVisibleRowRange()
  not (endRow <= visibleStart or visibleEnd <= startRow)

selectionIntersectsVisibleRowRange: (selection) ->
  {start, end} = selection.getScreenRange()
  @intersectsVisibleRowRange(start.row, end.row + 1)
```

Apparently this method is not directly used anywhere else in atom core. On the other hand, `selectionIntersectsVisibleRowRange` **is** used in `TextEditorComponent` but, unfortunately, changing it to always return `false` or `true` doesn't break any test, thus it's quite difficult to make assumptions on what is the current behavior.

However, I think that this piece of code tells something: 

```coffee
@intersectVisibleRowRange(start.row, end.row + 1)
```

Aside from being correct or not, which we cannot verify unfortunately, makes me bet that `@getVisibleRowRange` should return only, well, the visible row range: if lines need to go past such range, they can simply "cheat" on the end row being passed to the method, exactly as `@selectionIntersectsVisibleRowRange` does.

Quick aside: I managed to find out a TODO [in the past](https://github.com/atom/atom/commit/e365e51a2bac05beaa4f1614420475fef84ac803#diff-c74ccbac6e5951e339cf3b8fb0af1cf8R103):

```
@intersectsVisibleRowRange(start, end + 1) # TODO: Use closed-open intervals for change events
```

This may have been removed because no other piece of code was using `@intersectVisibleRowRange` anymore, but it probably pointed out the intention of making `@intersectsVisibleRowRange` parameterizable.

We may still want to wait for @nathansobo's reply, but (as stated earlier on) I think the default behavior should take into account the closed interval only. After we clarify this once for all, I think that writing some new tests is mandatory and I can take care of it. :white_check_mark: 

@izuzak: :+1: for your gist, it looks super! :rocket:

nathansobo:
I'm fine changing it to use closed intervals, since that's the approach we use elsewhere. I'm still not clear how this is exposed to the public API however...

> In turn, intersectsVisibleRowRange is being used by a wrapper which exposes it to the public API

Which wrapper? There is an `intersectsVisibleRowRange` method on `TextEditor` but it isn't public. I think anything dealing with what is "visible" should be moved to the view, because this information is only available once the editor is actually attached to the DOM. Exposing it in the model invites a situation where package authors call these methods in a scenario where the model isn't associated with an attached view.

as-cii:
> I'm fine changing it to use closed intervals, since that's the approach we use elsewhere.

Great, tomorrow I can write some new tests along with a brand new implementation of `intersectsVisibleRowRange` :+1: 

@izuzak: maybe you want to start off a new branch with the implementation you proposed and I simply add missing tests?

@neojski: you originally opened the PR, therefore if you're interested in continuing it, I could just provide some guidance. Otherwise, feel free to close this and I'll take care of it :blush: 

> Which wrapper? There is an intersectsVisibleRowRange method on TextEditor but it isn't public.

You're correct: the wrapper is not part of the public API, but I couldn't see any usage of it inside atom and  I assumed it was there to expose `intersectsVisibleRowRange`.

> I think anything dealing with what is "visible" should be moved to the view, because this information is only available once the editor is actually attached to the DOM. Exposing it in the model invites a situation where package authors call these methods in a scenario where the model isn't associated with an attached view.

I feel like this is the right path :ok_hand: However, it seems like `intersectsVisibleRowRange` [was already inside `EditorComponent` in the past](https://github.com/atom/atom/commit/e365e51a2bac05beaa4f1614420475fef84ac803#diff-c74ccbac6e5951e339cf3b8fb0af1cf8R116), but it was moved to `DisplayBuffer` afterwards. Any ideas why that happened?

Thanks everyone! :bow:

nathansobo:
> Any ideas why that happened?

I don't really remember. For a while, my thinking was to explicitly model various things related to screen dimensions in the model layer. Upon further thought, I think a better place for all of that logic is in the presenter to keep the model as pure as possible. However, some discussion of pixels can't be avoided in the model layer, such as information about characters widths for soft-wrap. I'd just like to channel public usage of anything that depends on it through the view to avoid confusion when the model is used independently of the view.

neojski:
@as-cii, I can work on it if it's not too complicated. It's my first atom patch so I don't want to do any big refactor. So, what do we do?

izuzak:
> maybe you want to start off a new branch with the implementation you proposed and I simply add missing tests?

@as-cii I'd be :+1: if you or @neojski want to work on this, either in this PR or a new one, using whatever implementation you think is best. I was just trying to push this issue forward and figure out where the problem is (especially because there are other issues blocking on this).

My wishlist would be:
* fix the bug
* add tests so that Future Atom Contributor doesn't break this behavior
* document the method so that Future Atom Contributor is less confused about expected behavior

We could refactor things in another PR, it doesn't necessarily need to be this one (and it would probably allow us to :ship: this fix sooner).

nathansobo:
> We could refactor things in another PR, it doesn't necessarily need to be this one (and it would probably allow us to :ship: this fix sooner).

Agreed. I think I muddied the waters a bit by dragging in the issue about where the public method should be. As @izuzak says, let's just unblock this and then deal with the API. I would like it to be sooner rather than later if people are using this though so we aren't promoting people using a private API.

PR #4602, opened 2014-12-13
Add dark-theme for gtk+3

ardeshirj:
In Ubuntu, the default theme is dark gray, so it is a nice match for the Atom dark default theme. However, in gnome-3 the default theme is white and it kind of contradicts with Atom dark theme. Here is how it looks:

atom-white-theme

Since atom-shell is now supporting [dark theme in gtk3+](https://github.com/atom/atom-shell/blob/master/docs/api/browser-window.md#class-browserwindow) I thought it would be nice to have the dark theme by default, however there is an open issue [#765](https://github.com/atom/atom-shell/issues/765) in atom-shell where the menu bar will still be white while the top bar is black. Here is how it looks:

atom-gnome-dark-1

I am going to leave this here so we can review it, and once the issue is resolved, you may want to merge it.

Thank you.

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -30,6 +30,7 @@ class AtomWindow
       'web-preferences':
         'direct-write': false
         'subpixel-font-scaling': false
+      'dark-theme': true
     global.atomApplication.addWindow(this)
 
     @handleEvents()


KeitIG:
Any news on this ? I'm really missing this "feature"

mnquintana:
This is still blocked on atom/electron#765 unfortunately :cry:

ai:
@mnquintana this issue is blocked only because we will have black title bar with white menu bar?

I hide menu bar abd this is why atom/electron#765 is not a problem for me, but I really want this PR in Atom?

Maybe we can merge it and enable black title bar by config option?

mnquintana:
@ai I'd love to get feedback from more Linux users before merging this, just to make sure that this is an okay tradeoff for now.

varemenos:
I'm also hiding the menu bar cause it's distracting but I don't see how this issue is blocked by the black menu issue. I mean it's better to have a black title bar with a white menu than a white title bar and a white menu.

Of course this is my opinion, not a rule :smile:

lee-dohm:
I'm going through old PRs to see if they can be salvaged or should be closed. Is this PR something that we want to push through? I think it would be best to not change the defaults but have this as an option if it isn't already.

stefansedich:
I just installed Linux on my laptop last night and this was the first thing that anoyed me when I switched to the dark gnome theme.

If I could configure it I would be happy.

damieng:
No response.

Issue #4138, opened 2014-11-10
Make project.getRepositories pluggable to support VCS other than Git


benad:
Right now, Project populates "@repo" only with the result GitRepository.open. There is no way to make Atom support other VCS like Subversion or Mercurial from a plugin because of that. Also, subclassing GitRepository is difficult because it does far more than simply forwarding Git status information, for example subscribing to window focus changes in the constructor.

There should be a simple, generic base class for VCS support, with a mean for Project to ask each VCS module if the current project has any repository for its directory, and a clean API to refresh any item status.

iorlas:
So?

lmcardle:
+1

tynril:
+1

Emnalyeriar:
+1

rodrigopmatias:
+1

ergo:
+1

diegomarquez:
+1

batisteo:
+1
1.0-roadmap please

nistvan86:
+1
Please merge this. I need Mercurial integration :)

andischerer:
Actually this is already possible by using the `atom.repository-provider`. Today i've released a package in early dev state that supports svn.
https://github.com/andischerer/atom-svn

BlaM:
There seems to be https://atom.io/packages/nuclide-hg-repository - but as far as I can tell that package - at least with Windows - does not work.

josser:
@BlaM You have to install nuclide-file-tree to get it work

BlaM:
I tried to use nuclide-installer for a full installation, but that didn't do anything. Not an issue of atom, though, so we shouldn't discuss that here. I just wanted to point users watching this issue to a possible solution.

bolinfest:
We upstreamed https://github.com/atom/atom/pull/5491 to Atom so that we could support Mercurial (via https://atom.io/packages/nuclide-hg-repository) as part of Nuclide. If anyone else wants to add support for another version control system in Atom, I recommend following the pattern we set forth in `nuclide-hg-repository`.

Since the necessary API is there, I am going to close this issue.

PR #4624, opened 2014-12-15
Add a hook so packages can support more version control systems

oclbdk:
The project API currently only supports Git. This diff makes it possible for 3rd-party packages to add support for other version control systems, addressing issue https://github.com/atom/atom/issues/4138.

Here's an example of a preliminary package that adds support for Mercurial: https://github.com/oclbdk/vcs-hg

Here's a demo of it in action:

vcs-hg demo


kevinsawicki:
Closing this out, I believe this use case was covered by the service added in https://github.com/atom/atom/pull/5491

PR #4731, opened 2014-12-21
add dependency for arch-based systems

NerdDiffer:
for manjaro (based on arch linux), I was getting this error after building & install atom: `/usr/local/share/atom/atom: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory`

after running `sudo pacman -S gconf`, i could open atom. hope this is of use to others.

### DIFF
--- a/docs/build-instructions/linux.md
+++ b/docs/build-instructions/linux.md
@@ -26,7 +26,7 @@ Ubuntu LTS 12.04 64-bit is the recommended platform.
 
 ### Arch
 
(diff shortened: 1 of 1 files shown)


Anachron:
@kevinsawicki got this on Ubuntu 14.04:
```
~ » atom                                                           eos@eos-gaz9
/opt/atom/atom: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory
```

PR #4733, opened 2014-12-22
WIP: Add a way to duplicate selected text instead of lines only

dsandstrom:
This is a feature I miss from another text editor.
1. When the selection is only on one line, `ctrl-shift-d` duplicates the text to the right
2. When the selection takes up more than one line, `ctrl-shift-d` duplicates text down

Right now you have `editor.duplicateLines()` to accomplish 2.  But, I would like to add support for 1.  I added a function `editor.duplicateSelections()` instead of adding to the current function because I wanted to get feedback and get something working.


dsandstrom:
Going to close this for now.  There is a package that does something similar: https://github.com/nick/duplicate-line-or-selection. I will work on it more if there are people you want this.

PR #4781, opened 2014-12-27
Update scroll position when opening from command line with line specified

### LINKED ISSUE #4780: Scroll isn't updated when using a line number from the command line
Currently when you open a file from the command line with a line number, i.e.:

    atom ~/cool-thing.js:42

there is a cursor set, but the editor is still scrolled to the top.  Ideally the editor should open up scrolled to the location of the cursor.

ultramiraculous:
Attempt to fix #4780.  

The event listener solution is probably hacky.  I mostly don't quite know when the DisplayBuffer is fully "initialized"/has a height, width, etc.  As of right now if I attempt to call @scrollToCursorPosition() in TextEditor's constructor (at say line 105), the DisplayBuffer has no size information when it gets to scrollToScreenRange() and no scrolling occurs.

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -102,6 +102,9 @@ class TextEditor extends Model
       initialLine = Math.max(parseInt(initialLine) or 0, 0)
       initialColumn = Math.max(parseInt(initialColumn) or 0, 0)
       @addCursorAtBufferPosition([initialLine, initialColumn])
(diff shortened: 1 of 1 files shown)


izuzak:
#4780 was closed as we can't reproduce the problem anymore, so I'm closing this PR as well. If you can still reproduce the problem -- please post the steps in the issue and we can start from there. Thanks again!

Issue #3923, opened 2014-10-22
Update the app icon on OS X to fit with the new Yosemite UI style


brandonweiss:

brandonweiss:
![fark_4glikgpuzfc4wtk5h1urhsbq0n0](https://cloud.githubusercontent.com/assets/4727/5674062/8e027398-9759-11e4-8b65-8692d21b2b70.jpg)

It's beautiful...

kevinsawicki:
:tada:

PR #3089, opened 2014-07-25
Update Mac icon to be flatter, to work with Yosemite styling

ParkerK:
Yosemite has a much flatter look to it, and all of the default Mac apps have had their icons changed. I updated the Atom icon to fit in with these theme a little better. Any feedback would be appreciated :)

Old: atom
New: f2b4523d650ae78d-64x64

PR #4807, opened 2014-12-30
Icon refresh

simurai:
:warning: Don't merge yet. Still needs legal clearance.

Replaces the Atom icon with a "Yosemite" version. Issue #3923

Current

screen shot 2014-12-30 at 11 10 27 pm

New "Yosemite" version

screen shot 2014-12-30 at 11 11 04 pm

In context

screen shot 2014-12-30 at 11 48 21 pm

The new version doesn't necessarily need to have the white border etc. Makes it fit in, but it might seem too close to a stock app from Apple? /cc @cameronmcefee
- [x] Mac
- [x] File icons
- [x] Win
- [x] Linux


wcandillon:
Looks amazing.

kevinsawicki:
> Another thing to think about.. should the Win and Linux version be different? 

I think using the same icon is fine, GitHub for Windows and GitHub for Mac use the same icon.

jeejkang:
:+1: lookin great! @simurai have you tried increasing the stroke ever so slightly?? The icon itself seems little thin next to the thick white border.

benogle:
It would be nice to fit into the metro aesthetic as well. Windows seems to have a super flat, single-color app icons (in the 'start' menu anyway). I dont know all the possibilities for windows, so maybe the icon on the desktop would be the same, but anywhere the metro look is used, IMO, we should go that route.

benogle:
Of course, that can be done in a separate pull, just responding to:

> should the Win and Linux version be different

paulcbetts:
@benogle You generally only single-color outline icons if they are Metro apps - even the built-in desktop apps have color:

cameronmcefee:
We have the :+1: to move forward re IP. Carry on, and thanks for postponing.

cameronmcefee:
Oh, also please ping me when you have finalized assets so I can do the necessary documenting.

benogle:
@paulcbetts noted

simurai:
@paulcbetts @benogle @kevinsawicki For this PR we could have the same icon everywhere and then as a seperate PR look into adding a "metro" tile: http://msdn.microsoft.com/en-us/library/windows/apps/hh465403.aspx#default_tiles But not sure how to add that to Atom's build script. It probably needs to be a separate file (+ manifest?) and not part of the current `atom.ico`.

paulcbetts:
@simurai Metro tiles only apply to Metro apps, Atom's would look different; there is no way to have a "metro-only" icon for a Desktop app. But really, it's fiiiiiiiine - just use this icon everywhere, it looks great!

simurai:
@paulcbetts Ok, I see. Thanks.

simurai:
@jeejkang Do you still have the original icon in vector format? I used this [SVG file](http://commons.wikimedia.org/wiki/File:Atom_editor_logo.svg) to create all the different sizes.

screen shot 2015-01-06 at 11 16 31 am

For the 3 smallest sizes, the thickness is increased a bit, but the rest is just scaled based on that SVG. I can increase it for the others too. It might also appear thinner because the new one has a dark drop-shadow instead of a glow?

jeejkang:
@simurai I uploaded the Atom file in this temporary [Dropbox folder](https://www.dropbox.com/s/p0io1hp9786carf/atom-icon.ai?dl=0). Please let me know if you have any trouble opening the file.

simurai:
Note to self if it becomes an issue: Used [Icon Slate](http://www.kodlian.com/apps/icon-slate) to create the Windows `atom.ico` file. For some reason it moves the `256px` version to the end and in the finder it shows up as a `128px` file. The reason might be to fix an XP issue. From this [comment](http://stackoverflow.com/a/3244679):

> Just a note about Windows XP compatibility: If you reuse the icon as window icon, then note that this can crash your application if you use a compressed 256 icon. The solution is to either not compress the icon or create a second version without the (compressed) 256 icon.

simurai:
Ok, here all icons in an overview:

Mac

screen shot 2015-01-08 at 2 39 01 pm

screen shot 2015-01-08 at 2 40 40 pm

Windows

screen shot 2015-01-08 at 2 37 24 pm

Linux

atom


Maybe a last point of discussion: The smallest size (+ 2x) of the file icon is filled with green instead of the circle in the middle. Because it's so small, you wouldn't really recognize the white part anyways.

screen shot 2015-01-08 at 2 45 45 pm

screen shot 2015-01-08 at 2 57 02 pm

If nobody hates it, I'll ping the legal gatekeepers for a last check.

lee-dohm:
:+1:

philipgiuliani:
It looks amazing! I really like the green file icons!

cameronmcefee:
Looks great @simurai. There shouldn't be any legal issues with this change since it doesn't modify the logo itself.

kevinsawicki:
@talniv can you confirm these changes before we merge this PR? Thanks.

talniv:
:+1:

kevinsawicki:
This will :ship: in 0.169, thanks @simurai

philipgiuliani:
Awesome!! :)

jeejkang:
awesome work @simurai!! I :green_heart: the green folder direction.

simurai:
@timgthomas made a good point on twitter:

> @AtomEditor I love it visually, but doesn’t it clash with OS X’s metaphors? Circles for consuming content, tilted rectangles for producing?

I wasn't aware but now when looking at my dock I totally see it. An argument to use the circle is to stay as close as possible to the previous one, so that people still feel it's the same. But I don't mind doing another round with a square version. Maybe as part of the 1.0 launch.

cameronmcefee:
> Maybe as part of the 1.0 launch.

:+1: That's my vote if you choose to change again.

garethbjohnson:
:+1: I love it.

cameronmcefee:
This just popped up on my dock today and looks great. Nice work @simurai!

ParkerK:
@benogle would something like this work?
https://github.com/atom/atom/pull/3089#issuecomment-50837759

simurai:
@ParkerK Are you asking if it's possible to have a "metro" icon on Windows? Since this PR is already merged, do you mind creating a new issue about it.

Issue #4834, opened 2015-01-03
Make Find UI position lockable


mdeck:
The Find UI is a bar across the bottom of the editor.  I also have a linter which places a bar across the bottom.  Depending on what line the cursor is on, the linter output will contain 0+ lines, so the linter bar height varies depending on the line.

When I search for a term, I click the 'Find' button.  The cursor moves to the next found instance, and the linter height changes, moving the 'Find' button away from my mouse cursor.  In effect, I have to look down at the mouse cursor and point it at the 'Find' button every time I want to find the next instance of the search term.  

It is common to browse code looking for instances of a search term, and this seems like a simple fix to make usable.  An option to affix the Find bar at the very bottom would do the trick.

lee-dohm:
There is API support for choosing what order panels stack in:

https://atom.io/docs/api/latest/Workspace#instance-addBottomPanel

This can either be fixed by the Find UI giving itself a lower priority, if we think it should always be at the bottom. Or it can be fixed by the linter UI giving itself a higher priority, to ensure that it is always on top relative to the Find UI. Personally, I think updating linter is the right choice since it is variable in size and would cause the same problems with any other panel that decides to place itself at the bottom.

I think the documentation should be updated to reflect this advice though too no matter how this is fixed.

lee-dohm:
Looks like I forgot to close this. Closing now since the outstanding issues were addressed in the above PRs.

PR #4836, opened 2015-01-03
Add priority recommendation for variable-sized panels

lee-dohm:
Applies to #4834 

I figured this was a good piece of advice for package authors that create panel UI. I believe this is the best place to put this documentation. If it should go somewhere else, let me know.

/cc @benogle @maxbrunsfeld


maxbrunsfeld:
:+1:

PR #4858, opened 2015-01-06
:memo: Prevent confusion over project.contains

lee-dohm:
See https://discuss.atom.io/t/testing-for-the-existence-of-a-file/14079

Perhaps this is overkill, but I wanted to prevent future confusion regarding this function. Got some of the language from node-pathwatcher.


benogle:
Thanks man

PR #4877, opened 2015-01-07
:memo: Add section on apm unlink to debugging doc

lee-dohm:
@kevinsawicki mentioned that his first debugging step to check is `apm unlink -a`. I figured since I refer people to the debugging doc all the time, I should add this suggestion.

I put the section where I felt it should go, but I could definitely see that if this is more intended for end users than Atom package hackers, it should go futher down in the document.


kevinsawicki:
:sunglasses:

benogle:
:+1:

lee-dohm:
Thanks guys!

PR #4901, opened 2015-01-07
:memo: Fix URL references for LESS.

danfuzz:
There were two problems:
- In the URL lists, the reference was listed as `[less]`, but in the prose it was listed as `[LESS]`. I made it `[LESS]` in the URL lists, which seems to match the rest of the docs.
- In both cases in the prose, the reference `[LESS]` was followed by a prose parenthetical. Markdown treats these as literal URLs. Adding a quoting backslash before the open paren fixes this.

### DIFF
--- a/docs/creating-a-package.md
+++ b/docs/creating-a-package.md
@@ -123,8 +123,8 @@ like you.
 
 Style sheets for your package should be placed in the _stylesheets_ directory.
 Any style sheets in this directory will be loaded and attached to the DOM when
-your package is activated. Style sheets can be written as CSS or [LESS] \(but
-LESS is recommended).
+your package is activated. Style sheets can be written as CSS or [LESS], but
+LESS is recommended.
 
 Ideally, you won't need much in the way of styling. We've provided a standard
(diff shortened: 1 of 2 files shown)


docs/creating-a-package.md
```diff
@@ -123,7 +123,7 @@ like you.
 
 Style sheets for your package should be placed in the _stylesheets_ directory.
 Any style sheets in this directory will be loaded and attached to the DOM when
-your package is activated. Style sheets can be written as CSS or [LESS] (but
+your package is activated. Style sheets can be written as CSS or [LESS] \(but
```
kevinsawicki: maybe make this `, but LESS is...` and drop the parens completely from that?

Seems like a bold statement like the recommended language can stand on its own.

danfuzz: Just to be clear I was just aiming to fix the technical problem; I don't have an opinion one way or the other on what's actually said there.

I'll go ahead and make the suggested change.

kevinsawicki:
Great, thanks for fixing this.

danfuzz:
Glad to be of service!

PR #4911, opened 2015-01-08
:memo: Fix PackageManager::onDidLoadAll description

postcasio:
It invokes the callback when packages are loaded, not activated.


kevinsawicki:
Thanks!

PR #4930, opened 2015-01-08
Add color support to config schema

kevinsawicki:
This adds first-class support for colors in the config schema.

``` coffee
atom.config.set('my-package.aColor', 'khaki')
atom.config.get('my-package.aColor') # {red: 240, green: 230, blue: 140, alpha: 1}
atom.config.get('my-package.aColor').toHexString() # '#f0e68c'
atom.config.get('my-package.aColor').toRGBAString() # 'rgba(240, 230, 140, 1)'

atom.config.set('my-package.bColor', 'rgba(10,20, 30, .75)')
atom.config.get('my-package.bColor') # {red: 10, green: 20, blue: 30, alpha: .75}
```

This will be used to make themes adjustable by allowing colors to be changed in the settings view and those color changes will be reflected in Less variables in the loaded theme stylesheets.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1291,7 +1291,7 @@ describe "Config", ->
         beforeEach ->
           schema =
             type: 'color'
-            default: true
+            default: 'white'
           atom.config.setSchema('foo.bar.aColor', schema)
 
         it 'coerces various types to a color object', ->
@@ -1305,6 +1305,13 @@ describe "Config", ->
           expect(atom.config.get('foo.bar.aColor')).toEqual {red: 1, green: 2, blue: 3, alpha: 1}
           atom.config.set('foo.bar.aColor', 'rgba(4,5,6,.7)')
           expect(atom.config.get('foo.bar.aColor')).toEqual {red: 4, green: 5, blue: 6, alpha: .7}
+
+        it 'reverts back to the default value when undefined is passed to set', ->
+          atom.config.set('foo.bar.aColor', 'rgb(255,255,255)')
+          expect(atom.config.get('foo.bar.aColor')).toEqual {red: 255, green: 255, blue: 255, alpha: 1}
+
+          atom.config.set('foo.bar.aColor', undefined)
+          expect(atom.config.get('foo.bar.aColor')).toEqual {red: 255, green: 255, blue: 255, alpha: 1}
 
       describe 'when the `enum` key is used', ->
         beforeEach ->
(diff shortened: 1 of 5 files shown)


spec/config-spec.coffee
```diff
@@ -1287,6 +1290,54 @@ describe "Config", ->
...
+      describe 'when the value has a "color" type', ->
+        beforeEach ->
+          schema =
+            type: 'color'
+            default: 'white'
+          atom.config.setSchema('foo.bar.aColor', schema)
+
+        it 'coerces various types to a color object', ->
+          atom.config.set('foo.bar.aColor', 'red')
+          expect(atom.config.get('foo.bar.aColor')).toEqual {red: 255, green: 0, blue: 0, alpha: 1}
```
benogle: I imagine the 99% case is that someone will want conversion to hex or rgba string. As a consumer of the api, it sucks that you need to manually create the string to actually use it.

Can provide an object that has a `toString` method? Is that a stick in the spokes for serialization. I think you could define a `toJSON` method on that object to get around any issues.

kevinsawicki: > I imagine the 99% case is that someone will want conversion to hex or rgba string

Yeah, I went back and forth about how much to put in this object initially.

I've update it to use a `Color` class with `toHexString` and `toRGBAString` methods on it.

benogle: Some tests for the toString methods?

benogle: Also :+1:

src/color.coffee
```diff
@@ -0,0 +1,46 @@
+_ = require 'underscore-plus'
+ParsedColor = require 'color'
+
+# Public: A simple Color class returned from `atom.config.get` when the value at
```
benogle: Essential? Extended? I suppose Essential.

kevinsawicki: Whoops, forgot that was the new jam, yeah, I could go either way, but Essential sounds good.

src/color.coffee
```diff
@@ -0,0 +1,88 @@
...
+
+    try
+      parsedColor = new ParsedColor(value)
+    catch error
+      return null
+
+    new Color(parsedColor.red(), parsedColor.green(), parsedColor.blue(), parsedColor.alpha())
+
+  constructor: (red, green, blue, alpha) ->
+    Object.defineProperties this,
```
kevinsawicki: This allows `JSON.stringify` to still work on `Color` objects while enforcing that color values are always valid.

kevinsawicki:
Any other feedback @benogle?

benogle:
:boat:

kevinsawicki:
This was held up for a bit because of some `CSON.stringify` issues discovered that were fixed in #4971

maxbrunsfeld:
:beers:

PR #4955, opened 2015-01-09
"Style sheet" is two words – Treat it that way in the API

nathansobo:
According to the W3C, ["style sheet" is two words](http://www.w3.org/TR/html401/present/styles.html).

I already updated some of the API to treat it that way a few weeks ago, but I forgot about the package directory layout and the `package.json` keys. This finishes the job. I opted for a one-word directory name, `styles`. The `package.json` keys use camel case.
- Package directory format:
  - `stylesheets/` => `styles/`
- Package metadata:
  - `stylesheets` => `styleSheets`
  - `stylesheetMain` => `mainStyleSheet`

This will unfortunately require another mindless upgrade pass after the next release. I will happily do it though to have correct grammar in our APIs.


benogle:
This seems like it's going to generate a lot of work. Do we really want to change it? I would be inclined to leave it the way it is. Conventionally, I feel like stylesheet is really interchangeable with style sheet. Also `styleSheet` feels really weird.

nathansobo:
I would really like it to be grammatically correct. I can do the work. It will probably take 30 minutes. Everyone else is already going to have deprecations anyway so it won't represent a large incremental difference in effort.

nathansobo:
I talked to @benogle offline and coaxed him to let me make this change. Three cheers for grammar.

mnquintana:
FWIW, I don't really get this change all of a sudden - that W3C doc also uses caps-locked HTML tag names exclusively - does that mean all of Atom's views should also have caps-locked element names? ¯\_(ツ)_/¯

mark-hahn:
There are way too many of these examples to justify trying to fix them.
What about `checkBox`?  We are writing code here.  Names are often bastard
versions of english.

nathansobo:
It's a pretty minor change for a given package and makes the API internally consistent. Obviously this stuff is all subjective but we're in the final stages of spit-polishing the API and this bugged me. I'm hoping today will be the last day of changes before we proceed.

mark-hahn:
>  I'm hoping today will be the last day of changes

You are freezing for 1.0?

benogle:
Yeah, for the API freeze

PR #4971, opened 2015-01-10
Stop eval'ing .cson files

kevinsawicki:
This pull request upgrades the [season](https://github.com/atom/season) version to `5.0` which uses the latest version of [cson-safe](https://github.com/groupon/cson-safe).

The main upside here is that `cson-safe` doesn't eval the files but instead parses them safely just like `JSON.parse` does.

This will break any `.cson` files that contain code since they will no longer parse correctly.  This is a good thing since the expectations of CSON are the same as JSON, they parse to plain old JavaScript objects with arrays, string, numbers, and booleans.

This PR also upgrades to CoffeeScript 1.8 across the board for `.cson` files and `.coffee` files.
# Performance Differences

Ran 100 calls of `CSON.readFileSync` and `CSON.stringify` against  https://github.com/atom/atom/blob/6cc1ffa7935ea013e822ddab06b7897320313016/keymaps/darwin.cson

Prior version used (1.0.2): ~2600 milliseconds
New version used (5.0.0):  ~600ms

### DIFF
--- a/build/package.json
+++ b/build/package.json
@@ -18,7 +18,7 @@
     "grunt-contrib-coffee": "~0.9.0",
     "grunt-contrib-csslint": "~0.1.2",
     "grunt-contrib-less": "~0.8.0",
-    "grunt-cson": "0.10.0",
+    "grunt-cson": "0.11.0",
     "grunt-download-atom-shell": "~0.10.0",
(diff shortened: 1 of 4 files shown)


lee-dohm:
:clap:

PR #4973, opened 2015-01-10
Log scopes to notification instead of console

### LINKED ISSUE #4969: Post an info notification in response to editor:log-cursor-scope
See: https://discuss.atom.io/t/how-to-see-what-syntax-markup-is-being-used/14172

Needing the console open inhibits the discoverability of this *very* useful command. If it popped up a notification, some Sublime users might try the key combination out of habit (I know I did when I first started using Atom) and get a little warm fuzzy out of it. (Besides putting notifications to use in a situation other than error logging :grinning:)

lee-dohm:
Fixes #4969

See: https://discuss.atom.io/t/how-to-see-what-syntax-markup-is-being-used/14172

screen shot 2015-01-09 at 4 31 11 pm

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -2522,7 +2522,11 @@ class TextEditor extends Model
 
   logCursorScope: ->
     scopeDescriptor = @getLastCursor().getScopeDescriptor()
-    console.log scopeDescriptor.scopes, scopeDescriptor
+    list = scopeDescriptor.scopes.toString().split(',')
+    list = _.map list, (item) -> "* #{item}"
+    content = "Scopes at Cursor\n#{list.join('\n')}"
+
+    atom.notifications.addInfo(content, dismissable: true)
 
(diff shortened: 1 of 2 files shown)


src/text-editor.coffee
```diff
@@ -2522,7 +2522,11 @@ class TextEditor extends Model
 
   logCursorScope: ->
     scopeDescriptor = @getLastCursor().getScopeDescriptor()
-    console.log scopeDescriptor.scopes, scopeDescriptor
+    list = scopeDescriptor.scopes.toString().split(',')
+    list = _.map list, (item) -> "* #{item}"
```
kevinsawicki: Really minor, but you could just call `list.map` instead of `_.map`

lee-dohm: Thanks! Momentito por favor :grinning:

kevinsawicki:
:ship: Very nice!

PR #4979, opened 2015-01-10
Adding link to blog post with Travis CI instructions

DinisCruz:
Here is an example of adding it to an Package https://github.com/jacogr/atom-lcov-info/issues/18


benogle:
Dope. Thanks.

PR #5044, opened 2015-01-14
Allow scoped default File Encoding

russlescai:
This pull request allows custom configuration for file encoding based on scope descriptors.

For example (my original issue), setting powershell files to UTF16-BE:

```
".powershell":
  core:
    fileEncoding: "utf16be"
```

I have created a package to do this (https://atom.io/packages/default-encoding) which works, however I added this pull request if you guys think it will be useful.

Note this is not in lieu of auto detection, just another way to customise atom.

Thanks

### DIFF
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -266,7 +266,6 @@ class Project extends Model
   # Still needed when deserializing a tokenized buffer
   buildBufferSync: (absoluteFilePath) ->
     buffer = new TextBuffer({filePath: absoluteFilePath})
-    buffer.setEncoding(atom.config.get('core.fileEncoding'))
     @addBuffer(buffer)
     buffer.loadSync()
     buffer
@@ -282,7 +281,6 @@ class Project extends Model
       throw new Error("Atom can only handle files < 2MB for now.")
 
     buffer = new TextBuffer({filePath: absoluteFilePath})
-    buffer.setEncoding(atom.config.get('core.fileEncoding'))
     @addBuffer(buffer)
     buffer.load()
       .then((buffer) -> buffer)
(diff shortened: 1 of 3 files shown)


src/project.coffee
```diff
@@ -311,6 +309,7 @@ class Project extends Model
 
   buildEditorForBuffer: (buffer, editorOptions) ->
     editor = new TextEditor(_.extend({buffer, registerEditor: true}, editorOptions))
+    editor.setEncoding(atom.config.get(editor.getRootScopeDescriptor(), "core.fileEncoding"))
```
kevinsawicki: This should probably go in the `TextEditor` constructor now so that no matter how it is created, it is setting the encoding from the root scope descriptor.

kevinsawicki:
Yeah, this sounds good to me, would you mind adding a spec for this new behavior?

russlescai:
Ok thanks, yep I'll move it to the constructor of text editor and add a spec.

russlescai:
I have added a spec and changed the config get/set to use options (to remove deprecated message).

src/text-editor.coffee
```diff
@@ -105,6 +105,8 @@ class TextEditor extends Model
 
     @languageMode = new LanguageMode(this)
 
+    @setEncoding(atom.config.get("core.fileEncoding", { scope: @getRootScopeDescriptor() }))
```
kevinsawicki: The `{}` can be dropped.

russlescai:
thanks @kevinsawicki I have removed the braces.

kevinsawicki:
Thanks for adding this.

russlescai:
No worries, glad to help :)

Issue #4811, opened 2014-12-30
Window icon is pixelated


GameplayJDK:
Status: waiting for the next release of atom shell.

The icon of the editor windows seems to be pixelated, at least on windows. (Atom v0.165.0; Windows 8.1)

![issue00](https://cloud.githubusercontent.com/assets/4069263/5584108/51b27cb6-9084-11e4-9409-27b1b07bfd4f.png)

jimgraham:
This seems to be an issue with which icon is referenced in the main window. The `atom.exe` has a 16x16 scaled icon, and that one is used in explorer and for short cuts

![atomicon2](https://cloud.githubusercontent.com/assets/430293/5620877/18c83494-94fe-11e4-93a4-18c21a19562f.png)

However, it does not appear to be the one used in the main window -- the main window seems to use a larger icon that is reduced to 16x16 by the OS. One can see that by using the "Magnify" tool -- the icon looks great when magnified up.

![atomicon](https://cloud.githubusercontent.com/assets/430293/5620876/16ac5564-94fe-11e4-8767-8cc27487eaaa.png)

simurai:
From this [comment](http://stackoverflow.com/questions/3236115/which-icon-sizes-should-my-windows-applications-icon-include#comment11003729_3244679):

> I did some more testing. For display on the desktop, when I set it to 144dpi ("150%") on Windows 7, for the smallest possible icon it used the 24px icon (if it is there, otherwise uprezzed 16px).

So adding a `24px` version might fix it. I'll try that part of #4807.

jimgraham:
I don't think the issue is with the embedded icon. I think that creating the atom window sets the icon to a resource PNG.

https://github.com/atom/atom/blob/master/src/browser/atom-window.coffee#L13

``` coffee
module.exports =
class AtomWindow
  _.extend @prototype, EventEmitter.prototype

  @iconPath: path.resolve(__dirname, '..', '..', 'resources', 'atom.png')
  @includeShellLoadTime: true

  browserWindow: null
  loaded: null
  isSpec: null

  constructor: (settings={}) ->
    {@resourcePath, pathToOpen, initialLine, initialColumn, @isSpec, @exitWhenDone, @safeMode, @devMode} = settings

    # Normalize to make sure drive letter case is consistent on Windows
    @resourcePath = path.normalize(@resourcePath) if @resourcePath

    @browserWindow = new BrowserWindow
      show: false
      title: 'Atom'
      icon: @constructor.iconPath
[...]

PR #5047, opened 2015-01-14
Replace pixelated BrowserWindow icon

simurai:
**Update II** We could wait with this since it might get fixed in Atom Shell: https://github.com/atom/atom-shell/issues/863

**Update**: It now only adds a `32px` icon, see https://github.com/atom/atom/pull/5047#issuecomment-69874810

This PR adds icons as `png` in multiple resolutions. See https://github.com/atom/atom/issues/4811#issuecomment-69702057

I tested it locally on Windows 8.1. Not by making a new build, but just adding it manually. It seemed to work, but I only have a non hi-res screen. So not sure if it will pick up the 1.5x or 2x version for higher DPIs.


paulcbetts:
@simurai Lemme have a look

paulcbetts:
Hm, no dice, though I think it's definitely better:

simurai:
:disappointed:  but thanks for testing.

I found another strange thing. I had Atom pinned in the dock. It looked fine:

capture2

But when un-pinned and opening Atom again, it picked the `16px` icon.

capture

Pinning again, makes it pick up the `ico` file and it looks fine.

paulcbetts:
I'm not sure if multi-DPI works here - can you just give the 48x48 one a unique name and use it?

simurai:
Yeah, we could.. but then we're back at the beginning that the top left icon in the main Atom window is pixelated.

capture3

simurai:
Just tested with a `32px` icon. Top right looks better, task bar less. Hmm.. maybe the best compromise?

capture4

Once you pin the icon, then it looks nice.

capture5

simurai:
Ok, I changed this PR to only use the `32px` icon. It's not ideal, but maybe the best compromise.

simurai:
Atom Shell is gonna allow to use the `ico` file: https://github.com/atom/atom-shell/issues/863#issuecomment-70197230

Gonna close this for now and make a new PR.

PR #5051, opened 2015-01-14
:memo: fix setHead & setTail docs in Marker

tony612:
Change range -> position


kevinsawicki:
Thanks so much! :ship:

Issue #4099, opened 2014-11-07
Folding broken after splitting


kevinsawicki:
* Open an editor
* Fold a function
* Split the pane
* Unfold the function in the new pane
* Fold something inside the initially folded function
* Gutter shows it as folded but is isn't

![fold-split](https://cloud.githubusercontent.com/assets/671378/4959970/29b5be8e-66be-11e4-978e-2a0c8ceb4569.gif)

PR #5060, opened 2015-01-14
Fix folds with splits

maxbrunsfeld:
The first commit fixes an error that caused #4099; we were accidentally comparing folds from two different `DisplayBuffer`s, on the same `TextBuffer`.

 The second commit is a refactor to make it easier to avoid this kind of error. It makes it so that any time a `DisplayBuffer` is used to find or create markers, they are associated with that `DisplayBuffer`.

### DIFF
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -351,10 +351,15 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(1).text).toMatch /^10/
 
       describe "when there is another display buffer pointing to the same buffer", ->
-        it "does not create folds in the other display buffer", ->
+        it "does not consider folds to be nested inside of folds from the other display buffer", ->
           otherDisplayBuffer = new DisplayBuffer({buffer, tabLength})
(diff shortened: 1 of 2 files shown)


src/display-buffer.coffee
```diff
@@ -952,7 +953,7 @@ class DisplayBuffer extends Model
   #
   # Returns an {Array} of existing {Marker}s.
   getMarkers: ->
-    @buffer.getMarkers().map ({id}) => @getMarker(id)
+    @findMarkers()
```
maxbrunsfeld: @nathansobo does this seem right to you?

nathansobo: Yeah seems reasonable.

nathansobo:
Only thing I'd suggest is mentioning this explicitly in the API docs, that you'll only find markers created via the editor when you use `TextEditor::findMarkers`, `::getMarkers` etc.

maxbrunsfeld:
On second thought, I don't think we should make the larger change. I had forgotten to look for usages of `TextEditor::decorateMarker`. It turns out that there are several third-party packages that use that method in conjunction with `TextBuffer::mark{Position,Range}`. I'm just going to go with the minimal fix instead.

PR #5066, opened 2015-01-14
Fix getOriginUrl calling itself

adrianlee44:
I see this error when using `git-blame` package which is using the deprecated method `getOriginUrl`. Since it's the method is referring to itself, it's causing `Maximum call stack size` (https://github.com/alexcorre/git-blame/issues/56).


kevinsawicki:
Thanks!

PR #5067, opened 2015-01-14
Don't call deprecated method in infinite loop

dmnd:
See also alexcorre/git-blame#56

Note: I didn't check for other instances of this refactoring error


dmnd:
Uhh, I used the web editor to write this but the diff looks way different. Not sure what happened there - maybe I was editing an old version? Closing.

dmnd:
Oh, I see it was fixed in #5066

PR #5068, opened 2015-01-15
Rename 'gutter' decorations to 'line-number'

nathansobo:
We're talking about adding APIs for multiple gutters, so the name 'gutter' for these decorations doesn't really make sense. What we're actually decorating is the _line numbers_, which will be just one of the potential gutters on an atom editor.


nathansobo:
The flaky markdown-preview test is also on master. Going to merge this and address the flakiness tomorrow so everything is baked for 1.0 stability.

PR #5110, opened 2015-01-16
Add ._* to ignored file names

amayer5125:
Macs generate ._ files when saving to a filesystem other then HFS
In shared environments this is a big problem

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -9,7 +9,7 @@ module.exports =
     properties:
       ignoredNames:
         type: 'array'
-        default: [".git", ".hg", ".svn", ".DS_Store", "Thumbs.db"]
+        default: [".git", ".hg", ".svn", ".DS_Store", "._*", "Thumbs.db"]
         items:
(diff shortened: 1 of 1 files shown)


izuzak:
I think this is okay. What do you think, @kevinsawicki? 

Some info on these: http://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them.

kevinsawicki:
> What do you think, @kevinsawicki?

I haven't tried it out but it seems useful.

amayer5125:
If/Whenever you would like to merge this if you let me know I can rebase if you would like

kevinsawicki:
> If/Whenever you would like to merge this if you let me know I can rebase if you would like

Rebasing now would be great, thanks.

amayer5125:
@kevinsawicki I rebased

izuzak:
Just tried this out and seems to work well, so going to :boat: this. Thanks again, @amayer5125.

PR #5152, opened 2015-01-20
Update atom-text-editor selector example

simurai:
Now that `.editor` and `.editor-colors` are replaced with `atom-text-editor`.


nathansobo:
Thanks. Man we really missed these everywhere.

PR #5176, opened 2015-01-20
Upgrade core to space-pen 5.0

kevinsawicki:
Previously core was on space-pen 3.8 which meant it wasn't getting deduped across packages and so we were shipping 28 versions of it in the app.

This also upgrades Atom to jQuery 2.1.3 via https://github.com/atom/space-pen/pull/63

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "semver": "2.2.1",
     "serializable": "^1",
     "service-hub": "^0.2.0",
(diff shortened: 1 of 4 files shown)


kevinsawicki:
Closing this out, this probably can't happen until after 1.0 since space pen 5.0 would remove shims needed currently for backwards compatibility.

PR #5201, opened 2015-01-22
:apple: try ~/Applications before using mdfind

wyand:
When Atom is installed with [Homebrew](https://github.com/Homebrew/homebrew), it will be in `~/Applications` instead of `/Applications`. This change updates the `atom.sh` startup script to check both `/Applications` and `~/Applications` before falling back to a metadata search.

Behavior is unchanged when `$ATOM_PATH` is unset or empty.


kevinsawicki:
Nice, thanks for this

PR #5227, opened 2015-01-22
Add optional package-name argument to run-specs grunt task

maxbrunsfeld:
This way we can do `grunt run-specs:markdown-preview` to debug a single package's specs on CI.

### DIFF
--- a/build/tasks/spec-task.coffee
+++ b/build/tasks/spec-task.coffee
@@ -35,8 +35,9 @@ module.exports = (grunt) ->
       when 'win32'
         path.join(contentsDir, 'atom.exe')
 
-  runPackageSpecs = (callback) ->
-    failedPackages = []
+  runSinglePackageSpecs = (packageName, callback) ->
+    modulesDirectory = path.resolve('node_modules')
+    packagePath = path.join(modulesDirectory, packageName)
     rootDir = grunt.config.get('atom.shellAppDir')
     resourcePath = process.cwd()
     appPath = getAppPath()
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Just curious, does `apm test` not work on CI if you cd into `node_modules/markdown-preview`?

maxbrunsfeld:
I didn't try because I couldn't find `apm`. Looking for it on the CI box now.

maxbrunsfeld:
It just outputs `Tests Failed` for me. Is there a flag I can pass?

kevinsawicki:
It is in `apm/node_modules/.bin/apm` inside the atom repo working directory.

build/tasks/spec-task.coffee
```diff
@@ -106,27 +112,37 @@ module.exports = (grunt) ->
...
-      method = async.series
+    if packageName?
+      runSinglePackageSpecs packageName, (error) ->
+        elapsedTime = Math.round((Date.now() - startTime) / 100) / 10
+        grunt.log.ok("Total spec time: #{elapsedTime}s using #{concurrency} cores")
+        grunt.log.error("[Error]".red + " #{packageName} spec(s) failed") if error?
+        if process.platform is 'win32' and process.env.JANKY_SHA1
+          done()
+        else
+          done(!error?)
```
kevinsawicki: Hmm, I thought `done(error)` would work here, does it need to be coerced into a boolean?

maxbrunsfeld:
I'm guessing I need to pass the path to the `atom` command for it to work.

kevinsawicki:
Yeah, probably need to pass in the atom path

maxbrunsfeld:
So it would seem like the Atom path would be in [a fixed location in `/tmp`](https://github.com/atom/atom/blob/master/build/Gruntfile.coffee#L40), but I'm finding that it's in a random temp directory like `/private/var/folders/rj/620tf4yn6pn9q8hj453s2j3r0000gp`.

Is there a better way to do this than this?

```
../../apm/node_modules/.bin/apm test --path $(find /private/var/folders/ -name Atom 2>/dev/null -print -quit)
```

kevinsawicki:
> Is there a better way to do this than this?

Probably not, you could always build to a consistent directory using `--build-dir` but that is pretty unpleasant too.

maxbrunsfeld:
Ok if the `Atom` binary isn't in a stable location on CI, then this probably won't be that helpful in terms of saving time debugging, since you'd have to run `build` in the same grunt session to be able to find it.

PR #5228, opened 2015-01-23
Make waitsForPromise() work with es6 promises as well as Q promises

benogle:


spec/spec-helper.coffee
```diff
@@ -298,13 +298,13 @@ window.waitsForPromise = (args...) ->
   window.waitsFor timeout, (moveOn) ->
     promise = fn()
     if shouldReject
-      promise.fail(moveOn)
-      promise.done ->
+      promise.catch(moveOn)
+      promise.then ->
```
benogle: `then` has slightly different semantics than `done` on Q promises. `done` just throws any exceptions that go unhandled. Not a big deal cause we are using a `catch` though.

kevinsawicki:
:+1:

PR #5240, opened 2015-01-24
[Docs] Rename LESS to Less

mnquintana:
[Less is now written without caps lock](http://lesscss.org/), so I went through and updated the docs to reflect that.


simurai:
http://sassnotsass.com

mnquintana:
@simurai Heh indeed :smiley:

thedaniel:
:sparkles:

PR #5277, opened 2015-01-27
Allow packages to specify services in their package.json files

### LINKED ISSUE #5165: Make services API public

maxbrunsfeld:
See https://github.com/atom/atom/issues/5165#issuecomment-71685691.
- [x] When activating a package, register its service-consuming methods
- [x] When activating a package, register its service-providing methods
- [x] When _deactivating_ a package, dispose of its service-consuming methods
- [x] When _deactivating_ a package, dispose of its service-providing methods
- [x] When activating a package that depends on a service, should we automatically activate all non-disabled packages that provide that service (see https://github.com/atom/atom/issues/5165#issuecomment-71697306)? I think not, for now.
- [x] Should there be an option for making a service dependency `optional` vs `required`? I'm inclined to not add this yet.
- [x] Should we make the `ServiceHub` private? We did.
- [x] Document usage of new `providedServices` and `consumedServices` fields in package.json

/cc @atom/core

Closes #5165


spec/fixtures/packages/package-with-service-dependencies/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "package-with-service-dependencies",
+
+  "serviceDependencies": {
+    "service-1": {
+      "0.3.1": "handleFirstServiceV3",
```
nathansobo: Worth doing a version *range* here just to be thorough?

maxbrunsfeld:
Currently, I have the nested `versions` field inside of the `serviceDependencies`, in case we want to add extra fields to those objects like `required: true` or `forceActivate: true`, or something.

src/atom.coffee
```diff
@@ -271,7 +267,6 @@ class Atom extends Model
     @contextMenu = new ContextMenuManager({resourcePath, devMode})
     @menu = new MenuManager({resourcePath})
     @clipboard = new Clipboard()
-    @services = new ServiceHub
```
joefitzgerald: Hey @nathansobo I thought the global wasn't going away?

joefitzgerald:
Not LGTM. :-1: Please don't remove the `atom.services` global.

nathansobo:
@joefitzgerald I meant to get back to you about it but you're reviewing faster than I can communicate :smile:.  Max really wasn't on board leaving it in there and it makes sense. It's still available via `atom.packages.serviceHub` which I don't think is too onerous.

joefitzgerald:
Ugh. If you release this, all of autocomplete-plus breaks. Please release it with the global in, and deprecate it for future removal if need be.

joefitzgerald:
Side note: 99% certain on `atom.services` remaining in (admittedly with no promises), then removing it the day we release autocomplete-plus dependent on it is crappy timing.

maxbrunsfeld:
@joefitzgerald I am fine with adding in a deprecated shim for `atom.services` before shipping this PR. @nathansobo are you cool with that?

nathansobo:
@joefitzgerald @maxbrunsfeld Why not just put some conditional code in the package that adapts to both situations? We did that multiple times in bundled packages:

```coffee
services = atom.services ? atom.packages.serviceHub
```

joefitzgerald:
That would assist us in migrating to a different approach.

One other unintended side-effect of forcing this into config-only mode on activation is I am at a loss to understand how I can reduce package load and activation time for autocomplete-plus now. This was to be the next item on the list given that it takes ~250-300 ms to get through both activities.

Service consumption in autocomplete-plus occurs inside `ProviderManager`, which is not in main. How do you propose we have a <10ms  load and activation timeline when service provision occurs immediately following and requires things to be ready to go?

joefitzgerald:
@nathansobo that's fine, but you can't ship a new API and remove the old API completely in the *same version* without breaking things. Give us a version or two to coordinate the provider updates.

joefitzgerald:
> Give us a version or two to coordinate the provider updates.

I'd rather just switch everyone over to the `package.json`technique than have a `services = atom.services ? atom.packages.serviceHub` halfway house. Would you be ok with that?

nathansobo:
> I'd rather just switch everyone over to the `package.json`technique than have a `services = atom.services ? atom.packages.serviceHub` halfway house. Would you be ok with that?

Most definitely. The other situation is for when you need dynamism. I'm okay with a shim for now though if that eases the process.

benogle:
@joefitzgerald You can collect them in the consumption method, then pass them to the ProviderManager or the autocomplete manager on activation, no? Maybe you can save the registration of the providers on a nextTick or something. I assume the provider registration is the slow part as it's dealing with settings and blacklists. 

Really the whole service consumption part of autocomplete plus should be pretty light.  The providers just create an object and provide it, and autocomplete plus just rolls them up in an array.

joefitzgerald:
@benogle yes, it is light currently: https://github.com/atom-community/autocomplete-plus/blob/420e8bb5afbc956641f266d3dce866afd763ec70/lib/provider-manager.coffee#L109-L112

My point is simply that this adds another layer of indirection (now I need to put registration requests into a queue until autocomplete-plus has finished activating which itself will be asynchronous). This will increase complexity, and increase the level of understanding required to successfully write packages that activate quickly.

But if that buys the core team the ability to provide enhanced functionality (ahem - dependent package installation) then so be it.

docs/creating-a-package.md
```diff
@@ -54,6 +54,21 @@ _snippets_ directory are added alphabetically.
 - `activationEvents` (**Optional**): an Array of Strings identifying events that
 trigger your package's activation. You can delay the loading of your package
 until one of these events is triggered.
+- `providedServices` (**Optional**): an Object describing the services that your
+packages provides, which can be used by other packages. The keys of this object
```
kevinsawicki: Should this be `your package provides`?

maxbrunsfeld: Yeah, thanks.

joefitzgerald:
Just for the record, I've completely come around to config-based registration. Wanted to update this thread, given my heavy objections earlier. Keeping the atom.sevvices global (temporarily) will help us coordinate autocomplete-plus provider migration. Many thanks!

maxbrunsfeld:
Thanks for the update @joefitzgerald. Excited to see what you and others on the package-development vanguard do with this.

nathansobo:
Haha, I hadn't hit refresh to see you already merged. Oh well.

joefitzgerald:
So... next feature = using this metadata in apm during install / upgrade to establish package dependencies, right :wink:?

thedaniel:
@joefitzgerald PDI ;)

joefitzgerald:
Hey @maxbrunsfeld what happens if a package has - for example - multiple autocomplete-plus providers, each of which must be provided on the same namespace. Does `providedServices` allow multiple keys for the same namespace? I think not...
```
"providedServices": {
  "my-service": {
    "description": "Does a useful thing"
    "versions": {
      "1.2.3": "provideMyServiceV1",
      "2.3.4": "provideMyServiceV2",
    }
  }
}
```

maxbrunsfeld:
@joefitzgerald I see your point. I think for right now though, the solution to that is to use multiple packages. *Maaybe* down the road we could add something like this -

```json
{
  "providedServices": {
    "my-service": {
      "description": "Does a useful thing",
      "versions": {
        "1.2.3": ["provideMyServiceWithFoo", "provideMyServiceWithBar"]
      }
    }
  }
}
```

- an Array of method names, corresponding to multiple implementations of the same service. Like I said though, we're trying to keep things tight and minimal. If you think this is an immediate concern and is blocking something you're trying to do, I'm happy to hear you out.

joefitzgerald:
There is more than one autocomplete-plus provider package that has multiple providers. So I would flag this as an immediate concern. It's not blocking me, but it'll block them.

maxbrunsfeld:
@joefitzgerald would you mind sharing links to the packages that do this? I'd like to have a look at them and see if there's another solution that would work for them.

joefitzgerald:
Best bet is trolling through https://github.com/atom-community/autocomplete-plus/wiki/Autocomplete-Providers - the reason I'm aware of it is I went through yesterday and updated the entire doc. I recall at least 2 but will need to go look again to find them.

thedaniel:
If it reduces the complexity of our API I'm comfortable telling 2-3 people to do something slightly inconvenient.

joefitzgerald:
Keep in mind we're talking about a sample size of ~10 so it's not an insignificant percentage.

I'm providing feedback as the first user of the new API, but obviously you're not obliged to make changes to the API to accommodate it. Just providing visibility.

maxbrunsfeld:
@joefitzgerald A couple of ways that this problem could be solved with the existing API:

* `autocomplete-plus` could *provide* a service that packages like `autocomplete-haskell` could *consume*. The service could expose a method called `registerCompletionSource` or whatever, and packages could call that as many times as they wanted.

* Packages like `autocomplete-haskell` could provide a single service that *contained* multiple completion-sources. Maybe an object that mapped editor scopes to completion sources:

```coffeescript
{
  ".source.literate.haskell": literateHaskellCompletionSource
  ".source.haskell": ordinaryHaskellCompletionSource
}
```

Anyway, my point is that this problem can be solved in package-space, given the existing primitives provided by Atom core.

joefitzgerald:
Your first suggestion is the way I initially implemented the autocomplete-plus api, and at @nathansobo's direction I inverted the directionality. Since then, I've come to realize this is the correct direction of dependency, and I'm unlikely to reverse it again. We've just gone through multiple breaking API changes and I doubt provider authors will tolerate another.

I do think you should be prescriptive on directionality because it will impact the experience of installing packages when you get to addressing package dependencies.

The second suggestion (as described) seems a bit complicated, but in the abstract it's the best option. I suppose could add a new version of the API with {providers: [provider1, provider2]} in place of {provider: provider1} and then bridge the two in autocomplete-plus in a backwards compatible way. This is the approach I will take.

nathansobo:
As I said when we talked last week, I really think I was wrong for suggesting that you guys invert the API. The dependency really would be clearer if the completion providers depended on a provider registration endpoint service.

joefitzgerald:
I think your initial instinct was correct @nathansobo, and that is manifested in the extreme simplicity with which providers integrate with autocomplete-plus. Inverting the relationship adds boilerplate with no clear benefit. 

If autocomplete-plus needed to provide something to providers, we could also provide a service they consume. But there is no clear need at this time.

nathansobo:
> Inverting the relationship adds boilerplate with no clear benefit.

I'd argue the case of needing to register multiple providers in a single package is a pretty clear benefit. It might not be enough to tip the scales in favor of inverting but you have to acknowledge it's something. We could of course complicate the core API by allowing packages to provide multiple services, but if autocomplete-plus just inverts the relationship back you could programmatically call `add` with every provider in packages. I think that simplicity at the core merits more complexity elsewhere in many cases, though not all cases.

joefitzgerald:
The providing package currently supplies an anonymous object in 1.0.0 of the API:

`{provider: myProvider}`

In 1.1.0 of the API, we will accept:

`{providers: [myProvider]`

And we will support both versions concurrently.

This is by far the simplest solution and stays backward compatible and doesn't cause an upheaval of the API providers yet again. It also has no boilerplate (providers don't need to track a disposable and orchestrate their registration / de registration, as atom handles it for them).

nathansobo:
Sounds reasonable. 🏂

PR #5278, opened 2015-01-27
Tab selection fix

adzenith:
These changes make selecting hard tabs work like it does in other editors. Currently in Atom, if you click in the middle of a hard tab the cursor always goes to the left edge. These changes make it go to the closest edge.

Tests have been updated, and new ones have been added for the new functionality.

I believe I've hit everything in the contribution guidelines. Let me know if I need to make any changes!

### DIFF
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -638,8 +638,11 @@ describe "DisplayBuffer", ->
         expect(displayBuffer.outermostFoldsInBufferRowRange(3, 18)).toEqual [fold1, fold3, fold5]
         expect(displayBuffer.outermostFoldsInBufferRowRange(5, 16)).toEqual [fold3]
 
-  describe "::clipScreenPosition(screenPosition, wrapBeyondNewlines: false, wrapAtSoftNewlines: false, skipAtomicTokens: false)", ->
+  describe "::clipScreenPosition(screenPosition, wrapBeyondNewlines: false, wrapAtSoftNewlines: false, clip: 'closest')", ->
     beforeEach ->
+      tabLength = 4
+
+      displayBuffer.setTabLength(tabLength)
       displayBuffer.setSoftWrapped(true)
       displayBuffer.setEditorWidthInChars(50)
 
@@ -698,19 +701,28 @@ describe "DisplayBuffer", ->
         expect(displayBuffer.clipScreenPosition([3, 58], wrapAtSoftNewlines: true)).toEqual [4, 4]
         expect(displayBuffer.clipScreenPosition([3, 1000], wrapAtSoftNewlines: true)).toEqual [4, 4]
 
-    describe "when skipAtomicTokens is false (the default)", ->
-      it "clips screen positions in the middle of atomic tab characters to the beginning of the character", ->
+    describe "when clip is 'closest' (the default)", ->
+      it "clips screen positions in the middle of atomic tab characters to the closest edge of the character", ->
         buffer.insert([0, 0], '\t')
         expect(displayBuffer.clipScreenPosition([0, 0])).toEqual [0, 0]
         expect(displayBuffer.clipScreenPosition([0, 1])).toEqual [0, 0]
+        expect(displayBuffer.clipScreenPosition([0, 2])).toEqual [0, 0]
+        expect(displayBuffer.clipScreenPosition([0, tabLength-1])).toEqual [0, tabLength]
         expect(displayBuffer.clipScreenPosition([0, tabLength])).toEqual [0, tabLength]
 
-    describe "when skipAtomicTokens is true", ->
+    describe "when clip is 'backward'", ->
+      it "clips screen positions in the middle of atomic tab characters to the beginning of the character", ->
+        buffer.insert([0, 0], '\t')
+        expect(displayBuffer.clipScreenPosition([0, 0], clip: 'backward')).toEqual [0, 0]
+        expect(displayBuffer.clipScreenPosition([0, tabLength-1], clip: 'backward')).toEqual [0, 0]
+        expect(displayBuffer.clipScreenPosition([0, tabLength], clip: 'backward')).toEqual [0, tabLength]
+
+    describe "when clip is 'forward'", ->
       it "clips screen positions in the middle of atomic tab characters to the end of the character", ->
         buffer.insert([0, 0], '\t')
-        expect(displayBuffer.clipScreenPosition([0, 0], skipAtomicTokens: true)).toEqual [0, 0]
-        expect(displayBuffer.clipScreenPosition([0, 1], skipAtomicTokens: true)).toEqual [0, tabLength]
-        expect(displayBuffer.clipScreenPosition([0, tabLength], skipAtomicTokens: true)).toEqual [0, tabLength]
+        expect(displayBuffer.clipScreenPosition([0, 0], clip: 'forward')).toEqual [0, 0]
+        expect(displayBuffer.clipScreenPosition([0, 1], clip: 'forward')).toEqual [0, tabLength]
+        expect(displayBuffer.clipScreenPosition([0, tabLength], clip: 'forward')).toEqual [0, tabLength]
 
   describe "::screenPositionForBufferPosition(bufferPosition, options)", ->
     it "clips the specified buffer position", ->
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -719,6 +719,14 @@ describe "DisplayBuffer", ->
       expect(displayBuffer.screenPositionForBufferPosition([100000, 0])).toEqual [12, 2]
       expect(displayBuffer.screenPositionForBufferPosition([100000, 100000])).toEqual [12, 2]
 
+    it "clips to the (left or right) edge of an atomic token without simply rounding up", ->
+      tabLength = 4
+      displayBuffer.setTabLength(tabLength)
+
+      buffer.insert([0, 0], '\t')
+      expect(displayBuffer.screenPositionForBufferPosition([0, 0])).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 1])).toEqual [0, tabLength]
+
   describe "position translation in the presence of hard tabs", ->
     it "correctly translates positions on either side of a tab", ->
       buffer.setText('\t')
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -739,6 +739,17 @@ describe "DisplayBuffer", ->
       expect(displayBuffer.screenPositionForBufferPosition([0, 0])).toEqual [0, 0]
       expect(displayBuffer.screenPositionForBufferPosition([0, 1])).toEqual [0, tabLength]
 
+    it "clips to the edge closest to the given position when it's inside a soft tab", ->
+      tabLength = 4
+      displayBuffer.setTabLength(tabLength)
+
+      buffer.insert([0, 0], '    ')
+      expect(displayBuffer.screenPositionForBufferPosition([0, 0])).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 1])).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 2])).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 3])).toEqual [0, 4]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 4])).toEqual [0, 4]
+
   describe "position translation in the presence of hard tabs", ->
     it "correctly translates positions on either side of a tab", ->
       buffer.setText('\t')
--- a/src/marker.coffee
+++ b/src/marker.coffee
@@ -286,7 +286,6 @@ class Marker
   # * `screenPosition` The new {Point} to use
   # * `properties` (optional) {Object} properties to associate with the marker.
   setHeadScreenPosition: (screenPosition, properties) ->
-    screenPosition = @displayBuffer.clipScreenPosition(screenPosition, properties)
     @setHeadBufferPosition(@displayBuffer.bufferPositionForScreenPosition(screenPosition, properties))
 
   # Extended: Retrieves the buffer position of the marker's tail.
@@ -313,7 +312,6 @@ class Marker
   # * `screenPosition` The new {Point} to use
   # * `properties` (optional) {Object} properties to associate with the marker.
   setTailScreenPosition: (screenPosition, options) ->
-    screenPosition = @displayBuffer.clipScreenPosition(screenPosition, options)
     @setTailBufferPosition(@displayBuffer.bufferPositionForScreenPosition(screenPosition, options))
 
   # Extended: Returns a {Boolean} indicating whether the marker has a tail.
--- a/src/tokenized-line.coffee
+++ b/src/tokenized-line.coffee
@@ -41,10 +41,20 @@ class TokenizedLine
   copy: ->
     new TokenizedLine({@tokens, @lineEnding, @ruleStack, @startBufferColumn, @fold})
 
+  # This clips a given screen column to a valid column that's within the line
+  # and not in the middle of any atomic tokens.
+  #
+  # column - A {Number} representing the column to clip
+  # options - A hash with the key clip. Valid values for this key:
+  #           'closest' (default): clip to the closest edge of an atomic token.
+  #           'forward': clip to the forward edge.
+  #           'backward': clip to the backward edge.
+  #
+  # Returns a {Number} representing the clipped column.
   clipScreenColumn: (column, options={}) ->
     return 0 if @tokens.length == 0
 
-    { skipAtomicTokens } = options
+    { clip } = options
     column = Math.min(column, @getMaxScreenColumn())
 
     tokenStartColumn = 0
@@ -55,10 +65,15 @@ class TokenizedLine
     if @isColumnInsideSoftWrapIndentation(tokenStartColumn)
       @softWrapIndentationDelta
     else if token.isAtomic and tokenStartColumn < column
-      if skipAtomicTokens
+      if clip == 'forward'
         tokenStartColumn + token.screenDelta
-      else
+      else if clip == 'backward'
         tokenStartColumn
+      else #'closest'
+        if column > tokenStartColumn + (token.screenDelta / 2)
+          tokenStartColumn + token.screenDelta
+        else
+          tokenStartColumn
     else
       column
 
--- a/src/tokenized-line.coffee
+++ b/src/tokenized-line.coffee
@@ -67,7 +67,7 @@ class TokenizedLine
     screenColumn = 0
     currentBufferColumn = 0
     for token in @tokens
-      break if currentBufferColumn > bufferColumn
+      break if currentBufferColumn + token.bufferDelta > bufferColumn
       screenColumn += token.screenDelta
       currentBufferColumn += token.bufferDelta
     @clipScreenColumn(screenColumn + (bufferColumn - currentBufferColumn))
--- a/src/cursor.coffee
+++ b/src/cursor.coffee
@@ -305,7 +305,7 @@ class Cursor extends Model
         columnCount-- # subtract 1 for the row move
 
       column = column - columnCount
-      @setScreenPosition({row, column})
+      @setScreenPosition({row, column}, clip: 'backward')
 
   # Public: Moves the cursor right one screen column.
   #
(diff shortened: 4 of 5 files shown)


50Wliu:
:+1: This will be nice to have implemented :).

nathansobo:
@as-cii Can we lean on you to review this PR? Looks valuable.

as-cii:
> Can we lean on you to review this PR? Looks valuable.

@nathansobo: sure thing! I'll have a look later on today :+1:

as-cii:
@adzenith: first of all, thank you for taking the time to contribute. Nice job here :+1: 

The repository has seen many changes since you opened the pull-request (many of them impact how clipping works), thus it would be awesome if you could merge master into this branch. I have already checked and, apparently, there aren't many conflicts, so it should be a straightforward task.

After merging you'll encounter some errors while running the tests: once again, I feel like fixing them should be quite easy, but I'll be happy to help you along the way if there's any problem.

In the meantime, I'll write some thoughts about your changes so that we can :ship: this! :fire:

src/tokenized-line.coffee
```diff
@@ -47,10 +57,15 @@ class TokenizedLine
...
 
     if token.isAtomic and tokenStartColumn < column
-      if skipAtomicTokens
+      if clipToAtomicTokenEdge == 'right'
         tokenStartColumn + token.screenDelta
-      else
+      else if clipToAtomicTokenEdge == 'left'
         tokenStartColumn
+      else #'closest'
+        if column > tokenStartColumn + token.screenDelta / 2
```
as-cii: Is this the correct logic? The expression will be evaluated like this:

```coffee
column > tokenStartColumn + (token.screenDelta / 2)
```

Shouldn't we sum first and then divide by 2? If the answer is yes, then our tests should be revisited to shield against this kind of error.

src/marker.coffee
```diff
@@ -311,7 +310,6 @@ class Marker
   # * `screenPosition` The new {Point} to use
   # * `properties` (optional) {Object} properties to associate with the marker.
   setTailScreenPosition: (screenPosition, options) ->
-    screenPosition = @displayBuffer.clipScreenPosition(screenPosition, options)
```
as-cii: :sparkles: Clever deletion here: this runs already inside `displayBuffer.bufferPositionForScreenPosition`, so there's no need to run it a second time.

src/tokenized-line.coffee
```diff
@@ -59,7 +74,7 @@ class TokenizedLine
     screenColumn = 0
     currentBufferColumn = 0
     for token in @tokens
-      break if currentBufferColumn > bufferColumn
+      break if currentBufferColumn + token.bufferDelta > bufferColumn
```
as-cii: Ideally, we should have a failing test for each change we make to the code: deleting the `+ token.bufferDelta`, however, has no impact on the test suite.

Could you explain why you changed this, please? Also, could you add a test for it?

src/tokenized-line.coffee
adzenith: Yeah, it's correct, but I should definitely add parentheses to make it more clear. What we're doing here is seeing if we're more than halfway through the token, so we're interested in just dividing the token width by 2.

If you add parentheses the wrong way (e.g., `column > (tokenStartColumn + token.screenDelta) / 2`) then it successfully fails a test:
```
TextEditor
  cursor
    .moveRight()
      it moves the cursor by one column to the right
        Expected { row : 3, column : 5 } to equal [ 3, 4 ].
```

src/tokenized-line.coffee
adzenith: Good catch that there's no tests. I'll add some.

The default behavior of clipScreenColumn changed from "left" to "closest". On a line with one tab, trying to find the screen position for `bufferColumn = 0`, it used to go something like this:
```
screenColumn = token.screenDelta = 4
currentBufferColumn = token.bufferDelta = 1 #which is now larger than bufferColumn
@clipScreenColumn(4 + (0 - 1)) = @clipScreenColumn(3) => 0 #clip to the left edge of a four-space tab
```
Now that final clip call will clip to the right edge instead, because it's closer. This changes it so that if the column is in the middle of a token (or on the left edge of it) we want to bail out of the for loop and clip right now, and clip will handle the logic correctly.

I'm going to add some tests of DisplayBuffer::screenPositionForBufferPosition (which is currently the only caller of TokenizedLine::screenColumnForBufferColumn) inside of display-buffer-spec. Sound good? The clip tests are passed through DisplayBuffer in the same way.

adzenith:
The additional unit tests are working; I'm going to try a merge and see what I can see.

adzenith:
I've merged it into master and all tests are passing. Let me know if there's anything else you'd like me to change!

spec/display-buffer-spec.coffee
```diff
@@ -147,19 +144,13 @@ describe "DisplayBuffer", ->
...
 
-          expect(tokenizedLine.tokens[0].value).toBe("  ")
+          expect(tokenizedLine.tokens[0].value).toBe("    ")
           expect(tokenizedLine.tokens[0].isSoftWrapIndentation).toBeTruthy()
 
-          expect(tokenizedLine.tokens[1].value).toBe("  ")
+          expect(tokenizedLine.tokens[1].value).toBe("   ") # hanging indent
           expect(tokenizedLine.tokens[1].isSoftWrapIndentation).toBeTruthy()
 
-          expect(tokenizedLine.tokens[2].value).toBe("  ") # hanging indent
```
as-cii: We actually want to preserve these tests. What about setting `tabLength` only on `::clipScreenPosition` test and leave everything as-is? Here's a diff you can apply to your current branch:

```diff
diff --git a/spec/display-buffer-spec.coffee b/spec/display-buffer-spec.coffee
index 9b38323..361e48c 100644
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -4,7 +4,7 @@ _ = require 'underscore-plus'
 describe "DisplayBuffer", ->
   [displayBuffer, buffer, changeHandler, tabLength] = []
   beforeEach ->
-    tabLength = 4
+    tabLength = 2
 
     buffer = atom.project.bufferForPathSync('sample.js')
     displayBuffer = new DisplayBuffer({buffer, tabLength})
@@ -127,10 +127,13 @@ describe "DisplayBuffer", ->
         it "breaks soft-wrap indentation into a token for each indentation level to support indent guides", ->
           tokenizedLine = displayBuffer.tokenizedLineForScreenRow(4)
 
-          expect(tokenizedLine.tokens[0].value).toBe("    ")
+          expect(tokenizedLine.tokens[0].value).toBe("  ")
           expect(tokenizedLine.tokens[0].isSoftWrapIndentation).toBeTruthy()
 
-          expect(tokenizedLine.tokens[1].isSoftWrapIndentation).toBeFalsy()
+          expect(tokenizedLine.tokens[1].value).toBe("  ")
+          expect(tokenizedLine.tokens[1].isSoftWrapIndentation).toBeTruthy()
+
+          expect(tokenizedLine.tokens[2].isSoftWrapIndentation).toBeFalsy()
 
       describe "when editor.softWrapHangingIndent is set", ->
         beforeEach ->
@@ -144,13 +147,19 @@ describe "DisplayBuffer", ->
         it "includes hanging indent when breaking soft-wrap indentation into tokens", ->
           tokenizedLine = displayBuffer.tokenizedLineForScreenRow(4)
 
-          expect(tokenizedLine.tokens[0].value).toBe("    ")
+          expect(tokenizedLine.tokens[0].value).toBe("  ")
           expect(tokenizedLine.tokens[0].isSoftWrapIndentation).toBeTruthy()
 
-          expect(tokenizedLine.tokens[1].value).toBe("   ") # hanging indent
+          expect(tokenizedLine.tokens[1].value).toBe("  ")
           expect(tokenizedLine.tokens[1].isSoftWrapIndentation).toBeTruthy()
 
-          expect(tokenizedLine.tokens[2].isSoftWrapIndentation).toBeFalsy()
+          expect(tokenizedLine.tokens[2].value).toBe("  ") # hanging indent
+          expect(tokenizedLine.tokens[2].isSoftWrapIndentation).toBeTruthy()
+
+          expect(tokenizedLine.tokens[3].value).toBe(" ") # odd space
+          expect(tokenizedLine.tokens[3].isSoftWrapIndentation).toBeTruthy()
+
+          expect(tokenizedLine.tokens[4].isSoftWrapIndentation).toBeFalsy()
 
     describe "when the buffer changes", ->
       describe "when buffer lines are updated", ->
@@ -615,6 +624,9 @@ describe "DisplayBuffer", ->
 
   describe "::clipScreenPosition(screenPosition, wrapBeyondNewlines: false, wrapAtSoftNewlines: false, clipToAtomicTokenEdge: 'closest')", ->
     beforeEach ->
+      tabLength = 4
+
+      displayBuffer.setTabLength(tabLength)
       displayBuffer.setSoftWrapped(true)
       displayBuffer.setEditorWidthInChars(50)
 
@@ -716,7 +728,6 @@ describe "DisplayBuffer", ->
 
     it "correctly translates positions on soft wrapped lines containing tabs", ->
       buffer.setText('\t\taa  bb  cc  dd  ee  ff  gg')
-      displayBuffer.setTabLength(2)
       displayBuffer.setSoftWrapped(true)
       displayBuffer.setEditorWidthInChars(10)
       expect(displayBuffer.screenPositionForBufferPosition([0, 10], wrapAtSoftNewlines: true)).toEqual [1, 4]

```

src/tokenized-line.coffee
```diff
@@ -40,10 +40,20 @@ class TokenizedLine
   copy: ->
     new TokenizedLine({@tokens, @lineEnding, @ruleStack, @startBufferColumn, @fold})
 
+  # This clips a given screen column to a valid column that's within the line
+  # and not in the middle of any atomic tokens.
+  #
+  # column - the column to clip
```
as-cii: Really minor, but what about the following?

```
# column - A {Number} representing the column to clip
```

src/tokenized-line.coffee
as-cii: I see, but I am wondering: what do you think if we simply preserve the original behavior and always clip to `left` instead?

spec/display-buffer-spec.coffee
adzenith: Yeah, that's a much better idea. I had changed the global tab length before because only one other test was using it, but I'll do it this way instead now that there are a lot of tests using 2-space tabs.

src/tokenized-line.coffee
as-cii: By the way, I think it would be awesome to add some more tests to document this behavior a bit better. Maybe handling edge cases like the following:

* Left edge of a token
* Middle of a token
* Right of a token

What do you think?

src/tokenized-line.coffee
adzenith: Done. I added `{Number}` for the return value doc as well.

adzenith:
I've taken out the `tabLength` change entirely. I've been making changes and then merging with master; in general, would you prefer this or a rebase?
I'll write some more tests around `screenColumnForBufferColumn`.
Thanks!

spec/display-buffer-spec.coffee
```diff
@@ -703,6 +715,11 @@ describe "DisplayBuffer", ->
       expect(displayBuffer.screenPositionForBufferPosition([100000, 0])).toEqual [12, 2]
       expect(displayBuffer.screenPositionForBufferPosition([100000, 100000])).toEqual [12, 2]
 
+    it "clips correctly around atomic tab characters", ->
+      buffer.insert([0, 0], '\t')
```
as-cii: I think here you need to set `tabLength` to `4`, otherwise this spec will always pass.

as-cii: Moreover, what do you think to handle also every column between 0 and 4 in this test?

src/tokenized-line.coffee
as-cii: Never mind. After completing specs around this I think we can just use your version, as it's symmetrical with `bufferColumnForScreenColumn` :+1:

as-cii:
> I've taken out the tabLength change entirely

Good job there! :+1: 

> I've been making changes and then merging with master; in general, would you prefer this or a rebase?

Merging is fine :wink: 

After completing those specs I mentioned earlier on, I think we can merge this right away :metal:

spec/display-buffer-spec.coffee
adzenith: Good catch - this had been affected by my global tabLength change before, and when I took that commit out I missed this. I will add more tests here as well.

adzenith:
Sounds great. I'll add some specs for `screenColumnForBufferColumn` and push them up.

thomasjo:
> I've been making changes and then merging with master; in general, would you prefer this or a rebase?

Either is fine, but generally a rebase is better in my opinion as it keeps the commit history cleaner. I don't even use the *Merge pull request* button on my personal projects because I really care about keeping the history as clean and "consumable" as possible :grinning: 

Regardless, thanks for working on this!

adzenith:
> a rebase is better in my opinion as it keeps the commit history cleaner.

I agree. I asked because I've also seen projects where they want a merge commit every time. I'll rebase it :)

spec/display-buffer-spec.coffee
adzenith: Wait, this should be independent of tab length. We're going from buffer position, which is either 0 (before the tab) or 1 (after the tab), and the result should be either 0 in the first case or tabLength in the second case.

src/tokenized-line.coffee
adzenith: Cool, that's what I was thinking too. I've been looking through the code and it looks like `screenPositionForBufferPosition` interacts more interestingly with soft tabs than with hard tabs. If you have a soft tab, then `screenPositionForBufferPosition` clips to the edges of it when called for any of the individual spaces making it up. Hard tabs only take up one buffer character, and `screenPositionForBufferPosition` calculates the tab width.
The specs are written to test around hard tabs, soft wraps, and folds, but not for soft tabs. I think those are the only tests we're missing.

adzenith:
Pushed the rebased code. I'm going to look into adding specs for soft tabs and `screenPositionForBufferPosition`.

spec/display-buffer-spec.coffee
as-cii: Correct, however if you leave `tabLength` as it is (namely, `= 2`) spec will pass anyway, even if we remove `+ token.bufferDelta`. That's because with `2`, if you execute:

```coffee
displayBuffer.screenPositionForBufferPosition([0, 0])
```

It will try to clip `1`, which is exactly in the middle and `closest` will always prefer to clip to `0` (behaving exactly as `left`). We need a higher value, such as `4`, which will try to clip to something past the middle (e.g. `3`), reporting an error if we leave out `+ token.bufferDelta`.

src/tokenized-line.coffee
as-cii: @adzenith I agree to put some boundaries around soft tabs behavior, therefore please go ahead with this :+1:  

By the way, I tested the code below on **master** and, unless we're exactly at the edge of an atomic token, it always clips to left, whereas the new code will clip to right if we're in the middle (thus making these tests fail).

```coffee
      tabLength = 4
      displayBuffer.setTabLength(tabLength)

      expect(displayBuffer.screenPositionForBufferPosition([2, 0])).toEqual [2, 0]
      expect(displayBuffer.screenPositionForBufferPosition([2, 1])).toEqual [2, 0]
      expect(displayBuffer.screenPositionForBufferPosition([2, 2])).toEqual [2, 0]
      expect(displayBuffer.screenPositionForBufferPosition([2, 3])).toEqual [2, 4]
      expect(displayBuffer.screenPositionForBufferPosition([2, 4])).toEqual [2, 4]
```

However, I believe that **master** behavior is not correct and that the change we're making here is the right way to go.

Great job, @adzenith :fire:

as-cii:
> Pushed the rebased code. I'm going to look into adding specs for soft tabs and screenPositionForBufferPosition.

@adzenith: I left some comments, but overall I think we're getting closer to merge this :metal: 

@atom/non-github-maintainers: would you be interested in testing this out to catch eventual regressions we're not taking into account? Your help would be much appreciated! :pray:

spec/display-buffer-spec.coffee
adzenith: Oh! I see what you're saying. On it.

adzenith: The new test correctly passes, and correctly fails when the ` + token.bufferDelta` is removed.

src/tokenized-line.coffee
adzenith: I've added these tests to test `screenPositionForBufferPosition`. Thanks!

adzenith: Oh, and this ended up in comments for a different diff, but there's a test now for the ` + token.bufferDelta`. :hammer:

adzenith:
Sweet. At this point I think I've updated the code based on all the comments and feedback you've provided: tests for the change to `screenColumnForBufferColumn`, new tests for `screenPositionForBufferPosition`'s handling of soft tabs, and changing the global `tabLength` in display-buffer-spec back to 2.

If there's anything I've missed, let me know. Otherwise ... maybe it's ready for merge :dart:?!

spec/display-buffer-spec.coffee
```diff
@@ -721,6 +733,26 @@ describe "DisplayBuffer", ->
       expect(displayBuffer.screenPositionForBufferPosition([100000, 0])).toEqual [12, 2]
       expect(displayBuffer.screenPositionForBufferPosition([100000, 100000])).toEqual [12, 2]
 
+    it "clips correctly around atomic tab characters", ->
+      #set tabLength higher than 2 to make sure we're not just rounding up; we have to actually clip
```
as-cii: :bulb: Maybe we could be a bit more verbose on this spec, what about:

```coffee
it "clips to the (left or right) edge of an atomic token without rounding up"
```

So that we can get rid of the comment as well :fire:

adzenith: Good idea. I wanted that info in there, but adding it to the spec name is smarter than adding a comment.

spec/display-buffer-spec.coffee
```diff
@@ -721,6 +733,26 @@ describe "DisplayBuffer", ->
...
+      tabLength = 4
+      displayBuffer.setTabLength(tabLength)
+
+      buffer.insert([0, 0], '\t')
+      expect(displayBuffer.screenPositionForBufferPosition([0, 0])).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForBufferPosition([0, 1])).toEqual [0, tabLength]
+
+    it "clips correctly around soft tab characters", ->
```
as-cii: What about `it "clips to the edge closest to the given position when it's inside a soft tab"`?

as-cii:
> If there's anything I've missed, let me know. Otherwise ... maybe it's ready for merge ?!

Two minor comments, but we should be good to go. Before merging this, however, I'd like to leave it open for a while and catch any regression we may have missed.

Awesome job, @adzenith! :rainbow:

spec/display-buffer-spec.coffee
adzenith: I think it's a good idea!

adzenith:
Excellent! Thanks so much for the great code review @as-cii! I enjoyed the emoji considerably. :tophat: :wave:

as-cii:
> Excellent! Thanks so much for the great code review @as-cii! I enjoyed the emoji considerably.

It was great to collaborate with you on this.

@nathansobo: could you have a final look so that we can :shipit:? Thank you :bow:

src/tokenized-line.coffee
```diff
@@ -41,10 +41,20 @@ class TokenizedLine
...
+  #           'closest' (default): clip to the closest edge of an atomic token.
+  #           'right': clip to the right edge.
+  #           'left': clip to the left edge.
+  #
+  # Returns a {Number} representing the clipped column.
   clipScreenColumn: (column, options={}) ->
     return 0 if @tokens.length == 0
 
-    { skipAtomicTokens } = options
+    { clipToAtomicTokenEdge } = options
```
nathansobo: What do you guys think about just calling this `clip` and doing `forward` or `backward` as option values like we're doing in the new `text-document`?

nathansobo:
I'd like to see the option name move more toward what we're working on in text-document if possible, but other than that, I think this is looking really good. Thanks to both of you!

src/tokenized-line.coffee
as-cii: :+1:

@adzenith: could you change that, please?

as-cii:
> I'd like to see the option name move more toward what we're working on in text-document if possible, but other than that, I think this is looking really good. Thanks to both of you!

@adzenith did a great job here :checkered_flag: Let's wait for him to update this PR with the renaming you proposed and then we can :shipit:

adzenith:
I don't know what you mean... it's been called `clip` this whole time :sparkles: :stuck_out_tongue:

Re-rebased onto master; all tests are still passing. Let me know if I can make any more changes! :doughnut:

as-cii:
Thanks a lot, @adzenith! :fireworks: :fireworks: :fireworks:

adzenith:
Woo merged! Thanks so much! :sunny::seedling:

PR #5279, opened 2015-01-27
Enable DirectWrite

anaisbetts:
Chrome 39 proper now has DirectWrite as the default renderer on Windows, which makes a huge difference in text clarity, especially on High DPI monitors.


kevinsawicki:
Thanks!

PR #5280, opened 2015-01-27
Reset all user config values when the schema changes

### LINKED ISSUE #5219: Package config schemas aren't used at startup
* With Atom closed, add the following to your `~/.atom/config.cson` file:

```coffee
'markdown-preview':
  liveUpdate: 4
```

* Open Atom
* Run `atom.config.get('markdown-preview.liveUpdate')` from the Console
* `4` is printed out

This schema for this package states this key path should be a boolean, https://github.com/atom/markdown-preview/blob/dfc0f9617dc42885ec719134931c78293d275524/lib/main.coffee#L26-L28 but no validation error is logged at startup.

### LINKED ISSUE #5306: atom.config.get '<some-color>' doesn't return an instance of Color if set in config.cson
When I try to `get` a config value of a setting with the `color` type, and the setting is saved to `config.cson` (so not the default value), I don't get an instance of the `Color` class, but rather just the object/string that was in `config.cson`.

Versions:
Atom: `0.176.0`
OS: Linux Debian 8 (testing)

benogle:
What's happening? A value in the user's config file is not run through the config schemas
- The user config is loaded from disk, and sets without running through the schemas
- The user config is loaded before package schemas are read

I had a couple options
1. Just run the value through the correct schema on `get`. This is not intrusive, but it slows down reads. And reads happen a lot
2. Split up initial config load to loading core config, loading package schemas, then loading the rest of the config values
3. Reset the all values in the user config on schema changes to take advantage of the new schemas

I opted for option 3 as it should keep the config in sync as packages are activated/deactivated/installed, etc. This may have an impact on startup time because the user config is reset (though, without save) for each schema addition. 

Closes #5219 
Closes #5306

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1135,8 +1135,16 @@ describe "Config", ->
               int:
                 type: 'integer'
                 default: 2
+              bar:
+                type: 'string'
+                default: 'def'
 
-        it 'the values set respect the new schema', ->
+        it "the values set respect the new schema", ->
+          expect(atom.config.set('foo.bar.bar', 'globalBar')).toBe true
+          expect(atom.config.set('foo.bar.bar', 'scopedBar', scopeSelector: '.source.js')).toBe true
+          expect(atom.config.get('foo.bar.bar')).toBe 'globalBar'
+          expect(atom.config.get('foo.bar.bar', scope: ['.source.js'])).toBe 'scopedBar'
+
           expect(atom.config.set('foo.bar.int', 'nope')).toBe true
           expect(atom.config.set('foo.bar.int', 'notanint', scopeSelector: '.source.js')).toBe true
           expect(atom.config.set('foo.bar.int', 23, scopeSelector: '.source.coffee')).toBe true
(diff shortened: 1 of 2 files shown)


spec/config-spec.coffee
```diff
@@ -1106,6 +1106,44 @@ describe "Config", ->
...
+        beforeEach ->
+          schema =
+            type: 'object'
+            properties:
+              int:
+                type: 'integer'
+                default: 2
+
+        it 'the values set respect the new schema', ->
+          expect(atom.config.set('foo.bar.int', 'nope')).toBe true
```
kevinsawicki: So if you have something like `foo.bar.str` set to `'hey'`, does it get unset after the schema gets set even though it doesn't have a schema?

Mind want to add some coverage for this case.

benogle: It does, yeah. I will add a case

benogle: Sorry, no it does not get unset. I thought you were asking something else.

kevinsawicki: Cool, still might be cool to verify it doesn't get unset in the specs

benogle:
Startup speed seems to be the same as without this change.

spec/config-spec.coffee
```diff
@@ -1106,6 +1126,62 @@ describe "Config", ->
...
+            properties:
+              int:
+                type: 'integer'
+                default: 2
+              str:
+                type: 'string'
+                default: 'def'
+
+        it "the values set respect the new schema", ->
+          expect(atom.config.set('foo.bar.str', 'global')).toBe true
```
kevinsawicki: What do you think about testing the case of a value that is set before the schema is applied but doesn't have an associated property in the applied schema?

benogle: that's the `noschema` property i'm setting 5 lines down, no?

kevinsawicki: Damn, sorry, was reviewing before I noticed https://github.com/atom/atom/commit/bfbeed8400ab404f7e9f0e0d3b2c7526b45083e7 landed, :+1:

spec/config-spec.coffee
```diff
@@ -1106,6 +1126,62 @@ describe "Config", ->
...
+            type: 'object'
+            properties:
+              int:
+                type: 'integer'
+                default: 2
+              str:
+                type: 'string'
+                default: 'def'
+
+        it "the values set respect the new schema", ->
```
kevinsawicki: This `it` reads a bit weird since the `it the..` doesn't really flow, maybe `respects the new schema when values are set` or something?

PR #5293, opened 2015-01-28
Move all text editor view state into a presenter object

nathansobo:
This PR changes our approach to text editor rendering by centralizing all state in a `TextEditorPresenter` object that sits between the view and the model.

presenter diagram

The presenter manages a mutable state object that is synchronously updated whenever there are changes from the model or changes to measurements from the view. The view consumes this state object when updating the DOM asynchronously so we don't need to recompute it before every render like we do now.

This reconfiguration paves the way to perform all DOM updates manually rather than relying on React's virtual DOM reconciliation system. The view will maintain a copy of the presenter state, and it will perform a comparison of the presenter's state against the copy when updating the DOM.

The net result will be a design very similar to React's reconciliation mechanism, but my hope is that a lower-level implementation targeting our specific problem domain will make this design easier to profile and open up specific optimizations that wouldn't be possible at a higher level of abstraction. We already perform DOM updates to lines and line numbers manually to optimize around reconciliation overhead; this would just complete that process.
### Optimizations
- [x] Perform a single decorations query when the visible row range changes to cache the visible decorations, then read decoration classes from that when building lines, line numbers, etc rather than querying multiple times on every change.
- [x] When decorations are added / removed, update only the decoration classes on lines and line numbers rather than doing a full pass on their content.
- [x] Cache `foldable` status for each line in the tokenized buffer instead of recomputing it every time we update the line numbers.
### Fixes
- [x] Fix velocity scrolling upward
- [x] Fix TextEditorComponent specs
- [x] Compute row id correctly when the first visible line is soft-wrapped.
- [x] Eliminate `TextEditorPresenter::getHeight`
- [x] Audit handling of scrollTop/scrollLeft
- [x] Enforce maximum scrollTop/Left based on height/width and content. This should hopefully address some flakiness in the random test.
- [x] Editor losing focus when switching grammars with grammar switcher. Does that happen on master too?
- [x] Gutter background color isn't changing when themes get changed.
- [x] Switching syntax themes is super slow


mark-hahn:
Is there a problem modifying the DOM directly at the same time?

nathansobo:
> Is there a problem modifying the DOM directly at the same time?

I might consider it, but I'm trying to break this up into pieces at the request of the team since it's already a big change as is.

mark-hahn:
I'm sorry.  I meant could a package modifying the DOM directly conflict
with this presenter modification?

Now that I think about it the answer is probably no.

nathansobo:
@mark-hahn Yeah it shouldn't interfere. Although one thing I'm hoping to solve is to provide a hook for appropriate *times* to modify the DOM yourself to prevent reflows. That's fairly fuzzy at the moment though and a ways off.

mark-hahn:
>  appropriate *times* to modify the DOM yourself to prevent reflows.

That would be cool.  Can a package post changes to the presenter and let it
handle everything?

nathansobo:
The basic structure is in place here, so now it's time to optimize. It breaks down to about 6ms updating the model and presenter state and 4ms updating the DOM for a single character insert. I have reason to believe I can improve on this substantially.

screenshot_2015-01-29_15_56_14

There's a ton of low hanging fruit on the presenter update because I did everything in the most naive way, so that's the work I'll do on this PR and save DOM update optimizations for a subsequent PR. To ensure optimizations don't introduce regressions, I'm going to build a spec that randomly mutates an editor and compares the state of a brand new presenter to that of a preexisting presenter to be sure updates were performed correctly.

After that, some of the presenter optimizations I have in mind:

* [ ] Perform a single decorations query when the visible row range changes to cache the visible decorations, then read decoration classes from that when building lines, line numbers, etc rather than querying multiple times on every change.
* [ ] When decorations are added / removed, update only the decoration classes on lines and line numbers rather than doing a full pass on their content.
* [ ] Cache foldability status on tokenized lines so we don't have to recompute it when updating line numbers.

I have more ideas, but I'd like to try these things first and re-evaluate.

I also want to re-introduce `shouldComponentUpdate` methods on all the React components which should improve reconciliation performance.

nathansobo:
We're going to try this out for a couple days before merging, but this should be ready to go. Next steps:

* Cache "foldable" status of lines so it's cheaper to update the line numbers in the presenter
* Switch to manual DOM manipulation
* Try tiled rendering

nathansobo:
Randomized spec is failing. Need to get that resolved before merging.

nathansobo:
I'm going to go ahead and cache `foldable` in the tokenized buffer before merging since we're recomputing the gutter more frequently with this design. I'm thinking that's the last step on this unless any other performance issues crop up.

nathansobo:
Finished caching "foldable" status on tokenized lines. The sync presenter update is down to about 3.4ms, and only about 1ms is spent in the presenter. Still room for optimization there, but I think we have other :fish: to :egg: before we worry about it. Want to spend a day using this in its final form before merging.

screenshot 2015-02-06 18 58 36

maxbrunsfeld:
:sparkles: hoorah! :beers:

nathansobo:
@maxbrunsfeld Give it a build and see if it feels more responsive to you, if you weren't already testing that branch.

PR #5299, opened 2015-01-29
Transpile all .js files beginning with the "use 6to5"; pragma with 6to5.

bolinfest:
Transpile all .js files with 6to5.

In the spirit of supporting JavaScript development for Atom packages,
this adds default support for es.next transpilation support in the way
that Atom already has default support for CoffeeScript transpilation.
There are many new features in ES6+ that make JavaScript development
easier and more enjoyable, particularly in terms of support for async code.

For reference, this was a much faster way to iterate on this than running `./script/build`
each time:

```
cp /Users/mbolin/src/atom/static/index.js /Applications/Atom.app/Contents/Resources/app/static/index.js
coffee --output /Applications/Atom.app/Contents/Resources/app/src --compile /Users/mbolin/src/atom/src/esnext.coffee
```

Run the following in the console to see how warm the cache was after startup:

```
global.require('../src/esnext/').getCacheHits()
global.require('../src/esnext/').getCacheMisses()
```


kevinsawicki:
> See if it's possible to have apm publish do the transpilation so users of the package do not have to waste startup time on transpilation.

Usually we prefer to warm the cache on `apm install` so publishes can always just be regular tarballs from the repo.

I'd be happy to add this into `apm install` when the time comes, https://github.com/atom/apm/blob/30569a16d87916f191ffb99415f61dac99bc0802/src/install.coffee#L473-L487

src/esnext.coffee
```diff
@@ -0,0 +1,92 @@
+fs = require 'fs'
+path = require 'path';
+to5 = require '6to5'
+
+ES_DIRECTORY = 'atom-es'
```
kevinsawicki: Could you explain a little about what this directory is for?

kevinsawicki: Nevermind, I see it is described in the body of the PR.

kevinsawicki:
> As an initial starting point for discussion, this change requires
developers to store all files that need to be transpiled in an atom-es
directory under the module root.

Is there a file extension that could be used/introduced instead?

kevinsawicki:
> This change also supports reading options to pass to 6to5 from the source file's
package.json in case the default 6to5 options used by Atom are
not appropriate.

Do you think this will be common? Just wondering if we can get away with a single configuration to start.

kevinsawicki:
Is the reason to not just transpile all `.js` files that the performance is noticeable even for files where nothing would be changed?

bolinfest:
@kevinsawicki I just addressed some of these issues [in the forum](https://discuss.atom.io/t/rfc-support-for-6to5-in-atom/14557), but I'll reiterate here as I'm hoping to keep the discussion on this pull request:

* I'm rethinking that whole `atom-es` thing and believe that listing the paths in the `package.json` would be more appropriate. Then you can organize the files in your package however you like, tests can be transpiled, etc.
* I don't think that requiring the use of a new file extension is a good idea. Especially if you're writing ES6, you shouldn't have to pretend that you aren't writing JavaScript.
* I don't think that it will be common to need to override the default 6to5 options, but I wanted to give myself an escape hatch for easier testing. In particular, as we play with newer versions of React in Atom, I would set `"reactCompat": false` in my test packages.
* I haven't profiled how long transpilation takes, but I guess I'm more of a "better safe than sorry" kinda guy in that respect. If it's fast enough now, but then gets slower over time, what do we do? For people who aren't using ES6, if 6to5 doesn't write out their code *exactly* as they wrote it originally, it seems like that could be confusing. But if you want to push that envelope, then I'm on board! (I was also a little hesitant because all of your transitive dependencies would also get transpiled, which includes everything from npm, so I don't know if that could possibly break anything.)

bolinfest:
Come to think of it, how do you feel about an .atom.js suffix? Will that
play well with Node? Would foo.atom.js have to be pulled in via
require('foo.atom') instead of just require('foo')?

lee-dohm:
My main concern is whether or not the class model, module model, Unicode support and other features mesh well with the CoffeeScript-generated ones. As can be seen in [this and other linked topics from Discuss](https://discuss.atom.io/t/views-in-a-package-without-coffee-script/5863) there already are issues with extending CoffeeScript classes from JavaScript. I suppose if 6to5 support is limited to packages then as before it is up to the package author to ensure compatibility ... though with the planned [ServiceHub](https://github.com/atom/atom/issues/5165) ... that might be more complicated than before.

My secondary concern is that I prefer CoffeeScript to what I've seen of ES6, so on a personal level I would hope that this is adding an option for package developers ... not adding an option with the hope of eventually supplanting CoffeeScript :grinning:

I'm sure the "do we compile or don't we and when" and compile caching stuff will all get worked out by smart people :smile:

bolinfest:
@lee-dohm I've done a bunch of subclassing of SpacePen's `View` class form ordinary JS and there's nothing fundamental that prevents it from working. I believe you just have to be a bit careful with how the `super` keyword is thrown around in CoffeeScript. I don't see any issues with mixing 6to5-generated classes with inheritance.

Also, in the JS console in Atom, I can already write `class Foo {}`, so it's probably not too long before CoffeeScript has an option to transpile to ES6 classes.

I don't mean for this diff to suggest that Atom should drop support for CoffeeScript, though once you start writing more async code, you may come around on your own :)

bolinfest:
@kevinsawicki Would this diff be more amenable if I changed it so I:

* Removed the option to customize flags to 6to5.
* Transpiled all files with a `.js` extension.
* Added caching support. If all files with a `.js` extension will go through 6to5, then I don't think this can go in until caching is set up.

If that will move things along, then I'm happy to go ahead and make these changes.

bolinfest:
I can tell you this much. If I turn on transpilation for all `.js` files and don't implement caching, Atom takes almost 2 minutes to start up. It makes it hard to iterate on testing this feature...I guess I'll just have to get it right the first time?

sebmck:
Potential extensions could be `.es`, `.es6`, `.jsx` or even `.6to5`. None of those are entirely suitable though.

You shouldn't really be concerned with class interopability as ES6 classes are just sugar around prototypes. ie.

```javascript
class Foo {}
```

is basically equivalent to

```javascript
function Foo() {}

Foo.prototype.bar = function () {};
```

Compiling every single file with the `.js` extension seems like an extreme overkill which is why you're seeing such long compile times.

package.json
```diff
@@ -19,6 +19,7 @@
   ],
   "atomShellVersion": "0.21.0",
   "dependencies": {
+    "6to5": "3.0.14",
```
sebmck: `6to5-core` would be better here since it's a build of `6to5` without certain dependencies so it'll install faster. You can see the script that generates the `6to5-core` `package.json` [here](https://github.com/6to5/6to5/blob/master/tools/generate-core-package-json).

src/esnext.coffee
```diff
@@ -0,0 +1,106 @@
...
+
+      # Because Atom is currently packaged with a fork of React v0.11,
+      # it makes sense to use the --react-compat option so the React
+      # JSX transformer produces pre-v0.12 code.
+      'reactCompat': true
+
+      # Blacklisted features do not get transpiled. Features that are
+      # natively supported in the target environment should be listed
+      # here. Because Atom uses a bleeding edge version of Node/io.js,
+      # I think this can include es6.arrowFunctions, es6.classes, and
```
sebmck: v8 doesn't support arrow functions and classes are still behind a flag as they aren't quite spec compliant yet.

src/esnext.coffee
```diff
@@ -0,0 +1,106 @@
...
+    cachePath = getCachePath(sourceCode)
+    js = getCachedJavaScript(cachePath)
+
+    unless js
+      options = createOptions filePath
+      try
+        js = to5.transform(sourceCode, options).code
+        stats.misses++
+      catch error
+        console.error('Error compiling ' + filePath + ': ' + error)
```
sebmck: `error` here will be an instance of `Error` which wont coerce very well to a string and you wont be able to see the code frame. Should be getting `err.stack` which automatically includes the filename anyway. He's an example:

```javascript
SyntaxError: script.js: Unexpected token (1:4)
> 1  | asdf!@#$
     |     ^
  2  |
```

bolinfest:
@sebmck Thanks for the tip about 6to5-core! I updated the diff to use it.

@kevinsawicki I think I have addressed all of the issues you raised. Now all `.js` files go through 6to5 and have a cache path that is a function of:

* The version of 6to5.
* The options passed to 6to5. (See `create6to5VersionAndOptionsDigest()`.)
* The file contents.

The first time you start Atom (or if you change the version of 6to5 or its options), it will take about 2 minutes to start up, but after that, it is much better. Atom tells me that it loads 1,224 files from cache on startup with my current set of packages. That's quite a bit of sha1'ing, but I'll let you test drive this for yourself to decide how you feel about its performance.

I originally tried incorporating `atom.appVersion` into the hash, but then every time you make a local commit to Atom, the `atom.appVersion` changes, invalidating the cache. Then you hit a 2 minute startup time again, so that was impractical.

mark-hahn:
> The first time you start Atom (or if you change the version of 6to5 or its options), it will take about 2 minutes to start up

This seems unacceptable to me.  If I saw this happening I'd kill the process.

sebmck:
Completely agree with @mark-hahn.

nathansobo:
After sleeping on this I have an idea. What if we decide to preprocess files based on the presence of a `.preprocess.json` file in a directory. Directives in this file would apply recursively to all files the containing directory and descendant directories.

I'm thinking the file could contain metadata about how to preprocess various extensions. So for the 6to5 case it might look like this:

```json
{
  "js": ["6to5"]
}
```

That would run `6to5` on all files with the `.js` extension. Or if you had specific options to pass to the pre-processor, you could use an expanded form:

```json
{
  "js": [{
    "name": "6to5",
    "options": {}
  }]
}
```

The nice thing about a file is that it doesn't affect the directory structure. On package installation we could run the desired preprocessors up-front and then delete the file entirely. If the file isn't there, we pay nothing.

Keying on extension means we could opt in to preprocessing for files other than just `.js`. Using arrays makes room for multiple preprocessor steps such as `sweet.js` in the future. Also, I'm thinking that preprocessor settings wouldn't cascade through the `node_modules` directory.

What do you guys think?

bolinfest:
OK, I think we're agreed that categorically transpiling all `.js` files is a bad idea and that we would like to be able to robustly pre-transpile code where possible to help improve startup times.

@nathansobo's proposal generally sounds good to me since it seems to achieve both of those goals. We might want to check upwards for a `.preprocess.json` file until we find either it or a `package.json` file, which I think would address the "don't cascase through `node_modules`" thing. Though I guess if you insist on having a `.preprocess.json` in your root, then including `node_modules` would still be problematic. Moving your `index.js` to `src/index.js` to avoid this issue doesn't seem like that big of a deal to me, though. (Or we could support a "norecurse" option in `.preprocess.json`?)

The only hangup I have about supporting the `options` in the `.preprocess.json` is that they need to be hashable in order for the path in the compile-cache to be sound. The current iteration of my pull request supports this for the 6to5 options that are used, but is not a general hashing solution for arbitrary JSON. I suppose it wouldn't be that hard to extend it, though.

bolinfest:
The other thing that we probably have to verify is that checking for (and parsing) these `.preprocess.json` files is that it's "cheap enough."

nathansobo:
> The only hangup I have about supporting the `options` in the `.preprocess.json` is that they need to be hashable in order for the path in the compile-cache to be sound.

Maybe we start with the simple case of not allowing options and keep that idea in our back pocket?

> The other thing that we probably have to verify is that checking for (and parsing) these .preprocess.json files is that it's "cheap enough."

Seems like checking for non-existence of the file should be pretty cheap for the fast-path case.

bolinfest:
"Seems like checking for non-existence of the file should be pretty cheap for the fast-path case."

Right, I guess the question is whether it makes sense to check all the way up to the root of the filesystem, or check for `.preprocess.json` and then `package.json` to prevent going all the way to the root.

bolinfest:
@sebmck Updated the PR to use the placeholders supported by Chrome's console API, e.g.:

    console.error('Error compiling %s: %o', filePath, error)

The error is also re-thrown, so between these two things, I think it should be sufficient for debugging.

kevinsawicki:
The solution we have for CoffeeScript is to use `.coffee` as your extension and Atom just compiles it on the fly and caches it based on the contents. This same solution is what we've used for Less files as well, a custom extension, no configuration, and caching based on the SHA-1 of the raw file(s). 

So my gut would say to just implement this the same exact way, a custom extension (I'd lean towards `.es6`), caching based on a SHA-1 of the raw content, and every package uses the same options that are controlled by core (just like Less and CoffeeScript).

Adding any configuration outside of the root `package.json` file will require extra synchronous file reads and JSON parses at startup and requires trickier cache strategies in order for busts to occur when options change but files are the same.

If packages need custom options during development, they can always handle the compilation themselves by opting out of the `.es6` extension and doing whatever they need to do to generate `.js` files.

bolinfest:
@kevinsawicki How do you feel about `.es` since 6to5 supports features beyond those of ES6? (It seems more future proof to me.) It doesn't seem like it's used for many other things: http://filext.com/file-extension/ES.

Having to use a different suffix makes me a little sad, but I admit that it will help us achieve the performance we want and simplify the implementation of this feature.

kevinsawicki:
> How do you feel about .es since 6to5 supports features beyond those of ES6?

I'm fine with that, I'm really flexible on the extension

kevinsawicki:
Looks like `.es` files are parsed as Erlang on GitHub: https://github.com/github/linguist/blob/168ff4c050e29361ec9f9b40536f953347d6f4d4/lib/linguist/languages.yml#L834

bolinfest:
@kevinsawicki Weird, [Wikipedia says](http://en.wikipedia.org/wiki/Erlang_%28programming_language%29) `.erl` and `.hrl` are the extensions.

kevinsawicki:
> Weird, Wikipedia says .erl and .hrl are the extensions.

Yeah, I have no idea if `.es` is "right" in the Erlang case, I just always consult that file in that repo when looking up extensions.

bolinfest:
@kevinsawicki would `.jsx` be acceptable?

bolinfest:
@kevinsawicki I've been sourcing suggestions at Facebook. New ones include `.njs` for "next JavaScript," or `.nes` for "next ECMAScript." As `.njs` is similar to "Ninjas," it seems apropos.

mark-hahn:
What is wrong with `file.6to5` ?

bolinfest:
@mark-hahn It seems to tied to the tool and @sebmck, I thought I saw some talk of changing the project name?

kevinsawicki:
We've been discussing this a bit in meatspace as well and are coming around to an extension not being the best solution here.

What do you think about instead just using something like `"use strict";` instead that marks what JS transpiler it should go through?

If a `.js` file starts with `"use 6to5";` then it goes through the 6to5 transpiler.  That way the natural-feeling `.js` extension is preserved but it comes with a directive of how to process the content.

And if someday we want to add [traceur](https://github.com/google/traceur-compiler) support, it could just be `"use traceur";` at the front of the file.

mark-hahn:
I meant use `.6to5` as the file extension instead of `.es6`

bolinfest:
FYI, I just pushed an update to the PR that implements a generic function for hashing a value parsed from JSON and included a test.

I'll try adding support for `"use 6to5"` now and see how that performs.

kevinsawicki:
> I'll try adding support for "use 6to5" now and see how that performs.

Yeah, for this I think making it a requirement that it be the **very first** line is okay, so I don't think we need to try to skip over comments and what not.

bolinfest:
@kevinsawicki Please take a look at my latest version of the PR. Startup times seem speedy again and my file with `'use 6to5';` at the top gets transpiled as expected.

kevinsawicki:
> Please take a look at my latest version of the PR

Okay, since if feels like we are getting close here I'm going to leave a few minor comments about code styling now.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
+###
+Cache for source code transpiled by 6to5.
+
+Inspired by https://github.com/atom/atom/blob/6b963a562f8d495fbebe6abdbafbc7caf705f2c3/src/coffee-cache.coffee.
+###
+
+crypto = require 'crypto'
+fs = require 'fs-plus'
+path = require 'path';
```
kevinsawicki: :fire: `;`

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+to5 = require '6to5-core'
+
+stats =
+  hits: 0
+  misses: 0
+
+defaultOptions =
+  # The Chrome dev tools will show the original version of the file
+  # when the source map is inlined.
+  'sourceMap': 'inline'
```
kevinsawicki: We typically drop the quotes around the keys unless one of the keys needs it (such as because if contains a `-` or something).  Since none of these keys appear to need it, I'd say drop all the quotes around them in the `defaultOptions` object.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+updateDigestForJsonValue = (shasum, value) ->
+  # Implmentation is similar to that of pretty-printing a JSON object, except:
+  # * Strings are not escaped.
+  # * No effort is made to avoid trailing commas.
+  # These shortcuts should not affect the correctness of this function.
+  if typeof value is 'string'
+    shasum.update('"', 'utf8')
+    shasum.update(value, 'utf8')
+    shasum.update('"', 'utf8')
+  else if typeof value is 'boolean' or typeof value is 'number'
```
kevinsawicki: This could be the shorter `else if typeof value in ['boolean', 'number']` and CoffeeScript will compile to the same thing as this JS-wise.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+  else if Array.isArray value
+    shasum.update('[', 'utf8')
+    for item in value
+      updateDigestForJsonValue(shasum, item)
+      shasum.update(',', 'utf8')
+    shasum.update(']', 'utf8')
+  else
+    # Must be an object: be sure to sort the keys.
+    keys = []
+    for key of value
```
kevinsawicki: Could you just call `Object.keys(value)` instead?

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+cacheDir = path.join(fs.absolute('~/.atom'), 'compile-cache')
+jsCacheDir = path.join(
+    cacheDir,
+    create6to5VersionAndOptionsDigest(to5.version, defaultOptions),
+    'js')
+
+getCachePath = (sourceCode) ->
+  digest = crypto.createHash('sha1').update(sourceCode, 'utf8').digest('hex')
+  return path.join(jsCacheDir, "#{digest}.js")
```
kevinsawicki: No `return` keyword needed here.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+getCachedJavaScript = (cachePath) ->
+  if fs.isFileSync(cachePath)
+    try
+      cachedJavaScript = fs.readFileSync(cachePath, 'utf8')
+      stats.hits++
+      return cachedJavaScript
+  return null
```
kevinsawicki: No `return` keyword needed here.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+# Returns the 6to5 options that should be used to transpile filePath.
+createOptions = (filePath) ->
+  options =
+    'filename': filePath
+  for key, value of defaultOptions
+    options[key] = value
+  return options
```
kevinsawicki: No `return` keyword needed here.

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+# Function that obeys the contract of an entry in the require.extensions map.
+# Returns the transpiled version of the JavaScript code at filePath, which is
+# either generated on the fly or pulled from cache.
+loadFile = (module, filePath) ->
+  sourceCode = fs.readFileSync(filePath, 'utf8')
+  if not (sourceCode.startsWith('"use 6to5"') or sourceCode.startsWith("'use 6to5'"))
```
kevinsawicki: We typically use `unless` instead of `if not` because then you don't need the `()`

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+    try
+      fs.writeFileSync(cachePath, js)
+    catch error
+      console.error('Error writing to cache at %s: %o', cachePath, error)
+      throw error
+
+  module._compile(js, filePath)
+
+register = (appVersion) ->
```
kevinsawicki: `appVersion` appears to be unused so :fire: it

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+    try
+      cachedJavaScript = fs.readFileSync(cachePath, 'utf8')
+      stats.hits++
+      return cachedJavaScript
+  return null
+
+# Returns the 6to5 options that should be used to transpile filePath.
+createOptions = (filePath) ->
+  options =
+    'filename': filePath
```
kevinsawicki: I'd say drop the quotes and make this a single line object since it is pretty simple

```coffee
options = filename: filePath
```

src/esnext.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+create6to5VersionAndOptionsDigest = (version, options) ->
+  shasum = crypto.createHash('sha1')
+  # Include the version of 6to5 in the hash.
+  shasum.update('6to5-core', 'utf8')
+  shasum.update('\0', 'utf8')
+  shasum.update(version, 'utf8')
+  shasum.update('\0', 'utf8')
+  updateDigestForJsonValue(shasum, options)
+  return shasum.digest('hex')
```
kevinsawicki: No `return` keyword needed here.

spec/esnext-spec.coffee
```diff
@@ -0,0 +1,27 @@
+{create6to5VersionAndOptionsDigest} = require '../src/esnext'
+crypto = require 'crypto'
+
+describe "esnext", ->
+
+  it "::create6to5VersionAndOptionsDigest", ->
```
kevinsawicki: Usually we put the method name in a `describe` and then the `it` would be something like:

```coffee
it "returns a digest for the library version and specified options", ->
```

src/esnext.coffee
```diff
@@ -0,0 +1,155 @@
...
+  # here. Because Atom uses a bleeding edge version of Node/io.js,
+  # I think this can include es6.arrowFunctions, es6.classes, and
+  # possibly others, but I want to be conservative.
+  blacklist: [
+    'useStrict'
+  ]
+
+  # Includes support for es7 features listed at:
+  # http://6to5.org/docs/usage/transformers/#es7-experimental-.
+  experimental: true,
```
kevinsawicki: :fire: `,`

bolinfest:
@kevinsawicki Sorry, I'm not used to using GitHub for code review. Before, I was just squashing my changes, but now I've been responding by stacking commits. What's the best way to amend a PR so it's easiest for you to review?

kevinsawicki:
> What's the best way to amend a PR so it's easiest for you to review?

No need to squash or amend/force-push, just keep the commits rolling.

src/esnext.coffee
```diff
@@ -0,0 +1,154 @@
...
+
+    try
+      fs.writeFileSync(cachePath, js)
+    catch error
+      console.error('Error writing to cache at %s: %o', cachePath, error)
+      throw error
+
+  module._compile(js, filePath)
+
+register = () ->
```
kevinsawicki: :fire: `()`

bolinfest:
@kevinsawicki When I stack commits, is there an easy way to see the full diff from the base of my PR to the head that I don't know about?

kevinsawicki:
> When I stack commits, is there an easy way to see the full diff from the base of my PR to the head that I don't know about?

Not sure exactly what you mean here, but you mean something different than the diff shown in https://github.com/atom/atom/pull/5299/files ?

src/esnext.coffee
```diff
@@ -0,0 +1,154 @@
...
+  require.extensions['.js'] = loadFile
```
kevinsawicki: What do you think about doing this technique here for consistency with the coffee cache?

https://github.com/atom/atom/blob/e0bac77fa5cda1fefe55a00531d51116a3a37b23/src/coffee-cache.coffee#L57-L61

src/esnext.coffee
```diff
@@ -0,0 +1,154 @@
...
+
+module.exports =
+  register: register
+  getCacheMisses: -> stats.misses
+  getCacheHits: -> stats.hits
```
kevinsawicki: Down the road, if you want to see these in Timecop, they can be added [here](https://github.com/atom/timecop/blob/12df77d91c7686e65d9825ffdb4db506c70b4b25/lib/cache-panel-view.coffee#L26-L29)

bolinfest: Good idea: I'll put up a separate PR for that once this is in.

bolinfest:
@kevinsawicki Also, is there any way to do anything like `apm test spec/esnext-spec.coffee`?

And thanks, yes, https://github.com/atom/atom/pull/5299/files was the thing I was looking for.

kevinsawicki:
> Also, is there any way to do anything like apm test spec/esnext-spec.coffee?

Make the method you want to run `fit` intead of `it` or `fdescribe` instead of `desribe` and run `Window: Run Package Specs` from the command palette to launch the spec window.

The `f` stands for focused, only specs with an `f` will run when one is present.

nathansobo:
Would you be down to rename `esnext.coffee` `6to5.coffee`? It seems like a better match for the library we're using and more clear.

bolinfest:
Do all of the commits in the PR get squashed before it is accepted? If so, what should I edit to update the commit message for what will ultimately go in?

nathansobo:
> Do all of the commits in the PR get squashed before it is accepted?

We've never done that sort of thing. I'll defer to @kevinsawicki for the merge decision.

kevinsawicki:
> Do all of the commits in the PR get squashed before it is accepted?

Nope, we just click the merge button, so they end just how they look in the **Commits** tab of this PR.

bolinfest:
@kevinsawicki Hrm, ok, I'll be more cognizant of that fact in the future. The commit at the bottom of the stack has references to `esnext` that are no longer accurate. I'm happy to sqaush 'n scrub or just leave things as-is.

kevinsawicki:
>  I'm happy to sqaush 'n scrub or just leave things as-is.

I'd say just leave it, it's all about the journey :mountain_bicyclist:

package.json
```diff
@@ -19,6 +19,7 @@
   ],
   "atomShellVersion": "0.21.0",
   "dependencies": {
+    "6to5-core": "3.0.14",
```
sebmck: Might be a good idea to have this as `^3.0.14` as I push out releases pretty often and don't want atom users to be disadvantaged.

kevinsawicki: :+1:

nathansobo:
Updated the title of the PR for clarity.

paulcbetts:
Pretty excited about this PR, great work @bolinfest!

kevinsawicki:
:ship: :chart_with_upwards_trend:

paulcbetts:
@bolinfest Are you going to do a follow-up with build precompilation for non-dev mode? That would be :heart_eyes: :gem:

kevinsawicki:
> Are you going to do a follow-up with build precompilation for non-dev mode? 

`apm install` will automatically warm the cache at the end of install.

paulcbetts:
@bolinfest @kevinsawicki I would still take a PR for that for atom/atom-shell-starter if you're Feeling Brave, so that people can write Atom Shell apps in ES6

kgrossjo:
Oh yes!  Writing Atom Shell apps in ES6 would be awesome!

paulcbetts:
@kgrossjo I Think So Too™  atom/atom-shell-starter already has support for it today in live-compile mode!

steelbrain:
I have a question, If Atom has switched to IoJs <sup>[[1]]</sup>, then why do we need to transpile ES6, Can't we use it directly?
And If we really want to use a transpiler, Then why not use the `react-tools` transpiler. It supports ES6 stuff, as well as TypeHinting. (I use typehinting in all my JS projects, it makes stuff easier for Humans & IDEs)
[1]:https://github.com/atom/atom/issues/5383#issuecomment-72743578

LegNeato:
Because the ES6 (and ES7!) support is uneven and will change as we update V8. This insulates package authors from that while letting them use more than native V8 support today.

@bolinfest will have to speak to react-tools but seeing how we are at FB and and work together with the React folks on tooling there likely is a good reason :-)

sebmck:
@steelbrain You check out a compatibility table [here](https://kangax.github.io/compat-table/es6/) for a list of supported features. react-tools and io.js don't even come close. 6to5 supports JSX out of the box and is completely compliant with the JSX transformer in the official react-tools.

eyko:
Isn't anybody worried about adding such a specific implementation mode to all your es6 files? The user's (or team's) _current_ transpiler preference shouldn't be reflected in the source code. This is something that should be decided by the build system and maybe specified via a setting (either in the atom settings panel or by reading `package.json` or whatever).

I'm not sure how much sense it makes to switch from `'use 6to5';` to `'use traceur';` when suddenly people decide they prefer traceur. It's a big search/replace job and it pollutes the commit history with what is essentially change in the build config. I personally think it would make more sense to just use `"use es6";` or something along those lines, and then abstract away the implementation details. If someday Atom decides to support switching between 6to5 and traceur, developers could simply add an entry in package.json. It also makes it easier to temporarily switch transpilers for testing/curiosity/whatever reasons.

sebmck:
@eyko You seem to be conflating the 6to5 support with just ES6 support. 6to5 supports [so](http://6to5.org/docs/usage/experimental/) [much](http://6to5.org/docs/usage/playground/) [more](http://6to5.org/docs/usage/jsx/).

eyko:
It was an example, I do know that 6to5 is not strictly ES6. Call it `"use
esnext";` if you would, hopefully you get my point.

eyko:
@sebmck it was an example, I do know that 6to5 is not strictly ES6. Call it `"use esnext";` if you would, hopefully you get my point.

benogle:
> Isn't anybody worried about adding such a specific implementation mode to all your es6 files?

We talked about this quite a bit. We wanted to be more general about the naming. `"use esnext";` would have been nice, but would have been misleading since the chosen transpiler supports only a subset of es6, and even some of [es7](http://kangax.github.io/compat-table/es7/). With a generic pragma, the author wouldnt know which subset was available, and may expect that all of es6 or es7 is supported. Then what if we decided to change the transpiler under the hood? It would have different levels of support and code written against the original transpiler may not work, or work differently. 

With `"use 6to5";` you know exactly what you're getting. If we decide to support `traceur` in the future, it can be alongside `6to5`, and you can migrate when you're ready. 

I agree it sucks to have the a pragma tied to a specific tool, but we believe it is the best way forward given the spotty support for es6.

eyko:
Okay, so that's a very convincing argument with which I agree, I guess. (now I'm wondering what this will mean for developers already using other pragmas but wanting to take advantage of the build feature...)

josh:
FWIW, Rails 5 + Sprockets is shipping 6to5 with the `.es6` extension approach. The extension name maybe subject to change, but for performance reasons detecting with just the filename available works better than a contents scan. Not that Atom needs to do the same thing, but just wanted to mention that.

nathansobo:
@josh Yeah we went back and forth on it. The main issue is we couldn't come up with an extension that felt truthful, since 6to5 isn't necessarily ES6. An extension does have the advantage of being detectable from the file path alone, but since since reading the file anyway in every circumstance in which it matters, looking at the first line isn't an issue.

PR #5314, opened 2015-01-30
Make rpm install package a little more flexible

eestrada:
:penguin: Took a couple things that were hard coded to be absolute paths and made them relative. This makes it much easier to install Atom in a non-standard location on Linux and have it just work.

However, with default settings on the build scripts, everything still behaves the same as before these changes.


kevinsawicki:
Cool, thanks for this :ship:

PR #5316, opened 2015-01-30
Badge component

### LINKED ISSUE #5271: Add badge UI component
It seems like it would be really useful to have some kind of badge component ([like Bootstrap's](http://getbootstrap.com/components/#badges)). The most immediate use for this I can think of would be any of the numbers in the headings in Settings > Packages (or Themes). They're currently in parentheses, but that's really just a text approximation of a badge anyway. :smile:

Something like:
screenshot 2015-01-27 09 11 47

Instead of:
screenshot 2015-01-27 09 12 54

(I'm opening this here and not in [settings-view](https://github.com/atom/settings-view) because this seems like a useful UI component to add to core, rather than a styling just useful for settings-view.)

simurai:
This PR adds a badge component to core and removes the import of the Bootstrap badge. Issue #5271.

screen shot 2015-01-30 at 8 08 54 pm

### DIFF
--- a/static/atom.less
+++ b/static/atom.less
@@ -11,6 +11,7 @@
 
 @import "workspace-view";
 @import "bootstrap-overrides";
+@import "badges";
 @import "buttons";
 @import "icons";
 @import "links";
--- a/static/badges.less
+++ b/static/badges.less
@@ -36,14 +36,17 @@
     min-width: @size + @padding*2;
     padding: @padding round(@padding*1.5);
   }
-  .badge-size();
+  .badge-size(); // default
+
+  // Fixed size
   &.badge-large  { .badge-size(18px); }
   &.badge-medium { .badge-size(14px); }
   &.badge-small  { .badge-size(10px); }
 
-  // TODO: un-dry this, no rouding of badge-size()
-  // adjusts to parent element, best used for larger sizes
-  &.badge-parent {
+  // Flexible size
+  // The size changes depending on the parent element
+  // Best used for larger sizes, since em's can cause rounding errors
+  &.badge-flexible {
     @size: .8em;
     @padding: @size/2;
     font-size: @size;
--- a/static/badges.less
+++ b/static/badges.less
@@ -7,9 +7,7 @@
   font-weight: normal;
   text-align: center;
   white-space: nowrap;
-  color: @text-color;
   border-radius: 1em;
-  background-color: @background-color-highlight;
 
   &:empty {
     display: none; // Hide when un-used
--- a/static/badges.less
+++ b/static/badges.less
@@ -1,0 +1,64 @@
+@import "ui-variables";
+
+.badge {
+  display: inline-block;
+  line-height: 1;
+  vertical-align: middle;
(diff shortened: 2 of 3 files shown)


benogle:
Looks good :+1:

mnquintana:
Looks great! :+1:

I just have one question – I noticed that badges with a single character in them tend to produce a circle, instead of a "pill" or "capsule" shape – was that intentional?

simurai:
> I just have one question – I noticed that badges with a single character in them tend to produce a circle, instead of a "pill" or "capsule" shape – was that intentional?

Yes, that's intentional. Good or bad? I wouldn't mind some more padding on the sides, but for single characters, if it's almost a circle but not quite, it kinda feels odd.

simurai:
Documented another feature. If you add the `badge-flexible` class, the size of the badge changes depending on the parent's size (using `em`s). For example if a badge is used inside an `<h1>`, the badge becomes bigger. Settings will look like this:

screen shot 2015-01-31 at 11 20 41 am

mnquintana:
> Good or bad? I wouldn't mind some more padding on the sides, but for single characters, if it's almost a circle but not quite, it kinda feels odd.

I'm not sure. I think I'm experiencing an optical illusion – even though the circle badges should be the same height as the capsule badges, they *seem* taller, and a little out of place. 

This has happened to me before with typography and optical alignment (versus "perfect" alignment). Say you're trying to center a fragment of text – depending on the characters used / capitalization, even if it's technically perfectly centered in the center of a document, it could *look* un-centered because of the shapes of the characters. I think something like that might be happening here?

simurai:
Could it be that it depends on the number? The `5` in the screenshot does look a bit strange. Here a couple different ones:

screen shot 2015-01-31 at 3 10 20 pm

mnquintana:
@simurai Oh sorry I meant when they were all in a line:

image

For some reason the single character badge (`'3'`) looks to me a little taller than the others, even though I'm sure it's the same height, because of the circle shape.

simurai:
Maybe the "roundness" makes it seem more "pointy" and thus taller? Here just a bit more padding to make them oval.

screen shot 2015-02-03 at 10 17 34 am

screen shot 2015-02-03 at 10 18 38 am

Hmm.. the "almost" a circle makes them look a bit weird when stand-alone.

Ohh.. wait. Maybe it's the color? Here they are all green:

screen shot 2015-02-03 at 10 12 28 am

Yes, I would say better, right?

mnquintana:
Ah yeah I see what you mean, they definitely look better as circles than as "almost circles." And it does look fine all the same color. Looks good to me :+1:

simurai:
Phew... :sweat_smile:

kevinsawicki:
Is this ready to :ship:?

simurai:
Yes, I think it's good to go. Next to the settings, the only other place I found them being used is in the Select List:

screen shot 2015-02-06 at 10 15 37 am

Maybe some other packages use them, but but since the class name stays `.badge`, it should be fine there too.

PR #5330, opened 2015-01-31
:memo: Fix a broken link.

xfq:
Fix a broken link in `docs/creating-a-theme.md`.

### DIFF
--- a/docs/creating-a-theme.md
+++ b/docs/creating-a-theme.md
@@ -24,7 +24,8 @@ a few things before starting:
   is used to help distribute your theme to Atom users.
 * Your theme's _package.json_ must contain a `"theme"` key with a value
   of `"ui"` or `"syntax"` for Atom to recognize and load it as a theme.
-* You can find existing themes to install or fork on [atom.io](atomio).
+* You can find existing themes to install or fork on
+  [atom.io](https://atom.io/themes).
 
 ## Creating a Syntax Theme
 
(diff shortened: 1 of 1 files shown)


docs/creating-a-theme.md
```diff
@@ -24,7 +24,8 @@ a few things before starting:
   is used to help distribute your theme to Atom users.
 * Your theme's _package.json_ must contain a `"theme"` key with a value
   of `"ui"` or `"syntax"` for Atom to recognize and load it as a theme.
-* You can find existing themes to install or fork on [atom.io](atomio).
+* You can find existing themes to install or fork on
+  [atom.io](https://atom.io/themes).
```
izuzak: Suggestion: instead of defining the URL for the link here, define it at the bottom similar to how it's done for other links.

Notice that there's already [an `atomio` URL defined at the bottom](https://github.com/atom/atom/pull/5330/files#diff-f90dd88c45fac608d3fc2fc043a4dde4R134):

```
[atomio]: http://atom.io/packages
```

But it should be really pointing to:

```
[atomio]: http://atom.io/themes
```

So, I think this broken link is actually caused by two typos. Instead of `[atom.io](atomio)` use `[atom.io][atomio]`, and instead of `[atomio]: http://atom.io/packages` use `[atomio]: http://atom.io/themes`.

What do you think?

PR #5336, opened 2015-01-31
Fixes RPM install path and icon location

### LINKED ISSUE #5331: RPM package Icon is using absolute paths, breaks icon themes
Currently the generated RPM includes the absolute path to the icon into the .desktop file. This should be just the icon name to be compatible with icon-themes.

edgard:
This makes Atom a better desktop citizen relocating to where the usual install
directory is (like on the debian package) and also fix the icon using absolute
paths, breaking icon-themes.

Fixes #5331

### DIFF
--- a/build/tasks/mkrpm-task.coffee
+++ b/build/tasks/mkrpm-task.coffee
@@ -30,9 +30,9 @@ module.exports = (grunt) ->
     rm rpmDir
     mkdir rpmDir
 
-    installDir = grunt.config.get('atom.installDir')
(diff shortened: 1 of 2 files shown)


eestrada:
Great work! One hitch though; the icons aren't actually copied to a place where any desktop environments can find them (i.e. /usr/share/icons/hicolor/). I pulled from your repo and added this to the rpm build script. The additions are at pull request #5339.

edgard:
Closing. Use @eestrada pull request #5339 instead.

PR #5339, opened 2015-02-01
Copy atom.png to standard icon path locations for RPM build

eestrada:
:penguin: This completes a missing feature of pull request #5336 which in turn references issue #5331. Basically, the new version of the atom.desktop file from pull request #5336 allows the desktop environment to scan its icon path for the "atom" icon. This is great! The only catch is the icons weren't actually being placed in that path yet. This pull request solves that issue. :smile:

As an added bonus, this pull request replaces a hardcoded path in the atom.spec.in file that pointed to the default build location ("/tmp/atom-build"). This would break if someone builds to a non-standard location and tries to build an RPM from it. Now, regardless of where the initial build is done, the RPM build should work.

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -207,6 +207,7 @@ module.exports = (grunt) ->
       loadingGif: path.resolve(__dirname, '..', 'resources', 'win', 'loading.gif')
       iconUrl: 'https://raw.githubusercontent.com/atom/atom/master/resources/win/atom.ico'
       setupIcon: path.resolve(__dirname, '..', 'resources', 'win', 'atom.ico')
+      remoteReleases: 'https://atom.io/api/updates'
 
     shell:
       'kill-atom':
--- a/build/package.json
+++ b/build/package.json
@@ -12,7 +12,7 @@
     "fs-plus": "2.x",
     "github-releases": "~0.2.0",
     "grunt": "~0.4.1",
-    "grunt-atom-shell-installer": "^0.20.0",
+    "grunt-atom-shell-installer": "^0.21.0",
     "grunt-cli": "~0.1.9",
     "grunt-coffeelint": "git+https://github.com/atom/grunt-coffeelint.git#cfb99aa99811d52687969532bd5a98011ed95bfe",
     "grunt-contrib-coffee": "~0.12.0",
--- a/build/tasks/publish-build-task.coffee
+++ b/build/tasks/publish-build-task.coffee
@@ -68,7 +68,7 @@ getAssets = ->
       ]
     when 'win32'
       assets = [{assetName: 'atom-windows.zip', sourcePath: 'Atom'}]
-      for squirrelAsset in ['AtomSetup.exe', 'RELEASES', "atom-#{version}-full.nupkg"]
+      for squirrelAsset in ['AtomSetup.exe', 'RELEASES', "atom-#{version}-full.nupkg", "atom-#{version}-delta.nupkg"]
         cp path.join(buildDir, 'installer', squirrelAsset), path.join(buildDir, squirrelAsset)
         assets.push({assetName: squirrelAsset, sourcePath: assetName})
       assets


edgard:
:+1: Beautiful! Much better!

docs/build-instructions/linux.md
```diff
@@ -21,7 +21,7 @@ Ubuntu LTS 12.04 64-bit is the recommended platform.
 
 ### Fedora / CentOS / RHEL
 
-* `sudo yum --assumeyes install make gcc gcc-c++ glibc-devel git-core libgnome-keyring-devel rpmdevtools`
+* `sudo yum --assumeyes install make gcc gcc-c++ glibc-devel git-core libgnome-keyring-devel rpmdevtools ImageMagick`
```
kevinsawicki: Just curious why `ImageMagick` was added here, what depends on it?

eestrada:
ImageMagick is included in order to convert the images to sizes that most desktop environments can use in "/usr/share/icons/hicolor/". The icon in the atom/atom repo is 1024x1024, which no Linux desktops will recognize, as far as I can tell. Thus, the included atom.png file is resized at rpm build time to play nicely with desktop environments. ImageMagick is used for this. It is often pre-installed with most Linux distributions.

One alternative option is to include the varying icon sizes in the git repo, but that seems like a redundancy that is unneeded, since they can be dynamically generated at build time.

The last option I can think of is I can try to modify the build rpm.spec.in build script so that it will create them if ImageMagick is installed, but will skip it otherwise. If they are not generated, the icon will still show up in the application, but it won't show up in any menus in the desktop; it will be whatever icon the desktop falls back on.

thedaniel:
> One alternative option is to include the varying icon sizes in the git repo, but that seems like a redundancy that is unneeded, since they can be dynamically generated at build time.

This is often preferable because large assets don't always resize / look nice at smaller sizes and are usually adjusted manually by a designer to change / remove details so they look right when very small. I don't necessarily think it's a big deal in terms of the RPM package but I want to get @simurai's thoughts on this - I think these assets might actually already exist.

eestrada:
I had noticed that that at really small sizes the icon got a bit mushy since the contrast between the white and the green becomes relatively low, so it would be great to have alternate designs for varying sizes. However, if they exist, I haven't been able to find them in the repo yet, in either png or svg format. I looked for a while before implementing this particular solution; I really didn't want to add another dependency if I could help it.

simurai:
Yes, the icons vary slightly depending on the size. For example in the 32px and 16px version, the white border is removed.

screen shot 2015-02-03 at 9 05 21 am

For OS X and Windows, the different sizes are "backed in" the [.icns](https://github.com/atom/atom/blob/master/resources/mac/atom.icns) and [.ico](https://github.com/atom/atom/blob/master/resources/win/atom.ico) file.

For Linux I could add all the `.png`s needed. Not sure where, maybe to `atom/resources/linux/icons/`? Currently these sizes are already made:

screen shot 2015-02-03 at 9 18 50 am

build/tasks/mkrpm-task.coffee
```diff
@@ -30,9 +30,9 @@ module.exports = (grunt) ->
     rm rpmDir
     mkdir rpmDir
 
-    installDir = grunt.config.get('atom.installDir')
+    installDir = '/usr'
```
kevinsawicki: Is there a reason this has to be hard-coded instead of allowing it to be specified?

edgard:
If you specify there you break the RPM package in multiple places. RPM already implements relocation feature for those that wish to install somewhere else. The current version package already supports this feature, it was implemented by @eestrada on #5314.

kevinsawicki:
> For Linux I could add all the .pngs needed. Not sure where, maybe to atom/resources/linux/icons/? 

Yeah, that sounds good, thanks @simurai

ardeshirj:
>RPM already implements relocation feature for those that wish to install somewhere else

Currently the recommended way to install the RPM package is through ```yum``` using ```localinstall``` argument. However the only way to change the installation directory is through ```rpm``` using ```--prefix``` argument. In this case, if someone want to install atom in different directory the only option would be  using ```rpm``` instead of ``yum`` 

Is the reason to move installation path from ```/usr/local/bin``` to ```/usr/bin``` just because the desktop environment wont' pick up the icons from ``` /usr/share/icons/hicolor/```?

edgard:
True, YUM doesn't support relocating packages. If someone wanted to relocate, that can be done only with RPM.

The installation path was moved because it's common for RPM apps to be installed on /usr instead of local. Actually checking on my setup (with multiple custom repos), I don't have a single file on local. The idea is that since it's package managed, it should feel 'native'.

The icons could be installed on /usr/local/share/icons also, it works. IIRC, the XDG base directory spec says it should be included on icon path search.

ardeshirj:
Then why not keep it the same, so it will be consistent with Ubuntu installation and leave it open to generate ```rpm``` package for custom installation directory? That way, all we need to change would be ```atom.spec``` and ```atom.desktop``` to pick up the new icons folder.

eestrada:
@ardeshirj , that seems like a good idea. We should probably also base the atom.spec "Prefix:" tag on the value of `grunt.config.get('atom.installDir')`, as well as all the paths in the atom.spec file, really. Then we get both a customized rpm build ***and*** an rpm that can be relocated at install time. As they are currently implemented, the paths are all hardcoded in atom.spec, so choosing a custom install dir through grunt would probably have just broken it anyway.

I can work on getting the atom.spec.in file in shape tomorrow.

edgard:
:+1: Great stuff, sounds good to me! Ping if you need some help @eestrada. :)

eestrada:
Thanks @edgard! I'll let you know.

eestrada:
If we are going to allow the RPM package build to fully control where the install location is through grunt, can we change the default RPM package that is downloadable from atom.io to point to "/usr" instead of "/use/share"? Are the install packages created by hand for each new release, or is there an automated build script somewhere we can modify?

PR #5347, opened 2015-02-01
Fix indent guides missing on whitespace only lines with invisible characters

driskell:
Using 0.175.0 on Mac OS X 10.10.2 (though likely affecting all platforms), the indent guides are missing on whitespace only lines. So when you start a new line which auto-indents the indent guides are visibly missing. Snapshot below:

screen shot 2015-02-01 at 17 59 23

This is down the fact that the "isWhitespaceOnly" detection only works if invisible characters is disabled. With invisible characters enabled, it sees the invisible characters and thinks the line does not contain only whitespace, thus disabling the guides.

This patch also means the whitespace only flag is stored, and there's no need to run any regex again when re-rendering the HTML for the line.

Snapshot after applying the patch:

screen shot 2015-02-01 at 18 01 01

### DIFF
--- a/src/lines-component.coffee
+++ b/src/lines-component.coffee
@@ -213,8 +213,7 @@ LinesComponent = React.createClass
     innerHTML = ""
 
     scopeStack = []
-    firstTrailingWhitespacePosition = text.search(/\s*$/)
-    lineIsWhitespaceOnly = firstTrailingWhitespacePosition is 0
+    lineIsWhitespaceOnly = line.isOnlyWhitespace()
     for token in tokens
       innerHTML += @updateScopeStack(scopeStack, token.scopes)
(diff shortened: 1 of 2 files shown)


src/lines-component.coffee
```diff
@@ -213,8 +213,7 @@ LinesComponent = React.createClass
     innerHTML = ""
 
     scopeStack = []
-    firstTrailingWhitespacePosition = text.search(/\s*$/)
-    lineIsWhitespaceOnly = firstTrailingWhitespacePosition is 0
+    lineIsWhitespaceOnly = line.isOnlyWhitespace()
```
driskell: FAO reviewer: This is the main issue. After substituteInvisibleCharacters has run the line text is no longer whitespace only and contains the invisible character glyphs.

My fix removes the regex check and simply calls isOnlyWhitespace on the TokenizedLine, which remembers if the original text in the buffer is whitespace only or not.

nathansobo:
Thanks very much for contributing this.

PR #5364, opened 2015-02-02
Updated doc to specify package language

JoeUX:
I just added a sentence in the first paragraph: "You can write a package in Javascript or CoffeeScript."

This doesn't seem to be declared anywhere. We can apparently write packages in JS but the word 'Javascript' is nowhere in the doc. index.coffee would presumably imply to readers that packages are written in CoffeeScript, but this should be stated explicitly, and early. (And the JS option will be hugely important to many.)

### DIFF
--- a/docs/creating-a-package.md
+++ b/docs/creating-a-package.md
@@ -1,6 +1,6 @@
 # Creating Packages
 
-Packages are at the core of Atom. Nearly everything outside of the main editor
+Packages are at the core of Atom. You can write a package in Javascript or CoffeeScript. Nearly everything outside of the main editor
 is handled by a package. That includes "core" pieces like the [file tree][file-tree],
 [status bar][status-bar], [syntax highlighting][cs-syntax], and more.
 


docs/creating-a-package.md
```diff
@@ -1,6 +1,6 @@
 # Creating Packages
 
-Packages are at the core of Atom. Nearly everything outside of the main editor
+Packages are at the core of Atom. You can write a package in Javascript or CoffeeScript. Nearly everything outside of the main editor
```
50Wliu: Maybe `Packages can be written in either JavaScript or CoffeeScript.`?
Also, this line now looks a bit longer than the 80-line limit that Atom likes :).

izuzak: I like that suggestion, @50Wliu. :+1: 

@JoeArizona Also, what do you think about moving this sentence to one of the next paragraphs? I think it might fit a bit better there -- it breaks the flow of this paragraph a bit, and it's not the most important thing to mention at the start of the guide, I think.

izuzak:
Sorry for not getting to this sooner -- could you reopen this on the new repository for the docs: https://github.com/atom/docs. Most docs in this repository will be removed so that we don't have duplicated content around. :book:

PR #5374, opened 2015-02-03
Add section on "composed" commands to Advanced Keymap document

lee-dohm:
This is intended to address the commonly-asked question about making a keybinding execute multiple commands.

It also updates the example code for a command by removing the reference to the deprecated `EditorView` and just using the example from the `CommandRegistry` documentation.


benogle:
Thanks! :heart:

PR #5382, opened 2015-02-03
Add asar support in Atom

zcbenz:
By using [asar](https://github.com/atom/asar) to [package the `app` folder](https://github.com/atom/atom-shell/blob/master/docs/tutorial/application-packaging.md) into a single file,  we should be able to speed up the install time and startup time on computers that run on HDD significantly.

Known issues:
- [x] `index.js` not loading on Windows.
- [x] Can not interpret Less stylesheets because of graceful-fs not compatible with `asar`.
- [x] Program not quitting on Windows.
- [x] Tasks not working.
- [x] `apm` and `atom` commands can not be installed.
- [x] `apm` can not show built-in packages.


benogle:
WIN. :boom:

paulcbetts:
This PR is much shorter than I thought it would be, maybe we'll move to asar sooner rather than later

build/tasks/generate-asar-task.coffee
```diff
@@ -0,0 +1,14 @@
+asar = require 'asar'
+path = require 'path'
+
+module.exports = (grunt) ->
+  {rm} = require('./task-helpers')(grunt)
+
+  grunt.registerTask 'generate-asar', 'Generate asar archive for the app', ->
+    done = @async()
+
+    appDir = grunt.config.get('atom.appDir')
```
paulcbetts: Check to make sure this dir exists, otherwise if you run `script/grunt generate-asar` explicitly, you regenerate an empty asar file

paulcbetts:
This PR will forever and ever kill all of the long path problems on Windows too, :sparkles:

Zireael07:
Any progress on the last bullet point?

zcbenz:
Everything is working now in this branch.

thomasjo:
> Everything is working now in this branch.

Awesome! :shipit: :rocket:

Zireael07:
Great!

src/task.coffee
```diff
@@ -100,6 +100,11 @@ class Task
     @childProcess.removeAllListeners()
     @childProcess.on 'message', ({event, args}) =>
       @emit(event, args...) if @childProcess?
+    # Catch the errors that happened before task-bootstrap.
+    @childProcess.stdout.on 'data', (data) ->
+      console.log data.toString()
```
kevinsawicki: Should these be reported somewhere else? Or just ignored?

build/tasks/build-task.coffee
```diff
@@ -145,9 +145,9 @@ module.exports = (grunt) ->
     cp 'src', path.join(appDir, 'src'), filter: /.+\.(cson|coffee)$/
     cp 'static', path.join(appDir, 'static')
 
-    cp path.join('apm', 'node_modules', 'atom-package-manager'), path.join(appDir, 'apm'), filter: filterNodeModule
+    cp path.join('apm', 'node_modules', 'atom-package-manager'), path.resolve(appDir, '..', 'apm'), filter: filterNodeModule
```
kevinsawicki: Since the apm path is changing, does this need to be updated? https://github.com/atom/atom/blob/dd271aae960039b35ce20e2ecfcfc5bc751b42da/src/package-manager.coffee#L132

src/task.coffee
zcbenz: This only prints when there is output before the `task-bootstrap` script starts to run, e.g. when there is regression in atom-shell. So as long as atom-shell works as expected, there will be nothing printed.

zcbenz:
I just found that apm can not list built-in packages in asar archive, I'll try to fix it in apm.

Zireael07:
There was an apm update in the meantime, I think...

Can't wait for this, should significantly speed up install/update.

kevinsawicki:
So changing the `atom.sh` and `apm` paths means that after upgrading, the existing symlinks will be broken right?

So if the person previously ran `Atom > Install Shell Commands` and auth-ed the operation, they will be out of luck until they run it again?

zcbenz:
> So if the person previously ran Atom > Install Shell Commands and auth-ed the operation, they will be out of luck until they run it again?

Yeah I believe so. However if they can install the shell commands without privilege I think the commands will be updated silently.

zcbenz:
There are some random TextEditorComponent specs failing.

Zireael07:
Woot, merged?!

kevinsawicki:
Looks like this broke the symbols view so I'm going to revert it for now so it isn't broken on master, https://github.com/atom/symbols-view/issues/89

m1ga:
asar will create a tar like file archive, right? How many files would be packed together? So instead of the 11000 files in the windows /app folder you would have only one (136 MB) file?

kevinsawicki:
> asar will create a tar like file archive, right? How many files would be packed together? So instead of the 11000 files in the windows /app folder you would have only one (136 MB) file?

Yup, exactly this.

mark-hahn:
Does this mean we wouldn't be able to go in and modify code?

kevinsawicki:
> Does this mean we wouldn't be able to go in and modify code?

Yes, dev mode will still load from source though.

mark-hahn:
Luckily I'm always in dev mode.  I'm one of those hackers who can't leave
any code untouched.  It's a disease.

PR #3320, opened 2014-08-18
Portable version.

ghost:
If an .atom folder is present within the installation directory it will be used.
https://github.com/atom/atom/issues/2939

Issue #2199, opened 2014-05-13
Move .atom config to $XDG_CONFIG_HOME on linux


ReAzem:
From feedesktop.org:

$XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

Source: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Thanks!

kevinsawicki:
There is now full support for configuring the config location via the `ATOM_HOME` environment variable via #5385.

I'm going to close this issue out since changing the default location at this point is unlikely given how many Atom installs are already out there.

ReAzem:
So Atom will never ever respect the freedesktop standard? Its a shame, 1.0 wasn't released yet. Now should be the right moment to make things right.

kevinsawicki:
Switching to this standard would require reworking more than just one path since `XDG_CACHE_HOME` should be used for Atom's compile cache, and `XDG_RUNTIME_DIR` should probably be used for installed packages & themes.

Currently these are all stored under `~/.atom`. Having all of them co-located makes it incredibly easy to backup, transfer, and clean out your machine of all Atom related files. Also using the same path on all platforms makes Atom predictable for people using multiple platforms.

cgtx:
One of the benefits of XDG Base Directory Specification is that a user can back up configurations and data, but not cache.  Forcing all of Atom's data under one directory means that it is much more difficult to do this.

vially:
> using the same path on all platforms makes Atom predictable for people using multiple platforms

I think people using multiple platforms are accustomed with different operating systems having different file system layouts. If we were to take this reasoning to the extreme you might be tempted to install the Atom binary in a "/usr/Program Files" directory while on a Linux/Mac machine and I don't think there's anyone who would argue that that's just plain wrong.

Given that Atom is still not considered stable I think it's still not too late to change this to support the XDG standard. And while I agree that doing it properly requires a little bit more work to be fully compliant, I think that if the Atom team shows interest in supporting the standard, someone might be able to come up with a pull request to implement it.

malko:
:+1:

malko:
:+1: 
I'm working on different platforms and i'm always looking first for platform specific path in first place.

Dremora:
Changing default location should never be an issue as long as it's done in a backwards-compatible way, no matter how large the user base is.
[...]

PR #5385, opened 2015-02-04
Respect ATOM_HOME env var for portability

### LINKED ISSUE #2939: Add portable version of Atom
I don’t really stick to one computer to program, and some computers don’t have editors. Currently I use Sublime Text portable, but it would be great if there was a portable version of Atom.

kevinsawicki:
This PR unifies all references to use the `ATOM_HOME` environment variable as the location to store packages, compile caches, node headers, etc.

The location still defaults to `~/.atom` when unset.

This should make Atom portable meaning it will write no files outside this location.

Closes #2939

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -75,9 +75,9 @@ elif [ $OS == 'Linux' ]; then
   SCRIPT=$(readlink -f "$0")
   USR_DIRECTORY=$(readlink -f $(dirname $SCRIPT)/..)
   ATOM_PATH="$USR_DIRECTORY/share/atom/atom"
-  DOT_ATOM_DIR="$HOME/.atom"
+  ATOM_HOME="${ATOM_HOME:-$HOME/.atom}"
 
-  mkdir -p "$DOT_ATOM_DIR"
+  mkdir -p "$ATOM_HOME"
 
   : ${TMPDIR:=/tmp}
 
@@ -88,9 +88,9 @@ elif [ $OS == 'Linux' ]; then
     exit $?
   else
     (
-    nohup "$ATOM_PATH" --executed-from="$(pwd)" --pid=$$ "$@" > "$DOT_ATOM_DIR/nohup.out" 2>&1
+    nohup "$ATOM_PATH" --executed-from="$(pwd)" --pid=$$ "$@" > "$ATOM_HOME/nohup.out" 2>&1
     if [ $? -ne 0 ]; then
-      cat "$DOT_ATOM_DIR/nohup.out"
+      cat "$ATOM_HOME/nohup.out"
       exit $?
     fi
     ) &
--- a/apm/package.json
+++ b/apm/package.json
@@ -6,6 +6,6 @@
     "url": "https://github.com/atom/atom.git"
   },
   "dependencies": {
-    "atom-package-manager": "0.134.0"
+    "atom-package-manager": "0.135.0"
   }
 }
--- a/docs/customizing-atom.md
+++ b/docs/customizing-atom.md
@@ -100,6 +100,14 @@ namespaces: `core` and `editor`.
 
 You can open this file in an editor from the _Atom > Open Your Config_ menu.
 
+### Custom Configuration Location
+
+You can override the location that Atom stores configuration files and folders
+by setting the `ATOM_HOME` environment variable. The `ATOM_HOME` path will be
+used instead of `~/.atom` when it is set.
+
+This option can be useful when you want to make Atom portable across machines.
+
 ### Configuration Key Reference
 
 - `core`
--- a/docs/customizing-atom.md
+++ b/docs/customizing-atom.md
@@ -103,7 +103,7 @@ You can open this file in an editor from the _Atom > Open Your Config_ menu.
 ### Custom Configuration Location
 
 You can override the location that Atom stores configuration files and folders
-by setting the `ATOM_HOME` environment variable. The `ATOM_HOME` path will be
+in by setting the `ATOM_HOME` environment variable. The `ATOM_HOME` path will be
 used instead of `~/.atom` when it is set.
 
 This option can be useful when you want to make Atom portable across machines.
--- a/src/6to5.coffee
+++ b/src/6to5.coffee
@@ -96,7 +96,7 @@ getCachePath = (sourceCode) ->
 
   unless jsCacheDir?
     to5Version = require('6to5-core/package.json').version
-    cacheDir = path.join(fs.absolute('~/.atom'), 'compile-cache')
+    cacheDir = path.join(process.env.ATOM_HOME, 'compile-cache')
     jsCacheDir = path.join(cacheDir, 'js', '6to5', create6to5VersionAndOptionsDigest(to5Version, defaultOptions))
 
   path.join(jsCacheDir, "#{digest}.js")
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -263,9 +263,6 @@ class Atom extends Model
     # Make react.js faster
     process.env.NODE_ENV ?= 'production' unless devMode
 
-    # Set Atom's home so packages don't have to guess it
-    process.env.ATOM_HOME = configDirPath
-
     @config = new Config({configDirPath, resourcePath})
     @keymaps = new KeymapManager({configDirPath, resourcePath})
     @keymap = @keymaps # Deprecated
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -109,7 +109,7 @@ class Atom extends Model
   #
   # Returns the absolute path to ~/.atom
   @getConfigDirPath: ->
-    @configDirPath ?= fs.absolute('~/.atom')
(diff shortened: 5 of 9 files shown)


lee-dohm:
This could also perhaps be not a "fix" for #2199, but at least allow people who prefer the `.atom` directory to be stored someplace else to move it where they want it?

kevinsawicki:
> This could also perhaps be not a "fix" for #2199

Yeah, I was planning on closing that PR after merging this since it will now be properly configurable

batjko:
This is excellent. Will make a lot of people happy.  :+1:

CumpsD:
The installer still installs under ```AppData/Local/atom``` in Windows. Should it respect the env variable if it exists to be truly portable?

batjko:
@CumpsD It's only just been added to the repo. It won't be available in the installer(s) until the next official release, i.e. 178 I'm thinking.

kevinsawicki:
> The installer still installs under AppData/Local/atom in Windows.

This isn't changed by this pull request. The installer install location will stay the same. This change only affects the location that Atom stores config files, package, themes, and caches.

CumpsD:
So for a truly portable setup, one would need to install it, then copy it from that location to a USB stick and then uninstall atom again, I presume?

CumpsD:
Or just download https://github.com/atom/atom/releases/download/v0.178.0/atom-windows.zip and unzip would do as well I guess

atom.sh
```diff
@@ -75,9 +75,9 @@ elif [ $OS == 'Linux' ]; then
   SCRIPT=$(readlink -f "$0")
   USR_DIRECTORY=$(readlink -f $(dirname $SCRIPT)/..)
   ATOM_PATH="$USR_DIRECTORY/share/atom/atom"
-  DOT_ATOM_DIR="$HOME/.atom"
+  ATOM_HOME="${ATOM_HOME:-$HOME/.atom}"
```
joefitzgerald: Shouldn't path.sep be used instead of `/` here?

batjko: Good point, but does it matter in an .sh file?

joefitzgerald: Sorry, you're right - I spaced on the fact this is a shell script.

danwatco:
How exactly do I change the ATOM_HOME env var to make it portable then>

steelbrain:
@danwatco Not sure about how to do it on windows, but for bash, You can do it like this
```sh
# I am Bash Yo!
# ENV_VAR=VALUE COMMAND [ARGS]
ATOM_HOME=~/.atom_second atom
# Or
export ATOM_HOME=~/.atom_second
atom
```

danwatco:
@steelbrain I'm on windows :/

steelbrain:
@danwatco According to [this article][article], If you want to create a bat script, then you'll have to do something like this
```sh
set ATOM_HOME="D:\AtomHome"
atom
```
Otherwise, You just have to open my computer properties, find the "Environment Variables" option in the tabs. Open it and Add a UserLevel Env Var, with name `ATOM_HOME` and value of the target folder.
[article]:http://ss64.com/nt/set.html

danwatco:
@steelbrain Just tried that, it came up with an error in the console

steelbrain:
I am curious what that error was.

danwatco:
@steelbrain It went as follows: `Fatal error in V8: CALL_AND_RETRY_LAST Allocation failed - process out of memory`

steelbrain:
@danwatco Oo Lala, You might want to report it as a bug along with the repro steps.

danwatco:
@steelbrain I'll open it as a bug!

iliyang:
Please forgive my pedantry about this, but to me the requirement for an environment variable makes the application non-portable? As was suggested in #2939, if Atom would look in the application directory for an `.atom` folder (or `.config` or whatever), then it would be truly portable, meaning that you can plug the USB drive on a new machine and launch the executable without the need to touch any files or settings. To me, this is the definition of a portable app.

If I make a patch that implements the above approach, what's the chance of it being considered for merging into the repository?

Cheers!

kevinsawicki:
> If I make a patch that implements the above approach, what's the chance of it being considered for merging into the repository?

The application directory is completely replaced when Atom auto updates so if `.atom` is in there, it will be deleted when Atom next auto updates.

Perhaps there is another approach, or maybe this would only be supported for versions of Atom that don't auto update.

A PR that improves the portability of Atom would definitely be considered and appreciated though :+1:

iliyang:
Thanks for the reply, Kevin! OK, so if the application directory is wiped clean before updating Atom, then I guess a command-line option for setting the configuration directory is probably the next best thing. When I have some time I'll investigate how this can be achieved and prepare a PR.

kevinsawicki:
> When I have some time I'll investigate how this can be achieved and prepare a PR.

Sounds great, thanks for taking a look at this.

iliyang:
I just tried the "portable" Atom on my Windows 8.1 x64 by setting the `ATOM_HOME` environment variable. Upon launch, Atom writes both in the directory pointed by `ATOM_HOME` *and* areates a `%APPDATA%/Atom` directory. is there no way to make it only touch the `ATOM_HOME` dir?

danwatco:
@iliyang I've just re tried again by setting the atom home, and it works for me! Off a usb stick too!

iliyang:
@danwatco What do you mean it works? That the settings are properly stored in `ATOM_HOME` and preserved? I can believe that, but I still think that writing to %APPDATA% is undesirable and does against the definition of a predictable, portable app.

danwatco:
@iliyang Sorry, I was referencing that it worked as I tried before with no luck. Still some errors though with a package I'm trying to install. Yeah it makes the same files that you stated. It's not a problem for me, but I understand what you mean

iliyang:
Bump. Any hope of Atom only working in `ATOM_HOME` and not touching `%APPDATA%` ?

konfuzius:
@yuki-takei [Add portable version of Atom · Issue #2939 · atom/atom](https://github.com/atom/atom/issues/2939#issuecomment-111917501)

Windows 8.1
Atom 0.210.0

Works neatly.

.atom portable

```cmd
@echo off
pushd %~dp0
if not exist .atom mkdir .atom
if not exist .atom\AppData mkdir .atom\AppData
if not exist .atom\AppData\Roaming mkdir .atom\AppData\Roaming

:: Trick emmet or other plugin to think profile directory is in our data directory
set "USERPROFILE=%~dp0\.atom"

:: Set Atom base directory
set "ATOM_HOME=%~dp0\.atom"

:: Boot
start atom
```

alexchandel:
@iliyang You're absolutely right, this partial implementation does not count as portable. **All** temporary files must be written to Atom's folder to be portable. And by definition there can be no modification of the environment, not even via a batch-file unless it's shipped with Atom. In fact, the information that an Atom folder is portable must be entirely contained within that folder. If the updater clobbers the entire Atom folder and doesn't re-save that information, then this is a bug in the updater, by definition. The closure of #2939 is erroneous.

enociz:
Quick question, how exactly do you make your install of atom portable? Are there instructions somewhere?

ericwiseman123:
Sorry to ask this question but...
I have read a lot of posts on how to make atom portable I have gone through the  Post #8380 and #2199, #3320, #2939 and many other posts about this but I still guess I am not getting how to make the windows version of atom portable...
I have added the .atom file to the zip file that I have downloaded (by using the .bat file implementation that is posted in here. However I often switch back and forth from mac to windows to linux and I worry that just adding the .atom file will not work across the different systems.
So I want to set the ATOM_HOME  var (because that seems to be the way to do it) however I am really not sure on how to do that. 

So basically what I am looking for (sorry to be a pain) is a easier to follow step by step guide to making Atom portable (especially how to set the ATOM_HOME var).

Tapemaster21:
Thank you konfuzius! That seems to work wonderfully.

CraigCarboni:
Using "atom-windows.zip" extracted in my Downloads folder to "atom-windows" I created the following batch file which may be of use to others.

var example = true
@ECHO OFF
SETLOCAL ENABLEDELAYEDEXPANSION
:: NOTE: %~dp0 already has a trailing backslash
:: enclose in double quotes in case of folder names with spaces
:: %1 to open the file when associated with a file extension in windows
:: %1 to open a file when said file "dropped" onto this batch file
PUSHD "%~dp0"
SET atom=atom.exe
SET "ATOM_HOME=%~dp0.atom"

IF EXIST "%~dp0%atom%" (
IF NOT EXIST "%~dp0.atom" MKDIR "%~dp0.atom"
START "atom.io" "%~dp0%atom%" %1
) 

IF NOT EXIST "%~dp0%atom%" (
ECHO.
ECHO.
ECHO.   Warning: %~dp0%atom% is not found!
ECHO.
ECHO.   Is atom.bat in the correct folder?
ECHO.
PAUSE
EXIT
)
EXIT
var example = false

justathoughtor2:
@CraigCarboni Your batch file works perfectly! Thanks!

PR #5392, opened 2015-02-04
Make atom.spec.in use grunt variables

eestrada:
:penguin: Now rpms created with grunt will respect the --install-dir flag. Also, the rpm summary uses the grunt description variable. This is a partial implementation of issues discussed in PR #5339 . Resolving icons in the atom.desktop file using the system icon path will be taken care of in another PR.

Also, automated builds of rpm packages for upload to atom.io should now build against '/usr' instead of '/usr/local'.


kevinsawicki:
Awesome, thanks!

Issue #5354, opened 2015-02-02
What version of Jasmine is Atom using?


xfq:
In [vendor/jasmine.js](https://github.com/atom/atom/blob/52abb4afc9098454cea8e220a363be3a9b958934/vendor/jasmine.js#L2662), I found that the Jasmine version is 1.3. In [docs/writing-specs.md](https://raw.githubusercontent.com/atom/atom/2f62346c585361591e6a9de7349401c7ebe360eb/docs/writing-specs.md), I found links to both Jasmine 1.3 and 2.0. It would be nice to document the version of Jasmine used in Atom and correct the wrong link(s) in `writing-specs.md`.

as-cii:
:+1: 

This has the added benefit of avoiding to write helpers like `fit`, which are already provided by new versions (`it.only`).

[This guide](http://jasmine.github.io/2.0/upgrading.html) provides some help in the upgrade process. If you consider this a feasible opportunity for upgrading I am interested in contributing (e.g. I can take care of it).

Meanwhile I'll make some experiments to see what breaks without changing tests.

thedaniel:
Moving to 2.0 would be great, and we look forward to the results of your investigation ;) Some caveats:

- There are changes to how async specs work (though I'm not super familiar with the specifics) that will probably require a lot of big changes to the atom specs
- All community package specs are written to 1.3 so there would need to be some kind of mechanism to allow a package to opt-in to 2.0 and run 1.3 by default
- There are dozens of core package repos (everything from tree-view to markdown-preview) that will need spec updates or to be set to run on 1.3

cc @maxbrunsfeld as we briefly discussed this in chat

kevinsawicki:
I believe this was fixed by https://github.com/atom/atom/pull/5368

UziTech:
For anyone who is looking for a way to transition to jasmine 2.x I created a test runner that uses jasmine 2.x and falls back to the default jasmine 1.3 test runner for legacy tests. 

[atom-jasmine2-test-runner](https://www.npmjs.com/package/atom-jasmine2-test-runner)

PR #5393, opened 2015-02-04
Upgrade to Jasmine 2.1.1

as-cii:
Partly related to #5354 I decided to investigate how difficult could it have been to upgrade Jasmine to 2.1.1.

:warning: **Please note that this is an experimental branch and there are a bunch of features that do not work yet (including some specs which do not pass due to Jasmine breaking changes)** :warning: 
### What doesn't work
- [ ] `TerminalReporter`. I excluded it for the time being and I am using Jasmine's `ConsoleReporter` 'cause it seems to provide almost all functionalities. On the other hand it seems like `apm test` doesn't exit (even when specs have finished) and I suppose it may be related to this; 
- [ ] `TimeReporter`. 2.0 introduced a new way of creating custom reporters and this has to be adapted as well.
- [ ] Old async specs. I provided some examples (e.g. upgrading some specs) about how Jasmine 2 is supposed to work with asynchronous tests. You can find more information [here](https://github.com/jasmine/jasmine/issues/526).
- [ ] Old custom matchers. The syntax has changed for those as well. I updated `vendor/jasmine-jquery.js` to allow some of the matchers to be compatible with the new specification, but there are still a bunch of them which aren't ported yet;
- [ ] Old spies expectation. Many expectation have changed since 1.3 and now need to be upgraded (`andCallThrough` vs. `and.callThrough`)
- [ ] `setIncludedTags`. This should filter out some specs as far as I understand, but Jasmine doesn't seem to provide an alternative. Needs further investigation. 
- [ ] `waits(0)`. This served as a way of yielding to the UI thread. `waits` method does not exist anymore, and therefore we need to find another solution.
- [ ] `setSpecField`. `currentRunner()` method has been deprecated, thus we need to find another solution. I wonder if this is still needed, therefore it would be great if you could provide some feedback :pray:
- [ ] Suspicious Jasmine initialization. I am not 100% sure the solution below is the best practice to load Jasmine, but I couldn't find any other feasible initialization strategy. Needs investigation but it should be straightforward.
  
  ``` coffee
  Jasmine = require "jasmine"
  window.jasmine = (new Jasmine).jasmine
  
  {ConsoleReporter} = require "jasmine"
  ```
- [ ] Running test via UI. For now I have only used `apm test` to run tests, but we need to make sure that `AtomReporter` works fine as well.

I am sure I left out many other things which are not evident to me (it's my first contribution to Atom, as you may already know), therefore please feel free to leave a comment wherever you find something suspicious so that I can update incrementally this main post to keep track of what's missing.
### Discussion

As @thedaniel mentioned on #5354 one of the main concerns of upgrading is to keep package specs working. Quoting his statement:

> All community package specs are written to 1.3 so there would need to be some kind of mechanism to allow a package to opt-in to 2.0 and run 1.3 by default

A possible solution may be to have two separate suites:
- A jasmine 2 bootstrap, which loads all the specs that are known to be compatible (package owners may tag such specs somehow, maybe a filename suffix or a subdirectory could be sufficient);
- A jasmine 1.3 bootstrap (the default one), which will load any other spec;

I am still not sure how difficult could it be to make this approach work, but I hope we can discuss about it :smile:

---

I couldn't tackle this all on my own, especially because I cannot contribute to Atom full-time and this is my first contribution. On the other hand I hope this Pull Request can be a good starting point for us to go on and :shipit: !


sedabull:
+1 on this, though I'm not sure I can be of much assistance. I'm working on a new package right now, and I'd like to be able to use jasmine 2.x right now and only have to write my tests once, instead of writing a bunch of tests that will have to be refactored in a month or two. The sooner I can opt-in to 2.0 the better :smile:

maxbrunsfeld:
@sedabull for the record, you won't have to upgrade your specs from jasmine 1.3 in a month or two, or anytime in the foreseeable future; even if we upgrade to jasmine 2, 1.3 will continue to be supported for a long, long time.

@as-cii thanks for investigating this. Indeed, it would be a lot of work for us to upgrade all of our specs to jasmine 2. Maybe you could help us come up with a incremental strategy for upgrading one test at time?

sedabull:
@maxbrunsfeld, I read over my post and realized I hadn't really communicated very well why I'd like to opt-in to 2.x as soon as possible. I'm new to both package development, and unit testing in general, and I'd really rather not have to learn jasmine *twice*. Nor do I want to write less than optimal specs now, and just be happy with them until I get around to learning a new API. If I'm going to learn something new, I'd like for it to have maximum utility for the longest possible amount of time. (Plus I really like shiny new things :smile:). Thank you for clarifying though, it's good to know that everything imaginable will be done to avoid any disastorous breaking changes from being introduced to the Atom environment. Just out of curiosity, how long do think it would be before 2.x is availiable to package developers, and could that be made a priority over upgrading all of the core specs to use the 2.x jasmine API?

thedaniel:
> how long do think it would be before 2.x is availiable to package developers

This honestly depends on community effort from folks like @as-cii as the core team only has time for guidance and code review for this for the forseeable future.

sedabull:
@thedaniel, it sounds like i'll have to just stop being so picky and suck it up :smile:. Is there anything (relatively easy) you can think of that I could do to help out?

mark-hahn:
Ask @as-cii <https://github.com/as-cii> ?

On Wed, Feb 4, 2015 at 3:55 PM, Seth David Bullock <notifications@github.com

sedabull:
@mark-hahn, a fantastic idea!
@as-cii, is there anything I can do to help make jasmine 2.x available to package developers ASAP?

as-cii:
> Indeed, it would be a lot of work for us to upgrade all of our specs to jasmine 2. Maybe you could help us come up with a incremental strategy for upgrading one test at time?

@maxbrunsfeld that's a perfectly reasonable concern: I agree with you it should be the first thing to consider before upgrading (branches like this one usually end up being way behind master, causing tons of other issues).

#### Approach n.1

I have made some experiments and it turns out that it should be possible to run two separate versions of jasmine side by side. One way is to disable globals export for jasmine 2.1.1 through a custom boot: to get a sense of what I am talking about, please [have a look at this repo](https://github.com/as-cii/many-jasmine-versions) which I created for you to consider. Of course there is quite a bit of work to do in Atom in order to support this: I am referring, in particular, to `spec-helper.coffee`, which performs a bunch of modifications to globals and therefore needs a more careful look.

**Benefits**: core and package developers would just need to follow a naming convention and specs will be run with the new *engine*. UX–wise this is the best option, in my opinion, as there won't be any difference between old and new specs from the "developer" perspective.

**Drawbacks**: I suspect this needs quite a bit of effort to implement and there may be some *risks* which would make this solution not resilient as it should be (see `spec-helper.coffee` notes above). 

#### Approach n.2

Alternatively, we may have another command (e.g. `atom --dev --test-v2`) which runs only a particular set of tests (e.g. `1.3`, `2.2`). This way we could load only the needed jasmine version and, gradually, port old matchers and spies as we convert old tests.

**Benefits**: I see this as a more resilient option; loading only one version effectively deletes the problem of incompatibility, globals aliasing, etc.

**Drawbacks**: UX gets a bit worser, given that as a package developer I'd need to run both my old tests and my new tests with two different commands (why should I write tests in two different versions? I don't know :stuck_out_tongue_closed_eyes:). CI scripts would need to be modified as well to include this new command.

I'd be glad to hear your opinion about these two approaches and/or to propose new ones, before proceeding in implementing one or the other :smile:

sedabull:
It's probably just because I'm a user, but I'm all for n.1! As far as naming conventions, what about a new field in `package.json`? Something like:
```json
"spec-dependancies": {
    "jasmine": "2.x"
}
```
I'm unsure of how hard this would be to implement, or how opposed the core team would be to it.

PR #5399, opened 2015-02-04
Reverted atom.desktop.in

eestrada:
I realize the debian package build depends on this. Although it
probably won't break anything, I would rather implement a solution
that doesn't change the behavior of the debian installer.


kevinsawicki:
Thanks

PR #5403, opened 2015-02-04
Let Linux packages fully determine executable path

eestrada:
:penguin: It is no longer hard coded into the atom.desktop.in file; the individual grunt tasks now determine how the executable path is built and accessed.


kevinsawicki:
:ship:

PR #5404, opened 2015-02-04
Add integration test to cover browser-side code

maxbrunsfeld:
This adds an integration test, `spec/integration/startup-spec.coffee` that tests Atom's handling of different command-line arguments. Currently, only a few simple cases are tested.

Questions:
- [x] Currently, `selenium-webdriver` has been added to the `dependencies`. We probably shouldn't require this module to be installed unless the user is running tests. Can we use `devDependencies`? _Moved the dependency to build/package.json_
- [x] Currently, the tests are skipped if `chromedriver` is not installed. Should we install `chromedriver` on CI? _We won't need to if we download it as part of the build; see below_
- [x] Should this test live in a separate suite so that it isn't run by default. The integration tests take about 14 seconds to run on my macbook pro. _Yes_
- [x] Is this even a good idea? I feel like something along these lines will help us in the future when developing the browser-side code. _Yes_
- [x] Make integration specs runnable separately from the main spec suite
  - [x] Make the integration specs _not_ run as part of the normal spec suite
  - [x] Make build script download `chromedriver` to the same location as `Atom.app`
  - [x] Make `run-specs` grunt task run the integration tests
  - [x] Publish new `grunt-download-atom-shell` version and upgrade it in atom
- [x] Fix the tests' flakiness

/cc @atom/core

### DIFF
--- a/build/tasks/spec-task.coffee
+++ b/build/tasks/spec-task.coffee
@@ -85,15 +85,27 @@ module.exports = (grunt) ->
     appPath = getAppPath()
     resourcePath = process.cwd()
     coreSpecsPath = path.resolve('spec')
+    chromedriverPath = path.join(resourcePath, "atom-shell", "chromedriver")
 
     if process.platform in ['darwin', 'linux']
       options =
         cmd: appPath
         args: ['--test', "--resource-path=#{resourcePath}", "--spec-directory=#{coreSpecsPath}"]
+        opts:
+          env: _.extend({}, process.env,
+            ATOM_INTEGRATION_TESTS_ENABLED: true
+            PATH: [process.env.path, chromedriverPath].join(":")
+          )
+
     else if process.platform is 'win32'
       options =
         cmd: process.env.comspec
         args: ['/c', appPath, '--test', "--resource-path=#{resourcePath}", "--spec-directory=#{coreSpecsPath}", "--log-file=ci.log"]
+        opts:
+          env: _.extend({}, process.env,
+            ATOM_INTEGRATION_TESTS_ENABLED: true
+            PATH: [process.env.path, chromedriverPath].join(";")
+          )
 
     spawn options, (error, results, code) ->
       if process.platform is 'win32'
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "scoped-property-store": "^0.16.2",
     "scrollbar-style": "^2.0.0",
     "season": "^5.1.2",
+    "selenium-webdriver": "^2.44.0",
     "semver": "2.2.1",
     "serializable": "^1",
     "service-hub": "^0.2.0",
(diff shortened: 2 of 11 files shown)


nathansobo:
I think test coverage for this area of our system is probably a good idea. It will have a cost in terms of extra build complexity, but it will be great to have confidence in browser to render process interactions. I wonder if there's also a role for browser process unit tests so we don't end up testing everything browser-related with the integration tests. That's another day though.

src/browser/atom-application.coffee
```diff
@@ -14,7 +14,7 @@ url = require 'url'
 {EventEmitter} = require 'events'
 _ = require 'underscore-plus'
 
-socketPath =
+DEFAULT_SOCKET_PATH =
```
kevinsawicki: I think we typically do `DefaultSocketPath` for things like this

package.json
```diff
@@ -58,6 +58,7 @@
     "scoped-property-store": "^0.16.2",
     "scrollbar-style": "^2.0.0",
     "season": "^5.1.2",
+    "selenium-webdriver": "^2.44.0",
```
kevinsawicki: We don't want to ship this with Atom itself right?

If not, it should go in `build/package.json` then.

spec/integration/startup-spec.coffee
```diff
@@ -0,0 +1,85 @@
+os = require "os"
+fs = require "fs"
+path = require "path"
+remote = require "remote"
+temp = require("temp").track()
+{spawn, spawnSync} = require "child_process"
+{Builder, By} = require("selenium-webdriver")
```
kevinsawicki: :fire: `()`

maxbrunsfeld:
Passed on CI after fixing some flaky parts. Gonna re-run it repeatedly before merging.

maxbrunsfeld:
Alright 4 in a row. After 8, I'm gonna :ship:.

mark-hahn:
That ship icon is not legible.  I have to blow up my page each time to see
it.  Of course my eyes aren't great.

lee-dohm:
@mark-hahn You can always just hover over the emoji to get the text equivalent.

maxbrunsfeld:
Alright, so several package specs have been failing intermittently on CI lately, but the core specs have now passed 12 times in a row, so I feel that they're ready to go.

build/tasks/spec-task.coffee
```diff
@@ -85,15 +85,27 @@ module.exports = (grunt) ->
...
+    chromedriverPath = path.join(resourcePath, "atom-shell", "chromedriver")
 
     if process.platform in ['darwin', 'linux']
       options =
         cmd: appPath
         args: ['--test', "--resource-path=#{resourcePath}", "--spec-directory=#{coreSpecsPath}"]
+        opts:
+          env: _.extend({}, process.env,
+            ATOM_INTEGRATION_TESTS_ENABLED: true
+            PATH: [process.env.path, chromedriverPath].join(":")
```
kevinsawicki: The `path.delimiter` constant is probably better here than hard-coding `':'`

nathansobo:
Thanks for taking the initiative and expanding our test coverage.

PR #5405, opened 2015-02-04
Pick up icons from icon path in RPM package build

eestrada:
:penguin: :package: This rounds out the issues addressed in PR #5339, PR #5336 and issue #5331. Now the ideal icon for a given resolution is available to the desktop environment.


eestrada:
:warning: I modified the the directory structure so that the new icons @simurai would be included in the build. However, now I am pretty sure they are included in every build (Linux, Mac and Windows). I doubt this is a big issue, but I wanted to make sure that it is known in case someone can provide a workaround so that extraneous data isn't copied around willy-nilly; I am personally not familar enough with the general build process yet to know how to take care of it.

kevinsawicki:
Would you mind renaming the icons back and adding an explicit copy for it in a similar pattern to this example?

https://github.com/atom/atom/blob/01bf346d0f1fb86fefaad66e80bc70dd66992bb3/build/tasks/build-task.coffee#L157-L158

It isn't a huge issue but it would be nice to not add any unnecessary size to the mac/windows distributions.

eestrada:
I would not mind at all. In fact I would prefer to move them back :smile: ; I just didn't know where to look for an example on how to copy them over for the build. The commit you pointed to should help. Thanks.

thedaniel:
Thanks for the notice @eestrada - it's probably not worth worrying about, it's one meg out of a hundred.

eestrada:
I agree @thedaniel, but at the same time, a meg here and meg there eventually adds up. All they are used for is copying over at rpm build time, then they just sit in the install directory doing nothing after that. I'd rather clean up it the now while I am already mucking around in this part of the code.

eestrada:
Ok, the icons are no longer copied into the main installed directory. They are only copied for the :package: build process and only on Linux. Let me know if there is anything else this PR needs to be copacetic.

kevinsawicki:
Looks great, thanks for this and all other solid PRs this week :ship: :ship: :ship:

eestrada:
:+1: No problem.

PR #5408, opened 2015-02-04
Add language-csharp to core packages

thedaniel:
C# is a pretty mainstream language to leave out of core. I think it at least meets the bar set by inclusion of language-clojure, language-mustache etc.


thomasjo:
Awesome! :sparkling_heart:

kevinsawicki:
This package has deprecations, let me fix em up real quick and publish a new version.

PR #5413, opened 2015-02-05
fix onWillDispatch asymmetry

hoosierhobbyist:
This PR come coupled with two questions for the core team. The first is `CommandRegistry::onWillDispatch` is undocumented, so does that mean it is either not a part of the public API, or soon to be deprecated? If so then please ignore this proposal. The second question, assuming this method is meant to be public, is doesn't it make sense to have a matching `::onDidDispatch`? I have a particular use case for this event. I'm currently writing a terminal emulator for Atom, and in it I'm defining a special `atom` builtin. It is essentially a small wrapper around `atom.commands.dispatch` that allows a user to send commands directly to atom via the command line. It works great, except that after the command has executed, my input element loses focus. This hook would allow me to do something like:

``` coffeescript
atom.commands.onDidDispatch -> document.querySelector('#quantum-shell-input').focus()
```

to return focus to my package after the command has executed.

### DIFF
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -174,6 +174,9 @@ class CommandRegistry
 
   onWillDispatch: (callback) ->
     @emitter.on 'will-dispatch', callback
+  
+  onDidDispatch: (callback) ->
+    @emitter.on 'did-dispatch', callback
 
   getSnapshot: ->
     snapshot = {}
@@ -223,6 +226,7 @@ class CommandRegistry
       for listener in listeners
         break if immediatePropagationStopped
         listener.callback.call(currentTarget, syntheticEvent)
+        @emitter.emit 'did-dispatch', syntheticEvent
 
       break if currentTarget is window
       break if propagationStopped


nathansobo:
This package you're working on sounds really cool, especially the `atom` built-in.

> The first is CommandRegistry::onWillDispatch is undocumented, so does that mean it is either not a part of the public API, or soon to be deprecated?

It's not part of the public API, but it's not going anywhere anytime soon. I'd be down to add another method if absolutely needed, but I have some questions about your use-case.

When you dispatch a command, what is causing the terminal to lose focus? Is some package handling the command and stealing it? Since commands are dispatched synchronously, have you tried just focusing your view immediately after dispatching the command like this?

```coffee
# in the terminal code...
atom.commands.dispatch(target, "command-from-the-terminal")
terminalElement.focus()
```

If the command had some async handler that steals focus then obviously that doesn't work, but adding the `onDidDispatch` variant seems like it wouldn't help that much either.

sedabull:
After playing with some commands what I've found is that most of them do not steal focus. A good example is `atom window:toggle-full-screen`. However there were two common commands that were causing issues, namely `atom tree-view:toggle` and `atom settings-view:open`. I took your advice and simply refocused my input element after dispatching the commands. `atom tree-view:toggle` now works as expected, however `atom settings-view:open` seemed to take a little longer to execute and did steal focus once the view had fully loaded. Correct me if I'm wrong here (I'm still trying to wrap my head around asynchrounous logic) but it sounds like what you're saying is I need to wait for the actual command to execute, not just be dispatched from `atom.commands`. If that is the case, is there an easy way to do that? All I can think to do right now is something like:
```coffeescript
# in the terminal code...
atom.commands.dispatch(target, "command-from-the-terminal")
setTimeout(terminalElement.focus, 50)
```
This seems less than elegant though, and is obviously still not guaranteed to work. (How long is too long to wait?)

nathansobo:
How about listening for the blur event and stealing focus back? You may need to do that in a timeout with a zero delay, but not totally sure.

nathansobo:
Of course how do you distinguish a genuine blur from an accidental one.

sedabull:
Yes, it would seem that any solution beyond simply waiting for some period of time would add an unacceptable level of complexity. The `setTimeout` solution may not be pretty, but it does *work*, and there's a lot to be said for that. I've tested as many commands as I can think of at a 100ms timeout, and it's not even noticeable at the human level and provides the desired behavior in all cases. I am satisfied with this solution, so this PR can be safely closed as far as I'm concerned. I would like to thank you @nathansobo, you've been very friendly and helpful to this newbie and I am very excited to publish this new feature in the next release of `quantum-shell`!

nathansobo:
Good luck! It sounds like an exciting package.

PR #5425, opened 2015-02-05
Introduce ProjectRoot.

bolinfest:
Introduce ProjectRoot.

Today, Project has three methods that return parallel arrays:

```
getPaths()
getDirectories()
getRepositories()
```

Instead, there should be one type, `ProjectRoot`, that encapsulates
all of this information. Then these three methods can be replaced
with one method:

```
getProjectRoots()
```

This commit introduces the `ProjectRoot` type and the `getProjectRoots()`
method. The other three methods are now defined in terms of `getProjectRoots()`.

Note that there are two outstanding methods on `Project` that assume a single
root that seem out of place:

```
relativize(fullPath)
contains(pathToCheck)
```

### DIFF
--- a/src/project-root.coffee
+++ b/src/project-root.coffee
@@ -1,0 +1,20 @@
+# A root folder of {Project}.
+#
+module.exports =
+class ProjectRoot
+  constructor: (@directory, @repo) ->
+
+  # Public: Returns the {Directory} that corresponds to the project root.
+  getDirectory: -> @directory
+
+  # Public: Returns the {GitRepository} for the project root or {null}.
+  getRepository: -> @repo
+
+  destroy: ->
+    @directory.off()
+    @destroyRepo()
+
+  destroyRepo: ->
+    if @repo
+      @repo.destroy()
+      @repo = null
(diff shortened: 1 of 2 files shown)


nathansobo:
Do we need a special type here, or are these facilities that all `Directory` objects could provide? For example, every directory could be associated with a repository, it just might be for a parent.

The `::relativize` method will work because it takes a full path, so we'd only make it relative to a project root that matches. Same is true with contains... if any project root contains, it can return true. We took the only show-stopper, `::resolve` out of the API before the freeze.

bolinfest:
Hi Nathan, so here's my thinking:

Pathwatcher is one of potentially many services that logically make sense
as something unique to a ProjectRoot. Once it's possible for a Directory to
be backed by a remote directory, other services such as "Find in project"
would also be better implemented as something available through a service
on ProjectRoot than an API available on every Directory instance.

I feel like Directory should have as small and specific an interface as
possible so that it can be properly separated from services. Admittedly,
getRepository() is not so much a service, but a property that may vary from
directory to directory, so it's probably right to make that a property of
Directory while ProjectRoot may still be a valuable abstraction.

At least as far as this diff is concerned, it does not make ProjectRoot a
public member of the API, so we haven't given up anything by introducing
the change in this form. That is, we could always take it out later without
disruption if it doesn't seem like a valuable abstraction.

bolinfest:
OK, I'm going to abandon/close this pull request in favor of the discussion on https://github.com/atom/atom/pull/5428.

PR #5428, opened 2015-02-06
Add hook for other implementations of [Git]Repository.

bolinfest:
With this change, `Project` registers as a consumer of `"atom.repository-provider"`.
When `atom.project.setPaths()` is called, if a path does not have a `GitRepository`
associated with it, it will see if any of the providers of `"atom.repository-provider"`
can produce an appropriate repository object.

I have a test package with a dummy implementation of `HgRepository` and the following
stanza in its `package.json`:

```
  "providedServices": {
    "atom.repository-provider": {
      "versions": {
        "0.0.1": "createHgRepositoryProvider"
      }
    }
  },
```

I verified that my provider was considered for a path with an Hg repo and that my
`HgRepository` was created and assigned to the `ProjectRoot` that was constructed.

### DIFF
--- a/src/project-root.coffee
+++ b/src/project-root.coffee
@@ -1,0 +1,20 @@
+# A root folder of {Project}.
+#
+module.exports =
+class ProjectRoot
+  constructor: (@directory, @repo) ->
+
+  # Public: Returns the {Directory} that corresponds to the project root.
+  getDirectory: -> @directory
+
+  # Public: Returns the {GitRepository} for the project root or {null}.
+  getRepository: -> @repo
+
+  destroy: ->
+    @directory.off()
+    @destroyRepo()
+
+  destroyRepo: ->
+    if @repo
+      @repo.destroy()
+      @repo = null
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -8,6 +8,7 @@ Q = require 'q'
 {Model} = require 'theorist'
 {Subscriber} = require 'emissary'
 {Emitter} = require 'event-kit'
+ProjectRoot = require './project-root'
 Serializable = require 'serializable'
 TextBuffer = require 'text-buffer'
 {Directory} = require 'pathwatcher'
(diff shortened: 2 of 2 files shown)


maxbrunsfeld:
So after our conversation today, we think we want to accomplish these goals with the following API changes:

```coffee-script
class Project

  # Will be removed in 2.0 because it does synchronous IO.
  # Only uses 'atom.repository-provider' services that implement ::createRepositorySync.
  getRepositories: ->
  
  # Returns a {Promise} that resolves with either:
  #   * {Repository} if a repository can be created for the given directory
  #   * `null` if no repository can be created for the given directory.
  #
  # Uses all 'atom.repository-provider' services.
  repositoryForDirectory: (directory) ->

# Registered at 'atom.repository-provider@0.1.0' service endpoint
class RepositoryProvider
  
  # Required.
  #
  # Returns a {Promise} that resolves with either:
  #   * {Repository} if the given directory has this type of repository.
  #   * `null` if the given directory does not have this type of repository.
  repositoryForDirectory: (directory) ->

  # Optional.
  #
  # Returns:
  #   * {Repository} if the given directory has this type of repository.
  #   * `null` if the given directory does not have this type of repository.
  createRepositorySync: (directory) ->

# See `GitRepository`
class Repository
```

On a related note, we'd like to also move toward allowing packages to provide their own `Directory` implementations via an API like this:

```coffee
# Registered at 'atom.directory-provider@0.1.0' service endpoint
class DirectoryProvider
  
  # Required.
  #
  #   * {Directory} if the given URI is compatible with this provider.
  #   * `null` if the given URI is not compatibile with this provider.
  directoryForURI: (uri) ->

# See pathwatcher's `Directory` class
class Directory
```

bolinfest:
OK, I implemented the `RepositoryProvider` component of this proposal in https://github.com/atom/atom/pull/5491.

bolinfest:
OK, this PR is superseded by:

https://github.com/atom/atom/pull/5491
https://github.com/atom/atom/pull/5535

Closing.

PR #5432, opened 2015-02-06
Refresh GitRepository status constantly

### LINKED ISSUE #5301: Find and Replace All in a project doesn't add the file changed icon in the sidebar until you view that file
If you use Find and Replace All to make changes in files in the project which aren't currently open, the sidebar won't show that the file has changed until you click to open and view the file.

**What should happen** - The sidebar should show that the file has changed, or open all those files in the editor with the changes. Sublime does this and leaves them unsaved, forcing the User to Save All.

as-cii:
Closes #5301 

/cc: @benogle

### DIFF
--- a/src/git-repository.coffee
+++ b/src/git-repository.coffee
@@ -79,7 +79,7 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
-    @watch = fs.watch path, { persistent: false }, (event, file) =>
+    @watch = fs.watch path, { persistent: false, recursive: true }, (event, file) =>
       @getPathStatus join(path, file)
 
     @statuses = {}
--- a/src/git-repository.coffee
+++ b/src/git-repository.coffee
@@ -79,7 +79,7 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
-    @watch = fs.watch path, { persistent: false }, (_, file) =>
+    @watch = fs.watch path, { persistent: false }, (event, file) =>
       @getPathStatus join(path, file)
 
     @statuses = {}
--- a/src/git-repository.coffee
+++ b/src/git-repository.coffee
@@ -79,8 +79,10 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
-    @watch = fs.watch path, { persistent: false, recursive: true }, (event, file) =>
-      @getPathStatus join(path, file)
+    @watch = setInterval (=>
+      @refreshIndex()
+      @refreshStatus()
+    ), 1000
 
     @statuses = {}
     @upstream = {ahead: 0, behind: 0}
@@ -103,7 +105,7 @@ class GitRepository
       @repo = null
 
     if @watch?
-      @watch.close()
+      clearInterval(@watch)
       @watch = null
 
     @subscriptions.dispose()
--- a/src/git-repository.coffee
+++ b/src/git-repository.coffee
@@ -79,24 +79,15 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
+    @watch = fs.watch path, { persistent: false }, (_, file) =>
+      @getPathStatus join(path, file)
+
     @statuses = {}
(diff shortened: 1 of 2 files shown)


as-cii:
It appears that, inside Atom, `fs.watch` works only on the top-level directory: in facts, changes in subdirectories won't raise any event, thus not updating the tree view (and other components which use `did-change-status` event). However, I performed some local tests (read: via a node script that has nothing to do with Atom) which highlighted that it *does* work on subdirectories as well (node version: v0.10.33).

I can simply recur over all the directories in the project (and it may just work), but there must be a reason why "recursive watching" doesn't work on Atom.

@benogle: recursion aside, what do you think about this change?

as-cii:
> I can simply recur over all the directories in the project (and it may just work), but there must be a reason why "recursive watching" doesn't work on Atom.

Locally it worked because, for some reason, recursion was enabled by default. Enabling it explicitly solves the issue :dancers: 

This approach involves another problem: changing git index (e.g. `git add something`) won't cause any **file change** event, thus showing a file as staged even when it's already committed; I guess that, by default, `fs.watch` doesn't look for changes in directories that begin with `.`

I suspect that before this PR the issue was still there, but it was solved through `window.onfocus` event. This is a bit coarse grained, therefore an alternative may be to explicitly look for changes in `.git` directory: when a file changes inside such directory, it means that git has been called and that we must invalidate the index. 

This would be particularly easy if we had [git_repository_discover](https://libgit2.github.com/libgit2/#HEAD/group/repository/git_repository_discover), which finds the folder automagically (modifying [https://github.com/atom/git-utils/blob/master/src/repository.cc](git-utils) to support the method should be relatively easy, I think). Otherwise we'd have to manually search for the directory inside Atom, but I believe it's way outside its boundaries.

benogle:
cc @kevinsawicki as you are the file watch master.

src/git-repository.coffee
```diff
@@ -79,6 +79,9 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
+    @watch = fs.watch path, { persistent: false, recursive: true }, (event, file) =>
```
kevinsawicki: According to the node docs:

> The fs.watch API is not 100% consistent across platforms, and is unavailable in some situations.
>
> The recursive option is currently supported on OS X. Only FSEvents supports this type of file watching so it is unlikely any additional platforms will be added soon.

So this won't be enough on Linux and Windows and so the focus handler should probably be kept.

as-cii: For some reason I totally skipped that part which mentioned the support for OS X only. That's unfortunate, but wouldn't be enough to build a directory tree and watch each directory? I don't think performance would degrade that much, but I haven't any benchmark to show so it's just a supposition :neutral_face: 

Alternatively there are other libraries which claim to solve the issue already: https://github.com/paulmillr/chokidar is one of them, and it has support for Windows, Linux and OS X (it seems to be well supported, but I haven't tried it yet).

@kevinsawicki: what do you think?

as-cii: I tried out chokidar (and alternatives, like watchr) but they won't solve the issue when there are directories with a large amount of files (e.g. `atom/atom` repository).

There seems to be a pull request for Windows support (https://github.com/joyent/libuv/pull/1473) but it has to be merged yet (and Linux would still be a problem; even the new [fanotify](http://man7.org/linux/man-pages/man7/fanotify.7.html) doesn't work out of the box). This is quite sad and I assumed this problem was something which operating systems had already solved long time ago.

Constantly polling the git repository could be a solution at this point: we may setup a timer which runs, for instance, every second and fetches updates from the repository. Other than removing the need of subscribing to window focus and to buffer change, it would still solve #5301.

@kevinsawicki: my last commit is the exact implementation of what I said above. It works quite well on my machine and it doesn't eat a lot of resources.

Meanwhile, I'll change the name of the PR as well, given that we've discarded the `fs.watch` option.

bolinfest:
Have you tested this on large Git repos?

as-cii:
@bolinfest I have tested it on Atom, but the more I think about this, the less I am convinced this is a suitable solution.

I'd like to close this for now, until we come up with something better. Thanks everyone for the great feedback! :heart:

PR #5436, opened 2015-02-06
External config documentation

mattberkowitz:
Per advice of @thomasjo in https://github.com/atom/atom/pull/5429#issuecomment-73186684, added the following to config.coffee documentation to help other users:
#### External Config Schemas

If you would like to store your schema in an external file and require it, you must require it prior to setting the `config` value in `module.exports`:

```
ConfigSchema = require('./configuration.coffee')
module.exports =
  config: ConfigSchema
```

not:

```
module.exports =
  config: require('./configuration.coffee')
```


src/config.coffee
```diff
@@ -115,6 +115,23 @@ ScopeDescriptor = require './scope-descriptor'
 # atom.config.get('my-package.thingVolume') # -> 11
 # ```
 #
+# #### External Config Schemas
+#
+# If you would like to store your schema in an external file and require it, you must require it prior to setting the `config` value in `module.exports`:
+#
+# ```
+# ConfigSchema = requrie('./configuration.coffee')
```
mnquintana: Looks like there's a typo here and in the other `require`

mattberkowitz:
Thanks @mnquintana, I've fixed that

PR #5445, opened 2015-02-07
Allow users to enable or disable the automatic changing of indent

h0dgep0dge:
This is less of a serious pull request, and more a prompt for discussion. I've written a proof of concept that disables automatic changing of indent based on a setting on the preferences screen.


h0dgep0dge:
For some reason, this only does anything when I'm running it straight from the build directory. I'm not familiar enough with the architecture of Atom to work this out at the moment, but maybe I'll work it out before someone else does. In addition, use of the options object breaks certain packages. This is likely a quite poor implementation, but one could consider this thread a feature request of sorts.

src/selection.coffee
```diff
@@ -387,7 +387,7 @@ class Selection extends Model
           @editor.setIndentationForBufferRow(row, newIndent)
     else if options.autoIndentNewline and text == '\n'
```
nathansobo: If you could avoid this section running entirely, wouldn't that mean you didn't need to change `language-mode.coffee` at all?

nathansobo:
Seems like a more specific config option name like "autoIndentNewlines" would be more descriptive.

Also, we'll definitely need tests for any new behavior in `text-editor-spec.coffee`.

h0dgep0dge:
That option already exists, and isn't what I want. The current implementation seems to be an all or nothing kind of deal; you either take the auto-indentation with the annoyance of having the editor up and change the indent level on you, or you take nothing, and it seems to me that these 2 features can be split into two different options.

And like I said above, this is a likely a poor implementation, it's only really meant to be a proof of concept and hack to make the editor usable for me.

AntonKhorev:
This is a *simple autoindent* vs *smart/full autoindent* option found in [some editors](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/296#issuecomment-116782726).

h0dgep0dge:
I'd like to say, I know very little about desktop javascript development, coffee, or Atom specifically. This implementation isn't meant to be an actual proposed merge, but a proof of concept and feature request.

damieng:
Contains merge conflicts and debug code, considered abandoned.

Issue #1612, opened 2014-02-28
Allow the gutter to show relative line numbers


probablycorey:
https://github.com/atom/vim-mode/issues/54#issuecomment-36278523

jamezb:
+1

vito-c:
+1

qianlifeng:
:+1:

murphyrandle:
+1

meznaric:
+1

zpconn:
:+1: :+1:

I use this all the time in vim and really miss it in Atom.

erickbarron:
:+1:

edazdarevic:
+1

isair:
+1

kai3x5:
Anyone working on this ? Is there an existing solution already ?

lee-dohm:
Looks like there are a couple different packages to do this:

* [relative-line-numbers](https://atom.io/packages/relative-line-numbers)
* [relative-numbers](https://atom.io/packages/relative-numbers)

I haven't tried them, so I can't endorse them. But there you go :grinning:

erickbarron:
relative-line-numbers: I have been using it for a while, works great. I installed it after relative-numbers didn't work. One problem it had before was that it was buggy if you were using soft-wrap. But it looks like he fixed it as I made opened an issue. 

relative-numbers: Have tried it before, it doesn't really work so hot. It crashes (for me at least)

kai3x5:
Huzzah! relative-line-numbers works! 

relative-numbers is broken for me.

spirityy:
relative-line-numbers is broken too?
Is there an solution ?

gggritso:
There was a [big pull request](https://github.com/atom/atom/issues/5038) a while back to add an API for custom gutters, but at the time I couldn't find documentation for it. I assume this is the "right" way to implement something like this.

yuriteixeira:
The current `relative-line-numbers` package (https://atom.io/packages/relative-line-numbers) isn't available anymore (404 on the repo's page). I'm using https://github.com/dimatter/atom-relative-line-numbers (although manually downloaded into my `~/.atom/packages` folder) and it is working very well, as you can see in the screenshot below: 

![screenshot 2015-08-28 14 49 29](https://cloud.githubusercontent.com/assets/461055/9546571/065f0a22-4d94-11e5-989f-cb54b21b1ec8.png)

maxbrunsfeld:
[...]

PR #5449, opened 2015-02-08
Allow the gutter to show relative line numbers

Nadrieril:
Displays vim-like relative line numbers in gutter, with option to show true current line number instead of zero on current line
https://github.com/atom/atom/issues/1612

### DIFF
--- a/src/gutter-component.coffee
+++ b/src/gutter-component.coffee
@@ -4,6 +4,7 @@ React = require 'react-atom-fork'
 {isEqual, isEqualForProperties, multiplyString, toArray} = require 'underscore-plus'
 Decoration = require './decoration'
 SubscriberMixin = require './subscriber-mixin'
+{$} = require './space-pen-extensions'
 
 WrapperDiv = document.createElement('div')
 
@@ -84,7 +85,7 @@ GutterComponent = React.createClass
   # since the real line numbers are absolutely positioned for performance reasons.
   appendDummyLineNumber: ->
     {maxLineNumberDigits} = @props
-    WrapperDiv.innerHTML = @buildLineNumberHTML(-1, false, maxLineNumberDigits)
+    WrapperDiv.innerHTML = @buildLineNumberHTML(-1, null, 0, false, maxLineNumberDigits)
     @dummyLineNumberNode = WrapperDiv.children[0]
     @refs.lineNumbers.getDOMNode().appendChild(@dummyLineNumberNode)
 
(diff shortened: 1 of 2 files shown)


50Wliu:
You'll probably want to add tests for the new behavior in order for this to get merged.

Nadrieril:
I'd like to do that indeed. Could you give me some pointers for how to run atom tests ? I can find it anywhere in the docs

src/gutter-component.coffee
```diff
@@ -3,6 +3,7 @@ React = require 'react-atom-fork'
 {div} = require 'reactionary-atom-fork'
 {isEqual, isEqualForProperties, multiplyString, toArray} = require 'underscore-plus'
 Decoration = require './decoration'
+{$} = require './space-pen-extensions'
```
thedaniel: We're trying to remove jQuery dependencies wherever we can, this probably won't get merged with this require.

Nadrieril: It was indeed not necessary, I removed it

50Wliu:
@Nadrieril try [this document](https://github.com/atom/atom/blob/master/docs/writing-specs.md) (scroll down near the bottom to see how to run them).  Basically-`window:run-package-specs`.

nathansobo:
@Nadrieril hey, so you should really be performing the logic for this in the `TextEditorPresenter` layer rather than in the view. Ideally it's implemented and tested mostly at the presenter layer and the view just has minor tweaks to consume it. We should probably replace "buffer row" with "display row" like you're doing now in the line number state objects in the presenter.

Nadrieril:
I did not see so much changes coming.
I don't think we can replace entirely the bufferRow, at least because I need it to calculate the relative row number, and it might be needed in the view anyway.
Where should I put the displayRow recalculation ? Ideally, with normal line numbers, recalculation should only be triggered when adding or removing lines in the editor, not on scroll nor on cursor move. So this update cannot be in updateLinesState, but it cannot be in updateLineNumbersState either, because updateLineNumbersState should only happen on content change...

nathansobo:
@Nadrieril You'll probably need a pretty extensive change to the presenter when in relative line number mode that updates line numbers when the cursor moves instead of based on content.

Nadrieril:
@nathansobo Yeah, I guessed... That's why I haven't I haven't progressed a lot since

Francescu:
Any updates? This feature would be amazing.

Nadrieril:
Atom has changed too much since I made this feature. I would have to code it from scratch, but I don't have time. If anyone is interested, feel free to look at my code

PR #5487, opened 2015-02-11
Upgrade and dedupe q

kevinsawicki:
In dev mode there is no module caching and so a bunch of versions of [q](https://github.com/kriskowal/q) are getting required since packages and modules were using a variety of versions.

Requiring q in dev mode is pretty slow since it create and inspects an error which causes source maps to load and a CoffeeScript compile to occur. Requiring it multiple times makes that even worse.

This PR speeds up dev mode by ensuring q is required only once.  It also allows 1 version of q to ship in the actual app bundle instead of 5.

This also upgrades Atom to the latest q version, 1.1.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "optimist": "0.4.0",
     "pathwatcher": "^3.1.0",
     "property-accessors": "^1",
-    "q": "^1.0.1",
+    "q": "^1.1.2",
     "random-words": "0.0.1",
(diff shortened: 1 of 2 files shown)


YuriSolovyov:
just curious: why do you use q over bluebird or native promises?

kevinsawicki:
> why do you use q over bluebird or native promises?

Native promises are a relatively recent addition to Atom Shell and Atom code has not been migrated yet.

PR #5491, opened 2015-02-11
Set up the atom.repository-provider service and implement GitRepositoryP...

bolinfest:
...rovider.

I tested this by creating a dummy implementation of an `HgRepositoryProvider`
(with the optional `createRepositorySync()` method implemented) and an `HgRepository`
in an Atom package with the following stanza in the `package.json`:

```
  "providedServices": {
    "atom.repository-provider": {
      "versions": {
        "0.1.0": "createHgRepositoryProvider"
      }
    }
  },
```

I opened a path with an Hg repository from the command line using Atom.
I verified that `atom.project.repositoryProviders` contains both a
`GitRepositoryProvider` and an `HgRepositoryProvider`.

I also verified that running the following printed out an `HgRepository`:

```
var Directory = require('pathwatcher').Directory;
atom.project.repositoryForDirectory(
    new Directory(atom.project.getPath(), /* symlink */ false)).then(
        function(repo) { console.log('repo: %o', repo); });
```

One thing that stands out to me about the current API for the
atom.repository-provider service is that the function used to create
a `RepositoryProvider` does not receive any arguments. If the creation
of the `GitRepositoryProvider` were done via the service, this would
be a problem because it needs a reference to `atom.project`, which is
not defined when it is created. (We work around this because it is
created in `Project`'s constructor, so it can pass `this` to
`new GitRepositoryProvider()`.)

We would have to create a `RepositoryProviderFactory` or something if
we wanted to specify arguments when creating a `RepositoryProvider`,
in general. Maybe that's too crazy / not an issue, in practice.

Though note that `GitRepository` cannot access `atom.project` lazily
because it uses it in its constructor to do the following:

```
if @project?
  @subscriptions.add @project.eachBuffer (buffer) => @subscribeToBuffer(buffer)
```

So long as we can guarantee that `atom.project` is defined before the
other providers are initialized, I think we should be OK.

Follow-up work:
- Replace the use of `RepositoryProvider.createRepositorySync()` with
  `RepositoryProvider.repositoryForDirectory()` in `Project.setPaths()`.
- Replace all uses of `Project.getRepositories()` with
  `Project.repositoryForDirectory()` in packages that are bundled with Atom
  by default.
- Implement `Directory.exists()` and/or `Directory.existsSync()` and update
  `git-repositor-provider.coffee`, as appropriate.
- Eliminate `GitRepositoryProvider.repositoryForDirectory()`'s use of
  synchronous methods.
- Somewhat orthogonal to this diff, but the following fields need to be
  removed from `Project` because they enforce the idea of a single root:
  `path`, `rootDirectory`, and `repo`. This has implications around the
  existing use of `@rootDirectory?.off()` and `@destroyRepo()`.

### DIFF
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -8,17 +8,13 @@ describe "GitRepositoryProvider", ->
 
     describe "when specified a Directory with a Git repository", ->
       it "returns a Promise that resolves to a GitRepository", ->
-        provider = new GitRepositoryProvider atom.project
-        theResult = null
-
         waitsForPromise ->
+          provider = new GitRepositoryProvider atom.project
           directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
-          provider.repositoryForDirectory(directory).then (result) -> theResult = result
-
-        runs ->
-          expect(theResult).toBeInstanceOf GitRepository
-          expect(provider.pathToRepository[theResult.getPath()]).toBeTruthy()
-          expect(theResult.statusTask).toBeTruthy()
+          provider.repositoryForDirectory(directory).then (result) ->
+            expect(result).toBeInstanceOf GitRepository
+            expect(provider.pathToRepository[result.getPath()]).toBeTruthy()
+            expect(result.statusTask).toBeTruthy()
 
       it "returns the same GitRepository for different Directory objects in the same repo", ->
         provider = new GitRepositoryProvider atom.project
@@ -38,13 +34,9 @@ describe "GitRepositoryProvider", ->
           expect(firstRepo).toBe secondRepo
 
     describe "when specified a Directory without a Git repository", ->
-      provider = new GitRepositoryProvider atom.project
-      theResult = 'dummy_value'
-
       it "returns a Promise that resolves to null", ->
         waitsForPromise ->
+          provider = new GitRepositoryProvider atom.project
           directory = new Directory '/tmp'
-          provider.repositoryForDirectory(directory).then (result) -> theResult = result
-
-      runs ->
-        expect(theResult).toBe null
+          provider.repositoryForDirectory(directory).then (result) ->
+            expect(result).toBe null
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -1,0 +1,42 @@
+path = require 'path'
+{Directory} = require 'pathwatcher'
+GitRepository = require '../src/git-repository'
+GitRepositoryProvider = require '../src/git-repository-provider'
+
+describe "GitRepositoryProvider", ->
+  describe ".repositoryForDirectory(directory)", ->
+
+    describe "when specified a Directory with a Git repository", ->
+      provider = new GitRepositoryProvider atom.project
+      the_result = 'dummy_value'
+      the_second_result = 'dummy_value2'
+
+      it "returns a Promise that resolves to a GitRepository", ->
+        waitsForPromise ->
+          directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
+          provider.repositoryForDirectory(directory).then (result) -> the_result = result
+
+        runs ->
+          expect(the_result).toBeInstanceOf GitRepository
+          expect(provider.pathToRepository[the_result.getPath()]).toBeTruthy()
+          expect(the_result.statusTask).toBeTruthy()
+
+        waitsForPromise ->
+          directory = new Directory path.join(__dirname, 'fixtures/git/master.git/objects')
+          provider.repositoryForDirectory(directory).then (result) -> the_second_result = result
+
+        runs ->
+          expect(the_second_result).toBeInstanceOf GitRepository
+          expect(the_second_result).toBe the_result
+
+    describe "when specified a Directory without a Git repository", ->
+      provider = new GitRepositoryProvider atom.project
+      the_result = 'dummy_value'
+
+      it "returns a Promise that resolves to null", ->
+        waitsForPromise ->
+          directory = new Directory '/tmp'
+          provider.repositoryForDirectory(directory).then (result) -> the_result = result
+
+      runs ->
+        expect(the_result).toBe null
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -8,35 +8,35 @@ describe "GitRepositoryProvider", ->
 
     describe "when specified a Directory with a Git repository", ->
       provider = new GitRepositoryProvider atom.project
-      the_result = 'dummy_value'
-      the_second_result = 'dummy_value2'
+      theResult = 'dummy_value'
+      theSecondResult = 'dummy_value2'
 
       it "returns a Promise that resolves to a GitRepository", ->
         waitsForPromise ->
           directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
-          provider.repositoryForDirectory(directory).then (result) -> the_result = result
+          provider.repositoryForDirectory(directory).then (result) -> theResult = result
 
         runs ->
-          expect(the_result).toBeInstanceOf GitRepository
-          expect(provider.pathToRepository[the_result.getPath()]).toBeTruthy()
-          expect(the_result.statusTask).toBeTruthy()
+          expect(theResult).toBeInstanceOf GitRepository
+          expect(provider.pathToRepository[theResult.getPath()]).toBeTruthy()
+          expect(theResult.statusTask).toBeTruthy()
 
         waitsForPromise ->
           directory = new Directory path.join(__dirname, 'fixtures/git/master.git/objects')
-          provider.repositoryForDirectory(directory).then (result) -> the_second_result = result
+          provider.repositoryForDirectory(directory).then (result) -> theSecondResult = result
 
         runs ->
-          expect(the_second_result).toBeInstanceOf GitRepository
-          expect(the_second_result).toBe the_result
+          expect(theSecondResult).toBeInstanceOf GitRepository
+          expect(theSecondResult).toBe theResult
 
     describe "when specified a Directory without a Git repository", ->
       provider = new GitRepositoryProvider atom.project
-      the_result = 'dummy_value'
+      theResult = 'dummy_value'
 
       it "returns a Promise that resolves to null", ->
         waitsForPromise ->
           directory = new Directory '/tmp'
-          provider.repositoryForDirectory(directory).then (result) -> the_result = result
+          provider.repositoryForDirectory(directory).then (result) -> theResult = result
 
       runs ->
-        expect(the_result).toBe null
+        expect(theResult).toBe null
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -7,11 +7,10 @@ describe "GitRepositoryProvider", ->
   describe ".repositoryForDirectory(directory)", ->
 
     describe "when specified a Directory with a Git repository", ->
-      provider = new GitRepositoryProvider atom.project
-      theResult = 'dummy_value'
-      theSecondResult = 'dummy_value2'
-
       it "returns a Promise that resolves to a GitRepository", ->
+        provider = new GitRepositoryProvider atom.project
+        theResult = null
+
         waitsForPromise ->
           directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
           provider.repositoryForDirectory(directory).then (result) -> theResult = result
@@ -21,13 +20,22 @@ describe "GitRepositoryProvider", ->
           expect(provider.pathToRepository[theResult.getPath()]).toBeTruthy()
           expect(theResult.statusTask).toBeTruthy()
 
+      it "returns the same GitRepository for different Directory objects in the same repo", ->
+        provider = new GitRepositoryProvider atom.project
+        firstRepo = null
+        secondRepo = null
+
+        waitsForPromise ->
+          directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
+          provider.repositoryForDirectory(directory).then (result) -> firstRepo = result
+
         waitsForPromise ->
           directory = new Directory path.join(__dirname, 'fixtures/git/master.git/objects')
-          provider.repositoryForDirectory(directory).then (result) -> theSecondResult = result
+          provider.repositoryForDirectory(directory).then (result) -> secondRepo = result
 
         runs ->
-          expect(theSecondResult).toBeInstanceOf GitRepository
-          expect(theSecondResult).toBe theResult
+          expect(firstRepo).toBeInstanceOf GitRepository
+          expect(firstRepo).toBe secondRepo
 
     describe "when specified a Directory without a Git repository", ->
       provider = new GitRepositoryProvider atom.project
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -214,19 +214,18 @@ describe "Project", ->
 
     it "resolves to a GitRepository and is cached when the given directory is a Git repo", ->
       the_result = "dummy_value"
-      the_promise = null
       directory = new Directory(path.join(__dirname, '..'))
 
       waitsForPromise ->
-        the_promise = atom.project.repositoryForDirectory(directory)
-        the_promise.then (result) -> the_result = result
+        atom.project.repositoryForDirectory(directory).then (result) -> the_result = result
 
       runs ->
         dirPath = directory.getRealPathSync()
         expect(the_result).toBeInstanceOf GitRepository
         expect(the_result.getPath()).toBe path.join(dirPath, '.git')
-        expect(atom.project.repositoryPromisesByPath.size).toBe 1
-        expect(atom.project.repositoryPromisesByPath.get dirPath).toBe the_promise
+
+        # Verify that the result is cached.
+        expect(atom.project.repositoryForDirectory(directory)).toBe(atom.project.repositoryForDirectory(directory))
 
   describe ".setPaths(path)", ->
     describe "when path is a file", ->
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -205,7 +205,7 @@ describe "Project", ->
     # be different than the atom.project used in a subsequent it(). To work
     # around this issue, we use runs() instead of it().
 
-    describe "there is no repository for /tmp even though there are RepositoryProviders", ->
+    it "there is no repository for /tmp even though there are RepositoryProviders", ->
       the_result = "dummy_value"
 
       waitsForPromise ->
@@ -218,7 +218,7 @@ describe "Project", ->
         expect(the_result).toBe null
         expect(atom.project.repositoryPromisesByPath.size).toBe 0
 
-    describe "when Git repository for directory, promise resolves to Atom's GitRepository and is cached", ->
+    it "when Git repository for directory, promise resolves to Atom's GitRepository and is cached", ->
       the_result = "dummy_value"
       the_promise = null
       directory = new Directory(path.join(__dirname, '..'))
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -5,6 +5,8 @@ _ = require 'underscore-plus'
 fs = require 'fs-plus'
 path = require 'path'
 BufferedProcess = require '../src/buffered-process'
+{Directory} = require 'pathwatcher'
+GitRepository = require '../src/git-repository'
 
 describe "Project", ->
   beforeEach ->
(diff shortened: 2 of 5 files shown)


src/project.coffee
```diff
@@ -96,11 +110,45 @@ class Project extends Model
 
   # Public: Get an {Array} of {GitRepository}s associated with the project's
   # directories.
+  #
+  # This method will be removed in 2.0 because it does synchronous I/O.
+  # Prefer the following, which evaluates to a Promise that resolves to an
+  # {Array} of {Repository} objects:
+  # ```
+  # project.getDirectories().map(
```
nathansobo: Doesn't this evaluate to an array of promises for repository objects?

src/project.coffee
```diff
@@ -39,6 +39,20 @@ class Project extends Model
     @emitter = new Emitter
     @buffers ?= []
 
+    # Mapping from the real path of a {Directory} to a {Promise} that resolves
+    # to either a {Repository} or null. Ideally, the {Directory} would be used
+    # as the key; however, there can be multiple {Directory} objects created for
+    # the same real path, so it is not a good key.
```
nathansobo: It would be great to maintain an identity map of pathname to `Directory` and `File` instances on `Project`.

src/project.coffee
```diff
@@ -39,6 +39,20 @@ class Project extends Model
...
+    @directoryToRepositoryMap = new Map();
+
+    # Note that the GitRepositoryProvider is registered synchronously so that
+    # it is available immediately on startup.
+    @repositoryProviders = [new GitRepositoryProvider(this)]
+    atom.packages.serviceHub.consume(
+        'atom.repository-provider',
+        '>=0.1.0',
```
nathansobo: We should probably constrain this version range more tightly, like `^0.1.0` so only patch releases are compatible.

src/project.coffee
```diff
@@ -96,11 +110,45 @@ class Project extends Model
...
+  # ```
+  # project.getDirectories().map(
+  #     project.repositoryForDirectory.bind(project))
+  # ```
   getRepositories: -> _.compact([@repo])
   getRepo: ->
     Grim.deprecate("Use ::getRepositories instead")
     @repo
 
+  # Public: Returns a {Promise} that resolves with either:
```
nathansobo: Our typical format is to put the return value at the end. So something like:

```coffee
# Public: Get the repository for a given directory asynchronously.
#
# * `directory` {Directory} for which to get a {Repository}.
#
# Returns a {Promise} that ...
```

src/project.coffee
nathansobo: Also, what do you think of the name `repositoryPromisesByDirectory`? That blends in with naming we use elsewhere for maps.

src/project.coffee
```diff
@@ -96,11 +110,45 @@ class Project extends Model
...
+  # * {Repository} if a repository can be created for the given directory
+  # * `null` if no repository can be created for the given directory.
+  #
+  # * `directory` {Directory} for which to get a {Repository}.
+  repositoryForDirectory: (directory) ->
+    path = directory.getRealPathSync()
+    promise = @directoryToRepositoryMap.get(path)
+    unless promise
+      promises = @repositoryProviders.map(
```
nathansobo: Nitpick here, but I'd prefer to see this formatted as follows:

```coffee
promises = @repositoryProviders.map (provider) ->
  provider.repositoryForDirectory(directory)
```

src/project.coffee
```diff
@@ -96,11 +110,45 @@ class Project extends Model
...
+          (provider) -> provider.repositoryForDirectory directory)
+      promise = Promise.all(promises).then((repositories) =>
```
nathansobo: I'm pretty sure the following formatting would work, which I think could be more clear:

```coffee
promise = Promise.all(promises).then
  (repositories) =>
    # now the function is indented and doesn't need the parens, which looks weird on line 148 with `repo)`
```

nathansobo: Also, should we be handling the rejection case in the event an exception is thrown by a provider?

src/project.coffee
```diff
@@ -126,9 +174,13 @@ class Project extends Model
     if projectPath?
       directory = if fs.isDirectorySync(projectPath) then projectPath else path.dirname(projectPath)
       @rootDirectory = new Directory(directory)
-      if @repo = GitRepository.open(directory, project: this)
-        @repo.refreshIndex()
-        @repo.refreshStatus()
+
+      # For now, use only the repositoryProviders with a sync API.
+      for provider in @repositoryProviders
+        if provider.createRepositorySync
```
nathansobo: This could be collapsed to:

```coffee
break if @repo = provider.createRepositorySync?(@rootDirectory)
```

As a bonus, calling methods that might not exist with `?` is slightly more correct in CoffeeScript because it actually verifies that the property is a function.

src/git-repository.coffee
```diff
@@ -111,7 +116,14 @@ class GitRepository
...
       @repo = null
 
-    @subscriptions.dispose()
+    if @subscriptions?
+      @subscriptions.dispose()
+      @subscriptions = null
+
+  # Public: Invoke the given callback when this GitRepository's destroy() method
+  # is invoked.
+  onDestroy: (callback) ->
```
nathansobo: We always include an auxiliary verb in event subscription methods for consistency, `onDidDestroy`. Maybe this was overkill and should have only been applied to events that had a `onWill` and `onDid` case, but the :boat: has sailed.

src/git-repository.coffee
```diff
@@ -111,7 +116,14 @@ class GitRepository
       @repo.release()
       @repo = null
 
-    @subscriptions.dispose()
+    if @subscriptions?
```
nathansobo:
```coffee
@subscriptions?.dispose()
```

src/git-repository.coffee
```diff
@@ -101,8 +101,13 @@ class GitRepository
   # Public: Destroy this {GitRepository} object.
   #
   # This destroys any tasks and subscriptions and releases the underlying
-  # libgit2 repository handle.
+  # libgit2 repository handle. This method is idempotent.
   destroy: ->
+    if @emitter?
+      @emitter.emit 'did-destroy'
+      @emitter.off()
```
nathansobo: Pretty sure this should be `@emitter.dispose()`.

src/git-repository-provider.coffee
```diff
@@ -0,0 +1,80 @@
+fs = require 'fs'
+GitRepository = require './git-repository'
+
+# Checks whether the specified directory has, or has a parent directory that
```
nathansobo: Could we invert the phrasing for clarity: "Checks whether a valid `.git` directory is contained within the given directory or one of its ancestors." Just a thought. Not a huge deal.

src/git-repository-provider.coffee
```diff
@@ -0,0 +1,80 @@
...
+# has, a valid .git folder, indicating the root of a Git repository.
+# If found, a Directory that corresponds to the .git folder will be returned.
+# Otherwise, returns `null`.
+#
+# * `directory` {Directory} to explore whether it is part of a Git repository.
+findGitDirectorySync = (directory) ->
+  # TODO: Fix node-pathwatcher/src/directory.coffee so the following methods
```
nathansobo: Ah, you're already ahead of me on the identity map idea.

src/git-repository-provider.coffee
```diff
@@ -0,0 +1,80 @@
...
+  return directoryExistsSync(directory.getSubdirectory('objects')) and
+      directory.getFile('HEAD').exists() and
+      directoryExistsSync(directory.getSubdirectory('refs'))
+
+# Returns a boolean indicating whether the specified directory exists.
+#
+# * `directory` {Directory} to check for existence.
+directoryExistsSync = (directory) ->
+  # TODO: Directory should have its own existsSync() method. Currently, File has
+  # an exists() method, which is synchronous, so it may be tricky to achieve
```
nathansobo: Damn. I'm pissed that we missed that! I think we should probably bit the bullet deprecate it post-freeze and favor existsSync to set ourselves up for an async API past 1.0. @kevinsawicki what would you think about that?

src/git-repository-provider.coffee
```diff
@@ -0,0 +1,80 @@
...
+  # * `null` if the given directory does not have a Git repository.
+  repositoryForDirectory: (directory) ->
+    # TODO: Currently, this method is designed to be async, but it relies on a
+    # synchronous API. It should be rewritten to be truly async.
+    Promise.resolve(@createRepositorySync(directory))
+
+  # Returns either:
+  # * {GitRepository} if the given directory has a Git repository.
+  # * `null` if the given directory does not have a Git repository.
+  createRepositorySync: (directory) ->
```
nathansobo: Should this be called `repositoryForDirectorySync`?

nathansobo:
Great work @bolinfest! @maxbrunsfeld can I pass the :flashlight: to you for merging on this since you're actively working in this area of the code?

src/project.coffee
bolinfest: Good catch: I'll change the code sample rather than the comment.

src/project.coffee
bolinfest: I'm not sure if `Project` should be maintaining that map or if a `DirectoryProvider` should be responsible for its own cache.

Renamed to `repositoryPromisesByDirectory`, as requested. One thing to note is that I made this a `Map` rather than a `{}` so that there would be fewer code changes if the key type changed (hence having `Map` in the name). However, I'm not sure if we'll every be confident that there will be exactly one `Directory` for path, in which case the key will probably always be a string and I can just change this to a `{}` now.

src/project.coffee
bolinfest: Done.

src/project.coffee
bolinfest: Good to know: fixed.

src/project.coffee
maxbrunsfeld: Aren't the keys in a `Map` compared via pointer equality? If instantiate two identical `Directory` objects, and store the `Repository` for one of them in this map, will I be able to read the repository out using the other?

src/project.coffee
bolinfest: np, done.

src/project.coffee
nathansobo: > I'm not sure if Project should be maintaining that map or if a DirectoryProvider should be responsible for its own cache.

You're right. It should definitely be the provider. But I'm comfortable assuming the provider does this correctly and making that part of the contract.

I think the name should clarify the key type. So if we're changing this to be indexed by path, then it should probably be `repositoryPromisesByPath` so as not to mislead. It would be great to actually make it a `Map` and fix the aliasing problems in pathwatcher :soon:.

src/project.coffee
bolinfest: I actually have to make heavy use of the Preview package to figure out what is legit and what's not.

It seems like I can drop the parens but have to keep `(repositories)` on the same line as `then`.

I suppose we could log it? A provider should really be defensive and return `null`. If we don't add any handling, unhandled Promises do appear in the console, which is awesome, btw.

src/project.coffee
bolinfest: CoffeeScript requires too much cleverness for me. Fixed.

src/git-repository.coffee
bolinfest: That's fine: works for me.

src/git-repository.coffee
bolinfest: Are you sure on this one? That would make it inconsistent with the other things in `destroy()`.

src/git-repository.coffee
bolinfest: Ah, it does appear to be from `event-kit`:

    {Emitter, Disposable, CompositeDisposable} = require 'event-kit'

Fixed.

src/git-repository-provider.coffee
bolinfest: Done.

src/git-repository.coffee
nathansobo: I can see an argument for leaving it. No big deal.

src/git-repository-provider.coffee
bolinfest: Yep! Again, I think each `DirectoryProvider` may have to take responsibility for this.

src/git-repository-provider.coffee
bolinfest: :thumbsup:

src/git-repository-provider.coffee
bolinfest: Makes sense: I was just going off of https://github.com/atom/atom/pull/5428. Will change.

src/project.coffee
lee-dohm: @bolinfest You can find the documentation on the AtomDoc format here:

https://github.com/atom/atomdoc

I'm probably going to be enhancing the documentation of it over the next couple weeks.

maxbrunsfeld:
Yeah, nice work @bolinfest. Can you write some specs for this functionality? Mainly, I'd like to see tests for `Project::repositoryForDirectory`. I'm happy to help w/ this if you have questions.

bolinfest:
@maxbrunsfeld Let me know whether you feel this needs more test coverage.

spec/project-spec.coffee
```diff
@@ -197,6 +199,43 @@ describe "Project", ->
           atom.project.bufferForPath("b").then (anotherBuffer) ->
             expect(anotherBuffer).not.toBe buffer
 
+  describe ".repositoryForDirectory(directory)", ->
+    # it() causes beforeEach() in spec-helper.coffee to be run, which reassigns
+    # atom.project. Therefore, the atom.project used in waitsForPromise() will
+    # be different than the atom.project used in a subsequent it(). To work
+    # around this issue, we use runs() instead of it().
```
maxbrunsfeld: :fire: comment

bolinfest: thanks: removed

spec/project-spec.coffee
```diff
@@ -197,6 +199,43 @@ describe "Project", ->
...
+
+    it "there is no repository for /tmp even though there are RepositoryProviders", ->
```
maxbrunsfeld: We try to phrase these so that they read nicely starting with an `it`. For this one, I'd say something like `it "resolves to null for the /tmp directory"`.

spec/project-spec.coffee
```diff
@@ -197,6 +199,43 @@ describe "Project", ->
...
+        directory = new Directory("/tmp")
+        atom.project.repositoryForDirectory(directory).then (result) ->
+          the_result = result
+
+      runs ->
+        expect(atom.project.repositoryProviders.length).toBeGreaterThan 0
+        expect(the_result).toBe null
+        expect(atom.project.repositoryPromisesByPath.size).toBe 0
+
+    it "when Git repository for directory, promise resolves to Atom's GitRepository and is cached", ->
```
maxbrunsfeld: Maybe something like `it "resolves to a GitRepository when the given directory is a git repo"`

spec/project-spec.coffee
```diff
@@ -197,6 +199,43 @@ describe "Project", ->
...
             expect(anotherBuffer).not.toBe buffer
 
+  describe ".repositoryForDirectory(directory)", ->
+    # it() causes beforeEach() in spec-helper.coffee to be run, which reassigns
+    # atom.project. Therefore, the atom.project used in waitsForPromise() will
+    # be different than the atom.project used in a subsequent it(). To work
+    # around this issue, we use runs() instead of it().
+
+    it "there is no repository for /tmp even though there are RepositoryProviders", ->
+      the_result = "dummy_value"
```
maxbrunsfeld: We usually just use `null` for this purpose.

bolinfest: But the assertion is:

    expect(the_result).toBe null

So that would succeed if the promise did not resolve.

spec/project-spec.coffee
```diff
@@ -197,6 +199,43 @@ describe "Project", ->
...
+        the_promise.then (result) ->
+          the_result = result
+
+      runs ->
+        dirPath = directory.getRealPathSync()
+        expect(the_result).toBeInstanceOf GitRepository
+        expect(the_result.getPath()).toBe path.join(dirPath, '.git')
+
+        expect(atom.project.repositoryPromisesByPath.size).toBe 1
+        expect(atom.project.repositoryPromisesByPath.get dirPath).toBe the_promise
```
maxbrunsfeld: I think that rather than using the private property `repositoryPromisesByPath`, you should verify the caching of the promises by saying:

```coffee
expect(atom.project.repositoryForDirectory(directory)).toBe(atom.project.repositoryForDirectory(directory))
```

spec/project-spec.coffee
```diff
@@ -197,6 +199,35 @@ describe "Project", ->
...
+        expect(the_result).toBe null
+        expect(atom.project.repositoryPromisesByPath.size).toBe 0
+
+    it "resolves to a GitRepository and is cached when the given directory is a Git repo", ->
+      the_result = "dummy_value"
+      the_promise = null
+      directory = new Directory(path.join(__dirname, '..'))
+
+      waitsForPromise ->
+        the_promise = atom.project.repositoryForDirectory(directory)
```
benogle: We use camel case for variable names

spec/project-spec.coffee
bolinfest: Done.

spec/project-spec.coffee
bolinfest: Changed to:

    "resolves to a GitRepository and is cached when the given directory is a Git repo"

spec/project-spec.coffee
bolinfest: Done.

spec/project-spec.coffee
bolinfest: Strangely, normally so do I...good catch: fixed!

spec/git-repository-provider-spec.coffee
```diff
@@ -0,0 +1,42 @@
...
+        waitsForPromise ->
+          directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
+          provider.repositoryForDirectory(directory).then (result) -> theResult = result
+
+        runs ->
+          expect(theResult).toBeInstanceOf GitRepository
+          expect(provider.pathToRepository[theResult.getPath()]).toBeTruthy()
+          expect(theResult.statusTask).toBeTruthy()
+
+        waitsForPromise ->
```
maxbrunsfeld: I think the code after its line should be in its own test:

```
describe "when given a Directory that is inside of a git repository", ->
  it "resolves to a GitRepository", ->

File: spec/project-spec.coffee [hunk 22]
```diff
@@ -197,6 +199,34 @@ describe "Project", ->
           atom.project.bufferForPath("b").then (anotherBuffer) ->
             expect(anotherBuffer).not.toBe buffer
 
+  describe ".repositoryForDirectory(directory)", ->
+    it "resolves to null when the directory does not have a repository", ->
+      theResult = "dummy_value"
```
maxbrunsfeld: You could remove some boilerplate from these tests by putting all of the code inside a single `waitsForPromise` call:

```coffee
waitsForPromise ->
  atom.project.repositoryForDirectory(directory).then (result) ->
    # other assertions...
    expect(result).toBeNull()
```

Though the indentation would be one level greater, you could avoid the extra `theResult` variable and the `runs`, so I think it would be simpler to read.

spec/git-repository-provider-spec.coffee
bolinfest: Done.

spec/project-spec.coffee
```diff
@@ -197,6 +199,26 @@ describe "Project", ->
...
+    it "resolves to a GitRepository and is cached when the given directory is a Git repo", ->
+      waitsForPromise ->
+        directory = new Directory(path.join(__dirname, '..'))
+        atom.project.repositoryForDirectory(directory).then (result) ->
+          expect(result).toBeInstanceOf GitRepository
+          dirPath = directory.getRealPathSync()
+          expect(result.getPath()).toBe path.join(dirPath, '.git')
+
+          # Verify that the result is cached.
+          expect(atom.project.repositoryForDirectory(directory)).toBe(atom.project.repositoryForDirectory(directory))
```
maxbrunsfeld: Sorry, I didn't say this correctly in an earlier comment. This should just be:

```coffee
expect(atom.project.repositoryForDirectory(directory)).toBe(result)
```

maxbrunsfeld: Disregard.

maxbrunsfeld:
Awesome work @bolinfest!

spec/project-spec.coffee
bolinfest: I added the variable back and shortened the assertion.

Made sure entire file still passes.

bolinfest:
@nathansobo OK, I'm happy to add more parens since that's closer to JS :) I was trying to make my CS as idiomatic as possible.

I don't suppose you have a linter for stuff like this?

nathansobo:
@bolinfest No we don't. CoffeeScript is so loose I wonder how useful that would be. Every line is a unique snowflake :wink:.

bolinfest:
Once I write my CoffeeScript-to-6to5 transpiler, we can use eslint and all this will be forgotten :)

maxbrunsfeld:
:100: :sparkles:

nathansobo:
:hamburger: :hammer:

PR #5494, opened 2015-02-11
Make getPaths() an async method that leverages repositoryForDirectory().

bolinfest:
This changes `getPaths()` to be an async method that iterates over the `RepositoryProvider`
objects registered via `atom.repository-provider` and asks each one to try to create a `Repository`
for the `Directory` that corresponds to the path. This uses the async `repositoryForDirectory()` API
instead of the `createRepositorySync()` API that was used before.

Currently, `getPaths()` maps a path to a `Directory` asynchronously via the following basic logic:

```
Promise.resolve(new Directory(path))
```

But in a follow up diff, I'll introduce the `DirectoryProvider` described in
https://github.com/atom/atom/pull/5428, so it will be possible to register a custom handler
to map special paths, such as `ftp://`.

This builds on https://github.com/atom/atom/pull/5491.

### DIFF
--- a/src/git-repository-provider.coffee
+++ b/src/git-repository-provider.coffee
@@ -1,0 +1,80 @@
+fs = require 'fs'
+GitRepository = require './git-repository'
+
+# Checks whether the specified directory has, or has a parent directory that
+# has, a valid .git folder, indicating the root of a Git repository.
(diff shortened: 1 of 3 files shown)


bolinfest:
This PR should be abandoned in favor of https://github.com/atom/atom/pull/5531.

PR #5522, opened 2015-02-12
Return pane items from the config opener, rather than using workspace.open

benogle:
This way, options like `split: ‘left’` can be used with the config URIs.


benogle:
Markdown preview html export error. Merging.

PR #5526, opened 2015-02-12
Move reactCompat option to transform as per 6to5 deprecation warning.

bolinfest:
Bumped the version of 6to5-core to `^3.6.0` to ensure `reactCompat` is
available as a transform rather than an option.

Verified that the deprecation warning went away in my console and that a
package that uses JSX+React still works as expected.


kevinsawicki:
Thanks!

PR #5528, opened 2015-02-12
:memo: a little hint added for 'Building from source for linux'

vpeil:
_script/build_ has the shebang #!/usr/bin/env node. On some system this command is no available but nodejs (e.g. on debian).
The softlink fixes this problem, and all the other steps work fine.


kevinsawicki:
Thanks!

mehcode:
@vpeil @kevinsawicki I don't think we should recommend users to modify things in `/usr` directly or outside of the package manager.

The correct solution to this would be (on debian):

```
sudo update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
```

kevinsawicki:
@mehcode sounds good, thanks for the tip.

kevinsawicki:
@mehcode Just pushed that change up :+1:

vpeil:
@mehcode thanks for the tip.
and thanks to the whole team for this awesome editor.

PR #5531, opened 2015-02-12
Change Project::setPaths() to be async.

bolinfest:


bolinfest:
In its current form, this causes 1148 failures when I run `apm test`.

Presumably this is because I have not updated `spec-helper.coffee` yet, which likely assumes `new Project` to finish synchronously.

nathansobo:
I'm not sure we can make this change without breaking the API. The current assumption is that setting paths is synchronous and immediately afterward the consequences of setting the paths have alread taken place.

bolinfest:
Yep, I agree. It's interesting to see how far up the chain you have to go
so soon once you start to replace sync APIs with async ones.

I'm afk right now, but feel free to close.

bolinfest:
Closing.

PR #5534, opened 2015-02-12
Optimize auto-indent on paste

### LINKED ISSUE #5411: Editor freezes after a huge paste
it's freeze,after i was paste some text (it's 79000's),i wish atom can handle 79898 lines of text from paste method.

### LINKED ISSUE #5529: Pasting large amounts of text has gotten really slow
Here I'm pasting ~800 lines of text and it's taking about 100ms. Looks like we're spending a ton of time in `setIndentationForBufferRow`, probably because the presenter is doing more work synchronously in these cases. @maxbrunsfeld do you think we could auto-indent before inserting the text to avoid this? 

screenshot 2015-02-12 13 31 12

Sync vs async presenter updates is a tricky balancing act because async will be slower in extremely common cases such as inserting a character but faster in less common ones like indenting 800 lines. I may need to revisit my approach and do more of the presenter work async.

nathansobo:
Previously, auto-indent-on-paste updated pasted text _after_ inserting it into the buffer, causing tons of buffer changes events to be emitted, which is very expensive. Now, we adjust all but the first line of the inserted text before inserting it into the buffer. The first line still needs to be post-processed in case it requires indentation to be less than the insertion column.

Fixes #5529
Fixes #5411

### DIFF
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -229,11 +229,20 @@ class LanguageMode
   #
   # Returns a {Number}.
   suggestedIndentForBufferRow: (bufferRow, options) ->
+    tokenizedLine = @editor.displayBuffer.tokenizedBuffer.tokenizedLineForRow(bufferRow)
+    @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, tokenizedLine, options)
+
+  suggestedIndentForLineAtBufferRow: (bufferRow, line, options) ->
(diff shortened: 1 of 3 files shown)


nathansobo:
@izuzak: @maxbrunsfeld and I tested this for a while, but maybe you could try to break it too? We're going to merge because the paste slowness is pretty egregious but it would be nice to get out in front of any issues we've missed.

izuzak:
@nathansobo Sure thing -- I just tried pasting crazy-indented blocks and seems to be working fine. :clap:

nathansobo:
@izuzak :bow:

PR #5535, opened 2015-02-12
Introduce atom.directory-provider service.

bolinfest:
A `Project` will always have a `DefaultDirectoryProvider` that
will be used if there are no other `DirectoryProvider` objects
that can produce a `Directory` for a path.

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -65,7 +65,7 @@ describe "Project", ->
       expect(otherDummyDirectory.getPath()).toBe otherRemotePath
       expect(otherDummyDirectory instanceof DummyDirectory).toBe true
 
-    it "a custom DirectoryProvider that returns null defaults to the DefaultDirectoryProvider", ->
+    it "uses the default directory provider if no custom provider can handle the URI", ->
       directoryProvider =
         directoryForURISync: (uri) -> null
         directoryForURI: (uri) -> throw new Error("This should not be called.")
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -14,6 +14,76 @@ describe "Project", ->
     atom.project.setPaths([atom.project.getDirectories()[0]?.resolve('dir')])
 
   describe "constructor", ->
+    it "enables a custom DirectoryProvider to supersede the DefaultDirectoryProvider", ->
+      remotePath = "ssh://foreign-directory:8080/"
+      class DummyDirectory
+        constructor: (@path) ->
+        getPath: -> @path
+        getFile: -> existsSync: -> false
+        getSubdirectory: -> existsSync: -> false
+        isRoot: -> true
+        off: ->
+        contains: (filePath) -> filePath.startsWith(remotePath)
+
+      directoryProvider =
+        directoryForURISync: (uri) ->
+          if uri.startsWith("ssh://")
+            new DummyDirectory(uri)
+          else
+            null
+        directoryForURI: (uri) -> throw new Error("This should not be called.")
+      atom.packages.serviceHub.provide(
+        "atom.directory-provider", "0.1.0", directoryProvider)
+
+      expect(atom.project.directoryProviders.length).toBe 2
+      expect(atom.project.directoryProviders[0]).toBe directoryProvider
+
+      tmp = temp.mkdirSync()
+      atom.project.setPaths([tmp, remotePath])
+      directories = atom.project.getDirectories()
+      expect(directories.length).toBe 2
+
+      localDirectory = directories[0]
+      expect(localDirectory.getPath()).toBe tmp
+      expect(localDirectory instanceof Directory).toBe true
+
+      dummyDirectory = directories[1]
+      expect(dummyDirectory.getPath()).toBe remotePath
+      expect(dummyDirectory instanceof DummyDirectory).toBe true
+
+      expect(atom.project.getPaths()).toEqual([tmp, remotePath])
+
+      # Make sure that DummyDirectory.contains() is honored.
+      remotePathSubdirectory = remotePath + "a/subdirectory"
+      atom.project.addPath(remotePathSubdirectory)
+      expect(atom.project.getDirectories().length).toBe 2
+
+      # Make sure that a new DummyDirectory that is not contained by the first
+      # DummyDirectory can be added.
+      otherRemotePath = "ssh://other-foreign-directory:8080/"
+      atom.project.addPath(otherRemotePath)
+      newDirectories = atom.project.getDirectories()
+      expect(newDirectories.length).toBe 3
+      otherDummyDirectory = newDirectories[2]
+      expect(otherDummyDirectory.getPath()).toBe otherRemotePath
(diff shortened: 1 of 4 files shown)


bolinfest:
I'll add a unit test now. `apm test` for the whole project passed, though.

src/project.coffee
```diff
@@ -39,6 +39,14 @@ class Project extends Model
     @emitter = new Emitter
     @buffers ?= []
 
+    @directoryProviders = [new DefaultDirectoryProvider()]
+    atom.packages.serviceHub.consume(
+      'atom.directory-provider',
+      '^0.1.0',
+      # New providers are added to the front of @directoryProviders because
+      # DefaultDirectoryProvider is a catch-all that will always provide a Directory.
+      (provider) => @directoryProviders.splice(0, 0, provider))
```
maxbrunsfeld:
```coffee
@directoryProviders.unshift(provider)
```

maxbrunsfeld:
Could you also add a test that exercises the code added in `Project`? I would do something like you did for the `RepositoryProvider` PR, where you register a dummy directory provider for some URI scheme and check that it is used.

src/project.coffee
maxbrunsfeld: Do we need to be able to define alternative `Directory` implementations for local paths? Because it'd be nice to have the most common case be first in line, rather than last. We would just have to have `DefaultDirectoryProvider` return `null` if the `uri` argument had a protocol other than `file://`. That way the `Project` would be able to tell if a given path was usable or not, possibly throwing an exception like this:

```
Could not open a directory for the URI 'ftp://some-host/some/path'.
You need to install a package which implements the 'atom.directory-provider' service for that URI.
```

nathansobo: :+1: @maxbrunsfeld I agree with your thoughts on this. It also ties into @bolinfest's idea about having package metadata for the service provider prefix. Another wild thought... could we use the provider's service name make the protocol explicit? `atom.directory-provider.webdav`? Not sure it's a good idea.

maxbrunsfeld: My gut says it's better to let the provider perform ad-hoc logic to decide if it can open the URI or not. For example, our default directory provider can work with `/the/path` or `file://the/path`.

nathansobo: @bolinfest You're largely responsible for pushing me toward avoiding the requirement of running code to make these kinds of decisions. What do you think?

bolinfest: I agree with @maxbrunsfeld that if this put is in a spot where you had to
use a file:// uri instead of a path everywhere, that would be very
frustrating.

nathansobo: Alright, we can use some other metadata mechanism if we need to. I'm a go on this.

src/default-directory-provider.coffee
```diff
@@ -0,0 +1,42 @@
...
+class DefaultDirectoryProvider
+
+  # Public: Create a Directory that corresponds to the specified URI.
+  #
+  # * `uri` {String} The path to the directory to add.
+  # * `existingPaths` {Array} The existing paths in atom.project.
+  # Returns:
+  # * {Directory} if the given URI is compatible with this provider.
+  # * {string} if the uri maps to a path an existingPath, in which case
+  #     the path in existingPaths is returned.
```
maxbrunsfeld: I think this `String` return value makes the method interface a little too complex. What's the use case that you're thinking of where this uniqueness check needs to defined by the provider, rather than just being done inside the project? Maybe we can solve the problem a different way.

bolinfest: The problem is that normalize() was being called on the path before the directory provider was being consulted, which is not appropriate for URIs. It's not clear to me that `Project` should be responsible for knowing when there is already an appropriate root folder.

There's also the special check for `(directoryPath.indexOf(path.join(existingPath, path.sep)) is 0)`, which seems specific to `DefaultDirectoryProvider`. The idea of `contains()` is specific to a `Directory`, which is the thing we're trying to create, so it doesn't exist yet.

maxbrunsfeld: But we have the list of existing `Directory` objects, so can we check whether any of them `contains` the new path?

bolinfest: Sure. I'll have to see if `Directory.contains(arg)` does the right thing when `arg` is not normalized.

maxbrunsfeld: Yeah, good point. If not, I'd like to implement said behavior so that we can simplify the `DirectoryProvider` interface.

spec/project-spec.coffee
```diff
@@ -14,6 +14,76 @@ describe "Project", ->
...
+          if uri.startsWith("ssh://")
+            new DummyDirectory(uri)
+          else
+            null
+        directoryForURI: (uri) -> throw new Error("This should not be called.")
+      atom.packages.serviceHub.provide(
+        "atom.directory-provider", "0.1.0", directoryProvider)
+
+      expect(atom.project.directoryProviders.length).toBe 2
+      expect(atom.project.directoryProviders[0]).toBe directoryProvider
```
maxbrunsfeld: I would say :fire: these assertions about `.directoryProviders`, as that property is private.

src/project.coffee
```diff
@@ -182,22 +190,22 @@ class Project extends Model
...
+      # Apparently a Directory does not believe it can contain itself, so we
+      # must also check whether the paths match.
+      return if directory.contains(projectPath) or directory.getPath() is projectPath
+
+    directory = null
+    for provider in @directoryProviders
+      break if directory = provider.directoryForURISync?(projectPath)
+    if directory is null
+      # This should never happen because DefaultDirectoryProvider should always
+      # return a Directory.
```
maxbrunsfeld: I still think that if an unrecognized URI is given, this *should* happen. The `DefaultDirectoryProvider` should return `null` unless the URI is for a local file. That way, this would actually be a useful error.

I also still think the `DefaultDirectoryProvider` should be tried first, not last, and that it should simply be called `LocalDirectoryProvider`.

spec/project-spec.coffee
```diff
@@ -14,6 +14,70 @@ describe "Project", ->
...
+      # DummyDirectory can be added.
+      otherRemotePath = "ssh://other-foreign-directory:8080/"
+      atom.project.addPath(otherRemotePath)
+      newDirectories = atom.project.getDirectories()
+      expect(newDirectories.length).toBe 3
+      otherDummyDirectory = newDirectories[2]
+      expect(otherDummyDirectory.getPath()).toBe otherRemotePath
+      expect(otherDummyDirectory instanceof DummyDirectory).toBe true
+
+    it "a custom DirectoryProvider that returns null defaults to the DefaultDirectoryProvider", ->
```
maxbrunsfeld: For readability with the `it`, I'd say:

```coffee
it "uses the default directory provider if no custom provider can handle the URI", ->
```

src/project.coffee
maxbrunsfeld: :cyclone: As per conversation with @bolinfest, I think this approach is actually better.

spec/project-spec.coffee
bolinfest: so rewritten

PR #5537, opened 2015-02-13
Try to set the Project's repo if it does not have one when a new Reposit...

bolinfest:
...oryProvider is registered.

I tested this using my test `HgRepositoryProvider`. Now when I run the following from the
command line:

```
atom <path-to-directory-with-hg-repository>
```

And then run the following in the console:

```
atom.project.getRepositories()
```

I get an array with an `HgRepository` in it. Previously, I got an empty array because the
`Project`'s paths were set before my `HgRepositoryProvider` was registered.

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -15,12 +15,13 @@ describe "Project", ->
 
   describe "constructor", ->
     it "tries to update repositories when a new RepositoryProvider is registered", ->
-      atom.project.setPaths(["/tmp"])
+      tmp = temp.mkdirSync('atom-project')
+      atom.project.setPaths([tmp])
       expect(atom.project.getRepositories()).toEqual [null]
       expect(atom.project.repositoryProviders.length).toEqual 1
 
       # Register a new RepositoryProvider.
-      dummyRepository = destroy: () ->
+      dummyRepository = destroy: ->
       repositoryProvider =
         repositoryForDirectory: (directory) -> Promise.resolve(dummyRepository)
         repositoryForDirectorySync: (directory) -> dummyRepository
@@ -37,7 +38,7 @@ describe "Project", ->
       expect(repository).toBeTruthy()
 
       # Register a new RepositoryProvider.
-      dummyRepository = destroy: () ->
+      dummyRepository = destroy: ->
       repositoryProvider =
         repositoryForDirectory: (directory) -> Promise.resolve(dummyRepository)
         repositoryForDirectorySync: (directory) -> dummyRepository
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -13,6 +13,39 @@ describe "Project", ->
   beforeEach ->
     atom.project.setPaths([atom.project.getDirectories()[0]?.resolve('dir')])
 
+  describe "constructor", ->
+    it "tries to update repositories when a new RepositoryProvider is registered", ->
+      atom.project.setPaths(["/tmp"])
+      expect(atom.project.getRepositories()).toEqual [null]
+      expect(atom.project.repositoryProviders.length).toEqual 1
+
+      # Register a new RepositoryProvider.
+      dummyRepository = destroy: () ->
+      repositoryProvider =
+        repositoryForDirectory: (directory) -> Promise.resolve(dummyRepository)
+        repositoryForDirectorySync: (directory) -> dummyRepository
+      atom.packages.serviceHub.provide(
+        "atom.repository-provider", "0.1.0", repositoryProvider)
+
+      expect(atom.project.repositoryProviders.length).toBe 2
+      expect(atom.project.getRepositories()).toEqual [dummyRepository]
+
+    it "does not update @repositories if every path has a Repository", ->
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
@bolinfest this seems pretty important. Can you add a test covering this case?

bolinfest:
Absolutely. Is there an easy way to create an atom package with the test
Project that has the metadata that would trigger this? Or should I just
make this logic exercisable from a "visible for testing" method?

maxbrunsfeld:
For now, could you just test it by using the private `atom.packages.serviceHub` variable to provide a dummy repo-provider?

bolinfest:
Blargh, `spec-helper.coffee` does:

```
  atom.project = new Project(paths: [projectPath])
  atom.workspace = new Workspace()
  atom.packages.serviceHub = new ServiceHub
```

so it resets `atom.packages.serviceHub` after `atom.project` has just used it. I'll see if reordering these lines breaks anything.

bolinfest:
Fortunately all tests still pass with:

```
  atom.packages.serviceHub = new ServiceHub
  atom.project = new Project(paths: [projectPath])
  atom.workspace = new Workspace()
```

bolinfest:
OK, added test cases both for when there are paths without repositories and when there aren't.

spec/project-spec.coffee
```diff
@@ -13,6 +13,37 @@ describe "Project", ->
...
+      repositoryProvider =
+        repositoryForDirectory: (directory) -> Promise.resolve(dummyRepository)
+        repositoryForDirectorySync: (directory) -> dummyRepository
+      atom.packages.serviceHub.provide(
+        "atom.repository-provider", "0.1.0", repositoryProvider)
+
+      expect(atom.project.repositoryProviders.length).toBe 2
+      expect(atom.project.getRepositories()).toEqual [dummyRepository]
+
+    it "does not update @repositories if every path has a Repository", ->
```
maxbrunsfeld: Is it true that every path has a repository in this case? It looks like the path is being set to `spec/fixtures/dir` here, which doesn't contain a git repo.

bolinfest: Yep, exactly. /tmp does not have a repository initially and then the
provider supplies one.

bolinfest: Oh, in the second case, I thought that atom.project points to the Atom repo
itself by default, but I should confirm.

maxbrunsfeld: It points to the `spec/fixtures` directory by default, but it is being set to `spec/fixtures/dir` in [the beforeEach](https://github.com/atom/atom/pull/5537/files#diff-f42f793b4715cfbf8d07c772bc142849L14) of the test.

bolinfest:
@maxbrunsfeld You're right. But `GitRepositoryProvider` takes a `Directory` and searches upwards until it finds a parent with a `.git` subdirectory or the root of the file tree. So in this case, it finds the Atom repo.

maxbrunsfeld:
Got it :+1:. Thanks for adding that extra assertion in there in case anyone else is confused about that.

bolinfest:
Yep, of course. Let me know if there are any other scenarios that you think should be tested.

bolinfest:
Also, if we could get this in before the 0.180.0 release, I'd be much
obliged.

spec/project-spec.coffee
```diff
@@ -13,6 +13,39 @@ describe "Project", ->
...
+      repositories = atom.project.getRepositories()
+      expect(repositories.length).toEqual 1
+      [repository] = repositories
+      expect(repository).toBeTruthy()
+
+      # Register a new RepositoryProvider.
+      dummyRepository = destroy: () ->
```
kevinsawicki: :fire: `()`

spec/project-spec.coffee
```diff
@@ -13,6 +13,39 @@ describe "Project", ->
   beforeEach ->
     atom.project.setPaths([atom.project.getDirectories()[0]?.resolve('dir')])
 
+  describe "constructor", ->
+    it "tries to update repositories when a new RepositoryProvider is registered", ->
+      atom.project.setPaths(["/tmp"])
```
kevinsawicki: This shouldn't use a hard-coded platform-specific path, it should use something like `temp.mkdirSync('atom-project')` instead.

spec/project-spec.coffee
```diff
@@ -13,6 +13,39 @@ describe "Project", ->
...
+      expect(atom.project.getRepositories()).toEqual [null]
+      expect(atom.project.repositoryProviders.length).toEqual 1
+
+      # Register a new RepositoryProvider.
+      dummyRepository = destroy: () ->
```
kevinsawicki: :fire: `()`

spec/project-spec.coffee
bolinfest: done

spec/project-spec.coffee
bolinfest: done

spec/project-spec.coffee
bolinfest: done

PR #5558, opened 2015-02-14
Fix typos in API docs

jeremy-engel:
The word `optional` was misspelled as `optinal` two times in the API docs. This PR corrects those errors.


izuzak:
Thanks!

PR #5561, opened 2015-02-14
Add reload step to the ASCII Art tutorial

silvenon:


izuzak:
Sorry for not getting to this sooner -- could you reopen this on the new repository for the docs: https://github.com/atom/docs. Most docs in this repository will be removed so that we don't have duplicated content around. :book:

silvenon:
The reload tip is already there :smiley:

izuzak:
> The reload tip is already there

:tada: :birthday: :gift:

PR #5567, opened 2015-02-15
Indent soft wrap

### LINKED ISSUE #1125: Softwrap should match indent
Ok, this one may be a personal preference, but it drives me nuts that softwrap doesn't match the indent.

screen shot 2013-11-18 at 5 18 17 pm

as-cii:
This fixes #1125.

It still needs a lot of refactoring (and some test cases too), but I just wanted to evaluate the approach that me and @nathansobo agreed on the aforementioned issue. To recap, the following is a brief explanation of how the solution works.

When lines with an indentation are soft wrapped, they are prepended with a phantom token which is as long as the indentation of the original line. This token is not selectable, nor reachable by the cursor. I had to change how `DisplayBuffer#clipScreenPosition` worked, since I could not treat the phantom token as an hard tab. I'd like to remark that this is still experimental code, but based on my tests I can confirm it works pretty well.

~~There's a subtle bug, though, which needs to be addressed and which I'll be looking at soon after creating this pull request. When a line is indented but no space is present, Atom freezes: I am quite sure this is due to how soft wrapping works on lines without spaces (namely, lines without words). I am quite sure this can be fixed easily, it just needs some more attention (and a better knowledge of the codebase).~~

Thanks everyone in advance for the support!

/cc: @atom/core


as-cii:
The bug mentioned above was just fixed. This code needs some :heart:, but I think we are on the right track (at least I hope so!) :smile: 

Please, feel free to leave opinions and critics about the approach I suggested. Thank you!

src/display-buffer.coffee
```diff
@@ -860,17 +863,19 @@ class DisplayBuffer extends Model
   # Returns `null` if a wrap wouldn't occur.
   findWrapColumn: (line, softWrapColumn=@getSoftWrapColumn()) ->
```
as-cii: This method seems a bit Feature Envy (it deals mostly with another class internal). Shall we move it into `TokenizedLine`?

abe33:
This is great to see this issue in the process of being fixed! :clap: 

Out of curiosity, how does it plays with decorations? Does a selection spanning the wrapped line includes the phantom token?

as-cii:
> This is great to see this issue in the process of being fixed!

:blush:

> Out of curiosity, how does it plays with decorations? Does a selection spanning the wrapped line includes the phantom token?

screen shot 2015-02-17 at 09 07 27

Yes, it does. I wondered whether it should be included or not and, to be honest, I am not sure I know the answer :sweat_smile: I have a gut feeling that it shouldn't be selected at all, but I'd like to have some feedback before proceeding to change how selection works.

abe33:
> Yes, it does. I wondered whether it should be included or not and, to be honest, I am not sure I know the answer :sweat_smile: I have a gut feeling that it shouldn't be selected at all, but I'd like to have some feedback before proceeding to change how selection works.

IMO, given that selections go all the way to the right edge of the buffer when spanning many lines, it would make sense it includes the phantom token to start the new line selection from the left edge of the buffer.

nathansobo:
The code looks reasonable at first glance. I'm going to build and play with this now. It concerns me that there's no test coverage however. For each addition to the code, there should be a spec that fails if that addition isn't present.

src/display-buffer.coffee
nathansobo: Yeah, let's try moving it into `TokenizedLine` and see how it feels. The method isn't part of the public API.

src/tokenized-line.coffee
```diff
@@ -97,25 +99,44 @@ class TokenizedLine
       leftTextLength += nextToken.value.length
       leftTokens.push nextToken
 
+    tab = leftTokens[0].buildPhantomToken(@indentLevel * @tabLength)
```
nathansobo: I think we should call this `leadingWhitespace` or something to clarify it isn't technically a "tab".

src/tokenized-line.coffee
```diff
@@ -97,25 +99,44 @@ class TokenizedLine
...
+      lineEnding: @lineEnding,
+      indentLevel: @indentLevel,
+      tabLength: @tabLength
     )
     [leftFragment, rightFragment]
 
   isSoftWrapped: ->
     @lineEnding is null
 
+  isOutsidePhantomToken: (column) ->
```
nathansobo: maybe `isColumnOutsidePhantomToken`?

nathansobo:
Okay, after test driving this everything is looking good. Really nice work @as-cii. We just need thorough test coverage... like I mentioned, no part of this diff should be able to be reverted without a spec failing.

nathansobo:
Sorry, I should also add that I agree with @abe33 about selection rendering. I think the selection going all the way to the left is the right approach.

grizzilus:
Glad to see this is progressing. One small issue: if the leading whitespace is set to display, then the phantom tokens render as spaces, even if the document uses hard tabs. See the screenshot below. I don't know what the best behavior would be (not showing those character, showing a matching whitespace character for the document, or something else), but I don't think this is the best option.

matching

nathansobo:
@grizzilus You're absolutely right. I'm such a bad QA tester.

grizzilus:
Also, I just want to throw out there that I don't think the selection should go to the left edge. It makes it less obvious when selecting a large chunk of text whether you are selecting multiple indented lines or a single soft-wrapped indented line. Line numbers and maybe some modifications to the way the whitespace on the left edge of the soft-wrapped indented line help the situation, but it still seems a bit off.

I know highlighting the selection to the left edge would match the highlighting to the right edge, but that may not necessarily be the correct approach either. There is a similar discussion in atom/find-and-replace#130. The reasoning there for searching for `\n` seems to apply to selecting whole lines as well.

as-cii:
First of all, let me say that I am really happy to see such involvement into this pull-request: thank you guys for testing and reporting issues.

> It concerns me that there's no test coverage however. For each addition to the code, there should be a spec that fails if that addition isn't present.

@nathansobo: thanks for pointing this out, that's a concern for me too. I'll make sure to write some new specs (e.g. phantom token should be invisible, phantom token shouldn't be reachable by the cursor, etc.) as well as fixing [the failing ones that are currently checking the soft wrapping behavior](https://github.com/atom/atom/blob/master/spec/display-buffer-spec.coffee#L64-234).

> IMO, given that selections go all the way to the right edge of the buffer when spanning many lines, it would make sense it includes the phantom token to start the new line selection from the left edge of the buffer.

@nathansobo @abe33: isn't it like this already? Just to make sure I have understood, here's another screenshot which illustrates how it works right now.

soft-wrapping

Selection will reach the left and the right edges. How is this different from your suggested approach?

> Also, I just want to throw out there that I don't think the selection should go to the left edge. It makes it less obvious when selecting a large chunk of text whether you are selecting multiple indented lines or a single soft-wrapped indented line.

@grizzilus: nice observation (indeed, that was part of the reason why I didn't stick with one approach or another). On the other hand, if you display invisibles you'll be able to understand whether indentation comes from real tabs or it is simply carried from a wrapped line (see previous screenshot).

grizzilus:
I understand that looking at the invisibles or the line numbers makes it simple to tell the difference between a soft-wrapped indented line and multiple indented lines,but I still think the selection should match what does into the selection buffer without the extra padding to the right or the left. So I guess it is less an issue of this PR and more a design issue (since selections extend to the right edge). 

One other small issue: I think the indent guide should appear on the soft-wrapped lines. I think hiding the leading invisibles on the soft-wrapped lines also hides the indent guide.

nathansobo:
> @nathansobo @abe33: isn't it like this already?

Yes, it is. I was just saying I think this is the correct behavior. This is all subjective, but my strong preference is to keep the behavior consistent with our treatment on the right hand side. When selections wrap between 2 locations on screen, they should always extend to either edge. If we change this on the left, we should change it on the right, but then we'd be inconsistent with selection rendering in most other editors.

Ping me when you have a green core spec suite and test coverage for all aspects of this. I'm very excited to get it merged. :sparkles:

as-cii:
> Ping me when you have a green core spec suite and test coverage for all aspects of this. I'm very excited to get it merged.

@nathansobo: I got rid of `wrapAtPhantomTokens` and used `wrapAtSoftNewlines` instead. It seems like everything is green now (both core and newly added specs) and, hopefully, we can :shipit:

Thanks to everyone who got involved on this: your help was very precious! :pray:

src/display-buffer.coffee
```diff
@@ -841,9 +841,15 @@ class DisplayBuffer extends Model
     if screenLine.isSoftWrapped() and column >= maxScreenColumn
       if wrapAtSoftNewlines
         row++
-        column = 0
+        column = @screenLines[row].clipScreenColumn(0)
       else
         column = screenLine.clipScreenColumn(maxScreenColumn - 1)
+    else if screenLine.isColumnInsidePhantomToken(column)
+      if wrapAtSoftNewlines
```
nathansobo: I would think this would be the other way around... when `wrapAtSoftNewlines` is `true` we're supposed to round *up* when clipping. Otherwise we round down. Am I missing something?

src/cursor.coffee
```diff
@@ -304,13 +304,14 @@ class Cursor extends Model
...
 
+      originalRow = row
       while columnCount > column and row > 0
         columnCount -= column
         column = @editor.lineTextForScreenRow(--row).length
         columnCount-- # subtract 1 for the row move
 
       column = column - columnCount
-      @setScreenPosition({row, column})
+      @setScreenPosition({row, column}, wrapAtSoftNewlines: originalRow == row)
```
nathansobo: My intuition would be that we *always* wrap at soft newlines when moving right, never when moving left. If we're trying to move rightward, landing in a phantom token moves us further rightward. If we're trying to move leftward, the opposite is true.

src/token.coffee
```diff
@@ -144,6 +144,15 @@ class Token
       isHardTab: isHardTab
     )
 
+  buildPhantomTabToken: (length) ->
```
nathansobo: How about `buildSoftWrapIndentToken`. I was confused at first thinking this had something to do with tab when it doesn't.

src/cursor.coffee
as-cii: > My intuition would be that we always wrap at soft newlines when moving right, never when moving left.

That was the behavior before https://github.com/as-cii/atom/commit/e5eaa2538189ad2424fc13e6b204c10460026be5. Unfortunately, that causes an issue when moving down from a soft wrapped line which has a phantom token right under the current column (e.g. `{ currentRow + 1, currentColumn }`). Basically the cursor moves to the end of the current row, rather than to the beginning of the next row, whereas I would expect the opposite. The same happens when clicking on a phantom token: the cursor gets moved at the end of the previous row, rather than to the first *real token* of the clicked line.

An alternative could be to force `wrapAtSoftNewlines` when moving down and up, but what about the scenario in which a user clicks the aforementioned empty space (namely, the phantom token)?

grizzilus:
Is this the desired behavior for indent guides on soft-wrapped lines? It seems that logically the indent guides should continue through the soft-wrapping.

screenshot from 2015-02-19 09 27 45

src/display-buffer.coffee
nathansobo: Sure enough, when I invert this logic and only move forward when `wrapAtSoftNewlines` is `true`, I no longer need to pass an option in `moveLeft`. However, it breaks some existing specs that I *think* might actually be wrong. The are all specs for clipping the position at the beginning of a continuation of a soft-wrapped line... Pretty sure we should always back up by 1 column when wrapping back up before the soft-wrap so we're not leaving the cursor on an invalid position.

thomasjo:
At least initially, I disagree with that @grizzilus. Wouldn't indent guides on soft-wrapped lines be somewhat confusing? You cannot affect the indentation level on a soft-wrapped line, only the "real" line.

Maybe we need to test both cases? Not sure how complicated it would be to implement though.

src/display-buffer.coffee
as-cii: > However, it breaks some existing specs that I think might actually be wrong. The are all specs for clipping the position at the beginning of a continuation of a soft-wrapped line...

Yes, as mentioned in [another comment](https://github.com/atom/atom/pull/5567#discussion_r25004660) the reason why I came up with such behavior (and its related specs) was that I wanted the cursor to bump to the first *real* character of the line whenever a user clicked on the empty space (aka phantom token) of an indented soft wrapped line, or when moving down (see previous comment for a better explanation of this issue). Assuming this behavior is correct, it seemed reasonable enough to change `moveLeft` to pass a parameter when clipping screenPosition, rather than changing both `moveDown` and the click handler.

nathansobo:
@as-cii I'm working on some commits for this PR. Could you add me as a collaborator so I can push them?

src/cursor.coffee
nathansobo: I'm trying a commit where `wrapAtSoftNewlines` and `skipAtomicTokens` are `true` when moving down. It probably should have been that way anyway.

nathansobo:
Actually I can just send a PR the other direction.

as-cii:
@nathansobo Never mind, already added you :wink:

nathansobo:
@as-cii Can you take a look at https://github.com/as-cii/atom/pull/1.

After trying it out I like the behavior. The one thing is that clicking in a phantom token moves you up to the previous line, but I think that behavior is actually more consistent with click handling elsewhere. For all other cases, we always move you to the closest valid position *before* the location of your click. I think this fits into that scheme.

nathansobo:
Okay, I closed my PR but at least it demonstrates an idea. The main thing is I think the position should always be clipped greater when `wrapAtSoftNewlines` is `true`. Maybe things get simpler if you go back to having an explicit option for skipping phantom tokens? I bet they do.

nathansobo:
And by the way, I agree with @grizzilus that the indent guides should continue through the soft-wrapped region. It just looks weird when they don't.

nathansobo:
@as-cii I'm still seeing the problem where moving down from a long soft-wrapped line to a shorter one skips a line:

move-down-across-soft-wrap

I'm really thinking you should restore the `skipPhantomTokens` option and just use *that* when moving up and down. I wish we had a better name than "phantom" for these tokens in the code. Maybe `skipSoftWrapIndent`?

as-cii:
@nathansobo I introduced some new commits:

1. I used `softWrapIndentation` name consistently across `Token`, `TokenizedLine` and `DisplayBuffer`: I believe it expresses intent better than just `softWrapIndent` (which seems more like an action, rather than a *property*);
2. I removed `wrapAtSoftNewlines: true` as you suggested, because it caused way too many issues without necessarily adding value to the code; instead we use `skipSoftWrapIndentation` in `moveDown`, `moveUp` and `moveToFirstCharacterOfLine`;
3. Indent guides are now showed consistently: basically, instead of only one `softWrapIndentationToken` I create one for each indentation level;

The specs we have written have proved to be really valuable, as they worked as a safety net when I changed the behavior without getting in my way while I was playing with the code :+1: 

/cc: @grizzilus @thomasjo

grizzilus:
@as-cii Just built; looks much better with the indent guides. I'll keep playing around with this branch and see if anything comes up, but this looks great!

nathansobo:
I've really loved collaborating with you on this @as-cii. Amazing work here. Maybe when manual DOM updates branch gets merged again you can rebase this and then we can reach out to the community to test drive it over the weekend? If everything is :sunglasses: we can merge it Monday in time for release next week.

as-cii:
> I'll keep playing around with this branch and see if anything comes up, but this looks great!

> I've really loved collaborating with you on this @as-cii. Amazing work here.

That's very kind of you guys :blush: Thank you so much for your great effort in supporting and helping me: I really appreciated it! :pray: 

> Maybe when manual DOM updates branch gets merged again you can rebase this and then we can reach out to the community to test drive it over the weekend

@nathansobo: I noticed that the "manual DOM updates branch" was eventually merged into master, therefore I aligned this branch with it. I preferred to merge instead of rebasing to avoid history rewriting, so it's a bit easier for everyone to pull recent changes (and to avoid all those notifications caused by mentions inside commit messages). Would you agree to wait for rebasing until we are ready to merge this into master?

By the way, I think I can say this is ready for test driving! :smile:

nathansobo:
Merge is fine. We don't need to re base before merging to master.

LucasKA:
Whoa, great job y'all, and a really good feature. :+1:

nathansobo:
@as-cii Can you merge master into this so we can test it out today?

as-cii:
> Can you merge master into this so we can test it out **today**?

@nathansobo I am sorry for answering a bit late to this, but it was already evening here in Italy :sleeping: 

I just merged master into this branch and pushed it already: now it should be ready for testing! :smile:

izuzak:
I've been playing with this and trying to break it :smile_cat:, but it seems that it's working really well. Great work @as-cii! :zap:

I did notice a few things, so I'll mention them below, but as far as I can tell -- all of those are problems on `master` as well and weren't introduced with this PR. (/cc @nathansobo -- if you agree, I'll open separate issues for these things):

1. When I click on the line's number in the gutter, I expected to see the whole logical line to be selected (that's the behavior in Sublime, for example). Instead, only the physical line is selected (kind of, it strange):

  

2. Each physical line has a fold icon if the line has nested lines. Instead, I expected only the first physical line to have the fold icon (that's also the behavior in Sublime).

  

3. Folding an indented and soft-wrapped line un-soft-wraps the line. 

  

Again -- great work @as-cii. :bow:

nathansobo:
@izuzak Yeah, I agree. These all seem to be problems on master. We can create separate issues for them.

nathansobo:
@kevinsawicki You had mentioned possibly guarding this behind a flag or making it an option. Can you elaborate on the reasons? Are you concerned about regressions or do you want to give people the option?

kevinsawicki:
> You had mentioned possibly guarding this behind a flag or making it an option. Can you elaborate on the reasons? Are you concerned about regressions or do you want to give people the option?

Yeah, my concerns are performance, changing the default behavior, and also supporting a way to disable this behavior if we do find issues after it is released.

Making it an option that is off by default addresses all 3 of those concerns and allows people to test it out, give feedback, and still go back to the old behavior if they do find issues.

If we are confident in the implementation, test coverage, and this being the default behavior, then I'm fine releasing it without a feature flag though.

nathansobo:
I'm confident in this being the default behavior, performance, and test coverage. The only concern I have is the possibility of unexpected regressions, but we've had multiple people test this and I'm not sure a feature flag would completely protect us based on the nature of the changes. I think we should merge it when we're feeling good about the most recent release to give it a couple days to bake on master.

kevinsawicki:
>  I think we should merge it when we're feeling good about the most recent release to give it a couple days to bake on master.

Sounds good, 0.183 should be out in the next hour, so I'd say merge any time after that.

nathansobo:
@kevinsawicki I'm going to merge this at the end of the day if no issues crop up with 0.183.0.

nathansobo:
:fireworks: Glad to have this in. Everyone be on the lookout for any issues we missed during testing, but I feel pretty good about this.

theDoug:
:+1:

javaporter:
Outstanding. Thanks much for the work on this. This was the one issue for me which made Atom an unpleasant experience (otherwise it's been fantastic). When this update ships it's going to be a good day.

lee-dohm:
:clap: Great work!

thomasjo:
:sparkling_heart:

ioquatix:
Thanks guys this is bloody brilliant.

Cangit:
Built this morning- not looking back. This is great work. :thumbsup:

pierrepavlov:
Really looking forward to this feature! How do I get it to work though? I'm on 0.184.0 but can't get it to work. Am I missing something? :)

as-cii:
@pierrepavlov v0.184.0 doesn't ship with this yet. [v0.185.0](https://github.com/atom/atom/releases/tag/untagged-fb072b5b1ecad1fd2a89) will have it :wink:

Cangit:
@pierrepavlov you can build your own version of atom if you are impatient. [Follow the doc's](https://github.com/atom/atom#building)

joshrowley:
:clap: This is fantastic, nice work.

kevincoleman:
I’ve gotta jump on and add a big +1 for this work. Stellar, folks.

aaronransley:
Excellent job, all. @as-cii, @nathansobo especially - thanks for all the hard work. 

This one's been a long time coming :+1:

pierrepavlov:
Thanks for info @as-cii and @Cangit!

jtkiley:
Is there a way to get hanging indents for soft–wrapped lines? It would be nice to have continued lines display one tab stop in from the original line.

javaporter:
This went live this afternoon, works like a charm. Thanks again guys. Sublime was great, but is now uninstalled. Atom full on. :thumbsup:

chodorowicz:
Is there a way to indent the wrapped line a bit more, e.g. two spaces? So it's more clear that it's part of previous line, not a new one?

as-cii:
> Is there a way to indent the wrapped line a bit more, e.g. two spaces? So it's more clear that it's part of previous line, not a new one?

Nope, I'd say that the only way to visually understand that a line is the continuation of the previous one is to look at the sidebar (please, note the dot instead of the line number):

screen shot 2015-03-14 at 11 52 28

I have never found this particularly annoying, though, and, for what is worth, Sublime Text behaves the same way (it hasn't dots, but expands the line number all the way down to the last wrapped line):

screen shot 2015-03-14 at 11 59 36

chodorowicz:
@as-cii Thanks for reply. Yeah, it's not a big thing,  I suppose just a matter of noticing the dot and lack of empty characters. thanks!

jtkiley:
@chodorowicz @as-cii I'd also like a setting for further indenting wrapped lines. It's not as big of a deal for a lot of code (at least that I use), as many languages have both style norms and a set of syntax features that enable hard wrapping lines (e.g. PEP8). However, it's a huge convenience for LaTeX when you're working on documents that get a lot of editing. It's nice to have sentences on individual lines for diffs, and those single breaks get ignored when the document is typeset. The wrapping makes it very easy to find specific sentences and to edit sentence-by-sentence. Other editors do this, and I miss it a lot. I'd be thrilled to see it added, as I use Atom almost exclusively now. 

I'd just add a variable to the length of the wrapped indent, and give users a setting for "Soft Wrap Hanging Indent" that provides the value (perhaps defaulting to zero).

ioquatix:
@jtkiley Yes it would be nice if there were support for a hanging indent, I prefer it when reading code. +1 your proposal, hanging indent should be multiples of the files existing indentation character(s).

izuzak:
@jtkiley @chodorowicz @ioquatix Let's open a new issue so that we can track that feature request. Sound good?

jtkiley:
@izuzak I'm walking out the door now, but I'll create it when I get back.

chodorowicz:
good idea!

PR #5573, opened 2015-02-16
Use double colons for "before" and "after" pseudo-elements

postcasio:
In CSS3, single colon for pseudo-elements is deprecated.


kevinsawicki:
Nice, thanks for cleaning this up, :vertical_traffic_light:

PR #5579, opened 2015-02-17
Rename 6to5 to Babel

sebmck:
6to5 has been renamed to Babel, more information [here](http://babeljs.io/blog/2015/02/15/not-born-to-die/).


thedaniel:
Quickly skimming this code, it looks like it's mostly find-and-replace, and removes refs to 6to5 entirely. Even if we change the directive to `'use babel';` we've publicly written about `'use 6to5`;` so we need to support using that directive to invoke babel as well.

spec/fixtures/babel/single-quotes.js
```diff
@@ -1,3 +1,3 @@
-'use 6to5';
+'use babel';
```
kevinsawicki: Instead of changing the specs, I'd prefer to just add new ones for babel, so we make sure the legacy support is still tested.

sebmck: Sure.

sebmck:
@thedaniel Is that not what this does?

```javascript
unless /^("use 6to5"|'use 6to5'|"use babel"|'use babel')/.test(sourceCode)

```

sebmck:
Updated to include backwards compatibility tests for the old `use 6to5` directive.

spec/fixtures/6to5/double-quotes.js
```diff
@@ -1,3 +1,3 @@
-"use 6to5";
+"use babel";
```
kevinsawicki: This file can go now right? Since all the fixtures are in the new `babel` folder?

spec/fixtures/6to5/single-quotes.js
```diff
@@ -1,3 +1,3 @@
-'use 6to5';
+'use babel';
```
kevinsawicki: This file too is no longer used right?

spec/fixtures/6to5/double-quotes.js
sebmck: Whoops, accidentally committed that, sorry!

src/babel.coffee
```diff
@@ -132,7 +132,7 @@ transpile = (sourceCode, filePath, cachePath) ->
 # either generated on the fly or pulled from cache.
 loadFile = (module, filePath) ->
   sourceCode = fs.readFileSync(filePath, 'utf8')
-  unless sourceCode.startsWith('"use 6to5"') or sourceCode.startsWith("'use 6to5'")
+  unless /^("use 6to5"|'use 6to5'|"use babel"|'use babel')/.test(sourceCode)
```
bolinfest: I don't know if V8 caches regexes, but in the event that it doesn't, I would extract this regex to a constant so it is not recreated each time.

Also, I don't know how well this performs compared to `startsWith()`.

bolinfest: Also, if it matches `6to5` instead of `babel`, I think it should print out a deprecation warning. It's early enough that we can probably get everyone to convert their pragmas now if there's a warning that tells them to.

sebmck: I copy and pasted that line from [atom-shell-starter](https://github.com/atom/atom-shell-starter) so I have no idea how well it fairs.

kevinsawicki:
Thanks for fixing this up, this will be included in the next Atom release, 0.183

PR #5598, opened 2015-02-17
 fix deb  for debian and ubuntu 14.04TLS comfortability 

a-moses:
Add friendly  depend for  debian (testing/unstable versions)  and ubuntu 14.04TLS


50Wliu:
There don't seem to be any changes in this PR.

a-moses:
some mistake with github.com sorry . 

the new one is pull

PR #5605, opened 2015-02-18
fix .deb  file for debian and ubuntu 14.04TLS comfortability 

a-moses:
fix .deb file  for Debian (SID/JESSY versions)  and Ubuntu 14.04TLS comfortability


kevinsawicki:
Cool, thanks for fixing this.

Issue #5569, opened 2015-02-15
Docs task unclear


silvenon:
I would like to create a [Dash](http://kapeli.com/dash) docset for Atom, but I can't seem to build the docs. I tried running `grunt docs` inside the `build` directory, but it can't find `markdown:guides` task. Running only `grunt build-docs` does generate something, `docs/output/api.json`, but I don't know where to go from here.

izuzak:
> I tried running grunt docs inside the build directory, but it can't find markdown:guides task. 

I think `markdown:guides` should have been deleted as a part of https://github.com/atom/atom/commit/db13a40f0a5004b1dcec1734ee907ea6d25a70ce /cc @kevinsawicki.

> Running only grunt build-docs does generate something, docs/output/api.json, but I don't know where to go from here.

The `api.json` file is what you get (it contains descriptions of classes and methods in Atom) -- we use that on [atom.io to render the API docs pages](https://atom.io/docs/api/latest/Atom). That file is also a part of Atom's [releases](https://github.com/atom/atom/releases). Just to make sure I understand -- can you clarify what exactly you expected to see instead?

silvenon:
> Just to make sure I understand -- can you clarify what exactly you expected to see instead?

I was hoping to be able to run the step which converts the JSON file into the API docs website (HTML/CSS/JS), so I can use it for Dash. Is that step open sourced too?

izuzak:
> I was hoping to be able to run the step which converts the JSON file into the API docs website (HTML/CSS/JS), so I can use it for Dash. Is that step open sourced too?

As far as I know, there is no such step currently -- we don't generate the API docs as a static site. /cc @thedaniel

silvenon:
Oh, you generate it on the fly with JavaScript or something? Anyway, I'm closing this then.

thedaniel:
@silvenon we import it into a db and index it in elasticsearch and render the markdown "on the fly" (well, once, then it's heavily cached) - makes it easier to have good search / more sophisticated indexing / doc versioning. It would be nice to generate a basic static HTML set with grunt though.

silvenon:
Cool, thanks for the explanation :smiley:

PR #5614, opened 2015-02-18
Remove markdown:guides task which doesn't exist anymore

izuzak:
I think this should have been removed as a part of db13a40. See https://github.com/atom/atom/issues/5569 for more context.

cc @kevinsawicki


kevinsawicki:
:+1:

PR #5623, opened 2015-02-18
Add "Checking for Update" to menu

brendannee:
When state is “checking”, menu should say “Checking for Update” instead of "Downloading Update".  This solves the issue of a slow network connection where it incorrectly shows "Downloading Update" even before it has been determined if an update is available.


50Wliu:
:+1:

kevinsawicki:
Awesome, thanks for this.

PR #5624, opened 2015-02-18
Implement text editor DOM updates manually instead of via React

nathansobo:
React is an amazing abstraction, but very few abstractions come without at least some overhead. In the case of Atom's text editor, it's worth the effort to avoid this overhead by hand coding all DOM updates. The update code is repetitive, but it's also pretty straightforward. In exchange, we get much simpler profile traces, drop a big dependency, and reduce the overall number of moving parts in Atom. Below are some flame graphs for entering a character at the end of a syntax-highlighted line. The comparison is somewhat unfair because the React case _could_ have been a bit more optimized, but I think the pictures also show how much simpler the manual approach is in our case.

screenshot_2015-02-18_15_03_39

screenshot_2015-02-18_15_01_42

React is a great tool for many cases, for our particular needs in this particular case I decided it would be easier to just do things for ourselves.
### Remaining Tasks:
- [x] Don't use React for EditorComponent
- [x] Fix autocomplete spec.
- [x] Unify DOM updates / polling across editors.


benogle:
Legit

nathansobo:
I'm going to keep the basic structure of the code with the `TextEditorComponent` separate from the `TextEditorElement` the same for now. It might be nice to clean it up but it's not really a huge deal and I think time could be spent better in other ways. Someday I would love to get to a more minimal DOM structure as well, but it doesn't seem worth the risk of breaking anything to lose a couple extra nodes.

nathansobo:
@atom/core @atom/non-github-maintainers Could some people build this branch and try working with it tomorrow. I need some more :eyes: looking for any regressions, and as a reward you'll see speed improvements.

a-moses:
@nathansobo  :+1: 

React is perfect for very spasific "component" or pakages  only and only with Flux Application Architecture (or similar to Flux) .

for example : facebook chat and so on.
and some component for  fast feedback to the user .

react is not good for all dom managment or for all "staff" in atom.

So hybrid application is the answer .

i think  :)

lee-dohm:
@nathansobo I've been working with it for the past hour. I haven't noticed any issues. But it does seem a bit more responsive :grinning:

thomasjo:
Been using this all day and haven't noticed any regressions. The constant console logging is very annoying though, hehe :angel:

thomasjo:
I've not really noticed any major performance issues in the past few months, so hard to tell whether the speedup is noticeable or not on my machine, but things do seem a bit more responsive. Could be confirmation bias though.

postcasio:
Same here. No issues so far, playing well with my other packages. Scrolling is silky smooth!

as-cii:
Same here, it has been working fine so far. Great job, @nathansobo! :+1:

nathansobo:
Thanks to all of you for your time! @thomasjo thanks for pointing out the console logging. I accidentally left it in on the last commit and just force-pushed an amendment.

kevinsawicki:
:chart_with_upwards_trend: :green_heart: :recycle:

lee-dohm:
:clap: Awesome work!

nathansobo:
I force pushed back the merge of this PR so we can get a release out this morning without me having to worry about this code over the weekend. Will merge it again once the release is out.

src/view-registry.coffee
```diff
@@ -150,3 +158,52 @@ class ViewRegistry
...
+
+  stopPollingDocument: ->
+    window.clearInterval(@pollIntervalHandle)
+
+  performDocumentPoll: =>
+    if @documentUpdateRequested
+      @performDocumentPollAfterUpdate = true
+    else
+      @performDocumentPollAfterUpdate = false
+      poller() for poller in @documentPollers
```
kevinsawicki: Might want to `return` after this line to prevent array return.

nathansobo: :+1: *Very* good catch.

nathansobo:
Okay, merged in 964809373bbf6e621a1b307c4fae372b8fe0e525. Sorry for the false :rotating_light:.

hemanth:
Interesting stats, but was curious about why was react opted before...

nathansobo:
@hemanth There was also a huge improvement over the first version of the editor when we added React. It was much like scaffolding.

hemanth:
:+1:

mcharytoniuk:
Why didn't you use React's PureRenderMixin (http://facebook.github.io/react/docs/pure-render-mixin.html), batched updates and other optimisations that would produce better results than manual updates and abandoned the entire solution instead?

kl3ryk:
@mcharytoniuk +1

benogle:
@mcharytoniuk The old implementation was doing the same thing as described in the doc page:

> Under the hood, the mixin implements shouldComponentUpdate, in which it compares the current props and state with the next ones and returns false if the equalities pass.

See https://github.com/atom/atom/blob/97a671cb496f288ba30703af4800a12551e99d50/src/lines-component.coffee#L74-L92

mcharytoniuk:
I know that maybe it's too late for such ideas but I'd try moving away at least cursor component out of lines component and position it absolutely on top of it to prevent lines component updates caused by cursor movement. Also, Flipboard published react-canvas recently. Did you ever consider using canvas for animations / some editor elements as they did? Many people criticised them for negative SEO impact but inside text editor every crazy trick is acceptable imo. :)

nathansobo:
A big selling-point of Atom is the ability to style content with CSS. Used properly, the DOM doesn't introduce prohibitive overhead for our use case.

React being removed is purely an implementation detail in this case and shouldn't affect package authors in any way. Appreciate your concern and suggestions here, but on balance it's easier to achieve our goals without it in this case.

Issue #5642, opened 2015-02-19
Copy paste of text aligned with spaces (e.g. asterisk-aligned PHPDoc) sometimes loses the alignment with soft tabs off


driskell:
Easiest to reproduce with PHP grammar, and the "doc_f" snippet. Soft tabs needs to be off (it's fine with it on.)

Create a new PHP file and use doc_f snippet. You'll get a PHPDoc block which uses the recommended formatting - the asterisks are aligned vertically using an additional leading space.

When copy and pasting... this space is lost if you paste on an already indented line. ANIGIF below shows the reproduction, and shows that on an empty line it pastes OK, but on pasting with indentation the spaces aligning the asterisks are lost.

![example](https://cloud.githubusercontent.com/assets/939815/6273202/41e81dbe-b865-11e4-8ea5-a5354bdbb50c.gif)

In case someone thinks aligning the asterisks is bad practice - it's actually customary for PHPDoc. See first glossary entry on the PHPDoc website:
http://www.phpdoc.org/docs/latest/glossary.html

**Edit: Removed the root cause bit - I've put it in the PR**

Thanks!

driskell:
Any updates? Appreciate things are prioritised but I've not really heard anything in a while.

nathansobo:
Taking a look at this now.

nathansobo:
Hate to say this after all this time, but I'm having trouble reproducing this. @driskell is this still happening? Am I doing something wrong in the animation below?

![php-test](https://cloud.githubusercontent.com/assets/1789/8414982/440ff28a-1e61-11e5-96a2-d4db55a68152.gif)

nathansobo:
Nevermind, I see what I was doing wrong now. Going to delete the above comment to reduce noise.

nathansobo:
Closed via #7563

PR #5658, opened 2015-02-20
:bug: Do not modify excess whitespace when modifying indentation

driskell:
This is a proposal for fixing #5642 and I'd welcome feedback on if this is the right approach or not?

The problem, is copying pasting PHPDoc code at an indented level, with hard tabs on, loses the asterisk alignment:

example

**Root Cause**

Seems inside Atom, in the case of encountering soft tabs in a hard tab file... it can return a _floating point_ indentation level, such as 1.5 in this instance (tab+half-tab). When rebuilding the indentation (it calls setIndentationForBufferRow for each line), this calls @buildIndentString with 1.5 which gets converted to integer: 1. ALL leading whitespace is then replaced by this new indentation - so we lose the space that aligns the asterisk.

The tests seem to point to floating point indentation levels being allowed. Yet I can find nowhere where it is useful. Most places just Math.ceil(). And with hard tabs it doesn't even make sense.

This patch attempts to remove all the floating points (it leaves the Math.ceil calls - they're harmless and can be cleaned later if this turns out the right approach.). It then has a function for "modifying" whitespace to change its indentation level, while paying attention to straggling spacing like that in the PHPDoc comments in the image above. It also adds a couple API for handling indentation changes that can be used elsewhere.

Only thing I'm not too happy with is the overlap between my "setIndentLevelForRow" and the existing "setIndentationForBufferRow" - I'm sure mine follows correct naming convention but existing doesn't. It is the existing one that is called when pasting though.

### DIFF
--- a/spec/tokenized-buffer-spec.coffee
+++ b/spec/tokenized-buffer-spec.coffee
@@ -722,7 +722,7 @@ describe "TokenizedBuffer", ->
         expect(tokenizedBuffer.tokenizedLineForRow(1).indentLevel).toBe 1
         expect(tokenizedBuffer.tokenizedLineForRow(2).indentLevel).toBe 2
         buffer.insert([2, 0], ' ')
-        expect(tokenizedBuffer.tokenizedLineForRow(2).indentLevel).toBe 2.5
+        expect(tokenizedBuffer.tokenizedLineForRow(2).indentLevel).toBe 2
 
     describe "when the line is empty", ->
       it "assumes the indentation level of the first non-empty line below or above if one exists", ->
@@ -900,3 +900,35 @@ describe "TokenizedBuffer", ->
       expect(tokenizedBuffer.tokenizedLineForRow(1).tokens[0].value).toBe 'b'
       expect(tokenizedBuffer.tokenizedLineForRow(2).tokens.length).toBe 1
       expect(tokenizedBuffer.tokenizedLineForRow(2).tokens[0].value).toBe 'c'
+
+  describe ".parseIndentation(leadingWhitespace)", ->
+    it "parses the indentation correctly", ->
+      buffer = atom.project.bufferForPathSync('sample-with-tabs-and-initial-comment.js')
+      tokenizedBuffer = new TokenizedBuffer({buffer})
+      fullyTokenize(tokenizedBuffer)
+
+      parseIndentation = (bufferRow) ->
+        leadingWhitespace = tokenizedBuffer.tokenizedLineForRow(bufferRow).text.match(/^[\t ]*/)[0]
+        tokenizedBuffer.parseIndentation(leadingWhitespace)
+
+      parsed = parseIndentation(0)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 0
+      parsed = parseIndentation(1)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 1
+      parsed = parseIndentation(2)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 1
+      parsed = parseIndentation(3)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 1
+      parsed = parseIndentation(4)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 0
+      parsed = parseIndentation(5)
+      expect(parsed.indentLevel).toBe 0
+      expect(parsed.excessSpace).toBe 0
+      parsed = parseIndentation(6)
+      expect(parsed.indentLevel).toBe 1
+      expect(parsed.excessSpace).toBe 0
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -2246,6 +2246,53 @@ class TextEditor extends Model
     return unless @getSoftTabs()
     @scanInBufferRange /\t/g, bufferRange, ({replace}) => replace(@getTabText())
 
+  # Essential: Set the indentation level for the given buffer row.
+  #
+  # Rebuilds the indentation based on the soft tabs and tab length settings of
+  # this editor in order to bring it to the given indentation level.
+  # Note that if soft tabs are enabled and the tab length is 2, a row with 4
+  # leading spaces would have an indentation level of 2.
+  #
+  # * `bufferRow` A {Number} indicating the buffer row.
+  # * `newLevel` A {Number} indicating the new indentation level.
+  setIndentLevelForRow: (bufferRow, newLevel) ->
+    @modifyIndentLevelForRow(bufferRow, newLevel)
+
+  # Essential: Adjust the indentation level for the given buffer row.
+  #
+  # Rebuilds the indentation based on the soft tabs and tab length settings of
+  # this editor in order to increase or decrease it by the requested number of
+  # levels.
+  # Note that if soft tabs are enabled and the tab length is 2, a row with 4
+  # leading spaces would have an indentation level of 2.
+  #
+  # * `bufferRow` A {Number} indicating the buffer row.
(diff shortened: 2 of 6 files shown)


driskell:
Edited to try reduce my usual walls of text :(

driskell:
Hi @50Wliu I raise this as a proposed fix for #5642 - thanks.

50Wliu:
@kevinsawicki @benogle @nathansobo?

driskell:
Rebased onto master.

Please let me know if there is anything I can do to help with this.

src/tokenized-buffer.coffee
```diff
@@ -349,13 +349,23 @@ class TokenizedBuffer extends Model
...
+      {indentLevel} = @parseIndentation(match[0])
+      indentLevel
     else
       0
 
+  parseIndentation: (leadingWhitespace) ->
+    tabLength = @getTabLength()
+    columnCount = 0
+    for c in leadingWhitespace
+      if c == '\t'
```
izuzak: Recommendation: use `is` instead of `==`. See [this section](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#coffeescript-styleguide) of the contributing guide.

src/text-editor.coffee
```diff
@@ -2236,6 +2236,60 @@ class TextEditor extends Model
...
+    {indentLevel, excessSpace} = @displayBuffer.parseIndentation(leadingWhitespace)
+
+    if newLevel?
+      newIndentString = @buildIndentString(newLevel + indentIncrease)
+    else
+      newIndentString = @buildIndentString(indentLevel + indentIncrease)
+
+    # Handle case where spaces were used for alignment (such as /* */ blocks)
+    # immediately after the indentation
+    if excessSpace == 0
```
izuzak: Recommendation: use `is` instead of `==`. See [this section](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#coffeescript-styleguide) of the contributing guide.

src/text-editor.coffee
```diff
@@ -2236,6 +2236,60 @@ class TextEditor extends Model
...
+      newIndentString
+    else
+      newIndentString + _.multiplyString(' ', excessSpace)
```
izuzak: Wondering if we can replace

```
if excessSpace == 0
  newIndentString
else
  newIndentString + _.multiplyString(' ', excessSpace)
```

with just

```
newIndentString + _.multiplyString(' ', excessSpace)
```

If `excessSpace` is `0`, then `newIndentString + _.multiplyString(' ', excessSpace)` becomes `newIndentString + ""`, which is what we want.

src/text-editor.coffee
```diff
@@ -2236,6 +2236,60 @@ class TextEditor extends Model
...
+  # * `adjustLevel` A {Number} indicating how many levels to increase by. A
+  #   negative value will remove levels.
+  adjustIndentLevelForRow: (bufferRow, adjustLevel) ->
+    @modifyIndentLevelForRow(bufferRow, undefined, adjustLevel)
+
+  # Helper for modifying indentation for a buffer row
+  modifyIndentLevelForRow: (bufferRow, newLevel=undefined, indentIncrease=0) ->
+    leadingWhitespace = @lineTextForBufferRow(bufferRow).match(/^\s*/)[0]
+    newIndentString = @modifyIndentWhitespace(leadingWhitespace, newLevel, indentIncrease)
+    @buffer.setTextInRange([[bufferRow, 0], [bufferRow, leadingWhitespace.length]], newIndentString)
```
izuzak: Wondering if we could replace this block above :point_up: 

```
leadingWhitespace = @lineTextForBufferRow(bufferRow).match(/^\s*/)[0]
newIndentString = @modifyIndentWhitespace(leadingWhitespace, newLevel, indentIncrease)
@buffer.setTextInRange([[bufferRow, 0], [bufferRow, leadingWhitespace.length]], newIndentString)
```

with a call to `modifyIndentLevelForLine`. Perhaps with something like this?

```
modifyIndentLevelForLine(@lineTextForBufferRow(bufferRow), newLevel, indentIncrease)
```

spec/tokenized-buffer-spec.coffee
```diff
@@ -900,3 +900,17 @@ describe "TokenizedBuffer", ->
       expect(tokenizedBuffer.tokenizedLineForRow(1).tokens[0].value).toBe 'b'
       expect(tokenizedBuffer.tokenizedLineForRow(2).tokens.length).toBe 1
       expect(tokenizedBuffer.tokenizedLineForRow(2).tokens[0].value).toBe 'c'
+
+  describe ".parseIndentation(leadingWhitespace)", ->
```
izuzak: `parseIndentation` returns [two parameters](https://github.com/atom/atom/pull/5658/files#diff-dc7bdf9554a6b952bca110e83059dae8R367), `indentLevel` and `excessSpace`. Do you think it would make sense to extend these specs so that both returned parameters are covered (currently, `excessSpace` isn't, as far as I can tell)? That would prevent future :octocat: from breaking the behavior of `parseIndentation` which you're introducing in this PR.

izuzak:
@driskell Thanks for submitting this PR. :bow: I don't review PRs normally, but wanted to give it a shot here -- a :school: experience for me. :smile: 

I think @nathansobo will have more :thought_balloon: and :bulb:, but just wanted to say that I think these changes make sense overall.

driskell:
Thanks @izuzak ! Appreciate the feedback and review!

src/tokenized-buffer.coffee
```diff
@@ -349,13 +349,23 @@ class TokenizedBuffer extends Model
 
   indentLevelForLine: (line) ->
     if match = line.match(/^[\t ]+/)
-      leadingWhitespace = match[0]
-      tabCount = leadingWhitespace.match(/\t/g)?.length ? 0
-      spaceCount = leadingWhitespace.match(/[ ]/g)?.length ? 0
-      tabCount + (spaceCount / @getTabLength())
```
izuzak: I'm a bit surprised at how this method was counting tabs and spaces before -- just counting the number of spaces and tabs, regardless of where they're located. I think your approach in `parseIndentation` is much better. :+1:

src/text-editor.coffee
driskell: @izuzak I remember now why I did this :)

It was the keep the buffer change small, so the "patch" generated by @buffer.setTextInRange is kept small (not sure if would be a performance thing - though it means it benefits from future enhancements I think). Also, unless there's a way to set a line of text (I hadn't seen one) we'd still be calling @buffer.setTextInRange() anyway but with the whole line. So I felt may as well do the minimum change.

src/tokenized-buffer.coffee
driskell: Yes it was surprising! I hope my method is better.

It also follows tab stops if there's spaces intermingled. I know mixed tab/space is REALLY bad anyway but it hopefully means indentLevelForLine returns nice correct numbers that match exactly the visual indentation we see, where as at moment it would be weird. For example "S\tS\t" would return 3 where visually it would actually be identical to "\t\t" which is 2 (assuming 2 space tab)

spec/tokenized-buffer-spec.coffee
driskell: This is now done! Just wrapped a helper and called parseIndentation directly.

driskell:
Again, thanks for the review @izuzak - I'll followed through and provided more info on the @modifyIndentLevelForRow logic.

driskell:
Not sure who to check with but wondering where this sits in the pipeline, if at all?

izuzak:
Sorry for the delay, @driskell. I believe @nathansobo wanted to look at this when he finds some time. :pager:

nathansobo:
Thanks for taking the time to do this. It served as a jumping off point for a different fix in #7563, which reduces some of the complexity that led to this issue in the first place. Please let me know if there's something you think I missed with that fix.

PR #5660, opened 2015-02-21
Make sure that GitRepositoryProvider.repositoryForDirectorySync() return...

bolinfest:
...s null rather than throws.

The UI locks up if this method does not return.

### DIFF
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -61,7 +61,7 @@ describe "GitRepositoryProvider", ->
       beforeEach ->
         provider = new GitRepositoryProvider atom.project
 
-        # Tolerate an implementation of Directory that does not implement existsSync().
+        # An implementation of Directory that does not implement existsSync().
         subdirectory = {}
         directory =
           getSubdirectory: ->
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -55,14 +55,15 @@ describe "GitRepositoryProvider", ->
           provider.repositoryForDirectory(directory).then (result) ->
             expect(result).toBe null
 
-    describe "when specified a Directory that throws an exception when explored", ->
-      it "catches the exception and returns null for the sync implementation", ->
+    describe "when specified a Directory without existsSync()", ->
+      it "returns null", ->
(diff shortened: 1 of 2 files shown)


bolinfest:
Unit test forthcoming.

bolinfest:
Unit test provided. There's a `TODO` about logging the error. I'm not sure what's standard practice in Atom for something like this.

spec/git-repository-provider-spec.coffee
```diff
@@ -54,3 +54,35 @@ describe "GitRepositoryProvider", ->
           directory = new Directory dirPath
           provider.repositoryForDirectory(directory).then (result) ->
             expect(result).toBe null
+
+    describe "when specified a Directory without existsSync()", ->
+      it "returns null", ->
+        provider = new GitRepositoryProvider atom.project
+
+        # Tolerate an implementation of Directory that does not implement existsSync().
```
maxbrunsfeld: Could you put the code that sets up `directory` into a `beforeEach`. My criteria for extracting something like that is roughly:
* It's the same for all of the tests in the `describe` block.
* It corresponds directly to the description of the `describe` block.

src/git-repository-provider.coffee
```diff
@@ -26,9 +26,9 @@ isValidGitDirectorySync = (directory) ->
   # To decide whether a directory has a valid .git folder, we use
   # the heuristic adopted by the valid_repository_path() function defined in
   # node_modules/git-utils/deps/libgit2/src/repository.c.
-  return directory.getSubdirectory('objects').existsSync() and
-      directory.getFile('HEAD').existsSync() and
-      directory.getSubdirectory('refs').existsSync()
+  return directory.getSubdirectory('objects').existsSync?() and
```
maxbrunsfeld: I would say leave `isValidGitDirectorySync` as it was; we only need to check once if the `Directory` (and the directories it returns) implements synchronous IO; no need to check each call.

spec/git-repository-provider-spec.coffee
bolinfest: Done.

src/git-repository-provider.coffee
bolinfest: Done.

Issue #5163, opened 2015-01-20
Optimize typing, cursor movement, and scrolling performance


nathansobo:

YuriSolovyov:
possibly related to scrolling: #3154

gregersrygg:
Good to see you're planning on improving the typing performance :) It can take many seconds to type 1 character when using a few hundred cursors. Not sure you need it, but here is a CPU-profile I saved while typing. The first plateau (in the profiler chart view) is when I pressed <kbd>⌘</kbd><kbd>⇧</kbd><kbd>L</kbd> to convert the selection into multiple cursors. The rest of the plateaus are typing.

[multiple-cursors-typing.cpuprofile](https://www.dropbox.com/s/4pw45ii6gxtai59/multiple-cursors-typing.cpuprofile?dl=0)

![screen shot 2015-02-18 at 19 40 06](https://cloud.githubusercontent.com/assets/149725/6254056/0553f720-b7a6-11e4-9ca5-e2dd899dcc32.png)

To me it looks like it's in the `reducer` inside `TextEditor.prototype.mergeIntersectingSelections` most of the time is spent (actually inside underscore-methods called from `reducer`).

If `mergeIntersectingSelections` does what it sounds like it does. Then why does it need to be called when I'm typing without any selection?

nathansobo:
@gregersrygg It probably doesn't. Would you be interested in playing around with some optimizations? Sounds like some quick wins are available here.

gregersrygg:
I really wish I had time to contribute, but I don't :(

nathansobo:
@gregersrygg That's okay, just investigating things is helpful. One big issue is probably that merging intersection selections is N^2, which doesn't having the greatest scaling properties. We do need to merge them because if cursors overlap they need to be merged.

benogle:
@nathansobo can this be closed for now? I assume the rest of the optimizations will come from the work to fix #307

nathansobo:
It really wasn't complete, but we can reopen it later if you want.

nathansobo:
Wait a minute. The story name in tracker didn't match. This is fine.

Issue #5025, opened 2015-01-13
Multi-line editing becomes unusable with large selections


Anteru:
I've just tried to multi-line edit a file with 400 lines produced by `find .` All of lines the started with ./, so I selected the first ./, kept on pressing CTRL+D to capture them all. Around line 250-300, the editor would slow down a bit, but the selection finished quickly. Then I started typing a few characters, and the first took several seconds to appear, while the second didn't appear even after 30 seconds. That makes multi-line editing pretty much useless.

Anteru:
Probably related to issue #2078, if not the same.

as-cii:
Bringing the discussion from #5678 (that's a nice number! :wink:)

> Fantastic work here.

@nathansobo: first off, let me thank you for your kind words. They're part of the reason why I enjoy contributing to open source: they really make my day.

> If you're interested in optimizing this further, I have some ideas.

Definitely :smile: Before your reply, I was collecting some benchmarks to check where the time was being spent while moving multiple cursors or typing a character on many lines. It turned out that it's exactly as you describe:

> Unfortunately, when there are ton of little changes before the next display update, we end up paying more for each incremental update than we would pay if we waited and just recomputed everything at the end.

In facts, the performance bottleneck arises because we recompute everything for each change, thus wasting quite a lot of time in performing over and over the same operation (which, as you point out, can be executed at the end of a change). Your redesign of `TextEditorPresenter` was essential to address such bottlenecks, therefore thank you for this :bow:

> I see two paths of attack on this.
1) We could do a better job on the incremental updates. Some of them are way more wasteful than they need to be in that they update every line when one line is updated, etc.
2) We could apply a heuristic for when we think a bunch of incremental changes might be applied. For example, we could handle 1 cursor movement or text change incrementally, but if we get a second in either category we could flip into batch mode where we defer any further updates until it's time to render to the screen, then recompute everything async.

Either choices will definitely speed things up, but I believe the first path would be more beneficial than the second. Indeed, the latter would still be a coarse grained way of handling updates, whereas I am convinced we need to be very strict on what changes we apply and when to apply them. I made some experiments with heuristics but they do not feel quite right (moreover, they need *always* some kind of threshold/configuration, which I'd prefer to skip wherever possible).

According to the first path, I believe that the key idea is to emit some more detailed events from the model. We could do this using two techniques:

1. Emit the same events with further informations (e.g. which lines changed exactly, and perform updates only on a certain range of lines)

[...]

PR #5678, opened 2015-02-22
Faster `mergeIntersectingSelections`

as-cii:
This pull-request introduces a performance improvement in `mergeIntersectingSelection`. Before changing it the cost of this function was `O(n^2)`: selections were accessed always and in sparse order, causing the traversing time to increase exponentially as more and more selections were added. 

Before merging selections, we now order them first and try to intersect the current selection with the previous one only. This brings the cost of `mergeIntersectingSelections` down to `O(n log n)`, changing this:

screen shot 2015-02-22 at 14 01 32

To this:

screen shot 2015-02-22 at 14 00 31

_(Benchmarks based on splitting into lines the entire selection of a 400 lines file)_

All the specs are green, therefore this "new algorithm" shouldn't have changed the behavior of `mergeIntersectingSelections`. It would be great to have some more :eyes: to detect any possible regression.

Hopefully, this should be a significant step forward towards #5163 :smile:

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -2206,11 +2206,18 @@ class TextEditor extends Model
       result = fn()
       @suppressSelectionMerging = false
 
+    findIntersectingSelection = (elements, selection) ->
+      return if elements.length == 0
+
+      otherSelection = _.last(elements)
+      exclusive = not selection.isEmpty() and not otherSelection.isEmpty()
+      intersects = otherSelection.intersectsWith(selection, exclusive)
+
+      return otherSelection if intersects
+
+
     reducer = (disjointSelections, selection) ->
-      intersectingSelection = _.find disjointSelections, (otherSelection) ->
-        exclusive = not selection.isEmpty() and not otherSelection.isEmpty()
-        intersects = otherSelection.intersectsWith(selection, exclusive)
-        intersects
+      intersectingSelection = findIntersectingSelection(disjointSelections, selection)
 
       if intersectingSelection?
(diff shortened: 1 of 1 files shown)


as-cii:
/cc: @nathansobo @gregersrygg

(as you were discussing about this in #5163 and I thought it might have been relevant to you :wink:)

nathansobo:
@as-cii Fantastic work here. 

If you're interested in optimizing this further, I have some ideas. I recently introduced the `TextEditorPresenter` layer to synchronously and incrementally maintain a state object based on model changes. The intuition on this was that for small changes like a single cursor movement or a single character insertion, recomputing the entire presenter state was wasteful and it would be better to make a small change. Unfortunately, when there are ton of little changes before the next display update, we end up paying more for each incremental update than we would pay if we waited and just recomputed everything at the end.

I see two paths of attack on this.

1) We could do a better job on the incremental updates. Some of them are way more wasteful than they need to be in that they update every line when one line updated, etc.

2) We could apply a heuristic for when we think a bunch of incremental changes might be applied. For example, we could handle 1 cursor movement or text change incrementally, but if we get a second in either category we could flip into batch mode where we defer any further updates until it's time to render to the screen, then recompute everything async.

Not sure if you're interested in investigating this but since you're already doing a great job in other areas, I thought I'd suggest it.

nathansobo:
Refs https://github.com/atom/atom/issues/5025.

Any further discussion should probably actually take place there.

gregersrygg:
@as-cii Awesome speed improvement :zap: :smile:

Issue #5492, opened 2015-02-11
Can't update or install any packages without root permission


sedabull:
* Ubuntu 14.04
* Atom 0.178.0
* apm 0.136.0

It is impossible for me to update or install any packages from the atom ui. I get an error message like this:
![atom-update-error](https://cloud.githubusercontent.com/assets/9311295/6141355/2a450964-b172-11e4-9780-d67ed56741e5.png)

apm gives the same error on the command line, unless sudo is used. In which case it works fine. On a possibly related note, packages linked via `apm link --dev` show up in the installed packages view like they should, but are not activateable.

gau1991:
Same here on Fedora

izuzak:
This looks very similar to https://discuss.atom.io/t/solved-new-package-installs-updates-generates-errors/13152 -- can you check that discussion and see if the advice there helps?

GeorgeSapkin:
Removing .atom directory didn't work for me on Fedora 21 x64.

gau1991:
This discussion not worked for me...

mehcode:
Something similar occurs for me on update that didn't occur on the last version.
![screenshot from 2015-02-11 06 28 21](https://cloud.githubusercontent.com/assets/753919/6148528/304b7ede-b1b7-11e4-8c7c-393338be7ddc.png)

joefitzgerald:
See also: https://github.com/atom-community/autocomplete-plus/issues/260 @nathansobo @kevinsawicki @thedaniel @maxbrunsfeld - v0.178.0 seems to be broken - at least on Linux.

Works for me on OS X though:

![screen shot 2015-02-11 at 8 03 23 am](https://cloud.githubusercontent.com/assets/744740/6149233/7f3708b2-b1c4-11e4-81c4-65428fa8a1e4.png)

![screen shot 2015-02-11 at 8 02 31 am](https://cloud.githubusercontent.com/assets/744740/6149234/7f3927a0-b1c4-11e4-8db1-487239b7c08d.png)

![screen shot 2015-02-11 at 8 02 38 am](https://cloud.githubusercontent.com/assets/744740/6149235/7f39363c-b1c4-11e4-8040-31bbdd0c7363.png)

edgard:
Same here, maybe it has something todo with pr #5385?

I tried setting ATOM_HOME but still doesn't work.

mnquintana:
Same issue here on Ubuntu 14.04 after updating to Atom 0.178.0.

[...]

PR #5701, opened 2015-02-23
fix #5602

### LINKED ISSUE #5602: --dev linked packages are undefined unless in --dev mode
Atom v-0.180.0
Ubuntu 14.04

See the last half of the discussion at #5492. Here are the highlights:
* git clone an atom package locally
* run apm install to bring in node_modules
* run apm link --dev
* open atom

In the settings view, you will see the package under the `development packages` section. However, you cannot adjust the settings, none of the activation events are available, and running `atom.packages.resolvePackagePath('my-package')` in the dev tools returns `undefined`. Instead a prompt to install the package is displayed (as in the install packages view). Running atom in dev mode behaves as expected. Here is a relevant screen shot on atom v-0.180.0:
atom-dev-error

cc @kevinsawicki @thedaniel

hoosierhobbyist:
Issue #5602 is that packages that are `apm link --dev`ed are undefined and unusable in a normal `atom` session. While reading through the `PackageManager` source today, I noticed that the path `~/.atom/dev/packages` is only added to the package search path when in `devMode`. This tiny fix should resolve that issue.

### DIFF
--- a/src/package-manager.coffee
+++ b/src/package-manager.coffee
@@ -34,8 +34,7 @@ class PackageManager
     @emitter = new Emitter
     @packageDirPaths = []
     unless safeMode
-      if @devMode
-        @packageDirPaths.push(path.join(configDirPath, "dev", "packages"))
+      @packageDirPaths.push(path.join(configDirPath, "dev", "packages"))
       @packageDirPaths.push(path.join(configDirPath, "packages"))
 
     @loadedPackages = {}


lee-dohm:
The fact that packages in `~/.atom/dev/packages` are only loaded in Dev Mode is an intentional feature, not a bug. It is made this way so that package developers can test out new versions of packages while keeping the production version of the same packages installed.

sedabull:
@lee-dohm, I had thought it might be but when I asked in my final comment in #5492, @thedaniel advised me to go ahead and open issue #5602, which was marked as a bug by @50Wliu

lee-dohm:
I haven't read through all of #5492, which is why I didn't close either of these out of hand. But I'm pretty certain that this is not the right solution for the issue.

thedaniel:
I think that it's a bug in settings-view and the UX around development/disabled packages - the point of dev linked packages is to load them in dev mode only. Sorry about the confusion!

lee-dohm:
Thanks @thedaniel!

sedabull:
@thedaniel, that's quite alright! I'll open up the issue over at atom/settings-view.

PR #5707, opened 2015-02-24
Report deprecation warning when 'use 6to5' is used instead of 'use babel'.

bolinfest:
In Deprecation Cop, the warning is associated with `atom core` rather than the package that loaded the file with the old pragma.

Also, even though the deprecation message is parameterized by `filePath`, multiple deprecation warnings about the 6to5 pragma appear to be coalesced into a single warning under `atom core`.

### DIFF
--- a/spec/babel-spec.coffee
+++ b/spec/babel-spec.coffee
@@ -40,7 +40,7 @@ describe "Babel transpiler support", ->
       expect(grim.getDeprecationsLength()).toBe 0
 
   describe "when a .js file starts with 'use 6to5';", ->
-    it "transpiles it using babel and adds a deprecation", ->
+    it "transpiles it using babel and adds a pragma deprecation", ->
       expect(grim.getDeprecationsLength()).toBe 0
       transpiled = require('./fixtures/babel/6to5-single-quotes.js')
       expect(transpiled(3)).toBe 4
@@ -53,7 +53,7 @@ describe "Babel transpiler support", ->
       expect(grim.getDeprecationsLength()).toBe 0
 
   describe 'when a .js file starts with "use 6to5";', ->
-    it "transpiles it using babel", ->
+    it "transpiles it using babel and adds a pragma deprecation", ->
       expect(grim.getDeprecationsLength()).toBe 0
       transpiled = require('./fixtures/babel/6to5-double-quotes.js')
       expect(transpiled(3)).toBe 4
--- a/spec/babel-spec.coffee
+++ b/spec/babel-spec.coffee
@@ -1,7 +1,14 @@
 babel = require '../src/babel'
 crypto = require 'crypto'
+grim = require 'grim'
 
 describe "Babel transpiler support", ->
+  beforeEach ->
+    jasmine.snapshotDeprecations()
+
+  afterEach ->
+    jasmine.restoreDeprecationsSnapshot()
+
   describe "::createBabelVersionAndOptionsDigest", ->
     it "returns a digest for the library version and specified options", ->
       defaultOptions =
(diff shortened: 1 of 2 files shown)


sebmck:
Looks good to me, I'm keen to clean up the 6to5 name legacy :stuck_out_tongue_winking_eye:

kevinsawicki:
> In Deprecation Cop, the warning is associated with atom core rather than the package that loaded the file with the old pragma.

Yeah, would be great to associate it with the package, this can be done by adding the `packageName` metadata to the deprecate call similar to here: https://github.com/atom/atom/blob/3e439f3c57ea69150b0173ec5175dbf11859a111/src/workspace.coffee#L518

bolinfest:
@kevinsawicki I tried adding a `atom.workspace.getCallingPackageName()` call in there, but it appears that `atom.workspace` is not available yet on startup. The `getCallingPackageName()` method relies on `atom.packages.getLoadedPackages()`, which I'm guessing does not have the package whose file is being loaded yet.

bolinfest:
@kevinsawicki I changed it to `atom.workspace?.getCallingPackageName()`, which helps.

However, even in the cases where `packageName` is not undefined, I only see one instance of the warning get logged in `atom core`. Is there any sort of "do not coalesce" option I can pass as part of the metadata? Presumably I only see it in `atom core` because the warning is only reported once per file:line:column instance?

bolinfest:
My suspicions appear to be correct:

https://github.com/atom/grim/blob/fc9442cb2033f82ca18d9d52a884237ce099142d/src/grim.coffee#L50

Is this something you would be willing to relax in Grim?

bolinfest:
Also, I don't see `packageName` anywhere in https://github.com/atom/grim when I grep for it. Is it only honored by Deprecation Cop?

kevinsawicki:
>  Is it only honored by Deprecation Cop?

Yeah, grim just has support for a arbitrary `metadata` object to associate with the deprecation

kevinsawicki:
Yeah, I think maybe the package path should just be parsed from the file path being required.

`atom.workspace.getCallingPackageName()` relies on creating an exception which is slow but needed when you don't have the path in question.

Since the path is known here, I'd say just parse the packge name from the path.

bolinfest:
@kevinsawicki It looks like `filePath` is the real path of the file being loaded (so it's not of the form `~/.atom/packages/somethingsomething`), so to figure out the package, I would have to look upwards from the `filePath` until I find a `package.json` file, which I would read synchronously, parse, and then get the package name from there.

I'm happy to do that, if that's what you want, but I just want to make sure there isn't an easier way.

kevinsawicki:
> so it's not of the form ~/.atom/packages/somethingsomething

Sorry if I'm missing something, but shouldn't installed packages still have file paths that originate from `~/.atom/packages`?

I think we can skip linked and core packages for something like this and just focus on installed packages.

bolinfest:
@kevinsawicki The packages that I am working on in development do not seem to work that way. When I `console.log(filePath)`, I see the real path printed out. It could be because I manually created a symlink to `~/.atom/packages` manually instead of using `apm link` (I didn't know about `apm link` originally).

kevinsawicki:
@bolinfest I looked into this a bit more and came up with something I think addresses the issue without having to read/parse package.json files.

Basically it just creates a manual stack trace since the file path and location is known and manually adds it to grim as a serialized deprecation.

Can you check if this properly correlates the deprecation with the right package for the packages you were previously testing? I tried it out in the wrap guide package and saw it properly associated:

screen shot 2015-02-25 at 9 30 12 am

kevinsawicki:
Thanks @bolinfest, this will be in 0.183, which should be out soon.

basarat:
Perhaps update the blog post : http://blog.atom.io/2015/02/04/built-in-6to5.html

Note: Since 6to5 was renamed we now support `'use babel'`

kevinsawicki:
> Perhaps update the blog post

:+1:  Thanks mentioning this, I've updated http://blog.atom.io/2015/02/04/built-in-6to5.html

PR #5714, opened 2015-02-24
Update needed cursors only

as-cii:
If one cursor changed, all of them were recomputed in `TextEditorPresenter`. With a small refactoring, we gain quite a lot of speed (noticeable especially on multiline selections when moving left and right).

As usual, a second pair of :eyes: is always welcome! Thanks :bow:


izuzak:
cc @nathansobo for :eyes:

nathansobo:
Excellent. There are surely more opportunities like this to be more precise in our incremental updates.

PR #5717, opened 2015-02-24
Remove unused variable

### LINKED ISSUE #5675: Single Equals in Conditional - likely :bug:
I ran into the some [odd code](https://github.com/atom/atom/blob/master/src/package.coffee#L526
) trying to fix a different bug in atom.

```coffeescript
  isCompatible: ->
    return @compatible if @compatible?

    if @path.indexOf(path.join(atom.packages.resourcePath, 'node_modules') + path.sep) is 0
      # Bundled packages are always considered compatible
      @compatible = true
    else if packageMain = @getMainModulePath()
      @incompatibleModules = @getIncompatibleNativeModules()
      @compatible = @incompatibleModules.length is 0
    else
      @compatible = true
```
I can't believe the ```else if``` does what was intended-  *packageMain* isn't used anywhere, and the expression evaluates to true when *@getMainModulePath* returns something truthy.

edit:
This sort of stuff is everywhere in atom...  please consider not doing this! In the one case, you must guess at the intent of the author by examining the entire context of this line.  In the other case, you just have to look at the line.  There is no good reason to use this sort of construct, ever.

Douglas Crockford says
[*Avoid forms that are difficult to distinguish from common errors...*](https://www.youtube.com/watch?v=taaEzHI9xyY#t=2420):

izuzak:
I guess this fixes https://github.com/atom/atom/issues/5675 since we're not going to stop using assignments in conditional tests (?).

/cc @nathansobo @kevinsawicki

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -523,7 +523,7 @@ class Package
     if @path.indexOf(path.join(atom.packages.resourcePath, 'node_modules') + path.sep) is 0
       # Bundled packages are always considered compatible
       @compatible = true
-    else if packageMain = @getMainModulePath()
+    else if @getMainModulePath()
       @incompatibleModules = @getIncompatibleNativeModules()
       @compatible = @incompatibleModules.length is 0
     else


kevinsawicki:
:ship:

PR #5720, opened 2015-02-24
Don't add project paths when opening a file in an existing window

### LINKED ISSUE #5705: Opening a file in an existing window shouldn't add a project path
Right now, if I run `atom some-directory` then `atom another-directory/some-file.txt`, the directory `another-directory` is added to the project as a new root directory. This is wrong.

maxbrunsfeld:
Fixes #5705.


kevinsawicki:
:heart_eyes:

PR #5729, opened 2015-02-25
Add note on "+1 comments" to CONTRIBUTING

thedaniel:
cc @atom/core 

In my heart I just want to say that "+1 comments will be deleted" but that's a strong hand. Let's start with a kind suggestion.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,10 @@ propose changes to this document in a pull request.
 * Perform a cursory search to see if a similar issue has already been submitted.
 * Please setup a [profile picture](https://help.github.com/articles/how-do-i-set-up-my-profile-picture)
   to make yourself recognizable and so we can all get to know each other better.
+* Issue comments should help move the issue closer to resolution. Comments that
+  consist only of `+1` or similar should be avoided - they make it hard to scan
+  the conversation to date and send notifications to everyone already on the
+  thread.
 
 ### Package Repositories
 


lee-dohm:
+1 :laughing:

Sorry, couldn't resist :wink:

thedaniel:
I should have expected that.

benogle:
Part of me kind of likes them now. It shows interest. It separates the popular or painful issues from the rest. I ignore then 100% of the time, but they are useful when going back to an issue to gauge popularity.

50Wliu:
`the conversation to date` sounds a bit weird to me.  Why not just `the conversation`?
Also, how about `...the conversation and also sends notifications...`?

50Wliu:
Adding on to @benogle, +1 comments *are* helpful to a new/empty issue, but once multiple people have started to +1, then it becomes useless.

benogle:
> then it becomes useless

Maybe after an insane amount. But difference between 1, 10, and 50-100 is significant.

lee-dohm:
I agree that +1s have a definite diminishing returns facet to them. Also, I feel like many +1s are not looking at the whole issue. There have been multiple occasions where I've had to point out that an issue has been solved when people just read the initial bug report, jump to the bottom and +1 it.

On the other hand, I think this is an opportunity for GitHub. I filed a feature request a while back for Starring Issues. This would:

1. Allow me to mark Issues that I want to remember, come back to or check up on (over and above the gigantic swath of Atom issues I look at in my notifications)
1. Allow me to register "me too" ... equating to a +1

This feature would remove the noise from the comment stream and still afford the function that @benogle is suggesting.

nathansobo:
I personally do not mind the :+1: comments, though I wouldn't mind having them aggregated into the body of the issue or something. I'm not subscribed on as many issues as you guys, but seeing a :+1: pop into my email for the one's I do watch can be a helpful gauge for me.

batjko:
Voting for issues can be a useful feature indeed. It's used to great success in Jira for example.
But it should really be a default feature on each comment, rather than adding a new one.

[Zenhub](https://www.zenhub.io/) adds a +1 box accordingly:
image

thedaniel:
> Voting for issues can be a useful feature indeed

I disagree. Everyone seems to agree that "design by committee" is bad, but still wants issue voting at the same time. People will come in with the expectation that the folks working on the software full time are beholden to the vote count, no matter what expectations are set. Would you put your software project on reddit and add whatever feature gets the most upvotes?

batjko:
There's really no need to go to the extreme.
Like @benogle said, it can be very useful to gauge the demand for an issue, without committing yourself to be "beholden to the vote count". More of an indicator for informational purposes than a mandate.

If it's implemented as a little +1 or :+1:  box in the corner, it also helps preventing this from becoming spam, while allowing you to ignore the vote count if you wish.

lee-dohm:
> People will come in with the expectation that the folks working on the software full time are beholden to the vote count, no matter what expectations are set.

People already come in with that expectation. Just look at the spate of anti-metrics-package bugs lately, even including ones that see that it has been asked, answered and closed ... but file a new bug anyway. Or the "Remove Trailing Whitespace shouldn't be the default" issue that has been closed for nearly a year now yet has been commented on as recently as two days ago. The "tyranny of the majority" is something that has to be dealt with on any open source project ... whether we have voting buttons or stars or just :+1: in the comments.

postcasio:
Worth pointing out that Google Code has issue starring, but Chromium *still* gets the occasional issue with hundreds of +1s comments.

benogle:
> Everyone seems to agree that "design by committee" is bad, but still wants issue voting at the same time.

I definitely dont think we should blindly make something in a super popular issue, but it is a nice indicator for pain people are having.

mark-hahn:
If the majority ruled we'd still have slaves in the US.  Or at least for
many more years.

It's obvious that instead of majority rule everyone should just do what I
want.

kevinsawicki:
Closing this out, I think we've found a nice middle ground of locking issues when appropriate and this doesn't need to be explicit in the docs.

PR #5744, opened 2015-02-25
Make package activation function optional

kevinsawicki:
Now that we have services and config that both use other exported properties from the package's main module,  implementing and exporting an `activate` function should be optional.

`deactivate` was already optional so this makes the two consistent with each other.

Packages just using services will no longer need a no-op `activate` function exported.


benogle:
:+1:

PR #5750, opened 2015-02-25
Remove the windows icon in updater MessageBox 's

### LINKED ISSUE #5670: Atom icon is still pixelated in certain occasions
This was mentioned in a different issue (https://github.com/atom/atom/issues/5500#issuecomment-73915709) but its own dedicated issue was never opened, so here we are.

The Atom icon still shows as pixelated (#4811) when the "No Update Available" box appears:
image

Something interesting to note is that unlike the main Atom window/taskbar icon, the popup has the white border, which makes me think that Windows is again trying to downsize the 1024px icon.

And unrelated: It would be nice if all the icon sizes could somehow get the white border applied to them for consistency :).

/cc @zcbenz?

jimgraham:
Do not set the icon in the updater dialog when
- on Windows
- calling the `showMessageBox` function.

Fixes atom/atom#5670

I only found two references to the `'atom.png'` file in `src/` and assumed this was the fix.

Looks like this when complete:
atom

### DIFF
--- a/src/browser/auto-update-manager.coffee
+++ b/src/browser/auto-update-manager.coffee
@@ -17,10 +17,12 @@ class AutoUpdateManager
 
   constructor: (@version) ->
     @state = IdleState
+    @iconPath = path.resolve(__dirname, '..', '..', 'resources', 'atom.png');
     if process.platform is 'win32'
       # Squirrel for Windows can't handle query params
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Thanks!

50Wliu:
:+1:

PR #5753, opened 2015-02-26
Add API for getting paths relative to project dirs

maxbrunsfeld:
In adding multi-folder support, a few packages have needed this, and I've had to write local helper functions:
- [tree-view](https://github.com/atom/tree-view/blob/master/lib/move-dialog.coffee#L23)
- [find-and-replace](https://github.com/atom/find-and-replace/blob/master/lib/project/result-view.coffee#L14)
- [fuzzy-finder](https://github.com/atom/fuzzy-finder/blob/d30e2b09988327d9436173386c22e918bd995901/lib/fuzzy-finder-view.coffee#L153)

I think before I write a fourth copy of this helper function for symbols-view, we should add _something_ like it to core. Any suggestions on this API or the naming? The way I have it, it's like this:

``` coffee
[rootPath, relativePath] = atom.project.splitPath(thePath)
if rootPath?
  console.log "#{thePath} is in project directory #{rootPath}. relative path: #{relativePath}"
else
  console.log "#{thePath} is not in any project directory"
```

/cc @nathansobo

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -438,22 +438,22 @@ describe "Project", ->
       randomPath = path.join("some", "random", "path")
       expect(atom.project.relativize(randomPath)).toBe randomPath
 
-  describe ".splitPath(path)", ->
+  describe ".relativizePath(path)", ->
     it "returns the root path that contains the given path, and the path relativized to that root path", ->
       atom.project.addPath(temp.mkdirSync("another-path"))
 
       rootPath = atom.project.getPaths()[0]
       childPath = path.join(rootPath, "some", "child", "directory")
-      expect(atom.project.splitPath(childPath)).toEqual [rootPath, path.join("some", "child", "directory")]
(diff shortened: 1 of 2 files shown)


nathansobo:
I like the functionality, but I'm grasping for a more obvious name.

My issue with `split` is `String::split` divides a string up into N parts, whereas this only divides the path in to 2 parts. Not sure how much that matters... `bisectPath` indicates the arity of the result but it's also a fancy word.

It feels like this is what `Project::relativize` should do, but that's already taken. I'm not totally sure that's the right name for this, but if it is, we could potentially soft-deprecate `::relativize` (just take it out of the docs) and replace it with `::relativizePath` that does this instead.

maxbrunsfeld:
Yeah, I was thinking something similar regarding `::relativize`. I like your idea of taking over that name anyway and just removing the docs for `::relativize`.

PR #5759, opened 2015-02-26
Do not hang on soft wrapping when editor is very small

### LINKED ISSUE #3999: Soft wrap causes editor to hang when less than 1 character wide
Steps to reproduce: 
   1) Make sure the tree view is visible by opening a folder.
   2) Open a file with some text in it
   3) Ensure soft wrap is enabled
   4) Resize the window so the editor is either barely or not at all visible.
   5) Editor should now hang, while the rest of Atom is responsive.

atom-soft-wrap-bug

OS: Windows 8.1
Version: 0.140.0

as-cii:
It appears that #5567 solved #3999 too, with a minor bug that happens during that scenario. This PR attempts to solve it, so that we can close #3999 once for all.

Current behavior (master):

soft-wrapping-before

After these changes:

soft-wrapping-after

_(the editor doesn't flash like this, it's a LICEcap issue I guess)_

Unfortunately I have found quite difficult to reproduce the issue via a test (so that we can guard against this regression), probably because when resizing the buffer we need to have a pane (like the tree-view). Any ideas how we can write one? Should we write one at all?

Thanks!

### DIFF
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -80,6 +80,14 @@ describe "DisplayBuffer", ->
           atom.config.set('editor.softWrapAtPreferredLineLength', false)
           expect(displayBuffer.tokenizedLineForScreenRow(10).text).toBe '    return '
 
+      describe "when editor width is negative", ->
+        it "does not hang while wrapping", ->
+          displayBuffer.setDefaultCharWidth(1)
+          displayBuffer.setWidth(-1)
+
+          expect(displayBuffer.tokenizedLineForScreenRow(1).text).toBe "  "
+          expect(displayBuffer.tokenizedLineForScreenRow(2).text).toBe "  var sort = function(items) {"
+
       describe "when the line is shorter than the max line length", ->
         it "renders the line unchanged", ->
           expect(displayBuffer.tokenizedLineForScreenRow(0).text).toBe buffer.lineForRow(0)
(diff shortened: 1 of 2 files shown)


src/tokenized-line.coffee
```diff
@@ -183,6 +185,9 @@ class TokenizedLine
   buildSoftWrapIndentationDelta: ->
     _.reduce @softWrapIndentationTokens, ((acc, token) -> acc + token.screenDelta), 0
 
+  isEmpty: ->
+    @tokens.length == 0 || @hasOnlySoftWrapIndentation()
+
   hasOnlySoftWrapIndentation: ->
     @tokens.length == @softWrapIndentationTokens.length
```
as-cii: This line would cover also the empty case, but I found it more expressive to have a separate method (`isEmpty`) that expresses intent better.

nathansobo:
Is there any way you could reproduce the bad scenario at the model layer only, but setting `DisplayBuffer::editorWidthInChars` to a small number or something? It would be really nice to have some sort of test on this, but it doesn't need to exactly reproduce the problem we see in the view layer.

as-cii:
@nathansobo: I finally managed to reproduce the issue through a test in `DisplayBuffer`. This led me to understand a bit better the problem, and thus to propose a better solution as well :bug: :bug: :bug: 

Basically we now sanitize `getEditorWidthInChars` to return always a column >= 0, as a negative width in chars would have little sense.

Thank you! :smile:

nathansobo:
Nice diligence sir!

PR #5762, opened 2015-02-26
Batch UI updates

as-cii:
This PR is still experimental, as it still needs some :heart:, but it demonstrates a few concepts discussed in #5025. It introduces, in facts, the notion of batching along with some other performance improvements :racehorse: You can find some pieces of information about batching in the aforementioned issue, but I'll report some key ideas here as well.

We noticed that a lot of computation was wasted in `TextEditorPresenter`, especially when performing multi-operations (e.g. select many things, write on many lines, etc.). If we write a character on 10 lines simultaneously, the whole UI would be recomputed 10 times, even though this was an atomic operation.

Therefore we came up with the concept of batching: before making an atomic operation, we explicitly put `TextEditor` on `batchMode`. This will emit two events: one on start and one at the end. `TextEditorPresenter` intercepts them and records what changes would have happened during the batching timeframe (if the same change would have been applied many times, it just keeps the last one). When batching ends, it applies all the recorded operation.

When `transact`ing, batching happens automatically. 

If we agree this could be a suitable redesign and we manage to refine this a bit, it seems like we get rid of almost any bottleneck in the presentation layer. Writing 5 characters on 1000 lines simultaneously shows that most of the time is now spent on `TextBuffer` and `mergeIntersectingSelections` (we need to make this even faster):

screen shot 2015-02-26 at 11 39 02

(the whole time was 87 seconds, and most of them were spent in the model layer)

/cc: @nathansobo


maxbrunsfeld:
Badass. :dragon:

src/text-editor.coffee
```diff
@@ -1140,7 +1144,23 @@ class TextEditor extends Model
   #   with a positive `groupingInterval` is committed while the previous transaction is
   #   still 'groupable', the two transactions are merged with respect to undo and redo.
   # * `fn` A {Function} to call inside the transaction.
-  transact: (groupingInterval, fn) -> @buffer.transact(groupingInterval, fn)
+  transact: (groupingInterval, fn) ->
+    @batch => @buffer.transact(groupingInterval, fn)
+
+  batch: (fn) ->
```
nathansobo: I wonder if we should be more explicit with the name, like `batchPresenterUpdates` or something. Just to clarify what the purpose of this is and to distinguish it from the model-level concept of `transact`.

src/text-editor.coffee
```diff
@@ -1140,7 +1144,23 @@ class TextEditor extends Model
...
+    @batchCount++
+    @emitter.emit "will-start-batch-operation" if @batchCount == 1
+    value = fn()
+    @emitter.emit "did-finish-batch-operation" if @batchCount == 1
+    @batchCount--
+
+    value
+
+  onWillStartBatchOperation: (callback) ->
```
nathansobo: `onWillBeginBatchingPresenterUpdates`?

src/text-editor.coffee
```diff
@@ -993,17 +996,18 @@ class TextEditor extends Model
   # selections to create multiple single-line selections that cumulatively cover
   # the same original area.
   splitSelectionsIntoLines: ->
-    for selection in @getSelections()
-      range = selection.getBufferRange()
-      continue if range.isSingleLine()
+    @batch => @mergeIntersectingSelections =>
```
nathansobo: Should the call to `mergeIntersectingSelections` always imply batching, similar to transactions?

nathansobo:
The code looks pretty good to me. As I read through it, I couldn't help but wonder if we could simplify our approach to batching. What if the presenter *always* batched operations, and only finalized the batch immediately before the view layer needs to update. I'm thinking the chain of events could look like this:

* Presenter receives update 1
  * Presenter sets a flag and does not update
  * Presenter tells the editor view it has updated, requests an animation frame
* Presenter receives update 2
  * Presenter sets another flag and does not update
  * Presenter tells the editor view it has updated, but an animation frame is already in flight so it's a no-op
* Animation frame fires
  * Editor tells the presenter to update its state based on information collected (boolean flags or otherwise) since the last update
  * Editor uses the presenter's state to update as normal.

The nice thing about that approach is we keep the behavior of the presenter completely separate from the model layer. We also ensure that the batch always extends as long as possible, waiting until the last possible moment before we need an update. It's similar to what we had with React where we recomputed the world at the last possible moment, except now we can be much smarter about how much of the world we recompute each time based on our knowledge of what changes have occurred.

This would eliminate the ability to handle individual updates without batching, but if we're always batching on transactions I can't imagine there are many updates that aren't being batched already anyway.

as-cii:
> The nice thing about that approach is we keep the behavior of the presenter completely separate from the model layer. We also ensure that the batch always extends as long as possible, waiting until the last possible moment before we need an update. It's similar to what we had with React where we recomputed the world at the last possible moment, except now we can be much smarter about how much of the world we recompute each time based on our knowledge of what changes have occurred.

That's a clever observation. Moreover, the code looks even more reasonable now. While the presenter hasn't changed a lot, we have moved the complexity of batching down into the presentation layer, which is where it should live.

Our issue was an "UI update overflow", and the responsible for keeping it normal must be who updates the UI. Kudos to @nathansobo for this brilliant idea :bulb: 

The routine works this way now:

1. When it receives an update from the model, `TextEditorPresenter` tracks it and emits a `did-update-state`, which in turn notifies `TextEditorComponent`.
2. Right before starting to `updateSync` the UI, `TextEditorComponent` applies all the collected changes (which causes a real update in `TextEditorPresenter`) and updates the world as usual.

----------------

I still need to work out a nice structure for the code, as I don't like it as it is right now. Meanwhile, though, it would be great if you could build this branch and play with it: there may be some glitch, but the performance improvement is really impressive (especially on multi-lines) :dash: :dash: :dash: 

Thank you!

as-cii:
@nathansobo: I refactored the code a bit and it looks quite right now to me. Maybe we still need to refine names a bit, but structure-wise I think it looks good overall.

I still need to write (and adapt) specs for `TextEditorPresenter`, because it now needs to be updated explicitly (through `TextEditorPresenter#update`). I am quite excited for this, therefore any feedback or recommendation is really appreciated :pray:

nathansobo:
Looks good. I want to see what you think about this one thought...

Right now each component in the view maintains a reference to the presenter. What if instead the `updateSync` method of each component took a state object as an argument. Then we could have the presenter continue to emit an `onDidUpdateState` event, but the `EditorComponent` would need to call `getState` in order to access the state rather than accessing it directly. The presenter could then perform last-minute updates when `getState` is called to return an appropriate state.

Basically, with that approach the batching behavior of the presenter would be fully encapsulated by the presenter. It's only promise would be to return an object from `getState` that would be usable for the duration of one view update. We would still recycle the same objects like we do now, but it might tell a bit clearer of a story by keeping the batching decisions localized.

as-cii:
> Right now each component in the view maintains a reference to the presenter. What if instead the updateSync method of each component took a state object as an argument. Then we could have the presenter continue to emit an onDidUpdateState event, but the EditorComponent would need to call getState in order to access the state rather than accessing it directly. The presenter could then perform last-minute updates when getState is called to return an appropriate state.

Agreed :+1: Unfortunately, `OverlayManager` needs to access the presenter directly, thus not allowing to remove the reference altogether, but I removed it wherever possible. On the other hand, I believe this has improved the quality of the code quite significantly: this is something that tests in `TextEditorPresenter` reflected as well. In facts, if we didn't use `getState()`, updating those tests would have been a nightmare because of the lack of encapsulation. We don't even need to write new tests for this behavior, as it's an implementation detail of `TextEditorPresenter`.

Moreover, I thought it may have been useful to avoid passing the whole state to sub-components, e.g.:

```coffee
# Inside TextEditorComponent#updateSync
@linesComponent.updateSync(state.content)
```

However, if we need to change how `LinesComponent` behaves in the future (e.g. if, for instance, needs to access other state properties), we would have to change the code in two places rather than one. Therefore, I decided to always pass the whole `state`.

All the specs pass now. I removed the `WIP` on the PR and I think this is ready for test-drive :+1:

src/text-editor-presenter.coffee
```diff
@@ -43,6 +48,59 @@ class TextEditorPresenter
     @model.setVerticalScrollbarWidth(@measuredVerticalScrollbarWidth) if @measuredVerticalScrollbarWidth?
     @model.setHorizontalScrollbarHeight(@measuredHorizontalScrollbarHeight) if @measuredHorizontalScrollbarHeight?
 
+  # Private: Determines whether {TextEditorPresenter} is currently batching changes.
+  # Returns a {Boolean}, `true` if is collecting changes, `false` if is applying them.
+  isBatching: ->
+    @updating == false
```
nathansobo: Minor detail: Let's use `is` here.

nathansobo:
Yes more great work @as-cii. Feel free to open your next PR directly on atom/atom if you'd like to. No need to use your fork.

thomasjo:
Awesome work! :sparkles:

PR #5775, opened 2015-02-26
Don't blink cursors unless editor is focused

### LINKED ISSUE #4378: Atom Cursor causes high CPU load (20% x 2 processors.)
The way atom does cursors causes a high CPU load when debugging anything inside of atom because it is constantly causing chrome inspector to update which is apparently very expensive.  It would be nice if the cursor would also pause when the inspector is in focus, or for there to even be an option to simply disable the blink entirely since it has many performance problems (actually, if the cursor weren't blinking there would be no CPU usage in general with Atom on my Linux system.)

There is probably a ticket for the latter but the former is an issue of it's own I think (I searched.)

nathansobo:
Fixes #4378


benogle:
Cool

wmertens:
:beers:

PR #5786, opened 2015-02-27
Don't show foldable icon on soft wrapped lines

### LINKED ISSUE #5739: Fold icon shown in gutter for all physical lines of soft-wrapped line
Taken from https://github.com/atom/atom/pull/5567#issuecomment-75758499

Each physical line has a fold icon if the soft-wrapped line has nested lines. Instead, I expected only the first physical line to have the fold icon (that's also the behavior in Sublime).

izuzak:
Fixes https://github.com/atom/atom/issues/5739.
-  574158d adds a failing spec to demonstrate the problem. Running the specs results in this (expected) output:
  
  ```
  TextEditorComponent
  gutter rendering
    fold decorations
      rendering fold decorations
        when soft wrapping is enabled
          it doesn't add the foldable class for soft-wrapped lines
            Expected true to be false. (spec/text-editor-component-spec.coffee:650:55)
  ```
- 09ed657 adds a check in the GutterComponent so that fold icons are added only on the first physical line of soft wrapped lines. Running the specs doesn't trigger the failure above. 

Before :tv::

foldy2

After :tv::

foldy

cc @benogle @nathansobo for :eyes:

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -637,6 +637,18 @@ describe "TextEditorComponent", ->
           nextAnimationFrame()
           expect(lineNumberHasClass(5, 'folded')).toBe false
 
+        describe "when soft wrapping is enabled", ->
+          beforeEach ->
+            editor.setSoftWrapped(true)
+            nextAnimationFrame()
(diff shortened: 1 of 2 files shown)


nathansobo:
:sparkling_heart: Thanks @izuzak!

Issue #4270, opened 2014-11-21
protocol isn't getting to my opener on atom load


mark-hahn:
I have this code to open http protocols in my package `web-browser`.  It runs in my activate method. ...

```
@opener = (filePath, options) =>
  if /^https?:\/\//.test filePath
    new Page @, filePath
atom.workspace.registerOpener @opener
```

I know this works properly because when I execute this code from a different package my opener catches it and everything works ...

```
atom.workspace.open url
```

However, if I start atom with the command `atom http://apple.com` then my opener never gets a chance to catch the url.  Instead atom opens up an empty editor tab with the url as the tab label.  The tree-view is empty.

I'm guessing that my package hasn't had a chance to register the opener yet.  My `activationEvents` in `package.json` is an empty array so I'm activating as fast as I can.

 This is a problem.  Any ideas on how I or the team can fix this?  If it is possible to fix I'd be happy to do a PR.

PR #5795, opened 2015-02-27
Default directory provider should handle only local filesystem directories

mostafaeweda:
I'm having an issue of:
When I add an atom directory provider with uri like: myuri://server:port/some_path and reload atom or close/open atom, the default directory provider normalizes my uri path (bc it calls path.normalize on it eve though it's not a local filesystem directory).
https://github.com/atom/atom/blob/master/src/default-directory-provider.coffee#L16-L24

How do you think about the following solution:
1- The deafult directory provider returns a directory only when it exists on the filesystem.
2- On deserializing and creating the project with paths, if a projectPath can't be handled by a directory provider, it's added to another array to be consumed later by atom packages, when they are loaded

That's trying to work-around the nature of atom core initializing before packages are loaded and able to provide their directory providers.

### DIFF
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -21,7 +21,10 @@ class DefaultDirectoryProvider
     else
       path.dirname(projectPath)
 
-    new Directory(directoryPath)
+    if fs.isDirectorySync(projectPath)
+      return new Directory(directoryPath)
+    else
+      return null
 
   # Public: Create a Directory that corresponds to the specified URI.
   #
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -16,15 +16,12 @@ class DefaultDirectoryProvider
   directoryForURISync: (uri) ->
     projectPath = path.normalize(uri)
 
-    directoryPath = if fs.isDirectorySync(projectPath)
-      projectPath
-    else
+    directoryPath = if !fs.isDirectorySync(projectPath) and fs.isDirectorySync(path.dirname(projectPath))
       path.dirname(projectPath)
-
-    if fs.isDirectorySync(projectPath)
-      return new Directory(directoryPath)
     else
-      return null
+      projectPath
+
+    new Directory(directoryPath)
 
   # Public: Create a Directory that corresponds to the specified URI.
   #
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -205,7 +205,7 @@ class Project extends Model
       break if directory = provider.directoryForURISync?(projectPath)
     if directory is null
       # DefaultDirectoryProvider doesn't handle specific directory protocols
-      # Atom Packages may use packageDirectoryPaths to add thier own paths later.
+      # Atom Packages may use packageDirectoryPaths to add their own paths later.
       @packageRootDirectoryPaths.push(projectPath)
       return
     else
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -39,6 +39,7 @@ class Project extends Model
     @emitter = new Emitter
     @buffers ?= []
     @rootDirectories = []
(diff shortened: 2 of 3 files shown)


mostafaeweda:
@kevinsawicki how do you find this?

mark-hahn:
> That's trying to work-around the nature of atom core initializing before
packages are loaded and able to provide their directory providers.

I really want this also.  This problem causes Atom to not be able to start
with web pages.

https://github.com/atom/atom/issues/4270

src/project.coffee
```diff
@@ -203,10 +204,12 @@ class Project extends Model
     for provider in @directoryProviders
       break if directory = provider.directoryForURISync?(projectPath)
     if directory is null
-      # This should never happen because DefaultDirectoryProvider should always
-      # return a Directory.
-      throw new Error(projectPath + ' could not be resolved to a directory')
-    @rootDirectories.push(directory)
+      # DefaultDirectoryProvider doesn't handle specific directory protocols
+      # Atom Packages may use packageDirectoryPaths to add thier own paths later.
```
50Wliu: `their`

mark-hahn: So you are saying we can misuse the directoryProvider to handle protocols
like `http://`?

mostafaeweda: No, that's for the case when you have a directory sitting on your server and have a directory provider that serves files from it to your Atom instance (in your package) - e.g. http://someserver:port/workspace_path
When you Atom restarts and tries to load this directory, it will use its default-directory-provider and hence, fail and destroy your project uri - or path, because the `default-directory-provider` and the default pathwatcher's `Directory` normalizes the given path (which could sometimes not be a filesystem path, but a uri).

So, this diff is to make the `default-directory-provider` skip loading a certain path, if it can't (doesn't exist on the local filesystem).

src/default-directory-provider.coffee
```diff
@@ -21,7 +21,10 @@ class DefaultDirectoryProvider
     else
       path.dirname(projectPath)
 
-    new Directory(directoryPath)
+    if fs.isDirectorySync(projectPath)
+      return new Directory(directoryPath)
```
kevinsawicki: This return statement isn't needed

src/default-directory-provider.coffee
```diff
@@ -21,7 +21,10 @@ class DefaultDirectoryProvider
...
+    else
+      return null
```
kevinsawicki: This return statement isn't needed

mostafaeweda:
I guess it's easier to merge if I stick to the bare minimum that's blocking which is getting rid of the destructive non-needed `path.dirname()` call when the `directoryPath` isn't really a directory.

Also, adding two tests for the expected behavior.

mostafaeweda:
@kevinsawicki does this look good?

maxbrunsfeld:
@mostafaeweda we'd prefer to avoid adding another instance variable (`@packageRootDirectoryPaths`) and requiring packages to use it instead of the existing methods in the project API.

What do you think about this solution instead: 
* In `Project::setPaths`, we store the original paths in a separate array:

```coffee
  setPaths: (projectPaths) -> 
    @projectPaths = projectPaths
    # ...
```

* We return these original paths from `Project::getPaths`:

```coffee
  getPaths: -> @projectPaths
```

* When a `atom:directory-provider` service becomes available, we recompute the `Directory` objects for each path:

```coffee
  atom.packages.serviceHub.consume(
    'atom.repository-provider',
    '^0.1.0',
    (provider) =>

      # ...

      if null in @directories
        @setPaths(@getPaths())
)
```

This is similar to the logic that @bolinfest added for the `atom.repository-provider` service. Would you mind trying this solution? I think it should work for your use case.

PR #4429, opened 2014-12-07
Add subword cursors - take 3

dsandstrom:
Fixes #1982 and an alternative to #3961. Also, hopefully can used with atom/vim-mode#377.

I can't figure out how to go left on the first line when there is only white space, see cursor.moveToSubwordBoundary(), /^[\t ]*$/ isn't working.

I haven't added keymaps yet. But maybe:

``` cson
'.platform-linux .editor, .platform-win32 .editor':
  'alt-left': 'editor:move-to-previous-subword-boundary'
  'alt-right': 'editor:move-to-next-subword-boundary'
  'alt-shift-left': 'editor:select-to-previous-subword-boundary'
  'alt-shift-right': 'editor:select-to-next-subword-boundary'
  'alt-backspace': 'editor:delete-to-beginning-of-subword'
  'alt-k': 'editor:delete-to-end-of-subword'

'.platform-darwin .editor':
  'ctrl-alt-left': 'editor:move-to-previous-subword-boundary'
  'ctrl-alt-right': 'editor:move-to-next-subword-boundary'
  'ctrl-alt-shift-left': 'editor:select-to-previous-subword-boundary'
  'ctrl-alt-shift-right': 'editor:select-to-next-subword-boundary'
  'ctrl-alt-backspace': 'editor:delete-to-beginning-of-subword'
  'ctrl-alt-k': 'editor:delete-to-end-of-subword'
```

I would not mind if both were the same, though (`ctrl-alt`).  But Sublime uses just `alt` on Linux/Win, I'm not sure about Mac.

PR #5806, opened 2015-03-02
Add subword cursors - take 4

### LINKED ISSUE #1982: Support subword navigation
SublimeText supports navigation by subwords (underscore separated or hills in camelcase):
```
{ "keys": ["alt+left"], "command": "move", "args": {"by": "subwords", "forward": false} },
{ "keys": ["alt+right"], "command": "move", "args": {"by": "subword_ends", "forward": true} },
{ "keys": ["alt+shift+left"], "command": "move", "args": {"by": "subwords", "forward": false, "extend": true} },
{ "keys": ["alt+shift+right"], "command": "move", "args": {"by": "subword_ends", "forward": true, "extend": true} }
```

It would be nice if Atom also supported `move*`, `delete*`, `backspace*`, `select*` editor commands based on subwords.

dsandstrom:
Add commands for moving, selecting, and deleting camelCase words.

Fixes #1982 and an alternative to #3961. Also, hopefully can used with atom/vim-mode#377.

Replaces #4429.


nathansobo:
Apologies for the long delay in reviewing this. Have been taking a look this afternoon. I'd like to investigate whether `TextBuffer.prototype.scanInRange` can be inclusive of its start position so we don't need the hack at the beginning of the file.

nathansobo:
Thanks very much for doing this. It's a really nice addition. Apologies again for the delay. We got pretty swamped in the approach to 1.0.

adrianoresende:
I installed now, Atom 1.7.4 in OS X El Capitan 10.11.4 and this default CTRL + ARROW not work for me in editor.

dsandstrom:
Its ctrl-alt-arrow on osx: https://github.com/atom/atom/blob/master/keymaps/darwin.cson#L143

adrianoresende:
@dsandstrom But CTRL + ARROW it is not standard?

dsandstrom:
Who's standard?  In Atom, ctrl-arrow runs the current OS' "native" command.  In OSX, I believe this is used by Mission Control to switch desktops by default.  Alt is already being used to go the beginning/end of the line, cmd is being used to go the beginning/end of a word.  So, we are stuck with ctrl-alt on OSX.

PR #5812, opened 2015-03-03
Add Workspace::onWillOpen 

benogle:
This will add an `Workspace::onWillOpen` event to allow fixing https://github.com/atom/welcome/issues/26. Usage would look something like this

``` coffee
# Suppress the opening of the initial empty buffer
atom.workspace.onWillOpen ({uri, options, cancel}) ->
  cancel() if options.isInitialEmptyBuffer
```

Refs https://github.com/atom/welcome/issues/26

How do you guys feel about this approach?

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -73,7 +73,8 @@ class AtomWindow
     @browserWindow.loadUrl @getUrl(loadSettings)
     @browserWindow.focusOnWebView() if @isSpec
 
-    @openLocations(locationsToOpen) unless @isSpecWindow()
+    hasPathToOpen = not (locationsToOpen.length is 1 and not locationsToOpen[0].pathToOpen?)
+    @openLocations(locationsToOpen) if hasPathToOpen and not @isSpecWindow()
 
   getUrl: (loadSettingsObj) ->
     # Ignore the windowState when passing loadSettings via URL, since it could
--- a/spec/atom-spec.coffee
+++ b/spec/atom-spec.coffee
@@ -152,3 +152,46 @@ describe "the `atom` global", ->
       loadSettings.initialPaths = [dir2, dir1]
       atom2 = Atom.loadOrCreate("editor")
       expect(atom2.state.stuff).toBe("cool")
+
+  describe "openInitialEmptyEditorIfNecessary", ->
+    describe "when there are no paths set", ->
+      oldPaths = null
+
+      beforeEach ->
+        oldPaths = atom.project.getPaths()
+        atom.project.setPaths([])
+
+      afterEach ->
+        atom.project.setPaths(oldPaths)
+
+      it "opens an empty buffer", ->
+        spyOn(atom.workspace, 'open')
+        atom.openInitialEmptyEditorIfNecessary()
+        expect(atom.workspace.open).toHaveBeenCalledWith(null, {isInitialEmptyEditor: true})
+
+      describe "when there is already a buffer open", ->
+        beforeEach ->
+          waitsForPromise -> atom.workspace.open()
+
+        it "does not open an empty buffer", ->
+          spyOn(atom.workspace, 'open')
+          atom.openInitialEmptyEditorIfNecessary()
+          expect(atom.workspace.open).not.toHaveBeenCalled()
+
+    describe "when the project has a path", ->
+      beforeEach ->
+        spyOn(atom.workspace, 'open')
+
+      it "does not open an empty buffer", ->
+        atom.openInitialEmptyEditorIfNecessary()
+        expect(atom.workspace.open).not.toHaveBeenCalled()
+
+    describe "when there is already a buffer open", ->
+      beforeEach ->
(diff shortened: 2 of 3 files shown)


nathansobo:
@benogle Can you talk a bit about the `options` parameter? Where does it come from and what other kinds of options would be present?

benogle:
It's just the options passed to `workspace.open`. Check out the diff

nathansobo:
:cool: Sorry for being lazy I could have read that myself.

src/browser/atom-window.coffee
```diff
@@ -73,6 +73,9 @@ class AtomWindow
     @browserWindow.loadUrl @getUrl(loadSettings)
     @browserWindow.focusOnWebView() if @isSpec
 
+    if locationsToOpen.length is 1 and locationsToOpen[0].pathToOpen is null
```
benogle: Dislike this. Maybe a better option would be to pass an empty `locationsToOpen` to the constructor, or an option `openInitialEmptyBuffer`, then build this `locationsToOpen` hash.

src/workspace.coffee
```diff
@@ -391,6 +394,11 @@ class Workspace extends Model
   #
   # Returns a promise that resolves to the {TextEditor} for the file URI.
   open: (uri, options={}) ->
+    cancelOpen = false
```
nathansobo: `cancelled`?

benogle: Ah, yeah that's nicer

src/workspace.coffee
```diff
@@ -391,6 +394,11 @@ class Workspace extends Model
...
+    cancel = -> cancelOpen = true
+    @emitter.emit('will-open', {uri, cancel, options})
+    return if cancelOpen
```
benogle: I should probably return a `Promise.resolve()` here, eh?

src/browser/atom-window.coffee
nathansobo: Can you do this earlier? At what point do we decide to open the initial empty buffer? Seems like you could do it then.

src/workspace.coffee
nathansobo: Kind of tricky since the caller may be expecting things to actually be open. Should it resolve with a boolean now?

src/browser/atom-window.coffee
benogle: Yeah, I'd like to. This is the beginning: https://github.com/atom/atom/blob/8fac0e87788aaf398adee0bb75f034ef497eb971/src/browser/atom-application.coffee#L93

I will look at propagating this with fresh eyes tomorrow. It seemed there would be some messiness in AtomApplication::openPath(). It's expecting a path. I could split out the [window creation things](https://github.com/atom/atom/blob/8fac0e87788aaf398adee0bb75f034ef497eb971/src/browser/atom-application.coffee#L378-L385) to another method. It just all felt dirty. This was the quickest thing I could get out to show the direction I wanted to take.

kevinsawicki: Just curious, I feel like I've seen `is null` being used lately, is there a reason to use it instead of just a `?` ?

benogle: Other than just being explicit, no

kevinsawicki: Yeah, so in this case, wouldn't you want it to go down that path if it is also `undefined`?

benogle: Yeah using `?` is fine

kevinsawicki:
Would be great to get specs for this in the new https://github.com/atom/atom/blob/master/spec/integration/startup-spec.coffee

benogle:
Yeah was planning on it

src/browser/atom-window.coffee
```diff
@@ -73,7 +73,8 @@ class AtomWindow
     @browserWindow.loadUrl @getUrl(loadSettings)
     @browserWindow.focusOnWebView() if @isSpec
 
-    @openLocations(locationsToOpen) unless @isSpecWindow()
+    hasPathToOpen = not (locationsToOpen.length is 1 and not locationsToOpen[0].pathToOpen?)
```
as-cii: Nice work here!

What do you think about moving this down into `openLocations` (maybe as a guard clause), so that if we call it from somewhere else in the future it just works?

benogle: I'd rather have the check outside `openLocations` for clarity when scanning this code. openLocations can handle the null case just fine (it opens an empty editor). 

The empty editor logic is moving to the renderer process.

benogle: Also, I want to minimize edits to the browser code. Changes to it are scary as it's more on the opaque and hard-to-test side of things.

as-cii: Makes sense :+1:

benogle:
No longer want to add the event. There are implications for the cancellation of `workspace.open`. Packages rely on the lack of cancellation. 

For atom/welcome#26, I am happy having the welcome package open its panes over top the untitled buffer.

PR #5813, opened 2015-03-03
Make it possible to restore a root folder created by a DirectoryProvider.

bolinfest:
Previously, the challenge in restoring a root folder created by a `DirectoryProvider`
was that the `DirectoryProvider` that created the root folder during the previous
session may not be available when the `Project` is restored via deserialization.

This adds a new method to the `DirectoryProvider` API named `getHostPackage()`
(it also bumps the version of `atom.directory-provider` from 0.1.0 to 0.2.0).
A `DirectoryProvider` must return the name of the Atom package that provided it
or `null`. (Surprisingly, it is not possible to get this information programmatically
today, though ServiceHub could potentially be expanded to provide this. However,
it would only work for providers statically declared in `package.json`, but not for
those that are registered dynamically, so it might not be worth it.)

When a `Directory` is created via a `DirectoryProvider`, the correspoding Atom
package via `DirectoryProvider::getHostPackage()` is recorded as part of the `Project`
so it can be serialized along with the `Project`'s paths.

This way, when `Project` is deserialized, paths that have a corresponding "host package"
can be filtered from ones that do not. Paths without a corresponding host package will
be loaded normally. Those with a host package will have their package activated so
the `DirectoryProvider` can be restored, which will hopefully be able to recreate the
root folder from the previous session.

When a new `DirectoryProvider` is registered, the `Project` checks to see whether
there were any paths from the previous session that correspond to its host package.
If so, each path is fed to the provider's async `directoryForURI` method. Assuming it
returns a `Promise` that returns a non-null value, the path will be added to the
`Project` via `atom.project.addPath()`. This allows the provider to do any async I/O
or network communication it needs to do to recreate the `Directory` (which it should
cache) before its synchronous `directoryForURISync()` method is invoked as a side-effect
of `atom.project.addPath()`. Assuming the provider caches the path->`Directory` mapping
during the async invocation, it can return it immediately as part of the sync invocation.

### DIFF
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -5,6 +5,14 @@ path = require 'path'
 module.exports =
 class DefaultDirectoryProvider
 
+  # Public: Identify the Atom package that must be activated to load this
+  # provider.
+  #
+  # Returns:
+  # * {String} name of the Atom package.
+  # * `null` if this provider is loaded automatically by Atom core.
+  getHostPackage: () -> null
+
   # Public: Create a Directory that corresponds to the specified URI.
   #
   # * `uri` {String} The path to the directory to add. This is guaranteed not to
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -5,14 +5,6 @@ path = require 'path'
 module.exports =
 class DefaultDirectoryProvider
 
-  # Public: Identify the Atom package that must be activated to load this
-  # provider.
-  #
-  # Returns:
-  # * {String} name of the Atom package.
-  # * `null` if this provider is loaded automatically by Atom core.
-  getHostPackage: () -> null
-
   # Public: Create a Directory that corresponds to the specified URI.
   #
   # * `uri` {String} The path to the directory to add. This is guaranteed not to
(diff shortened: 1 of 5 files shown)


bolinfest:
Working on the tests now.

src/default-directory-provider.coffee
```diff
@@ -5,6 +5,14 @@ path = require 'path'
 module.exports =
 class DefaultDirectoryProvider
 
+  # Public: Identify the Atom package that must be activated to load this
+  # provider.
+  #
+  # Returns:
+  # * {String} name of the Atom package.
+  # * `null` if this provider is loaded automatically by Atom core.
+  getHostPackage: () -> null
```
kevinsawicki: :fire: `()`

bolinfest:
OK, this is now in a state where things work. Now I'll see if I can backpedal and remove the `getHostPackage()` method so we can stay with v0.1.0 for the `atom.directory-provider` API. The only downside is that packages that provide such an API must not have activation commands because they must load on startup. Such packages should make sure that their startup costs are as small as possible.

bolinfest:
Moving to new, simpler approach in https://github.com/atom/atom/pull/5830.

mnquintana:
@bolinfest Does that mean this can be closed?

izuzak:
> Moving to new, simpler approach in #5830.

Closing this in favor of that PR then.

PR #5814, opened 2015-03-03
Select buffer lines when clicking the gutter, not screen lines

### LINKED ISSUE #5738: Clicking the gutter for soft-wrapped lines selects the physical line instead of the logical line
Take from https://github.com/atom/atom/pull/5567#issuecomment-75758499.

When I click on the line's number in the gutter, I expected to see the whole logical line to be selected (that's the behavior in Sublime, for example). Instead, only the physical line is selected (kind of, it strange):

izuzak:
Fixes https://github.com/atom/atom/issues/5738:
-  d21ae2c modifies the click handlers for the gutter to select buffer lines, rather than screen lines. Clicking the gutter for any screen line of a wrapped buffer line selects the whole buffer line (previously, only the clicked screen line would have been selected).
- 03a064b adds more tests for checking how click handlers for the gutter behave when soft wrap is turned on (I duplicated all the existing tests, turned on soft wrap, and then tweaked them to test the new functionality).

Before :tv: (notice screen lines being selected):

After :tv: (notice buffer lines being selected):

gutter

cc @nathansobo for :eyes:


spec/text-editor-component-spec.coffee
```diff
@@ -1845,6 +1845,135 @@ describe "TextEditorComponent", ->
             nextAnimationFrame()
             expect(editor.getSelectedScreenRange()).toEqual [[3, 4], [9, 0]]
 
+    describe "when soft wrap is enabled", ->
```
nathansobo: These are excellent, thorough tests!

nathansobo:
Kicked CI again to see if the build failures were a fluke. The code looks good to me. Thanks for taking care of this @izuzak. You are a champion.

nathansobo:
:clap:

PR #5824, opened 2015-03-03
Move the logic for opening empty initial buffer to the renderer process

### LINKED ISSUE #26: Guide shows behind untitled editor on first run
Wasn't sure if this was intentional or not, but it felt weird to have the guide behind the untitled buffer instead of in front of it when Atom is first launched.

screen shot 2015-03-02 at 9 04 06 am

benogle:
Using this solution to atom/welcome#26 instead of #5812

Closes atom/welcome#26

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -632,7 +632,7 @@ class Atom extends Model
     @windowEventHandler?.unsubscribe()
 
   openInitialEmptyEditorIfNecessary: ->
-    if @project.getPaths().length is 0 and @workspace.getPaneItems().length is 0
+    if @getLoadSettings().initialPaths?.length is 0 and @workspace.getPaneItems().length is 0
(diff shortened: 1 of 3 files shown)


src/atom.coffee
```diff
@@ -629,6 +631,10 @@ class Atom extends Model
 
     @windowEventHandler?.unsubscribe()
 
+  openInitialEmptyEditorIfNecessary: ->
+    if @getLoadSettings().initialPaths?.length is 0 and @workspace.getPaneItems().length is 0
+      @workspace.open(null, {isInitialEmptyEditor: true})
```
kevinsawicki: Is the `isInitialEmptyEditor` option used anywhere?

benogle: I removed it from master

PR #5830, opened 2015-03-03
Optional API for DirectoryProvider: excludeFromSerialization(Directory).

bolinfest:

PR #5832, opened 2015-03-04
Fix a bug when a URI is passed to atom.project.removePath().

bolinfest:


kevinsawicki:
Can you add a spec for this?

bolinfest:
@kevinsawicki I can in exchange for https://github.com/atom/atom/pull/5830 :grinning:

Working on a test now.

bolinfest:
@kevinsawicki Test added.

spec/project-spec.coffee
```diff
@@ -436,6 +436,29 @@ describe "Project", ->
       expect(atom.project.getPaths()).toEqual([path.join(__dirname, "..", "src")])
       expect(atom.project.getRepositories()[0].isSubmodule("src")).toBe false
 
+    it "removes a path that is represented as a URI", ->
+      ftpUri = "ftp://example.com/some/folder"
```
maxbrunsfeld: ftpURI is how we roll these days.

spec/project-spec.coffee
```diff
@@ -436,6 +436,29 @@ describe "Project", ->
...
+      directoryProvider =
+        directoryForURISync: (uri) ->
+          if uri is ftpUri
```
maxbrunsfeld: Since this `directoryProvider` is just here to allow us to set `ftpURI` as a path, I would get rid of the `if`/`else`/`throw` logic in `directoryForURISync` and simply return your fake `Directory` unconditionally. That way the test can get straight to the point.

maxbrunsfeld:
Makes sense. Thanks for the fix!

spec/project-spec.coffee
bolinfest: done

spec/project-spec.coffee
bolinfest: Sure thing: I wasn't sure how defensive you wanted the test to be. All cleaned up.

PR #5839, opened 2015-03-04
Export TextBuffer for external package usage

mostafaeweda:
@kevinsawicki @bolinfest

### DIFF
--- a/exports/atom.coffee
+++ b/exports/atom.coffee
@@ -1,4 +1,5 @@
-{Point, Range} = require 'text-buffer'
+TextBuffer = require 'text-buffer'
+{Point, Range} = TextBuffer
 {Emitter, Disposable, CompositeDisposable} = require 'event-kit'
 {deprecate} = require 'grim'
 
@@ -7,6 +8,7 @@ module.exports =
   BufferedProcess: require '../src/buffered-process'
   GitRepository: require '../src/git-repository'
   Notification: require '../src/notification'
+  TextBuffer: TextBuffer
   Point: Point
   Range: Range
   Emitter: Emitter


bolinfest:
@kevinsawicki Is there any reason why this wasn't exported before?

kevinsawicki:
>  Is there any reason why this wasn't exported before?

Probably because no packages ever needed it. They always accessed the buffer from the editor and so they never needed to create one directly.

bolinfest:
OK, cool, just making sure there wasn't any trickiness like there is for `Directory`.

kevinsawicki:
So are you planning on creating the text buffer manually and passing into a manually created text editor?

mostafaeweda:
Yep, that's our existing case but we require the original text-buffer it in a tricky way.
We will extend `TextBuffer`'s `setPath` function (https://github.com/atom/text-buffer/blob/master/src/text-buffer.coffee#L326) to create the remote file instance we have instead of it being created using `new File()` which is pathwatcher's version of the `File`

Also, after this, I'll do another PR to the `text-buffer` that makes the existing setPath call a method: `createFile(path)` - where we can simply extend and provide our version of the file instead of having to copy the contents of the `setPath` function.

And this way, we don't require the added `setFile` function to the API introduced in https://github.com/atom/text-buffer/pull/42 - then we don't even need this PR.

kevinsawicki:
Merging this sounds good to me then, @nathansobo/@benogle any objections?

mostafaeweda:
@kevinsawicki could we also have this with a new atom version release?

mostafaeweda:
PING

nathansobo:
No objections, though we may be replacing `TextBuffer` with a new document model some time in the next few months. We can make it API-compatible however and we're already exposing instances anyway.

PR #5843, opened 2015-03-04
Speed up `selectAll` on a buffer with many folded lines

### LINKED ISSUE #5791: Select all very slow when lots of folded content
Doing &lt;ctrl>a to select the complete buffer, when most of the content is hidden in folded sections is very slow, and it unfolds everything.

I do not see why either of those should happen.

I do this regularly with some javascript files.  I start with everything folded, expand a small section, make changes, then select / copy the whole buffer to paste into jslint.com.  I have not gotten around to installing the jslint package.

as-cii:
As pointed out on #5791, selecting the whole buffer with many folded lines caused a serious performance degradation. I tracked down the issue to `Selection#selectAll` trying to unfold those lines which contained the selected buffer range. However, since we know in advance that we're going to select the whole text, we can make the clever assumption that everything would be unfolded anyways, thus calling the faster `@editor.unfoldAll` (e.g. faster because it simply scans each line one by one, instead of traversing the buffer back and forth for each line in the selection).

So, we go from this:

screen shot 2015-03-04 at 23 49 43

To this:

screen shot 2015-03-04 at 23 50 05

_(Benchmark performed on `benchmark/fixtures/huge.js`, on my machine it's 70% faster)_

With this change, `selectAll` on a buffer with many folded lines takes no longer than unfolding everything :racehorse: :racehorse: :racehorse: 

Close #5791 

/cc: @nathansobo @izuzak

### DIFF
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -251,7 +251,8 @@ class Selection extends Model
 
   # Public: Selects all the text in the buffer.
   selectAll: ->
-    @setBufferRange(@editor.buffer.getRange(), autoscroll: false)
+    @editor.unfoldAll()
+    @setBufferRange(@editor.buffer.getRange(), autoscroll: false, preserveFolds: true)
 
   # Public: Selects all the text from the current cursor position to the
   # beginning of the line.


nathansobo:
Ideally, I'd like to only expand folds that *contain* the range we are selecting. But considering that this optimizes the existing behavior, I think we should merge it and revisit the behavior later. :sparkles: work once again @as-cii.

izuzak:
:zap: @as-cii

PR #5875, opened 2015-03-05
Always open all CLI paths in the same window

### LINKED ISSUE #770: Multiple folders in projects
In Sublime I can have multiple folders in the side bar. If Atom already has this, I haven't been able to figure out how to do it.

screen shot 2013-08-29 at 10 17 32 am

maxbrunsfeld:
No more `atom --multi-folder`

Closes #770.

### DIFF
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -87,7 +87,7 @@ describe "Starting Atom", ->
       nestedDir = path.join(otherTempDirPath, "nested-dir")
       fs.mkdirSync(nestedDir)
 
-      runAtom [tempDirPath, otherTempDirPath, "--multi-folder"], {ATOM_HOME: AtomHome}, (client) ->
+      runAtom [tempDirPath, otherTempDirPath], {ATOM_HOME: AtomHome}, (client) ->
         client
           .waitForExist("atom-workspace", 5000)
           .treeViewRootDirectories()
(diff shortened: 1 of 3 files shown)


philipgiuliani:
There is no way to disable this behaviour or? Im used to have a window for every project.. :)

maxbrunsfeld:
We definitely want this new behavior to be the default, but I'm willing to consider adding a flag if this is really important. So you're saying you often want to quickly open multiple folders in separate windows at the same time? Can't you just do this?

```shell
atom dir1
atom dir2
```

philipgiuliani:
Ah if that still works its ok thanks, i though this PR overwrote that behaviour. Currently im always starting atom with `atom .`

maxbrunsfeld:
:+1: This PR is just for handling multiple path arguments (`atom dir1 dir2`).

philipgiuliani:
Ahh i understand thanks :) than i like the nrw behaviour, i dont even knew thats its possible to open multiple folders like this ;) will use it more often now :)

Zireael07:
Thanks a lot for this change, it made it possible to use the Project Manager package to open two or three root folders at once. 

I don't think I'll go back to ST 3... :D

kimegede:
`atom --help` doesn't says anything about a flag for `append`.

This makes 2 windows, where I only wanted 1.
```bash
atom dir1
atom dir2
```

PR #5893, opened 2015-03-07
Update clean.cmd

ghost:
Clean code.


izuzak:
I'm not :-1: on changes like these, but this makes the style a bit inconsistent -- open paren `(` in same line for the `if` part, and on a new line for the `else` part. Also, the user who created the pull request deleted their account, so noone can really update it.

Issue #5888, opened 2015-03-07
Windows build readme need update


basarat:
From : https://github.com/atom/atom/blob/master/docs/build-instructions/windows.md#common-errors I did not get the error : `error MSB8020: The build tools for Visual Studio 2010 (Platform Toolset = 'v100') cannot be found`, but getting past `Installing build modules` was done by the solution recommended for it `$env:GYP_MSVS_VERSION=2013`.

I also needed to run `npm install -g node-gyp` before `gypgypgyp` started showing.

The error I got was: 
```
C:\atom [master +0 ~1 -0]> script/build
Node: v0.10.26
npm: v2.5.1
Python: v2.7.9
Installing build modules...
gypgypgypnpm ERR! Windows_NT 6.2.9200
npm ERR! argv "node" "C:\\atom\\build\\node_modules\\npm\\bin\\npm-cli.js" "--userconfig=C:\\atom\\.npmrc" "install" "--
loglevel" "error"
npm ERR! node v0.10.26
npm ERR! npm  v2.5.1
npm ERR! code ELIFECYCLE

npm ERR! minidump@0.8.0 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the minidump@0.8.0 install script 'node-gyp rebuild'.
npm ERR! This is most likely a problem with the minidump package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls minidump
npm ERR! There is likely additional logging output above.
```

Either the build script needs to output the source of the error OR we need to update the readme. Either way wanted to create this issue for tracking / google results.

Currently I am seem to be stuck (5 minutes) at `installing apm`. Lets see how long that goes. 
![image](https://cloud.githubusercontent.com/assets/874898/6539329/824a6ae4-c4c5-11e4-85ce-d2875d233f64.png)

PR #5895, opened 2015-03-07
windows build: potentially common knowledge for node native module builders

basarat:
But worth mentioning that you need an explicit install. Refs https://github.com/atom/atom/issues/5888


paulcbetts:
I don't think this actually true, you don't need to install node-gyp, it comes as part of npm, which is brought in via `script/bootstrap`

thomasjo:
This is indeed false. Brought in by `script/bootstrap` which is in turn invoked by `script/build`.

basarat:
Thanks for the clarification :heart:

PR #5898, opened 2015-03-08
Transpile all .ts files

basarat:
This is similar to https://github.com/atom/atom/pull/5299 

Provide TypeScript support to ease package development.
# Motivation
- Provides built in support for sourcemaps 
- No longer need to check in `.js` files when creating packages using TypeScript 
# Sample

You can see a sample package here : https://github.com/basarat/atom-typescript-test 
image
image

same as other PR 

``` js
global.require('../src/typescript/').getCacheHits()
global.require('../src/typescript/').getCacheMisses()
```
## TODO:
- [x] Add transpiling support
- [x] Add tests

/cc @vvakame @csnover


paulcbetts:
How does cross-compiling with TypeScript work when tsc has to have every file in the project to compile a file? (i.e. TypeScript by-design isn't a simple 1:1 mapping unlike Babel or CoffeeScript, to compile input file X, you need the type definitions of all the types it uses including the other files in the project).

basarat:
> cross-compiling with TypeScript work when tsc has to have every file in the project 

Not essentially. Consider

`foo.ts`
```ts
var foo = 123; 
export = foo;
```
`bar.ts`
```ts
import foo = require('./foo');
console.log(foo); // 123 
```

Each file can be compiled on its own. However file `bar` in the absence of *type information* of `foo` despite being *syntactically* valid, is *semantically* invalid. It would be *semantically valid* only if it had reliable information about `foo.ts`. However the TypeScript compiler *doesn't block JS emit* in the presence of *semantic* errors. 

I have *semantic validation* disabled by passing `false` here : https://github.com/atom/atom/pull/5898/files#diff-1bb8c8dc24534560f9903363ebb9a332R97

basarat:
Also, call sites in TypeScript do not rely on source type information. So absence of type information does not change the emit. This is by design.

basarat:
@paulcbetts I've just pushed a multifile demo : https://github.com/basarat/atom-typescript-test/commit/1dfe56bdc6bf51c867b6bb706086ff11650e165b

The files are compiled separately (in the absence of type information about the other files) and still work : 
image

paulcbetts:
@basarat Aha, now it makes sense. It'd be cool to start making a more generalized framework for loading cross-compilers, because they're going to start hurting startup time and require perf (and hurting require perf is pretty bad, though in the TypeScript case it's a very quick "Does file end in .ts" check)

thedaniel:
> hurting require perf is pretty bad

In general hurting require perf even a little is gonna run counter to this goal (and the 1.0 goal of reducing startup time) in a big way and will need the most convincing of justifications: https://github.com/atom/atom-shell/issues/169

paulcbetts:
@thedaniel I think this can probably be pretty easily done, since these require hooks are so easy to register. We just put it off until we find typescript files

vvakame:
:+1: https://discuss.atom.io/t/how-to-use-the-npmignore/9208/6
https://github.com/vvakame/language-review/tree/master/lib I want to ignore compiled .js code.

basarat:
Also I couldn't get sourcemaps to work because of fancy JS wrapping : https://github.com/TypeStrong/atom-typescript/pull/156#issuecomment-77518310 

Note: I can *get them to work* with more custom code but I'd rather atom do it for me + any other TS package authors.

kevinsawicki:
@basarat So is this something that is ready to review? Or did you still have any planned changes left?

kevinsawicki:
Could you add some tests for this as well? You can model them after the babel specs.

basarat:
@kevinsawicki was ready for review. I'll add the tests tonight and ping you.

basarat:
@kevinsawicki ready for review. Also updated (and marked as completed) the todo list :heart:

spec/typescript-spec.coffee
```diff
@@ -0,0 +1,36 @@
+typescript = require '../src/typescript'
+crypto = require 'crypto'
+
+describe "TypeScript transpiler support", ->
+  beforeEach ->
+    jasmine.snapshotDeprecations()
```
kevinsawicki: I don't think these should be needed since the TypeScript addition shouldn't have any deprecations to start with, unless I'm missing something.

src/typescript-transpile.js
```diff
@@ -0,0 +1,152 @@
+// From https://github.com/teppeis/typescript-simple/blob/master/index.js
```
kevinsawicki: Would you mind converting this file to CoffeeScript with a 2-space indent? We use CoffeeScript in this repository when possible.

If you'd like to keep it as JavaScript then I'd suggest extracting it to a library, publishing it to npm, and then just using it as a library here.

kevinsawicki: After doing so, you can run `script/grunt lint` from the root of the repository to ensure there are no linter errors.

basarat:
@kevinsawicki ready for review again. Did:

* [x] didn't need `jasmine.snapshotDeprecations()` so removed
* [x] The js file was there as a PR was pending. Since been merged so depending on the library `typescript-simple` directly. So js file removed. 
* [x] ran `script/grunt lint` zero errors.

Also I'm now using `JSON.stringify`. Don't see why that wasn't used for babel.

src/typescript-transpile.js
basarat: Removed the js file, there is already a repo that has it (previously was waiting on a PR), so referencing that. Also ran `grunt lint`

spec/typescript-spec.coffee
basarat: Wasn't needed. Removed. Sorry.

package.json
```diff
@@ -66,6 +66,8 @@
     "temp": "0.8.1",
     "text-buffer": "^4.1.5",
     "theorist": "^1.0.2",
+    "typescript": "^1.4.1",
```
kevinsawicki: Is this require used anymore now that typescript-simple is used?

basarat:
@kevinsawicki done:

* [x] remove `typescript` dependency as it is not required explicitly

kevinsawicki:
Thanks for making all these updates, this feels like it is getting really close.

One thing I thought of those was this should be added to the compile-cache helper class that is called directly by apm: https://github.com/atom/atom/blob/11a5b71fb3446ab6eda023fc6606d7509597f71c/src/compile-cache.coffee#L26

And also to https://github.com/atom/atom/blob/11a5b71fb3446ab6eda023fc6606d7509597f71c/spec/compile-cache-spec.coffee#L10

kevinsawicki:
`compile-cache` is called directly by apm when `apm install` is run so that the cache is populated before the package is loaded in Atom so installed packages don't slowdown startup when they are first installed.

basarat:
@kevinsawicki done: 

* [x] add TypeScript to the `compile-cache` + `compile-cache-spec`

mhegazy:
Thanks @basarat, this looks really nice.

Just chiming in, as @basarat mentioned TypeScript errors do not block the generation of JavaScript. we have always thought of the type check errors as warnings that the compiler provides the developer at design time but do not block the rest of the tool chain. 

In general TypeScript does not have emit that depends on global program optimizations, we have steered away from these for the most part. So in Atom's context it should be safe to use typescript as a single file transpiler without loosing correctness. For completeness sake, there are
two places where the compiler relies on the full program information to direct emit, 1. TypeScript internal modules (this is not relevant in atom's case everything is treated as node modules) and 2. use of const enums from .d.ts files (for this i have created an [issue](https://github.com/Microsoft/TypeScript/issues/2499) to ensure that there is an API interface that handles these correctly).

We would like to also add new API for single file transpilation, making [typescript-simple](https://github.com/teppeis/typescript-simple)'s life much simpler (no pun intended :)); and to ensure this scenario is well supported in the future.

It is really nice to see TypeScript plugins in Atom. can not wait to try writing one myself!

src/typescript.coffee
```diff
@@ -0,0 +1,104 @@
...
+fs = require 'fs-plus'
+path = require 'path'
+tss = null # Defer until used
+
+stats =
+  hits: 0
+  misses: 0
+
+defaultOptions =
+  target: 1 # ES5
```
kevinsawicki: Any reason to target ES5 vs. ES6 here?

basarat: Yes. Typescript supports a superset of ES6 futures supported by IO.js

If Typescript spits out ES6 it might not run natively.

kevinsawicki: Oh okay, thanks for the details :+1:

kevinsawicki:
Awesome, thanks so much for adding this :ship:

basarat:
:heart: can't thank you enough for this.

teppeis:
> We would like to also add new API for single file transpilation, making [typescript-simple](https://github.com/teppeis/typescript-simple)'s life much simpler (no pun intended :)); and to ensure this scenario is well supported in the future.

Awesome! I'm waiting.

Issue #274, opened 2013-02-14
Allow panes to be resized


Caged:
Panels split the difference between the space available.  It'd be nice if we could manually resize them.  Should take a look at what @nakajima did with the tree view.

Moter8:
:+1: @yjw9012 Achieved any progress on this?

philipgiuliani:
http://codepen.io/pprice/pen/splkc/ Here a demo of a resizing flexbox :)

360fun:
I'm in the waiting queue too: I really need this feature please!!!

In the meanwhile...is there any package that implements it with mouse support? (like in the @philipgiuliani example)

davidgilbertson:
@yjw9012 How are you progressing? I started using Atom yesterday and love it, this feature is the only thing missing for me. I'd like to try and write it myself (specifically, dragging the pane divider to adjust widths), but if you're almost done I shall wait patiently.

yjw9012:
I've been slow on this because I've been super busy searching for a new job. Finally, I just got a new job, and I am planning to spend significant efforts on the issue.

prescottprue:
Thanks @yjw9012 ! It is one of the few things left that are holding back Atom compared to sublime.

yjw9012:
@davidgilbertson 
I am becoming busy again looking for a new apartment in San Francisco. If you want, you can go ahead and try it.

rmartin:
hey @davidgilbertson ,

I'd be interested in teaming up with you on this. This is a feature that i'm strongly interested in as well and have been reviewing the code to potentially resolve this along with ability to split panes https://github.com/atom/atom/issues/4738 . Let me know if you or others would be interested in collaborating on this as hopefully that will allow us to get this added to the core module sooner than later. I think @philipgiuliani codepen example above is a good start along with the file tree pane view that's already in the build. 

Cheers,
Roy

Caged:
Here's the relevant tree-view code - https://github.com/atom/tree-view/blob/47894ce1ebd68ab7c5c02214d945d7674a3fdfaf/lib/tree-view.coffee#L205-L224

rmartin:
Thanks @Caged, I made some good initial progress and got my head wrapped around the tree view and how they added in the resize handler into the tree view. Thanks for the pointer. 

[...]

PR #5902, opened 2015-03-09
add  implementation for resizable panes, just solve the issue #274

liuxiong332:
### this commit request implement resizable panes and achieve the feature of  #274.

**printscreen**:
resizepane
#### approach

the approach is inserting resizable view (named pane-resize-handle-view) between two panes. when user move the resizable view,  I will change the pane's flexGrow style value by two pane's width or height ratio. for example, if the left pane is  200px and the right pane is 600px, then I will modify flexGrow style to 0.5 and 1.5.
This way is like tree view's resizable view.


src/pane-resize-handle-view.coffee
```diff
@@ -0,0 +1,67 @@
+{$, View} = require 'atom-space-pen-views'
```
benogle: Can you remove all references of view and jQuery? We're no longer using space pen or jQuery in core. In place of jQuery, please use the raw DOM APIs.

benogle:
This looks interesting! A couple things before we can merge:

* Can you please write some specs? 
* Please remove all references to jQuery and `View`

liuxiong332:
I have remove all references to jQeury and `View` @benogle and add specs for pane axis element to make sure the insertion logic of resizable view is correct.
But, I have no idea how to write specs for resizable element. Maybe the good test way is the UI test using some tools such as chrome driver, but now it cannot achieve.

philipgiuliani:
Really cool! Thanks for this :) Finally split-views make sense! :+1: :clap:

src/pane-element.coffee
```diff
@@ -37,8 +38,21 @@ class PaneElement extends HTMLElement
...
 
+    handleDragOver = (event) =>
+      event.preventDefault()
+      event.stopPropagation()
+
+    handleDrop = (event) =>
+      event.preventDefault()
+      event.stopPropagation()
+      @getModel().activate()
+      pathsToOpen = _.pluck(event.dataTransfer.files, 'path')
```
philipgiuliani: You could write i think, so you could remove the dependency to `_` (i just saw they are trying that in new packages currently)...
```coffeescript
pathsToOpen = event.dataTransfer.files.map (file) -> file.path
```
See [array.map](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

liuxiong332: Now I have removed the dependency to _ and replaced with Array::map.
```javascript
pathsToOpen = Array::map.call event.dataTransfer.files, (file) -> file.path
```
Thank you.

rmartin:
hey @liuxiong332 ,

I had a chance to take a look at this today and overall this looks great. The approach I had taken was to integrate this within the resize-pane package, but that had several challenges when purely relying on the callbacks. I like the approach you took here of adding this to core and making the necessary changes to the Pane Axis to adjust the children. I also liked a few additions that I hadn't considered like double-click to balance the frames. 

Overall this looks like a great fix and if this gets integrated into core then I can just shelve the work I was doing on the other package. Looking through your repo, it looks like you've also updated but not yet committed the resize-pane package to accommodated for this new approach, which is great. 

I just had a few comments on my end based on my experience working on that the last few weeks, but nothing that should prevent moving forward with this. 
* Resize Pane Width - When doing research on determining what size this should be, I found Chris Coyier's approach helpful (http://codepen.io/chriscoyier/blog/design-is-hard). My suggestion would be to follow suite there with the height / width of the resize panes and adjust them from the current 8px down to 6px. This is an opinion more than anything but after playing around with them for a while this seemed like the 'right' size. 
* Rebalance Panes After Close - One thing I noted in your solution, that I also struggled with, is that after you resize a pane it 'retains' the flex width value. You took the similar approach that I did by clearing this out for the last element. One suggestion I would have is to 'rebalance' the panes when one of the panes close. This is also likely a minor item that can be improved in the future. Here is one scenario to see this would be to do the following:
** Open a single editor, split right, split right. So you have three total panes. 
** resize the first pane so that it's smaller. 
** close the second and third pane
** split right the first pane. Notice that because the first pane still has a flexbox width that the pane is not 50/50 width instead it retained the old value.
* Spec Tests - One thing to add here would be to actually test a few different scenarios by splitting the panes. That way you can simulate the user scenarios of splitting the pane. Although it needed to be refactored,  you can see this in action within this spec test. https://github.com/rmartin/resize-panes/blob/issue-7-resize-with-mouse/spec/resize-pane-view-spec.coffee

In either case, glad to see that this is getting added and if there is anything that I can do to help get that merged into core let me know. Overall great work.

Cheers,
Roy

src/pane-element.coffee
nathansobo: Did you push this change? I'm still seeing `_` being used at the head of your branch.

philipgiuliani: He removed it in the second PR https://github.com/atom/atom/pull/5910

nathansobo:
@simurai Can you weigh in on the design aspects of this? Note the thickened draggable border that's added between panes.

screenshot 2015-03-16 12 20 42

@liuxiong332 The updated pane size isn't being serialized down into the model, so it gets blown away on refresh. Could you explore adding that? We really want refresh to be a first-class concept so the view state is retained as much as possible when people do it.

spec/pane-axis-element-spec.coffee
```diff
@@ -0,0 +1,58 @@
+PaneAxisElement = require '../src/pane-axis-element'
+
+describe "PaneResizeHandleElement", ->
+  describe "add and remove", ->
+    class TestItemElement extends HTMLElement
+      createdCallback: ->
+        @classList.add('test-root')
+
+    TestItemElement = document.registerElement 'atom-test-item-element', prototype: TestItemElement.prototype
```
nathansobo: I think you should be able to use a generic `div` element to test this rather than registering your own. Adding a custom element here adds complexity and it also can never be registered with the current custom elements API.

spec/pane-axis-element-spec.coffee
```diff
@@ -0,0 +1,58 @@
...
+
+    [paneAxis, model, Model] = []
+    beforeEach ->
+      # This is a dummy item to prevent panes from being empty on deserialization
+      class Model
```
nathansobo: Rather than creating a custom `PaneAxis` model just for this test, I'd prefer to just use the real model in the specs. If you want to keep the specs focused on `PaneAxisElement`, you should be able to construct a `PaneAxis` model directly.

spec/pane-axis-element-spec.coffee
```diff
@@ -0,0 +1,58 @@
...
+        onDidReplaceChild: (@replaceCallback) ->
+        getOrientation: -> 'horizontal'
+        getChildren: -> []
+
+      paneAxis = document.createElement('atom-pane-axis')
+      model = new Model
+      paneAxis.initialize(model)
+      document.querySelector('#jasmine-content').appendChild(paneAxis)
+
+    it "should insert the correct postion in one pane axis", ->
```
nathansobo: how about `"inserts draggable resize elements between pane axis children"`

spec/pane-axis-element-spec.coffee
nathansobo: On second thought, I think this might work better as an integration test on `PaneContainer` or even by enhancing existing tests that do splitting to add assertions about the draggable handles between axis children. Also, we need some tests that actually tested the resize functionality.

src/pane-element.coffee
nathansobo: I really would like to merge this stuff, but I need a single consolidated PR with good test coverage first. Can someone else from the community collaborate on the test side of things?

simurai:
> Note the thickened draggable border that's added between panes.

Wouldn't be a huge issue, but to keep most themes visually untouched, I made a PR on top of this one: https://github.com/liuxiong332/atom/pull/1 It removes the "thickened draggable border" and instead the `atom-pane-resize-handle` overlap the editors on each side a bit. Themes could still override it if the space between fits better.

resize-pane

philipgiuliani:
Looks really good @simurai :) Already happy to get this one merged!

liuxiong332:
Looks good @simurai . Thank you!

pitaj:
I implemented something like this as a temporary fix before this is merged

https://github.com/pitaj/atom-resize-panes/

liuxiong332:
@nathansobo I have added  specs for the resize functionality in `pane-resize-handle-element-spec.coffee`. Please review it, thank you.

zephraph:
Hey, great work @liuxiong332! This is looking awesome!

rmartin:
Hey @nathansobo ,

Is there anything else here that needs to be adjusted before this is merged in?

Cheers,
Roy

liuxiong332:
I also want to know what else I can do to make this PR be merge in.

nathansobo:
Overall, really great job on this. It's looking pretty close, but the language in the specs need to conform to the [specs style guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#specs-styleguide) before we can merge. Also, do we have test coverage for everything added to the implementation? It seems like we don't cover serialization.

spec/pane-axis-element-spec.coffee
```diff
@@ -0,0 +1,46 @@
+PaneAxisElement = require '../src/pane-axis-element'
+PaneAxis = require '../src/pane-axis.coffee'
+
+describe "PaneResizeHandleElement", ->
```
nathansobo: Why do we have `PaneResizeHandleElement` both here and in another file? Could both be integration tested in a `pane-container-element-spec`? I realize the specs in this area are kind of a mess because they've never been ported away from deprecated SpacePen.

spec/pane-resize-handle-element-spec.coffee
```diff
@@ -0,0 +1,131 @@
...
+
+      resizeElementMove(resizeElements[0], getElementWidth(paneElements[0]) / 2)
+      expectPaneScale(0.5, 1.5, 1)
+
+      clientX = getElementWidth(paneElements[0]) + getElementWidth(paneElements[1]) / 2
+      resizeElementMove(resizeElements[1], clientX)
+      expectPaneScale(0.5, 0.75, 1.75)
+
+    it "drag the horizontal element, the size of other vertical pane will not change", ->
+      upPane = container.getActivePane()
```
nathansobo: `upperPane` and `lowerPane` would be clearer

liuxiong332:
Hey, @nathansobo 
I have completed the following changes.
1. integrate the specs for `PaneResizeHandleElement` to `pane-container-element-spec.coffee`.
2. change the description of specs to conform to the specs style guide.
3. simplify the specs to make it clear and easy to understand.
4. fix some errors and conflicts.

spec/pane-container-element-spec.coffee
```diff
@@ -0,0 +1,110 @@
+PaneContainer = require '../src/pane-container'
+PaneAxisElement = require '../src/pane-axis-element'
+PaneAxis = require '../src/pane-axis'
+
+describe "PaneResizeHandleElement", ->
+  describe "when add or remove panes in PaneAxisElement", ->
```
50Wliu: >when panes are added or removed

spec/pane-container-element-spec.coffee
```diff
@@ -0,0 +1,110 @@
...
+    [paneAxisElement, paneAxis] = []
+
+    beforeEach ->
+      paneAxisElement = document.createElement('atom-pane-axis')
+      paneAxis = new PaneAxis({})
+      paneAxisElement.initialize(paneAxis)
+      document.querySelector('#jasmine-content').appendChild(paneAxisElement)
+
+    it "inserts or remove resize elements", ->
```
50Wliu: >inserts or removes resize elements

spec/pane-container-element-spec.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+      paneAxis.addChild(models[2])
+      expectPaneAxisElement(i) for i in [0, 2, 4]
+      expectResizeElement(i) for i in [1, 3]
+
+      # test removeChild
+      paneAxis.removeChild(models[2])
+      expectResizeElement(i) for i in [1]
+
+  describe "when user drag the resize element", ->
```
50Wliu: >when the resize element is dragged

spec/pane-container-element-spec.coffee
```diff
@@ -0,0 +1,110 @@
...
+      getElementWidth = (element) ->
+        element.getBoundingClientRect().width
+
+      # assert the pane's flex scale. arguments is list of pane-scale pair
+      expectPaneScale = ->
+        args = Array::slice.call(arguments, 0)
+        for paneScale in args
+          expect(paneScale[0].getFlexScale()).toBeCloseTo(paneScale[1], 0.1)
+
+    it "add and remove panes in the same direction", ->
```
50Wliu: >adds and removes panes in the direction that the pane is being dragged
Not sure about this one, but from the context of the describe it might work

spec/pane-container-element-spec.coffee
```diff
@@ -0,0 +1,110 @@
...
+      resizeElementMove(resizeElements[1], clientX)
+      expectPaneScale [leftPane, 0.5], [middlePane, 0.75], [rightPane, 1.75]
+
+      middlePane.close()
+      expectPaneScale [leftPane, 0.44], [rightPane, 1.55]
+
+      leftPane.close()
+      expectPaneScale [rightPane, 1]
+
+    it "split or close panes in orthogonal direction", ->
```
50Wliu: >splits or closes adjacent panes

Again, not sure if this is accurate

spec/pane-container-element-spec.coffee
liuxiong332: Yes, exactly. Thank you.

spec/pane-container-element-spec.coffee
liuxiong332: I just want to test that the scale of horizontal panes is not changed when the vertical panes are added or removed.

maxbrunsfeld:
Thanks a lot @liuxiong332. A lot of people will be excited to have this feature.

philipgiuliani:
Whaaat its mergeeed :boom: Thanks @liuxiong332 for working on this!! :)

thomasjo:
Awesome :zap::exclamation:

benogle:
:tada: :tada: Pulling and building...

rmartin:
Yahoo! well done @liuxiong332

liuxiong332:
Awesome!1

JordiOrlando:
Wow, fantastic work @liuxiong332!

simurai:
:heart_eyes_cat:

mmcbride1007:
:clap: so pumped about this. awesome work @liuxiong332. thank you!

ajwhite:
:clap: thanks @liuxiong332!

trkoch:
Thanks @liuxiong332! One minor issue: The resize icon is not the same that appears when resizing the file drawer.

nathansobo:
> One minor issue: The resize icon is not the same that appears when resizing the file drawer.

/cc @simurai on this slight inconsistency.

chrishough:
I noticed this recently broke in the latest version of atom.  I can no longer use cmd+alt-- to shrink or expand windows, thoughts?

mmcintyre123:
Can you add this to atom in ubuntu?  I'm running on ubuntu 14.04 LTS and atom 1.0.2 and none of the panes are resizable.  Also commented on #274. Thanks!

Issue #4049, opened 2014-11-04
Open drag&dropped file in correct split view


alexandernst:
Atom should open files in the split view a file was dropped into. It currently opens the files in the last used split view iirc.

mnquintana:
Implemented in https://github.com/atom/atom/pull/5902

PR #5910, opened 2015-03-09
this commit make Atom open files in the pane view that files was dropped into

liuxiong332:
- invalidate document's drop event to prevent the browser's default behavior that open the file URL.
- this commit fix bug #4049. Every pane element will listen `drop` event to make drop action take in this pane view, therefore the drop files will open in the correct split view.

### DIFF
--- a/src/pane-axis-element.coffee
+++ b/src/pane-axis-element.coffee
@@ -1,5 +1,6 @@
 {CompositeDisposable} = require 'event-kit'
 {callAttachHooks} = require './space-pen-extensions'
+PaneResizeHandleView = require './pane-resize-handle-view'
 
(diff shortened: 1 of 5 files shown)


liuxiong332:
**printscreen** 
atom-drop

Zireael07:
Whee!

mnquintana:
@liuxiong332 Would you mind merging in / rebasing master and fixing the merge conflicts?

liuxiong332:
@mnquintana this commit has merged into origin through #5902. So this PR is not used.

mnquintana:
@liuxiong332 Oh awesome to hear! :+1:

PR #5923, opened 2015-03-10
add package manager for install Atom on Windows

gutierri:
An alternative to Chocolatey


izuzak:
Thanks for suggesting this :tophat:, but we document things we don't officially support. Sharing this on the forum would be neat, though: https://discuss.atom.io/

Issue #5882, opened 2015-03-06
Join lines into one long line cases Atom slow or hangs


hokein:
Type `cmd+j` short-cut to join multiple lines into one line, and as the line gets longer, Atom will become slower, even hangs.

This happens on highlight-switch on, while I changed the file type to `txt` and repeated the same steps, Atom seems work much better. The test file is [here](https://gist.github.com/hokein/48a9ed680d69fbf25305).

This may also related to #979.

izuzak:
Another observation: if soft wrap is turned on, then joining these lines takes forever (I waited for over a few minutes). However, if you turn off soft wrap, it's much faster. And if you turn on soft wrap after after joining lines -- it's instant.

Issue #979, opened 2013-10-16
Files with large extremely long lines cause hangs


bnferguson:
May be a known issue, but files with large embedded images such as github's public/enterprise/maintenance.html will often cause Atom to freeze (MacVim actually struggles as well). 

Seems some of them (github's public/maintenance.html for instance) work fine when in soft wrap mode, or until you move your cursor to the long line when out of soft wrap.

benogle:
Via @tdegrunt in #5270 

> The following file will not open with Atom, but Brackets (and SublimeText, etc) will open the file instantly.
It's likely due to the very long line (with strange characters), but still ...

https://gist.github.com/tdegrunt/1404dc0c35be458bdb59

ilanbiala:
@benogle even the web editor renders it very quickly.

hokein:
I have done some investigations on this issue. Below are some details.

The benchmark runs with two test [files](https://gist.github.com/hokein/48a9ed680d69fbf25305)(test-oneline.js, test-multiple.js), which are the same except line organization. The test-oneline.js is a line containing about 150000 characters.

**Environment**

* Matchine: MacBook Pro on OS X 10.9.4 with 2.4GHz Intel Core i5, 8GB RAM, 256G SSD.
* Atom version: 0.179.0

**Steps**

1. Open Atom's dev-tools, switch to Profiles tab and click `Start` button to start profiling.
2. Open the test file in Atom, wait until finishing.
3. Click `Stop` button to stop profiling.

**Result**

File | Parser Time | Total Time
----- | -----------------|----------------
test-oneline.js |21s | 22.7s
test-multiple.js | 1.8s | 2.7s

I also run the same benchmark in [first-mate](https://github.com/atom/first-mate) directly under nodejs v0.11.13, it gave similar results:

![image](https://cloud.githubusercontent.com/assets/2557445/6526123/39971738-c449-11e4-9cad-75a220f7650e.png)

Below are the detailed profile charts in dev-tools.

test-oneline.js:
![image](https://cloud.githubusercontent.com/assets/2557445/6526114/1b2c1ce4-c449-11e4-9cac-3258ff12509e.png)

test-multipleline.js

![image](https://cloud.githubusercontent.com/assets/2557445/6526132/66ccc00e-c449-11e4-9a58-7c8c5cf7e572.png)

[...]

PR #5931, opened 2015-03-11
Faster lines joining without soft-wrap

as-cii:
This partially addresses #5882, read on to understand why it doesn't fully solve it.
### Benchmark

_Benchmark performed using the `test-multiple-lines.js` file found on the aforementioned issue with soft-wrap disabled._
#### Steps
- Select all
- Hit Cmd+J (`core:join-lines`)
#### Before :chart_with_downwards_trend:

screen shot 2015-03-11 at 11 35 04
#### After :chart_with_upwards_trend:

screen shot 2015-03-11 at 11 36 12

On my machine this is almost 3x faster :racehorse: :racehorse: :racehorse: 
### Considerations

Most of the time was spent in autoscrolling (which is something we can do at the end of `joinLines`) and in updating the UI (some operations were not batched). Now time is spent mostly on scanning tokens, which is something that needs to be addressed in another layer (see #979).

screen shot 2015-03-11 at 11 46 33

The situation is dramatically different for soft-wrapped lines. For this particular scenario, in facts, there's a significant amount of throwaway computation involved. When multiple lines are joined together, the operation does not happen atomically but causes `DisplayBuffer` to continuously update the first screen line as more and more lines are joined into the first one. This is quite cheap when soft wrapping is off, but it becomes very expensive when we turn it on:

screen shot 2015-03-11 at 12 02 40

:fearful:
- One solution could be to make `updateScreenLines` smarter, and reuse previous screen lines as a starting point for updates. However, that would involve a radical change in how we tokenize lines and I feel like this is an improvement that we should make at a lower layer, e.g. while addressing #979 
- Another way to tackle this would be to start batching `DisplayBuffer#handleTokenizedBufferChange` on `TextEditor#transact`. On one hand I find this quite smart, but on the other hand there's a lot of event-handling involved (e.g. markers) and I am afraid this change could break many things. Based on my first trials, I would say it will. @nathansobo: batching, batching everywhere :scream: :smile: 
- Alternatively, all those methods like `Selection#joinLines` could directly operate on the buffer, as if an external program performs the action, and `DisplayBuffer` sees only the final result. This is similar to the solution provided above, but addresses the issue from a different perspective, probably breaking abstraction.

Personally, I feel like the second solution provides quite a lot of benefits without breaking abstractions or changing how we deal with tokenized line. What do you think?

### DIFF
--- a/src/cursor.coffee
+++ b/src/cursor.coffee
@@ -54,6 +54,7 @@ class Cursor extends Model
       @emitter.emit 'did-destroy'
       @emitter.dispose()
     @needsAutoscroll = true
+    @autoscrollEnabled = true
 
   destroy: ->
     @marker.destroy()
@@ -669,7 +670,14 @@ class Cursor extends Model
     {row, column} = @getScreenPosition()
     new Range(new Point(row, column), new Point(row, column + 1))
 
-  autoscroll: (options) ->
+  autoscroll: (options, fn) ->
+    return unless @autoscrollEnabled
+
+    if fn?
+      @autoscrollEnabled = false
+      fn()
+      @autoscrollEnabled = true
+
     @editor.scrollToScreenRange(@getScreenRange(), options)
 
   getBeginningOfNextParagraphBufferPosition: ->
--- a/src/cursor.coffee
+++ b/src/cursor.coffee
@@ -673,6 +673,10 @@ class Cursor extends Model
   autoscroll: (options, fn) ->
     return unless @autoscrollEnabled
 
+    if typeof options is "function"
+      fn = options
+      options = fn
+
     if fn?
       @autoscrollEnabled = false
       fn()
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -487,47 +487,48 @@ class Selection extends Model
   #
   # If there selection spans more than one line, all the lines are joined together.
   joinLines: ->
-    selectedRange = @getBufferRange()
-    if selectedRange.isEmpty()
-      return if selectedRange.start.row is @editor.buffer.getLastRow()
-    else
-      joinMarker = @editor.markBufferRange(selectedRange, invalidationStrategy: 'never')
+    @cursor.autoscroll {}, =>
+      selectedRange = @getBufferRange()
+      if selectedRange.isEmpty()
+        return if selectedRange.start.row is @editor.buffer.getLastRow()
+      else
+        joinMarker = @editor.markBufferRange(selectedRange, invalidationStrategy: 'never')
 
-    rowCount = Math.max(1, selectedRange.getRowCount() - 1)
-    for row in [0...rowCount]
-      @cursor.setBufferPosition([selectedRange.start.row])
-      @cursor.moveToEndOfLine()
+      rowCount = Math.max(1, selectedRange.getRowCount() - 1)
+      for row in [0...rowCount]
+        @cursor.setBufferPosition([selectedRange.start.row])
+        @cursor.moveToEndOfLine()
 
-      # Remove trailing whitespace from the current line
-      scanRange = @cursor.getCurrentLineBufferRange()
-      trailingWhitespaceRange = null
-      @editor.scanInBufferRange /[ \t]+$/, scanRange, ({range}) ->
-        trailingWhitespaceRange = range
-      if trailingWhitespaceRange?
-        @setBufferRange(trailingWhitespaceRange)
+        # Remove trailing whitespace from the current line
+        scanRange = @cursor.getCurrentLineBufferRange()
+        trailingWhitespaceRange = null
+        @editor.scanInBufferRange /[ \t]+$/, scanRange, ({range}) ->
+          trailingWhitespaceRange = range
+        if trailingWhitespaceRange?
+          @setBufferRange(trailingWhitespaceRange)
+          @deleteSelectedText()
+
+        currentRow = selectedRange.start.row
+        nextRow = currentRow + 1
+        insertSpace = nextRow <= @editor.buffer.getLastRow() and
+                      @editor.buffer.lineLengthForRow(nextRow) > 0 and
+                      @editor.buffer.lineLengthForRow(currentRow) > 0
+        @insertText(' ') if insertSpace
+
+        @cursor.moveToEndOfLine()
+
+        # Remove leading whitespace from the line below
+        @modifySelection =>
+          @cursor.moveRight()
+          @cursor.moveToFirstCharacterOfLine()
         @deleteSelectedText()
 
-      currentRow = selectedRange.start.row
-      nextRow = currentRow + 1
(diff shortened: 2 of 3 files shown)


as-cii:
I am also thinking that we should address how `hasPairedCharacters`, or other string-related methods, work: maybe we may even stop using `String` as the base for text operations, and start working on something more specialized instead.

as-cii:
Speaking about batching, I feel like we could manage this straight inside `TextBuffer`. @nathansobo: do you see any drawback in emitting events only at the end of a transactional operation? This way it would be much easier on the other side (e.g. `DisplayBuffer`) to elide duplicate operations (e.g. two events that act on the first row). We already `pauseChangeEvents` on `TextBuffer#applyPatch` but I think we could push this further, maybe resuming change events only at the end of a transactional operation and emitting some more events, e.g. `onWillTransact`, `onDidTransact`. Not sure how much troublesome this may be, though.

`onDidStopChanging` performs a similar job, but I wouldn't rely on it as it is asynchronous.

thomasjo:
Wow! :scream:
Looks super promising.

src/cursor.coffee
```diff
@@ -669,7 +670,14 @@ class Cursor extends Model
     {row, column} = @getScreenPosition()
     new Range(new Point(row, column), new Point(row, column + 1))
 
-  autoscroll: (options) ->
+  autoscroll: (options, fn) ->
```
nathansobo: Can we rig this to make the `options` parameter optional?

src/selection.coffee
```diff
@@ -487,47 +487,48 @@ class Selection extends Model
...
-      @cursor.moveToEndOfLine()
-
-      # Remove trailing whitespace from the current line
-      scanRange = @cursor.getCurrentLineBufferRange()
-      trailingWhitespaceRange = null
-      @editor.scanInBufferRange /[ \t]+$/, scanRange, ({range}) ->
-        trailingWhitespaceRange = range
-      if trailingWhitespaceRange?
-        @setBufferRange(trailingWhitespaceRange)
+    @cursor.autoscroll {}, =>
```
nathansobo: Do you think we could perform the line join as a single change operation on the buffer? You would have to keep more local state in the method but it would reduce the change events dramatically. Batching everything in the model would be a more universal solution, but it's not a simple change. I'll add some more thoughts about that subject to the PR in a bit. In the meantime, seems like combining here could be a win.

nathansobo:
@as-cii Could you explain about the changes you're making to the presenter in this PR. I'm sure it's all justified, but it would help my understanding if you explained it at a high level so I don't have to derive it all from the diff.

> do you see any drawback in emitting events only at the end of a transactional operation?

I actually am pretty concerned about doing this. One benefit of a synchronous, per-change event is that the state of the buffer is well understood both before and after the change. If we batched multiple changes together, we'd start to run into the same kind of issues that make the `Object.observe` API a bit awkward to work with. For any given change event, it wouldn't be clear whether subsequent change events invalidated it in some way. I like that we know that at the time of the event, the state of the buffer reflects only a single change, which is fully described by the event object.

That said, I'm not totally opposed to exploring in this direction... maybe there are solutions to my concerns. For example, we could easily roll *contiguous* changes up into a single event. Or we could pretend that non-contiguous changes were actually a single big change with some of the content in between repeated.

I'm reluctant to make a bunch of changes in `TextBuffer` and `DisplayBuffer` at this point however, because we plan on replacing them with `TextDocument` before 1.0. I know you're interested in helping out with that, and at this point I'm pretty confident that the new `text-document` repo will be the winner. @maxbrunsfeld and I are still ironing out the basic primitives, but once we get the lines layer fully operational there will probably more opportunities for parallel development.

For now, what do you think about batching just the line join operation in a situation-specific way for now?

as-cii:
> Could you explain about the changes you're making to the presenter in this PR. I'm sure it's all justified, but it would help my understanding if you explained it at a high level so I don't have to derive it all from the diff.

@nathansobo: sure, sorry about that. Here's a more thorough explanation of what I changed (and why I changed it) in the presentation layer.

Basically I started batching some operations that were not batched before. Specifically I noticed that, when `TextEditor` needs to perform a lot of tasks (which involve change events), smart updates in the presenter, such as updating only one cursor when it changed, were way too much expensive and our coarse grained batching strategy outperformed them quite significantly.

screen shot 2015-03-15 at 12 39 16

For this reason, I opted to:

* Avoid updating each cursor whenever it changes: I only track that we should update cursors on next `getState`;
* Avoid updating each highlightState whenever some decoration changes: I only track that we should update decorations on next `getState`, along with some specific information about flashes;

screen shot 2015-03-15 at 12 39 30

------------------------

> I'm reluctant to make a bunch of changes in TextBuffer and DisplayBuffer at this point however, because we plan on replacing them with TextDocument before 1.0.

That's definitely a nice point you bring up: we already planned to replace `TextBuffer` with `TextDocument`, therefore it's definitely reasonable to wait for it to be rolled out before changing how the current implementation works.

> I know you're interested in helping out with that, and at this point I'm pretty confident that the new text-document repo will be the winner. @maxbrunsfeld and I are still ironing out the basic primitives, but once we get the lines layer fully operational there will probably more opportunities for parallel development.

Great job there! :fire:

> Do you think we could perform the line join as a single change operation on the buffer? You would have to keep more local state in the method but it would reduce the change events dramatically.

Yeah, I think this matches with the third option described in this PR's description, where I basically suggested to collect changes right inside `joinLines`: as I pointed out, though, my biggest concern in doing that is breaking abstraction. Right now the method fundamentally acts as a *user recorded macro* (selects portions of text, deletes them, moves cursors, etc.), whereas if we keep local state we'd have to deal with strings instead. As always, I believe it's a matter of tradeoffs. However, we don't know yet how far we will get with the new `TextDocument`, thus any performance improvement we can think of today will be probably useless with the new buffer. That said, what do you think if we simply refine and merge this as it is, and address the remaining issues once `TextDocument` gets released?

Thanks :bowtie:

mark-hahn:
Are new things like TextDocument discussed anywhere?  I'd like to get a
peek.

On Sun, Mar 15, 2015 at 6:12 AM, Antonio Scandurra <notifications@github.com

as-cii:
> Are new things like TextDocument discussed anywhere?  I'd like to get a peek.

@mark-hahn: there isn't a real discussion going on, because @nathansobo and @maxbrunsfeld are building the foundations first, but you can find a few attempts to implement a new buffer here:

* https://github.com/atom/text-transform
* https://github.com/atom/text-document (this should be the *good one*)

nathansobo:
Where is this at? Should we catch this up and merge it?

as-cii:
> Where is this at? Should we catch this up and merge it?

I should benchmark this specific case again to see if anything has changed, but I think we can close this for now as I would like to address the problem with a more radical solution such as batching in the `DisplayBuffer` layer.

Thanks for the heads-up! :bow:

PR #5936, opened 2015-03-11
Destroy only containing folds on selection

as-cii:
As reported by @nathansobo on #5858:

> I'm fine with leaving the optimization in, although extra bonus :christmas_tree: would be to only destroy folds containing a selection when the selection changes. I don't think selectAll should destroy anything. If you get inspired I would love to remove the optimization combined with that behavior change.

:fire:

Any feedback or suggestion is very welcome, as usual. Thank you! :bow: 

/cc: @izuzak


nathansobo:
:surfer: The waves of @as-cii goodness keep crashing in.

izuzak:
> The waves of @as-cii goodness keep crashing in.

:joy: :clap:

as-cii:
> The waves of @as-cii goodness keep crashing in.

haha thank you! :smile:

PR #5942, opened 2015-03-11
Update documentation to point to atom/docs

schacon:
All these docs have been moved to the atom/docs repository, integrated into the Flight Manual.


izuzak:
https://github.com/atom/atom/blob/master/CONTRIBUTING.md has a few links to the docs which we should update (debugging guide link and link to contributing to atom packages docs)

schacon:
@izuzak technically those links will still work, they'll just fall back to older versions of the guides. I tried not to break any existing URLs. However, I should find where they moved to in the book and update the URLs, so I'll do that. Thanks!

PR #5951, opened 2015-03-12
Adjust `DisplayBuffer#getVisibleRowRange` logic

as-cii:
Supersedes and closes #4597

Thanks to @izuzak, @neojski and @nathansobo for reporting the issue, providing an alternative algorithm and bringing the issue forward: you are great guys! :clap: 

As usual, an additional pair of :eyes: is most welcome :bow:

### DIFF
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -348,12 +348,13 @@ class DisplayBuffer extends Model
   getScrollWidth: ->
     @scrollWidth
 
+  # Returns an {Array} of two numbers representing a closed interval of visible rows.
   getVisibleRowRange: ->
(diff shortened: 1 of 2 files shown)


src/display-buffer.coffee
```diff
@@ -348,12 +348,13 @@ class DisplayBuffer extends Model
   getScrollWidth: ->
     @scrollWidth
 
+  # Returns an {Array} of two numbers representing a closed interval of visible rows.
```
as-cii: Here I simply avoided to write about 0-based indexes, as everything is 0-based.

nathansobo:
Thanks so much @as-cii. I really appreciate your input and the thorough, careful thinking you put into everything you do.

PR #5959, opened 2015-03-13
Use screen ranges to select above and below

### LINKED ISSUE #5788: Weird vertical text selection behaviour
optimised

I used alt + shift + arrow up/down
Syntax perl
```perl
my $sth = $dbh->prepare(q~
	SELECT
		TableA.id																AS TaskObserverId
		,TableA.i_user_id														AS TaskObserverUserId
		,TableB.s_displayname													AS UserDisplayName
		,TableB.s_email															AS UserEMail
	FROM
		TableA WITH(NOLOCK)
		LEFT OUTER JOIN TableB WITH(NOLOCK)ON TableA.id = TableB.i_user_id
~);
```

Atom 0.184.0 on Ubuntu 14.10 x64

as-cii:
This fixes #5788 

As reported on the issue, Atom behaved inconsistently while adding selections (above or below) on lines made of tokens with a `screenDelta > bufferDelta` (e.g. tabs). The behavior was not compatible with Sublime Text, too, which handles the scenario like this:
- Selecting above or below works with what's visible on screen (thus, no mismatch between buffer length and screen length)
- Soft-wrapped lines are therefore included while selecting
- When a soft-wrapped line has indentation, skip it and select the first real character

This is the behavior I would expect in Atom as well and, hopefully, this PR includes everything that's needed to achieve it :fire: 

An additional pair of :eyes: is most welcome, but even if code-wise this looks good, I'd prefer to leave this open for a while so that the community has the chance to test-drive it.

Thank you! :bow: 

/cc: @benogle @izuzak @nathansobo @kevinsawicki

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -1635,6 +1635,19 @@ describe "TextEditor", ->
             [[7, 20], [7, 25]]
           ]
 
+        it "takes atomic tokens into account", ->
+          waitsForPromise ->
+            atom.project.open('sample-with-tabs-and-leading-comment.coffee', autoIndent: false).then (o) -> editor = o
+
+          runs ->
+            editor.setSelectedBufferRange([[2, 1], [2, 3]])
+            editor.addSelectionBelow()
+
+            expect(editor.getSelectedBufferRanges()).toEqual [
+              [[2, 1], [2, 3]]
+              [[3, 1], [3, 2]]
+            ]
+
       describe "when the selection is empty", ->
         it "does not skip soft-wrapped lines shorter than the current column", ->
           editor.setSoftWrapped(true)
@@ -1722,6 +1735,19 @@ describe "TextEditor", ->
             [[7, 20], [7, 25]]
             [[6, 20], [6, 25]]
           ]
+
+        it "takes atomic tokens into account", ->
+          waitsForPromise ->
+            atom.project.open('sample-with-tabs-and-leading-comment.coffee', autoIndent: false).then (o) -> editor = o
+
+          runs ->
+            editor.setSelectedBufferRange([[3, 1], [3, 2]])
+            editor.addSelectionAbove()
+
+            expect(editor.getSelectedBufferRanges()).toEqual [
+              [[3, 1], [3, 2]]
+              [[2, 1], [2, 3]]
+            ]
 
       describe "when the selection is empty", ->
         it "does not skip soft-wrapped lines shorter than the current column", ->
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -1456,8 +1456,8 @@ describe "TextEditor", ->
         expect(editor.getSelectedBufferRanges()).toEqual [[[5, 5], [6, 6]]]
 
       it "merges intersecting selections", ->
-        editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 0], [5, 5]]])
-        expect(editor.getSelectedBufferRanges()).toEqual [[[2, 2], [5, 5]]]
+        editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 0], [5, 4]]])
+        expect(editor.getSelectedBufferRanges()).toEqual [[[2, 2], [5, 4]]]
 
       it "does not merge non-empty adjacent selections", ->
         editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 3], [5, 5]]])
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -1623,16 +1623,15 @@ describe "TextEditor", ->
             [[6, 22], [6, 28]]
           ]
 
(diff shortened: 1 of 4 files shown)


nathansobo:
Sorry @as-cii that I've been having trouble keeping up with you this week. We've all been in San Francisco for a summit so I've gotten behind. Next week I will be way more responsive and we'll get all these improvements merged.

spec/text-editor-spec.coffee
```diff
@@ -1623,7 +1623,56 @@ describe "TextEditor", ->
             [[6, 22], [6, 28]]
           ]
 
+        it "selects also soft-wrapped lines", ->
```
nathansobo: :bulb: `"can add selections to soft-wrapped line segments"`

spec/text-editor-spec.coffee
```diff
@@ -1623,7 +1623,56 @@ describe "TextEditor", ->
...
+            expect(editor.getSelectedBufferRanges()).toEqual [
+              [[2, 1], [2, 3]]
+              [[3, 1], [3, 2]]
+            ]
+
       describe "when the selection is empty", ->
+        describe "when lines are soft-wrapped", ->
+          beforeEach ->
+            editor.setSoftWrapped(true)
+            editor.setDefaultCharWidth(10)
```
nathansobo: Setting the default character width is actually messing up the subsequent call to `setEditorWidthInChars` and causing it not to work. So basically none of the lines are actually soft-wrapping in this test.

src/selection.coffee
```diff
@@ -183,7 +183,7 @@ class Selection extends Model
 
   # Public: Clears the selection, moving the marker to the head.
   clear: ->
-    @marker.setProperties(goalBufferRange: null)
+    @marker.setProperties(goalBufferRange: null, goalScreenRange: null)
```
nathansobo: What's your thinking behind retaining the `goalBufferRange` here now that you've added `goalScreenRange`? Are we still using it in other ways?

spec/text-editor-spec.coffee
as-cii: Nice catch. Thank you @nathansobo, you rock! :guitar:

src/selection.coffee
as-cii: That's totally reasonable. The only method using it is `merge`, which can be refactored to work directly with screen ranges as well. I'll explain this refactoring a bit more thoroughly in a comment later on :+1:

as-cii:
> Sorry @as-cii that I've been having trouble keeping up with you this week. We've all been in San Francisco for a summit so I've gotten behind. Next week I will be way more responsive and we'll get all these improvements merged.

No need to apologize, @nathansobo. To be honest, I really appreciate you dedicated some of your time to answer me during this week, so thank you for that :bow: 

> What's your thinking behind retaining the goalBufferRange here now that you've added goalScreenRange? Are we still using it in other ways?

As reported on another comment, apparently we were using it only on `Selection#merge`: I adapted this method to use screen ranges instead, so that I could get rid of `goalBufferRange` entirely. One spec failed because of clipping issues (I'll report the original version here for the sake of clarity):

```coffee
it "merges intersecting selections", ->
  editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 0], [5, 5]]])
  expect(editor.getSelectedBufferRanges()).toEqual [[[2, 2], [5, 5]]]
```

After changing the method this returned `[[[2, 2], [5, 4]]]`: since we use screen ranges now, `[5, 5]` gets clipped to `[5, 4]` thus causing it to fail. How do we want to deal with this? UX-wise I find it a bit confusing.


#### Using buffer ranges (original behavior)

```coffee
editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 0], [5, 5]]])
editor.getSelectedBufferRanges()) #=> [[[2, 2], [5, 5]]]
```

screen shot 2015-03-14 at 09 30 43

#### Using screen ranges

```coffee
editor.setSelectedBufferRanges([[[2, 2], [3, 3]], [[3, 0], [5, 5]]])
editor.getSelectedBufferRanges()) #=> [[[2, 2], [5, 4]]]
```

screen shot 2015-03-14 at 09 34 03

To the user they present in the same exact way, but they're actually selecting two different things. On one hand I'd say we should keep the original behavior, since selecting a buffer range should not have anything to do with clipping. On the other hand, selections are a user-facing feature and, as pointed out earlier on, I find the second behavior more consistent.

What do you think? For now I sticked with the original behavior (e.g. use buffer ranges while merging), but if we choose to go with screen ranges instead, I can simply revert the last commit and we should be good to go :sunglasses:

nathansobo:
So if I understand this situation correctly, it seems like we aren't clipping `[5, 5]` to `[5, 4]` until the selection is merged. But `[5, 5]` always corresponds to an invalid location due to soft-tabs, right? The most consistent thing to do would probably be to always clip ranges when creating selections even via buffer coordinates. But in absence of that, clipping it only when merging does seem a bit surprising. I think your compromise is the correct one for now. Ideally I'd like to always clip but this whole layer is changing soon so we can keep going.

PR #5960, opened 2015-03-13
Enable multiple users

Bjoernsen:
Currently, it is not possible to use atom for multiple users on the same linux machine. The socket file '/tmp/atom.sock' belongs to the first user that starts atom.
With this small change, a socket file will be created for each user.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -18,7 +18,7 @@ DefaultSocketPath =
   if process.platform is 'win32'
     '\\\\.\\pipe\\atom-sock'
   else
-    path.join(os.tmpdir(), 'atom.sock')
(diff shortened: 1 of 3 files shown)


src/browser/atom-application.coffee
```diff
@@ -18,7 +18,7 @@ DefaultSocketPath =
   if process.platform is 'win32'
     '\\\\.\\pipe\\atom-sock'
   else
-    path.join(os.tmpdir(), 'atom.sock')
+    path.join(os.tmpdir(), "atom_#{process.env['USER']}.sock")
```
kevinsawicki: Really minor, but this could be `process.env.USER` which is a little simpler

Bjoernsen:
@kevinsawicki Thank you for your hint.

kevinsawicki:
Thanks for this

PR #5969, opened 2015-03-14
Add pane:close to atom-pane context menu

mnquintana:
I've found myself wanting to do this quite frequently – this PR adds "Close Pane" to `atom-pane`'s context menu, allowing you to close the active pane from the context menu just like you can do with the "Split" pane commands.


kevinsawicki:
:+1: nice addition

PR #5972, opened 2015-03-15
Use core:cancel command to dissmiss notifications

olmokramer:
Listen to `core:cancel` on `atom-workspace`. When triggered dismiss notifications.


50Wliu:
:+1:

benogle:
I think this is interesting, but I think it should happen in the notifications package.

olmokramer:
Ok, I'll move this PR to the notifications package then. Closing this one.

PR #6007, opened 2015-03-17
:racehorse: Don't tokenize on NullGrammar

octref:
For file format that Atom doesn't recognize as a
programming language, such as log files,
skip Tokenization since no syntax highlighting should be done.


octref:
For a 500k file on my machine, Atom used to open it in 5 sec, but then UI freezes for 20 sec before I can even scroll with no lag.
With this commit Atom opens the file in less than 2 sec, and no lag for scrolling/moving around.

steelbrain:
+1

kevinsawicki:
Awesome, thanks so much for this :ship:

PR #6010, opened 2015-03-17
Soft wrap hanging indentation spaces

### LINKED ISSUE #6003: Add option for hanging indent for soft-wrapped lines.
Please consider adding an option for hanging indents for soft-wrapped lines. It's really helpful for identifying wrapped lines, especially for languages like LaTeX where lines tend to be long and frequent editing makes manual hard wrapping needlessly tedious.

Example (adapted from @AlexLog's nicely-done example from Discuss): 
~~~latex

% Current behavior
\begin{itemize}
    \item Aaaa aaa aaa aaaaaaa aaaaaa aaaaaa aaaaaa
    aaa aaaa aaaaaa aaa.
    \item Aaaa aaa aaa aaaaaaa aaaaaa aaaaaa aaaaaa
    aaa aaaa aaaaaa aaa.
\end{itemize}

% Proposed behavior with 4-space hanging indent specified
\begin{itemize}
    \item Aaaa aaa aaa aaaaaaa aaaaaa aaaaaa aaaaaa
        aaa aaaa aaaaaa aaa.
    \item Aaaa aaa aaa aaaaaaa aaaaaa aaaaaa aaaaaa
        aaa aaaa aaaaaa aaa.
\end{itemize}

~~~

I'd add a variable to the soft wrap indent formula whose value is added to the indent level of the first line. By default, the value could be zero. In the UI, it should be on the `Settings` tab below `Tab Length` and named `Soft Wrap Hanging Indent`. If it uses a dropdown list like `Tab Length`, the options could be the same as the tab length options with the default zero option added.

While it could be designed as adding a configurable number of tab stops, it may be simpler to just make it a number of spaces, as there may be people who want a hanging indent level to be distinct from tab stops. Since the default would be zero, any level would have to be configured, so the number of spaces option would be no more configuration effort and more flexible.

Related discussions:

Discuss thread: https://discuss.atom.io/t/indented-soft-wrap/4808/1
Indent soft wrap (PR, Merged): https://github.com/atom/atom/pull/5567

as-cii:
Closes https://github.com/atom/atom/issues/6003

/cc: @nathansobo

### DIFF
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -140,6 +140,7 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[2].isSoftWrapIndentation).toBeTruthy()
+          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[3].isSoftWrapIndentation).toBeTruthy()
 
     describe "when the buffer changes", ->
       describe "when buffer lines are updated", ->
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -128,9 +128,9 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
 
-      describe "when editor.softWrapHangingIndentationSpaces is set", ->
+      describe "when editor.softWrapHangingIndent is set", ->
         beforeEach ->
-          atom.config.set('editor.softWrapHangingIndentationSpaces', 3)
+          atom.config.set('editor.softWrapHangingIndent', 3)
 
         it "further indents wrapped lines", ->
           expect(displayBuffer.tokenizedLineForScreenRow(11).text).toBe "       sort(left).concat(pivot).concat(sort(right)"
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -128,6 +128,12 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
 
+        it "correctly tokenizes hanging indentation spaces", ->
+          atom.config.set("editor.softWrapHangingIndentationSpaces", 3)
+          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
+          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
+          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[2].isSoftWrapIndentation).toBeTruthy()
+
     describe "when the buffer changes", ->
       describe "when buffer lines are updated", ->
         describe "when whitespace is added after the max line length", ->
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -133,6 +133,7 @@ describe "DisplayBuffer", ->
           atom.config.set('editor.softWrapHangingIndent', 3)
 
         it "further indents wrapped lines", ->
+          expect(displayBuffer.tokenizedLineForScreenRow(10).text).toBe "    return "
           expect(displayBuffer.tokenizedLineForScreenRow(11).text).toBe "       sort(left).concat(pivot).concat(sort(right)"
           expect(displayBuffer.tokenizedLineForScreenRow(12).text).toBe "       );"
 
--- a/spec/display-buffer-spec.coffee
+++ b/spec/display-buffer-spec.coffee
@@ -124,9 +124,16 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(3).tokens[1].isHardTab).toBeTruthy()
 
       describe "when a line is wrapped", ->
-        it "correctly tokenizes soft wrap indentation tokens", ->
-          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
-          expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
+        it "breaks soft-wrap indentation into a token for each indentation level to support indent guides", ->
+          tokenizedLine = displayBuffer.tokenizedLineForScreenRow(4)
+
+          expect(tokenizedLine.tokens[0].value).toBe("  ")
+          expect(tokenizedLine.tokens[0].isSoftWrapIndentation).toBeTruthy()
+
+          expect(tokenizedLine.tokens[1].value).toBe("  ")
+          expect(tokenizedLine.tokens[1].isSoftWrapIndentation).toBeTruthy()
+
+          expect(tokenizedLine.tokens[2].isSoftWrapIndentation).toBeFalsy()
 
       describe "when editor.softWrapHangingIndent is set", ->
         beforeEach ->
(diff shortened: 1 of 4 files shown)


spec/display-buffer-spec.coffee
```diff
@@ -128,6 +128,12 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
 
+        it "correctly tokenizes hanging indentation spaces", ->
```
as-cii: Please note that I wrote only a test that verifies the correct tokenization of `hangingIndentationSpaces`, as positioning was already thoroughly tested by other specs. I believe that adding similar ones for these hanging spaces was redundant and not worth the maintenance.

thomasjo:
Bloody brilliant! :sparkling_heart:

src/tokenized-line.coffee
```diff
@@ -118,20 +118,27 @@ class TokenizedLine
     oddIndentLevel = @indentLevel - Math.floor(@indentLevel)
     Math.round(@tabLength * oddIndentLevel)
 
-  buildSoftWrapIndentationTokens: (token) ->
+  buildSoftWrapIndentationTokens: (token, hangingIndentationSpaces) ->
```
nathansobo: What's your reasoning behind breaking the "hanging" indentation into a separate token rather than just making the initial indentation token longer?

nathansobo:
Excited about the feature, but I'm wondering if anyone has any ideas for the name of the setting. `softWrapHangingIndentationSpaces` feels a bit wordy. Maybe `softWrapIndentationOverhang`? Not loving that either. We could go with it I'd just like to kick around some alternatives.

src/tokenized-line.coffee
as-cii: @nathansobo: if I understand your question correctly, we need to break the "hanging spaces" into tokens so that we could show indent guides.

After a bit of sleep (I love Hammock Driven Development :sleeping:), however, I realized this wasn't doing what I expected: indeed indent guides weren't shown, as only one token was created for all the hanging indentation spaces.

f0b9bb7ce3e75f716c0d0297bbcacee5fe289f81 solves this by calculating the `hangingIndentLevel` and summing it to `TokenizedLine#indentLevel`.

as-cii:
> Excited about the feature, but I'm wondering if anyone has any ideas for the name of the setting. softWrapHangingIndentationSpaces feels a bit wordy. Maybe softWrapIndentationOverhang? Not loving that either. We could go with it I'd just like to kick around some alternatives.

Yeah, naming things is hard :disappointed: To be honest I simply borrowed the term from the referenced pull-request but it feels wordy to me as well. Maybe we could involve @atom/non-github-maintainers to help us with this? Your help would be much appreciated :+1: 

As usual, thank you @nathansobo for the interesting discussion and for your always great feedback! :pray:

grizzilus:
I just stumbled across this PR and the naming conversation.

From the typography perspective I like either the current `softWrapHangingIndentationSpaces` or a shorter `softWrapHangingIndentation`. In typography, the general paragraph shape that results from line-wrapping like that which results from having the option in this PR enabled can come about one of two ways: either the paragraph has an `outdentation` or a `hanging indentation`. (I personally have not seen any other names, and a quick flip through all the typography books I have at hand seems to confirm this.)
 
An `outdentation` is when the paragraph extends into the left margin by some amount. The analogue in a text editor would by a negative-valued indentation width, which doesn't really make sense. But this brings up another interesting concern I'll mention at the end of this comment.
 
A `hanging indentation` is is an indent of all lines after some _n_th line of the paragraph by some amount, which, in my opinion, corresponds exactly to the behavior of this PR. So having `HangingIndentation` somewhere in the naming makes it very clear what is going on. In particular, the original issues was created used `TeX` as an example, where this type of output is controlled by the lengths `\hangindent` and `\hangafter`, which may be why the author of that issue used the term `hanging` in the name of the issue.

The `outdentation` discussion bring up an interesting point: this PR results in some odd behavior if `softWrapHangingIndentationSpaces` is negative and its magnitude is larger than the main indent level. I would expect the soft-wrapped lines to be aligned with the left side of the editor view, but they indent themselves by the difference between the value of `softWrapHangingIndentationSpaces` and the number of spaces by which the first line is indented. (It will be clear if you try some negative values.)

50Wliu:
`indentSoftWrap`?  That sounds like it could be a bit misleading though so perhaps a comment could be added stating "this is relative to the original indent" or something?

as-cii:
@50Wliu @grizzilus: thank you for answers! 

Based on @grizzilus thorough explanation (thank you for that), I renamed the config variable for now to `softWrapHangingIndent` but, most importantly, we discovered an interesting edge case, which I didn't take into account. As I intended the feature, the value of this config setting **must always be positive** and, therefore, I fixed that accordingly (https://github.com/atom/atom/commit/e5fa44171df82962efb185d6b5d36ab2b8329f86).

Let's wait for @nathansobo final take on this :fireworks:

grizzilus:
I'm actually not getting any handing indent if I build the last commit here but I can't find any error (or anything missed) in 7c33b9b...

as-cii:
@grizzilus: could you post here your `config.cson`, please? Also, make sure you have updated the setting to:

```
softWrapHangingIndent: <number>
```

grizzilus:
@as-cii Nevermind, it is fine, sorry about that. I rebuilt but when changing the value in the GUI it wasn't updating. Removing the old setting from the config file allowed me to set directly or from the GUI.

On a side note, I wonder if it is reasonable to allow for negative indents. I know you originally intended for positive values only, but what if we did allow negative indents and just cap it at the maximum of 0 and the sum of the regular indent and the hanging indent? While I can't think of an really useful example (other than easily seeing where there are long lines nested very deep more easily), it seems like a straightforward change to allow more flexibility.

as-cii:
> While I can't think of an really useful example (other than easily seeing where there are long lines nested very deep more easily), it seems like a straightforward change to allow more flexibility.

Thank you for pointing this out @grizzilus. I must say I have considered this option while implementing the change, but as you're pointing out I couldn't find a meaningful example and, therefore, I opted to simply make the configuration stricter allowing only positive values. Supposedly, this makes its intent a bit clearer (or does it?) sacrificing a bit of flexibility (e.g. negative values).

On the other hand, if we find a good use-case for negative values I'll be happy to implement that feature as well. :dancers: 

/cc: @AlexLog @jtkiley

grizzilus:
The more I think about it, you are right. I can't imagine the magnitude of this being more than twice the indentation amount, and if text starts that close to the margin, there are bigger issues to deal with than the hanging indent. But great work @as-cii on this and the original indented softwrap; it is such a help to people who use TeX-based files regularly.

paulcbetts:
How about `wrapHangingIndentation` - Atom doesn't do hard-wrap, and "IndentationSpaces" seems redundant

jtkiley:
@as-cii @grizzilus I can't think of a good use of negative values, either. I thought about it briefly when I was writing up the issue, and I thought about it after seeing the discussion this morning. Someone may come up with a cool use case, but a negative indent would actually look like the old behavior (i.e. wrapped lines beginning at the left margin) in some cases. If anyone is clamoring for the old behavior as an option, I haven't seen it.

Echoing @grizzilus, thank you @as-cii for the awesome work on indented soft wrap and the hanging indent. It's a huge help for TeX work.

src/config-schema.coffee
```diff
@@ -149,6 +149,10 @@ module.exports =
       softWrapAtPreferredLineLength:
         type: 'boolean'
         default: false
+      softWrapHangingIndent:
```
nathansobo: :+1: I like the new shorter name.

src/tokenized-line.coffee
```diff
@@ -111,27 +111,32 @@ class TokenizedLine
...
+    if oddIndentationSpaces = @oddIndentationSpacesForIndentLevel(indentLevel)
       indentTokens.concat(
-        token.buildSoftWrapIndentationToken @getOddIndentationSpaces()
+        token.buildSoftWrapIndentationToken(oddIndentationSpaces)
       )
     else
       indentTokens
 
-  softWrapAt: (column) ->
+  softWrapAt: (column, options = {}) ->
```
nathansobo: What do you think about just passing the overhang in as a direct parameter rather than using an options object. It's not as pretty but it saves an object allocation and since this code is internal, seems like a reasonable trade-off.

spec/display-buffer-spec.coffee
```diff
@@ -128,6 +128,20 @@ describe "DisplayBuffer", ->
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[0].isSoftWrapIndentation).toBeTruthy()
           expect(displayBuffer.tokenizedLineForScreenRow(4).tokens[1].isSoftWrapIndentation).toBeTruthy()
 
+      describe "when editor.softWrapHangingIndent is set", ->
+        beforeEach ->
+          atom.config.set('editor.softWrapHangingIndent', 3)
+
+        it "further indents wrapped lines", ->
+          expect(displayBuffer.tokenizedLineForScreenRow(11).text).toBe "       sort(left).concat(pivot).concat(sort(right)"
```
nathansobo: Could you also show the text for row 10 here? I think that would tell the :book: a bit more.

spec/display-buffer-spec.coffee
```diff
@@ -128,6 +128,20 @@ describe "DisplayBuffer", ->
...
+          expect(displayBuffer.tokenizedLineForScreenRow(12).text).toBe "       );"
+
+        it "correctly tokenizes hanging indentation spaces", ->
```
nathansobo: Some suggestions for clarifying this example:

* Talk about the content of each token.
* Assert that the token at index 4 is *not* soft-wrap indentation to clarify the boundary.
* Maybe mention why we're breaking the indentation up in the description: maybe "breaks soft-wrap indentation into a token for each indentation level to support indent guides".

src/tokenized-line.coffee
```diff
@@ -111,27 +111,32 @@ class TokenizedLine
...
-  # Calculates how many trailing spaces in this line's indentation cannot fit in a single tab.
+  # For a given `indentLevel`, calculates how many trailing spaces cannot fit in a single tab.
   #
-  # Returns a {Number} representing the odd indentation spaces in this line.
-  getOddIndentationSpaces: ->
-    oddIndentLevel = @indentLevel - Math.floor(@indentLevel)
+  # indentLevel - {Number}
+  # Returns a {Number} representing the odd indentation spaces in `indentLevel`.
+  oddIndentationSpacesForIndentLevel: (indentLevel) ->
+    oddIndentLevel = indentLevel - Math.floor(indentLevel)
```
nathansobo: Thought here: couldn't you also do it this way?

```coffee
(indentLevel * @tabLength) % @tabLength
```

src/tokenized-line.coffee
```diff
@@ -111,27 +111,32 @@ class TokenizedLine
...
     Math.round(@tabLength * oddIndentLevel)
 
-  buildSoftWrapIndentationTokens: (token) ->
-    indentTokens = [0...Math.floor(@indentLevel)].map =>
+  buildSoftWrapIndentationTokens: (token, hangingIndent) ->
```
nathansobo: If you change the implementation of this method, you don't need `oddIndentationSpacesForIndentLevel` at all. 

```coffee
  buildSoftWrapIndentationTokens: (token, hangingIndent) ->
    totalIndentSpaces = (@indentLevel * @tabLength) + hangingIndent
    indentTokens = []
    while totalIndentSpaces > 0
      tokenLength = Math.min(@tabLength, totalIndentSpaces)
      indentToken = token.buildSoftWrapIndentationToken(tokenLength)
      indentTokens.push(indentToken)
      totalIndentSpaces -= tokenLength
    indentTokens
```

I tested this version locally, but wanted to avoid committing to this branch without your :+1:.

src/tokenized-line.coffee
nathansobo: Actually, I have a suggestion for removing this method entirely by changing the implementation of `buildSoftWrapIndentationTokens` below...

as-cii:
@nathansobo: thank you for all the great suggestions: hopefully, I have implemented them all :+1: 

Please, have one final look and feel free to commit on this branch if you need to :smile:

nathansobo:
:joy: :joy_cat: Members of every species thank you for your contribution.

PR #6016, opened 2015-03-18
:memo: document where to add the tests and how to run them

basarat:
This is purely from code review. I haven't managed to get tests working on windows.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ For more information on how to work with Atom's official packages, see
   [JavaScript](https://github.com/styleguide/javascript),
   and [CSS](https://github.com/styleguide/css) styleguides.
 * Include thoughtfully-worded, well-structured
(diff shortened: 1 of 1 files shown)


as-cii:
@basarat: thank you for your contribution! When I started contributing on Atom, running tests was one of the major pain points I encountered, therefore I think this PR will be much appreciated by newcomers as well :+1: 

Running specs from cmd with `script/grunt run-specs` doesn't work particularly well for me: as a result, I usually run tests via `apm test`, which by the way works greatly to run single-package specs as well. By the way, a similar discussion took place [here](https://discuss.atom.io/t/testing-the-core-on-ubuntu-14-10/) and you may want to have a look at it :wink: 

What do you think if we also specify that tests can be run also through the GUI (e.g. `Window: Run Package Specs`)?

Thank you again for this :space_invader:

basarat:
`Window: Run package specs` does not work for me (on windows): 
image

`apm test` works like a charm. Also I used `fdescribe` to run just my tests, but not going to document it as people might commit with that :rose:. I've updated this pull request to take `apm test` into account.

as-cii:
Thank you @basarat!

PR #6026, opened 2015-03-18
Check Comsumed/Provided Service Method Exists, fixes #416

jlord:
This is a short term fix for https://github.com/atom/settings-view/issues/416 (because it's causing quite a bit of errors)—making Atom to check first that the method exists and skip if it doesn't.  

A more ideal, but more complicated, fix down the line will involve updating requires when upgrading a package. 

:pear: 'd with @kevinsawicki 
cc @izuzak @maxbrunsfeld


kevinsawicki:
I believe the CI failures here are fixed by https://github.com/atom/atom/commit/414b82f05ec60562563b9b94fa9cea42ac9a53d9 so you might want to rebase or merge master into this branch.

izuzak:
:zap: @jlord -- happy to see this, it will stop all those exceptions. :heart: 

You mentioned that this is a short term fix -- should we open a new issue so that we can track the underlying problem and don't forget to fix it in the future?

Issue #4924, opened 2015-01-08
Rendering artifacts since Chrome 39 upgrade


kevinsawicki:
Most noticeable in the tree view, resizing it, scrolling, etc. makes it redraw and go away.

![screen shot 2015-01-07 at 2 47 31 pm](https://cloud.githubusercontent.com/assets/671378/5666877/fe272bc4-9717-11e4-9d81-c6f737e03def.png)

/cc @zcbenz Any ideas on this one?

zcbenz:
I saw this too sometimes, I believe it is a rendering bug of Chrome.

paulcbetts:
We're getting rendering issues too, though ours are worse; here's a video:

http://cl.ly/3q3x2t3F3f30

On both OS X and Windows, these boxes blank out (they should normally have content in them). The weird thing is, opening the Dev Tools seems to sometimes make the bug go away

kevinsawicki:
@zcbenz So should we try reporting it to Chromium or maybe search their issue tracker?

zcbenz:
We have to make it reproducible on Chrome browser, otherwise it would be unlikely for them to look into this.

I think we can check if this bug exists in Chrome 40 first, it is still beta but the next stable release should be quite close.

simurai:
I saw it too on OS X, when I enabled "General > Show scroll bars: always" on a retina screen. But can't reproduce it when trying the same again.

Ohh.. It happened when using a theme that has a transparent scrollbar track. So might be related to #4578. The Atom Dark and One Dark theme now have an opaque scrollbar background, which might fix it.

wyand:
@simurai I am seeing the tree view artifacts while using Atom Dark for both UI and syntax, on OSX.

simurai:
@wyand Did you see it happen again just recently? I haven't seen it since making the scrollbars have an opaque color #4578.

wyand:
@simurai I messed up! I I have not made the scrollbars opaque. Misunderstood the previous comment to imply that Atom Dark had opaque scrollbars by default. Sorry about that.

simurai:
@wyand NP. Yeah, it used to be semi transparent, but got changed to opaque. Not sure when exactly, maybe since 1-2 weeks ago. So if you still see it happen from now on, then this issue might not be related to the scrollbar transparency.

iclanzan:
It is still happening to me, on Atom 0.177  on OS X using One Dark UI.

aldendaniels:
[...]

PR #6027, opened 2015-03-19
Upgrade to Chrome 41

### LINKED ISSUE #309: Native Emoji Support
It would be awesome to support native emoji.

Right now I just see nothing when I try to paste one in:

image

Compare to Safari:

image

Though unfortunately Chrome is broken too, so maybe this isn't something we can fix:

image

zcbenz:
This PR upgrades atom-shell to v0.22.1, which uses Chrome 41 and io.js 1.5.1.

Fixes #309.

screen shot 2015-03-19 at 9 22 46 am


zcbenz:
Chrome 41 nows [forces using sub-pixel font scaling](https://code.google.com/p/chromium/issues/detail?id=362112), bringing back the [incorrect character size issue](https://github.com/atom/atom/pull/3495). /cc @atom/core

nathansobo:
Taking a look at the character measurement issue now. @zcbenz would it be feasible to investigate the measurement problems at the native level? In the meantime I'll try to find a workaround.

zcbenz:
@nathansobo I can patch libchromiumcontent to turn off sub-pixel font scaling if you want. The measurement problem is a bug of blink, which is hard to investigate in a short time.

kevinsawicki:
@zcbenz just to confirm, we won't need to bump any of our native libraries to go from chrome 40 to 41?

kevinsawicki:
>  I can patch libchromiumcontent to turn off sub-pixel font scaling if you want.

I think this is the right move so we can get chrome 41 out without having the font measurement stuff block it. 

@zcbenz has you noticed if the tree view rendering problems went away in chrome 41? #4924

zcbenz:
@kevinsawicki I haven't tried this branch enough because the measurement bug made Atom hard to use, I'll retry after patching libchromiumcontent.

zcbenz:
> @zcbenz just to confirm, we won't need to bump any of our native libraries to go from chrome 40 to 41?

No. The version of Nan used by our native libraries are compatible with Chrome 41.

zcbenz:
After playing with this branch for some time, I think everything is working fine. The tree-view rendering issue is still there, exactly the same with Chrome 40.

One possible problem is `String.contains` has been removed completely in Chrome 41, though it has never been a standard method, some packages are still using it (`japanese-wrap` is one of them, but it could be the only one). We can polyfill it, but I don't think we should since it has never been a standard.

thomasjo:
> After playing with this branch for some time, I think everything is working fine.

I haven't noticed any regressions or problems either. This looks good to me! Merge after v0.188.0 ships?

kevinsawicki:
> Merge after v0.188.0 ships?

Yeah, I'd like to get 0.188 out first, and then have this go in in 0.189 (or later if needed).

PR #6038, opened 2015-03-19
Remove unwanted loop returns

kevinsawicki:
Profiling some of the lower level libraries in the past has shown GC pauses and slow downs when methods accidentally build-up and return an array accidentally when the function ends with a `for` or `while` loop.

This cleans up a bunch of areas in core that were doing this involuntarily as well as adds a grunt task to lint for them in the future.

This also updates array copies to consistently use `slice` instead of `for` which is significantly faster, https://jsperf.com/slice-vs-loop-return/2

I don't have any metrics on before/after speedups but it does reduce the total JS size of the `src` folder by ~225 lines.

This also prevents some public APIs from returning arrays accidentally when they should have been returning `undefined`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,6 +104,9 @@ For more information on how to work with Atom's official packages, see
   should be lower-case:
   * `getURI` instead of `getUri`
   * `uriToOpen` instead of `URIToOpen`
+* Use `slice()` to copy an array
+* Add an explicit `return` when your function ends with a `for`/`while` loop and
+  you don't want it to return a collected array.
 
 ## Documentation Styleguide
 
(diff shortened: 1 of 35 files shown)


as-cii:
@kevinsawicki: well done! :+1: :+1: :racehorse: :horse_racing: 

I hope Coffeescript purists won't feel offended by this but, what about enforcing an *explicit return* policy for every method longer than one line? This would result in:

1. Forcing the implementer to think about what is being returned and why (thereby avoiding the kind of problems this PR is highlighting);
2. Easier code reviews, as methods will be more "uniform" when there are early return statements;

On the other hand this would add some clutter to the code and, to be honest, I don't know if this is the right way to go either. I just wanted to share this quick thought with you :thought_balloon: 

Once again, nice job! :wink:

kevinsawicki:
@as-cii yeah, it would be really nice to enforce this for public APIs at least so they don't end up leaking things accidentally that people end up relying on.

PR #6041, opened 2015-03-19
Revert "Revert "Don't soft-wrap on indentation""

nathansobo:
Reverting the revert... the spec failures weren't caused by the PR I reverted as I originally thought. My apologies @as-cii.


as-cii:
No need to apologize, @nathansobo :relaxed:

Thank you for reporting the issue: it should have been fixed in #6042 :wink:

PR #6042, opened 2015-03-19
Fix random mutation spurious errors

as-cii:
...which were caused by the recent soft-wrap hanging indent change. The random
mutation editor was not calling `softWrapAt` appropriately. In facts, it was not
passing `hangingIndent` as a parameter, which means it was treated as `null`:
when summed to `totalIndentationSpaces` the result was always `NaN`, thus
exhibiting strange behaviors.

@nathansobo: sorry about this issue :disappointed: I have managed to reproduce it on master as well, though, and I'd say we can safely bring back the reverted branch into master :+1:  

This brings up another important question, which I am not sure I know the answer of: how could we can reduce the coupling between the random text editor and our real production code?

### DIFF
--- a/spec/random-editor-spec.coffee
+++ b/spec/random-editor-spec.coffee
@@ -3,7 +3,7 @@ randomWords = require 'random-words'
 TextBuffer = require 'text-buffer'
 TextEditor = require '../src/text-editor'
 
-describe "TextEditor", ->
+fdescribe "TextEditor", ->
   [editor, tokenizedBuffer, buffer, steps, previousSteps] = []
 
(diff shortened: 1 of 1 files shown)


as-cii:
@nathansobo: I have been burning my CPU to reproduce other issues through the random editor :joy: They seem to be gone now, but I think it's better to run CI tests two or three times before merging this :wink:

as-cii:
Closing this, as it was superseded by 5f76979fc83367d916e8962a6dd31c12983eff57.

nathansobo:
Oh wow, I totally missed this.

PR #6047, opened 2015-03-20
Fix deprecated calls and recommendations in deprecated methods

postcasio:
Noticed that a couple of deprecated methods on project were themselves calling deprecated methods, and recommending the use of deprecated methods.


kevinsawicki:
Thanks for fixing these up :+1:

PR #6062, opened 2015-03-22
Defer keydown event to refresh screen while fast key typing

ypresto:
I feel atom is very laggy while typing fast or key repeat (than vim :) ).
Especially lag on repeating j/k caret move is huge, freezes for seconds.

With time profiler, I found continuous keydown events block main thread for >10sec (!).

This happens even if vim-mode is disabled.

It is significant because I use [Karabiner](https://pqrs.org/osx/karabiner/index.html.en) to make key repeat even faster than maximum speed of OS X preferences.

This PR increases rendering rate by deferring keydown event handling until next setTimeout dispatch.
This enables to refresh screen before handling next key event.
Lag on fast typing in insert mode is reduced too.

Left: before, right: after!

atom-key-typing-speed-720p

(I'm wondering that there is so much idle time after releasing key and before main thread is released. See below.)

### DIFF
--- a/src/window-event-handler.coffee
+++ b/src/window-event-handler.coffee
@@ -130,7 +130,7 @@ class WindowEventHandler
     bindCommandToAction('core:select-all', 'selectAll:')
 
   onKeydown: (event) ->
-    atom.keymaps.handleKeyboardEvent(event)
+    _.defer -> atom.keymaps.handleKeyboardEvent(event)
     event.stopImmediatePropagation()
 
   onDrop: (event) ->


ypresto:
I found `Shift+Command+P` (Command Palette) and other accelerator keys not working (seems to be opened but close immediately). I'll investigate about it.

ypresto:
I found that source of problem is some kind of installed plugin.
I also read implementation of atom-keymap's KeymapManager.
This changeset should not be applied because it breaks event consuming logic like event.preventDefault() in manager.

Sorry for bad PR m(_ _)m

PR #6075, opened 2015-03-23
Display unsaved buffer while prompting to save.

### LINKED ISSUE #6070: Show the unsaved file while displaying the 'unsaved changes' alert at closing.
If you have several files with unsaved changes, when you try to exit the application you will only see the current open buffer when clicking discard. Atom should show you the particular buffer about to be saved or discarded.

I have atom 0.144.0 on an ubuntu 14.04

Similar problem here: #5823

mathiasvr:
This should fix #6070

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -457,13 +457,20 @@ class Pane extends Model
     else
       return true
 
+    currentActiveItem = @activeItem
+    @activateItem item
     chosen = atom.confirm
       message: "'#{item.getTitle?() ? uri}' has changes, do you want to save them?"
       detailedMessage: "Your changes will be lost if you close this item without saving."
       buttons: ["Save", "Cancel", "Don't Save"]
+    
+    @activateItem currentActiveItem if chosen isnt 0
 
     switch chosen
-      when 0 then @saveItem(item, -> true)
+      when 0 
+        saved = @saveItem(item, -> true)
+        @activateItem currentActiveItem
+        saved
       when 1 then false
       when 2 then true
 


botverse:
Looks good to me
:+1:

botverse:
Oh wait.

Is it going to show the current item after hitting `Don't save` too?
Why do you restore the original file after `Save`?

abe33:
Having specs for this feature along with the patch would be great, so that future change don't break it.

mathiasvr:
@botverse The file being closed will be shown while showing the dialog, if you choose to save, the file remains visible until the save dialog closes as well. After any dialog the file which was active before pressing close will be re-activated. This is of course redundant when the file being closed is already the active one, or when closing all files at once. Maybe there is a better way?

src/pane.coffee
```diff
@@ -457,13 +457,20 @@ class Pane extends Model
     else
       return true
 
+    currentActiveItem = @activeItem
+    @activateItem item
```
mathiasvr: This is the only line which should be merged to sublime text like functionality.

postcasio:
If you have multiple unsaved items, will this keep reactivating the originally active item between activating each unsaved item? Ideally it should activate each item, then finally return to the original item once all unsaved items have been prompted.

Alternatively, just don't activate the original item and leave the unsaved item active. ST2 has this behaviour for what it's worth.

botverse:
I've been playing around with the both options and I actually prefer what @mathiasvr has developed, this is for you to compare:

Only with this: `@activateItem item`

dumber

With the returning to tab function:

clever

So for me this is a :+1: as it is.

mathiasvr:
@botverse Very good idea with the comparison :smiley: 

@postcasio Yes it will keep reactivating the original item when closing multiple items, however this is not visible to the user. It would be nice to avoid this, but then this feature would have to be spread out to multiple functions. Alternatively we could pass reactivation as an option?

alexandernst:
Any news on this one?

mnquintana:
@mathiasvr Can you add some specs for this?

alexandernst:
Original dev @mathiasvr seems missing which is sad as this feature looks really awesome :(

Zireael07:
Yeah, this would be awesome!

mnquintana:
I might try to take this up in the next week if I have some free time :smile:

thedaniel:
I added the `help-wanted` label in case anyone from the community wants to pick this back up.

stereobooster:
Tried to pick this up https://github.com/atom/atom/compare/master...stereobooster:6075-focus-tab-on-save-prompt?expand=1

But for some reason Atom doesn't want to switch active tab before it shows save prompt. It treats currently destroyed buffer as active, while it is not displayed :/

Siilwyn:
I gave it a shot but this is actually pretty hard to do now. All dialog (related) code is written synchronous and thus while the dialog is open it's not possible to switch from tab. So I tried to re-write some functions to accept callbacks but it goes a lot deeper than I initially thought...

https://github.com/atom/atom/compare/master...Siilwyn:display-unsaved-buffer

damieng:
PR is very out of sync and hasn't been merged in 2 years.

Issue #3821, opened 2014-10-14
Ligatures aren't spaced correctly


YawarRaza7349:
I'm using the Atom for Windows Alpha from the Atom website (not Chocolatey), version 0.136.0 (as viewed in the Help menu and with `--version` in the command line, the `version` file in the Atom folder says v0.17.1; it was downloaded around 12:30AM US Eastern on October 14, 2014 if that helps) on Windows 8.1. I set `text-rendering: optimizeLegibility;` for `.editor` in `styles.less` (because otherwise the ligatures aren't in effect and I get each individual character). This happens regardless of whether hardware acceleration is enabled. It also happens in `--safe` mode, and the developer tools console shows no errors.

![temp](https://cloud.githubusercontent.com/assets/4254261/4624455/bad18f5a-5364-11e4-9a0f-8c59596bcd59.png)
```
(<*>)::Functor f=>(a->b)->(f a->f b)
fmap::Functor f=>(a->b)->(f a->f b)
```
in [Hasklig](https://github.com/i-tu/Hasklig), which has ligatures for `<*>`, `::`, `=>`, and `->`.

![temp](https://cloud.githubusercontent.com/assets/4254261/4624502/6b794cee-5365-11e4-8d85-a129dee8ae1d.png)
```
fifififififififififi
```
in Calibri, using the ligature for `fi`. The cursor is at the end of the line, and the misspell underline only leads up to that point as well. The ligature for `ff` in Calibri has the opposite effect; the cursor and underline extend farther than text on the screen if you put enough f's on a line.

I can provide more details if you need them.

ai:
```css
body {
    -webkit-font-feature-settings: "liga" 1;
    font-feature-settings: "liga" 1;
    text-rendering: optimizeLegibility;
}
```

makes little bit better, but not enough

ai:
Also same problems are with [Fire Code](https://github.com/tonsky/FiraCode)

benogle:
Try disabling `Use hardware acceleration` in the settings. Does that help?

ai:
@benogle thanks for advice. But it really not help. LIgatures behaviour is totally same without “Use hardware acceleration”.

50Wliu:
Refs #2880 (possible duplicate).

ai:
@50Wliu nope. #2880 is about disabled rendering ligatures (two symbols instead of one) and this issue is about wrong caret position on font with ligatures.

50Wliu:
Oh, whoops.

[...]

PR #6083, opened 2015-03-24
Char measurement

### LINKED ISSUE #6055: We need a better approach to text measurement
When Chrome 39 added support for subpixel font scaling, it broke our approach to horizontal position measurement. We disabled subpixel scaling with a flag, but that was only a temporary stopgap measure. The truth is our existing approach to measuring horizontal positions on screen is somewhat ill-conceived, and it needs to be changed.

## The existing approach

In our existing approach, we attempt to measure the width of characters in a given context. So, for example, we can determine that the width of an `a` character in the `.source.coffee .meta.function.coffee` is 11px at the current font size. Then, when we calculate the pixel position for a given row and column, we just simply total up the recorded width for every character we encounter.

The problem is that for non fixed-width fonts, a given character doesn't have a single width. Kerning, ligatures, and other stuff the type rendering engine does can result in an `a` taking up different amount of space in different contexts. I am not an expert in this area yet. What's worse, when subpixel font scaling is turned on, Chrome starts reporting different widths *even for fixed width fonts*. We're going to need a different approach.

## Idea for a better approach

My current thinking is that when calculating a pixel position, we need to explicitly measure the distance from the beginning of the line to the column in question rather than consulting our encyclopedia of character widths. This was our original approach but I moved away from it because we were naively reading the DOM whenever we needed to translate positions, which was causing reflows.

But we actually *can* safely read this information, so long as we do so either at the start of an update *before* we actually update the DOM, or at the end of an update, after we are finished updating the DOM. When simply moving the cursor, we can measure before the update. When typing, however, we need to measure *after* the update in order for the content to be on screen, then trigger another update based on our measurement.

We may also be able to make these measurements off-DOM, like in an iframe or something. My main concern here is making sure that we capture all the styling that's applied to the text. We may be able to draw a line at the shadow boundary, and just say the only styling that we will apply for purposes of character measurement is the syntax themes. Theme stylesheets could easily be loaded in an iframe via the `atom-styles` element. Or we could set up the entire world in this iframe and measure there.

Or maybe there's some other, better approach to character measurement out there that Atom should apply instead. The main requirements are as follows:

* Measurement should not force a blocking reflow.
* Measurement should capture user styling as seamlessly as possible.
* Measurement should handle different widths being reported for the same character in different contexts, as is currently the case with subpixel font scaling enabled.

Can anyone from the community help out with this? Ideas are welcome if they're specific. Code would be even more helpful. @as-cii I know I'm pinging you on a ton of stuff lately but this seems right in your wheelhouse based on your recent presenter work.

/cc @atom/non-github-maintainers

as-cii:
Fixes #6055 

This is still a work in progress: code is messy and specs are not passing yet. ~~With this PR we basically wrap every character with a span so that we're able to measure it. This helps to calculate the real character's width on screen: whenever this is not possible, however, the current solution falls back to `@defaultCharWidth`.~~

We have split the `TextEditorPresenter` and the `TextEditorComponent` updates in two phases: pre-measure and post-measure. The former should calculate anything related to lines, whereas the latter  performs updates which need to be based on lines measurement.
- [x] Write new specs to consolidate this behavior
- [x] Make existing specs pass
- [x] Remove references to scoped char width
- [x] Scrolling into an unmeasured line with decorations (or cursors) has issues (since we don't measure such lines on scroll anymore and updating them doesn't require any user action)


src/display-buffer.coffee
```diff
@@ -332,6 +322,14 @@ class DisplayBuffer extends Model
     @scopedCharacterWidthsChangeCount++
     @characterWidthsChanged() unless @batchingCharacterMeasurement
 
+  setCharWidthsForLineId: (lineId, charWidths) ->
```
nathansobo: I very much like this approach of just keeping a parallel array.

src/display-buffer.coffee
```diff
@@ -332,6 +322,14 @@ class DisplayBuffer extends Model
...
+    @charWidthsForLineId[lineId] = charWidths
+    @characterWidthsChanged() unless @batchingCharacterMeasurement
+
+  getCharWidthForLine: (line, index) ->
```
nathansobo: I wonder if this should just be `getCharacterWidth` and take a `Point`. It could then consult its lines internally.

src/lines-component.coffee
```diff
@@ -270,8 +270,19 @@ class LinesComponent
...
-    @measureCharactersInNewLines()
+    @measureCharactersInVisibleLines()
+
+  measureCharactersInVisibleLines: ->
+    for id, state of @oldState.lines
+      lineNode   = @lineNodesByLineId[id]
+      characters = lineNode.querySelectorAll("span.char")
+      charWidths = []
+
+      for character in characters
```
nathansobo: Take a look at the previous code for an approach to measurement that doesn't depend on spans. You can build an iterator over the text nodes then set the start and end index of a DOM selection for a given pair of offsets in a text node to get a bounding rectangle from that.

nathansobo:
I'm wondering if rather than storing the widths of each character at each position if it wouldn't be easier to just store a running total. So if every character were 10px wide on a line with 4 characters, it would be `[10, 20, 30, 40]`. Basically, rather than an array of widths it is now an array of the horizontal position of the right side of that character in absolute terms.

nathansobo:
:rocket: Looking forward to collaborating on this problem with you.

as-cii:
> I'm wondering if rather than storing the widths of each character at each position if it wouldn't be easier to just store a running total. So if every character were 10px wide on a line with 4 characters, it would be [10, 20, 30, 40]. Basically, rather than an array of widths it is now an array of the horizontal position of the right side of that character in absolute terms.

I was just thinking about this and it's definitely something we could do. I wonder, however, if we could make it even faster :open_mouth: What if, while measuring lines, we stored:

* A matrix of rows and columns, with the absolute left position of each column
* An interval skip list (or a similar data structure) of absolute left positions (this is needed for clipping when in the middle)

If we manage to do this, then `screenPositionForPixelPosition` and `pixelPositionForScreenPosition` would basically avoid tokens iteration altogether, making finally use of the existing computation performed by `measureCharactersInVisibleLines`. What do you think @nathansobo? This would be huge in terms of speed! :racehorse: 

By the way, specs are :green_heart: now (exception made for those referring to scoped char widths) and it would be great if you could have a second look at this. Thank you :bow:

In the meantime I'll try to put a bit of thought into a reflow-less solution, even though I feel like this PR performs quite well.

src/display-buffer.coffee
as-cii: :+1:

I'll do that once we agree on pushing this solution forward. What do you think?

as-cii:
@nathansobo: tomorrow I'll write a bunch of specs for this and collect some useful metrics in order to make a more informed decision :rocket:

nathansobo:
> An interval skip list (or a similar data structure) of absolute left positions (this is needed for clipping when in the middle)

This seems like it might be more than is needed. If you just store a running total in an array, it would be pretty easy to binary search it when converting from pixels to columns.

I'd love to see some profiling / timeline results for this change at some point. I know we need to fix this issue but I'm worried about regressing performance. Thanks so much!

src/display-buffer.coffee
```diff
@@ -37,7 +37,7 @@ class DisplayBuffer extends Model
 
   verticalScrollMargin: 2
   horizontalScrollMargin: 6
-  scopedCharacterWidthsChangeCount: 0
+  characterWidthsDidChange: false
```
nathansobo: Maybe `characterWidthsChanged`? We have been using `will` and `did` in the names of hooks.

src/lines-component.coffee
```diff
@@ -302,29 +322,24 @@ class LinesComponent
...
+
+        while nextTextNodeIndex <= charIndex
+          textNode = iterator.nextNode()
+          textNodeIndex = nextTextNodeIndex
+          nextTextNodeIndex = textNodeIndex + textNode.textContent.length
+
+        i = charIndex - textNodeIndex
+        rangeForMeasurement.setStart(textNode, i)
+        rangeForMeasurement.setEnd(textNode, i + charLength)
+        charWidth = rangeForMeasurement.getBoundingClientRect().width
```
nathansobo: Glad you're using this approach instead of spans. You could leave the start of the measurement range at index 0 of the first text node if you want to store a running total rather than individual widths.

as-cii:
I have performed some benchmarks and captured some screenshots so that we can consider them together.

All the benchmarks were performed on a 1920x650 px window.

## Move 300 cursors simultaneously

**master**
master

**as-char-measurement**
dev

**Conclusion**: we have significantly reduced the amount of time spent in updating cursors. This has nothing to do with our char measurement strategy, however, because it's mainly due to 8a153f16596ba2c334a6c5e08dc2d942e60a4dbc :tada: 

## Copy and paste a paragraph till we reach ~300 lines

**master**
master

**as-char-measurement**
dev

**Conclusion**: I cannot seem to find any significant difference between the two. Fps are similar too and therefore I think we haven't hit a performance regression here :tada: 

## Scroll down line-by-line on ~200 lines

**master**
master

**as-char-measurement**
dev

**Conclusion**: I'd say that we have a performance degradation here, but it's not really significant. Earlier on I had introduced a performance optimization to avoid calculating lines widths while scrolling or for previously calculated lines. The main drawback of this approach is that we cannot apply it to cursor scrolling. One approach may be to simply skip already measured lines: however we have to keep in mind that sub-pixel rendering may occur if, for example, a line has changed position, and therefore we need a always fresh value.

## Scroll down via mouse on ~3000 lines

**master**
master

**as-char-measurement**
dev

**Conclusion**: the same considerations of the above scenario apply here.

----------------------------

All in all, I'd say that the editor feels quite responsive. It's definitely a pity to not be able to measure text before actually adding it to the DOM, but I am not sure we have many alternatives here.

@nathansobo: I recall having a conversation with you about the usefulness of `<canvas>`. I know, it would be a quite drastic change, but it would allow us to [measure text](http://www.w3schools.com/tags/canvas_measuretext.asp) before actually rendering it on screen. That API does not seem to have sub-pixel rendering support: however, it's something we could workaround ourselves [by turning floating points into integers](http://www.html5rocks.com/en/tutorials/canvas/performance/) (see **Avoid floating point coordinates**), so that we can be sure that no mismatch occurs between our measured text and the rendered one. Something to keep in mind for the future?

I'd :heart: to have your feedback on this. Thank you :bow:

nathansobo:
Looking good. What does typing a character look like?

nathansobo:
Scrolling down via mouse looks to have slowed down quite a lot. I wonder if the presenter could have some state that maintains what positions we'll actually need to measure and only measure those locations. Something like `screenRangesToMeasure` or something.

as-cii:
> I wonder if the presenter could have some state that maintains what positions we'll actually need to measure and only measure those locations. Something like `screenRangesToMeasure` or something.

That's an optimization we could make and I had that in mind too, but quoting what I've written in my previous comment:

> One way to gain some speed here may be to simply skip already measured lines: however, we need to keep in mind that **sub-pixel rendering may occur** if, for example, a line has changed position, and therefore we need an always fresh char width value.

So, basically, even though we have already measured characters' width for a certain line, when scrolling up or down we have no guarantee that such widths are correct because sub-pixel rendering *may* have happened. What do you think?

as-cii:
screen shot 2015-03-27 at 09 48 35

While scrolling via mouse we're now operating as fast as **master** :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend:

Basically, we now simply avoid to mark lines as changed while scrolling: we just keep in mind that we'll need to change them once scrolling ends in `TextEditorPresenter`. This solves also the sub-pixel rendering issue I was mentioning earlier on, because right before any user action (after scrolling has ended) we will compute the char widths.

Specs are :green_heart: and I wrote some new ones for this "changed lines performance improvement". I think, however, that we need to update all the existing specs in `text-editor-presenter-spec.coffee` to use `getPreMeasureState` and `getPostMeasureState`, otherwise they won't honor the *real* contract between `TextEditorPresenter` and `TextEditorComponent`.

@nathansobo: this afternoon I will have limited availability, therefore please feel free to jump on this if you want. I feel like we're quite close to :shipit: and I am looking forward to hearing what you think.

Moreover, I'd like to try this out with sub-pixel rendering enabled: any ideas how we could do that (since it will involve patching again `libchromiumcontent`)?

Thank you!

as-cii:
> Now you can actually just iterate over leaf spans with unique scopes instead of text nodes since you only need computed styles. That could probably save some work.

I ended up assigning a unique id to each token. I am not super happy with how I access it, given that I use `document.querySelector` (with the `/deep/` marker). All in all, however, it performs quite nicely and it's faster compared to the iterator-based version. I am open to suggestions on how to make this even better.

Our most performance critical area is scrolling: one thing to keep in mind is that now we do an additional operation (i.e. measuring lines) while rendering lines, so it'll be difficult to reach the same exact level of smoothness we had before these changes.

*(Benchmarks performed on `text-editor-component-spec.coffee`)*

**master**
screen shot 2015-04-01 at 18 44 05

**as-char-measurement**
screen shot 2015-04-01 at 18 37 52

There's a significant performance boost compared to the graphs I showed earlier on. Still, however, we're not always operating at 60fps but I do think we've got several things we may do better. I'll write below a series of facts and, sometimes, possible solutions to explore.

* Things get garbage collected during animation frames. It is the cause of some of those spikes you can see in the graph, but I still need to perform benchmarks to see how memory is actually allocated and how we can make a better use of it.
* On average, measuring lines usually takes less than 6ms. I am not sure how much we can reduce this: in other words it means that we may have only ~10ms to perform everything else.
* Updating scroll top causes both a reflow and a layout (~7ms). What about ditching the default scrollbar for a virtual one (which would not trigger any expensive operation)? Positioning elements will become slightly more complicated, but we could save something by doing that. 
* The remaining time is spent updating line nodes (~11ms). I conducted a quick experiment with `DocumentFragment`: apparently, however, creating nodes by hand is slower than appending strings as we do now. I must admit I didn't dig as much as I'd have liked on this, but it would be really awesome to find a cheaper alternative. An idea could be to have a pool of objects, instead of  always creating and removing new lines (this could even solve some GC slowdowns?), but I haven't any metrics at hand to confirm it would work.

@nathansobo: I know there's quite a lot of stuff in this comment but I wanted to keep you posted about this. I am going to spend some more time on it, especially trying out some of the optimizations I outlined above. I am sorry if this is taking longer than expected: I am truly looking forward to see this :ship:ped as soon as possible, and I feel like we're halfway there :fire:. 

Any suggestion or comment is always welcome and appreciated :bow: :bow: 

Thank you.

nathansobo:
No worries about timing. I'm keeping an eye on this and want to support you, but this is still elective surgery for at least a few more Chrome releases.

nathansobo:
@as-cii What's the status on this? Do you think this is ready to focus on merging? As long as we don't lose performance I think it would be better to fix this sooner rather than later.

as-cii:
@nathansobo: as discussed yesterday, I am still working to make this mergeable and this will hopefully happen by the end of this week :fire: 

Right now there are still a few things I need to address. Specifically:

1. [ ] :green_heart: Look into two tests that are currently failing for no apparent good reason (it should be quite straightforward, though);
2. [ ] :racehorse: Avoid computing scroll width in `DisplayBuffer` every single time `setCharLeftForPoint` is called;
3. [ ] :bug: Scrolling from model (i.e. `editor.setScrollTop`) should trigger a character measurement on newly built lines;
4. [ ] Resolve conflicts with `master`

In the meantime, however, it would be great if you could start having a look at this and leave some comments wherever you feel we should change something. Code-wise this is not perfect yet and, since it has been up for a while and I experimented a lot, there may be a bunch of things we want to refine and do better.

Thanks! :bow:

src/display-buffer.coffee
```diff
@@ -293,36 +293,39 @@ class DisplayBuffer extends Model
...
     fn()
     @batchingCharacterMeasurement = false
-    @characterWidthsChanged() if oldChangeCount isnt @scopedCharacterWidthsChangeCount
+    @handleCharacterWidthsChanged() if @characterWidthsChanged
 
-  setScopedCharWidth: (scopeNames, char, width) ->
-    @getScopedCharWidths(scopeNames)[char] = width
-    @scopedCharacterWidthsChangeCount++
-    @characterWidthsChanged() unless @batchingCharacterMeasurement
+  setCharLeftPositionForPoint: (row, column, charWidth) ->
```
as-cii: I am not super happy with the name of this method. I couldn't come up with something clever and everything always felt like a mouthful. Any ideas how could we improve it? :thought_balloon:

nathansobo: Only a slight improvement: `setLeftPixelPositionForPoint`. I think emphasizing *pixels* is a good idea.

aceLren:
Will this address #3821?  Ligatures broke a few versions back and it would be great to have it for fonts like Hasklig and Fire Code.

nathansobo:
Yeah, it should fix ligatures. We need to try something more efficient than the current state of this PR because it has a bit too much overhead.

nathansobo:
Wanted to provide a status update on this because implementing correct character measurement is becoming a bottleneck for other efforts, such as CJK-aware softwrap. @as-cii feel free to chime in and fill in any details I may be missing.

@as-cii and I had a screen sharing session about this PR a few weeks back, and concluded that the measurement overheads for the current state of this PR are probably too high. Our overall goal is to be able to compute the horizontal pixel position of a column at any time without reading from the document (which causes reflows).

* The existing solution on master caches the width of each character in each scope whenever the DOM is updated, allowing us to compute the horizontal location of a column by summing up characters and reading out of this cache. Unfortunately, this approach is broken with ligatures and doesn't work when subpixel font scaling is enabled.

* The solution on this PR is to use the canvas API to measure text. The advantage to measuring via canvas is that we don't interact with the document at all, but unfortunately, the canvas text measurement API seems to be too slow for our purposes, requiring every prefix of the line to be measured in a separate call.

* The idea @as-cii and I explored on our call is to instead perform measurement in a separate document, via an iframe. We would need to *sample* styling information for every unique scope in the document via dummy DOM nodes, similar to @abe33's approach for the [minimap](https://github.com/abe33/atom-minimap). Of course all we're interested in is font styling, since that's the only thing that affects measurement. Then, to measure the position of all columns on a line, we would construct a simpler version of that line in a different document with the same styling applied inline, then use the DOM range API to iterate over the characters and measure. The advantage of this is that we only need to perform layout once.

Will this new approach work? Not sure. But it's the next thing we'd like to try. Other ideas are welcome, but remember that it can't rely on reading from the document.

superlou:
@nathansobo: Will this approach allow positioning on lines that have margins or padding, or allow some other means of changing softwrap boundaries throughout the text editor?

nathansobo:
@superlou I'm not sure how much CSS applied to lines it's going to be realistic to support. Any supported styling will need to be sampled for measurement in the other document, and covering every possible thing people may want to style positioning wise could end up being pretty complicated. So we can try, but we may need to stick to CSS for styling fonts and colors and add other editor-specific mechanisms for controlling soft wrap behavior etc. We'll see how it goes.

larsenwork:
Not sure if it's related to this but I'm having issues with opentype `calt ` stopped working

nathansobo:
@larsenwork Nothing has really changed in this area for a while, so I'd say it's unlikely. You can still open another issue however.

alexandernst:
Is this still a thing?

nathansobo:
@alexandernst Yep. It's still happening. @as-cii has been churning through a zillion alternatives looking for the best approach. It's looking increasingly like just eating a small forced reflow might be the best approach. We're getting there.

PR #6087, opened 2015-03-25
fix typo

rossPatton:
`apm test` -> `npm test`

no contribution too small?


50Wliu:
I believe it is supposed to be `apm test`.

as-cii:
@rossPatton: thank you for taking the time to contribute :+1: 

However, @50Wliu is right: `apm` stands for `Atom Package Manager` and handles package and core testing as well. This change in docs was introduced with 2aa566b7c6eaa81b5650e7e1e7ece6cc68a10eac and was actually intended.

Thanks :bow:

PR #6098, opened 2015-03-25
Enable deprecations in all windows

kevinsawicki:
Previously deprecations were only shown in dev and test windows.

Removing this check enables it for all windows.

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -207,12 +207,6 @@ class Atom extends Model
   #
   # Call after this instance has been assigned to the `atom` global.
   initialize: ->
-    # Disable deprecations unless in dev mode or spec mode so that regular
-    # editor performance isn't impacted by generating stack traces for
-    # deprecated calls.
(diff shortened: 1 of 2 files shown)


src/atom.coffee
```diff
@@ -207,12 +207,6 @@ class Atom extends Model
   #
   # Call after this instance has been assigned to the `atom` global.
   initialize: ->
-    # Disable deprecations unless in dev mode or spec mode so that regular
-    # editor performance isn't impacted by generating stack traces for
-    # deprecated calls.
-    unless @inDevMode() or @inSpecMode()
-      require('grim').deprecate = ->
```
kevinsawicki: @nathansobo I'm guessing this probably should have been removed after https://github.com/atom/grim/pull/4 was merged and just never was. Any reason not to :fire: it now?

nathansobo: Not that I know of.

PR #6099, opened 2015-03-25
Add support for jasmine 2.0

maxbrunsfeld:
#### Motivation

Currently, any npm modules we work on (e.g. `text-document`) need to use jasmine 1.3 so we can run specs in Atom, which is really useful for debugging. Without Atom's monkey-patched `waitsFor`, Jasmine 1.3 is basically unusable for testing async code (it only supports polling, no callbacks). Rather than have `text-document` and other libraries vendor atom/atom's customized version of jasmine 1.3, we should just use jasmine 2.
#### Changes

This allows packages and npm-modules to opt-in to using Jasmine-2.0 for the `application:run-package-specs` command in Atom. Packages opt in by adding `atom-jasmine: "^2.0.0"` to the `engines` object in their `package.json`.

Currently, when using Jasmine 2, `spec-helper` is not loaded. I think it would be nice to eventually stop using the miscellaneous global functions defined in that file. Some of them (e.g. the non-standard `jasmine.advanceClock`) duplicate functionality that is built into jasmine.

I am thinking that some community members might be interested in upgrading individual package's test suites to use Jasmine-2.0. We could make `good-first` issues for doing that in various packages.
- [x] Make the `AtomReporter` support the jasmine 2 reporter API
- [x] Make sure `fit`, `fdescribe` et al work w/ jasmine 2
- [x] Make sure `apm test` works w/ jasmine 2
#### Questions
- [ ] This PR also removes the `application:run-all-specs` command. It doesn't seem like that command is used anymore, and it would have been difficult to preserve it while adding support for Jasmine-2.0 in the 
  AtomReporter. Also, it would be very difficult to support that command's functionality if individual packages started to migrate to Jasmine-2.0.
- [ ] Is the `engines.atom-jasmine` the right way to do this?

/cc @atom/core @atom/non-github-maintainers I'm not sure if it's overly aggressive to get rid of `run-all-specs` command. What do you think?

Refs #5393.

### DIFF
--- a/spec/spec-bootstrap.coffee
+++ b/spec/spec-bootstrap.coffee
@@ -8,8 +8,11 @@ atom.getCurrentWindow().show() unless atom.getLoadSettings().exitWhenDone
 
 try
   document.title = "Spec Suite"
-  {runSpecSuite} = require './jasmine-helper'
-  runSpecSuite('./spec-suite', atom.getLoadSettings().logFile)
+  {runSpecSuite, runSpecSuiteV2} = require './jasmine-helper'
+  if atom.getLoadSettings().useJasmine2
+    runSpecSuiteV2('./spec-suite', atom.getLoadSettings().logFile)
+  else
+    runSpecSuite('./spec-suite', atom.getLoadSettings().logFile)
 catch error
   if atom?.getLoadSettings().exitWhenDone
     console.error(error.stack ? error)
--- a/spec/spec-bootstrap.coffee
+++ b/spec/spec-bootstrap.coffee
@@ -1,27 +1,15 @@
-# Start the crash reporter before anything else.
-require('crash-reporter').start(productName: 'Atom', companyName: 'GitHub')
+require '../src/window'
+Atom = require '../src/atom'
+window.atom = Atom.loadOrCreate('spec')
 
-path = require 'path'
+# Show window synchronously so a focusout doesn't fire on input elements
+# that are focused in the very first spec run.
+atom.getCurrentWindow().show() unless atom.getLoadSettings().exitWhenDone
 
 try
-  require '../src/window'
-  Atom = require '../src/atom'
-  window.atom = Atom.loadOrCreate('spec')
-
-  # Show window synchronously so a focusout doesn't fire on input elements
-  # that are focused in the very first spec run.
-  atom.getCurrentWindow().show() unless atom.getLoadSettings().exitWhenDone
-
-  {runSpecSuite} = require './jasmine-helper'
-
-  # Add 'exports' to module search path.
-  exportsPath = path.join(atom.getLoadSettings().resourcePath, 'exports')
-  require('module').globalPaths.push(exportsPath)
-  # Still set NODE_PATH since tasks may need it.
-  process.env.NODE_PATH = exportsPath
-
   document.title = "Spec Suite"
-  runSpecSuite './spec-suite', atom.getLoadSettings().logFile
+  {runSpecSuite} = require './jasmine-helper'
+  runSpecSuite('./spec-suite', atom.getLoadSettings().logFile)
 catch error
   if atom?.getLoadSettings().exitWhenDone
     console.error(error.stack ? error)
--- a/spec/spec-bootstrap.coffee
+++ b/spec/spec-bootstrap.coffee
@@ -8,11 +8,8 @@ atom.getCurrentWindow().show() unless atom.getLoadSettings().exitWhenDone
 
 try
   document.title = "Spec Suite"
-  {runSpecSuite, runSpecSuiteV2} = require './jasmine-helper'
-  if atom.getLoadSettings().useJasmine2
-    runSpecSuiteV2('./spec-suite', atom.getLoadSettings().logFile)
-  else
-    runSpecSuite('./spec-suite', atom.getLoadSettings().logFile)
+  {runSpecSuite} = require './jasmine-helper'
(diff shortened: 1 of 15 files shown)


thomasjo:
OMG OMG OMG! :heartpulse:

thomasjo:
> 'm not sure if it's overly aggressive to get rid of that command. What do you think?

I have no problems with removing it. It hasn't worked for me for a long while now. From a UX perspective I also consider it to be potentially confusing. Let it :fire:

thomasjo:
> Is the engines.atom-jasmine the right way to do this?

I like the this manner of opting in, since it leaves room for future upgrades to Jasmine, and hopefully make it easier. Using the *engines* block seems to make sense. Can't really think of a better approach right now.

kevinsawicki:
> Is the engines.atom-jasmine the right way to do this?

The only other idea I have is looking for a 2.0 compatible `jasmine-core` entry in `devDependencies` since it kind of feels more like a dependency requirement than an engine requirement.

spec/spec-bootstrap.coffee
```diff
@@ -1,27 +1,15 @@
-# Start the crash reporter before anything else.
-require('crash-reporter').start(productName: 'Atom', companyName: 'GitHub')
+require '../src/window'
```
kevinsawicki: I think these requires and the `Atom.loadOrCreate` call should still be in a try block so it exits on caught errors or shows the window, unless something else has changed where that is no longer an issue.

maxbrunsfeld:
Yeah, that would work too. I was thinking it might confuse people though, since they might not depend on `jasmine-core` directly, but through an intermediate module like `jasmine` or `jasmine-node`. Is there a clean way for us to recursively look for `jasmine-core` in their dep graph?

kevinsawicki:
> Is there a clean way for us to recursively look for jasmine-core in their dep graph?

Not really, but maybe just start with two checks, one for jasmine-core and one for jasmine, jasmine-node won't be an issue to start since it doesn't support jasmine 2 yet.

nathansobo:
What about packages that want to use other testing frameworks, such as mocha. Could a package declare an entry point for tests instead, similar to how `main` indicates an entry point for its main module? If such an entry point were defined, we would use it instead of our default runner infrastructure.

thedaniel:
cc @as-cii and @sedabull who did some work on #5393

thomasjo:
> What about packages that want to use other testing frameworks, such as mocha. Could a package declare an entry point for tests instead, similar to how main indicates an entry point for its main module? If such an entry point were defined, we would use it instead of our default runner infrastructure.

I really like this idea.

as-cii:
> What about packages that want to use other testing frameworks, such as mocha. Could a package declare an entry point for tests instead, similar to how main indicates an entry point for its main module? If such an entry point were defined, we would use it instead of our default runner infrastructure.

@nathansobo: :+1: That would definitely remove quite a lot of headaches related to testing, hopefully enabling us to change how we test things in the future independently from how packages do. To enforce this flexibility, we may even provide a default testing configuration when creating new packages, so that package developers don't have to choose their own without relying on Atom's one.

I think this is the kind of flexibility we should give to people and I am really looking forward to see this :ship:ped. @maxbrunsfeld: what do you think? I am open to collaborating with you on this if you want to :smiley_cat: 

By the way, I think this will eventually supplant #5393 and I am going to close it :fire:

sedabull:
@nathansobo :+1: I really like this idea!

maxbrunsfeld:
Ok, I'm going to close this one for now. I like @nathansobo's idea of moving this functionality to packages, rather than adding to our existing jasmine-based infrastructure. For testing non-atom npm modules, https://github.com/nathansobo/atom-jasmine-runner works for now.

Glavin001:
I apologize for bumping this, however I have been looking around and do not yet have a solution. I am fine with using Jasmine, although I prefer Mocha, however most of my tests are async and it is not pleasant without Jasmine 2.0's async support. 

Is an implementation of @nathansobo's solution already in progress? This issue is closed but I do not see where to continue. I do not see any issues referenced to this one in the tread above and [the docs do not provide nicer async functions](https://atom.io/docs/latest/hacking-atom-writing-specs#asynchronous-functions-with-callbacks) so I do not know the status of this issue. Specs are important and putting the choice of testing framework in the hands of the package is a great idea. Hopefully this can be a reality soon. Thanks!

nathansobo:
Yes, our current testing situation is a bit messy. As you can imagine its one of the oldest areas of the codebase and could use improvement.

I think the way forward is to create a package that runs jasmine tests for a project in a clean JavaScript context, maybe even an iframe in an Atom pane. Ideally, we'd also have a test helpers library that could help setup Atom globals when people wanted them. Or authors could instantiate their own versions of these globals and use dependency injection. The main thing I'd like to get away from is all the default setup we're doing before every test run.

This can definitely be done in a package, and if you're interested in experimenting, a [quick package](https://github.com/nathansobo/atom-jasmine-runner) I wrote may be a place to start. It uses a different window and last time I tried it was crashing Electron. But the same ideas could be adapted to use an iframe in a pane pretty effectively.

If you want to fork that, copy and paste code, or open pull requests, that may be a way forward.

david-driscoll:
Any idea how get the helpers without adding anything to atom core itself?  Looking at the source, at the very least need a way to get a hold of the AtomWindow constructor... not sure how to do that at the moment.

Would love to be able to use mocha for testing my packages if it were possible :)

Issue #5164, opened 2015-01-20
Remove deprecated APIs


benogle:

fulldecent:
Link: https://github.com/atom/atom/search?q=deprecated&type=Code&utf8=%E2%9C%93

thomasjo:
I started working on this in #8682.

mnquintana:
Closed by https://github.com/atom/atom/pull/8682

PR #6103, opened 2015-03-25
1.0 API preview mode

kevinsawicki:
This pull request adds a new command line flag (`-1` or `--one`) that will launch Atom with no core deprecated APIs so you can test out the 1.0 API experience where APIs will be missing instead of present but logging deprecations.

So things like `atom.workspaceView` will be unavailable and so you can use this flag to test out that your packages will work correctly when 1.0 comes out.

The basic flow here is to check `Grim.includeDeprecatedAPIs` in places and to co-locate all the prototype methods at the bottom of the file so they can all be deleted easily once 1.0 ships.
### Deprecations To Update
- [x] atom/atom
- [x] https://github.com/atom/atom-keymap/pull/77
- [x] https://github.com/atom/event-kit/pull/5
- [x] https://github.com/atom/first-mate/pull/41
- [x] markdown-preview
- [x] https://github.com/atom/node-pathwatcher/pull/68
- [x] https://github.com/atom/scoped-property-store/pull/13
- [x] https://github.com/atom/text-buffer/pull/55
- [x] wrap-guide

Refs #5164


benogle:
Gnar

kevinsawicki:
I'm going to merge this now, there might be a few places that still need `Grim.includeDeprecatedAPIs` guards but managing this diff given the churn on master will only be harder the longer this sits.

The specs are passing which at least verifies that no methods called by the specs were accidentally deleted and that the deprecation blocks are still running as expected.

PR #6132, opened 2015-03-27
Bump deprecation-cop@0.38.0

jlord:
Use the latest version of `deprecation-cop` which uses the latest version of the `status-bar` Service API. :cat2:


jlord:
greens! :green_heart:

PR #6167, opened 2015-03-30
Fix screenPositionForPixelPosition for case above first row.

### LINKED ISSUE #6168: Selecting text in atom based text editor doesn't work like native platform
In particular on Mac OS X if you:

1. start creating a drag selection.

2. Drag to above the text editor.

3. Then the text selection will always extend to the 0 column of the text.

But in Atom this doesn't happen. Instead the selection extends to the column based on the mouse x position, even when the mouse y position is above the text editor. (Atom does correctly handle the case (set selection to end of text) when the mouse extends below the text editor... it's only the above case that's not working correctly).

This pull request fixes the issue:

https://github.com/atom/atom/pull/6167

jessegrosjean:
Currently screenPositionForPixelPosition special cases pixel positions
below the last row so that the column position is always set to the
last column of the last line, even if the pixel 'x' position is less
then that column.

This patch special cases picks above the first row so that the  text
column position will be in the first column even if the pixel 'x'
position is greater then that column.

At a higher level, this patch fixes the problem where you can’t select
to the start of a text field by just clicking and dragging up. Instead
you have to click and drag back (x axis) beyond the start of the text
field.

Closes #6168


benogle:
Can you write a spec for this?

jessegrosjean:
Not sure if you get automatically notified anyway, but I just added a spec for this commit.

benogle:
We dont, so thanks for the ping

spec/display-buffer-spec.coffee
```diff
@@ -724,6 +724,15 @@ describe "DisplayBuffer", ->
         expect(displayBuffer.clipScreenPosition([0, 1], clip: 'forward')).toEqual [0, tabLength]
         expect(displayBuffer.clipScreenPosition([0, tabLength], clip: 'forward')).toEqual [0, tabLength]
 
+  describe "::screenPositionForPixelPosition(pixelPosition)", ->
+    it "pixel positions above buffer to map to start", ->
+      expect(displayBuffer.screenPositionForPixelPosition(top: -Infinity, left: -Infinity)).toEqual [0, 0]
+      expect(displayBuffer.screenPositionForPixelPosition(top: -Infinity, left: Infinity)).toEqual [0, 0]
+
+    it "pixel positions below buffer map to end", ->
```
as-cii: :bulb: Do you think we could make these specs a bit more descriptive? I was thinking to something like:

```coffee
it "clips pixel positions above buffer start"
it "clips pixel positions below buffer end"
```

Maybe we could even add some more expectations, just to make clear that we don't treat `Infinity` as a special case, e.g.

```coffee
expect(displayBuffer.screenPositionForPixelPosition(top: -1, left: Infinity)).toEqual [0, 0]
expect(displayBuffer.screenPositionForPixelPosition(top: -1, left: 4)).toEqual [0, 0]
```

Nice job here, by the way :wink:

as-cii:
:+1: 

I've read through the linked issue, and I'd say this behavior proposal is quite consistent with how text editors usually handle such scenario.

@jessegrosjean: it'd be great to add a spec covering this case also in `text-editor-component-spec.coffee`. If you feel like tackling this, you can take inspiration from how other specs are written in `describe "mouse interactions on the lines"` (I'll be happy to help you if you encounter any issue, thus please feel free to ask any question) :smiley_cat:

benogle:
@as-cii feel free to merge this when you're happy with it.

jessegrosjean:
@as-cii I've just added some updated specs for this.

src/display-buffer.coffee
```diff
@@ -692,6 +692,7 @@ class DisplayBuffer extends Model
     targetLeft = pixelPosition.left
     defaultCharWidth = @defaultCharWidth
     row = Math.floor(targetTop / @getLineHeightInPixels())
+    targetLeft = -Infinity if row < 0
```
jessegrosjean: I'm not sure. I guess that would be OK, but as I don't really understand the internals of display-buffer (and fact that Infinity is used for the positive overflow case) I think Infinity might make more sense. Maybe it doesn't work this way, but I can imagine cases where there's some negative text indent that causes lines of text to layout in negative x territory, and so that's why Infinity seems a bit safer.

as-cii: > I'm not sure. I guess that would be OK, but as I don't really understand the internals of display-buffer (and fact that Infinity is used for the positive overflow case) I think Infinity might make more sense. Maybe it doesn't work this way, but I can imagine cases where there's some negative text indent that causes lines of text to layout in negative x territory, and so that's why Infinity seems a bit safer.

`Infinity` is the highest value possible and it's used to ensure we stop at the end of the line while iterating. In this scenario, however, we actually know in advance what the beginning of the line is and, as a result, I'd say we can use `0` quite safely. As of today we don't support negative screen positions and using `-Infinity` would be misleading when reading the code later on.

@jessegrosjean: what do you think?

jessegrosjean:
Ok, thanks for feedback. I've updated to retarget to 0 instead of -Infinity.

PR #6197, opened 2015-04-01
Allow overlay decorations to break out of the editor; bound overlays to window

benogle:
Here is an overlay outside the editor bounds with a `-250px` `left-margin`:

screen shot 2015-03-31 at 7 13 04 pm

Here is autocomplete using the new overlay. Note how it is up against the window, and it is shifted horizontally and flipped vertically to keep within the window.

autocomplete-recompute-better
### Notes
- The overlay parent is the `atom-text-editor`. The editor and containing pane are now both `overflow: visible`. This allows the pane item the flexibility to have things jump out if necessary.
- When bumping up against the left or right edge of the window, the overlay will _shift_ horizontally, not flip. This always keeps it in the window frame. Shifting felt more natural than flipping.
- When bumping up against the bottom edge of the window, the overlay will flip. It will not necessarily be always visible in the window frame (due to scrolling), but does its best with the flip.
- A `left-margin` is supported on the overlay. It will take the margin into account when doing the window bounds-checking. Note the blue box in the first image has a `-250px` left margin.
- In the future, it would be nice to allow configuration of the shift or flip, or bounds checking at all, but for right now, keeping it simple.
### TODO
- [x] Make it work
- [x] Clean up the overlay caching :art:
- [x] Move specs to the presenter
- [x] All specs passing

cc @nathansobo @bolinfest

### DIFF
--- a/src/overlay-manager.coffee
+++ b/src/overlay-manager.coffee
@@ -4,99 +4,38 @@ class OverlayManager
     @overlaysById = {}
 
   render: (state) ->
-    editorDimensionsHaveChanged = !@editorDimensionsAreEqual(state)
-
     for decorationId, overlay of state.content.overlays
-      overlayHasChanged = not @overlayStateIsEqual(decorationId, overlay)
-      if editorDimensionsHaveChanged or overlayHasChanged
+      if @shouldUpdateOverlay(decorationId, overlay)
         @renderOverlay(state, decorationId, overlay)
-        @cacheOverlayState(decorationId, overlay)
 
     for id, {overlayNode} of @overlaysById
       unless state.content.overlays.hasOwnProperty(id)
         delete @overlaysById[id]
         overlayNode.remove()
 
-    @cacheEditorDimensions(state)
-
-  overlayStateIsEqual: (decorationId, overlay) ->
-    return false unless @overlaysById[decorationId]?
-    @overlaysById[decorationId].itemWidth is overlay.itemWidth and
-      @overlaysById[decorationId].itemHeight is overlay.itemHeight and
-      @overlaysById[decorationId].contentMargin is overlay.contentMargin and
-      @overlaysById[decorationId].pixelPosition?.top is overlay.pixelPosition?.top and
-      @overlaysById[decorationId].pixelPosition?.left is overlay.pixelPosition?.left
-
-  cacheOverlayState: (decorationId, overlay) ->
-    return unless @overlaysById[decorationId]?
-    @overlaysById[decorationId].itemWidth = overlay.itemWidth
-    @overlaysById[decorationId].itemHeight = overlay.itemHeight
-    @overlaysById[decorationId].contentMargin = overlay.contentMargin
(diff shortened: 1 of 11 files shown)


src/overlay-manager.coffee
```diff
@@ -1,38 +1,101 @@
...
+    return false unless @cachedEditorDimensions?
+    @cachedEditorDimensions.lineHeight is @presenter.lineHeight and
+      @cachedEditorDimensions.contentFrameWidth is @presenter.contentFrameWidth and
+      @cachedEditorDimensions.editorTop is @presenter.boundingClientRect?.top and
+      @cachedEditorDimensions.editorLeft is @presenter.boundingClientRect?.left and
+      @cachedEditorDimensions.editorWidth is @presenter.boundingClientRect?.width and
+      @cachedEditorDimensions.windowWidth is @presenter.windowWidth and
+      @cachedEditorDimensions.windowHeight is @presenter.windowHeight and
+      @cachedEditorDimensions.scrollTop is state.content.scrollTop and
+      @cachedEditorDimensions.scrollLeft is state.content.scrollLeft
```
benogle: We _could_ ignore all this caching, and render them all the time...

If keeping the cache, maybe it goes in the presenter?

src/text-editor-presenter.coffee
```diff
@@ -994,6 +1012,20 @@ class TextEditorPresenter
 
       regions
 
+  setOverlayDimensions: (decorationId, itemWidth, itemHeight, contentMargin) ->
```
benogle: I'm caching the overlay dimensions in the state. Abuse? I feel it is cleaner than keeping track of two structures for the overlays.

src/overlay-manager.coffee
```diff
@@ -1,38 +1,101 @@
...
+  measureOverlays: ->
+    for decorationId, {item} of @overlaysById
+      @measureOverlay(decorationId, item)
 
-  renderOverlay: (state, decorationId, item, pixelPosition) ->
+  measureOverlay: (decorationId, item) ->
+    contentMargin = parseInt(getComputedStyle(item)['margin-left']) ? 0
+    @presenter.setOverlayDimensions(decorationId, item.offsetWidth, item.offsetHeight, contentMargin)
+
+  renderOverlay: (state, decorationId, {item, pixelPosition}) ->
```
nathansobo: Most of the logic in this method should probably be moved into the presenter. You shouldn't need to cache anything in this component other than the last rendered overlay state. That way, the presenter can decide when this overlay state needs to be updated based on dimensions that get written into it and you don't need to cache a bunch of stuff at this layer.

src/text-editor-presenter.coffee
nathansobo: For clarity and consistency, the state object should only be used to store information that is consumed by the view. Since this information came from the view, it doesn't seem like it should be placed here. It should be just one more object for maintaining these dimensions, right? That means state update methods are free to blow away and reconstruct state without worrying about destroying data that's required to build the state. That might not happen here, but keeping a separation as a general principle means any future changes are free to reason about the state as a write-only artifact.

PR #5499, opened 2015-02-11
Export File and Directory through require('atom')

jessegrosjean:
This exports the pathwatcher File and Directory classes through require('atom').

I think this makes sense since those classes are already publicly documented as part of the [Atom platform API](https://atom.io/docs/api/v0.178.0/Atom), but currently there's no way for a package to require and instantiate them. The alternative is to add the pathwatcher module to my own package, but this seems redundant since File and Directory are already documented as part of the Atom platform.

PR #6202, opened 2015-04-01
Export File and Directory through require('atom')

jessegrosjean:
This exports the pathwatcher File and Directory classes through require('atom').

I think this makes sense since those classes are already publicly documented as part of the Atom platform API, but currently there's no way for a package to require and instantiate them. The alternative is to add the pathwatcher module to my own package, but this seems redundant since File and Directory are already documented as part of the Atom platform. Also since pathwatcher is a native package it's even more problematic then it would be with a normal package.

### DIFF
--- a/exports/atom.coffee
+++ b/exports/atom.coffee
@@ -1,5 +1,6 @@
 TextBuffer = require 'text-buffer'
 {Point, Range} = TextBuffer
+{File, Directory} = require 'pathwatcher'
 {Emitter, Disposable, CompositeDisposable} = require 'event-kit'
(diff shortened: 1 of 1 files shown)


izuzak:
@atom/core any objections to this? I think it would be useful -- I'd use this in my package as well, where problems like https://github.com/atom/atom/pull/5499#issuecomment-75169244 happen.

maxbrunsfeld:
Sounds good to me.

kevinsawicki:
:+1:

izuzak:
Thanks for this, @jessegrosjean

PR #6208, opened 2015-04-01
Add styleguide for specs

jlord:
A wee PR here to add a section on spec writing. Coming from a suggestion from @nathansobo [here](https://github.com/atom/markdown-preview/pull/220#issuecomment-84234793).

Nathan, I tweaked the orginal dog scenario to be a happier one :dog: :grinning: 

[r E n D e R e D](https://github.com/atom/atom/blob/0f43b246b4ef5c1759c81440e8b0b2abbf9230d3/CONTRIBUTING.md)


nathansobo:
Looking good. Would it be worth including real syntax, (that is the `, ->`s) in the example spec structure?

jlord:
Ha, yes I think that would be useful! Adding...

PR #6220, opened 2015-04-02
Always mutate selections in order

### LINKED ISSUE #6190: Multiple cursors shuffle order on paste when selecting characters on more than 7 lines
atom-mult-cursor-bug
Hi! 
I noticed a bug when selecting 7 or more lines with multiple cursors. In the gif above, I press ctrl-alt-down 7 times, then once shift-right to select the numbers, then ctrl-c to copy, then esc to go back to 1 cursor. When I press ctrl-v on an empty line, the just copied numbers are shuffled.

as-cii:
Closes #6190 

When copying, cutting, pasting and mutating selections in general, we were accessing selections as they were added to the editor. This caused unexpected behaviors, especially when dealing with multi-line selections.

With this pull-request we now compute all the aforementioned operations ordering the selections first. This is also consistent with how Sublime Text works :sparkles:

/cc: @nathansobo

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2730,6 +2730,20 @@ describe "TextEditor", ->
 
             """
 
+        describe "when many selections get added in shuffle order", ->
+          it "cuts them in order", ->
+            editor.setSelectedBufferRanges([
+              [[2,8], [2, 13]]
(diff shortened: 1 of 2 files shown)


nathansobo:
Ah, great idea!

Issue #5982, opened 2015-03-16
Lint package builds on Travis


kevinsawicki:
We currently have coffee linting set up to run on core CI builds but package CI builds don't have this.

This causes problems where green PRs get merged on packages but they fail when updated in core since the lint task fails.

mnquintana:
![image](http://www.quickmeme.com/img/d4/d4a888dd9facf35543ae429f296cda146c8239223681279804642557db0fa75c.jpg)

I should've gotten everything – feel free to reopen / ping me if I missed anything @atom/feedback!

kevinsawicki:
:blue_heart: :boat:

Issue #6222, opened 2015-04-02
Setup CI on Travis and AppVeyor


as-cii:
As discussed on Slack, it would be great to make CI available for everyone :heart:

Ideally we should be able to provision a cibuild on both Travis and AppVeyor, so that we can check if specs pass on Windows and *nix environments.

mnquintana:
:+1: Yesssss this would be awesome!

thomasjo:
The free AppVeyor offering is such a slow, unstable product though. I've had to disable it for my projects since it produced so many false negatives. Randomly failed downloading and installing Atom for instance. They really need to improve their stability, or the Atom team needs to work out some sort of deal to get access to AppVeyor's "Pro" infrastructure which is reportedly more stable.

paulcbetts:
@thomasjo It's not :(

PR #6230, opened 2015-04-03
Lint for more styleguide errors

mnquintana:
This PR adds more coffeelint rules to cover code style requirements from CONTRIBUTING.md, and one rule that wasn't explicitly defined in CONTRIBUTING.md, but was already used everywhere in the codebase.

Refs atom/settings-view#449, atom/settings-view#44, #5982 

/cc @kevinsawicki I wanted to update the coffeelint rules first in core before I go and add coffeelint.json to all the package repos.


benogle:
Note that the colon spacing rule will annoyingly fail in a regex containing
a colon.

mnquintana:
@benogle Aww :cry: – alright then, I'll take that one out. How does it look otherwise?

mnquintana:
@benogle Is that still the case? I tested it with this regex:

```coffeescript
/[A-Za-z0-9.,:-]*/
```

And didn't get any coffeelint errors.

benogle:
May have been a different version of coffee lint. I had problems in the
autocomplete linter. If it doesn't throw an error, carry on.

kevinsawicki:
I like these changes but the build will fail if this is merged since there are cases currently where these aren't being followed.

Did you want to fix these in this pull request or just change these additions to warnings to start?

mnquintana:
@kevinsawicki I was planning on fixing the code style errors in this PR; I just wanted to make sure the coffeelint changes looked good first. If it all looks good then I'll take care of that when I get home!

kevinsawicki:
>  If it all looks good then

Yup, looks good to me :+1:

mnquintana:
@kevinsawicki This should be ready for review now. (How do you trigger CI builds? I can't tell if it's passing or not.)

spec/fixtures/coffee.coffee
```diff
@@ -13,6 +13,7 @@ class quicksort
       if current < pivot
         left.push(current)
       else
+        # coffeelint: disable=no_trailing_semicolons
```
kevinsawicki: I think we probably want to omit all `spec/fixtures` files from any linting.

I'm guessing adding this line might cause some specs to fail that may be targeting specific lines in this file by number.

kevinsawicki:
> How do you trigger CI builds?

CI builds only run for local branches currently, not branches from forks, sorry, this is being fixed in #6222  though

spec/fixtures/coffee.coffee
mnquintana: That makes sense - we could handle that by adding a `.coffeelintignore` (oh how I wish we could just configure it in `coffeelint.json`), but that would be redundant with the coffeelint grunt task config. So I could just revert the changes to fixtures, if that would work?

kevinsawicki: > So I could just revert the changes to fixtures, if that would work?

That sounds good to me

mnquintana:
@kevinsawicki This should be ready now! (I kept `spec/fixtures/packages/package-with-config-defaults/index.coffee` as is though because it shouldn't do any harm)

benchmark/fixtures/medium.coffee
```diff
@@ -209,13 +209,13 @@ template = (str) ->
...
-       .split('<%').join("');")
-       .split('%>').join("p.push('") +
-       "');}return p.join('');"
+      .replace(/'(?=[^<]*%>)/g,"\t")
+      .split("'").join("\\'")
+      .split("\t").join("'")
+      .replace(/<%=(.+?)%>/g, "',$1,'")
+      .split('<%').join("');")
+      .split('%>').join("p.push('") +
+      "');}return p.join('');"
```
kevinsawicki: Mind backing this out since it is a fixture as well?

spec/fixtures/packages/package-with-config-defaults/index.coffee
```diff
@@ -1,5 +1,5 @@
 module.exports =
   configDefaults:
-    numbers: { one: 1, two: 2 }
+    numbers: {one: 1, two: 2}
```
kevinsawicki: Also mind reverting the changes to this fixture as well?

mnquintana:
@kevinsawicki Addressed the comments!

kevinsawicki:
Awesome, thanks for cleaning these all up :green_heart:

thomasjo:
Thanks for doing this @mnquintana! :sparkling_heart:

PR #6239, opened 2015-04-03
Fix `deleteLine()` when multiple selections are on the same line

### LINKED ISSUE #6108: Delete Line deletes more than one line if the line has multiple selections

as-cii:
Closes #6108.

Before these changes, when a line contained multiple selections and `deleteLine()` was called, two lines would have been deleted instead of just one. This happens because, when the line containing the first selection is deleted, the other selection doesn't get destroyed but it simply moves.

This PR tries to fix this by merging selections that have intersecting rows. However, I do feel like there may be some other elegant solutions to this: I am thinking to something like `Marker` invalidation, for example.

/cc: @nathansobo


src/selection.coffee
```diff
@@ -174,6 +174,18 @@ class Selection extends Model
   intersectsScreenRow: (screenRow) ->
     @getScreenRange().intersectsRow(screenRow)
 
+  # Public: Identifies if this selection's rows intersects with another selection's rows.
```
nathansobo: I'm not sure this belongs in the API. Mainly I'm hanging up on the `ByRow` part of the name feeling a little confusing. What if you just used an existing method in your editor code?

```coffee
selectionA.getScreenRange().intersectsRowRange(selectionB.getScreenRange().getRowRange()...)
```

It's a bit longer but it's nice that the API doesn't gain surface area.

src/text-editor.coffee
```diff
@@ -2232,6 +2233,17 @@ class TextEditor extends Model
   # the function with merging suppressed, then merges intersecting selections
   # afterward.
   mergeIntersectingSelections: (args...) ->
+    @mergeSelections args..., (previousSelection, currentSelection) ->
+      exclusive = not currentSelection.isEmpty() and not previousSelection.isEmpty()
+
+      previousSelection.intersectsWith(currentSelection, exclusive)
+
+  mergeIntersectingSelectionsByRow: (args...) ->
```
nathansobo: The code is great here, but I'm sticking on the naming. The selections we'll merge aren't actually always intersecting, right? They're intersecting-by-row, which feels like a mouthful. What about a name like `mergeSelectionsOnSameRows`?

as-cii: @nathansobo: great suggestion here, it sounds much better! :+1:

as-cii:
Thanks for taking the time to review this: I absolutely agree with your suggestions and I refactored the code accordingly :rainbow: 

My main concern with this solution is that we merge selections in advance just because we don't want them to become invalid during the deletion process. That seems like something that `Selection` could handle on its own, probably through some invalidation strategy at the marker-level.

On the other hand I think I am also fine with merging the code as it is, since it's self-contained and looks reasonably clean to me.

@nathansobo: what do you think? :thought_balloon:

src/text-editor.coffee
```diff
@@ -2048,6 +2049,17 @@ class TextEditor extends Model
...
   # afterward.
   mergeIntersectingSelections: (args...) ->
+    @mergeSelections args..., (previousSelection, currentSelection) ->
+      exclusive = not currentSelection.isEmpty() and not previousSelection.isEmpty()
+
+      previousSelection.intersectsWith(currentSelection, exclusive)
+
+  mergeSelectionsOnSameRows: (args...) ->
+    @mergeSelections args..., (previousSelection, currentSelection) ->
+      previousSelection.intersectsScreenRowRange(currentSelection.getBufferRowRange()...)
```
nathansobo: Careful, you're mixing coordinate spaces here: screen vs. buffer.

nathansobo:
Agreed. One more critical thing to address in my inline comment, then I'd say go ahead and :ship:.

as-cii:
> Agreed. One more critical thing to address in my inline comment, then I'd say go ahead and .

@nathansobo: great catch there! Thank you for the awesome review :fireworks:

PR #6240, opened 2015-04-03
Fix screen/buffer range error in paragraph cursor motions

### LINKED ISSUE #496: Previous Paragraph ("{") is broken
This seemed related to #333 but now I'm not so sure. In previous versions of Atom, Previous paragraph ("{") has been intermittently broken, but as of the latest versions 0.175 / 0.29.0, this is now completely broken, which is making the editor much less usable for me

### LINKED ISSUE #602: {  got broken
Jumping paragraphs with `{` and `}` used to work normally before.
After the last update (atom 0.189.0, vim-mode 0.42.0), I've seen this behavior broken when editing Markdown with live preview on the right side.

maxbrunsfeld:
Fixes atom/vim-mode#602
Fixes atom/vim-mode#496

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -831,35 +831,29 @@ describe "TextEditor", ->
 
     describe ".moveToBeginningOfNextParagraph()", ->
       it "moves the cursor before the first line of the next paragraph", ->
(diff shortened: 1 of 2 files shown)


paulcbetts:
THANK YOU

PR #6242, opened 2015-04-03
Revert "Add asar support in Atom"

kevinsawicki:
Reverts atom/atom#5382 because of https://github.com/atom/symbols-view/issues/89


kevinsawicki:
@zcbenz I'd like to test the asar stuff a bit more on the different platforms before merging this again and also try and find a workaround for not losing the atom/apm symlinks on mac.

zcbenz:
> @zcbenz I'd like to test the asar stuff a bit more myself on the different platforms before merging this again

It sounds good to me.

> find a workaround for not losing the atom/apm symlinks on mac

The only workaround I can think of is to put symbols link under `Atom.app/Contents/Resources/app/atom.sh` and `Atom.app/Contents/Resources/app/apm/bin/apm`, which point to the new locations. Does it look good to you?

kevinsawicki:
> Does it look good to you?

Yeah, I was curious about that, if you have both an `app/` folder and an `app.asar` file, will atom shell still load from the asar file?

zcbenz:
> Yeah, I was curious about that, if you have both an app/ folder and an app.asar file, will atom shell still load from the asar file?

If there is no valid app under `app/`, atom-shell will still load from `app.asar`.

PR #6245, opened 2015-04-03
Added an option to disable keymaps for packages.

jussi-kalliokoski:
Introduces a config entry called packagesWithKeymapsDisabled that allows users to disable specific packages' keymaps.

atom-disable-keymaps-for-packages

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -357,6 +357,26 @@ describe "PackageManager", ->
             expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-n', target: element1[0])[0].command).toBe 'keymap-2'
             expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-y', target: element3[0])).toHaveLength 0
 
+      describe "when the package is listed in the config 'core.packagesWithKeymapsDisabled'", ->
+        it "doesn't load the keymaps", ->
+          atom.config.set("config.packagesWithKeymapsDisabled", ["package-with-keymaps"])
+
+          element1 = $$ -> @div class: 'test-1'
+          element2 = $$ -> @div class: 'test-2'
+          element3 = $$ -> @div class: 'test-3'
+
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element2[0])).toHaveLength 0
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element3[0])).toHaveLength 0
+
+          waitsForPromise ->
+            atom.packages.activatePackage("package-with-keymaps")
+
+          runs ->
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element2[0])).toHaveLength 0
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element3[0])).toHaveLength 0
+
       describe "when the keymap file is empty", ->
         it "does not throw an error on activation", ->
           waitsForPromise ->
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -26,6 +26,13 @@ module.exports =
         default: []
         items:
           type: 'string'
+      packagesWithKeymapsDisabled:
+        type: 'array'
+        default: []
+        title: "Disable Keymaps for Packages"
+        description: "Used to disable keymaps for individual packages."
+        items:
+          type: 'string'
       themes:
         type: 'array'
         default: ['one-dark-ui', 'one-dark-syntax']
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -222,7 +222,9 @@ class Package
     return
 
   loadKeymaps: ->
-    if @bundledPackage and packagesCache[@name]?
+    if _.contains(atom.config.get("core.packagesWithKeymapsDisabled") ? [], @name)
+      @keymaps = []
+    else if @bundledPackage and packagesCache[@name]?
       @keymaps = (["#{atom.packages.resourcePath}#{path.sep}#{keymapPath}", keymapObject] for keymapPath, keymapObject of packagesCache[@name].keymaps)
     else
       @keymaps = @getKeymapPaths().map (keymapPath) -> [keymapPath, CSON.readFileSync(keymapPath) ? {}]


jussi-kalliokoski:
Note: not sure if the test actually works, couldn't get the test suite to run. :cry:

d2s:
This would be useful feature to have. :+1:

spec/package-manager-spec.coffee
```diff
@@ -357,6 +357,26 @@ describe "PackageManager", ->
             expect(atom.keymaps.findKeyBindings(keystrokes:'ctrl-n', target:element1[0])[0].command).toBe 'keymap-2'
             expect(atom.keymaps.findKeyBindings(keystrokes:'ctrl-y', target:element3[0])).toHaveLength 0
 
+      describe "when the package is listed in the config 'core.packagesWithKeymapsDisabled'", ->
+        it "doesn't load the keymaps", ->
+          atom.config.set("config.packagesWithKeymaps", ["package-with-keymaps"]);
```
postcasio: Should be `atom.config.set("config.packagesWithKeymapsDisabled", ["package-with-keymaps"]);`

jussi-kalliokoski: Thanks, fixed! Also fixed lint errors. Looks like the tests pass now (only get unrelated test fails).

slang800:
Can't you already do this with your `keymap.cson` & the [`unset!`](https://github.com/atom/atom-keymap/issues/9) value?

xlaywan:
https://github.com/atom/atom/pull/6245#issuecomment-101761370, No if you need to remap hotkeys. I think this is useful feature.

slang800:
@xlaywan: that's exactly what the issue I linked to is talking about: https://github.com/atom/atom-keymap/issues/9

"hotkeys" are another term for keybindings, if that was the confusion...

lee-dohm:
@slang800 You can do this with `unset!`, it is true. But if a package has 50 key bindings, it is a pain to have to go in and copy the various keybindings (especially if they're spread across multiple files), paste them into your `keymap.cson` and set them all to `unset!` ... just to get back to zero.

jussi-kalliokoski:
It's also more maintenance in terms of when packages / APIs change, since after each update to the packages / the editor itself, you have to check that the packages didn't manage to slip in any nasty keymappings.

slang800:
@jussi-kalliokoski - that sounds like an issue with the packages themselves having poor defaults, rather than something in the atom core.

slang800:
@lee-dohm - I just use the keybindings page in settings to copy the ones I don't like & unset them... maybe we could just make that process easier by adding a "disabled" checkbox on the side of each keybinding, and clicking that would add the `unset!` declaration for that particular keybinding to your keymap.cson?

As for a package with 50 keybindings, that sounds like a really bad package. Does anyone actually do that?

xlaywan:
Setting `unset!` in `keymap.cson` is not the solution, because it diables **all** keybindings.

slang800:
@xlaywan: that doesn't match the way that unset is supposed to work... maybe report what you're seeing to https://github.com/atom/atom-keymap/issues?

xlaywan:
@slang800, I think unset works as expected(for example: `enter: 'unset!'` will disable all bindings from all packages and core for `enter` keystroke) or I didn't understand you?

slang800:
no, you'd use a selector... for example, to disable all the key-bindings from open-on-github, you'd use this:
```cson
'atom-workspace':
  'alt-g b': 'unset!'
  'alt-g c': 'unset!'
  'alt-g g': 'unset!'
  'alt-g h': 'unset!'
  'alt-g o': 'unset!'
  'alt-g r': 'unset!'
```

Because that's the same selector that you get from the keybindings page when you hit the copy button.

xlaywan:
But `'alt-g b': 'unset!'` completely disables `alt-g b` for all packages that used this keybinding.

slang800:
You can't have multiple packages using the same keybinding with the same selector, one would take precedence over the other. Other packages that use the `alt-g b` keybinding under a different selector are (or at least should be) unaffected.

xlaywan:
Ok, got it. But I still think this feature is useful for cases were I want to use only one command from package and bind it to some key and disable rest bindings. And if I install another package that uses some other bindings from first package,  I will  don't have to think about that keybindings which I forcibly disabled. :)

slang800:
@xlaywan Yeah, that makes sense. However, I'm worried about that kind of logic being added to the core, rather than being in an auxiliary package for people who want such things.

jussi-kalliokoski:
The problem is not as much that packages are misbehaving - batteries included is good - but that often the shortcuts are in conflict. I use the `vim-mode` package which has a nice boon of handy shortcuts, but other packages' shortcuts don't necessarily play well with them. Then there's common keybindings such as tab, that every package wants, so I've disabled the keymaps for the most of them and added them in my own keymap with the correct precedence (so that e.g. autocomplete doesn't happen when I want a snippet or emmet templating).

jussi-kalliokoski:
> Yeah, that makes sense. However, I'm worried about that kind of logic being added to the core, rather than being in an auxiliary package for people who want such things.

How would one do this? Preventing keymaps from loading would require tapping into the internals (`Package`), that currently aren't exposed, and surely it shouldn't be recommended for packages to depend on un-exported behavior **and** monkey patch that? However, if there's a good way to tap into the `Package` module, I'd be happy to make a package for this in the meantime as this issue is further discussed.

slang800:
@jussi-kalliokoski: What the package would do is add `unset!` declarations for each of the packages that the user wants to disable the key-bindings of. This way, there would be no monkey-patching or extra magic in the core. If you wanted to be really fancy, you could even add more granular control and let specific key-bindings be disabled with `unset!`.

d2s:
@jussi-kalliokoski One nice option could be to scan every package’s key-bindings and build visual list of them, allowing user to see overlapping bindings on one list (or even a generic keyboard layout, though that might be much more complex to build).

lee-dohm:
@d2s Actually, I think @mark-hahn already made a package to do keybinding visualizations like that?

slang800:
@lee-dohm https://github.com/atom/keybinding-resolver?

db47h:
Here's a good example where using unset! just doesn't cut it: the navigate package which maps F2 and F3:

```
'.platform-win32 atom-text-editor:not(.mini), .platform-linux atom-text-editor:not(.mini) ':
  'f2': 'navigate:forward'
```

F2 is also mapped to:

```
'atom-text-editor':
  'f2': 'bookmarks:jump-to-next-bookmark'
```

Note the different selector. I used the following unset!:

```
'.platform-win32 atom-text-editor:not(.mini), .platform-linux atom-text-editor:not(.mini) ':
  'f2': 'unset!'
```

but it won't give me back the default behavior for the F2 key. I have to map it again. Granted, using Atom core keybinds is a major flaw in the navigate package, but why should the end user pay the price for poor keybind choices in a package?

@slang800:

>  I just use the keybindings page in settings to copy the ones I don't like & unset them... maybe we could just make that process easier by adding a "disabled" checkbox on the side of each keybinding, and clicking that would add the unset! declaration for that particular keybinding to your keymap.cson?

A check box on the side of each keybinding would be just great, as long as it does not use the unset! mechanism, or we'd be back at square one.

solarsailer:
I recently started to use Atom — and in just a few hours, I have already encountered a situation where this modification would have made a lot of sense (eg. emmet and subword movements).

I don't use Emmet a lot, so I'm fine with calling its functions only via the command palette (and disabling all the package's keybindings — I can still add one or two keybindings after, if I want to).

This would make a lot of sense. :+1:

maxbrunsfeld:
This functionality exists on master now. It was added in #8130 and #8173. Sorry for neglecting this PR @jussi-kalliokoski; I became aware of #8130 before this one even though this one was opened earlier. Closing this out for now. Let me know if I should reopen it.

jussi-kalliokoski:
@maxbrunsfeld no problem, the other PR seems better anyway, I'm just happy to finally get this feature. ^.^ :tada:

PR #6246, opened 2015-04-03
:art: :green_heart: Fixed the test script.

jussi-kalliokoski:
Also removed duplicated logic to run grunt.

### DIFF
--- a/script/build
+++ b/script/build
@@ -1,13 +1,12 @@
 #!/usr/bin/env node
 var cp = require('./utils/child-process-wrapper.js');
+var runGrunt = require('./utils/run-grunt.js');
 var path = require('path');
 
 process.chdir(path.dirname(__dirname));
 
 cp.safeExec('node script/bootstrap', function() {
   // build/node_modules/.bin/grunt "$@"
-  var gruntPath = path.join('build', 'node_modules', '.bin', 'grunt') + (process.platform === 'win32' ? '.cmd' : '');
-  var args = ['--gruntfile', path.resolve('build', 'Gruntfile.coffee')];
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
Could you explain what bug you fixed in the script? Thanks!

jussi-kalliokoski:
Yes, so the test script didn't find grunt:

```
Node: v0.10.33
npm: v2.5.1
Installing build modules...
Installing apm...
Installing modules ✓
Deduping modules ✓
/bin/sh: node_modules/.bin/grunt: No such file or directory
```

(because `grunt` is located in `build/node_modules/.bin/grunt` instead) nor the gruntfile.

izuzak:
@maxbrunsfeld I was just looking at this I think this makes sense. The actual bug is in [this line](https://github.com/atom/atom/pull/6246/files#diff-0e2bc0f03fe54e5e0fff129205e6ee92L8) -- it's trying to run grunt from `node_modules/.bin/grunt`, instead of `build/node_modules/.bin/grunt`. I _think_ that should have been updated as a part of https://github.com/atom/atom/pull/1145.

The other changes are refactoring.

maxbrunsfeld:
This looks good to me. Thanks for the fix, @jussi-kalliokoski!

izuzak:
@kevinsawicki just wondering if you have any :thought_balloon: on this fix+refactoring before it's merged -- you've been maintaining these scripts and making sure things are working as smooth as possible. :blue_heart:

kevinsawicki:
Very cool, thanks for fixing this :ship:

jussi-kalliokoski:
:+1: :tada:

PR #6248, opened 2015-04-04
Public Continuous Integration

### LINKED ISSUE #6222: Setup CI on Travis and AppVeyor

as-cii:
Closes #6222  :green_heart: :green_heart: :green_heart:
- [x] Setup Travis
  - [x] OS X setup. [You can see a passing build here.](https://travis-ci.org/as-cii/atom/builds/57160743)
  - [x] `script/cibuild` didn't work out of the box on Travis; on my OS X machine it doesn't work either, [showing the same failures](https://travis-ci.org/as-cii/atom/builds/57005243). @kevinsawicki: any idea why we're getting those errors? `run-specs` task is not really reliable on my machine and I usually run tests through `apm test`: do you think we can keep it in Travis as well?
  - [x] Linux setup (to test builds only, as per current `cibuild` behavior).
    - [x] Ask Travis to enable multi-os setup, [as it is disabled by default](http://docs.travis-ci.com/user/multi-os/)

Any idea or suggestion is most welcome, as usual. Thank you! :pray: 

/cc: @atom/non-github-maintainers @atom/owners


kevinsawicki:
> any idea why we're getting those errors?

Those looks like integration spec errors, @maxbrunsfeld might know better

maxbrunsfeld:
> {"code":"ECONNREFUSED", "errno":"ECONNREFUSED", "syscall":"connect", "address":"127.0.0.1", "port":9515}

Yeah, it looks like `chromedriver` isn't running. Those tests try to run an executable at the path `atom-shell/chromedriver/chromedriver`. That executable is created by the `download-atom-shell-chromedriver` grunt task. Maybe we need to run that task as an additional `install` step.

maxbrunsfeld:
Oh, nevermind, I just realized that the build you linked to was already running `script/cibuild`, which runs the aforementioned grunt task. Thinking...

maxbrunsfeld:
@as-cii if you can reproduce these failures on your local machine, could you try focusing [the integration tests](https://github.com/atom/atom/blob/master/spec/integration/startup-spec.coffee#L13), bumping [this timeout](https://github.com/atom/atom/blob/master/spec/integration/helpers/start-atom.coffee#L117) to a larger number (a couple of seconds maybe) and re-running `ATOM_INTEGRATION_TESTS_ENABLED=true apm test`?

The tests currently have a very naive way of waiting for chromedriver to start up, just because it worked in our existing test and CI environments. If bumping the timeout fixes the problem on your machine, then I'll add a more complex polling-based strategy in place of this simple timeout, and hopefully that will fix the failures on Travis.

as-cii:
@maxbrunsfeld: thank you for your extensive investigation :metal:

The first thing I had to change on my machine was [`concurrency = 2` to `concurrency = 1`](https://github.com/as-cii/atom/commit/859a6dcf6867b4b930b7c96215987b1200cc62e8#diff-1dc361262257553d817b132c3c6b284fR7):  apparently this causes some race condition, which in turn arises issues such as `ECONNREFUSED` (may be related to [this issue on selenium](https://code.google.com/p/selenium/issues/detail?id=6989&noredir=1)), or, even worse, it hangs the build (both on my machine and [on Travis](https://travis-ci.org/as-cii/atom/builds/57460242)).

Bumping the timeout to `1000ms` as you suggested [seems to fix the issue](https://travis-ci.org/atom/atom/builds/57478065): on my machine it always works now, whereas on Travis [it randomly fails](https://travis-ci.org/as-cii/atom/builds/57478047); using `5000ms` allegedly works fine on Travis as well, thus confirming your hypothesis that our timeout is the culprit and that we may want to switch to a less flaky alternative. :sparkles: 

@maxbrunsfeld: I think we can handle this as the selenium guys do https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/webdriver/DriverService.cs#L173-205. Namely, they constantly issue a `GET` request to the service, waiting for it to return `200`. I don't know if you had something else in mind, but I feel like it could be a good starting point. What do you think?

maxbrunsfeld:
@as-cii that looks perfect :guitar:.

as-cii:
Everything works nicely on Travis and, apparently, I managed to make AppVeyor work as well (I say "apparently" because the [builds times out](https://ci.appveyor.com/project/as-cii/atom/build/1.0.34); read more below)

Installing packages and building Atom, however, takes a significant amount of time: on Travis this time is  ~18 minutes, whereas on AppVeyor it is more than 40 minutes, thus causing the build to timeout. We need to make that faster :zap: 

Running specs takes a relatively short amount of time, therefore I wouldn't start optimizing from there (e.g. by splitting tests across many workers). By contrast, installing modules seems to be pretty slow and one idea could be to start caching dependencies; if we decide to go with that, I think there are a couple of problems with it:

1. **AppVeyor**: in order to cache dependencies we need to reach the end of the build. This can be easily worked around by running only one build with a modified `appveyor.yml` (e.g. spawn `script/bootstrap` only and cache everything). This feels somehow flaky as a workaround, though, and we may want to contact their support to ask for an extended build timeout (which we should reach anyway only when cache gets completely invalidated).
2. **Travis CI**: caching is enabled for private repositories and on the new "container infrastructure" only. I see two solutions to this issue:
   1. Contact the support team and ask them if we could make use of their cache despite Atom being public;
   2. Build our own cache layer;

@maxbrunsfeld @kevinsawicki: what's your take on this? Any feedback or suggestion is very welcome :+1: 

Thank you! :pray:

mnquintana:
@as-cii Re: Travis, after a quick look-over of the cibuild script I didn't see anything that would require `sudo`. It looks like we should be able to move to the container infrastructure without much hassle, unless I'm missing something. (The caching is also [free now for public repositories with the container-based infrastructure.](http://docs.travis-ci.com/user/workers/container-based-infrastructure/))

as-cii:
@mnquintana: thank you for your reply :bow: The requirement, however, is to run specs across several platforms: where possible I'd love to make use of the existing caching infrastructure, but, as far as I know, this is simply not possible on their OS X public boxes by default.

Moreover, when we'll settle on a caching option (assuming we should cache at all), my idea would be to have as little differences as possible among our CI scripts: namely, I'd prefer not to have two different caching strategies for two different platforms. I agree, however, that we shouldn't try to reinvent the wheel if we have other alternatives at hand :+1:

as-cii:
@maxbrunsfeld @kevinsawicki: seems like we have our first properly configured build! https://travis-ci.org/atom/atom/builds/57928777 :tada: :tada: :tada: 

Unfortunately it's not green: I don't think the failures are related to the changes in this pull-request though, and it would be great if you could confirm you see those errors in your private cibuild as well every now and then.

Moreover, as stated in a previous comment, the build takes quite a while to complete and I have to say it's very slow compared to your private infrastructure (i.e. it takes more than 30 minutes on Travis). I feel like caching dependencies could be a quick win (e.g. it should get us under 10 minutes) but that would require a custom caching strategy, 'cause Travis does not provide it for public Mac projects.

The two issues illustrated above are my biggest concerns before :shipit: and I'd love to have your feedback on how to proceed, especially because we may need to setup a custom storage service (such as Amazon S3) to upload/download cached dependencies.

I am looking forward to collaborating with you on this :raised_hands: :bow:

maxbrunsfeld:
@as-cii what do you think about only running the tests on Linux, and just using the OSX infrastructure to check that we can build the app? That way, we could use caching for the build that we rely on the most: the one that runs the tests.

as-cii:
@maxbrunsfeld: I think we're not ready yet to use Linux as our main infrastructure for testing, as a bunch of package specs fail on Ubuntu (even on my local VM) and https://github.com/travis-ci/travis-ci/issues/3643 hasn't been "fixed" yet. The path to make everything bullet-proof on that platform is likely complex and could take a fair amount of time.

On the other hand we have a quite solid OS X environment where, even if they take ~30min., specs run smoothly (especially after we'll complete #6385).

So, here's my proposal :green_heart: 

* Ditch AppVeyor for the time being, as it's a lot slower than Travis and running specs on Windows seems to be even more cumbersome than on Linux;
* Keep triggering private CI builds when GitHub employees push a commit;
* Use always Travis CI (along with the private CI when needed)
  * Run a full `script/cibuild` on :apple: ;
  * Only test builds on :penguin: ;

This way we can provide a CI for our contributors already and check if it works fine for them. In the meantime we could optimize stuff to achieve a better speed (e.g. use a caching layer, use the `sudo: false` infrastructure, etc.).

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+
+os:
+  - linux
+  - osx
+
+install:
+  - git clone https://github.com/creationix/nvm.git /tmp/.nvm
+  - source /tmp/.nvm/nvm.sh
+  - nvm install $NODE_VERSION
+  - nvm use $NODE_VERSION
```
maxbrunsfeld: Just curious; why did you choose to install node.js like this, rather than w/ travis's node support?

```yaml
node_js:
  - "0.12"
  - "0.10"
```

Does builtin node support not work w/ multi-os builds?

as-cii: Exactly (see https://github.com/travis-ci/travis-ci/issues/2311 for further information). To make it work consistently I ended up using a configuration similar to nodegit's one (https://github.com/nodegit/nodegit/blob/master/.travis.yml).

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      sudo apt-get install build-essential git libgnome-keyring-dev fakeroot;
+      export DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
```
as-cii: If we decide not to run specs on Linux, I think these last two lines may be commented out.

spec/spec-helper.coffee
```diff
@@ -51,7 +51,7 @@ Object.defineProperty document, 'title',
 
 jasmine.getEnv().addEqualityTester(_.isEqual) # Use underscore's definition of equality for toEqual assertions
 
-if process.env.JANKY_SHA1 and process.platform is 'win32'
+if (process.env.JANKY_SHA1 and process.platform is 'win32') or process.env.CI
```
kevinsawicki: The windows/janky_sha1 check can actually just be completely removed here since the specs no longer run on Windows on janky.

maxbrunsfeld:
> So, here's my proposal :green_heart:

I agree on all fronts. This change already adds a ton of value as it is.

>  In the meantime we could optimize stuff to achieve a better speed (e.g. use a caching layer, use the sudo: false infrastructure, etc.

I would say we could leave it as is for a while. I'd be reluctant to add the complexity of a custom caching layer. And as far as I know `sudo: false` only works with Linux, because it opts-in to the docker infrastructure. I could be wrong on both these fronts though; let me know what ideas you have.

Other than that, :ship:.

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+    fi
+
+script: script/cibuild
+
+notifications:
+  email:
+    on_success: never
+    on_failure: change
```
kevinsawicki: A bunch of the other `.travis.yml` files for packages use the following:

```yml
git:
  depth: 10
```

Think it is worth adding here?

maxbrunsfeld: :+1:

Why not 1 though, rather than 10?

kevinsawicki: https://github.com/atom/ci/pull/4#discussion_r16486942

as-cii:
As you can see our latest build has successfully run our `script/cibuild` :tada: :tada: :tada:

I agree with all the changes you proposed and I already applied them: please, have a last look so that we can :shipit:

@kevinsawicki @maxbrunsfeld: what do you think if we run a build on Janky just to make sure we didn't break anything there and that it is able to co-exist with Travis gracefully :rainbow:?

Thanks to all of your for your extensive review and for the great support :sparkles:

.travis.yml
as-cii: Agreed :+1: 

Probably not a significant speedup (since Travis already uses a depth of `50`), but it's nice to be consistent with other repos :+1:

kevinsawicki:
> what do you think if we run a build on Janky just to make sure we didn't break anything there and that it is able to co-exist gracefully with Travis :rainbow:?

Let's just ship it :ship:

Thanks so much for setting this up :green_heart:

maxbrunsfeld:
This is great. :balloon:

thomasjo:
> This is great.

Agreed! Awesome job as usual @as-cii. King amongst men! :bow:

as-cii:
A huge thanks goes to Travis CI for making Multi-OS possible: you rock! :metal: (/cc: @rkh)

@kevinsawicki @maxbrunsfeld: I have spotted a bunch of errors during some of our OS X builds on Travis: I suspect this could be due to running package and core specs in parallel. I plan to give some :heart: to `spec-task.coffee`: this will include some refactoring :art: and a speedup :zap: for our build times (more on that this weekend, hopefully).

Thanks everyone! :bow:

bhuga:
:sweat_smile: :metal:

PR #6323, opened 2015-04-09
Update help message to reflect changes to multi-path handling

maxbrunsfeld:
/cc @atom/core - could you double check my wording of this message?


jlord:
Makes sense to me :pizza:

PR #6333, opened 2015-04-10
Precompile bootstrap

kevinsawicki:
Precompile the `static/bootstrap.less` file and ship a `static/bootstrap.css` instead.

This shaves ~10ms off of startup since the less import cache checks aren't run and the file system is only hit once for the `.css` file instead of 10 times for all the bootstrap imports.

### DIFF
--- a/build/tasks/prebuild-less-task.coffee
+++ b/build/tasks/prebuild-less-task.coffee
@@ -75,7 +75,7 @@ module.exports = (grunt) ->
         themeMains.push(mainPath) if grunt.file.isFile(mainPath)
         importPaths.unshift(stylesheetsDir) if grunt.file.isDir(stylesheetsDir)
 
-      grunt.verbose.writeln("Building LESS cache for #{configuration.join(', ').yellow}")
+      grunt.verbose.writeln("Building Less cache for #{configuration.join(', ').yellow}")
       lessCache = new LessCache
         cacheDir: directory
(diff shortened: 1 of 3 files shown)


kevinsawicki:
@simurai and @benogle it should be fine to just ship bootstrap.css right? It didn't look like tweaking the bootstrap variables via a theme was currently supported.

build/tasks/prebuild-less-task.coffee
```diff
@@ -1,10 +1,28 @@
...
+    lessCache = new LessCache
+      cacheDir: temp.mkdirSync('atom-less-cache')
+      resourcePath: path.resolve('.')
+
+    bootstrapCss = lessCache.readFileSync(bootstrapLessPath)
+    grunt.file.write(bootstrapCssPath, bootstrapCss)
+    rm(bootstrapLessPath)
+    rm(path.join(appDir, 'node_modules', 'bootstrap', 'less'))
+
   grunt.registerMultiTask 'prebuild-less', 'Prebuild cached of compiled LESS files', ->
```
50Wliu: If LESS was changed down at line 78, it should also be changed here right?

kevinsawicki: Yup, thanks!

simurai:
Yes, I'm not aware of a way to change Bootstrap's variables from themes. So precompiling is a good idea.

PR #6338, opened 2015-04-11
Fix doc on getting the home directory

subesokun:
`atom.fs` is undefined in the current Atom version. The function `getHomeDirectory()` seems to be now only available in the Atom `fs-plus` package.


izuzak:
I believe you're right -- thanks for updating the docs :+1:

subesokun:
Sure, you're welcome :) Thanks for merging the PR.

PR #6346, opened 2015-04-12
Changes ATOM_HOME on Windows to be APPDATA.

### LINKED ISSUE #6345: ATOM_HOME on Windows should follow Windows conventions, not Linux conventions.
In the Linux world, the convention is to put user-specific application configuration in `~/.folder` or `~/.file`.  In the Windows world, this is not the case.  User-specific settings have a well defined place they should go which is `~/AppData/(Local|Roaming|LocalLow)`.

Details on exactly which of the three folders to use can be found here: http://windows.microsoft.com/en-us/windows-8/what-appdata-folder

Additional information about how to access the different paths can be found here (a bit old, but still valuable):
http://blogs.msdn.com/b/patricka/archive/2010/03/18/where-should-i-store-my-data-and-configuration-files-if-i-target-multiple-os-versions.aspx

Reasoning: `~/.(file|folder)` is not hidden on Windows and it is not the place an experienced Windows developer would look for user settings.  It also pollutes the user directory which is meant to be the place for user documents and files that they interact with directly (word document, photos, videos, downloads, etc.).

MicahZoltu:
Fixes https://github.com/atom/atom/issues/6345.

This doesn't address the migration issue where a user may be upgrading from a previous version of ATOM and have files currently stored in `~/.atom`.  How this is resolved is a much bigger design decision than I am qualified to make so I will leave it up to the Atom maintainers.

My 2 cents is that migration occurs on first-run after an upgrade.  It is unclear to me if Atom has mechanisms in place to make this easy or not.  Alternatively, every run of atom could check to see if there are files in the old location and if so, migrate them.

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -76,13 +76,13 @@ var setupCoffeeCache = function(cacheDir) {
 
 var setupAtomHome = function() {
   if (!process.env.ATOM_HOME) {
-    var home;
+    var atomHome;
     if (process.platform === 'win32') {
-      home = process.env.USERPROFILE;
-    } else {
-      home = process.env.HOME;
+      atomHome = path.join(process.env.APPDATA, 'atom');
     }
-    var atomHome = path.join(home, '.atom');
+    else {
+      atomHome = path.join(process.env.HOME, '.atom');
+    }
     try {
       atomHome = fs.realpathSync(atomHome);
     } catch (error) {


Zireael07:
+1

paulcbetts:
I'm pretty :-1: on this change:

* All of the documentation / 3rd party blogs point to `~\.atom`, this will lead to a ton of "I don't have a ~\.atom folder!" questions because they followed some guide

* `%AppData%\Atom` already has lots of stuff in it because it's the folder where Atom Shell puts its data, whereas we expect users to directly edit files in this folder like `styles.less`

In exchange for these pretty significant drawbacks (especially that first one), we don't really get anything in return, other than some "cleanliness". If Atom weren't already released software it'd be a different story, but this really will cause problems for users if we try to move the folder.

Zoltu:
TL;DR: *When in Rome, do as the Romans do.*
`s/Rome/Windows/`
`s/Romans/Windows Users/`

One could pretty easily argue that for windows users, `~\.atom` does not point to `%USERPROFILE%\.atom`.  Windows does not have a `~` directory so the documentation is already pretty poor for Windows users.  I believe this significantly weakens point 1 since its foundation is built on the premise that "existing docs will be worse" when in reality, I believe the "existing docs will be equally bad".  As a Windows user, when I see a reference to `~` I know that I likely have to check in 6 locations to figure out what that translates to in Windows for this particular package port.  In the *nix world `~` is a dumping ground for all things.  In Windows, there are well defined places for each of these things (depending on the type of data) which include `%USERPROFILE%`, `%APPDATA%`, `%LOCALAPPDATA%`, `%ALLUSERSPROFILE%`, `%PROGRAMFILES86%` and `%PROGRAMFILES%`.

For point 2, Atom shell is doing the right thing (sort of).  If Atom shell files weren't meant to sit in the same place as Atom files then Atom shell should have created an atom-shell folder.  More specifically, there is a reason that Microsoft encourages `%APPDATA%\vendor\product` naming, it is to avoid collisions like this.  In this case, I believe atom-shell to be in the wrong here but that is a discussion for another place.  In either case, atom and atom-shell are both authored by the same vendor, so presumably this conflict can be dealt with in-house (though you still have to worry about another vendor creating something called atom and clobbering your application).

For point 3, I would say that this is the most compelling argument, but it still doesn't align with the Windows world and very much is a nix-ism.  The idea that a *user* knows where all of the files generated by an application go is something that Windows applications shy away from.  This sort of tribal knowledge hurts self-discovery which is a big part of the Windows user experience that people have come to expect.  In the case of files the user is intended to edit, several of them are already available via the menus and it would not be hard to add a menu item or command for something like "Open Atom settings folder in explorer" or even "Open Atom settings folder in Atom".  Both of these are far more discoverable than a folder on disk that is intended to be hidden (`.` prefixed) but is not hidden and auto-created in an unexpected location.

paulcbetts:
> Windows does not have a ~ directory so the documentation is already pretty poor for Windows users.

You've misinterpreted my meaning when I wrote `~` for shorthand - my issue is that documentation / blogging / Stack Overflow will direct people to the old folder, which won't exist / won't be read, however you notate it.

50Wliu:
The main problem is Atom itself uses `~./atom` when referring to it.  I had a hard time finding it since, like @Zoltu said, that doesn't really exist on Windows (but thankfully I found the nice button in the Settings View that opened it up for me :smile:).

Atom changing its docs to point to either `%userprofile%/.atom` or `%appdata%/Atom` would be better.

batjko:
I think there's still the idea on a lot of people's wishlists, to have it install to `Program Files` or some other more direct location.

I would be hesitant to change this path now, when a more dramatic change is likely to come in the future.

Zireael07:
I believe the documentation should be updated as @50Wliu says.
And I still support the migration - fishing out where Atom installed to was really difficult.

thomasjo:
I'm :+1: for updating the documentation with regards to Windows (e.g. `%USERPROFILE` should replace `~`), but I'm :-1: on changing the default location. Hiding it away in `%APPDATA%` is not user-friendly. While the current location might not be optimal either, at least it's more visible and accessible to the average user.
Alternative locations might be along the lines of Visual Studio, MATLAB, etc, which use `%USERPROFILE/Documents/<app name>`.

---
By the way `~` is not a special folder in UNIX-like systems, it's a shell expansion. The file system and OS has no concept of `~`, it's merely expanded by the shell and replaced with the value of `$HOME`.

paulcbetts:
Definitely :+1: for updating docs to reflect the right location

paulcbetts:
>  fishing out where Atom installed to was really difficult.

This is a different proposition than what the PR contributor is suggesting - you're talking about the installation directory `C:\Users\NAME\AppData\Local\Atom` whereas this PR wants to move the user's configuration files `C:\Users\NAME\.atom`

aharpervc:
re @thomasjo, dumping config in `%userprofile%` is at least as unfriendly as putting it in the correct location in `%appdata%`. Not only is it "wrong" per correct use of these folders, it's hostile because you're dumping things in my user profile folder.

Zireael07:
Oops. For said configuration, haven't yet figured out where they live, so please move them somewhere that's findable. :)

jhasse:
I'm also for keeping the current location.

> Alternative locations might be along the lines of Visual Studio, MATLAB, etc, which use `%USERPROFILE/Documents/<app name>`.

Please don't :O

lee-dohm:
You bring up a good point that we aren't being good Windows citizens by flaunting convention. On the other hand, so as to not throw current users for a loop, we are going to leave the directories where we store configuration information where they are for now.

Thanks for contributing and helping to shape the conversation!

Issue #6274, opened 2015-04-06
TextEditorComponent specs are flaky


kevinsawicki:
Saw three different failures in three different builds on CI with some that passed.

```
>> TextEditorComponent
>>   gutter rendering
>>     it applies the background color of the gutter or the editor to the line numbers to improve GPU performance
>>       Expected 'rgba(0, 0, 0, 0)' to be 'rgb(255, 0, 0)'. (spec/text-editor-component-spec.coffee:579:53)
```

```
>> TextEditorComponent
>>   gutter rendering
>>     fold decorations
>>       mouse interactions with fold indicators
>>         it does not fold when the line number componentNode is clicked
>>           Expected Function to be Function. (spec/text-editor-component-spec.coffee:690:38)
```

```
>> TextEditorComponent
>>   hiding and showing the editor
>>     when the editor is hidden when it is mounted
>>       it defers measurement and rendering until the editor becomes visible
>>         Expected 0 to be greater than 0.
>>           at [object Object].<anonymous> (/private/var/lib/jenkins/workspace/atom/spec/text-editor-component-spec.coffee:2400:64)
```

```
>> TextEditorComponent
>>   mousewheel events
>>     updating scrollTop and scrollLeft
>>       it updates the scrollLeft or scrollTop according to the scroll sensitivity
>>         Error: No animation frame requested
>>           at noAnimationFrame (/private/var/lib/jenkins/workspace/atom/spec/text-editor-component-spec.coffee:22:39)
>>           at [object Object].<anonymous> (/private/var/lib/jenkins/workspace/atom/spec/text-editor-component-spec.coffee:2088:9)
>>       it parses negative scrollSensitivity values at the minimum
>>         Error: No animation frame requested
>>           at noAnimationFrame (/private/var/lib/jenkins/workspace/atom/spec/text-editor-component-spec.coffee:22:39)
>>           at [object Object].<anonymous> (/private/var/lib/jenkins/workspace/atom/spec/text-editor-component-spec.coffee:2105:9)
```
```
>> TextEditorComponent
>>   scrolling
>>     it updates the horizontal scrollbar and the x transform of the lines based on the scrollLeft of the model
>>       Error: No animation frame requested
>>         at noAnimationFrame (/Users/travis/build/atom/atom/spec/text-editor-component-spec.coffee:22:39)
>>         at [object Object].<anonymous> (/Users/travis/build/atom/atom/spec/text-editor-component-spec.coffee:1927:7)
```

```
[...]

PR #6359, opened 2015-04-13
Fix some of the specs which fail on Linux

as-cii:
While addressing #6248 I discovered some flaky tests: some of them failed only on Linux, some others failed just because they were executed in a slow machine. In both cases, there wasn't a good reason why they should have failed and, therefore, I decided to fix them :green_heart: 

Please note that, although my ultimate goal is to make every spec bullet-proof, the changes here proposed are deliberately not comprehensive: there are a bunch of other tests which need to be inspected (see #6274) but I think it's better to address each of them incrementally, without creating gargantuan pull requests.

Also, I think this highlights the importance of running specs on every platform: even though all the tests hereby addressed were failing because of a bug in the test itself (and not in our production code), there's a quite high risk of introducing bugs on other platforms without noticing it.

An additional pair of :eyes: is most welcome, especially for some changes (in which I am going to write an additional comment).

Thanks :bow: 

/cc: @maxbrunsfeld @nathansobo @kevinsawicki

### DIFF
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -24,7 +24,7 @@ TextEditorElement = require '../src/text-editor-element'
 TokenizedBuffer = require '../src/tokenized-buffer'
 TextEditorComponent = require '../src/text-editor-component'
 pathwatcher = require 'pathwatcher'
-clipboard = require "../src/native-clipboard"
+clipboard = require "../src/safe-clipboard"
 
 atom.themes.loadBaseStylesheets()
 atom.themes.requireStylesheet '../static/jasmine'
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -24,7 +24,7 @@ TextEditorElement = require '../src/text-editor-element'
 TokenizedBuffer = require '../src/tokenized-buffer'
 TextEditorComponent = require '../src/text-editor-component'
 pathwatcher = require 'pathwatcher'
-clipboard = require 'clipboard'
+clipboard = require "../src/native-clipboard"
 
 atom.themes.loadBaseStylesheets()
 atom.themes.requireStylesheet '../static/jasmine'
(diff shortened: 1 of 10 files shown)


src/native-clipboard.coffee
```diff
@@ -0,0 +1,6 @@
+# Using clipboard in renderer process is not safe on Linux.
```
as-cii: What do you think about the name used here? To be honest, I don't like it much as it doesn't really express the intent of having a separate file. Maybe `safe-clipboard` may work better?

spec/spec-helper.coffee
```diff
@@ -24,7 +24,7 @@ TextEditorElement = require '../src/text-editor-element'
 TokenizedBuffer = require '../src/tokenized-buffer'
 TextEditorComponent = require '../src/text-editor-component'
 pathwatcher = require 'pathwatcher'
-clipboard = require 'clipboard'
+clipboard = require "../src/safe-clipboard"
```
kevinsawicki: Super minor, but typically single quotes are used for requires since they almost never have interpolated values in them

spec/text-editor-component-spec.coffee
```diff
@@ -1352,17 +1352,18 @@ describe "TextEditorComponent", ->
         expect(overlay.style.left).toBe position.left + gutterWidth + 'px'
         expect(overlay.style.top).toBe position.top + editor.getLineHeightInPixels() + 'px'
 
-        editor.insertText('a')
-        nextAnimationFrame()
+        if process.platform isnt "linux" # see TextEditorComponent#measureWindowSize
```
kevinsawicki: Are the asserts before this line still worth running on linux?

If not, you can mark this spec as not running on linux by changing the `it` to be:

```coffee
it "slides horizontally left when near the right edge on #win32 and #darwin", ->
```

as-cii:
I just updated the code with the suggested changes: thank you guys for your help! :pray: 

Also, please feel free to jump on this if you want to refine it further before :ship:ping :art: 

Thanks! :bow:

maxbrunsfeld:
Looks good to me! Thanks a lot for fixing this stuff.

kevinsawicki:
Thanks!

PR #6050, opened 2015-03-20
:bug: fix rowRangeForParagraphAtBufferRow using \S

lisongx:
Fixes https://github.com/atom/atom/issues/5963 , seems there shoud be using `\S`. 
But I'm not be able to run the test locally, please review

PR #6364, opened 2015-04-13
Fix rowRangeForParagraphAtBufferRow using \S -- Round two

### LINKED ISSUE #5963: Paragraphs selection misses nonempty lines
rowRangeForParagraphAtBufferRow misses nonempty lines

From the comment "Right now, a paragraph is a block of text bounded by and empty line or a block of text that is not the same type (comments next to source code)."

But the code will not match lines which contain non-word characters such as punctuation.

https://github.com/atom/atom/blob/f7bb8aab9e33090cc8455cfdf045efe6257568d9/src/app/language-mode.coffee#L183

This means that blocks of code are incorrectly selected if there is for example a trailing bracket on its own line.

benogle:
This is a version of #6050 that handles all cases and passes the specs. cc @seansay

Closes #6050 
Closes #5963

### DIFF
--- a/spec/language-mode-spec.coffee
+++ b/spec/language-mode-spec.coffee
@@ -124,6 +124,11 @@ describe "LanguageMode", ->
               // lines
               var sort = function(items) {};
               // comment line after fn
+
+              var nosort = function(items) {
+                  return item;
+              }
+
             };
           '''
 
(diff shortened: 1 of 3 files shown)


seansay:
:clap:

PR #6368, opened 2015-04-14
Consolidate selector caches

kevinsawicki:
This pull request consolidates the selector validation and specificity caching into the `clear-cut` library and upgrades to the faster specificity calculations via https://github.com/atom/clear-cut/pull/2.

Previously selector specificity was calculate and cached multiple places meaning cached values were not shared between keymaps, command registry, and context menus.

Now all places that parse selectors and calculate specificity share one common cache.

:racehorse:


as-cii:
:sparkles: :sparkles: :sparkles: 

There was an error in the specs but it should be already fixed with https://github.com/atom/clear-cut/issues/3 (more details there).

Great job here!

kevinsawicki:
Thanks for your help on this @as-cii :ship:

as-cii:
Thank **you** for bringing this speed improvement :zap:

PR #6379, opened 2015-04-14
Portable: Uses "userdata" folder if exists inside the application path

furious:
I like portable apps, but I don't wanna create an env. variable in every computer I plug my USB drive. So I did this to solve my (or everyone's?) problem.

### DIFF
--- a/src/browser/main.coffee
+++ b/src/browser/main.coffee
@@ -69,7 +69,9 @@ setupCrashReporter = ->
 setupAtomHome = ->
   return if process.env.ATOM_HOME
 
-  atomHome = path.join(app.getHomeDir(), '.atom')
+  atomHome = path.join(path.dirname(process.execPath), '.atom')
+  if not fs.existsSync(atomHome)
+    atomHome = path.join(app.getHomeDir(), '.atom')
   try
     atomHome = fs.realpathSync(atomHome)
   process.env.ATOM_HOME = atomHome
--- a/src/browser/main.coffee
+++ b/src/browser/main.coffee
@@ -69,7 +69,7 @@ setupCrashReporter = ->
 setupAtomHome = ->
   return if process.env.ATOM_HOME
 
-  atomHome = path.join(path.dirname(process.execPath), '.atom')
+  atomHome = path.join(path.dirname(process.execPath), 'userdata')
   if not fs.existsSync(atomHome)
     atomHome = path.join(app.getHomeDir(), '.atom')
   try


mark-hahn:
This seems a little dangerous.  I'm afraid I'd forget some .atom folder and
get the wrong one.  I copy them sometimes.

Maybe there should be some file also that says something like
'.atom-portable'?

furious-:
I just changed the portable folder name to "userdata", so that will avoid users complaining about copying/creating the ".atom" folder.

That way, on a fresh install, you just need to create the "userdata" on the atom executable path and it's ready to go, no need to set up ATOM_HOME env. variable.

src/browser/main.coffee
```diff
@@ -69,7 +69,9 @@ setupCrashReporter = ->
 setupAtomHome = ->
   return if process.env.ATOM_HOME
 
-  atomHome = path.join(app.getHomeDir(), '.atom')
+  atomHome = path.join(path.dirname(process.execPath), 'userdata')
```
kevinsawicki: `process.execPath` is something like `/Applications/Atom.app/Contents/Frameworks/Atom Helper.app/Contents/MacOS/Atom Helper` on Mac. This folder will be completely replaced whenever Atom automatically updates which means the `userdata` folder, if there, would be deleted.

This needs to be supported in a way where the config folder is not in a location that is deleted by automatic updates.

furious-: Well, I don't own a Mac to know this, but we're talking about a simple portable way for Atom, how it will be made using a Mac?

Clarifying, this only will work if you put an "userdata" folder (empty or a copy of ".atom" contents) inside the executable directory. In my case, I'm using a USB drive to use Atom on any computer I plug it.

kevinsawicki: So on Windows when you use the installer, the `proces.execPath` will be something like `C:\Users\kevin\AppData\Local\atom\app-0.191.0\atom.exe` so if you put a `userdata` folder in `C:\Users\kevin\AppData\Local\atom\app-0.191.0\` it will also be deleted the next time Atom automatically updates.

If this is something that will be supported by the app, the location needs to be someplace that isn't deleted when Atom updates so the person doesn't lose their packages, config, etc.

furious-: I understand... Actually I haven't installed the Atom, I just extracted the setup/nupkg to a directory and worked on this solution for portability, it doesn't have the "Update.exe", without this Atom will not auto-update, right?

But let's think that users would copy `C:\Users\kevin\AppData\Local\atom` folder with the updater. We can change the path to `'..\userdata'`, so it will stay with the updater and will not accidentally erased?

kevinsawicki: >  it doesn't have the "Update.exe", without this Atom will not auto-update, right?

Yes, it won't auto update.

> We can change the path to '..\userdata', so it will stay with the updater and will not accidentally erased?

I'm not sure exactly what you mean here, sorry, are you talking about looking for the `userdata` at `C:\Users\kevin\AppData\Local\atom\userdata`?

furious-: Yes! The resolved path from `process.execPath` + `..\userdata`

```coffeescript
atomHome = path.join(path.dirname(process.execPath), '..', 'userdata')
```

lee-dohm:
To my understanding, this PR was superseded by #8442. I'm going to close this PR. Please let me know if I've misunderstood anything. Thanks for contributing and helping shape the discussion!

PR #6380, opened 2015-04-14
Reopen windows when starting app w/ no arguments

### LINKED ISSUE #1603: reopen windows when relaunching Atom
A feature I love about ST2 and iA Writer, which I'm missing in Atom, is when you quit the app with open windows (including unsaved documents), they all appear again when you reopen. 

This has saved me countless times in my workflow to handle having to do a system restart for software update, etc and not have to take time to save all my unsaved docs, etc. 

It's also nice because I generally have five or six open projects that I just keep minimized in the editor for ease of access. It would be great to have such a feature in Atom.

maxbrunsfeld:
- [x] Write integration test
- [x] Make it pass
- [x] Fix duplicated logic for working w/ `~/.atom/storage`
- [x] Debug integration test failures

Closes #1603.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -3,6 +3,7 @@ ApplicationMenu = require './application-menu'
 AtomProtocolHandler = require './atom-protocol-handler'
 AutoUpdateManager = require './auto-update-manager'
 BrowserWindow = require 'browser-window'
+StorageFolder = require '../storage-folder'
 Menu = require 'menu'
 app = require 'app'
 fs = require 'fs-plus'
@@ -77,6 +78,7 @@ class AtomApplication
     @listenForArgumentsFromNewProcess()
     @setupJavaScriptArguments()
     @handleEvents()
+    @storageFolder = new StorageFolder(process.env.ATOM_HOME)
 
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test
       @openWithOptions(options)
@@ -426,29 +428,18 @@ class AtomApplication
             'safeMode'
             'apiPreviewMode'
           ))
-    fs.writeFileSync(@getStatePath(), JSON.stringify(states))
+    @storageFolder.store('application.json', states)
 
   loadState: ->
-    try
-      stateString = fs.readFileSync(@getStatePath(), 'utf8')
-    catch error
-      if error.code is 'ENOENT'
-        return false
-      else
-        throw error
-
-    states = JSON.parse(fs.readFileSync(@getStatePath()))
-    for state in states
-      @openWithOptions({
-        pathsToOpen: state.initialPaths
-        urlsToOpen: []
-        devMode: state.devMode
-        safeMode: state.safeMode
-        apiPreviewMode: state.apiPreviewMode
-      })
-
-  getStatePath: ->
-    path.join(process.env.ATOM_HOME, "storage", "application.json")
+    if states = @storageFolder.load('application.json')
+      for state in states
+        @openWithOptions({
+          pathsToOpen: state.initialPaths
+          urlsToOpen: []
+          devMode: state.devMode
+          safeMode: state.safeMode
+          apiPreviewMode: state.apiPreviewMode
+        })
 
   # Open an atom:// url.
   #
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -431,7 +431,7 @@ class AtomApplication
     @storageFolder.store('application.json', states)
 
   loadState: ->
-    if states = @storageFolder.load('application.json')
+    if (states = @storageFolder.load('application.json'))?.length > 0
       for state in states
         @openWithOptions({
           pathsToOpen: state.initialPaths
@@ -440,6 +440,9 @@ class AtomApplication
           safeMode: state.safeMode
           apiPreviewMode: state.apiPreviewMode
         })
+      true
+    else
+      false
 
   # Open an atom:// url.
   #
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -328,7 +328,7 @@ class AtomApplication
   focusedWindow: ->
     _.find @windows, (atomWindow) -> atomWindow.isFocused()
 
-  # Public: Opens multiple paths, in existing windows if possible.
+  # Public: Opens a single path, in an existing window if possible.
   #
   # options -
   #   :pathToOpen - The file path to open
@@ -341,7 +341,7 @@ class AtomApplication
   openPath: ({pathToOpen, pidToKillWhenClosed, newWindow, devMode, safeMode, apiPreviewMode, window}) ->
     @openPaths({pathsToOpen: [pathToOpen], pidToKillWhenClosed, newWindow, devMode, safeMode, apiPreviewMode, window})
 
-  # Public: Opens a single path, in an existing window if possible.
+  # Public: Opens multiple paths, in existing windows if possible.
   #
   # options -
   #   :pathsToOpen - The array of file paths to open
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -419,13 +419,13 @@ class AtomApplication
     states = []
     for window in @windows
       if loadSettings = window.getLoadSettings()
-        states.push(_.pick(loadSettings,
-          'initialPaths'
-          'isSpec'
-          'devMode'
-          'safeMode'
-          'apiPreviewMode'
-        ))
+        unless loadSettings.isSpec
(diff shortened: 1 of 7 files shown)


src/browser/atom-application.coffee
```diff
@@ -411,6 +415,41 @@ class AtomApplication
...
+  loadState: ->
+    try
+      stateString = fs.readFileSync(@getStatePath(), 'utf8')
+    catch error
+      if error.code is 'ENOENT'
+        return false
+      else
+        throw error
+
+    states = JSON.parse(fs.readFileSync(@getStatePath()))
```
kevinsawicki: Should this use `stateString` here?

src/browser/atom-application.coffee
```diff
@@ -411,6 +415,41 @@ class AtomApplication
         console.log("Killing process #{pid} failed: #{error.code ? error.message}")
     delete @pidsToOpenWindows[pid]
 
+  saveState: ->
+    states = []
+    for window in @windows
+      if loadSettings = window.getLoadSettings()
+        states.push(_.pick(loadSettings,
+          'initialPaths'
+          'isSpec'
```
kevinsawicki: So does this enable spec windows to be reopened at launch?

src/browser/atom-application.coffee
maxbrunsfeld: Yeah, thanks. I think I'm going to :fire: this and extract a coupe of helper functions for working with the `storage` directory here and in `atom.coffee`.

src/browser/atom-application.coffee
maxbrunsfeld: I was planning on doing that, but I think I should actually just filter out the spec windows for now. What do you think of that?

maxbrunsfeld: It'd be unappealing to fire up Atom, only to have your laptop fan start to immediately spin because specs are running.

kevinsawicki: > What do you think of that?

Sounds good, I think no spec windows is the right move.

maxbrunsfeld:
Ok, specs are good; the TextEditorComponent spec fixes just aren't on this branch.

src/storage-folder.coffee
```diff
@@ -0,0 +1,27 @@
...
+class StorageFolder
+  constructor: (containingPath) ->
+    @path = path.join(containingPath, "storage")
+
+  store: (name, object) ->
+    fs.writeFileSync(@pathForKey(name), JSON.stringify(object), 'utf8')
+
+  load: (name) ->
+    try
+      stateString = fs.readFileSync(@pathForKey(name), 'utf8')
```
kevinsawicki: You don't need the `utf8` here, `JSON.parse` handles `Buffer` objects.

src/storage-folder.coffee
```diff
@@ -0,0 +1,27 @@
+path = require "path"
+fs = require "fs-plus"
+
+module.exports =
+class StorageFolder
+  constructor: (containingPath) ->
+    @path = path.join(containingPath, "storage")
+
+  store: (name, object) ->
+    fs.writeFileSync(@pathForKey(name), JSON.stringify(object), 'utf8')
```
kevinsawicki: Should a `try`/`catch` be around this? Or are errors handled above this layer?

src/storage-folder.coffee
kevinsawicki: Should it be deleted after it is read? So if Atom is crashing because of the window state it doesn't keep retrying on subsequent launches.

src/storage-folder.coffee
maxbrunsfeld: I extracted this from `Atom::saveSync()`, which didn't wrap it in a `try/catch`. What would the `try/catch` be for in this case?

kevinsawicki: > I extracted this from Atom::saveSync()

Oh okay.

> What would the try/catch be for in this case?

I would think if `fs.writeFileSync` fails then atom wouldn't quit possibly since an error is thrown from the `beforeunload` handler that this runs in.

And since `store` is also called in the browser process, I wasn't sure how an error thrown during `before-quit` handler would behave.

maxbrunsfeld: Yeah, interesting. Looking...

benogle:
SHIIIIIT. This will make soo many people happy

src/storage-folder.coffee
kevinsawicki: I think the `Atom.saveSync` code was very very old, and as we've learned since then, the fs is far from reliable writing stuff, https://github.com/atom/atom/blob/c665747879161991f413db2a967527c8abdae1c0/src/pane.coffee#L666-L682

maxbrunsfeld: I just tried closing windows/quitting the app after running `sudo chown -R ~/.atom/storage`. I see `Uncaught Error: EACCES: permission denied` in the console, but otherwise everything works as expected.

So I don't have a strong opinion about it, but generally I err on the side of not swallowing errors. Does this case warrant it?

basarat:
image

src/storage-folder.coffee
kevinsawicki: > So I don't have a strong opinion about it, but generally I err on the side of not swallowing errors. Does this case warrant it?

To me this feels like the same case to wrap the `JSON.parse` in a try/catch down on line 22, better safe than sorry.

benogle:
One piece of feedback: it doesnt keep the active window. So on restart the focused window may not be the window focused on quit. Other than that, it's great :tada:

src/storage-folder.coffee
```diff
@@ -0,0 +1,27 @@
...
+
+  load: (name) ->
+    try
+      stateString = fs.readFileSync(@pathForKey(name), 'utf8')
+    catch error
+      if error.code is 'ENOENT'
+        return undefined
+      else
+        throw error
```
kevinsawicki: Looks like Atom fails to start if this error is thrown, should it just resort to the default window in this case instead?

maxbrunsfeld: Yeah, thanks for catching that. Fixing that up on master.

kevinsawicki: :heart:

maxbrunsfeld:
> So on restart the focused window may not be the window focused on quit.

Yeah, we should do this. I opened a separate issue for it since it's not totally trivial to do.

src/storage-folder.coffee
maxbrunsfeld: Maybe, but I think we should make that it's own issue, just because it'd be a little bit of a behavior change from what we have now, and I think it'd take a little testing to make sure we didn't ever unintentionally wipe out the user's state.

nathansobo:
:heart: You're amazing @maxbrunsfeld.

jacekkopecky:
This PR doesn't do autosave, does it? Is there another issue beside #1603, or another PR, on that?
There are two aspects to auto-save:
- the open windows and buffers - so that re-opening works also on crash
- the changes in unsaved buffers (including unnamed ones) - so that work doesn't get lost on crash

maxbrunsfeld:
@jacekkopecky does the built-in [autosave](https://github.com/atom/autosave) package do what you want? It saves all buffers whenever the application loses focus.

benogle:
@maxbrunsfeld I think @jacekkopecky might mean to autosave of the application state, rather than the buffers themselves.

maxbrunsfeld:
Oh, yeah. Currently, the application state is only saved on quit.

maxbrunsfeld:
Regarding the application crash use case: I think autosaving the contents of open buffers is critical because it prevents work from being lost. I don't think autosaving the opened windows falls into that category though; I'd rather address that problem by reducing the frequency of crashes.

@jacekkopecky That said, if Atom is crashing on you often enough that reopening your opened windows would be a major improvement, then maybe we should implement that feature. Feel free to open a new issue for this.

mehcode:
@maxbrunsfeld I think it's critical. It's not about atom crashing or not crashing. It's about a myriad of other things crashing on a computer. It's about being able to not kill a junior because he or she accidentally hit the reset button on your computer. Sublime gives you the "don't worry about it" guarantee. It's a state of mind and affects your workflow. I was used to having many unnamed buffers open with random scratch data. I didn't need to save it anywhere but sublime kept them around.

maxbrunsfeld:
Ok @mehcode could you open an issue for that feature?

> It's about being able to not kill a junior because he or she accidentally hit the reset button on your computer.

I think this use case is covered though, because Atom should quit normally during a system restart (at least on OSX and Linux).

denieler:
We are moving forward with this idea with [save-workspace](https://atom.io/packages/save-workspace) plugin. It give opportunity to save whole workspace (tabs, splitting, project folders opened, etc.) and move between them very fast.

PR #6383, opened 2015-04-14
Indicate whether buffer has a file path

mnquintana:
This adds `data-active-path` and `data-active-name` to `atom-pane` for any active buffer that implements `::getPath` – refs atom/tree-view#406, https://github.com/atom/tree-view/pull/406#issuecomment-83653324

This lets you do things based on whether or not a buffer has a file path (like enabling or disabling commands, for instance).

/cc @kevinsawicki @maxbrunsfeld


src/pane-element.coffee
```diff
@@ -71,6 +71,8 @@ class PaneElement extends HTMLElement
     hasFocus = @hasFocus()
     itemView = atom.views.getView(item)
 
+    if item.getPath? then $(itemView).addClass('has-file-path')
```
maxbrunsfeld: It'd be nice to use the native DOM API rather than jQuery here, so that we can eventually remove our jQuery dependency entirely. Also, we usually use the postfix `if` for these one-liners.

```coffee
itemView.classList.add('has-file-path') if item.getPath?
```

mnquintana: No problem – done!

mark-hahn: Way off topic ...

> we usually use the postfix if for these one-liners.

I've always been curious why everyone does this.  To me it hurts
readability because I can't scan down the left looking for branch points.

mnquintana:
@maxbrunsfeld I rewrote this to add some attributes on `atom-pane` instead like we discussed. Let me know how this looks now. :smile:

spec/pane-view-spec.coffee
```diff
@@ -118,6 +118,14 @@ describe "PaneView", ->
...
         expect(pane.itemViews.find('#view-2').length).toBe 1
 
+    describe "when the new activeItem implements ::getPath", ->
+      it "adds its file path to the active pane", ->
+        paneModel.activateItem(editor1)
+        expect(pane).toHaveAttr('data-active-path')
+      it "adds its file name to the active pane", ->
+        paneModel.activateItem(editor1)
+        expect(pane).toHaveAttr('data-active-name')
+
```
maxbrunsfeld: It'd be good to have a test for the converse case as well: when the new active item does not implement ::getPath, it *removes* the attributes. You should be able to use a simple `div` element (`document.createElement("div")`) as the active-item for that case.

To make sure that the attributes are removed, you may want to test both adding and removing the attributes in a single `it`: first activate a `TextEditor`, then activate a `div`.

maxbrunsfeld:
Looks good to me other than the one comment.

mnquintana:
@maxbrunsfeld Addressed your comment!

src/pane-element.coffee
```diff
@@ -71,6 +71,14 @@ class PaneElement extends HTMLElement
     hasFocus = @hasFocus()
     itemView = atom.views.getView(item)
 
+    if itemPath = item.getPath?()
+      path = require 'path'
+      @dataset.activeName = path.basename(itemPath)
```
kevinsawicki: What do you think about calling these `activeItemName` and `activeItemPath` just to be a little clearer?

spec/pane-view-spec.coffee
```diff
@@ -118,6 +118,23 @@ describe "PaneView", ->
...
+        paneModel.activateItem(editor1)
+      it "adds its file path to the active pane", ->
+        expect(pane).toHaveAttr('data-active-path')
+      it "adds its file name to the active pane", ->
+        expect(pane).toHaveAttr('data-active-name')
+
+    describe "when the new activeItem does not implement ::getPath", ->
+      beforeEach ->
+        paneModel.activateItem(editor1)
+        paneModel.activateItem(document.createElement('div'))
```
kevinsawicki: Super minor, but could you add a newline after this line?

spec/pane-view-spec.coffee
```diff
@@ -118,6 +118,23 @@ describe "PaneView", ->
...
+      it "removes its file path from the active pane", ->
+        expect(pane).not.toHaveAttr('data-active-path')
```
kevinsawicki: Super minor, but could you add a newline after this line?

maxbrunsfeld:
Looks good to me other than Kevin's comments.

mnquintana:
@kevinsawicki Done!

src/pane-element.coffee
```diff
@@ -71,6 +71,14 @@ class PaneElement extends HTMLElement
     hasFocus = @hasFocus()
     itemView = atom.views.getView(item)
 
+    if itemPath = item.getPath?()
+      path = require 'path'
+      @dataset.activeItemName = path.basename(itemPath)
+      @dataset.activeItemPath = itemPath
+    else
+      delete @dataset.activeName
```
kevinsawicki: I think these need to still be updated to have `Item` in them

mnquintana: Whoops :grimacing:

mnquintana: Fixed now

kevinsawicki:
Great, thanks for this :ship:

kevinsawicki:
Just a heads up, after testing for a bit, I make one small tweak which is to clear the attributes when the pane items are closed. I noticed they were sticking around when the pane was empty.

https://github.com/atom/atom/commit/9a0c3f94a9c455a7f84af2d0955cb0ade89c618c

mnquintana:
Good catch :+1:

Issue #5469, opened 2015-02-09
Overlay decorations don't overflow out of the editor


oclbdk:
It would be nice if overlay decorations could be positioned outside of the editor, such as for autocomplete. With additional type information, it would be great if we could keep suggestions lined up with what the user is typing.

![screen shot 2015-02-06 at 5 08 19 pm](https://cloud.githubusercontent.com/assets/136982/6114867/24a7bf46-b06f-11e4-8bcf-d6327d6a539a.png)

benogle:
This has been fixed, and has been released. Sorry forgot to close this. Autocomplete-plus also defaults to displaying as shown in the image.

PR #6419, opened 2015-04-17
Add popover panel API

jessegrosjean:
Currently Atom has panel layers for top/left/right/bottom and modal. I think it would also be useful to have a 'popover' panel layer. This would be an absolutely positioned layer above all other window content (except modal panel layer). Panels are absolute positioned within it. Benefits include:
1. A standard place (using existing panel API) for inserting overlay elements. Now it seems the common practice is for packages to insert popover UI arbitrarily into the DOM.
2. Support more use cases then TextEditor 'overlay' decorations, because the popover panel layer would allow popovers anywhere in the Atom window, not just within a text editor. This layer might also be a place to solve https://github.com/atom/atom/issues/5469.
3. Optionally implement popover layout logic once instead of having each package implement it badly once. Popover layout logic is somewhat complex since generally popovers want to be positioned relative to an existing element while also not being clipped by screen edges.

This pull request is a minimal start. It defines the popover layer and API for adding panels. But it doesn't perform any positioning logic. Assuming people like this idea I would also like to add positioning options and logic with the basic goal being:
1. Uses should be able to position panel at arbitrary position on screen
2. Or users should be able to position panel relative to an existing element on screen similar to Cocoa's [NSPopover](https://developer.apple.com/library/mac/documentation/AppKit/Reference/NSPopover_Class/index.html#//apple_ref/occ/instm/NSPopover/showRelativeToRect:ofView:preferredEdge:).

Thoughts?

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -645,6 +645,21 @@ describe "Workspace", ->
         expect(itemView instanceof TestItemElement).toBe(true)
         expect(itemView.getModel()).toBe(model)
 
+    describe '::addPopoverPanel(model)', ->
+      it 'adds a panel to the correct panel container', ->
+        expect(atom.workspace.getPopoverPanels().length).toBe(0)
+        atom.workspace.panelContainers.popover.onDidAddPanel addPanelSpy = jasmine.createSpy()
+
+        model = new TestItem
+        panel = atom.workspace.addPopoverPanel(item: model)
+
+        expect(panel).toBeDefined()
+        expect(addPanelSpy).toHaveBeenCalledWith({panel, index: 0})
+
+        itemView = atom.views.getView(atom.workspace.getPopoverPanels()[0].getItem())
+        expect(itemView instanceof TestItemElement).toBe(true)
+        expect(itemView.getModel()).toBe(model)
+
     describe '::addModalPanel(model)', ->
       it 'adds a panel to the correct panel container', ->
         expect(atom.workspace.getModalPanels().length).toBe(0)
--- a/spec/workspace-view-spec.coffee
+++ b/spec/workspace-view-spec.coffee
@@ -292,5 +292,8 @@ describe "WorkspaceView", ->
       expect(topContainer.nextSibling).toBe workspaceElement.paneContainer
       expect(bottomContainer.previousSibling).toBe workspaceElement.paneContainer
 
+      popoverContainer = workspaceElement.querySelector('atom-panel-container.popover')
+      expect(popoverContainer.parentNode).toBe workspaceElement
+
       modalContainer = workspaceElement.querySelector('atom-panel-container.modal')
       expect(modalContainer.parentNode).toBe workspaceElement
--- a/src/panel-element.coffee
+++ b/src/panel-element.coffee
@@ -23,6 +23,7 @@ class PanelElement extends HTMLElement
   attachedCallback: ->
     callAttachHooks(@getItemView()) # for backward compatibility with SpacePen views
     @visibleChanged(@getModel().isVisible())
+    @positionPopover()
 
   visibleChanged: (visible) ->
     if visible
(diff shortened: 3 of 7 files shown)


jessegrosjean:
Here's a proposal for positioning:

```
formattingBarPanel = atom.workspace.addPopoverPanel
  item: formattingBar
  position:
    xAnchor: 0.5 # center
    yAnchor: 1.0 # bottom
    targetXAnchor: 0.5 # center
    targetYAnchor: 0 # top
    targetRect: =>
      @getClientRectForItemOffset(item, offset)
```

In this example I'm showing a Medium style formatting bar over a custom (non TextEditor) view. Positioning is based on x,y anchor points between two rectangles: the panel rect and the target rect. In the above example the formatting bar panel is horizontally centered over the target rect and vertically the formatting bar panel's bottom is aligned with the target rects top.

Target rect can be defined statically as a `top,bottom,left,right rectangle`, or it can be defined as a function that returns a rectangle. If it is defined as a function a `requestAnimationFrame` loop is started updating the panel position in each frame. This way the panel can track items as they are scrolled or otherwise moved on screen.

I'm using this in my own app testing and it's working pretty well and is flexible to support a number of different positioning strategies. Though the `xAnchor, yAnchor` etc names aren't particularly clear. Thoughts on improvements? Does this seem like reasonable positioning strategy for popover panels?

Jesse

benogle:
Can you show a screenshot of this? And maybe outline some usecases? What packages would use this and why?

jessegrosjean:
Hi,

Bootstrap provides a good generic popover example:

http://getbootstrap.com/javascript/#popovers

Important features are:

1. Displayed in layer over normal UI
2. Positioned relative to some other object, for example above, below, left, right
3. Optionally include logic for auto positioning to avoid window edge clipping. So for example if displaying below a given position will put the popover out of the viewport then instead position above it.
4. Optionally include logic to update the popover position anytime something changes, such as window being resized or related object changing position.

It's the same concept and use case as Atom already provides with `overlay` decorations (so uses such as autocomplete window or color-picker, though at the moment color-picker isn't built as `overlay` decoration... probably should be). The problem with overlay decorations is that they only work for overlaying over the text editor.

This API would allow you to overlay over any aspect of Atoms UI. Places where this might be used:

1. Atom Tree-View might implement a "show-info" command. It could popup file info (size, created, modified, etc) for the select file.

2. Tree-View when your rename a file you are show a modal dialogue. The rename interaction might be a more clear if instead it was shown in a popover attached to the file that's being renamed.

3. In Settings > Packages when click to see a packages settings you are taken to another settings/screen and it's a pain to "go back" to your previous packages list view. If instead of being taken to a separate screen the package settings were displayed in a popover that wouldn't be a problem.

4. There are a number of built in features that are creating their own popover style UIs. For example notifications, tooltips, and TextEditor `overlay` decorations. I don't know if it makes sense in all cases, but if this popover API was in place I think they could be implemented in terms of the popover UI instead of building things from scratch each time.

5. For my particular use case, I'm creating a new outline file type and it's edited in its own view (ie not a TextEditor). I'd like to create UI such as Medium's formatting bar, autocomplete, etc. This API would allow me to do that using standard Atom technology (ie workspace Panels) instead of just picking some arbitrary location in the DOM to insert it all.

Heres an in progress screenshot of a "tag editor" UI that I popup when the user wants to edit tags for a selected outline item:

tageditor2

Jesse

jessegrosjean:
The positioning spec example above is just a direct port of what I've been using in my own version of this stuff. I think for something public it would probably be better to base it off one of the existing popover APIs such as:

- http://getbootstrap.com/javascript/#popovers-options

mark-hahn:
Having this would have saved me a ton of work in my `got-diff-popup`
package and a bit of work in other ones.

jessegrosjean:
I've just updated the request with positioning logic that I'm using in my own app. The basic idea is that when you create a popover your provide:

1. target - A target rectangle to position around
2. viewport - A constraint rectangle in which to keep the popover
3. placement - Desired popover placement 'top center', 'top left', 'left bottom', etc.

Both `target` and `viewport` can be specified as rects, elements, or a function. If specified as an element or function then the positioning is continuous scheduled view requestAnimationFrame. This makes it easy to keep the popover positioned over a scrolling or otherwise dynamic position.

Of course this type of loop isn't ideal, but I think it's a simple way to handle lots of different cases. It's smart in that it only updates the DOM when the position has changed. And generally popovers are modal, there shouldn't be many on screen, and they shouldn't stick around long.

I'm including this positioning logic as a proposal looking for feedback. If it's a direction that seems useful then I can polish it up a bit more. Otherwise I can remove it from this pull request.

Jesse

steelbrain:
:+1:

damieng:
PR is out of sync and hasn't been merged in 2 years.

Issue #1783, opened 2014-03-22
Atom doesn't soft wrap CJK(Chinese/Japanese/Korean) correctly


izuzak:
Reported in Halp:
* support/b47caffcaba511e3903bfcd96bef4da2

> If I have a paragraph in CJK characters and toggle soft wrap, the wrapping point (should be the right edge of current window) is far away than the current edge and some characters are displayed out of window which is not visible.
> I think this may because of the wrong processing of character wide. A latin letter has a width "one character width", but usually a CJK character has a width twice as a latin letter, it's "two character width". Hope this information may help you.
> The debug info auto attached should include the test text I open now. It's include (from up to down): Chinese, Japanese, Korean, English. Actully they tell the same thing:)

I can reproduce this in Atom 0.75.0. (notice how CJK characters are not wrapped correctly at the right edge of the app)

![screen shot 2014-03-22 at 7 24 05 pm](https://f.cloud.github.com/assets/38924/2491493/404ad7da-b1ef-11e3-9143-272b55530c6c.png)

allenfantasy:
Thank you very much! Though [japanese-wrap] is for Japanese but I found it works perfect in Chinese. @SaschaNaz

DeckerCHAN:
@SaschaNaz Thankyou! Works good for me!

clkao:
@SaschaNaz++! works perfectly for zh-hant as well.

hanfeisun:
hope the issue could be fixed officially +1

Maki-Daisuke:
Great!!  @SaschaNaz++!

kmxz:
After updating my Atom to the latest version, the AtomicChar package no longer works... so sad.
Why isn't this issue officially fixed yet?

SaschaNaz:
I'm investigating. Maybe I should just send a pull request...

SaschaNaz:
@kmxz New version should now work. Thanks.

kmxz:
@SaschaNaz The new version works perfectly. Thank you very much,

SaschaNaz:
I'm working on this for a pull request.

raccy:
@SaschaNaz Great working!
I'm looking forward to that your work is complete.
and... I'll be happy, because I don't have to fix japanese-wrap with `measureText`,
and for Chinese and Korean.
やっとメンテナンスの重圧から逃れられる(-д-; (Do not translate!)

SaschaNaz:
[...]

PR #6425, opened 2015-04-17
New text wrapper, measuring actual text widths

saschanaz:
Fixing #1783. Both Soft Wrap and Soft Wrap At Preferred Line Length works.

PS: Well, I just found there's another pull request which seems to have similar purpose. I don't know... Hmm.


benogle:
Hey @SaschaNaz thanks for the PR. We're currently changing the underpinnings of buffer storage, including the way soft wrapping is handled in https://github.com/atom/text-document. We'd rather not potentially destabilize the current implementation at this time, and should fix this issue with the https://github.com/atom/text-document upgrade. Please check out and follow that repo for more information.

SaschaNaz:
@benogle Okay, let's just stick to the current external package. Well... I think `text-document` still soft-wraps based on "column", not actual text width. Does it currently receive pull requests?

nathansobo:
Hey @SaschaNaz. Reopening this for conversation purposes, but I think we'll need to structure things differently. As I mentioned in your other PR, I think we need to tackle text measurement in a general way and then apply it to soft wrapping as one specific case. Any interest in experimenting with the approach to measurement I discussed in #6083?

SaschaNaz:
@nathansobo Sorry for my late response. I'm not sure what's next, did you mean we have to replace canvas with iframe+DOM Range API? Was there some performance comparison test between these two approach?

nathansobo:
@as-cii will probably resuming work on #6083 in a couple of weeks. He can probably speak more to this, because he's tried both approaches. @as-cii, maybe we could post some profile information comparing the two? Ultimately, I want to use whatever is fastest. In the event of a tie, I'd rather use the DOM approach because I'm not that excited about depending on the canvas and the DOM having the same rendering code path forever. It's true now, and on this browser, but if we can avoid tying ourselves to it I think it would be better.

as-cii:
I apologize for replying a bit late but hopefully the following insights will be worth the delay.

So, at the moment I do not have any fancy graphic at hand but the canvas approach takes approximately ~ 4ms to measure every new line, whereas the `getBoundingClientRect` technique can take up to twice more. In a tight loop like scrolling, however, neither are sufficiently fast and, therefore, we need to tackle this problem in a different way. Moreover, as @nathansobo pointed out, it is better not to rely on canvas to measure text as its rendering path could diverge from the DOM one unexpectedly. We need a resilient, yet fast solution.

For these reasons, we opted to try out a new approach where we don’t measure every new line eagerly but, instead, we do it lazily using `getBoundingClientRect`. This should reduce wasted computation to a minimum, thereby allowing to complete measurements before reaching the 16ms budget we have for each animation frame. On a side note, using the canvas here could potentially be wasteful, as we would need to compute all the positions preceding the requested character’s one in order to calculate it. Any solution that I can think of feels quite hacky, but I am afraid this is what we’ll end up employing before a real measurement API lands in Chromium.

You can find more insights about this technique on https://github.com/atom/atom/issues/6055#issuecomment-110858502: I plan to resume the work I made on this pretty soon, so I will definitely keep you guys posted meanwhile I make progress.

SaschaNaz:
> it is better not to rely on canvas to measure text as its rendering path could diverge from the DOM one unexpectedly

This is a real problem as my current package cannot correctly measure a text with a thickened text inside it.

> For these reasons, we opted to try out a new approach where we don’t measure every new line eagerly but, instead, we do it lazily using `getBoundingClientRect`.

I think it may make our code much more complicated, but it's anyway interesting. I'd love to test it with my package.

PR #6438, opened 2015-04-20
use electron instead of atom-shell

Jonahss:
The one thing I can't understand is why the built app is called Electron.app instead of Atom.app. Help?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ node_modules
 npm-debug.log
 debug.log
 /tags
-/atom-shell/
+/electron/
 docs/output
 docs/includes
 spec/fixtures/evil-files/
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -22,7 +22,7 @@ module.exports = (grunt) ->
   grunt.loadNpmTasks('grunt-contrib-coffee')
   grunt.loadNpmTasks('grunt-contrib-less')
   grunt.loadNpmTasks('grunt-shell')
-  grunt.loadNpmTasks('grunt-download-atom-shell')
+  grunt.loadNpmTasks('grunt-download-electron')
   grunt.loadNpmTasks('grunt-atom-shell-installer')
   grunt.loadNpmTasks('grunt-peg')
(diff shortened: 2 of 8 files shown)


kevinsawicki:
I was hoping to rename the `grunt-download-atom-shell` library to `grunt-download-electron` today after getting it working again with the new name.

Looks like you published a package by that name though, would you be okay unpublishing that so that we don't have to pick a different name and it is less confusing for people migrating their apps?

I'd be happy to add you as a collaboration on the repo under the Atom org. if you'd like.

kevinsawicki:
> The one thing I can't understand is why the built app is called Electron.app instead of Atom.app.

Yeah, atom-shell defaulted to calling everything `Atom` so Atom was using the release as is instead of having to make any modifications.

This was one of the motivations about renaming it, so Atom would have to start using it like other apps do and would feel any pain with the build process that could be simplified or automated.

So it is going to take more work to upgrade Atom to use Electron that just upgrading the grunt plugins. For example, the Windows build will need to change to completely rebuild Electron so that it has an `atom.exe` instead of `electron.exe`.

Jonahss:
Ah! I suspected that the default package name had changed, thanks for confirming.
Yeah, halfway through I noticed that separate grunt-atom-shell-installer. I was just fooling around on the weekend :)

Happy to unpublish grunt-download-electron.

kevinsawicki:
> Happy to unpublish grunt-download-electron.

Awesome, thanks so much :green_heart:

Jonahss:
@kevinsawicki it looks like `npm unpublish` would still disallow republishing it with an already-used version number. Can I just transfer ownership to you?

kevinsawicki:
> Can I just transfer ownership to you?

Sure, that would be great

Jonahss:
`kevinsawicki` added as owner, feel free to `npm owner rm jonahss`

kevinsawicki:
great, thanks so much!

kevinsawicki:
Just a heads up, I renamed the repo and published a 2.0.0 version as `grunt-download-electron`, thanks again!

PR #6446, opened 2015-04-20
:arrow_up: babel-core@5.1.11

bolinfest:
Bump babel-core from 4.0.2 to 5.1.11.

Note that this is a major version bump, so `babel.coffee` also had to be updated to reflect the minor API changes.

Notable new features from the [changelog](https://github.com/babel/babel/blob/master/CHANGELOG.md) since 4.0.2:
- Add [trailing function comma proposal](https://github.com/jeffmo/es-trailing-function-commas).
- Decorators based on [@wycat's stage 1 proposal](https://github.com/wycats).

There is a number of bug fixes and performance improvements, as well.

One new error that users are likely to run into is this one in subclass constructors that reference `this` but not `super`:

```
'this' is not allowed before super()
```

As best I can tell, this was a recent-ish update to the ES6 constructor semantics spec that Babel is now enforcing: https://gist.github.com/allenwb/53927e46b31564168a1d. /cc @sebmck

### DIFF
--- a/src/babel.coffee
+++ b/src/babel.coffee
@@ -28,10 +28,6 @@ defaultOptions =
     'useStrict'
   ]
 
-  # Includes support for es7 features listed at:
-  # http://babeljs.io/docs/usage/transformers/#es7-experimental-.
-  experimental: true
-
   optional: [
     # Target a version of the regenerator runtime that
     # supports yield so the transpiled code is cleaner/smaller.
@@ -42,6 +38,11 @@ defaultOptions =
     # JSX transformer produces pre-v0.12 code.
     'reactCompat'
   ]
+
+  # Includes support for es7 features listed at:
+  # http://babeljs.io/docs/usage/experimental/.
+  stage: 0
+
 
 ###
 shasum - Hash with an update() method.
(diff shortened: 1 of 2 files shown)


bolinfest:
Sorry, I shouldn't have put this up for review quite yet. I need to see which flags, if any, I need to change in `babel.coffee`.

bolinfest:
OK, I tested this with our internal packages, and the only think I needed to fix were the `super()` calls that I mentioned in the commit message. Violators are not ES6-compliant, so I don't think it's unreasonable to require them to fix their own code.

src/babel.coffee
```diff
@@ -43,6 +39,11 @@ defaultOptions =
     'reactCompat'
   ]
 
+  # Includes support for es7 features listed at:
+  # http://babeljs.io/docs/usage/experimental/.
+  'stage': 0
```
kevinsawicki: :fire: the quotes

thomasjo:
In order to get this to build on CI, you should rebase it on top of `master` (or merge if you prefer that).

bolinfest:
Rebased and killed quotes.

maxbrunsfeld:
Looks good to me. :ship:

PR #6447, opened 2015-04-20
Do not bundle package specs

kevinsawicki:
#6395 removed core specs and fixtures and this PR removes package specs and fixtures from the bundle to reduce the size.

It also removes a bunch of other unneeded files from the app bundle that I noticed while auditing.

This removes around **~2,500** unneeded files from the app which shrinks `app.asar` by 3MB.

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -69,6 +69,8 @@ module.exports = (grunt) ->
       expand: true
       src: [
         'src/**/*.coffee'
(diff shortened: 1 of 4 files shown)


50Wliu:
Wow.  :clap:

benogle:
How much does this reduce the slug by?

kevinsawicki:
> How much does this reduce the slug by?

It reduced `app.asar` by 3 megs

PR #6449, opened 2015-04-21
Perform DOM updates requested from DOM reads in the same frame, after we finish reading

nathansobo:
A common pattern is to put something on the DOM, measure it, then update the DOM again based on that measurement. This change ensures that any updates requested as a result of reading from the DOM get scheduled for the end of the current frame. If you want to read _again_ after these follow-on updates, you will need to wait for the next frame. But at least this way we ensure instant feedback with minimal thrashing (1 reflow) when we have no choice but to read the DOM before updating.

/cc @benogle

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -2624,7 +2624,7 @@ describe "TextEditorComponent", ->
       expect(componentNode.querySelector('.placeholder-text')).toBeNull()
 
   describe "legacy editor compatibility", ->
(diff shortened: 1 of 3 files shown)


benogle:
Cool, thanks for doing this.

benogle:
I'm going to feature detect in AC+ by checking for `atom.views.documentReadInProgress?` so it will be jitter-free in all versions. Good?

nathansobo:
That sounds fine.

PR #6450, opened 2015-04-21
Make Travis build more resilient

as-cii:
As you may have noticed, our Travis builds fail more often than not. That's due to our specs running in parallel, which slow down one another since (I suppose) we're on a single-core machine.

So, the important changes which we're going to introduce with this PR are:
- Set concurrency to `1` when we're on Travis;
- Use `sudo: false` infrastructure;

This should lead to a much consistent test result across builds (hopefully reducing spurious errors).

_Side note: once we manage to make specs working flawlessly on Linux, things should get much faster (and, in addition, we get caching for free)._


mnquintana:
Just for posterity, it turns out that [Travis VMs run on 1.5 cores](http://docs.travis-ci.com/user/speeding-up-the-build/#Parallelizing-your-build-on-one-VM), which... I'm not quite sure what that means. (What is half a core??)

maxbrunsfeld:
Great work @as-cii! 

On a sidenote, what does everybody think about only building with node 0.12 on travis? I think that it might be overkill to test with node 0.10, because the version of node.js we use only affects our build scripts, not our actual application code. I think running fewer total builds might reduce our queue time and the number of flaky failures.

as-cii:
> Just for posterity, it turns out that Travis VMs run on 1.5 cores, which... I'm not quite sure what that means. (What is half a core??)

@mnquintana: yeah, but I am not sure this holds true for their Mac infrastructure as well. I think the documentation referred to the old Linux workers (e.g. the containerized ones use 2 cores).

> Great work @as-cii!

> On a side-note, what does everybody think about only building with node 0.12 on travis? I think that it might be overkill to test with node 0.10, because the version of node.js we use only affects our build scripts, not our actual application code. I think running fewer total builds might reduce our queue time and the number of flaky failures.

@maxbrunsfeld: ideally, each build in the matrix shouldn't impact the queue time. However, they're at capacity on their Mac infrastructure right now (more workers are coming soon, apparently), therefore I believe your suggestion totally makes sense! I'll make sure to remove `0.10` in a while :+1: 

By the way, I think Travis is currently experiencing some troubles (http://www.traviscistatus.com/incidents/crzxldg09nnr), therefore the build results that are showing up may not be super reliable :disappointed:

kevinsawicki:
Looks good, thanks, hopefully travis resolves that incident soon.

PR #6456, opened 2015-04-21
core:select-up/down should work in mini editor as well as main editor.

### LINKED ISSUE #6445: Shift-up at the end of input should select all.
In `atom-text-editor:mini`, when I get to the end of the line, `shift-up` should select all of the text in the input like a native input box would. This is not working in Atom 0.192.0 on OS X.

bolinfest:
Fixes #6445.


kevinsawicki:
Starting to wonder if these were intentionally not included originally since `core:move-up` and `core:move-down` aren't included as well.

@nathansobo, @maxbrunsfeld, @benogle do you remember and/or have feelings about this?

maxbrunsfeld:
I like it. We should probably include `move-up` and `move-down` too, assuming they work.

kevinsawicki:
> We should probably include move-up and move-down too, assuming they work.

Yeah, I wonder how this will play with the select list.

bolinfest:
@kevinsawicki What's an example of where the select list is used nowadays?

kevinsawicki:
>  What's an example of where the select list is used nowadays?

Fuzzy finder, command palette, and symbols view.

bolinfest:
Ah, I've been testing the current PR with fuzzy finder and that works fine.

I haven't tried `core:move-up` and `core:move-down` yet, but I suspect they are probably trickier, as you point out.

Could we get this in and we can file a separate issue for `core:move-up` and `core:move-down`?

kevinsawicki:
> Could we get this in and we can file a separate issue for core:move-up and core:move-down?

PR #6457, opened 2015-04-21
Add david-dm dependency status badge to Readme

mnquintana:
This adds a dependency status badge (from https://david-dm.org) to the Readme to make it a bit easier to keep track of dependency up-to-date-ness. Unfortunately, this only works for dependencies in `atom/atom/package.json` and not `atom/atom/build/package.json` (they don't support subdirectories... yet). This should make it a bit easier to keep track of dependency versions and see when we're out of date.

Eventually I'd like to add this to all package Readmes too, but that will come with @atom/issue-triage's big rewrites.


kevinsawicki:
Sounds good, thanks for adding this :anchor:

PR #6459, opened 2015-04-21
Upgrade insecure dependencies

kevinsawicki:
Result of looking at initial report from #6457 

/cc @mnquintana


mnquintana:
:+1: There are probably more from build/package.json (and definitely more from individual packages), but this should be good to start

PR #6477, opened 2015-04-22
Move some Grim deprecation calls to correctly attribute them with packages

izuzak:
_See https://github.com/atom/deprecation-cop/issues/36#issuecomment-94856987 for motivation._

@kevinsawicki @maxbrunsfeld I moved the two calls mentioned in the comment above :point_up: and that resolved the problem for those two calls. Here's how those two deprecations look now in Deprecation Cop:

screen shot 2015-04-22 at 15 27 18

:sparkles: -- notice that they are attributed to pdf-view, not to Atom core.

However, there are two other `Grim.deprecate` calls which are still being incorrectly attributed to Atom core -- the ones which check if the value returned from `onDidChangeModified` and `onDidChangeTitle` are `Disposable`s. 

screen shot 2015-04-22 at 15 31 08

Those two calls are [here](https://github.com/atom/atom/blob/ecd04446434e47c505d7f1d3d8cb0fb76e91aa64/src/pane-view.coffee#L119) and [here](https://github.com/atom/atom/blob/ecd04446434e47c505d7f1d3d8cb0fb76e91aa64/src/pane-view.coffee#L128). As far as I can tell, I can't apply the same approach to these because they check if the _return value_ has a specific type. So it needs to stay there, I _think_.

Also, the "magic" which helps us solve the problem for the two deprecations in this PR is that we explicitly pass a package name to Grim.deprecate to tell it what package is involved. That "magic" is implemented in [atom.workspace.getCallingPackageName()](https://github.com/atom/atom/blob/ecd04446434e47c505d7f1d3d8cb0fb76e91aa64/src/workspace.coffee#L483). So, I tried calling this method from `pane-view.onActiveItemChanged()`, where the two deprecations checking for a `Disposable` return values are located, but that call returned `undefined`.

I'm not sure how to fix those two deprecations -- would love your help with those, if you have ideas. :pray: Also, maybe we can just remove them?

### DIFF
--- a/src/pane-view.coffee
+++ b/src/pane-view.coffee
@@ -116,7 +116,6 @@ class PaneView extends View
 
     if item.onDidChangeTitle?
       disposable = item.onDidChangeTitle(@activeItemTitleChanged)
-      deprecate 'Please return a Disposable object from your ::onDidChangeTitle method!' unless disposable?.dispose?
       @activeItemDisposables.add(disposable) if disposable?.dispose?
     else if item.on?
       disposable = item.on('title-changed', @activeItemTitleChanged)
@@ -124,7 +123,6 @@ class PaneView extends View
 
     if item.onDidChangeModified?
       disposable = item.onDidChangeModified(@activeItemModifiedChanged)
-      deprecate 'Please return a Disposable object from your ::onDidChangeModified method!' unless disposable?.dispose?
       @activeItemDisposables.add(disposable) if disposable?.dispose?
     else if item.on?
       item.on('modified-status-changed', @activeItemModifiedChanged)
--- a/src/pane-view.coffee
+++ b/src/pane-view.coffee
@@ -119,7 +119,6 @@ class PaneView extends View
       deprecate 'Please return a Disposable object from your ::onDidChangeTitle method!' unless disposable?.dispose?
       @activeItemDisposables.add(disposable) if disposable?.dispose?
     else if item.on?
-      deprecate 'If you would like your pane item to support title change behavior, please implement a ::onDidChangeTitle() method. ::on methods for items are no longer supported. If not, ignore this message.'
       disposable = item.on('title-changed', @activeItemTitleChanged)
       @activeItemDisposables.add(disposable) if disposable?.dispose?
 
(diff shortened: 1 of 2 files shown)


izuzak:
:circus_tent: note: the tests are failing with this error in the notifications package:

```
"Invalid Version: 0.195.0-0207016", source: /private/var/lib/jenkins/workspace/atom/node_modules/notifications/lib/notification-element.coffee (39)
>> [8835:0422/065012:INFO:CONSOLE(40)] "TypeError: Invalid Version: 0.195.0-0207016
>>   at new SemVer (/private/var/lib/jenkins/workspace/atom/node_modules/semver/semver.js:295:11)
>>   at compare (/private/var/lib/jenkins/workspace/atom/node_modules/semver/semver.js:571:10)
>>   at Function.gte (/private/var/lib/jenkins/workspace/atom/node_modules/semver/semver.js:620:10)
>>   at /private/var/lib/jenkins/workspace/atom/node_modules/notifications/lib/user-utilities.coffee:120:47
>>   at process._tickCallback (node.js:367:9)
```

The reason why that version is invalid is that the pre-release part `0207016` starts with a leading zero, which is [not allowed](http://semver.org/).

```
> semver = require("semver")
> semver.valid("0.195.0-0207016")
null
> semver.valid("0.195.0-1207016")
'0.195.0-1207016'
```

In other words, I don't believe that this PR broke anything or that there's a problem in those tests.

:carousel_horse:

kevinsawicki:
> Also, maybe we can just remove them?

I think we can, I think they were trying to cover the case where old packages implemented `onDidChangeTitle` before core started using it and it was to warn people that it is now an official API that their item should implement a specific way.

Since these method names are pretty formal, I'm optimistic either no packages ever implemented them previously the "wrong" way or those packages have since been updated.

I think just assuming disposables are returned and throwing a real error when they aren't is the right move now.

kevinsawicki:
> I think just assuming disposables are returned and throwing a real error when they aren't is the right move now.

Actually since `PaneView` is already going away, I think just removing the deprecations and things failing the right way with `atom --one` is the right move going forward.

kevinsawicki:
Thanks for fixing these up :blue_heart:

izuzak:
:zap: for finishing this, @kevinsawicki :checkered_flag:

PR #6489, opened 2015-04-23
Fix #6466 by calling app.quit() before last window are closes.

### LINKED ISSUE #6466: Reopening windows feature from 193 not working
Does the Desktop shortcut on Windows need updating for this to work, maybe? Because it does not work for me atm, see video: https://youtu.be/SETJog-VnkY

brettle:
`app.saveState()` was being called after the window was closed, resulting in no state being saved.


anthonyraymond:
:+1:

maxbrunsfeld:
@brettle thanks for the PR; it looks totally correct to me. I'm sorry that I didn't see this one earlier; I woudl have merged it, but I have another PR going (#6684) that goes a little further and saves the state whenever a window is added, removed, or loses focus. I think this will be beneficial because some users are in the habit of killing Atom forcibly through some task-manager application, or running it in the foreground from the terminal and killing it with `ctrl-c`.

Zireael07:
@maxbrunsfeld: Some (well, most) people close on windows by using the 'X' instead of File-Quit and it didn't call the functions correctly either.

Please merge either this or #6684 asap to stop the 'I closed via X and have to reopen everything' situation.

brettle:
Fixed by other commits in upstream.

PR #6497, opened 2015-04-23
Add a handler for an `unhandledRejection`.

bolinfest:


thedaniel:
Whether this should be in Electron or not (I think it should be) I'm in favor of merging this asap and pulling it out later when it lands over there.

bolinfest:
@thedaniel Thanks, that would be a big help!

PR #6499, opened 2015-04-23
Update csslint version to 0.10.0 so that FreeBSD is happier.

pusateri:
csslint removed their architecture limitations in 0.10.0 and so it no longer stops building on FreeBSD.


kevinsawicki:
Thanks for updating these.

PR #6501, opened 2015-04-23
Update encoding-selector@0.20.0, spell-check@0.56.0, symbols-view@95.0

mnquintana:
:arrow_up:


kevinsawicki:
Thanks for upgrading these :boat:

PR #6505, opened 2015-04-24
Replace to electron

tomoya:
Replace to electron from atom-shell


kevinsawicki:
Closing this out, this has drifted quite and bit and is now being taken up in https://github.com/atom/atom/issues/7877

tomoya:
:+1:

Issue #6310, opened 2015-04-08
After updating to Atom 0.190.0 I get following deprecation notification of Atom core


rukshn:
**After updating to Atom 0.190.0 I get the following 3 deprecation notices of Atom core.** Anyone else noticed this?

Use a config schema instead. See the configuration section of https://atom.io/docs/latest/hacking-atom-package-word-count and https://atom.io/docs/api/latest/Config for more details
Called 1 time
Package.activateConfig - /usr/share/atom/resources/app/src/package.js:250:11
Package.activateNow - /usr/share/atom/resources/app/src/package.js:221:14

--------------------------------------------------------------------------------------------------------------

Use activationCommands instead of activationEvents in your package.json Commands should be grouped by selector as follows:

  "activationCommands": {
    "atom-workspace": ["foo:bar", "foo:baz"],
    "atom-text-editor": ["foo:quux"]
  }
Called 1 time
Package.getActivationCommands - /usr/share/atom/resources/app/src/package.js:764:9
Package.hasActivationCommands - /usr/share/atom/resources/app/src/package.js:689:20

----------------------------------------------------------------------------------------------------------------
ContextMenuManager::add has changed to take a single object as its argument. Please see https://atom.io/docs/api/latest/ContextMenuManager for more info.
Called 1 time
ContextMenuManager.add - /usr/share/atom/resources/app/src/context-menu-manager.js:57:18
Package.activateResources - /usr/share/atom/resources/app/src/package.js:299:61

izuzak:
This looks very similar to https://github.com/atom/atom/issues/6278. @kevinsawicki -- https://github.com/atom/atom/issues/6278 turned out to be a problem in two packages. Would it be possible to somehow not report these as Atom core problems, but rather package problems?

rukshn:
@izuzak It's shown as deprecation notices for atom core, doesn't show as a problem with any packages. That's why reported it as an atom core problem because I couldn't find any packages that's causing the problem

rukshn:
Starting Atom in safe mode generates only 2 error,

Use a config schema instead. See the configuration section of https://atom.io/docs/latest/hacking-atom-package-word-count and https://atom.io/docs/api/latest/Config for more details
Called 1 time
Package.activateConfig - /usr/share/atom/resources/app/src/package.js:250:11
Package.activateNow - /usr/share/atom/resources/app/src/package.js:221:14

----------------------------------------------------------------------

[...]

PR #6522, opened 2015-04-26
Add new Grim.deprecate call for context menu CSON format so that it can be correctly attributed to packages

izuzak:
@kevinsawicki @maxbrunsfeld I _think_ this is the last of the fixes for the reports in https://github.com/atom/atom/issues/6310. :sweat_smile: 

Currently, there's no special `Grim.deprecate` call for the old context menu CSON format -- it's being detected [within `ContextMenuManager.add()`](https://github.com/atom/atom/blob/14161a63fe1b56aed834050bd8bb56180e6acfb3/src/context-menu-manager.coffee#L115-L123), which is called from [`Package.activateResources()`](https://github.com/atom/atom/blob/14161a63fe1b56aed834050bd8bb56180e6acfb3/src/package.coffee#L194). So, the `Grim.deprecate` call within `ContextMenuManager.add()` catches both calls to `ContextMenuManager.add()` from packages (adding menu items after activation) and calls to `ContextMenuManager.add()` from Atom when it adds menu items from packages' cson file at activation.

For that reason, the `Grim.deprecate` call in cases when `ContextMenuManager.add()` is called from Atom at package activation is incorrectly attributed to Atom itself. Also, the deprecation text is a bit confusing because it says that `ContextMenuManager.add()` was called, but the package didn't make such a call directly.

This PR adds a new `Grim.deprecate` call to `Package.activateResources()` _before_ `ContextMenuManager.add()` is called. This catches the old-context-menu-CSON-format situation so that it is correctly attributed to packages and a clearer message can be shown. In addition, the conversion from the old format to the new format happens in `Package.activateResources()` as well so that the `Grim.deprecate` call within `ContextMenuManager.add()` isn't triggered in that case. The `Grim.deprecate` call within `ContextMenuManager.add()` wasn't removed because it still needs to catch direct calls to that public method from packages.

Before (call attributed to Atom):

screen shot 2015-04-26 at 11 08 40

After (call attributed to package + a bit more direct deprecation message)

screen shot 2015-04-26 at 11 11 41

Does this approach look okay?

### DIFF
--- a/src/context-menu-manager.coffee
+++ b/src/context-menu-manager.coffee
@@ -105,9 +105,9 @@ class ContextMenuManager
       # Detect deprecated file path as first argument
       if itemsBySelector? and typeof itemsBySelector isnt 'object'
         Grim.deprecate """
-          ContextMenuManager::add has changed to take a single object as its
+          `ContextMenuManager::add` has changed to take a single object as its
           argument. Please see
-          https://atom.io/docs/api/latest/ContextMenuManager for more info.
+          https://atom.io/docs/api/latest/ContextMenuManager#context-menu-cson-format for more info.
         """
         itemsBySelector = arguments[1]
         devMode = arguments[2]?.devMode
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Looks good to me. :ship:

PR #6526, opened 2015-04-27
:bug: Put a limit of 100 on max font size

### LINKED ISSUE #4880: Add input checking on Settings, Font Size
My young son typed in 1000000 for the Font Size in settings and Atom froze. I had to edit the config back to a sane number with a different editor.

Might I suggest limiting the setting to a range, maybe 4 to 1000?

lee-dohm:
This will prevent Atom from crashing, but still allow folks who need large
font sizes for visual impairments to have them.

Fix #4880

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -108,6 +108,7 @@ module.exports =
         type: 'integer'
         default: 16
         minimum: 1
+        maximum: 100
       lineHeight:
         type: ['string', 'number']
         default: 1.3


mnquintana:
Hmm, I just noticed an interesting bug - I tried entering in a value > 100, and when Atom bumped the value back down, it wiped my config.cson completely.

lee-dohm:
Interesting, I'll have to try that.

lee-dohm:
I can't replicate the behavior you mention, though I did get it stuck saying 1,000 even though visually it was still 100. I'll have to see if I can track that down with a solid repro.

mnquintana:
It might've been that I tried doing it multiple times in rapid succession (I tried entering a value >100, waited for it to change to 100, then did it again as fast as I could).

lee-dohm:
Yes, that's exactly what I was doing ... though it still didn't wipe my `config.cson`:

1. Open Settings View
1. Move to "Font Size" field
1. Add three zeros
1. Wait for the value to change to 100
1. Add three zeros again

**Expected:** Atom to change the value back to 100
**Actual:** Atom leaves the bad value in the text field (though the good value of 100 in the `config.cson` is saved)

benogle:
> Expected: Atom to change the value back to 100
> Actual: Atom leaves the bad value in the text field (though the good value of 100 in the config.cson is saved)

This is probably a settings view issue

This change seems good. The config wiping is sketch, though. @mnquintana have you reproduced it again?

mnquintana:
@benogle Not yet, I'll have to try it again when I get home (and actually backup my config first this time :grimacing: )

mnquintana:
I haven't been able to reproduce that - not sure what happened there :confused:

lee-dohm:
I opened the issue with the settings view. Hopefully the config wiping was a fluke, though it is disturbing even as a fluke.

kevinsawicki:
:ship: :+1:

PR #6527, opened 2015-04-27
Add font size commands to View menu

mnquintana:
Adds "Increase Font Size", "Decrease Font Size", and "Reset Font Size" to the View menu, which is typical in other editors (TextMate, Brackets) and should help in discoverability.

### DIFF
--- a/menus/linux.cson
+++ b/menus/linux.cson
@@ -125,6 +125,10 @@
         ]
       }
       { type: 'separator' }
+      { label: '&Increase Font Size', command: 'window:increase-font-size' }
+      { label: '&Decrease Font Size', command: 'window:decrease-font-size' }
+      { label: '&Reset Font Size', command: 'window:reset-font-size' }
+      { type: 'separator' }
       { label: 'Toggle Soft &Wrap', command: 'editor:toggle-soft-wrap' }
     ]
   }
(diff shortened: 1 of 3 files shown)


izuzak:
What do you think about grouping these as sub-items in a `View > Font` menu? That's what TextMate does.

mnquintana:
@izuzak I wouldn't mind that, but I do tend to dislike too many submenus, so I'm not sure.

lee-dohm:
I think the View menu is compact enough that I could easily go either way on this one.

mnquintana:
I think I'll leave it as-is for now, and if the View menu gets really long we can always move it to a Font submenu then.

menus/linux.cson
```diff
@@ -125,6 +125,10 @@
         ]
       }
       { type: 'separator' }
+      { label: '&Increase Font Size', command: 'window:increase-font-size' }
+      { label: '&Decrease Font Size', command: 'window:decrease-font-size' }
+      { label: '&Reset Font Size', command: 'window:reset-font-size' }
```
izuzak: Will this cause problems with this command in the same menu:

```
       { label: '&Reload', command: 'window:reload' }
```

We would now have two menu options starting with `&R...`

izuzak:
@mnquintana I'm :cool: with that. Thanks :+1:

menus/linux.cson
mnquintana: Oh damn! Good catch, I didn't see that one. Fixing...

PR #6539, opened 2015-04-27
Reset font size to window's starting font size

mnquintana:
This changes `workspace::resetFontSize`'s behavior from restoring Atom's default font size to restoring the window's starting font size. 

This is, I think, far more useful behavior than restoring Atom's default font size, which you can do by just emptying the Font Size input in Settings View. If you've ever had to change the font size to make code more readable to others temporarily (ie. a presentation or a talk), having a command to return to what you had configured originally is a godsend. (I've found myself in that situation quite often.)

Please try this branch out and let me know what you think! :wink:

/cc @atom/feedback

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -425,6 +425,19 @@ describe "Workspace", ->
       expect(atom.config.get('editor.fontSize')).toBe 1
       workspace.decreaseFontSize()
       expect(atom.config.get('editor.fontSize')).toBe 1
+  describe "::resetFontSize()", ->
+    it "resets the font size to the window's starting font size", ->
+      originalFontSize = 6
+
+      atom.config.set('editor.fontSize', originalFontSize)
+      workspace.increaseFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize + 1
+      workspace.resetFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize
+      workspace.decreaseFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize - 1
+      workspace.resetFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize
 
   describe "::openLicense()", ->
     it "opens the license as plain-text in a buffer", ->
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -425,6 +425,7 @@ describe "Workspace", ->
       expect(atom.config.get('editor.fontSize')).toBe 1
       workspace.decreaseFontSize()
       expect(atom.config.get('editor.fontSize')).toBe 1
+
   describe "::resetFontSize()", ->
     it "resets the font size to the window's starting font size", ->
       originalFontSize = 6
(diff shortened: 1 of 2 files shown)


50Wliu:
I'd enjoy this.

src/workspace.coffee
```diff
@@ -612,16 +612,18 @@ class Workspace extends Model
...
   decreaseFontSize: ->
+    @originalFontSize ?= atom.config.get("editor.fontSize")
     fontSize = atom.config.get("editor.fontSize")
     atom.config.set("editor.fontSize", fontSize - 1) if fontSize > 1
 
-  # Restore to a default editor font size.
+  # Restore to the window's original editor font size.
   resetFontSize: ->
-    atom.config.unset("editor.fontSize")
+    atom.config.set("editor.fontSize", @originalFontSize)
```
lee-dohm: Are you sure this will work the same as `atom.config.unset` in the case where `@originalFontSize` is `undefined`? Let's say my normal font size is 28. If I *do not* increase or decrease the font size and accidentally hit restore, will it set me to the Atom default font size? Or will it do something freaky?

There should also be a spec for this, that reset with no increase or decrease first should leave the font size where it is.

mnquintana: Good catch! Should be fixed now.

lee-dohm:
:+1:

src/workspace.coffee
```diff
@@ -612,16 +612,19 @@ class Workspace extends Model
 
   # Increase the editor font size by 1px.
   increaseFontSize: ->
+    @originalFontSize ?= atom.config.get("editor.fontSize")
```
kevinsawicki: Not sure if this is worth changing things for, but this won't catch the case where someone calls `atom.config.set('editor.fontSize', 10)` directly after startup via a package or init script. In that case the "originalFontSize" would be stored as 10 even though it was a different value "originally".

kevinsawicki: Thinking about this more, if I change the value from the settings view and then run the reset command, it won't reset it right?

mnquintana: > Thinking about this more, if I change the value from the settings view and then run the reset command, it won't reset it right?

That's true! I'll fix that later today.

mnquintana: I think what might do here is emit a `did-change-font-size` event here and in settings-view.

mnquintana: Fixed! Reset will now work if `editor.fontSize` is changed from anywhere.

braver:
:+1:

spec/workspace-spec.coffee
```diff
@@ -426,6 +426,37 @@ describe "Workspace", ->
       workspace.decreaseFontSize()
       expect(atom.config.get('editor.fontSize')).toBe 1
 
+  describe "::resetFontSize()", ->
+    it "resets the font size to the window's starting font size", ->
+      originalFontSize = atom.config.get('editor.fontSize')
+
+      atom.config.set('editor.fontSize', originalFontSize)
```
kevinsawicki: I don't think this line is needed since the value is read on the previous line and the value isn't changed to `atom.config.set` will ignore it I think since the value hasn't changed.

spec/workspace-spec.coffee
```diff
@@ -426,6 +426,37 @@ describe "Workspace", ->
...
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize
+      workspace.decreaseFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize - 1
+      workspace.resetFontSize()
+      expect(atom.config.get('editor.fontSize')).toBe originalFontSize
+
+    it "does nothing if the font size has not been changed", ->
+      originalFontSize = atom.config.get('editor.fontSize')
+
+      atom.config.set('editor.fontSize', originalFontSize)
```
kevinsawicki: Same here, I think this line can go :fire:

mnquintana: Both lines have been :fire:ed

kevinsawicki:
Awesome, thanks!

PR #6543, opened 2015-04-28
Add --profile-startup option

kevinsawicki:
This pull request adds the `--profile-startup` command line option that will profile render process for the created window.

The goal here is to make it easy to see why an Atom window might be loading slowly.

It covers the time between the `window.onload` handler firing and the completion of the bootstrap script. So it will include deserialization, all requires, package loading, theme loading, etc.

``` sh
atom --profile-startup
```

screen shot 2015-04-27 at 6 17 00 pm

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -32,44 +32,56 @@ window.onload = function() {
 
     var devMode = loadSettings.devMode || !loadSettings.resourcePath.startsWith(process.resourcesPath + path.sep);
 
-    setupCoffeeCache(cacheDir);
-
-    ModuleCache = require('../src/module-cache');
-    ModuleCache.register(loadSettings);
-    ModuleCache.add(loadSettings.resourcePath);
-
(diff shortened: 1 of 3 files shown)


static/index.js
```diff
@@ -121,3 +133,21 @@ var setupVmCompatibility = function() {
...
+      console.profile('startup');
+      try {
+        setupWindow(cacheDir, loadSettings);
+      } catch (error) {
+        handleSetupError(error);
+      } finally {
+        console.profileEnd('startup');
+        console.log("Switch to the Profiles tab to view the startup profile")
+      }
+    }, 100);
```
kevinsawicki: @zcbenz this timeout seemed to be required. Without it the profile didn't actually get created. Do you think this is an okay workaround?

lee-dohm:
:clap: This sounds really cool!

izuzak:
:clap: :clap: Yesss, this will be super helpful!

steveoh:
is this still implemented?

PR #6547, opened 2015-04-28
add missing keybindings to show specific item on windows

Cynede:
related issue on discuss: https://discuss.atom.io/t/ctrl-x-tab-selection/13615

I strongly suggest to keep `alt` there and not `ctrl`
- I just can't rotate my hand to press `ctrl+1` or alike, I don't know how people use it, it's very uncomfortable
- sumblime has got `alt+n` even on windows


lee-dohm:
I think using <kbd>Alt</kbd> as the modifier key there is fine.

Heather:
@lee-dohm great, I've tested it on user-config file.

kevinsawicki:
Awesome, thanks for adding these :+1:

PR #6563, opened 2015-04-29
Fix scroll width calculation when longest line is folded

### LINKED ISSUE #6562: Can't scroll right when unfolding causes content to be wider than the editor's scroll view
If the longest line is folded, we clip the screen position at the end of the line to 0 before translating it to a pixel position. This is due to the fact that we clip cursors to the beginning of folded lines, a decision I'd actually like to revisit in its own right.

nathansobo:
With the presenter we started clipping screen positions prior to translating them to pixel positions. This interacts with the current clipping behavior on folded lines (which should change) where the cursor is always clipped to 0. So when the longest line was also folded we were miscalculating the width. :see_no_evil:!

The removal of clipping also causes us to calculate the width based on the trailing whitespace of soft-wrapped lines, which I actually think is an improvement but it is slightly different.

Fixes #6562

### DIFF
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -498,6 +498,11 @@ describe "TextEditorPresenter", ->
 
           expect(presenter.getState().content.scrollWidth).toBe 10 * editor.getMaxScreenLineLength() + 1
 
+        it "isn't clipped to 0 when the longest line is folded (regression)", ->
+          presenter = buildPresenter(contentFrameWidth: 50, baseCharacterWidth: 10)
+          editor.foldBufferRow(0)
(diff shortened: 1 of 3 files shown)


nathansobo:
This has an issue where we scroll horizontally a tiny bit when the text is soft wrapped because we're not clipping back to the end of the actual content. May need to explore a better path.

nathansobo:
Forced push a change that preserves the original clipping behavior on soft-wrapped lines. Seems to cover both cases well. I tried disabling clipping of folded lines but there are too many implications to take that on at this moment. After the document model work is finished we can revisit it.

maxbrunsfeld:
Didn't get time to try this branch before 0.195 was released. Trying it now, with no problems so far.

PR #6566, opened 2015-04-29
Refactor TextEditorPresenter batching

as-cii:
As we kept adding features or fixing stuff, our `TextEditorPresenter` has grown more and more complex (as of now it counts more than 1k LOC). As you might imagine, this leads to difficult maintenance and introducing regressions becomes much easier. :fearful:

That said, this layer is not intrinsically convoluted and, therefore, I think we have some chance to make it better. This PR attempts to bring some clarity to one aspect @nathansobo pointed during a conversation and which I totally agree with.

When dealing with batching, I introduced an helper function (`@batch`) which enabled me to, more or less, decorate a function and have it batched automagically. While this removed a bit of duplication, it didn't improve code quality at all, making the batching process look a bit opaque.

The intent of this PR is to replace this mechanism with a simpler one, which solely relies on booleans. I have payed attention to preserve the original behavior but, as illustrated earlier on, it's easy to overlook mistakes and it would be great if you could try this out to spot regressions of any kind. (/cc: @atom/feedback)

Going forward I think that the path should be to split `TextEditorPresenter` into many smaller objects (e.g. one for lines, one for gutters, etc... maybe the structure could even match how we currently organize _UI Components_). This way `TextEditorPresenter` could act as the glue which makes those smaller presenters stick together. If we follow this design I think we can make our test suite even more precise (i.e. keep some integration tests in `TextEditorPresenter` and test each component separately). @nathansobo: what are your :thought_balloon: on this?

My plan is to achieve this incrementally :footprints:: this PR is just the first, tiny step; as we touch related areas of this code we should take the time to give them some :heart:.

Any suggestion, proposal or feedback is very welcome! :bow:


lee-dohm:
So it looks like this is a transitory refactoring on the way to something like this?

```coffee
class Presenter
  constructor: (@textEditorPresenter) ->
    @shouldUpdate = false

  getShouldUpdate: ->
    @shouldUpdate

  setShouldUpdate: ->
    @shouldUpdate = true

  update: (state) ->
    return unless @shouldUpdate
    @performUpdate(state)
    state

# One of many that can be plugged in to `TextEditorPresenter` in the proper order
class UpdateHeightPresenter extends Presenter
  constructor: (presenter) ->
    super(presenter)

  performUpdate: (state) ->
    if @textEditorPresenter.autoHeight
      state.height = @textEditorPresenter.contentHeight
    else
      state.height = null
```

I'm of course assuming that the order in which updates are performed is deterministic. It's just that certain actions are performed or not performed depending on what collection of changes were batched?

I just want to make sure I'm following along :grinning:

src/text-editor-presenter.coffee
```diff
@@ -94,40 +82,68 @@ class TextEditorPresenter
...
-      @updateHorizontalScrollState()
-      @updateScrollbarsState()
-      @updateContentState()
-      @updateDecorations()
-      @updateLinesState()
-      @updateLineNumberGutterState()
-      @updateLineNumbersState()
-      @updateGutterOrderState()
-      @updateCustomGutterDecorationState()
+      @shouldUpdateHeightState = true
```
benogle: :+1: much less confusing

as-cii: Totally. Credits to @nathansobo :dragon: for suggesting this refactoring! :sparkles: 

On a side note, I have noticed we handle *batched operations* in more than one object. Lately I have been thinking about some abstraction which could handle batching for us through a more descriptive API:

```coffee
# Record actions to execute later
foo.defer(@updateHeightState) # or foo.push(@updateHeightState)
foo.defer(@updateWidthState)
foo.defer(@updateBar)

# Apply recorded operations just-in-time (each "operation type" should be executed once and only once, or we could have several strategies to handle that if we need to)
foo.apply()
```

Where `foo` could be `LazyWorkQueue`, or something similar.

Anyway, booleans already express the intent much better than the previous implementation, so I am also fine to keep this stuff as it is for the time being :+1:

benogle: I like the booleans for now as they are super straight forward. I do like the idea of the `LazyWorkerQueue`, but someone coming in would need to understand how it worked to grok the presenter code.

as-cii:
> I'm of course assuming that the order in which updates are performed is deterministic. It's just that certain actions are performed or not performed depending on what collection of changes were batched?

@lee-dohm: correct. Check out `@getState` to understand how it works, but basically we have a bunch of function calls which we execute based on the associated boolean.

> So it looks like this is a transitory refactoring on the way to something like this?

Yep, that's quite close. :+1: However, I was thinking that each presenter should be responsible of knowing when to update, whereas your design seems to suggest that an external object would do. In my mind, `TextEditorPresenter` would simply instantiate these smaller objects and call `getState` on each of them right before returning the state to `TextEditorComponent`. Please, tell me if I am missing something there.

lee-dohm:
> However, I was thinking that each presenter should be responsible of knowing when to update,

Yes, that totally makes more sense.

> instantiate these smaller objects and call getState on each of them right before returning the state to TextEditorComponent. Please, tell me if I am missing something there.

No, I don't see any fault with that design. I think it is a *huge* boon for testability too.

Thanks for taking the time to catch me up! :grinning:

src/text-editor-presenter.coffee
nathansobo: @as-cii What are some of the other batched cases you're thinking of?

I tend to agree with @benogle that it's hard to beat the straightforwardness of a list of subroutines that get called conditionally. I worry that it might be another case of an abstraction that dries things up but also adds some opacity.

If we're going to get more dynamic, there's lots more we could do. For example, could we somehow express the dependencies between various pieces of state to derive the list of what needs to update automatically, like a spreadsheet. That would make things less error prone, but also less explicit. I'm not sure.

src/text-editor-presenter.coffee
```diff
@@ -94,40 +82,68 @@ class TextEditorPresenter
...
     @disposables.add @model.onDidChangeMini =>
+      @shouldUpdateScrollbarsState = true
+      @shouldUpdateContentState = true
+      @shouldUpdateDecorations = true
+      @shouldUpdateLinesState = true
+      @shouldUpdateLineNumberGutterState = true
+      @shouldUpdateLineNumbersState = true
+      @shouldUpdateGutterOrderState = true
+      @shouldUpdateCustomGutterDecorationState = true
       @updateScrollbarDimensions()
```
nathansobo: Why is this one running synchronously?

src/text-editor-presenter.coffee
```diff
@@ -153,24 +169,32 @@ class TextEditorPresenter
...
+      @emitDidUpdateState()
 
   didChangeGrammar: ->
     @observeConfig()
-    @updateContentState()
-    @updateLineNumberGutterState()
+    @shouldUpdateContentState = true
+    @shouldUpdateLineNumberGutterState = true
+    @shouldUpdateGutterOrderState = true
     @updateCommonGutterState()
```
nathansobo: Same here. Everything else is deferred, why not this?

src/text-editor-presenter.coffee
```diff
@@ -94,40 +82,68 @@ class TextEditorPresenter
...
+    @shouldUpdateOverlaysState = false
+    @shouldUpdateLineNumberGutterState = false
+    @shouldUpdateLineNumbersState = false
+    @shouldUpdateGutterOrderState = false
+    @shouldUpdateCustomGutterDecorationState = false
+
   observeModel: ->
     @disposables.add @model.onDidChange =>
       @updateContentDimensions()
       @updateEndRow()
```
nathansobo: Why not async?

src/text-editor-presenter.coffee
as-cii: > What are some of the other batched cases you're thinking of?

@nathansobo: `DisplayBuffer` and `TextEditorPresenter`, for example, defer some operations whenever character widths are being changed (https://github.com/atom/atom/blob/master/src/display-buffer.coffee#L307-L312).

> I tend to agree with @benogle that it's hard to beat the straightforwardness of a list of subroutines that get called conditionally. I worry that it might be another case of an abstraction that dries things up but also adds some opacity.

> If we're going to get more dynamic, there's lots more we could do. For example, could we somehow express the dependencies between various pieces of state to derive the list of what needs to update automatically, like a spreadsheet. That would make things less error prone, but also less explicit. I'm not sure.

Agreed to keep things simple right now: booleans clearly express the intent of deferring work and I think we should favor explicitness over duplication removal if it causes code to be less clear.

src/text-editor-presenter.coffee
as-cii: Some updates are not deferred until the last possible moment: when a decoration marker changes, for example, we need to add/remove it to/from the decorations cache, e.g.

```coffee
if oldRange.intersectsRowRange(@startRow, @endRow - 1)
  @removeFromLineDecorationCaches(decoration, oldRange)
  intersectsVisibleRowRange = true

if newRange.intersectsRowRange(@startRow, @endRow - 1)
  @addToLineDecorationCaches(decoration, newRange)
  intersectsVisibleRowRange = true
```

I agree that, eventually, we should make (almost) any presenter operation deferred. However, my intent with this PR was to apply behavior-preserving changes just to make the code easier to read. Batching these operations would make the refactoring more invasive and I would feel more confident if we do so when we will have smaller presenters (which will, hopefully, be tested independently).

What do you think?

nathansobo: Sounds reasonable to just preserve the existing structure. Seems like this is ready to merge when you're confident it's working correctly from a testing standpoint.

as-cii:
This has been up for a while and I couldn't spot any evident regression or performance degradation since I opened the pull request. CI is :green_heart: too, therefore I would say to :ship: this!

as-cii:
Thank you guys for the always amazing and precious help :sparkles:

Issue #4201, opened 2014-11-16
Consider bundling default font


wutong:
wish u addon Noto fonts to editor and support laravel syntax hints and highlights

mnquintana:
I updated the title to better reflect what it seemed like you were asking for @wutong – does it look alright to you?

I think shipping with a default font is probably a good idea, especially for Linux, since some distros might not have great default monospace fonts. Especially since @simurai has started using it for some screenshots (and it looks awesome), I'd be a big fan of bundling [Fira Mono](http://mozilla.github.io/Fira/). It's also an [open source font](https://github.com/mozilla/Fira), so we should be good for licensing (as long as we include the license).

/cc @atom/non-github-maintainers

lee-dohm:
:+1: I think this would just save us a lot of headache as far as fonts on Linux go.

paulcbetts:
Droid Sans Mono also seems to look (generally) pretty good on all three operating systems

mnquintana:
@paulcbetts Droid Sans Mono seems to only have one weight though :confused:

paulcbetts:
@mnquintana Legit

simurai:
Somehow I just liked the unique `@` in Fira Mono.

There is the [Fonts](https://atom.io/packages/fonts) package if you wanna try out some fonts. Here my 3 favs:

- Fira Mono -> Used in Firefox OS.
- DejaVu Mono
- Source Code Pro -> Used in Brackets.

![fonts](https://cloud.githubusercontent.com/assets/378023/7362401/35817cb4-eda3-11e4-802e-91e2b5548532.gif)

mark-hahn:
+1 for Source Code Pro.

lee-dohm:
Actually, if people were looking for a comprehensive list, [this one](http://programmingfonts.org/list) was pointed out in [this topic on Discuss](https://discuss.atom.io/t/hebrew-not-showing-in-0-190-on-ubuntu/16035).

And something to keep in mind. If we package a "default font", we might want to be sure that it doesn't override the font that people already have set. We don't want to [surprise people negatively](https://discuss.atom.io/t/you-changed-my-theme-when-i-wasnt-looking/15731).

simurai:
> If we package a "default font", we might want to be sure that it doesn't override the font that people already have set.

Right, the font-family set in the settings should still have priority over the bundled one.

simurai:
> If we package a "default font", we might want to be sure that it doesn't override the font that people already have set.

[...]

PR #6567, opened 2015-04-29
Improve font stack

simurai:
This PR adds `DejaVu Sans Mono` to the Atom font stack. It should fix font issues on Linux. See #4201

The default fonts will be:
- `Inconsolata` on all platforms, but must be manually installed by the user. Otherwise..
- `Monaco` on OS X
- `Consolas` on Windows
- `DejaVu Sans Mono` on most Linux distros. Thx @fvsch.
#### Unpleasant surprises

This change should only affect users on Linux that have not defined a custom font in the settings. But arguably that will be a pleasant surprise.

### DIFF
--- a/static/text-editor-light.less
+++ b/static/text-editor-light.less
@@ -4,7 +4,7 @@
 
 atom-text-editor {
   display: block;
-  font-family: Inconsolata, Monaco, Consolas, 'DejaVu Sans Mono', 'Courier New', Courier;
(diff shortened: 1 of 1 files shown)


simurai:
:+1: Especially since `monospace` seems to be mapped to `Courier New` and `Courier` anyways. -> http://fontfamily.io/monospace

lee-dohm:
:+1: Looks good to me. Great solution :grinning:

kevinsawicki:
:+1:

PR #6570, opened 2015-04-29
[Gutter] Clear custom gutter decorations from previous usages of a custo...

jssln:
...m gutter element

This is @nathansobo's fix for line numbers (https://github.com/atom/atom/commit/192997c8cf4479c4cae155207ad3b175c0ee9498) applied to custom gutters.


kevinsawicki:
Thanks for the fix :rocket:

PR #6086, opened 2015-03-25
Added delete-to-next/previous-word-boundary

ghost:
This is better behavior for alt-del/backspace operations, is less "greedy" and doesn't eat up excess characters.
Also set default key mappings to these, feel free to trash my changes to keymaps/darwin.cson. Though I do feel this should be default behavior, personally.

PR #6575, opened 2015-04-30
Fixes for PR 6086

lee-dohm:
See #6086 

@CaptSaltyJack, I went ahead and fixed the tests. I did more research and found that the behavior wasn't the bug it was just an obscure flaw in the tests. They're all fixed now :grinning:


src/selection.coffee
```diff
@@ -404,6 +404,28 @@ class Selection extends Model
     @selectLeft() if @isEmpty() and not @editor.isFoldedAtScreenRow(@cursor.getScreenRow())
     @deleteSelectedText()
 
+  # Deprecated: Use {::deleteToBeginningOfWord} instead.
+  backspaceToBeginningOfWord: ->
```
kevinsawicki: Sorry if I missed this in the other pull request, buy why are these being added as deprecated APIs?

kevinsawicki: Looks like this might have been a merge gone wrong, these are at the bottom now in the deprecated block so I think they can just be removed completely from this diff.

lee-dohm: I believe because they're identical to their `deleteTo*` counterparts and to make the naming consistent? I can take it out though.

kevinsawicki: They are located on master here: https://github.com/atom/atom/blob/15b4cefcf4ee7ec37bda0da08bc0dd391581e2c6/src/selection.coffee#L783-L791

src/text-editor.coffee
```diff
@@ -1057,6 +1057,14 @@ class TextEditor extends Model
   deleteToBeginningOfWord: ->
     @mutateSelectedText (selection) -> selection.deleteToBeginningOfWord()
 
+  # Similar to above, but deletes back to the previous word boundary
```
kevinsawicki: Since the methods on `Selection` are public, then I'd say mark these as `Extended` for consistency.

kevinsawicki:
Looks great, :ship:

lee-dohm:
Thanks @CaptSaltyJack for getting this going! :grinning:

CaptSaltyJack:
@lee-dohm Thanks!! Sorry it took so much hand-holding. ;)

joostverdoorn:
Thanks! I was just looking for this :).

PR #6582, opened 2015-04-30
Normalize short urls for repositories in package.json

### LINKED ISSUE #387: Supports short repository url
NPM supports 
```js
  "repository": "atom-minimap/minimap",
```

But in Atom, that was broken.
screen shot 2015-02-20 at 4 03 43 pm
screen shot 2015-02-20 at 4 03 51 pm

jlord:
This adds a dependency and replicates the way npm handles 'shortcut' repository urls in package.json, turning them into urls. 

Closes atom/settings-view#387

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "fuzzaldrin": "^2.1",
     "git-utils": "^3.0.0",
     "grim": "1.4.0",
+    "hosted-git-info": "^2.1.2",
     "jasmine-json": "~0.0",
     "jasmine-tagged": "^1.1.4",
     "jquery": "^2.1.1",
--- a/spec/fixtures/packages/package-with-short-url-package-json/package.json
+++ b/spec/fixtures/packages/package-with-short-url-package-json/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "package-with-short-url-package-json",
+  "repository": "example/repo"
+}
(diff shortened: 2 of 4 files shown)


jlord:
:balloon:

sindresorhus:
Why not just use [`read-package-json`](https://github.com/npm/read-package-json) as suggested in https://github.com/atom/settings-view/issues/387#issuecomment-90846012 ? It will handle all package.json normalizations, not just this one. There's lots of more. You don't want to handle them all manually.

kevinsawicki:
> Why not just use read-package-json as suggested in atom/settings-view#387 (comment) ?

One reason was that that library is quite large and takes 35-40ms to require which would negatively impact startup.

It also corrected certain typos in package keys which I think is overreaching and confusing to the package author on which one is right since both normalize to the same value.

sindresorhus:
> One reason was that that library is quite large and takes 35-40ms to require which would negatively impact startup.

https://github.com/npm/normalize-package-data might be faster. If there's a perf problem you could always try to fix it at the source.

> It also corrected certain typos in package keys which I think is overreaching and confusing to the package author on which one is right since both normalize to the same value.

I would argue that you're using package.json, hence you should support what it does, even it's idiosyncrasies. Do you have an example?

kevinsawicki:
>  Do you have an example?

The `bundleDependencies` example listed here: https://github.com/npm/normalize-package-data#what-normalization-currently-entails

> Seems like a lot of extra work to handle all these normalizations manually and basically duplicating the work

Yeah, since package metadata normalization has never come up before this short repo url issue it seemed like we could just be surgical about that specific case and use the one library needed to support it.

@jlord do you want to give `normalize-package-data` a spin and see how it goes?

Looking through the other normalization things in the README list, it seems like normalization of authors and keywords might be helpful too to have handled in core.

> Just trying to help here

Much appreciated :+1:

jlord:
> @jlord do you want to give normalize-package-data a spin and see how it goes?

Yep :+1: Opened an issue for it :arrow_heading_up:

PR #6602, opened 2015-05-01
Remove Completions from Settings

### LINKED ISSUE #353: What is "Completions"? Need info/help caption
I find this "Completions" box to be fairly vague/ambiguous. What kind of completions? What's an example of what this box expects? I have no idea what this does. I'd recommend an explanatory caption.

settings-completions

jlord:
We don't really need these here (and it can be confusing) as languages can define them. 

_before_

Closes https://github.com/atom/settings-view/issues/353

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -98,8 +98,6 @@ module.exports =
         type: ['string', 'null']
 
       # These can be used as globals or scoped, thus defaults.
-      completions:
-        type: ['array', 'object']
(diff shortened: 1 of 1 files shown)


benogle:
Does it disappear if you just take out the default? Or do you need to remove the entire entry?

jlord:
> Does it disappear if you just take out the default? 

Yep :+1: So I've just added back the type.

PR #6631, opened 2015-05-04
Add commands to move selections left and right by one column

mnquintana:
_See https://discuss.atom.io/t/shift-selection-by-column/16545?u=mnquintana for more background_

screencap1

This adds the column-equivalent of `TextEditor::moveLineUp()` and `TextEditor::moveLineDown()` - instead of moving a line up or down, these commands move any active selections left or right by one column. This is especially useful if you're refactoring / generally trying to clean up your code.

TODO:
- [x] Add keybindings for Linux and Windows (anyone have any suggestions for good keybindings for this?)
- [x] Consider renaming commands? (Is Move Selection Left / Right too vague?)
- [x] Fix case when moving multiple one-line selections to either the beginning or end of the line

Shoutout to @lee-dohm for helping with the implementation :wave: 

/cc @atom/feedback


YuriSolovyov:
>  Add keybindings for Linux and Windows (anyone have any suggestions for good keybindings for this?)

Not sure, but `Alt + (Left) Shift + Left | Right` Arrows seem to do nothing on Ubuntu ATM

thomasjo:
> Not sure about Windows, but Alt + (Left) Shift + Left | Right Arrows seem to do nothing on Ubuntu ATM

If you read the TODO list, you'll see this

> Add keybindings for Linux and Windows (anyone have any suggestions for good keybindings for this?)

:see_no_evil:

YuriSolovyov:
@thomasjo I was just _suggesting_ these keys. "seem to do nothing" -> can be taken for this PR's functionality.

thomasjo:
@YuriSolovyov Aaah, I misunderstood what you were saying; so you were suggesting those as possible keybinds. Sorry about the confusion! And yeah, those seems like sensible suggestions :+1:

lee-dohm:
Things look good to me! :+1:

src/text-editor.coffee
```diff
@@ -1074,6 +1074,50 @@ class TextEditor extends Model
...
+    selections = @getSelectedBufferRanges()
+    noSelectionAtStartOfLine = selections.every((selection) ->
+      selection.start.column isnt 0
+    )
+
+    translationDelta = [0, -1]
+    translatedRanges = []
+
+    @transact =>
+      if noSelectionAtStartOfLine
```
mnquintana: Would it make more sense for this conditional to encompass the `transact`ion too?

mnquintana:
This should be ready for review again! (I'm happy I managed to get back to this one just under the year mark... :sweat_smile:)

Right now I'm debating switching this over to screen ranges to keep behavior consistent if soft wrap's enabled, but I'm not sure right now which would be more intuitive. :thought_balloon:s?

Otherwise, @atom/feedback plz try this out and let me know if you catch anything weird! :bow:

src/text-editor.coffee
lee-dohm: Yeah, I think if the transaction is a no-op when the selection is at the start/end of line, then we shouldn't create the transaction.

lee-dohm:
In testing it out, I'm getting the same error beep behavior that we've seen when using <kbd>Cmd+Ctrl+Down</kbd> for each left/right movement on OS X. I don't think that's an issue with the PR, just something I noticed. Also, it appears that <kbd>Cmd+Ctrl+Up/Down</kbd> move the entire line, rather than just the selection. Again, not something specifically to block this, but to be aware of.

mnquintana:
> In testing it out, I'm getting the same error beep behavior that we've seen when using <kbd>Cmd+Ctrl+Down</kbd> for each left/right movement on OS X.

Yup :disappointed: that one's blocked on https://github.com/atom/electron/issues/2617 right now.

> Also, it appears that <kbd>Cmd+Ctrl+Up/Down</kbd> move the entire line, rather than just the selection. Again, not something specifically to block this, but to be aware of.

That's true – that was something I considered early on with this. Like, the literal counterpart of Move Line Up / Down would be to shift an entire column left or right (since the former is just shifting an entire row up or down), but that wouldn't be very useful. This is more the functional counterpart to Move Line Up / Down.

:ship:?

lee-dohm:
👍 I think this is good to go, yep :rocket:

legovini:
It seems to me that on Linux alt+shift+(left|right) conflict with these keybindings:

```
    "alt-shift-left": "editor:select-to-previous-subword-boundary",
    "alt-shift-right": "editor:select-to-next-subword-boundary",
```

See `linux.json:99` and `:100`.

marat93:
same for me. `alt-shift-right` just selects next word. Ubuntu 14.04, atom 1.8.0

davesmiths:
Thank you so much for this, love it! 
on OSX CMD+CTRL+LEFT/RIGHT

AlexWayfer:
image
Looks like too many key-bindings, and works in fact only one (not clear which)

mnquintana:
Ah well that's a bummer – that's what happens when you merge a PR 1 year later I guess 😞 (this definitely didn't conflict at the time!)

I'll look into a new keybinding for this. Thanks everyone for bringing it to my attention! 😄

Dayjo:
It was but 2 weeks ago when I thought to myself "I wish I could move selections left and right as well as up and down".. and today this! 

Good luck getting the binding sorted, and thanks! :)

not-a-hacker:
I was disappointed to find that the default key bindings for move-selection-left and move-selection-right don't work in Linuxmint 17.3 Cinnamon 64 bit on my MSI laptop. I was able to get them working with the windows-alt-left and windows-alt-right key combos by putting the following in my keymap file, keymap.cson:

```
'atom-text-editor':
    'cmd-alt-left': 'editor:move-selection-left'
    'cmd-alt-right': 'editor:move-selection-right'

```
So does anyone know if atom will always recognize the Linux super key when you define a key binding with 'cmd?' On Windows boxes super is the Windows key, but what about on PCs built to run Linux like those from System76, zareason, Entroware, etc?

defking:
Rigth now on windows i'm using:
`'atom-text-editor':
  'ctrl-alt-pageup': 'editor:move-selection-left'
  'ctrl-alt-pagedown': 'editor:move-selection-right'`
maybe those are not the most intuitive, but they work and don't conflict with anthing.

By the way, @not-a-hacker cmd-alt-* combinations don't work on windows.

ForeverWinter:
For shortcuts: What about Ctrl+Shift+, for shift left and Ctrl+Shift+. for shift right (i.e. looking at the < and > on the keys for intuitiveness)

AlexWayfer:
@ForeverWinter So we need to replace exist shortcut:

image

ForeverWinter:
Ah yes, I missed that.  Could make "scroll-to-cursor" ctrl-? (ctrl-shift-/).  Alternatively, could make the shift left and right Alt-< and Alt-> (alt-shift-, and alt-shift-.) instead.

AlexWayfer:
`Alt->` is occupied.

image

ForeverWinter:
> Alt-> is occupied.
> 
>` alt->    linter:next-error`

Yeah, but not by any core packages.  Granted, Linter is one of the most common community packages; it should probably be taken under consideration given its popularity, but I didn't have it installed when checking for shortcut conflicts.

ZdrytchX:
Grouping Does not work well with python because python does not have brackets 😢 

Indenting should use the system set by the user (in my case for example that is 2 spaces, not a tab-whitespace)

PR #6640, opened 2015-05-04
Allow 0 to be passed as initialLine when opening files via command line

### LINKED ISSUE #1743: Problems with opening a file at a specific line
Running `atom file.coffee:20` will open something.coffee at line 20. But if the editor already has file.coffee open, it will not scroll to line 20. This is causing problems for people wanting to create links from stacktraces.

/via http://discuss.atom.io/t/open-file-at-specific-line-and-column/3883/6

izuzak:
Fixes https://github.com/atom/atom/issues/1743 which was reopened due to this observation:

> This doesn't quite work yet. When opening a file to line 0, Atom will just try to open a file called "filename.txt:0" rather than just defaulting to the top of the file.

I think it's worth supporting this edge case where we'd normally expect the user to specify "1" as the initial line, not "0".

cc @kevinsawicki

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -518,11 +518,11 @@ class AtomApplication
 
     [fileToOpen, initialLine, initialColumn] = path.basename(pathToOpen).split(':')
     return {pathToOpen} unless initialLine
-    return {pathToOpen} unless parseInt(initialLine) > 0
+    return {pathToOpen} unless parseInt(initialLine) >= 0
 
     # Convert line numbers to a base of 0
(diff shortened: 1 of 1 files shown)


lee-dohm:
Looks good to me! :+1:

50Wliu:
Does this also fix #6654?

lee-dohm:
@50Wliu it'll fix the case of `filename.txt:0`, but I don't think it will "fix" the `filename.txt:` issue that is also mentioned.

kevinsawicki:
:+1:

PR #6648, opened 2015-05-04
Add TextEditorElement::isCursorOnScreen()

mnquintana:
This adds the method `::isCursorOnScreen()` to `TextEditorElement`, which makes it a lot easier to determine whether or not a cursor is visible on screen without having to do the rows comparison yourself. This is especially handy for testing, but could certainly have other uses for package authors.

/cc @atom/feedback @nathansobo

### DIFF
--- a/spec/text-editor-element-spec.coffee
+++ b/spec/text-editor-element-spec.coffee
@@ -234,7 +234,7 @@ describe "TextEditorElement", ->
       jasmine.attachToDOM(element)
       expect(element.getDefaultCharacterWidth()).toBeGreaterThan(0)
 
-  fdescribe "::isCursorOnScreen", ->
+  describe "::isCursorOnScreen", ->
     [element, model] = []
 
     beforeEach ->
--- a/spec/text-editor-element-spec.coffee
+++ b/spec/text-editor-element-spec.coffee
@@ -233,3 +233,23 @@ describe "TextEditorElement", ->
       element = new TextEditorElement
       jasmine.attachToDOM(element)
       expect(element.getDefaultCharacterWidth()).toBeGreaterThan(0)
+
+  fdescribe "::isCursorOnScreen", ->
+    [element, model] = []
+
+    beforeEach ->
+      atom.project.setPaths([atom.project.getDirectories()[0]?.resolve('.')])
+
+      waitsForPromise ->
+        atom.workspace.open('two-hundred.txt').then (o) -> model = o
+
+    it "returns true if the cursor is visible on the screen", ->
+      element = atom.views.getView(model)
+      expect(model.getCursorScreenPosition()).toEqual [0, 0]
+      expect(element.isCursorOnScreen()).toBe true
+
+    it "returns false if the cursor is not visible on the screen", ->
+      element = atom.views.getView(model)
+      expect(model.getCursorBufferPosition()).toEqual [0, 0]
+      model.setCursorBufferPosition([element.getLastVisibleScreenRow() + 1, 0], autoscroll: false)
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
In general, we're going for a pattern where our model objects (`TextEditor`, `Cursor`, etc) don't depend on our view objects (`TextEditorElement` etc). The view objects depend on the models, but not the other way around. Because this method requires using the `TextEditorElement`, I'd rather the method not be on `Cursor`. Maybe it could be on `TextEditorElement` instead; something like `TextEditorElement::isCursorOnScreen(cursor)`?

mnquintana:
@maxbrunsfeld Sure, makes sense - just updated it!

as-cii:
Nice work here, @mnquintana! :rainbow: 

On a side note, I wonder if we should state clearly in our documentation that this method doesn't take into account cursors that are off-screen because they are past the viewport width. As far as I understand if, for instance, a cursor is at `[0, 90]` but in the viewport we only see characters from `[0, 0]` to `[0, 80]`, this method will return `true` anyway. Is that correct? 

In order to perform such calculation for the viewport width as well, I think we need to store char left pixel positions for each row: since this is a bit tricky (because of how we measure/store character widths right now) I believe that it's sufficient to describe this behavior in the :memo: for the time being. What do you think?

mnquintana:
> As far as I understand if, for instance, a cursor is at [0, 90] but in the viewport we only see characters from [0, 0] to [0, 80], this method will return true anyway. Is that correct?

Yup that's true. That's a good point though – I originally intended this method to only return true if the cursor's *row* is in the viewport. – I wonder if I should rename it? (e.g. to something like `::isCursorRowOnScreen`?) 

I think having a method that does take into account viewport width would also be really useful though at some point.

nathansobo:
I'm not a huge fan of the highly specific method name. I'd be open to the `isCursorOnScreen`. The question is, is it reasonable to release it half-baked with only vertical support and then add horizontal support later? I'm worried things could be built on the assumption of it only being vertical and then break. Maybe the answer is to get it working in both vertical and horizontal dimensions and then add this.

as-cii:
> Maybe the answer is to get it working in both vertical and horizontal dimensions and then add this.

@nathansobo: agreed :+1: 

@mnquintana: even though we don't have an API which stores characters' absolute positions, you may still be able to calculate it through `pixelPositionForScreenPosition`. That information (along with `scrollLeft` and `scrollWidth`) should be sufficient to understand if the cursor is actually visible.

mnquintana:
@as-cii @nathansobo Should be working horizontally now too!

However, the spec I wrote for horizontal behavior is throwing a `TypeError: Cannot read property 'getTokenIterator' of undefined` on `screenPositionForPixelPosition` – does this have to do with the recent Tokenized Lines changes? Or am I just doing it wrong? :stuck_out_tongue_closed_eyes:

nathansobo:
You may be calling the method on a row that doesn't exist. I believe it would have been an error before, but perhaps a different one. Would be worth putting a breakpoint in that method and seeing what row it's resolving to.

mnquintana:
Actually just kidding, `scrollWidth` does not work the way I thought it did – I'm actually not sure how to get the visible width of an editor here.

nathansobo:
There's a `::clientWidth` property on the presenter you could add a getter for. That refers to the actual width of the editor minus the (possibly visible) vertical scrollbar.

lee-dohm:
@mnquintana Is this still in progress? Do you want to push it forward? (Sorry that it went so long without update :frowning:)

damieng:
No response, closing.

PR #6656, opened 2015-05-05
Remove babel compatibility option for legacy React versions

jgebhardt:
This change depends on https://github.com/atom/react/pull/2.

After updating https://www.npmjs.com/package/react-atom-fork to v0.13.2 per the above PR, we need to remove the compatibility transform.

### DIFF
--- a/src/babel.coffee
+++ b/src/babel.coffee
@@ -32,11 +32,6 @@ defaultOptions =
     # Target a version of the regenerator runtime that
     # supports yield so the transpiled code is cleaner/smaller.
     'asyncToGenerator'
-
-    # Because Atom is currently packaged with a fork of React v0.11,
-    # it makes sense to use the reactCompat transform so the React
-    # JSX transformer produces pre-v0.12 code.
-    'reactCompat'
   ]
 
   # Includes support for es7 features listed at:


bolinfest:
For clarification, let's go ahead and get this in so that users who are using Babel can get their JSX transpiled to code that targets a modern version of React.

Because `require('atom').React` is already on the chopping block, those who were relying on this behavior should move to a modern version of React that they vendor with their own package.

Submitting this commit on its own, without https://github.com/atom/react/pull/2, will only be problematic for people who are doing both of the following:

* Transpiling JSX via Atom's support for `use babel`.
* Using React via `require('atom').React`.

We are committed to publishing an npm package that will make it easy to pull in the latest React in Atom such that it works seamlessly with the updated version of Babel options provided by this PR. The package we create will ensure that there is only one instance of `React` in Atom, which is the special guarantee that `require('atom').React` has historically provided.

bolinfest:
@jgebhardt @thedaniel I think that `babel-spec.coffee` has to be updated as part of this PR.

bolinfest:
Ah, OK, `babel-spec.coffee` does not need to be updated. It tests the hash logic and uses the existing options as a sample input, but it's just a sample (it could be anything), so the test is fine.

kevinsawicki:
:ship: Thanks!

PR #6659, opened 2015-05-05
Use React with addons

jgebhardt:
This makes [React addons](https://facebook.github.io/react/docs/addons.html) available on `atom.React.addons`.

Without this change, developers using `React.addons` either need to pull in a completely different React build that includes addons (bad), or use addons from another react module (also bad, because the module may have diverged from `atom.React`).

### DIFF
--- a/exports/atom.coffee
+++ b/exports/atom.coffee
@@ -122,7 +122,7 @@ unless process.env.ATOM_SHELL_INTERNAL_RUN_AS_NODE
 
     Object.defineProperty module.exports, 'React', get: ->
       deprecate "Please require `react-atom-fork` instead: `React = require 'react-atom-fork'`. Add `\"react-atom-fork\": \"^0.11\"` to your package dependencies."
-      require 'react-atom-fork'
+      require 'react-atom-fork/addons'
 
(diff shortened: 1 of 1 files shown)


thomasjo:
Please note that React bundled with Atom (and as such exposed via `atom.React`) has been deprecated—just look at the line above the one you altered. Therefore I don't think we'll accept any PRs related to React since it will most likely be yanked out soon (v1.0.0).

/cc @kevinsawicki

kevinsawicki:
Yeah, this export will be completely removed in Atom 1.0 and is already not exported by default in the latest version of Atom.

Closing this out, sorry for not replying sooner.

jgebhardt:
No worries. I wanted to add that we've moved to a new fork of react: https://www.npmjs.com/package/react-for-atom

The plan is to have that be the recommended way to require React in any Atom package going forward to avoid conflicts between different version of React, though admittedly, it's hard to enforce.

`react-for-atom` tracks upstream react more closely, lazily loads `React.addons`, and provides a few additional patches that make it more suitable for use in atom (basically applying the modifications from `atom-react-fork` by injection rather than forking).

Issue #6521, opened 2015-04-26
Integration tests failing


brettle:
Are the integration tests just generally broken at the moment or should I expect them to work? 

Here's what I get:
```
[brettle@localhost atom]$ LD_LIBRARY_PATH=/opt/google/chrome ATOM_INTEGRATION_TESTS_ENABLED=true /home/brettle/apps/atom/share/atom/resources/app/apm/bin/apm test
[13254:0425/215459:ERROR:browser_main_loop.cc(170)] Running without the SUID sandbox! See https://code.google.com/p/chromium/wiki/LinuxSUIDSandboxDevelopment for more information on developing with the sandbox on.
ATTENTION: default value of option force_s3tc_enable overridden by environment.
[13304:0425/215459:INFO:renderer_main.cc(212)] Renderer process started
[13254:0425/215501:INFO:CONSOLE(59)] "Window load time: 1707ms", source: file:///home/brettle/git/atom/static/index.js (59)
.........................[13254:0425/215502:INFO:CONSOLE(104)] "Uncaught BufferedProcessError: Failed to spawn command `bad-command-nope`. Make sure `bad-command-nope` is installed and on your PATH", source: /home/brettle/git/atom/src/buffered-process.coffee (104)
.....-.......................................................................[13254:0425/215505:INFO:CONSOLE(889)] "Error setting item in object: Validation failed at null.foo.int, null cannot be coerced into an int", source: /home/brettle/git/atom/src/config.coffee (889)
[13254:0425/215505:INFO:CONSOLE(889)] "Error setting item in object: Validation failed at foo.int, null cannot be coerced into an int", source: /home/brettle/git/atom/src/config.coffee (889)
.......[13254:0425/215505:INFO:CONSOLE(612)] "'foo.baz' could not set the default. Attempted default: undefined; Schema: {"type":"string"}", source: /home/brettle/git/atom/src/config.coffee (612)
[13254:0425/215505:INFO:CONSOLE(612)] "'foo.scoped' could not set the default. Attempted default: undefined; Schema: {"type":"boolean"}", source: /home/brettle/git/atom/src/config.coffee (612)
.[13254:0425/215505:INFO:CONSOLE(612)] "'foo.baz' could not set the default. Attempted default: undefined; Schema: {"type":"string"}", source: /home/brettle/git/atom/src/config.coffee (612)
[...]

PR #6672, opened 2015-05-06
Fix some integration test flakiness

maxbrunsfeld:
Refs #6521 (May fix, but I'm not sure)


maxbrunsfeld:
:ship:ing because the failing specs are unrelated to this.

PR #6673, opened 2015-05-06
Keep the cursor in view when changing font size

### LINKED ISSUE #4084: cmd+ and cmd- should keep you at the same line in the file

mnquintana:
Fixes #4084 

The spec is failing but it really shouldn't be... it looks like it's because `DisplayBuffer::getVisibleRowRange` seems to always return `[0, 0]` when running specs, which `TextEditorElement::getLast/FirstVisibleScreenRow()` makes use of. @nathansobo any idea what might be going on there?

/cc @atom/feedback


nathansobo:
If you attach the workspace to the document with `jasmine.attachToDOM(workspaceElement)` it will be able to measure what "on screen" actually means and return you better values. You may need to adjust its size for it to work.

mnquintana:
Ah, not sure how I missed that - thanks @nathansobo! One more question – how can I adjust the size of the workspace element? I tried setting `clientHeight` et al., but to no avail.

spec/workspace-spec.coffee
```diff
@@ -455,6 +455,33 @@ describe "Workspace", ->
       workspace.resetFontSize()
       expect(atom.config.get('editor.fontSize')).toBe originalFontSize
 
+    fdescribe "when the cursor goes off-screen", ->
```
izuzak: `fdescribe` -> `describe` :microscope:

mnquintana:
@nathansobo I moved the behavior and specs to `TextEditorComponent`, and I think I got the specs working properly now - can you take a look?

nathansobo:
I just realized something... maybe rather than scrolling to the cursor's position, you should instead scroll the editor to the same relative position in the file. You can determine the relative position by dividing the scroll top by the line height. Then after the font size changes, reset the scroll position to the result of the previous division multiplied by the new line height. That way the position will be preserved regardless of whether the cursor is current on or off screen. It could be a bit jarring to have the scroll position completely change if the cursor is off screen at the time the size is changed. Thoughts?

mnquintana:
> I just realized something... maybe rather than scrolling to the cursor's position, you should instead scroll the editor to the same relative position in the file. ... That way the position will be preserved regardless of whether the cursor is current on or off screen. It could be a bit jarring to have the scroll position completely change if the cursor is off screen at the time the size is changed. Thoughts?

This sounds way better to me :+1: I'll update this with the new behavior.

mnquintana:
@nathansobo I ran into an interesting snag with this - it turns out that the `lineHeight` property is `undefined` for inactive editors, which means that using `TextEditor::screenPositionForPixelPosition()` doesn't work the way I've been using it.

An example in `TextEditorComponent`:
```coffeescript
  observeConfig: ->
    @disposables.add atom.config.onDidChange 'editor.fontSize', =>
      @sampleFontStyling()
      @editor.scrollToScreenPosition(@editor.screenPositionForPixelPosition({top: 0, left: 0}))
```

Assuming you have multiple editors open, `@editor.screenPositionForPixelPosition({top: 0, left: 0})` here will return a proper value for the active editor, and throw an `Uncaught TypeError: Cannot read property 'tokens' of undefined` for the first inactive editor.

Should we only update the scroll position of the active editor?

nathansobo:
You should interact with the `TextEditorPresenter` to retrieve information like this, as we're trying to move away from the model properties. That said, if the editor hasn't yet appeared on screen, you're right that we won't have a line height yet.

The ideal solution is to store the scroll position on the model in terms of row/column coordinates, so that whenever the editor comes on screen we can always restore to the correct relative position. That's a bit of a bigger job however. Perhaps at first we could only apply this to editors that actually have a line height stored on their presenter?

lee-dohm:
@mnquintana Is this something that still needs fixing? Let us know :grinning:

gareth-ferneyhough:
Bump. Would love to have this fixed.

mnquintana:
@lee-dohm Yup, definitely still an issue – I'm hoping to have time to work on some of these in the next week or so.

gabegattis:
@mnquintana Is this still something you plan to work on?

This is an issue I run into on an almost daily basis. I use a super small font size. If I want to show something to one of my coworkers, I bump up the font size so they can read it more easily. This usually moves the relevant text off of the screen and I have to go looking for it again.

mnquintana:
Haven't touched this in awhile – gonna close this out, and open a new PR if I have enough time

PR #6684, opened 2015-05-07
Save state of open windows more frequently

maxbrunsfeld:
We now save the open windows whenever:
- A new window is opened
- A window loses focus
- A window is closed (unless it's the last window open on linux or windows)
- `atom.exit()` is called

Refs #6466

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -99,7 +99,11 @@ class AtomApplication
   # Public: Removes the {AtomWindow} from the global window list.
   removeWindow: (window) ->
     @windows.splice @windows.indexOf(window), 1
-    @applicationMenu?.enableWindowSpecificItems(false) if @windows.length is 0
+    if @windows.length is 0
+      @applicationMenu?.enableWindowSpecificItems(false)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
@kevinsawicki could you :eyes: this code when you get a chance? I moved the logic for quitting when closing the last window on linux/windows to support the saving behavior. I tested it on Ubuntu and everything seemed :ok_hand:.

kevinsawicki:
Looks good to me.

Just want to confirm, the window closed event fires when you quit the app via `cmd-q` and also calling `atom.exit()`?

I always forget if there are certain cases where it doesn't.

maxbrunsfeld:
It looks like that event is not triggered if you call `atom.exit()`. Do you think we should we save the open windows in the `will-exit` handler so that they still get saved if you call that method?

kevinsawicki:
> Do you think we should we save the open windows in the will-exit handler so that they still get saved if you call that method?

Sounds good :+1:

rebornishard:
nice :+1:   
thank you, now i can click 'x' button properly :dancers:

PR #6686, opened 2015-05-07
Move selected line up and down in linux default conf 

israelcrux:
'ctrl-shift-up': 'editor:move-line-up'
  'ctrl-shift-down': 'editor:move-line-down'
For some reason this is default for OSX but not for Linux, while Sublime's default configuration for this keymap in Linux and Windows is indeed moving the line up or down


kevinsawicki:
Did you close this by accident? Is this something you'd still like to see merged in?

dantaex:
Yes I did, sorry I created a new one

PR #6690, opened 2015-05-07
Make 'atom' command w/ no args reopen previous windows

### LINKED ISSUE #6464: open previous window from command line?
> Opening Atom without the command line (e.g. using the dock on OSX) now re-opens any previously opened windows

Would it be hard for me to do a PR to add a command-line option to open atom with the previous windows?  I know my use-case is rare but I open X11 windows from the command-line and I have no desktop at all.

I know this isn't going to happen but logically using `atom` would open previous windows and `atom .` would open the working directory.

maxbrunsfeld:
Closes #6464

### DIFF
--- a/src/browser/main.coffee
+++ b/src/browser/main.coffee
@@ -133,7 +133,6 @@ parseCommandLine = ->
   safeMode = args['safe']
   apiPreviewMode = args['one']
   pathsToOpen = args._
-  pathsToOpen = [executedFrom] if executedFrom and pathsToOpen.length is 0
   test = args['test']
   specDirectory = args['spec-directory']
   newWindow = args['new-window']


izuzak:
I was a bit surprised by this new behavior when I built Atom today, but I think I'll get used to it. :smile: 

One thing I'm not sure about is how this works with the `--safe` option. Here are two examples that made me blink:

First:

1. run `atom --safe .` -- Atom opens the current folder in safe mode
2. quit Atom and run `atom` -- Atom reopens the previous window in safe mode even if I didn't ask for safe mode.

Second:

1. run `atom .` -- Atom opens the current folder
2. quit Atom and run `atom --safe` -- Atom reopens the previous window, but doesn't open it in safe mode even though I requested that.

I'm _guessing_ that the `--safe` option is ignored when windows are reopened? Honestly, I'm not sure what the correct behavior is here. If other folks are fine with this, I'll definitely get used to it. Just wanted to point it out. :doughnut:

maxbrunsfeld:
Thanks for pointing that out @izuzak. There's probably a similar behavior with `--dev` and `--one`. We should decide whether we want to persist these mode settings, and if we want to let the user override them with a new flag. Currently, we persist them. The relevant code is [here](https://github.com/atom/atom/blob/master/src/browser/atom-application.coffee#L427). @atom/feedback Opinions?

kevinsawicki:
>  Opinions?

After reading @izuzak's scenarios, I don't think we should persist these settings. All three are specialized modes that aren't intended to always be used. I would think persisting them will often be surprising to people since they are more like one-offs for specialized tasks.

nathansobo:
Agreed. These modes should only apply when the flag is explicitly supplied.

CharlesSocial:
I think this behavior is broken. Running `atom` always open a New File window. I have not seen it open a previous window.

maxbrunsfeld:
@CharlesSocial what version of Atom and what OS are you using? How are you exiting Atom? After exiting Atom, what is the contents of your `~/.atom/storage/application.json` file?

CharlesSocial:
OSX Yosemite,
Exiting with `command-q`
Atom Version 0.200.0
`~/.atom/storage/application.json` contents:
```
[{"initialPaths":["path/to/project/open/when/i/quit"]}]
```

Actually quitting this final time seems to have started the functionality. I guess it just updated? But this feature is now really problematic because it's opening several folders, one being `/` and trying to index every file.

maxbrunsfeld:
@CharlesSocial could you provide some exact steps to reproduce your problem, and report what the content of `application.json` is at the beginning of the steps?

CharlesSocial:
@maxbrunsfeld I see what's happening. If you already have an atom window open, and use `atom` again, it opens a New File window on top.

maxbrunsfeld:
:+1: Thanks for following up @CharlesSocial. Yeah, so it sounds like what you're experiencing is the expected behavior currently. The details you mentioned are by no means set in stone though. We're open to suggestions about what should happen in those cases.

PR #6694, opened 2015-05-07
:racehorse: Avoid repeated array allocation in ::getDecorations

maxbrunsfeld:
Some good low-hanging :cherries:. Screen updates:

Before:
before
300 ms / 3 fps. Spikes at the bottom are GC.

After:
after
20 ms / 50 fps. No GC.


benogle:
:+1:

PR #6706, opened 2015-05-08
Ignore trailing colons and whitespace when opening files from command line

### LINKED ISSUE #6654: atom filename:0 bug
Hi,
on atom 0.196 and mac os 10.9.3

For a file "file.rb", that opens fine with 

> atom file.rb:10

When i leave out the 10, so

> atom file.rb:

atom opens a new file with that filename (including the :)
The same happens when i put a 0 (which never really happens to me, but was the easier subject)

This happens for me when i click in iterm2 , as item always adds the ":" but not always supplies a number.

For sane usage of filenames one could just ignore the trailing ":"

izuzak:
This is the little brother of https://github.com/atom/atom/pull/6640 and together they should fix https://github.com/atom/atom/issues/6654. :family: :facepunch: 

cc @kevinsawicki for :eyes:

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -517,6 +517,8 @@ class AtomApplication
     return {pathToOpen} unless pathToOpen
     return {pathToOpen} if fs.existsSync(pathToOpen)
 
+    pathToOpen = pathToOpen.replace(/[:\s]+$/, '')
+
     [fileToOpen, initialLine, initialColumn] = path.basename(pathToOpen).split(':')
     return {pathToOpen} unless initialLine
     return {pathToOpen} unless parseInt(initialLine) >= 0
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -41,6 +41,22 @@ describe "Starting Atom", ->
           .then ({value}) -> expect(value).toBe "Hello!"
           .dispatchCommand("editor:delete-line")
 
+    it "removes all trailing whitespace and colons from the specified path", ->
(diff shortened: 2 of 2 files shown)


src/browser/atom-application.coffee
```diff
@@ -517,6 +517,8 @@ class AtomApplication
     return {pathToOpen} unless pathToOpen
     return {pathToOpen} if fs.existsSync(pathToOpen)
 
+    pathToOpen = pathToOpen.replace(/[:\s]*$/, '')
```
izuzak: I _think_ this isn't too aggressive and won't cause problems, but not 100% sure (because regexes make me :cold_sweat:).

izuzak: > Super minor, but I'd go with `+` instead of `*` here just to make it a bit clearer what the intent is.

Yeah, that's definitely more clear -- using that in https://github.com/atom/atom/commit/6a7d5e69459cfd6dc14c2d4e25c97353f84c49c9 (I rebased that change).

kevinsawicki:
@izuzak I'd like to try out adding some integration specs for this myself because I've never written any before.

I'll push a couple up to this pull request that try to test the trailing `:` case as well as the `:8` style case.

izuzak:
> I'd like to try out adding some integration specs for this myself because I've never written any before.

:+1: Yesss, thank you! I noticed there were no tests for this, but didn't feel like I could write them myself. I'll follow along and learn :car: :eyes:.

kevinsawicki:
> Ahhh, didn't know these tests existed -- never looked inside the integration folder before.

Yeah, they are relatively newish, added by @maxbrunsfeld to greatly improve the coverage of browser process code that was previously completely untested.

kevinsawicki:
:ship: Thanks @izuzak!

PR #6733, opened 2015-05-11
Render lines via tiles

as-cii:
This PR introduces a change in how we render lines. It does so by organizing them into tiles, which we then move up and down through a compositing transformation. The main advantage is that paint times dramatically reduce, because now we simply need to paint a small tile, instead of all the lines. This is especially noticeable during scrolling which, as a result, will be the subject of the benchmarks I am going to show later on.

Please, note that:
- [x] Specs are broken. The solution you can see here was achieved by spiking a bunch of ideas, and therefore I didn't pay attention to change specs/code gradually. As a result, I plan to adapt previous specs and write some new ones as soon as possible.
- [x] There's some evident structural duplication in code. However, I'd like to start refactoring that stuff as soon as I have unit tests to prevent naive errors.
- [x] Naming-wise I am not entirely happy with `TilesPresenter` and `LinesPresenter` and they're probably going to change. Any suggestion or improvement is very welcome here.
- I have intentionally left out line numbers from using this new approach. I'd like to keep this PR focused and work on that right after :ship:ping this one.

:racehorse: And now some benchmarks! :racehorse:
## Benchmarks

Tests conducted using:
- A `1920x1080` window
- `editor.scrollSensitivity = 130`
- A custom code (associated to a keyboard key) to simulate `mouseWheel` events
  
  ``` coffee
  document.querySelector("html /deep/ .editor-contents--private").dispatchEvent(new WheelEvent("mousewheel", wheelDeltaY: -250))
  ```
- Keyboard Key Repeat: `Fast` (Mac OS X setting)
- Delay Until Repeat: `Short` (Mac OS X setting)

I needed a (more or less) repeatable way to benchmark scrolling. To try this out, you can simply scroll up and down as you would normally do.
### master

master
### as-tiled-rendering

as-tiled-rendering
### Why this matters?

As you can notice, paint times are dramatically reduced. This may not be noticeable with your :eyes:, but it's nevertheless important 'cause it leaves some room for additional operations (such as a more expensive character measurement computation). As said earlier, things are going to speed up further once we introduce tiling for line numbers as well.

It would be great if you could try this out, reporting any eventual regression or glitch that I may have overlooked while implementing the tiling strategy. Moreover, any code review or additional :eyes: are always welcome.

Thank you!


src/lines-presenter.coffee
```diff
@@ -0,0 +1,41 @@
+module.exports =
+class LinesPresenter
```
as-cii: This should really be `TilePresenter`.

maxbrunsfeld:
:sparkles: Really exciting work.

src/text-editor-presenter.coffee
```diff
@@ -963,7 +963,7 @@ class TextEditorPresenter
   hasPixelPositionRequirements: ->
     @lineHeight? and @baseCharacterWidth?
 
-  pixelPositionForScreenPosition: (screenPosition, clip=true) ->
+  pixelPositionForScreenPosition: (screenPosition, clip=true, {absolute}={}) ->
```
nathansobo: Talk to me a little about this `absolute` flag... seems like you should always just choose the position relative to the viewport now since the lines layer is no longer moving. Either way, I think we should avoid allocating the options object on this hot path. I'm paranoid about garbage objects these days.

src/text-editor-presenter.coffee
```diff
@@ -298,50 +302,46 @@ class TextEditorPresenter
...
+      0, Math.floor(@startRow / linesPerTile) - @tileOverdrawMargin
+    )
+    endIndex = Math.min(
+      Math.ceil(@model.getScreenLineCount() / linesPerTile),
+      Math.ceil(@endRow / linesPerTile) + @tileOverdrawMargin
+    )
+
+    visibleTiles = {}
+    for index in [startIndex...endIndex]
+      presenter = @linesPresentersByTileIndex[index] ?= new LinesPresenter(@)
```
nathansobo: Is it too nasty to just build out the tile in a subroutine of the current presenter?

nathansobo:
I've been talking about wanting to do tiles for a year now. Thanks so much for pushing this forward!

benogle:
Wow, looks like a huge speedup

Zireael07:
Always good to see speedups!

src/text-editor-presenter.coffee
as-cii: I have inlined it for the time being :fire: 

The main idea with having a separate object, though, was to push the design of presenters/components in another direction. This object, as it is right now, tends to couple different behaviors in the same place and I can feel a lot of cognitive overhead every time I add new features or fix :bug:s (e.g. when I change how lines construction behave I have to think about the details of how cursors work, and vice versa).

:bulb: By separating each responsibility, I think we could be able to understand each presenter at a glance and, as a result, feel more confident to change it.

Probably I have been a bit too abstract here, but I'd love to hear your opinion about moving towards more cohesive presenters in the future :thought_balloon:

spec/text-editor-presenter-spec.coffee
```diff
@@ -657,140 +657,103 @@ describe "TextEditorPresenter", ->
...
-            screenRow: 8
-            text: line8.text
-            tokens: line8.tokens
-            top: 10 * 8
-          }
+          expect(presenter.getState().content.tiles[8]).toBeUndefined()
 
-          expect(lineStateForScreenRow(presenter, 9)).toBeUndefined()
-
-        it "does not overdraw above the first row", ->
```
as-cii: @nathansobo: is line-overdrawing strictly a performance "feature"? Overdrawing tiles doesn't seem to provide any significant speedup and, therefore, I simply got rid of them. What do you think?

nathansobo: You're correct to drop it with tiling. It used to be required to prevent a full screen repaint when adding/removing lines at the top/bottom. If the repainted regions were too close together they would be grouped. This approach bypasses all such nonsense.

nathansobo:
I lost your comment about breaking the presenter up into separate objects. I'm somewhat on board, at least with the idea of breaking the presenter up into objects that persist for the lifetime of the presenter. I'm less excited about objects that will come into and out of existence because that puts pressure on the GC. And whatever we do, I guess I want to make it well-encapsulated. Are there aspects of the problem that could be cleanly isolated? Right now, a single model update can update many aspects of the presenter state, which leaves us room to keep the view simple. If we break things up into objects but the objects aren't cleanly encapsulated, I think we've made things worse.

as-cii:
I have rebased this branch against master and fixed `TextEditorPresenter` specs. To make everything green I still need to polish `TextEditorComponent`'s tests, which are currently failing (I presume this won't take long, though).

Production code, on the other hand, is **ready for review**: there may still be a few things to refine but I feel quite good about it. I have tested it over the last days and it doesn't seem to break anything. Your mileage may vary, though, so if anyone wants to try this out I'll be happy to hear your feedback (in return you should notice some performance boosts, especially while scrolling).

### wrap-guide

After "virtualizing" vertical and horizontal scrolling, I have noticed that wrap-guide sticks in a fixed place on screen, even when scrolling horizontally. This actually makes sense, because wrap-guide is declared like this:

screen shot 2015-05-18 at 14 01 28

We should change it to update its position every time the editor scrolls using `pixelPositionForScreenPosition` instead. This, however, makes me think about a few questions:

1. How can we make sure we don't break existing packages which rely on absolute positions?
2. `DisplayBuffer#pixelPositionForScreenPosition` still returns the absolute pixel position for a given screen position. @nathansobo: I think this method should be changed to use relative positions, as `TextEditorPresenter#pixelPositionForScreenPosition` does. Do you think we can finally move all these duplicated position-related methods into `TextEditorPresenter`?

Thank you! :bow:

YuriSolovyov:
Works amazing even on:
```
Intel® Core™ i5 CPU 650 @ 3.20GHz
Intel® Ironlake Desktop
```
I can even turn on minmap animation flag!

src/lines-component.coffee
```diff
@@ -19,11 +20,7 @@ class LinesComponent
   placeholderTextDiv: null
 
   constructor: ({@presenter, @hostElement, @useShadowDOM, visible}) ->
```
as-cii: We may still introduce a couple of improvements here:

1. Avoid allocating a `TileComponent` object whenever new tiles need to be rendered. This will remove some pressure from GC and will improve performance as well. I somewhat liked the idea of having a tile encapsulated in its own component, but this area of the code is particularly performance sensitive and we should pay attention to each allocation.
2. When more than one tile gets rendered on screen for a given frame, we observe multiple `Parse HTML` events (one for each new tile). An alternative approach could be to parse all the lines at the same time and append them later to their respective tile. Not sure if the performance implications here are very serious, but this could make us save some cycles.
3. Avoid removing tile nodes while scrolling. Removing a tile node from the DOM takes between 0.5 and 1.0 ms: for performance sensitive scenarios like scrolling, we may simply put `display: none` and cleanup dangling tiles once scrolling ends.

Not sure if we should tackle 3), but I feel pretty confident we should have a look at 1) and 2)

spec/text-editor-presenter-spec.coffee
```diff
@@ -1893,20 +1848,18 @@ describe "TextEditorPresenter", ->
 
             getLineNumberGutterState(presenter).content.lineNumbers[key]
 
-          it "contains states for line numbers that are visible on screen, plus and minus the overdraw margin", ->
+          it "contains states for line numbers that are visible on screen", ->
```
nathansobo: I'm worried about the repaint behavior of the gutter if it doesn't overdraw and isn't tiled. Who knows what's changed in Chromium's graphics implementation, but previously when I didn't overdraw the red paint rectangle always covered the entire gutter when scrolling.

src/lines-component.coffee
nathansobo: Like I said in chat, I'm skeptical about 1. We're allocating plenty of objects in the typical render frame. One per tile doesn't seem to represent a significant increase.

I *would* like to see you try out 2, because in the timeline it definitely seems like there's overhead for the discrete parse calls. You could try out the wrapper div approach and just generate all the DOM nodes in masse, then populate the tiles. Not sure it will be a win by the time we do all that, but worth a look.

Not sure about 3. Probably easy to experiment with and see how much it helps?

nathansobo:
I'm super stoked about the apparent improvements to the Chrome compositor since I experimented with tiled rendering last year. When I tried it, leaving the tiles with transparent backgrounds was absolutely fatal to performance, so I had to give them opaque backgrounds and break selection rendering across tiles as well.

I'd be curious to see how much your compositing time drops if you give the tiles opaque backgrounds. Also, Chrome is *supposed* to render text with subpixel anti-aliasing if it's rendered on an opaque background and translated by an integral amount. It's never done that, but perhaps there's hope it will someday work. If we leave the tile backgrounds transparent, however, it will never be possible because the text is already rasterized by the time it gets to the GPU, meaning the information required to do subpixel anti-aliasing is lost.

So it's a toss up. Now that composite performance is <= 1ms for transparent tiles, it's definitely simpler to leave them that way and let the wrap guide and highlight decorations show through from underneath. But compositing may be even cheaper with opaque tiles, and with opaque tiles we theoretically have a possibility of getting subpixel AA someday.

nathansobo:
Also, what was your experience with larger tile sizes? I notice they're only 2 lines tall. The gutter is indeed doing a full repaint on every scroll now, but it's still pretty fast. But I bet you you could shave a bit more time if you could avoid that.

as-cii:
@nathansobo: thanks for the extensive feedback!

> I'd be curious to see how much your compositing time drops if you give the
tiles opaque backgrounds.

Assigning an opaque background doesn't impact layers compositing performance:
some experiments I conducted seem to report the same time for both (both ~
0.400ms - 0.500ms)

> Also, Chrome is supposed to render text with subpixel anti-aliasing if it's rendered on an opaque background and translated by an integral amount. It's never done that, but perhaps there's hope it will someday work. If we leave the tile backgrounds transparent, however, it will never be possible because the text is already rasterized by the time it gets to the GPU, meaning the information required to do subpixel anti-aliasing is lost.

That's very interesting and I didn't consider that: I've always assumed that the
background color on `.lines` was enough to maintain the original behavior. Now
that I think about it, however, I'd prefer to wait until Chrome supports AA for
two reasons:

* It would require changing how wrap-guide and highlights are rendered on
screen, as you're pointing out. Probably not a drastic change, but will still
introduce some complexity;

* It enables us to explicitly opt-in sub pixel anti-aliasing in the future: the
advantage here is avoiding to inadvertently introduce new behaviors (and bugs)
when we upgrade Chrome to a new version.

@nathansobo: that said, I trust your judgement on this and I am totally fine if
you insist we should give tiles an opaque background.

> Also, what was your experience with larger tile sizes?

I observed that having smaller tiles doesn't have an impact on performance and,
given I primarily work on a 1920x1080 window, I arbitrarily chose 6 as the
default value. This may be too large when the screen is smaller, so I think we
can safely decrease that to `4`. That said, however, I am not a huge fan of
*magic constants* and I would :heart: to analyze performance metrics of other
devices so that we can make a more informed decision. Is anyone on
@atom/feedback interested in trying this out? Your help would be much
appreciated! :bow:

> The gutter is indeed doing a full repaint on every scroll now, but it's still pretty fast. But I bet you you could shave a bit more time if you could avoid that. It takes a little over .2ms for the full gutter repaint.

I couldn't spot any difference with how the gutter is painted on **master**: am
I missing anything there? Do you mean we should introduce tiling for gutter as
well in this PR?

When I first opened it, my intention was to simply let this bake for a while
after :ship:ping it, and proceed to tile the gutter after we validated the
quality of this strategy. If that sounds unnecessary, I am ok to handle line
numbers tiling here as well :+1:

src/lines-component.coffee
as-cii: @nathansobo: I experimented with 2) and apparently we're not saving a significant amount of time. Moreover, batching HTML parsing requires us to add stuff in a two-step fashion: namely, HTML accumulation and rendering. Not a huge boost in complexity, but neither a "free lunch" (TANSTAAFL), therefore I'd say it's not worth going further down that path.

As for 3) it seems like we may be able to save ~1ms if we do this for gutter as well. What about addressing it after merging this?

nathansobo: :+1: on both.

src/tile-component.coffee
```diff
@@ -0,0 +1,285 @@
...
+  buildEndOfLineHTML: (id) ->
+    {endOfLineInvisibles} = @newTileState.lines[id]
+
+    html = ''
+    if endOfLineInvisibles?
+      for invisible in endOfLineInvisibles
+        html += "<span class='invisible-character'>#{invisible}</span>"
+    html
+
+  updateScopeStack: (scopeStack, desiredScopeDescriptor) ->
```
nathansobo: You're going to need to integrate my changes on https://github.com/atom/atom/pull/6757 unfortunately but it shouldn't be a huge deal.

nathansobo:
I feel like I was seeing more than 6 tiles in my test. I'm curious about whether it's better to specify tile count or tile size, then just pay for more tiles if the window is taller. Not sure what the more important metric is.

As for the opaque background, thanks for exploring that. I agree that leaving it transparent for now is way easier and probably the right move to limit ripples.

I think addressing the gutter can be done in a separate PR. So basically this is looking good as soon as the tests are passing and we test it out for a bit.

Nice work as usual @as-cii. The rendering layer and Atom as a whole are much better thanks to you.

spec/text-editor-component-spec.coffee
```diff
@@ -68,48 +68,111 @@ describe "TextEditorComponent", ->
       expect(nextAnimationFrame).not.toThrow()
 
   describe "line rendering", ->
-    it "renders the currently-visible lines plus the overdraw margin", ->
-      wrapperNode.style.height = 4.5 * lineHeightInPixels + 'px'
+    expectLineRender = (tileNode, {screenRow, top}) ->
```
as-cii: This is minor, but I am not very happy with how this test helper is named. Probably this would look better with a matcher?

```coffee
expect(tilesNodes[0]).toHaveRenderedLine(screenRow: 0, top: 10)
```

src/tile-component.coffee
```diff
@@ -0,0 +1,364 @@
...
+TokenTextEscapeRegex = /[&"'<>]/g
+MaxTokenLength = 20000
+
+cloneObject = (object) ->
+  clone = {}
+  clone[key] = value for key, value of object
+  clone
+
+module.exports =
+class TileComponent
```
as-cii: This diff is the result of an unfortunate false positive: ideally, git should have recognized `LinesComponent` as a whole new file, and `TileComponent` as a rename from the previous `lines-component.coffee`.

as-cii:
> I feel like I was seeing more than 6 tiles in my test. I'm curious about whether it's better to specify tile count or tile size, then just pay for more tiles if the window is taller. Not sure what the more important metric is.

:+1: 

I ended up managing `tileSize` as a configuration parameter, rather than an automatically calculated value (inferred from `tileCount`): I think it's just easier to reason about, and we save a few cycles as well.

> I think addressing the gutter can be done in a separate PR. So basically this is looking good as soon as the tests are passing and we test it out for a bit.

I wrote some new specs and fixed all the failing ones as well. I think this is now ready for a test drive! :checkered_flag:

spec/text-editor-component-spec.coffee
```diff
@@ -68,48 +68,111 @@ describe "TextEditorComponent", ->
...
+      lineNode = tileNode.querySelector("[data-screen-row='#{screenRow}']")
+      tokenizedLine = editor.tokenizedLineForScreenRow(screenRow)
+
+      expect(lineNode.offsetTop).toBe(top)
+      if tokenizedLine.text is ""
+        expect(lineNode.innerHTML).toBe("&nbsp;")
+      else
+        expect(lineNode.textContent).toBe(tokenizedLine.text)
+
+    it "renders the currently-visible lines into a tiled fashion", ->
```
nathansobo: `in a tiled fashion`

spec/text-editor-component-spec.coffee
nathansobo: This helper doesn't represent a clear concept to me, which is why it's tough to name. If I could articulate in a sentence I think it means this:

"Expect this tile node to contain a line node at the following offset."

What about `expectTileContainsRow` and passing the tile and the row as positional args. Then you could follow with `top: 10`. I think that might tell the story a bit better.

nathansobo:
So far so good after about a day.

benogle:
Hows the `wrap-guide`? You know the `wrap-guide` has to work. Muy importante.

nathansobo:
It works fine because Chromium's GPU compositing is fast now regardless of the opacity of the tiles, so they still have transparent backgrounds and not much has to change in that respect. Major win.

benogle:
Oh sweet. I was just being a jerk, but that's great.

as-cii:
@benogle @nathansobo: there's actually a subtle issue which affects the `wrap-guide` during horizontal scrolling. I have already made a PR to fix that (you can find it above this comment :point_up_2:) and ideally we should release it as soon as we release this one.

nathansobo:
@as-cii I'd like to merge this as soon as we have a backward-compatible fix for the wrap-guide ready to go.

as-cii:
@nathansobo: I have narrowed the scope of the flag and renamed it as well. Please, have one last look at this as well as at https://github.com/atom/wrap-guide/pull/31.

Thank you! :bow:

nathansobo:
Plan to merge this after the next release.

nathansobo:
Hey @as-cii, I found a regression. When saving strips whitespace at the end of the file and forces a scroll, the cursor's position isn't being updated correctly on screen until the next cursor movement.

cursor bug

as-cii:
@nathansobo: thanks for spotting the issue! :pray: :pray: :pray: 

I have fixed it in 9b4d62b: before that commit, when a change in the model caused `scrollTop` to change, `TextEditorPresenter` didn't update cursor state (which now use relative coordinates and needs to be recalculated).

My initial thought was to fix this :bug: by updating cursors right inside `updateScrollTop`, `updateScrollLeft`, etc., but those methods do not really deal with state right now and eventually I went with adding one more state update in `model.onDidChange`, which happen to solve the problem as well.

nathansobo:
@atom/feedback Can anyone test this on Windows and Linux (without a VM)? I want to make sure there are no performance regressions on those platforms. It seems unlikely, but there may be difference in the GPU compositing pipeline and I just want to make sure before putting this on master.

YuriSolovyov:
Tested here https://github.com/atom/atom/pull/6733#issuecomment-103051316, on Ubuntu 14.04, was great!

src/lines-component.coffee
```diff
@@ -343,59 +114,18 @@ class LinesComponent
 
   measureCharactersInNewLines: ->
     @presenter.batchCharacterMeasurement =>
-      for id, lineState of @oldState.lines
-        unless @measuredLines.has(id)
-          lineNode = @lineNodesByLineId[id]
-          @measureCharactersInLine(id, lineState, lineNode)
-      return
```
kevinsawicki: Should this `return` be retained so that the results of the `for` loop aren't built up in an array and returned?

kevinsawicki: Nevermind, look like it is, just way down in the diff, sorry.

as-cii: No worries about that. I am actually sorry [this diff got so ugly](https://github.com/atom/atom/pull/6733/files#r30785003) :pray:

nathansobo:
Tested on Windows and it's working correctly.

nathansobo:
Thanks very much @as-cii. People are really going to be happy to see scroll speed improvements.

alexandernst:
Let's see if this helps with my #6981

v3ss0n:
@nathansobo  @as-cii  i am testing on linux 64bit.
I opened up a compiled js file , with long lings and 1.2MB.
Selection is smooth, editing is a bit laggy
Scrolling is fine until it goes to bottom.

When scrolled to bottom  part , it freeze up , taking 100% CPU and seems to be generating tiles, for 20-40 secs? 
That bottom part is 1.1 mil chars , within 2 lines only.
It happens everytime when scroll back to top and bottom part again

Is it possible to move tiles generation to a worker process? I have tested with 205 and it rendered fast and fine.

as-cii:
Hi @v3ss0n, thank you for your feedback!

Could you please attach somewhere the file you're talking about? If it's private and you don't want to share it with everyone, it'd be awesome if you could share it via e-mail (you can find my address on my profile, https://github.com/as-cii).

Thanks a lot!

v3ss0n:
@as-cii  thanks a lot for fast response. I restarted atom and reopened, first it stop responding while atom start , and loading the file  , so i start from scratch (previously auto reload have a lot of tabs and split panes opened ). Now it rendering fine . I tested without --safe.

So only problem remain is when previous session restored , and large files are opened previously atom stop responding and  has to force close (can reproduce with --safe )

Rendering is smooth and fine , i copy / paste that file 10 times , making 11mb , and it dosen't slow down scrolling at all!

Thanks a lot for hard work , i will upload the file i testing to gist in a bit.

v3ss0n:
Here are things i've tested so far:

###Works

- [x] 11 MB file
- [x] More than 100k char per line
- [x] Compile Javascript file
- [x] Just caracters
- [x] Select word , Select Line , Select Multiple line , Select all

##Fails

- [ ] Reopen session with many tabs and one huge file.

alexandernst:
This looks very interesting https://github.com/nodejs/io.js/pull/1159
I wonder if Atom could use it for heavy tasks? @nathansobo

nathansobo:
@alexandernst That looks super interesting. Thanks for the heads up.

ilanbiala:
@nathansobo does Atom already use io.js or is it still on Node?

mnquintana:
@ilanbiala Atom uses io.js – Electron uses it.

nathansobo:
@ilanbiala Yep. It could take a while for this feature to get to us however since it's not even merged to master in iojs. Then it has to be released, integrated into Electron, and finally Atom has to upgrade to that version. But definitely worth keeping an eye on. It's something we could definitely make lots use of once it's available.

v3ss0n:
This may be a bit offtopic but curious.
@mnquintana atom is built using node 0.12 on my machine , is that already using IO.js ? 
Atom current version is not using electron  , yet right?

mehcode:
@v3ss0n From what I understand atom has a packaged version. The version you have is only used to bootstrap the process.

YuriSolovyov:
@v3ss0n it is currently io v1.6.3 and [being updated](https://github.com/atom/electron/commits/master) to io v2.2.1

v3ss0n:
ah i got it , it have packaged own IO.js separated from OS version ?

v3ss0n:
We have a problem with tile rendering , effecting themes with background transparency : https://github.com/v3ss0n/steam-pirate-ui
It renders tiles in more darkened color than my theme's color , i had compared it against atom version without tiled rendering. See attached.

tile_rendering_theme

from styling side , should i theme ```.lines``` or ```.tiles```

olmokramer:
Hey, awesome work on the tile rendering! I've got one problem with it, however, as it breaks my [block-cursor](https://atom.io/packages/block-cursor) package.

To have a block-cursor,  I used to add a (opaque) background color to the cursor element, and set its `z-index` to `-1`. That way, the cursor would render behind the text, and text would still be legible.

Now, with the new tile rendering, all the lines are inside a new element with an opaque background, so I can't render the cursor behind the lines anymore (or it would also be behind the tile, and thus invisible anyway)...

There's the option to set a very low alpha value on the cursor color, and render it in front of the text, but that way the cursor changes the text color, and it's tricky to find a value that produces a nice color *and* keeps text legible.

v3ss0n:
it seems we have no choice but to embrace it ? The benefits outweighs a few styling inflexibility , but i miss my beautiful theme tho .

olmokramer:
Hmmm, I just thought that highlights might suffer from the same issue, but then I found that a highlights container is added to each tile. I wonder if this is also possible for the cursors, or would it be too much of a performance issue? Or just too much work for too little gain?

v3ss0n:
Highlights? the Highlight selected package ? or Editor's highlight? I am gonna give a try. I was too busy  these days and haven't hack atom much last 2-3 weeks.

as-cii:
> Hmmm, I just thought that highlights might suffer from the same issue, but then I found that a highlights container is added to each tile. I wonder if this is also possible for the cursors, or would it be too much of a performance issue? Or just too much work for too little gain?

Given that tiles have an opaque background, I am afraid this is the only possible solution. Moreover, that would be consistent with how we handle highlights, as you’re pointing out. I kinda overlooked it because Atom’s default cursor works just fine as it stands in front of the tiles; the implementation will definitely bring some complexity but, ultimately, I think it’s the right way of doing it and, after all, it won’t mess things up that much.

@olmokramer: that said, what do you think about opening a brand new issue where we can keep track of this? I plan to work on it, but I’ll be traveling over the next days and I think I’ll be able to tackle it starting from the next month.

> Highlights? the Highlight selected package ? or Editor's highlight? I am gonna give a try. I was too busy these days and haven't hack atom much last 2-3 weeks.

@v3ss0n: highlights are an abstraction useful to decorate markers; you can have a look at [the related blog post](http://blog.atom.io/2014/07/24/decorations.html) for all the nitty-gritty details.

v3ss0n:
Thanks , interesting , i will look into it.

olmokramer:
@as-cii Thanks! That's truly awesome :) I'll submit a new issue in the evening.

PR #6735, opened 2015-05-11
Use /usr/bin/env bash in hashbang

mnquintana:
From https://github.com/atom/apm/issues/282#issuecomment-100289741

Refs https://github.com/atom/apm/pull/361


kevinsawicki:
:+1:

PR #6737, opened 2015-05-11
Tweak how + and - keys are handled when constructing menu keystrokes

### LINKED ISSUE #6713: View > Increase Font Size doesn't show keybinding
screenshot 2015-05-09 15 49 58

Atom 0.197.0-a79a8fa in safe mode, OS X 10.10.3

Can anyone on Linux or Windows confirm?

izuzak:
This is an attempt to fix https://github.com/atom/atom/issues/6713 and another problematic case when constructing keystrokes for menus.

The problem in https://github.com/atom/atom/issues/6713 was that the keybinding for `window:increase-font-size` is `cmd-+`, but Electron doesn't accept `+` characters in keystrokes -- [it accepts `Plus`](https://github.com/atom/electron/blob/master/docs/api/accelerator.md#available-key-codes). 29a4ab1 fixes this by detecting `+` keys and switching them with `Plus`. 

screen shot 2015-05-11 at 14 08 08

Another thing I noticed is that keystrokes are split on `-` (which is the separator Atom uses), which causes problems for keystrokes which use `-` as the key. For example, `window:decrease-font-size` has `cmd--` as one of its keybindings. The reason why this problem doesn't manifest currently is that [`window:decrease-font-size` has two keybindings currently](https://github.com/atom/atom/blob/f1f8c3d0a72de7d20bcde4fd145261c34d3e7462/keymaps/darwin.cson#L79-L80), and the second one is getting picked when constructing the menu (the one which doesn't use `-`):

```
  'cmd--': 'window:decrease-font-size'
  'cmd-_': 'window:decrease-font-size'
```

If you remove that second keybinding, the menu entry for `Decrease font size` will not have a keystroke as well. 71fba99 fixes this by splitting the keystroke on `-`, but only if there's another character after it (so, it doesn't split on trailing `-`).

I tried to figure out how to write tests for this, but failed. :disappointed: @kevinsawicki @maxbrunsfeld if you have some :hourglass: to point me in the right direction -- I'd be happy to add those.

### DIFF
--- a/src/browser/application-menu.coffee
+++ b/src/browser/application-menu.coffee
@@ -163,7 +163,7 @@ class ApplicationMenu
     return null unless firstKeystroke
 
     modifiers = firstKeystroke.split('-')
-    key = modifiers.pop()
+    key = modifiers.pop().toUpperCase().replace('+', 'Plus')
 
     modifiers = modifiers.map (modifier) ->
       modifier.replace(/shift/ig, "Shift")
@@ -171,5 +171,5 @@ class ApplicationMenu
               .replace(/ctrl/ig, "Ctrl")
               .replace(/alt/ig, "Alt")
 
-    keys = modifiers.concat([key.toUpperCase()])
+    keys = modifiers.concat([key])
     keys.join("+")
--- a/src/browser/application-menu.coffee
+++ b/src/browser/application-menu.coffee
@@ -162,7 +162,7 @@ class ApplicationMenu
     firstKeystroke = keystrokesByCommand[command]?[0]
     return null unless firstKeystroke
 
-    modifiers = firstKeystroke.split('-')
+    modifiers = firstKeystroke.split(/-(?=.)/)
     key = modifiers.pop().toUpperCase().replace('+', 'Plus')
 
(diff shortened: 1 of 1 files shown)


50Wliu:
My personal opinion on this:
Wouldn't it be better if a PR was made to Electron so that it could accept <kbd>+</kbd> as keystrokes?  That sounds like a better solution than converting it to <kbd>Plus</kbd>.

mnquintana:
> Wouldn't it be better if a PR was made to Electron so that it could accept <kbd>+</kbd> as keystrokes?  That sounds like a better solution than converting it to <kbd>Plus</kbd>.

That does seem like a better long-term solution to me - it does seem a little strange that Atom should need to handle this much about keystrokes instead of Electron.

But, in the meantime, this does solve that issue so :+1:

mnquintana:
Although actually now that I think about it, the Font Size menu items should be showing the simpler keybindings instead (which is more standard in other apps anyway) - is that as simple as reordering the keybindings in keymaps/{platform}.cson? (This may also be a topic for another issue, I'm not sure.)

kevinsawicki:
Thanks for fixing this :ship:

PR #6750, opened 2015-05-12
Remove extra function wrapper in deprecated Decoration::update method

izuzak:
While I was [updating](https://github.com/atom/decoration-example/pull/1) decoration-example to 1.0 APIs, I noticed that this `Decoration::update` call was not triggering a deprecation warning and might be wrong (because it will return a function which would need to be called again). I think the extra function wrapper is not needed so this removes it.

After removing it, the deprecation is correctly reported in Deprecation cop when calling `Decoration::update`.

screen shot 2015-05-12 at 12 58 43


kevinsawicki:
Great catch :green_heart:

PR #6755, opened 2015-05-12
Check for updates every 4 hours

### LINKED ISSUE #4936: Periodically check for available updates
Could we get an indicator when there's an update available for Atom, in addition to the file menu?  If someone leaves Atom open for days at a time, they could easily end up several versions behind.

I'm picturing something like Chrome's indicator, where the hamburger icon goes from green->yellow->red depending on how stale the user's version is.  Maybe just a little light in the status bar.

kevinsawicki:
Closes #4936

### DIFF
--- a/src/browser/auto-update-manager.coffee
+++ b/src/browser/auto-update-manager.coffee
@@ -53,7 +53,11 @@ class AutoUpdateManager
       @emitUpdateAvailableEvent(@getWindows()...)
 
     # Only released versions should check for updates.
-    @check(hidePopups: true) unless /\w{7}/.test(@version)
+    unless /\w{7}/.test(@version)
+      checkForUpdates = => @check(hidePopups: true)
+      thirtyMinutes = 1000 * 60 * 30
+      setInterval(checkForUpdates, thirtyMinutes)
+      checkForUpdates()
 
     switch process.platform
       when 'win32'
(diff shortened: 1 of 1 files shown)


kevinsawicki:
@paulcbetts it is okay to run `Update.exe --download` every 30 minutes right? It won't download anything if it already exists on disk right? Thanks.

kevinsawicki:
> It won't download anything if it already exists on disk right?

I mean, it won't re-download packages already downloaded, I would guess it still hits `RELEASES` which is fine.

paulcbetts:
@kevinsawicki Should be fine, yeah

Zireael07:
Will it crash or something if I don't have net access at the moment it runs Update.exe --download? I often work without net access at all (wireless card off/cord not plugged in)

kevinsawicki:
> Will it crash or something if I don't have net access at the moment it runs Update.exe --download?

Nope, it will just do nothing and check again in 30 minutes.

kevinsawicki:
@thedaniel do you have any concerns with all Atoms hitting the `https://atom.io/api/updates` endpoint every 30 minutes?

I could definitely bump it to an hour or 4 hours if you think if would be a better starting value.

paulcbetts:
Hm. So, there is one problem - checking every 30mins has the potential to spam the update log so much that it gets hella big. Probably want to delete it from time to time.

thedaniel:
@kevinsawicki that stuff is pretty heavily cached, I think it's fine.

kevinsawicki:
Downgraded this to every 4 hours just to be conservative to start.

bendem:
That should totally be configurable

Issue #929, opened 2013-10-05
Resizing the window leaves a white trail 


izuzak:
I started noticing a white trail when resizing the Atom window if a medium-sized file is loaded (or larger). This GIF was made with a file that has 1000 lines and `This is a test` text in every other line:

![out](https://f.cloud.github.com/assets/38924/1275351/516ad54e-2df4-11e3-8240-53702ad391e8.gif)

(sorry for the quality! still learning how to create GIFs)

The trail is shown only when the resizing increases the size of the window, not when the size is decreased. 

Atom version 31.0. OSX 10.8.5

jordwalke:
Is there really no way to solve this? Sublime and other modern IDEs don't have this ugly window resize issue. It's just one more reminder that "this is a web view" - a glitch in the matrix.
Last time I checked, Safari/Webkit doesn't have this issue.

jordwalke:
Visual Studio Code Insiders version has solved this now. What do they do to solve it?

faustinoaq:
Ping, Atom 1.19 still has this issue :sweat_smile:

Redsleeve:
Ping, Atom 1.25.0 still has this issue :(

codeblooded:
@jordwalke is right about VSCode not doing it? What are they doing differently?

PR #6759, opened 2015-05-12
Show window sooner

kevinsawicki:
This changes Atom to show the editor window as quickly as possibly, before package and theme loading begins.

It stores the background color in local storage to minimize the color changes once the editor loads.

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -196,6 +196,10 @@ class Atom extends Model
   #
   # Call after this instance has been assigned to the `atom` global.
   initialize: ->
+    dimensions = @restoreWindowDimensions()
+    maximize = dimensions?.maximized and process.platform isnt 'darwin'
+    @displayWindow({maximize})
+
     sourceMapCache = {}
 
     window.onerror = =>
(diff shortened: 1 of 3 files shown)


kevinsawicki:
@benogle want to give this branch a try when you get a chance?

benogle:
Yeah will do in a few

src/atom.coffee
```diff
@@ -572,6 +581,13 @@ class Atom extends Model
     dimensions = @getWindowDimensions()
     @state.windowDimensions = dimensions if @isValidDimensions(dimensions)
 
+  storeWindowBackground: ->
```
mnquintana: I'm actually really excited about this :raised_hands:

mnquintana: Would this also resolve https://github.com/atom/atom/issues/929? (Seems like it would.)

Zireael07:
Great!

src/atom.coffee
kevinsawicki: @mnquintana I'm not sure it will, I think that was an underlying Chrome issue that is still present.

kevinsawicki:
shipping this :fountain:

benogle:
:tractor:

mark-hahn:
What does the tractor mean?  I always feel like I'm left out of an inside
joke.

benogle:
He's plowing, man

kevinsawicki:
> What does the tractor mean? 

My interpretation of it is sometimes "shipping" with a :ship: gets old, so people sprinkle in other vehicle types like :speedboat: :dash:, :boat:, :car: :dash:, and :tractor: just to add a little variety to signal that something is "moving", "motoring", "sailing", etc. along.

yongkangchen:
+1

PR #6764, opened 2015-05-13
:arrow_up: fuzzy-finder@0.84.0

mnquintana:
Adds fuzzy match character highlighting

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "encoding-selector": "0.20.0",
     "exception-reporting": "0.24.0",
     "find-and-replace": "0.162.0",
-    "fuzzy-finder": "0.83.0",
+    "fuzzy-finder": "0.84.0",
     "git-diff": "0.55.0",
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Awesome, I'll take a look at this today

mnquintana:
Also, I'm not totally sure on this, but I feel like we should add highlighting to all the other SelectLists before we ship this (just for consistency) – symbols-view and command-palette off the top of my head

kevinsawicki:
> but I feel like we should add highlighting to all the other SelectLists before we ship this (just for consistency)

Yeah, that sounds good

kevinsawicki:
I think this is worth shipping in the next release before we it into the symbols view and command palette, it is probably most helpful in the fuzzy finder where you have lots of paths that may be very similar and the visual cue is really helpful.

Thanks so much for reviewing and merging the fuzzy finder pull request :ship:

PR #6773, opened 2015-05-13
Do not create an AutoUpdateManager when in test mode.

bolinfest:
Tests should be 100% reproducible, but `AutoUpdateManager` introduces unpredictable behavior, so it should not be installed when Atom is launched via `apm test`.

### DIFF
--- a/src/browser/application-menu.coffee
+++ b/src/browser/application-menu.coffee
@@ -9,11 +9,10 @@ _ = require 'underscore-plus'
 # and maintain the state of all menu items.
 module.exports =
 class ApplicationMenu
-  constructor: (@version) ->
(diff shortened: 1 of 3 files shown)


bolinfest:
Ah, good catch @kevinsawicki!

kevinsawicki:
@bolinfest we have a few specs that verify events out of the auto updater so instead of not creating it in test mode, I think we should just disable it from checking for updates so it can still be tested but it won't ever actually cause an update to be performed.

kevinsawicki:
:ship: Thanks for this @bolinfest

bolinfest:
@kevinsawicki sounds good to me!

PR #5217, opened 2015-01-22
[Gutter] Implement API for Creating Custom Gutters

jssln:
Summary: This implements the gutter API described in https://github.com/atom/atom/issues/5038.

Test Plan: Added unit tests. Tested manually as well -- if you rebase before the commit "Remove test-only CSS", there is some default CSS that makes gutters and decorations obviously visible when you add them via commands in the console. Example test commands, on top of this default CSS:

``` javascript
var e = atom.workspace.getActiveEditor()
var g = e.addGutter({name: 'test'})

var m = e.markBufferRange([[1, 0], [10, 0]])
var d = e.decorateMarker(m, {type: 'gutter', gutterName: 'test', class: 'test-class'})

var item = document.createElement('div')
item.className = 'test-item'
item.style['background-color'] = 'green'
var m2 = e.markBufferRange([[12, 0], [20, 0]])
var di = e.decorateMarker(m2, {type: 'gutter', gutterName: 'test', class: 'test-class', item: item})
```

Updating:

``` javascript
var item2 = document.createElement('div')
item2.className = 'test-item2'
item2.style['background-color'] = 'red'
di.setProperties({type: 'gutter', gutterName: 'test', class: 'test-class2', item: item2})

g.hide()
g.show()
g.destroy()
g = e.addGutter({name: 'test'})
```

PR #6776, opened 2015-05-13
Refactor `state.gutters` in TextEditorPresenter; pass minimal state to gutter components

jssln:
This is the first part in a series of follow-ups to polish off https://github.com/atom/atom/pull/5217.

This PR does the following:
- Changes `@state.gutters` in TextEditorPresenter to be an Array of Objects of the form:
  - gutter: Gutter model object
  - visible: boolean, whether the gutter is visible
  - content: either what used to be @state.gutters.lineNumberGutter or @state.gutters.customDecorations[gutterName], depending on the gutter
  - styles: an Object with fields: scrollHeight, scrollTop, backgroundColor
- Changes the `state` that is passed to each gutter component's `::updateSync` to only contain the state that it needs.

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -481,12 +481,12 @@ class TextEditorPresenter
     for gutter in @model.getGutters()
       gutterName = gutter.name
       gutterDecorations = @customGutterDecorations[gutterName]
-      if not gutterDecorations
-        @customGutterDecorations[gutterName] = {}
-      else
+      if gutterDecorations
         # Clear the gutter decorations; they are rebuilt.
         # We clear instead of reassigning to preserve the reference.
         @clearDecorationsForCustomGutterName(gutterName)
+      else
+        @customGutterDecorations[gutterName] = {}
       return if not @gutterIsVisible(gutter)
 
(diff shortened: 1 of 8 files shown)


nathansobo:
@as-cii I gave this a quick look and everything seems reasonable. Wanted to give you a chance to weigh in before merging since you have a lot of ownership over the presenter.

src/text-editor-presenter.coffee
```diff
@@ -461,23 +473,43 @@ class TextEditorPresenter
...
+    if @model.isMini()
+      # Mini editors have no gutter decorations.
+      # We clear instead of reassigning to preserve the reference.
+      @clearAllCustomGutterDecorations()
 
     for gutter in @model.getGutters()
       gutterName = gutter.name
-      @state.gutters.customDecorations[gutterName] = {}
+      gutterDecorations = @customGutterDecorations[gutterName]
+      if not gutterDecorations
```
as-cii: :art: Super minor, but what do you think if we reverse the conditional here (this way we should be able to remove `not`)?

jssln: Done!

PR #6777, opened 2015-05-14
Improve performance of opening files

maxbrunsfeld:
Currently, opening large files takes a long time. Most of the time is spent tokenizing the lines and computing the screen lines (accounting for soft-wraps, etc). Because of a small logic error, this was all happening _twice_, every time we opened a file.

These flame graphs are for opening an example 2.3MB javascript file (jQuery 2.0.3, concatenated 10 times).
### Before

Opening the file took about **8.2 seconds**. The span of time between the two red arrows is unnecessary work.

screen shot 2015-05-13 at 8 18 58 pm
### After

No more duplicated work. Opening the file took about **3.2 seconds**.

screen shot 2015-05-13 at 8 17 45 pm

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -36,23 +36,21 @@ describe "TextEditor", ->
 
     it "preserves the invisibles setting", ->
       atom.config.set('editor.showInvisibles', true)
-      previousInvisibles = editor.displayBuffer.invisibles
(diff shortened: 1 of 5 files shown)


thomasjo:
:sunglasses:

kevinsawicki:
So awesome :+1:

benogle:
wowowowow o_O

PR #6778, opened 2015-05-14
Add opened files and folders to recently opened list

andrewrothman:
This pull request adds the recent files list to the Atom dock icon's menu on Mac OS.

screen shot 2015-05-13 at 9 37 23 pm

Electron's API supposedly also supports Windows, so theoretically this patch should also work on that platform, but I'm not able to test it.

No change will be visible on Linux.

Unfortunately I attempted to add a Clear Recent Documents dock menu item using the Electron API, but I wasn't able to make that work. Maybe someone else has some guidance on that or can implement it themselves.

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -168,6 +168,8 @@ class AtomWindow
     if @loaded
       @focus()
       @sendMessage 'open-locations', locationsToOpen
+      for location in locationsToOpen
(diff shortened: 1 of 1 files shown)


damieng:
Given the merge conflicts, lack of spec tests and no facility for clearing recent documents I'm going to close this.  If you wish to add these things please feel free to update the PR.

PR #6781, opened 2015-05-14
Remove hard coded restrictions to enable more correct indentions

svanharmelen:
Currently these hard coded restrictions prevent some very valid use cases (will add a before and after screencasts at the bottom) and limit the possibilities to implement correct indentation. As far as I’ve tested with it, I did not
find any negative and/or unexpected behaviour with the restrictions removed, but of course I did not test with all possible languages Atom supports.

If it turns out there is a language that does have negative effects from this change, I would argue that the correct place to fix that is in the specific language package by altering the increaseIndentPattern and/or (more likely) the decreaseIndentPattern, instead of relying on these hard coded restrictions.

Before:
before
As you can see here the bracket isn't indented correctly when moving it to the next line. The decreaseIndentPattern is correct as it does increase when removing and typing in the bracket again on the same line.

After:
after
After removing the hard coded restrictions, the bracket is indented as expected.

### DIFF
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -311,7 +311,7 @@ class LanguageMode
 
     desiredIndentLevel = @editor.indentationForBufferRow(precedingRow)
     desiredIndentLevel -= 1 unless increaseIndentRegex.testSync(precedingLine)
-    if desiredIndentLevel >= 0 and desiredIndentLevel < currentIndentLevel
+    if desiredIndentLevel >= 0
       @editor.setIndentationForBufferRow(bufferRow, desiredIndentLevel)
 
   getRegexForProperty: (scopeDescriptor, property) ->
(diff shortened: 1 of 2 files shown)


mnquintana:
Would you mind adding some specs for the cases this fixes? Thanks!

thomasjo:
> Would you mind adding some specs for the cases this fixes? Thanks!

Would be great if said specs cover both C-style languages (bracket bonanza) and whitespace significant languages such as Python. I just want to make sure this doesn't break anything, and also guard against future regressions.

svanharmelen:
@mnquintana @thomasjo yes of course... Will look up the current test for this code and add some tests for the updated behaviour.

thomasjo:
Let us know if you need any help :bow:

svanharmelen:
Thanks @mnquintana and @thomasjo for pointing me to the tests! It turns out this "fix" causes some new unwanted behaviour:

hmm
It should of course not keep decreasing indents with every new line...

So now I understand why this code was added in the first place :wink: Unfortunately I no longer know how to fix the indent issues as shown earlier in the screencasts. It seems the current auto indentation has a few (at least two) annoying limitations which most likely need more then a little tweak to solve them.

Will need to think it over again I guess, so will close this one for now.

PR #6789, opened 2015-05-14
Run package specs from the active item's project folder

maxbrunsfeld:
I was editing `find-and-replace` and `atom` in the same window, and it was annoying not being able to run specs for both projects. Now it tries to run the package specs for the active pane item. If there's no active pane item or it doesn't have a path, it uses the first project directory like it did before.


kevinsawicki:
Love it :+1:

PR #6808, opened 2015-05-15
Change hard coded restrictions to enable more correct indentions

svanharmelen:
Currently the hard coded restrictions prevent some very valid use cases (will add a before and after screencasts at the bottom) and limit the possibilities to implement correct indentation.

Before:
before
As you can see here the bracket isn't indented correctly when moving it to the next line. The decreaseIndentPattern is correct as it does increase when removing and typing in the bracket again on the same line.

After:
after
After removing the hard coded restrictions, the bracket is indented as expected.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2226,12 +2226,56 @@ describe "TextEditor", ->
         expect(editor.lineTextForBufferRow(1)).toBe '  '
         expect(editor.lineTextForBufferRow(2)).toBe '}'
 
-    describe "when a new line is appended before a closing tag (e.g. by pressing enter before a selection)", ->
-      it "moves the line down and keeps the indentation level the same when editor.autoIndent is true", ->
-        atom.config.set('editor.autoIndent', true)
-        editor.setCursorBufferPosition([9, 2])
-        editor.insertNewline()
-        expect(editor.lineTextForBufferRow(10)).toBe '  };'
+    describe ".insertNewLine()", ->
+      describe "when a new line is appended before a closing tag (e.g. by pressing enter before a selection)", ->
+        it "moves the line down and keeps the indentation level the same when editor.autoIndent is true", ->
+          atom.config.set('editor.autoIndent', true)
+          editor.setCursorBufferPosition([9, 2])
+          editor.insertNewline()
+          expect(editor.lineTextForBufferRow(10)).toBe '  };'
+
+      describe "when a newline is appended with a trailing closing tag behind the cursor (e.g. by pressing enter in the middel of a line)", ->
+        it "indents the new line to the correct level when editor.autoIndent is true and using a curly-bracket language", ->
+          waitsForPromise ->
+            atom.packages.activatePackage('language-javascript')
+
+          runs ->
+            atom.config.set("editor.autoIndent", true)
+            editor.setGrammar(atom.grammars.selectGrammar("file.js"))
+            editor.setText('var test = function () {\n  return true;};')
+            editor.setCursorBufferPosition([1, 14])
+            editor.insertNewline()
+            expect(editor.indentationForBufferRow(1)).toBe 1
+            expect(editor.indentationForBufferRow(2)).toBe 0
+
+        it "indents the new line to the correct level when editor.autoIndent is true and using a off-side rule language", ->
+          waitsForPromise ->
+            atom.packages.activatePackage('language-coffee-script')
+
+          runs ->
+            atom.config.set("editor.autoIndent", true)
+            editor.setGrammar(atom.grammars.selectGrammar("file.coffee"))
+            editor.setText('if true\n  return trueelse\n  return false')
+            editor.setCursorBufferPosition([1, 13])
+            editor.insertNewline()
+            expect(editor.indentationForBufferRow(1)).toBe 1
+            expect(editor.indentationForBufferRow(2)).toBe 0
+            expect(editor.indentationForBufferRow(3)).toBe 1
+
+      describe "when a newline is appended on a line that matches the decreaseNextIndentRegex", ->
+        it "indents the new line to the correct level when editor.autoIndent is true", ->
+          waitsForPromise ->
+            atom.packages.activatePackage('language-go')
+
+          runs ->
+            atom.config.set("editor.autoIndent", true)
+            editor.setGrammar(atom.grammars.selectGrammar("file.go"))
+            editor.setText('fmt.Printf("some%s",\n	"thing")')
+            editor.setCursorBufferPosition([1, 10])
+            editor.insertNewline()
+            expect(editor.indentationForBufferRow(1)).toBe 1
+            expect(editor.indentationForBufferRow(2)).toBe 0
+
 
     describe ".backspace()", ->
       describe "when there is a single cursor", ->
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -246,8 +246,12 @@ class LanguageMode
     iterator.next()
     scopeDescriptor = new ScopeDescriptor(scopes: iterator.getScopes())
 
+    increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
+
     currentIndentLevel = @editor.indentationForBufferRow(bufferRow)
-    return currentIndentLevel unless increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    return currentIndentLevel unless increaseIndentRegex
 
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
@@ -256,13 +260,17 @@ class LanguageMode
       precedingRow = bufferRow - 1
       return currentIndentLevel if precedingRow < 0
 
-    precedingLine = @buffer.lineForRow(precedingRow)
     desiredIndentLevel = @editor.indentationForBufferRow(precedingRow)
-    desiredIndentLevel += 1 if increaseIndentRegex.testSync(precedingLine) and not @editor.isBufferRowCommented(precedingRow)
+    return desiredIndentLevel if @buffer.isRowBlank(precedingRow)
 
-    return desiredIndentLevel unless decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    line = @buffer.lineForRow(bufferRow)
-    desiredIndentLevel -= 1 if decreaseIndentRegex.testSync(line)
+    unless @editor.isBufferRowCommented(precedingRow)
+      precedingLine = @buffer.lineForRow(precedingRow)
+      desiredIndentLevel += 1 if increaseIndentRegex?.testSync(precedingLine)
+      desiredIndentLevel -= 1 if decreaseNextIndentRegex?.testSync(precedingLine)
+
+    unless @editor.isBufferRowCommented(bufferRow)
+      bufferLine = @buffer.lineForRow(bufferRow)
+      desiredIndentLevel -= 1 if decreaseIndentRegex?.testSync(bufferLine)
 
     Math.max(desiredIndentLevel, 0)
 
@@ -298,21 +306,26 @@ class LanguageMode
   # bufferRow - The row {Number}
   autoDecreaseIndentForBufferRow: (bufferRow) ->
     scopeDescriptor = @editor.scopeDescriptorForBufferPosition([bufferRow, 0])
-    increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    return unless increaseIndentRegex and decreaseIndentRegex
+    return unless decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
 
     line = @buffer.lineForRow(bufferRow)
     return unless decreaseIndentRegex.testSync(line)
 
     currentIndentLevel = @editor.indentationForBufferRow(bufferRow)
     return if currentIndentLevel is 0
+
     precedingRow = @buffer.previousNonBlankRow(bufferRow)
     return unless precedingRow?
-    precedingLine = @buffer.lineForRow(precedingRow)
 
+    precedingLine = @buffer.lineForRow(precedingRow)
     desiredIndentLevel = @editor.indentationForBufferRow(precedingRow)
-    desiredIndentLevel -= 1 unless increaseIndentRegex.testSync(precedingLine)
+
+    if increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
+      desiredIndentLevel -= 1 unless increaseIndentRegex.testSync(precedingLine)
+
+    if decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
+      desiredIndentLevel -= 1 if decreaseNextIndentRegex.testSync(precedingLine)
+
     if desiredIndentLevel >= 0 and desiredIndentLevel < currentIndentLevel
       @editor.setIndentationForBufferRow(bufferRow, desiredIndentLevel)
 
@@ -325,6 +338,9 @@ class LanguageMode
 
   decreaseIndentRegexForScopeDescriptor: (scopeDescriptor) ->
     @getRegexForProperty(scopeDescriptor, 'editor.decreaseIndentPattern')
+
+  decreaseNextIndentRegexForScopeDescriptor: (scopeDescriptor) ->
+    @getRegexForProperty(scopeDescriptor, 'editor.decreaseNextIndentPattern')
 
   foldEndRegexForScopeDescriptor: (scopeDescriptor) ->
     @getRegexForProperty(scopeDescriptor, 'editor.foldEndPattern')
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "language-css": "0.32.0",
     "language-gfm": "0.78.0",
     "language-git": "0.10.0",
-    "language-go": "0.27.0",
+    "language-go": "0.28.0",
     "language-html": "0.40.0",
     "language-hyperlink": "0.14.0",
     "language-java": "0.15.0",
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -395,10 +395,7 @@ class Selection extends Model
       @editor.setIndentationForBufferRow(oldBufferRange.start.row, desiredIndentLevel)
 
     if options.autoIndentNewline and text is '\n'
-      currentIndentation = @editor.indentationForBufferRow(newBufferRange.start.row)
       @editor.autoIndentBufferRow(newBufferRange.end.row, preserveLeadingWhitespace: true, skipBlankLines: false)
-      if @editor.indentationForBufferRow(newBufferRange.end.row) < currentIndentation
-        @editor.setIndentationForBufferRow(newBufferRange.end.row, currentIndentation)
     else if options.autoDecreaseIndent and NonWhitespaceRegExp.test(text)
       @editor.autoDecreaseIndentForBufferRow(newBufferRange.start.row)
 


svanharmelen:
@thomasjo did some more thinking and came to this solution. All existing tests now run successfully. So I wouldn't mind adding more/specific tests for this, but I could use some help/pointers as to where to place those tests (in `text-editor-spec.coffee`, `selection-spec.coffee` or `language-mode-spec.coffee`) and what you think should be in the test exactly?

I looked around in the existing tests and there are already lots of tests around the indentation logic, so the only tests I can think of that would make sense are a test that verifies the behaviour as shown in the screencasts for both a language using brackets and a language using white spaces.

svanharmelen:
This last commit adds another improvement which I will describe in a little more detail here...

Currently there are only two regexps that can influence the indents. Those are the `increaseIndentPattern` (which tells Atom the indent of **next** line should be **increased**) and `decreaseIndentPattern` (which tells Atom the indent of the **current** line should be **decreased**).

But I found that a couple of languages would need a 3rd regexp in order to support their use cases. This 3rd regexp should be a mixture of the existing two so it tells Atom that the indent of the **next** line should **decrease**.

I’ll add a screencast to show a use case for this 3rd regexp which I would like to call `decreaseNextIndentPattern`. And so to keep things in sync and so they more accurate describe what they do, I would opt for renaming the current `increaseIndentPattern` to
`increaseNextIndentPattern`.

I did this already in the PR, but in a fully backwards compatible way so this can be changed over time.

As for an example, see this before screencast:
before
As you can see the new line indents to the same level as the previous one. But in this case (in Go) it should decrease the indent level.

And after:
after
Here you can see the (expected) result after these commits.

Again all existing tests are successful, but I fully understand we would need some tests before merging something like this in. But any help/pointers on those tests would be great to make sure they are what you expect and what would be needed to get this merged.

Thanks...

svanharmelen:
@thomasjo @kevinsawicki just added some tests (and squashed the commits) for the first change (removing the hard coded restriction for decreasing indents) for both a curly-bracket language and a off-side rule language. Does these tests make sense and are they enough to cover the first use case?

If so I will move on and add some tests for the second use case (the one needing the new/additional `decreaseNextIndentPattern`) hoping to get this PR up-to-spec so it can be merged.

So please let me know any comments and/or concerns. Thanks!

thomasjo:
I'll try to find time to play around with this PR tomorrow. Just wanted to let you know it's not being ignored. Maybe someone from @atom/feedback has time to give it a go?

svanharmelen:
Cool... And thanks for the ping :smiley:

svanharmelen:
@thomasjo any comments and/or feedback you can share already?

svanharmelen:
@thomasjo @kevinsawicki @mnquintana guess you guys are pretty busy as is, but would be great to get some feedback on this one.

I myself and a few colleagues are using this for about 2 weeks now and it really improved our user experience. So would be great to get this reviewed and (if no issues/concerns) merged into Atom.

Thanks!

maxbrunsfeld:
Wow thanks so much for diving in to fix this fairly gnarly issue @svanharmelen! Looking at the code now...

spec/text-editor-spec.coffee
```diff
@@ -2187,12 +2187,37 @@ describe "TextEditor", ->
...
+          waitsForPromise ->
+            atom.packages.activatePackage('language-coffee-script')
+
+          runs ->
+            editor.setText('if true\n  return trueelse\n  return false')
+            editor.setCursorBufferPosition([1,13])
+            editor.insertNewline()
+            expect(editor.lineTextForBufferRow(1)).toBe '  return true'
+            expect(editor.lineTextForBufferRow(2)).toBe 'else'
+            expect(editor.lineTextForBufferRow(3)).toBe '  return false'
```
maxbrunsfeld: Does this test verify the new feature that you added? I would think that in order to test the new behavior, you would need to change the indent-pattern settings in the `language-coffee-script` or `language-javascript` packages.

maxbrunsfeld: Oh, I think I see. I didn't realize that there was more to this PR than the addition of `decreaseNextIndentPattern`. So these tests must be about *other* fixes that you made to the indentation logic.

svanharmelen: Yes, you are absolutely right! Please see the first comment I made to this PR, as the existing tests are to verify that behaviour. And indeed they do not require any changes to the existing regexps as they seem to fix something that wasn't working as expected before.

As for the behaviour as described in the 3rd comment, for that indeed a regexp should be added to a language that needs/uses such behaviour. Go does, so I created a PR there which was merged just a few hours ago (https://github.com/atom/language-go/pull/44) whit an additional regexp. So now that this one is merged I can add some more tests to verify the additional behaviour as well. Will do that shortly...

src/language-mode.coffee
```diff
@@ -255,12 +255,32 @@ class LanguageMode
...
+          desiredIndentLevel = @editor.indentationForBufferRow(lastIndentingRow) + 1
+
+          for row in [lastIndentingRow..bufferRow]
+            continue if @editor.isBufferRowCommented(row)
+
+            rowLine = @buffer.lineForRow(row)
+            desiredIndentLevel -= 1 if decreaseNextIndentRegex?.testSync(rowLine)
+            desiredIndentLevel -= 1 if decreaseIndentRegex?.testSync(rowLine)
+
+          return Math.max(desiredIndentLevel, 0)
```
maxbrunsfeld: I don't fully understand what this new code does, and I'm a little bit concerned about the performance implications of calling `testSync()` in doubly-nested loop. Just to help me understand: could you explain in a little more detail your intent behind the `while` loop and the inner `for` loop?

svanharmelen: Of course... To determine the correct level of indentation it will search for the most recent line that was indented and then calculate the current indent level based on the rows between the last indented row and the current row.

So the first part of the while loop searches for a previous line that matches the indent pattern. As soon as 1 is found, it will enter the for loop and process every row between the two. After entering the for loop, it will never get back to the while loop but will return the function there.

Does this help or do you want/need more insights/reasoning?

svanharmelen: @maxbrunsfeld additionally please note that it will only ever get to the `while` and `for` loops when you insert a newline (press return) and the trailing chars that will then move to the newline, match a decrease or decreaseNext pattern.

So this code path itself will be triggered by every newline (return), but in most cases the loops will never be reached.

Still it should be efficient of course! But I do believe it is. I myself (and some colleagues) are using this for some time now and I can tell any delays or something alike. Don't know if there are benchmark tests for something like this to show the impact?

mnquintana:
@svanharmelen Would you mind merging in / rebasing master and fixing the merge conflicts?

svanharmelen:
@mnquintana done...

svanharmelen:
@maxbrunsfeld just updated the PR so Atom now uses version 0.27.0 of `language-go` package as that version is specifically updated to contain a `decreaseNextIndentPattern`. So with this new version I could add a test for the last part of this PR.

I believe all changes are now fully covered with tests so I would like to know what I can do next to get this merged? It would be really awesome to get this is Atom v1.0, but even if that's not an option we should really try to get this in as it improved the auto indentation a lot.

So please let me know your thoughts and how to continue.

/CC @thomasjo @kevinsawicki

svanharmelen:
@maxbrunsfeld @mnquintana just rebased again to have a clean/merge-able PR. Any update on how to proceed?

svanharmelen:
@thomasjo @maxbrunsfeld @kevinsawicki I really appreciate you guys a very busy, but this PR is open or about a month now and even through I did receive some comments, after following up on them it got pretty silent again.

I really believe this is a good fix that improves the overall indenting capabilities of Atom and does not have any general performance drawbacks. If you know of a way how to benchmark that (if that is a concern) then please let me know how to do that so I can take care of that.

Otherwise I don't know what else I can/should do to get this merged, so please if any of you can spare a minute? Thanks!

/CC @atom/feedback @anyone?

thomasjo:
Sorry! Been super busy, but will try to play around with it over the weekend.

svanharmelen:
@thomasjo I understand... And that would be really great! Will be awaiting your comments...

Thx!

svanharmelen:
Hi @thomasjo... Hope you had a good weekend and had some time to play with this? Any feedback would be great in order to get this one going. Thx...

maxbrunsfeld:
Ok, really sorry for the delay in giving you further feedback on this @svanharmelen. We've been super busy and this change is pretty involved. Thanks again for taking it on. Here is my take on it:

1. It looks like you must have squashed the commits that removed the hard-coded indentation restrictions in `selection.coffee` and the ones that added the `decreaseNextIndentPattern` setting. I'd really like to see this work *un-squashed* because my feeling is that we *should* definitely remove the hard-coded restrictions as you described, but I don't think that we should add the new indentation regexp (or rename the existing indentation regexps) right now.

2. In `LanguageMode::suggestedIndentForTokenizedLineAtBufferRow`, I don't think it should be necessary to search all the way up the previous line that matches the `increaseIndentPattern`. Maybe this was only needed to support the `decreaseNextIndentPattern` feature? I could be mistaken, but I think that because of the way these patterns work, it should always be sufficient to find the previous non-blank line.

Do you still have the un-squashed version of your work from this PR? Could you submit a second PR that references this one, and just addresses the problems that originally described at the beginning of the PR?

svanharmelen:
@thomasjo @maxbrunsfeld thanks for the update!

I must admit I think it's a bad idea to *only* fix the hard-coded indentation restriction and not add the `decreaseNextIndentPattern` feature as I believe with these two fixes together we are really fixing and improving the overall indentation capabilities of Atom to a level that will allow 99% (maybe even a 100%, but there probably some more corner cases out there I don't know about) of all use cases to be handled properly. Personally I feel that with this PR the indentation support and capabilities are now better in Atom than I'm used to have in other editors like Sublime (which already does a pretty good job for most languages).

As for your reasoning why you would like to hold of on the `decreaseNextIndentPattern` feature, I think your having the wrong impression of the code needed to fix the hard-coded indentation restriction properly. Looking all the way up for the last line matching the `increaseNextIndentPattern` is not needed for the `decreaseNextIndentPattern` feature at all. It's needed for fixing the hard-coded indentation restriction for languages like coffee-script (so any off-side rule languages that is).

But I just had another look (spent a couple of hours actually) to see if there really wasn't another way to get the same result without the additional `while` and `for` loops and I think I might have found a much, much better solution (sometimes looking at it again after a little while helps giving new insights :wink:).

So let me run some tests and work with the new code for a day or two to see if everything still works as intended after which I will update the PR and give you a ping again for another review.

As for renaming `increaseIndentPattern` to `increaseNextIndentPattern`, of course there is no *real* need for this one. Still I do believe the updated name describes the purpose and use of the regexp much better and with that it improves the readability of the code. Especially taken into account that `increaseIndentPattern` and `decreaseIndentPattern` seem to imply the same behaviour, but actually they behave very different (`increaseIndentPattern` only has effect on the **next line** where `increaseIndentPattern` has effect on the **current line**).

The code in this PR is currently allowing both pattern names so it doesn't require any immediate changes in existing (language) packages. So we can safely merge/update this now and then start updating all the existing (language) packages in the upcoming time (I will personally make a PR for that in every language package I can find :wink:). And then later on we can make the `increaseIndentPattern` deprecated for a while and again a bit later really remove support for the old pattern name.

But as said let me test and work with the updated solution first and if everything looks good, I'll update the PR and we can have another look. I think the current solution looks very clean and so I would say adding the `decreaseNextIndentPattern` feature should almost be a no-brainer (it requires only 2 additional lines in the `suggestedIndentForTokenizedLineAtBufferRow` function and if `decreaseNextIndentPattern` doesn't exist in your language package it will just skip the related code anyway).

I'll ping you again in a day or two. Thx!

maxbrunsfeld:
> But as said let me test and work with the updated solution first and if everything looks good

@svanharmelen That sounds great. Thanks so much for sticking with this.

> I must admit I think it's a bad idea to only fix the hard-coded indentation restriction and not add the `decreaseNextIndentPattern` feature

I'm not dead-set opposed to adding this feature, but because it constitutes a new API that we would need to support indefinitely, we need to be more cautious about adding it than the other change that you had originally addressed with this PR. I'm also considering the fact that currently, we support a subset of TextMate's indentation system (see section **24.2.2** of [the texmate docs](https://manual.macromates.com/en/appendix)). TextMate does not have a separate regex for `decreaseNextIndent`, and generally I find its indentation suggestions to be good. That said, there could still be room for improvement.

svanharmelen:
@maxbrunsfeld could you give Travis a little push to build/test the PR again (as the reason it currently failed wasn't due to code itself)?

Also please take a minute to review the PR again. I changed/updated the solution so it doesn't need/use any loops anymore, so I think the current solution is much better and cleaner.

As for renaming `increaseIndentPattern `to `increaseNextIndentPattern` and for adding the `decreaseNextIndentPattern` feature, I would really like to hold on to these changes. I really believe they improve the capabilities a lot and with these changes it enabled Atom to handle indentations better then other existing text editors currently can.

You referenced the TextMate docs in which I noticed they also acknowledge the need for a `increaseNextIndentPattern` like feature, but they called it `indentNextLinePattern`. Reading their docs, it seems the logic/reasoning behind the `indentNextLinePattern` is actually the logic currently implemented in Atom but then Atom uses the `increaseIndentPattern` name for it.  So if you like to stay close to the TextMate API then of course I'm fine with renaming `increaseIndentPattern` to `indentNextLinePattern` instead of `increaseNextIndentPattern`.

As for the `decreaseNextIndentPattern` (which could then become the `dedentNextLinePattern`), I see they currently don't offer a solution like this in TextMate, but I think they should. How else can a text editor dedent a line after it came across a continuation indent? The only why to automatically dedent after that, is by being able to dedent based on the content of the previous line.

Thanks!

thomasjo:
Bumped Mr. Travis CI. Hopefully it goes :green_heart:

svanharmelen:
@thomasjo unfortunately it seems I got the same error again: `Request for package information failed: atom.io is temporarily unavailable, please try again later.` Can you maybe check/confirm what is going on and let me know if I can/must do anything to get past this? Thanks...

thomasjo:
@svanharmelen Not your fault. The problem is caused by a license change in [jschardet](https://github.com/aadsm/jschardet). Either we need to fix the version of the dependency to v1.1.1, or we need to replace the dependency.

/cc @atom/feedback

svanharmelen:
@thomasjo check... Then I'll just wait for the feedback from @maxbrunsfeld :smiley:

maxbrunsfeld:
> You referenced the TextMate docs in which I noticed they also acknowledge the need for a increaseNextIndentPattern like feature, but they called it indentNextLinePattern.

I think TextMate's `indentNextLinePattern` is unrelated to the changes we're talking about here. It's for the special case where only *one* line should be indented, like this C code:

```c
if (a)
  b();
c();
```

We could definitely add this in Atom at some point, but I don't think it needs to affect the naming of our other patterns.

svanharmelen:
@maxbrunsfeld I just read the TextMate docs again and your absolutely right! Think I didn't read it close enough the first time...

So in that case please ignore the part about renaming `increaseIndentPattern` to `increaseNextLinePattern` instead of to `increaseNextIndentPattern` :wink:

src/language-mode.coffee
```diff
@@ -246,23 +246,30 @@ class LanguageMode
...
+    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
+    decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
+
     currentIndentLevel = @editor.indentationForBufferRow(bufferRow)
-    return currentIndentLevel unless increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    return currentIndentLevel unless increaseNextIndentRegex
 
-    if options?.skipBlankLines ? true
+    bufferLine = @buffer.lineForRow(bufferRow)
+    if options?.skipBlankLines ? true or (decreaseIndentRegex?.testSync(bufferLine) and not @editor.isBufferRowCommented(bufferRow))
```
maxbrunsfeld: :zap: This is looking very clean now.

I think we still need to respect the `skipBlankLines` option here: if it is set to false, we should only consider the previous row, and not search upward for the previous *non-blank* row. The reason is that when inserting a newline by pressing `enter`, we want to avoid altering the indentation based on 'far-away' lines of the buffer.

For example, suppose that somewhere above the cursor, there's a line whose indentation doesn't match the tab width. This happens:

indent

Even though the 'correct' indentation is being suggested, based on the settings, it can feel unintuitive to the user because they're just trying to insert a newline.

Thinking about alternative ways to make your tests pass...

svanharmelen: Pff... This one doesn't come easy right :disappointed:

Having yet another look...

svanharmelen:
@maxbrunsfeld it took me an hour or so, but I think I found a (again better) solution :smiley: The current code seem to be awfully close the the original code (which is of course a good thing) but it still brings the same improvements as in the earlier versions.

So please have yet another look and run all the manual and automatic tests you need to confirm this is a good fix that doesn't have any nasty side effects.

svanharmelen:
@maxbrunsfeld I worked (quite intensively) with the updated code today, and it seems to work perfectly. So if you could have another look that would be great :smiley:

Thx!

svanharmelen:
@thomasjo @maxbrunsfeld I know you guys are super busy, but it would make me very happy if we can continue the discussion/review on this PR as I believe the current code is rock solid and it would be great if one of you could confirm that so we could move on to getting this merged.

Thanks again guys!

maxbrunsfeld:
@svanharmelen Yeah, this looks great to me. I haven't had time to merge and test locally, but I definitely want to merge this. We're at Codeconf for the rest of this week, and we shipped 1.0 today, so I'll have to do this next week. Thanks for staying with this so persistently; I think it's going to be a good improvement.

svanharmelen:
Yes, I just noticed the 1.0 release. Really great work so congrats to everyone involved!

And sounds good, let's pick this up next week and see if we can get it in by then. Thx!

svanharmelen:
Please note that the Travis failure has noting to do with the code itself. It seems atom.io was unavailable... Travis was triggers as I rebased my PR on v1.0.0.

svanharmelen:
@maxbrunsfeld it's been a pretty silent week... Any change you could pick this one up again?

Thanks!

maxbrunsfeld:
Sorry @svanharmelen. I forgot to let you know that I was on vacation last week. Testing now...

maxbrunsfeld:
Ok, looks good. I like the way that you cleaned up `::suggestedIndentForTokenizedLineAtBufferRow`. One thing though: I don't want to rename `increaseIndentPattern` to `increaseNextIndentPattern` right now; I'd prefer to leave it as `increaseIndentPattern`. We can always rename it later, separately from this PR. Would you mind removing the code for dealing w/ the new config setting name?

svanharmelen:
@maxbrunsfeld sure thing, I'll update the PR shortly so we keep using `increaseIndentPattern` just like TextMate does.

And to be better inline with the naming TextMate uses, I will also rename `decreaseNextIndentPattern` to `dedentNextLinePattern` (matching it's existing counter part `indentNextLinePattern`).

Will update as soon as https://github.com/atom/language-go/pull/47 is merged so the tests can use the update name to confirm all is still good and we can get this one merged in :smiley:

joefitzgerald:
Kudos to @svanharmelen for the persistence and patience on this!

svanharmelen:
@maxbrunsfeld just updated the PR like discussed yesterday...

maxbrunsfeld:
> And to be better inline with the naming TextMate uses, I will also rename `decreaseNextIndentPattern` to `dedentNextLinePattern` (matching it's existing counter part `indentNextLinePattern`).

I'm not sure if makes sense to adopt that naming convention. TextMate's `indentNextLine` is specifically for the case where you only want to indent *one* line, not all subsequent lines. I actually thought `decreaseNextIndentLinePattern` was more consistent with our other setting names.

/cc @nathansobo what do you think about naming here?

svanharmelen:
@maxbrunsfeld you could have saved me some time and effort if you would have told me this yesterday after I suggested the change, instead of now after I updated `language-go` and this PR.

I'm fine either way so please let me know how you want it and I'll make it so...

nathansobo:
Yeah, please keep in mind that we're dealing with hundreds of existing language bundles. The value of changes, even if the names are more optimal, needs to be weighed against the cost of forcing everyone to change or the potential confusion caused by supporting two options that do the same exact thing. We need to be conservative with changes and find ways to blend in with what exists the best we can. I wish that weren't the case, and sometimes it will be worth it to make a breaking change. In the case of these regex names I don't think it is.

nathansobo:
So, to clarify, I think `decreaseNextIndentPattern` fits most cleanly into our existing naming scheme. I agree that `increaseIndentPattern` would be better if it were called `increaseNextIndentPattern`, but I don't think it's worth the cost of changing it on an ecosystem-wide level.

svanharmelen:
@nathansobo I'm fully aware of the impact and agree it's important to get the naming right... So I'm perfectly fine with having this discussion :smiley:

Will update and rebase the PR in a few...

maxbrunsfeld:
Alright. This thing's done. :fork_and_knife:

Thanks for sticking with it @svanharmelen.

thomasjo:
Great job @svanharmelen!

svanharmelen:
@maxbrunsfeld thank you too for sticking with me as well :wink:

kevinsawicki:
This was just released in Atom 1.0.2, thanks @svanharmelen :+1:

svanharmelen:
Cool :smiley:

One thing left I assume... Documenting the new `decreaseNextIndentPattern` that can be used from now on in the language bundles. Any pointers on how and where to do that @maxbrunsfeld @kevinsawicki ?

thomasjo:
@svanharmelen Ideally in the *Flight Manual* at least, specifically in https://github.com/atom/docs/blob/master/book/03-hacking-atom/sections/xx-language.asc, but unfortunately that section was never finished it seems like.

Maybe you feel like taking on the challenge of beginning to document the language grammar support in Atom? That'd be amazing :sparkles:

svanharmelen:
Uuh... Hmm... Think I'll have to think about that one :wink:

But thanks for the info!

PR #6813, opened 2015-05-15
Initial paths shouldn't be normalized on save and restore (could contain uris)

mostafaeweda:
Including url-style paths like `remote://host:port/some/path` in the `initialPaths` causes state loss issues, because the hash creation in: https://github.com/atom/atom/blob/master/src/atom.coffee#L93 creates a key that's different from the key got when restoring the state, because the `initialPaths` are normalized and not existing (see: https://github.com/atom/atom/blob/master/src/browser/atom-application.coffee#L372).

So, when the initialPaths contain a url, the state is never restored

This diff makes sure that project paths and `initialPaths` aren't normalized when not local paths

@bolinfest @maxbrunsfeld @benogle

### DIFF
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -227,3 +227,13 @@ describe "Starting Atom", ->
                 [tempDirPath]
                 [otherTempDirPath]
               ].sort()
+
+  describe "opening a remote directory", ->
+    it "opens the parent directory and creates an empty text editor", ->
+      remoteDirectory = 'remote://server:3437/some/directory/path'
+      runAtom [remoteDirectory], {ATOM_HOME: atomHome}, (client) ->
+        client
+          .waitForWindowCount(1, 1000)
+          .waitForExist("atom-workspace", 5000)
+          .treeViewRootDirectories()
+          .then ({value}) -> expect(value).toEqual([remoteDirectory])
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -522,9 +522,8 @@ class AtomApplication
     new AtomWindow({bootstrapScript, @resourcePath, exitWhenDone, isSpec, specDirectory, devMode})
 
   locationForPathToOpen: (pathToOpen) ->
-    {protocol} = url.parse(pathToOpen)
-    return {pathToOpen} if protocol?
     return {pathToOpen} unless pathToOpen
+    return {pathToOpen} if url.parse(pathToOpen).protocol?
     return {pathToOpen} if fs.existsSync(pathToOpen)
 
     pathToOpen = pathToOpen.replace(/[:\s]+$/, '')
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -369,7 +369,12 @@ class AtomApplication
   #   :windowDimensions - Object with height and width keys.
   #   :window - {AtomWindow} to open file paths in.
   openPaths: ({pathsToOpen, pidToKillWhenClosed, newWindow, devMode, safeMode, apiPreviewMode, windowDimensions, profileStartup, window}={}) ->
-    pathsToOpen = (fs.normalize(pathToOpen) for pathToOpen in pathsToOpen)
+    pathsToOpen = pathsToOpen.map (pathToOpen) ->
+      if fs.existsSync(pathToOpen)
+        fs.normalize(pathToOpen)
+      else
+        pathToOpen
+
     locationsToOpen = (@locationForPathToOpen(pathToOpen) for pathToOpen in pathsToOpen)
 
     unless pidToKillWhenClosed or newWindow
@@ -517,6 +522,8 @@ class AtomApplication
     new AtomWindow({bootstrapScript, @resourcePath, exitWhenDone, isSpec, specDirectory, devMode})
 
   locationForPathToOpen: (pathToOpen) ->
+    {protocol} = url.parse(pathToOpen)
+    return {pathToOpen} if protocol?
     return {pathToOpen} unless pathToOpen
     return {pathToOpen} if fs.existsSync(pathToOpen)
 
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -1,6 +1,7 @@
 {Directory} = require 'pathwatcher'
 fs = require 'fs-plus'
 path = require 'path'
+url = require 'url'
 
 module.exports =
 class DefaultDirectoryProvider
@@ -21,7 +22,13 @@ class DefaultDirectoryProvider
     else
       projectPath
 
-    new Directory(directoryPath)
+    # TODO: Stop normalizing the path in pathwatcher's Directory.
+    directory = new Directory(directoryPath)
+    if (url.parse(directoryPath).protocol)
+      directory.path = directoryPath;
+      if (fs.isCaseInsensitive())
+        directory.lowerCasePath = directoryPath.toLowerCase();
+    directory
 
   # Public: Create a Directory that corresponds to the specified URI.
   #
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -25,9 +25,9 @@ class DefaultDirectoryProvider
     # TODO: Stop normalizing the path in pathwatcher's Directory.
     directory = new Directory(directoryPath)
     if (url.parse(directoryPath).protocol)
-      directory.path = directoryPath;
+      directory.path = directoryPath
       if (fs.isCaseInsensitive())
-        directory.lowerCasePath = directoryPath.toLowerCase();
+        directory.lowerCasePath = directoryPath.toLowerCase()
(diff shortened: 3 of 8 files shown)


maxbrunsfeld:
I think I would rather avoid normalizing paths if they are not local:

```coffee
for pathToOpen, i in pathsToOpen
  if fs.existsSync(pathToOpen)
    pathsToOpen[i] = fs.normalize(pathToOpen)
```

I don't think we should omit any project paths from hash, because then you might do `atom path-1`, and end up with a window that has `path-1`, `remote://path-2`, `remote://path-3`.

mostafaeweda:
@maxbrunsfeld agree and changed the code accordingly.
If you approve the required changes, I'll take out the TODO `pathwacther` directory change into its own PR and update the `pathwatcher` version here.

src/browser/atom-application.coffee
```diff
@@ -517,6 +522,8 @@ class AtomApplication
     new AtomWindow({bootstrapScript, @resourcePath, exitWhenDone, isSpec, specDirectory, devMode})
 
   locationForPathToOpen: (pathToOpen) ->
+    {protocol} = url.parse(pathToOpen)
+    return {pathToOpen} if protocol?
```
kevinsawicki: I think this should be moved below the ` return {pathToOpen} unless pathToOpen` check since `url.parse` blows up with a `null` or `undefined` `pathToOpen`

src/default-directory-provider.coffee
```diff
@@ -21,7 +22,13 @@ class DefaultDirectoryProvider
     else
       projectPath
 
-    new Directory(directoryPath)
+    # TODO: Stop normalizing the path in pathwatcher's Directory.
+    directory = new Directory(directoryPath)
+    if (url.parse(directoryPath).protocol)
```
kevinsawicki: :fire: Remove the `()`

src/default-directory-provider.coffee
```diff
@@ -21,7 +22,13 @@ class DefaultDirectoryProvider
...
+      directory.path = directoryPath
+      if (fs.isCaseInsensitive())
```
kevinsawicki: :fire: Remove the `()`

kevinsawicki:
Would be great cover these changes with tests before merging this.

src/browser/atom-application.coffee
mostafaeweda: done, and looks cleaner as one-liner

src/default-directory-provider.coffee
mostafaeweda: done

src/default-directory-provider.coffee
mostafaeweda: done

mostafaeweda:
@kevinsawicki added relevant tests in `window-spec.js` and `default-directory-provider-spec.coffee` - but couldn't find `application-spec.coffee` for the changes in `atom-application.coffee`

kevinsawicki:
> but couldn't find application-spec.coffee for the changes in atom-application.coffee

`atom-application` stuff is typically tested via the new integration spec, https://github.com/atom/atom/blob/ba98013eebcbe590820a6159fabd98338ca4fc6a/spec/integration/startup-spec.coffee

mostafaeweda:
@kevinsawicki I added an integration test, but all integration tests are failing for me either ways - which got me a bit frustrated. I hope the travis CI gets it or if you could help run the integration test to verify it works.

maxbrunsfeld:
@mostafaeweda I would actually say don't worry about adding an integration test for this. If I understand correctly, the purpose of your change is to fix problems that occur when using packages that provide custom `directory-provider` services. Without such packages, opening a URL like `remote://server/...` is not something you would normally do. So I would say this change is OK as long as it does not break any existing integration tests.

Currently though, it *does* break existing integration tests. This is the output on my machine, running these tests:

```
Starting Atom
  opening a new file
    it opens the parent directory and creates an empty text editor
      Expected false to be true.
        at WebdriverIO.<anonymous> (/Users/maxbrunsfeld/github/atom/spec/integration/helpers/start-atom.coffee:90:26)
        at /Users/maxbrunsfeld/github/atom/build/node_modules/webdriverio/lib/utils/PromiseHandler.js:85:52
        at process._tickCallback (node.js:357:13)
      Expected false to be true.
        at WebdriverIO.<anonymous> (/Users/maxbrunsfeld/github/atom/spec/integration/startup-spec.coffee:37:44)
        at /Users/maxbrunsfeld/github/atom/build/node_modules/webdriverio/lib/utils/PromiseHandler.js:85:52
        at process._tickCallback (node.js:357:13)

...

    it opens the file to the specified line number and column number
      Expected 0 to be 1.
        at WebdriverIO.<anonymous> (/Users/maxbrunsfeld/github/atom/spec/integration/startup-spec.coffee:92:34)

....

  when there is already a window open
    it reuses that window when opening files, but not when opening directories
      Expected false to be true.
        at WebdriverIO.<anonymous> (/Users/maxbrunsfeld/github/atom/spec/integration/helpers/start-atom.coffee:90:26)
        at /Users/maxbrunsfeld/github/atom/build/node_modules/webdriverio/lib/utils/PromiseHandler.js:85:52
        at process._tickCallback (node.js:357:13)
      Expected false to be true.
        at WebdriverIO.<anonymous> (/Users/maxbrunsfeld/github/atom/spec/integration/helpers/start-atom.coffee:90:26)
        at /Users/maxbrunsfeld/github/atom/build/node_modules/webdriverio/lib/utils/PromiseHandler.js:85:52
        at process._tickCallback (node.js:357:13)
```

I can help you with debugging these failures. You said all integration specs fail when you run them on your machine? Can you share some of the output?

mostafaeweda:
I was getting different kind of outputs from the integration tests, sometimes, I get the expected stuff failures, and other times, chrome driver timeout & most of the time: chrome driver io already in use error.

Fortunately, @maxbrunsfeld helped me debug this & killing the chromedriver stale processes.
Then, I could reproduce the integration test bug and fix it - thanks @maxbrunsfeld :+1:

spec/integration/startup-spec.coffee
```diff
@@ -227,3 +227,13 @@ describe "Starting Atom", ->
...
+
+  describe "opening a remote directory", ->
+    it "opens the parent directory and creates an empty text editor", ->
+      remoteDirectory = 'remote://server:3437/some/directory/path'
+      runAtom [remoteDirectory], {ATOM_HOME: atomHome}, (client) ->
+        client
+          .waitForWindowCount(1, 1000)
+          .waitForExist("atom-workspace", 5000)
+          .treeViewRootDirectories()
+          .then ({value}) -> expect(value).toEqual([remoteDirectory])
```
maxbrunsfeld: I'm thinking maybe we should just remove this test, unless you think it's testing something important. What exactly is it testing?

kevinsawicki: When I saw this PR included changes to how `atom-application.coffee` behaves I suggested adding an integration test since those are the only tests for that code currently.

maxbrunsfeld:
It's too bad that we do path normalization in so *many* different places, so that you had to change this in so many places. I'd really like to refactor the code such that the path stays un-normalized for as long as possible, giving the `Project` and the `DirectoryProvider` objects total control over how the paths are treated.

For now, though, I think this is the right thing to do. Except for my question about the integration test you added, this looks good to me.

spec/integration/startup-spec.coffee
mostafaeweda: This is testing that the remote project path survives as a uri, without any other code normalizing it on the way.

maxbrunsfeld: I see your point, but I don't know if the asserted behavior makes sense. It's asserting that the tree view's "root directory" is `remote://something`, even though there is no such directory, and there is no directory provider for remote URLs. It seems like we've gotten the editor into a weird state here.

mostafaeweda: That weird state was already the case, and it was handled by us later when our package is activated to replace the root directory with its remote version.
All what this PR guarantees is that the path survives normalization, but doesn't fix the fact that Atom core initializes and deserializes state before the third party packages get a say in how that should happen.

However, I'd be happy to help in a following PR that does something like what @bolinfest suggested before:

**Either:**
When serializing the project state, include which directory provider (by names) created this directory and when reloading the state, wait for that specific directory provider to register and create the directories at that time

**OR:**
Add a function to directory providers:
```js
handlePath(path: string): boolean
```
where that would be in `default-directory-provider.coffee`:
```coffee
handlePath(directoryPath) ->
  fs.existsSync(directoryPath) or fs.existsSync(path.dirname(directoryPath))
```
and in remote directory provider:
```coffee
handlePath(directoryPath) ->
  directoryPath.startsWith('remote://')
```

maxbrunsfeld: > Atom core initializes and deserializes state before the third party packages get a say in how that should happen.

I see. You're right; this test seems useful :+1:.

PR #6843, opened 2015-05-18
Bundle the hyperclick package

oclbdk:
This will allow 3rd parties to start creating packages to allow users to perform actions by clicking on text.

Note the BSD-3-Clause license and patent grants.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "git-diff": "0.55.0",
     "go-to-line": "0.30.0",
     "grammar-selector": "0.47.0",
+    "hyperclick": "0.1.4",
     "image-view": "0.54.0",
     "incompatible-packages": "0.24.0",
     "keybinding-resolver": "0.32.0",


maxbrunsfeld:
@oclbdk Thanks for starting up this discussion in an issue. We like what you've done with hyperclick so far, but we aren't ready to bundle it with Atom yet.

Here are some we would need you to do:
* [ ] Extract the `ctags` logic from `symbols-view` into a new package (called `hyperclick-ctags` or something) that provides a `hyperclick.provider` service. Our team uses this ctags functionality every day while developing Atom. This will allow us to start using hyperclick instead, so that we can start to find any bugs, missing features or performance issues with the package.
* [ ] Show us some other of your open-source packages that use hyperclick.

Some things that your team and the @atom/core team need to discuss together:
* [ ] Consider moving the ability to define mouse event handlers into Atom core, so that other packages besides hyperclick can do things like this using Atom's command system.
* [ ] Add a way of selecting a suggestion when there are multiple suggestions for the same symbol. This should work like Atom's other select lists.

And some for the Atom team to do:
* [ ] Discuss with our legal team whether we can bundle a package w/ hyperclick's license and patent grants.
* [ ] Discuss if we have any issues w/ bundling a package that uses Flow and Babel.

These lists are a work in progress. @atom/core please feel free to add or adjust items.

thedaniel:
I added a list item for moving the package repo into @atom.

oclbdk:
Thanks for the comments! I'll get working on the features (i.e. ctags provider, selection for multiple suggestions).

If this code gets moved into @atom, we could probably use the MIT license. @jamesgpearce @bolinfest

oclbdk:
After talking with a legal representative from Facebook, we should be able to use the MIT license.

kevinsawicki:
Closing this out for now, will reopen once this is closer to a shippable state.

PR #6847, opened 2015-05-18
Favor installed packages when breaking grammar score ties

kevinsawicki:
If two grammars are both registered for the same file type, use the one from the non-bundled package.

This allows bundled grammars to be easily overridden by installed packages.

This takes away the stress from adding/removing more file types to core grammars in fear that they will break installed packages.

Basically installed packages will always win.

Refs https://github.com/atom/language-python/pull/61
Refs https://github.com/atom/language-php/pull/50
Refs https://github.com/atom/language-xml/issues/23

/cc @atom/feedback 
/cc @lee-dohm @thomasjo

### DIFF
--- a/spec/fixtures/packages/package-with-rb-filetype/grammars/rb.cson
+++ b/spec/fixtures/packages/package-with-rb-filetype/grammars/rb.cson
@@ -1,0 +1,11 @@
+'name': 'Test Ruby'
+'scopeName': 'test.rb'
+'fileTypes': [
+  'rb'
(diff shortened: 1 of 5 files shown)


thomasjo:
Awesome! :sparkling_heart:

50Wliu:
Fantastic :grinning:

lee-dohm:
Looks good to me :+1:

nathansobo:
Looks good.

PR #6848, opened 2015-05-18
Upgrade Bootstrap and fix tooltip arrow positioning

### LINKED ISSUE #5009: Tooltip arrow wrongfully positioned
This happens if the file system pane is closed.
hnnho9zjh6wvyz_y-d5uheutchljuloo-3etj22umfk

mnquintana:
Fixes https://github.com/atom/atom/issues/5009

screenshot 2015-05-19 10 05 55

This PR migrates Atom to use the latest source Bootstrap instead of Atom's outdated fork of Bootstrap. There's really no need for Atom's fork anymore because the `viewport` option was added to source Bootstrap awhile ago, and that's all we were using the fork for anyway. [Source Bootstrap fixed the tooltip arrow positioning issue awhile ago too.](https://github.com/twbs/bootstrap/commit/4c9850701069954375b19e19bd1513654130eacf)

_Note: this doesn't mean Atom won't still be moving away from Bootstrap eventually, but Atom does heavily rely on it for such components as tooltips, so upgrading it is still a good idea._

TODO:
- [x] Figure out why bootstrap.less is overriding atom.less for the tooltips' `font-family` (ie. the tooltips are always Helvetica Neue because they're defined to be in bootstrap.less)
- [x] Settings View
  - [x] Fix checkboxes
- [x] Fix padding to left of tabs
- [x] Check Styleguide for other regressions

/cc @atom/feedback @simurai

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "atom-keymap": "^5.1.2",
     "atom-space-pen-views": "^2.0.4",
     "babel-core": "^5.1.11",
-    "bootstrap": "git+https://github.com/atom/bootstrap.git#6af81906189f1747fd6c93479e3d998ebe041372",
+    "bootstrap": "^3.3.4",
     "clear-cut": "^2.0.1",
     "coffee-cash": "0.8.0",
     "coffee-script": "1.8.0",
--- a/static/bootstrap-overrides.less
+++ b/static/bootstrap-overrides.less
@@ -26,3 +26,9 @@ body {
   font-family: inherit; // inherit from html
   font-size: inherit; // inherit from html
 }
+
+// Latest Bootstrap specifies the font properties again instead of inheriting
+.tooltip {
+  font-family: @font-family;
+  font-size: @font-size;
+}


mnquintana:
Settings View is the only thing broken by this – I think we'll need to update some of the structure / styling there to be Bootstrap-compliant.

simurai:
Added a PR for the checkboxes. Otherwise haven't seen any other surprises.

lee-dohm:
I've played around with it. I can't find anything other than the checkboxes in the Settings View.

mnquintana:
@atom/feedback This should be ready for review now – we'll need to pull in https://github.com/atom/settings-view/pull/508 once this is in.

lee-dohm:
Looks good to me. :+1:

mnquintana:
@kevinsawicki Should we hold off till after the next release to merge just in case?

kevinsawicki:
> Should we hold off till after the next release to merge just in case?

I think so :+1:

mnquintana:
Hmm two specs are failing, but one is [definitely unrelated](https://travis-ci.org/atom/atom/jobs/63097455#L252-L255) and [the other is probably unrelated](https://travis-ci.org/atom/atom/jobs/63097455#L269-L285).

PR #6849, opened 2015-05-18
Use bash highlighting instead of bat

winstliu:
Last I checked, Git Shell uses bash-like syntax, not bat.  This fixes comments not being recognized and instead being parsed (such as the `for` receiving special highlighting).
Also some wording changes (and remove some indentation that's not needed).


kevinsawicki:
:+1: thanks

PR #6877, opened 2015-05-20
Upgrade to Electron and Chrome 42

kevinsawicki:
Updates the build to use electron 0.26 and gets Atom running on Chrome 42
- [ ] Test on Mac
- [ ] Test on Linux
- [ ] Test on Windows

Closes https://github.com/atom/atom/pull/6505


paulcbetts:
Did this actually work for you? How are your native modules not all broken? We had to use `npm shrinkwrap` and hand-edit the shrinkwrap file to force install nan 1.8.4 everywhere

kevinsawicki:
> Did this actually work for you?

Yeah, it ran on my mac.

> How are your native modules not all broken?

Maybe because they all use like this? https://github.com/atom/node-keytar/blob/d801e60eb06a36ccbc0322e56d9e77abcf9cb21e/package.json#L44

paulcbetts:
> Maybe because they all use nan like this? https://github.com/atom/node-keytar/blob/d801e60eb06a36ccbc0322e56d9e77abcf9cb21e/package.json#L44

Yeah, that version of `nan` is _definitely_ busted for us. Weird.

YuriSolovyov:
Looks like there will be [chrome 43](https://github.com/atom/brightray/tree/chrome43), not sure about timing though, but maybe it is better to update straight to that?

izuzak:
Similar PR was started here: https://github.com/atom/atom/pull/6505. Just adding it here as a reference so that we can close that after this is merged. Also, that PR has some other nice changes, e.g. changes to code comments where atom-shell was used as a name to use electron -- might be nice to do that here as well?

YuriSolovyov:
Looks like Chrome 43 electron is [just released](https://github.com/atom/electron/commit/6ee8f6e4f2b66e3ded52390538cb8ef039359c4b)

v3ss0n:
Straight to electron 27 would be nice.

kevinsawicki:
Due to some v8 changes in Electron 0.27 that may break native node modules, the plan is to stay on 0.26 until more is known about which native modules are broken.

See https://github.com/atom/electron/releases/tag/v0.27.0 for more details.

YuriSolovyov:
@kevinsawicki did you tried it or just guessing?

kevinsawicki:
@YuriSolovyov there were some native changes needed in Electron to prevent crashes after upgrading to Chrome 43 and @zcbenz mentioned that native node module could be similarly affected.

zcbenz:
I need some work on Electron side to make snapshot work, so I'll try updating Atom to Chrome 43 in this PR.

@paulcbetts already reported `node-spellcheck` crashes on Chrome 43, I think many of the modules using nan would have the same problem.

zcbenz:
Encountered a Chrome bug:

https://code.google.com/p/chromium/issues/detail?id=495437

v3ss0n:
thats a lot of bugs in chrome 43 then

zcbenz:
I have updated this branch with snapshot support, there is still something not working because of Chrome's bug, but it is enough to test how well snapshot works.

The conclusion is snapshot doesn't save loading time for Atom.

Snapshot boosts startup time by reducing file reading time and JS compilation time to minimum. For file reading time, since Atom is using asar archive, we already have done the best we can, and snapshot itself has its own overhead, it needs to read the huge snapshot file (which is larger than the asar archive) and create a huge object on the heap.

For the JS compilation time, V8 compiles the JS file already very fast, the actual time spent on compilation is quite minor, and since V8 tends to do lazy optimizations when running the code, most time are spent on running the code instead of compiling the code.

So compared to the time saved by snapshot, the overhead of snapshot actually makes things slower.

paulcbetts:
@zcbenz sounds like our princess is in another castle. One idea I had recently was to somehow make a lazy version of `require`, or somehow defer all of those imports that always happen at the top of a file but aren't necessarily used immediately (or at all!) in the code until later

zcbenz:
> One idea I had recently was to somehow make a lazy version of require, or somehow defer all of those imports that always happen at the top of a file but aren't necessarily used immediately (or at all!) in the code until later

Some modules might do some initialization work when being "require"d, one example is graceful-fs. But generally I think it is a good idea.

IMO the key to quicker startup is to show the initial page as quick as possible, it is hard to make things faster if we have to do lots of computations or IO operations before actually showing the content.

thomasjo:
I'm contemplating picking this up and continuing the effort started by @kevinsawicki and @zcbenz. Anything in particular I should know that has not been discussed here?

despairblue:
This would be cool, this would allow opening links under linux with chrome again (https://github.com/atom/electron/issues/1348).

kevinsawicki:
Closing in favor of #7877

PR #6881, opened 2015-05-21
Micro-optimize TokenizedLine::transformContent

nathansobo:
:racehorse: Transforming the tokens and text for display when constructing a tokenized line is a bottleneck when opening large files. Above a certain threshold it needs to be made asynchronous, but this PR has some quick wins for the synchronous path by avoiding a bunch of allocation and unnecessary concatenation.

### DIFF
--- a/src/tokenized-line.coffee
+++ b/src/tokenized-line.coffee
@@ -8,7 +8,25 @@ LeadingWhitespaceRegex = /^\s*/
 TrailingWhitespaceRegex = /\s*$/
 RepeatedSpaceRegex = /[ ]/g
 CommentScopeRegex = /(\b|\.)comment/
+TabCharCode = 9
+SpaceCharCode = 32
+SpaceString = ' '
+TabStringsByLength = {
+  1: ' '
+  2: '  '
+  3: '   '
+  4: '    '
+}
+
 idCounter = 1
+
+getTabString = (length) ->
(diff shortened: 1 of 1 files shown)


src/tokenized-line.coffee
```diff
@@ -8,8 +8,26 @@ LeadingWhitespaceRegex = /^\s*/
...
+TabCharCode = 9
+SpaceCharCode = 32
+SpaceString = ' '
+TabStringsByLength = {
+  1: ' '
+  2: '  '
+  3: '   '
+  4: '    '
+}
+
```
YuriSolovyov: Use an array? Though you will have to either do i+1 lookup or make [0] = '' or something

nathansobo:
Closing this out as the PR it's based on has been reverted. When the reverted code is restored I will include these optimizations as well.

src/tokenized-line.coffee
nathansobo: Seems like a good idea. Any idea what the trade-off is between the two in this situation?

YuriSolovyov: Arrays should be generally faster for numeric loads, trade-off is mostly just style.
ps.
I've posted bench on jsperf http://jsperf.com/array-vs-obj-num-load, though it may be too naive, but it shows no difference in chrome, so looks like V8 is smart enough to use array for object in this case.

PR #6888, opened 2015-05-21
Add lsb_release as Linux dependency

mnquintana:
Refs https://github.com/atom/notifications/pull/64#issuecomment-103884342

This helps ensure that Linux installs have `lsb_release` installed for more accurate OS reporting in issues.


kevinsawicki:
:+1:

PR #6757, opened 2015-05-12
Use 70% less memory for TokenizedLines by using a different representation

nathansobo:
Previously, we naively stored every token on every line as its own object, and made things even worse by storing an array of scopes on every token. This caused `TokenizedLine`s to consume much more memory than is strictly necessary.

This PR switches `TokenizedLine` to a much more compact (but also more cryptic) representation. Instead of repeating the scopes on each token, we instead represent the entire tokenized line as a _nested word_, which intermingles start and end tags with text much like HTML.

Consider the following JavaScript fragment:

``` js
function() {}
```

Previously, we would store this as an array of tokens, each with an array of its containing scopes:

``` js
[
  {value: 'function', scopes: ['source.js', 'meta.function.js', 'storage.type.function.js']},
  {value: ' ', scopes: ['source.js', 'meta.function.js']},
  {value: '(', scopes: ['source.js', 'meta.function.js', 'punctuation.definition.parameters.begin.js']},
  {value: ')', scopes: ['source.js', 'meta.function.js', 'punctuation.definition.parameters.end.js']},
  {value: ' ', scopes: ['source.js', 'meta.function.js']},
  {value: '{', scopes: ['source.js', 'punctuation.section.scope.begin.js']},
  {value: '}', scopes: ['source.js', 'punctuation.section.scope.end.js']}
]
```

Now this is stored more like HTML, with strings intermingled with scope start / end tags.

``` js
// Note: This JSON is just for example purposes. Read below for the actual representation.

[
  {start: 'source.js'},
    {start: 'meta.function.js'},
      {start: 'storage.type.function.js'},
        "function",
      {end: 'storage.type.function.js'},
      " ",
      {start: 'punctuation.definition.parameters.begin.js'},
        "(",
      {end: 'punctuation.definition.parameters.begin.js'},
      {start: 'punctuation.definition.parameters.end.js'},
        ")",
      {end: 'punctuation.definition.parameters.end.js'},
    {end: 'meta.function.js'},
    " ",
    {start: 'punctuation.section.scope.begin.js'},
      "{",
    {end: 'punctuation.section.scope.begin.js'},
    {start: 'punctuation.section.scope.end.js'},
      "}",
    {end: 'punctuation.section.scope.end.js'},
  // note 'source.js' isn't closed yet
]
```

[...]

PR #6892, opened 2015-05-21
Take 2: Use 70% less memory for TokenizedLines

nathansobo:
#6757 contained a couple of regressions that required me to back it out. This should fix both of them. I'm going to merge this after we release today.


nathansobo:
Trying again. Hold onto your :tophat:s. Please ping me if you see anything :fish:y.

PR #6898, opened 2015-05-22
Lint for even more style errors

mnquintana:
The sequel to https://github.com/atom/atom/pull/6230

I'm sure making the package linting job harder for myself :stuck_out_tongue_closed_eyes:, but I figured I should do this now before I add linting support to any more packages – this updates coffeelint.json to lint for more code style errors.

Let me know if these changes seem reasonable!

/cc @atom/feedback

PR #6907, opened 2015-05-22
Skip linting fixtures

mnquintana:
Refs #6898, https://github.com/atom/atom/issues/5982, and https://github.com/atom/atom/pull/6230#discussion_r27896062

This adds a `.coffeelintignore` file to ensure that no one accidentally lints the `spec/fixtures` directory, lest we break the universe.

### DIFF
--- a/.coffeelintignore
+++ b/.coffeelintignore
@@ -1,0 +1,1 @@
+spec/fixtures


thomasjo:
For future consideration: have a separate `coffeelint.json` for the specs—just add one in the `/spec` directory. I used to have that for my specs when my (primary) package was written in CoffeeScript. I can't remember if it overrides (aka. inherits from the parent) duplicate keys or if takes precedence completely, but it's a decent way to ensure some form of conformity there as well, but being less strict.

mnquintana:
The main reason for this is that a lot of specs are relying on fixtures remaining completely unchanged (especially TextEditor specs) because they make assumptions about what is / should be in each column and row, for instance (which would change when fixing most linter errors). We can't really add any linting for them unless we update the corresponding tests, which would be a pretty big task. It's a lot easier to just ignore them for now.

thomasjo:
Sorry, ignore me! I read *specs* and not *fixtures*. Derp :expressionless:

PR #6910, opened 2015-05-22
:memo: Document command name standard

### LINKED ISSUE #6893: Multiple `:` colons in commands
I can't find any docs that says that using more then one `:` in a command isn't allowed. 
Executing a command with multiple colons works as expected, however the [Command Palette](https://github.com/atom/command-palette) doesn't show the text after the second colon.

**Example commands:**
[`tool-bar:position:top`](https://github.com/suda/tool-bar/blob/master/lib/tool-bar-view.coffee#L34)
[`tool-bar:position:right`](https://github.com/suda/tool-bar/blob/master/lib/tool-bar-view.coffee#L37)
[`tool-bar:position:bottom`](https://github.com/suda/tool-bar/blob/master/lib/tool-bar-view.coffee#L40)
[`tool-bar:position:left`](https://github.com/suda/tool-bar/blob/master/lib/tool-bar-view.coffee#L43)

**Screenshot Command Palette:**
[tool-bar command palette](https://github.com/suda/tool-bar/issues/40)

**Trace:**
After some research I traced it down to Underscore Plus [`humanizeEventName`](https://github.com/atom/underscore-plus/blob/master/src/underscore-plus.coffee#L142), which then only takes the first and second parts.

**Specs:**
Windows 7 64-bits
Atom 0.200.0
tool-bar 0.1.1

thomasjo:
Not entirely sure about the wording, and location of the paragraph, but perhaps not too bad?

/cc @atom/feedback 

---

Resolves #6893

### DIFF
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -22,6 +22,7 @@ SequenceCount = 0
 # will typically be the name of your package, and `action` describes the
 # behavior of your command. If either part consists of multiple words, these
 # must be separated by hyphens. E.g. `awesome-package:turn-it-up-to-eleven`.
+# All words should be lowercased.
 #
 # As the event bubbles upward through the DOM, all registered event listeners
 # with matching selectors are invoked in order of specificity. In the event of a
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -18,10 +18,10 @@ SequenceCount = 0
 # command event listeners globally on `atom.commands` and constrain them to
 # specific kinds of elements with CSS selectors.
 #
-# Command names must follow the `namespace:event` pattern, where both parts can
-# be further split into words by hyphens. Typically `namespace` will be the
-# name of your package, and `event` describes the behavior of your command.
-# E.g. `awesome-package:toggle`.
+# Command names must follow the `namespace:action` pattern, where `namespace`
+# will typically be the name of your package, and `action` describes the
+# behavior of your command. If either part consists of multiple words, these
+# must be separated by hyphens. E.g. `awesome-package:turn-it-up-to-eleven`.
 #
 # As the event bubbles upward through the DOM, all registered event listeners
 # with matching selectors are invoked in order of specificity. In the event of a
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -20,7 +20,7 @@ SequenceCount = 0
 #
 # Command names must follow the `namespace:event` pattern, where both parts can
 # be further split into words by hyphens. Typically `namespace` will be the
-# name of your package, and `event` the name of your command.
+# name of your package, and `event` describes the behavior of your command.
 # E.g. `awesome-package:toggle`.
 #
 # As the event bubbles upward through the DOM, all registered event listeners
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -18,6 +18,11 @@ SequenceCount = 0
 # command event listeners globally on `atom.commands` and constrain them to
 # specific kinds of elements with CSS selectors.
 #
+# Command names must follow the `namespace:event` pattern, where both parts can
+# be further split into words by hyphens. Typically `namespace` will be the
+# name of your package, and `event` the name of your command.
+# E.g. `awesome-package:toggle`.
+#
 # As the event bubbles upward through the DOM, all registered event listeners
 # with matching selectors are invoked in order of specificity. In the event of a
 # specificity tie, the most recently registered listener is invoked first. This


src/command-registry.coffee
```diff
@@ -18,6 +18,11 @@ SequenceCount = 0
 # command event listeners globally on `atom.commands` and constrain them to
 # specific kinds of elements with CSS selectors.
 #
+# Command names must follow the `namespace:event` pattern, where both parts can
+# be further split into words by hyphens. Typically `namespace` will be the
+# name of your package, and `event` the name of your command.
+# E.g. `awesome-package:toggle`.
```
mnquintana: I think it would be good to have the `event` name example also be something separated by hyphens.

src/command-registry.coffee
```diff
@@ -18,6 +18,11 @@ SequenceCount = 0
 # command event listeners globally on `atom.commands` and constrain them to
 # specific kinds of elements with CSS selectors.
 #
+# Command names must follow the `namespace:event` pattern, where both parts can
+# be further split into words by hyphens. Typically `namespace` will be the
+# name of your package, and `event` the name of your command.
```
mnquintana: This could potentially be confusing because we consider the whole `namespace:event` to be the command name

src/command-registry.coffee
thomasjo: Yeah, I was thinking the same. Hmm, any suggestions?

src/command-registry.coffee
thomasjo: Yeah, I was thinking the same. Hmm, any suggestions?

src/command-registry.coffee
thomasjo: My thought behind leaving one of them un-hyphenated was to indicate that it's optional, but perhaps that's already obvious?

src/command-registry.coffee
mnquintana: Maybe ``and `event` describes the behavior of your command``

src/command-registry.coffee
mnquintana: Oh that's a good point – I think it's fine as-is then :+1:

src/command-registry.coffee
thomasjo: I like that! New commit incoming.

src/command-registry.coffee
```diff
@@ -18,6 +18,11 @@ SequenceCount = 0
 # command event listeners globally on `atom.commands` and constrain them to
 # specific kinds of elements with CSS selectors.
 #
+# Command names must follow the `namespace:event` pattern, where both parts can
+# be further split into words by hyphens. Typically `namespace` will be the
```
lee-dohm: I would go ahead and make the same **strong** recommendation that `namespace` *must* be the dasherized version of your package name.

src/command-registry.coffee
```diff
@@ -18,6 +18,11 @@ SequenceCount = 0
...
+# name of your package, and `event` describes the behavior of your command.
+# E.g. `awesome-package:toggle`.
```
lee-dohm: I'd also select a multi-word command name to drive the point home that both components should be dasherized.

src/command-registry.coffee
mnquintana: :+1: We should probably also enforce that in code somewhere too, but that's a good idea.

(It would be pretty jarring if you installed `my-awesome-package` and all the commands were namespaced as `settings-view` :stuck_out_tongue_closed_eyes:)

lee-dohm: Well, like configuration settings ... I don't think we want to make it an absolute requirement. I prefer to leave some things as "Ok, if you need to break the rules ... that's fine, but it's ice skating uphill".

thomasjo: OK, what about something like 

> Command names must follow the `namespace:event` pattern, where `namespace` will typically be the name of your package, and `event` describes the behavior of your command. If either part consists of multiple words, these must be separated by hyphens.
E.g. `awesome-package:turn-it-up-to-eleven`.

lee-dohm: :+1:

nathansobo:
What do you think about `action` as opposed to `event`? It's more active, indicating that a command is an action initiated by the user rather than an event that's emitted by the system.

thomasjo:
> What do you think about action as opposed to event?

That seems to make sense! My original naming came straight out of `underscore-plus`, not very attached to them :smile:

src/command-registry.coffee
mnquintana: What about:

> Command names must follow the `namespace:event` pattern, where `namespace` **is** the name of your package and `event` describes the behavior of your command. If either part consists of multiple words, these must be separated by hyphens.
E.g. `awesome-package:turn-it-up-to-eleven`.

mnquintana: Should we also make a note about preferring lowercase / case sensitivity or something?

thomasjo: 98% of the time that's likely fine. But only a few lines further down there's a reference to a user-defined command example `user:insert-date`, so I didn't want to be overly restrictive here, since there are valid scenarios where it's **not** tied to a package. Hmm...

mnquintana: Okay, I think that'd be fine then.

thomasjo:
I think @mnquintana suggestion regarding lowercase command names is good. Should I add it before, or after the remark about word hyphenation? Perhaps somewhere else, or inlined with different wording?

mnquintana:
How about:

> ...If either part consists of multiple words, these must be separated by hyphens (e.g. `awesome-package:turn-it-up-to-eleven`). **All words should be lowercased.**

mnquintana:
Looks good to me :shipit:

PR #6942, opened 2015-05-25
Document build flags

mnquintana:
Refs https://discuss.atom.io/t/install-atom-as-standard-user-on-os-x/17142/2?u=mnquintana

`script/build`'s option flags are currently undocumented for every platform but Linux – this documents it for other platforms.

/cc @atom/feedback

### DIFF
--- a/docs/build-instructions/freebsd.md
+++ b/docs/build-instructions/freebsd.md
@@ -19,4 +19,18 @@ FreeBSD -RELEASE 64-bit is the recommended platform.
   sudo script/grunt install # Installs command to /usr/local/bin/atom
   ```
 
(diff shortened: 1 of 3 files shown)


lee-dohm:
On OS X, the `--install-dir` parameter is a bit of a misnomer. What you should really pass to this parameter on OS X is the full path to the application package you want to create: `foo/bar/Atom.app`. (Technically, this application package is just a directory with some special stuff on it, but still.)

mnquintana:
Huh, that's a bit strange – I'd like to update that for OS X at some point to make it more consistent with other platforms, but for now, :memo: it is good.

PR #6946, opened 2015-05-26
Don't save `null` repositories

akonwi:
Putting instances of null in the `@repositories` array can be misleading for packages that want to use `atom.project.getRepositories().length` to know how many repos there actually are instead of how many potential repositories there could be. Right now, I'm filtering the array for non-null values each time I call it.

### DIFF
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -203,7 +203,7 @@ class Project extends Model
     repo = null
     for provider in @repositoryProviders
       break if repo = provider.repositoryForDirectorySync?(directory)
-    @repositories.push(repo ? null)
+    @repositories.push(repo) if repo
 
     unless options?.emitEvent is false
       @emit "path-changed" if includeDeprecatedAPIs


maxbrunsfeld:
@akonwi The reason that `getRepositories` sometimes contains `null` values is that there needs to be a way to identify which project-path a repository corresponds to. Currently, `::getPaths()`, `::getDirectories()` and `::getRepositories()` return parallel arrays.

We know that this API is not ideal. We also want to move toward accessing repository objects *asynchronously* for performance reasons. That's why new code should use `atom.project.repositoryForDirectory(directory)` which takes a `Directory` object as returned from `.getDirectories()`, and returns a `Promise` that resolves with either a `Repository` or `null`.

So unfortunately, we can't change the behavior of `getRepositories()` at this point. Let me know if you have any questions or problems using `repositoryForDirectory`. We're hoping that it will be suitable for every package to use.

akonwi:
@maxbrunsfeld I didn't dig much further into the code but I had a feeling `getDirectories` and `getRepositories` were parallel arrays. `repositoryForDirectory` will work pretty well. 

Thanks!

PR #6957, opened 2015-05-26
Pin the Babel dependency to a specific version number.

bolinfest:
Third-party Atom packages may want to use the exact same version of Babel
that Atom uses to transpile Babel files automatically. Specifying the version
as a range makes this hard to do.


bolinfest:
We'll find a better way to do this.

PR #6958, opened 2015-05-26
Add coffeelint config files to ignored build paths

mnquintana:
This adds `.coffeelintignore` and `coffeelint.json` to the list of ignored paths so we don't ship with them.

Refs https://github.com/atom/atom/issues/5982

/cc @kevinsawicki


kevinsawicki:
:+1:

Issue #6969, opened 2015-05-27
Uncaught TypeError: Parameter 'url' must be a string, not undefined


izuzak:
Steps to reproduce:

1. Hit `cmd-n` to create a new tab with empty text editor.

**Atom Version**: 0.202.0-60fbb1d
**System**: Mac OS X 10.10.3
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Parameter 'url' must be a string, not undefined

```
At url.js:90

TypeError: Parameter 'url' must be a string, not undefined
    at Url.parse (url.js:90:11)
    at Object.urlParse [as parse] (url.js:84:5)
    at EventEmitter.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/src/window-event-handler.js:46:30)
    at emitTwo (events.js:87:13)
    at EventEmitter.emit (events.js:169:7)
```

### Commands

```
     -0:00.6.0 application:new-file (atom-pane.pane.active)
```

### Config

```json
{
  "core": {},
  "editor": {
    "fontSize": 20,
    "invisibles": {}
  }
}
```

### Installed Packages

```coffee
# User
No installed packages

# Dev
No dev packages
```

cc @maxbrunsfeld @kevinsawicki, in case it's related to https://github.com/atom/atom/issues/6968

izuzak:
Looks related to #6813 again -- working for me at 2953d67, but not working at 8dff5bb (throws this exception). cc @mostafaeweda

maxbrunsfeld:
Thanks for catching this @izuzak!

Issue #6968, opened 2015-05-27
Some menu commands are no longer working


izuzak:
As far as I can tell, the following menu commands are no longer working on latest master (0.202.0-60fbb1d) -- clicking them has no effect:

![screen shot 2015-05-27 at 14 09 45](https://cloud.githubusercontent.com/assets/38924/7835538/13644862-047a-11e5-964a-31e90c61d0ae.png)

Other menu commands seem to be working.

@maxbrunsfeld @kevinsawicki @mostafaeweda: my :8ball: says that https://github.com/atom/atom/pull/6813 _might_ be involved -- the menu commands are working for me at 2953d67, but not working at 8dff5bb (which was merged right after 2953d67).

mnquintana:
Confirmed on Atom 0.202.0-8dff5bb and OS X 10.10.3

maxbrunsfeld:
Thanks @izuzak, @mnquintana!

PR #6977, opened 2015-05-27
Revert "Initial paths shouldn't be normalized on save and restore (co…

maxbrunsfeld:
Reverts atom/atom#6813

That PR introduced some bugs and we need to unblock the release of 0.202.

### DIFF
--- a/src/window-event-handler.coffee
+++ b/src/window-event-handler.coffee
@@ -5,7 +5,6 @@ ipc = require 'ipc'
 shell = require 'shell'
 {Subscriber} = require 'emissary'
 fs = require 'fs-plus'
-url = require 'url'
 
 # Handles low-level events related to the window.
 module.exports =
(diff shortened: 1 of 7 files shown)


maxbrunsfeld:
Fixes #6968.
Fixes #6969.

maxbrunsfeld:
@mostafaeweda We can re-merge your PR, with some fixes, after Atom 0.202 is released.

src/window-event-handler.coffee
```diff
@@ -24,13 +23,12 @@ class WindowEventHandler
...
-                atom.project.addPath(path.dirname(pathToOpen))
               else
-                atom.project.addPath(pathToOpen)
+                dirToOpen = path.dirname(pathToOpen)
+                if fs.existsSync(dirToOpen)
+                  atom.project.addPath(dirToOpen)
 
-            {protocol} = url.parse(pathToOpen)
-            unless fs.isDirectorySync(pathToOpen) or protocol?
+            unless fs.isDirectorySync(pathToOpen)
```
maxbrunsfeld: @mostafaeweda The 2 problems we found that this PR introduced were both in this file:

1. `pathToOpen` can be `undefined`, if we're trying to open a new empty editor.
2. If `pathToOpen` is a URI, we still need to open it.

mostafaeweda: 1- will check for the undefined and fix it
2- how about: if the uri starts with `atom://`, then it needs to be opened.
or how do I differentiate between the reload/restart case and the normal open cases?

maxbrunsfeld: I think that we always need to open it. Other packages could define other 'opener` URIs, like `markdown-preview://view`, or whatever. In general, remote URIs could represent files, not just directories.

mostafaeweda: agreed - addressed in: https://github.com/atom/atom/pull/6978

PR #6978, opened 2015-05-27
Initial paths shouldn't be normalized on save and restore [2]

mostafaeweda:
This PR fixed the `undefined` case and not being able to open new files and opening all url-style paths, config, preferences, ..etc.

Including url-style paths like remote://host:port/some/path in the initialPaths causes state loss issues, because the hash creation in: https://github.com/atom/atom/blob/master/src/atom.coffee#L93 creates a key that's different from the key got when restoring the state, because the initialPaths are normalized and not existing (see: https://github.com/atom/atom/blob/master/src/browser/atom-application.coffee#L372).

So, when the initialPaths contain a url, the state is never restored

This diff makes sure that project paths and initialPaths aren't normalized when not local paths

@bolinfest @maxbrunsfeld @benogle

### DIFF
--- a/src/window-event-handler.coffee
+++ b/src/window-event-handler.coffee
@@ -5,6 +5,7 @@ ipc = require 'ipc'
 shell = require 'shell'
 {Subscriber} = require 'emissary'
 fs = require 'fs-plus'
+url = require 'url'
 
 # Handles low-level events related to the window.
 module.exports =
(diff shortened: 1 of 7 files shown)


src/window-event-handler.coffee
```diff
@@ -5,6 +5,7 @@ ipc = require 'ipc'
 shell = require 'shell'
 {Subscriber} = require 'emissary'
 fs = require 'fs-plus'
+url = require 'url'
```
maxbrunsfeld: :fire:

mostafaeweda: :+1:

maxbrunsfeld:
@mostafaeweda Thanks for the quick fixup. I will merge this later today, after we :ship: Atom 0.202.

mostafaeweda:
Great to hear, thanks Max

mostafaeweda:
ping

maxbrunsfeld:
@mostafaeweda since our Travis build is a little flaky right now and does not run the integration tests, could you just confirm that all tests (including integration tests) pass on your machine w/ this change?

mostafaeweda:
Yea, I just double-checked that, including integration tests.

PR #6986, opened 2015-05-28
Add a rootDirectories filtering option to atom.workspace.scan().

bolinfest:
By default, `atom.workspace.scan()` scans everything in `atom.project.getDirectories()`.
This introduces the option to restrict `scan()` to a specific set of `Directory` objects.

This will be useful for excluding special root `Directory` objects (such as those created
via a custom `DirectoryProvider`) from calls to `scan()`, if appropriate.

Because I was in the neighborhood, I also made some small documentation fixes:
- Documented the previously undocumented `onPathsSearched` option.
- Clarified that the returned `Promise` has a `cancel()` method.

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -2,6 +2,7 @@ path = require 'path'
 temp = require 'temp'
 Workspace = require '../src/workspace'
 Pane = require '../src/pane'
+{Directory} = require 'pathwatcher'
 {View} = require '../src/space-pen-extensions'
 platform = require './spec-helper-platform'
 _ = require 'underscore-plus'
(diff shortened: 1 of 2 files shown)


bolinfest:
This change depends on https://github.com/atom/atom/pull/6986.

Also, I'm hoping to get feedback on the API for `DefaultSearchDirectoryProvider` before investing the time to write tests for it. I think all of the methods are necessary: it's mainly a question of whether all of the parameters should be grouped into one object literal named `options`.

bolinfest:
I added a unit test. Please let me know if there is anything else you would like to see tested.

maxbrunsfeld:
Left some comments on https://github.com/atom/find-and-replace/pull/399 that potentially affect this PR too.

bolinfest:
Take 2: https://github.com/atom/atom/pull/7022

bolinfest:
Deleting in favor of https://github.com/atom/atom/pull/7022

PR #6987, opened 2015-05-28
Use previous definition of scope selector match to fix API breakage

nathansobo:
I switched to first-mate Selector because I didn’t want to replicate the poorly-defined Token::matchesScopeSelector method now that tokens are not stored on lines. However, the first-mate semantics are stricter and that broke the API. Perhaps using selector-kit here would be better, but I just wanted to put back exactly to how it was for now.

@kevinsawicki we need to merge this and release ASAP to fix railroad diagrams and probably some other packages. Sorry about this everybody.

/cc @ypresto


kevinsawicki:
Thanks for fixing this :+1:

ypresto:
https://github.com/atom/atom/pull/6757#discussion_r31212258

ypresto:
Thanks!!!

PR #6998, opened 2015-05-28
Style links and code blocks in text-* classes

benogle:
They were unstyled. So this: 

screen shot 2015-05-28 at 4 10 06 pm

to this:

screen shot 2015-05-28 at 4 10 16 pm

Maybe should go in the themes, but I wanted this to be in a central place. What do you think @simurai?

### DIFF
--- a/static/atom.less
+++ b/static/atom.less
@@ -25,5 +25,6 @@
 @import "text-editor-light";
 @import "select-list";
 @import "syntax";
+@import "text";
 @import "utilities";
 @import "octicons";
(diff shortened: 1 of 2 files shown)


benogle:
One with a link: 

screen shot 2015-05-28 at 4 11 21 pm

simurai:
Looks good. Does it work when switching to a dark theme?

I think it's ok to add it to core, just so that all themes will have it. And if some theme authors are unhappy with the styling, they can still change it.

benogle:
Yep, looks good in dark as well:

screen shot 2015-05-29 at 10 32 25 am

Issue #3051, opened 2014-07-23
Make invisibles tab character span multiple spaces


alexandernst:
In other editors I'm used to have a wide character (em dash) representing a tab and while Atom has something like that, it doesn't work quite how I'd expect it to work.

Currently, if you set an em dash in "Invisibles Tab" field and "Tab width" to 4 spaces, the em dash would be painted with the width of a single character. Here is a screenshot showing this:

![alt text](https://virtual-host-discourse.global.ssl.fastly.net/uploads/github_atom/23682/5e3564883ffa487d.png "Bug invisibles tab")

Note that the em dash should have the width of 4 spaces.

Brackets fixed this in a really cool way, VIM has it also. This way, IMHO, should be how this should work:

The "Invisibles Tab" setting should have 2 fields. The first field is for the character that should be used to fill the space of the tab. And the second field is for the character that should finish the drawing (if any).

Practical examples of this setting:

Example A
First character: -
Second character: >
Result for 4 spaces: --->

Example B
First character: -
Second character: -
Result for 4 spaces: ----

Regards

Felicis:
I found a fix, that achieves almost the desired result: 
just use the "LONG ARROW RIGHTWARDS" (0x27f6)  character in the unicode table, which is really wide and comes really near to the wide dash I'm used to: "⟶"
![selection_017](https://cloud.githubusercontent.com/assets/11877488/10863853/39782d68-7fd9-11e5-9f6f-eef7e4331111.png)

Best

alexandernst:
This is indeed a really nice trick!

alexandernst:
Spoke too soon. It's not looking good here: http://ibin.co/2KwgTvtt2TEr

Felicis:
you sure you took the right character?
for me it looks like the closely related simple right arrow...
Best

natbraun:
@alexandernst I'd be happy to work on this issue. However, we need to specify it precisely. Here is a tentative specification. There are several 'kinds' of tabs users could want to use:

A. Mandatory first  character, and repeating secondary character. For the '>' and '-' couple, and for length 1, 2 and 5 the tabs would respectively show like this '>', '>-', '>----'

B. Mandatory last character, and repeating preceding character. For the '-' and '>' couple, and for length 1, 2 and 5 the tabs would respectively show like this '>', '->', '---->'

[...]

PR #7020, opened 2015-05-29
Constrain length of invisible character strings in settings

nathansobo:
We don't currently support invisible character strings that exceed a single character. Previously, rendering went haywire when users specified strings that were longer. This PR disallows that.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -805,7 +805,6 @@ describe "Config", ->
           atom.config.loadUserConfig()
           expect(atom.config.get("foo.bar")).toBe "baz"
 
-
     describe ".observeUserConfig()", ->
       updatedHandler = null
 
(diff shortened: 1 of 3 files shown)


thomasjo:
Nice! :+1:

nathansobo:
Thanks a ton for your help @thomasjo. It was nice to know someone else cared about this and was looking too. :dancers:

alexandernst:
Does this fix https://github.com/atom/atom/issues/3051 ?

nathansobo:
Sorry @alexandernst, it doesn't fix that issue, no. Your idea about 2 fields seems reasonable.

alexandernst:
@nathansobo Oh, :disappointed: For 1 second I thought my issue would finally get fixed. Just out of curiosity, how "reasonable" is my feature request, from "we'll put it on the TODO list for the distant future" to "this should get implemented right after v1.0"?

nathansobo:
Probably in the former category, though we would review a well-written pull request once we get out of crunch mode for 1.0. It's a good idea, but there's a lot of them on the list. Sorry.

Issue #6867, opened 2015-05-20
Atom packages that require an upgrade to 1.0 API


benogle:
We've been collecting data on which packages have deprecations since 0.201. This is a list of the packages using deprecated APIs, their owners and each package's impact. This data reflects a subset of the user population (those who have upgraded to 201). Affected numbers are based on unique users catching the deprecations.

If you are a maintainer on this list please note that we will be removing deprecated APIs on __June 1st__. See the [blog post](http://blog.atom.io/2015/05/01/removing-deprecated-apis.html) for more info. Note that upgrading to the new APIs will decrease breakage, increase speed, and generally be better for everyone. 

### Upgrading your package

We have a few guides for upgrading your package or theme: 

* [Upgrading your package](https://atom.io/docs/latest/upgrading-to-1-0-apis-upgrading-your-package)
* [Upgrading your package styles or UI theme](https://atom.io/docs/latest/upgrading-to-1-0-apis-upgrading-your-ui-theme-or-package-selectors)
* [Upgrading your syntax theme](https://atom.io/docs/latest/upgrading-to-1-0-apis-upgrading-your-syntax-theme)
* If you do only one thing, use [atom-space-pen-views](https://github.com/atom/atom-space-pen-views) for the `View` class rather than importing them from atom. See the [package upgrade guide](https://atom.io/docs/latest/upgrading-to-1-0-apis-upgrading-your-package#upgrading-your-views) for details.
* Not sure what deprecations your package is generating? See https://gist.github.com/benogle/6d09e295c84b717ef9b4 for a list

### Need help upgrading?

The community and the core team can offer some help:

* Post an issue on the [forum](https://discuss.atom.io/)
* Get help in the [slack room](http://atom-slack.herokuapp.com/)
* If you're stuck, post an issue on [atom/atom](https://github.com/atom/atom/issues)

### Package abandoned?

If your package is abandoned or supplanted by something in core, please comment with the package name and we can take steps to remove it from the package listing and people's install.

However, ***do not delete the package repo from GitHub before unpublishing***, as this will make it so that you can't unpublish from the package registry.

### Notes

* We are currently working on ways to minimize pain from deprecations. We will try hard to keep things going smoothly through the transition
* It appears there are many many people still using outdated packages. We have taken steps in v0.200 to be more aggressive about upgrading packages. 
* If you want to tinker with the data, it is all here: https://github.com/benogle/deprecation-data

[...]

PR #7021, opened 2015-05-29
Don't load packages with deprecations

kevinsawicki:
Adds a list of package name and versions not to load since they contain deprecations and will break once the deprecated Atom APIs are removed.

Refs #6867


benogle:
Can I also get a method to get the deprecation info for a package name? `getPackageDeprecationMetadata(name)` or something

kevinsawicki:
> Can I also get a method to get the deprecation info for a package name?

Yup, just pushed an API:

```coffee
metadata = atom.packages.getDeprecatedPackageMetadata('wordcount')
metadata.version # "<=2.1.0"
```

PR #7022, opened 2015-05-29
Introduce atom.directory-searcher service v0.1.0.

bolinfest:
The contract for a provider for the `atom.directory-searcher` service
is defined by the spec of the `DefaultDirectorySearcher`.

This modifies `Workspace::scan()` to use the appropriate `DirectorySearcher`
for each member of `atom.project.getDirectories()` when scanning the workspace
for files that match the specified regex.

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -1,6 +1,5 @@
 path = require 'path'
 temp = require 'temp'
-{Disposable} = require 'event-kit'
 Workspace = require '../src/workspace'
 Pane = require '../src/pane'
 {View} = require '../src/space-pen-extensions'
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -985,6 +985,37 @@ describe "Workspace", ->
               expect(onPathsSearched.mostRecentCall.args[0]).toBe(
                 numPathsToPretendToSearchInCustomDirectorySearcher + numPathsSearchedInDir2)
 
+          it "can be cancelled when the object returned by scan() has its cancel() method invoked", ->
+            lastCustomDirectorySearchCreated = null
+            class CustomDirectorySearch
+              constructor: ->
+                lastCustomDirectorySearchCreated = this
+                @promise = Promise.resolve()
+              then: (args...) ->
+                @promise.then.apply(@promise, args)
+              cancel: ->
+
+            class CustomDirectorySearcher
+              canSearchDirectory: (directory) -> directory.getPath() is dir1
+              search: (directory, delegate, options) ->
+                new CustomDirectorySearch
+
+            atom.packages.serviceHub.provide(
+              "atom.directory-searcher", "0.1.0", new CustomDirectorySearcher())
+
+            thenable = atom.workspace.scan /aaaa/, ->
+            cancelSpy = spyOn(lastCustomDirectorySearchCreated, 'cancel').andCallThrough()
+            expect(cancelSpy).not.toHaveBeenCalled()
+            thenable.cancel()
+            expect(cancelSpy).toHaveBeenCalled()
+
+            resultOfPromiseSearch = null
+            waitsForPromise ->
+              thenable.then (promiseResult) -> resultOfPromiseSearch = promiseResult
+
+            runs ->
+              expect(resultOfPromiseSearch).toBe('cancelled')
+
           it "will have the side-effect of failing the overall search if it fails", ->
             # Note that hoisting reject in this way is generally frowned upon.
             hoistedReject = null
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -1,5 +1,6 @@
 path = require 'path'
 temp = require 'temp'
+{Disposable} = require 'event-kit'
 Workspace = require '../src/workspace'
 Pane = require '../src/pane'
 {View} = require '../src/space-pen-extensions'
@@ -943,9 +944,13 @@ describe "Workspace", ->
             foreignFilePath = 'ssh://foreign-directory:8080/hello.txt'
             numPathsSearchedInDir2 = 1
             numPathsToPretendToSearchInCustomDirectorySearcher = 10
-            class CustomDirectorySearcher
-              canSearchDirectory: (directory) -> directory.getPath() is dir1
-              search: (directory, regexSource, onSearchResult, onSearchError, onPathsSearched, options) ->
+            class CustomDirectorySearch
+              constructor: () ->
+                @promise = Promise.resolve()
+              then: (args...) ->
+                @promise.then.apply(@promise, args)
+              onDidMatch: (callback) ->
+                # Invoke the callback with the only result we plan to return.
                 searchResult1 =
                   filePath: foreignFilePath,
                   matches: [
@@ -956,11 +961,20 @@ describe "Workspace", ->
                       range: [[0, 0], [0, 5]],
                     }
                   ]
-                onSearchResult(searchResult1)
-                onPathsSearched(numPathsToPretendToSearchInCustomDirectorySearcher)
-                promise = Promise.resolve()
-                promise.cancel = ->
-                promise
+                callback(searchResult1)
+                new Disposable
+              onDidError: (callback) ->
+                new Disposable
+              onDidSearchPaths: (callback) ->
+                # Invoke the callback with the one notification we plan to send.
+                callback(numPathsToPretendToSearchInCustomDirectorySearcher)
+                new Disposable
+              cancel: ->
+
+            class CustomDirectorySearcher
+              canSearchDirectory: (directory) -> directory.getPath() is dir1
+              search: (directory, options) ->
+                new CustomDirectorySearch
 
             atom.packages.serviceHub.provide(
               "atom.directory-searcher", "0.1.0", new CustomDirectorySearcher())
@@ -982,16 +996,27 @@ describe "Workspace", ->
 
           it "can be cancelled by cancelling one of the DirectorySearchers", ->
             customDirectorySearcherPromiseInstance = null
+            class CustomDirectorySearchToCancel
+              constructor: () ->
+                # Note that hoisting reject in this way is generally frowned upon.
+                @promise = new Promise (resolve, reject) =>
+                  @hoistedReject = reject
+                customDirectorySearcherPromiseInstance = this
+              then: (args...) ->
+                @promise.then.apply(@promise, args)
+              onDidMatch: (callback) ->
+                new Disposable
+              onDidError: (callback) ->
+                new Disposable
+              onDidSearchPaths: (callback) ->
+                new Disposable
+              cancel: ->
+                @hoistedReject()
+
             class CustomDirectorySearcherToCancel
               canSearchDirectory: (directory) -> directory.getPath() is dir1
-              search: (directory, regexSource, onSearchResult, onSearchError, onPathsSearched, options) ->
-                # Note that hoisting reject in this way is generally frowned upon.
-                hoistedReject = null
-                promise = new Promise (resolve, reject) ->
-                  hoistedReject = reject
-                promise.cancel = -> hoistedReject()
-                customDirectorySearcherPromiseInstance = promise
-                promise
+              search: (directory, options) ->
+                new CustomDirectorySearchToCancel
 
             atom.packages.serviceHub.provide(
               "atom.directory-searcher", "0.1.0", new CustomDirectorySearcherToCancel())
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -978,7 +978,8 @@ describe "Workspace", ->
               ]
             onFakeSearchCreated = (fakeSearch) ->
               fakeSearch.options.didMatch(searchResult)
-              fakeSearch.options.didSearchPaths(numPathsToPretendToSearchInCustomDirectorySearcher)
+              directory1 = atom.project.getDirectories()[atom.project.getPaths().indexOf(dir1)]
+              fakeSearch.options.didSearchPaths(directory1, numPathsToPretendToSearchInCustomDirectorySearcher)
               fakeSearch.hoistedResolve()
 
             resultPaths = []
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -945,12 +945,8 @@ describe "Workspace", ->
             numPathsSearchedInDir2 = 1
             numPathsToPretendToSearchInCustomDirectorySearcher = 10
             class CustomDirectorySearch
-              constructor: ->
+              constructor: (delegate) ->
                 @promise = Promise.resolve()
-              then: (args...) ->
-                @promise.then.apply(@promise, args)
-              onDidMatch: (callback) ->
-                # Invoke the callback with the only result we plan to return.
                 searchResult1 =
                   filePath: foreignFilePath,
                   matches: [
@@ -959,22 +955,18 @@ describe "Workspace", ->
                       lineTextOffset: 0,
                       matchText: 'Hello',
                       range: [[0, 0], [0, 5]],
-                    }
+                    },
                   ]
-                callback(searchResult1)
-                new Disposable
-              onDidError: (callback) ->
-                new Disposable
-              onDidSearchPaths: (callback) ->
-                # Invoke the callback with the one notification we plan to send.
-                callback(numPathsToPretendToSearchInCustomDirectorySearcher)
-                new Disposable
+                delegate.onDidMatch(searchResult1)
+                delegate.onDidSearchPaths(numPathsToPretendToSearchInCustomDirectorySearcher)
+              then: (args...) ->
+                @promise.then.apply(@promise, args)
               cancel: ->
 
             class CustomDirectorySearcher
               canSearchDirectory: (directory) -> directory.getPath() is dir1
-              search: (directory, options) ->
-                new CustomDirectorySearch
+              search: (directory, delegate, options) ->
+                new CustomDirectorySearch(delegate)
 
             atom.packages.serviceHub.provide(
               "atom.directory-searcher", "0.1.0", new CustomDirectorySearcher())
@@ -1004,12 +996,6 @@ describe "Workspace", ->
                 customDirectorySearcherPromiseInstance = this
               then: (args...) ->
                 @promise.then.apply(@promise, args)
-              onDidMatch: (callback) ->
-                new Disposable
-              onDidError: (callback) ->
-                new Disposable
-              onDidSearchPaths: (callback) ->
-                new Disposable
               cancel: ->
                 @hoistedReject()
 
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -1014,12 +1014,12 @@ describe "Workspace", ->
             cancelableSearch = atom.workspace.scan /aaaa/, ->
             fakeSearch.hoistedReject()
 
-            resultOfPromiseSearch = null
+            didReject = false
             waitsForPromise ->
-              cancelableSearch.then (promiseResult) -> resultOfPromiseSearch = promiseResult
+              cancelableSearch.catch -> didReject = true
 
             runs ->
-              expect(resultOfPromiseSearch).toBe('cancelled')
+              expect(didReject).toBe(true)
 
   describe "::replace(regex, replacementText, paths, iterator)", ->
     [filePath, commentFilePath, sampleContent, sampleCommentContent] = []
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -939,36 +939,47 @@ describe "Workspace", ->
                   expect(resultPaths).toEqual([file2])
 
         describe "when a custom directory searcher is registered", ->
+          fakeSearch = null
+          # Function that is invoked once all of the fields on fakeSearch are set.
+          onFakeSearchCreated = null
+
+          class FakeSearch
+            constructor: (@options) ->
+              # Note that hoisting resolve and reject in this way is generally frowned upon.
+              @promise = new Promise (resolve, reject) =>
+                @hoistedResolve = resolve
+                @hoistedReject = reject
+                onFakeSearchCreated?(this)
+            then: (args...) ->
+              @promise.then.apply(@promise, args)
+            cancel: -> @cancelled = true
+
+          beforeEach ->
+            fakeSearch = null
+            onFakeSearchCreated = null
+            atom.packages.serviceHub.provide('atom.directory-searcher', '0.1.0', {
+              canSearchDirectory: (directory) -> directory.getPath() is dir1
+              search: (directory, regex, options) -> fakeSearch = new FakeSearch(options)
+            })
+
           it "can override the DefaultDirectorySearcher on a per-directory basis", ->
             foreignFilePath = 'ssh://foreign-directory:8080/hello.txt'
             numPathsSearchedInDir2 = 1
             numPathsToPretendToSearchInCustomDirectorySearcher = 10
-            class CustomDirectorySearch
-              constructor: (delegate) ->
-                @promise = Promise.resolve()
-                searchResult1 =
-                  filePath: foreignFilePath,
-                  matches: [
-                    {
-                      lineText: 'Hello world',
-                      lineTextOffset: 0,
-                      matchText: 'Hello',
-                      range: [[0, 0], [0, 5]],
-                    },
-                  ]
-                delegate.didMatch(searchResult1)
-                delegate.didSearchPaths(numPathsToPretendToSearchInCustomDirectorySearcher)
-              then: (args...) ->
-                @promise.then.apply(@promise, args)
-              cancel: ->
-
-            class CustomDirectorySearcher
-              canSearchDirectory: (directory) -> directory.getPath() is dir1
-              search: (directory, delegate, options) ->
-                new CustomDirectorySearch(delegate)
-
-            atom.packages.serviceHub.provide(
-              "atom.directory-searcher", "0.1.0", new CustomDirectorySearcher())
+            searchResult =
+              filePath: foreignFilePath,
+              matches: [
+                {
+                  lineText: 'Hello world',
+                  lineTextOffset: 0,
+                  matchText: 'Hello',
+                  range: [[0, 0], [0, 5]],
+                },
+              ]
+            onFakeSearchCreated = (fakeSearch) ->
+              fakeSearch.options.didMatch(searchResult)
+              fakeSearch.options.didSearchPaths(numPathsToPretendToSearchInCustomDirectorySearcher)
+              fakeSearch.hoistedResolve()
 
             resultPaths = []
             onPathsSearched = jasmine.createSpy('onPathsSearched')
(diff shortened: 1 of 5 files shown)


src/default-directory-searcher.coffee
```diff
@@ -0,0 +1,61 @@
...
+      task = Task.once require.resolve('./scan-handler'), rootPaths, regexSource, options, resolve
+      task.on 'task:cancelled', reject
+    promise.cancel = ->
+      task.cancel()
+
+    task.on 'scan:result-found', onSearchResult
+    task.on 'scan:file-error', onSearchError
+    task.on 'scan:paths-searched', onPathsSearched
+
+    promise
```
maxbrunsfeld: I think if this interface is going to be a public API, we should probably find a way to avoid the `Promise that includes a cancel() method` thing. I know that's the way `Workspace::scan` behaves currently, but that's currently undocumented, and seems like a pattern we should prevent from spreading.

It might fit in better with Atom's other APIs if `::search()` returned a `DirectorySearch` object that allows the user to subscribe to different events:

```js
let search = searcher.search(theDirectory, "the-regexp", options);

search.onDidMatch((result) => console.log("Got result", result));
search.onDidError((error) => console.log("Error", error));
search.onDidSearchPaths((count) => console.log("Finished searching ", count, "files"));
search.onDidComplete((error) => console.log("Done"));

// ...

search.cancel();
```

@bolinfest @atom/core what do you think we should do for this API?

src/workspace.coffee
```diff
@@ -791,36 +800,72 @@ class Workspace extends Model
...
-        options.onPathsSearched(numberOfPathsSearched)
+      # Maintain a map of providers to the number of search results. When notified of a new count,
+      # replace the entry in the map and update the total.
+      onPathsSearchedOption = options.onPathsSearched
+      totalNumberOfPathsSearched = 0
+      numberOfPathsSearchedForProvider = new Map()
+      onPathsSearched = (provider, numberOfPathsSearched) ->
+        oldValue = numberOfPathsSearchedForProvider.get(provider)
+        if oldValue
+          totalNumberOfPathsSearched -= oldValue
```
maxbrunsfeld: What if a given `DirectorySearcher` object is being used for multiple searches simultaneously? Should this `Map` have `Directories` as keys, rather than `DirectorySearchers`?

src/workspace.coffee
```diff
@@ -791,36 +800,72 @@ class Workspace extends Model
...
+      inclusions: options.paths or []
       includeHidden: true
       excludeVcsIgnores: atom.config.get('core.excludeVcsIgnoredPaths')
       exclusions: atom.config.get('core.ignoredNames')
       follow: atom.config.get('core.followSymlinks')
 
-    task = Task.once require.resolve('./scan-handler'), atom.project.getPaths(), regex.source, searchOptions, ->
-      deferred.resolve()
+    # Find a search provider for every Directory in the project.
+    providersAndDirectories = []
```
maxbrunsfeld: I think we should not use the term `provider` to refer to these objects. Probably just `searcher`.

src/workspace.coffee
```diff
@@ -791,36 +800,72 @@ class Workspace extends Model
...
+      {provider, directory} = entry
+      recordNumPathsSearched = onPathsSearched.bind(undefined, provider)
+      allSearchPromises.push(provider.search(
+        directory,
+        regex.source,
+        onSearchResult,
+        onSearchError,
+        recordNumPathsSearched,
+        searchOptions))
+    searchPromise = Promise.all(allSearchPromises)
```
maxbrunsfeld: This means that when there are, for example 8 local root directories, we'll spawn 8 tasks to search them, rather than one task that searches all of them. This might be OK, but if we wanted to avoid it, we could de-dup the `atom.directory-searcher` services and make the `::search()` method take an *array* of `Directory` objects instead of just one `Directory`.

@atom/core @bolinfest do you think it matters how many tasks we spawn?

src/workspace.coffee
```diff
@@ -829,11 +874,30 @@ class Workspace extends Model
...
+    cancellablePromise.cancel = ->
+      # Note that cancelling all (or actually, any) of the members of allSearchPromises
+      # will cause searchPromise to reject, which will cause cancellablePromise to resolve
+      # in the desired way.
+      promise.cancel() for promise in allSearchPromises
+
+    # Although this method claims to return a `Promise`, the `ResultsPaneView.onSearch()`
+    # method in the find-and-replace package expects the object returned by this method to have a
+    # `done()` method. Include a done() method until find-and-replace can be updated.
+    cancellablePromise.done = (f) ->
```
kevinsawicki: what is `f`?  Mind giving it a more descriptive name?

src/workspace.coffee
```diff
@@ -791,36 +800,72 @@ class Workspace extends Model
...
+        totalNumberOfPathsSearched += numberOfPathsSearched
+        onPathsSearchedOption(totalNumberOfPathsSearched)
+    else
+      onPathsSearched = ->
+
+    # Kick off all of the searches and unify them into one Promise.
+    allSearchPromises = []
+    for entry in providersAndDirectories
+      {provider, directory} = entry
+      recordNumPathsSearched = onPathsSearched.bind(undefined, provider)
```
kevinsawicki: Seems a bit inconsistent to have `totalNumberOfPathsSearched` above but then shorten this to `recordNumPathsSearched` instead of `recordNumberOfPathsSearched`

src/workspace.coffee
```diff
@@ -791,36 +800,72 @@ class Workspace extends Model
...
+    for directory in atom.project.getDirectories()
+      providerForDirectory = null
+      for provider in @searchProviders
+        if provider.canSearchDirectory(directory)
+          providerForDirectory = provider
+          break
+      if providerForDirectory
+        providersAndDirectories.push({provider, directory})
+      else
+        throw Error("Could not find search provider for #{directory.getPath()}")
```
kevinsawicki: Since this methods return a promise, shouldn't the error be reported to that promise instead of thrown explicitly?

src/default-directory-searcher.coffee
bolinfest: @maxbrunsfeld I can't remove support for `cancel()` in this change because it is used in the `find-in-replace` package:

    @inProgressSearchPromise.cancel()

Also, I think it is valuable to continue to use a `Promise` to make it easier to compose with `Promise.all()`, use with the `await` keyword, etc.

At a minimum, I think it would have to continue to be a *thenable* such that is "Promise compatible," though I would have to test that out to be sure. (See http://www.html5rocks.com/en/tutorials/es6/promises/#toc-promise-terminology for info on *thenable*.)

bolinfest: In terms of the callbacks, I think it would be cleaner to pass in a single object with an API similar to how delegates are often used in iOS (e.g., https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIWebViewDelegate_Protocol/):

```
DirectorySearchDelegate {
  searcherDidMatch(result) {}
  searcherDidError(error) {}
  searcherDidSearchPaths(count) {}
  searcherDidComplete() {}
}
```

Then the caller passes in an instance of `DirectorySearchDelegate`, which avoids a bunch of event registration, deregistration. (I think.)

src/workspace.coffee
bolinfest: Good catch! Fixed.

src/workspace.coffee
bolinfest: Changed to searcher throughout.

src/workspace.coffee
bolinfest: My gut feeling is no. In general, I expect users will appreciate the reduction in latency.

Also, if this is a problem, we can fix this later by changing how the `DefaultDirectorySearcher` works. The `Promise` it returns could periodically `yield` to other active searchers, etc.

src/workspace.coffee
bolinfest: Renamed to `onSuccessOrFailure`.

src/workspace.coffee
bolinfest: Changed to `recordNumberOfPathsSearched`.

src/workspace.coffee
bolinfest: I would say no: this is basically an assertion and we should fail fast. I wouldn't want this swept under the rug as an unhandled promise rejection.

If this error is thrown, we would have violated our invariant that there should always be a member of `@directorySearchers` for which `canSearchDirectory()` returns `true` for a given `Directory`, which we expect to uphold by having `DefaultDirectorySearcher` at the end of the list, as its `DefaultDirectorySearcher::canSearchDirectory()` indiscriminately returns `true`.

src/default-directory-searcher.coffee
maxbrunsfeld: > At a minimum, I think it would have to continue to be a thenable such that is "Promise compatible,"

That seems fine to me. I just think that since the returned object needs to have a `.cancel()` method, it would be clearer if were an instance of our own class, not of `Promise`.

> Then the caller passes in an instance of DirectorySearchDelegate, which avoids a bunch of event registration.

I see your point about the smaller amount of indirection involved when using the delegate pattern (one callback instead of potentially n callbacks), but in our situation I don't think there is much overhead or complexity involved in using event registration. All of Atom's existing APIs use the observer pattern, and it fits in well with our other tools like `CompositeDisposable`.

Also, I think it would be simpler implementation-wise as well, at least in the case of the `DefaultDirectorySearcher`. Rather than having to define a callback, like this:

bolinfest:
@maxbrunsfeld I updated the code and tests so that a `DirectorySearcher` must return a `DirectorySearch` object that has the following methods:

* `then()`
* `onDidMatch()`
* `onDidError()`
* `onDidSearchPaths()`
* `cancel()`

Admittedly, I need to update all of the docs in `default-directory-search.coffee`, but I wanted to make sure you were OK with this API before I invested in that.

maxbrunsfeld:
Looks good to me. Thanks @bolinfest.

src/default-directory-searcher.coffee
```diff
@@ -0,0 +1,91 @@
...
+
+# Public:
+# Implements thenable so it can be used with `Promise.all()`.
+class DirectorySearch
+  # Public:
+  constructor: (directory, options) ->
+    @task = new Task(require.resolve('./scan-handler'))
+    rootPaths = [directory.getPath()]
+    @promise = new Promise (resolve, reject) =>
+      myResolve = (arg) ->
```
maxbrunsfeld: for debugging?

bolinfest: Good catch: fixed.

spec/workspace-spec.coffee
```diff
@@ -941,6 +942,100 @@ describe "Workspace", ->
...
+              search: (directory, options) ->
+                new CustomDirectorySearchToCancel
+
+            atom.packages.serviceHub.provide(
+              "atom.directory-searcher", "0.1.0", new CustomDirectorySearcherToCancel())
+
+            resultPaths = []
+            cancelableSearch = atom.workspace.scan /aaaa/, ({filePath}) ->
+              resultPaths.push(filePath)
+            customDirectorySearcherPromiseInstance.cancel()
```
maxbrunsfeld: The *individual* `DirectorySearch` objects are not exposed to callers of `Workspace::scan`, right? Currently, this test is verifying the behavior of one of these individual searches. I think that instead, the test should assert is that calling `.cancel()` on the return value of `Workspace::scan` has the effect of calling `cancel` on each individual `DirectorySearch`.

bolinfest:
I just rebased because apparently my previous stack of diffs would not have merged cleanly because of the update to `spec/task-spec.coffee`.

bolinfest:
@maxbrunsfeld This fixes/amends the tests as we discussed.

bolinfest:
Rebased.

spec/workspace-spec.coffee
```diff
@@ -941,6 +941,113 @@ describe "Workspace", ->
...
+                @promise.then.apply(@promise, args)
+              cancel: ->
+
+            class CustomDirectorySearcher
+              canSearchDirectory: (directory) -> directory.getPath() is dir1
+              search: (directory, delegate, options) ->
+                new CustomDirectorySearch(delegate)
+
+            atom.packages.serviceHub.provide(
+              "atom.directory-searcher", "0.1.0", new CustomDirectorySearcher())
```
maxbrunsfeld: I think these tests could be less verbose if you registered a searcher in a generic way in a `beforeEach`:

```coffee
fakeSearch = null

class FakeSearch
  constructor: (@delegate) ->
  then: (@simulateSuccess, @simulateFailure) ->
  cancel: -> @cancelled = true

beforeEach ->
  fakeSearch = null
  atom.packages.serviceHub.provide("atom.directory-searcher", "0.1.0", {
    canSearchDirectory: (directory) -> directory is dir1
    search: (directory, delegate) -> fakeSearch = new FakeSearch(delegate)
  })
```

Then the contents of the `it` blocks could be more focused on what's being tested:

```coffee
  it "can be cancelled...", ->
    # ...
    expect(fakeSearch.cancelled).toBe true

  it "fails if any particular search fails", ->
    # ...
    fakeSearch.simulateFailure("some-error");

  it "can override the DefaultDirectorySearcher...", ->
    # ...
    fakeSearch.delegate.didMatch({
      filePath: "the-path",
      matches: []
    })
```

spec/workspace-spec.coffee
```diff
@@ -941,6 +941,113 @@ describe "Workspace", ->
...
+
+            cancelableSearch = atom.workspace.scan /aaaa/, ->
+            hoistedReject()
+
+            resultOfPromiseSearch = null
+            waitsForPromise ->
+              cancelableSearch.then (promiseResult) -> resultOfPromiseSearch = promiseResult
+
+            runs ->
+              expect(resultOfPromiseSearch).toBe('cancelled')
```
maxbrunsfeld: Does this mean that if the search fails, then `.scan()` will always resolve with the string `"cancelled"`?

src/default-directory-searcher.coffee
```diff
@@ -0,0 +1,107 @@
...
+  #       * `filePath` {String} absolute path to the matching file.
+  #       * `matches` {Array} with object elements with the following keys:
+  #         * `lineText` {String} The full text of the matching line (without a line terminator character).
+  #         * `lineTextOffset` {Number} (This always seems to be 0?)
+  #         * `matchText` {String} The text that matched the `regex` used for the search.
+  #         * `range` {Range} Identifies the matching region in the file. (Likely as an array of numeric arrays.)
+  #   * `didError` {Function} call with an Error if there is a problem during the search.
+  #   * `didSearchPaths` {Function} periodically call with the number of paths searched thus far.
+  # * `options` {Object} with the following properties:
+  #   * `regexSource` {String} regex to search with. Produced via `RegExp::source`.
```
maxbrunsfeld: I think that `regexSource` should be a separate parameter, since it's so central to the point of the method. I would also say that the `delegate` and `options` objects could be merged. I don't think we should frame this like iOS/Cocoa, where the delegate is its own long-lived entity; it's just a collection of optional callbacks, which we're going to be creating for each call to `::search()`.

So I'd say a call should look like:

```coffee
search = searcher.search(theDirectory, "the-pattern", {
  ignoreCase: true,
  didError: ->
  didMatch: ->
  didSearchPaths: ->
})
```

spec/workspace-spec.coffee
bolinfest: Yes, though note this comment from workspace.spec:

```coffee
    # Make sure the Promise that is returned to the client is cancelable. To be consistent
    # with the existing behavior, instead of cancel() rejecting the promise, it should
    # resolve it with the special value 'cancelled'. At least the built-in find-and-replace
    # package relies on this behavior.
```

I would love to fix this in a subsequent diff, but it requires coordination between `find-and-replace` and this package.

src/default-directory-searcher.coffee
```diff
@@ -0,0 +1,107 @@
+{EventEmitter} = require 'events'
+Task = require './task'
+
+# Maintain a queue of ids of searches to run. When a search is complete, the active
+# search is cleared and the next search (if any) in the queue is notified to run.
+# This ensures there is at most one scan-handler task running at a time.
+searchQueue = []
+nextId = 1
+activeSearchId = 0
+emitter = new EventEmitter
```
maxbrunsfeld: I'd like to avoid using global state here. I'm open to any alternatives you can think of, but my preference would be to just let `DirectorySearcher::search` take multiple directories, so we don't need queuing.

src/default-directory-searcher.coffee
bolinfest: Point taken, will do.

bolinfest: Though I think it should just get the raw `RegExp` and the searcher should decide whether it needs to invoke the `.source` param, no? I believe the only reason this was done originally is so it could be serialized over the process boundary, so that was leaking an implementation detail.

src/workspace.coffee
```diff
@@ -791,36 +800,76 @@ class Workspace extends Model
...
 
-    task.on 'scan:file-error', (error) ->
-      iterator(null, error)
+    # Now that we are sure every Directory has a searcher, construct the search options.
+    delegateProto = {
+      didMatch: (result) ->
+        iterator(result) unless atom.project.isPathModified(result.filePath)
+      didError: (error) ->
+        iterator(null, error)
+    }
```
maxbrunsfeld: Was this to address some performance problem you were seeing in this method? It seems a little odd to use `Object.create` for this purpose. I would say just include these functions directly in the argument to `.search()` below, unless you're seeing that this speeds up the function significantly.

src/default-directory-searcher.coffee
maxbrunsfeld: > Though I think it should just get the raw RegExp and the searcher should decide whether it needs to invoke the .source param.

:+1: I like that.

spec/workspace-spec.coffee
maxbrunsfeld: But that's only if the search is cancelled, right? The way I'm reading this test, it seems to assert that the promise is resolved with that value if a search fails for any reason.

src/workspace.coffee
bolinfest: OK, options is a much bigger object now, but I'll just create a new one each time.

src/workspace.coffee
```diff
@@ -791,36 +800,76 @@ class Workspace extends Model
...
+    for directory in atom.project.getDirectories()
+      searcher = null
+      for directorySearcher in @directorySearchers
+        if directorySearcher.canSearchDirectory(directory)
+          searcher = directorySearcher
+          break
+      if searcher
+        searchersAndDirectories.push({searcher, directory})
+      else
+        throw Error("Could not find directory searcher for #{directory.getPath()}")
```
maxbrunsfeld: This is similar to what you did for the directory-provider right, where this exception is just here to signal an internal error? I was thinking it might be more clear and direct to explicitly choose the `DefaultDirectorySearcher` here, rather than putting it into the `@directorySearchers` array. Then we wouldn't need this `throw` here at all.

spec/workspace-spec.coffee
bolinfest: FYI, this is made difficult by:

```coffee
            waitsForPromise ->
              atom.workspace.scan /aaaa/, {onPathsSearched}, ({filePath}) ->
                resultPaths.push(filePath)
```

Because the `Promise` returned by `atom.workspace.scan` needs `fakeSearch` to do work in order for it to ever resolve.

maxbrunsfeld: Yeah, I think you'd have to save off the return value of `scan`. So that you could manipulate the fake-search before waiting for the scan to complete.

bolinfest: OK, I'm inserting this optional callback to make this work: you'll see.

src/workspace.coffee
bolinfest: Good call: will do.

src/default-directory-searcher.coffee
bolinfest: OK, let me try multiple directory support. This should be easier to implement now that the return value of a `DirectorySearcher` is only required to have `then()` and `cancel()` as methods. (This was more difficult when it was also responsible for `onDidMatch`, `onDidPathsSearched`, etc.)

bolinfest: Ugh, I think I still have to create yet-another-custom-class to achieve this. We'll see what you think.

bolinfest: Actually, this may not be so bad.

bolinfest: OK, you can see how I cleaned up `DefaultDirectorySearcher`, but I think I'm going to change the contract of `didSearchPaths` where the client must specify both the `Directory` and the count.

bolinfest: All fixed.

spec/workspace-spec.coffee
bolinfest: I assume it's OK to actually reject the thenable returned by `atom.workspace.scan()` in this case?

bolinfest: On further consideration, this behavior might be hard to change because of the following:

```coffee
    cancellablePromise = new Promise (resolve, reject) ->
      onSuccess = ->
        resolve(null)
        return
      onFailure = ->
        resolve('cancelled')
        return
      searchPromise.then(onSuccess, onFailure)
```

I guess we want to say:
* If all the searches succeed, `cancellablePromise` resolves with `null`.
* If a search is cancelled, `cancellablePromise` resolves with `'cancelled'`.
* If a search fails, `cancellablePromise` rejects.

I think the tricky part is that it requires us to distinguish between a search that is cancelled vs. one that fails. This will probably require hoisting of resolve/reject functions...

bolinfest: OK, I was able to do this without hoisting and the unit test now reflects this.

src/workspace.coffee
```diff
@@ -46,6 +47,15 @@ class Workspace extends Model
     @paneContainer ?= new PaneContainer()
     @paneContainer.onDidDestroyPaneItem(@didDestroyPaneItem)
 
+    @directorySearchers = []
+    @defaultDirectorySearcher = new DefaultDirectorySearcher()
+    atom.packages.serviceHub.consume(
+      'atom.directory-searcher',
+      '^0.1.0',
+      # New providers are added to the front of @directorySearchers because
+      # DefaultDirectorySearcher is a catch-all that will always claim to search a Directory.
```
maxbrunsfeld: I think you can :fire: this comment now.

src/workspace.coffee
```diff
@@ -791,36 +801,67 @@ class Workspace extends Model
...
+      searcher = @defaultDirectorySearcher
+      for directorySearcher in @directorySearchers
+        if directorySearcher.canSearchDirectory(directory)
+          searcher = directorySearcher
+          break
+      directories = directoriesForSearcher.get(searcher)
+      unless directories
+        directories = []
+        directoriesForSearcher.set(searcher, directories)
+      directories.push(directory)
```
maxbrunsfeld: Nice :+1:

src/default-directory-searcher.coffee
```diff
@@ -0,0 +1,99 @@
...
+  #
+  # Returns a *thenable* `DirectorySearch` that includes a `cancel()` method. If `cancel()` is
+  # invoked before the `DirectorySearch` is determined, it will reject the `DirectorySearch`.
+  search: (directories, regex, options) ->
+    # Make a mutable copy of the directories array.
+    directories = directories.slice(0)
+    isCancelled = false
+    promise = new Promise (resolve, reject) ->
+      run = ->
+        if isCancelled
```
maxbrunsfeld: I think you could simplify this even further: since the `scan-handler` task accepts multiple paths as arguments, you could just pass the whole array of directories into `DirectorySearch`, and have it pass all of their paths to the task.

src/workspace.coffee
bolinfest: :fire:'d

src/default-directory-searcher.coffee
bolinfest: Does the task take care of making sure only one directory is searched at a time?

Also, I fear that might make it hard to implement `didSearchPaths` correctly since it is currently parameterized by `Directory`:

```coffee
    @task.on 'scan:paths-searched', (count) -> options.didSearchPaths(directory, count)
```

bolinfest: Though if you want, I guess we can change the contract so that the first argument to `didSearchPaths` is the `provider` rather than the `directory`?

bolinfest: OK, I just added two more commits that fix this.

Now, only one `scan-handler` task is created per call to `search()`.
I solved the `didSearchPaths` thing by passing each searcher its own `didSearchPaths` method.

maxbrunsfeld:
Looks great. I'd say let's :ship: tomorrow.

bolinfest:
Thanks @maxbrunsfeld: sounds great!

src/workspace.coffee
```diff
@@ -791,36 +799,65 @@ class Workspace extends Model
   # * `regex` {RegExp} to search with.
   # * `options` (optional) {Object} (default: {})
   #   * `paths` An {Array} of glob patterns to search within
+  #   * `onPathsSearched` (optional) {Function}
   # * `iterator` {Function} callback on each file found
   #
-  # Returns a `Promise`.
+  # Returns a *thenable* object with a `cancel()` method that will cancel all
+  # of the underlying searches that were started as part of this scan.
   scan: (regex, options={}, iterator) ->
```
kevinsawicki: This method still returns a `Promise` right? Can you change the doc back to a `Promise` from a `thenable` if it does?

bolinfest: No, it now returns a *thenable*, which is an object with a Promise-compatible `then()` method.

A `Promise` has only two instance methods (http://www.html5rocks.com/en/tutorials/es6/promises/#toc-api):

* `then()`
* `cancel()`

I could also support `cancel`, I guess, but the returned object would not pass an `instanceof Promise` test because the code is:

```coffee
return {
  then: promise.then.bind(promise)
  cancel: ->
    isCancelled = true
    directorySearch.cancel()
}
```

In this case, it also supports an undocumented `done()` method, but as you can see in the code, that is only for backwards compatibility with `find-and-replace`.

kevinsawicki: Looks like it returns this right: https://github.com/atom/atom/blob/3f03317b08647aaba5ef4c470af618a36ca1b76a/src/workspace.coffee#L874

Which is a `Promise`, am I missing something?

bolinfest: Whoops, I went to the wrong file. Will fix.

bolinfest: Fixed.

bolinfest:
@kevinsawicki all good by you?

bolinfest:
@maxbrunsfeld @kevinsawicki I just rebased this on top of HEAD:master.

maxbrunsfeld:
:rocket: Thanks @bolinfest!

maxbrunsfeld:
Hey @bolinfest, we need to back this out for today's release, because it broke some specs in the `find-and-replace`. It looks like something to do with canceling? Can you take a look at these? We can re-merge shortly.

```
 ResultsModel
   cancelling a search
     it populates the model with all the results, and updates in response to changes in the buffer
       Expected spy unknown to have been called.
         at [object Object].<anonymous> (/private/var/lib/jenkins/workspace/atom/node_modules/find-and-replace/spec/results-model-spec.coffee:125:30)
         at process._tickCallback (node.js:367:9)
     it populates the model with all the results, and updates in response to changes in the buffer
       Expected spy unknown to have been called. (spec/results-model-spec.coffee:136:30)
```

PR #7051, opened 2015-06-01
Include minor deprecations in 1.0 API mode

kevinsawicki:
Keep deprecating but don't remove certain minor deprecations that are easy to keep supporting for a little while longer and don't slow down Atom:
- Loading stylesheets from `stylesheets/` (now `styles/`)
- Loading settings from `scope-properties/` (now `settings/`)
- `activationEvents` in `package.json` (now `activationCommands`)
- `configDefaults` in package main (now `config`)
- `atom.config.unobserve` API (currently a no-op)
- Old `context-menu` format

This is an effort to not break packages that have very minor deprecations.

Refs #6867

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -1158,6 +1158,7 @@ withoutEmptyObjects = (object) ->
     resultObject = object
   resultObject
 
+# TODO remove in 1.0 API
 Config::unobserve = (keyPath) ->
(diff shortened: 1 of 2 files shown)


benogle:
Should we remove those `on` calls in pane-view too?

kevinsawicki:
> Should we remove those on calls in pane-view too?

After looking that a bit more, `PaneView` is gone from 1.0 mode anyway so I don't think those calls need to be updated.

PR #7055, opened 2015-06-01
Rename --one to --include-deprecated-apis

kevinsawicki:
Switches 1.0 API preview mode to be enabled by default and renames the flag to be `--include-deprecated-apis` to still run Atom with deprecated APIs.


benogle:
:+1:

PR #7060, opened 2015-06-02
Load large files in background

nathansobo:
:rotating_light: Work In Progress :rotating_light:

Loading a 3.7MB file created by concatenating jQuery to itself multiple times:

async-load

Constructing the initial set of `TokenizedLine`s hangs the UI thread for large files. This PR performs this work in a background task for files above a certain size to avoid the pause, indicating progress with a progress bar. To integrate smoothly with existing packages, we open a `TextEditorLoader` item to display progress until the editor is fully loaded, then replace it with the editor afterward. This ensures that no packages attempt to interact with an editor in a partially loaded state.
### Can't change grammars

We currently use tokenized lines to store horizontal spatial information such as hard tabs, paired unicode characters, soft-wrap indents, etc. This is in addition to their primary purpose of representing syntactic scopes. This entanglement makes it awkward to change the grammar once the file is loaded, because doing so replaces all the tokenized lines with the same placeholder lines we're constructing in the background to clear all the styling, then starts tokenizing again from the beginning. However, this takes a lot of time for a large file. Once the editor is loaded, I don't want to put it back in "background loading" mode because that implies a lot of complexity. Ideally I'd like to store the horizontal spatial information outside of the tokenized lines, so styling could be cleared without throwing away all this computation.
### Can't change tab length, invisibles, etc

Again, these require recomputing the tokenized lines from scratch. Invisibles can be dealt with by separating them out into a second rendering phase once our position calculation is more decoupled from the content of the model. Tab lengths are not so easy. We may be able to use our existing metadata about tab locations to update the lengths more efficiently in a synchronous way (without having to scan every token). Or we may want to just disable the ability to change it over a certain size. Or we may need to bite the bullet and figure out how to put the editor back in "background loading" mode after all. I'd like to put that off though.
### Long pauses before / after progress bar

The initial pause is due to loading the entire buffer in the background task before beginning processing. I just did this as a first pass implementation and will fix it before merging this. The second pause is autocomplete-plus populating its symbol table. We may want to disable it for now for files above a certain size, then work on optimizing this.
## Tasks
- [x] Display progress bar for loading in background
- [ ] Disallow operations that require recomputation of tokenized lines
- [ ] Optimize background task by computing results in a streaming fashion
- [ ] Do something about autocomplete plus

I may want to push tokenization to a background task in a different PR. That would allow us to at least select an initial grammar for large files instead of the null grammar as is currently the case on this PR. But I think loading huge files without highlighting is enough of an improvement to do that in a second pass.

### DIFF
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -24,13 +24,13 @@ class DisplayBuffer extends Model
   horizontalScrollMargin: 6
   scopedCharacterWidthsChangeCount: 0
 
-  constructor: ({tabLength, @editorWidthInChars, @tokenizedBuffer, buffer, ignoreInvisibles}={}) ->
+  constructor: ({tabLength, @editorWidthInChars, @tokenizedBuffer, buffer, ignoreInvisibles, @largeFileMode}={}) ->
     super
 
     @emitter = new Emitter
     @disposables = new CompositeDisposable
 
-    @tokenizedBuffer ?= new TokenizedBuffer({tabLength, buffer, ignoreInvisibles})
+    @tokenizedBuffer ?= new TokenizedBuffer({tabLength, buffer, ignoreInvisibles, @largeFileMode})
     @buffer = @tokenizedBuffer.buffer
     @charWidthsByScope = {}
     @markers = {}
@@ -41,6 +41,12 @@ class DisplayBuffer extends Model
     @disposables.add @tokenizedBuffer.onDidChange @handleTokenizedBufferChange
     @disposables.add @buffer.onDidCreateMarker @handleBufferMarkerCreated
     @foldMarkerAttributes = Object.freeze({class: 'fold', displayBufferId: @id})
+    if @largeFileMode
+      @tokenizedBuffer.onDidLoad(=> @finalizeConstruction())
+    else
+      @finalizeConstruction()
+
+  finalizeConstruction: ->
     folds = (new Fold(this, marker) for marker in @buffer.findMarkers(@getFoldMarkerAttributes()))
     @updateAllScreenLines()
     @decorateFold(fold) for fold in folds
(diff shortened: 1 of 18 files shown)


Zireael07:
Woo!

src/text-editor-loader-element.coffee
```diff
@@ -0,0 +1,25 @@
+class TextEditorLoaderElement extends HTMLElement
+  createdCallback: ->
+    progressBar = document.createElement('div')
+    progressBar.classList.add('progress-bar')
```
kevinsawicki: Any reason to not use the `<progress>` element from the style guide?

src/display-buffer.coffee
```diff
@@ -41,6 +41,12 @@ class DisplayBuffer extends Model
     @disposables.add @tokenizedBuffer.onDidChange @handleTokenizedBufferChange
     @disposables.add @buffer.onDidCreateMarker @handleBufferMarkerCreated
     @foldMarkerAttributes = Object.freeze({class: 'fold', displayBufferId: @id})
+    if @largeFileMode
+      @tokenizedBuffer.onDidLoad(=> @finalizeConstruction())
```
YuriSolovyov: I wonder if some day code path for large files will be good enough to use it unconditionally for both large and small files.

src/text-editor-loader-element.coffee
nathansobo: Only that I didn't know it existed. Will update.

src/display-buffer.coffee
nathansobo: I definitely think it will be. This is a temporary stopgap, although we may not ever initialize smaller files in a background process because it might not be worth the overhead.

nathansobo:
Rather than doing work in the background, I've found a way to drop a couple features and just not do the work at all.

Zireael07:
Uh, I'm sure more than a few people are having large files that should be highlighted for grammar, maybe consider getting this merged?

nathansobo:
I'm going to look into background highlighting next week, but that's a different story. We won't support folds and soft wraps for a while though which is the only reason we'd really need to do this.

PR #7071, opened 2015-06-02
Add libcap2 as a dependency for Debian

### LINKED ISSUE #7066: libcap2 missing from .deb installer
Can someone give me a hint why I can't run atom? This is the error. I've installed atom 204 using the .deb from the home page on a new installation of ubuntu 14.04.  I've had no problem running atom on this ubuntu installation in the past.

```
/usr/share/atom/atom: error while loading shared libraries: libcap.so.2: cannot open shared object file: No such file or directory
```

winstliu:
Closes #7066


kevinsawicki:
:+1:

PR #7072, opened 2015-06-03
Add atom.keymaps.dispatch

thedaniel:
This is intended for simulating key input in specs. Its implementation is based on a spec 
helper from vim-mode: 
https://github.com/atom/vim-mode/blob/7926ffd777c71763597a045a3f79f670ffef025e/spec/spec-helper.coffee#L47-L64

cc @nathansobo @orderedlist @maxbrunsfeld

### DIFF
--- a/src/keymap-extensions.coffee
+++ b/src/keymap-extensions.coffee
@@ -62,6 +62,44 @@ KeymapManager::subscribeToFileReadFailure = ->
 
     atom.notifications.addError(message, {detail, dismissable: true})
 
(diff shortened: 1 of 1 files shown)


nathansobo:
This is looking like a good start. Could you open it as a PR on the [atom-keymap](https://github.com/atom/atom-keymap), add some tests, and do something to reconcile it with the existing `buildKeydownEvent` method on that class? We could deprecate it, or maybe even keep it and use the code in this PR for its implementation.

thedaniel:
Sure thing.

PR #7084, opened 2015-06-03
Log message about --one being removed

kevinsawicki:
`-1`/`--one` has been removed so add a warning about it, see #7055 for more details

This ensure people that aliased `atom --one` to `atom` don't have breakage where the path after the `--one` argument isn't treated as a path to open.


maxbrunsfeld:
:+1:

PR #7086, opened 2015-06-03
Add 'Show Preferences' command to open Settings

jlord:
We use 'Preferences' as the option within the app menu (e.g. Atom) as is typical in OS X applications. In Chrome and in Atom (and likely elsewhere) we actually call the page you land on 'Settings'. 

This PR adds 'Show Preferences' as an option so that if you happen to search for 'Preferences' rather than 'Settings' you can still make your way to the page. 

Closes https://github.com/atom/settings-view/issues/80

cc @kevinsawicki


kevinsawicki:
:ship: :+1:

PR #7101, opened 2015-06-04
Tiled Line Numbers

as-cii:
Please, note that this is still a work in progress but it makes use of the tiling technique adopted in #6733 for line-numbers as well.

On average, with these improvements we basically halve paint times. This is a screenshot before gutter tiling:

before

And this is after:

after

_To see the difference, please have a look at them at full-screen. Benchmarks were performed using the following script on `text-editor-component-spec.coffee`:_

``` coffee
startBenchmark: (time=10, delta=250) ->
  @benchmarkInterval = setInterval(->
    document.querySelector("html /deep/ .editor-contents--private").dispatchEvent(new WheelEvent("mousewheel", wheelDeltaY: -delta))
  , time)
```

Chrome further confirms we have a ~50% speedup on paint times:

| Before | After |
| --- | --- |
| screen shot 2015-06-04 at 19 29 37 | screen shot 2015-06-04 at 19 29 31 |

A byproduct of this change is that we'll be able to paint more than ~750k lines! :tada: 

I took the chance to extract out a `TiledComponent` which aims to cleanly abstract away all the complexity of creating a tile on screen and moving it up and down. I think there is a similar duplication in `TextEditorPresenter` but I'd like to a wait a bit more before introducing an abstraction in there, too.
### What's missing
- [x] Find a better name for `TileComponent` (it clashes with `TiledComponent`). I think we should keep the latter and get rid of the former, e.g. `TileComponent -> LinesTileComponent`.
- [x] Adjust existing specs to support this new behavior
- [x] Spec coverage for tiling (having a reusable contract test for this one seems like a good path, as we can use it for every `TiledComponent`)
- [x] Make sure custom gutters continue to work. @jssln: I think nothing huge should be screwed up with this PR (although changes like the one in `gutter-component-helpers.coffee` look suspicious), but you did a great job on the gutter and I'd :heart: to have your feedback :bow: 
- [x] Uncomment `if @newState.maxLineNumberDigits isnt @oldState.maxLineNumberDigits` and handle that scenario
- [x] Refine `text-editor-presenter-spec.coffee`'s tests names


benogle:
:bow: :bow:

nathansobo:
This is looking very nice. Any thoughts on how we could use composition for the `TiledComponent` rather than inheritance. I'm not super zealous about it but I wanted to raise the idea since I usually end up happier working with composition-oriented code rather than inheritance.

mnquintana:
I've been using this regularly for the past day (just normal use) and it's been smooth sailing so far :boat:

nathansobo:
Pretty excited to get this merged. What kind of support / feedback do you need to reach the :dart:?

as-cii:
> Pretty excited to get this merged. What kind of support / feedback do you need to reach the ?

Yeah, I am sorry for the delays on this but I have been quite busy with the tiled editor bugs that some users reported and wanted to fix them first. I feel quite good about those now and I should be able to deliver this quite soon (this weekend I will probably have more time to actually work on it).

I understand this is quite crucial, as it allows a more expensive computation on the text measurement side without affecting UX. Hopefully, we can :ship: this ~ by the end of this week and start making some progress on #6083 (although I suspect I am probably going to start off a new branch, as the new mechanism we'd like to try is fairly different and **master** has changed quite a lot).

@jssln: have you had the chance to have a look at this? Do you think this new strategy could break custom gutters? If yes, which issues do you think could arise? Any feedback is very welcome and much appreciated! :sparkles: :bow:

jssln:
@as-cii Just saw this! I will take a look asap -- this is probably in a few hours, but definitely today.

src/gutter-component-helpers.coffee
```diff
@@ -19,10 +19,6 @@ module.exports =
       domNode.style.height = newState.scrollHeight + 'px'
       oldState.scrollHeight = newState.scrollHeight
 
-    if newState.scrollTop isnt oldState.scrollTop
```
jssln: Can you update the comment on line 16 above?

jssln:
by the way, here are some snippets you can use to test a custom decoration in a custom gutter (just paste in the console):
```
// Create a custom gutter
var e = atom.workspace.getActiveTextEditor()
var g = e.addGutter({name: 'test'})
var gv = atom.views.getView(g)
gv.style.width = '25px'

// Add a decoration
var item = document.createElement('div')
item.className = 'test-item'
item.style.width = '25px'
item.style['background-color'] = 'green'
var m = e.markBufferRange([[12, 0], [20, 0]]) // or any other buffer range
var di = e.decorateMarker(m, {type: 'gutter', gutterName: 'test', class: 'test-class', item: item})
```

Some things to test:
* Create the decoration when the buffer lines it spans are off screen. Then scroll the editor until the decoration comes into view.
* Scroll up and down and see the decoration appear/disappear as expected.
* Test destroying the decoration and gutter

```
// Destroy the decoration & gutter
m.destroy() // the decoration should disappear
g.destroy() // the gutter should disappear
```

as-cii:
I have performed a couple of refactorings as well as completing the missing parts. This should be a drop-in replacement and no external behavior should appear as changed. I'd like to let this bake for a while but I think we're almost ready to :shipit:! :tada: 

@jssln: I ended up not touching the custom gutter at all, therefore the tests you suggested work just fine and there shouldn't be any difference in there. Still, I'd be super grateful if you could have a final :eyes: on this to catch eventual regressions.

@nathansobo: any feedback from your side is very welcome as usual and it'd be great to have your opinion about these changes. :sparkles: 

@mnquintana: thanks for trying this out! I feel like this is almost finished so it would be awesome if you could try it again and attempt to break it. We've caught a couple of bugs after we've released #6733 and I'd :heart: to have a reasonable confidence before :ship:ping this (although I am quite sure this won't be as disruptive as that PR). Would you be willing to help out with that? :pray: 

Thanks everyone! :metal:

src/line-number-gutter-component.coffee
```diff
@@ -1,13 +1,13 @@
...
 WrapperDiv = document.createElement('div')
+DummyLineNumberComponent = LineNumbersTileComponent.createDummy()
 
 module.exports =
-class LineNumberGutterComponent
+class LineNumberGutterComponent extends TiledComponent
   dummyLineNumberNode: null
 
-  constructor: ({@onMouseDown, @editor, @gutter}) ->
+  constructor: ({@onMouseDown, @editor, @presenter, @gutter}) ->
```
jssln: We should avoid passing the `presenter` around. Are you planning to refactor this later?

as-cii: Not really, as we are already passing it in `LinesComponent` as well as in `LinesTileComponent`. I agree, though, this is an unhealthy dependency and, as far as I remember, we use it only for testing purposes (e.g. `lineNodeForScreenRow` and `lineNumberNodeForScreenRow`).

I am going to get rid of it everywhere, exception made for `TextEditorComponent` which should be the only object to know about it. The correspondence between tiles and rows will still be handled by `TextEditorPresenter` but its conversion methods will be called in one place only (i.e. `TextEditorComponent`). I'll /cc you in the relevant upcoming commit.

Nice suggestion! :sparkler:

nathansobo:
@as-cii I think the only thing holding this back is that the tiles need an opaque background for the same reasons the lines did, subpixel anti-aliasing. Good seeing you in Nashville and hope you had a good trip back!

as-cii:
> I think the only thing holding this back is that the tiles need an opaque background for the same reasons the lines did, subpixel anti-aliasing.

For sure! I also introduced some assertions to ensure we don’t lose this behavior in the future. Thank you for spotting this! :bow:

nathansobo:
Thanks for finishing this out. It'll be a great relief to GPUs worldwide.

PR #7119, opened 2015-06-05
Update linux.md

verrazanof:
For Arch, `export PYTHON=/usr/bin/python2` should correct for `export python=/usr/bin/python2`


kevinsawicki:
Thanks for updating this.

PR #7123, opened 2015-06-05
Re-enable tree-view and tabs when sublime-tabs is installed but not loaded

kevinsawicki:
The sublime-tabs package disables the tree-view and tabs packages so re-enable them when sublime-tabs is installed but not loaded because it is deprecated.

Refs https://github.com/atom/tree-view/issues/467
Refs https://github.com/atom/tabs/issues/162


iamstarkov:
thanks a lot

PR #7127, opened 2015-06-05
Bring Subpixel Anti-Aliasing Back

as-cii:
After merging #6733 there have been [some reports of text looking blurred](https://discuss.atom.io/t/text-rendering-seems-blurry-compared-to-other-editors/13927/15). The reason for that is we started rendering tiles with a transparent background: this doesn't play well with Chrome, which disables sub pixel rendering, thereby making text less crisp.

This PR attempts to bring text clarity back. It's a bit more cumbersome than it seems on the surface because, in order to make tiles opaque, we need to distribute highlights across tiles. We plan to :ship: this in time for the next release and so it's quite urgent. :warning: 

If anyone wants to give feedback meanwhile we make progress on this it would be highly appreciated! The code still needs to be polished, but feature-wise everything should already work as expected.

Thanks to anyone who will take the time to have a look! :bow: 

/cc: @atom/feedback


YuriSolovyov:
any before / after screenshots ?

nathansobo:
selection-bugs-2

Looks like find and replace decorations are hanging around when destroyed. Here is me using `cmd-g` to move to the next search result in a file. Note the old highlight is retained.

thomasjo:
Yeah, seeing similar behavior with <kbd>cmd-d</kbd> functionality.

src/text-editor-presenter.coffee
```diff
@@ -1236,7 +1254,12 @@ class TextEditorPresenter
     range = marker.getScreenRange()
 
     if decoration.isDestroyed() or not marker.isValid() or range.isEmpty() or not range.intersectsRowRange(@startRow, @endRow - 1)
-      delete @state.content.highlights[decoration.id]
+      tileStartRow = @tileForRow(range.start.row)
```
as-cii: This code is repeated a couple of times now. I am thinking we could extract something like:

```coffee
iterateTiles: (startRow, endRow, fn)
```

Which could basically perform the conversion from row to tile along with the necessary setup (e.g. initialize the tile state).

as-cii:
I confirm this actually brings sub pixel anti-aliasing back! :tada: :tada: :tada: 

I wasn't able to spot the difference because, at some point, I disabled the anti-aliasing setting on OS X (`System Preferences -> General`):

screen shot 2015-06-06 at 20 26 25

After enabling it again we can easily see the difference:

| Before | After |
| --------- | ------- |
| screen shot 2015-06-07 at 15 05 09 | screen shot 2015-06-07 at 15 04 40 |

I have also adapted the old specs to cover this new behavior and everything is green now. :green_heart:

paulcbetts:
Hm, so on Windows, while this doesn't hurt text rendering in High DPI, it doesn't _help_ either. Compare the text rendering of the Treeview with the editor text (built against 515ae7a):

capture

nathansobo:
@paulcbetts Did this regress in 0.206 or has it always been an issue. Do we have no choice but to use the CPU instead of the GPU on high-DPI windows?

nathansobo:
The following thought is related to this PR but does not need to block it from merging:

In reviewing this again, I just realized that we probably should disable the GPU entirely for mini editors. We don't assign a background color on mini editors because certain themes were animating the selection of a mini editor with a highlight flash, and we weren't able to pick it up and transfer it to the GPU layer in time. As a fix I just made the background of the GPU layer transparent in mini editors, since they don't scroll anyway. But this also means that subpixel AA probably doesn't work in mini editors. Would be good to revisit this at some point and perhaps not use `translate3d` on editors that aren't intended to scroll. That said, merging the tiled gutter is more important.

paulcbetts:
> Did this regress in 0.206 or has it always been an issue. 

I'm not sure when this showed up, but it's definitely been broken in pre-0.206.0

> Do we have no choice but to use the CPU instead of the GPU on high-DPI windows?

So, one idea I had recently that might help (though I haven't tried it yet!), is to always ensure that the values passed to translate3d are a multiple of `window.devicePixelRatio` - I'll give this a go and report back

nathansobo:
@paulcbetts Interesting idea. Very curious to hear your findings.

as-cii:
This PR should be good to go now and it's (hopefully) going to restore text clarity, at least for some users. If no issues arise in the meantime, we will :shipit: in `0.208`! :tada: 

> So, one idea I had recently that might help (though I haven't tried it yet!), is to always ensure that the values passed to translate3d are a multiple of window.devicePixelRatio - I'll give this a go and report back

That would be huge, thanks @paulcbetts. I think it's probably good to continue this conversation into its own issue (or an existing one which fits the topic) and discuss about further improvements there.

Thanks everyone for all the hard work and for the awesome support! :bow:

PR #7129, opened 2015-06-05
Avoid redundant decoration state updates in presenter

maxbrunsfeld:
In #7023, when we switched to a fully batched system of computing decoration states, I neglected to remove a couple of unnecessary incremental decoration state updates. They are now gone.

This exposes another performance benefit of the batched system: large find-and-replace searches are now much faster because we avoid thousands of unnecessary presenter state updates as highlights are created.

A search in an 8800-line file with 20k results now takes about **2.7 seconds** on my machine, whereas before it took **4 seconds**.

/cc @nathansobo


nathansobo:
:dollar:

PR #7130, opened 2015-06-05
Load large files with fewer features but much better performance

nathansobo:
Pictured below is Atom opening an 18.5 MB file (repeated copies of jQuery):

large-file
### Less Is More

The lion's share of overhead when working with text files goes to supporting the following features:
- Soft wrap
- Folds
- Syntax analysis

As files get larger, this overhead becomes more problematic. Yes, we can reduce this overhead and plan to do so, but it's also true that these features are less likely to be critical when editing really huge files. So this PR disables them for files above 2MB.

In return we can do the following:
- Compute placeholder `TokenizedLine`s on demand instead of up-front. This skips all of the processing I was trying to perform in the background on #7060, which can now be closed. It also avoids the memory hit of storing horizontal spatial metadata for any lines we haven't yet viewed. I'm leaving lines cached once we compute them once to avoid creating the same line over and over again. We could potentially use an LRU ("least recently used") cache eviction strategy if we want to cap consumed memory, but I'm not sure it's worth prioritizing.
- Delegate straight through to `TokenizedBuffer` from `DisplayBuffer` rather than storing a second copy of lines, since there are no folds and soft wraps. This isn't a huge win but it does avoid an extra potentially huge array of references.
- Not block the UI thread as we perform the initial tokenization. This is really long overdue to be backgrounded, but this is a good stopgap.
### Metrics

Now that we've cleared a huge bottleneck, new bottlenecks are revealing themselves in the initial load of a large file. At least on an SSD, I/O really isn't high on the list yet. Here's a flame graph for loading that 18.5 MB file:

screenshot_2015-06-05_22_11_18

There's plenty of low hanging :apple: in here. I've highlighted a few things that stood out to me.

@maxbrunsfeld: `TextBuffer` might be a good first candidate for `Patch` after all, at least in terms of getting that data structure optimized for insertion. `SpanSkipList` insertion is by far the biggest remaining cost.

From a memory perspective, the `SpanSkipList` is also our biggest offender when `TokenizedLine`s are out of the way:

screenshot 2015-06-05 22 10 07
### Closing Thoughts

I still want to push the efficiency of the structures that support Atom's full feature set, and we have a bunch of ideas for how to do so. But in the meantime, this is a much better compromise than an error message.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -66,6 +66,25 @@ describe "TextEditor", ->
 
       expect(editor.tokenizedLineForScreenRow(0).invisibles.eol).toBe '?'
 
+  describe "when the editor is constructed with the largeFileMode option set to true", ->
+    it "loads the editor but doesn't tokenize", ->
+      editor = null
+
+      waitsForPromise ->
+        atom.workspace.open('sample.js', largeFileMode: true).then (o) -> editor = o
+
+      runs ->
+        buffer = editor.getBuffer()
+        expect(editor.tokenizedLineForScreenRow(0).text).toBe buffer.lineForRow(0)
+        expect(editor.tokenizedLineForScreenRow(0).tokens.length).toBe 1
+        expect(editor.tokenizedLineForScreenRow(1).tokens.length).toBe 2 # soft tab
+        expect(editor.tokenizedLineForScreenRow(12).text).toBe buffer.lineForRow(12)
+        expect(editor.tokenizedLineForScreenRow(0).tokens.length).toBe 1
+        expect(editor.getCursorScreenPosition()).toEqual [0, 0]
+        editor.insertText('hey"')
+        expect(editor.tokenizedLineForScreenRow(0).tokens.length).toBe 1
+        expect(editor.tokenizedLineForScreenRow(1).tokens.length).toBe 2 # sof tab
+
   describe "when the editor is constructed with an initialLine option", ->
     it "positions the cursor on the specified line", ->
       editor = null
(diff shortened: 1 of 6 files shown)


mnquintana:
Awesome! I am a bit concerned about the experience for users though – I suspect that without some kind of extra feedback, loading a large file only to find that it has no syntax highlighting and no soft wrap support would still feel "broken". For those looking at plain text log files it wouldn't matter, but for those who are looking at files that would normally be associated with a grammar, I think it would be quite noticeable. What do you think?

nathansobo:
@mnquintana Yeah, it's not ideal. I'm not sure popping up a notification to apologize makes it any better though. And not loading the file at all is even worse. So considering this is something we could conceivably ship next week and that there's probably no way to support everything in time for 1.0, I still think it's the right way to go. After all, I've seen people disable syntax highlighting in vim when working with big files.

hastebrot:
Interesting solution. I like it. :+1:

maxbrunsfeld:
So much improvement. So little code. :ship:

alexandernst:
I'd still vote for a "loading" progress bar. It's not really clear from the gif if Atom is just freezing trying to open to file or is it actually doing *something*, so imho a progress bar-like (or even a spinning dent) would be better than not showing anything.

Zireael07:
Yeah, a progress bar would be nice.

nathansobo:
This is just the improvement I could achieve quickly. The pause you see is a blocking computation that would freeze the progress bar anyway. It could be broken into 10ms chunks but that complicates things and still impacts responsiveness. It could be backgrounded but the majority of time is spent populating a reference-heavy data structure that I want to replace and would be awkward to serialize. I think we should revisit the progress bar after switching to a simpler line indexing structure and seeing what we can gain in linear performance first. I want a progress bar too, but I don't want to waste time serializing a data structure I want to replace.

nathansobo:
Okay, after thinking about this for a bit I think I will try to yield to the UI thread when building the line index so we can display a progress bar. True background computation in another process will have to wait for a new data structure, but it may be possible to give some feedback (with reduced responsiveness) before them.

PR #7144, opened 2015-06-07
Expose `screenPositionForPixelPosition` on TextEditorElement

jssln:
This PR exposes the `screenPositionForPixelPosition` method currently on the `TextEditor` model. We expose this method on the view instead of the model because it should only be called when the editor is on screen.


nathansobo:
@jssln Are you actually able to use this method now that we have tiled rendering? What do you use it for?

jssln:
@nathansobo What release number was tiled rendering?
We recently used this in a patch to our click-to-symbol package, though this is probably short term because @oclbdk is working on hyperclick.

jssln:
I'm gonna close this PR because this isn't actually useful.

PR #7145, opened 2015-06-07
Item provides save panel options

jessegrosjean:
This changes the pane.saveItemAs logic to look for a getSaveOptions method and if present those options will be passed into the save dialogue. This add flexibility, giving items more control over how the save dialog works. I need it so that I can include filter types in the save dialogue. But I expect there will be other uses cases going forward.

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -783,11 +783,12 @@ class Atom extends Model
   showSaveDialog: (callback) ->
     callback(showSaveDialogSync())
 
-  showSaveDialogSync: (defaultPath) ->
-    defaultPath ?= @project?.getPaths()[0]
+  showSaveDialogSync: (options={}) ->
     currentWindow = @getCurrentWindow()
     dialog = remote.require('dialog')
-    dialog.showSaveDialog currentWindow, {title: 'Save File', defaultPath}
+    options.title ?= 'Save File'
+    options.defaultPath ?= @project?.getPaths()[0]
+    dialog.showSaveDialog currentWindow, options
 
   saveSync: ->
     if storageKey = @constructor.getStateKey(@project?.getPaths(), @mode)
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -786,6 +786,8 @@ class Atom extends Model
   showSaveDialogSync: (options={}) ->
     if _.isString(options)
       options = defaultPath: options
+    else
+      options = _.clone(options)
     currentWindow = @getCurrentWindow()
     dialog = remote.require('dialog')
     options.title ?= 'Save File'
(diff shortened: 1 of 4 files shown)


src/atom.coffee
```diff
@@ -783,11 +783,12 @@ class Atom extends Model
   showSaveDialog: (callback) ->
     callback(showSaveDialogSync())
 
-  showSaveDialogSync: (defaultPath) ->
-    defaultPath ?= @project?.getPaths()[0]
+  showSaveDialogSync: (options={}) ->
```
kevinsawicki: I know this isn't a public API but I think it might be nice to still preserve the old behavior.

Basically just test of the first argument is a string and retain the old behavior in this case.

Does that sound okay?

src/pane.coffee
```diff
@@ -492,8 +492,9 @@ class Pane extends Model
   saveItemAs: (item, nextAction) ->
     return unless item?.saveAs?
 
-    itemPath = item.getPath?()
-    newItemPath = atom.showSaveDialogSync(itemPath)
+    saveOptions = item.getSaveOptions?() or {}
```
kevinsawicki: Usually we use `? {}` instead of `or {}` since it handles fall backs pretty well.

src/atom.coffee
jessegrosjean: Yes, I'm just updated to fix.

src/pane.coffee
jessegrosjean: Thanks I didn't know about that syntax… I've switched to use it.

src/atom.coffee
```diff
@@ -783,11 +783,14 @@ class Atom extends Model
...
+    if _.isString(options)
+      options = defaultPath: options
     currentWindow = @getCurrentWindow()
     dialog = remote.require('dialog')
-    dialog.showSaveDialog currentWindow, {title: 'Save File', defaultPath}
+    options.title ?= 'Save File'
```
kevinsawicki: We probably shouldn't modify a passed in object, mind cloning it?

jessegrosjean: Ok, is there an existing util method to do this… I see underscore has a shallow clone, but I don't to a deep clone. Currently (and seems reasonable for the future) the options will be JSON so if there's no existing util method I can just convert to JSON and back again to deep clone. What do you suggest?

kevinsawicki: `_.clone()` should be fine since the only properties being touched are at the root level.

jessegrosjean: Ahh right, thanks. I've pushed that change.

kevinsawicki:
Awesome, thanks for this :+1:

Issue #7096, opened 2015-06-04
Require just lsb-core in rpm specfile, not whole lsb


kparal:
In #6988 the rpm spec file got changed to:
```
Requires: lsb 
```
However, that pulls the *whole* lsb package, including all submodules. On Fedora 21, this pulls:
```
redhat-lsb                   
redhat-lsb-core              
redhat-lsb-cxx               
redhat-lsb-desktop           
redhat-lsb-languages         
redhat-lsb-printing          
redhat-lsb-submod-multimedia 
redhat-lsb-submod-security   
```
and those packages in turn pull a lot of other dependencies. That seems to be completely unnecessary. IIUIC, you just need `lsb_release` to be present, nothing else, so just the lsb core module.

Please adjust this to:
```
Requires: lsb-core-noarch
```
That should ensure that only lsb core module is required, nothing else. I tested this in Fedora 21 (`sudo yum install 'lsb-core-noarch'`) and in OpenSUSE 13.2 (`sudo zypper install 'lsb-core-noarch'`) and it seems to work just fine.

Thank you.

mnquintana:
Thanks for the heads up! I'll change this shortly.

kparal:
Thanks, I'll try the next build :-)

kparal:
Looks great, atom dependencies are no longer excessive.

PR #7157, opened 2015-06-07
Only use lsb-core instead of full suite of lsb packages

mnquintana:
Fixes https://github.com/atom/atom/issues/7096

`lsb` is quite a large package – we should just use what we need (which is `lsb-core`)

Tested on Fedora 21 and OpenSUSE 13.2

/cc @kevinsawicki


kevinsawicki:
:+1:

PR #7158, opened 2015-06-07
Enable "Downloading Update" context menu entry

### LINKED ISSUE #7128: [Windows] Update mechanism menu text
The dropdown menu entry stays at "Checking for Update" and then switches to "Restart to Update" never showing "Downloading Update" in between.

winstliu:
Previously `update-downloaded` would get emitted right after `update-available` which would suppress the "Downloading Update" menu entry.

I'm not too literate at Coffeescript so hopefully I put `@emit update-available` where it actually belongs :P.

Fixes #7128.


kevinsawicki:
Awesome, thanks for this :+1:

PR #7169, opened 2015-06-08
Update Project::getRepositories() example

mnquintana:
The example was using a `project` global instead of `atom.project`.


kevinsawicki:
:+1:

PR #7170, opened 2015-06-08
Fixed missing icon when building from source

erfanio:
When building from source there is a problem with the `atom.desktop`. The icon that it points to doesn't exit.
I added `atom.png` to get unpacked from app.asar and changed `atom.desktop` to point to and fixed the problem


kevinsawicki:
Thanks for this

atrotors:
It's OK, I'm really existed, this my first conttibution to a big broject

PR #7173, opened 2015-06-08
Show a warning prompt when opening files over 20MB

nathansobo:
screenshot 2015-06-08 23 52 18

I'm now considering chunking the population of the line index on the UI thread to get us a progress bar before 1.0. I want to fix some bugs first, so until then, this will represent a soft limit to keep people from straying into the danger zone with larger files.

### DIFF
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -329,9 +329,9 @@ class Project extends Model
 
     if fileSize >= 20 * 1048576 # 20MB
       choice = atom.confirm
-        message: 'Atom can have issues with files over 20MB.'
-        detailedMessage: "Do you still want to try to load this?"
-        buttons: ["Open", "Cancel"]
+        message: 'Atom currently freezes during the loading of very large files.'
+        detailedMessage: "Do you still want to load this file?"
+        buttons: ["Proceed", "Cancel"]
       if choice is 1
         error = new Error
         error.code = 'CANCELLED'
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -323,8 +323,22 @@ class Project extends Model
         # allow ENOENT errors to create an editor for paths that dont exist
         throw error unless error.code is 'ENOENT'
 
-    @bufferForPath(filePath).then (buffer) =>
-      @buildEditorForBuffer(buffer, options)
+    absoluteFilePath = @resolvePath(filePath)
+
+    fileSize = fs.getSizeSync(absoluteFilePath)
+
+    if fileSize >= 20 * 1048576 # 20MB
+      choice = atom.confirm
+        message: 'Atom can have issues with files over 20MB.'
+        detailedMessage: "Do you still want to try to load this?"
+        buttons: ["Open", "Cancel"]
+      if choice is 1
+        error = new Error
+        error.code = 'CANCELLED'
+        throw error
+
+    @bufferForPath(absoluteFilePath).then (buffer) =>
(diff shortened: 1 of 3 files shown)


src/project.coffee
```diff
@@ -323,8 +323,22 @@ class Project extends Model
...
 
-    @bufferForPath(filePath).then (buffer) =>
-      @buildEditorForBuffer(buffer, options)
+    absoluteFilePath = @resolvePath(filePath)
+
+    fileSize = fs.getSizeSync(absoluteFilePath)
+
+    if fileSize >= 20 * 1048576 # 20MB
+      choice = atom.confirm
+        message: 'Atom can have issues with files over 20MB.'
```
izuzak: I like that the message is short :zap:, but wonder if we could tell the user a bit more about what's about to happen:

1. the performance might be reduced
2. some features will be disabled (e.g. syntax highlighting)

Here's a draft suggestion for discussion:

"This file if over 20MB and Atom performance might be reduced. In order to improve performance, some editing features will be disabled, e.g. syntax highlighting and soft wrap."

src/project.coffee
```diff
@@ -323,8 +323,22 @@ class Project extends Model
...
+        detailedMessage: "Do you still want to try to load this?"
```
izuzak: Perhaps we can drop "try to":

> "Do you still want to load this file?"

src/project.coffee
izuzak: Oh wait, these editor features are disabled when the file is larger than 2MB, not 20MB. Not sure why I thought these limits were the same.

Would showing that message for files larger than 2MB be something we'd consider? I kind of like the idea of informing users that some features of the editor were disabled. Since a 2MB file limit would be much lower, we might consider adding a config setting to disable this warning so that users who often open large files don't need to click the button every single time?

Just through out some ideas. :bulb:

izuzak:
:zap: for this, @nathansobo, and all your other work on improving large file support in Atom

src/project.coffee
YuriSolovyov: > This file if over 20MB, __so__ Atom performance might be reduced

kevinsawicki:
:ship: ?

nathansobo:
I've updated the message. It's embarrassing so I'm going to work on the progress bar for 1.0.

nathansobo:
What do you guys think? I kinda think we should just have no message for a bit until I can get a progress bar in there.

v3ss0n:
I tested 20MB + . still responsive. ( Except random 200ms-500ms lags) . So i think we can live with it.

kevinsawicki:
Is this still something we want to ship before 1.0?

nathansobo:
@izuzak Do you want to massage the text to your liking and merge this?

nathansobo:
It's worth noting that the bad experience were worried about is a hard freeze when processing the file upon load.

izuzak:
> @izuzak Do you want to massage the text to your liking and merge this?

Sure -- just made a small tweak in d32f307 and tested everything manually one more time, so going to merge this. I think it's worth having this in 1.0 -- it will really help in situations when someone accidentally clicks on a super large file in the tree view, for example. :+1: 

Thanks again for this, @nathansobo! :guitar:

v3ss0n:
@nathansobo @izuzak  thanks alot , that prevent me from clicking huge build files after Project wide search.

PR #7176, opened 2015-06-09
Only use package directory name as fallback

### LINKED ISSUE #7174: Enable/disable depends on package directory name, not `name` from package.json
We have some packages whose directory names do not correspond to their package name (e.g. `nuclide/file-tree` for `nuclide-file-tree`, `nuclide/hack` for `nuclide-hack`), and I noticed that the Package Manager (i.e. Settings > Packages) doesn't allow me to enable/disable them.

I looked at `~/.atom/config.cson`, and it seems like just the base name of the directory is being written into the `core.disabledPackages` config. I believe it needs to be the `name` field from the `package.json` for it to work properly.

mnquintana:
Fixes #7174

/cc @kevinsawicki

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -43,8 +43,8 @@ class Package
           @normalizeMetadata(metadata)
         catch error
           throw error unless ignoreErrors
-      else
-        metadata.name ?= packageName
+
+    metadata.name ?= packageName
 
     if includeDeprecatedAPIs and metadata.stylesheetMain?
       deprecate("Use the `mainStyleSheet` key instead of `stylesheetMain` in the `package.json` of `#{packageName}`", {packageName})
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -44,8 +44,7 @@ class Package
         catch error
           throw error unless ignoreErrors
       else
-        metadata ?= {}
-        metadata.name = packageName
+        metadata.name ?= packageName
 
     if includeDeprecatedAPIs and metadata.stylesheetMain?
       deprecate("Use the `mainStyleSheet` key instead of `stylesheetMain` in the `package.json` of `#{packageName}`", {packageName})
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -44,6 +44,7 @@ class Package
         catch error
           throw error unless ignoreErrors
 
+    metadata ?= {}
     metadata.name ?= packageName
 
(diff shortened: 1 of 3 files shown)


src/package.coffee
```diff
@@ -43,9 +43,9 @@ class Package
           @normalizeMetadata(metadata)
         catch error
           throw error unless ignoreErrors
-
-    metadata ?= {}
-    metadata.name = packageName
+      else
+        metadata ?= {}
```
kevinsawicki: This still needs to run in all cases since ` CSON.readFileSync(metadataPath)` might return null or if an error is thrown and `ignoreErrors` is `true`

src/package.coffee
```diff
@@ -43,9 +43,9 @@ class Package
...
+        metadata.name = packageName
```
kevinsawicki: I think this should be `metadata.name ?= packageName`

mnquintana:
@kevinsawicki Addressed your comments!

mnquintana:
I'll take a look at the failing specs later today.

kevinsawicki:
Yeah, now I'm remembering why it is this way and I think it was for the following case:

`atom.packages.loadPackage` takes a name or a path and if a name is specified it is looked up by folder name in `~/.atom/packages`.

So you could end up with something:

* Package with a `name` value of `foo` in  `~/.atom/packages/bar/package.json`
* `atom.packages.loadPackage('bar').name` would be `foo` in this case which is kind of weird since the person requested a bar package by name and got back a foo package.

I'm not saying we shouldn't change this behavior but I just wanted to point out this case in case we want to further refine.

This would really only affect linked/local packages since installed packages from atom.io will always have consistent folder/name values.

kevinsawicki:
Thanks for this :+1:

mnquintana:
@kevinsawicki There was one failing spec: https://travis-ci.org/atom/atom/jobs/65983846 Should we be concerned about that?

kevinsawicki:
Whoops, yeah, taking a look now, thanks for mentioning it.

kevinsawicki:
Looks like the `normalize-package-data` library that gets called before this sets `name` to `''` when unset so the `?=` check wasn't taking affect originally, pushed a fix in https://github.com/atom/atom/commit/0edce07446a7a4f424ce621c20d83c8fea955267

Issue #7178, opened 2015-06-09
Resize pane cursor is inconsistent with tree view cursor


rmartin:
As noted by @trkoch in the https://github.com/atom/atom/pull/5902#issuecomment-95063910 thread, the resize pane cursor is inconsistent with the resize cursor. Specifically the tree view cursor uses the col-resize cursor where the resize panes use the ew-resize for horizontal and ns-resize for vertically split panes. This should be updated to using the split panes leverage col-resize cursor for horizontal and row-resize for vertically split panes.

The following screen capture highlights the issue:
![atomresizepanecursorbug](https://cloud.githubusercontent.com/assets/119151/8053169/e7d78bbe-0e42-11e5-9048-00cccbd61ca8.gif)

mnquintana:
Added in https://github.com/atom/atom/pull/7179

PR #7179, opened 2015-06-09
Fix inconsistency with pane resize cursors.

rmartin:
Resolves inconsistency as raised by https://github.com/atom/atom/issues/7178 and originally raised by  @trkoch in https://github.com/atom/atom/pull/5902#issuecomment-95063910. Updates the pane resize cursor to leverage col-resize for horizontal and row-resize for vertically split panes.

Cursor Reference: 
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Screen Capture showing fix:
atomresizepanecursorfix

### DIFF
--- a/static/panes.less
+++ b/static/panes.less
@@ -25,7 +25,7 @@ atom-pane-container {
       width: 100%;
       height: 8px;
       margin-top: -4px;
-      cursor: ns-resize;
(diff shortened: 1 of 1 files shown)


mnquintana:
Thanks for this! :ship:

PR #7182, opened 2015-06-09
Update linux.md

verrazanof:
Some one have both python2 and python3 in archlinux will have some problem, there python will links to python3, so they should `ln -s /usr/bin/python2 /usr/bin/python` before building atom.

### DIFF
--- a/docs/build-instructions/linux.md
+++ b/docs/build-instructions/linux.md
@@ -31,6 +31,7 @@ Ubuntu LTS 12.04 64-bit is the recommended platform.
 
 * `sudo pacman -S gconf base-devel git nodejs libgnome-keyring python2`
 * `export python=/usr/bin/python2` before building Atom.
+* if you have both python2 and python3, it would be best to `ln -s /usr/bin/python2 /usr/bin/python`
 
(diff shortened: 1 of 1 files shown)


mnquintana:
Have you looked at the updated line at https://github.com/atom/atom/pull/7182/files#diff-cde9cda8108b3777e5d97e5806a4f51fR33? It now explicitly says to link to python2, so it shouldn't be an issue if both Python 2 and 3 are installed, unless I'm misunderstanding?

despairblue:
```
 $ pacman -Qo /usr/bin/python
/usr/bin/python is owned by python 3.4.3-2
```
overwriting a file that is owned by a package is not a good idea. That might break other applications that depend on python being python3 in archlinux (also, updating python will just link it to python3 again). Exporting it before building and configuring npm to use python2 is a perfectly acceptable solution.

mnquintana:
> Exporting it before building and configuring npm to use python2 is a perfectly acceptable solution.

:+1: I agree.

@verrazanof Thanks for the submission, but I think the existing build step should be sufficient (and is a lot safer).

Issue #6909, opened 2015-05-22
Remove cursor when there are multiple cursors if Cmd-Clicking on it


mnquintana:
From a discussion in Slack – it would be really nice if <kbd>Cmd / Ctrl-Click</kbd>ing on a cursor when you have multiple cursors removed that cursor. This would make it a lot easier to fine tune your set of multiple cursors after the fact.

albertorestifo:
:+1:

PR #7225, opened 2015-06-11
Add remove cursor when there are multiple cursors and cmd clicking on it

lpommers:
Closes https://github.com/atom/atom/issues/6909

deselect

 `Cmd / Ctrl-Click` -ing on a cursor when there are multiple cursors will remove that cursor. 

If there is only the one cursor and it is `Cmd / Ctrl-Click`-ed, nothing will happen.

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1578,19 +1578,19 @@ describe "TextEditorComponent", ->
 
         describe "when there are multiple cursors, and one of the cursor's screen position is the same as the mouse click screen position", ->
           it "removes a cursor at the mouse screen position", ->
-            editor.setCursorScreenPosition([3,4])
-            editor.addCursorAtScreenPosition([5,2])
+            editor.setCursorScreenPosition([3, 4])
+            editor.addCursorAtScreenPosition([5, 2])
             editor.addCursorAtScreenPosition([7, 5])
-            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3,4]), metaKey: true))
+            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3, 4]), metaKey: true))
             nextAnimationFrame()
-            expect(editor.getSelectedScreenRanges()).toEqual [[[5,2], [5,2]], [[7,5], [7,5]]]
+            expect(editor.getSelectedScreenRanges()).toEqual [[[5, 2], [5, 2]], [[7, 5], [7, 5]]]
 
         describe "when there is a single cursor and the click occurs at the cursor's screen position", ->
           it "neither adds a new cursor nor removes the current cursor", ->
             editor.setCursorScreenPosition([3, 4])
-            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3,4]), metaKey: true))
+            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3, 4]), metaKey: true))
             nextAnimationFrame()
-            expect(editor.getSelectedScreenRanges()).toEqual [[[3,4], [3,4]]]
+            expect(editor.getSelectedScreenRanges()).toEqual [[[3, 4], [3, 4]]]
 
     describe "when a non-folded line is double-clicked", ->
       describe "when no modifier keys are held down", ->
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1569,11 +1569,28 @@ describe "TextEditorComponent", ->
           expect(editor.getSelectedScreenRange()).toEqual [[3, 4], [5, 6]]
 
       describe "when the command key is held down", ->
-        it "adds a cursor at the nearest screen position", ->
-          editor.setCursorScreenPosition([3, 4])
-          linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([5, 6]), metaKey: true))
-          nextAnimationFrame()
-          expect(editor.getSelectedScreenRanges()).toEqual [[[3, 4], [3, 4]], [[5, 6], [5, 6]]]
+        describe "the current cursor position and screen position do not match", ->
+          it "adds a cursor at the nearest screen position", ->
+            editor.setCursorScreenPosition([3, 4])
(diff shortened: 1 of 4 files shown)


spec/text-editor-spec.coffee
```diff
@@ -916,6 +916,12 @@ describe "TextEditor", ->
         editor.setCursorBufferPosition([3, 1])
         expect(editor.getCurrentParagraphBufferRange()).toBeUndefined()
 
+    describe "getCursorAtScreenPosition(screenPosition)", ->
+      it "returns the cursor at the given screenPosition", ->
+        cursor1 = editor.addCursorAtScreenPosition([0, 2])
+        cursor2 = editor.getCursorAtScreenPosition(cursor1.getScreenPosition())
+        expect(cursor2.marker).toBe cursor1.marker
```
maxbrunsfeld: I would think that `expect(cursor2).toBe(cursor1)` would work here, right? I'd prefer that because comparing the markers is a bit indirect; it makes it sound like `getCursorAtScreenPosition` might return a different `Cursor` instance.

src/text-editor.coffee
```diff
@@ -1561,6 +1561,17 @@ class TextEditor extends Model
   setCursorBufferPosition: (position, options) ->
     @moveCursors (cursor) -> cursor.setBufferPosition(position, options)
 
+  # Essential: Get a {Cursor} at given screen coordinates {Point}
+  #
+  #* `position` A {Point}
```
maxbrunsfeld: Just needs a space before the `*` there:
```
# * `position` A {Point}
```

src/text-editor.coffee
```diff
@@ -1561,6 +1561,17 @@ class TextEditor extends Model
...
+  #
+  # Returns the first matched {Cursor} or undefined
+  getCursorAtScreenPosition: (position) ->
+    cursorsAtPoint = @cursors.filter (cursor) ->
+      cursorPosition = cursor.getScreenPosition()
+      cursorPosition.row is position.row and cursorPosition.column is position.column
+    if cursorsAtPoint.length > 0 then cursorsAtPoint[0] else undefined
```
maxbrunsfeld: For performance, it'd be nice to avoid the temporary array allocated by `filter`. Maybe instead:

```coffee
for cursor in @cursors
  return cursor if cursor.getScreenPosition().isEqual(position)
undefined
```

spec/text-editor-component-spec.coffee
```diff
@@ -1569,11 +1569,28 @@ describe "TextEditorComponent", ->
...
+            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3, 4]), metaKey: true))
+            nextAnimationFrame()
+            expect(editor.getSelectedScreenRanges()).toEqual [[[5, 2], [5, 2]], [[7, 5], [7, 5]]]
+
+        describe "when there is a single cursor and the click occurs at the cursor's screen position", ->
+          it "neither adds a new cursor nor removes the current cursor", ->
+            editor.setCursorScreenPosition([3, 4])
+            linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([3, 4]), metaKey: true))
+            nextAnimationFrame()
+            expect(editor.getSelectedScreenRanges()).toEqual [[[3, 4], [3, 4]]]
```
maxbrunsfeld: :+1:

maxbrunsfeld:
I like this! Looks good except for some pretty minor comments that I left.

lpommers:
@maxbrunsfeld thanks for the feedback and the encouragement. It's always a bit scary/intimidating making a the first pull to a large, mature project.  I'm implemented your suggestions and made some new commits.

src/text-editor.coffee
```diff
@@ -1561,6 +1561,16 @@ class TextEditor extends Model
...
 
+  # Essential: Get a {Cursor} at given screen coordinates {Point}
+  #
+  # * `position` A {Point} or {Array} of `[row, column]`
+  #
+  # Returns the first matched {Cursor} or undefined
+  getCursorAtScreenPosition: (position) ->
+    for cursor in @cursors
+      return cursor if cursor.getScreenPosition().isEqual(position)
+    undefined
```
mnquintana: Would it maybe be better to return null here instead of undefined?

lpommers: @mnquintana I think you're right. `null` would be a better fit if there is no matching cursor. I'll make another commit

spec/text-editor-spec.coffee
maxbrunsfeld: Just realized that the code that I commented on here was actually based on an existing test right below. So what you did was totally reasonable. I see that in a later commit, you actually updated that other test and not this one. Could you do the same thing here?

src/text-editor.coffee
maxbrunsfeld: I would say don't use `return` here, just write either `undefined` or `null`. The one gotcha with coffee-script is that you can't just remove the last line: if the last statement in a function is a `for` loop, `coffee-script` will use that loop as a *value*, collecting its results into an array.

`null` and `undefined` are both fine. Sometimes I use `undefined` for this purpose, to be consistent with `object["non-existent-key"]` or `array[outOfBoundsIndex]`.

lpommers: @maxbrunsfeld that's interesting about coffee-script using a `for` as a value if it's the last line - i'm pretty new to coffee-script so I had no idea about that. 

The reason I was considering a `return` was from the contribution documentation:
>Using a plain return when returning explicitly at the end of a function.
>Not return null, return undefined, null, or undefined

Not sure how I should proceed here, `undefined` or `null`.

maxbrunsfeld: I interpret that rule from the docs as applying when the method is not called for its return value. I would say leave the code as is.

spec/text-editor-spec.coffee
lpommers: @maxbrunsfeld done.

maxbrunsfeld:
Great; thanks for adding this @lpommers!

mnquintana:
This is awesome :sparkles:

lpommers:
thanks @maxbrunsfeld and @mnquintana for making so easy to get involved and contribute. I'm looking forward to contributing more. And thanks for making `beginner` labels as it's a really great way to jump in and I probably wouldn't have known where to get started otherwise

philipgiuliani:
Niceee :) Until now i had to start from beginning when i had to do 50 selections and clicked one wrong 😂

PR #7234, opened 2015-06-11
Avoid saving state when closing windows due to auto-update

### LINKED ISSUE #7156: Restore session after install upgrade and relaunch
When 'install upgrade and relaunch' atom can reload all opened files before.

Liu Jiang.

### LINKED ISSUE #7214: All windows not remembered if exiting with unsaved files.
With multiple windows open and some of them with files with unsaved changes, when quitting Atom it prompts about those unsaved files. Regardless of the option chosen for the prompt, only the window with the last prompt to close is reopened.

Experiencing this on _OS X 10.10.3_ and _Atom 0.208_.

maxbrunsfeld:
This also fixes some cases where the window state would incorrectly be saved in the middle of exiting.
- [x] Avoid incorrectly saving state as windows are closed while restarting to install an update.
- [x] Make sure that `quitting` state doesn't incorrectly persist if quit is cancelled due to unsaved buffers.

Fixes #7156
Fixes #7214

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -105,7 +105,7 @@ class AtomApplication
         app.quit()
         return
     @windows.splice(@windows.indexOf(window), 1)
-    @saveState() unless window.isSpec or @quitting
+    @saveState() unless window.isSpec
 
   # Public: Adds the {AtomWindow} to the global window list.
   addWindow: (window) ->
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
/cc @kevinsawicki?

Do you know if there's a better way to do this? I always want to call `saveState()` when a window is removed, unless we're in the process of quitting. I'm not sure how to detect that the app is quitting except for manually setting some state (`@quitting`) in some known cases. But then quitting could be aborted because of unsaved changes, so would need to somehow detect that and set `@quitting` back to false, which seems kind of gnarly.

kevinsawicki:
> so I need to somehow detect that and set @quitting back to false, which seems kind of gnarly.

Yeah...that is gnarly. Maybe emit an event to the browser process when unload is aborted from the render process? Like after here perhaps? https://github.com/atom/atom/blob/9ab57884e8701d5b000fe41a65903f7e91249aff/src/window-event-handler.coffee#L67

maxbrunsfeld:
@kevinsawicki could you review? Testing on Ubuntu 14 now.

maxbrunsfeld:
Ok, this seems to fix the above issues from my testing on Ubuntu.

PR #7239, opened 2015-06-11
Handle native cut command in render process too

viddo:
I imagine the cut command was missed when this was originally done in https://github.com/atom/atom/commit/bb1bcc233ae37f6e72f3172db30908885c768919 ?

I verified it working manually by [re-building atom](https://github.com/atom/atom#building) and adding an input element with the `native-key-bindings` class through the developer console, e.g.

``` js
i=document.createElement('input'); i.classList.add('native-key-bindings'); i.type="text"; i.tabIndex="-1"; document.body.appendChild(i); i.style.position="absolute"; i.style.zIndex=3; i.style.bottom=0; i.style.border="5px solid red"; i;
```

and verifying that the cut now works with this change.


kevinsawicki:
Awesome, thanks for adding this :shipit:

PR #7281, opened 2015-06-15
New default editor font

### LINKED ISSUE #7220: Why is the font so big by default?
The default font size seems extremely large compared to every other editor I've used. Why is it so big?

simurai:
This PR replaces the current default font for the editor. Fixes #7220

**Before**: The default is `Inconsolata`, but only if it's installed. If not, the fallback font is `Monaco`. The problem is that Inconsolata is quite smaller than other fonts, making it look bigger when changing the font in the settings. Or it's already too big when Inconsolata isn't installed. It's also a problem in mini editors.

**After**: New default is `Menlo`.
- Much closer to other fonts.
- Already comes preinstalled with OS X from Snow Leopard or newer.
- Is also the default in other editors/applications, like Xcode.

default-font

Text in mini editors in some/most themes will look bigger. But they already do that, if Inconsolata isn't installed or if a user changes the font in the settings. So with this change, themes hopefully make the `@input-font-size` a bit smaller and then it would be more consistent with other fonts.

screen shot 2015-06-15 at 4 16 11 pm
### Possible alternatives
1. `Monaco`. It's the current fallback font if Inconsolata is not installed.
   - Pro: Less people would be affected by this change.
   - Con: Monaco doesn't have _italic_ and _bold_ version, which some themes use. Also in MD files. Menlo is newer and arguably more legible.
2. Bundle a new default font (see: #4201).
   - Pro: Could be any font, not just a system font. Would be consistent across all platforms.
   - Con: Even more users would be affected (also Win+Linux). Staying close to the platform might be a good idea because people are already used to that font from other editors/applications. Like Xcode uses Menlo and Visual Studio uses Consolas, in Atom it would be the same.
### Who is affected?
- Users that haven't changed the Font Family in the settings and have Inconsolata installed or are on OS X.
- Users that haven't changed the Font Size in the settings.
- Users that haven't changed the Line Height in the settings.
### I prefer the old default

For users that prefer the old default, it can be changed back in the settings to:
- Font Family: `Inconsolata, Monaco, Consolas, 'DejaVu Sans Mono', monospace`
- Font Size: `16`
- Line Height: `1.3`

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -108,7 +108,7 @@ module.exports =
         maximum: 100
       lineHeight:
         type: ['string', 'number']
-        default: 1.44
+        default: 1.5
       showInvisibles:
         type: 'boolean'
         default: false
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -108,7 +108,7 @@ module.exports =
         maximum: 100
       lineHeight:
         type: ['string', 'number']
-        default: 1.4
+        default: 1.44
       showInvisibles:
         type: 'boolean'
         default: false
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -103,12 +103,12 @@ module.exports =
         default: ''
       fontSize:
         type: 'integer'
-        default: 16
+        default: 14
         minimum: 1
         maximum: 100
       lineHeight:
         type: ['string', 'number']
-        default: 1.3
+        default: 1.4
       showInvisibles:
         type: 'boolean'
         default: false
--- a/static/text-editor-light.less
+++ b/static/text-editor-light.less
@@ -4,8 +4,7 @@
 
 atom-text-editor {
   display: block;
-  font-family: Inconsolata, Monaco, Consolas, 'DejaVu Sans Mono', monospace;
-  line-height: 1.3;
+  font-family: Menlo, Consolas, 'DejaVu Sans Mono', monospace;
 }
 
 atom-text-editor[mini] {


thomasjo:
LGTM :+1: 

This makes much more sense than the current default; this aligns with other editors (as mentioned in the PR). I say with :boat: this.

simurai:
The only thing I wasn't sure is the new Line Height of `1.4`. It's "visually" a bit smaller than currently with Inconsolata's, but `1.5` seemed too big. Maybe an inbetween of `1.44` could also be considered:

default-font-1-44

In pixels it would be
- 1.4: `19px`
- 1.44: `20px`

mnquintana:
Looks good to me :+1:  although I think when we communicate this change we should be clear that it affects more than just OS X (because it's removing Inconsolata for everyone, and defining a new default font size and line-height for everyone).

I prefer the spacing with `line-height: 1.44`; it's just sort of an aesthetically random-looking number :laughing: 

How does it look with `1.5`?

simurai:
> although I think when we communicate this change we should be clear that it affects more than just OS X (because it's removing Inconsolata for everyone, and defining a new default font size and line-height for everyone).

Yeah, right. Should definitely go into the release notes.

> aesthetically random-looking number

Since it snaps to whole pixels, it could be anything between `1.43` and `1.49`.

> How does it look with 1.5?

Sorry, the gif above was actually `1.4` and `1.5`. Here `1.4`, `1.44` and `1.5`.

default-font-lh
menlo-1 4
menlo-1 44
menlo-1 5

philipgiuliani:
Im for 1.4 or 1.44 (prefered). The font looks good :)

v3ss0n:
How about directly packaging Menlo font ?

mnquintana:
I like 1.44 or 1.5 the best of the three.

simurai:
Ok, how about `1.5`.

- It's a nice ratio of `3:2`.
- With 14px font size, line height will be exactly `21px` instead of `20.1599998474121px` with 1.44. It should snap to 20px, but still feels better to have a round number.
- And maybe the more important one.. with a bit larger line height, it's easier to go down in font size and still be able to read.

izuzak:
> Ok, how about 1.5.

:+1: that looks good to me

mnquintana:
> Ok, how about 1.5.

:+1: Sounds great to me

kevinsawicki:
:ship: This looks great to me using Source Code Pro at 13px.

PR #7309, opened 2015-06-17
Revert "Fix inconsistency with pane resize cursors."

basarat:
Reverts atom/atom#7179 

Tree view uses `ew-resize` as shown here : https://github.com/atom/tree-view/blob/f0530bca9217ef6c2bb9fc63d815c9d153d2c898/styles/tree-view.less#L58

`row-resize` and `col-resize` don't work on windows (for whatever reason). 

Screenshots : https://github.com/atom/atom/pull/7179#commitcomment-11715003

### DIFF
--- a/static/panes.less
+++ b/static/panes.less
@@ -25,7 +25,7 @@ atom-pane-container {
       width: 100%;
       height: 8px;
       margin-top: -4px;
-      cursor: row-resize;
+      cursor: ns-resize;
       border-bottom: none;
     }
   }
@@ -37,7 +37,7 @@ atom-pane-container {
       width: 8px;
       height: 100%;
       margin-left: -4px;
-      cursor: col-resize;
+      cursor: ew-resize;
       border-right: none;
     }
   }


mnquintana:
Ah, looks like we were working on this at the same time! I didn't want to revert this for all platforms though - https://github.com/atom/atom/pull/7310 keeps the `col-` and `row-resize` cursors for Mac and Linux, and fixes them for Windows.

rmartin:
hey @basarat and @mnquintana ,

Thanks for brining this up. I wasn't aware of the differences between platforms but it looks like that was also pointed out in the Electron thread in general. https://github.com/atom/electron/issues/1361. 

I'm not sure if there is a standard way of system targeting or if that's how the treeview does it, but https://github.com/atom/atom/pull/7310 is more correct as your fix would cause the inconsistency again in OSX which is the problem I was originally trying to resolve.

Cheers,
Roy

basarat:
Cool. Thanks guys :rose:

mnquintana:
> I'm not sure if there is a standard way of system targeting or if that's how the treeview does it

Yup, the approach I used in https://github.com/atom/atom/pull/7310 is the standard way of targeting styles by platform in Atom ([and is what's used in tree-view](https://github.com/atom/tree-view/blob/master/styles/tree-view.less#L55)).

PR #7320, opened 2015-06-17
Fix TextEditorComponent Race Condition

### LINKED ISSUE #7306: Orphaned line numbers
While screensharing with @nathansobo I got into a situation where there were some orphaned gutter elements. 

screen shot 2015-06-16 at 2 59 54 pm

The gutter component only had a reference to the line numbers at the top of the screen. Not sure of exact repro steps, but i wasnt able to get rid of them.

I was working on a long md doc with softwrap on. I had been working on this doc for 2 days. The atom window had been through several resizes.

cc @as-cii maybe you have insight into this.

as-cii:
Closes #7306

We started noticing that when a `TextEditor` pane got split, the same view for `Gutter` was being shared amongst several models, thereby making the same DOM element accessible simultaneously by more than one object. This made us experience _orphaned line numbers_, caused by two instances of `LineNumberGutterComponent` mutating `.line-numbers` at the same time.

This is a typical race condition which I would normally address by understanding and possibly locking the correct order in which operations should happen. However, I believe in this situation we shouldn’t actually care about ordering at all, since I think views should be kept “local”, thus avoiding to expose them to the world and/or reusing them across other views (either accidentally as in this case or on purpose).

@jssln @nathansobo: I assume the `ViewRegistry` came into play because it was handy to get the view straight from `atom.views.getView`. If this is the use case and I am not missing something, we may expose a method under `TextEditorElement` which, in turn, would interrogate `TextEditorComponent` to fetch the related DOM node. What do you think? :thought_balloon:

### DIFF
--- a/spec/custom-gutter-component-spec.coffee
+++ b/spec/custom-gutter-component-spec.coffee
@@ -16,9 +16,6 @@ describe "CustomGutterComponent", ->
     decorationsWrapperNode = gutterComponent.getDomNode().children.item(0)
     expect(decorationsWrapperNode.classList.contains('custom-decorations')).toBe true
 
-  it "makes its view accessible from the view registry", ->
-    expect(gutterComponent.getDomNode()).toBe atom.views.getView(gutter)
-
   it "hides its DOM node when ::hideNode is called, and shows its DOM node when ::showNode is called", ->
     gutterComponent.hideNode()
     expect(gutterComponent.getDomNode().style.display).toBe 'none'
--- a/spec/custom-gutter-component-spec.coffee
+++ b/spec/custom-gutter-component-spec.coffee
@@ -16,6 +16,9 @@ describe "CustomGutterComponent", ->
     decorationsWrapperNode = gutterComponent.getDomNode().children.item(0)
     expect(decorationsWrapperNode.classList.contains('custom-decorations')).toBe true
 
+  it "makes its view accessible from the view registry", ->
+    expect(gutterComponent.getDomNode()).toBe atom.views.getView(gutter)
+
   it "hides its DOM node when ::hideNode is called, and shows its DOM node when ::showNode is called", ->
     gutterComponent.hideNode()
     expect(gutterComponent.getDomNode().style.display).toBe 'none'
--- a/spec/gutter-container-component-spec.coffee
+++ b/spec/gutter-container-component-spec.coffee
@@ -107,24 +107,22 @@ describe "GutterContainerComponent", ->
 
       gutterContainerComponent.updateSync(testState)
       expect(gutterContainerComponent.getDomNode().children.length).toBe 2
-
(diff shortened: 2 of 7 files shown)


as-cii:
On a side note, this is somewhat similar to what was fixed in https://github.com/atom/atom/pull/6570. I think the root cause is the same, although I believe we should just go for a different solution to avoid this problem altogether in the future. :rocket:

nathansobo:
@as-cii The original intent of storing gutter elements in the view registry was to maintain consistency with the overall approach to accessing views in Atom. Everywhere else so far, we can always call `atom.views.getView` with a model object to access the view. The trade-off of course is that we can only support a single model instance in the workspace, but I considered this simpler than having a bunch of localized entry points. For example, to do this for the workspace, panes, etc, we still would have needed some global way to get the workspace element. Then, on the workspace, we would have needed individual methods for obtaining views for different model objects. After agonizing over it, I opted for a 1:1 relationship between models and views globally for the workspace, so accessing a view is always the same.

I'm a bit confused about why this is still happening. When we construct the gutter component and re-obtain our element, we always clear its contents [here](https://github.com/atom/atom/blob/0a7f6ae18743d76a40c567dc2cd893ad702c72f3/src/line-number-gutter-component.coffee#L16). How is this still going wrong?

jssln:
From my understanding, one `TextEditorComponent` corresponds to one `LineNumberGutterComponent`, but one `TextEditor` may correspond to several `TextEditorComponent`s(?). 

So the race condition really arises from that fact. (At least there must be some place the 1:1 correspondence breaks down in order to create the multiple `LineNumberGutterComponent`s.) It does seem the straightforward thing to do is either to make that not the case, or to make separate DOM views as in this PR.

nathansobo:
After talking this through with @as-cii, I'm starting to think that exposing the raw gutter nodes to API users is unnecessary. For example, there are other cases where we expose model objects without their corresponding views: cursors and selections. @jssln is there anything critical you need direct node access for? If not, I think life is simpler if we remove it.

jssln:
One thing we do now is register a click listener on the entire gutter node.

nathansobo:
@as-cii That's a pretty compelling use case. We could add an official API to replicate it, but it sounds somewhat annoying for now.

as-cii:
@nathansobo @jssln: agreed, although I wouldn’t go down the path of creating a `1:1` correspondence between `TextEditor` and `TextEditorComponent`. It certainly sounds like a good move, but it feels somewhat risky and we should probably avoid doing it right now.

Another thing @nathansobo brought up in chat was that at some point we could replace the global `ViewRegistry`, but that sounds complicated as well.

So, I will simply ditch the first commit and keep the last one: this means that there will be no visible changes but the race condition will be addressed by making `TextEditorComponent` aware of its destruction.

If everyone is :cool: with that, I am going to :ship: it as soon as the builds gets :green_heart:.

jssln:
:+1:

PR #7338, opened 2015-06-19
styles.less examples

simurai:
This PR adds some examples (commented). Fixes https://github.com/atom/welcome/issues/34

It also changes the line about Less and mentions that you can use normal CSS (or Less). Some people might get discouraged when they read about having to learn Less.

### DIFF
--- a/dot-atom/styles.less
+++ b/dot-atom/styles.less
@@ -4,8 +4,7 @@
  * This stylesheet is loaded when Atom starts up and is reloaded automatically
  * when it is changed.
  *
- * If you are unfamiliar with LESS, you can read more about it here:
- * http://www.lesscss.org
+ * Uncomment the examples below to test or add your own CSS or Less.
  */
 
 // style the background color of the tree view
--- a/dot-atom/styles.less
+++ b/dot-atom/styles.less
@@ -8,18 +8,18 @@
  * http://www.lesscss.org
  */
 
+// style the background color of the tree view
 .tree-view {
-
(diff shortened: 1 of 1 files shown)


dot-atom/styles.less
```diff
@@ -4,22 +4,21 @@
  * This stylesheet is loaded when Atom starts up and is reloaded automatically
  * when it is changed.
  *
- * If you are unfamiliar with LESS, you can read more about it here:
- * http://www.lesscss.org
```
izuzak: I kind of liked the link here which helps users learn more about less. If you don't feel strongly about removing this, could I talk you into keeping the link or adding some other links to point users to :books:?

izuzak:
Other than that one comment/question: :+1: :zap:

dot-atom/styles.less
simurai: Yeah, sure. I rephrased it a bit and added the line with the link back.

dot-atom/styles.less
```diff
@@ -4,22 +4,23 @@
...
+ * Uncomment the examples below to try it out. Or add your own CSS or Less.
```
mnquintana: > try it out

Try what out? Probably better to be more explicit here.

simurai: Refers to "reloaded automatically", but I shuffled it around. Now it's under the "Examples".

Still not sure about `(To see them, uncomment and save)`.

kevinsawicki:
:+1:

PR #7341, opened 2015-06-19
Remove event listeners from orphaned gutters

### LINKED ISSUE #7337: Wrong Folding Operation
Originally reported by @maxbrunsfeld:

> I'm seeing another weird gutter behavior, and I'm wondering if you think it's the same bug. Occasionally the fold buttons in the gutter cause incorrect fold operations:
> 
> fold-bug
> 
> Unfolding works as expected, but when I click the fold icon next to line 9 to create a fold, instead of creating a fold at line 9, it scrolls the editor slightly, and then creates a fold at line 6.
> 
> Creating folds from the keyboard (via editor:fold-current-row) works correctly, so the problem seems to be related to the gutter.
> 
> This only happens to me intermittently. Is there any code that I can run when I get into this state to verify whether or not it's the same bug?

I am able to reproduce this similarly to the bug we had in #7306:

* Write a few paragraphs of text (ensure it enables vertical scrolling).
* Scroll to the very bottom.
* Split the pane (I split it right, but I think left or anything else will do).
* Ensure that splitting the pane has caused the editor on the left to scroll up.
* Scroll the left panel again to the very bottom.
* Close the right panel.
* Fold any line.
* The wrong line was folded. :boom:

as-cii:
Fixes #7337.

We caught another :bug: related to #7320. Basically, when multiple `TextEditorComponent`s were created for the same instance of `TextEditor`, we weren’t removing some event listeners from the recycled gutter DOM node which, in turn, had multiple handlers listening and acting when a certain event was fired (e.g. `onMouseDown` and `onMouseClick`).

@maxbrunsfeld: could you give this a shot and look out for the bug you reported? I am quite confident this is going to fix it but I’d like to have your feedback as well. :bow: 

I think this is likely to be the last fix for all the so-called “global state” issues that we have encountered over the last days. However, I won’t feel very confident about changing these areas of the code without introducing new :beetle:s. I like @jssln's proposal :bulb: of having a `1:1` correspondence between `TextEditorComponent` and `TextEditor` and I think we have two ways of handling it:
1. When panes are split, always create a new instance of `TextEditor`.
2. When `TextEditorElement` is detached, avoid creating a new instance of `TextEditorComponent` and simply use the previous one.

@nathansobo: this PR can be merged without addressing the above proposal, but I am curious about your :thought_balloon: on this. My two cents are that we should go for 2), as it’s kinda pointless to always create new instances as proposed in 1). I am sure, though, that there are good reasons for that behavior to exist  and it’d be great if you could share your thoughts about them.

Thanks everyone! :bow:

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -743,6 +743,16 @@ describe "TextEditorComponent", ->
         beforeEach ->
           gutterNode = componentNode.querySelector('.gutter')
 
+        describe "when the component is destroyed", ->
+          it "stops listening for folding events", ->
+            component.destroy()
+
+            lineNumber = component.lineNumberNodeForScreenRow(1)
+            target = lineNumber.querySelector('.icon-right')
+            target.dispatchEvent(buildClickEvent(target))
+
+            expect(nextAnimationFrame).toBe(noAnimationFrame)
+
         it "folds and unfolds the block represented by the fold indicator when clicked", ->
           expect(lineNumberHasClass(1, 'folded')).toBe false
 
@@ -1726,6 +1736,14 @@ describe "TextEditorComponent", ->
 
     beforeEach ->
       gutterNode = componentNode.querySelector('.gutter')
+
+    describe "when the component is destroyed", ->
+      it "stops listening for selection events", ->
+        component.destroy()
+
+        gutterNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenRowInGutter(1)))
+
+        expect(editor.getSelectedScreenRange()).toEqual [[0, 0], [0, 0]]
 
     describe "when the gutter is clicked", ->
       it "selects the clicked row", ->
--- a/src/gutter-container-component.coffee
+++ b/src/gutter-container-component.coffee
@@ -16,7 +16,12 @@ class GutterContainerComponent
 
     @domNode = document.createElement('div')
     @domNode.classList.add('gutter-container')
-    @domNode.style.display = 'flex';
+    @domNode.style.display = 'flex'
+
+  destroy: ->
+    for {name, component} in @gutterComponents
+      component.destroy?()
+    return
 
   getDomNode: ->
     @domNode
--- a/src/line-number-gutter-component.coffee
+++ b/src/line-number-gutter-component.coffee
@@ -18,6 +18,10 @@ class LineNumberGutterComponent
     @domNode.addEventListener 'click', @onClick
     @domNode.addEventListener 'mousedown', @onMouseDown
 
+  destroy: ->
+    @domNode.removeEventListener 'click', @onClick
+    @domNode.removeEventListener 'mousedown', @onMouseDown
+
   getDomNode: ->
     @domNode
 
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -106,6 +106,7 @@ class TextEditorComponent
     @mounted = false
     @disposables.dispose()
     @presenter.destroy()
+    @gutterContainerComponent?.destroy()
     window.removeEventListener 'resize', @requestHeightAndWidthMeasurement
 
   getDomNode: ->


nathansobo:
Yeah, let's just properly recycle the node for now. Open to proposals about the association of views with models in general, but I'd like to put it off because it's a pretty nuanced situation where a lot of concerns have to be balanced. I view these bugs as a fundamental misunderstanding when the code was written that the gutter node might be recycled when the editor model is moved. I think that if we want to make view elements conveniently available from the model layer, there is going to be a set of issues like this in one way or another. But perhaps it can be arranged a different way.

nathansobo:
Sorry, I didn't address your question: Reusing the component sounds like a good idea. My only concern with recycling the component is that we need to tear down some global handlers like the DOM polling in the scheduler. So long as we leave the component in a state where it can be garbage collected on detachment and then reestablish these things when reattached, I think it's worth trying.

maxbrunsfeld:
Oops, commented in the wrong place. To reiterate: I just verified that this fixes the problem. :ship:

as-cii:
> So long as we leave the component in a state where it can be garbage collected on detachment and then reestablish these things when reattached, I think it's worth trying.

It seems to me like this may cause more complexity than I initially thought: the resulting effect could be, again, code that’s risky to change and we should probably think to something simpler.

> Yeah, let's just properly recycle the node for now.

--

> Oops, commented in the wrong place. To reiterate: I just verified that this fixes the problem. :ship: 

:+1: 

> I view these bugs as a fundamental misunderstanding when the code was written that the gutter node might be recycled when the editor model is moved. I think that if we want to make view elements conveniently available from the model layer, there is going to be a set of issues like this in one way or another. But perhaps it can be arranged a different way.

One first thought is that we should avoid letting clients access DOM nodes: after all, they are *used inside a view* but they are *not* a view (in Atom terms, at least). This allows us to have more control over objects’ lifetime and expose only a subset of events (such as `TextEditorElement#onGutterClicked`), so that when destroying a `TextEditorComponent` (or any other component, for that matter) we can handle DOM cleanup and creation without affecting the client at all.

Also, I think it may be interesting to explore an alternative to `ViewRegistry` where things are evented, and clients just don’t look for a certain instance in the application's global state, but rather rely on events being fired and respond to them accordingly: this could clarify that no assumptions should be made on views lifetime. I am not clear on what the entry point for subscribing to such events should be but I feel like this could be promising.

@nathansobo: what do you think? I am not sure if this is just another way of complicating things; I’ll spawn a background thread on my brain and think more about it over the next days.

nathansobo:
We may want to start another issue with some starter proposals, but I'll give you my initial thoughts.

I agree that giving access to DOM nodes *anywhere* is tricky and potentially hazardous. I think this statement applies to almost any area of the system however, from the workspace element all the way down. Direct DOM access is an invitation to do things that are bad ideas.

On the other hand, a core value we've adhered to so far in Atom is to try to always provide people with the tools they need to accomplish their objective, while trying to anticipate their needs and guide them down safer and better-abstracted paths. I think providing access to DOM elements is a part of that. As was the case with @jssln's click handler, it allows people to move forward with their plans without waiting for us to build APIs, and then we can follow behind and pave over the cow paths. We can't anticipate what people are going to want to do perfectly, but if we give them full access to the DOM we don't have to.

Rather than cutting DOM access off, I'd rather establish patterns and tools to help people work with the DOM properly. The DOM interaction scheduling methods on `atom.views` are an example of this, and we should be thinking about making them official and promoting them. Lifecycle hooks on specific elements may be another good place to explore. Then, we can try to stay on top of what people are doing with this access and try to build better APIs for achieving it. I definitely agree that a generalized mechanism for handling clicks in the editor is sorely needed.

I'm still learning the limits of this "open by default" philosophy. Maybe I'm taking it too far, but so far I haven't seen any concrete examples to convince me it's a bad idea. However, I have changed my mind on enough things so far in this project to keep an open mind.

If you want to keep discussing this overall question of direct element access, how far to take it, and how to express it in the API, we can open an issue. Keep in mind that the current approach to view retrieval is–for better or worse–baked into the 1.0 API. Any alternatives we design will have to co-exist with the existing system. That said, I think it can be enlightening and valuable to discuss completely different ways of doing things, if only to inform iteration on what we have.

Thanks for taking a broad, long-term view!

PR #7348, opened 2015-06-20
fix doc for TextEditor setText

tony612:


kevinsawicki:
Thanks for adding this

Issue #5905, opened 2015-03-09
Uncaught TypeError: Cannot read property 'isComment' of undefined


TobEnd:
1. Editing an unsaved .xml file

**Atom Version**: 0.182.0
**System**: linux 3.18.6-1-ARCH
**Thrown From**: Atom Core
### Stack Trace

Uncaught TypeError: Cannot read property 'isComment' of undefined

```
At /usr/share/atom/resources/app/src/tokenized-buffer.js:401

TypeError: Cannot read property 'isComment' of undefined
  at TokenizedBuffer.module.exports.TokenizedBuffer.isFoldableCodeAtRow (/usr/share/atom/resources/app/src/tokenized-buffer.js:401:71)
  at TokenizedBuffer.module.exports.TokenizedBuffer.isFoldableAtRow (/usr/share/atom/resources/app/src/tokenized-buffer.js:396:19)
  at TokenizedBuffer.module.exports.TokenizedBuffer.updateFoldableStatus (/usr/share/atom/resources/app/src/tokenized-buffer.js:384:25)
  at TokenizedBuffer.module.exports.TokenizedBuffer.handleBufferChange (/usr/share/atom/resources/app/src/tokenized-buffer.js:351:20)
  at /usr/share/atom/resources/app/src/tokenized-buffer.js:56:24
  at Emitter.module.exports.Emitter.emit (/usr/share/atom/resources/app/node_modules/event-kit/lib/emitter.js:82:11)
  at TextBuffer.module.exports.TextBuffer.applyPatch (/usr/share/atom/resources/app/node_modules/text-buffer/lib/text-buffer.js:594:20)
  at BufferPatch.module.exports.BufferPatch.applyTo (/usr/share/atom/resources/app/node_modules/text-buffer/lib/buffer-patch.js:74:21)
  at Transaction.module.exports.Transaction.applyTo (/usr/share/atom/resources/app/node_modules/text-buffer/lib/transaction.js:60:29)
  at History.module.exports.History.abortTransaction (/usr/share/atom/resources/app/node_modules/text-buffer/lib/history.js:165:24)
  at History.module.exports.History.transact (/usr/share/atom/resources/app/node_modules/text-buffer/lib/history.js:122:16)
  at TextBuffer.module.exports.TextBuffer.transact (/usr/share/atom/resources/app/node_modules/text-buffer/lib/text-buffer.js:687:27)
[...]

PR #7350, opened 2015-06-20
Add assertion mechanism

nathansobo:
There are a few hard-to-reproduce uncaught exceptions lurking in Atom that are giving a subset of our users a hard time. Until now, in order to solve these kinds of mysteries, our only option other than raw luck with reproduction steps or a brilliant stroke of insight was to be more aggressive about throwing exceptions when getting into an invalid state. But, of course, we're extremely reluctant to do that because who wants to introduce an even greater chance of throwing an exception?!

This PR enables a better approach by adding a new tool to our box: `atom.assert`. This global method takes a condition, a description message, and an optional metadata callback as follows:

``` coffee
atom.assert a is b, "a is b", -> {a, b}
```

Now, whenever we have uncaught exceptions, our goal should be as follows:
- If appropriate, do something to suppress the exception. This may lead to other, weirder errors, but we also may survive without the user ever noticing. This is obviously an extremely short-term strategy, and it will need to be applied judiciously.
- Start forming hypotheses and seasoning in efficient-to-run assertions that can investigate how and where things might be going wrong. This will enable us to start chasing errors to their source as we put out releases.

An example of this approach is included in this PR, referencing #5905.

@atom/feedback Once we start doing this, it could be fairly important to know if an assertion has failed when someone is reporting a bug. We'll be reporting assertion failures to bugsnag and logging them to the console. Do we want some other indicator? My main thing is that we need to be able to insert assertions without interrupting a user's flow.

It's also worth noting that this changes the API of `atom.onDidThrowError` to yield the entire `Error` object to subscribers. It's made backward-compatible by adding the old event properties to the error object. I found no packages that use this method, so it should be pretty safe to change this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "link": "0.30.0",
     "markdown-preview": "0.150.0",
     "metrics": "0.51.0",
-    "notifications": "0.56.0",
+    "notifications": "0.57.0",
     "open-on-github": "0.37.0",
     "package-generator": "0.39.0",
     "release-notes": "0.53.0",
--- a/spec/atom-spec.coffee
+++ b/spec/atom-spec.coffee
@@ -142,6 +142,37 @@ describe "the `atom` global", ->
         expect(atom.openDevTools).not.toHaveBeenCalled()
         expect(atom.executeJavaScriptInDevTools).not.toHaveBeenCalled()
 
+  describe ".assert(condition, message, metadata)", ->
+    errors = null
+
+    beforeEach ->
+      errors = []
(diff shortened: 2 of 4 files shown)


lee-dohm:
I might want assertions to be logged to a file in `~/.atom` when running in Dev Mode, perhaps trimmed to a certain size on exit. It looks though like you have an assertion failed event, so what I'm suggesting could easily be provided by a package.

thomasjo:
Nice, I like this concept. Will definitely make it easier to track down some of the more elusive bugs.

nathansobo:
Reverting this in the name of hyper-conservatism for 1.0.

PR #7351, opened 2015-06-20
Mark Clipboard::write's metadata argument as optional

mnquintana:
`Clipboard::write()`'s `metadata` argument is optional, but it's currently marked as required - this updates the docs to mark it as optional.

/cc @atom/feedback


kevinsawicki:
:balloon:

PR #7368, opened 2015-06-22
Add Fedora 22 Install Directions To README

amayer5125:
Fedora 22 uses dnf
yum command will not work


amayer5125:
I wasnt sure if this should be a sub-direction under the Red Hat section or if it should be its own section. I can change it if needed.

kevinsawicki:
Awesome, thanks so much :ship:

Issue #3377, opened 2014-08-25
Make application:about open internally


kevinsawicki:
Similar to chrome, map the about menu internally instead of using the Mac OS X-only API that shows a simple dialog with only the version and app name.

Possible other content to put in this view:

* Packages bundled
* Packages installed

This would enable the about command to work on all platforms, refs https://github.com/atom/atom/issues/2536

Other ideas welcome.

PR #7377, opened 2015-06-22
Add cross-platform "About Atom" view

mnquintana:
Closes https://github.com/atom/atom/issues/3377
Closes https://github.com/atom/about/issues/2

This PR adds a new package - [`about`](https://github.com/mnquintana/atom-about) - dedicated to viewing information about your Atom installation within Atom, across all platforms.

About screenshot

_Note: The color on the Atom logo is the hover state._

Right now it doesn't do a whole lot (you can copy the version number though), but I have several enhancements planned right now (showing all Atom org contributors, checking for updates, checking if latest commit if using built Atom, installed packages, etc), but that should make it a bit easier to review. I plan on continuing to maintain this package if it gets merged (along with the core team and the rest of the maintainers and anyone else who wants to pitch in :smile:).

If the team decides to go forward with this, then we should:
- [x] Transfer repo to Atom org
- [x] Transfer atom.io ownership to atom
- [x] Remove Command Palette entry for About: About Atom

Let me know what you think!

/cc @simurai @atom/feedback

### DIFF
--- a/menus/linux.cson
+++ b/menus/linux.cson
@@ -181,6 +181,8 @@
       { label: 'Report Issue', command: 'application:report-issue' }
       { label: 'Search Issues', command: 'application:search-issues' }
       { type: 'separator' }
+      { label: 'About Atom', command: 'application:about' }
+      { type: 'separator' }
     ]
   }
 ]
--- a/menus/win32.cson
+++ b/menus/win32.cson
@@ -184,6 +184,8 @@
       { label: 'Report Issue', command: 'application:report-issue' }
       { label: 'Search Issues', command: 'application:search-issues' }
       { type: 'separator' }
+      { label: 'About Atom', command: 'application:about' }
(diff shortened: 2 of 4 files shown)


batjko:
I think a dedicated About view makes sense, and it is indeed pretty standard nowadays to have information about available new releases there as well, so that could be moved right in there.

I'm not sure if listing the repo contributors wouldn't be too much, though?

mnquintana:
@benogle The same specs that are failing in this branch are also failing on master, so I think this is safe to merge, unless you also wanna give it a run through Janky just to make sure.

mnquintana:
Would it be worth adding some code to remove user-installed `about` for a few versions to prevent conflicts?

kevinsawicki:
This shipped today, thanks so much :+1:

thomasjo:
Bloody brilliant! :heart:

PR #7395, opened 2015-06-24
Add Pane::onWillRemoveItem event.

oggy:
Hi,

I'd love to have this here `will-remove-item` event on Pane. The motivation is for oggy/tab-switcher#11, where I want to determine in a `did-change-active-item` hook if an item is being activated because the user clicked the tab, or if it's just because the active item was removed and this was next to it (which I could do by setting a flag between `will-remove-item` and `did-remove-item`).

There are probably many possible solutions here, but this seemed the most generic. WDYT?

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -209,6 +209,12 @@ describe "Pane", ->
       expect(item2.isDestroyed()).toBe true
       expect(events).toEqual [{item: item2, index: 1}]
 
+    it "invokes ::onWillRemoveItem() observers", ->
+      events = []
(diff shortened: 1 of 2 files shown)


thomasjo:
I haven't tested this yet, but it looks good to me. Specs and everything. Great work!

Pretty sure we need to wait for v1.0 to land before we merge this though, as we need to extremely conservative about adding and changing new behavior at this point.

oggy:
Ok, fair enough. Thanks for the update!

benogle:
Thanks!

PR #7411, opened 2015-06-25
Mini editor with TextEditor::setMini

fstiewitz:
`TextEditor` has a function called `setMini` that toggles the mini state of an editor. The problem is that it only "changes" the `TextEditor` itself and `TextEditorPresenter`, but not `TextEditorElement`.
Before fix:
unfixed-setmini
After fix:
fixed-setmini
Code:

``` coffeescript
@content: ->
    ...
    @div class:'editor-item', =>
            @subview 'test_data', new TextEditorView({mini:true, placeholderText:'Test Input'})
    ...

toggleMini: ->
    editor = @test_data.getModel()
    editor.setMini(not editor.isMini())
```


benogle:
Seems legit. Can you write a spec?

deprint:
Added a spec and it passes.

PR #7426, opened 2015-06-25
Fix Arch build instructions

drewkett:
PYTHON environment variable needs to be in all caps


mnquintana:
Thanks for fixing this!

PR #7427, opened 2015-06-26
Updating Linux.md

fusion809:
Adding some techniques for installing Atom that have worked for me on 32 bit Linux distros. All the techniques on this page that I've been able to try on 32-bit systems, at least, have failed, hence why I am proposing these changes.

### DIFF
--- a/docs/build-instructions/linux.md
+++ b/docs/build-instructions/linux.md
@@ -14,18 +14,54 @@ Ubuntu LTS 12.04 64-bit is the recommended platform.
       * You might need to run this command as `sudo`, depending on how you have set up [npm](https://github.com/joyent/node/wiki/Installing-Node.js-via-package-manager#ubuntu-mint-elementary-os).
   * development headers for [GNOME Keyring](https://wiki.gnome.org/Projects/GnomeKeyring)
 
+
+### Installing Node.js from source
+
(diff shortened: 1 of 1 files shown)


mnquintana:
@atom/feedback Do we want to maintain our own Node installation instructions? I feel like we should continue to defer to Node's own installation instructions.

thomasjo:
> Do we want to maintain our own Node installation instructions? I feel like we should continue to defer to Node's own installation instructions.

I agree. We should definitely not maintain our own which would have to be kept in sync with the official instructions. It's destined to be outdated.

If there are deviations from the official documentation, then we should definitely document that, but otherwise let's simple refer/link to the official story :+1:

fusion809:
OK, well how about we include a link to the [source installation instructions](https://github.com/joyent/node/wiki/Installation#installing-on-linux) as sometimes installing via package managers fail, hence it's nice to give users a backup method of installation.

thomasjo:
Linking to *official* documentation is fine with me at least since that doesn't put any burden on us beyond simply ensuring it points to a valid URI. So :+1: from me on the link suggestion.

joshaber:
@fusion809 Are you still interested in pursuing this?

fusion809:
Na, I no longer work on 32-bit systems so I've forgotten what I learnt from operating on them.

joshaber:
Cool, I'll close this then. Thanks!

PR #7434, opened 2015-06-26
Make Alt-F behaviour native on Windows

rjattrill:
Make Alt-F work natively - to open file menu - as per standard behaviour on Windows platform.

Refer issue #474


benogle:
Is this still an issue? Can you elaborate on the issue? Issue #474 doesnt look like it is in reference to this change.

50Wliu:
What about keybindings such as `Alt-H` or `Alt-E`?  This doesn't seem like it'll fix all of them.

damieng:
Seems to be no more traction on this so going to close. If somebody wants to answer the questions and push it through please reopen.

PR #7521, opened 2015-06-28
Update license in package.json

ghost:
Use of license objects or a "licenses" property containing an array of license objects are now deprecated. Instead, use SPDX expressions. https://docs.npmjs.com/files/package.json#license


mnquintana:
Thanks for fixing this! :sparkles:

Issue #7339, opened 2015-06-19
Disable "Check for updates at startup"


klaszlo:
I'm often developing en route, that means via tethering my mobile phone internet access.
So I have a pretty limited data access for each month (1GiB in my case).

So I would like to minimize internet usage. So I don't need automatical update checking at each startup.
It consumes about 2MB of data. Since I restart atom a lot, it piles up quickly.

I searched through the forum, looked around settings, but could not find anything similar of disabling update checking.

Can an option be added to the settings page? It would be greatly appreciated. 
Sometimes, when I'm home, does not bother much this additional internet usage.

I hope you appreciate my feature request.

lee-dohm:
As a workaround, you can build Atom from source. When you do (even if it is the same commit as a released version), Atom disables update checking.

robyoder:
+1, but ooh nice tip, @lee-dohm

Zireael07:
I believe this is a dupe of an existing issue, but I can't find the other one atm - I only remember arguments such as being on a shitty modem connection in Indonesia.

50Wliu:
@Zireael07 Found it: #1882.

AnwarShah:
@lee-dohm I built atom from source. But, it is still checking for update from version 1.5.3!

Zireael07:
This hasn't been implemented yet.

50Wliu:
@Zireael07 Yes it has - it's been available since Atom 1.5.0.

Zireael07:
So why is it still checking for update? I'm on 1.7.x...

50Wliu:
You have to disable the auto-checker in Help->About.

AnwarShah:
50Wliu I don't find anything like that in Help -> About

robyoder:
On Mac it's in _Atom > About Atom_

<img width="656" alt="screen shot 2016-04-26 at 1 30 42 pm" src="https://cloud.githubusercontent.com/assets/622513/14831465/2a5915c0-0bb3-11e6-84a0-8282b99aecb1.png">

liapis:
Settings > Core >  Automatically Update

/home/bduser/Pictures/atom.png

stevekm:
[...]

PR #7525, opened 2015-06-29
Auto updater

Njanderson:
image

First contribution to any open source project. Hoping that this solves the issue, which is issue #7339 about auto updating,

### DIFF
--- a/src/browser/auto-update-manager.coffee
+++ b/src/browser/auto-update-manager.coffee
@@ -1,6 +1,7 @@
 autoUpdater = null
 _ = require 'underscore-plus'
 {EventEmitter} = require 'events'
+atom = require 'atom'
 path = require 'path'
 
 IdleState = 'idle'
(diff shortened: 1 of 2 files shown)


src/browser/auto-update-manager.coffee
```diff
@@ -1,6 +1,7 @@
 autoUpdater = null
 _ = require 'underscore-plus'
 {EventEmitter} = require 'events'
+atom = require 'atom'
```
lee-dohm: I'm not sure if `atom` can be required here. The files in the `browser` directory execute in the main process, not the renderer process where most of the rest of Atom runs, including all the places where `require 'atom'` is usually used.

Njanderson:
Thanks for the feedback; I removed that requires statement. 

Love to work on anything you suggest I do to improve this PR and contribute to Atom in general.

lee-dohm:
The problem is that if you remove the `require 'atom'`, then you shouldn't be able to call `atom.config.get()` later in the same file. This is why making changes in the main process is a little sticky :smile: I'm not sure I have a good solution, which is why I called in the feedback team.

You can use [this search](https://github.com/search?utf8=%E2%9C%93&q=user%3Aatom+label%3Abeginner+is%3Aissue+is%3Aopen&type=Issues&ref=searchresults) to find anything that the triage team has labeled as "beginner". Items carrying that label are things that we think are a great starting point for people wanting to get their feet wet :sunglasses:

Njanderson:
Awesome! Thank you so much Lee! I'll take a look right now.

batjko:
For my benefit, is the `@setupAutoUpdater()` call not just checking for updates (as opposed to actually auto-updating)?

Because if that's the case, the option in the Settings View might be better named as "Auto Update Check", which sounds terrible, but you know what I'm getting at.

PR #7558, opened 2015-06-29
Reintroduce assertion API

nathansobo:
This was briefly introduced in #7350, but reverted before we went 1.0 to minimize risk. It's the logical next step in attacking our uncaught exceptions.

### DIFF
--- a/src/atom.coffee
+++ b/src/atom.coffee
@@ -198,27 +198,30 @@ class Atom extends Model
   initialize: ->
     sourceMapCache = {}
 
-    window.onerror = =>
-      @lastUncaughtError = Array::slice.call(arguments)
-      [message, url, line, column, originalError] = @lastUncaughtError
+    window.onerror = (message, url, line, column, error) =>
+      @lastUncaughtError = error
 
-      convertedLine = convertLine(url, line, column, sourceMapCache)
-      {line, column} = convertedLine if convertedLine?
-      originalError.stack = convertStackTrace(originalError.stack, sourceMapCache) if originalError
-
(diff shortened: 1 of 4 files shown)


src/atom.coffee
```diff
@@ -321,18 +324,16 @@ class Atom extends Model
 
   # Extended: Invoke the given callback whenever there is an unhandled error.
   #
-  # * `callback` {Function} to be called whenever there is an unhandled error
-  #   * `event` {Object}
-  #     * `originalError` {Object} the original error object
```
kevinsawicki: Are these fields no longer available on the object passed to the callback? Just curious why they are removed from the docs since it looks like they are still there in the emitted object.

src/atom.coffee
```diff
@@ -715,6 +716,22 @@ class Atom extends Model
   Section: Private
   ###
 
+  assert: (condition, message, metadata) ->
+    return true if condition
+
+    error = new Error("Assertion failed: " + message)
```
kevinsawicki: :art: any reason to not use string interpolation here?

src/atom.coffee
nathansobo: I would like to deprecate those fields for 2.0 since they lead to inconsistency with the assertion failure API. Just like we did pre-1.0, I think it's okay to make changes to the public API, so long as they don't actually break packages. No published package I could find is actually using this method, but I added the fields there just to be safe.

nathansobo:
@kevinsawicki scaled this back again. You cool to merge now?

kevinsawicki:
> You cool to merge now?

Looks good :+1:

Issue #7552, opened 2015-06-29
Problems overriding alt-up + alt-down in keymap.cson


arahlf:
I'm trying to override alt-up + alt-down to move a line up/down.  Using the Settings helper, I copied the bindings as a template and then added the following to my keymap.cson file:

    'atom-workspace atom-text-editor':
      'alt-up': 'editor:move-line-up'

    'atom-workspace atom-text-editor':
      'alt-down': 'editor:move-line-down'

I can now use alt-down to move a line down, but the binding for alt-up isn't getting picked up.  If I swap the order of them, so that the line-down block is before the line-up block, then line-up works but line-down does not.

Any ideas on what I'm doing wrong?

mnquintana:
See https://github.com/atom/atom/issues/7302#issuecomment-112564548

> Atom's config files (including keymap.cson) are a simple key-value format just like JSON (but CSON), which means that keys must always be unique. If you repeat keys, then only one of those values will actually be registered.

In your example above, you repeat the `'atom-workspace atom-text-editor'` key, so only one of them is registered. It should be:

```cson
'atom-workspace atom-text-editor':
  'alt-up': 'editor:move-line-up'
  'alt-down': 'editor:move-line-down'
```

arahlf:
Ahh ok that makes sense now, thanks!

Might be worth specifically calling that out or something in the comments of that file at some or providing a validation warning/error, because the phrase:

> Just as style sheets use selectors to apply styles to elements, Atom keymaps use selectors to associate keystrokes with events in specific contexts.

kinda makes it sound like you can repeat selectors (at least to someone just getting started).

PR #7562, opened 2015-06-29
Clarify that config files require unique keys

mnquintana:
From https://github.com/atom/atom/issues/7552#issuecomment-116735581

A lot of people try to use the config files as they would a stylesheet, trying to override keybindings for instance by repeating the same selector multiple times in the file, which causes issues because CSON keys must be unique. This makes it a bit more explicit that keys need to be unique.

### DIFF
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -2,7 +2,8 @@
 #
 # Atom keymaps work similarly to style sheets. Just as style sheets use
 # selectors to apply styles to elements, Atom keymaps use selectors to associate
-# keystrokes with events in specific contexts.
+# keystrokes with events in specific contexts. (Unlike style sheets however,
+# each selector can only be declared once.)
 #
 # You can create a new keybinding in this file by typing "key" and then hitting
 # tab.
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -2,8 +2,8 @@
 #
 # Atom keymaps work similarly to style sheets. Just as style sheets use
 # selectors to apply styles to elements, Atom keymaps use selectors to associate
-# keystrokes with events in specific contexts. (Unlike style sheets however,
-# each selector can only be declared once.)
+# keystrokes with events in specific contexts. Unlike style sheets however,
+# each selector can only be declared once.
 #
 # You can create a new keybinding in this file by typing "key" and then hitting
 # tab.
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -22,5 +22,6 @@
 # * https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 #
 # This file uses CoffeeScript Object Notation (CSON).
-# If you are unfamiliar with CSON, you can read more about it here:
-# https://github.com/bevry/cson#what-is-cson
+# If you are unfamiliar with CSON, you can read more about it in the 
+# Atom Flight Manual:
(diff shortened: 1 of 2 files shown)


thomasjo:
Looks good to me! :boat:

dot-atom/keymap.cson
```diff
@@ -2,7 +2,8 @@
 #
 # Atom keymaps work similarly to style sheets. Just as style sheets use
 # selectors to apply styles to elements, Atom keymaps use selectors to associate
-# keystrokes with events in specific contexts.
+# keystrokes with events in specific contexts. (Unlike style sheets however,
+# each selector can only be declared once.)
```
izuzak: I think we can drop the parens here. 

Also, it might be helpful to give an example of the error, and then how the error should be corrected (to make things perfectly clear for users who just copy/paste examples). For example:

```
# This is not correct
#
# 'atom-text-editor':
#   'enter': 'editor:newline'
# 'atom-text-editor':
#   'end': 'editor:move-to-end-of-screen-line'
#
# This is correct:
#
# 'atom-text-editor':
#   'enter': 'editor:newline'
#   'end': 'editor:move-to-end-of-screen-line'
```

Another approach would be to fix https://github.com/atom/docs/issues/43 :memo:, and then link to those docs with a big WARNING. I think I like that approach a bit more.

dot-atom/snippets.cson
```diff
@@ -13,6 +13,8 @@
 #     'prefix': 'log'
 #     'body': 'console.log $1'
 #
+# Each scope can only be declared once.
```
izuzak: Since the text above this line doesn't mention the word "scope", I'm again concerned that this won't be clear enough for some users (they won't know what a "scope" is and what the problem is exactly). For that reason, I think an example would be helpful here, similar to https://github.com/atom/atom/pull/7562/files#r33752061

izuzak:
:zap: @mnquintana for this -- helping users not make these mistakes would be great. I left two comments  -- wondering what your :thought_balloon: are.

dot-atom/keymap.cson
mnquintana: I think I prefer the latter option too – putting the examples inline here would make the comments section really long.

mnquintana:
@izuzak Just opened up https://github.com/atom/docs/pull/100 - I'll link to that new section from within these comments.

dot-atom/keymap.cson
```diff
@@ -21,5 +22,6 @@
 # * https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 #
 # This file uses CoffeeScript Object Notation (CSON).
-# If you are unfamiliar with CSON, you can read more about it here:
-# https://github.com/bevry/cson#what-is-cson
+# If you are unfamiliar with CSON, you can read more about it in the 
+# Atom Flight Manual:
+# https://atom.io/docs/latest/using-atom-basic-customization#cson
```
mnquintana: Btw @izuzak, I changed this link to the CSON Primer I added in https://github.com/atom/docs/pull/100 – do you think that's okay? There is a footnote in that section to the original link here, and it expands upon the source CSON docs quite a bit.

izuzak: > Btw @izuzak, I changed this link to the CSON Primer I added in atom/docs#100 – do you think that's okay?

Yep, looks great to me :+1:

izuzak:
:sunglasses: Thanks again for this, @mnquintana

PR #7563, opened 2015-06-29
Consider tabs or spaces mutually exclusively as indentation

### LINKED ISSUE #5642: Copy paste of text aligned with spaces (e.g. asterisk-aligned PHPDoc) sometimes loses the alignment with soft tabs off

nathansobo:
Closes #5658 
Closes #5642

A huge huge thanks to @driskell for identifying the issue and taking action to fix it. I used a slightly adapted version of his test case in this PR, but took a different approach to solving the problem.

Previously, we allowed a mix of tabs and spaces to be counted toward the indentation of a line. This caused problems when auto-indenting lines in a hard-tabbed file that contained “alignment spaces”, such as occurs in aligned comment blocks. For this case, it makes more sense to assume that the line is indented via tabs only, and consider the subsequent alignment space as part of the line’s content. This means it's actually more correct to only consider tabs **or** spaces, but not both when reading or updating a line's indentation.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2943,6 +2943,20 @@ describe "TextEditor", ->
               expect(editor.lineTextForBufferRow(7)).toBe "          c(x);"
               expect(editor.lineTextForBufferRow(8)).toBe "      current = items.shift();"
 
+            it "auto-indents lines with a mix of hard tabs and spaces without removing spaces", ->
+              editor.setSoftTabs(false)
+              expect(editor.indentationForBufferRow(5)).toBe(3)
+
+              atom.clipboard.write("/**\n\t * testing\n\t * indent\n\t **/\n", indentBasis: 1)
+              editor.setCursorBufferPosition([5, 0])
+              editor.pasteText()
+
+              # Do not lose the alignment spaces
+              expect(editor.lineTextForBufferRow(5)).toBe("\t\t\t/**")
+              expect(editor.lineTextForBufferRow(6)).toBe("\t\t\t * testing")
+              expect(editor.lineTextForBufferRow(7)).toBe("\t\t\t * indent")
+              expect(editor.lineTextForBufferRow(8)).toBe("\t\t\t **/")
+
           describe "when pasting a single line of text", ->
             it "does not auto-indent the text", ->
               atom.clipboard.write("a(x);", indentBasis: 0)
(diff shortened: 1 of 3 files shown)


driskell:
@nathansobo One reason I took space into account too is so in the event of a mixed space/tab file the indentation is still correct after copy paste.

I realise that file is badly formatted anyway. But it would now mean copying pasting from a bad file to a good file goes crazy if spaces in the middle of the original file's tabs.

Secondly accounting for spaces as well as tabs meant I could run a plugin to normalise indentation. If we don't include the spaces relying on indent level to normalise indentation isn't possible unless we only include tabs OR spaces in which case it has no effect.

Would you reconsider including space in calculations after taking account the above?

nathansobo:
@driskell I'm open to continuing to support it. One concern I had was that the original order of tabs and spaces won't be preserved when re-indenting a line. So a line that mixed tabs and spaces together would end up getting indented with all the excess spaces grouped together at the end of the line. But I guess that would preserve the indentation better than an outright breakage.

driskell:
The PR I put forward, though it was more complex, it counted the spaces and tabs the same as they would be rendered. That meant "<space><tab><space><tab>" would actually produce indentLevel of 2 because the <tab> would jump to the next tab stop.

I am beginning to second guess myself on benefit vs complexity though, but I do wonder if it would be better to handle the tab space mixture in that way - so the indentLevel is exactly the same as is rendered.

nathansobo:
@driskell Do you actually have files where tabs and spaces are mixed on the same line and where both are intended to be treated as indentation? I'm sure they exist, but I wonder how common they are.

thomasjo:
> Do you actually have files where tabs and spaces are mixed on the same line and where both are intended to be treated as indentation?

This is somewhat common in C/C++ land. I've been trying to remember concrete examples, but there are big code bases out there with this type of mixed indentation where hard tabs are used for the primary indentation followed by optional spaces used for alignment. I personally hate this style, but it's not entirely uncommon, sadly.

driskell:
@thomasjo Those are fine now after this PR. The type referred to are where indentation is tabs but spaces appear within the indentation rather than after.

I will check in more detail but I'm sure Munin project (Perl) had some of this happening. I'm sure I saw in FreeIPA too. It's mostly in older projects where editors were more... dumb... and I have some internal at my company too. Mostly where developers have (annoyingly) tabbed in either side of spaces or spaced in either side of tabs in Eclipse for example.

driskell:
To be clear, supporting them would purely be for the purpose of making it easier to fix, and reduce likely hood of it spreading. So losing the messy indentation after indent operations is actually desirable.

nathansobo:
@driskell can you take a look at my latest PR? It looks more like your original one but I wanted to stick to a single number for our indent level.

driskell:
I will move discussion to #7602 so not crossing issue boundary :+1:

PR #7593, opened 2015-06-30
Position lines without absolute coordinates

as-cii:
At CodeConf, we had a quick chat with a Chromium team member about how we could improve the rendering layer in Atom.

It turned out that Chromium is heavily optimized for non-absolutely positioned elements and, as a result, we may want to avoid using them for our lines nodes.

The main change we had to make in this PR was to ensure that lines were correctly sorted when inserting them in the DOM, so that the browser could properly figure out how to lay things out. I ended up employing a naive insertion sort algorithm which, although not super efficient, does its job and is pretty fast anyway given the small amount of lines.

Using the timeline, however, I couldn’t notice any major speedup:

| Before | After |
| --- | --- |
| screen shot 2015-06-30 at 15 31 12 | screen shot 2015-06-30 at 15 30 20 |

It is subtle, but I can kinda feel it is a bit faster, hence it is likely that the Chrome timeline is not the right tool for the job and `about:traces` could tell a different story. 

@nathansobo: what’s your take on this? Should we go for this route? My opinion is that we may just continue using the absolute coordinates mechanism, as the benefits of this approach do not seem to be extremely evident. 

Thanks! :bow:

### DIFF
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -850,7 +850,6 @@ describe "TextEditorPresenter", ->
               firstNonWhitespaceIndex: line3.firstNonWhitespaceIndex
               firstTrailingWhitespaceIndex: line3.firstTrailingWhitespaceIndex
               invisibles: line3.invisibles
-              top: 0
             }
 
             line4 = editor.tokenizedLineForScreenRow(4)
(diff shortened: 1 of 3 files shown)


nathansobo:
You can take a look at the native tracing by running `window.open('chrome://tracing')` in the dev tools console if you want to take a look at a lower level. In general I'd say we should avoid changing anything unless it brings tangible benefit, since change is always associated with risk. But maybe it's worth digging in with a sharper tool?

as-cii:
> But maybe it's worth digging in with a sharper tool?

I have tried opening a tracing window but I couldn’t get any further insight. I definitely need to better understand how that benchmarking tool works, but overall I don’t think this is going to make a difference anyway.

> In general I'd say we should avoid changing anything unless it brings tangible benefit, since change is always associated with risk.

I totally agree. Let’s close this and maybe refer to it when we’ll inspect performance with the Chromium guys.

Thanks for taking the time to have a :eyes:!

PR #7596, opened 2015-06-30
Parallelize Travis build

mnquintana:
This PR parallelizes the Travis build so that core specs and package specs can be run concurrently, significantly cutting down on build time.

| Before | After |
| --- | --- |
| [~27 min](https://travis-ci.org/atom/atom/builds/68887449) | [~22 min](https://travis-ci.org/atom/atom/builds/68996020) |

~22 minutes is still a pretty long time, so there's still some work to be done here. :wink: 

Possible next steps:
- Set up custom caching for packages on S3
  - If we manage to fix the issue below, then we might be able to use [the custom caching scheme built into Travis.](http://docs.travis-ci.com/user/caching/#Arbitrary-directories) (It's only supported for the container-based infrastructure on Linux.) Although, to do this, I think we'd have to add an option to `apm install` to only install if the package isn't already installed.
- Fix flaky specs so they can be run on Travis' (much faster and more stable) Linux infrastructure instead of the Mac infrastructure (was discussing this with @as-cii at CodeConf – this is probably the trickiest of the two)

/cc @as-cii @atom/feedback

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ git:
 env:
   global:
     - ATOM_ACCESS_TOKEN=da809a6077bb1b0aa7c5623f7b2d5f1fec2faae4
+    - NODE_VERSION=0.12
 
   matrix:
-    - NODE_VERSION=0.12
     - ATOM_SPECS_TASK=packages
     - ATOM_SPECS_TASK=core
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,18 @@ git:
 env:
   global:
     - ATOM_ACCESS_TOKEN=da809a6077bb1b0aa7c5623f7b2d5f1fec2faae4
-    - NODE_VERSION=0.12
 
-  matrix:
+  env:
+    - NODE_VERSION=0.12
     - ATOM_SPECS_TASK=packages
     - ATOM_SPECS_TASK=core
+
+  matrix:
+    exclude:
+      - os: linux
+        env: ATOM_SPECS_TASK=core
+      - os: linux
+        env: ATOM_SPECS_TASK=packages
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,11 @@ git:
 env:
   global:
     - ATOM_ACCESS_TOKEN=da809a6077bb1b0aa7c5623f7b2d5f1fec2faae4
-    
+
   matrix:
     - NODE_VERSION=0.12
+    - ATOM_SPECS_TASK=packages
+    - ATOM_SPECS_TASK=core
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ env:
     - ATOM_ACCESS_TOKEN=da809a6077bb1b0aa7c5623f7b2d5f1fec2faae4
     - NODE_VERSION=0.12
 
-  matrix:
-    - ATOM_SPECS_TASK=packages
-    - ATOM_SPECS_TASK=core
-
 matrix:
   include:
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,9 @@ matrix:
       env: ATOM_SPECS_TASK=core
     - os: osx
       env: ATOM_SPECS_TASK=packages
-  exclude:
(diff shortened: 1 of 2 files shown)


as-cii:
I am fine with merging this, although it could probably fill our Travis queue sooner because it'd spawn one more job whenever a build is triggered. My biggest worry here is to trim 5 minutes from a build only to add them to the time we wait before a build is actually started.

Since the performance gaining here may not be really huge, what do you think about trying to investigate why the Linux specs fail so that we can run tests on that platform and have caching enabled by default? If we manage to do so, I think we could cut times by 50%. At that point we may split core and package specs as well, since we would take advantage of Travis containerized machines which, AFAIK, are pretty cheap compared to the Mac ones.

I recall having a look at the Linux failures and I should have fixed many of them already in https://github.com/atom/atom/pull/6359. Things may have changed since then, but I have a pretty good feeling this is a nice route we may try. One drawback of this approach is that we wouldn’t run specs on the Mac infrastructure anymore, which is a bit weird given that the core team (and many maintainers) use it as their development environment.

Maybe we should really roll out our own package caching mechanism so that we can still use the Mac infrastructure?

mnquintana:
> I am fine with merging this, although it could probably fill our Travis queue sooner because it'd spawn one more job whenever a build is triggered. My biggest worry here is to trim 5 minutes from a build only to add them to the time we wait before a build is actually started.

Hmm that's a good point – I'm not really sure what the impact would be. However, we could always merge this, try it out for awhile, later assess if it seems to produce stalled builds more often, and if it does, revert this.

> Since the performance gaining here may not be really huge, what do you think about trying to investigate why the Linux specs fail so that we can run tests on that platform and have caching enabled by default?

I'm down to investigate this – I'll see what I can do. :smile: 

> One drawback of this approach is that we wouldn’t run specs on the Mac infrastructure anymore, which is a bit weird given that the core team (and many maintainers) use it as their development environment.

I actually think this might be a good thing – since most of us develop on Macs, OS X is a pretty well-tested platform for Atom. It seems like it would be beneficial to run specs on a platform most of us don't develop on to help catch some possible platform-specific bugs.

as-cii:
> I actually think this might be a good thing – since most of us develop on Macs, OS X is a pretty well-tested platform for Atom. It seems like it would be beneficial to run specs on a platform most of us don't develop on to help catch some possible platform-specific bugs.

Agreed: I was just thinking to the times where a build could pass locally but not on Travis. Investigating and tracking down the issue, especially if it is on Linux, could be somewhat tricky and therefore annoying. That might be ok though because, as you’re pointing out, it would basically put some pressure on us to take care of Linux as well.

> Hmm that's a good point – I'm not really sure what the impact would be. However, we could always merge this, try it out for awhile, later assess if it seems to produce stalled builds more often, and if it does, revert this.

Sounds good, let’s merge it and see how that plays when many builds are spawned. :ship:

.travis.yml
```diff
@@ -4,10 +4,17 @@ git:
...
-  matrix:
     - NODE_VERSION=0.12
 
+matrix:
+  include:
+    - os: osx
+      env: ATOM_SPECS_TASK=core
+    - os: osx
+      env: ATOM_SPECS_TASK=packages
+  exclude:
```
as-cii: Sorry for noticing this now, but why are we excluding `osx` from here? Does that spawn unnecessary builds? If yes, is there a way for making this clearer, e.g. removing the `os` section below?

mnquintana: It does spawn unnecessary builds - I should be able to just remove the `osx` entry below though.

.travis.yml
```diff
@@ -4,13 +4,17 @@ git:
...
+matrix:
+  include:
+    - os: osx
+      env: ATOM_SPECS_TASK=core
+    - os: osx
+      env: ATOM_SPECS_TASK=packages
+
 os:
   - linux
-  - osx
```
as-cii: Great! :sparkles: Do you think we can remove `- linux` from here and move it to the `matrix` section for the sake of consistency?

Then, if at some point we’ll manage to run specs on Linux as well we can simply ditch `matrix: include`, set the environment variables in the global matrix and reintroduce the `os` section.

kevinsawicki:
:+1: :ship:

PR #7602, opened 2015-06-30
Restore support for mixed tabs and spaces in line indentation

nathansobo:
In #7563, I fixed an issue where alignment spaces were getting stripped when auto-indenting lines that combined tabs and spaces by only allowing one kind of character to count as indentation. There were concerns, however, that this would make Atom's indentation handling less robust in the face of sloppy code where tabs and spaces are interleaved.

This takes an approach more similar to @driskell's original PR (#5658), but we continue to use floating point numbers for the indent level. When hard tabs are enabled, we handle the fractional portion of the indent level by appending an appropriate number of trailing spaces.

@driskell Any thoughts on this one?

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -3707,11 +3707,12 @@ describe "TextEditor", ->
       expect(editor.indentLevelForLine("\t\thello")).toBe(2)
 
     it "returns the indent level based on the character starting the line when the leading whitespace contains both spaces and tabs", ->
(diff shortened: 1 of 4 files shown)


driskell:
Yep that initially looks exactly as I would expect. So any operations involving current indent level should behave as expected even with invisibles off. I'm a floatophobe currently but if you're happy with it that's fine :smile: 

I can do some testing tomorrow with some files where this is happening and feedback.

PR #7612, opened 2015-07-01
Support subword cursor navigation

nathansobo:
Continuation of #5806 in preparation to merge.

@dsandstrom I've changed the regular expression a little bit so we don't need a hack to move the cursor to the beginning of the buffer. It also required a change to text-buffer to allow regex matches at the beginning of the scan range.

One of the tests changed slightly in behavior, but it's still seeming reasonable and things are working intuitively in all my tests.

I'm considering adding the following emacs-style bindings for those of us who are averse to the arrow keys:
- `ctrl-alt-f` Move subword forward
- `ctrl-alt-b` Move subword backward
- `ctrl-alt-d` Delete subword forward
- `ctrl-alt-h` Delete subword backward
- `ctrl-alt-shift-F` Select subword forward
- `ctrl-alt-shift-B` Select subword backward


benogle:
> ctrl-alt-f Move subword forward

Are these the defaults in any other editor? They are strange on a qwerty keyboard because the `f` key is left of the `b` key. If there are default bindings, my vote would be for arrow key chord defaults.

nathansobo:
@benogle We already have very similar character and word movement bindings (derived from Emacs). I would definitely leave the arrow bindings, but I think it makes sense to provide Emacs equivalents since we already have Emacs word and character bindings.

benogle:
> I would definitely leave the arrow bindings

Noted

dsandstrom:
Awesome, thanks.  I have not tried building this version, but does subword navigation work on mini editors?  If not, the keymaps should be under the `:not([mini])` selectors.

Btw, congrats on 1.0.

benogle:
:tada: glad to have this in

nathansobo:
>  I have not tried building this version, but does subword navigation work on mini editors? If not, the keymaps should be under the :not([mini]) selectors.

I just confirmed that the bindings and commands work in mini-editors, as I think they should.

PR #7624, opened 2015-07-01
:apple: Menus: replace three period with ellipsis

isghe:
From:
[OS X Human Interface Guidelines -Terminology and Wording](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/TerminologyWording.html#//apple_ref/doc/uid/20000957-CH15-SW3)

> IMPORTANT
> 
> Be sure to create the ellipsis character using the key combination Option-; (Option-semicolon). This ensures that an assistive app can provide the correct interpretation of the character to a disabled user. If you use three period characters to simulate an ellipsis, many assistive apps will be unable to make sense of them. Also, three period characters and an ellipsis don't look the same because the periods are spaced differently than the points of an ellipsis.

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -10,7 +10,7 @@
       { label: 'Checking for Update', enabled: false, visible: false}
       { label: 'Downloading Update', enabled: false, visible: false}
       { type: 'separator' }
-      { label: 'Preferences...', command: 'application:show-settings' }
+      { label: 'Preferences…', command: 'application:show-settings' }
       { label: 'Open Your Config', command: 'application:open-your-config' }
       { label: 'Open Your Init Script', command: 'application:open-your-init-script' }
       { label: 'Open Your Keymap', command: 'application:open-your-keymap' }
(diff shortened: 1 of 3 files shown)


isghe:
It looks https://travis-ci.org/atom/atom/jobs/69127662 fails because of a bug in tree-view-spec.coffee introduced in https://github.com/atom/tree-view/commit/d47a1841116a4ebb9f3027a10e0a9c5ce360f5ce
and solved in pull request https://github.com/atom/tree-view/pull/500

mnquintana:
Good catch! Thanks for this. I'm wondering though if we should also update the Windows and Linux menus too – I'm not sure what the conventions are for those platforms, though.

paulcbetts:
> I'm wondering though if we should also update the Windows and Linux menus too

:+1:

mnquintana:
@paulcbetts Is that the standard for Windows too then? I was just taking a look at the [Windows HIG](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742392.aspx#usingellipses) and their examples all use `...` (3 dot characters) instead of a proper ellipsis (`…`).

paulcbetts:
@mnquintana Looking at VS, it appears to use ellipses, I could be wrong though, it's a little hard to tell

kubetz:
It seems that it is a standard for Windows too:

> Indicate a command that needs additional information (including a confirmation) by adding an ellipsis at the end of the label.

Source: https://msdn.microsoft.com/en-us/library/dn742392.aspx#usingellipses

mnquintana:
@kubetz Right, that's the doc I linked to. But they don't make it explicit like Apple does that "ellipsis" means "the ellipsis character" and not 3 dots that look like an ellipsis, especially confusing because none of their examples use the ellipsis character.

kubetz:
@mnquintana Sorry for not noticing that link :blush: . 

Isn't Word autocorrecting three dots into ellipsis character? I found on web peopel complaining that it does, but don't have it installed to verify it myself. If that it the case then it can be used as an argument for switching to "…"

isghe:
That is the reason I named this branch "osx_ellipsis"… to understand what to do on others platforms It's another job ;-)

mnquintana:
@isghe We looked into it a bit further, and it would be great if you could update the menus for Windows and Linux too. :smile:

isghe:
@mnquintana I will, but… I cannot test by myself, because actually I don't have a Windows or Linux development platform.

isghe:
I did also for Windows and Linux… but take care, I cannot test those modifications by myself right now.

kubetz:
I build atom on Windows with these changes and here is the screenshot of file menu:

menu

So it looks fine.

mnquintana:
Thanks for this! :boat:

Issue #4059, opened 2014-11-04
Make package.activate() async.


bolinfest:
I think that `PackageManager` and `Package` should be resigned to be truly async and make heavy use of ES6 Promises (removing the dependency on `Q`). For example, `package.activate()` should have the option to return a `Promise`, and the package should not be considered "loaded" until that `Promise` is resolved.

Then it would be possible to depend on other packages as follows:

``` coffee
  activate: (state) ->
    # This package is not considered activated until the shared-logging package is activated.
    return atom.packages.activate 'shared-logging'
```

Then `PackageManager` and `Package` would have to be updated to use these `Promise`s all the way through. For example, `Package.activateNow()` would have to become something like:

``` coffee
  activateNow: ->
    try
      @activateConfig()
      @activateStylesheets()
      if @requireMainModule()
        maybePromise = @mainModule.activate(atom.packages.getPackageState(@name) ? {})
        if maybePromise instanceof Promise: # or maybe just check for then() method
          return maybePromise
        else
          return Promise.resolve null
    catch e
      console.warn "Failed to activate package named '#{@name}'", e.stack
```

and callers of `activateNow()` would invoke `then()` on the return value, as appropriate.

PR #7638, opened 2015-07-01
Add activationHooks + Trigger `{grammar-package-name}:grammar-used` Hook When A text-editor-element's Grammar Is Set

### LINKED ISSUE #6629: Load packages on-demand (using scopes)
I got the idea from reading through #6619 

---

The idea would be to have a way for package authors to scope their package loading and activation to a specific language scope. For instance, we don't need the `python-isort` package loading if I've never opened a python file.

joefitzgerald:
This PR:
- Adds `activationHooks` to package.json
- Triggers the `{grammar-package-name}:grammar-used` hook when a text-editor-element's grammar is set

This allows package authors to defer package activation for language-specific packages until the first use of the grammar(s) their package serves. It also leaves open the possibility to trigger other activation hooks in the future with minimal code.
- Fixes #6629

/cc @kevinsawicki

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -189,7 +189,7 @@ describe "PackageManager", ->
             expect(atom.config.get('package-with-config-defaults.numbers.two')).toBe 2
 
       describe "when the package metadata includes `activationCommands`", ->
-        [mainModule, promise, workspaceCommandListener] = []
+        [mainModule, promise, workspaceCommandListener, registration] = []
 
         beforeEach ->
           jasmine.attachToDOM(workspaceElement)
@@ -200,9 +200,13 @@ describe "PackageManager", ->
           spyOn(Package.prototype, 'requireMainModule').andCallThrough()
 
           workspaceCommandListener = jasmine.createSpy('workspaceCommandListener')
-          atom.commands.add '.workspace', 'activation-command', workspaceCommandListener
+          registration = atom.commands.add '.workspace', 'activation-command', workspaceCommandListener
 
           promise = atom.packages.activatePackage('package-with-activation-commands')
+
+        afterEach ->
+          registration?.dispose()
+          mainModule = null
 
         it "defers requiring/activating the main module until an activation event bubbles to the root view", ->
           expect(promise.isFulfilled()).not.toBeTruthy()
@@ -285,6 +289,41 @@ describe "PackageManager", ->
           runs ->
             expect(addErrorHandler.callCount).toBe 1
             expect(addErrorHandler.argsForCall[0][0].message).toContain("Failed to load the package-with-invalid-settings package settings")
+
+    describe "when the package metadata includes `activationHooks`", ->
+      [mainModule, promise] = []
+
+      beforeEach ->
+        mainModule = require './fixtures/packages/package-with-activation-hooks/index'
+        spyOn(mainModule, 'activate').andCallThrough()
+        spyOn(Package.prototype, 'requireMainModule').andCallThrough()
+
+        promise = atom.packages.activatePackage('package-with-activation-hooks')
+
+      it "defers requiring/activating the main module until an triggering of an activation hook occurs", ->
+        expect(promise.isFulfilled()).not.toBeTruthy()
+        expect(Package.prototype.requireMainModule.callCount).toBe 0
+        atom.packages.triggerActivationHook('language-fictitious:grammar-used')
+
+        waitsForPromise ->
+          promise
+
+        runs ->
+          expect(Package.prototype.requireMainModule.callCount).toBe 1
+
+      it "activates the package immediately when activationHooks is empty", ->
+        mainModule = require './fixtures/packages/package-with-empty-activation-hooks/index'
+        spyOn(mainModule, 'activate').andCallThrough()
+
+        runs ->
+          expect(Package.prototype.requireMainModule.callCount).toBe 0
+
+        waitsForPromise ->
+          atom.packages.activatePackage('package-with-empty-activation-hooks')
+
+        runs ->
+          expect(mainModule.activate.callCount).toBe 1
+          expect(Package.prototype.requireMainModule.callCount).toBe 1
 
     describe "when the package has no main module", ->
       it "does not throw an exception", ->
@@ -750,8 +789,7 @@ describe "PackageManager", ->
       package1 = atom.packages.loadPackage('package-with-main')
       package2 = atom.packages.loadPackage('package-with-index')
       package3 = atom.packages.loadPackage('package-with-activation-commands')
-      spyOn(atom.packages, 'getLoadedPackages').andReturn([package1, package2])
-
+      spyOn(atom.packages, 'getLoadedPackages').andReturn([package1, package2, package3])
       activateSpy = jasmine.createSpy('activateSpy')
       atom.packages.onDidActivateInitialPackages(activateSpy)
 
--- a/src/package-manager.coffee
+++ b/src/package-manager.coffee
@@ -30,7 +30,7 @@ module.exports =
 class PackageManager
   constructor: ({configDirPath, @devMode, safeMode, @resourcePath}) ->
     @emitter = new Emitter
-    @hooks = new Emitter
+    @activationHookEmitter = new Emitter
     @packageDirPaths = []
     unless safeMode
       if @devMode
@@ -412,11 +412,11 @@ class PackageManager
 
   triggerActivationHook: (hook) ->
     return new Error("Cannot trigger an empty activation hook") unless hook? and _.isString(hook) and hook.length > 0
-    @hooks.emit(hook)
+    @activationHookEmitter.emit(hook)
 
   onDidTriggerActivationHook: (hook, callback) ->
     return unless hook? and _.isString(hook) and hook.length > 0
-    @hooks.on(hook, callback)
+    @activationHookEmitter.on(hook, callback)
 
   # Deactivate all packages
   deactivatePackages: ->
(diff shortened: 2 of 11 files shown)


park9140:
:+1:  This is awesome.
@basarat, atom typescript needs all it's initialization behind this.

spec/text-editor-spec.coffee
```diff
@@ -4768,3 +4768,28 @@ describe "TextEditor", ->
...
+
+    afterEach: ->
+      registration?.dispose?()
+      atom.packages.deactivatePackages()
+      atom.packages.unloadPackages()
+
+    it 'dispatches the grammar-used command', ->
+      called = false
+      registration = atom.commands.add editor, 'language-coffee-script:grammar-used', (event) ->
+        console.log event
```
kubetz: Was this `console.log` committed by accident :angel:?

joefitzgerald:
Yes, console.log shouldn't be there. This is WIP, I'm moving some stuff around and will clean up history.

joefitzgerald:
Ready for review.

spec/text-editor-spec.coffee
joefitzgerald: Yes, console.log shouldn't be there. This is WIP, I'm moving some stuff around and will clean up history.

nathansobo:
@joefitzgerald I only want to use commands for communication from the user to Atom, not for communication from Atom to packages.

I would be open to a method on text editor such as `TextEditor::onDidChangeGrammar` instead of this however.

joefitzgerald:
@nathansobo this is the only way we can have activationCommands kick in on first use of a grammar. This allows us to defer activation of packages that are language-specific until the first use of the grammar for that language. `onDidChangeGrammar` doesn't achieve that objective.

thedaniel:
:-1:

I support a hard line that commands should be user-invoked. If this delayed activation thing is really, truly needed, I'd support baking this into core as activationGrammar in package.json or something.

nathansobo:
@joefitzgerald Got it. I wasn't fully understanding what you're trying to do here.

Yeah, echoing what @thedaniel is saying, I'm worried about expanding the role of commands in this way.  I see some ways forward though...

* As @thedaniel suggested, some grammar specific activation hook in packages. This is nice and focused, but may be too narrow. What happens when packages need to activate on some other event?
* Adding a new, more flexible package activation API. This would be something like `atom.packages.activate('grammar:.foo.bar')` that referenced a new top-level package property (perhaps bring back `activationEvents`?). The key difference here is we wouldn't expose an API for responding to these events. They would be *purely* for package activation (write-only).

park9140:
@nathansobo, @thedaniel, you make good points. 

Delayed activation is a simple thing to achieve for code that we control inside our activation method, by not activating any package functionality on activation but just bind a event listener for any given event we want to activate on.

This doesn't work for any dynamically activated functionality like keybinds and services however.

What if we could return a promise from our activation function that determined when to activate the other automatic features of our package.

If we add `onGrammarLoaded` to https://github.com/atom/first-mate/blob/v4.1.7/src/grammar-registry.coffee#L36?

Then delayed activation based on grammar is something like
```coffeescript
activate: ->
  return new Promise resolve -> 
    grammerLoadOff() = atom.grammars.onGrammarLoaded (grammar) ->
      if grammar == "coffeeScript"
       grammerLoadOff()
       @customInitialization();
       resolve()
```

joefitzgerald:
@park9140 you don't want to activate on load of a grammar as - I believe - all grammars are loaded on startup? Core team, is that accurate?

I do like the idea of allowing activate to return a promise. It would somewhat resolve #4059 also. I'm unsure how `activationEvents` were implemented before - how did it hook into the `emitter`s in _all the things_?

joefitzgerald:
> Q: [...] how did it hook into the emitters in all the things?

A: https://github.com/atom/atom/blob/8c2bcf39431133da43f365df46c4ddd4099f0e40/src/package.coffee#L334-L341
It didn't it registered on workspaceView. So, is that really what we'd want to go back to? Or should there be something like `atom.commands` (i.e.`atom.events`)?

joefitzgerald:
OK folks, I've pivoted to a new approach that leverages `activationHooks` in `package.json` and leaves the option open for adding other hooks in the future. Please review.

spec/package-manager-spec.coffee
```diff
@@ -286,6 +290,34 @@ describe "PackageManager", ->
             expect(addErrorHandler.callCount).toBe 1
             expect(addErrorHandler.argsForCall[0][0].message).toContain("Failed to load the package-with-invalid-settings package settings")
 
+    describe "when the package metadata includes `activationHooks`", ->
+      [mainModule, promise] = []
+
+      beforeEach ->
+        jasmine.attachToDOM(workspaceElement)
```
nathansobo: Seems like we shouldn't need the workspace element to be on the DOM here. Is this a leftover?

joefitzgerald: Yep.

spec/package-manager-spec.coffee
```diff
@@ -286,6 +290,34 @@ describe "PackageManager", ->
...
+        mainModule = require './fixtures/packages/package-with-activation-hooks/index'
+        spyOn(mainModule, 'activate').andCallThrough()
+        spyOn(Package.prototype, 'requireMainModule').andCallThrough()
+
+        promise = atom.packages.activatePackage('package-with-activation-hooks')
+
+      it "defers requiring/activating the main module until an triggering of an activation hook occurs", ->
```
nathansobo: What do you think about making some mock assertions here? Did you want to use the `requireMainModule` mock?

src/package-manager.coffee
```diff
@@ -30,6 +30,7 @@ module.exports =
 class PackageManager
   constructor: ({configDirPath, @devMode, safeMode, @resourcePath}) ->
     @emitter = new Emitter
+    @hooks = new Emitter
```
nathansobo: What do you think about naming this more pedantically? `activationHookEmitter`

src/package-manager.coffee
```diff
@@ -405,6 +406,14 @@ class PackageManager
     else
       Q.reject(new Error("Failed to load package '#{name}'"))
 
+  triggerActivationHook: (hook) ->
+    return unless hook? and _.isString(hook) and hook.length > 0
```
nathansobo: Appreciate the caution here, but I wonder if we should blow up clearly rather than failing silently...

src/package.coffee
```diff
@@ -122,7 +123,7 @@ class Package
         @loadMenus()
         @loadStylesheets()
         @settingsPromise = @loadSettings()
-        @requireMainModule() unless @hasActivationCommands()
+        @requireMainModule() unless @activationShouldBeDeferred()
```
nathansobo: :+1:

src/package-manager.coffee
joefitzgerald: How would you propose blowing up? Would you go as far as an `Error`?

src/text-editor-element.coffee
```diff
@@ -153,8 +153,10 @@ class TextEditorElement extends HTMLElement
     @component.focused() if event.relatedTarget is this
 
   addGrammarScopeAttribute: ->
```
nathansobo: I really think the activation hook should be triggered at the model level. Is there a reason you're confining it to the view? There might be a situation where an editor is hidden and we could delay package activation, but that does seem pretty rare.

nathansobo:
Everything seems pretty straightforward except where the grammar hook is being triggered and how we're going to format the string. I just want to put some thought into our scheme.

I think I see where you're going with using the package-name prefix in the hook name. However, there's also the grammar scope name. And a single language package could contain multiple grammars. So we're losing a bit of information here. That said, this does lay down a nice pattern for packages to add their own hooks when we expose the activation hook API.

src/text-editor-element.coffee
joefitzgerald: Well I started in text-editor. And then moved to text-editor-element. I think you and @kevinsawicki should chat. I can put it wherever it makes sense.

src/package-manager.coffee
joefitzgerald: 6 of 1, half dozen of the other. Your call. Either way, if people want to abuse it they will abuse it however it is named (if avoiding abuse is your intent for naming it more pedantically).

joefitzgerald:
> I think I see where you're going with using the package-name prefix in the hook name. However, there's also the grammar scope name, and a single language package could contain multiple grammars.

I could just emit once for the package name and then again for the grammar scope name?

src/text-editor-element.coffee
nathansobo: The only reason I can see it needing to be in the view would be to support the previous command-based approach, since commands are tied to the view.

nathansobo:
> I could just emit once for the package name and then again for the grammar scope name?

I guess let's just start with the package name and see where it goes.

src/package-manager.coffee
nathansobo: Yeah I would. If you trigger a hook with an invalid string you should know about it.

joefitzgerald:
@nathansobo I moved the triggering of activationHooks to tokenized-buffer. Please review.

nathansobo:
Looking good. Running a build on our internal CI and giving core team members one last chance to check this out. If all goes well, will merge this afternoon.

thedaniel:
Don't forget to doc it over in atom/docs after it lands!

nathansobo:
@joefitzgerald Would you mind opening a PR documenting this?

nathansobo:
Thanks by the way. I really appreciate the thought you're putting into leveling up our package system.

PR #7698, opened 2015-07-03
WIP: Use default cursor instead of pointer

mnquintana:
Refs https://github.com/atom/settings-view/issues/531

This implements something @simurai and I have been throwing around for awhile now – it removes all Atom core usage of `cursor: pointer` in favor of `cursor: default`. As @jerone put it in https://github.com/atom/settings-view/issues/531:

> Showing the cursor as a "hand" when hovering over things in the UI is a very "web" oriented convention.

This will go a long way, I think, toward making Atom feel less like a """website""" running in a semi-headless Chrome, and far more like a native app with a custom UI. :smile: 

_Note: This won't affect the styling of any components that explicitly declare another `cursor` type._

> But how will I know I can click on a thing? How will I know if a thing can be interacted with?

Here's how various OSes and desktop shells handle this:
- **OS X**: Occasionally a hover state, often a thing that looks like a button, but usually no way of knowing other than by clicking on it.
- **Windows**: Hover states.
- **GNOME 3**: Hover states.
- **Ubuntu Unity**: Hover states.

I think the safest bet here is to give actionable elements appropriate hover states as a replacement for `cursor: pointer`.

Although this is quite a sweeping change, it doesn't _entirely_ do away with `cursor: pointer` – the pointer should still be used when an element actually is an external link (whether it's an `a` element or not). Ideally this will be combined with https://github.com/atom/status-bar/issues/89 eventually.

TODO:
- [x] What do we do about actionable text? Right now it'll have a standard link-like underline hover state, but without `cursor: pointer`.
- [x] ~~We could be more extreme and add an `!important` flag to really discourage use of `cursor: pointer` – thoughts?~~
- [ ] Ensure that no other Bootstrap classes need overrides
- [ ] Add `.link-external` class example to styleguide
- [ ] Add `.link-internal` class example to styleguide
- [ ] Add `.text-selectable` class example to styleguide
- [x] Remove all remaining uses of `cursor: pointer` from settings-view (https://github.com/atom/settings-view/pull/575)
- [ ] Ensure that all actionable status bar tiles have hover states
  - [ ] Update status bar tile hover states to look less like links (background highlight maybe?)
- [ ] Communicate this change well so package authors can adapt if need be
  - [ ] Also encourage package authors to use `cursor: default` + hover state for actionable elements in their packages over `cursor: pointer`

/cc @atom/feedback

### DIFF
--- a/static/links.less
+++ b/static/links.less
@@ -7,8 +7,12 @@ a {
     text-decoration: underline;
   }
 
-  &.link-external,
   &[href^="http"] {
     cursor: pointer;
   }
+}
+
+// For non anchor elements that open external links
+.link-external {
+  cursor: pointer;
 }
--- a/static/links.less
+++ b/static/links.less
@@ -7,7 +7,8 @@ a {
     text-decoration: underline;
   }
 
-  &[href^="http"], .link-external {
+  & .link-external,
+  &[href^="http"] {
     cursor: pointer;
   }
 }
--- a/static/links.less
+++ b/static/links.less
@@ -7,7 +7,8 @@ a {
     text-decoration: underline;
   }
 
-  &[href^="http"] {
+  &[href^="http"],
+  & * {
     cursor: pointer;
   }
 }
--- a/static/links.less
+++ b/static/links.less
@@ -1,10 +1,13 @@
 @import "ui-variables";
 
 a {
-  cursor: pointer;
   color: @text-color-highlight;
   &:hover {
     color: @text-color-highlight;
     text-decoration: underline;
+  }
+
+  &[href^="http"], .link-external {
+    cursor: pointer;
   }
 }
--- a/static/links.less
+++ b/static/links.less
@@ -7,7 +7,7 @@ a {
     text-decoration: underline;
   }
 
-  & .link-external,
+  &.link-external,
   &[href^="http"] {
     cursor: pointer;
   }
--- a/static/links.less
+++ b/static/links.less
@@ -13,7 +13,38 @@ a {
   }
 }
 
-// For non anchor elements that open external links
+// For any elements that open external links
 .link-external {
   cursor: pointer;
+}
+
+// For any textual elements that are not styled like buttons but open something
+// "internally" in Atom
+.link-internal {
+  cursor: default !important;
+  text-decoration: none;
+  color: @text-color-info;
+  // transition: padding 0.5s;
+
+  &:active,
+  &:hover,
+  &:focus {
+    border-radius: 2px;
+    text-decoration: none;
+    // padding: 1px;
+  }
+
+  &:hover,
+  &:focus,
+  &:active {
+    color: contrast(@text-color-info, darken(@text-color-info, 26%), lighten(@text-color-info, 26%), 43%);
+  }
+
+  &:hover, &:focus {
+    background-color: @text-color-info;
+  }
+
+  &:active {
+    background-color: lighten(@text-color-info, 4%);
+  }
 }
--- a/static/links.less
+++ b/static/links.less
@@ -13,9 +13,9 @@ a {
   }
 }
 
-// For any elements that open external links
+// For any non-button elements that open external links
 .link-external {
-  cursor: pointer;
+  cursor: pointer !important;
 }
 
 // For any textual elements that are not styled like buttons but open something
@@ -24,14 +24,12 @@ a {
   cursor: default !important;
   text-decoration: none;
   color: @text-color-info;
-  // transition: padding 0.5s;
 
   &:active,
   &:hover,
   &:focus {
     border-radius: 2px;
     text-decoration: none;
-    // padding: 1px;
   }
 
   &:hover,
--- a/static/links.less
+++ b/static/links.less
@@ -39,10 +39,10 @@ a {
   }
 
   &:hover, &:focus {
-    background-color: @text-color-info;
+    background-color: @background-color-selected;
   }
 
   &:active {
-    background-color: lighten(@text-color-info, 4%);
+    background-color: lighten(@background-color-selected, 4%);
   }
 }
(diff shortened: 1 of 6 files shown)


simurai:
> What do we do about actionable text? Right now it'll have a standard link-like underline hover state, but without `cursor: pointer`.

Keep the underline for now until we have another option? Yeah, the "background highlights" for the status-bar could be an option. It would be another "type" of button, where it looks just like text but has a background on hover. I guess it depends a bit if the text has to left/right align to something since there is an invisible padding.

> We could be more extreme and add an !important flag to really discourage use of cursor: pointer – thoughts?

Maybe too extreme? Having the default cursor should already be enough of a hint.

static/links.less
```diff
@@ -1,10 +1,13 @@
...
 
 a {
-  cursor: pointer;
   color: @text-color-highlight;
   &:hover {
     color: @text-color-highlight;
     text-decoration: underline;
   }
+
+  &[href^="http"], .link-external {
```
simurai: I think the `.link-external` needs an `&` in front.

Also that class gets lost a bit being the second selector, but arguably it's the more important one. We don't really have any naming conventions, but how about:

```less
&.link-external,
&[href^="http"] {
  cursor: pointer;
}
```

raulsntos:
Since you're trying to get a more desktop feel, you may want to add:
```css
*{
  -webkit-user-select: none;
}
```
Since text in desktop apps is not selectable if it's not in an editable area like an input or something.

static/links.less
mnquintana: Whoops, not sure how I missed that :laughing: – fixed and reordered the selector.

batjko:
@raulsntos That's not actually true, I don't think. 
I tend to select a whole lot of text on panels and fields that are not editable at all.

mnquintana:
@raulsntos @batjko We've talked about doing that too, but that'll probably be in a separate PR. What we'd do is disable `webkit-user-select` on `*`, and then selectively enable it wherever we (or package authors) want the text to be user selectable.

batjko:
@mnquintana That kind of opt-in approach would indeed make sense for the sake of a more Desktop-like user experience. I can see that point.

What I wonder is how difficult it may be to avoid breaking people's expectations and assumptions. The status bar for example is practically made of actionable elements now, and my impression is that the majority of packages with text content on views may use it to output important information / data.

When text is data (rather than just labels for example), you can be assured that people will want to select and copy it, and they will be discouraged to even try if the cursor doesn't have the expected hover state for it.

I don't have any statistics to back that up, though. It's just my impression and I may be wrong about how many views with texts are like that. 
It's just something to keep in mind.

simurai:
I would be pro `* { -webkit-user-select: none; }` if it already came from the start, but now with lots of packages out there, it might cause package authors trouble.

Since `-webkit-user-select` cascades we could also do it on a per package basis where we have more control. Like `.settings-view { -webkit-user-select: none; }` and then opt-in where needed. So other people's packages are still unaffected.

static/links.less
```diff
@@ -1,10 +1,19 @@
...
 
 a {
-  cursor: pointer;
   color: @text-color-highlight;
   &:hover {
     color: @text-color-highlight;
     text-decoration: underline;
   }
+
+  &[href^="http"],
```
mnquintana: @simurai Do you have an idea why grunt-lesslint is failing on this line? This is valid Less, so I'm kinda confused here :confused:

simurai: Maybe an issue with grunt-lesslint? [linter-less](https://atom.io/packages/linter-less) seems fine.

mnquintana: Found it - [grunt-lesslint](https://github.com/jgable/grunt-lesslint) actually uses [csslint](https://github.com/CSSLint/csslint) for linting once the Less is compiled, and csslint wasn't configured properly for the lesslint task (it was complaining about regex-like selectors being slow).

@kevinsawicki Is there any reason why we have both grunt-csslint and grunt-lesslint set up? As far as I can tell they seem a bit redundant, since grunt-lesslint is basically just csslint + sourcemaps to source Less files.

mnquintana:
@simurai Before I document it in the styleguide, what do you think about `.link-internal`, `.link-external`, and `.text-selectable`?

simurai:
Hmm.. What about a shorter `.link` instead of `.link-internal`? Everything would just be a link and only if it's a variation, like external, the name gets extended?

Quickly tried it out  how that would feel?

links

Shouldn't have the internal link some padding on the sides? But then it almost feels like a "button". Here they call them [flat buttons](http://www.getmdl.io/components/index.html#buttons-section). Also heard "silent button", or "link button". I guess the problem with adding a padding on the sides is that they don't look visually aligned when on the left or right of something.

Also, should we enforce that no internal links are allowed to have underlines? Or is that still up to package or theme authors?

simurai:
Ok, after thinking about it some more.. how do you like this:

links

- `.btn-silent` would be part of `buttons.less`. A good use case might in the status bar, where normal buttons are too strong and they are not really links to navigate, but more for triggering modals and tooltips. So using a button like control would make sense.
- `.link` this is a generic (internal) link. Used to navigate, e.g. the breadcrumbs in the package settings.
- `.link-external` used for opening a browser outside of Atom. The visual difference to an internal link is that it's blue and uses a pointer cursor.

thedaniel:
Hi, friends! I am going through the open PRs on settings-view right now, and https://github.com/atom/settings-view/pull/575 is blocked on this. @mnquintana @simurai is this something you are still pursuing?

mnquintana:
@thedaniel Yup! Sorry I've let this sit for so long (things have been crazy here), but hopefully I'll have time to take a look at this again this weekend!

lee-dohm:
@mnquintana Is this something you want to finish? Let us know :grinning:

simurai:
It might be worth to split it up into smaller PRs instead of trying to catch everything at once.

Or maybe first introduce the new UI components like the `.btn-silent`, `.link`, `.link-external` and then once that's merged, start to use it everywhere.

damieng:
Has been open with no activity for over a year, closing.

PR #7699, opened 2015-07-03
Different title for file dialogs and folder dialogs.

### LINKED ISSUE #6796: Change dialog titles for "open project folder" and "open file"
With there being multiple now and being able to be opened in the same window it would be nice if the dialog titles reflected such so that mistakes in keystrokes can be quickly corrected before they become a problem (in that you browse to a place and can't open a file.)

wbrbr:
Fixes #6796
If the dialog opens a file its title is changed to "Open File", and if it opens a directory its title is changed to "Open Folder".

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -576,7 +576,7 @@ class AtomApplication
       title: switch type
         when 'file' then 'Open File'
         when 'folder' then 'Open Folder'
-        when 'all' then 'Open'
+        else 'Open'
 
     if process.platform is 'linux'
       if projectPath = @lastFocusedWindow?.projectPath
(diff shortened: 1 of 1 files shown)


src/browser/atom-application.coffee
```diff
@@ -573,7 +573,10 @@ class AtomApplication
 
     openOptions =
       properties: properties.concat(['multiSelections', 'createDirectory'])
-      title: 'Open'
+      title: switch type
+        when 'file' then 'Open File'
+        when 'folder' then 'Open Folder'
+        when 'all' then 'Open'
```
batjko: Should this last one perhaps be an `else` instead?

kevinsawicki:
Looks good, thanks!

envygeeks:
Thanks!

PR #7702, opened 2015-07-04
:art: Create console intro.

### LINKED ISSUE #1490: Add welcome text to the dev tools console.
When you open a dev tools console from [MDN](https://developer.mozilla.org) you see this: 

element getboundingclientrect - web api interfaces mdn

Would a few lines of docs (and ascii art, of course) be helpful for users? Or do you think it would just be annoying?

Sticksman:
Logs ascii art from a file and into the dev console.
Couldn't find an icon that actually made sense.
fix #1490

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -32,7 +32,7 @@ window.onload = function() {
 
     fs.readFile(path.join(loadSettings.resourcePath, 'static', 'intro_ascii_art'), function(err, data) {
       if (err) {
-        throw(err);
+        throw err;
       }
       console.log(data.toString());
     });
--- a/static/index.js
+++ b/static/index.js
@@ -30,6 +30,13 @@ window.onload = function() {
       setupDeprecatedPackages();
     }
 
+    fs.readFile(path.join(loadSettings.resourcePath, 'static', 'intro_ascii_art'), function(err, data) {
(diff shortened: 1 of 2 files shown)


benogle:
Could we make the mark wider and take less lines? Would be cool if it were square and smaller. Also not a super fan of the type.

benogle:
Some of [these](http://patorjk.com/software/taag) are cool. like http://patorjk.com/software/taag/#p=display&f=Swamp%20Land&t=ATOM

simurai:
How does it look in the console? On GitHub lines probably have a bigger line height, making it look stretched.

> take less lines

I agree that making it smaller would be nice. Just so that you don't have to scroll down a lot everytime you wanna see the console output.

mnquintana:
> I agree that making it smaller would be nice. Just so that you don't have to scroll down a lot everytime you wanna see the console output.

Putting it all in one line seems like a good way to do this (essentially an ASCII art version of the [Atom logo](https://cloud.githubusercontent.com/assets/72919/2874231/3af1db48-d3dd-11e3-98dc-6066f8bc766f.png)).

benogle:
Closing this as I dont think we want something so large.

PR #7712, opened 2015-07-05
:bug: Ask user to 'save as' if save fails when closing tab, fixes #7708

### LINKED ISSUE #7708: Unsaved changes lost if permission denied when saving on exit
If I open a file that I do not have permission to save and try to save it I get a popup telling me I lack permissions, e.g. `Unable to save file: permission denied '/etc/hosts'`. This seems reasonable (although I wish it would ask me for my sudo password instead).
However, if I modify it and then close it, Atom will ask me if I want to save it, and when I click "yes", it closes the file and gives me that same popup that it failed to save. This means my work is lost. I don't think I have used an editor before that doesn't keep the file open if it fails to save and I believe this should be considered a bug, right?

I am on Chakra linux running 1.0.0 via http://chakraos.org/ccr/packages.php?ID=7252

einarmagnus:
This probably needs testing on mac where users are prompted for credentials when saving files they don't have permissions for.

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -671,6 +671,104 @@ describe "Pane", ->
       expect(item1.save).not.toHaveBeenCalled()
       expect(pane.isDestroyed()).toBe false
 
+    it "does not destroy the pane if save fails and user clicks cancel", ->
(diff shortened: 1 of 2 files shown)


einarmagnus:
This fails the list test, I'll fix it and push a new one

PR #7733, opened 2015-07-06
Remove leftover update statements

as-cii:
This PR gets rid of a couple of `update` statements in `TextEditorPresenter` that were getting executed when some event was triggered. They are unnecessary because they’ll get executed anyway as soon as `getState` is called.

Although this is pretty straightforward, another pair of :eyes: is most welcome. Thanks! :bow: 


maxbrunsfeld:
:+1: Looks good to me.

PR #7748, opened 2015-07-07
Emit will-insert-text event when pasting

maxbrunsfeld:
Packages like `go-to-line` [use the `::onWillInsertText` event](https://github.com/atom/go-to-line/blob/70428c943c4e1c5c2675d90e418bdcf2fd62833b/lib/go-to-line-view.coffee#L24) to stop invalid text from being inserted into a text-editor. Previously, this event was not emitted when pasting. This allowed invalid data to enter the editor in the `go-to-line` dialog.

Fixes atom/go-to-line#22

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -772,31 +772,24 @@ class TextEditor extends Model
   # Returns a {Range} when the text has been inserted
   # Returns a {Boolean} false when the text has not been inserted
   insertText: (text, options={}) ->
-    willInsert = true
-    cancel = -> willInsert = false
-    willInsertEvent = {cancel, text}
-    @emit('will-insert-text', willInsertEvent) if includeDeprecatedAPIs
-    @emitter.emit 'will-insert-text', willInsertEvent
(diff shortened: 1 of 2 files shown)


src/text-editor.coffee
```diff
@@ -2977,6 +2972,14 @@ class TextEditor extends Model
 
   logScreenLines: (start, end) -> @displayBuffer.logLines(start, end)
 
+  emitWillInsertTextEvent: (text) ->
+    result = true
+    cancel = -> result = false
+    willInsertEvent = {cancel, text}
+    @emit('will-insert-text', willInsertEvent) if includeDeprecatedAPIs
```
mnquintana: Do we still need `includedDeprecatedAPIs` post 1.0? Is it just because we haven't removed core's dependency on `emissary` yet?

maxbrunsfeld: I think we can remove a lot of these; core doesn't depend on emissary anymore. I'm not sure when exactly we want to remove them.

PR #7752, opened 2015-07-07
:arrow_up: apm@1.0.0

Colean128:
To just upgrade the as-odd-tiles apm


as-cii:
Hi @spongecattle, thanks for your PR. I have updated that branch against master, so that it can be tried again. I am going to close this PR for now, but it’d be great to know if you tried it out and if it worked for you.

Thank you! :bow:

spongecattle:
OK, but thanks anyway @as-cii

PR #7754, opened 2015-07-07
`updateContentDimensions` when model changes

as-cii:
In #7733 we have removed a couple of update statements that were no longer needed. Unfortunately, we didn’t catch a :bug: that caused the editor not to scroll automatically when the model changed.

This PR fixes it and also adds a spec to prevent this to happen in the future. The :beetle: was caused by the following scenario:
1. A new line was inserted at the end of the buffer, causing the content height to be bigger;
2. The model triggered a `setScrollTop` on `TextEditorPresenter` with the correct value;
3. `TextEditorPresenter::setScrollTop` didn’t know about the change in the model yet and tried to constrain `@scrollTop` using a bad value (i.e. the content height before the change) as the upper bound;
4. :bug: 

/cc: @maxbrunsfeld @nathansobo


nathansobo:
Thanks. Glad you got this in before release!

PR #7717, opened 2015-07-05
Check if error message exists before using it in Pane.handleSaveError

izuzak:
This should fix https://github.com/atom/atom/issues/7714, though I'm not really sure when an error won't have a message defined (other than manually setting it to undefined, which a package could do I guess?). 

cc @kevinsawicki for :thought_balloon: :eyes:

PR #7768, opened 2015-07-07
Guard against non-string messages in Pane.handleSaveError

izuzak:
See https://github.com/atom/atom/pull/7717#issuecomment-119280666 for context. 

cc @kevinsawicki :zap: for catching that case as well


kevinsawicki:
:+1:

PR #7769, opened 2015-07-07
Validate atom.notifications input

kevinsawicki:
Previously the API documented on `atom.notifications` was not validated so you could create a notification with a non-string message, non-object options, or non-string detail. 

This would cause the notifications package to throw errors when it tried to convert a non-string markdown message.

Now it verifies the options on creation and throws error when they are invalid.

Closes https://github.com/atom/notifications/issues/79

/cc @benogle


benogle:
:+1:

kevinsawicki:
I backed out the `options.detail` validation since it seemed like from failing specs in the notifications package that existing packages were using non-string details and I didn't want to break them.

benogle:
I think allowing for non-string details is fine.

PR #7809, opened 2015-07-09
Move view registration for Pane & Gutter related views earlier

thedaniel:
Registering them in pane-container led to packages deserializing before e.g. Gutter was available.


nathansobo:
:green_heart:

PR #7825, opened 2015-07-10
:BUG:About menu language

chinakids:
`...`It turned into a `…`  ，It will result in failure of language packs, although we have to repair, but I believe this is a bug~

This change in all *.cson,I hope you can confirm that the problem is an update or bug

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -10,7 +10,7 @@
       { label: 'Checking for Update', enabled: false, visible: false}
       { label: 'Downloading Update', enabled: false, visible: false}
       { type: 'separator' }
-      { label: 'Preferences…', command: 'application:show-settings' }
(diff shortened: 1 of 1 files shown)


batjko:
Does the fact that you closed the PR mean that you found this change to be unnecessary?

mnquintana:
This was an intentional change made in https://github.com/atom/atom/pull/7624 and is not a bug. [Apple's own Human Interface Guidelines](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/TerminologyWording.html#//apple_ref/doc/uid/20000957-CH15-DontLinkElementID_20) make it clear that the ellipsis character `…` is preferred over three periods `...` for accessibility reasons, so I am skeptical that this would break international support.

chinakids:
yes，I look at your's commit,and I believe that the change is justified.

PR #7858, opened 2015-07-12
Fix indentation when pasting lines

svanharmelen:
The fix to ignore invisibles (64dfda572dd3db46ff841db9659f413a5f9b3b21) introduces a new bug when pasting lines from the clipboard:

before

As the commit takes the content of `@buffer.lineFromRow(bufferRow)` to test against the `decreaseIndentRegex`, it will actually test the content of the row the cursor is on instead of the content that is being pasted. And this (of course) could cause unexpected indentations...

In turn the same goes for checking if the `bufferRow` is a commented line, so this check is removed again.

After this commit pasting above a line matching the `decreaseIndentRegex` works as expected again:

after

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2262,7 +2262,7 @@ describe "TextEditor", ->
             expect(editor.indentationForBufferRow(2)).toBe 0
             expect(editor.indentationForBufferRow(3)).toBe 1
 
-      describe "when a newline is appended on a line that matches the decreaseNextIndentRegex", ->
+      describe "when a newline is appended on a line that matches the decreaseNextIndentPattern", ->
         it "indents the new line to the correct level when editor.autoIndent is true", ->
           waitsForPromise ->
             atom.packages.activatePackage('language-go')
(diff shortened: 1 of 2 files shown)


svanharmelen:
@maxbrunsfeld think you should be able to judge this one fairly easy :wink:

Did have a look at some other approaches, but they all ended up being (a lot) more expensive and then this pretty straight forward solution...

src/language-mode.coffee
```diff
@@ -234,14 +234,15 @@ class LanguageMode
...
-    @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, tokenizedLine, options)
+    @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, line, tokenizedLine, options)
 
   suggestedIndentForLineAtBufferRow: (bufferRow, line, options) ->
     tokenizedLine = @editor.displayBuffer.tokenizedBuffer.buildTokenizedLineForRowWithText(bufferRow, line)
-    @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, tokenizedLine, options)
+    @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, line, tokenizedLine, options)
 
-  suggestedIndentForTokenizedLineAtBufferRow: (bufferRow, tokenizedLine, options) ->
+  suggestedIndentForTokenizedLineAtBufferRow: (bufferRow, line, tokenizedLine, options={}) ->
```
maxbrunsfeld: In this case, I'd rather leave the default argument off of the `options` parameter and instead null-guard it on line 257, just because it's nice to avoid the extra object allocation.

spec/text-editor-spec.coffee
```diff
@@ -3012,7 +3012,19 @@ describe "TextEditor", ->
...
+              atom.clipboard.write("a(x);\n  b(x);\n    c(x);\n", indentBasis: 0)
+              editor.setCursorBufferPosition([7, 0])
+              editor.pasteText()
+
+              expect(editor.lineTextForBufferRow(7)).toBe "      a(x);"
+              expect(editor.lineTextForBufferRow(8)).toBe "        b(x);"
+              expect(editor.lineTextForBufferRow(9)).toBe "          c(x);"
+              expect(editor.lineTextForBufferRow(10)).toBe "    }"
+
+
```
maxbrunsfeld: Could you remove this second blank line?

maxbrunsfeld:
:+1: This is a great, simple solution. Thanks so much for catching this. Left some very minor comments above.

spec/text-editor-spec.coffee
svanharmelen: Whoeps, missed that one... Removed in latest commit.

src/language-mode.coffee
svanharmelen: Check... Adjusted...

svanharmelen:
@maxbrunsfeld thanks for the review, both comments are addressed in the latest commit.

svanharmelen:
@maxbrunsfeld have a minute to review the updates I did after your earlier review? Think we are good to merged this one right?

svanharmelen:
Thx!

PR #7869, opened 2015-07-13
Fix config.set scope selector example

izuzak:
It seems that the `config.get` and `config.set` APIs are not fully compatible. The `set` API takes a scope selector which must be prefixed with a `.`, while the `get` API takes a scope descriptor which doesn't need to have a `.` (it's first converted into a ScopeDescriptor and then then that object prepends a `.` if it's missing, as far as I can tell). Would be nice if they were compatible in this way, but for now -- I think fixing the example would be :sparkles:.

cc @maxbrunsfeld for :eyes: since Git blame says you were polishing these examples in the past.

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -561,7 +561,7 @@ class Config
   # atom.config.get('editor.tabLength', scope: ['source.js']) # => 4
   #
   # # Set ruby to 2
-  # atom.config.set('editor.tabLength', 2, scopeSelector: 'source.ruby') # => true
+  # atom.config.set('editor.tabLength', 2, scopeSelector: '.source.ruby') # => true
   #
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks for fixing that; looks good to me.

I'm not sure if we still need to automatically prepend the dot in `ScopeDescriptor`. It seems like something that was there for backwards-compatibility. I think @benogle or @nathansobo would know.

nathansobo:
This was a bit of an oversight in our API modernization program. Since we didn't emit a deprecation warning from these call sites when the `.` wasn't present, it might be risky to remove it because there could be code relying on it. I would ideally like to not mess with the selector.

Issue #4966, opened 2015-01-09
Have metrics disabled by default, or completely removed


greencopper:
You need to leave the metrics disabled by default and then at least ask users, first time they activate the editor, whether they would allow it to be enabled.

I for one will not allow spyware to run on my box, hence I disable or remove such applications, but Atom doesn't provide a clue in the application, so if someone haven't seen the FAQ, they don't know.

This is a problem because Atom is then collecting data WITHOUT consent, even if it is in the FAQ and even if the user has downloaded the application.

Besides, I find the whole concept of tracking users with the excuse of collecting useful data ridiculous!

In the entire history of Open Source this has never been needed and users are extremely helpful in sharing information, data, and reports about bugs, how they use the application, and other stuff.

The metrics package in Atom should to be removed from the default install, and then you can leave the package for anyone who want to install it by himself from the installation menu. Or as suggested, leave it OFF at first.

Another important point is that Atom is NOT simply collecting statistical data! If that where the case there certainly is no need to create a SHA1 of each installation MAC address. That is NOT collecting statistical data, that is specific and individual surveillance, which is unacceptable.

50Wliu:
I guess putting a notice in `Welcome.md` might help, or do what lots of applications do on first startup: ask for permission (I'm guessing this is possible with Squirrel?).

About your point with 'this has never been needed':  There are currently [68 open issues](https://github.com/atom/atom/labels/more-information-needed) with the `more-information-needed` tag (and that's not counting issues where it was removed after the needed info was added).  I'd guess at least half of the issues don't include the OS or Atom version, and half of the issues coming from Notifications don't include reproduction steps.

greencopper:
It should atleast be as "opt in", people need to make the choice when using the application the first time, and collecting SHA1 of MACs are unacceptable too! These can be reversed using rainbow tables.

Atom can generate a unique uuid on first time usage and use that instead!

And data, are they send encrypted? If not, they should be.

There's a whole bunch of problems with the approach of the metrics package.

aoighost:
This needs to be opt-in, not opt-out. This is a massive privacy issue otherwise.

lee-dohm:
Also relevant:

https://github.com/atom/metrics/issues/16#issuecomment-36307675

ferdy-galema:
+1

Pretty sneaky "feature" that is well hidden in the Metrics package. Just respect your users and disable the package by default.

Skalman:
[...]

PR #7871, opened 2015-07-13
metrics removed

kantorkel:
Metrics. A package that reports usage information to Google Analytics. A hackable text editor for the 21st century should be free as in freedom.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,6 @@
     "keybinding-resolver": "0.33.0",
     "link": "0.30.0",
     "markdown-preview": "0.150.0",
-    "metrics": "0.51.0",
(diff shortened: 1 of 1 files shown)


Zireael07:
+1.

kubetz:
Do you expect this to get merged or it is just some kind of attempt to pressure atom team from a different angle?

AFAIK atom developers are agains removing the package (see https://github.com/atom/atom/issues/4966) so creating PRs instead of continuing the discussion in the issue dedicated to it seems kinda strange.

lee-dohm:
No matter what the strategy decision is around opt-in, removing the package from the default installation is not the solution. Thanks for your interest.

PR #7878, opened 2015-07-13
Added docs for `PackageManager::activatePackage` and `PackageManager::deactivatePackage`

rameshvarun:
:memo: I noticed that the `PackageManager::activatePackage` function is used frequently in specs for packages (ex: [1](https://github.com/lee-dohm/indentation-indicator/blob/master/spec/indentation-indicator-spec.coffee#L12) [2](https://github.com/atom/tree-view/blob/17fb96308437ac63d2cb16da9d787cd20740a418/spec/tree-view-spec.coffee#L30) [3](https://github.com/atom/find-and-replace/blob/8e357b79e590ddc8b952aa78f77dbb3200a38967/spec/select-next-spec.coffee#L18)), however is undocumented in the [API docs](https://atom.io/docs/api/v1.0.2/PackageManager).

This PR adds a new section `Activating and deactivating packages` to the `PackageManager` docs, with documentation for the `activatePackage` and `deactivatePackage` functions.

### DIFF
--- a/src/package-manager.coffee
+++ b/src/package-manager.coffee
@@ -373,6 +373,10 @@ class PackageManager
     else
       throw new Error("No loaded package for name '#{name}'")
 
+  ###
+  Section: Activating and deactivating packages
+  ###
+
   # Activate all the packages that should be activated.
   activate: ->
     promises = []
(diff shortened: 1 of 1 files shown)


benogle:
@nathansobo we comfortable making these public?

nathansobo:
I feel weird about the API of `activatePackage`. I'm wondering if we should change the implementation to always activate the package, then use another method or an option internally to do the deferred version.

rameshvarun:
@nathansobo This seems to be a common point that confuses people when writing specs - see https://discuss.atom.io/t/cant-activate-package-in-specs/13672 and https://discuss.atom.io/t/can-you-force-the-activation-of-another-package/10885.

Right now there is `Package#activateNow` which I think immediately activates the package - https://github.com/atom/atom/blob/master/src/package.coffee#L156.

benogle:
Have we decided to leave these private for now?

nathansobo:
Yes, I think we should fix the API before making it public. @kuychaco maybe this is something you could look into and we could discuss.

kuychaco:
We're considering evolving our API so I'm going to close this issue for now. Thanks @rameshvarun for your thought and effort!

PR #7879, opened 2015-07-13
Add docblocks for `Notification` public functions

ssorallen:
- Mark `dismiss` as an "extended" API because its use case is uncommon.
- Mark event handler functions as public because responding to a
  notification being displayed or dismissed is useful.

### DIFF
--- a/src/notification.coffee
+++ b/src/notification.coffee
@@ -19,18 +19,40 @@ class Notification
     unless _.isObject(@options) and not _.isArray(@options)
       throw new Error("Notification must be created with an options object: #{@options}")
 
+  ###
+  Section: Event Subscription
+  ###
+
+  # Public: Invoke the given callback when the notification is dismissed.
+  #
+  # * `callback` {Function} to be called when the notification is dismissed.
+  #
+  # Returns a {Disposable} on which `.dispose()` can be called to unsubscribe.
   onDidDismiss: (callback) ->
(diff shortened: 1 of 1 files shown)


src/notification.coffee
```diff
@@ -19,18 +19,40 @@ class Notification
...
 
   getOptions: -> @options
 
+  ###
+  Section: Methods
+  ###
+
   # Public: Retrieves the {String} type.
+  #
+  # Returns a {String}.
```
benogle: These can be on the same line:

```
# Public: Returns the {String} type.
```

src/notification.coffee
```diff
@@ -19,18 +19,40 @@ class Notification
...
   getType: -> @type
 
   # Public: Retrieves the {String} message.
+  #
+  # Returns a {String}.
```
benogle: Similar to the other method above

src/notification.coffee
ssorallen: :+1: I didn't know that worked, thanks. Does it look for the "Returns" keyword?

benogle: yeah it does

izuzak:
@ssorallen Thanks for making those updates @benogle suggested. 

@benogle I think this looks good. Want to give it another look?

benogle:
:+1:

PR #7880, opened 2015-07-13
Update windows.md

ghost:


benogle:
Thanks!

netmml:
@benogle :+1:

docs/build-instructions/windows.md
```diff
@@ -54,7 +54,7 @@ If none of this works, do install Github for Windows and use its Git shell. Make
 ### Common Errors
 * `node is not recognized`
 
-  * If you just installed node you need to restart your computer before node is
+  * If you just installed node, you'll need to restart your computer before node is
   available on your Path.
```
batjko: Not necessarily true, actually. Just the command window needs to be closed and re-opened, at least that's how it was for me on Win7, Win8.1 and Win10.

The most common reason for this error is likely to be that they tried to re-use the command window which had not been started with the updated PATH variable.

PR #7881, opened 2015-07-13
:arrow_up: babel-core@5.6.17

bolinfest:
Upgrade Babel to pick up bug fixes and to ensure a version that
has support for transforms such as `es7.classProperties`.

This transform is particularly useful for classes that subclass `React.Component`
as shown in http://babeljs.io/blog/2015/06/07/react-on-es6-plus/#property-initializers.


thomasjo:
This looks good to me! Thanks @bolinfest :bow:

PR #7905, opened 2015-07-15
:apple: Add more subword navigation keybindings for Mac OS X

jimmydo:
These `alt`-less variants are commonly available in Mac OS X apps --
such as Sublime Text 3, TextMate 2, and Xcode.

Related pull request to change a conflicting `ctrl-backspace` keybinding:
https://github.com/atom/bracket-matcher/pull/154


thedaniel:
I don't have a good idea how many people depend on the bracket-matcher binding, but ctrl-shift-(left,right) are also currently bound to pane:move-item-(left,right) in core, and ctrl-arrows are used by the OS to switch workspaces by default.

jimmydo:
Ah I missed the existing bindings for pane:move-item-(left,right). My personal observation is that the bindings are more commonly used for subword navigation, so I would favor that instead. :) Moving the current tab left and right within the current pane doesn't seem like it would be as common.

As for ctrl-arrows, I usually use 3-finger swipes on my trackpad, which is another default mechanism for moving between workspaces, so I forgot that I had disabled the corresponding keybindings in the OS X Keyboard preferences pane. I'm thinking most Atom users would probably be okay with disabling the workspace keybindings if they wanted to use ctrl-arrows for subword navigation?

benogle:
> My personal observation is that the bindings are more commonly used for subword navigation, so I would favor that instead. :) Moving the current tab left and right within the current pane doesn't seem like it would be as common.

> I'm thinking most Atom users would probably be okay with disabling the workspace keybindings if they wanted to use ctrl-arrows for subword navigation?

If we're anecdata-ing, I use both of these things 100 times a day. 

I think we can come up with better keybindings for this that dont clash with other bindings.

PR #7916, opened 2015-07-15
Support descriptions for enum values in config.

oggy:
The "enum" property may now specify options as a list of [value,
description] pairs.

This goes with atom/settings-view#589.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1572,6 +1572,14 @@ describe "Config", ->
                 items:
                   type: 'string'
                   enum: ['one', 'two', 'three']
+              str_options:
+                type: 'string'
+                default: 'one'
+                enum: [
+                  value: 'one', description: 'One'
+                  'two',
+                  value: 'three', description: 'Three'
+                ]
 
           atom.config.setSchema('foo.bar', schema)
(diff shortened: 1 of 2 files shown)


benogle:
I like this idea cc @atom/feedback

thomasjo:
Same here! This seems like a valuable addition.

benogle:
Though my vote would be for a an object rather than an array: `{value: 'one', description: 'One'}`

oggy:
@benogle I think I went for an array rather than an object so you have control over the order. Granted, perhaps supporting an object (and just sorting the descriptions alphabetically in that case) would be good too.

oggy:
@benogle Oh, I think I misread you -- you meant an array of objects. I'm on board with that.

thomasjo:
@oggy But wouldn't it be an array of objects? You can still control the order.

thomasjo:
LGTM! :ship:

benogle:
cc @maxbrunsfeld

oggy:
Is there anything else needed here? I think the requires-changes tag can be removed.

50Wliu:
I've also restarted the build.

Issue #7428, opened 2015-06-26
cmd+n should default to cmd+shift+n when no windows are open on OS X


void-validation:
Atom: 1.0
Mac OS X: 10.10.3

On OS X, if you close all windows so only the menubar is active, cmd+n does nothing at all. It should default to cmd+shift+n if no windows are open in Atom.app.

Apple mention suggestions for default keys here: http://support.apple.com/kb/HT1343

dominikwilkowski:
:+1:

cesarandreu:
Similarly, when you no longer have open tabs, I would expect pressing Command-W to close the front window and focus the next window in the stack unless the window is minimized. I'm on OS X 10.10.4 and Atom 1.0.

This is how Sublime Text 3 behaves. Google Chrome also behaves similarly, except that they do it when you have a single open tab. Pressing Command-W when you have a single open tab will close the window and shift focus to the next open window unless it's minimized.

Should I create a separate issue for this or piggyback on this one?

lee-dohm:
@cesarandreu it's generally best to keep one item per Issue. This way there is no confusion as to when to close it.

jacekkopecky:
To get the described behaviour, what works for me is to put the following in `.atom/keymap.cson`:
```
'body':
  'cmd-n': 'application:new-window'
'atom-workspace':
  'cmd-n': 'application:new-file'
```
Except now the File menu shows `cmd-N` for both commands, but what the heck, nobody ever looks.

joshaber:
Fixed by https://github.com/atom/atom/pull/10233.

PR #7933, opened 2015-07-16
Open new window on new file without open windows

### LINKED ISSUE #7171: New file command does nothing with no windows open
I'm running Atom 0.207 on Mac. I ran into a weird scenario today where hitting cmd-n and clicking "File -> New File" did absolutely nothing. Unfortunately, I'm not sure exactly what caused this. I was adding and renaming folders and maybe using Git Plus a little bit but that's all. Anyways, here's a gif of the behavior: http://g.recordit.co/njk9NwI6M8.gif.

fb55:
Fixes #7171

OSX allows applications to be in focus without an active window, which leads to some unexpected behavior.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -167,7 +167,7 @@ class AtomApplication
     @on 'application:run-benchmarks', -> @runBenchmarks()
     @on 'application:quit', -> app.quit()
     @on 'application:new-window', -> @openPath(_.extend(windowDimensions: @focusedWindow()?.getDimensions(), getLoadSettings()))
-    @on 'application:new-file', -> (@focusedWindow() ? this).openPath()
+    @on 'application:new-file', -> @openPath()
     @on 'application:open', -> @promptForPathToOpen('all', getLoadSettings())
     @on 'application:open-file', -> @promptForPathToOpen('file', getLoadSettings())
     @on 'application:open-folder', -> @promptForPathToOpen('folder', getLoadSettings())


benogle:
Seems like we would like to keep the same behavior when there is a focused window, but fix when there is no focused window. Is that what happens here?

cc @maxbrunsfeld

fb55:
> Seems like we would like to keep the same behavior when there is a focused window, but fix when there is no focused window. Is that what happens here?

The current behavior is kept when there is a focused window and the semantics for opening a new window without a focused window are applied otherwise.

pydolan:
FYI: this is probably related to #7428, so update accordingly if it addresses.

fb55:
That is the same issue.

>

src/browser/atom-application.coffee
```diff
@@ -167,7 +167,7 @@ class AtomApplication
     @on 'application:run-benchmarks', -> @runBenchmarks()
     @on 'application:quit', -> app.quit()
     @on 'application:new-window', -> @openPath(_.extend(windowDimensions: @focusedWindow()?.getDimensions(), getLoadSettings()))
-    @on 'application:new-file', -> (@focusedWindow() ? this).openPath()
+    @on 'application:new-file', -> @openPath()
```
DouweM: I don't understand how this would fix anything. If there is no focused window, both evaluate to exactly the same thing, so how does this fix anything if that's exactly the case we want to change the behavior of?

maxbrunsfeld: I have the same question. Assuming that the `focusedWindow()` works properly, this should not change the behavior in the case where there isn't a focused window.

fb55: Late response: On OSX, it is possible that there is no focused window (eg. when all windows are closed), but the application might still be active and receive keyboard events. Native applications still allow users to open new files via keyboard shortcuts, which isn't supported by Atom.

If there is no focused window, then `focusedWindow` returns undefined and the command is ignored. Removing this check seemed like a quick fix. Having a second look, the `newWindow` option of `openPath` should instead probably be set to `Boolean(@focusedWindow())`.

Any thoughts?

maxbrunsfeld: > If there is no focused window, then focusedWindow returns undefined and the command is ignored. Removing this check seemed like a quick fix.

I don't think your fix will have the desired effect. With the original code, if `@focusedWindow()` returned `undefined`, then `this.openPath()` would be called. Your change makes it so that `this.openPath()` is *always* called, even if there *is* a focused window.

joshaber:
https://github.com/atom/atom/pull/10233 fixed that bug :boom:

PR #7944, opened 2015-07-17
Bump one-dark/light-syntax@1.1.0

thomasjo:
Bumps both `one-dark-syntax` and `one-light-syntax` to v1.1.0 which adds highlighting for the `.support.type` scope. 

See https://github.com/atom/one-dark-syntax/issues/36 for details.


thomasjo:
Build is :green_heart: so I'm :ship: this.

PR #7945, opened 2015-07-17
Character Measurements (iframe)

as-cii:
This PR is intended to supersede #6083. Please, note that it’s still a :construction_worker: _work in progress_ :construction_worker: and, as a result, there are still some minor hiccups and a few performance regressions, which I am going to expound later in this description.
#### Background

Have a look at #6055 to get a sense of why we want to do this. @nathansobo did a great job at explaining the drawbacks of the current approach and how we could possibly solve them, so I strongly recommend to read that issue in depth.
#### Solution: `<iframe>`

The idea with the approach proposed in this PR is to have an independent context where we can arbitrarily build a bunch of DOM nodes exclusively for measuring purposes. We employed an `<iframe>` for this, so that a reflow in the measurement context does not affect the main document and viceversa.

To accomplish this we have introduced `LinesYardstick` (naming is hard :sweat_smile:), an abstraction whose purpose is to hide the complexity of dealing with the measurement context housekeeping. The API is simple:
- `buildDomNodesForScreenRows`
- `leftPixelPositionForScreenPosition`

We need to explicitly have two separate steps, as otherwise we’d end up continously changing the DOM and reading from it, thus causing extra reflows.
#### The Measurement Context

The context, as stated previously, is an `<iframe>` that replicates some of the functionalities featured on the fully fledged document, namely:
- Has a bunch of line nodes
- Includes a custom `<style>` element which contains font-size information
- Includes `<atom-styles>` which, in turn, contains the syntax theme

You can think of it as a skinny replica of the larger document, where line characters occupy the exact same space and are, therefore, measurable. It’s extremely important for this context to be independent, as we will use it in `TextEditorPresenter` as well as in `DisplayBuffer` (in the future).
#### Performance

**Side note**: this section is a collection of personal thoughts and may not be accurate or necessarily _correct_. However, it will hopefully serve as a starting point for further optimizations and discussions.

We have encountered several performance regressions along the way, especially for performance sensitive scenarios such as scrolling. Keep in mind that for such _intensive_ activities we need to respect the 16ms budget.

As of today, the steps we perform are the following:
1. Right before updating (i.e. at the beginning of `TextEditorPresenter#getState`), we build the DOM nodes for a bunch of screen rows we know we are going to measure:
   - All the rows that will be rendered on screen after the update (to make sure decorations, cursors, etc. will be able to measure stuff).
   - The longest row (to calculate the maximum width of the open document).
   - The last cursor row (to calculate the hidden input position).
2. As we update things (`@updateDecorations`, `@updateCursors`), we call `@pixelPositionForScreenPosition` which, in turn, will query `LinesYardstick` for positions. The first time this method is called, a `Recalculate Style` and a `Layout` will be triggered.

These are the two main operations and you can see how they perform on a quite heavy scrolling session (profiler disabled to avoid performance overhead):

screen shot 2015-07-17 at 11 59 41

And, if we zoom in one of those spikes:

screen shot 2015-07-17 at 12 01 21

The first half of the picture is the illustration of the aforementioned steps:
1. Building HTML: `1.33ms`
2. Parsing HTML: `1.28ms`
3. Recalculate Style: `0.839ms`
4. Layout: `1.408ms`

The whole frame took `20ms`; measuring the first half via Chrome (Shift+Click) reported `6ms`.

Now the question is: what could we improve? There are a couple of possibilities:
1. Build the same HTML for the lines in the context as well as for the lines in the real document. This way we can avoid reparsing the lines twice, although cloning the nodes still takes some time. Not sure if this is really useful.
2. Build skinnier lines for the measurement context. Ideally, we should avoid as much nesting as possible and, reducing the scopes complexity by building only those ones needed for styling, could improve that. It's also true that it would complicate things a bit (involves dealing with CSS rules, etc.). (eb8a1c0)
3. Be smarter about updates. Instead of including every possible line, we build and layout only those lines needed for position-based updates. The drawback here is that, when e.g. we have decoration for each visible row, this optimization is useless. This means I want to make sure to speedup the current process instead of relying on not having certain rows. Anyway, this improvement has a positive impact on `Layout`, `Recalculate Style` and HTML manipulation. To get a sense of what this could look like, here's a relevant snippet:
   
   ``` coffee
   getMeasurableScreenRows: ->
     measurableRows = new Set
   
     @addOverlaysRowsToSet(measurableRows)
     @addCursorsRowsToSet(measurableRows)
     @addDecorationsRowsToSet(measurableRows)
   
     if lastCursorRange = @model.getLastCursor()?.getScreenRange()
       measurableRows.add(lastCursorRange.start.row)
       measurableRows.add(lastCursorRange.end.row)
   
     if longestScreenRow = @model.getLongestScreenRow()
       measurableRows.add(longestScreenRow)
   
     measurableRows
   ```
4. Cut non-syntax CSS from styles. Removing `<atom-styles>` does not yield huge speedups, so I am not sure it's the first thing we want to improve. Still an improvement I’d like to make, though. (2257688)
5. Reduce layout times by trying to modify the measurement context without affecting the whole document. If we manage to always change the DOM in a way that affects only smaller layout boundaries, we may be able to halve reflow times. (28405e09afe097a8f40f90cb2166a47619c8feef)

I believe a combination of 2), 3), 4) and 5) could be interesting. HTML creation (and parsing) and `Layout` are the expensive operations right now and I am thinking to ways in which we could speed them up. I am still kinda making some research but I feel we could reach decent performance if we put some more effort on this. I'll investigate other potential speedups in the meantime. @nathansobo: what do you think?
#### Conclusion

If you have read everything so far, I extremely appreciate your attention. I feel like there are many ways to solve this problem, none of which is guaranteed to be the right one, thus I wanted to include as much of my knowledge as possible so that you and the future me can try things out based on it. :pray:

Any feedback would be much appreciated! :bow: Thanks!


src/lines-yardstick.coffee
```diff
@@ -0,0 +1,164 @@
...
+    context.style.height = "600px"
+    context.style.display = "block"
+    context
+
+  setupIframe: =>
+    @initialized = true
+    @domNode = @iframe.contentDocument.body
+    @headNode = @iframe.contentDocument.head
+    @headNode.appendChild(@stylesNode)
+    @headNode.appendChild(@syntaxStyleElement)
```
as-cii: Here I am adding the whole `<style>` element, but with a little effort we may generate a minimal CSS document that contains only those selectors with font informations.

Similarly, we can reduce scope HTML complexity by dropping those scopes that don’t have a matching CSS selector in the previously generated CSS.

src/lines-yardstick.coffee
```diff
@@ -0,0 +1,164 @@
...
+    if foundToken = @findTokenByColumn(position.column, tokens)
+      textNode = foundToken.childNodes[0]
+      positionWithinToken = position.column - parseInt(foundToken.dataset.start)
+    else
+      textNode = last(tokens).childNodes[0]
+      positionWithinToken = textNode.textContent.length
+
+    @leftPixelPositionForCharInTextNode(textNode, positionWithinToken)
+
+  findTokenByColumn: (column, tokens, startIndex = 0, endIndex = tokens.length) ->
```
as-cii: To avoid finding a token through `TokenIterator` (which, by its nature, is sequential) I decorated the HTML tokens with some data attributes that contain screen position information. This allows us to perform a binary search on all the tokens in the current line.

The idea here is that we trade a few microseconds more on HTML parsing for a faster retrieval on the measurement side: while the former should be negligible, the latter brings a boost of `~ 2ms`.

alexandernst:
Is there a bug/feature request in Chrom(e|imum)'s bugzilla for chars measurement API?

as-cii:
I have changed the original description of this PR, linking the proposed solutions to real commits which implement them in this branch. This is how they compare to **master**.

|                            |                                                                  Timeline                                                                 |
|:--------------------------:|:-----------------------------------------------------------------------------------------------------------------------------------------:|
|         **master**         | screen shot 2015-07-21 at 12 05 17 |
| **as-iframe-measurements** | screen shot 2015-07-21 at 12 04 24 |

Admittedly we have a performance degradation, but I’d say that the editor speed overall is pretty good, as we go past the 60fps limit quite rarely.

#### Another possible improvement

Looking at the graph this morning I realized something. Below you can find a smaller portion of the above timeline:

screen shot 2015-07-21 at 12 04 24

The spikes in the graph correspond exactly to the generation of new tiles. As you notice, the frames next to such spikes are super cheap (`~5 ms`), which means we potentially have some time (`~11ms`) to perform other tasks. This pattern repeats all over the graph, as you can see from the larger picture above.

So, now that we have noticed this interesting correlation, the question is: what if we generated some HTML ahead of time in these cheaper frames? If we’re scrolling down there’s a relatively high chance that the user will scroll down again in the next frame, and the same goes for scrolling up. 

Having something like `10ms` of *spare time* means that we can even up the frames in the graph, effectively decreasing the possibility of going past 60fps.

#### Ending Notes

Even without this last improvement, I feel pretty good about the current performance. There may be something I am missing, so if anyone wants to chime in and give their feedback it would be super welcome! :bow:

as-cii:
> Is there a bug/feature request in Chrom(e|imum)'s bugzilla for chars measurement API?

@alexandernst: not that I know, but I might be outdated on that. However, there’s an API being discussed [over here](https://wiki.css-houdini.org/explaining-css-layout) and we should definitely reach out to the Chromium team as soon as we figure out how a possible API could look like.

Personally, I haven’t think too much about it, but I feel pretty good about the proposal in the link above.

spec/text-editor-component-spec.coffee
```diff
@@ -77,6 +82,37 @@ describe "TextEditorComponent", ->
 
       expect(component.lineNodeForScreenRow(0).textContent).not.toBe("You shouldn't see this update.")
 
+  describe "measurements", ->
+    it "is equivalent to TextEditorPresenter::pixelPositionForScreenPosition", ->
```
nathansobo: I like this comparison test for now, but will we be able to drop the old method before merging this and test the new method directly?

spec/text-editor-component-spec.coffee
```diff
@@ -371,7 +407,7 @@ describe "TextEditorComponent", ->
       it "displays newlines as their own token outside of the other tokens' scopeDescriptor", ->
         editor.setText "var\n"
         nextAnimationFrame()
-        expect(component.lineNodeForScreenRow(0).innerHTML).toBe "<span class=\"source js\"><span class=\"storage modifier js\">var</span></span><span class=\"invisible-character\">#{invisibles.eol}</span>"
+        expect(component.lineNodeForScreenRow(0).innerHTML).toBe "<span class=\"source js\"><span class=\"storage modifier js\"><span data-start=\"0\" data-end=\"3\" class=\"token\">var</span></span></span><span class=\"invisible-character\">E</span>"
```
nathansobo: I'm thinking we should avoid annotating the start/end positions on lines outside of the measurement context, since it adds complexity that isn't needed in the actual display lines. You could parameterize the line construction routine to only include it in certain cases.

spec/text-editor-component-spec.coffee
```diff
@@ -888,7 +924,7 @@ describe "TextEditorComponent", ->
       cursor = componentNode.querySelector('.cursor')
       cursorRect = cursor.getBoundingClientRect()
 
-      cursorLocationTextNode = component.lineNodeForScreenRow(0).querySelector('.storage.type.function.js').firstChild
+      cursorLocationTextNode = component.lineNodeForScreenRow(0).querySelector('.storage.type.function.js .token').firstChild
```
nathansobo: Would be great to not need to touch these tests.

benogle:
cc @esprehn At CodeConf we talked about getting examples to you regarding text measurement. There is a discussion at https://github.com/atom/atom/issues/6055, and another approach using canvas at https://github.com/atom/atom/pull/6083. If you have ideas on what we should/shouldnt be doing wrt char measurement or if/how we can help out with a measurement API, please let us know.

nathansobo:
Just got a stack trace using this...

```
TypeError: Cannot read property 'textContent' of undefined
  at LinesYardstick.module.exports.LinesYardstick.leftPixelPositionForCharInTextNode (/Applications/Atom.app/Contents/Resources/app.asar/src/lines-yardstick.js:257:52)
  at LinesYardstick.module.exports.LinesYardstick.leftPixelPositionForScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/lines-yardstick.js:226:19)
  at TextEditorComponent.module.exports.TextEditorComponent.pixelPositionForScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:1076:34)
  at TextEditorPresenter.module.exports.TextEditorPresenter.pixelPositionForScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-presenter.js:1367:33)
  at TextEditorPresenter.module.exports.TextEditorPresenter.pixelRectForScreenRange (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-presenter.js:1391:22)
  at TextEditorPresenter.module.exports.TextEditorPresenter.updateHiddenInputState (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-presenter.js:420:20)
  at TextEditorPresenter.module.exports.TextEditorPresenter.getState (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-presenter.js:152:14)
  at TextEditorComponent.module.exports.TextEditorComponent.updateSync (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:197:38)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:313:28
  at ViewRegistry.module.exports.ViewRegistry.performDocumentUpdate (/Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:168:9)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:3:61
```

Issue #4738, opened 2014-12-22
Pane split also splits file


halleycarleton:
When splitting a pane you end up with the same file open in two panes. While splitting a file is a useful feature it should be separate from splitting panes. It is frustrating having to close the extra file view every time I split a pane.

I should be able to split panes without ending up with duplicate views of the same file, but I should also be able to split a file into multiple views if I want.

tylerFowler:
+1 

I find this pretty annoying. I always end up with a floating tab that I don't need (and usually forget about it until it creates too much clutter).

acontreras89:
:+1:

austincrft:
:+1:

phaazon:
:+1:

elliotlarson:
:+1: this would be a nice addition, I agree with @tylerFowler

acontreras89:
I'm working on this right now :)

arxpoetica:
Progress?

T-Grave:
:+1: 

I always close the tab after splitting it, so this would be an awesome feature!

ramirezd42:
Any progress on this? It's very annoying to have to close a file every time i open a pane

justinko:
:pray:

acontreras89:
This will be coming shortly. Sorry for the delay.
You can keep an eye on PR atom/pull/7953 for updates.

50Wliu:
#7953 has been merged and the improvements should be available in Atom 1.6.0.  Thanks @acontreras89!

acontreras89:
My pleasure, @50Wliu! Glad to see my PR finally merged :blush:

Trudko:
@acontreras89 thank you , you wonderful human being :D. It has bothered me for some time and I decide today , that I will look  for solution, so I am happy to see that 1.6 will include this.

whomwah:
@acontreras89 Great stuff on this one. Thank you!

kehd07:
Hello! This is already close, but I am also getting this issue on 1.22.1
What am I missing? Thanks!

acontreras89:
@kehd07 this is not a bug but a design decision. Others commands are available which implement different behaviors.

[This comment on the discuss forum](https://discuss.atom.io/t/how-to-replace-a-keymap-binding/16834/3) explains how to override a default keybinding.

I guess what you want is to add something like this to your `keymap.cson`:

```
'body':
  'ctrl-k down': 'pane:split-down-and-move-active-item'
[...]

PR #7953, opened 2015-07-18
Improve multi-pane API

acontreras89:
**Updated** as of Feb 2016 (see discussion below):
Improve multi-pane API to allow creating empty panes, as well as moving or copying files among them. 
Addresses #4738.
- [x] Add an option (namely `moveActiveItem`) to `Pane::split`, allowing to move the active item when splitting, instead of copying it
- [x] Rewrite `pane:split-<direction>` commands to create empty panes
- [x] Add `pane:split-<direction>-and-<action>-active-item` commands, to achieve original behaviour of the split commands and, alternatively, move the current item instead of copying it
- [x] Add `window:<action>-active-item-to-pane-<direction'>` to allow moving and copying files between panes

Where
- `<direction>` is either `up`, `down`, `right` or `left`
- `<action>` is either `copy` or `move`
- `<direction'>` is either `above`, `below`, `on-right` or `on-left`

Behavior for current key bindings is kept. For example, <kbd>cmd-k left</kbd> (<kbd>ctrl-k left</kbd>) will now trigger `pane:split-left-and-copy-active-item` instead of `pane:split-left`.


acontreras89:
I forgot a question mark :sweat_smile: 
Sorry for that!

benogle:
How would this behave when there is only one editor in a pane, and you split? Does it keep the initial pane around?

acontreras89:
@benogle thanks for looking into this! :smile: 
If `Destroy Empty Panes` setting is enabled, it will _flash_. Otherwise the pane will stay open with no items.
Should it not do anything at all when both options are used?

mehcode:
@acontreras89 I want to jump in and say that I don't think this addresses #4378 all the way. That issue is asking for the separation of **split pane** and **open file in a new pane**.

I also would like the ability to split the viewport without having the active file buffer go *anywhere*.

There are a few actions one could want (that I see).

 1. Pane: Split `direction` — open a new, empty pane in `direction`
 2. File: Split `direction` — open a new pane with the same buffer in `direction`
 3. File: Move `direction` — move file in `direction` making a new pane only if at the end (this PR)

In closing -- I don't want a config option to toggle between `2` and `3`. I want to be able to execute `2` or `3` depending on my needs at the time.

benogle:
I am with @mehcode. Seems like we should have three commands: 

* `pane:split-right` - creates a new empty pane to the right
* `pane:split-right-copy-item` - copies current item into new pane
* `pane:split-right-move-item` - moves current item into new pane

Then the current keybinding could bind to `pane:split-right-copy-item`. 

Though thinking that maybe `*-copy-item` and `*-move-item` only create a pane when there is nothing already in that direction. So if you have a left and right pane, you're focused in the left pane and run `split-right-move-item` it will _not_ create a new pane, but move to the existing pane.

What do you guys think?

acontreras89:
@benogle I really like the idea of moving files between panes if possible, creating new panes otherwise, although I'd bind the current keybindings to the `*-move-item` commands. I guess I'll have to override it :sweat_smile:

Regarding the `pane:split-right`, how should that work if `Destroy Empty Panes` config option is enabled?

benogle:
> I'd bind the current keybindings to the *-move-item commands

I just think we need to keep the current behavior for now

> Regarding the pane:split-right, how should that work if Destroy Empty Panes config option is enabled?

Maybe it just does nothing? Or shows a warning in the console or something? Not sure yet.

mehcode:
> Though thinking that maybe *-copy-item and *-move-item only create a pane when there is nothing already in that direction. So if you have a left and right pane, you're focused in the left pane and run split-right-move-item it will not create a new pane, but move to the existing pane.

Agreed.

For commands; how about:

 - `pane:split-right` - creates a new empty pane to the right
 - `pane:move-item-right` - copies current item into a pane to the right (new pane only if needed)
 - `pane:copy-item-right` - moves current item into a pane to the right (new pane only if needed)

I'm not sure I like the word `split` in there for the second two.

> Regarding the pane:split-right, how should that work if Destroy Empty Panes config option is enabled?

I'd honestly vote to transition this to activate only on item close. In other words.. without the option enabled, closing the last buffer in the pane does not close the pane; however, with the option enabled,  closing the last buffer in the pane does close the pane. In either case, <kbd>Ctrl-W</kbd> on an active pane with no buffers would close it. It'd technically be backwards-compatible as splitting a pane without touching the item is a new thing.

----

One more thing I thought of that is slightly related to this. It'd be nice to be able to move the focus between active panes. Something like `pane:focus-right`.

acontreras89:
> I'm not sure I like the word `split` in there for the second two.

@mehcode I agree but `pane:move-item-*` commands already exists and serve a different purpose (they're intended to move the file within the same pane). Any other ideas?

> transition this to activate only on item close

:+1: I thought of this yesterday and it really sounds like the way to go to me

> One more thing I thought of that is slightly related to this. It'd be nice to be able to move the focus between active panes. Something like `pane:focus-right`.

Do you mean like `window:focus-pane-*` commands? :stuck_out_tongue:

mehcode:
> Do you mean like window:focus-pane-* commands? :stuck_out_tongue: 

Exactly. Didn't see those. Haha.

----

> I agree but pane:move-item-* commands already exists and serve a different purpose (they're intended to move the file within the same pane). Any other ideas?

Perhaps something along the lines of `window:move-item-right` and `window:copy-item-right` as we're operating between panes and not within a pane.

benogle:
Ok, how about `pane:move-item-to-right-pane` and `pane:copy-item-to-right-pane`.

> I'd honestly vote to transition this to activate only on item close.

This seems reasonable. It lets you open an empty pane, but will remove it when an item in that pane is destroyed.

So just to recap: these are the things we've agreed on

* `pane:split-<direction>` - open a new empty pane in that direction. When `Destroy Empty Panes` is true, it will still open an empty pane. 
* `pane:copy-item-to-<direction>-pane` - Copies to pane in that direction. Creates a new pane if none in that direction. Make this the default keybinding so things dont change on people.
* `pane:move-item-to-<direction>-pane` - Moves to pane in that direction. Creates a new pane if none in that direction.

This is basically what https://github.com/vidius/atom-pain-split does. 

I'm admittedly hesitating a little on the move and copy commands not creating panes in some situations. Thinking maybe they should be atomic, but intuitively, I want it to just move or copy to the existing pane when it already exists in that direction. The question is, will anyone have a left and right pane, then want to create a 'middle' pane with the move or copy commands? If this is common, then these commands should always create panes. If not, I feel like what we've outlined is good.

mehcode:
Agreed on command names.

---

> The question is, will anyone have a left and right pane, then want to create a 'middle' pane with the move or copy commands? If this is common, then these commands should always create panes. If not, I feel like what we've outlined is good.

I feel that one could easily make a 4th command that runs `pane:split-<direction>` followed by `pane:copy-item-to-<direction>-pane` if that is their use case (or this 4th command could be added to core). I'm fairly biased here because I'm envisioning this magical world of dynamic panes where I can now move files in arbitrary directions to an existing/new pane and as I close files the panes go away; and, it sounds awesome. Think of like the "Dynamic Workspaces" feature of gnome shell.

benogle:
Ok, I'm in for non-atomic commands. We can try them out and see how it goes.

acontreras89:
I updated the title and description of the pull request and will start working on the changes right away. Thanks for the help, @mehcode @benogle!

Just one last thing. Should we be worried about packages using the current `pane:split-*` commands?

benogle:
Are there packages using those commands? We should definitely be careful of renames, we need to assess the impact.

acontreras89:
Added to the check list. I'll look into that too, @benogle

acontreras89:
I've added the new commands and rewritten the old ones. Also updated keybindings to keep current default behavior.

I've been looking into making these commands create new panes only when necessary. I've found a couple of options and hope to have something ready soon :)

nathansobo:
Just checking in here... looks like there's still some boxes unchecked.

antcodd:
Could we please include something like #8228? The current behaviour would be significantly more useful if the scroll position was also copied and is consistent with other editors and the idea of 'copying'.

Note that the cursor position already seems to be copied. The option to not activate the pane might also be useful.

acontreras89:
First of all, apologies for the delay.

After 6859b16cadc5c1b554bcba78b0f385669b16d664, something didn't quite feel right. I was using the `nearestPaneInDirection` method in `workspace`'s view' to check whether an existing pane was available to use.

This solution worked pretty nicely but readability was compromised and specs were hard to design, as actual rendered positions are used to find the closest existing pane, if any.

After discussing this matter a bit further with @benogle on slack, it felt like changing the commands' namespace to `window` was the way to go, and so I've done. Commands have been renamed to 
* `window:copy-item-to-pane-<direction>` and 
* `window:move-item-to-pane-<direction>`, 

where `<direction>` can be one of `on-left`, `on-right`, `above` or `below`. This makes them consistent with the `window:focus-pane-<direction>` commands.

I'll add specs for this during the weekend, and hopefully have it ready to ship next week. I'm also taking a look at the problems pointed out by @antcodd and @t9md in #8228.

Could you give me your comments on this, @benogle?

benogle:
@nathansobo or @maxbrunsfeld will take over on comments for this one. 

As far as the other PR, I think this PR should be scoped to what it currently is. I think we should keep #8228 as a separate unit of work.

src/workspace-element.coffee
```diff
@@ -95,6 +95,50 @@ class WorkspaceElement extends HTMLElement
 
   focusPaneViewOnRight: -> @paneContainer.focusPaneViewOnRight()
 
+  copyItemToPaneOnLeft: (destroyOriginal) ->
```
nathansobo: I'm a little confused about this `destroyOriginal` flag. Seems like if we wanted to destroy the original, maybe we should just *move* the item. Maybe another set of methods, `moveItemToPaneOn*` would be clearer and more efficient in this case.

nathansobo:
This is looking pretty close. Packages shouldn't really be dispatching commands to make things happy anyway, and there are low-level APIs on `Pane` for achieving splits programmatically. This seems more like giving users the option to bind keys to more atomic operations for split, copy, move.

My main concern is that we should implement move with the `Pane::moveItemToPane` method rather than copying and destroying. Other than that this is very close.

nathansobo:
Of course we'll need some spec coverage on what you have.

benogle:
Hey @acontreras89 how are you doing on this? This is a great addition, just requires some specs.

corysimmons:
:100: 🐎So close

acontreras89:
Hey @benogle (and everyone),
I'm sorry I have had this on hold for so long.

If I remember correctly, last time I talked with @nathansobo we agreed on some changes that were necessary. In short:
* First, code should be rearranged to have separate functions for copying and moving, instead of using `destroyOriginal` flag 
* Commands from `window` namespace (to copy/move files between panes) should not create new panes when none is found in the given direction. It felt kind of awkward on usage and (specially) when trying to write specs. In addition, this would make them more consistent with `window:focus-pane-*` commands.
* To achieve this and still maintain current behaviour, as well as keep the possibility of getting empty panes on split, `Pane`'s split API has still to be modified slightly.

In my opinion (given that default behaviour must be kept,) best way to achieve this would be having three separate commands (namely empty-split, copy-split, move-split) in `pane` namespace and three more (focus pane, and move and copy file to pane) in `window` namespace.

I can make this changes later today, but I might still need some help with the specs.

Anyone who thinks we're looping the loop, please give your comments

benogle:
No problem, glad you're still around! :+1: @maxbrunsfeld would be your point man this week.

keymaps/win32.cson
```diff
@@ -67,10 +67,10 @@
   'ctrl-_': 'window:decrease-font-size'
   'ctrl-0': 'window:reset-font-size'
 
-  'ctrl-k up': 'pane:split-up' # Atom Specific
-  'ctrl-k down': 'pane:split-down' # Atom Specific
-  'ctrl-k left': 'pane:split-left' # Atom Specific
-  'ctrl-k right': 'pane:split-right' # Atom Specific
+  'ctrl-k up': 'window:copy-item-to-pane-above' # Atom Specific
```
nathansobo: Maybe this should this be more explicit at the cost of being a bit longer... `copy-active-item-to-pane-above`. @maxbrunsfeld what do you think?

src/workspace-element.coffee
nathansobo: Same thought on the method name... maybe slightly more explicit at the expense of brevity: `copyActiveItemToPaneOnLeft`

nathansobo: One more comment... all these methods belong in `PaneContainerElement` and you should just delegate to them from `WorkspaceElement` like we do with `focusPaneViewOn*` methods.

nathansobo:
> First, code should be rearranged to have separate functions for copying and moving, instead of using destroyOriginal flag

I would even be comfortable renaming the method `moveActiveItemToPaneOn*` and passing an option to express that we want to copy the item. The opposite with the `destroyOriginal` flag feels counterintuitive to me however.

> Commands from window namespace (to copy/move files between panes) should not create new panes when none is found in the given direction. 

Seems reasonable. No reason to move an item to a pane that doesn't exist, and wouldn't really make sense to create it.

> To achieve this and still maintain current behaviour, as well as keep the possibility of getting empty panes on split, Pane's split API has still to be modified slightly.

If by this you mean that you're simply not passing `copyActiveItem: true` when splitting panes now by default, this makes sense. Splitting empty panes initially seems like it could be a better experience. But this doesn't seem like an API change to me, rather just the command using the API differently, so perhaps I'm misunderstanding your meaning here.

> but I might still need some help with the specs.

If you could give it a try, it would be helpful. There are probably specs you could emulate that already test the positional focusing methods.

acontreras89:
Thanks for your comments @nathansobo. I had some commits on my local repository that hadn't been pushed yet, so part of the work was already done.
I've updated my branch, included those commits, and made some new changes following your suggestions.

Now the split API in the `pane` namespace consists of three sets of commands:
* `split-direction?` (where `direction?` is one of `up`, `down`, `right` and `left`), which create empty panes
* `split-direction-and-action?-active-item` (where `action?` is either `copy` or `move`), which create a new pane and copy or move the active item to it

Similarly, two sets of convenience commands have been added to the `window` namespace:
* `action?-active-item-to-pane-direction?` (where `direction?` is one of `above`, `below`, `on-right` and `on-left`), which copy or move the active item to the nearest pane in the given direction, if one is available

Let me know if I missed something. In the meanwhile, I'll be trying to write some specs.

nathansobo:
:+1: That all sounds great. Thanks!

acontreras89:
I've updated the PR with:
* A fix for #8765 and a test to avoid regressions in the future
* Specs for the new functionalities in `Pane` and `PaneContainerElement`
* Minor readability changes

@nathansobo please, feel free to give me your feedback on the specs. Hope they are :ok_hand: but let me know if we need some more assertions

I haven't added specs for the new methods in `PaneContainer` yet.

src/pane-container-element.coffee
```diff
@@ -36,6 +36,27 @@ class PaneContainerElement extends HTMLElement
...
+    @moveActiveItemToNearestPaneInDirection('right', params)
+
+  moveActiveItemToNearestPaneInDirection: (direction, params) ->
+    destPane = @nearestPaneInDirection(direction)?.getModel()
+    return unless destPane?
+    if params?.keepOriginal
+      @model.copyActiveItemToPane(destPane)
+    else
+      @model.moveActiveItemToPane(destPane)
+    destPane.focus()
```
acontreras89: Should we call `destPane.focus()` on both `copyActiveItemToPane` and `moveActiveItemToPane` (in `PaneContainer`) instead?
Once I get some feedback on this, I'll add some specs for those methods.

thedaniel:
@acontreras89: I'll let @nathansobo give you feedback on the `destPane.focus()` calls (my vote would be yes, just call it in `{copy,move}ActiveItemToPane` but anyway), there are also a couple existing specs failing in `changing focus, copying and moving items directionally between panes` that need to be green. I looked through the test setup but figured you had better context on the numbered pane items under test and would be able to fix up the specs more quickly than I could.

acontreras89:
Hey @thedaniel, thanks for your comments
Yeah, I know which tests are failing and why.. :sweat_smile: I made some changes on my code and forgot to update the specs (sorry about that!)
I was waiting for opinions on the `destPane.focus()` thing to push the fix along with the remaining specs.

acontreras89:
Last week I updated the specs pointed out by @thedaniel but one of the builds still failed. I've no idea what the new problem is :sweat_smile: can I get some help? @maxbrunsfeld @nathansobo

50Wliu:
@acontreras89 I've restarted the build - looks like it was just a network issue.

acontreras89:
Thanks @50Wliu! :kissing_heart:

src/pane-container.coffee
```diff
@@ -164,6 +164,17 @@ class PaneContainer extends Model
     else
       false
 
+  moveActiveItemToPane: (destPane) ->
+    item = @activePane.getActiveItem()
+    @activePane.moveItemToPane(item, destPane)
+    destPane.setActiveItem(item)
+    destPane.focus()
```
nathansobo: I think it might be better to perform these `focus` calls in the view layer, for two reasons.

First, focus is a view-level concept. That method is really intended to be called from the view layer currently, leaving `activate` as the corresponding model level concept. The difference between them is pretty subtle: a pane can be "active" without being focused. We usually activate the most recently focused pane.

Second: It might be nice to keep these model-level methods a bit more focused on what they claim to do in their names, which is moving/copying items. I can even see them graduating to the public API someday, in which case messing with focus state might be surprising.

Let me know what you think of my reasoning here... This may have some impact on your tests, which are looking pretty good.

nathansobo:
@lee-dohm I think this is a pretty decent improvement... One side-effect however is that anyone who has custom bindings for `split-[direction]` commands is going to get a change in behavior when this rolls out. Instead of the current item getting copied, the new pane will now be created empty. I think it's worth it to make the commands explicit though. If people want the old behavior they'll need to re-bind to `split-[direction]-and-copy-active-item`. Just wanted to give you a heads up about that and give you a chance to give feedback.

nathansobo:
Other than my minor suggestion about changing focus, this is looking good to me. Thanks for the thorough spec coverage @acontreras89 and sticking with this PR over *months* to get it merged. We really appreciate it. :sparkles:

lee-dohm:
Thanks for the heads up. I'll put together a communication plan.

src/pane-container.coffee
acontreras89: I definitely agree. The only reason that led me to make this change is that it felt _better_ when writing specs. I can remove/update the assertions on the active pane, though.
I'll change it back right away.

acontreras89:
Hey @nathansobo, @lee-dohm, just one comment: I've kept current behaviour for the default key bindings -- they're currently set to the `pane:split-[direction]-and-copy-active-item` commands.
I was told long time ago I shouldn't make changes to default behaviours until the new features have been field-tested thoroughly.

I do agree that making split create empty panes by default is the best approach, though. I just want to make sure you want me to change it in this PR.

lee-dohm:
I talked with @acontreras89 via Slack and I mentioned that we want to maintain the built-in keybindings behavior (such as <kbd>Cmd+K</kbd> <kbd>Cmd+Right</kbd>). I clarified that I will be informing the community that *custom* keybindings in `keymap.cson` or packages calling the commands directly (though that is generally discouraged) will need to be updated.

nathansobo:
Thanks for this improvement @acontreras89! It's gonna be nice.

lee-dohm:
Posted about this here:

https://discuss.atom.io/t/upcoming-change-to-the-pane-commands-in-v1-6-0/25407

Let me know if I got any of it wrong :blush:

acontreras89:
Thanks for the good work @nathansobo, @lee-dohm!
:ship: :tada:

nathansobo:
@oorahduc That behavior should be pretty straightforward to script. Something like this in your `init.js` should work:

```js
atom.commands.add('atom-workspace', {
  'user:split-right-and-open-empty-editor': () => {
    atom.workspace.getActivePane().splitRight()
    atom.workspace.open()
  }
})
```

PR #7957, opened 2015-07-18
Enable subpixel AA for tabs, status-bar etc.

### LINKED ISSUE #7904: Showing line numbers disables sub-pixel AA on tabs, status-bar etc.
In https://github.com/atom/one-dark-ui/issues/82, @wiox noticed that sub-pixel anti aliasing is not only disabled for the tabs, but also on the fuzzy finder and the status bar.

sub-pixel

It seems it's caused by the line number tiles getting promoted to a GPU layer and probably related to #7101. Kinda strange that it affects different areas.

Something that might be worth looking into is using `will-change`. It also adds a GPU layer for the line numbers, but doesn't affect the rest. Notice the orange "composited layer borders" only appear in the status bar with `-webkit-transform: translate3d(0px, 0px, 0px);` but not with `will-change: transform;`.

tiles

Now, I *heard* that `will-change` should only be used when interacting, like on hover or so.. and not kept permanently enabled. So not sure if that would help. /cc @as-cii @nathansobo 

__Edit:__ Sorry, I just realized that the `transform` is needed to actually move line numbers. So can't be replaced. :grin:

simurai:
### Problem

The GPU layer added to the .gutter .tiles cause the surounding UI to also composite layers, like the tas or the status-bar. This disables subpixel anti-aliasing and text doesn't look that sharp.

Eventhough the `.gutter` has `overflow: hidden`, it seems Chrome still thinks there is a chance of overlapping and thus creates the extra layers.
### Solution

Creating an own stacking context with `z-index` for the `.gutter` seems to fix it.

z-index

Fixes #7904

### DIFF
--- a/static/text-editor-shadow.less
+++ b/static/text-editor-shadow.less
@@ -17,6 +17,7 @@
 
 .gutter {
   overflow: hidden;
+  z-index: 0;
   text-align: right;
   cursor: default;
   min-width: 1em;


as-cii:
Looks good to me, @simurai! Nice job! :sparkles:

as-cii:
/cc: @nathansobo for one last look :eyes:.

nathansobo:
:heart: Thanks to you both. :heart:

PR #7960, opened 2015-07-19
Make activatePackage and deactivatePackage public.

oggy:
The former is used in the tests in generated packages.


benogle:
Closing in favor of #7878

PR #7968, opened 2015-07-19
switch Slackware build to iojs

ryanpcmcquen:


ryanpcmcquen:
Here is the upstream page:

http://slackbuilds.org/repository/14.1/development/atom/

As you can see, it lists:

    This requires: iojs

ryanpcmcquen:
We are switching back to node, so I will close this.

PR #7996, opened 2015-07-20
Upgrade to Babel 5.8.3

bolinfest:
Once https://github.com/babel/babel/issues/2028 is fixed, we can start
picking up new versions of Babel again.

/cc @sebmck

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "async": "0.2.6",
     "atom-keymap": "^5.1.6",
     "atom-space-pen-views": "^2.0.4",
-    "babel-core": "^5.6.17",
+    "babel-core": "^5.8.3",
     "bootstrap": "^3.3.4",
     "clear-cut": "^2.0.1",
     "coffee-cash": "0.8.0",


sebmck:
Should have this fixed in the next couple hours.

bolinfest:
@sebmck I see v5.8.3 is now available on https://www.npmjs.com/package/babel-core, but I only see up to v5.8.1 in https://github.com/babel/babel/commits/master.

In any event, what do you think I should change this to? `^5.8.1`?

sebmck:
@bolinfest Travis is supposed to be merging from the [`development`](https://github.com/babel/babel/commits/development) branch into `master` on CI success, looks like that malfunctioned. `^5.8.3` would be best since there were some issues with the publishing process were a file wasn't included with borked the install.

bolinfest:
Sounds good: changed to `^5.8.3`.

thomasjo:
Pretty confident the one CI failure was caused by problems on the Travis side of things, but I kicked off a new build anyway to be on the safe side. I'll merge when it goes :green_heart:

thomasjo:
PR build is :green_heart:, so I'll ignore the failing "push" build (likely due to a forced push or some such).

PR #7999, opened 2015-07-21
Support Package Sets

joefitzgerald:
- Ensure that activate and deactivate are no-ops
- Rationalize package type, retain legacy getType
- This should be merged with https://github.com/atom/apm/pull/385

### DIFF
--- a/src/package-set.coffee
+++ b/src/package-set.coffee
@@ -1,0 +1,11 @@
+Q = require 'q'
+Package = require './package'
+
+module.exports =
+class PackageSet extends Package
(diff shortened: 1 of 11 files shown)


src/package-set.coffee
```diff
@@ -0,0 +1,11 @@
+Q = require 'q'
```
steelbrain: Why not use the native promises? This line is redundant.
SideNote: You are requiring the variable `Q` and doing `Promise.resolve`, so it already uses native ones

Issue #4019, opened 2014-10-31
Atom has excessive battery impact compared to other apps on Mac OS Yosemite


AlexYursha:
Atom's energy impact goes up to 25 points of energy impact with just 4 tabs open when idle. To compare Sublime Text 3 uses just 0.1, Xcode - 0.9 and Intellij IDEA - 8 points when idle. When moving keyboard focus to Intellij IDEA it still uses less power than idle Atom. Thus Yosemite reports Atom as an app using significant energy in system bar.

rdlugosz:
While there may be some "artificial reporting" going on, in general the energy footprint of Chromium is high relative to other browsers running on OS X. For example, spend a little time just doing regular web browsing with Safari and then similar activity with Chrome. You'll notice that your time with Chrome has used a lot more battery than Safari had—even if you aren't hitting sites with a lot of flash, etc.

This bug should probably be closed as "not an Atom bug" unless someone is able to directly show that Atom itself is contributing to resource usage above and beyond what Chromium introduces as a baseline.

paulcbetts:
There are certainly things that Atom could do even at a JavaScript level to improve battery usage. Removing as many persistent timers as possible is high on the list

rdlugosz:
I did a bit of experimenting and would agree—Atom certainly seems to drain battery faster than just browsing in Chrome. In fact, opening just an empty Atom window and minimizing it costs you a constant 4-8% CPU usage. Color me convinced that there's a problem!

bitardo:
Any update on this?
I confirm – *idle* Atom uses *more* than a VMware machine running Windows and about as much as Safari during active browsing. This is a major concern since a lot of us work unplugged.

![screenshot at may 24 11-27-28](https://cloud.githubusercontent.com/assets/12165130/7787204/ff5c197c-0207-11e5-8b16-e55348adbe26.png)

mnquintana:
@bitardo I think you may be misreading that – idle Atom has an energy impact of 0.4 which is pretty negligible, but an average energy impact of 100.62 when in use. Since Atom is built on Chromium it's not surprising that it would have an average energy impact comparable to another WebKit browser.

SolomonBier:
Experiencing same problem-surprising clocking 100+ when idle

carlhopf:
running OS X, atom 1.0:

i'm getting an energy impact of ~11 when idling while window is focused.

[...]

Issue #7995, opened 2015-07-20
The Quality Initiative


benogle:
The Atom team will be focusing on quality for the next month or so. This effort is timeboxed: the plan is to finish around the __beginning of Sept__. This means __we probably won't get everything done on this list__ during the allotted time window. After this time window, unfinished things will work into our normal issue handling routine.

### What does this mean?

The goal of the quality initiative is to focus on fixing bugs and adding features that impact a lot of people, but are reasonably low effort. Why? There are a lot of low effort things affecting many people, and a lot of issues that slipped through the cracks while we were focused on the API and 1.0. We're just giving our selves some time to focus on them.

These will be small projects, ideally 1-2 days max. We will avoid rewrites or major refactoring.

The quality initiative will be our main focus, but we will still handle issues and review PRs&mdash;think of this effort as our 80% time.

### What about X bug/feature?

After this quality initiative, we plan on shifting our 80% time to focus on things like [performance](https://github.com/atom/atom/issues/307), better support for [international users](https://github.com/atom/atom-keymap/issues/37), and some features we have in mind.

If you know of a bug or feature that is high-impact / low-effort, not in the list below, and does not fall under performance or internationalization, comment on this issue and we'll figure out if it fits in.

#### Bug

* [ ] Fix largest uncaught errors in bugsnag
* [ ] Continue fixing uncaught errors from https://github.com/atom/atom/issues/7740
  * [x] Cannot read property 'dispose' of undefined https://github.com/atom/tabs/issues/183
  * [x] Uncaught TypeError: Cannot read property 'isEmpty' of undefined https://github.com/atom/bracket-matcher/issues/102
* [x] Fix erasing files when no disk space 
  * [x] https://github.com/atom/atom/issues/3158 @nathansobo
* [ ] Fix app auto-update issues
  * [ ] https://github.com/atom/atom/issues/7833
  * [ ] https://github.com/atom/atom/issues/6612
  * [ ] https://github.com/atom/atom/issues/7854
  * [ ] better feedback https://github.com/atom/atom/issues/8038
* [ ] Fix app update and release notes issues
  * [ ] https://github.com/atom/atom/issues/5692
  * [ ] https://github.com/atom/atom/issues/6922
  * [ ] https://github.com/atom/atom/issues/3757
  * [ ] https://github.com/atom/atom/issues/6482
* [ ] Investigate issues when changing branches
  * [ ] https://github.com/atom/atom/issues/4982
[...]

PR #8009, opened 2015-07-21
Change Document Polling from Time Interval to Event 

jlord:
:pear: ing with @nathansobo 

Previously document polling happened through timed intervals causing continued use of CPU when the editor was unfocused. 

This PR changes the polling to occur when the DOM is mutated or the window is resized (since that doesn't fall under a DOM mutation). Additionally it changes the way Atom updates the CSS so that it occurs as a mutation and the event is captured. Furthermore! It prevents the cursor from ever starting to blink if it wasn't blinking already which can be caused, for instance, when content is changed in the Settings View.

We've added test coverage but are still testing this branch out to be sure we've:
- covered all the DOM changing scenarios
- not added additional CPU usage when editor _is_ active

cc @atom/feedback, try out this branch too if ya can :+1: 

Fixes https://github.com/atom/atom/issues/4019
Quality Initiative https://github.com/atom/atom/issues/7995

### DIFF
--- a/spec/view-registry-spec.coffee
+++ b/spec/view-registry-spec.coffee
@@ -7,6 +7,9 @@ describe "ViewRegistry", ->
   beforeEach ->
     registry = new ViewRegistry
 
+  afterEach ->
+    registry.clearDocumentRequests()
+
   describe "::getView(object)", ->
     describe "when passed a DOM node", ->
       it "returns the given DOM node", ->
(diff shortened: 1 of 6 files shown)


spec/view-registry-spec.coffee
```diff
@@ -1,12 +1,15 @@
 ViewRegistry = require '../src/view-registry'
 {View} = require '../src/space-pen-extensions'
 
-describe "ViewRegistry", ->
+fdescribe "ViewRegistry", ->
```
benogle: nof

nathansobo:
Thanks @jlord!

jlord:
wowoowo! Awesome! Thank you @nathansobo!

FipFop:
These PR's haven't changed much.

screen shot 2016-05-11 at 12 13 35 am

PR #8010, opened 2015-07-21
Default maintainHistory marker option to false

maxbrunsfeld:
This is a change I've been meaning to make for a while. I made `find-and-replace` and `spell-check` explicitly opt out of marker snapshotting, but in other situations where there are lots of markers (e.g. the linter w/ lots of errors, a big file w/ lots of folds) the marker snapshotting causes noticeable lag.
#### Before

In a 2000-line file w/ 150 folds, `MarkerStore::createSnapshot` takes a long time before and after each change. This character insertion **took 15 ms**.

screen shot 2015-07-21 at 2 43 16 pm
#### After

Now that part of the flame graph is gone, because folds aren't snapshotted. Folds shouldn't be snapshotted, because they are destroyed if touched by a change anyway. The same character insertion **took 4 ms**.

screen shot 2015-07-21 at 2 43 33 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "space-pen": "3.8.2",
     "stacktrace-parser": "0.1.1",
     "temp": "0.8.1",
-    "text-buffer": "6.3.11",
+    "text-buffer": "6.4.0-0",
     "theorist": "^1.0.2",
     "typescript-simple": "1.0.0",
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
I'm going to build this branch and use it for a while to make sure it doesn't cause problems.

PR #8019, opened 2015-07-22
Document return type of `Gutter::decorateMarker()`.

bolinfest:
@jssln @nathansobo I just want to make sure this wasn't deliberately omitted.


bolinfest:
Oops, this is not supposed to be a stacked diff...

bolinfest:
eff me

Issue #7103, opened 2015-06-04
'copy' keyboard shortcut should not copy empty space


DavidAwad:
It's too frequent that someone might accidentally hit the 'c' key instead of the 'v' key when pasting in atom. 

instead of copying empty space, the copy shortcut should simply not do anything. 

I can write the pull request myself if someone can point me in the right direction, either way if we could  improve this very annoying problem it would be great!

benogle:
This was a deliberate decision. cc @maxbrunsfeld. I admit, I dislike it as well, but many people wanted this behavior.

mnquintana:
@benogle This seems like an ideal candidate for a config option though

DavidAwad:
A config option would be perfect!

maxbrunsfeld:
Just to clarify, in general it copies the line under the cursor, not empty space. I'd be ok w/ a config option.

DavidAwad:
So how would I implement this? I have no problem submitting a PR for it, it seems like a simple request.

vemv-kantox:
I second this issue. People actually leveraging the current behavior may like a similiar but different key-binding: `command-shift-c` (free by default) could copy the current line regardless of selection.

maxbrunsfeld:
> People actually leveraging the current behavior may like a similar but different key-binding.

This is good idea. @DavidAwad I think that the best way to do this would be to add a new command, `editor:copy-selection` that would write to the clipboard *only* if text were selected. That way, users who prefer that behavior could simply add this to their keymap:

```coffee
'atom-text-editor':
  `command-c`: `editor:copy-selection` # mac os
  `ctrl-c`: `editor:copy-selection`    # linux and windows
```

lpommers:
I've playing with this locally and just looking for some input. For me I think it would be nice if copy didn't work unless something was selected *or* the cursor is at the beginning of the line. In that case I think I would find it useful if `copy` copied the line. Any thoughts?

jschroeder9000:
For me, I would prefer copy to do nothing if nothing is selected, even if the cursor is at the beginning of the line.  If I accidentally hit ^c instead of ^v, I don't want it to blow away my clipboard contents, even if the cursor is at the beginning of the line.  Nothing selected => nothing to do.

MichaelJCole:
Thanks so much!  This drove me mad.  Here's how to implement it:

Edit -> Preferences -> Keybindings -> Click link at very top "your keymap file"
This should open a file in the editor.

Paste this:
```
'body':
[...]

PR #8025, opened 2015-07-22
Copy selection command to only write to clipboard if text selected

lpommers:
Closes https://github.com/atom/atom/issues/7103

Add a new command, `editor:copy-selection`, that writes to the clipboard only if the user has selected text

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2994,6 +2994,31 @@ describe "TextEditor", ->
               sort
               items
             """
+      describe ".copyOnlySelectedText()", ->
+        it "copies selected text onto the clipboard", ->
+          editor.setSelectedBufferRanges([[[0, 4], [0, 13]], [[1, 6], [1, 10]], [[2, 8], [2, 13]]])
+
+          editor.copyOnlySelectedText()
+          expect(buffer.lineForRow(0)).toBe "var quicksort = function () {"
+          expect(buffer.lineForRow(1)).toBe "  var sort = function(items) {"
+          expect(buffer.lineForRow(2)).toBe "    if (items.length <= 1) return items;"
+          expect(clipboard.readText()).toBe 'quicksort\nsort\nitems'
+          expect(atom.clipboard.read()).toEqual """
+            quicksort
+            sort
+            items
+          """
+
+        describe "when no text is selected", ->
+          it "copies the line on which the cursor is at the begginning of the line", ->
+            editor.setCursorBufferPosition([1, 0])
+            editor.copyOnlySelectedText()
+            expect(atom.clipboard.read()).toEqual "  var sort = function(items) {\n"
+
+          it "does not copy anything if the cursor is not at the beginning of the line", ->
+            editor.setCursorBufferPosition([1, 5])
+            editor.copyOnlySelectedText()
+            expect(atom.clipboard.read()).toEqual "initial clipboard content"
 
       describe ".pasteText()", ->
         copyText = (text, {startColumn, textEditor}={}) ->
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -3010,12 +3010,7 @@ describe "TextEditor", ->
           """
 
         describe "when no text is selected", ->
-          it "copies the line on which the cursor is at the begginning of the line", ->
-            editor.setCursorBufferPosition([1, 0])
-            editor.copyOnlySelectedText()
-            expect(atom.clipboard.read()).toEqual "  var sort = function(items) {\n"
-
-          it "does not copy anything if the cursor is not at the beginning of the line", ->
+          it "does not copy anything", ->
             editor.setCursorBufferPosition([1, 5])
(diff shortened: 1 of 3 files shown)


src/text-editor.coffee
```diff
@@ -2587,6 +2587,18 @@ class TextEditor extends Model
...
 
+
+  # Essential: For each selection, only copy hightlighted text. Line is coptied if no selected text and cursor is at the beginning of the line
+  copyOnlySelectedText: ->
+    maintainClipboard = false
+    for selection in @getSelectionsOrderedByBufferPosition()
+      if not selection.isEmpty()
+        selection.copy(maintainClipboard, true)
+      else if selection.isEmpty() and selection.cursor.isAtBeginningOfLine()
+        @copySelectedText()
```
izuzak: Just wondering why you decided to call `copySelectedText()` here instead of inlining the part for copying the line? Calling `copySelectedText()` will cause _all selections_ to be processed again within that function, not just the selection currently being processed within `copyOnlySelectedText()`, right? While that normally won't be a performance problem, I think avoiding doing the same thing over and over again (since this would be a loop nested within the same loop) is always nice. 

I also wonder if calling `copySelectedText()` here might cause problems in some cases because _all selections_ will be processed according to the rules of that function, and not just by the rules of `copyOnlySelectedText()`? I'm thinking of cases with multiple selections, where one selection is empty and at the beginning of the line, and the other is empty but not at the beginning of the line.

izuzak:
cc @maxbrunsfeld for :eyes: and :thought_balloon: since you were looking at #7103.

src/text-editor.coffee
lpommers: @izuzak I decided to use `copySelectedText()` because I thought reusing code would be nicer than inlining that part, but I hadn't considered that it would process all selections again - so I think you're right - it would be best to avoid doing the same thing twice. I'll make some changes :)

lpommers:
I'm not entirely sure why this PR is failing the build. It doesn't look like the error `Error Downloading Update: Could not get code signature for running application` is related to any of my changes

src/text-editor.coffee
```diff
@@ -2587,6 +2587,16 @@ class TextEditor extends Model
...
     return
 
+
+  # Essential: For each selection, only copy hightlighted text. Line is coptied if no selected text and cursor is at the beginning of the line
+  copyOnlySelectedText: ->
+    maintainClipboard = false
+    for selection in @getSelectionsOrderedByBufferPosition()
+      if not selection.isEmpty()
+        selection.copy(maintainClipboard, true)
+      maintainClipboard = true
```
maxbrunsfeld: I would think that the `maintainClipboard = true` statement should be *inside* of the `if not selection.isEmpty()` block. Otherwise, if the first selection is empty and the second selection is not, the existing clipboard entry will be mutated unexpectedly.

lpommers:
@maxbrunsfeld this PR is now up to date based on your comment

src/text-editor.coffee
```diff
@@ -2587,6 +2587,16 @@ class TextEditor extends Model
       maintainClipboard = true
     return
 
+
+  # Essential: For each selection, only copy hightlighted text. Line is coptied if no selected text and cursor is at the beginning of the line
```
maxbrunsfeld: I'd actually like to *not* add this to our public API right now. Could you change `Essential` -> `Private` here?

Also, there's a small typo: `hightlighted` -> `highlighted`.

Also, I don't think the last sentence of the comment is true; it doesn't look like there's any special behavior of the cursor is at the beginning of the line. Did you intend to implement that behavior? If not, we should just remove that sentence from the comment.

lpommers: yeah, in an earlier version of this branch, that comment was true - but then based on some feedback from what I read in the discussion in #7103 I removed it, so it's no longer true. I'll remove that and update the rest

lpommers:
I'm not sure why this failed build `#2699.2`. As for ``#2699.3`, i'm not sure what I could have changed to make it fail a package spec. Any guidance on how i can look into see what would cause this package spec to fail

src/text-editor.coffee
```diff
@@ -2587,6 +2587,16 @@ class TextEditor extends Model
       maintainClipboard = true
     return
 
+
```
maxbrunsfeld: Could you remove this second blank line?

spec/text-editor-spec.coffee
```diff
@@ -2994,6 +2994,27 @@ describe "TextEditor", ->
               sort
               items
             """
+      describe ".copyOnlySelectedText()", ->
```
maxbrunsfeld: Could you add a blank line here, separating the two `describe` blocks?

maxbrunsfeld:
Other than the placement of two `\n`s, this looks great. Thanks for taking the time to add this!

lpommers:
shoot, sorry about the newlines, i'll make a quick change and push

maxbrunsfeld:
Great!

izuzak:
:clap:

PR #8041, opened 2015-07-23
Use `border-box` to prevent losing subpixel AA

### LINKED ISSUE #7316: Some text within code loses boldness
You'll note that some of the keywords are not bold while others are. This seems to occur frequently and arbitrarily. If I scroll the non-bold text out of view and bring it back, the bold comes back.

### LINKED ISSUE #7382: Subpixel anti-aliasing not working on some lines
schermafbeelding 2015-06-23 om 12 01 09

I think the screenshot says it all.
I'm using Atom 0.211 on Max OS X 10.3

The brighter colors (although I prefer them) are "wrong". The are rendered with greyscale anti-aliasing, in contrast to the other lines which have normal subpixel anti-aliasing.

This particular occurrence of the bright lines is 12 lines high, one of them being soft wrapped.

The occurrence and position of a bright area is rather arbitrary. I can't readily reproduce it, because slightly resizing the window horizontally removes it or produces it elsewhere. Switching tabs seems to fix it.

### LINKED ISSUE #7772: Subpixel antialiasing isn't consistently applied to all text in document
* Using Atom 1.0.0.
* Using Windows 7.
* The effect seems to happen on both of my monitors:
 * Primary monitor is 1920 x 1200 resolution.
 * Secondary monitor is 1280 x 1024 resolution.
* Using One Light for both UI Theme and Syntax Theme, but it also happens if I use Atom Light on both.
* Running atom --safe didn't make any difference (still has the same alternating blurry).

It seems that when editing some files, that subpixel antialiasing works for a couple lines, then doesn't work for a couple lines, then works for a few lines, then doesn't ...

capture

When you zoom in, you can see that lines 34-36 are working properly, but then line 37 doesn't have subpixel antialiasing applied.

zoom

The strange thing is that if I make a document of all "a" characters, for example, it *all* looks blurry, but at least it is consistent.

all_blurry

If I fill the document with all "0" characters, it looks like the top few lines are blurry, but the rest renders correctly.

as-cii:
When highlights are positioned outside the line box, they could get hidden by a subsequent tile (because of its opaque background). As a result, subpixel anti-aliasing gets disabled by Chrome.

Using `border-box` as the default `box-sizing` allows highlights to always fit within the line. This won’t cover cases where a style like `margin` is applied (causing the highlight to go outside its boundaries), but I am not sure we should consider such scenarios given that highlights should be positioned right inside lines.

Fixes #7382
Fixes #7316
Fixes #7772

@karai17 @w1res @burodepeper: it would be great if you guys could try this branch out to see if it actually fixes the issues you reported. If you don’t know how to build this branch, you can also put the following snippet in your personal stylesheet:

``` less
atom-text-editor::shadow .region {
  box-sizing: border-box;
}
```

Thank you! :pray: 

/cc: @nathansobo @atom/feedback for additional :eyes:

### DIFF
--- a/src/highlights-component.coffee
+++ b/src/highlights-component.coffee
@@ -76,6 +76,10 @@ class HighlightsComponent
       unless oldHighlightState.regions[i]?
         oldHighlightState.regions[i] = {}
         regionNode = document.createElement('div')
+        # This prevents highlights at the tiles boundaries to be hidden by the
+        # subsequent tile. When this happens, subpixel anti-aliasing gets
+        # disabled.
+        regionNode.style.boxSizing = "border-box"
         regionNode.classList.add('region')
         regionNode.classList.add(newHighlightState.deprecatedRegionClass) if newHighlightState.deprecatedRegionClass?
         @regionNodesByHighlightId[id][i] = regionNode


burodepeper:
I have no idea how to try out a branch, but if I look at the changes you made, I assume I can also try this out by adding the fix to my stylesheet. If so, what should I add?

as-cii:
> I have no idea how to try out a branch, but if I look at the changes you made, I assume I can also try this out by adding the fix to my stylesheet. If so, what should I add?

Thanks for pointing this out, @burodepeper. :sparkles: I have updated the PR description to include the styles you should add in the stylesheet. :art:

burodepeper:
After using the patch for about two hours, I haven't seen any glitches. Removed the patch, and immediately spotted a glitch. Applied the patch again, and it seems to have fixed my issues! Woohoo!

nathansobo:
@as-cii Merge at will.

w1res:
@as-cii I'm not sure how to compile Atom, but when I added 

```
atom-text-editor::shadow .region {
  box-sizing: border-box;
}
```

to my Stylesheet, everything looks correct, even with the spell-check package enabled.

mark-hahn:
That's weird because everything in Atom already has  box-sizing: border-box.

nathansobo:
Shadow DOM?

as-cii:
> Shadow DOM?

Yup! :ok_hand: 

Thanks everyone! :sparkles: :bow:

PR #8053, opened 2015-07-23
Fire event before a pane is destroyed.

oggy:
Rationale: I don't like losing my tabs when I close a pane (or accidentally close the wrong pane), so want to be able to move them into another pane just before the pane is destroyed.

This adds both `pane.onWillDestroy -> ...` and `workspace.onWillDestoryPane (pane) -> ...`, analogous to the pane item callbacks.


maxbrunsfeld:
:+1: Looks good to me.

thomasjo:
Looks good to me as well.

thomasjo:
Thanks @oggy!

PR #8082, opened 2015-07-25
Update language-javascript to 0.84.0

winstliu:
0.83.0 fixes an issue with regex highlighting after a `const` declaration.
0.84.0 fixes the `setInterval` and `setTimeout` snippets not being properly escaped.


50Wliu:
`language-javascript` specs are passing and the failing core specs are also broken on master, so :ship:ing.

Issue #2956, opened 2014-07-14
Add Linux software repositories for app update support


netroby:
launchpad.net
would you please release atom for ubuntu , as a launchpad ppa repository?

190n:
+1

thedaniel:
IIRC this is blocked until Atom can be built without network access.

50Wliu:
How does the Webupd8 PPA do it?  Perhaps their method could be used for this?

Joshua-Anderson:
They zip up builds of atom and extract it into the deb. It isn't a great solution. Ideally we would have a command the downloads but does not build all dependencies.

Dannyzen:
+1

Joshua-Anderson:
As a alternative to launchpad, [packagecloud](packagecloud.io) also offers free apt repository posting and allows you to directly upload Debian packages. It's a single command to upload and would be easy to do as part of a release process. I'd be happy to help set it up if there is interest, but it would need to mostly be done by whoever uploads the official Debian packages to the atom website.

amingilani:
+1000

Please, please, please.

wreckah:
:+1:

SudoPluto:
For the rpm side of things, the openSUSE build service could be used to provide repositories for rpm based distros.

devurandom:
openSUSE can also build Ubuntu packages, afaik. (See e.g. the ownCloud Client repository it builds.)

Yajo:
@devurandom Posting the link FWIW: http://software.opensuse.org/download/package?project=isv:ownCloud:community&package=owncloud

sudopluto:
Since nobody else wants to take the initiative, I guess I will. I wanted to avoid doing this because I am new to Linux, but maybe it will be a good learning experience. I have never compiled source code before, so if you want to help, pls let me know.

alanfranz:
@sudopluto the point is that many people want an "official" repository from atom developers, not a 3rd party repository.

If you want a 3rd party repository hosting **exactly** the release .deb which are released by atom developers, there's already my repo at www.a9f.eu

sudopluto:
@thedaniel
>IIRC this is blocked until Atom can be built without network access.

Copr (the Fedora version of ppa) has a build service that has internet access. There is already unofficial copr repo [here:](https://copr.fedoraproject.org/coprs/helber/atom/)
[...]

PR #8093, opened 2015-07-27
Add update mechanism for :penguin:

stanier:
Temporary(?) solution to #2956 while better solution is discussed.

### DIFF
--- a/src/browser/auto-update-manager.coffee
+++ b/src/browser/auto-update-manager.coffee
@@ -23,15 +23,17 @@ class AutoUpdateManager
       @feedUrl = 'https://atom.io/api/updates'
     else
       @iconPath = path.resolve(__dirname, '..', '..', 'resources', 'atom.png')
-      @feedUrl = 'https://atom.io/api/updates'
+      @feedUrl = "https://atom.io/api/updates?version=#{@version}"
 
     process.nextTick => @setupAutoUpdater()
 
   setupAutoUpdater: ->
     if process.platform is 'win32'
       autoUpdater = require './auto-updater-win32'
-    else
+    else if process.platform is 'linux'
       autoUpdater = require './auto-updater-linux'
+    else
+      autoUpdater = require 'auto-updater'
 
     autoUpdater.on 'error', (event, message) =>
       @setState(ErrorState)
@@ -41,8 +43,8 @@ class AutoUpdateManager
 
     autoUpdater.on 'checking-for-update', =>
       @setState(CheckingState)
-
     autoUpdater.on 'update-not-available', =>
+
       @setState(NoUpdateAvailableState)
 
     autoUpdater.on 'update-available', =>
@@ -53,7 +55,7 @@ class AutoUpdateManager
       @emitUpdateAvailableEvent(@getWindows()...)
 
     # Only released versions should check for updates.
-    @scheduleUpdateCheck()# unless /\w{7}/.test(@version)
+    @scheduleUpdateCheck() unless /\w{7}/.test(@version)
 
     switch process.platform
       when 'win32'
--- a/src/browser/auto-update-manager.coffee
+++ b/src/browser/auto-update-manager.coffee
@@ -23,7 +23,7 @@ class AutoUpdateManager
       @feedUrl = 'https://atom.io/api/updates'
     else
       @iconPath = path.resolve(__dirname, '..', '..', 'resources', 'atom.png')
-      @feedUrl = "https://atom.io/api/updates?version=#{@version}"
+      @feedUrl = 'https://atom.io/api/updates'
 
     process.nextTick => @setupAutoUpdater()
 
@@ -31,7 +31,7 @@ class AutoUpdateManager
     if process.platform is 'win32'
       autoUpdater = require './auto-updater-win32'
     else
-      autoUpdater = require 'auto-updater'
+      autoUpdater = require './auto-updater-linux'
 
     autoUpdater.on 'error', (event, message) =>
       @setState(ErrorState)
@@ -53,13 +53,13 @@ class AutoUpdateManager
       @emitUpdateAvailableEvent(@getWindows()...)
 
     # Only released versions should check for updates.
-    @scheduleUpdateCheck() unless /\w{7}/.test(@version)
+    @scheduleUpdateCheck()# unless /\w{7}/.test(@version)
 
     switch process.platform
       when 'win32'
         @setState(UnsupportedState) unless autoUpdater.supportsUpdates()
       when 'linux'
-        @setState(UnsupportedState)
+        @setState(UnsupportedState) unless autoUpdater.supportsUpdates()
 
   emitUpdateAvailableEvent: (windows...) ->
     return unless @releaseVersion?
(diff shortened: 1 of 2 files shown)


p-e-w:
This PR is a good shot at fixing the single biggest issue on Linux. It has now been tagged "needs-review" for almost a month. Any chance of speeding up the review process so there can be progress on this issue?

The way I see it, while the approach of trying to identify the distro and then downloading packages is not pretty, it's the only way to move forward given the circumstances. There is one – and only one – proper way to install and update packages on Linux: From repositories. If the Atom team is unwilling to create official aptitude/yum repos, the "semi-manual" approach is the next best thing. Using third-party repos is simply not an option in many environments due to security concerns.

antond:
This would be great to have. I hope it gets merged soon.

Zireael07:
Has it been reviewed at all? There are no comments to this effect...

stanier:
@Zireael07 I can't really attest to the functionality of this fix as it was only tested on two local builds and I didn't really care to mess with CoffeeScript's syntactic bull.  I think the only part that really needs "reviewing" is the decision of whether or not to continue using the release channel for updates or to make use of a ppa (as we await a response in #2956).

In my personal opinion, the update process should be automated regardless.  I might push a commit later today to touch up some things and make it more likely to be implemented.

src/browser/auto-update-manager.coffee
```diff
@@ -41,8 +43,8 @@ class AutoUpdateManager
 
     autoUpdater.on 'checking-for-update', =>
       @setState(CheckingState)
-
     autoUpdater.on 'update-not-available', =>
+
```
50Wliu: Why was this line added, and why was the above one removed?

src/browser/auto-update-manager.coffee
```diff
@@ -59,7 +61,7 @@ class AutoUpdateManager
       when 'win32'
```
50Wliu: This can be combined into `when 'win32', 'linux'`

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+    error?.code ?= code
+    error?.stdout ?= stdout
+    callback?(error, stdout)
+
+class AutoUpdater
+  _.extend @prototype, EventEmitter.prototype
+
+  stdout = ChildProcess.execSync('cat /etc/issue').toString().replace(/\\./g, "")
+
+  if distros.ubuntu.test(stdout) | distros.debian.test(stdout)
```
50Wliu: Shouldn't this be using `||` instead of `|`?

stanier:
Passes unit tests now.

As a side note, this is why I generally dislike CoffeeScript (see a6d4b91 d9c68a4)

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+        return
+
+      @emit 'update-available'
+
+      @installUpdate (error) =>
+        if error?
+          @emit 'update-not-available'
+          return
+
+        console.log 'will update'
```
50Wliu: ;o

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+  spawnedProcess.on 'close', (code, signal) ->
+    error ?= new Error("Command failed: #{signal ? code}") if code isnt 0
+    error?.code ?= code
+    error?.stdout ?= stdout
+    callback?(error, stdout)
+
+class AutoUpdater
+  _.extend @prototype, EventEmitter.prototype
+
+  stdout = ChildProcess.execSync('cat /etc/issue').toString().replace(/\\./g, "")
```
50Wliu: I believe Atom uses `lsb_release` to determine the OS - could you do that instead?
https://github.com/atom/notifications/blob/4716ca24f74c685f6db85447e432e8a173d86a67/lib/user-utilities.coffee#L64-L78 (though it does seem a bit broken on Fedora)

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+    @pkgType = 'deb'
+    @installCmd = 'dpkg --install -y'
+  else if distros.redhat.test(stdout) or distros.centos.test(stdout)
+    @pkg = 'atom.x86_64.rpm'
+    @pkgType = 'rpm'
+    @installCmd = 'yum localinstall -y'
+  else if distros.fedora.test(stdout)
+    @pkg = 'atom.x86_64.rpm'
+    @pkgType = 'rpm'
+    @installCmd = 'dnf install -y'
```
50Wliu: Does this work on Fedora 21 and under?

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+    app.quit()
+
+  downloadUpdate: (callback) ->
+    spawn 'curl', ['-L', @updateUrl], (error, stdout) ->
+      return callback(error) if error?
+
+      try
+        update = JSON.parse(stdout)
+
+        spawn 'atom', ['--version'], (error, stdout) ->
```
50Wliu: The command isn't necessarily `atom` anymore with the introduction of beta releases - `atom-beta` is valid now and will launch the beta version which also supports auto-updates.  You'll need to catch that.

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+
+  setFeedUrl: (@updateUrl) ->
+
+  quitAndInstall: ->
+    app.once 'will-quit', -> spawn 'sudo', [installCmd, @pkg, '&&',
+      'rm', '-rf', @pkg, '&&', 'atom']
+    app.quit()
+
+  downloadUpdate: (callback) ->
+    spawn 'curl', ['-L', @updateUrl], (error, stdout) ->
```
50Wliu: Is there any chance at all that `curl` isn't installed?  If so we'll need to add it as a dependency.

src/browser/auto-updater-linux.coffee
```diff
@@ -0,0 +1,115 @@
...
+  else if distros.fedora.test(stdout)
+    @pkg = 'atom.x86_64.rpm'
+    @pkgType = 'rpm'
+    @installCmd = 'dnf install -y'
+  else distro = 'unsupported'
+
+  setFeedUrl: (@updateUrl) ->
+
+  quitAndInstall: ->
+    app.once 'will-quit', -> spawn 'sudo', [installCmd, @pkg, '&&',
```
50Wliu: `sudo` :(
I'm guessing that this can't be avoided though.

50Wliu:
The linux packages are only for 64-bit right now, but I'm fairly certain that 32-bit users can build it manually and run Atom (possibly pinning it to a release version and enabling updates).  If so, that should be addressed in the auto-updater so that it doesn't try installing 64-bit builds on 32-bit machines.

stanier:
... I'll just wait for the official repositories

lee-dohm:
Thanks @stanier for all your hard work on this. We talked this over and we're not comfortable with supporting a partial solution like this. We're still planning on having support for the official repositories. I don't have an ETA on when that will be accomplished right now though.

Thanks again for driving this!

PR #8094, opened 2015-07-27
Specify class attributes directly for overlay nodes

smashwilson:
Specifying a `.class` parameter when creating an overlay decoration now applies that CSS class to the created `atom-overlay` element:

``` coffee
decoration = editor.decorateMarker(marker, {type: 'overlay', item: element, class: 'my-overlay'})
```

Usually, this isn't necessary, because you can apply styles to the item's view instead. I hit a problem with that when I tried to apply positioning styles to the item, though, because the overlay element's absolute positioning interfered with the positioning I was trying to apply. I've had [a workaround](https://github.com/smashwilson/merge-conflicts/blob/1088c7fbcf32924998ff6c88c41f088bf0da37c9/lib/view/covering-view.coffee#L23) in place for a while, but being able to apply styles to the overlay element itself would make it a bit more graceful, I think. It would also make overlay parameters a bit more consistent with those accepted by the other decorations.

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1475,6 +1475,17 @@ describe "TextEditorComponent", ->
         overlay = component.getTopmostDOMNode().querySelector('atom-overlay .overlay-test')
         expect(overlay).toBe null
 
+      it "renders the overlay element with the CSS class specified by the decoration", ->
+        marker = editor.displayBuffer.markBufferRange([[2, 13], [2, 13]], invalidate: 'never')
+        decoration = editor.decorateMarker(marker, {type: 'overlay', class: 'my-overlay', item})
+        nextAnimationFrame()
+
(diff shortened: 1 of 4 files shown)


nathansobo:
This looks good to me. I just kicked Travis again to hopefully get a green build. Not sure what's going on with that failure.

nathansobo:
@benogle I'm okay merging this because it does feel consistent with aspects of the API and provides something that's awkward to do otherwise (add a class to the overlay wrapper). Any objections?

benogle:
Nope, let's do it.

benogle:
:heart: :heart: @smashwilson

smashwilson:
Excellent, thanks! :bow:

I was seeing a few other spec failures in my local `apm test` too, by the way (on master, too). I'll have a deeper look at those when I get the chance.

PR #8098, opened 2015-07-27
Render tiles upper in the stack in front of the ones below

### LINKED ISSUE #8026: Atom does not render underscore characters (_) every few lines
This is Atom 1.0.2 on Linux.

I am not seeing Underscore characters rendered every few lines, which is making the editor dangerous to work with in programming, as I cannot determine the difference between ```foo_bar``` and ```foo bar```.
underscore_missing

This screen shot is taken with Font size 15. The line height is set to 1.1. This happens every 12 lines, but not reliably.

It does not happen with font size 14, but also happens with 16 and above. See this video that shows the effect:
atom_underscore

as-cii:
Fixes #8026.

screen shot 2015-07-27 at 20 34 41

As a temporary stopgap measure until we explore better alternatives, we have decided to tackle the above issue by putting tiles upper in the stack in front of the ones below.

We need to put tiles into a separate div with its own stacking context, so that tiles do not overlap other elements such as the cursors or the wrap-guide.

/cc: @nathansobo @abe33 @rbu

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -88,6 +88,28 @@ describe "TextEditorComponent", ->
       else
         expect(lineNode.textContent).toBe(tokenizedLine.text)
 
+    it "renders tiles upper in the stack in front of the ones below", ->
+      wrapperNode.style.height = 6.5 * lineHeightInPixels + 'px'
+      component.measureDimensions()
+      nextAnimationFrame()
+
+      tilesNodes = componentNode.querySelector(".lines").querySelectorAll(".tile")
+
+      expect(tilesNodes[0].style.zIndex).toBe("2")
+      expect(tilesNodes[1].style.zIndex).toBe("1")
+      expect(tilesNodes[2].style.zIndex).toBe("0")
+
+      verticalScrollbarNode.scrollTop = 1 * lineHeightInPixels
+      verticalScrollbarNode.dispatchEvent(new UIEvent('scroll'))
+      nextAnimationFrame()
+
+      tilesNodes = componentNode.querySelector(".lines").querySelectorAll(".tile")
+
+      expect(tilesNodes[0].style.zIndex).toBe("3")
+      expect(tilesNodes[1].style.zIndex).toBe("2")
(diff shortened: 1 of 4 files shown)


as-cii:
@rbu: it'd be great if you could try this out and check if it actually solves the problem you reported. In the screenshot above it seems like this did the trick, but I'd like to make sure it works on the other platforms as well. Thank you! :pray: :sparkles: (if you don't know how to build an Atom branch, please follow the [build instructions](https://github.com/atom/atom/tree/master/docs/build-instructions) or hit me up on Slack)

nathansobo:
Code looks good to me if it solves the problem.

as-cii:
|                      |                                                                  Timeline                                                                 |
|:--------------------:|:-----------------------------------------------------------------------------------------------------------------------------------------:|
|      **master**      |               master               |
| **as-tiles-z-index** | z-index |

:+1:

dwt:
You might want to inset that unicode oddity in the first line of a new tile.

Out of interest: since this is meant as a stopgap, shouldn't you let the bug #8026 open to track resolution of the bigger issue?

rbu:
@as-cii, thanks for the ping. I could not build Atom yet, as it would require an npm update on my system (which I have no package for) and now you were faster than myself. Anyway, thanks for fixing it, I'll check it out in the next release!

as-cii:
> You might want to inset that unicode oddity in the first line of a new tile. :)

@dwt: yeah, we are aware it's not a super comprehensive solution but we wanted to make sure that a common case such as the underscore was covered in the meantime.

> Out of interest: since this is meant as a stopgap, shouldn't you let the bug #8026 open to track resolution of the bigger issue?

That specific issue should be fixed as of now, but I agree we should probably track the bigger problem in a new issue (linking to #8026 and similar bug reports). Thanks for pointing that out! :bow: 

> I could not build Atom yet, as it would require an npm update on my system (which I have no package for) and now you were faster than myself. Anyway, thanks for fixing it, I'll check it out in the next release!

@rbu: sounds great! Thanks for taking the time to report this problem and to follow its progress! :sparkles:

PR #8103, opened 2015-07-28
Update language-javascript to 0.85.0

winstliu:
Fixes an issue where strings would be incorrectly highlighted in the following scenario:

``` js
write("){");
```

(Also, sorry for the many JS PRs lately - maybe I should try bumping the version less often :P)


benogle:
> (Also, sorry for the many JS PRs lately - maybe I should try bumping the version less often :P)

It's fine

benogle:
Merge when green :+1:

PR #8105, opened 2015-07-28
Emit `did-insert-text` on `pasteText()`

### LINKED ISSUE #8070: TextEditor::pasteText does not emit did-insert-text
`insertText` calls `@emitWillInsertTextEvent(text)` and later `@emitter.emit 'did-insert-text', didInsertEvent`, but `pasteText` only does the former and never the latter.

as-cii:
Fixes #8070

/cc: @atom/feedback @maxbrunsfeld @nathansobo

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -2627,14 +2627,19 @@ class TextEditor extends Model
         if containsNewlines or not cursor.hasPrecedingCharactersOnLine()
           options.indentBasis ?= indentBasis
 
+      range = null
       if fullLine and selection.isEmpty()
         oldPosition = selection.getBufferRange().start
         selection.setBufferRange([[oldPosition.row, 0], [oldPosition.row, 0]])
-        selection.insertText(text, options)
+        range = selection.insertText(text, options)
         newPosition = oldPosition.translate([1, 0])
         selection.setBufferRange([newPosition, newPosition])
       else
-        selection.insertText(text, options)
+        range = selection.insertText(text, options)
+
+      didInsertEvent = {text, range}
+      @emit('did-insert-text', didInsertEvent) if includeDeprecatedAPIs
+      @emitter.emit 'did-insert-text', didInsertEvent
 
   # Public: For each selection, if the selection is empty, cut all characters
   # of the containing line following the cursor. Otherwise cut the selected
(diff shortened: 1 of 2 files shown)


src/text-editor.coffee
```diff
@@ -2627,14 +2627,19 @@ class TextEditor extends Model
...
-        selection.insertText(text, options)
+        range = selection.insertText(text, options)
         newPosition = oldPosition.translate([1, 0])
         selection.setBufferRange([newPosition, newPosition])
       else
-        selection.insertText(text, options)
+        range = selection.insertText(text, options)
+
+      didInsertEvent = {text, range}
+      @emit('did-insert-text', didInsertEvent) if includeDeprecatedAPIs
```
thomasjo: Is the deprecated API change really necessary? This is new functionality, so we wouldn't need to support this?

as-cii:
On a side note, I have noticed there are other places where a `did-insert-text` is not emitted. Do you think we should move this "emitting events" behavior right inside `mutateSelectedText`?

Is there any use case where we call `@mutateSelectedText` but we don't want to emit such events?

thomasjo:
Other than my one comment, this looks good to me.

nathansobo:
> Do you think we should move this "emitting events" behavior right inside mutateSelectedText?

I wonder if emitting of this event should be moved to `Selection`, since text insertion is routed through that location. `mutateSelectedText` just takes a callback so it could be awkward to figure out what to emit in there.

as-cii:
> I wonder if emitting of this event should be moved to Selection, since text insertion is routed through that location.

That makes perfectly sense. And I believe that the [duplication](https://github.com/atom/atom/blob/master/spec/text-editor-spec.coffee#L2066) in [tests](https://github.com/atom/atom/blob/master/spec/text-editor-spec.coffee#L3007) points out exactly towards that direction: a centralized place where to emit those events.

I don't see it as an urgent thing, but I feel like it would increase code quality while reducing the possibility of making mistakes. I'll create an issue so that we can tackle it at some point.

Thanks @nathansobo and @thomasjo! :sparkles:

jacekkopecky:
Yes, emitting of the event should be in Selection, and here's why:

In vim-mode we have an operation to copy the character from the line above: for example when typing a line that starts the same as the previous one, one can just press `ctrl-y` that many times. 

With multiple cursors, this operation naturally takes what's above each cursor; therefore, it needs to do `selection.insertText` and not `editor.insertText` so the editor has no idea the insert has happened and cannot notify whoever is listening to `will-insert-text` and `did-insert-text`.

This bites in vim-mode's *replace mode*, which simulates over-type by deleting a character for every one that is being inserted, as notified by `will-inset-text`.

PR #8107, opened 2015-07-28
Correct path separator in Windows Build instructions

batjko:
Just a little oversight.

### DIFF
--- a/docs/build-instructions/windows.md
+++ b/docs/build-instructions/windows.md
@@ -33,7 +33,7 @@ cd atom
 script/build # Creates application in the `Program Files` directory
 ```
 
-### `script/build` Options
+### `script\build` Options
   * `--install-dir` - Creates the final built application in this directory.
   * `--build-dir` - Build the application in this directory.
   * `--verbose` - Verbose mode. A lot more information output.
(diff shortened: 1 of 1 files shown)


mnquintana:
Nice catch - thanks!

batjko:
:+1:

docs/build-instructions/windows.md
```diff
@@ -33,7 +33,7 @@ cd atom
 script/build # Creates application in the `Program Files` directory
```
izuzak: I noticed you didn't change this instance of `script/build`. Just wondering why that's so.

izuzak:
I can't run my windows VM anymore, so I can't try this out, but I think it was possible to run `script/build` in Git shell. Can anyone give give that a try?

mnquintana:
@izuzak Ah that's a good point, that should work. Should we revert this?

batjko:
I can try it via the git shell... will update shortly.

I wouldn't revert the change but clarification may be necessary, because in command or powershell it still needs to be a backslash. So if it does depend on from where the build is executed, we may have to mention that.

docs/build-instructions/windows.md
batjko: My oversight in this instance. 
I'll open a new PR to correct this, as soon as we are also clear about whether further clarification needs to be added.

batjko:
Confirmed. The shell does take fwd slashes.
So I have added #8109 to partially revert this one and add a note to mention backslashes if the shell isn't used.

PR #8109, opened 2015-07-28
Clarify path separators, partially reverting #8107

batjko:
Thanks @izuzak for pointing out that the git shell can take forward slashes, and the original instructions do assume usage of that, as opposed to cmd or Powershell.

### DIFF
--- a/docs/build-instructions/windows.md
+++ b/docs/build-instructions/windows.md
@@ -32,8 +32,10 @@ git clone https://github.com/atom/atom/
 cd atom
 script/build # Creates application in the `Program Files` directory
 ```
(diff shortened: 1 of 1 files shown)


kevinsawicki:
:+1:

izuzak:
Sorry for missing this yesterday, @batjko -- I'm switching from email notification to web notifications in GitHub and I'm still trying to find the best workflow to handle all the threads. :sweat_smile: 

Thanks for this!

batjko:
No worries at all, it's not like this is a particular time-critical change, but @kevinsawicki picked it up quickly anyway. :wink:

I tend to extensively use the Notifications screen and just work my way through that always-way-too-large list.

PR #8110, opened 2015-07-28
Only build pushes to master on Travis

mnquintana:
This PR adds some Travis config to only trigger builds for pushes to `master`, instead of building pushes for every branch in the repo. This should help keep our jobs queue from being flooded - it should cut down our number of build jobs from 6 to 3. However, this does preserve triggering builds for PRs.

I think we should probably add this config to every repo in the Atom org.

/cc @as-cii @atom/feedback

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 git:
   depth: 10
 
+branches:
+  only:
+    - master
+
 env:
   global:
     - ATOM_ACCESS_TOKEN=da809a6077bb1b0aa7c5623f7b2d5f1fec2faae4


as-cii:
I think we can safely proceed to merge this, as there will almost always be a PR for non-master branches.

:shipit: :green_heart: :sparkles:

thomasjo:
> I think we can safely proceed to merge this, as there will almost always be a PR for non-master branches.

Agreed. :boat:

maxbrunsfeld:
Great idea.

> I think we should probably add this config to every repo in the Atom org.

:+1:

mnquintana:
> I think we should probably add this config to every repo in the Atom org.

Alright, added this config to every repo I have push access to. Hopefully this'll help keep our build queue clear :green_heart:

PR #8112, opened 2015-07-28
Create new stacking context for tiles

as-cii:
Refs: #8098

To create a new stacking context we need to isolate the container div (https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Understanding_z_index/The_stacking_context).

This fixes a visual glitch on the wrap-guide whenever there are more than 5 tiles (i.e. the wrap guide gets hidden). I am sorry for not catching this early, but apparently I overlooked it while testing #8098.

/cc: @nathansobo

### DIFF
--- a/src/lines-component.coffee
+++ b/src/lines-component.coffee
@@ -14,6 +14,9 @@ class LinesComponent extends TiledComponent
     @domNode = document.createElement('div')
     @domNode.classList.add('lines')
     @tilesNode = document.createElement("div")
+    # Create a new stacking context, so that tiles z-index does not interfere
+    # with other visual elements.
+    @tilesNode.style.isolation = "isolate"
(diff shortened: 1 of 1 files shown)


nathansobo:
Wow. I have never seen that CSS directive.

as-cii:
> Wow. I have never seen that CSS directive.

Yeah, I had never used it as well but apparently its job is exactly to create a new stacking context. From https://developer.mozilla.org/en-US/docs/Web/CSS/isolation:

> The isolation CSS property defines if the element must create a new stacking context.

`position: absolute` could have done the trick as well, but it feels more like an hack given that we don't really care about that `<div>` positioning.

nathansobo:
:+1: Merge at will.

PR #8120, opened 2015-07-28
Use `Returns` keyword to highlight vals in docs

ssorallen:
Use the `Returns` keyword in `PackageManager::getAvailablePackage*`
docs so the return values are called out in panels like with other
documentation.


ssorallen:
This follows a similar change to the `Notification` docs that use "Public: Returns" in the same line: https://github.com/atom/atom/pull/7879

kevinsawicki:
:+1: :memo: thanks

PR #8121, opened 2015-07-28
Use `Returns` keyword in `ThemeManager` public docs

ssorallen:
Use the `Returns` keyword in `get*Theme` function docs so they are put
in panels like other return values in docs.


kevinsawicki:
:chart_with_upwards_trend:

PR #8122, opened 2015-07-28
Use `Returns` keyword in global `atom` object docs

ssorallen:
Use the `Returns` keyword in function docs so they are put
in panels like other return values in docs.


kevinsawicki:
:fireworks:

PR #8123, opened 2015-07-28
Don’t autoscroll to selection when double- or triple-click dragging

### LINKED ISSUE #6675: scrolling behavior while mouse-selecting whole lines of text is hard to control
Atom v.0.196.0
Mac OS X 10.10.3

Trying to select text that spans more than just the visible scroll region is much more difficult in Atom than in other editors when selecting entire lines because of the auto-scrolling behavior.  When you stretch a selection past the edge of the scroll region, the editor scrolls to accommodate you, but once you overshoot your destination, pulling back starts scrolling back in the other direction even though it's not necessary.  Most other editors will only scroll when necessary.

Steps:
1. open a file that is at least twice as long as what you can display in your window
2. make sure you're scrolled to the top
3. triple-click and hold somewhere to begin selecting entire lines
4. drag down just past the bottom of the window, causing the window to scroll
5. once you've scrolled all the way to the bottom (it's important here that you scroll at least past the point where the top of the selection scrolls out of the window), pretend you overshot the bottom of what you wanted selected by a few lines and you actually want the selection to end at the 10th line from the bottom of the file.  continue to hold the mouse button down while you move your mouse back up to that line.
6. observe that if you move your mouse up past about the 4th line from the bottom of the window, it actually starts scrolling back up and will keep going until the top of your selection is visible.  This makes it difficult to get the selection to stop on a specific line because you have to time the scrolling to get your target line within 4 lines of the bottom of the window.

Expected: after mouse movement downward causes a scroll, moving the mouse back up into the window should not cause scrolling back in the other direction unless you attempt to move the end of the selection past the top of the visible part of the file.

nathansobo:
Instead just autoscroll to the cursor, which is what we do when single-click dragging.

Fixes #6675

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1807,6 +1807,58 @@ describe "TextEditorComponent", ->
           expect(window.removeEventListener).toHaveBeenCalledWith('mouseup')
           expect(window.removeEventListener).toHaveBeenCalledWith('mousemove')
 
+    describe "when the mouse is double-clicked and dragged", ->
+      it "expands the selection over the nearest word as the cursor moves", ->
+        jasmine.attachToDOM(wrapperNode)
+        wrapperNode.style.height =  6 * lineHeightInPixels + 'px'
+        component.measureDimensions()
+        nextAnimationFrame()
+
+        linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([5, 10]), detail: 1))
+        linesNode.dispatchEvent(buildMouseEvent('mouseup'))
+        linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([5, 10]), detail: 2))
+        expect(editor.getSelectedScreenRange()).toEqual [[5, 6], [5, 13]]
+
+        linesNode.dispatchEvent(buildMouseEvent('mousemove', clientCoordinatesForScreenPosition([11, 11]), which: 1))
+        nextAnimationFrame()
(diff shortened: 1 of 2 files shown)


izuzak:
@nathansobo Thanks for fixing this. :zap:

I'm seeing different behavior for single-click drag-to-scroll, double-click drag-to-scroll, and triple-click drag-to-scroll. 

When I single-click (I'm testing all of these on text-editor.coffee):

1click

When I double-click:

2click

When I triple-click:

3click

Differences:
* when I single-click or triple-click and then drag down -- scrolling is super fast. When I double-click and then drag down -- scrolling is slow and choppy. 
* when I single-click, then scroll down, and then scroll all the way up and move the mouse over the the top edge of the window, and then try to scroll down -- scrolling continues to work. When I double-click or triple-click and then do the same -- I can no longer scroll down after scrolling all the way up and move the mouse over the top edge.

Are you seeing the same behavior? It's possible that this PR didn't cause these behaviors, but just wanted to mention it since I was testing this and noticed it. If you see the same and think it's worth fixing -- I'd be happy to create issues for them.

nathansobo:
@izuzak Good catch. I actually agree with you that's it's seeming choppier than it was. I'll investigate today.

nathansobo:
@izuzak Fixed the autoscroll problems in #8141. Looking into the choppiness now.

nathansobo:
@izuzak Fixed choppiness in #8146.

Thanks so much for pointing out these issues. Let me know if I missed anything.

izuzak:
Yessss! Thanks for the fixes, @nathansobo -- things look great now.  :heart_eyes:

maxbrunsfeld:
:clap:

PR #8125, opened 2015-07-29
Update language-go and yaml

winstliu:
PRing because two languages changed.

language-go@0.32.0 improves the auto-indent patterns and also adds specs for it :green_heart:!
language-yaml@0.23.0 adds support for `.sls` files and properly handles pound (`#`) symbols in multiline blocks.

Will merge when green.


50Wliu:
The only failing spec comes from `symbols-view`, which is also failing intermittently on master and on the `symbols-view` repo.

PR #8126, opened 2015-07-29
Fix process leak

benogle:
The process for each project search would stick around:

screen shot 2015-07-28 at 2 11 12 pm

Closes https://github.com/atom/find-and-replace/issues/459

### DIFF
--- a/src/default-directory-searcher.coffee
+++ b/src/default-directory-searcher.coffee
@@ -18,7 +18,9 @@ class DirectorySearch
     @task.on 'scan:paths-searched', options.didSearchPaths
     @promise = new Promise (resolve, reject) =>
       @task.on('task:cancelled', reject)
-      @task.start(rootPaths, regex.source, scanHandlerOptions, resolve)
+      @task.start rootPaths, regex.source, scanHandlerOptions, =>
+        @task.terminate()
+        resolve()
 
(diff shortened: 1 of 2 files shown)


kevinsawicki:
:syringe: :heart: awesome, maybe 1.0.4 tomorrow?

benogle:
Yeah, this sucks

maxbrunsfeld:
Ah, so stoked you found it!

src/default-directory-searcher.coffee
```diff
@@ -89,6 +91,8 @@ class DefaultDirectorySearcher
           reject()
     return {
       then: promise.then.bind(promise)
+      catch: promise.catch.bind(promise)
+      directorySearch: directorySearch
```
maxbrunsfeld: I think it'd be kinda nice to avoid adding the `directorySearch` to the return value. It's because you need to spy on `terminate()` for the specs, right? What do you think about using a spy on `Task.prototype` instead? Maybe just:

```coffee
spyOn(Task::, 'start').andCallThrough()
spyOn(Task::, 'terminate').andCallThrough()

# ...

expect(Task::start.callCount).toBe(1)
expect(Task::terminate.callCount).toBe(1)
```

benogle: Yeah spying on the proto makes more sense

PR #8130, opened 2015-07-29
Add support for activating and deactivating package-specific keymaps

tmunro:
Adds support for activating and deactivating the keymaps for individual packages. Currently, there's no good way to quickly disable conflicting keymaps. Adding `unset!` entries for every binding in your own keymap is tedious and error-prone. This adds `Package` support for activating and deactivating keymaps, so you can easily choose which keymaps from which packages are important to you.

See https://github.com/atom/settings-view/pull/610 for UI support.

Related: https://github.com/atom/atom-keymap/issues/82

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -438,6 +438,38 @@ describe "PackageManager", ->
           runs ->
             expect(atom.packages.isPackageActive("package-with-empty-keymap")).toBe true
 
+      describe "when the package's keymaps have been disabled", ->
+        it "does not add the keymaps", ->
+          element1 = $$ -> @div class: 'test-1'
+
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+
+          atom.config.set("core.disabledKeymaps", ["package-with-keymaps-manifest"])
+
+          waitsForPromise ->
+            atom.packages.activatePackage("package-with-keymaps-manifest")
+
+
+          runs ->
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+
+      describe "when the package's keymaps are disabled at activation and re-enabled later", ->
+        it "re-adds the keymaps", ->
+          element1 = $$ -> @div class: 'test-1'
+
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+
+          atom.config.set("core.disabledKeymaps", ["package-with-keymaps-manifest"])
+
+          waitsForPromise ->
+            atom.packages.activatePackage("package-with-keymaps-manifest")
+
+          runs ->
+            atom.config.set("core.disabledKeymaps", [])
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])[0].command).toBe 'keymap-1'
+
+
+
     describe "menu loading", ->
       beforeEach ->
         atom.contextMenu.definitions = []
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -452,18 +452,17 @@ describe "PackageManager", ->
           runs ->
             expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
 
-      describe "when the package's keymaps are disabled at activation and re-enabled later", ->
-        it "re-adds the keymaps", ->
+      describe "when the package's keymaps are disabled and re-enabled after it is activated", ->
+        it "removes and re-adds the keymaps", ->
           element1 = $$ -> @div class: 'test-1'
-
-          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
-
-          atom.config.set("core.disabledKeymaps", ["package-with-keymaps-manifest"])
 
           waitsForPromise ->
             atom.packages.activatePackage("package-with-keymaps-manifest")
 
           runs ->
+            atom.config.set("core.disabledKeymaps", ['package-with-keymaps-manifest'])
+            expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+
             atom.config.set("core.disabledKeymaps", [])
             expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])[0].command).toBe 'keymap-1'
 
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -449,7 +449,6 @@ describe "PackageManager", ->
           waitsForPromise ->
             atom.packages.activatePackage("package-with-keymaps-manifest")
 
-
           runs ->
             expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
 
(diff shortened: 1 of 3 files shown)


spec/package-manager-spec.coffee
```diff
@@ -438,6 +438,38 @@ describe "PackageManager", ->
...
+      describe "when the package's keymaps have been disabled", ->
+        it "does not add the keymaps", ->
+          element1 = $$ -> @div class: 'test-1'
+
+          expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0
+
+          atom.config.set("core.disabledKeymaps", ["package-with-keymaps-manifest"])
+
+          waitsForPromise ->
+            atom.packages.activatePackage("package-with-keymaps-manifest")
```
bronson: Extra newline

bronson:
Great feature!  You thinking about adding UI in another PR or is this good enough?

tmunro:
Yes! UI support is in https://github.com/atom/settings-view/pull/610, forgot to put a link in when I added the other PR.

spec/package-manager-spec.coffee
tmunro: Fixed!

maxbrunsfeld:
This feature makes a lot of sense. Thanks for taking the time to add it!

Right now, I can't tell if it works to add a package to the `disabledKeymaps` *after* it has been activated. Could you make sure that is tested? I would suggest just modifying your second test, to something like this:

```coffee
describe "when a package's keymaps are disabled and re-enabled after it is activated", ->
  it "removes and re-adds the keymaps", ->
    element1 = $$ -> @div class: 'test-1'

    waitsForPromise ->
      atom.packages.activatePackage("package-with-keymaps-manifest")

    runs ->
      atom.config.set("core.disabledKeymaps", ['package-with-keymaps-manifest'])
      expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])).toHaveLength 0

      atom.config.set("core.disabledKeymaps", [])
      expect(atom.keymaps.findKeyBindings(keystrokes: 'ctrl-z', target: element1[0])[0].command).toBe 'keymap-1'

    
```

tmunro:
@maxbrunsfeld Good point, I knew I was missing something in the specs. Added.

src/package.coffee
```diff
@@ -200,7 +200,13 @@ class Package
...
   activateResources: ->
     @activationDisposables = new CompositeDisposable
-    @activationDisposables.add(atom.keymaps.add(keymapPath, map)) for [keymapPath, map] in @keymaps
+
+    @activationDisposables.add atom.config.observe "core.disabledKeymaps", {}, (map) =>
+      value = not _.include(map ? [], @name)
+      if value
+        @activateKeymaps()
+      else if not value
+        @deactivateKeymaps()
```
maxbrunsfeld: It would be nice to do this config-observing in the `PackageManager`. Performance-wise, it would mean there'd be one observer instead of one for each package, and it would also be more consistent with [the way we observe `disabledPackages`](https://github.com/atom/atom/blob/1d9ecc1d597983c0ac88b3c7f9df1bde3f798c3e/src/package-manager.coffee#L305). I think the code could be very similar to the code for `observeDisabledPackages`.

Would you mind refactoring this code in that way?

tmunro: Sure, that makes more sense, thanks for the suggestion.

maxbrunsfeld: > Would you suggest splitting it into onDidChange for the observer and an initial config check for activation?

I don't have a strong opinion on that; go with whichever way makes the most sense as you're writing the code.

src/package.coffee
```diff
@@ -200,7 +200,12 @@ class Package
 
   activateResources: ->
     @activationDisposables = new CompositeDisposable
-    @activationDisposables.add(atom.keymaps.add(keymapPath, map)) for [keymapPath, map] in @keymaps
+
+    value = not _.include(atom.config.get("core.disabledKeymaps") ? [], @name)
+    if value
+      @activateKeymaps()
+    else if not value
+      @deactivateKeymaps()
```
maxbrunsfeld: You can remove the second `if`, since its condition is the inverse of the first `if`. How about this:

```coffee
keymapIsDisabled = _.include(atom.config.get("core.disabledKeymaps") ? [], @name)
if keymapIsDisabled
  @deactivateKeymaps()
else
  @activateKeymaps()
```

src/package.coffee
```diff
@@ -232,6 +237,31 @@ class Package
...
+    @keymapDisposables?.dispose()
+    atom.menu.update()
+
+    @keymapActivated = false
+
+  hasKeymaps: ->
+    for [path, map] in @keymaps
+      if map.length > 0
+        return true
+
```
maxbrunsfeld: Could you :fire: this blank line, since the method's so short?

maxbrunsfeld:
I left some very minor additional feedback, and restarted the travis build because it failed for reasons unrelated to your change. Looks good!

maxbrunsfeld:
Alright! :ship: :money_with_wings:

t9md:
Thanks @tmunro and nice work.

I want to put comment for name `core.disabledKeymaps`.
If we add another keymap disabling feature for specific keystroke(keystroke based disabling).
The name `core.disabledKeymaps` become ambiguous.

How about rename `core.disabledKeymaps` to `core.keymapDisabledPackages` its also well sync current `core.disabledPackages`.

maxbrunsfeld:
> If we add another keymap disabling feature for specific keystroke(keystroke based disabling.

If you wanted to disable a specific keystroke, couldn't you just `unset!` in your own keymap?

bronson:
> If you wanted to disable a specific keystroke, couldn't you just unset! in your own keymap?

Yes, if you already have some Atom expertise.  Even then, figuring out the right selectors gets tedious.

If it gets @t9md to start including more bindings in his packages then it's totally worth it.  ;-)

t9md:
I'm not so sure but I want some easy to use UI in settings-view to disable specific keymap by searching keystroke from Keybindings left menu on settings-view.

But anyway, even if we don't introduce keystroke based disabling UI,  I want to vote `core.keymapDisabledPackages` since its store package list.

maxbrunsfeld:
> I want to vote `core.keymapDisabledPackages` since its store package list.

Yeah, that's a good point @t9md. It would be good to include `packages` in the name, since the value is a list of package names.

How about `core.packagesWithDisabledKeymaps`?

/cc @atom/feedback

t9md:
I like `core.packagesWithKeymapDisabled' if it goes `with` version ;)
But I think `core.keymapDisabledPackages` is simple and clear enough.

bronson:
`core.packagesWithoutKeymaps`?  Naming things is tough.

t9md:
Not 'without', since its originally 'with' keymap and disabled intentionally by user, so keyword 'disable' should be included to name..  but I agree naming is tough and but very important.
I frequently hit the situation that I want to rename after release.

bronson:
`core.packagesNotKeymaps`  :)

Gotta say, I think `core.packagesWithDisabledKeymaps` is the best so far.  Readable and unambiguous. `core.keymapDisabledPackages` is shorter but awkward.  It's trying to make keymap-disabled a verb.

maxbrunsfeld:
Maybe `packagesWithKeymapsDisabled` makes it a bit more clear that *all* of the package's keymaps are disabled (not just some).

t9md:
Maybe reason for I don't think keymapdisabkedPackages is because I'm not English native. So I will agree as long as keyword packages and disable included. :smile:

bronson:
`packagesWithKeymapsDisabled` :+1: It sounds a little strange but it's more descriptive.

t9md:
:+1:

PR #8142, opened 2015-07-29
Upgrade Squirrel for Windows

kevinsawicki:
Long time coming, upgrades to Squirrel for Windows 0.99.3 via upgrading the grunt plugin for creating electron installers.

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -23,7 +23,7 @@ module.exports = (grunt) ->
   grunt.loadNpmTasks('grunt-contrib-less')
   grunt.loadNpmTasks('grunt-shell')
   grunt.loadNpmTasks('grunt-download-atom-shell')
-  grunt.loadNpmTasks('grunt-atom-shell-installer')
+  grunt.loadNpmTasks('grunt-electron-installer')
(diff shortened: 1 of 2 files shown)


paulcbetts:
:+1: :sparkles:

kevinsawicki:
@paulcbetts there is nothing special we need to do for migrations right? Squirrel should handle this upgrade like others right?

I only ask based on the bit changelog of things since we last upgraded, thanks.

paulcbetts:
@kevinsawicki The only thing I can think of is when we added back in BSDiff, you should test an upgrade first. Build the app, install the public version then:

```sh
cd \Users\NAME\AppData\Local\Atom
update.exe --update \path\to\the\atom-build\Installers
```

If it works, you're :gem:

kevinsawicki:
>  you should test an upgrade first.

Followed your instructions and was able to go from official 1.0.3 to the 1.0.4 draft release, thanks for your help :+1:

maaarghk:
I have been on 1.0.3 for a while and turns out 1.0.5 has been downloaded on my system. Atom.sh / atom.cmd were not updated during this download and the "Check for updates" menu said 1.0.3 was the latest version? Related? Windows 10, FYI.

PR #8144, opened 2015-07-29
Allow custom association between file names and grammars

### LINKED ISSUE #1718: Make file type -> language mapping locally configurable
Several halp issues about people wanting to add extensions to language plugins.

Currently they have to fork the package to do this, might be nice to support this through the config.

maxbrunsfeld:
Closes #1718


benogle:
Cool! People will be really happy with this.

maxbrunsfeld:
Right now, the config key is `core.fileTypesByScope`, and works like so:

```coffee
'*':
  core:
    fileTypesByScope:
      'source.ruby': ['Cheffile', '.also-ruby']
      'source.cpp': ['h']
```

What do you all think @atom/feedback?

thomasjo:
> Right now, the config key is `core.fileTypesByScope` [..]

Not falling in love with the name, but other than that this is awesome. The name might grow on me.

src/config.coffee
```diff
@@ -665,12 +665,23 @@ class Config
...
-      schema = schema.properties?[key]
+      if schema.type is 'object'
+        childSchema = schema.properties?[key]
+        unless childSchema?
+          if isPlainObject(schema.additionalProperties)
+            childSchema = schema.additionalProperties
+          else if schema.additionalProperties is false
+            return false
+          else
+            return null
```
as-cii: I am not 100% familiar with this code but I am wondering if we could avoid `null` and/or `false` as this function's return values. Without looking at the documentation, the calling code seems a bit obscure to me.

I am probably ok with keeping `null`, as it sounds reasonable that a schema cannot be found. But can we encode `false` differently? Can we tell what that value represents without indicating it in the documentation?

Maybe we could make the return value consistent across the various `return` statements, and return an object that always responds to the following messages:

* `hasSchemaSpecified`
* `isAccessibleFromRootSchema`

@maxbrunsfeld: what do you think?

as-cii:
One comment, otherwise it looks good to me! Great job, @maxbrunsfeld! :sparkles:

maxbrunsfeld:
> Not falling in love with the name

Yeah, I feel that. `fileTypesByScope` communicates the structure of the object pretty exactly, but maybe it would be better to give it a simpler name like `customFiletypes` and leave it up to the description to describe the name.

src/config.coffee
maxbrunsfeld: I agree that the `null/false` thing is weird. But I feel like there's something intuitive about  `::getSchema` returning the same thing that was passed to `::setSchema`, as opposed to an object with additional structure. Maybe we could think of another way to encode the difference between "not accessible" and "whatever value you want" though.

How about we change `null` to represent "not accessible", and in the case where any value is allowed (e.g. `atom.config.getSchema("editor.unknownKey")`), we return a new schema like `{type: "any"}` that corresponds to no validation whatsoever.

benogle:
> Right now, the config key is core.fileTypesByScope, and works like so:

Any reason these couldnt be a minimatch pattern. Maybe overkill, but could be a little more flexible:

```coffee
'*':
  core:
    fileTypesByScope:
      'source.ruby': ['Cheffile', '*.this-is-also-ruby']
      'source.cpp': ['*.h']
      'source.js': ['i-am-crazy*.*']
```

benogle:
For the record, I dig the name for its clarity.

maxbrunsfeld:
> Any reason these couldn't be a minimatch pattern.

I wanted to make them use the same system as the `grammars` use for matching files ([exact suffix matches](https://github.com/atom/language-coffee-script/blob/master/grammars/coffeescript.cson#L3)), so it would be easiest to understand which matches take precedence. Right now, longer matches win, and ties are broken by choosing the users' custom filetypes.

benogle:
> I wanted to make them use the same system as packages' grammars use

:+1:

src/config.coffee
as-cii: > Another thing we could do is change null to represent "not accessible", and in the case where any value is allowed (e.g. atom.config.getSchema("editor.unknownKey")), we return a new schema like {type: "any"} that corresponds to no validation whatsoever.

:+1:

maxbrunsfeld:
Ok, I'm leaning toward renaming it from `fileTypesByScope` to `customFiletypes`. Anybody care?

nathansobo:
I like `customFileTypes`... less precise but more descriptive somehow.

maxbrunsfeld:
Ok, I think this is ready. Anywhere else I should document this new setting?

kevinsawicki:
nice ship :ship: :sunglasses:

rubidot:
For anyone looking for how to use this featured: [Customizing Language Recognition](http://flight-manual.atom.io/using-atom/sections/basic-customization/#customizing-language-recognition)

PR #8182, opened 2015-07-31
Safeguard the caching of Babel transpilations by disallowing .babelrc…

bolinfest:
… files.

Also update the version of Babel to pick up a JSON-parsing fix. This was an
issue that was discovered in the course of seeing how reading .babelrc files
would interact badly with Atom's Babel cache.

/cc @sebmck

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "async": "0.2.6",
     "atom-keymap": "^5.1.8",
     "atom-space-pen-views": "^2.0.4",
-    "babel-core": "^5.8.3",
+    "babel-core": "^5.8.20",
     "bootstrap": "^3.3.4",
     "clear-cut": "^2.0.1",
     "coffee-cash": "0.8.0",
--- a/src/babel.coffee
+++ b/src/babel.coffee
@@ -15,6 +15,18 @@ stats =
   misses: 0
 
 defaultOptions =
+  # Currently, the cache key is a function of:
+  # * The version of Babel used to transpile the .js file.
+  # * The contents of this defaultOptions object.
+  # * The contents of the .js file.
+  # That means that we cannot allow information from an unknown source
+  # to affect the cache key for the output of transpilation, which means
+  # we cannot allow users to override these default options via a .babelrc
+  # file, because the contents of that .babelrc file will not make it into
+  # the cache key. It would be great to support .babelrc files once we
+  # have a way to do so that is safe with respect to caching.
+  breakConfig: true
+
   # The Chrome dev tools will show the original version of the file
   # when the source map is inlined.
   sourceMap: 'inline'


sebmck:
Yeah, this seems like an appropriate patch. Although I'm not sure how widespread the usage of Babel is in the atom community but this would be a breaking change for those relying on `.babelrc` files. In the next major all the transformers are going to be optional which means that specifying Babel configuration is going to be critical, not sure if there's some way to specify that currently to the Atom compiler.

thomasjo:
There's not that much usage of Babel yet, and I doubt many (if any) use `.babelrc` files. I can grab all packages and check. If there's no usage of it, I'll :ship: this if that's OK with you @bolinfest?

bolinfest:
Sounds great -- thanks @thomasjo!

sebmck:
@thomasjo Sounds good to me. I don't really mind what happens, just flagging it just in case it's an issue :blush:

thomasjo:
Here's the list of packages that has a `.babelrc` file in the root of their repo

```
format
hyperclick
nuclide-arcanist
nuclide-code-format
nuclide-diagnostics-store
nuclide-diagnostics-ui
nuclide-diff-view
nuclide-file-tree
nuclide-file-watcher
nuclide-find-references
nuclide-flow
nuclide-hack
nuclide-hg-repository
nuclide-installer
nuclide-language-hack
nuclide-ocaml
nuclide-quick-open
nuclide-remote-projects
nuclide-type-hint
zephir-ide
```

@bolinfest I guess we can ignore the nuclide packages from that list?

bolinfest:
@thomasjo Yes, all of the `nuclide-` and `hyperclick` since that's us, too.

thomasjo:
I don't know that much about Babel (yet), so I'm listing the contents of the two remaining `.babelrc` files;
(1) zephir-ide:
```json
{
  "stage": 0,
  "loose": true,
  "sourceMap": "inline",
  "blacklist": [
    "es6.classes"
  ]
}
```

(2) format:
```json
{
  "stage": 2,
  "loose": true
}
```

Would it be problematic for those packages if we ship this PR?

sebmck:
They're very likely using experimental syntax that is enabled via the `stage` option which would break them.

thomasjo:
> They're very likely using experimental syntax that is enabled via the stage option which would break them.

We are using `stage: 0` by default though. Looks like the only problem for zephir-ide is the blacklisting of `es6.classes` (unless I'm misunderstanding the implication of `stage`).

bolinfest:
Hold up @sebmck, here are the options that are used in Atom today:

https://github.com/atom/atom/blob/master/src/babel.coffee

So for `zephir-ide`:

* We're already using `stage: 0`.
* I'm not sure about `loose: true` since the docs say it should take an Array: https://babeljs.io/docs/usage/options/.
* We're already using `sourceMap: 'inline'`.
* The different blacklist just means that the generated code will be slightly larger. If ES6 classes are working natively in the version of io.js that is bundled with Atom, then we should include it in the blacklist in https://github.com/atom/atom/blob/master/src/babel.coffee. (I was the one who recently added `es6.forOf` to that list.)

For  `format`
* `stage: 0` means that `stage: 2` should be supported
* Again, I don't know about `loose`.

sebmck:
Ah my bad, I talked out of turn. Those other options are fine then.

> ES6 classes are working natively in the version of io.js that is bundled with Atom, then we should include it in the blacklis

This will break decorators and class properties which rely on the class desugaring.

bolinfest:
@sebmck Ah, good to know about `es6.classes`. I guess we'll hold off on that then.

@thomasjo It sounds like we're good to go! Without this fix, any file that is bundled in `Atom.app` with `'use babel'` will fail spectacularly and hilariously. (The problem won't manifest in development, however.) This is why we had to introduce https://github.com/facebook/nuclide/commit/6bbe3f440cb0d13dd3815f5c135e8562aab5b6c0 as a stopgap.

Issue #3719, opened 2014-10-03
Add new editor.tabType scoped setting


benogle:
While digging into scoping `editor.softTabs` in https://github.com/atom/atom/pull/3718, it's pretty clear that the setting has very little influence on the actual value chosen. It's effectively the worst-case fallback: it only works on an empty file.

I propose doing a couple things:

* adding a new `editor.tabType` setting that accepts 3 values: Auto, Hard, Soft. With the default being `auto`. Then in the scoped settings for each language, they can set it to whatever they want. 
* Changing `TextEditor::setSoftTabs` to being a settings override accepting 4 values: unset / use-the-settings, Auto, Hard, Soft
* Reading the settings on each time we need to get the tab type. Something like (pseudo code): 

```coffee
getTabType: (scopeDescriptor) ->
  @userTabTypeOverride or atom.config.get(scopeDescriptor, 'editor.tabType') or atom.config.get(scopeDescriptor, 'editor.softTabs') or 'soft'
```

See https://github.com/atom/atom/blob/4231d6942101c3e77a1d0371cf3b61045b8fa6fb/src/text-editor.coffee#L93 and https://github.com/atom/atom/blob/4231d6942101c3e77a1d0371cf3b61045b8fa6fb/src/text-editor.coffee#L2319 for reference.

Refs 
* https://github.com/atom/atom/issues/3584 

Comment if you know of more issues, complaints. I know there have been several complaints about this, but cant find issues.

jplatte:
+1, I would really appreciate a fix!

TeckniX:
+1 on getting this actually working properly, since certain file extensions do not respect the tab/spacing rules

lexicalunit:
:+1: This would be neato!

remillc:
+1 This soft tab issue is pretty annoying...

NioTeX:
:+1: We need this

ScottPeterJohnson:
:+1:  Really hard to work in indentation-sensitive languages when suddenly Atom inserts spaces instead of tabs.

alexandernst:
Not really sure how such a critical issue didn't make it for 1.0. Shame on Atom for still not being able to manage spaces/tabs

xaviergmail:
:+1: I find myself having to replace all tabs with spaces and re-open a file to get tab to insert spaces properly, which isn't very intuitive.o

JeffreyVdb:
[...]

PR #8186, opened 2015-08-01
Add settings to enable and disable softTabs autodetection

Abdillah:
This solved the issue (partly), discussed on #3719.

I create new config item to specify tabType used by default. But, it clearly conflicting with softTabs config item. I didn't remove it, even I think it's safe to be removed, is it?

tabtype

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -3753,16 +3753,38 @@ describe "TextEditor", ->
   describe "soft-tabs detection", ->
     it "assigns soft / hard tabs based on the contents of the buffer, or uses the default if unknown", ->
       waitsForPromise ->
-        atom.workspace.open('sample.js', softTabs: false).then (editor) ->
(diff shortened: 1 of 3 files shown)


mnquintana:
Thanks for contributing! [This seems to break a lot of specs](https://travis-ci.org/atom/atom/builds/73663507) – you'll need to fix those before this can be considered for merging.

Abdillah:
Ok, I don't aware of that tailing spec test and neither I have test it myself.
Thanks for pointing.

Abdillah:
__Checkpoint One__
Up to this point is test-error free, at least in my computer. 
If travis say so, I'm ready for review..

benogle:
I've created a new PR to add this setting at #8356. This PR is trying to do too much (new API methods, setting config in the editor), and it is lacking specs. I have used a couple commits from this PR in the new one.

Abdillah:
@benogle Great! Thanks.
I will closes this, as I intent the same things as #8356.

PR #8189, opened 2015-08-01
Add text color to atom-workspace

simurai:
Currently the `atom-workspace` background-color uses `@app-background-color`, but text color isn't defined and falls back to Bootstrap's default color of `#333`. This can make text hard to read with dark themes.

This PR adds `color: @text-color;` which should match the background.


thomasjo:
Looks good to me! Let's :shipit:

PR #8208, opened 2015-08-04
Defer selection merge until dragging completes

### LINKED ISSUE #3489: Selection merging doesn't behave as expected sometimes
It seems that selection merging gives different results based on how a selection was made exactly. Here are two scenarios which give different results, even though the selections involved are the same.

First, I double click to select a word, then cmd-click-draw to create another selection and merge it with the first one:

1se

Notice that the selection which survives the merging is the one I created first.

Second, I click-draw to select a word, then cmd-click-draw to create another selection and merge it with the first one:

2se

Notice that the selection which survives the merging is the one I created second.

In the first case, I expected that the selection which I was currently dragging would survive the merging (as in the second case).

cc @benogle and @nathansobo (who was recently looking at how merging works)

nathansobo:
Fixes #3489

This PR changes our approach to creating and manipulating multiple selections with the mouse. Previously, as soon as selections started overlapping we merged them, but this led to unintuitive behavior when dragging with the mouse, where existing selections would snap to the current pointer location. It also created inconsistencies when the first selection was wordwise or linewise.

In this PR, we defer selection merging during dragging. You can drag a new selection to overlap an existing one, then change your mind and still leave the old selection intact. As soon as you release the mouse button, selections are merged as normal.

selection-merging

This is the first time we've ever allowed overlapping selections to exist, so I do wonder if there might be some fallout, but it's a very narrow window. @izuzak interested in giving this a spin since you originally brought this issue to my attention?

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -1948,10 +1948,11 @@ class TextEditor extends Model
   # This method may merge selections that end up intesecting.
   #
   # * `position` An instance of {Point}, with a given `row` and `column`.
-  selectToScreenPosition: (position) ->
+  selectToScreenPosition: (position, suppressMerge) ->
     lastSelection = @getLastSelection()
     lastSelection.selectToScreenPosition(position)
-    @mergeIntersectingSelections(reversed: lastSelection.isReversed())
+    unless suppressMerge
+      @mergeIntersectingSelections(reversed: lastSelection.isReversed())
 
   # Essential: Move the cursor of each selection one character upward while
   # preserving the selection's tail position.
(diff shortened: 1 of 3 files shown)


thomasjo:
:+1: Looks good.

izuzak:
@nathansobo Just took this branch for a test drive and couldn't trigger any problems with the new approach -- seems to work well! :zap: Also, I think I like this approach better than the last one where selections were merged while dragging -- really neat idea. :bulb: 

Should we use the same approach for gutter selections as well?

src/text-editor.coffee
```diff
@@ -1948,10 +1948,11 @@ class TextEditor extends Model
   # This method may merge selections that end up intesecting.
   #
   # * `position` An instance of {Point}, with a given `row` and `column`.
```
mnquintana: Should we document the new `suppressMerge` option here?

nathansobo: I'm not sure I want to make it part of the public API, since it leaves us in a weird state where selections overlap.

nathansobo:
> Should we use the same approach for gutter selections as well?

Good thought. I'll add that before merging this.

izuzak:
:sparkle:

PR #8225, opened 2015-08-05
Fix searching explicitly in directories that contain dots

### LINKED ISSUE #309: Unexpected search results from "Search in directory"

### LINKED ISSUE #388: Find in project does not search in folders with a dot in the name
In folders like: `desktop.level`.
If dot is at the beginning – all works fine.

### LINKED ISSUE #6134: "Search in Directory" not working when root directory name contains a period
The `Search in Directory` command does not seem to work properly when there are multiple root folders (as below illustrates), but works as expected with a single root folder.

out2

I'm on Ubuntu 14.10 and v0.188.0 of Atom.

benogle:
Scandal now stats path patterns that do not contain any `*` chars to better determine if a pattern is a file or directory. So the case of "Search in directory" when the directory contains dots will now work.

screen shot 2015-08-04 at 5 41 58 pm

Closes https://github.com/atom/atom/issues/6134
Closes https://github.com/atom/find-and-replace/issues/388
Closes https://github.com/atom/find-and-replace/issues/309

@jugglingnutcase feel free to try out this branch to see if it fixes your problem. You will need to `script/bootstrap` after pulling down this branch to get the new scandal version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "q": "^1.1.2",
     "random-words": "0.0.1",
     "runas": "2.0.0",
-    "scandal": "2.1.1",
+    "scandal": "2.1.2",
     "scoped-property-store": "^0.17.0",
     "scrollbar-style": "^3.1",
     "season": "^5.3",


jugglingnutcase:
image

image

I ran the same test in the original issue I filed and... looks like it's working! Thanks :smile_cat:

jugglingnutcase:
Thanks for digging into this one @benogle, I had trouble finding this myself :+1:

izuzak:
:clap: :clap: @benogle

VodkaBears:
:+1:

PR #8228, opened 2015-08-05
Add syncScrollRatio option and activate option.

t9md:
Currently,  splitRight, Left, Up, Down don't sync scrollRatio of original editor.
Since copyActiveItem: is true by default, I want newPane scrolled to same ratio of original item if item was instance of TextEditor.
- activate: false won't activate newPane
- syncScrollRatio: true sync original scrollRatio on split.

Why I added `activate` option is I want split pane without activation, in Vim it split without activation(and of course sync scrollRatio).

Currently spec was not finished. I couldn't getHeight() of editor, its always say 0, I need help to fix this.

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -539,7 +539,7 @@ describe "Pane", ->
           expect(pane2.isDestroyed()).toBe true
           expect(item4.isDestroyed()).toBe false
 
-  fdescribe "split methods", ->
+  describe "split methods", ->
     [pane1, container, editor, scrollTop, textPane] = []
 
     setEditorEnvironment = (editor) ->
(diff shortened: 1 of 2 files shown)


t9md:
I could set editorEnvironment like height, lineHeightInPixel etc, but these value is not copied and most importantly even original editor's height is not updated after split( it should be half height of original height after splitDown()).

So I need help, I'm currently thinking to write test for `syncScrollRatioToPane()` function, and let each split function only check if its call `syncScrollRatioToPane()`.

lee-dohm:
I'm going through old PRs to see if they can be salvaged or should be closed. Is this something that still needs fixing? Or is it superseded by #7953?

Let us know :grinning:

t9md:
Still need to fix. Not superseded by other issue.
But no one responded for long time, I have no motivation to push this currently. So Im OK to close this.

Issue #2470, opened 2014-05-31
"Move Line Up" and "Move Line Down" with multiple selections


ifthenelse:
After making multiple selections, "Move Line Up" (<kbd>Ctrl</kbd> <kbd>Cmd</kbd> <kbd>Up</kbd>) and "Move Line Down" (<kbd>Ctrl</kbd> <kbd>Cmd</kbd> <kbd>Down</kbd>) works only on the last selected line, instead it should move all the lines.

![atom-multiple-selection-move-issue](https://cloud.githubusercontent.com/assets/50681/3138639/e528d254-e8a2-11e3-9608-e67da60182cf.gif)

jonathandelgado:
+1

sobi3ch:
+1

tonka3000:
+1

stokescomp:
+1

xaviergmail:
+1

blqke:
+1

devonestes:
+1

ryanjuve-porch:
+1

quinHD:
+1

fnfilho:
+1

mnquintana:
Thanks for the feedback everyone! However, all these +1s and :+1:s are just adding noise, so to minimize that I'm going to lock this issue for now – we may unlock it once someone starts working on this enhancement to get further feedback.

50Wliu:
This has been fixed in #8232 (thanks to @lpommers and @abe33) and should be available in the next Atom beta or the Atom 1.2.0 stable release.

nathansobo:
Clarification: 1.3.0 stable since it was merged into 1.3.0-dev. 1.2.0-betaN is already out.

Issue #7653, opened 2015-07-02
Move-Line-Down doesn't move entire line below last line


stenverbois:
Selecting an entire line (or lines) prevents `editor:move-line-down` to swap it
with the last line in the editor.

Note: It does work when not selecting the entire line (i.e. not selecting the newline char)

![GIF:](http://media.giphy.com/media/xTiTnoUoSJL5vHiLKw/giphy.gif)

## Repro Steps

1. Open a file in the main editor
2. Select an entire line or lines
3. Use `editor:move-line-down` (or default `Ctrl-Down`)
4. Try to move it below the last line in the editor

**Expected:** The selected line swaps with the last line
**Actual:** It doesn't do anything

## Versions

* **Atom:**       1.0.0
* **Electron:** 0.22.3
* **OS:**         linux 3.19.0-15-generic
* **Misc**
    * apm  1.0.1
    * npm  2.5.1
    * node 0.10.35
    * python 2.7.9
    * git 2.1.4

---

<small>This report was created using the package `bug-report` v0.7.1.</small>

kubetz:
I am currently on Windows 7 using Atom 1.0.0 and it is works for me :open_mouth: .

Have you tried closing all instances of atom and running `atom --safe`?

![works](https://cloud.githubusercontent.com/assets/1128248/8477997/c40ad454-20ce-11e5-89b9-77a612bf45ca.gif)

kubetz:
Was able to reproduce it on Windows 7 Atom 1.0.0.

Exactly as written above - occurs *only* when whole line is selected.

stenverbois:
Nice find, that's indeed the code that is causing this I think. The issue may be that `selection.end` of a fully selected line is on position 0 of the next line? That's where the cursor appears if you press `right` after selecting anyway.

kubetz:
@stenverbois Yeah, indeed.

I have no idea if it should fixed in `moveLineDown` directly or deeper in code that is setting `selection.end.row` and if it is fixed deeper then I am not sure if it won't break something else.

That is the reason I am not trying to fix anything :blush: . It also looks like there are no specs for moving lines :cold_sweat: .

benogle:
This was fixed by #8232

Issue #8023, opened 2015-07-22
editor:cut-to-end-of-line on empty line above fold deletes the whole fold


KnutHelland:
I feel there are a couple of inconsistencies with code folding. At least I can prove this one:

`editor:cut-to-end-of-line` is by default bound to `ctrl-k` on darwin, so I use it a lot for deleting current line.

The default behavoiur when cursor is on an _empty_ line is to delete the current line (delete the newline).

But if the line below is folded, that one is deleted as well (and not even cut. It's deleted). I've tried to make a demonstration in this gif. It first demonstrates normal behaviour, and then what happens when the line below is folded.

![Demo](http://zippy.gfycat.com/ShortWeightyHorseshoebat.gif)

KnutHelland:
One additional and similar bug: `editor:newline-above` on the line below a folded are should add a newline between current line and the folded are, but it is inserted above the folded area (and the fold is expanded).

I wrote that "I feel there are a couple of inconsistencies", but I'm almost sure that these two bugs are the only one I've encountered.

renestalder:
Same problem when pressing backspace just below a folded code block. The whole block will be deleted. Very annoying behaviour.

PR #8232, opened 2015-08-05
Move lines up and down with multiple selections

lpommers:
Closes https://github.com/atom/atom/issues/2470

move-line

`ctrl-cmd-up` and `ctrl-cmd-down` now supports moving multiple selections up and down.


benogle:
cc @nathansobo I wonder if we can find the old specs for these methods. They may have been blown away inadvertently.

nathansobo:
I think we should probably just roll forward unless you want to go hunting.

spec/text-editor-spec.coffee
```diff
@@ -1986,6 +1986,82 @@ describe "TextEditor", ->
         expect(editor2.getSelectedBufferRanges()).not.toEqual editor.getSelectedBufferRanges()
 
   describe "buffer manipulation", ->
+    describe ".moveLineUp", ->
+      describe "when there is only one line selected", ->
+        it "moves the line up by one row", ->
+          editor.setSelectedBufferRange([[3, 2], [3, 9]])
+          expect(editor.getSelectedBufferRange()).toEqual {start: {row: 3, column: 2}, end: {row: 3, column: 9}}
```
nathansobo: In specs we typically use a shorter array-based notation: `[[3, 2], [3, 9]]`

spec/text-editor-spec.coffee
```diff
@@ -1986,6 +1986,82 @@ describe "TextEditor", ->
...
+
+          editor.moveLineUp()
+
+          expect(editor.getSelectedBufferRange()).toEqual {start: {row: 2, column: 2}, end: {row: 2, column: 9}}
+
+      describe "when there is multiple selections", ->
+
```
nathansobo: :fire: extra newline

spec/text-editor-spec.coffee
```diff
@@ -1986,6 +1986,82 @@ describe "TextEditor", ->
...
   describe "buffer manipulation", ->
+    describe ".moveLineUp", ->
+      describe "when there is only one line selected", ->
+        it "moves the line up by one row", ->
+          editor.setSelectedBufferRange([[3, 2], [3, 9]])
+          expect(editor.getSelectedBufferRange()).toEqual {start: {row: 3, column: 2}, end: {row: 3, column: 9}}
+
+          editor.moveLineUp()
+
+          expect(editor.getSelectedBufferRange()).toEqual {start: {row: 2, column: 2}, end: {row: 2, column: 9}}
```
nathansobo: In addition to the range checks, you should also make assertions about the content of these lines.

spec/text-editor-spec.coffee
```diff
@@ -1986,6 +1986,82 @@ describe "TextEditor", ->
...
+            editor.setSelectedBufferRanges([[[1, 2], [1, 9]], [[3, 2], [3, 9]], [[5, 2], [5, 9]]])
+            expect(editor.getSelectedBufferRanges()).toEqual [[[1, 2], [1, 9]], [[3, 2], [3, 9]], [[5, 2], [5, 9]]]
+
+            editor.moveLineUp()
+            editor.moveLineUp()
+
+            expect(editor.getSelectedBufferRanges()).toEqual [
+              {start: {row: 0, column: 2}, end: {row: 0, column: 9}},
+              {start: {row: 2, column: 2}, end: {row: 2, column: 9}},
+              {start: {row: 4, column: 2}, end: {row: 4, column: 9}} ]
```
nathansobo: Another stupid code styling tweak: This `]` should be on the next line.

src/text-editor.coffee
```diff
@@ -818,114 +818,126 @@ class TextEditor extends Model
...
-      for row in rows
-        if fold = @displayBuffer.largestFoldStartingAtBufferRow(row)
-          bufferRange = fold.getBufferRange()
-          startRow = bufferRange.start.row
-          endRow = bufferRange.end.row
-          foldedRows.push(startRow - insertDelta)
+    newSelectionBufferRanges = []
+    selections = @getSelectedBufferRanges()
+    selections.sort (a, b) ->
+      return a.compare(b)
```
nathansobo: Don't need `return` in CoffeeScript.

src/text-editor.coffee
```diff
@@ -818,114 +818,126 @@ class TextEditor extends Model
...
+    for selection in selections
+      return if selection.start.row is 0
+      lastRow = @buffer.getLastRow()
+      return if selection.isEmpty() and selection.start.row is lastRow and @buffer.getLastLine() is ''
```
nathansobo: Is there test coverage for this line? Can you explain why it's needed?

spec/text-editor-spec.coffee
lpommers: @nathansobo I noticed this is as well... but whenever I ran my tests they failed, asking for the specific range objects instead of the array style notation. I didn't know what else to do; but I did notice the array based styled used elsewhere

src/text-editor.coffee
```diff
@@ -818,114 +818,126 @@ class TextEditor extends Model
...
-          endRow = row
+          insertDelta = 1
 
-        insertPosition = Point.fromObject([startRow - insertDelta])
-        endPosition = Point.min([endRow + 1], @buffer.getEndPosition())
-        lines = @buffer.getTextInRange([[startRow], endPosition])
-        if endPosition.row is lastRow and endPosition.column > 0 and not @buffer.lineEndingForRow(endPosition.row)
-          lines = "#{lines}\n"
+        for row in rows
+          if fold = @displayBuffer.largestFoldStartingAtBufferRow(row)
```
nathansobo: I'm actually a bit unclear why the existing code needed to talk explicitly about folds... Is it possible to work strictly in terms of screen coordinates and have things work? Meaning if we're moving a selection that starts at screen row 10 up, it should be safe to insert it at screen row 9 and that should automatically account for folds. Am I missing anything with this thinking? Did you try working exclusively in screen coordinates?

src/text-editor.coffee
```diff
@@ -818,114 +818,126 @@ class TextEditor extends Model
...
+            @unfoldBufferRow(insertPosition.row)
+            foldedRows.push(insertPosition.row + endRow - startRow + fold.getBufferRange().getRowCount())
 
-      # Restore folds that existed before the lines were moved
-      for foldedRow in foldedRows when 0 <= foldedRow <= @getLastBufferRow()
-        @foldBufferRow(foldedRow)
+          @buffer.insert(insertPosition, lines)
 
-      @setSelectedBufferRange(selection.translate([-insertDelta]), preserveFolds: true, autoscroll: true)
+        # Restore folds that existed before the lines were moved
```
nathansobo: Yeah... my intuition tells me we shouldn't need to mess with unfolding and refolding folds here. Help me understand if I'm wrong.

nathansobo:
:heart: Thanks for taking a crack at this. I'd like to see if we can simplify the implementation a bit with respect to fold handling, but this is a great start.

spec/text-editor-spec.coffee
nathansobo: Wow, that's odd. There may be some method that's not doing a `Range.fromObject` on its arguments and causing issues, but the equality checks should work. Could you try one more time with this particular assertion and if it doesn't work I'll investigate?

lpommers: Feeling a bit like the boy who cried wolf because I didn't have any problems using the array syntax today.. strange. changes made

src/text-editor.coffee
lpommers: I didn't really experiment with screen coordinates. I will spend some time playing around and seeing what happens

lpommers:
@nathansobo I will spend some more time digging into the folds and possibly using screen coordinates. I didn't really investigate too deep when examining this as the main culprit that jumped out was that it just didn't use all the selected buffer ranges.

nathansobo:
@lpommers Checking in on this... you ready for another review?

lpommers:
@nathansobo unfortunately not. I've spent a bit of time playing with using screen coordinates and am still having some trouble making it work with folds. When I come across a fold, instead of putting the row above it, it places it on the same line as the starting fold. Not sure why yet...

src/text-editor.coffee
lpommers: @nathansobo Been playing with screen coordinates. Just want to make sure I have this straight: In order to insert text at a certain screen row, it's position needs to be converted using something like `bufferPositionForScreenPosition` and then inserted using `insert`, correct?

nathansobo: Correct. Are you encountering problems with that approach. Perhaps there's something I didn't consider.

nathansobo:
Taking a deeper look here... you're definitely right that my initial suggestion of leaning heavily on screen coordinates isn't enough... folds will have to be dealt with explicitly. Which is a bummer.

nathansobo:
Okay, I've made some changes to your original PR for `move-line-up` on this [branch](https://github.com/atom/atom/compare/move-lines-up-and-down-with-multiple-selections). Would you be interested in taking a look at my adjustments and if you're in favor, add them to this PR and apply a similar treatment to `move-lines-down`?

src/text-editor.coffee
```diff
@@ -818,114 +818,133 @@ class TextEditor extends Model
...
+        lines += @buffer.lineEndingForRow(linesRange.end.row - 1) unless lines[lines.length - 1] is '\n'
+        @buffer.delete(linesRange)
+        @buffer.insert([precedingBufferRow, 0], lines)
 
-      # Restore folds that existed before the lines were moved
-      for foldedRow in foldedRows when 0 <= foldedRow <= @getLastBufferRow()
-        @foldBufferRow(foldedRow)
+        # Restore folds that existed before the lines were moved
+        for rangeToRefold in rangesToRefold
+          @displayBuffer.createFold(rangeToRefold.start.row, rangeToRefold.end.row)
```
lpommers: @nathansobo here don't we want to refold at `rangeToRefold.start.row - 1 ` instead? The fold has moved down one row after the insert, so it's start row is now different. Creating the fold on the start row pushes the folding down one row too many. 
refold

src/text-editor.coffee
```diff
@@ -818,114 +818,133 @@ class TextEditor extends Model
...
 
-  # Move lines intersection the most recent selection up by one row in screen
+  # Move lines intersection the most recent selection or multiple selections up by one row in screen
   # coordinates.
   moveLineUp: ->
-    selection = @getSelectedBufferRange()
-    return if selection.start.row is 0
-    lastRow = @buffer.getLastRow()
-    return if selection.isEmpty() and selection.start.row is lastRow and @buffer.getLastLine() is ''
+    selections = @getSelectionsOrderedByBufferPosition()
```
nathansobo: Sorry, but I realized when I woke up this morning that we probably don't actually have to refer to the selection objects directly... I had changed it so we could reassign the selection ranges one at a time, but that had bad interactions with selection merging, so I went back to your approach of just resetting the selected buffer ranges at the end. So you can probably switch back to your original approach of just working with the sorted ranges.

src/text-editor.coffee
nathansobo: @lpommers Definitely looks wrong, but I think we should correct the problem when we're originally pushing the range onto `rangesToRefold`. Would be good to add a spec that reproduces the behavior you've pictured.

nathansobo:
@lpommers Ping me on this PR when you're ready for another review!

lpommers:
@nathansobo I've implemented the changes into `moveLineDown` and I think this is ready for another review. One thing I just noticed was that if you move a selection down a row so it is one before a fold it unfolds the row. but this isn't called by `moveLineDown`'s `fold.destroy()`. I looked into it and it might be coming from `@displayBuffer.unfoldBufferRow`, but I don't know why, hoping you can give me some insight here.
unfold

nathansobo:
Sorry for the delay on this... taking a look at your changes from this weekend now.

nathansobo:
So, by default creating a selection that crosses a fold destroys the fold. This is so we don't create hidden selections. You can suppress that behavior by passing `preserveFolds: true` as an option to `setSelectedBufferRanges`.

One thing I'm noticing when I do that however is that lines are getting merged into the fold below them when they're moved down. Are you sure you're re-folding the correct rows? I bet what's happening is that the fold is somehow getting created to include the moved line, then you restore the selection and it destroys that line... check this out with folds preserved on creating the selections:

move-lines-and-folds

lpommers:
@nathansobo yep, that was definitely it!

On a somewhat unrelated note, should I make an issue about the documentation? - I didn't know  `setSelectedBufferRanges` could accept `preserveFolds`, the docs only refer to a `reversed` parameter in the `options` object. https://atom.io/docs/api/v1.0.7/TextEditor#instance-setSelectedBufferRanges

nathansobo:
Sorry out the docs. Even better, if you want just change them in a PR and I'll merge it.

nathansobo:
Once you think this is working correctly from a behavior perspective, I'll do another review of the actual code. Ping me and @izuzak and we'll hammer on it and try to find holes. Great work on this by the way. This problem is way more challenging than I would have thought and really suggests some improvements to our underlying modeling of folds.

lpommers:
@nathansobo and @izuzak I think we can take another look :eyes:  at this again. I think we're getting there. And thanks for all the help with this, it's been a bit challenging for me and you've all been really helpful

lpommers:
this PR now somewhat closes https://github.com/atom/atom/issues/7653, since moving down at the end of the file can now continue infinitely, but you will still always see the next screen row as the last 
lines

izuzak:
Hey @lpommers :wave: -- thanks so much for working on this! I was just testing this and noticed one strange thing: when you move selections down then the screen scrolls down correctly when the bottom selection reaches the bottom of the screen, but when you move selections up then the screen scrolls up only after the top selection passes over the top of the screen. Here's a GIF which demonstrates this:

lines

Do you see the same behavior?

I'll play with this a bit more and report if I notice anything else :smile:

izuzak:
Noticed one other strange thing -- here's a GIF:

lines2

I make two selections on separate lines, and then start moving them down. But notice that what's selected changes as I move the selections down (unexpected). When I move the selections up, the selected text doesn't change (expected)

lpommers:
Hi @izuzak, I'm able to reproduce both of those. As to your first comment, I'm not sure what's happening there.. gonna look into it. And the second, I only get that behavior when the two lines are right next to each other - I have an idea as to what might be happening.. gonna look into these (and anything else that pops up) sometime this week when I have some free time.

lpommers:
@izuzak I have implemented fixes for those two bugs you spotted, but I think there is a better way to solve the second bug (screen not scrolling at the proper cursor) - perhaps @nathansobo might have a suggestion.

izuzak:
:zap: @lpommers thanks for the ping :bow:

> but I think there is a better way to solve the second bug (screen not scrolling at the proper cursor) - perhaps @nathansobo might have a suggestion.

@nathansobo Since you've been :hammer: at scrolling recently (and https://github.com/atom/atom/pull/8548 will be merged soon), do you perhaps want to take a quick look at how things are going here before I test things manually again?

src/text-editor.coffee
```diff
@@ -818,114 +818,131 @@ class TextEditor extends Model
...
+        # Restore folds that existed before the lines were moved
+        for rangeToRefold in rangesToRefold
+          @displayBuffer.createFold(rangeToRefold.start.row, rangeToRefold.end.row)
 
-      @setSelectedBufferRange(selection.translate([-insertDelta]), preserveFolds: true, autoscroll: true)
+        for selection in selectionsToMove
+          newSelectionRanges.push(selection.translate([-insertDelta, 0]))
 
-  # Move lines intersecting the most recent selection down by one row in screen
+      @setSelectedBufferRanges(newSelectionRanges)
```
nathansobo: I'd suggest passing `autoscroll: false` here to suppress any autoscroll before scrolling explicitly.

src/text-editor.coffee
```diff
@@ -818,114 +818,131 @@ class TextEditor extends Model
...
+      @scrollToBufferPosition([newSelectionRanges[0].start.row + 2, 0])
```
nathansobo: This seems okay to me if autoscrolling didn't work correctly when setting the ranges. Sometimes it's better to just ask for what you want explicitly and I think this case may be complex enough for it to be one of those times.

src/text-editor.coffee
nathansobo: Maybe you want to `preserveFolds` too just for symmetry?

src/text-editor.coffee
```diff
@@ -818,114 +818,131 @@ class TextEditor extends Model
...
-        @foldBufferRow(foldedRow)
+        # Restore folds that existed before the lines were moved
+        for rangeToRefold in rangesToRefold
+          @displayBuffer.createFold(rangeToRefold.start.row, rangeToRefold.end.row)
+
+        for selection in selectionsToMove
+          newSelectionRanges.push(selection.translate([insertDelta, 0]))
 
-      @setSelectedBufferRange(selection.translate([insertDelta]), preserveFolds: true, autoscroll: true)
+       @setSelectedBufferRanges(newSelectionRanges, {preserveFolds: true})
```
nathansobo: Same here... suppress autoscroll.

nathansobo:
@izuzak aside from a couple minor adjustments, if you can't break this, I'm ready to merge.

src/text-editor.coffee
lpommers: cool. The reason I wasn't sure about going this route is because I couldn't (and still can't) really see why the autoscrolling wasn't working properly. And I wasn't sure if this kind of solution was sort of duct-tapey without addressing the root of the problem.

izuzak:
I'm still seeing some problems when moving multiple selected lines up. Here's a GIF:

lines3

Notice how when I move the selections down -- the same lines are still selected. But when I move them up -- different lines are selected (then numbers in the selections change).

izuzak:
I'm also seeing a strange exception being thrown if I select multiple line, and then press and hold `cmd-z` to undo:

lines4

The exception is `Uncaught TypeError: Cannot read property 'getScreenPosition' of undefined` and it's thrown from: `/Users/izuzak/github/atom/src/text-editor.coffee:420`. Wasn't able to reproduce the same thing on the master branch.

@lpommers Are you able to reproduce that exception? @nathansobo might also be interested in this as well.

izuzak:
I should have probably mentioned that I'm testing these changes on a merge between the latest master and this branch (so that I'm testing what the behavior will be when this is merged).

lpommers:
@izuzak I can only reproduce the first of the two issues. I can't seem to reproduce the `Uncaught TypeError` on a merge between master and the branch. I will look into what's going on with the `moveLineUp` functionality

src/text-editor.coffee
```diff
@@ -818,114 +818,131 @@ class TextEditor extends Model
...
-          endRow = bufferRange.end.row
-          foldedRows.push(startRow - insertDelta)
+      newSelectionRanges = []
+
+      while selections.length > 0
+        # Find selections spanning a contiguous set of lines
+        selection = selections.shift()
+        selectionsToMove = [selection]
+
+        while selection.end.row is selections[0]?.start.row
```
lpommers: @nathansobo can you explain to me why this check is necessary? this seems to be the cause of @izuzak most recent finding. When selecting two "sibling" lines via the line numbers, this hits true and the first is removed causing the top line not to move upward. I've only been able to reproduce this gif when selecting lines from the line numbers, not from a cursor in the editor or click and drag - here's his gif of the issue. 

https://cloud.githubusercontent.com/assets/38924/9610796/fb1b59b8-50db-11e5-9788-5f7abae8499c.gif

lpommers:
@izuzak I've just encountered this bug, can you see it as well? It only happens when selecting lines via the line number gutter.. 
line_down

izuzak:
@lpommers Yep, seeing that as well:

lines6

I'm seeing strange behavior with moving a line that is directly above a folded line even on master (not on your branch):

lines7

Notice how when I'm not moving the whole line -- things work as expected. But if I select the whole line -- then the folded line moves as well. :confused:

nathansobo:
This is getting incredibly tricky. I do think we can keep hacking at this and get it right eventually. I'm wondering if some changes to lower level abstractions might make things easier.

* Folds should really be represented as markers that span the elided text. Right now they start at the beginning of the first line that remains visible, which causes us to stray into them too easily. If the marker for the folded text extended from the end of the first folded line to the end of the fold, maybe we could avoid accidentally going into folds.
* I've always wanted to allow folds to be cut/copied and pasted along with text. I'm wondering if such a facility could make this easier. We'd just take the union of all lines spanned by contiguous selections, cut the content with its fold metadata, then paste it one line down.

I feel like my time might be spent fixing some of these lower level issues, but I will merge this if we can get it working sooner.

jeancroy:
Bug https://github.com/atom/atom/issues/8023 migth be related to this discussion of special behavior on the line above the fold.

lpommers:
@izuzak So, I think we're slowly getting there... I've been able to fix two of those bugs. 
* w/r/t the bug where moving two lines down (selecting via the gutter) through a fold caused the lines to separate: I've fixed the worst part - but currently the fold gets destroyed - so i need more time with that.
* Moving lines up should no long move the wrong selections
* still unable to reproduce the thrown exception

here's a gif of what I mean about the fold getting destroyed, but the lines moving properly

fold

lpommers:
I need to change the two commits i just made - forgot to take into consideration multiple selections on the same line..

src/text-editor.coffee
lpommers: I get this now

lpommers:
@izuzak I think we're getting there... some notes on two of the things I've fixed 

* w/r/t the bug where moving two lines down below a fold (selecting via the line gutter), the issue of the lines breaking apart is solved - but the fold gets destroyed - will need more time
* moving lines up should now have the proper selections
* still haven't been able to reproduce the uncaught exception

gif of what I mean about moving lines down through the fold 
fold

lpommers:
@izuzak  @nathansobo I think we've :hammer: out most of the kinks with this. I'm wondering if it's worth merging this and then making a new issue or PR to handle any specific issues may that pop up as this has become quite large.

izuzak:
@lpommers I applaud your efforts here :clap: :clap: -- thanks for sticking with this. :bow:

Just took this for a test drive again and had a hard time finding problems. I noticed two:

1. The exceptions thrown when I hold and press undo, as described here: https://github.com/atom/atom/pull/8232#issuecomment-136799859. You mentioned that you're unable to reproduce these, so I'd love it if @nathansobo could take a look. I'm unable to trigger these on `master` so I suspect there's something on this branch triggering this. Not sure how serious it is or if it's worth blocking this PR on it.

2. There seems to be a few small problems when moving lines around wrapped lines. Here's an example:

  ```
1
2
Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
3
4
```

  And here's how that looks on my screen wrapped:

  screen shot 2015-09-11 at 18 42 57

  A few problems I observed:
  * If I select both lines 1 and 2 (by cmd-clicking in the gutter) and then try to move them down -- nothing happens. If I disable soft wrapping -- I can move them down just fine.
  * If I put a cursor on line 1 and another on line 2 (so, similar case as above, but whole lines are not selected), a strange thing happens when I try to move the lines down -- lines 1 and 2 switch places and it's not possible to move them down anymore. Here's a GIF:

    p1

    Even if you just place a single cursor on line 2 and try to move it down over the wrapped line -- it won't work. As far as I can tell, this problem doesn't happen on master.

> I'm wondering if it's worth merging this and then making a new issue or PR to handle any specific issues may that pop up as this has become quite large.

I wouldn't be :-1: on that, but will defer to @nathansobo to make the call on that.

Thanks again :zap:

**Comment by lpommers** [comment]
izuzak I should be applauding you guys :bow: for all the help here. It's easy to stick with this, since it's so fun. Without your help, I would have missed a ton of use cases/scenarios; it's something I want to get better at. I'm gonna spend some time this weekend taking another crack at the issues you brought up

**Comment by lpommers** [comment]
@izuzak @nathansobo Thanks to some great work from @abe33 :bow:, the issue with soft wrapped lines is fixed. Maybe we can get some new :eyes: on this since it's been a little while since there was activity, mostly due to me being too busy lately to have any time to focus on this. hopefully that will change soon.

**Comment by izuzak** [comment]
:zap: @abe33 and @lpommers :zap:

Just gave this branch a try again and the soft wrap problem does seem to be fixed, as does the exception I reported in https://github.com/atom/atom/pull/8232#issuecomment-136799859. Great work!

If someone else from @atom/feedback would like to give this branch a try as well -- I think that would be swell.

Also /cc @nathansobo for code reviewing the new changes.

**Comment by nathansobo** [comment]
Code looks good. If this is behaving correctly, I'd say :ship: at your discretion @izuzak.

**Comment by izuzak** [comment]
> Code looks good. If this is behaving correctly, I'd say :ship: at your discretion @izuzak.

Just gave it another go and I did find two differences compared to master. @lpommers @abe33 Do you see the same difference? Sorry for not noticing them when I first tested this!

1. Moving a wrapped line over the top edge doesn't show the full line (the top of the wrapped line is not shown on this branch, but is shown on master).

  Behavior on master (I just move a wrapped line down and back up -- notice how the top of the line is visible when I start moving it back up and over the top edge of the window):
  
  bug1_master

  Behavior on this branch (notice how the top of the wrapped line is not visible when I move it over the top edge. Also notice how the screen "jumps" when I start moving the line up -- that doesn't happen on master either):

  bug1_branch

2. Moving a folded line breaks the folded line.

  Behavior on master (I just fold a line and move it down, then unfold -- works as expected):

  bug2_master

  Behavior on this branch (notice how the folded line breaks apart when I try to move it down):

  bug2_branch

**Comment by abe33** [comment]
@izuzak Thanks for the feedback, I'll try merging back master in the to see if it's related to changes in master or just caused by an issue in our implementation.

**Comment by lpommers** [comment]
@izuzak @abe33 I merged back master locally and still seeing the issues you presented here. Gonna try and spend some time with this..

**Comment by lpommers** [comment]
@izuzak I seem to be seeing these issues you mentioned on both `master` and this branch now..

**Comment by izuzak** [comment]
> @izuzak I seem to be seeing these issues you mentioned on both master and this branch now..

@lpommers That's strange, I'm still seeing the behavior I described in https://github.com/atom/atom/pull/8232#issuecomment-149463992. On master, moving a folded line doesn't break that line on master, and moving wrapped lines above the top edge always shows the cursor. On this branch (with master merged in), a folded line breaks when you try to move it and cursors on wrapped lines are not always visible when you move them over the top edge.

If you're seeing different behavior, can you share some GIF which clearly show that you're using the latest version of master and the two problems reproduced? That might help us figure out why we're not seeing the same behavior.

Thanks! :bow:

**Comment by lpommers** [comment]
@izuzak you're correct. I was looking at the wrong window when I reported that... sorry for the confusion, I got mixed up :flushed:

**Comment by abe33** [comment]
@izuzak @lpommers I'm going to take a look at that this evening.

If I understand correctly: 

1. Solution here is to always make the screen row with the cursor visible at the end of the motion
2. When moving a folded line, it also moves all the hidden lines in the fold without unfolding it.

I'll add tests for these two cases as well.

**Comment by abe33** [comment]
Some news, I fixed most issues with moving folded rows. As folds in `rangesToFold` were not destroyed there was a lot of side effects when moving multiple selections containing folds and close enough that one selection was refolding the previously moved fold (but still present in the display buffer as it wasn't destroyed) with wrong coordinates. 

I just have to fix the issue with scrolling when moving a wrapped line and it should be good for another :eyes:.

**Comment by lpommers** [comment]
@abe33  seeing some interesting behavior w/ folded rows on this branch now... are you seeing this? I've been experiencing some strange behavior on my local build recently, so it might just be my copy at the moment.. 

fols
fols2

**Comment by maxbrunsfeld** [comment]
FYI, I just merged https://github.com/atom/atom/pull/9104, a small PR that adds autoindent when moving lines. It introduced a conflict with this branch, but it should be pretty easy to resolve.

**Comment by abe33** [comment]
@lpommers Interesting, actually I'm not a heavy user of folds so I'm mostly working in the dark here, relying on tests to make sure I'm not missing anything, maybe we should another grammar that js in tests as make sure moving lines with folds operate smoothly in every cases. And there's so many cases it's hard to get the whole picture without a good set of tests. I think I'll try to add more tests using the cases you described to make sure I didn't missed some crucial factor.

@maxbrunsfeld thanks for the heads up, I'll take some time tomorrow to review the PR with the new changes.

**Comment by abe33** [comment]
@lpommers Are you sure you tried with a version of Atom running the branch? I can't reproduce either issues: 
moving-html-lines-with-fold

**Comment by abe33** [comment]
@maxbrunsfeld as you had anticipated it was pretty easy to resolve, I took the liberty to moves the autoindent tests with the other line moves tests so that we don't need to look in several place in the future.

As far as I can tell, the only remaining issue is when moving a wrapped line triggers a scroll. I'm looking into it now.

**Comment by abe33** [comment]
Alright, I think we're pretty much done with this PR, if someone want to give it a look.

For the fun, here's an example of moving a multiple selection with a fold and autoindent enabled: 
moving-html-lines-with-fold-and-autoindent

:sparkles:

**Comment by maxbrunsfeld** [comment]
👏 Nice!

**Comment by lpommers** [comment]
Tested :eyes:  this locally and things are looking good. haven't been able to notice anything. :bow: :pray: :clap: to @abe33 for the work on this. And thanks to everyone else.

not entirely sure what that build error is referring to though.. not sure what we can do about the license of a package? 
```
Running "check-licenses" task
Found dependencies without permissive licenses:
spdx-expression-parse@1.0.1
  License: (MIT AND CC-BY-3.0)
  Repository: git+https://github.com/kemitchell/spdx-expression-parse.js
```

**Comment by abe33** [comment]
Yes, I just noticed that too, @50Wliu had identified the issue coming from atom/legal-eagle not expecting a parenthesis in the license string (https://github.com/atom/legal-eagle/blob/master/src/legal-eagle.coffee#L47-L56).

**Comment by 50Wliu** [comment]
Just restarted the build since legal-eagle has been updated on master.  Might take a while to complete though since I also restarted 3 others before this one.

File: src/text-editor.coffee [hunk 15]
```diff
@@ -866,116 +866,177 @@ class TextEditor extends Model
       @transact groupingInterval, =>
         fn(selection, index) for selection, index in @getSelectionsOrderedByBufferPosition()
 
-  # Move lines intersection the most recent selection up by one row in screen
+  # Move lines intersection the most recent selection or multiple selections up by one row in screen
```
50Wliu: :memo: "intersection" -> "intersecting"

abe33: Thank you for the heads up, I pushed a fix on @lpommers repo, it'll be available when he'll merge the branch.

abe33:
@izuzak @nathansobo If you guys could give it a quick :eyes: before we get too far from master again. The build failure is due to symbols-view whose tests was broken by language-javascript, as far as I can tell @50Wliu fixed that on master as.

50Wliu:
That was actually an accidental commit that I reverted.  The failing symbols-view spec has been super-flaky for a long time now and can be safely ignored: atom/symbols-view#128.

nathansobo:
Insanely good job @lpommers, @abe33 and everyone else involved here. As mentioned, I have some ideas for improvements in core that could simplify this implementation, because it shouldn't be this hard. For now though, happy to have the feature.

izuzak:
I :clap: you for the work here, @lpommers and @abe33, and also for sticking with this PR for so long. 

:dancers:

PR #8243, opened 2015-08-06
Destroy Markers created by DisplayBuffer when buffer is destroyed.

bolinfest:
The implementation of `Marker::destroy()` is:

``` coffee
  destroy: ->
    @bufferMarker.destroy()
    @disposables.dispose()
```

so if `marker.disposables.dispose()` is called instead of `marker.destroy()`, then `marker.bufferMarker.destroy()` does not get called. As best I can tell, `marker.bufferMarker` is a `Marker` returned by a call to `TextBuffer::getMarker(id)` that happens in `DisplayBuffer`.

However, `TextBuffer::getMarker(id)` returns this type of `Marker`:

https://github.com/atom/text-buffer/blob/master/src/marker.coffee

which is distinct from the `Marker` associated with the `DisplayBuffer`:

https://github.com/atom/atom/blob/master/src/marker.coffee

which is a little confusing.

By failing to call `marker.bufferMarker.destroy()`, callbacks registered via `onDidDestroy()` on the `Marker` from the `text-buffer` package will not get called.

Incidentally, there is also this line in the constructor of `Marker`, but that does not seem to help us here:

``` coffee
@disposables.add @bufferMarker.onDidDestroy => @destroyed()
```

I verified that the associated unit test fails without this change and succeeds when the change is added.

### DIFF
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -1174,7 +1174,7 @@ class DisplayBuffer extends Model
 
   destroyed: ->
     fold.destroy() for markerId, fold of @foldsByMarkerId
-    marker.disposables.dispose() for id, marker of @markers
+    marker.destroy() for id, marker of @markers
     @scopedConfigSubscriptions.dispose()
     @disposables.dispose()
     @tokenizedBuffer.destroy()
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -4044,6 +4044,14 @@ describe "TextEditor", ->
       editor.destroy()
       expect(destroyObserverCalled).toBe true
 
+    it "notifies ::onDidDestroy observers of Markers created for this editor", ->
+      gutterMarker = editor.markBufferPosition([0, 0])
+      onDidDestroyObserver = jasmine.createSpy('onDidDestroy')
+      gutterMarker.onDidDestroy(onDidDestroyObserver)
+
+      editor.destroy()
+      expect(onDidDestroyObserver.callCount).toBe 1
+
   describe ".joinLines()", ->
     describe "when no text is selected", ->
       describe "when the line below isn't empty", ->


src/display-buffer.coffee
```diff
@@ -1174,7 +1174,7 @@ class DisplayBuffer extends Model
 
   destroyed: ->
     fold.destroy() for markerId, fold of @foldsByMarkerId
-    marker.disposables.dispose() for id, marker of @markers
```
kevinsawicki: @nathansobo this used to be `unsubscribe()` instead calling dispose directly before the 1.0 API changes, do you happen to remember why that was and why it didn't always call `destroy()` on the marker?

kevinsawicki: Nevermind, look like it was originally added by me quite a long time ago in https://github.com/atom/atom/commit/1764677b7d039fbcbd9febef550703b2b9a1c04c to fix some sort of regression

kevinsawicki:
@nathansobo would still be great to get a pair of :eyes: on this in case you see any problems with this, looks good to me though

kevinsawicki:
In re-reading #593 it looks like maybe we did try this approach at one time but there was an issue, granted this was a very long time ago.

> This is because we don't properly destroy markers when an edit session is destroyed. When we tried the simple fix of destroying them, it created problems with the cursor on the surviving edit session.

nathansobo:
I'm in favor of this, but it looks like we're having some failures in package specs.

This goes hand-in-hand with another change that @maxbrunsfeld was thinking about making several months ago, but didn't out of fear of breaking packages.

Ideally, markers created in one editor shouldn't be visible in any other editor. But what about the following questions?

* Should markers created directly  via the buffer API be visible from editors? I think yes.
* Should markers created via an editor be visible via the buffer API? I think yes.

Also, are there packages that currently depend on seeing markers created in one editor from another editor? Looks like yes from the failures in our spec suite.

bolinfest:
@nathansobo Which specs are failing for you? When I ran `./script/grunt lint`, everything succeeded. How reliable is the signal from Travis-CI these days?

nathansobo:
@bolinfest Specs for a couple bundled packages are failing [here](https://travis-ci.org/atom/atom/jobs/74348797)... Travis can timeout and stuff but these actually look legit to me.

bolinfest:
OK, I'll look into the `autocomplete-plus` and `whitespace` specs.

lee-dohm:
I'm going through old PRs to see if they can be salvaged or should be closed. @bolinfest was this superseded by the new MarkerLayer stuff?

damieng:
No response.

Issue #8246, opened 2015-08-06
Windows defender quarantining Atom 1.0.4 as malware


simonbuerger:
See discussion https://discuss.atom.io/t/windows-10-defender-atom-exe-malware/19645 - apparently it's not just me. Seems to mostly be people on Windows 10.

atom.exe is being quarantined as being infected with TrojanDownloader:Win32/Lerspeng.B ([Malware protection center page](http://www.microsoft.com/security/portal/threat/encyclopedia/entry.aspx?name=TrojanDownloader%3aWin32%2fLerspeng.B&threatid=2147686694&enterprise=0&wa=wsignin1.0#tab=1))

Hopefully it's just a false positive, but obviously it's a massive problem for Windows users.

CmdrCupcake:
My 1.0.4 atom.exe is being quarantined as being infected with Backdoor:Win32/Kelihos.F ([Maleware center page](http://www.microsoft.com/security/portal/threat/encyclopedia/entry.aspx?name=Backdoor%3aWin32%2fKelihos.F&threatid=2147655469&enterprise=0))on Windows 10.

Zireael07:
Duplicate of #3927. (seriously, just search for "malware" before making an issue)

tinymins:
![image](https://cloud.githubusercontent.com/assets/1808990/9106601/14ce411e-3c54-11e5-9085-88d012d5367d.png)
![image](https://cloud.githubusercontent.com/assets/1808990/9106593/0e942a70-3c54-11e5-8c80-6e3cd2d188a7.png)

paulcbetts:
@Zireael07 I'm not so sure, the others were all reputation-based or generally bullshit, this seems far worse

paulcbetts:
@tinymins, @CmdrCupcake, @simonbuerger Can all of you get a SHA1 or MD5 hash and the file size of your Atom.exe?

tinymins:
文件: C:\Users\diyiming\AppData\Local\atom\app-1.0.4\atom.exe
大小: 7207936 字节
文件版本: 0.22.3
修改时间: 2015年8月6日, 16:01:42
MD5: 3621AF2D067448B969557DFDA7066DD7
SHA1: 5A2788BEE32D6A4E1A4B11E8CD987CC2FF074C1A
CRC32: 49674D34

paulcbetts:
[...]

PR #8255, opened 2015-08-06
Codesign Update.exe before building installer

kevinsawicki:
Follow on to #8254, sign `Update.exe` as well before building the installer.

Refs #8246


paulcbetts:
:+1:

PR #7877, opened 2015-07-13
Upgrade to Electron v0.30.6

thomasjo:
This is a continuation of #6877.
## Blocking issues
- [x] Building with io.js v2.x is not supported.
- [x] Crash caused by invalid memory access during exit/closing window.
- [x] 8 failing spec assertions.
- [x] In OS X closing all windows exits the application.
- [x] Building on Linux (Ubuntu 12.04.5 LTS) is failing due to compilation errors.
- [x] Failing settings-view specs.
- [x] apm builds incompatible native modules.
- [x] Experiment with upgrading to Electron v0.31.x.  
  **Conclusion**: Not viable in this PR since it requires bumping every single dependency to NAN 2.x.
- [x] Wait until https://github.com/atom/incompatible-packages/pull/9 is ready to merge.
- [x] Preserve `CustomEvent` as type of object dispatched by command registry
- [x] Investigate why symbols-view is failing to find any symbols
- [x] Investigate extremely long paint times for cursor movement on first-gen MBP retinas
#### Building with io.js v2.x is not supported

**Resolution**: Do not update npm. The version shipped with io.js is patched to work around the problems with node-gyp.

Any attempt to execute `script/build` when using io.js v2.x fails with

```
Installing build modules...
gyp ERR! configure error
gyp ERR! stack Error: 404 response downloading https://nodejs.org/dist/v2.4.0/node-v2.4.0.tar.gz
gyp ERR! stack     at Request.<anonymous> (/Users/thomasjo/.nvm/versions/io.js/v2.4.0/lib/node_modules/npm/node_modules/node-gyp/lib/install.js:251:14)
gyp ERR! stack     at emitOne (events.js:82:20)
gyp ERR! stack     at Request.emit (events.js:169:7)
gyp ERR! stack     at Request.onRequestResponse (/Users/thomasjo/.nvm/versions/io.js/v2.4.0/lib/node_modules/npm/node_modules/request/request.js:985:10)
gyp ERR! stack     at emitOne (events.js:77:13)
gyp ERR! stack     at ClientRequest.emit (events.js:169:7)
gyp ERR! stack     at HTTPParser.parserOnIncomingClient (_http_client.js:415:21)
gyp ERR! stack     at HTTPParser.parserOnHeadersComplete (_http_common.js:88:23)
gyp ERR! stack     at TLSSocket.socketOnData (_http_client.js:305:20)
gyp ERR! stack     at emitOne (events.js:77:13)
gyp ERR! System Darwin 14.4.0
[...]

PR #8256, opened 2015-08-06
Use native promises instead of Q library

nathansobo:


YuriSolovyov:
Why not use bluebird as a faster drop-in replacement? Just for sake of removing deps?

nathansobo:
> Why not use bluebird as a faster drop-in replacement?

I just want the simplest, most minimal API that can work. Bluebird seems better than Q from afar but I'd rather just go with the most minimal, standard solution if I'm going to bother making these changes.

Unfortunately, this is now blocked on [upgrading to the latest Electron](https://github.com/atom/atom/pull/7877), and making sure we're on a version that includes the change made in https://github.com/atom/electron/pull/2447.

nathansobo:
`atom-rpm` build failure is spurious. This is building cleanly. Merging at long last.

nathansobo:
Oh, after I resolve these conflicts. :grimacing:

maxbrunsfeld:
awesome :fire: awesome :fire: awesome :rocket:

PR #8308, opened 2015-08-10
Make bufferRangeForScopeAtPosition work with last column of scope

### LINKED ISSUE #7961: editor.bufferRangeForScopeAtCursor('link') returns undefined
How to reproduce, given an empty markdown file:

``` markdown
[link](http://example.com)_
```

Where `_` represents the cursor position (at the end of the link)

And I try to get the buffer range for scope `link`:

``` javascript
editor = atom.workspace.getActiveTextEditor()

// check link scope selector is there
cursor = editor.getLastCursor()
cursor.getScopeDescriptor().getScopesArray()
// => ["source.gfm", "link", "markup.underline.link.gfm", "punctuation.definition.end.gfm"]

// since link scope is there, when i try to get the buffer range
editor.bufferRangeForScopeAtCursor('link')
// => undefined
editor.bufferRangeForScopeAtCursor('markup.underline.link.gfm')
// => undefined
```

nathansobo:
Fixes #7961

### DIFF
--- a/spec/tokenized-buffer-spec.coffee
+++ b/spec/tokenized-buffer-spec.coffee
@@ -585,6 +585,7 @@ describe "TokenizedBuffer", ->
     describe "when the selector matches a single token at the position", ->
       it "returns the range covered by the token", ->
         expect(tokenizedBuffer.bufferRangeForScopeAtPosition('.storage.modifier.js', [0, 1])).toEqual [[0, 0], [0, 3]]
+        expect(tokenizedBuffer.bufferRangeForScopeAtPosition('.storage.modifier.js', [0, 3])).toEqual [[0, 0], [0, 3]]
 
     describe "when the selector matches a run of multiple tokens at the position", ->
       it "returns the range covered by all contigous tokens (within a single line)", ->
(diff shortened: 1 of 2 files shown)


izuzak:
@nathansobo I just wanted to double-check my understanding of scopes, related to my question in https://github.com/atom/atom/issues/7961#issuecomment-128679083.

Let's say I have two lines of Markdown:

```
[link](http://example.com)
[link](http://example.com)foo
```

And let's say that I ask for the scopes at the cursor position for these two cases:

1. The cursor is in the first line, right after `)` (position 1:27)
2. The cursor is in the second line, right after `)` (position 2:27)

My expectation was that the scopes would be the same in both cases -- only `source.gfm`. However, the scopes which are reported in the first case are `source.gfm, link, markup.underline.link.gfm, punctuation.definition.end.gfm`, and in the second case it's `source.gfm`. 

Can you help me understand why that's so? :pray:

nathansobo:
@izuzak Here's the reason:

* We always return a scope descriptor for the token *starting* at the given position
* If there is no token starting at the given position, because we're at the end of the line, we return a scope descriptor for the previous token.

I'm not sure this is the best system, but I experimented briefly with favoring the scope of the preceding token and that isn't always intuitive either.

izuzak:
> If there is no token starting at the given position, because we're at the end of the line, we return a scope descriptor for the previous token.

Thank you, @nathansobo -- that explains things. :bow:

PR #8321, opened 2015-08-11
Clear pending frames on folding specs

### LINKED ISSUE #8301: Failing OS X builds on Travis CI
The OS X builds on Travis CI are currently quite consistently failing a singular test;
```
TextEditorComponent
  gutter rendering
    fold decorations
      mouse interactions with fold indicators
        when the component is destroyed
          it stops listening for folding events
            Error: No animation frame requested
              at noAnimationFrame (/Users/travis/build/atom/atom/spec/text-editor-component-spec.coffee:22:39)
              at [object Object].<anonymous> (/Users/travis/build/atom/atom/spec/text-editor-component-spec.coffee:827:13)
```

For a concrete example of a failing build, see https://travis-ci.org/atom/atom/jobs/74894506#L276.
This is also happening against [`master` builds](https://travis-ci.org/atom/atom/jobs/74537917#L265).

Seems to be related to gutter components. It has happened in the past, but infrequently, but seems to happen quite consistently now. Timing related perhaps? :racehorse: 

/cc @as-cii

as-cii:
Fixes #8301.

It seems like, for a time related problem, we're getting intermittent build failures on Travis [for this line](https://github.com/atom/atom/blob/master/spec/text-editor-component-spec.coffee#L827) (which was added as part of #8009).

This PR simply bypasses the `nextAnimationFrame` call when there is no pending frame (as we did for https://github.com/atom/atom/blob/master/spec/text-editor-component-spec.coffee#L852).

/cc: @atom/feedback @thomasjo

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -824,7 +824,8 @@ describe "TextEditorComponent", ->
 
         describe "when the component is destroyed", ->
           it "stops listening for folding events", ->
-            nextAnimationFrame()
+            nextAnimationFrame() unless nextAnimationFrame is noAnimationFrame # clear pending frame request if needed
+
             component.destroy()
 
             lineNumber = component.lineNumberNodeForScreenRow(1)


thomasjo:
Travis build just went :green_heart: so this looks promising!

thomasjo:
Thanks again @as-cii!

PR #8350, opened 2015-08-12
Autoscroll correctly when dragging gutter selections

### LINKED ISSUE #2696: Clicking and dragging up in gutter does not scroll
Clicking on the gutter and dragging up does not expand the selection and scroll the editor when you reach the top of the gutter.

This didn't work in the old editor but I noticed in the new editor it doesn't work either.

It does scroll down though correctly when clicking and dragging.

nathansobo:
Fixes #2696
- Explicitly assign the orientation of the selection when it crosses the initially clicked row.
- Autoscroll to the cursor, not the entire selection.

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1953,6 +1953,32 @@ describe "TextEditorComponent", ->
           gutterNode.dispatchEvent(buildMouseEvent('mouseup', clientCoordinatesForScreenRowInGutter(2)))
           expect(editor.getSelectedScreenRange()).toEqual [[2, 0], [7, 0]]
 
+      it "orients the selection appropriately when the mouse moves above or below the initially-clicked row", ->
+        gutterNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenRowInGutter(4)))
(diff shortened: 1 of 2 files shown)


izuzak:
:clap: Thanks for fixing this, @nathansobo. I was just playing with this and noticed that the problem still happens if you press and hold `cmd` or `shift` before select-scrolling :cactus:. Would that be an easy followup fix or would it be helpful to open a new issue for this?

nathansobo:
I can do a followup.

nathansobo:
@atom/feedback As I tested this, I also noticed that selecting with the mouse in the gutter selects entire buffer lines instead of screen lines. This feels a bit counter-intuitive to me and leads to the selections feeling jumpy instead of smooth. What would people think about always selecting screen lines in the gutter?

mnquintana:
> What would people think about always selecting screen lines in the gutter?

:+1: This would make the most sense to me too

izuzak:
> What would people think about always selecting screen lines in the gutter?

Ahhh, just noticed this comment -- sorry, @nathansobo, I left a note in your PR: https://github.com/atom/atom/pull/8377#issuecomment-131007707. Would love to hear more thoughts on this and also how other editors behave.

lee-dohm:
I would suggest using screen lines in the case of soft wrap, but in the case of folding, it should use buffer lines.

PR #8356, opened 2015-08-13
Support `editor.tabType`

benogle:
This does the simplest thing possible. When `editor.tabType` is set to `auto`, it does what it did before. When it is set to `hard` or `soft`, it uses those values all the time unless `setSoftTabs()` is explicitly called. The API is unchanged. Specs have been consolidated a bit as there were a couple places where the tab stuff was being tested.

Eventually (another PR), I'd like to change the name of `editor.softTabs` setting to `editor.defaultTabType` or something more clear. 

Closes https://github.com/atom/atom/pull/8186 @Abdillah, I've used a couple of your commits.
Closes https://github.com/atom/atom/issues/3719

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -87,9 +87,6 @@ class TextEditor extends Model
     @displayBuffer ?= new DisplayBuffer({buffer, tabLength, softWrapped, ignoreInvisibles: @mini, largeFileMode})
     @buffer = @displayBuffer.buffer
 
-    @disposables.add atom.config.observe 'editor.tabType', scope: @getRootScopeDescriptor(), =>
-      @softTabs = @shouldUseSoftTabs(defaultValue: @softTabs)
-
     for marker in @findMarkers(@getSelectionMarkerAttributes())
       marker.setProperties(preserveFolds: true)
       @addSelection(marker)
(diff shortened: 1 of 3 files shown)


src/text-editor.coffee
```diff
@@ -176,9 +176,15 @@ class TextEditor extends Model
       @subscribe @displayBuffer.onDidAddDecoration (decoration) => @emit 'decoration-added', decoration
       @subscribe @displayBuffer.onDidRemoveDecoration (decoration) => @emit 'decoration-removed', decoration
 
+  subscribeToTabTypeConfig: ->
+    @tabTypeSubscription?.dispose()
```
benogle: This is not super pretty. As I need to either add it to @disposables then remove on a grammar change, or explicitly destroy it in the `destroyed` method. I chose the latter

maxbrunsfeld:
Do you think we should deprecate `editor.softTabs`?

benogle:
One caveat with this is: if the user has explicitly set the soft tabs value to something and an event happens (config or grammar change), the user's setting is ignored and computed based on the rules of the setting. I thought it added extra complexity to support tracking the explicitly set choice, and I wasnt sure if it would be the correct behavior anyway.

benogle:
> Do you think we should deprecate editor.softTabs?

We could change the name to `editor.defaultTabType` or something? Or yeah, we could deprecate it.

benogle:
Anyone opposed to merging this?

maxbrunsfeld:
:ship:

Issue #7332, opened 2015-06-18
Add Resize - Panel Keyboard Commands


chrishough:
References: https://github.com/santip/resize-panes/issues/21#issuecomment-113251749 

Before the resize-pane project was consumed by core it had shortcuts for resizing the panes. I know many people here use the mouse to resize, but this was a **key** feature for those of us who love the speed of the keyboard. 

Can we add this back somehow? 

@mnquintana

izuzak:
> Can we add this back somehow?

@chrishough I like that idea, but the core Atom team probably won't have time to tackle this in the near future. If someone from the community wants to submit a pull request to add this ability (similar to how @liuxiong332 added the resizing ability in https://github.com/atom/atom/pull/5902), the core team would be happy to help by reviewing that PR. Does that sound good?

chrishough:
While I would not have the skill set to pull that off just yet, I would toss a git-tip for that feature if someone would be game to build it in. @liuxiong332

liuxiong332:
Ok, I have added the resize commands #7354 and expect to be merged.

chrishough:
@liuxiong332 damn, that was fast.  Thank you so much, I truly appreciate it.  This feature from the old plugin for us keyboard junkies was crucial.  It was driving me crazy not having it.

liuxiong332:
reply so fast!

chrishough:
lol, always online! basically a robot :fireworks:

PR #7354, opened 2015-06-21
add pane resize keyboard commands

liuxiong332:
This PR add commands `pane:shrink` and `pane:enlarge` to make it enable resize the panes with keyboard.

The following is the keymap binding:

| Command name | Default key binding | Comment |
| --- | :-: | --- |
| `pane:enlarge` | `cmd-alt-=` | enlarge the current pane by 10% |
| `panes:shrink` | `cmd-alt--` | shrink the current pane by 10% |

Closes https://github.com/atom/atom/issues/7332

PR #8357, opened 2015-08-13
Add pane increase/decrease size commands

kevinsawicki:
Adds two new pane commands, `pane:increase-size` and `pane:decrease-size` with some default keybindings on all platforms.

Continuation of #7354

Closes https://github.com/atom/atom/issues/7332

### DIFF
--- a/keymaps/linux.cson
+++ b/keymaps/linux.cson
@@ -143,6 +143,10 @@
   'ctrl-k ctrl-8': 'editor:fold-at-indent-level-8'
   'ctrl-k ctrl-9': 'editor:fold-at-indent-level-9'
 
+'atom-workspace atom-pane':
+  'ctrl-alt-=': 'pane:enlarge'
+  'ctrl-alt--': 'pane:shrink'
+  
 # allow standard input fields to work correctly
 'body .native-key-bindings':
   'ctrl-z': 'native!'
--- a/keymaps/linux.cson
+++ b/keymaps/linux.cson
@@ -144,9 +144,9 @@
   'ctrl-k ctrl-9': 'editor:fold-at-indent-level-9'
 
 'atom-workspace atom-pane':
-  'ctrl-alt-=': 'pane:enlarge'
-  'ctrl-alt--': 'pane:shrink'
-  
+  'ctrl-alt-=': 'pane:increase-size'
+  'ctrl-alt--': 'pane:decrease-size'
+
 # allow standard input fields to work correctly
 'body .native-key-bindings':
   'ctrl-z': 'native!'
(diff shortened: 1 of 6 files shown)


kevinsawicki:
@izuzak :eyes: any concerns about these keybindings? They seem pretty intuitive to me.

amfarrell:
As a user new to atom (coming from vim) I forked it to try out @liuxiong332's commits and they were pretty darn intuitive even though they don't match vim's :+1:

keymaps/linux.cson
```diff
@@ -143,6 +143,10 @@
   'ctrl-k ctrl-8': 'editor:fold-at-indent-level-8'
   'ctrl-k ctrl-9': 'editor:fold-at-indent-level-9'
 
+'atom-workspace atom-pane':
+  'ctrl-alt-=': 'pane:enlarge'
+  'ctrl-alt--': 'pane:shrink'
```
izuzak: Should these be `pane:increase-size` and `pane:decrease-size`, same as in the OSX keybindings?

keymaps/win32.cson
```diff
@@ -147,6 +147,10 @@
   'ctrl-k ctrl-8': 'editor:fold-at-indent-level-8'
   'ctrl-k ctrl-9': 'editor:fold-at-indent-level-9'
 
+'atom-workspace atom-pane':
+  'ctrl-alt-=': 'pane:enlarge'
+  'ctrl-alt--': 'pane:shrink'
```
izuzak: Should these be `pane:increase-size` and `pane:decrease-size`, same as in the OSX keybindings?

izuzak:
> @izuzak  any concerns about these keybindings? They seem pretty intuitive to me.

Yeah, they seem intuitive to me too :+1: 

Also, just tried searching if any other packages use these keybindings and found some, and they all use those keybindings for resizing panes:

https://github.com/search?utf8=%E2%9C%93&q=%22ctrl-alt--%22+extension%3Acson&type=Code&ref=searchresults

keymaps/win32.cson
kevinsawicki: Whoops, absolutely, thanks for catching that :green_heart:

kevinsawicki:
Thanks for this @liuxiong332 

And thanks @izuzak and @amfarrell for the feedback.

This will be included in the next Atom release, 1.0.8

Glutexo:
On Windows with Czech keyboard, because of this binding it is effectively impossible to write an asterisk `*`. It is written as `AltGr`+`-` or `Ctrl`+`Alt`+`-`. `-` is located in the lower-right corner next to the Right Shift key.

PR #8373, opened 2015-08-13
Make gutter dragging experience consistent across selection modes

nathansobo:
Refs #8350

Regardless of whether you're holding `shift` or `cmd`, the autoscroll behavior should be consistent.

@izuzak Can you try to break this?


izuzak:
> @izuzak Can you try to break this?

I tried and failed. :clap: :clap: 

And :zap: for cleaning this up, I believe I was the author of those `cmd`-click and `shift`-click handlers :sweat_smile:

PR #8377, opened 2015-08-13
Select screen rows when click-dragging in gutter

nathansobo:
/cc @izuzak for :microscope: 
### Before:

gutter-selection-before
### After:

gutter-selection-after

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -2073,123 +2073,123 @@ describe "TextEditorComponent", ->
         nextAnimationFrame()
 
       describe "when the gutter is clicked", ->
-        it "selects the clicked buffer row", ->
+        it "selects the clicked screen row", ->
           gutterNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenRowInGutter(1)))
-          expect(editor.getSelectedScreenRange()).toEqual [[0, 0], [2, 0]]
+          expect(editor.getSelectedScreenRange()).toEqual [[1, 0], [2, 0]]
 
       describe "when the gutter is meta-clicked", ->
-        it "creates a new selection for the clicked buffer row", ->
+        it "creates a new selection for the clicked screen row", ->
           editor.setSelectedScreenRange([[1, 0], [1, 2]])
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Nice! :gem:

izuzak:
Thanks for the ping, @nathansobo! Yeah, this is how things were before, but we intentionally switched to buffer rows in https://github.com/atom/atom/pull/5814 (which was a fix for https://github.com/atom/atom/issues/5738). 

I'm not strongly against this change, but just wanted to mention that I personally find it more intuitive if buffer rows are selected when you click in the gutter, and that's also the behavior in Sublime (as far as I can tell by reading my previous comments on this topic). I'm on a brand new laptop this week so don't have all the other editors installed -- would be great if someone could see how some other editors behave. And would be nice to hear from other folks what's more intuitive for them.

mehcode:
**Sublime**
When you click on a line number, it selects the _buffer row_.

When you click and drag to select text, it only highlights whole _buffer rows_ on the line number gutter but it selects one _screen row_ of text at a time when dragging up.

---

If we could mimic Sublime here I think we could solve both use cases nicely (smooth drag-to-select and click to select line).

izuzak:
Thanks for double-checking, @mehcode :heart: -- yeah, I'd be :+1: on that approach.

nathansobo:
@lee-dohm In https://github.com/atom/atom/pull/8350#issuecomment-131154926 other issue you said this:

> I would suggest using screen lines in the case of soft wrap, but in the case of folding, it should use buffer lines.

Can you expand a bit on this and get more specific about the behavior you'd prefer?

lee-dohm:
Absolutely, sorry I wasn't clear in the first place ...

If I click here:

napkin 08-14-15 9 23 08 am

I would expect it to only select from `Status]` through the end of that screen line. But if I go to a folded section of code and click here:

napkin 08-14-15 9 25 29 am

I would expect it to select the entire folded section in the buffer, not just the screen line.

nathansobo:
I've updated the PR to select the initially clicked *buffer row*, but then select screen rows on drag. It feels smooth but still intuitive on the initial click.

select-by-buffer-then-screen

izuzak:
Just took this for a test drive and it feels great. :zap: again @nathansobo

PR #8397, opened 2015-08-15
Add disableInitialEmptyEditor setting

### LINKED ISSUE #4573: Empty files after opening Atom
Every time after I open Atom the new empty "untitled" file is being created. Makes sense.
Problem is, that after I close Atom, this "untitled" file stays there and if I reopen Atom the number of "untitled" files is increasing.

screen shot 2014-12-12 at 13 20 17

jonathanrdelgado:
Adds the configuration option to disable the initial empty editor. [Many people](https://discuss.atom.io/t/preference-not-to-open-new-file-when-opening-atom/8476), myself included, find it annoying to constantly close these files.

Fixes #4573 by proxy, as well.


nathansobo:
I'd prefer to express the option in a positive way. `openEmptyEditorOnStart` or something similar, which we would default to `true`. Also, we'll need some sort of test coverage for this option, probably [here](https://github.com/atom/atom/blob/master/spec/integration/startup-spec.coffee).

jonathandelgado:
@nathansobo I made the changes you requested.

Can you please double check me on the spec? The only thing I am unsure of is the `.waitForPaneItemCount(0, 5000)`. I assume the spec will only pass if a pane never opens, but just want to confirm it doesn't trigger at initial start before a pane opens. Additionally, is `5000` too large of a value?

maxbrunsfeld:
@jonathandelgado Glad you added that integration spec! Those can be a little tricky to write, and unfortunately we don't run them on travis because they're resource-intensive.

Are you able to run them on your machine? Does that one fail reliably without your implementation changes, and pass reliably now?

jonathandelgado:
@maxbrunsfeld I actually don't have the ability to run them locally, I was relying on travis. On windows 10, I get a  `Cannot find module 'temp'` error which prevents the atom spec from even starting.

I assumed the tests ran on travis, so I'm unaware if they actually pass or not. Everything I wrote was based on existing integration tests, though.

Do you have the ability to test this? If not I can spin up a VM this weekend and make sure the spec passes.

maxbrunsfeld:
I'm running your branch on our internal CI now.

spec/integration/fixtures/atom-home/config-openEmptyEditorOnStart.cson
```diff
@@ -0,0 +1,6 @@
+"*":
+  welcome:
+    showOnStartup: false
+  "exception-reporting":
+    userId: "7c0a3c52-795c-5e20-5323-64efcf91f212"
+	openEmptyEditorOnStart: false
```
maxbrunsfeld: I think this just needs to be nested properly under `core`, like this:

```coffee
"*":
  core:
    openEmptyEditorOnStart: false

  # ...
```

maxbrunsfeld:
The test failed on CI, probably because of the error in your `fixtures` file that I commented on above.

jonathandelgado:
@maxbrunsfeld Great catch, sorry about that. Pushed a fix.

maxbrunsfeld:
Ok, it failed again on the same line. I'm not sure why. Can you take a crack at running them on a VM later? I can help you troubleshoot if you have problems getting those tests running.

jonathandelgado:
@maxbrunsfeld Absolutely, I'll take a crack at it. Thanks again!

jonathandelgado:
@maxbrunsfeld Spun up a VM today and worked on this. Built the latest Atom from source on a clean install of Ubuntu and I got 26 integration spec failures. I am assuming there are some discrepancies between the travis environment and a clean install of Ubuntu.

To be honest, not sure where to go from here on my side. Unless you are able to fix the spec, might just need to scrap this PR?

maxbrunsfeld:
Ok, just needed to tweak the test a little. Thanks so much for adding this!

jonathandelgado:
Good catch on that CSON, would have never thought of it. Thanks for the merge and your time! Appreciate it.

dotku:
I have already setup `disableInitialEmptyEditor: true` but there still is a untitled page on atom start.

Ben3eeE:
@dotku The setting name is `openEmptyEditorOnStart`. Which is true to open the untitled editor on start and false to not.

PR #8403, opened 2015-08-16
Add eol to word regex

dsandstrom:
I added the EOL symbol as an option for the word regex.  This will be used in atom/vim-mode#581.  Let me know if tests should be added.

### DIFF
--- a/src/cursor.coffee
+++ b/src/cursor.coffee
@@ -641,6 +641,8 @@ class Cursor extends Model
   # * `options` (optional) {Object} with the following keys:
   #   * `includeNonWordCharacters` A {Boolean} indicating whether to include
   #     non-word characters in the regex. (default: true)
+  #   * `includeEol` A {Boolean} indicating whether to include the end of the
+  #     line in the regex. (default: false)
   #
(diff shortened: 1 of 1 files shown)


src/cursor.coffee
```diff
@@ -641,15 +641,20 @@ class Cursor extends Model
   # * `options` (optional) {Object} with the following keys:
   #   * `includeNonWordCharacters` A {Boolean} indicating whether to include
   #     non-word characters in the regex. (default: true)
+  #   * `includeEol` A {Boolean} indicating whether to include the end of the
```
nathansobo: Can we call the option `includeEOL`? We standardized on all caps for initialisms for 1.0.

nathansobo:
Even a test that called the method with the option and tested the regex would be nice, just to make sure this doesn't regress in core.

dsandstrom:
For some reason, whenever I rebase a PR, I end up with double commits.  I'll submit a new one.

nathansobo:
Should we close this one then? Let me know when you've submitted the new one and correct the capitalization of the option.

PR #8420, opened 2015-08-18
Destroy a TextEditor's Gutters when it is destroyed

jssln:
It seems expected for custom `Gutter`s to be destroyed when its `TextEditor` is destroyed.
We decided that the `line-numbers` Gutter could not be deliberately `destroy`-ed, which is why there is a special case for it. 

The change in this PR also prevents an exception that would be thrown if you call `Gutter::destroy` after the TextEditor has been destroyed:
- `TextEditor::destroyed` calls `GutterContainer::destroy`
- ... which used to set the GutterContainer's `@gutters = null`.
- Then if you call `Gutter::destroy`, it calls `GutterContainer::removeGutter`
- ... which attempts to call `@gutters.indexOf`. But `@gutters` would be `null`, not an Array.

### DIFF
--- a/spec/gutter-container-spec.coffee
+++ b/spec/gutter-container-spec.coffee
@@ -50,3 +50,13 @@ describe 'GutterContainer', ->
       otherGutterContainer = new GutterContainer fakeOtherTextEditor
       gutter = new Gutter 'gutter-name', otherGutterContainer
       expect(gutterContainer.removeGutter.bind(null, gutter)).toThrow()
+
+  describe '::destroy', ->
+    it 'clears its array of gutters and destroys custom gutters', ->
+      newGutter = gutterContainer.addGutter {'test-gutter', priority: 1}
+      newGutterSpy = jasmine.createSpy()
+      newGutter.onDidDestroy(newGutterSpy)
+
+      gutterContainer.destroy()
+      expect(newGutterSpy).toHaveBeenCalled()
+      expect(gutterContainer.getGutters()).toEqual []
--- a/src/gutter-container.coffee
+++ b/src/gutter-container.coffee
@@ -13,7 +13,12 @@ class GutterContainer
     @emitter = new Emitter
 
   destroy: ->
-    @gutters = null
+    # Create a copy, because `Gutter::destroy` removes the gutter from
+    # GutterContainer's @gutters.
+    guttersToDestroy = @gutters.slice(0)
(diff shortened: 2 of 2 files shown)


spec/gutter-container-spec.coffee
```diff
@@ -50,3 +50,13 @@ describe 'GutterContainer', ->
       otherGutterContainer = new GutterContainer fakeOtherTextEditor
       gutter = new Gutter 'gutter-name', otherGutterContainer
       expect(gutterContainer.removeGutter.bind(null, gutter)).toThrow()
+
+  describe '::destroy', ->
+    it 'clears its array of gutters and destroys custom gutters', ->
+      newGutter = gutterContainer.addGutter {'test-gutter', priority: 1}
+      newGutterSpy = jasmine.createSpy()
+      newGutter.onDidDestroy(newGutterSpy);
```
MaximSokolov: Remove a trailing semicolon.

spec/gutter-container-spec.coffee
```diff
@@ -50,3 +50,13 @@ describe 'GutterContainer', ->
...
+
+      gutterContainer.destroy()
+      expect(newGutterSpy).toHaveBeenCalled();
```
MaximSokolov: Remove a trailing semicolon.

jssln:
Oops, sorry about the semicolons. 
The reason I can't destroy the `line-number` gutter is because it throws: https://github.com/atom/atom/blob/master/src/gutter.coffee#L27

bolinfest:
Possibly related: https://github.com/atom/atom/pull/8243

jssln:
@bolinfest That's a similar type of issue but shouldn't be affected by this PR.

jssln:
looks like I have fix some tests

MaximSokolov:
```
Installing one-dark-syntax@1.1.0 ✗
Request for package information failed: atom.io is temporarily unavailable, please try again later.
```
```
Unable to download https://www.atom.io/api/packages/atom-dark-syntax/versions/0.27.0/tarball: 503
```
I dont't think test have failed bacause of your PR.

jssln:
There are some failing TextEditor and TextEditorPresenter tests.

nathansobo:
:+1: Once this builds cleanly.

jssln:
Adjusted to fix failing tests.

nathansobo:
Appreciate it @jssln!

PR #8431, opened 2015-08-18
Extend ViewRegistry with support for objects with an element property

thomasjo:
By adding this extension to `ViewRegistry::getView` we're paving the way for [Etch](https://github.com/nathansobo/etch)-like view frameworks which promotes the usage of plain objects and classes with an `element` property which is an instance of `HTMLElement`.

/cc @nathansobo @maxbrunsfeld

### DIFF
--- a/spec/view-registry-spec.coffee
+++ b/spec/view-registry-spec.coffee
@@ -26,6 +26,14 @@ describe "ViewRegistry", ->
         expect(node.textContent).toBe "Hello"
         expect(node.spacePenView).toBe view
 
+    describe "when passed an object with an element property", ->
+      it "returns the element property if it's an instance of HTMLElement", ->
+        class TestComponent
+          constructor: -> @element = document.createElement('div')
+
+        component = new TestComponent
+        expect(registry.getView(component)).toBe component.element
+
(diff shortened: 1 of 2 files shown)


spec/view-registry-spec.coffee
```diff
@@ -26,6 +26,14 @@ describe "ViewRegistry", ->
         expect(node.textContent).toBe "Hello"
         expect(node.spacePenView).toBe view
 
+    describe "when passed an object with an element property", ->
+      it "returns the element property if it's an instance of HTMLElement", ->
```
thomasjo: Not in love with the naming here, so I'm open for suggestions to improvements. /cc @atom/feedback

maxbrunsfeld:
:+1: Looks good to me.

nathansobo:
It would be good to also document this somewhere.

thomasjo:
> It would be good to also document this somewhere

I can take care of that; any thoughts on where? Expand `getView` docs maybe?

nathansobo:
> Expand getView docs maybe?

That sounds good. Thanks!

maxbrunsfeld:
Thanks @thomasjo!

Issue #8416, opened 2015-08-17
Support Babel in all of the same ways as CoffeeScript


maxbrunsfeld:
We want to move toward writing new code in javascript, not coffeescript. For example: https://github.com/atom/line-ending-selector. But before we start bundling packages written in Babel, we need to improve some infrastructure around it.

### Tasks

[The CI script](https://github.com/atom/ci/blob/master/build-package.sh)
* [x] Make sure it runs specs properly for packages that use babel
* [x] Somehow detect when packages are using [Standard Style](http://standardjs.com/) and lint them properly

Bundled Packages
* [x] Make the [build script](https://github.com/atom/atom/blob/master/build/tasks/build-task.coffee) treat babel source files the same as coffee-script files
* [x] Make the Gruntfile compile babel files

Opting in to Babel
* [x] Create an alternative to the `'use babel'` string that fits in better with other babel directives. Maybe this:

```js
/** use babel */
/** @jsx dom */
```

nathansobo:
I'm unsure about the wording of the comment header... I see where you're going with `@atom` but I also think it's just a declaration that Babel is being used for the current document, not necessarily specific to Atom. But I suppose there are a bunch of options to Babel that would need to be specified if this were to be a general flag... I guess it would be nice if those could be enabled by their own directives.

Honestly just supporting `/** use babel */` might be enough to blend in more cleanly with their existing header syntax.

mnquintana:
Just a thought - do we need to have an opt-in for Babel? Would there be any significant harm in passing all JS through the Babel parser? In principle I get why we would want to avoid any unnecessary compilation, but if we're really considering making developing with Babel *feel* more like developing with Coffee, then it would be nice from a "user" perspective to be able to omit the pragma.

nathansobo:
@mnquintana My main concern is the overhead of always assuming the need for Babel for every single JS file, both in terms of performance and complexity. I haven't fully thought through the implications of dropping the pragma, but if you or anyone else wants to do an impact analysis on it, I'd be open to discussing it. It just has to be as close to free as possible.

maxbrunsfeld:
> `/** use babel */` might be enough to blend in

Yeah, that probably makes the most sense; use the same wording as the other style.

paulcbetts:
electron-compile uses `'use nobabel'` as its directive. One thing you can do is modify your Babel compile-cache to look more like this, where we early-exit from any path involving `node_modules`:

[...]

PR #8433, opened 2015-08-18
Support bundling babel packages and bundle the line-ending-selector package 

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/8416


build/lib/uses-babel.coffee
```diff
@@ -0,0 +1,18 @@
+fs = require 'fs'
+
+BABEL_PREFIXES = [
+	"'use babel'"
```
kevinsawicki: Super minor, but this looks like a hard tab (`\t`) here

build/lib/uses-babel.coffee
```diff
@@ -0,0 +1,18 @@
...
+PREFIX_LENGTH = Math.max(BABEL_PREFIXES.map((prefix) -> prefix.length)...)
+
+buffer = Buffer(PREFIX_LENGTH)
+
+module.exports = (filename) ->
+  file = fs.openSync(filename, 'r')
+  fs.readSync(file, buffer, 0, PREFIX_LENGTH)
+  fs.closeSync(file)
+  BABEL_PREFIXES.some (prefix) ->
+    prefix.equals(buffer.slice(0, prefix.length))
```
kevinsawicki: Should this be `prefix is buffer.slice(0, prefix.length)`? is `equals` a String method? Not familiar with it.

kevinsawicki: Oh, they are buffers, nevermind.

build/Gruntfile.coffee
```diff
@@ -77,6 +80,11 @@ module.exports = (grunt) ->
       dest: appDir
       ext: '.js'
 
+  babelConfig =
+    options: {}
```
kevinsawicki: Should this include any of the options in the `defaultOptions` object from `src/babel.coffee`? Not sure what the default options are for source maps.

build/lib/uses-babel.coffee
maxbrunsfeld: Oh damn, thanks. I had turned of `softTabs` for testing something, and was accidentally sprinkling hard tabs everywhere.

build/Gruntfile.coffee
maxbrunsfeld: :+1: Yeah

maxbrunsfeld:
@kevinsawicki So I'm compiling the files *in place* at the moment, so that the `main` field in the package's `package.json` doesn't have to change. Is that going to cause problems when you re-build though, because the file gets overridden?

kevinsawicki:
> So I'm compiling the files in place at the moment, so that the main field in the package's package.json doesn't have to change

So they are changing the files in the `node_modules` that sits inside the clone atom/atom repository?

Could a path relative to `appDir` be specified as the `dest` value when you are pushing to the babel config array?

maxbrunsfeld:
> Could a path relative to appDir be specified as the dest value when you are pushing to the babel config array?

:+1: Oh yeah, of course. I didn't realize that the `compile` was already writing to the app dir. For some reason I thought that happened as a separate step.

build/Gruntfile.coffee
maxbrunsfeld: Ok, so I just grabbed the data by requiring `src/babel`, which is a bit weird. Do you think I should extract those options into some separate file, or just leave it like that for now?

kevinsawicki: > Do you think I should extract those options into some separate file, or just leave it like that for now?

I think leaving it for now is fine.

ssorallen:
`/** @babel */` seems more appropriate than using `use`.

Turning `'use babel;` into `/** use babel */` mixes two ideas. 'use' is a keyword for [JavaScript Pragmas](http://www-archive.mozilla.org/js/language/js20-2002-04/core/pragmas.html) that can be used in '`use strict;'` to enforce strict mode within a given scope. `use` in the comment alludes to some use of actual Pragmas, but this comment is ignored by the JS interpreter.

This also follows the late JSX comment `/** @jsx ... */`, which did not have the `use` keyword.

nathansobo:
I'm down with `/** @babel */`

thomasjo:
> I'm down with /** @babel */

:+1: from me too. This also leaves the door open for potentially allowing developers to set some parameters as well. Not that we ever need to, but it makes a lot more sense with `/** @babel ... */`.

nathansobo:
Thanks for the suggestion @ssorallen.

maxbrunsfeld:
@ssorallen :+1:

maxbrunsfeld:
Ok, I think this is ready to go. @kevinsawicki can you give this some final :eyes:?

build/Gruntfile.coffee
```diff
@@ -1,6 +1,9 @@
 fs = require 'fs'
 path = require 'path'
 os = require 'os'
+glob = require 'glob'
+usesBabel = require './lib/uses-babel'
+babelOptions = require('../static/babelrc.json')
```
kevinsawicki: `.json` isn't needed here

src/typescript.js
```diff
@@ -0,0 +1,53 @@
...
+    var version = require('typescript-simple/package.json').version
+    typescriptVersionDir = path.join('ts', createVersionAndOptionsDigest(version, defaultOptions))
+  }
+
+  return path.join(
+    typescriptVersionDir,
+    crypto
+      .createHash('sha1')
+      .update(sourceCode, 'utf8')
+      .digest('hex') + ".js"
```
kevinsawicki: Might be nice to use consistent quotes on this line, looks like single is the pattern in this file

src/compile-cache.js
```diff
@@ -0,0 +1,131 @@
...
+}
+
+exports.addPathToCache = function (filePath, atomHome) {
+  this.setAtomHomeDirectory(atomHome)
+  var extension = path.extname(filePath)
+
+  if (extension === '.cson') {
+    CSON.readFileSync(filePath)
+  } else {
+    var compiler = COMPILERS[extension]
```
kevinsawicki: Was there a reason not to follow this pattern for `.cson` files as well?

src/compile-cache.js
```diff
@@ -0,0 +1,131 @@
...
+    while ((match = INLINE_SOURCE_MAP_REGEXP.exec(fileData))) {
+      lastMatch = match
+    }
+    if (lastMatch == null){
+      return null
+    }
+
+    var sourceMappingURL = lastMatch[1]
+    var rawData = sourceMappingURL.slice(sourceMappingURL.indexOf(',') + 1)
+    var sourceMap = JSON.parse(new Buffer(rawData, 'base64').toString())
```
kevinsawicki: JSON.parse handles Buffers, the `toString` isn't needed.

src/compile-cache.js
```diff
@@ -0,0 +1,131 @@
...
+    if (fileData == null) {
+      return null
+    }
+
+    var match, lastMatch
+    INLINE_SOURCE_MAP_REGEXP.lastIndex = 0
+    while ((match = INLINE_SOURCE_MAP_REGEXP.exec(fileData))) {
+      lastMatch = match
+    }
+    if (lastMatch == null){
```
kevinsawicki: Missing space before `{`

src/compile-cache.js
```diff
@@ -0,0 +1,131 @@
...
+var INLINE_SOURCE_MAP_REGEXP = /\/\/[#@]\s*sourceMappingURL=([^'"\n]+)\s*$/mg
+
+require('source-map-support').install({
+  handleUncaughtExceptions: false,
+
+  // Most of this logic is the same as the default implementation in the
+  // source-map-support module, but we've overridden it to read the javascript
+  // code from our cache directory.
+  retrieveSourceMap: function (filePath) {
+    if (!fs.isFileSync(filePath)){
```
kevinsawicki: Missing space before `{`

src/compile-cache.js
```diff
@@ -0,0 +1,131 @@
+'use strict'
+
+var path = require('path')
+var CSON = require('season')
+var fs = require('fs-plus')
+var _ = require('underscore-plus')
```
kevinsawicki: It looks like this file is required before the module cache is set up in `index.js`.

That means any libraries required here and downstream won't pass through it. It would be nice to address that so there aren't any performance regressions.

src/browser/main.coffee
```diff
@@ -77,14 +77,9 @@ setupAtomHome = ->
...
-setupCoffeeCache = ->
-  CoffeeCache = require 'coffee-cash'
-  cacheDir = path.join(process.env.ATOM_HOME, 'compile-cache')
-  # Use separate compile cache when sudo'ing as root to avoid permission issues
-  if process.env.USER is 'root' and process.env.SUDO_USER and process.env.SUDO_USER isnt process.env.USER
-    cacheDir = path.join(cacheDir, 'root')
-  CoffeeCache.setCacheDirectory(path.join(cacheDir, 'coffee'))
-  CoffeeCache.register()
+setupCompileCache = ->
+  compileCache = require('../compile-cache')
```
kevinsawicki: Just curious, how long does this require take in the browser process?  Just want to make sure the browser process load time is minimally impacted.

spec/atom-reporter.coffee
```diff
@@ -15,8 +13,8 @@ formatStackTrace = (spec, message='', stackTrace) ->
   for line in stackTrace.split('\n')
     convertedLines.push(line) unless jasminePattern.test(line)
     break if firstJasmineLinePattern.test(line)
+  stackTrace = convertedLines.join('\n')
 
-  stackTrace = convertStackTrace(convertedLines.join('\n'), sourceMaps)
   lines = stackTrace.split('\n')
```
kevinsawicki: Looks like the `convertedLines` are joined and then immediately split again now on the next line.

spec/compile-cache-spec.coffee
```diff
@@ -1,39 +0,0 @@
-path = require 'path'
-CSON = require 'season'
-CoffeeCache = require 'coffee-cash'
-
-babel = require '../src/babel'
-typescript = require '../src/typescript'
-CompileCache = require '../src/compile-cache'
-
-describe "Compile Cache", ->
-  describe ".addPathToCache(filePath)", ->
```
kevinsawicki: Where are these covered now?

src/babel.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+var _ = require('underscore-plus')
+var crypto = require('crypto')
+var path = require('path')
+var defaultOptions = require('../static/babelrc.json')
```
kevinsawicki: It would be nice to inline this into Atom's root mega `package.json` file since all other `.json` and `.cson` files are currently placed there instead of bundled separately.

src/coffee-script.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
```
kevinsawicki: Are there specs for this? I would imagine something similar to the babel/typescript ones

src/compile-cache.js
maxbrunsfeld: :+1: Good call. It seems like this would be addressed if we did the following:

1. Defer the require of `coffee-script` like we talked about
2. Move the setup for CSON back *out* of `compile-cache.js`, and into `index.js` after the `ModuleCache` is set up
3. Don't use `_.each` in `compile-cache`

That way, no new libraries will be required before the module cache. What do you think?

src/compile-cache.js
maxbrunsfeld: We had to change the code for typescript, coffee and babel to make source maps work properly. It would have been nice to refactor the CSON code to use the same pattern, but there wasn't a pressing need to touch `season`, so we left it as is for now.

src/babel.js
kevinsawicki: After thinking about this more, it probably isn't worth the effort/complexity, feel free to disregard.

src/compile-cache.js
kevinsawicki: > What do you think?

Sounds good.

spec/atom-reporter.coffee
maxbrunsfeld: :see_no_evil: Thanks; I was kind of blindly hacking away there.

src/coffee-script.js
maxbrunsfeld: I figured this was tested via the fact that the test suite loaded at all. It seemed kinda weird to add a coffee-script spec that tests that another coffee-script file can be required.

kevinsawicki: Yeah, would be nice though to somewhere test that cache files are written and files with the same content are only compiled once.

Basically, what coffeestack was testing now that that logic is being done here.

package.json
```diff
@@ -20,9 +20,7 @@
     "babel-core": "^5.8.21",
     "bootstrap": "^3.3.4",
     "clear-cut": "^2.0.1",
-    "coffee-cash": "0.8.0",
```
kevinsawicki: This is going to require a change in https://github.com/atom/timecop/blob/master/lib/cache-panel-view.coffee to get timecop working again.

src/compile-cache.js
```diff
@@ -0,0 +1,141 @@
...
+
+  // Most of this logic is the same as the default implementation in the
+  // source-map-support module, but we've overridden it to read the javascript
+  // code from our cache directory.
+  retrieveSourceMap: function (filePath) {
+    if (!fs.isFileSync(filePath)) {
+      return null
+    }
+
+    var sourceCode = fs.readFileSync(filePath, 'utf8')
```
kevinsawicki: Are read errors handled internally by `source-map-support` or would a read error here cause a notification?

package.json
maxbrunsfeld: Woah thanks for the heads up. PR for that here: https://github.com/atom/timecop/pull/13

src/coffee-script.js
maxbrunsfeld: Ok, I think the caching is tested through the compile-cache spec.

kevinsawicki: Awesome :+1:

src/compile-cache.js
maxbrunsfeld: Ok, I'm just gonna let it behave like there is no source map, in the event that there's an IO error retrieving the source map.

src/browser/main.coffee
maxbrunsfeld: On my MBP:

```sh
node -p 't1 = new Date().getTime(); require("coffee-cash"); new Date().getTime() - t1'
24
node -p 't1 = new Date().getTime(); require("./src/compile-cache"); new Date().getTime() - t1'
38
```

So there's a 16ms increase without asar. I would expect a smaller number when using the asar bundle. In my opinion, we should try to get that down, but not by maintaining what is now a completely duplicated code-path. We should just convert the browser-side code to JS at some point soon and eliminate that whole chunk of time.

kevinsawicki: >  In my opinion, we should try to get that down, but not by maintaining what is now a completely duplicated code-path.

Yeah, I'm totally on board with de-duplication but there were only a couple lines of duplication here between the browser/render processes right? Just the part that sets the cache directory path and checks for the sudo case?

A 28ms startup time increase (14ms on the browser side, 14ms presumably on the render side) does seem a bit excessive and something to possibly investigate further before merging.

maxbrunsfeld: > 14ms presumably on the render side

I would expect close to zero difference on the render side, because `babel.js`, `typescript.js`, `coffee-stack`, `coffee-cash` were all previously eager-loaded by `index.js`. This adds `source-map-support` but removes 2 other npms (`coffee-stack` and `coffee-cash`).

maxbrunsfeld: Profiling startup w/ timecop now...

kevinsawicki: > I would expect close to zero difference on the render side, because

Ah yeah, good point

maxbrunsfeld: Ok, I see identical numbers in time-cop for every parameter (`Window load`, `Shell load`, `Workspace load`...). Are you ok :ship:ing this like this?

kevinsawicki: > I see identical numbers on this branch and master for every time-cop

Great to hear :dash: 

> Are you ok :ship:ing this like this?

:shipit:

nathansobo:
:zap: Nice work.

:sunrise: The era of Babel is dawning. :sunrise:

thomasjo:
Bloody brilliant! Thank you so much for doing this @maxbrunsfeld! 

> :sunrise: The era of Babel is dawning. :sunrise: 

Oooooh yeah :smiling_imp:

kevinsawicki:
Great cleanup here as well, really nice to see this whole system level up :green_heart: :blue_heart:

PR #8436, opened 2015-08-19
Batch cursor updates

as-cii:
While working on character measurement, I spotted a couple of _cursor rendering computations_ that were not batched yet.

This PR fixes that, so that when a cursor changes (e.g. gets added, removed, or moves) we wait until the next frame to update it. This is consistent with how we handle other updates in the presenter and should as well yield some speedups when many cursors are added/removed/moved at once.

Any feedback is very welcome and much appreciated! :eyes: :bow: 

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -410,10 +410,7 @@ class TextEditorPresenter
     @updateCursorState(cursor) for cursor in @model.cursors # using property directly to avoid allocation
     return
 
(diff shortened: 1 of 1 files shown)


nathansobo:
Batching batching everywhere. Thanks @as-cii.

as-cii:
> Batching batching everywhere

:joy: 

I have tested this for a day now and it seems to work properly.

Thanks @nathansobo!

Issue #5412, opened 2015-02-05
Atom fails to fully start


nickelpro:
Arch Linux, latest git

Atom dies on something, spits this out in chrome console

```
TypeError: Arguments to path.join must be strings
  at Object.posix.join (path.js:472:13)
  at Object.<anonymous> (/usr/share/atom/resources/app/src/coffee-cache.js:14:19)
  at Object.<anonymous> (/usr/share/atom/resources/app/src/coffee-cache.js:106:4)
  at Module._compile (module.js:475:26)
  at Object.loadFile [as .js] (/usr/share/atom/resources/app/src/6to5.js:138:21)
  at Module.load (/usr/share/atom/resources/app/node_modules/coffee-script/lib/coffee-script/register.js:45:36)
  at Function.Module._load (module.js:325:12)
  at Module.require (module.js:380:17)
  at require (module.js:399:17)
  at window.onload (file:///usr/share/atom/resources/app/static/index.js:55:5)
```

stramel:
I am running into the same thing when building from the latest git.

**OS:** Ubuntu 14.04
**Atom:** master branch (83c8d904cf)

It seems to fail off on the
```console
$ sudo script/grunt install
```

when running in verbose mode:
```console
$ sudo script/grunt install -v
...
Writing /usr/local/share/atom/version...OK
Copied /tmp/atom-build/Atom to /usr/local/share/atom.
Reading package.json...OK
Parsing package.json...OK
Warning: executable is not defined Use --force to continue.

Aborted due to warnings.
```

When building from the latest tag, there is no issue.

thomasjo:
@stramel Your issue is not the same as reported by @nickelpro; his problem is a runtime issue, whereas yours is related to installation. I would suggest you open a separate issue for your problem.

stramel:
@thomasjo When running `--force` on the `install` task, I get the exact same error message. But I will go ahead and open it in a separate issue.

kevinsawicki:
@nickelpro in the dev tools console tab can you run `process.env.ATOM_HOME` and see what the value is?

jbt:
I'm seeing this as well, ubuntu 14.04, for me `process.env.ATOM_HOME` is `undefined`

stramel:
@kevinsawicki I am now seeing the same issue. 

```console
> process.env.ATOM_HOME
< undefined
```

kevinsawicki:
Sorry about this one, this should be fixed on master now.

[...]

PR #8442, opened 2015-08-19
Enable Portable Mode

### LINKED ISSUE #2939: Add portable version of Atom

anaisbetts:
This PR sets the default ATOM_HOME to be relative to `atom.exe`, if the following cases are true:
1. We're not in DevMode
2. We're on Windows
3. The EXE path is not in the normal installed location

This allows users to take the entire Atom folder and use it as a "Portable" application (i.e. portableapps.com)

Fixes #2939
## TODO:
- [x] Make sure this actually works
- [ ] ~~Add `.alwaysportable` directive~~ (Gonna see if people want this or not)
- [ ] Code Review


thomasjo:
Looks promising :8ball:

static/index.js
```diff
@@ -103,6 +103,19 @@ var setupCoffeeCache = function(cacheDir) {
...
+  // No portable mode on non-Windows
+  if (process.platform !== 'win32') return false;
+  
+  // DevMode? Nope
+  var devMode = loadSettings.devMode || !loadSettings.resourcePath.startsWith(process.resourcesPath + path.sep);
+  if (devMode) return false;
+  
+  // Compare our EXE's path to where it would normally be in an installed app
+  var ourPath = process.execPath.toLowerCase();
+  return (ourPath.indexOf(process.env.USERPROFILE.toLowerCase()) === 0);
```
kevinsawicki: Should this comparison include a trailing slash on `USERPROFILE`?

Currently this check will pass for the following `process.execPath`/`USERPROFILE` right?

`process.execPath = C:\Users\KevinSawicki\atom\atom.exe`
`USERPROFILE = C:\Users\Kevin`

static/index.js
```diff
@@ -103,6 +103,19 @@ var setupCoffeeCache = function(cacheDir) {
   CoffeeCache.register();
 }
 
+var isPortableMode = function() {
+  // No portable mode on non-Windows
+  if (process.platform !== 'win32') return false;
+  
+  // DevMode? Nope
+  var devMode = loadSettings.devMode || !loadSettings.resourcePath.startsWith(process.resourcesPath + path.sep);
```
kevinsawicki: `loadSettings` might be null, see `isRunningCoreSpecs` for an example of guards typically around it.

kevinsawicki:
So the idea here is that is they install to `USERPROFILE` then it is portable? And a squirrel installer never goes there in any cases?

kevinsawicki:
> So the idea here is that if they install to USERPROFILE then it is portable?

Nevermind, it is the other way around right? **Not** in `USERPROFILE` means portable mode?

static/index.js
paulcbetts: This is a typo that I fixed but didn't push apparently, this should be `LOCALAPPDATA`, which makes *way* more sense :)

paulcbetts:
> So the idea here is that is they install to USERPROFILE then it is portable? And a squirrel installer never goes there in any cases?

Peep the line comment above, things will make way more sense then

static/index.js
kevinsawicki: Yeah, totally :+1:

static/index.js
paulcbetts: :+1:, will fix

lee-dohm:
Something I thought of because someone was asking about installing to the `D:\` drive the other day. If someone installs applications to a non-standard directory for disk space reasons, it will automatically start behaving as a portable application?

paulcbetts:
@lee-dohm Correct, but because this person probably (maybe?) already has an existing `.atom` folder in their home directory, we'll prefer that over the portable path

batjko:
I don't think this will satisfy those people who keep waiting for a portable version.
Portable should be entirely independent, by default always read and write its settings and plugins only from the installation directory (wherever was chosen by the installing user) - entirely independent of what is already installed or not. It shouldn't care at all if there are already 50 other Atoms installed, and only use its own .atom folder.

This conditional checking of already installed versions and preferring the "remote" .atom folder over its own... it just brings us back to the same situation that exists currently.

What people keep asking for is a separate version of Atom. One that, as described, doesn't need to be installed (it is already runnable from the folder people unzipped), and which doesn't care about where it runs from or whether there are other versions installed elsewhere on the system.


I don't need a portable version myself, but just wanted to point that out.

paulcbetts:
@batjko This might be true, but then that gets us into creating separate builds for Portable Windows which sucks. I can add a check for an .alwaysPortable dot-file that makes us always ignore ~/.atom

batjko:
> but then that gets us into creating separate builds for Portable Windows which sucks

True, unless the portable version is cross-platform, but it would still need to be maintained separately.

> I can add a check for an .alwaysPortable dot-file that makes us always ignore ~/.atom

That sounds like a pretty nifty way to do it. Generally, if an `.alwaysPortable` file exists, rely only on your local install folder.

paulcbetts:
> I can add a check for an .alwaysPortable dot-file that makes us always ignore ~/.atom

Let's punt on this for now and see if people ask for it

paulcbetts:
@kevinsawicki Plz to review!

src/browser/main.coffee
```diff
@@ -56,11 +56,24 @@ handleStartupEventWithSquirrel = ->
 setupCrashReporter = ->
   crashReporter.start(productName: 'Atom', companyName: 'GitHub')
 
-setupAtomHome = ->
+isPortableInstall = (args) ->
+  return false unless process.platform is 'win32'
+  return false unless (process and process.type)
```
kevinsawicki: I don't think you need the `process and` check here, since the line before already calls `process.platform` without checking `process`.

kevinsawicki:
Will this work if they have installed Atom directly to Program Files via downloading `atom-windows.zip`?

src/browser/main.coffee
```diff
@@ -56,11 +56,24 @@ handleStartupEventWithSquirrel = ->
...
+  return false if args.devMode or args.testMode
+
+  ourPath = process.execPath.toLowerCase()
```
kevinsawicki: Could you add a comment above this line that just adds some example paths and says which ones are considered portable and which ones are not?

src/browser/main.coffee
paulcbetts: This is copy-pasta from the apm version, will fix

paulcbetts:
> Will this work if they have installed Atom directly to Program Files via downloading atom-windows.zip

No, because we'll try to write to `C:\Program Files\Atom\.atom` which will fail because of no permissions

src/browser/main.coffee
PikachuEXE: What is the use of checking `process.type` here?
I am newbie :P

paulcbetts: @PikachuEXE No use in this case, copy pasta from apm

src/browser/main.coffee
```diff
@@ -57,11 +57,14 @@ handleStartupEventWithSquirrel = ->
...
-setupAtomHome = ->
+setupAtomHome = (args) ->
   return if process.env.ATOM_HOME
   atomHome = path.join(app.getHomeDir(), '.atom')
-  try
-    atomHome = fs.realpathSync(atomHome)
+  AtomPortable = require './atom-portable'
+  atomHome = AtomPortable.portableAtomHomePath() if AtomPortable.isPortableInstall process.platform, process.env.ATOM_HOME
+  atomHome = fs.realpathSync(atomHome)
+  fs.statSync(atomHome)
```
kevinsawicki: Is this line needed? The return value doesn't appear to be used

src/browser/main.coffee
```diff
@@ -57,11 +57,14 @@ handleStartupEventWithSquirrel = ->
...
 
-setupAtomHome = ->
+setupAtomHome = (args) ->
   return if process.env.ATOM_HOME
   atomHome = path.join(app.getHomeDir(), '.atom')
-  try
-    atomHome = fs.realpathSync(atomHome)
+  AtomPortable = require './atom-portable'
+  atomHome = AtomPortable.portableAtomHomePath() if AtomPortable.isPortableInstall process.platform, process.env.ATOM_HOME
+  atomHome = fs.realpathSync(atomHome)
```
kevinsawicki: I think `fs.realpathSync` throws an error when it doesn't exist so I think we might still want to have a `try` statement around it and just ignore the error when it doesn't exist.

src/browser/atom-portable.coffee
```diff
@@ -0,0 +1,15 @@
+fs = require 'fs-plus'
+path = require 'path'
+
+module.exports =
+class AtomPortable
+  @portableAtomHomePath: ->
+    execDirectoryPath = path.dirname(process.execPath)
+    return path.join(execDirectoryPath, "../.atom")
```
kevinsawicki: I think we usually write this as `path.join(execDirectoryPath, '..', '.atom')` so that the `/` isn't hard-coded

paulcbetts:
I'll try to have a look at this at the weekend

sketchpunk:
is it a big deal to make the app portable first, remote second? for apps i've made it check if a folder exists locally, if it doesn't then mkdir. if it errors because of permission then you make the folder in appdata. we sell our apps and some clients dont have permissions to install things on their work computer. been years, its easy and a single function that drives all our apps. i tried to make atom portable a year ago and the only issue i saw was lots of duplication of code to determine where .atom existed instead of having a single function called from the whole codebase.

iliyang:
I think making Atom portable (in the way a Windows user expects a portable app to behave) is as simple as making `atom.exe` check if there is an `.atom` or `.home` directory next to it. If there is, then simply create/use the user profile there.

No need to make a special portable version of Atom, and no need to have an `.alwaysPortable` file. The Atom installer can simply ask the user if he/she wants a portable installation. And creating this portable installation simply means creating an empty `.atom` or `.home` directory next to `atom.exe`. If creating/writing to such a directory requires admin privileges (because the installation location is protected), the installer will ideally know this (not sure how exactly) and warn the user.

paulcbetts:
> I think making Atom portable (in the way a Windows user expects a portable app to behave) is as simple as making atom.exe check if there is an .atom or .home directory next to it. If there is, then simply create/use the user profile there.

Yep, that's what this mostly does

> No need to make a special portable version of Atom, and no need to have an .alwaysPortable file. 

Most people won't need to use the .alwaysPortable file, it's more for superusers that want to have >1 Atom profile (i.e. they want to use the portable one *and* have an installed one, and keep them separate)

iliyang:
Thanks for the quick response @paulcbetts ! I'm eagerly waiting for this portable version, as I'm sure quite a lot of other people are.

Sisyphuss:
My laboratory uses Debian and only installs Debian-official packages. Since I have no superuser authority, I can not use Atom in my laboratory. Any solution?

spec/atom-portable-spec.coffee
```diff
@@ -0,0 +1,71 @@
+path = require "path"
+fs = require 'fs-plus'
+temp = require "temp"
+rimraf = require "rimraf"
+AtomPortable = require "../src/browser/atom-portable"
+
+describe "Set Portable Mode", ->
+  portableAtomHomePath = path.join(path.dirname(process.execPath), "../.atom").toString()
```
kevinsawicki: `path.join` always returns a string so I don't think you need the `toString` here.

static/index.js
```diff
@@ -80,24 +77,6 @@
     require('ipc').sendChannel('window-command', 'window:loaded')
   }
 
-  function setupAtomHome () {
```
kevinsawicki: I should have left a comment in here originally about this but I think this is still needed on Linux based on this comment: https://github.com/atom/atom/issues/5412#issuecomment-73119795

Maybe we can pass `atomHome` as a load setting and set it as an env var still which would remove the code duplication.

raelyard:
Sample warning if portable home directory not writable (just noticed it's missing a (back)slash) - What do you think of this?
portablehomewritefailed

static/index.js
```diff
@@ -14,6 +14,8 @@
       })
 
       // Ensure ATOM_HOME is always set before anything else is required
+      // This is because of a difference in Linux not inherited between browser and render processes
+      // issue #5142
```
kevinsawicki: :heart:

paulcbetts:
@raelyard :+1:

raelyard:
Updated screenshot:
portablehomewritefailed

raelyard:
@thomasjo Thanks, that is a big improvement - needed to also add an escape to the dash in the character set.

kevinsawicki:
@raelyard it looks like this branch no longer merges cleanly with master, can you merge or rebase master? Thanks.

src/browser/main.coffee
```diff
@@ -57,11 +57,14 @@ handleStartupEventWithSquirrel = ->
 setupCrashReporter = ->
   crashReporter.start(productName: 'Atom', companyName: 'GitHub')
 
-setupAtomHome = ->
+setupAtomHome = (args) ->
```
kevinsawicki: It looks like args is no longer used here so would you mind backing out this change here?

raelyard:
@kevinsawicki merged and now I can't get the app to run.  Checked out master, seeing the same thing.  Something amiss, I think, with master.  Trying to understand what.

src/browser/main.coffee
raelyard: done

raelyard:
@kevinsawicki merge complete and stupid mistakes resolved

src/browser/atom-portable.coffee
```diff
@@ -0,0 +1,32 @@
...
+
+  @isPortableInstall: (platform, environmentAtomHome, defaultHome) ->
+    return false unless platform is 'win32'
+    return false if environmentAtomHome
+    return false if not fs.existsSync(@getPortableAtomHomePath())
+    # currently checking only that the directory exists  and is writable,
+    # probably want to do some integrity checks on contents in future
+    @portableAtomHomePathWritable(defaultHome)
+
+  @portableAtomHomePathWritable: (defaultHome) ->
```
kevinsawicki: Could you rename this to `isPortableAtomHomePathWritable` to match the style of `isPortableInstall`?

src/browser/atom-portable.coffee
```diff
@@ -0,0 +1,32 @@
...
+    message = ""
+    try
+      writePermissionTestFile = path.join(@getPortableAtomHomePath(), "write.test")
+      fs.writeFileSync(writePermissionTestFile, "test") if not fs.existsSync(writePermissionTestFile)
+      fs.removeSync(writePermissionTestFile)
+      writable = true
+    catch error
+      message = "Failed to use portable Atom home directory (#{@getPortableAtomHomePath()}).  Using the default instead (#{defaultHome}).  #{error.message}"
+
+    ipc.on 'check-portable-home-writable', (event, arg) ->
```
kevinsawicki: `arg` appears to be unused here so it looks like it is save to just remove it.

src/browser/main.coffee
```diff
@@ -12,6 +12,8 @@ yargs = require 'yargs'
 console.log = require 'nslog'
 
 start = ->
+  args = parseCommandLine()
```
kevinsawicki: Does this need to be moved here in this pull request? Or could it go back to under the `app.setAppUserModelId('com.squirrel.atom.atom')` line?

raelyard:
@kevinsawicki addressed line comments

src/browser/main.coffee
```diff
@@ -18,7 +18,7 @@ start = ->
 
   # NB: This prevents Win10 from showing dupe items in the taskbar
   app.setAppUserModelId('com.squirrel.atom.atom')
-
+
```
kevinsawicki: :fire: looks like some trailing whitespace here

iliyang:
Thanks, guys, for the work on this ticket! Could a developer please let us know here once this feature has been released in a beta/stable version of Atom so we can try it out? Cheers!

kevinsawicki:
:ship: :green_heart:

paulcbetts:
@raelyard Thanks for bringing this over the finish line!

iliyang:
Thanks again! One question. Will the release notes of the next version include instructions how to enable the portable mode? Also, will (automatic) updating of Atom and add-ons be possible in portable mode?

kevinsawicki:
> Will the release notes of the next version include instructions how to enable the portable mode?

Yes it will

> will (automatic) updating of Atom and add-ons be possible in portable mode?

It should as long as your originally installed using the `AtomSetup.exe` installer. The `atom-windows.zip` does not currently auto-update and so it is not impacted by this change.

nengxu:
Thanks! Eager to try.

LanderN:
Can someone tell me how this works? I tried moving the atom folder to another directory and deleting the .atom folder in the user directory, but it just creates a new .atom folder in the user directory instead of next to the exe. Am I missing something? I'm on the latest beta.

raelyard:
@LanderN The expectation is that regardless of where the exe is running, it will seek a .atom directory as a sibling of the directory in which the .exe lives.  If it finds it, it will use that as the Atom home.  If not, it will use the default .atom in the home directory as the Atom home.  If that also doesn't exist (existing functionality), it will create it in the default location.  Thus, to use this feature, you need to create a .atom directory as a sibling of the directory holding the atom executable (if contents missing, they will generate).  Instead of manually creating a directory, atom can be executed with the --portable parameter, which will trigger a copy of your default home to the portable location relative to where you are executing.

That's a lot of words, please let me know if this is still unclear.

iliyang:
I just tried the ZIP version of 1.3.0-beta0, and when I ran `atom.exe --portable` it didn't create an `.atom` directory neither in the folder containing `atom.exe` nor in the outer folder. Instead it seems to be picking up the settings in `C:\Users\<UserName>\AppData\Roaming\Atom`

nengxu:
@raelyard People may generally expect the .atom folder inside the Atom portable folder, not in parallel to the portable folder. Otherwise, we need to copy both the Atom dir and the parallel .atom dir when moving around. This is against common practice.

I tried manually creating the .atom folder as the sibling of the portable folder, then Atom can pick it up.

Please change the mechanism to put the .atom dir in parallel to the atom.exe, ie, inside the Atom portable folder. Thanks.

raelyard:
@iliyang I will try with the beta builds and see if I see what you see and report back.  @nengxu I agree having it as a sibling to the .exe would be more intuitive.  The location choice for sibling to the directory was made because that plays nicer with the automatic updating structure with Squirrel.Windows and enables straighforward keeping of your settings with updates.  Feedback and thoughts are welcome.

raelyard:
@iliyang When I ran atom.exe --portable on 1.3.0-beta0, I get the expected directory created.  Is it possible you're not executing the exe you think you are?  If you have multiple versions on the machine, possible it's running from somewhere other than what you expect.

LanderN:
@raelyard Thank you for that trick, might come in handy sometime ;) I think the most intuitive way would be to have an option in the installer or even a separate installer, that way it also won't create those explorer context menu items that obviously stop working once you move the installation directory! I know you can adjust them in het registry, but it's something no-one really wants to deal with ;) Or maybe have the zip version be auto-updatable, that would solve those issues too.

sbrl:
A separate installer would make sense - when setting up a portable version of atom people won't always have administrative privileges.

LanderN:
I don't think the installer requires admin privileges to install to the appdata directory. But that doesn't remove the need for another installer for the portable version ;)

raelyard:
I hadn't even thought about the Windows Explorer context menu.  It's worth looking into that and seeing if there's an easy way to make that work that wouldn't require a whole separate installer.

LanderN:
Of course, the same installer with an option to install it to a different location along with an option to not make context menu items/start menu items would be just as good. Better probably.

n054:
work always on portable, going there now. Wishes for green.

proimage:
Regarding Windows Explorer context menu integration, I always liked how 7-Zip (yes, portable) did it. There's a checkbox in the app settings screen to integrate 7-Zip into the context menu:

image

Let the user decide whether to modify the OS the program is running under, and to make that decision easily.

sketchpunk:
I'm not familiar with Atom's build process but isn't the application compiled then assembled. The complete release version is then put inside an installer as another step in the build process. I usually use NAnt for my build process with each step defined and stringed together. I usually have a step where it takes the assembled release copy and just zip it up for portable version, then create an installer from the same release copy. So in whatever you're using to automate the build process, can you just inject an extra step at the end, after making the installer, just zip up your release. You don't need an installer for a portable, ideally a portable app should run just be unzipping it. So without an installer, you won't get all those nasty little OS changes that you don't need if its going to be portable.

sketchpunk:
oh. on top of that. When you zip up the file why not add a batch file called launch_portable.bat that calls the atom.exe with the param to make it run in portable mode. Thats how Portable Node.js works.

paulcbetts:
Atom in Portable mode should not install Explorer hooks at all, setting up Explorer extensions that are sourced from USB drives is a Very Bad Idea

sbrl:
While setting up explorer integration in the portable version of atom 
would be awesome, I'd be a bit concerned that it would leave clutter 
behind on the machine I was using.

proimage:
"setting up Explorer extensions that are sourced from USB drives is a Very Bad Idea"

Indeed it would be, but shouldn't that be an opt-in option left to the advanced user? After all, not everybody runs portable apps from USB drives. I run them from my Dropbox, which syncs most all of the apps I use between home and work computer. The Dropbox folder will always be there, so I relish anytime a portable app gives me the option to enable integration.

I say, if the user is advanced enough to opt for a portable app in the first place, why not give them the option to enable OS hooks? For Pete's sake, this is ATOM we're talking about here - a tool mostly used by computer-savvy programmers in the first place! :)

paulcbetts:
> I say, if the user is advanced enough to opt for a portable app in the first place, why not give them the option to enable OS hooks?

Nope! If they want to do that, they can be advanced enough to compile their own patched copy

jyrkive:
I just tested Atom 1.2.2 and 1.3.0 beta 4 for portability over at portablefreeware.com: http://www.portablefreeware.com/forums/viewtopic.php?p=77080#p77080

The **--portable** switch didn't do anything on my system with the 1.3.0 beta 4 build.

([See here](http://www.portablefreeware.com/faq.php#stealth) what we mean with "stealth". It is just an ideal, and it's entirely optional for portable programs to be stealth.)

raelyard:
@jyrkive I'm not sure I understand what you are reporting.  My understanding of the intent of portability is that you can run the app from location that has the portable home directory and it will use the settings in that portable home directory.  If there is no such directory, it will use the default location of .atom home directory, which is in the user profile.  This enables running in portable mode from a USB drive or cloud storage drive, or in non-portable mode from the normal installation.  On the forum, it seems your expectation is that the user profile .atom directory would go away.  Is that really desired?  I guess that would be necessary for the stealth designation.  Are you reporting that --portable does not create a .atom directory as a sibling of the directory holding the atom.exe?

jyrkive:
Yes, your understanding of the intent of portability is correct.

I didn't have a .atom directory in the user profile directory beforehand. Atom created it when I launched it with --portable. Yes, I'm reporting that --portable didn't create the .atom directory as a sibling of the directory holding atom.exe. (Besides, I thought you might be interested to know the results of our testing even if Atom had passed it, because we are essentially gatekeepers of portable software.)

n054:
Unable to download https://www.atom.io/api/packages/settings-view/versions/0.232.0/tarball: 503
This resul after running Travis.

raelyard:
@jyrkive Can you tell me more about how it failed to create the portable home directory?  I can't seem to replicate your results.  Which Windows are you using?  Where are you executing the atom exe?  Is it a USB drive or other type of medium?

jyrkive:
I'm using Windows 10 Pro. I had atom.exe in "I:\Atom Beta\atom.exe". The I:\ partition is a FAT32 partition in an external hard drive. I launched Atom with this batch file placed in the "Atom Beta" folder:

`START atom.exe --portable`

iliyang:
I'm still having issues with the latest beta 6. On my native Windows 8 system, running `atom --portable` does *not* create an `.atom` folder that is a sibling of the `atom.exe` folder. I already have another Atom "installation" (from a .zip file, not installer) on that system. Even if I manually copy the directory `C:\Users\<MyUser>\AppData\Roaming\Atom` and rename it to `.atom`, the beta still picks up the settings from the old location.

On a virtual Windows 10 system, where I don't have other Atom installations, the `--portable` option does work and creates an `.atom` folder. **However**, it also creates an `C:\Users\<MyUser>\AppData\Roaming\Atom` folder, which is a big no no.

It's also not clear from the release notes whether one should run `atom.exe --portable` only once, after which just opening `atom.exe` will work with the portable `.atom` directory. That's certainly how I'd expect it to work.

Lastly, I don't think it's a problem that the `.atom` directory has to be a sibling of the `atom.exe` directory. Yes, it's a little more inconvenient than having `.atom` in the same directory as the .exe, but one could always put the two directories in a parent directory: `G:\Apps\Atom\.atom` and `G:\Apps\Atom\App\atom.exe`.

Zireael07:
We were supposed to get instructions on how to enable portable mode after updating. I've updated at least twice since the merging of this PR and have yet to see said instructions.

theRealPadster:
I tried downloading the latest windows zip () and running atom.exe --portable to test this out, and it seems to save settings (I changed the theme to test it), but I it didn't make a sibling .atom directory. There is one in my User folder, but not where I ran the exe from (or a sibling folder to the Atom directory). It does seem to read the settings from a sibling .atom folder if I create one. I moved my Users  .atom folder to be a sibling, and changed the theme again and it picked it up. So it looks like it works but -- portable doesn't create the sibling directory. This is with Windows 10 Pro.

r44083:
Atom 1.3.2: Portable mode works. For example: you have "D:\Atom\atom.exe", so you need to make dir "D:\.atom" for enable portable mode.
BUT atom still automaticaly create one folder: "C:\Users\username\AppData\Roaming\Atom".
It seems like a garbage while working. Maybe atom should create this folder nearby .atom folder?

CraigCarboni:
For a Windows environment there are many ways to handle a "portable" or "compact" mode.
You could follow the guidelines from portableapps.com or do things differently.

A portable or compact mode is primarily about running on different systems with means different drive letters and folder paths! Secondary is the location of the .atom folder

Personally I would avoid the use of an environment variable to tell atom.exe that it is in portable/compact mode or where .atom folder is as that can be changed/lost/or need to be set every time for every different system it is run on. (e.g.: from Dropbox) and avoids the creation of batch/script files.

Currently whether installed using the setup program or extracted from atom-windows.zip file, atom.exe creates the .atom folder in the same manner which is great.

When atom.exe runs it could do the following: 

1. Check the folder where it is run from and if a .atom folder exists, use that or else create in the default location


2. Check the folder structure where it is run from and is if it complies with the portableapps.com standard

3. Check the folder where it is run from and see if a "special" file exists
a) does the "special" specifiy the mode?
b) does the "special" file specify the location of the .atom folder

It should not matter to the normal or power user, whether using the installer or unzipping the "windows-atom.zip"

GGG-KILLER:
@CraigCarboni the ATOM_HOME variable is really useful and is one of the features I used in my [portable batch implementation](https://gist.github.com/GGG-KILLER/2851d305302945f69530) I think it all depends on the way you use it.
In my opinion there shouldn't be no need for a `--portable` option, but instead atom should check its path and then if it was on appdata it would realize that it was not a portable "installation" and therefore use the user folder to store the `.atom` folder. If it was not located on appdata then it should proceed to use an `.atom` folder in the executable location instead of the user path.
Of course this is only my opinion, but I consider this behavior more user friendly than having to make an shortcut (or use command line) to open it in a portable manner.

katacarbix:
I think it would be nice to have an option to make it portable for Mac as well.

Sisyphuss:
@katacarbix #9443

ViperCode7:
Well, on my end this works as intended, if I got it right.

I have:
\Atom\core\ (with atom.exe, which from what I understand gets wiped clean on update)
\Atom\.atom\ (with config.cson and all the other packages installed)

--portable switch is useful only when creating a portable version from an already locally configured machine, and all that it does is to create the directory structure I wrote above and copy the .atom from Users to the portable location as a sibling folder to the core folder.

There's literally no need for a ATOM_HOME variable as that would defeat the purpose of a portable app, what would it do on a restricted public computer where variables are not allowed to be set?

The purpose of a portable app is to use as little write operations for variables, registry and files and use only the base path to do its storing.

Now all a user has to do is to launch atom.exe, without any switches and if that directory structure is there, it will use it as the main config folder.

The issue is that the core still creates a Chrome cache in Users\Roaming\Atom. It's not a big problem if that location is known to be r&w enabled by default. IDK, maybe some MSDN junkie can clarify this.

raelyard:
The intent of the --portable command line switch is not that it needs to be run every time.  It's a one-time invocation that takes the existing Atom home in the user profile and copies it to the portable location (sibling of the directory in which atom.exe lives).  From that point on (as long as the portable home directory exists, Atom will see that portable home directory and use it).

I did not try to get rid of the default Atom home directory when going portable.  My thinking is that a person may want to have a default home location for use on their own machine with one set of configuration and packages and such for what they do at home for personal things and have another set for work they do with clients or something of the sort and be able to run those independently.  It also provides a fallback for if the portable home is deleted, there will still be a home directory with a configuration that was at one time desired.

The ATOM_HOME environment variable is not needed for using portable mode via the home directory in the portable location.  It's an alternative that enables using a custom location.  Having it as another option provides additional flexibility to use the editor as you choose.

iliyang:
The only problem for me is that Atom insists on creating a cache directory outside its own folder (in Users\Roaming\Atom). That's not acceptable for a true portable app. Would it be difficult to create that cache folder in the portable `.atom` home instead?

eugenesvk:
Doesn't seem to work for me.
I have Atom installed locally with a custom path set to a Dropbox folder (so `ATOM_HOME` is defined in my environment).
Then I've downloaded and unpacked the zipped Atom version to my USB drive. However, when I run it, it still reads the Dropbox settings.
Starting from a command line with `atom.exe --portable` doesn't change anything either (it doesn't even create any sibling folders (on a side note: a portable app shouldn't have any siblings, everything should be in one folder)).

Would you kindly implement the `.alwaysportable` option or something instead of the not working rules-based system from first option in the **TODO** list? I'm not sure it's worth trying to make up a set of rules that would work universally while there is a much simpler and prone-free solution.
So that nothing in the system (no preset environment variables, nothing) will interfere with the portable installation.

Xcodo:
Probably  bit late to the party now, but would `--make_portable` be less confusing as a command line switch (instead of `--portable`)? It might make it clearer that you only have to run it once

Zireael07:
The .always_portable option was discussed above. Was this implemented or not? I need to move Atom off C for disk space reasons...

tobiasxyz:
As a few people pointed out now Atom still creates directories and files in AppData on Windows (C:\Users\username\AppData\Roaming\Atom to be precisely).

That's not quite yet what Portable means. Would it be possible to move that directory into the sibling .atom directory or maybe as another sibling directory to that? Maybe even inside the actual Atom directory considering it's just a "cache"(maybe create a setting for that in the config)?

n054:
Regarding the always on portable comment, I use a portable device using my chrome browser on desktop mode to push commits. No special file, just a desktop browser on my phone. It's how I contribute to builds but always on my portable device.

ZombieHippie:
Even changing the `ATOM_HOME` and `APPDATA` environment variables on Windows 8.1, and then copying `~/.atom` to my set `ATOM_HOME` uninstalling packages yields error **Failed to delete minimap-cursorline: Does not exist**

image

ZombieHippie:
At this moment in the codebase, it looks like the only thing that Atom should need is the `ATOM_HOME` environment variable, I feel that even if Atom is installed to `LOCALAPPDATA` we should be able to set different `ATOM_HOME` locations in case we have different needs for different projects. This would be similar to having specific workspaces, or environments.

Atom has been built / put into an install directory.
`<Atom Install Directory>/resources/cli` has been added to the PATH.
No other folders under AppData or User folder have Atom installation files.

 1. Use Regshot to capture the state of the drive.
 2. create a new empty folder to set as Atom home
 3. Set `ATOM_HOME` environment variable to this path
 4. Launch `atom` from command line
 5. Use Regshot to capture the state of the drive
 6. Notice that there are many database/indexing files created in the `%APPDATA%\Atom` folder
image
 7. Notice `init.coffee`, `keymap.cson`, `styles.less` are all initiallized as empty files.

methodbox:
WTF is this still using coffeescript ??

damieng:
You do not need to specify atom_home at all. Details on portable mode are in the flight manual. The stuff stored in app data are caches and temp files. It's quite common for portable apps to continue to use temp files in this way. Let us know if you find it storing something else there you think should really be on the portable drive.

ZombieHippie:
@damieng Thank you for clarifying, but a "portable" software should not leave traces on a computer, see [the portable application guidelines followed by Portable Apps](http://portableapps.com/about/what_is_a_portable_app#guidelines). Ensuring that no files are stored in a folder like the APPDATA, enables someone to run multiple versions of Atom, with different packages and different styling set ups.


Thus, if they want to use Atom for Go, they can set it up with the best syntax theme for go, and the best linters and plugins, and config, but still be able to use another installation of Atom for lightweight text editing.

damieng:
Nothing Atom's mode does today precludes you from running any of your scenarios. The files in AppData today are temp/cache files. You can do everything you just said.

PortableApp's definition doesn't say you can't use AppData for cache/temp (storing those files on a USB stick would be awful for performance). The solution here might be to just aggressively clean up the cache/temp files when running in portable mode.

dogancelik:
* I store Atom in `E:\Portable\Atom`
* I start `E:\Portable\Atom\Atom.exe --portable`
* It creates `E:\Portable\.atom`

**That's not what Portable means.** Portable means it would store `.atom` inside the program folder where `Atom.exe` resides.

If you try to move your Atom folder, you have to move `.atom` too every time you need to move Atom to somewhere else. And that's not handy (or portable).

That's like having to hold a suitcase but you also have to hold another one because they love each other and can't do without the other. So wearing a backpack instead (Portable mode) is much better (Atom configuration folder inside Atom program folder) and less burden on [Portable people](http://www.imagebase.net/var/albums/People/IMG_9288%20copy.jpg).

What happens if I put all Atom files in root `E:\`, is Atom going to create a folder `E:\..\.atom` (basically inside *My Computer*)?

--

Using a file `portablemode` to activate Portable mode is a common practice among Windows developers. Many popular programs in Windows use this convention. Examples: Foobar2000, TeraCopy, [ShareX](https://github.com/ShareX/ShareX), [QuiteRSS](https://github.com/QuiteRSS/quiterss), [RetroShare](https://github.com/RetroShare/RetroShare) etc.

damieng:
Because of the way the auto-updater works it can't store .atom inside Atom as the Atom folder gets replaced with the new version.

> That's not what Portable means. Portable means it would store .atom inside the program folder where Atom.exe resides.

Portable means just that, portable between machines. There isn't a specific definition that states config has to be stored directly in the folder or that the app can't use temporary storage on the machine.

dogancelik:
I'm getting errors when uninstalling packages as they get removed from the filesystem but doesn't get removed from Atom itself.

* `ATOM_HOME=E:\Portable\.atom`
* No `--portable` is passed
* Local Atom installation, `Local\Atom\app-1.7.0`

altgr

Console log:
```
C:\Users\dogan\AppData\Local\atom\app-1.7.0\resources\app.asar\node_modules\settings-view\lib\package-card.js:721 Uninstalling package altgr failed Error: Uninstalling “altgr” failed.
    at C:\Users\dogan\AppData\Local\atom\app-1.7.0\resources\app.asar\node_modules\settings-view\lib\package-manager.js:504:21
    at exit (C:\Users\dogan\AppData\Local\atom\app-1.7.0\resources\app.asar\node_modules\settings-view\lib\package-manager.js:73:16)
    at triggerExitCallback (C:\Users\dogan\AppData\Local\atom\app-1.7.0\resources\app.asar\src\buffered-process.js:215:47)
    at ChildProcess.<anonymous> (C:\Users\dogan\AppData\Local\atom\app-1.7.0\resources\app.asar\src\buffered-process.js:237:18)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:200:12) Failed to delete altgr: Does not exist
```

proimage:
I appreciate all the hard work going into this, but wouldn't it be easier not to re-invent the (portable) wheel? Why not just implement according to the de facto portable standard set by PortableApps.com?

damieng:
This is not a case of reinventing a definition to be different but rather working with some hard limitations.

If you wanted to not use the auto update you could easily create a cmd file to launch atom with the config file inside the folder. This would live inside the Atom folder perhaps named atomp.exe and the contents would be:

```
@echo off
set ATOM_HOME=%~dp0
atom.exe %*
```

(Note the way SET works would not leave anything on this machine)

ZombieHippie:
And there is a `--user-data-dir` flag you need to pass to Atom if you want
all its files contained in the install directory, ://
github.com/atom/atom/blob/f48069c5f848b5ae624b110a1df7800e919eb142/src/browser/main.coffee

Gubith:
This works for me:
`\portable\edit\atom\atom\ `
`\portable\edit\atom\.atom\ `

No need to change the folder structure.  

I put my custom startup here:
`\portable\edit\atom\StartAtom.cmd `


Also, I tested multiple installs -  ssd portable install copied to an IMDisk ramdrive.  Both open about the same with a minimum install   (3300 files) and work fine individually.    But when one is open already, the second uses the first ones profile.   This may be expected behavior,  I don't need to run more than one at at time, just want to point this out in case it's unexpected.   After exiting both and running one at a time, they have their own profiles intact.

neojp:
Someone asked about documentation on this matter a while ago. May I mention PortableApps? They have been doing this for years.

http://portableapps.com/development

n054:
Prefer mobile friendly sites over mobile app. Mobile friendly usually allows access to all the files and sites connected to an application or website. It doesn't load up a mobile device with #? megabytes and can be cleaned by clearing the search history and cache. Mobile friendly is recommended by many businesses over mobile apps.

michaelmanoCOM:
If you want you can also do this.

please note the usernames and also make sure your google drive has the folders already made

Programs
Programs->Atom
Programs->Atom->Local
Programs->Atom->Local->atom
Programs->Atom->Roaming
Programs->Atom->Roaming->Atom

mklink /D "C:\Users\USERNAME HERE\.atom" "C:\Users\USERNAME HERE\Google Drive\Programs\Atom"
mklink /D "C:\Users\USERNAME HERE\AppData\Local\atom" "C:\Users\USERNAME HERE\Google Drive\Programs\Atom\Local\atom"
mklink /D "C:\Users\USERNAME HERE\AppData\Roaming\Atom" "C:\Users\USERNAME HERE\Google Drive\Programs\Atom\Roaming\Atom"

Once these links have been created you can now install atom.

If you wish to use atom on another pc just make a batch file with the above symlinks or copy paste them into the new pc and your done as long as you have drive installed (This can be on usb but why bother)

Arlen22:
Really curious where this is at. I would also like portable capability. I really don't care if I have to put the Atom files folder inside a AtomPortable folder and have the .atom folder next to it. That is how PortableApps are usually setup. But it would be good to have the ability to set each kind of folder (profile/cache/temp) from the command line, so that people running off a thumb drive can use the standard temp folder, and people running off a more durable storage drive would be able put the other stuff on as well. 

`atom.exe -profile ..\data\profile -cache ..\data\cache -temp ..\data\temp`

Why can't we implement ALL of the options that have been proposed? I'm sure some of them have problems related to specific implementations. I also like the idea of checking for a file or checking in the parent directory. 

The point is that there are several different use cases here, and I think we can satisfy all of them. We don't need to make this simple and easy, rather we should make it granular and versatile. And well documented, of course.

Arlen22:
Actually, just comprehended that this got merged. Thanks for making a way. My suggestion still stands, but one way is better than no way :)

TeejMonster:
I dug into the PortableApps spec a little bit and found out that they have a really nice template for people wishing to make their own.  Unzip this file, and then unzip the contents of the latest Atom zip release for windows into the "AtomPortable/App/Atom" directory.

The launcher in the root of the zip will launch "AtomPortable/App/Atom/atom.exe".  If a C:\*user*\.atom folder exists, it will be temporarily renamed and the portable version's folder will be placed there.  Once the portable version is closed, the contents of .atom are restored to the portable device and the old .atom directory is restored.  This will, of course, result in slower load times, but it does get the job done.

Also, here is a pre-packaged version of the 1.15 release.  Haven't tested this method heavily, but so far it seems to provide satisfactory results.
https://www.dropbox.com/s/tdlcpv08y2q3d4g/AtomPortable_1.15.paf.exe?dl=0

[AtomPortable.zip](https://github.com/atom/atom/files/891920/AtomPortable.zip)

lee-dohm:
@TeejMonster I applaud the effort and work you put into this. However, we aren't going to allow a link to an unofficial distribution of Atom here. If you want to create a repository that helps people replicate your process and link to that here, I think that would be great. Let me know if you need any help getting that going and I'd be happy to assist.

TeejMonster:
I understand removing the paf file as that contained a full distribution of Atom, however, the zip file that I attached included no Atom files whatsoever and was merely a wrapper.  Is that an important distinction?

lee-dohm:
@TeejMonster Ah, I didn't look inside it ... I thought they were different forms of the same thing. In any case, setting up a repo still seems the right thing to do so that people can examine the files before downloading them.

TeejMonster:
@lee-dohm Fair enough.  Thanks for the push.

You can find the wrapper in its current state [here](https://github.com/TeejMonster/Atom-PortableApps).  If anyone wants to help out with refining the wrapper, I welcome the help as I have just dipped my toe into things so far.

sergeevabc:
PortableApps lists Atom 1.13.1 (Jan ’17), latest stable Atom is 1.18 (Jun ’17), let alone beta 1.19 (Jul ’17).
Could we get portable zip-packed Atom with latest additions at last, please?

damieng:
@sergeevabc You can download atom-windows.zip from our releases page here and follow the instructions at http://flight-manual.atom.io/getting-started/sections/installing-atom/#portable-mode 

There is no need to download a 3rd party version of Atom such as those from PortableApps. We do not control what versions they publish nor do we test or support them.

PR #8446, opened 2015-08-19
Terminate selection drag on text input

### LINKED ISSUE #5200: Pasting or typing instead of selected text results in random selection
When I select some text and then paste something instead of it, sometimes I get random piece of the pasted text being selected. Look at the screencast:
atom_paste_instead_selection
Here I'm selecting pieces of text and pasting some other text. In one of three times it results in a part of the pasted text being selected.

It leads to some inconvenience in my work. After I replace selected text with text from the clipboard, I should check whether something is selected before I type anything (because the text I'm typing will replace the selection).

The same when I start typing while some text is selected. The typed symbol replaces the selection, but sometimes it becomes selected itself. So when I type the second symbol, it replaces the first one.

The behaviour which I observe in other text editors is: replace selected text with pasted/typed one, deselect anything, place the cursor after the pasted text. This is the most convenient option, I guess.

My Atom version is 0.175.0 (and I've observed it in all previous versions), my system - OS X 10.9.5.

### LINKED ISSUE #5243: Partial selection is left behind
_(I am able to reproduce this in safe mode in latest Atom 0.175.0)_

Not always, but quiet often, I have partial selection left behind when I select a portion of code on one line and then start typing hoping to replace the selected chars. When I type my first char, this causes selection jump to the characters on the right and they get deleted when I type the second char. For example, I have this `test.js` file:

```js
  var where = _.pick({
    classified_id: query.classId,
    section_id: query.secId,
    category_id: query.catId
  }, function(val, key) {
    console.log('  **** val:', val);
    console.log('  **** key:', key);
    return !_.isUndefined(val);
  });
```

To replicate:
1. Select a portion of code on any line with a mouse or a touchpad. Release the mouse button.
1.  Switch to keyboard and press the space bar
1. You should see that the selected chars were deleted, but some of the chars on the right are selected now too. 
1. Press space bar again, and those extra selected chars get deleted
1. Do steps 1 and 2 a few times if you can't replicate it on first time.

This is a real pain since I always have to do `Command + Z` to undo  these side effects. Sometimes, I don't notice it and my code no longer runs.

### LINKED ISSUE #8369: Strange selection problem in latest master version
waaat

I select some text, press `r` (or some other key), and the selected text is replaced with `r`, but the selection stays behind and is now selecting some other text. I'm in safe mode, Atom 1.0.8-469662a, OSX 10.10.4.

@nathansobo @benogle @maxbrunsfeld Any of you seeing the same thing and know what might be up? Doesn't happen on all selections but it seems to always happen when there's some text like `foo bar, baz`, and then I drag-select `bar` (up to but not including the `,`), and then press a key to replace the selection.

nathansobo:
Fixes #8369
Fixes #5243 
Fixes #5200

@izuzak, @mnquintana Either of you interested in giving this a try on Yosemite? I'll probably merge it once it's green regardless, but it would be good to confirm it fixes the issue you're experiencing that I can't repro.

On recent versions of OS X Yosemite, there is a delay before we receive the `mouseup` event when dragging selections with the trackpad. This means that we’re frequently accidentally inserting text before the selection process terminates, leading to unexpected selections after the text insertion. This fixes that behavior and makes the behavior more straightforward even in the case where the mouse button is remains held down during text insertion.

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -1792,6 +1792,24 @@ describe "TextEditorComponent", ->
         expect(nextAnimationFrame).toBe noAnimationFrame
         expect(editor.getSelectedScreenRange()).toEqual [[2, 4], [6, 8]]
 
+      it "stops selecting if a textInput event occurs during the drag", ->
+        linesNode.dispatchEvent(buildMouseEvent('mousedown', clientCoordinatesForScreenPosition([2, 4]), which: 1))
+        linesNode.dispatchEvent(buildMouseEvent('mousemove', clientCoordinatesForScreenPosition([6, 8]), which: 1))
+        nextAnimationFrame()
+        expect(editor.getSelectedScreenRange()).toEqual [[2, 4], [6, 8]]
+
+        inputEvent = new Event('textInput')
+        inputEvent.data = 'x'
+        Object.defineProperty(inputEvent, 'target', get: -> componentNode.querySelector('.hidden-input'))
+        componentNode.dispatchEvent(inputEvent)
+        nextAnimationFrame()
+
+        expect(editor.getSelectedScreenRange()).toEqual [[2, 5], [2, 5]]
+
+        linesNode.dispatchEvent(buildMouseEvent('mousemove', clientCoordinatesForScreenPosition([8, 0]), which: 1))
+        expect(nextAnimationFrame).toBe noAnimationFrame
+        expect(editor.getSelectedScreenRange()).toEqual [[2, 5], [2, 5]]
+
       describe "when the command key is held down", ->
         it "adds a new selection and selects to the nearest screen position, then merges intersecting selections when the mouse button is released", ->
           editor.setSelectedScreenRange([[4, 4], [4, 9]])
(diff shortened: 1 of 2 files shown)


mnquintana:
I think this might also fix https://github.com/atom/atom/issues/5243 and https://github.com/atom/atom/issues/5200 (they seem like symptoms of the same underlying issue).

I'll try to test this out soon!

thomasjo:
Will test once I get home.

nathansobo:
> I think this might also fix #5243 and #5200 (they seem like symptoms of the same underlying issue).

I'll be optimistic and add them to the "fixes" list.

thomasjo:
This definitely fixes the problem on my machine running OS X 10.10.5 (14F27).

izuzak:
Yep, fixes the problem for me as well. :zap: @nathansobo

sethlopezme:
@nathansobo You're awesome for figuring this out!

blah-blah:
Is this in 1.0.9? I'm still seeing the odd selection behavior.

mnquintana:
@blah-blah Which selection behavior are you referring to? Just want to make sure it's the same thing this PR fixed.

blah-blah:
I can reproduce the behavior seen in the GIF and described at the outset of issue [#8369](https://github.com/atom/atom/issues/8369) ([this one](https://github.com/atom/atom/issues/8369#issue-100800733)), but I *cannot* reproduce the problematic behavior described later in that issue, where people select text and replace it with 'r' ([here](https://github.com/atom/atom/issues/8369#issuecomment-132326035)).

However, I don't have problems in safe mode, so it looks like I'll need to take a closer look.

izuzak:
> I can reproduce the behavior seen in the GIF and described at the outset of issue [#8369](https://github.com/atom/atom/issues/8369) ([this one](https://github.com/atom/atom/issues/8369#issue-100800733)), but I *cannot* reproduce the problematic behavior described later in that issue, where people select text and replace it with 'r' ([here](https://github.com/atom/atom/issues/8369#issuecomment-132326035)).

@blah-blah I'm a bit confused by this. The two different GIFs you link to show that same problem -- some text is selected, "r" is pressed, and the selection is not updated properly. Can you clarify what you mean by that you can reproduce one problem but not the other (again, since both GIF show the same problem)?

blah-blah:
Sorry. I had myself convinced that one of them involved selecting text and then deleting it, but they both replace selected text with "r."

Here's my problem:

delete_selection

Select text, delete it, and the selection area remains. But I can't reproduce this in safe mode, so I'll try to narrow it down.

izuzak:
@blah-blah Thanks for clarifying -- you're right, I'm able to reproduce that as well.

@nathansobo this seems like the same problem as before. Select some text by dragging, release trackpad, then press `delete` within a short period of time. Not sure why it happens only with `delete` and not with other keys. Do you see the same behavior?

nathansobo:
It makes perfect sense that there would be issues with `delete`, because I made drag termination conditional on text *input*. I didn't think about the delete case causing the same problem. We can reopen that issue or open a new one and I'll extend the solution to cover delete.

nathansobo:
Have #8627 open with a broader fix.

PR #8452, opened 2015-08-19
:memo: fix pane flex scale doc

### LINKED ISSUE #8401: pane.observeFlexScale documentation error
[pane.observeFlexScale](https://github.com/atom/atom/blob/master/src/pane.coffee#L95) description seems to be off. The callback is invoked when pane is resized and its argument is a Number representing the panes `flex-grow`; ability for a flex item to grow if necessary.

It looks like a copy-paste error from `pane.observeItems`.

Also it wouldn't hurt to add a comment to [pane.onDidChangeFlexScale](https://github.com/atom/atom/blob/master/src/pane.coffee#L92), that the panes current flex can be extracted using `pane.getFlexScale()`.

I can create a PR if everybody agrees.

mrodalgaard:
Fixes #8401.

Might consider making `::getFlexScale` and `::setFlexScale` public, as they are pretty essential for controlling / registering the pane’s size from a package.

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -84,7 +84,8 @@ class Pane extends Model
 
   # Public: Invoke the given callback when the pane resize
   #
-  # the callback will be invoked when pane's flexScale property changes
+  # The callback will be invoked when pane's flexScale property changes.
+  # Use {::getFlexScale} to get the current value.
   #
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Thanks for this

PR #8456, opened 2015-08-20
Remove `::mouseWheelScreenRow`

as-cii:
While working on character measurement, I stumbled upon `mouseWheelScreenRow` and tried to reason a bit about it.

The reason why it exists is to let the browser continue firing `mousewheel` events even when the original line goes off-screen. Using this approach, instead of deleting it as we would do for other lines, we keep it around in the DOM until scrolling finishes.

This PR attempts to change that by forcing the browser to fire `mousewheel` events on the parent element. To achieve this, I made use of the following CSS:

``` css
.lines {
  * {
    pointer-events: none;
  }
}
```

This will prevent any kind of mouse interaction with `.lines` and `.line-numbers` descendants and therefore I am not sure about the implications of this solution. Fold icons in the gutter stopped working, but everything else seems to work nicely. I guess the question is: can we find a way to make `.lines` or `.line-numbers` the target of `mouse wheel` events?

This was a pretty quick fix to demonstrate the idea so I just opened a PR but we can easily discard it if it's unfeasible.


nathansobo:
Wow. If this works I'm pretty embarrassed by the complexity of my original solution.

nathansobo:
This breaks the ability to fold/unfold lines by clicking in the gutter with the mouse. May need to re-enable pointer events for those icons.

as-cii:
> Wow. If this works I'm pretty embarrassed by the complexity of my original solution.

The original solution makes a lot of sense and has proven to be bulletproof over time, so a huge :+1: for that. Also, as you are pointing out, the solution I proposed breaks things like folding.

The alternative here would be to manually fire events, probably using [`Document.elementFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) to figure out the `event.target` of the synthesized event and [`EventTarget.dispatchEvent`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) to handle everything else. Something like:

```coffee
[lineNumbersNode, linesNode].forEach (container) => 
  container.addMultipleEventListeners(
    "mousewheel", "mouseenter", "mouseleave", "etc.",
    @forwardEvent
  )

forwardEvent: (event) ->
  # here we should avoid forwarding events bubbled from a previous forwarded event
  element = document.elementFromPoint(event.x, event.y)
  element.dispatchEvent(event)
```

So, assuming all of the above is feasible, I guess the question is: would this approach be less complex than our current `mouseWheelScreenRow` technique? Our DOM should become a little skinnier while scrolling, thereby bringing some speedups on the layout/style-recalculation side of the things.

@nathansobo: what do you think? I am fine with closing this and re-evaluate it once we have more information on synchronous scrolling events (which, if we manage to make them work, should solve this problem as a byproduct).

nathansobo:
I think it's probably less complicated to just preserve our existing implementation. Who knows what kind of implications we'll encounter if we disable pointer events inside the editor... I'm not really keen to find out. If we can really use native scrolling with the touchevents hack, then we probably don't need to synthesize scrolling from mousewheel events and can drop this code in that way.

as-cii:
Agreed. :+1: 

@nathansobo: thank you for evaluating this! :sparkles:

PR #8475, opened 2015-08-22
Add API for rebuilding native modules of packages

nathansobo:
This is to support https://github.com/atom/incompatible-packages/pull/9

This PR adds `Package::rebuild()` and `Package::getBuildFailureOutput()` methods, persists build failures, and considers a package incompatible when it has a build failure.

It also changes the storage location of the cached incompatible native modules in local storage so we can distinguish them more easily from build failures, so we should wait to merge this until the Electron upgrade lands which will force us to invalidate the cache anyway.


nathansobo:
I'd like to merge this immediate following the merge of #7877. Then we can upgrade the incompatible-packages package on that branch and test everything for a while on master.

PR #8480, opened 2015-08-23
Fix config.get for schemas with objects

jeremyramin:
The config.get method does not return the default values for objects with properties in a config schema.
### Example

For example the editor config has:
(excerpt taken from [config-schema.coffee](https://github.com/atom/atom/blob/master/src/config-schema.coffee))

``` coffeescript
...
  backUpBeforeSaving:
    type: 'boolean'
    default: false
    description: 'Ensure file contents aren\'t lost if there is an I/O error during save by making a temporary backup copy.'
  invisibles:
    type: 'object'
    properties:
      eol:
        type: ['boolean', 'string']
        default: '\u00ac'
        maximumLength: 1
      space:
        type: ['boolean', 'string']
        default: '\u00b7'
        maximumLength: 1
      tab:
        type: ['boolean', 'string']
        default: '\u00bb'
        maximumLength: 1
      cr:
        type: ['boolean', 'string']
        default: '\u00a4'
        maximumLength: 1
...
```

A call to `atom.config.get('editor')` returns an object with an 'invisibles' value that is an empty object. We are expecting an object containing the default values for the properties in 'invisibles'. 

This result is caused by `_.defaults()` in `@getRawValue()` not being recursive. The lack of a `_.deepDefaults()` lead me to use a bit of a workaround with `_.deepExtends()`. Now with `_.deepExtends()`, the default values are the base while any new values read from the config.cson overwrite the defaults. This way all default values, even nested ones, appear in the result.
### Side-Effect

A side effect of this issue was that atom/settings-view does not properly present the available configuration options from the config schema. The `appendObject()` method in [settings-view/lib/settings-panel.coffee](https://github.com/atom/settings-view/blob/5a115f9e051da32e7af676daf6076d890d8d5e75/lib/settings-panel.coffee) uses the values provided by `config.get(packageName)`. The lack of keys in the value provided to appendObject would lead to objects in the schema disappearing from the SettingsPanel. With this proposed change, the SettingsPanel now properly displays objects and their properties.

See atom/settings-view#386 and atom/settings-view#518.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1110,6 +1110,24 @@ describe "Config", ->
           nestedObject:
             superNestedInt: 36
 
+        expect(atom.config.get("foo")).toEqual {
+          bar:
+            anInt: 12
+            anObject:
+              nestedInt: 24
+              nestedObject:
+                superNestedInt: 36
+        }
+        atom.config.set("foo.bar.anObject.nestedObject.superNestedInt", 37)
+        expect(atom.config.get("foo")).toEqual {
+          bar:
+            anInt: 12
+            anObject:
+              nestedInt: 24
+              nestedObject:
+                superNestedInt: 37
+        }
+
       it 'can set a non-object schema', ->
         schema =
           type: 'integer'
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -865,7 +865,7 @@ class Config
 
     if value?
       defaultValue = @deepClone(defaultValue)
-      value = _.deepExtends(defaultValue, value) if isPlainObject(value) and isPlainObject(defaultValue)
+      value = _.deepExtend(defaultValue, value) if isPlainObject(value) and isPlainObject(defaultValue)
     else
       value = @deepClone(defaultValue)
 
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -864,8 +864,8 @@ class Config
       defaultValue = _.valueForKeyPath(@defaultSettings, keyPath)
 
     if value?
-      value = @deepClone(value)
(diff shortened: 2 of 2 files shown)


jeremyramin:
It seems that this method does not return a deep clone of the value. I am currently looking for a better alternative to cloning both the value and default value.

thomasjo:
Thanks for helping out! This will require at least one test to verify functionality and to prevent regressions.

thomasjo:
Also note that this currently has failing core specs, and 1 failing settings-view spec.

jeremyramin:
Sorry to trouble you, but how would I go about seeing exactly which specs this is failing? I tried going into atom safe mode and running the atom specs, but that displays errors on both the master branch and my fix branch.

jeremyramin:
Never mind, it seems to have passed this time. I recommend that the deepDefaults method I introduced be moved to the atom/underscore-plus repository. If you want, I can take that code and submit a pull request there.

maxbrunsfeld:
:zap: Thanks for tracking this down and fixing it!

> I recommend that the deepDefaults method I introduced be moved to the atom/underscore-plus repository.

I actually think it's ok to leave this function where it is, because we have specific concerns here about only merging *plain* objects, not instances of custom constructors like `Snippet`.

> it seems to have passed this time.

Awesome. Can you still add a test for this functionality? You might want to just add one or more assertions to [this test](https://github.com/atom/atom/blob/4560914273127105eac7ce2d46bcb8903b972dae/spec/config-spec.coffee#L1108).

jeremyramin:
I went ahead and added two assertions to the test you specified.

maxbrunsfeld:
Great work! I'm so glad you fixed this. The failing travis build is something that was failing on master briefly and has been fixed now.

Issue #8038, opened 2015-07-23
Add a status when Atom is downloading a new update


Ivoz:
When I know there is a new update available for Atom, I can run "Check for Update" in the Help menu. Sometimes it is already running. The problem is lack of feedback: it seems, if it found an update, and is now downloading one, it doesn't tell the user it is doing this. Which would be useful.

Instead, as it is downloading the update, it simply looks like it is stuck checking for one. It would be nice if the Menu item could change to something like "Downloading Update (1.0.2) 43%..." or similar while it is doing this, to give feedback to the user that something is actually going on.

50Wliu:
Checking Update and Downloading Update should be displayed correctly (but without a version number and %) - what are you seeing instead?

Ivoz:
I am running Windows 7 x64, with atom installed from `atom.io` installer.

I am always only seeing a greyed out `Checking for Update` while I am pretty sure there is the new update being downloaded. After it is completely finished, (takes a while for me with ADSL over in Australia) IIRC it updates to text that a new update is ready to be installed (been a week since it happened, sorry).

I can quite clearly remember never seeing a `Downloading Update` text.

Zireael07:
Seconding @Ivoz's observations. Never seen downloading update...

50Wliu:
I'm seeing `Downloading Update` right now as I'm updating to 1.0.3...
![downloading-update](https://cloud.githubusercontent.com/assets/2766036/8947568/7aeb4d8e-356a-11e5-87b5-a03f78636591.png)

Does this also happen with auto-updates or only when you click "Check for updates"?  In my case, Atom started auto-updating and displayed "Downloading update".

batjko:
A percentage or other progress indicator would indeed be nice.

I didn't pay attention when I just updated, so not sure which text it displayed actually.

Ivoz:
I can report I see Downloading Update now, on 1.0.2 -> 1.0.3

![Downloading Update](http://i.imgur.com/cjn48dB.png)

I'm guessing this could have been fixed by something in 1.0.2, haven't verified that yet though.

It seemed to take a while to show up, though. I would have thought a 'new version check' would normally be a <1 second network operation after which you'd quickly update the text.

I'd still advocate making the text more detailed, as a cosmetic upgrade. Adding ellipses at the end, for instance, is a common standard for indicating that something is ongoing. So text like `Downloading Update... (<new_version_>)` would be cool.

50Wliu:
[...]

PR #8483, opened 2015-08-23
Issue #8038: add ellipses to "Downloading Update" message

charleswhchan:
Partially fix to address the lack of download progress by adding ellipses to the "Downlading Update" message.

Solution suggested by @Ivoz in https://github.com/atom/atom/issues/8038#issuecomment-127039291.

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -8,7 +8,7 @@
       { label: 'Restart and Install Update', command: 'application:install-update', visible: false}
       { label: 'Check for Update', command: 'application:check-for-update', visible: false}
       { label: 'Checking for Update', enabled: false, visible: false}
-      { label: 'Downloading Update', enabled: false, visible: false}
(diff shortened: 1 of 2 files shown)


charleswhchan:
This is my first PR. Let me know if I am missing anything. :smiley:

thomasjo:
Thanks for helping out, but ellipsis in menus does not indicate progress. It indicates that more user input is required. See for instance Apple's [OS X Human Interface Guidelines](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/TerminologyWording.html#//apple_ref/doc/uid/20000957-CH15-SW3) for more details.

Ivoz:
@thomasjo, those do not apply, since the item is greyed out and cannot be, and isn't expected to be, interacted with by the user. i.e: it is not a button.

PR #8494, opened 2015-08-24
:memo: add preserveFolds documentation to setSelectedBufferRange(s)

lpommers:
update to the docs

`setSelectedBufferRange` and `setSelectedBufferRanges` can also take a `preserveFolds` bool in the options object.


nathansobo:
:zap: Thanks!

PR #8496, opened 2015-08-24
Add CommandRegistry::onDidDispatch

maxbrunsfeld:
Refs https://github.com/atom/vim-mode/pull/827

### DIFF
--- a/src/command-registry.coffee
+++ b/src/command-registry.coffee
@@ -182,8 +182,19 @@ class CommandRegistry
       stopImmediatePropagation: value: ->
     @handleCommandEvent(eventWithTarget)
 
+  # Public: Invoke the given callback before dispatching a command event.
+  #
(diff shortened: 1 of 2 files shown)


src/command-registry.coffee
```diff
@@ -239,6 +250,8 @@ class CommandRegistry
       break if propagationStopped
       currentTarget = currentTarget.parentNode ? window
 
+    @emitter.emit 'did-dispatch', syntheticEvent
```
jacekkopecky: When this gets invoked, `currentTarget` has become `window`, unless `propagationStopped`. See my comment on atom/vim-mode#827.

maxbrunsfeld: Yeah, I think `target` would be the field you'd want to check; not `currentTarget`.

PR #8497, opened 2015-08-24
Document MenuItem `enabled` and `visible`

ssorallen:
The Electron MenuItem[1] `enabled` and `visible` options are exposed to
Atom context menus in menu_helpers.coffee[2] but are not yet documented.
Both options are passed through to Electron.

[1] http://electron.atom.io/docs/v0.31.0/api/menu-item/
[2] https://github.com/atom/atom/blob/v1.0.7/src/menu-helpers.coffee#L49

### DIFF
--- a/src/context-menu-manager.coffee
+++ b/src/context-menu-manager.coffee
@@ -86,9 +86,14 @@ class ContextMenuManager
   #   * `label` (Optional) A {String} containing the menu item's label.
   #   * `command` (Optional) A {String} containing the command to invoke on the
   #     target of the right click that invoked the context menu.
+  #   * `enabled` (Optional) A {Boolean} indicating whether the menu item
+  #     should be clickable. Disabled menu items typically appear grayed out.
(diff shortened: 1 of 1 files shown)


src/context-menu-manager.coffee
```diff
@@ -86,9 +86,14 @@ class ContextMenuManager
   #   * `label` (Optional) A {String} containing the menu item's label.
   #   * `command` (Optional) A {String} containing the command to invoke on the
   #     target of the right click that invoked the context menu.
+  #   * `enabled` (Optional) A {Boolean} indicating whether this menu item
+  #     should be clickable. Disabled items often appear grayed out. Default
```
nathansobo: Can we say "Defaults to `true`" here? After that I will merge.

ssorallen:
@nathansobo I reworded to "Defaults to `true`".

nathansobo:
Thanks very much!

PR #8500, opened 2015-08-25
Fix a bug with refreshing the gutter UI.

natthu:
I work on a package that adds a gutter to show "dot" next to a line (setting breakpoints). I noticed that sometimes the gutter UI would not update even if the code to decorate the marker successfully executed. When I debugged Atom, I realized that since the order in which gutters are added is undeterministic (unless you set the right priorities), `updateCustomGutterDecorationState` would sometimes return early.


50Wliu:
Can you please add specs for this?

natthu:
Added a spec as requested.

spec/text-editor-presenter-spec.coffee
```diff
@@ -2300,6 +2300,15 @@ describe "TextEditorPresenter", ->
 
             expect(decorationState[decoration3.id]).toBeUndefined()
 
+          it "updates when another gutter with higher priority is hidden", ->
```
as-cii: :bulb: What about something along the lines of:

```coffee
it "updates all the gutters, even when a gutter with higher priority is hidden", ->
```

spec/text-editor-presenter-spec.coffee
```diff
@@ -2300,6 +2300,15 @@ describe "TextEditorPresenter", ->
...
+            editor.addGutter({name: 'test-gutter-1', priority: -150, visible: false})
```
as-cii: What about extracting a variable here, to make clear we are adding an hidden gutter? e.g.

```coffee
hiddenGutter = {name: 'any-name', priority: -150, visible: false}
editor.addGutter(hiddenGutter)
```

spec/text-editor-presenter-spec.coffee
```diff
@@ -2300,6 +2300,15 @@ describe "TextEditorPresenter", ->
...
+
+            # This update will scroll decoration1 out of view, and decoration3 into view.
+            expectStateUpdate presenter, -> presenter.setScrollTop(scrollTop + lineHeight * 5)
+
+            decorationState = getContentForGutterWithName(presenter, 'test-gutter')
+            expect(decorationState[decoration1.id]).toBeUndefined()
```
as-cii: What do you think about testing also for `decoration3` to be defined? :thought_balloon:

as-cii:
I left a couple of comments, but otherwise it looks good. :ok_hand: 

Thanks @natthu for addressing this! :sparkles:

as-cii:
Thank you, @natthu! Great job here! :fireworks:

PR #8507, opened 2015-08-25
Bump symbols-view to v0.103.0

thomasjo:
This PR is necessary as part of the Electron upgrade (https://github.com/atom/atom/pull/7877). A bug was discovered in ASAR that causes hidden files to be ignored. We've worked around that problem by renaming the `.ctags` file in symbols-view to `ctags-config`.


maxbrunsfeld:
:tractor: to electron 30!

PR #8513, opened 2015-08-26
Add 'New View For File' to file menu

### LINKED ISSUE #8409: Ability to open file in new tab multiple times
There is a neat little feature in ST called "New View for File", which basically opens the file a second time in a new tab. Atom Editor has support for this by "splitting" the view, but this isn't always the desired behaviour.

For example, if I have two panes already split and all I want to do is create a new view of a file, then I'd have to split then drag the file back out to destroy the newly created pane.

There should be an option to just open the file in a new tab, in the same pane which it exists in.

jeremyramin:
New option in the file menu to create a new tab for the current active pane item. Only works if the active item is a TextEditor object. The new item is then set as the active item. newViewForActiveTextEditor() returns the new TextEditor or 'undefined' if the active pane item is not a TextEditor. This closes atom/atom#8409.
### Preview

new-view-for-file
### Note

No spec for this yet, because I am not sure which spec this should belong to. Might want to change the command name from `application:new-view-for-file` to something else. I was not sure if it should be 'application' or 'pane' in the command.

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -35,6 +35,7 @@
       { label: 'New File', command: 'application:new-file' }
       { label: 'Open…', command: 'application:open' }
       { label: 'Add Project Folder…', command: 'application:add-project-folder' }
+      { label: 'New View For File', command: 'application:new-view-for-file'}
       { label: 'Reopen Last Item', command: 'pane:reopen-closed-item' }
       { type: 'separator' }
       { label: 'Save', command: 'core:save' }
--- a/menus/linux.cson
+++ b/menus/linux.cson
@@ -7,6 +7,7 @@
       { label: '&Open File…', command: 'application:open-file' }
       { label: 'Open Folder…', command: 'application:open-folder' }
       { label: 'Add Project Folder…', command: 'application:add-project-folder' }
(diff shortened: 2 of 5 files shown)


jerone:
Maybe rename it to: "Duplicate View For File"...

50Wliu:
How about just "Duplicate File"?  Though that might suggest you're also duplicating it on the disk.

mnquintana:
Personally I like Sublime's name for this (`New View into File`). I think the `into` makes it more clear that you're just opening a second "peek" into the file, rather than creating something that might be saved to disk.

jeremyramin:
@jerone That could work if the Atom team likes it.
@50Wliu I think "Duplicate File" would be confusing. 
@mnquintana The original issue mentioned that. That would work well, just wasn't sure if the Atom team wanted to use those exact words.

robsonsobral:
Sadly, it looks like this feature will never see the light of the day.

benogle:
I think it needs to start with a verb. I like @jerone's suggestion: `Duplicate View For File`.

This will need specs for the new command.

damieng:
No response.

PR #8514, opened 2015-08-26
Add Remove Trailing Whitespace editor option

### LINKED ISSUE #7828: New blank indented lines keeps spaces or tabs
When adding multiple indented new lines, the blank one(s) still have tabs or spaces.

```python
# Present Behaviour
class PotOfFlower(object):
····height = 12000
····
····def fall():

# Expected behaviour
class Whale(object):
····height = 12000

····def think():
```

jeremyramin:
The editor settings now include an option to remove the trailing whitespace before an inserted newline. This option is disabled by default in order to prevent the sudden change of the editor's indent functionality.

screen shot 2015-08-25 at 11 50 39 pm
### In Action

remove-trailing-whitespace
The boxes signify lines where there was a leftover indent after a newline command. Once enabling remove trailing whitespace, new lines have their indents set to nothing. Lines with anything besides whitespace will not be effected. The newline will continue to have it's indent.
### Note

This should be pretty easy to write a spec for if it is approved as a good change. This closes #7828.

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -217,6 +217,10 @@ module.exports =
         type: 'boolean'
         default: process.platform isnt 'darwin'
         description: 'Increase/decrease the editor font size when pressing the Ctrl key and scrolling the mouse up/down.'
+      removeTrailingWhitespace:
+        type: 'boolean'
+        default: 'false'
+        description: 'Remove the trailing whitespace from the previous line on newline.'
 
 if process.platform in ['win32', 'linux']
   module.exports.core.properties.autoHideMenuBar =
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -361,6 +361,8 @@ class Selection extends Model
   #     (for example, when a closing bracket is inserted).
   #   * `normalizeLineEndings` (optional) {Boolean} (default: true)
   #   * `undo` if `skip`, skips the undo stack for this operation.
+  #   * 'removeTrailingWhitespace' if 'true', removes the trailing whitespace
+  #     from the previous line on newline.
   insertText: (text, options={}) ->
     oldBufferRange = @getBufferRange()
     @editor.unfoldBufferRow(oldBufferRange.end.row)
@@ -398,6 +400,8 @@ class Selection extends Model
 
     if options.autoIndentNewline and text is '\n'
       @editor.autoIndentBufferRow(newBufferRange.end.row, preserveLeadingWhitespace: true, skipBlankLines: false)
+      if options.removeTrailingWhitespace and @editor.isBufferRowBlank(newBufferRange.start.row)
+        @editor.setIndentationForBufferRow(newBufferRange.start.row, 0)
     else if options.autoDecreaseIndent and NonWhitespaceRegExp.test(text)
       @editor.autoDecreaseIndentForBufferRow(newBufferRange.start.row)
 
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -787,6 +787,7 @@ class TextEditor extends Model
 
     options.autoIndentNewline ?= @shouldAutoIndent()
     options.autoDecreaseIndent ?= @shouldAutoIndent()
+    options.removeTrailingWhitespace ?= @shouldRemoveTrailingWhitespace()
     @mutateSelectedText(
       (selection) =>
         range = selection.insertText(text, options)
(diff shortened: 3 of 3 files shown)


mnquintana:
Awesome! :tada: 

What do you think about renaming the option? It might be a little confusing because there's already a "Remove Trailing Whitespace" option in the whitespace package, which does something slightly different.

settings_-__users_machistequintana_code_campus-map_-_atom

And actually, now that I think about it, this enhancement seems like it belongs in the whitespace package, rather than core. What do you think?

jeremyramin:
So that is where it was. A developer did comment that there was a remove trailing whitespace option. I didn't see it in the editor settings so I was confused.

It could go in the whitespace package. I'll have to take a look at the whitespace package and see where a good place for it might be. Though this feature seems like it belongs with 'Auto Indent' since it is more like an 'Auto Remove Indent on Newline'.

As for a new name, how about 'Conservative Auto Indent' or 'Strip Whitespace on Blank Line'. The name needs work. Another option is to attach it to the 'Remove Trailing Whitespace' option that is already there.

maxbrunsfeld:
@jeremyramin Thanks for this :+1:.

/cc @atom/feedback 

I'd personally be ok with just *always* doing this, and not even having a setting, at least for now. If you *really* wanted to create a line w/ only whitespace, you'd have to type the whitespace and then move your cursor away without hitting 'enter'.

This only affects *blank* lines right, not all trailing whitespace? So if we do make an option, i'd suggest a name like `Clear Blank Lines on Newline`.

50Wliu:
I'm :+1: for this change, but a setting might still be needed because some people seem to like trailing whitespace.

Also :+1: on moving it into atom/whitespace.

lee-dohm:
I agree with @maxbrunsfeld, that we should just *always* do this and let people's current whitespace settings handle the fallout. For example:

* I want to remove the leading whitespace on blank lines
    1. Set `Remove Trailing Whitespace` to `true`
    1. Set `Ignore Whitespace Only Lines` to `false`
* I want to leave leading whitespace on blank lines
    1. Set `Remove Trailing Whitespace` to either `true` or `false`
    1. Set `Ignore Whitespace Only Lines` to `true`

The only problem that this could cause is if there is a discrepancy in the whitespace standard between old code and new code, i.e. leaving trailing whitespace on lines where it already exists but I don't want to allow trailing whitespace on lines that are new or modified. But we already have this problem ... and I don't believe this change would make things any worse.

bronson:
The only place I've seen anyone care about trailing whitespace is Markdown files.  And even then, it only matters when it follows printable chars.  So, having this patch always on sounds fine to me.

Maybe grammar files could repair the [few places](https://en.wikipedia.org/wiki/Whitespace_(programming_language)) that actually need nonzero-length all-blank lines?

lee-dohm:
@bronson If you want to see examples of people who care very deeply about whitespace, check https://github.com/atom/whitespace/issues/10

bronson:
@lee-dohm thanks for that link.  I totally agree with #10: Atom should never EVER make sweeping changes to my files.  That's just wrong.  The Atom team needs to reconsider their position.

However, fixing up the lines that I'm editing anyway?  That seems OK to me.

jeremyramin:
@maxbrunsfeld It checks to see if the user entered a newline, then it checks if the previous line has anything in it besides whitespace. If there is anything that doesn't match the whitespace regex class then the line isn't modified.

example

@lee-dohm I like that configuration setup, very intuitive. From a user's perspective though, I had no idea the whitespace package existed for adjusting the whitespace settings. If the next update stopped suddenly stopped leaving the whitespace since 'Ignore Blank Lines' is off by default, I think some people would be confused.

If anything, atom/whitespace#10 shows that many people rarely ever see the internal package settings like whitespace. If this feature will attach to the whitespace settings, it would probably be best to figure out a way to have these settings more apparent and visible.

maxbrunsfeld:
> it would probably be best to figure out a way to have these settings more apparent and visible.

Yeah, I strongly agree. There's an open issue for this: https://github.com/atom/settings-view/issues/284. We should probably prioritize this soon.

mehcode:
Is there anyway to just not add the whitespace? I ask it like this because of linter warnings about trailing whitespace would still show up while I'm typing if all this does is remove it after I hit enter.

What I'd prefer is that until I start typing non-whitespace the cursor moves but does not actually insert whitespace.

lee-dohm:
@mehcode No, there is currently no support for "virtual whitespace" in the Atom data model. There is/was an Issue for adding it, but I can't find it right now.

maxbrunsfeld:
> the cursor moves but does not actually insert whitespace

I think this would add a lot of complexity, both in the implementation and for the user and package author.

An alternative solution might be to pass the current cursor positions to the linter so that it can filter out warnings about whitespace on the lines where the cursors are.

mehcode:
Going to make an issue for this with those two things mentioned @lee-dohm and @maxbrunsfeld. One of the primary things that bugs me about having the linter on a change interval instead of on save.

jeremyramin:
Just to clarify, I should close this pull request and move the changes to the whitespace package?

jeremyramin:
Closing in favor of atom/whitespace#91.

PR #8515, opened 2015-08-26
fixed pre-1.0-API-call in doc-example

florianb:
Fixed pre-1.0-API-call in the documentation-example of the GitRepository-class, to fit the actual implementation of Project.


lee-dohm:
Apologies for leaving this open so long. It appears that the documentation has been updated in the interim. Thank you for contributing to Atom!

PR #8520, opened 2015-08-26
Change lines container height based on `TextEditorElement` dimensions

as-cii:
After merging the tiles PRs, I have overlooked that the lines/line-numbers containers were still using `scrollHeight` as their height. This is clearly wrong and, although this seemed to have no effect performance-wise, @nathansobo found it had an impact while benchmarking stuff with `chrome://tracing`.

To fix this, I have introduced a `state.content.maxHeight` that's calculated using `TextEditorElement`'s bounding client rect and that gets updated every time its dimensions change.

Sorry about this but hopefully we can get it fixed pretty soon with this PR.

/cc: @atom/feedback for :eyes:

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -88,6 +88,21 @@ describe "TextEditorComponent", ->
       else
         expect(lineNode.textContent).toBe(tokenizedLine.text)
 
+    it "gives the lines container the same height as the wrapper node", ->
+      linesNode = componentNode.querySelector(".lines")
+
+      wrapperNode.style.height = 6.5 * lineHeightInPixels + 'px'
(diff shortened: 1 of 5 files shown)


nathansobo:
@as-cii I hypothesized that it was a problem but it's not clear that it was. The memory *estimate* for large files was pretty massive, but I couldn't find a way to determine actual memory consumption. Regardless this seems prudent. Why hit the GPU with any large layers than necessary?

as-cii:
Agreed! :+1: (and I am truly sorry I haven't caught this early on :disappointed_relieved:)

Please, merge at will when/if Travis becomes :green_heart: (I know it's having some issues today).

nathansobo:
Thanks @as-cii :clapper:

PR #8526, opened 2015-08-26
use undo in IME composition

### LINKED ISSUE #8486: IME composition with multiple cursors
In `text-editor-component.coffee` L249–251 ([here](https://github.com/atom/atom/blob/4560914273127105eac7ce2d46bcb8903b972dae/src/text-editor-component.coffee#L249)) the `compositionend` event causes the editor to put back in the text that was in the original selection, except it breaks with multiple cursors. 

Steps to reproduce:
 1. have `foobar` in a text editor
 2. put cursor before `f` and add a new cursor before `b`
 3. press `shift-right`, now you have `f` and `b` selected
 4. type a composited text (e.g., on a mac with US keyboard, `alt-e`+`e` to get `é`)
 5. you will now have `éooéar` – so far so good, but now do `undo` (`cmd-z`)
 6. end up with `boobar` instead of `foobar`

Possibly related to #5414.

jacekkopecky:
Fixes #8486 I think.

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -2716,33 +2716,6 @@ describe "TextEditorComponent", ->
 
       describe "when a string is selected", ->
         beforeEach ->
-          editor.setSelectedBufferRange [[0, 4], [0, 9]] # select 'quick'
-
-        it "inserts the chosen completion", ->
-          componentNode.dispatchEvent(buildIMECompositionEvent('compositionstart', target: inputNode))
-          componentNode.dispatchEvent(buildIMECompositionEvent('compositionupdate', data: 's', target: inputNode))
-          expect(editor.lineTextForBufferRow(0)).toBe 'var ssort = function () {'
-
-          componentNode.dispatchEvent(buildIMECompositionEvent('compositionupdate', data: 'sd', target: inputNode))
-          expect(editor.lineTextForBufferRow(0)).toBe 'var sdsort = function () {'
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Ah, good solution. Could you add a test for this? You could add it somewhere in [here](https://github.com/atom/atom/blob/master/spec/text-editor-component-spec.coffee#L2742).

jacekkopecky:
thanks for the pointer @maxbrunsfeld

maxbrunsfeld:
Those tests look great. Do you think that they could actually *replace* the corresponding tests where there is only one selection? On `atom/atom` we have a little bit more concern than on `vim-mode` about not having *redundant* tests. If you think there's any reason why the single-selection tests still have their own value, I'm ok with leaving them. What do you think?

jacekkopecky:
I can't see any value in them, so they're :fire:d

maxbrunsfeld:
Beautiful. Just waiting on CI to go green to merge.

jacekkopecky:
:heart:

jacekkopecky:
oh boy; but the same commit just without `:fire:` passed: https://travis-ci.org/atom/atom/builds/77396995

maxbrunsfeld:
Yeah, looks like travis is having issues again.

maxbrunsfeld:
Glad to have you hacking on this repo.

jacekkopecky:
Thank you :blush:

PR #8533, opened 2015-08-27
Document all options for creating Notifications

ssorallen:
- The defaults for the 'icon' values come from [Notification::getIcon][0].
- `dismissable` defaults to `!!options.dismissable`, which equates to `false` for `undefined`.

[0]:
https://github.com/atom/atom/blob/v1.0.7/src/notification.coffee#L79


steelbrain:
:+1:

maxbrunsfeld:
Thanks @ssorallen!

PR #8536, opened 2015-08-27
Remove redundant calls to buffer-to-screen conversion methods

as-cii:
Some methods in `TextEditorPresenter` were calling buffer-to-screen conversion methods more than once, causing unnecessary slowdowns when many cursors/decorations were in place.

Before:

screen shot 2015-08-27 at 11 36 19

After:

screen shot 2015-08-27 at 11 36 04

_(benchmarks executed on a file with ~ 3k rows full of cursors and decorations)_

 :racehorse: 

Ideally, we shouldn't need to be conservative about such kinds of calls but that would require to address the problem from a totally different perspective and, on the other hand, this is a fairly quick win we can take advantage of in the meantime.

/cc: @maxbrunsfeld @nathansobo @atom/feedback

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -416,9 +416,10 @@ class TextEditorPresenter
 
   updateCursorState: (cursor) ->
     return unless @startRow? and @endRow? and @hasPixelRectRequirements() and @baseCharacterWidth?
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
:+1: Looks good to me.

nathansobo:
:spaghetti:

PR #8537, opened 2015-08-27
Gutter z-index

### LINKED ISSUE #71: Deletion mark in gutter is partially covered when the mark's bottom row is selected
In the GIF below, something was deleted between row 16 and 17, so the icon is shown. Notice how when I click in the gutter for row 16 the upper part of the triangle is still visible, but when I click in the gutter for row 17 the lower part of the triangle is no longer visible.

gutter

Atom 1.0.8-a87dd05 in safe mode, on OSX.

Some :pager:: @simurai @as-cii @nathansobo

as-cii:
Fixes https://github.com/atom/git-diff/issues/71.

Some decorations on line-numbers exceed the line height, thus overlapping the next row. When the next row has e.g. a `background-color`, such decorations get hidden. For more information, please have a look at the aforementioned issue; here's a relevant screenshot of the glitch:

visual-glitch

After this PR, similarly to the tiles within `.lines`, we assign a decreasing `z-index` to the tiles in the gutter. We do so for `.line-number`s as well, so that we don't have to deal with DOM ordering while still solving the problem.

Among all the solutions proposed in the above issue, I feel like this is the most unobtrusive (assuming we don't want to change the visuals). Although this doesn't solve an analogous scenario (i.e. when a decoration overlaps the previous row), it fixes the pretty common issue we are observing on the gutter.

I'll test drive this to check if anything got screwed up. Feedback is very welcome! :eyes: :bow: 

/cc: @simurai @nathansobo @izuzak @atom/feedback

### DIFF
--- a/spec/text-editor-component-spec.coffee
+++ b/spec/text-editor-component-spec.coffee
@@ -623,6 +623,26 @@ describe "TextEditorComponent", ->
       expect(tilesNodes[2].style.zIndex).toBe("1")
       expect(tilesNodes[3].style.zIndex).toBe("0")
 
+    it "renders lines upper in the stack in front of the ones below in each tile", ->
+      wrapperNode.style.height = 6.5 * lineHeightInPixels + 'px'
+      component.measureDimensions()
+      nextAnimationFrame()
+
+      # Tile 0
+      expect(component.lineNumberNodeForScreenRow(0).style.zIndex).toBe("2")
+      expect(component.lineNumberNodeForScreenRow(1).style.zIndex).toBe("1")
+      expect(component.lineNumberNodeForScreenRow(2).style.zIndex).toBe("0")
+
+      # Tile 1
+      expect(component.lineNumberNodeForScreenRow(3).style.zIndex).toBe("2")
+      expect(component.lineNumberNodeForScreenRow(4).style.zIndex).toBe("1")
+      expect(component.lineNumberNodeForScreenRow(5).style.zIndex).toBe("0")
+
+      # Tile 2
+      expect(component.lineNumberNodeForScreenRow(6).style.zIndex).toBe("2")
+      expect(component.lineNumberNodeForScreenRow(7).style.zIndex).toBe("1")
+      expect(component.lineNumberNodeForScreenRow(8).style.zIndex).toBe("0")
+
     it "gives the line numbers container the same height as the wrapper node", ->
       linesNode = componentNode.querySelector(".line-numbers")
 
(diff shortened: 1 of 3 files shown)


alexandernst:
I have been watching all this z-index problem with fonts (taller than "normal" chars, symbols) and artifacts. It's starting to feel very dirty and hacky now. Let's hope this won't turn around and byte.

as-cii:
> I have been watching all this z-index problem with fonts (taller than "normal" chars, symbols) and artifacts. It's starting to feel very dirty and hacky now. Let's hope this won't turn around and byte.

@alexandernst: agreed, I feel the same about this kind of hacks. I believe we should try to solve the problem from a different angle but, first, we need to evaluate what other options do we have, making sure they respect our performance/features constraints.

Although tiling complicates things further, the problem lies in lines having a position which does not affect consequent lines (i.e. `position: absolute`). As a result, I think we may always have this kind of problems in the future. On the other hand I am not sure we can get rid of it, as `lineHeight` is the basis for many Atom visual elements (e.g. scrollbars, cursors, highlights).

@nathansobo: while re-evaluating the problem, I spotted a correlation between the "removed line indicator" and inline decorations (although we haven't really discussed them yet) which could employed to solve this issue. Instead of decorating the line with a class, we could put such indicator between (and on top of) the two lines through an appropriate Atom primitive.

What do you think?

spec/text-editor-component-spec.coffee
```diff
@@ -601,6 +601,48 @@ describe "TextEditorComponent", ->
       expect(lineNode.offsetTop).toBe(top)
       expect(lineNode.textContent).toBe(text)
 
+    it "renders tiles upper in the stack in front of the ones below", ->
```
maxbrunsfeld: The phrase "upper in the stack" sounds a bit unusual. I'd maybe say "it renders higher tiles in front of lower ones".

maxbrunsfeld:
> Although this doesn't solve an analogous scenario (i.e. when a decoration overlaps the previous row), it fixes the pretty common issue we are observing on the gutter.

Yeah, I think this is the right move for now.

spec/text-editor-component-spec.coffee
Zireael07: Maybe "higher in the stack" would be better?

simurai:
> Let's hope this won't turn around and byte.

It will be an issue again if a decoration would want to overlap a `.line-number` above. As long as `.line-number` has `position: absolute` we have to make a choice if it should overlap the `.line-number` above __or__ below.

> Find a proper solution for the "z-index problem" #8540

Two options that probably are flawed for some reason.

1. Change `.line-number`s to `position: relative`. But then there are still the tiles that are absolutely positioned.

2. Introduce an "overlay decoration" container for the gutter. An element that is stacked on top of all tiles/lines. But that might defeat the purpose of having tiles, since it could create a huge element that needs to scroll with the rest.

I think it's not easy to solve and there needs to be some kind of compromise or decided on a case by case basis. Like in this PR.

PR #8556, opened 2015-08-28
Update specs for language-javascript changes

winstliu:
Hopefully I got all of them :smile:.

This change allows for the scope change in atom/language-javascript#206 to pass on master.


thomasjo:
I kicked the failed build just in case, but the 10 failures looked genuine.

50Wliu:
Yeah, I'm not too sure how to fix those.  In the meantime though I'll revert the scope change.

PR #8561, opened 2015-08-28
Add a notification when toggling the menu bar

winstliu:
Refs #4289.  People keep hitting it so I thought a notification would be helpful.
Didn't think a spec was necessary, but if someone wants one, I'll add it :).


thomasjo:
Looks good to me! :ship:

PR #8571, opened 2015-08-29
prevent use of socket in apm test

jacekkopecky:
When I run `apm test` while having another Atom window open (on a mac, if that matters), the new Atom instance screws with the socket file: it deletes the original socket on start [here](https://github.com/atom/atom/blob/5e5139b753b9c24b2671064f3b1eceabb31112d9/src/browser/atom-application.coffee#L136), creates its own and deletes that on exit.

That means that after `apm test`, `atom .` no longer opens in the original (still running) window.

Even worse, during `apm test`, `atom .` will open a window in the instance running the tests, which then happily dies (taking the new window with it) when the specs are done.

Comments on the PR welcome.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -65,6 +65,8 @@ class AtomApplication
   constructor: (options) ->
     {@resourcePath, @devResourcePath, @version, @devMode, @safeMode, @socketPath} = options
 
+    @socketPath = null if options.test
+
     global.atomApplication = this
 
     @pidsToOpenWindows = {}
@@ -130,6 +132,7 @@ class AtomApplication
   # the other launches will just pass their information to this server and then
   # close immediately.
   listenForArgumentsFromNewProcess: ->
+    return unless @socketPath?
     @deleteSocketFile()
     server = net.createServer (connection) =>
       connection.on 'data', (data) =>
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Ha, I've seen this happen before and never bothered to look into it. Thanks so much for tracking this down! 

@kevinsawicki Is there any reason we'd want to reuse browser processes for concurrent `apm test` commands? I guess if we do, an alternative solution would be to *rename* the socket path in test mode.

kevinsawicki:
> That means that after apm test, atom . no longer opens in the original (still running) window.

Yeah, I've seen this too, pretty annoying.

> Is there any reason we'd want to reuse browser processes for concurrent apm test commands? I guess if we do, an alternative solution would be to rename the socket path in test mode.

I don't think reusing browser processes prevents anything right now regarding concurrent apm test commands.

So I'm good merging something that is either a test socket path or no socket path at all in test mode. And then seeing if we run into any problems on CI and adjust accordingly once we know more.

jacekkopecky:
Yes, this one is in the direction of "no socket path at all in test mode".

kevinsawicki:
@jacekkopecky this pull request no longer merges cleanly, can you take a look at the conflicts and resolve them? Apologies for the delay in getting this in.

jacekkopecky:
Hi, I can have a look on the weekend, thanks for getting back to this.

jacekkopecky:
This is a quick rebase-and-fix, hope it works. 8-)

maxbrunsfeld:
Yep, still works! Thanks @jacekkopecky.

PR #8578, opened 2015-08-29
WIP RFC: link from package spec stack trace to the files

### LINKED ISSUE #8383: stack trace in spec window should be hyperlinked
It would be great if file pointers in reports of failed specs were hyperlinked so that clicking on them opens the file on the right line. For example, I'd love a hyperlink on this:

```
at operators-spec.coffee:259:32
```

in the screeshot:

screen shot 2015-08-14 at 15 09 43

jacekkopecky:
Fixes #8383.
This is a first attempt, and I can't say I'm confident that it's done right. I'll add some comments in the patch.
Any comments and suggestions on this one will be welcome.

### DIFF
--- a/spec/atom-reporter.coffee
+++ b/spec/atom-reporter.coffee
@@ -29,6 +29,37 @@ formatStackTrace = (spec, message='', stackTrace) ->
   lines = lines.map (line) -> line.trim()
   lines.join('\n').trim()
 
+processStackTrace = (spec, message='', stackTrace) ->
+  return [] unless stackTrace
+
+  jasminePattern = /^\s*at\s+.*\(?.*[/\\]jasmine(-[^/\\]*)?\.js:\d+:\d+\)?\s*$/
+  firstJasmineLinePattern = /^\s*at [/\\].*[/\\]jasmine(-[^/\\]*)?\.js:\d+:\d+\)?\s*$/
+  lines = []
+  for line in stackTrace.split('\n')
+    lines.push(line) unless jasminePattern.test(line)
+    break if firstJasmineLinePattern.test(line)
+
+  # Remove first line of stack when it is the same as the error message
+  errorMatch = lines[0]?.match(/^Error: (.*)/)
+  lines.shift() if message.trim() is errorMatch?[1]?.trim()
+
+  links = []
+
+  for line in lines
+    links.push link = {}
+
+    # Remove prefix of lines matching: at [object Object].<anonymous> (path:1:2)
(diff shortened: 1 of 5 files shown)


spec/atom-reporter.coffee
```diff
@@ -29,6 +29,37 @@ formatStackTrace = (spec, message='', stackTrace) ->
   lines = lines.map (line) -> line.trim()
   lines.join('\n').trim()
 
+processStackTrace = (spec, message='', stackTrace) ->
```
jacekkopecky: haven't replaced `formatStackTrace` because I don't know its other use.

spec/atom-reporter.coffee
```diff
@@ -76,6 +107,10 @@ class AtomReporter extends View
     @on 'click', '.stack-trace', ->
       $(this).toggleClass('expanded')
 
+    @on 'click', '.stack-trace a', ->
+      if href = $(this)[0].dataset.href
```
jacekkopecky: using `dataset` because the `href` attribute was screwing with the path, making it a URI, which then prevented parsing the line/char numbers in `AtomApplication::locationForPathToOpen`

src/browser/atom-application.coffee
```diff
@@ -254,8 +254,8 @@ class AtomApplication
       win = BrowserWindow.fromWebContents(event.sender)
       @applicationMenu.update(win, template, keystrokesByCommand)
 
-    ipc.on 'run-package-specs', (event, specDirectory) =>
-      @runSpecs({resourcePath: global.devResourcePath, specDirectory: specDirectory, exitWhenDone: false})
+    ipc.on 'run-package-specs', (event, specDirectory, senderDevMode) =>
+      @runSpecs({resourcePath: global.devResourcePath, specDirectory: specDirectory, exitWhenDone: false, specRequesterDevMode: senderDevMode})
```
jacekkopecky: this extra parameter `senderDevMode` -> `specRequesterDevMode` is necessary because the spec window must request the `open` with the right `devMode`, otherwise a new window may be created.

src/browser/atom-window.coffee
```diff
@@ -49,6 +49,8 @@ class AtomWindow
     loadSettings.devMode ?= false
     loadSettings.safeMode ?= false
 
+    @devMode ?= false
```
jacekkopecky: without this, the check in `AtomApplication::windowForPaths` was failing: it tried to compare `null` with `undefined` and coffeescript doesn't see them as close enough to pass the `is` test.

benogle:
@maxbrunsfeld what is the status of this?

damieng:
No response.

PR #8583, opened 2015-08-30
:bug: make atom protocol urls more robust

darwin:
I wanted to bundle some fonts to my package as described here: https://atom.io/docs/v0.186.0/creating-a-package#bundle-external-resources.

It didn't work and after diving into Atom's sources I realised that the problem is translation of urls to file-system paths. In my case urls contained extra url parameters, e.g. `atom://plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010`

I have modified the code to strip parts after question mark or hash from urls. That was a quick fix.

Possible future improvement could be using some robust library for proper url parsing.

### DIFF
--- a/spec/atom-protocol-handler-spec.coffee
+++ b/spec/atom-protocol-handler-spec.coffee
@@ -1,15 +1,23 @@
 {$} = require '../src/space-pen-extensions'
 
+fetch = (url, callback) ->
+  $.ajax
+    url: url
+    success: callback
+    # In old versions of jQuery, ajax calls to custom protocol would always
+    # be treated as error eventhough the browser thinks it's a success
+    # request.
+    error: callback
+  
+
 describe '"atom" protocol URL', ->
   it 'sends the file relative in the package as response', ->
-    called = false
-    callback = -> called = true
-    $.ajax
-      url: 'atom://async/package.json'
-      success: callback
-      # In old versions of jQuery, ajax calls to custom protocol would always
-      # be treated as error eventhough the browser thinks it's a success
-      # request.
-      error: callback
+    called = 0
+    callback = -> called += 1
+    fetch('atom://async/package.json', callback)
+    fetch('atom://async/package.json#some-hash', callback)
+    fetch('atom://async/package.json?some&params=value', callback)
+    fetch('atom://async/package.json?', callback) # test edge-case
+    fetch('atom://async/package.json#', callback) # test edge-case
 
-    waitsFor 'request to be done', -> called is true
+    waitsFor 'request to be done', -> called is 5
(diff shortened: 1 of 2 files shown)


ssorallen:
Node bundles [`url.parse`](https://nodejs.org/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost) to parse URLs like this. The following is the output for your example:

```
> url.parse('atom://plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010')
{ protocol: 'atom:',
  slashes: true,
  auth: null,
  host: 'plastic-clojurescript',
  port: null,
  hostname: 'plastic-clojurescript',
  hash: null,
  search: '?v=2.010',
  query: 'v=2.010',
  pathname: '/fonts/eot/hack-regularoblique-webfont.eot',
  path: '/fonts/eot/hack-regularoblique-webfont.eot?v=2.010',
  href: 'atom://plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010' }
```

This brings up the point that `atom:` URLs intended to be file paths should maybe follow the pattern of `file:` URLs and start with 3 slashes like `atom:///` because the parsing then works as expected with `pathname` being everything after the second slash:

```
> url.parse('atom:///plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010')
{ protocol: 'atom:',
  slashes: true,
  auth: null,
  host: '',
  port: null,
  hostname: '',
  hash: null,
  search: '?v=2.010',
  query: 'v=2.010',
  pathname: '/plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot',
  path: '/plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010',
  href: 'atom:///plastic-clojurescript/fonts/eot/hack-regularoblique-webfont.eot?v=2.010' }
```

maxbrunsfeld:
@darwin This is a good change. Could you use `url.parse` as @ssorallen suggested, rather than using `String.prototype.split`? Also, this needs to a merge or rebase to resolve conflicts.

spec/atom-protocol-handler-spec.coffee
```diff
@@ -1,15 +1,23 @@
 {$} = require '../src/space-pen-extensions'
 
+fetch = (url, callback) ->
```
mnquintana: We actually have `window.fetch` natively now - what do you think about using that now that we've moved away from jQuery in core?

darwin:
I'm sorry, I'm not going to update this PR. I'm not really proficient in node.js APIs and I don't understand  Atom to judge what is the best approach.

Feel free to drop my PR and implement it however you feel is best from scratch. Should be just an one-liner.

PR #8604, opened 2015-08-31
Pressing esc on multiple cursors returns to original cursor or selection

### LINKED ISSUE #7738: Feature Request: Pressing `esc` on multiple cursors returns you to original cursor location.
After selecting multiple words by pressing `cmd-d` multiple times, then pressing `esc` key to go back to a single cursor, it seems to place the single cursor at the position of the last cursor that was added.  I would like it to place the single cursor at the original selection.

Explanation:
I got used to Sublime positioning the single cursor at the original cursor location.  At first, I thought it was an arbitrary decision and that I would adjust to atom's way of doing it.  Now I've decided that it's not entirely subjective.  In many cases, my original cursor was intentionally placed at the original position, and that's where I would like it to return to.  For example, maybe I was editing something in-line and thought that this particular word (variable name) I was editing should be edited in a few places at the same time.  After doing this, I would like to return to my original train of thought - editing that line that I was on before creating multiple selections.  I would also think that in any cases where I didn't intentionally place my cursor on a particular instance, the decision is arbitrary whether it takes me to the original or final cursor upon pressing `esc`, so this might be a safe change to make for all use cases.  At least an option for altering this behavior would be much-appreciated.

lpommers:
closes https://github.com/atom/atom/issues/7738

Now, when you have multiple cursors or selections and press `esc`, the single cursor or selection that will remain is the original cursor/selection 

cursors


maxbrunsfeld:
:+1: /cc @atom/feedback

lpommers:
hmm.. not entirely sure why the build fails

50Wliu:
Looks like some package specs (bracket-matcher, snippets) that were testing multi-cursor behavior broke with this change.  You might also need to go update those specs.

lpommers:
@50Wliu thanks. But how can I change them? If I make PRs to those repos, and they get merged before this PR, won't things be out of sync? The spec will have been changed before the code in the core is updated?

as-cii:
This looks good to me. Nice work, @lpommers! :+1: 

> If I make PRs to those repos, and they get merged before this PR, won't things be out of sync? The spec will have been changed before the code in the core is updated?

As for the specs, you raise a good concern here. What do you think about fixing those specs in a way that does not depend on `consolidateSelections`? For example, [in this spec](https://github.com/atom/bracket-matcher/blob/master/spec/bracket-matcher-spec.coffee#L108) or [in this one](https://github.com/atom/snippets/blob/master/spec/snippets-spec.coffee#L473) we may simply destroy cursors manually while still checking that those packages behave correctly.

We can either merge this and fix the specs later, but I think that not depending on this particular behavior would make those specs more bulletproof for the future. :green_heart:

as-cii:
@lpommers: could you please bump the version of those packages in `package.json`? CI should get :green_heart: after that.

Thanks!

as-cii:
I have upgraded the packages directly on master, so you should be able to simply merge from it.

Thanks again!

lpommers:
@as-cii thanks for the code review! This branch is now up to date!

src/text-editor.coffee
```diff
@@ -2264,11 +2264,11 @@ class TextEditor extends Model
     @consolidateSelections()
     @getLastSelection().clear(options)
 
-  # Reduce multiple selections to the most recently added selection.
+  # Reduce multiple selections to the first, original selection.
```
as-cii: :bulb: What about "least recently added" here?

spec/text-editor-spec.coffee
```diff
@@ -1890,7 +1890,7 @@ describe "TextEditor", ->
         expect(editor.getSelectedBufferRanges()).toEqual [[[0, 0], [0, 3]]]
 
     describe ".consolidateSelections()", ->
-      it "destroys all selections but the most recent, returning true if any selections were destroyed", ->
+      it "destroys all selections until the first, original, selections returning true if any selections were destroyed", ->
```
as-cii: :thought_balloon: What do you think if we name this spec:

```
destroys all selections but the least recent, returning true if any selections were destroyed
```

as-cii:
Two small comments and this looks good to me! Thanks, @lpommers!

lpommers:
@as-cii done, and I agree, much clearer now

lpommers:
@as-cii looks like `install apm` failed this time..

lpommers:
and we're back on track here :+1:

as-cii:
Thanks!

maxbrunsfeld:
@lpommers This is a nice improvement; thanks!

lpommers:
thanks! I love working on (and using), so it's my pleasure

colllin:
@lpommers one tiny fix: #9642 Thanks again for this feature!

Issue #8551, opened 2015-08-28
Uncaught TypeError: Cannot redefine property: .ts


Hushpar:
Steps to reproduce below:

1. Start Atom
2. Watching

**Atom Version**: 1.0.8
**System**: Unknown Windows Version
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Cannot redefine property: .ts

```
At c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\compile-cache.js:164

TypeError: Cannot redefine property: .ts
    at Function.defineProperty (native)
    at c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\compile-cache.js:164:10
    at Array.forEach (native)
    at Object.<anonymous> (c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\compile-cache.js:161:24)
    at Module._compile (module.js:452:26)
    at Object.defineProperty.value (c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\compile-cache.js:169:21)
    at Module.load (module.js:347:32)
    at Function.Module._load (module.js:302:12)
    at Module.require (module.js:357:17)
    at require (module.js:376:17)
    at new Task (c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\task.js:32:26)
    at Function.module.exports.Task.once (c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\task.js:19:14)
    at GitRepository.module.exports.GitRepository.refreshStatus (c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\git-repository.js:398:37)
    at c:\Users\All\AppData\Local\atom\app-1.0.8\resources\app.asar\src\git-repository.js:77:26
```

### Commands

```
     -0:34.1.0 application:add-project-folder (atom-pane.pane.active)
     -0:00.7.0 tree-view:toggle (a.btn-issue.btn.btn-error)
```

### Config

```json
{
  "core": {
    "themes": [
      "one-light-ui",
      "one-light-syntax"
    ],
    "autoHideMenuBar": true
  },
  "editor": {
    "invisibles": {}
  }
}
```

### Installed Packages

```coffee
# User
autocomplete-ruby, v0.1.0
emmet, v2.3.12
file-icon-supplement, v0.8.3
file-icons, v1.6.9

# Dev
No dev packages
```

[...]

Issue #3936, opened 2014-10-22
Computer drive (C:) is lowercase on Windows


50Wliu:
A really small nitpick, but I feel like the drive letter should be capitalized.
Explorer capitalizes it (`C:\Servers\[...]`) and so does Notepad++.
It's been lowercase since I started trying out Atom (before 0.100.0).  Also replicated in safe mode (kinda, since I didn't have any packages at the very beginning).

50Wliu:
I'm not running into this anymore, at least in 1.0.0.  A bit weird since the linked node issues haven't been marked as closed, but I'm not complaining :wink:.

PR #8608, opened 2015-09-01
Normalize drive letter case on windows

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/8551

### DIFF
--- a/src/browser/main.coffee
+++ b/src/browser/main.coffee
@@ -54,17 +54,20 @@ start = ->
       else
         path.resolve(pathToOpen)
 
-    if args.devMode
-      AtomApplication = require path.join(args.resourcePath, 'src', 'browser', 'atom-application')
-    else
-      AtomApplication = require './atom-application'
+    AtomApplication = require path.join(args.resourcePath, 'src', 'browser', 'atom-application')
 
     AtomApplication.open(args)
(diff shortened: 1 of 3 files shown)


mnquintana:
Huh, that's super weird that `path.normalize` doesn't catch this. Is this something we should report to the Node team?

50Wliu:
I feel like drive letter capitalization shouldn't matter (see https://github.com/atom/atom/issues/3936#issuecomment-60172258), but if this fixes it, then yay?

src/browser/main.coffee
```diff
@@ -54,17 +54,20 @@ start = ->
       else
         path.resolve(pathToOpen)
 
-    if args.devMode
-      AtomApplication = require path.join(args.resourcePath, 'src', 'browser', 'atom-application')
-    else
-      AtomApplication = require './atom-application'
+    AtomApplication = require path.join(args.resourcePath, 'src', 'browser', 'atom-application')
```
maxbrunsfeld: @kevinsawicki Could you :eyes: this part? My goal removing this conditional was to cut off the problematic drive name at the earliest point possible, which is basically that `__filename` and `__dirname` and friends can be incorrect in `main.js`. This problem get propagated to other files if we call `require` with a relative path.

maxbrunsfeld:
Ok this appears to fix the problem now.

PR #8623, opened 2015-09-01
Fixed link for instructions Node.js.

aranajhonny:
Fixed link for instructions Node.js debian, ubuntu and fedora. It is not something remarkable for us , but it is better to be specific.


mnquintana:
Thanks for this! :rocket:

PR #8632, opened 2015-09-02
Update notification-manager.coffee

devoncarew:
Fix a broken link in the docs page for https://atom.io/docs/api/v1.0.10/NotificationManager. The link from `getNotifications` points to a `Notifications` class instead of `Notification`.


maxbrunsfeld:
Thanks!

PR #8640, opened 2015-09-02
Stop highlighting at the end of the line

vjeux:
Editors like Sublime and Code do not highlight the entire line when doing a multi-line selection, instead they stop after the last character on that line.

I wanted to implement the same thing on Atom. The way I did it is very hacky and not production ready but it gets the job done. Right now, if you highlight 5 lines, the first and last lines are their own "tile" and all the remaining ones are grouped into a single rectangular tile. Since we want to have each line with a different width, I forced the tile size to 1 and removed the logic that makes it expand all the way.

I'd really like this feature to be in atom, let me know how I can refactor the code to make it shippable. Thanks

Different editors (top left: sublime, bottom left: code, top right: atom):
screen shot 2015-09-01 at 12 36 01 pm

Before:
screen shot 2015-09-01 at 5 33 00 pm

After:
screen shot 2015-09-01 at 5 32 29 pm

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -20,7 +20,7 @@ class TextEditorPresenter
     @measuredHorizontalScrollbarHeight = horizontalScrollbarHeight
     @measuredVerticalScrollbarWidth = verticalScrollbarWidth
     @gutterWidth ?= 0
-    @tileSize ?= 12
+    @tileSize = 1
 
     @disposables = new CompositeDisposable
     @emitter = new Emitter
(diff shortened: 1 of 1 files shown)


nmat:
+1

Zireael07:
Loving this!

mnquintana:
Thanks for contributing! It looks like this breaks several core specs though - can you take a look at those? This will also need some specs to be merged.

mehcode:
@vjeux Any chance you can come back and adjust specs? This would be awesome to have.

Furthermore it looks you're not selecting the new line characters. That's part of what makes the feature work / seem to flow.

vjeux:
I'm sorry but I'm not going to update this pull request. Feel free to create a new one.

alexandernst:
^ Totally expected. Guy asked for help, he got a `this breaks some things, fix it, and also add specs`.

mehcode:
Shame. This is a very nice feature.

PR #8644, opened 2015-09-02
remove atomShellVersion

fscherwi:
Remove "atomShellVersion" in package.json.

Whats about "ElectronVersion"?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


mnquintana:
Sorry about any confusion here - `atomShellVersion` is used to control the version of Electron Atom uses (it's still called `atomShellVersion` only because Atom is still on an old, pre-rename version of Electron). [There's an open PR to upgrade Atom to a current version of Electron](https://github.com/atom/atom/pull/7877).

Issue #1654, opened 2014-03-01
Invert I-bar cursor when a dark theme is used


tekkub:
The I-bar cursor is currently always black in Atom.  This makes it very hard to track when using a theme with a dark background.  Sublime inverts this cursor when you're using a dark theme... and it's quite wonderful.  I'd love if Atom did this too.

I don't think you can take a screenshot of the cursor, so here's some crazy photos.  Top is Atom, bottom is Sublime.  Both are using a dark solarized theme.

![2014-02-28 21 30 32](https://f.cloud.github.com/assets/706/2300567/95771684-a0fa-11e3-8351-881401e7a672.jpg)

muhasturk:
+1

muhasturk:
Why dont you embed [this package](https://atom.io/packages/white-cursor "White Cursor") into Atom?

ravinggenius:
Here are screenshots comparing light and dark backgrounds. (from #8659).

<img width="273" alt="atom_light" src="https://cloud.githubusercontent.com/assets/21517/9666201/b7bc55d8-5242-11e5-88bd-5f4f4f7fca22.png">

<img width="277" alt="atom_dark" src="https://cloud.githubusercontent.com/assets/21517/9666200/b7bb0746-5242-11e5-8421-02060312fa22.png">

mnquintana:
@lee-dohm What do you think about bringing this into core? I'd be :+1: for adding this to core, but I think the rest of @atom/feedback should weigh in too. :smile:

lee-dohm:
I'd be fine merging [white-cursor](https://atom.io/packages/white-cursor) into core. This is only a problem on OS X though, right?

DepthDeluxe:
The cursor is also hard to see in Linux as well.  It would be nice to have the option to turn this feature on.

maxbrunsfeld:
What if we just added the [white cursor styling](https://github.com/lee-dohm/white-cursor/blob/1d9e205d00e328db3f73811e1f293b1ebd78d26e/styles/white-cursor.less#L5) to core, as a [less mixin](http://lesscss.org/features/#features-overview-feature-mixins)? That way, dark syntax themes could just include it.

lee-dohm:
That's essentially what my white-cursor package does, provides a Less mixin. There's just a tiny bit of code to configure when to force it if the dark theme doesn't support it.

maxbrunsfeld:
[...]

PR #8663, opened 2015-09-03
Add white-cursor Less mixin for use in dark syntax themes

lee-dohm:
Applies to #1654 

@simurai would you like to take a look at my implementation here and let me know if there is a better approach?

/cc @atom/feedback

### DIFF
--- a/static/atom.less
+++ b/static/atom.less
@@ -28,4 +28,4 @@
 @import "text";
 @import "utilities";
 @import "octicons";
-@import "white-cursor";
+@import "cursorsr";
--- a/static/atom.less
+++ b/static/atom.less
@@ -28,3 +28,4 @@
 @import "text";
 @import "utilities";
 @import "octicons";
+@import "white-cursor";
--- a/static/atom.less
+++ b/static/atom.less
@@ -28,4 +28,4 @@
 @import "text";
 @import "utilities";
 @import "octicons";
-@import "cursorsr";
+@import "cursors";
--- a/static/cursors.less
+++ b/static/cursors.less
@@ -1,20 +1,20 @@
 @ibeam-1x: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAQAAAC1+jfqAAAAL0lEQVQoz2NgCD3x//9/BhBYBWdhgFVAiVW4JBFKGIa4AqD0//9D3pt4I4tAdAMAHTQ/j5Zom30AAAAASUVORK5CYII=');
 @ibeam-2x: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAQAAADZc7J/AAAAz0lEQVRIx2NgYGBY/R8I/vx5eelX3n82IJ9FxGf6tksvf/8FiTMQAcAGQMDvSwu09abffY8QYSAScNk45G198eX//yev73/4///701eh//kZSARckrNBRvz//+8+6ZohwCzjGNjdgQxkAg7B9WADeBjIBqtJCbhRA0YNoIkBSNmaPEMoNmA0FkYNoFKhapJ6FGyAH3nauaSmPfwI0v/3OukVi0CIZ+F25KrtYcx/CTIy0e+rC7R1Z4KMICVTQQ14feVXIbR695u14+Ir4gwAAD49E54wc1kWAAAAAElFTkSuQmCC');
 
-.white-cursor-image() {
+.cursor-white() {
   cursor: -webkit-image-set(@ibeam-1x 1x, @ibeam-2x 2x) 5 8, text;
 }
 
 // Editors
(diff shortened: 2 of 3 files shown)


thomasjo:
This looks good to me!

simurai:
Added the white-cursor to dark editors:

- In editors it's based on `@syntax-background-color`
- In mini editors it's based on `@input-background-color`

Works with most themes, also when mixing light/dark themes. The only issue I found is if a theme uses a semi-transparent color for the mini editors. In that case, the theme should need to use the mixin themselves.

static/atom.less
```diff
@@ -28,3 +28,4 @@
 @import "text";
 @import "utilities";
 @import "octicons";
+@import "white-cursor";
```
simurai: How about renaming it to just `cursor`? There might be a chance that we add more custom cursors that would go into the same file.

simurai:
How about renaming it to just `cursor.less`? Or plural `cursors.less` to match the rest that are also plural (buttons.less, icons.less). And there might be a chance that we add more custom cursors.

And the mixin to just `white-cursor()`? Or `cursor-white()`, `cursor-light()`. Then it would be more like `property-value`.

lee-dohm:
I changed the name of the file and the mixin as suggested.

static/atom.less
thomasjo: This needs to be changed to `cursors`.

lee-dohm:
Thanks @thomasjo ... still not awake yet, it would seem.

static/cursors.less
```diff
@@ -0,0 +1,20 @@
+@ibeam-1x: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAQAAAC1+jfqAAAAL0lEQVQoz2NgCD3x//9/BhBYBWdhgFVAiVW4JBFKGIa4AqD0//9D3pt4I4tAdAMAHTQ/j5Zom30AAAAASUVORK5CYII=');
+@ibeam-2x: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAQAAADZc7J/AAAAz0lEQVRIx2NgYGBY/R8I/vx5eelX3n82IJ9FxGf6tksvf/8FiTMQAcAGQMDvSwu09abffY8QYSAScNk45G198eX//yev73/4///701eh//kZSARckrNBRvz//+8+6ZohwCzjGNjdgQxkAg7B9WADeBjIBqtJCbhRA0YNoIkBSNmaPEMoNmA0FkYNoFKhapJ6FGyAH3nauaSmPfwI0v/3OukVi0CIZ+F25KrtYcx/CTIy0e+rC7R1Z4KMICVTQQ14feVXIbR695u14+Ir4gwAAD49E54wc1kWAAAAAElFTkSuQmCC');
+
+.cursor-white() {
+  cursor: -webkit-image-set(@ibeam-1x 1x, @ibeam-2x 2x) 5 8, text;
+}
+
+// Editors
+& when ( lightness(@syntax-background-color) < 50% ) {
+  .platform-darwin atom-text-editor:not([mini])::shadow .editor-contents--private {
```
mnquintana: So this means the cursor will still be dark against dark-themed mini-editors?

mnquintana: Could these two selectors be simplified into just `.platform-darwin atom-text-editor::shadow`? I'm thinking that we would want this to apply to *any* editor, not just "regular" and mini editors.

lee-dohm: Last I checked, it only works if you target the --private section.

mnquintana: Oops sorry, I didn't mean to exclude that part - I meant reduce the two cases in this file (one for regular and one for mini editors) to one case for all editors. So the selector would be `.platform-darwin atom-text-editor::shadow .editor-contents--private`.

lee-dohm: Ah ok. The issue with that is the two editor flavors key off of different colors, i.e. `@syntax-background-color` vs. `@input-background-color`. This is because some people run a dark syntax but a light UI or a light UI might have dark mini-editors or whatever. See @simurai's comments on the PR itself.

mnquintana: Ah okay then, never mind :+1:

static/atom.less
```diff
@@ -28,3 +28,4 @@
 @import "text";
 @import "utilities";
 @import "octicons";
+@import "cursorsr";
```
mnquintana: :art: `cursorsr` -> `cursors`

lee-dohm: Thanks

lee-dohm:
Rebased on latest. Now that the Electron merge happened, I'll work on testing this extensively.

lee-dohm:
Took me a while to get the time to finish this, but I've tested it pretty well I think. It works like a charm. Someone want to give it a once-over and sign off on it?

maxbrunsfeld:
:clap: Great solution here; nice job guys.

PR #8672, opened 2015-09-04
Add a clarification about the usage of addOpener

fedorov:
Documentation augmented with a note about the need of using fake protocol in
order to invoke opener on a file already opened.

See discussion at

https://discuss.atom.io/t/difficulties-using-atom-workspace-addopener/20444

### DIFF
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -509,9 +509,12 @@ class Workspace extends Model
   #
   # Note that the opener will be called if and only if the URI is not already open
   # in the current pane. The searchAllPanes flag expands the search from the
-  # current pane to all panes. If the file path is converted from a standard path, 
-  # e.g. /foo/bar/baz.html, to a URI with a fake protocol, e.g. custom://foo/bar/baz.html, 
-  # it wouldn't conflict and the opener will be invoked.
+  # current pane to all panes. If you wish to open a view of a different type for 
+  # a file that is already open, consider changing the protocol of the URI. For 
+  # example, perhaps you wish to preview a rendered version of the file /foo/bar/baz.quux 
+  # that is already open in a text editor view. You could signal this by calling 
+  # {Workspace::open} on the URI quux-preview://foo/bar/baz.quux. Then your opener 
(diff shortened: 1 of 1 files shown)


src/workspace.coffee
```diff
@@ -506,6 +506,12 @@ class Workspace extends Model
   #
   # Returns a {Disposable} on which `.dispose()` can be called to remove the
   # opener.
+  #
+  # Note that the opener will be called if and only if the URI is not already open
+  # in the current pane. The searchAllPanes flag expands the search from the
+  # current pane to all panes. If the file path is converted from a standard path,
```
lee-dohm: Starting with `If the file path`:

I think this might be confusing for someone coming to this documentation without being in your specific situation. How about replacing it with something like:

> If you wish to open a view of a different type for a file that is already open, consider changing the protocol of the URI. For example, perhaps you wish to preview a rendered version of the file `/foo/bar/baz.quux` that is already open in a text editor view. You could signal this by calling {Workspace::open} on the URI `quux-preview://foo/bar/baz.quux`. Then your opener can check the protocol for `quux-preview` and only handle those URIs that match.

What do you think?

lee-dohm:
I think this would make a solid addition to the docs.

/cc @atom/feedback

src/workspace.coffee
maxbrunsfeld: :+1:

fedorov: I totally agree, this makes it much more clear.

fedorov:
let me know if you prefer to have one commit - I can squash and issue a new PR

lee-dohm:
No, thank you. The Atom team doesn't squash commits. There's just one final change. It wasn't apparent in my comment, but could you surround the URIs with backticks? So /foo/bar/baz.quux &rarr; \`/foo/bar/baz.quux\`, etc. This marks them as "code" in the eventual documentation `like this`.

fedorov:
Not sure if you received the notification about the commit, but I made the change @lee-dohm suggested.

lee-dohm:
Thanks for the reminder. Sorry I forgot to check back.

PR #8665, opened 2015-09-03
Preserve undo stack on reload

maxbrunsfeld:
- [x] Once Atom's CI passes w/ this text-buffer upgrade, publish a new patch release of text-buffer.

Fixes https://github.com/atom/atom/issues/148
Depends on https://github.com/atom/text-buffer/pull/95

PR #8676, opened 2015-09-04
Preserve undo history after reload, even when file has no unsaved changes

### LINKED ISSUE #8758: Opening a file and triggering an undo removes all content from the file
Here's a GIF:

undoooo

I start Atom, open a file, type something, and then trigger undos -- after the text I added is removed another undo removes all content from the file. Not the behavior I expected - I expect undo to only undo changes I made. I can reproduce this even when I don't type anything -- a single undo removes the content of the file. I'm on OSX 10.10.5, Atom 1.0.12-bc68056

cc @maxbrunsfeld I'm wondering if this might be related to the recent undo support across sessions?

maxbrunsfeld:
#8665 didn't completely fix the problem.

Fixes https://github.com/atom/atom/issues/8758


lee-dohm:
Looks good to me :rocket:

PR #8680, opened 2015-09-05
Remove jQuery from Atom core

nathansobo:
This has been a long time coming. I made the early mistake of basing Atom's initial view system, SpacePen, on jQuery, which solves a lot of problems that aren't at all relevant when targeting a single, modern browser engine. Last summer we removed the reliance on jQuery from Atom's public API and setup a pretty complicated layer of shims to keep the old jQuery-based views working through the deprecation period. With that finally behind us, this PR will remove jQuery and SpacePen entirely. We'll need to transition bundled packages away from SpacePen to finally get the benefit of reduced require time by not loading jQuery, but this is a first step.

Some things to note:
- Our default tooltips are based on bootstrap, which ships its tooltip support as a jQuery plugin. I reworked their code to remove the dependency on jQuery, which could _slightly_ alter a couple corners of the API. It would be great if people could help me test that the new tooltips work correctly in all cases.
- Some of our rarely used global spec helpers have been removed.
- The last incomplete step is to switch our test reporter from SpacePen to raw DOM APIs.

I do worry that old packages that depend on jQuery _might_ break in unforeseen ways, since we were augmenting the jQuery that shipped with SpacePen with extra stuff. For example, we had shimmed `trigger` to run Atom commands, but that no longer works. I'm not totally sure what's going to happen, but most packages developed in newer styles should have nothing to do with this.


batjko:
Maybe a blog post, linked to on Slack and Discuss, would help raise awareness among package authors to confirm whether this will impact them.

Yomi0:
Will those tooltips be created by Blink?
(The same way, or similarly enough?) 
If not, disregard this post. 

Right now, there's a particularly terrible bug¹ open that would make Atom unusable in Enlightenment. 

1: https://code.google.com/p/chromium/issues/detail?id=381732

nathansobo:
There are some spurious API-related failures, but our main build is passing, so I'm going to merge this.

@Yomi0 We're still creating the tooltips manually via the DOM... I just ported the Bootstrap code to native DOM APIs. So everything should be okay on that front at least. Happily the next Atom release will be accompanied by a 1-week beta phase we'll be rolling out next week.

spec/atom-protocol-handler-spec.coffee
```diff
@@ -1,15 +1,9 @@
...
     called = false
-    callback = -> called = true
-    $.ajax
-      url: 'atom://async/package.json'
-      success: callback
-      # In old versions of jQuery, ajax calls to custom protocol would always
-      # be treated as error eventhough the browser thinks it's a success
-      # request.
-      error: callback
+    request = new XMLHttpRequest()
```
mnquintana: Could this be `window.fetch` now that we have Chrome 43?

nathansobo: Seems like it probably could.

PR #8690, opened 2015-09-06
Support query strings for atom protocol.

kaitoy:
I created an atom package [disturb-me](https://atom.io/packages/disturb-me) which adds an image randomly moving in the atom window by an `img` tag.
When I specify a gif image located in the package using atom protocol (e.g. `atom://disturb-me/assets/atom/white/atom_die.gif`), the package set the `src` of the `img` tag to it and then the image is shown. But the gif animation is run only when the image is loaded for the first time.
It seems this is because the image was cached. So, to avoid such caching, I want to add a query string with a random value to the url. (e.g. `atom://disturb-me/assets/atom/white/atom_die.gif?time=1441559906660`)
I locally tested the random query method with this change I'm pull-requesting and confirmed the caching issue was solved.

### DIFF
--- a/src/browser/atom-protocol-handler.coffee
+++ b/src/browser/atom-protocol-handler.coffee
@@ -32,6 +32,10 @@ class AtomProtocolHandler
     protocol.registerProtocol 'atom', (request) =>
       relativePath = path.normalize(request.url.substr(7))
 
+      qMarkIdx = relativePath.indexOf('?')
+      if qMarkIdx isnt -1
+        relativePath = relativePath.substring(0, qMarkIdx)
+
       if relativePath.indexOf('assets/') is 0
         assetsPath = path.join(process.env.ATOM_HOME, relativePath)
         filePath = assetsPath if fs.statSyncNoException(assetsPath).isFile?()


ssorallen:
Use [`url.parse`](https://nodejs.org/docs/latest/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost) instead of `indexOf` to let Node do the URL parsing.

steelbrain:
@ssorallen `url.parse` would've been appropriate if he was using those query strings anywhere, We shouldn't parse it unnecessarily, the approach he is using is good IMO.

ssorallen:
If `url.parse` were used at the top of the function, there would be no need to do this at all. Using `substr(7)` has no meaning until you stop to count the characters in "atom://".

```cs
# Creates the 'atom' custom protocol handler.
registerAtomProtocol: ->
  protocol.registerProtocol 'atom', (request) =>
    urlParts = url.parse(request.url)
    relativePath = path.normalize(
      path.join(urlParts.hostname, urlParts.pathname)
    )

    if relativePath.indexOf('assets/') is 0
      assetsPath = path.join(process.env.ATOM_HOME, relativePath)
      filePath = assetsPath if fs.statSyncNoException(assetsPath).isFile?()

    unless filePath
      for loadPath in @loadPaths
        filePath = path.join(loadPath, relativePath)
        break if fs.statSyncNoException(filePath).isFile?()

    new protocol.RequestFileJob(filePath)
```

steelbrain:
@ssorallen in the example you provided, You are only using the value of `urlParts` once, to get a `relativePath`. And to get that, You are parsing a whole url. @kaitoy's approach is still simpler because we just need the url of it.

I believe we should not overuse the APIs. I wouldn't have objected about the url parsing if you were using the query map after parsing it. but you are not. Parsing a whole url instead of using a hard-coded `.substr(7)` and query string trim is completely unnecessary IMO.

ssorallen:
This isn't an overuse of an API, this is an appropriate use of an API. `request.url` is a URL. `url.parse` parses URLs. Saying `.substr(7)` is more appropriate than `url.parse(request.url)` is sacrificing readability for an unproven improvement on performance. Given how infrequently this function is called, the CPU and memory usage `url.parse` is insignificant.

steelbrain:
I guess it's a matter of personal taste then :)

ssorallen:
I try to write code as if I won't be the next person to edit it, particularly in open source. `substr(7)` had no significance to me when I first saw this, and I had to poke around to see Atom URLs looked like "atom://". `url.parse` is explicit in its functionality.

kaitoy:
Is it allowed to use uppercase letters for a package name? If yes, we cannot use url.parse because it lowercases hostname.

ssorallen:
@kaitoy Aha great point, I didn't know that. Seems like Atom URLs should stick to the pathname with triple slashes like `atom:///PACKAGE-name/foo`.

damieng:
Closing as this doesn't really add ? query parameter support, rather it just strips it off so you can use it as a cache bypass.

We are intending on reviewing the atom protocol support with regards to features, facilities and multiple platform support which will include a review of how we handle query strings.  Cl

PR #8697, opened 2015-09-07
CSP policy prevents using Polymer in Atom plugins

anaisbetts:
I'm trying to use [Polymer](https://polymer-project.org) in an Atom plugin, but the CSP policy blocks inline scripts in [HTML imports](http://www.html5rocks.com/en/tutorials/webcomponents/imports/), which is core to how Polymer works. I'm not sure that this actually protects against any actual security boundary in Atom/Electron's security model (which isn't equivalent to the web)

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <meta http-equiv="Content-Security-Policy" content="default-src *; img-src blob: data: *; script-src 'self'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: *;">
+  <meta http-equiv="Content-Security-Policy" content="default-src *; img-src blob: data: *; script-src 'self'
+  'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: *;">
   <script src="index.js"></script>
 </head>
 <body tabindex="-1">


nathansobo:
This isn't something we're really comfortable doing from a security perspective. Installing an Atom package is one thing... you expect it to run code. But this opens up a hole where a carelessly written package could execute arbitrary code just by inserting some malicious content into the DOM. Just opening content in Atom shouldn't be risky.

We've talked to the Chrome team about fixing the CSP issue with HTML imports, which is the real problem. If there were some solution in place to honor scripts in legal imports, such as propagating a nonce through to imported script tags, this wouldn't be an issue. For now, though, you'll need to use another approach.

paulcbetts:
>  But this opens up a hole where a carelessly written package could execute arbitrary code just by inserting some malicious content into the DOM. Just opening content in Atom shouldn't be risky.

Can you help me understand the security boundary here? Packages are already executing arbitrary code, adding the ability to create a script element just means they can execute arbitrary code in a less straightforward way

>  Just opening content in Atom shouldn't be risky.

This is a compelling argument against `eval`, but I don't see how it applies to the `script` element, because an Atom plugin would have to explicitly choose to create the script element first. I don't think our boundary is "malicious plugin", since plugins can still download an arbitrary executable and just run it. I definitely agree with the overall principle, I just don't see how banning script tags will further that principle

nathansobo:
What if a package loads content with an unescaped script tag? Now we're running code that wasn't created by a package author, right?

paulcbetts:
> What if a package loads content with an unescaped script tag? Now we're running code that wasn't created by a package author, right?

Maybe? It seems like a weird edge case where an author is intentionally doing something stupid, and this doesn't even stop them, because they can use the `<webview>` tag with node integration enabled and do the same thing, or create a `BrowserWindow` and load arbitrary content into it with node integration (tbh a much more plausible scenario because it's something I could imagine someone actually wanting to do).

paulcbetts:
Threw some tags on this one - this bug is blocking me from being able to publish https://github.com/paulcbetts/atom-twitch

benogle:
> https://github.com/paulcbetts/atom-twitch

Neat package

paulcbetts:
Updated this against master

justinfagnani:
@paulcbetts while I sympathize with you on this issue (a lot), it shouldn't really be blocking as you can run Cripser[1] on all HTML files in your project. It's a PITA, but a viable workaround.

[1]: https://github.com/PolymerLabs/crisper

paulcbetts:
@justinfagnani That's where the (sanely) part comes in

TimvdLippe:
I had plans to write a package tooz with Polymer. What is the official stance of the team on the designated way to develop such plugins? Using Crisper or integrating this change (or maybe something completely different)?

nathansobo:
We're still officially unwilling to run arbitrary code based only on writing content to the DOM. It's very unfortunate but until Chrome addresses the issue in some better way, it's the way it is. I would entertain a well-done PR adding an option to disable the CSP check at the user's own risk, but it seems to risky to do as a default experience.

paulcbetts:
> I would entertain a well-done PR adding an option to disable the CSP check at the user's own risk, but it seems to risky to do as a default experience.

This doesn't seem super useful to enable writing plugins in Polymer

paulcbetts:
I figured out a way around this if any other plugin writer needs it:

https://github.com/paulcbetts/atom-twitch/commit/4fe4399229b3b4517b2869cc33f6a68afff409d7

PR #8699, opened 2015-09-07
Add "Continuous Reflow" mode

as-cii:
When this option is enabled, we reflow lines and line numbers at a minimum interval rate of `200ms`, similarly to Chrome's **Continuous Repaint** feature. If another frame is requested in the meantime, we dispatch it anyways but still reflow everything.

This feature is particularly useful to inspect **Layout** and **Recalculate Styles**, possibly tracking down performance issues related to them. A typical workflow to inspect slow styles would look like the following:
1. Settings -> Enable continuous reflow
2. Open the document you'd like to inspect
3. Record a few seconds with the Timeline
4. Open your personal stylesheet
5. Add/Remove styles
6. Go to 2)

Please, note that reflowing happens only when the editor is focused (so to avoid noise from benchmarking multiple editors at once).

I am a bit unsure we should expose this as a user-facing setting: it would probably be okay to show it only when dev mode is enabled. What do you think? :thought_balloon: 

Thanks!


nathansobo:
I almost wonder if this should be a setting you enable via the console so people don't accidentally turn it on and ruin their experience. Kinda like Chrome flags are a bit hidden away.

lee-dohm:
I agree. And the setting should not be persisted between sessions if it is hidden like that.

as-cii:
Agreed. :+1: 

What about exposing this on a per `TextEditorElement` basis? Like, opening up the console and calling a method on it:

```coffee
editorElement = $0 # or atom.views.getView(...)
editorElement.setContinuousReflow(true)
```

nathansobo:
:+1: on an element API.

as-cii:
Okay, I have just added it. :+1: 

Please, have one last :eyes: and merge at will. Thanks!

nathansobo:
Thanks!

Issue #4279, opened 2014-11-21
Indent guides are disconnected


mc10:
Indent guides do not connect to each other; they are separated by a small but noticeable blank. It'd be better if the lines were connected as one straight line.

simurai:
I can reproduce this as well by increasing/decreasing Line Height in the settings. I think it happens in all themes, but you only see the dots if the line is semi transparent.

@thomasjo's suggestion seems to work fine. But maybe to limit side effects, it could be applied to just `.indent-guide` instead of any `<span>`.

```css
atom-text-editor::shadow .indent-guide {
  display: inline-block;
}
```

50Wliu:
My line height is set to the default but I still get this.  I'm using Inconsolata size 14 on Ubuntu and Consolas size 14 on Windows.

simurai:
@50Wliu Sorry, this hasn't been fixed yet. But it's it's now on master.

PR #8700, opened 2015-09-07
Display `.indent-guide`s inline

as-cii:
This is part of our ongoing effort to cut down rendering times.

Apparently, our indent-guides have a modest impact on Layout. Currently, they are styled with `display: inline-block` so that they have the same height of line nodes. This produces the effect of a contiguous vertical line:

screen shot 2015-09-07 at 14 21 42

My assumption is that, when rendering text runs, Chrome has to take a slightly different route to lay out inline blocks, thereby slowing down reflows.

As a result, if we use `display: inline` for indent guides as well, **Layout** decreases from `3.2ms – 3.5ms` to `2.0ms – 2.5ms` for a full reflow. This, however, has the consequence of showing indent guides as the following:

screen shot 2015-09-07 at 14 25 59

I have worked with @simurai to find a performant replacement which didn't alter the UI, but we had no luck in our investigation. We might probably render indent guides _outside_ the line, but I am not sure it's really worth it, especially since users may be okay with the slowdown and still keep the contiguous line effect with a simple change in their stylesheet:

``` less
atom-text-editor::shadow .indent-guide {
  display: inline-block;
}
```

Thanks!

### DIFF
--- a/static/text-editor-shadow.less
+++ b/static/text-editor-shadow.less
@@ -129,7 +129,6 @@
 }
 
 .indent-guide {
-  display: inline-block;
   box-shadow: inset 1px 0;
 }
 


thomasjo:
Initial knee-jerk reaction is that if we don't render the indent guides as continuos lines, we might as well not render them at all. The way the are presented in the second screenshot makes them look like a visual artifact, and hence attention grabbing which is very unfortunate.

Is the performance gain here actually useful? Does it outweigh the very obvious negative side-effect of this change? Perhaps you can provide some additional context to shed some light on why a ~1 ms improvement is substantial enough to warrant this change? How many reflows/redraws are we typically doing under normal editor usage? Thanks :bow:

as-cii:
@thomasjo: thank you for your feedback! :sparkles: 

> Initial knee-jerk reaction is that if we don't render the indent guides as continuos lines, we might as well not render them at all. The way the are presented in the second screenshot makes them look like a visual artifact, and hence attention grabbing which is very unfortunate.

Yeah, that makes sense and I agree it looks somewhat weird. I am still okay with the resulting effect but I'd understand if anyone disagrees with that.

> Is the performance gain here actually useful? Does it outweigh the very obvious negative side-effect of this change? Perhaps you can provide some additional context to shed some light on why a ~1 ms improvement is substantial enough to warrant this change? How many reflows/redraws are we typically doing under normal editor usage? Thanks

For any visual change in the editor, we have to render all the elements within ~16ms. :sweat_drops: 

As of today, layout times (excluding style recalculation) under typical usage (i.e. when only part of the document is reflowed) take from `1.3ms` up to `2.2ms` (this increases as we add more indent guides), while with this change they almost always take `<= 1ms` (I have discovered another bottleneck in the gutter that may allow us to get some more speedups).

I definitely agree it's not a huge saving in terms of absolute values, but a `2.2ms` layout is ~14% of a single frame budget: along with **Recalculate Style**, reflows take usually ~50% of the time spent in dispatching an animation frame.

As I said, this is still an ongoing effort to cut down rendering times and, although we can still do way better in our javascript, I feel like we will need to make small adjustments as the one in this PR to make that happen.

Ultimately, even if we end up not merging this PR because of the UI glitch, I believe it's still good to know which parts of the rendering layer could be tweaked. :racehorse:

50Wliu:
Just going to link #4279 here since that was the issue that triggered the adding of `inline-block` to the styles.

Personally, I'm also :-1: on rendering indentation guides as disconnected, but if there's absolutely no other way to stay under the 16ms limit, then I have no objections to merging.

nathansobo:
I don't think we can insert the breaks. Maybe there's another approach we could take that's more similar to the wrap guide, but that's obviously a much bigger project.

simurai:
I fiddled around a bit more and this seems to work:

```less
atom-text-editor::shadow .indent-guide {
  display: inline;
  box-shadow: none;
  &::before {
    content: "";
    position: absolute;
    width: 1px;
    height: 100%;
    box-shadow: inset 1px 0;
  }
}
```

It moves the guide to a pseudo element. I already tried that the other day, but somehow missed that `height: 100%;` uses the height of the `.line` and not `.indent-guide`, which is what we want.

@as-cii if you wanna test it again with that. Although not sure if it helps. By absolutely position the pseudo element, it uses `display: block` which might also "has to take a slightly different route to lay out". I mean we just move the problem into a child element.

as-cii:
@simurai: thanks for trying that out. It turns out that there is a small performance improvement with the pseudo-element technique, but unfortunately it's almost unnoticeable.

> Maybe there's another approach we could take that's more similar to the wrap guide, but that's obviously a much bigger project.

Yeah, I was thinking to something along those lines but it could be quite a lot more complicated than our current approach.

-----------------------------------------

@nathansobo: although the solution hereby proposed is unfeasible from a UX perspective, I'll keep this PR around so that you can use it as a playground during the meeting, and maybe ask the Chrome team if there's anything we can do about it. Feel free to close it as soon as you're done with it.

Thanks guys for taking the time to have a look. :pray:

olmokramer:
Adding the following to my `styles.less` achieved the desired result. I don't know if it's any faster but scrolling through a file doesn't feel any different from the `inline-block` solution (though I never noticed bad performance anyway, so I might not be the best judge of that :stuck_out_tongue_closed_eyes:).

```less
atom-text-editor::shadow .indent-guide {
  display: inline;
  float: left;

  &:last-child {
    display: inline-block;
    float: none;
  }
}
```

The `:last-child` part was required to fix lines that contain only `.indent-guide`s (no text) which would otherwise have a height of `0px`, thus pulling the next line up. So it won't `inline` all the indent guides, but at least it's a significant portion of them :) Or maybe you can come up with something else.

simurai:
Nice. Ya, would be interesting to see if reducing inline-blocks also has an effect.

as-cii:
@olmokramer: thanks for trying that out. Unfortunately, performance-wise your snippet is kinda similar to the other approaches we have tried out: I believe this happens because Chrome has to take a different path to render non-inline (or floated) pieces of text.

By the way, I think we were able to reach some good reflow times after :ship:ping #8730 and, therefore, I feel like we can close this as it is too disruptive from a UI point of view.

Thanks everyone for the precious feedback! :sparkles:

PR #8703, opened 2015-09-07
Pop scope from the top of scopeStarts directly

### LINKED ISSUE #8701: Tokenized buffer scopes bleed from previous line to the end of next one
The symptoms can be seen at atom/first-mate#56

The problematic case is when a scope ends in the beginning of the next line before any characters (at row 0?).

Ingramz:
Fixes #8701  

If scopeStarts contains the end element we encounter from tags, we should pop it directly from scopeStarts instead of pushing it to scopeEnds. If the top element wasn't the same (or it didn't exist), we add scopeStarts element back and push the tag to scopeEnds just like we used to.

Also I am interested in knowing if there is a safe way to peek the top of stack without popping.

### DIFF
--- a/src/token-iterator.coffee
+++ b/src/token-iterator.coffee
@@ -32,7 +32,14 @@ class TokenIterator
       tag = tags[@index]
       if tag < 0
         if tag % 2 is 0
-          @scopeEnds.push(atom.grammars.scopeForId(tag + 1))
+          scopeName = atom.grammars.scopeForId(tag + 1)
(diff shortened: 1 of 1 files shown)


izuzak:
cc @nathansobo for :eyes:

nathansobo:
Looking at this now. Adding some tests and making some stylistic adjustments, but @Ingramz  you're absolutely right that this was a total oversight on my part. Fantastic investigation. I really appreciate it.

PR #8715, opened 2015-09-08
Correctly terminate scopes that end at the beginning of a line

### LINKED ISSUE #8701: Tokenized buffer scopes bleed from previous line to the end of next one

nathansobo:
Closes #8703
Fixes #8701

Major props to @Ingramz for discovering this obscure bug in `TokenIterator`. This PR adds a test and adjusts the implementation slightly, while incorporating his original commits. Thanks so much for your contribution :metal:.

### DIFF
--- a/src/token-iterator.coffee
+++ b/src/token-iterator.coffee
@@ -36,7 +36,7 @@ class TokenIterator
           if @scopeStarts[@scopeStarts.length - 1] is scopeName
             @scopeStarts.pop()
           else
-            @scopeEnds.push()
+            @scopeEnds.push(scopeName)
           @scopes.pop()
         else
           scope = atom.grammars.scopeForId(tag)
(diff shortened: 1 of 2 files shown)


src/token-iterator.coffee
```diff
@@ -32,7 +32,11 @@ class TokenIterator
       tag = tags[@index]
       if tag < 0
         if tag % 2 is 0
-          @scopeEnds.push(atom.grammars.scopeForId(tag + 1))
+          scopeName = atom.grammars.scopeForId(tag + 1)
```
as-cii: Maybe we can name this `scope` as we do in a few lines later? :thought_balloon:

as-cii:
Janky is complaining for some reason, whereas Travis seems to be :green_heart:.

Just one comment, otherwise this looks pretty good to me! :+1:

nathansobo:
Good thinking. Obviously did this a bit rushed.

PR #8721, opened 2015-09-09
Add --no-auto-update while run `scripts/build` to disable autoupdate

scv119:
Add `--no-auto-update` to script/build as an option to disable Atom autoupdate.

This PR create a new grunt task that set `_enableAutoUpdate` to Atom's package.json while building. And in AutoupdateManager, it loads package.json and check if `_enableAutoupdate` is set.

I verified that `script/build --no-auto-update` will set `"_enableAutoUpdate" : false` in Atom's package.json whereas `script/build` will  set `"_enableAutoUpdate" : true`. 

However, it seems that the build generated by `script/build` won't trigger autoupdate even if I made the the version to be `1.0.9` instead of `1.0.9-somehash`.

Any thought on how to verify the autoupdate feature using `script/build`?

cc @bolinfest  @maxbrunsfeld  @kevinsawicki

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -36,6 +36,7 @@ module.exports = (grunt) ->
   buildDir = grunt.option('build-dir') ? path.join(tmpDir, 'atom-build')
   buildDir = path.resolve(buildDir)
   installDir = grunt.option('install-dir')
+  disableAutoUpdate = grunt.option('no-auto-update') ? false
 
   channel = grunt.option('channel')
   channel ?= process.env.JANKY_BRANCH if process.env.JANKY_BRANCH in ['stable', 'beta']
@@ -155,7 +156,7 @@ module.exports = (grunt) ->
   grunt.initConfig
     pkg: grunt.file.readJSON('package.json')
 
-    atom: {appDir, appName, symbolsDir, buildDir, contentsDir, installDir, shellAppDir, channel}
+    atom: {appDir, appName, symbolsDir, buildDir, contentsDir, installDir, shellAppDir, channel, disableAutoUpdate}
 
     docsOutputDir: 'docs/output'
 
--- a/build/tasks/build-task.coffee
+++ b/build/tasks/build-task.coffee
@@ -190,4 +190,5 @@ module.exports = (grunt) ->
     dependencies = ['compile', 'generate-license:save', 'generate-module-cache', 'compile-packages-slug']
     dependencies.push('copy-info-plist') if process.platform is 'darwin'
     dependencies.push('set-exe-icon') if process.platform is 'win32'
+    dependencies.push('disable-autoupdate') if grunt.config.get('atom.disableAutoUpdate')
     grunt.task.run(dependencies...)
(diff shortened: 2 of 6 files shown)


scv119:
I noticed that there is no way to just check the update but not update it without touching electron code.  So I instead disable the periodically check if no-autoupdate is set, whereas user could still click the check update button to trigger autoupdate manually.

Also, do you have any suggestions on how to test that? I tried to hardcode the version number as `1.0.3` instead of something like `1.0.12-hash` before building, but the autoupdate just doesn’t do anything.

maxbrunsfeld:
> I tried to hardcode the version number as 1.0.3 instead of something like 1.0.12-hash before building, but the autoupdate just doesn’t do anything.

If the app's version is just `1.x.y`, the auto-update *should* run. What is the output of this command from your built version of Atom?

```js
remote.require("app").getVersion()
```

What you've done looks good so far. There are a couple of changes I'd like you to make:

* [ ] Rather than passing the `@resourcePath` into the `AutoUpdateManager`, so that it can read the `package.json` and possibly disable itself, I would rather pass a boolean `disabled` flag into the `AutoUpdateManager` if auto update has been disabled in the package.json, or the version string has a SHA in it.
* [ ] When Atom has been built with `--no-auto-update`, the 'Atom' application menu should not have a `Check for Update` option. You could add an `isDisabled` method to `AutoUpdateManager` that can be checked in [`ApplicationMenu::getDefaultTemplate`](https://github.com/atom/atom/blob/f2aeb635fcea93e0ed6b72d1e56294cba6639274/src/browser/application-menu.coffee#L123).

scv119:
@maxbrunsfeld  ready for review again.

I still can't trigger autoupdate, here is what I did:
1) I checkout tag v1.0.13
2) I apply this patch to v1.0.13 https://github.com/scv119/atom/commit/e9e46f9fbf6c958eaf768ba65f3899651b9d9769
3) run `pkill Atom && rm -rf /Applications/Atom.app`
4) run `./script/build` to get Atom.app
5) open `/Applications/Atom.app`

Here I verified the version is `1.0.1`, but the `Check for Update` doesn't do anything if I click it.

scv119:
Here is how I tested:
* apply this patch (https://github.com/scv119/atom/commit/32c758b0fafc6a2631e3c609c9b894bcda7b70cd) to the commit.
* run `./script/build` to get Atom.app
* verified that `2015-09-22 11:05:05.692 Atom[93618:5819640] called` showed up while starting Atom using `atom -f`, and this log also showed up when `Check for Update` is clicked.
  * However, this didn't actually autoupdate Atom, which I suspect something related to this error `Error Downloading Update: Could not get code signature for running application`.
* run `./script/build --no-auto-update` to get Atom.app without autoupdater.
* verified that `.... called` didn't show up while starting Atom using `atom -f` 
* verified that `Check for Update` didn't show up in Atom menu.

maxbrunsfeld:
> Here is how I tested

@scv119 :+1: Ok, that sounds good to me.

We can merge this as soon as you resolve the conflicts.

scv119:
@maxbrunsfeld  rebased :)

maxbrunsfeld:
Thanks @scv119!

Issue #8540, opened 2015-08-27
[Metabug] Find a proper solution for the "z-index problem"


alexandernst:
- https://github.com/atom/atom/pull/8537
- https://github.com/atom/atom/pull/8098
- https://github.com/atom/atom/pull/8112
- https://github.com/atom/git-diff/issues/71

Those feel dirty/hacky (@as-cii agrees, I believe), so a better/proper solution should be found.

simurai:
Also #6857

The `position: fixed` solution from @Varanas [comment](https://github.com/atom/atom/issues/6857#issuecomment-128733330) might work. But only for things that keep position and are not part of a scrollview.

PR #8730, opened 2015-09-09
Faster Style Recalculations

as-cii:
- [x] Reopen https://github.com/atom/git-diff/issues/71 after :ship:ping this.
- [x] Merge https://github.com/atom/git-diff/pull/73.
- [x] Merge https://github.com/atom/bookmarks/pull/39.

In this PR I have removed inline styles from lines and got rid of a couple of classes that were not used for styling. To do that, I sorted line and line numbers by their screen row.

Before (full reflow):

before-1
before-2

After (full reflow):

after-1
after-2

Please note that this regresses https://github.com/atom/git-diff/issues/71 because we cannot set inline z-indexes anymore on line nodes. @simurai: in the meantime, I feel like I'd be down to have the git deletion marker styled like:

deletion-marker

Or, alternatively, to avoid assigning an opaque background to the cursor line number.

Thoughts? :thought_balloon: 


alexandernst:
Seems like a good moment to think about https://github.com/atom/atom/issues/8540

nathansobo:
@as-cii Still seems like it might be worth it to try dropping tiles and going back to a fixed size single layer that we reposition infrequently at scroll seams.

as-cii:
> Still seems like it might be worth it to try dropping tiles and going back to a fixed size single layer that we reposition infrequently at scroll seams.

Yep, that's something we should introduce regardless of the measurement mechanism we will end up choosing. :+1: 

Speaking about this, I think that with the performance improvements in this PR, we'd still need to keep tiles around, as sorting line nodes wouldn't be enough to properly position lines as we scroll. However, since all the problems related to tiling come from Chrome rules for subpixel anti-aliasing (i.e. background color, fixed height, etc.), we may simply avoid to promote tiles to rendering layers. This would allow to:

1. Keep this PR's performance optimizations derived from removing inline styles from line/line-number nodes.
2. Avoid problems like overlapping tiles (even if they overlap, they'll have no background).
3. Have a single `.highlights` container.
4. Perform a less invasive refactoring.

@nathansobo: what do you think?

nathansobo:
@as-cii So to be clear you're basically talking about moving to the single layer but leaving in the tiles as an interim step? We'll still need to solve the repositioning issue right? Otherwise we'll see our line rendering cut off after a few hundred thousand lines.

It seems reasonable to leave the tiles in place if it's not too big a drag on performance. If you think it would take a super long time to go back to plain lines rendering I'd be okay with it.

as-cii:
> We'll still need to solve the repositioning issue right? Otherwise we'll see our line rendering cut off after a few hundred thousand lines.

Exactly, repositioning must happen at an arbitrary scroll seam, as you correctly pointed out.

> So to be clear you're basically talking about moving to the single layer but leaving in the tiles as an interim step?

> It seems reasonable to leave the tiles in place if it's not too big a drag on performance. If you think it would take a super long time to go back to plain lines rendering I'd be okay with it.

I totally agree to move to a single layer, but part of the performance gainings in this PR are the result of deleting inline styles (that's why we resorted to ordering line nodes). If we remove tiles, however, we'd have no way of correctly positioning lines within the fixed sized layer without also re-introducing inline styles for each line.

By contrast, I am suggesting to keep tiles (assigning them an absolute position within the huge layer) so that we have to style only one element (i.e. the tile) instead of setting an inline style for each line (12 elements per tile).

In other words, tiles would become logical containers used solely for positioning (and performance) purposes. This means that we wouldn't promote them to rendering layers, thereby solving all those pesky issues related to subpixel anti-aliasing.

@nathansobo: specs are now :green_heart:. If we want to let this bake for a while, I'll create a new branch on top of this one, showing the single layer + tiles approach above discussed (hopefully in time for tomorrow morning, so that you can show it to the Chrome team). :balloon:

nathansobo:
> By contrast, I am suggesting to keep tiles (assigning them an absolute position within the huge layer) so that we have to style only one element (i.e. the tile) instead of setting an inline style for each line (12 elements per tile).

Ah ha. Makes sense. I'm fine merging this if it works and is performant. I'll build right now and run it for a while, but the best test is probably careful measurement so I'll trust you on it unless you want me to dive in.

simurai:
> I'd be down to have the git deletion marker styled like:

Yeah, I'm ok having the deletion marker a horizontal line instead of the triangle. Some might miss it, but can be added back in `styles.less` or in a theme.. if that selection glitch doesn't matter.

nathansobo:
> Yeah, I'm ok having the deletion marker a horizontal line instead of the triangle.

What if we just wait and until we go back to the single layer?

as-cii:
> What if we just wait and until we go back to the single layer?

I am afraid the single layer won't cut it. So long as two lines overlap (in this case, because of the git deletion maker crossing the line container) and one of them has an opaque background, we'll always face this kind of problem.

> I'll build right now and run it for a while, but the best test is probably careful measurement so I'll trust you on it unless you want me to dive in.

Performance-wise I am pretty confident this brings some decent speedups and allows us to get closer to the minimum threshold for reflows.

However, I'd prefer to make sure that the new line insertion logic works as expected: so far so good, but let's keep this open for a little more. In case anyone in the @atom/feedback is interested in trying this out it would be super appreciated! :pray:

PR #8749, opened 2015-09-10
:memo: CompositeDisposable is exported by event-kit, not atom

BrainCrumbz:


mnquintana:
Sorry for any confusion here, but `CompositeDisposable` *is* exported by `atom`, here: https://github.com/atom/atom/blob/master/exports/atom.coffee#L19

Importing it from Atom lets packages ensure they're using the same version as Atom core.

BrainCrumbz:
Oh, I see. Could you please clarify this to me: 
 1. in a file part of atom core, should I import *CompositeDisposable* from *atom* or from *event-kit* ? I think all references I found in atom core code import that from *event-kit*.

Thanks!

PR #8751, opened 2015-09-11
Fix a typo in Electron URL

Ingramz:


izuzak:
Thank you! :microscope:

Issue #8656, opened 2015-09-03
Atom does not use the system font on Linux (and Mac and Windows)


p-e-w:
Atom does not actually use the system font for its UI. [The "font stack" solution currently implemented](https://github.com/atom/atom/pull/8612) hardcodes a bunch of fonts in the hope that the default font for every possible installation is included. This is not a big issue on Mac and Windows where it is ~~impossible~~ almost impossible – and very unusual – to change the system font.

Not so on Linux. As an example, I use Fedora which comes with Cantarell by default. I don't like Cantarell and have changed my system font to Source Sans Pro instead (easily done in GNOME). All of my desktop applications respect the change, except Atom because it has Cantarell hardcoded in its font stack and since Cantarell is still present on my system it is used instead.

The solution is to make Atom use the system fonts like other desktop applications. Instead of guessing which font the user wants, it should query the system for the default font to use. IMO this should also extend to the fixed-width font and themes explicitly setting the font family should be the exception rather than the rule. No matter which theme I choose, there is almost never a good reason for one application to use a different font than the others.

mnquintana:
That's legit - we may have to wait on https://www.webkit.org/blog/3709/using-the-system-font-in-web-content/ to get standardized though. When it is, then using a future version of Electron we should be able to just set the UI font to `-system`.

paulcbetts:
We don't have to wait, we just have to add code to Atom to figure out the system font, It Can Be Done(tm)

peduxe:
Would be great, it's using Ubuntu font on mine and I would like to set SFNS or Noto.

benklebe:
Would it not be simpler to allow the user to change the UI font themselves?

thomasjo:
> Would it not be simpler to allow the user to change the UI font themselves?

Which can easily be done, just not via a GUI. But I think @paulcbetts approach is where the effort should be spent. Working on a GUI solution that would still require manual intervention by the user, is not worth the effort in my opinion. An automated approach is the way to go.

jerone:
To quote @mnquintana on [3 jul. 2015 20:13 CEST](https://github.com/atom/atom/pull/7698#issue-92918401 "2015-07-03T18:13:33Z - Replied by Github Comment Enhancer"):
> This will go a long way, I think, toward making Atom feel less like a "website" running in a semi-headless Chrome, and far more like a native app with a custom UI.

johnmuhl:
I haven't tested it across every platform but:

``` css
body { font: menu }
```

[...]

PR #8778, opened 2015-09-13
Use system menu font for Atom UI

mnquintana:
Closes https://github.com/atom/atom/issues/8656

If `@font-family` is undefined by a theme, then Atom will use the OS's default menu font in its place. Theme authors won't have to do anything unless they want their themes to use the OS default font - if that's the case, they need only :fire: their `@font-family` declaration.

This PR is pending changes to the One ~~and Atom~~ themes though for there to actually be an effect. :sweat_smile: 

Possible consequences:
- Anyone using SF UI on anything but El Capitan will be disappointed :disappointed: (since it'll now use the OS default - Helvetica Neue on Yosemite)
- It's possible that some OS font will have weird / unexpected consequences on `line-height` / kerning that we won't necessarily be able to anticipate. If that's the case though, themes can always override the OS default if necessary.

@simurai I was thinking of leaving Atom Dark & Light alone for now, [since they still have some issues with SF UI](https://github.com/atom/atom-dark-ui/pull/48#issuecomment-117368696) - what do you think?

Tested on OS X 10.10.5, Windows 8.1, Fedora 22, and Ubuntu 15.04

/cc @simurai @atom/feedback

### DIFF
--- a/static/variables/ui-variables.less
+++ b/static/variables/ui-variables.less
@@ -78,8 +78,3 @@
 @component-border-radius: 2px;
 
 @tab-height: 30px;
-
-
-// Other
-
-@font-family: '.SFNSText-Regular', 'SF UI Text', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
--- a/static/variables/ui-variables.less
+++ b/static/variables/ui-variables.less
@@ -78,3 +78,8 @@
 @component-border-radius: 2px;
 
 @tab-height: 30px;
+
+
+// Other
+
+@font-family: '';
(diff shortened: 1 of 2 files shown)


thomasjo:
LGTM! I say :ship: this including the One theme PRs.

---
*Keep in mind I'm still on vacation, so my brain is not running at 100%*

lee-dohm:
I agree with @thomasjo :rocket:

mnquintana:
Btw @atom/feedback, I think we should add this change to the release notes - it seems like something theme authors and users might wanna know about. :smile:

PR #8660, opened 2015-09-03
Avoid pairing null byte characters

as-cii:
Fixes #2537.

I am relatively new to this area, so feedback is super appreciated! :bow: 

When a document had null bytes, we were previously detecting whether such bytes were paired with other characters. Thus, if we had a string like the following in the buffer:

```
{someLetter}{nullByte}{variationSelector}
```

`TokenIterator::isPairedCharacter` returned true for `{nullByte}{variationSelector}`.

Although I _suspect_ it would be technically possible for null bytes to be paired with other unicode characters, as soon as we insert a null byte into the DOM it gets ripped off. 

This screwed up our measurement mechanism, which wasn't able to match DOM positions against the model. A solution for this could have been to substitute [this line](https://github.com/atom/atom/blob/master/src/lines-tile-component.coffee#L346) with:

``` coffee
continue if char.indexOf('\0')
```

However, this would skip the whole paired character and it's inconsistent with what gets displayed to the user. This PR changes the behavior of `isPairedCharacter` so that it skips null bytes when detecting pair code points.

/cc: @maxbrunsfeld @nathansobo @atom/feedback for :eyes:

PR #8783, opened 2015-09-14
Recycle DOM nodes

as-cii:
After meeting with the Chrome team, we have found out that our GC times were quite severe when many DOM nodes were appended and removed from the tree (e.g. a typical scenario where you can see this in practice is scrolling).

This is a chart that shows how **master** behaves when scrolling down for a while:

before

As you can see, the browser is triggering minor garbage collections during the scroll; at some point (which, _I think_, coincides with some Chrome threshold), a full GC is triggered and it takes quite a long time:

screen shot 2015-09-14 at 18 26 32

With this PR we are introducing a DOM elements pool, which efficiently allocates and deallocates nodes to minimize GC times. The result is the following:

after

screen shot 2015-09-14 at 18 25 47

:chart_with_downwards_trend: :racehorse: 

Moreover, we were able to ditch all the code that dealt with strings that contained null bytes (e.g. #8660) because now those characters are retained in the actual text nodes.

Some other pair of :eyes: are most welcome, especially to look out for performance regressions (it _feels_ like `innerHTML` was a bit faster, but I couldn't confirm this).

Thanks!


spec/dom-elements-pool-spec.coffee
```diff
@@ -0,0 +1,41 @@
+DomElementsPool = require '../src/dom-elements-pool'
+
+describe "DomElementsPool", ->
```
lee-dohm: Some tests around failure cases would be good, especially for such an important piece of infrastructure. I would suggest tests like:

* Free something that was already freed
* Free `undefined` and `null`
* Free a child and then free the parent and descendants
* Build `undefined` and `null`

If I can think of any others, I'll mention them.

as-cii: @lee-dohm: sounds good! Thank you for the suggestions, will add more specs shortly.

lee-dohm:
It appears that the memory usage of the elements pool will *never* go down. It should reach an upper bound, of course, but I'm still concerned that someone could accidentally reach an edge case and then their editor is :fire: Can you describe what a "worst case scenario" would look like? I'll take a stab at it ...

It appears that this pool will reach an upper bound based on the complexity of the UI on the screen. So some things to watch out for might be?

* Large screens? (Able to fit more stuff)
* More elements tightly packed? (Minimized JavaScript perhaps?)
* Lots of tabs open? (Or do non-visible elements get freed?)
* Lots of panes? (Same as above)

maxbrunsfeld:
> Moreover, we were able to ditch all the code that dealt with strings that contained null bytes.

:heart: Awesome.

Some minor style suggestions:
* I would call it `DOMElementPool` instead of `DomElementsPool`. We've been doing initialisms like `DOM` as uppercase in our code lately, and I think that singular noun seems to a more common convention for other pool-type objects, (e.g. "connection pool", "object pool").
* Is `DomElementsPool::free` meant to be called from the outside? If not, I would probably inline at its one call-site in `::freeElementAndDescendants`, to make it as clear as possible that you should always use the latter method.

spec/dom-elements-pool-spec.coffee
```diff
@@ -0,0 +1,41 @@
...
+  domElementsPool = null
+
+  beforeEach ->
+    domElementsPool = new DomElementsPool
```
nathansobo: Nitpick: `DOMElementsPool`

src/line-numbers-tile-component.coffee
```diff
@@ -106,9 +113,11 @@ class LineNumbersTileComponent
     else
       lineNumber = (bufferRow + 1).toString()
 
-    padding = _.multiplyString('&nbsp;', maxLineNumberDigits - lineNumber.length)
-    iconHTML = '<div class="icon-right"></div>'
-    padding + lineNumber + iconHTML
+    padding = _.multiplyString("\u00a0", maxLineNumberDigits - lineNumber.length)
+    iconRight = @elementsPool.build("div", "icon-right")
+
+    lineNumberNode.innerText = padding + lineNumber
```
nathansobo: Why this vs `textContent =`? Just curious.

src/lines-tile-component.coffee
```diff
@@ -114,64 +114,70 @@ class LinesTileComponent
...
-
-      lineHTML
+        invisible = endOfLineInvisibles[invisibleIndex++]
+        lineNode.appendChild(
+          @elementsPool.build("span", "invisible-character", invisible)
+        )
     else
-      @buildEndOfLineHTML(id) or '&nbsp;'
+      unless @appendEndOfLineNodes(id, lineNode)
+        lineNode.insertAdjacentText("beforeend", "\u00a0")
```
nathansobo: I think we only need to insert a non-breaking space on empty lines... or am I missing something. That might allow you to avoid this weird `insertAdjacentText` method. Don't care that much though.

nathansobo:
Love the way you structured this with the elements pool. I'm inclined to accept a tiny slowdown of line construction to avoid the big pauses. I'm going to build this and try it out, but in general I'm more worried about janky performance than slightly slower performance. But if you have ideas about how we can have the best of both worlds that would be good too.

src/lines-tile-component.coffee
as-cii: Thanks! :+1:

src/line-numbers-tile-component.coffee
as-cii: I guess I got confused with `innerHTML`: apparently `textContent` is faster and, moreover, is consistent with our codebase.

So, :+1: Thank you

as-cii:
> It appears that the memory usage of the elements pool will never go down. It should reach an upper bound, of course, but I'm still concerned that someone could accidentally reach an edge case and then their editor is  Can you describe what a "worst case scenario" would look like? I'll take a stab at it ...

@lee-dohm: that's a really good concern but I think this won't be worse than what we have currently. Let me explain why:

1. **Large screens**. Memory usage will reach an upper bound pretty quickly and, since the lines on screen will always be quite the same amount, we shouldn't waste memory. The worst case scenario here is when we haven't got much content on screen and, as a result, the pool is almost full. In practice, this could happen only at the end of a file when "scroll past end" is true.
2. **More elements tightly packed**. I think the same reasoning of 1) applies here. The upper bound of the pool is the number of lines that can fit on a screen (which, roughly, is also our lower bound, except in the worst case scenario mentioned above).
3. **Lots of tabs/panes open**. Currently, when we have more than one tab open, the ones that are hidden still retain all the nodes inside them and, as a result, we shouldn't see any difference in memory usage.

What do you think? Am I missing something there? On the other hand, I realized we need to make sure that the pool gets disposed properly so that its nodes can be garbage collected when the editor is destroyed.

> Some minor style suggestions:
> 
> I would call it DOMElementPool instead of DomElementsPool. We've been doing initialisms like DOM as uppercase in our code lately, and I think that singular noun seems to a more common convention for other pool-type objects, (e.g. "connection pool", "object pool").
> Is DomElementsPool::free meant to be called from the outside? If not, I would probably inline at its one call-site in ::freeElementAndDescendants, to make it as clear as possible that you should always use the latter method.

---------------------------------------------

> Love the way you structured this with the elements pool. I'm inclined to accept a tiny slowdown of line construction to avoid the big pauses. I'm going to build this and try it out, but in general I'm more worried about janky performance than slightly slower performance. But if you have ideas about how we can have the best of both worlds that would be good too.

Thanks for the suggestions guys! :sparkles: Eventually I went with `DOMElementPool`, but kept `::free` as a private method (i.e. not testing it in the specs) 'cause it felt a bit cleaner code-wise.

Overall I feel pretty good about this, exception made for the lines construction routines which _seems_ a bit slower. I'd like to hear your feedback as well after you try this out; in the meantime, I am going to research how we can minimize those times when/if we resort to a background thread.

lee-dohm:
> What do you think? Am I missing something there?

I see now that you're releasing the freed elements in the `build` method, so that takes care of my concern :grinning:

abe33:
Hi there, nice work! Object pools have always been an efficient way to fix performances problem and have proven to be very effective (I'm using it for both pigments and in my csv editor package and it achieved great results). 

I think, in the case of text editor tiles and lines, I can imagine a much more worst scenario even if it won't happen on each run:

When changing the font size from the settings view, if you don't type the size quickly enough you can end with lines with a height of 1px for a short moment. In that case your pool will reach an amount of elements completely meaningless after you revert to a proper font size. It'll also occurs if you unintentionally change the font size with the other available means, but I believe it will have less impact as the change should be less radical than when going from 14 to 1 and then to 12 for instance.

I frequently faced these kind of sudden burst of instances back when I was working in the flash game industry. You can have special events in a game that trigger a massive amount of particles (and we all love that :grin:). But to avoid having a too huge memory footprint after that I was frequently capping the freed instances collection based on data collected during the testing.

as-cii:
Thank you for sharing this super useful insight! That's an edge case I hadn't think of and which we should definitely address.

I think we have a couple of ways in which this could be hypothetically fixed:

1. Clearing out `DOMElementPool` when `fontSize` or `lineHeight` change. This is the most straightforward in my opinion and exploits the assumption that those parameters won't change that frequently under typical usage. 
2. Setting an absolute upper bound, as @lee-dohm proposed. This needs to be configured, though, and there's the risk of choosing a value that's not quite right.
3. Setting a relative upper bound. The idea here could be to work with ratios, configuring the pool to rebalance automatically once the ratio reaches a certain value (i.e. `ratio` here could be defined as `freedElements / builtElements` for a given series of moments `t0, t1, t2...`). Although this is probably the best solution possible, I'd prefer to keep the pool code simple and straight to the point unless some more complex interaction becomes mandatory.

I went ahead and implemented 1) in daf4316. @abe33: your experience with games would be super helpful here, so please let me know if you think there could be anything wrong with this solution. Thanks again! :+1:

benogle:
@as-cii :heart: :green_heart: :yellow_heart:

abe33:
@as-cii Solution 1 seems fine too. 

Solution 2 woud require some metrics to be really accurate and solution 3 seems more complex to me without having good metrics to analyze. 

For instance if I remove all the code in a file we'll end with a ratio used/freed of 1/N where N is the number of lines used before the deletion, but we should not rebalance anything because this is kinda the expected result in this case, but at the same time, the same ratio could be obtained in the scenario I described above but in that case a rebalance would be required.

as-cii:
Thanks everyone! :heart:

jeancroy:
If the pool re-balance become an issue I'll throw one idea:

At each generation, free a small percent of the pool. 
That way you'll avoid steady state error without having to maintain complicated statistic about ideal condition.

Suppose you free 1% of the pool per generation. Then suppose some event occurs that drastically reduce the number of node required (say snap to half of screen). After about 300 generations you'll come to about 5% of the new operating condition whatever that new condition is (0.99^300≈0.05) .

If you sync those generation with screen refresh (say about 60 per second), it means you'll adapt in about 5 seconds to the new condition. Which is relatively short in term of wasted memory for the user, but at the same time an eternity if you compare to the 30ms timescale of the garbage collection screenshot.

PR #8788, opened 2015-09-14
Add railcar script

maxbrunsfeld:
This adds a script, `railcar` that merges branches and updates version numbers in the way that we've talked about. With this script, the workflow for doing regularly scheduled releases will be:
1. Check that all CI builds are green on `beta` and `master`
2. Run `script/railcar`
3. Wait for all CI builds to finish on `stable` and `beta`
4. Click `Publish Release` on the two current draft releases (both normal and pre-release)

### DIFF
--- a/script/railcar
+++ b/script/railcar
@@ -1,103 +1,89 @@
 #!/usr/bin/env node
 
+var fs = require('fs')
 var exec = require('child_process').exec
-var async = require('../build/node_modules/async')
+var series = require('async').series
+var semver = require('semver')
 
-var branchName, leadingBranchName
-
-async.series([
-  parseArguments,
+series([
+  section('Preparing to roll the railcars'),
   checkCleanWorkingTree,
-  checkoutBranch,
-  pullUpstreamChanges,
-  mergeLeadingBranch,
-  fetchUpstreamTags,
-  updateVersionNumber,
-  pushLocalChanges,
-  pushLocalTags,
+  run('git fetch origin master:master beta:beta stable:stable'),
+  run('git fetch origin --tags'),
+
+  section('Updating stable branch'),
+  run('git checkout stable'),
+  run('git merge --ff-only origin/stable'),
+  run('git merge --ff-only origin/beta'),
+  bumpStableVersion,
+
+  section('Updating beta branch'),
+  run('git checkout beta'),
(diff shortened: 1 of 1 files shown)


script/railcar
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env node
+
+var fs = require('fs')
```
kevinsawicki: This appears to be unused

script/railcar
```diff
@@ -0,0 +1,102 @@
...
+      break
+    default:
+      return next(new Error('Usage: script/railcar <beta|stable>'))
+  }
+  next()
+}
+
+function checkCleanWorkingTree(next) {
+  console.log('Checking for a clean working tree');
+  exec('git status --porcelain', function(error, output) {
```
kevinsawicki: Is the `error` here handled? It looks like `next` might be being called with `null` even when `error` is present since `output` will be an empty string.

maxbrunsfeld:
Ok, after talking with @nathansobo about this, I've rewritten it so that it updates the `stable` and `beta` branches and tags in one shot. This allows us to more easily maintain the invariant that `master` is a fast-forward of `beta` and `beta` is a fast-forward of `stable`.

script/railcar
maxbrunsfeld: This comment was addressed, but then it popped back up after back after I rewrote the script. :unamused:

nathansobo:
:sparkles: Love the literate style. This is a work of :art:.

script/railcar
```diff
@@ -0,0 +1,89 @@
...
+  if (error) {
+    console.log('Error: ' + error.message)
+    process.exit(1)
+  }
+
+  console.log('OK, now just wait for all CI builds to pass on beta and stable')
+}
+
+function getCurrentVersion () {
+  return JSON.parse(fs.readFileSync(require.resolve('../package.json'))).version
```
kevinsawicki: Could this just be `require('../package.json').version` instead?

maxbrunsfeld: I considered doing that, but wouldn't `require` cache the content? Then I'd get the same version next time this method was called, even after changing branches.

kevinsawicki: Yeah, it definitely will, didn't know it needed to be called twice in one run.

PR #8811, opened 2015-09-16
DOM-based measurements

as-cii:
Supersedes and closes #6083.
Supersedes and closes #7945.
## A Brief Flashback

Over the past month we have evaluated several strategies to attack the problem, among which:
#### iframe

Build the lines inside an `<iframe>` element and measure from them.

**Pros**: measurements happen in a separate context, completely decoupled from the current DOM. Some optimization can, therefore, be applied: skinnier DOM, less stylesheet, and so on.
**Cons**: building the lines twice has a cost which is higher than the optimizations we tried out. Moreover, we cannot alter the original DOM structure as, otherwise, measurements could be wrong. 
#### Canvas

Have a canvas element and use the `measureText` function to provide all the metrics we need to position visual elements.

**Pros**: measurements happen in a separate context and the overhead is minimal.
**Cons**: CSS typography is not fully supported (e.g. ligatures/contextual alternates get screwed up) and the risk of reporting corrupted measurements is very high.
#### No Measurements

Instead of positioning decorations, cursors, etc. with absolute coordinates, embed them inside each line and let the browser figure out all the metrics.

**Pros**: no need for measurements and more semantical HTML. In addition, this enriches the power of styling customizations (e.g. the user has the possibility to style a piece of text surrounded by a decoration).
**Cons**: worse performance, because the DOM gets more complicated and nesting increases. Overlays need to be displayed on top of lines node, which forces us to employ tricks to let them appear in front of everything. 
## The Present

In this PR we have implemented DOM-based measurements. The idea here is to append the real lines to the DOM, measure from them, and render everything else that needs measurements. 

Every time `TextEditorElement::pixelPositionForScreenPosition` or `TextEditorElement::screenPositionForPixelPosition` is called, we trigger a reflow (a bit like the DOM does when you try to read from it when its information is stale).

In `TextEditorPresenter` we update the DOM with all lines we'll need to measure before computing any state, in order to avoid multiple reflows caused by reading from line nodes more than once.

What's missing:
- [x] Autoscroll when reaching the right end of the editor.
- [x] Implement `screenPositionForPixelPosition` in `LinesYardstick`.
- [x] Fix remaining specs.
- [x] Avoid using `NodeIterator` as it is slow. Cache text nodes and measure those instead.
- [x] Evaluate if caching already computed positions yields some improvements.
- [x] Remove measurement in `TextEditorComponent::readAfterUpdateSync`.
- [x] Ensure we have no significant regression performance-wise.
- [x] Investigate why Hasklig behaves oddly.
- [x] Fix bug where the cursor disappears when moving it left/right.
- [x] :arrow_up: autocomplete-plus.
- [x] :arrow_up: wrap-guide.

Feedback is super welcome and appreciated! Thanks! :bow: 

/cc: @nathansobo @atom/feedback for :eyes:


spec/text-editor-presenter-spec.coffee
```diff
@@ -832,11 +889,46 @@ describe "TextEditorPresenter", ->
...
+          presenter.getState().content.tiles[tileRow]?.lines?[lineId]
 
         tiledContentContract (presenter) -> presenter.getState().content
 
+        it "contains the state for the longest tile on screen", ->
+          presenter = buildPresenter(explicitHeight: 4, scrollTop: 0, lineHeight: 1, tileSize: 2)
+
+          expectValues presenter.getState().content.tiles[6], {
+            visibility: "hidden"
+          }
```
benogle: can we get rid of these curlies?

benogle:
Excited for this!

spec/text-editor-presenter-spec.coffee
nathansobo: Does it really matter? Looking forward to ES6. :wink:

as-cii: I am okay with removing them but this way the code feels a bit more consistent (i.e. the other specs use `expectValues` with curly brackets). 

What do you think if we just keep them for now?

spec/text-editor-presenter-spec.coffee
```diff
@@ -209,10 +220,36 @@ describe "TextEditorPresenter", ->
         expect(stateFn(presenter).tiles[0]).toBeDefined()
 
     describe "during state retrieval", ->
-      it "does not trigger onDidUpdateState events", ->
+      it "does not trigger ::onDidUpdateState events", ->
         presenter = buildPresenter()
         expectNoStateUpdate presenter, -> presenter.getState()
 
+      it "triggers ::onWillMeasure events before computing any state that needs measurement", ->
```
nathansobo: Think I asked this question on another PR with a similar method but... shouldn't this method be called like `onWillNeedMeasurements` or something?

as-cii: Sounds good! :+1:

spec/text-editor-presenter-spec.coffee
maxbrunsfeld: :+1:

as-cii:
Travis is playing a bit strangely with the network today, but specs should be :green_heart: now and I feel like this is ready to be reviewed. :tada: 

I'll have a second look with fresh :eyes: tomorrow, but I feel pretty good about the code changes (I am sure there's room for improvement, though, especially in `TextEditorPresenter`). I have this feeling sometimes that the presenter wants to be split into smaller objects, but I haven't come up with a cohesive design yet (of course this is outside the scope of this PR).

This definitely needs to be tested out and I'd like to make sure we have no evident regression (e.g. :bug:s  or slowdowns). I am so happy that after this PR we should be able to :ship: subpixel font scaling pretty smoothly! :sparkles: 

Some more :eyes: are most welcome on this, and I'd like to invite @atom/feedback to try this out as well: ideally you shouldn't notice any change from **master**, exception made for ligatures and kerning which should now work properly.

-------------------------------

@larsenwork: this should fix many of the issues you encountered with your font (e.g. #3821), so I'd :heart: to have your feedback on this as well. You should be able to test drive it by:

* Cloning this repository;
* Checking out this branch;
* Following the [build instructions for your platform](https://github.com/atom/atom/tree/master/docs/build-instructions);
* Running the atom version you just built;

Thanks!

src/text-editor-presenter.coffee
```diff
@@ -595,33 +641,24 @@ class TextEditorPresenter
...
-    visibleLineNumberIds = {}
+  isSoftWrappedRow: (bufferRow, screenRow) ->
+    return false if screenRow is 0
 
-    if startRow > 0
-      rowBeforeStartRow = startRow - 1
-      lastBufferRow = @model.bufferRowForScreenRow(rowBeforeStartRow)
-    else
-      lastBufferRow = null
+    @model.bufferRowForScreenRow(screenRow - 1) is bufferRow
```
as-cii: @nathansobo: I feel like we should be able to optimize this. Instead of converting screen rows, do you think we could cache this information into `TokenizedLine` when we soft wrap a line in `DisplayBuffer`? This call is quite expensive in terms of performance, and it grows as the file gets longer.

Until we implement a faster `RowMap` I feel like we should be restrictive about calling those methods.

benogle:
> shouldn't notice any change from master, exception made for ligatures and kerning which should now work properly.

:boom:

alexandernst:
Extremely nice to see this happening. One little question: will ligatures work out-of-the-box or will there be some kind of option in the settings/change in the config file that should be enabled/made for ligatures to work?

src/text-editor-presenter.coffee
as-cii: I went ahead and implemented this in 116f92d, as it seemed quite straightforward. Please, let me know if you think this could have any unintended side effects.

izuzak:
I'm seeing this exception when testing this branch:

screen shot 2015-10-01 at 11 58 15

Step to reproduce:

1. Copy this text

  ```
क ख ग घ ङ च छ ज झ ञ ट ठ ड ढ ण त थ द ध न प फ ब भ म य र ऱ ल ळ व श ष स ह क़ ख़ ग़ ज़ ड़ ढ़ फ़ य़ ॹ ॻ ॼ ॾ ॿ
```

2. Open a new tab in Atom
3. Paste the text a few times.
4. Use the mouse to position the cursor near the beginning of line 2
5 Press `delete` a few times so that line 2 merges with line 1
6. Select all text (`cmd-a`)
7. Press `delete`

GIF:

exc2


Wasn't able to trigger it on `master`. Are you seeing the same, @as-cii?

izuzak:
@as-cii Also, do you know which issues in this repository this PR should fix (other than #3821), so that we can perhaps verify/test that this PR fixes those issues? :pray:

izuzak:
I was testing to see how #3821 worked now (Hasklig font + `text-rendering: optimizeLegibility;`), and noticed this behavior:

exec3

When I try to go over the `===` character from right side to left side, the cursor stays in place for the first two key presses, and then goes over the character on the third key press (notice how the position of the cursor changes in the status bar). After that, I hop over to the right side again, press `delete` and the cursor jumps back in front of the character.

I'm not sure if this PR is supposed to address this or if this is even a problem, but wanted to point it out as something I noticed. :v:

izuzak:
Also notice this problem with the cursor vanishing:

exec4

I type `======` then press `left`.

as-cii:
First off, thanks @izuzak for the thorough feedback! :pray: 

So, most of my tests were performed using `Monoid`, which seemingly works as expected:

monoid

On the other hand, Hasklig behaves differently (i.e. exactly as @izuzak pointed out) also on Chrome Canary (e.g. tests performed on a `<textarea>`):

hasklig

As you can see, the cursor cannot jump between physical characters. I still have to understand why we have this substantial difference between the two fonts (@larsenwork: maybe you can tell us more about this?), but I am not sure we can do much better than this unless we start to insert "breaks" in the HTML (which is a completely different beast, I think).

Nevertheless, and aside from the cursor disappearing (I'll fix this in a while), I still think this is a huge improvement over **master**, where fonts like Hasklig or Monoid or even subpixel font scaling messed up measurements on the whole line:

master

What do you think?

izuzak:
> Nevertheless, and aside from the cursor disappearing (I'll fix this in a while), I still think this is a huge improvement over master, 

Definitely -- it is a huge improvement :+1: :purple_heart:. 

Let me know when you're done with those changes -- I'd be happy to test this PR one more time.

as-cii:
> Let me know when you're done with those changes -- I'd be happy to test this PR one more time.

The hidden cursor issue is a tricky one, as it's kinda difficult to reproduce it through a test case. I'll gather some more insights (also about the Hasklig font) tomorrow: in the meantime you can already try this out as every other issue should be fixed and your super helpful feedback would be truly appreciated, @izuzak. :sparkles:

izuzak:
> in the meantime you can already try this out as every other issue should be fixed and your super helpful feedback would be truly appreciated, @izuzak.

:cool: sounds good. Here's one I found on this branch (latest commit) but couldn't reproduce on master:

lorem

Notice how when I trigger soft-wrap -- the width of the window isn't updated and I can still scroll.

izuzak:
Sorry, I should have mentioned this in the comment above: the soft wrap problem is present when I use the Hasklig font and `text-rendering: optimizeLegibility;`. It is _not_ present when I use the default font and remove `text-rendering: optimizeLegibility;` from my styles.less file.

as-cii:
In the latest commit I have introduced subpixel font scaling: I wanted to defer this to a later moment, but this is mandatory if we want to fully support `optimizeLegibility`. Specs are :x: because `wrap-guide` and `autocomplete-plus` need to be upgraded in order to support subpixel precision (I have already opened two PRs to do that).

Also, please note that with this PR we will fix cursor position when there are RTL characters (i.e. arabic):

arabic

This, of course, doesn't mean we provide a complete RTL editing experience, but I feel like this is a nice improvement towards it.

@izuzak: all the other issues (i.e. hidden cursor, weird soft-wrapping behavior) should have been fixed, so it would be awesome if you could give this another try! :pray:

izuzak:
> @izuzak: all the other issues (i.e. hidden cursor, weird soft-wrapping behavior) should have been fixed, so it would be awesome if you could give this another try!

:zap: @as-cii -- I couldn't find any other problems and confirmed those :point_up: were fixed.

nylki:
Has anybody tested if different font-sizes in one document behaves better with the new changes? Previously there where some wiered cursor positionings and line overlappings etc.

as-cii:
> I couldn't find any other problems and confirmed those  were fixed.

@izuzak: :green_heart: :heart: :purple_heart: :blue_heart:. If you happen to test this once more, it'd be great if you could confirm there are no performance regression and Atom is still fast. :zap: :racehorse: 

> Has anybody tested if different font-sizes in one document behaves better with the new changes? Previously there where some weired cursor positionings and line overlappings etc.

@nylki: as far as overlapping lines are concerned, this won't actually alter the existing behavior, but cursor positioning should improve with the changes in this PR. Maybe you could point out at some specific scenarios you had in mind so that we can test them out as well? (bonus points :sparkles: if you could give some feedback by building it locally :heart:)

Anyway, specs should now pass :green_heart: and I think we can :shipit: as soon as we have asserted there are no performance regression (I haven't noticed anything so far) and code-wise this looks good.

Thank you!

nathansobo:
Because lines are grouped into tiles that are absolutely positioned by multiplying the start row of the tile by the height of a "typical" line, we'll still have issues dealing with varying line heights, which will still cause troubles with changing line heights. It's something we'll need to address separately and it could be a while until we're ready, but definitely open to supporting it in the long run.

pradyunsg:
:heart:

wismill:
Sounds very good! I am testing it with the amazing font [PragmataPro](http://www.fsd.it/fonts/pragmatapro.htm).

I highly recommend following the [instruction from Laurian](https://github.com/atom/atom/issues/3821#issuecomment-110017100) that deactivate the ligature on the current line, in order to see exactly on which character the cursor is positioned while moving over a ligature ([some ligatures](http://www.fsd.it/fonts/pragmatapro/OpenType_features.png) involve 3 characters).

A package could be very useful to manage `text-rendering: optimizeLegibility;` and the option to deactivate it on the current line: simple options on/off.

nathansobo:
Bravo @as-cii. An incredible amount of work went into this PR, much of it not even visible in the commit history. This is the first big step toward good support for internationalization!

benogle:
:boom:

maxbrunsfeld:
:fireworks: This is huge!

50Wliu:
Yesss! :100:

kevinsawicki:
So awesome :+1: :+1: :+1:

rajadain:
Ligature support! Incredible! I've been waiting for this for more than a year!

dikiaap:
:metal:

PR #8817, opened 2015-09-17
Use system menu font for Atom UI (more robustly)

mnquintana:
Refs https://github.com/atom/atom/pull/8778, https://github.com/atom/one-light-ui/pull/38, https://github.com/atom/one-dark-ui/pull/98

This PR fixes a few oversights from my last pass at this by properly overriding Bootstrap's default font styles for tooltips, and providing a new `'system'` fallback for `@font-family`. Omitting `@font-family` from bundled themes was breaking the build before, because these fallback variables weren't being loaded for bundled themes (and only bundled themes) for some reason. 

Now theme authors should be able to either omit `@font-family` or set `@font-family` to `'system'` to leverage OS default fonts.

/cc @atom/feedback


simurai:
An issue with using `@font-family: 'system';` in themes might be that if packages try to use `@font-family`, it would use `system`, but since that doesn't exist, it would fall back to Chromium's default which is `Times New Roman`.. I think.

Maybe it's not a big issue since packages shouldn't need `@font-family`. It already gets inherited anyways.

Too bad `font-family: menu` doesn't seem to work.

static/workspace-view.less
```diff
@@ -22,7 +22,9 @@ atom-workspace {
   position: relative;
   color: @text-color;
   background-color: @app-background-color;
-  font-family: @font-family;
+  font: menu;
```
simurai: I think we can remove all the font properties here, since they are already set on `html`. I guess it got initially added because Bootstrap added a font to `body`, but that is now overridden https://github.com/atom/atom/blob/master/static/bootstrap-overrides.less#L25-L28, so it's not needed here anymore.

mnquintana:
> An issue with using @font-family: 'system'; in themes might be that if packages try to use @font-family, it would use system, but since that doesn't exist, it would fall back to Chromium's default which is Times New Roman.. I think.

Hmm yeah that's a good point... do you think this is a worthwhile tradeoff? I'm not sure if there's any way around this, since the logic here needs to use `@font-family` to preserve the theme "API".

lee-dohm:
I'm Not a CSS Guy&trade;, so maybe this is a stupid idea, but could we create a custom `@font-face` like in this topic from Discuss?

https://discuss.atom.io/t/how-can-i-select-plural-fonts-one-for-roman-and-another-for-italic-in-atom/20611/17?u=leedohm

Would that allow us to then assign that custom font face (containing `menu`) to `@font-family` and solve this issue?

static/workspace-view.less
mnquintana: Ah okay. I think we need to keep `font-size` and `line-height` though, because interestingly enough the `menu` shorthand also defines new a new `font-size` and `line-height`, not just a new `font-family`

mnquintana:
> could we create a custom @font-face like in this topic from Discuss?

Unfortunately no – `menu` is only supported in the `font` property, so we wouldn't be able to specify it as `src` in `@font-face`. :confused: 

I'll have to think on this one. This might end up requiring a totally different approach.

simurai:
Hey.. A custom `@font-face` might still be useful. It has the `local()` option to reference local fonts. But I guess it would need to be generated with the right system font name.

If you add `html { font: menu; }` and then do

```js
window.getComputedStyle(document.documentElement,null).getPropertyValue("font-family");
```

you get the system font. In my case it's `'.HelveticaNeueDeskInterface-Regular'`. Seems like the private name for `Helvetica Neue`.

So maybe it could be done something like this:

1. create a "dummy element"
2. add to DOM
3. add inline style of `font: menu`
4. pull out the `font-family` with `getComputedStyle`
5. generate a style sheet with a `@font-face` (like below) and inject it into the `<head>`

```css
@font-face {
  font-family: 'system';
  src: local('.HelveticaNeueDeskInterface-Regular');
}
```

Then themes could add `system` to the front that has highest priority

```
@font-family: 'system', '.SFNSText-Regular', 'SF UI Text', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
```

and it should pick up the font defined in `@font-face`, in my case: `'.HelveticaNeueDeskInterface-Regular'`.

static/variables/ui-variables.less
```diff
@@ -82,4 +82,4 @@
 
 // Other
 
-@font-family: '.SFNSText-Regular', 'SF UI Text', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
+@font-family: 'system';
```
kuon: Is it safe to use system as name?

I just read some adventure about the system font: https://medium.com/@mwichary/system-shock-6b1dc6d6596f#.v0lk6vj4x

mnquintana: @kuon Good point! This whole thing will need to be redone anyway - I'm gonna close this out for now till I have a better solution.

PR #8821, opened 2015-09-17
Add margin left to icons

faustinoaq:
I'm using Kubuntu 15.04 and Atom 1.0.11
I just added one line `margin-left: 1px` because when I use atom with half screen I can see the icons cutted.

Screenshot
Margin

### DIFF
--- a/static/lists.less
+++ b/static/lists.less
@@ -43,6 +43,7 @@
   }
 
   .icon::before {
+    margin-left: 1px;
     margin-right: @component-icon-padding;
     position: relative;
     top: 1px;


as-cii:
Thanks for this PR, @faustinoaq!

/cc: @simurai for :eyes:

simurai:
I'll try to reproduce it, but maybe it's related to the scrollbar on the right? What happens if you start typing so and the scrollbar disappears, are the icons still cut off?

faustinoaq:
I try `atom --safe .` and the problem remain.

screen2

as-cii:
:sparkles:

simurai:
I could reproduce it on Win `8.1` as well.

It seems to happen only with the One themes when the `font-size` is larger than `12px`. I guess it's because of a rounding/rendering error when converting `em` to `px`.

It should be fixed in the next Atom update.

@faustinoaq Thanks for taking the time to make a PR. :pray:

PR #8826, opened 2015-09-17
Read _atomModuleCache to get all the native dependencies while initializing package

scv119:
Currently we traverse through submodules recursively to get all native modules while loading a given atom package. This could be quite slow for some package like `nuclide-remote-projects`, which takes 14 seconds to load for the first time.

In this PR, we utilize the native dependency information generated from apm, (which lies in _atomModuleCache['extensions']['.node']), to get all the transitive native dependencies for given package. And fallback to old approach if the field `_atomModuleCache` doesn't exist in package.json.

cc @kevinsawicki


src/package.coffee
```diff
@@ -590,10 +590,21 @@ class Package
...
-  # This will recurse through all dependencies.
+  #
+  # First try to get this information from
+  # @metadata._atomModuleCache.extensions. If @metadata._atomModuleCache doesn't
+  # exist, recurse through all dependencies.
   getNativeModuleDependencyPaths: ->
     nativeModulePaths = []
 
+    if @metadata._atomModuleCache?
+      nativeModuleBindingPaths = @metadata._atomModuleCache.extensions?['.node'] ?[]
```
kevinsawicki: Space needed between `?` and `[]`

src/package.coffee
```diff
@@ -590,10 +590,21 @@ class Package
...
+      for nativeModuleBindingPath in nativeModuleBindingPaths
+        # The `.node` file lies in nativeModulePath/build/Release/ folder.
+        nativeModulePath = path.normalize(path.join(path.dirname(nativeModuleBindingPath), '..', '..'))
```
kevinsawicki: I believe `path.join` normalizes the result so you don't need an additional call to `path.normalize`

spec/package-spec.coffee
```diff
@@ -19,6 +19,11 @@ describe "Package", ->
       expect(pack.incompatibleModules[0].name).toBe 'native-module'
       expect(pack.incompatibleModules[0].path).toBe path.join(packagePath, 'node_modules', 'native-module')
 
+    it "utilizes _atomModuleCache to get native modules and skips traversing through submodules", ->
+      packagePath = atom.project.getDirectories()[0]?.resolve('packages/package-with-both-incompatible-and-compatible-native-modules')
+      pack = new Package(packagePath)
+      expect(pack.isCompatible()).toBe true
```
kevinsawicki: So if the package is named `package-with-both-incompatible-and-compatible-native-modules` I would expect the spec to assert that `pack.isCompatible()` is `false` since the words `incompatible-native-modules` are in the name of the package.

scv119:
Comments addressed @kevinsawicki 
Also, once the Atom is autoupdated (and it's buildin electron) , will Atom do anything automatically to rebuild all the native dependencies, or they just failed to load?

kevinsawicki:
> Also, once the Atom is autoupdated (and it's buildin electron) , will Atom do anything automatically to rebuild all the native dependencies, or they just failed to load?

There is a UI for rebuilding them after upgrade, https://github.com/atom/incompatible-packages/pull/9

scv119:
Woohoo, that looks great.

PR #8828, opened 2015-09-17
Add command for cutting text to the end of the cursor's buffer line

### LINKED ISSUE #2607: editor:cut-to-end-of-line should cut to end of actual line in soft wrap
^k only kills to the edge of the editor when soft wrap is on. I expect the behavior to be the same as if the file wasn't soft-wrapped - that is to say, a line's a line (and it's the line represented in the file, bounded by newlines), even if I have the editor temporarily showing it as three or four.

izuzak:
This is an attempt to close https://github.com/atom/atom/issues/2607. 

Currently, the `TextEditor.cutToEndOfLine` method and associated `editor:cut-to-end-of-line` command cut text from the cursor to the end of the cursor's _screen_ line. For some users, this is not the expected behavior when the cursor is in a soft wrapped line -- for those users, a more intuitive behavior would be to cut to the end of the cursor's _buffer_ line.

This PR adds a new method and command which cuts text to the end of the buffer line. Users can add a new keybinding for that or change the existing keybinding for `editor:cut-to-end-of-line` to point to the new command.

Notes:
- I decided not to change the current method `TextEditor.cutToEndOfLine` so that it cuts to the end of the buffer line because that would be considered a breaking change for that API. Still, if we consider this to be a bug in that API method, I'd be happy to change that method instead.
- I decided not to change the current command `editor:cut-to-end-of-line` so that it cuts to the end of the buffer line because that would might surprise users who are relying the current behavior. Also, `editor:select-to-end-of-line` select to the end of the screen line as well, not buffer line, so it made sense to keep these in sync (selecting and cutting to end of line should work with the same line concept so that it's not surprising for users). 
- I decided not to add a new keybinding for this command because it seemed a bit strange to have keybinding for both `editor:cut-to-end-of-line` and `editor:cut-to-end-of-buffer-line`. Normally, users wouldn't use both, they'd use one or the other. 
- I decided not to change the existing keybinding for `editor:cut-to-end-of-line` so that it calls `editor:cut-to-end-of-buffer-line` just so it doesn't surprise users who are used to the existing behavior.

I'm not 100% sure in any of those decisions above. I opened this PR to have a discussion about them -- happy to make changes if anyone feels differently.

cc @thedaniel since you opened that issue https://github.com/atom/atom/issues/2607 and @nathansobo @maxbrunsfeld for :thought_balloon: and :eyes:

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2899,6 +2899,35 @@ describe "TextEditor", ->
               expect(buffer.lineForRow(3)).toBe '    var pivot = item'
               expect(atom.clipboard.read()).toBe ' <= 1) ret\ns.shift(), current, left = [], right = [];'
 
+      describe ".cutToEndOfBufferLine()", ->
+        describe "when soft wrap is on", ->
+          it "cuts up to the end of the buffer line", ->
+            editor.setSoftWrapped(true)
+            editor.setEditorWidthInChars(10)
+            editor.setCursorScreenPosition([2, 2])
+            editor.cutToEndOfBufferLine()
+            expect(editor.tokenizedLineForScreenRow(2).text).toBe '= '
+
+        describe "when soft wrap is off", ->
+          describe "when nothing is selected", ->
+            it "cuts up to the end of the buffer line", ->
+              editor.setCursorBufferPosition([2, 20])
+              editor.addCursorAtBufferPosition([3, 20])
+              editor.cutToEndOfBufferLine()
+              expect(buffer.lineForRow(2)).toBe '    if (items.length'
+              expect(buffer.lineForRow(3)).toBe '    var pivot = item'
+              expect(atom.clipboard.read()).toBe ' <= 1) return items;\ns.shift(), current, left = [], right = [];'
+
+          describe "when text is selected", ->
+            it "only cuts the selected text, not to the end of the buffer line", ->
+              editor.setSelectedBufferRanges([[[2, 20], [2, 30]], [[3, 20], [3, 20]]])
+
+              editor.cutToEndOfBufferLine()
+
+              expect(buffer.lineForRow(2)).toBe '    if (items.lengthurn items;'
+              expect(buffer.lineForRow(3)).toBe '    var pivot = item'
+              expect(atom.clipboard.read()).toBe ' <= 1) ret\ns.shift(), current, left = [], right = [];'
+
       describe ".copySelectedText()", ->
         it "copies selected text onto the clipboard", ->
           editor.setSelectedBufferRanges([[[0, 4], [0, 13]], [[1, 6], [1, 10]], [[2, 8], [2, 13]]])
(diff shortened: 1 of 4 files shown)


nathansobo:
Looking good.

Ideally every method would be explicit with regards to `Screen` or `Buffer` if there were a chance for ambiguity. So if anything I would deprecate anything ambiguous and mention `Screen` explicitly if we ever wanted to rev our API again. Not sure it's really worth it right now though.

spec/text-editor-spec.coffee
```diff
@@ -2899,6 +2899,35 @@ describe "TextEditor", ->
...
               expect(atom.clipboard.read()).toBe ' <= 1) ret\ns.shift(), current, left = [], right = [];'
 
+      describe ".cutToEndOfBufferLine()", ->
+        describe "when soft wrap is on", ->
+          it "cuts up to the end of the buffer line", ->
+            editor.setSoftWrapped(true)
+            editor.setEditorWidthInChars(10)
+            editor.setCursorScreenPosition([2, 2])
+            editor.cutToEndOfBufferLine()
+            expect(editor.tokenizedLineForScreenRow(2).text).toBe '= '
```
nathansobo: You should probably make an assertion about the state of the buffer here, rather than talking about screen rows. Does this actually prove we cut to the end of the buffer line?

spec/text-editor-spec.coffee
```diff
@@ -2899,6 +2899,35 @@ describe "TextEditor", ->
...
+
+        describe "when soft wrap is off", ->
```
nathansobo: I might eliminate the soft wrap on / off sections and just test everything with soft wrap off.

nathansobo:
Aside from a few test adjustments, the implementation looks good to go.

thedaniel:
I personally consider it a bug in the implementation since we're editing files, not screens, and a line has a pretty well-defined meaning in a file, but I do understand that it could be perceived as a breaking change (by people that are wrong :trollface:), so the cautious approach described by @izuzak seems fine to me.

spec/text-editor-spec.coffee
izuzak: > Does this actually prove we cut to the end of the buffer line?

For this specific fixture and cursor placement -- it does. But you're right, this line doesn't communicate the intention of the test well. I've fixed this in bed902e -- thanks for pointing this out. :bow:

spec/text-editor-spec.coffee
izuzak: > I might eliminate the soft wrap on / off sections and just test everything with soft wrap on.

:sparkles: I agree -- I removed that in bed902e and now testing everything with soft wrap on.

izuzak:
Thanks for the review, @nathansobo and @thedaniel -- I made some changes based on the review comments. I'll let this sit for the rest of the day just in case anyone else has any thoughts and merge after that. :construction_worker:

izuzak:
> Ideally every method would be explicit with regards to `Screen` or `Buffer` if there were a chance for ambiguity. So if anything I would deprecate anything ambiguous and mention `Screen` explicitly if we ever wanted to rev our API again. Not sure it's really worth it right now though.

That ambiguity caught my eye as well, @nathansobo, and I think we should do that but I wouldn't say it's urgent. I'll add it to my list so that I can open an issue and perhaps tackle it myself. :memo:

Issue #8837, opened 2015-09-18
Can't rebuild package after updating to 1.0.13


burodepeper:
The 'red bug' appeared in the bottom right corner of the status bar after updating to 1.0.13, and led me to an UI where I could rebuild my package. Rebuilding kept failing with the following error message:
```
/Applications/Atom.app/Contents/Resources/app/apm/node_modules/npm/bin/node-gyp-bin/node-gyp: line 3: node: command not found

npm ERR! Darwin 14.5.0
npm ERR! argv "/Applications/Atom.app/Contents/Resources/app/apm/bin/node" "/Applications/Atom.app/Contents/Resources/app/apm/node_modules/npm/bin/npm-cli.js" "--globalconfig" "/Users/david/.atom/.apm/.apmrc" "--userconfig" "/Users/david/.atom/.apmrc" "rebuild" "--target=0.30.6" "--arch=x64"
npm ERR! node v0.10.40
npm ERR! npm  v2.13.3
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! oniguruma@4.2.2 install: `node-gyp rebuild`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the oniguruma@4.2.2 install script 'node-gyp rebuild'.
npm ERR! This is most likely a problem with the oniguruma package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls oniguruma
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/david/git/language-markdown/npm-debug.log
```

Eventually, after discussion in the #atom channel on Slack with @as-cii and @simurai, and after `apm rebuild`-ing numerous times via the command line, and after a re-install of Atom, rebuilding finally worked. So, my problem is fixed, but I suppose this isn't intended behavior.

The package that caused this was [language-markdown](http://atom.io/packages/language-markdown/).

`npm-debug.log` got unfortunately erased once the rebuild succeeded.

burodepeper:
And the bug is back.

simurai:
Had a similar issue with the `styleguide` package when opening Atom in "Dev Mode". I think also because of `oniguruma`. After trying a couple things, it eventually worked when clicking the rebuild button.

[...]

PR #8852, opened 2015-09-18
Hotfix: Use bundled Node when running `apm rebuild` from within Atom

nathansobo:
Fixes https://github.com/atom/atom/issues/8837


nathansobo:
The windows failure won't be an issue on stable and has a pending fix on master. Merging this and getting the fix out.

PR #8854, opened 2015-09-18
Create draft release with correct target branch

maxbrunsfeld:
This way, we won't accidentally tag the head of master when clicking 'Publish' for beta and stable releases.

/cc @nathansobo


nathansobo:
👍🏻

PR #8859, opened 2015-09-19
Fix coffelint error causing Travis builds to fail

izuzak:
Currently, running coffeelint produces this error and causes Travis builds to fail:

```
Running "coffeelint:src" (coffeelint) task
src/command-installer.coffee
  ✖  line 73  Line contains inconsistent indentation  Expected 2 got 0
✖ 1 error
Warning: Task "coffeelint:src" failed. Use --force to continue.
Error: Task "coffeelint:src" failed.
```

See clutchski/coffeelint#189 for more information on this problem and suggested approach, and for a different place in Atom where this approach is used see this:

https://github.com/atom/atom/blob/ca39c106d6570b7e46129dd445dde3b9ba84cf9f/src/config.coffee#L984

cc @mnquintana @kevinsawicki for :eyes: and :thought_balloon: in case there's a better way to fix this.

### DIFF
--- a/src/command-installer.coffee
+++ b/src/command-installer.coffee
@@ -70,7 +70,7 @@ module.exports =
           try
             error = null
             symlinkCommandWithPrivilegeSync(commandPath, destinationPath)
-          catch error
-            undefined
+          catch err
+            error = err
 
         callback?(error)
--- a/src/command-installer.coffee
+++ b/src/command-installer.coffee
@@ -71,5 +71,6 @@ module.exports =
             error = null
             symlinkCommandWithPrivilegeSync(commandPath, destinationPath)
           catch error
+            undefined
 
         callback?(error)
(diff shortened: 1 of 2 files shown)


as-cii:
Sorry guys for chiming in, but I think we have one more linting problem which should probably be addressed in this PR (see https://travis-ci.org/atom/atom/jobs/81269172):

```
src/workspace.coffee
  ⚠  line 936  Unnecessary fat arrow
```

(It seems like using `->` should be enough to fix it.)

As for the issue in `command-installer.coffee`, what about assigning the error variable in the `catch` block? e.g.

```coffee
catch err
  error = err
```

It's a bit redundant but on the other hand I like its explicitness (the first time I read the code it wasn't immediately clear to me that we were interested in assigning the error).

It seems okay to use `undefined` if we care about the return value (as I think we do in https://github.com/atom/atom/blob/ca39c106d6570b7e46129dd445dde3b9ba84cf9f/src/config.coffee#L984), but in this specific scenario it feels a bit misleading.

P.s. thanks @izuzak for addressing this! :clap:

izuzak:
Both those suggestions make sense to me, @as-cii -- we're now living in that future as of d798559 and fd9f483. :smile_cat:  Thanks for the suggestions!

nathansobo:
:+1:

Issue #8168, opened 2015-07-31
Atom does not recognize git worktree repositories


snaury:
When `git-repository-provider.coffee` searches for the git repository it looks for the `.git` directory, however `git worktree` creates repositories that have `.git` file that has `gitdir: path/to/gitdir` as contents. As a result atom cannot currently work with repositories created with `git worktree`. Unfortunately even making `.git` a symlink doesn't work, since worktree gitdir doesn't have `objects` or `refs` subdirectories.

Atom should recognize `.git` files that start with `gitdir:` as valid git repositories.

phord:
@snaury I am interested in this bug and I'm working on a fix, but for me it primarily affects submodules.  I'm not sure what you mean about your symlink solution failing.  If you can provide a simple script to setup the failure condition, it would help ensure that is covered, too.

phord:
@snaury Nevermind.  I see what you mean.  I didn't notice the new 'git worktree' command show recently.  The bug I was working on was only to fix the ".git-file" issue.   I think some more work needs to be done to support the new '.git/worktrees' arrangment.

phord:
I do have a fix for the first half of this bug.  But the worktree problem (still marked "experimental" in git documentation) is not supported by libgit2 yet, and I don't have a working fix for that.

fooishbar:
@phord libgit2 support is WIP: https://github.com/libgit2/libgit2/pull/3436

rymohr:
Just ran into this one too. It hasn't been a showstopper for me but it does cause find-in-project to search across .gitignore'd files.

sos4nt:
Any updates on this issue? I find `git worktree` quite useful and a proper support (editor wise) would be really nice.

50Wliu:
@sos4nt this issue is blocked until https://github.com/libgit2/libgit2/pull/3436 is merged.

kriomant:
Worktree support in libgit2 just landed!

50Wliu:
Nice!  Now I think we'll need to wait for a new libgit2 release containing worktree support, update atom/git-utils to use that new version, and then update atom/atom to use the new git-utils version (phew).

yao23:
I'm in top folder and it works only after save (cmd+s) the change, then it will show in git.

pleonex:
[...]

PR #8862, opened 2015-09-19
:bug: Teach Atom to read .git files

phord:
This fixes part of #8168.  The git-worktree model still is not supported, but this is because it is so new that libgit2 doesn't know about it yet, either.  But the .git-file feature which is heavily used by submodules works with this fix.

### DIFF
--- a/spec/git-repository-provider-spec.coffee
+++ b/spec/git-repository-provider-spec.coffee
@@ -11,7 +11,7 @@ describe "GitRepositoryProvider", ->
       it "returns a Promise that resolves to a GitRepository", ->
         waitsForPromise ->
           provider = new GitRepositoryProvider atom.project
-          directory = new Directory path.join(__dirname, 'fixtures/git/master.git')
+          directory = new Directory path.join(__dirname, 'fixtures', 'git', 'master.git')
           provider.repositoryForDirectory(directory).then (result) ->
             expect(result).toBeInstanceOf GitRepository
             expect(provider.pathToRepository[result.getPath()]).toBeTruthy()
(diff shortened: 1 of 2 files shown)


spec/git-repository-provider-spec.coffee
```diff
@@ -56,6 +56,21 @@ describe "GitRepositoryProvider", ->
           provider.repositoryForDirectory(directory).then (result) ->
             expect(result).toBe null
 
+    describe "when specified a Directory with a valid gitfile-linked repository", ->
+      it "returns a Promise that resolves to a GitRepository", ->
+        waitsForPromise ->
+          provider = new GitRepositoryProvider atom.project
+          gitDirPath = path.join(__dirname, 'fixtures/git/master.git')
```
maxbrunsfeld: For windows compatibility, I think this should be:

```coffee
path.join(__dirname, 'fixtures', 'git', 'master.git')
```

phord: @maxbrunsfeld I expect you are correct.  But I copied that line from a previous test (line 27).

maxbrunsfeld: I see. Sorry there are some bad examples in our test suite. Would you mind fixing this anyway, and adjusting the line that you copied it from in the same way?

maxbrunsfeld:
@phord Other than my one small piece of feedback, this looks great. Thanks for diving in and fixing this!

spec/git-repository-provider-spec.coffee
phord: I am happy to do so.

PR #8865, opened 2015-09-19
As single layer

hirk7788:


as-cii:
We are considering to switch to a single layer but the implementation isn't quite ready yet and this was only a spike to prove some concepts and collect some metrics.

Closing this for now.

PR #8873, opened 2015-09-21
Remove deprecated code

as-cii:
In #8680 we have removed some deprecated events but we were still performing some operations which should be now unnecessary.

Other than tightening the codebase, this should bring some speedups when adding selections as well, 'cause we don't need to call `editor.selectionIntersectsVisibleRowRange` anymore.

/cc: @nathansobo for :eyes:


nathansobo:
Thanks for going deeper on this one.

PR #8892, opened 2015-09-22
Throw exception when adding a pane item that has already been destroyed

nathansobo:
Closes atom/bracket-matcher#178

@izuzak: This will help place blame properly when we attempt to add a pane item that reports itself as destroyed.


izuzak:
:zap: Great -- thanks for fixing this, @nathansobo

Issue #8762, opened 2015-09-12
Uncaught TypeError: Cannot read property 'offsetHeight' of undefined


oscaroceguera:
[Enter steps to reproduce below:]
1. ...
2. ...

**Atom Version**: 1.0.11
**System**: Mac OS X 10.10.5
**Thrown From**: Atom Core
### Stack Trace

Uncaught TypeError: Cannot read property 'offsetHeight' of undefined

```
At /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:829

TypeError: Cannot read property 'offsetHeight' of undefined
  at TextEditorComponent.module.exports.TextEditorComponent.isVisible (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:829:26)
  at TextEditorComponent.module.exports.TextEditorComponent.measureLineHeightAndDefaultCharWidth (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:915:16)
  at TextEditorComponent.module.exports.TextEditorComponent.sampleFontStyling (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:896:14)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:3:61
  at /Applications/Atom.app/Contents/Resources/app.asar/src/config.js:587:20
  at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:82:11)
  at Config.module.exports.Config.emitChangeEvent (/Applications/Atom.app/Contents/Resources/app.asar/src/config.js:772:29)
  at Config.module.exports.Config.setRawValue (/Applications/Atom.app/Contents/Resources/app.asar/src/config.js:564:19)
  at Config.module.exports.Config.set (/Applications/Atom.app/Contents/Resources/app.asar/src/config.js:233:14)
  at Workspace.module.exports.Workspace.increaseFontSize (/Applications/Atom.app/Contents/Resources/app.asar/src/workspace.js:663:26)
  at atom-workspace.atom.commands.add.window:increase-font-size (/Applications/Atom.app/Contents/Resources/app.asar/src/workspace-element.js:164:30)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app.asar/src/command-registry.js:245:29)
[...]

PR #8895, opened 2015-09-22
Allow null TextEditorComponent::domNode during visibility check

as-cii:
...and use `atom.assert` to verify that the constructor was run before the exception.

Ref.: #8762 

/cc: @nathansobo @izuzak

### DIFF
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -34,6 +34,7 @@ class TextEditorComponent
   stylingChangeAnimationFrameRequested: false
   gutterComponent: null
   mounted: true
+  initialized: false
 
   constructor: ({@editor, @hostElement, @rootElement, @stylesElement, @useShadowDOM, tileSize}) ->
     @tileSize = tileSize if tileSize?
@@ -102,6 +103,7 @@ class TextEditorComponent
 
     @updateSync()
     @checkForVisibilityChange()
+    @initialized = true
 
   destroy: ->
     @mounted = false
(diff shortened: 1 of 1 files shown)


izuzak:
Thanks for tackling this, @as-cii :zap:

In https://github.com/atom/atom/issues/8762#issuecomment-142209322 you said:

> A package is setting the variable to null. This seems a bit unlikely, though.

Would it make sense to try and confirm or refute that theory as well (if you still think that's a possibility)? I'm not sure, but just wondering if it's possible to add another assert which would catch `@domNode` being set to `null` (e.g. would it be possible to wrap it somehow)?

as-cii:
Thanks for the feedback! :sparkles: 

I still feel like a package is not arbitrarily setting `::domNode` to `null`, but the assertion code is quite straightforward so the tradeoff seems reasonable. :+1: 

Please, have a last :eyes: and :ship: at will. Thanks!

src/text-editor-component.coffee
```diff
@@ -34,6 +34,13 @@ class TextEditorComponent
   stylingChangeAnimationFrameRequested: false
   gutterComponent: null
   mounted: true
+  initialized: false
+
+  Object.defineProperty @prototype, "domNode",
```
nathansobo: Does this perform well? That's my one concern with adding it.

nathansobo: I guess this isn't really a hot code path so whatever.

as-cii:
Thanks!

PR #8897, opened 2015-09-22
Use lowercase 'optional' keyword used by docs

ssorallen:
The docs generator expects 'optional' starting with a lowercase 'o' to
mark arguments as optional. Lowercase 'O' to properly generate the docs.


maxbrunsfeld:
Thanks @ssorallen!

PR #8905, opened 2015-09-23
Use logical coordinates in DisplayBuffer

as-cii:
This PR moves all the code that handles pixel positions into `TextEditorPresenter`, where it belongs. This is part of the effort to merge #8811, as we realized we needed the `DisplayBuffer` to be clean before proceeding with measurements.

This refactoring was quite big, so I'd :heart: to gather some feedback on the changes I have made (both on code quality and regressions).

What's missing:
- [x] Fix missing failing specs.
- [x] Fix package specs (by shimming the original API).
- [x] Maintain a scroll queue to preserve scroll history.
- [x] Serialize scroll state in terms of logical coordinates.
- [x] Confirm that packages work as expected.
- [x] Confirm that atom core works as expected.
- [x] Shim `TextEditor` APIs to use their `TextEditorElement` counterpart.
- [x] Fix discovered bugs.

Thanks! :bow:


src/display-buffer.coffee
```diff
@@ -324,84 +253,17 @@ class DisplayBuffer extends Model
...
-      if desiredScrollBottom > @getScrollBottom()
-        @setScrollBottom(desiredScrollBottom)
-
-      if desiredScrollLeft < @getScrollLeft()
-        @setScrollLeft(desiredScrollLeft)
-      if desiredScrollRight > @getScrollRight()
-        @setScrollRight(desiredScrollRight)
+  scrollToScreenRange: (screenRange, options = {}) ->
+    scrollEvent = {screenRange, options}
+    @emitter.emit "did-change-scroll-position", scrollEvent
```
nathansobo: How would you feel about expressing the position logically as `scrollTopRow`, and `scrollLeftColumn`, rather than ephemerally as an event. That would mean you could assign the row/column to request a scroll.

src/display-buffer.coffee
```diff
@@ -1314,6 +1154,25 @@ class DisplayBuffer extends Model
       # this one. Only emit when the marker still exists.
       @emitter.emit 'did-create-marker', marker
 
+  # TODO: serialize state in TextEditorElement, rather than saving scroll
+  # positions here.
```
nathansobo: Imagine if your font is a different size next time you load the editor... don't you actually want the same logical location rather than the same pixel location to be in view? That's what makes me wonder about expressing the scroll top as state in terms of rows... but maybe it's something that is only assigned by the view and read one time when it is attached to the DOM.

src/display-buffer.coffee
as-cii: > How would you feel about expressing the position logically as scrollTopRow, and scrollLeftColumn, rather than ephemerally as an event. That would mean you could assign the row/column to request a scroll, and also help with serialization.
> 
> Edit: On further thought, I'm less sure about this. It would require us to pull more view state into the display buffer. Maybe a 1-time request to scroll to a point is better.

That's how it looked like in the first commits: basically, I wanted to express scroll positions in terms of logical coordinates. At some point, however, I realized that scrolling is merely a view concern: in other words, if a view isn't attached to a `DisplayBuffer`, it's kinda weird to _talk_ about scrolling in that object.

And I think the same goes for scoped character widths or other similar concerns that are still in the display buffer: I feel like this is just a step towards removing all of that knowledge from an object whose purpose is to handle spatial coordinates independently of how they get presented to the user.

What do you think?

nathansobo: :+1:

src/display-buffer.coffee
as-cii: > but maybe it's something that is only assigned by the view and read one time when it is attached to the DOM.

:+1: for this. Ideally, I'd :heart: the view to have a serialized state of its own because of the reasons pointed out in https://github.com/atom/atom/pull/8905#discussion_r40288745, but I am :cool: with keeping the model up to date with the current logical coordinates as well.

as-cii:
Okay, the build is :green_heart: now and I think I have addressed all the concerns we had during our call. :tada:

I have found, however, a method on `Marker` which makes use of the display buffer's pixel conversion methods (i.e. `Marker::getPixelRange`) and a cursory search on GitHub showed that someone is actually using it :disappointed:.

We cannot shim it to a corresponding method on the view, though, as we don't have any reference to `TextEditor` in `DisplayBuffer`: passing a `TextEditor` instance when constructing `DisplayBuffer` seems like a bad move to me (other than being an unhealthy dependency, I think we'd have serialization issues) so I went ahead and simply deleted it in a027772 and 7cd318f.

If we want to keep it, I think the only option would be to deprecate it for a few versions and then delete it afterwards (so that package developers have some time to upgrade). On the other hand I'd :heart: to clean the `DisplayBuffer` as soon as possible... What do you think?

Other than that, I think this is ready to :ship:! @nathansobo: please, have one last :eyes:.

nathansobo:
> and a cursory search on GitHub showed that someone is actually using it

It wasn't in the public API right? How popular is the package in question? Maybe we should open an issue or even open a PR if it's a simple fix.

src/text-editor-presenter.coffee
```diff
@@ -1438,3 +1474,179 @@ class TextEditorPresenter
...
+    targetLeft = Infinity if row > @model.getLastScreenRow()
+    row = Math.min(row, @model.getLastScreenRow())
+    row = Math.max(0, row)
+
+    left = 0
+    column = 0
+
+    iterator = @model.tokenizedLineForScreenRow(row).getTokenIterator()
+    while iterator.next()
+      charWidths = @getScopedCharacterWidths(iterator.getScopes())
```
nathansobo: It's going to be so great to drop this.

src/text-editor.coffee
```diff
@@ -2854,23 +2859,27 @@ class TextEditor extends Model
 
   # Essential: Scrolls the editor to the top
   scrollToTop: ->
-    @setScrollTop(0)
```
nathansobo: Should we drop deprecated methods from the public API to avoid encouraging further use? Ideally we'd have a *Deprecated* label, but I'm not sure it's worth the time to add it to the docs system at this time.

nathansobo:
Gave this a quick scan. Ultimately if tests are green and the structure hasn't changed beyond our discussion, I'm good to merge once we drop deprecated methods from the docs.

src/text-editor.coffee
as-cii: Agreed to drop methods from the public API. :+1:

src/text-editor-presenter.coffee
as-cii: Agreed! :fireworks:

as-cii:
> It wasn't in the public API right? How popular is the package in question? Maybe we should open an issue or even open a PR if it's a simple fix.

Nope, the method wasn't in the public API. The package in question is `color-picker` but I think the fix is straightforward enough to deserve a PR. I'll do that shortly and :ship: this PR afterwards.

Thanks, @nathansobo!

nathansobo:
💥🎉🇮🇹

maxbrunsfeld:
Awesome!

maxbrunsfeld:
Hey @as-cii, apparently our internal CI goes red when deprecated methods are called by core or bundled packages. I actually didn't know this either, until now, but I guess it makes sense, because we don't want the somewhat-expensive `Grim.deprecate` calls to be happening in the next release. I [temporarily commented out](https://github.com/atom/atom/commit/8fb4e887df800438b2deeb23a2c23004b9ff358a) the `Grim.deprecate` calls that you added in `TextEditor`, just to get CI green.

When you get a chance, could you update core and bundled packages to use the new APIs? The deprecations occurred when these tests:
* [ ] find-and-replace
* [ ] symbols-view
* [ ] wrap-guide
* [ ] atom core
  * [ ] `::page{Up,Down}`
  * [ ] `::selectPage{Up,Down}`

PR #8906, opened 2015-09-23
Fix behavior of Cursor methods that are aware of word and non-word characters

### LINKED ISSUE #3650: getCurrentWordBufferRange doesn't seem to work correctly for words like change-case
I'm trying to get the current word under the cursor and the default options of `getCurrentWordBufferRange` should AFAIK enable matching words with non-space separators, but currently the regex seems to have been misconfigured so that it doesn't match word correctly.

Current code:

``` coffee
wordRegExp: ({includeNonWordCharacters}={}) ->
includeNonWordCharacters ?= true
nonWordCharacters = atom.config.get('editor.nonWordCharacters')
segments = ["^[\t ]*$"]
segments.push("[^\\s#{_.escapeRegExp(nonWordCharacters)}]+")
if includeNonWordCharacters
  segments.push("[#{_.escapeRegExp(nonWordCharacters)}]+")
new RegExp(segments.join("|"), "g")
```

Correct code:

``` coffee
wordRegExp: ({includeNonWordCharacters}={}) ->
includeNonWordCharacters ?= true
nonWordCharacters = atom.config.get('editor.nonWordCharacters')
segments = ["^[\t ]*$"]
if includeNonWordCharacters
  segments.push("[#{_.escapeRegExp(nonWordCharacters)}]+")
  segments.push("[^\\s]+")
else
  segments.push("[^\\s#{_.escapeRegExp(nonWordCharacters)}]+")
new RegExp(segments.join("|"), "g")
```

### LINKED ISSUE #6538: Confusing behavior of Cursor.getCurrentWordBufferRange()
I have a very simple file containing nothing but the text `-12`.
When I execute the `getCurrentWordBufferRange()`-method on this text with the cursor on different positions of the text I get 3 different results:
- "-" cursor position [0,0]
- "-12" cursor position [0,1]
- "12" cursor position [0, 2]

Is this really the expected behaviour of `getCurrenWordBufferRange()`? I would expect it to always return the string `"-12"` or at least not different results for the same word.

I tested it on atom 0.194.0 on Mac OS X.

Here is a simple jasmine spec with the confusing results which all pass on my machine:

```
getEditorElement = (callback) ->
  textEditor = null

  waitsForPromise ->
    atom.project.open().then (e) ->
      textEditor = e

  runs ->
    element = document.createElement('atom-text-editor')
    element.setModel(textEditor)
    callback(element)

describe 'getCurrentWordBufferRange tests', ->
  editor = null

  beforeEach ->
    getEditorElement (element) ->
      editorElement = element
      editor = editorElement.getModel()
      editor.setText('-12')

  describe 'getCurrentWordBufferRange tests', ->

    it 'current word at position [0,0] is "-"', ->
      cursor = editor.addCursorAtScreenPosition([0, 0])
      cursorText = editor.getTextInBufferRange(cursor.getCurrentWordBufferRange())
      expect(cursorText).toBe('-')

    it 'current word at position [0,1] is "-12"', ->
      cursor = editor.addCursorAtScreenPosition([0, 1])
      cursorText = editor.getTextInBufferRange(cursor.getCurrentWordBufferRange())
      expect(cursorText).toBe('-12')

    it 'current word at position [0,2] is "12"', ->
      cursor = editor.addCursorAtScreenPosition([0, 2])
      cursorText = editor.getTextInBufferRange(cursor.getCurrentWordBufferRange())
      expect(cursorText).toBe('12')
```

izuzak:
This is an attempt to close https://github.com/atom/atom/issues/3650 and close https://github.com/atom/atom/issues/6538.

Sorry for the wall of text below, I think it will be helpful in understanding the confusion and figuring out what the expected behavior is. @benogle (since you commented on https://github.com/atom/atom/issues/3650) @maxbrunsfeld @nathansobo -- I'd love to hear your advice here. Also /cc @atom/feedback for extra :eyes:.

My plan is to split the work around this problem into several steps (within this PR):
- [ ] Describe the problem and write specs to help us understand the expected behavior
- [ ] Update the API documentation to explain the behavior of methods and the supported options
- [ ] Fix the incorrect behavior for all affected methods

This text below if about the first task -- describing the problem and adding specs to confirm expected behavior.

https://github.com/atom/atom/issues/3650 and https://github.com/atom/atom/issues/6538 bring up a problem with how `Cursor.getCurrentWordBufferRange` works in some cases. For text like `atom-+electron`, here are results for different positions of the cursor and different values of the `includeNonWordCharacters` option:

| Cursor position | `includeNonWordCharacters = true` | `includeNonWordCharacters = false` |
| --- | --- | --- |
| <code>at&#124;om&#8209;+electron</code> | `atom` | `atom` |
| <code>atom&#8209;&#124;+electron</code> | `-+` | _nothing_ |
| <code>atom&#8209;+&#124;electron</code> | `-+electron` | `electron` |

The problem with the above is the inconsistency of behavior, lack of specs, and unclear documentation:
- When the cursor is between a word character and a non-word character (<code>atom&#8209;+&#124;electron</code>) and `includeNonWordCharacters = true`, then both the word characters and the non word characters are grouped into a single word (`-+electron`), but only those groups next to the cursor (notice that `atom` wasn't grouped into that).
- When the cursor is between two word characters (or two non-word characters) and `includeNonWordCharacters = true`, then only word characters or only non-word characters are grouped into a single word.
- The [documentation](https://github.com/atom/atom/blob/e1fb376a084f463c4f1e641cb89803dc3095f9b2/src/cursor.coffee#L616-L617) for the `includeNonWordCharacters` option is unclear -- it says: "A Boolean indicating whether to include non-word characters in the regex." But it's not clear if that means that the regex then groups both word and non-word characters into a single word, or that the regex groups non-word characters into words, but not together with word characters. The name of the option implies the former, but the behavior _mostly_ implies the latter.
- There are no specs around these methods which would indicate expected behavior, as far as I can tell.

My _guess_ is that the intended meaning of the `includeNonWordCharacters` option is "should non-word characters be grouped together into a word (but never together with word characters)". In other words, the `-+electron` result from table above should not happen since it contains both word and non-word characters, and would be considered a bug. Instead, the result should have been `electron` since that's the word next to the cursor (we favor words consisting of word characters, not words consisting of non-word characters).

Based on this guess -- I added specs in this PR to cover the cases shown in the table, and the only spec which fails is the one for `-+electron`:

```
TextEditor
  cursor
    .getCurrentWordBufferRange()
      when cursor is between a word character and non-word character
        when includeNonWordCharacters is true
          it returns the word next to the cursor, defined with word characters
            Expected { start : { row : 0, column : 4 }, end : { row : 0, column : 14 } } to equal [ [ 0, 6 ], [ 0, 14 ] ].
```

The reason why this problem happens lies in how the `Cursor.getCurrentWordBufferRange` methods works. That method scans to the left and to the right [to find the edge of words](https://github.com/atom/atom/blob/e1fb376a084f463c4f1e641cb89803dc3095f9b2/src/cursor.coffee#L551). However, it doesn't check if the cursor is between a word and a non-word character. That's relevant because if `includeNonWordCharacters` is `true` -- the method will group word and non-word characters together by joining the left and right part together. Some other methods in the `Cursor` class have the same problem (for example, [`getBeginningOfCurrentWordBufferPosition`](https://github.com/atom/atom/blob/e1fb376a084f463c4f1e641cb89803dc3095f9b2/src/cursor.coffee#L476) and [`getEndOfCurrentWordBufferPosition`](https://github.com/atom/atom/blob/e1fb376a084f463c4f1e641cb89803dc3095f9b2/src/cursor.coffee#L506) which `getCurrentWordBufferRange` uses for scanning) -- they don't check if the cursor is between a word and non-word character.

Interestingly, this check is made as a part of one other public method -- in [Selection.selectWord](https://github.com/atom/atom/blob/abf6d40cc4e3462a38fe2ce041d222d369e44147/src/selection.coffee#L315). Notice how this method uses `Cursor.isBetweenWordAndNonWord()` first, and modifies behavior based on that. I believe that logic should be in `Cursor` methods. In other words, `Cursor` methods should check if the cursor is between a word and non-word character and handle the situation when `includeNonWordCharacters` is `true`.

I'm not 100% sure about this, so definitely want to hear if this sounds good to everyone.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -211,6 +211,50 @@ describe "TextEditor", ->
         editor.getLastCursor().destroy()
         expect(editor.getLastCursor().getBufferPosition()).toEqual([0, 0])
 
+    describe ".getCurrentWordBufferRange()", ->
+      cursor = null
+
+      beforeEach ->
+        editor.setText('atom-+electron')
+        atom.config.set 'editor.nonWordCharacters', '-+'
+        cursor = editor.getLastCursor()
+
+      describe "when cursor is between two word characters", ->
+        beforeEach ->
+          editor.setCursorBufferPosition([0, 2]) # at|om-+electron
+
+        describe "when includeNonWordCharacters is false", ->
+          it "returns the word under the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": false})).toEqual [[0, 0], [0, 4]] # atom
+
+        describe "when includeNonWordCharacters is true", ->
+          it "returns the word under the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": true})).toEqual [[0, 0], [0, 4]] # atom
+
+      describe "when cursor is between two non-word characters", ->
+        beforeEach ->
+          editor.setCursorBufferPosition([0, 5]) # atom-|+electron
+
+        describe "when includeNonWordCharacters is false", ->
+          it "returns a zero-width range", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": false})).toEqual [[0, 5], [0, 5]] # <nothing>
+
+        describe "when includeNonWordCharacters is true", ->
+          it "returns the word under the cursor, defined with non-word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": true})).toEqual [[0, 4], [0, 6]] # -+
+
+      describe "when cursor is between a word character and a non-word character", ->
+        beforeEach ->
+          editor.setCursorBufferPosition([0, 6]) # atom-+|electron
+
+        describe "when includeNonWordCharacters is false", ->
+          it "returns the word next to the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": false})).toEqual [[0, 6], [0, 14]] # electron
+
+        describe "when includeNonWordCharacters is true", ->
+          it "returns the word next to the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": true})).toEqual [[0, 6], [0, 14]] # electron
+
     describe ".getCursors()", ->
       it "creates a new cursor at (0, 0) if the last cursor has been destroyed", ->
         editor.getLastCursor().destroy()


spec/text-editor-spec.coffee
```diff
@@ -211,6 +211,50 @@ describe "TextEditor", ->
...
+          editor.setCursorBufferPosition([0, 6]) # atom-+|electron
+
+        describe "when includeNonWordCharacters is false", ->
+          it "returns the word next to the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": false})).toEqual [[0, 6], [0, 14]] # electron
+
+        describe "when includeNonWordCharacters is true", ->
+          it "returns the word next to the cursor, defined with word characters", ->
+            expect(cursor.getCurrentWordBufferRange({"includeNonWordCharacters": true})).toEqual [[0, 6], [0, 14]] # electron
+
```
as-cii: Thanks for tackling this, @izuzak! :clap: 

Do you think we could express the specs in terms of `editor.getWordUnderCursor`? Seems like we should be able to drop comments with that, and to let specs tell a better story. :thought_balloon:

izuzak: > Do you think we could express the specs in terms of `editor.getWordUnderCursor`? Seems like we should be able to drop comments with that, and to let specs tell a better story. :thought_balloon:

That's what my first version of these specs looked like, and then I decided to switch them to use `cursor.getCurrentWordBufferRange` directly. My plan was actually to add spec for both methods because I believe that every public API method should have it's own specs, even in cases when one public API method is mostly a wrapper for another (as is the case here -- both are public, and one is mostly a wrapper for the other). I think that approach will help us feel safer about not breaking any behavior and breaking our promise to package developers. I can definitely imagine situations where changes are made to both methods and we find out that tests for `editor.getWordUnderCursor` still pass, but `cursor.getCurrentWordBufferRange` is not covered at all anymore or an edge case is introduced which the tests for `editor.getWordUnderCursor` don't cover.  

Still, this issue and pull request have been a big learning experience for me :school_satchel:, and I would love to hear what everyone thinks about my reasoning above. /cc @atom/feedback Thanks for bringing this up, @as-cii. :blue_heart:

lee-dohm: > My plan was actually to add specs for both methods because I believe that every public API method should have its own specs, even in cases where one public API method is mostly a wrapper for another

I agree. The specs should document the entire public API, at a minimum.

jeancroy:
With a word like `snake_case_test` isn't the proper behavior of move-to-next-word this:

````
|snake_case_test
snake_|case_test
snake_case_|test
````

IE the setting `includeNonWordCharacters` means don't stop a non words when scaning for next word start.

**Comment by izuzak** [comment]
Thanks for the feedback, @jeancroy

> With a word like snake_case_test isn't the proper behavior of move-to-next-word this:
> 
> ```
|snake_case_test
snake_|case_test
snake_case_|test
> ```

Can you provide the exact name of the command which you're referring to? I don't see a `move-to-next-word` command. Are you perhaps referring to `move-to-next-word-boundary` or `move-to-beginning-of-next-word`? If you're referring to one of those, does the current behavior of Atom match what you described above?

> IE the setting includeNonWordCharacters means don't stop a non words when scaning for next word start. If that is a good behavior then I beleive it's more consistent with previous way of doing things than this PR

I don't think it means that. See the table I provided above: https://github.com/atom/atom/pull/8906#issue-107940584. Currently, there's no combination which would give the behavior you described above. For example, for this position of the cursor: `at|om‑+electron` and `includeNonWordCharacters=true`, the result is `atom`, not `atom-+`. If I'm wrong or misunderstanding, can you point out which combination does give the behavior you described?

Also, it think it shouldn't mean that -- it think that would make the function less general and also tied to a specific use-case. That use-case could be achieved by a more general approach where the regex is used with includeNonWordCharacters=true to skip blocks of only word or only non-word characters. Since you can model a string of text as a list of such blocks (word, non-word, word, non-word, ...), you can use that to implement skipping to the next word by skipping blocks of word and non-word characters. What do you think?

> At contrario I don't see an immediate use case where I would want to select a word made only of non word character.

That's how the editor behaves currently -- try typing "####$$$$$" and double-click within that string. I see the same behavior in SublimeText.

**Comment by izuzak** [comment]
Would love some feedback on what this parameter should really mean from @atom/core. :pray:

**Comment by nathansobo** [comment]
@izuzak I agree that the current behavior seems wrong when we're at a word boundary. I support this effort to get test coverage and clarify things, and hope you'll proceed.

**Comment by izuzak** [comment]
> @izuzak I agree that the current behavior seems wrong when we're at a word boundary. I support this effort to get test coverage and clarify things, and hope you'll proceed.

Thanks, @nathansobo. At this point - I think I need feedback to proceed. I've added basic tests in this PR in order to clarify/demonstrate the problem, but need confirmation that the added tests really do describe expected behavior. In other words, I'm not really sure what the expected behavior is, so I picked one based on my intuition and wrote tests for that.

If you (or someone else) can confirm that I made the right choice, I think I can continue in that direction. :arrow_left: :blue_car:

**Comment by nathansobo** [comment]
Your specs seem reasonable to me. I can't predict what the implications of changing it will be exactly, but it seems correct from an intuitive perspective. I think you should proceed to make your tests pass.

**Comment by jeancroy** [comment]
I'm sorry for the confusion. The `-+electron` entry effectively looks wrongs.
Your proposal looks like a minimum change solution to that problem.

Interesstingly there's at least one gard against `+-electron` because between a word and a non word char `includeNonWordCharacters` is set to false. [here](https://github.com/atom/atom/blob/7d743827820969871e8724e6f353887a3ddfc105/src/selection.coffee#L314-L315)

-----------------

While looking at linked issue it looks like your first intuition was something like that

With selection `### foo-ba|r-baz ###` :
- `bar` (includeNonWordCharacters = false)
- `foo-bar-baz` (includeNonWordCharacters = true)

That is, it sometime make sens to mix word and non word. And basically that was the issue I had,  but looking at the table it might be the scope of another method.

**Comment by nathansobo** [comment]
@izuzak Yeah, upon thinking about it more, it might be more intuitive to always expand over word and non-word characters and only stop at whitespace when `includeNonWordCharacters` is true. To me that matches the meaning of "include" the best... it's something additional, not something different.

**Comment by DanEEStar** [comment]
I tripped over this problem again when trying to do some work on vim-mode.
I also think the `+-electron` entry looks wrong.

I think as you proposed this problem should be solved at the `cursor` level. Because that is the API that gets used in the vim-mode code.

In my opinion it would also make sense to get the "whole" word (with non word characters) when `includeNonWordCharacters` is set to `true`

I hope you can proceed soon with some changes.

**Comment by benmccormick** [comment]
> My guess is that the intended meaning of the includeNonWordCharacters option is "should non-word characters be grouped together into a word (but never together with word characters)". In other words, the -+electron result from table above should not happen since it contains both word and non-word characters, and would be considered a bug. Instead, the result should have been electron since that's the word next to the cursor (we favor words consisting of word characters, not words consisting of non-word characters).

@izuzak I think the behavior you described above makes sense, and would allow us to implement correct behavior around word objects in vim-mode as well.  (see https://github.com/atom/vim-mode/issues/868#issuecomment-152072850 for the problems that this would solve for vim-mode)

Its also easy to take that behavior you're describing and add a slight wrapper to it to get the current behavior, whereas to reproduce the behavior you're describing right now you have to rebuild a method from primitives.

**Comment by mamenama** [comment]
I think the issue is with the way that `wordRegExp` is built:
```coffeescript
  wordRegExp: ({includeNonWordCharacters}={}) ->
    includeNonWordCharacters ?= true
    nonWordCharacters = atom.config.get('editor.nonWordCharacters', scope: @getScopeDescriptor())
    segments = ["^[\t ]*$"]
    segments.push("[^\\s#{_.escapeRegExp(nonWordCharacters)}]+")
    if includeNonWordCharacters
      segments.push("[#{_.escapeRegExp(nonWordCharacters)}]+")
    new RegExp(segments.join("|"), "g")
```
This creates a RegExp by calling anything that is EITHER a stretch of WORDS or a stretch of NONWORDS but not one that is a mix of both. Probably should look more like:

```coffeescript
  wordRegExp: ({includeNonWordCharacters}={}) ->
    includeNonWordCharacters ?= true
    nonWordCharacters = atom.config.get('editor.nonWordCharacters', scope: @getScopeDescriptor())
    segments = ["^[\t ]*$"]
    if includeNonWordCharacters
      segments.push("[^\\s]+")
    else
      segments.push("[^\\s#{_.escapeRegExp(nonWordCharacters)}]+")

    new RegExp(segments.join("|"), "g")
```

izuzak:
Thanks for the feedback, everyone. After reading all the comments and the related vim-mode issue, I'm not sure I have a clear idea about which direction to go in, so I've asked a few other folks from the Atom core team for help. Just wanted to leave a note since there's been more than 10 days of :radio: silence. :v:

benogle:
@izuzak what's the status of this? Did you get the help you needed?

izuzak:
> @izuzak what's the status of this?

@benogle The status is that there's still a problem that needs to be fixed here. I tried to describe the problem in the body of the pull request and tried to write specs which demonstrate expected behavior. However, I don't think we have agreement on what the expected behavior should be.

> Did you get the help you needed?

No. I think someone else (someone from the core team) should pick this up because 1) I feel it's not a trivial problem to solve, and 2) I no longer have time to work on this. We can also close this pull request and open an issue to track the underlying problem using the description and discussion above.

benogle:
Thanks for the info @izuzak

damieng:
No response.

PR #8919, opened 2015-09-24
Add top/bottom split options when opening uri in a workspace

bencolon:
Exactly the same behaviour as left/right for top/bottom.

FYI, I need this to update the [find-and-replace](https://github.com/atom/find-and-replace) package with an option to open Find results on the bottom. For now, the package allows to open Find results only on the right.

Now : 

capture d ecran 2015-09-24 a 10 27 07

With the bottom option : 

capture d ecran 2015-09-24 a 10 27 36

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -281,6 +281,90 @@ describe "Workspace", ->
               expect(workspace.paneContainer.root.children[0]).toBe pane1
               expect(workspace.paneContainer.root.children[1]).toBe pane4
 
+      describe "when the 'split' option is 'up'", ->
+        it "opens the editor in the topmost pane of the current pane axis", ->
+          pane1 = workspace.getActivePane()
+          pane2 = pane1.splitDown()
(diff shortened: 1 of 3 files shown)


benogle:
cc @nathansobo @maxbrunsfeld

nathansobo:
I'd like to see `top` and `bottom` renamed to `up` and `down` to blend in with the existing `splitUp/Down/Left/Right` APIs.

The existing API seems a bit strange in that we'll create a new pane when passing `split: 'right'` but won't create a new pane when splitting left. I see @bencolon blended in with that approach here, but @bencolon do you see any reason not to just create a new pane when splitting upward as well? I'm unclear what would break if we changed the current behavior when passing `split: 'left'`.

nathansobo:
Thanks @bencolon for adding this. I still wish our behavior for `up` and `left` were consistent (we always split if only a single pane existed in the desired direction). If you or anyone else can investigate the impact of changing that, that would be great.

PR #8923, opened 2015-09-24
Fix runas requirement

g2p:
runas 2 doesn't build with a recent node.js

Closes #8922
Closes #8508

PR #8779, opened 2015-09-14
Update to Electron v0.33.x

zcbenz:
Node modules to update:
- [x] runas
- [x] pathwatcher
- [x] nslog
- [x] oniguruma
- [x] git-utils
- [x] spellchecker
- [x] scrollbar-style
- [x] keyboard-layout
- [x] atom-keymap
- [x] first-mate
- [x] scandal
- [x] text-buffer
- [x] apm
- [x] highlights
- [x] ctags

Atom packages to update:
- [x] bracket-matcher
- [x] link
- [x] markdown-preview
- [x] symbols-view
- [x] tree-view
- [x] spell-check

Fixes #8508
Fixes #7061
Fixes #3593

PR #8926, opened 2015-09-24
Add Node 4 to Travis Build Matrix

as-cii:
Closes #8923.

This is a continuation of #8923 that will allow us to test whether node 4 works correctly with our dependencies.

This will increase a bit the Travis backlog but this is only temporary, as we should be able to keep only the latest node version after we merge #8779.

/cc: @atom/feedback @mnquintana

### DIFF
--- a/build/package.json
+++ b/build/package.json
@@ -33,7 +33,7 @@
     "rcedit": "~0.3.0",
     "request": "~2.27.0",
     "rimraf": "~2.2.2",
-    "runas": "^2",
+    "runas": "^3",
     "tello": "1.0.5",
     "temp": "~0.8.1",
     "underscore-plus": "1.x",
--- a/build/package.json
+++ b/build/package.json
@@ -33,7 +33,7 @@
     "rcedit": "~0.3.0",
     "request": "~2.27.0",
     "rimraf": "~2.2.2",
-    "runas": "^3",
+    "runas": "^3.1",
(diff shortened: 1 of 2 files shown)


build/package.json
```diff
@@ -33,7 +33,7 @@
     "rcedit": "~0.3.0",
     "request": "~2.27.0",
     "rimraf": "~2.2.2",
-    "runas": "^2",
+    "runas": "^3",
```
50Wliu: ^3.1 instead?  I know it doesn't make a difference, but the main package.json currently uses ^3.1 right now.

50Wliu:
The build matrix doesn't seem to be setup correctly: right now there's a build with matrix `NODE_VERSION=0.12 NODE_VERSION=4` :frowning: (which is the equivalent of NODE_VERSION=4).

as-cii:
Thanks for the heads-up, @50Wliu! It was kinda late yesterday :sleeping: and I put this on pause.

So, I ended up configuring the new node only to build on Linux: the reason is that, otherwise, we'd have 6 jobs per push (vs. 4) and that seems a good tradeoff to ensure that everything builds correctly with both node versions.

What do you think? /cc: @mnquintana

mnquintana:
Looks good to me :rocket:

as-cii:
Thanks!

PR #8929, opened 2015-09-24
:bug: fix atom.isValidDimensions under Mac (fixes #7555, fixes #8909)

### LINKED ISSUE #7555: Window position not remembered after Reload
Atom 1.0, Mac OS 10.10.

Steps to reproduce:
- Launch Atom
- Move the Atom window to a secondary monitor and resize the Atom window
- Go to View > Reload
- (BUG) When Atom is reloaded it's moved back to it's original position and size

Expected behaviour:
- Atom should remember where the user can moved the window and what size it was between reloads

This is not an issue on Windows (where the window reloads in the correct location). Not sure if this is Atom-specific or an issue with Electron.

### LINKED ISSUE #8909: Cannot change window height after entering target display mode
Seeing this on OS X 10.10.5.

1) Launch Atom on a laptop (screen dimensions 2880X1800).
2) Resize Atom window so it occupies most/all of the screen height.
3) Connect an iMac to laptop and enter [target display mode](https://support.apple.com/en-us/HT204592). External display resolution 2560x1440
4) Once the new display kicks in, you cannot change the height of the Atom window.  (Width resizing works just fine.)

I wonder if this is related to #7555. Or if perhaps moving to a shorter display freaks Atom out.

darwin:
Additionally I have also:
- added logging when detected invalid dimensions
- changed logic for determining width and height after recovery from invalid dimensions

Original implementation was not symmetric. And I believe Math.min was a mistake. Original intent probably was to use Math.max instead.
It didn't guard against case of generating invalid width if screen.getPrimaryDisplay().workAreaSize returned 0 or some negative value.
And it was clamping width for displays wider than 1024 which makes little sense.

New implementation makes sure width is at least 1024 and height is at least 512 after reset.

Fixes #7555
Fixes #8909

### DIFF
--- a/spec/atom-spec.coffee
+++ b/spec/atom-spec.coffee
@@ -11,6 +11,8 @@ describe "the `atom` global", ->
       it 'sets the position of the window, and can retrieve the position just set', ->
         atom.setPosition(22, 45)
         expect(atom.getPosition()).toEqual x: 22, y: 45
+        atom.setPosition(-22, -45) # negative coordinates are fine on external displays under OS X
+        expect(atom.getPosition()).toEqual x: -22, y: -45
 
     describe '::getSize and ::setSize', ->
       originalSize = null
--- a/spec/atom-spec.coffee
+++ b/spec/atom-spec.coffee
@@ -11,8 +11,8 @@ describe "the `atom` global", ->
       it 'sets the position of the window, and can retrieve the position just set', ->
         atom.setPosition(22, 45)
         expect(atom.getPosition()).toEqual x: 22, y: 45
-        atom.setPosition(-22, -45) # negative coordinates are fine on external displays under OS X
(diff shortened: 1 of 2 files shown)


spec/atom-spec.coffee
```diff
@@ -11,6 +11,8 @@ describe "the `atom` global", ->
       it 'sets the position of the window, and can retrieve the position just set', ->
         atom.setPosition(22, 45)
         expect(atom.getPosition()).toEqual x: 22, y: 45
+        # atom.setPosition(-22, -45) # negative coordinates are fine on external displays under OS X
```
nathansobo: Do these assertions pass on OS X? If so, you could run them conditionally on `process.platform is 'darwin'`.

damieng:
Out of date PR, addresses two bugs that have been closed by alternate means.

PR #8945, opened 2015-09-25
Fix build failures for Electron 32 upgrade

as-cii:
This introduces a couple of commits that fix [those spec failures](https://travis-ci.org/atom/atom/builds/81909632) we are observing in #8779 :green_heart:.

These are pretty straightforward, but it'd be great to have a second pair of :eyes:.

Thanks! :bow: 

/cc: @atom/feedback @zcbenz


zcbenz:
Awesome, I'm merging this and we can continue the review in #8779.

PR #8958, opened 2015-09-26
Remove hardcoded Makefile tab behavior

winstliu:
With the merge of atom/language-make#23, this hardcoded behavior is no longer needed and is set in the Make package itself.

Refs atom/language-make#21.

/cc @benogle


thomasjo:
LGTM! :rocket:

PR #8966, opened 2015-09-27
upgrade packages

fscherwi:
| package | current | upgraded | status | age |
| :-: | :-: | :-: | :-: | :-: |
| async | 0.2.6 |  | :arrow_left: | 31 months :heavy_exclamation_mark: |
| coffee-script | 1.8.0 |  | :arrow_left: | 13 months |
| color | ^0.7.3 | ^10.0.1 | :arrow_up: | 12 months |
| fstream | 0.1.24 | 1.0.8 | :arrow_up: | 26 months :heavy_exclamation_mark: |
| semver | ^4.3.3 | ^5.0.3 | :arrow_up: | 4 months |
| service-hub | ^0.6.2 | ^0.7.0 | :arrow_up: | 3 months |
| stacktrace-parser | 0.1.1 | 1.1.3 | :arrow_up: | 20 months :heavy_exclamation_mark: |
| temp | 0.8.1 | 0.8.3 | :arrow_up: | 14 months |
| typescript-simple | 1.0.0 | 2.0.0 | :arrow_up: | 7 months |

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "stacktrace-parser": "0.1.3",
     "temp": "0.8.3",
     "text-buffer": "7.1.2",
-    "typescript-simple": "1.0.0",
+    "typescript-simple": "2.0.0",
     "underscore-plus": "^1.6.6",
     "yargs": "^3.23.0"
   },
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "scrollbar-style": "^3.2",
     "season": "^5.3",
     "semver": "^5.0.3",
-    "service-hub": "^0.6.2",
+    "service-hub": "^0.7.0",
     "source-map-support": "^0.3.2",
     "stacktrace-parser": "0.1.3",
     "temp": "0.8.3",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "event-kit": "^1.3.0",
     "first-mate": "^5.0.0",
     "fs-plus": "^2.8.0",
-    "fstream": "0.1.24",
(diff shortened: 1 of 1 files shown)


50Wliu:
What's your motivation behind these changes?  Is there any reason you want these packages to be updated other than the fact that we're a few versions behind?

fscherwi:
Because they have all small changes, e.g. beautification, so that's not a big problem!
But a few of the packages are older than 1 ½ years!
And newer packages are simply looking better!

fscherwi:
Are there any updates about this PR?

thomasjo:
I'm a little bit concerned with these updates; some of these are major version bumps? Are we guaranteed there are zero breaking changes? Why are these packages updated? Merely to use the latest version, fix bugs or security holes? If the latter, some details would be great.

lee-dohm:
I seem to remember another PR like this that was closed by @thedaniel.

thedaniel:
> But a few of the packages are older than 2 years!

Also in 2 years, those packages went from 0.x to 1.x with who knows how many API changes. Upgrading dependencies is fine, but we have to do it in a more careful way, and for a specific reason (bug fixes, security, etc.)

It's OK for a dependency to be 10 years out of date if the system using it is stable :) I noticed the tests failed on some of the commits in this PR as well.

I don't mean to discourage you from contributing, and from updating dependencies, but I'm going to close this particular PR since it combines too many big updates. Feel free to open more focused PRs with some more explanation about the dependencies, and also note that if you're changing a package's major version its API might have changed..

PR #8968, opened 2015-09-28
Enable custom package testing environments

nathansobo:
This PR aims to improve the experience of running tests in Atom in a few ways.
- Package authors will be able to specify an `atomTestRunner` field in their `package.json` containing a require path. If the field is present, Atom will resolve the path relative to the package's root, expecting the required module to export a function which it will invoke with an object containing the following parameters.
  - `testPaths` An array of paths to tests to run. Could be paths to files or directories.
  - `buildAtomEnvironment` A function that can be called to construct an instance of the `atom` global (it won't automatically be assigned to `global.atom`, that is left to the test runner to decide.
  - `logFile` An optional path to a log file to which test output should be logged.
  - `headless` A boolean indicating whether or not the tests are running headless.
  - `legacyTestRunner` This function can be invoked to run the legacy Jasmine runner, giving packages a chance to transition to a new test runner while maintaining a subset of their tests in the old environment.
- The test runner function will be expected to return a promise that resolves with an exit code. When tests are being run in headless mode Atom will exit with this code when the promise resolves. In interactive mode the promise will be ignored.
- Running tests via the `atom` CLI has been streamlined. Simply run `atom --test path1 path2` to invoke the test runner on the listed paths, expanded relative to the working directory. The `--spec-directory` option will be deprecated.
- The `buildAtomEnvironment` function will construct a clean Atom environment whose state can be reset via a call to `reset` in between test runs.

Refs #8917
/cc @bolinfest @maxbrunsfeld
### Additional Thoughts

Atom's current testing environment has been around since the very first day of Atom development, and is quite a mess. We assign all kinds of globals and mock the clock in evil and inconsistent ways that can cause quite a few headaches when writing tests for new packages.

This PR will give us a good way to make a fresh start. Ideally, a default test runner isn't something we should be supporting in Atom core. Every package should specify its preferred runner, which it can pull in as a development dependency and which Atom can interact with via an evolving interface. It would be good to have a recommended test framework. As it stands I'd probably want to switch bundled packages over to Mocha/Chai, unless people have other suggestions. Someday in the distant future we can finally delete the tangled mess that is our current default testing environment, but we'll keep it around for now.
### Tasks
- [x] Basic clean up around test window initialization
- [x] Support specifying multiple test paths as arguments from the CLI.
- [x] Support `atomTestRunner` field containing a path.
- [x] Support multiple test paths in the default jasmine test runner.
- [x] Clean up Atom environment construction so we can instantiate the Atom environment without assigning to globals.
- [x] Pass `buildAtomEnvironment` to test runners and use it to assign the `atom` global in the default test runner, rather than in the test window initialization script.
- [x] Rename `Atom` to `AtomEnvironment`
- [x] Add a `reset` method to Atom environment
- [x] Pass `legacyTestRunner` parameter.
- [x] Fix docs generation.
- [x] Drop the `--spec-directory` CLI option
  - [ ] Stop using `--spec-directory` in `apm test` (https://github.com/atom/apm/pull/445)
  - [x] Stop using `specDirectory` parameter when running tests interactively
- [x] Drop the `Run All Tests` command from Atom since it's broken.
- [x] Shim `process.stdout.write` and `process.stderr.write` in the test window
- [x] Support a timeout flag
- [x] Expect the test runner to return a promise
- [x] Update packages to not use deprecated methods
  - [x] https://github.com/atom/welcome/pull/39
  - [x] https://github.com/atom/markdown-preview/pull/316
  - [x] https://github.com/atom/language-php/pull/105
  - [x] https://github.com/atom/autocomplete-plus/pull/592
  - [x] https://github.com/atom/snippets/pull/176
  - [x] https://github.com/atom/autosave/pull/49
  - [x] https://github.com/atom/tabs/pull/214
  - [x] https://github.com/atom/language-javascript/pull/239
  - [x] https://github.com/atom/find-and-replace/pull/551
  - [x] https://github.com/atom/language-c/pull/93
  - [x] https://github.com/atom/find-and-replace/pull/552
  - [x] https://github.com/atom/fuzzy-finder/pull/144
- [x] Add user-facing documentation for the new workflow and options 
- [x] Move remaining uses of APIs that interact with main process such as `ipc` to methods on `ApplicationDelegate`
- [x] Assign the `defaultTarget` of the keymaps in the `AtomEnvironment` constructor.
- [x] Parameterize the `configDirPath` and don't attempt any disk I/O in `Config` or `AtomEnvironment` if it is undefined. This will allow us to remove several mocks from the spec helper that prevent reading and writing state.
- [x] Move `config.load()` into the `AtomEnvironment` constructor, which will be a no-op in specs where the config dir is undefined. This will allow us to move `loadBaseStyleSheets` back into the constructor as well so they are available in the environment's default state.
- [x] Move subscription to `core.autoHideMenuBar` into `AtomEnvironment` constructor.
- [x] Get build green.
- [ ] Document environment parameters (`applicationDelegate`, `window`, `document`)


izuzak:
Just a note -- I added this to the list of tasks above:

> Add user-facing documentation for the new workflow and options

I think that will help package authors understand and switch to the new workflow. It can definitely be done as a separate PR, but wanted to add it here so we don't forget. :v:

thomasjo:
I'm struggling to find the words that accurately express how insanely excited I am about this, so I must resort to expressing myself with this beauty

giphy

nathansobo:
> Add user-facing documentation for the new workflow and options

Thanks @izuzak. I actually realized I forgot to list that as I was falling asleep last night.

bolinfest:
What will the value of `testPaths` be when `apm test` is run by default: will it be empty? Will the user always be required to specify paths?

An easy solution is probably to include the absolute path to the Atom package being tested. That way, if the test runner wants to have a default place where tests (or fixtures) live, it can find that by resolving the expected subpath against the location of the Atom package. So if the property were named `pathToAtomPackageUnderTest`:

```js
import path from 'path';

// Note this assumes testPaths can be a mix of files and directories.
function runTests({pathToAtomPackageUnderTest, testPaths}) {
  if (testPaths.length === 0) {
    testPaths = [path.join(pathToAtomPackageUnderTest, 'spec')];
  }

  // Proceed...
}
```

The nice thing about this is that running tests for the entire package is still as simple as invoking `apm test` while only one property needs to be added to `package.json` (for `"atomTestRunner"`).

Also, what is the return value of the function exported by `"atomTestRunner"` supposed to be? I'd suggest a `Promise` that resolves to some sort of object that, at a minimum, indicates whether the test passed or failed so we know what exit code to return from `apm test`.

bolinfest:
Thinking about this more, `pathToAtomPackageUnderTest` is really important because if there is other metadata that the test runner expects to be baked into the package's `package.json` file, or dotfiles in the same directory as `package.json`, it should be able to locate those things easily.

nathansobo:
> Will the user always be required to specify paths?

Currently `apm test` already provides its own default (`package-root/spec`), but there really shouldn't be a reason to run tests via `apm` anymore at all with this more convenient CLI. I considered adding a default path, but I'm wondering if it's really that hard to supply the directory... This would be the command for running all tests on a package from its directory.

`atom -t spec`

I'm not convinced a default path is worth the added complexity. The lines of code aren't that big of a deal, but it's just one more magical thing to remember about the command rather than the simple story that the command does what you tell it to do.

> Also, what is the return value of the function exported by "atomTestRunner"

Hadn't thought about it yet, but a promise resolving to an error code makes sense. I'm going to provide access to `process.stdout.write` and `process.stderr.write` in case `headless` runners want to write to do terminal output.

bolinfest:
@nathansobo I'm not saying the user should have to supply a default path: I'm saying the test runner should be given the option to infer one. (Again, this makes it possible to eventually support a feature in Atom that lets you right click on an Atom package and run all of its tests. Without this, you would always have to prompt the user to ask where the tests are.)

But even if you don't agree with that, I think it makes sense to pass the location of the Atom package, or at least its `package.json` file. There's no reason to make the test runner try to figure that out.

bolinfest:
Also, as I'm currently dealing with tests that fail to terminate, it would be nice if `atom --test` accepted an optional `timeout` argument. It seems like most Linux distros have a nice built-in `timeout` command, but BSD/OS X do not: http://unix.stackexchange.com/questions/43340/how-to-introduce-timeout-for-shell-scripting/43346.

EasyHard:
mocha would be great! IMHO, features like `-g` (only run tests matching given regex pattern) and `describe.only` could help a lot to reduce time wasting on running irrelevant specs.

nathansobo:
Okay, the globals are cleaned up. Still have to move a bunch of logic out of the spec helper and into the atom environment's reset method, which could take some time. The other tasks are easy.

nathansobo:
We're in a tricky situation with the deprecations. We can't update any packages to the replacements for deprecated APIs until this is released, but until then all our packages are failing. @maxbrunsfeld, should we deprecate in a second step once this goes out?

nathansobo:
@as-cii What do you think about making your pull requests backwards-compatible so we can merge them now? Maybe you could make things conditional on the presence of `atom.workspace.buildTextEditor`? Then we could merge this as an actual green build which would make me feel much better.

src/browser/atom-application.coffee
```diff
@@ -483,25 +487,64 @@ class AtomApplication
...
+  runTests: ({headless, resourcePath, executedFrom, pathsToOpen, logFile, safeMode, timeout}) ->
     if resourcePath isnt @resourcePath and not fs.existsSync(resourcePath)
       resourcePath = @resourcePath
 
+    timeout = Number.parseFloat(timeout)
+    unless Number.isNaN(timeout)
+      timeoutHandler = ->
+        console.log "The test suite has timed out because it has been running for more than #{timeout} minutes."
+        process.exit(124) # Use the same exit code as the UNIX timeout util.
+      setTimeout(timeoutHandler, timeout * 60 * 1000)
```
nathansobo: What do you think about expressing timeout in seconds instead of minutes? Seems a bit more standard.

nathansobo:
Just realized we're not returning a promise. As soon as we land that and get a green build, we're ready to merge this and start getting feedback from @bolinfest or others trying it out.

bolinfest:
@nathansobo @as-cii @maxbrunsfeld This is amazing: I'm so psyched for this! Thank you!!!

nathansobo:
We now inject `document` and `window` globals into `AtomEnvironment` on construction, so that building an `AtomEnvironment` doesn't have global side effects. We also pass an `applicationDelegate` into the environment to perform all interaction with the application...

@as-cii I've added a few more tasks to the list if you want to hammer on this some more. I really think there's not much left to do before we're ready for testing. I'm starting to think we should merge this *after* cutting the next beta so it has maximal time to bake.

bolinfest:
@nathansobo When is the next beta supposed to be cut?

nathansobo:
@bolinfest: @maxbrunsfeld is actively working on it now... there's no remaining obstacles so should happen very soon, potentially today.

nathansobo:
@as-cii Came across a failure in settings view related to an undefined `configDirPath` in specs. Wondering if we should just go ahead and always assign it and introduce another parameter, `enablePersistence`, that enables / disables file system I/O for `Config` and `AtomEnvironment.prototype.load/saveSync`.

nathansobo:
Okay, everything is green but core specs but I can't see any output from CI or reproduce locally. Almost there though.

thomasjo:
> Okay, everything is green but core specs but I can't see any output from CI or reproduce locally. Almost there though.

OMG, it's almost upon us!
This has got to be the most impressive PR I've ever seen. 224 commits and still going :eyes:

nathansobo:
Was building green before a trivial merge conflict, so I'm merging into master.

@bolinfest now that this is out, it would be great to get feedback on the interface for building test runners so we can refine this if needed. Thanks for your patience... as you can see, we had some serious cleanup to do. Looking forward to seeing some diversity of testing environments come out of this.

maxbrunsfeld:
:sparkles: :chart_with_upwards_trend:

nathansobo:
Classic. Forgot to save the merge conflict resolution. Fixed on master.

bolinfest:
:sparkles: Super excited for this! Yes: I will definitely start writing code against this. I should have time tomorrow -- can't wait!

jamesgpearce:
@bolinfest I also wonder if we can do something better with our benchmarker now too.

abe33:
@as-cii @nathansobo: What do you recommend as replacement in tests that previously spied on `atom.confirm`? Should we just spy on the app delegate?

bolinfest:
I just built Atom from master (6495f1a) and when I startup, I get:

> Uncaught TypeError: deprecationSite.getFileName

when I run:

    require('atom').TextEditor

It seems like the error is a bug in the Grim, but I don't know why `require('atom').TextEditor` is triggering Grim in the first place. Is this no longer an appropriate way to get the `TextEditor` class?

bolinfest:
Ah, this appears to be the source of my woes:

https://github.com/atom/atom/commit/4d2d4c3cc47693062dfc93475d397268b47cbdc9

I'll audit how/why we're using it.

bolinfest:
This one is a real problem: https://github.com/facebook/nuclide/blob/master/pkg/nuclide/ui/atom-text-editor/lib/ReadOnlyTextEditor.js

bolinfest:
We also have four instances of testing via `instanceof TextEditor`. The recommendation to test for a method such as `getText()` seems a little weak. Methods such as `getText()` are extremely generic: they could conceivably exist on objects that are not instances of `TextEditor`.

It seems like we're generally doing this in a callback to `onDidChangeActivePaneItem()`.

nathansobo:
Run apm install to get a new Grim. As for the deprecated export, we don't want people calling that constructor because we now inject the editors dependencies. What about some other scheme for identifying the editor?

bolinfest:
@nathansobo Did https://github.com/atom/grim/commit/8807010397e64f499b6815090d7b575aaf4a27c3#diff-dd248f5f751e9429ed611987b50f19b6R48 fix or create the problem with Grim? Because I thought that was the code I saw when I hit a breakpoint in the debugger.

bolinfest:
Also, `apm test` from the command line fails with:

```
Error: Specify at least one test path

Tests failed
```

I have tried the following commands, but all yield the same result:

```
apm test
apm test spec
apm test spec/projects-spec.js
```

nathansobo:
The problem came from making Error.prepareStackTrace write only... you can revert that commit for now if Grim is definitely up to date, and we'll try again. I'm not at my machine or I would do it.

bolinfest:
:+1:

nathansobo:
Just run `atom --test`. We may have had a miscommunication that dropped support for the --specDirectory flag prematurely.

bolinfest:
Hmm, `atom --test spec` yields:

> Error Downloading Update: Could not get code signature for running application

To make sure I'm running the right executables, I tried to do **File** -> **Install Shell Commands**, but that pops up an error with **Error Downloading Update: Could not get code signature for running application**. I can't seem to copy/paste the stacktrace :(

thomasjo:
@nathansobo Should I merge https://github.com/atom/apm/pull/445? I don't have access to publish a new release however, so I guess it won't be of much help.

bolinfest:
I'm trying a clean build now.

bolinfest:
Nope, same stuff with a clean build :( There's no other output besides "Error Downloading Update" and the exit code is 1.

thomasjo:
@bolinfest I can reproduce what you're seeing.

bolinfest:
Oh hey, `atom --dev --test` did stuff! Seemingly the right stuff!

bolinfest:
Though the Atom windows the test runner popped up seem to be taking awhile to die...

thomasjo:
> Oh hey, atom --dev --test did stuff! Seemingly the right stuff!

Confirmed! I thought that was fixed in this PR. Never really understood why we need this flag. I first encountered it in apm a while back.

@nathansobo Can we drop the need for the `--dev` flag? Shouldn't it be implicit that specs should run in dev mode?

thomasjo:
@bolinfest They died pretty quick here (basically instantaneous). Although my specs are rather trivial with minimal UI interaction... could be a factor.

bolinfest:
Where does `console.log()` from my test runner go?

bolinfest:
Hmm, I guess it does the normal thing when I run the specs from Atom. But when I run from the command line, I get nothing. My first attempt at a testrunner was:

```
'use babel';

module.exports = function({testPaths, buildAtomEnvironment, logFile, headless, legacyTestRunner}) {
  console.log(arguments);
  return Promise.resolve(0);
};
```

But I can iterate/inspect from doing things through the UI instead. However, I think it's important to be able to caveman debug, especially from unit tests. Is the test runner supposed to redirect to the logFile or something itself?

as-cii:
This wasn't changed with this PR, but with the recent electron upgrade.

With the new chrome version, logging gets disabled by default. Therefore,
when running tests, you should pass `--enable-logging` (there should be
more info in the electron release notes).

On sab 17 ott 2015 at 09:54 bolinfest <notifications@github.com> wrote:

> Hmm, I guess it does the normal thing when I run the specs from Atom. But
> when I run from the command line, I get nothing. My first attempt at a
> testrunner was:
>
> 'use babel';
>
> module.exports = function({testPaths, buildAtomEnvironment, logFile, headless, legacyTestRunner}) {
>   console.log(arguments);
>   return Promise.resolve(0);
> };
>
> But I can iterate/inspect from doing things through the UI instead.
> However, I think it's important to be able to caveman debug, especially
> from unit tests. Is the test runner supposed to redirect to the logFile or
> something itself?
>
> —
> <https://github.com/atom/atom/pull/8968#issuecomment-148893765>.
>
-- 
Antonio

nathansobo:
The `--dev` and `--enable-logging` flags should no longer be required on latest master. Still can't run via `apm test` but I'm going to deal with that on Monday because there's really no reason to run things that way over `atom --test`. I couldn't reproduce any issues with `Grim` and `Error.prepareStackTrace`.

bolinfest:
@as-cii Is it also Electron that prepends the enormous `[12122:1017/115954:INFO:CONSOLE(19)] ` to every line? (So does the file and line number at the end.) It's pretty meaningless/distracting, and it would be nice to disable that, if possible. It really messes with things like copy/paste and other stuff.

thomasjo:
Indirectly. I'm pretty sure that's actually from Chromium. @zcbenz definitely knows.

bolinfest:
Is the idea that I should do the following in my test?

```
global.atom = buildAtomEnvironment();
```

I'm getting a failure with:

```
[15521:1017/120920:INFO:CONSOLE(61)] "TypeError: Cannot read property 'setRepresentedFilename' of undefined
    at Workspace.module.exports.Workspace.updateWindowTitle (/Users/mbolin/src/atom/src/workspace.coffee:172:27)
    at Workspace.updateWindowTitle (/Users/mbolin/src/atom/src/workspace.coffee:1:1)
    at Workspace.module.exports.Workspace.subscribeToActiveItem (/Users/mbolin/src/atom/src/workspace.coffee:123:6)
    at Workspace.module.exports.Workspace.subscribeToEvents (/Users/mbolin/src/atom/src/workspace.coffee:77:6)
    at new Workspace (/Users/mbolin/src/atom/src/workspace.coffee:52:6)
    at new AtomEnvironment (/Users/mbolin/src/atom/src/atom-environment.coffee:175:22)
    at testRunner.buildAtomEnvironment (/Users/mbolin/src/atom/src/initialize-test-window.coffee:48:43)
    at module.exports (/Users/mbolin/fbsource/fbobjc/Tools/Nuclide/pkg/nuclide/test/framework-atom/lib/main.js:17:17)
    at Object.<anonymous> (/Users/mbolin/src/atom/src/initialize-test-window.coffee:44:13)
    at Object.<anonymous> (/Users/mbolin/src/atom/src/initialize-test-window.coffee:2:1)", source: /Users/mbolin/src/atom/src/initialize-test-window.coffee (61)
```

bolinfest:
Note that I am launching the test runner via the command line.

bolinfest:
Oh, so it looks like the constructor for `AtomEnvironment` has default constructor args:

```coffee
  constructor: (params={}) ->
```

but then it seems like all these fields will be undefined if you don't supply them via `params`:

```coffee
    {@applicationDelegate, @window, @document, configDirPath, @enablePersistence} = params
```

and then the error I am getting is because `@applicationDelegate` is `undefined`.

bolinfest:
Also, I don't know what this comment means:

```coffee
  # Call .loadOrCreate instead
  constructor: (params={}) ->
    {@applicationDelegate, @window, @document, configDirPath, @enablePersistence} = params
```

That is, I don't see `loadOrCreate` anywhere.

bolinfest:
Hmm, I just rebased to master (eea75b6) and now I don't see the `[15521:1017/120920:INFO:CONSOLE(21)]` prefixes to my console.log() statements anymore when I run `atom --enable-logging --test path/to/spec`.

bolinfest:
OK, I put up a quick fix so that I can use `buildAtomEnvironment()`: https://github.com/atom/atom/pull/9195.

bolinfest:
Hmm, I see there is a `buildDefaultApplicationDelegate` function that is also passed in. Should I be using that instead?

bolinfest:
OK, for reference, here is the new skeleton I am using for a custom test runner. Is this what is expected?

```js
'use babel';
/* @flow */

/*
 * Copyright (c) 2015-present, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the license found in the LICENSE file in
 * the root directory of this source tree.
 */

type TestRunnerParams = {
  /** Absolute paths to tests to run. Could be paths to files or directories. */
  testPaths: Array<string>,

  /** A boolean indicating whether or not the tests are running headless. */
  headless: boolean,

  /** Creates the `atom` global object. */
  buildAtomEnvironment: (params: BuildAtomEnvironmentParams) => AtomGlobal,

  /** Currently undocumnted, but seemingly necessary to use buildAtomEnvironment(). */
  buildDefaultApplicationDelegate: () => Object,

  /** An optional path to a log file to which test output should be logged. */
  logFile: ?string,

  /** Unclear what the contract of this is, but we will not be using it. */
  legacyTestRunner: () => void,
}

type BuildAtomEnvironmentParams = {
  applicationDelegate: Object,
  window: Object,
  document: Object,
  configDirPath?: string,
  enablePersistence?: boolean,
}

/**
 * This function satisfies the contract for a custom Atom test runner.
 */
export default async function({
  testPaths,
  buildAtomEnvironment,
  buildDefaultApplicationDelegate,
  }: TestRunnerParams,
): Promise<number> {
  let applicationDelegate = buildDefaultApplicationDelegate();
  let params = {
    applicationDelegate,
    window,
    document,
  };
  global.atom = buildAtomEnvironment(params);

  // TODO: Run tests for testPaths.

  // Exit with zero if successful.
  return 0;
}
```

nathansobo:
Argh. Still having trouble reproducing any problem with the `TextEditor` export on a fresh build.

screen shot 2015-10-19 at 11 49 42 am

As for your description, it's exactly what I had in mind. Is it working for you?

bolinfest:
Just so everyone knows, we figured this out: https://github.com/atom/grim/commit/8807010397e64f499b6815090d7b575aaf4a27c3#commitcomment-13887073

We also concluded that the `TextEditor` constructor needs to be patched up so it still maintains its existing, albeit not-guaranteed since it was never published as part of 1.0, contract.

nathansobo:
> We also concluded that the TextEditor constructor needs to be patched up so it still maintains its existing, albeit not-guaranteed since it was never published as part of 1.0, contract.

Sorry, it turns out that the error we were seeing was due to an attempt to subclass the existing `TextEditor` constructor, which wasn't covered by our shim. @maxbrunsfeld I'm wondering if we should continue to export `TextEditor` and just make invoking it as a constructor a deprecated action. We could maintain an internal `Symbol` that can be passed in by `Workspace.prototype.buildTextEditor` to suppress the deprecation warning. Then it will "just work" for people attempting to subclass, and people can continue to use `instanceof` checks beyond 2.0, which I don't really have a problem with.

bolinfest:
When running a test, is there any way to remove all of the output that surrounds what I write to `console.log()`? It would be helpful to have the option to have "clean" output that doesn't have the `[6237:0226/152313:INFO:CONSOLE(39)]` prefix or the `source: ...` suffix.

bolinfest:
Currently, it seems like specifying a `--log-file` and doing `fs.appendFileSync()` is my best workaround, but it's pretty tedious to get the path to the log file everywhere I need it as compared with `console.log()`.

50Wliu:
@bolinfest see #10952.

bolinfest:
Hmm, so there's also no way to specify additional flags/arguments to your test runner? Maybe we could use `--` and pass a list of strings specified after that?

Alhadis:
Hrm, is there currently no way of running a custom test-runner inside the usual spec window? (Via <kbd>View</kbd> -> <kbd>Developer<kbd> -> <kbd>Run Package Specs</kbd>).

I'd give my left nut to use Mocha/Chai instead of Jasmine.

nathansobo:
@Alhadis https://github.com/binarymuse/atom-mocha-test-runner

nathansobo:
We're working on making this more official and documented.

Alhadis:
Nah it's cool, I already wrote my own... =) Just realised too damn late the reason it wasn't running is because the `package.json` had been cached in memory (and reloading the window wasn't doing anything, I literally had to quit and restart).

Alhadis:
It'd be lovely if we could use `package.json` to pass arbitrary values to the test-runner:

```json
"atomTestRunner": {
	"path": "./path/to/test-runner",
	"config": {
		…
	}
}
```

The `atomTestRunner` setting could be extended to accept an object instead of a plain string, treating the latter as if `{"path": "…"}` were passed instead. Which is similar to NPM's treatment of `repository` and `author` properties, among many others.

Alhadis:
@nathansobo [Uh, this kinda snowballed, sorry...](https://www.npmjs.com/package/atom-mocha)

yamsellem:
@nathansobo, making it official and documented will be great, but it takes some time, meanwhile, may you please explain in a nutshell how we can take use of it today?

It seems there is [already a mocha plugin](https://github.com/binarymuse/atom-mocha-test-runner), and I can wait to launch some mocha test in Atom. Existing mocha integration are not good, it seems Nuclide starts something in the way but too much bloated to even understand it.

Thanks.

Alhadis:
@nathansobo [Try this](https://github.com/Alhadis/Atom-Mocha/). Seriously. :)

nathansobo:
@Alhadis Looks pretty cool. I like the black background. I'll keep it in mind next time I'm considering a runner, but right now I'm pretty deep into fixing some international keybindings stuff.

nathansobo:
@yamsellem You can specify a path in your `package.json` to point to a file that exports a function. This function will be invoked in your test window with various parameters that are useful for implementing a custom test runner. Take a look at the source of one of the existing runners for ideas on how to use those parameters to build your own runner.

PR #8972, opened 2015-09-28
Update reference to deprecated method Project::getRepo

### LINKED ISSUE #8971: deprecated method call in api documentation
https://atom.io/docs/api/v1.1.0-beta.0/GitRepository

getRepo() should be getRepositories()

https://github.com/jacogr/atom-git-control/issues/64

thomasjo:
GitRepository class documentation references deprecated method `Project::getRepo`. This resolves that by instead using the replacement `Project::getRepositories`.

Although, it seems as if that will be removed in v2.0, so perhaps we should update this to instead using the async `Project::getDirectories` method, coupled with `Project::repositoryForDirectory`?

/cc @atom/feedback

---

Fixes #8971

### DIFF
--- a/src/git-repository.coffee
+++ b/src/git-repository.coffee
@@ -10,8 +10,8 @@ Task = require './task'
 # Extended: Represents the underlying git operations performed by Atom.
 #
 # This class shouldn't be instantiated directly but instead by accessing the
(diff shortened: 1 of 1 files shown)


lee-dohm:
Looks good to me :ship:

PR #8985, opened 2015-09-29
Fix deprecations

as-cii:
This is a continuation of #8905.

We deprecated a bunch of methods, but still used them in our core and in bundled packages as well. This PR changes that so that we no longer use the deprecated methods anymore.

Also, Travis will now fail if any deprecated method is hit by our specs.
- [x] Merge https://github.com/atom/find-and-replace/pull/540 and bump version.
- [x] Merge https://github.com/atom/symbols-view/pull/124 and bump version.
- [x] Merge https://github.com/atom/wrap-guide/pull/44 and bump version.
- [x] atom core
  - [x] `::page{Up,Down}`
  - [x] `::selectPage{Up,Down}`

/cc: @maxbrunsfeld @nathansobo


nathansobo:
Thanks for doing this. These changes look fine to me, and I doubt the package PRs are anything controversial so proceed at your discretion.

as-cii:
Everything is :green_heart: so I'll go ahead and :ship: this. In addition to fix deprecated methods I have included some other things:

1. Provide a sensible default for getters when `component` is `null`.
2. Add a feature flag to detect when the logical display buffer gets released (so that we could avoid to break packages).
3. Allow to listen for `TextEditorElement` events (namely, only `onDidChangeScrollLeft` and `onDidChangeScrollTop`) even when the element is not attached.

Thanks everybody! :+1:

PR #8992, opened 2015-09-29
Master

FlowerBoot:
README汉化完成

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 [![Dependency Status](https://david-dm.org/atom/atom.svg)](https://david-dm.org/atom/atom)
 [![Join the Atom Community on Slack](http://atom-slack.herokuapp.com/badge.svg)](http://atom-slack.herokuapp.com/)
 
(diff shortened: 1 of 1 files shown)


thomasjo:
Sorry, we are not going to translate the (main) README from English to Chinese.

steelbrain:
You could however make a subdir of translated manual. Like it's done in the electron repo. Or maybe we could create a separate docs repo where the translations live.
What do you think @thomasjo ?

thomasjo:
I vote no for translating anything in this repo. Translations in https://github.com/atom/docs would be nice in theory, but seeing how outdated some aspects of the current (english) documentation is, I'm super reluctant about adding translations anywhere.

I'll let other folks from @atom/feedback chime in since I'm likely being overly pessimistic.

lee-dohm:
I agree with @thomasjo. My priorities for translation would be:

1. Atom itself
1. Flight Manual
1. Developer documentation

PR #8999, opened 2015-09-30
Request feature: Run system shell command

KSXGitHub:
Dear Atom's developers,
Notepad++ editor has a feature called "Run" which run system command directly in the editor, I can assign a keystroke to each command, it makes conveniences.
I see no reason why Atom doesn't supports this.
So, please support this.


thomasjo:
I'm assuming you meant to open an issue and not a PR, so I'm going to close this.

As for your feature request, there are already many community packages which implement this concept; see https://atom.io/packages/search?q=run as a start.

Issue #7345, opened 2015-06-19
Proposal: Improve tab titles when opening files with the same name


patrick-steele-idem:
When opening multiple files with the same base name (but in different directories) it is hard to distinguish which tab is associated with which file. Currently, when Atom detects that multiple files with the same name are opened it appends the complete project path after the filename, but it is cut off (making it not very helpful). The user must use their mouse to hover over the tab to know the full path. I think the logic can be improved by including the parent directory name to the left of the file name as shown below:
# One tab

![one-tab](https://cloud.githubusercontent.com/assets/978214/8258063/4da379d8-166f-11e5-937f-e7b1e8f9413f.png)
# Two tabs (current)

![two-tabs](https://cloud.githubusercontent.com/assets/978214/8258067/517bbc5a-166f-11e5-88a9-867c8e91acdc.png)
# Two tabs (proposed)

![two-tabs-good](https://cloud.githubusercontent.com/assets/978214/8258068/53509da2-166f-11e5-8e08-a6bc5552ba23.png)

Thoughts?

maxbrunsfeld:
I like this idea. There are cases where it would do a *worse* job disambiguating the paths than the current behavior (e.g. `some-dir/src/stuff.c` and `another-dir/src/stuff.c`), but I think that in many cases it will do a better job. I also just think it's more readable this way - a single path rather than a name *and* a path separated by a hyphen.

mehcode:
For that .. how about `some-dir/../stuff.c` and `another-dir/../stuff.c`.
The idea would be to truncate starting from the common point or the middle (if no common point).

maxbrunsfeld:
> For that .. how about `some-dir/../stuff.c` and `another-dir/../stuff.c`.

:+1+ Perfect.

patrick-steele-idem:
I'm also in favor of the following:

> For that .. how about some-dir/../stuff.c and another-dir/../stuff.c.

:+1: 

On a related note, I do see a problem with tabs currently having a `max-width: 160px;` style. I feel like it is more important for a tab to take up the needed width than setting some arbitrary max width. I would drop the `max-width` style, but I'm not sure what other implications that would have on the UI.

maxbrunsfeld:
>  I do see a problem with tabs currently having a max-width: 160px; style.

That property is set by the UI theme. You might want to open an issue on the [Atom Dark UI](https://github.com/atom/atom-dark-ui) theme for that.

ollpu:
[...]

PR #9023, opened 2015-10-02
Add TextEditor.getUniqueTitle()

EasyHard:
This PR tries to address [issue 7345](https://github.com/atom/atom/issues/7345)

This method is intended to replace TextEditor.getLongTitle
and gives a better tab title when there are multiple opened
files have identical file name. It will returns 
`<unique-parent-directory>/.../<filename>`  when opened files
have identical file name. With this PR and updating `atom/tabs`
multiple files with identical name will look like this.

image


EasyHard:
Oops. I some typos are caught by travis-ci. I will create a new one to fix them.

PR #9024, opened 2015-10-02
Add TextEditor.getUniqueTitle()

EasyHard:
This PR tries to address issue #7345

This method is intended to replace TextEditor.getLongTitle
and gives a better tab title when there are multiple opened
files have identical file name. It will returns 
<unique-parent-directory>/.../<filename> when opened files
have identical file name. With this PR and updating atom/tabs
multiple files with identical name will look like this.

image

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -547,6 +547,45 @@ class TextEditor extends Model
     else
       'untitled'
 
+  # Essential: Get unique title for display in other parts of the UI
+  # such as the window title.
+  #
+  # If the editor's buffer is unsaved, its title is "untitled"
+  # If the editor's buffer is saved, its unique title is formatted as one
+  # of the following,
+  # * "<filename>" when it is the only editing buffer with this file name.
+  # * "<unique-dir-prefix>/.../<filename>", where the "..." may be omitted
+  # if the the direct parent directory is already different.
+  #
+  # Returns a {String}
+  getUniqueTitle: ->
+    if sessionPath = @getPath()
+      title = @getTitle()
+
+      # find text editors with identical file name.
+      paths = []
+      for textEditor in atom.workspace.getTextEditors() when textEditor isnt this
+        if textEditor.getTitle() is title
+          paths.push(textEditor.getPath())
+      if paths.length is 0
+        return title
+      fileName = path.basename(sessionPath)
+
+      # find the first directory in all these paths that is unique
+      nLevel = 0
+      while (_.some(paths, (apath) -> path.basename(apath) is path.basename(sessionPath)))
+        sessionPath = path.dirname(sessionPath)
+        paths = _.map(paths, (apath) -> path.dirname(apath))
+        nLevel += 1
+
+      directory = path.basename sessionPath
+      if nLevel > 1
+        path.join(directory, "...", fileName)
+      else
+        path.join(directory, fileName)
+    else
+      'untitled'
+
   # Essential: Get the editor's long title for display in other parts of the UI
   # such as the window title.
   #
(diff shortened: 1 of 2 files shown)


src/text-editor.coffee
```diff
@@ -547,6 +547,45 @@ class TextEditor extends Model
...
+  # If the editor's buffer is unsaved, its title is "untitled"
+  # If the editor's buffer is saved, its unique title is formatted as one
+  # of the following,
+  # * "<filename>" when it is the only editing buffer with this file name.
+  # * "<unique-dir-prefix>/.../<filename>", where the "..." may be omitted
+  # if the the direct parent directory is already different.
+  #
+  # Returns a {String}
+  getUniqueTitle: ->
+    if sessionPath = @getPath()
```
nathansobo: What do you mean by `sessionPath` in this case?

src/text-editor.coffee
```diff
@@ -547,6 +547,45 @@ class TextEditor extends Model
...
+      paths = []
+      for textEditor in atom.workspace.getTextEditors() when textEditor isnt this
+        if textEditor.getTitle() is title
+          paths.push(textEditor.getPath())
+      if paths.length is 0
+        return title
+      fileName = path.basename(sessionPath)
+
+      # find the first directory in all these paths that is unique
+      nLevel = 0
```
nathansobo: What's a more descriptive name for this?

nathansobo:
Hey @EasyHard, thanks a lot. This looks like a much nicer approach to building a unique title for the tabs when files have the same name. I have a couple clarifying questions. After those get clarified, I can get this caught up with master and inject the workspace dependency (we're trying to avoid using globals in core). Then we can start using this in packages. Thanks again.

src/text-editor.coffee
EasyHard: Probably `absolutePath` is a better name. `this.getPath` will return absolute path of the current buffer if I understand correctly.

src/text-editor.coffee
EasyHard: This is the distance from the closest distinct parent directory to the current file. By distance I mean how many directory level one needs to pass though to open the current file. `nLevel` is indeed too short for this, maybe `nLevelToDistinceParent` is a good idea?

EasyHard:
Hi @nathansobo . Thanks a lot. I add some comments according to your questions. Please ping me if I did not put myself clearly. And by the way, your effort of improving atom's testing environments is awesome. Easier testing will definitely make Atom more attractive.

nathansobo:
Thanks @EasyHard. I made some minor updates to the local variables and rebased. This PR will need to make its way into a release before we can use it in packages.

DouweM:
Any reason why we don't just change `getLongTitle` to get this unique title, so that we don't need to change the call in packages?

nathansobo:
@DouweM, Good idea. Changed on the `beta` branch (af22a149bdd8989f5c279863ff319bdc652d7c66). Heads up @EasyHard that `getUniqueTitle` will be gone in the next `beta` hotfix update.

Ben3eeE:
@nathansobo What happened to this? I still see the old format on tab names.

nathansobo:
I'm actually not sure. Maybe you could pickaxe the git history?

Ben3eeE:
I found it, this version only lived for a few releases and was replaced in/around 1.3.3 with https://github.com/atom/atom/pull/9636.

Shame cause this looks better than the current. Looks like the images in #9636 does not use the format described in this PR? I don't know what happened, also the last comment in #9636 is confused as well.

Tag with change.
https://github.com/atom/atom/blob/v1.3.3/src/text-editor.coffee

lee-dohm:
@Ben3eeE This PR caused a crashing bug: https://github.com/atom/atom/issues/10043. This is why the behavior was essentially reverted.

Ben3eeE:
@lee-dohm I see. Maybe https://github.com/atom/atom/issues/7345 can be closed? The current look is similar to what it was before #9636 but still improved over what it was before this.

I found https://github.com/atom/tabs/issues/256 which sounds like what I really want. That way I can make whatever look I feel like!

PR #9029, opened 2015-10-02
Add `onDidStopChangingActivePaneItem` for async callbacks

### LINKED ISSUE #8997: Consider an `onDidStopChangingActivePaneItem`
Similar to the motivation behind [`TextEditor::onDidStopChanging`](https://atom.io/docs/api/v1.1.0-beta.0/TextEditor#instance-onDidStopChanging),  clients of `onDidChangeActivePaneItem` should nearly always wait for a small time period before acting on the new pane item. If the user is quickly switching through tabs or closing several tabs in quick succession, work done on [`Workspace::onDidChangeActivePaneItem`](https://atom.io/docs/api/v1.1.0-beta.0/Workspace#instance-onDidChangeActivePaneItem) is likely wasted and can become a bottleneck for giving feedback to the user if they are switching/closing tabs quickly.

There's at least one candidate in core that could defer its actions until the active item has stopped changing: [`FileInfoView`](https://github.com/atom/status-bar/blob/master/lib/file-info-view.coffee#L71) in the status bar.

Nuclide wrapped `onDidChangeActivePaneItem` to get this effect: https://github.com/facebook/nuclide/blob/master/pkg/nuclide/atom-helpers/lib/atom-event-debounce.js#L29

I think a `DidStopChangingActivePaneItem` should be recommended similar to the one in `TextEditor` if it's created.

ssorallen:
`onDidChangeActivePaneItem` is called synchronously when the active pane
item changes, and several non-critical actions preform work on that
event. Critical UI feedback, like changing the active tab, needs to
happen synchronously, but most other functionality should be run
asynchronously.

The code follows the same naming scheme and design pattern as
[`TextBuffer::onDidStopChanging`](https://github.com/atom/text-buffer/blob/v7.1.2/src/text-buffer.coffee#L184), which is exposed as
`TextEditor::onDidStopChanging`.

Fixes #8997.

### DIFF
--- a/src/pane-container.coffee
+++ b/src/pane-container.coffee
@@ -12,6 +12,8 @@ class PaneContainer extends Model
   @version: 1
 
   root: null
+  stoppedChangingActivePaneItemDelay: 100
+  stoppedChangingActivePaneItemTimeout: null
 
   @deserialize: (state) ->
     container = Object.create(@prototype) # allows us to pass a self reference to our child before invoking constructor
@@ -81,6 +83,9 @@ class PaneContainer extends Model
 
   onDidChangeActivePaneItem: (fn) ->
     @emitter.on 'did-change-active-pane-item', fn
+
+  onDidStopChangingActivePaneItem: (fn) ->
+    @emitter.on 'did-stop-changing-active-pane-item', fn
 
   observeActivePaneItem: (fn) ->
     fn(@getActivePaneItem())
@@ -189,12 +194,18 @@ class PaneContainer extends Model
 
   # Called by Model superclass when destroyed
   destroyed: ->
+    @cancelStoppedChangingActivePaneItemTimeout()
     pane.destroy() for pane in @getPanes()
     @subscriptions.dispose()
     @emitter.dispose()
 
+  cancelStoppedChangingActivePaneItemTimeout: ->
+    if @stoppedChangingActivePaneItemTimeout?
+      clearTimeout(@stoppedChangingActivePaneItemTimeout)
+
   monitorActivePaneItem: ->
     childSubscription = null
+
     @subscriptions.add @observeActivePane (activePane) =>
       if childSubscription?
         @subscriptions.remove(childSubscription)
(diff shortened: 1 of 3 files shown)


src/pane-container.coffee
```diff
@@ -15,6 +15,8 @@ class PaneContainer extends Model
   @version: 1
 
   root: null
+  stoppedChangingActivePaneItemDelay: 100
```
ssorallen: 100ms is arbitrary. `TextEditor::didStopChanging` uses 300ms, which seemed too long here but for no measured reason. Does 100ms seem too short or too long?

benogle:
cc @nathansobo @maxbrunsfeld

nathansobo:
This seems like a good idea. I wish the async version could be named `onDidChangePaneItem` and have the synchronous version be called `onDidChangePaneItemSync` to steer people more toward the async version, but that :ship: has already sailed.

Can you add documentation to `onDidChangePaneItem` to steer them toward this method?

ssorallen:
Sublime Text uses an `_async` suffix for its [`sublime_plugin.EventListener`](https://www.sublimetext.com/docs/3/api_reference.html#sublime_plugin.EventListener) API, but using `Sync` is definitely more Node-like. The `onDidStop` prefix is too wordy, but I picked up the naming pattern from `TextEditor`. We could choose something else for this and future async functions like it.

nathansobo:
Yeah, I'm not wild about the `Async` suffix either. I think your choice of naming was pretty decent from a blending-in perspective.

ssorallen:
:+1: I'll point the sync version to this one and fix the failing spec.

ssorallen:
@nathansobo I reused the description from `TextEditor::didChange` to point developers to `::onDidStopChangingActivePaneItem`. The specs are set as well.

src/pane-container.coffee
maxbrunsfeld: I think 100ms is a good starting point. We can adjust it once we start using this method in packages.

nathansobo:
Thanks very much @ssorallen!

maxbrunsfeld:
:ok_hand: This is nice. @ssorallen Could you ping me on any PRs that you open for bundled packages to use this method?

ssorallen:
@maxbrunsfeld Definitely, will do. I found 3 or 4 places that were prime use cases for this. I'll add both you and @nathansobo on the PRs.

PR #9042, opened 2015-10-04
add option to choose whether atom rememebers your last windows and project folders

### LINKED ISSUE #6605: Add setting to disable restoring last session
It's useful that atom can remember the files/folders you had open last time and then re-open them next time you start atom. I can't seem to figure out how to turn off this behavior though, I generally _want_ to start clean each time I open atom.

Is there a config option to enable/disable this feature? Or could we implement a dialog box when closing atom that asks if you want it to remember what you had open?

jordanbtucker:
A new Core Setting: **Restore Previous  Windows On Start**

Enabled by default, so it's backward compatible.

Fixes #6605

### DIFF
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -268,6 +268,36 @@ describe "Starting Atom", ->
                 [otherTempDirPath]
               ].sort()
 
+    it "doesn't reopen any previously opened windows if restorePreviousWindowsOnStart is disabled", ->
+      configPath = path.join(atomHome, 'config.cson')
+      config = CSON.readFileSync(configPath)
+      config['*'].core = {restorePreviousWindowsOnStart: false}
+      CSON.writeFileSync(configPath, config)
+
+      runAtom [tempDirPath], {ATOM_HOME: atomHome}, (client) ->
+        client
+          .waitForExist("atom-workspace")
+          .waitForNewWindow(->
+            @startAnotherAtom([otherTempDirPath], ATOM_HOME: atomHome)
+          , 5000)
+          .waitForExist("atom-workspace")
+
+      runAtom [], {ATOM_HOME: atomHome}, (client) ->
+        windowProjectPaths = []
+
+        client
+          .waitForWindowCount(1, 10000)
+          .then ({value: windowHandles}) ->
+            @window(windowHandles[0])
+            .waitForExist("atom-workspace")
+            .treeViewRootDirectories()
+            .then ({value: directories}) -> windowProjectPaths.push(directories)
+
+            .call ->
+              expect(windowProjectPaths).toEqual [
+                []
+              ]
+
   describe "opening a remote directory", ->
     it "opens the parent directory and creates an empty text editor", ->
       remoteDirectory = 'remote://server:3437/some/directory/path'
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -506,18 +506,17 @@ class AtomApplication
 
   saveState: (allowEmpty=false) ->
     return if @quitting
-    restorePreviousState = @config.get('core.restorePreviousWindowsOnStart') ? true
     states = []
-    if restorePreviousState
-      for window in @windows
-        unless window.isSpec
-          if loadSettings = window.getLoadSettings()
-            states.push(initialPaths: loadSettings.initialPaths)
-    if states.length > 0 or allowEmpty or not restorePreviousState
+    for window in @windows
+      unless window.isSpec
+        if loadSettings = window.getLoadSettings()
+          states.push(initialPaths: loadSettings.initialPaths)
+    if states.length > 0 or allowEmpty
       @storageFolder.storeSync('application.json', states)
 
   loadState: (options) ->
-    if (states = @storageFolder.load('application.json'))?.length > 0
+    restorePreviousState = @config.get('core.restorePreviousWindowsOnStart') ? true
+    if (states = @storageFolder.load('application.json'))?.length > 0 and restorePreviousState
       for state in states
         @openWithOptions(_.extend(options, {
           initialPaths: state.initialPaths
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -3,6 +3,7 @@ ApplicationMenu = require './application-menu'
 AtomProtocolHandler = require './atom-protocol-handler'
 AutoUpdateManager = require './auto-update-manager'
 StorageFolder = require '../storage-folder'
+Config = require '../config'
 ipcHelpers = require '../ipc-helpers'
 {BrowserWindow, Menu, app, dialog, ipcMain, shell} = require 'electron'
 fs = require 'fs-plus'
@@ -70,7 +71,11 @@ class AtomApplication
     @pidsToOpenWindows = {}
     @windows = []
 
-    @autoUpdateManager = new AutoUpdateManager(@version, options.test, @resourcePath)
+    @config = new Config({configDirPath: process.env.ATOM_HOME, @resourcePath, enablePersistence: true})
+    @config.setSchema null, {type: 'object', properties: _.clone(require('../config-schema'))}
+    @config.load()
+
+    @autoUpdateManager = new AutoUpdateManager(@version, options.test, @resourcePath, @config)
     @applicationMenu = new ApplicationMenu(@version, @autoUpdateManager)
     @atomProtocolHandler = new AtomProtocolHandler(@resourcePath, @safeMode)
 
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -506,12 +506,14 @@ class AtomApplication
 
   saveState: (allowEmpty=false) ->
     return if @quitting
+    restorePreviousState = @config.get('core.restorePreviousWindowsOnStart') ? true
     states = []
-    for window in @windows
-      unless window.isSpec
-        if loadSettings = window.getLoadSettings()
-          states.push(initialPaths: loadSettings.initialPaths)
-    if states.length > 0 or allowEmpty
+    if restorePreviousState
+      for window in @windows
+        unless window.isSpec
+          if loadSettings = window.getLoadSettings()
+            states.push(initialPaths: loadSettings.initialPaths)
+    if states.length > 0 or allowEmpty or not restorePreviousState
       @storageFolder.storeSync('application.json', states)
 
   loadState: (options) ->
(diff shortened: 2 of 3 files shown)


thedaniel:
:-1: on loading config twice. Is there a way you could achieve the same goal without this? We have to be tediously strict about counting milliseconds on window load.

thedaniel:
Also, what's your use case here? I'm personally not convinced this is a common enough need to warrant another checkbox in core settings. cc @atom/feedback

50Wliu:
> :-1: on loading config twice.

Ditto.  My window load time is long enough as it is :wink:.

As for use case, a lot of people have been asking for this in #6605.

m1ga:
@jordanbtucker  wouldn't it be enough just to save an empty state when closing atom and the setting is switched on. So you won't need to change to opening part at all.

jordanbtucker:
@thedaniel @50Wliu I totally agree with you on the loading config issue, but I haven't come up with a better solution, and config option is what 99% of the users wanted in their comments. I'm okay with removing the checkbox and making it config file only.

I'm against having a command that disables saving on exit that you have to call every time. This should be a set-and-forget setting.

Honestly, I feel this is on par with the "open empty editor" config setting. Seems odd to give that a checkbox, but not this one.

Please leave any suggestions to improve this. I'll be so glad when this is fixed.

m1ga:
> I'm against having a command that disables saving on exit that you have to call every time. This should be a set-and-forget setting.

I meant in your example you are returning out of the save function when your new setting is switched on and you have the same for the opening function. If you just change the saving part to save [] if you setting is switched on you don't need to do any changes on the loading part (less code to test, less side-effects). saveState/loadState do their work normally but you end up with a clean session every time.

jordanbtucker:
@m1ga I hadn't responded to your comment yet. With my method, you can modify the config file with the editor closed and it will honor your setting when it starts. If we go with your method, the option should be changed to "Save Your Previously Opened Windows" or something like that. Or maybe I'm just nitpicking. I'm not opposed to your method, but I just want everyone to be aware before we decide which way to go.

samis:
I am personally fine with it not being a checkbox. However, there should be some way that the user can find out about the option if they want to. Perhaps the main settings page links to a 'Configuration Reference' saying it contains additional options not displayed here?

jordanbtucker:
I removed the checkbox and it now only loads the config file when saving. I'm not sure the best way to use a config option that doesn't have a UI. Do I set the default in the code like I did on [line 438][L438]? What other examples are there of settings that do not have a UI?

[L438]: https://github.com/jordanbtucker/atom/commit/dd6571493c7de132b7f9c0bdc9083cfa82b349bb#diff-9929202e800354a2caa775ed44c61bfeR438

spec/integration/startup-spec.coffee
```diff
@@ -241,6 +241,42 @@ describe "Starting Atom", ->
...
+            .window(windowHandles[1])
+            .waitForExist("atom-workspace")
+            .treeViewRootDirectories()
+            .then ({value: directories}) -> windowProjectPaths.push(directories)
+
+            .call ->
+              expect(windowProjectPaths).toEqual [
+                []
+                []
+              ]
```
maxbrunsfeld: Nice work adding this integration test. The assertion doesn't seem right to me though; this is saying that there should be *two* Atom windows, each with *no* project paths. If we've chosen to *not* restore the session though, shouldn't there just be *one* Atom window?

src/browser/atom-application.coffee
```diff
@@ -428,12 +428,22 @@ class AtomApplication
...
-      unless window.isSpec
-        if loadSettings = window.getLoadSettings()
-          states.push(initialPaths: loadSettings.initialPaths)
-    if states.length > 0 or allowEmpty
+    if restorePreviousState
+      for window in @windows
+        unless window.isSpec
+          if loadSettings = window.getLoadSettings()
+            states.push(initialPaths: loadSettings.initialPaths)
+    if states.length > 0 or allowEmpty or not restorePreviousState
```
maxbrunsfeld: I think it would be better to change the `loadState` method instead of `saveState`. With your current code, if there was a pre-existing `application.json` file, it would not be *updated*, but it would always be restored on startup. I think instead, you should make `loadState` return `false` immediately if `core.restorePreviousWindowsOnStart` is `false`.

maxbrunsfeld:
@jordanbtucker Thanks for diving in and taking a look at this. I'm :+1: on this change; a number of people have asked for it. I agree w/ @thedaniel that we don't need to add UI for it yet. I left some feedback on your code above.

jordanbtucker:
Good catch on the integration test. I can't actually get integration tests to work on my Ubuntu box, but hopefully the test is correct now.

src/browser/atom-application.coffee
```diff
@@ -428,12 +428,22 @@ class AtomApplication
 
   saveState: (allowEmpty=false) ->
     return if @quitting
+
+    Config = (require '../config')
+    config = new Config
+      configDirPath: process.env.ATOM_HOME
+      resourcePath: @resourcePath
+    config.setSchema null, {type: 'object', properties: _.clone(require('../config-schema'))}
+    config.load()
```
maxbrunsfeld: Currently, I believe this will throw an exception when the user's config is invalid, because it will [try to reference](https://github.com/atom/atom/blob/73728f501a9a7cde5b473e38f8e773a1c3efd847/src/config.coffee#L782) `atom.notifications`, which won't be defined in the browser process. If we're going to use the `Config` class here, I think we'll need to remove its dependency on the `NotificationManager`.

That code is also going to change a little bit when https://github.com/atom/atom/pull/8968 lands. Maybe instead of creating the notification inside of `Config`, we should add a method `::onDidFailToLoad(callback)`, and we can use that method in the renderer process to set up a callback that creates the notification.

docrinehart:
+1 Bump! I'd really like to have Atom open clean every time and stop saving my session when I close it

maxbrunsfeld:
@jordanbtucker Are you still interested in doing this? It has actually become a little bit *easier* since we last talked, because of https://github.com/atom/atom/pull/9990. We are now using the `Config` class in the browser process for other purposes.

jordanbtucker:
Sure! I'll take a look at #9990 and start working on this again. Thanks.

docrinehart:
@jordanbtucker Yay! You rock! And thanks for the quick reply and review @maxbrunsfeld

jordanbtucker:
@maxbrunsfeld I looked through #9990, and the relevant code looks almost identical to what I've already written. Sorry if I'm missing something, but what do I still need to change to get this merged?

maxbrunsfeld:
@jordanbtucker We just need to avoid loading the config twice in the browser process. Currently, it's loaded inside of `auto-update-manager`, so I think we need to move that code out into `atom-application`, passing the `config` into the `AutoUpdateManager`.

jacobgraf:
Excited for this! Personally, I want a clean slate every time Iaunch Atom.

mehcode:
You can shim this in atom right now with [project-plus](https://atom.io/packages/project-plus) and the following line in `init.coffee` (as long as you only use one atom window):

```
atom.commands.dispatch(atom.views.getView(atom.workspace), 'project-plus:close')
```

jordanbtucker:
@maxbrunsfeld I made some changes that I think satisfy your requirements. Please take a look and let me know. Thanks.

lee-dohm:
@jordanbtucker Max is on parental leave currently. I've added this to the queue to review but I don't have an ETA on when someone will be able to get to it.

as-cii:
Thanks for the patience and the hard work on this, @jordanbtucker. :sparkles:

I have started reviewing this and I think right now we may not be handling the scenario where the `restorePreviousWindowsOnStart` setting is set to `false` outside of Atom (e.g. `nano ~/.atom/config.cson`). In that case, we'd still reopen the previous windows even if the configuration explicitly states we shouldn't.

So, what do you think about moving that `if` clause down to the `loadState` method? Basically, we'd always store the currently open windows, but we open them on the next launch only if the configuration states so. If we do that, we might also want to change the integration test to reflect this behavior; in particular, we could move this piece of code down, so that it gets executed before opening the second instance:

```coffee
configPath = path.join(atomHome, 'config.cson')
config = CSON.readFileSync(configPath)
config['*'].core = {restorePreviousWindowsOnStart: false}
CSON.writeFileSync(configPath, config)
```

What do you think? /cc: @atom/core

jacobgraf:
I don't know much about the technicalities of this, but based on your comments, why does this have to be so difficult? Why save the previously opened windows at all unless explicitly told to do so? From the people I have talked to, none of them want the windows saved when they fully close the editor. Frankly, it's a pain in the butt that it does that currently.

mnquintana:
@jacobgraf This was a highly requested feature by the community: https://github.com/atom/atom/issues/1603

FWIW, Sublime also remembers its open windows and files by default, so this is actually expected behavior for those coming from Sublime.

Once this PR lands then you'll be able to disable this.

jordanbtucker:
@as-cii Thanks for the feedback. You make a good point.

I've pushed the changes you suggested.

jordanbtucker:
One thing I want to make clear to everyone (in case there is any confusion). This only applies to opening Atom _without_ a path argument.

If your config file looks like this:

```cson
'*'
  core:
    restorePreviousWindowsOnStart: false
```

Then the following command will not open the previous session:

```bash
$ atom    # will not restore previous session
```

However, the following commands will restore the previous session _for that directory_.

```bash
# this will restore the state for this directory (if it was
# recently opened) regardless of restorePreviousWindowsOnStart
$ atom /home/me/project

# this will restore the state for the parent directory (if it was recently
# opened) of this file regardless of restorePreviousWindowsOnStart
$ atom /home/me/project/package.json
```

spec/integration/startup-spec.coffee
```diff
@@ -268,6 +268,36 @@ describe "Starting Atom", ->
                 [otherTempDirPath]
               ].sort()
 
+    it "doesn't reopen any previously opened windows if restorePreviousWindowsOnStart is disabled", ->
+      configPath = path.join(atomHome, 'config.cson')
+      config = CSON.readFileSync(configPath)
+      config['*'].core = {restorePreviousWindowsOnStart: false}
+      CSON.writeFileSync(configPath, config)
```
as-cii: What do you think about moving this section before [this LOC](https://github.com/atom/atom/pull/9042/files#diff-a5dc259060c4c3c53c128b8bd0f02d23R285)? This way we make sure that the setting works independently of when we change the configuration.

as-cii:
Just left one comment, I think we can :shipit: after addressing it and specs are :green_heart:. :fireworks:

PR #9052, opened 2015-10-05
beautify package.json

fscherwi:


steelbrain:
:+1:

kevinsawicki:
This looks like it is just adding optional `.0` minor and patch values. What is the motivation to add them since they are optional?

steelbrain:
IMO this consistency with other values looks better.

benogle:
Closing as this is super out of date

PR #9062, opened 2015-10-06
upgrade asar for builds

fscherwi:


50Wliu:
What does this fix?

thomasjo:
This resolves the regression in asar where hidden files (i.e. files prefixed with a period) are not packaged. Basically this is what broke symbols-view in the Electron upgrade, which I worked around by renaming `.ctags` to `ctags-config`.

thomasjo:
We don't really need this right now, but I'm going to merge this anyway. Thanks @fscherwi.

PR #9067, opened 2015-10-07
Avoid leaking DOM objects

as-cii:
With this PR we make sure to clear out all the references that caused `TextEditorComponent`s to be retained (e.g. closures, etc.). Below you can have a look at how the heap looked like after scrolling for a while on multiple editors and then closing them.

screen shot 2015-10-07 at 11 09 04

screen shot 2015-10-07 at 11 16 35

As you can see, a lot of elements were leaking because they were still retained by `TextEditorComponent`.

This is how the heap looks after the changes in this PR:

screen shot 2015-10-07 at 11 09 11

:non-potable_water: :non-potable_water: :non-potable_water:

/cc: @nathansobo @atom/feedback for :eyes:

### DIFF
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -119,6 +119,14 @@ class TextEditorComponent
     @gutterContainerComponent?.destroy()
     @domElementPool.clear()
 
+    @verticalScrollbarComponent.destroy()
+    @horizontalScrollbarComponent.destroy()
+
+    @verticalScrollbarComponent = null
+    @horizontalScrollbarComponent = null
+    @onVerticalScroll = null
+    @onHorizontalScroll = null
+
   getDomNode: ->
     @domNode
 
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -563,7 +563,7 @@ class TextEditorComponent
     @sampleBackgroundColors()
     @remeasureCharacterWidths()
 
-  handleDragUntilMouseUp: (dragHandler) =>
+  handleDragUntilMouseUp: (dragHandler) ->
     dragging = false
     lastMousePosition = {}
     animationLoop = =>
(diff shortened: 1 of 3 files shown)


src/text-editor-component.coffee
```diff
@@ -119,6 +119,14 @@ class TextEditorComponent
     @gutterContainerComponent?.destroy()
     @domElementPool.clear()
 
+    @verticalScrollbarComponent.destroy()
+    @horizontalScrollbarComponent.destroy()
+
+    @verticalScrollbarComponent = null
+    @horizontalScrollbarComponent = null
+    @onVerticalScroll = null
+    @onHorizontalScroll = null
```
as-cii: These references needs to be cleared out in order to allow `TextEditorComponent` to be freed up. Not sure what's causing them to retain `TextEditorComponent` as, although there's a cycle between references in there, I'd expect V8 mark-and-sweep GC to handle this properly. Maybe a problem with Coffeescript itself or a bug in V8 GC?

nathansobo:
This seems fine to me, though it would like to find out why we're leaking the model to begin with. But I think this is actually a pretty good mitigating strategy in general... when something bulky like a text editor gets destroyed, null out all of its components. That way if anyone leaks a reference, the cost will be much lower... just a small wrapper object.

as-cii:
> though it would like to find out why we're leaking the model to begin with

Ugh, I guess my comment was stripped off by an unfortunate diff mismatch. (https://github.com/atom/atom/pull/9067#discussion_r41417558) Sorry about that.

By the way, my suspects point towards a problem with the GC or Coffeescript causing the cyclic reference not to be resolved correctly. What do you think, @nathansobo?

nathansobo:
Very skeptical of a bug in the garbage collector. I've always found a reason but it's usually hard.

PR #9070, opened 2015-10-07
Allow a panesActiveItem show a saved Dialog

billybonks:
Currently if you want to create a new file type eg:
- writes and reads over ssh
- writes and reads to a remote server
- writes and reads via ftp

This functionality can be elegantly achieved by:
-inheriting from File from pathwatcher and over-riding read/write methods and nativeEvent Listeners.
-making a few minor changes to an extended text-buffer.

The only way of providing saveAs functionality is by overriding the `activePane`. or shim `atom.showSaveDialogSync`, both are quite large changes to the atom for such a small requirement.

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -290,15 +290,17 @@ describe "Pane", ->
         describe "when the item has no uri", ->
           it "presents a save-as dialog, then saves the item with the given uri before removing and destroying it", ->
             itemURI = null
-
             showSaveDialog.andReturn("/selected/path")
             confirm.andReturn(0)
+            spyOn(pane, 'saveItemAs')
             pane.destroyItem(item1)
-
-            expect(showSaveDialog).toHaveBeenCalled()
-            expect(item1.saveAs).toHaveBeenCalledWith("/selected/path")
-            expect(item1 in pane.getItems()).toBe false
-            expect(item1.isDestroyed()).toBe true
+            setTimeout( ->
+              expect(showSaveDialog).toHaveBeenCalled()
+              expect(pane.saveItemAs).toHaveBeenCalled()
+              expect(item1.saveAs).toHaveBeenCalledWith("/selected/path")
+              expect(item1 in pane.getItems()).toBe false
+              expect(item1.isDestroyed()).toBe true
+            , 500)
 
       describe "if the [Don't Save] option is selected", ->
         it "removes and destroys the item without saving it", ->
@@ -384,6 +386,27 @@ describe "Pane", ->
       pane.destroyInactiveItems()
       expect(pane.getItems()).toEqual [item2]
 
+  describe "::showSaveDialog()", ->
+    pane = null
+    item =  null
+
+    beforeEach ->
+      item = new Item("A")
+      pane = new Pane(items: [item], applicationDelegate: atom.applicationDelegate)
+      atom.applicationDelegate.showSaveDialog = jasmine.createSpy("atom.showSaveDialog")
+
+    describe "when the item implements a showSaveDialog method", ->
+      it "calls item.showSaveDialog", ->
+        item.showSaveDialog = jasmine.createSpy("item.showSaveDialog")
+        pane.showSaveDialog(item)
+        expect(item.showSaveDialog).toHaveBeenCalled()
+        expect(atom.applicationDelegate.showSaveDialog.callCount).toEqual(0)
+
+    describe "when the item doesn't implement a showSaveDialog method", ->
+      it "calls atom.showSaveDialog", ->
+        pane.showSaveDialog(item)
+        expect(atom.applicationDelegate.showSaveDialog).toHaveBeenCalled()
+
   describe "::saveActiveItem()", ->
     pane = null
 
@@ -408,9 +431,12 @@ describe "Pane", ->
 
     describe "when the current item has no uri", ->
       describe "when the current item has a saveAs method", ->
-        it "opens a save dialog and saves the current item as the selected path", ->
+        beforeEach ->
           pane.getActiveItem().saveAs = jasmine.createSpy("saveAs")
-          pane.saveActiveItem()
+          waitsForPromise ->
+            pane.saveActiveItem()
+
+        it "opens a save dialog and saves the current item as the selected path", ->
           expect(showSaveDialog).toHaveBeenCalled()
           expect(pane.getActiveItem().saveAs).toHaveBeenCalledWith('/selected/path')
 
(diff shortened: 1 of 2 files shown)


thedaniel:
cc @atom/feedback

src/pane.coffee
```diff
@@ -514,7 +514,8 @@ class Pane extends Model
 
     saveOptions = item.getSaveDialogOptions?() ? {}
     saveOptions.defaultPath ?= item.getPath()
-    newItemPath = atom.showSaveDialogSync(saveOptions)
+    showSaveDialogSync = if item.showSaveDialogSync else atom.showSaveDialogSync
```
benogle: Does this even work? Wouldn't it be `item.showSaveDialogSync ? atom.showSaveDialogSync`. Also the `this` would be `window` for this call. Not ideal.

benogle:
cc @nathansobo is this something we want to add?

src/pane.coffee
nathansobo: Yeah, this line of code doesn't look legit to me. It would have to use the `?` operator as you said. Also it would need to be bound to the item...

```coffee
showSaveDialogSync = item.showSaveDialogSync?.bind(item) ? atom.showSaveDialogSync.bind(atom)
```

Also, we'd need tests for this. I'd be open to adding a point of flexibility here but we need tests.

billybonks: Sorry for the rubbish code, wanted to know if the change would be welcomed.
I'll add tests and documentation.

billybonks:
not sure what the test naming conventions are. Let me know if i got it wrong :smile:

nathansobo:
Sorry, I should have thought of this sooner, but I'm a bit worried to have a synchronous design on this. Presumably you'll want to be able to show save dialogs that are rendered in HTML? If we do things synchronously, we won't be able to present it in a modal way. What's your use case for a synchronous API?

billybonks:
@nathansobo good catch. i'm building my package on this branch so i don't miss anything else :smile:. I will make the required changes

spec/pane-spec.coffee
```diff
@@ -281,15 +281,17 @@ describe "Pane", ->
         describe "when the item has no uri", ->
           it "presents a save-as dialog, then saves the item with the given uri before removing and destroying it", ->
             itemURI = null
-
             showSaveDialog.andReturn("/selected/path")
```
nathansobo: Doesn't this spy need to return a promise now?

nathansobo: Oh nevermind, it's a different method.

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+  describe "::showSaveDialog()", ->
+    pane = null
+    item =  null
+
+    beforeEach ->
+      item = new Item("A")
+      pane = new Pane(items: [item])
+      atom.showSaveDialogSync = jasmine.createSpy("item.showSaveDialog");
+
+    describe "item implements ::showSaveDialog", ->
```
nathansobo: `"when the item implements a showSaveDialog method"`

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+      it "calls item::showSaveDialog", ->
```
nathansobo: `"calls showSaveDialog and saves the item with the results of the returned promise"`

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+        item.showSaveDialog = jasmine.createSpy("item.showSaveDialog");
```
nathansobo: Please remove these semicolons

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+        pane.showSaveDialog(item)
+        expect(item.showSaveDialog).toHaveBeenCalled();
+        expect(atom.showSaveDialogSync.callCount).toEqual(0)
+
+    describe "item doesn't implement ::showSaveDialogSync", ->
```
nathansobo: `"when the item doesn't implement a showSaveDialog method"`

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+      it "calls atom::showSaveDialog", ->
```
nathansobo: `atom.showSaveDialog` (the `::` is for prototype methods)

nathansobo:
We need to document all the optional interface methods on pane items, including this one. We could deal with that in a separate PR though.

spec/pane-spec.coffee
```diff
@@ -281,15 +281,17 @@ describe "Pane", ->
...
             showSaveDialog.andReturn("/selected/path")
             confirm.andReturn(0)
+            spyOn(pane, 'saveItemAs')
             pane.destroyItem(item1)
-
-            expect(showSaveDialog).toHaveBeenCalled()
-            expect(item1.saveAs).toHaveBeenCalledWith("/selected/path")
-            expect(item1 in pane.getItems()).toBe false
-            expect(item1.isDestroyed()).toBe true
+            setTimeout( ->
```
billybonks: the only way to improve this spec is having, `destroyItem`, return a promise, that resolves when the item has been destroyed. or emit a `did-destroy-item` event

nathansobo: At the very least you can do a Jasmine-style `waitsFor "item to be destroyed", -> item1.isDestroy()` to hold the test back until that condition is satisfied. Or do an `onDidDestroy` on the item and a `waitsFor` block that takes a parameter.

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+
+    describe "when the item implements a showSaveDialog method", ->
+      it "calls item.showSaveDialog and saves the item with the results of the returned promise", ->
+        item.showSaveDialog = jasmine.createSpy("item.showSaveDialog")
+        pane.showSaveDialog(item)
+        expect(item.showSaveDialog).toHaveBeenCalled()
+        expect(atom.applicationDelegate.showSaveDialog.callCount).toEqual(0)
+
+    describe "when the item doesn't implement a showSaveDialog method", ->
+      it "calls atom.showSaveDialog and saves the item with the results of the returned promise", ->
```
nathansobo: You're describing more behavior than is actually tested here. Should fix one or the other.

spec/pane-spec.coffee
```diff
@@ -375,6 +377,27 @@ describe "Pane", ->
...
+    pane = null
+    item =  null
+
+    beforeEach ->
+      item = new Item("A")
+      pane = new Pane(items: [item], applicationDelegate: atom.applicationDelegate)
+      atom.applicationDelegate.showSaveDialog = jasmine.createSpy("atom.showSaveDialog")
+
+    describe "when the item implements a showSaveDialog method", ->
+      it "calls item.showSaveDialog and saves the item with the results of the returned promise", ->
```
nathansobo: Same here, let's show it saving.

src/pane.coffee
```diff
@@ -513,15 +514,29 @@ class Pane extends Model
   saveItemAs: (item, nextAction) ->
     return unless item?.saveAs?
 
+    @showSaveDialog(item).then( (newItemPath) =>
```
nathansobo: :fire: extra space

nathansobo:
Ping me again when you've addressed my comments.

benogle:
@billybonks have all the changes been made?

billybonks:
@benogle  sorry for the late reply. i still need to change one more test. i am having trouble rebuilding atom. i know it uses apm, but i upgraded npm to v3 and then probably upgrade node-gyp, which may be causing this error

screen shot 2016-02-15 at 11 19 33 am

billybonks:
going to close this for now, i don't need the api at the moment and i am sure the code has changed quite a bit.

PR #9073, opened 2015-10-07
Bug in moving selected lines down

Aakash1312:
There was a bug in moving lines down. 
The bug can be reproduced in the following steps:
- Make a text file with last line blank
- Select the entire row of third last line
- Try moving down. It won't move

output

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -849,7 +849,7 @@ class TextEditor extends Model
     selection = @getSelectedBufferRange()
     lastRow = @buffer.getLastRow()
     return if selection.end.row is lastRow
-    return if selection.end.row is lastRow - 1 and @buffer.getLastLine() is ''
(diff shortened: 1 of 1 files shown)


nathansobo:
Hi @Aakash1312... after testing this out it's not clear how the behavior differs. Can you post 2 animations, 1 with the original behavior and one with the change applied.

Aakash1312:
@nathansobo 
Here is the gif without change:
output2

and here is the fix
output

as you can see, you cannot move the selected line down beyond line number 7 in the first part.

benogle:
Is this still an issue @Aakash1312? If so, can you rebase on master and write a spec?

benogle:
I cannot reproduce at this time. Closing. @Aakash1312 if you can reproduce it still, feel free to reopen.

PR #9096, opened 2015-10-09
Add descriptions for config settings

izuzak:
This adds descriptions for core and editor config settings. I think this will help users understand the purpose of config settings and use them properly. I've definitely seen confusion about what some settings mean and I've been confused about some as well. Once this is merged, I'll start working on clarifying settings for individual packages.

@atom/feedback I'd love your help in reviewing this and making these descriptions as clear and as helpful as possible. Please let me know if you have any wordsmithing advice. :pray: 

Before:

screen shot 2015-10-09 at 12 58 22

After

screen shot 2015-10-09 at 12 57 27

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -97,7 +97,7 @@ module.exports =
           'windows866'
         ]
       openEmptyEditorOnStart:
-        description: 'Automatically open an empty editor when on startup.'
+        description: 'Automatically open an empty editor on startup.'
         type: 'boolean'
         default: true
 
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -147,7 +147,6 @@ module.exports =
         type: 'boolean'
         default: true
         description: 'Automatically indent the cursor when inserting a newline.'
-        type: 'boolean'
       autoIndentOnPaste:
         type: 'boolean'
         default: true
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -177,7 +177,7 @@ module.exports =
         type: 'string'
         default: 'auto'
         enum: ['auto', 'soft', 'hard']
-        description: 'Determine character inserted when Tab key is pressed. Possible values: "auto", "soft" and "hard". When set to "soft" or "hard", soft tabs (spaces) or hard tabs (tab characters) are used. When set to "auto", the editor auto-detects the tab type based on the contents of the buffer, or uses the value of the Soft Tabs config setting if auto-detection fails.'
+        description: 'Determine character inserted when Tab key is pressed. Possible values: "auto", "soft" and "hard". When set to "soft" or "hard", soft tabs (spaces) or hard tabs (tab characters) are used. When set to "auto", the editor auto-detects the tab type based on the contents of the buffer (it uses the first leading whitespace on a non-comment line), or uses the value of the Soft Tabs config setting if auto-detection fails.'
       softWrapAtPreferredLineLength:
         type: 'boolean'
         default: false
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -17,7 +17,7 @@ module.exports =
         type: 'boolean'
         default: true
         title: 'Exclude VCS Ignored Paths'
-        description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, project\'s using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
+        description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, projects using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
       followSymlinks:
         type: 'boolean'
         default: true
@@ -177,7 +177,7 @@ module.exports =
         type: 'string'
         default: 'auto'
         enum: ['auto', 'soft', 'hard']
-        description: 'Determine character inserted when Tab key is pressed. Possible values: "auto", "soft" and "hard". When set to "soft" or "hard", soft tabs (spaces) or hard tabs (tab characters) are used. When set to "auto", the editor auto-detects the tab type based on the contents of the buffer, or uses the value of the Soft Tabs config setting if auto-dection fails.'
+        description: 'Determine character inserted when Tab key is pressed. Possible values: "auto", "soft" and "hard". When set to "soft" or "hard", soft tabs (spaces) or hard tabs (tab characters) are used. When set to "auto", the editor auto-detects the tab type based on the contents of the buffer, or uses the value of the Soft Tabs config setting if auto-detection fails.'
       softWrapAtPreferredLineLength:
         type: 'boolean'
         default: false
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -12,24 +12,27 @@ module.exports =
         default: [".git", ".hg", ".svn", ".DS_Store", "._*", "Thumbs.db"]
         items:
           type: 'string'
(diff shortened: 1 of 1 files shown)


src/config-schema.coffee
```diff
@@ -12,24 +12,27 @@ module.exports =
         default: [".git", ".hg", ".svn", ".DS_Store", "._*", "Thumbs.db"]
         items:
           type: 'string'
+        description: 'List of string glob patterns. Files and directories matching these patterns will be ignored by some packages, such as the fuzzy finder and tree view. Individual packages might have additional config settings for ignoring names.'
       excludeVcsIgnoredPaths:
         type: 'boolean'
         default: true
         title: 'Exclude VCS Ignored Paths'
+        description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, project\'s using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
```
50Wliu: :art: second `project's` should just be `projects`

src/config-schema.coffee
```diff
@@ -12,24 +12,27 @@ module.exports =
...
       followSymlinks:
         type: 'boolean'
         default: true
         title: 'Follow symlinks'
-        description: 'Used when searching and when opening files with the fuzzy finder.'
+        description: 'Follow symbolic links when searching files and when opening files with the fuzzy finder.'
```
50Wliu: when searching and opening files?

src/config-schema.coffee
```diff
@@ -155,32 +173,36 @@ module.exports =
       softTabs:
         type: 'boolean'
         default: true
+        description: 'If the `Tab Type` config setting is set to "auto" and autodetection of tab type from buffer content fails, then this config setting determines whether a soft tab or a hard tab will be inserted when the Tab key is pressed.'
       tabType:
         type: 'string'
         default: 'auto'
         enum: ['auto', 'soft', 'hard']
-        description: 'Determine character inserted during Tab keypress.'
+        description: 'Determine character inserted when Tab key is pressed. Possible values: "auto", "soft" and "hard". When set to "soft" or "hard", soft tabs (spaces) or hard tabs (tab characters) are used. When set to "auto", the editor auto-detects the tab type based on the contents of the buffer, or uses the value of the Soft Tabs config setting if auto-dection fails.'
```
50Wliu: `auto-dection` -> `auto-detection`

src/config-schema.coffee
```diff
@@ -155,32 +172,36 @@ module.exports =
       softTabs:
         type: 'boolean'
         default: true
+        description: 'If the `Tab Type` config setting is set to "auto" and autodetection of tab type from buffer content fails, then this config setting determines whether a soft tab or a hard tab will be inserted when the Tab key is pressed.'
```
benogle: This is really only for a new file. Auto will autodetect, and fall back to this for new files

benogle:
Other than clarification around the tabType thing, I think it looks good

src/config-schema.coffee
izuzak: > This is really only for a new file. Auto will autodetect, and fall back to this for new files

Thanks for mentioning this, @benogle -- I think that config settings related to soft tabs are among the ones causing most confusion so I'd like the descriptions to be :gem: clear.

Can you clarify what _exactly_ "only for a new file" means? Does this mean new files created in Atom, files opened in Atom for the first time (I'm _guessing_ it's this), or something else? In other words, in which situations is the description above not true?

lee-dohm: I believe what it means is a file that is empty, i.e. has no content at all.

benogle: Yeah, if there's no content, the auto detect doesnt work, and will fall back to this setting.

izuzak: > Yeah, if there's no content, the auto detect doesnt work, and will fall back to this setting.

@benogle Just to make sure I'm following along -- can you share a link to the code where this behavior is implemented? Here's what I'm seeing:

1. Auto-detect kicks in [here](https://github.com/atom/atom/blob/3636eb35d8a7dbdc00cfbdaba7280c26abb08ca2/src/text-editor.coffee#L2387-L2388) if the tabType option is set to "auto". If auto-detect fails, then the defaultValue is used (what's this?) or the value of the softTabs config setting.
2. Auto-detect is implemented [here](https://github.com/atom/atom/blob/3636eb35d8a7dbdc00cfbdaba7280c26abb08ca2/src/text-editor.coffee#L2356-L2364) and fails for buffers which have no non-comment lines with leading whitespace (not necessarily an empty file).

I'm not sure what this defaultValue is, at which moments is the tab type of a buffer determined (When the file is first opened in Atom? Every time it's opened? Every time it's changed?), and how does the ability to manually change the tab type for a single buffer come affect everything (using the `Editor: Toggle Soft Tabs` command)?

benogle: We could change it to:

> Uses the value of this setting to determine whether to insert a soft or hard tab in an empty file when the `Tab Type` config setting is set to "auto".

izuzak: @benogle We could, but is an empty file the only time when this option will be chosen? I think it's not -- it's chosen any time auto-detect fails, and it can fail even when the file is non-empty, as far as I can tell (see my comment above).

benogle: So detection would happen everytime a file opened. IIRC, it works like this:

* Editor created :tada: 
* `defaultValue` is null at first, either no content or no content with leading whitespace -> fall back to `editor.softTabs`
* `@softTabs` is set the result of `@shouldUseSoftTabs(defaultValue: null)`
* When any soft tabs related config is changed, or tokenization happens, and calls `@shouldUseSoftTabs(defaultValue: @softTabs)`; so the `defaultValue` is then whatever the previous value was

`Editor: Toggle Soft Tabs` would straight up override all the settings for that editor. If it was determined to be hard tabs by auto detect, then toggle makes them soft no matter how the original detection happened.

Long story short: It is only used when a file is opened on files that are either empty or have no leading whitespace on a non-comment line. We should probably rename it to something like `fallbackTabType`.

thomasjo: > We should probably rename it to something like fallbackTabType

This. :100:

izuzak: Great -- thanks for explaining that flow, @benogle. :bow: I just pushed a clarification for how auto-detection works, and going to :boat: now.

izuzak:
I've opened PRs to update add/update descriptions for package config settings:

* [bracket-matcher](https://github.com/atom/bracket-matcher/pull/181) 
* [find-and-replace](https://github.com/atom/find-and-replace/pull/561)
* [fuzzy-finder](https://github.com/atom/fuzzy-finder/pull/145)
* [git-diff](https://github.com/atom/git-diff/pull/76)
* [grammar-selector](https://github.com/atom/grammar-selector/pull/25)
* [markdown-preview](https://github.com/atom/markdown-preview/pull/317)
* [notifications](https://github.com/atom/notifications/pull/96)
* [open-on-github](https://github.com/atom/open-on-github/pull/50)
* [package-generator](https://github.com/atom/package-generator/pull/32)
* [spell-check](https://github.com/atom/spell-check/pull/89)
* [symbols-view](https://github.com/atom/symbols-view/pull/126)
* [tabs](https://github.com/atom/tabs/pull/220)
* [tree-view](https://github.com/atom/tree-view/pull/607)
* [welcome](https://github.com/atom/welcome/pull/40)
* [whitespace](https://github.com/atom/whitespace/pull/95)

@atom/feedback I'll leave the PRs open for at least a few days -- would appreciate any feedback on the text, to make sure I didn't misunderstand the purpose of those config settings and that the text is understandable. :pray: Thank you! :purple_heart:

izuzak:
> I've opened PRs to update add/update descriptions for package config settings:

I'll probably merge these PRs at the end of the day or tomorrow. If anyone fancies reading long descriptions :book: of configuration settings during their lunch break :hamburger:, let me know if you have any feedback on these :point_up:. :smile_cat:

PR #9101, opened 2015-10-09
Fix #4300 by preventing 'will-navigate' event from firing

### LINKED ISSUE #4300: Dragging and dropping a file from the file system occasionally opens it as Chromium instead of Atom
Atom's drag&drop feature seem quite buggy. Sometimes it just won't open a dragged file, but even worse bug is when decides to open the file as if it was a real browser, efectively replacing the entire editor with the content of the file.

Video: http://youtu.be/EPRRFs9SQbs

thomasjo:
According to https://github.com/atom/atom/issues/4300#issuecomment-139103318 preventing the `will-navigate` event from firing/propagating will resolve the file dragging and dropping problem reported in #4300.

I've been unable to reproduce the problem on OS X, so I'm unsure if this actually fixes it or not. Will test on a VM in the morning (UTC) unless someone can verify that this fixes the problem.

In addition I wasn't sure whether we wanted to add the event subscription in `atom-window.coffee` or `atom-application.coffee`. 

/cc @atom/feedback 

---

Fixes #4300 (**unverified**)

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -142,6 +142,8 @@ class AtomWindow
         when 0 then @browserWindow.destroy()
         when 1 then @browserWindow.restart()
 
+    @browserWindow.webContents.on 'will-navigate', (event) -> event.preventDefault()
+
     @setupContextMenu()
 
     if @isSpec


maxbrunsfeld:
:clap:

zcbenz:
:+1:

kevinsawicki:
Thanks for this, I think this should definitely fix the issue.

benogle:
Thanks for taking this on @thomasjo

thomasjo:
> Thanks for taking this on @thomasjo

My pleasure :bow:

Issue #7546, opened 2015-06-29
Ctrl+<arrow> move of line should update its indentation


NDavidBrown:
Being able to move a line with Ctrl+arrow is very useful.

I suggest this feature be improved so the line matches the indentation in its new scope, e.g. being "unindented" if moving out of a FOR loop.

batjko:
:+1: Makes total sense to me.

eddiemonge:
merged in https://github.com/atom/atom/pull/9104 but perhaps not the best solution

50Wliu:
Ok, I'm going to close this then.  If the above PR is reverted then I'll re-open this.

PR #9104, opened 2015-10-10
autoindent lines with moveLineUp/moveLineDown

joseramonc:
Related to #7546, added the `autoIndentSelection if should autoIndent`. A demo of the change:

autoindentmoveline

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -4145,6 +4145,37 @@ describe "TextEditor", ->
       """
       expect(editor.getSelectedBufferRange()).toEqual [[13, 0], [14, 2]]
 
+  describe ".moveLineUp()", ->
+    it "moves line up", ->
+      editor.setCursorBufferPosition([1, 0])
+      editor.moveLineUp()
+      expect(editor.getTextInBufferRange([[0, 0], [0, 30]])).toBe "  var sort = function(items) {"
+      expect(editor.indentationForBufferRow(0)).toBe 1
+      expect(editor.indentationForBufferRow(1)).toBe 0
+      
+    it "update indentation when config editor.autoIndent is true", ->
+      atom.config.set('editor.autoIndent', true)
+      editor.setCursorBufferPosition([1, 0])
+      editor.moveLineUp()
+      expect(editor.indentationForBufferRow(0)).toBe 0
+      expect(editor.indentationForBufferRow(1)).toBe 0
+      
+  describe ".moveLineDown()", ->
+    it "moves line down", ->
+      editor.setCursorBufferPosition([0, 0])
+      editor.moveLineDown()
+      expect(editor.getTextInBufferRange([[1, 0], [1, 31]])).toBe "var quicksort = function () {"
+      expect(editor.indentationForBufferRow(0)).toBe 1
+      expect(editor.indentationForBufferRow(1)).toBe 0
+         
+    it "update indentation when config editor.autoIndent is true", ->
+      atom.config.set('editor.autoIndent', true)
+      editor.setCursorBufferPosition([0, 0])
+      editor.moveLineDown()
+      expect(editor.indentationForBufferRow(0)).toBe 1
+      expect(editor.indentationForBufferRow(1)).toBe 2
+      
+
   describe ".shouldPromptToSave()", ->
     it "returns false when an edit session's buffer is in use by more than one session", ->
       jasmine.unspy(editor, 'shouldPromptToSave')
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -842,6 +842,7 @@ class TextEditor extends Model
         @foldBufferRow(foldedRow)
 
       @setSelectedBufferRange(selection.translate([-insertDelta]), preserveFolds: true, autoscroll: true)
+      @autoIndentSelectedRows() if @shouldAutoIndent()
 
   # Move lines intersecting the most recent selection down by one row in screen
   # coordinates.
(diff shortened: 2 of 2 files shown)


YuriSolovyov:
Shouldn't there be an option in settings for that?

maxbrunsfeld:
> Shouldn't there be an option in settings for that?

It looks like it already respects the existing `core.autoIndent` scoped config option. It makes sense to me to use the same option as auto-indenting on newline insertion. Thoughts?

maxbrunsfeld:
Could you add a test for this @joseramonc?

YuriSolovyov:
@maxbrunsfeld if so, it would be nice to have this metioned in setting description

joseramonc:
@maxbrunsfeld Added tests for `editor.moveLineUp()` and `editor.moveLineDown()`, hope they are right, they were passing locally

maxbrunsfeld:
Sorry for letting this go quiet for a while; there were some network-related CI failures. I re-ran the build.

joseramonc:
@maxbrunsfeld It seems that just one build failed, is there some guide or help to debug this locally? Tests are passing on my computer :disappointed:

50Wliu:
You can safely ignore the symbols-view error; that spec is super-flaky right now.

maxbrunsfeld:
@joseramonc Thanks for the ping. Great work!

stefanbuck:
@joseramonc Thanks for this great feature, love it :tada:

dorian-marchal:
Thanks a lot for your contribution but this breaks a pretty common use case : https://discuss.atom.io/t/move-line-without-auto-indent/23754
Could we have an other option in settings to disable this specific behaviour ?

__Original message :__

> I don't want my lines to auto indent when I move them.

> How can I change this behaviour ?

> Example :

>

alexandernst:
Perhaps an option to disable auto-indent when moving lines in plain-text files?

dorian-marchal:
> Perhaps an option to disable auto-indent when moving lines in plain-text files?

Note that this is not specific to Markdown / Plain Text :

optimised

alexandernst:
@as-cii As per the gif @dorian-marchal just posted, I'd say auto-indent shouldn't trigger when changing lines from the same "lexical" line. Or make it understand what lexical line breaks are and make it work accordingly.

maxbrunsfeld:
OK, it looks like this feature seems to have undesirable consequences pretty often, even in contexts where auto-indent is desirable for text insertion.

@joseramonc I think we may need to remove this feature until Atom's auto-indentation system is good enough to avoid these problems.

snowsunny:
Just like @dorian-marchal has demonstrated this feature wreaks havoc on indentation for code spanning multiple lines. It would be nice to have a separate setting to enable or disable auto-indent specifically for this feature. EG: "editor.indentOnLineMove", when set to false it won't automatically indent but editor.autoIndent will remain untouched/active.

@dorian-marchal I kludged it with this custom command.
```coffee
atom.commands.add 'atom-text-editor', 'my:move-line-up', ->
  editor = atom.workspace.getActiveTextEditor()
  if atom.config.get('editor.autoIndent')
    atom.config.set('editor.autoIndent', false)
    editor.moveLineUp()
    atom.config.set('editor.autoIndent', true)
  else
    editor.moveLineUp()

atom.commands.add 'atom-text-editor', 'my:move-line-down', ->
  editor = atom.workspace.getActiveTextEditor()
  if atom.config.get('editor.autoIndent')
    atom.config.set('editor.autoIndent', false)
    editor.moveLineDown()
    atom.config.set('editor.autoIndent', true)
  else
    editor.moveLineDown()
```
and
```cson
'.platform-linux atom-workspace atom-text-editor:not([mini])':
  'ctrl-shift-up': 'my:move-line-up'
  'ctrl-shift-down': 'my:move-line-down'
```

dorian-marchal:
@snowsunny It works just fine, thank you :+1:

snowsunny:
@dorian-marchal I’m happy to help :+1:

jeancroy:
Maybe make this an option ? It seems useful for people that mostly use language with clear indent scoping block (like html, JavaScript)

I also feel that re-arranging item in hanging indent is somewhat rare.

mnquintana:
Making this an option sounds :cool: to me – that way we could scope it only to grammars that play nicely with this.

geoffkembel:
Adding another vote for making this a separate option from auto-indent. For someone who moves multiple lines regularly, this is a real killer.

In the meantime, thanks to @snowsunny for the workaround.

jeancroy:
In what kind of scenario auto-indent works well but this does not ?
I was under the impression that this only make auto-indent flaws more apparent.

eddiemonge:
Yes please for separating this into its own flag.

snowsunny:
@eddiemonge  Please try this keymap.
```cson
# for mac
'.platform-darwin atom-workspace atom-text-editor:not([mini])':
  'ctrl-cmd-up': 'my:move-line-up'
  'ctrl-cmd-down': 'my:move-line-down'
```

phwh:
Also, auto-indent while moving doesn't work well with Python code. Here is an example:

atom-indent-issue

radiac:
This can be pretty dangerous when writing python. For example, moving a line from above an `if` statement to underneath it makes the code conditional. If you don't notice the indentation change, it can take quite a while to figure out why your code has stopped working.

Definitely needs to be an option, and probably off by default in python - and I'd imagine any other language with significant whitespace. In the meantime, thanks @snowsunny - you are my internet hero of the day.

There is now a related issue where this is being discussed further: https://github.com/atom/atom/issues/10524

PR #9108, opened 2015-10-11
:bug: `BufferedProcess.kill` now fully kills any child-process-eption that happens

### LINKED ISSUE #7252: Problem with killing children processes
Hi there. I seem to be running into a problem using the atom-script editor, you can see it here: rgbkrk/atom-script#453

Basically, atom-script runs go, which in turn starts up a server. Killing the process with a control-c kills off the child process, but not the child of the child. 

```
13:53|~/go/src/github.com/davewv31/helloworld/helloworld% ps ax | grep helloworld
84187   ??  S      0:00.22 go run /Users/dave/go/src/github.com/davewv31/helloworld/helloworld/helloworld.go
84218   ??  S      0:00.22 /var/folders/2v/53vdc3j9187f6zj1ljjfc10h0000gn/T/go-build234219847/command-line-arguments/_obj/exe/helloworld
```

After control-c in atom:

```
13:53|~/go/src/github.com/davewv31/helloworld/helloworld% ps ax | grep helloworld
84218   ??  S      0:00.22 /var/folders/2v/53vdc3j9187f6zj1ljjfc10h0000gn/T/go-build234219847/command-line-arguments/_obj/exe/helloworld
```

In buffered-process.coffee, the windows and non-windows terminations seem to be different. Windows seems to kill children of children, non-windows seems to just kill the child. 

```
  killProcess: ->
    @process?.kill()
    @process = null
```

So I guess what this comes down to is what should the expected behavior of BufferedProcess.kill be? It does just wrap child_process, but doesn't allow a full process group kill (which I want) or a signal type to be sent(which I'm not sure would work). 

Thanks!

acburdine:
closes #7252
- fixes :bug: where child processes spawned by spawned child processes on *nix systems were not getting killed correctly

### DIFF
--- a/src/buffered-process.coffee
+++ b/src/buffered-process.coffee
@@ -2,6 +2,7 @@ _ = require 'underscore-plus'
 ChildProcess = require 'child_process'
 {Emitter} = require 'event-kit'
 path = require 'path'
+psTree = require 'ps-tree'
 
 # Extended: A wrapper which provides standard error/output line buffering for
 # Node's ChildProcess.
@@ -159,6 +160,20 @@ class BufferedProcess
           process.kill(pid)
       @killProcess()
 
+  # Kill all child processes spawned by a spawned child process
+  #
+  # Uses ps-tree, so it works on *nix systems only
+  killOnNix: ->
+    return unless @process?
+
+    parentPid = @process.pid
+    psTree parentPid, (err, children) =>
+      pidsToKill = children.map (p) -> p.PID
+      for pid in pidsToKill
+        try
+          process.kill(pid)
+      @killProcess()
+
   killProcess: ->
     @process?.kill()
     @process = null
@@ -187,7 +202,7 @@ class BufferedProcess
     if process.platform is 'win32'
       @killOnWindows()
     else
-      @killProcess()
+      @killOnNix()
 
     undefined
 
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "oniguruma": "^5",
     "pathwatcher": "^6.2",
     "property-accessors": "^1.1.3",
+    "ps-tree": "^1.0.1",
     "random-words": "0.0.1",
     "runas": "^3.1",
     "scandal": "^2.2",


acburdine:
[relevant article cred](http://krasimirtsonev.com/blog/article/Nodejs-managing-child-processes-starting-stopping-exec-spawn) for where this code comes from

src/buffered-process.coffee
```diff
@@ -159,6 +160,27 @@ class BufferedProcess
           process.kill(pid)
       @killProcess()
 
+  # Kill all child processes spawned by a spawned child process
+  #
+  # Uses ps-tree, so it works on *nix systems only
+  killOnNix: ->
+    return unless @process?
+
+    self = this
```
thomasjo: :fire: This line can be removed

src/buffered-process.coffee
```diff
@@ -159,6 +160,27 @@ class BufferedProcess
...
+    parentPid = @process.pid
+    psTree(parentPid, (err, children) ->
```
thomasjo: Use a fat arrow (`=>`) here to capture the scope of the function.

src/buffered-process.coffee
```diff
@@ -159,6 +160,27 @@ class BufferedProcess
...
+      [parentPid].concat(
+        children.map((p) ->
+          return p.PID
+        )
+      ).forEach((tpid) ->
```
thomasjo: Rewrite the pid block as 
```coffee
pidsToKill = psTree parentPid, (err, children) -> [parentPid].concat(children.map (p) -> p.PID)
```
or equivalent.

src/buffered-process.coffee
```diff
@@ -159,6 +160,27 @@ class BufferedProcess
...
+        try
+          process.kill(tpid)
+        catch ex
+      )
```
thomasjo: The `.forEach` block should be rewritten as
```coffee
for pid in pidsToKill
  try
    process.kill(pid)
```

src/buffered-process.coffee
```diff
@@ -159,6 +160,27 @@ class BufferedProcess
...
+      self.killProcess()
```
thomasjo: Make this `@killProcess()`.

src/buffered-process.coffee
thomasjo: On second thought, considering we're invoking `@killProcess`, we don't want the parent PID here, so we should be able to change this code to
```coffee
pidsToKill = psTree parentPid, (err, children) -> children.map (p) -> p.PID
```

acburdine:
@thomasjo It should be good to go now, checking to make sure changes still retain functionality

src/buffered-process.coffee
```diff
@@ -159,6 +160,21 @@ class BufferedProcess
...
+  # Uses ps-tree, so it works on *nix systems only
+  killOnNix: ->
+    return unless @process?
+
+    parentPid = @process.pid
+    pidsToKill = psTree parentPid, (err, children) -> children.map(p) -> p.PID
+
+    for pid in pidsToKill
+        try
+            process.kill(pid)
```
thomasjo: These two lines should be indented with 2 spaces, not 4.

src/buffered-process.coffee
acburdine: @thomasjo both of these seem to break the functionality introduced, as the array isn't returned by the (err, children) callback function

acburdine: Also, I'm pretty sure psTree is asynchronous, meaning that it won't work to set it to something outside the callback. The forEach has to be handled inside the callback

thomasjo:
> Also, I'm pretty sure psTree is asynchronous, meaning that it won't work to set it to something outside the callback. The forEach has to be handled inside the callback

Aha, that changes the lay of the land slightly. I guess this should work then
```coffee
parentPid = @process.pid
psTree parentPid, (err, children) =>
  pidsToKill = children.map(child) -> child.PID

  for pid in pidsToKill
    try
      process.kill(pid)
  @killProcess()
```

acburdine:
Got it done, tested it so it should be good to go!

src/buffered-process.coffee
```diff
@@ -159,6 +160,22 @@ class BufferedProcess
...
+  killOnNix: ->
+    return unless @process?
+
+    parentPid = @process.pid
+    psTree parentPid, (err, children) ->
+        pidsToKill = children.map (p) -> p.PID
+        for pid in pidsToKill
+          try
+            process.kill(pid)
+        return
```
thomasjo: You can remove this `return` statement.

src/buffered-process.coffee
```diff
@@ -159,6 +160,22 @@ class BufferedProcess
...
+  # Kill all child processes spawned by a spawned child process
+  #
+  # Uses ps-tree, so it works on *nix systems only
+  killOnNix: ->
+    return unless @process?
+
+    parentPid = @process.pid
+    psTree parentPid, (err, children) ->
+        pidsToKill = children.map (p) -> p.PID
+        for pid in pidsToKill
```
thomasjo: :art: The indentation here is wrong; it has to be 2 spaces, not 4.

src/buffered-process.coffee
```diff
@@ -159,6 +160,22 @@ class BufferedProcess
...
+          try
+            process.kill(pid)
+        return
+
+    @killProcess()
```
thomasjo: Given the fact the `psTree` call is async, I think this invocation of `@killProcess` should be nested inside the callback handler (as seen in https://github.com/atom/atom/pull/9108#issuecomment-147311776). This is to guard against any possible side-effects of killing the primary process first. What do you think?

acburdine: I would agree. I'll fix that

thomasjo: Remember to change the outermost arrow to a fat arrow to capture the scope, otherwise `@killProcess()` is going to target the wrong scope. 

Thanks for sticking with me on this!

thomasjo:
Great, the code looks good to me now. Although I haven't tested this yet, which brings me to the next point; it would be great if you could add a test for this to ensure we're protected against regressions in functionality.

Maybe we should add a separate `describe` block for `"on Linux and OS X"` (or some such name) along the lines of the `"on Windows"` block. And given the added functionality is dependent upon actually running on a *NIX system, the block should not run on Windows, so simply `return` if the platform is set to `win32`.

/cc @atom/feedback

acburdine:
@thomasjo to that effect, shouldn't the killOnWindows function also have a check to make sure it is windows?

acburdine:
also, I'm not 100% sure how to test for the behavior. I know the windows spec block tests to make sure the process is wrapped in a cmd.exe wrapper, but I'm not sure how to run a command that spawns another child process and keeps it running

thomasjo:
> to that effect, shouldn't the killOnWindows function also have a check to make sure it is windows?

I was specifically talking about specs in that context, so we are not changing e.g. `killOnWindows` or `killOnNix` to check for platforms. What we need to do though, is to ensure we don't try to run the specs covering `killOnNix` when we're on a Windows platform.

thomasjo:
If you don't feel up to the challenge with regards to adding specs for this — I realize testing this is not trivial — perhaps one of folks from @atom/non-github-maintainers (or someone else) would like to assist. If nobody hops on this, I'll try to find some time by the end of the week.

acburdine:
Makes sense. I'll try and get this finished up over the next couple of days.

acburdine:
@thomasjo I've racked my brain trying to figure out a way to test this out. I can't think of any system commands that would spawn their own child processes (except maybe a fork bomb, but that's probably not a good idea :wink:). I'm not sure exactly where to go from here.

thedaniel:
> I can't think of any system commands that would spawn their own child processes

probably [foreman](https://github.com/ddollar/foreman), or perhaps a web server that does concurrency via child processes?

acburdine:
@thedaniel yeah that would work, the only issue is that the test suite would have to install foreman just for doing testing. There are a lot of npm packages as well that could work to test this behavior out, but to add another package solely for testing this one behavior seems like overkill to me.

I'm willing to write a test using an npm library to generate the child processes, but only if it's deemed worth adding.

thedaniel:
ah, sorry, i just skimmed this and didn't get the full context.

ghost (approved):

1Uno:
Does this really work? This is just terminology ain't it. No killing

Issue #9116, opened 2015-10-12
Source maps are broken in the debugger on Windows


ForNeVeR:
1. Install the latest Atom on Windows.
2. Install any coffeescript package (for example, [color-picker](https://atom.io/packages/color-picker)).
3. Open the debugger (Ctrl-Alt-I by default).
4. Try to open any coffeescript file in the debugger. It will show the JS code instead of coffee (problem 1, see screenshot).
   
   ![image](https://cloud.githubusercontent.com/assets/92793/10421947/b22a5c7c-70d4-11e5-90a4-e0fa94961f38.png)

5. Try to set the breakpoint; it will not show the breakpoint and show the empty file instead (problem 2, see screenshot).
   ![image](https://cloud.githubusercontent.com/assets/92793/10421959/1686bbac-70d5-11e5-8051-b81e62eab78a.png)
6. Try to find the corresponding sources in the tree, it will show this (problem 3):
   ![image](https://cloud.githubusercontent.com/assets/92793/10421991/cb2e8bf2-70d5-11e5-8076-6662def4bf08.png)

   (that's weird because it places full path `C:\Users\USERNAME\etc\etc\*.coffee` inside of a `C:\Users\USERNAME`, but it should show only file name such as `file.coffee` **without path** there as shown in the next tree leaf)

There is a `//# sourceMappingURL` comment in the JS code shown on the first screenshot. I've decoded it from base64 and the relevant part seems to be the following:

```
  "sources": [
    "C:\\Users\\USERNAME\\.atom\\packages\\color-picker\\lib\\ColorPicker-view.coffee"
  ],
```

I am not sure but it seems that the code contributed in #3062 is no longer working. It seems that that contributed code is still there: https://github.com/atom/atom/blob/d9410716d2c39fa4d62b2e1ab2fe1251a5805487/src/compile-cache.js#L97 but maybe it somehow don't get applied in this case.

ForNeVeR:
It seems that maybe that hack should be also applied to `compile` function of `coffee-script` module: https://github.com/atom/atom/blob/fb2a45d71dcb2f9ac60be29089382f866ec3813f/src/coffee-script.js#L21

izuzak:
Closing since https://github.com/atom/atom/pull/9122 was merged.

PR #9122, opened 2015-10-12
Fix coffeescript source maps on Windows

ForNeVeR:
I've done some research of current code and PR #3062 and have found that we not only should convert the slashes in the path, but also prepend `file:///` to it.

After that source maps started to work on Windows for me. Please note that the similar code has been added to [`compile-cache` module](https://github.com/atom/atom/blob/d9410716d2c39fa4d62b2e1ab2fe1251a5805487/src/compile-cache.js#L96) but I'm not sure whether we should/want to generalize and reuse this (because it doesn't prepend `file:///` there and seems to be working fine).

This patch will fix issue #9116.

### DIFF
--- a/src/coffee-script.js
+++ b/src/coffee-script.js
@@ -29,6 +29,10 @@ exports.compile = function (sourceCode, filePath) {
     Error.prepareStackTrace = previousPrepareStackTrace
   }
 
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks so much for figuring this out!

Issue #9032, opened 2015-10-03
Warning in specs: Error setting item in object: Validation failed at editor.useShadowDOM...


benogle:
Running any package spec produces this warning:

![screen shot 2015-10-02 at 5 12 37 pm](https://cloud.githubusercontent.com/assets/69169/10260300/d747734a-6928-11e5-8e95-9fee152d3803.png)

Here's why it's happening:

* `config.set "editor.useShadowDOM", true` is used in the [spec-helper](https://github.com/atom/atom/blob/37503b4d03f3f0162738332ff9693641b9ffdd40/spec/spec-helper.coffee#L138)
* true is the default value for this setting, so [`value = undefined if _.isEqual(defaultValue, value)`](https://github.com/atom/atom/blob/37503b4d03f3f0162738332ff9693641b9ffdd40/src/config.coffee#L835) in config.coffee:setRawValue() is run
* At this point `atom.config.settings.editor.useShadowDOM` is actually set to undefined.
* When the package is loaded, it resets the config schema, and pushes `atom.config.settings` back through the schema validators. __But__ `atom.config.settings.editor.useShadowDOM` is `undefined`, which is not a valid value! So warning

Not sure what to do here. I assume the undefined value is supposed to remove the key from the settings, but I dont know the current state of things. @maxbrunsfeld @nathansobo any insight?

PR #9131, opened 2015-10-13
Fix setting default values in config

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/9032


as-cii:
Thanks for this, @maxbrunsfeld. :heart:

PR #9135, opened 2015-10-14
Update Node versions in the docs

winstliu:
Instead of specifying both node.js and io.js, just say "node.js 0.10.x or above" now that node.js and io.js have merged.


thomasjo:
The official name is actually *Node.js*, not *node.js*; would you mind reverting those changes and fixing the other changed descriptions where it was previously incorrect?

50Wliu:
> The official name is actually Node.js, not node.js

I knew I should have double-checked :grin:.  Fixed!

maxbrunsfeld:
:ship:

PR #9157, opened 2015-10-15
Remove stacktrace-parser

winstliu:
I couldn't find any part of the code that used this, so here goes!


nathansobo:
Looks like we use it in notifications, but have our own dependency there. Thanks!

PR #9162, opened 2015-10-15
CJK soft wrap

as-cii:
Instead of considering every character to have the same width when we perform soft-wrapping, we now make a distinction between default characters, double width characters, half width characters and Korean characters .

The strategy consists in measuring these kinds of characters once when sampling font styles and then using the calculated width to find where a certain line wraps.

Please, note that this won't work with non-monospaced fonts, but it's consistent with how we currently handle soft-wrapping for latin characters.


benogle:
OMG, if this fixes the 80% case, I :heart: :heart: you. :tada:

as-cii:
cjk

@benogle: :joy: :point_up_2: :point_up_2: :point_up_2: :heart: 

I had to treat Korean characters differently, as their width isn't the same as japanese and chinese. I couldn't find a good name to express this in code (as opposed to double width and half width), so I simply named them `koreanCharacter`s (I could have probably mentioned "Hangul" but as a non-native speaker it didn't feel super memorable).

Honestly I performed only super basic tests to check if this worked fine: probably a native speaker could give a more detailed feedback. I'll make sure to mention some of the people who actively participated in the CJK discussion if this looks good code-wise.

Thanks!

SaschaNaz:
Korean characters use spaces but usually they don't wrap by words. For example:

```
아이돌마스터 신데렐라걸즈 스타라이트   |
스테이지

아이돌마스터 신데렐라걸즈 스타라이트 스| <- This is usually preferred
테이지
```

Web browser behavior example: https://twitter.com/newstapa/status/654793987543306241

Zireael07:
<3 this is amazing for CJK users

as-cii:
@SaschaNaz: I wonder if, as a first step, it could be acceptable to wrap by words. How bad do you think that would be? The :cool: thing about this approach is that it fits quite nicely our current code structure: it's far from being perfect, but it _seems_ like it's way better than how we handle this scenario currently. What do you think? By the way, thanks for all the effort you have put into this. :heart:

SaschaNaz:
@as-cii I think there can be some more progress but I agree that this is far better than current behavior. Great work :+1:

maxbrunsfeld:
:ship:

nathansobo:
Thanks @as-cii for your continued excellence.

as-cii:
:heart:

benogle:
I feel like this should close #1783 Thoughts?

maxbrunsfeld:
Yeah, I agree. We can create more specific issues for whatever remaining problems we find.

kevinsawicki:
so awesome :fireworks:

ethanluoyc:
This is awesome

PR #9192, opened 2015-10-17
Fix issues with running headless tests

nathansobo:
This pull requests addresses a couple issues with #8968
- It fixes an issue where the `resourcePath` wasn't being correctly assigned for `devMode`, which required `--dev` to always be passed to avoid throwing an exception when loading bundled keymaps.
- It overrides `console.log` and friends in headless tests to log to the terminal, working around a change in Electron 33 that requires `--enable-logging` to be passed on the command line to see console output. This output is now always visible in headless tests.

/cc @bolinfest @thomasjo @as-cii


thomasjo:
Great, going to take this for a spin!

PR #9195, opened 2015-10-17
Assign defaults for applicationDelegate, document, window to AtomEnvironment

bolinfest:
I tested this using the new custom testrunner feature introduced in
https://github.com/atom/atom/pull/8968. My testrunner function was defined as follows:

``` js
'use babel';

module.exports = function({testPaths, buildAtomEnvironment, logFile, headless, legacyTestRunner}) {
  global.atom = buildAtomEnvironment();
};
```

Without this change, `buildAtomEnvironment()` threw a bunch of exceptions.

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -13,6 +13,7 @@ StorageFolder = require './storage-folder'
 {getWindowLoadSettings} = require './window-load-settings-helpers'
 registerDefaultCommands = require './register-default-commands'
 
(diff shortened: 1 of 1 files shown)


nathansobo:
Why not just pass these parameters in? We provide you with a `buildDefaultApplicationDelegate` method, and you can easily pass in a `window` and a `document`. I'd really prefer these things to be explicit.

PR #9198, opened 2015-10-17
permit any whole number for tabLength

### LINKED ISSUE #8261: Cannot easily specify variable Tab Length
Why does Atom feel the need to choose which tab lengths I'm allowed to use? So many other editors let you specify this value (XCode/ST2/ST3/TW/KE/VisualStudio/even Word). And it's useful for all kinds of random tasks: sometimes I need to view/edit a massive variable-length TSV, so, elsewhere, I just throw my tab length to 40/60/whatever and edit faster than Excel thanks to regexes and [column selection](#2306).

The Tab Length item in the Settings panel should be a field, like Font Size, Line Height, and every other reasonable option: not a dropdown.

### LINKED ISSUE #9147: Long tab lengths fail, and default back to 4
Manually setting the tab length beyond 8, say to `16` or `40`, causes the tab length to become 4. This is true both when set from the user's config file and when specified in the settings of a package. Long tabs are useful for spreadsheet editing, given that neither Excel nor Numbers support multiple cursors or regex-find-replace.

This is indirectly related related to #8261. I can't fathom why Atom moralizes on tab length. Sure, 4-length soft tabs are superior, but there are legitimate reasons for using other permutations.

alexchandel:
Fixes #8261 
Fixes #9147

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -139,6 +139,15 @@ describe "TextEditor", ->
         expect(editor2.getSoftTabs()).toBe true
         expect(editor2.getEncoding()).toBe 'macroman'
 
+        atom.config.set('editor.tabLength', -1)
+        expect(editor2.getTabLength()).toBe 1
+        atom.config.set('editor.tabLength', 2)
+        expect(editor2.getTabLength()).toBe 2
+        atom.config.set('editor.tabLength', 17)
+        expect(editor2.getTabLength()).toBe 17
+        atom.config.set('editor.tabLength', 128)
+        expect(editor2.getTabLength()).toBe 128
+
     it "uses scoped `core.fileEncoding` values", ->
       editor1 = null
       editor2 = null
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -171,7 +171,7 @@ module.exports =
       tabLength:
         type: 'integer'
         default: 2
-        enum: [1, 2, 3, 4, 6, 8]
+        minimum: 1
         description: 'Number of spaces used to represent a tab.'
       softWrap:
         type: 'boolean'


50Wliu:
Do you think there should be a maximum for this?  Would setting a tab length of eg 2000 have any negative performance effects?

alexchandel:
@50Wliu no, [there shouldn't be a maximum](https://en.wikipedia.org/wiki/Zero_one_infinity_rule). There isn't for any other parameter, including line height, and I just tested a tab length of 2000 in ST3 with no issues. Specs?

benogle:
IMO, there should be a max of something like 100 or 200. 

> Specs?

Yes, please.

50Wliu:
For context, specs: https://github.com/atom/atom/blob/master/CONTRIBUTING.md#specs-styleguide

alexchandel:
@benogle the thing is, there are always use cases where artificial caps are hit. For example, last week I had to do some text work a massive TSV with 512 length tabs; horizontal selection & regexs in ST made the task easy. I'm probably unusual for eschewing clunky spreadsheet programs for text editors, but I'm sure there are other situations. I just don't understand. Why is it so necessary to artificially limit tabs? Just give users the freedom to push the editor.

@50Wliu Could you give me an example of what I should edit in the spec folder? I only see four references to `tabLength` in `spec/config-spec.coffee`.

benogle:
> Why is it so necessary to artificially limit tabs?

Because 99.99% of the time, a huge tab length like that is not what people want. If they did enter something that large, it will have been in error. So I'd rather protect the majority of users than allow this strange edge case.

50Wliu:
I say that if there's no performance decrease then it might not be necessary to add an upper limit.  I mostly asked in the first place because I remember one issue where someone accidentally made their font size 20000 or something which froze Atom.

benogle:
We can merge this when a spec has been written.

alexchandel:
@benogle Sure. However I don't understand what that entails. Afaik there weren't any specs to begin with. What file should I change?

benogle:
You should add a spec in [text-editor-spec.coffee](https://github.com/atom/atom/blob/master/spec/text-editor-spec.coffee)

alexchandel:
@benogle So around [https://github.com/atom/atom/blob/master/spec/text-editor-spec.coffee#L107](text-editor-spec.coffee#L107), change the 4->8 `tabLength` test to test a larger range of `tabLength`s instead ?

benogle:
Yep, that sounds good.

alexchandel:
So, from L124, something like this?
```coffee
        atom.config.set('editor.tabLength', 8)
        atom.config.set('editor.softWrap', false)
        atom.config.set('editor.softTabs', true)
        atom.config.set('core.fileEncoding', 'macroman')

      waitsForPromise ->
        atom.workspace.open('dir/b').then (o) -> editor2 = o

      runs ->
        expect(editor2.getTabLength()).toBe 8
        expect(editor2.isSoftWrapped()).toBe false
        expect(editor2.getSoftTabs()).toBe true
        expect(editor2.getEncoding()).toBe 'macroman'
        for tab in [5..128]
          atom.config.set('editor.tabLength', tab)
          expect(editor2.getTabLength()).toBe tab
```

50Wliu:
@alexchandel I think testing every possibility from 5 to 128 seems a bit excessive...how about trying -1 (which should fail and set it to 1), 2, and maybe one or two other numbers?

alexchandel:
@50Wliu Ok like this? And it's ok to leave it in the same `runs` block?
```coffee
        atom.config.set('editor.tabLength', 8)
        atom.config.set('editor.softWrap', false)
        atom.config.set('editor.softTabs', true)
        atom.config.set('core.fileEncoding', 'macroman')

      waitsForPromise ->
        atom.workspace.open('dir/b').then (o) -> editor2 = o

      runs ->
        expect(editor2.getTabLength()).toBe 8
        expect(editor2.isSoftWrapped()).toBe false
        expect(editor2.getSoftTabs()).toBe true
        expect(editor2.getEncoding()).toBe 'macroman'
        atom.config.set('editor.tabLength', -1)
        expect(editor2.getTabLength()).toBe 1
        atom.config.set('editor.tabLength', 2)
        expect(editor2.getTabLength()).toBe 2
        atom.config.set('editor.tabLength', 17)
        expect(editor2.getTabLength()).toBe 17
        atom.config.set('editor.tabLength', 128)
        expect(editor2.getTabLength()).toBe 128
```

50Wliu:
Maybe separate it out into a different test.  @benogle?

alexchandel:
Would that involve just another
```coffee
waitsForPromise ->
        atom.workspace.open('dir/b').then (o) -> editor2 = o
```
except with `dir/c`?

benogle:
I think adding a couple lines to the current test is fine

alexchandel:
Tests pass after rebase.

benogle:
Thanks! It will be in 1.7. We only do patch releases for regressions and major issues.

alexchandel:
@benogle Since 1.6.0 is still in beta, could it make it there?

benogle:
No, sorry. 1.7 will be in beta next week.

Issue #2456, opened 2014-05-30
Atom over sshfs Unusably Slow


martindale:
When editing files over an sshfs connection, it takes long periods of time to perform basic actions (directory expansion, file opening).  This is likely because git is running locally, but querying a remote repository many times -- can these behaviors be migrated into a non-blocking request?  e.g., allow the file to open before the git repo has been read for any changes.

zuphu:
Problem still exists, I'd like to see this fixed :+1:.

ysle:
+1

Nikita240:
+1

macbury:
this flags for sshfs make it less laggy:
sshfs macburybury@localhost:/home/macbury/my_app ~/mounts/my_app/ -oauto_cache,reconnect,Ciphers=arcfour

mamoit:
:+1: Unusably slow for me.
Had to go back to sublime.
Running on Debian Jessie.

Nikita240:
@macbury Those flags did help with the performance a bit. Thanks!

conormce:
Had the same issues as everyone here, went into packages and disabled all git related installed packages. (git-log, git-diff-popup, git-history, git-tab-status). For now it at least makes it possible to work in atom with sshfs. Also this only happens when using a VPN for me

ypresto:
Plain git command on sshfs is very slow too.
Perhaps this is because git accesses many files in .git simultaneously.
My idea is proxying [git-utils](https://github.com/atom/git-utils) methods to remote (micro) atom instance over ssh or sockets.

mosbth:
As @Nikita240 says, the flags provided by @macbury made it a bit better. It got from unusable to workable. It still hangs on occasions though, for a few seconds.

One place it hangs is where I try to mark text, using the mouse, by double-clicking. It does not always hang, but quite often when a switch between applications and comes back to atom.

I'm running on Debian Sid. The filesystem I mounted is quite large, I guess more than 1000 files. Its a smaller part, of that filesystem, that I open i Atom, less than 100 files in three subdirectories.

More sshfs-flags to try? Any settings in Atom I might change to improve performance?

ypresto:
Now I'm using [unison](http://www.cis.upenn.edu/~bcpierce/unison/) instead of sshfs..!

https://discuss.atom.io/t/working-with-ssh/1737/34?u=ypresto

willwhitney:
[...]

PR #9213, opened 2015-10-19
Async Git Repository class

thedaniel:
This is an experimental spike of an alternate GitRepository class that uses [nodegit](http://www.nodegit.org/) rather than [git-utils](https://github.com/atom/git-utils). We would love to speed up the responsiveness of various aspects of Atom (tree-view, etc) by doing more Git operations asynchronously, and it would also be nice to get out of the business of maintaining our own libgit2 wrapper.

Here are a few notes about the reasoning behind some decisions here:
- It follows the existing GitRepository API closely, changing everything sync to Promise-based async.
  - The assumption is that having a close mapping between the sync and aysnc APIs will make it less painful to refactor client packages.
  - I'm not 100% sure we should keep this design, so it's a tradeoff between design quality and incurred labor. Ideally we'll keep this private and experimental all the way until Atom 2.0 so that we can change it as we need to.
- The class is written in ES6, because Coffeescript, while lovely, is the past, and the community seems to have consensus on transpilers like babel.
  - This includes an .eslintrc, because I've found it extremely useful.
  - The spec is still in coffee because I thought it might be easier to just modify it as I go to make sure I had the same coverage on both classes, though at the moment I wish I'd run it through a coffee -> js converter first.
- I haven't done a lot of ES6 or constructed a Promise-based API for anything serious before, so I'm sure to misuse some APIs.
- We'll need to keep the sync GitRepository class around for backwards compat, so I haven't decided quite how to integrate this with the current provider system.
  - Perhaps we add a private method on the sync repo like `GitRepository.async` that returns the Promise for an async repo, so that internal callers can use async without changing anything in the `GitRepositoryProvider`?
- Documentation is sparse because I wasn't sure this was a good path to follow until today. I'll flesh it out more as I go.

cc @atom/core @orderedlist 
- [ ] Determine a good test case (even if manual / subjective) for git-related UI responsiveness
  - many simultaneous changes, check tree-view update?
  - simulate degraded disk performance / network drive?
- [ ] cf #9240


nathansobo:
What's the plan for transitioning to this API? I was thinking our existing synchronous `GitRepository` could always have an `async` property that points to this new version. We can then encourage users of today's APIs to use the async version like this:

```js
let repository = someMethodReturningARepository().async
```

When we hit Atom 2.0 and transition to the repository *always* being async, we can retain the `async` property and just point it to self, such that `repository.async === repository` returns `true`.

However, I'm not sure how this all meshes with your idea of keeping this API private for a while. In my mind, we should start transitioning the community to any async APIs we feel confident about. We might not want to make every method on the class public, but I'd like to deprecate existing synchronous functionality as soon as we can.

thedaniel:
>  I was thinking our existing synchronous GitRepository could have an async property that points to this new version. 

Yep, I agree with this, it's suggested in the bullet points above.

> However, I'm not sure how this all meshes with your idea of keeping this API private for a while. In my mind, we should start transitioning the community to any async APIs we feel confident about. 

This is not necessarily a conflict with how I feel, I'm just not expecting we'll feel confident about the APIs for a while.

spec/git-repository-async-spec.coffee
```diff
@@ -0,0 +1,406 @@
...
+
+  describe ".getPath()", ->
+    # XXX HEAD isn't a git directory.. what's this spec supposed to be about?
+    xit "returns the repository path for a .git directory path", ->
+      # Rejects as malformed
+      repo = GitRepositoryAsync.open(path.join(__dirname, 'fixtures', 'git', 'master.git', 'HEAD'))
+
+      onSuccess = jasmine.createSpy('onSuccess')
+
+      waitsForPromise ->
```
nathansobo: If you also write the specs in ES6, `async/await` syntax could really clean up these async tests...

```js
it('does something async', function () {
  waitsForPromise(async function() {
    expect(await repo.getPath()).toBe('/some/path')
  })
})
```

src/git-repository-async.js
```diff
@@ -0,0 +1,191 @@
...
+const _ = require('underscore-plus')
+
+module.exports = class GitRepositoryAsync {
+  static open (path) {
+    // QUESTION: Should this wrap Git.Repository and reject with a nicer message?
+    return new GitRepositoryAsync(path)
+  }
+
+  constructor (path) {
+    this.repo = null
```
nathansobo: What's this used for?

src/git-repository-async.js
```diff
@@ -0,0 +1,191 @@
...
+
+  getPath () {
+    return this.repoPromise.then((repo) => {
+      return Promise.resolve(repo.path().replace(/\/$/, ''))
+    })
+  }
+
+  isPathIgnored (_path) {
+    return this.repoPromise.then((repo) => {
+      return Promise.resolve(Git.Ignore.pathIsIgnored(repo, _path))
```
nathansobo: Is `pathIsIgnored` synchronous?

src/git-repository-async.js
```diff
@@ -0,0 +1,191 @@
...
+    })
+  }
+
+  isPathModified (_path) {
+    return this._filterStatusesByPath(_path).then(function (statuses) {
+      var ret = statuses.filter((status) => {
```
nathansobo: V8 supports the new ES6 `Array.prototype.every` method, which would be clearer and save you allocation of a new array.

nathansobo:
:zap: Thanks for moving us in the right direction on this.

nathansobo:
> it's suggested in the bullet points above.

Sorry, I didn't comprehend it initially for some reason. Of course `async` will have to be a promise, and when we transition all our repository-returning APIs to be asynchronous, we'll have to have an `async` property on the returned *promise* that references the original promise.

smashwilson:
Oh hey, this will be excellent for packages that integrate with git. :+1: :grin:

alexandernst:
:heart: ES6 instead of :coffee:script!

thedaniel:
Thought: it seems like a good place to start using this in prod & packages is to proxy all the `did-change-status` events from the sync implementation to the async implementation. At this moment I'm just writing tests that assert the events fire correctly in async-land at all and calling through from sync to async in a few methods (e.g. `getPathStatus(path)`) just so the side-effects happen, but ultimately we could remove anything solely concerned with generating events out of the sync implementation entirely and just re-emit the ones generated by the async implementation.

src/git-repository-async.js
```diff
@@ -0,0 +1,272 @@
...
+// Just using this for _.isEqual and _.object, we should impl our own here
+const _ = require('underscore-plus')
+
+module.exports = class GitRepositoryAsync {
+  static open (path, options = {}) {
+    // QUESTION: Should this wrap Git.Repository and reject with a nicer message?
+    return new GitRepositoryAsync(path, options)
+  }
+
+  constructor (path, options) {
```
philipgiuliani: What do you think about removing the spaces in front of the arguments? https://github.com/airbnb/javascript#18.3

thedaniel: Good catch - this currently conforms to [JS Standard Style](https://github.com/feross/standard) (a very presumptuous name, isn't it?) rather than AirBnB style and I'm happy with Standard.

philipgiuliani: Thanks for the resource. Didn't knew about it! Looks of course also good :+1: Just important that there is one style over the whole repository :+1: :clap:

src/git-repository-async.js
```diff
@@ -0,0 +1,255 @@
...
+    this.repo = null
+    this.emitter = new Emitter()
+    this.subscriptions = new CompositeDisposable()
+    this.pathStatusCache = {}
+    this._gitUtilsRepo = GitUtils.open(path) // TODO remove after porting ::relativize
+    this.repoPromise = Git.Repository.open(path)
+
+    var {project, refreshOnWindowFocus} = options
+    this.project = project
+    if (refreshOnWindowFocus === undefined) {
```
steelbrain: If you want a boolean value from `refreshOnWindowFocus`, then use `refreshOnWindowFocus = Boolean(refreshOnWindowFocus)`. It'll turn `false`, `null` and others to `false` and `true` remains the same.

src/git-repository-async.js
```diff
@@ -0,0 +1,255 @@
...
+      this.project.getBuffers().forEach((buffer) => { this.subscribeToBuffer(buffer) })
+    }
+  }
+
+  destroy () {
+    this.subscriptions.dispose()
+  }
+
+  getPath () {
+    return this.repoPromise.then((repo) => {
```
steelbrain: in ES6, if you are only getting one parameter, you can write `(repo) => { }` as just `repo => { }`

src/git-repository-async.js
```diff
@@ -0,0 +1,255 @@
...
+      return Promise.resolve(Git.Ignore.pathIsIgnored(repo, _path))
+    })
+  }
+
+  isPathModified (_path) {
+    return this._filterStatusesByPath(_path).then(function (statuses) {
+      var ret = statuses.filter((status) => {
+        return status.isModified()
+      }).length > 0
+      return Promise.resolve(ret)
```
steelbrain: You don't necessarily have to return a promise from `Promise.prototype.then`

.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "env": {
```
steelbrain: May I suggest the eslint [`no-var`](http://eslint.org/docs/rules/no-var.html) clause?

thedaniel:
Thanks for the feedback @steelbrain!

src/git-repository-async.js
thedaniel: I actually want to check undefined here, because it should default to `true` - that is to say, `true` is `true` and `undefined` is `true`, but only explicitly set`false` is `false`, which isn't quite how the Boolean constructor works.

.eslintrc
thedaniel: That's not a bad idea. There's a bunch of other JS in the repo that won't comply to this, but we don't need to have CI fail on eslint failures right away..

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+
+  destroy () {
+    this.subscriptions.dispose()
+  }
+
+  getPath () {
+    return this.repoPromise.then((repo) => {
+      return Promise.resolve(repo.path().replace(/\/$/, ''))
```
steelbrain: You might want to replace most of these `Promise.resolve`s here with just values.

peduxe:
You guys really hate placing semicolons don't you? :D

SeanJM:
By not adding semi colons, are you adding overhead to the JIT?

peduxe:
@SeanJM I don't know why github doesn't use them in their code, is this because they have most of their code in Ruby?

SeanJM:
https://www.quora.com/Are-there-performance-differences-between-using-semi-colons-and-not-using-semi-colons-in-JavaScript <https://www.quora.com/Are-there-performance-differences-between-using-semi-colons-and-not-using-semi-colons-in-JavaScript>
According to this the answer is no.

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+
+  constructor (path, options) {
+    this.repo = null
+    this.emitter = new Emitter()
+    this.subscriptions = new CompositeDisposable()
+    this.pathStatusCache = {}
+    this._gitUtilsRepo = GitUtils.open(path) // TODO remove after porting ::relativize
+    this.repoPromise = Git.Repository.open(path)
+
+    var {project, refreshOnWindowFocus} = options
```
thomasjo: This should be `let` (or `const`), doesn't look like we want or need hoisting here.

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+  isPathIgnored (_path) {
+    return this.repoPromise.then((repo) => {
+      return Promise.resolve(Git.Ignore.pathIsIgnored(repo, _path))
+    })
+  }
+
+  isPathModified (_path) {
+    return this._filterStatusesByPath(_path).then(function (statuses) {
+      return statuses.filter((status) => {
+        return status.isModified()
```
thomasjo: This can be changed into
```js
return statuses.filter(status => status.isModified())
```

You should probably also be able to rewrite the whole block as
```js
return this._filterStatusesByPath(_path).then(statuses => statuses.filter(status => status.isModified()))
```

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+      }).length > 0
+    })
+  }
+
+  isPathNew (_path) {
+    return this._filterStatusesByPath(_path).then(function (statuses) {
+      var ret = statuses.filter((status) => {
+        return status.isNew()
```
thomasjo: Same deal here with the `var`, and collapsing the single return function.

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+      }).length > 0
+      return Promise.resolve(ret)
+    })
+  }
+
+  checkoutHead (_path) {
+    return this.repoPromise.then((repo) => {
+      var checkoutOptions = new Git.CheckoutOptions()
```
thomasjo: `let`

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+      checkoutOptions.paths = [this._gitUtilsRepo.relativize(_path)]
+      checkoutOptions.checkoutStrategy = Git.Checkout.STRATEGY.FORCE | Git.Checkout.STRATEGY.DISABLE_PATHSPEC_MATCH
+      return Git.Checkout.head(repo, checkoutOptions)
+    }).then(() => {
+      return this.getPathStatus(_path)
```
thomasjo: Can be collapsed.

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+    })
+  }
+
+  checkoutHeadForEditor (editor) {
+    return new Promise(function (resolve, reject) {
+      var filePath = editor.getPath()
```
thomasjo: `let`

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+      if (filePath) {
+        if (editor.buffer.isModified()) {
+          editor.buffer.reload()
+        }
+        resolve(filePath)
+      } else {
+        reject()
+      }
+    }).then((filePath) => {
+      return this.checkoutHead(filePath)
```
thomasjo: Can be collapsed.

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+    })
+  }
+
+  // Returns a Promise that resolves to the status bit of a given path if it has
+  // one, otherwise 'current'.
+  getPathStatus (_path) {
+    var relativePath = this._gitUtilsRepo.relativize(_path)
```
thomasjo: `let`

src/git-repository-async.js
```diff
@@ -0,0 +1,271 @@
...
+    }).then((statuses) => {
+      var cachedStatus = this.pathStatusCache[relativePath] || 0
+      var status = statuses[0] ? statuses[0].statusBit() : Git.Status.STATUS.CURRENT
+      if (status !== cachedStatus) {
+        this.emitter.emit('did-change-status', {path: _path, pathStatus: status})
+      }
+      this.pathStatusCache[relativePath] = status
+      return Promise.resolve(status)
+    })
+  }
```
thomasjo: `let`s, and collapses :smiley:

src/git-repository-async.js
thedaniel: I had been trying to only use `=>` where I needed lexical this (though it snuck in here), is it typical to just use it everywhere?

thomasjo: They are typically used everywhere. I use them just as I use skinny arrows in CS.

thedaniel:
Hm, I only have 2 (seemingly unrelated) failures on my machine, but there are a bunch of :red_circle: on travis. In any case this is starting to come together nicely.

nathansobo:
> but there are a bunch of :red_circle: on travis

You may want to run tests against the current released version rather than master locally just in case that's a factor.

thedaniel:
Whoops! @nathansobo i meant to post that comment on the tree-view PR. Tabs are hard.

spec/git-repository-async-spec.coffee
thedaniel: I think that's ES7+ only

thomasjo: Yes, which we have access to :smile:

thedaniel:
Over in #9469 as of this moment I have just the buffer callback stuff implemented (i.e. the specs that are weirdly flaky on this branch) and consistently green. I'm not sure what the essential difference is though. One difference is that I implemented `::relativize` in that PR rather than just proxying to the git-utils version, but I'm not sure how that's relevant. I also implemented the specs slightly differently, with some gentle use of async/await. In any case, tomorrow I'll port some of the differences over to this branch and see if it makes things green here, else I'll port things from here to there one section of specs at a time and see if any one area starts to make async-land flaky.

joshaber:
So I *think* this branch is in a pretty good spot now.

1. Ported more tests to JS.
2. Moved from Jasmine waiting to async/await as much as I know how.
3. Fixed a race between file writing/git reading.

The tests _seem_ to consistently pass locally now. I'd be really curious to know how they behave for you now too :pray: 

My plan for Monday was to:

- [ ] Continue porting tests to JS.
- [ ] Make `terribleWait` and `terribleSleep` less terrible.
- [ ] See about getting CI building again so we can have some consistency to testing.

If you wanted, we could divide and conquer that.

thedaniel:
> Fixed a race between file writing/git reading.

😻

> If you wanted, we could divide and conquer that.

Sure thing, let me attack this CI thing first.

thedaniel:
The last build failed on a bunch of license checks, at least a couple related to eslint & standard added as `devDependencies`. One could argue that those libs are only used to develop, not in the distribution, so it's fine that their licenses aren't permissive. though I'm not 100% sure that holds water. For now I'll just take them out of the package.json (but leave in the config for standard) and see if that clears up this check license task.

joshaber:
So the tests are in a pretty good spot now, I think. There's still a potential race condition but I haven't seen in practice. I'm thinking over how to best fix it.

But the more pressing problem is that CI is timing out.

`master` currently builds in ~1200s, where this is timing out at 1800s. So somehow this added ~10 minutes to the build.

I'm not entirely sure *where* the time's being spent. Anecdotally we seem to spend a lot of time in the `Installing apm...` phase. We're using `node-pre-gyp` to download the nodegit binary (or at least, that happens locally; unclear how to verify that from CI output).

@atom/feedback Anyone have ideas? If not I'll probably try to break some of the steps down and time them.

orderedlist:
nodegit takes forever to build in apm because there aren't any binaries. This is probably something we should fix in nodegit.

thedaniel:
Also, atom.io can have high latency sometimes, and we hit it many times to install core packages. I do plan to do a performance pass there at some point in the future, but in any case it would be so cool if we could just bundle / cache those dependencies rather than going over the network.

joshaber:
>nodegit takes forever to build in apm because there aren't any binaries. This is probably something we should fix in nodegit.

nodegit uses node-pre-gyp to provide binaries, but it looks like it special-cases Electron apps to always build from source: https://github.com/nodegit/nodegit/blob/b259b0587c6e44e3939e49055b90845fbfce9221/lifecycleScripts/install.js#L27 Presumably because Electron wasn't supported until fairly recently? https://github.com/mapbox/node-pre-gyp/pull/175

>but in any case it would be so cool if we could just bundle / cache those dependencies rather than going over the network

I'd love to do some sort of binary caching but I'd really love for it to not block this.

thedaniel:
@joshaber I brought up the `Installing apm...` thing in chat the other day because it takes forever on my laptop as well though it "should be fast", worth looking into. Though it would be a shame to yak shave too much. The nodegit node-gyp path seems like the lowest-hanging fruit.

orderedlist:
Installing nodegit actually has to compile libgit2 as well, which takes some time. Realistically, an `apm` binary for nodegit would make this wayyyyy faster.

joshaber:
Yeah I'm looking into adding Electron binaries to nodegit.

johnhaley81:
@orderedlist is right with building libgit2 but it's even worse than that. NodeGit has to build libgit2/libssh2/openssl which is as horrible as you think it is. Luckily we don't have a C++11 requirement anymore but it's still super slow.

Anyways, so nodegit/nodegit#783 is in and ready to go so really the next step is to have the CI build/publish binaries for linux/darwin/win32 during a release. That way everything is just automated and ready to go. The only thing that would break is if a new version of electron is released there's no way for nodegit to pick up on that and we won't be able to grab a binary for the latest.

The only way around that (that I can think of right now) is just having a box that hooks into releases of electron and automates building/publishing for nodegit.

This might not be too much of a problem when Electron gets to v1 but still, it's going to be a pain to always go back and build/publish this guy when Electron gets a version bump.

joshaber:
:metal: Tests are passing, just need to get build times down so it doesn't time out :metal:

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,457 @@
...
+  fit(name, () => {
+    waitsForPromise(fn)
+  })
+}
+
+function xasyncIt(name, fn) {
+  xit(name, () => {
+    waitsForPromise(fn)
+  })
+}
```
maxbrunsfeld: @nathansobo ended up adding functions like these in [async-spec-helpers.coffee](https://github.com/atom/atom/blob/master/spec/async-spec-helpers.coffee). They detect if a promise is returned from the `it` function, so you can just shadow `it` and friends.

.npmrc
```diff
@@ -1 +1,4 @@
 cache = ~/.atom/.npm
+runtime = electron
+target = 0.34.5
```
joshaber: I don't like that we have to put this here and in package.json. Anyone have ideas about how we could not do that?

joshaber: Maybe set the environment variable from within bootstrap? Or something?

src/git-repository-async.js
```diff
@@ -0,0 +1,719 @@
...
+      .then(repo => repo.getHeadCommit())
+      .then(commit => commit.getEntry(_path))
+      .then(entry => entry.getBlob())
+      .then(blob => {
+        const options = new Git.DiffOptions()
+        if (process.platform === 'win32') {
+          // Ignore eol of line differences on windows so that files checked in
+          // as LF don't report every line modified when the text contains CRLF
+          // endings.
+          options.flags = Git.Diff.OPTION.IGNORE_WHITESPACE_EOL
```
50Wliu: :heart_eyes: Will this close #9424?

joshaber: Huh... Unfortunately I doubt it. Our [current implementation](https://github.com/atom/atom/blob/77a540962312ecf2a9cedd483cb4329564590458/src/git-repository.coffee#L387) is supposed to use this flag too.

joshaber:
Alright, I believe this is ready for review :pray: @atom/feedback 

There are two caveats:

1. The Windows builds are still timing out because we're having issues getting Windows nodegit binaries. Still a work in progress. I won't merge this until that's taken care of.
2. Submodule support is incomplete. I'm [working](https://github.com/nodegit/nodegit/pull/821) [on](https://github.com/nodegit/nodegit/pull/825) it, but I think this can land without it, so long as everyone's aware.

My current plan is:

1. Add submodule support once a new nodegit version is out.
2. Convert `tree-view`, `status-bar`, and `git-diff` to async.

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,598 @@
...
+  })
+
+  describe('.isPathIgnored(path)', () => {
+    beforeEach(() => {
+      repo = openFixture('ignore.git')
+    })
+
+    it('resolves true for an ignored path', async () => {
+      const ignored = await repo.isPathIgnored('a.txt')
+      expect(ignored).toBeTruthy()
```
kevinsawicki: Might be worth making these a little "stronger" with `toBe(true)` since this API should always return a `boolean` right? Looks like the `isStatusModified` specs do that currently.

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,598 @@
...
+
+  describe('.destroy()', () => {
+    it('throws an exception when any method is called after it is called', async () => {
+      repo = new GitRepositoryAsync(require.resolve('./fixtures/git/master.git/HEAD'))
+      repo.destroy()
+
+      let threw = false
+      try {
+        await repo.getShortHead()
+      } catch (e) {
```
kevinsawicki: Would it be better to catch this exception and assert the message and/or type? To ensure it throws the expected error and not just any error.

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,598 @@
...
+      await repo.refreshStatus()
+
+      expect(await repo.getCachedPathStatus(cleanPath)).toBeUndefined()
+      expect(repo.isStatusNew(await repo.getCachedPathStatus(newPath))).toBeTruthy()
+      expect(repo.isStatusModified(await repo.getCachedPathStatus(modifiedPath))).toBeTruthy()
+    })
+  })
+
+  describe('buffer events', () => {
+    let repository
```
kevinsawicki: Super minor, but it looks like most of the other times in other specs this variable is called `repo` and not `repository`, not sure if it would be worth using a consistent name throughout, but just wanted to mention it.

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,598 @@
...
+      waitsFor(() => !repository._isRefreshing())
+    })
+
+    it('emits a status-changed event when a buffer is saved', async () => {
+      const editor = await atom.workspace.open('other.txt')
+
+      editor.insertNewline()
+
+      let called
+      repository.onDidChangeStatus(c => called = c)
```
kevinsawicki: I think it would be better to use a Jasmine spy here and wait on the `callCount` and then assert on  `toHaveBeenCalledWith` helper.

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,598 @@
...
+    })
+  })
+
+  describe('.getLineDiffs(path, text)', () => {
+    beforeEach(() => {
+      const workingDirectory = copyRepository()
+      repo = GitRepositoryAsync.open(workingDirectory)
+    })
+
+    it('can get the line diff', async () => {
```
kevinsawicki: Maybe something like "returns the old and new lines of the diff" would be a clearer name for this spec.

src/git-repository-async.js
```diff
@@ -0,0 +1,870 @@
...
+  // Returns a {Promise} that resolves true if the given path is a submodule in
+  // the repository.
+  isSubmodule (_path) {
+    return this.repoPromise
+      .then(repo => repo.openIndex())
+      .then(index => {
+        // TODO: This'll probably be wrong if the submodule doesn't exist in the
+        // index yet? Is that a thing?
+        const entry = index.getByPath(_path)
+        const submoduleMode = 57344 // TODO compose this from libgit2 constants
```
kevinsawicki: This seems like a constant that might make sense to put at the top with the status flags.

src/git-repository-async.js
```diff
@@ -0,0 +1,870 @@
...
+
+    const {project, subscribeToBuffers} = options
+    this.project = project
+    if (this.project && subscribeToBuffers) {
+      this.project.getBuffers().forEach(buffer => this.subscribeToBuffer(buffer))
+      this.subscriptions.add(this.project.onDidAddBuffer(buffer => this.subscribeToBuffer(buffer)))
+    }
+  }
+
+  // Public: Destroy this {GitRepositoryAsync} object.
```
kevinsawicki: Minor question, does our documentation parser work for JavaScript docs already?

src/git-repository-async.js
```diff
@@ -0,0 +1,870 @@
...
+  // is ignored.
+  isPathIgnored (_path) {
+    return this.repoPromise.then(repo => Git.Ignore.pathIsIgnored(repo, _path))
+  }
+
+  // Get the status of a directory in the repository's working directory.
+  //
+  // * `directoryPath` The {String} path to check.
+  //
+  // Returns a promise resolving to a {Number} representing the status. This value can be passed to
```
kevinsawicki: `promise` -> `{Promise}`

src/git-repository-async.js
joshaber: I don't think so. I think that's what https://github.com/atom/donna/issues/9 is about.

src/git-repository-async.js
```diff
@@ -0,0 +1,870 @@
...
+        })
+      })
+  }
+
+  // Refresh the status bit for the given path.
+  //
+  // Note that if the status of the path has changed, this will emit a
+  // 'did-change-status' event.
+  //
+  // path    :: String
```
kevinsawicki: This appears to be in a different documentation format than the other methods in this file.

src/git-repository-async.js
```diff
@@ -0,0 +1,870 @@
...
+      buffer.onDidReload(refreshStatusForBuffer),
+      buffer.onDidChangePath(refreshStatusForBuffer),
+      buffer.onDidDestroy(() => {
+        bufferSubscriptions.dispose()
+        this.subscriptions.remove(bufferSubscriptions)
+      })
+    )
+
+    this.subscriptions.add(bufferSubscriptions)
+    return
```
kevinsawicki: I don't think this explicit `return` is needed since it is the last line of this function.

src/git-repository.coffee
```diff
@@ -75,6 +76,15 @@ class GitRepository
     unless @repo?
       throw new Error("No Git repository found searching path: #{path}")
 
+    asyncOptions = {}
+    for key, val of options
```
kevinsawicki: Super minor, but this could be `asyncOptions = _.clone(options)` instead.

kevinsawicki:
> Alright, I believe this is ready for review :pray: /cc @atom/feedback

Gave it a review pass, left some comments. I know a lot of this code was ported over from the original sync version so I apologize if any of the comments pertain to things that were legacy and are just showing up in the diff because of the movement.

maxbrunsfeld:
:heart: This looks great.

spec/git-repository-async-spec.js
joshaber: That's a really great point :+1:

joshaber:
@kevinsawicki Thanks! :sparkling_heart: I think I :hammer:'d them all.

Any other feedback is welcome. Otherwise this is just waiting on Windows nodegit binaries.

src/git-repository-async.js
```diff
@@ -0,0 +1,871 @@
...
+  //
+  // Returns a {Promise} resolving to a {Number} representing the status. This
+  // value can be passed to {::isStatusModified} or {::isStatusNew} to get more
+  // information.
+  getDirectoryStatus (directoryPath) {
+    let relativePath
+    // XXX _filterSBD already gets repoPromise
+    return this.repoPromise
+      .then(repo => {
+        relativePath = this.relativize(directoryPath, repo.workdir())
```
kevinsawicki: This variable appears to be only used in this promise but is declared outside this scope on line 402, could the `let` move here instead?

src/git-repository-async.js
```diff
@@ -0,0 +1,871 @@
...
+
+  // Public: Returns a {Promise} which resolves to the {String} working
+  // directory path of the repository.
+  getWorkingDirectory () {
+    return this.repoPromise.then(repo => repo.workdir())
+  }
+
+  // Public: Returns a {Promise} that resolves to true if at the root, false if
+  // in a subfolder of the repository.
+  isProjectAtRoot () {
```
kevinsawicki: Looks like this method returns `null` instead of a `Promise` if `this.project` is falsey. Should it return a rejected Promise in that case instead?

kevinsawicki: Or maybe in that case it should return a promise that just resolves to false.

src/git-repository-async.js
```diff
@@ -0,0 +1,871 @@
...
+  }
+
+  // Public: Returns true if the given status indicates the path is ignored.
+  //
+  // * `statusBit` A {Number} representing the status.
+  //
+  // Returns a {Boolean} that's true if the `statusBit` indicates the path is
+  // ignored.
+  isStatusIgnored (statusBit) {
+    return (statusBit & (1 << 14)) > 0
```
kevinsawicki: Super minor, worth pulling `1 << 14` to a `ignoredStatusFlags` constant for consistency?

kevinsawicki:
> Any other feedback is welcome.

Took another pass, left a few minor comments. This is looking really good though :+1:

src/git-repository-async.js
joshaber: Yup, good catch :+1:

joshaber:
:zap: :eyes:

src/git-repository-async.js
```diff
@@ -0,0 +1,873 @@
...
+
+  // Public: Returns a {Promise} that resolves to true if at the root, false if
+  // in a subfolder of the repository.
+  isProjectAtRoot () {
+    if (!this.project) return Promise.resolve(false)
+
+    if (this.projectAtRoot) {
+      return this.projectAtRoot
+    } else {
+      this.projectAtRoot = Promise.resolve(() => {
```
kevinsawicki: Looks like `this.projectAtRoot` won't be returned when this case is hit since there isn't a return below here.

joshaber: :astonished:

spec/git-repository-async-spec.js
```diff
@@ -0,0 +1,620 @@
...
+import Project from '../src/project'
+
+temp.track()
+
+function openFixture (fixture) {
+  return GitRepositoryAsync.open(path.join(__dirname, 'fixtures', 'git', fixture))
+}
+
+function copyRepository () {
+  let workingDirPath = temp.mkdirSync('atom-working-dir')
```
YuriSolovyov: I guess can be `const` ?

src/git-repository-async.js
```diff
@@ -0,0 +1,873 @@
...
+    //
+    // The original implementation also handled null workingDirectory as it
+    // pulled it from a sync function that could return null. We require it
+    // to be passed here.
+    if (!_path || !workingDirectory) {
+      return _path
+    }
+
+    if (process.platform === 'win32') {
+      _path = _path.replace(/\\/g, '/')
```
YuriSolovyov: Can't we make use of `node`s core libs like `path` here?

src/git-repository-async.js
```diff
@@ -0,0 +1,873 @@
...
+
+  // Public: Resolves true if the given path is modified.
+  //
+  // * `path` The {String} path to check.
+  //
+  // Returns a {Promise} which resolves to a {Boolean} that's true if the `path`
+  // is modified.
+  isPathModified (_path) {
+    return this._filterStatusesByPath(_path).then(statuses => {
+      return statuses.filter(status => status.isModified()).length > 0
```
YuriSolovyov: Can we use [Array.prototype.some](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/some)
```javascript
return statuses.some(status => status.isModified())
```
here?

The thing is that it can return earlier without iterating whole collection and also not producing new array.

src/git-repository-async.js
```diff
@@ -0,0 +1,873 @@
...
+      })
+      .then(statuses => {
+        return Promise.all(statuses.map(s => s.statusBit())).then(bits => {
+          let directoryStatus = 0
+          const filteredBits = bits.filter(b => b > 0)
+          if (filteredBits.length > 0) {
+            filteredBits.forEach(bit => directoryStatus |= bit)
+          }
+
+          return directoryStatus
```
YuriSolovyov: how about:
```javascript
return bits.filter(b => b > 0).reduce((status, bit) => { return status |= bit }, 0);
```
?

src/git-repository-async.js
joshaber: That's a good question. As the comment indicates, this was cargo-culted from `git-utils` so I don't have any context on it. @thedaniel any idea?

src/git-repository-async.js
joshaber: Fantastic :metal:

joshaber:
Thanks @YuriSolovyov!

joshaber:
Look at that lovely green checkmark :heart_eyes:

joshaber:
I've verified Mac and Windows binaries are working. Linux will have to wait until after the new year.

thedaniel:
Coincidentally, 2016 is the year of Linux on the desktop, so that works out great.

joshaber:
:penguin: :fire:

joshaber:
Even though it's not yet 2016, I verified that the Linux build is working :metal: :penguin:

joshaber:
Hoping to merge this tomorrow :pray:

joshaber:
There's reportedly a problem with the Linux nodegit binaries.

From @johnhaley81:

>It seems to only affect LTS linux...
>It's very strange. Like it's executing a CPU instruction that's not available to older versions of ubuntu.

So I'm going to hold off on merging this until that's resolved.

johnhaley81:
I'll keep you posted as I find out more.

alexandernst:
CPU instructions are not bound to software, but to hardware. If Linux binaries are not working on LTS Ubuntu (as there is no such thing as LTS Linux) it means you're using unavailable libraries (most probably). `ldd` should be enough to check that.

joshaber:
FWIW, it worked fine for me in testing on Ubuntu 64-bit 14.0.4.3

johnhaley81:
The LTS Linux was a typo. I meant Ubuntu. The issue was when requiring the NodeGit library it would throw a `SIGILL`. It throws this when it's initializing OpenSSL for the library but not at a specific line of code. It dies when it returns from the first method in the call stack after `init` is called.

This happens with the prebuilt binary off of Travis. I'm attempting to configure a VM to allow @joshaber to have a consistent repo since it seems to be a heisenbug as well.

alexandernst:
`SIGILL` errors are expected in OpenSSL. Check https://www.openssl.org/docs/faq.html#PROG17

johnhaley81:
That says under debugger situations which we aren't in. We're in a full production build without an attached debugger. Is that expected under that scenario as well?

johnhaley81:
Ahh I re-read it and it's saying that during it's query for a CPU's capabilities it could get a SIGILL. Running the code though it was happening during a request to `dev/random` and it did successfully get a result from that. It dies when it pops the function from the call stack.

johnhaley81:
@alexandernst here's the `ldd -v` output for each binary. The shared libs looks the same to me. Also, are those hex values the address in the binary where the lib is loaded? 

https://gist.github.com/johnhaley81/617f36595b35513cc2a4

alexandernst:
@johnhaley81 Can we get the backtrace too?

johnhaley81:
@alexandernst https://gist.github.com/johnhaley81/abbc6dd801eca780ae7f

alexandernst:
@johnhaley81 It looks either like nodegit was compiled for the wrong arch (can you show me how is it compiled?) or an OpenSSL bug (less likely, but possible).

johnhaley81:
@alexandernst this is the prebuilt binary compilation log: https://travis-ci.org/nodegit/electron-npg-automator/jobs/97494428

johnhaley81:
Also so you know, I have a "magic" Ubuntu box that's configured the same way that works. Clearly not everything is the same but it's a clean install, no updates, only those packages are install, same env-variables and same compiler and compiler version. And that works. I can give you any outputs from that machine that you like.

alexandernst:
@johnhaley81 Just to make sure I understand the situation. The bug you're getting exist only in a VM that you use to test? Or can you repro the bug easily in any LTS Ubuntu?

johnhaley81:
@alexandernst I've had reports of users experiencing this across multiple versions of Ubuntu. I have only been able to repro this consistently on this on 2 VMs and only when running Ubuntu. Mint and CentOS work.

johnhaley81:
Also my local install of Ubuntu works fine with either build.

johnhaley81:
@alexandernst @joshaber mentioned that the [disassmble](https://gist.github.com/johnhaley81/cae460ffee3c5876a236) of the running app might help

joshaber:
```
(gdb) x/i $pc
=> 0x7fc83ef3851c <lh_new+332>:    vxorps %xmm0,%xmm0,%xmm0
```

This seems to be the instruction it's SIGILL'ing on.

alexandernst:
@johnhaley81 @joshaber http://www.felixcloutier.com/x86/XORPS.html

Do the users's CPUs that are reporting the problem support AVX?
Also, I'm yet to see how nodegit is exactly compiled (the link @johnhaley81 posted doesn't have that info). Are you compiling with `-march=native` or something?

johnhaley81:
@alexandernst that travis box loads the repo from github and then we have the GYP file loaded https://github.com/nodegit/nodegit/blob/master/vendor/libgit2.gyp which sets the flags. That is how the makefile is generated. We then use `node-gyp` to compile the module.

johnhaley81:
@alexandernst as far as AVX support, I'll get back to you on that. Looking into it as much as I can now.

alexandernst:
Maybe something weird is going on in the Makefile. Can we get a copy of the generated Makefile?

About the AVX, check if available with `cat /proc/cpuinfo| grep avx`

alexandernst:
Also, the version of clang, if possible.

johnhaley81:
The definitely are lacking avx. So here's a potential fix: https://github.com/nodegit/electron-npg-automator/commit/d4a8725b7d42011cd39edf938d1927f97500eb04

The `use_obsolete_assembly` should force OpenSSL to NOT use that but I think that setting it to true might have not set the flag correctly? Trying this out now.

alexandernst:
There you go, it's a compiler flag. That's why it was working on your machine (which I guess it does support AVX) and not in the machines of some of your users. Find how to force the compilation process to avoid using unsupported CPU ops.

johnhaley81:
Compiler flag didn't fix. Turns out it's possibly a bug in OpenSSL. Potential fix: https://github.com/nodegit/nodegit/pull/855

alexandernst:
Indeed, turns out it's an OpenSSL bug :)
https://github.com/openssl/openssl/commit/817ddb9fb07e78b451e838a77b2b272b0dd23e5f

johnhaley81:
Confirmed, that fixes it.

johnhaley81:
Bumping to NodeGit 0.7.0 should resolve that linux issue and make this PR gtg :)

thedaniel:
Hm, I guess I didn't cross-reference this here: https://github.com/atom/atom/pull/10352

^ This is a fix for ::relativize on this branch that still needs a little polish

joshaber:
I'm seeing intermittent failures about file watching, e.g.:

```
>> GitRepositoryAsync
>>   .checkoutHeadForEditor(editor)
>>     it does not display a dialog when confirmation is disabled
>>       Error: Leaking subscriptions for paths: /private/var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/atom-working-dir116011-49688-dfmojy/a.txt
>>         at ensureNoPathSubscriptions (/Users/travis/build/atom/atom/spec/spec-helper.coffee:122:15)
>>         at [object Object].<anonymous> (/Users/travis/build/atom/atom/spec/spec-helper.coffee:115:3)
>>         at process._tickCallback (node.js:368:9)
>>   .destroy()
>>     it throws an exception when any method is called after it is called
>>       Error: Leaking subscriptions for paths: /private/var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/atom-working-dir116011-49688-1pr3ez3/a.txt
>>         at ensureNoPathSubscriptions (/Users/travis/build/atom/atom/spec/spec-helper.coffee:122:15)
>>         at [object Object].<anonymous> (/Users/travis/build/atom/atom/spec/spec-helper.coffee:115:3)
>> 
>> 
```

`GitRepositoryAsync` itself doesn't do any file watching directly. Is this a bug somewhere else? Some cleanup we're not doing? @atom/core

joshaber:
I moved some stuff around which maybe fixed that? I dunno.

I think we're just waiting on #10352, and then :tada: :balloon: :birthday:

kevinsawicki:
Awesome work here :+1:

johnhaley81:
w00t!

image

martindale:
Does this address #2456?

50Wliu:
@martindale it would be great if you could download the 1.7.0 beta and find out :).

martindale:
Subjectively, the problem appears to be worse with 1.7.0 beta.  Atom now completely hangs when launching it inside of a remote mount, and eventually requests direction on whether to close or keep waiting.

yannickcr:
Same problem here, Atom 1.6.0 or 1.7.0 beta now hangs up when trying to open a big git repository inside of a remote mount making Atom unusable in this configuration :cry:

bynite:
Can't use Atom anymore. Probably because of this feature. I mounted a vagrant via SSHFS and Atom freezes when opening a file. Could you release a fix for this, please.

mehcode:
I'm wondering why we switched to using nodegit. I realize that maintaining a wrapper around a git executable is lame but nodegit wraps libgit2 which is glacial in comparison to git. This is completely anecdotal but I've always found it far slower.

Perhaps using something like https://www.npmjs.com/package/simple-git instead?

thedaniel:
@mehcode Before nodegit we were already using libgit2, with a sync api wrapper we were maintaining ourselves (https://github.com/atom/git-utils) - so we can't blame any performance regressions on libgit2 here.

mehcode:
Ah. I was under the impression we were just doing shell execution.

Issue #8766, opened 2015-09-12
"Open In Dev Mode..." doesn't open in dev mode when no files are open


burodepeper:
Atom: 1.0.11
Platform: Mac OS X 10.10.5

### Issue
When I have no files open, and use `View > Developer > Open In Dev Mode` to open a folder, that folder is opened, but not in dev mode. If I have any file opened, `Open In Dev Mode` works as expected/intended.

izuzak:
Thanks for the report, @burodepeper. Strange, I'm not able to reproduce this on OSX 10.10.5 and Atom 1.0.12-dev-c7d4871. `View > Developer > Open In Dev Mode` opens a new window in dev mode regardless of whether I have any files opened or not.

@maxbrunsfeld I noticed that you assigned yourself to this -- were you able to reproduce this?

burodepeper:
I've been able to narrow it down a bit more.

If I open Atom, an empty untitled file is visible. It's _then_ that I can't `Open In Dev Mode`. If I close that window, so that there is no window open and only the `Panes >` and `Developer >` items are enabled in the `View` menu, then I can open in dev mode directly.

izuzak:
Thanks, @burodepeper -- I can reproduce now as well.

Here's what I did:

1. Start Atom
2. Open new window with `cmd-shift-n`
3. In the new window (which will have an untitled tab opened), run the `View > Developer > Open In Dev Mode` command and then select a folder.

PR #9218, opened 2015-10-19
Don't try to reuse existing normal windows as dev mode windows

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/8766


burodepeper:
Thanks!

PR #9229, opened 2015-10-20
Command Line Parameter to set Atom to Portable Mode

raelyard:
This is an enhancement to Portable Mode PR #8442 to provide a command line parameter for Atom to automatically set up a Portable Mode directory (a .atom directory as a sibling to the directory containing the Atom executable).

This uses -p and --set-portable
as a straw-man example of what would be the desired parameter.

This copies the .atom home from the default location to the portable location, assuming someone who has been using Atom and wants to go portable would want to take with them what they have already installed/configured.  This may or may not be a good assumption.


kevinsawicki:
Closing in favor of #9419 which is the same code but targets master.

PR #9235, opened 2015-10-21
Take 2: Pass current options when loading from previous window state

### LINKED ISSUE #9233: --profile-startup flag doesn't record a profile when launching Atom for the first time with no path
More context here: https://github.com/atom/docs/pull/136#issuecomment-149606128

@kevinsawicki mentioned:

> Hmm, I wonder if the timeout needs to be bumped in https://github.com/atom/atom/blob/806b46bb6c981f110a34ed20f566de50288900d0/static/index.js#L143-L145

> There seemed to be issues in past Electron versions where you needed to wait a bit to start profiling after the devtools opened event fired.

I tried bumping the timeout to 500ms and that did fix the problem for me. But I'm not sure what a good timeout would be so opening this to track the issue.

cc @kevinsawicki @nathansobo

nathansobo:
Fixes #9233

This makes atom --profile-startup work correctly even if Atom isn't running when the command is executed.

I accidentally left the commit on master which marked the previous PR (#9234) as merged. I force-pushed to master seconds after it happened to correct the issue, so I'm opening this again.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -87,7 +87,7 @@ class AtomApplication
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test
       @openWithOptions(options)
     else
-      @loadState() or @openPath(options)
+      @loadState(options) or @openPath(options)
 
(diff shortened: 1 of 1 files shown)


izuzak:
@nathansobo Strange, this doesn't fix the problem for me, see GIF below (I'm on the latest version of master):

profilestartup

However, using a longer timeout like @kevinsawicki suggested in https://github.com/atom/atom/issues/9233 does fix the problem for me.

PR #9238, opened 2015-10-21
Increase timeout for starting a profile after opening devtools

### LINKED ISSUE #9233: --profile-startup flag doesn't record a profile when launching Atom for the first time with no path

izuzak:
Hopefully this closes https://github.com/atom/atom/issues/9233, see https://github.com/atom/atom/issues/9233#issuecomment-149674829.

cc @kevinsawicki @nathansobo

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -141,7 +141,7 @@
     } else {
       currentWindow.openDevTools()
       currentWindow.once('devtools-opened', function () {
-        setTimeout(profile, 100)
+        setTimeout(profile, 1000)
       })
     }
   }


nathansobo:
Thanks @izuzak. Between our two changes this should hopefully be good for people.

PR #9241, opened 2015-10-21
Fix Build Issues For Node v4 and apm v1.3.0+

joefitzgerald:


.node-version
```diff
@@ -1 +1 @@
-v0.12.0
+v4.2.1
```
thomasjo: We should outright delete this file. It's not used by nvm at least.

script/bootstrap
```diff
@@ -66,7 +66,7 @@ function bootstrap() {
 
   var buildInstallCommand = initialNpmCommand + npmFlags + 'install';
   var buildInstallOptions = {cwd: path.resolve(__dirname, '..', 'build')};
-  var apmInstallCommand = npmPath + npmFlags + '--target=0.10.35 ' + 'install';
+  var apmInstallCommand = npmPath + npmFlags + '--target=4.2.1 ' + 'install';
```
thomasjo: Have you tested by removing the `target` argument? This is a code smell. We should not be be hardcoding versions in build scripts in order to avoid the "nightmare" scenario we just experienced.

PR #9262, opened 2015-10-23
Ignore git repo which has .atomignore file

codingbelief:
This PR is to make atom ignores git repository which has .atomignore in the working directory.
Atom will consume 90%+ I/O usage and get sick , when open folders which belong to a huge git repository. Maybe we can have chance to ignore huge git repository.

image


damieng:
We discussed this today and decided we don't want to special-case these repos and instead want to work on making sure that we handle huge git repositories.

PR #9265, opened 2015-10-23
Startup Snapshot

as-cii:
This is an experimental branch which makes use of the [v8 snapshot feature](http://www.hashseed.net/2015/03/improving-v8s-performance-using.html) to preload some of the files Atom requires on startup.

:warning: Please, note that this is far from being finished and could potentially never be: I aim to constantly update this PR's description with what I find out, along with the performance savings that snapshotting brings us, but we could experience some limitation that will force us to ditch this technique. :warning:
### Why snapshotting?

If we look closely at the startup profile, we will soon notice that most of the time is spent in requiring files. This is somewhat a downside of using an interpreted language: everything from loading files into memory to executing them must happen at runtime.

With snapshotting, however, we're allowed to take a snapshot of the heap and make it available to the V8 context without actually running any code at runtime. Indeed, we can exploit this to load files during our build pipeline (i.e. `script/build`) and simply refer to the already required modules when Atom is executed.

We _think_ this could bring a decent speedup in our startup phase, which on my machine currently takes `1.0s`-`1.2s`.
### How does snapshotting works?

Electron provides a `mksnapshot` utility that allows to take a javascript file and turn it into a blob ready to be loaded in memory. After generating it, loading it is just a matter of copying it inside the Electron.framework directory:

``` bash
$ cp snapshot_blob.bin /Applications/Atom.app/Contents/Frameworks/Electron\ Framework.framework/Resources/snapshot_blob.bin
```

The capabilities that the _snapshot phase_ provides, though, are somewhat limited. Unfortunately, we won't have access to many of the handy facilities that we're allowed to use in a normal Electron application. For example:
1. Node;
2. Functions such as `Object::toString` (or anything else defined in v8 `js_natives`);
3. Native modules;

This means we have to change many (here I should say "most") of our files to be arranged in order to overcome the above limitations (e.g. by deferring `require`s of native modules).
### What's currently in this PR

I have managed to load the entire `text-buffer` library and to use it within Atom. I plan to go on with other expensive require paths, such as `src/project` and `src/config`.

I am still a bit unsure on how to proceed exactly, but before automating the generation of `snapshot.js`, I'd prefer to carefully include source files so that I can patch external libraries without actually opening PRs on the other repos. This is because I'd like to understand what are the realistic savings that this technique could allow us before delving into all the other details that concern building, etc.
### What's missing

Here I'll include what this PR would need in order to be :ship:ped. I do not expect to follow any of this point in a particular order, but they will serve as a good TODO list (also in case anyone wants to help out with this).
- [ ] Generate `snapshot.js` automatically. (right now I have manually included the source files)
- [ ] Change all the core source files so that they can be required in the snapshot.
- [ ] Change all the libraries source files so that they can be required in the snapshot.
- [ ] Include benchmarks to demonstrate the potential of snapshotting.
- [ ] Investigate how debugging changes when having functions already loaded in memory. (can we use source maps here somehow?)

Any kind of feedback here would be super appreciated. Thanks! :bow: 


lee-dohm:
I could totally be misremembering, but I thought I read somewhere that V8 snapshotting is platform-specific? So an OS X snapshot would not necessarily work on a Windows machine?

as-cii:
> I could totally be misremembering, but I thought I read somewhere that V8 snapshotting is platform-specific? So an OS X snapshot would not necessarily work on a Windows machine?

Yeah, I think you're right. I should have probably mentioned that I am still unclear on when we should generate the snapshot: I think we could either do it at runtime, or simply have another step in our build pipeline (which already works on a per-platform basis).

alexandernst:
This is very interesting. Please post the performance benefits of this when you're done with the benchmarking/testing.

YuriSolovyov:
How well it works with modules containing native parts?

50Wliu:
@as-cii You might find this comment by @zcbenz helpful: https://github.com/atom/atom/pull/6877#issuecomment-107878462

YuriSolovyov:
Just curious: are js/css files minified or just compiled on release builds?

nathansobo:
> Just curious: are js/css files minified or just compiled on release builds?

Just compiled and packaged in the ASAR bundle. Any evidence that minification substantially improves compilation / execution time? It's not intuitively obvious that it would.

alexandernst:
I can't stop thinking of https://top.fse.guru/nodejs-a-quick-optimization-advice-7353b820c92e#.qctfhcko7
Maybe give it a try?

paulcbetts:
Didn't @zcbenz already try this and figure out it actually made things slower?

nathansobo:
> Didn't @zcbenz already try this and figure out it actually made things slower?

Depends on what "this" actually is. Our global environment was in a really weird tangled state at the time and he can't find the branch where he tried it for us to look at. In light of that, I think it could be worth exploring again just to be sure and gather some more specifics for ourselves.

as-cii:
> Just curious: are js/css files minified or just compiled on release builds?

Reading files is, unfortunately, not the bottleneck: there's still a fair amount of time that we spend on constructing prototypes and so on.

I am not sure this will end up being merged into master, as the tradeoffs for snapshotting a pseudo-working version of `AtomEnvironment` are several. I have put this on pause as I think there is another part of the initialization process we could exploit to get some decent speedups: I'll come back to this as soon as I validated the other approach (which still relies on cutting down compile times) and profiled the startup once again.

> Didn't @zcbenz already try this and figure out it actually made things slower? (Not to rain on your parade, just perhaps save you some work to come up with the same conclusion)

As @nathansobo said, we're in a better situation now to understand what causes slowdowns during startup and to possibly change it. This should differ from the approach @zcbenz tried out because the intent of this PR is not (only) to include the compiled JS files, but rather to execute them. This means that, after creating the snapshot, Atom could take advantage of an already instantiated `AtomEnvironment`: I think this is feasible in theory but it requires quite a lot of work to cleanly dissect Atom so that we can run some parts of it inside a constrained environment such as the snapshot one (i.e. a bare v8 context without  `jsnatives`).

mehcode:
> Just compiled and packaged in the ASAR bundle. Any evidence that minification substantially improves compilation / execution time? It's not intuitively obvious that it would.

In theory it would reduce overall file read time and compilation time by a small percentage. I'd be very interested what the startup time difference is (even if its a very small percentage, atom is a huge project, it could help some). If someone can help point where to poke in the atom build process I wouldn't mind trying it out.

benogle:
@as-cii is this dead?

as-cii:
Yeah, I don't feel like this is feasible from a practical standpoint, although it led us to implement the native compile cache eventually.

Thanks everyone for the feedback! And thank you @benogle for the heads-up. :+1:

PR #9268, opened 2015-10-23
Use ELECTRON_ENABLE_LOGGING instead of console.log shims

thomasjo:
This PR attempts to leverage the `ELECTRON_ENABLE_LOGGING` environment variable flag added to Electron in v0.34.1. Setting the environment variable enables `console.log` support in Electron, just as the `--enable-logging` application argument does.

I haven't tested my changes on Windows yet, so it'd be brilliant if someone (@50Wliu perhaps?) could quickly check if they work. Otherwise I'll boot up one of my Windows VMs later and give it a whirl.

Note that I tried setting this in various places in code, but it seemed that wasn't early enough in the application lifecycle, hence why I ended up setting the environment variable in the shell scripts. Also note that I'm setting the variable whenever `EXPECT_OUTPUT` is set which I feel is appropriate, but if someone disagrees, I can easily change it so that it's only set in the case of launching Atom in test mode.

/cc @atom/feedback

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -50,6 +50,10 @@ if [ $REDIRECT_STDERR ]; then
   exec 2> /dev/null
 fi
 
+if [ $EXPECT_OUTPUT ]; then
+  export ELECTRON_ENABLE_LOGGING=1
+fi
+
 if [ $OS == 'Mac' ]; then
   if [ -n "$BETA_VERSION" ]; then
     ATOM_APP_NAME="Atom Beta.app"
--- a/resources/win/atom.cmd
+++ b/resources/win/atom.cmd
@@ -16,6 +16,7 @@ FOR %%a IN (%*) DO (
 )
 
 IF "%EXPECT_OUTPUT%"=="YES" (
+  SET ELECTRON_ENABLE_LOGGING=YES
   "%~dp0\..\..\atom.exe" %*
 ) ELSE (
   "%~dp0\..\app\apm\bin\node.exe" "%~dp0\atom.js" %*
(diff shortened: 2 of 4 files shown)


nathansobo:
Thanks for this. Seems like we could just use the flag instead of the environment variable if you're moving stuff into the shell scripts... am I missing something.

thomasjo:
> Thanks for this. Seems like we could just use the flag instead of the environment variable if you're moving stuff into the shell scripts... am I missing something.

Yeah, that's true! I can easily switch to that instead if you prefer that. I wish we could set this somewhere else, but doesn't seem like that's possible at the moment.

thomasjo:
@zcbenz I think v0.34.1 might have had a regression with respect to scrollbars in OS X. My system is setup to *auto*, and when I'm scrolling in Atom using this branch, the scrollbars only show up after I've stopped scrolling. They are actually hidden during scrolling. In other words, the behavior is inverse of what it should be.

Double checking now whether this bug is also present on master (in case this has regressed earlier in the week). Will update.

50Wliu:
I'm getting console output, so it seems to work on Windows.  The only thing that might be off is a missing `TESTS PASSED`/`TESTS FAILED` line, though that might be because specs still crash on Windows after they finish.

50Wliu:
@thomasjo :ship:?

thomasjo:
@50Wliu We could, but I was contemplating switching from using the environment variable `ELECTRON_ENABLE_LOGGING` to simply passing the `--enable-logging` flag (as proposed by @nathansobo). I just never got around to it.

thomasjo:
Upon checking what's involved in switching from the environment variable to the feature flag, I believe it's actually cleaner to stick to variable. It requires fewer changes, and is likely more robust (up to a certain point).

I think this is ready to :ship:

nathansobo:
Cool, as we can confirm long as logging still works in tests, :ship:.

thomasjo:
Just double-checked, and logging still works.

PR #9274, opened 2015-10-24
Add top/bottom bars

### LINKED ISSUE #8939: Make it easier to add top/bottom bars
From this discussion: https://discuss.atom.io/t/placing-element-at-top-of-atom-for-package/21118

#### Problem:

Currently you can't easily add a bar to the top or bottom that spans across the whole width. If you try, it pushes down the status bar. I guess because `atom-workspace-axis` has `height: 100%`.

screen shot 2015-09-25 at 3 22 20 pm

An option is to add it above/below the editor, like `tool-bar` currently does, but that might not be desired for all packages.

#### Possible solution:

Add two empty `<atom-panel-container>` as children of `<atom-workspace>`. They're just placeholders and won't take up space if they are empty, but it would allow other packages to fill them.

atom-panels

simurai:
This PR adds a top and bottom bar to `atom-workspace`.

``` html
<atom-workspace class="workspace">
  <atom-panel-container class="header"></atom-panel-container><!-- New -->
  <atom-workspace-axis class="horizontal"></atom-workspace-axis>
  <atom-panel-container class="footer"></atom-panel-container><!-- New -->
  <atom-panel-container class="modal"></atom-panel-container>
</atom-workspace>
```

atom-panels

It allows packages to add `<atom-panel>`s that span across the whole width and not just above/under the editor.

The `status-bar` or `tabs` _could_ go on top and under the tree-view.

screen shot 2015-10-24 at 3 15 51 pm

screen shot 2015-10-24 at 3 04 02 pm
#### Todo
- [x] Change to flexbox
- [x] Add top/bottom bar elements
- [x] Add API
- [x] Add specs
#### Todo after merging
- [ ] Merge: https://github.com/atom/tree-view/pull/627
#### Packages that break or have issues
- [ ] [nuclide-file-tree](https://atom.io/packages/nuclide) (Scrolling is broken)
- [x] [tool-bar](https://github.com/suda/tool-bar/) (scrolling is broken) -> [PR](https://github.com/suda/tool-bar/pull/111)
- [x] [atom-chat](https://atom.io/packages/atom-chat) (Overflowing content, no scrolling. Fix: Use Flexbox, only scroll `chat-messages`) -> [PR](https://github.com/mertkahyaoglu/atom-chat/pull/7)
- [x] [youtube-pane](https://atom.io/packages/youtube-pane) (footer will cover the youtube panel). -> [PR](https://github.com/alexandruionascu/youtube-pane/pull/3)

Closes #8939

### DIFF
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -47,6 +47,8 @@ class Workspace extends Model
       left: new PanelContainer({location: 'left'})
       right: new PanelContainer({location: 'right'})
       bottom: new PanelContainer({location: 'bottom'})
+      topBar: new PanelContainer({location: 'top-bar'})
+      bottomBar: new PanelContainer({location: 'bottom-bar'})
       modal: new PanelContainer({location: 'modal'})
 
     @subscribeToEvents()
@@ -66,6 +68,8 @@ class Workspace extends Model
       left: new PanelContainer({location: 'left'})
       right: new PanelContainer({location: 'right'})
       bottom: new PanelContainer({location: 'bottom'})
+      topBar: new PanelContainer({location: 'top-bar'})
+      bottomBar: new PanelContainer({location: 'bottom-bar'})
       modal: new PanelContainer({location: 'modal'})
 
     @originalFontSize = null
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -47,8 +47,8 @@ class Workspace extends Model
       left: new PanelContainer({location: 'left'})
       right: new PanelContainer({location: 'right'})
       bottom: new PanelContainer({location: 'bottom'})
-      topBar: new PanelContainer({location: 'top-bar'})
-      bottomBar: new PanelContainer({location: 'bottom-bar'})
+      header: new PanelContainer({location: 'header'})
+      footer: new PanelContainer({location: 'footer'})
       modal: new PanelContainer({location: 'modal'})
 
     @subscribeToEvents()
@@ -68,8 +68,8 @@ class Workspace extends Model
       left: new PanelContainer({location: 'left'})
       right: new PanelContainer({location: 'right'})
       bottom: new PanelContainer({location: 'bottom'})
-      topBar: new PanelContainer({location: 'top-bar'})
-      bottomBar: new PanelContainer({location: 'bottom-bar'})
+      header: new PanelContainer({location: 'header'})
+      footer: new PanelContainer({location: 'footer'})
       modal: new PanelContainer({location: 'modal'})
 
     @originalFontSize = null
@@ -838,11 +838,11 @@ class Workspace extends Model
   addTopPanel: (options) ->
     @addPanel('top', options)
 
-  # Essential: Get an {Array} of all the panel items at the top of the window.
-  getTopBarPanels: ->
-    @getPanels('topBar')
+  # Essential: Get an {Array} of all the panel items in the header.
+  getHeaderPanels: ->
+    @getPanels('header')
 
-  # Essential: Adds a panel item to the top of the window. It will take up full width.
+  # Essential: Adds a panel item to the header.
   #
   # * `options` {Object}
   #   * `item` Your panel content. It can be DOM element, a jQuery element, or
@@ -854,14 +854,14 @@ class Workspace extends Model
   #     forced closer to the edges of the window. (default: 100)
   #
   # Returns a {Panel}
-  addTopBarPanel: (options) ->
-    @addPanel('topBar', options)
+  addHeaderPanel: (options) ->
+    @addPanel('header', options)
 
-  # Essential: Get an {Array} of all the panel items at the bottom of the window.
-  getBottomBarPanels: ->
-    @getPanels('bottomBar')
+  # Essential: Get an {Array} of all the panel items in the footer.
+  getFooterPanels: ->
+    @getPanels('footer')
 
-  # Essential: Adds a panel item to the bottom of the window. It will take up full width.
+  # Essential: Adds a panel item to the footer.
   #
   # * `options` {Object}
   #   * `item` Your panel content. It can be DOM element, a jQuery element, or
@@ -873,8 +873,8 @@ class Workspace extends Model
   #     forced closer to the edges of the window. (default: 100)
   #
   # Returns a {Panel}
-  addBottomBarPanel: (options) ->
-    @addPanel('bottomBar', options)
+  addFooterPanel: (options) ->
+    @addPanel('footer', options)
 
   # Essential: Get an {Array} of all the modal panel items
   getModalPanels: ->
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -838,6 +838,44 @@ class Workspace extends Model
   addTopPanel: (options) ->
     @addPanel('top', options)
 
+  # Essential: Get an {Array} of all the panel items at the top of the window.
+  getTopBarPanels: ->
+    @getPanels('topBar')
+
+  # Essential: Adds a panel item to the top of the window. It will take up full width.
+  #
+  # * `options` {Object}
+  #   * `item` Your panel content. It can be DOM element, a jQuery element, or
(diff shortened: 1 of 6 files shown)


thomasjo:
This looks very promising! :sparkles:

simurai:
I think *if* we allow other packages to add a panel that spans across the full width at the bottom, the `status-bar` should also go full width and be below it. Otherwise, this looks a bit odd having it above the `tool-bar`.

screen shot 2015-10-24 at 9 25 24 pm

mertkahyaoglu:
I'm working on a package that needs to be fullwidth. This will be :cool: .

src/workspace.coffee
```diff
@@ -47,6 +47,8 @@ class Workspace extends Model
       left: new PanelContainer({location: 'left'})
       right: new PanelContainer({location: 'right'})
       bottom: new PanelContainer({location: 'bottom'})
+      topBar: new PanelContainer({location: 'top-bar'})
+      bottomBar: new PanelContainer({location: 'bottom-bar'})
```
jerone: Why name it `bar` when they are panels?

thomasjo: They're not panels, they're (panel) containers.

simurai: Yeah, wasn't too sure about the `bar` name. Any good alternative? 

- `top-full`
- `top-full-width`
- `top-outer`
- `top-edge`
- `top-most`

jerone: [**@thomasjo**](https://github.com/thomasjo) commented on [24 okt. 2015 21:11 CEST](https://github.com/atom/atom/pull/9274#discussion_r42937216 "2015-10-24T19:11:10Z - Replied by Github Comment Enhancer"):
> They're not panels, they're (panel) containers.

I'm sorry, I meant when the content are panels.

I like the `top-most` and `bottom-most`, nut will leave that up to you guys. Just wasn't sure about `bar` when the containers aren't limited to bars only.

thomasjo: Naming things is probably the single most difficult thing in software development, hehe. I like `top-most` and `bottom-most` too, but something about those names don't "jive".

What about something simple like `header` and `footer`?

thedaniel: `header` and `footer` are my favorites, especially since the terms are extremely common in the web world and we're using web tech in atom :)

mnquintana: `header` and `footer` sound :+1: to me too

thedaniel:
@simurai Looks good to merge when you're satisfied. I don't think moving the status-bar when the panels are there is actually crucial because it's generally full of metadata for the current editor and it seems a little strange to separate it from the bottom of the editor by an arbitrary amount.

simurai:
@thedaniel I tested a few packages and made some tweaks so they don't break. So I think it should be good now.

Well, I still found [one](https://atom.io/packages/youtube-pane) that inlines the `height` based on the whole window, which doesn't work when header/footer gets populated. There are probably more that do the same, but maybe ok?

thedaniel:
I think it's maybe OK. What happens, does the youtube pane cover up the header/footer? We could always give package authors like @alexandruionascu feedback on a better way to calculate height after this gets merged.

simurai:
No, the footer covers the youtube pane. Which is not *too* problematic in this case.

screen shot 2015-10-28 at 1 01 49 am

Right, we can notify package authors about it.

thomasjo:
> Right, we can notify package authors about it.

We can even open PRs. When we merge this, just point me at repos and I can try to fix a few :wink:

simurai:
Sounds good. I started a "Packages that break or have issues" list in the description.

Only packages that add an `atom-panel` on the left/right should be affected.

thomasjo:
@simurai Great, thanks!

alexandernst:
I'm sorry for the complete offtopic, but what plugin is the one with the GitHub icon (among others) that you're using, @simurai ?

simurai:
@alexandernst It's https://atom.io/packages/tool-bar. Plus https://atom.io/packages/flex-tool-bar so you can easily map any command to a button.

alexandernst:
@simurai Thank you so much! This plugin should be part of core imho, it completely rocks! \o/

as-cii:
Code-wise this looks good to me: nice work here, @simurai! :sparkles: 

I know we haven't done so for other kinds of panels, but what would you think about testing the actual dimensions/coordinates in `workspace-element-spec.coffee`? So that we make sure that our flex box CSS code is actually working as expected. I can add those specs myself or we can :pear: and add them together. 

Also, I gave this a quick test-drive and I have jotted down a couple of concerns (which maybe are not related to this PR, but to the respective packages instead?):

1. Header panels cause tabs' top-border to look inconsistent/missing:
screen shot 2015-12-08 at 12 16 24

2. Header panels cause tree-view projects to look off:
screen shot 2015-12-08 at 12 20 52

I noticed you opened a PR on tree-view which is maybe addressing 2), but https://github.com/atom/tree-view/pull/627#issuecomment-151683080 seems to state it's not needed anymore. Thoughts?

simurai:
@as-cii Thanks for testing. Point 2 should be fixed with the PRs above.

Point 1: Yeah, it looks a bit odd if you just add text with One light. I guess it depends on the theme and the package that would occupy the `header`. For example the `tool-bar` has a bottom border so it should look ok.

screen shot 2015-12-08 at 12 24 34 pm

I think we can leave it up to the package/theme to either add a bottom border or a different background color to have some kind of separation.

>  I can add those specs myself or we can :pear: and add them together.

:pear:ing would be great. I'm flying home tonight, so I'll ping you about it.

rm-code:
Wow, very nice!

alexandernst:
@rm-code This hasn't been merged yet

rm-code:
@alexandernst I know :grin: Looking forward to it though. Especially if it enables #10208 to be merged as well.

iolsen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;RFEkawmU&#x2F;1-add-top-bottom-bars"> Add top&#x2F;bottom bars</a>

maxbrunsfeld:
The code looks good to me. @as-cii are you still interested in :pear:ing with @simurai to add some DOM-based specs?

as-cii:
Looks good! :sparkles: 

:shipit: after specs go :green_heart:!

benogle:
Travis builds are legit. Merging.

alexandernst:
@benogle Don't forget to merge https://github.com/atom/tree-view/pull/627 please

benogle:
Thanks for the heads up

zertosh:
This broke some Nuclide stuff too. Will this go out with 1.5.0 or 1.6.0?

benogle:
Sorry about that. It'll be out in 1.6

simurai:
@zertosh Sorry. Yeah, noticed that scrolling of Nuclide's file-tree will break.

Any idea where that <div> with inline `height: 100%;` comes from? If that <div> can be removed, I think it should work.

screen shot 2016-02-09 at 6 00 00 pm

Or if that <div> needs to stay, removing the `height` and adding flexbox should work too:

```less
atom-panel.left > div {
  height: initial !important;
  display: flex;
}
```

zertosh:
@simurai: No worries, @ssorallen is the one looking into. I think he has a solution to support both the new and old behavior- something we have to do since we can't upgrade everyone at fb to 1.6.0 at the same time.

ssorallen:
@simurai Thanks for checking out the Nuclide DOM. I will try out your fix.

simurai:
> I think he has a solution to support both the new and old behavior

Good point. Most packages currently use `.my-package { height: 100%; }`. Then in `1.6.0` core styles will override it with [initial](https://github.com/atom/atom/blob/127e4dbae404257f14e8a050c44c56964902c8ac/static/panels.less#L64-L68) `atom-panel-container > atom-panel.left > * { height: initial; }` to make it work with Flexbox and the added header/footer.

If you wanna keep the current markup, inheriting the height and adding flexbox seems to work with both versions.

```html
<atom-panel class="left tool-panel panel-left">
  <div style="height: inherit; display: flex;">
```

PR #9309, opened 2015-10-27
Move async `didStopChangingActiveItem` to `Pane`

ssorallen:
Functionality like focusing the newly active text editor lives in the
`Pane` class and needs access to the async `didStop` callback. Move the
functionality into `Pane` and keep the same API for `PaneContainer` by
adding a new subscription similar to its `observeActiveItem`
subscription.

Throttle the async version of `didChangeActiveItem` because debouncing
can lead to a UI that is significantly out-of-date compared with the
state expected by the user. For example, pressing and holding ctrl-tab
(next tab) would keep the event debounced for the duration of the
keypress. Throttling the event ensures most callbacks are still async
but are able to update the UI periodically if events are happening
quickly.

### DIFF
--- a/src/pane-container.coffee
+++ b/src/pane-container.coffee
@@ -9,8 +9,6 @@ module.exports =
 class PaneContainer extends Model
   serializationVersion: 1
   root: null
-  stoppedChangingActivePaneItemDelay: 100
-  stoppedChangingActivePaneItemTimeout: null
 
   constructor: (params) ->
     super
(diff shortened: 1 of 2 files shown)


ssorallen:
I changed this implementation after moving most core packages to use `onDidStopChangingActivePaneItem` instead of `onDidChangeActivePaneItem` and noticing the UI would be stagnant while pressing and holding the key combination to cycle through tabs.

ssorallen:
I will add pull requests for changes to core packages once this implementation is settled.

thedaniel:
cc @atom/feedback

nathansobo:
Can we think of a better name for this, since it's throttled now rather than debounced? Just a thought:  maybe you could pass in an amount of time you're *willing* to wait and/or an amount of time you must wait between firing when subscribing via onDid methods, and provide a general solution in event-kit. Trouble is that it complicates the interface.

nathansobo:
Maybe it just makes more sense to throttle the handlers when they're attached.

ssorallen:
@nathansobo It's possible to add `throttle` at all locations, but async should be normal rather than the current sync version. It might not be obvious to 3rd-party packages to throttle, debounce, or delay work on active item changes. I've found 7 calls in core packages that do synchronous work on `onDidChangeActiveItem` that would benefit from moving to async.

ssorallen:
@nathansobo I will send PRs for throttling the callbacks in a few core packages and see what you think. If it's a pattern than core wants to endorse, it'll be simple to replace the `throttle` calls with calls to `onDidStop...`.

PR #9313, opened 2015-10-27
Remove keyPath as a possible return value on the event object for onDidChange in config.coffee

### LINKED ISSUE #7918: config onDidChange does not return keypath in event
Per docs (https://github.com/atom/atom/blob/v1.0.2/src/config.coffee#L418) there should be oldValue, newValue, keyPath in the event passed to the callback. In fact only the first two are present

I think https://github.com/atom/atom/blob/v1.0.2/src/config.coffee#L871 is the issue. Correct?

yakirac:
Updated the documentation for onDidChange in config.coffee. The value is never returned on the object so the documentation was updated to reflect that. Should close https://github.com/atom/atom/issues/7918

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -422,7 +422,6 @@ class Config
   #   * `event` {Object}
   #     * `newValue` the new value of the key
   #     * `oldValue` the prior value of the key.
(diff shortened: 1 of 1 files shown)


thedaniel:
Thanks!

PR #9318, opened 2015-10-28
Introduce NativeCompileCache

as-cii:
This PR speeds up source code compile times by using a hacked version of the `vm` module (https://github.com/as-cii/cached-run-in-this-context).

We've been searching very hard for a way to make startup faster: the biggest low hanging fruit seemed to be `vm.runInThisContext` which, before this improvement, took more than `250ms-300ms`. That function is used inside `Module.prototype._compile` for, surprise, compiling the source code: although v8 compiler is quite fast already, our codebase has grown quite a lot and the more files we added, the worse the situation would get.

The snapshot approach tried in #9265 should have indirectly solved this, but it felt way too complicated to construct an `AtomEnvironment` inside the snapshot context. Instead, we are taking advantage of a v8 compiler feature that allows to speed up source code compilation by producing/consuming a cache blob. By using this mechanism, we have managed to change startup times from (without any tab open):

screen shot 2015-10-28 at 11 51 42

to:

screen shot 2015-10-28 at 11 48 05

:tada: :tada: :tada:

There's still some aspect we could improve, such as where these cache blobs get stored: I used files as the storage mechanism, as it allows us to keep the code pretty simple, but they take ~40ms to get read. There are alternatives to this (e.g. asar), but I am not sure it would improve the situation that much: even a 50% improvement, would make this just ~20ms faster at the expense of making code a bit more complex.

Please, note that whenever DevTools are open, v8 seems to make this optimization a bit worse: if you, however, spawn Atom:
- via cmd;
- by running the executable;
- by reloading the window with DevTools closed;

You will notice the speedup. @zcbenz: any ideas why this happens? I guess this covers already 100% of the average use cases and won't matter to the end user, but if possible it seems to be still worth investigating.

After this lands, I will go on hunting other things that are slowing startup down: we feel like IO could be the main cause but we'll get a better picture after this improvement gets :ship:ped.


zcbenz:
:sparkles: 

> whenever DevTools are open, v8 seems to make this optimization a bit worse @zcbenz: any ideas why this happens?

I guess devtools disabled the cache optimization, devtools is known to affect V8's optimization when opened since it needs to receive full information of the JavaScript context.

as-cii:
This seems to work pretty nicely and the I/O improvements made in b365ba5 seems to let Atom load within 670ms-690ms almost always on my machine.

@maxbrunsfeld @nathansobo: I have polished up the code as well, so your feedback would be super appreciated! :pray: 

@izuzak: features-wise nothing should have changed, but could you have a look and try to break :bomb:  it?

src/file-system-cache-blob-storage.js
```diff
@@ -0,0 +1,93 @@
+var fs = require('fs-plus')
+var path = require('path')
+
+module.exports = (function () {
+  FileSystemCacheBlobStorage.load = function (directory) {
```
maxbrunsfeld: What do you think about naming this class something generic like `BlobStore`, since its interface isn't specific to a caching use case?

as-cii: Good call. :+1: I think I still like the explicitness of mentioning that this is a file-system cache, though: so, what about naming it `FileSystemBlobStore`?

src/file-system-cache-blob-storage.js
```diff
@@ -0,0 +1,93 @@
...
+    var instance = new FileSystemCacheBlobStorage(directory)
+    instance.load()
+    return instance
+  }
+
+  function FileSystemCacheBlobStorage (directory) {
+    this.inMemoryCache = new Map()
+    this.cacheBlobFilename = path.join(directory, 'v8-compile-cache.blob')
```
as-cii: Based on the above comment, I think these filenames need to be changed as well. I cannot think of anything more generic than: `BLOB` and `MAP`. Ideas?

src/native-compile-cache.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+      var args = [self.exports, require, self, filename, dirname, process, global]
+      return compiledWrapper.apply(self.exports, args)
+    }
+  }
+
+  return NativeCompileCache
+
+})()
```
maxbrunsfeld: I think it'd be nice to use es6 class syntax for these two classes, since electron supports it natively now.

as-cii: :+1:

src/file-system-cache-blob-storage.js
maxbrunsfeld: Yeah, I like that. It implements the `BlobStore` interface using the file-system.

src/file-system-cache-blob-storage.js
maxbrunsfeld: Yeah, blob and map are perfect.

maxbrunsfeld:
:heart: Clean design. Clean tests. Super nice.

thomasjo:
Is there a reason the new bits are not written using Babel/ES2015? :confused:

maxbrunsfeld:
I commented above about using es6. Don't think we need babel though, now that we're on electron 34, unless we're using `async` or `await`.

src/file-system-cache-blob-storage.js
maxbrunsfeld: Another very minor thing, maybe `NativeCompileCache::setCacheStorage -> ::setBlobStore`?

thomasjo:
> I commented above about using es6. Don't think we need babel though, now that we're on electron 34, unless we're using async or await.

Does the ES6 bits in Electron support new module loader syntax (e.g. `import`)? Makes code so much nicer :art: 

Also `var` is generally evil, and should be avoided due to hoisting nonsense. Prefer `let` (and `const` when applicable).

Other than the ES syntax comment, this looks great! Awesome job as usual @as-cii :clap:

maxbrunsfeld:
No, it doesn't support `import`. Personally, I probably wouldn't introduce transpilation just for that feature, but I guess I'd be open to it. Agreed about `const` and `let`, which electron also supports natively. You might have to `use strict`, but that's fine.

maxbrunsfeld:
Also, in order to work in dev mode, in which files are transpiled on-the-fly, files like these, that are loaded early in the startup process (`compile-cache.js` and earlier), need to *not* be transpiled.

maxbrunsfeld:
I've been using this branch this morning, and I can definitely feel the improvement in startup.

as-cii:
> I've been using this branch this morning, and I can definitely feel the improvement in startup.

> Clean design. Clean tests. Super nice.

> Awesome job as usual

@thomasjo @maxbrunsfeld: :heart: :heart: :heart: 

I have just ported the code over ES6 and removed references to `cache`. It reads so better now, thanks for the suggestion guys! I'll give this another spin and merge in the weekend if everything is :cool:. :sunglasses:

YuriSolovyov:
Is that `cached-run-in-this-context` module is something worth upstreaming?

maxbrunsfeld:
:ship:

static/index.js
```diff
@@ -1,9 +1,19 @@
...
+  var fs = require('fs-plus')
   var path = require('path')
+  var FileSystemBlobStore = require('../src/file-system-blob-store')
+  var NativeCompileCache = require('../src/native-compile-cache')
 
   var loadSettings = null
   var loadSettingsError = null
+  var blobStore = null
+
+  app.on('before-quit', function () {
```
maxbrunsfeld: We might want to do this a different way, to avoid doing these two synchronous IPC calls to the browser process (`.require` and `.on`) during startup, and also to avoid contentious writes by multiple windows during shutdown. Maybe the browser process could actually control this in [its `before-quit` callback](https://github.com/atom/atom/blob/69dfdd0745209814b41c0c14e01f3d41ad0d27e8/src/browser/atom-application.coffee#L216), telling only one window to save its state, using some synchronous IPC method like [`WebContents::executeJavascript`](http://electron.atom.io/docs/v0.34.0/api/web-contents/#webcontents-executejavascript-code-usergesture).

as-cii:
Thanks everyone for all the great suggestions! :sparkles: If needed we can refine this it on a separate branch/commit.

nathansobo:
🐎🌠

maxbrunsfeld:
Yeah!! Nice job @as-cii

thomasjo:
Fantastic work as usual, @as-cii! :bow:

nathansobo:
Hey @as-cii: @kevinsawicki and I were talking yesterday and he brought up that we should probably include the v8 version in the compile cache key in case they change their internal formats. Did a quick audit and not sure this is currently happening, but we should probably get that in before this goes to beta if we can.

maxbrunsfeld:
>we should include the v8 version in the compile cache key in case they change their internal formats

V8 does have a built-in notion of [rejecting](https://github.com/atom/cached-run-in-this-context/blob/master/src/cached-run-in-this-context.cc#L35) a cache blob, and @as-cii has handled this case by [deleting it](https://github.com/atom/atom/blob/80bd25cf80e89643680919af8db1ccb94ff04fb6/src/native-compile-cache.js#L63) from our blob-store, so we may not need to worry about this.

nathansobo:
Oh right, that makes sense. They handle their own invalidation internally.

PR #9362, opened 2015-10-30
Focus pane items async, improve tab responsiveness

ssorallen:
The native `focus` call is the bulk of the time spent in response to
an active item change event. By debouncing the event and always focusing
asynchronously (async is the default for `debounce` callbacks), tab
changing is more responsive because the 'tabs' package typically updates
its UI only after the native `focus` call completes.

For the crazy perf test of pressing and holding next/previous tab
(ctrl-tab on OS X), many unnecessary `focus` calls are avoided because
of debouncing.

### DIFF
--- a/spec/pane-element-spec.coffee
+++ b/spec/pane-element-spec.coffee
@@ -58,8 +58,16 @@ describe "PaneElement", ->
       jasmine.attachToDOM(paneElement)
       paneElement.focus()
 
+      pane.activateItem(item1)
+      # Advance the clock to trigger the pane element's debounced active item
+      # change event handler.
+      advanceClock(51)
+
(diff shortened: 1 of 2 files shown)


ssorallen:
@thedaniel How does core handle implementation changes that affect specs in other packages? The failures in the CI build mostly break because they are testing an implementation detail. Namely, that when a pane element item is activated it also focuses its text editor.

thedaniel:
@ssorallen If the specs are truly just testing an implementation detail that's not relevant to the actual desired behavior under test, the only option is to go fix the package specs - in this case it sounds like just updating sync tests to use the waitsFor / waitsForPromise / runs queue to test that the correct item is focused asyncly.

PR #9383, opened 2015-11-01
Fix incorrectly reported width when measuring lines

### LINKED ISSUE #9382: Cursors are slightly off when font-size is smaller than 12px
Originally reported by @thomasjo. When setting font-size to, e.g. `11px`, we're experiencing some weird cursor positioning, which can cause a cursor to be positioned almost on top of another character.

screen shot 2015-11-01 at 19 30 47

as-cii:
This commit fixes what seems to be a bug in Chromium. Fixes #9382.

Before:

before

After:

screen shot 2015-11-01 at 19 33 27

When measuring lines (and with a special character sequence), it could happen
that Range(0, 0). getBoundingClientRect().width reports a number greater than 0.
This seems to happen when the font size is smaller than 12px and it's probably
due to subpixel font scaling. To solve it we've explicitly included a guard
clause that prevents this problem to happen.

/cc: @nathansobo @50Wliu @thomasjo @atom/feedback  for :eyes:

### DIFF
--- a/spec/lines-yardstick-spec.coffee
+++ b/spec/lines-yardstick-spec.coffee
@@ -2,7 +2,7 @@ LinesYardstick = require "../src/lines-yardstick"
 {toArray} = require 'underscore-plus'
 
 describe "LinesYardstick", ->
-  [editor, mockPresenter, mockLineNodesProvider, createdLineNodes, linesYardstick] = []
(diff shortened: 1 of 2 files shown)


nathansobo:
@as-cii Does this affect beta or stable?

as-cii:
I think it affects stable, as it's most likely related to the enablement of subpixel font-scaling.

as-cii:
For posterity, after double checking this we've concluded this affects only beta and not stable.

PR #9419, opened 2015-11-03
Command Line Parameter to set Atom to Portable Mode

kevinsawicki:
I'm porting this from #9229 and retargeting to master.

From @raelyard:

 This is an enhancement to Portable Mode PR #8442 to provide a command line parameter for Atom to automatically set up a Portable Mode directory (a .atom directory as a sibling to the directory containing the Atom executable).

This uses -p and --set-portable
as a straw-man example of what would be the desired parameter.

This copies the .atom home from the default location to the portable location, assuming someone who has been using Atom and wants to go portable would want to take with them what they have already installed/configured.  This may or may not be a good assumption.

### DIFF
--- a/src/browser/main.coffee
+++ b/src/browser/main.coffee
@@ -66,7 +66,7 @@ setupAtomHome = ({setPortable}) ->
     try
       AtomPortable.setPortable(atomHome)
     catch error
-      console.log("Failed copying portable directory: '#{atomHome}' to '#{AtomPortable.getPortableAtomHomePath()}'")
+      console.log("Failed copying portable directory '#{atomHome}' to '#{AtomPortable.getPortableAtomHomePath()}'")
       console.log("#{error.message} #{error.stack}")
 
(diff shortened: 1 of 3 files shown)


src/browser/main.coffee
```diff
@@ -102,6 +113,7 @@ parseCommandLine = ->
   options.boolean('profile-startup').describe('profile-startup', 'Create a profile of the startup execution time.')
   options.alias('r', 'resource-path').string('r').describe('r', 'Set the path to the Atom source directory and enable dev-mode.')
   options.boolean('safe').describe('safe', 'Do not load packages from ~/.atom/packages or ~/.atom/dev/packages.')
+  options.boolean('set-portable').describe('set-portable', 'Set portable mode. Copies the ~/.atom folder to be a sibling of the installed Atom location.')
```
50Wliu: I think just specifying `--portable` sounds nicer.

kevinsawicki: :+1: great feedback

PR #9422, opened 2015-11-04
Add portable mode on Linux

kevinsawicki:
/cc @raelyard

### DIFF
--- a/spec/atom-portable-spec.coffee
+++ b/spec/atom-portable-spec.coffee
@@ -3,6 +3,38 @@ fs = require 'fs-plus'
 temp = require "temp"
 AtomPortable = require "../src/browser/atom-portable"
 
+portableModeCommonPlatformBehavior = (platform) ->
+  describe "with ATOM_HOME environment variable", ->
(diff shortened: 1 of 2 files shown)


50Wliu:
No way! :heart_eyes:  I thought portable mode was just going to be for Windows :open_mouth:.

kevinsawicki:
> I thought portable mode was just going to be for Windows

Yeah, the plan is to support all platforms, there will be another PR for Mac soon.

ryanpcmcquen:
Nice work! :sparkles:

remzr7:
How does one use the portable mode?

raelyard:
@remzr7 Having a directory sibling to the Atom installation directory called .atom is the trigger for portable mode and that .atom functions as the Atom home directory.  Running atom with the --portable switch will create such a portable .atom home directory if it doesn't already exist.  For instance, if you run Atom from a directory in your Dropbox and execute atom --portable, you'll now have a .atom in your Dropbox that will take your customizations with you.

PR #9425, opened 2015-11-04
Add PPA instructions for Ubuntu

Calvein:


mnquintana:
Thanks for contributing! Unfortunately we don't support installing Atom from unofficial sources, including this PPA. Feel free to subscribe to https://github.com/atom/atom/issues/2956 for updates on adding an official Atom PPA.

Calvein:
Thanks for the heads up.

bilderbuchi:
@mnquintana well, you could just add information that this is not an officially supported source (or something to that effect). (For what it's worth, I've been successfully using that PPA for the last couple months myself). Especially considering that someone has invested the work to get a working solution together, while "the core team hasn't had time for this yet.", as you [said](https://github.com/atom/atom/issues/2956#issuecomment-154853501). This solution could tide people over until the core team gets to do it "eventually".

Issue #9208, opened 2015-10-19
Proposal: Marker layers


nathansobo:
Supplants #9137 

Find and replace currently scales poorly with the number of results. Based on a test of searching for the `e` character in `jquery.js` that took 2932ms, the biggest bottlenecks as follows:
- 670ms ~20% – Creating and subscribing to individual decorations for each result.
- 1752ms 60% – Creating markers, of which:
  - 893ms – Inserting markers into current marker index.
  - 533ms – Propagating marker creation events
- 125ms ~4% – Translating character offsets to rows and columns.

Based on discussion with @maxbrunsfeld and @as-cii, this is a proposal for changes to core APIs that attacks these bottlenecks directly.
## Marker Layers

`TextBuffer.prototype.addMarkerLayer` creates and returns a `MarkerLayer` object that represents a set of markers.
### Creating Markers

This object has a `markRange` method that takes a 2D rows, column range as normal, but it also has a `markCharacterIndexRange` that will eliminate the need to translate character indices to points during our scan. Both marker creation methods return an integer id instead of an object, which can be passed to `getMarker` to retrieve the actual marker object for direct manipulation. Working in terms of ids rather than objects saves us from allocating large quantities of what are essentially "convenience" objects until they are needed. We could potentially avoid the need to _ever_ allocate marker objects by manipulating individual markers via methods on `MarkerLayer` that take an id, such as `updateMarkerRange(id, range)` or `destroyMarker(id)`. However, this may interact poorly with existing APIs and would need to be carefully considered.
### Observing Updates

`MarkerLayer`s will also have an asynchronous, batch-oriented change subscription function called `onUpdate`. For more information on our need to transition asynchronous events in the model and the naming scheme that could facilitate this transition, see #9207.

Subscribers to `onUpdate` will be called once for any number of marker changes occurring in a given tick of the event loop. If needed, we can emit an event object containing arrays of `created`, `updated`, and `destroyed` marker ids, but since the view re-queries the marker index anyway during render, we may want to wait until we have a use case. I do however have a feeling that subscribers could be interested in this information.
## Decorating Marker Layers

To avoid the overhead of creating and subscribing to thousands of decorations, we can add `TextEditor::decorateMarkerLayer` that accepts a marker layer along with a decoration descriptor object that should apply to every range in that layer. This will allow us to create and decorate the search results layer _once_, and from then on just manipulate markers.
## Other Methods

[...]

PR #9426, opened 2015-11-04
Optimize markers via new layers API and treap-based index

nathansobo:
This PR improves the responsiveness of find-and-replace and other marker-intensive tasks by adding a layer-oriented API for working with collections of related markers and upgrading to a new implementation of the marker index. Searching for `e` in `jquery.js` now blocks the UI thread for ~1 second instead of ~3.2 seconds. Inserting newlines in this file with the markers present now takes ~10 ms instead of ~35 ms.

I still need to make the changes to find-and-replace backwards compatible and merge them after merging these changes.
### Marker Layers

A huge problem with our existing implementation is that creating markers emits synchronous events to support existing packages and core APIs. We needed a way to opt out of these event handlers for high-volume use cases, and marker layers were a good way to do this.
- `TextBuffer`
  - `.addMarkerLayer({maintainHistory})` Creates and returns a `MarkerLayer` object. If `maintainHistory` is `true`, markers will be restored on calls to undo/redo.
  - `.getMarkerLayer` Retrieve a marker layer based on its id.
- `MarkerLayer` Some highlights:
  - `.markRange` Creates a marker.
  - `.getMarker` Gets a marker by id.
  - `.findMarkers` Queries the marker layer.
  - `.onDidUpdate` Called asynchronously whenever markers are created, updated, or destroyed. Can be used without being a drag on performance, but you'll have to re-query any state you're interested in between calls.
  - `.onDidCreateMarker` This is needed to support our current synchronous selection APIs, but it shouldn't be used on layers that will contain a large number of markers because it slows down insertion.
  - `.copy` Create and return a new marker layer with all the same markers.
- `TextEditor`
  - `.addMarkerLayer({maintainHistory})` Creates and returns a `TextEditorMarkerLayer`.
  - `.getMarkerLayer` Retrieve an editor-level marker layer based on its id.
- `TextEditorMarkerLayer` This is a wrapper around a buffer-level `MarkerLayer` object and shares the same id.
  - `.markBufferRange` Create a marker.
  - `.markScreenRange` Create a marker.
  - `.findMarkers` Query markers with the same API as `TextEditor.prototype.findMarkers`
### Marker Layer Decorations

Another huge source of slowness in find and replace is that we decorated each marker individually. Now that we have marker layers, we can decorate an entire layer in the same way, avoiding a ton of allocations.
- `TextEditor`
  - `.decorateMarkerLayer(layer, properties)` Accepts a `MarkerLayer` or `TextEditorMarkerLayer` and the same properties that `decorateMarker` takes. It creates a decoration for every marker in the layer based on these parameters. This won't work for custom gutter decorations or overlay decorations, but _will_ work for highlights, line, and line number decorations. Returns a `LayerDecoration`.
- `LayerDecoration`
  - `setPropertiesForMarker(marker, properties)` This can be used to override the properties of a particular marker in the layer. We needed this to support the "current result" in find and replace without major changes to the package code. Pass `null` to clear the override.
## New Marker Index

marker-binary-tree

This PR also incorporates a new binary-tree-based [marker-index](https://github.com/atom/marker-index), which replaces the previous b-tree-based index in text-buffer. It offers better and more predictable performance... about 33% faster on my insertion tests (from 1.6 second to 1 second) and an order of magnitude faster at splicing. 1 second is still too long to block the UI thread, so we may still need to background the index population. I would like to experiment with a C++ implementation and see what it buys us, since it's a straightforward implementation that should be fairly easy to port.
## Performance Improvements

Now on to the good part... the improvement. This shows results from 3 implementations. The current version, marker layers with the old index, and marker layers with the new index.
#### Inserting ~20k markers:

~3.2 seconds

screen shot 2015-11-03 at 1 20 25 pm
#### Inserting newlines

~35 ms per keystroke

screen shot 2015-11-03 at 1 20 11 pm
#### Inserting ~20k markers:

1.6 seconds

screen shot 2015-11-03 at 11 48 37 am
#### Inserting newlines

~30 ms per keystroke

screen shot 2015-11-03 at 11 51 12 am

Zooming in, the newline insertion contains 3 ~5ms calls to `MarkerIndex.prototype.splice`:

screen shot 2015-11-03 at 12 21 51 pm
#### Memory consumed

36.2 MB

screen shot 2015-11-03 at 1 15 50 pm
#### Inserting ~20k markers

~1 second

screen shot 2015-11-03 at 12 45 31 pm
#### Inserting newlines

~7-10 ms per keystroke

screen shot 2015-11-03 at 12 48 16 pm

Only 1 significant splice operation, 0.8ms.

screen shot 2015-11-03 at 12 52 55 pm
#### Memory consumed

34.8 MB (not a huge savings)

screen shot 2015-11-03 at 1 13 42 pm


benogle:
:boom:

EasyHard:
Hi @nathansobo, I have a feeling that maybe we can use splay tree instead of treap since operations have locality. How do you think? If you think it may be a good idea. I would like to give it a shot.

nathansobo:
> Hi @nathansobo, I have a feeling that maybe we can use splay tree instead of treap since operations have locality

@EasyHard I considered using splay trees early on, but my concern is that a splay tree might not deliver consistent performance. While it is true that the *amortized* cost of access is optimal, splay trees can get in really unbalanced states such that a given operation would take longer. For batch workloads, this isn't a problem, but since Atom is an interactive application it's important that all operations have a fairly uniform cost.

Consider the example of populating the marker index in find and replace... We'll insert a bunch of markers in order, which without rebalancing will create a reversed linked-list structure. Then when we query the list for the first time, we'll have to walk the entire list to get to the beginning of the list. At this point we'll rebalance, but this first operation will be expensive.

It may be the case that we'll always find a way to amortize the cost without users noticing, but this wasn't obvious enough to me to want to allow our access patterns to be the only thing balancing the tree. Let me know if you think I'm missing something, but I'd prefer to stick to something more predictable.

nathansobo:
I converted `text-editor-component-spec.coffee` to Babel as part of this PR. Kind of off topic, but I was having trouble understanding asynchronous sequencing in that test after removing all the mocking, and I really wanted `async/await`. Turns out most of my issues were probably related to throttling, but the conversion was a ton of work so I'm keeping it.

nathansobo:
Really wondering what's going on with the build. Worked out most of the flakiness but I'm still timing out on animation frames that never fire on CI. From what I can tell there *is* a pending frame when I time out, but for some reason it's not occurring.

Now testing with an animation loop that logs if there's a frame > 500ms. May throw in the towel and go back to mocking the event loop, but I want to spend a bit more time understanding what the heck is going on with the event loop.

nathansobo:
Okay, I have at least one green build on our internal CI, which hasn't happened yet. So either I got lucky or have found the fix. Now I need to narrow down which of the many things I tried are necessary for stable passing builds.

nathansobo:
Refs #9208

PR #9440, opened 2015-11-04
Add package name metadata for load errors

winstliu:
Fatal notifications created during package load/activation/etc. now contain additional package name metadata for use in the Notifications package.

Required for atom/notifications#100.

/cc @kevinsawicki @benogle


benogle:
Legit 👍

PR #9444, opened 2015-11-05
Allow synchronous control of scroll position through TextEditor model

maxbrunsfeld:
This PR adds the ability to manipulate a TextEditor's vertical scroll position at the model level, via a new method, `::setFirstVisibleScreenRow` method, which is marked as `Experimental` for the time being. It also un-deprecates `TextEditor::getFirstVisibleScreenRow`, making it possible to synchronously read the editor's vertical scroll position.

This was driven out by vim-mode's [`ScrollFullPage` commands](https://github.com/atom/vim-mode/blob/e1d1a3edbaa16b6010e688b07c6cb3b9cd5c95a9/lib/motions/general-motions.coffee#L454), which need to synchronously update the editor's scroll position.

I've refactored the presenter so that it doesn't store its own copy of the `scrollRow` and `scrollColumn` state, using the model instead as the source of truth for these concepts. As part of this effort, I refactored the way the presenter's `scrollTop` and `scrollLeft` are initialized; they're now computed based on the `firstVisibleScreen{Row,Column}` as soon as the necessary measurements (line-height and base-character-width) are assigned.

Refs https://github.com/atom/vim-mode/pull/906


src/text-editor-presenter.coffee
```diff
@@ -1109,6 +1104,8 @@ class TextEditorPresenter
   setLineHeight: (lineHeight) ->
     unless @lineHeight is lineHeight
       @lineHeight = lineHeight
+      unless @scrollTop?
```
as-cii: :art: I think here we're checking if the position has been already restored: if it hasn't, we calculate it only the first time the line height is set.

Checking if `@scrollTop` has been defined, though, is somewhat misleading in this context: what would you think about extracting a method that tells this story a bit more explicitly? Something like: `::shouldRestoreScrollTop()`, `::shouldRestoreScrollLeft()`. Maybe we could also add another method that wraps `::updateScrollTop`, and call it `::restoreScrollTop()` (or `::deserializeScrollTop`) so that the code reads like:

```coffee
@restoreScrollTop() if @shouldRestoreScrollTop()
@restoreScrollLeft() if @shouldRestoreScrollLeft()
```

as-cii:
Awesome work, @maxbrunsfeld. Thank you for addressing this! :sparkles: 

I wonder if, analogously, we should emit an event whenever `::setFirstVisibleScreenColumn` is called on the model. What do you think?

By the way, I've left just one other minor comment, otherwise it looks good to me! :boom:

src/text-editor.coffee
```diff
@@ -3088,11 +3083,40 @@ class TextEditor extends Model
...
+        lineHeightInPixels = @displayBuffer.getLineHeightInPixels()
+        if height? and lineHeightInPixels?
+          maxScreenRow -= Math.floor(height / lineHeightInPixels)
+      screenRow = Math.max(Math.min(screenRow, maxScreenRow), 0)
 
-  getScrollColumn: -> @scrollColumn
-  setScrollColumn: (@scrollColumn) ->
+    unless screenRow is @firstVisibleScreenRow
+      @emitter.emit 'did-change-first-visible-screen-row', screenRow unless fromView
+      @firstVisibleScreenRow = screenRow
```
t9md: This line should come BEFORE emitting did-change-first-visible-screen-row?

maxbrunsfeld: :+1: Thanks @t9md. This method isn't intended for public consumption, but that would still be better style.

maxbrunsfeld:
> I wonder if, analogously, we should emit an event whenever ::setFirstVisibleScreenColumn is called on the model. What do you think?

I left the `setFirstVisibleScreenColumn` private because I don't we need it as a public API in the same way as we need `setFirstVisibleScreenRow`. The `onDidChangeFirstVisibleScreenRow` method is also private, and is only used by the presenter at the moment, so that it can respond to external calls to `setFirstVisibleScreenRow`.

So even though it's asymmetrical, I'd like to avoid adding an `onDidChangeFirstVisibleScreenColumn` method until we either need it as a public API, or we need a public API for setting the horizontal scroll position programatically.

as-cii:
> So even though it's asymmetrical, I'd like to avoid adding an onDidChangeFirstVisibleScreenColumn method until we either need it as a public API, or we need a public API for setting the horizontal scroll position programatically.

:+1: 

I think there's some lint error that's causing CIs to fail: other than that, I guess we can :shipit: :sparkles:.

src/text-editor.coffee
```diff
@@ -3192,11 +3187,40 @@ class TextEditor extends Model
     Grim.deprecate("This is now a view method. Call TextEditorElement::getWidth instead.")
     @displayBuffer.getWidth()
 
-  getScrollRow: -> @scrollRow
-  setScrollRow: (@scrollRow) ->
+  # Experimental: Scroll the editor such that the given screen row is at the
+  # top of the visible area.
+  setFirstVisibleScreenRow: (screenRow, fromView) ->
+    unless fromView
```
aleclarson: Thoughts on having some `NaN` detection here?

Without, you get this confusing error message:
```
Error: Undefined screen line when clipping screen position
    at DisplayBuffer.module.exports.DisplayBuffer.clipScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/display-buffer.js:878:15)
    at DisplayBuffer.module.exports.DisplayBuffer.bufferPositionForScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/display-buffer.js:830:20)
    at TextEditorMarker.module.exports.TextEditorMarker.setHeadScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-marker.js:185:60)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/cursor.js:67:31
    at Cursor.module.exports.Cursor.changePosition (/Applications/Atom.app/Contents/Resources/app.asar/src/cursor.js:673:7)
    at Cursor.module.exports.Cursor.setScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/cursor.js:65:19)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:1560:23
    at TextEditor.module.exports.TextEditor.moveCursors (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:1780:9)
    at TextEditor.module.exports.TextEditor.setCursorScreenPosition (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:1559:19)
    at TextEditorComponent.module.exports.TextEditorComponent.onMouseDown (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:757:25)
    at HTMLDivElement.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:3:61)
```

src/text-editor.coffee
```diff
@@ -3192,11 +3187,40 @@ class TextEditor extends Model
     Grim.deprecate("This is now a view method. Call TextEditorElement::getWidth instead.")
     @displayBuffer.getWidth()
 
-  getScrollRow: -> @scrollRow
-  setScrollRow: (@scrollRow) ->
+  # Experimental: Scroll the editor such that the given screen row is at the
+  # top of the visible area.
+  setFirstVisibleScreenRow: (screenRow, fromView) ->
```
aleclarson: This may need `NaN` detection. Otherwise we get a nasty `displayBuffer` error. May require closing and re-opening tab after calling `editor.setFirstVisibleScreenRow(NaN)`.

PR #9449, opened 2015-11-05
OS X File Icons

Michael-Thibault:
I've created file icons in the ICNS format for use in on OS X, as well as updating the atom-Info.plist file's icon paths. These icons cover every file type that was listed in the Info.plist.

Edit: Here's a preview of some of the icons, as they appear full-size:

Full Size Icons

Here's how they appear in a list or in column view:

32x32 and smaller

Note that for other platforms, it is possible to easily convert these from ICNS to ICO or other formats by extracting the PNGs using [IConvert Icons](https://iconverticons.com/online/#).

For anyone wondering, [here's the PSD on Dropbox](https://www.dropbox.com/s/z521d6bby4izrgo/Fortran.psd?dl=0) for the Fortran 1024x1024 image. It can easily be used to add other file extensions/formats in future.

One other note: I followed the TextMate 2 icons' general style, preferring all-caps when possible, but switching to camelcase whenever it's a part of the language's name (e.g. BibTeX or OCaml).


thomasjo:
Can you explain the reasoning behind this change? Now we have lots of duplicated files instead of a single file, and the end result should be identical.

Michael-Thibault:
If you haven't, please look at the icns files themselves. Each is specific to the file format(s) it represents, so the end result, as you say, would actually be somewhat different. That means that the icons provided in Finder should then be appropriate to the file type, which can be handy. Granted QuickLook often replaces the icon with something else, but for those times when it doesn't, I figured might as well have them.

thomasjo:
Aha, I see now. They've got the file extension embedded. I'm fine with this then, but since this somewhat touches upon the branding, I'll have to defer this to the core team. Also would be great to get @simurai's eyes on this.

/cc @atom/feedback @nathansobo

lee-dohm:
For people who aren't on OS X but still would like to review this, would you include a screen shot of a few sample icons?

50Wliu:
The build task will have to be updated to copy over all of the new icons.

Michael-Thibault:
@lee-dohm 
> For people who aren't on OS X but still would like to review this, would you include a screen shot of a few sample icons?

Absolutely. I've just updated the original post with appropriate images and a link to the original PSD used to get all the other images for good measure.

@50Wliu
> The build task will have to be updated to copy over all of the new icons.

Good point. If I stick all the files in a new subfolder it should be easy to do with one line. (I could/probably should add the PSDs and PNGs too for good measure. They wouldn't need to go into the build, obviously, but could prove useful for anyone making other icons in future.)

lee-dohm:
Thanks for updating the original post!

Yeah, I think it would be good to include the source material so that they could be added upon in the future when necessary.

Michael-Thibault:
I've moved the ICNS files into a subfolder, and added PNG and PSD versions of each file. I also updated the build task, but I'd really appreciate someone checking my work on that (and offering a suggestion if I got it wrong, which is entirely possible, or even likely >.> ).

simurai:
@Michael-Thibault Thanks for your PR. Although it's just altering the existing file icon by adding the file extensions, it probably still falls under "GitHub can't accept brand-related pull requests". See https://github.com/atom/atom/pull/3089#issuecomment-50504224

Something else I'm wondering... by how much would it increase the download size of Atom? When compressing the `icns` folder the size is about `50mb`. Currently Atom's download size is `67.5mb`.

benogle:
We've decided that we'd like to keep the simplicity and space savings of the current implementation. These icons look nice, but they dont add any more information to the file as they just repeat the extension in the icon.

nathansobo:
Appreciate the work you put into it however. Thanks.

simurai:
Something that could still be improved is adding more file extensions to https://github.com/atom/atom/blob/master/resources/mac/atom-Info.plist. Then they would also show the `file.icns` icon. Maybe at least for all languages that get bundled with Atom.

I'll make a new issue about it.

PR #9452, opened 2015-11-05
:arrow_up: language-go@0.40.0

svanharmelen:


50Wliu:
I'm pretty confident that this won't break specs (I don't think Atom has any core specs that rely on language-go), so :ship:.

PR #9468, opened 2015-11-06
Make LanguageMode::unfoldAll faster

as-cii:
_All the benchmarks were performed by opening `src/text-editor.coffee` on startup_.

In the previous implementation we were constantly querying `TextBuffer` for folds whenever the grammar on `TextEditor` changed:

screen shot 2015-11-06 at 14 29 22

Indeed, when changing the grammar we issue a call to `LanguageMode::unfoldAll` which tries to get a fold for each row in the buffer. 

screen shot 2015-11-06 at 14 33 40

This happens also on startup because grammars are loaded after a `TextEditor` instance gets created. Moreover, user's perceived responsiveness gets worse as more files (or larger files) need to be opened/restored because we change the grammar for every instance of `TextEditor` as soon as the packages' grammars are loaded and a new grammar is recognized.
### Solution

With this PR, instead of issuing a `findMarker` call to `TextBuffer` for each buffer row, we just get all the folds intersecting the whole buffer and destroy them in reverse order. This is how the first graph looks like after the improvements:

screen shot 2015-11-06 at 13 22 04

Calling `::unfoldAll` is almost instantaneous now and for a single ~3000 lines file this saves almost `150ms`. :racehorse: :racehorse: :racehorse: 

In the same code path there's still some work we could do to make loading settings faster, as it would seem like we spend a considerable amount of time in just setting keys/values in `Config` (at least during startup). Nevertheless, this should result in a significant responsiveness increase when users boot up Atom with some open files.

### DIFF
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -465,8 +465,8 @@ class DisplayBuffer extends Model
 
     folds
 
-  # Returns all the folds that contain the given row range.
-  foldsContainingBufferRowRange: (startRow, endRow) ->
+  # Returns all the folds that intersect the given row range.
+  foldsIntersectingBufferRowRange: (startRow, endRow) ->
     @foldForMarker(marker) for marker in @findFoldMarkers(intersectsRowRange: [startRow, endRow])
 
   # Public: Given a buffer row, this returns folds that include it.
--- a/src/display-buffer.coffee
+++ b/src/display-buffer.coffee
@@ -465,6 +465,10 @@ class DisplayBuffer extends Model
 
     folds
 
+  # Returns all the folds that contain the given row range.
+  foldsContainingBufferRowRange: (startRow, endRow) ->
+    @foldForMarker(marker) for marker in @findFoldMarkers(intersectsRowRange: [startRow, endRow])
+
   # Public: Given a buffer row, this returns folds that include it.
   #
   #
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -97,8 +97,8 @@ class LanguageMode
 
   # Unfolds all the foldable lines in the buffer.
   unfoldAll: ->
-    for row in [@buffer.getLastRow()..0] by -1
-      fold.destroy() for fold in @editor.displayBuffer.foldsStartingAtBufferRow(row)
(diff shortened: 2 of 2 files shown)


src/display-buffer.coffee
```diff
@@ -465,6 +465,10 @@ class DisplayBuffer extends Model
 
     folds
 
+  # Returns all the folds that contain the given row range.
+  foldsContainingBufferRowRange: (startRow, endRow) ->
```
nathansobo: I think this should be named `foldsIntersectingBufferRowRange`, based on the query.

src/language-mode.coffee
```diff
@@ -97,8 +97,8 @@ class LanguageMode
 
   # Unfolds all the foldable lines in the buffer.
   unfoldAll: ->
-    for row in [@buffer.getLastRow()..0] by -1
-      fold.destroy() for fold in @editor.displayBuffer.foldsStartingAtBufferRow(row)
+    for fold in @editor.displayBuffer.foldsContainingBufferRowRange(0, @buffer.getLastRow()) by -1
```
nathansobo: Great job here. Querying every single line is insane. Definitely leftover from the dark ages.

nathansobo:
Having a little trouble following the flame graphs but the code changes are a no-brainer. Nice work. One naming-related comment, then :ship:.

src/display-buffer.coffee
as-cii: Yeah, I thought to this as well but this _seems_ inconsistent with the method below (`::foldsContainingBufferRow`) which performs a similar query. What do you think?

nathansobo: Makes sense, but in that case, if a fold intersects a single row it also contains it. In this case it's a range of rows, and we know these folds don't fully contain the range, otherwise there would only be one big fold. I'd be down renaming the other method for clarity/consistent if it isn't being used by packages (it's private but still might as well avoid breakages just in case).

as-cii: A cursory search showed up that there's some reference to `foldsContainingBufferRow` in packages (some typescript header files), so I think it's just better to rename the new one only (7d97241). 

Thanks! :sparkles:

as-cii:
:heart:

PR #9469, opened 2015-11-06
Async repo take 2

thedaniel:
After implementing almost everything else in #9213, I ran into some
sporadic spec failures that I was having trouble reasoning about. This
is an attempt to start over with the class _beginning_ with the
functionality that went flaky, to better understand how it is tested
and whether it was being broken by an interaction with something else
in the new implementation, broken intrinsically, or had buggy tests.

This may supersede the other PR, or it may not. Let’s see what we learn.

Also, the spec is written in es6 and not coffee in this branch, and I do plan to take Nathan’s advice and update it to use async/await rather than the jasmine queue.

cc @joshaber


src/git-repository-async.js
```diff
@@ -0,0 +1,202 @@
+'use babel'
+
+const path = require('path')
```
vramana: @thedaniel Why not use es6 import syntax ??

thedaniel:
Moving back over to the original PR, @joshaber found the underlying cause of the flakiness there.

thedaniel:
False alarm, reopening

joshaber:
@thedaniel https://github.com/atom/atom/pull/9213 is back on track, I believe. The tests pass locally and on Travis. The build's just timing on.

controversial:
@thedaniel Why was this closed the second time?

thedaniel:
@controversial because this was an alternate strategy for the same goal as #9213, which was ultimately merged, making this unnecessary.

PR #9476, opened 2015-11-06
Always set devMode to true for spec windows

nathansobo:
Otherwise dev mode isn't enabled when running specs interactively, which can have unintended consequences.


thomasjo:
This looks good to me! I can't think of any (sensible) reason why we would not want to do this.

maxbrunsfeld:
:+1:

PR #9480, opened 2015-11-07
Adjust icon size for small buttons

### LINKED ISSUE #8412: Icons in small buttons increase button height
The style guide shows regular buttons with icons via `<button class='btn btn-primary icon icon-cloud-download inline-block-tight'>Install</button>`:

screen shot 2015-08-17 at 10 53 00 am

However, small buttons styled the same way are too tall, making text not vertically aligned, because `.icon` has a static height and font-size of 16px `<button class='btn btn-sm btn-primary icon icon-cloud-download inline-block-tight'>Install</button>`:

screen shot 2015-08-17 at 10 53 26 am

Icons should shrink when inside a `.btn-sm`, something like this:

screen shot 2015-08-17 at 10 56 29 am

simurai:
This PR makes the icons in small and extra small buttons a bit smaller too. Fixes #8412

Before:

screen shot 2015-11-07 at 9 52 47 pm

After:

screen shot 2015-11-07 at 9 50 42 pm

Screenshots are with core styles only (no theme).

### DIFF
--- a/static/buttons.less
+++ b/static/buttons.less
@@ -32,7 +32,7 @@
   height: auto;
   line-height: 1.3em;
   &.icon:before {
-    font-size: 12px;
+    font-size: @font-size - 2px;
   }
 }
 .btn.btn-sm,
@@ -41,7 +41,7 @@
   height: auto;
   line-height: 1.3em;
   &.icon:before {
-    font-size: 14px;
+    font-size: @font-size + 1px;
   }
 }
 .btn.btn-lg,
@@ -50,6 +50,9 @@
   padding: @component-padding - 2px  @component-padding + 2px;
   height: auto;
   line-height: 1.3em;
+  &.icon:before {
+    font-size: @font-size + 6px;
+  }
 }
 
 .btn-group > .btn {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Nice!

thomasjo:
LGTM! :ship:

static/buttons.less
```diff
@@ -31,12 +31,18 @@
   font-size: @font-size - 2px;
   height: auto;
   line-height: 1.3em;
+  &.icon:before {
+    font-size: 12px;
```
ssorallen: Should this be relative to `@font-size` like the `font-size` itself?

ssorallen:
Awesome! Curious what you think about reusing `@font-size`, but otherwise looks good to me.

static/buttons.less
simurai: Ya, it could. Maybe also for `sm` and `lg` buttons. And keep default buttons fixed at `16px` so they look sharp on non-retina.

simurai:
Changed it so that all icon sizes are based on `@font-size`, except for the default button. There is will always be `16px` to make it look sharp on non-retina.

PR #9483, opened 2015-11-07
Show non breaking space if showInvisibles is true

### LINKED ISSUE #8080: [feature request] Display non-breaking space character when showing invisibles.
It would be really nice if the editor displayed an "invisibles" symbol for the character UNICODE NON BREAK SPACE `U+00A0` as it most often is added by mistake and gives pretty perplexing error messages. It is currently visually indistinguishable from the normal space. 

image

(You can't see it but it's there...)

cakecatz:
Fixes #8080 

screen_shot_2015-11-07_at_11_38_18_pm

### DIFF
--- a/src/tokenized-line.coffee
+++ b/src/tokenized-line.coffee
@@ -10,6 +10,7 @@ RepeatedSpaceRegex = /[ ]/g
 CommentScopeRegex = /(\b|\.)comment/
 TabCharCode = 9
 SpaceCharCode = 32
+NonBreakingSpaceCharCode = 160
 SpaceString = ' '
 TabStringsByLength = {
   1: ' '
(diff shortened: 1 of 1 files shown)


src/tokenized-line.coffee
```diff
@@ -97,7 +98,7 @@ class TokenizedLine
         tokenOffset = 0
 
       # split out leading soft tabs
-      else if charCode is SpaceCharCode
+      else if charCode is SpaceCharCode or charCode is NonBreakingSpaceCharCode
```
jeancroy: Can you use two different cases ?

issue  #8080 basically state that non breaking space are undesirable as they break code
so maybe we should use a different visual than plain space.

cakecatz: Sure.

PR #9484, opened 2015-11-07
Remove subpixel-font-scaling option

winstliu:
This option was removed in Electron v0.33.9, and we're on v0.34.0 now (soon to be v0.34.3).


nathansobo:
:+1:

PR #9485, opened 2015-11-07
Disable throttling in test windows via WebAudio hack

nathansobo:
Throttling in test windows is annoying when tests run interactively, and _seems_ to cause issues with animation frames not firing even in headless tests, though it’s difficult to fully confirm since this issue is intermittent.

Refs atom/electron#3225
/cc @zcbenz

### DIFF
--- a/src/initialize-test-window.coffee
+++ b/src/initialize-test-window.coffee
@@ -57,6 +57,12 @@ module.exports = ({blobStore}) ->
 
     document.title = "Spec Suite"
 
+    # Avoid throttling of test window by playing silence
+    context = new AudioContext()
+    source = context.createBufferSource()
+    source.connect(context.destination)
+    source.start(0)
+
     testRunner = require(testRunnerPath)
     legacyTestRunner = require(legacyTestRunnerPath)
     buildDefaultApplicationDelegate = -> new ApplicationDelegate()


nathansobo:
@thedaniel I'd be interested to see if this helps with some of the flaky async tests you've been dealing with.

benogle:
Cray, I feel like this comment needs a link.

nathansobo:
Based on some comments from Cheng I'm not sure it's even working now, though I thought it was in my testing. Going to test again and maybe revert it.

nathansobo:
Nope, this actually seems to be working. @zcbenz is there a good URL I can link to from this code to explain this?

zcbenz:
@nathansobo 

The code that controls the behavior:
https://code.google.com/p/chromium/codesearch#chromium/src/content/browser/renderer_host/render_process_host_impl.cc&sq=package:chromium&type=cs&l=2368
The issue that reports a bug on this:
https://code.google.com/p/chromium/issues/detail?id=491895

nathansobo:
Just linked back to here in ab6f26368b84dce01683bb74e7c865dac324a979.

src/initialize-test-window.coffee
```diff
@@ -57,6 +57,12 @@ module.exports = ({blobStore}) ->
 
     document.title = "Spec Suite"
 
+    # Avoid throttling of test window by playing silence
+    context = new AudioContext()
+    source = context.createBufferSource()
+    source.connect(context.destination)
+    source.start(0)
+
     testRunner = require(testRunnerPath)
```
francoislaberge: @nathansobo Any idea if this trick works as far back as 0.29.2? We're having odd throttling behavior (networking and cpu) currently but can't upgrade our Electron because we're having build issues for our ARM environment.

We're working on getting unblocked on upgrading, but in the meantime any techniques to disable all throttling behavior in chromium is desired. Related, any advice on making Chrome only perform at maximum would be helpful. We are adding our own resource throttling at a higher level in our custom linux build.

nathansobo: There was a `page-visibility` option that stopped working, requiring this. I referenced it  at https://github.com/atom/electron/issues/3225. Not sure when that was introduced though.

nathansobo: I should note that as of Chrome 47, the visibility doesn't appear to be working to keep our tests running in the foreground. A new `--disable-background-timer-throttling` flag was [added](https://bugs.chromium.org/p/chromium/issues/detail?id=485425) in Chrome 49, but that's still in beta.

zcbenz: @nathansobo `disable-background-timer-throttling` only disables throttling for timers, for `requestAnimationFrame` it doesn't work. I tried lots of ways but still can not find the correct way to make `requestAnimationFrame` work for background pages, we probably need help from Chrome developers.

nathansobo: So the original flag to disable background rendering doesn't work anymore?

zcbenz: It is still working for timers like `setInterval`, but not for `requestAnimationFrame`.

PR #9490, opened 2015-11-08
Fix activatePackages

aki77:
When a package having `activationHooks` exists, `did-activate-initial-packages` is not called.


nathansobo:
Thanks @aki77!

aki77:
Thanks!

PR #9508, opened 2015-11-10
Faster settings load

as-cii:
When activating a package, we load its settings asynchronously. Then, as soon as Node is done with IO, it calls us back with the content of those files, at which point we call `ScopedProperties::activate` to transfer those settings into our `Config` instance.

When setting a key in our `Config`, we emit an event that notifies attached listeners that something has just changed: the implications of this are pretty important, as there is a remarkable amount of operations we need to do in order to notify those clients, such as comparing the newly set key with the previous one, etc. Although the current implementation could be somewhat improved in terms of performance, it handles the configuration semantics pretty smartly, which means that, whenever setting something in the configuration, it's better to emit as little events as possible. For this reason, before the changes introduced with this PR, we were [wrapping packages activation with a call to `Config::transact`](https://github.com/atom/atom/blob/46665c1f8b86664f9431d66cfd9e291bdc0689cf/src/package-manager.coffee#L419-L423).

The problem with this approach, however, is that it doesn't play well with the asynchronous nature of loading the configuration, as the transaction will end as soon as we've finished iterating over the packages to activate (i.e. _before_ `ScopedProperties::activate` has the chance to execute). This means we emit way more events than we really should, paying an extra amount of time during startup:

before

The breakdown here is the following:
- Parsing CSON: `~ 39.2ms`
- Activating Grammars: `~ 71.0ms`
- Activating Settings: `122.4ms` + `37.9ms`
- **Total: 294.9ms**

For this reason, we have introduced a `Config::beginTransaction` and `Config::endTransaction`, which are aimed to handle the asynchronous scenario in a more performant manner:

after

We can observe a a `~ 125ms` speedup, as this approach allows us to emit just a minimal amount of events, thereby dramatically reducing the overhead of loading setting files.

For the future, I think it's also worth taking a look at how we might improve CSON (which, for bundled packages, is actually JSON) parsing and grammars activation, in order to reduce those remaining `169.7ms` to the minimum.

/cc: @atom/feedback @nathansobo @maxbrunsfeld


alexandernst:
I'm not sure how JSON is parsed, but maybe this could help (async JSON parse): http://azimi.me/2015/07/30/non-blocking-async-json-parse.html

joshaber:
Awesome :metal:

nathansobo:
Great work. I'm a bit concerned about people calling these methods incorrectly and leaving `Config` in a bad state. We could potentially add a somewhat more abstract API that takes a promise, but it's probably not worth it for now.

nathansobo:
:ship: at will.

as-cii:
> I'm a bit concerned about people calling these methods incorrectly and leaving Config in a bad state. We could potentially add a somewhat more abstract API that takes a promise, but it's probably not worth it for now.

I :heart: the idea of using a `Promise` to express the idea of an asynchronous transaction. Good call! :sparkles: 

I feel like we shouldn't let the user pass a `Promise`, though, because it could have already been executed before beginning a transaction. In this scenario I believe it would be probably better to let the users of this API pass in a `callback` function that _returns_ a `Promise`, so that it's more explicit that the action being executed should be wrapped in its totality (i.e. from start to finish) within the transaction. Moreover, this would make it symmetrical to the current `::transact` method, where we have full control on the moment in which the action gets executed. Of course, people could still pass in a callback where they return an already instantiated/executing `Promise`, but I feel like a `callback` tells this _execution control_ story a bit more explicitly.

For the time being, I'll go ahead and :ship: this as-is. I will open a new PR where I'll address these concerns.

Thanks everyone for the feedback! :pray: :fireworks:

PR #9513, opened 2015-11-10
Prioritize file type matching over first line matching in grammar scoring system

### LINKED ISSUE #2734: PHP and HTML Syntax Coloring issue
I am using Elementary OS and version 0.107.0, and when I write PHP and this happens:
screenshot from 2014-06-24 18 59 36

The surrounding HTML code is highlighted, but PHP isn't.

It's even worse when you have a string in the PHP code, because the whole following text (even after `?>`) is seen as html value and has the same color.

I believe this package should be responsible for this: https://atom.io/packages/language-php , please tell me if this is the wrong place, but i'm not sure if it is a general atom issue.

Please note that javascript and CSS are highlighted when written inside of the HTML.

Ingramz:
Fixes atom/atom#2734

This is a pull request version of https://github.com/atom/atom/issues/2734#issuecomment-155230824 in order to work together on solving the currently unfair grammar scoring system. The main problem with the current system is that it gives a higher score to HTML grammar for any file that begins with HTML doctype, but the file might be using PHP or some other (templating) language that utilizes HTML syntax (with a different file type, eg `php`).

@nathansobo suggested adding an option to fall back to using the old system in case the new one breaks something, but I am not sure how to do that exactly. Also another question would be whether this transition period is needed at all, because all it "breaks" right now is grammar selection for files, for which there is a known file type, but the first line is matched from some other grammar. A fix for that case would be just adding the file type via user configuration, to override the known type from the other language.

### DIFF
--- a/src/grammar-registry.coffee
+++ b/src/grammar-registry.coffee
@@ -46,7 +46,7 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
     contents = fs.readFileSync(filePath, 'utf8') if not contents? and fs.isFileSync(filePath)
 
     score = @getGrammarPathScore(grammar, filePath)
-    if not grammar.bundledPackage
+    if score > 0 and not grammar.bundledPackage
       score += 0.25
     if @grammarMatchesContents(grammar, contents)
       score += 0.125
--- a/src/grammar-registry.coffee
+++ b/src/grammar-registry.coffee
@@ -36,21 +36,21 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
       if score > highestScore or not bestMatch?
         bestMatch = grammar
         highestScore = score
-      else if score is highestScore and bestMatch?.bundledPackage
-        bestMatch = grammar unless grammar.bundledPackage
     bestMatch
 
   # Extended: Returns a {Number} representing how well the grammar matches the
   # `filePath` and `contents`.
   getGrammarScore: (grammar, filePath, contents) ->
+    return Infinity if @grammarOverrideForPath(filePath) is grammar.scopeName
+
     contents = fs.readFileSync(filePath, 'utf8') if not contents? and fs.isFileSync(filePath)
 
-    if @grammarOverrideForPath(filePath) is grammar.scopeName
-      2 + (filePath?.length ? 0)
-    else if @grammarMatchesContents(grammar, contents)
-      1 + (filePath?.length ? 0)
-    else
-      @getGrammarPathScore(grammar, filePath)
+    score = @getGrammarPathScore(grammar, filePath)
+    if not grammar.bundledPackage
+      score += 0.25
+    if @grammarMatchesContents(grammar, contents)
+      score += 0.125
+    score
 
   getGrammarPathScore: (grammar, filePath) ->
     return -1 unless filePath
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "language-make": "0.20.0",
     "language-mustache": "0.13.0",
     "language-objective-c": "0.15.0",
-    "language-perl": "0.30.0",
+    "language-perl": "0.31.0",
     "language-php": "0.34.0",
     "language-property-list": "0.8.0",
     "language-python": "0.41.0",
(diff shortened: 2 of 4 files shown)


nathansobo:
@Ingramz Thanks. I buy your argument that the existing override structure is enough. Let me review this...

nathansobo:
@Ingramz If you look at Travis, It looks like some tests related to the change code are failing in core specs. It looks like we're asserting exactly the opposite of the new behavior, so you'll need to update that. Also, you should add a test that asserts that the path-based selection wins over the first line regexp.

src/grammar-registry.coffee
```diff
@@ -36,21 +36,21 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
...
         bestMatch = grammar
         highestScore = score
-      else if score is highestScore and bestMatch?.bundledPackage
-        bestMatch = grammar unless grammar.bundledPackage
     bestMatch
 
   # Extended: Returns a {Number} representing how well the grammar matches the
   # `filePath` and `contents`.
   getGrammarScore: (grammar, filePath, contents) ->
+    return Infinity if @grammarOverrideForPath(filePath) is grammar.scopeName
```
nathansobo: This is good.

src/grammar-registry.coffee
```diff
@@ -36,21 +36,21 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
...
 
-    if @grammarOverrideForPath(filePath) is grammar.scopeName
-      2 + (filePath?.length ? 0)
-    else if @grammarMatchesContents(grammar, contents)
-      1 + (filePath?.length ? 0)
-    else
-      @getGrammarPathScore(grammar, filePath)
+    score = @getGrammarPathScore(grammar, filePath)
+    if score > 0 and not grammar.bundledPackage
+      score += 0.25
```
nathansobo: Curious... why these numbers? Just arbitrary?

nathansobo:
You had commented with a long description of this logic elsewhere. Could you transfer that into the body of this PR?

Ingramz:
This will return Infinity as a score for the overridden grammar, which in my opinion is a little more elegant than 2 + length of file path.

The path scoring system is already good as-is. Longer match will get a higher score, plus 0.5 bias for user defined file types. I find it very clever that the user can select between two grammars by adding the file type in question to the user defined list, even if it already was defined in grammar so one of the two grammars always gets priority with the same matched longest file type.

From the other post.

Also there is a bias of 0.25 so that a non-bundle package would win over a bundled package, if their matches were the same length (ergo the file types matched are the same), this is the same as the code that was on lines 39/40. However one thing to note - if we get -1 as a file type score for all grammars and there is a non-bundle grammar, then that would win over all bundled grammars with a score of -0.75, it seems that this oddity exists also in original code. I think that this bias shouldn't be applied if the path score itself is -1. <- update: this has been fixed in PR

Finally a bias for the case where the first line of the file matches a file, which is currently the weakest - 0.125. This should help in case the two grammars were equal in the previous cases. Setting this with the smallest bias makes more sense to me as it should be used only as a last resort, which works really well for files which the type is unknown.

nathansobo:
Cool, thanks for the explanation. @maxbrunsfeld do you think you could take a look at this too? Maybe we could screen share and just try to think through the implications of changing this and be sure I'm not missing anything.

Ingramz:
The numbers come from the sequence 1/pow(2, n), for which the sum for n=1...infinity is 1, so if we keep going smaller, we hopefully never reach 1 (so let's make sure we're not summing _all_ of the possible terms in that sequence). This way the small increments never override the file type's priority. This also applies recursively, sum for n=2..infinity is 1/2 etc, so all increments smaller than 0.5 will also stay smaller than 0.5, which helps with the priority.

nathansobo:
Cool deal about the numbers. Very nice thought process. Now just I need to see tests of new functionality and a green build.

Ingramz:
@nathansobo seems like the tests are passing now.

nathansobo:
Okay. I'm a bit nervous that I'm not thinking through some implication but this is why we have a beta phase. Merging. Thanks @Ingramz for the thoughtful implementation here. Awesome work.

PR #9514, opened 2015-11-10
Guard against not loaded packages when updating disabled keymaps

kevinsawicki:
Previously an uncaught exception was thrown if `core.packagesWithKeymapsDisabled` was set to an array that included a package name that wasn't loaded, such as a package that was uninstalled.

Closes atom/settings-view#695


benogle:
> Closes #695

Is this right?

kevinsawicki:
> Is this right?

Right number, wrong repo, the issue was reported to settings-view, I've updated the link, thanks for pointing it out.

PR #9524, opened 2015-11-11
Introduce Config::transactAsync

as-cii:
Refs: #9508 

This method is supposed to wrap asynchronous code within a `Config` transaction. It accepts a `callback` which must return a `Promise`: when this resolves or gets rejected (or the `callback` throws an error) we complete the transaction.

I am not super happy with the name I've chosen for this API, as it would have probably been better to follow Node's conventions (i.e. rename `::transact` to `::transactSync` and name the new one `::transact`): unfortunately that method name was already taken and it's part of the public API.

Another thought was to include this feature right inside the existing `::transact` method: although possible, the resulting method would need to conform to two different contracts based on the return type of `callback`, which feels somewhat misleading for clients. In the end having a separate method felt somewhat cleaner.

Feedback super welcome! :bow:  

/cc: @nathansobo @joshaber @atom/feedback

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -687,7 +687,7 @@ class Config
   #
   # Returns a {Promise} that is either resolved or rejected according to the
   # `{Promise}` returned by `callback`. If `callback` throws an error, a
-  # rejected {Promise} will be returned.
+  # rejected {Promise} will be returned instead.
   transactAsync: (callback) ->
     @beginTransaction()
     endTransaction = (resolveOrReject) => (args...) =>
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -677,10 +677,14 @@ class Config
     finally
       @endTransaction()
 
-  # Extended: Suppress calls to handler functions registered with
-  # {::onDidChange} and {::observe} for the duration of the {Promise} returned
-  # by `callback`. After the {Promise} is either resolved or rejected, handlers
-  # will be called once if the value for their key-path has changed.
+  ###
+  Section: Internal methods used by core
+  ###
+
+  # Private: Suppress calls to handler functions registered with {::onDidChange}
+  # and {::observe} for the duration of the {Promise} returned by `callback`.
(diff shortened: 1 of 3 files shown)


nathansobo:
Yeah, someday I think we can fix these warts in our API, but for now I think 2 methods is a good idea.

src/config.coffee
```diff
@@ -677,6 +677,34 @@ class Config
...
+  # * `callback` {Function} that returns a {Promise}, which will be executed
+  #   while suppressing calls to handlers.
+  #
+  # Returns a {Promise} that is either resolved or rejected according to the
+  # `{Promise}` returned by `callback`. If `callback` throws an error, a
+  # rejected {Promise} will be returned instead.
+  transactAsync: (callback) ->
+    @beginTransaction()
+    try
+      endTransaction = (fn) => (args...) =>
```
nathansobo: Are the multiple arguments needed here? Seems like you only call `resolve` or `reject`.

as-cii: Yeah, I guess so: this is a function that returns a function (i.e. curried), so when called with `endTransaction(resolve)`, for instance, will return a function that:

* Ends the transaction.
* Calls `resolve` with all the passed arguments.

When used like `result.then(endTransaction(resolve))`, `resolve` will be invoked with all the arguments available in `then`'s callback. This means that a potential client could use this API like the following:

```coffee
atom.config.transactAsync(-> Promise.resolve(1)).then((n) -> console.log("Output: #{n}"))
#=> Output: 1
```

Although we're not using it in our production code, it seemed reasonable to follow the `Promise` API semantics when I implemented this. What do you think?

nathansobo: Ah, I missed the currying. So long as you test it I'm cool having the functionality.

as-cii: :white_check_mark: I tested this behavior in c6cb37a :+1:

I am going to :shipit: as soon as the build gets :green_heart:

thomasjo:
Nice!

PR #8557, opened 2015-08-28
Remap pane:show-item-9 to show last tab

SantanMaddi:
Activate the last tab on pane on cmd / ctrl + 9 shortcut

Closes #6609

PR #9528, opened 2015-11-11
Remap pane:show-item-9 to activate last pane item

### LINKED ISSUE #6609: Request: map cmd/alt + 9 to show last (rightmost) tab
I wanted to see what people thought about remapping this key combination to instead to to the last tab open in a pane. Currently, it is set to [view the ninth tab](https://github.com/atom/atom/blob/master/keymaps/darwin.cson#L103): `'cmd-9': 'pane:show-item-9'`

My reasons for suggesting are as follows:
* I believe it is easier to have a dedicated key to always go to this specific position, as compared to counting how many tabs from the left your desired tab is
* if someone has more than nine tabs open, a keymapping to go directly to the end would be beneficial
* Atom shares many existing keyboard shortcuts with Chrome, and Chrome uses this key combination to navigate to the last tab

Curious to hear your thoughts on it. Thanks.

efatsi:
Activate the last tab on pane on cmd / ctrl + 9 shortcut.

Closes #6609.

Noticed work for this was started on #8557 but never finished. This adds the functionality and accompanying spec.

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -168,6 +168,15 @@ describe "Pane", ->
       pane.activateNextItem()
       expect(pane.getActiveItem()).toBe item1
 
+  describe "::activateLastItem()", ->
+    it "sets the active item to the last item", ->
+      pane = new Pane(paneParams(items: [new Item("A"), new Item("B"), new Item("C")]))
+      [item1, item2, item3] = pane.getItems()
+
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
:ok_hand: Great! Thanks @efatsi.

djpowers:
Awesome! Thanks for picking this up @efatsi.

PR #9575, opened 2015-11-13
Set initialLoadComplete to true only in test mode

### LINKED ISSUE #9189: Scrollbar not getting styled correctly on Windows and Linux
I just pulled the master branch and got this odd looking scrollbar in UI. 
screenshot from 2015-10-17 15 08 12

as-cii:
Fixes #9189 

After the changes in #8968, `ThemeManager::isInitialLoadComplete` [was always being set to true](https://github.com/atom/atom/blob/master/src/atom-environment.coffee#L185-L186), whereas in the past we used to set it only after all the theme packages had been activated.

How does this relate to scrollbars? When restoring the previous state we attach the workspace along with the needed `TextEditorElement`s. When `TextEditorComponent` gets instantiated, it uses `isInitialLoadComplete` to decide whether it should listen to `ThemeManager::onDidChangeActiveThemes`, which is what triggers `::refreshScrollbars` after e.g. `one-dark-ui` loads.

Since `::isInitialLoadComplete` was true before the themes had the chance to actually being loaded, we weren't refreshing scrollbars at all, thus not allowing webkit to apply the needed stylesheets (when we refresh the scrollbars we explicitly trigger a reflow to let Chrome handle stylesheets correctly).

This PR changes the existing behavior so that we mark the initial theme load as complete only in test mode (through a configurable option).

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -117,7 +117,7 @@ class AtomEnvironment extends Model
 
   # Call .loadOrCreate instead
   constructor: (params={}) ->
-    {@blobStore, @applicationDelegate, @window, @document, configDirPath, @enablePersistence} = params
+    {@blobStore, @applicationDelegate, @window, @document, configDirPath, @enablePersistence, loadBaseStylesheetsOnly} = params
 
     @state = {version: @constructor.version}
 
@@ -183,7 +183,7 @@ class AtomEnvironment extends Model
 
     @themes.loadBaseStylesheets()
     @initialStyleElements = @styles.getSnapshot()
-    @themes.initialLoadComplete = true
+    @themes.initialLoadComplete = true if loadBaseStylesheetsOnly
     @setBodyPlatformClass()
 
     @stylesElement = @styles.buildStylesElement()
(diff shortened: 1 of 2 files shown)


src/atom-environment.coffee
```diff
@@ -117,7 +117,7 @@ class AtomEnvironment extends Model
 
   # Call .loadOrCreate instead
   constructor: (params={}) ->
-    {@blobStore, @applicationDelegate, @window, @document, configDirPath, @enablePersistence} = params
+    {@blobStore, @applicationDelegate, @window, @document, configDirPath, @enablePersistence, loadBaseStylesheetsOnly} = params
```
as-cii: :art: Any ideas on how to call this parameter? I am not super happy with `loadBaseStylesheetsOnly` but it's probably okay.

as-cii:
As a side note, this can be reproduced in OS X through the following steps:

1. Open a buffer
2. Make sure that it is scrollable horizontally/vertically
3. Make sure you've set "Always" in your `System Settings -> General -> Show Scrollbars`  
4. Close Atom so that it saves the current state
5. Open Atom so that it reloads the previous state
6. :boom:

src/atom-environment.coffee
nathansobo: Maybe `onlyLoadBaseStyleSheets`? "Style sheet" should definitely be two words. After that, :ship:

PR #9585, opened 2015-11-13
Add apm.cmd

winstliu:
Allows Windows users who build Atom to conveniently type `apm` in the console just as they would for `atom`.

/cc @raelyard @joefitzgerald @kevinsawicki


kevinsawicki:
:+1:

PR #9592, opened 2015-11-14
upgrade semver

fscherwi:
Upgrade semver to newest version!

It only has small changes and no significant API changes!

But it has an new license, from BSD to ISC!

See: [changes](https://github.com/npm/node-semver/compare/v4.3.3...v5.0.3)


lee-dohm:
We need to be careful about updating this dependency because various parts of Atom all have to use the same exact version to unify behavior. We're not ready to update the semver library right now.

Thanks for contributing!

Issue #863, opened 2013-09-20
Add option to have cmd-w close the window when there are no editors open


matthewmccullough:
When I have multiple buffers open in tabs and a tree view, my past editors (TextMate, Sublime) have conditioned me to expect that a repeated ⌘+w close each buffer's tab, then close the frame and its tree list after the last buffer has been closed.  Atom does not behave this way, and I'm curious if this is by design.

It is worth noting that Brackets does also hold the frame open after repeated ⌘+w presses. Is this similarly by design in Atom? Must I switch fingers to ⌘+⇧+w to close the frame?

fulldecent:
For me, this is intuitive:

**A file is open and you press Cmd-W:** This closes the last file, and focus returns to the file tree so you can up/down/enter to open another file.

**No file is open and you press Cmd-W:** The full MDI window closes.

ryanseys:
Seems strange that Cmd+W doesn't close the window on Mac because this is just regular convention across all applications. This felt very much "broken" when I realized I couldn't close the window with Cmd+W.

mcandre:
+1

cthielen:
+1

jacekkopecky:
would it be easily possible to make background-tips receive keyboard events? Then we could bind `cmd-w` to `window:close` for an easy workaround.

jacekkopecky:
Oh, how is the package `the-closer` not mentioned here? It addresses this very issue.

eparejatobes:
@jacekkopecky it was! https://github.com/atom/atom/issues/863#issuecomment-47467785 works just great and it makes atom usable as your `$EDITOR`

jacekkopecky:
Ah, @izuzak neglected to mention the name of his package when introducing it. 8-)

PaulCapestany:
+1

df-sean:
+1 *** Atom is literally the *only* app on my Mac that doesn't behave this way.

smilingpoplar:
+1

Gerald09:
+1

groue:
@benogle, you have changed the title of the issue, by adding "option". I don't know if it's a good idea. On OSX, such behavior has no meaning being an option, since it is a de facto standard.

benogle:
Yes, an option is important because this behavior has been in since the beginning. There will inevitably be new people who are frustrated by `cmd-w` closing the window. Whenever we change behavior [we see this happen](https://github.com/atom/atom/issues/6605), so we need to minimize frustration.

[...]

PR #9607, opened 2015-11-15
Close empty active window when Close command is given

DouweM:
On OS X, Command-W is expected to close the current window, unless the current window has tabs (Safari, Chrome, Finder), in which case it will close the current tab, until only one is left, in which case it closes the entire window (and thus also the tab).

Currently in Atom, Command-W will close every pane item in the current pane, then the current pane, and so on until every pane is empty, which makes sense since pane items and panes ~ tabs. However, when only the empty root pane is left, which the user experiences as there being no panes left at all, hitting Command-W will not close the window, but will instead have no effect at all. To actually close the window, one needs to either manually hit the red close button in the upper left corner or the window, choose "Close Window" in the File menu or use the Cmd-Shift-W shortcut, all of which break the OS X UX contract.

With this PR, Command-W will behave exactly as it does right now, with one exception: when only the root pane is left, Command-W closes the entire window.

Hitting Command-W with only one pane item ~ tab open will NOT close the window, which is inconsistent with tabbed web browsing but makes more sense for a text editor.

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -21,7 +21,6 @@ module.exports =
       followSymlinks:
         type: 'boolean'
         default: true
-        title: 'Follow symlinks'
         description: 'Follow symbolic links when searching files and when opening files with the fuzzy finder.'
       disabledPackages:
         type: 'array'
@@ -54,7 +53,12 @@ module.exports =
       destroyEmptyPanes:
         type: 'boolean'
         default: true
-        description: 'When the last item of a pane is removed, remove that pane as well.'
+        title: 'Remove Empty Panes'
+        description: 'When the last tab of a pane is closed, remove that pane as well.'
+      closeEmptyWindows:
+        type: 'boolean'
+        default: true
+        description: 'When a window with no open tabs or panes is given the \'Close Tab\' command, close that window.'
       fileEncoding:
         description: 'Default character set encoding to use when reading and writing files.'
         type: 'string'
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -1438,11 +1438,11 @@ describe "Workspace", ->
         save = -> atom.workspace.saveActivePaneItem()
         expect(save).toThrow()
 
-  describe "::destroyActivePaneItemOrEmptyPane", ->
+  describe "::closeActivePaneItemOrEmptyPaneOrWindow", ->
     beforeEach ->
       waitsForPromise -> atom.workspace.open()
 
-    it "closes the active pane item until all that remains is a single empty pane", ->
+    it "closes the active pane item, or the active pane if it is empty, or the current window if there is only the empty root pane", ->
       atom.config.set('core.destroyEmptyPanes', false)
 
       pane1 = atom.workspace.getActivePane()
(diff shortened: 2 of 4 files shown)


mnquintana:
Would you mind adding specs for this? Thanks!

DouweM:
@mnquintana I amended the existing test: https://github.com/atom/atom/pull/9607/files#diff-013fa854db95a703465103ef15c4b0edR1470. Is that enough?

benogle:
Refs https://github.com/atom/atom/issues/863

nathansobo:
Great job @DouweM! :zap: We'll see if everyone yells at us for changing it to work this way now. :grimacing:

nathansobo:
Code looks good, restarting the failed Travis build.

benogle:
IMO, there should be an option to preserve the current behavior

benogle:
@DouweM Can you add a setting for this? Something like `closeEmptyWindowOnCoreClose`? @nathansobo feel free to come up with a better name.

DouweM:
@benogle May I ask why? The old behavior has never made sense to me or any other Mac/Linux Atom user I know (I don't know any Windows Atom users). You'll always have people complain about changes in behavior, but I think this is a change for the better and I'm seriously wary about adding configuration options just because people don't like change.

benogle:
For behavior changes like this, we need to give people who rely on the current behavior an out. Similar to the recent changes in fuzzy-finder matching algorithm, if it's clear that almost no one wants the old behavior, we can pull the option in the future.

> I think this is a change for the better

I dont necessarily think this is a change for the better. It's just different. I for one really love the current behavior, commands do explicitly one thing: close an item or close a window and do not change based on state. I can mash cmd-w 100 times, and know that I wont accidentally close the window. We honestly have had little to no complaining about the current behavior in the last year or so. I'm inclined to think that many people are now used to it.

nathansobo:
Agreed on adding the option. There's no reason to have a bike shed debate on something that's so easy to customize. It originally worked this way and we changed it based on internal feedback from GitHubbers, so I think there's a good argument that it's going to be controversial. Let's avoid the controversy and let people design their own experience.

DouweM:
@nathansobo Fair enough, I will add an option!

DouweM:
@nathansobo Option was added!

nathansobo:
Kicking off another Travis build.

nathansobo:
Our Travis build is timing out. Think it may have to do with Blink throttling our timers and animation frames on CI. I'm running a build locally.

nathansobo:
Okay, I figured it out. @DouweM locally I'm seeing the spec you added actually closing the test window, which prevents us from informing the browser process that the tests completed. Could you move the `spyOn(atom, 'close')` into your before each and see if that gets us a :green_apple: Travis build?

nathansobo:
Actually, nevermind, I'll just move it and merge.

nathansobo:
:metro:  Thanks for your help @DouweM. :metro:

DouweM:
@nathansobo :heart:

src/config-schema.coffee
```diff
@@ -21,7 +21,6 @@ module.exports =
       followSymlinks:
         type: 'boolean'
         default: true
-        title: 'Follow symlinks'
```
matiasgarciaisaia: Was this deletion intentional?

DouweM: It was, te default title is the key (`followSymlinks`) capitalized, which would be `Follow Symlinks`. The downcase `s` was inconsistent with the other titles.

Issue #3935, opened 2014-10-22
Feature: OSX - Recent Projects on right click in Dock


habermann24:
Other editors (e.g. Textmate) give a list of recently used projects/files when right clicking on the icon in the dock.

See these screenshots:

![screen shot 2014-10-23 at 00 36 31](https://cloud.githubusercontent.com/assets/7480/4745559/f54196de-5a3b-11e4-905c-7cdd97f11dd9.png)
![screen shot 2014-10-23 at 00 36 09](https://cloud.githubusercontent.com/assets/7480/4745560/f542b992-5a3b-11e4-8a03-95bc82685ef8.png)

It would be great if Atom would do the same.

I was not sure what package to post this issue in. Maybe it belongs to atom-shell.
There needs to be some code that can hook into native OSX APIs, since that is how i believe that it works.

aequitas:
I would love to see this feature, as I use it frequently with Textmate.

janbaykara:
+1

brianmoran:
Is there a way to Clear closed projects? When I right click on the dock icon, it shows projects that I have deleted or no longer want to show up.

grahm:
needs to be a preference setting to turn this off

thecodeassassin:
Take a look at this:

http://tca0.nl/Alp

The recent items setting is turned to None in my general settings and yet my atom context menu is flooded with recent items.

starriv:
> Is there a way to Clear closed projects? When I right click on the dock icon, it shows projects that I have deleted or no longer want to show up.

I also encountered the same problem

Issue #4241, opened 2014-11-19
Feature request: Most recently used (MRU) files and folders


ghost:
I would like to be able to find files or folders that I have recently opened in previous editing sessions. On Windows this is usually done under the File menu. For example: File > Recent Files > (displays a sub menu of top 10 most recently used files in descending order).

File

> Recent Files
>        > (top 10 most recent files)
> Recent Folders
>        > (top 10 most recent folders)

mnquintana:
:+1: This would be an awesome addition. I think the issue title and tags should be updated though because this isn't actually a Windows-only enhancement – all OSes should get some kind of File > Open Recent menu item and Windows and OS X should both get recent files / directories listings on their taskbar / dock right-click menus, respectively.

(It might've been lost in that whole paragraph, but to reiterate **the `windows` tag should be removed and the issue title should be updated to remove "Windows"**)

50Wliu:
I think I do have a "Most recent" on Ubuntu when I do File-->Open Folder.

mnquintana:
@50Wilu Isn't that "Recently Used" Place in the File > Open Folder dialog on Ubuntu just "recently used" by *any* program, not just Atom?

50Wliu:
Ah, yes, you're correct.

jerone:
There's a package for this https://atom.io/packages/open-recent

Zren:
Hey, author of open-recent here. The one thing I was wondering about if I wanted to merge the plugin into atom itself is if using `localStorage` for serializing the recent paths is okay.

Using atom's config to serialize the paths doesnt seem right since it updates the file every time you open something and might annoy people whove synched/use version control on their `config.cson`.

lee-dohm:
This feature was recently added for OS X and I believe Windows too in https://github.com/atom/atom/pull/9609.

Leaving this open and labeling it for Linux.

pflannery:
> This feature was recently added for OS X and I believe Windows too in #9609.

This isn't working on Windows 10 in Atom 1.6.2.

50Wliu:
@pflannery Can you try Atom 1.7.0?

pflannery:
@50Wliu ok. I've just tried 1.7.0 and I've managed to get files to appear but only when using the "File->Open File" menu.

![image](https://cloud.githubusercontent.com/assets/1727302/14233130/64304524-f9b7-11e5-8ed3-dba648ac46ba.png)

[...]

PR #9609, opened 2015-11-15
 Add opened files and folders to recent documents menu

DouweM:
Title says it all!

screen shot 2015-11-15 at 20 51 23

Closes https://github.com/atom/atom/issues/3935

### DIFF
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -109,6 +109,9 @@ class ApplicationDelegate
   setRepresentedFilename: (filename) ->
     ipc.send("call-window-method", "setRepresentedFilename", filename)
 
+  addRecentDocument: (filename) ->
+    ipc.send("add-recent-document", filename)
+
   setRepresentedDirectoryPaths: (paths) ->
     loadSettings = getWindowLoadSettings()
     loadSettings['initialPaths'] = paths
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -885,9 +885,9 @@ class AtomEnvironment extends Model
         else
           @project.addPath(pathToOpen)
 
-      if fs.isDirectorySync(pathToOpen)
-        @applicationDelegate.addRecentDocument(pathToOpen)
-      else
+      @applicationDelegate.addRecentDocument(pathToOpen)
+
+      unless fs.isDirectorySync(pathToOpen)
         @workspace?.open(pathToOpen, {initialLine, initialColumn})
 
     return
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -885,7 +885,9 @@ class AtomEnvironment extends Model
         else
           @project.addPath(pathToOpen)
 
-      unless fs.isDirectorySync(pathToOpen)
+      if fs.isDirectorySync(pathToOpen)
+        @applicationDelegate.addRecentDocument(pathToOpen)
+      else
         @workspace?.open(pathToOpen, {initialLine, initialColumn})
 
     return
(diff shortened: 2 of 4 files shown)


50Wliu:
Will this also work on Windows and Linux?

alexandernst:
This uses https://github.com/atom/electron/blob/master/docs/api/app.md#appaddrecentdocumentpath-os-x-windows , so it's not actually *in* Atom itself, but a helper (kind of) for the OS. I'd vote :+1: for this only if this is implemented for Linux too (inside Atom's menu).

DouweM:
@50Wliu It uses an Electron helper method that works on OS X and Windows, but is currently unimplemented for Linux.

@alexandernst It would be nice if `app.addRecentDocument` were supported on Linux, but that would need to happen in Electron, not Atom. Once this PR is accepted into Atom, if someone ever implements `addRecentDocument` for Linux, it will start working for Atom immediately. I feel like that's outside the scope of this particular PR, though.

50Wliu:
Sounds good to me then.

alexandernst:
This is not (easily) possible in Linux. xdg might have something like a "standart" for such a thing, but it's up to each DE how this standart should be followed.

On the other hand, IMHO, this isn't as simple as having the recent files list inside Atom itself, in a submenu of "File".

On Windows, you'd have to go to the start menu, then find Atom, then click on the right arrow: win

On Linux, well, it depends on the DE as I said, but in KDE 5 for example that would be main menu -> recent documents -> scroll down to the "recent docs" section.

Now compare how easier would it be if Atom had that same feature in it's own menu: menu

alexandernst:
All that said, I'm totally :+1: for this feature, it's just that this should be extended to Linux support too, which means a submenu in Atom's main menu.

DouweM:
@alexandernst That makes sense, many OS X apps have an "Open Recent" menu under File too, in addition to the menu under the dock icon. 

It would be nice to get that menu on all platforms, but this sounds like something that would need to be done in Electron using C++ since it should use the same API that already exists for the OS X/Windows-specific bindings and be available to all Electron apps, which is unfortunately outside of my ability at this point.

mnquintana:
@alexandernst That would definitely be great to have, but I think that can be a separate PR. Adding a menu item for recent documents is being tracked here: https://github.com/atom/atom/issues/4241

DouweM:
@mnquintana Anything I can do to help this PR along?

kevinsawicki:
>  Anything I can do to help this PR along?

@DouweM looks like there are conflicts, mind updating them?

DouweM:
@kevinsawicki Done.

kevinsawicki:
One last quick question before merging, do you know if this list grows infinitely? Does the OS handle capping it at a reasonable number like a dozen or so?

DouweM:
@kevinsawicki Yep, that's handled by OS X, which is the only OS for which `app.addRecentDocument` is currently implemented:

```objc++
void Browser::AddRecentDocument(const base::FilePath& path) {
  NSURL* u = [NSURL fileURLWithPath:base::mac::FilePathToNSString(path)];
  [[NSDocumentController sharedDocumentController] noteNewRecentDocumentURL:u];
}
```

See [`noteNewRecentDocumentURL:`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSDocumentController_Class/#//apple_ref/occ/instm/NSDocumentController/noteNewRecentDocumentURL:) and [`maximumRecentDocumentCount`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSDocumentController_Class/#//apple_ref/occ/instp/NSDocumentController/maximumRecentDocumentCount).

kevinsawicki:
> Yep, that's handled by OS X

Awesome, thanks for this :+1:

DouweM:
@kevinsawicki Thanks for merging!

PR #9620, opened 2015-11-16
Use em-dash in window title and exclude app name on OS X

DouweM:
On OS X, the convention is to use the em-dash as the segment separator in window titles, and to not include the app name: (The two items at the top are window titles)

## OLD

screen shot 2015-11-16 at 12 02 36
screen shot 2015-11-16 at 12 03 00
## NEW

screen shot 2015-11-16 at 11 57 31
screen shot 2015-11-16 at 11 57 59

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -76,7 +76,7 @@ describe "Workspace", ->
           expect(editor4.getCursorScreenPosition()).toEqual [2, 4]
 
           expect(atom.workspace.getActiveTextEditor().getPath()).toBe editor3.getPath()
-          expect(document.title).toBe "#{path.basename(editor3.getLongTitle())} - #{atom.project.getPaths()[0]} - Atom"
+          expect(document.title).toMatch ///^#{path.basename(editor3.getLongTitle())}\ \u2014\ #{atom.project.getPaths()[0]}///
 
     describe "where there are no open panes or editors", ->
       it "constructs the view with no open editors", ->
@@ -732,7 +732,7 @@ describe "Workspace", ->
     describe "when the project has no path", ->
       it "sets the title to 'untitled'", ->
         atom.project.setPaths([])
-        expect(document.title).toBe 'untitled - Atom'
+        expect(document.title).toMatch ///^untitled///
 
     describe "when the project has a path", ->
       beforeEach ->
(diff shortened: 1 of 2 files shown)


spec/workspace-spec.coffee
```diff
@@ -76,7 +76,7 @@ describe "Workspace", ->
           expect(editor4.getCursorScreenPosition()).toEqual [2, 4]
 
           expect(atom.workspace.getActiveTextEditor().getPath()).toBe editor3.getPath()
-          expect(document.title).toBe "#{path.basename(editor3.getPath())} - #{atom.project.getPaths()[0]} - Atom"
+          expect(document.title).toMatch /^#{path.basename(editor3.getPath())} \u2014 #{atom.project.getPaths()[0]}/
```
50Wliu: You'll need to use an extended regex here (`///`) because the normal one doesn't support string interpolation.

DouweM:
@50Wliu Anything I can do to help this PR along?

thomasjo:
LGTM! /cc @atom/feedback

lee-dohm:
Looks good to me too :+1:

spec/workspace-spec.coffee
```diff
@@ -76,7 +76,7 @@ describe "Workspace", ->
           expect(editor4.getCursorScreenPosition()).toEqual [2, 4]
 
           expect(atom.workspace.getActiveTextEditor().getPath()).toBe editor3.getPath()
-          expect(document.title).toBe "#{path.basename(editor3.getPath())} - #{atom.project.getPaths()[0]} - Atom"
+          expect(document.title).toMatch ///^#{path.basename(editor3.getPath())}\ \u2014\ #{atom.project.getPaths()[0]}///
```
mnquintana: Super minor, but any reason why you're using multi-line regexes here instead of a standard regex?

DouweM: @mnquintana It was pointed out to me by @50Wliu that string interpolation doesn't work in standard regexes.

mnquintana: Ah yeah good point, neeeever mind :laughing:

spec/workspace-spec.coffee
```diff
@@ -777,7 +777,7 @@ describe "Workspace", ->
         })
         workspace2.deserialize(atom.workspace.serialize(), atom.deserializers)
         item = atom.workspace.getActivePaneItem()
-        expect(document.title).toBe "#{item.getTitle()} - #{atom.project.getPaths()[0]} - Atom"
+        expect(document.title).toMatch ///^#{item.getTitle()}\ \u2014\ #{atom.project.getPaths()[0]}///
```
maxbrunsfeld: Why `toMatch` here instead of asserting an exact string like before?

lee-dohm: The end of the title is variable based on platform. These new tests only cover the platform-agnostic portions, from what I saw anyway.

DouweM: That's exactly it. On Windows/Linux, the title will end in "— Atom", on OS X it will not. We're verifying that the title _starts_ with the expected item title / project path combination.

kevinsawicki:
Cool, thanks so much for this :+1:

Issue #9294, opened 2015-10-26
Initial window load UI can be improved


50Wliu:
![initial load](https://cloud.githubusercontent.com/assets/2766036/10740328/0df89310-7bf8-11e5-8c91-9f01974ef968.png)
There are a couple of issues here:
1. The only menu that appears is `Atom`.  If you click on that the commands that normally go in the `Atom` menu on Mac appear, but with the `Cmd-` part of the keyboard shortcut stripped off (eg `Cmd-R` --> `R`).
2. A large portion of the screen is white, which seems related to #8847.
3. As briefly mentioned in #8293, the window is showing up in an unmaximized state, but with dimensions that correspond to a maximized state.

Please note that I could only reproduce the first two issues on Windows 10 x64 using Atom 1.0.19, and not Atom 1.2.0-dev-9b46f7c.

mnquintana:
> 1. The only menu that appears is Atom. If you click on that the commands that normally go in the Atom menu on Mac appear, but with the Cmd- part of the keyboard shortcut stripped off (eg Cmd-R --> R).

Definitely :+1: on this - I've seen a lot of coworkers attempt to click that Atom menu before Atom actually finishes loading, only to be confused about the result (on both Mac and Windows).

Zireael07:
Win 7 x64 here, seen 1 & 3 but not 2.

hiddewie:
3. When the window was maximized before it was closed last session, a new window will show up with  maximized dimensions (but not maximized) and become maximized on load of the content of the window.
(Windows 10 x64)

50Wliu:
Some updates since 1.0.19.  Note that all of the images below were taken on Atom 1.5.0-dev-dbf7bed (the Electron 0.36 branch).
![initial-screen](https://cloud.githubusercontent.com/assets/2766036/12060809/82395626-af45-11e5-9286-d8290d29525c.png)
This is now the initial (unmaximized) screen that appears when starting Atom.  The shortcut keys in the `Atom` menu no longer appear though.

Wait a bit more for Atom to load, and this (maximized) screen appears:
![still-loading](https://cloud.githubusercontent.com/assets/2766036/12060791/09a297d6-af45-11e5-8732-d8deec9bd500.png)
Notice how everything is scaled down appropriately to match the available area.  Not sure why the rest of that whitespace is there.  Also note how no syntax highlighting has been applied yet.  This image in particular seems to be new behavior from Electron 0.35 and up, and isn't present on master yet.

[...]

PR #9627, opened 2015-11-16
Update to Electron 0.36

### LINKED ISSUE #666: Cannot delete file or folder (atom 1.3.0)
See subject. Answering `Move to Trash` to the delete dialog results in nothing.

Works fine in atom 1.2.4.

Windows 7 (zip file install)

This seems to be an Electron issue (since v0.34.3): atom/electron#3656 (`shell.moveItemToTrash`)

---
*[Moved from: atom/atom#9909]*
*[edit] ai, 666.. :P ;)*

### LINKED ISSUE #9534: Error "auto-updater is not implemented on this platform" on Linux
I'm getting the following message in `~/.atom/nohup.out` or in the terminal when I start atom with the `--foreground` option. Atom seems to function properly, though. I use Atom v1.3.0-dev-62f4f32 on Debian 9.

```
$ cat ~/.atom/nohup.out 
auto-updater is not implemented on this platform
Error: auto-updater is not implemented on this platform
    at Object.<anonymous> (/usr/share/atom/resources/atom.asar/browser/api/lib/auto-updater.js:11:13)
    at Object.<anonymous> (/usr/share/atom/resources/atom.asar/browser/api/lib/auto-updater.js:14:4)
    at Module._compile (module.js:434:26)
    at Object.defineProperty.value [as .js] (/usr/share/atom/resources/app.asar/src/compile-cache.js:190:21)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at AutoUpdateManager.module.exports.AutoUpdateManager.setupAutoUpdater (/usr/share/atom/resources/app.asar/src/browser/auto-update-manager.js:51:23)
    at /usr/share/atom/resources/app.asar/src/browser/auto-update-manager.js:42:24
App load time: 256ms
```

### LINKED ISSUE #9616: Atom hangs when docked to left or right side of screen - Windows
After moving Atom to the left side of my screen using windows shortcut (Win+left arrow key) or (Win+right arrow key), Atom hangs. However, when I move it around or make it full screen. Atom starts to work again.

I've tried this in safe mode and I experience the same problem.
image

Edit: I've found an easy way to reproduce it somewhat consistently. If you open one atom instance and move it to the right or left side and then open a second atom instance and move it to the other side. The second atom instance will hang. As shown in the picture below.
image

### LINKED ISSUE #9901: No response when running atom at one side of desktop (Windows)
I have a problem with atom (version 1.2.4) on my Windows 10 laptop. In Windows we can split the view and run applications side by side using Win+LEFT or Win+RIGHT key. When we run atom at one side, even if we keep atom at front, typing keyboard won't have any response until the window is restored or maximized.

Here is the way to reproduce the problem.

1. Open atom, maximize the window, type some content.
2. Type Win+RIGHT to set atom to the right of the screen.
3. Keep atom at front, type keyboard (no response this time).
4. Maximize the window, the content will now show up.

winstliu:
Figured I'd get this started and get some more experience with the core atom code, as I'm pretty sure lots of things will need to be changed this time.

https://github.com/atom/electron/releases/tag/v0.35.0
https://github.com/atom/electron/releases/tag/v0.36.0

Fixes #9534
Fixes #9616
Fixes #9901
Fixes atom/tree-view#666
# TODO:
- [x] Fix core specs failing almost instantly with no output
- [x] Fix specs randomly hanging - at least, they didn't hang in the last build.
- [x] atom/tabs#238
- [x] Fix notifications specs that rely on `executeJavaScriptInDevTools`...kinda?  Not sure if I actually fixed the underlying deprecation or not.
- [x] Attempt to update to Electron v0.36.0 - in progress
  - [x] Fix failing specs on Mac (https://github.com/atom/atom/pull/9627#issuecomment-166756294)
  - [x] Remove `ipc.sendSync()` calls on startup
- [ ] Test for and fix regressions
# Regressions:
- [x] Opening a new tab no longer works
- [x] Adding a project folder does nothing through the context menu or file menu
- [x] Adding a project folder through the keyboard shortcut/command palette opens up the folder chooser but doesn't actually add the new folder
- [x] Open Your ... menu items are all broken, even when executed through the command palette
- [x] Open File does not actually open the chosen file
- [x] Help -> About Atom does nothing
- [x] Open Settings does not work through the file menu
- [x] ~~Opening Settings or any other file that uses a URI will cause a fake folder to appear in the Tree View (eg `config`, `init-script`) and `Settings - atom://config` or similar to show up in the titlebar.  Reproduce by using a new window.~~ - also an issue on master (tested on f6da4a9139bd08b4b89e85203caf20a4af72d195)
- [x] On 0.36 - menu mousedown effect is not removed when switching menus by hovering over to a new one (atom/electron#3964)
- [x] Closing the window when Atom is launched using `atom --foreground` yields an error:

```
Error: Object has been destroyed
    at Error (native)
    at AtomWindow.module.exports.AtomWindow.getLoadSettings (file:///C:/Users/user/Documents/GitHub/atom/src/browser/atom-window.coffee:91:8)
    at AtomApplication.module.exports.AtomApplication.saveState (file:///C:/Users/user/Documents/GitHub/atom/src/browser/atom-application.coffee:452:34)
    at EventEmitter.<anonymous> (file:///C:/Users/user/Documents/GitHub/atom/src/browser/atom-application.coffee:211:8)
    at emitOne (events.js:77:13)
    at EventEmitter.emit (events.js:169:7)
    at AtomApplication.module.exports.AtomApplication.removeWindow (file:///C:/Users/user/Documents/GitHub/atom/src/browser/atom-application.coffee:104:13)
    at BrowserWindow.<anonymous> (file:///C:/Users/user/Documents/GitHub/atom/src/browser/atom-window.coffee:126:30)
    at emitOne (events.js:82:20)
    at BrowserWindow.emit (events.js:169:7)
```

The error occurs on the following line:

``` coffee
getLoadSettings: ->
    if @browserWindow.webContents? and not @browserWindow.webContents.isLoading()
      hash = url.parse(@browserWindow.webContents.getURL()).hash.substr(1)  # Here!
      JSON.parse(decodeURIComponent(hash))
```


50Wliu:
Is there any way to get more output from Travis?  Core specs are (surprisingly) passing but I'm sure there's a bunch of deprecated Electron APIs being used.

50Wliu:
And...specs are failing to run.  Not sure how to proceed from here, though I suspect I probably messed up the `ipc` renaming.

50Wliu:
Ran the tests locally and it looks like we're hitting atom/electron#3484:
```
Insufficient number of arguments.
TypeError: Insufficient number of arguments.
    at TypeError (native)
    at parseCommandLine (C:\Program Files\Atom\resources\app.asar\src\browser\main.js:167:100)
    at start (C:\Program Files\Atom\resources\app.asar\src\browser\main.js:32:12)
    at Object.<anonymous> (C:\Program Files\Atom\resources\app.asar\src\browser\main.js:209:3)
    at Object.<anonymous> (C:\Program Files\Atom\resources\app.asar\src\browser\main.js:211:4)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Object.<anonymous> (C:\Program Files\Atom\resources\atom.asar\browser\lib\init.js:130:10)
```
The build then proceeds to hang.  I'll see if I can figure out what got updated; if not it looks like we'll just have to wait for Electron 0.35.1.

src/browser/atom-application.coffee
```diff
@@ -9,7 +9,7 @@ app = require 'app'
 dialog = require 'dialog'
 shell = require 'shell'
 fs = require 'fs-plus'
-ipc = require 'ipc'
+ipc = require 'ipc-main'
```
50Wliu: I *think* this file uses `ipc-main`, but if someone could double-check, that would be great.

50Wliu:
Specs are still timing out :disappointed:.  I can run them fine locally (albeit with a whole bunch of errors), so it might be a Mac/Linux-only issue.

All the deprecations in Atom core should be fixed though.

src/browser/atom-portable.coffee
```diff
@@ -1,6 +1,6 @@
 fs = require 'fs-plus'
 path = require 'path'
-ipc = require 'ipc'
+ipc = require 'ipc-main'
```
50Wliu: I'm thoroughly confused by this right now.  Tried both `ipc-main` and `ipc-renderer`, but both throw "Cannot find module" errors.

(It's probably also why core specs finish in 13 secs and fail)

50Wliu: And...switching it back to ipc-renderer makes it work again.  Really hope there isn't a race condition of some sort going on here.

50Wliu:
@zcbenz A few questions:
1. Requiring either `ipc-main` or `ipc-renderer` in `atom-application.coffee` and `atom-portable.coffee` throws "Cannot find module" errors.  Is there any way around that?
2. `webContents.executeJavaScriptInDevTools` doesn't seem to be a method.  In fact, looking at v0.34.x, I can't even find `executeJavaScriptInDevTools`, just `executeJavaScript`.  The 0.35 upgrade appears to have broken `executeJavaScriptInDevTools` as well with multiple `cannot read "apply" of undefined` errors, though it's only supposed to be a deprecation.
3. Not a question, but I'm pretty sure the deprecation message for `getHomeDir()` didn't tell me to use `getPath('home')`, just `getPath()`.

glangford:
> The 0.35 upgrade appears to have broken executeJavaScriptInDevTools as well with multiple cannot read "apply" of undefined errors, though it's only supposed to be a deprecation.

Spelunking to learn...It looks like `executeJavaScriptInDevTools` may have broke here:
<https://github.com/atom/electron/commit/91c4ed26fc9e0ed2792abd2bf594dd0d2e4bb336>

glangford:
@zcbenz From atom/electron@91c4ed2 it looks like the original code had

```coffee
BrowserWindow::executeJavaScriptInDevTools = (code) -> @devToolsWebContents?.executeJavaScript code
```

then deprecated to 
```coffee
deprecate.member BrowserWindow, 'executeJavaScriptInDevTools', 'webContents'
```

which on the surface looks wrong, since `deprecate.member` forwards the method to `webContents` not `devToolsWebContents`, and the original code ran a different method `executeJavaScript`.

Another possible issue is:
```coffee
BrowserWindow::getPageTitle = -> @webContents.getTitle()
```

was deprecated with
```coffee
deprecate.member BrowserWindow, 'getPageTitle', 'webContents'
```
where I'm guessing that deprecate.rename is required.

zcbenz:
@glangford Those deprecation wrappers were wrong, thanks for the notice!

Announcement:
I have an upgrade for npm-latest and node-latest on my laptop with a broken screen, it actually was a LOT of work, so if anyone else hasn't already done this I could fetch it while we're still upgrading API's

glangford:
@50Wliu  -
> ...All of these seem like issues caused by the ipc renaming, but I can't get any errors or deprecation notices to appear

Fix wrong deprecation wrappers of BrowserWindow
<https://github.com/atom/electron/commit/932cd92bf6dded7756196254f5d7593fc5d2e14d>

Improve the deprecation notice for ipc module
<https://github.com/atom/electron/commit/9c62be8fc998856fb6f5dbf032f85cacd3818659>

package.json
```diff
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "0.34.4",
+  "electronVersion": "0.35.1",
```
glangford: 0.35.2 has some relevant changes.

50Wliu: Thanks for the heads up, just rebased and updated!

50Wliu:
Specs still can't find `ipc-main` in `atom-portable.coffee` for whatever reason.

glangford:
> Specs still can't find ipc-main in atom-portable.coffee for whatever reason.

:confused: 
I went hunting for clues and saw this in the updated api docs:

`const ipcMain = require('electron').ipcMain;`

Refs
<https://github.com/atom/electron/blob/master/docs/api/ipc-main.md>
<https://github.com/atom/electron/pull/3415>
<https://github.com/atom/electron/commit/eac2e7cc614d4e98e2828921f63d55f7b5e2c85c>

50Wliu:
Aha.  I missed atom/electron#3415 so I assumed Atom was special in that it didn't need the `require('electron')` part.  Surprised Electron didn't deprecate the old way.

Testing now...

50Wliu:
Upon further inspection moving to `require('electron').x` won't solve anything, since that maps to the same thing as `require 'x'`.  The issue still remains that `ipc-main`/`ipcMain` can't be found.

Also, since we don't have electron as a node module, attempting to do the former will result in `Error: Cannot find module 'electron'`.  That's probably an easy fix (adding `electron-prebuilt` as a dependency or something?) but I feel like I'm getting way out of my comfort zone here.

glangford:
Does it help to back off to 0.35.1 to see if the same inability to find the module occurs?

50Wliu:
> Does it help to back off to 0.35.1 to see if the same inability to find the module occurs?

It's been happening since the initial upgrade to 0.35.0.

glangford:
Ok, one more kick at this problem:

The electron specs for ipc include this:
```coffee
{ipcRenderer, remote} = require 'electron'
{ipcMain, BrowserWindow} = remote.require 'electron'
```

<https://github.com/atom/electron/blob/master/spec/api-ipc-spec.coffee#L5>

I don't understand the use of `remote.require` here, just pointing out the different incantation in case it helps.

50Wliu:
Nope, `remote` isn't defined.  Explicitly requiring it through `remote = require 'remote'` (which is what other portions of the code does) fails with `Error: cannot find module 'remote'`.

zcbenz:
I have updated the PR with the new styles of Electron API.

50Wliu:
You're magical @zcbenz.  `require 'electron'` just refused to work for me.  Thanks so much! :star2: :100:

50Wliu:
For the record I know what I did wrong now: I was launching through `atom --dev` without running `script\build` beforehand, so Electron wasn't the correct version (I think).

50Wliu:
The error that occurs when closing a window (as documented in the PR body) is pretty interesting.  It seems to be happening because the browser window is being destroyed before we access it in `getLoadSettings` to save the serialization state in `application.json`.  I've made sure that the sequence of close events hasn't changed between 0.34.5 and 0.35.4 - it's always `close` -> `closed` -> `before-quit`.

Now the interesting part is that we're calling `getLoadSettings` in `before-quit`, which happens *after* `closed` and when the Electron documentation says to stop using the window object.  I'm not sure why continuing to access the destroyed window didn't throw an "object is destroyed" error in previous versions, but now it is.

In any case the error seems to be harmless as Atom is still able to access the copy of the window object to save the open tabs.

src/browser/atom-window.coffee
```diff
@@ -121,6 +119,9 @@ class AtomWindow
         false
 
   handleEvents: ->
+    @browserWindow.on 'close', ->
+      global.atomApplication.saveState(false)
```
50Wliu: I'm not sure if this breaks anything (as it gets called when every window is closed now, not just when the entire app is closing), but it did solve the uncaught exception.

50Wliu:
I've been running this for around a week now and haven't noticed any new regressions, but it would be great to have a few more :eyes: :mag:: /cc @atom/feedback.

50Wliu:
Oh, and also: I tried looking for any bundled packages that were using the old way of requiring Electron modules, but it's pretty likely that I missed some.

pleonex:
What about trying to update to electron 0.36.0?
That version seems to fix atom/electron#2645 which is crashing my Atom so many times

50Wliu:
1 failing core spec:
```
>> TextEditorComponent
>>   overlay decoration rendering
>>     positioning the overlay when near the edge of the editor
>>       it slides horizontally left when near the right edge on #win32 and #darwin
>>         Expected '387px' to be '375px'.
>>           Error: Expected '387px' to be '375px'.
>>           at /Users/travis/build/atom/atom/spec/text-editor-component-spec.js:1979:36
>>           at [object Generator].next (native)
>>           at step (/Users/travis/build/atom/atom/spec/text-editor-component-spec.js:1:411)
>>         Expected '399px' to be '375px'.
>>           Error: Expected '399px' to be '375px'.
>>           at /Users/travis/build/atom/atom/spec/text-editor-component-spec.js:1986:36
>>           at [object Generator].next (native)
>>           at step (/Users/travis/build/atom/atom/spec/text-editor-component-spec.js:1:411)
```

and one failing ac+ spec:
```
>> Autocomplete Manager
>>   when an external provider is registered
>>     when autocomplete-plus.suggestionListFollows is 'Word'
>>       it displays the suggestion list with a negative margin to align the prefix with the word-container
>>         Expected '-144px' to be '-143px'.
>>           at [object Object].<anonymous> (/Users/travis/build/atom/atom/node_modules/autocomplete-plus/spec/autocomplete-manager-integration-spec.coffee:969:55)
>>           at Immediate._onImmediate (/Users/travis/build/atom/atom/node_modules/autocomplete-plus/spec/spec-helper.coffee:26:9)
```

Seems like we have some alignment issues with 0.36.

50Wliu:
@pleonex working on it!  If it turns out to require too much work, I'll revert the changes and try updating to 0.36 in a different PR.

pleonex:
@50Wliu Thanks for giving a try!

50Wliu:
Just tried reproducing the failures locally and both specs pass (which in itself is pretty surprising since almost a quarter of Atom's specs fail on my laptop).
Looks like someone who has a Mac will have to try.

qazwx:
With electron 0.36 there's an issue:
If autoHideMenuBar is true and Alt is pressed, the menu bar remains hidden. (same in safe-mode, Xubuntu 15.04)

50Wliu:
I'm *pretty* sure someone here left a comment about menus not toggling after the <kbd>Alt</kbd> key was pressed, but either my memory is a bit confuzzled or the comment got deleted.  At any rate that bug has been fixed in Electron 0.36.2 which we're now on.

qazwx:
Sorry, I deleted my comment when I saw the bug fix on electron 0.36.2

zcbenz:
There is a small issue on the upgrade, since v0.36.1 calling `ipc.sendSync` won't pump the message loop anymore, so when waiting for the main process to respond the rendering of page would be stopped. There are a few places in Atom sending synchronous messages on startup, which would result in a small period of white screen now, so for users they may feel the loading is becoming slower. The actual startup time is not affected.

We need to remove the `sendSync` calls on startup to fix the small period of white screen, or we can delay the showing of window until it is fully loaded.

50Wliu:
@zcbenz I just did a global find for `sendSync` but that returned 0 results.  Can you please give a few examples of where we're calling that?

zcbenz:
@50Wliu `remote` module uses `sendSync` under the hood, so every call of a remote object would send a synchronous message.

I just did a profile of startup code, there is no synchronous message sent on startup anymore, apparently recent performance improvements have solve this. :+1:

maxbrunsfeld:
OK, tests are now passing on Mac CI. The problem was that `remote.getCurrentWindow().setSize` seems to no longer take immediate effect; there is a test that now needs to poll until `window.innerWidth` eventually changes to reflect the new size.

One remaining problem, which may be related, is that on startup, content starts to render before the window size is fully restored, which causes some visual glitches.

@zcbenz We still are [using the `remote` module](https://github.com/atom/atom/blob/f4be23049d537b78d58a9125d71c28b3f5b2b6f8/src/application-delegate.coffee#L29) on startup, and its behavior seems to have changed such that it returns before the window's size changes.

I'm guessing we need to switch to using an async `send`, and figure out our own way of waiting until the resize is finished. Do you have any suggestions for this?

50Wliu:
> content starts to render before the window size is fully restored, which causes some visual glitches.

I reported that in a different issue (#9294)...looks like I forgot to also post the picture over here.
startup

I've reverted the Electron version to 0.36.4 since 0.36.5 crashes Atom on startup for me.  Dump file is attached...
[electron-0.36.5-crash-windows.zip](https://github.com/atom/atom/files/105779/electron-0.36.5-crash-windows.zip)

50Wliu:
@maxbrunsfeld Travis timed out with the new await in https://travis-ci.org/atom/atom/jobs/105022601.  Seems like there's a chance that it's a bit flaky.

maxbrunsfeld:
> Travis timed out with the new await

Thanks for the heads up @50Wliu. I think that the flakiness may be due to the window size being 1 pixel off from the requested size, or something like that. I've changed the spec so that now it `awaits` the resize message being processed, instead of polling for an exact window size. Maybe this will help.

maxbrunsfeld:
OK, I think I have addressed the rendering artifacts on startup, and this is ready for testing.

maxbrunsfeld:
/cc @amytruong: Some areas that could use some testing

* [ ] Startup (Mac, Linux, Windows)
  * Any brief visual oddities like the one @50Wliu captured [above](https://github.com/atom/atom/pull/9627#issuecomment-175266702)
  * The window should appear with the correct size, and with the correct background color based on the theme you were previously using. It shouldn't be initially white.
  * If the window was previously full-screen, it should still appear full-screen
* [ ] Overlays, like the autocomplete pop-up should appear in the right spot. After resizing the window in various ways, when typing a word near the edge of the screen, the pop-up should always appear in a reasonable spot.
* [ ] Adding project folders (the `Application: Add Project Folder` command) should open the native file picker dialog correctly.

maxbrunsfeld:
@zcbenz I'm seeing a lot of this in the console on startup:

```
[75796:0127/133819:WARNING:id_weak_map.cc(81)] Removing unexist object with ID 1
```

Is this something to worry about?

zcbenz:
> I'm guessing we need to switch to using an async send, and figure out our own way of waiting until the resize is finished. Do you have any suggestions for this?

I think we should probably pass the window size to the page as part of the url, instead of reading window size in the page, it can save us the resize, which is very expensive.

> I've reverted the Electron version to 0.36.4 since 0.36.5 crashes Atom on startup for me. Dump file is attached... (/cc @paulcbetts, @raelyard)

I created an issue for the crash: https://github.com/atom/electron/issues/4258.

> I usually see this in the console the first time the Atom window loses focus after startup:

It is an warning message, there is no need to worry about it. It has been fixed in v0.36.5.

50Wliu:
So the upgrade to 0.36.7 somehow broke Standard linting... `Fatal error: TypeError: Cannot read property 'match' of undefined` :disappointed:.  The error didn't happen in 0.36.5 so I'll see if 0.36.6 also breaks it.

50Wliu:
So for what it's worth this `script\grunt standard` still works on my machine...
```cmd
>script\grunt standard
Running "standard:src" (standard) task

Linting files...

✓ No Problems

Done, without errors.


Execution Time (2016-01-31 20:30:39 UTC)
loading tasks  4.1s  ████████████████████████████████ 67%
standard:src     2s  ████████████████ 33%
Total 6.1s
```

I'll try an entire `script\cibuild` next to see if that repros.

50Wliu:
Still can't reproduce, even after a full `script\cibuild`.  This isn't fun.  Running Windows 10 x64.

amytruong:
**Windows** (8.1)
Atom Smoke Test 
 (High level test to ensure basic functionality and there's no immediate showstoppers)
 - [x] Run through Welcome Guide
 - [x] Install and update Packages
 - [x] Install and update Themes
 - [x] Verify speed and response time (Windows is just not fast enough - ever)
 - [x] Verify scrolling is fast (Windows just not fast enough - ever)
 - [x] Verify File menu bars/navigation is functioning 

* [x] Startup 
  * Any brief visual oddities like the one @50Wliu captured [above](https://github.com/atom/atom/pull/9627#issuecomment-175266702)
  * The window should appear with the correct size, and with the correct background color based on the theme you were previously using. It shouldn't be initially white.
  * If the window was previously full-screen, it should still appear full-screen
Note: There is an existing bug where if you are full screen and hit restore down, it doesn't restore the window back to the original smaller size. 
* [x] Opening a new window - the window should be offset from the initial window, similarly to how new windows behave for other text editors on the platform (`TextEdit` on mac, `Notepad` on windows).
* [x] Overlays, like the autocomplete pop-up should appear in the right spot. After resizing the window in various ways, when typing a word near the edge of the screen, the pop-up should always appear in a reasonable spot.
* [x] Adding project folders (the `Application: Add Project Folder` command) should open the native file picker dialog correctly.

**Mac Yosemite 10.10.5**
Atom Smoke Test 
 (High level test to ensure basic functionality and there's no immediate showstoppers)
 - [x] Run through Welcome Guide
 - [x] Install and update Packages
 - [x] Install and update Themes
 - [x] Verify speed and response time (Windows is just not fast enough - ever)
 - [x] Verify scrolling is fast (Windows just not fast enough - ever)
 - [x] Verify File menu bars/navigation is functioning 

* [x] Startup 
  * Any brief visual oddities like the one @50Wliu captured [above](https://github.com/atom/atom/pull/9627#issuecomment-175266702)
  * The window should appear with the correct size, and with the correct background color based on the theme you were previously using. It shouldn't be initially white.
  * If the window was previously full-screen, it should still appear full-screen
* [x] Opening a new window - the window should be offset from the initial window, similarly to how new windows behave for other text editors on the platform (`TextEdit` on mac, `Notepad` on windows).
* [x] Overlays, like the autocomplete pop-up should appear in the right spot. After resizing the window in various ways, when typing a word near the edge of the screen, the pop-up should always appear in a reasonable spot.
* [x] Adding project folders (the `Application: Add Project Folder` command) should open the native file picker dialog correctly.

**Linux (Ubuntu 15.00? whatever the latest is)**
Atom Smoke Test 
 (High level test to ensure basic functionality and there's no immediate showstoppers)
 - [x] Run through Welcome Guide
 - [x] Install and update Packages
 - [x] Install and update Themes
 - [x] Verify speed and response time (Windows is just not fast enough - ever)
 - [x] Verify scrolling is fast (Windows just not fast enough - ever)
 - [x] Verify File menu bars/navigation is functioning 
 - [x] Create new file. Save. (Verify it's saved)

Notes: 
* Changing themes can sometimes gives a really quick white flicker in the background. Happens so little and sporadically that it might not be worth mentioning. 
* Resizing the window takes awhile and I'll see white or black space before the app displays again fully but it happens on the Beta too. /shrug
screen shot 2016-02-12 at 1 38 19 pm
* Did have an unusual crash opening up a simple window

* [x] Startup 
  * Any brief visual oddities like the one @50Wliu captured [above](https://github.com/atom/atom/pull/9627#issuecomment-175266702)
  * The window should appear with the correct size, and with the correct background color based on the theme you were previously using. It shouldn't be initially white.
  * If the window was previously full-screen, it should still appear full-screen
* [x] Opening a new window - the window should be offset from the initial window, similarly to how new windows behave for other text editors on the platform (`TextEdit` on mac, `Notepad` on windows).
**Does NOT open offset from initial window but rather right on top of it**
* [x] Overlays, like the autocomplete pop-up should appear in the right spot. After resizing the window in various ways, when typing a word near the edge of the screen, the pop-up should always appear in a reasonable spot.
* [x] Adding project folders (the `Application: Add Project Folder` command) should open the native file picker dialog correctly.

would like to do some more exploratory testing when I get back into town.

maxbrunsfeld:
Thanks for the update Amy! I merged this into master so that we can use it in some further PRs. There should be ample time to do exploratory testing on master once you get back from vacation!

PR #9333, opened 2015-10-29
Add section on reporting bugs to contributing guide

izuzak:
This adds a section on reporting bugs to the contributing guide and removes the generic section on submitting issues. My plan is to add a section on submitting feature requests and sections on other types of "issues" and pull requests, so that we cover all the different types of contributions. I think that will be more helpful than a single generic section since it will allow us to provide specific advice for each type of contribution, and I also think users will find the right section easier.

The section on submitting bug reports is structured in three parts: what you should do before submitting a bug report, which information you should provide in a bug report, and a template for structuring the information. For the part on which information should be provided, I tried to structure this as a list, and bold key sections off each note so that it's easy to scan. I think it will be helpful to list all the questions we often ask when debugging issues -- sometimes that will help users figure out things on their own and provide more helpful reports, and it will allow maintainers to link to this section and say "Please provide more information, as listed here". 

@atom/feedback @atom/issue-triage As always, I'd appreciate your feedback. :honey_pot: Is there any other information we normally ask from users when debugging issues? How do you like the idea of providing a template here, and should some other questions/information be a part of it?

PR #9633, opened 2015-11-16
Add section on feature suggestions to CONTRIBUTING guide

izuzak:
Similar to https://github.com/atom/atom/pull/9333 which added a section on reporting bugs, this adds a section on suggesting features.

@atom/feedback @atom/issue-triage As always, I'd appreciate your feedback. If you can think of anything else that might be helpful here -- please let me know. :koala:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 [How Can I Contribute?](#how-can-i-contribute)
   * [Reporting Bugs](#reporting-bugs)
-  * [Suggesting Features](#suggesting-features)
+  * [Suggesting Enhancements](#suggesting-enhancements)
   * [Your First Code Contribution](#your-first-code-contribution)
   * [Pull Requests](#pull-requests)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -157,6 +158,60 @@ Include details about your configuration and environment:
     * Problem can be reliably reproduced, doesn't happen randomly: [Yes/No]
     * Problem happens with all files and projects, not only some files or projects: [Yes/No]
 
+### Suggesting Features
```
mnquintana: Should we maybe call these "enhancements" to be more consistent with our labeling scheme?

izuzak: > Should we maybe call these "enhancements" to be more consistent with our labeling scheme?
 
I've debated this with myself when I was writing this section, and couldn't come up with any good arguments to use one word over the other. I like the word "feature" a bit more than "enhancement" because I think it's more familiar to most users (I never use "enhancement" when I talk to people, for example). Still, that's a good point about the label name -- I've switched to using "enhancement" instead of "feature" in 3585950. If it becomes an issue of any kind -- we can continue the quest for a better words then. :book: :mag: 

Thanks for pointing this out, @mnquintana :bow:

PR #9636, opened 2015-11-16
Use unique text editor title in window and tab titles	

DouweM:
I have two files, `app/services/notes/create_service.rb` and `app/services/issues/create_service.rb`.

Right now it's impossible to find which is which using either the window or tab title:

screen shot 2015-11-16 at 20 16 02

screen shot 2015-11-16 at 20 16 10

With these changes, it is:

screen shot 2015-11-16 at 20 18 28
screen shot 2015-11-16 at 20 18 36

Both now use `<item>::getLongTitle` which delegates to `TextEditor::getUniqueTitle` in the case of `TextEditor` items.

Besides this, `TextEditor::getUniqueTitle` was modified to match Sublime Text's behavior, where `features/index.html`, `features/main/index.html` and `contact/index.html` will show as `index.html — features`, `index.html — features/main` and `index.html — contact` rather than `features/index.html — features`, `features/.../index.html` and `features/index.html`. With the original implementation, the difference between the first two would still be difficult to spot, and the actual filename could be pushed out of the tab if the directory names were especially long.

PS. I used em-dashes rather than 'normal' dashes since this was developed in conjunction with #9620.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -168,7 +168,7 @@ describe "TextEditor", ->
         buffer.setPath(undefined)
         expect(editor.getLongTitle()).toBe 'untitled'
 
-      it "returns <parent-directory>/<filename> when opened files has identical file names", ->
+      it "returns '<filename> — <parent-directory>' when opened files have identical file names", ->
         editor1 = null
         editor2 = null
         waitsForPromise ->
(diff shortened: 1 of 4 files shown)


DouweM:
Once af22a149bdd8989f5c279863ff319bdc652d7c66 lands in master, I will move `getUniqueTitle` to `getLongTitle`.

DouweM:
@benogle Anything I can do to help this PR along?

benogle:
cc @maxbrunsfeld

n054:
4242.3
 no language set
 ATOM_SPECS_TASK=core NODE_VERSION=0.12
 24 min 13 sec
 This was the only failing test on 4bc4c68. But when I came here to comment all checks passed.

maxbrunsfeld:
Thanks for improving this. Your algorithm is easier to understand than the old one too :+1:.

mik01aj:
I wonder why you were seeing:

    create_service.rb - app/service/......  |  create_service.rb - app/service/......

Because in such situations, I normally saw:

    notes/create_service.rb  |  issues/create_service.rb

...which was imho way better. I recently upgraded and I now see:

    create_service.rb - notes  |  create_service.rb - issues

I don't like the change. Is there any way to get the old behaviour back?

PR #9654, opened 2015-11-17
Only remove node_modules on ci if changed

thedaniel:
_This is just a spike, the hash probably needs more stuff in it (node version, etc), just curious how it will interact with Travis_

Just opening this PR to kick off a couple Travis builds as it's the simplest way to test this change. Basically, are now using their container infrastructure which allows caching, but `script/cibuild` manually removes `node_modules` before running anyway. This just writes a sha into node_modules based on package.json and process.platform that prevents us from wiping it out of the build succeeds & the deps didn't change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,13 @@ install:
 
 script: script/cibuild
 
+cache:
+  directories:
+    - $HOME/.atom/.apm
+    - $HOME/.atom/.node-gyp/.atom
+    - $HOME/.atom/.npm
+    - node_modules
+
 notifications:
   email:
     on_success: never
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -284,6 +284,7 @@ module.exports = (grunt) ->
   ciTasks.push('download-electron')
   ciTasks.push('download-electron-chromedriver')
   ciTasks.push('build')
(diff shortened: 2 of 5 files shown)


thedaniel:
Anecdotally: a run that generated a cache successfully took 17:34, and upon restart, it loaded the cache from travis, verified the fingerprint, and completed in 8:23. Of course this isn't scientific and it doesn't help us on that first build for the branch / new package.json, but it's something!

cc @joefitzgerald @joshaber

thedaniel:
This is pretty simple, but I feel like it's not crazy to merge. @atom/feedback how do you feel about this?

joshaber:
I think this is great. Anything that brings our CI times down without being invasive.

thomasjo:
This seems to make a hell of a lot of sense!

lee-dohm:
:ship: I think this is a great and sensible idea.

thedaniel:
I'm going for it. Should be obvious if we get weird CI failures what else needs to be in the fingerprint other than platform, node version, package.json contents.

philschatz:
Howdy! To help decrease build times #9689 skips the build entirely when no code actually changed. Would something similar be useful for Electron (i.e. check if only `.md` files changed in the commits)?

PR #9668, opened 2015-11-18
Suggest [ci skip] in documentation-only commits

winstliu:
Helps get builds that actually change code started faster.  For example, I'm waiting on Travis to run my Electron PR, and right now it's waiting on 3 documentation-only PRs before it.

If this gets merged I'll also propose the same to atom/electron, since that seems to attract more documentation PRs :).


n054:
Let me run through real quick and see if it triggers anything.

mnquintana:
:100: 

Would definitely be good to help keep our build queue clear.

n054:
Running again.

PR #9670, opened 2015-11-18
Don't create a desktop shortcut when building

winstliu:
Remove the `create-shortcut` task (reasoning is below).


kevinsawicki:
I'm fine adding an option for this but I'd also be fine completely removing this from `script/build`.

This was added very early on when the Windows build was very primitive.

I'm fine leaving this completely up to the builder to create the shortcut manually and since `script/build` always builds to the same place, they could create it once and then just leave it there and it should keep working when they rebuild.

50Wliu:
@kevinsawicki done!

paulcbetts:
I actually think we should :fire: all of install-task on Windows, it tries to copy to some weirdo place as I recall (Program Files?), it doesn't mimic any standard way that Atom is installed

50Wliu:
> it tries to copy to some weirdo place as I recall (Program Files?), it doesn't mimic any standard way that Atom is installed

Yeah, I was planning to address this in a future PR.  Where do you think it should install?  %localappdata% or just %appdata%?

paulcbetts:
@50Wliu It depends:

* You could have it become the newest version of your installed Atom (i.e. install it under `%LocalAppData%\atom\app-VERSION`, then shortcuts to Atom will run it instead
* Same deal, only with `atom-beta`
* You could create an entirely separate fake Squirrel install `atom-dev`. This isn't as hard as it sounds, just mkdirp `%LocalAppData%\atom\app-VERSION`, copy Update.exe in, then run `atom.exe --squirrel-install VERSION`

thedaniel:
The fake squirrel install is appealing, but that diverges from what happens on *nix where our freshly built atom launches when the `atom` command is run. The first option you list seems most appropriate unless we want to change the behavior on other platforms / drop install-task entirely.

paulcbetts:
> The fake squirrel install is appealing, but that diverges from what happens on *nix where our freshly built atom launches when the atom command is run

Ah, in that case you probably want to do Option 1 or Option 2 (personally, I'm always in favor of Option 3)

kevinsawicki:
Going to merge this one in and we can tweak the install location in another pull request.

Thanks for this @50Wliu :+1:

Issue #9526, opened 2015-11-11
Files don't open in initial window on OS X


palant:
I am using Atom 1.2.0-beta2 on OS X 10.10. I open files like that:

    open -a "Atom Beta" foo.js

If Atom isn't running yet then `foo.js` will be opened in the initial window. Opening another file after that will create a second application window, additional files will be added as tabs to that window. I would expect these files to be added to the initial window.

Originally I thought that the initial window is opened in dev mode because Cmd+Alt+I shortcut works there. However, `node-inspector` shows `window.devMode === false` for the initial window and `window.devMode === undefined` for the second one.

palant:
This seems to be caused by the strict comparison for the `window.devMode` flag. For externally opened files `openPath` function will be [called with `devMode === undefined`](https://github.com/atom/atom/blob/ff812749e67ed55518f91fa9d56286f8f3ce4ed6/src/browser/atom-application.coffee#L224). But this makes the comparison in [function `windowForPath`](https://github.com/atom/atom/blob/ff812749e67ed55518f91fa9d56286f8f3ce4ed6/src/browser/atom-application.coffee#L341) fail (`false !== undefined`) so that the existing window is considered unsuitable for the file. Adding something like this to `openPaths` function should do:    

    devMode = if devMode then true else false

izuzak:
Thanks for the report, @palant. I'm going to cc @maxbrunsfeld for more :eyes:.

samrayner:
I can reproduce this in 1.2.0. It seems to be a bug introduced since 1.1.0.

Navigate to a folder in Terminal > `atom .` to open the folder in Atom > `open .` to open the folder in Finder > double click a file within the folder in Finder > opens a new atom window instead of opening a tab in the existing window.

ryanaltvater:
I am also seeing this in the 1.2.0 release. When clicking on file outside of Atom and the current project/files open in Atom, the external file open a second window with the same files as the first window, only this time including the additional external file. This didn't occur in the previous 1.1.0 release.

davydden:
i noticed something interesting: Atom creates a new window when the current one does NOT have a project folder with a target file. When I drag-and-drop the second file from the same folder, the new window is not created.

ryanaltvater:
I use the Project Ring package, so I always have a project folder loaded in my main/current window. I just think it has more to do with the second part you mentioned, when the external file is not a part of the project loaded in the current window.

benogle:
[...]

PR #9672, opened 2015-11-18
Fix opening multiple files from the finder in the same window

### LINKED ISSUE #9574: When opening multiple files Atom (1.3.0.beta0) creates a new window for each file
In Atom 1.2.0 when I opened a lot of files at once from the finder it was just adding them to a previously opened window. Now I'm trying the version 1.3.0.beta0 and it created a new window for each file, each of them having the folder the file was into as the project folder.

This is the result:

Every white rectangle is an Atom window, a crashed one. Is it a bug or a minded change? If so I'd like to point out that those files where all part of a folder that had already an Atom window, if you want to keep this "feature" at least you should notice it, and in that case attach them to the already opened window.

nathansobo:
Fixes #9574

Previously, we were storing the project directory paths as the `initialPaths` key in load settings, which were accessed in the browser process by reading the URL hash. However, this URL hash was not always available, subjecting us to timing issues when opening multiple files in the same folder in rapid succession.

We now store the project directory paths directly on AtomWindow instances on creation, then RPC changes from the render process to the browser process with a custom code path.

Shout out to :airplane::finnadie:’d @as-cii on this for pairing with me.

@izuzak @amytruong This changes browser process code on which test coverage is extremely thin. If you could help test different aspects of opening stuff with Atom once this lands on `beta`, that would be super helpful.

### DIFF
--- a/spec/atom-environment-spec.coffee
+++ b/spec/atom-environment-spec.coffee
@@ -151,7 +151,7 @@ describe "AtomEnvironment", ->
       [dir1, dir2] = [temp.mkdirSync("dir1-"), temp.mkdirSync("dir2-")]
 
       loadSettings = _.extend atom.getLoadSettings(),
-        initialPaths: [dir1]
+        projectDirectoryPaths: [dir1]
         windowState: null
 
       spyOn(atom, 'getLoadSettings').andCallFake -> loadSettings
@@ -165,7 +165,7 @@ describe "AtomEnvironment", ->
       atom.loadStateSync()
       expect(atom.state.stuff).toBeUndefined()
 
-      loadSettings.initialPaths = [dir2, dir1]
+      loadSettings.projectDirectoryPaths = [dir2, dir1]
       atom.state = {}
       atom.loadStateSync()
       expect(atom.state.stuff).toBe("cool")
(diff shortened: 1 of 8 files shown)


izuzak:
Thanks for jumping on this, @nathansobo and @as-cii. 

As far as I can tell, this doesn't completely fix the problem for me. When I follow the steps shown in https://github.com/atom/atom/issues/9574#issuecomment-157411492, I get better behavior now, but still not expected.

In that GIF, I first open a single file from atom/atom, then three more. That cause four windows to be created -- one for the first file, and one for each of the three after that. 

When I use a build of this PR, following the same steps causes two windows to be created -- one for the first file, and one for the three I open after that. The behavior I expected to see was that one window would be created in total -- after I open the first file, the same window would be used when I opened the other three (because they're from the same project/folder). 

Do you see the same behavior? I this behavior by any chance expected? 

I'm not sure if this is the same issue, but @benogle pointed out that https://github.com/atom/atom/issues/9526 seemed similar. Perhaps the problem pointed out in that issue is the missing piece to this?

nathansobo:
@izuzak Thanks for finding that. Fixed. When this goes green I'm going to merge into beta.

izuzak:
@nathansobo I'm not sure if I'm missing something, but I can now reproduce the original problem (described in https://github.com/atom/atom/issues/9574) with multiple windows being opened, both on master and on beta. 

Do you see the same behavior on master? I'm wondering if something got lost again like in that other PR. :confused:

nathansobo:
@izuzak Yep. I just cherry picked it back in. I was merging with `--strategy ours` expecting the only conflict to be in the `package.json` for the version, which in retrospect was really really dumb. I combed through the last few merges to try to spot any other places I might have made a mistake, but I think this is the only one. My apologies. Coding at a conference isn't always the best idea.

izuzak:
:boom: Things are working now, @nathansobo -- thanks!

PR #9687, opened 2015-11-18
Provide package.json fields so we can defer requiring packages' main modules

maxbrunsfeld:
#### The Problem

In order to interoperate with the deserialization of the Atom workspace, there are some APIs that packages have to call as soon as they are loaded (as opposed to inside of their `activate` method):
- `atom.views.addViewProvider`
- `atom.deserializers.add`

This forces Atom to load all packages' main modules at an early point in startup, even though many packages can be _activated_ lazily. This allows packages to affect the app even if they are disabled, and may reduce our ability to optimize startup time.
#### The Proposed API Change

This PR will add 3 new optional fields to packages' `package.json` files:
- `configSchema` - an object specifying the package's config schema
- `deserializers` - an object whose keys are deserializer names, and whose values are paths to files that export deserialization functions
- `viewProviders` - an array whose values are paths to files that export view-providing functions. These will need to work a little bit differently than normal view providers, because they can't be specified in relation to a particular model constructor. Maybe they will just be functions that return `null` if they do not apply to the given model.

It will also add a deprecation warning when packages accesses the `atom` global at load time.
#### Changes to Package Loading

When packages use the new `package.json` fields, we can defer loading any of their code until it is used, because we can know that there is no reason for them to run code at load time.
- The main module will be loaded when the package is activated.
- View providers will be loaded when either the package is activated.
- Deserializers will be loaded when `deserialize` is called with one of the specified deserializer names.

Once the deprecation period is over, we can use this approach for all packages.
#### Todo
- [x] Respect the 3 new fields in package.json.
- Use new fields to optimize package loading
  - [x] If a package has `configSchema`, `viewProviders`, or `deserializers`, don't load its main module until it is activated.
  - [x] Don't load a package's view providers until either it is activated or one of its deserializers is used.
- [x] Remember which packages use the atom global at require time, and for those that don't, don't require the main module until they are activated.

/cc @nathansobo


nathansobo:
Great. We've needed this for a very long time. Some thoughts on the key names.

* Seems like they should be camelCased rather than dash-separated to blend in with other keys in `package.json`.
* We haven't prefixed several keys with the word "atom" thus far. Maybe we should have, but these are Atom packages after all and starting now seems to introduce inconsistencies.

maxbrunsfeld:
>  Some thoughts on the key names

D'oh; of course. Updated.

@nathansobo How do you think the view providers should work? I was thinking we could have a simple array of view providers functions that will always be called, but should return `null` unless they apply to a given model.

PR #9688, opened 2015-11-18
Fix updating

### LINKED ISSUE #9658: Atom Beta - v1.3.0-beta0 - Unable to update SSL error
When attempting to update the Beta I am seeing the following message 

    An SSL error has occurred and a secure connection to the server cannot be made.: (null)

The release version updated successfully, so I doubt this is a local environment issue.

What additional information can I provide?

joshaber:
Give github-cloud.s3.amazonaws.com a pass. Fixes #9658.

We’ll use an exception for now until we can get an updated certificate in place.

After this lands I’ll cherry-pick it to `beta`.

### DIFF
--- a/resources/mac/atom-Info.plist
+++ b/resources/mac/atom-Info.plist
@@ -38,6 +38,17 @@
 	<string>speakeasy.pem</string>
 	<key>SUScheduledCheckInterval</key>
 	<string>3600</string>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSExceptionDomains</key>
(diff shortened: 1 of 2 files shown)


lee-dohm:
:rocket: Looks good to me.

joshaber:
I mentioned this is the issue but I'll mention it here too. This comes with a big caveat: those already on the beta with this problem won't see the new update. They'll need to manually download a new version, or wait until we get the new cert deployed.

So depending on how long the infrastructure work takes, we may not want to merge this.

PR #9689, opened 2015-11-18
[WIP] Skip CI build entirely when code did not change

philschatz:
This checks if the code changed by building a glob of all the minified code (& JSON files) in this repo and compares it to a generated glob of the previous commit.

Comments, local variable renames, and lint fixes should not start an entire build. But changing package versions should.

Hopefully Travis won't :sweat: over this PR too much :smile: 

**Side note:** I was not sure where to install `uglify` since it seems `build/package.json` is too late.

You can test it out locally:
1. check out this branch
2. make a code change or add a comment and commit
3. run `script/cibuild-when-code-changed`
4. See if it says "Just a documentation change. No need to run tests!"

If this approach seems promising I can get the rest of the "kinks" out:
- [ ] switch to using `node` instead of shell scripts
- [x] include JSON/CSON/scripts/spec tests in the glob
- [x] build babel files
- [ ] generate a checksum instead of using `diff`
# Console Output

```
$ script/cibuild-when-code-changed
DID_CODE_CHANGE: CASE 1: This is the continuous-integration/travis-ci/pr case
DID_CODE_CHANGE: Building slug for this commit
DID_CODE_CHANGE: END_COMMIT=c1002e587ead0c237feb75963c0427c765c3a1e7
DID_CODE_CHANGE: Building slug for the previous commit (to compare against)
DID_CODE_CHANGE: START_COMMIT=cad74060855aa37c5de528d6f03761ac74170103
DID_CODE_CHANGE: Comparing slugs
DID_CODE_CHANGE: Just a documentation change. No need to run tests
The command "script/cibuild-when-code-changed" exited with 0.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ debug.log
 docs/output
 docs/includes
 spec/fixtures/evil-files/
+
+.ci-cache-previous.txt
+.ci-cache-this.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
   - nvm install $NODE_VERSION
   - nvm use $NODE_VERSION
 
-script: script/cibuild
+script: script/cibuild-when-code-changed
 
 notifications:
   email:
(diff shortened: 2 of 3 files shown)


damieng:
Out of date, no further activity.

PR #9709, opened 2015-11-20
Compute content width when the longest row is the first one

### LINKED ISSUE #9649: Problem With Horizontal Scroll in Specific Condition
I have created a file to replicate this problem here:
https://gist.github.com/mssio/2e559b6d53de7a2e2438

Before you copy the content of the files in Atom please make sure that the width of your atom window is not super wide (can't display more than 150 chars) and the height of your atom window is not super tall (can't display more than 60 lines). Also make sure that the atom configuration on soft wrap is turned off.

After you copy the file you can see on line 1, everything is normal, you can use horizontal scroll and continue typing on line 1 and automatically scrolled to follow the cursor. But let's try to see the line 105, in that line we can't do a horizontal scroll and if I type on that line, the window will not follow the cursor.

This is really annoying problem because a lot of my code have a format like this and when I am in the middle / bottom, I will not be able to edit the code using Atom. I have to switch to another text editor.

I hope this is not a duplicate issue, please just close it if it is a duplicate. And I am sorry for my bad English.

as-cii:
Fixes #9649

There was a subtle problem with an `if` statement that went overlooked when implementing the new measurement strategy which caused the first screen row to be ignored when computing content width if that row was both hidden and the longest one.

This PR fixes it and adds a spec to prevent this from happening in the future.

/cc: @abe33 @nathansobo

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -377,7 +377,8 @@ class TextEditorPresenter
     endRow = @constrainRow(@getEndTileRow() + @tileSize)
 
     screenRows = [startRow...endRow]
-    if longestScreenRow = @model.getLongestScreenRow()
+    longestScreenRow = @model.getLongestScreenRow()
+    if longestScreenRow?
       screenRows.push(longestScreenRow)
     if @screenRowsToMeasure?
       screenRows.push(@screenRowsToMeasure...)
--- a/spec/fake-lines-yardstick.coffee
+++ b/spec/fake-lines-yardstick.coffee
@@ -7,6 +7,7 @@ class FakeLinesYardstick
 
   prepareScreenRowsForMeasurement: ->
     @presenter.getPreMeasurementState()
+    @screenRows = new Set(@presenter.getScreenRows())
 
   getScopedCharacterWidth: (scopeNames, char) ->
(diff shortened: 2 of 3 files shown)


src/text-editor-presenter.coffee
```diff
@@ -377,7 +377,8 @@ class TextEditorPresenter
     endRow = @constrainRow(@getEndTileRow() + @tileSize)
 
     screenRows = [startRow...endRow]
-    if longestScreenRow = @model.getLongestScreenRow()
+    longestScreenRow = @model.getLongestScreenRow()
+    if longestScreenRow?
```
nathansobo: Classic. Damn you falsy zero! :smile:

nathansobo:
:ship: Should we hotfix beta or stable with this?

as-cii:
So, I think the :bug: is present on both `beta` and `stable` so I think we should apply the commit on both branches. (although I am not sure I'll have the time today to actually do this :airplane:)

nathansobo:
I just cherry picked to `stable` and merged back into `beta` and `master`. Building now.

as-cii:
Thank you! :green_heart: :heart: :yellow_heart: :blue_heart:

PR #9711, opened 2015-11-20
Export ::isTextEditor function

as-cii:
To understand if an object was a `TextEditor`, we previously suggested users to check for the existence of some of its methods. It turns out that this approach could lead to issues like https://github.com/atom/autocomplete-plus/issues/614, caused by checking for the wrong methods.

The alternative approach herein proposed tries to solve this problem by exporting an `::isTextEditor` function in `exports/atom.coffee`. This allows us to not export the internal `TextEditor` constructor while still letting user to understand whether a given object is a `TextEditor` or not.

I am going to open a PR that makes use of this new function in `autocomplete-plus`, where the problem first appeared, and search for other spots where we could use it.

### DIFF
--- a/exports/atom.coffee
+++ b/exports/atom.coffee
@@ -35,7 +35,6 @@ unless process.env.ATOM_SHELL_INTERNAL_RUN_AS_NODE
         The `TextEditor` constructor is no longer public.
 
         To construct a text editor, use `atom.workspace.buildTextEditor()`.
-        To check if an object is a text editor, look for for the existence of
-        a public method that you're using (e.g. `::getText`).
+        To check if an object is a text editor, use `atom.workspace.isTextEditor(object)`.
       """
       TextEditor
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
What do you think about having `isTextEditor` be on `workspace`, together with `buildTextEditor` and `observeTextEditors` and friends?

nathansobo:
Putting it on workspace is interesting. Does seem to slot in nicely there and avoids yet another property on the `atom` global.

as-cii:
That sounds good, it still feels a bit weird for `isTextEditor` to be an instance method on `Workspace` given that it doesn't really depend on any of `Workspace`'s state. However, I agree it's more cohesive and I feel like users would be less surprised to find it in `Workspace` rather than on the atom global exports. :+1:

nathansobo:
I also agree that it's a bit weird.

as-cii:
Thanks!

PR #9721, opened 2015-11-20
Fix Long tab lengths fail, and default back to 4 Issue #9147 

g23-aw:
Implement fix for issue #9147. Valid tab lengths are any number between 1 and 100.

tablength


benogle:
Closing in favor of https://github.com/atom/atom/pull/9198

PR #9723, opened 2015-11-20
Insert relative path when tree view file is dropped on editor

DouweM:
drag-and-drop

The GIF is a little misleading, since it appears that the path is inserted at the location where the drag item is dropped. In reality, it's always inserted at the cursor location, where no cursor is visible while the text editor is out of focus.

I'll gladly add tests, but I would like some feedback on the functionality first.

### DIFF
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -1,4 +1,5 @@
 _ = require 'underscore-plus'
+path = require 'path'
 scrollbarStyle = require 'scrollbar-style'
 {Range, Point} = require 'text-buffer'
 {CompositeDisposable} = require 'event-kit'
@@ -227,6 +228,7 @@ class TextEditorComponent
   listenForDOMEvents: ->
     @domNode.addEventListener 'mousewheel', @onMouseWheel
     @domNode.addEventListener 'textInput', @onTextInput
+    @domNode.addEventListener 'drop', @onDrop
     @scrollViewNode.addEventListener 'mousedown', @onMouseDown
     @scrollViewNode.addEventListener 'scroll', @onScrollViewScroll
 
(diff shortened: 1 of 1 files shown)


velaia:
Hello Douwe, I would expect the drop of a file in the editor to open the file. Similar to dropping files on other programs opens these files or uploads them (HTML5 upload e.g.)
But I see that the functionality comes in extremely handy for programmers. Would it be possible to combine it with a key-press, like the CTRL + drag Item with mouse on many office applications that inserts a copy of the dragged item?

velaia:
I've slept over it and had an intuition. If you drop a file on atom from Finder e.g. it already opens the file in a new tap. And of course double-clicking the file in the file-browser panel also opens it. I still feel most (regular) users would not expect that behavior - and it's very programmer specific anyway. If I read your code properly, you don't differentiate by filetype in your onDrop function.
There's a great SCADA software I've had the chance of working with recently called Ignition by Inductive Automation. If you drop a (OPC) tag onto the Ignition Designer area it gives you a small menu right next to the mouse pointer where you can chose what you want to insert - maybe that would be an option ...

DouweM:
@velaia:

> I still feel most (regular) users would not expect that behavior 

Is that a problem? They don't currently expect any other behavior either, since currently drag/drop from the tree view doesn't do anything.

> and it's very programmer specific anyway

Right, Atom is a developer tool after all :)

> If you drop a (OPC) tag onto the Ignition Designer area it gives you a small menu right next to the mouse pointer where you can chose what you want to insert - maybe that would be an option ...

I think this would make the flow a lot more cumbersome, with very little benefint.

velaia:
@DouweM I think you're right. The feature doesn't hurt anyone and by putting it in people can see whether they like it or not. I personally would rather have the functionality than not have it.

One question: What is the best behavior for dropping a folder? Attach a trailing slash or not?

DouweM:
@velaia I think we don't need the trailing slash, in most cases I can think of where my code expects the path to a folder, it wouldn't need it.

benogle:
This is pretty neat. Can you add a spec for this behavior?

DouweM:
@benogle Sure! As an aside, do you know of a way to insert the path at the mouse pointer location rather than the original text cursor location?

benogle:
There is [screenPositionForMouseEvent](https://github.com/atom/atom/blob/65e19b7a882ba66d284faaaa5bffa112c5029477/src/text-editor-component.coffee#L909) which is a member of TextEditorComponent.

damieng:
Out of date, no further activity.

PR #9729, opened 2015-11-21
Upload releases to s3

joshaber:
Upload releases to a bucket separate from Releases.

This is a step toward fixing #9658.


package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "async": "0.2.6",
     "atom-keymap": "^6.1.1",
+    "aws-sdk": "^2.2.18",
```
joshaber: Given that we only need for this CI, should we install it in `cibuild` (or some such) instead? @atom/feedback

maxbrunsfeld: What we've been doing is putting these dependencies into `build/package.json`.

joshaber: :+1:

joshaber:
I think this is ready :pray:

@maxbrunsfeld I'd love your :eyes: on this to make sure I didn't mess up releases.

@thedaniel Do you think this makes sense? Does this seem ok for the atom.io side of things?

maxbrunsfeld:
:+1: This looks good to me @joshaber, though I'm still not totally fully fluent in these scripts myself. I think we should go for it.

joshaber:
@thedaniel @maxbrunsfeld I set the object to not be cached (https://github.com/atom/atom/commit/7b976bc689e479995ca1ce40a5f2cc357aa5984a) and it seems to work as we'd expected. **But** I accidentally stumbled upon what I think is an even better solution: we can [configure CloudFront](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/QueryStringParameters.html) to take query string parameters into account with respect to caching.

So if atom.io throws the asset's last modified date onto the download URL, we should get consistent caching and the right download regardless of how many times the file is changed in place :metal:

joshaber:
(I hit submit a bit prematurely there.)

So given that we probably weren't gonna ship this until next week anyways, I'm gonna take the time to reconfigure CloudFront and then allow caching again.

joshaber:
@atom/feedback I believe this is ready :pray: 

My current plan:

1. Merge into `master`.
2. Cherry-pick into `stable` and `beta`.
3. Ensure the build binaries are uploaded for beta.
4. Upload the current release build binaries manually.

lee-dohm:
It looks good to me. I don't see any glaring issues. Though I also am not terribly familiar with these scripts.

build/tasks/publish-build-task.coffee
```diff
@@ -221,15 +222,43 @@ uploadAssets = (release, buildDir, assets, callback) ->
...
+
+    key = "releases/#{release.tag_name}/#{assetName}"
+    uploadParams =
+      Bucket: s3Bucket
+      ACL: 'public-read'
+      Key: key
+      Body: fs.createReadStream(assetPath)
+    s3.upload uploadParams, (error, data) ->
+      if error?
+        console.log("Upload release asset #{assetName} to S3 failed", error)
```
kevinsawicki: Should this use `logError`?

joshaber: Yup, good catch :+1:

joshaber:
Dunno what's up with Travis.

PR #9730, opened 2015-11-21
bump Contributor Covenant version to 1.3

zeke:
This updates the version from 1.2 to 1.3 in the README and CONTRIBUTING docs.

http://contributor-covenant.org/version/1/3/0/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 ### Code of Conduct
 
-This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0)  (see [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)).
+This project adheres to the Contributor Covenant [code of conduct](CODE_OF_CONDUCT.md).
 By participating, you are expected to uphold this code.
 Please report unacceptable behavior to [atom@github.com](mailto:atom@github.com).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 ### Code of Conduct
 
-This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0).
+This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0)  (see [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)).
 By participating, you are expected to uphold this code.
 Please report unacceptable behavior to [atom@github.com](mailto:atom@github.com).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 ### Code of Conduct
 
-This project adheres to the [Contributor Covenant 1.2](http://contributor-covenant.org/version/1/2/0).
+This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0).
 By participating, you are expected to uphold this code.
 Please report unacceptable behavior to [atom@github.com](mailto:atom@github.com).
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,24 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free experience for everyone, regardless of level of experience, gender, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, or nationality.
+
+Examples of unacceptable behavior by participants include:
+
+- The use of sexualized language or imagery
+- Personal attacks
+- Trolling or insulting/derogatory comments
+- Public or private harassment
+- Publishing other's private information, such as physical or electronic addresses, without explicit permission
+- Other unethical or unprofessional conduct
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, or to ban temporarily or permanently any contributor for other behaviors that they deem inappropriate, threatening, offensive, or harmful.
+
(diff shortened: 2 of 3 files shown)


izuzak:
@zeke Thanks for taking the time to open a pull request for this. :heart:

In the future, could I trouble you to also provide a diff which clearly shows the differences in the text between the two versions? Let me try and explain why that's important. In a way, we're bumping a dependency here (a very important one) and we don't really have any tests which would verify that the new version still works for the project. For that reason, we need humans to verify that manually, and it is much easier to do that when you can see the whole and exact diff in one place (instead of just the diff for changing the URL, and without asking people to figure out the diff themselves). Does that make sense?

I've gone ahead and did that below:

```diff
@@ -12,10 +12,12 @@ Trolling or insulting/derogatory comments
 Public or private harassment
 Publishing other's private information, such as physical or electronic addresses, without explicit permission
 Other unethical or unprofessional conduct
-Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct. By adopting this Code of Conduct, project maintainers commit themselves to fairly and consistently applying these principles to every aspect of managing this project. Project maintainers who do not follow or enforce the Code of Conduct may be permanently removed from the project team.
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, or to ban temporarily or permanently any contributor for other behaviors that they deem inappropriate, threatening, offensive, or harmful.
+
+By adopting this Code of Conduct, project maintainers commit themselves to fairly and consistently applying these principles to every aspect of managing this project. Project maintainers who do not follow or enforce the Code of Conduct may be permanently removed from the project team.
 
 This code of conduct applies both within project spaces and in public spaces when an individual is representing the project or its community.
 
-Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by opening an issue or contacting one or more of the project maintainers.
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting a project maintainer at [INSERT EMAIL ADDRESS]. All complaints will be reviewed and investigated and will result in a response that is deemed necessary and appropriate to the circumstances. Maintainers are obligated to maintain confidentiality with regard to the reporter of an incident.
 
-This Code of Conduct is adapted from the Contributor Covenant, version 1.2.0, available from http://contributor-covenant.org/version/1/2/0/
+This Code of Conduct is adapted from the Contributor Covenant, version 1.3.0, available from http://contributor-covenant.org/version/1/3/0/
```

@atom/core @atom/feedback How do you all feel about these changes? I'll add my thoughts when I find a moment to read this thoroughly.

I did notice one change, though:

```
... by contacting a project maintainer at [INSERT EMAIL ADDRESS] ...
```

That makes me think that just linking to the Contributor Covenant isn't going to work anymore and we need to have a local copy of it with that part substituted with an actual email address.

lee-dohm:
> That makes me think that just linking to the Contributor Covenant isn't going to work anymore and we need to have a local copy of it with that part substituted with an actual email address.

Agreed. Didn't Contributor Covenant once have a query parameter version that would put the right things in the right places?

lee-dohm:
The [Elixir language project](https://github.com/elixir-lang/elixir/) uses a [`CODE_OF_CONDUCT.md`](https://github.com/elixir-lang/elixir/blob/master/CODE_OF_CONDUCT.md) in their project. Perhaps we could do the same.

izuzak:
> The Elixir language project uses a CODE_OF_CONDUCT.md in their project. Perhaps we could do the same.

@lee-dohm Yeah, I prefer that approach over linking to an external site. :+1: 

I just had another look at the changes in the code of conduct and they look find to me. @atom/feedback if you don't see any :red_circle: :flags: in these changes, we can create a local copy of the code of conduct and add our email address to it (or just a link to https://atom.io/contact ?).

50Wliu:
Changes look fine to me!  I like the extra language about banning.

lee-dohm:
I'm :+1: on the changes. Thank you for the diff @izuzak. I think the email address would be better than the contact form. Email addresses are a little more clear on to whom they are going, which is better in a situation that is potentially embarrassing.

zeke:
Sorry for the slow response. I subscribed to the whole atom/atom repo and it got lost in sea of issues! I'm happy to update this PR with a local `CODE_OF_CONDUCT.md`.

What email address should be used?

50Wliu:
I believe it's atom@github.com.

zeke:
Added a file and `mailto` link to atom@github.com

This is ready for review.

zeke:
Fetched upstream master and rebased to see if that fixes the (seemingly-unrelated) CI failure.

50Wliu:
:+1:.

In the future, just add `[ci skip]` to these kinds of commits (ie ones that don't touch code or change functionality) since there's no point in testing them.

zeke:
> [ci skip]

Nice. Will do!

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 ### Code of Conduct
 
-This project adheres to the [Contributor Covenant 1.2](http://contributor-covenant.org/version/1/2/0).
+This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0).
```
lee-dohm: This link and the link in the README should probably link to the CODE_OF_CONDUCT document.

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 ### Code of Conduct
 
-This project adheres to the [Contributor Covenant 1.2](http://contributor-covenant.org/version/1/2/0).
+This project adheres to the [Contributor Covenant 1.3](http://contributor-covenant.org/version/1/3/0)  (see [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md)).
```
izuzak: @zeke I think @lee-dohm wanted to say that this should only link to the CODE_OF_CONDUCT.md file, not to http://contributor-covenant.org/version/1/3/0. The CODE_OF_CONDUCT.md file contains a link to http://contributor-covenant.org/version/1/3/0 (notice the section at the bottom), so it's not needed here and it might be confusing for users who will click that link first, land on http://contributor-covenant.org/version/1/3/0, and not find an email address filled out there. 

I think we should link to CODE_OF_CONDUCT.md only, both here and in the README, and link to http://contributor-covenant.org from CODE_OF_CONDUCT.md (as it is now).

lee-dohm: Yes, that's what I meant. Thanks @izuzak :grinning:

zeke: Done!

lee-dohm:
Thanks @zeke!

mnquintana:
:+1: Looks good to me! Totally fine with the changes here.

kevinsawicki:
Thanks for this @zeke 

Would you be up for doing a similar pull request to Electron?

Issue #6035, opened 2015-03-19
Atom fails to build on FreeBSD 11.0 x86_32


cpu82:
I tried to build it following this instructions: 
https://github.com/atom/atom/blob/8db47bd7b2226b858a255867aaa16419c652f64d/docs/build-instructions/freebsd.md

Here is the full build log:
http://pastebin.com/VEtB9WuY

TIA,

cpu82:
Some outputs:
```
% uname -a
FreeBSD box.underbuild.com 11.0-CURRENT FreeBSD 11.0-CURRENT #0 r279847M: Wed Mar 11 01:46:19 CET 2015     charly@box.underbuild.com:/usr/obj/usr/src/sys/PROD  i386
```
```
% pkg info node
node-0.12.0_1
Name           : node
Version        : 0.12.0_1
Installed on   : Thu Mar 19 15:16:37 CET 2015
Origin         : www/node
Architecture   : freebsd:11:x86:32
Prefix         : /usr/local
Categories     : www
Licenses       : MIT
Maintainer     : linpct@gmail.com
WWW            : http://nodejs.org/
Comment        : V8 JavaScript for client and server
Annotations    :
Flat size      : 12.7MiB
Description    :
node - evented I/O for V8 javascript

WWW: http://nodejs.org/
```
```
% pkg info npm
npm-2.7.1
Name           : npm
Version        : 2.7.1
Installed on   : Thu Mar 19 20:23:14 CET 2015
Origin         : www/npm
Architecture   : freebsd:11:x86:32
Prefix         : /usr/local
Categories     : www
Licenses       : MIT
Maintainer     : sunpoet@FreeBSD.org
WWW            : http://npmjs.org/
Comment        : Node package manager
Options        :
	IOJS           : off
	NODE           : on
	NODE010        : off
	NODE_DEVEL     : off
Annotations    :
	cpe            : cpe:2.3:a:npmjs:node_packaged_modules:2.7.1:::::freebsd11:x86
Flat size      : 7.44MiB
Description    :
npm is a package manager for node. You can use it to install and publish your
node programs. It manages dependencies and does other cool stuff.

WWW: http://npmjs.org/
WWW: https://github.com/npm/npmff
```

cpu82:
I'm willing to help with this issue.

pusateri:
The problem with csslint is fixed in later versions. Applying this patch gets past this problem but on to the next:

[...]

PR #9748, opened 2015-11-22
Upgrade pathwatcher

fscherwi:
Upgrade pathwatcher, which has no API-changes and better support for UNIX-like systems!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "normalize-package-data": "^2.0.0",
     "nslog": "^3",
     "oniguruma": "^5",
-    "pathwatcher": "~6.3",
+    "pathwatcher": "~6.5",
     "property-accessors": "^1.1.3",
     "random-words": "0.0.1",
     "resolve": "^1.1.6",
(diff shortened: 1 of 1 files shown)


lee-dohm:
@maxbrunsfeld @50Wliu is there anything between v6.2 and v6.3 of pathwatcher that necessitates this upgrade? From looking at the history, it looks mostly like CI and build changes?

50Wliu:
Looks like 6.3.0 might help out people trying to build Atom on OpenBSD/FreeBSD (#6035, though they don't seem to have had troubles with pathwatcher yet).  Other than that, it looks like the upgrade is just for the sake of upgrading.  I never got any warnings when building Atom on Windows and VS2015, so *shrug*.  Perhaps that's only when directly building pathwatcher.

damieng:
We should probably bump right up to 6.5 as there are some fixes for watching root of drives on Windows, treating move to trash as delete on the Mac etc.

fscherwi:
@damieng changed pathwatcher to 6.5

lee-dohm:
Thanks for this!

If you could include whatever testing you've done before submitting the PR, that would be helpful :grinning:

Issue #7662, opened 2015-07-02
ctrl-delete does not remove CRLF


dwelle:
By `ctrl-delete` I mean `editor:delete-to-next-word-boundary`, `editor:delete-to-end-of-word` or  `editor:select-to-beginning-of-word` (which is probably used by the first two commands).

I.e. when issuing the above commands in this position:

![image](https://cloud.githubusercontent.com/assets/5153846/8487587/5285ae6e-210d-11e5-868b-e41bb9378dd0.png)

__Nothing__ happens, the `CRLF` windows line-ending is not consumed.

Linux's `LF` works fine, though (`editor:delete-to-next-word-boundary`):

![image](https://cloud.githubusercontent.com/assets/5153846/8487775/7565e1b4-210e-11e5-9206-085256a411e9.png)

Also, the other way round (`editor:delete-to-previous-word-boundary` etc.) works fine.

Windows 7, Atom 1.0

nycdotnet:
This is also a problem with keyboard navigation.  When holding down CTRL and using the left arrow key, Atom will successfully navigate backward through a document one "word" at a time.  CTRL and right arrow also works *until* the end of the line is reached.  Then atom will stay there stuck at the end of the line when CTRL+Right arrow is pressed.  Right-arrow alone works fine to navigate to the first position of the next line, and then CTRL+Right will work again until the end of that line is reached.

billybonks:
@nycdotnet is this on windows ? seems to be working on mac but i use alt instead of control

davidspiess:
This problem is a nasty one.
It's caused by a regex in `wordRegExp` function in [cursor.coffee](https://github.com/atom/atom/blob/v1.1.0/src/cursor.coffee#L609)

`["^[\t ]*$"]` causes 0-length matches with CRLF line ending.

JavaScript treats CRLF pairs as two line breaks. ^ matches in the middle of and after CRLF, while $ matches before and in the middle of CRLF.

You can read more about this here https://github.com/atom/find-and-replace/issues/454

I tried to fix it with `["^[\t ]+$"]` as suggested in the issue above, but by running the testsuite empty lines do not count as words anymore. Maybe someone knows a solution for it.

sanuj:
Please close this issue as it was fixed in #9937.

50Wliu:
Thanks @sanuj.

nycdotnet:
[...]

PR #9763, opened 2015-11-23
:bug: Fix windows CR LF issue in wordRegExp

dranzerashi:
Issue #7662

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -723,20 +723,19 @@ describe "TextEditor", ->
         editor.moveToBeginningOfWord()
         expect(editor.getCursorBufferPosition()).toEqual [9, 2]
 
-      buffer.setText(buffer.getText().replace(/\n/g, "\r\n"))
-
-      it "treats lines with only whitespace as a word with CRLF line ending", ->
+      it "treats lines with only whitespace as a word (CRLF line ending)", ->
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks for this! Could you write a test for this @dranzerashi?

dranzerashi:
I didn't find a spec for cursor, Is it ok if I try updating the word selection and delete methods in text-editor-spec to cover the changes.

dranzerashi:
The test case concerning the issue already exists in Text-editor-spec, however the sample.js file that is loaded as test-case has unix style line endings. So the test-case always passes even on windows.

dranzerashi:
The text-editor-spec already has suitable test cases concerning the issue, however the sample.js file is written in unix format and hence passes all the test cases without raising errors.

maxbrunsfeld:
Yeah, so we just need a test that converts `sample.js` to use CRLF line endings and checks that `ctrl-delete` still works properly at the end of a line.

maxbrunsfeld:
I included your work in my pull request #9937, which did some additional fixes for `CRLF` line endings. Unfortunately, without thinking, I rebased my branch afterward, so these commits did not get marked as included. My apologies for that, @dranzerashi. Thanks for this work. Through https://github.com/atom/atom/pull/9937, it will be released in Atom 1.4.0.

PR #9788, opened 2015-11-24
Interact with DevTools asynchronously

### LINKED ISSUE #9697: Open DevTools with alt-cmd-i on OSX
When opening the devTools with alt-cmd-i on OSX it added an **`1`** to opened file cursor position

as-cii:
Fixes #9697.

The above issue can be consistently reproduced by switching to Dvorak and trying to open the DevTools through `Cmd-Alt-I` (which, if you have a Qwerty keyboard, is `Cmd-Alt-G`). This doesn't happen on most Qwerty keyboards and, quite oddly, I tracked down the problem to be related to DevTools being opened or closed. Indeed, if we try to replace `ApplicationDelegate::toggleWindowDevTools` with a simple log statement this problem doesn't seem to be reproducible. The reason why only certain keyboard layouts trigger this particular behavior is obscure to me at the moment, but somehow I believe this has to do with the DevTools appearing and disappearing, thereby confusing Chromium's input events. (@zcbenz: maybe you have some ideas on what's triggering this behavior?)

I think this wasn't caused by a recent Electron upgrade, but rather to the way we have changed Atom DevTools interaction: in the past, this was asynchronous but it got changed to be sync during the `AtomEnvironment` refactoring. I switched back to using an asynchronous approach, so that our keybinding/commands event handlers have the chance to fully execute without waiting for DevTools to be opened/closed.

I _think_ we were using a sync version because [these lines of code](https://github.com/atom/atom/blob/master/src/atom-environment.coffee#L672-L674) need to be executed in succession, so that the javascript code can be run when DevTools are already open. It turns out that, before these changes, the console wasn't getting shown if DevTools were closed because we still have to wait until the `devtools-opened` event gets triggered. This PR fixes also this problem by implementing a Promise-based API on top of Electron DevTools API.

### DIFF
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -66,15 +66,30 @@ class ApplicationDelegate
     ipc.send("call-window-method", "setFullScreen", fullScreen)
 
   openWindowDevTools: ->
-    if remote.getCurrentWindow().isDevToolsOpened()
-      Promise.resolve()
-    else
-      new Promise (resolve) ->
-        remote.getCurrentWindow().once("devtools-opened", -> resolve())
-        ipc.send("call-window-method", "openDevTools")
+    new Promise (resolve) ->
+      process.nextTick ->
+        if remote.getCurrentWindow().isDevToolsOpened()
+          resolve()
+        else
+          remote.getCurrentWindow().once("devtools-opened", -> resolve())
+          ipc.send("call-window-method", "openDevTools")
+
+  closeWindowDevTools: ->
+    new Promise (resolve) ->
+      process.nextTick ->
+        unless remote.getCurrentWindow().isDevToolsOpened()
+          resolve()
+        else
+          remote.getCurrentWindow().once("devtools-closed", -> resolve())
+          ipc.send("call-window-method", "closeDevTools")
 
   toggleWindowDevTools: ->
-    ipc.send("call-window-method", "toggleDevTools")
+    new Promise (resolve) =>
+      process.nextTick =>
+        if remote.getCurrentWindow().isDevToolsOpened()
+          @closeWindowDevTools().then(resolve)
+        else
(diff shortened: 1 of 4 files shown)


src/application-delegate.coffee
```diff
@@ -66,13 +66,33 @@ class ApplicationDelegate
     ipc.send("call-window-method", "setFullScreen", fullScreen)
 
   openWindowDevTools: ->
-    remote.getCurrentWindow().openDevTools()
+    new Promise (resolve) ->
+      process.nextTick ->
```
nathansobo: Why the next tick here?

nathansobo:
:clap:  Great investigation. This changes the API slightly by making it asynchronous, but I don't think it's a huge deal since we never guaranteed it would be synchronous.

src/application-delegate.coffee
as-cii: Basically because any synchronous interaction with the DevTools makes the issue reappear (i.e. we cannot even call `isDevToolsOpened`). I am not sure about what exactly means to interact with the DevTools, but there must be something odd happening in that code path either in Electron or in Chromium.

nathansobo: Interesting. Probably worth a comment in the source.

as-cii:
> This changes the API slightly by making it asynchronous, but I don't think it's a huge deal since we never guaranteed it would be synchronous.

Agreed :+1:. As a side note, even though we were calling it synchronously, it appeared to be asynchronous anyways because we couldn't really rely on calling `openDevTools` to call also `executeJavascriptInDevTools` subsequently.

There is a minor issue with the *metrics* package which is mocking `openDevTools` without returning a `Promise`, thus causing the build to fail: I already fixed this on `metrics 0.53.1` and bumped the version in 972d483. I'm going to :ship: this once the build gets :green_heart: if this looks good.

PR #9796, opened 2015-11-24
Use CommandRegistry to listen for native-key-bindings

### LINKED ISSUE #9548: copy/paste in native inputs no longer works
Hi,

Nuclide debugger uses chrome debugger UI as it front end UI. Atom seems to hijack the copy/paste operations which caused these operations to be disabled in Nuclide watch/console window. 

I think it is regressed in recent Atom releases per user reporting. 

Any idea how to fix this?

Jeffrey

as-cii:
Fixes #9548.

Before this PR, we were using `addEventListener` to listen for commands on elements with the `.native-key-bindings` class. The only way a command (i.e. not a standard event) can be triggered on an HTML element, however, is through `atom-keymap`, but that doesn't happen whenever a native keymap is fully matched.

In the old jQuery days, whenever adding an event listener on the document, we also called `atom.commands.add`, thus allowing to listen on the document and still being notified for events dispatched manually from `CommandRegistry`. After dropping it, though, we didn't update the `handleNativeKeybindings` code path to use `CommandRegistry`, thus not hitting that code path anymore when e.g. copy-pasting because those kinds of event are _synthesized_ (and not triggered on the DOM) whenever a command from `ApplicationMenu` or `ContextMenu` is dispatched.

It would be awesome if anyone could verify that the above rationale is sound, 'cause I'm relatively new to this area of the codebase. :smile_cat: 

### DIFF
--- a/spec/window-event-handler-spec.coffee
+++ b/spec/window-event-handler-spec.coffee
@@ -200,3 +200,34 @@ describe "WindowEventHandler", ->
 
       expect(dispatchedCommands.length).toBe 1
       expect(dispatchedCommands[0].type).toBe 'foo-command'
+
+  describe "native key bindings", ->
+    it "correctly dispatches them to active elements with the '.native-key-bindings' class", ->
+      webContentsSpy = jasmine.createSpyObj("webContents", ["copy", "paste"])
+      spyOn(atom.applicationDelegate, "getCurrentWindow").andReturn({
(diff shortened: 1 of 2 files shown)


nathansobo:
Makes total sense. I think I just ported the jQuery listeners over in the most direct way without really thinking through the implications. Awesome work!

despairblue:
Awesome :+1:

PR #9797, opened 2015-11-24
Upgrade to Electron 0.34.4

### LINKED ISSUE #9454: Gigantic and blurred fonts
Under **Linux**, using new **Atom 1.2.0-beta1** (didn't try -beta0) with Atom Dark UI+Syntax styles, I experience very big font size on File Tree, status bar, file tabs. Editor's font can be adjusted under settings but the font size X appears much bigger (approximately 5pt) than the regular X size of other editors (and previous Atom's versions), and the font is blurred. The font is blurred also in status bar and file tabs, the only place where is not is on file tree explorer (but gigantic anyway). Even MiniMap preview is much larger and blurred.
My styles.less is empty, and Atom produces no errors on startup. Using Atom Material UI+Syntax styles, editor's font is not blurred, but other fonts around are still gigantic and blurred.

Building and installing back versions 1.1.0 or 1.1.0-beta1, solves the issue and everything is back to normal.

Perhaps a bug with new Electron's version?

as-cii:
This should fix https://github.com/atom/atom/issues/9454 by rounding down any device scale factor between 1.0 and 1.3.

Could any Linux user build this locally to check if this solves the problem? 

/cc: @atom/feedback @dappiu @martinb3 @NikoUY @anywhim @fiws

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "0.34.3",
+  "electronVersion": "0.34.4",
   "dependencies": {
     "async": "0.2.6",
     "atom-keymap": "^6.1.1",


maxbrunsfeld:
> This should fix #9454 by rounding down any device scale factor between 1.0 and 1.3.

:tada: Great!

fiws:
Looks good!

## before
screenshot from 2015-11-24 19-32-13


## after
screenshot from 2015-11-24 20-04-35

as-cii:
I am going to :shipit: as the only thing that has changed in Electron is this [scale factor rounding](https://github.com/atom/electron/compare/v0.34.3...v0.34.4).

Thanks!

PR #9798, opened 2015-11-24
Make --wait work on Windows

### LINKED ISSUE #3074: atom --wait doesn't wait
Version 0.118.0 on Windows 7 x64

    PS C:\ > atom --wait myfile.txt

Immediately terminates, but does continue to open the file. Expected result is to wait until the file is closed in the editor.

https://github.com/atom/language-git
and
http://blog.atom.io/2014/03/13/git-integration.html 
say that the syntax that I'm using is correct. Is there a new syntax to enable this or is this feature no longer supported.

raelyard:
especially to enable using Atom as the git commit editor

Fixes #3074

Works in Powershell, cmd, and cygwin

### DIFF
--- a/resources/win/atom.cmd
+++ b/resources/win/atom.cmd
@@ -23,10 +23,11 @@ FOR %%a IN (%*) DO (
 )
 
 rem Getting the process ID in cmd of the current cmd process: http://superuser.com/questions/881789/identify-and-kill-batch-script-started-before
-set T=%TEMP%\sthUnique.tmp
+set T=%TEMP%\atomCmdProcessId-%time::=%.tmp
 wmic process where (Name="WMIC.exe" AND CommandLine LIKE "%%%TIME%%%") get ParentProcessId /value | find "ParentProcessId" >%T%
 set /P A=<%T%
 set PID=%A:~16%
+del %T%
 
 IF "%EXPECT_OUTPUT%"=="YES" (
   SET ELECTRON_ENABLE_LOGGING=YES
--- a/resources/win/atom.cmd
+++ b/resources/win/atom.cmd
@@ -34,7 +34,7 @@ IF "%EXPECT_OUTPUT%"=="YES" (
   IF "%WAIT%"=="YES" (
     "%~dp0\..\..\atom.exe" --pid=%PID% %*
     rem If the wait flag is set, don't exit this process until Atom tells it to.
-	  goto waitLoop
+    goto waitLoop
   )
   ELSE (
     "%~dp0\..\..\atom.exe" %*
@@ -46,7 +46,7 @@ IF "%EXPECT_OUTPUT%"=="YES" (
 goto end
 
 :waitLoop
-	sleep 1
-	goto waitLoop
+  sleep 1
+  goto waitLoop
 
 :end
--- a/resources/win/atom.cmd
+++ b/resources/win/atom.cmd
@@ -1,6 +1,7 @@
 @echo off
 
 SET EXPECT_OUTPUT=
+SET WAIT=
 
 FOR %%a IN (%*) DO (
   IF /I "%%a"=="-f"           SET EXPECT_OUTPUT=YES
@@ -11,13 +12,40 @@ FOR %%a IN (%*) DO (
   IF /I "%%a"=="--test"       SET EXPECT_OUTPUT=YES
   IF /I "%%a"=="-v"           SET EXPECT_OUTPUT=YES
   IF /I "%%a"=="--version"    SET EXPECT_OUTPUT=YES
-  IF /I "%%a"=="-w"           SET EXPECT_OUTPUT=YES
-  IF /I "%%a"=="--wait"       SET EXPECT_OUTPUT=YES
+  IF /I "%%a"=="-w"           (
+    SET EXPECT_OUTPUT=YES
+    SET WAIT=YES
+  )
+  IF /I "%%a"=="--wait"       (
+    SET EXPECT_OUTPUT=YES
+    SET WAIT=YES
+  )
 )
+
+rem Getting the process ID in cmd of the current cmd process: http://superuser.com/questions/881789/identify-and-kill-batch-script-started-before
+set T=%TEMP%\sthUnique.tmp
+wmic process where (Name="WMIC.exe" AND CommandLine LIKE "%%%TIME%%%") get ParentProcessId /value | find "ParentProcessId" >%T%
+set /P A=<%T%
+set PID=%A:~16%
 
 IF "%EXPECT_OUTPUT%"=="YES" (
   SET ELECTRON_ENABLE_LOGGING=YES
-  "%~dp0\..\..\atom.exe" %*
+  IF "%WAIT%"=="YES" (
+    "%~dp0\..\..\atom.exe" --pid=%PID% %*
+    rem If the wait flag is set, don't exit this process until Atom tells it to.
+	  goto waitLoop
+  )
+  ELSE (
+    "%~dp0\..\..\atom.exe" %*
+  )
 ) ELSE (
   "%~dp0\..\app\apm\bin\node.exe" "%~dp0\atom.js" %*
 )
+
+goto end
+
+:waitLoop
+	sleep 1
+	goto waitLoop
+
+:end
--- a/resources/win/atom.sh
+++ b/resources/win/atom.sh
@@ -33,7 +33,6 @@ directory=$(dirname "$0")
 
 WINPS=`ps | grep -i $$`
 PID=`echo $WINPS | cut -d' ' -f 4`
-echo $PID
 
 if [ $EXPECT_OUTPUT ]; then
   export ELECTRON_ENABLE_LOGGING=1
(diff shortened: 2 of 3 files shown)


resources/win/atom.cmd
```diff
@@ -11,13 +12,40 @@ FOR %%a IN (%*) DO (
...
+    SET WAIT=YES
+  )
+  IF /I "%%a"=="--wait"       (
+    SET EXPECT_OUTPUT=YES
+    SET WAIT=YES
+  )
 )
 
+rem Getting the process ID in cmd of the current cmd process: http://superuser.com/questions/881789/identify-and-kill-batch-script-started-before
+set T=%TEMP%\sthUnique.tmp
```
alexandernst: I'd say you must change the literal `sthUnique` by something actually unique.

raelyard: You make a good point - I was thinking it was inconsequential because you wouldn't be doing atom --wait multiple times simultaneously and not worth worrying about it, but with a generic name, there could be another script doing something similar.  Addressing.

resources/win/atom.sh
```diff
@@ -17,9 +31,20 @@ done
 
 directory=$(dirname "$0")
 
+WINPS=`ps | grep -i $$`
+PID=`echo $WINPS | cut -d' ' -f 4`
+echo $PID
```
kevinsawicki: Will the pid be logged to the command line by doing this?

src/browser/squirrel-update.coffee
```diff
@@ -139,7 +139,7 @@ addCommandsToPath = (callback) ->
 
     atomShCommandPath = path.join(binFolder, 'atom')
     relativeAtomShPath = path.relative(binFolder, path.join(appFolder, 'resources', 'cli', 'atom.sh'))
-    atomShCommand = "#!/bin/sh\r\n\"$(dirname \"$0\")/#{relativeAtomShPath.replace(/\\/g, '/')}\" \"$@\""
+    atomShCommand = "#!/bin/sh\r\n\"$(dirname \"$0\")/#{relativeAtomShPath.replace(/\\/g, '/')}\" \"$@\"\r\necho"
```
kevinsawicki: Just curious, why was this `echo` call needed?

resources/win/atom.cmd
```diff
@@ -11,13 +12,41 @@ FOR %%a IN (%*) DO (
...
+    "%~dp0\..\..\atom.exe" %*
+  )
 ) ELSE (
   "%~dp0\..\app\apm\bin\node.exe" "%~dp0\atom.js" %*
 )
+
+goto end
+
+:waitLoop
+	sleep 1
```
kevinsawicki: Do you need to use hard tabs here? Or would two space indent work?

src/browser/squirrel-update.coffee
raelyard: This is something I don't fully understand.  Using atom --wait from Powershell or cmd in Windows works using the cmd script.  Using as the git commit editor executes this Squirrel-generated sh script in Cygwin.  The script process doesn't return 0 and git won't accept the commit without the echo.  I tried exiting, returning the process return code and it didn't work unless there was an echo.  This makes the git commit work.

resources/win/atom.sh
raelyard: Oops - debugging effort that shouldn't have gotten into the commit.

Announcement:
are you sure its not just the line break doing it?

raelyard:
@Announcement Tried several things and it's only happy with the echo.  Line break alone didn't work for me.

amytruong:
Hey @raelyard. Was taking a look at at this PR and had one weird thought about it. Terminating it by ctrl+c seems to terminate it right away rather than wait for you to enter in Y or N. I don't think this is expected behavior but then again, can't be sure and wanted to verify. :smile: 
atom wait

amytruong:
Also, I get the invalid accelerator token but after some googling, I think this is a known issue or considered OK after asking Max. However, just wanted to put it out there just to be noted.
screen shot 2016-01-26 at 12 57 50 pm

raelyard:
@amytruong - I guess that's a little unexpected that Atom goes away if you kill it before answering yes to the prompt - probably due to having a cmd process with Atom as a child.  I don't think there's any reason to be concerned about that and the typical use case will be closing Atom gracefully from the app rather than killing it by cancelling from the command line.  If you find there's a problem caused by it, please let me know, but I don't expect it's an issue.

raelyard:
It looks like the expectation is that those warnings and error messages would be resolved with an Electron update.  It seems that hasn't resolved it, but definitely not related to this.

amytruong:
@raelyard Thanks for the clarification! Whew. :sweat: 
 Tested the following on: Windows 8.1 via cmd and power shell but through GitShell (I have Desktop for Windows on my machine)
- [x] just regularly opening atom with atom --wait
- [x] git config --global core.editor (through GitShell)
- [x] have atom open, open with atom --wait again 
A few other trivial scenarios but let me know if I've missed any big ones. :pray:

mamidenn:
Is this tested on Windows 7? Neither cygwin nor cmd.exe work for me.
* ``sleep`` is not a valid cmd.exe command.
* The ``%time::=%`` hack does not work, because there still remain invalid characters (whitespace, comma).
* The ``.sh`` script will wait forever after closing Atom.

Thanks for your contribution, I would really love to finally see this issue fixed.

raelyard:
@mamidenn I've tested everything I've done with Atom on 7, 8, and 10.  This was a while back, though, so my memory is not perfectly clear - may have had an oversight.  I'll try it again today and see if I have the same result as you.

benogle:
Thanks for looking into this on Win 7 @raelyard!

raelyard:
@mamidenn Confirmed, not working on Windows 7.  Will make it work.

benogle:
> @mamidenn Confirmed, not working on Windows 7. Will make it work.

:heart: :green_heart:

damieng:
Also did some testing here in Windows 7:

- Successfully waits from both PowerShell and Command Prompt
- Command Prompt closes when atom closes (probably not intended?)
- ``find: 'ParentProcessId': No such file or directory`` error running from Git Shell Powershell

martinlong1978:
For me this breaks in Windows 10. From a cmd prompt: 

 - open a file normally  (atom hello.txt)
 - open a second file with -w switch (atom -w bye.txt)
 - close the second file
 - cmd prompt gets killed

The point of opening the normal file is that I am using the -w switch for git, but I typically will have other atom windows open in the background. 

Interestingly it breaks from cmd and bash (on Windows), but works fine in Powershell. 

See gif for demo. 

atom

50Wliu:
@martinlong1978 What Atom version are you using?

martinlong1978:
@50Wliu 1.6.0-beta8

damieng:
Hi @martinlong1978 I also had same issues and made additional fixes. They are in master but just missed beta8.  You can just overwrite atom.cmd and atom.sh in your app-version/resources/cli folder with the contents from master if you want to use the fixes before next release.

https://github.com/atom/atom/tree/master/resources/win

martinlong1978:
@damieng Thanks, I'll give that a try.

martinlong1978:
@damieng Thanks, that worked a charm.

Tyleo:
Hey, just wanted to say that I was experiencing the same issue. @damieng's suggestion worked like a charm though.

avonmoll:
I've been trying to get Atom to work as my git commit editor, however, I'm still having issues.

## My Setup
* Windows 7
* Atom 1.6.0
* Powershell (also tried CMD and Git Bash)
* Using `atom.cmd` from [4d9e2b8](https://github.com/atom/atom/commit/4d9e2b865b1f136bd1400eb3c67cb98dfc7571f2)

## Behavior
* Running `atom --wait` with no other Atom instances open doesn't seem to have the correct behavior. A new atom window opens, but when I close it the PS window doesn't change (i.e. I don't get the `PS C:\Users\...\` prompt back).
* Running `git commit` with no other Atom instances __works perfectly__!
* Running `git commit` with another Atom instance open __always fails__, saying `Aborting commit due to empty commit message`. I tried the following in my .gitconfig:
    * `atom --wait`
    * `atom --wait --new-window`
    * `atom --new-window --wait`
    * `atom --wait --foreground`

The above behavior is the same whether I'm using the latest `atom.cmd` or the one that was released with v 1.6.0.

n3dst4:
@avonmoll I had the same problem, but I have just got it to work using

    atom --new-window --foreground --wait

avonmoll:
Updated to Atom 1.6.1, same behavior, even with that combination, @n3dst4

damieng:
Yes, this is only fixed in 1.7.

avonmoll:
I think I discovered part of the issue. In my .gitconfig I have:
`editor = atom --new-window --foreground --wait`
but this is actually calling __atom.exe__ instead of __atom.cmd__. Is this correct behavior? If it matters, I did not run the installer; rather, I downloaded the ZIP package.

damieng:
Yes, you'll need it to call atom.cmd.

avonmoll:
Atom now waits, but when I save the commit message and close the window (ctrl-w ctrl-w), git tells me the following:

```
error: There was a problem with the editor 'atom.cmd --wait'
Please supply the message using -m or -F option
```

PR #9802, opened 2015-11-24
Correct config.coffee doc references to `scope`

ssorallen:
The scope selector is referred to with the key `scope` on the options
object, not `scopeDescriptor` as the current docs claim. Reference
`scope` correctly.

Mark `options` as optional for `::observe` because the arguments
collection is checked for length, and a length of 2 assumes the second
argument is `callback`. `options` is actually optional.

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -381,8 +381,8 @@ class Config
   # ```
   #
   # * `keyPath` {String} name of the key to observe
-  # * `options` {Object}
-  #   * `scopeDescriptor` (optional) {ScopeDescriptor} describing a path from
+  # * `options` (optional) {Object}
+  #   * `scope` (optional) {ScopeDescriptor} describing a path from
   #     the root of the syntax tree to a token. Get one by calling
   #     {editor.getLastCursor().getScopeDescriptor()}. See {::get} for examples.
   #     See [the scopes docs](https://atom.io/docs/latest/behind-atom-scoped-settings-scopes-and-scope-descriptors)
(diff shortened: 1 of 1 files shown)


src/config.coffee
```diff
@@ -381,8 +381,8 @@ class Config
   # ```
   #
   # * `keyPath` {String} name of the key to observe
-  # * `options` {Object}
-  #   * `scopeDescriptor` (optional) {ScopeDescriptor} describing a path from
+  # * `options` (optional) {Object}
+  #   * `scope` (optional) {ScopeDescriptor} describing a path from
```
ssorallen: See line 400 for `scope`, not `scopeDescriptor` as currently referenced.

src/config.coffee
```diff
@@ -413,7 +413,7 @@ class Config
   # * `keyPath` (optional) {String} name of the key to observe. Must be
   #   specified if `scopeDescriptor` is specified.
   # * `optional` (optional) {Object}
-  #   * `scopeDescriptor` (optional) {ScopeDescriptor} describing a path from
+  #   * `scope` (optional) {ScopeDescriptor} describing a path from
```
ssorallen: See line 435 for `scope`, not `scopeDescriptor`.

src/config.coffee
```diff
@@ -413,7 +413,7 @@ class Config
   # * `keyPath` (optional) {String} name of the key to observe. Must be
   #   specified if `scopeDescriptor` is specified.
   # * `optional` (optional) {Object}
```
50Wliu: Should this be `options (optional)`, not `optional (optional)`?

ssorallen: I noticed that too, but it was unrelated to the `scope` change I was making here. I can add that too though if you like.

50Wliu: I say go for it.  Let's make the docs for this function 100% correct rather than 75% correct :smile:.

ssorallen: Cool, done.

kevinsawicki:
Thanks for fixing this up :+1:

PR #9817, opened 2015-11-25
Make --wait work on Linux

raelyard:
including using Atom as the git commit editor


sadick254:
`--wait` now works on the latest Atom release. This was probably fixed a while back. I will close this and mark it as resolved. 

PR #9831, opened 2015-11-26
Do not override user-defined keymaps when reloading packages

as-cii:
Depends on https://github.com/atom/atom-keymap/pull/105.
Fixes https://github.com/atom/atom-keymap/issues/100.

In this PR we use the new `priority` parameter to prevent packages from overriding user-defined keymaps.
- [x] :arrow_up: atom-keymap after :ship:ping the related PR (this will make specs :green_heart:)


as-cii:
@nathansobo @maxbrunsfeld: do you think we can :ship: this and the related atom-keymap PR?

maxbrunsfeld:
:+1: @as-cii Yeah, they both look good to me.

as-cii:
@maxbrunsfeld: I just realized I cannot publish a new version for `atom-keymap` on npm; could you add me to that package's owner list? In the meantime, I merged atom/atom-keymap#105 into master and this should be ready to go as soon as we bump the dependency. :+1:

Issue #8794, opened 2015-09-15
"(null)" text appears in front-facing message


frewsxcv:
Steps to repeat:
1. Disable/turn off internet
2. Atom -> Check for Updates

![screen shot 2015-09-14 at 7 50 47 am](https://cloud.githubusercontent.com/assets/416575/9874884/dfdae58c-5b77-11e5-8e72-5d450bac491e.png)

PR #9843, opened 2015-11-27
Fix for Issue #8794 Remove 'null' text from front-facing messages

g23-aw:
Remove 'null' from alert messages if the item that is being printed is 'null'
This is the original message when the element is 'null':

messageoriginal
This is the message now:

message

### DIFF
--- a/node_modules/open-on-github/lib/github-file.coffee
+++ b/node_modules/open-on-github/lib/github-file.coffee
@@ -1,0 +1,213 @@
+Shell = require 'shell'
+{Range} = require 'atom'
+parseUrl = require('url').parse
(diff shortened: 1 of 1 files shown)


philipgiuliani:
Hi! This does not belongs to the atom repository. `open-on-github` is a seperate package, that's why it was in node_modules...

Open on Github: https://github.com/atom/open-on-github/
The file: https://github.com/atom/open-on-github/blob/master/lib/github-file.coffee

50Wliu:
Like @philipgiuliani mentioned, this PR should be made to the open-on-github repository, not here.

PR #9853, opened 2015-11-28
Fix #9806 small issue in scrollbar component

### LINKED ISSUE #9806: Scrollbar and file position don't match
Hello there.

After start Atom with last project scroll is up, but redactor position different.

Atom 1.2.4.

Screenshot:

gzzhanghao:
This fix the issue #9806 [Small issue with scroll position]. Since the scrollTop is always 0 when the element is `display: none`, we need to update the visibility of the element before we update the scrollTop.


50Wliu:
Please add specs to this.

damieng:
No response to request to add specs, closing.

Issue #9860, opened 2015-11-28
RangeError: Maximum call stack size exceeded at prepareStackTraceWithRawStackAssignment


gnestor:
Atom 1.4.0-dev

Steps to reproduce: Clone from master, build, and run

Result: 
```
RangeError: Maximum call stack size exceeded
    at prepareStackTraceWithRawStackAssignment (/Users/grant/Downloads/atom/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:165:50)
    at prepareStackTraceWithRawStackAssignment (/Users/grant/Downloads/atom/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:167:10)
    at prepareStackTraceWithRawStackAssignment 
...
handleSetupError @ index.js:61
```

https://github.com/atom/atom/blob/8b76c3d57cf3f74ead5fdd2eee747c3fc18f6d26/static/index.js#L61

50Wliu:
I happened upon this while trying to run specs through `atom --test spec` for atom/atom.  More details to come soon.

50Wliu:
Fixed by #9878.

Issue #9660, opened 2015-11-17
package broken by 1.2.0 update


willwhitney:
My package Hydrogen no longer works after updating Atom to v1.2.0: https://github.com/willwhitney/hydrogen/issues/135

Atom fails to find or parse the `package.json` for a dependency of a dependency of mine.

**Atom Version**: 1.2.1
**System**: Mac OS X 10.11.1
**Thrown From**: [Hydrogen](https://github.com/willwhitney/hydrogen) package, v0.6.5
At Could not find module root given file: "undefined". Do you have a `package.json` file? 

```
Error: Could not find module root given file: "undefined". Do you have a `package.json` file? 
    at Function.getRoot (/Users/will/code/hydrogen/node_modules/zmq/node_modules/bindings/bindings.js:159:13)
    at bindings (/Users/will/code/hydrogen/node_modules/zmq/node_modules/bindings/bindings.js:55:32)
    at Object.<anonymous> (/Users/will/code/hydrogen/node_modules/zmq/lib/index.js:6:30)
    at Module._compile (module.js:434:26)
    at Object.defineProperty.value [as .js] (/Applications/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:190:21)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/will/code/hydrogen/node_modules/zmq/index.js:2:18)
    at Module._compile (module.js:434:26)
    at Object.defineProperty.value [as .js] (/Applications/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:190:21)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/will/code/hydrogen/lib/main.coffee:4:7)
    at Object.<anonymous> (/Users/will/code/hydrogen/lib/main.coffee:1:1)
    at Module._compile (module.js:434:26)
    at Object.defineProperty.value [as .coffee] (/Applications/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:190:21)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
[...]

PR #9878, opened 2015-11-30
Avoid infinite recursion in Error.prepareStackTrace

nathansobo:
Refs #9660

Previously, prepareStackTraceWithStackAssignment could end up calling itself when third-party code assigned Error.prepareStackTrace back to its original value. Now, we short-circuit this process if the rawStack property has already been assigned.

:pear:d on this with @maxbrunsfeld.


50Wliu:
Also fixes #9860.

gnestor:
Working for me 👌

steelbrain:
:heart:

PR #9894, opened 2015-12-01
:memo: Add examples for Windows script/build parameters [ci skip]

BrainCrumbz:
The command line parameters for script/build command where missing examples on how they should be run. That might be usefule for people starting active contribution to the project.

### DIFF
--- a/docs/build-instructions/windows.md
+++ b/docs/build-instructions/windows.md
@@ -31,8 +31,14 @@ Note: If you use cmd or Powershell instead of Git Shell, use a backslash instead
 These instructions will assume the use of Git Shell.
 
 ### `script/build` Options
-  * `--install-dir` - Creates the final built application in this directory.
-  * `--build-dir` - Build the application in this directory.
+  * `--install-dir` - Creates the final built application in this directory. Example (trailing slash is optional):
+```bash
+./script/build --install-dir Z:\Some\Destination\Directory\
+```
+  * `--build-dir` - Build the application in this directory. Example (trailing slash is optional):
+```bash
+./script/build --build-dir Z:\Some\Temporary\Directory\
+```
   * `--verbose` - Verbose mode. A lot more information output.
 
 ## Why do I have to use GitHub Desktop?


docs/build-instructions/windows.md
```diff
@@ -31,8 +31,14 @@ Note: If you use cmd or Powershell instead of Git Shell, use a backslash instead
 These instructions will assume the use of Git Shell.
 
 ### `script/build` Options
-  * `--install-dir` - Creates the final built application in this directory.
-  * `--build-dir` - Build the application in this directory.
+  * `--install-dir` - Creates the final built application in this directory. Example (trailing slash is optional):
+```bash
+./script/build --install-dir Z:\Some\Destination\Directory\
```
50Wliu: If using Git Shell, does the install dir still have to be back-slashed?

BrainCrumbz: We actually found that when opening the shell from Github Desktop (so, the Git Shell), `script\build` works, `script/build` does not work, `./script/build` does not work either. Thinking that it was just us, we did not changed the convention found in the doc page.

---
On a totally different topic: 
This PR just affects an .md file, still Travis CI reports (random) errors similar to:

```
Unable to download https://www.atom.io/api/packages/<some package name, version, etc>

```

where the *random* part is related to which package actually fails to download.

50Wliu: Yeah...in the future, please include `[ci skip]` in your commit messages for commits that only affect documentation.

BrainCrumbz: Thanks, I'll try to remember this bit, I read in contribution guidelines but forgot.

Is that something one should place in each commit message that does not touch code? Or should that be inserted in PR title? PR description?

lee-dohm:
I'm going through old PRs to see if they can be salvaged or should be closed. Is this something you're still working on @BrainCrumbz?

BrainCrumbz:
Not really. But we were not expecting to. If you look at the change, that's just altering a documentation page. So build should not be affected in the first place.

Then, if you look at travis output, 3 builds over 4 run successfully. One fails because build server had problems while [downloading some dependency](https://travis-ci.org/atom/atom/jobs/94523121#L182), so probably nothing to do with the actual change.

All in all, this PR should be merged for us.

lee-dohm:
Thanks. I understand that the build breaks aren't important, I just didn't know if you were planning on updating anything from @50Wliu's feedback.

BrainCrumbz:
You were right on asking in the first place: we decided to stick with convention. Cheers

PR #9910, opened 2015-12-02
Only add absolute paths to project

### LINKED ISSUE #9899: atom.project.addPath : Relative paths add super folder instead.
### Reproduce:

Add a directory to the path using a relative path, e.g.:

```
atom.project.addPath("DirectoryName")
atom.project.addPath("./DirectoryName")
```

When using the relative paths, the final result in the project explorer window is a directory with a name consisting only of a period (e.g: '.') and the contents of the specified directory listed.
screenshot after adding relative path - 1 of 2
Closing and re-opening (Tipping) the containing directory '.', results in all files/directories that are siblings to 'DirectoryName' being added to the project explorer window.  I suspect the super folder of 'DirectoryName' is the directory is the one that is actually added to the project instead and the contents of the specified directory are temporarily/initialized in the listing.

screenshot after closing-reopening - 2 of 2

Compare to using absolute paths:

```
atom.project.addpath("/home/user-name/DirectoryName")
```

The project explorer window has a folder with a name of 'DirectoryName' and all files.  Closing and re-opening (Tipping) the directory leaves the files unchanged.

Atom Version: 1.2.4
System: Ubuntu 14.04.3
### Done When:

Adding relative paths behaves similarly to adding absolute paths does in the project explorer window.

thedaniel:
fixes #9899 - Adding a relative path seems to do ambiguous and incorrect
things. On my system, it attempts to add `/`, on the linked issue it
seems relative to something else. This change rejects relative paths. 

We could try to resolve paths, but relative to what? If there's only one path in the project, we could try relative to that, but if there are multiple, it becomes totally ambiguous.

cc @atom/feedback

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -406,6 +406,15 @@ describe "Project", ->
       expect(atom.project.getPaths()).toEqual([oldPath, newPath])
       expect(onDidChangePathsSpy).toHaveBeenCalled()
 
+    it "doesn't add relative paths", ->
+      onDidChangePathsSpy = jasmine.createSpy('onDidChangePaths spy')
+      atom.project.onDidChangePaths(onDidChangePathsSpy)
+      [oldPath] = atom.project.getPaths()
+
+      atom.project.addPath('a-dir')
+      expect(atom.project.getPaths()).toEqual([oldPath])
+      expect(onDidChangePathsSpy).not.toHaveBeenCalled()
+
   describe ".removePath(path)", ->
     onDidChangePathsSpy = null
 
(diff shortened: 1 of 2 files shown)


thedaniel:
Eh, looks like this simple change unfortunately breaks URI paths and probably more. The breakage that happens when ambiguous relative paths are added will have to be addressed in a more sophisticated way.

lee-dohm:
Is this something that should be closed out @thedaniel?

thedaniel:
@lee-dohm yeah, it would be nice to fix the underlying issue, but my attempt at a quick fix here isn't going to help.

PR #9921, opened 2015-12-03
Update specs for language-javascript changes

winstliu:
`let` and `var` are now tokenized as `storage.type.var.js` instead of `storage.modifier.js`.
# TODO:
- [x] Get core specs to pass
- [x] Update language-html (atom/language-html#97)
- [x] Get package specs to pass

Refs atom/language-javascript#277

Closes #8556

### DIFF
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -611,7 +611,7 @@ describe "TextEditorPresenter", ->
             expect(presenter.getState().hiddenInput.width).toBe 15
 
             expectStateUpdate presenter, ->
-              presenter.getLinesYardstick().setScopedCharacterWidth(['source.js', 'storage.modifier.js'], 'r', 20)
+              presenter.getLinesYardstick().setScopedCharacterWidth(['source.js', 'storage.type.var.js'], 'r', 20)
               presenter.characterWidthsChanged()
             expect(presenter.getState().hiddenInput.width).toBe 20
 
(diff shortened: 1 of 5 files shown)


spec/text-editor-presenter-spec.coffee
```diff
@@ -611,7 +611,7 @@ describe "TextEditorPresenter", ->
             expect(presenter.getState().hiddenInput.width).toBe 15
 
             expectStateUpdate presenter, ->
-              presenter.getLinesYardstick().setScopedCharacterWidth(['source.js', 'storage.modifier.js'], 'r', 20)
+              presenter.getLinesYardstick().setScopedCharacterWidth(['source.js', 'storage.type.var.js'], 'r', 20)
```
50Wliu: I have no idea why these specs required changes.  @as-cii maybe you have an idea?

as-cii: Basically to test the integration with `LinesYardstick` and make sure `TextEditorPresenter` delegates characters coordinates computation to that object instead of doing it on its own (i.e. as we did before measuring from the DOM).

This is dependent on how we tokenize stuff, hence the required change: I wonder if all these tests (and the other ones that use `language-javascript`) should depend on a different, fake language package so that we don't have to perform this annoying task of keeping them up-to-date.

as-cii: /cc: @nathansobo

50Wliu: Ok, that makes sense.  Thanks!

And I definitely agree that the specs should use a dummy package.

PR #9930, opened 2015-12-03
Block Decorations

### LINKED ISSUE #9819: Proposal: Block Decorations
The aim of this document is to write down some of the ideas @nathansobo and I discussed over the past days. I don't think we should implement everything herein proposed in one single, big step. Still, I feel like it's valuable to have a place to discuss the API and to share these ideas with the :globe_with_meridians:.

As of today, Atom doesn't provide a way to create decorations that affect screen rows layout. More specifically, we think it would be :cool: to expose an API that allows users to create something like the following:

block-decorations

You can notice how lines of text break when a block decoration begins and how they resume later when the decoration ends. This hypothethical abstraction would cover several scenarios for package authors, among which:

* Show diffs;
* Show previews;
* And, more generally, provide a richer experience within the editor;

This primitive would cover only decorations in between screen rows; that is, users won't be able to insert inline elements between `<span>`s inside a screen row by using a block decoration.

## API

The code for using this API could look like the following:

```js
let div = document.createElement("div")
let marker = editor.markScreenPosition([7, 0], {invalidate: "never"}) // is "never" the right invalidation strategy?
let decoration = editor.decorateMarker(marker, {type: "block", item: div})
```

I am also thinking to provide a shorthand for the above code:

```js
editor.addBlockDecorationBeforeScreenRow(7, document.createElement("div"))
```

But I guess I'd like to see how this gets used before dealing with this kind of extra details, although I prefer the latter version more in terms of readability. :book:

Also, I feel like we might take this a bit further by enforcing a better separation of concerns between the model and the view, by avoiding to pass an element as the `item` when decorating the marker and using the `ViewRegistry` instead. This seems to stray from how we implemented other decoration APIs, so take this thought with a grain of salt.

### Performance

We would need to know the whole content height in advance in order to show the appropriate vertical scrollbar. Right now, we measure a dummy line's height and calculate `dummyLineHeight * model.getScreenLineCount()`; if we introduced this kind of custom content, however, we'd need to measure block decorations upfront and then calculate height like: `dummyLineHeight * model.getScreenLineCount() + measuredBlockDecorationsHeight`. When having lots of these, loading/refreshing them could become problematic.

On the core side, the scrollbar problem _seems_ the only issue preventing block decorations to be fetched, rendered and measured incrementally (as we do for other decorations); we might even argue that people are used to this sort of progressive content loading on the web, thus making this a non-problem but I'd :heart: to have some more feedback on this. For example, what about minimap? I would think that it will need to know block decorations' height in advance rather than as the user scrolls.

Another optimization tactic we could employ is to split up the measurement tasks in chunks to be executed across multiple animation frames.

To tell the truth, I believe we should just go for the "measure everything upfront" approach first, see how that feels and then re-iterate on that later (if needed). :racehorse: 

### Styling

Because of Shadow DOM, we will have to insert elements outside the shadow root and find a way to teleport them in the right spot inside the editor content. The most straightforward way of doing this could be the following:

```html
<atom-text-editor>
  <#shadow-root>
    <div class='tile-1'>
      <content select=".block-decoration-before-screen-row-1"></content>
      <div data-screen-row='1'></div>
      <div data-screen-row='2'></div>
    </div>
    <div class='tile-2'>
      <div data-screen-row='3'></div>
      <content select=".block-decoration-before-screen-row-4"></content>
      <div data-screen-row='4'></div>
    </div>
  </#shadow-root>

  <div class='block-decoration-before-screen-row-1'></div>
  <div class='block-decoration-before-screen-row-1'></div>
  <div class='block-decoration-before-screen-row-4'></div>
</atom-text-editor>
```

### UX

When a block decoration above the current scroll area changes its height, we need to make sure to adjust the current scroll position seamlessly, so that the editor doesn't scroll down when the user is, for example, editing something.

**Question:** What about the gutter? Should we have block decorations in there as well?

### Naming

Should we consider other names for this API? "Block decoration" seems a pretty sensible choice to me, as it seems to reflect CSS' semantics for elements with a `display: block` style. What do you think? :bulb:

Feedback is super welcome! :sparkles: :bow:

as-cii:
Implements and closes #9819.

block-decorations

The above GIF lasts `~ 50s` and showcases the integration of the block decorations API with the existing primitives within Atom (so, bear with it if it takes a bit to load :sweat_smile:). By default, we attach a decoration to a screen row: when the buffer changes or the underlying marker moves, we make sure to move the block decoration as well so that we can correctly calculate the other visual elements positions with respect to such decoration.

A block decoration can be created through the standard API to decorate markers:

``` js
editor.decorateMarker(
  editor.markScreenPosition([screenRow, 0]),
  {type: "block", item: item}
)
```

Decorations will be (re)measured if:
1. They are visible.
2. They are invisible but they were just added.
3. The editor width changes.
4. `TextEditorElement::invalidateBlockDecorationDimensions(blockDecoration)` is called.
5. They intersect with the buffer splicing region.

There could be some edge cases I haven't taken into account, but here's what I think it's missing:
- [x] Implement a non-linear (logarithmic) algorithm to handle screen-to-pixel positions conversion.
- [x] `offsetHeight` doesn't take margins into account.
- [x] Remove `addBlockDecorationForScreenRow`.
- [x] What happens when resizing the editor? Some block decorations above could need to be resized, thus changing all the items' position after them.
- [x] Correctly handle buffer vs. screen rows by invalidating the splicing region.
- [x] Handle custom gutters as well.
- [x] Inspect the `~ 3-4ms` slowdown when adding/removing many block decorations outside the shadow root.
  - This doesn't seem to be related to block decorations having a unique `id` or being teleported inside the shadow DOM because of `<content>`. Instead, it appears that some styles that cross the shadow boundary make style recalculation a bit slower when many block decorations get swapped in and out from screen. 
- [x] Add missing documentation (mark everything as experimental)
- [x] Allow to position block decorations before or after the marked position
- [x] Update documentation
- [ ] Test-drive! :car:
- [x] **Question:** should we implement a version that works without the shadow DOM? Maybe we can defer this to another PR, when the API stabilizes.


spec/line-top-index-spec.js
```diff
@@ -0,0 +1,189 @@
+/** @babel */
```
as-cii: Maybe the specs here could benefit from some property (randomized) tests, as we do for other data structures such as the marker-index?

src/block-decorations-presenter.js
```diff
@@ -0,0 +1,150 @@
+/** @babel */
+
+const {CompositeDisposable, Emitter} = require('event-kit')
+const LineTopIndex = require('./linear-line-top-index')
+
+module.exports =
+class BlockDecorationsPresenter {
```
as-cii: I opted to move block decorations handling out of `TextEditorPresenter` because it seemed like this code felt quite tight and cohesive. I am a bit unsure about the naming, though, because this doesn't seem to prepare any state for the view but rather to encapsulate information about block decorations + measurements and provide a cleaner API on top of them. Maybe this is just a `BlockDecorationsManager`? (although it feels a bit generic)

philipgiuliani:
Are there still users that have not enabled the Shadow DOM? I already wondered why its not removed yet!

src/block-decorations-presenter.js
```diff
@@ -0,0 +1,149 @@
+/** @babel */
+
+const {CompositeDisposable, Emitter} = require('event-kit')
+
+module.exports =
+class BlockDecorationsPresenter {
```
as-cii: Putting a link to https://github.com/atom/atom/pull/9930#discussion_r46543078 which is not getting shown anymore because some commits were introduced after the comment.

spec/text-editor-component-spec.js
```diff
@@ -1651,6 +1651,106 @@ describe('TextEditorComponent', function () {
     })
   })
 
+  describe('block decorations rendering', function () {
```
mostafaeweda: since this will be transpiled by babel, why not preferring the `() => {}` over `function() {}`

src/linear-line-top-index.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+    })
+
+    this.setMaxRow(this.maxRow + newExtent - oldExtent)
+  }
+
+  topPixelPositionForRow (row) {
+    row = Math.min(row, this.maxRow)
+    let linesHeight = row * this.defaultLineHeight
```
mostafaeweda: revisit `let` --> should be `const` throughout the code.

mostafaeweda: consider adding the eslint rules:
`prefer-const` & `no-const-assign` & `no-var`

mostafaeweda:
I can't express how awesome this is, especially that it aligns with Nuclide's Diff View (avoiding the screen lines to buffer lines hack to make hat happen).

mikeball:
Nice work @as-cii, thank you so much for working on this!

spec/text-editor-component-spec.js
as-cii: Thanks for the suggestion, @mostafaeweda. I am not sure we have any convention to follow in this case, so I preferred consistency with other specs over syntax terseness. We could probably update all the other specs in another PR, although I have no strong preference over one syntax or the other, unless `this` needs to be captured.

as-cii:
This seems to be fine and dandy after a test-drive. I will make sure to keep working on a build with these changes so that I can catch potential mistakes, but overall I feel like this is almost good to go.

@atom/feedback @mostafaeweda @mikeball @philipgiuliani: would love if someone could give this a :eyes: and poke around to ensure this works as expected.

@nathansobo @maxbrunsfeld: I think it's okay to :ship: this without support for non-shadow DOM editors. Adding such support would force us to introduce a different code path (which doesn't make use of `<content>` tags, etc.) and I feel like it's okay not add more complexity until this API stays experimental.

Feedback is super welcome! :tada:

mikeball:
Excellent I was able to get a very basic package(my first) working! A couple of things I noticed.

When the cursor is at the beginning of the line directly below the inserted block, and press enter to get a new line, the block is dragged down and new line inserted above the block. It would seem like the block should stay attached to the line above and the newline should be inserted below the block.

It would also seem logical that if you select text that contains a block with a standard text selection, and delete lines above and below, it should delete blocks contained within the selection?

Is it possible to make the text inside of the block item selectable? I tried various CSS rules but can’t seem to allow text contained in the block to be selectable with the mouse.

Thank you @as-cii !

as-cii:
> Is it possible to make the text inside of the block item selectable? I tried various CSS rules but can’t seem to allow text contained in the block to be selectable with the mouse.

I am not sure about that, but I feel like each package would need to handle this on its own. Block decorations are meant to be a very general purpose abstraction and, as such, I don't think that the core of Atom should provide anything that doesn't fit really common scenarios. Maybe selection is one of those scenarios, but I'd :heart: to see other examples before dealing with it.

> When the cursor is at the beginning of the line directly below the inserted block, and press enter to get a new line, the block is dragged down and new line inserted above the block. It would seem like the block should stay attached to the line above and the newline should be inserted below the block.

> It would also seem logical that if you select text that contains a block with a standard text selection, and delete lines above and below, it should delete blocks contained within the selection?

That's completely up to the marker you are decorating: the decoration will move with it and, therefore, it will follow any convention we have for markers. In this case, I think you'll want to listen for changes on that marker and move it somewhere else when it becomes invalid or your package's domain logic requires so.

mikeball:
>> When the cursor is at the beginning of the line directly below the inserted block, and press enter to get a new line, the block is dragged down and new line inserted above the block. It would seem like the block should stay attached to the line above and the newline should be inserted below the block.

> That's completely up to the marker you are decorating: the decoration will move with it and, therefore, it will follow any convention we have for markers. 

To clarify a bit what I was trying to say, when you insert a marker at say line 3, like this:
```coffeecript
  editor.markScreenPosition([3, 0]), {type: 'block', item: blockitem})
```
it's displayed above line 3 and moves along with line 3. I guess it would seem more natural for inline blocks to be displayed below line 3.

mikeball:
> > Is it possible to make the text inside of the block item selectable? 

> I am not sure about that, but I feel like each package would need to handle this on its own. Block decorations are meant to be a very general purpose abstraction and, as such, I don't think that the core of Atom should provide anything that doesn't fit really common scenarios. Maybe selection is one of those scenarios, but I'd :heart: to see other examples before dealing with it.

Also, completely agree, and it's great as is. I wasn't clear in asking, was trying to ask if it's possible for the package itself to make the contents selectable?

mostafaeweda:
I just tried replacing my diff view hack usage with this block decorations and it mostly worked fine.
However, when the editor has many block decorations and scrolling down fast, it errors:
```
Uncaught TypeError: Cannot read property 'blockHeight' of undefinedadjustNodeBlockHeight @ /Applications/Atom.app/Contents/Resources/app.asar/node_modules/line-top-index/dist/line-top-index.…:292resizeBlock @ /Applications/Atom.app/Contents/Resources/app.asar/node_modules/line-top-index/dist/line-top-index.…:97module.exports.TextEditorPresenter.setBlockDecorationDimensions @ /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-presenter.js:1645module.exports.BlockDecorationsComponent.measureBlockDecorations @ /Applications/Atom.app/Contents/Resources/app.asar/src/block-decorations-component.js:78module.exports.TextEditorComponent.readAfterUpdateSync @ /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:272__bind @ /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:3module.exports.ViewRegistry.performDocumentUpdate @ /Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:219__bind @ /Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:3
```
I'm suspecting this is a race condition in creating the map nodes: `blockEndNodesById` because this only happens the first time where I debugged and found the passed block item, but not the node entry in the map, then further fast scrolls are fine.

screenshot 2015-12-28 11 41 37

as-cii:
> I'm suspecting this is a race condition in creating the map nodes: blockEndNodesById because this only happens the first time where I debugged and found the passed block item, but not the node entry in the map, then further fast scrolls are fine.

@mostafaeweda: thanks so much for trying this out, I really appreciate it! :clap: The issue you have reported should have been fixed in c7a7f0c and I think it was caused by measuring a destroyed block  decoration which was rendered in a previous animation frame and hadn't been removed from the DOM yet.

> it's displayed above line 3 and moves along with line 3. I guess it would seem more natural for inline blocks to be displayed below line 3.

@mikeball: I am wondering if we should allow users to supply an option to specify whether the decoration should follow or precede the marked screen row. Specifically, I am thinking to cases where we want to support adding a block decoration before the first or after the last screen row. With the current approach, we cannot add a block decoration for the latter, which seems a bit unfortunate.

@nathansobo @maxbrunsfeld: thoughts? :thought_balloon:

nathansobo:
> I am wondering if we should allow users to supply an option to specify whether the decoration should follow or precede the marked screen row.

I think that makes sense.

as-cii:
@mikeball @mostafaeweda: we have now introduced the possibility of inserting block decorations after a specified row. I've updated the docs to describe this new feature: basically you now have the possibility of passing a `position` when decorating a marker: `'before'` if you want the decoration to be shown before the marked row, or `'after'` if the decoration should stick after the row. :rainbow: 

I feel like this is pretty stable and shouldn't really affect users/packages that do not interact with block decorations. Still, I'll let this bake for a while before :ship:ping it so it'd be really super awesome if you could try to break it during this period. :boom:

Thanks! :heart:

mostafaeweda:
Cool, will try it - but can't wait any longer to have it in an Atom release

PR #9947, opened 2015-12-05
Don’t force DOM updates in LinesYardstick

nathansobo:
In contemplating a radically different approach to rendering lines based on the new display layers, @maxbrunsfeld and I got a bit confused following the logic of editor DOM updates. We think it will be easier to follow if the `LinesYardstick` never forces a DOM update, but instead requires all lines being measured to be on the DOM as part of its contract.

That means we break `TextEditorComponent.updateSync` into two explicit pre-measurement and post-measurement phases directly in the component so it's more explicit to see, then add visibility checks and synchronous updates on any of the exposed measurement methods on the component.

This might have conflicts with block decorations, but it's going to make it easier for us to maneuver as we render via the `DisplayLayer`'s `TokenIterator` instead of from `TokenizedLines`.

### DIFF
--- a/spec/fake-lines-yardstick.coffee
+++ b/spec/fake-lines-yardstick.coffee
@@ -2,12 +2,8 @@
 
 module.exports =
 class FakeLinesYardstick
-  constructor: (@model, @presenter) ->
+  constructor: (@model) ->
     @characterWidthsByScope = {}
-
-  prepareScreenRowsForMeasurement: ->
-    @presenter.getPreMeasurementState()
-    @screenRows = new Set(@presenter.getScreenRows())
 
   getScopedCharacterWidth: (scopeNames, char) ->
     @getScopedCharacterWidths(scopeNames)[char]
@@ -35,8 +31,6 @@ class FakeLinesYardstick
     left = 0
     column = 0
 
-    return {top, left: 0} unless @screenRows.has(screenPosition.row)
-
     iterator = @model.tokenizedLineForScreenRow(targetRow).getTokenIterator()
     while iterator.next()
       characterWidths = @getScopedCharacterWidths(iterator.getScopes())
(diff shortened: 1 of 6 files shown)


as-cii:
> We think it will be easier to follow if the LinesYardstick never forces a DOM update, but instead requires all lines being measured to be on the DOM as part of its contract.
> 
> That means we break TextEditorComponent.updateSync into two explicit pre-measurement and post-measurement phases directly in the component so it's more explicit to see, then add visibility checks and synchronous updates on any of the exposed measurement methods on the component.

Good work, I like the explicitness of these changes. :+1: 

Originally, I designed the `LinesYardstick` to be an object whose responsibility was to handle reads/updates from/to the DOM transparently, although I agree the interaction with the presenter was somehow complex (and it was an unhealthy, ciclic dependency). Recently, I was exploring an alternative design that you might want to consider:

```js
class TextEditorComponent {
  constructor () {
    let yardstick = new LinesYardstick(this.editor, this.linesComponent, this.grammarRegistry)
    yardstick.onWillMeasureScreenRows(this.flushScreenRowsToMeasure.bind(this))
  }

  // LinesYardstick lazily accumulates screen rows to measure, and calls this
  // method just in time before performing the first DOM read.
  flushScreenRowsToMeasure (screenRows) {
    this.presenter.setScreenRowsToMeasure(screenRows)
    this.linesComponent.updateSync(this.presenter.getPreMeasurementState())
  }

  pixelRectForScreenRange (screenRange) {
    this.linesYardstick.addRowToMeasure(screenRange.start.row)
    this.linesYardstick.addRowToMeasure(screenRange.end.row)
    return this.linesYardstick.pixelRectForScreenRange(screenRange)
  }
}
```

Where `LinesYardstick` gives `TextEditorComponent` the chance to update `LinesComponent` with the needed screen rows. Although slightly less explicit, what I like about this approach is that we don't add another concern to the `TextEditorComponent` object, which would be solely responsible for updating the component (i.e. something it already does). Thoughts? :thought_balloon:

On the other hand, I am not sure how the new changes will impact the rendering layer, so the above design might be incompatible with that new approach. Anyway, I like the proposed changes more than the current version, so feel free to go on with this. :sparkles:

PR #9948, opened 2015-12-05
save config.cson in alphabetic order

dirk-thomas:
Currently the `config.cson` is being written in semi-random order (insertion order of the object properties). That makes it pretty difficult so visually find certain options or compare two `config.cson` files from different atom installations. Especially after changing configuration options back and forth the same set of config values results in a very different file.

By ordering the object properties alphabetically the output of that file becomes deterministic. The same configuration results in the same file. And diffing or comparing changes becomes much easier.
- The first commit only contains the new spec (which obviously fails without any other changes):

```
it writes properties in alphabetical order
    Expected [ 'foo', 'bar', 'baz' ] to equal [ 'bar', 'baz', 'foo' ]. (spec/config-spec.coffee:697:29)
    Expected [ 'foo', 'bar' ] to equal [ 'bar', 'foo' ]. (spec/config-spec.coffee:700:29)
```
- The second commit updates the save function to store the data with a deterministic order.

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -827,6 +827,7 @@ class Config
 
     allSettings = {'*': @settings}
     allSettings = _.extend allSettings, @scopedSettingsStore.propertiesForSource(@getUserConfigPath())
+    allSettings = sortObject(allSettings)
     try
       CSON.writeFileSync(@configFilePath, allSettings)
     catch error
@@ -1189,6 +1190,13 @@ Config.addSchemaEnforcers
 
 isPlainObject = (value) ->
   _.isObject(value) and not _.isArray(value) and not _.isFunction(value) and not _.isString(value) and not (value instanceof Color)
+
+sortObject = (value) ->
+  return value unless isPlainObject(value)
+  result = {}
+  for key in Object.keys(value).sort()
+    result[key] = sortObject(value[key])
+  result
 
 withoutEmptyObjects = (object) ->
   resultObject = undefined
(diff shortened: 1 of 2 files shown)


dirk-thomas:
It took me some time to get the new spec right but now it passes. This should be ready for review now.

src/config.coffee
```diff
@@ -1190,6 +1191,13 @@ Config.addSchemaEnforcers
 isPlainObject = (value) ->
   _.isObject(value) and not _.isArray(value) and not _.isFunction(value) and not _.isString(value) and not (value instanceof Color)
 
+sortObject = (value) ->
+  return value unless _.isObject(value) and not _.isArray(value)
```
maxbrunsfeld: I think it'd be good to use the `isPlainObject` function here instead of `_.isObject and _.isArray`, because that function contains this same loigic, but it looks like it handles a few additional edge cases (e.g. `Color` objects).

maxbrunsfeld:
Other than one very small change I suggested above, this looks great. Thanks for making this better!

src/config.coffee
dirk-thomas: Good point. I have addressed it in https://github.com/atom/atom/commit/fbbcdbf5185f2a08dd1705d7a11767920e662f71#diff-1dca5d0fd1e63d76f43245422a3a18efR1195

maxbrunsfeld:
Thanks @dirk-thomas!

PR #9968, opened 2015-12-07
Deserialize untitled editors

### LINKED ISSUE #942: Auto-save untitled buffers
One feature I love of an increasing number of Mac OSX Lion and above apps is the auto-saving of untitled buffers.  If I quit an app that supports this behavior and re-open it, even untitled buffers (never explicitly named and saved) are restored. This lets me have greater confidence about using these apps as "clipboards," knowing that even if I crash or accidentally close the app, I'll have some reasonably up-to-date copy of the file restored.

Sublime is the front-of-mind example for this in my daily workflow.

as-cii:
Depends on https://github.com/atom/text-buffer/pull/116.
Fixes #942.

Our serialization code already supported saving state for untitled `TextBuffer` instances, but the problem was that we used the file's path to identify and search for buffers in `TokenizedBuffer` and `atom.project`. This PR changes it so that, in addition to paths, we can make use of a uniquely generated id for each `TextBuffer`.

If the `TokenizedBuffer`'s state has a buffer path, we deserialize it exactly as before. Otherwise, we try to search by buffer id among all the project's buffers. It might be tempting to use IDs even when we have a path, but it still seems quite reasonable to just use them as a fallback for two potential problems:
1. Possible weird mismatches or false duplicates (e.g. two buffers have the same path, but different IDs)
2. Even if a buffer is not present in `atom.project`, we still want re-open a buffer at that location when deserializing a `TokenizedBuffer` (I am not sure we have any guarantee that this never happens).

Feedback welcome! :balloon: 


src/text-editor.coffee
```diff
@@ -677,7 +677,7 @@ class TextEditor extends Model
   # this editor.
   shouldPromptToSave: ({windowCloseRequested}={}) ->
     if windowCloseRequested
-      @isModified()
+      false
```
as-cii: Based on #942, we now avoid to prompt the user when a window close request is in place. What are your thoughts on this, @atom/feedback? :thought_balloon:

50Wliu:
> Fixes #942.

Yessssssss

nathansobo:
> It might be tempting to use IDs even when we have a path

I'm unclear on whether either of the concerns you raise can actually happen in a real deserialization scenario. Problems with either would emerge from the workspace getting into an inconsistent state...

We ensure we don't duplicate buffers every time we open a path. So long as we only deserialize from workspaces in which that rule was followed for every open, we shouldn't get duplicates. Also, we pretty explicitly guarantee we serializing a workspace that any buffer we reference in panes is present in our overall set of buffers.

Unless I'm missing something, I think the code would be simpler if we just went all the way with ids. I do think using a random string is smart to avoid needing to update an incrementing counter after deserializing.

as-cii:
> Problems with either would emerge from the workspace getting into an inconsistent state...

Exactly, I wasn't sure about which exact guarantees we had for the deserialization path and went down the safe, defensive path instead. Like, one of my concerns was having a scenario like the following:

1. Open an untitled editor, say _Untitled 1_.
2. Open an untitled editor, say _Untitled 2_.
3. Save _Untitled 1_ into `a.txt`.
4. Save _Untitled 2_ into `a.txt`, thus replacing the previously created file.

I am quite sure that the above steps cause `Project` to serialize two different buffers and, as a result, we would introduce a different behavior with this PR: that is, by using IDs we would use both buffers when deserializing, as opposed to reusing the same one by using paths. 

I guess it's not the end of the world, but more generally speaking, I was concerned about the potential edge cases such as the one above that we might introduce by changing the logical key we use to refer to buffers. Thoughts? Like I said, I was tempted to use IDs only as well and I am :cool: with doing so after hearing what you think.

Thanks!

nathansobo:
> having a scenario like the following:

Interesting. I hadn't considered this edge case. I actually think our current behavior is probably wrong, in that we don't fully enforce one buffer per physical file on disk. Fixing this would probably require some weird juggling where we merge two buffers into one when paths get changed...

I agree this is a weird case, but presumably having two duplicate buffers after a serialization wouldn't make it any worse, right? It's actually a bit weird that we "fix" this edge case with deserialization at this point. Ideally, deserialization reproduces the state of the workspace as faithfully as possible and we fix this at a different layer.

Making a smaller change and leaving the ids in place introduces less risk, but I also think we should consider the total complexity of the system to a minimum. So that leans me toward just using ids.

**Another thing:** Did you consider backwards compatibility? What happens when we deserialize workspaces created with different versions. Conversely, what happens when we fall back a version. Might be nice to make sure that's smooth if it's cheap to do.

as-cii:
> I agree this is a weird case, but presumably having two duplicate buffers after a serialization wouldn't make it any worse, right? It's actually a bit weird that we "fix" this edge case with deserialization at this point. Ideally, deserialization reproduces the state of the workspace as faithfully as possible and we fix this at a different layer.

Agreed. Hopefully we're not missing other edge cases similar to the one above but I think using IDs will be just fine.

> Another thing: Did you consider backwards compatibility? What happens when we deserialize workspaces created with different versions. Conversely, what happens when we fall back a version. Might be nice to make sure that's smooth if it's cheap to do.

Good point. Using IDs requires introducing a conditional where we check for id presence first, falling back to path when we don't find it in the serialized state. Other than that, it seems to work just fine across multiple versions. :+1: 

Please, merge at will after one last :eyes:. Thank you!

benogle:
Thanks for taking this on!

src/text-editor.coffee
50Wliu: Just tested with Notepad++ and it doesn't prompt to save on close.  I really don't see any downside to this because even if someone closed Atom accidentally, they can just open it up again?

benogle: When would these buffers be deserialized? Would it only be when atom is opened in the same project context? i.e. What would happen in this case:

* open atom via `atom /somepath`
* edit untitled buffers
* kill atom 
* open atom via `atom /anotherpath`

Will the scratch buffers be in this new atom window? Or only if I started it via `atom /somepath`?

nathansobo: It will be tied to the workspace state, which is keyed on the project paths. Making it independent of the workspace state is a substantial increase in scope.

as-cii: Exactly, so following the above reproduction steps would only reopen the untitled buffers when starting it via `atom /somepath`. 

I feel like this would be a nice byproduct, as it would allow users to have several untitled editors based on the project they're working on. I cannot think of any other drawback in not prompting to *save* when closing a window, 'cause serializing the state acts a nice safety net in case users want to restore their changes.

I think I should clearly state here that this will affect also editors with a path (i.e. non-untitled), although I think this is actually fine and doesn't change the thoughts I expressed above.

What do you think?

as-cii:
@benogle: :blue_heart: :green_heart: :heart: 

@nathansobo: I just noticed an inconsistency in `TextBuffer::isModified`, which returns `false` when deserializing an unsaved buffer (and not loading it). I fixed it in https://github.com/atom/text-buffer/pull/118, so we may probably want to merge that first and :arrow_up: text-buffer here before :ship:ping this.

as-cii:
Everything is :green_heart: and this seems to work quite nicely after a test-drive, so I'll go ahead and :shipit: (we can refine this in a later PR if needed).

bhadra:
I upgraded to 1.4.0-beta0 now. I do not see #942 fixed on Windows 8.1.

1. Open Atom
2. Create an untitled buffer. Type some random text.
3. Close and reopen Atom
4. The untitled buffer does not show up.

In case of Sublime Text 3, the untitled buffer shows up as expected.

Do I have to change any settings?

50Wliu:
@bhadra does it still not work in safe mode (`atom --safe`)?  If so, please create a new issue for that.

nathansobo:
Keep in mind that the untitled buffers are associated with the workspace, which is saved based on the open folders in the current project.

bhadra:
@50Wliu It does not work in safe mode. I have followed the test case as described in my previous message.

@nathansobo You are right. As mentioned by @as-cii,  untitled buffers are persisted if and only Atom is re-started via atom /somepath. This is significantly limited compared to Sublime Text. In Sublime Text, untitled buffers are not associated with the workspace and are persisted independent of how Sublime Text is re-started.

schmurfy:
I know the issue is closed but I just discovered save-session was merged except I don't see any of its behavior in current Atom version...

In an open project I create a new tab, type some text and then quit Atom: it asks to save Oo
What was included exactly and how to enable it ?

mnquintana:
@schmurfy This is available now in the beta channel and will be available in the next stable release of Atom (1.4.0)

schmurfy:
thanks :)

mattpker:
This is a great new feature and I really love it, but I think more work needs to be done when you have not opened a project. A lot of the time when I need to jot down some notes I will open a new atom window (not apart of a project). When I go to close that window or restart my computer, the pop up to save the untitled notes was really great, because a lot of the time I would forget to save a note that I actually needed to keep. Because this was not a part of a project, I just lost all my notes. I think having a feature that only works on projects is a bit of an oversight, it gets you in the mindset that "It's okay if I close this without saving, it will be back when I re-open it." and then when it was not opened as a project it is gone with no prompt warning you of this. 

I suggest adding the prompt back for untitled tabs that are not apart of a project, or making all non project tabs all be apart of some "no project" project or something. I hope that makes sense, let me know if you have any questions.

asbjornu:
Why does `untitled` files need to be associated with a project? What about files saved outside a project? I think both should be handled the same way, perhaps through an implicit project that always exist, to hold `untitled` files and files outside of any explicit projects.

This implicit project and all of its files (tabs) should be opened every time Atom is started, no matter how it is started, but only in the first window. If several windows are opened after that, they should contain their own implicit project that can have files added to them.

If Atom is closed with several windows open, all windows should be re-opened the next time Atom is started (no matter how it is started). This will preserve all `untitled` tabs and unsaved changes to previously saved files. It may become cluttered to have several windows open like this, but that's better than losing work. It should also be as easy as dragging tabs from one window to another to move files from one implicit project to another. When the last tab is dragged from a window, its implicit project should be deleted.

And lastly, to solve the several `untitled` tabs problem, how about doing like Sublime and Apple's Notes app and naming the tabs after the first line of the content within them?

splashx:
Are you guys fixing `untitled` outside a project? I'm running 1.7.3 and still can't see this behaviour:
- open atom, `untitled` opens
- write `Hello World`
- CMD+Q on Mac
- Open atom 

`untitled` is blank.

akamuza:
1.8.0 - the same behavior with untitled outside a project - the tab is blank on reopening.

mnquintana:
@splashx That's being tracked in https://github.com/atom/atom/issues/10474

Since this has already been merged and there's already another issue open tracking further improvements, I'm going to lock this to keep noise levels down.

PR #9990, opened 2015-12-09
Add a core config `core.automaticallyUpdate` to disable Atom autoupdate.

scv119:
This PR does following:
- Added a config `core.automaticallyUpdate`. This config is enabled by default, but if user disable it, the new Atom window won't check update/schedule update on startup.
- I didn't observer the config changes after initialization, which means if user change the automaticallyUpdate config, it won't change the autoupdate behavior until user quit and start Atom again, but we could observer this changes if we think it's necessary.
- Remove the build option `--no-auto-update` and revert all related code introduced in https://github.com/atom/atom/pull/8721

What I have tested:
- run the `scrip/build` to build a new Atom.app.
- noticed the `enableAutoupdate` is selected by default, and the checkUpdate button showed up in Atom menu.
- disable `enableAutoupdate` option, open a new Atom window, the checkUpdate button disappeared from Atom menu.
- Note that since the build of Atom don't have certs, so I can't tell if the download is happening or not.


Zireael07:
Great!

philipgiuliani:
If i have autoupdate disabled, its still possible to check and install manually an update?

scv119:
@philipgiuliani My current plan is hide the `checkUpdate` button as well, so there is no way manually trigger the update.

benogle:
cc @maxbrunsfeld @nathansobo

Zireael07:
@scv119: If I were you, I wouldn't hide the button. Disabled autoupdate should still allow me to check for updates manually if I want to. This way, autoupdate wouldn't eat my limited bandwidth or slow my net down to a crawl when I want to do other stuff, but I could check for updates once I was back to the unlimited connection and/or stopped watching YT. (just an example, btw)

maxbrunsfeld:
> Disabled autoupdate should still allow me to check for updates manually if I want to.

I am ok with that. @scv119 What do you think?

scv119:
Makes sense for me. I will update the PR.

scv119:
Addressed the comments:
- Disable the config still allow user to manually check update
- Toggle disable/enable of the config will schedule/cancel the autoupdate interval.

philipgiuliani:
Really cool :+1: Like this, the feature makes sense to me!

src/config-schema.coffee
```diff
@@ -100,6 +100,10 @@ module.exports =
         description: 'Automatically open an empty editor on startup.'
         type: 'boolean'
         default: true
+      enableAutoupdate:
```
50Wliu: `enableAutoUpdate`?

src/config-schema.coffee
```diff
@@ -100,6 +100,10 @@ module.exports =
...
+        description: 'Automatically update Atom when new release is available.'
```
50Wliu: ...`when a new release`

src/config.coffee
```diff
@@ -790,7 +790,7 @@ class Config
     @watchSubscription = null
 
   notifyFailure: (errorMessage, detail) ->
-    @notificationManager.addError(errorMessage, {detail, dismissable: true})
+    @notificationManager.addError(errorMessage, {detail, dismissable: true}) if @notificationManager
```
kevinsawicki: This might look better as:

```
@notificationManager?.addError(errorMessage, {detail, dismissable: true})
```

src/config-schema.coffee
```diff
@@ -100,6 +100,10 @@ module.exports =
         description: 'Automatically open an empty editor on startup.'
         type: 'boolean'
         default: true
+      enableAutoUpdate:
```
kevinsawicki: What do you think about just making this `automaticallyUpdate` instead of `enableAutoUpdate`.

Looks like most of (or all) the other config items don't start with `enable`, such as `scrollPastEnd` instead of `enableScrollPastEnd`.

kevinsawicki:
While looking at the fs calls in `config.coffee` I came across these lines: https://github.com/scv119/atom/blob/test/src/config.coffee#L752-L754

If these lines trigger an fs exception in the main process then the app will fail to load and fail to update so the person will be in a permanently bad state. I know they are unguarded currently in the render process but it would seem worth wrapping them in  a `try` / `catch` in this pull request to ensure they don't impact people.

maxbrunsfeld:
Note to self (:tophat: @kevinsawicki): Update [this text](https://github.com/atom/notifications/blob/4716ca24f74c685f6db85447e432e8a173d86a67/lib/notification-element.coffee#L215-L219) to remind people to update Atom before reporting issues.

scv119:
hmm, I might have done something wrong

maxbrunsfeld:
> I'm not sure what's the best way to handle the exception in this case.

I think that what you did is fine, but could you add a test for that handling? It would be good to have a test that sets up one of those file system APIs to throw an exception, and assert that the config keeps working consistently.

Something like this:

```coffee
describe ".loadUserConfig", ->
  # ...

  describe "when the config file fails to load", ->
    beforeEach ->
      spyOn(fs, 'existsSync').andCallFake -> throw new Error("EPERM")

    it "does not try to save later changes to disk", ->
      atom.config.loadUserConfig()
      atom.config.set("foo.bar", "baz")
      advanceClock(100)
      expect(atom.config.save).not.toHaveBeenCalled()
      expect(atom.config.get("foo.bar")).toBe "baz"
```

scv119:
@maxbrunsfeld  test added.

maxbrunsfeld:
Alright @scv119, thanks for this!

PR #9997, opened 2015-12-10
Wrap line at boundary if it includes a CJK character

### LINKED ISSUE #9447: Provide different soft-wrapping rules for CJK languages
Text:

```
文字文字 English 文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字
```

1.2.0-beta0:

When mixing English and Chinese(or Japanese) with whitespace, Chinese(or Japanese) sentence can be break, instead of break in whitespace. 

Browser:

> 文字文字 English 文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字文字

as-cii:
Fixes #9447.

This PR provides a different strategy for wrapping lines whenever CJK characters are involved. With this PR we try to detect a CJK character before or exactly at the desired wrap boundary: if we do, we perform the wrap; otherwise, we use the old line-breaking algorithm.

Although not compliant with the Unicode spec, this allows users to mix CJK characters with Latin ones without the need to introduce a complex line breaking logic.

Before: 

screen shot 2015-12-10 at 10 29 22

After: 

screen shot 2015-12-10 at 10 29 44


as-cii:
@chloerei @yukihop: it would be great to hear what you think and, even more awesome if you could build and try this out. :sparkles:

chloerei:
Example looks good to me.

I want to test in local, but have a problem with building:

```
Node: v0.12.9
npm: v3.5.1
Installing build modules...
WARN engine prettyjson@0.8.1: wanted: {"node":">= 0.6.0 < 0.11.0"} (current: {"node":"0.12.9","npm":"3.5.1"})
npm ERR! Linux 3.13.0-67-generic
npm ERR! argv "node" "/usr/local/bin/npm" "--userconfig=/vagrant/build/.npmrc" "install" "--loglevel" "error"
npm ERR! node v0.12.9
npm ERR! npm  v3.5.1
npm ERR! path /vagrant/build/node_modules/npm/node_modules/columnify/node_modules/strip-ansi/cli.js
npm ERR! code ENOENT
npm ERR! errno -2

npm ERR! enoent ENOENT, chmod '/vagrant/build/node_modules/npm/node_modules/columnify/node_modules/strip-ansi/cli.js'
npm ERR! enoent ENOENT, chmod '/vagrant/build/node_modules/npm/node_modules/columnify/node_modules/strip-ansi/cli.js'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent
```

I'm not familiar with npm, after google I can not find some similar problems, any ideas?

spec/text-utils-spec.coffee
```diff
@@ -74,3 +74,23 @@ describe 'text utilities', ->
       expect(textUtils.isKoreanCharacter("ㄼ")).toBe(true)
 
       expect(textUtils.isKoreanCharacter("O")).toBe(false)
+
+  describe ".isCjkCharacter(character)", ->
```
nathansobo: Style nitpick: `isCJKCharacter`

src/tokenized-line.coffee
```diff
@@ -321,8 +321,11 @@ class TokenizedLine
     return unless maxColumn?
     return unless @text.length > maxColumn
 
-    if /\s/.test(@text[maxColumn])
-       # search forward for the start of a word past the boundary
+    if isCjkCharacter(@text[maxColumn])
+      # always wrap when a CJK character is at the wrap boundary
+      return maxColumn
```
nathansobo: Seems like there would still be an issue if you had latin characters, then CJK characters, then more latin at the wrap boundary. Would we want to use this approach if the line contained any CJK characters at all? I guess you get into a weird state where someone might be writing principally in a latin character set but used a single CJK character for some reason, which this more limited approach avoids.

nathansobo:
Thanks for jumping on this!

src/tokenized-line.coffee
as-cii: Yeah, I tried to reason about always wrapping when the line contained CJK characters and the edge case you're pointing out is basically the reason why I avoided doing so. In particular, I was trying understand what we would have done by using that approach with the following javascript line of code:

```js
/* 화화화화화화화화화화화화화화화화화화화화화화화화 */ const hello = "OK"
```

If the wrap occurs in the middle of either `const` or `hello`, for instance, I don't think we'd want to break those words in two like this:

```js
/* 화화화화화화화화화화화화화화화화화화화화화화화화 */ con
st hello = "OK"
```

chloerei: I think Chrome render result is fine, break on whitespace and between CJK characters.

yukihop:
@as-cii Thanks, the new screenshot looks perfect to me!

chloerei:
I success build atom on local by downgrade npm to 1.4.28.

It is a big improvement. But I found it will leave a large space  in some situation, for example:

Line length 80

```
中文中文中文中文中文中文中文中文 English 中文中文中文中文中文中文中文中文中文中文中文中文。En
```

When add one more English character, it becomes:

```
中文中文中文中文中文中文中文中文 English
中文中文中文中文中文中文中文中文中文中文中文中文。Eng
```

as-cii:
@chloerei: what behavior would you expect in that case? Seems like we should wrap at the beginning of the english word, as Chrome does.

I have pushed a couple of commits that refine the previous behavior to address your concerns. Could you pull this again and try it out? Thanks so much for helping out with this, @chloerei! :sparkles:

chloerei:
:+1: I test with my blog post and feel it's great!

as-cii:
Great, I am going to :ship: this as soon as the build gets :green_heart:.

Thank you!

yukihop:
Hi, I'm still having trouble building Atom on Windows, but judging from the committed code, here's how I feel. I'm totally new to Atom dev, and I've never treated Unicode seriously, so I might be wrong.

- `isJapaneseCharacter` function seems to also cover [CJK symbols and punctuation](https://en.wikipedia.org/wiki/CJK_Symbols_and_Punctuation), which is not really Japanese-specific.
- Aside from this, `isJapaneseCharacter` basically checks if the character is a [hiragana](https://en.wikipedia.org/wiki/Hiragana_(Unicode_block)) or [katakana](https://en.wikipedia.org/wiki/Katakana_(Unicode_block)), but hiragana/katakana are only a part of Japanese characters. This function may be renamed to `IsJapaneseKanaCharacter` or `IsJapaneseKana`.
- Shouldn't `isCJKUnifiedIdeograph` also cover CJK unified ideographs extensions [described here](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs)? These extensions contain very rare characters (kanji) which are unlikely to cause big problems in real life situations, though.
- I believe at least one hiragana and one katakana character should be included in the `isCJKCharacter` test. Naturally-written Japanese sentences consist of many of those characters.
- Despite what I've said, I kind of feel all the characters at least from U+3000 to U+9FFF in [this table](https://en.wikipedia.org/wiki/Unicode_block) can be treated in the same way as ordinary kanji (U+4E00 - U+9FFF) (i.e., line breaks can happen anywhere between these characters)

as-cii:
Thanks @yukihop for writing down these thoughts, I appreciate them a lot! :+1: 

> isJapaneseCharacter function seems to also cover CJK symbols and punctuation, which is not really Japanese-specific.

> Aside from this, isJapaneseCharacter basically checks if the character is a hiragana or katakana, but hiragana/katakana are only a part of Japanese characters. This function may be renamed to IsJapaneseKanaCharacter or IsJapaneseKana.

> I believe at least one hiragana and one katakana character should be included in the isCJKCharacter test. Naturally-written Japanese sentences consist of many of those characters.

Those are really good points. :+1: I will address them shortly.

> Despite what I've said, I kind of feel all the characters at least from U+3000 to U+9FFF in this table can be treated in the same way as ordinary kanji (U+4E00 - U+9FFF) (i.e., line breaks can happen anywhere between these characters)

> Shouldn't isCJKUnifiedIdeograph also cover CJK unified ideographs extensions described here? These extensions contain very rare characters (kanji) which are unlikely to cause big problems in real life situations, though.

Yeah, that was basically the reason why I did not include them originally. I think, however, we should add them just to be sure to cover every edge case and be more bulletproof. Will change this PR to fix those as well.

Thanks a lot, @yukihop! Having a second pair of :eyes: on CJK characters is definitely awesome!

as-cii:
Okay, I slightly adjusted names and partially extended the ranges checked in `isCJKUnifiedIdeograph`. I will open another PR after :ship:ping this one which addresses CJK Unified Ideograph Extensions Tables.

bicycle1885:
@as-cii Thank you! I hope this fix will be included in the next bug fix release. I really need this one :pray:

as-cii:
Yeah, I think we are going to release this one under `1.5.0` and `1.5.0-beta0`.

PR #10004, opened 2015-12-10
There is no ES6 in compile-cache.js

maxbrunsfeld:
Fixes https://github.com/atom/apm/issues/476

### DIFF
--- a/src/compile-cache.js
+++ b/src/compile-cache.js
@@ -161,8 +161,7 @@ require('source-map-support').install({
 Error.stackTraceLimit = 30
 
 var prepareStackTraceWithSourceMapping = Error.prepareStackTrace
-
-let prepareStackTrace = prepareStackTraceWithSourceMapping
+var prepareStackTrace = prepareStackTraceWithSourceMapping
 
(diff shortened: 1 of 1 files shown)


src/compile-cache.js
```diff
@@ -1,5 +1,10 @@
 'use strict'
 
+// For now, we're not using babel or ES6 features like `let` and `const` in
+// this file, because `apm` requires this file directly in order to pre-warm
+// Atom's compile-cache when installing or updating packages, using an older
+// version of node.js
```
mnquintana: Why do we use an older version of Node for that? I remember @thomasjo and I running into that with apm in the past

maxbrunsfeld:
CI failed with an atom.io 503, but the build before the comment passed, so I'm going to merge.

src/compile-cache.js
maxbrunsfeld: @mnquintana There's an open PR to upgrade apm's node version to 4.x, but there are some issues with compatibility with npm. I forget the details. We definitely would like to upgrade it soon.

PR #10010, opened 2015-12-11
Fix visible screen row line count checks.

CentricStorm:
Visible screen row calculations should use getScreenLineCount so that correct values are found when text is wrapped or folded.


alexandernst:
Atom's devs will probably ask for unit tests for this

CentricStorm:
Fixed in #10967.

PR #10027, opened 2015-12-11
Use internal scroll assignment methods when committing logical positions

nathansobo:
This prevents an additional pending scroll top/left assignment and cuts directly to adjusting the positions, saving us an update cycle. It prevents the only major case of emitting an update event during state computation.

@as-cii Am I missing anything here?


as-cii:
Looks good to me! Thanks, @nathansobo. :fireworks:

PR #10028, opened 2015-12-11
Remove non-essential update guard flags from TextEditorPresenter

nathansobo:
The only extremely expensive operation we need to avoid unless it's needed is decoration computation. The rest of these flags don't really warrant the added complexity for the tiny amount of computation they save. Eventually, we'll make the decoration computation so fast so that we can always execute the entire code path whenever we update the editor. This will be easier to reason about both in terms of adding functionality and thinking about performance in a variety of circumstances. Removing the bulk of the flags gets us part of the way there.

/cc @as-cii


as-cii:
:art: :+1: :shipit:

PR #10029, opened 2015-12-11
Escape paths in bootstrap and clean scripts

puls:
I just tried to check out and build the repository, and apparently I'm the only weirdo who has tried to do this in a folder with spaces or other special characters in its name in a year and a half?


damieng:
Build system has been rewritten, this PR is now out of date. Let us know if the new build system still gives you issues.

PR #10031, opened 2015-12-12
:memo: Add tip about Keybinding Resolver to keymap.cson

### LINKED ISSUE #9994: Keybinding not consistent. Works one way through the command palette, different through the keybinding
I'm trying to invoke a command from the `atom-turbo-javascript` package:
>End New Line CTRL-ENTER
Terminates the current line with a colon or semicolon, followed with a new line. A comma is inserted when the cursor is inside an object literal, otherwise a semicolon is inserted.

I want the keybinding to be <kbd>⌘</kbd>+<kbd>⇧</kbd>+<kbd>⏎</kbd> and so in my `keymap.cson` I have:

    'atom-text-editor':
        'cmd-shift-;': 'turbo-javascript:end-line'
        'cmd-shift-,': 'turbo-javascript:end-line-comma'
        'cmd-shift-enter': 'turbo-javascript:end-new-line'

Now, if I use cmd-shift-enter I get (i guess) the core functionality: a newline above the caret line. But if I search the same command in the command-palette, I see that it correctly has cmd-shift-enter and executing through the palette, it works!

So using the same shortcut the palette acknowledges directly, it gives me a new line above, not the turbo-javascript functionality. (in shift-cmd-enter order also doesn't work).

lee-dohm:
Fixes #9994

### DIFF
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -24,7 +24,7 @@
 # If you're having trouble with your keybindings not working, try the
 # Keybinding Resolver: Cmd+. on OS X and Ctrl+. on other platforms. See the
 # Debugging Guide for more information:
-# * https://atom.io/docs/v1.3.1/hacking-atom-debugging#check-the-keybindings
+# * https://atom.io/docs/latest/hacking-atom-debugging#check-the-keybindings
 #
 # This file uses CoffeeScript Object Notation (CSON).
 # If you are unfamiliar with CSON, you can read more about it in the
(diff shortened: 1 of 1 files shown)


dot-atom/keymap.cson
```diff
@@ -21,7 +21,12 @@
 # * https://atom.io/docs/latest/using-atom-basic-customization#customizing-key-bindings
 # * https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 #
+# If you're having trouble with your keybindings not working, try the
+# Keybinding Resolver: Cmd+. on OS X and Ctrl+. on other platforms. See the
+# Debugging Guide for more information:
+# * https://atom.io/docs/v1.3.1/hacking-atom-debugging#check-the-keybindings
```
johnmuhl: Should this be https://atom.io/docs/latest/hacking-atom-debugging#check-the-keybindings instead?

lee-dohm: :anguished: Yes ... told myself to do that and totally missed it!

benogle:
I would wrap the keybindings in backticks, but other than that :+1:

PR #10034, opened 2015-12-12
Add "Yes for All" and "No for All" buttons to prompts from pane:close

Areadrill:
When using the _pane:close_ command, if there're more than one unsaved items open in the pane, the save prompt will now display two additional buttons as seen in the GIF below:

newbehavior

This also fixes an issue where using pane:close when only one pane exists makes atom prompt to save an unsaved item twice, as you can see below. The issue was verified on Windows, Mac and Linux.

 This was due to @promptToSaveItem being called twice in different methods in the case of there only being one pane. 

Also, thank you to @joaomnb for contributing to this new functionality.

### DIFF
--- a/src/pane-container.coffee
+++ b/src/pane-container.coffee
@@ -137,7 +137,7 @@ class PaneContainer extends Model
 
     for pane in @getPanes()
       for item in pane.getItems()
-        unless pane.promptToSaveItem(item, options)
+        unless pane.promptToSaveItem(item, options) is 0
           allSaved = false
           break
 
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -439,7 +439,7 @@ class Pane extends Model
       @emitter.emit 'will-destroy-item', {item, index}
       @container?.willDestroyPaneItem({item, index, pane: this})
       if promptToSave
-        if @promptToSaveItem(item)
+        if @promptToSaveItem(item) is promptToSaveReturns.Single
           @removeItem(item, false)
           item.destroy?()
           true
@@ -460,15 +460,21 @@ class Pane extends Model
     @destroyItem(item) for item in @getItems() when item isnt @activeItem
     return
 
+  promptToSaveReturns =
+    Single: 0
+    Cancel: 1
+    NoForAll: 2
+    YesForAll: 3
+
   promptToSaveItem: (item, options={}, multiple=false) ->
-    return true unless item.shouldPromptToSave?(options)
+    return promptToSaveReturns.Single unless item.shouldPromptToSave?(options)
 
     if typeof item.getURI is 'function'
       uri = item.getURI()
     else if typeof item.getUri is 'function'
       uri = item.getUri()
     else
-      return true
+      return promptToSaveReturns.Single
 
     chosen = @applicationDelegate.confirm
       message: "'#{item.getTitle?() ? uri}' has changes, do you want to save them?"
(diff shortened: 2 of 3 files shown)


lee-dohm:
Is it possible to make this work for Close All Tabs and Close Other Tabs as well?

Areadrill:
@lee-dohm I've been taking a look at [atom/tabs](https://github.com/atom/tabs/), and I can adapt close all tabs to use what I've done here with minimal effort. Close other tabs would be trickier, since it'd have to take in to account what tab not to prompt for. Both are definitely feasible, tough I think it's better to wait and see if this gets accepted before trying to make other packages work with it.

src/pane.coffee
```diff
@@ -468,12 +473,14 @@ class Pane extends Model
...
       message: "'#{item.getTitle?() ? uri}' has changes, do you want to save them?"
       detailedMessage: "Your changes will be lost if you close this item without saving."
-      buttons: ["Save", "Cancel", "Don't Save"]
+      buttons: if multiple then ["Save", "Cancel", "Don't Save", "No For All", "Yes For All"] else ["Save", "Cancel", "Don't Save"]
 
     switch chosen
       when 0 then @saveItem(item, -> true)
       when 1 then false
       when 2 then true
+      when 3 then return 1
```
joshaber: I don't think we should use magic numbers and mix bools and numbers. Could we use string constants, or number constants, or something more descriptive?

joshaber:
@Areadrill Thanks for opening this up and sorry for taking so long to review it! This looks good apart from the one comment above.

Areadrill:
@joshaber Just to clarify, you mean using string constants just on the two return values I added in, right? Because changing all of them would disrupt the functionality of other components that depend on this method. In that case yes I totally agree and can make that change in a couple of hours.

mel111:
stoppp sending or i will report !!!!!!!!!!!!

joshaber:
>Just to clarify, you mean using string constants just on the two return values I added in, right? Because changing all of them would disrupt the functionality of other components that depend on this method.

Given that this isn't marked public, I think we're OK to change what this returns (obviously so long as we update all the callsites). @atom/core do I have that right?

nathansobo:
Yeah, it's an internal method so fixing call sites in core should be sufficient. If you're worried you can use [atom-package-downloader](https://github.com/atom/atom-package-downloader) and scan packages for calls, but it's hard to see why someone would call this outside of core.

Areadrill:
Alrigthy then, I'll get to work either tonight or tomorrow night, depending on how my college projects are going.

joshaber:
Great, thanks @Areadrill!

Areadrill:
@joshaber Would this be better? I didn't know if it was OK to add an enumeration in but I think it makes everything descriptive enough. I wracked my brain trying to find a good name for the 1st return value, but that's the best that came out...

src/pane-container.coffee
```diff
@@ -137,7 +137,7 @@ class PaneContainer extends Model
 
     for pane in @getPanes()
       for item in pane.getItems()
-        unless pane.promptToSaveItem(item, options)
+        unless pane.promptToSaveItem(item, options) is 0
```
joshaber: Could we export `promptToSaveReturns` so that we can use the appropriate value here?

joshaber:
Thanks @Areadrill! That looks good, just wondering about the one other spot.

Areadrill:
@joshaber Forgive me, but I haven't messed with Coffeescript in a long time (or enough for that matter). How would I go about exporting stuff? A google search didn't help me...

PR #10035, opened 2015-12-12
Clear the require cache for a package files and dependencies

abe33:
This is an attempt to fix the issues after a package update by clearing the `require.cache` object of all the paths related to the package.

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -225,6 +225,20 @@ describe "PackageManager", ->
         atom.packages.unloadPackage(pack.name)
         expect(atom.packages.isPackageLoaded(pack.name)).toBeFalsy()
 
+      it "clears the require cache entries for all the package files and dependencies", ->
+        pathsForPackage = (pack) ->
+          Object.keys(require.cache).filter (p) ->
+            p.indexOf(pack.path + path.sep) is 0
+
+        pack = atom.packages.loadPackage('package-with-main')
+        require.cache[path.join(pack.path, 'main-module.coffee')] = {}
+        require.cache[path.join(pack.path + '-other', 'main.coffee')] = {}
+
+        expect(pathsForPackage(pack).length > 0).toBeTruthy()
+        atom.packages.unloadPackage(pack.name)
+        expect(pathsForPackage(pack).length > 0).toBeFalsy()
+        expect(require.cache[path.join(pack.path + '-other', 'main.coffee')]).not.toBeUndefined()
+
     it "invokes ::onDidUnloadPackage listeners with the unloaded package", ->
       atom.packages.loadPackage('package-with-main')
       unloadedPackage = null
--- a/src/package-manager.coffee
+++ b/src/package-manager.coffee
@@ -403,6 +403,9 @@ class PackageManager
 
     if pack = @getLoadedPackage(name)
       delete @loadedPackages[pack.name]
+      Object.keys(require.cache)
+      .filter (p) -> p.indexOf(pack.path + path.sep) is 0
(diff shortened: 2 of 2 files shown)


abe33:
Would love to know if you see any pitfalls with this approach?

cc/ @atom/feedback

lee-dohm:
I would think that the major pitfall would be slower unload and reload times. So anything that unloads or reloads a lot of packages would be an issue? Particularly `window:reload` might be slower. Some people use this a *lot* for package development workflow.

I don't think that should block this (unless it is egregious), but we should be aware.

abe33:
Aren't we already starting a new node process on `window:reload`? Because if the `require.cache` is persisted that means that no code will be effectively reloaded, and that's not what I experience (as you point out, this is part of the development workflow for pretty much every package devs).
I don't think clearing the cache with `delete` costs that much, but we could make that behavior available only with an argument passed to `unloadPackage` (that would default to false).

I'm also starting to wonder if the `compile-cache` shouldn't be cleared as well? I'm not sure how it works in regard with node's `require` and its cache.

lee-dohm:
You're right. I guess my concern is unfounded 😀

abe33:
For more background on the reason of this change, we had this issue in the minimap: https://github.com/atom-minimap/minimap/issues/418

This is an error reported after the conversion to ES6, where some mixins had to change to adapt to ES6 features. Most people reporting the error seems to have made the update from the settings view.

The error message says `this.initializePlugins is not a function` which mean than the new `main.js` file was properly reloaded (https://github.com/atom-minimap/minimap/blob/master/lib/main.js#L75) but the `PluginManagement` was still the coffeescript version (which didn't had the `initializePlugins` method: https://github.com/atom-minimap/minimap/commit/bd7583ddabb10c02cef72184755a6cc18c4ed4d9).
 
Which means we are in a situation where some files are properly reloaded after a package update and some aren't. I guess only the main module file is reloaded, but I couldn't find where this was done in Atom. As far as I can tell the `Package` class is just using `require` when requiring the main module (https://github.com/atom/atom/blob/master/src/package.coffee#L437) so the require's cache should play nonetheless. 

Is there anything else I'm missing here?

maxbrunsfeld:
@abe33 I think the reason that we didn't make an effort to do this in the past is that the package's dependencies may have been upgraded, and they may contain native modules, which *cannot* be loaded twice. At least on node 4.x, if you `delete` a `.node` file from `require.cache`, and then try to require the file again, it throws an exception `Error: Module did not self-register.`

As far as I can tell, this removes the ability to every robustly reload a new version of a package. When a package is upgraded from within Atom, we should probably force the user to reload their window.

What are your thoughts on this? I imagine that your main concern is stopping the flood of issues on your repositories due to users upgrading them from within Atom. What do you think about the proposal to modify settings view to reload the editor window after a package update?

lee-dohm:
> What do you think about the proposal to modify settings view to reload the editor window after a package update?

I'm :+1: on this if this is the only way to ensure that packages get robustly updated. So long as when I do an "Update All" it doesn't reload my window once for each package :laughing:

abe33:
@maxbrunsfeld Right, I didn't consider native module. Do we have an idea of how many packages use native modules VS package using only JS code?

> What are your thoughts on this? I imagine that your main concern is stopping the flood of issues on your repositories due to users upgrading them from within Atom.

That, and the fact that the more Atom will gets adopted the more we'll see non-technical users use it. And from a user point of view, having random errors on update isn't giving a good signal in term of robustness and reliability.

> What do you think about the proposal to modify settings view to reload the editor window after a package update?

That's good for me, as long as we make it clear to the user than updating will reload the window.

PR #10062, opened 2015-12-15
Emit state update when model's scroll position is changed

maxbrunsfeld:
Previously, manipulating the model's scroll position would not emit a presenter state update.


maxbrunsfeld:
Build failures are caused by some 503's from atom.io.

PR #10078, opened 2015-12-15
Copy empty .lproj dirs to fix `navigator.language` on OS X

bengotow:
(Related: https://github.com/atom/electron/issues/2484)

In Atom 1.3.1, navigator.language is always set to "English" on Mac OS X, even if the user is using another language. This is particularly troublesome because English isn't even an appropriate value for that field. (It should be something like `en-US`). This PR is a copy of a change we made to nylas/N1 to fix this issue. It turns out the empty `.lproj` directories inside Electron must be copied for language detection to work in Chromium.


jeancroy:
also linked to this discution
https://github.com/atom/electron/pull/363

damieng:
Sorry we have rewritten our build system and this PR is now invalid as it targets the old system.

PR #10084, opened 2015-12-16
Serialize language state for untitled buffers

### LINKED ISSUE #9992: Untitled buffers don't save language state
This is a pretty common usage scenario for me as I open a temporary tab in language repos to test out changes that I make:
1. Open a new untitled buffer
2. Set the language to anything other than Plain Text (eg Java)
3. write some text
4. Reload Atom

**Expected**: Language is still Java
**Actual**: Language is reset to Plain Text

Atom 1.4.0-dev-77a5409 **non safe-mode**, Windows 10 x64

/cc @as-cii

as-cii:
Fixes #9992.

Although we're adding a new parameter to the serialized state, this should be a 100% backwards-compatible change. :raised_hands: 

/cc: @atom/feedback


benogle:
Cool

lycon2014:
How to sentranize

Issue #4893, opened 2015-01-07
Make setting Atom as Default Editor for filetypes easier


hebbet:
Currently it is not so easy to assign atom as default editor on windows eg for all .css files

the path to atom.exe is changed with every update.

no exe in C:\Users\hebbet\AppData\Local\atom
![image](https://cloud.githubusercontent.com/assets/1423115/5648585/df87f05a-968f-11e4-87bc-ae2b5b6f9280.png)

Chrome has eg an chrome.exe in:
C:\Users\username\AppData\Local\Google\Chrome SxS\Application which starts chrome with the currect folder.

![image](https://cloud.githubusercontent.com/assets/1423115/5648575/c9bdcde4-968f-11e4-9f63-42fc08e3f0b5.png)

kevinsawicki:
Have you tried using `C:\Users\hebbet\AppData\Local\atom\bin\atom`?

That should be a consistent location that always points to the latest installed release

hebbet:
With atom it doesn't work. Tested with Windows 7 and 8.

With atom.cmd it works but with an pretty annoying cmd Window for a very
short time.

sick51:
I tried to use .lnk file in the start menu instead of the .cmd file and it seems to work. The annoying cmd window is no longer appearing.

ArtskydJ:
Thanks @sick51, it works for me!

samirotiv:
Doesn't work for me - I tried opening a C file in Windows 8 using the Start Menu shortcut, and it showed a Command Prompt window (which itself is alright), but then didn't proceed after that.

ArtskydJ:
@samirotiv The shortcut target should be `C:\Users\[you]\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\GitHub, Inc\atom.exe`

samirotiv:
I think there's something wrong with that since the executable surely
wouldn't be in the Start Menu (and it indeed isn't), and the shortcut in
the Start Menu won't point to an executable in the same folder.

The location I wrote above is the target of the shortcut in the Start Menu.

Regards,
Samir.

dogancelik:
It's easy to set Atom as default editor.

1. Go to Windows Explorer
2. Select a `.txt` file, right click, go to *Open with*, then *Choose default program*
3. In dialog, copy and paste this: `%LOCALAPPDATA%\atom\bin\atom.cmd` and click *OK*

[Here's a GIF](http://cl.ly/image/0H1T1W3a013G)

[...]

Issue #5901, opened 2015-03-08
Make Windows shell integration opt-in via Settings


BrainCrumbz:
Hello all
we're on windows 7 64bits. After installing Atom, basically everywhere I right-click in windows explorer I see a "Open with Atom" item. We tried cleaning those many elements manually from registry, but at the next update of course they're back again.

How can we opt-out from the Winsows shell integration? Possibly at install time and, also, afterwards through some program preferences,

Thanks

paulcbetts:
We don't need to do this at install time, we just need to have a setting that will enable / disable this

BrainCrumbz:
Well, if you think about many program install wizards, they can show in one of their steps the option for (Windows?) shell integration, usually in the form of a pre-selected checkbox.

Just for completeness:  
Depending on the type of application, the wizard can have two install paths (default and custom): in such cases, the shell integration options shows up only in custom path. In default path integration is implicitly enabled.

Our 2 cents

paulcbetts:
So, I'm suggesting a solution where anyone can change it whenever they want, and you're telling me, "No, no, make them choose at exactly one time, a time where they'll have no idea what to pick because they won't know the consequences of their decisions or how they'll feel about it later".

BrainCrumbz:
Not at all, I was not clear enough. As said at the beginning, it would be nice to have both ways: at install time, and later on via preferences. The last comment ("*Well, if you think about many...*") was just a digression specific to the install-time feature, it was not meant to exlcude the preferences feature.

art-ist:
+1: I just uninstalled Atom because it took at least 10-20 secs! to open any context-menu in Windows explorer since I installes Atom. That was unacceptable as I'm hevily using it.
After I now uninstalled Atom the context-menu pops up again.

Zireael07:
I am not having any issues with context menus, but I would like an option to opt out, too.

nguyenkinh:
I uninstalled and the menu is still there, only with a broken icon!!!

MarkusRehberg:
Same here: I installed Atom and directly uninstalled it again (which is sad), because I found it extremely invasive to have this "integration" (on Win 8.1. 64 bit). Please allow to easily install atom without such surprises. Thanks!

BramVanroy:
I'd like to get more attention to this issue. Also on Windows 10 I experienced this.

+1 for the option during install *and* in settings

kulukimak:
[...]

PR #10085, opened 2015-12-16
Add Atom to Open With context menu list in Windows Explorer

raelyard:
This modifies the registry to include Atom in the context menu for the same file extensions for which it appears in the Open With context menu list in Finder on OS X.
#4893


Stanzilla:
:+1:

src/browser/squirrel-update.coffee
```diff
@@ -21,6 +21,9 @@ fileKeyPath = 'HKCU\\Software\\Classes\\*\\shell\\Atom'
...
...s', 'lisp', 'cl', 'l', 'lsp', 'mud', 'el', 'log', 'logo', 'lua', 'markdown', 'mdown', 'markdn', 'md', 'mk', 'wiki', 'wikipedia', 'mediawiki', 's', 'mips', 'spim', 'asm', 'm3', 'cm3', 'moinmoin', 'm', 'mm', 'ml', 'mli', 'mll', 'mly', 'mustache', 'hbs', 'pas', 'p', 'patch', 'pl', 'pod', 'perl', 'pm', 'php', 'php3', 'php4', 'php5', 'ps', 'eps', 'dict', 'plist', 'scriptSuite', 'scriptTerminology', 'py', 'rpy', 'cpy', 'python', 'r', 's', 'rl', 'ragel', 'rem', 'remind', 'rst', 'rest', 'rhtml', 'erb', 'erbsql', 'rb', 'rbx', 'rjs', 'rxml', 'sass', 'scss', 'scm', 'sch', 'ext', 'sh', 'ss', 'bashrc', 'bash_profile', 'bash_login', 'profile', 'bash_logout', 'slate', 'sql', 'sml', 'strings', 'svg', 'i', 'swg', 'tcl', 'tex', 'sty', 'cls', 'text', 'txt', 'utf8', 'text/plain', 'TEXT', 'sEXT', 'ttro', 'textile', 'toml', 'xhtml', 'xml', 'xsd', 'xib', 'rss', 'tld', 'pt', 'cpt', 'dtml', 'xsl', 'xslt', 'vcf', 'vcard', 'vb', 'yaml', 'yml', 'nfo', 'g', 'vss', 'd', 'e', 'gri', 'inf', 'mel', 'build', 're', 'textmate', 'fxscript', 'lgt', 'cfm', 'cfml', 'dbm', 'dbml', 'dist', 'dot', 'ics', 'ifb', 'dwt', 'g', 'in', 'l', 'm4', 'mp', 'mtml', 'orig', 'pde', 'rej', 'servlet', 's5', 'tmp', 'tpl', 'tt', 'xql', 'yy']
```
kevinsawicki: This array seems pretty long, what do you think about storing this in a `.json` file and just doing:

```coffee
openWithFileExtensions = require './win32-open-file-extensions'
```

raelyard: That's reasonable.  Thinking through the implications of making sure that file is included in the installation package and available in the path to be picked up in running specs.  Would be ideal to have one source that also works with the OS X plist, but might require more effort than it's worth.

resources/win/win32-open-file-extensions.json
```diff
@@ -0,0 +1,227 @@
+[
+"scpt",
```
kevinsawicki: What do you think about indenting these two spaces and also sorting this list alphabetically to make future additions easier?

src/browser/squirrel-update.coffee
```diff
@@ -5,6 +5,7 @@ path = require 'path'
 appFolder = path.resolve(process.execPath, '..')
 rootAtomFolder = path.resolve(appFolder, '..')
 binFolder = path.join(rootAtomFolder, 'bin')
+binFolder = path.join(appFolder, 'resources', 'cli') if not fs.existsSync(binFolder)
```
kevinsawicki: We usually use `unless` instead of `if not`

src/browser/squirrel-update.coffee
```diff
@@ -21,6 +22,9 @@ fileKeyPath = 'HKCU\\Software\\Classes\\*\\shell\\Atom'
 directoryKeyPath = 'HKCU\\Software\\Classes\\directory\\shell\\Atom'
 backgroundKeyPath = 'HKCU\\Software\\Classes\\directory\\background\\shell\\Atom'
 environmentKeyPath = 'HKCU\\Environment'
+openWithKeyPath = 'HKCU\\Software\\Classes\\Atom'
+filetypeKeyPath = (fileExtension) -> "HKCU\\Software\\Classes\\.#{fileExtension}\\OpenWithProgIds"
+openWithFileExtensions = require path.join(binFolder, 'win32-open-file-extensions.json')
```
kevinsawicki: I think it might be better to keep the `win32-open-file-extensions.json` in the `src/` folder somewhere instead of in `resources/` since it is a `.json` file and not a platform specific file type like a `.plist`.

src/browser/squirrel-update.coffee
```diff
@@ -72,9 +76,28 @@ installContextMenu = (callback) ->
         args = ["#{keyPath}\\command", '/ve', '/d', "\"#{process.execPath}\" \"#{arg}\""]
         addToRegistry(args, callback)
 
+  recursivelyRegisterFileTypesForOpenWith = (index, callback) ->
+    if(index >= openWithFileExtensions.length)
```
kevinsawicki: :fire: No need for parens here

src/browser/squirrel-update.coffee
```diff
@@ -119,12 +142,27 @@ getPath = (callback) ->
...
     spawnReg(['delete', keyPath, '/f'], callback)
 
-  deleteFromRegistry fileKeyPath, ->
-    deleteFromRegistry directoryKeyPath, ->
-      deleteFromRegistry(backgroundKeyPath, callback)
+  deleteValueFromRegistry = (keyPath, valueName, callback) ->
+    spawnReg(['delete', keyPath, '/v', valueName, '/f'], callback)
+
+  recursivelyRemoveFileTypesForOpenWith = (index, callback) ->
+    if(index >= openWithFileExtensions.length)
```
kevinsawicki: :fire: no need for parens here

src/browser/squirrel-update.coffee
raelyard: Started moving this, but then thinking about it - it's not a platform specific file type, but it is a file for a specific platform.  There's also a .js file in the win folder.  .sh files as well (which are platform specific, but not Windows types).  Is the intent really to separate file types?  Will move it if desired, but wanted to ask the question.  If so, should it just go in the root of src or create a new subdirectory?

kevinsawicki: > Will move it if desired

Yeah, I do think we should move it to somewhere in `src/`

> If so, should it just go in the root of src or create a new subdirectory?

I think that is a good place to put it :+1:

src/browser/squirrel-update.coffee
spongecattle: :+1: Could help

amytruong:
Did various testing around this. :+1:  @raelyard suggested downloading random selection of file types (especially media) to check the Open With options.

damieng:
I've merged the basic work on registering on https://github.com/atom/atom/pull/10818 so the application **can** be associated without any path/update issues.

However I think we should avoid actively associating with file extensions for now until we can provide opt-out options as people find t he current level of shell integration already disruptive https://github.com/atom/atom/issues/5901

lee-dohm:
Sounds good to me @damieng. Should this be closed then?

damieng:
Yes, let's close this for now and revisit specific pre-selected file types later.

PR #10122, opened 2015-12-18
Naive implementation of non breaking spaces in Show Invisibles.

paradoxxxzero:
Invisible non breaking space breaking code is driving me nuts so this is a really naive implementation of showing non breaking space with show invisible.

screenshot from 2015-12-18 18-02-01

References: #8080 #9483


paradoxxxzero:
Rewrote in #10400

PR #10126, opened 2015-12-19
Print steps time during Grunt tasks and script/bootstrap

as-cii:
After talking to @maxbrunsfeld about how we may speed up our CI times, we agreed it could have been beneficial to have a better understanding about where the build time gets spent.

As a result, this PR introduces two more pieces of information when building Atom:
1. Package/Modules (un)install times.
2. Grunt tasks times. (these include running specs, linting, transpiling, etc.)

screen shot 2015-12-19 at 17 52 22

screen shot 2015-12-19 at 17 52 32

(You can see a more detailed report at https://travis-ci.org/atom/atom/jobs/97848561)

/cc: @maxbrunsfeld @nathansobo @atom/core

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -14,6 +14,8 @@ _ = require 'underscore-plus'
 packageJson = require '../package.json'
 
 module.exports = (grunt) ->
+  require('time-grunt')(grunt)
+
   grunt.loadNpmTasks('grunt-babel')
   grunt.loadNpmTasks('grunt-coffeelint')
   grunt.loadNpmTasks('grunt-lesslint')
--- a/build/package.json
+++ b/build/package.json
@@ -37,6 +37,7 @@
     "runas": "^3.1",
     "tello": "1.0.5",
     "temp": "~0.8.1",
+    "time-grunt": "1.2.2",
     "underscore-plus": "1.x",
     "unzip": "~0.1.9",
     "vm-compatibility-layer": "~0.1.0",
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -5,8 +5,16 @@ var verifyRequirements = require('./utils/verify-requirements');
 var safeExec = require('./utils/child-process-wrapper.js').safeExec;
 var path = require('path');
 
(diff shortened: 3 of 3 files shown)


joshaber:
OMG :heart_eyes:

joshaber:
(BTW, one thing I noticed when looking into this a while back is that we're [deleting all of `~/.atom`](https://github.com/atom/atom/blob/6a1f311f394070acfdd35d1e932981973c9de949/script/cibuild#L108), so all of the caching that we try to do on `~/.atom/.apm`, etc is wasted.)

as-cii:
> (BTW, one thing I noticed when looking into this a while back is that we're deleting all of ~/.atom, so all of the caching that we try to do on ~/.atom/.apm, etc is wasted.)

Thanks for this insight, @joshaber! :sparkles: I am not sure why we nuke that directory, but I guess it may have something to do with modules being incompatible across electron versions, or something. If that's the case, we may delete that folder only e.g. when we bump the electron version (which doesn't happen that often).

I'd be curious to see how Janky behaves with this new information in place. :thought_balloon: 

Another interesting insight is that, on Travis, running package specs takes 6 minutes more than running core ones.

| Core                                                                                                                                       | Packages                                                                                                                                   |
|--------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
| screen shot 2015-12-19 at 17 52 32 | screen shot 2015-12-19 at 18 01 37 |

So, this probably means we can further parallelize specs (across multiple builds), as the lower bound seems to be around ~ 5 min. 

Another thought could be to parallelize (within the same build) linting, transpilation and electron download/compilation, so that we can probably save other ~4 minutes.

Anyways, my sense is that there are many angles from which we could attack build times and these metrics seem to be good companions to do so. :muscle:

lee-dohm:
Awesome :grinning:

as-cii:
I think this is good to go. :tada: Hopefully it will be useful to gather some more information about our build in order to fix part (or all) of its slowness. :zap:

PR #10139, opened 2015-12-21
Prevent committing `blob-store`

sparanoid:
`blob-store` is accidentally committed to my .atom repo recently: https://github.com/sparanoid/.atom/tree/master/blob-store

I think these files should be avoided.


as-cii:
@sparanoid: :cool:, thanks so much for submitting this!

as-cii:
Just noticed the specs are failing for a git-related problem. I think it's a false negative caused by a problem we had on master when this PR was first opened: could you rebase this against master so that a new build spawns and we can :shipit:?

sparanoid:
@as-cii now LGTM :D

as-cii:
Thanks, @sparanoid! :balloon:

PR #10145, opened 2015-12-21
Fix bug in grammar path scoring

winstliu:
This PR fixes a bug in grammar path scoring where a period would be treated the same as a path separator.  So instead, we only treat the _last_ period as the file extension separator.


50Wliu:
For reference, the failing spec is [this line](https://github.com/atom/atom/blob/5d85ea18ae48c9c9fd54931630a99a4d44962a9e/spec/grammars-spec.coffee#L61):
```coffee
expect(atom.grammars.selectGrammar("/hu.git/config").name).toBe "Null Grammar"
```

Upon digging into this I think this is a bug in Atom's implementation of file path checking.  [`getGrammarPathScore`](https://github.com/atom/atom/blob/master/src/grammar-registry.coffee#L55) tries to split the given path into different components, separated by either a `.` or a `/`.  This works fine for most cases, but when a folder name has a period in it like `hu.git/`, and a grammar tries to match the `git` filetype (in this case `git/config`), it will be inadvertently matched.

The fix to this is probably similar to the one for Notifications and the Tree View - ignore all periods until the last one.

src/grammar-registry.coffee
```diff
@@ -64,7 +63,8 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
       fileTypes = fileTypes.concat(customFileTypes)
 
     for fileType, i in fileTypes
-      fileTypeComponents = fileType.toLowerCase().split(PathSplitRegex)
+      fileTypeComponents = fileType.toLowerCase().split("/")
```
50Wliu: This isn't `path.sep` because it's supposed to match the file types listed in the grammar files, all of which use `/` for consistency.

src/grammar-registry.coffee
```diff
@@ -54,9 +53,9 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
 
   getGrammarPathScore: (grammar, filePath) ->
     return -1 unless filePath
-    filePath = filePath.replace(/\\/g, '/') if process.platform is 'win32'
 
-    pathComponents = filePath.toLowerCase().split(PathSplitRegex)
+    pathComponents = filePath.toLowerCase().split(path.sep)
```
50Wliu: After sleeping on this I'm no longer sure that this is a clean solution.  `file:///` handlers use forward-slashes, and so might other remote files.  I'm not 100% sure every file path passed to `getGrammarPathScore` already uses that OS's path separator.

/cc @maxbrunsfeld

maxbrunsfeld: Yeah, I'm not sure if `filePath` is always a file path, or if it's sometimes a URL. It might be better to just stick w/ the current approach of converting all separators to `/`.

damieng:
Hey, is this still worth keeping open?

50Wliu:
Yep.  It's been waiting on review.  I fix the merge conflicts from time to time.

src/grammar-registry.coffee
```diff
@@ -57,15 +55,17 @@ class GrammarRegistry extends FirstMate.GrammarRegistry
     return -1 unless filePath
     filePath = filePath.replace(/\\/g, '/') if process.platform is 'win32'
 
-    pathComponents = filePath.toLowerCase().split(PathSplitRegex)
+    pathComponents = filePath.toLowerCase().split('/')
+    pathComponents = pathComponents.concat(pathComponents.pop().split('.'))
```
maxbrunsfeld: Now that we're not using the regex anymore, maybe we could get rid of the conditional on line 56 that replaces `\` with `/` on windows.

```coffee
pathComponents = filePath.split(path.sep)
pathComponents.push(pathComponents.pop().split('.')...)
```

maxbrunsfeld (changes requested):
Can we add test coverage for this in a way that doesn't require using an old version of `language-git`? Maybe using the `core.customFileTypes` setting?

spec/grammars-spec.coffee
```diff
@@ -49,18 +49,19 @@ describe "the `grammars` global", ->
         Object.defineProperty process, 'platform', value: originalPlatform
 
       it "normalizes back slashes to forward slashes when matching the fileTypes", ->
-        expect(atom.grammars.selectGrammar('something\\.git\\config').scopeName).toBe 'source.git-config'
+        expect(atom.grammars.selectGrammar(path.win32.join('something', '.git', 'config')).scopeName).toBe 'source.git-config'
 
     it "can use the filePath to load the correct grammar based on the grammar's filetype", ->
-      waitsForPromise ->
-        atom.packages.activatePackage('language-git')
```
50Wliu: This was already activated in a beforeEach.

src/grammar-registry.coffee
50Wliu: I just implemented this, but then realized we had a previous discussion about it: https://github.com/atom/atom/pull/10145#discussion-diff-52680158.  Is this something that we're sure won't run into `file:///` or other weird Windows normalization difficulties?

lee-dohm:
Hey @50Wliu, we're going to close this for now. Let us know if you want to resubmit this. If you have any questions, feel free to ping me directly.

Thanks for all your hard work!

Issue #9165, opened 2015-10-15
Fix preview tabs


benogle:
_From @benogle on October 15, 2015 22:46_

There have been a lot of issues created related to preview tabs. It seems like nearly all of these issues could be solved by a couple things:

1. Create all tabs as 'permanent' by default and only open in a preview tab in a couple cases:
  * Single clicking in the tree-view
  * Single clicking in project find results pane
  * And eventually (when open files is suuuper fast) opening preview tabs when arrowing through the fuzzy finder
2. Close the preview tab when it loses focus

The issues

* https://github.com/atom/tabs/issues/170
* https://github.com/atom/tabs/issues/206
* https://github.com/atom/tabs/issues/202
* https://github.com/atom/tabs/issues/196
* https://github.com/atom/tabs/issues/180
* https://github.com/atom/tabs/issues/173
* https://github.com/atom/tabs/issues/170
* https://github.com/atom/tabs/issues/164
* https://github.com/atom/tabs/issues/207
* https://github.com/atom/tabs/issues/163
* https://github.com/atom/fuzzy-finder/issues/108

## The short term approach

Currently preview-tabs is implemented entirely in the tabs package. When a new pane item is opened, the tabs package closes the old pane item associated with the preview tab, and creates a new tab. So atom core (or other packages) has no idea that there are preview tabs. This is kind of crappy as

* There are hacks in the tabs package to handle things like double clicking in the tree view to make a tab permanent.
* There is no way for other packages to open preview tabs, or even know about them without accessing the tabs package
* Along the same lines, a couple folks want an API, and the only way to provide an API would be to add a service or hack to the tabs package.
  * https://github.com/atom/tabs/issues/188
  * https://github.com/atom/tabs/issues/185

That said, we could probably do a quick fix to change the default behavior to open a 'permanent' tab, and hack the other packages to open preview tabs in the exceptions listed above.

## The long term approach

A cleaner and more robust approach would be to bake the notion of a preview pane item into atom core. Packages could then decide to open preview items if they so choose. Then there would be no hacks: tree-view and the find results could open a preview item on single click; tabs could watch for preview items and have them all share one tab element.

@nathansobo @maxbrunsfeld not sure how much effort this longer term approach would be. Can you estimate how long it would take to add a new option to `open()` indicating the notion of a preview pane-item + related events?

cc @atom/feedback

_Copied from original issue: atom/tabs#218_

benogle:
[...]

PR #10171, opened 2015-12-23
(WIP) Move all logic related to preview tabs into core

kuychaco:
Work in progress, just starting a conversation here. 

See corresponding pull requests for tree-view and tabs packages:
- [tree-view #681](https://github.com/atom/tree-view/pull/681)
- [tabs #244](https://github.com/atom/tabs/pull/244)

Still need to write more tests and clean out tabs package to remove unnecessary code. 

Also, considering refactoring to move pending item logic to the Text Editor. This seems to be the better approach and can be found here #10178.

ku-pending-pane-items

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -3,6 +3,7 @@
 Pane = require '../src/pane'
 PaneAxis = require '../src/pane-axis'
 PaneContainer = require '../src/pane-container'
+TextEditor = require '../src/text-editor'
 
 describe "Pane", ->
   [confirm, showSaveDialog, deserializerDisposable] = []
@@ -153,7 +154,7 @@ describe "Pane", ->
       pane.activateItem(pane.itemAtIndex(1))
       expect(observed).toEqual [pane.itemAtIndex(1)]
 
-    describe "when the `pending` option is true", ->
+    fdescribe "when the `pending` option is true", ->
       it "reports the new active item as pending", ->
         item = new Item("C")
         pane.activateItem(item, pending: true)
@@ -191,7 +192,7 @@ describe "Pane", ->
         expect(pane.getActiveItemIndex()).toBe 1
         expect(pane.isActiveItemPending()).toBe false
 
-    describe "when pending item is re-activated", ->
+    fdescribe "when pending item is re-activated", ->
       it "invokes ::onDidConfirmPendingItem() observers if `pending` is not true", ->
         itemC = new Item("C")
         events = []
@@ -215,6 +216,26 @@ describe "Pane", ->
         expect(pane.getActiveItemIndex()).toBe 1
         expect(events).toEqual []
         expect(pane.isActiveItemPending()).toBe true
+
+    fdescribe "when pending item's buffer content is changed", ->
+      it "invokes ::onDidConfirmPendingItem() observers", ->
+        itemC = new Item("C")
+        events = []
+        pane.onDidConfirmPendingItem (event) -> events.push(event)
+        pane.activateItem(itemC, pending: true)
+        console.log 'active item', pane.getActiveItem()
+        console.log '####', pane.getActiveItem() instanceof TextEditor
+        expect(itemC in pane.getItems()).toBe true
+        expect(pane.getActiveItem()).toBe itemC
+        expect(pane.getActiveItemIndex()).toBe 1
+        expect(events).toEqual [{item: itemC, index: 1}]
+        expect(pane.isActiveItemPending()).toBe false
+
+
+    fdescribe "when pending item is saved or destroyed", ->
+      it "invokes ::onDidConfirmPendingItem() observers", ->
+        # handle saved
+        # handle destroyed
 
   describe "::activateNextItem() and ::activatePreviousItem()", ->
     it "sets the active item to the next/previous item, looping around at either end", ->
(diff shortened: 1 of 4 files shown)


spec/workspace-spec.coffee
```diff
@@ -381,6 +381,32 @@ describe "Workspace", ->
...
+
+        waitsForPromise ->
+          workspace.open('a', pending: true).then (o) -> editor = o
+
+        runs ->
+          expect(workspace.getActivePane()).toBe pane
+          expect(pane.items).toEqual [editor]
+          expect(pane.isActiveItemPending()).toBe true
+
+    describe "when the 'pending' option is NOT true", ->
```
lee-dohm: You may want to add an explicit test case for `pending: false` also, just in case.

lee-dohm:
I'd vote against placing the pending item logic in the `TextEditor` class. Some folks might have ideas for ways to use the pending item system that do not involve a `TextEditor`.

steelbrain:
Can I request an API change?
Currently there's no way to check if a pane item (aka text editor) is a preview one other than hacking through internal APIs and checking the tab element's state

It would help us defer linting for preview tabs

spec/pane-spec.coffee
```diff
@@ -153,6 +153,89 @@ describe "Pane", ->
       pane.activateItem(pane.itemAtIndex(1))
       expect(observed).toEqual [pane.itemAtIndex(1)]
 
+    fdescribe "when the `pending` option is true", ->
```
steelbrain: Is `fdescribe` a new helper that I haven't heard of yet? I only know of `describe`

50Wliu: @steelbrain `fdescribe` focuses that describe, just like `fit` focuses individual tests.

steelbrain: @50Wliu Interesting, Thanks :+1:

lee-dohm:
@steelbrain I'm not sure if this is intended to simply be a refactoring or if it is open for changes to functionality also. You probably should post your suggestion as a new Issue regardless, so that it doesn't get lost.

steelbrain:
@lee-dohm just found it requested here https://github.com/atom/atom/issues/9165#issuecomment-165405260 but that issue wasn't dedicated to it so I created https://github.com/atom/atom/issues/10176

lee-dohm:
Awesome! Thanks @steelbrain :grinning:

spec/pane-spec.coffee
```diff
@@ -153,6 +153,89 @@ describe "Pane", ->
...
+        console.log '####', pane.getActiveItem() instanceof TextEditor
+        expect(itemC in pane.getItems()).toBe true
+        expect(pane.getActiveItem()).toBe itemC
+        expect(pane.getActiveItemIndex()).toBe 1
+        expect(events).toEqual [{item: itemC, index: 1}]
+        expect(pane.isActiveItemPending()).toBe false
+
+
+    fdescribe "when pending item is saved or destroyed", ->
+      it "invokes ::onDidConfirmPendingItem() observers", ->
```
nathansobo: Hmm, not sure it makes sense to confirm the pending item if it's destroyed... what's your thinking behind doing it if the item is saved. Seems like to be saved it will have to be modified, which should confirm it instead.

spec/workspace-spec.coffee
nathansobo: In other places we've said `"when the 'pending' option is false (the default)"`.

spec/pane-spec.coffee
lee-dohm: The current behavior confirms it if it is saved, even with no changes.

spec/workspace-spec.coffee
lee-dohm: Yeah, I suppose if the default changes by accident, this test should break.

nathansobo:
> Also, considering refactoring to move pending item logic to the Text Editor.

Curious to hear more about your ideas. Pondering this a bit, it does seem like we could move *more* of the logic into the pane items themselves, but I can't see a way to move *all* of the logic. I may be missing something though. What do you think of this adjustment to our initial approach...

If an item wants to be displayed in "preview mode", it implements the following interface:

* `isPending` Whenever an item is activated, we'll call this method on the *current* active item if it is defined. If `true`, we'll destroy the current active item and replace it with the new one we are adding.
* `onDidChangePending` Any package interested in responding to the change on an item's pending status can use this to subscribe. In this case, the tabs package would subscribe directly to the item.
* `setPending` If an existing item is opened via `Workspace.prototype.open` *without* the pending option, we'll get the existing item and explicitly assign `pending` to false.

A potential issue with this proposal is that it allows a non-pending item to *become* pending if someone assigns `setPending(true)`. An alternative would be to replace `setPending` and `onDidChangePending` to `confirmPending` and `onDidConfirmPending` or something. What I like about expressing pending as a simple boolean property you assign however is that it avoids the awkward language for what to call the transition from pending to not pending. It's just a property that can change. And who knows, maybe someone will find a use for putting items back into a pending state. It shouldn't be that hard to handle it.

kuychaco:
@nathansobo yeah, you basically described my approach. In my implementation there's no way to go from permanent to pending. 

Thanks for you comments everyone! I wrote out an alternative approach (#10178) that I think is cleaner and more straightforward. Curious to know your thoughts.

kuychaco:
Closing this and proceeding with approach in #10178

PR #10178, opened 2015-12-24
Open file in pending state on single click

kuychaco:
Addresses issue #9165 

Single-clicking on a file in tree-view or on a search result from find-in-project opens a 'pending tab'. 

See corresponding pull requests for tree-view and tabs packages:
- [tree-view #682](https://github.com/atom/tree-view/pull/682)
- [tabs #246](https://github.com/atom/tabs/pull/246)
- [find-and-replace #644](https://github.com/atom/find-and-replace/pull/644)

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -18,6 +18,8 @@ describe "Pane", ->
     onDidDestroy: (fn) -> @emitter.on('did-destroy', fn)
     destroy: -> @destroyed = true; @emitter.emit('did-destroy')
     isDestroyed: -> @destroyed
+    isPending: -> @pending
+    pending: false
 
   beforeEach ->
     confirm = spyOn(atom.applicationDelegate, 'confirm')
@@ -152,6 +154,26 @@ describe "Pane", ->
       pane.onDidChangeActiveItem (item) -> observed.push(item)
       pane.activateItem(pane.itemAtIndex(1))
       expect(observed).toEqual [pane.itemAtIndex(1)]
+
+    it "replaces pending items", ->
+      itemC = new Item("C")
+      itemD = new Item("D")
+      itemC.pending = true
+      itemD.pending = true
+
+      expect(itemC.isPending()).toBe true
+      pane.activateItem(itemC)
+      expect(pane.getItems().length).toBe 3
+      expect(pane.getActiveItem()).toBe pane.itemAtIndex(1)
+
+      expect(itemD.isPending()).toBe true
+      pane.activateItem(itemD)
+      expect(pane.getItems().length).toBe 3
+      expect(pane.getActiveItem()).toBe pane.itemAtIndex(1)
+
+      pane.activateItem(pane.itemAtIndex(2))
+      expect(pane.getItems().length).toBe 2
+      expect(pane.getActiveItem()).toBe pane.itemAtIndex(1)
 
   describe "::activateNextItem() and ::activatePreviousItem()", ->
     it "sets the active item to the next/previous item, looping around at either end", ->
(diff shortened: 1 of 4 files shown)


nathansobo:
I like this better, too. Nice work.

Looks like you still need to thread the option through `Workspace.prototype.open` and test `confirmPendingState` and that modifying the buffer also confirms the state.

lee-dohm:
To me "confirm pending state" means that one is confirming that it *is* pending ... rather than making a pending item not-pending or confirmed. I feel like the term "pending" is causing a naming issue. Is there a reason we can't use the terms "temporary" and "permanent"? (I realize that "pending" came along with the original code, but I can't find the original discussion around the terminology.)

I'm totally :+1: on the API though. Well designed.

nathansobo:
Now that all the methods are on the object rather than on the pane, maybe something like `terminatePendingState` works? I avoided it on the initial Pane-oriented API because of concerns about confusion with a pending item being *removed*, but that's not an issue now.

lee-dohm:
`terminatePendingState` is much better, thanks :grinning:

kuychaco:
Thanks for the feedback and great suggestions for improving the nomenclature! I changed the code to use "terminate" rather than "confirm". 

> Looks like you still need to thread the option through Workspace.prototype.open and test confirmPendingState and that modifying the buffer also confirms the state.

No need to make any changes to Workspace::open, it's already set up to handle the `pending` option. What's nice is that with this approach there's no need to touch the workspace.coffee file. 

Added a few tests. The test for modifying the buffer isn't passing because for some reason using `::insertText` does not trigger the `did-change-modified` event that `::terminatePendingState` is listening for. Is there another method I should be using to modify the buffer in the test?

nathansobo:
> is listening for. Is there another method I should be using to modify the buffer in the test?

It's an async event... the clock is mocked in the global environment and you should be able to use `advanceClock` to force time ahead until it fires. I'm not stoked about the way that evolved but it should work for now.

nathansobo:
What happens if I single click, then double click a file in the tree view?

kuychaco:
Doh! Async, of course... I'm curious to know what you're not stoked about @nathansobo... is there a better way you know of? I tried using `waits`/`runs` but it doesn't work:

```javascript
it "should terminate pending state when buffer is changed", ->
      events = []
      editor1.onDidTerminatePendingState (event) -> events.push(event)
      expect(editor1.isPending()).toBe true
      editor1.insertText('I\'ll be back!')
      waits 500
      runs ->
        expect(editor1.isPending()).toBe false
        expect(events).toEqual [editor1]
```

I've seen `waits` being used in [other specs](https://github.com/atom/tabs/blob/85c58d69af8ed9a69ae45dbfbb2d3d22e7634591/spec/tabs-spec.coffee#L231-L236) and from what I can tell it waits a fixed period of time before moving to the next block. Unsure why it's not working here. Anyhow, using `advanceClock` works just fine. Thanks @nathansobo!

kuychaco:
Single-clicking and then double-clicking a file in tree view should work as expected - single click opens the file in a pending state, and a subsequent double-click terminates the pending state. (Note that for the double-click the first click re-opens the file as pending and the second click terminates the pending state)

You can view a test verifying this [here](https://github.com/atom/tree-view/blob/ab6e96ce3c9dd0c007f4688cd7ff5b9d6d78f549/spec/tree-view-spec.coffee#L587-L609)

Also note that single-clicking a file that is already open does not put the file back in pending state.

nathansobo:
> Single-clicking and then double-clicking a file in tree view should work as expected

:+1:

nathansobo:
Last commit is superficial change. Merging. Congrats @kuychaco on landing your first PR on atom/atom!

lee-dohm:
Congratulations! :tada:

steelbrain:
Note to self: Defer linting on preview tabs

steelbrain:
Is there any config to disable preview tabs? Existing config was [removed in `atom/tabs`](https://github.com/atom/tabs/pull/246/files#diff-2e1bb36917ff76ab7308da52bc968501L19)

benogle:
Not yet, though we plan to add one and patch it into 1.6 beta

Issue #4599, opened 2014-12-13
Hide system window title bar


dustContributor:
Hi! Could it be possible to implement system window title bar/borders hiding like Chromium/Chrome has?

If you have used Chromium/Chrome you can notice it lacks the "system window title bar" by default (ie, the bar that says the name of the window and has the minimize, maximize and close buttons), while at the same time providing its own tiny minimize, maximize and close buttons in the top, rightmost part of the window.

Some window managers for Linux let you hide the window title bar, but since Atom lacks its own minimize/maximize/close buttons, it becomes inconvenient. Besides, it would be something nice to have in Windows/OSX too.

IMHO, this feature manages to save (precious!) vertical space, and looks much more pleasant.

EDIT: Made a discussion thread about this https://discuss.atom.io/t/hide-system-window-title-bar/14578

mnquintana:
Ooooh, this would be pretty cool. I don't know if there should be no title bar by default, but having a config setting for it would be nice. I think whether or not this is currently possible to implement would depend on [atom-shell](https://github.com/atom/atom-shell), and not really Atom core. @zcbenz would this be possible with atom-shell?

zcbenz:
@mnquintana Yeah it is possible with [frameless window](https://github.com/atom/atom-shell/blob/master/docs/api/frameless-window.md).

luqmaan:
If you want a borderless window now, just add

```js
transparent: true,
frame: false,
```

to https://github.com/atom/atom/blob/master/src/browser/atom-window.coffee#L26-L31

![image](https://cloud.githubusercontent.com/assets/1275831/6157402/abbc7d1c-b206-11e4-85f8-0ca66ac576dd.png)

dustContributor:
I'm pretty sure I explained thoroughly what I want and it isn't a plain borderless window (much less one that doesn't uses the gained space like you're showing), you should read the OP and the linked discussion.

dustContributor:
Nevermind, I tried it and I can maximize the window just fine, it also removes the menu bar sadly :cry:

mathieucivel:
@luqmaan thanks, I have now retrieved the perfect distraction free mode I used to like on Sublime. 

I guess there is now way to keep the change across versions yet ?

yutzhead:
Also relevant: http://askubuntu.com/questions/75284/remove-titlebar-from-maximized-terminal-window

bramcordie:
[...]

PR #10208, opened 2015-12-29
Hide title bar on OS X 

anaisbetts:
This PR is mostly to get folx excited, because my CSS-fu is far too poor to make it actually work - on OS X, remove the Title Bar to fit in with other Yosemite apps - based on @simurai's design at https://discuss.atom.io/t/hide-system-window-title-bar/14578/8:

In safe mode:

## TODO:
- [x] Set title-bar-style on the window
- [x] Ensure window is draggable in the parts that it should be (window background, top-left corner), but buttons are still clickable
- [x] Solve weird race condition in CSS where padding doesn't apply when tree view is collapsed / recreated
- [x] Solve how to make tabs indented to the right a bit when left panes are collapsed
- [ ] Figure out a way to make tabs draggable
- [ ] Special case OS X 10.7/10.8, where this API isn't supported

Closes https://github.com/atom/atom/issues/4599


joshaber:
:heart_eyes_cat:

simurai:
:yum:

> Solve how to make tabs indented to the right a bit when left panes are collapsed

The following should add extra padding if the left `atom-panel-container` is empty. Kinda a hack to rely on an empty element, but maybe ok to test it out.

```less
.platform-darwin {
  atom-panel-container.left:empty + atom-workspace-axis.vertical .tab-bar {
    padding-left: 100px;
  }
}
```

traffic-lights

simurai:
It seems that you can't drag the tabs. Probably issue https://github.com/atom/electron/issues/3009.

If we don't wanna wait for a fix, we could mimic a title bar that is full width. And maybe add the title to make some use of the space.

```less
.platform-darwin {
  atom-panel.left {
    padding-top: 0;
  }

  atom-workspace {
    padding-top: 36px;
  }
}
```

screen shot 2015-12-29 at 9 34 57 pm

mnquintana:
Refs https://github.com/atom/atom/issues/4599

paulcbetts:
@simurai I bet we can work around this by creating dummy divs that take up the extra space in places we want draggable - basically, the bug revolves around visually overlapping `-webkit-app-region`s and if you can make them not overlap, it works

paulcbetts:
@zcbenz Is there any clever trick that we can do to make tabs draggable in the non-client region?

zcbenz:
> @zcbenz Is there any clever trick that we can do to make tabs draggable in the non-client region?

No, it has been [a long hurting issue](https://github.com/atom/electron/issues/3009) but we still don't know how to solve it.

benogle:
Oh cool. Crazy it's so little code.

simurai:
For now.. how about this:

1. Wait till https://github.com/atom/atom/pull/9274 is merged.
2. Change to `'title-bar-style': 'hidden'`. I think it makes the traffic lights more compact, have less padding.
3. Add a title-bar. It could say the same as it currently does. Directory and current file. Not sure if the "right click to open parent folder" is needed.

title-bar

It's not as compact as having the traffic lights next to the tabs. But on the other hand it gives a larger reserved area to drag around the window vs. the smaller space next to the tabs.

benogle:
Is this an osx only thing? How does it look on other platforms?

paulcbetts:
> Is this an osx only thing? How does it look on other platforms?

Yeah, this only applies on OS X, other platforms will have the normal title bar

paulcbetts:
This will be harder than one would assume from a CSS perspective, a _lot_ of components use `height: 100%` which breaks being able to add any vertical component, we probably have to retool the panes to use Flexbox

lee-dohm:
@paulcbetts Is this something that you're still interested in pursuing? Is it blocked by something?

benogle:
@lee-dohm I think we're going to do what @simurai mentioned above:

> * Wait till #9274 is merged.
> * Change to 'title-bar-style': 'hidden'. I think it makes the traffic lights more compact, have less padding.
> * Add a title-bar. It could say the same as it currently does. Directory and current file. Not sure if the "right click to open parent folder" is needed.

(#9274 is merged now!) Not sure if it will happen in this PR or not though.

paulcbetts:
@benogle I'll have a look soonish

jneubrand:
It's been two months—can we get an update? Can we help somehow? :grin:

samwillis:
It looks to me like the Brave guys have fixed it in their falk of Electron https://github.com/brave/electron/pull/13 and it just needs merging upstream.

simurai:
The dragging issue seems fixed in https://github.com/electron/electron/pull/5557

🔜 🔜 🔜  :heart_eyes_cat:

paulcbetts:
@simurai :sparkling_heart:

FezVrasta:
Is there a way to apply these changes without compiling the whole app from scratch?

paulcbetts:
@FezVrasta Nope

The-Penultimate-Defenestrator:
@simurai Any update? I assume your last comment about the electron fix means we _won't_ have to settle for the look where the buttons are above the tabs, it can all be inline 😃, and we can have the ideal solution proposed in the initial post. Is this correct? I hope so, that would mean there's only one more checkbox to go. 

So, any updates? (this is the feature I want the *most* out of Atom, it would really give it an awesome native feel.)

simurai:
> @The-Penultimate-Defenestrator I assume your last comment about the electron fix means we won't have to settle for the look where the buttons are above the tabs

Yeah, once Atom uses Electron `1.1.1` the tabs _could_ be at the top (without title-bar).

But if, it probably should stay opt-in to not remove existing functionality, see https://github.com/atom/atom/pull/11790#issuecomment-222168228. Maybe like https://github.com/atom/atom/pull/11790#issuecomment-222191019.

The-Penultimate-Defenestrator:
@simurai Wasn't Electron just updated with the 1.8/1.9 beta?

keithamus:
@The-Penultimate-Defenestrator to `0.37.8` according to the blog post http://blog.atom.io/2016/06/06/atom-1-8-and-1-9-beta.html.

The-Penultimate-Defenestrator:
Wow. That's pretty far behind.

50Wliu:
@The-Penultimate-Defenestrator we're stuck on 0.37.8 because 1.0 and above introduce breaking changes that packages will need time to update for.  So Electron 1.0 will _probably_ be included with Atom 2.0.

The-Penultimate-Defenestrator:
Ah. Are changes still being made to the 0.X version because of this, or is
it static?

lee-dohm:
To be clear, we currently have no plans for version 2 of Atom or even a timeline of when a major version bump might be considered.

Esya:
Wouldn't it possible to have that either part of a package, or part of a theme?

kylerphillips:
What's the best way to implement this as a working solution in my version of Atom?

simurai:
@kylerphillips In Atom `1.11` you can switch to a custom title bar #11790. Then you could hide the title bar in your `styles.less` file. The only issue that remains is that you couldn't drag (reorder) the tabs. If that's a dealbreaker you'd have to wait for Electron to be updated first #12300.

@Esya Yeah, at some point there probably will be themes or a package that will allow it, currently still blocked by the dragging issue ☝️.

FezVrasta:
@simurai I see that when you maximize Atom, the custom titlebar get hidden just like the native one. I'm not sure how this behavior works but I think it could made harder to apply your suggestion to @kylerphillips

Do you have any insights on it?

FezVrasta:
This is what I hacked together using the custom stylesheet feature of Atom and the latest Atom compiled from `master`:

image

My fear was founded, in fact in fullscreen this is the result:

image

As you see, the tabs are cut out.  
Luckily we can use the `.fullscreen` class to style Atom in this case:

image


I'm sharing the CSS in case someone wants to apply these edits:
```less

// Hide title
.title-bar .title {
  display: none;
}

// Remove background from titlebar and toolbars
.title-bar,
.tool-bar,
.tool-bar button.tool-bar-btn {
  background-color: transparent;
  border-width: 0;
}

// Shift the tabs up
atom-workspace atom-workspace-axis.horizontal {
  margin-top: -20px;
}

// Keep the side panels aligned
atom-panel-container{
  &.left {
    margin-top: 42px;
  }
  &.right {
    margin-top: 10px;
  }
}

.fullscreen {
  atom-workspace atom-workspace-axis.horizontal {
    margin-top: 0px;
  }
  atom-panel-container{
    &.left {
      margin-top: 0px;
    }
    &.right {
      margin-top: 0px;
    }
  }
}

// Avoid tabs from overlap window controls when sidebar is closed
atom-workspace atom-panel-container.left:empty + atom-workspace-axis.vertical .tab-bar {
  margin-left: 70px;
}
```

I'd like to move the window controls a bit down to vertical align them with the tabs but I can't find how...

simurai:
@FezVrasta Wouldn't it also work to just hide the `.title-bar` (instead of only the title and make the background transparent)?

```
.title-bar {
  display: none;
}
```

You'd still have to adjust the panels, but then the `title-bar`'s [negative margin](https://github.com/atom/atom/blob/master/static/title-bar.less#L8-L10) in fullscreen mode wouldn't affect the rest.

> I'd like to move the window controls a bit down to vertical align them with the tabs but I can't find how...

Once the dragging issue is fixed, we could remove the title bar and change the window controls to use `hidden-inset`, then it adds a bit more padding. The custom title bar is needed till then.

FezVrasta:
Yes I think it makes sense, I was just hacking it to barely make it work :P

simurai:
> About the `hidden-inset`, this would make the draggable area higher, making impossible to drag the tabs

Yes, right. Let's wait till after the dragging issue has been fixed. In the meantime, feel free to hack as much as you like. 🤘

FezVrasta:
I think a good solution would be if Electron provided a way to define draggable areas using a CSS property like it happens with the Google Chrome Packaged Apps

simurai:
You can use `-webkit-app-region: drag;`, see the [docs](http://electron.atom.io/docs/api/frameless-window/#draggable-region). Just in the current Electron version that Atom uses, there is an "invisible layer" on top that can't be changed. Here the [WIP](https://github.com/atom/atom/pull/12300) to follow along.

FezVrasta:
Ah that's great! So once Electron is updated we can just define a custom area and everything is good :) Great!

kylerphillips:
@simurai What's the best way to go about upgrading to Atom `1.11`? Sorry if that's a stupid question. I have the beta installed but I know 1.11 hasn't been released for Beta yet.

FezVrasta:
You have to compile it yourself

rjsteinert:
This design direction reminds me of how hyperterm looks. Hyperterm also uses Atom if anyone wants to look at a reference of how they do it https://github.com/zeit/hyperterm

Might also have some code in there for making it cross platform compatible.

50Wliu:
> Hyperterm also uses Atom

It uses _Electron_, which is the framework Atom is built on.

rjsteinert:
@50Wliu Hehe, ya, I meant Electron. Woops :P

Some screenshots of hyperterm...

screen shot 2016-08-25 at 11 36 22 am


screen shot 2016-08-25 at 11 36 38 am

smlombardi:
@FezVrasta Thanks for sharing, but adding that code to my styles.less did **not** remove the titlebar in  1.11.0-beta0.  The tabs and tree did move up, but I still have the default Mac titlebar.

FezVrasta:
@smlombardi you have to compile the master to use this feature

smlombardi:
Ah, I was confused as the release notes for 1.11 simply say

> An option to show a custom title bar on macOS got added by @brumm ✨

Nothing about having to build my own atom.

FezVrasta:
Ops sorry I didn't noticed it, then you just need to enable it from the settings 🙃

CvX:
@smlombardi To hide macOS titlebar go to "Atom Beta" -> "Config…", and then under "core" add: `useCustomTitleBar: true`.

steverandy:
I really like the look with titlebar hidden. Much cleaner.

screen shot 2016-09-01 at 8 02 57 pm

nicooprat:
Works great! Any way to move these 3 Mac buttons? It's outside Atom view so I guess it's not possible with CSS only... Here's what I mean, I would like to move them a few pixels to bottom/right to make it more consistent:

capture d ecran 2016-09-14 a 12 50 38

It's really pixel perfection but, you know... Useless, so essential :)

acondiff:
Regarding "Figure out a way to make tabs draggable", isn't it possible to hide the titlebar altogether and make your own traffic lights using css like Nylas does? Wouldn't that make tabs draggable again because an invisible titlebar isn't in the way?

50Wliu:
No longer blocked.

simurai:
Ok, this PR currently looks like this:

screen shot 2016-09-21 at 5 04 44 pm

Difference to `master`: Uses `hidden-inset`, which adds more padding to the traffic lights. But also the title-bar is bigger in height.

## Hiding that custom title-bar

Now about entirely **hiding** the title-bar. It's not too hard in the default configuration, but it can get tricky if you consider all the possible options how people might use/customize Atom. Like when hiding the tree-view, the tabs need some padding:

screen shot 2016-09-21 at 4 53 48 pm

But **not** when you add a top panel (like the tool-bar package):

screen shot 2016-09-21 at 4 54 18 pm

Or what should happen if the left panel isn't wide enough, should the tab-bar add the padding?

screen shot 2016-09-21 at 4 50 48 pm

There are probably more edge cases. I can think of the following:

1. Keep the custom title-bar in core, let themes or `style.less` do the hiding.
2. Add some warning to the toggle in the settings. Like: Hiding the title-bar could cause issues when adding top/left panels.
3. Try to solve all those edge cases.. probably only doable with JS and not that easy.
4. Have a minimum width for the `atom-panel-container.left` so that the tabs and top panels don't have to worry about adding padding. But it adds extra space for people that wanna hide the tree-view. See below:

title-bar

And not ideal for panels that are smaller than the min width:

screen shot 2016-09-21 at 5 46 56 pm

Any other ideas?

nicooprat:
I'd vote for solutions #1 + #2, for now at least, and provide here or wherever some fixes for `styles.less`?

simurai:
For `#1` there is now a PR: https://github.com/atom/atom/pull/12767

Maybe good thing to test it out in a package/theme first.

simurai:
Superseded by #13616

abduljamac:
screen shot 2017-08-05 at 01 52 16
screen shot 2017-08-05 at 01 52 05

simurai:
@abduljamac Easiest way is to open the Atom `Settings > Core` and then at the bottom there is a "Title Bar" option with a drop down:

image

non-binary:
@simurai bless you. thanks!

PR #10226, opened 2015-12-30
Update deprecated-packages.json

danielbayley:
Deprecate `language-jxa` package.

### DIFF
--- a/build/deprecated-packages.json
+++ b/build/deprecated-packages.json
@@ -862,6 +862,10 @@
     "hasDeprecations": true,
     "latestHasDeprecations": false
   },
+  "language-jxa": {
+    "hasDeprecations": true,
+    "latestHasDeprecations": true
+  },
   "language-rspec": {
     "version": "<=0.2.1",
     "hasDeprecations": true,


lee-dohm:
This file was for packages that still had deprecated API calls just prior to the release of v1.0. It wasn't intended to be a general package deprecation system. 

You can unpublish your package if you don't want to support it anymore.

After the holidays, I'll check in with the person who created this system and see what other alternatives there are.

danielbayley:
> You can unpublish your package if you don't want to support it anymore.

I would, but unfortunately I made the mistake of deleting the repository on GitHub before unpublishing it (See [#221](https://github.com/atom/apm/issues/221))…

danielbayley:
Hmm… Wouldn’t it have been better to just remove [`language-jxa`](https://atom.io/packages/language-jxa) (and [`test`](https://atom.io/packages/test) for that matter…) instead, rather than _adding_ code (this PR)?

It feels like there needs to be some kind of “garbage collection” for abandoned packages… Iv’e come across a number of them over time.

See also [#221](https://github.com/atom/apm/issues/221).

PR #10233, opened 2015-12-30
Fix New File with no open windows

### LINKED ISSUE #7171: New file command does nothing with no windows open

joshaber:
Fixes #7171 and #7428.

It'd fail previously with:

```
TypeError: Cannot read property 'pathToOpen' of undefined
```

/cc @atom/feedback


joshaber:
(Supersedes #7933.)

lee-dohm:
:ship: Looks good to me!

PR #10236, opened 2015-12-30
Cascade new windows on OS X

### LINKED ISSUE #10057: New windows should be offset in mac
Repro in Mac:
1. Drag a folder to Atom.
2. Drag another folder to Atom:

Observed: Second window opens on top of the first.
Expected: Second window is offset down and to the right by ~20px, and is smaller in each dimension by the same amount.

joshaber:
Fixes #10057.

I don’t know if this is The Right Thing® on other platforms, so for now it’s OS X-only.

/cc @atom/feedback


50Wliu:
Also a thing on Windows:
cascading-windows

No idea how many pixels that's offset though.

joshaber:
Awesome, thanks @50Wliu!

FWIW, I also changed the behavior to base the new window's dimensions on the last opened window. That seemed to more closely match the behavior on OS X, but I don't feel strongly either way.

joshaber:
>FWIW, I also changed the behavior to base the new window's dimensions on the last opened window. 

I changed my mind on this in the name of not fixing what ain't broke.

mnquintana:
Awesome! :smile:  Should we add tests for this?

joshaber:
It doesn't look like we're testing `atom-application` currently. Mostly it's UI glue code.

benogle:
Would it be worth writing a [integration test](https://github.com/atom/atom/tree/master/spec/integration) for it? Not sure if this is in its purview.

joshaber:
Oh interesting! I'll give that a look.

joshaber:
Managed to pull together an integration test :metal:

maxbrunsfeld:
> Managed to pull together an integration test :metal:

Nice!

50Wliu:
After testing this out on Windows it looks like there needs to be some extra Windows-specific code: opening a new window that spawns from a maximized one should stay maximized and not cascade.

PR #10247, opened 2016-01-01
Bump git-utils

### LINKED ISSUE #3627: Editor crashes on saving or on opening a file that has been modified
In the last 2 days, Atom has started to:

* Crash immediately after saving an edited file
* Crash whenever you try to open a file that has been modified but not committed (according to git) - verified by committing the file and then trying to open it, which then works

It appears to have issues reconciling the git status of the file on opening or saving, maybe? I'm not using any complex arrangement of plugins.

joshaber:
Pick up https://github.com/atom/git-utils/pull/61.

_Should_ fix #3627.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "fs-plus": "^2.8.0",
     "fstream": "0.1.24",
     "fuzzaldrin": "^2.1",
-    "git-utils": "^4.0.7",
+    "git-utils": "^4.0.8",
     "grim": "1.5.0",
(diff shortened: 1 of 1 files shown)


joshaber:
:boat:

viatsko:
Shouldn't this be merged in stable and beta branches also as bug affects them too?

joshaber:
We generally only target stable or beta when fixing a regression. This'll end up in those branches as part of the usual rolling of the railcars.

PR #10256, opened 2016-01-02
Remove "Open Your" from menu items

### LINKED ISSUE #9952: Hide all "Open Your ..." menu items behind a single one
Could we please combine all of those items behind a single menu item called `Open Your ...`, `Open Your` or `Modifications`? If you ask me, that would be a lot better than repeating "Open Your" over and over again.

The new item should then contain "Config", "Init Script", "Keymap", "Snippets" and "Stylesheet".

menu

leo:
This resolves #9952 (please read the issue for more infos). Here’s a quick screenshots of what this PR does:
before-after


mnquintana:
Thanks for contributing! :tada: 

I'm personally :-1: on this change because it dramatically reduces usability in favor of a purely aesthetic improvement.

**Before:** Clicking on an Open Your item requires 2 actions: 
1. Open the Atom (or Edit on other platforms) menu
1. Click on an Open Your Item

**After:** Clicking on an Open Your item requires 4 actions:
1. Open the Atom (or Edit on other platforms) menu
1. Move cursor to Customization
1. Move cursor into expanded menu
1. Click on an Open Your item

I do agree that this looks cleaner, and makes sense from an organizational point of view, but I'm not sure that it's a worthwhile tradeoff for the hit in usability.

What do you think @atom/feedback?

leo:
Thank you, too, for your feedback, @mnquintana! :blush:

I actually had this in mind while I was making those changes. I partly agree with your point of view. While it definitely reduces usability, I definitely don't think it's a dramatic change. However, I think it's a tradeoff that needs to be made to organise the menu items logically and aesthetically.

If you ask me, one of the most important things I've learned while creating software is, that you should use whatever options are available to you to avoid repeating yourself. Not only in the code, but also within the interface. I definitely think that it's important to make certain actions as accessible as possible, but in this case, I think we should keep in mind that we need to treat every pixel of our software that appears on a user's screen as valuable as possible.

Therefore I think simply repeating "Open Your ..." over and over again is just disrespectful towards the fact that a user is trusting us with a certain space on his computer.

Following this reasoning, we could of course just show all of those five items without the prefix and they would be easily accessible AND halfway aesthetic. However I also think that it's important to group things together that have a collective meaning when designing UIs. Not only does it save space, but it also makes the interface clearly structured and easier to understand.

lee-dohm:
I'm :-1: on having a submenu for this. These are important items that we want users to know about. Leaving aside usability for a moment, simply putting things on a submenu makes it that much more likely that these commands will not be found.

In [About Face](http://www.amazon.com/About-Face-Essentials-Interaction-Design/dp/0470084111), the author talks about how software design should consider the needs of three different expertise levels in users:

> Our goal should be neither to pander to beginners nor to rush intermediates into expertise. Our goal is threefold: to rapidly and painlessly get beginners into intermediacy, to avoid putting obstacles in the way of those intermediates who want to become experts, and most of all, to keep perpetual intermediates happy as they stay firmly in the middle of the skill spectrum.

I see having these not in a submenu as specifically part of the "rapidly and painlessly get beginners into intermediacy" mission. The menu system is there for beginners and somewhat for intermediates. The menu system is not there for experts. Experts and most intermediates will use the command palette or keybindings for the things they need.

Because of this, I feel that the need to optimize the menu system for beginners to use and easily discover what capabilities Atom has outweighs the desire to make the menu system more compact and perhaps esthetically pleasing to intermediates and power users.

simurai:
> we could of course just show all of those five items without the prefix and they would be easily accessible AND halfway aesthetic. However I also think that it's important to group things together that have a collective meaning when designing UIs.

How about keeping them on the first level but give them its own group (then "Preferences..." has its own group too). And remove the `Open your`. It think it makes it easier to scan.

no-label

If it's not clear enough, we could add an extra "read-only" line at the top, so it doesn't need to be repeated. It could say `Customize your...`

label

But maybe not needed. If you click any of the items.. you can assume that it opens "something".

mnquintana:
> @simurai: How about keeping them on the first level but give them its own group (then "Preferences..." has its own group too). And remove the `Open your`. It think it makes it easier to scan.
>
> no-label

I like this :+1: 

Maybe we should add ellipses to those items, to better reflect that they open a new "window" that would require some input from the user. From the [Apple HIG](https://github.com/atom):
> Use an ellipsis in the name of a button or menu item when the associated action... [i]s performed by the user in a separate window or dialog.
>
> For example, Preferences, Customize Toolbar, and App Store all use an ellipsis because they open a window or dialog in which the user sets preferences, customizes the toolbar, or shops for new apps.
To see why such commands must include an ellipsis, consider that the absence of an ellipsis implies that the app performs the action for the user. For example, if the Customize Toolbar command does not include an ellipsis, it implies that there is only one way to customize the toolbar and the user has no choice in the matter.

simurai:
:+1:

no-label

I guess the `About Atom` and `View License` should also have ellipses.

leo:
:blush: That sounds awesome! Now I only need to know if `About Atom` and `View License` need ellipses, too and if the `Customize your...` is necessary, too. Then I can adjust the PR.

If you ask me, they need ellipses, too. But the read-only line isn't necessary. Also: What about the ellipses on win32 and linux?

lycon2014:
Marshmellow

simurai:
My vote:

- if "About Atom" and "View License" need ellipses? Yes
- if the `Customize your...` is necessary. No
- What about the ellipses on win32 and linux? Looking at some screenshots, they seem to follow the same convention. So ellipses on all 3 platforms.

mnquintana:
> if "About Atom" and "View License" need ellipses? Yes

My vote's for no ellipses on those two, if I'm interpreting Apple's HIG correctly –

> Don't use an ellipsis in the name of a button or menu item when the associated action... [d]oes not require specific input from the user.
>
>For example, the New, Save a Version, and Duplicate commands don't use an ellipsis because either the user has already provided the necessary information or no user input is required. That is, New always opens a new document or window, Save a Version saves a snapshot of the current document, and Duplicate creates a new copy of the current document.

The View License command never requires further input from the user, and neither does About Atom. Also, I don't think I've ever seen an OS X app add ellipses after About <app name>.

lee-dohm:
I agree with @mnquintana on the placement of ellipses. And I agree with @simurai that we should use the same system across platforms.

simurai:
> My vote's for no ellipses on those two, if I'm interpreting Apple's HIG correctly –

I think you're right. :+1: No ellipses for "read only" stuff. Ok, the consensus seems to be this:

no-label

For all 3 platforms.

leo:
Done!

leo:
@simurai @lee-dohm What else needs to be done to get this merged?

kandros:
I'm :-1:  for both, the submenu and the removal of **your**
To me the **your** does make explicit the cascading nature of the overriding happening.

Without **your** it look like there are no init operations, nor keymaps nor snippets or stylesheets at all already implemented.

lee-dohm:
Thanks @leo for contributing and your patience while we get things through the pipeline :grinning:

Thanks for the feedback @kandros. I don't believe that it is a big risk that people will think there are no keymaps, snippets or styles already implemented. But we'll keep an eye on it!

leo:
@lee-dohm :blush:

PR #10277, opened 2016-01-04
Fix measurement of rows that are present in the model but haven't been rendered yet

### LINKED ISSUE #10051: pixelPositionForScreenPosition randomly returns invalid x position
As you can see in the gif below, some of my markers views get invalid x coordinates when I modify the buffer:

screen-position-issue

In the gif above, the update of the views is scheduled on a new frame during the `did-change` event of the text editor. I have more or less the same result if I do the update immediately. And as far as I can tell it was still working properly last week.

nathansobo:
Fixes #10051

This PR moves clipping of measured positions up to the editor and out of `LinesYardstick`. This allows us to check for the presence of a row in the editor after clipping is applied and saves complexity since `LinesYardstick` is only used internally and we can always control whether we clip the position we pass in.

Instead of just checking whether a row is visible in the model, we also check whether a DOM node with the line's assigned id is present. If It isn't, we force a synchronous render.

### DIFF
--- a/spec/fake-lines-yardstick.coffee
+++ b/spec/fake-lines-yardstick.coffee
@@ -19,9 +19,8 @@ class FakeLinesYardstick
   setScopedCharacterWidth: (scopeNames, character, width) ->
     @getScopedCharacterWidths(scopeNames)[character] = width
 
-  pixelPositionForScreenPosition: (screenPosition, clip=true) ->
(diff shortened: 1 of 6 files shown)


as-cii:
Thanks for taking care of this! :pray: Looks good to me. :+1:

JesseLeite:
Awesome, thanks @nathansobo!

PR #10293, opened 2016-01-05
Ignore errors when loading an invalid blob store

as-cii:
Supersedes #10292.

@nathansobo: this should be good to go now! :boom:


nathansobo:
As for the code, seems reasonable. Looks like Appveyor is failing but once we're green I'd say let's hotfix.

joshaber:
CI won't pass since this is targeting beta :sad: But this is legit and the fix needs to go out.

as-cii:
❤️💜💙💚💛

PR #10296, opened 2016-01-05
This change gives custom repositoryProviders more priority

viatsko:
This will allow us to override base git repository implementation if needed


joshaber:
Makes sense to me :+1: 

/cc @atom/feedback in case there's anything I'm not thinking of.

lee-dohm:
Looks good to me :+1: Especially since this is the way directory providers are written.

joshaber:
:boom:

Issue #7757, opened 2015-07-07
ATOM erases PATH Variable when installing on Win7


SpectreJan:
When I installed ATOM 1.0.0 and one of the previous versions on Windows7 using the Windows installer it erased the user PATH Variable instead of appending it to the existing PATH variable, which is kind of annoying. Sorry I forgot which was the version I installed before, but that behaviour is still present in 1.0.0.
Not sure if this behaviour is intended, I sincerely hope not!

Cheers,
Jan

SpectreJan:
With appending it to the PATH variable I meant appending it's own bin directory to PATH. Sorry for the bad phrasing.

Cheers,
Jan

50Wliu:
Refs #5092
/cc @kevinsawicki

batjko:
It sure isn't intended behavior, but it didn't happen for me on Win7 or Win8.
Have you tried completely uninstalling Atom and re-installing it from scratch, see if it does it again?

SpectreJan:
I did that,
It still shows the same behaviour. I also had a look at the Issue 50Wliu linked (thanks), and it seems that it is the same problem.

thomasjo:
@SpectreJan So running `reg query HKCU\Environment /v Path` fails just like it does for others in the referenced issue (#5092)?

SpectreJan:
@thomasjo Yes it fails.

thomasjo:
@SpectreJan Can you show us the error message? I'm guessing it's in German? If so, then that's the problem since @kevinsawicki's quick fix only works if the error message is in English unfortunately.

SpectreJan:
@thomasjo  No problem
 `C:\Users\krae_jn>reg query HKCU\Environment /v Path`
`FEHLER: Die Registrierungsbearbeitung wurde vom Administrator deaktiviert.`

thomasjo:
As I expected then. @kevinsawicki looks like the old fix needs to be revisited (somehow). Perhaps we can always fiddle with `Path` on Windows in a safe way?

batjko:
What both the German and English error messages have in common is the words 'error' and 'administrator'.
That may be true for a bunch of other languages, though surely not for all.
But it could be a quick adjustment to the existing fix.

Stanzilla:
Sounds like Atom tried to edit the system PATH variable instead of the user PATH variable?

benogle:
@damieng is this fixed?

damieng:
[...]

Issue #5092, opened 2015-01-16
Atom Windows Installer Deletes User Path


robatron:
The latest Atom Windows installer completely overwrites the user's `PATH` environment variable with `c:\Users\<user>\AppData\Local\atom\bin`.

robatron:
CC @jskjons

simonjosefsson:
New install or auto-update?

Didn't happen for me with auto-updated 0.174.0 on win 8.1 64bit

jskjons:
I had the same issue with a fresh install of 0.170.0 on win7x64

50Wliu:
@paulcbetts would this be an issue with Squirrel?

robatron:
@simonjosefsson, a new install on Windows 7 x64. First time using the installer. (Had previously been using Chocolatey.)

kevinsawicki:
>  would this be an issue with Squirrel?

Most likely not, Atom manages the path

kevinsawicki:
For people having this issue can you run the following command and include the output in a comment?

```
reg query HKCU\\Environment /v Path
```

robatron:
@kevinsawicki , from an admin `cmd.exe`:

    C:\>reg query HKCU\Environment /v Path

    HKEY_CURRENT_USER\Environment
        Path    REG_SZ    ;C:\Program Files\Boot2Docker for Windows

kevinsawicki:
@robatron what about from a non-admin `cmd.exe`?

robatron:
@kevinsawicki , unfortunately, I don't have perms. (Work machine.)

    c:\>reg query HKCU\Environment /v Path
    ERROR: Registry editing has been disabled by your administrator.

kevinsawicki:
@robatron so did you run the Atom installer as an administrator?

jskjons:
from admin: C:\tools\atlassian-plugin-sdk\bin;C:\tools\boot2docker
non-admin: (had to do powershell for the same reason as @robatron) 
get-itemproperty HKCU:\Environment
PATH         : C:\tools\devops-toolkit;c:\Users\jskjons\AppData\Local\atom\bin;C:\tools\apache-maven-3.2.1\bin;C:\tools\c
               hromedriver_win32-2.11\bin;C:\tools\git-1.9.3\bin;C:\tools\groovy-2.2.2\bin;C:\tools\httpd-rei.com-2.2.22.
[...]

Issue #5063, opened 2015-01-14
Atom installation/updating garbles %PATH%


renecrapts:
The character *é* in my user %PATH% environment variable gets garbled after installing Atom. It seems to happen after an auto-update as well.

Before installation:
```
C:\Users\René\AppData\Roaming\npm
```
After installation (actual):
```
C:\Users\Ren�\AppData\Roaming\npm;c:\Users\René\AppData\Local\atom\bin
```
Expected:
```
C:\Users\René\AppData\Roaming\npm;c:\Users\René\AppData\Local\atom\bin
```
Windows 7 SP1
Atom 0.172.0

kevinsawicki:
I've added a check so it will no longer corrupt the path, we still need to find a better long term solution

PR #10326, opened 2016-01-07
Fix Windows installer path update woes

raelyard:
#5092 and #7757 were worked around with not updating the path if registry tools access disabled in the Windows Group Policy
#5063 was worked around with not updating the path if the path contains characters outside the ASCII range

This addresses both of these with fixes that correctly update the user path under these conditions with the deleting or corruption of the user path resulting from these issues.

### DIFF
--- a/src/browser/squirrel-update.coffee
+++ b/src/browser/squirrel-update.coffee
@@ -11,9 +11,11 @@ exeName = path.basename(process.execPath)
 if process.env.SystemRoot
   system32Path = path.join(process.env.SystemRoot, 'System32')
   regPath = path.join(system32Path, 'reg.exe')
+  powershellPath = path.join(system32Path, 'WindowsPowerShell', 'v1.0', 'powershell.exe')
   setxPath = path.join(system32Path, 'setx.exe')
 else
   regPath = 'reg.exe'
+  powershellPath = 'powershell.exe'
   setxPath = 'setx.exe'
 
 # Registry keys used for context menu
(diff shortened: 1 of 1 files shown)


src/browser/squirrel-update.coffee
```diff
@@ -43,11 +45,26 @@ spawn = (command, args, callback) ->
...
 
 # Spawn reg.exe and callback when it completes
 spawnReg = (args, callback) ->
   spawn(regPath, args, callback)
 
+# Spawn powershell.exe and callback when it completes
+spawnPowershell = (args, callback) ->
+  # set encoding and execute the command, capture the output, and return it via .NET's console in order to have consistent UTF-8 encoding
+  # http://stackoverflow.com/questions/22349139/utf-8-output-from-powershell
+  args[0] = "[Console]::OutputEncoding=[System.Text.Encoding]::UTF8\r\n$output=#{args[0]}\r\n[Console]::WriteLine($output)"
```
kevinsawicki: Do you think it might look better to split this over multiple lines for readability like:

```coffee
args[0]  = [
  '[Console]::OutputEncoding=[System.Text.Encoding]::UTF8'
  "$output=#{args[0]}"
  '[Console]::WriteLine($output)'
].join('\r\n')
```

mnquintana: @kevinsawicki Or maybe a block string?

```coffee
args[0]  = """
  '[Console]::OutputEncoding=[System.Text.Encoding]::UTF8'
  "$output=#{args[0]}"
  '[Console]::WriteLine($output)'
"""
```

Not sure if CoffeeScript interprets that to be Unix line endings though.

kevinsawicki: @mnquintana yeah, I didn't think CoffeeScript could force `\r\n` with `"""` strings, not 100% sure though

raelyard: After some testing, turns out Powershell (on Windows 10 (Powershell 5)) doesn't care about the Windows line endings.

Both of these work:
args[0]  = [
  '[Console]::OutputEncoding=[System.Text.Encoding]::UTF8'
  "$output=#{args[0]}"
  '[Console]::WriteLine($output)'
].join('\r\n')

args[0]  = [
  '[Console]::OutputEncoding=[System.Text.Encoding]::UTF8'
  "$output=#{args[0]}"
  '[Console]::WriteLine($output)'
].join('\n')

Net result, the block string also works:
args[0]  = """
  [Console]::OutputEncoding=[System.Text.Encoding]::UTF8
  $output=#{args[0]}
  [Console]::WriteLine($output)
"""

Definitely a readability improvement.
Not committing this until tested on older versions of Powershell, but working in initial run.

iolsen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;SUSvJm07&#x2F;4-fix-windows-installer-path-update-woes"> Fix Windows installer path update woes</a>

damieng:
Tested this out today:

- [x] Merges with master and builds
- [x] Creates Windows installer

On Windows 7 32-bit with registry tools **enabled**:
- [x] Installer correctly adds to existing path with a unicode glyphs
- [x] Uninstaller correctly removes from existing path containing unicode glyphs 

On Windows 7 32-bit with registry tools **disabled**:
- [x] Installer correctly adds to existing path with a unicode glyphs
- [x] Uninstaller correctly removes from existing path containing unicode glyphs

PR #10327, opened 2016-01-07
Async submodule support

joshaber:
Well, I split the submodule work out into its own PR hoping that'd make it more easily reviewable, but then I went an added a new fixture with submodules so this PR is back to being unreviewable :cry:


joshaber:
That test failure is a race that's over on `dh-async-repo` too. I'll fix it over there.

PR #10328, opened 2016-01-07
Default to using standard GitHub API token during build

lee-dohm:
Resolves an issue where the GitHub API is rate-limited during the build when attempting to download Electron.

/cc @kevinsawicki


kevinsawicki:
:+1: :key:

PR #10332, opened 2016-01-08
Relocate and rename 'View > Reload' to prevent confusion

### LINKED ISSUE #10331: Relocate or rename "Reload" menu item to prevent confusion
See: https://github.com/atom/text-buffer/issues/124

lee-dohm:
Fixes #10331

/cc @joshaber @mnquintana

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -163,8 +163,8 @@
         label: 'Developer'
         submenu: [
           { label: 'Open In Dev Mode…', command: 'application:open-dev' }
-          { label: 'Run Package Specs', command: 'window:run-package-specs' }
           { label: 'Reload Window', command: 'window:reload' }
+          { label: 'Run Package Specs', command: 'window:run-package-specs' }
           { label: 'Toggle Developer Tools', command: 'window:toggle-dev-tools' }
         ]
       }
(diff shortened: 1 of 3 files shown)


menus/darwin.cson
```diff
@@ -165,6 +164,7 @@
         submenu: [
           { label: 'Open In Dev Mode…', command: 'application:open-dev' }
           { label: 'Run Package Specs', command: 'window:run-package-specs' }
+          { label: 'Reload Window', command: 'window:reload' }
```
mnquintana: Just an idea – I wonder which of these is the most commonly used command for package developers? If it's Reload Window, maybe it would be worth putting it first? Otherwise, maybe alphabetical order would be best.

lee-dohm: My theory on this was essentially:

* Open in Dev Mode might be targeted by people as being the first in the list
* Toggle Dev Tools might be targeted by people as being the last in the list (and it is targeted that way by me)
* Run Package Specs feels more related to Open in Dev Mode to me

But yeah, now that you mention it ... alphabetic order makes a little more sense.

mnquintana:
Looks good to me! :rocket:

PR #10336, opened 2016-01-08
Backport identical tab names fix

winstliu:
Fixes Atom hanging indefinitely upon encountering two identical filenames located in different directories in 1.3.x.

(Is cherry-picking the correct way to do this?  It feels really weird creating a PR that uses someone else's commits)


maxbrunsfeld:
Thanks for this @50Wliu!

maxbrunsfeld:
And yeah, once we've already merged it into master, we have no choice but to :cherries: pick. It creates some duplicate entries in the history, but it's the approach we're going with for now.

Issue #10302, opened 2016-01-06
Menu ordering on OS X doesn't match system patterns


danielcompton:
On OS X, most apps have a similar pattern of menus:

![screenshot of safari 7-01-16 12-02-28 am](https://cloud.githubusercontent.com/assets/811954/12141142/f326b0e0-b4d1-11e5-869c-400c40fbaf46.png)

![screenshot of iterm2 7-01-16 12-02-41 am](https://cloud.githubusercontent.com/assets/811954/12141149/fb8e1494-b4d1-11e5-83eb-1474d43081f8.png)

![screenshot of finder 7-01-16 12-02-52 am](https://cloud.githubusercontent.com/assets/811954/12141151/030087ac-b4d2-11e5-89cd-33c7d0b7250b.png)

![screenshot of mail 7-01-16 12-03-29 am](https://cloud.githubusercontent.com/assets/811954/12141163/1ad9f2aa-b4d2-11e5-8a14-2edb081d7863.png)

![screenshot of textedit 7-01-16 12-03-58 am](https://cloud.githubusercontent.com/assets/811954/12141171/2e8bee16-b4d2-11e5-9f5f-e06cf8f07584.png)

The general pattern is: File, Edit, (Format), View, Application menus, Window, Help. However Atom uses File, Edit, Selection, Find, View, Packages, Window, Help. 

![screenshot of atom 7-01-16 12-04-22 am](https://cloud.githubusercontent.com/assets/811954/12141230/a8dec562-b4d2-11e5-8282-814b3d12474a.png)

I think View should probably be next to File and Edit. 

Apologies if this is a duplicate, I tried searching but I couldn't find anything.

simurai:
:+1:  on `Atom   File   Edit   View   Selection   Find   Packages   Window   Help`

lee-dohm:
This is also backed up by the [OS X Human Interface Guidelines](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/MenuBarMenus.html#//apple_ref/doc/uid/20000957-CH29-SW1).

danielcompton:
Perfect, I was looking for those guidelines but couldn't find them. Thanks!

PR #10341, opened 2016-01-08
Use OS X Human Interface Guidelines compliant menu order

mnquintana:
Fixes https://github.com/atom/atom/issues/10302

Might be worth checking to see if we're compliant on other platforms too, and also checking the order of items in the File, Edit, View, Window, and Help menus.


joshaber:
Awesome, thanks!

PR #10349, opened 2016-01-09
Fix F11 keymaps

### LINKED ISSUE #10287: Listening to window.toggle-full-screen using atom.commands.onWillDispatch does not retain originalEvent
*Atom 1.3.2, Ubuntu 12.04*

Run the following in the console:

```js
atom.commands.onWillDispatch(function (e) {console.log(e.originalEvent)})
```

When using the default <kbd>F11</kbd> to trigger `window.toggle-full-screen` the `originalEvent` property is not set like with other keybindings.

Output:

```
(left)
KeyboardEvent {}

(home)
KeyboardEvent {}

(F11)
undefined
```

Tyriar:
Function key keymaps don't work when they feature upperface 'F' characters.
They still tend to work because they typically have an application menu item
which is associated with it and falled back upon. The main difference is that
the originalEvent (KeyboardEvent) is not retained when doing this fallback as
Atom thinks that the menu item was pressed instead of the keymap.

Fixes #10287

### DIFF
--- a/keymaps/linux.cson
+++ b/keymaps/linux.cson
@@ -13,7 +13,7 @@
   'ctrl-alt-o': 'application:add-project-folder'
   'ctrl-shift-pageup': 'pane:move-item-left'
   'ctrl-shift-pagedown': 'pane:move-item-right'
-  'F11': 'window:toggle-full-screen'
+  'f11': 'window:toggle-full-screen'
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:+1: Thanks for figuring this out @Tyriar.

PR #10352, opened 2016-01-09
Support symlinked repo roots when relativizing

thedaniel:
With this change, async tree-view specs are green (https://github.com/atom/tree-view/pull/624) but it seems this blows up sync tree-view specs somehow. In any case, we need to use relativizeToWorkingDirectory in this class so that symlinks can be relativized.

cc @joshaber


joshaber:
:sunglasses: Is it crazy to try to test that?

thedaniel:
> Is it crazy to try to test that?

`::relativizeToWorkingDirectory`? not crazy, though i intended it as an internal utility function and figured any errors would bubble up to tests for functions that use it. I'll throw in a couple tests though, more coverage is better than less.

joshaber:
Yeah, specifically that it Does The Right Thing with symlinks?

thedaniel:
Well, I'm glad you poked me on testing that, since I found a problem with the implementation doing so. Tree-view is green on this, and master has a few flaky specs related to NativeCompileCache that I suspect aren't related. I'll merge #9213 into this next and see how that goes.

src/git-repository-async.js
```diff
@@ -161,7 +161,11 @@ export default class GitRepositoryAsync {
   // Returns a {Promise} which resolves to the relative {String} path.
   relativizeToWorkingDirectory (_path) {
     return this.getRepo()
-      .then(repo => this.relativize(_path, repo.workdir()))
+      .then((repo) => {
+        let workingDirectory = repo.workdir()
```
joshaber: It looks like this change ended up being unnecessary?

src/git-repository-async.js
```diff
@@ -171,23 +175,21 @@ export default class GitRepositoryAsync {
   //
   // Returns the relative {String} path.
   relativize (_path, workingDirectory) {
-    // Cargo-culted from git-utils. The original implementation also handles
```
joshaber: :fire: :package: :pray:

src/git-repository-async.js
thedaniel: ah yes, good call, i didn't fully clean up from another change

thedaniel:
I merged in #9213 and ... one deserialization spec went :red_circle: :person_frowning: Just one though!

thedaniel:
I am throwing in the towel on these flaky buffer event specs. @joshaber maybe tomorrow we can look at them together?

joshaber:
Yeah definitely :metal:

joshaber:
https://github.com/atom/atom/commit/f30f2c0d629e83ccf173e5f43296979efe658bbb seems to have fixed the race, though it's troubling we don't really know why :\

Issue #10358, opened 2016-01-10
PackageManager.disablePackage() does not check if a package is already disabled


burodepeper:
See https://github.com/burodepeper/language-markdown/issues/70

I assumed `atom.packages.disablePackage()` would do this for me, but it didn't, and now I've polluted a whole bunch of people's configs with duplicates... :oopsie:

mnquintana:
Confirmed on Atom 1.5.0-dev-6d949d3

burodepeper:
Thanks for picking this up so quickly!
Is there an easy way to remove all the duplicate entries (although I assume they're harmless) without enabling a package and then disabling it again...

PR #10361, opened 2016-01-10
Only disable a package if it isn't already disabled

mnquintana:
Fixes https://github.com/atom/atom/issues/10358

I wasn't sure if this check was better suited for `Package` or `PackageManager`, but I think it's more within the scope of `PackageManager` to manage the state of the `core.disabledPackages` keypath. Let me know if you think I should move it though!

/cc @atom/feedback

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -1026,6 +1026,16 @@ describe "PackageManager", ->
         expect(atom.packages.enablePackage("this-doesnt-exist")).toBeNull()
         expect(console.warn.callCount).toBe 1
 
+      fit "does not disable an already disabled package", ->
+        packageName = 'package-with-main'
+        atom.config.pushAtKeyPath('core.disabledPackages', packageName)
+        atom.packages.observeDisabledPackages()
+        expect(atom.config.get('core.disabledPackages')).toContain packageName
+
+        atom.packages.disablePackage(packageName)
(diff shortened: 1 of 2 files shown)


Fank:
@mnquintana FYI: checks have failed

mnquintana:
Hmmm these look like unrelated failures – the same specs fail on https://travis-ci.org/atom/atom/jobs/101451615 (https://github.com/atom/atom/pull/10341), and that one's just a simple app menu change that should have no effect on the areas that are failing.

Looking into it more, it looks like `master`'s also failing on those same specs (https://travis-ci.org/atom/atom/jobs/101166457), since https://github.com/atom/atom/pull/10178 was merged.

cc @kuychaco @nathansobo

nathansobo:
Not getting local failures with either tree-view or settings-view. Kicking off an internal build for comparison.

nathansobo:
Okay, the problem was that jQuery moved from 2.1.4 to 2.2.0 a couple days ago and there were incompatible changes. I've pinned Atom and space-pen back to 2.1.4. Hopefully we can eliminate the jQuery dependency before it becomes issue that we're staying behind.

mnquintana:
:eyes: @atom/feedback?

50Wliu:
Looks like a simple enough change, so :+1:

joshaber:
This looks good, thanks!

PR #10362, opened 2016-01-10
Only build pushes to master on Appveyor

mnquintana:
To mirror Travis config, refs https://github.com/atom/atom/pull/8110

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,9 @@
 version: "{build}"
 
+branches:
+  only:
+    - master
+
 skip_tags: true
 clone_folder: c:\projects\atom
 clone_depth: 10


mnquintana:
Hmmm, weird that this didn't kick off an Appveyor build – does [that config](https://www.appveyor.com/docs/appveyor-yml) not mean what I think it means? :disappointed:

lee-dohm:
The way that reads to me is that it only builds `master`, since this isn't on `master` yet ... it won't kick off an Appveyor build?

mnquintana:
That's true for branches, but I was expecting Appveyor to still build PRs with this config like Travis does. After looking into it more, it seems that we don't have the setting for building PRs enabled on Appveyor itself (see http://help.appveyor.com/discussions/questions/203-auto-run-tests-on-pull-requests).

benogle:
Yo @mnquintana, whats the status of this?

mnquintana:
So I think someone with admin access on Appveyor needs to flip the switch for allowing builds on PRs

joshaber:
@mnquintana I updated that setting :+1: Wanna make a dummy push to make sure this works?

mnquintana:
@joshaber Looks good, thanks!

PR #10363, opened 2016-01-10
:arrow_up: language-go@0.42.0

### LINKED ISSUE #10272: editor hangs for very long time with 7.5k file
I've found similar reports when using Atom with large files but my file is less than 300 lines with less than 150 characters each.

**Reproduction Steps:**

1. Download and open the file https://github.com/go-gcfg/gcfg/blob/083575c3955c85df16fe9590cceab64d03f5eb6e/set.go#L242
2. Go to the end of line 242 (btw I find it weird that in Windows the last character is not displayed in full so the cursor at the end is also not visible)
3. Press Enter

**Expected behavior:**

New line is started at L243

**Observed behavior:**

Atom editor window hangs for a very long time. (I thought it was indefinite but it came back after an hour or so.) After that, new line is started.

**Screenshots and GIFs**

screenshot

**Atom version:** 1.3.1 and 1.4.0beta3
**OS and version:** Win8.1 64bit, Debian8.2 64bit

**Installed packages:**

Just the core packages.

**Additional information:**

* Problem can be reproduced in safe mode: Yes
* Problem started happening recently, didn't happen in an older version of Atom: don't know
* Problem can be reliably reproduced, doesn't happen randomly: Yes
* Problem happens with all files and projects, not only some files or projects: No

svanharmelen:
This version fixes #10272

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "language-css": "0.36.0",
     "language-gfm": "0.82.0",
     "language-git": "0.11.0",
-    "language-go": "0.41.0",
+    "language-go": "0.42.0",
     "language-html": "0.43.1",
     "language-hyperlink": "0.16.0",
     "language-java": "0.17.0",


50Wliu:
@svanharmelen would you mind targeting stable instead for this?  I think the freeze is dangerous enough to warrant a hotfix.

svanharmelen:
Sure... I'll close this one and create a new PR.

PR #10367, opened 2016-01-11
Fixed the indentation logic

chfritz:
Previously the logic lacked handling of multiple matches of the
increase/decrease regexs, so things like the following javascript
would not be handled correctly (the last line would be indented).

``` js
if (test) {
  foo({a:1
  })};
var t = "this should be at indentation 0";
```

Fixed that.

Also revisited the logic regarding decreaseIndent and
decreaseNextIndent. It did not seem very logical that decreaseIndent
did _not_ correspond to increaseIndent: decreaseIndent regarded the
current line whereas increaseIndent regarded the next line (just like
decreaseNextIndent). For that reason I renamed decreaseIndent to
decreaseThisIndent and decreaseNextIndent to decreaseIndent. This way
we also avoid a whole lot of issues related to the concept of
decreaseNextIndent, because most of the language packages I looked at
have not yet adopted it but the semantics of decreaseIndent had
changed. So now the semantics of decreaseIndent is back to what it was
and decreaseThisIndent is the new concept.

Also now using the tokenized line to determine whether a in-/decrease
pattern appears within a quoted string (and ignoring it in that case),
rather than requiring regexes for checking whether or not the
occurrence is within a quoted string or not -- the examples I saw for
that were wrong, too (e.g., things like `/} [^"] /` are non-sense,
because things like `} var x = "hey"` are perfectly fine). One really
needs to reason about balanced quotation to get it right, and the
tokenizer already does that for us, so let's use it rather than
reinventing the wheel. The decrease/increase regexes can now be much
simpler in the language packages.

### DIFF
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -252,6 +252,9 @@ class LanguageMode
     decreaseThisIndentRegex = @decreaseThisIndentRegexForScopeDescriptor(scopeDescriptor)
     decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
 
+    if not decreaseThisIndentRegex
+      decreaseThisIndentRegex = decreaseIndentRegex
+
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
       return 0 unless precedingRow?
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -243,13 +243,14 @@ class LanguageMode
     @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, line, tokenizedLine, options)
 
   suggestedIndentForTokenizedLineAtBufferRow: (bufferRow, line, tokenizedLine, options) ->
+    # console.log tokenizedLine
     iterator = tokenizedLine.getTokenIterator()
     iterator.next()
     scopeDescriptor = new ScopeDescriptor(scopes: iterator.getScopes())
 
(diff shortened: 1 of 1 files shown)


chfritz:
This new logic is best tried with the following PR for javascript: https://github.com/atom/language-javascript/pull/301.

alexandernst:
I don't usually post `+1` answers. But, please oh please, +1 for this.

src/language-mode.coffee
```diff
@@ -298,15 +316,30 @@ class LanguageMode
...
+
+  # Count the number of occurence of regex in string, given its tokenized
+  # version, and ignoring any occurence that start inside a quoted string
+  countRegex: (regex, string, tokenized) ->
+    count = 0
+    m = regex.searchSync(string)
+    while (m)
+      quoted = false
+      if scopes = tokenized.tokenAtBufferColumn(m[0].start)?.scopes
+        quoted = scope for scope in scopes when scope.match '^string.quoted'
```
50Wliu: Just saying, YAML also has `string.unquoted`, so you might want to adjust for that.

chfritz:
Found a way to make the transition smoother. Will create a new PR.

chfritz:
The new PR is https://github.com/atom/atom/pull/10384#issuecomment-170925427.

src/language-mode.coffee
chfritz: Good point. I didn't know about that. However it seems that unquoted strings must not contain any special characters, so I'm not sure I can think of a language and example where this would be necessary. Can you think of a good test case (incl. language) we could add for that?

PR #10375, opened 2016-01-11
:fire: CHANGELOG.md

joshaber:
Having this file still here is confusing. Do we use it, or do we use https://github.com/benogle/pr-changelog?

Let's apply :fire: if we're not gonna keep it up to date.

/cc @mnquintana @atom/core


nathansobo:
Failed experiment.

Issue #5692, opened 2015-02-23
Restart and Update Doesn't from Release Notes Page


joshvickerson:
Using the Restart and Update button on the release notes view doesn't properly restart and update. Atom closes, but it never restarts. When launched again, it hasn't updated. Restart and update from the Atom Menu works fine.

2010 iMac, OSX 10.10.2

mnquintana:
Which version of Atom?

joshvickerson:
I've noticed this on the past few releases and have just gotten around to submitting the issue.

In the most recent case I was trying to update from .181 to .182

judahstevenson:
I'm experiencing this problem as well, but only when I open atom from the command line, and even then it only happens on certain repos. If i open atom using spotlight it doesn't happen. I'm on version .188 but the notes page show's the latest release as .186

Clicking on the "restart and update" button closes Atom, but then it never re-opens. If I open from the command line again after that it opens on the release notes page as before.

cemsbr:
It is happening with me from 191 to 192

judahstevenson:
hallelujah I was finally able to fix the behavior in my installation of atom. I stumbled across this forum discussion about cache in atom: 

https://discuss.atom.io/t/clean-cache-when-atom-starts/13196/5

I closed atom, deleted all the content in ```` ~/.atom/storage ````

restarted atom via spotlight, then opened the repo that I had been experiencing the problem with. No files opened up by default, including the release notes view. 

When I posted my original comment I was on version 181/182 but I have kept current with releases, even though the problem persisted. I'm on 192 right now. 

It might be worth noting that prior to trying this solution I had also tried just nuking the entire repo and re-cloning but it had no effect.

mark-hahn:
> Using the Restart and Update button on the release notes view doesn't properly restart and update. 

I don't see any such button in my Atom release notes page.  Where is it supposed to be?  I'm on atom 190 on Ubuntu 14.0.4.

BTW, I haven't gotten an auto-update in months when on Windows.  I'm now on Linux.  Does it work differently than on Windows?

I just tried clearing storage.  I'll see if I get an update.  If I don't I'll file an issue.  I've been putting that off for a long time.

50Wliu:
Linux doesn't have auto-updates yet.

cemsbr:
@joshvickerson , it also worked for me. Before, I noticed that there is no Restart and Update button if atom is launched in a folder where it hasn't been launched before. Probably because there is no entry for that folder in the storage files.

[...]

PR #10055, opened 2015-12-14
[WIP] Fix wrong version on release notes

ysakasin:
The bug is mixed in that commit : 9e59ab1e45ed4cbb00a6ed0408ce76572d5ab32b

ref:
[atom/src/browser/auto-update-manager.coffee](https://github.com/atom/atom/blob/c90aabed923c49d2b0d5e9d601d806de41b98c6f/src/browser/auto-update-manager.coffee#L65)
f989ed65e1ffd684b338867d964c5de62e83eab3
atom/release-notes#39
#9682

PR #10377, opened 2016-01-11
Remove release-notes

joshaber:
Since it’s been reporting the wrong release notes, it’s doing more harm than good :disappointed: 

We’ll link to the release notes from About now.

/cc @atom/feedback


joshaber:
I'll back port this onto `stable` once it's merged.

package.json
```diff
@@ -102,7 +102,6 @@
     "notifications": "0.62.1",
     "open-on-github": "0.40.0",
     "package-generator": "0.41.0",
-    "release-notes": "0.53.0",
```
kevinsawicki: :fire: :green_heart:

50Wliu:
Refs (and should close) #10055.

mnquintana:
Is this going to change the update flow at all? (Like, will there now be no status bar squirrel indicating a new version's available / been installed, for instance?)

benogle:
> Is this going to change the update flow at all? (Like, will there now be no status bar squirrel indicating a new version's available / been installed, for instance?)

This. I think the status bar squirrel is done in the release notes package, but that logic could move to the about package.

lee-dohm:
Also should close https://github.com/atom/atom/issues/5692

joshaber:
>I think the status bar squirrel is done in the release notes package, but that logic could move to the about package.

Or perhaps the `status-bar` package itself?

lee-dohm:
> Or perhaps the status-bar package itself?

:+1: I like this idea.

simurai:
There is also https://github.com/atom/about/pull/5, so the about package would fit too in case that direction will move forward.

joshaber:
Ah, good point @simurai. I'll go with `about` then :+1:

joshaber:
Once this goes green I'll merge and cherry-pick it into stable.

benogle:
Cool :+1: 

> cherry-pick it into stable

beta?

joshaber:
>>cherry-pick it into stable

>beta?

Yeah or that. Basically the same since they'll both be rolled into 1.4.0.

joshaber:
Looks like the about specs are hanging on janky for some reason.

joshaber:
Well, it passed on janky-osx4 and Travis so it seems to be a problem specific to janky-osx2.

joshaber:
Cherry picked into `beta`: https://github.com/atom/atom/commit/d229fed7e73c0f89631ed3b6c2cf16497d0ccbab

PR #10384, opened 2016-01-12
Fixed the indentation logic (updated)

chfritz:
Previously the logic lacked handling of multiple matches of the
increase/decrease regexs, so things like the following javascript
would not be handled correctly (the last line would be indented).

``` js
if (test) {
  foo({a:1
  })};
var t = "this should be at indentation 0";
```

Fixed that.

Also revisited the logic regarding decreaseIndent and
decreaseNextIndent. It did not seem very logical that decreaseIndent
did _not_ correspond to increaseIndent: decreaseIndent regarded the
current line whereas increaseIndent regarded the next line (just like
decreaseNextIndent). For that reason I renamed decreaseIndent to
decreaseThisIndent and decreaseNextIndent to decreaseIndent. This way
we also avoid a whole lot of issues related to the concept of
decreaseNextIndent, because most of the language packages I looked at
have not yet adopted it but the semantics of decreaseIndent had
changed. So now the semantics of decreaseIndent is back to what it was
and decreaseThisIndent is the new concept.

Also now using the tokenized line to determine whether a in-/decrease
pattern appears within a quoted string (and ignoring it in that case),
rather than requiring regexes for checking whether or not the
occurrence is within a quoted string or not -- the examples I saw for
that were wrong, too (e.g., things like `/} [^"] /` are non-sense,
because things like `} var x = "hey"` are perfectly fine). One really
needs to reason about balanced quotation to get it right, and the
tokenizer already does that for us, so let's use it rather than
reinventing the wheel. The decrease/increase regexes can now be much
simpler in the language packages.

In this version we use decreaseIndentPattern as the default value for decreaseThisIndentPattern, which is OK, because that pattern is never applied (matched) more than once, which was the previous behavior.

### DIFF
--- a/spec/fixtures/sample.js
+++ b/spec/fixtures/sample.js
@@ -11,3 +11,19 @@ var quicksort = function () {
 
   return sort(Array.apply(this, arguments));
 };
+
+foo({
+    atIndentTwo: true
+  },
+  atIndentOne
+);
+atIndentZero = true;
+
+bar({
+    atIndentTwo: true
+  });
+atIndentZero = true;
+
+bar({
+    atIndentTwo: true });
+atIndentZero = true;
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -249,11 +249,11 @@ class LanguageMode
     scopeDescriptor = new ScopeDescriptor(scopes: iterator.getScopes())
 
     increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    decreaseThisIndentRegex = @decreaseThisIndentRegexForScopeDescriptor(scopeDescriptor)
     decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
 
-    if not decreaseThisIndentRegex
-      decreaseThisIndentRegex = decreaseIndentRegex
+    if not decreaseNextIndentRegex
+      decreaseNextIndentRegex = decreaseIndentRegex
 
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
@@ -276,18 +276,18 @@ class LanguageMode
       # desiredIndentLevel += 1 if increaseIndentRegex?.testSync(precedingLine)
       if increaseIndentRegex
         desiredIndentLevel += @countRegex(increaseIndentRegex, precedingLine, tokenizedPrecedingLine)
-      # desiredIndentLevel -= 1 if decreaseIndentRegex?.testSync(precedingLine)
-      if decreaseIndentRegex
-        desiredIndentLevel -= @countRegex(decreaseIndentRegex, precedingLine, tokenizedPrecedingLine)
+      # desiredIndentLevel -= 1 if decreaseNextIndentRegex?.testSync(precedingLine)
+      if decreaseNextIndentRegex
+        desiredIndentLevel -= @countRegex(decreaseNextIndentRegex, precedingLine, tokenizedPrecedingLine)
       # we need to add back one if we already closed one scope on the previous line itself
-      if decreaseThisIndentRegex and @countRegex(decreaseThisIndentRegex, precedingLine, tokenizedPrecedingLine)
+      if decreaseIndentRegex and @countRegex(decreaseIndentRegex, precedingLine, tokenizedPrecedingLine)
         desiredIndentLevel += 1
 
     unless @buffer.isRowBlank(precedingRow)
-      # desiredIndentLevel -= 1 if decreaseThisIndentRegex?.testSync(line)
-      if decreaseThisIndentRegex and @countRegex(decreaseThisIndentRegex, line, tokenizedLine)
+      # desiredIndentLevel -= 1 if decreaseIndentRegex?.testSync(line)
+      if decreaseIndentRegex and @countRegex(decreaseIndentRegex, line, tokenizedLine)
         desiredIndentLevel -= 1
-      # desiredIndentLevel -= @countRegex(decreaseThisIndentRegex, line) if decreaseThisIndentRegex
+      # desiredIndentLevel -= @countRegex(decreaseIndentRegex, line) if decreaseIndentRegex
 
     # console.log("suggestedIndentForTokenizedLineAtBufferRow: rtv", desiredIndentLevel)
     Math.max(desiredIndentLevel, 0)
@@ -339,10 +339,10 @@ class LanguageMode
   # bufferRow - The row {Number}
   autoDecreaseIndentForBufferRow: (bufferRow) ->
     scopeDescriptor = @editor.scopeDescriptorForBufferPosition([bufferRow, 0])
-    return unless decreaseThisIndentRegex = @decreaseThisIndentRegexForScopeDescriptor(scopeDescriptor)
+    return unless decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
 
     line = @buffer.lineForRow(bufferRow)
-    return unless decreaseThisIndentRegex.testSync(line)
+    return unless decreaseIndentRegex.testSync(line)
 
     currentIndentLevel = @editor.indentationForBufferRow(bufferRow)
     return if currentIndentLevel is 0
@@ -356,9 +356,9 @@ class LanguageMode
     if increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
       desiredIndentLevel -= 1 unless increaseIndentRegex.testSync(precedingLine)
 
-    if decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
-      desiredIndentLevel -= 1 if decreaseIndentRegex.testSync(precedingLine)
-      # desiredIndentLevel -= @countRegex(decreaseIndentRegex, precedingLine)
+    if decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
+      desiredIndentLevel -= 1 if decreaseNextIndentRegex.testSync(precedingLine)
+      # desiredIndentLevel -= @countRegex(decreaseNextIndentRegex, precedingLine)
 
     # console.log("autoDecreaseThisIndentForBufferRow: desiredIndentLevel", desiredIndentLevel)
     if desiredIndentLevel >= 0 and desiredIndentLevel < currentIndentLevel
@@ -376,13 +376,13 @@ class LanguageMode
   # when the *current* line matches this pattern than decrease indentation of
   # the current line (this, e.g., puts closing brackets with nothing before them
   # at the level of the opening statement, as in the KR style for C)
-  decreaseThisIndentRegexForScopeDescriptor: (scopeDescriptor) ->
-    @getRegexForProperty(scopeDescriptor, 'editor.decreaseThisIndentPattern')
+  decreaseIndentRegexForScopeDescriptor: (scopeDescriptor) ->
+    @getRegexForProperty(scopeDescriptor, 'editor.decreaseIndentPattern')
 
   # when the previous line matches this pattern than decrease indentation of the
   # current line
-  decreaseIndentRegexForScopeDescriptor: (scopeDescriptor) ->
-    @getRegexForProperty(scopeDescriptor, 'editor.decreaseIndentPattern')
+  decreaseNextIndentRegexForScopeDescriptor: (scopeDescriptor) ->
+    @getRegexForProperty(scopeDescriptor, 'editor.decreaseNextIndentPattern')
 
   foldEndRegexForScopeDescriptor: (scopeDescriptor) ->
     @getRegexForProperty(scopeDescriptor, 'editor.foldEndPattern')
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -326,11 +326,11 @@ class LanguageMode
     count = 0
     m = regex.searchSync(string)
     while (m)
-      quoted = false
+      isQuoted = false
       if scopes = tokenized.tokenAtBufferColumn(m[0].start)?.scopes
-        quoted = scope for scope in scopes when scope.match '^string.quoted'
+        isQuoted = scopes.some (scope) -> scope.startsWith('string.quoted')
       # console.log quoted
-      count += 1 unless quoted
+      count += 1 unless isQuoted
       m = regex.searchSync(string, m[0].start + 1)
     count
 
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -252,6 +252,9 @@ class LanguageMode
     decreaseThisIndentRegex = @decreaseThisIndentRegexForScopeDescriptor(scopeDescriptor)
     decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
 
+    if not decreaseThisIndentRegex
+      decreaseThisIndentRegex = decreaseIndentRegex
+
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
       return 0 unless precedingRow?
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -243,13 +243,14 @@ class LanguageMode
     @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, line, tokenizedLine, options)
 
   suggestedIndentForTokenizedLineAtBufferRow: (bufferRow, line, tokenizedLine, options) ->
+    # console.log tokenizedLine
     iterator = tokenizedLine.getTokenIterator()
     iterator.next()
     scopeDescriptor = new ScopeDescriptor(scopes: iterator.getScopes())
 
     increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
+    decreaseThisIndentRegex = @decreaseThisIndentRegexForScopeDescriptor(scopeDescriptor)
     decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
 
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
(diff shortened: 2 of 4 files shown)


50Wliu:
Looks like some tests also need to be updated for this.

chfritz:
I don't really think so. This is simply a cyclic dependency with the language packages being used in the tests (in particular the javascript one). In which order do you suggest trying to resolve them? The language-package maintainers may point out the same thing (failing test cases on my PR), which are in turn due to the not-yet-merged PR in atom/atom.

chfritz:
Also created a corresponding PR for language-c:
https://github.com/atom/language-c/pull/110

src/language-mode.coffee
```diff
@@ -298,15 +319,30 @@ class LanguageMode
...
+
+  # Count the number of occurence of regex in string, given its tokenized
+  # version, and ignoring any occurence that start inside a quoted string
+  countRegex: (regex, string, tokenized) ->
+    count = 0
+    m = regex.searchSync(string)
+    while (m)
+      quoted = false
+      if scopes = tokenized.tokenAtBufferColumn(m[0].start)?.scopes
+        quoted = scope for scope in scopes when scope.match '^string.quoted'
```
maxbrunsfeld: It would be good to break this loop once we find a match, and also to avoid performing a `RegExp` match here. What do you think about this:

```coffee
isQuoted = scopes.some (scope) -> scope.startsWith('string.quoted')
```

chfritz: Thanks for the suggestion of using `startsWith`, that is indeed what I was thinking.
However, regarding the loop, no, we cannot break it because the next occurrence may be outside of the quote again, e.g.,

```javascript
foo({
     a: 1
   /* an oddly placed comment */  });
var thisShouldBeAtIndent0;
```

maxbrunsfeld: Yeah, I meant breaking your `for` loop over the scopes (i.e. using `some` instead of a full loop), not the outer `while` loop. No reason to continue looping over scopes once a quoted scope is found, right?

maxbrunsfeld:
Thanks for digging into this! It's a pretty tricky area of the code.

> For that reason I renamed `decreaseIndent` to `decreaseThisIndent` and `decreaseNextIndent` to  `decreaseIndent`.

I think we're going to need to keep the existing names, in order to avoid breaking other people's language packages.

> I don't really think so.

The tests do need to be updated. We would at least will need to add one or more specs that test the new capabilities that you've added (e.g. dedenting multiple times if multiple `}` tokens appear on the same line).

> the semantics of `decreaseIndent` had changed

I'm not sure what you mean by this; how have the semantics of `decreaseIndent` changed? I don't think that they were changed at all when the `decreaseNextIndent` regexp. was introduced.

chfritz:
Hi Max,

Thanks for attending to this. I saw the dialog you had with @svanharmelen regarding his contributions last summer, so I'm glad we are all on the same page regarding status-quo.

Responding to your points:

1) renaming: The renaming is incremental. Only two language packages currently set the `decreaseNextIndentPattern` parameter, go and javascript:

```shell
[cfritz@astatine:~/github/atom/node_modules Wed Jan 13 13:39:35]
> grep decreaseNext language-*/settings/*.cson
language-go/settings/language-go.cson:    'decreaseNextIndentPattern': '^\\s*[^\\s()}]+(?<m>[^()]*\\((?:\\g<m>|[^()]*)\\)[^()]*)*[^()]*\\)[,]?$'
```

(javascript doesn't show up because I've already patched it).
Regarding the renaming of `decreaseIndent` and `decreaseThisIndent`: the new code uses the former as the value for the latter if the latter is not defined. This matches the understanding of the language-pack authors at the time they wrote their packages.

2) Regarding test: again, there are cyclic dependencies with the language packs (in particular javascript, c, and go), which are used in the tests. If the PRs of those got merged then the atom/atom tests would pass. So I'll look to you guys to tell me how to best break this cycle (start with the language packs, atom/atom, or somehow changes the tests temporarily).

3) changed semantics of `decreaseIndent`: Yeah, I think you might be right. After looking at more language packs it seems that the regexes were indeed meant for outdenting the current line, which just isn't such a great approach because things like 

```javascript
if (test) {
  var x = 1; }
var shouldBeAtZero;
```

would become:

```javascript
if (test) {
var x = 1; }
var shouldBeAtZero;
```

or (worse), be completely wrong, like:

```javascript
if (test) {
  var x = 1; }
  var shouldBeAtZero;
```

The latter is what I have seen in several languages, and which is why my first interpretation of `decreaseIndent` was actually what `decreaseNextIndent` is meant to do. Perhaps some of the language pack authors had that same misconception of its semantics.

BTW, please let me know where to document the intended semantics of these parameters. No matter what we decide, we should capture that somewhere so language pack authors can get clarity.

maxbrunsfeld:
Can you explain a little further about the cyclic dependency? I would think that since your rename was done in a backwards-compatible way, the existing language packages would continue to work. I know that you removed the need for the regexps to check for quotation marks, but I would think that could be done in a backwards compatible way as well (it doesn't hurt to check for quotation marks, it just is no longer necessary).

chfritz:
No they won't because the semantics for `increaseIndent` has changed (you can now match more than once), and `decreaseIndent` can also be applied more than once per line. But honestly I don't think the indentation of the language packs is working right now anyways. I think it's quite broken and the tests are just too narrow to show that. I think the number of open issues relating to this show that quite clearly. The regexes for outdenting, e.g., are very brittle. For instance, most of them assume that there is only whitespace before the string/char that indicates the end of a block. For instance in C++ right now you get:

```c++
if (x) {
  cout << ok; }
  
  int thisIsWronglyIndented;
```

This is because `decreaseIndent` is really meant for the case where the *current* line can and has to be outdented already (as with the `}` in KR style). Since that is not the case above, and since there is no `decreaseNextIndent` defined for C, the behavior is wrong.

I think we really need to allow ourselves to take a larger scoop here to out get out of this hole. Frankly, inductive (line-to-line) + regex based indentation logic seems error prone to me. I would think that using the parse tree is a better solution approach, so maybe that is something to reconsider, too. But that would be an even longer discussion and larger change.

chfritz:
The only way I see to make this merge *not* require simultaneous changes in the language packs is if we literally set some sort of flag in the language pack settings indicating to use the new semantics, and then keep the old logic around and use it for any language packs that do not set that flag to true. I think it's a little ugly, but perhaps as a transitional and temporary step that's OK.

Would you prefer that?

chfritz:
@maxbrunsfeld could you let me know how you want us to move forward with this? I don't want this to be forgotten and many people are waiting on a fix to the indentation problems in several languages.

Please let me know which path forward you think is the most promising, both in terms of long-term correctness of the logic and short-term merge-ability:

1. merge the current PRs (this one and the language ones) and create corresponding PRs for all other languages. Pro: getting to the right solution faster. Con: intermediate even more broken indentation logic and failing tests. 

2. change the logic to detect somehow (perhaps using a flag in the settings) whether to use the old or new indentation logic. Pro: no disruptions whatsoever. Con: some ugly code for some time.

3. write a completely new indentation logic, e.g., based on tokens (I'm actually toying with this right now using emacs as the baseline). Con: same as 2. Pro: potentially much better indentation (emacs indentation is pretty fantastic), including for multi-line expressions such as:

```js
var x = someVeryLongVariableNameThatIsLooooooooooooooooooooooooooooooooooooong
  * 134723458236457165328762587365487623548756237847523948572345
  / numberOfElementsInTheUniverse;

// and pipelines:
svg.selectAll("g")
    .data(tiler
      .scale(projection.scale() * 2 * Math.PI)
      .translate(projection([0, 0])))
  .enter().append("g")
    .each(function(d) {
      var g = d3.select(this);
      d3.json("http://" + ["a", "b", "c"][(d[0] * 31 + d[1]) % 3] + ".tile.openstreetmap.us/vectiles-highroad/" + d[2] + "/" + d[0] + "/" + d[1] + ".json", function(error, json) {
        g.selectAll("path")
           .data(json.features.sort(function(a, b) { return a.properties.sort_key - b.properties.sort_key; }))
           .enter().append("path")
           .attr("class", function(d) { return d.properties.kind; })
           .attr("d", path);
      });
    });
```

The latter would currently be indented as:
```js
var x = someVeryLongVariableNameThatIsLooooooooooooooooooooooooooooooooooooong
* 134723458236457165328762587365487623548756237847523948572345
/ numberOfElementsInTheUniverse;

// and pipelines:
svg.selectAll("g")
.data(tiler
   .scale(projection.scale() * 2 * Math.PI)
   .translate(projection([0, 0])))
.enter().append("g")
.each(function(d) {
   var g = d3.select(this);
   d3.json("http://" + ["a", "b", "c"][(d[0] * 31 + d[1]) % 3] + ".tile.openstreetmap.us/vectiles-highroad/" + d[2] + "/" + d[0] + "/" + d[1] + ".json", function(error, json) {
     g.selectAll("path")
     .data(json.features.sort(function(a, b) { return a.properties.sort_key - b.properties.sort_key; }))
     .enter().append("path")
     .attr("class", function(d) { return d.properties.kind; })
     .attr("d", path);
  });
});
```

Please let me know your preference so I know whether it's worth the trouble implementing this further.

maxbrunsfeld:
> Please let me know which path forward you think is the most promising:

This is what I'd like to do:

1. Revert the naming changes that you've introduced to the settings. In Atom, we try not to introduce unnecessary API changes, in order to provide a consistent platform for packages. In some cases, changes are necessary, but I don't see a strong reason for to change the names in this case.

2. See what specs fail after the original names are restored. I'm still not understanding why these changes (without the renames) can't be done in a backwards-compatible way. What would help me to understand is if you could take a look at the test failures on Travis, and explain specifically why one or more individual tests are failing.

3. If there are indeed unavoidable spec failures, open PRs on the relevant language packages with the minimal changes required to make the tests pass (no setting renames, just regexp changes or setting additions). I will publish pre-release versions of those packages (e.g. `v0.54.0-test-indentation-fixes`) based on those PRs' branches.

4. On this branch of atom/atom, as part of this PR, update Atom's package.json to point to those pre-release versions.

5. See the tests go green.

> write a completely new indentation logic, e.g., based on tokens

I agree with you that this would be a better approach.

I'm a little bit concerned that Atom's current tokenization is not reliable enough to provide truly correct indentation *either*. In the next year, we plan to start rolling out a new parsing system in Atom, one language at a time. The new system will provide an exact syntax tree for the program, instead of a sequence of tokenized lines based on regex matches. As we roll out that system, I am very excited about updating Atom's auto indentation system to make use of the syntax tree. I'm concerned that until that time, a complete rewrite of Atom's indentation system which will require changes to all bundled and third-party language packages may not be worth it.

That said, I don't want to discourage you from experimenting! Just be aware that such a PR will probably be very difficult to land.

chfritz:
Awesome, @maxbrunsfeld, I'll proceed as suggested.

Great to hear about the upcoming syntax trees. That did come to my mind very quickly as well as I was thinking more about it. I agree that that will lend itself to a much better logic. In that case I'll put my token-based work on hold and just prepare the line up of the PRs as suggested for now.

BTW, the short answer to 2. is that in the new logic the same regex can be matched multiple times per line, so 

```js
foo({ // one one opening scope is recognized
  a: 1,
  b: 2
  },
c);  // this is wrong

var d; // this would be at indentation -1, which is prevented by Math.max(desiredIndentLevel, 0)
```

now becomes

```js
foo({
    a: 1,
    b: 2
  },
  c);
```

Which is correct, and no need to rely on `max` to save us.

However, the test cases may desire the first behavior. But I'll investigate further. I should be able to make the tests go green.

alexandernst:
@maxbrunsfeld When you say `In the next year, we plan to start rolling out a new parsing system in Atom` do you mean <b>next</b> year (2017) or <b>this</b> (2016) year?

maxbrunsfeld:
I meant 2016, but as always, that is a very rough estimate.

chfritz:
@maxbrunsfeld I've renamed the regexes back and the tests seem to pass again. But now there seems to be some issue with the testing procedure itself, or something like that. If you look at the log of why the checks fail you'll see: 

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

Any idea how to fix that? I can't see how this is related to my code.

50Wliu:
Travis can be flaky at times...I've restarted the build.

chfritz:
Thanks, @50Wliu. Travis got past that issue now (it only seems to happen when checking the specs takes more than 10minutes). Now there is another error that I don't know how to interpret.

50Wliu:
> Now there is another error that I don't know how to interpret.

Another case of flakiness that you can safely ignore :)

maxbrunsfeld:
@chfritz Thanks for keeping at this. It's looking like a very tightly-focused change now :+1:. Can you add one or more tests that demonstrate the new and improved indentation? The example that you gave in your initial PR description seems like it would make a good test. I think that there should be a test that each regexp (`increaseIndent`, `decreaseIndent`, and `decreaseNextIndent`) can match multiple times.

chfritz:
@maxbrunsfeld : with pleasure! I was thinking the same.

chfritz:
ok, @maxbrunsfeld, I added three cases, one each for checking a case where one of the three regexs matches multiple times.

spec/fixtures/sample.js
```diff
@@ -10,4 +10,20 @@ var quicksort = function () {
...
+atIndentZero = true;
+
+bar({
+    atIndentTwo: true
+  });
+atIndentZero = true;
+
+bar({
+    atIndentTwo: true });
+atIndentZero = true;
```
maxbrunsfeld: Instead of modifying the `sample.js` fixture, could you just use a multi-line string in your test?

```coffee
beforeEach ->
  editor.setText """
    foo({
        atIndentTwo: true
      },
      atIndentOne
    );
    atIndentZero = true;
  """
```

spec/fixtures/sample.js
```diff
@@ -10,4 +10,20 @@ var quicksort = function () {
   };
 
   return sort(Array.apply(this, arguments));
-};
\ No newline at end of file
+};
+
+foo({
+    atIndentTwo: true
```
maxbrunsfeld: Some of this indentation looks incorrect to me. For example, take this code from the [node.js docs](https://nodejs.org/api/http.html#http_class_http_agent):

```js
http.get({
  hostname: 'localhost',
  port: 80,
  path: '/',
  agent: false  // create a new agent just for this one request
}, (res) => {
  // Do stuff with response
})
```

The keys in object literal, and the body of the function should be indented only one level.

spec/fixtures/sample.js
chfritz: ok, no problem.

spec/fixtures/sample.js
chfritz: well, define "correct", because there is absolutely no way to accomplish this indentation if you go with an inductive (line-to-line) approach unless you want to access incorrect indentations like the following:

```js
http.get({
  hostname: 'localhost',
  port: 80,
  path: '/',
  agent: false  // create a new agent just for this one request
}, 
(res) => {
// Do stuff with response
})
```

Here the `(res)` line has to be outdented because a scope ended (induction), but that's not correct. We would want this to still be indented.

Also, I would argue that the content of the object in the first argument should *not* be on the same indentation as the second argument, but rather that all arguments should be on the same indentation (which implies that the content of the first argument should be on indentation two. This becomes more clear when you rewrite this:

```js
http.get(
  {
    hostname: 'localhost',
    port: 80,
    path: '/',
    agent: false  // create a new agent just for this one request
  }, 
  (res) => {
    // Do stuff with response
  }
)
```

maxbrunsfeld: > incorrect indentations like the following

I think I agree with you that there is an inconsistency, but to be fair: the problem with your first example is more than just indentation: if the callback argument appears on a new line, the object should start a new line as well, as shown in your second example.

But the example I gave is very common. Another even *more* common pattern is this (also from the node.js docs):

```js
fs.readFile('/etc/passwd', (err, data) => {
  // indented one level
});
```

We need to preserve the indentation levels in these cases.

chfritz: well, as I said, that won't be possible with the current single-line look-back logic. Then we'll need to leave all these indentation bugs in the code until you release the AST logic you mentioned or bite the bullet and search back to the beginning of the scope -- which I believe I saw @svanharmelen trying to argue for but failed.

Frankly, I feel that putting aesthetics in specific case over general correctness (that follows some logic) is not a good idea. I don't take node.js docs to be the ground truth of what is correct indentation. And I don't think your editor should only indent correctly when you are writing clean code. It should *help* you write clean code, not try to mislead you when you don't. I really don't see what's wrong with

```js
fs.readFile('/etc/passwd', (err, data) => {
    // indented two levels
});
```

But if that's what you want, then again, you will have to give up correct indentation in many other cases such as this one:

```js
const child = exec('cat *.js bad_file | wc -l',
  (error, stdout, stderr) => {
    console.log(`stdout: ${stdout}`);
    console.log(`stderr: ${stderr}`);
    if (error !== null) {
      console.log(`exec error: ${error}`);
    }
  });
  // this should be at indentation zero, not one!
```

maxbrunsfeld: >  I really don't see what's wrong with: ...

Can you give some examples of popular open source code that's formatted like that? It seems highly unusual to me.

I feel your concern regarding consistency, but it's important not to lose the forest for the trees here. The purpose of formatting code is to make it easier for people to read. In that regard, formatting  has more in common with natural language than it does with mathematics; familiarity is important, and fitting in with the style of the document you're editing is important.

chfritz: I agree with the desirability of all of the above. I'm just pointing out the trade-off you guys will need to make until we can do all of the above right simultaneously. With the current logic we can't, so we need to pick the lesser evil. Sounds like I may just be an outlier here, although I haven't really seen too many issues regarding too much indentation. Most people seem to complain about inconsistent indentation -- such as indentation that is different before and after a block. That kind of consistency you won't get unless you buy into the counting logic I propose. For me it's a complete deal-breaker if I cannot auto-indent the whole file without risking screwing it up entirely. With all the files I've tested my logic has never done that, but the current one did on virtually any file I tried (the last line always had indentation much greater than zero in the end, which it shouldn't).

Please let me know what you want to do. There is nothing more I can so at this point (other than fix the test case).

muppetjones:
Wow...thanks for smoothing this out. Fantastic job.

Regarding the last discussed example:
```
fs.readFile('/etc/passwd', (err, data) => {
    // indented two levels
});
```
Personally, I would much rather have the two-level indentation and overall "more correct" indentation (i.e., indentation I have to spend less time correcting). If the code with the function is all indented correctly, I certainly would not mind the extra indentation for the entire block.

maxbrunsfeld:
I'm going to close this one out because we never really reached consensus on a way forward.

By the way, syntax trees have been available in Atom for a while now, if you enable the `core.useTreeSitterParsers` setting. We are pretty close to enabling this setting by default. In the next few months, I'd like to start designing a new auto-indent system that takes advantage of these syntax trees. I'll try to remember to CC some of you on any issues that I open.

chfritz:
@maxbrunsfeld I've taken a look at what the new tree-sitter parser produces. Pretty sweet! Since my [sane-indentation package](https://atom.io/packages/sane-indentation) was broken anyways, due to some recent changes I hadn't yet caught up with, I decided to "fast-forward" to tree-sitter and [rebuild the package based on that](https://github.com/chfritz/atom-sane-indentation/releases/tag/v0.9.3), at least for JavaScript. Thought you might be curious to give it a try. Note that it sets `core.useTreeSitterParsers` to `true`.

I think it's working pretty well. My biggest issue right now is dealing with parse-errors. When the program doesn't parse, then the entire indentation logic is broken, which only makes it harder to fix the syntax error in one's program. This is a bummer, because I always felt that good indentation is the key to finding syntax errors quickly. Any thoughts on how to overcome that problem are welcome!

Issue #10056, opened 2015-12-14
Failed to load the line-ending-selector package(EPERM: operation not permitted, write)


LendLord:
[Enter steps to reproduce below:]

1. Install atom
2. Error occured

**Atom Version**: 1.2.4
**System**: Unknown Windows Version (Windows 10, version 1511, build 10586.29)
**Thrown From**: [line-ending-selector](https://github.com/atom/line-ending-selector) package, v0.3.0

### Stack Trace

Failed to load the line-ending-selector package

```
At C:/Users/Lord/AppData/Local/atom/app-1.2.4/resources/app.asar/node_modules/line-ending-selector/lib/status-bar-item.js: EPERM: operation not permitted, write

Error: C:/Users/Lord/AppData/Local/atom/app-1.2.4/resources/app.asar/node_modules/line-ending-selector/lib/status-bar-item.js: EPERM: operation not permitted, write
    at Error (native)
    at Object.fs.writeSync (fs.js:663:20)
    at SyncWriteStream.write (fs.js:1981:6)
    at Function.log (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\node_modules\debug\node.js:106:17)
    at enabled (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\node_modules\debug\debug.js:117:11)
    at Logger.debug (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\lib\transformation\file\logger.js:92:31)
    at File.parse (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\lib\transformation\file\index.js:573:14)
    at File.parseCode (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\lib\transformation\file\index.js:691:20)
    at C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\lib\transformation\pipeline.js:167:12
    at File.wrap (C:\Users\Lord\AppData\Local\atom\app-1.2.4\resources\app.asar\node_modules\babel-core\lib\transformation\file\index.js:639:16)
```

### Commands

```
```

### Config

```json
{}
```

### Installed Packages

```coffee
# User
linter, v1.11.3
linter-php, v1.1.6

# Dev
No dev packages
```

[...]

PR #10395, opened 2016-01-12
Stop babel from logging to stderr during compilation

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/10056
Refs https://github.com/atom-minimap/minimap/issues/426

This stops babel from logging to stderr in the case where the user has the `DEBUG` environment variable set to `*` or `babel`. Babel doesn't have an API (except for the env var) for disabling this logging, so I've monkey-patched its `Logger` class, and created a feature request for adding a programmatic API: https://phabricator.babeljs.io/T6955.

### DIFF
--- a/spec/babel-spec.coffee
+++ b/spec/babel-spec.coffee
@@ -1,3 +1,8 @@
+# Users may have this environment variable set. Currently, it causes babel to
+# log to stderr, which causes errors on Windows.
+# See https://github.com/atom/electron/issues/2033
+process.env.DEBUG='*'
+
 path = require('path')
 temp = require('temp').track()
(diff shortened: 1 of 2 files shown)


joshaber:
Amazing. Thanks for digging into that!

maxbrunsfeld:
Ok, that snippets test failure looks unrelated to this change. Gonna :ship: this to get it into v1.5.0-beta0.

maxbrunsfeld:
Made an issue for the flaky snippets spec: https://github.com/atom/snippets/issues/188.

abe33:
Thank you for taking care of that! :bow:

PR #10400, opened 2016-01-13
Add non breaking space in Show Invisible (with tokenizer)

paradoxxxzero:
Implement non breaking space as a Show Invisibles character.

screenshot from 2015-12-18 18-02-01

This is an improved version of #10122 since it uses the tokenizer (and not a hack when rendering).

References #8080 #9483 
Closes #10122


wismill:
There is a special symbol to represent a non breaking space: "⍽" (Unicode: 237D).

There are also [variants](https://en.wikipedia.org/wiki/Non-breaking_space#Width_variation) of the non-breaking space. And also invisible operators: U+2061 to U+2064. Could they be included please?

paradoxxxzero:
Does not merge anymore with the recent changes, which is not that bad since https://atom.io/packages/highlight-bad-chars provides now a similar feature as a plugin.

I still personally think that show-invisibles should really show invisibles.

jerone:
> I still personally think that show-invisibles should really show invisibles.

Yeah, agree

Microsoft Word (and other text processors) use the `°` character to display non-breaking spaces.

ms-word-nonbreaking-spaces

sadick254:
I am closing this PR based on paradoxxxzero comment.
> Does not merge anymore with the recent changes, which is not that bad since https://atom.io/packages/highlight-bad-chars provides now a similar feature as a plugin.

PR #10409, opened 2016-01-14
Further improvements to pending item functionality

kuychaco:
Further improvements to #10178

Improve terminate pending state tests.
Serialize pending state for editor.
Keep pending item upon losing focus.


spec/native-compile-cache-spec.coffee
```diff
@@ -68,12 +68,12 @@ describe "NativeCompileCache", ->
...
   describe "when a previously required and cached file changes", ->
     beforeEach ->
-      fs.writeFileSync path.resolve('./spec/fixtures/native-cache/file-5'), """
+      fs.writeFileSync path.resolve(__dirname + '/fixtures/native-cache/file-5'), """
       module.exports = function () { return "file-5" }
       """
 
     afterEach ->
-      fs.unlinkSync path.resolve('./spec/fixtures/native-cache/file-5')
+      fs.unlinkSync path.resolve(__dirname + '/fixtures/native-cache/file-5')
```
nathansobo: I'm not sure there's actually a good reason to call `path.resolve` here. I think it's meant to be called with multiple arguments.

PR #10415, opened 2016-01-14
Don't cascade maximized windows

### LINKED ISSUE #10335: When the current window is maximized, opening a new window should open the new one maximized
From **@50Wliu** on https://github.com/atom/atom/pull/10236#issuecomment-169855151

> After testing this out on Windows it looks like there needs to be some extra Windows-specific code: opening a new window that spawns from a maximized one should stay maximized and not cascade.

winstliu:
Fixes #10335

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -369,7 +369,7 @@ class AtomApplication
   # Get the dimensions for opening a new window by cascading as appropriate to
   # the platform.
   getDimensionsForNewWindow: ->
-    return if (@focusedWindow() ? @lastFocusedWindow)?.isMaximized() and process.platform isnt 'darwin'
+    return if (@focusedWindow() ? @lastFocusedWindow)?.isMaximized()
     dimensions = (@focusedWindow() ? @lastFocusedWindow)?.getDimensions()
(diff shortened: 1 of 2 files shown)


src/browser/atom-application.coffee
```diff
@@ -369,6 +369,7 @@ class AtomApplication
   # Get the dimensions for opening a new window by cascading as appropriate to
   # the platform.
   getDimensionsForNewWindow: ->
+    return if (@focusedWindow() ? @lastFocusedWindow)?.isMaximized() and process.platform isnt 'darwin'
```
joshaber: I don't think we need to bother conditionalizing this since that's the right behavior for OS X, too.

joshaber:
Awesome, thank you! Just one note.

50Wliu:
@joshaber Changed!

joshaber:
Awesome, thanks again!

joshaber:
What do you think about cherry picking this onto beta so that this is fixed when the cascading windows ships?

50Wliu:
> What do you think about cherry picking this onto beta so that this is fixed when the cascading windows ships?

Sure thing!

Issue #6762, opened 2015-05-13
Ctrl-O should open the file chooser in the current directory


olsonpm:
This is exactly the same issue as #3618 .  I'm re-submitting an issue because the old fix does not work anymore as a result of AtomWindow.projectPath no longer existing (assumed to be the result of #770 where multiple project paths are now supported.

I was unable to figure out how to get console.log statements or errors reported to the dev console.  If someone can shed some light on how to get that working, then I can submit a PR.

MadLittleMods:
This also happens in Mac OSX. I would prefer to "open the currently open file's folder as default."

sneakypete81:
@olsonpm: The title of this bug is too specific:
 * The correct behaviour is to open in the directory of the currently open file
 * Mac OSX and Windows are affected too

Can you please update the title to something like:
"Ctrl-O should open the file chooser in the current directory".

As the original report in #3618 stated:
> Using keyboard shortcut ctrl-o in Ubuntu launches open file dialog with "Recently used" as default
> folder. The default in basically all other apps (including Sublime Text) is to open the currently open 
> file's folder as default.
>
> Yes, the tree view is there to quickly open files from current directory, but breaking convention is 
> bad.

afandria:
I agree. Please fix Ctrl+O. I am having a very hard time with Atom's current behavior.

olsonpm:
@sneakypete81 - I kept the title the same in order to keep it consistent with the original issue.  Changing the title assumes something different is occurring - I would like to keep it the same for that purpose.

monocasual:
Confirmed also in Ubuntu 14.04 with Atom 1.2.2.

kleinbottle:
I am a new atom-editor user who was previously using KDE's kate and kile. I am on Netrunner Rolling Release 2015-09, which is a derivative of Manjaro Linux, and ultimately Arch. The version of atom editor is 1.2.4.

Ctrl-O gives me "Recently Used" as reported here. What is worse, some of recent those files are PDFs, not intended for editing. 

Is there an installable package that allows circumventing this problem as quick workaround?

olsonpm:
@kleinbottle - if you find one please let me know.

The workaround I've been using is to utilize a different workflow.  I'm now accustomed to opening 'projects' (for which there are many plugins depending on how you want to manage those).  Within an open project, the fuzzy finder works very well - and there's even an "open buffer fuzzy finder" which  searches the files that you currently have open.

[...]

PR #10451, opened 2016-01-15
Change default path when opening files

afandria:
This is an attempt to use the active pane and active item to determine the default path when opening a file (on all platforms).

Note: Not tested yet. See #6762 for the discussion around this bug/feature.


lee-dohm:
This has been fixed in #11618. Thanks very much for your work on fixing this bug.

PR #10484, opened 2016-01-19
Honor created() function for an item in a submenu.

### LINKED ISSUE #10483: The created function is not honored for a menu item in a submenu
I believe it is due to this code:

https://github.com/atom/atom/blob/v1.4.0/src/context-menu-manager.coffee#L140-L144

Specifically, `Object.create()` is called on a root `item`, is screened for visibility via `shouldDisplay()`, and then its `created()` function is called, if it exists.

Unfortunately, this sequence is not applied recursively for submenu items.

bolinfest:
Fixes #10483.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "atom",
   "productName": "Atom",
-  "version": "1.4.0-beta3",
+  "version": "1.4.0",
   "description": "A hackable text editor for the 21st Century.",
   "main": "./src/browser/main.js",
   "repository": {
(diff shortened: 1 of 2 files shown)


bolinfest:
Hmm, I guess I screwed something up when I put up this PR? I built off of 1.4.0 because that's the latest stable revision.

bolinfest:
This is a mess: I'm going to try again here.

bolinfest:
I meant to delete this not merge it. Will fix ASAP.

PR #10486, opened 2016-01-19
Honor `created()` function for an item in a submenu.

### LINKED ISSUE #10483: The created function is not honored for a menu item in a submenu

bolinfest:
Fixes #10483.

### DIFF
--- a/spec/context-menu-manager-spec.coffee
+++ b/spec/context-menu-manager-spec.coffee
@@ -157,7 +157,7 @@ describe "ContextMenuManager", ->
         addError = error
       expect(addError.message).toContain('<>')
 
-    it "recursively applies templateForEvent on submenu items", ->
+    it "calls `created` hooks for submenu items", ->
       item = {
         label: 'A',
         command: 'B',
--- a/spec/context-menu-manager-spec.coffee
+++ b/spec/context-menu-manager-spec.coffee
@@ -156,3 +156,30 @@ describe "ContextMenuManager", ->
       catch error
         addError = error
       expect(addError.message).toContain('<>')
+
+    it "recursively applies templateForEvent on submenu items", ->
+      created = (event) -> @label = 'D'
+      item = {
+        label: 'A',
+        command: 'B',
+        submenu: [
+          {
+            label: 'C',
+            created,
+          }
+        ]
+      }
+      contextMenu.add('.grandchild': [item])
+
+      dispatchedEvent = {target: grandchild}
+      expect(contextMenu.templateForEvent(dispatchedEvent)).toEqual(
+        [
+          label: 'A',
+          command: 'B',
+          submenu: [
+            {
+              label: 'D',
+              created,
+            }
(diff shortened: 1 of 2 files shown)


bolinfest:
I'm curious if there's any chance of getting this patched in on the v1.5.0 release branch?

src/context-menu-manager.coffee
```diff
@@ -150,6 +147,19 @@ class ContextMenuManager
...
+  # Returns an object compatible with `::add()` or `null`.
+  cloneItemForEvent: (item, event) ->
+    return null if item.devMode and not @devMode
+    item = Object.create(item)
+    if typeof item.shouldDisplay is 'function'
+      return null unless item.shouldDisplay(event)
+    item.created?(event)
+    if Array.isArray(item.submenu)
+      item.submenu = item.submenu
+        .map((item) => @cloneItemForEvent(item, event))
```
ssorallen: These `item` variables shadow the outer `item`, which made this a bit tough to read the first time.

bolinfest: OK, changed.

bolinfest:
@nathansobo @maxbrunsfeld I'm hoping this should not be contentious since I believe it fixes a bug where the implementation does not match the spec: would either of you be able to take a look?

I'm not sure if there are CoffeeScript style changes or extra comments you'd like to see, but let me know and I'll make it happen. Thanks guys!

bolinfest:
I'd really like to get this in before the next branch is cut because this is blocking some menu management improvements I'm trying to make to Nuclide. Still trying to find a reviewer: @joshaber @mnquintana would you be able to take a look?

maxbrunsfeld:
@bolinfest Last week was the Atom team summit, but we will be able to take a look at this on Monday.

bolinfest:
@maxbrunsfeld Ah, sorry, didn't realize: thanks and have a good weekend!

spec/context-menu-manager-spec.coffee
```diff
@@ -156,3 +156,28 @@ describe "ContextMenuManager", ->
       catch error
         addError = error
       expect(addError.message).toContain('<>')
+
+    it "recursively applies templateForEvent on submenu items", ->
```
maxbrunsfeld: This description is a little bit implementation-specific. How about something like

```coffee
it "calls `created` hooks for submenu items", ->
```

maxbrunsfeld:
I left one small suggestion about the test. Otherwise, this looks great. Thanks for the fix!

bolinfest:
@maxbrunsfeld Thanks: done!

PR #10526, opened 2016-01-20
Fix open dialog default behavior on Linux

jrkoenig:
Open dialog now defaults to the same directory as the open file, or the user's home directory if there is no file active (such as when viewing the Settings page). This is the convention for gedit and other text editors on Linux. Because of the render/browser process split, this requires async messages to get the path over. The alternative is passing this data when calling application:open, but this requires more extensive changes.

Note: this change is only for linux but it could easily be expanded to other platforms assuming the dialog module works the same way.

Fixes at least #6762

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -213,6 +213,14 @@ class AtomEnvironment extends Model
 
     checkPortableHomeWritable()
 
+    setupOpenDefaultPath = ->
+      if process.platform is 'linux'
+        ipc.on 'get-open-default-path', (responseChannel) ->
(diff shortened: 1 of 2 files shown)


olsonpm:
fyi - I just built this now on debian 8 and your fix works.  I didn't realize your wording "user's home directory if there is no file active" - what should happen is that the current project directory opens if there is one.  The home directory should be the backup to that.  We also need to consider osx.  I'm looking into this now on a branch - given the debugging tips you gave.  I looked up the browser/renderer processes and can move forward with that reference now.

PR #10549, opened 2016-01-22
:arrow_up: nodegit@0.9.0

joshaber:
Most notably pick up https://github.com/nodegit/nodegit/pull/876.


maxbrunsfeld:
I've used this a fair amount and haven't seen any crashes.

PR #10552, opened 2016-01-23
improve docs about grouped settings

dirk-thomas:
See https://github.com/atom/settings-view/pull/698#issuecomment-171725215

Mentioning the grouping feature as "just another type" makes it pretty hard to find. Therefore I moved it into a separate section to make it more obvious.


Arcanemagus:
Ah, I had skimmed over `object` every time as I never had a need for an object type setting, that makes far more sense this way.

lee-dohm:
Thanks for contributing :grinning:

PR #10570, opened 2016-01-25
Improve working with folds

### LINKED ISSUE #8023: editor:cut-to-end-of-line on empty line above fold deletes the whole fold

wvanlint:
Hi,

I've been looking into making working with folds a bit smoother to fix #8023.

The functionality in this PR:
- Allows the end of the fold line on the screen to represent the end of the fold in the buffer. This makes us able to delete up until the boundaries of folds.
- More conservative fold expansion: only expand a fold when necessary.
- More conservative deletion: removed some cases where the entire fold would be deleted.

Let me know what you think!
I'm pretty new to CoffeeScript so any comments are welcome!

A demo:
demo


src/text-editor.coffee
```diff
@@ -2952,7 +2952,13 @@ class TextEditor extends Model
 
   # TODO: Rename to foldRowRange?
   createFold: (startRow, endRow) ->
-    @displayBuffer.createFold(startRow, endRow)
+    fold = @displayBuffer.createFold(startRow, endRow)
+    # Update the buffer positions based on the screen positions.
+    # For example, the end of the screen fold line maps to the end of the buffer fold.
+    # @moveCursors (cursor) ->
```
wvanlint: I would like to do update the cursors so the screen position stays in sync with the buffer position.
Let me know if this would be useful.

This breaks some tests though and I can't seem to fix some of them due to #10571:
- The fold creation would update the cursors, changing selections inside the fold.
- Adding the selections afterwards would remove the fold.

wvanlint:
In case you're considering merging this, I was also wondering if it would be better to set the invalidation strategy of folds to `inside`. This would help with, for example, deleting from the start to the end of the fold. The markers would also become exclusive automatically.

Issue #10435, opened 2016-01-15
Flaky text-editor-presenter spec


maxbrunsfeld:
https://travis-ci.org/atom/atom/jobs/102501079

Internal CI passed for this commit.

/cc @kuychaco

as-cii:
This seems related to #10089 and I suspect it's because of the call to `waitsForStateToUpdate`. Maybe bumping the timeout a little bit could solve the problem?

joshaber:
`GitRepositoryAsync` has a similarly flaky failure: https://travis-ci.org/atom/atom/jobs/102702554

```
>> GitRepositoryAsync
>>   .checkoutHeadForEditor(editor)
>>     it does not display a dialog when confirmation is disabled
>>       Error: Leaking subscriptions for paths: /private/var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/atom-working-dir116015-49834-1jigq4q/a.txt
>>         at ensureNoPathSubscriptions (/Users/travis/build/atom/atom/spec/spec-helper.coffee:123:15)
>>         at [object Object].<anonymous> (/Users/travis/build/atom/atom/spec/spec-helper.coffee:116:3)
>>         at process._tickCallback (node.js:368:9)
>> 
>> 
```

kuychaco:
Similar failure here - https://janky.githubapp.com/2716352/output
Passed just fine on travis - https://travis-ci.org/atom/atom/builds/106365370
(part of PR #10605)

```
>> TextEditorPresenter
>>   ::get(Pre|Post)MeasurementState()
>>     .content
>>       .cursors
>>         it updates when block decorations change
>>           timeout: timed out after 1000 msec waiting for presenter state to update
>>         it updates when ::scrollTop changes
>>           Error: Leaking subscriptions for paths: /private/var/lib/jenkins/workspace/atom/spec/fixtures/sample.js
>>             at ensureNoPathSubscriptions (/private/var/lib/jenkins/workspace/atom/spec/spec-helper.coffee:122:15)
>>             at [object Object].<anonymous> (/private/var/lib/jenkins/workspace/atom/spec/spec-helper.coffee:115:3)
>>             at process._tickCallback (node.js:368:9)
```

joshaber:
That one :point_up: is pretty annoyingly consistent.

kuychaco:
Another TextEditorPresenter spec failure, plus some Metadata.framework Error (that didn't cause any failures) - https://travis-ci.org/atom/atom/builds/108673841

Issue #2957, opened 2014-07-14
Add --reset option for wiping ~/.atom/storage


lee-dohm:
Occasionally we get instances where someone's editor has some wedged serialized state that is preventing them from opening Atom properly. When this happens, they need to be instructed on how to find `~/.atom/storage` (or the platform-specific equivalent) and to delete or move it. It would be very helpful to have an `atom --reset` command, like the `atom --safe` command that would either:

1. Delete `~/.atom/storage` prior to loading the rest of Atom
2. Package up `~/.atom/storage` and send it off to GitHub before deleting it (with the user's permission)

maherbeg:
:+1:

In reality, if Atom crashes multiple times in a row, then that directory should be blown away. This happens often enough that I'm starting to go a bit crazy. This has been particularly apparent when working on ruby code bases for me. I never had this problem on JS code bases before.

lee-dohm:
This will be available in Atom v1.7.x as `--clear-window-state`.

TomCaserta:
`--safe` should automatically enable this option, had the same issue and nothing was working until I found this cli option.

50Wliu:
`--clear-window-state` is permanent and potentially destructive.  Associating safe mode with completely wiping out all data, including unsaved, isn't what the Atom team is going for.  Safe mode is supposed to temporarily start Atom with no community packages installed, while `--clear-window-state` is supposed to allow you to break out of situations where Atom keeps hanging/crashing because of certain editor scenarios.

lee-dohm:
There is also complete documentation for all of this in the [Atom Flight Manual](http://flight-manual.atom.io) in the [Debugging section](http://flight-manual.atom.io/hacking-atom/sections/debugging/) including how and why to use each switch.

TomCaserta:
Ah, thank you, that makes sense. Hadn't seen that page to be honest lee-dohm, searched google for my issue with atom crashing opening a large file but it kept coming up with people requesting option to disable the window saving feature entirely.

tarah34:
I'm still not understanding these instructions, can you please help? I've read these and related posts via atom.io searching for resolving a bunch of tabs of files opening at startup that I haven't used for weeks.  I also read the debugging section in the Flight Manual.  However, I can't get the --clear-window-state function to work in Terminal.  Can you give easy steps (I'm a neophyte with Terminal) for how to try to resolve this problem with Atom tabbed files opening at startup that are from a very old session of using the app?  Thanks!
[...]

PR #10605, opened 2016-01-27
Periodically save state and store in indexedDB

### LINKED ISSUE #4161: Feature Request: add crash restore
Today my Atom crashed and I realized that there is no such thing as restore your unsaved work.
Since this is more than standard nowadays - please add this!

kuychaco:
Fixes [atom/atom#4161](https://github.com/atom/atom/issues/4161)
- [x] separate serialization from package deactivation
- [x] extract serialization logic from unloadEditorWindow
- [x] fix package manager tests
- [x] save state asynchronously on keydown and mousedown, write tests
- [x] get rid of AtomEnvironment state instance variable
- [x] fix AtomEnvironment tests
- [x] profile serialization time for various environment states
- [x] store state in IndexedDB
- [x] add command-line flag to clear indexedDB object store


spec/atom-environment-spec.coffee
```diff
@@ -174,6 +174,19 @@ describe "AtomEnvironment", ->
...
+      keypress = new KeyboardEvent('keypress')
+      atom.document.dispatchEvent(keypress)
+      advanceClock 1100
+      expect(atom.saveStateSync).toHaveBeenCalled()
+
+      mousedown = new MouseEvent('mousedown')
+      atom.document.dispatchEvent(mousedown)
+      advanceClock 1100
+      expect(atom.saveStateSync).toHaveBeenCalled()
+
```
kuychaco: Spy not working as expected and causing test to fail:

screen shot 2016-01-27 at 11 18 09 am

spec/package-manager-spec.coffee
```diff
@@ -821,6 +819,35 @@ describe "PackageManager", ->
           expect(atom.packages.isPackageActive("package-with-missing-provided-services")).toBe true
           expect(addErrorHandler.callCount).toBe 0
 
+  describe "serialization", ->
+
```
nathansobo: Super minor nitpick drive-by comment: :fire: this empty line. :smile:

jeancroy:
Is this saved to .atom/storage ?
On the forum, when atom crash on startup, deleting the storage folder is a standard debug step.
Maybe there's an interest in discarding most of the state but restoring unsaved file content.

spec/atom-environment-spec.coffee
kuychaco: Figured out this was an issue with binding `saveStateSync` before spying on it

kuychaco:
@jeancroy thanks for your comment. To answer your question - yes, the atom environment state is saved to .atom/storage. That said, I'm going to create an unsaved text buffers store that will be located elsewhere (possibly using IndexedDB). So in the event of a crash and deletion of the storage folder, your unsaved buffers will be preserved. 

The unsaved buffer store will be a part of a separate pull request. I've updated the title of this one to be more specific as to what aspect of crash recovery will be addressed here.

maxbrunsfeld:
This is looking good! How long does `saveStateSync` on average? Does it vary much depending on how many open buffers you have?

YuriSolovyov:
I wonder if we can use [requestIdleCallback](https://developers.google.com/web/updates/2015/08/using-requestidlecallback) here once we have chrome 47+

maxbrunsfeld:
> I wonder if we can use requestIdleCallback here once we have chrome 47+

Yeah, we should start using that, as soon as https://github.com/atom/atom/pull/9627 lands

src/storage-folder.coffee
```diff
@@ -6,11 +6,16 @@ class StorageFolder
   constructor: (containingPath) ->
     @path = path.join(containingPath, "storage") if containingPath?
 
-  store: (name, object) ->
+  storeSync: (name, object) ->
     return unless @path?
 
     fs.writeFileSync(@pathForKey(name), JSON.stringify(object), 'utf8')
 
+  storeAsync: (name, object) ->
```
nathansobo: I think this should be called `store`... the convention in Node is to have async be implied and only mention sync vs async on a sync alternative.

kuychaco:
Did some profiling. Here are durations for saveState (synchronous serialization, and async writing to disk):

| State | Time |
| ------- | ------- |
| 20 tabs opened | 2 ms |
| large unsaved buffer (jquery, 10K lines)* | 11 ms |

*Marker Layer serialization took 8.3 ms

20 tabs open:
screen shot 2016-02-01 at 5 58 09 pm

large unsaved buffer (jquery, 10K lines)
screen shot 2016-02-01 at 5 16 20 pm

as-cii:
Great job, @kuychaco. I think those are really impressive results: the ideal idle time is `50ms`, which, if I am interpreting the results correctly, means we have room for serializing ~ 5 large files or ~ 400 open tabs.

Also, I find very interesting that most of the time is spent in dumping markers: I guess it's because we have to perform a full traversal of the tree, which can be time consuming for large files. I still wonder, though, if we can dissect that `MarkerIndex` call and maybe optimize based on that.

Also, @kuychaco, what's the time spent in `JSON.stringify`?

I am really excited about this! :sparkles:

nathansobo:
> Also, I find very interesting that most of the time is spent in dumping markers

I'd like to implement serialization of the `MarkerIndex` directly for this case. Then we don't have to traverse the tree and build an alternate structure. Instead we can just convert the tree to JSON directly.

kuychaco:
/cc @atom/core feedback welcome. Note that the travis failure can be disregarded ([flakey spec](https://github.com/atom/atom/issues/10435))

This PR is intended to improve crash recovery by serializing and saving the AtomEnvironment state regularly to minimize data loss.

**Serializing**
Serializing will occur upon debounced mousedown and keydown events (debouce interval is set to 1 second). As part of this PR, I've cleaned up the `AtomEnvironment` code and done away with the `state` instance variable. Assuming that the serialized state is successfully saved, that means that worst-case-scenario a lovely Atom user will lose only 1 second worth of work upon crash. And speaking of successful saving...

**Saving**
With this PR, the state is now stored in IndexedDB rather than the ~/.atom/storage folder. Here are some of the benefits of using IndexedDB:
- transactional nature means:
  - when the disk is full we won't lose data as we would in the case of truncated files on disk
  - when two Atom windows have the same project folders open we don't have to worry about race conditions
- IndexedDB can store JS objects so there is no need to JSON.stringify/parse the state (possible performance improvements for startup time if deserializing large state object)
- state can be easily accessed in chrome dev tools, which is pretty cool I think
 screen shot 2016-02-11 at 5 36 13 pm

TODO: come up with a user-friendly way to clear the state from IndexedDB for [debugging purposes](https://github.com/atom/atom/issues/2957)

src/atom-environment.coffee
```diff
@@ -640,7 +648,7 @@ class AtomEnvironment extends Model
     @registerDefaultTargetForKeymaps()
 
     @packages.loadPackages()
-
+
```
nathansobo: :fire: this trailing whitespace

src/storage-folder.coffee
```diff
@@ -6,11 +6,16 @@ class StorageFolder
   constructor: (containingPath) ->
     @path = path.join(containingPath, "storage") if containingPath?
 
-  store: (name, object) ->
+  storeSync: (name, object) ->
     return unless @path?
 
     fs.writeFileSync(@pathForKey(name), JSON.stringify(object), 'utf8')
 
+  store: (name, object) ->
```
nathansobo: Is this async version used now that you're using IndexedDB?

nathansobo:
Very solid work @kuychaco. I did a pretty quick scan and things look good to me. Are you thinking of adding the `--clear-workspace-state` flag before merging?

kuychaco:
@nathansobo thanks for the review! I think I will add the command-line flag before merging. Hope to get that done tonight.

benogle:
:boom:

jeancroy:
Hi @kuychaco 

Is there mechanism to migrate state data from previous format to the new IndexedDB format ?
Here's one thread where a beta user think he lost all his state data on upgrade, but it was there, just ignored/not loaded.

https://discuss.atom.io/t/upgrading-to-1-7-0-beta0-blew-away-my-session-any-way-to-restore/26918
It could be interesting to avoid a widespread fear of data loss.

iolsen:
cc @as-cii ☝️

nathansobo:
Talked with @as-cii and he's planning on falling back to reading the state out of `~/.atom` if it is present and there's nothing in the database.

mehcode:
Great to hear there is a fallback in the works. :+1: 

---

> [...] People care more about their session state than I realized.

People (me) have even written a [package](https://github.com/mehcode/atom-project-plus) to list / inspect / make use of session state. It really is a nice feature of Atom.

miguelbarrio (commented):
8d55bbcdeadc41b07a1e5e710e45b4a939c399eb

PR #10610, opened 2016-01-28
Add TextEditor::{onWill,onDid}MutateSelectedText

as-cii:
Before this PR, we could catch edits to the buffer (or the editor) only on a per-change basis (e.g. `onDidChange` or `onDidInsertText`), thus preventing packages to schedule work at the beginning or at the end of a multi-cursor change.

This PR adds `onWillMutateSelectedText` and `onDidMutateSelectedText` on `TextEditor`, which will respectively trigger at the beginning and at the end of a `mutateSelectedText` operation.

@nathansobo @maxbrunsfeld @kuychaco: although similar, these new APIs serve a different purpose than the `TextBuffer.prototype.setTextsInBufferRanges` API (and the possible associated events) we have talked about, as they're meant to catch only user-triggered changes to the editor and they're much higher level, as they won't provide information about the changes (adding such information would indeed be tricky, as selections could do an arbitrary number of changes to the underlying buffer, and it would require a much more invasive refactoring).

On the other hand, adding these is super cheap and provides an entry point for coarse-grained batching operations. What do you think? I am planning to use them in autocomplete-plus (a PR is on the way to that repo which will show how it'd work).

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2984,6 +2984,22 @@ describe "TextEditor", ->
           expect(willInsertSpy).toHaveBeenCalled()
           expect(didInsertSpy).not.toHaveBeenCalled()
 
+      describe "when there are ::onWillMutateSelectedText and ::onDidMutateSelectedText observers", ->
+        it "notifies the observers just once before and after mutating the selections", ->
+          events = []
+
+          editor.onWillMutateSelectedText -> events.push("will-mutate")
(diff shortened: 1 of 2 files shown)


nathansobo:
I definitely see how these events help out at the moment, but I am a bit unclear about how they fit into the big picture once we start batching all change events. Not totally opposed, but it does seem like we'll have somewhat redundant APIs once we can get change events from the buffer / editor that include multiple changes... thoughts?

PR #10628, opened 2016-01-31
Properly restore initialPaths so project paths are not loaded as files

jrkoenig:
If project folders no longer exist or are not mounted, then Atom would open them as buffers, which are empty because the corresponding file did not exist (and if it did, it would be a directory, which would make little sense). This threads the right state through the startup sequence so that `initialPaths`, which actually are project roots, are restored correctly.


nathansobo:
I've solved this is a different way in #11023 if you're interested @jrkoenig. Let me know if I've missed something.

PR #10645, opened 2016-02-01
Only use electron version in fingerprint

kevinsawicki:
Previously the CI fingerprint used a SHA-1 of the entire `package.json` contents to determine when to clear the `node_modules` folder.

This would cause lots of clearings of that folder on CI since `master`, `beta`, and `stable` branches all have different `version` values that can change often.

Now only the `electronVersion` value is used in the fingerprint meaning that `node_modules` will only be cleared when the minor version of Electron changes which is why I believe this clearing was originally added before the fingerprint check was incorporated.

/cc @thedaniel @maxbrunsfeld

### DIFF
--- a/script/utils/fingerprint.js
+++ b/script/utils/fingerprint.js
@@ -7,7 +7,11 @@ var fingerprintPath = path.resolve(__dirname, '..', '..', 'node_modules', '.atom
 module.exports = {
   fingerprint: function () {
     var packageJson = fs.readFileSync(path.resolve(__dirname, '..', '..', 'package.json'))
-    var body = packageJson.toString() + process.platform + process.version
+
+    //Include the electron minor version in the fingerprint since that changing requires a re-install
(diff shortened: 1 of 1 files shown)


kevinsawicki:
This should save around 10 minutes on the build since that is how longer installing packages on CI seems to take currently.

/cc @iolsen

thedaniel:
What if the electron version remains the same but we add a dependency to package.json?

kevinsawicki:
> What if the electron version remains the same but we add a dependency to package.json?

That dependency will still get installed, this just prevents blowing away the entire `node_modules` when the build starts.

Is there a case where adding a dependency was causing issues previously when this was cached?

maxbrunsfeld:
:+1: This makes sense to me.

thedaniel:
> Is there a case where adding a dependency was causing issues previously when this was cached?

Nah, I was just thinking that we weren't installing at all with a cached node_modules but I see from the code that's not the case.

iolsen:
Nice!

PR #10646, opened 2016-02-01
Reuse Less cache across builds

kevinsawicki:
Previously each build recompiled all `.less` files for all the shipped themes so that the first Atom launch required no Less compiles.

This pull requests adds a cache in `~/.atom/compile-cache/prebuild-less` so that only changed files need to be recompiled across builds.

This should save around 30-60 seconds when no Less changes are made between builds and it will be incremental compiles whenever a file does change.
### Before (first compile)

```
prebuild-less:src    55.1s
```
### After (fully cached)

```
prebuild-less:src    1.5s
```

/cc @iolsen 

Refs https://github.com/atom/less-cache/pull/9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ script: script/cibuild
 cache:
   directories:
     - node_modules
+    - $HOME/.atom/compile-cache
 
 notifications:
   email:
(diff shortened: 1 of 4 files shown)


iolsen:
37x improvement after first build? :boom: :sunglasses:

joshaber:
Awesome :metal:

PR #10658, opened 2016-02-02
Default build directory to out/

kevinsawicki:
I'm struggling to remember why Atom originally built to a temp folder but this pull request changes the default build directory to be inside the repository at `out/` instead of `$TMPDIR/atom-build`.

This should make it easier to access the build once it completes when you are building but not installing since you'll be able to just do `open out/Atom.app` on Mac or `.\out\Atom\atom.exe` on Windows.

The reason I am making this change now is that AppVeyor's artifact upload system requires your artifacts be inside your project folder which got me thinking why Atom didn't follow that pattern like many other projects do like Node, Electron, etc.

/cc @atom/feedback

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ debug.log
 docs/output
 docs/includes
 spec/fixtures/evil-files/
+out
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ debug.log
 /tags
 /atom-shell/
 /electron/
+/out/
 docs/output
 docs/includes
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
:+1: :+1: :+1:

lee-dohm:
I like it! :grinning:

iolsen:
:+1: I struggled to find the output recently because I expected it to be somewhere under within the source tree!

mnquintana:
:+1: :heart_eyes:

Issue #10071, opened 2015-12-15
Highlight Atom Beta in Spotlight freezes/crashes Spotlight


maxsteenbergen:
Title says it all really. As soon as I highlight Atom Beta in Spotlight, it freezes Spotlight and I have to kill it in order to get it back. This issue is consistently reproducible.

Mac OS X 10.11.2
Atom Beta 1.4.0-beta0
├── atom-beautify@0.28.19
├── atom-material-syntax@0.2.10
├── atom-material-syntax-light@0.1.0
├── atom-material-ui@0.8.0
├── autoprefixer@3.1.0
├── color-picker@2.0.14
├── json-converter@0.2.1
├── less-than-slash@0.11.1
├── minimap@4.18.0
├── nucleus-dark-ui@0.6.14
└── pigments@0.19.3

lee-dohm:
Isn't this a bug for Apple if Spotlight is crashing?

maxbrunsfeld:
I can't reproduce this. I agree that it sounds like a Mac OS bug.

joshaber:
Yup, sounds like it. Anything interesting logged in Console.app?

lee-dohm:
Not that I can tell: https://gist.github.com/lee-dohm/9e40c230849f55ed87c0

joshaber:
Eh, yeah, definitely a problem in Spotlight.

maxsteenbergen:
I'm seeing the same error, so that's likely the culprit: ``Invalid parameter not satisfying: aString != nil``

https://gist.github.com/maxsteenbergen/b5f561f7f7e40fd8beca

ggamel:
I am receiving this same crash with Atom exclusively in Spotlight in OS X 10.11.2.

See screenshot of what happens when it begins/crashes:
![screen shot 2016-01-13 at 7 34 06 pm](https://cloud.githubusercontent.com/assets/34730/12313409/d44d61a4-ba2c-11e5-9175-96fa6cd4dc52.png)

I'll note this occurs multiple times a day/week. Appears random. I currently have three (3) Atom windows open with six (6) total tabs open between them.

ayushya:
I am also getting the same issue..
<img width="1280" alt="screen shot 2016-01-15 at 10 07 34 am" src="https://cloud.githubusercontent.com/assets/7705367/12345522/db9d5c0e-bb6f-11e5-89c3-b21ebe9782c7.png">

While Searching for atom in spotlight makes spotlight disappear and restarts it.
[...]

PR #10660, opened 2016-02-02
Only add local files to the recent documents, not URLs

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/10071

Also, add them in `Workspace::open` and not `AtomEnvironment::openLocations`, so that files opened in an existing window will also be added to the list.

/cc @atom/core

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -885,8 +885,6 @@ class AtomEnvironment extends Model
         else
           @project.addPath(pathToOpen)
 
-      @applicationDelegate.addRecentDocument(pathToOpen)
-
       unless fs.isDirectorySync(pathToOpen)
         @workspace?.open(pathToOpen, {initialLine, initialColumn})
 
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -471,6 +471,27 @@ describe "Workspace", ->
           workspace.open("bar://baz").then (item) ->
             expect(item).toEqual {bar: "bar://baz"}
 
+    it "adds the file to the application's recent documents list", ->
+      spyOn(atom.applicationDelegate, 'addRecentDocument')
+
+      waitsForPromise ->
+        workspace.open()
+
+      runs ->
+        expect(atom.applicationDelegate.addRecentDocument).not.toHaveBeenCalled()
+
+      waitsForPromise ->
+        workspace.open('something://a/url')
+
+      runs ->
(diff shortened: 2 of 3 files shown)


src/atom-environment.coffee
```diff
@@ -885,8 +885,6 @@ class AtomEnvironment extends Model
         else
           @project.addPath(pathToOpen)
 
-      @applicationDelegate.addRecentDocument(pathToOpen)
-
```
maxbrunsfeld: This method is only called when the browser process tells a render process to open a file (e.g. when initially opening a file from the terminal, or using a special command like `Application:  Open Your Stylesheet). It seems like we should add to the recent documents list *whenever* we open a file.

joshaber:
Really great work figuring this out :clap:

benogle:
> Really great work figuring this out

:+1: :8ball:

kevinsawicki:
:+1: Great work!

lee-dohm:
giphy

ayushya:
Thanks @maxbrunsfeld for addressing the issue and finally fixing it.

src/atom-environment.coffee
DouweM: @maxbrunsfeld I added the Recent Documents feature in https://github.com/atom/atom/pull/9609 because I wanted a way to quickly jump to "subjects" I recently worked on, where the subject can be either a project (folder) or a single file, but always an item I explicitly opened in a new Atom window.

screen shot 2016-02-03 at 21 58 04

If all files that are ever opened in a tab are added to Recent Documents, it loses all of its value for me, since the menu will always show 10 random files from the last project I worked with.

I don't see how that would actually be useful for anyone.

maxbrunsfeld: @DouweM I see your point about wanting to use the recent documents list as something of a 'Recent Projects' list, but I don't think it makes sense to use the recent documents list to represent that concept.

With the previous code, whether or not a file got added to the recent documents list depended on an implementation detail of *how* the file was opened. Opening a file with `command-o` would add it, but opening the same file with `command-p` would not. Opening your stylesheet with the `Application: Open Your Stylesheet` command would add it, but adding a root folder with `command-shift-o` would not add it.

Also, Atom's concept of a 'Project' can in general consist of multiple root folders. So the objects that were added to the recent documents list didn't correspond to projects consistently either.

You might want to use the [recent-projects](https://github.com/EXSyst/recent-projects) package for the workflow that you're describing.

PR #10673, opened 2016-02-03
Copy apm.cmd to the build directory when building

winstliu:
In #9585 I added `apm.cmd`, but forgot to copy it over when building...:weary:.

Thanks to @PaulAik for asking why `apm.cmd` wasn't available!


joshaber:
Nice catch. Thanks!

PR #10675, opened 2016-02-04
Expose application updater lifecycle events to packages

thedaniel:
This is WIP, I'm not sure the best way to write specs for this. There'll be a corresponding PR in atom/about for the UI.

The general idea is that the `AutoUpdateManager` which runs on the main thread will pass on events from the auto updater in Electron (or whatever needs to manage update checking in the future) corresponding with the update lifecycle - checking, available, not available, downloading, downloaded - to the renderer over IPC, the app delegate will proxy them, and then the Update wrapper (available at `atom.update`) will bind to the delegate and proxy them for clients (such as the about package)

---

Edit from @benogle:

Previously there was just an `update-available` and a `update-not-available` event. Respectively, these meant "there is an update, it's been downloaded and it's ready to install", and "atom is up to date". The only one exposed was `update-available` via `atom.onUpdateAvailable`. We needed more granularity for atom/about#12, so now there are more events.

This change adds a renderer process `AutoUpdateManager` which is hanging off of `atom.autoUpdater`. The events it supports are:

``` js
onDidBeginCheckingForUpdate(callback)
onDidBeginDownloadingUpdate(callback)
onDidCompleteDownloadingUpdate(callback)
onUpdateNotAvailable(callback)
```

This is probably an 80% point to something cleanish regarding update code. The underlying AutoUpdateManager still emits `update-available` events when the entire update has been downloaded. I was worried about breaking things, so that work can be done as a separate effort. These event names should be changed and `atom.onUpdateAvailable` deprecated in favor of events on `atom.autoUpdater`.


kuychaco:
@atom/core Ready to merge this. Feedback welcome!

benogle:
Mind if I do some renaming? `Update` seems like a weird name for a class. `AutoUpdateManager` would be more appropriate, even though there is already one in the main process.

benogle:
This would be more complete if we had something in electron to report a new version, but without auto-downloading. I've opened an electron issue: https://github.com/atom/electron/issues/4587

.gitignore
```diff
@@ -15,3 +15,4 @@ debug.log
 docs/output
 docs/includes
 spec/fixtures/evil-files/
+out/
```
benogle: @thedaniel what was this for?

benogle:
Ok I think this is ready. @nathansobo want to give this a once over?

src/application-delegate.coffee
```diff
@@ -175,8 +174,38 @@ class ApplicationDelegate
 
   onUpdateAvailable: (callback) ->
     outerCallback = (event, message, detail) ->
-      if message is 'update-available'
-        callback(detail)
+      # TODO: Yes, this is strange that `onUpdateAvailable` is listening for
+      # `did-begin-downloading-update`. We currently have no mechanism to know
+      # if there is an update, so begin of downloading is a good proxy.
+      callback(detail) if message is 'did-begin-downloading-update'
```
benogle: Note that this is changing the semantics of `onUpdateAvailable()`. It used to indicate that the update is downloaded and ready to be installed. Now, it will indicate that there is an update available, but before download and all that.

src/atom-environment.coffee
```diff
@@ -880,7 +886,8 @@ class AtomEnvironment extends Model
     @emitter.emit 'update-available', details
 
   listenForUpdates: ->
-    @disposables.add(@applicationDelegate.onUpdateAvailable(@updateAvailable.bind(this)))
+    # listen for updates available locally (that have been successfully downloaded)
+    @disposables.add(@autoUpdater.onDidCompleteDownloadingUpdate(@updateAvailable.bind(this)))
```
benogle: This change retains the old semantics.

.gitignore
benogle: Nevermind, I suppose I have an `out` folder as well. Maybe that's what the build does now?

src/auto-update-manager.js
```diff
@@ -0,0 +1,84 @@
...
+      applicationDelegate.onDidCompleteDownloadingUpdate((details) => {
+        this.emitter.emit('did-complete-downloading-update', details)
+      }),
+      applicationDelegate.onUpdateNotAvailable(() => {
+        this.emitter.emit('update-not-available')
+      })
+    )
+  }
+
+  dispose () {
```
nathansobo: For consistency with all other classes in Atom, we should probably call this `destroy`. Disposables are intended to represent an indirect claim on a specific resource, whereas all our application objects such as the workspace, panes, editors, selections, etc implement a `destroy` method enabling them to be destroyed. Destruction is a simpler concept. This object exists, and now it shouldn't any longer.

It's a subtle difference and maybe not one worth maintaining, but for now I think we should be consistent. If we really want to change this, we should deprecate `destroy` everywhere and call it `dispose`.

src/atom-environment.coffee
```diff
@@ -183,6 +187,8 @@ class AtomEnvironment extends Model
     })
     @themes.workspace = @workspace
 
+    @autoUpdater = new AutoUpdateManager({@applicationDelegate})
```
nathansobo: I notice you aren't calling `dispose` (which I suggested renaming to `destroy` in another comment). Should you do that in `AtomEnvironment.prototype.destroy`?

src/auto-update-manager.js
```diff
@@ -0,0 +1,84 @@
...
+    this.subscriptions.dispose()
+    this.emitter.dispose()
+  }
+
+  checkForUpdate () {
+    ipcRenderer.send('check-for-update')
+  }
+
+  restartAndInstallUpdate () {
+    ipcRenderer.send('install-update')
```
nathansobo: Technically, these interactions should also be routed through the application delegate. Its purpose is to stand in for all interactions with the global application state which facilitates testing and improves isolation.

src/auto-update-manager.js
```diff
@@ -0,0 +1,84 @@
...
+
+  onUpdateNotAvailable (callback) {
+    return this.emitter.on('update-not-available', callback)
+  }
+
+  getPlatform () {
+    return process.platform
+  }
+
+  // TODO: We should move this into atom env or something.
```
nathansobo: Feel free to do that.

nathansobo:
Cool. It will be great to have these details available in the render process. The basics look good. I made some structural suggestions.

src/atom-environment.coffee
benogle: done

benogle:
Alright, when the specs pass, I'm merging this.

nathansobo:
Thanks for making those adjustments.

benogle:
> Thanks for making those adjustments.

:+1:

benogle:
Thanks @thedaniel and @kuychaco for getting this started

kunkinkan:
What is the status of this? When can we expect it in a release?

iolsen:
@kunkinkan this shipped on stable in [1.7.0](https://github.com/atom/atom/releases/tag/v1.7.0) in April. 😎

kunkinkan:
@iolsen will this be made available for electron also? I don't believe they have all the events listed above? I thought atom was built on electron?

iolsen:
This change simply exposes the [underlying Electron autoUpdater events](http://electron.atom.io/docs/api/auto-updater/) to Atom packages. So either I'm misunderstanding or the functionality you're looking for is already available in Electron.

PR #10679, opened 2016-02-04
Add --build-dir example in custom installs documentation

n-riesco:
If atom was built in a custom directory, a custom install will require the use of `--build-dir`.


lee-dohm:
Can you make the build directories in both examples match?

n-riesco:
Done!

lee-dohm:
Thanks very much for contributing!

PR #10680, opened 2016-02-04
Automatically resize command-palette for small screens

### LINKED ISSUE #10603: Atom cuts command palette window if global window size is too small
I'm sure this one is cosmetic, but Atom cuts command palette window if global window size is set too small.
screen shot 2016-01-27 at 18 10 49

Aakash1312:
Fixes #10603 

Now command palette is automatically resized when global window size is very small.

### DIFF
--- a/src/panel-element.coffee
+++ b/src/panel-element.coffee
@@ -13,7 +13,22 @@ class PanelElement extends HTMLElement
     @classList.add(@model.getClassName().split(' ')...) if @model.getClassName()?
     @subscriptions.add @model.onDidChangeVisible(@visibleChanged.bind(this))
     @subscriptions.add @model.onDidDestroy(@destroyed.bind(this))
+    @subscriptions.add window.onresize = @updateModalSize
     this
+  
+  updateModalSize:=>
+    modalPanels = atom.views.getView(atom.workspace).panelContainers.modal.children
+    screenWidth = atom.getSize().width
+    fontSize = atom.config.get('editor.fontSize')
+    screenEM = screenWidth/fontSize
+    for a in modalPanels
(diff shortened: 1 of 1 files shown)


lee-dohm:
Are there any unit tests that can be written for this to prevent us from accidentally breaking it in future changes?

benogle:
Also, it looks like this makes the modals always scale to the width of the window. I think, if we are to fix this, we should keep the current width, and only scale when the window is too small.

Aakash1312:
Although the resize function is called every time window is resized, modals are scaled only when window width is less than the default modal size, ie, 50em. otherwise it sets the modal size to default value.
Is there any problem with this?

philipgiuliani:
I think this could be completly solved in CSS, or do i miss something? @simurai

```css
width: 80%; /* or different */
max-width: 50em;
transform: translateX(-50%); /* instead of - margin left */
```

simurai:
> And what if a theme wants to overwrite the width?

Yeah, some themes are probably overriding it. The `50em` comes actually from the One themes. Default from core is `500px`.

I can take a look.. maybe it's also possible with Flexbox or using a Media Query.

simurai:
Ok, here a fix using a CSS Media Query: https://github.com/atom/atom/pull/10782

@philipgiuliani `transform: translateX(-50%);` would work too, but it seems to make text blurry.

Flexbox would be nice, but more themes might be affected and needs more testing for side effects.

simurai:
#10782 that uses a CSS Media Query is now merged and closes this one. It should be easier to customize and maybe less taxing on performance.

@Aakash1312 Thanks alot for your initial PR. 🙇

PR #10691, opened 2016-02-04
Fix Javascript-related specs

winstliu:
These specs broke yesterday after Atom was updated to use `language-javascript@0.109.0`.  That version provided further control over numbers by splitting hex/octal/decimal numbers into different tokens.  Normal numbers are now `constant.numeric.decimal` instead of `constant.numeric`.

Will merge once specs are green.


50Wliu:
:green_heart:

PR #10708, opened 2016-02-05
Flow implies Babel

joshaber:
Something’s gotta remove all those sweet, sweet type annotations.

/cc @maxbrunsfeld @as-cii to make sure I didn’t miss anywhere else where we determine Babel-ness.


maxbrunsfeld:
:+1: Looks good to me.

joshaber:
Failing spec is unrelated and tracked in #10435.

PR #10712, opened 2016-02-05
Ignore more native source files

kevinsawicki:
Atom maintains a bunch of regexes of things to ignore when bundling the app to prevent bloat.

This pull request adds ignore patterns for the latest native modules added to the build; `nodegit`, `marker-index`, and `cached-run-in-this-context`. These patterns prevent `.c`, `.cc`, and `.h` files from being included in the built app.

This shrinks the `app.asar` overall size from `~151M` to `~105M` and the number of files in the bundle from `~18.5K` to `~14.1K`.  The bulk of the savings was from `nodegit` which was bundling all of the `openssl`, `libssh`, and `libgit2` native source code.

@iolsen and I were pairing today and we noticed this when running the `script/grunt output-build-filetypes` to display what types of files were bundled in the app.

/cc @joshaber @thedaniel :eyes:

### DIFF
--- a/build/tasks/build-task.coffee
+++ b/build/tasks/build-task.coffee
@@ -124,6 +124,7 @@ module.exports = (grunt) ->
     ignoredPaths.push "#{_.escapeRegExp(path.join('git-utils', 'src') + path.sep)}.*\\.(cc|h)*"
     ignoredPaths.push "#{_.escapeRegExp(path.join('nodegit', 'src') + path.sep)}.*\\.(cc|h)?"
     ignoredPaths.push "#{_.escapeRegExp(path.join('nodegit', 'generate') + path.sep)}.*\\.(cc|h)?"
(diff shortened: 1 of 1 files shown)


joshaber:
So they'll still be included in the app, but not in the asar? Or am I misunderstanding?

kevinsawicki:
> So they'll still be included in the app, but not in the asar? Or am I misunderstanding?

Previously all the `.c`, `.cc`, and `.h` files included in nodegit were ending up in `Atom.app`. This pull request removes them so only the required files like `.js` and `.a` files will be included.

kevinsawicki:
Also `node-pre-gyp` was previously being included as well via nodegit and it looks like it is only needed at install time, not at runtime.

joshaber:
:cool: :+1:

PR #10737, opened 2016-02-09
Add MRU tab switching functionality

### LINKED ISSUE #5344: Tab switching (ctrl-tab) should be in MRU order
Current switching order is as tab display order, which is useless when opening many tabs.
Other editors uses MRU order to improve it (Intellij IDEA, Sublimetext, VS.net, Windows alt-tab etc..)

It's better to popup an open file list (in MRU order) when do switching (like IntelliJ IDEA / Windows alt-tab does).
tabswitch_2366

natalieogle:
Fixes #5344 

First thing I did was give the `Pane` class an `itemStack` property to keep track of the order the items were used. The main function I added is `activateNextRecentlyUsedItem` which walks down the `itemStack` each time the user hits tab after ctrl is held down.

I set up the keymap for ctrl-tab but I still don't have the keymaps figured out for when ctrl is held down and tab is pressed over and over before lifting up on ctrl.  Something else I need to figure out is how to call a specific function to move an item to the top of the stack once ctrl is lifted.  

So, based on the assumption that I can make a keymap for holding down ctrl and pressing tab over and over (suggestions are welcome), I set up my code so that `activateNextRecentlyUsedItem` would be called each time the user hits tab when holding down ctrl.  Each call to this function would walk down the `itemStack`.  When the user lifted the ctrl key, somehow we would need to call `moveItemToTopOfStack` which would move the item the user stopped on to the top of the stack.

I modified the serialization, too, so the `itemStack` would persist.  The 4 scenarios I was focusing on are:
- nothing is wrong
- no `itemStack` serialized
- some items not serialized
- for some reason there is a different number of items in the `itemStack` than in the `items` array

I wrote tests for the MRU functionality and for the different serialization scenarios.

So yeah, this won't work correctly until I can figure out the key maps for holding ctrl down and hitting tab more than once and until I can figure out how to call `moveItemToTopOfStack` when ctrl is lifted.


keymaps/darwin.cson
```diff
@@ -73,7 +73,8 @@
   'cmd-alt-right': 'pane:show-next-item'
   'ctrl-pageup': 'pane:show-previous-item'
   'ctrl-pagedown': 'pane:show-next-item'
-  'ctrl-tab': 'pane:show-next-item'
+  'ctrl-tab': 'pane:show-next-recently-used-item'
+  'ctrl-tab ^ctrl': 'pane:move-item-to-top-of-stack'
```
benogle: Maybe this should be called `pane:move-active-item-to-top-of-stack`. I dunno. @nathansobo you are the naming master.

src/pane.coffee
```diff
@@ -278,12 +285,28 @@ class Pane extends Model
...
       @emitter.emit 'did-change-active-item', @activeItem
     @activeItem
 
+  # Build the itemStack after deserializing
+  addItemsToStack: (itemStackIndices) ->
+    if @items.length > 0
+      if itemStackIndices.length is 0 or itemStackIndices.length isnt @items.length or itemStackIndices.indexOf(-1) >= 0
+        itemStackIndices = (i for i in [0..@items.length-1])
+      for itemIndex in itemStackIndices
+        @addItemToStack(@items[itemIndex])
```
benogle: Should probably put a return at the end of this function, otherwise it will implicitly return an array because of the loop.

keymaps/darwin.cson
nathansobo: :+1: Specifying `active` is a good idea.

src/pane.coffee
```diff
@@ -39,18 +39,25 @@ class Pane extends Model
...
+    @addItemsToStack(params?.itemStackIndices ? [])
     @setFlexScale(params?.flexScale ? 1)
 
   serialize: ->
     if typeof @activeItem?.getURI is 'function'
       activeItemURI = @activeItem.getURI()
+    itemsToBeSerialized = compact(@items.map((item) -> item if typeof item.serialize is 'function'))
+    itemStackIndices = []
+    for item in @itemStack
+      itemStackIndices.push(itemsToBeSerialized.indexOf(item)) if typeof item.serialize is 'function'
```
benogle: So this looks a little awkward. Would be nice to use `compact`, but compact removes any falsy, so `0` items. A list comprehension works though: 

```
itemStackIndices = (itemsToBeSerialized.indexOf(item) for item in @itemStack when typeof item.serialize is 'function')
```

spec/pane-spec.coffee
```diff
@@ -837,3 +868,30 @@ describe "Pane", ->
...
+      newPane = Pane.deserialize(pane.serialize(), atom)
+      expect(newPane.getItems()).toEqual newPane.itemStack
+
+    it "rebuilds the itemStack if items.length does not match itemStack.length", ->
+      [item1, item2, item3] = pane.getItems()
+      pane.itemStack = [item2, item3]
+      newPane = Pane.deserialize(pane.serialize(), atom)
+      expect(newPane.getItems()).toEqual newPane.itemStack
+
+    it "does not serialize items in the itemStack if they will not be serialized", ->
```
benogle: Spec descriptions arent really my skill, but how about: `does not serialize items in the itemStack for items that will not be serialized themselves` @nathansobo?

benogle:
There should probably be a spec for when an item is removed. Other than that spec and the junk I mentioned, I think this is good to go :+1:

spec/pane-spec.coffee
nathansobo: I'm not sure I have enough context... do the items get serialized twice, once in the pane and once in the stack? Or are we somehow storing off a reference?

benogle: Effectively: when an item has no serializer, it shouldn't end up in the serialized `itemStack`.

nathansobo: Makes perfect sense, but it seems like a bad idea to serialize anything twice, if that's whats happening. Sounds like I better read the diff closer.

natalieogle: If I understand the question correctly, it is only serializing the indices of where each item in the `itemStack` is in `items`.  The items are only being serialized once.  Then `itemStack` is built with actual items after deserialization using those serialized indices.

nathansobo: Great, nevermind, I see you are just storing indices. :tada:

nathansobo: @natalieogle Yeah, I should have just looked before asking. Thanks!

natalieogle:
Made all changes and additions from your comments, yo.

maxbrunsfeld:
Just tried this out, and it's really nice.

There's one thing that I think we should add before merging: `ctrl-shift-tab` should cycle backwards through the stack, so that if you `ctrl-tab` too far, you can easily get back where you intended. @natalieogle Are you up for adding that?

natalieogle:
@maxbrunsfeld Definitely! I'll start that tonight.

benogle:
Ok, this is good to go!

jefkoslowski:
Hey,

Thank you for adding this feature. This was a heavily requested feature, but on the other side, a lot of users prefer the older way of switching tabs and expect that the editor would behave like that.

What was my surprise today when I started switching tabs and it was not working as expected. Did I broke something? Did I installed a new package that did that? Was this enabled by a keyboard shortcut that I removed? I didn't know! Then I went to Google and found some old [discussion](https://discuss.atom.io/t/keyboard-shortcut-for-cycling-through-tabs/698) suggesting me to add a shortcut config and it didn't work.

Finally, I found some people complaining on Twitter and a link to the 1.7 blog post with release notes.

> I don’t get in my car and find a message saying “for your convenience, the functions of the accelerator and brake pedals have been swapped”. ~ [source](https://reprog.wordpress.com/2014/08/29/dear-git-stop-changing-default-behaviour/)

So please, on the next releases, don't change the default behaviour. Instead, make it "opt-in" explicitly.

PavelVanecek:
I came here with the same problem as @jefkoslowski. For other poor souls searching for the solution how to disable the MRU behaviour and revert back to left-and-right:

Put this code in your _keymap.cson_:
```
'body':
  'ctrl-tab ^ctrl': 'unset!'
  'ctrl-tab': 'pane:show-next-item'
  'ctrl-shift-tab ^ctrl': 'unset!'
  'ctrl-shift-tab': 'pane:show-previous-item'
```

Source of this snippet: https://github.com/atom/atom/issues/5344#issuecomment-198216180

lemontheme:
Thank you so much, @PavelVanecek! I can see why people might prefer the new style of tab-switching, but it's definitely not for me. I would advise against making it the new default, but rather adding the option  to enable it somewhere in the GUI –  under the 'tabs' package, for instance.

PR #10743, opened 2016-02-09
Don't cascade on reload

### LINKED ISSUE #10648: On reload, window cascades
1. Open an atom window
1. Open another atom window
1. Move the windows so one is on the left of the screen, and one is on the right of the screen
1. Reload _the second window_ via `cmd-option-ctrl-l`
1. Note that it does not retain it's original position, but cascades under the other open window :(

I assume it's related to https://github.com/atom/atom/pull/10236

joshaber:
Fixes #10648.

So this fix is a little fussy.
1. Ideally we’d simply update the window load settings with the window dimensions before we reload, so that the reloaded window has the same dimensions.
2. **But** any changes we make to the window load settings in our `beforeunload` handler aren’t reflected in the reload. (I suspect Chromium’s grabbing the location for reload before calling the handler?)
3. So the next best thing: we detect a reload before `beforeunload` is called and update the window load settings there. This worked for _most_ reloads but I wasn’t able to find a hook for reloads triggered by cmd-r with the dev tools focused. That inconsistency would have been super annoying, to me at least.
4. So instead we track whether this is the first time the window’s been loaded. If it is, we use the default dimensions. If it isn’t, we use the dimensions we’re already saving in state.

@atom/core I’m curious if I’ve totally overcomplicated this or missed anything obvious.

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -384,6 +384,9 @@ class AtomEnvironment extends Model
   inSpecMode: ->
     @specMode ?= @getLoadSettings().isSpec
 
+  isFirstLoad: ->
+    @firstLoad ?= @getLoadSettings().firstLoad
+
   # Public: Get the version of the Atom application.
   #
   # Returns the version text {String}.
@@ -492,8 +495,6 @@ class AtomEnvironment extends Model
 
   # Extended: Reload the current window.
   reload: ->
-    @saveWindowDimensions()
-
     @applicationDelegate.restartWindow()
 
   # Extended: Returns a {Boolean} that is `true` if the current window is maximized.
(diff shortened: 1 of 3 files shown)


joshaber:
@maxbrunsfeld and I talked about removing the initial window dimensions from the load settings entirely. I'ma go down that road and see how it works out.

joshaber:
So I went down the road of putting `AtomWindow`/`AtomApplication` in charge of dimensions, and found it to have some other catches:

1. We store the default window dimensions in `localStorage` (https://github.com/atom/atom/blob/017bd2f716609731d2c927b88f8127cbe44f5c1f/src/atom-environment.coffee#L582). So the browser process can't get to it.
2. At the same time, in order to cascade windows properly, we have to take into account other windows, which is what passing the dimensions in from `AtomApplication` gives us.

I'd be really great to stop passing state between the browser and renderer using `location`, but that's more than I care to bite off for this PR.

@atom/core Any other thoughts? Are we OK with this as-is?

maxbrunsfeld:
>  Are we OK with this as-is?

:+1:

PR #10758, opened 2016-02-10
Fix status in subdir

### LINKED ISSUE #10746: All files in Tree marked as New(Green)
I just updated to Atom 1.5 and now when opening any folder within a local repo each file in the tree is marked as green (marked as a new file). If I click and open the file it un-marks it green and goes back to clean. Note: The files are marked as green but Git doesn't pick them up as new files so I can still work with the files per normal.

I ran in Safe mode and was still able to reproduce the bug. I also tried different UI and Syntax themes and it still occurs.

I did notice that if I open the project from the root of the repo it works as intended, and the issue only seems present when opening a folder deeper within the repo.

#### Tree in root Git project
root

#### Tree in sub dir of same project (files in screenshot are **NOT** new)
subdir

joshaber:
Fixes #10746.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "fs-plus": "^2.8.0",
     "fstream": "0.1.24",
     "fuzzaldrin": "^2.1",
-    "git-utils": "^4.1.0",
+    "git-utils": "^4.1.2",
     "grim": "1.5.0",
     "jasmine-json": "~0.0",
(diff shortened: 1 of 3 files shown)


joshaber:
This is ready :pray: There's an underlying libgit2 bug (https://github.com/libgit2/libgit2/pull/3609) but we can paper over it for now.

nathansobo:
:heart:

Issue #9974, opened 2015-12-08
Update all bundled packages to use new package.json fields, so their main modules can be required later


maxbrunsfeld:
Depends on https://github.com/atom/atom/pull/9975.
Don't merge these PRs until 1.4.0 is released.
Packages with ~~slashes~~ through them require no changes.

* [x] one-dark-ui https://github.com/atom/one-dark-ui/pull/113
* [x] one-light-ui https://github.com/atom/one-light-ui/pull/48
* [x] about ~~https://github.com/atom/about/pull/6~~, https://github.com/atom/about/pull/13
* [x] archive-view https://github.com/atom/archive-view/pull/32
* ~~autocomplete-atom-api~~
* ~~autocomplete-css~~
* ~~autocomplete-html~~
* [x] autocomplete-plus https://github.com/atom/autocomplete-plus/pull/637
* ~~autocomplete-snippets~~
* ~~autoflow~~
* [x] autosave https://github.com/atom/autosave/pull/56
* ~~background-tips~~
* ~~bookmarks~~
* [x] bracket-matcher https://github.com/atom/bracket-matcher/pull/197
* command-palette
* [x] deprecation-cop ~~https://github.com/atom/deprecation-cop/pull/65~~ https://github.com/atom/deprecation-cop/pull/67
* ~~dev-live-reload~~
* ~~encoding-selector~~
* ~~exception-reporting~~
* [x] find-and-replace https://github.com/atom/find-and-replace/pull/617
* [x] fuzzy-finder https://github.com/atom/fuzzy-finder/pull/160
* [x] git-diff https://github.com/atom/git-diff/pull/81
* ~~go-to-line~~
* [x] grammar-selector https://github.com/atom/grammar-selector/pull/29
* ~~image-view~~
* [x] incompatible-packages ~~https://github.com/atom/incompatible-packages/pull/11~~ https://github.com/atom/incompatible-packages/pull/12
* ~~keybinding-resolver~~
* [x] line-ending-selector https://github.com/atom/line-ending-selector/pull/17
* ~~link~~
* [x] markdown-preview ~~https://github.com/atom/markdown-preview/pull/349~~ https://github.com/atom/markdown-preview/pull/367
* ~~metrics~~
* [x] notifications https://github.com/atom/notifications/pull/105
* [x] open-on-github https://github.com/atom/open-on-github/pull/59
* [x] package-generator https://github.com/atom/package-generator/pull/37
[...]

PR #10764, opened 2016-02-10
Let packages define deserializers & view providers as main module methods

maxbrunsfeld:
This makes some further changes to the new `package.json`-based deserializer and view-provider APIs introduced in https://github.com/atom/atom/pull/9687. After updating some packages (see https://github.com/atom/atom/issues/9974) to use these new fields, I realized that it was somewhat awkward to specify deserializers and view providers in separate files from the packages's main module. I think it is better if these fields work similarly to `consumedServices` and `providedServices`, by specifying names of methods on the package's main module.
#### Example

`package.json`

``` json
{
  "name": "my-package",
  "main": "index",
  "deserializers": {
    "MyObject": "deserializeMyObject"
  },
  "viewProviders": [
    "buildMyObjectView"
  ]
}
```

`index.js`

``` js
const MyObject = require('./my-object')

module.exports = {
  activate () {
    // ...
  },

  deserializeMyObject (state) {
    return new MyObject(state)
  },

  buildMyObjectView (model) {
    if (model instanceof MyObject) {
      return new MyObjectView
    }
  }
}
```
#### Advantages
- This approach is more consistent with `consumedServices` and `providedServices`, which seem to be working well for a variety of use cases.
- Many packages currently maintain singleton instances of some view or model object, and keep a reference to it in their main module. This makes it easier for view providers and deserializers to create or update that singleton instance. The alternative would be for the instance to be stored as a global variable in the deserializer file, or in some third file that is required by both the main module and the deserializer.
- Similarly, this allows the deserializer to interact with any of the package's consumed or provided services with less boilerplate.
- Overall, I like the idea of the package's main module being the only file that needs global state, and all other files in the package exporting classes or functions.
#### Drawbacks
- The main module will be required earlier now, for packages that either 1) have view providers, or 2) have deserializers that are needed immediately. But _some_ code file would have to be required under these conditions regardless, so I think as long as the main module defers its requires effectively, it shouldn't make any difference in terms of load time.

/cc @atom/core

### DIFF
--- a/spec/fixtures/packages/package-with-deserializers/deserializer-1.js
+++ b/spec/fixtures/packages/package-with-deserializers/deserializer-1.js
@@ -1,6 +1,0 @@
-module.exports = function (state) {
-  return {
-    wasDeserializedBy: 'Deserializer1',
-    state: state
-  }
-}
--- a/spec/fixtures/packages/package-with-deserializers/deserializer-2.js
+++ b/spec/fixtures/packages/package-with-deserializers/deserializer-2.js
@@ -1,6 +1,0 @@
-module.exports = function (state) {
-  return {
-    wasDeserializedBy: 'Deserializer2',
-    state: state
-  }
-}
--- a/spec/fixtures/packages/package-with-deserializers/index.js
+++ b/spec/fixtures/packages/package-with-deserializers/index.js
@@ -1,3 +1,17 @@
 module.exports = {
-  activate: function() {}
+  activate () {},
+
+  deserializeMethod1 (state) {
+    return {
+      wasDeserializedBy: 'deserializeMethod1',
+      state: state
+    }
+  },
+
+  deserializeMethod2 (state) {
+    return {
+      wasDeserializedBy: 'deserializeMethod2',
+      state: state
+    }
+  }
 }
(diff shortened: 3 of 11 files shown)


izuzak:
@maxbrunsfeld Wondering if you can help me figure out something. I'm on `1.7.0-dev-9720541` (latest master version) and doing this:

1. start Atom
2. open Settings View
3. run "Window: reload" from command palette
4. observe that Settings View is not shown (this is unexpected behavior -- I expected Settings View to be shown)
5. open devtools console
6. observe a "No deserializer found" warning in console for the settings view state

Same thing happens for Archive View as well. 

Any chance this PR caused this behavior to change/break?

maxbrunsfeld:
Thanks for the heads up @izuzak! Looking into this now.

maxbrunsfeld:
@izuzak @joshaber Actually, I think this was caused by https://github.com/atom/atom/pull/10743. We need to load packages before deserializing our state, because packages provide deserializers that are used during state loading. Looking into how we can restore that ordering while addressing the window cascade problem....

joshaber:
Ugh, yeah, good catch. Could we make loading `@state` separate from deserializing?

PR #10768, opened 2016-02-11
Temporarily disable deserializers & viewProviders metadata fields

maxbrunsfeld:
See https://github.com/atom/atom/pull/10764

After merging the above PR in master, I'd like to avoid waiting for v1.7.0 to be released before I can merge PRs that update packages to use the new deserializer and view-provider APIs. I can change the packages in a backwards-compatible way, but I need to temporarily remove the logic that I added in https://github.com/atom/atom/pull/9687 for handling the slightly-different API that I had introduced previously.

This should be pretty risk-free, since it just removes calls to some methods that I recently added. Once it goes :green_heart:, my intent is to merge it directly into stable and publish a new patch release, just to unblock continuing efforts to use the improved APIs.

/cc @nathansobo

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -83,75 +83,6 @@ describe "PackageManager", ->
 
       expect(loadedPackage.name).toBe "package-with-main"
 
-    it "registers any deserializers specified in the package's package.json", ->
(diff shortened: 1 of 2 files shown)


Maxusz666 (commented):
#

> ``

PR #10782, opened 2016-02-12
Shrink modals when window gets narrow

### LINKED ISSUE #10603: Atom cuts command palette window if global window size is too small

simurai:
This solves the same as #10680 but using a CSS Media Query instead.
## Before:

When Atom's window get's narrower than `500px`, modals get cut off.

screen shot 2016-02-12 at 10 07 56 am
## After:

When Atom's window get's narrower than `500px`, modals will resize automatically.

screen shot 2016-02-12 at 10 07 12 am

Fixes #10603 closes #10680
## Affected themes

Themes that override the default of `500px` need to add their own Media Query.

``` less
@media (max-width: ???px) {
  atom-panel.modal {
    width: 100%;
    left: 0;
    margin-left: 0;
  }
}
```

When using `em`s, here another approach: https://github.com/atom/one-dark-ui/commit/4567ee785b7ffa8067684266b1fa7f3f5590dbbb


philipgiuliani:
Looks very good!

simurai:
✅  All checks have passed

Finally! Ok, merging.

PR #10789, opened 2016-02-12
Upgrade autocomplete-plus to use new text-buffer API

as-cii:
Refs: https://github.com/atom/autocomplete-plus/pull/659

What's missing:
- [x] `apm publish -t v2.27.1` for autocomplete-plus (it's currently failing with `Application Error`)

@maxbrunsfeld: I noticed you published a new version of autocomplete-plus but didn't upgrade it on atom/atom. The new version I plan to publish will include your changes: are you okay with this? Is there anything you were waiting for before updating ac+?

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "service-hub": "^0.7.0",
     "source-map-support": "^0.3.2",
     "temp": "0.8.1",
-    "text-buffer": "8.2.1",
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
@as-cii The package failed to publish due to an atom.io error for me yesterday. I am fine w/ upgrading it on master, just wasn't able to.

PR #10792, opened 2016-02-12
Fix TextEditorPresenter spec timeout.

### LINKED ISSUE #10435: Flaky text-editor-presenter spec

joshaber:
Fixes #10435.

I haven’t been able to reproduce this locally yet, but let’s see if it’s as easy as giving it a longer timeout.

### DIFF
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -20,7 +20,6 @@ describe "TextEditorPresenter", ->
 
       buffer = new TextBuffer(filePath: require.resolve('./fixtures/sample.js'))
       editor = atom.workspace.buildTextEditor({buffer})
-      editor.setUpdatedSynchronously(true)
       waitsForPromise -> buffer.load()
 
     afterEach ->
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -91,7 +91,7 @@ describe "TextEditorPresenter", ->
     expectNoStateUpdate = (presenter, fn) -> expectStateUpdatedToBe(false, presenter, fn)
 
     waitsForStateToUpdate = (presenter, fn) ->
-      waitsFor "presenter state to update", 1000, (done) ->
+      waitsFor "presenter state to update", 2000, (done) ->
         fn?()
         disposable = presenter.onDidUpdateState ->
           disposable.dispose()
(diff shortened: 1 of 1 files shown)


joshaber:
A parallel `atom` build failed: https://janky.githubapp.com/2759555/output

This kinda seems like a legit race.

joshaber:
This is ready.

The problem was with `waitsForStateToUpdate`. We'd call a function that emitted a state update, and _then_ start watching for state updates. So we were racing between emitting the event and watching for it.

I introduced `waitsForStateToUpdateAsync` instead of fixing `waitsForStateToUpdate`, because when I changed `waitsForStateToUpdate` it broke some other tests which I didn't understand.

/cc @as-cii @nathansobo :eyes: I think you two did most of the work here.

spec/text-editor-presenter-spec.coffee
```diff
@@ -20,6 +20,7 @@ describe "TextEditorPresenter", ->
 
       buffer = new TextBuffer(filePath: require.resolve('./fixtures/sample.js'))
       editor = atom.workspace.buildTextEditor({buffer})
+      editor.setUpdatedSynchronously(true)
```
nathansobo: Interested in why this was needed...

nathansobo:
Thanks for this. Would be nice to only have one `waitsForStateToUpdate` method, but I'd rather fix the flakiness and get back to it later I suppose.

spec/text-editor-presenter-spec.coffee
joshaber: It wasn't needed, but it made the test failures deterministic so I could ensure it was fixed.

nathansobo: Ah, do you think we could drop it now then? I feel like the test is more realistic overall without it.

joshaber:
>Would be nice to only have one waitsForStateToUpdate method, but I'd rather fix the flakiness and get back to it later I suppose.

Yeah, I agree. I time boxed this so I could get to some of the more pressing FR issues. Maybe the next FR could dig into this further.

spec/text-editor-presenter-spec.coffee
joshaber: :+1:

nathansobo:
> Yeah, I agree. I time boxed this so I could get to some of the more pressing FR issues. Maybe the next FR could dig into this further.

Totally reasonable.

joshaber:
We're back down to one `waitsForStateToUpdate` now :metal:

nathansobo:
> We're back down to one waitsForStateToUpdate now :metal:

:pray: Namaste

maxbrunsfeld:
:moneybag: :train: :dollar: :steam_locomotive:!!

as-cii:
Thank you, @joshaber! :boom:

PR #10817, opened 2016-02-14
Fixed atom documentation hyperlink

arnav-gudibande:
Changed the atom/link to Atom Documentation


mnquintana:
Thanks for contributing! However, this link is for the source of the documentation, not the generated documentation, which is why the link text was `atom/docs`.

Issue #6860, opened 2015-05-20
Unable to set atom as default program to open files


cytsunny:
I am using atom on Windows 7. After the recent update to version 0.200.0, I am not able to set atom as my default program to open files. 

Before the (auto)update, atom is set as my default program to open .js, .php, .blade.php, .json and .html files. However, after the update, the system is unable to find atom on the default program list and therefore ask me for a default program again. I am able to find and select atom when browsing the file system when asked to set the default program, but when I click open, it fails to set as default.

![atom_error](https://cloud.githubusercontent.com/assets/3927120/7718471/6130ce68-fee3-11e4-8f76-40b0da233adb.png)

As shown above, it can be found but when I click open, nothing happens.

The atom runs smoothly with no error message shown.

TouPye:
Did you try to reboot your computer after the update? 
I know that WIndows sometimes has weird issues after some program updates..

50Wliu:
Related to #4893.

cytsunny:
I have tried rebooting the computer after the udpate. (or to be exact, I turned off my computer when I left office yesterday and turn it on again today) The problem still exists.

kokuou:
I have the same issue. Using the executable in the newest update as the default program will not launch Atom or open the file. The atom.cmd in the bin folder will work, but then you get an ugly icon and a flash of a command prompt before opening every file.

cytsunny:
Thanks @kokuou  your feedback Your feedback helps me at least being able to set atom as default program, so now I am having the same problem as you. A falsh of a command prompt is not a big problem to me, but the icon is really ugly.

Yet at least it is not a big deal now and the atom team can lower the priority a bit.

dmase44:
I have the same issue on Win7. Previous to the 1.2.0 update my file association were working fine and the atom.ico was present in default programs. Now every workaround I have tried does not work. Most end up pointing to Update.exe witch does nothing. The atom/app-1.2.0/atom.exe does open atom but doesn't show in the default programs to select for file associations even after setting it in the variable path. This is a shame having to right click on a Brackets icon to open with Atom.

jotun-from-muspelheim:
+1, same issue.

damieng:
Fix checked in to master. Will now appear in the list of apps from "Open With..." and the update will take care of updating/correcting the exe path.

DingBEN:
NOPE... first install on WIN 7 total ZERO... no shortcuts, and all the other problems listed...
[...]

Issue #6177, opened 2015-03-31
Setting Atom as my default editor in Cyberduck


TafariD:
So when Atom installed itself, I have no idea where it placed its files... 
I've been trying to set Atom as my default editor in Cyberduck, the issue is, I can't find the "Atom" file in my computer to select in the file window.

I tried selecting the Atom shortcut that was installed on my desktop, but Cyberduck claims it points to a file called "update" and then refuses to work. I was able to set Cyberduck to work with Sublime, but I can't figure out how to set Atom as my default editor...

Does anyone know how to fix this?

![where](https://cloud.githubusercontent.com/assets/10626101/6910174/7ab7866c-d71d-11e4-9735-bdd18a6321a7.png)

nramin:
I'm having the same problem. Anyone know?

batjko:
There is a `.cmd` file for atom (next to one for apm) that your link should use as a target in:
`C:\Users\[username]\AppData\Local\atom\bin`

![image](https://cloud.githubusercontent.com/assets/7287076/6996306/43301df8-db78-11e4-87ed-be2b04e787cc.png)

TafariD:
It doesn't work. The edit tab in Cyber Duck is still greyed out,  and I can't open files with Atom.

On Sun, Apr 05, 2015 at 4:45 AM, batjko <notifications@github.com<mailto:notifications@github.com>> wrote:


There is a .cmd file for atom (next to one for apm) that your link should use as a target in:
C:\Users\[username]\AppData\Local\atom\bin

[image]<https://cloud.githubusercontent.com/assets/7287076/6996306/43301df8-db78-11e4-87ed-be2b04e787cc.png>

batjko:
I don't know at all what your CyberDuck's requirements are for declaring editors. 

Maybe it needs to be the actual executable, which theoretically you could get by going directly into the latest version's `bin` folder. But then you'd have to manually change it every time an update comes out.

damieng:
Atom is now registered as an application for association with files.

Cyberduck however still has Edit With grayed out and I wasn't even able to get that working with Notepad. I suspect there is either an issue their end or they are expecting the application to be able to open url's.  Atom worked fine from the open command on the transfer window.

I think our part of the work here is done and you might want to follow up with Cyberduck - I found issues on their tracker relating to Edit With not working but they were all OS X related and subsequently fixed.

If you can get Cyberduck Edit With working with another app please post back so we can find out what the difference is.

pentracode:
[...]

PR #10818, opened 2016-02-14
Register Atom for file associations in Windows

damieng:
:checkered_flag: Register atom.exe as a file handler for the current user including updating the .exe location on upgrade and removing it on uninstall. 

Addresses #4893 and #6860. Partially addresses #6177 in that we now do the right thing although Cyberduck on Windows still seems to have some issues - I couldn't get Edit With to even work with Notepad although the Open With from the transfer Window now works with Atom.

Tested on Win7/x86 and Win10/x64:
handlerwin7
handlerwin10

### DIFF
--- a/src/browser/squirrel-update.coffee
+++ b/src/browser/squirrel-update.coffee
@@ -20,6 +20,7 @@ else
 fileKeyPath = 'HKCU\\Software\\Classes\\*\\shell\\Atom'
 directoryKeyPath = 'HKCU\\Software\\Classes\\directory\\shell\\Atom'
 backgroundKeyPath = 'HKCU\\Software\\Classes\\directory\\background\\shell\\Atom'
+applicationsKeyPath = 'HKCU\\Software\\Classes\\Applications\\atom.exe'
 environmentKeyPath = 'HKCU\\Environment'
 
 # Spawn a command and invoke the callback when it completes with an error
@@ -64,6 +65,10 @@ installContextMenu = (callback) ->
     args.push('/f')
     spawnReg(args, callback)
 
+  installFileHandler = (callback) ->
+    args = ["#{applicationsKeyPath}\\shell\\open\\command", '/ve', '/d', "\"#{process.execPath}\" \"%1\""]
+    addToRegistry(args, callback)
+
   installMenu = (keyPath, arg, callback) ->
     args = [keyPath, '/ve', '/d', 'Open with Atom']
     addToRegistry args, ->
@@ -74,7 +79,8 @@ installContextMenu = (callback) ->
 
   installMenu fileKeyPath, '%1', ->
     installMenu directoryKeyPath, '%1', ->
-      installMenu(backgroundKeyPath, '%V', callback)
+      installMenu backgroundKeyPath, '%V', ->
+        installFileHandler(callback)
 
 isAscii = (text) ->
   index = 0
@@ -124,7 +130,8 @@ uninstallContextMenu = (callback) ->
 
   deleteFromRegistry fileKeyPath, ->
     deleteFromRegistry directoryKeyPath, ->
-      deleteFromRegistry(backgroundKeyPath, callback)
+      deleteFromRegistry backgroundKeyPath, ->
+        deleteFromRegistry(applicationsKeyPath, callback)
 
 # Add atom and apm to the PATH
 #


Zireael07:
Great!

Riddler84:
Is this actually implemented? I'm using Windows 10 and i don't see a symbol like the one in the pics above. It don't appear in the list. I use the latest version.

50Wliu:
@Riddler84 this will be available in Atom 1.7.0.

Riddler84:
@50Wliu ok, thanks.

fcathala:
I'm using Atom 1.7.1 and don't see any change in the way file types are registered. Can you please confirm this fix is shipped and if there are any known caveats (as in "full uninstall", etc...)?

damieng:
It doesn't associate itself with any file types by default (e.g. .js) if that's what you were expecting.

When you do Open With... however you should see Atom as one of the applications as shown in the screenshots above.

NJDEN:
Will Atom be getting a separate file type icon independent of the application icon? I'm not really a fan of having all my Atom associated files using Atom's icon.

damieng:
That is a good idea, do you want to raise an issue for it and label it Windows and Feature?

fcathala:
Thanks @damieng, I know it doesn't associate automatically.
I can't see it in the list of "Open With" capable apps to open a specific file (I tried with a .js and a .txt) which is the documented feature.
Is it supposed to be dealt with at installation time or am I expected to run a separate patch?
Happy to send logs or any troubleshooting information if you ask for. :)

damieng:
It should just automatically be there as part of the upgrade/install. We do however only associate it for the current user.

Can you check your registry to see if you have the key HKCU\Software\Classes\Applications\atom.exe

fcathala:
I don't have a branch called "Applications" within "HKCU\Software\Classes"... Let me know if you think of anything. I'll continue the investigation my side and post my findings later.

damieng:
That's strange. What version of Windows are you on?

Do you have a HKLM\Software\Classes\Applications ?

fcathala:
Windows 10 with all updates.

There's no Application key at all...

On Sunday, 17 April 2016, Damien Guard <notifications@github.com> wrote:

> That's strange. What version of Windows are you on?
>
> Do you have a HKLM\Software\Classes\Applications ?
>
> On Sun, Apr 17, 2016 at 3:28 PM Fabrice Cathala <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > I don't have a branch called "Applications" within
> > "HKCU\Software\Classes"... Let me know if you think of anything. I'll
> > continue the investigation my side and post my findings later.
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/atom/atom/pull/10818#issuecomment-211124448>
> >
>
> —
> You are receiving this because you commented.
> <https://github.com/atom/atom/pull/10818#issuecomment-211125223>
>


-- 
Sent from Gmail Mobile

damieng:
It's possible no applications have registered themselves this way yet (there are other ways to associate with specific files etc).

Did your machine auto-update or was this a fresh install?

Are you running on a Windows domain-joined machine that might have additional policies applied?

fcathala:
Auto-update via Microsoft Windows 10 campaign.

No domain, no policy it's a personal PC.


On Monday, 18 April 2016, Damien Guard <notifications@github.com> wrote:

> It's possible no applications have registered themselves this way yet
> (there are other ways to associate with specific files etc).
>
> Did your machine auto-update or was this a fresh install?
>
> Are you running on a Windows domain-joined machine that might have
> additional policies applied?
>
> —
> You are receiving this because you commented.
> <https://github.com/atom/atom/pull/10818#issuecomment-211126527>
>


-- 
Sent from Gmail Mobile

damieng:
Sorry I meant how did Atom get to 1.7 - fresh install or upgrade from 1.6?

fcathala:
Fresh install.

fcathala:
Works well on a different Win 7 PC though. So, I guess that's a local environment issue and I will investigate. Thanks for your help.

damieng:
Hmm.  Did it add the general purpose top-level "Open in Atom" context menu
options to the shell?

fcathala:
Indeed. It did from a fresh install on PC 2 (running Windows 7). Let me have a look and I'll update this thread. :)

DingBEN:
**FAIL** and some success!
Could someone(sysOP) clean-up this thread and consolidate all the key information in a concise post AT THE TOP of the thread!?

**My scenario, ex:** WIN7, first install of 1.8.0 atom.exe...

**1. run after install**, luckily I checked it or I would never have even seen the tool.

**2. once I exited**, had no way to start atom!? - no shortcuts ANYWHERE
       - had to do a FileSystem search to find where installed

**3. clicking on atom.exe**
    does nothing
    O! yes it does! it starts it but does not show the visual interface, only shows as an entry in the Task-Manager.!?

**4. going to DOS**
    and using .cmd or atom FAIL (spews 1/2 screen of 'crash' lines on the DOS window and hangs)

**5. context menu(right-click)**
       - **NO (Open with Atom)** even after having manually associated extensions to be ATOM default and icon shows correctly on the file type, like .css, .js, ...
       ...AND THE BAD PART ABOUT THAT
       ...ASSOCIATING .htm/.aspx/.php/etc...simply a CLUMSY APPROACH - GETS IN THE WAY...
          these extensions need to be
         - DefaultBrowser-associated to **EXECUTE**,
         - and instead, **(Open with Atom)** in the context menu!

      - **Open with** - Choose default program - Browse to fetch Atom FAILs (as in 4 - above)
         ...2 to 4 clicks and delays - RIDICULOUS

**6. adding a registry key,**
      to place (Open with Atom) in the context menu(right-click)
      have not tried that yet, but have had problems with other editors...
      was able to get (Open with notepad2) and (Open with Wordpad) to appear in context menu
     ...but it was a struggle (with Wordpad definitely)
**Will post an update HERE if I succeed with that!  (any suggestions?!)**

**7. made a shortcut:**
   C:\Users\User\AppData\Local\atom\app-1.8.0\atom.exe (with same target)
   - that seems to dig the atom from the depth of KLUDGEs...
   - atom comes up and shows as was last exited
     and remembers the file displayed is an ASPX which was set as HTML manually or no syntax color would show!   **(wonder if that will stick... where is it stored in the config!?)**

damieng:
@DingBEN there is something very wrong with your installation. Please create a new issue and include your squirrelsetup.log file

DingBEN:
ok... my post reappeared now LOL
thank you for this rapid answer.
what do I name this new issue?

sannidhi01:
Team: As this is a very handy feature, please consider to move up in the priority list. And, do push in an early release ASAP. Thanks!
Note: I'm aware of work arounds.

Arcanemagus:
This has been implemented in Atom for several years now, you should be able to enable it if it isn't by default under the Settings -> System menu:
image

If you are having issues getting it working please file a _new_ issue, making sure you fill out the template.

Thanks!

PR #10797, opened 2016-02-12
Fix status with multiple paths

joshaber:
Fixes #10757.

PR #10827, opened 2016-02-15
Port repository subdirectory status fixes to async repo

joshaber:
Bring the fixes over from https://github.com/atom/atom/pull/10758 and https://github.com/atom/atom/pull/10797.


joshaber:
Test failure on that last build is fixed in https://github.com/atom/atom/pull/10828.

PR #10835, opened 2016-02-16
Add TextEditor.prototype.cursorsForScreenRowRange

as-cii:
With this PR, `TextEditorPresenter` can avoid to scan through all the cursors to understand whether they're visible on screen. This dramatically reduces the calls to `getScreenRange` and, thus, makes `updateCursorsState` faster for scenarios where the number of cursors is elevated (e.g. editing, scrolling through a file, etc.).

Although this code path will partially, and positively, be affected by the in-progress `DisplayLayer` redesign, I believe that reducing the number of cursors we need to iterate is still valuable in terms of performance and reduces the number of `getEnd()` queries we need to perform against `MarkerIndex`.
#### Before

screen shot 2016-02-16 at 14 44 49

_(Total: `~ 1183ms`, benchmarks performed by editing 3000 lines, writing 5 letters)_
#### After

screen shot 2016-02-16 at 14 46 45

_(Total: `~ 98ms`, 12x faster :racehorse:)_

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -432,18 +432,14 @@ class TextEditorPresenter
     return
 
   updateCursorsState: ->
-    @state.content.cursors = {}
-    @updateCursorState(cursor) for cursor in @model.cursors # using property directly to avoid allocation
-    return
-
-  updateCursorState: (cursor) ->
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Good idea! Makes a lot of sense.

as-cii:
Thanks, @maxbrunsfeld! :sparkles: The failure on AppVeyor is a false negative, so I am going ahead to :ship: this.

PR #10838, opened 2016-02-16
Log output for core specs

### LINKED ISSUE #10837: Core specs task sometimes time out on Travis
Travis times out builds if they haven't had any output in 10 minutes. Our core specs currently take right around 10 minutes and they don't log any output. So it's pretty hit-or-miss as to whether our core specs complete or get timed out.

We should just pass through the output of the core specs to stdout so that Travis knows its still running.

joshaber:
Fixes #10837.


joshaber:
Opened https://github.com/atom/atom/issues/10840 about the flaky TextEditorPresenter spec.

PR #10841, opened 2016-02-16
Add the -a, --add CLI option

ArnaudRinquin:
This commit adds the `-a, --add` option to the CLI which allows users to add paths to the last focused window (similar to Sublime), like shown here:

add-option

This allows you to open many projects in the same Atom window (but not necessarily at the same time) without the hassle of clicking through the menus (right click > Add Project Folder > find my way to the folder...) but doing it from the CLI instead.

This comes with:
- a single additional test, I'm not sure if adding more is necessary
- documented CLI usage (automated)

I'm not sure if that feature needs to be documented elsewhere, if it does I'd be happy to do it.

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -897,8 +897,8 @@ class AtomEnvironment extends Model
   openLocations: (locations) ->
     needsProjectPaths = @project?.getPaths().length is 0
 
-    for {pathToOpen, initialLine, initialColumn} in locations
-      if pathToOpen? and needsProjectPaths
+    for {pathToOpen, initialLine, initialColumn, forceAddToWindow} in locations
+      if pathToOpen? and (needsProjectPaths or forceAddToWindow)
         if fs.existsSync(pathToOpen)
           @project.addPath(pathToOpen)
         else if fs.existsSync(path.dirname(pathToOpen))
--- a/spec/atom-environment-spec.coffee
+++ b/spec/atom-environment-spec.coffee
@@ -275,6 +275,14 @@ describe "AtomEnvironment", ->
         atom.openLocations([{pathToOpen}])
         expect(atom.project.getPaths()[0]).toBe __dirname
 
+      describe "then a second path is opened with forceAddToWindow", ->
+        it "adds the second path to the project's paths", ->
+          firstPathToOpen = __dirname
+          secondPathToOpen = path.resolve(__dirname, './fixtures')
+          atom.openLocations([{pathToOpen: firstPathToOpen}])
+          atom.openLocations([{pathToOpen: secondPathToOpen, forceAddToWindow: true}])
+          expect(atom.project.getPaths()).toEqual([firstPathToOpen, secondPathToOpen])
+
     describe "when the opened path does not exist but its parent directory does", ->
       it "adds the parent directory to the project paths", ->
         pathToOpen = path.join(__dirname, 'this-path-does-not-exist.txt')
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -123,6 +123,34 @@ describe "Starting Atom", ->
           .waitForPaneItemCount(0, 1000)
           .treeViewRootDirectories()
           .then ({value}) -> expect(value).toEqual([otherTempDirPath])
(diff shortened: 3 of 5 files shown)


src/atom-environment.coffee
```diff
@@ -891,10 +891,8 @@ class AtomEnvironment extends Model
     @commands.dispatch(@contextMenu.activeElement, command, args)
 
   openLocations: (locations) ->
-    needsProjectPaths = @project?.getPaths().length is 0
-
     for {pathToOpen, initialLine, initialColumn} in locations
-      if pathToOpen? and needsProjectPaths
+      if pathToOpen?
```
maxbrunsfeld: I think the reason that this guard was here is for this case:

```sh
$ atom directory-1 
# opens directory-1 in a new window

$ atom directory-2/file-2
# opens file-2 in the same window, but *does not* add directory-2 to the project
```

Does that still work? In other words, when I open a single file from the command line, does it still add it to the most recently focused Atom window, but *not* add the parent directory to the project?

ArnaudRinquin:
@maxbrunsfeld I just pushed an update that passes all tests, at least on my mac, including the use case you pointed.

It's also rebased on top of latest `master`

thedaniel:
I found myself wanting this feature today. :+1: from me.

iolsen:
:+1:

lee-dohm:
@maxbrunsfeld Does it still need changes? Or is it ready to merge?

maxbrunsfeld:
@ArnaudRinquin are you able to run the integration tests locally? I think this behavior would be best tested through an integration test. You could even just add a few lines to [this one](https://github.com/atom/atom/blob/fa08fba60e9cfd0c7acc82baec149f52cae7cf0b/spec/integration/startup-spec.coffee#L101). These tests can be a bit tricky, but could you give it a try? I can help you if you run into problems.

ArnaudRinquin:
Like I said, yes, current integration tests run locally + I just commited [some more](https://github.com/atom/atom/commit/bb1c048e12d5a3930b4c536d3a3ac453673ac4f2#diff-a5dc259060c4c3c53c128b8bd0f02d23R126).

maxbrunsfeld:
>  I just commited some more.

Nice!

maxbrunsfeld:
:clap: Really solid work @ArnaudRinquin!

maxbrunsfeld:
This will land in Atom v1.7.0 and v1.7.0-beta0.

ArnaudRinquin:
Great! Thank you all for the support!

asbjornu:
This is awesome, but can this be made the default behavior so I don't have to write the `-a` argument? And does this add support for drag-dropping folders into Atom on OS X in any way?

thomasjo:
> This is awesome, but can this be made the default behavior so I don't have to write the -a argument?

No, since that would break the normal usage pattern of opening multiple projects/windows. Essentially we'd go from a sensible workflow to something extremely cumbersome.

If you know you want to open multiple folders as part of a single project, you can simply do something like

```shell
~ atom folder1 folder2
```

In summary, adding new folders to an existing project, from a command line, is an auxiliary action and should remain as such.

cybertk:
+1

asbjornu:
@thomasjo Although I disagree with your definition of "normal usage pattern", I agree. I don't really care whether this is the default, what I meant to write was that I'd like to be able to **configure** Atom to behave this way by default **on my machine**.

thomasjo:
> [..] what I meant to write was that I'd like to be able to configure Atom to behave this way by default on my machine.

@asbjornu Just create an alias? `alias atom="atom --add"`

thedaniel:
^ I was just in the middle of typing this suggestion - it's a great way to configure atom to behave this way on your machine without adding any more complexity to the editor itself.

asbjornu:
@thomasjo That would have helped if I **only** opened Atom from the command line. I almost **never** do.

shane0:
just found this feature, I'll be using it many times a day great for automation, thanks!

I'm wondering if you can open multiple projects or if there's plans to implement that?

mnquintana:
@shane0 If you mean can you supply `atom -a` with multiple paths to open, then you can!

```
atom -a . ./path-to-other-folder ./path/to/some/other/folder
```

PR #10843, opened 2016-02-16
Bump packages to use async git

joshaber:
:pray:


joshaber:
Looks like a legit failure from `fuzzy-finder`.

PR #10851, opened 2016-02-17
Registry for TextEditors

wvanlint:
Currently, only `TextEditors` that are items of `Panes` are observable.
From the model point-of-view, the `TextEditors` are linked to the workspace through the `PaneContainer` and `Panes`.

There are other `TextEditors` that are of interest though.
For example, the editor in `find-and-replace` might want autocomplete functionality (atom/autocomplete-plus#383).

With this change, any editor created through `workspace.buildTextEditor` can be observed, which seemed the cleanest way to observe `TextEditors` across a varied use in the model.

Note: the tests seem to succeed locally, let me know if they are failing due to flakiness or if I am doing something wrong.

### DIFF
--- a/spec/text-editor-registry-spec.coffee
+++ b/spec/text-editor-registry-spec.coffee
@@ -1,0 +1,38 @@
+TextEditorRegistry = require '../src/text-editor-registry'
+
+describe "TextEditorRegistry", ->
+  [registry, editor] = []
+
+  beforeEach ->
+    registry = new TextEditorRegistry
+
+  describe "when a TextEditor is added", ->
+    it "gets added to the list of registered editors", ->
+      editor = {}
+      registry.add(editor)
+      expect(registry.editors.size).toBe 1
+      expect(registry.editors.has(editor)).toBe(true)
+
+    it "returns a Disposable that can unregister the editor", ->
+      editor = {}
+      disposable = registry.add(editor)
+      expect(registry.editors.size).toBe 1
+      disposable.dispose()
+      expect(registry.editors.size).toBe 0
+
+  describe "when the registry is observed", ->
+    it "calls the callback for current and future editors until unsubscribed", ->
+      [editor1, editor2, editor3] = [{}, {}, {}]
+
+      registry.add(editor1)
+      subscription = registry.observe spy = jasmine.createSpy()
+      expect(spy.calls.length).toBe 1
+
+      registry.add(editor2)
+      expect(spy.calls.length).toBe 2
+      expect(spy.argsForCall[0][0]).toBe editor1
+      expect(spy.argsForCall[1][0]).toBe editor2
+
+      subscription.dispose()
+      registry.add(editor3)
+      expect(spy.calls.length).toBe 2
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -40,6 +40,7 @@ Project = require './project'
 TextEditor = require './text-editor'
 TextBuffer = require 'text-buffer'
 Gutter = require './gutter'
+TextEditorRegistry = require './text-editor-registry'
 
 WorkspaceElement = require './workspace-element'
 PanelContainerElement = require './panel-container-element'
@@ -110,6 +111,9 @@ class AtomEnvironment extends Model
 
   # Public: A {Workspace} instance
   workspace: null
+
+  # Public: A {TextEditorRegistry} instance
+  textEditors: null
 
   saveStateDebounceInterval: 1000
 
@@ -182,6 +186,8 @@ class AtomEnvironment extends Model
       notificationManager: @notifications, @applicationDelegate, @clipboard, viewRegistry: @views, assert: @assert.bind(this)
     })
     @themes.workspace = @workspace
+
+    @textEditors = new TextEditorRegistry
 
     @config.load()
 
--- a/src/text-editor-registry.coffee
+++ b/src/text-editor-registry.coffee
@@ -1,0 +1,33 @@
+{Emitter, Disposable} = require 'event-kit'
+
+# This global registry tracks registered `TextEditors`.
+#
+# Packages that provide extra functionality to `TextEditors`, such as
+# autocompletion, can observe this registry to find applicable editors.
+module.exports =
+class TextEditorRegistry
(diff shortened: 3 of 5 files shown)


nathansobo:
The core team discussed this today... We're concerned about changing the existing semantics of `observeTextEditors` because packages will end up running code for editors they didn't expect. We need to do something additive.

* Callers to `buildTextEditor` can pass a `register: true` option to indicate they want to "register" their text editor.
* Observers call `observeRegisteredTextEditors`, which is called with all current and future registered editors.
* When an editor is destroyed, it is unregistered.

We'd like it to be opt-in because we're concerned that many package authors might not be calling `destroy`, which would be our only mechanism for removing an editor from the set of registered editors. If they opted in, they'd know what they were signing up for. Obviously this still presents issues for editors that don't get registered, but we could accept pull requests changing the option in any internal packages people would want to extend. Other package authors will need to work it out amongst themselves.

nathansobo:
Okay, after sleeping on this I have more thoughts on how we should approach this. I don't think we should provide a global API in Atom for this sort of thing. Instead, I think specific packages should expose services that allow specific editors to be observed.

The reason for this is as follows: Say that an autocomplete package wants to extend the find and replace editors. Isn't the fact that they're find-and-replace editors important to the autocomplete user experience? For example, in the find field, autocomplete could provide assistance building regular expressions. Whereas for the filter field, maybe it wants to provide path completion based on your project. Context really matters. For this reason, I think it's better for this relationship to be more custom and high-touch between packages than a global registry would allow. I'd like to encourage a PR to find-and-replace to add the requisite service for the functionality you'd like to build.

wvanlint:
Thanks for all the comments!

I totally agree that this change should be additive. Note that I didn't actually change `observeTextEditors` in this PR, I only added some methods. I just wanted to check if you think that the naming of the new methods isn't confusable with the existing methods. :)
The registration definitely makes sense to ensure correct behavior if there are a lot of `TextEditors` that don't call `destroy`.

For your second comment, it is true that for a specific use case, a specific `observe` method would be better. However, for example, the `autocomplete-plus` framework observes `TextEditors` in general as it is a general framework. Making the observation specific would inhibit an autocomplete package for find-and-replace from leveraging the power of the `autcomplete-plus` framework.

I fully agree that context matters though! See atom/autocomplete-plus#676 to see how I'm planning to handle this.

nathansobo:
Wow, I'm really sorry. I *swear* I submitted a comment on this PR and it appears to be gone. That's really frustrating.

nathansobo:
Okay, retyping my comment because I don't know where it went. After talking with @wvanlint and @maxbrunsfeld, we agree we should move forward with a global API after all with the following rough plan:

* Assign a `TextEditorRegistry` to `atom.editors`
* Implement `atom.editors.add(editor)`. It takes and editor and returns a disposable. This notifies future and current observers.
* Implement `atom.editors.observe(fn)`. It takes a callback and returns a disposable. It calls the callback for all current and future editors.

It turns out we're actually leaking editors that don't get destroyed anyway, as discussed in https://github.com/atom/atom/issues/10979, but I don't want to add more fuel to that fire. It still makes sense to opt in to being stored in the set of observed editors so you get a disposable you know you have to deal with. It will be easy to spot PRs where people are registered editors and not dealing with the disposable, whereas tracking down `destroy` calls sounds much harder.

As for focus tracking, you can just obtain the editor's DOM element and bind the `focus` and `blur` events directly. That's much more a product of the standard DOM behavior and we don't think it needs special treatment in the Atom API.

Thanks for taking the initiative here @wvanlint and your patience with my slow response. I had a comment written an hour after we talked and apparently my :dog: ate my :memo:.

wvanlint:
Thanks for the help with the design, that sounds great!
I adjusted the PR, let me know what you think. :)

src/editor-registry.coffee
```diff
@@ -0,0 +1,35 @@
...
+  # * `editor` The editor to register.
+  #
+  # Returns a {Disposable} on which `.dispose()` can be called to remove the
+  # added editor. To avoid any memory leaks this should be called when the
+  # editor is destroyed.
+  add: (editor) ->
+    @editors.push(editor)
+    @emitter.emit 'did-add-editor', editor
+    new Disposable =>
+      @editors = @editors.filter (e) -> e isnt editor
```
nathansobo: I think this might be better implemented as a `Set` so that removal is O(1). You can still use `forEach` to iterate it when new observers are added. Any objections?

spec/editor-registry-spec.coffee
```diff
@@ -0,0 +1,38 @@
+EditorRegistry = require '../src/editor-registry'
+
+describe "EditorRegistry", ->
```
nathansobo: Thanks for writing these nice tests!

src/editor-registry.coffee
```diff
@@ -0,0 +1,35 @@
+{Emitter, Disposable} = require 'event-kit'
+
+# This global registry tracks registered `TextEditors`.
+#
+# Packages that provide extra functionality to `TextEditors`, such as
+# autocompletion, can observe this registry to find applicable editors.
+module.exports =
+class EditorRegistry
```
nathansobo: Small nitpick, but I'd really like this to be called `TextEditorRegistry` for consistency with the `TextEditor` class. I know I proposed calling the global `atom.editors`, but I wonder if we *should* actually call it `atom.textEditors` for more precision and consistency.

I originally wanted to keep it one word so it looked better, but now that I think about it, it would be the first unqualified use of the word "editor" in our API. Many different kinds of editors could potentially exist within Atom, so maybe we should be clear that we're talking about text editors. Would you be down to change this? Thanks.

nathansobo:
This is looking really nice. I made some minor suggestions. Once we reach consensus on my comments we are good to go.

wvanlint:
Fully agree with all your suggestions, I applied the changes.
Thanks for the review!

nathansobo:
Build appears to be a false negative. I'm moving forward with a merge and will follow up with some documentation adjustments. Thanks a ton @wvanlint for your careful collaboration and raising this overall issue. Ping me on PRs to packages you want to opt in with such as `find-and-replace`.

Issue #9879, opened 2015-11-30
TextEditor.getPath only providing realpath


dirk-thomas:
Currently `TextEditor.getPath` returns the realpath of the file. When a project directory contains symlinks somewhere under it it becomes impossible to decide which project root a specific file belongs to. Please see noseglid/atom-build#245 for a longer explanation of the use case as well as an example folder hierarchy.

The only way to make a correlation between a TextEditor and a project root requires the TextEditor to provide the non-real path of the file.

Getting the realpath from the non-real path is trivial (`fs.realpath())`) but the other way around is simply not possible.

thedaniel:
@atom/feedback thoughts on this? I'm not familiar with a foolproof way to associate a TextEditor to a project path - I can see the benefit of adding a getter for something like `nonRealPath` / `literalPath` (naming is hard), or perhaps even better, a method that returns the project root, if any, that the file in the editor belongs to. My gut tells me that a lot of people are inferring project root from the file path.. i certainly have in packages before.

lee-dohm:
If we could get the project root and project-relative path (both I think would be useful in other contexts too) from a `TextEditor`, that should fix this issue I think?

But this is going to be a really, really deep rabbit hole, I feel. The problem is that there can be an infinite number of paths that can reach a file. Should the `TextEditor` instance have to know about all of them? What if one of the `n` paths changes after the `TextEditor` is created? Say one of the non-real (imaginary?) paths to the file in question is `a/foo/c.txt` and gets renamed to `a/bar/c.txt`. Do we have to have file watchers on everything to manage this?

maxbrunsfeld:
I agree with @dirk-thomas that we should simply *not* compute the real path, and leave that to the caller. Does anyone know of circumstances where that would break things in a way that wasn't trivially fixable?

lee-dohm:
> Does anyone know of circumstances where that would break things in a way that wasn't trivially fixable?

I'm pretty sure it was this way to begin with (because it was easier), but it changed for some reason ¯\_(ツ)_/¯ I'll see if I can backtrace it.

lee-dohm:
So I devised a test just to see how consistent this is:

1. Created a new repository `lee-dohm/test-atom-9879`
1. Created a directory `real-dir` inside it
1. Created a symlink to `real-dir` named `fake-dir`
1. Created a file inside `real-dir` named `test.md`

[...]

PR #10856, opened 2016-02-17
[WIP] add project.resolveNonRealpath to return non-realpath compared to project.resolvePath

dirk-thomas:
This PR is trying to address #9879.

~~The first commit simply changes the semantic of `project.resolvePath()` to return a non-realpath. Since this might break the callers expectation I will try to keep the semantic of existing API on-top as outlined in https://github.com/atom/atom/issues/9879#issuecomment-162253183 as necessary.~~ See newer comment for current behavior.

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -106,7 +106,7 @@ describe "Project", ->
 
       runs ->
         editor.saveAs(tempFile)
-        expect(atom.project.getPaths()[0]).toBe path.dirname(tempFile)
+        expect(atom.project.getPaths()[0]).toBe path.dirname(atom.project.normalizeRealpath(tempFile))
 
   describe "before and after saving a buffer", ->
     [buffer] = []
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -5,6 +5,16 @@ fs = require 'fs-plus'
 {Emitter, Disposable} = require 'event-kit'
 TextBuffer = require 'text-buffer'
 
+TextBuffer.getUnresolvedPath = TextBuffer.getPath
+TextBuffer.getPath = ->
+  fullPath = @getUnresolvedPath()
+  return unless fullPath
+  try
+    fullPath = fs.realpathSync(fullPath)
+  catch e
+    # nothing
+  path.normalize(fullPath)
+
 DefaultDirectoryProvider = require './default-directory-provider'
 Model = require './model'
 GitRepositoryProvider = require './git-repository-provider'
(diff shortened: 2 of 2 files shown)


dirk-thomas:
The current patch changes each `TextBuffer` to store the non-realpath (instead of the normalized realpath). This allows providing new methods to access the non-realpath:

* Project.resolveNonRealpath
* Text.getNonRealpath

The existing methods of `Project` and `TextBuffer` have been updated to continue having the same semantic as before. They use the helper function `Project.normalizeRealpath`.

Currently this PR monkey patches the TextBuffer in the `project.coffee`. Once the path forward is becoming clearer this should be moved to the corresponding package.

I am not yet sure why the one test needed to be updated in the second commit (in order to pass).

But I would like to get feedback on the general approach. Any comments are welcome.

maxbrunsfeld:
@dirk-thomas Could you explain what you'd like to do with this API when and if we add it? Are you planning on using it in packages of your own, or submitting later PRs to make existing packages use this API?

dirk-thomas:
Please see the referenced issue #9879 for some more information. The immediate use case would be the Atom `build` package. Currently it has no way to work with projects which contain symlinked paths. The problem is described in a ticket referenced from the issue: noseglid/atom-build#245

I plan to provide patches to the `build` package (and their providers) so that it supports symlinked files within projects correctly. Without support for this in the Atom core API I don't think this is possible.

maxbrunsfeld:
Ah ok, thanks for the explanation. I would definitely be ok with adding an API for this to `TextBuffer`. I think I might prefer the name `getOriginalPath`, rather than `getNonRealPath`. The idea of a non-real path sounds kind of confusing.

dirk-thomas:
For the naming imo the best would be to have `getPath` and `getRealPath` ("realpath" because that is what the two methods differentiates http://man7.org/linux/man-pages/man3/realpath.3.html). But obviously we don't want to change the semantic of the existing method. Therefore I called the new one `getNonRealpath`.

If you prefer `getOriginalPath` I can for sure update the PR. I just think that "realpath" is the right term here since its meaning is clearly defined by the same named low level API.

Once we agree on the current state of the patch (only changing `Project`) the next step is for me to also add the new API to `TextBuffer`.

maxbrunsfeld:
> I just think that "realpath" is the right term here since its meaning is clearly defined by the same named low level API.

Good point. What about `getUnresolvedPath()`? The linux [man page for `realpath(1)`](http://man7.org/linux/man-pages/man1/realpath.1.html) refers to the output of `realpath()` as the "the resolved absolute file name".

dirk-thomas:
I was able to rebase to master and remove some code from the patch since now `fs-plus` 2.9.2 is being used. I also changed `TextBuffer.getNonRealpath` to `TextBuffer.getUnresolvedPath` following your naming recommendation.

I am just not sure what a good name for the new method on `Project` is. Before I named it `resolveNonRealpath`. Based on your comments I think avoiding "resolve" is better. Therefore I updated the PR to call it `Project.absolutePath` but any feedback on this is welcome.

dirk-thomas:
@maxbrunsfeld Can you please take another look at the updated PR when you have some time. Thank you.

maxbrunsfeld:
I think `absolutePath` is a good name for the new method. So I think the next steps for this PR are:

1. Submit a text-buffer PR adding `getUnresolvedPath`
2. Add specs for the new `Project.absolutePath` method.

dirk-thomas:
Ok, thank you for the feedback. I will get on those items as soon as I find a spare minute.

maxbrunsfeld:
This problem was addressed in https://github.com/atom/atom/pull/13545.

dirk-thomas:
Great. Thanks!

PR #10858, opened 2016-02-17
Update grunt-electron-installer to latest version

damieng:
This gets us the latest version of Squirrel updater with many Windows fixes.


kevinsawicki:
:+1:

PR #10861, opened 2016-02-17
Compute line foldability lazily

as-cii:
Understanding whether a row can be folded or not requires to scan surrounding lines and it's relatively slow. As you might imagine, doing so for every changed line after a multi-cursor edit can become quite painful as the amount of edited lines increases. For this reason, with this PR we now compute foldability on-demand (lazily) only for on-screen lines.
#### Before

screen shot 2016-02-17 at 15 24 01

_(Total: `~ 482ms`, benchmark performed by editing 1500 lines, writing 5 characters)_
#### After

screen shot 2016-02-17 at 15 22 22

_(Total: `0 ms` in `handleBufferChange`, and the impact on the rendering path is negligible thanks to 2a35d19 :racehorse: :racehorse: :racehorse:)_

There's one caveat I illustrated in 90c75fd which needs to be discussed, but I feel like it _might_ be addressed separately and was not a problem introduced with these changes.

/cc: @atom/core


maxbrunsfeld:
:clap:

nathansobo:
Nice work here @as-cii. :bullettrain_front: keeps on rolling.

as-cii:

I believe specs are failing because of network issues, so I went ahead and merged this anyways. :ship:

Issue #7099, opened 2015-06-04
Dropping multiple files into an empty atom window makes each file a root "folder"


nshtg:
OS: Windows 7 x64, freshly build `atom-v0.206.0-e1514cf` in safe mode.

I made a gif to show:

![animation 0](https://cloud.githubusercontent.com/assets/5219415/7989076/6428ce62-0ae7-11e5-8d60-c76487570229.gif)

mnquintana:
I can't reproduce this with Atom 0.206.0-e1514cf and OS X 10.10.3 – maybe this is a Windows only thing? I'll try this with my VM to see.

maschs:
I can confirm it's not happening on OSX. Only windows. Don't know about linux.

MadMac:
Reproduced on Windows 8.1 32-bit. Atom version 1.0.0.

50Wliu:
According to @ben3eeE this is no longer reproducible on Atom 1.9.0, so I'm going to close this.

PR #10862, opened 2016-02-17
Fix for Default Directory Provider incorrectly identifying file paths as URLs on Windows

PaulAik:
Atom's `default-directory-provider` incorrectly identified Windows file paths (e.g. c:\users\paul\text.txt) as URL's. This has a knock-on effect for features such as the Tree View, which shows individually opened files as though they are directories (as opposed to the correct functionality, whereby Tree View shows the file's parent directory and highlights the opened file).

This was because the module was checking for the presence of a protocol in the path passed to it in order to determine whether it was a local file path or a URL (such as atom://config/something). Unfortunately, on Windows the drive letter (e.g. `c:`) was identified as a protocol.

The fix put in place replaces the check for a protocol with one that checks for a hostname. In valid URLs this should come back as a non-empty string, but on Windows it'll come back empty meaning the module correctly identifies the string as a file path.

Tested on Mac and Windows.

Linked tree view issue is [here](https://github.com/atom/tree-view/issues/471)


PaulAik:
Hmm, Travis CI build shouldn't be broken now!

50Wliu:
@PaulAik Any chance you can add additional tests for this? :sparkles:

PaulAik:
@50Wliu I had a go at creating some new tests, but had issues because you need the right path format for each platform (e.g. can't run a test for 'c:\somepath\test.txt on Mac/Linux).

Looking at the existing tests, _"creates a Directory for its parent dir when passed a file"_ and _"creates a Directory with a path as a uri when passed a uri"_ should cover it when tests are run on Windows.

50Wliu:
You can make it so the test only runs on Windows (see below).  My question then is why the specs consistently pass on the `atom-windows` CI box.

`it 'Doesn't treat filenames as file protocols on #win32', ->`

PaulAik:
Yeah it's strange. Can you show me where to find the Windows CI build test results? 

Looked at the AppVeyor tests tab, but it just says _"The build job has not produced any test results"_.

50Wliu:
When Github employees push something, their commits trigger additional private CI builds.  Currently only the private Windows CI server runs core specs, Appveyor just makes sure everything builds correctly right now.

So unfortunately I can't direct you to the spec results :(.

PaulAik:
So I replaced the `default-directory-provider.coffee` file with the one from atom master on Windows and re-ran the tests. The following tests fail:

- "normalizes its input before creating a Directory for it" 
- "creates a Directory for its parent dir when passed a file"

When I re-introduce the change in this pull request, all tests (for that module) pass.

No idea why what I'm seeing differs from the internal Github CI server to be honest. Based on my (limited!) understanding, the current code would fail these tests.

PaulAik:
Hi guys - let me know if there's anything I can do to push this one along :+1:

J-qak:
Hello, I've been reading these threads for some time since I'm frustrated by this bug every day, but I don't see any progress on pullling this fix. May I ask when we will see this issue finally fixed?

nixel2007:
>  I've been reading these threads for some time since I'm frustrated by this bug every day

+1. Please, review

damieng:
I think the problem here is there are no specs to cover the test either before or after this change. @PaulAik is it possible you could add a positive and negative case?

PaulAik:
@damieng there's actually a full suite of [tests](https://github.com/atom/atom/blob/master/spec/default-directory-provider-spec.coffee) for that module, so I think the updates should already be covered.

What I was confused about was why the Windows CI server was passing these tests before; they should have failed from what I could see (and what happened on my Windows test run). 

However, the server is (or was) private so I can't see the official test results unfortunately!

damieng:
When I run the specs locally on Windows today without this PR they pass... Hence why I think we need a failure case to prove before/after the PR works as intended.

50Wliu:
I'm seeing the same failures as @PaulAik when running specs for `DefaultDirectoryProvider`.
```
DefaultDirectoryProvider
  .directoryForURISync(uri)
    it normalizes its input before creating a Directory for it
      Expected 'C:\Users\username\AppData\Local\Temp\d-116216-11320-p8gps7\..\d-116216-11320-p8gps7' to equal 'C:\Users\username\AppData\Local\Temp\d-116216-11320-p8gps7'. (file:\C:\Users\username\Documents\GitHub\atom\spec\default-directory-provider-spec.coffee:23:35)
    it creates a Directory for its parent dir when passed a file
      Expected 'C:\Users\username\AppData\Local\Temp\d-116216-11320-1am9lm7\example.txt' to equal 'C:\Users\username\AppData\Local\Temp\d-116216-11320-1am9lm7'. (file:\C:\Users\username\Documents\GitHub\atom\spec\default-directory-provider-spec.coffee:32:35)


Finished in 6.376 seconds
5 tests, 7 assertions, 2 failures, 2056 skipped
```

damieng:
Can confirm this fixes a bunch of spec failures on Windows. Should make sure it does not regress on Mac.

src/default-directory-provider.coffee
```diff
@@ -26,7 +26,7 @@ class DefaultDirectoryProvider
 
     # TODO: Stop normalizing the path in pathwatcher's Directory.
     directory = new Directory(directoryPath)
-    if protocol?
+    if host
```
damieng: Shouldn't this be host?  i.e. only if not a URL given it is treat it as a plain directory.

PaulAik: Host is the new test. 

It's the same approach as used previously, only using a different part of the URL that is more reliably detected on Windows (but gives the same result as before on Mac, etc)

Ben3eeE:
Does this close https://github.com/atom/atom/issues/7099?

damieng:
Would have to test - it's not obvious why it would.

Ben3eeE:
@damieng I did test and I can reproduce https://github.com/atom/atom/issues/7099 in Atom 1.7.4 but not in 1.9.0-dev, this is the only change I could find related to files showing up as folders.

PR #10870, opened 2016-02-17
Add Issue template and extra version info

lee-dohm:
/cc @atom/feedback for more :eyes:

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -25,6 +25,7 @@
 3. [and so on...]
 
 **Expected:** [What you expected to happen]
+
 **Actual:** [What actually happened]
 
 ### Versions
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -19,9 +19,9 @@ For more information on how to write a good [bug report](https://github.com/atom
 2. [Second Step]
 3. [and so on...]
 
-**Expected:** [What you expected to happen]
+**Expected behavior:** [What you expected to happen]
 
-**Actual:** [What actually happened]
+**Actual behavior:** [What actually happened]
 
 ### Versions
 
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -5,13 +5,13 @@
     * [ ] Did you check the [debugging guide](https://atom.io/docs/latest/hacking-atom-debugging)?
     * [ ] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
     * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
-    * [ ] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+user%3Aatom) to see if it is already reported?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if it is already reported?
     * For more information on how to write a good bug report [look at the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report).
 * [ ] Enhancement
     * [ ] Are you running the [latest version of Atom](https://atom.io/docs/latest/hacking-atom-debugging#update-to-the-latest-version)?
     * [ ] Did you check to see if [a package](https://atom.io/packages) already offers the feature?
     * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
-    * [ ] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+user%3Aatom) to see if it is already requested?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if it is already requested?
     * For more recommendations on how to write a good enhancement suggestion [see the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion).
 
 ### Description
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,41 @@
+### Flow
+
+* [ ] Bug
+    * [ ] Can you reproduce the problem in [safe mode](https://atom.io/docs/latest/hacking-atom-debugging#check-if-the-problem-shows-up-in-safe-mode)?
+    * [ ] Did you check the [debugging guide](https://atom.io/docs/latest/hacking-atom-debugging)?
+    * [ ] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
+    * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+user%3Aatom) to see if it is already reported?
+    * For more information on how to write a good bug report [look at the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report).
+* [ ] Enhancement
+    * [ ] Are you running the [latest version of Atom](https://atom.io/docs/latest/hacking-atom-debugging#update-to-the-latest-version)?
+    * [ ] Did you check to see if [a package](https://atom.io/packages) already offers the feature?
+    * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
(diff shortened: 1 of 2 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,41 @@
+### Flow
+
+* [ ] Bug
+    * [ ] Can you reproduce the problem in [safe mode](https://atom.io/docs/latest/hacking-atom-debugging#check-if-the-problem-shows-up-in-safe-mode)?
+    * [ ] Did you check the [debugging guide](https://atom.io/docs/latest/hacking-atom-debugging)?
+    * [ ] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
+    * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+user%3Aatom) to see if it is already reported?
```
50Wliu: This should include `-repo:atom/electron`

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,41 @@
...
+
+[Description of the bug or feature]
+
+### Repro Steps
+
+1. [First Step]
+2. [Second Step]
+3. [and so on...]
+
+**Expected:** [What you expected to happen]
```
50Wliu: I think an extra line break is needed after this line, or else the Expected and Actual will display on the same line.

50Wliu:
Also, the whole file seems to be indented using 4 spaces and not 2...not sure if that was intentional or not.

lee-dohm:
I indent Markdown using 4 spaces habitually because some Markdown parsers require that. (It's one of the places where original Markdown isn't clear.) I can change it though.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,42 @@
...
+    * [ ] Did you check to see if [a package](https://atom.io/packages) already offers the feature?
+    * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if it is already requested?
+    * For more recommendations on how to write a good enhancement suggestion [see the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion).
+
+### Description
+
+[Description of the bug or feature]
+
+### Repro Steps
```
joefitzgerald: Non-technical users may not know what "Repro" means. "Steps To Reproduce"?

src/browser/main.coffee
```diff
@@ -83,6 +83,12 @@ setupCompileCache = ->
   compileCache = require('../compile-cache')
   compileCache.setAtomHomeDirectory(process.env.ATOM_HOME)
 
+fullVersion = ->
+  process.stdout.write("Atom    : #{app.getVersion()}\n")
+  process.stdout.write("Electron: #{process.versions.electron}\n")
+  process.stdout.write("Chrome  : #{process.versions.chrome}\n")
+  process.stdout.write("Node    : #{process.versions.node}\n")
```
maxbrunsfeld: How about this:

```coffee
process.stdout.write """
  Atom    : #{app.getVersion()}
  Electron: #{process.versions.electron}
  Chrome  : #{process.versions.chrome}
  Node    : #{process.versions.node}
 """

Comment (lee-dohm) [inline src/browser/main.coffee]: Good point ... testing now.

File: ISSUE_TEMPLATE.md [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if it is already reported?
+    * For more information on how to write a good bug report [look at the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report).
+* [ ] Enhancement
+    * [ ] Are you running the [latest version of Atom](https://atom.io/docs/latest/hacking-atom-debugging#update-to-the-latest-version)?
+    * [ ] Did you check to see if [a package](https://atom.io/packages) already offers the feature?
+    * [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
+    * [ ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if it is already requested?
+    * For more recommendations on how to write a good enhancement suggestion [see the CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion).
```
maxbrunsfeld: I think that this checklist is a bit too verbose to actually include in the issue body. What if we just had a single checkbox with a link to a section of `CONTRIBUTING.md`?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,42 @@
...
+You can get this information from executing `atom --version` and `apm --version` at the command line:
+
+* **Atom:**     x.y.z
+* **Electron:** x.y.z
+* **OS:**       OS x.y.z
+* **APM**
+    * apm    x.y.z
+    * npm    x.y.z
+    * python x.y.z
+    * git    x.y.z
```
maxbrunsfeld: I'd say since the formatting is controlled by `atom` and `apm`, we can omit this example template for the versions.

ISSUE_TEMPLATE.md
lee-dohm: The idea for this is that it puts the common debugging steps that the triage team very often has to ask people about in follow up questions right into the issue up front. It is verbose ... but so is the back and forth on an Issue in a bunch of comments before we get at the real problem. I know that some people will skip by it ... or just check all the boxes without trying things, but I think many more will continue to not know the common things to help themselves if all we have is a single checkbox that points to a `CONTRIBUTING.md` that they already ignore the, admittedly, less prominent link to.

src/browser/main.coffee
```diff
@@ -83,6 +83,15 @@ setupCompileCache = ->
   compileCache = require('../compile-cache')
   compileCache.setAtomHomeDirectory(process.env.ATOM_HOME)
 
+fullVersion = ->
```
benogle: This should probably be named something like `writeFullVersion`

ISSUE_TEMPLATE.md
benogle: I agree that this is pretty verbose. Is there a way to have some of the most common things, but not everything? Enhancements are less bothersome than 'omg broken' bug reports; maybe we just have something that applies to both?

src/browser/main.coffee
```diff
@@ -118,7 +127,7 @@ parseCommandLine = ->
   options.boolean('portable').describe('portable', 'Set portable mode. Copies the ~/.atom folder to be a sibling of the installed Atom location if a .atom folder is not already there.')
   options.alias('t', 'test').boolean('t').describe('t', 'Run the specified specs and exit with error code on failures.')
   options.string('timeout').describe('timeout', 'When in test mode, waits until the specified time (in minutes) and kills the process (exit code: 130).')
-  options.alias('v', 'version').boolean('v').describe('v', 'Print the version.')
+  options.alias('v', 'version').boolean('v').describe('v', 'Print the version information.')
```
mnquintana: I'm not totally sure, but I'm wondering if we'd want to keep the `--version` flag as-is, but then be able to modify it to display the full version info with a `--verbose` flag too. Maybe that's too convoluted, though.

lee-dohm: I thought about that at first. I want to make it drop-dead simple though. I also want Atom to always display the same version information ... whether at the command line or in the About view. This way if Atom won't launch (or only shows the white screen), you can still get to the version info from the command line.

benogle: I wonder if people are using `atom -v` in scripts or otherwise that will break stuff. Probably a long shot, but something to consider?

lee-dohm: Actually, that's a good point. A couple of my support snippets use `atom -v` :laughing:

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+### Flow
```
maxbrunsfeld: I think this section should be called 'Prerequisites' or 'First Steps', or something.

src/browser/main.coffee
maxbrunsfeld: :+1: Good call @mnquintana. Maybe `--versions / -V`?

maxbrunsfeld: Or else we should just change the scripts. I'm ok w/ making the `version` more geared toward humans than scripts.

lee-dohm: I'm cool with making `--version` more geared towards humans. I'll change my script. We already have colors and stuff in the apm commands that have to be stripped out or given another parameter, i.e. `apm list --bare`.

benogle: :+1:

ISSUE_TEMPLATE.md
nathansobo: Expected *behavior*?

nathansobo:
:ok_hand:

benogle:
:surfer: Will be super interesting to see how people interact with this!

lee-dohm:
Yes, it will! I'm all settled in for a long day :laughing:

PR #10872, opened 2016-02-17
Terminate pending state for opened file if pending option is false

kuychaco:
@joshaber pointed out that opening a file from fuzzy-finder currently does not make a pending pane item permanent. This PR makes it such that if you open a file without the `pending` option and the file is currently opened in pending state, the pending state will be terminated. 

TODO: Merge PRs for find-and-replace and tree-view once this is on stable. I simplified code in these packages by opening items without `pending` option rather than explicitly invoking `terminatePendingState`

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -585,6 +585,22 @@ describe "Workspace", ->
           open = -> workspace.open('file1', workspace.getActivePane())
           expect(open).toThrow()
 
+    describe "when the file is already open in pending state", ->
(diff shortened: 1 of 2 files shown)


50Wliu:
@kuychaco Does this address atom/fuzzy-finder#173?  Seems like it's asking for the opposite behavior.

kuychaco:
@50Wliu thanks for the heads up. Will discuss with the team about opening items in pending state from fuzzy-finder

joshaber:
:metal:

PR #10874, opened 2016-02-18
Avoid Windows 260-character path limits in script\clean

damieng:
Kind of experimental but currently clean on Windows fails most of the time because of the path lengths.

### DIFF
--- a/script/clean
+++ b/script/clean
@@ -4,15 +4,16 @@ var fs = require('fs');
 var path = require('path');
 var os = require('os');
 
-var removeCommand = process.platform === 'win32' ? 'rmdir /S /Q ' : 'rm -rf ';
+var isWindows = process.platform === 'win32';
+var removeCommand = isWindows ? 'rmdir /S /Q ' : 'rm -rf ';
 var productName = require('../package.json').productName;
 
 process.chdir(path.dirname(__dirname));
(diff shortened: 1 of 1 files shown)


iolsen:
:+1: I'm good with merging this.

joshaber:
Oh man this makes me happy :heart:

benogle:
:checkered_flag: Everytime a windows improvement gets merged, a developer gets their wings. I think that's how the saying goes.

script/clean
```diff
@@ -39,12 +40,32 @@ var run = function() {
...
+    }
   }
-
-  cp.safeExec(next, run);
+  else
+    cp.safeExec(next, run);
 };
 run();
+
+// Windows has a 260-char path limit for rmdir etc. Just recursively delete in Node.
```
paulcbetts: Doesn't `rimraf` handle this?

damieng: Somebody else told me that - it looks like it does but who wants to bring in a package to replace 11 lines of code only used in a clean script on Windows.

paulcbetts: The big reason to use rimraf is that it uses `graceful-fs` behind the scenes, so it automatically retries transient ACCESS_DENIEDs that Windows AV vendors love to throw

PR #10881, opened 2016-02-18
Prevent hangs on opening files with long lines by only tokenizing the first 2000 characters of a line

wiggzz:
Works with https://github.com/atom/first-mate/pull/67 to prevent hangs on opening files with extremely long lines. Should fix many of the issues in https://github.com/atom/atom/issues/979.


Zireael07:
Great!

wiggzz:
Looks like some downloads failed on the CI, someone may need to restart the builds.

wiggzz:
Just wondering if there is any progress on this?

This does fix a lot of hanging bugs for me, primarily ones that result from long lines. While there is already a provision in the code to prevent creation of more than 100 tokens, that check is not made in the inner loop of the oniguruma scanner, so if the line is really long and no tokens are found, scanning can take a really long time.

thomasjo:
This + atom/first-mate#67 makes sense to me, but where does the 8000 character limit come from? Any specific reason you chose that specific number and not something significantly lower?

/cc @atom/feedback

wiggzz:
Yea.... pretty much picked it at random. All the hangs I have seen have been when lines are on the order of 100k characters or more... I just guessed that oniguruma should be able to very quickly handle around 8000 chars, but I agree, would be nice to justify that... let me see if I can add a performance test somewhere for this.

wiggzz:
Ok I've looked into this a pretty good amount. I've put together a performance test that uses a snippet of an example file I have been playing with and tokenizes it using the basic html grammar. I suspect there is some kind of runaway matcher happening which is part of the problem, but it's compounded by the fact that the line length is so long, oniguruma spends a long time processing the match.

Anyway, according to the benchmarking, restricting line lengths to 1000 characters seems to keep the tokenizing under 30ms, which is really fast, and seems like a more reasonable default than 8000 characters, which tokenizes in around 300ms on my machine, so I have changed my pull request to restrict line lengths to 1000 characters.

If you want to run the benchmarking yourself, just run

```>coffee spec/benchmark/tokenize.coffee```

from the atom base directory. You can adjust the parameters within that file.

(For reference the buggy grammar is here: https://github.com/atom/language-html/blob/b6abe7923b85906c607e58ded4961b67f161ff38/grammars/html.cson#L217, which I determined by commenting that block, whereby the hang goes away. I didn't spend much time trying to debug that, but with this pull request, buggy grammars should not cause big time hangs any more)

wiggzz:
AppVeyor build timed out.

MaximSokolov:
This should be closed as #979 is solved

50Wliu:
Yeah.  Sorry for not acting on this @wiggzz!

PR #10885, opened 2016-02-18
Cache regexes in LanguageMode.prototype.getRegexForProperty

as-cii:
This adds a simple layer of caching into `LanguageMode.prototype.getRegexForProperty` to avoid allocating `OnigRegExp` objects when e.g. calling `autoDecreaseIndentForBufferRow` (which is called for each line when performing a multi-cursor edit).
#### Before

screen shot 2016-02-18 at 14 31 43

_(Total: `~ 293ms`, benchmark performed by editing 1500 lines, writing 5 characters)_
#### After

screen shot 2016-02-18 at 14 29 03

_(Total: `~ 135ms`, ~ 2x speedup :racehorse:)_

For the same edit, the method now takes `~ 135ms` to execute, but most of that time (`> 80ms`) is spent in getting the scope descriptor for a certain buffer position. That should get better with @nathansobo's improvements at the `DisplayLayer`, making `autoDecreaseIndentForBufferRow` a pretty cheap function call.

/cc: @atom/core

### DIFF
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -10,6 +10,7 @@ class LanguageMode
   # editor - The {TextEditor} to associate with
   constructor: (@editor, @config) ->
     {@buffer} = @editor
+    @regexesByPattern = {}
 
   destroy: ->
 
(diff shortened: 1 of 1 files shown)


nathansobo:
Looking good. Simple change. Is the `scopeDescriptorForPosition` call slow due to position translation? If not, my current work won't help. But we can still improve it.

as-cii:
> Is the scopeDescriptorForPosition call slow due to position translation? If not, my current work won't help. But we can still improve it.

Yes, I should have included that in the profiler but it's basically position translation and instantiating `TokenIterator`s. So, theoretically, there will be a big improvement with `DisplayLayer` which will make that function's cost negligible. :+1:

:ship:ping this! :boom:

benogle:
If there were a steamroller emoji, I'd use it right here.

PR #10886, opened 2016-02-18
Update Atom build directory in build instructions

daxlab:
The atom build directory has been changed. The new build directory is `out/Atom`.


50Wliu:
Thanks!

PR #10888, opened 2016-02-18
Load packages before deserializing state

maxbrunsfeld:
Fixes issue caught by @izuzak in https://github.com/atom/atom/pull/10764#issuecomment-183709726
Refs https://github.com/atom/atom/pull/10743

I took this as an opportunity to clean up some of the craziness. Thus, I have probably broken something.

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -41,13 +41,11 @@ class AtomWindow
     @handleEvents()
 
     loadSettings = _.extend({}, settings)
-    loadSettings.windowState ?= '{}'
     loadSettings.appVersion = app.getVersion()
     loadSettings.resourcePath = @resourcePath
     loadSettings.devMode ?= false
     loadSettings.safeMode ?= false
     loadSettings.atomHome = process.env.ATOM_HOME
-    loadSettings.firstLoad = true
     loadSettings.clearWindowState ?= false
 
     # Only send to the first non-spec window created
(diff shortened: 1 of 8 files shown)


maxbrunsfeld:
napkin

src/window-load-settings-helpers.coffee
```diff
@@ -5,15 +5,6 @@ windowLoadSettings = null
...
   windowLoadSettings ?= JSON.parse(window.decodeURIComponent(window.location.hash.substr(1)))
-  clone = _.deepClone(windowLoadSettings)
-
-  # The windowLoadSettings.windowState could be large, request it only when needed.
-  clone.__defineGetter__ 'windowState', ->
-    remote.getCurrentWindow().loadSettings.windowState
-  clone.__defineSetter__ 'windowState', (value) ->
-    remote.getCurrentWindow().loadSettings.windowState = value
-
-  clone
```
maxbrunsfeld: The window state used to be considered part of the 'load settings', even though it actually used a completely different IPC mechanism under the hood. AFAIK, the reason that it ever had anything to do with the load settings is that for windows w/ no project paths (and hence no key in our on-disk storage system), we needed a place to store the window state when refreshing the window.

There are now separate APIs on the `ApplicationDelegate` for reading and writing this in-process, in-memory state: `getTemporaryWindowState` and `setTemporaryWindowState`. They are async.

src/browser/atom-window.coffee
```diff
@@ -64,23 +62,18 @@ class AtomWindow
...
     loadSettings.initialPaths.sort()
 
-    @browserWindow.loadSettings = loadSettings
     @browserWindow.once 'window:loaded', =>
       @emit 'window:loaded'
       @loaded = true
 
     @setLoadSettings(loadSettings)
     @browserWindow.focusOnWebView() if @isSpec
+    @browserWindow.temporaryState = {windowDimensions} if windowDimensions?
```
maxbrunsfeld: Now, instead of making the window dimensions a top-level field on the load settings, we pre-populate the window's 'temporary state' (see below). Since this temporary window state is overwritten periodically and when unloading the window, we should avoid the unwanted cascade positioning.

src/window-event-handler.coffee
```diff
@@ -141,8 +141,8 @@ class WindowEventHandler
       @atomEnvironment.hide()
     @reloadRequested = false
 
-    @atomEnvironment.storeDefaultWindowDimensions()
     @atomEnvironment.storeWindowDimensions()
+    @atomEnvironment.saveState()
```
maxbrunsfeld: Now we fire off one last async save before unloading the window. @kuychaco had chosen not to bother with this initially, since we're already saving periodically, and it was unclear whether the async save would complete before the unload. I think it does work though, and it's important to ensure we store the latest window dimensions.

joshaber: Ah, hopefully solves https://github.com/atom/atom/issues/10648 again?

maxbrunsfeld: Ah yeah. https://github.com/atom/atom/pull/10605 must have broken that in the case where you reloaded the window without ever interacting with it and going idle afterwards. This should fix that.

maxbrunsfeld:
The mac build failure is another flaky test.

PR #10895, opened 2016-02-18
Fix some spec issues

maxbrunsfeld:


src/browser/main.coffee
```diff
@@ -33,6 +34,8 @@ start = ->
 
   if args.userDataDir?
     app.setPath('userData', args.userDataDir)
+  else if args.test
+    app.setPath('userData', temp.mkdirSync('atom-test-data'))
```
maxbrunsfeld: /cc @kuychaco You were talking about adding this, but I found I needed to do this for a spec I added.

PR #10926, opened 2016-02-22
Fix case-preserving path relativization

joshaber:
Fixes https://github.com/atom/git-diff/issues/86.


joshaber:
After this is merged I'll back port it to `beta`.

Issue #10852, opened 2016-02-17
Command-tilde no longer cycles through open windows


joshaber:
On `master` (41161dd462a1be9073752a34b4809612f933da6e).

This is a regression, presumably introduced by https://github.com/atom/atom/pull/9627.

/cc @maxbrunsfeld

maxbrunsfeld:
@joshaber I can't reproduce this; command-tilde works for me. I *was* seeing some strange behavior when I first built Atom after the upgrade. Atom was actually failing to show up in the app switcher and in the dock. This problem persisted for me until I rebooted. I thought maybe something about the electron upgrade put OSX into a weird state with regard to Atom.

Is this happening consistently even after `script/clean && script/build`, and maybe rebooting your machine?

abe33:
I built Atom freshly this morning (1.7.0-dev-41161dd) and I have the same issue with cmd-backtick.

Someone seems to have reported the same kind of issue in atom/electron#4512

joshaber:
I can still reproduce this after a clean, rebuild, and restart.

bdielman:
Still seeing this issue intermittently on U.S. keyboard. Reloading Atom will fix it sometimes, but it's not consistent.

PR #10927, opened 2016-02-22
Upgrade electron to fix command-backtick bug

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/10852

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "0.36.7",
+  "electronVersion": "0.36.8",
   "dependencies": {
     "async": "0.2.6",
     "atom-keymap": "^6.2.0",


maxbrunsfeld:
Ok for me, command-backtick now works immediately, instead of failing to work for the first few seconds after startup. @joshaber Hopefully the worse behavior that you were seeing will be resolved as well.

joshaber:
:boom: :camel:

lee-dohm:
@amytruong you may want to play with this too?

amytruong:
:+1: @lee-dohm

amytruong:
I think I may have found a (pretty small) issue. 
Verifying this on master, I noticed:
cmd-backtick switches windows from oldest to newest
cmd-shift-backtick switches from oldest to newest

On beta:
cmd-backtick: newest to oldest
cmd-shift-backtick: from oldest to new

joshaber:
Great find @amytruong! I opened https://github.com/atom/electron/issues/4577 about this.

abe33:
Hmm, sorry to butt-in this late, but running on master (1.7.0-dev-f88051e) I don't see any fix on this. I'm using a french keyboard with custom layout to remove the dead key on backticks. Has it been reported already?

joshaber:
@abe33 I haven't seen any reports like that. An issue would be great!

abe33:
@joshaber sorry, I didn't see your reply, creating one right away.

simplyandyg:
I'm running Atom Version 1.15 and still seam to have this issue where command-back tick does not seem to register with Atom. When using the Keybinding Resolver and pressing command + back tick, only cmd shows up in the Resolver.

Command + back tick combo is working fine for other apps on my comp such as Chrome.

lee-dohm:
@simplyandyg Please open a new issue and fill out the bug template so that we can properly triage this.

simplyandyg:
@lee-dohm my bad. Figured out my issue. Not a problem with Atom. I'll try to stop being such a newb.

lee-dohm:
> But the Keybinding Resolver does still show that only cmd was pressed. I can open an issue for that.

This isn't a bug. The OS intercepts some keybindings and Atom doesn't receive them. There's really nothing we can do about that.

simplyandyg:
@lee-dohm okay, that makes sense. I was curious if something like that might be the reason, which is why I hadn't opened a bug yet. Thanks for the info!

PR #10930, opened 2016-02-22
BufferedProcess: search only new data for new lines rather than entire buffer

chuckhendo:
I've noticed for some time that viewing installed packages inside of settings-view was very slow. I did a little research on it, and reported the issue on the settings-view repo here atom/settings-view#734. It seems the issue is due to BufferedProcess trying to find new lines on the entire stream buffer, including data that it's already searched for new lines on. 

This PR fixes that by only searching the new data from the stream. Viewing installed packages is now significantly faster. I've attached a before and after screenshot of the CPU profile when opening settings-view and then going to the Packages tab.
### Before

screen shot 2016-02-22 at 3 24 51 pm
### After:

screen shot 2016-02-22 at 3 25 30 pm

### DIFF
--- a/src/buffered-process.coffee
+++ b/src/buffered-process.coffee
@@ -111,11 +111,12 @@ class BufferedProcess
 
     stream.on 'data', (data) =>
       return if @killed
+      bufferedLength = buffered.length
       buffered += data
-      lastNewlineIndex = buffered.lastIndexOf('\n')
+      lastNewlineIndex = data.lastIndexOf('\n')
       if lastNewlineIndex isnt -1
-        onLines(buffered.substring(0, lastNewlineIndex + 1))
-        buffered = buffered.substring(lastNewlineIndex + 1)
+        onLines(buffered.substring(0, lastNewlineIndex + bufferedLength + 1))
+        buffered = buffered.substring(lastNewlineIndex + bufferedLength + 1)
 
     stream.on 'close', =>
       return if @killed


50Wliu:
Very nice!  Can you please add some specs to this?

Zireael07:
3300 ms -> 8.5 ms. Is it a speedup record? :checkered_flag: :boom:

mehcode:
>  Very nice! Can you please add some specs to this?

I'm curious what specs would be needed. This is a few line change that significantly improves performance. If the existing specs pass, isn't that enough? It's not like new (and testable) functionality is being introduced here.

If we have no specs for that function and you're asking for @chuckhendo to write the specs from ground zero ... that's not something I'd do as a project maintainer but maybe Atom works differently.

chuckhendo:
I certainly don't mind writing specs, but agreed with @mehcode - I'm not exactly sure what I should write. Maybe something to test stdout to make sure it's properly separating lines?

lee-dohm:
I don't think that asking for specs is unreasonable, especially given that `BufferedProcess` is a part of the public Atom API. Even just a test that it doesn't duplicate or drop content would be good :grinning: Properly separating lines would be even better!

joshaber:
Thanks for doing all the heavy lifting on this @chuckhendo!

chuckhendo:
@joshaber Thank you for taking care of the specs! I had that on my todo list, but then I left for vacation. Glad to see this get merged in

Issue #10840, opened 2016-02-16
Flaky TextEditorPresenter spec


joshaber:
See https://travis-ci.org/atom/atom/jobs/109626270.

Similar to https://github.com/atom/atom/issues/10435.

joshaber:
I thought I fixed this in https://github.com/atom/atom/pull/10828 but apparently not :disappointed: I'm betting we're expecting something to emit a state change when it really doesn't.

PR #10933, opened 2016-02-23
Fix block decorations spec in text editor presenter

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/10840

### DIFF
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -1340,7 +1340,9 @@ describe "TextEditorPresenter", ->
               blockDecoration3 = addBlockDecorationBeforeScreenRow(7)
               blockDecoration4 = null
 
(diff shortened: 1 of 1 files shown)


spec/text-editor-presenter-spec.coffee
```diff
@@ -1338,7 +1340,9 @@ describe "TextEditorPresenter", ->
               blockDecoration3 = addBlockDecorationBeforeScreenRow(7)
               blockDecoration4 = null
 
-              waitsForStateToUpdate presenter, blockDecoration4 = addBlockDecorationAfterScreenRow(7)
```
maxbrunsfeld: This lil' block decoration was being passed as the callback argument to `waitsForStateToUpdate`, which was silently accepting it because of the `fn?()` syntax.

joshaber: Thanks for tracking this down!

Issue #10762, opened 2016-02-10
Add config option to disable the concept of "pending tabs"


bjmiller:
After updating to atom 1.6.0-beta0, I found that my usual habit of opening tabs through single clicking was disrupted by having the tabs replace each other instead of opening next to each other.

I had to dig in to the release notes to find the problem - I was automatically opted in to the feature of "pending tabs", with no way out.  I never want to see a tab in a "pending" state, ever.  There should be a way to turn this feature off, and return to the previous behavior, in which every tab is always "non-pending".  A checkbox in the settings should be sufficient for this.

kuychaco:
Planning to add a config option and patch it into 1.6 beta. Stay tuned.

bjmiller:
Thank you!

bjmiller:
Excellent!  Thank you so much for fixing this.  If I'd realized it was that easy, I might have submitted a patch myself. :)

kuychaco:
@bjmiller you're welcome. Thanks for your feedback! If you do decide to open a pull request in the future, we'd totally welcome it :)

lorenzos:
I'm still unable to get the old behaviour in Atom 1.8.0.

By default, if I single click items in the tree view, panes replaces themselves. If I uncheck the new *Allow Pending Pane Items* in core settings, single clicking an item does nothing. What me and I guess @bjmiller want is that the single click always opens new non-pending pane.

50Wliu:
@lorenzos You might be looking for https://atom.io/packages/single-click-open.

lorenzos:
@50Wliu Exactly! Thank you.

sekmo:
We have it now! https://flight-manual.atom.io/using-atom/sections/pending-pane-items/#disabling-pending-pane-items

PR #10951, opened 2016-02-24
return default to default open file behavior

tomByrer:
In stable, click on a file in the tree, Atom opens a new file tab.
In beta, by default, clicking on the file opens over the active tab sometimes.
Switching `allowPendingPaneItems: default: false` returns behavior to expected from v1.5.x.
ref #10762

Not sure if [these lines](https://github.com/atom/atom/blob/ae26e6c8ca284bb2858c65074c8d1d64564d1593/src/workspace.coffee#L417-L418) need to be changed.

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -111,7 +111,7 @@ module.exports =
       allowPendingPaneItems:
         description: 'Allow items to be previewed without adding them to a pane permanently, such as when single clicking files in the tree view.'
         type: 'boolean'
-        default: true
+        default: false
 
   editor:
     type: 'object'


Zireael07:
There is going to be an option to switch the pending pane option on or off.

tomByrer:
I'm confused, isn't this the switch?

BTW, "pending pane" is a confusing nomenclature; I wish it was called something more direct, like `openSameTab` or the like.

steelbrain:
:+1: to disabling it by default

benogle:
Sorry, we've decided to change the default behavior for now. If we did retain the existing behavior we would need to migrate the config setting from tabs.

tomByrer:
Sorry, but I disagree; shouldn't change expected behaviors by default, esp for a minor version bump.  Opening files is something people do many times a session....

PR #10955, opened 2016-02-24
TextEditor customization

### LINKED ISSUE #5503: Using atom-text-editor in packages
For reference I've made a package to display all the standalone `atom-text-editor` permutations. https://github.com/atom/atom-text-editor-exercise

Currently using an `atom-text-editor` in a package as a replacement for `textarea` isnt that great:

```html
<atom-text-editor>
    This editor
    grows
</atom-text-editor>
```

Produces:

screen shot 2015-02-11 at 12 10 10 pm

A few issues

* It shows line numbers by default
* it shows invisibles if I have them enabled
* It grows by default
* The styling is not like the mini editors

Constraining height makes it behave a little more like a text area, but you need to set the height in the `style` attribute or the editor does weird things:

```html
<atom-text-editor style="height: 100px">
  Some text
</atom-text-editor>
```

Produces (border added for clarity)

screen shot 2015-02-11 at 12 14 43 pm

Also not great. It introduces something that is nice in the pane-item editor, but not in a textarea: you cant get close to the bottom of the editor with the cursor, there is always a 3 line padding. 

editor-padding

### Proposal

I propose two new tags:

* `atom-text-area`: textarea emulation. This would be an editor that
  * does not show invisibles
  * does not show the gutter
  * does not have the 3 line padding 
  * does not grow by default
  * allows the themes to style the editor as a text area
* `atom-text-input`: a replacement for the mini editors

Then we could add attributes to enable things

* `show-gutters` would show the gutter
* `auto-height` would ignore css sizing
* `read-only` would set to read only instead of the remove `tabindex` hack 

The markdown preview can just use the `atom-text-editor` to get full code support.

joshaber:
From https://github.com/atom/atom/issues/5503:
- [ ] Don’t grow with content by default (`grow-vertically` attribute).
- [x] Don’t show gutters by default (`show-gutters` attribute).
- [ ] Don’t show invisibles by default (`show-invisibles` attribute).
- [ ] Don’t include 3 line padding.
- [ ] Add `read-only` attribute.

Did I miss anything? Does this seem like a reasonable way forward? /cc @atom/core
## 

This has evolved a bit since :point_up:.


joshaber:
Also is there any way for it to inherent the theme of `atom-text-editor` by default?

benogle:
One other thing to note is that markdown-preview uses `atom-text-editor`s to show code fences. Maybe as an example or point of reference.

maxbrunsfeld:
> Also is there any way for it to inherent the theme of atom-text-editor by default?

What's the advantage of having a custom tag for this, vs one or more html attributes on the `atom-text-editor`? A lot of existing third party code probably uses `atom-text-editor` in selectors or command bindings.

lee-dohm:
> Also is there any way for it to inherent the theme of atom-text-editor by default?

Isn't that why we have the `<atom-text-editor class="mini">` convention? Maybe we should do `<atom-text-editor class="textarea">`?

joshaber:
@maxbrunsfeld @lee-dohm A new tag was the consensus of https://github.com/atom/atom/issues/5503, so @nathansobo and @benogle would be in the best position to address that.

benogle:
I defer to @nathansobo on this fo sho.

joshaber:
Maybe the answer is "it shouldn't inherit styles, that's entirely the point of the new tag." But I'll certainly defer to @nathansobo.

nathansobo:
Writing up some thoughts right now.

nathansobo:
# 1:23 PM – atom-textarea

In general, custom elements are a non-starter because there's no facility for versioning them. Tags are a global namespace. In core, they're somewhat more okay, because they're part of the core API which is global anyway (for now). There was a reason I advocated for them. Custom tags are essentially new primitives in the language of HTML and are really easy for package authors to access regardless of their view technology choice. They can set `innerHTML` on a node and still easily build a text editor inside it.

My concern with custom elements, even for core, is that they're forever. I can envision a scheme in which we version the Atom core APIs by loading each package in a custom context and exposing it to a versioned facade for our latest API. With custom elements, that's impossible because they rely on a non-negotiable global variable, the `document`. So whatever APIs we come up with today, we'll have to stick with pretty much forever. That makes me nervous.

Ideally, I think all components should just be plain JavaScript objects with an `.element` property or a `.getElement()` method (I haven't decided which is preferable, the flexibility of `.getElement()` or the simplicity of `.element`). Ideally, that would also include any UI components that we bake into the core API. You'd access these components like this:

```js
import {TextEditorComponent} from 'atom'

// to use imperatively
let area = new TextEditorComponent({textarea: true})
elementNeedingAnEditor.appendChild(area.getElement())

// to use with Etch or another JSX library supporting the `getElement` interface
class ComponentNeedingAnEditor {
  render () {
    <div class='component-needing-an-editor'>
      <TextEditorComponent textarea={true} />
    </div>
  }
}
```

This is less deeply embedded than the custom HTML tag approach, but it gives us a lot more flexibility.

If we move away from custom tags, it also starts to seem silly to build a different component for what is essentially a different mode of the text editor. With tags it made more sense because they felt like language primitives. With the component object for some reason it seems less appealing.

If anyone feels strongly we should also export an `AtomTextareaComponent`, I'd like to hear their arguments. Otherwise I tend to agree with @maxbrunsfeld. Sorry for changing my mind... I forgot I even wrote that stuff until now.

joshaber:
Awesome, thanks for the brain dump @nathansobo.

So the new direction: make `TextEditorElement` more configurable, in the ways I outlined in the issue body. @atom/core :+1:?

maxbrunsfeld:
:+1: :exclamation:

nathansobo:
@joshaber 

> So the new direction: make TextEditorElement more configurable, in the ways I outlined in the issue body.

This is a bigger pivot, but I actually think that we should pivot toward `TextEditorComponent` (which the element currently instantiates as an implementation detail) being our public interface. :grimacing:

nathansobo:
While I'm talking about it, another blunder of the "embrace web components" era was using the shadow DOM instead of just prefixing scope classes with `scope--` or something. Would love to back ourselves out of that corner someday.

joshaber:
>This is a bigger pivot, but I actually think that we should pivot toward TextEditorComponent (which the element currently instantiates as an implementation detail) being our public interface. :grimacing:

And just for my own sake: this would mean users would instantiate `TextEditorComponent`s directly, vs. going through the view provider system, correct?

joshaber:
Alright, I pushed up some new changes. I'd love to get some feedback on the new direction. I'm trying balance the new aim of `TextEditorComponent` being the user-facing API, while also not refactoring more than I care to take on.

To that end, I've created `TextEditorComponent` which uses a `TextEditorElement` internally and exposes `element` so that it can be easily used from Etch / anything that just wants a DOM element.

1. Is this an acceptable direction?
2. Should `TextEditorComponent` accept a `TextEditor`, or should it construct one based on arguments?

I really appreciate all the feedback :heart:

nathansobo:
@joshaber give me a few minutes to write up more thoughts after talking on the :phone: to @maxbrunsfeld.

nathansobo:
We're rethinking the model/component separation. Originally, with SpacePen, our components were jQuery wrappers around DOM nodes with a bunch of extra methods added. That meant our components had a huge API surface that invited people to screw with the DOM... it ended up being a bad idea to have `atom.workspace` expose a `.find()` or `.hide()` method... so I separated out all these model objects.

Unfortunately, I think I may have gone too far. The view provider system is a really complicated trade-off for having the models be completely pure. In reality, it was just their jQuery-polluted API surface area that posed the problem. Instead, we should just own up to the fact that there's a 1:1 relationship between these "models" and DOM elements and call them components. We just have to add a `getElement` method.

I still think we should keep the objects organized exactly as they are. The `TextEditorComponent` might be better described as the `TextEditorRenderer` since it has complex view update logic that won't be found in most views that using a virtual DOM. The `TextEditor` can just build a renderer instance lazily when `getElement` is called to avoid the overhead in logic-only tests.

* We should just add a `getElement` method to the `TextEditor` "model". It can construct a component as needed lazily and return its associated element.
* Rename `TextEditorComponent` to `TextEditorRenderer`.
* We should change the `ViewRegistry` to prefer calling `getElement` on any object if that method is defined.
* We should change Etch to call `getElement` on a child component if it exists rather than referencing the `.element` field.
* To build a text editor in an Etch component, do this:
  ```js
  // We can't export a constructor from the `'atom'` `require` because
  // TextEditor needs to be dependency-injected with environment-specific
  // services. But it looks pretty ugly having a verb in your JSX expressions so
  // saving off a noun-y const is a nice workaround.
  const TextEditor = atom.workspace.buildTextEditor

  class MyComponent {
    render () {
      <div>
        <TextEditor />
      <div>
    }

  }
  ```
* Bind `atom.workspace.buildTextEditor` in `Workspace` so it can be called in a detached way.

simurai:
Also these?

- [ ]  Don’t scroll past end
- [ ]  Don’t change font size
- [ ]  Don’t soft wrap

But I guess depends on the situation. e.g. soft wrap is probably better when used as text input like a `<textarea>`, but not when used to show fenced code blocks like a `<pre>`.

joshaber:
>Don’t change font size

@simurai What's that one mean?

joshaber:
@nathansobo Could you give this a lil lookyloo to make sure it's taking the shape you expected?

nathansobo:
:+1:

src/workspace.coffee
```diff
@@ -550,7 +556,7 @@ class Workspace extends Model
   # Extended: Create a new text editor.
   #
   # Returns a {TextEditor}.
-  buildTextEditor: (params) ->
+  buildTextEditor_: (params) ->
```
nathansobo: I don't think you should *need* to suffix this with an underscore because this version is on the prototype whereas the version bound in the constructor will be on the instance. Maybe that's too vague though.

src/workspace.coffee
```diff
@@ -43,6 +43,12 @@ class Workspace extends Model
     @defaultDirectorySearcher = new DefaultDirectorySearcher()
     @consumeServices(@packageManager)
 
+    # One cannot simply .bind here since it could be used as a component with
+    # Etch, which means it'd be `new`d in which case `this` would the new
+    # object.
+    realThis = this
+    @buildTextEditor = (params) -> realThis.buildTextEditor_(params)
```
nathansobo: Did you try `@buildTextEditor = @buildTextEditor.bind(this)`?

nathansobo:
Looks good modulo a couple stylistic comments. Thanks a ton for rolling with this.

src/workspace.coffee
maxbrunsfeld: Yeah agreed, it's better if the object only has a `buildTextEditor` method, not both `buildTextEditor` and `buildTextEditor_`.

src/workspace.coffee
joshaber: We talked about this in chat, but for posterity: basically when used as `new buildTextEditor()`, `this` is always the newly allocated object, instead of what we bound:

```
$ node -e 'var A=function(){console.log(this)}; B=A.bind(1); B(); new B();'
[Number: 1]
{}

**Comment by maxbrunsfeld** [comment]
yes

File: src/workspace.coffee (hunk 1 above)
Comment (joshaber) [inline src/workspace.coffee]: The problem I found was that if `@buildTextEditor` was called, which then called `realThis.buildTextEditor`, we'd infinitely recurse.

Comment (maxbrunsfeld) [inline src/workspace.coffee]: What about `Workspace.prototype.buildTextEditor.apply(realThis, arguments)`?

Comment (maxbrunsfeld) [inline src/workspace.coffee]: or capturing `this.buildTextEditor` as a local `buildTextEditor` before building the bound function

File: src/workspace.coffee [hunk 3]
```diff
@@ -43,6 +43,12 @@ class Workspace extends Model
     @defaultDirectorySearcher = new DefaultDirectorySearcher()
     @consumeServices(@packageManager)
 
+    # One cannot simply .bind here since it could be used as a component with
+    # Etch, in which case it'd be `new`d. And when it's `new`d, `this` is always
+    # the newly created object.
+    realThis = this
+    @buildTextEditor = -> Workspace.prototype.buildTextEditor.apply(realThis, arguments)
```
nathansobo: Love this approach and really tells the story. :sparkles:

simurai:
> Don’t change font size

@joshaber Will it be a new custom element `<atom-textarea>`? Then it should be fine.

If it's also a `<atom-text-editor>` then it uses the "Font Size" from the settings or changes when you `cmd +/-`. It could just overridden with `@input-font-size`, like we do with [mini editors](https://github.com/atom/atom/blob/master/static/text-editor-light.less#L10-L11).

joshaber:
>@joshaber Will it be a new custom element <atom-textarea>? Then it should be fine.

Ah, no, it won't. In my mind I imagine package developers specifying `font-size` manually. That should accomplish the same ends, right?

simurai:
>  In my mind I imagine package developers specifying `font-size` manually. 

Yes, they could override it. For `<atom-text-editor mini>`, most UI themes already to it.

joshaber:
@atom/core This is ready for :eyes: 

@nathansobo has outlined A Better World for `TextEditor` in #10979, but in the short term, I think this'll do :pig:

lee-dohm:
Looks good from my point of view :+1:

nathansobo:
So to summarize your approach, you're just allowing override flags to suppress things like invisibles and scroll past end?

joshaber:
>So to summarize your approach, you're just allowing override flags to suppress things like invisibles and scroll past end?

Right. In combination that should let package authors use `TextEditor` more generally for text entry.

nathansobo:
@joshaber How would you feel about phrasing the new ignore flags as positive overrides instead? So instead of `ignoreInvisibles: true`, you would pass `invisibles: false`. Similarly you'd pass `scrollPastEnd: false`.

This would pave the way for a world in which we don't inject the config system at all and always assign these things from the outside. Eventually, the default would actually be to *not* scroll past end or show invisibles, but at least the option names would slot in nicely.

Also, what do you think about injecting a `grammar` instance instead of a `grammarName` string? The reason for this is similar. According to my longer-term plan, we would stop injecting the grammar registry entirely, so this would co-exist much more peacefully with that reality than passing in a name.

joshaber:
@nathansobo Yeah, I went do that road at first and then switched to the negative phrasing because it felt weird since we were still pulling values from the config. But you're totally right that positive phrasing is more future-friendly :+1:

nathansobo:
Other than those changes I'm :cool: with this. Thanks!

joshaber:
Changes made :+1:

@atom/core Anyone else have feels?

nathansobo:
:sparkles:

PR #10959, opened 2016-02-25
Refactor pending state to live in pane instead of items

kuychaco:
- ~~New public API `workspace.setItemNotPending` that
  packages can use to set an item to set an item to not pending
  (e.g. when the user interacts with the item)~~
- Pending state for newly opened items with `{pending: true}`
  is now tracked by `Pane` instead of the item, and packages like
  `tabs` that query this information now get it from the Pane.

See corresponding changes:
- https://github.com/atom/tabs/pull/268
- https://github.com/atom/tree-view/pull/741

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -18,8 +18,8 @@ describe "Pane", ->
     onDidDestroy: (fn) -> @emitter.on('did-destroy', fn)
     destroy: -> @destroyed = true; @emitter.emit('did-destroy')
     isDestroyed: -> @destroyed
-    isPending: -> @pending
-    pending: false
+    onDidTerminatePendingState: (callback) -> @emitter.on 'terminate-pending-state', callback
+    terminatePendingState: -> @emitter.emit 'terminate-pending-state'
 
   beforeEach ->
     confirm = spyOn(atom.applicationDelegate, 'confirm')
@@ -136,10 +136,8 @@ describe "Pane", ->
       pane = new Pane(paneParams(items: []))
       itemA = new Item("A")
       itemB = new Item("B")
-      itemA.pending = true
-      itemB.pending = true
-      pane.addItem(itemA)
-      pane.addItem(itemB)
+      pane.addItem(itemA, undefined, false, true)
+      pane.addItem(itemB, undefined, false, true)
       expect(itemA.isDestroyed()).toBe true
 
   describe "::activateItem(item)", ->
@@ -172,19 +170,17 @@ describe "Pane", ->
       beforeEach ->
         itemC = new Item("C")
         itemD = new Item("D")
-        itemC.pending = true
-        itemD.pending = true
 
       it "replaces the active item if it is pending", ->
-        pane.activateItem(itemC)
+        pane.activateItem(itemC, true)
         expect(pane.getItems().map (item) -> item.name).toEqual ['A', 'C', 'B']
-        pane.activateItem(itemD)
+        pane.activateItem(itemD, true)
         expect(pane.getItems().map (item) -> item.name).toEqual ['A', 'D', 'B']
 
       it "adds the item after the active item if it is not pending", ->
-        pane.activateItem(itemC)
+        pane.activateItem(itemC, true)
         pane.activateItemAtIndex(2)
-        pane.activateItem(itemD)
+        pane.activateItem(itemD, true)
         expect(pane.getItems().map (item) -> item.name).toEqual ['A', 'B', 'D']
 
   describe "::activateNextItem() and ::activatePreviousItem()", ->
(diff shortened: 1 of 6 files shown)


BinaryMuse:
This is still a WIP, but would love to get some early feedback on aspects of this from @atom/core.

Previously, the pending state of an item was owned by the item itself; if packages like [`tabs`](https://github.com/atom/tabs/blob/f6234cf8cbe8ec3d88c3b8f8ea71931b1eeaba92/lib/tab-bar-view.coffee#L343-L345) or [`tree-view`](https://github.com/atom/tree-view/blob/184990266c2d0b234266eb09adbb5a0e815f5d9d/lib/tree-view.coffee#L216) wanted to mark an item as being permanent, it would call `item.terminatePendingState?()`. Additionally, items needed to implement `isPending()` [and `onDidTerminatePendingState(callback)`](https://github.com/atom/tabs/blob/f6234cf8cbe8ec3d88c3b8f8ea71931b1eeaba92/lib/tab-view.coffee#L40-L45) to properly support having a pending state — meaning that items had to fully opt in to a specific, implicit interface in order to ever be set as pending.

```
┌─────────────────────────────────────────────────────────────┐
│                                                             │
│                     tabs/tree-view/etc.                     │
│                                                             │
└──┬──────────┬────────────────▲──────────────────────────────┘
   │          │                │
   │     (on double-click)     │
   │  terminatePendingState()  │
   │          │                │
   │          │           ┌────┘
isPending()   │           │
   │          │     (when item thinks it shouldn't
   │          │          be pending any more)
   │          │           │
┌──▼──────────▼───────────┴───┐ ┌─────────────────────────────┐
│                             │ │                             │
│            Item             │ │                             │
│         #isPending          │ │            Item             │
│   #terminatePendingState    │ │    (no pending support)     │
│ #onDidTerminatePendingState │ │                             │
│                             │ │                             │
└─────┬──────────────▲────────┘ └─────────────────────────────┘
      │              │
      (on interaction)
   terminatePendingState()
      │              │
      │              │
      └──────────────┘
```

For this reason, openers that add non-`TextEditor` items to panes created items that were never set as pending (e.g. atom/tabs#266).

This PR moves the pending state off the item and into its containing `Pane`. When an item is opened with `{pending: true}`, the `Pane` keeps track of whether or not the item is pending. When a package wants to change the item to be permanent, it can call into a new method to clear the pending state.

This means that *all* items support being pending (because Atom manages that state upon adding the item to a `Panel`), and items can optionally *enhance* the notion of what it means to be pending or not (e.g. `TextEditor`s stop being pending when you type in them).

```
┌───────────────────────────────────────────────┐
│                                               │
│              tabs/tree-view/etc.              │
│                                               │
└──────┬──────────────────────────────────┬─────┘
       │                        (on double-click)
isItemPending(item)            clearPendingItem()
───────┼──────────────────────────────────┼──────
       │                                  │
       │                                  │
┌──────▼──────────────────────────────────▼─────┐
│                                               │
│                     Pane                      │
│                                               │
└───────────────────────────────────────▲───────┘
                                        │
                              clearPendingItem()
    ┌────(on interaction)──────┐        │
    │   clearPendingItem()     │        │
    │                          │        │
┌───┴─────────────────┐ │ ┌────▼────────┴───────┐
│                     │ │ │                     │
│        Item         │ │ │      Workspace      │
│                     │ │ │                     │
└─────────────────────┘ │ └─────────────────────┘
```

One big question I have is around the new API for clearing the pending state for an item; to get this idea moving, we implemented it on `Workspace` (`Workspace#setItemNotPending(item)`), but I think this could probably be better. I wonder if it's more reasonable to allow an item to figure out which pane it's in, and call a method on `Pane`, e.g.

```coffeescript
myPane = atom.workspace.getPaneForItem(this)
myPane.clearPendingItem() if myPane.getPendingItem() == this
```

nathansobo:
Naming-wise, I still like something like `terminateItemPendingState(item)`. Or something to make it clearer that the verb is operating on the "pendingness" of the item rather than the item itself.

Adding a pass-through method on the workspace itself does seem to make this more usable, because you don't have to care about the item's pane. Obviously it's two methods instead of one though, so that's a downside. Some changes we're contemplating to the view system could potentially open up the ability for the same item instance to appear in multiple panes in the distant future, but this could deal with that pretty gracefully.

lee-dohm:
Does this break people's expectations of things that should *not* be pending? Examples:

* Markdown Preview (for that matter PDF preview, ASCIIdoc preview, perldoc preview, etc)
* Find In Project results
* More?

BinaryMuse:
@lee-dohm If I understand your question, you can still open those items as non-pending if you'd like (you have to pass `{pending: true}` in the options for `Workspace#open` or `Pane#addItem` for an item to be pending, just as before)

lee-dohm:
@BinaryMuse Thanks, that answers my question :grinning:

nathansobo:
Please hold off on merging this until we talk. I'm not sure it's a good idea due to some realizations @maxbrunsfeld and I are having about `TextEditor`.

BinaryMuse:
Specs have been updated, and along with atom/tabs#268 and atom/tree-view#741 wraps up the refactor for this feature. Happy to get more :eyes:/opinions from @atom/core, otherwise I'll merge later today.

Issue #10996, opened 2016-02-28
Show tooltip immediately if the tooltip trigger is manual


oclbdk:
I think tooltips should display immediately for `trigger: 'manual'` (currently, setting that trigger type has no effect).

The Bootstrap tooltip has a `trigger` option that's `click | hover | focus | manual` http://getbootstrap.com/javascript/#tooltips-options For `'manual'` you need to call `show()` on the tooltip to get it to display, but the current Atom tooltip API doesn't expose that method. This makes it hard to get a tooltip to show immediately (e.g. you'd have to set `trigger: 'click'` and dispatch a click event on the target element).

Ben3eeE:
Closed via https://github.com/atom/atom/pull/11063.

PR #10997, opened 2016-02-28
Show tooltip immediately if the tooltip trigger is manual

oclbdk:
https://github.com/atom/atom/issues/10996


nathansobo:
Just for the record, I've merged this via #11063 with a small change to make the build pass.

PR #10999, opened 2016-02-28
Upgrade language-javascript to 0.111.0

winstliu:
Time to update some more specs!


50Wliu:
Well, it seems like I managed to crash Electron.  That's pretty impressive.

On a different note, I have no idea how to fix the core specs this time around.

50Wliu:
This crash started occurring after running specs a few times locally:
```
<--- Last few GCs --->

   79973 ms: Scavenge 698.5 (739.0) -> 698.5 (739.0) MB, 4.6 / 0 ms (+ 5.0 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
   81362 ms: Mark-sweep 698.5 (739.0) -> 698.5 (739.0) MB, 1389.4 / 0 ms (+ 7.0 ms in 2 steps since start of marking, biggest step 5.0 ms) [last resort gc].
   82585 ms: Mark-sweep 698.5 (739.0) -> 698.5 (739.0) MB, 1223.2 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 2C024291 <String[7]: file://>
    2: emitString [C:\Users\username\Documents\GitHub\atom\vendor\jasmine.js:~1959] [pc=48B2A229] (this=11D41939 <JS Object>,value=21DDED95 <String[12]: Null Grammar>)
    3: format [C:\Users\username\Documents\GitHub\atom\vendor\jasmine.js:1905] [pc=48B2B5F2] (this=11D41939 <JS Object>,value=21DDED95 <String[12]: Null Grammar>)
    4: /* anonymous */(aka /* anonymous */) [C:\Users\u...

[120:0229/191828:ERROR:atom_bindings.cc(39)] Fatal error in V8: CALL_AND_RETRY_LAST Allocation failed - process out of memory
[11488:8624:0229/191829:VERBOSE1:crash_service.cc(394)] dump for pid = 120 is C:\Users\username\AppData\Local\Temp\Atom Crashes\391376cf-c36b-4ce5-9ce4-f866aaeb5911.dmp
[11488:8164:0229/191830:VERBOSE1:crash_service.cc(442)] trying to send report for pid = 120
Error processing argument at index 9, conversion failure from
[11488:3376:0229/191830:VERBOSE1:crash_service.cc(331)] client end. pid = 120
[11488:9720:0229/191830:VERBOSE1:crash_service.cc(331)] client end. pid = 120
TypeError: Error processing argument at index 9, conversion failure from
    at TypeError (native)
    at Object.module.exports.showMessageBox (C:\Program Files\Atom\resources\atom.asar\browser\api\lib\dialog.js:155:20)
    at EventEmitter.<anonymous> (file:///C:/Users/username/Documents/GitHub/atom/src/browser/atom-window.coffee:136:23)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
[11488:9720:0229/191830:VERBOSE1:crash_service.cc(331)] client end. pid = 8664
[11488:8164:0229/191832:VERBOSE1:crash_service.cc(472)] dump for pid =120 crash2 id =<network issue>
[11488:9720:0229/191832:VERBOSE1:crash_service.cc(352)] zero clients. exiting
[11488:9076:0229/191832:VERBOSE1:crash_service.cc(490)] session ending..
[11488:9076:0229/191832:VERBOSE1:crash_service.cc(495)] clients connected :3
clients terminated :3
dumps serviced :0
dumps reported :0
[11488:9076:0229/191832:VERBOSE1:crash_service_main.cc(89)] Session end. return code is 0
```
Dump is attached.
[90dd8456-1fdf-4cf9-be6a-3a9a502e1bc9.zip](https://github.com/atom/atom/files/152028/90dd8456-1fdf-4cf9-be6a-3a9a502e1bc9.zip)

:grimacing:

50Wliu:
Summarizing my findings after working on this for a good couple of hours:
The problem seems to be that the internal screen lines aren't updating correctly (`DisplayBuffer::buildScreenLines`) after the tokenized buffer changed.  It works the first time but fails subsequently.  Not sure why a language update would break this though.

The screen lines should contain this after the deletion on line 837 of display-buffer-spec.coffee:
```js
var quicksort = function () {
  var sort = function(items) {
    if (items.length <= 1) return items;
      current = items.shift();
    }
    return sort(left).concat(pivot).concat(sort(right));
  };

  return sort(Array.apply(this, arguments));
};
```
but instead, they contain this:
```js
var quicksort = function () {
  var sort = function(items) {
    if (items.length <= 1) return items;
      current = items.shift();
    }

  return sort(Array.apply(this, arguments));

  return sort(Array.apply(this, arguments));
};
```

lee-dohm:
@as-cii Can you take a look at this? I was able to replicate the test failures (though not the crash) and there seems to be some weirdness in the display-buffer.

50Wliu:
The crash was probably a one-time thing, as I haven't been able to reproduce it after that either.

50Wliu:
After discussion with @as-cii the problem seems to be unbalanced rules somewhere caused by the refactor in 0.111.0.  I'll be adding extra scopes later to make sure everything's being closed correctly.

as-cii:
So, this seems to be a problem with having unbalanced stacks (after https://github.com/atom/language-javascript/pull/312), either because of the grammar or `first-mate` not working properly.

@50Wliu: could you post your findings here? I think I won't be able to follow up on this next week, and it would be nice to have a summary of what we think is happening on this issue.

50Wliu:
As a recap:
The aforementioned PR drastically refactored function matching to enclose the entire function in a `function_body` scope that was controlled by opening and closing brackets.  Brackets are matched recursively through `$self` to prevent functions from closing at the wrong time.

So now this works fine for normal files, but what the failing spec is doing is deleting a line which has an opening bracket on it.  This results in an unbalanced bracket scenario which is illustrated in the following snippet:
```js
var quicksort = function () {
  var sort = function(items) {
    if (items.length <= 1) return items;
      current = items.shift();
    }  //<-- this bracket is matched with the one on line 2
    return sort(left).concat(pivot).concat(sort(right));
  };  //<-- and this one with the one on line 1

  return sort(Array.apply(this, arguments));
};  //<-- this bracket isn't assigned a scope
```
You can see this behavior visually through Bracket Matcher.  I assume this scenario is causing unbalanced stacks, and completely confusing the display buffer.

Still working on figuring out which part of the code is to blame though.

MaximSokolov:
@50Wliu the snippet you posted for the illustration works fine:
```js
var quicksort = function () { // 1
  var sort = function(items) {  // 2
    if (items.length <= 1)
      return items;
    current = items.shift();
    }  // <- extra bracket
  }; // 2

  return sort(Array.apply(this, arguments));
}; // 1
```
Any progress on this?

50Wliu:
I've updated this PR to language-javascript@0.116.0 and pulled in the latest changes from master, but the failures remain.

The brighter news is that #11414 seems to fix this issue.

50Wliu:
@as-cii I think I've "fixed" the remaining TokenizedBuffer spec failures.  These failures are closely related to the dangling bracket scenario above: all but one of the failures comment out an opening bracket, which causes changes to cascade down to the next one.  Fixing those specs has the unfortunate side effect of not really being able to tell if everything works as intended though, since the default tokenization chunk is also 5 :/.

BinaryMuse:
@as-cii What do we need to do to get this into master before the next railcar release?

as-cii:
I was somewhat concerned about performance, because `language-javascript` seems to apply rules that are more expensive now after `v0.110.0`. I guess that's fine for now though, so I'll go ahead and merge this so that I have some time to test it before the next railcar release.

Thanks, @50Wliu! ✨

50Wliu:
Thanks @as-cii!  I'll be happy to work with you to fix any performance problems that show up.

PR #11001, opened 2016-02-29
onDidTerminatePendingState ➡︎ onItemDidTerminatePendingState

BinaryMuse:
Continuation of #10959

### DIFF
--- a/spec/pane-spec.coffee
+++ b/spec/pane-spec.coffee
@@ -183,6 +183,41 @@ describe "Pane", ->
         pane.activateItem(itemD, true)
         expect(pane.getItems().map (item) -> item.name).toEqual ['A', 'B', 'D']
 
+  describe "::setPendingItem", ->
+    pane = null
+
+    beforeEach ->
+      pane = atom.workspace.getActivePane()
+
+    it "changes the pending item", ->
+      expect(pane.getPendingItem()).toBeNull()
+      pane.setPendingItem("fake item")
+      expect(pane.getPendingItem()).toEqual "fake item"
+
+  describe "::onItemDidTerminatePendingState callback", ->
+    pane = null
+    callbackCalled = false
+
+    beforeEach ->
+      pane = atom.workspace.getActivePane()
+      callbackCalled = false
+
+    it "is called when the pending item changes", ->
+      pane.setPendingItem("fake item one")
+      pane.onItemDidTerminatePendingState (item) ->
+        callbackCalled = true
+        expect(item).toEqual "fake item one"
+      pane.setPendingItem("fake item two")
+      expect(callbackCalled).toBeTruthy()
+
+    it "has access to the new pending item via ::getPendingItem", ->
+      pane.setPendingItem("fake item one")
+      pane.onItemDidTerminatePendingState (item) ->
+        callbackCalled = true
+        expect(pane.getPendingItem()).toEqual "fake item two"
+      pane.setPendingItem("fake item two")
+      expect(callbackCalled).toBeTruthy()
+
   describe "::activateNextRecentlyUsedItem() and ::activatePreviousRecentlyUsedItem()", ->
     it "sets the active item to the next/previous item in the itemStack, looping around at either end", ->
       pane = new Pane(paneParams(items: [new Item("A"), new Item("B"), new Item("C"), new Item("D"), new Item("E")]))
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -441,8 +441,10 @@ class Pane extends Model
     item
 
   setPendingItem: (item) =>
-    @pendingItem = item if @pendingItem isnt item
-    @emitter.emit 'did-terminate-pending-state' if not item
+    if @pendingItem isnt item
+      mostRecentPendingItem = @pendingItem
+      @pendingItem = item
+      @emitter.emit 'item-did-terminate-pending-state', mostRecentPendingItem
 
   getPendingItem: =>
     @pendingItem or null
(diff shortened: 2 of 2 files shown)


BinaryMuse:
This PR renames `Pane::onDidTerminatePendingState` to `Pane::onItemDidTerminatePendingState` and makes the following API changes:

* Calling `Pane::setPendingItem(item)` *always* calls back to `onItemDidTerminatePendingState` callbacks assuming the item being set as pending isn't the same as the currently pending item (if any)
* Callbacks for `onItemDidTerminatePendingState` are passed the item that is no longer pending as the first parameter

Most of the package code that was utilizing this API wanted to answer the question, "Was the thing that just stopped being pending the item I'm managing?" and tended to do so via awkward internal tracking of state and a check to `pane.getPendingItem() is @myItem`; this should simplify most of these use cases to

```coffeescript
pane.onItemDidTerminatePendingState (item) =>
  @doSomethingWhenNotPending() if @myItem is item
```

paulpflug:
I think there is a little problem with this pr.
`clearPendingItem()` calls [`@setPendingItem(null)`](https://github.com/atom/atom/blob/master/src/pane.coffee#L467), which doesn't call `terminatePendingState()` on the item and therefore doesn't emit `onDidTerminatePendingState` on the item.

I think this should be:
```coffee
  setPendingItem: (item) =>
    if @pendingItem isnt item
      mostRecentPendingItem = @pendingItem
      @pendingItem = item
      mostRecentPendingItem.terminatePendingState()
      @emitter.emit 'item-did-terminate-pending-state', mostRecentPendingItem
```

BinaryMuse:
@paulpflug A method called `terminatePendingState()` on items, with an associated `onDidTerminatePendingState`, was used briefly to control the pending-ness of pane items. Starting in v1.6.0, however, we don't call `terminatePendingState()` on items — the pendingness of items is controlled exclusively through the `Pane`; see [this comment](https://github.com/atom/atom/pull/10959#issuecomment-188910672) for a little discussion around why we made that change.

We do still have a clause for supporting `item.onDidTerminatePendingState` for backward compatibility with items that don't know about the new API; if the callback we pass in is called, then we assume the item is trying to make itself permanent. As far as notifications go, going forward, if an item needs to be notified when it has been made permanent, [it can subscribe to](https://github.com/atom/tabs/blob/1200c00a7a7478f0e96efbf2e8abbca06b0a8c22/lib/tab-view.coffee#L42-L44) (the as of yet undocumented) `onItemDidTerminatePendingState` callback.

I hope that helps; please let me know if I've misunderstood your comment!

paulpflug:
hmm I understand..

I wanted to use `paneItem.hasTerminatedPendingState`
but that isn't working either..
What would be the right way to get the containing pane of a paneItem?
```coffee
atom.workspace.observeTextEditors (editor) =>
  # get the containing pane
  pane.isItemPending(editor) #right?
```

BinaryMuse:
@paulpflug Unfortunately, there's not a great API for this right now; I had planned on adding one, but some potential changes to some APIs made me hold off. For now, you could use [`Workspace::getPanes`](https://atom.io/docs/api/v1.6.1/Workspace#instance-getPanes) and check for the item in each Pane manually.

paulpflug:
New problem: `onItemDidTerminatePendingState` gets called on the old item when I open a new one in pending state and the old gets destroyed.
```coffee
atom.workspace.observeTextEditors (editor) =>
  pane = atom.workspace.paneForItem(editor)
  if pane.getPendingItem() != editor
    # not pending
  else
    disposable = pane.onItemDidTerminatePendingState (paneItem) =>
      if paneItem == editor
        # pending terminated
        # gets called when I open 2 pending items consecutively 
        disposable.dispose()
     disposable2 = editor.onDidDestroy ->
        disposable.dispose()
        disposable2.dispose()
```
other question, is there an easier way of achieving, what I want?

thanks for your fast and profound help btw ;)

paulpflug:
this could be fixed like this
````coffee
addItem: (item, options={}) ->
    # ...
    lastPendingItem = @getPendingItem()
    @pendingItem = null
    @setPendingItem(item) if pending
    # ...

  setPendingItem: (item) =>
    if @pendingItem isnt item
      mostRecentPendingItem = @pendingItem
      @pendingItem = item
      if mostRecentPendingItem?
        @emitter.emit 'item-did-terminate-pending-state', mostRecentPendingItem
```

here my workaround:
````coffee
destroyed = false
disposable = pane.onItemDidTerminatePendingState (paneItem) =>
  if paneItem == editor
    setTimeout ( =>
      unless destroyed
        # actually terminated pending state
        disposable.dispose()
      ),100
disposable2 = editor.onDidDestroy ->
  destroyed = true
  disposable.dispose()
  disposable2.dispose()
```

BinaryMuse:
@paulpflug 

> New problem: `onItemDidTerminatePendingState` gets called on the old item when I open a new one in pending state and the old gets destroyed.

I apologize, maybe it's just the ☀️ early morning time ☀️ for me but I read over this a couple times and I'm not sure I'm fully understanding the problem; the behavior described there is, I *think*, correct. It'd be helpful for me to understand the problem well enough to be able to write a failing test for `Pane`.

paulpflug:
I will try to clarify:
In my opinion `onItemDidTerminatePendingState` should ONLY get called when an item is made permanent (however this is achieved).
If this is wrong, the behavior is right, otherwise the following would be necessary to replicate the problem:

- open item1 in pending state
- open item2 in pending state

item1 will get closed but `onItemDidTerminatePendingState` will be called on item1 [here](https://github.com/atom/atom/blob/master/src/pane.coffee#L450), despite it isn't made permanent but closed

BinaryMuse:
@paulpflug Thanks for the clarification; I see what you mean now. I've created #11352 to address this.

PR #11008, opened 2016-02-29
Add documentation for order key in config.

jonjitsu:
The config documentation:

https://atom.io/docs/api/latest/Config

was missing info on the order key.


lee-dohm:
Thanks! :rainbow:

PR #11023, opened 2016-03-02
Avoid creating empty buffers for deleted previous project paths

nathansobo:
Closes #10628 

To test this:
- Open Atom with 2 directories.
- Delete one of the directories and reload Atom.
- Atom should be opened with only one directory.

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -21,6 +21,14 @@ describe "Project", ->
     afterEach ->
       deserializedProject?.destroy()
 
+    it "does not deserialize paths to non directories", ->
+      deserializedProject = new Project({notificationManager: atom.notifications, packageManager: atom.packages, confirm: atom.confirm})
+      state = atom.project.serialize()
+      state.paths.push('/directory/that/does/not/exist')
+      state.paths.push(path.join(__dirname, 'fixtures', 'sample.js'))
+      deserializedProject.deserialize(state, atom.deserializers)
+      expect(deserializedProject.getPaths()).toEqual(atom.project.getPaths())
+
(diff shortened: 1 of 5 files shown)


jrkoenig:
This doesn't help #10628, I tested it. The reason it doesn't work is the bad paths are stored as initialPaths in `.atom/storage/application.json`, so they are scheduled to load before deserialize even runs. I don't think silently dropping non-existent project roots is a good idea as then we'd be implicitly undoing an explicit user action to add the project folder.

The real problem in #10628 is that editor state is saved in application.json (through some complex initialPaths marshalling through to and from the browserwindow). Fortunately all the state is properly handled in the regular Project/AtomEnvironment loading/saving, so all we have to do is make sure that properly loaded state is restored properly. I'll submit a new PR, the previous one is outdated.

nathansobo:
Can you articulate exactly the steps I should take to reproduce any problems with the current implementation? I think I'm not understanding it fully.

jrkoenig:
I tested this with f26c0c0b15231f853f82a8fc6c7e527a27154d89:

1. Create test folder
2. Add project path -> add test folder
3. Close Atom
4. Delete test folder
5. Open Atom. Folder name is now a empty buffer (in addition to the path in tree-view).

This situation arises naturally when you have project paths on remote network drives that aren't always mounted. You would want to preserve these paths, just not load them as buffers.

spec/integration/startup-spec.coffee would be a great place to add a test, but that file doesn't work on my machine, every test fails even on a clean checkout :(

nathansobo:
> Close Atom

I see. I was only reloading. I'll give this a look. Thanks!

nathansobo:
Closing in favor of #11036.

PR #10983, opened 2016-02-26
Update Electron to 0.36.11

drewmnoel:
Pulls in a fix for freetype2 on some systems

PR #11024, opened 2016-03-02
bump up electron + bootstrap versions

z11h:


50Wliu:
What is your rationale behind upgrading Bootstrap?

In addition, there is already an Electron PR open at #10983.

lee-dohm:
Thanks for your contribution @ZakariaRidouh. We're not looking to continue our dependency on Bootstrap. We also like to keep our Electron upgrades separate from other things so that we can test independently.

ZakariaRidouh:
@lee-dohm Sure :smile:

Also may I ask what Bootstrap is currently being used for? If nothing why not remove it completely + other unused dependencies? Thanks!

lee-dohm:
It's being used for certain parts of the UI that we haven't been able to take off of Bootstrap yet.

PR #11026, opened 2016-03-02
move templates to ./github folder to declutter root

z11h:


lee-dohm:
Thanks for contributing and the effort you put into this. We prefer the templates and such remain where they are.

PR #11032, opened 2016-03-02
Use persistence stateKey directly rather than initialPaths

jrkoenig:
Saves and restores application state using the stateKey rather than initialPaths themselves. This avoids having Atom "open" these initialPaths, instead using the deserialized project to restore these paths.

Replaces #10628, see #11023

spec/integration/startup-spec.coffee should have a spec for this, but I couldn't get the tests to run.


nathansobo:
Ah, I see you closed this. Let me know what you think of the other one. Sorry to open a new one. Maybe someday we'll make it easier for maintainers to collaborate on contributor branches directly. You are the author on the commit though.

PR #11036, opened 2016-03-03
Skip deleted directories when restoring application windows

nathansobo:
Refs #10628
Closes #11032

@jrkoenig I understand your original PR #10628 better now and after considering some alternatives, I think your idea of making `initialPaths` explicit is our best option. I cherry-picked that commit and fixed merge conflicts, along with some other small changes that I needed to get a clean restoration experience. Without these changes, I was getting a spurious entry for a parent directory in my project paths.

Can you give this a try and make sure I haven't missed some case that's important to you?

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -21,6 +21,14 @@ describe "Project", ->
     afterEach ->
       deserializedProject?.destroy()
 
+    it "does not deserialize paths to non directories", ->
+      deserializedProject = new Project({notificationManager: atom.notifications, packageManager: atom.packages, confirm: atom.confirm})
+      state = atom.project.serialize()
+      state.paths.push('/directory/that/does/not/exist')
+      state.paths.push(path.join(__dirname, 'fixtures', 'sample.js'))
+      deserializedProject.deserialize(state, atom.deserializers)
+      expect(deserializedProject.getPaths()).toEqual(atom.project.getPaths())
+
     it "does not include unretained buffers in the serialized state", ->
       waitsForPromise ->
         atom.project.bufferForPath('a')
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -676,7 +676,7 @@ class AtomEnvironment extends Model
         @document.body.appendChild(@views.getView(@workspace))
         @backgroundStylesheet?.remove()
 
-        @watchProjectPath()
+        @watchProjectPaths()
 
         @packages.activate()
         @keymaps.loadUserKeymap()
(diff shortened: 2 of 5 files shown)


nathansobo:
@lee-dohm @amytruong You both might be interested in testing how this affects startup. The following steps depend on a directory with two subdirectories, like this:

```
test/
  foo/
  bar/
```

#### Before this PR:
* Open Atom with the two directories: `cd test` then  `atom foo bar`
* Create an untitled buffer and write some text in it
* Quit Atom
* Delete a directory `rm -rf bar`
* Start Atom with no arguments `atom`
* Window restores with the untitled buffer, but *also* an empty buffer named `bar`.

#### After this PR:
* After executing the same sequence of tests as above, we don't create an empty buffer named `bar`

jrkoenig:
This fixes the folders-as-buffers issue I was seeing, and it's a better/less invasive solution than #11032. The filtering of non-existent directories from 47a348 isn't so great for my workflow though.

nathansobo:
How does that impact your workflow?

nathansobo:
screen shot 2016-03-03 at 3 39 44 pm

@jrkoenig This is what happens to me when we don't filter out non-existent directories during project deserialization. Still trying to understand where it's messing with your workflow. Can you explain a bit? I'd like to merge this and move on but don't want to create a new problem for you.

jrkoenig:
I have a bunch of remote repos saved, so it's easy to see what has changes to commit, etc, and some are on different servers that aren't always mounted. It's mildly inconvenient to re-add them all.

I replicated the strangeness you were seeing. Turns out it only does that if the project root is non-existent but the parent dir still exists (if both don't exist it just shows up as an empty directory, but is properly displayed if the folder shows up later -- this is usually what happens with my remote drives because the whole remote path doesn't exist).

I'd say merge this and I'll look into the weirdness later. I still think the "right" thing to do is save the directory but indicate it is missing somehow, so the user can decide to remove it or mount it or whatever, but that's a new day and a new PR.

jrkoenig:
And weirdness explained. When deserializing, a directoryProvider gives us a Directory for the parent directory. Presumably this is so that if you ask for the directory of /path/to/file.txt, you get the directory for /path/to and not something else like a Directory representing /path/to/file.txt. There doesn't appear to be a simple solution that isn't super hacky or changes the behavior of directoryForURISync. See `src/default-directory-provider.coffee`.

nathansobo:
Classic spurious failure in our flaky text editor component tests. But the integration tests have passed twice now. I'm going to merge.

PR #11039, opened 2016-03-03
Add checkbox styling

simurai:
This adds the checkbox styles from the [settings-view](https://github.com/atom/settings-view/blob/master/styles/settings-view.less#L113-L179) to core. Some packages might wanna [use them](https://github.com/atom/settings-view/issues/740) too. Also could be used for the [About](https://github.com/atom/about/pull/12) package.

screen shot 2016-03-03 at 10 40 20 am
## Questions
1. **Selector**? Currently uses `.checkbox[type="checkbox"]`, which means you need to use an `<input type="checkbox" class="checkbox">` to get the styling. Alternatives:
   1. `input[type="checkbox"]`: Then the class is not needed, but it would make it harder to use a custom checkbox. One day we might wanna add a "toggle switch" that also uses `<input type="checkbox">`.
   2. `.checkbox`: Some packages might already use that name in other ways (e.g. settings-view uses it as the wrapper for a checkbox). So a little more risky.
2. **Color**? Currently uses `@background-color-success`. Which is green in most themes. Alternative:
   1. Use `@background-color-info` which is blue in most themes. Would follow OS X default convention, but feel a bit uncanny if you have it set to OS X's grey theme.
## Todo
- [x] Add checkbox styles
- [x] Add `indeterminate` state
## Todo after merge
- Use in `settings-view`
- Use in `about`


simurai:
Added an `indeterminate` state:

indeterminate

simurai:
Is the reason the build is failing the `(lesslint) task`?

```bash
Running "lesslint:src" (lesslint) task
static/checkbox.less (14)
Unqualified attribute selectors are known to be slow. (unqualified-attributes)
>> static/checkbox.less 3: .checkbox[type="checkbox"] {
>> static/checkbox.less 17: &:focus {
>> static/checkbox.less 21: &:active {
>> static/checkbox.less 25: &:before,
>> static/checkbox.less 25: &:before,
>> static/checkbox.less 40: &:before {
>> static/checkbox.less 44: &:after {
>> static/checkbox.less 51: &:checked {
>> static/checkbox.less 53: &:active {
>> static/checkbox.less 56: &:before {
>> static/checkbox.less 61: &:after {
>> static/checkbox.less 68: &:indeterminate {
>> static/checkbox.less 70: &:active {
>> static/checkbox.less 73: &:after {

>> 14 lint errors in 30 files.
Warning: Task "lesslint:src" failed. Use --force to continue.
Error: Task "lesslint:src" failed.

...

Aborted due to warnings.

```

Should we relax the `(unqualified-attributes)` rule? In recent years, selector performance improved a lot and it's now not really an issue anymore.

simurai:
Closing this. _If_ it should go into https://github.com/atom/atom-ui.

PR #11041, opened 2016-03-03
Move to Electron v0.36.9

lee-dohm:
According to @zcbenz, this will prevent crashes on Linux.


lee-dohm:
I did some basic testing, but nothing fancy. @amytruong, let's sync up on this the next time we talk.

lee-dohm:
Whoops, this is a duplicate of https://github.com/atom/atom/pull/10983.

PR #11042, opened 2016-03-03
Use apm from `core.apmPath` config if it exists

kuychaco:
Add the ability to install packages via git urls. See https://github.com/atom/apm/pull/518 and https://github.com/atom/settings-view/pull/743.

Signed-off-by: Michelle Tilley binarymuse@github.com /cc @BinaryMuse


lee-dohm:
Actually, since this is a change to the public API behavior, it should be documented in the doc comments too.

BinaryMuse:
@lee-dohm Updated the doc comment and added a couple specs for good measure

lee-dohm:
Thanks! Looks good to me :rocket:

PR #11051, opened 2016-03-04
Add new item before destroying pending item

kuychaco:
Fixes atom/tabs#278

Signed-off-by: Michelle Tilley binarymuse@github.com


BinaryMuse:
To note: this wouldn't be required in some cases if the pendingness of a tab was observable from `Pane` (like the active item is) but we want to keep the public API smaller if possible.

kuychaco:
@BinaryMuse thanks for adding that context :point_up:

PR #11053, opened 2016-03-04
Ensure atom.cmd --wait correctly waits in Windows cmd & powershell

damieng:
Existing solution caused Command Prompt to exit and sleep command is not built in to Windows.

New solution works by spawning powershell child process that passes it's PID to atom and then waits indefinitely. When atom terminates that process the parent process continues.

Tested in Command Prompt and Powershell in Windows 7 32-bit and Windows 10 64-bit.

### DIFF
--- a/resources/win/atom.cmd
+++ b/resources/win/atom.cmd
@@ -22,30 +22,13 @@ FOR %%a IN (%*) DO (
   )
 )
 
-rem Getting the process ID in cmd of the current cmd process: http://superuser.com/questions/881789/identify-and-kill-batch-script-started-before
-set T=%TEMP%\atomCmdProcessId-%time::=%.tmp
-wmic process where (Name="WMIC.exe" AND CommandLine LIKE "%%%TIME%%%") get ParentProcessId /value | find "ParentProcessId" >%T%
-set /P A=<%T%
-set PID=%A:~16%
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Awesome, love seeing a simpler (and functionally better) solution here :+1: :chart_with_upwards_trend:

joefitzgerald:
:tada:

joefitzgerald:
:shipit:

lee-dohm:
:rocket:

joefitzgerald:
Can we please make sure this gets into the next beta? Will fix CI pipeline stuff, and I may need to make an atom/ci PR on the back of it.

lee-dohm:
@joefitzgerald It'll definitely be in v1.7.x when it hits beta. Or did you mean to ask for this to be pulled in to the next v1.6-beta?

joefitzgerald:
@lee-dohm, I'm asking it to be pulled into 1.6.x beta; beta is currently *broken* in this regard so will impact CI for people doing CI against windows stable, and people using atom on the command line if we don't fix it in beta.

damieng:
I'm all for including this in the 1.6 beta.  The caveat is while it works in Powershell and Command Line it's still broken in Cygwin.

Issue #6956, opened 2015-05-26
load $PATH when launched as an app


willwhitney:
When Atom is launched as an application, it doesn't get the user's PATH, because it was launched by `launchctl`. This means that packages like [script](https://github.com/rgbkrk/atom-script) and [hydrogen](https://github.com/willwhitney/hydrogen) have to tell the user to launch Atom from the command line. Some packages (like hydrogen) can't even be installed without the PATH because their dependencies use native extensions built with [`node-gyp`](https://github.com/TooTallNate/node-gyp). There exist tools such as [`fix-path`](https://github.com/sindresorhus/fix-path), but they clobber the user's PATH even if they've set something on purpose (with e.g. a virtualenv).

Having access to the user's PATH will become more and more important as people build packages that interface more deeply with languages, other tools, and the user's environment. Forcing the user to configure each package with complete paths to the binaries it needs is possible, but inelegant, and becomes intractable with general-purpose tools.

This seems like a clear situation where making this change to Atom itself would dramatically simplify life for many packages and eliminate many bugs arising from poor attempts to capture the PATH.

rgbkrk:
At least from the node runtime, a process can be found on the `launchctl list` filtering out for the `process.pid`:

```
$ launchctl list | grep Atom
83917	-	0x7fcf11d092e0.anonymous.Atom
84086	-	0x7fcf11d0c8d0.anonymous.Atom Helper
83917	-	0x7fcf11e02cd0.anonymous.Atom
84089	-	0x7fcf11d0bcf0.anonymous.Atom Helper
84088	-	0x7fcf11d0ba40.anonymous.Atom Helper
84086	-	0x7fcf11c11930.anonymous.Atom Helper
83919	-	0x7fcf11c0f8a0.anonymous.Atom Helper
```

Here's the `init.coffee` I'm now using (as of seconds ago):

```CoffeeScript
try
  childProcess = require('child_process')

  atomProcs = childProcess.execSync("launchctl list | grep #{process.pid} | grep Atom").toString().split()
  launchRegex = ///^#{process.pid}\t[0\-]\t.*\.Atom///
  matches = (line for line in atomProcs when launchRegex.exec(line))

  if matches.length > 0
    console.log("Fixing paths for Atom")
    process.env.PATH = childProcess.execFileSync('/bin/bash', ['-c', 'echo $PATH']).toString().trim()
[...]

Issue #4126, opened 2014-11-10
Inconsistent environment and behaviour


lloeki:
# Problem

Various tools don't operate correctly depending on the order and the manner with which Atom is started.

Also, persisted state is not restored consistently.
# Cause

Depending the manner in which Atom is started, a different environment is inherited.
- **launchd event**: inherits launchd mostly blank env, which means no shell env vars
- **`atom` in the terminal**: inherits currently set vars

This applied to when the app is _started_, not when a new window is opened, in which case that window will inherit the current process env (which is normal OS X behaviour)

I venture the creation of the `atom` command this way is on purpose to work around the lack of env when run from launchd. But please, dear reader, continue on to the next step.
# Current workaround / Anti-feature

Manually control Atom startup each time you want to use it in a specific environment.

Why is it an anti feature? If you want two projects open at the same time but with different environments, you're busted.

Also, long running sessions and _"wait what env was that started with already?"_ (session can also mean the idiomatic "document != app" OS X way: "Atom.app is started but no windows are opened") .

Also also, frustration quickly sets in when you start it up the wrong way and have to quit and restart halfway through your work and notice your linter does not actually run (manual config does not cut it because 1. it's manual and 2. constantly changing versions means constantly changing path).
# Possible solution
1. wrap Atom.app startup in a login shell (i.e make `Atom.app/Contents/MacOS/Atom` a script that calls whatever the user shell is† as a login, non-interactive one before calling the real `Atom` binary), to inherit global env, so that chruby/rbenv/GOPATH/whatever works.
2. make `atom` command send a signal to `Atom.app` instead of starting atom itself via `Contents/MacOS/Atom`. Basically, `atom foo/bar` has no reason to be different from `open -a Atom foo/bar`. ~~Precisely, it _does not_ inherit the local environment.~~
3. local environment management is bestowed to package implementor (meaning detection of `.git`, `Gemfile`, `.rubocop.yml`, `.gopath`, which they have to do anyway else at least starting via launchd is broken, if not more), not to manual handling by developer.
# Consequences
- `PATH` & al. is properly set and tools like linters can leverage the env to use chruby/rbenv/whatever.
- Clicking the icon, starting via spotlight, opening a file from Finder, using `open -a Atom` and using `atom foo` on the terminal are all consistent, whatever the order they are executed in.
- Everything is now consistent and we can now rejoice as glittery roses spontaneously bloom out of Bose-Einstein condensates.
# Additional note

[...]

PR #11054, opened 2016-03-04
Patch Environment On OSX And Allow A Different Environment Per Window

### LINKED ISSUE #11037: Normalize The Environment And Allow Atom Windows To Have Different Environments
### Prerequisites

* [x] ~~Can you reproduce the problem in [safe mode](https://atom.io/docs/latest/hacking-atom-debugging#check-if-the-problem-shows-up-in-safe-mode)?~~
* [x] Are you running the [latest version of Atom](https://atom.io/docs/latest/hacking-atom-debugging#update-to-the-latest-version)?
* [x] ~~Did you check the [debugging guide](https://atom.io/docs/latest/hacking-atom-debugging)?~~
* [x] ~~Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?~~

### Description

There are some issues with the way the user's environment is made available within Atom:

* On OS X, if a user launches `Atom.app` from `Finder`, `Spotlight`, `Dock`, `open`, or another `launchd` launched process, the user's environment is anaemic and potentially varies from the user's intended environment
* If a user launches `atom` from the command line multiple times, any changes in the environment for attempts `2...n` are not reflected in Atom's environment

I am conflating these two issues because I propose they should be solved by adding a ~~project-level API to `atom.project` (e.g. `atom.project.getEnv()` or `atom.project.env`)~~ functionality to Atom that:
* Is aware of the OS X "anaemic environment" issue and is capable of patching the environment for the implicit project that exists on first launch of Atom, if Atom was indeed launched from `Finder`, `Dock`, `Spotlight`, `open`, et a
* Allows projects to have a varying environment, so that multiple invocations of `atom` from the command line can result in some change in the environment, if changes were made

~~Note that this specifically does *not* endorse changing `process.env` nor does it seek to patch it. The actual environment for the running Atom process should be preserved, unpatched as it is still a valid environment for the original process. This also allows existing functionality to work as it always has, while providing a mechanism for package authors to opt-in to the new project-level API.~~

### Open Questions

* Is `atom.project` the right place for this?
* ~~What "marker" can be used to determine - on OS X - that the environment should be patched (see https://github.com/atom-community/environment/blob/3ccde732f213452e9b897482e1a3fba1f7a4a3e2/lib/environment.js#L113-L120 for prior art)~~ We are going to use the presence of ~~`TERM`~~ `PWD` as a marker on OSX that the environment does not need to be patched
* ~~Should `atom.project.getEnv()` return a promise to permit the patching, or should any patching be done synchronously (e.g. `atom.project.env`)?~~ process.env will be patched directly.
* ~~Should there be lifecycle events that are associated with changes in a project's environment (e.g. `onWillChangeEnv`, `onDidChangeEnv`, etc.)?~~ No.
* I'm certain there are a bunch of issues that this proposal would solve; I'd love help from @50Wliu and @lee-dohm to find them all

### Prior Art

* https;//atom.io/packages/environment
* https://www.npmjs.com/package/consistent-env

### Steps to Reproduce

1. Use OS X
1. Launch Atom from `Finder`, `Spotlight`, `Dock`, `open`
1. Open the Atom console, and type `process.env`
1. Observe the environment is anaemic
1. Quit Atom
1. Launch Atom from the terminal (`atom .`)
1. Open the Atom console, and type `process.env`
1. Observe the environment is as you would expect

**Expected behavior:** The environment is the same, regardless of how Atom is launched on OS X.

untitled 2

**Actual behavior:** The environment is different, depending on how Atom is launched.

untitled

### Versions

All, <= 1.7.x.

### Steps to Reproduce

* Atom does not update its environment when a project is opened twice
  1. Use any operating system
  1. Launch a terminal
  1. Make a directory (e.g. `directory-1`), and change your working directory to it
  1. Set an environment variable called `SOMEVAR` to `true` (e.g. `export SOMEVAR=true` on OS X / Linux; `SET SOMEVAR=true` on Windows)
  1. Launch `atom` from the terminal (e.g. `atom .`)
  1. Observe a new Window launches, with a new project in the tree view, for the `directory-1` directory
  1. Open the Atom console, and type `process.env`
  1. Observe the environment has the `SOMEVAR` key set with a value of `true`
  1. Go back to the terminal, and set the SOMEVAR environment variable to `false` (e.g. `export SOMEVAR=true` on OS X / Linux; `SET SOMEVAR=true` on Windows)
  1. Launch Atom from the terminal (`atom .`)
  1. Observe a new process launches temporarily, but then exits, and the original atom window focuses, with the `directory-1` project open
  1. Open the Atom console, and type `process.env`
  1. Observe the environment has the `SOMEVAR` key set with a value of `true` (although one would expect it to have a value of `false` now)
screen shot 2016-03-02 at 5 58 47 pm
* Atom does not allow different environments for different projects / windows
  1. (Continued from above)
  1. Change your working directory to the parent folder of `directory-1`
  1. Make a directory (e.g. `directory-2`) and change your working directory to it 
  1. Set an environment variable called `SOMEOTHERVAR` to `true` (e.g. `export SOMEOTHERVAR=true` on OS X / Linux; `SET SOMEOTHERVAR=true` on Windows)
  1. Launch `atom` from the terminal (e.g. `atom .`)
  1. Observe a new Window launches, with a new project in the tree view, for the `directory-2` directory
  1. Observe that the existing Window for `directory-1` is still open
  1. Open the Atom console, and type `process.env`
  1. Observe the environment has the `SOMEVAR` key still set to a value of `true`
  1. Observe the environment has the `SOMEOTHERVAR` key undefined
screen shot 2016-03-02 at 5 59 58 pm

**Expected behavior:** There is some way for a package author to get access to the state of the environment as it existed when `atom` was last invoked to launch a Window.

**Actual behavior:** Any changes in the environment for `atom` invocation `2...n` are not available to the package author.

### Versions

All, <= 1.7.x.

/cc @lee-dohm

lee-dohm:
- [X] ~~Create an `atom.project.getEnv` API~~
- [X] Fixup `PATH` on OS X when `TERM` is not set
- [X] Patch `process.env` on startup
- [x] Receive passed in environment variables on subsequent launches

Fixes #11037
Fixes https://github.com/atom/atom/issues/6956


thomasjo:
giphy

src/project.coffee
```diff
@@ -271,9 +274,51 @@ class Project extends Model
...
+        else
+          k = components.shift()
+          v = components.join("=")
+          env[k] = v
+
+    env.PATH
+
+  getShellEnv: ->
+    shell = process.env.SHELL ? "/bin/bash"
+    results = child_process.spawnSync shell, ["--login"], input: "env", encoding: "utf8"
```
steelbrain: I just tried it and it doesn't give accurate terminal results for `$PATH` me on my mac
```
# From terminal
/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/steel/n/bin:/usr/games/bin
```
```
# From executing this
/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin
```

steelbrain: We ran into the same problem when working with linters, we ended up creating [`consistent-path`](https://www.npmjs.com/package/consistent-path) and [`consistent-env`](https://www.npmjs.com/package/consistent-env). Both are very battle tested and used by almost all of the linter providers.

`consistent-path` is a `consistent-env` dependent package which provides just the PATH, but looks like you could use `consistent-env` here

thomasjo: @lee-dohm If you replace the `--login` flag with `--interactive` this should work as intended :sparkles:

spec/fixtures/sample.txt
```diff
@@ -1 +1 @@
-Some textSome textSome text.
+Some textSome textSome textSome text.
```
thomasjo: This change seems unintentional. Revert?

src/project.coffee
steelbrain: @thomasjo That would break linux because Atom is spawned on linux with `nohup`, see https://github.com/steelbrain/nohup-shell-hanging-bug

joefitzgerald: @steelbrain the shell launching should not be done on any platform other than OS X. Consequently, nothing should break on Linux.

steelbrain: @joefitzgerald There are several pairs of Desktop Environments and shells on Linux that do not supply ENV correctly to apps not run from terminal, most notably GNOME Shell and ZSH. Therefore we do need to fix it there as well

joefitzgerald: Well we don't need to in this PR; this PR is intended to solve for the OS X issue, and add a project level API for the environment.

That doesn't preclude more cases beyond OS X; can you please open up (or link) an issue for each additional scenario, describing steps to reproduce, and desired behavior?

thomasjo: @steelbrain Correct me if I'm wrong, but that has nothing to do with the DE nor the shell. What you're describing would be distribution-specific configuration problems. There are lots of silly *nix distributions out there with horrible defaults.

But if this truly is a problem (in particular the HUP scenario described earlier), then we'll just have to deal with it.

thomasjo: > Well we don't need to in this PR; this PR is intended to solve for the OS X issue, and add a project level API for the environment.

Agreed! We can patch other problems at a later point, once we can reliably reproduce the anecdotal reports.

steelbrain: @joefitzgerald @thomasjo What do you guys have against the npm modules I suggested? It takes care of all the scenarios and fixes this for everyone not just the handful few

spec/fixtures/sample.txt
lee-dohm: Thanks, yeah ... not sure why that happens on my machine :angry:

mnquintana: Do we have a spec that inadvertently edits that fixture? :eyes:

lee-dohm: I suspect we do, yes.

lee-dohm:
This is still not complete but I wanted to push up what I have so far. I decided to include patching `process.env` on startup per @joshaber's suggestion on #11037 (though the approach I chose is probably wrong). I also did a bit more reading on #4126 (and there's still #6956 that I haven't caught up on in a while). There are a lot of nuances that will need to be covered for this to be acceptable. I'm going to keep plugging away at it :grinning:

src/atom-environment.coffee
```diff
@@ -229,6 +229,10 @@ class AtomEnvironment extends Model
 
     checkPortableHomeWritable()
 
+    # Patch the `process.env` on startup to fix the problem first documented
+    # in #4126
+    process.env = @project.getEnv()
```
joefitzgerald: This is going to make it less clean, but this should happen prior to packages being loaded. I.e. prior to @project being defined. Think of all the activities that packages may do on activation that they _should not be doing_ (but still are doing)n that are dependent on `process.env`.

joefitzgerald: If you move the patching logic such that it is executed in the first atom-environment, then maybe `@project.getEnv()` could be renamed to be `@project.env`. Perhaps we could also externalize the logic into a file that is only required if patching is needed (`javascript` FTW?).

lee-dohm: Thanks for the ideas. I'll take a deeper look at this later tonight or tomorrow.

src/project.coffee
```diff
@@ -271,9 +274,66 @@ class Project extends Model
...
+  #
+  # On OS X, the `PATH` can be different depending on whether Atom is launched
+  # from the Dock, Finder, Spotlight or the terminal. This detects how Atom was
+  # started and corrects the `PATH` environment variable before returning a copy
+  # of the environment.
+  getEnv: ->
+    unless @env?
+      @env = _.clone(process.env)
+      if process.platform is "darwin" and not process.env.TERM?
+        shellPath = @getShellPath()
```
joefitzgerald: We shouldn't just be setting the path; we should swap out the entire environment.

thomasjo: > We shouldn't just be setting the path; we should swap out the entire environment.

I agree fully with this. The goal here should be to make sure all launch experience are equivalent.

src/atom-environment.coffee
lee-dohm: > then maybe `@project.getEnv()` could be renamed to be `@project.env`

That would lose us the current feature that packages can't muck up the environment for other packages (at least through the public API). Right now, everyone gets a copy and then they can do what they want with their copy.

I'll definitely move the first initialization further up though.

joefitzgerald: Yeah, fair enough on making it a function:

* `project.getEnv()`?
* `project.env()`?
* `project.environment()`?

lee-dohm: Looking deeper, it appears that packages aren't loaded until `AtomEnvironment.startEditorWindow`:

https://github.com/atom/atom/blob/ld-osx-environment/src/atom-environment.coffee#L674

So doing the patching of the environment in the `AtomEnvironment` constructor does make it happen before packages are loaded.

src/project.coffee
lee-dohm: Done.

src/atom-environment.coffee
lee-dohm: Personally, I was sticking with just "env" as the main part of the name to make it familiar to Node developers and `process.env`. I added "get" to the front to make it clear it is a function rather than just an attribute (and all of the other public APIs have a verb in the front). I can totally be swayed though.

src/environment.coffee
```diff
@@ -0,0 +1,41 @@
+child_process = require('child_process')
```
joefitzgerald: Why are we creatomg new files in `.coffee`? Want me to convert to `.js`?

thedaniel: yeah this should probably be js

nathansobo:
I don't think we should make a new API for fetching environment variables. I think we should extend the existing API so it works correctly despite our multi-process setup, which shouldn't be that relevant to Atom users. The only justifications I could see for inventing something new would be the following:

1) Someone might write into `process.env`. :shrug: People can always write places they shouldn't. Doesn't justify building a new construct in my mind.
2) People don't expect `process.env` to have these particular variables. I think we can reinterpret a bit since we own the platform.

joefitzgerald:
Yep @nathansobo i've already made the requisite changes :) Pushing the commits shortly.

spec/environment-spec.js
```diff
@@ -0,0 +1,161 @@
...
+describe('Environment handling', () => {
+  let originalEnv
+  let options
+
+  beforeEach(() => {
+    originalEnv = process.env
+    delete process._originalEnv
+    options = {
+      platform: process.platform,
+      env: _.clone(process.env)
```
nathansobo: `Object.assign`?

src/atom-environment.coffee
```diff
@@ -4,6 +4,7 @@ path = require 'path'
 
 _ = require 'underscore-plus'
 {deprecate} = require 'grim'
+environment = require('./environment')
```
nathansobo: This is a weird name to me... what do you think of `environmentHelpers` or something?

lee-dohm: Works for me, I'll change it.

spec/environment-spec.js
lee-dohm: I'll try it out. I haven't used it before. If it works the same, I'll update it.

joefitzgerald:
:tada: :joy_cat:

basnijholt:
Finally 👍😄

sbromberger:
Apparently this didn't make 1.6.0. Will we see it in 1.6.1? (Fingers crossed....)

joefitzgerald:
It's in the beta channel. You can run stable and beta side by side. atom.io/beta

sbromberger:
Ah, thanks. I hesitate to run beta since Atom is now a huge part of my workflow and I don't really want to disrupt / switch between things. If I understand the blog post, this feature will be merged into -stable at some point. Since it's in -beta now, does that mean 1.6.1?

thomasjo:
> Since it's in -beta now, does that mean 1.6.1

@sbromberger 1.7.0 (unless it's reverted for whatever reason).

lierdakil:
> Should -i be used when spawning the shell?

I found that many Linux/OSX users tend to customize their environment in rc file (`.bashrc`/`.zshrc`), which only works for *interactive* shells. This is certainly not a preferred way, but it's used by many people nonetheless. Not all of them understand differences between login and interactive shell, and not all of them care.

It's true that running interactive shell can lead to problems. But running login shell can too. For example, I run `ssh-agent` from `.profile` -- not something you want to do when starting Atom (I'm on Linux though, so make of it what you will)

ioquatix:
Hey, since I tried to solve this problem in [script-runner](https://atom.io/packages/script-runner) I thought I'd chime in here.

Firstly, for people running things like RVM, or the equivalent for python, you definitely want to get the environment from an interactive login shell, or you won't get the right things.

We tried tied to extract this into a separate module, called [shell-environment](https://www.npmjs.com/package/shell-environment).

We also had some issues on Linux where things like DISPLAY were not getting set and this might be unintuitive for users who are expecting things to "just work".

If you look at the approach we took in shell-environment, you'll see we avoided many of the issues of using an interactive shell by piping the results of `env` back through a custom pipe.

gnestor:
I have discovered another minutia missing from this patch. Unless Atom is launched from the terminal, `process.env.PWD` is set to '/'. If `env` (bash) is run as a child/buffered process in Atom, PWD is correctly set to my project's root directory. Therefore, it appears like the method used in this PR to get the shell environment is failing for PWD. It's worth mentioning that PWD can be inferred from `atom.workspace.project.getPaths()[0]`. @lee-dohm What do you say?

lee-dohm:
@gnestor I think `PWD` is set correctly. It is set by the environment that Atom is launched within. The goal of this fix isn't to make the environment identical to what it would be if it were launched from the terminal at some pre-determined location. It is to import the user's standard profile. This change achieved that.

gnestor:
@lee-dohm Ok so it sounds like this is a separate issue. Do you agree though that Atom should correctly set `process.env.PWD` to the project's root (e.g. `atom.workspace.project.getPaths()[0]`) vs. "/" when Atom is launched outside of the terminal? If so, I will open a new issue.

lee-dohm:
@gnestor No, I don't agree that `PWD` should be set to the project's root. The concept of this change wasn't to make it as if someone launched from the project directory but to import the user's standard environment (as much as possible).

driskell:
In most of 1.12 betas (I'm on 1.12.0-beta7 at the moment on Sierra 10.12) there's a problem with new windows.

If you open from OS X dock and inspect process.env, all is fine and wonderful in the world/window. When you go to File -> New Window, and inspect process.env on the new window, it does not include any of the environment variables, anything from login is gone. I haven't had change to inspect things but seems might be an issue with inheriting environment when creating new windows. I can't find a related issue to this either.

thomasjo:
I can reproduce the behavior described by @driskell. Unsure whether this is a regression or not — was this specific workflow tested/fixed? /cc @lee-dohm @atom/feedback

driskell:
It looks like the new-window events etc don't pass in the current `env` to the `openPaths` (https://github.com/atom/atom/blob/e5cfc6b6e4b36b6c443c0526bfb4c816c666b0f2/src/main-process/atom-application.coffee#L210) so it doesn't do any environment update, unlike with the first window opening (or window opened from command line), which does. It also means dragging a folder onto dock to open a project means you get no env.

I'll raise a new issue for it this week, maybe even attempt a fix.

lee-dohm:
@driskell Have you filed an Issue on this behavior?

flujan:
Same issue here.

joefitzgerald:
@flujan have you updated to the latest version of atom? A fix for the issue described by @driskell was included in the latest patch.

flujan:
Yes, when I start ATOM from the terminal, it works. When I try to launch from the Dock it does not have the Path configured.

joefitzgerald:
@flujan https://github.com/atom/atom/pull/13200 your package(s) may need to be updated to make use of the `core:loaded-shell-environment` activationHook. Which package(s) are complaining abou the path?

flujan:
Guys, the Iterm2 plugin was causing the bug. Thanks.

PR #11057, opened 2016-03-04
Move Pane::addItem 'pending' option to options object

BinaryMuse:
The API for adding pending pane items is pretty verbose; combined with ddb08d0c46ccf6408c5e7618fedd6cfc93bbad50, the argument list to `Pane::addItem` was getting pretty long, and it's easy to imagine it getting longer. This collapses things into an options object.

/cc @nathansobo re: the commit above; since the part of the API that changed was never publicly documented I assume it's fine to change for 1.6.

/cc @kuychaco the pending-ness :crown:

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -404,7 +404,7 @@ class Pane extends Model
     @setPendingItem(item) if pending
 
     @emitter.emit 'did-add-item', {item, index, moved}
-    @destroyItem(lastPendingItem) if lastPendingItem?
+    @destroyItem(lastPendingItem) if lastPendingItem? and not moved
     @setActiveItem(item) unless @getActiveItem()?
     item
 
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -348,16 +348,17 @@ class Pane extends Model
   # Public: Make the given item *active*, causing it to be displayed by
   # the pane's view.
   #
-  # * `pending` (optional) {Boolean} indicating that the item should be added
-  #   in a pending state if it does not yet exist in the pane. Existing pending
-  #   items in a pane are replaced with new pending items when they are opened.
-  activateItem: (item, pending=false) ->
+  # * `options` (optional) {Object}
+  #   * `pending` (optional) {Boolean} indicating that the item should be added
+  #     in a pending state if it does not yet exist in the pane. Existing pending
+  #     items in a pane are replaced with new pending items when they are opened.
+  activateItem: (item, options={}) ->
     if item?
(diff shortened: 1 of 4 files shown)


nathansobo:
@BinaryMuse Could you also add an `index` option and just make the options object the second argument? A `typeof` check could be used to support passing an integer for backward compatibility.

BinaryMuse:
@nathansobo Push a commit to implement this. Mind taking a quick :eyes::question:

src/pane.coffee
```diff
@@ -348,30 +349,49 @@ class Pane extends Model
...
+  #     in a pending state if it does not yet exist in the pane. Existing pending
+  #     items in a pane are replaced with new pending items when they are opened.
+  activateItem: (item, options={}) ->
     if item?
       if @getPendingItem() is @activeItem
         index = @getActiveItemIndex()
       else
         index = @getActiveItemIndex() + 1
-      @addItem(item, index, false, pending)
+      @addItem(item, _.extend({index: index}, options))
```
nathansobo: Can we use `Object.assign`? I think we have it now. Hate underscore & friends.

src/pane.coffee
```diff
@@ -348,30 +349,49 @@ class Pane extends Model
...
+  #     If omitted, the item is added after the current active item.
+  #   * `pending` (optional) {Boolean} indicating that the item should be
+  #     added in a pending state. Existing pending items in a pane are replaced with
+  #     new pending items when they are opened.
   #
   # Returns the added item.
-  addItem: (item, index=@getActiveItemIndex() + 1, moved=false, pending=false) ->
+  addItem: (item, options, args...) ->
+    # Backward compat with old API:
+    #   addItem(item, index=@getActiveItemIndex() + 1, moved=false, pending=false)
```
nathansobo: I'm okay *only* being backward with the API on stable if that simplifies this at all.

nathansobo:
Couple comments. Would love to drop that extra parameter if we can get away with it. I think we can based on the previous public stable API.

src/pane.coffee
BinaryMuse: We went this direction but `extend` was being used in the rest of the file, so we kept it for parity's sake. Maybe we can make a pass to replace them all at once or something?

src/pane.coffee
BinaryMuse: :+1: removed this in the latest commit. thanks

nathansobo:
As for `_.extend`, makes sense. I didn't realize there were other instances in the file.

PR #11060, opened 2016-03-04
Serialize MarkerLayers only on quit

### LINKED ISSUE #10966: Uncaught TypeError: Converting circular structure to JSON

1. . Just editing javascript files.  

**Atom Version**: 1.7.0-dev-5655518
**System**: Mac OS X 10.11.2
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Converting circular structure to JSON

```
At /Applications/Atom.app/Contents/Resources/app.asar/src/state-store.js:29

TypeError: Converting circular structure to JSON
    at Object.stringify (native)
    at StateStore.save (/Applications/Atom.app/Contents/Resources/app.asar/src/state-store.js:29:26)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/atom-environment.js:947:140
```

### Commands

```
  3x -1:03.6.0 core:move-left (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -1:02.7.0 core:move-right (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -1:00.4.0 core:save (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:49.9.0 core:copy (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:49.5.0 project-find:show (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:49.3.0 core:paste (atom-text-editor.editor.mini.is-focused)
     -0:49.2.0 core:confirm (atom-text-editor.editor.mini.is-focused)
     -0:47.5.0 core:cancel (ol.results-view.list-tree.focusable-panel.has-collapsable-children)
     -0:43.2.0 core:copy (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:42 core:paste (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
  8x -0:37.8.0 core:backspace (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:22 core:select-right (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
  3x -0:21.7.0 core:backspace (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:13.7.0 core:copy (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:11.4.0 core:paste (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
     -0:10.3.0 core:save (atom-text-editor.editor.vim-mode.insert-mode.is-focused)
```

### Config

```json
{
  "core": {
    "ignoredNames": [
      ".git/**/*",
      "coverage/**/*",
      "*.min.*",
      "*.scssc",
      "report/files/**/*",
      "node_modules/**/*",
      "node_modules/**/*"
    ],
    "themes": [
      "seti-ui",
      "atom-dark-syntax"
    ],
    "projectHome": "/Users/NickCapito/work",
    "disabledPackages": [
      "language-shellscript",
      "language-ruby-on-rails",
      "language-ruby",
      "language-python",
      "language-php",
      "language-perl",
      "language-objective-c"
    ]
  },
  "editor": {
    "invisibles": {},
    "showInvisibles": true,
    "showIndentGuide": true,
    "scrollPastEnd": true,
    "fontSize": 13
  }
}
```

### Installed Packages

```coffee
# User
angularjs, v0.3.4
atom-beautify, v0.28.22
atom-ternjs, v0.13.2
editor-stats, v0.17.0
language-csharp, v0.11.0
language-powershell, v2.1.0
linter, v1.11.3
linter-csslint, v1.3.1
linter-jshint, v2.0.2
node-debugger, v1.4.3
seti-ui, v0.8.1
vim-mode, v0.64.0
wakatime, v5.0.2

# Dev
No dev packages
```

as-cii:
Depends on https://github.com/atom/text-buffer/pull/136.

Closes #10966.

After #10605 we have noticed some slowdowns during multi-cursor edits which we have tried to address in #10925 by switching from structured cloning to `JSON.stringify` when we had to serialize state. Even though that was faster, it led to several issues like stringifying cyclic structures and taking more and more time as the `History` grew over time.

With this PR we still continuously save the `AtomEnvironment`'s state, but we try to address the above concerns by: 
1. Incrementally converting history changes to a binary format. (https://github.com/atom/atom-patch/pull/7)
2. Serialize marker snapshots only before quitting. Although this won't maintain selection after hard crashes, the user history will always be preserved.

The result is the following: 

screen shot 2016-03-04 at 17 43 44
screen shot 2016-03-04 at 17 44 03

(**Total:** `~ 40ms`, benchmarks performed by simultaneously writing 5 characters on 1500 lines)

Those `~ 34ms` will only be payed the first time new history entries need to be serialized, because we will reuse a cached binary representation of them whenever they need to be serialized again.

As a byproduct, and thanks to history's internals restructuring, undos/redos with many overlapping changes will get faster, as we compose and merge them without repeating some of the work for the intersecting regions.

It would be awesome to have some feedback on this, especially to understand how this feels to you performance-wise after a test-drive.

What's missing:
- [x] Bump text-buffer after merging the related PR.

/cc: @atom/core


src/atom-environment.coffee
```diff
@@ -817,12 +817,15 @@ class AtomEnvironment extends Model
 
     @blobStore.save()
 
-  saveState: ->
+  saveState: (options) ->
```
as-cii: In general, I think the `isQuitting` property could be passed also to packages as well, so that they can make a more informed decision (as we do) on how to perform the serialization. What do you think? :thought_balloon:

src/project.coffee
```diff
@@ -65,15 +65,15 @@ class Project extends Model
           fs.closeSync(fs.openSync(bufferState.filePath, 'r'))
         catch error
           return unless error.code is 'ENOENT'
-      deserializerManager.deserialize(bufferState)
+      TextBuffer.deserialize(bufferState)
```
as-cii: Here we can get rid of the `DeserializerManager` because we have removed the `Serializable` mixin in `TextBuffer` on https://github.com/atom/text-buffer/pull/136

src/state-store.js
```diff
@@ -52,9 +49,8 @@ class StateStore {
 
         request.onsuccess = (event) => {
           let result = event.target.result
-          if (result) {
-            // TODO: remove this when state will be serialized only via JSON.
-            resolve(result.isJSON ? JSON.parse(result.value) : result.value)
+          if (result && !result.isJSON) {
```
as-cii: Even if we haven't shipped this to beta/stable yet, I kept the `isJSON` conditional just to make sure that when rebuilding master there is no hard crash also for our users on the "dev channel".

as-cii:
The impact of long histories on IndexedDB seems to be minimal in terms of flushing the state to disk:

screen shot 2016-03-07 at 10 50 39
screen shot 2016-03-07 at 10 50 48

I believe this makes a lot of sense now that we encode histories as byte buffers: 

* Structured cloning doesn't need to traverse nested/long objects/arrays and just needs to copy bytes, which is a very fast :zap: operation.
* Storing the buffer clone in the file system should be fast, because ideally no additional operation needs to be done on the byte array.

I am going to :ship: this once it goes :green_heart:, so that we can start test driving it :car: on dev and beta.

nathansobo:
Nice work here.

PR #11070, opened 2016-03-05
store disabled packages in deterministic order

dirk-thomas:
Follow up of #9948. Storing the list of disabled package names in a deterministic order.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1104,6 +1104,17 @@ describe "Config", ->
         expect(atom.config.get("foo.bar.baz")).toEqual ["a", "b"]
         expect(observeHandler).toHaveBeenCalledWith atom.config.get("foo.bar.baz")
 
+    describe ".insertInOrderAtKeyPath(keyPath, value)", ->
+      it "insert the given value to the array in alphabetic order at the key path and updates observers", ->
+        atom.config.set("foo.bar.baz", ["b"])
+        observeHandler = jasmine.createSpy "observeHandler"
+        atom.config.observe "foo.bar.baz", observeHandler
+        observeHandler.reset()
+
+        expect(atom.config.insertInOrderAtKeyPath("foo.bar.baz", "a")).toBe true
+        expect(atom.config.get("foo.bar.baz")).toEqual ["a", "b"]
+        expect(observeHandler).toHaveBeenCalledWith atom.config.get("foo.bar.baz")
+
     describe ".unshiftAtKeyPath(keyPath, value)", ->
       it "unshifts the given value to the array at the key path and updates observers", ->
         atom.config.set("foo.bar.baz", ["b"])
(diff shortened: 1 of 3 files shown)


dirk-thomas:
I added a spec for the new function.

lee-dohm:
What happens if this gets out of order outside of Settings View? Will Settings View reorder it? For example:

1. Launch Atom
1. Open Settings View
1. Disable some packages
1. Close Settings View
1. Open `config.cson` and ensure the list of disabled packages is ordered
1. Open the Terminal and enter the command `apm disable aaapackage-name`

* What's the expected behavior in the above scenario?
* Does a change need to be made to apm as well?
* Should Settings View or `Config` simply reorder things on write?

dirk-thomas:
The new function makes sure that the list is in alphabetic order after a package name is inserted. While the config could always reorder on write I didn't do that since I though ordering this list only when it is being modified is more efficient and less intrusive.

I don't know anything about `apm` and how it interacts with this config file.

as-cii:
I believe we need to support the really good point @lee-dohm brought up, so that we can make the disabled packages list consistent. We might always order things in the settings-view, but I think it's nice to see the disabled packages ordered even when reading the textual version of `config.cson`.

One thing we might do is having a configuration schema that enforces ordered values on a specific keypath: this is something other packages can use as well, and we would use it for the schema of `disabledPackages`. For instance:

```cson
disabledPackages:
  type: 'sorted-array'
  default: []
  items:
    type: 'string'
  description: 'List of names of installed packages which are not loaded at startup.'
```

Upon load or write of `disabledPackages`, we enforce this schema and sort lexicographically its entries. We might also allow more complex objects to be sorted as well:

```cson
something:
  type: 'sorted-array'
  default: []
  items:
   sort: 'foo.bar.baz'
   type: 'object'
```

As for `apm disable` I can see three different solutions:

1. Let `apm` store the disabled package in the wrong order. Then, when opening Atom, the correct order is enforced;
2. Make Atom and `apm` share the `Config` class somehow, so that we can change the configuration and enforce the same constrains on both;
3. Sort entries when settings the `disabledPackages` key path.

It's probably ok to just go with 1), but 2) seems worth considering as well.

@dirk-thomas: if you're interested in implementing this and @atom/core agrees is a good idea, I can provide additional guidance and support. :+1:

dirk-thomas:
@as-cii I was focused on other project recently but I would like to pick this topic up again. If you could provide my with a few pointer I am happy to work on patches adding support for ordered arrays.

dirk-thomas:
I just asked on Slack and @lee-dohm mentioned that they will talk about the option to add a new type ("ordered-array") in one of the next meetings.

dirk-thomas:
I heard back from @lee-dohm and the Atom team doesn't want to maintain this additional functionality even if all the missing pieces are contributed through PRs. While I would like to see this feature being accepted I understand why they decided this way :neutral_face: 

Therefore I will close this PR.

Issue #10719, opened 2016-02-06
ATOM_DEV_RESOURCE_PATH needs to be set to run specs


PaulAik:
Having some trouble running the test specs on OSX on version 1.6.0-dev-17bf2b4 (latest from master).

The command `atom --test spec` produces:

```
Downloading Update: Could not get code signature for running application
[26691:0206/133702:INFO:CONSOLE(85)] "Error: Failed to execute 'registerElement' on 'Document': Registration failed for type 'atom-styles'. A type with that name is already registered.
  at Error (native)
  at Object.<anonymous> (/Users/paulaikman/dev/Git/external/unchangedAtom/src/styles-element.coffee:117:43)
  at Object.<anonymous> (/Users/paulaikman/dev/Git/external/unchangedAtom/src/styles-element.coffee:1:1)
  at Module._compile (/Applications/Atom.app/Contents/Resources/app.asar/src/native-compile-cache.js:103:30)
  at Object.defineProperty.value [as .coffee] (/Applications/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:208:21)
  at Module.load (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/coffee-script/lib/coffee-script/register.js:45:36)
  at Function.Module._load (module.js:310:12)
  at Module.require (module.js:365:17)
  at require (/Applications/Atom.app/Contents/Resources/app.asar/src/native-compile-cache.js:50:27)
  at Object.<anonymous> (/Users/paulaikman/dev/Git/external/unchangedAtom/src/atom-environment.coffee:12:17)
  at Object.<anonymous> (/Users/paulaikman/dev/Git/external/unchangedAtom/src/atom-environment.coffee:1:1)
  at Module._compile (/Applications/Atom.app/Contents/Resources/app.asar/src/native-compile-cache.js:103:30)
  at Object.defineProperty.value [as .coffee] (/Applications/Atom.app/Contents/Resources/app.asar/src/compile-cache.js:208:21)
  at Module.load (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/coffee-script/lib/coffee-script/register.js:45:36)
  at Function.Module._load (module.js:310:12)
  at Module.require (module.js:365:17)
  at require (/Applications/Atom.app/Contents/Resources/app.asar/src/native-compile-cache.js:50:27)
[...]

PR #11072, opened 2016-03-05
Modify selection to stop highlight at new-line (instead of continuing to end-of-screen)

mehcode:
Continued work by @vjeux from https://github.com/atom/atom/pull/8640

It's mostly the same as what @vjeux had but I further extend the selection to select the new line (as it does in sublime and vs code).

---

Some general direction on how to go from there and I don't mind going through the specs and adjusting.

Now.. I'm not sure how to get the rounded corners that VS Code and Sublime have on the selection. I tried playing with border-radius and it sort of works but we need these triangles to fill in the bottom-right border radius (but only if the line is shorter than the line under it). An idea I had is we could perhaps switch to constructing a single SVG Path that is filled for selection. Perhaps @simurai knows a creative method to make it work in CSS. If there is a reasonable method to do this with CSS; the rounded selection could be left up to a theme to implement (and leave corner with pointy corner selection).

---

A question before I delve too much deeper. Is this something that would be accepted if we figure out a good solution? @nathansobo @benogle

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -18,7 +18,7 @@ class TextEditorPresenter
     {@contentFrameWidth} = params
 
     @gutterWidth = 0
-    @tileSize ?= 6
+    @tileSize ?= 1
     @realScrollTop = @scrollTop
     @realScrollLeft = @scrollLeft
     @disposables = new CompositeDisposable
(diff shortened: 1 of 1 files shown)


50Wliu:
@mehcode for your spec failures you might want to take a look at this issue: #10719 (specifically [this comment](https://github.com/atom/atom/issues/10719#issuecomment-184716688)).

mehcode:
Thanks @50Wliu; was having the exact problem.

nathansobo:
@mehcode can you post a screenshot to illustrate what you have in mind?

nathansobo:
Oh I see some screenshots in the referenced PR. It's funny. We worked pretty hard to go full-width with the selection because that's how TextMate and some other OS X apps looked. Can you talk about why you prefer it to be the other way? I'd prefer not to rock the boat unless there's a good reason.

simurai:
You could overlap 2 pseudo elements and then add a border-radius to one, like:

```less
atom-text-editor::shadow .selection .region {
  border-top-right-radius: .33em;

  &:before,
  &:after {
    content: "";
    position: absolute;
    top: 0;
    bottom: 0;
    right: -1em;
    width: 1em;
  }
  &:before {
    background-color: @syntax-selection-color;
  }
  &:after {
    background-color: @syntax-background-color;
    border-bottom-left-radius: .33em;
  }
}
```

screen shot 2016-03-09 at 10 31 03 pm

But you would need some extra elements if there is more than one corner. SVG might work too, yeah. I guess most important is to not slow down performance. Could this be done as a separate package to test first or does it need to be implemented in core?

jeancroy:
> Can you talk about why you prefer it to be the other way?

On sublime I'd use the close selection feature to show trailing white-space.
Sublime also has a default config of show invisible on selection. 

Here's a screenshot of the same text on both editor, see that for someone interested in white space sublime show a lot more useful informations.

screenshot_2

screenshot_3

nathansobo:
@simurai I think this would be pretty tough to do in a package with current APIs. With clever interaction with the scheduler it might be possible, but it's probably something we should consider changing in core if there's popular demand. Assuming we could apply any CSS classes you wanted, any ideas for creating the desired rounding effect without resorting to SVG?

alexandernst:
@jeancroy I have completely forgotten about ST's "show invisible chars only in selection", which is **SO** useful and makes the editor look so clearer.

From a user POV, I think there should be 2 settings for this:

1. Enable/disable this (not everybody will like this feature)
2. Enable/disable showing invisible chars only on selection.

mehcode:
@jeancroy That's a great point.

Trailing whitespace is a lot easier to see with sublime / VS Code / Visual Studio / Eclipse selection highlighting.

My other point was going to be all other editors do it this way (even chrome text areas). But apparently TextMate does not. That seems to be an aberration.

I'd don't have another point. I was thinking this was an oversight as most editors I look at do it one way and I guess TextMate does it another.

---

Here is how Atom looks on this branch. Slightly different than how it looks on the linked PR.

selection_014

jeancroy:
Does using class for rounding means to conditionally apply classes depending of the length of previous / next line ?

screenshot_4

simurai:
> @jeancroy Does using class for rounding means to conditionally apply classes depending of the length of previous / next line ?

Yes, it would need to check if the previous and next line is longer or shorter and then add the a class(es) for it. Since there could be an inverted corner left and right, each `.selection .region` could have two extra child elements.

```html
<div class="highlight selection">
  <div class="region">
    <div class="corner-left"></div>
    <div class="corner-right"></div>
  </div>
</div>
```

Then have some classes. Not sure how to call them, maybe:
```html
<div class="region has-corner-top-left has-corner-top-right has-corner-bottom-right--outwards">
```

- `.has-corner-top-left` or `.has-corner-top-left--outwards`
- `.has-corner-top-right` or `.has-corner-top-right--outwards`
- `.has-corner-bottom-left` or  `.has-corner-bottom-left--outwards`
- `.has-corner-bottom-right` or `.has-corner-bottom-right--outwards`

Then core or themes could apply the right `border-radius`.

PR #11078, opened 2016-03-06
Add TextEditors to the registry only when opting in

wvanlint:
This PR makes it so `TextEditors` are added to the registry only when they explicitly opt in.
This was mentioned in atom/atom#10851 (where I forgot to include it) to avoid leaking memory due to package authors not calling `destroy`.

/cc @nathansobo

### DIFF
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -542,7 +542,10 @@ class Workspace extends Model
         throw error
 
     @project.bufferForPath(filePath, options).then (buffer) =>
-      @buildTextEditor(_.extend({buffer, largeFileMode}, options))
+      editor = @buildTextEditor(_.extend({buffer, largeFileMode}, options))
+      disposable = atom.textEditors.add(editor)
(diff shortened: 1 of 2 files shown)


src/workspace.coffee
```diff
@@ -565,8 +565,9 @@ class Workspace extends Model
       @grammarRegistry, @project, @assert, @applicationDelegate
     }, params)
     editor = new TextEditor(params)
-    disposable = atom.textEditors.add(editor)
```
nathansobo: I would rather just require the editor to be explicitly registered rather than having this extra parameter.

nathansobo:
Maintaining registration across serialization/deserialization is a good idea, but I'd prefer not to add a parameter to `buildTextEditor` when someone can just call register *and* get a disposable back that would unregister.

wvanlint:
Makes sense, I removed the parameter to `buildTextEditor`.

nathansobo:
:cherry_blossom: :arrow_right: :cherries:

PR #11089, opened 2016-03-07
Update nodegit

joshaber:
Fixes https://github.com/atom/git-diff/issues/88 and https://github.com/atom/git-diff/issues/89.


joshaber:
(Once this lands I'll cherry-pick it into beta.)

PR #11096, opened 2016-03-08
Add semicolons to README.MD

alex-teren:


mnquintana:
Thanks for contributing! However, since these items are already in a bulleted list, they don't need semicolons.

PR #11097, opened 2016-03-08
Make atom.sh work in Cygwin

andlrc:
Node cannot figure out to resolve atom.js while running though cygwin:

```
$ atom --wait .
module.js:340
    throw err;
          ^
Error: Cannot find module 'C:\cygdrive\c...'
```

But if we change current working directory to `$directory` then node will be able to resolve atom.js


damieng:
Hi @ginkoms I'm not able to reproduce this and I don't see how the fix changes behavior as $directory prefixed atom.js. About the only scenario I could see would be if $directory had a trailing slash.

Where exactly are you seeing the error and what version of atom are you using?

damieng:
I'm actually about to entirely replace this script with a call to the powershell one in PR https://github.com/atom/atom/pull/11103 so will close this one.

PR #11099, opened 2016-03-08
Add zero to hexadecimal numbers below F (16)

### LINKED ISSUE #712: Color picker returns invalid HEX string.
When using the color picker in a package's settings, the `.toHexString()` method sometimes returns invalid HEX values for dark colors (close to black).

I resorted to catching those invalid colors using `tinycolor`'s `isValid()` method…

```javascript
atom.config.onDidChange('atom-material-ui.colors.accentColor', (value) => {
    var accentColor = tinycolor(value.newValue.toHexString());

    if (!accentColor.isValid()) {
        return atom.notifications.addError(`The selected color <code>${value.newValue.toHexString()}</code> is not a valid HEX value. Try another!`, { dismissable: true });
    }
    amu.writeConfig()
});
```

Just to be clear, I'm not talking about the [color-picker](https://atom.io/packages/color-picker) package. I'm talking about the color picker Atom gives us developers for our package's settings:

```javascript
"baseColor": {
    "title": "Primary color",
    "description": "Changes the main theme color.",
    "type": "color", // <-- This one!
    "default": "#009688"
},
```

Looks like this on OS X:

captura de pantalla 2015-12-13 a las 2 58 24

When changed, the `value.newValue` property is a `Color` object:

captura de pantalla 2015-12-13 a las 3 04 38

And its `toHexString()` method *sometimes* produces invalid HEX color strings. It happens with dark colors mostly.

rbottema:
Should fix https://github.com/atom/settings-view/issues/712

The `numberToHexString` method incorrectly only put a 0 for numbers below 10 instead of 16, outputting `A` instead of `0A` for 11, for example. I've added a test at the only place that uses `toHexString` that fails on the current master branch. I'm not sure if it's in the right place, please let me know if I should move it/create seperate test cases.

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1621,6 +1621,16 @@ describe "Config", ->
           expect(color.toHexString()).toBe '#ff0000'
           expect(color.toRGBAString()).toBe 'rgba(255, 0, 0, 1)'
 
+          color.red = 11
+          color.green = 11
+          color.blue = 124
+          color.alpha = 1
+          atom.config.set('foo.bar.aColor', color)
+
+          color = atom.config.get('foo.bar.aColor')
+          expect(color.toHexString()).toBe '#0b0b7c'
+          expect(color.toRGBAString()).toBe 'rgba(11, 11, 124, 1)'
+
         it 'coerces various types to a color object', ->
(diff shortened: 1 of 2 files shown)


lee-dohm:
Awesome, thanks!

PR #11101, opened 2016-03-08
Emit status changes when anything changes

joshaber:
Fixes https://github.com/atom/status-bar/issues/128.

This mirrors the behavior of `GitRepository`.


joshaber:
(When this lands I'll cherry-pick it into beta.)

PR #11111, opened 2016-03-09
Default the options parameter to an empty object

mehcode:


as-cii:
Hey @mehcode, thanks for contributing!

Could you explain why you are supplying a default value for `options`? Initially I did the same in #11060, but then realized that there was only one caller and that passing `options` explicitly forced callers to pass the `isUnloading` parameter.

What's the scenario you have in mind for this?

mehcode:
@as-cii A couple reasons.

1. Mainly petty as it broke my [project-plus](https://github.com/mehcode/atom-project-plus) package but that's my fault for targeting master and using non-stable APIs. However it was nice to get an indication that "hey, something changed" and I adjusted it to pass the options object.

2. Convention. Every other `.serialize` call takes no parameters. This means that we can no longer iterate through serializable objects (not that I do) and wantonly serialize so it breaks intuitive expectations (aka one would have to dive into the code / read the docs to understand).

as-cii:
> However it was nice to get an indication that "hey, something changed" and I adjusted it to pass the options object.

Yeah, that was the original intent of making the options parameter mandatory, especially since it is a private API.

> Convention. Every other .serialize call takes no parameters. This means that we can no longer iterate through serializable objects (not that I do) and wantonly serialize so it breaks intuitive expectations (aka one would have to dive into the code / read the docs to understand).

My long term plan was to pass the `isUnloading` parameter to packages as well, so that they can decide whether to handle the unloading scenario in a different way. Similarly, they could be able to iterate through the serializable objects and pass those options, if they want to. 

I believe that the way serialization is handled for `Project`, `Workspace`, etc., is part of the private API and shouldn't be accessed by packages, but at the same time I think people might be using it and ideally I'd like this change to be 100% backwards compatible.

So, I am going to :ship: this and add a test to make sure we don't make the mistake of not passing the `isUnloading` parameter internally. Thanks, @mehcode!

mehcode:
> I believe that the way serialization is handled for Project, Workspace, etc., is part of the private API and shouldn't be accessed by packages, but at the same time I think people might be using it and ideally I'd like this change to be 100% backwards compatible.

Off-topic.. but is there any chance of this becoming a public API? Or perhaps would it be an acceptable improvement to make some methods around it (to facilitate project switching) like something for "getting all projects" and something for "switching the editor window to a different project"?

as-cii:
> Off-topic.. but is there any chance of this becoming a public API? Or perhaps would it be an acceptable improvement to make some methods around it (to facilitate project switching) like something for "getting all projects" and something for "switching the editor window to a different project"?

I think right now we expose a bunch of methods in `Project` (e.g. `::getPaths`, `::getBuffers`) that allow package authors to replicate the `serialize` behavior in a custom way. I don't think this scenario would be so general purpose to deserve a new API for that, but definitely feel free to come up with a proposal in a new issue and mention me for further discussion. Some things you might mention are how you plan to use that API, how it would look like and what new and enticing scenarios that would enable. :memo: 

Thanks! :sparkles:

PR #11115, opened 2016-03-09
Update nodegit

joshaber:
Fixes a crash when creating a diff. See https://github.com/nodegit/nodegit/pull/945.


joshaber:
Looks like it hasn't been published on npm yet.

PR #11122, opened 2016-03-10
Compute a state key even if there are no project folders

### LINKED ISSUE #10474: Unsaved buffers aren't restored if there is no project folder added to the window
According to this [release notes for 1.4.0beta](https://github.com/atom/atom/releases/tag/v1.4.0-beta0) and this [issue report](https://github.com/atom/atom/issues/942) Atom is supposed to automatically save any unsaved/untitled buffer upon closing.

Steps to reproduce:

1. Open Atom.
2. Open a new tab
3. write something without saving it
4. close atom (click X or file->exit)
5. reopen atom
6. the unsaved new untitled tab is missing. It should be there with all the content that you wrote.

This is important feature for me, please fix it. Thanks.

mehcode:
This allows for state to be saved without a project folder added.

Would fix #10474 

---

I can add a spec for this if this something we want.

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -880,7 +880,8 @@ class AtomEnvironment extends Model
       sha1 = crypto.createHash('sha1').update(paths.slice().sort().join("\n")).digest('hex')
       "editor-#{sha1}"
     else
-      null
+      # Represents an empty editor (no project folders)
+      "editor-<empty>"
 
   getConfigDirPath: ->
     @configDirPath ?= process.env.ATOM_HOME


src/atom-environment.coffee
```diff
@@ -880,7 +880,8 @@ class AtomEnvironment extends Model
       sha1 = crypto.createHash('sha1').update(paths.slice().sort().join("\n")).digest('hex')
       "editor-#{sha1}"
     else
-      null
+      # Represents an empty editor (no project folders)
+      "editor-<empty>"
```
mehcode: Completely blanking on a good name here

alexandernst:
How will this behave with multiple instances without a project folder? Will it restore all instances or will it restore just one instance with all the tabs from all the instances?

50Wliu:
@alexandernst Wouldn't having separate instances share the same serialization state?  So if you edit some files without a project, then open a new Atom window, shouldn't those same files show up?

(Also, this should adhere to the openEmptyEditorOnStartup option)

alexandernst:
@50Wliu  No, not really. At least here, opening multiple instances opens completely separate windows that do not share any of the open files (neither existing nor unsaved buffers).

leobasilio:
I might be wrong, but isn't this return value used as the file name? < and > are not valid file name characters. Why don't you just return the sha for an empty string by removing the "if paths?.length > 0"?

mehcode:
@leobasilio States got moved to IndexedDB. They are no longer in `~/.atom/storage`. I wanted the state name to be something that clearly indicated empty.. but your way would work nice too.

rreinhardt:
This is great! Looking forward to this getting rolled in but I will share that it does not quite work as expected right now if multiple windows are opened. 

e.g.

Start Atom
Add text in Untitled (1) in Window (1)
Create New Tab Untitled (2) in Window (1) & add text
Create New Window & add text in Untitled (1) in Window (2)...

If you bail on Atom, anything in Window (2) is lost. Atom will restore two Windows but both will contain tabs from the first Window (1) in the above scenario. Continuing to edit in Window (2) will not get acknowledged in subsequent exits/restores.

Additional weird behavior is if you create a new window, after serialization, the new document in the new window will contain the text from the previous window.

All in all, this is a huge improvement to behavior in v1.6 — currently there is not even 'save' dialogs before exiting/dataloss.  

As long as there is one 'no project' window open, things should work as you'd think.

mehcode:
> [...] As long as there is one 'no project' window open, things should work as you'd think.

Hmm.. I'm not sure how to solve what you're describing. This probably requires some more thought rather than a quick patch, as the solution to what you describe sounds involved. Perhaps each non-project window has some sort of unique identifier and we remember these identifiers.

PR #11139, opened 2016-03-12
Add Selection.onDidChangeRange event object to emitter

### LINKED ISSUE #11133: Selection.onDidChangeRange callbacks do not get an event passed
According to [the docs](https://atom.io/docs/api/v1.5.4/Selection#instance-onDidChangeRange), a callback should get an `event` as argument. However, [Atom doesn't do this](https://github.com/atom/atom/blob/v1.5.4/src/selection.coffee#L784).

lee-dohm:
Fixes #11133


smashwilson:
:sparkles: _N-n-n-n-necromerge_ :sparkles:

PR #11153, opened 2016-03-13
Fix node env

lee-dohm:
https://github.com/atom/atom/pull/11054 wasn't preserving the `NODE_PATH` environment variable when the `process.env` was changed. This caused at least one problem: fuzzy-finder could no longer index files. This fix preserves all `NODE_` environment variables from the original environment, no matter what other changes are made.

/cc @joefitzgerald @nathansobo

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -4,7 +4,6 @@ path = require 'path'
 
 _ = require 'underscore-plus'
 {deprecate} = require 'grim'
-environmentHelpers = require('./environment-helpers')
 {CompositeDisposable, Disposable, Emitter} = require 'event-kit'
 fs = require 'fs-plus'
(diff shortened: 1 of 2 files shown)


lee-dohm:
Since `NODE_PATH` and `NODE_ENV` are explicitly set here:

https://github.com/atom/atom/blob/6b38049b8d10deb9abf66fbdb19d2280e04075f2/src/initialize-application-window.coffee#L12-L15

I figured this was prudent.

joefitzgerald:
You should just move the normalize call into initialize-application-window, instead of having the replaceEnv function. Put it before the NODE_ vars are set but after env is set.

nathansobo:
Seems like a refinement of #11054, so I'm :cool: with it.

Issue #10623, opened 2016-01-29
"Open Empty Editor On Startup" option not working.


ghost:
Option "Open Empty Editor On Startup" is not working.
image: http://i.stack.imgur.com/zJrRp.png

It doesn't matter if i tick or untick it, it always opens my last opened project and opened files.

Doesn't work on windows (win10) and linux (Mint 17 Cinnamon).
Tried to disable all plugins. Still having this issue.

If i can give You any information, log file, etc, let me know.

lee-dohm:
Related to: https://github.com/atom/atom/issues/10517

brentshermana:
Confirmed on Ubuntu 14.04.3 with Atom 1.6.0-dev-f914d18

To be clear, is it just the tabs which should not be brought up, or that and all files in the project tree on the left hand side?

ExClouds:
It always opens both:
* project folder on the left side
* previously opened files (doesn't matter if the files were project or manually opened files) (even atom settings pages, like init, keymaps, settings, etc)

brentshermana:
I understand that, but what is it *meant* to do?

ExClouds:
Well, the checkbox say's **Open Empty Editor On Startup**, so, when it's ticked, i expect the editor to open blank.. No open tabs, no open projects..

demhydraz:
Can confirm on Arch Linux with 1.6.0-beta3.

minsis:
Confirmed on Ubuntu Mint 15.10 and Windows 7 Ent.

zzeroo:
Also here on debian sid with 1.6.0-beta3

demhydraz:
Can still confirm on beta6.

sp82:
Same on Windows7

Kingdaro:
And on Windows 10. Sorry to add more confirmations than solutions, but at least we're a _little_ closer to knowing that this is totally an issue on every platform!

demhydraz:
Still happening on beta7, same platform.

delliott2016:
The option causes a different problem on Mac OS X (v10.11.3) using Atom 1.8.0-dev-3959259. When the box is checked, the workspace opens up empty with a single untitled tab. When the box is unchecked, the workspace opens up completely empty.

leobasilio:
[...]

PR #11161, opened 2016-03-14
:bug: Make "Open Empty Editor on Start" config effective

leobasilio:
As reported on issue #10623, this config isn't effective or at least doesn't work as it sounds it should. This is due to the state persistence feature, which is unaware of this setting, restoring the editor to its last state on every startup. I agree with @ExClouds (the guy who opened the issue), when he says

> i expect the editor to open blank.. No open tabs, no open projects..

My strategy was to prevent state saving/loading when the setting is on. This is my first contribution to atom and I'm not that familiar with the code. I did what I thought would be the least intrusive changes.


leobasilio:
Giving up on this PR in favor of #11324.

PR #11166, opened 2016-03-14
Note where GitRepositoryAsync deviates from its synchronous predecessor

joshaber:
Documents changes wrt https://github.com/atom/git-diff/issues/93 and https://github.com/atom/git-diff/issues/94.

### DIFF
--- a/src/git-repository-async.js
+++ b/src/git-repository-async.js
@@ -442,9 +442,6 @@ export default class GitRepositoryAsync {
   // Returns a {Promise} which resolves to a {Boolean} that's true if the `path`
   // is ignored.
   isPathIgnored (_path) {
-    // NB: We're matching the behavior of `GitRepository` here.
-    if (!_path) return Promise.resolve(false)
-
     return this.getRepo()
       .then(repo => {
         const relativePath = this.relativize(_path, repo.workdir())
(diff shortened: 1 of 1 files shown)


joshaber:
@atom/core I'd like some Opinions® on this. Should we maintain API compatibility, or require users to change?

joshaber:
To expound a bit more, users would be opting into using `GitRepositoryAsync` over `GitRepository` to begin with. But should we maintain the same API contract, or are we ok with behaving differently in the edge cases?

iolsen:
Because it's opt-in, this seems like a good time to make these kinds of changes: minor API contract breakage netting better behavior. I'm :+1:

lee-dohm:
I'd feel better about the API breakage if it was discoverable without running headlong into it, for example a note in the API documentation saying something like:

```
`GitRepository` and `GitRepositoryAsync` behave differently in some cases, see the [list of known differences][link-tbd] for more information.
```

thedaniel:
I don't think there's any kind of contract we're breaking. As I recall we started with the GitRepository API as a familiar starting point / make sure we covered all the required functionality, but I expected it to diverge more than it has already, tbh. In my opinion as long as we can support all the actions the core packages need, the API of GitRepositoryAsync is fine to diverge widely from the sync class, if necessary.

Being similar is nice, but we won't get an opportunity to redesign this API for a while.

thedaniel:
> GitRepository and GitRepositoryAsync behave differently in some cases, see the [list of known differences][link-tbd] for more information.

They behave differently in ~~some~~ basically all cases because one returns promises instead of values ;) (Yes, I know the promises usually resolve to things that match the sync api :trollface:) but I'm a lot less worried about this because no one's using it in the wild - it's not even on stable yet!

joshaber:
:+1: Ended up just documenting the differences. Thanks all!

nathansobo:
Would like to add my agreement that since we're adding API surface area, we should change what we need to.

PR #11184, opened 2016-03-17
cli: Open stdin as a file when present

### LINKED ISSUE #1729: Piping input into Atom doesn't work
support/4aa96dda9fd511e3961e21f79eba7c86

> It'd be nice if piping input into atom were possible.
> `git diff | atom`
> Right now this seems to open a second instance of Atom that immediately closes. The data goes nowhere.

This probably belongs in the feature-request arena, but also seems like something to keep track of here. Personally, I think this would be great to have. :doughnut: 

Other reports:
- support/e9fab3a2a94a11e39ef7d68ca2b55878
- support/ddc261f2a5df11e39bf33f35dc393db2

mkhl:
The [atom cli discards its input from stdin](https://discuss.atom.io/t/it-would-be-nice-to-be-able-to-pipe-standard-out-to-atom-and-have-it-open-in-an-unsaved-document/4781) when other editors would open a new document containing that text.

This change writes input from stdin to a new temporary file and inserts that file as the first argument to the atom executable.
It only does that when stdin is not at tty, i.e. running `git diff | atom` creates and opens a temporary file but running `atom some/file` does not.

Closes https://github.com/atom/atom/issues/1729
## Caveats
- [ ] Errors from `mktemp` are not handled.
- [ ] The call to `mktemp` appears before ensuring that `TMPDIR` is set on line 105.
- [ ] `mktemp` is called without arguments, resulting in a filename like `tmp.XXXXXXXX`.  
  A filename prefix like `atom-stdin` would be nicer but the mktemp commands on [Linux](http://www.unix.com/man-page/linux/1/mktemp/) and [OS X](http://www.unix.com/man-page/osx/1/mktemp/) are sufficiently different to make this more involved. (A call like `mktemp -t atom-stdin` would work on OS X but wouldn't include a random suffix on Linux, if I understand the man page correctly.)

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -52,6 +52,12 @@ if [ $EXPECT_OUTPUT ]; then
   export ELECTRON_ENABLE_LOGGING=1
 fi
 
+if [ ! -t 0 ]; then
+  TEMPFILE="$(mktemp)"
+  cat > "$TEMPFILE"
+  set "$TEMPFILE" "$@"
+fi
+
 if [ $OS == 'Mac' ]; then
   if [ -n "$BETA_VERSION" ]; then
     ATOM_APP_NAME="Atom Beta.app"


damieng:
Hi there, what platforms have you tried this on?

mkhl:
I'm using the script on OS X.

I also tried this proof of concept script on GNU/Linux (a Fedora 23 VM inside a Docker container):

```bash
#!/bin/bash

if [[ ! -t 0 ]]; then
  TEMPFILE="$(mktemp)"
  cat > "$TEMPFILE"
  set "$TEMPFILE" "$@"
fi

echo "$@"
```

damieng:
I'll test this on Windows.

50Wliu:
I'll test on Ubuntu 16.04 later today.

atom.sh
```diff
@@ -52,6 +52,12 @@ if [ $EXPECT_OUTPUT ]; then
   export ELECTRON_ENABLE_LOGGING=1
 fi
 
+if [ ! -t 0 ]; then
+  TEMPFILE="$(mktemp)"
+  cat > "$TEMPFILE"
+  set "$TEMPFILE" "$@"
```
damieng: Isn't this going to wipe out the other command arguments?

damieng:
Doesn't work on Windows as the magic is in the Bash file. We'd have to do the same thing in the command/batch file there. Let me check out what' involved.

atom.sh
mkhl: No, the `"$@"` preserves them.

50Wliu:
Tested on Ubuntu and everything seems to work fine (other than the tmp folder opening as a project root, but that's a different issue).

Tried:
`git log | atom`
`git log | atom --foreground`
`ps aux | grep "atom" | atom` (yes I know there's `pgrep`, just wanted to make sure multiple pipes worked)
`ps aux | grep "atom" | atom package.json` (this opens up both files but focuses `package.json`)
`ps aux | grep "atom" | atom --foreground --safe --dev package.json`

KayvanMazaheri (approved):

Blink18260000:
@mkhl @KayvanMazaheri Can this get pulled in sometime relatively soon? It seems like this PR has been ready since October, and I would appreciate having this functionality without having to set up my own hacky bash macro.

damieng:
The hold-up is the fact it doesn't work on Windows. We don't want to introduce functionality like this that only works on a subset of our supported platforms.

drosenstark:
For those of you who want to add this functionality **now** on OS X (and don't want to wait for the PR to be merged):

Edit your `/usr/local/bin/atom` (in Atom, for instance)

Below this code:

```
if [ $EXPECT_OUTPUT ]; then
  export ELECTRON_ENABLE_LOGGING=1
fi
```

insert this code

```
if [ ! -t 0 ]; then
   TEMPFILE="$(mktemp)"
   cat > "$TEMPFILE"
   set "$TEMPFILE" "$@"
fi
```

Save the file and now piping should work.

drosenstark:
The code has changed recently and now this hack no longer works. Any ideas?

tithonium:
@drosenstark: I stuck it after
```
if [ $REDIRECT_STDERR ]; then
  exec 2> /dev/null
fi
```
and it worked for me.

jguffey:
Is this feature dead?

drosenstark:
The hack above is no longer working, I think. Anyway, I've started using backticks like this:

```
atom `find . | grep -i newrelic_post`
```

good enough.

sadick254:
Hello @mkhl, I know this PR has been open for a very long time now. Apologies we didn't get to it on time. Having this functionality in Atom would be really great. However, looking at the implementation, this would only work on macOS and Linux systems leaving out Windows. A better approach would be to handle this at the entry points of atom i.e `start.js` or `main.js`. That would ensure we get a working implementation that would work for all systems.
I am going to close this for now. I am happy to reopen when the implementation is updated.

Thank you for your contribution @mkhl

PR #11187, opened 2016-03-17
[WIP] Ensure cursor is visible after showing panels

### LINKED ISSUE #8068: Shrink or scroll editor when panels at bottom open up
If this has been raised before or if this is the wrong repo, please correct me. I couldn't find a previous issue for it. I've raised this issue at the atom linter repository originally, see https://github.com/atom-community/linter/issues/758, but we felt this was a more widespread issue and not the responsibility of the linter package.

Considering that [linter might get integrated into the atom core](https://github.com/atom/atom/issues/7353), it's probably a good idea to handle this issue in atom core anyway.

The problem is with panels that open up at the bottom of the Atom window (e.g. find & replace, linter) while you are working on code which is at the bottom of a file. These panels overlap your code when they open, hiding the code from view and forcing you to scroll up.

Type something erroneous at the bottom:
ss

Panel shows up and overlaps your code:
ss

The problem is particularly troublesome with automatically opened windows, like linter. The find & replace panel never bothered me like that, because it's triggered manually with a keyboard command, so you're not focussed on typing code anyway. But with the linter panel, you are, and it's very disturbing to have the panel suddenly show up and block your view of your code.

A proposed work around has been to enable the setting "Scroll Past End", but while this alleviates the issue a little bit, it doesn't solve it and it can still happen when you forgot to scroll further up.

A better solution would probably be for the panels to not overlap the editor window, but rather push it up causing it to shrink. I realize this might cause the opposite problem though, e.g. if you're typing at the top of the file, then that code would be pushed out of view.

Any idea's?

joshaber:
Fixes #8068.

So this is an interesting one.

We want to ensure the cursor is visible after we show a panel. The tricky bit is that we have to do this _after_ the text editor’s been laid out to include the new panel.

@atom/core I’d love to get some thoughts on how we can accomplish this in a way that doesn't make me sad inside.


nathansobo:
@joshaber I have an idea. What if we detected the fact that we weren't visible when the autoscroll request occurs and stored it away, then autoscrolled when we became visible. Would that make you less sad?

joshaber:
@nathansobo Good call :metal:

joshaber:
Closing this for now. Testing it has proven to be more complicated than I want right now.

PR #11188, opened 2016-03-17
Bump status-bar

joshaber:


joshaber:
(The original bump was reverted because it broke some specs.)

PR #11201, opened 2016-03-18
Remove an unnecessary call to then().

bolinfest:
I see that this file has a little use of async/await and many uses of `then()`.
Things would be much less verbose and much more linear if async/await were used throughout.
Would Atom be open to a PR that changes this?

For bonus points, we could also replace `'use babel'` with `/* @flow */` at the top of the file :)

### DIFF
--- a/src/git-repository-async.js
+++ b/src/git-repository-async.js
@@ -158,8 +158,7 @@ export default class GitRepositoryAsync {
 
     if (!this.projectAtRoot) {
       this.projectAtRoot = this.getRepo()
-        .then(repo => this.project.relativize(repo.workdir()))
-        .then(relativePath => relativePath === '')
+        .then(repo => this.project.relativize(repo.workdir()) === '')
     }
 
     return this.projectAtRoot


joshaber:
>Things would be much less verbose and much more linear if async/await were used throughout.
Would Atom be open to a PR that changes this?

We've had some concerns about the performance of the code Babel translates async/await into, so I'm a bit hesitant to translate it fully until that improves, or we have a better handle on how it performs in practice.

>For bonus points, we could also replace 'use babel' with /* @flow */ at the top of the file :)

Maybe some day :sweat_smile:

joshaber:
Thanks @bolinfest!

bolinfest:
@joshaber Thanks for taking the PR. Do you have any data to back up your performance concerns around async/await? I believe you would end up creating fewer closures in many cases.

joshaber:
>Do you have any data to back up your performance concerns around async/await?

No, it was just something we talked about in chat. I'd love to be shown wrong.

PR #11216, opened 2016-03-20
:racehorse: wrap loop of config.setDefaults() in transact()

ypresto:
When config.setDefaults() is called "did-change" event emitted per one config key.
This hits performance when package has large set of default configs (like [atom-beautify](https://github.com/Glavin001/atom-beautify/issues/679).

Without transact: 

With transact:

Thanks!


maxbrunsfeld:
Thanks for noticing this @ypresto!

ypresto:
I noticed this overlaps with #12226 because setDefaults is only called by setSchema. Anyway thanks..!

PR #11240, opened 2016-03-22
More accurate block decoration measurements

### LINKED ISSUE #11228: Block decorations not measured properly in all cases
Running the following snippet in the console (the example from the block decorations blog post, with a child added that has a margin-bottom) causes this unfortunate rendering problem:

screen shot 2016-03-21 at 4 31 28 pm

```js
var element = document.createElement('div')
element.textContent = 'Block decorations! 🐲'
var child = document.createElement('div')
child.setAttribute('style', 'margin-bottom: 50px;')
element.appendChild(child)
var editor = atom.workspace.getActiveTextEditor()
var marker = editor.markScreenPosition([0, 0])
editor.decorateMarker(marker, {type: 'block', position: 'before', item: element})
```

`offsetHeight` on the containing element is returning the wrong value in this case. It seems we'll need to find a more accurate way to measure block decorations, or (much less ideal) document what methods to set an item's size are allowed to be used in block decorations.

cc @as-cii @nathansobo 

(atom 1.7.0-dev-f48069c)

as-cii:
Fixes #11228.

Using `offsetHeight + marginTop + marginBottom` for any given block decoration item isn't appropriate, because it doesn't take into account the decoration's children and how they affect the surrounding content spatially.

With this PR, we are adding an element before and one after every block decoration, and use them as rulers that help us computing the appropriate height. Although it introduces two extra elements for each decoration, it shouldn't be a huge deal since those elements are supposedly pretty straightforward to lay out for Chrome.

/cc: @thedaniel @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1840,17 +1840,22 @@ describe('TextEditorComponent', function () {
       expect(component.lineNodeForScreenRow(2).dataset.screenRow).toBe("2")
     })
 
-    it('measures block decorations taking into account both top and bottom margins', async function () {
+    it('measures block decorations taking into account both top and bottom margins of the element and its children', async function () {
       let [item, blockDecoration] = createBlockDecorationBeforeScreenRow(0, {className: "decoration-1"})
+      let child = document.createElement("div")
+      child.style.height = "7px"
+      child.style.width = "30px"
+      child.style.marginBottom = "20px"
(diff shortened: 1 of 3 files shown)


nathansobo:
Seems like a good approach.

as-cii:
I restarted the build a bunch of times since I have opened this, but there seems to be a flaky spec that's unrelated to the change.

:ship:ping this. :rocket:

PR #11246, opened 2016-03-23
Improve the Windows build process

damieng:
Gradually improving the Windows build experience, so far:
- **clean** deletes all specified folders (doesn't stop on first missing one)
- **build** removes any path entry containing a msbuild.exe (avoids node-gyp errors)

### DIFF
--- a/script/build
+++ b/script/build
@@ -2,8 +2,23 @@
 var cp = require('./utils/child-process-wrapper.js');
 var runGrunt = require('./utils/run-grunt.js');
 var path = require('path');
+var fs = require('fs');
 
 process.chdir(path.dirname(__dirname));
+
+if (process.platform === 'win32') {
+  process.env['PATH'] = process.env['PATH']
+    .split(';')
+    .filter(function(p) {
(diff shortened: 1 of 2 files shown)


SaadLaggoune:
yeah, you are right, but windows version have a lot things wrong, sometimes i can't write '}' it's make me made, Atom Editor is good but needs to improve Windows OS experience .

damieng:
Hi @SaadLaggoune this PR is specifically about improving the build process itself. There are various issues and improvements for Windows happening right now - keep an eye on the Windows tags on PR and Issues for more updates.

ZombieHippie:
@damieng are there things we can work on to expedite run spec test suites on Windows? Thanks for putting some focus in this direction.

damieng:
Feel free to investigate any specs failures on Windows and set-up PR's for them.  Right now a couple of them actually crash atom so I'd recommend using the prefix-with-f trick to run individual specs - e.g. 'fdescribe' instead of describe to just run that spec suite. Once fixed remove the f.

I'm trying to nail down the path issues right now of which there are many as various url/libs try and treat windows paths as linux paths and invert the slashes.

PR #11251, opened 2016-03-23
Fallback to ~/.atom/storage when no state is found in IndexedDb

as-cii:
This PR addresses the concerns raised by @jeancroy  in https://github.com/atom/atom/pull/10605#issuecomment-199603786 by falling back to reading state from the storage folder when we can't find one under IndexedDb.

Moreover, this introduces a `clear()` method on `StorageFolder` that deletes the storage folder whenever the user supplies the `--clear-window-state` argument (so that no state can be loaded in both IndexedDb and `~/.atom/storage`).

/cc: @nathansobo @kuychaco @iolsen

### DIFF
--- a/src/storage-folder.coffee
+++ b/src/storage-folder.coffee
@@ -12,7 +12,7 @@ class StorageFolder
     try
       fs.removeSync(@path)
     catch error
-      console.warn "Error deleting #{statePath}", error.stack, error
+      console.warn "Error deleting #{@path}", error.stack, error
 
   storeSync: (name, object) ->
     return unless @path?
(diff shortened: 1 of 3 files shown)


src/storage-folder.coffee
```diff
@@ -6,6 +6,14 @@ class StorageFolder
   constructor: (containingPath) ->
     @path = path.join(containingPath, "storage") if containingPath?
 
+  clear: ->
+    return unless @path?
+
+    try
+      fs.removeSync(@path)
+    catch error
+      console.warn "Error deleting #{statePath}", error.stack, error
```
t9md: statePath is out of scope?

thomasjo: Yup! `statePath` should be replaced by `@path`.

as-cii: Oops, good call @t9md. Thanks!

Issue #8648, opened 2015-09-03
Linux middle mouse paste inserts text twice


cloos:
- mark somewhere text with the mouse
- File -> New File
- press middle mouse button to insert text

The text is inserted twice.

This happens since 1.0.8 and even in the latest 1.0.10.

atom-amd64.deb
Ubuntu 15.04

peduxe:
[Here is a solution.](https://atom.io/packages/sublime-style-column-selection)

cloos:
If I install 1.0.7, the text is inserted only once, so everything is fine.
The issue appeared since 1.0.8.

As this seams to be a core issue, I don't think installing a community package is a right solution.

Also I think that this wouldn't solve the issue, as by "mark somewhere text with the mouse" I mean mark text for example inside the browser, and not text inside the atom editor.

peduxe:
I couldn't reproduce this. It just pasted once.

nicksanders:
I'm experiencing this too after upgrading from 1.0.7 to 1.0.11

dpanter:
Could not reproduce this issue with 1.0.11, fresh install.
Following OP instructions, it only pasted once.
Linux Mint Debian Edition, KDE4.

nicksanders:
This is still happening with Atom 1.0.13

To reproduce
mv ~/.atom ~/.atom-saved
atom --safe

In untitled tab 
type "test" then press enter twice
double click left mouse button on test to highlight
move mouse to line 3 and press middle mouse button

![atom1](https://cloud.githubusercontent.com/assets/809448/9961499/c512ff7a-5e18-11e5-9227-d6db3513476d.png)

peduxe:
@nicksanders I can confirm this happens to me as well.

i3wm, Linux 15.04 64-bits Atom 1.0.14

dpanter:
@nicksanders I'll revise my earlier statement and say I can in fact reproduce this exact behavior as well, a left click before pasting with a middle click changes the result.

However... as I tried middle clicking more than once, I realized each subsequent middle click makes my 1.0.11 doubly paste what it just pasted. Preceding left click or not makes no difference.
I get "test" -> "testtest" -> "testtesttesttest" -> "testtesttesttesttesttesttesttest" -> "testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttest" and so on.

Probably not the intended behavior.

hainesr:
This is happening in version 1.0.19 (from the PPA) on Ubuntu 14.04 too.

[...]

PR #11262, opened 2016-03-24
workaround: do not paste if stopped dragging

iivvoo:
I'm not sure if this is the correct solution and if it has any side effects. I suspect that the mouse button 2 action on linux can be detected as a mini-drag, causing onMouseUp to be invoked twice. 

Issue: https://github.com/atom/atom/issues/8648#issuecomment-200905442

### DIFF
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -650,7 +650,8 @@ class TextEditorComponent
         stopDragging()
         @editor.finalizeSelections()
         @editor.mergeIntersectingSelections()
-      pasteSelectionClipboard(event)
+      else
+        pasteSelectionClipboard(event)
 
     stopDragging = ->
       dragging = false


iivvoo:
as far as I can tell, tests are failing because of CI-brokeness. Restarting might fix it. But please correct me if I'm wrong :)

50Wliu:
Specs are flaky, so as long as you aren't seeing any failures that may be related to your fix, you should be fine.

Siilwyn:
Looking good, can confirm this issue is there on Linux (Ubuntu 14.04) even in safe mode. And the change made by @iivvoo fixes it.

Siilwyn:
Ping @50Wliu, this should be clear to merge right?

g3kk0:
Is anyone looking at this? It's been an on going problem since Sept 2015. I've just installed the latest stable release (1.11.1) and the middle mouse button still pastes multiple times.

iivvoo:
Installing vscode fixed it for me ;)

Siilwyn:
It's sad to see such a simple fix as this be open for so long. :/

iivvoo:
If it were an OSX issue it would have been fixed yesterday...

jlsalvador:
I checked this patch and sometimes Atom still pastes twice and three times. Has someone validated it?
I proposed this some time ago: https://github.com/atom/atom/issues/8648#issuecomment-172554041

Siilwyn:
@jlsalvador yes I validated it, see my reply on May 12.

50Wliu:
Superseded by #14063.

Issue #9045, opened 2015-10-05
 Pasting with a newline in front adds last lines indentation to pasted one


hitodama:
Example ("-" are indentations, "\n" is a leading newline from the previous line):

```
\n
---copy me
```

```
---indented line <-- copied here, to an indented line
------copy me <-- will end up indented as such
---------copy me <-- when done again on above line
```

I expect:

```
\n
---copy me
```

```
---indented line <-- copied here, to an indented line
---copy me <-- should end up as such, not further indented
---copy me <-- when done again on above line
```

To reproduce, open a new (language independent) file and copy-paste as shown. I am not sure this works in all other languages (it does in C), because I don't know the extend of their potential interference. (But I guess so basic rule (regex?) is broken.)

I typically turn of all indentation management but in this case **it seems to happen regardless of the options** (Auto Indent, Indent on Paste, Normalize Indent on Paste) set. In my opinion, **it should (almost) never happen**, because we don't want to indent most expressions in programming, unless they are scoped. Likewise, for texts it makes no sense (to me).

So:
1) For auto indentation this seems wrong, the indentation should not increase.
2) For "Indent On Paste" (or any other indent options) set to false, it is a **bug**, because the indentation should never be touched at all.

Infos: Linux, atom 1.0.0, also tested with git repo version 1.2.0-dev (commit 4b507be99769a38e0b8e9b26a950fc408951f986)

hitodama:
I may have spoken to soon, the issue still exist in C as well. Just try pasting around with newlines in front and different (lower levels of) indentations as seen above. The indentation will change in an unwanted manner, regardless of options (but potentially differently).

oseiskar:
+1 for "bug" because this happens when all auto-indenting options are turned off.
Tested on Linux (Debian) with v. 1.3.1.

brentshermana:
hitodama, I've solved this issue [here](https://github.com/atom/atom/pull/11267). Please help me test my code on Windows so we can get this pull request accepted!

pauloschilling:
Same problem here, and that's really annoying. =/

Kzone272:
How has something as simple as copying and pasting code in an editor as popular as Atom been broken for over a year?

I'm starting to regret the switch from Sublime.

patstew:
I've had another go at fixing this - see this package: https://atom.io/packages/fix-indent-on-paste
It would be useful to get feedback on whether that package works as people expect - if so I'll update PR #12793 with the same code

Issue #5473, opened 2015-02-10
Auto indent on paste behaves oddly with already indented lines.


NatalieWolfe:
When copying a segment of code that starts with a new-line followed by tabbed-in code while "Auto Indent On Paste" is enabled, consecutive pastes will each be an additional tab deeper. I tested this with C++, Java, and JavaScript which makes me believe it is not a problem with the language specs.

Here is an example in the form of an animated gif:
![autoindent-2](https://cloud.githubusercontent.com/assets/466761/6118758/d4ce2e04-b075-11e4-864b-46726fcd809e.gif)

It isn't just comments that this happens with, but it does require the previous line to be selected. Specifically, this happens when the copy-buffer matches a regex like `/^\n\s+[^\s]/`. That is, it starts with a newline, followed by one or more whitespace characters and then a non-whitespace character. And the previous line must be a blank line.

50Wliu:
What version of Atom is this on?

/cc @maxbrunsfeld

NatalieWolfe:
Oops, should have included that. I'm using Mac v0.177.0, but I've been experiencing this bug for a few versions now (not entirely sure when it first presented).

Interesting, just tried to reproduce this on my Windows build box too. It has v0.175.0 installed. On that machine the behavior is reversed. When Auto Indent On Paste is enabled then pasting behaves as expected (each consecutive line is pasted at the same depth). However when that option is disabled then the first paste behaves as expected, but the 2nd through nth pastes are all one level deeper than the last. Not sure how to update the Windows version to see if it persists in 0.177.0.

izuzak:
@NatalieWolfe I'm having trouble reproducing this in the latest version of Atom:

```
$ atom -v
0.180.0-2dac7fd
```

Here's what I'm doing:

![paste](https://cloud.githubusercontent.com/assets/38924/6190840/29790b7c-b3a5-11e4-90b2-9a12c08adc91.gif)

Is there something I'm doing differently than you are?

Can you try reproducing this again and record a gif in which the invisible characters are shown as well as the keybinding resolver (`cmd+.`)? Also, can you record the gif so that includes the point at which you select and copy text into the clipboard for the first time (it's not shown in the gif above)?

Atom should update to the latest version on it's own, but you can download the latest version here as well:

https://github.com/atom/atom/releases/latest

NatalieWolfe:
@izuzak: The link you provided goes to 0.179.0, which is what I have installed on my Mac at the moment. 

[...]

Issue #4461, opened 2014-12-09
Auto indent still misbehaving


ghost:
Atom 0.154 on Mac OS.

![auto-ind](https://cloud.githubusercontent.com/assets/440132/5354669/deb66bfa-7f3f-11e4-8fd8-9d62e7b221d6.gif)

Code below (make sure your syntax mode is CoffeeScript). Follow the instructions therein.

```
$ ->
  $(window).load ->

# Copy the below and paste it on the line below $(window).load ->
# Make sure the cursor is indented one level in from $(window).load (col = 5)
    if 2 is 2
      celebrate()
```

brentshermana:
Please help me test my pull request on Windows :)

PR #11267, opened 2016-03-24
Fixed selection.copy, which caused bugs in selection.insertText when pasting

brentshermana:
Referenced Issues: [1](https://github.com/atom/atom/issues/9045) [2](https://github.com/atom/atom/issues/5473) [3](https://github.com/atom/atom/issues/4461) 

In the specific case where you copy text beginning with a newline, the metadata field 'indentBasis' can be improperly stored, which doesn't create any issues until you try to paste somewhere else and the 'autoIndentOnPaste'  setting is on. It uses that incorrect indentBasis which results in incorrect indentation.
### evidence for the bug

1) Take a build of atom besides this one, and make a new file. add the following text:

```
one
  two
```

2) Add `\n\ttwo` to your clipboard by clicking and dragging from the right of 'two' until the right on 'one'. It's important that you capture that newline character.

3) With your cursor to the right of 'two', press cmd+v or ctrl+v depending on your OS. You should have had  the twos at the same indentation level, but instead the second two will be indented twice.

4) If you're not convinced yet, you can try this again with the first 'two' indented twice, and you'll see the second one indented four times.
### What Happens

copy() takes the indent level of the first selected line and stores that as the indentBasis. The user will want an indentBasis of 1, but get an indentBasis of 0 because \n is technically on the previous line.
### Why this matters

When people copy/paste large bodies of text, they will often drag their cursor up and to the right of whatever they're copying, which will give them that initial newline character. Knowing to avoid this is a workaround people shouldn't have to consider when they're focused on writing whatever they're writing.

### DIFF
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -608,19 +608,23 @@ class Selection extends Model
     {start, end} = @getBufferRange()
     selectionText = @editor.getTextInRange([start, end])
     precedingText = @editor.getTextInRange([[start.row, 0], start])
-    if selectionText.length > 0
-      if selectionText.charAt(0) is '\n'
-        newLineChars = true
-        myCounter = 1
-        while newLineChars and selectionText.length > myCounter
-          if selectionText.charAt(myCounter) is '\n'
-            myCounter++
-          else
-            newLineChars = false
-        startLevel = @editor.indentLevelForLine(@editor.lineTextForBufferRow(start.row + myCounter))
-      else
-        startLevel = @editor.indentLevelForLine(precedingText)
-    else startLevel = 0
+
+    countInitialNewLines = (string) ->
+      newLines = 0
+      counter = 0
+      while counter < string.length
+        if string.charAt(counter) is '\n'
+          counter++
+          newLines++
+        else if counter+1 < string.length and string.substring(counter, counter+2) is "\r\n"
+          counter+=2
+          newLines++
+        else
+          break
+      newLines
+
+    initialNewLines = countInitialNewLines(selectionText)
+    startLevel = @editor.indentLevelForLine(@editor.lineTextForBufferRow(start.row + initialNewLines))
 
     if maintainClipboard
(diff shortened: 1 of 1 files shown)


src/selection.coffee
```diff
@@ -608,7 +608,19 @@ class Selection extends Model
     {start, end} = @getBufferRange()
     selectionText = @editor.getTextInRange([start, end])
     precedingText = @editor.getTextInRange([[start.row, 0], start])
-    startLevel = @editor.indentLevelForLine(precedingText)
+    if selectionText.length > 0
+      if selectionText.charAt(0) is '\n'
```
50Wliu: Does this work on Windows where the newlines are usually `\r\n`?

brentshermana:
Sorry for the delay in response, 50Wliu. After some searching around, it seems like Atom uses /n within the application regardless of what OS you're using, though some people have described this as a bug.

50Wliu:
If this PR doesn't work on Windows, I would definitely make it `\r\n`.  Windows support is slightly lacking at this point unfortunately :/.

brentshermana:
Does \r\n count as a single character, or two? I can dive right back into the code and add a simple branch to account for that.

src/selection.coffee
envygeeks: It'll work up until you save the file.

seesemichaelj:
This is a real issue and currently the *only* reason why I don't like using Atom. How can I help get this integrated (I have not developed for Atom before)?

50Wliu:
@brentshermana looks like this PR broke one of the autoIndent specs - can you please look into that?

klaperman:
@brentshermana, @50Wliu, what is the status of this fix? The current behavior is a daily source of rage for me. Thanks

brentshermana:
It appears that because my changes broke a few specs, so my modification conflicts with what others have decided Atom should do. My code is there for anyone who has the authority to change the specs to do so if they'd like, but unfortunately I've gotten my hands deep in other projects and can't see this issue through to the end.

PR #11271, opened 2016-03-25
Fix typo in error message.

mhw:
I was just reading the source code and noticed this minor cut-and-paste error in the error message.


thomasjo:
Thanks!

PR #11274, opened 2016-03-25
Relative position atom-workspace-axis

simurai:
This let's packages use `position: absolute` and be contained to `atom-workspace-axis` elements.

It fixes https://github.com/thomaslindstrom/color-picker/pull/176 where color-picker relied on `background-tips` to add `position: relative;`, but broke when `background-tips` got disabled.

color-picker

:warning: Since `background-tips` only styled `atom-workspace .horizontal .vertical { position: relative; }`, there is a chance that some packages rely `atom-workspace-axis.horizontal` to be `position: static;`. So something to keep an eye on.


simurai:
Merging this.

Reminder in case something is shifted:

> there is a chance that some packages rely `atom-workspace-axis.horizontal` to be `position: static;`.

PR #11277, opened 2016-03-25
Less disk IO

### LINKED ISSUE #11214: Atom 1.6 Frequently Freezes with Remote Project Folders (All Platforms)
### Description

*Note - Title was edited to reflect that the issue is reported on all platforms, though the following report was originally for Ubuntu 14.04*

Since updating to version 1.6.0 (from 1.5.4, where the problem didn't exist) Atom freezes when trying to open files in networked project folders (i.e. remote folders added to an Atom Project) which are connected via SFTP _(edit: Confirmed to also occur on Samba, SSHFS, etc )._ I have run Atom in safe mode and also tried 1.7 beta and had the same problem.  The networked folders are connected through the built-in "Connect to Server" feature in Ubuntu's Nautilus file browser.  Ubuntu 14.04 was the platform tested.

The **Editor is not responding** message appears after waiting for nearly thirty seconds, and choosing **Keep Waiting** sometimes unfreezes the window, though sometimes not. The problem happens again as soon as another file is selected from the project tree. Atom also commonly freezes when restarting the program with these project folders added, and if it does eventually unfreeze it _will_ soon freeze again when trying to open a project file.

The problem occurs with linked folders on local virtual servers as well as remote servers, so high WAN latency does not appear to be the issue. Atom does not seem to have any CPU spikes during the freeze either as far as I can tell.

### Steps to Reproduce

1. Mount a remote folder via SSH/SFTP in Ubuntu using its Nautilus file browser ('Connect to server' option).
2. Add the folder to an Atom project using the standard folder browser.
3. Once folder is added try to view folder structure or open a file in Atom's tree view.

**Expected behavior:** 
Expected behavior is that the remote project folder operates identically to a local project folder.

**Actual behavior:**
Atom repeatedly freezes for an extended period of time when using the project tree viewer for remote folders and files.

### Versions
Problem occurs in Atom 1.6.0 and later.  Problem did not exist in Atom 1.5.4.  Platform is Ubuntu 14.04 LTS.

joshaber:
Partly fixes #11214.

The biggest problem was that calling `repo.workdir()` on the main thread involved acquiring a lock, which could already be held by a background thread doing disk IO. And on network filesystems, that IO was really expensive, which meant the lock would be held for a looooooong time. See also https://github.com/nodegit/nodegit/pull/968 which removes the need for `repo.workdir()` and friends to acquire the lock.

There are a couple other cases that we found where a lock could be acquired on the main thread:
- [x] `repo.path()`
- [x] `Git.Ignore.pathIsIgnored` (https://github.com/nodegit/nodegit/pull/970)

Even with these changes, performance will still be worse than 1.5.4 because both `GitRepository` and `GitRepositoryAsync` are refreshing their status from git which requires a lot of IO—where in 1.5.4 we only had `GitRepository`.

Ideally `GitRepository` would only do work if some package is using it. Currently that’ll always be the case as `tree-view` hasn’t been updated to use `GitRepositoryAsync` yet (see https://github.com/atom/tree-view/pull/624). But once `tree-view` is updated, that would be a good optimization to do as well.

/cc @nathansobo


joshaber:
We'll need to open a PR on nodegit to make `Git.Ignore.pathIsIgnored` async.

joshaber:
Opened https://github.com/nodegit/nodegit/pull/970.

nathansobo:
Appreciate the 💩  out of you taking this on. Now we just gotta get the tree-view off the old sync repository.

joshaber:
Nodegit 0.12 is [out](https://github.com/nodegit/nodegit/releases/tag/v0.12.0), which includes the async-ifying of `Git.Ignore.pathIsIgnored`. Just waiting on the prebuilt binaries to be available and then I'll bump nodegit and get this fix out.

nathansobo:
:heart:

PR #11284, opened 2016-03-26
Use %F instead of %U for file variable in desktop file

### LINKED ISSUE #2320: Open file from `Open With` dialog messes up file URL
I opened a file from the `Open With` dialog in PCManFM. However, this messes up the file URL and thus causes the editor to open just an empty file.

image

Basically it constructs an URL path in the form of:

`/home/<username>/file:/home/<username>...`

I am using the latest git version on Arch Linux.

Let me know if I can provide you with more information.

jonmast:
The desktop entry specification states that %U is for URLs and %F is for
files. Since atom doesn't support URLs, we should use %F. Fixes #2320.

### DIFF
--- a/resources/linux/atom.desktop.in
+++ b/resources/linux/atom.desktop.in
@@ -2,7 +2,7 @@
 Name=<%= appName %>
 Comment=<%= description %>
 GenericName=Text Editor
-Exec=<%= installDir %>/share/<%= appFileName %>/atom %U
+Exec=<%= installDir %>/share/<%= appFileName %>/atom %F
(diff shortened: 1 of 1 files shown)


StatelessCat:
I confirm this proposal fixes #2320.
Without it, the behavior of Atom is not compliant against the desktop file, according to [the desktop file specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html).

This PR need to be merged.

mnquintana:
Nice catch! Thanks so much for fixing this – sorry it took so long to review. 🚀 🌔

PR #11295, opened 2016-03-28
Take the submodule into account in more instances.

joshaber:
Fixes https://github.com/atom/git-diff/issues/97.


joshaber:
I reviewed all the callsites for `getRepo` and compared them to `GitRepository`, and found a few more spots that needed fixing up.

PR #11296, opened 2016-03-28
Improve build on Windows

nathansobo:
Worked with @damieng on these changes. There may be other improvements to make but we might as well get these in.
- Fix the clean script to actually work on windows.
- Omit any `PATH` entry containing `msbuild.exe`, which confuses `node-gyp`.


nathansobo:
Merging this since the Windows build is green and that's all I changed.

PR #11306, opened 2016-03-29
Warn rather than failing if we detect leaked pathwatcher subscriptions

nathansobo:
For some reason, we're failing intermittently in certain tests with leaked subscriptions. I've spent some time trying to figure out how this could happen, but I can't.
- In the flaky tests, we're leaking paths that are being watched by buffers that I _know_ are being created via `Workspace::open`.
- That means these buffers should be destroyed in `atom.clear()` in the `afterEach` block that's defined in our spec helper to follow every test. As part of destruction, we destroy our watch subscriptions.
- Yet even after destroying all editors and buffers, we still seem to be leaking watches. This leads me to believe there is some subtle error in pathwatcher itself.

Ultimately, it's not clear that the value of this global assertion really warrants this level of hassle. I'm making a pragmatic call to convert this to a warning. If we're seeing it a lot as we develop code that watches paths, then we're probably leaking. If we see it once in a blue moon on CI it's probably just this issue. I don't think it's worth it to invest more time understanding it right now.

/cc @joshaber

### DIFF
--- a/spec/git-repository-async-spec.js
+++ b/spec/git-repository-async-spec.js
@@ -230,9 +230,7 @@ describe('GitRepositoryAsync', () => {
     })
   })
 
-  // @joshaber: Disabling for now. There seems to be some race with path
-  // subscriptions leading to intermittent test failures, e.g.: https://travis-ci.org/atom/atom/jobs/102702554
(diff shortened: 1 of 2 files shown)


joshaber:
👍 💯

PR #11314, opened 2016-03-30
Handle auto-updater errors

as-cii:
This PR propagates the auto-update errors to the renderer process so that we can handle the error state in the about package and show an appropriate error label.

/cc: @thedaniel @joshaber @atom/core


Zireael07:
Yay! Good for those on flaky connections.

PR #11315, opened 2016-03-30
Serialize async git

joshaber:
Another entry in the saga of #11214.

The remaining problem comes down to nodegit's threading model. Nodegit uses a lock to protect against cross-thread memory corruption, race conditions, etc. It also uses libuv's thread pool to do git operations off the main thread.

The problem is that if a lot of git work gets queued up (as happens any time the window comes to the foreground), we can end up with git operations on all of libuv's threads, waiting to acquire the lock. That sucks, because node uses that same thread pool to do all its async work. So we end up blocking node from doing most anything until our git operations are done.

This is most noticeable on slow file systems (e.g., network drives). We’d essentially end up blocking node from doing anything while we waited on IO.

Instead of using nodegit’s locking, we’ll use our own serialization mechanism.

This should also afford us more concurrency in the future if/when we need it.


joshaber:
As of this PR, performance on a network drive feels comparable to 1.5.4 to me. So I'll hesitantly say and 🙏 that this fixes #11214.

PR #11316, opened 2016-03-30
Add request for OS and version to Issue Template

lee-dohm:


damieng:
Wonder if we should add `atom --report` as a command that provides a filled-in portion of the template with atom/os version  etc.

lee-dohm:
That's definitely on my list of things I want to do, yes.

PR #11324, opened 2016-03-31
Add restorePreviousWindowsOnStart setting

as-cii:
Supersedes and closes #9042.

This PR adds a couple of small changes to the above PR, so that:
- :racehorse: Atom opens slightly faster when `restorePreviousWindowsOnStart` is `false`;
- :white_check_mark: Specs ensure that such config parameter gets checked on start, rather than on exit;

Many thanks to @jordanbtucker for providing the original implementation and moving this forward! :sparkles: 

/cc: @atom/core


jordanbtucker:
@as-cii Thanks for implementing your own suggestions and merging! You can close #6605 with this, too.

PR #11325, opened 2016-03-31
Fix emoji rendering on OSX

### LINKED ISSUE #10403: Atom fails to render this emoji --> ▶️ (U+25B6)
It displays as a blank box, and also counts as two characters. Other emojis seem to render fine and only count as one character.

on osx 10.11.3
atom version 1.3.3 (and version 1.4)

the two character problem is perhaps related to https://github.com/atom/atom/issues/8974

Can anyone else reproduce this?

as-cii:
Fixes #10403.

On OS X, we noticed that some emojis were not being displayed correctly. This is a problem with the rendering mechanism that Chromium uses and is present in Chrome 50 as well:

screen shot 2016-03-31 at 11 26 59

However, it seems like it has been fixed on Chrome 51 Canary, where the test above works perfectly:

screen shot 2016-03-31 at 11 26 34
### A workaround

Even with Chrome 50, however, I noticed that GitHub.com was rendering emojis correctly without using PNGs:

screen shot 2016-03-31 at 11 48 34

It seems, in fact, that adding the "Apple Color Emoji" font to the list of font families allows Chrome to pick up the right glyph to render instead of the ugly box:

screen shot 2016-03-31 at 11 31 04

screen shot 2016-03-31 at 11 31 09

As expected, removing it, causes the box to appear (again, this works fine in Chrome Canary 51):

screen shot 2016-03-31 at 11 30 31
screen shot 2016-03-31 at 11 30 24

This is the exact workaround I have employed in Atom as well; indeed, we should be able to remove it  as soon as Chrome 51 gets released and Electron is updated.

/cc: @atom/core

### DIFF
--- a/spec/workspace-element-spec.coffee
+++ b/spec/workspace-element-spec.coffee
@@ -47,9 +47,14 @@ describe "WorkspaceElement", ->
 
     it "updates the font-family based on the 'editor.fontFamily' config value", ->
       initialCharWidth = editor.getDefaultCharWidth()
-      expect(getComputedStyle(editorElement).fontFamily).toBe atom.config.get('editor.fontFamily') + ", 'Apple Color Emoji'"
+      fontFamily = atom.config.get('editor.fontFamily')
+      fontFamily += ", 'Apple Color Emoji'" if process.platform is 'darwin'
+      expect(getComputedStyle(editorElement).fontFamily).toBe fontFamily
+
       atom.config.set('editor.fontFamily', 'sans-serif')
-      expect(getComputedStyle(editorElement).fontFamily).toBe atom.config.get('editor.fontFamily') + ", 'Apple Color Emoji'"
+      fontFamily = atom.config.get('editor.fontFamily')
+      fontFamily += ", 'Apple Color Emoji'" if process.platform is 'darwin'
+      expect(getComputedStyle(editorElement).fontFamily).toBe fontFamily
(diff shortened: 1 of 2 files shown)


thomasjo:
This looks good to me! :boat:

spec/workspace-element-spec.coffee
```diff
@@ -47,9 +47,9 @@ describe "WorkspaceElement", ->
 
     it "updates the font-family based on the 'editor.fontFamily' config value", ->
       initialCharWidth = editor.getDefaultCharWidth()
-      expect(getComputedStyle(editorElement).fontFamily).toBe atom.config.get('editor.fontFamily')
+      expect(getComputedStyle(editorElement).fontFamily).toBe atom.config.get('editor.fontFamily') + ", 'Apple Color Emoji'"
```
50Wliu: This spec will fail on non-OSX now, right?

joshaber:
Nice detective work 👏

spec/workspace-element-spec.coffee
as-cii: Right, but I don't think it's worth adding an extra conditional in the test considering we don't run them on other platforms. Given how easy this change was, I believe it's okay to keep it as-is for now, and revisit it later when adding compatibility for other platforms. :+1:

nathansobo: I think a quick `process.platform` conditional might be worth it in terms of saving us a bit of grief later.

PR #11326, opened 2016-03-31
Fix pinch-to-zoom disabling

### LINKED ISSUE #8850: Atom can be pinch and tap zoomed on OS X
It is possible to zoom in the whole editor on OS X using the pinch zoom gesture on the trackpad. If you do this accidentally I can be really hard to zoom out back. I don’t believe this is an intended feature, is it? 

Occurs in Atom 1.0.13.
bug

as-cii:
Fixes #8850.

There seems to be an issue with how Chromium handles the input data for the `setZoomLevelLimits` API: I suspect there's some kind of guard clause that ignores that function call whenever the `min` and the `max` level limits are the same.

As a stopgap measure, I have changed the max limit to be slightly larger than the minimum so that the configuration gets enforced by Chromium without actually allowing the user to zoom.

/cc: @zcbenz for :eyes: on the electron side.
/cc: @atom/core

### DIFF
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -234,7 +234,7 @@ class ApplicationDelegate
     shell.openExternal(url)
 
   disablePinchToZoom: ->
-    webFrame.setZoomLevelLimits(1, 1)
+    webFrame.setZoomLevelLimits(1, 1.0000000001)
 
   checkForUpdate: ->
     ipcRenderer.send('check-for-update')
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -234,6 +234,9 @@ class ApplicationDelegate
     shell.openExternal(url)
 
   disablePinchToZoom: ->
+    # Setting the same min/max zoom level limit, prevents Chrome to constrain
+    # the zoom level. As a workaround, we use a slightly larger max limit so
+    # that the content cannot be zoomed and the limit is enforced.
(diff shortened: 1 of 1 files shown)


nathansobo:
Wow the magic of floating point. So it doesn't zoom with this value? Can you add a comment to this code line for posterity?

as-cii:
> Wow the magic of floating point. So it doesn't zoom with this value?

Yeah, I cannot reproduce the problem with my Macbook's trackpad.

@atom/feedback: could you try this out and see if it works for you as well? Reproduction steps can be found in #8850. Thanks! :bow:

mnquintana:
Huh, I can't reproduce this on `master` anymore at 0f615ff4e669fccedfa8b8b42dea8d5d7c292e63 ¯\\_(ツ)_/¯

joshaber:
>Huh, I can't reproduce this on master anymore at 0f615ff ¯\(ツ)/¯

Huh, I can't either.

as-cii:
I tried running `script/bootstrap` on master and now I can't reproduce either.

I am a little lost honestly... Maybe some outdated package that caused this to fail? I'll keep trying to reproduce it. :disappointed:

as-cii:
Okay, I think I found what's wrong. Will open a PR in a bit to show the problem and a solution. Thanks everyone for the feedback! :sparkles:

PR #11337, opened 2016-04-01
Save and restore the environment for windows

### LINKED ISSUE #4126: Inconsistent environment and behaviour

joefitzgerald:
This PR more robustly handles the environment in atom-window and the browser window. It:
- Allows an environment to persist when Atom is completely exited and then re-launched, by serializing the environment and restoring it along with the window ("Should process.env be serialized and reused when re-opening an existing project?" from #4126)
- Allows multiple invocations of `FOO=blah atom .` and `FOO=bleh atom .` to result in an updated environment for an existing window ("Should process.env be updated if atom is invoked from the shell for the same directory multiple times?" from #4126)
- Extends #11054
- Partially fixes  #4126

**Note:** This fix is recommended for cherry picking into the beta channel once merged, as it fixes issues present in the 1.7.x beta.

/cc @lee-dohm

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -187,6 +187,7 @@ class AtomApplication
     @on 'application:search-issues', -> shell.openExternal('https://github.com/issues?q=+is%3Aissue+user%3Aatom')
 
     @on 'application:install-update', =>
+      @saveState(true)
       @quitting = true
       @autoUpdateManager.install()
 
@@ -213,6 +214,7 @@ class AtomApplication
     @openPathOnEvent('application:open-license', path.join(process.resourcesPath, 'LICENSE.md'))
 
     app.on 'before-quit', =>
+      @saveState(true)
       @quitting = true
 
     app.on 'will-quit', =>
@@ -318,6 +320,11 @@ class AtomApplication
 
     ipcMain.on 'execute-javascript-in-dev-tools', (event, code) ->
       event.sender.devToolsWebContents?.executeJavaScript(code)
+
+    ipcMain.on 'environment-updated', (event, env) =>
+      window = @windowForEvent(event)
+      window.setEnvironment(env) if window? and env?.PATH?
+      @env = env if env?.PATH?
 
   setupDockMenu: ->
     if process.platform is 'darwin'
@@ -511,7 +518,9 @@ class AtomApplication
     for window in @windows
       unless window.isSpec
         if loadSettings = window.getLoadSettings()
-          states.push({initialPaths: loadSettings.initialPaths, env: process.env})
+          env = window.getEnvironment()
+          env = process.env unless env?
+          states.push({initialPaths: loadSettings.initialPaths, env: env})
     if states.length > 0 or allowEmpty
       @storageFolder.storeSync('application.json', states)
 
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -520,7 +520,10 @@ class AtomApplication
         if loadSettings = window.getLoadSettings()
           env = window.getEnvironment()
           env = process.env unless env?
-          states.push({initialPaths: loadSettings.initialPaths, env: env})
+          if loadSettings.initialPaths.length > 0
+            states.push({initialPaths: loadSettings.initialPaths, env: env})
+          else
+            states.push({initialPaths: loadSettings.initialPaths})
     if states.length > 0 or allowEmpty
       @storageFolder.storeSync('application.json', states)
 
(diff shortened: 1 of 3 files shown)


lee-dohm:
I talked about this with the team this morning and there's some concern that persisting the environment across an exit is the right thing to do. Can you clarify what this is useful for? What if my environment changes between invocations of Atom? Shouldn't Atom always have the most up-to-date environment?

Regarding updating the environment of an already running window from the command line, is this much more useful than changing it via the console in the Developer Tools?

joefitzgerald:
> What if my environment changes between invocations of Atom?

This logic only kicks in when Atom is launched via Dock / Finder / Spotlight. When it is launched from a terminal, the environment is reset to that of the terminal. I suppose there is an edge case here where a user launches a bare `atom` from the terminal (i.e. without an initial path); this should cause a reset of the environment too.

The purpose of this part of the PR (there are others) is: suppose you start atom from the terminal. The environment is set very specifically to work for a specific project folder / window. Then you quit atom, and relaunch it (accidentally forgetting to launch from the terminal, and instead launching from the Dock or Spotlight or Finder). Your prior window(s) launch and the environment for each remains the way you intended. You can resume your work without futzing with closing all Atom windows, and launching each one individually from the right terminal window.

Without this change, the environment for the Atom window is broken in many scenarios depending on the flow with which the user launches Atom windows. This presents a cognitive burden to the user who has to grok "how did I launch this last time??" (if they even remember).

> Regarding updating the environment of an already running window from the command line, is this much more useful than changing it via the console in the Developer Tools?

Yes, because a user shouldn't have to open the developer tools to update their environment. If the user invokes `atom .` or `atom somedir` multiple times, what is the downside of allowing the environment to change? It allows the user to control the environment in their original environment and signal to atom that it has changed by relaunching atom with the changes.

lee-dohm:
Modifying the already open environment is a much easier sell than persisting an environment. Perhaps it should be in its own PR?

PR #11352, opened 2016-04-02
Pending items shouldn't be made permanent before being replaced

BinaryMuse:
See [this comment discussion](https://github.com/atom/atom/pull/11001#issuecomment-204294465). Unfortunately the solution suggested there will not work, as it breaks other functionality. So this is still 🚧.

/cc @kuychaco

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -551,7 +551,6 @@ class Pane extends Model
       @emitter.emit 'will-destroy-item', {item, index}
       @container?.willDestroyPaneItem({item, index, pane: this})
       if @promptToSaveItem(item)
-        @pendingItem = null if item is @getPendingItem()
         @removeItem(item, false)
         item.destroy?()
         true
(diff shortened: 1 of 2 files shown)


BinaryMuse:
@kuychaco I *believe* that it would be safe (and more correct) to make it so that, when pending items are replaced with other pending items, `Pane::itemDidTerminatePendingState` is not called for the original item. Can you think of anything I'm forgetting?

src/pane.coffee
```diff
@@ -442,15 +442,14 @@ class Pane extends Model
       if typeof item.onDidTerminatePendingState is "function"
         itemSubscriptions.add item.onDidTerminatePendingState =>
           @clearPendingItem() if @getPendingItem() is item
-      itemSubscriptions.add item.onDidDestroy => @removeItem(item, false)
```
BinaryMuse: Removed this because it was duplicated from line 441. o_o

kuychaco:
This looks good to me 👍. I agree that this seems safe and more correct. Nothing comes to mind that you're forgetting @BinaryMuse. Great catch @paulpflug, thanks for pointing this out.

PR #11085, opened 2016-03-07
Add option editor.showCursorWhenSelecting - Cursor Visible While Selecting

colelawrence:
This is a very common feature for most IDEs, I've used the same identifier as CodeMirror's option in this case `showCursorWhenSelecting`.

This is a very small fix for something that cannot be addressed with a plugin AFAIK, thank you.

image

PR #11353, opened 2016-04-02
Show Cursor When Selecting - Add setting to show cursor when selecting

colelawrence:
Redo #11085  commits with passing tests.

screenshot from 2016-04-02 14-49-29

Use the option `editor.showCursorWhenSelecting`

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1272,6 +1272,18 @@ describe('TextEditorComponent', function () {
       expect(cursorNodes[0].style['-webkit-transform']).toBe('translate(' + (Math.round(8 * charWidth)) + 'px, ' + (6 * lineHeightInPixels) + 'px)')
     })
 
+    it('does render cursors that are associated with non-empty selections when editor.showCursorWhenSelecting is true', async function () {
+      atom.config.set('editor.showCursorWhenSelecting', true)
+      editor.setSelectedScreenRange([[0, 4], [4, 6]])
+      editor.addCursorAtScreenPosition([6, 8])
+      await nextViewUpdatePromise()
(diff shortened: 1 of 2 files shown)


ZombieHippie:
Build error unrelated to changes:
https://travis-ci.org/atom/atom/jobs/120327020#L247

lee-dohm:
I've confirmed this works. A few things that would need to happen before this could be accepted:

1. The setting would need to be initialized and documented with all the other editor configuration settings
1. The setting should not be scope-specific
1. The setting should be a property on the `TextEditor` and observe the configuration value like with `subscribeToTabTypeConfig`

ZombieHippie:
Thank you for the additional information. May I ask why 2 is important,
because it seems like it would restrict the user config more than
necessary. I don't entirely understand scoped settings but I want to be
able to have different settings based on the grammars used.

Thank you!

lee-dohm:
Is it useful to have the cursor visible while selecting in certain file types but not in others? There are three levels of configurability to settings:

1. Global settings - set once and it is always that way until changed, example "Back Up Before Saving"
1. File-level scope-specific settings - Set by the recognized language of the file, examples "Tab Length", "Preferred Line Length", "Soft Wrap"
1. Scope-specific settings - Set by the recognized language at the position of the cursor, example "Non Word Characters"

Perhaps you could make a case for file-level, but not pure scope-specific as you have it coded now. I think that would be too confusing for people, to have their cursor disappear and reappear as they are selecting across language boundaries.

sadick254:
Thank you for your pull request!

There has been no activity on this PR since @lee-dohm's comment above. I'm going to close this but don't hesitate to reach out if you want to continue working on this PR, we'll be happy to reopen the pull request.

Issue #1766, opened 2014-03-18
Accessing files on network drives is slow


izuzak:
Reported via Halp:
- support/6290a694aa0611e38d8242e11b72b07d

> Accessing files on network drives appears to be much slower than Sublime for the same mount & file. Additionally, the "Keep Waiting" decision box pops up almost every time.

I don't have a network drive, so I didn't verify this. :car: 

Edit: more reports:
- support/4e6fe392af5a11e38b62f01b0c57459b

Nikita240:
+1, slow for me too.

Makes webdev really difficult.

GM-Alex:
Same for me with 0.177.0

toddbu:
Same for me on the last few releases including 0.187.0.  I'm running SMB shares on a CentOS server running under VirtualBox on the same Mac that I'm running Atom on.  It's really random but happens once or twice a day.  When I close the process Atom stays running and I have to force quit the app.

pheuter:
Same on Atom 0.188.0, having a hard time opening an smb drive from virtualbox on OS X.

JohnRoach:
I will like to inform you that this(loading mounted drive folders or files) also is slow in Windows 7.

benjamindean:
In addition, `fs.realpathSync` is, actually, slow for almost everything. This happens only when opening file using context menu (Open With) on remote server (using nautilus in my case). Don't know the reasons, but that's was the issue for a long time. This is, probably, avoidable by not using something other than `readFileSync` and `writeFileSync` when path are not in the System context. Anyway, nothing as fast as a native implementation. 
Nothing bad happens when opening the same file from Atom, thou.

medfreeman:
Same on Ubuntu 14.04, accessing a SFTP server through Nautilus.

eikeco:
I am accessing files via a Parallels Virtual Machine and having the same problem. I'm on a Mac.
Atom Version 1.0.0.

Even when a file is open it takes a moment to catch up with what you type.

leonardfactory:
Same here with 1.0.0. Parallels VM with Windows 8, I'm working on aMmac.

jryanad:
Having an issue as well. Atom 1.0.7 with running Virtual Box and connecting via smb on Mac 10.10.5. Just continually loads most times and have to restart the process to get it to "kick in".

incognick:
[...]

PR #11367, opened 2016-04-04
Learn configuration of FS paths for which to skip VCS integration.

adamfranco:
This patch allows users to configure paths such as `/mnt/` or `/Volumes/`
which map to network shares or other targets with slow I/O performance.
By skipping VCS integration on these paths, I/O is significantly reduced at
the cost of not having VCS integration and allowing Atom to at least be usable
as a text-editor when pointed at mounted network shares.

This is a attempt to mitigate some of the effects of #1766. Please consider this
a somewhat naive proposal for a solution that could use some more discussion
around ways to solve #1766. 

Feedback requests:
- Naming of the config variable
- The way this is wired into the git-repository-provider (maybe the `Project` or another location would be better).
- Alternative ways to stop Git integration on a per-project basis.

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -18,6 +18,12 @@ module.exports =
         default: true
         title: 'Exclude VCS Ignored Paths'
         description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, projects using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
+      skipVcsPaths:
+        type: 'array'
+        default: []
+        items:
+          type: 'string'
+        description: 'List directories under which VCS/Git integration will be skipped.'
       followSymlinks:
         type: 'boolean'
         default: true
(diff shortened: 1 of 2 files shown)


adamfranco:
Note: I've been using this patch against a Samba share to a linux node over a gigabit network and Atom is now quite usable with a Project Folder of 9500 files. Without this patch FS operations get backed up waiting for Git status-refresh to complete and opening/saving files takes many minutes. There is still a bit of I/O caused by the tree-view listing files and a full traversal of the entire parent-path on file-save (which I haven't located the source for yet), but the super-heavy I/O related to git disappears.

adamfranco:
Other questions for which feedback would be helpful:

- Should the default be empty `[]` or `['/mnt/', '/Volumes/']` to skip Git integration for network shares on Linux/OSX by default? What do you think makes more sense?
- Is there a sane default path for network shares on Windows?

adamfranco:
An additional note: By placing this check in the `GitRepositoryProvider` rather then higher up the stack in the `Project` this patch also disables Git operations for modules like tree-view (as requested in atom/tree-view#784).

sadick254:
Hey @adamfranco I strongly think that performance issues with network shared mounts should be addressed without disabling any functionality. We should find a way to optimise for network mapped drives. We placed this issue on our tech backlog and when we have capacity we will surely tackle it. For now, I have to close this since it doesn't reflect the direction we would like to go.

adamfranco:
Thanks for the note, @sadick254. What's the issue or task to follow for updates on that work?

sadick254:
I am going to create an Issue and post the links here.

PR #11369, opened 2016-04-04
:racehorse: Avoid setting hidden input value on textInput

hellendag:
Atom currently sets the `value` of the input on every `textInput` event, in an effort to appropriately handle changes made via the OSX diacritic menu (for accents, umlauts, etc).

The drawback of this is approach is that updating the value of the input will trigger layout and a subsequent layer tree update.

To resolve this, here is my proposal:
- Track a flag for `keypress` events. When the diacritic menu is used, there are two `textInput` events, with no `keypress` in between. Therefore, when no `keypress` has occurred just prior to a `textInput`, the editor model can select the previous character to be replaced by the new accented character.
- Track a flag for `compositionstart` events. When a user is in IME mode, the diacritic menu cannot be used, so the editor can skip the backward selection.

Test Plan:

Tested in a plaintext file.
- Type Latin characters, verify proper character insertion.
- Press and hold <kbd>a</kbd>. Diacritic menu appears. Select an option using the keyboard or mouse. Verify that the `a` is replaced by an accented `a`, with no extra characters.
- Type test strings in Katakana, 2-Set Korean, Telex (Vietnamese), Simplified Pinyin. Verify that characters are inserted correctly while composing, and after committing strings.

cc @ssorallen @nathansobo

### DIFF
--- a/src/text-editor-component.coffee
+++ b/src/text-editor-component.coffee
@@ -244,12 +244,27 @@ class TextEditorComponent
   listenForDOMEvents: ->
     @domNode.addEventListener 'mousewheel', @onMouseWheel
     @domNode.addEventListener 'textInput', @onTextInput
-    @domNode.addEventListener 'keypress', @onKeyPress
     @scrollViewNode.addEventListener 'mousedown', @onMouseDown
     @scrollViewNode.addEventListener 'scroll', @onScrollViewScroll
 
+    @detectAccentedCharacterMenu()
     @listenForIMEEvents()
     @trackSelectionClipboard() if process.platform is 'linux'
+
+  detectAccentedCharacterMenu: ->
+    # We need to get clever to detect when the accented character menu is
+    # opened on OS X. Usually, every keydown event that could cause input is
+    # paired with a corresponding keypress. However, when pressing and holding
+    # long enough to open the accented character menu causes additional keydown
+    # events to fire that aren't followed by their own keypress and textInput
+    # events. We exploit this by assuming the accented character menu is open
+    # until a subsequent keypress event proves otherwise.
+    @domNode.addEventListener 'keydown', (event) =>
+      unless event.keyCode is 229 # Skip keydown events associated with IME compositionupdate events
+        @openedAccentedCharacterMenu = true
+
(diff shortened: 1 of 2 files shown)


50Wliu:
Looks like this broke some specs.

nathansobo:
Taking a look at the test failures.

nathansobo:
@hellendag I noted in the spec that the `keypress` event is deprecated. Did you look for any other non-deprecated events with a similar behavior? Odds are pretty low that Chrome will drop this event any time soon but it would be nice if we could avoid building on deprecated functionality.

nathansobo:
I guess `textInput` isn't standardized either. Forgot about that 😬. I've never looked into the [edit events](https://www.w3.org/TR/uievents/#event-type-beforeedit) that replace them but perhaps that's a topic for another PR. Disregard my previous comment.

nathansobo:
@hellendag I think I've found a hole. When you open the accented character menu, you can use the left and right arrows to cycle through the options in addition to pressing a number or the mouse. I wonder if we need to filter out the left and right arrow keys when setting the keypress flag?

hellendag:
Yeah, I'm still hoping that `beforeEdit`/`beforeInput` gets implemented at some point. In the meantime, at least we've got `textInput`. :)

nathansobo:
Our internal build passed and the Travis failure is due to connectivity issues. Going to merge.

src/text-editor-component.coffee
```diff
@@ -319,26 +321,27 @@ class TextEditorComponent
...
+  onKeyPress: (event) =>
+    @detectedKeyPress = true
+
   onTextInput: (event) =>
     event.stopPropagation()
-
-    # If we prevent the insertion of a space character, then the browser
-    # interprets the spacebar keypress as a page-down command.
-    event.preventDefault() unless event.data is ' '
+    event.preventDefault()
```
nathansobo: @hellendag Removing this guard caused a regression where we interpret spaces as page downs (per the comment). After restoring the code preventing the `preventDefault` call for spaces, IME and the press-and-hold menu still seem to work. Am I missing anything?

Issue #9561, opened 2015-11-13
Atom 1.2.0 (1.3.0-beta) rendering is slow


alexbeletsky:
Sorry, for such subjective and non-technical bug description.

I've installed Atom 1.2.0 (as well as Atom 1.3.0-beta) yesterday. Release notes saying about performance improvements (both rending and startup time) that are hardly could be noticed.

1.2.0 (and 1.3.0-beta) rendering is very slow. What I mean, if I click some file, I first see the "grey" text and only after the coloring syntax applies. So, it blinks and for me it's really hurting experience.

Had to downgrade to 1.1.0 again. If I can provide more information, please let me know.

benogle:
To be clear, it is only slower when you are initially loading a file? Or is it slower in other areas as well? What kind of files are you loading? How large? On a fresh window open? Or is it slower the longer the window is open?

Can you try starting atom in [safe mode](https://atom.io/docs/latest/hacking-atom-debugging#check-if-the-problem-shows-up-in-safe-mode)?

If safe mode is faster, can you paste in the contents of `apm list`?

alexbeletsky:
@benogle 

1. Yes, it's slow on initial file rendering. With 1.1 I'm having the same effect, so then I open the file it always "blinks", but on 1.2 the blink is much longer, so it's really annoying.

2. My files are typically `.js` (Javascript-JSX), `.scss` and `.html` relatively low size.

3. I've tried safe mode. I still see a little blinks, but in general it works much more faster. 

My [apm list](https://gist.github.com/alexbeletsky/20e2da5f02b97b564e50).

patoi:
I have same issue with 1.2 on UHD (3840x2160), but not on HD (1920x1080). One laptop with dual screen HD + UHD (HDMI1.2@30Hz).

patoi:
There are some CPU and TimeLine data from DevTools (zip 2.8MB):
https://drive.google.com/open?id=0B6pleSdQXj0YV245UXc4ZVQxblE

benogle:
@alexbeletsky ok, it may be a package slowing things down if it's faster in safe mode. 

My guesses for culprits would be: minimap, pigments, linter, with enhanced-tabs, and editorconfig coming in second. Can you try disabling them one by one and reloading? Is there a difference?

patoi:
I've turned off all Community Packages, and all Core packages but it is slow rendering on UHD.
(settings-view, tabs and tree-view enabled)

[...]

PR #11370, opened 2016-04-04
Tokenize visible buffer chunk immediately

hellendag:
When opening a new file, there is currently a flash of unstyled content ("FOUC") between the point where the file is first rendered and when syntax highlights are rendered. This gap can be hundreds of milliseconds, and the flash can be jarring.

To resolve this, I propose paying some tokenizing costs up front so that the first rendering of the file is already highlighted. Tokenize the first chunk, then defer the rest.

This means implicitly adding 20ms or so to the initial render pass, but there is no longer a jarring style change.

Test Plan:

Open JS files of varying sizes. Verify that the initial render of the file is already highlighted, with no flash.

### DIFF
--- a/src/tokenized-buffer.coffee
+++ b/src/tokenized-buffer.coffee
@@ -154,6 +154,7 @@ class TokenizedBuffer extends Model
     @emitter.emit 'did-change', event
 
   setVisible: (@visible) ->
+    @tokenizeNextChunk()
     @tokenizeInBackground() if @visible
 
   getTabLength: ->


50Wliu:
Refs #9561.

Zireael07:
Nice change.

hellendag:
I have no idea how to make this pass the build. There doesn't seem to be much consistency about it.

massa84:
Thanks @hellendag , that FOUC was driving me crazy ( enough to force me to write a quick css fix that delayed the display of the editor scroll view by 0.1s 😮 )
Think this pr should be merged asap.

50Wliu:
/cc @atom/feedback

This change looks good to me as it reduces the amount of visual change after the window is loaded.

lee-dohm:
I'm concerned about the edge cases of this, for example with files with excessively long lines, this might cause the text to not be displayed for a noticeably long time. This is something that I would consider much more jarring than waiting a moment for the text to be highlighted.

as-cii:
> I'm concerned about the edge cases of this, for example with files with excessively long lines, this might cause the text to not be displayed for a noticeably long time. This is something that I would consider much more jarring than waiting a moment for the text to be highlighted.

I think I agree with this, although we should consider that in the context of a long line Atom would still block synchronously after rendering the text, making the UI completely un-responsive. Anyways, we specify a maximum of 100 tokens per line [here](https://github.com/atom/atom/blob/master/src/grammar-registry.coffee#L18), which should supposedly shield against such edge cases. 

@hellendag: have you tried performing some tests on that specific scenario? Also, it'd be awesome to add a spec to `tokenized-buffer-spec` that captures this new behavior, as suggested by @50Wliu.

jordwalke:
Can this be merged? This is driving me nuts.

lee-dohm:
@jordwalke No, the open questions that @as-cii posed haven't been answered.

jordwalke:
@as-cii seems to provide pretty convincing justification for this change -  `GrammarRegistry.maxTokensPerLine` supposedly guards against extremely large blocking times, no?

lee-dohm:
@jordwalke I'm specifically talking about:

> have you tried performing some tests on that specific scenario? Also, it'd be awesome to add a spec to tokenized-buffer-spec that captures this new behavior, as suggested by @50Wliu.

These have not been answered or addressed.

Issue #8295, opened 2015-08-10
Overlay items flicker when near the edge of the window


ssfrr:
This happens frequently when using the [Hydrogen package](https://github.com/willwhitney/hydrogen), which displays the return values from evaluated code as overlays on the window. It seems to be exacerbated when there are more overlays. It looks like the layout engine is trying to figure out where to render them but the act of rendering changes the location solution, so they start flickering like crazy.

Originally filed as an issue on Hydrogen: willwhitney/hydrogen#93, but it seems this is an issue with the Atom rendering.

izuzak:
@ssfrr Just to make sure we have all the relevant information -- can you please record a short GIF which clearly demonstrates the problem, and also provide the exact steps which reproduce this for you (+ version of Atom and OS)?

ssfrr:
This gif doesn't really capture the speed of the flicker (it's very fast). Video is [here] (https://vimeo.com/136067838) (and even that seems slower then it looks on my screen).

![atom_flicker](https://cloud.githubusercontent.com/assets/734386/9216613/1497f656-408a-11e5-8f3e-fb3ea649ad81.gif)

I'm on OSX, using Atom 1.0.5.

It seems to take a while to reproduce, and seems to be exacerbated when I have a lot of overlay items, which happens commonly after working in Hydrogen for a while. Unfortunately I don't have a specific sequence of steps, but if I start to see some commonalities I'll post back.

simurai:
I couldn't get the Hydrogen package to install, but based on the video, it looks like some GPU layer issue.

If you're familiar with the DevTools, you could try to "Inspect Elements" and see what class that overlay has. It might be [.hydrogen.output-bubble](https://github.com/willwhitney/hydrogen/blob/master/styles/hydrogen.less#L14). If so, here some random things you can add to your `styles.less` and see if it helps.

```less

// uncomment only one line at the time, not all together

.hydrogen.output-bubble,
atom-text-editor::shadow .hydrogen.output-bubble {
  //backface-visibility: hidden;
  //z-index: 0; // also try increasing the number by one: 1, 2, 3 etc.
  //will-change: transform;
}

atom-text-editor {
  //overflow: hidden; // this might cause autocomplete to be cut off
}
```

izuzak:
Thanks, @ssfrr. 

@nathansobo Does the behavior shown in the GIF above ring any :bell: for you?

ssfrr:
[...]

PR #11388, opened 2016-04-06
:bug: Fix window size is not updated on resize

ypresto:
https://github.com/atom/atom/issues/8295#issuecomment-206484993

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2251,6 +2251,17 @@ describe('TextEditorComponent', function () {
 
         expect(overlay.style.left).toBe(windowWidth - itemWidth + 'px')
         expect(overlay.style.top).toBe(position.top + editor.getLineHeightInPixels() + 'px')
+
+        // window size change
+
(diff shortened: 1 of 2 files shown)


50Wliu:
Hi @ypresto, can you please add a spec to test this change?  Thanks!

ypresto:
wow.. test-editor-component-spec.js is too long...

ypresto:
Added test!

50Wliu:
Woo, thanks!

ypresto:
Fixed conflict..!

ypresto:
one more rebase..!

nathansobo:
Thanks for tracking this down. What do you think about fixing this via a window `resize` event handler? We had one at some point and I'm not sure how it was removed. That would keep the `pollDOM` method lighter which will save battery power.

nathansobo:
I see, looks like we bind the `resize` event in the view scheduler and just call the pollers. These properties should probably be pretty cheap to read so I'm just going to merge. Thanks very much for finding and fixing this. Pretty embarrassing bug! ✨ ✨ ✨

nathansobo:
Fixes #8295

Issue #11396, opened 2016-04-07
Tree view doesn't update git status for files not in the root


joshaber:
I suspect this was introduced in https://github.com/atom/atom/pull/11280, which is why I'm filing it here and not on `tree-view`.

1. Edit a file in a subdir of the project (e.g., `some/cool/file.txt`).
2. Commit the file from the command line.
3. Switch back to Atom.

Expected: The file's not shown as modified anymore.
Actual: The file's still marked as modified.

PR #11397, opened 2016-04-07
Defer the callback to the next tick.

joshaber:
This gives `GitRepository` the chance to clear its path cache before the callback is invoked. Otherwise reads from the cached status state within the callback would be wrong.

Fixes https://github.com/atom/atom/issues/11396.


joshaber:
Build is failing because of some unrelated lint. Fixed in https://github.com/atom/atom/pull/11398.

PR #11398, opened 2016-04-07
Fix build lint

joshaber:
Looks like this was introduced in https://github.com/atom/atom/pull/11369. :see_no_evil:


joshaber:
(Another reason why having flaky builds is bad 😬)

Issue #5262, opened 2015-01-26
atom:// url package installation


thedaniel:
Elaborating on #2037:

To allow some interesting integrations with atom.io, we should allow users to initiate package installation from the `atom://` url handler. There are a number of questions to answer:
- [ ] What work in atom core & atom-shell is required to support this on all 3 platforms?
  - Windows registry doc: https://msdn.microsoft.com/en-us/library/aa767914(v=vs.85).aspx
  - The current top answer to this is relevant, showing how to use gconf on a system with gnome: http://stackoverflow.com/questions/1947209/registering-a-url-protocol-handler-in-a-multiple-platforms
- [ ] What are the user concerns around security and predictability and how can we address them?
  - Users should always know what is being installed and have to explicitly confirm installations initiated outside the editor
- [ ] What's a general-purpose UI look like for this, given the editor can be in any state when installation is initiated?
  - Probably a modal in the active window?

zcbenz:
Atom already has `atom://` registered in Info.plist:

```plist
	<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>atom</string>
			</array>
			<key>CFBundleURLName</key>
			<string>Atom Shared Session Protocol</string>
		</dict>
	</array>
```

And it seems that there is already some meaning of `atom://` url handler:
https://github.com/atom/atom/blob/master/src/browser/atom-application.coffee#L402

thedaniel:
@zcbenz That's mac-only though, we don't register the url with the Windows registry / try to set gnome/kde config for it do we?

zcbenz:
@thedaniel Right, we only have it on Mac.

Registering URL on other platforms are quite similar, they are implemented usually by writing some settings to some registry, and the program specified will be called with the URL appended in command line. The work will mostly be in installers/packages.

thedaniel:
Fortunately we can write and test all the app code for this without having to wait for the cross-platform URI handler stuff to be done. I probably won't have time this week but I'll wireframe or prototype this.

jeancroy:
Btw GitHub Desktop use that feature for the `Clone In Desktop` feature. So you may have some solution in house.

benogle:
> Btw GitHub Desktop use that feature for the Clone In Desktop feature. So you may have some solution in house.

[...]

Issue #2037, opened 2014-05-07
open files with atom:// handler


matreyes:
It should open an html handler like:
The handler opens atom, but doesn't works with params.
atom://open?url=file://Users/myuser/myapp/myfile.rb&line=32&column=12

it would be very useful for rails_panel !

jaredmoody:
+1

jmuheim:
:+1: This is the only thing keeping me from abandoning TextMate2 in favour Atom at the moment.

martinratinaud:
agreed

lifeiscontent:
@izuzak any updates on this?

izuzak:
> @izuzak any updates on this?

@lifeiscontent I think @thedaniel has been making some progress in this area, see for example https://github.com/atom/welcome/pull/38 and https://github.com/atom/settings-view/pull/614.

lifeiscontent:
@izuzak thanks! :+1:

EloB:
+1

shanept:
+1

WizardOfOgz:
I would also like to see this feature built into Atom directly. Until that happens, and if you are on OSX, you can use this project I started which will open links with the atm:// protocol in Atom. https://github.com/WizardOfOgz/atom-handler

shanept:
Interesting, WizardOfOzg - I'm working on a plugin to give us an atm:// handler too

jmatsushita:
Apart from @WizardOfOgz OSX handler there is also a Linux based handler (which I haven't tested) https://github.com/eclemens/atom-url-handler

UziTech:
Looks like they just merged the pull request for this :tada: #11399 

Hopefully out soon!!

pabloab:
Going to [Atom home page](https://atom.io/#teletype) and clic on [atom://settings-view/show-package?package=teletype](atom://settings-view/show-package?package=teletype) don't do anything on my Ubuntu 16.04 LTS. So [I do this](https://askubuntu.com/a/1023143/349837):

- Edit `/usr/share/applications/atom.desktop`
- Append `x-scheme-handler/atom;` to line starting `MimeType=text/plain;`
- Register the protocol: `xdg-mime default atom.desktop x-scheme-handler/atom`

Now browser runs atom but just open a new file. With `xdg-open` on CLI: _Client failed to connect to the D-BUS daemon:_ and _Failed to connect to socket /tmp/dbus-2i6A77jamf: Connection refused_. I'm missing something or should file a new bug...?

[...]

PR #11348, opened 2016-04-02
Update Electron to 0.36.12

winstliu:
Continuation of #10983.

Closes #10983.
Fixes #11198.

/cc @drewmnoel

PR #11399, opened 2016-04-07
Package URL handlers - mapping atom:// URIs to package functions

BinaryMuse:
See update below: https://github.com/atom/atom/pull/11399#issuecomment-330396536

---

This PR implements the `MessageRegistry`, which is similar to the `CommandRegistry` in that it maps string types of some kind to callback functions, but it differs in that messages originate from outside the application via `atom://` URIs handled via [the `open-url` event](http://electron.atom.io/docs/v0.37.4/api/app/#event-open-url-os-x).

This partially implements https://github.com/atom/atom/issues/5262, provides a mechanism for hitting things like https://github.com/atom/settings-view/pull/614 from outside the application (e.g. "View in Atom" links on package pages on the Atom site), and would enable package authors to implement things like https://github.com/atom/atom/issues/2037 for external debuggers, etc.

There is still some additional work to do around registering Atom as a system-level handler for `atom://` URIs on non-Mac systems, but that probably will be installer-based work.

/cc @atom/core @atom/feedback for review. Some open questions I have are:
- I restricted messages to URIs starting with `atom://atom` as we already [have a mechanism that utilizes URIs with hosts that match package names](https://github.com/atom/atom/blob/822e0c95108aadb91631570e386d8cf96d7cfdd0/src/browser/atom-application.coffee#L550-L557) to perform some work. (What is this used for? I'm not familiar with it.) I'd love to get rid of this, as it means we can't use any host that might match a package name. If anyone has bright ideas for URI schemes, that'd be awesome. (I thought maybe something like `atom://.message/...`?)
- Right now, the callback function is called with the string message (just the `namespace:name` portion) and the params, but maybe it makes more sense to pass the entire URI, or the parsed URI object.


lee-dohm:
> already have a mechanism that utilizes URIs with hosts that match package names to perform some work. (What is this used for? I'm not familiar with it.)

This is used for importing resources into Less without needing to know the path where packages are stored: https://github.com/lee-dohm/red-wavy-underline/blob/a7c36b5b8790016f81f7cf48d840ed44333df7fa/styles/red-wavy-underline.less#L2

zeke:
`red-wavy-underline` is a great package name.

BinaryMuse:
It looks like https://github.com/electron/electron/pull/4896#issue-143098316 took care of the work for allowing apps to be protcol handlers on OS X and Windows. Linux is still an open question for me.

BinaryMuse:
> This is used for importing resources into Less without needing to know the path where packages are stored

Lee and I discussed this and it's a slightly different thing.

thedaniel:
> It looks like electron/electron#4896 (comment) took care of the work for allowing apps to be protcol handlers on OS X and Windows.

Does that mean this PR is now waiting on an Electron upgrade to 0.37.4+?

BinaryMuse:
@iolsen This is just a reminder ping that we need to chat about this. :)

BinaryMuse:
@thedaniel That's correct; we would get this from #11474, which we can move forward on now that #11348 has been merged.

It looks like setting a default protocol handler via the Electron API [is not supported by Linux](http://electron.atom.io/docs/v0.37.8/api/app/#appsetasdefaultprotocolclientprotocol-os-x-windows); I'm actually not sure what it would take to do such a thing on Linux, or if it's even reasonable to be able to do so in a way that works for everyone.

bastilian:
@BinaryMuse I did a bit of digging around and it seems that [xdg-utils](https://www.freedesktop.org/wiki/Software/xdg-utils/) is the best way to add a protocol handler on Linux, which seems to be used or integrated in most popular desktop environments XFCE, KDE and Gnome.

nathansobo:
I'd prefer a name like "external URL handler" or something similar. My concern about the term "message registry" is that it's a bit general, and presumably there are other ways to message the application such as an OS-provided RPC facility. A name that more closely mirrors the actual implementation might be clearer.

I would definitely like to figure out if we're using the existing top-level namespace for anything critical. Even if we are, if we could move those URLs to something like `atom://packages/my-package`, I'd prefer it. Just like for services, I think we should encourage users to namespace their external URL handlers based on their package name... so for example if I had a package named foo, I would register the namespace `atom://foo/...`. I also think the handler should take the entire URL. Perhaps absent the `atom://` prefix?

As for registering these handlers, I think we should consider a setup where handlers are registered *statically* via the `package.json`. We've moved in this direction with other global facilities in Atom. The big advantage to doing something static is that you don't have to activate the package or evaluate any code to know a handler has been registered. This makes it easier to defer loading for packages that only define these kinds of handlers.

Obviously, some amount of arbitrary code execution might be required to interpret a given URL, but I don't think it would be too onerous to require packages to at least register for a specific URL prefix.

jgebhardt:
Just wanted to re-iterate that having this functionality would be huge. I already used it to prototype a few use cases for potential handlers in Nuclide, and it would enable some very powerful integrations. 

@nathansobo's point about renaming the API makes sense, since `MessageRegistry` is not super intuitive imho. 

Other than deciding on the name, registry, and namespacing convention, is anything else (maybe the Linux issue?) blocking this from making it into a release?

DavidSnider:
If we could make this happen, I would be suuuper happy

bolinfest:
@BinaryMuse Any chance you'd be willing to pick this up again soon?

bolinfest:
I agree with @nathansobo's point that these should be able to be registered _statically_. I think that this could also sidestep the conversation about the `MessageRegistry` naming for now. We would only have to bikeshed the name of the property in the `package.json` :).

In that vein, it could conceivably be exposed as a service, though up until now, I don't believe that services could include their own static configuration (though really, why not)?

bolinfest:
Actually, how bad would it be if it were just a service and you couldn't specify the prefix? That is, what if you were forced to use a prefix that matched the name of your extension? So if you had the following as your extension's `package.json`:

```
{
  "name": "an-example",
  "providedServices": {
    "atom.url-handler": {
      "versions": {
        "1.0.0": "onAtomURI"
      }
    }
  }
}
```

Then if Atom received a URI of the form `atom://url-handler/an-example/...`, then `an-example` would be activated and the URI would be passed as an argument to the registered provider, `onAtomURI()`.

hansonw:
+1 to @mbolin's suggestion. Maybe we could even simplify things a little and have `atom://package-name/path?params` trigger `package-name`'s URL handler service callback?

From searching on github, there's only three `package.json`s that contain a `urlMain` - one of them looks experimental and another is a homedir backup, so there's only one package (https://atom.io/packages/mastermind) that actually relies on this behaviour. (We could preserve the old urlMain behaviour if no path is specified).

No matter what we go with, this has tremendous value to internal Facebook users so I'd be willing to try and push this through myself if @BinaryMuse is busy :P

BinaryMuse:
I'm revisiting this and re-working it based on feedback. Curious about thoughts from folks who have commented so far and other @atom/core members. Here's the new docs:

---

Private: Associates listener functions with URLs from outside the application.

The global URL handler registry maps URLs to listener functions. URLs are mapped
based on the hostname of the URL; the format is atom://package/command?args.
The "core" package name is reserved for URLs handled by Atom core (it is not possible
to register a package with the name "core").

Because URL handling can be triggered from outside the application (e.g. from
the user's browser), package authors should take great care to ensure that malicious
activities cannot be performed by an attacker. A good rule to follow is that
**URL handlers should not take action on behalf of the user**. For example, clicking
a link to open a pane item that prompts the user to install a package is okay;
automatically installing the package right away is not.

Packages can register their desire to handle URLs via a special key in their
`package.json` called "urlHandler". The value of this key should be an object
that contains, at minimum, a key named "method". This is the name of the method
on your package object that Atom will call when it receives a URL your package
is responsible for handling. It will pass the full URL as the only argument, and you
are free to do your own URL parsing to handle it.

By default, Atom will defer activation of your package until a URL it needs to handle
is triggered. If you need your package to activate right away, you can add
`"deferActivation": false` to your "urlHandler" configuration object. When activation
is deferred, once Atom receives a request for a URL in your package's namespace, it will
activate your pacakge and then call `methodName` on it as before.

If your package specifies a deprecated `urlMain` property, you cannot register URL handlers
via the `urlHandler` key.

## Example

Here is a sample package that will be activated and have its `handleUrl` method called
when a URL beginning with `atom://my-package` is triggered:

`package.json`:

```javascript
{
  "name": "my-package",
  "main": "./lib/my-package.js",
  "urlHandler": {
    "method": "handleUrl"
  }
}
```

`lib/my-package.json`

```javascript
module.exports = {
  activate: function() {
    // code to activate your package
  }

  handleUrl(url) {
    // parse and handle the url
  }
}
```

hansonw:
This looks really good! At least I think this would solve our use case (on Nuclide) quite well.

I have a question around the case where the URL has to be opened in a new window, though. Since the URL message is emitted as soon as the `window:loaded` event is emitted, is it guaranteed that packages will be ready/subscribed to the URL registry at that point?

Thanks so much for picking this back up! The current Nuclide URL handler is a huge hack job.

BinaryMuse:
@hansonw I haven't wired things up yet, but yeah — the idea would be to defer the handling of the URL until all the packages' metadata have been loaded, appropriate packages activated, and handlers hooked up.

BinaryMuse:
With a test package called `my-package` with appropriate `package.json` config, this now works with `open -a /Applications/Atom.app atom://my-package/testurl --args -d` on macOS.

src/application-delegate.coffee
```diff
@@ -233,6 +233,14 @@ class ApplicationDelegate
     new Disposable ->
       ipcRenderer.removeListener('context-command', outerCallback)
 
+  onUrlMessage: (callback) ->
```
nathansobo: Stupid nitpick... we typically do initialisms like URL and HTML in all capital letters unless they're the first word of the identifier, so this should be `onURLMessage`. We do this to blend in with the majority of DOM APIs.

nathansobo:
So am I reading this right that a `urlMain` allows a package to specify its own window initialization code?

smashwilson:
> Because URL handling can be triggered from outside the application (e.g. from
the user's browser), package authors should take great care to ensure that malicious
activities cannot be performed by an attacker. A good rule to follow is that
URL handlers should not take action on behalf of the user. For example, clicking
a link to open a pane item that prompts the user to install a package is okay;
automatically installing the package right away is not.

Can we make it easier for package authors to get this right? Maybe we could build a prompt into the handler before it's handed off to the package's callback function, and show the URL, a package-provided prompt message, and ok/cancel buttons.

thomasjo:
> Can we make it easier for package authors to get this right? Maybe we could build a prompt into the handler before it's handed off to the package's callback function, and show the URL, a package-provided prompt message, and ok/cancel buttons.

I think this approach is _very_ sensible. Let's not assume that all developers always have the best intentions. Even if they are not knowingly doing something Evil™, they might be doing something "bad" as an effect of incompetence or gross negligence.

While the prompt might be somewhat annoying, I think it should be implemented. We can consider an explicit action to permanently disable it (globally, or on a per-package basis).

BinaryMuse:
> So am I reading this right that a urlMain allows a package to specify its own window initialization code?

Yes, that's correct. Only a few packages use it (Nuclide, mastermind, and a couple others). It should be *relatively* easy to convert them to the new system, but it's also quite easy to support a deprecated `urlMain` property as long as the package doesn't also attempt to opt-in to the new URL handlers.

> Maybe we could build a prompt into the handler

Yeah, definitely have some ideas for a nice prompt in Atom. This PR won't ship without addressing the necessary security considerations.

BinaryMuse:
Thinking more about the prompt thing; this has come up in internal conversations. Some thoughts from the past:

> Thinking about the UI concerns, I tend to veer toward not warning. I feel like, if we are going to warn at all, that should happen at package installation time (if ever). Any UI that requires the user to make a decision while they are working is going to go one of two ways:
> - Blindly click "OK" since I just want to get my work done.
> - Blindly click "No" since I value my security and have no real capacity to understand the implications of clicking "OK".
> 
> In either case the user isn't really able to make an educated decision. I mean, even if we surfaced UI and said "Hey, the extension 'Foo' is about to perform a 'bar'", what can they do? There is nothing that ties the invocation name to any specific functionality. On mobile devices this form of prompting is semi-useful, as they have defined specific classes of access: "Hey, this app wants to get your current location". If Atom had specific classes of access like that then it would make sense. But, prompting without being able to provide context doesn't seem useful. 
> 
> [...]
> 
> I don't think pushing the decision onto the user is compelling unless you can actually surface what they are authorizing. Given that isn't built-in to the security model of Atom currently, I'd tend to skip it.

smashwilson:
You're absolutely right that a mandatory prompt isn't useful for preventing protocol handlers from performing malicious actions; I was thinking that it would be useful for preventing the exploitation of protocol handlers that _don't_ prompt to perform actions without a user's knowledge.

Here's the scenario I'm thinking this will address:

1. A package X registers a URL handler that does not prompt and that does some kind of potentially exploitable action: writing to a file based on the URL contents, or installing a package, or opening a socket, ...
2. An attacker discovers the handler and embeds a crafted `atom://` URL on a page that Atom users are likely to visit (a package README maybe, spam link, HN clickbait?).
3. Atom users with package X installed unwittingly click the `atom://` URL.
4. Atom silently performs some action without notification. A malicious package is installed, or malware is dropped on their system. The user's machine is compromised.

In other words, I wasn't thinking of the prompt as a way to defend against malicious URL handlers in packages, I was thinking of it as a way to defend against turning URL handlers into an attack vector from the web.

> Any UI that requires the user to make a decision while they are working is going to go one of two ways:
>
> * Blindly click "OK" since I just want to get my work done.
> * Blindly click "No" since I value my security and have no real capacity to understand the implications of clicking "OK".
>
> In either case the user isn't really able to make an educated decision.

The decision we would want to present should be as simple as "did you just do something that you _expect_ Atom to respond to? If so, click OK. If not, click Cancel."

Crafting the wording to get that across without spooking anyone would be the trick :smile:

jgebhardt:
@smashwilson's scenario is compelling (and doesn't seem far-fetched, either). My guess is that exploiting existing packages to e.g. download malware would be a more likely attack vector than compelling Atom users to install a malicious package containing the handler. Users might not even see the actual URL passed to Atom (because of redirects, etc), so showing a prompt along the lines of
> "Hey, the extension 'Foo' is about to perform a 'bar'"

seems like an important first (and last?) line of defense. While I totally get the convenience argument, showing at least a one-time prompt with an option to "always allow Foo to perform Bar" seems like an acceptable compromise. Browsers already do this for eg. `atom://` links: 
screenshot 2017-09-20 12 41 36

...which unfortunately means that clicking an Atom link for the first time would result in a prompt by the browser, followed by another prompt from Atom.

BinaryMuse:
@ptoomey3 Wondering if you have any thoughts regarding ☝️

92161898515:
Saber? é tudo aquilo que ele não tem, sempre estou aprendendo, e ainda não
sei nada só estudei até a sétima série sozinho abri minhas redes sociais
nunca fiz curso ninguém me ensinou nada
você vem falar de saber!, estou em aprendizagem sozinho serto.🍍🚽
Edvaldo Silva dos santos

Em 21/09/2017 12:48 da manhã, "Michelle Tilley" <notifications@github.com>
escreveu:

ptoomey3:
> Wondering if you have any thoughts regarding 

My original ambivalence with warnings was the lack of useful context. I was under the impression that the most we would know is what extension a URI handler would invoke. So, I was thinking the best we could do is "Allow extension Foo?". Is there discussion of more granularity than that (i.e. we know the specific operation the extension will invoke)? If we know more context there could be some value in prompting. Since Atom has no intent like system (ala mobile apps), the context largely has to be left in the hands of the extension. 

> While I totally get the convenience argument, showing at least a one-time prompt with an option to "always allow Foo to perform Bar" seems like an acceptable compromise.

If the reason we are prompting is because an operation is potentially harmful, what does "always allow" accomplish? The first time they encounter it from a non-malicious site they check the "allow always" button. But, what happens when the same handler is invoked later by a malicious site?

> A package X registers a URL handler that does not prompt and that does some kind of potentially exploitable action: writing to a file based on the URL contents, or installing a package, or opening a socket

This goes to the "context" argument above. If an extension has a URI handler to install a package, yeah, that shouldn't happen automatically. I'd argue it is probably a bad idea for that to happen even semi-auto-magically...it gives me flashbacks back to the days of [ActiveX](https://en.wikipedia.org/wiki/ActiveX) :smile:. Even for less scary things, like writing a file, it still feels like you need more context than "Allow extension RepositoryImporter?". In that case, I would expect the URI handler to be responsible for triggering a more detailed flow from the extension itself that has the actual context. For example, it would surface UI such as what repository is about to be imported, where it is coming from, where it will be written to, etc. In other words, I do think that some form of UI interstitial can be handy for these (i.e. not having things blindly happen). But, I'm just not sure a generic interstitial has enough of the context to be a meaningful security boundary.

jgebhardt:
> what does "always allow" accomplish?

It would purely be an escape hatch for the user. They might have full trust in the extension, or the action might be clearly safe. In those cases, showing a warning every time a link is followed would be super annoying. It's not so much about achieving perfect security as it is about putting the user in control.

I think of this scenario as similar to OSX app install permissions– I believe there used to be a prompt to "always allow" when trying to install an app outside of the app store. The prompt has since been replaced by an option in system preferences, which makes it much more annoying to change.

Perfect security cannot be achieved in Electron almost by definition, due to the lack of sandboxing and the dynamism of JS. Putting the user in charge of risky actions is achievable though. Sure, people might blindly click "always allow", but at least they get to make a conscious choice. 

Regarding context / what to show in a potential prompt: Maybe there could be an (optional?) `description` field in the handler definition? So the prompt might look like

```
+------------------------------------------------------------------------------+
| Package <Package name> is trying to invoke URL handler <Method name>.        |
| <Handler Description, e.g. "Launches the debugger for your current project"> |
|                                                                              | 
| Do you want to proceed?                                                      |
|                                                                              | 
| ( ) Always allow <Package Name> to execute this handler                      |
| ( ) Always allow any package to run any URL handler                          |
|                                                          [ Cancel ][ Allow ] |
+------------------------------------------------------------------------------+
```

ptoomey3:
> Regarding context / what to show in a potential prompt: Maybe there could be an (optional?) description field in the handler definition? So the prompt might look like

Yeah..I had a private chat with @BinaryMuse about something like this. My thought was encouraging this by adding a secondary callback..one that basically renders the authorization dialog, but from the extension writer, with all the necessary context passed in from the URI data. In theory the extension writer could do all of that in their "main" callback. But, by separating it into an explicit argument at registration time it encourages correctness. And, for those extensions that truly need no authorization (ex. extension renders JPG images given a URL), they could provide a no-op authorization handler. As you noted, it isn't perfect, but it would make it harder to do the wrong thing. I'd tend to favor a full on callback instead of a basic description, since the actual arguments of the callback could be useful in rendering the message with context.

BinaryMuse:
Thought about this over the weekend and discussed with a few team members during stand-up this morning. Some thoughts.

> Regarding context / what to show in a potential prompt: Maybe there could be an (optional?) `description` field in the handler definition?

The main problem with this is a single URI handler method is responsible for handling *any* url that starts with `atom://package-name/`, so for packages with multiple possible endpoints (or endpoints that change from version to version), this wouldn't be effective.

> My thought was encouraging this by adding a secondary callback..one that basically renders the authorization dialog

This idea hasn't sat right with me since we discussed it, and I think I realized why — the proposal here is to basically have to callbacks, one for e.g. `handleUriPresentingConfirmation` and one for e.g. `handleUriPerformingAction`. In the example for installing a package, the former would do something like open the settings-view package page, and the latter would do the actual install.

But we actually don't want the package author to write `handleUriPerformingAction` *at all*, and the dual-callback API seems to encourage this behavior. We *only* want the user to create URI handlers that do the former (present a UI option, open a file, etc). I feel like having the two-method API will encourage package authors to create URI handlers that we don't actually want.

Furthermore, in most cases, the UI presented would already be wired up to an action. For example, for a "Clone in Atom" URI handler, we would show the clone dialog, and the clone dialog *already* has a button wired up to a handler that clones the repo; the second callback would not be useful. Same with installing a package; if the auth callback presents the settings-view page for the package, the second callback isn't useful since there is already a button on that page that installs the package and there's no good way to call any sort of "confirm" callback.

While discussing it with @nathansobo and others this morning, we agreed that the API that would most match the existing Atom APIs is a simple single-callback system with social engineering/encouragement/documentation around doing the right thing as a package author. In addition, we want to provide information on package installation pages that a package registers URI handlers (e.g. an icon and a text callout at the top of the package description), as well as a new tab in settings-view that helps manage URI handling in general (including being able to disable misbehaving packages).

ptoomey3:
> Furthermore, in most cases, the UI presented would already be wired up to an action

Yup..the dual callbacks system was merely a (likely bad) stab at trying, via the API, to encourage good behavior. In theory it might be easier to make an extension that does the safe thing (prompting in some way) if you are explicitly asked to pass some callback related to security. If not asked, it is on you to have read docs, have thought about it, etc. 

> we agreed that the API that would most match the existing Atom APIs is a simple single-callback system with social engineering/encouragement/documentation around doing the right thing as a package author.
> ..
> Finally, we will also make it clear in the documentation that having a misbehaving URI handler is grounds for removal from the Atom package registry.

Sounds sensible.

BinaryMuse:
Related settings-view PR: https://github.com/atom/settings-view/pull/1002

src/protocol-handler-installer.js
```diff
@@ -0,0 +1,49 @@
...
+  constructor () {
+    this.subscriptions = new CompositeDisposable()
+    this.supported = ['win32', 'darwin'].includes(process.platform)
+  }
+
+  initialize (config) {
+    this.config = config
+
+    this.subscriptions.add(
+      this.config.observe('core.defaultProtocolHandler', this.onValueChange.bind(this))
```
BinaryMuse: 🤔  I wonder if it's too obnoxious to immediately prompt the user if they switch to `"prompt"` and Atom isn't the default protocol handler — maybe we should only check at startup?

src/protocol-handler-installer.js
```diff
@@ -0,0 +1,101 @@
...
+class ProtocolHandlerInstaller {
+  constructor () {
+    this.subscriptions = new CompositeDisposable()
+  }
+
+  initialize (config, notifications) {
+    this.config = config
+    this.notifications = notifications
+
+    this.subscriptions.add(this.config.observe('core.uriHandlerRegistration', this.onValueChange.bind(this)))
```
BinaryMuse: 🤔  I wonder if it's too obnoxious to immediately prompt the user if they switch to `"prompt"` and Atom isn't the default protocol handler — maybe we should only check at startup?

nathansobo:
I haven't reviewed the code, but I like the API as proposed.

BinaryMuse:
The new settings-view panel from atom/settings-view#1002

settings_ ___atom

spec/url-handler-registry-spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    registry.handleUrl('atom://test-package/path')
+    expect(testPackageSpy).toHaveBeenCalledWith(url.parse('atom://test-package/path', true), 'atom://test-package/path')
+    expect(otherPackageSpy).not.toHaveBeenCalled()
+
+    registry.handleUrl('atom://other-package/path')
+    expect(otherPackageSpy).toHaveBeenCalledWith(url.parse('atom://other-package/path', true), 'atom://other-package/path')
+  })
+
+  it('keeps track of the most recent URIs', () => {
```
BinaryMuse: This is for a the new settings-view panel, which will show the most recent URLs.

BinaryMuse:
@atom/core I think this is ready for a review! A couple of open questions:

1. On macOS, it's not possible to stop being the default protocol handler if there are no other apps on the system to promote to be the new default. In this case, how can the user (if we even need to provide such a facility) tell Atom to not process URLs even if it's registered as a handler? (See the current settings-view UI above)
2. For URLs that should trigger opening a file (e.g. think links from a test window or debugger), it's likely that the file should be opened in a specific window (the window that has the file's project open) instead of the most recent window. How do you think we should best handle this? I was thinking probably a new core API, something like `atom.workspace.openInBestWindow(...)` or similar with a better name that packages could use to handle this. Alternatively, I thought about registering a special core URI, something like `atom://core/open?file=/path/to/file&line=n&col=n` that would do just do the right thing, but it's easy to imagine scenarios that this doesn't cover (e.g. selecting text in a specific file).

nathansobo:
> For URLs that should trigger opening a file (e.g. think links from a test window or debugger), it's likely that the file should be opened in a specific window (the window that has the file's project open) instead of the most recent window. How do you think we should best handle this?

An awesome solution seems like it would be allowing packages to optionally run URL handlers in the main process, but I don't know if we want to expand scope to having package code run there right now (or ever?).

Short of that a cross-window API like the one you describe seems sane.

BinaryMuse:
I'm pretty nervous about letting packages run code in the main process. My off-the-cuff opinion is that we should avoid that as long as possible. 😅 

I think we can handle the cross-window issue tangentially from this PR so I won't address it here.

benzen9113:
really expecting this release. Do you have cross-platform issues? hmm cross-window?

src/package.coffee
```diff
@@ -318,6 +320,21 @@ class Package
...
+        @handleUrl(methodName, args)
+
+  unregisterUrlHandler: ->
+    @urlHandlerSubscription?.dispose()
+
+  handleUrl: (methodName, args) ->
+    @activate().then =>
+      @mainModule[methodName]?.apply(@mainModule, args)
+    unless @mainActivated
+      @activateNow()
```
BinaryMuse: Not _100%_ sure this is the correct way to set things up, but it seems to work in all cases (including deferred loading).

spec/url-handler-registry-spec.js
```diff
@@ -0,0 +1,75 @@
+/** @babel */
+
+import url from 'url'
+
+import {it} from './async-spec-helpers'
+
+import UrlHandlerRegistry from '../src/url-handler-registry'
```
nathansobo: Some deep code review insight... To blend in with other names containing initialisms in the Atom codebase, URL should be capitalized.

spec/url-handler-registry-spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    registry.handleUrl('atom://another/url')
+    expect(changeSpy.callCount).toBe(6)
+    const history = registry.getRecentlyHandledUrls()
+    expect(history.length).toBe(5)
+    expect(history[0].url).toBe('atom://another/url')
+    expect(history[4].url).toBe(urls[1])
+  })
+
+  it('refuses to handle bad URLs', () => {
```
nathansobo: 👍🏻

src/atom-environment.coffee
```diff
@@ -1092,6 +1100,14 @@ class AtomEnvironment extends Model
   dispatchContextMenuCommand: (command, args...) ->
     @commands.dispatch(@contextMenu.activeElement, command, args)
 
+  dispatchUrlMessage: (uri) ->
+    if @packages.hasLoadedInitialPackages()
+      @urlHandlerRegistry.handleUrl(uri)
```
nathansobo: `s/Url/URL`

src/atom-environment.coffee
```diff
@@ -1092,6 +1100,14 @@ class AtomEnvironment extends Model
   dispatchContextMenuCommand: (command, args...) ->
     @commands.dispatch(@contextMenu.activeElement, command, args)
 
+  dispatchUrlMessage: (uri) ->
+    if @packages.hasLoadedInitialPackages()
```
nathansobo: Not saying you should do it as part of this PR, but it would be cool to have a `getInitialPackagesLoadPromise` method.

src/config-schema.js
```diff
@@ -55,6 +55,25 @@ const configSchema = {
           }
         }
       },
+      uriHandlerRegistration: {
```
nathansobo: More naming feedback... I'm seeing a mix of "URL" and "URI". I prefer URI myself because I think it's a more general concept.

src/main-process/atom-application.coffee
```diff
@@ -648,6 +648,41 @@ class AtomApplication
...
+    if @lastFocusedWindow?
+      @lastFocusedWindow.sendUrlMessage url
+    else
+      windowDimensions = @getDimensionsForNewWindow()
+      @lastFocusedWindow = new AtomWindow(this, @fileRecoveryService, {resourcePath, windowInitializationScript, devMode, safeMode, windowDimensions, env})
+      @lastFocusedWindow.on 'window:loaded', =>
+        @lastFocusedWindow.sendUrlMessage url
+
+  findPackageWithName: (packageName, devMode) ->
+    _.find @getPackageManager(devMode).getAvailablePackageMetadata(), ({name}) -> name is packageName
```
nathansobo: Don't JS arrays have a built-in `find` method now?

src/main-process/atom-application.coffee
```diff
@@ -648,6 +648,41 @@ class AtomApplication
...
+    parsedUrl = url.parse(urlToOpen)
+    return unless parsedUrl.protocol is "atom:"
+
+    pack = @findPackageWithName(parsedUrl.host, devMode)
+    if pack?.urlMain
+      @openPackageUrlMain(parsedUrl.host, pack.urlMain, urlToOpen, devMode, safeMode, env)
+    else
+      @openWithAtomUrl(urlToOpen, devMode, safeMode, env)
+
+  openWithAtomUrl: (url, devMode, safeMode, env) ->
```
nathansobo: I'm a bit confused about what makes this an "Atom" URL. I wonder if the method could indicate the distinction between the package URL main more clearly?

src/package.coffee
nathansobo: Would it be better to activate the package before handling the URL in case the handler depends on any state that gets established during activation?

nathansobo (dismissed):
I'm not the world's best asynchronous code reviewer, but this all looks reasonable to me. On the pedantic naming front, I'd like to consistently use the term "URI" over URL since there's precedent for using the name "URI" elsewhere in the API, and also we should always capitalize the full initialism in camel-cased words. The only major question I have is about the ordering of package activation and URL handlers. Seems like the package should always activate first. Maybe that's what it does already and I'm misunderstanding, or maybe there's a good reason not to do that I'm not thinking of.

Also, where are we planning on documenting the new package.json features? I want to be sure documentation gets updated in sync with this feature.

Thanks for adding this! It's going to be useful in a variety of places.

src/package.coffee
BinaryMuse: @nathansobo Unless I'm misreading the code, that's what's happening:

1. `activate()` creates the activation promise (a `Promise.all` of a few internal promises)
2. `activateNow()` resolves the activation promise
3. The `then` hook runs and calls the URL handler method

nathansobo: Oh of course, the `then` is deferred. Carry on.

nathansobo: Would it be possible to swap the statement order?

BinaryMuse:
@nathansobo I switched up the naming across the three relevant PRs (two of them in settings-view).

Documentation will be added to the flight manual as part of this effort (see the URI Handling project for more).

nathansobo (approved):
🚢 when ready!

iolsen:
Some damn fine work pushing this monster effort over the finish line, @BinaryMuse! 💯

akonwi:
This is great and I'm excited to use this. My biggest question right now is whether this means an `atom://cool-package` link will go to the package details in atom settings to allow the user to install it if it isn't already installed. From what I gathered skimming this thread, this api is more for already installed packages responding to URIs

BinaryMuse:
@akonwi `settings-view` sets up a URI handler that can show the installation/details page for a package; see https://github.com/atom/settings-view/pull/1011

akonwi:
thanks!

PR #11403, opened 2016-04-08
Fix dead atom.io/docs links in CONTRIBUTING.md

winstliu:
This file wasn't updated for the flight-manual conversion.

I also fixed two `https://atom.io/docs/latest` URLs, one of which was inadvertently going to the Notifications API when it was talking about core, and the other leading to a 404.

Would love some :eyeglasses: in case I missed anything!  /cc @atom/feedback


lee-dohm:
Looks good to me, thanks @50Wliu :grinning:

PR #11412, opened 2016-04-08
Signing support on Windows with P12 keys

damieng:
Allows signing of the Windows executables (atom.exe, update.exe, atomsetup.exe) using the standard Windows SDK signing tool. 

Included is a test signing pfx/p12 signing key with the password 'password'. You can specify your own using the WIN_P12KEY_PASSWORD and WIN_P12KEY_URL as the Atom team will do on official builds when signing with the official key.


nathansobo:
Merging this without testing on Janky because it's on a fork and I can't run a build for it. If it causes problems we can back it out.

PR #11420, opened 2016-04-10
Fixed positioning for overlay

wvanlint:
This PR changes the positioning of the overlay from absolute to fixed.
This ensures that no overflow will be triggered in surrounding elements, which is important especially for mini TextEditors.
cc @nathansobo


nathansobo:
Tested this out locally and everything seems to be working. @atom/core if anyone sees anything weird with overlay decorations, this may be a culprit.

PR #11436, opened 2016-04-12
Allow A Window's Environment To Reflect The Most Recent atom Invocation

joefitzgerald:
This PR splits out portions of #11337, to address:
- Allows multiple invocations of `FOO=blah atom .` and `FOO=bleh atom .` to result in an updated environment for an existing window ("Should `process.env` be updated if atom is invoked from the shell for the same directory multiple times?" from #4126)

/cc @lee-dohm


lee-dohm:
Thanks for bearing with me @joefitzgerald :bow:

PR #11459, opened 2016-04-14
Fix deprecations reporting

as-cii:
Previously builds that contained deprecations were failing without logging out the reason. I tracked down the issue to `ELECTRON_ENABLE_LOGGING`, which was preventing deprecations from being logged out. This is how this PR changes the `run-specs` task to handle this scenario correctly:
1. :bug: Use `ELECTRON_ENABLE_LOGGING=true`. This allows us to capture the output of calls to `console.warn` and `console.log`, which are useful to log out Grim deprecations.
2. :fire: Remove calls to `logDeprecations`. This method used to format the output captured in `stderr` to strip out "[Console]" noise from deprecations. This code path was not running anymore because we started using `stdio: 'inherit'` in #10838 for core specs, which prevents stderr output to be captured. I feel like it's okay to remove it, as the output is already readable as it is.

You can see a sample of a new build here: https://travis-ci.org/atom/atom/jobs/123049543#L254-L257.

### DIFF
--- a/build/tasks/spec-task.coffee
+++ b/build/tasks/spec-task.coffee
@@ -5,6 +5,13 @@ temp = require('temp').track()
 _ = require 'underscore-plus'
 async = require 'async'
 
+# TODO: This should really be parallel on every platform, however:
+# - On Windows, our fixtures step on each others toes.
+if process.platform is 'win32'
+  concurrency = 1
+else
+  concurrency = 2
+
 module.exports = (grunt) ->
   {isAtomPackage, spawn} = require('./task-helpers')(grunt)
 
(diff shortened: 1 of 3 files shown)


as-cii:
@atom/core: what do you think?

joshaber:
> Remove Windows/Linux run-specs code. There is little value in keeping those code paths, given that we don't run specs on those platforms; if we ever need it again, we can restore it from the git history.

I thought @damieng was actively working on getting the Windows specs running?

as-cii:
> I thought @damieng was actively working on getting the Windows specs running?

Thanks for the heads-up, @joshaber. I'll put that code back so that @damieng can refine it later when we need to run those specs on CI. 👍

PR #11463, opened 2016-04-14
Fixes #11190

### LINKED ISSUE #4126: Inconsistent environment and behaviour

### LINKED ISSUE #11190: Atom 1.7.0-beta0 fails to open from dock
### Description

Since Atom 1.7.0-beta0, Atom fails to open when initiated from the dock (Mac). It seems to work fine when opened from the terminal.

When called from the dock, a debugger window appear with the traceback

```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:8)
    at Object.posix.join (path.js:479)
    at Object.fsPlus.resolve (/Applications/Atom Beta.app/Contents/Resources/app.asar/node_modules/fs-plus/lib/fs-plus.js:461)
    at new Config (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/config.js:81)
    at new AtomEnvironment (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/atom-environment.js:181)
    at module.exports (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/initialize-application-window.js:23)
    at setupWindow (index.js:87)
    at window.onload (index.js:41)
```

### Steps to Reproduce

1. Open Atom from the dock icon.

### Versions

Atom    : 1.7.0-beta0
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1

OSX: 10.11.3

Installed packages:
├── Sublime-Style-Column-Selection@1.5.1
├── auto-detect-indentation@1.0.0
├── autocomplete-paths@1.0.2
├── autocomplete-python@1.7.0
├── file-icons@1.6.18
├── git-plus@5.13.0
├── highlight-selected@0.11.2
├── language-fish-shell@1.0.3
├── language-idl@0.4.0
├── language-latex@0.6.1
├── latex@0.32.0
├── linter@1.11.3
├── linter-flake8@1.13.0
├── merge-conflicts@1.3.7
├── minimap@4.21.0
├── minimap-find-and-replace@4.5.0
├── nucleus-dark-ui@0.7.14
├── open-terminal-here@2.1.4
├── project-manager@2.7.6
├── python-indent@0.4.2
├── set-syntax@0.3.0
└── synced-sidebar@0.4.3

alazaro:
This is the error thrown when starting Atom from Spotlight on Mac or by double clicking on the App and the default shell is `fish-shell`.
I attach a Snapshot of the error, that doesn't happen anymore after the fix.
screen shot 2016-04-14 at 16 19 48

### DIFF
--- a/src/environment-helpers.js
+++ b/src/environment-helpers.js
@@ -58,7 +58,7 @@ function getFromShell () {
 function needsPatching (options = { platform: process.platform, env: process.env }) {
   if (options.platform === 'darwin' && !options.env.PWD) {
     let shell = getUserShell()
-    if (shell.endsWith('csh') || shell.endsWith('tcsh')) {
+    if (shell.endsWith('csh') || shell.endsWith('tcsh') || shell.endsWith('fish')) {
       return false
     }
     return true


50Wliu:
Shouldn't this fix #11190 and not #4126?

alazaro:
@50Wliu You're right

lee-dohm:
Thanks for this!

artemruts:
After this pull request `Atom` is not reading paths added by `fish-shell` when starting from `Spotlight` or with double click. Correctly reads paths only when started directly from shell, e.g. with `atom` command.

Issue #5359, opened 2015-02-02
Smooth scrolling (Linux)


oktayacikalin:
I've seen some work in improving the smoothness of scrolling so far. But I'm wondering if it's actually possible to get real smooth scrolling or at least if it's planned for the near future?
Just reducing the scroll sensitivity to 1 line at a time is not a real solution...

benogle:
@nathansobo will be soon working on new rendering techniques that will hopefully speed up typing and scrolling performance, which should make things smoother.

theblacklion:
Awesome to hear that :)

Ehsanizadi:
When do you expect smooth scrolling to be added to atom text editor?

Tenrys:
For now, put this in your style sheet:
    atom-text-editor::shadow .editor--private .tile {
        transition: -webkit-transform .05s ease-out; // customize the duration.
    }
Courtesy of @Lixquid.

UltCombo:
Looks like @Tenrys's selector no longer works, so I've modified it:

```css
atom-text-editor::shadow .scroll-view > .lines > div > div:not(.cursor) {
  transition: -webkit-transform .1s linear;
}
```

This is still an ugly hack, and does not work very well when scrolling up.

Chrome has a smooth scroll flag under `chrome://flags/#enable-smooth-scrolling`, I wonder if this flag can be enabled in Electron? If so, Atom could add a setting to enable this flag.

ebkalderon:
Thanks for the information, @UltCombo. Any word on when smooth scrolling will work natively?

groovy354:
:+1:  I'd love to see that implemented, as well :)

thedaniel:
@zcbenz any comments on the Chrome smooth scroll flag and whether it's available / relevant in Electron?

halohalospecial:
Would this suffice as a temporary solution?

https://atom.io/packages/animated-page-scroll

ArniPL:
Is this still in the works? I'd love Atom to do smooth scrolling like Sublime does. That CSS hack just isn't cutting it. Thanks! :)

alvarolm:
interested as well, btw that css hack doesnt seems to be working on 1.7 beta 5

alvarolm:
https://github.com/atom/atom/issues/10276

kevinrambaud:
Bump, same thing on OS X El Captain, it's the only feature I miss from Sublime text, the current Atom scroll attacks my eyes 😁

[...]

PR #11474, opened 2016-04-14
Update Electron to 0.37

### LINKED ISSUE #6320: Atom freezes when clicking web links in settings tab without the web browser already open
Atom 0.190.0
Linux.

This happens both in normal, safe and --one mode; it seems to happen only when the browser (I use firefox) is not open yet.

Steps to reproduce: packages > [a package] > settings > view on atom; or release notes > view on atom. If the browser was already open, everything works fine; otherwise, the browser opens but Atom freezes and must be killed from task manager.

winstliu:
Naïve bump to Electron 37 to see what broke and what got fixed.

~~Blocked on #11348~~
### TODO:
- [x] ~~Fix background throttling~~ (`text-editor-component-spec.coffee` fails even if we disable throttling, which means those specs need to be revamped because they're too time sensitive)
- [x] Fix autocomplete-plus specs (bb46933)
- [x] Fix find-and-replace specs (bb46933)
- [x] Fix tree-view specs (atom/tree-view#649)
- [x] Fix subpixel AA on non-editor elements
- [x] Test for regressions

Closes #6320
Unblocks atom/atom-keymap#35
If anyone knows of any other issues that are resolved with the upgrade to 0.37 please let me know!

/cc @lee-dohm


50Wliu:
Three failing specs, one of which is already flaky.
The other two reference invalid `atom.asar` archives, which is suspiciously similar to the error reported in https://github.com/atom/atom/pull/11348#issuecomment-209614024.

50Wliu:
@nathansobo I noticed you also bumped some timeouts in a35e24658d2466c08f185e330badc24b90294dcd and 21f8ad69583b659b57b87976afcac9b42cbec9cf - do those need to be reverted as well (and do you recall any other parts of the code with extended timeouts)?

50Wliu:
Crash while running specs in https://travis-ci.org/atom/atom/jobs/123212713#L246.

nathansobo:
@50Wliu Yes, if we've fixed throttling we should revert to the normal timeouts.

Nikki1993:
@50Wliu I assume we need to include commits from 0.36.12 PR as well with this PR?

50Wliu:
@Nikki1993: Yup, that's why it has the `blocked` label right now.  After 0.36.12 is on master I'll rebase from master.

Nikki1993:
@50Wliu Cool, I went ahead and created a patch for both PR, so that I can easily apply it each time I sync atom from master. 

I thing I noticed immediately is much faster cold startup time and general increase in responsiveness. This entire time I worked with it, I had 0 issues or new regressions introduced (apart from already reported on github).

Nikki1993:
Electron 37.7 has been released 2 days ago btw, https://github.com/electron/electron/releases/tag/v0.37.7

Nikki1993:
Electron 37.8 released https://github.com/electron/electron/releases/tag/v0.37.8

libeclipse:
I've been running a manual update for weeks. No problems. Arch linux.

50Wliu:
Looks like specs are _still_ being throttled even with the `backgroundThrottling` option :/.

50Wliu:
In addition I don't have any subpixel AA on the tree-view, status-bar, tabs, etc. anymore.  Only the editor itself has subpixel AA (might have to do with the removal of the `direct-write` option, though it's supposed to be enabled by default now...).  The menu options are colored a weird blue when selecting them.  Does seem to load considerably faster though as @Nikki1993 said.  Syntax highlighting is near-instantaneous.

50Wliu:
Tree View specs seem to be failing because of atom/tree-view#464 (it looks like there's a period in the temporary directory now) and should be fixed by atom/tree-view#649.

Looking into the ac+ and fnr spec failures.

50Wliu:
The ac+ and fnr spec failures were due to `@mini` being `undefined` and not `false` when not passed to the TextEditor constructor.  Then in `text-editor-component.coffee` there was a `.toggle('mini', @mini)` (pseudocode) call which, because `@mini` was undefined, ended up toggling the mini attribute an odd amount of times and finally making the editor a mini one.  Not too sure why this worked before, but it's still probably better to explicitly declare `@mini` as `false` so that methods like `isMini` don't return undefined.

Regarding subpixel AA...I started Atom today to discover that the tree-view and status-bar now have subpixel AA again.  Only the tabs and menu bar don't (though I don't remember the menu bar ever having subpixel AA).  Will update with more info as I see if last night was just a fluke.

I have no idea how to fix the background throttling, since it looks like we're setting it correctly.

Nikki1993:
@50Wliu speaking of performance improvements. Not sure if this related to Atom editor, but there is a considerable lag when you opening packages tab in settings menu.

I have 16 community packages and while they are loading everything is fine, however the moment atom starts constructing list view of those packages it freezes for solid 5 seconds as if it struggles/tries to load everything at once causing editor to hang until all packages in the list are loaded (text, buttons, images etc). I can shoot a short video if necessary to demonstrate. Present everywhere, stable, beta and latest dev with all the changes from this PR.

as-cii:
> have no idea how to fix the background throttling, since it looks like we're setting it correctly.

Does putting the timeouts back make the tests pass? Maybe disabling throttling isn't enough to run time-sensitive specs like the ones we have in `text-editor-component.spec` on a slow platform like Travis (although hopefully tests fail because of some other issue).

thorade:
Electron 1.0 has been released:
http://electron.atom.io/blog/2016/05/11/electron-1-0

Nikki1993:
@thorade 1.0.1 already :D https://github.com/electron/electron/releases/tag/v1.0.1

thorade:
@Nikki1993 impossible to keep up with that... But the 1.x promise is to keep the API stable.

Nikki1993:
@thorade yep. Unfortunately, this PR, most likely, will have to stop at electron 0.37.8 since 1.0.1 removes deprecated, but not flagged as such, APIs. Better upgrade to electron 0.37.8, see if there are any used deprecated APIs, refactor the code and create a separate 1.x PR. Or so that's what @50Wliu would do I gather :D

50Wliu:
Exactly :).  Need to give package authors time to update all their code as well!

damieng:
I think some of the crashing bugs we're getting in 1.7.0-3 are related to Electron 0.36.8

Where are we with this PR?

50Wliu:
I've been using it for about a week and the only regression I've found so far is lack of subpixel anti-aliasing (but I have no idea how to fix that).  The other problem is background throttling but I've reverted the timeout decreases so those specs are passing again.

as-cii:
🎉 🎉 🎉 

Thanks for your work on this, @50Wliu!

v3ss0n:
When we are going to  upgrade directly to electron 1.0?

damieng:
@v3ss0n https://github.com/atom/design-decisions/blob/master/electron-update-procedure.md

v3ss0n:
@damieng  thanks !

pratyushbarik:
Resolves #5359

s0ber:
@50Wliu Chromium version, used in this update, is very slow with rendering huge SVG documents (with lots of text nodes). I am right now creating atom package, which allows to draw different diagrams. And now it is super slow. It is really frustrating, because latest Chrome version has fixed this issue.

So, previously everything was fast in Atom and slow in Chrome. And now it is vice versa. =(

Issue #11471, opened 2016-04-14
Electron 0.36 broke stdout on Windows


Arcanemagus:
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)? (I think...)
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

It seems that the version of Electron bundled in Atom fails to report any data back on `stdout` when called as Node. `options.env.ELETRON_RUN_AS_NODE` is set (to `1`), but the `stdout` callbacks are never called. If you have Node installed on the machine and replace `process.execPath` with the path to Node everything works as expected (tested with Node v5.9.0).

This has been reported as an issue by users updating to v1.7.0 on the following linter plugins:

* `linter-jshint`
* `linter-htmlhint`
* `linter-csslint`
* Possibly `linter-bootlint` as well (Looks like this might just be user error)

Note that none of those linters have AppVeyor (Windows) CI testing configured.

The following Node based linters _are_ being tested on AppVeyor, but it turns out all of these are directly being `require`'d, and as such aren't being ran as an "external" Node process:

* linter-stylelint
* linter-js-yaml

`linter-eslint` is interesting in that it _is_ working, but it utilizes a worker process that is being created with [`child_process.fork()`](https://github.com/AtomLinter/linter-eslint/blob/v7.2.0/src/helpers.js#L15).

[...]

PR #11475, opened 2016-04-14
Add ELECTRON_NO_ATTACH_CONSOLE in BufferedNodeProcess

BinaryMuse:
Potential fix for #11471

### DIFF
--- a/src/buffered-node-process.coffee
+++ b/src/buffered-node-process.coffee
@@ -47,6 +47,7 @@ class BufferedNodeProcess extends BufferedProcess
     options ?= {}
     options.env ?= Object.create(process.env)
     options.env['ELECTRON_RUN_AS_NODE'] = 1
+    options.env['ELECTRON_NO_ATTACH_CONSOLE'] = 1
 
     args = args?.slice() ? []
     args.unshift(command)


Arcanemagus:
This fixes the test case for me in v1.8.0-beta0 :tada:

Arcanemagus:
Fixed it in `linter-jshint` as well.

Altiano:
Is this fix the issue of developing electron app on Atom (need to add `set ATOM_SHELL_INTERNAL_RUN_AS_NODE=" ` on terminal-plus plugin before run `electron .`)

lee-dohm:
@Altiano the environment variable `ATOM_SHELL_INTERNAL_RUN_AS_NODE` was renamed to `ELECTRON_RUN_AS_NODE` some time ago. I don't know about running Electron apps from inside Atom using terminal-plus. This isn't intended as a fix for that since it isn't a bug that I am aware of. If there is an open Issue somewhere for it, please post a link so that I can read up on it :grinning:

Altiano:
I don't know if there is an issue in here but there is on termianl-plus: https://github.com/jeremyramin/terminal-plus/issues/96

PR #11478, opened 2016-04-15
Prompt user to save as if save fails

### LINKED ISSUE #7708: Unsaved changes lost if permission denied when saving on exit

kuychaco:
Thanks @einarmagnus for adding this feature :sparkles:! I resolved the merge conflicts and updated code to use the current APIs.

Closes https://github.com/atom/atom/pull/7713
Closes https://github.com/atom/atom/issues/7708

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -454,13 +454,26 @@ class Pane extends Model
     else
       return true
 
+    saveError = (error) =>
+      if error
+        chosen = atom.confirm
+          message: @getSaveErrorMessage(error)
+          detailedMessage: "Your changes will be lost if you close this item without saving."
+          buttons: ["Save as", "Cancel", "Don't save"]
+        switch chosen
+          when 0 then @saveItemAs item, saveError
+          when 1 then false
+          when 2 then true
+      else
+        true
+
     chosen = atom.confirm
       message: "'#{item.getTitle?() ? uri}' has changes, do you want to save them?"
       detailedMessage: "Your changes will be lost if you close this item without saving."
       buttons: ["Save", "Cancel", "Don't Save"]
 
     switch chosen
-      when 0 then @saveItem(item, -> true)
+      when 0 then @saveItem(item, saveError)
       when 1 then false
       when 2 then true
 
@@ -479,8 +492,10 @@ class Pane extends Model
   # Public: Save the given item.
   #
   # * `item` The item to save.
-  # * `nextAction` (optional) {Function} which will be called after the item is
-  #   successfully saved.
+  # * `nextAction` (optional) {Function} which will be called with no argument
+  #   after the item is successfully saved, or with the error if it failed.
+  #   The return value will be that of `nextAction` or `undefined` if it was not
+  #   provided
   saveItem: (item, nextAction) ->
     if typeof item?.getURI is 'function'
       itemURI = item.getURI()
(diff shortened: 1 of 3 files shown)


src/pane.coffee
```diff
@@ -577,13 +577,26 @@ class Pane extends Model
     else
       return true
 
+    saveError = (error) =>
+      if error
+        chosen = @applicationDelegate.confirm
```
lee-dohm: I'd like to see this unified with the nearly duplicate code starting on line 593.

src/pane.coffee
```diff
@@ -577,13 +577,26 @@ class Pane extends Model
...
+          message: @getMessageForErrorCode(error)
```
lee-dohm: The other version starting on line 593 definitely identifies the file that has changes and will be saved. This message doesn't seem to do so. Can that be added?

amytruong:
**SCENARIOS TESTED**
 **Mac 10.10**
Close out edited read-only file - **PASS**
1. Make file read-only
2. Edit file
3. Close out file 
4. Save dialog box displays. Click Save
5. Atom Helper wants to make changes box displays.
6. Click Cancel (instead of putting in password)
**Expected Results:** Permission denied message displayed with **Save As** dialog box displaying
**Actual Results:** Permission denied message displayed with dialog box displaying. 

The save-as dialog box doesn't display, if I click Cancel after entering in my password, just the error message if you click Save/CMD-S - **FAIL** _(Not sure if this scenario falls in the scope of the fix but tried it out anyway just in case)_
1. Make file read-only
2. Make changes in Atom
3. Click Save/cmd-S
4. Atom Helper wants to make changes box displays
5. Click Cancel (instead of putting in password)
**Expected Results:** Permission denied message displayed and  **Save As** dialog box also displayed
**Actual Results:** Permission denied message displayed **_without_** Save As dialog box 
The message is only displayed. :cry:
screen shot 2016-04-18 at 12 08 45 pm

 **Windows 8.1**
Close out edited read-only file **PASS**
1. Make file read-only
2. Edit file
3. Close out file 
4. Save dialog box displays. Click Save
**Expected Results:** Permission denied message displayed with **Save As** dialog box displaying
**Actual Results:** Permission denied message displayed with **Save As** dialog box displaying

The save-as dialog box doesn't display if I click Cancel after entering in my password, just the error message if you click Save/CMD-S  _(Not sure if this scenario falls in the scope of the fix but tried it out anyway just in case)_
1. Make file read-only
2. Make changes in Atom
3. Click Save/cmd-S
4. Atom Helper wants to make changes box displays.
5. Click Cancel (instead of putting in password)
**Expected Results:** Permission denied message displayed with **Save As** dialog box displaying
**Actual Results:** Permission denied message displayed **_without_** Save As dialog box 
The message is only displayed. :cry:
screen shot 2016-04-18 at 12 50 48 pm

lee-dohm:
I think this is a definite improvement over the current behavior and I'm not sure that I was able to find the same issues that @amytruong did. I'm going to merge this for now so we don't hold up the improvement on the edge case. Especially because the edge case doesn't enable people to lose data like the old behavior (the tab stays open in all cases from what I can see).

PR #11488, opened 2016-04-15
First batch of Windows specs fixes

damieng:
This sorts out a number of specs issues but is not yet complete. Want to merge anyway as it's better than what we have right now.

### DIFF
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -172,8 +172,8 @@ jasmine.useRealClock = ->
 addCustomMatchers = (spec) ->
   spec.addMatchers
     toBeInstanceOf: (expected) ->
-      notText = if @isNot then " not" else ""
-      this.message = => "Expected #{jasmine.pp(@actual)} to#{notText} be instance of #{expected.name} class"
+      beOrNotBe = if @isNot then "not be" else "be"
+      this.message = => "Expected #{jasmine.pp(@actual)} to #{beOrNotBe} instance of #{expected.name} class"
       @actual instanceof expected
 
     toHaveLength: (expected) ->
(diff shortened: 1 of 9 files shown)


spec/text-editor-component-spec.js
```diff
@@ -4028,9 +4028,9 @@ describe('TextEditorComponent', function () {
       component.setFontSize(10)
       await nextViewUpdatePromise()
       expect(editor.getDefaultCharWidth()).toBeCloseTo(6, 0)
-      expect(editor.getKoreanCharWidth()).toBeCloseTo(9, 0)
+      expect(editor.getKoreanCharWidth()).toBeCloseTo(10, 0)
       expect(editor.getDoubleWidthCharWidth()).toBe(10)
-      expect(editor.getHalfWidthCharWidth()).toBe(5)
+      expect(editor.getHalfWidthCharWidth()).toBeCloseTo(6, 5)
```
as-cii: Do these specs pass on Mac? If not, I think we might make this spec more general to test for differences before/after setting the font size, e.g.

```js
let originalDefaultCharWidth = editor.getDefaultCharWidth()
component.setFontSize(10)
expect(editor.getDefaultCharWidth()).toBeLessThan(originalDefaultCharWidth)
```

spec/spec-helper.coffee
```diff
@@ -172,41 +172,47 @@ jasmine.useRealClock = ->
...
     toShow: ->
-      notText = if @isNot then " not" else ""
+      toOrNotTo = this.isNot and "not to" or "to"
       element = @actual
       element = element.get(0) if element.jquery
-      @message = -> return "Expected element '#{element}' or its descendants#{notText} to show."
+      @message = -> return "Expected element '#{element}' or its descendants #{toOrNotTo} show."
       element.style.display in ['block', 'inline-block', 'static', 'fixed']
 
+    toBePath: (expected) ->
```
as-cii: What do you think about renaming this to `toEqualPath`? Usually, we use the `be` matchers for primitive types and references, as opposed to `equal` where we mean that two things are logically equal.

as-cii:
Nice work here, @damieng. Left a couple of comments, but these changes look good so long as they make the Mac build 💚.

Out of curiosity, did you get a sense about what are the other "groups" of specs that still fail on Windows (e.g. UI, interaction with the browser process, environment, etc.)?

spec/spec-helper.coffee
damieng: The calls were previously toBe so I made this toBePath but I'm open to a better name. toSamePath perhaps to indicate there is some level of fuzzyness given the normalization on both sides.

damieng:
A large number of the failures are waiting to receive messages that are sent over the IPC communication mechanism.  Once that's fixed I think we're going to be most of the way there.  There is however still at least one spec that causes Atom to quit although I haven't found out which.

spec/spec-helper.coffee
as-cii: > toSamePath perhaps to indicate there is some level of fuzzyness given the normalization on both sides.

`toSamePath` sounds somewhat weird to me, as `expect(...).toSamePath(...)` doesn't read like a valid sentence. It seems like `toEqualPath` might be a better name:

1. It indicates _value equality_.
2. We use the `toEqual` variant elsewhere in the codebase. 
3. It reads nicely when writing assertions.

Ideally I think another solution would be to not use strings to represent paths, and just have a different object (e.g. `Path`) that encapsulates this behavior, so that we can just use `toEqual`. That seems like a high-impact breaking change, though, and therefore it seems like introducing a new matcher for this use-case could be the most pragmatic path.

spec/text-editor-component-spec.js
as-cii: I took a 👀 at https://travis-ci.org/atom/atom/jobs/123412388#L291-L303, and it seems like these specs are failing. What do you think about changing them to look similar to the solution proposed above?

as-cii:
> A large number of the failures are waiting to receive messages that are sent over the IPC communication mechanism. Once that's fixed I think we're going to be most of the way there. There is however still at least one spec that causes Atom to quit although I haven't found out which.

Sounds good. I think we can go ahead merging this after addressing the concerns above. 👍

Thanks, @damieng!

PR #11506, opened 2016-04-18
Always favor more recently added inline listeners

wvanlint:
This allows newer listeners to be prioritized based on when they are added.
This is useful as more recent contexts often have to be resolved first.

For example, when `core:move-down` is called in a `find-and-replace` editor, it will go through its history.
But when an autocompletion listener is added later, it should move through the possible autocompletions.

I am curious how this use case is handled in for plain DOM listeners, as they will always be called [in the order of their registration](https://www.w3.org/TR/uievents/#event-flow). It seems possible that listeners can be backed up, and restored after the context is resolved. I'm not very experienced in front-end development though, do you guys have insight?

cc @nathansobo

### DIFF
--- a/spec/command-registry-spec.coffee
+++ b/spec/command-registry-spec.coffee
@@ -74,6 +74,13 @@ describe "CommandRegistry", ->
       grandchild.dispatchEvent(new CustomEvent('command', bubbles: true))
       expect(calls).toEqual ['.foo.bar', '.bar', '.foo']
 
+    it "orders inline listeners by reverse registration order", ->
+      calls = []
+      registry.add child, 'command', -> calls.push('child1')
+      registry.add child, 'command', -> calls.push('child2')
+      child.dispatchEvent(new CustomEvent('command', bubbles: true))
+      expect(calls).toEqual ['child2', 'child1']
+
     it "stops bubbling through ancestors when .stopPropagation() is called on the event", ->
       calls = []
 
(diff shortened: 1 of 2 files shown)


thomasjo:
At first glance this only seems like a partial fix to the scenario you listed; what if there are multiple registrations of `first` or `last`? Aren't you in almost the same predicament? I'll have to mull this one over for a bit, but I'm not sure this really fixes the "problem".

Let's get some more feedback and 👀 on this. /cc @atom/feedback

lee-dohm:
Is this for adding autocomplete to the find field in the find-and-replace package?

wvanlint:
Time of registration is still the main ordering for listeners, which I think is correct,.
It is what is used for DOM events (see the link above).
If there are multiple registrations, developers should register as late or as early as possible in the code, based on what they want.

However, I believe it would be useful for the listeners that are registered later to be called first, which is where `first` comes in. The argument `last` replicates the existing behavior.

@lee-dohm, that is correct. I do think it is a bit more general though: it applies when two features want to use the same command on the same element.

nathansobo:
@wvanlint I'm inclined to just *always* favor more recently added inline listeners, which would be consistent with always favoring more recently-added selector-based listeners. I agree there's a bit of drift from native event handling, but since this is our own custom command system we don't have to be totally hostage to that. What do you think?

wvanlint:
@as-cii I've talked to Nathan before he left on paternity leave and it seems like it would indeed be cleaner to always favor more recently added listeners, similar to what is done now for selector-based listeners.
Nathan also looked through all the packages for affected calls to `atom.commands`, and I verified the ones that may have been affected.

Let me know what you think.

as-cii:
> Let me know what you think.

Looks good to me. Plus, there's still some time left before publishing the next beta release, which should let us catch potential unexpected behaviors caused by this change.

Thanks again @wvanlint for all your great work! ✨

PR #11537, opened 2016-04-20
Add Project::createFile and ::createDirectory

BinaryMuse:
This PR implements four new methods on `Project` to create files, create directories, and watch for the creation of both:
- `Project::createFile(path)`
- `Project::createDirectory(path)`
- `Project::onDidCreateFile(callback)`
- `Project::onDidCreateDirectory(callback)`

Currently packages (like `tree-view`) manually call out to the filesystem APIs to create files, and other packages aren't notified of such changes. This leads to bugs like https://github.com/atom/fuzzy-finder/issues/175. By centralizing file and directory creation and adding notification hooks, packages can be notified any time a `Project`'s file structure changes.

/cc @nathansobo @atom/core

### DIFF
--- a/spec/project-spec.coffee
+++ b/spec/project-spec.coffee
@@ -478,6 +478,52 @@ describe "Project", ->
       atom.project.removePath(ftpURI)
       expect(atom.project.getPaths()).toEqual []
 
+  describe ".createFile(path)", ->
+    base = path.join(__dirname, 'fixtures', 'dir')
+    filePath = path.join(base, 'newfile.txt')
+
+    beforeEach ->
+      fs.removeSync(filePath) if fs.existsSync(filePath)
+
+    afterEach ->
+      fs.removeSync(filePath) if fs.existsSync(filePath)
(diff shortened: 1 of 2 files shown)


nathansobo:
👍🏻

lee-dohm:
Should we also standardize deletion in the same way?

BinaryMuse:
Good call, @lee-dohm; the [same `fuzzy-finder` bug](https://github.com/atom/fuzzy-finder/issues/175) manifests itself but in reverse for deletions.

BinaryMuse:
@nathansobo and I talked about this at length and have decided to back out of it; we don't think duplicating all of the `fs-plus` methods in `Project` is the right way to solve this problem. Instead, we'd like to develop a signaling mechanism where packages that modify the filesystem can notify Atom that they did so, and Atom can let any arbitrary subscribers know in turn. In the future, we could make the notification process automatic by watching the filesystem for changes ourselves.

PR #11414, opened 2016-04-09
Use display layers facility of text-buffer; delete all the code they replace

nathansobo:
This PR replaces a bunch of old and twisty code in `TextEditor` with usage of a new facility in the `text-buffer` library introduced in https://github.com/atom/text-buffer/pull/149. You can read that pull request for more information about what replaces all the code we're deleting here.

PR #11539, opened 2016-04-20
🏇 Improve perf for clicking below buffer of long line to seek cursor

### LINKED ISSUE #10769: Clicking on empty space below long line freezes Atom
I'm aware of #979, but I'm not entirely sure this is the same issue, and if it is, this may be helpful info that I can't add to the issue b/c it's locked.

To reproduce, open a large one line file. This example creates a file with 150k `#`'s and opens Atom in safe mode. (And make sure soft wrap is disabled?)
```sh
dd if=/dev/zero ibs=1k count=150 | tr "\000" "#" > hashes.txt
atom --safe hashes.txt
```

Then click on the empty non-numbered space below the first line. This causes Atom to move the cursor to the end of the long first line. Unfortunately, Atom freezes for about ~30s on my MBP before the cursor move is complete.

Clicking anywhere *on* the line doesn't cause a freeze, nor does navigating with the keyboard. Even pressing the end key moves the cursor to the end of the line instantly.

Here's a profile of that frozen period:

hashes txt users rahat downloads 2016-02-10 22-26-27

I figure this is different than #979 because the frozen period involves figuring out cursor position and not syntax highlighting, as the profiles in that issue imply.

Confirmed on both Windows 10 and El Capitan on Atom v1.5.1

rahatarmanahmed:
In cases where the last line of a file is very long and the user clicks below that line to seek to the end of the last line, it can take a very long time due to calculating the position of every single column. See #10769 for a test case.

This commit checks the special case of clicking below the buffer and returns the position without going into the whole calculation loop. I'm not super familiar with how the internals of Atom work, and I assumed going through all this calculation to find the very last cursor position was needed due to surrogate pairs or something. But it seems that's not the case as the last column position can be found by just summing the row's textNode lengths, so short circuiting that logic brings a perf win.

This stops freezing for clicking below long lines, but doesn't solve the whole perf problem with seeking long lines. If you were to scroll to the end of a long line and click somewhere close to the end, it will still hang. Perhaps you could calculate the position backwards if the click position is in the latter half of the long line? Either way, that's a discussion for another issue/PR.

As far as tests go, I'm not really sure how you run the line-yardstick specs? Running `script/test` had a lot of failures on master for other specs. I don't think any new specs for line-yardstick is required since it already covers testing clicking below the buffer.

Resolves #10769

### DIFF
--- a/src/lines-yardstick.coffee
+++ b/src/lines-yardstick.coffee
@@ -36,6 +36,10 @@ class LinesYardstick
     previousColumn = 0
     previousLeft = 0
 
+    if targetLeft is Infinity
+      column = textNodes.reduce(((lineLength, nodeLength) -> lineLength + nodeLength.textContent.length), 0)
+      return Point(row, column)
+
     @tokenIterator.reset(line, false)
     while @tokenIterator.next()
       text = @tokenIterator.getText()


lee-dohm:
@as-cii can you take a look at this?

as-cii:
Thanks for contributing, @rahatarmanahmed!

I wonder if, instead of optimizing only the `Infinity` case, we could improve the general approach we have to calculate the screen position for a certain pixel position. In particular, I am thinking that we could optimize iterating over the text nodes by skipping those ones whose width is less than the requested pixel position. This would exploit the fact that, usually, lines are made of several text nodes and so we can perform just one call to `getBoundingClientRect` for each text node, rather than for each text node's character.

Another (pretty big) improvement could also be to binary search the text nodes. The algorithm that mixes both approaches would work like the following:

1. Is the text node's width smaller than the requested pixel position?
  1. No. Advance to next text node. Go to 2).
  2. Yes. Go to 3).
2. Advance  `TokenIterator` till it matches the current text node's start position, incrementing the `column` appropriately. Go to 1).
3. Binary search within the text node's characters positions, calling `@leftPixelPositionForCharInTextNode` to determine the character index corresponding to the pixel position.
4. Advance `TokenIterator` till it matches the found character index, incrementing the `column` appropriately.

We can even skip using the iterator, as we're doing in #11414, but I think it's fine to keep it for now as far as this improvement goes.

What do you think? Would you be interested in trying this out?

Thanks!

/cc: @nathansobo

rahatarmanahmed:
@as-cii So I'm a bit confused, now. The `lines-yardstick.coffee` #11414 seems to already do what you're suggesting, minus the binary search within a text node. It also pretty much does what this PR is supposed to do, so I guess it's kinda obsolete.

I can certainly attempt to add binary search on top of that PR, if @nathansobo is okay with that. I might need some short advice on how text nodes work, though (like when are there more than one?)

as-cii:
> So I'm a bit confused, now. The lines-yardstick.coffee #11414 seems to already do what you're suggesting, minus the binary search within a text node. It also pretty much does what this PR is supposed to do, so I guess it's kinda obsolete.

Good catch, I didn't realize that branch was already using the optimization technique suggested in the above comment. So, I think it might be really nice to actually add binary search not only within a single text node, but also to search among *all text nodes*: I have updated the algorithm above to reflect this further optimization, which means we will always call `getBoundingClientRect` only a `O(log N)` number of times. ⚡ 

> I can certainly attempt to add binary search on top of that PR.

If #11414 is still open when you create a pull-request, you can target `ns-switch-to-display-layers` as the head branch, and I'll happily merge it. Otherwise, you can also wait for the display layers pull-request to be merged and target `master` too. 👍

> I might need some short advice on how text nodes work, though (like when are there more than one?)

There might be an arbitrary number of text nodes, and they all get stored in an array sequentially: that is, the first text node is spatially the leftmost, and the last is spatially the rightmost. You can get the bounding client rect of each text node using the [Range APIs](https://developer.mozilla.org/en-US/docs/Web/API/Range).

I am going to close this for now: please, feel free to open a new pull-request either against `master` or `ns-switch-to-display-layers`, as suggested above. Thanks, @rahatarmanahmed!

rahatarmanahmed:
@as-cii I'm not sure it's possible to find the text node with a binary search. We need to know the column position of the node, which afaik requires we do a linear scan to sum up the lengths of all previous text nodes.

I'll go ahead and try adding the binary search within a node. Thanks for the help!

as-cii:
> I'm not sure it's possible to find the text node with a binary search. We need to know the column position of the node, which afaik requires we do a linear scan to sum up the lengths of all previous text nodes.

That's true, but the same reasoning applies to searching within a single text node. However, we would still reduce the number of calls to `getBoundingClientRect` (the expensive part) to `O(log number_of_text_nodes + log longest_text_node_length)` and compute the cheap part (i.e. line lengths) _linearly_.

> I'll go ahead and try adding the binary search within a node. Thanks for the help!

Thanks for contributing! 👍

PR #11550, opened 2016-04-20
build: [ci skip] add lint npm script

BrainCrumbz:
Aim: to be able to invoke only linting from command line, as opposed to the whole (long) build process. With this, one can call `npm run lint` from project root directory.

### DIFF
--- a/build/package.json
+++ b/build/package.json
@@ -45,5 +45,8 @@
     "unzip": "~0.1.9",
     "vm-compatibility-layer": "~0.1.0",
     "webdriverio": "^2.4.5"
+  },
+  "scripts": {
+    "lint": "grunt --gruntfile Gruntfile.coffee lint"
   }
 }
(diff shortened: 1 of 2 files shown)


BinaryMuse:
Although this doesn't quite match the other scripts in the repo, it also does less and I'm perfectly okay with implementing it like this.

If we decide we'd rather have the project bootstrapped first, we could have a `scripts/lint` with something like

```javascript
#!/usr/bin/env node
var safeExec = require('./utils/child-process-wrapper.js').safeExec;
var runGrunt = require('./utils/run-grunt.js');
var path = require('path');

process.chdir(path.dirname(__dirname));

safeExec('node script/bootstrap', function() {
  runGrunt(["lint"], process.exit);
});
```

but I kind of think this is overkill.

damieng:
This targets our old grunt based system which is now gone, therefore closing this PR.

PR #11552, opened 2016-04-20
Content Security Policy fixes

ggreer:
Add blob: protocol to img-src and media-src sources. Add data: and mediastream: protocols to media-src.

These changes are needed to fix video chat in the Floobits Atom plugin (issue https://github.com/Floobits/floobits-atom/issues/114). A CSP directive of `default-src: *` doesn't cover protocols like `blob:`, `data:`, or `mediastream:`, which the Floobits package uses.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <meta http-equiv="Content-Security-Policy" content="default-src *; img-src data: *; script-src 'self'; style-src 'self' 'unsafe-inline';">
+  <meta http-equiv="Content-Security-Policy" content="default-src *; img-src blob: data: *; script-src 'self'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: *;">
   <script src="index.js"></script>
 </head>
 <body tabindex="-1">


ggreer:
I'm not sure what the Travis CI failure is about. It looks like some sort of unrelated timeout.

mertkahyaoglu:
I need this in my project as well. The code below is not working right now;
`video.src = window.URL.createObjectURL(stream)`

ggreer:
Can you please merge my PR? This bug is costing me money.

lee-dohm:
We're still looking in to this to ensure there aren't any unforeseen consequences. Relaxing the security could cause problems that we haven't seen before. The developers are looking into it and if they give the thumbs up, we'll go ahead and merge this. I'll have another update for you on Friday.

ggreer:
This patch actually gets CSP behavior closer to what it used to be. [Chromium has been tightening down how they parse CSP headers](https://codereview.chromium.org/1178373004). The CSP line in the Atom source has been the same for years, but only recently did a new enough version of Chromium (in the form of Electron) ship in Atom to cause breakage. 

See https://bugs.chromium.org/p/chromium/issues/detail?id=473904 for more patchsets related to stricter CSP parsing.

lee-dohm:
Talked it over with the devs and the security experts. They gave the 👍

PR #11562, opened 2016-04-21
Use ohnogit

joshaber:
We’ve pulled the brains of `GitRepositoryAsync` out into its own package: https://github.com/github/ohnogit

This lets others use it for projects other than Atom.


joshaber:
@atom/core FYI. The conversion was pretty mechanical so I don't expect any fallout but feel free to cc me if you see something suspicious.

nathansobo:
Quite a name!

PR #11580, opened 2016-04-22
Makes script/build experience better for Windows users.

damieng:
This does two things:
1. Changes the default install folder to better match atomsetup.exe location, e.g. `c:\users\bob\appdata\local\atom\app-dev\` instead of `c:\program files\atom`
2. Only requests Admin rights if it can't write to the target instead of always prompting UAC


50Wliu:
Should the `install` step also add the `atom` and `apm` commands to the PATH?

damieng:
I wonder that myself. There are reasons for (you can use it in one step) and against (what if you already had a release version of atom on your path). Open to debate.

PR #11581, opened 2016-04-22
Use an async dialog box

### LINKED ISSUE #11551: "Keep Waiting" dialog blocks until closed
### Prerequisites
- [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
- [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?
### Description

The "keep waiting" dialog that appears when Atom becomes unresponsive blocks until it is closed, preventing Atom from ever becoming responsive again until "keep waiting" is selected.  This is because Atom is performing a synchronous call to `dialog.showMessageBox` instead of an async one.  In addition, there is no code in place to automatically close the dialog if/when Atom does regain responsiveness.
### Steps to Reproduce
1. Get Atom into an unresponsive state and have the "keep waiting" dialog appear
2. Wait for Atom to become responsive again
3. :zzz:

**Expected behavior:** Atom to regain responsiveness

**Actual behavior:** Atom hangs forever until an option is selected
### Versions

```
Atom    : 1.9.0-dev-8ece1ee
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1
```

```
apm  1.9.2
npm  2.13.3
node 0.10.40
python 2.7.10
git 2.7.0.windows.1
visual studio 2015
```

Windows 10 x64.

winstliu:
This way we don't block the main process until the dialog box is closed.  I'm not sure that this actually works - I experienced Atom hanging once with this branch built but didn't receive a dialog.

(See http://electron.atom.io/docs/v0.37.7/api/dialog/#dialogshowmessageboxbrowserwindow-options-callback for the notes about sync/async)

Fixes #11551

### DIFF
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -131,7 +131,7 @@ class AtomWindow
         message: 'Editor is not responding'
         detail: 'The editor is not responding. Would you like to force close it or just keep waiting?'
       },
-      callback(response) => @browserWindow.destroy() if response is 0
+      (response) => @browserWindow.destroy() if response is 0
 
     @browserWindow.webContents.on 'crashed', =>
       global.atomApplication.exit(100) if @headless
--- a/src/browser/atom-window.coffee
+++ b/src/browser/atom-window.coffee
@@ -125,12 +125,13 @@ class AtomWindow
     @browserWindow.on 'unresponsive', =>
       return if @isSpec
 
-      chosen = dialog.showMessageBox @browserWindow,
+      dialog.showMessageBox @browserWindow, {
         type: 'warning'
         buttons: ['Close', 'Keep Waiting']
         message: 'Editor is not responding'
         detail: 'The editor is not responding. Would you like to force close it or just keep waiting?'
-      @browserWindow.destroy() if chosen is 0
+      },
+      callback(response) => @browserWindow.destroy() if response is 0
 
     @browserWindow.webContents.on 'crashed', =>
       global.atomApplication.exit(100) if @headless


PhiLhoSoft:
Mmm, I volunteered for a test, but I am not sure how to use this PR.
As said, I am on Windows, I see a 50 MB Atom binary, looks like generated JS is embedded in it. At least, I haven't found any JS file that I could patch.

50Wliu:
If you want to invest in the effort you can try [building from source](https://github.com/atom/atom/blob/master/docs/build-instructions/windows.md).

PhiLhoSoft:
I fear I won't have the time to do that, sorry.

Ben3eeE:
@50Wliu I invested in the effort and tested this on Windows 10, mostly cause I wanted a reason to build Atom from source and learn something in the process.
To get Atom to hang and show the dialog I alt-clicked on a big folder in the tree-view.
With this change I am able to scroll the text editor when the dialog is open but not do anything else like selecting text or accessing menu.

#### Result gifs
Getting lag, closing Atom and waiting for Atom to close works:
working maybe

Getting lag, waiting and Atom updates after dialog is shown:
new one

When you have many folders expanded in the tree view Atom is very laggy, closing Atom during this state also shows a keep waiting dialog after Atom has closed that closes itself after a short while. No idea what is going on:
close lag

#### Also noticed
If Atom crashes while hanging the "Atom has crashed" dialog never shows, even after you keep waiting in the "Editor is not repsonding" multiple times. So it seemed like this PR didn't work rarely. This is not limited to this version however, Atom 1.7.3 shows the same behavior.
I actually had quite many crashes while doing this both on this branch and on 1.7.3 so that might be worth investigating maybe but there are multiple issues related to opening/closing in the tree-view crashing atom so maybe not.

Also congratulations to me for managing to build from source :tada:

50Wliu:
> Atom updates after dialog is shown, dialog does not automatically close

Electron currently doesn't provide any way to programatically remove dialog boxes.  I'll create an issue shortly.

> closing Atom during this state also shows a keep waiting dialog after Atom has closed that closes itself after a short while. No idea what is going on

This has happened to me for forever.  Somehow the browser window is going into an unresponsive state while closing.

50Wliu:
Turns out that all of the dialog boxes are currently implemented synchronously, but most of them are heavily tied into Atom's internals.  Working on converting the rest now.

Ben3eeE:
> Somehow the browser window is going into an unresponsive state while closing.

It seemed easier to get the keep waiting dialog to appear when I closed Atom while it was hanging as my first gif. Like the timeout for showing the dialog when it doesn't respond to a close is much shorter. I can actually reproduce it with a small folder that only gives 1-2 seconds of hang when opening.

I did some quick testing on the crashes. Maybe some of them are related to corrupted project state from closing during a hang. I was able to reproduce this on another computer with 1.7.3 when testing now. Crashing by just alt-clicking and waiting is something I have only been able to reproduce in 1.7.3 so maybe that has been fixed?

1. Alt-click on a folder that gives a short amount of lag.
2. Close Atom before lag ends.
3. Get (un)lucky and a keep waiting dialog appears.
4. Force close Atom with the dialog.
5. Atom crashes all subsequent starts until state is cleared.

ADTC (approved):
I would like these changes to be adopted, because the dialog box is in fact automatically killed off (dismissed) when it is no longer needed. This is better than the current behavior, which keeps the dialog box open and waiting indefinitely for user input, even though Atom is ready to exit/quit anyway.

ADTC:
@Ben3eeE Regarding "No idea what is going on:" if my observation is correct, Atom does not have any ability to programmatically close the dialog box (whether sync or async). The reason why the async dialog box seems to close is because when a dialog box is async, it can be killed off by the operating system without the X button being clicked. If you noticed, that only happens when you try to close Atom. In this case, Atom's main window closes, then the async dialog box gets **killed off by the OS** because Atom's main window has already closed. It's not Atom that closes the dialog box. In the case of a sync dialog box, the OS cannot kill it off, because it's a sync dialog box - it needs to wait for user input before proceeding further.

This is actually great for the use case of trying to close or exit Atom - async dialog boxes would make more sense to notify about stuck threads like the Editor, but allowing it to become unstuck and close properly. And since the dialog box is async, when the stuck threads become unstuck and close properly, the OS will kill off the unnecessary async dialog box.

Perhaps an expansion of this idea, that you can explorer further is, what if you create a new thread and invoke an async (or sync) dialog from that thread? Then Atom's main thread can simply kill off that dialog thread when it's no longer necessary.

Issue #10979, opened 2016-02-26
Disentangling `TextEditor` dependencies and reviving the use of the constructor


nathansobo:
@joshaber and I were talking this morning about how unfortunate it is to require `TextEditor` instances to be constructed via the `atom.workspace.buildTextEditor` factory method instead of just via a normal constructor.

In https://github.com/atom/atom/pull/8968, we disentangled all the global references and moved to straight dependency injection. Because `TextEditor` had many dependencies, at that time we decided to deprecate the `TextEditor` export in the module returned by `require('atom')` and encourage people to use the factory so that these dependencies could be injected by the method without references to globals.

Now we're realizing that was probably a mistake. If you look at the dependencies of `TextEditor`, all but `clipboard` can actually be eliminated via various approaches. Some of them absolutely _should_ be eliminated, because we're using them to subscribe to global events such as `atom.grammars.onDidAddGrammar`. This creates a requirement that package authors call `destroy` on `TextEditor` instances they construct or otherwise they leak editor instances because these subscriptions are never disposed. Other dependencies _can_ be eliminated in various ways but aren't actually causing leaks.

Let's go through injected dependencies one at a time:
- [x] `notificationManager` This is for notifying the user of the cursor scope. We can add a `getCursorScope` method and implement this elsewhere as well.
- [x] `packageManager` This is just to emit grammar loaded hooks. Again, seems like we could emit events on individual editors and do this elsewhere.
- [x] `viewRegistry` We won't need this at all if people implement `getElement`.
- [x] `project` We set the project's path when the file is saved for the first time. Doesn't need to happen inside editor.
- [x] `assert` A method for sending stack traces of unexpected states to the exception tracker. We don't necessarily need this everywhere and could even maybe look the other way using a global.
- [x] `applicationDelegate` This is used to open prompts about file size. Could be moved and hopefully someday won't matter.
- [ ] `grammarRegistry` Again, when grammars are loaded, match them against editors from the outside.
- [ ] `config` Everything this controls can be managed from the outside (the workspace) and set options on the `TextEditor` directly. The only downside is that it would make it harder for package authors to match the settings of editors in panes.
- [ ] `clipboard` This one is needed or you can't copy and paste. But could we have a private global for it via the module system. Essentially a singleton you `require`?

Some of these cases are more clear than others. What's less clear for me is the config and grammar registry. What if the package author _wants_ a grammar to be automatically assigned based on the editor's path? What if they want to follow along with the user's settings for editors in panes?

[...]

PR #11603, opened 2016-04-25
Remove TextEditor's dependency on PackageManager

joshaber:
This is step 2 as laid out by @nathansobo in https://github.com/atom/atom/issues/10979.

Remove `TextEditor`s dependency on `PackageManager`.

### DIFF
--- a/src/tokenized-buffer.coffee
+++ b/src/tokenized-buffer.coffee
@@ -129,7 +129,7 @@ class TokenizedBuffer extends Model
     @emitter.emit 'did-change-grammar', grammar
 
     # Delay this to the next tick to ensure whoever created the buffer has the
-    # change to listen for this event before we send it.
+    # chance to listen for this event before we send it.
     process.nextTick =>
       @emitter.emit 'did-use-grammar', grammar
 
--- a/src/tokenized-buffer.coffee
+++ b/src/tokenized-buffer.coffee
@@ -29,14 +29,13 @@ class TokenizedBuffer extends Model
       state.buffer = atomEnvironment.project.bufferForPathSync(state.bufferPath)
     state.config = atomEnvironment.config
     state.grammarRegistry = atomEnvironment.grammars
-    state.packageManager = atomEnvironment.packages
     state.assert = atomEnvironment.assert
     new this(state)
 
   constructor: (params) ->
     {
       @buffer, @tabLength, @ignoreInvisibles, @largeFileMode, @config,
-      @grammarRegistry, @packageManager, @assert, grammarScopeName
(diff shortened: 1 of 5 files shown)


src/tokenized-buffer.coffee
```diff
@@ -126,9 +125,17 @@ class TokenizedBuffer extends Model
     @disposables.add(@configSubscriptions)
 
     @retokenizeLines()
-    @packageManager.triggerActivationHook("#{grammar.packageName}:grammar-used")
+
     @emitter.emit 'did-change-grammar', grammar
 
+    # Delay this to the next tick to ensure whoever created the buffer has the
+    # change to listen for this event before we send it.
+    process.nextTick =>
```
joshaber: @atom/core Thoughts on this? Alternatively, consumers could assume this happens synchronously and Deal With the fact that they'll miss this. But I'd rather not build in more synchronous assumptions.

src/tokenized-buffer.coffee
```diff
@@ -126,9 +125,17 @@ class TokenizedBuffer extends Model
     @disposables.add(@configSubscriptions)
 
     @retokenizeLines()
-    @packageManager.triggerActivationHook("#{grammar.packageName}:grammar-used")
+
     @emitter.emit 'did-change-grammar', grammar
 
+    # Delay this to the next tick to ensure whoever created the buffer has the
+    # chance to listen for this event before we send it.
+    process.nextTick =>
```
joshaber: @atom/core Thoughts on this? Alternatively, consumers could assume this happens synchronously and Deal With the fact that they'll miss this. But I'd rather not build in more synchronous assumptions.

joshaber: Or maybe `TokenizedBuffer` could emit the current grammar when someone subscribes?

as-cii: I am not sure how I feel about using `process.nextTick` here, because it seems like the pattern we're trying to prevent here is that the users of this API write something like:

```js
let buffer = new TokenizedBuffer
this.handleGrammarUsed(buffer.getGrammar())
buffer.onDidChangeGrammar(this.handleGrammarUsed)
```

Which isn't necessarily wrong or bad, but I get that we could have a nicer API for this. Similar to your proposal above, how would you feel about using the `observeFoo` pattern, as we do [elsewhere in the codebase](https://github.com/atom/atom/blob/master/src/workspace.coffee#L255-L262)?

```js
let buffer = new TokenizedBuffer
buffer.observeGrammar(this.handleGrammarUsed)
```

joshaber: Good call 👍 It turns out that already exists!

PR #11606, opened 2016-04-25
Fix infinite recursion in GitRepositoryAsync

joshaber:
Introduced in https://github.com/atom/atom/pull/11562.


as-cii:
Looks good to me! What do you think about adding specs to cover these APIs and make sure we don't regress again?

joshaber:
Good call 👍 It turns out `getUpstreamBranch` was buggy as hell: https://github.com/github/ohnogit/pull/4

PR #11616, opened 2016-04-26
Remove TextEditor's dependency on assert

joshaber:
This is step 5 as laid out by @nathansobo in https://github.com/atom/atom/issues/10979.

Remove `TextEditor`s dependency on `assert`.

I’m totally open to debate on how we want to handle this. @atom/core opinions wanted.


src/text-editor.coffee
```diff
@@ -101,8 +101,8 @@ class TextEditor extends Model
     throw new Error("Must pass a config parameter when constructing TextEditors") unless @config?
     throw new Error("Must pass a clipboard parameter when constructing TextEditors") unless @clipboard?
     throw new Error("Must pass a grammarRegistry parameter when constructing TextEditors") unless @grammarRegistry?
-    throw new Error("Must pass an assert parameter when constructing TextEditors") unless @assert?
 
+    @assert ?= @defaultAssert.bind(this)
```
joshaber: This way Atom can continue to use `assert` as it does today and other consumers can provide their own assert if they want.

as-cii: I think I am fine with this way of handling it: we could even use a global, but I like that all the dependencies get listed in this object's constructor.

I find `defaultAssert` to be somewhat confusing, though. How would you feel about inlining it to show it's a no-op?

```coffee
@assert ?= (condition) => condition
```

And if we find ourselves repeating this pattern over and over, we could extract a file like `no-op-assert.js' and `require` it:

```coffee
@assert ?= require('no-op-assert')
```

What do you think?

joshaber: >we could even use a global, but I like that all the dependencies get listed in this object's constructor.

Me too 👍 

>I find defaultAssert to be somewhat confusing, though. How would you feel about inlining it to show it's a no-op?

👍

PR #11617, opened 2016-04-26
🏇 Improve perf for clicking on long buffer rows

### LINKED ISSUE #10769: Clicking on empty space below long line freezes Atom

rahatarmanahmed:
As discussed in https://github.com/atom/atom/pull/11539#issuecomment-213368868 by @as-cii, this PR reimplements the `screenPositionforPixelPosition` method in `lines-yardstick.coffee` to use two binary searches to find the position in order to minimize the number of `getBoundingClientRect` calls. The first stage performs a binary search on the text nodes of a row to find the node containing the pixelPosition. The second stage performs a binary search within the text node to find the character that contains the pixelPosition. This improves responsiveness when clicking near the end of a very long line of text.

~~This adds a dependency on [`binary-search-with-index`](https://www.npmjs.com/package/binary-search-with-index), which is my fork of [`binary-search`](https://www.npmjs.com/package/binary-search). This fork passes the index to the comparator which is needed in the `characterComparator`, and also relicenses the package to MIT. The original package was CC0-1.0 and waived all copyrights, so this should be okay.~~

**Update**: The original package accepted my PR, so the only thing preventing us from using it is that its `CC0-1.0` license doesn't pass the [`legal-eagle`](https://github.com/atom/legal-eagle) check. Might be a better idea to patch `legal-eagle` to accept `CC0-1.0` since it does accept public domain.

Resolves #10769

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "async": "0.2.6",
     "atom-keymap": "6.3.2",
     "babel-core": "^5.8.21",
-    "binary-search-with-index": "^1.3.0",
     "bootstrap": "^3.3.4",
     "cached-run-in-this-context": "0.4.1",
     "clear-cut": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "async": "0.2.6",
     "atom-keymap": "6.3.2",
     "babel-core": "^5.8.21",
+    "binary-search-with-index": "^1.3.0",
     "bootstrap": "^3.3.4",
     "cached-run-in-this-context": "0.4.1",
     "clear-cut": "^2.0.1",
--- a/src/lines-yardstick.coffee
+++ b/src/lines-yardstick.coffee
@@ -1,6 +1,5 @@
 {Point} = require 'text-buffer'
 {isPairedCharacter} = require './text-utils'
-binarySearch = require 'binary-search-with-index'
 
 module.exports =
 class LinesYardstick
@@ -20,8 +19,8 @@ class LinesYardstick
     targetTop = pixelPosition.top
     targetLeft = pixelPosition.left
     defaultCharWidth = @model.getDefaultCharWidth()
-    return Point(0, 0) if targetTop < 0
     row = @lineTopIndex.rowForPixelPosition(targetTop)
+    targetLeft = 0 if targetTop < 0
     targetLeft = Infinity if row > @model.getLastScreenRow()
     row = Math.min(row, @model.getLastScreenRow())
     row = Math.max(0, row)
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "async": "0.2.6",
     "atom-keymap": "6.3.2",
     "babel-core": "^5.8.21",
+    "binary-search-with-index": "^1.3.0",
```
as-cii: I'd prefer not to pull in a dependency for binary search, especially since that package contains extra stuff we don't need in this code path (such as checking for `undefined`). I'd prefer having something like:

```js
function binarySearch (array, compare) {
  let low = 0
  let high = array.length
  while (low <= high) {
    let mid = low + (high - low >> 1)
    let comparison = compare(array[mid], mid)
    if (comparison < 0) {
      low = mid + 1
    } else if (comparison === 0) {
      return mid
    } else {
      high = mid - 1
    }
  }

  return -1
}
```

It's not exactly DRY, but the cost of depending on a library seems higher in this case.

src/lines-yardstick.coffee
```diff
@@ -19,8 +20,8 @@ class LinesYardstick
     targetTop = pixelPosition.top
     targetLeft = pixelPosition.left
     defaultCharWidth = @model.getDefaultCharWidth()
+    return Point(0, 0) if targetTop < 0
```
as-cii: Interesting. Why are we returning early here? Seems like the routine below can handle it, even if the cost is a little higher. On the other hand, performance will be consistent in all scenarios.

src/lines-yardstick.coffee
```diff
@@ -32,32 +33,27 @@ class LinesYardstick
     lineOffset = lineNode.getBoundingClientRect().left
     targetLeft += lineOffset
 
-    textNodeStartColumn = 0
-    for textNode in textNodes
-      {length: textNodeLength, textContent: textNodeContent} = textNode
-      textNodeRight = @clientRectForRange(textNode, 0, textNodeLength).right
+    textNodeComparator = (textNode, position) =>
```
as-cii: How would you feel about inlining these comparators? Seems like it increasing locality increases clarity as well in this case.

src/lines-yardstick.coffee
```diff
@@ -32,32 +33,27 @@ class LinesYardstick
...
-          if isPairedCharacter(textNodeContent, characterIndex)
-            nextCharacterIndex = characterIndex + 2
-          else
-            nextCharacterIndex = characterIndex + 1
+    textNodeIndex = binarySearch(textNodes, targetLeft, textNodeComparator)
 
-          rangeRect = @clientRectForRange(textNode, characterIndex, nextCharacterIndex)
+    if textNodeIndex >= 0
+      textNodeStartColumn = textNodes
+        .slice(0, textNodeIndex)
```
as-cii: Can we avoid duplicating the array contents in this case? We could scan through all the characters and use an imperative approach, which would make this code path even faster.

src/lines-yardstick.coffee
```diff
@@ -107,3 +103,22 @@ class LinesYardstick
     @rangeForMeasurement.setStart(textNode, startIndex)
     @rangeForMeasurement.setEnd(textNode, endIndex)
     @rangeForMeasurement.getClientRects()[0] ? @rangeForMeasurement.getBoundingClientRect()
+
+  charIndexForScreenPosition: (textNode, targetLeft) ->
```
as-cii: Seems like we could inline this function too. 💭

src/lines-yardstick.coffee
```diff
@@ -107,3 +103,22 @@ class LinesYardstick
...
+    {textContent: textNodeContent} = textNode
+    rangeRect = null
+    nextCharIndex = -1
+    characterComparator = (char, position, charIndex) =>
```
as-cii: Similar to the comment above, can we inline this comparator? The extra cognitive overhead of jumping to another function doesn't seem required in this case.

as-cii:
Thanks for this, @rahatarmanahmed! I left a bunch of comments, and I'll happily merge it as soon as they get addressed. 👍

src/lines-yardstick.coffee
rahatarmanahmed: It was a small optimization, but I can certainly revert this if needed.

rahatarmanahmed:
@as-cii Added a commit to address your comments. Let me know if you need me to rebase if you can't use github's squash merge.

as-cii:
This should go out with #11414. Thanks, @rahatarmanahmed!

PR #11618, opened 2016-04-26
File Open dialog defaults to directory of active editor or project root if no active editor

dav-ell:
Based on #6762 and suggestion by @lalomartins:

> if this is a huge project, it might be worth it as a temporary fix to open the file chooser in the direction it was last. So, whenever you select a file using the chooser, store the location as state somewhere, then next time start in the same place, if available. It would already be a lot less annoying, I believe, and that's what browsers do with the “save as” dialogue.

Also note that the code this replaces (`openOptions.defaultPath = projectPath`) was completely non-functional. Apparently `AtomWindow.projectPath` was removed a while ago, possibly as a result of #770.

### DIFF
--- a/src/register-default-commands.coffee
+++ b/src/register-default-commands.coffee
@@ -1,6 +1,7 @@
 {ipcRenderer} = require 'electron'
 
 module.exports = ({commandRegistry, commandInstaller, config, notificationManager}) ->
+
   commandRegistry.add 'atom-workspace',
     'pane:show-next-recently-used-item': -> @getModel().getActivePane().activateNextRecentlyUsedItem()
     'pane:show-previous-recently-used-item': -> @getModel().getActivePane().activatePreviousRecentlyUsedItem()
@@ -31,11 +32,15 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
     'application:unhide-all-applications': -> ipcRenderer.send('command', 'application:unhide-all-applications')
     'application:new-window': -> ipcRenderer.send('command', 'application:new-window')
     'application:new-file': -> ipcRenderer.send('command', 'application:new-file')
-    'application:open': -> ipcRenderer.send('open-command', 'application:open')
-    'application:open-file': ->
+    'application:open': -> 
+      defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
+      ipcRenderer.send('open-command', 'application:open', defaultPath)
+    'application:open-file': -> 
       defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
       ipcRenderer.send('open-command', 'application:open-file', defaultPath)
-    'application:open-folder': -> ipcRenderer.send('open-command', 'application:open-folder')
+    'application:open-folder': -> 
+      defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
+      ipcRenderer.send('open-command', 'application:open-folder', defaultPath)
     'application:open-dev': -> ipcRenderer.send('command', 'application:open-dev')
     'application:open-safe': -> ipcRenderer.send('command', 'application:open-safe')
     'application:add-project-folder': -> atom.addProjectFolder()
--- a/src/register-default-commands.coffee
+++ b/src/register-default-commands.coffee
@@ -31,11 +31,13 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
     'application:unhide-all-applications': -> ipcRenderer.send('command', 'application:unhide-all-applications')
     'application:new-window': -> ipcRenderer.send('command', 'application:new-window')
     'application:new-file': -> ipcRenderer.send('command', 'application:new-file')
-    'application:open': -> ipcRenderer.send('command', 'application:open')
-    'application:open-file': -> ipcRenderer.send('command', 'application:open-file')
-    'application:open-folder': -> ipcRenderer.send('command', 'application:open-folder')
-    'application:open-dev': -> ipcRenderer.send('command', 'application:open-dev')
-    'application:open-safe': -> ipcRenderer.send('command', 'application:open-safe')
+    'application:open': -> ipcRenderer.send('open-command', 'application:open')
+    'application:open-file': ->
+      defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
+      ipcRenderer.send('open-command', 'application:open-file', defaultPath)
+    'application:open-folder': -> ipcRenderer.send('open-command', 'application:open-folder')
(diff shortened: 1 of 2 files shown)


lee-dohm:
I don't believe this solution will work the way it is intended. Since `src/browser/atom-application.coffee` runs in the main process and there is only one main process shared between all the windows you have open, you can run into unexpected behavior in a situation like this:

1. Launch Atom with `atom ~/foo`
1. Launch a second window with `atom ~/bar`
1. Open the file `~/bar/baz/quux.js`
1. Press <kbd>Alt+Tab</kbd> to switch to the first window (the one with the `~/foo` project in it)
1. Press <kbd>Ctrl+O</kbd> to open the file dialog

**Expected:** Open file dialog to default to path `~/foo`
**(Suspected) Actual:** Open file dialog defaults to path `~/bar/baz`

I suspect what needs to happen is for the `application:open*` commands to pass across the path to the file in the active text editor so that the code you're working with can use it to set the `projectPath` option.

delliott2016:
> I suspect what needs to happen is for the application:open* commands to pass across the path to the file in the active text editor so that the code you're working with can use it to set the projectPath option.

That's exactly the problem, though. Getting the active text editor is not possible (to my knowledge) within the browser process. A reference to the model, workspace, or atom itself is needed, and none of those can be reached from within the file where thee different application:open* directs. That's why this idea was used, because if the active editor cannot be reached, then there isn't really any choice but to attempt and solve it the way a browser (that can also have multiple windows) would.

It seems to me that if this method is not used, then the code for opening files (and possibly folders/all) would need to be moved out of `atom-application.coffee`.

lee-dohm:
> what needs to happen is for the application:open* commands to pass across the path

This code:

https://github.com/atom/atom/blob/44d3f2dd0774b79dbbfabfbad91d114bac81ace6/src/register-default-commands.coffee#L34-L36

executes in the renderer process, not the main process (formerly referred to as the browser process). Since it executes in the renderer process, it has access to the `atom` global and the full Atom API, including the active editor. These commands would need to be modified to pass across the path to the file in the active editor to the main process where the code you have edited could use it.

delliott2016:
Ah, that'll work. I assume the code in the main process would then just save a different path for each active editor path that it has received?

lee-dohm:
It wouldn't have to save anything. Here's the flow:

1. Launch Atom using `atom ~/foo`
1. Open the file dialog using <kbd>Ctrl+O</kbd> (which defaults to the project path at `~/foo`, also passed across since there is no active editor)
1. Open the file `~/foo/bar/baz.js`
1. Open the file dialog using <kbd>Ctrl+O</kbd> ... this time defaulting to the path `~/foo/bar` because the command passed across `~/foo/bar/baz.js` from the active editor
1. Cancel out of the file open dialog
1. Open the Settings View
1. Open the file dialog using <kbd>Ctrl+O</kbd> ... once again defaulting to the project path at `~/foo` because there is no active editor

Make sense?

delliott2016:
I think so. It looks like this is pretty similar to the original idea of defaulting to the current project directory, thanks to the possibility of using IPC to pass the path. Do you happen to have a reference I could use for how to pass and receive parameters through the IPC renderer? I've tried to use Electron's documentation before, but couldn't translate it to what we have in coffescript.

delliott2016:
We have the following from Electron's documentation:

The `ipcRenderer` module has the following method to listen for events:

#### `ipcRenderer.on(channel, listener)`

 - `channel` String
 - `listener` Function

Listens to `channel`, when a new message arrives `listener` would be called with `listener(event, args...)`.

---

The code from `atom-application.coffee` is:

    @on 'application:open-file', -> @promptForPathToOpen('file', getLoadSettings())

The `listener` seems to be the second argument, the anonymous function calling `promptForPathToOpen`. Should the code then be:

    @on 'application:open-file', (event, args...) -> @promptForPathToOpen('file', getLoadSettings())

Except that I have tried this syntax and both `event` and `args` always come back undefined.

delliott2016:
I got the path to the active editor to be passed through. However, I'm unable to get the project path. I'm currently using this to get the path to the active editor:

    @getModel().getActivePane().getActiveItem().getPath?()

Do you know if there's a similar or better command I can use for the project path?

lee-dohm:
Yes, here's some code that should work much better:

```coffee
defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
```

Then just pass `defaultPath` through.

delliott2016:
What does this mean in the travis-ci build procedure?

>    [50557:0427/183327:INFO:CONSOLE(235)] "(node) warning: possible EventEmitter memory leak detected. %d %s listeners added. Use emitter.setMaxListeners() to increase limit.", source: events.js (235)

lee-dohm:
@delliott2016 That's something that we're trying to work out of the system, I've rerun the specs and they're failing for a different reason that shouldn't be a problem. I'm going to try testing this locally and see how it works.

lee-dohm:
This doesn't appear to be working on OS X. @50Wliu can you try this out on your machine and see if it works for you?

lee-dohm:
Oh, I see why. <kbd>Cmd+O</kbd> on OS X and <kbd>Ctrl+O</kbd> on Linux and Windows is mapped to `application:open`, which this change doesn't fix. @delliott2016, you should add the `defaultPath` change to all three `application:open` derivatives.

src/register-default-commands.coffee
```diff
@@ -31,11 +32,15 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
...
-    'application:open-file': ->
+    'application:open': -> 
+      defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
+      ipcRenderer.send('open-command', 'application:open', defaultPath)
+    'application:open-file': -> 
       defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
       ipcRenderer.send('open-command', 'application:open-file', defaultPath)
-    'application:open-folder': -> ipcRenderer.send('open-command', 'application:open-folder')
+    'application:open-folder': -> 
+      defaultPath = atom.workspace.getActiveTextEditor()?.getPath() ? atom.project.getPaths()?[0]
```
delliott2016: This bothers me, having `defaultPath` defined three places. But `atom` is only available when these functions are called later, so it has to be defined inside the callback. If you know of a better way to write this, let me know.

delliott2016:
I added what you suggested, but it seems to have a caused a conflict somewhere. I'm still trying to work out what it might have conflicted with.

lee-dohm:
Great work! Looks like it is doing the right thing to me now 👍

jdeisenberg:
Just downloaded the x86_64.rpm file for version 1.7.4; still opens to recently used.

50Wliu:
@jdeisenberg This feature will only be available starting in Atom 1.9.0.

jdeisenberg:
Thanks.

PR #11634, opened 2016-04-28
:bug: Wait for connection end to get completed data;

livelazily:
When env contains lots of data,eg: custom `LS_COLORS` or a very long long `PATH`, the connection will receive data multiple times in chunk, need to wait for its end to get completed data


lee-dohm:
Thanks for contributing!

Issue #11632, opened 2016-04-27
Tabs open without showing the content of the file


gassius:
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

When clicking on a file in the project tree (single or double clicking) a tab is open with the name of the file (the name being in cursive if is one click, regular if is double-click) but the content of the file is not shown.

After a variable time, the content may be displayed. Sometimes is not displayed even after 2 min (the longest I have waited)

If the same file in the project tree is left-clicked and "Open in New Window" is selected, the content is displayed properly.

This has been tested with atom --safe option and after running atom --clear-window-state with similar results.

The problem don't seems to appear in the first file that is being opened. Only in files afterward.

### Steps to Reproduce

1. Open a file from the project's tree
2. Click on a second file of the project's tree (one or double-click)

**Expected behavior:** 
New tab should open and the file content should be displayed

**Actual behavior:** 
A tab is opened with the name of the file as title, but no content is displayed. Content may appear after a variable period of time or not appear at all.

### Versions

Atom    : 1.7.2
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1

apm  1.9.2
npm  2.13.3
node 0.10.40
python 2.7.6
git 1.9.1

Ubuntu 14.04.4 LTS
[...]

PR #11635, opened 2016-04-28
Add setting to disable vcs integration

febLey:
In addition to this pull request (https://github.com/atom/atom/pull/11367) I've added a setting to disable the vcs integration completely.

It also solves some other issues like:
- https://github.com/atom/tree-view/issues/784
- https://github.com/atom/atom/issues/1766
- https://github.com/atom/atom/issues/11632

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -22,7 +22,7 @@ module.exports =
         type: 'boolean'
         default: false
         title: 'Disable VCS Integration'
-        description: 'Disable all aspects of the VCS integration.'
+        description: 'Disable the VCS integration (tree-view highlighting, line highlighting, branch display in the status bar, file watching, etc.).'
       followSymlinks:
         type: 'boolean'
         default: true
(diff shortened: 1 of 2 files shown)


lee-dohm:
I'm not certain we'll want to accept this PR. I'll bring it up with the team. If we decide to accept it, it will need specs.

febLey:
I'm not sure what could speak against it. It could be a temporary fix for some people with performance issues over the network. It would be nice if those could be fixed without disabling the vcs stuff, but i don't think this will happen soon. And the people that aren't using the vcs features anyway could disable them (in our company we all do our git stuff in the cli). I think the premise of Atom is customization, so the more options the better.

sadick254:
Hey @febLey I strongly think that performance issues with network shared mounts should be addressed without disabling any functionality. We should find a way to optimise for network mapped drives. We placed this issue on our tech backlog and when we have capacity we will surely tackle it. For now, I have to close this since it doesn't reflect the direction we would like to go.

PR #8780, opened 2015-09-14
:checkered_flag: Show/hide Atom shell context menu from settings

BrainCrumbz:
### Features

This PR aims at addressing some issues (see e.g. #5901 ) about Atom integration in Windows shell context menu (i.e., the _right click_ menu in Explorer). In particular, with those changes:
1. user will be able to disable Atom commands in context menu by acting on Atom settings, without resorting to manually editing the registry. Of course s/he can also enable it back again
2. when Atom is updated, again user will not have to re-edit the registry (nor settings) because Atom will "remember" that items were disabled and keep them like that.
### Implementation
1. Windows Registry operations needed to be invoked by more than one place, so they have been extracted from `squirrel-update.coffee` and moved to a new `WinRegistry` module in `win-registry.coffee`.
2. Because registry operations need ChildProcess.spawn functionalities, those have been extract from  `squirrel-update.coffee` as well and moved to a new `Spawner` in `spawner.coffee`.
3. A new checkbox <-> boolean setting has been added to the core settings pane in `config-schema.coffee`.
4. The new setting is monitored by a dedicated module, `WinContextMenuSettings` in `win-context-menu-settings.coffee`. When setting changes, that is responsible for updating the registry by invoking `WinRegistry`
5. `atom.coffee` creates an instance of that monitor, and dispose that when it's time to
6. `WinContextMenuSettings` monitor got its own specs
7. Some `squirrel-update-spec.coffee` specs, related to ignoring errors, where actually a test on spawning functionality, so they have been moved to a new `spawner-spec.coffee`
8. Specs for the shell context menu integration upon update found place in `squirrel-update-spec.coffee`
9. In `squirrel-update-spec.coffee` then there was a repetition on how to invoke a callback, and that has been factored

PR #11639, opened 2016-04-28
:checkered_flag: refactor registry and PoSH out of squirrel-update

BrainCrumbz:
The first aim is to extract Windows Registry and PowerShell operations from squirrel-update: having those has dependencies, it is easier to mock them and thus squirell-update tests are simplified.

In perspective, this PR is functional to further changes addressing issue #5901. Here's why:

The need for such PR arose from another currently open one, #8780 , which combines together a refactoring and a feature improvement related to Windows right-click menu. If this current PR gets merged, then #8780 would be basically superseded, hence closed, and the actual feature improvement could be discussed and brought forward separately.

As things are now, the refactoring part in #8780 needs manual updates and conflict resolution everytime the codebase moves further in the squirrel-update area. By accepting this current PR, that would be set aside once for all.

HTH


BrainCrumbz:
Doh! A lint warning landed in PR.... going to fix that shortly

BrainCrumbz:
Now the [last build](https://travis-ci.org/atom/atom/jobs/126444138) on Travis, the *packages* one on Mac, fails ... but on a first look it does not seem related to current PR changes.

Anyone taking some time to check this?

damieng:
@BrainCrumbz I'll try and take a look over the weekend.  We want to be able to enable/disable shell integration from settings so this would definitely help!

src/browser/spawner.coffee
```diff
@@ -0,0 +1,36 @@
...
+  error = null
+  spawnedProcess.on 'error', (processError) -> error ?= processError
+  spawnedProcess.on 'close', (code, signal) ->
+    error ?= new Error("Command failed: #{signal ? code}") if code isnt 0
+    error?.code ?= code
+    error?.stdout ?= stdout
+    callback?(error, stdout)
+  # This is necessary if using Powershell 2 on Windows 7 to get the events to raise
+  # http://stackoverflow.com/questions/9155289/calling-powershell-from-nodejs
+  spawnedProcess.stdin.end()
```
damieng: Would it be possible to use buffered-process for this? It already takes care of spawning commands and capturing output and it would seem redundant to have two. It doesn't yet do that PowerShell workaround but that could be something to add there.

src/browser/win-registry.coffee
```diff
@@ -0,0 +1,62 @@
...
+spawnReg = (args, callback) ->
+  Spawner.spawn(regPath, args, callback)
+
+# Install the Open with Atom explorer context menu items via the registry.
+#
+# * `callback` The {Function} to call after registry operation is done.
+#   It will be invoked with the same arguments provided by {Spawner.spawn}.
+#
+# Returns `undefined`.
+exports.installContextMenu = (callback) ->
```
damieng: We should break these up into installContextMenuForFolder, installContextMenuForFile and installFileHandler but that can be done as part of the subsequent options work.

src/browser/spawner.coffee
BrainCrumbz: I just gave a first look at `buffered-process` now, of course there are similarities. Given that this PR is the first step in extracting things out of `squirrel-update`, and that it is propaedeutic to other long postponed changes, I'd say to split things first and then improve them further. For another pending PR we have, merging such refactoring on every codebase advance was a pita.

src/browser/win-registry.coffee
BrainCrumbz: Yep, those changes make sense: I would favour clearer separation as well. As said for the previous not, I'd go for the first step of extracting things out of `squirrel-update`, then proceed with further refactoring where needed.

damieng:
I tested this on Windows 10 and it gets my :+1:

PR #11640, opened 2016-04-28
Remove TextEditor's global workspace dependency

joshaber:
Part of https://github.com/atom/atom/issues/10979.

Remove `TextEditor`s dependency on the global workspace.


src/text-editor.coffee
```diff
@@ -609,15 +609,19 @@ class TextEditor extends Model
   # * "<filename>" when it is the only editing buffer with this file name.
   # * "<filename> — <unique-dir-prefix>" when other buffers have this file name.
   #
+  # * `textEditors` An optional {Array} of the other {TextEditor} instances that
```
joshaber: @atom/core This is a change to a public API. Should we default to using the global to preserve backwards compatibility?

joshaber: Or more precisely, the API is still the same (you can still call it with no args) but the behavior is different (if called with no args, don't take other editors into account).

lee-dohm: I feel like we should default to using the global but deprecate it. Give package authors at least a couple minor releases to catch up.

as-cii: I wonder if we can take a slightly different design direction for this. Namely, I am concerned about keeping the `get...Title` responsibility within `TextEditor`: these methods seem to belong to UI models rather than to this object, which ultimately should be just a wrapper around `TextBuffer` + `TokenizedBuffer`. Therefore, how would you feel about extracting a `TextEditorPaneItem` object? It could look similar to this:

```js
class TextEditorPaneItem {
  getTitle ()
  getLongTitle ()
  getPath ()
  // all the other methods that belong to the `PaneItem` interface (which I couldn't find in the docs)
}
```

We could make this change backwards compatible in two steps:

1. Whenever a user tries to call `TextEditor`'s methods on `TextEditorPaneItem`'s prototype, forward them to the underlying editor instance and emit a deprecation.
2. Whenever a user tries to call `TextEditorPaneItem` methods on `TextEditor`'s prototype, forward them to `atom.workspace.paneForItem(this)` and emit a deprecation.

I feel like this solution could be what we might adopt long-term, but I understand this design decision has a larger impact on the users of this APIs, especially because calling `atom.workspace.open` returns a `TextEditor` instance, which means people would get quite a lot of deprecations if it starts returning `TextEditorPaneItem` instances. But I also feel like we should take this as an opportunity to improve Atom's internals design, without being afraid of changing them too much.

What do you think? 💭

joshaber: I agree, separating TextEditor/TextEditorPaneItem makes sense 👍 

The surface area for deprecation makes me nervous. It's gonna be annoying for ... a lot of packages. Do we have much president for this sort of large change?

lee-dohm: I also agree that separating `TextEditor` and `TextEditorPaneItem` makes sense. On the other hand, this is a *huge* change to the API. Just a cursory glance at `Workspace` reveals a bunch of APIs that will need to be touched, altered or rethought.

> Do we have much precedent for this sort of large change?

Only before v1.0. Which makes me feel like we should hold off on splitting these two roles apart until the run up to v2.0.

spec/text-editor-spec.coffee
```diff
@@ -160,15 +160,15 @@ describe "TextEditor", ->
   describe "title", ->
     describe ".getTitle()", ->
       it "uses the basename of the buffer's path as its title, or 'untitled' if the path is undefined", ->
-        expect(editor.getTitle()).toBe 'sample.js'
+        expect(editor.getTitle(atom.workspace.getTextEditors())).toBe 'sample.js'
```
as-cii: Do you think we could avoid passing `atom.workspace.getTextEditors()` here? Seems like the (short) title shouldn't need to refer to other editors.

src/text-editor.coffee
as-cii: > Which makes me feel like we should hold off on splitting these two roles apart until the run up to v2.0.

I think I am okay with that and I can see how it would let us move forward here, but it also makes me feel nervous mainly because of two reasons:

1. `textEditors` feels like a _false extension point_. I am not sure package authors (nor outside libraries/software) are actually going to use it.
2. I think we haven't talked about when `v2.0` is going to happen precisely, which could mean we'd have to stick with this new API for a while.

So I think we basically have three alternatives here:

1. Separate `TextEditorPaneItem` and `TextEditor`, dealing with all the corner cases (e.g. what happens if someone calls `getActivePaneItem() instanceof TextEditor`?) , and handling deprecations appropriately.
2. Pass `textEditors` as a parameter. However: avoid documenting it, avoid emitting a deprecation, and default it to `this.workspace.getTextEditors()`, as we know already this API is unstable and it's gonna change at some point in the future.
3. Pass a `getTextEditors` function in the constructor, and default it to `() => []`. When instantiating it through `Workspace`, pass `this.getTextEditors.bind(this)`.

Personally, I'd favor (in order of preference) 1), 3), 2), but I can definitely be talked out of it. @joshaber @lee-dohm: thoughts? 💭

spec/text-editor-spec.coffee
joshaber: Good catch 👍

src/text-editor.coffee
joshaber: Ugh so many not great options 😬 

@as-cii and I talked about this in chat. Given that:

1. Doing It Right® would require some big deprecations, and
2. Doing It Wrong® doesn't leave us in a much better position than we're in today.

We're just gonna let this global dependency slide for now. Once we have a better idea of what/when 2.0, we can re-visit Doing It Right®.

PR #11642, opened 2016-04-28
Fix some Buffered-Process specs, honor options.shell in Atom

damieng:
Do not do out magic shell logic if options.shell set to anything. Although we don't use a version of node that honors options.shell today this enables us to do the right thing when we hit 5.7 without breaking changes as well as allowing opt-out today.


spec/buffered-process-spec.coffee
```diff
@@ -117,6 +111,7 @@ describe "BufferedProcess", ->
       expect(stdout).toEqual ''
 
   it "calls the specified stdout callback only with whole lines", ->
+    # WINSPEC: Fails because command is too long, /bin/echo is *nix only, odd newlining and quoting
```
as-cii: Can we use any other command that works on Windows too and that verifies the same behavior? Or, alternatively, we can also write a different spec that runs on Windows only: 

```coffee
describe "on #darwin and #linux", ->
  it "calls the specified stdout callback only with whole lines", ->

describe "on #windows", ->
  it "calls the specified stdout callback only with whole lines", ->
```

damieng: I tried to make it run echo on Windows as well by specifying 'shell' but the line length was too short. I reduced that and then quoting was off and it only returned the first block.  

I've been trying to think what would return a lot of lines consistently between the two - apm perhaps? Not sure that's always enabled at this point.  It's difficult to come up with a command on Windows that will return multiple lines consistently regardless of Windows version or language.  Open to suggestions.

as-cii: I am not really familiar with the tools that Windows provides which we might use in this case, but maybe we could have a fixture script/exe? Something that we can execute like the following:

```bash
$ node fixtures/echo.js
```

damieng: A large text file that we `cat` on Linux/Mac and `type` on Windows would probably do it. I'll add that today.

spec/buffered-process-spec.coffee
```diff
@@ -116,29 +112,25 @@ describe "BufferedProcess", ->
...
+    content = fs.readFileSync(loremPath).toString()
+    baseContent = content.split('\n')
     stdout = ''
-    endLength = 10
-    outputAlwaysEndsWithStew = true
+    allLinesEndWithNewline = true
     process = new BufferedProcess
-      command: '/bin/echo'
-      args: [content]
+      command: if platform.isWindows() then 'type' else 'cat'
```
50Wliu: `process.platform is 'win32'`?  Then the extra `platform` require can be removed.

damieng: We're going to support x64 at some point...

spec/buffered-process-spec.coffee
```diff
@@ -116,29 +85,51 @@ describe "BufferedProcess", ->
...
 
     waitsFor -> exitCallback.callCount is 1
 
     runs ->
-      expect(outputAlwaysEndsWithStew).toBeTrue
-      expect(stdout).toBe content += '\n'
+      expect(allLinesEndWithNewline).toBeTrue
+      expect(stdout).toBe content
+
+  describe "on Windows", ->
```
as-cii: Do we want to run these tests only on windows? If so, our version of jasmine supports tags like:

```
it "does foo on #win32"
```

Which will exclude that spec from being run on `darwin` and `linux`. Maybe this way we can avoid mocking `process.platform`?

damieng: The good thing about that mock of process.platform is it ensures the right logic happens on Windows even when you run the specs on the Mac.

PR #11649, opened 2016-04-29
Try to cache *only* node_modules

as-cii:
This pull-request is a test to try out the recently-enabled caching feature in Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,6 @@ script: script/cibuild
 cache:
   directories:
     - node_modules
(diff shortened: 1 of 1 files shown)


as-cii:
Seems like it got cached [here](https://travis-ci.org/atom/atom/jobs/126678827#L369) but in the [next build](https://travis-ci.org/atom/atom/jobs/126700284#L52) it was not retrieved. Closing this, as it doesn't seem to work properly yet.

Issue #11659, opened 2016-04-30
Double Indent for line continuation. 


yufanyufan:
For language c family, when a line continued in the next line, 4 spaces is needed instead of 2 space.
One way to support this is that, language-model add two regex config: doubleIncreaseIndentPattern, doubleDecreaseNextIndentPattern
Expected:

``` c++
void main () {
  function(
      a,
      b);
  }
}
```

Version 1.7.3 gives:

``` c++
void main () {
  function(
    a,
    b);
  }
}
```

albertosantini:
Did you give a try with indentation settings?

However it seems to me this might be, finally, an issue in the language repo or a `Discuss` topic.

Generally speaking, you may read the recent post about [issue guidelines](http://blog.atom.io/2016/04/19/managing-the-deluge-of-atom-issues.html).

yufanyufan:
Hi, albertosantini,

Language repo can't support this unless changed is made to language-mode.coffeescript.

yufanyufan:
Hi, lee-dohm, how do you want to proceed?
An alternative solution is "Allow grammar to override suggestedIndentForTokenizedLineAtBufferRow' https://github.com/atom/atom/issues/9059

PR #11683, opened 2016-05-03
Fix Double Indent for line continuation #11659

yufanyufan:


sadick254:
Thanks for your PR!

PR #11695, opened 2016-05-05
Update language-sass

winstliu:
autocomplete-css specs are testing scopes that changed with the bump to 0.50.0.  Will merge once I fix those specs and the build goes :green_heart:.


50Wliu:
The inconsistency introduced with the scope boundary changes ended up not being worth it, so I've temporarily reverted them.  Once the rest of the language-sass PRs get merged I'll work on improving the autocomplete-css scope matching.

PR #11750, opened 2016-05-13
Add Devtron to the Developer Tools window

lee-dohm:
I decided to be conservative here and:
1. Pin to a specific version of Devtron
2. Only load Devtron when the Developer Tools are first opened

[More information on Devtron.](http://electron.atom.io/devtron/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "clear-cut": "^2.0.1",
     "coffee-script": "1.8.0",
     "color": "^0.7.3",
+    "devtron": "1.1.0",
     "event-kit": "^1.5.0",
     "find-parent-dir": "^0.3.0",
     "first-mate": "^5.1.1",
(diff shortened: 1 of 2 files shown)


as-cii:
How does the extension behave when installed multiple times? e.g. Does toggling the DevTools more than once trigger any issue? I assume it does, but maybe it's worth to double check.

Other than that, :shipit:! Thanks for this, @lee-dohm! ✨

lee-dohm:
> Does toggling the DevTools more than once trigger any issue?

I specifically tested that scenario and I found no issue with it.

as-cii:
I am seeing a bunch of these whenever opening DevTools:

screen shot 2016-05-17 at 11 21 36

I guess it might be fine, but probably worth adding a boolean or something to install that extension only once.

lee-dohm:
Until we update to Electron v1.0.1, we'll have to leave this as-is: https://github.com/electron/devtron/issues/59

as-cii:
@lee-dohm: what do you think about reverting this for the time being and re-adding it later when it will work properly? I feel like it's fine to wait a bit to include devtron if we know it has some issues on the current electron version.

lee-dohm:
@as-cii How do you feel about leaving in the dependency but just not activating it automatically? That way people can activate it manually.

Issue #7765, opened 2015-07-07
Uncaught TypeError: Cannot read property 'match' of undefined


telos-io:
[Enter steps to reproduce below:]
1. ...
2. ...

**Atom Version**: 1.0.0
**System**: Mac OS X 10.10.4
**Thrown From**: Atom Core
### Stack Trace

Uncaught TypeError: Cannot read property 'match' of undefined

```
At /Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/q/q.js:128

TypeError: Cannot read property 'match' of undefined
    at TokenizedBuffer.module.exports.TokenizedBuffer.indentLevelForLine (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:647:23)
    at TokenizedBuffer.module.exports.TokenizedBuffer.indentLevelForRow (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:641:21)
    at TokenizedBuffer.module.exports.TokenizedBuffer.retokenizeWhitespaceRowsIfIndentLevelChanged (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:409:69)
    at TokenizedBuffer.module.exports.TokenizedBuffer.handleBufferChange (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:387:18)
    at /Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:62:24
    at Emitter.module.exports.Emitter.emit (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:82:11)
    at TextBuffer.module.exports.TextBuffer.applyChange (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:626:20)
    at TextBuffer.module.exports.TextBuffer.revertToCheckpoint (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:777:16)
    at TextBuffer.module.exports.TextBuffer.transact (/Users/Koan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:743:14)
[...]

PR #11760, opened 2016-05-16
Fixed calling lineTextForBufferRow with unindented lines.

braian-i-de-roo:
Fixes calling `lineTextForBufferRow()` with unindented rows. Should address #7765

### DIFF
--- a/src/tokenized-buffer.coffee
+++ b/src/tokenized-buffer.coffee
@@ -420,7 +420,7 @@ class TokenizedBuffer extends Model
 
       indentLength / @getTabLength()
     else
-      0
+      indentLength = 0
 
   scopeDescriptorForPosition: (position) ->
     {row, column} = @buffer.clipPosition(Point.fromObject(position))


50Wliu:
Please add specs for this.  Thanks!

blacksunrise56:
Mhh... I'm not very good with specs, [this line](https://github.com/atom/atom/blob/master/spec/tokenized-buffer-spec.coffee#L540) confuses me. Does it mean that when calling `indentLevelForRow` with the row `0` it expects the returned value to be the number of the row? I can't make sense out of it.

50Wliu:
If you open `sample.js`, you should notice that line 1 (aka line 0 internally) has no leading whitespace, line 2 has one soft tab, and line 3 has two soft tabs.

blacksunrise56:
now all makes sense, thanks! I should've read the code thoroughly instead of jumping to conclusions. 

Sadly, after some quick tests, I found this fix works only sometimes. What's more, now sometimes the error doesn't appear even without the fix. I'll check everything from the start again. And I thought it would be easy.

sadick254:
Hey @braian-i-de-roo I am going to close this based on your comment.

> now all makes sense, thanks! I should've read the code thoroughly instead of jumping to conclusions.
Sadly, after some quick tests, I found this fix works only sometimes. What's more, now sometimes the error doesn't appear even without the fix. I'll check everything from the start again. And I thought it would be easy.

PR #11787, opened 2016-05-19
add includeMarkerText param to highlight decorations

abe33:
When passed in the decoration properties, this flag will render the text of the marker into the highlight's regions. This allow a highlight decoration to have a background color and a contrasting enough color for the text on top of it.

Actually, when using highlight decorations, we can't really use them if we have a background color that may conflict with the foreground color: 

styles_less_ __users_cedric_development_ruby_leclerc-es-v2

It means than these decorations, while essential in any good text editor, can be only used to outline/underline a marker.

Now, with the `includeMarkerText` flag enabled the following is now possible: 

json_form_js_coffee_ __users_cedric_development_ruby_leclerc-es-v2

It's far from an ideal solution, but given than since this problem was first reported (almost two years ago), we still don't have a proper solution for that issue. This one is at least simple, don't cost much performance-wise and can be reverted/changed quite easily as well.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2040,6 +2040,20 @@ describe('TextEditorComponent', function () {
       expect(regions.length).toBe(2)
     })
 
+    it('renders the marker text in the highlight', async function () {
+      decoration = editor.decorateMarker(marker, {
+        type: 'highlight',
+        'class': 'test-highlight-with-text',
+        includeMarkerText: true
+      })
+      await decorationsUpdatedPromise(editor)
+      await nextViewUpdatePromise()
+      let regions = componentNode.querySelectorAll('.test-highlight-with-text .region')
+      expect(regions.length).toBe(2)
+      expect(regions[0].textContent).toBe('.length <= 1) return items;')
+      expect(regions[1].textContent).toBe('    var pivot =')
+    })
+
     describe('when flashing a decoration via Decoration::flash()', function () {
       let highlightNode
 
--- a/spec/text-editor-presenter-spec.coffee
+++ b/spec/text-editor-presenter-spec.coffee
@@ -1843,6 +1843,10 @@ describe "TextEditorPresenter", ->
           marker9 = editor.markBufferRange([[0, 0], [2, 0]], invalidate: 'touch')
           highlight9 = editor.decorateMarker(marker9, type: 'highlight', class: 'h')
 
+          # partially off-screen above, 2 of 3 regions on screen, include marker text
+          marker10 = editor.markBufferRange([[0, 6], [3, 6]])
+          highlight10 = editor.decorateMarker(marker10, type: 'highlight', class: 'c', includeMarkerText: true)
+
           presenter = buildPresenter(explicitHeight: 30, scrollTop: 20, tileSize: 2)
 
           expectUndefinedStateForHighlight(presenter, highlight1)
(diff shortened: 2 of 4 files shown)


abe33:
ping @atom/feedback

lee-dohm:
@as-cii I know you were looking at doing something for this. Does this mesh with what you were thinking?

lee-dohm:
ping @atom/core

nathansobo:
*Ideally*, I'd like to address this issue with a different feature, called text decorations.

Display layers, our new system for presenting text on screen, include a facility for wrapping spans around arbitrary regions of text via *text decoration layers*. [Here](https://github.com/atom/text-buffer/blob/3b4cf629a56ed90fcc1820fd64dab832baadd2ae/src/display-layer.coffee#L807) is where we obtain an iterator from the decoration layer and use it to add scopes when we build screen lines, which are transformed into HTML spans in the view layer.

This is currently a private API, and we only support a single decoration layer, which we use to apply syntax highlighting. However, it should be fairly straightforward to allow multiple text decoration layers to interleave. Instead of interacting with a single text decoration layer, the display layer would instead interact with a *composite* text decoration layer, which would cleanly combine decorations from all layers.

We might need a couple more features to make this maximally useful:

First, text decoration layers are currently implemented via a fairly low-level iterator-based API. Exposing an API based on *markers* would probably be more convenient for situations like the color highlighting package.

Second, text decoration layers can currently only apply classes, but it would be possible to allow arbitrary inline styles to be applied to the spans as well. This would make it possible to arbitrarily style the text surrounded by a given marker... something like the following.

```js
editor.decorateMarker(m, {
  type: 'text',
  style: {color: '#f00', fontWeight: 'bold'}
})
```

So that's what I had in mind as the ideal system for styling text. This pull request addresses a much narrower use case, but has the advantage of existing today.

Perhaps we can present this feature in a way that will cleanly slot in with the more complete text styling system I'm describing when it exists in the future. Instead of the boolean `includeMarkerText` option, what about a `textStyle` hash containing styling properties? When we see this hash, we can go ahead and render the text in a span on top of the highlight and apply styling properties from the hash. In the future, we can just create a text decoration instead.

Asking the user to specify an inline style might also make it clearer that the syntax-based styling of the text will be overridden. When text decorations are fully operational, behavior will be slightly different because we'll inherit existing styling on the text whereas with this feature we'll be rendering completely new text, but it probably won't cause too many problems.

Anyway, those are my thoughts. Interested in what others think... @as-cii?

david-driscoll:
This would also solve a bunch of use cases with omnisharp-atom and highlighting know symbols.

as-cii:
I definitely agree that the ideal solution would be using text decoration layers; I do have a couple of concerns about this strategy, although, depending on the priority of having this feature in Atom soon, it might still be okay to 🚢 this anyways.

1. As far as I remember, highlights are rendered behind text: this, for example, allows selections to color the background of a certain piece of text. In order for this to work, and unless I am missing something, it seems like the "marker text" should be in the foreground. Will this approach blend in cleanly with the current way of rendering highlights? 
2. If the decorated text range would break a ligature, it's possible that the replaced text glyphs don't overlap cleanly with the underlying text glyphs rendered on the `.line`. It might be an edge case, but it's probably still worth considering it for other scenarios (e.g. selecting only half of a surrogate pair, or whatnot).
3. Inspecting the DOM might be confusing for package authors because of the _redundant_ text rendering.

> First, text decoration layers are currently implemented via a fairly low-level iterator-based API. Exposing an API based on markers would probably be more convenient for situations like the color highlighting package.
> 
> Second, text decoration layers can currently only apply classes, but it would be possible to allow arbitrary inline styles to be applied to the spans as well. This would make it possible to arbitrarily style the text surrounded by a given marker... something like the following.

Another thing I believe we need to work on before using text decoration layer is implementing an _aggregate_ decoration iterator which is able to report tags based on the state of many decoration layers. A nice idea to make the implementation easier could probably be to have a `CompositeDecorationIterator` that accepts two iterators and adheres to the default decoration iterator interface; if we want to have more than two iterators, we would create another `CompositeDecorationIterator` that combines the third decoration iterator and the previous composite iterator.

@nathansobo @abe33: what do you think?

sadick254:
Hey @abe33. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

PR #11790, opened 2016-05-19
replace OSX window title bar with custom title-bar

brumm:
This pull request is intended to naively implement a custom titlebar on OSX.

I've included fallback styling to match El Capitan's titlebars, which is probably not necessary as I modified the four included themes to provide their own titlebar styling. I'll create PRs for these in their respective repositories if this PR gets considered.

I'd like to hear your thoughts on the implementation, I just banged on the code until it worked 😁 

This is what it looks like by default: 
image

This is the titlebar provided by one-dark-ui, with the oceanic-next syntax
image

This is one-light-ui, with the atom-light syntax
image

Things left to do:
- [x] ✅ specs

This is a partial continuation of https://github.com/atom/atom/pull/10208 albeit without the combined title-bar-and-tabs functionality, for which this would be the groundwork. (As requested in https://github.com/atom/atom/issues/4599 for example.)

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -684,7 +684,7 @@ class AtomEnvironment extends Model
         @deserialize(state) if state?
         @deserializeTimings.atom = Date.now() - startTime
 
-        @document.body.appendChild(@views.getView(@titleBar)) if @titleBar
+        @workspace.addHeaderPanel({item: @views.getView(@titleBar)}) if @titleBar
         @document.body.appendChild(@views.getView(@workspace))
         @backgroundStylesheet?.remove()
 
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -196,13 +196,15 @@ class AtomEnvironment extends Model
       notificationManager: @notifications, @applicationDelegate, @clipboard, viewRegistry: @views, assert: @assert.bind(this)
     })
 
-    @titleBar = new TitleBar() if process.platform is 'darwin'
     @themes.workspace = @workspace
 
     @textEditors = new TextEditorRegistry
     @autoUpdater = new AutoUpdateManager({@applicationDelegate})
 
     @config.load()
+
+    # This needs to happen after config.load()
+    @titleBar = new TitleBar() if process.platform is 'darwin' and @config.get('core.useCustomTitleBar')
 
     @themes.loadBaseStylesheets()
     @initialStyleElements = @styles.getSnapshot()
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -31,6 +31,7 @@ ContextMenuManager = require './context-menu-manager'
 CommandInstaller = require './command-installer'
 Clipboard = require './clipboard'
 Project = require './project'
+TitleBar = require './title-bar'
 Workspace = require './workspace'
 PanelContainer = require './panel-container'
 Panel = require './panel'
@@ -44,6 +45,7 @@ Gutter = require './gutter'
 TextEditorRegistry = require './text-editor-registry'
 AutoUpdateManager = require './auto-update-manager'
 
+TitleBarElement = require './title-bar-element'
 WorkspaceElement = require './workspace-element'
 PanelContainerElement = require './panel-container-element'
 PanelElement = require './panel-element'
@@ -193,6 +195,8 @@ class AtomEnvironment extends Model
       @config, @project, packageManager: @packages, grammarRegistry: @grammars, deserializerManager: @deserializers,
       notificationManager: @notifications, @applicationDelegate, @clipboard, viewRegistry: @views, assert: @assert.bind(this)
     })
+
+    @titleBar = new TitleBar() if process.platform is 'darwin'
     @themes.workspace = @workspace
 
     @textEditors = new TextEditorRegistry
@@ -259,6 +263,8 @@ class AtomEnvironment extends Model
     registerDefaultCommands({commandRegistry: @commands, @config, @commandInstaller, notificationManager: @notifications, @project, @clipboard})
 
   registerDefaultViewProviders: ->
+    @views.addViewProvider TitleBar, (model, env) ->
+      new TitleBarElement().initialize(model, env)
     @views.addViewProvider Workspace, (model, env) ->
       new WorkspaceElement().initialize(model, env)
     @views.addViewProvider PanelContainer, (model, env) ->
@@ -678,6 +684,7 @@ class AtomEnvironment extends Model
         @deserialize(state) if state?
         @deserializeTimings.atom = Date.now() - startTime
 
+        @document.body.appendChild(@views.getView(@titleBar)) if @titleBar
         @document.body.appendChild(@views.getView(@workspace))
         @backgroundStylesheet?.remove()
 
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -557,10 +557,6 @@ class AtomEnvironment extends Model
   # Extended: Set the full screen state of the current window.
   setFullScreen: (fullScreen=false) ->
     @applicationDelegate.setWindowFullScreen(fullScreen)
-    if fullScreen
-      @document.body.classList.add("fullscreen")
-    else
-      @document.body.classList.remove("fullscreen")
 
   # Extended: Toggle the full screen state of the current window.
   toggleFullScreen: ->
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -40,13 +40,13 @@ class AtomWindow
     if process.platform is 'linux'
       options.icon = @constructor.iconPath
 
-    if process.platform is 'darwin' and global.atomApplication.config.get('core.useCustomTitleBar')
+    if @applyTitleBarSetting()
       options.titleBarStyle = 'hidden'
 
     @browserWindow = new BrowserWindow options
     global.atomApplication.addWindow(this)
 
-    if process.platform is 'darwin' and global.atomApplication.config.get('core.useCustomTitleBar')
+    if @applyTitleBarSetting()
       @browserWindow.setSheetOffset(23)
 
     @handleEvents()
@@ -204,6 +204,11 @@ class AtomWindow
     [x, y] = @browserWindow.getPosition()
     [width, height] = @browserWindow.getSize()
     {x, y, width, height}
+
+  applyTitleBarSetting: ->
+    not @isSpec and
+    process.platform is 'darwin' and
+    global.atomApplication.config.get('core.useCustomTitleBar')
 
   close: -> @browserWindow.close()
 
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -23,7 +23,6 @@ class AtomWindow
     options =
       show: false
       title: 'Atom'
-      titleBarStyle: 'hidden'
       # Add an opaque backgroundColor (instead of keeping the default
       # transparent one) to prevent subpixel anti-aliasing from being disabled.
       # We believe this is a regression introduced with Electron 0.37.3, and
@@ -40,6 +39,9 @@ class AtomWindow
     # taskbar's icon. See https://github.com/atom/atom/issues/4811 for more.
     if process.platform is 'linux'
       options.icon = @constructor.iconPath
+
+    if process.platform is 'darwin'
+      options.titleBarStyle = 'hidden'
 
     @browserWindow = new BrowserWindow options
     global.atomApplication.addWindow(this)
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -40,7 +40,7 @@ class AtomWindow
     if process.platform is 'linux'
       options.icon = @constructor.iconPath
 
-    if process.platform is 'darwin'
+    if process.platform is 'darwin' and global.atomApplication.config.get('core.useCustomTitleBar')
       options.titleBarStyle = 'hidden'
 
     @browserWindow = new BrowserWindow options
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -46,6 +46,9 @@ class AtomWindow
     @browserWindow = new BrowserWindow options
     global.atomApplication.addWindow(this)
 
+    if process.platform is 'darwin' and global.atomApplication.config.get('core.useCustomTitleBar')
+      @browserWindow.setSheetOffset(23)
+
     @handleEvents()
 
     loadSettings = Object.assign({}, settings)
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -46,9 +46,6 @@ class AtomWindow
     @browserWindow = new BrowserWindow options
     global.atomApplication.addWindow(this)
 
-    if @applyTitleBarSetting()
-      @browserWindow.setSheetOffset(23)
-
     @handleEvents()
 
     loadSettings = Object.assign({}, settings)
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -23,6 +23,7 @@ class AtomWindow
     options =
       show: false
       title: 'Atom'
+      titleBarStyle: 'hidden'
       # Add an opaque backgroundColor (instead of keeping the default
       # transparent one) to prevent subpixel anti-aliasing from being disabled.
       # We believe this is a regression introduced with Electron 0.37.3, and
--- a/static/title-bar.less
+++ b/static/title-bar.less
@@ -23,7 +23,6 @@
     text-overflow: ellipsis;
   }
 
-
   background-image: linear-gradient(@title-bar-gradient-focused);
   border-bottom: 1px solid @title-bar-border-color-focused;
 
@@ -31,24 +30,5 @@
     color: @title-bar-text-blurred;
     background-image: linear-gradient(@title-bar-gradient-blurred);
     border-bottom-color: @title-bar-border-color-blurred;
-  }
-
-  [title-bar-style="combined"] & {
-    display: none;
-    position: absolute;
-  }
-}
-
-[title-bar-style="combined"] .tab-bar {
-  height: 37px;
-  padding-left: 80px;
-  background-image: linear-gradient(@title-bar-gradient-focused);
-
-  .tab {
-    top: 7px;
-    border-radius: 3px 3px 0 0 !important;
-    &:after {
-      border-radius: 3px 3px 0 0 !important;
-  }
   }
 }
--- a/static/title-bar.less
+++ b/static/title-bar.less
@@ -23,6 +23,7 @@
     text-overflow: ellipsis;
   }
 
+  box-shadow: 0 1px 0 @title-bar-highlight inset;
   background-image: linear-gradient(@title-bar-gradient-focused);
   border-bottom: 1px solid @title-bar-border-color-focused;
 
--- a/static/title-bar.less
+++ b/static/title-bar.less
@@ -1,4 +1,8 @@
-@import "./variables/ui-variables";
+@import "ui-variables";
+
+@title-bar-text-size: 13px;
+@title-bar-background-color: @base-background-color;
+@title-bar-border-color: @base-border-color;
 
 .title-bar {
   height: 23px;
@@ -8,12 +12,11 @@
   align-items: center;
   justify-content: center;
 
-  font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue";
   font-size: @title-bar-text-size;
-  color: @title-bar-text-focused;
   -webkit-user-select: none;
+  -webkit-app-region: drag;
 
-  padding: 0 10px 0 70px;
+  padding: 0 70px;
   overflow: hidden;
 
   .title {
@@ -23,13 +26,10 @@
     text-overflow: ellipsis;
   }
 
-  box-shadow: 0 1px 0 @title-bar-highlight inset;
-  background-image: linear-gradient(@title-bar-gradient-focused);
-  border-bottom: 1px solid @title-bar-border-color-focused;
+  background-color: @title-bar-background-color;
+  border-bottom: 1px solid @title-bar-border-color;
 
   .is-blurred & {
-    color: @title-bar-text-blurred;
-    background-image: linear-gradient(@title-bar-gradient-blurred);
-    border-bottom-color: @title-bar-border-color-blurred;
+    color: @text-color-subtle;
   }
 }
--- a/static/title-bar.less
+++ b/static/title-bar.less
@@ -1,15 +1,17 @@
 @import "ui-variables";
 
 @title-bar-text-size: 13px;
+@title-bar-height: 23px;
 @title-bar-background-color: @base-background-color;
 @title-bar-border-color: @base-border-color;
 
 body.fullscreen .title-bar {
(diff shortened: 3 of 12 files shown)


FezVrasta:
Any change to get rid of the black separation line between the title bar and the three file view? I think it would look much better.

image

brumm:
@FezVrasta These screenshots are just a first stab at theming the title-bar and intended to show what it could look like. At any rate, you could always remove that line by tinkering with your user stylesheet :)

@simurai Would you like me to go ahead and create some PRs on https://github.com/atom/one-dark-ui/ and https://github.com/atom/one-light-ui/?

simurai:
I wonder if the background should just be `@base-background-color`? So that "abandoned, but still in use" themes would adapt too. Then maybe make a PR with the current grey gradient for themes that try to stay close to OS X. Like [unity-ui](https://atom.io/themes/unity-ui) or [native-ui](https://atom.io/themes/native-ui).

src/atom-environment.coffee
```diff
@@ -678,6 +684,7 @@ class AtomEnvironment extends Model
         @deserialize(state) if state?
         @deserializeTimings.atom = Date.now() - startTime
 
+        @document.body.appendChild(@views.getView(@titleBar)) if @titleBar
```
simurai: Would it make sense to add the title bar into the `atom-panel-container.header` with [addHeaderPanel()](https://atom.io/docs/api/v1.7.3/Workspace#instance-addHeaderPanel).

Then you _could_ do some funky stuff like (visually) combining the title-bar with the tool-bar:

screen shot 2016-05-20 at 5 30 32 pm

Probably not by default, but maybe just restyle it in `styles.less` or as an option in the tool-bar package.

static/title-bar.less
```diff
@@ -0,0 +1,54 @@
...
+    border-bottom-color: @title-bar-border-color-blurred;
+  }
+
+  [title-bar-style="combined"] & {
+    display: none;
+    position: absolute;
+  }
+}
+
+[title-bar-style="combined"] .tab-bar {
```
simurai: I guess these `[title-bar-style="combined"]` styles are meant be used as an option to show/hide the title-bar? Maybe we should wait with it and make a new PR once that [dragging fix](https://github.com/electron/electron/pull/5557) made it into Atom.

brumm: Oh dang, these ended up in there by accident. You guessed correctly, I was already working on a compact/combined (`hidden` and `hidden-inset`) titlebar.
I'll remove them for now, since the dragging fix you mentioned is blocking any serious attempts at this.

src/atom-environment.coffee
brumm: Yeah, that's exactly what I was wondering as well. It seemed easier for now to stick the title-bar above the atom-workspace, but I was hoping to also support the `combined` title-bar as well.

I thought the combined titlebar might primarily be hosting the tab-bar, which might be pragmatic for a start, toolbars do sound funky though.

brumm:
>I wonder if the background should just be @base-background-color?

That's a much better idea than using the grey gradient! I'll change it.

brumm:
atom-dark with a title-bar gradient based on `@base-background-color`
image

FezVrasta:
What's the status of this PR? I think the top grey bar is one of the most annoying things of Atom. If I have to use it the whole day I need a pleasant environment, and that gray bar doesn't help at all 😓

In my opinion, this PR should have high priority.

olegakbarov:
I just can't stress how awesome and cool this thing looks to me. Would *love* to see it!

lee-dohm:
Let us know when you have the specs complete for this @brumm :grinning:

src/atom-environment.coffee
simurai: > I thought the combined titlebar might primarily be hosting the tab-bar, which might be pragmatic for a start, toolbars do sound funky though.

Only visually maybe, but tab-bars have to live inside panes. Because they can be split into different directions and each pane as its own tab-bar.

Anyways, I think for now it would be safer to add the title-bar into the .header. Because the changes to `workspace-view.less` might invite other packages to add their bars or other UI stuff outside of `atom-workspace`. With the `addXXXPanel()` all panels can be kept a bit more under control.

static/title-bar.less
```diff
@@ -0,0 +1,35 @@
...
+
+.title-bar {
+  height: 23px;
+
+  flex-shrink: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue";
```
simurai: Maybe we don't have to force a `font-family` and just let it inherit whatever the theme defined. The One themes use the system font, but some themes might wanna use something else.

static/title-bar.less
```diff
@@ -0,0 +1,35 @@
...
+  font-size: @title-bar-text-size;
+  color: @title-bar-text-focused;
+  -webkit-user-select: none;
```
simurai: I think it would be good to already add `-webkit-app-region: drag;`. See http://electron.atom.io/docs/api/frameless-window/#text-selection

Not 100% sure, but otherwise dragging might stop working once that dragging fix lands in Atom.

static/variables/ui-variables.less
```diff
@@ -79,6 +79,19 @@
 
 @tab-height: 30px;
 
+// TitleBar
```
simurai: Not sure about adding lots of new variables to `ui-variables.less`. Since this file gets loaded into many packages, we couldn't make any changes to it later. It's a bit like a public API. For now, or at least till we settle on the "combined" option, might be safer to just keep everything inside `title-bar.less`.

Another idea would be to use `@tab-bar-background-color;`. Since the tab-bar is just underneath, it should match most of the time? And maybe just flat without a gradient. Usually the core styles only style layout and sizing, but don't have much of an opinion on gradients, shadows etc. Maybe something like

```less
.title-bar {
  color: @text-color;
  background-color: @tab-bar-background-color;
  border-bottom: 1px solid @tab-bar-border-color;

  .is-blurred & {
    color: @text-color-subtle;
  }
}
```

Then themes that don't like it, can override it with their own styles.

simurai:
Made a few more comments.. sorry.. :innocent: I tried to make a PR to your PR, but your fork somehow didn't show up in the list to compare against.

brumm:
Thanks for the comments! All good points. I'll be able to work on this in a few hours 👍

brumm:
@simurai i made a couple of changes:
- followed your advice and added title-bar as a headerPanel instead, which allowed me to undo the changes I made to workspace styling
- moved title-bar variables into title-bars' less file and simplified the styling to a single `background-color: @base-background-color`. After much back-and-forth I decided against `@tab-bar-background-color` to preserve some contrast between title-bar and the rest of the UI, but I do like both.
- cleaned up some superfluous code from `title-bar-element` and added a simple test

FezVrasta:
I'm for `@tab-bar-background-color`, it would look much better (and will allow a better integration in the existing themes)

brumm:
`@tab-bar-background-color`
image

`@base-background-color`
image

brumm:
As far as I can tell, the failed travis build doesn't seem to be related to my changes.
Should I squash now or after another once-over?
@lee-dohm @simurai

simurai:
>  After much back-and-forth I decided against `@tab-bar-background-color` to preserve some contrast between title-bar and the rest of the UI, but I do like both.

Yeah, sounds good. For the bundled One + Atom themes, we can still thinking about customizing it. Also having a border by default should help make it clear where you can start dragging the window.

@FezVrasta In case your favorite theme doesn't look seamless, you could always make it transparent in your `styles.less` with:

```css
.title-bar {
  background-color: transparent;
  border-color: transparent;
}
```

> Should I squash now or after another once-over?

Only if you want, so far we haven't really bothered about it. Maybe one day we will enable it on the merge button, now that GitHub added that feature.

> As far as I can tell, the failed travis build doesn't seem to be related to my changes.

Ya, looks like an unlreated timeout.

simurai:
ok, so the styling is 👍 

So hopefully we can get somebody from @atom/core to take a look at the code. And maybe @amytruong can give it another test drive?

src/main-process/atom-window.coffee
```diff
@@ -23,6 +23,7 @@ class AtomWindow
     options =
       show: false
       title: 'Atom'
+      titleBarStyle: 'hidden'
```
BinaryMuse: Is there any chance of this ever doing something on platforms other than OS X? If so, I think it'd be nice to specifically check for the platform here, since we're also explicitly checking it in `atom-environment`, and we don't want it to accidentally get out of sync.

BinaryMuse:
I love this! Other than my above comment, I say it's ready for 🚀 !

src/main-process/atom-window.coffee
brumm: I changed it to check for darwin as well, thanks for the headsup!

brumm:
Thanks for the reviews and comments, everyone! :shipit:

lee-dohm:
So, one thing that this doesn't appear to offer is the title bar icon:

screen_shot_2016-05-27_at_9_48_33_am

This would appear to be a regression. We would either need to maintain this functionality or come up with a way to enable/disable the feature. Both are pretty complex since we don't currently have a system for either having configuration options specific to a platform or restarting when a configuration option is changed.

Is this something that you would be able to cover @brumm?

lee-dohm:
Sorry, the correct term for the thing I was calling a "title bar icon" is "proxy icon":

https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/WindowAppearanceBehavior.html#//apple_ref/doc/uid/20000957-CH33-SW1

brumm:
Hah, I knew I was missing something ;)

I'd hazard a guess that while it would be possible to approximate the proxy icon, it'd be hard/impossible to really get right (a cursory google search suggests that getting a file icon from node is not a solved problem for instance).

I'd be happy to work on providing a settings interface for this feature, which is something I wanted to do in a later PR anyway.

That would mean extending the config-schema with a way to
- Define a `platform` for the setting
- Indicating that this setting will need a restart/reopened window.
  We could probably get away with just mentioning the fact inside the setting description, and then reacting to a setting change with a dialog.

Maybe like this:
```coffeescript
customTitleBar:
  platform: 'darwin' # 'darwin', 'freebsd', 'linux', 'sunos' or 'win32'
  type: 'boolean'
  default: false
  description: 'Use custom window and theme-aware title-bar. Note: This currently does not include a file icon or title context menu.<br />This setting will require a restart to take effect.'
```

If 'polluting' the config namespace is a concern, we could abstract these two new use cases for the config schema to mean
- Setting visibility/applicability
- Setting side effects

```coffeescript
customTitleBar:
  show: { platform: 'darwin' }
  sideEffect: ['restart-app']
  type: 'boolean'
  default: false
  description: 'Use custom window and theme-aware title-bar. Note: This currently does not include a file icon or title context menu.<br />This setting will require a restart to take effect.'
```

FezVrasta:
What's the purpose of such icon?

brumm:
It's a representation (or in osx terms, proxy) of the currently opened file. Click-holding the icon allows you to drag the file out of the title bar to move/copy/alias it.
The title bar also provides a context menu of the currently opened file's parent directories on right-click.

FezVrasta:
It's a bit tricky, but couldn't Electron provide the proxy icon directly from the OS X API like it does for the window control buttons?
Probably it would be difficult/impossible to position the icon on the left of the text, but you could just put in on the right corner maybe.

thomasjo:
@zcbenz Do you think it would be feasibly (aka not too much work) to add an API for this in Electron? And by this, I mean an API that will allow us to basically re-implement the "proxy icon" functionality. We'd ideally need access to the icon somehow, as well as being able to trigger (and position) the menu.

simurai:
I can imagine positioning the native proxy icon would be pretty hard. Adding a config option, as proposed in ☝️ https://github.com/atom/atom/pull/11790#issuecomment-222191019 feels more realistic. Then people can choose to use the native title bar with all its functionality or use the simple, but customizable one.

Maybe depends if platform specific configs would be beneficial for other things too? Or added just for this use-case?

brumm:
>Maybe depends if platform specific configs would be beneficial for other things too? Or added just for this use-case?

That's what I was wondering :)

I'll start with the simple `platform` version and see what others like @lee-dohm might say in the meantime.

brumm:
Right, finally got around to actually doing this.
There are some wording problems in `promptForRelaunch`, since we're not actually relaunching Atom, for which I'll happily accept better phrasing.

lee-dohm:
I'm running into the same error message as the CI, `badges.less` isn't found.

brumm:
Ah, seems like a bad rebase conflict resolution. Weird that the build script did not complain about this...

brumm:
Looks like the cursor blink spec timed out, travis looks okay otherwise

simurai:
Tried it out and works!!! 👍 

> There are some wording problems in promptForRelaunch, since we're not actually relaunching Atom, for which I'll happily accept better phrasing.

Should the button say "Quit Atom"? The word "Relaunch" suggest that it will quit **AND** re-open Atom automatically.

title-bar

static/workspace-view.less
```diff
@@ -1,4 +1,19 @@
 @import "ui-variables";
+@import "octicon-mixins";
```
simurai: Do you mind removing all these additions in this file?

Sorry, there has been some major reshuffling of the core styles and all these styles are now in [scaffolding.less](https://github.com/atom/atom/blob/master/static/scaffolding.less)

 It's probably also part of the:

> seems like a bad rebase conflict resolution.

brumm: Ah yes, absoutely 👍

brumm:
>Should the button say "Quit Atom"? I feel like the word "Relaunch" suggests that it will quit AND re-open Atom automatically.

Thanks, I changed it to
```js
{
  message: "You will need to relaunch Atom for this change to take effect."
  buttons: ['Quit Atom', 'Cancel']
}
```

We'll be able to actually relaunch Atom [once we're using electron 1.2.2](https://github.com/electron/electron/pull/5837)

brumm:
Made sure that we set a sheet offset when using the custom titlebar:

image

controversial:
Is this almost ready to merge? If so, that'd be great; I can't wait to have this. 

I also think that `@tab-bar-background-color` looks better, fwiw, so I hope that will be configurable. 

Overall, great work 👍

brumm:
@simurai Hey, it'd be great if you could take a look at https://github.com/atom/atom/pull/11790/commits/3e53a03871ac25e68269f3897896fc93170ef471.

I decided to transition this instead of a hard `display: none` because the change happens after the fullscreen animation on OSX and is kind of abrupt.

brumm:
I found a weird bug: If `.tree-view-directory` is overflowing `.tree-view`, toggling treeview visibility causes `<atom-workspace>` to scroll down by at least the height occluded and at most the height of the titlebar, effectively hiding the titlebar.

As far as I can tell, this is not caused by programmatically setting `scrollTop` of `<atom-workspace>`.

Screencast: http://i.imgur.com/hG6Zl2Z.gifv

hultberg:
@brumm Can confirm this. I observed this behaviour when building atom with @paulcbetts PR on this same feature (#10208).

lee-dohm:
> Note: This currently does not include a file icon

This should read "Note: This currently does not include a proxy icon"

static/title-bar.less
```diff
@@ -1,15 +1,17 @@
...
 
 body.fullscreen .title-bar {
-  display: none;
+  margin-top: -@title-bar-height;
 }
 
 .title-bar {
-  height: 23px;
+  height: @title-bar-height;
+  transition: margin-top 200ms ease-out 700ms;
```
simurai: Ya, too bad the `.fullscreen` styles stay until after the "fullscreen animation" finishes.

How do you like just

```css
transition: margin-top 160ms;
```

A bit shorter, default easing and no delay. That way the animation maybe still feels part of the fullscreen animation and doesn't call for too much attention.

simurai:
> @brumm I found a weird bug

That feels like a deja-vu.. do you have the `color-picker` package installed? It looks like it's the [same issue](https://github.com/suda/tool-bar/issues/130) that `tool-bar` also has. A [fix got merged](https://github.com/thomaslindstrom/color-picker/pull/180), but I _think_ it didn't get published yet?

brumm:
Thanks @simurai! This was indeed the problem. I opened a new issue and asked @thomaslindstrom to release a new version which includes the fix.

thomaslindstrom:
@simurai @brumm Sorry, I thought I'd done this already. It's out in `2.2.0`!

static/title-bar.less
brumm: Yeah, I changed it as per your suggestion. `160ms` is a tad on the short side, but it helps selling the part-of-fullscreen-animation feeling like you said.

thomaslindstrom:
@simurai @brumm Just a quick update: `v2.2.0` including the fix for this issue broke `color-picker` completely. I've published `v2.2.1` that reverts the fix, but that means that this issue is still present.

src/main-process/atom-window.coffee
```diff
@@ -40,9 +40,15 @@ class AtomWindow
...
       options.icon = @constructor.iconPath
 
+    if @applyTitleBarSetting()
+      options.titleBarStyle = 'hidden'
+
     @browserWindow = new BrowserWindow options
     global.atomApplication.addWindow(this)
 
+    if @applyTitleBarSetting()
+      @browserWindow.setSheetOffset(23)
```
lee-dohm: This hard-coded value makes me nervous. If I change the height for the title bar, this will be off. Is there a way to calculate this and update it?

brumm: Hey, good call. I think we might be able to read `offsetHeight` after the title bar has been rendered to account for CSS-induced height changes

brumm: Fixed in https://github.com/atom/atom/pull/11790/commits/76357473ade7dadb6a2e318fb60ff48b704ece97

hultberg:
Whats the latest here?

onbjerg:
Looking forward to this

nathansobo:
Ugh, sorry, I was trying to push a branch based on this to to atom/atom and accidentally pushed to master. I'm making some adjustments and working toward merging this, but it's not there yet.

nathansobo:
✨  Thanks @brumm for putting this together. I'm already enjoying the more stylish title bar in my dev environment. This should go out with Atom 1.11.0-beta0 in the next release cycle. ⚡

onbjerg:
Now to wait for theme support once this is released 😁

silvestreh:
@onbjerg Material UI already supports it 😄 

captura de pantalla 2016-08-03 a las 1 21 04 a m

brumm:
Thanks you so much @nathansobo for picking this up! I'm stoked to see this getting merged 🎉

maxwellzdeamon:
I'm trying to implement this work in my beta (1.10.0-beta1).  When I add the below snippet to my style.less and save, restart atom, nothing happens. Is what I'm doing supposed to work?

Thanks!!

.theme-one-dark-ui {
  .tree-view::before,
  .tab-bar:before,
  .tab.active:before {
    display: none;
  }
}

nathansobo:
@maxwellzdeamon This won't land until 1.11.0, so you'll have to build from source until then if you want it.

FezVrasta:
@brumm is there any way to change the position of the window controls using CSS?

brumm:
@FezVrasta The window controls are drawn by the OS and positions can't be changed indivdually.
You might get some milage out of changing `titleBarStyle` to `hidden-inset` which moves the controls a bit further down, but also increases the size of the title bar.

I always figured that we'd have to completely hide the titlebar to support the 'tabs-on-top' style proposed in #10208, and dynamically switch draggable areas on and off with css.

That's what I came up with the other day:

screen shot 2016-07-11 at 18 12 38

FezVrasta:
Thanks for the reply. If it can be helpful I posted my custom CSS on https://github.com/atom/atom/pull/10208#issuecomment-241346442

paulcbetts:
@brumm We came up with that design too - it looks great but the problem is that objects in the non-client area (i.e. the invisible bar from the traffic lights onward) cannot be dragged, so reordering tabs is busted :cry:

controversial:
Couldn't we hide the title bar altogether, and create fake window buttons with CSS, as shown [here](http://kapetan.github.io/titlebar/demo/index.html)? @paulcbetts @brumm

silvestreh:
@controversial In Atom's case, window buttons are rendered by the OS and can't be changed.

controversial:
@silvestreh Why couldn't that be changed? This is a PR to the atom core...

silvestreh:
@controversial This is what the titlebar's DOM element looks like:

captura de pantalla 2016-09-02 a las 11 12 50 a m

As you can see, the buttons aren't part of the DOM. You can't style them with CSS because the window manager, not Chrome, is rendering them.

AFAIK, you can't move, resize, nor hide them.

controversial:
You can spawn a frameless window, which would hide the title bar completely, then you could *emulate* the buttons using custom CSS. https://github.com/electron/electron/blob/master/docs/api/frameless-window.md

Maybe this isn't available in the version of Electron atom is on though.

thomasjo:
> Maybe this isn't available in the version of Electron atom is on though.

This does indeed require a more modern Electron. We're very close to shipping #12300 though 🎉

controversial:
Ah, great. I'll look forward to #12300 😃

sveggiani:
@controversial I think it's not a good practice to replace native OS UI controls. They are a hell to maintain and avoids OS level customization and third party software control over them.

controversial:
Yeah, you're probably right. Waiting for the newer Electron version will be a much better choice.

keplersj:
I love this addition so much! I'm just trying it out right now but I've noticed a little bit of extra black under the bar with Nuclide. Not sure if this an Atom issue or a Nuclide issue.
screen shot 2016-10-10 at 12 17 00 pm
Not a huge issue, just thought it was worth noting.

Daniel15:
Any plans to have something like this for other operating systems, given [57% of the Atom userbase is on Linux or Windows](http://blog.atom.io/2016/05/06/two-years-open-source.html)? 😄 

> You can spawn a frameless window, which would hide the title bar completely, then you could emulate the buttons using custom CSS

I agree with @sveggiani... Emulating native OS controls is tricky and very difficult to maintain. It's better to use the native APIs to render the native controls (for example, I believe the title bar of Google Chrome does it this way). This used to be reasonably common with native apps on Windows, I don't know if Electron exposes this functionality though.

controversial:
@Daniel15 Electron does expose this, but only in a version of Electron that's later than the one Atom uses. I think a fake control would be good for the short term, using the native controls once we get to that version of Electron.

controversial:
Now that Atom is on a modern Electron version can we do this? I really like @brumm's [solution](https://github.com/atom/atom/pull/11790#issuecomment-241357545), even though the tabs can't be reordered. [Hyper](https://hyper.is) does this and the tabs aren't reorderable, but it never really bothers me. If this was an option that could be enabled by users, I think it'd be great to include some kind of solution in the Atom core.

lacostenycoder:
I'm probably late to the party, but is there any way I can do this currently?  Is there a dev branch or beta release with this enabled?

MaximSokolov:
@lacostenycoder, look for `Title Bar` option in latest stable Atom release

lacostenycoder:
@MaximSokolov, already found that thanks!

PR #11791, opened 2016-05-19
Sign Windows builds with Squirrel for MSI compatibility

damieng:
This switches us from signing directly with the Windows SDK to using Squirrel. This is necessary as Squirrel builds both the atomsetup.exe and the atomsetup.msi (which contains atomsetup.exe) so signing atomsetup.exe ourself is too late if we want to distribute the msi.

This is a little convoluted as we currently support remote URLs for keys to enable signing keys to not be checked in which Squirrel does not support so we continue to do that in our signing step but leave the actual signing to Squirrel.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,6 +33,8 @@ deploy: off
 artifacts:
   - path: out\**\AtomSetup.exe
     name: AtomSetup.exe
+  - path: out\**\AtomSetup.msi
+    name: AtomSetup.msi
 
 cache:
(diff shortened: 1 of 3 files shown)


damieng:
Testing

- [x] Local builds continue to sign atom, update, atomsetup.exe with local test key
- [x] Local builds continue to download and sign with WIN_P12KEY_URL
- [x] AppVeyor builds continue to build and publish atomsetup.msi
- [x] MSI now contains signed atomsetup.exe
- [ ] Official builds continue to sign as before

PR #11792, opened 2016-05-19
replace _.extend in favor of Object.assign

Connormiha:
Since Chromium 45 supports native `Object.assign` for merging objects. Let's get rid `extends` from `underscore` in favor of native method.

### DIFF
--- a/src/browser/atom-application.coffee
+++ b/src/browser/atom-application.coffee
@@ -25,7 +25,7 @@ LocationSuffixRegExp = /(:\d+)(:\d+)?$/
 #
 module.exports =
 class AtomApplication
-  _.extend @prototype, EventEmitter.prototype
+  Object.assign @prototype, EventEmitter.prototype
 
   # Public: The entry point into the Atom application.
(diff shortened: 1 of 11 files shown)


YuriSolovyov:
I guess you might need [`Object.setPrototypeOf`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) for doing inheritance

Connormiha:
> I guess you might need Object.setPrototypeOf for doing inheritance

`Object.assign` and `_.extends` for merge objects (not inheritance) and the are equals.
`Object.setPrototypeOf`does something quite different.

YuriSolovyov:
sorry, got confused with `@prototype`

lee-dohm:
What benefit does this give?

Connormiha:
> What benefit does this give?

Just a little less depending of extend library. Why use extend function if browser has equal function?

lee-dohm:
Thanks for contributing!

PR #11803, opened 2016-05-20
Following replaceSelectedText(), update the selected text range accordingly.

lexicalunit:
Following a call to `replaceSelectedText()`, the previous selected text range is not necessarily still relevant as that text has been replaced. The newly inserted text can possibly have a totally different range. Use the range of the newly inserted text as the final selection range.


lexicalunit:
I completely forgot about this PR after working around the odd behavior of `replaceSelectedText()`. In [`pretty-json`](https://github.com/federomero/pretty-json/blob/master/index.coffee#L88). With the latest update to this PR I've finally made changes to the spec tests so they pass as well as added new tests to verify the corrected behavior. My apologies for letting this languish for as long as I did.

iolsen:
Great fix, thanks!

Issue #11782, opened 2016-05-18
Unable to disable 80 char guide


jglovier:
### Prerequisites

* [ ] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
  - The `atom` command and `apm` doesn't work on my new machine.
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
  - See note below
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

I'm getting Atom setup on a new machine. I'm trying to get rid of the 80 character guide, but cannot seem to find a way to disable it. 

I thought it used to be in Settings as a checkbox with a value you could adjust for the length, but either my memory is incorrect, or it has been changed. I also checked under the View menu, but there is no toggle for it there, either.

@shiftkey pointed out that on his machine setting `Preferred Line Length` to `0` seems to be doing it, but for me that doesn't actually work as setting a `0` value results in Atom automatically changing the value to `1`, which just overlays the line at the 1 character column.

Even more strange about the previous point is that I don't even have `Soft Wrap At Preferred Line Length` checked, so if anything I would assume that `Preferred Line Length` should not be in effect until this is checked.

### Steps to Reproduce

1. Download Atom from Atom.io on a new machine.
2. Open a file.
3. Attempt to turn off the 80 character guide.

**Expected behavior:** I expected there to be a checkbox in `Settings` or a toggle option in the `View` menu to allow me to turn the guide on or off.

[...]

PR #11804, opened 2016-05-21
Add a note how to hide the wrap-guide

simurai:
Ideally this would be an option in the settings, but maybe a note will already help too.

screen shot 2016-05-21 at 10 27 45 am

Refs: https://github.com/atom/settings-view/issues/790 + https://github.com/atom/atom/issues/11782. There are many more issues and discussions about this, but these are just the recent two.

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -197,7 +197,7 @@ module.exports =
       softWrapAtPreferredLineLength:
         type: 'boolean'
         default: false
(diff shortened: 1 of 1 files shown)


TimvdLippe:
There is a typo. "I" should be "If". Change it to "If you would like to"?

jglovier:
Yay!! :bow: :tada:

lee-dohm:
Made a little change to bold the note and make the wording more clear. 👍 from me if you want to merge this @simurai :grinning:

simurai:
@lee-dohm Thanks!

Ok, the build fails, but (to me) looks like unrelated timeouts, so 🚢

PR #11820, opened 2016-05-23
Update ohnogit@0.0.12

joshaber:
Hopefully fix the crash in https://github.com/atom/atom/issues/11708.


joshaber:
Looks like that test has been racy as hell before this too.

Issue #3158, opened 2014-08-02
Atom erasing the content of files when there is not enough disk space


xaviervia:
I have a hard time replicating this issue since it will require me filling up the disk again, but in a nutshell the problem is that when I edit a file in Atom, if the disk space is not enough to save it, I find that the file content is erased. 

This happened to me more than once (yeah I need more Gbs of disk space) so I can certify that it is Atom the one causing the issue. The issue appears whenever I edit the file, not when I just open it. The problem is not that Atom is not being able to save the file (that would make sense if there is no disk space available), the problem is that Atom _removes all the content_, and of course I'm afterwards not able to recover it.

I have no clue of why may this be happening

jonas-schievink:
The same thing happens for me, and I still have more than 80 GB free (I'm on Windows). Why does Atom not tell me that something went wrong? Why doesn't it create backups before saving? Not cool!

Zireael07:
An error message is definitely needed in such a case.

forivall:
Yeah, this happened to me and it wiped out my config file, leading to me being unable to open atom due to conflicting packages that were disabled.

alexandernst:
Not really sure why such a ridiculos and dangerous bug hasn't been fixed already. This should be nominated as "stop-everything-and-fix-this-right-now" instead of "1.0-roadmap" imho.

xaviervia:
Right. I think there has a been a weird distortion in which the developers do not run at all into this issue and they are overlooking the fact that people with passable equipment are running into this once and again. Now, the issue with solving this is that it touches a very core part of how Atom works, I don't think anybody is ready to just contribute with a quick fix considering that it has architecture implications.

It would be nice to hear more thoughts from the core team though. If writing contents to a temporary file and then renaming is considered a reasonable idea then it is possible to fix it soon.

50Wliu:
This is already on the 1.0 roadmap, so it'll most likely get fixed in the coming weeks/months.

bronson:
Hoping for 1.0.1...?  I'm definitely concerned that Atom is writing important data without using the [save-to-new-file-and-rename](http://stackoverflow.com/questions/1812115/how-to-safely-write-to-a-file) trick.

nathansobo:
[...]

PR #11828, opened 2016-05-24
File Recovery Service

### LINKED ISSUE #11708: Atom 1.7.3 crashes on OSX local network drive while saving scss files, erases file
Howdy folks, 

While working on an scss file on our company's local network drive, Atom will crash when using cmd + s to save. This issue is a few versions old, but I can always reopen and access the file I was working on and start over from the last save. Now, for the third time, Atom has entirely deleted the file I was working on. The file on the network drive is now blank. Zero bytes. I am on the latest version. I have run it in safe mode. I have uninstalled and reinstalled. 

Anyone else having trouble?

as-cii:
~~Depends on #11826 (to run [spec/browser/file-recovery-service.spec.js](https://github.com/atom/atom/blob/as-file-recovery-service/spec/browser/file-recovery-service.spec.js)).~~

Fixes #11708.

Recently, we have been observing several issues related to losing data when Atom hard-crashes. We think that this problem manifests whenever the following happens:
1. User/Package triggers a file save.
2. `fs.writeFileSync(path)` gets called.
   1. `fd = fs.openSync(path, "w")` gets called. This causes the file to be truncated.
   2. Before `fs.writeSync(fd, contents)` has the chance to run, a hard crash occurs.

The scenario described above leaves the file in a "truncated state", thus erasing all of its contents. In the past we have experimented with backup files to solve this issue, but that creates other problems such as paths that are too long on Windows, and unnecessary events triggered on fs watchers that are monitoring the saved file's directory. It turns out that atomic saves are not a good trade-off either, as described on https://github.com/atom/atom/issues/3158#issuecomment-124225148.

With this pull-request we want to provide a file recovery service running on the main process that keeps track of pending changes, and restores files to their original state when a renderer process crashes while a save is in flight. If the recovery is unsuccessful for any reason (e.g. problems with permissions, unmounted network drives, etc.), we log a message like the following in the system console and in a message box too, so that users can perform the recovery themselves:

```
There was a crash while saving "/private/etc/profile", so this file might be blank or corrupted.
Atom couldn't recover it automatically, but a recovery file has been saved at: "/Users/as-cii/.atom/recovery/profile-950635".
```

By default we will use `~/.atom/recovery` as our recovery directory, where we temporarily store a backup of the saved file and delete it upon successful save/recovery. The nice thing about using the file system as our recovery storage (as opposed to IndexedDb), is that users can inspect it pretty easily and copy/paste content via Terminal.app or Finder.app (or their equivalent on other platforms). 

As a convention, backup files stored under the recovery directory, will have the following format:

```
{originalFilename}-{randomHexNumber}.{originalExtension}
```

Where:
- `originalFilename` is the name of the file without its extension. This string will be truncated if it's longer than `34` characters.
- `randomHexNumber` is a randomly generated hexadecimal number. This string will be 6 characters long by default.
- `originalExtension` is the extension of the file.

This should be long enough to make filenames readable (and more inspectable), but short enough to avoid path length issues on platforms like Windows.


src/project.coffee
```diff
@@ -360,7 +360,6 @@ class Project extends Model
 
   addBuffer: (buffer, options={}) ->
     @addBufferAtIndex(buffer, @buffers.length, options)
-    @subscribeToBuffer(buffer)
```
as-cii: Apparently, we were subscribing to buffers twice, once here and once in `addBufferAtIndex`.

src/browser/atom-window.coffee
```diff
@@ -141,14 +141,15 @@ class AtomWindow
     @browserWindow.webContents.on 'crashed', =>
       global.atomApplication.exit(100) if @headless
 
-      chosen = dialog.showMessageBox @browserWindow,
+      dialog.showMessageBox @browserWindow,
```
as-cii: As described in de599f9c66b1838540cfd263754225908a237048, we need to make this dialog asynchronous so that other `crashed` event handlers (i.e. the recovery service ones) have the chance to run.

BinaryMuse:
This is lookin' great, @as-cii, thanks very much for taking this on.

I have a couple questions / comments:

1. Is there any chance that the timing won't work out in some cases? I'm imagining the following scenario:
  1. `onWillSave` calls `emitWillSavePath`
  2. `emitWillSavePath` emits `will-save-path`. This is *asynchronous*.
  3. The renderer process beings saving, truncating the file
  4. The main process receives the `will-save-path` event. It's too late to back up the file

  If so, perhaps this is one of the few occasions where [`sendSync`](http://electron.atom.io/docs/api/ipc-renderer/#ipcrenderersendsyncchannel-arg1-arg2-) would be appropriate? 

2. This has all got me wondering if this service is actually necessary at all... would it be reasonable to just ask the main process to do all file saving? The renderer process could just ship it the contents to save, either via IPC or some other mechanism (temp file?) and then let the main process handle the saving. I'm not actually sure if this is better or not, or even reasonable, but it's a thought. :)

src/browser/file-recovery-service.js
```diff
@@ -0,0 +1,65 @@
...
+    const recoveryPath = Path.join(this.recoveryDirectory, recoveryFileName)
+    fs.writeFileSync(recoveryPath, fs.readFileSync(path))
+
+    if (!this.recoveryPathsByWindowAndFilePath.has(window)) {
+      this.recoveryPathsByWindowAndFilePath.set(window, new Map())
+    }
+    this.recoveryPathsByWindowAndFilePath.get(window).set(path, recoveryPath)
+
+    if (!this.crashListeners.has(window)) {
+      window.on('crashed', () => this.recoverFilesForWindow(window))
```
kuychaco: Do you think we should remove the listener when the window closes? Would the callback's reference to `window` retain it and keep it from being removed from the WeakSet?

src/browser/file-recovery-service.js
```diff
@@ -0,0 +1,65 @@
...
+export default class FileRecoveryService {
+  constructor (recoveryDirectory) {
+    this.recoveryDirectory = recoveryDirectory
+    this.recoveryPathsByWindowAndFilePath = new WeakMap()
+    this.crashListeners = new WeakSet()
+  }
+
+  start () {
+    ipcMain.on('will-save-path', this.willSavePath.bind(this))
+    ipcMain.on('did-save-path', this.didSavePath.bind(this))
```
kuychaco: Should we keep a `CompositeDisposable` for removing these listeners? Might not be a big deal since everything will be torn down when the main process is killed, but might be good to clean up for testing purposes?

src/browser/file-recovery-service.js
```diff
@@ -0,0 +1,65 @@
...
+  willSavePath (event, path) {
+    if (!fs.existsSync(path)) {
+      // Unexisting files won't be truncated/overwritten, and so there's no data to be lost.
+      return
+    }
+
+    const window = event.sender
+    const recoveryFileName = crypto.randomBytes(5).toString('hex')
+    const recoveryPath = Path.join(this.recoveryDirectory, recoveryFileName)
+    fs.writeFileSync(recoveryPath, fs.readFileSync(path))
```
BinaryMuse: Should we ensure this is written by flushing?

src/browser/file-recovery-service.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  recoverFilesForWindow (window) {
+    const recoveryPathsByFilePath = this.recoveryPathsByWindowAndFilePath.get(window)
+    for (let [filePath, recoveryPath] of recoveryPathsByFilePath) {
+      try {
+        fs.writeFileSync(filePath, fs.readFileSync(recoveryPath))
+        fs.unlinkSync(recoveryPath)
+      } catch (error) {
+        console.log(`Cannot recover ${filePath}. A recovery file has been saved here: ${recoveryPath}.`)
```
BinaryMuse: Overall I think this PR is a great first step and we shouldn't get too complicated, but I'm wondering if it's worth surfacing this message in a more obvious way to users.

BinaryMuse:
> This has all got me wondering if this service is actually necessary at all... would it be reasonable to just ask the main process to do all file saving?

Now that I think of it, @as-cii didn't you experiment with saving via a different process and it was too slow? The suggested process where the renderer process creates a temp file somewhere containing the data to be saved and the main process takes that and copies it into the real save location isn't that different from what we have here... except maybe we don't have to worry as much about crashes? What do you think?

src/browser/file-recovery-service.js
as-cii: I am not sure about flushing, as I don't think it's enough to ensure the file was persisted. From `man 2 fsync`:

```
     Specifically, if the drive loses power or the OS crashes, the application may find that only some or none of their data was writ-
     ten.  The disk drive may also re-order the data so that later writes may be present, while earlier writes are not.

     This is not a theoretical edge case.  This scenario is easily reproduced with real world workloads and drive power failures.

     For applications that require tighter guarantees about the integrity of their data, Mac OS X provides the F_FULLFSYNC fcntl.  The
     F_FULLFSYNC fcntl asks the drive to flush all buffered data to permanent storage.  Applications, such as databases, that require a
     strict ordering of writes should use F_FULLFSYNC to ensure that their data is written in the order they expect.  Please see
     fcntl(2) for more detail.
```

So we might use `fcntl` but that wouldn't be cross-platform and I am not sure that investing time in flushing data to disk is actually worth it. What I like about this solution is that, worst-case scenario, users can find the contents of their file under `~/.atom/recovery` after rebooting their machine when e.g. the system crashes.

as-cii:
> This has all got me wondering if this service is actually necessary at all... would it be reasonable to just ask the main process to do all file saving?

What makes me nervous about this solution is handing possibly huge chunks of memory to another process through ipc. Maybe this isn't a huge deal in practice, but we ran into some troubles when sending lots of contents down an ipc channel in the past (for `spell-check`), so it felt more natural to keep the communication between the two processes as lightweight as possible.

> Now that I think of it, @as-cii didn't you experiment with saving via a different process and it was too slow?

Yeah, in that case we also had the huge overhead of spawning a new process, whereas here we have the main process already running, which makes it an (almost) zero cost facility that we can control through ipc.

> Is there any chance that the timing won't work out in some cases? I'm imagining the following scenario:

Good call, @BinaryMuse! 👍

src/browser/file-recovery-service.js
as-cii: Yeah, we aren't exercising `start` in our specs and, as you mentioned, the main process will tear everything down when killed, so I figured it wasn't a huge deal to make those handlers disposable. Thoughts? 💭

src/browser/file-recovery-service.js
as-cii: That's a great point, thanks! 💯

src/browser/file-recovery-service.js
kuychaco: Ah yes, makes sense 👍

src/browser/file-recovery-service.js
BinaryMuse: Okay, that sounds reasonable to me. Would it be worth using the original file basename as part of the filename so that they can be more easily identified?

src/browser/file-recovery-service.js
as-cii: Maybe we can add a message box too? I'll experiment with this. 👍

src/browser/file-recovery-service.js
as-cii: > Would it be worth using the original file basename as part of the filename so that they can be more easily identified?

My main concern there is path length on Windows, which is why I went with a 10-letter random filename for these recovery files. Introducing a variable-length file name could cause weird edge cases on Windows, although it's admittedly more user-friendly. Maybe we can truncate the filename at some arbitrary length, and add a random, fixed-size suffix string? I'll try that shortly.

spec/main-process/file-recovery-service.spec.js
```diff
@@ -0,0 +1,148 @@
...
+import FileRecoveryService from '../../src/main-process/file-recovery-service'
+import temp from 'temp'
+import fs from 'fs-plus'
+import {Emitter} from 'event-kit'
+import sinon from 'sinon'
+
+describe("FileRecoveryService", () => {
+  let recoveryService, recoveryDirectory, windows, previousConsoleLog
+
+  function createWindow () {
```
as-cii: These could be extracted in a `spec-helper` probably, but I'd like to wait for more usages before doing so.

src/main-process/file-recovery-service.js
```diff
@@ -0,0 +1,119 @@
...
+    this.recoveryPath = recoveryPath
+    this.refCount = 0
+  }
+
+  storeSync () {
+    fs.writeFileSync(this.recoveryPath, fs.readFileSync(this.originalPath))
+  }
+
+  recoverSync () {
+    fs.writeFileSync(this.originalPath, fs.readFileSync(this.recoveryPath))
```
BinaryMuse: If the files are large, will `storeSync` and `recoverSync` greatly increase memory usage? I wonder if reading some number of bytes at a time, or using some synchronous version of `pipe`, would be better?

BinaryMuse:
@as-cii Do you have a testing strategy I can use to test this on Windows? I'm not 100% sure how to reproduce artificially.

/cc @lee-dohm

as-cii:
🚀 

Thanks everyone for the great feedback! ❤️💚💙💜

Issue #8237, opened 2015-08-05
Ordering of styles makes some button Bootstrap styles unusable


ssorallen:
'static/bootstrap' is always included before 'static/atom', which means Atom's styles win on document order precedence. Atom's [buttons.less](https://github.com/atom/atom/blob/master/static/buttons.less) re-declares the `.btn` class, which means its styles have the same specificity as Bootstrap's button "sub"classes like `.btn-link`.

`<button class="btn btn-link"></button>` inside Atom looks like a regular button because Atom's `.btn` declares background color and border, overriding the `.btn-link` declaration.

Is it feasible to import Bootstrap as less and set Less variables for Bootstrap rather than re-declaring the class names? Bootstrap exposes dozens of variables for customization: http://getbootstrap.com/customize/

benogle:
Ideally, we'd like to get rid of bootstrap at some point. cc @simurai

mnquintana:
Refs https://github.com/atom/atom/issues/6351

IMO, if we are going to phase out Bootstrap, we should do it piece-by-piece (and honestly, we may end up deciding to keep Tooltips, if nothing else - that component is rather handy).

ssorallen:
I'm not sold that Bootstrap was a bad idea, but the way customization was done (re-declaring classes from Bootstrap) makes it difficult to know which pieces of Bootstrap actually work as expected inside Atom.

benogle:
> IMO, if we are going to phase out Bootstrap, we should do it piece-by-piece

I agree with this totally.

> we may end up deciding to keep Tooltips, if nothing else

We would write our own implementation. They are currently kind of a jank part of the system, and it's tough to get fixes in if we need them.

simurai:
The plan is to "merge" `bootstrap/less/buttons.less` into `atom/static/buttons.less`. Then there is less overriding needed and should simplify things a bit. Same with the rest of the components.

> we may end up deciding to keep Tooltips

Not sure how to organize components that need JS. Maybe `atom/static/js/tooltip.js`. Or  `atom/static/tooltip/` that would contain CSS __and__ JS.

And then maybe move the whole `atom/atom/static/` into an own package like `atom/ui-kit` or so.

simurai:
The button styles got moved to https://github.com/atom/atom-ui a while ago. Please holler if this is still an issue.

Issue #6351, opened 2015-04-12
Replace Bootstrap CSS with hand-written CSS designed for Atom


bolinfest:
The bootstrap CSS is enormous. It is designed to solve a problem that Atom does not have. That is, a set of CSS rules to normalize presentation across all browsers that may be used for rendering, using whatever browser-specific hacks necessary to achieve that goal.

The project was also conceived before the advent of flexbox, which means that bootstrap has a ton of cruft to deal with pre-flexbox browsers. Such workarounds are verbose and make it difficult to understand what can be removed.

In sum, this puts Atom into a situation where:
* The CSS is much larger than what Atom needs.
* It requires much more work than should be necessary to understand why the current CSS works the way it does.
* Bootstrap is the source of truth, rather than something that is directly under Atom's control.

For example, take a look at the issue I recently filed against Bootstrap: https://github.com/twbs/bootstrap/issues/16276. The recommended solution is to pile on another hack (i.e., use an even more specific CSS selector) rather than figure out why the existing CSS works the way it does and fix the underlying problem with the model. Presumably this is because trying to untangle why things work the way they do is too hard.

Atom should not be saddled with the baggage of layout engines of old and should be exploiting the fact that it only needs to support Blink (which supports flexbox), and write its own small, specific, and far more maintainable CSS.

thomasjo:
I'm fully behind this, and I vote for getting this on the 1.0 roadmap, if possible. That might mean us folks from outside the core team might need to really pitch in on this, hopefully with @simurai's help.

mnquintana:
I totally support this too, and I'd be willing to help with this in any way I can.

mdo:
Bootstrap's CSS is not enormous, but I imagine Atom could be using less than what all of Bootstrap includes. I do agree though that there is a mis-match between much of what Bootstrap offers—cross browser normalization, components, etc—and what Atom needs given it's built only for Chromium.

> For example, take a look at the issue I recently filed against Bootstrap: twbs/bootstrap#16276. The recommended solution is to pile on another hack (i.e., use an even more specific CSS selector) rather than figure out why the existing CSS works the way it does and fix the underlying problem with the model. Presumably this is because trying to untangle why things work the way they do is too hard.

[...]

PR #11839, opened 2016-05-25
Move Bootstrap into Core

simurai:
⚠️  This is in parallel to #11858 + 11875. Don't merge all PRs, only pick one. ⚠️ 

This moves most of the Bootstrap styles into Core. The styles are based on Bootstrap `3.3.6`. It probably will be the last Bootstrap version that supports Less.

Refs: #6351, #8237 
### Moved + merged into Core
- [x] Tooltips
- [x] Buttons + Button Groups
- [x] Lists
- [x] Code
- [x] Nav
- [x] Scaffolding
- [x] Utilities
- [x] Alerts
- [x] Tables
- [x] Forms
- [x] Type
- [x] Normalize
### Removed
- 🔥  grid
- 🔥  input-groups
- 🔥  labels
- 🔥  thumbnails
- 🔥  close
## Risks

Since some styles got removed, there is a chance that something might look unstyled or off. Packages bundled with Atom should be fine, but it's hard to predict if community packages rely on any of the removed styles.

/cc @atom/design @atom/feedback

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -235,8 +235,10 @@ module.exports = (grunt) ->
         'outline-none': false
         'overqualified-elements': false
         'qualified-headings': false
+        'regex-selectors': false
         'unique-headings': false
         'universal-selector': false
+        'unqualified-attributes': false
         'vendor-prefix': false
       src: [
         'static/**/*.css'
--- a/static/alerts.less
+++ b/static/alerts.less
@@ -1,0 +1,114 @@
+@import "variables/variables";
+@import "ui-variables";
+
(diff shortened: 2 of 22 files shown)


simurai:
FYI: This PR relaxes the Less linting rules: `unqualified-attributes` and `regex-selectors`

1. Modern browsers became much faster in selector performance. https://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles/#h-H1_11
2. We have been using the Bootstrap styles (with "slow" selectors) without linting them too.

A general CSS performance audit with **all** styles might be worth to look into.

PR #11852, opened 2016-05-26
Option for avoiding overlay movements

MikeInnes:
Right now overlays will often move up and down as you scroll, attempting to stay within the editor view. This makes sense for autocomplete, but less sense for things like annotations in code (see e.g. Ink, Hydrogen packages). Indeed in these cases the jumpiness feels very unstable.

So I added a simple option which turns it off. The name could probably do with refinement.

Edit: [more helpful diff](https://github.com/atom/atom/pull/11852/files?w=0). Also, same testing concern as #11864.


OmarShehata:
+1! I'm working on a tool to do some code analysis and annotation, definitely need this.

lee-dohm:
We would need some specs to ensure we don't accidentally break this in the future too.

MikeInnes:
Understood – do you have any comments on the testing concerns I mentioned in the linked issue? In particular,

> I can't seem to find current tests and it seems like being headless causes difficulties for testing things like positioning.

Advice on what would make for an acceptable test would be much appreciated.

Haacked:
I'm not sure there's a good way to test this. I'll follow up on that.

In the meanwhile, there's a legitimate build failure here.

> src/text-editor-presenter.coffee
>    line 478  Don't use &&, ||, ==, !=, or !  Replace "!" with "not"

Looks like an easy fix to at least get the build passing. :smile:

MikeInnes:
Ah yes, missed that – fixed!

nathansobo:
Maybe we should  phrase the naming in the inverse. Maybe an option like `avoidOverflow` which defaults to `true` but you can assign to `false`? I'd love to hear some other name ideas if you don't like `avoidOverflow`... that's just off the top of my head.

Haacked:
@MikeInnes regarding the testing question. I don't think we have any good way right now to write an automated test that exercises the UI and ensures that overlays stay in the same position as you scroll.

But there's probably a way to write a unit test to test this code in isolation such that the test fails when you revert your changes, but succeeds with your changes. Does that make sense?

OmarShehata:
Any news on this being merged?

MikeInnes:
Apologies for letting this slide for a while. @Haacked Good news is I found the overlay positioning tests (perhaps they are new or I just missed them), so the tests I've just added are checking this really works :)

@nathansobo I'm not wildly opinionated about the name so will be happy to change to whatever the consensus is. `avoidOverflow` seems sensible – perhaps `fixed` or `stayVisible` as alternatives? Aside, it may be worth considering making this mode the default, assuming we can figure out an acceptable approach to backwards compatibility.

nathansobo:
Taking a look at the failing tests.

nathansobo:
@MikeInnes can you give me push access to this branch?

MikeInnes:
Thanks @nathansobo, I've added you as a collaborator on my fork of atom.

nathansobo:
Thanks for the add. I also just added some documentation for the new option. For future reference, it's now possible to allow repository owners to make changes just to your pull request branch when you open the PR. But being a full collaborator works as well. Thanks for taking the initiative with this change. Hopefully we get a green build.

nathansobo:
Figured out a CI failure caused by something on our end. Running another build now and will merge once it's green.

nathansobo:
Thanks for this contribution @MikeInnes! ⚡️

MikeInnes:
Awesome, thanks for the help @nathansobo!

PR #11858, opened 2016-05-28
Move Bootstrap to a separate package

simurai:
⚠️  This is in parallel to #11839 + 11875. Don't merge all PRs, only pick one. ⚠️ 

This moves Bootstrap (and some core components) to a separate package. 

Refs: #6351, #8237 
### Stays in core `atom/atom/static`:
- [x] Octicons
- [x] Normalize
- [x] Scaffolding
- [x] Cursors
- [x] Panels
- [x] Panes
- [x] Syntax
- [x] `text-editor-light`
- [x] `workspace-view`
### Moved + merged into the [atom/ui](https://github.com/atom/ui) package:
- [x] Tooltips
- [x] Buttons + Button Groups
- [x] Lists
- [x] Code
- [x] Nav
- [x] Scaffolding
- [x] Utilities
- [x] Alerts
- [x] Tables
- [x] Forms
- [x] Type
- [x] Normalize
### Removed:
- 🔥  grid
- 🔥  input-groups
- 🔥  labels
- 🔥  thumbnails
- 🔥  close
## Motivation

Currently we load Bootstrap and then try to override it in core. Moving some of it into its own package allows more control and adjustments can be made directly. Also, making changes in a separate package feels less intimidating and can easier be rolled back in `package.json`.
## Concerns

What happens if you disable the UI package?

ui

Because the main layout styles are still in core and themes override a lot of the components, it doesn't look too broken. Just a few things seem off or un-styled. For the tree-view or nav in the settings, we could remove the list styles in those packages to make a bit more robust. Being able to disable the UI package might inspire the community to create their own version or add additional components.

Some styles got removed, so there is a chance that something might look unstyled or off. Packages bundled with Atom should be fine, but it's hard to predict if community packages rely on any of the removed styles. Maybe in the Beta phase we can catch some regressions.

@atom/design @atom/core


thomasjo:
@simurai Why is this being used as an apm package? Wouldn't it make a lot more sense for this to be an npm package that we simply add a dependency to? That avoids a LOT of problems, in particular the package being disabled like you've discovered.

thomasjo:
@simurai Let me know if you need any assistance btw. After tomorrow I should have a some spare time on my hands again, so I can pitch in wherever it is needed 🎉

simurai:
@thomasjo Ohh.. you're right. npm sounds like the much better option. We should look into that.

PR #11864, opened 2016-05-30
Fix overlay scrolling

MikeInnes:
Right now editor scrolling is not sensitive to overlays that are present, so that a line annotation at the end of a long line – or at the end of a document, if scrollPastEnd is not enabled – may be occluded by the edge of the window. This patch simply takes into account the overlay size when setting the scroll bounds so that these overlays will always be visible.

What's the best way to test things like scrolling and positioning of overlays? I can't seem to find current tests and it seems like being headless causes difficulties for testing things like positioning.

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -46,6 +46,8 @@ class TextEditorPresenter
 
     @scrollPastEndOverride = scrollPastEnd ? true
 
+    @overlayBounds = {bottom: 0, right: 0}
+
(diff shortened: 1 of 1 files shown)


lionkingsound:
Aoto

sadick254:
Hey @MikeInnes, Apologies we didn't get to review this ASAP.  This PR has been here for quite a long time. In that time Atom has seen lots of changes. This PR is now out of date. I am closing this for now. I am happy to reopen this PR when the PR is updated using the current state of Atom.

PR #11866, opened 2016-05-30
Fix flexbox issues after upgrading to Electron v0.37.8

### LINKED ISSUE #732: find-and-replace width does not update when window size is decreased
captura de pantalla de 2016-05-25 13-19-09

## Repro Steps

1. Right-click on any folder in the tree view.
2. Press *Search in Directory*.
3. Increase the text editor panel width.
4. Decrease it.

**Expected:** Find & replace panel matches the text editor panel's width.
**Actual:** Find & replace panel overflows itself and the above panel.

It happens with `atom --safe` too.

## Versions

* **Atom:**       1.7.3
* **Atom-Shell:** 
* **OS:**         linux 4.4.9-300.fc23.x86_64
* **Misc**
    * apm  1.10.0
    * npm  1.3.6
    * node 0.10.42
    * python 2.7.11
    * git 2.5.5

---

<small>This report was created in and posted from the Atom editor using the package `bug-report` v0.7.1.</small>

as-cii:
1) Fixes atom/find-and-replace#732.
2) Fixes an issue with resizing atom panes when one or more panes had many open tabs.

After [upgrading to Electron v0.37.8](https://github.com/atom/atom/pull/11474), we have noticed a couple of regressions when resizing some of Atom's UI elements. In particular, it seems like some of the flexbox rules for laying content out have changed since v0.36.12, and this causes elements like `atom-pane` or find-and-replace input boxes, to not shrink correctly when they are resized.

This pull-request fixes 1) by removing the explicit width on `atom-text-editor`'s `.lines` element (setting it should be unnecessary after #6733), which sometimes caused `atom-text-editor`'s width to determine the width of its parent, and 2) by allowing `atom-pane`s to be shrunk indefinitely, as it was before the Electron bump.

### DIFF
--- a/spec/pane-element-spec.coffee
+++ b/spec/pane-element-spec.coffee
@@ -1,12 +1,13 @@
 PaneContainer = require '../src/pane-container'
 
 describe "PaneElement", ->
-  [paneElement, container, pane] = []
+  [paneElement, container, containerElement, pane] = []
 
   beforeEach ->
     spyOn(atom.applicationDelegate, "open")
 
     container = new PaneContainer(config: atom.config, confirm: atom.confirm.bind(atom))
+    containerElement = atom.views.getView(container)
     pane = container.getActivePane()
     paneElement = atom.views.getView(pane)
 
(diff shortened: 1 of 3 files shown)


as-cii:
Thanks @simurai for the ping on 1) and thanks @BinaryMuse and @kuychaco for spotting 2). ✨ I'll go ahead and 🚢 this for now, let me know if you notice other odd behaviors related to resize/flexbox. 🙇

/cc: @atom/core

simurai:
Awesome.. Yes, works. 👍 🎉

kuychaco:
✨  thanks for fixing this @as-cii! was driving me crazy >_<

PR #11875, opened 2016-06-01
Merge Bootstrap + Core styles

simurai:
⚠️  This is in parallel to #11839 + #11858. Don't merge all PRs, only pick one. ⚠️ 

This PR does:
1. Remove `bootstrap` dependency
2. Merge Bootstrap `3.3.6` + some Core styles and publish it as [atom-ui](https://www.npmjs.com/package/atom-ui) module
3. Add `atom-ui` dependency

Refs: #6351, #8237 
#### Stays in core `atom/atom/static`:
- [x] Octicons
- [x] Normalize
- [x] Scaffolding
- [x] Cursors
- [x] Panels
- [x] Panes
- [x] Syntax
- [x] `text-editor-light`
- [x] `workspace-view`
#### Moved + merged into the [atom/atom-ui](https://github.com/atom/atom-ui) package:
- [x] Tooltips
- [x] Buttons + Button Groups
- [x] Lists
- [x] Close
- [x] Code
- [x] Nav
- [x] Scaffolding
- [x] Utilities
- [x] Alerts
- [x] Tables
- [x] Forms
- [x] Type
- [x] Normalize
#### Removed:
- 🔥  grid
- 🔥  input-groups
- 🔥  labels
- 🔥  thumbnails
## Motivation

Currently we load Bootstrap and then try to override it in core. Merging them allows more control and adjustments can be made directly. Also, moving out of core and into a separate package feels less intimidating and can easier be rolled back in `package.json`.
## Concerns

Some styles got removed, so there is a chance that something might look unstyled or off. Packages bundled with Atom should be fine, but it's hard to predict if community packages rely on any of the removed styles. Maybe in the Beta phase we can catch some regressions.

Affected packages:
- [x] `autocomplete-plus` fixed by [reverting](https://github.com/atom/atom-ui/pull/4) some refactoring.
- [x] [godoc](https://github.com/zmb3/godoc/pull/17) uses `.close`. Fixed by [adding](https://github.com/atom/atom-ui/pull/5) `.close` too.


simurai:
@thomasjo Did I get this npm thing right?

I had to import the atom-ui files in [atom.less](https://github.com/atom/atom/blob/sm-move-bootstrap-to-npm/static/atom.less#L27-L48) (instead of `bootstrap.less`), since atom-ui uses some of the `ui-variables`. Maybe ok.

simurai:
One downside I found is that the "live-reload" doesn't work and you have to manually reload when making changes to the `atom-ui` package. Maybe ok and still easier than building Atom.

simurai:
Although the build fails (I think unrelated), this branch would be ready for a test drive. Things to test: 

- if there are no errors when building Atom
- if the UI looks still the same and nothing is unstyled, broken or off in some ways.

/cc @atom/feedback

simurai:
Ouch.. apparently [adding some padding to highlights](https://github.com/atom/atom-ui/commit/ba923ffb5770ef769958590b96136cdc6970c249#diff-a0950c1c432fb6733f259e3fd811501dR182) causes [this find-and-replace spec](https://github.com/atom/find-and-replace/blob/master/spec/results-view-spec.coffee#L289) to fail. :see_no_evil:

thomasjo:
@simurai Once `bootstrap.less` is removed, this looks good to me! ✨ I'll keep testing for a day or two ✌️

thomasjo:
Been using this for a few days now, and haven't noticed any regressions. Let's 🔥 `bootstrap.less` and 🚢  this?

/cc @atom/feedback

simurai:
> Let's 🔥  `bootstrap.less`

I guess it's this: https://github.com/atom/atom/blob/f48069c5f848b5ae624b110a1df7800e919eb142/src/theme-manager.coffee#L173?

Also, searched for `bootstrap` and found these places that probably should be cleaned up:

- https://github.com/atom/atom/blob/f48069c5f848b5ae624b110a1df7800e919eb142/build/tasks/prebuild-less-task.coffee#L11-L36
- https://github.com/atom/atom/blob/ca041384fecf8a17bae078d2813fc94a90498bb8/build/tasks/build-task.coffee#L62-L67


Not sure about:

- https://github.com/atom/atom/blob/f48069c5f848b5ae624b110a1df7800e919eb142/src/task.coffee#L71-L82

thomasjo:
> Should we do that in a separate PR? Currently bootstrap.less is just empty.

Sure, we could remove those references in a separate PR to limit the scope of this one. But simply commenting out the lines in `bootstrap.less` and not removing it feels a bit "messy". I'll see how much work is involved in simply killing it with 🔥

thomasjo:
@simurai I pushed a commit that nukes `bootstrap.less` and (hopefully) all references to it.

@atom/feedback Some 👀 on this would be great. I believe this PR is ready to 🚢

thomasjo:
🚀

Issue #9247, opened 2015-10-21
Multiple Atom release channels can't be installed side-by-side on Windows


MethodGrab:
**Expected**
According to the [beta info page](https://atom.io/beta) Atom Beta should "run side-by-side with Atom stable". To me that reads like I should be able to have both the beta and stable installed (and even running) simultaneously on the same system (like with Chrome's stable & beta releases) and there should be separate links to run each build.

**Actual**
On Windows (7 x64), when I install the beta (`1.1.0-beta1`), it replaces the context menu "Open with Atom" shortcut and the start menu shortcuts with the beta so there is no apparent way to run stable without reinstalling.
Reinstalling stable works but it replaces all the shortcuts with links to the stable version so there is no way to run the beta without re-installing.

![image](https://cloud.githubusercontent.com/assets/2421069/10649437/18a1a58c-783c-11e5-8c7c-fe4be8b6a49d.png)

benogle:
cc @raelyard

maxbrunsfeld:
Ok, I had to stop work on this for the time being because of some PathLength issues on windows. We need to upgrade our build infrastructure to use npm 3, which does a better job of reducing path nesting. Until that time, I'm reopening this.

damieng:
With the file associations and shell integration having the beta executable named something different - atombeta.exe for example - would also make this work better. Thoughts?

maxbrunsfeld:
> having the beta executable named something different - atombeta.exe for example - would also make this work better.

:+1: Ah, I didn't realize that. We already rename a bunch of stuff based on the `channel` (`stable` vs `beta`) in the build scripts, so this would just be one more.

damieng:
Now the shell integration is coming out of the installer and into Atom settings we should make sure we use productName instead of atom.exe everywhere when we do that work in #5901

damieng:
Included in the forthcoming shell integration options.

calebmeyer:
@damieng what's the process/timeline for getting both installed side by side on windows?

damieng:
It's checked into master now on both atom/atom and atom/settings-view and should ship in Atom 1.10.

Ben3eeE:
Reopening because I am not able to install Atom beta side by side with Atom stable on Windows 7. Installing beta removes the stable install. See below for a full list of what I did to try and install `Atom 1.12.0-beta4` side by side with `Atom 1.11.2`.

I can test it out on Windows 10 later if required to see if stable and beta side by side works there.

[...]

PR #11929, opened 2016-06-07
Allow different versions of Atom to run simultaneously on Win32

damieng:
Allow side-by-side versions to run rather than whichever version of Atom was launched first ends up opening all subsequent requests.

### DIFF
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -32,7 +32,7 @@ class AtomApplication
   @open: (options) ->
     unless options.socketPath?
       if process.platform is 'win32'
-        options.socketPath = '\\\\.\\pipe\\atom-sock'
+        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-sock"
       else
         options.socketPath = path.join(os.tmpdir(), "atom-#{options.version}-#{process.env.USER}.sock")
 


lee-dohm:
Looks good to me!

lee-dohm:
This should at least apply #9247, if not fix it, yes?

paulcbetts:
Maybe instead :fire: this whole code path and replace it with `app.makeSingleInstance` instead?

damieng:
@paulcbetts Does that auto-distinguish between versions and can we also still provide an override? - we do that in specs to allow multiple side-by-side https://github.com/atom/atom/blob/720ad9c0b4b46e5e10b6d2381f150d23cf775433/spec/integration/helpers/start-atom.coffee#L14

@lee-dohm Would we be okay with deprecating the socket-path command line option?

paulcbetts:
@damieng It distinguishes via the path to the executable so it should Just Work. You can always choose not to call it in certain circumstances too (you definitely should _not_ call it in the Squirrel event handler for example)

damieng:
@lee-dohm It doesn't completely solve that issue - we still need the shell integration options to honor atombeta.exe but that will happen as part of moving the shell options into settings.

damieng:
While I agree that long-term we should switch to the electron version that's more time and consideration than I have right now and this unblocks the current desired scenarios so I'm going to go ahead and merge unless I see an objection.

PR #11930, opened 2016-06-08
Fix overflowing of the tabs

simurai:
This is a follow up for #11866.

In this case when there are too many tabs, they start to push `atom-workspace-axis.vertical` beyond the window width.

image

Fixes https://github.com/atom/one-dark-ui/issues/138


simurai:
Should this also go in `1.9` and it might be broken there too.

iurevych:
Hey @simurai congrats on your new occupation! I'm sure Atom will look better going forward!

PR #11942, opened 2016-06-08
Complete documentation for Notification creation

ssorallen:
Notifications support several options that were yet to be documented:
`buttons`, `description`, and `stack`. Add descriptions for each option
and its defaults (where applicable, e.g. .buttons.className).

`stack` is technically available on all notification types, but it
doesn't make sense to use it on types other than 'error' and 'fatal'.
Therefore it's documented only in those two cases.

The usage and expected display, like preformatted vs. Markdown, taken
from [atom/notifications/lib/notification-element.coffee](https://github.com/atom/notifications/blob/v0.64.1/lib/notification-element.coffee).

### DIFF
--- a/src/notification-manager.coffee
+++ b/src/notification-manager.coffee
@@ -33,8 +33,18 @@ class NotificationManager
   #
   # * `message` A {String} message
   # * `options` (optional) An options {Object} with the following keys:
-  #    * `detail` (optional) A {String} with additional details about the
-  #      notification.
+  #    * `buttons` (optional) An {Array} of {Object} where each {Object} has the
+  #      following options:
+  #      * `className` (optional) {String} additional class name add to the
+  #        button. It will already receive `btn btn-success`.
+  #      * `onClick` (optional) {Function} callback to call when the button is
+  #        clicked. The context will be set to the {NotificationElement}
+  #        instance.
+  #      * `text` {String} inner text for the button
+  #    * `description` (optional) A {String} that will be rendered as Markdown
+  #      to describe the notification.
+  #    * `detail` (optional) A preformatted {String} that will be rendered as
+  #      plain text with details about the notification.
   #    * `dismissable` (optional) A {Boolean} indicating whether this
   #      notification can be dismissed by the user. Defaults to `false`.
   #    * `icon` (optional) A {String} name of an icon from Octicons to display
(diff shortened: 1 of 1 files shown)


Gawdl3y:
Isn't `onDidClick` the correct event for the buttons?

ssorallen:
@Gawdl3y You are right, thanks for catching that.

https://github.com/atom/notifications/blob/a14d3a0cc771687439176ec949ab4079add15341/lib/notification-element.coffee#L120

src/notification-manager.coffee
```diff
@@ -46,8 +56,18 @@ class NotificationManager
   #
   # * `message` A {String} message
   # * `options` (optional) An options {Object} with the following keys:
-  #    * `detail` (optional) A {String} with additional details about the
-  #      notification.
+  #    * `buttons` (optional) An {Array} of {Object} where each {Object} has the
+  #      following options:
+  #      * `className` (optional) {String} additional class name add to the
```
as-cii: What about writing this as: 

```
A {String} representing a class name to add to the button's default class name (`btn btn-info`).
```

src/notification-manager.coffee
```diff
@@ -46,8 +56,18 @@ class NotificationManager
...
+  #        button. It will already receive `btn btn-info`.
+  #      * `onDidClick` (optional) {Function} callback to call when the button
+  #        has been clicked. The context will be set to the
+  #        {NotificationElement} instance.
+  #      * `text` {String} inner text for the button
+  #    * `description` (optional) A {String} that will be rendered as Markdown
+  #      to describe the notification.
+  #    * `detail` (optional) A preformatted {String} that will be rendered as
```
as-cii: Do you think we could remove `preformatted` here? When reading it it felt like the string the user needs to pass has to be preformatted (i.e. be an HTML string or something). What about rewriting it as:

```
A plain-text {String} containing additional details about the notification.
```

Similarly we could update `description` like this:

```
A Markdown {String} with a longer description of the notification.
```

Thoughts? 💭

as-cii:
This looks good to me, @ssorallen 👍 Just two small comments and then I think we can 🚢 it.

src/notification-manager.coffee
ssorallen: The "preformatted" is important because the `description` is rendered with `white-space: pre;`, as preformatted text. White-space is preserved, like in a stack trace that is indented with spaces or tabs.

Both `description` and `detail` are rendering the string `"this is an error\n      buddy"` in the following screenshot:

screen shot 2016-06-21 at 7 33 38 am

Is there a way to make the "preformatted" more clear?

src/notification-manager.coffee
ssorallen: 👍 Sounds good to me.

src/notification-manager.coffee
as-cii: That's a good point. 👍 I wonder if, however, this is just an implementation detail of the `notifications` package rather than being intrinsic to the `detail` field of the `Notification` object.

That said, I can see how it could cause some misunderstandings, so what about something like:

```
A plain-text {String} containing additional details about the notification. By default, this **will** preserve newlines and whitespace when it is rendered.
```

Similarly, we can document the `description` field like:

```
A Markdown {String} containing a longer description about the notification. By default, this **will not** preserve newlines and whitespace when it is rendered.
```

ssorallen:
@as-cii I added the descriptions you mentioned in your comment about whitespace and newlines.

as-cii:
Thanks, @ssorallen! 🙇

PR #11949, opened 2016-06-09
Add .github/ folder

igor-dyatlov:


50Wliu:
Thanks for contributing and the effort you put into this. We prefer the templates and such remain where they are.

PR #11960, opened 2016-06-11
Calmer Spec Suite

simurai:
This makes the Spec Suite less "flickery" by absolutely position the reporter to the bottom.
## Before

specs
## After

specs


jerone:
Doesn't this make reading failing specs harder...

YuriSolovyov:
Maybe add some redundant background so the only blinking part would be fields and buttons?

50Wliu:
Is there any way to not show the underlying editor at all in the specs window?

simurai:
> @jerone Doesn't this make reading failing specs harder...

If there are failures, the reporter (white box) just grows towards the top:

screen shot 2016-06-15 at 9 09 12 am

If there are tons of failures you could resize the window.. or maybe we could add some max-height + overflow scrolling. 👍

simurai:
> Maybe add some redundant background so the only blinking part would be text, fields and buttons?
> Is there any way to not show the underlying editor at all in the specs window?

Yeah, we could do that too. But somehow I liked seeing what is being tested. Although not too useful since it happens so fast (and often there is no UI).

A cool feature would be to have a "pause/stop on failure" checkbox, then you would have time to inspect and maybe understand the failure better. Useful in this recent case: https://github.com/atom/autocomplete-plus/pull/729#issuecomment-224748192 @patrick-steele-idem: "Since I was not able to visually see the editor when running the tests I took the time to reproduce the tests by creating the same provider that the failing tests were using."

But that's out of scope for this PR. Ok, let's just cover the whole "underlying editor".

> @50Wliu whoops, wrong button

Must be muscle memory :stuck_out_tongue_winking_eye:

simurai:
Ok, how about this:

find

Initially fully covered, but when you resize the window, you can get a peek what's going on underneath?

simurai:
I'll merge this, since it's finally green. :sweat_smile:

PR #11961, opened 2016-06-11
Remove workaround for old Windows bug

samhunta:
The code removed is no longer needed as it was fixed in this issue of electron

https://github.com/electron/electron/issues/380

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -157,6 +157,9 @@ class AtomWindow
 
     @setupContextMenu()
 
+    @browserWindow.on 'blur', =>
+      @browserWindow.focusOnWebView()
+
   openPath: (pathToOpen, initialLine, initialColumn) ->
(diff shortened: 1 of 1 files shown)


src/main-process/atom-window.coffee
```diff
@@ -157,17 +157,6 @@ class AtomWindow
...
 
-    if @isSpec
-      # Workaround for https://github.com/atom/atom-shell/issues/380
-      # Don't focus the window when it is being blurred during close or
-      # else the app will crash on Windows.
-      if process.platform is 'win32'
-        @browserWindow.on 'close', => @isWindowClosing = true
-
-      # Spec window's web view should always have focus
-      @browserWindow.on 'blur', =>
```
50Wliu: This code is still needed, just without the `unless @isWindowClosing` part.

samhunta: I will update my pull request

PR #11962, opened 2016-06-11
Add short text overview for untitled file

dabou:
When lot of untitled file are created, we must search in lot of 'untitled' tab. So with this feature, untitled file have a short description for locate quickly the good text-editor.

text-overview-feature

getTitle and getLongTitle return the first 50 characters of the first line or 'untitled' if blank.

(It's my first contribution in atom so shouldn't hesitate if i don't respect guidelines and sorry for my approximative English)

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -217,6 +217,10 @@ class TextEditor extends Model
     @disposables.add @buffer.onDidChangeModified =>
       @terminatePendingState() if not @hasTerminatedPendingState and @buffer.isModified()
 
+    if not @getPath()
+      @disposables.add @buffer.onDidChangeText =>
+        @emitter.emit 'did-change-title', @getTitle()
+
     @preserveCursorPositionOnBufferReload()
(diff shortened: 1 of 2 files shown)


src/text-editor.coffee
```diff
@@ -674,16 +678,26 @@ class TextEditor extends Model
...
   # If the editor's buffer is saved, its title is the file name. If it is
-  # unsaved, its title is "untitled".
+  # unsaved, its title is "untitled" only if the first 50 characters of first
+  # line isnt blank
   #
   # Returns a {String}.
   getTitle: ->
-    @getFileName() ? 'untitled'
+    if @getFileName()
+      return @getFileName()
```
YuriSolovyov: I think it would be better to cache result of `@getFileName()` somewhere

dabou:
YuriSolovyov,
Yes, you are absolutely right.

dabou:
The sources files modifications are outdated.

PR #11973, opened 2016-06-13
:memo: Add label property info on enum settings

Gawdl3y:


damieng:
This adds 'label' to the comments but settings-view doesn't look for such a property when displaying it on-screen.  Without that this PR makes no sense unless I'm missing something.

Gawdl3y:
It actually did use the undocumented `label` property up until 1.9 (IIRC). The `description` property only set the `title` attribute on the select items originally, resulting in blank options.

PR #11991, opened 2016-06-16
Use preferred line length as a *maximum* for soft wrapping if softWrapAtPreferredLineLength is enabled

### LINKED ISSUE #11990: Editor not soft wrapping when editor is narrower than the preferred line length when softWrapAtPreferredLineLength is enabled
screen shot 2016-06-16 at 8 11 50 am

nathansobo:
This will need to be merged into 1.9.x as well.

Fixes #11990


as-cii:
The failure is totally unrelated and will be fixed with #11998. 🚢ping this!

maxbrunsfeld:
This didn't get into `1.9.1`, but we need to publish `1.9.2` tomorrow anyway, so it can go out w/ that.

PR #11994, opened 2016-06-16
Make interactive package tests work with a top-level 'test' directory

nathansobo:
I realize this is vain, but I really hate the word "spec" and would like the option to just call them tests. That works fine when specifying paths on the command line, but the interactive runner was defaulting the path to `spec`. This allows `test` instead if no `spec` directory is present.

cc @as-cii @BinaryMuse


as-cii:
Looks great! Specs failures are unrelated, so I'll go ahead and 🚢 this (they'll get fixed with #11998).

PR #12002, opened 2016-06-18
Update font stack

williambout:
Since Chrome 46 supports `BlinkMacSystemFont`, I think we should use it instead of  `.SFNSText-Regular` (also the private font is not working anymore on macOS Sierra).


mnquintana:
cc @simurai 👀

simurai:
@williambout Yes.. thanks. 🚢

PR #12003, opened 2016-06-18
Replace "Mac OS" and "OS X" with "macOS"

leo:
OS X has been renamed to "macOS". More details [here](http://www.macworld.co.uk/news/mac-software/mac-os-sierra-x-10-12-fuji-features-uk-release-date-macos-touch-id-siri-apfs-wwdc-2016-3630374/).


README.md
```diff
@@ -95,6 +95,6 @@ repeat these steps to upgrade to future releases.
 ## Building
 
 * [Linux](docs/build-instructions/linux.md)
-* [OS X](docs/build-instructions/os-x.md)
+* [macOS](docs/build-instructions/os-x.md)
```
thomasjo: We should rename the target file as well (and update the link).

leo: Done!

README.md
```diff
@@ -95,6 +95,6 @@ repeat these steps to upgrade to future releases.
 ## Building
 
 * [Linux](docs/build-instructions/linux.md)
-* [macOS](docs/build-instructions/os-x.md)
+* [macOS](docs/build-instructions/mac-os.md)
```
thomasjo: Extreme nitpicking, but with the current naming scheme (look at FreeBSD), the file should be named `macos.md`.

---
_PS. Not deliberately trying to :trollface: you by the way, hehe._

thomasjo:
@leo Thanks for this! 🙇 

Looks good to me. @atom/feedback should we 🚢 this sooner rather than later to minimize potential merge conflicts caused by the markdown table commit bombs?

README.md
leo: Fixed! No problem 😄

50Wliu:
:+1: on merging.

mnquintana:
There are a few instances of `OSX` (without the space) left in here – would you mind getting those too? (Except for in `.travis.yml`, unless Travis CI already updated their config values 😆)

README.md
```diff
@@ -95,6 +95,6 @@ repeat these steps to upgrade to future releases.
 ## Building
 
 * [Linux](docs/build-instructions/linux.md)
-* [OS X](docs/build-instructions/os-x.md)
+* [macOS](docs/build-instructions/macos.md)
```
paulcbetts: Please don't change paths for branding purposes without having redirects, this isn't worth breaking links over

leo: IMHO, the day on which we'll have to change it will definitely come. And the longer we wait, the more links will break...

@thomasjo?

thomasjo: @paulcbetts While I can agree in general, is it really a problem in this case? We should rename the file at some point regardless, and I say sooner rather than later.

My vote is to keep this rename. /cc @atom/feedback 

---
_As a side note, an external link directly to a file on a `master` branch from is a bad idea exactly because of reasons such as this. Always, always link to an explicit version/commit._

leo:
@mnquintana Done! 😊 Also opened an issue on Travis CI: https://github.com/travis-ci/travis-ci/issues/6200

mnquintana:
Thanks for doing this! 🚀 🎸

README.md
Daniel15: Can't you just symlink the old name to the new name to ensure links don't break?

PR #12006, opened 2016-06-19
🐛 Fix $NODE_PATH being unset when using -a to add project

rahatarmanahmed:
When the -a flag is used, atom replaces env of the existing window. Since this existing window doesn't rerun `init-application-window.coffee`, the $NODE_PATH env variable becomes unset. This commit moves setting $NODE_PATH earlier in the window creation process so that it gets set regardless. This fixes the [issue](https://github.com/atom/fuzzy-finder/issues/205) where fuzzy-finder's indexing task cannot find the 'atom' module because the $NODE_PATH becomes unset.

### DIFF
--- a/src/initialize-application-window.coffee
+++ b/src/initialize-application-window.coffee
@@ -11,11 +11,6 @@ module.exports = ({blobStore}) ->
   environmentHelpers.normalize({env: env})
   env = process.env
 
-  # Add application-specific exports to module search path.
-  exportsPath = path.join(resourcePath, 'exports')
-  require('module').globalPaths.push(exportsPath)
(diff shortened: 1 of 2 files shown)


neoziro:
👍  please!

maxbrunsfeld:
Hi @rahatarmanahmed, thanks so much for this contribution. We ended up deciding to solve the problem a different way; we put the `NODE_PATH` logic in the renderer process rather than the main process, mainly because renderer process code is easier to test, and some related code was already there.

Still, your diagnosis of the problem was super valuable, and I'm really sorry we let this PR go unanswered for so long.

PR #12008, opened 2016-06-20
Recreate Repository if the same project is added after being removed

as-cii:
Previously promises resolving to destroyed repositories remained in our cache after removing a project folder. This caused subsequent calls to `repositoryForDirectory` to return the previous instance of the `GitRepository`, even after adding the working directory back to the project.

/cc: @atom/core


as-cii:
I'll go ahead and 🚢 this, as the failure on Janky is unrelated to these changes.

PR #12009, opened 2016-06-20
Add line-height to tips to fix appearance on line-breaks

michaelgmcd:


static/messages.less
```diff
@@ -18,6 +18,7 @@ ul.background-message {
   padding: 0;
 
   li {
+    line-height: 42px;
```
thomasjo: We should use some manner of relative size here. Pretty sure this is well-behaved for non-default font sizes. /cc @simurai. 

---
_The choice of 42 is... "cute"_ 😁

mgmcdermott: Yes. That makes a lot of sense. I'll update accordingly.

mgmcdermott:
@thomasjo I've updated to use `font-size * 4`

simurai:
@mgmcdermott do you mind posting a screenshot? Currently the `line-height` gets inherited from `body` which is `1.42857143`. So it should look ok for multi-line tips:

screen shot 2016-06-21 at 11 05 35 am

Maybe you have a theme installed that overrides it?

mgmcdermott:
screen shot 2016-06-20 at 10 31 04 pm

simurai:
Thanks, still can't reproduce it with those two themes installed.

Maybe another package? How does it look when you start Atom with `atom . --safe`. Then only bundled packages will be loaded.

Also, are you using Atom stable? `1.8`?

simurai:
@mgmcdermott Or in case you're familiar with the DevTools, do you see some `line-height` applied when inspecting:

screen shot 2016-06-21 at 1 09 27 pm

mgmcdermott:
Sorry for the lack of info here. It looks like trying to set the editor line-height also set the line-height of these messages. My fault.

PR #12028, opened 2016-06-22
Don't erase NODE_ENV from environment.

### LINKED ISSUE #12024: Running `atom` from the command line can un-set `process.env.NODE_DEV`, causing performance degradation & JS crashes
Unless Atom is running in dev mode, it sets `process.env.NODE_ENV` to "production". (The relevant code is here: https://github.com/atom/atom/blob/a5fdf3e18a512349e7efb91b3c297b1a2b91bf63/src/initialize-application-window.coffee#L20) This improves the speed of React by disabling its debug mode.

But if Atom is already running, and already has `file.txt` open, and one runs `atom file.txt` from the command line, the command line's environment is used to overwrite Atom's current `process.env`. Most people don't have `NODE_DEV` set in their environment, so React falls back to its slow debug mode. 

The code responsible for "cloning" the environment is https://github.com/atom/atom/blob/a5fdf3e18a512349e7efb91b3c297b1a2b91bf63/src/environment-helpers.js#L75 Notice that it deletes the existing environment.

If React starts in production mode, it won't create some key debugging data structures. If it later changes to dev mode (because `process.env.NODE_DEV` becomes undefined), it will try to access those structures and crash.

This bug is responsible for several reported issues in packages that I maintain: https://github.com/Floobits/atom-term3/issues/63
https://github.com/Floobits/floobits-atom/issues/121

ggreer:
Atom sets NODE_ENV to 'production' earlier, but some code paths can cause it to be un-set. This degrades performance and sometimes crashes React. Fixes #12024.

I've tested this on Linux and macOS, but not Windows.

### DIFF
--- a/src/environment-helpers.js
+++ b/src/environment-helpers.js
@@ -72,7 +72,10 @@ function needsPatching (options = { platform: process.platform, env: process.env
 // underlying functionality.
 function clone (to, from) {
   for (var key in to) {
-    delete to[key]
+    // Don't erase NODE_ENV. Fixes #12024
+    if (key !== 'NODE_ENV') {
+      delete to[key]
+    }
   }
 
   Object.assign(to, from)


ggreer:
Whoops. Didn't see the style rule about single quotes. Force-pushed a fix.

ggreer:
The build error seems to be a failure to download a dependency.

50Wliu:
The failing spec has been notoriously flaky for a while now on Travis and can be safely ignored.

/cc @lee-dohm

lee-dohm:
I have this on my plate to look at. I'll get to it as soon as I can.

lee-dohm:
I got this all tested. It looks good. Thanks for contributing!

PR #12031, opened 2016-06-22
:memo: Mention how models must implement getTitle() [ci skip]

Jonahss:


src/workspace.coffee
```diff
@@ -589,7 +589,10 @@ class Workspace extends Model
 
   # Public: Register an opener for a uri.
   #
-  # An {TextEditor} will be used if no openers return a value.
+  # Atom loops through opener functions until one returns a value for a given uri.
```
lee-dohm: I think this line is confusing. You should probably mention that it does this looping you mention when someone uses the Atom API to open a file.

Jonahss:
updated. I can squash commits if good to merge.

lee-dohm:
Squashing isn't necessary. Thanks for contributing!

PR #12038, opened 2016-06-24
Renaming Linux sections; adding SUSE section

fusion809:
Hi,

In this pull request I added a SUSE section, I also improved the accuracy of the Linux section titles (as these instructions should work on derivative distributions too), specifically:
- Debian Linux (Ubuntu) &rarr; Debian / Ubuntu / Linux Mint / derivatives thereof
- Red Hat Linux (Fedora) &rarr; CentOS / Fedora / Red Hat Enterprise Linux / derivatives thereof

Thanks for your time,
Brenton

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ The `.zip` version will not automatically update.
 Using [chocolatey](https://chocolatey.org/)? Run `cinst Atom` to install
 the latest version of Atom.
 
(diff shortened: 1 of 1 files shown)


damieng:
I think the 'derivatives thereof' is redundant. If you know your Linux distro is a derivative of it then you are used to following instructions for that.

PR #12068, opened 2016-06-29
-webkit cleanup

simurai:
- [x] Remove all `-webkit-` prefixes that aren't needed anymore.
- [x] Add missing `min-width: 0;` to prevent #11866


simurai:
Merging.. CI fails, but to my understanding, unrelated.

PR #12077, opened 2016-06-30
Introduce TextEditor.prototype.update

as-cii:
This will make it easier for views/components that use etch to use a `TextEditor` in their `render` method.

To make `scrollPastEnd` and `autoHeight` updatable, I segregated the knowledge of those two parameters into the `TextEditor` object, so that `TextEditorPresenter`, `TextEditorComponent` and `TextEditorElement` can just use it, applying the necessary changes upon initialization and when the `onDidChangeScrollPastEnd` and `onDidChangeAutoHeight` events trigger.

/cc: @nathansobo @atom/core


as-cii:
This seems ready for a second 👀: the latest commit returns a `Promise` in the `update` method which either:

1. Resolves on the next animation frame if we have a `TextEditorElement` instance assigned on the `TextEditor`;
2. Resolves right away if this editor has no element associated with it.

@nathansobo: for now this feels like a good tradeoff. Seems like a good next step would be to start extracting the editor into its own package, so that we isolate the dependency, and then we can refactor its internals as appropriate.

maxbrunsfeld:
This is being included in https://github.com/atom/atom/pull/12125

PR #12086, opened 2016-07-02
Improve incompatible native module error message

### LINKED ISSUE #6671: Auto Indent isn't idempotent
I'm using Atom 0.196.0 on Ubuntu. I thought it might have been a plugin causing the problem, but you can see from the video below that the bug exists in safe mode too.

It seems like I can pretty reliably get Auto Indent to do the right thing if I strip all indentation before triggering it. I know this was working in 0.188, but because you don't [provide a PPA](https://github.com/atom/atom/issues/2956) or any other [automated update mechanism](https://github.com/atom/atom#debian-linux-ubuntu) I don't get every version.

autoindent

### LINKED ISSUE #6771: Atom should specify which incompatible native module caused `require` to fail
It would be nice in a package with many native modules, if the logged message in the console was not:

> Failed to require the main module of 'git' because it requires an incompatible native module.
Run `apm rebuild` in the package directory to resolve.

but was instead:

> Failed to require the main module of 'git' because it requires 'module-name', an incompatible native module.
Run `apm rebuild` in the package directory to resolve.

Glavin001:
Close #6771.

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -426,8 +426,8 @@ class Package
     return @mainModule if @mainModuleRequired
     unless @isCompatible()
       console.warn """
-        Failed to require the main module of '#{@name}' because it requires an incompatible native module.
-        Run `apm rebuild` in the package directory to resolve.
+        Failed to require the main module of '#{@name}' because it requires an incompatible native module (#{_.map(@incompatibleModules, 'name').join(', ')}).
+        Run `apm rebuild` in the package directory and restart Atom to resolve.
       """
       return
     mainModulePath = @getMainModulePath()


Ben3eeE:
I think you linked the wrong issue.

Glavin001:
Absolutely did. Sorry about that. Fixed linked issue.

Glavin001:
Also force pushed an amended commit message with the correct issue reference.

thomasjo:
@Glavin001 How do you feel about either of these variants?

1. _[…] requires incompatible native **modules** […]_
~~~
Failed to require the main module of '#{@name}' because it requires incompatible native modules (#{_.map(@incompatibleModules, 'name').join(', ')}).
Run `apm rebuild` in the package directory and restart Atom to resolve.
~~~
2. _[…] requires **one or more** incompatible native **modules** […]_
~~~
Failed to require the main module of '#{@name}' because it requires one or more incompatible native modules (#{_.map(@incompatibleModules, 'name').join(', ')}).
Run `apm rebuild` in the package directory and restart Atom to resolve.
~~~

Just makes it read better in case there are actually more than one incompatible module.
Extremely minor thing though, I know. Let me know what you think, then I'll merge this.

Glavin001:
I like the `[…] requires **one or more** incompatible native **modules** […]` message. I'll make the appropriate changes when I get home late today. Thanks!

lee-dohm:
Updated and merged as #12158.

PR #12095, opened 2016-07-04
update nodejs installation instructions link for Fedora/CentOS/RHEL

ammarnajjar:
In the nodejs installation instructions for Fedora/Centos/RHEL, the old [link](https://github.com/nodejs/node-v0.x-archive/wiki/Installing-Node.js-via-package-manager#enterprise-linux-and-fedora) redirects the user to another [link](https://github.com/nodejs/node/wiki) which is the nodejs wiki, and where there are no installation instructions.
The correct link should be [https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora](https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora), which contains the most recent instructions to install nodejs.


mnquintana:
Thanks for fixing this! 🎉

Issue #3362, opened 2014-08-22
Unexpected Ctrl + Del behaviour


aerouk:
In other programs, pressing **Ctrl + Del** on the end of a line would delete newline characters and bring the next line after the cursor.

Example in Sublime Text:

![](https://s.kpvp.in/gifs/WfVaC.gif)

([https://s.kpvp.in/gifs/WfVaC.gif](https://s.kpvp.in/gifs/WfVaC.gif))

jhasse:
I think #4026 is a duplicate of this.

aerouk:
@jhasse that issue does show the problem I'm having, yeah

dwelle:
As mentioned in #4026, you can rebind via:

```
'ctrl-backspace': 'editor:delete-to-previous-word-boundary'
'ctrl-delete': 'editor:delete-to-next-word-boundary'
```

glen-84:
```
Atom                    delete-to-end-of-word
Brackets                delete-to-next-word-boundary
Eclipse                 delete-to-next-word-boundary
Programmer's Notepad    delete-to-next-word-boundary
Sublime Text            delete-to-next-word-boundary
Visual Studio 2015      delete-to-next-word-boundary
```

Spot the odd one out.

Please change the default.

YemSalat:
I also think that the default needs to be changed. In #4026 it is justified by the fact that this behavior "is similar to most text boxes on your system"
But that is exactly why I don't write code in text boxes and if I ever have to - I miss my editor at those times and one of the things I miss is exactly removing only whitespace on Ctrl+Backspace/Del

I do not think that Atom should be copying other editors, but this is not "copying" - this is complying with common code editor logic.

lucasvw:
:+1: change the default

joshrhoderick:
So Atom *really is* going to become the new Emacs: incredibly powerful but unusable out-of-the-box due to unintuitive behavior. I shouldn't need to modify my configuration just to retain the standard behavior of every other editor I have ever used. My text editor shouldn't behave like a text field, even if it really is one under the covers.

jims:
:+1: This is probably my nr 1 grief with text editing in atom!

DKhalil:
This is so incredibly annoying, +1

splintor:
:thumbsup:

westtrade:
:+1:

efkan:
+1

zenbaku:
+1, number one reason to keep going back to sublime.

[...]

PR #12098, opened 2016-07-05
Tweak / Change ctrl-backspace behaviour - delete to word boundary

YemSalat:
This addresses the unexpected `ctrl+backspace` behaviour: https://github.com/atom/atom/issues/3362

Changes the default `ctrl + backspace/del` (`alt + backspace/del` on osx) behaviour from deleting to the end of word to deleting to the word boundaries.
This makes atom behave similar to all other editors and fixes an issue that has been bothering a lot of people for a while (check the link above)


thomasjo:
@atom/feedback How do we feel about this? This is a breaking change, and given our recent track record of breaking people's workflow, I think we need to be super careful 🎱 

Not sure how to go about this though. Adding yet another checkbox is **not** a good idea. Create a tiny little package? 😕

YemSalat:
I think this is different from breaking people's workflows. Because this change actually _fixes_ workflows.

Let me explain:
- ALL  editors except for Atom have this behaviour (emphasis on 'all' here because literally every  other editor does)
  -- this means that atom behaviour is familiar only to those for whom Atom is the _first_ text editor ever.
- Removing to word boundary (this PR) is extremely useful for removing whitespace, this behaviour is not covered by other key combinations
- Current Atom behaviour for `ctrl-backspace` can also be achieved with `ctrl-shift-leftArrow` and then removing selection.
- The original issue (https://github.com/atom/atom/issues/3362) has a few duplicates and many +1s which I think is a pretty strong signal that the workflow is currently broken.

thomasjo:
It doesn't matter if every single editor ever invented by mankind has a different behavior — this has been the default behavior in Atom since its inception. Hence, if we change this, it will change default behavior that people very likely have gotten used to. Those who don't like the default behavior will have undoubtedly already customized their keybindings.

> The original issue (#3362) has a few duplicates, many +1s and is still getting comments (first open 2 years ago), which I think is a pretty strong signal that the workflow is currently broken.

20 or so (I'm being generous) +1s out of the million or so users that Atom has, is hardly strong evidence of a broken workflow. Especially given how trivial it is to customize this.

thomasjo:
And for the record, I'm not saying we will **not** accept this PR, but we should be extremely careful. We need to weigh the pros and cons first.

YemSalat:
I am definitely not trying to say that Atom did something wrong by having a different behaviour from the beginning - I am just providing arguments for why I think it should be changed (being different from all other editors is just one of them).
I also doubt that many people have gotten used to the Atom behaviour as I think it has less use cases then the 'classic' one. This of course is just speculation, but I am honestly having hard times coming up with use cases where deleting to end of word will be more useful then deleting to word boundary (as 'boundary' stops at ends of words but also provides more 'flexibility')

> 20 or so (I'm being generous) +1s out of the million or so users that Atom has
I counted 47 +1s :) (don't forget to count the reactions as well)
Also, out of million Atom users I think only a small fraction is commenting on github repo (otherwise we could just make a poll and see what everyone prefers)

glen-84:
@thomasjo Don't forget about *future* users of Atom as well. Comments like [this](https://github.com/atom/atom/issues/3362#issuecomment-230430967) will continue to appear, and additional issues will most-likely be submitted.

You're inconveniencing more users by keeping this behaviour than by changing it.

ElliottJRo:
@thomasjo could you specify a use-case where the user would actually use the current `ctrl + bksp` behavior? I have always used it to fix spacing issues and its really annoying when Atom deletes actual code (either a semicolon or `>` in case of html)

thomasjo:
I think it makes perfect sense to have the keybind be changed as proposed, but let me reiterate; it is a *breaking change*. That is why I am saying we need to *carefully* review this.

lee-dohm:
There are a bunch of bits of functionality wrapped up in here that I suspect are being conflated but I don't really have time to dig in to all the little edge cases. Here is the test case I looked at:

1. Open the editor under test
1. Open a new blank file
1. Type in `testing testing testing` and leave the cursor flush against the last `g` on that line
1. Press <kbd>Ctrl/Alt+Backspace</kbd> multiple times and observe

Here is what I see (pipe character `|` is used to denote the cursor position):

* Atom default on macOS
    1. `testing testing |`
    1. `testing |`
    1. `|`
* TextEdit (built-in text editor) on macOS
    1. `testing testing |`
    1. `testing |`
    1. `|`
* Sublime Text on macOS
    1. `testing testing |`
    1. `testing |`
    1. `|`
* TextMate on macOS
    1. `testing testing |`
    1. `testing |`
    1. `|`

* Notepad (built-in text editor) on Windows - <kbd>Ctrl+Backspace</kbd> not supported
* Sublime Text on Windows
    1. `testing testing |`
    1. `testing |`
    1. `|`
* Atom on Windows
    1. `testing testing |`
    1. `testing |`
    1. `|`

The only editor that I found that didn't work this way was [BBEdit](http://www.barebones.com/products/bbedit/):

* BBEdit on macOS
    1. `testing testing |`
    1. `testing testing|`
    1. `testing |`
* Atom proposed change
    1. `testing testing |`
    1. `testing testing|`
    1. `testing |`

Because of the above, I am 👎 on accepting this PR as-is. I agree with @thomasjo, that any changes to defaults at this point have to be addressed carefully. I'll bring this up to the team, but I don't see any burning need to make this change right now especially since everyone can customize this very easily with a simple keymap change.

ElliottJRo:
@lee-dohm the use case that differs is when there are multiple lines see #3362 

```
Atom                    delete-to-end-of-word
Brackets                delete-to-next-word-boundary
Eclipse                 delete-to-next-word-boundary
Programmer's Notepad    delete-to-next-word-boundary
Sublime Text            delete-to-next-word-boundary
Visual Studio 2015      delete-to-next-word-boundary
Visual Studio Code      delete-to-next-word-boundary
```

lee-dohm:
@ElliottJRo I realize that, but this change doesn't confine itself to only when dealing with multiple lines, so I have to look at the whole picture.

YemSalat:
> Because of the above, I am 👎 on accepting this PR as-is. 

Sorry, I did not understand your reasoning, are you against accepting this because BBEdit does the same as Atom? Can you please clarify a bit?

lee-dohm:
It boils down to:

1. The current behavior matches the default behavior of the underlying OS for macOS and Windows for all except maybe the multi-line edge case
1. The current behavior matches the default behavior of multiple programmer's editors on macOS and Windows for all except maybe the multi-line edge case
1. The proposed change will break items 1 and 2 above for a debatable improvement in the multi-line edge case

I don't believe item 3 is worth the price of breaking 1 and 2.

YemSalat:
@lee-dohm how are cases 1 and 2 broken if the behaviour is the same for both "delete-to-end-of-word" and "delete-to-word-boundary"

lee-dohm:
@YemSalat Cases 1 and 2 are broken because the proposed behavior will be a change for the cases I specified above. I won't advocate for a fix that breaks other behavior in the process.

If you want to fix #3362, the fix must be limited to multi-line behavior only. All other behavior must stay constant.

YemSalat:
@lee-dohm I am sorry, but I still do not follow your logic.

In a non-multiline case both behaviours are same. What will be a change?

>  All other behavior must stay constant.
It already does??

lee-dohm:
@YemSalat I can only refer you to [the above comment](https://github.com/atom/atom/pull/12098#issuecomment-230624611) where I outlined how Atom's proposed behavior would change relative to the current behavior. I realize it is long, but if you look for "proposed behavior" you will see there is a change.

YemSalat:
@lee-dohm are you refering to not removing the space between the words?
Can you re-test this please? As I am getting the same result in both Atom and Sublime on OSX.

PR #12117, opened 2016-07-07
Change select-to-end-of-line to work like move-to-end-of-line

cinnamonKale:
See discussion in atom#7159

### DIFF
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -258,7 +258,7 @@ class Selection extends Model
   # Public: Selects all the text from the current cursor position to the end of
   # the screen line.
   selectToEndOfLine: ->
-    @modifySelection => @cursor.moveToEndOfScreenLine()
+    @modifySelection => @cursor.moveToEndOfLine()
 
   # Public: Selects all the text from the current cursor position to the end of
   # the buffer line.


nathansobo:
@lee-dohm can you take a look at this and maybe help it across the 🏁 line? I definitely think it makes sense to have `selectToEndOfLine` match the behavior of `moveToEndOfLine`, but we don't want to behavior of default key bindings.

kylesimmonds:
@nathansobo thanks for the comment. I'm trying to get the build to work locally to test it, but having some issues. See #12120 for all my issues :expressionless:.

That's all to say, I prematurely made this pull request without testing it locally first. I definitely welcome any help with this, but I also don't want to waste anyone's time with something I may need to work out myself.

lee-dohm:
@kylesimmonds There are a few things that need to happen all at the same time in this PR:

* [x] `TextEditor.moveToEndOfLine` and `TextEditor.selectToEndOfLine` need to work the same
* [ ] Create a `TextEditor.selectToEndOfScreenLine` function that works the way `TextEditor.selectToEndOfLine` works currently
* [ ] Write specs for the above
* [ ] Create an `editor:select-to-end-of-screen-line` command that calls the new `TextEditor.selectToEndOfScreenLine` function
* [ ] Map all keybindings that currently map to `editor:select-to-end-of-line` to the new `editor:select-to-end-of-screen-line`

kylesimmonds:
As I stated in the new PR, I completely screwed up my local repo and just started over. New PR is here: atom/atom#12186

PR #12123, opened 2016-07-08
Windows shell integration system settings

damieng:
This is the Atom side of atom/settings-view#811 that provides the parts necessary for that one and facilitates;
1. Only installing the file handler by default (no more context menus by defaut but you keep them if you already have them)
2. Allows side-by-side of Atom.exe and AtomBeta.exe which can set, unset and upgrade independently


src/main-process/win-shell.coffee
```diff
@@ -0,0 +1,53 @@
...
+
+exeName = Path.basename(process.execPath)
+appPath = "\"#{process.execPath}\""
+appName = exeName.replace('atom', 'Atom').replace('beta', 'Beta').replace('.exe', '')
+
+class ShellOption
+  constructor: (key, parts) ->
+    @key = key
+    @parts = parts
+  isRegistered: (callback) =>
```
lee-dohm: Is there a reason to use fat arrows for these functions?

*nit* It is typical to have a blank line between functions.

src/main-process/win-shell.coffee
```diff
@@ -0,0 +1,53 @@
...
+    new Registry({hive: 'HKCU', key: "#{@key}\\#{@parts[0].key}"})
+      .get @parts[0].name, (err, val) =>
+        callback(not err? and val.value is @parts[0].value)
+  register: (callback) =>
+    doneCount = @parts.length
+    @parts.forEach((part) =>
```
lee-dohm: It's more idiomatic in CoffeeScript to use `for part in @parts`.

src/main-process/win-shell.coffee
damieng: The fat arrows are required so that the @ operator points to this object not to the window when the callback functions are executed.

I can add spacing.

Ben3eeE:
> Allows side-by-side of Atom.exe and AtomBeta.exe which can set, unset and upgrade independently

Do we know when this is set to release? As in which beta version would be the first I can install side by side with stable? I am very interested in some 1.10 features but wouldn't want to lose safety of a stable install on main workstation.

PR #12125, opened 2016-07-08
Remove some of the TextEditor class's Atom-specific dependencies

maxbrunsfeld:
The goal here is to remove some of the barriers to extracting the `TextEditor` class into a library. This means removing the following atom-specific dependencies:
- [x] Config
  
  Instead of each `TextEditor` internally observing the config and updating its behavior accordingly, the `TextEditorRegistry` (an atom-specific class) will observe the config and update text editors from outside, using new public APIs like `setUndoGroupingInterval`.
  
  In order to handle settings like `nonWordCharacters` which can vary according to the cursor's position, the `TextEditor` will have an optional 'scoped settings delegate', which will allow the editor to query for settings at particular scopes. Example:
  
  ``` js
  editor.setScopedSettingsDelegate({
    getCommentStartString (scopes) {
      if (languageLooksLikeC(scopes)) {
        return "//"
      } else {
        return "#"
      }
    },
  
    getNonWordCharacters (scopes) {
      // logic ...
    }
  })
  ```
- [x] GrammarRegistry

This also removes a duplicate call to `DisplayLayer::reset` that was happening in the `TextEditor` constructor. For a big file like `text-editor-component-spec.js`, this makes the editor visible about  120 ms sooner 🎉 .


src/text-editor.coffee
```diff
@@ -125,12 +124,11 @@ class TextEditor extends Model
     {
       @softTabs, @firstVisibleScreenRow, @firstVisibleScreenColumn, initialLine, initialColumn, @tabLength,
       @softWrapped, @decorationManager, @selectionsMarkerLayer, @buffer, suppressCursorCreation,
-      @mini, @placeholderText, lineNumberGutterVisible, @largeFileMode, @config, @clipboard, @grammarRegistry,
+      @mini, @placeholderText, lineNumberGutterVisible, @largeFileMode, @clipboard, @grammarRegistry,
       @assert, grammar, @showInvisibles, @autoHeight, @scrollPastEnd, @editorWidthInChars,
       @tokenizedBuffer, @ignoreInvisibles, @displayLayer
     } = params
 
-    throw new Error("Must pass a config parameter when constructing TextEditors") unless @config?
```
maxbrunsfeld: ❗ 💥  💣

src/text-editor.coffee
```diff
@@ -124,13 +123,12 @@ class TextEditor extends Model
...
       @softTabs, @firstVisibleScreenRow, @firstVisibleScreenColumn, initialLine, initialColumn, @tabLength,
       @softWrapped, @decorationManager, @selectionsMarkerLayer, @buffer, suppressCursorCreation,
-      @mini, @placeholderText, lineNumberGutterVisible, @largeFileMode, @clipboard, @grammarRegistry,
+      @mini, @placeholderText, lineNumberGutterVisible, @largeFileMode, @clipboard,
       @assert, grammar, @showInvisibles, @autoHeight, @scrollPastEnd, @editorWidthInChars,
       @tokenizedBuffer, @ignoreInvisibles, @displayLayer
     } = params
 
     throw new Error("Must pass a clipboard parameter when constructing TextEditors") unless @clipboard?
-    throw new Error("Must pass a grammarRegistry parameter when constructing TextEditors") unless @grammarRegistry?
```
maxbrunsfeld: :bomb: :boom: :boom:

src/text-editor.coffee
```diff
@@ -3338,42 +3338,99 @@ class TextEditor extends Model
...
   setScopedSettingsDelegate: (@scopedSettingsDelegate) ->
 
+  # Experimental: Retrieve the {Object} that provides the editor with settings
+  # for specific syntactic scopes.
   getScopedSettingsDelegate: -> @scopedSettingsDelegate
 
+  # Experimental: Enable or disable automatic language-based indentation when
+  # inserting a line-break in the editor.
+  #
+  # * `autoIdent` a {Boolean} indicating whether to enable or disable automatic
```
50Wliu: autoIndent

maxbrunsfeld: Thanks 🎩

maxbrunsfeld:
I think this is ready to go. All redundant calls to `DisplayLayer.reset` are gone.

spec/fixtures/sample.txt
```diff
@@ -1 +1 @@
-Some textSome textSome text.
+Some textSome textSome textSome textSome text.
```
50Wliu: Was this intended?

maxbrunsfeld: No. Thanks!

src/null-grammar.js
```diff
@@ -0,0 +1,13 @@
+/** @babel */
+
+import {Disposable} from 'event-kit'
+
+module.exports = Object.freeze({
```
as-cii: Since we're using the `import` notation above, what do you think about changing this into `export default Object.freeze(…)`?

as-cii:
Other than a super minor comment, this looks good to me. 🚢

nathansobo:
Great work grinding to this. We're getting close.

Stanzilla:
@nathansobo close to what, if I may ask?

nathansobo:
@Stanzilla Close to extracting the editor as a library for use in other applications. We have to drop the shadow DOM which is going to be tricky, and then we can do it. Once it's extracted we have some cleanup to do in the rendering layer.

PR #12145, opened 2016-07-12
Update package names

pkgdemon:
These packages now exist as node012, and npm012.


mnquintana:
Those numbers at the end aren't referring to a specific version number, right? Just want to make sure.

damieng:
No response.

PR #12150, opened 2016-07-12
Introduce @use-custom-controls variable

simurai:
This allows themes to **opt-out** of custom styling for controls, like radio, checkbox, select etc. Useful for themes that like to mimic a more native look and feel, like [unity-ui](https://atom.io/themes/unity-ui), [native-ui](https://atom.io/themes/native-ui) etc.

| `@use-custom-controls: true;` | `@use-custom-controls: false;` |
| --- | --- |
| screen shot 2016-07-12 at 3 30 56 pm | screen shot 2016-07-12 at 3 30 13 pm |
### Questions
- [ ] Variable name?
  - Drop the `use-` and just have `@custom-controls: true;`?
  - Invert it to `@use-native-controls: false;`

### DIFF
--- a/static/variables/ui-variables.less
+++ b/static/variables/ui-variables.less
@@ -83,3 +83,4 @@
 // Other
 
 @font-family: 'BlinkMacSystemFont', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
+@use-custom-controls: true; // false uses native controls


simurai:
Let's 🐑  this, so it makes it into the same release as https://github.com/atom/atom-ui/pull/8.

PR #12170, opened 2016-07-14
WIP: Use apm with npm v3 and Node v4

as-cii:
Refs: #11897

PR #12172, opened 2016-07-14
Clear out apm folder in cibuild

maxbrunsfeld:
Previously, the atom master build was failing because the `apm/node_modules` folder was somehow getting into an invalid state. Probably related to https://github.com/atom/atom/pull/12170.


maxbrunsfeld:
The failing test on CI is a flaky tree view spec.

PR #12176, opened 2016-07-15
Upgrade packages that use old versions of graceful-fs

thomasjo:
Bumps packages that have dependencies on graceful-fs to versions that uses graceful-fs v4 or higher. 

This is **necessary** for Node.js v6 support.

TODO:
- [x] Bump fstream.
- [x] Open PRs against [less-cache](https://github.com/atom/less-cache) — https://github.com/atom/less-cache/pull/10
  - [x] Bump less-cache once PRs have shipped.
- [ ] Fix regression during the `prebuild-less` task. See https://github.com/atom/atom/pull/12176#issuecomment-240339043


thomasjo:
@atom/feedback I think this should be good to go. Would appreciate some 👀 and maybe a bit of 🚬 testing.

thomasjo:
Nevermind, turns out this causes a problem during `script/build` (and possibly elsewhere)

```
Running "prebuild-less:src" (prebuild-less) task
Warning: variable @fold is undefined Use --force to continue.

Aborted due to warnings.
```

maxbrunsfeld:
> this causes a problem during `script/build`

I've tracked this problem down to what I believe is a bug in `less`. I have opened a PR fixing the bug: https://github.com/less/less.js/pull/2953

static/variables/octicon-mixins.less
```diff
@@ -1,3 +1,5 @@
+@import "octicon-utf-codes.less";
+
```
maxbrunsfeld: See https://github.com/less/less.js/pull/2953

maxbrunsfeld:
Interesting; the janky build passed, but the circle build failed with

```
Expected '#header {
  color: #4D926F;
}
h2 {
  color: #4D926F;
}
' to be '#header {
  color: #4d926f;
}
h2 {
  color: #4d926f;
}
'
```

Maybe the new version of `less` normalizes color literals to always use capital letters?

maxbrunsfeld:
OK, it turns out that the less bug I mentioned above is the expected behavior, so I've changed our base stylesheets to avoid imports in nested scopes. Now I've run into another less bug:

https://github.com/less/less.js/pull/2955

PR #12179, opened 2016-07-15
trial hotfix 1.8 release

maxbrunsfeld:


nathansobo:
This PR upgrades `status-bar` from `1.2.3` to `1.4.0`. The `status-bar` package specs are failing, which I'm assuming is due to some issue with the Atom core API having changed since 1.2.4. I think we should just merge the hotfix into beta and roll the rail-cars on Monday instead of trying to fix this PR.

PR #12196, opened 2016-07-19
Fixed undefined ATOM_HOME

uhop:
`ATOM_HOME` is `undefined`, which breaks Atom on startup (at least on Mac). Interesting that it is `undefined` even when I explicitly set this variable in `.bashrc`.

The patch is a one-liner for simplicity:
- It checks, if `ATOM_HOME` is falsy. If it is:
  - It checks if `HOME` is truthy. If it is:
    - It sets `ATOM_HOME` to `path.join(HOME, '.atom')`.
  - Otherwise it sets it to an empty string.

The description above dwarves the code.

I don't know what caused the problem, because I saw it before intermittently, but recently I updated node to 6.3.0, and Mac to 10.11.6. The combined change made the problem persistent.

Before debugging the code I tried to reinstall Atom 1.8.0, and 1.9.0-beta0 &mdash; both fail the same way (`undefined` is passed to `path.join()`).

### DIFF
--- a/src/initialize-application-window.coffee
+++ b/src/initialize-application-window.coffee
@@ -21,6 +21,7 @@ module.exports = ({blobStore}) ->
 
   AtomEnvironment = require './atom-environment'
   ApplicationDelegate = require './application-delegate'
+  !process.env.ATOM_HOME && (process.env.ATOM_HOME = process.env.HOME && path.join(process.env.HOME, '.atom') || '');
   window.atom = new AtomEnvironment({
     window, document, blobStore,
     applicationDelegate: new ApplicationDelegate,
--- a/src/initialize-application-window.coffee
+++ b/src/initialize-application-window.coffee
@@ -21,7 +21,7 @@ module.exports = ({blobStore}) ->
 
   AtomEnvironment = require './atom-environment'
   ApplicationDelegate = require './application-delegate'
-  !process.env.ATOM_HOME && (process.env.ATOM_HOME = process.env.HOME && path.join(process.env.HOME, '.atom') || '');
+  not process.env.ATOM_HOME and (process.env.ATOM_HOME = process.env.HOME and path.join(process.env.HOME, '.atom') or '')
   window.atom = new AtomEnvironment({
     window, document, blobStore,
     applicationDelegate: new ApplicationDelegate,


uhop:
One of four Travis CI builds has failed because of a timeout. Now the whole pull request is marked as failing. My understanding is this bug is never going to be fixed. Cheers.

50Wliu:
Just because the build fails doesn't mean it's not ready to be merged.

uhop:
It is a one-liner. It mostly needs a review to see what it does. The testing (if needed) should come after review.

What it does: explicitly sets the default value to `ATOM_HOME`, if it is missing, which should be a no-op, because the default should be used anyway, if not explicitly set, yet it breaks Atom.

uhop:
Today my Atom got automatically updated, and promptly crashed on start. I have to apply the online-liner fix in this pull request manually.

uhop:
That's how the unpatched stack trace looks like with today's release:

screen shot 2016-08-09 at 1 57 53 pm

(Sorry for the image &mdash; in Atom's error console cut/copy is turned off.)

lukasoppermann:
I think I have the same problem. However, for me it is the `/Applications/Atom.app/Contents/Resources/app.asar/src/initialize-application-window.js` not `.coffee`. How can I apply your fix in the js version, to test if it fixes my problem as well?

maxbrunsfeld:
@uhop @lukasoppermann How are you launching Atom when you see this? From the terminal? From Finder?

lukasoppermann:
When I see this I open it from finder, by dragging a folder onto the atom icon or by going to `file > new window`.

Opening Atom from the terminal via `atom` or `atom .` from within a dir, works fine.

maxbrunsfeld:
Do you use a default shell other than `/bin/bash`?

uhop:
@maxbrunsfeld I did more digging. The unpatched Atom runs fine from a terminal (iTerm2 in my case). Yet from a finder (the standard application folder) it fails. I fiddled with my default shell (bash), and its settings. I singled out one command that broke Atom:

```txt
alias env='command env | sort'
```

It was in `.bash_aliases`, which is itself taken from http://www.askapache.com/linux/bash_profile-functions-advanced-shell.html

Commenting out that line helped. I don't know how Atom ended up using it, and where exactly.

Interesting that I have this setup in `.bash_aliases` for a long time, way before Atom started to crash. It probably means that Atom started to rely on certain `env` features recently.

maxbrunsfeld:
Thanks so much for investigating further @uhop; that's very helpful. I have a different solution in mind for this problem, I will get back to you shortly.

maxbrunsfeld:
>  It probably means that Atom started to rely on certain env features recently.

Yep.

maxbrunsfeld:
@uhop I actually still don't understand why that alias causes a problem. What version of Atom are you using? Could you share the output of the following command (with any sensitive content removed)?

```sh
$SHELL -ilc env
```

maxbrunsfeld:
Ok, I believe we have a fix for this. The fix will be released in Atom v1.9.8 and v1.10.0-beta5.

uhop:
One more note: when I inspected `process.env` in debugger (before commenting out the `env` alias), the object has looked corrupted. For example, I have some long prompts in different environment variables (some used by `git-prompt`, some by iTerm2). I can see that a piece of such prompt value is a variable name, and its value consists of several other variables, which include `ATOM_HOME` and some other stuff. That may explain why it was not there, in spite of being explicitly set in `.bashrc`. It does not explain how `process.env` got corrupted, or why a missing environment variable is not defaulted, but used verbatim (as `undefined` value).

uhop:
The requested output:

```txt
✔ 15:43 ~ $ $SHELL -ilc env
Now using node v6.3.1 (npm v3.10.5)
MANPATH=/Users/elazutkin/.nvm/versions/node/v6.3.1/share/man:/usr/share/man:/usr/local/share/man:/opt/X11/share/man:/Applications/Xcode.app/Contents/Developer/usr/share/man:/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/share/man
TERM_PROGRAM=iTerm.app
NVM_CD_FLAGS=
SHELL=/bin/bash
TERM=xterm-256color
TMPDIR=/var/folders/n0/hz24lz8d4rnbc8szyy3qxdc40000gp/T/
Apple_PubSub_Socket_Render=/private/tmp/com.apple.launchd.yFfy5bIGkp/Render
NVM_PATH=/Users/elazutkin/.nvm/versions/node/v6.3.1/lib/node
PERL5LIB=/Users/elazutkin/perl5/lib/perl5:/Users/elazutkin/perl5/lib/perl5:/Users/elazutkin/perl5/lib/perl5
TERM_PROGRAM_VERSION=3.0.5
TERM_SESSION_ID=w0t2p1:F61D8493-60DE-4B9F-BE18-BDDB1CFF358F
PERL_MB_OPT=--install_base "/Users/elazutkin/perl5"
ATOM_HOME=/Users/elazutkin/.atom
USER=elazutkin
NVM_DIR=/Users/elazutkin/.nvm
COMMAND_MODE=unix2003
SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.NMbnCiWYNH/Listeners
__CF_USER_TEXT_ENCODING=0x1F6:0x0:0x0
PATH=/Users/elazutkin/perl5/bin:/Users/elazutkin/.nvm/versions/node/v6.3.1/bin:/Users/elazutkin/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/X11/bin:/Users/elazutkin/perl5/bin:/Users/elazutkin/bin:/Users/elazutkin/Library/Android/sdk/platform-tools:/Users/elazutkin/Library/Android/sdk/tools:/Users/elazutkin/Library/Android/sdk/platform-tools:/Users/elazutkin/Library/Android/sdk/tools
NVM_NODEJS_ORG_MIRROR=https://nodejs.org/dist
PWD=/Users/elazutkin
LANG=en_US.UTF-8
ITERM_PROFILE=Default
XPC_FLAGS=0x0
PS1=\[\e]0;\u@\h: \w\a\]\[\033[01;33m\]\w\[\033[00m\] \$
ITERM_ORIG_PS1=\[\033[0;32m\]✔ \[\033[0;37m\]$(date +%H:%M)\[\033[0;0m\] \[\033[1;33m\]\w\[\033[0;0m\]\[\033[0;0m\] $
XPC_SERVICE_NAME=0
COLORFGBG=7;0
HOME=/Users/elazutkin
SHLVL=2
ITERM_PREV_PS1=\[\e]0;\u@\h: \w\a\]\[\033[01;33m\]\w\[\033[00m\] \$
ITERM_SESSION_ID=w0t2p1:F61D8493-60DE-4B9F-BE18-BDDB1CFF358F
PERL_LOCAL_LIB_ROOT=/Users/elazutkin/perl5:/Users/elazutkin/perl5:/Users/elazutkin/perl5
LOGNAME=elazutkin
NVM_BIN=/Users/elazutkin/.nvm/versions/node/v6.3.1/bin
NVM_IOJS_ORG_MIRROR=https://iojs.org/dist
DISPLAY=/private/tmp/com.apple.launchd.HHPW2pj8Gm/org.macosforge.xquartz:0
SECURITYSESSIONID=186a8
PERL_MM_OPT=INSTALL_BASE=/Users/elazutkin/perl5
_=/usr/bin/env
✔ 09:26 ~ $
```

uhop:
My thoughts on a bug: when aliasing something, options stop working. For example, if I define an alias like that:

```txt
alias env='command env | sort'
```

And invoke it as `env --help` I will see an output from ... `sort` &mdash; it will get all extra options as per textual substitution.

So if somebody runs `env` from a bash session, and uses some options (in fact, any options) they will be ignored by `env` and possibly misinterpreted by `sort`.

As to `process.env` corruption &mdash; I tried to reproduce it in a terminal, but failed. An output of `env | sort` looks fine, and not corrupted in any way. I suspect either Atom manipulates `process.env` directly, and mangles it, or there is some bizarre bug deep down in node, which is triggered by sorting (?) of the environment.

Thank you for taking care of the bug. I am looking forward to test it.

lukasoppermann:
Damn. I had an `env` alias as well. Removing it "fixed" the bug.

uhop:
Upgraded to 1.9.8, and it starts up fine now with and without `env` defined as an alias.

PS: Interesting moment is that I have the same `.bash_aliases` file on my Linux box (with the aliased `env`), yet it worked without problems with the old code. So it must be Mac-specific somehow, e.g., triggering some quirks of Mac's implementation of `env`.

maxbrunsfeld:
> it must be Mac-specific somehow

Yeah, for the time being, we've only added this feature (setting environment variables based on a login shell) on Mac.

raphaelokon:
I have my Atom in a custom location and pointing to it via `export ATOM_HOME` … after the last update all my settings are gone – Atome uses the default location ~/.atom instead of $ATOM_HOME on Mac.

maxbrunsfeld:
@raphaelokon are you still seeing this on v1.9.9?

raphaelokon:
@maxbrunsfeld No – `v1.9.8` That is the latest I can download from atom.io

maxbrunsfeld:
Sorry, I meant are you still seeing this on v1.9.8?

raphaelokon:
Aye – after restarting Atom when it updated to v1.9.8 my custom location is ignored.

maxbrunsfeld:
@raphaelokon I can't reproduce this. What OS are you using? How are launching Atom, and how are you setting the `ATOM_HOME` environment variable?

raphaelokon:
Hmm. I am setting the `ATOM_HOME` in my `.bash_profile` → `export ATOM_HOME="$HOME/path/to/Atom"` →  OS  El Capitan 10.11.6

maxbrunsfeld:
Ok thanks. And how are you opening Atom?

raphaelokon:
Both, clicking on the dock symbol and directly clicking on the app icon in the custom location yield the same. I wiped `~/.atom` and `~/Library/Application Support/Atom` … still same behavior.

raphaelokon:
Any ideas @maxbrunsfeld ?

maxbrunsfeld:
@raphaelokon Yeah, I'm sorry. I didn't realize it was ever possible to use a custom `ATOM_HOME` while running Atom from the finder. This will be fixed in Atom v1.9.9. In the meantime, you can workaround this bug by launching Atom from the terminal.

raphaelokon:
Okay. I restored my settings by downgrading to `v1.9.7` – thanks for fixing it @maxbrunsfeld

raphaelokon:
I recently upgraded to `v1.12.2` → getting the same error in that my custom atom location (and its settings) are not being picked up from my profile :/
Any ideas @maxbrunsfeld ?

PR #11897, opened 2016-06-04
Use apm@1.12.2 with Node v4 and npm v3 \o/

BinaryMuse:
```
                  __              __ __
                 /\ \            /\ \\ \
  ___     ___    \_\ \     __    \ \ \\ \
/' _ `\  / __`\  /'_` \  /'__`\   \ \ \\ \_
/\ \/\ \/\ \L\ \/\ \L\ \/\  __/    \ \__ ,__\
\ \_\ \_\ \____/\ \___,_\ \____\    \/_/\_\_/
 \/_/\/_/\/___/  \/__,_ /\/____/       \/_/


                                    __
                                  /'__`\
     ___   _____     ___ ___     /\_\L\ \
   /' _ `\/\ '__`\ /' __` __`\   \/_/_\_<_
   /\ \/\ \ \ \L\ \/\ \/\ \/\ \    /\ \L\ \
   \ \_\ \_\ \ ,__/\ \_\ \_\ \_\   \ \____/
    \/_/\/_/\ \ \/  \/_/\/_/\/_/    \/___/
             \ \_\
              \/_/
```

This bumps apm to 1.11.1, which includes atom/apm#562 and atom/apm#563. Since npm 3 flattens packages by default, the build had to change relatively significantly, and I'm sure there are more edge cases to track down. This also bumps Atom's own npm to 3.9.5.

Of particular note, I removed the `apm clean` command that runs as part of the build. It didn't seem to be working in the first place, and seems to have trouble with npm 3's package flattening; maybe we can delegate to `npm prune` here. Since published builds get built from scratch, I think this is probably not a huge deal.

@atom/core, @atom/feedback, @damieng, I would love for you to pull down this branch and try to build Atom and make sure it runs correctly and that package-related operations work. It's _extremely likely_ you will have to run `script/clean` first; sorry, things are just too different.

Fixes #5109
Makes #7822 (more) possible
Probably a bunch of others too

PR #12207, opened 2016-07-21
Introduce CircleCI for macOS builds

as-cii:
Depends on #11897. This is the [actual diff](https://github.com/atom/atom/compare/mkt-use-apm-with-npm3-and-node-4...as-circle-ci).

This pull-request tries to supplant Travis and replace it with CircleCI. After trying it out for a while, the build process seems more reliable and faster as well. Moreover, if CircleCI enables parallelism for macOS containers in the future, we could be able to speed up our test suite very easily.

Right now the build fails because of some flaky tests that are not related to CircleCI at all (i.e. Janky and Travis fail too for the same specs), but overall I think we might be able to 🚢 this and let it run alongside Travis for a while so that we can assess how well it performs. If after the test-drive everything looks good, we can :fire: Travis and use CircleCI only.

This is the plan we're on right now:

screen shot 2016-07-20 at 20 47 37

After the experimentation phase, I believe we should switch initially to **Startup** and consider **Growth** too if we run out of minutes.

/cc: @nathansobo @iolsen @atom/core

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![Atom](https://cloud.githubusercontent.com/assets/72919/2874231/3af1db48-d3dd-11e3-98dc-6066f8bc766f.png)
 
-[![macOS Build Status](https://travis-ci.org/atom/atom.svg?branch=master)](https://travis-ci.org/atom/atom) [![Windows Build Status](https://ci.appveyor.com/api/projects/status/1tkktwh654w07eim?svg=true)](https://ci.appveyor.com/project/Atom/atom)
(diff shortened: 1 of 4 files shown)


as-cii:
As a bonus, in 084c5c6 and fbc5a51, I have added the possibility of downloading `Atom.app` when the build completes, as we do for AppVeyor. This should make it easier for everyone to test changes out in the future without downloading and compiling the app manually.

iolsen:
+:100!!!

PR #12212, opened 2016-07-22
Create separate "syntax stylesheet" scoped to atom-text-editor

### LINKED ISSUE #4590: No way to create CSS animations for syntax elements in styles.less?
I'm probably getting something wrong because I haven't worked extensively enough with the Shadow DOM yet. But if you put this in your `styles.less`:

``` less
@keyframes pulsate {
  from {
    color: black;
  }

  to {
    color: #f33;
  }
}

.editor, atom-text-editor::shadow {
  .comment {
    font-style: normal;
    -webkit-animation: pulsate 1s infinite alternate;
  }
}
```

You get spooky pulsing comments with the Shadow DOM turned off and default-colored comments with the Shadow DOM turned on ... but you get un-italic comments in both cases.

/cc @nathansobo @simurai

crazymykl:
Fixes #4590

### DIFF
--- a/menus/win32.cson
+++ b/menus/win32.cson
@@ -16,6 +16,7 @@
       { label: 'Keymap…', command: 'application:open-your-keymap' }
       { label: 'Snippets…', command: 'application:open-your-snippets' }
       { label: 'Stylesheet…', command: 'application:open-your-stylesheet' }
+      { label: 'Syntax Stylesheet…', command: 'application:open-your-syntax-stylesheet' }
       { type: 'separator' }
       { label: '&Save', command: 'core:save' }
       { label: 'Save &As…', command: 'core:save-as' }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,7 +124,7 @@ Include details about your configuration and environment:
 * **What's the name and version of the OS you're using**?
 * **Are you running Atom in a virtual machine?** If so, which VM software are you using and which operating systems and versions are used for the host and the guest?
 * **Which [packages](#atom-and-packages) do you have installed?** You can get that list by running `apm list --installed`.
-* **Are you using [local configuration files](http://flight-manual.atom.io/using-atom/sections/basic-customization/)** `config.cson`, `keymap.cson`, `snippets.cson`, `styles.less` and `init.coffee` to customize Atom? If so, provide the contents of those files, preferably in a [code block](https://help.github.com/articles/markdown-basics/#multiple-lines) or with a link to a [gist](https://gist.github.com/).
+* **Are you using [local configuration files](http://flight-manual.atom.io/using-atom/sections/basic-customization/)** `config.cson`, `keymap.cson`, `snippets.cson`, `styles.less`, `syntax-styles.less`, and `init.coffee` to customize Atom? If so, provide the contents of those files, preferably in a [code block](https://help.github.com/articles/markdown-basics/#multiple-lines) or with a link to a [gist](https://gist.github.com/).
 * **Are you using Atom with multiple monitors?** If so, can you reproduce the problem when you use a single monitor?
 * **Which keyboard layout are you using?** Are you using a US layout or some other layout?
 
--- a/dot-atom/styles.less
+++ b/dot-atom/styles.less
@@ -7,6 +7,9 @@
  * Add your own CSS or Less to fully customize Atom.
  * If you are unfamiliar with Less, you can read more about it here:
  * http://lesscss.org
+ *
+ * This sytlesheet is for the Atom UI outside the text editor itself, to
+ * style content in the text editor's shadow DOM, use the syntax stylesheet.
  */
 
 
@@ -20,13 +23,7 @@
   // background-color: whitesmoke;
 }
 
-// style the background and foreground colors on the atom-text-editor-element itself
-atom-text-editor {
-  // color: white;
-  // background-color: hsl(180, 24%, 12%);
-}
-
-// To style other content in the text editor's shadow DOM, use the ::shadow expression
-atom-text-editor::shadow .cursor {
-  // border-color: red;
+// make your tabs' titles a bit askew
+.tab-bar .tab .title {
+  // transform: rotate(-3deg);
 }
--- a/dot-atom/syntax-styles.less
+++ b/dot-atom/syntax-styles.less
@@ -1,0 +1,29 @@
+/*
+ * Your Syntax Stylesheet
+ *
+ * This stylesheet is loaded when Atom starts up and is reloaded automatically
+ * when it is changed and saved.
+ *
+ * Add your own CSS or Less to fully customize Atom.
+ * If you are unfamiliar with Less, you can read more about it here:
+ * http://lesscss.org
+ *
+ * This stylesheet is for the text editor's shadow DOM, to style other UI
+ * elements, use the regular stylesheet.
+ */
+
+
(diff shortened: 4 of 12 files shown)


crazymykl:
Do I have to do something to get the ball rolling on the review process?

50Wliu:
> Do I have to do something to get the ball rolling on the review process?

Nope, there's just a large PR backlog.

menus/win32.cson
```diff
@@ -16,6 +16,7 @@
       { label: 'Keymap…', command: 'application:open-your-keymap' }
       { label: 'Snippets…', command: 'application:open-your-snippets' }
       { label: 'Stylesheet…', command: 'application:open-your-stylesheet' }
```
mnquintana: Hmmm I wonder if we should call this `UI Stylesheet` now?

simurai: Hmm.. yeah.. Was thinking the same, but it's also less confusing if label and file name match:

Label | File
--- | ---
Stylesheet | `styles.less`
Syntax Stylesheet | `syntax-styles.less`

Changing `styles.less` to `ui-styles.less` would make sense but that's probably a bigger effort and maybe cause even more confusion.

simurai:
Maybe a silly idea but what about using/loading the same `styles.less` file in both places? Outside of the editor **and** inside the editor.

That would keep things simple and animations should work. It's against the idea of the Shadow DOM that protects against style clashes, but maybe ok in this _special_ case because `styles.less` is meant as "customize on your own risk". :thinking:

crazymykl:
Making one stylesheet do double duty looks like a recipe for high-specificity rules. I'm not a fan.

Alhadis:
> That would keep things simple and animations should work. It's against the idea of the Shadow DOM that protects against style clashes, but maybe ok in this special case because styles.less is meant as "customize on your own risk". 🤔

@simurai I'm sure we've all seen this by now:

Go away

I'm very much out of the loop regarding web components, but it seems like `::shadow` [isn't facing a very bright future](https://bugs.chromium.org/p/chromium/issues/detail?id=489954). ~~No pun intended~~ (bullshit, I'm hilarious). Last update from a couple days ago:

> [Comment 29](https://bugs.chromium.org/p/chromium/issues/detail?id=489954#c29) by [hayato@chromium.org](https://bugs.chromium.org/u/4036257750/), Sep 30 (2 days ago)
> Status update:
>
> As of now, we do not have any ETA to remove these features.
> 
> Note: These features are between at stage 4 and 5 in a deprecation process:
> http://www.chromium.org/blink#TOC-Launch-Process:-Deprecation

There's no ETA, but sitting between steps 4-5 of a 6-step process of deprecation shouldn't fill you with confidence. This PR is a forward-thinking decision.

Furthermore, it's not immediately clear to every user how syntax elements should be customised. Newcomers to Atom might even be confused as to why editor scopes aren't picking up CSS styling.

50Wliu:
@Alhadis https://github.com/atom/atom/tree/as-deprecate-shadow-dom is arguably an even more forward-thinking solution :).

crazymykl:
Wait, so the shadow DOM was a css-namespace thing and not a performance thing?

Alhadis:
@50Wliu Like I said, I'm *very* much OOTL with web components as a whole. Last I checked, `document.registerElement` was being used for defining custom HTML elements. I later learned that was scrapped by the working group in favour of something else. It's like CSS flex-boxes all over again...

I don't know how volatile web components still are, so I'm similarly uncertain of where the Atom team is going with their implementation of the shadow DOM (perhaps @nathansobo could enlighten me).

All I know is having a separate stylesheet to target syntax highlighting makes sense. Atom differentiates between UI styling and syntax styling in the form of two different theme types, so implementing a syntax-specific user stylesheet would be falling in line with this pattern.

It also obviates having to [purge stylesheets](https://github.com/atom/atom/commit/8144c678a43d198ffab0811ed8caa5e3523c76e9) of deprecated code automatically. One doesn't need to read a Dijkstra letter to know why separation of concerns is always a good thing.

nathansobo:
@Alhadis I think I got a bit fooled by Google's "HTML 5 Rocks" project into thinking web components were far more baked than they really ended up being. After abandoning SpacePen I was on the hunt for a good standard to replace it, and at the time web components seemed like a good candidate. Unfortunately, with the deprecation of the `::shadow` selector and just general usability problems with the shadow DOM, using it in the editor has proven to be a mistake. Custom elements have their own problems because the global namespace is at odds with the versioning guarantees provided by Node's require system.

My current conclusion, which I think is simple enough to be the final word on this, is that a "component" in Atom is just a javascript object with the following interface:

* A constructor that takes a plain properties object
* An `element` property
* An `update` method that takes a plain properties object and returns a promise that resolves when `element` is updated
* A `destroy` method returning a promise that resolves when the `element` is removed

That leaves a lot of freedom in exactly how you map the properties to the component's DOM element and it doesn't require us to use any crazy web platform features that are untested or might change out from under us. We've put together the [etch](https://github.com/atom/etch) library as a lightweight tool for building components that conform to this interface.

This is all tough because until we really put any kind of proposal to the test ourselves, it's difficult to fully endorse it. We're in a bit of a limbo state while we experiment with etch internally. Based on the track record of two previous solutions that didn't work out for the long run, I want to be really sure about this one before sending people in a third direction. So far it's feeling pretty good to me though.

nathansobo:
> One doesn't need to read a Dijkstra letter to know why separation of concerns is always a good thing.

As to this comment, I don't think any one principle can be held up as a universal good in software. There are always trade-offs. Obviously I agree that separation of concerns is a virtue to strive for, which is why I put the shadow DOM in place in the first place. But we also something in the other side of that trade off, which is the simplicity and familiarity of being able to style any element on the DOM in an easy and boring way.

Before the shadow DOM, our style handling code was really simple. Just put the style element in the head. Now we have all this elaborate mechanism for targeting a specific context that users need to understand and all our code paths dealing with styles need to deal with. We're in the process of extracting the editor to a library, and this complexity is a hurdle in our way that would complicate the API dramatically for other users of the library. The shadow DOM also makes our internal implementation of overlays and block decorations really crazy and makes it impossible to target context menu selectors based on syntax elements. 

There are many, often contradictory principles in software and I guess the trick is balancing them. In this case, I'm leaning toward "worse is better", that simplicity of implementation should be favored over correctness of interface. The "worse is better" solution to style leakage is well understood and battle-tested: just accept CSS as a global system and use a prefixing scheme to keep things separate. It's not as pretty as a magical encapsulation boundary, but it's also way simpler. The longer I work on this project, the more I favor simple over other virtues. Hopefully I don't live to regret it.

crazymykl:
I'm very much for the removal of the shadow DOM: it's super confusing for new users to hear they can edit their editor with css, then the editor itself is not effected by the most obvious place to write said css.

Alhadis:
To be honest, I sometimes feel the entire concept of the Shadow DOM is a mistake. I agree with what you've said, and in fact have always adhered to simplicity of implementation, which lead to me coining the term **ISOTRY**: "It's Sometimes Okay To Repeat Yourself" (then again, I'm a guy who prefers hand-writing CSS and natively-supported JS over the toxic complexity of precompilers, so...).

Good points made though, and I'm beginning to see why the Atom team are in a difficult spot at this point in development. Hopefully things pan out smoothly (and simply). =)

crazymykl:
Closing since the Shadow DOM will die in 1.13.0.

PR #12216, opened 2016-07-22
Improve build on AppVeyor

nathansobo:
- Fix asset uploading
- Fix automatic kick-off of builds
- Improve caching
- Add logging to upload tasks

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,5 @@
 version: "{build}"
 
-branches:
-  only:
-    - master
-
 skip_tags: true
 clone_folder: c:\projects\atom
 clone_depth: 10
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -1,9 +1,5 @@
 version: "{build}"
 
-branches:
-  only:
-    - master
```
50Wliu: Is there any way to reconfigure this so that Appveyor only builds on master or on `*-releases` branches?  So that commits on other branches that don't have a PR yet don't get built.

nathansobo: @50Wliu there may be... this was making it so that the *-releases branches weren't building. Maybe we should do `master` *and* a regex pattern for the releases branches. Do you want to open a tiny PR to do so? That's probably what I should have done but my brain was mush from waiting on builds all day.

PR #12226, opened 2016-07-24
Speedup Packages view loading time

ypresto:
Every config updates in config.setSchema causes did-change event to be emitted.
This PR reduces num of events by transact config changes.

image


maxbrunsfeld:
These both make perfect sense to me, and I don't think either change requires a test, since they are both just optimizations. Thanks for looking into this @ypresto!

maxbrunsfeld:
The CI failures are both known issues with our test suite.

PR #12231, opened 2016-07-25
Delete .travis.yml

as-cii:
After #12207, we are going to switch atom/atom to use CircleCI. The web hook remains in-place for existing pull-requests, but from now on we won't be building new PRs on Travis.


Alhadis:
So long, pal...

Die, Travis, Die

PR #12235, opened 2016-07-26
Experiment: import APM into the atom repo

maxbrunsfeld:
⚠️  This may be a terrible idea ⚠️ 

This adds `apm` (with all of its history) into this repo, under the `apm` directory. The idea is that `apm` should just be a part of Atom, rather than its own node module that needs to _find_ Atom on your hard drive in order to work properly.

/cc @nathansobo


nathansobo:
Hmm I like the idea of running apm via Atom but I still think it should be its own module in npm.

maxbrunsfeld:
Nothing to see here, folks.

PR #12237, opened 2016-07-26
Cleanup activationHookSubscriptions when deactivating package

zertosh:
Without this fix, the included test would fail with:

```
$ ./out/Atom.app/Contents/MacOS/Atom --test --resource-path="$(pwd)" -- spec/package-manager-spec.coffee
(electron) submitURL is now a required option to crashReporter.start
Error Downloading Update: Could not get code signature for running application
---------------------------------.F.---------------------------------------

PackageManager
  ::activatePackage(id)
    when the package metadata includes `activationHooks`
      it does not double register activation hooks when deactivating and reactivating
        Expected 2 to be 1.
          at .<anonymous> (/Users/asuarez/src/atom/atom/spec/package-manager-spec.coffee:444:49)
        Expected 3 to be 2.
          at .<anonymous> (/Users/asuarez/src/atom/atom/spec/package-manager-spec.coffee:455:49)
```

cc: @hansonw

### DIFF
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -431,6 +431,27 @@ describe "PackageManager", ->
         runs ->
           expect(Package.prototype.requireMainModule.callCount).toBe 1
 
+      it "does not double register activation hooks when deactivating and reactivating", ->
+        expect(mainModule.activate.callCount).toBe 0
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
⚡ Great test. Thanks for tracking this down!

PR #12258, opened 2016-07-29
Add JS style guide to the Contributing guide

mnquintana:
This PR adds a section for JavaScript style to the Contributing guide, based on feedback from https://github.com/atom/about/pull/35.

👀 @atom/feedback @nathansobo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,7 @@ These are just guidelines, not rules, use your best judgment and feel free to pr
 
 [Styleguides](#styleguides)
   * [Git Commit Messages](#git-commit-messages)
+  * [JavaScript Styleguide](#javascript-styleguide)
   * [CoffeeScript Styleguide](#coffeescript-styleguide)
   * [Specs Styleguide](#specs-styleguide)
   * [Documentation Styleguide](#documentation-styleguide)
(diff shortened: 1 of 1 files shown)


nathansobo:
Thanks @mnquintana. Good to get this started. Luckily most of it is in standard style.

mnquintana:
@nathansobo So I was just thinking – one advantage of the object spread syntax `{…someOtherObj}` over `Object.assign()` is that it’s much more foolproof. Take, for example:

```js
let state = {
  thing: true
}

let newState = Object.assign(state, {
  thing: false
})

state.thing !== newState.thing // returns false
```

If you forget to provide a new object as the first argument, then `Object.assign()` will both return
this new merged object *and* mutate the first object passed to it, which is almost never what we want to do.

On the other hand, this will always return a new object so we don't have to worry about unintended
mutations:

```js
let state = {
  thing: true
}

let newState = {
  ...state,
  thing: false
}

state.thing !== newState.thing // returns true
```

What do you think?

BinaryMuse:
I personally *greatly* prefer object spread, and since it's used heavily in JSX/Etch, I'd personally recommend it for consistency as well. Just my 2¢ :)

BinaryMuse:
It's also much less verbose when doing deeply nested things:

```javascript
{
  ...state,
  subobj: {
    ...state.subobj,
    key: "val"
  }
}
```

nathansobo:
Cool, let's use it then.

PR #12262, opened 2016-07-29
Delete Obsolete Code

as-cii:
This pull-request gets rid of `RowMap` along with its specs (after the transition to `DisplayLayer` we don't need this object anymore). It also deletes `spec/random-editor-spec.coffee`, because the tested behavior is already extensively covered in text-buffer (specifically, in `DisplayLayer` tests).

/cc: @atom/core


nathansobo:
Oops. Thanks for catching this. 😬

as-cii:
:fire: :fire: :fire:

PR #12276, opened 2016-08-01
Add ability to see passed specs via tooltips

pbiggar:
This is my first PR against atom, so let me know if I've gotten anything wrong.

It's not currently possible to see what specs pass when you ctrl-alt-cmd-P. You see that some tests pass (via the red and green dots), but you can't see which ones (except by process of deduction by looking at which fail)

This PR adds the description of the spec, as well as all it's parent suites, to a tooltip so you can see what's passing and failing on mouseover.

I couldn't find tests for atom-reporter. Given the simplicity of the change and the overhead of learning how to write tests for this component, it didn't seem worthwhile. If there are already specs for it, please point me to them and I'll be happy to add a quick test.

Before: (no tooltip)
screenshot 2016-08-01 10 24 12

After: (tooltip, showing suite and test name)
screenshot 2016-08-01 11 23 07

### DIFF
--- a/spec/atom-reporter.coffee
+++ b/spec/atom-reporter.coffee
@@ -197,6 +197,21 @@ class AtomReporter
     time = "0#{time}" if time.length < 3
     @time.textContent = "#{time[0...-2]}.#{time[-2..]}s"
 
+  specTitle: (spec) ->
+    parentDescs = []
+    s = spec.suite
+    while s
+      parentDescs.unshift(s.description)
+      s = s.parentSuite
+
+    suiteString = ""
+    indent = ""
+    for desc in parentDescs
+      suiteString += indent + desc + "\n"
+      indent += "  "
+
+    "#{suiteString} #{indent} it #{spec.description}"
+
   addSpecs: (specs) ->
     coreSpecs = 0
     bundledPackageSpecs = 0
@@ -204,6 +219,7 @@ class AtomReporter
     for spec in specs
       symbol = document.createElement('li')
       symbol.setAttribute('id', "spec-summary-#{spec.id}")
+      symbol.setAttribute('title', @specTitle(spec))
       symbol.className = "spec-summary pending"
       switch spec.specType
         when 'core'


50Wliu:
I'm :+1: on this.  Any chance you can upload an image of this in action?

pbiggar:
@50Wliu Cool, thanks. Just added an image.

pbiggar:
I looked at the tests on circleci (heh), and it looks like an intermittent failure in a different component. Tests pass on the other services.

50Wliu:
In your picture it looks like the two describes are flipped?  Shouldn't it be "paredit-el" first and then "Basic Insertion Commands tests"?  Also, the actual spec is missing the leading `it`,

pbiggar:
@50Wliu I forgot to reverse the order when traversing the parents. And great suggestion on adding "it" - I'm new to BDD so it hadn't occurred to me.

Code and "after" screenshot are updated.

50Wliu:
@atom/feedback any comments?

spec/atom-reporter.coffee
```diff
@@ -197,13 +197,27 @@ class AtomReporter
     time = "0#{time}" if time.length < 3
     @time.textContent = "#{time[0...-2]}.#{time[-2..]}s"
 
+  specTitle: (spec) ->
+    s = spec.suite
+    parentDescs = while s = s.parentSuite
```
mnquintana: This loop's a bit confusing – mind refactoring this so all these assignments aren't inlined?

maxbrunsfeld: It looks like the innermost suite (`spec.suite`) will not have its `description` added to the `parentDescs` array. Is that true / intentional? I'm not too familiar with how the specs and suites are represented in Jasmine, but I would think that `spec.suite` would correspond to the innermost `describe` call.

maxbrunsfeld:
⚡  This is a great improvement; thanks so much for adding it @pbiggar! And nice to see you again, also. I left one question, but otherwise I'd love to 🚢  this.

maxbrunsfeld:
> I couldn't find tests for atom-reporter. Given the simplicity of the change and the overhead of learning how to write tests for this component, it didn't seem worthwhile.

Yeah, there are no tests on it currently; thanks for checking.

spec/atom-reporter.coffee
pbiggar: You are correct. I discovered the same thing in the refactor :)

maxbrunsfeld:
This will be released in Atom v1.11.

pbiggar:
Awesome, thanks!

PR #12285, opened 2016-08-02
Add null checks for ShellOption.isRegister/update

Ingramz:
I am getting `Uncaught TypeError: Cannot read property 'value' of null` at the isRegistered function.

see https://github.com/fresc81/node-winreg/blob/master/lib/registry.js#L649

cc @damieng


damieng:
Hi.

Thanks for this. The null checks and some other things (e.g. beta checking) are actually in our 1.10-releases branch already and I just need to merge them back to master.  I'll do that now.

PR #12295, opened 2016-08-02
Continued: replace OSX window title bar with custom title-bar

nathansobo:
Picks up from #11790


nathansobo:
Fixed the only failure on CI, so going ahead and merging.

PR #12300, opened 2016-08-03
Upgrade Electron

### LINKED ISSUE #12591: Mouse scroll misbehaves on Linux after alt-tabbing
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?

### Description

With a window that partially or fully overlaps Atom's window, if that other window is on top and mouse scrolls are made on the part that overlaps Atom, and then one alt-tabs directly back into Atom and attempts to mouse scroll, it "re-plays" all those scrolls made and applies them along with the mouse scroll actually made.

### Steps to Reproduce

1. Open Atom, preferably with a very long file, open around the middle, to make it more obvious
2. Open any other window and place it on top of Atom. It doesn't actually matter if it has any scrollable content at all.
3. Focus the other window and mouse scroll a lot over it in the same direction. Say, downwards.
4. Alt-tab to Atom. Make sure not to move your mouse, at least not outside the boundaries of the Atom window.
5. Perform a single mouse scroll "click" in the opposite direction, i.e. if you scrolled downwards a lot in step 3, now scroll upwards once
6. Repeat steps 3 to 5, swap directions around a bit

**Expected behavior:** The content scrolls up and down accordingly, only as far as I expect it to

**Actual behavior:** The first time you do step 6, no scrolls happen. Afterwards, the content scrolls by as much as you've scrolled in step 3, and in that direction, even though in step 6 you scrolled once.

[I've recorded a video demonstrating this](https://my.mixtape.moe/nrpgkw.webm). Note that each time I alt-tab to Atom, I do a single scroll 'click', but the content jumps by a lot each time.

Switching to Atom by any method other than alt-tab (i.e. clicking on the Atom window to raise it, clicking the icon on any dock or taskbar, etc.) does not trigger this behaviour. Only alt-tabbing, and only if the mouse cursor doesn't leave the boundaries of the Atom window right before or after alt-tabbing.

Atom 1.9 also had the same problem, which I worked around by downgrading electron to 0.36.12 at compile time. This workaround makes Atom 1.10 fail to start (unrelated issue), so downgrading electron is no longer an option for me.

[This Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=608246) that I've also reported is probably related, but now it happens even without xf86-xinput-libinput or even libinput installed.

### Versions

~~~
Atom    : 1.10.0
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0
~~~

Atom installed from this [AUR package](https://aur.archlinux.org/packages/atom-editor/)
OS: Arch Linux 64-bit, fully up-to-date.

thomasjo:
The purpose of this PR is to begin the process of upgrading Electron to the latest release. As of right now, there are a few things that need to be dealt with before that can happen.

The primary problem is that as of Electron v1.1.0, Node.js was upgraded to v6 and that causes some issues. ~~I believe (and hope) that the only major issue is the usage of older version of graceful-fs in a few of our dependencies. Once those have been resolved, I think we'll be able to sort out the remaining issues in Atom core without too many problems.~~ OK, there are more problems than first anticipated — I'll keep updating the task list below as I identify (and fix) problems.

There is another concern though, since Electron v1 removed some [deprecated module `require` bits](https://github.com/electron/electron/pull/5373).

I propose that in this PR, we perform the upgrade process in stages, one minor version bump at a time, to minimize the number of issues and hopefully keep it reasonable.

---
- [x] Upgrade to v1.0.x
- [x] Upgrade to v1.1.x
  - [x] Resolve breaking changes to `path` module. See https://github.com/atom/atom/pull/12300#issuecomment-241507011.
    - [x] tree-view (atom/tree-view#926)
    - [x] autocomplete-plus (atom/autocomplete-plus#753)
    - [x] Add deprecated shims, so that community packages don't break
- [x] Upgrade to v1.2.x
  - [x] Fix use of removed `ATOM_SHELL_INTERNAL_RUN_AS_NODE` env variable in `exports/atom.js`
  - [x] Fix beforeunload handler to conform to new contract
  - [x] Open issue on atom-package-deps (https://github.com/steelbrain/package-deps/issues/57)
- [x] Upgrade to v1.3.x
  - [x] Upgrade oniguruma to work with Node 6
  - [x] Upgrade marker-index to work with Node 6
  - [x] Investigate why `atom-pane-resize-handles` are positioned incorrectly, making it impossible to resize panes with the mouse.
- [x] Add deprecated shims for legacy electron APIs (where feasible)
- [x] Open issues on packages that use removed Electron APIs that we aren't able to shim (because they involve methods called on remote proxy objects).
  
  APIs:
  - `remote.require`
  - `WebContents.prototype.loadUrl` (renamed to `loadURL`)
  - `WebContents.prototype.getUrl` (renamed to `getURL`)
  
  Packages:
  - [x] atom-cos-studio (https://github.com/MDevelopmentGroup/Atom-COS-Studio/issues/35)
  - [x] atom-force (https://github.com/mikegedelman/atom-force/issues/3)
  - [x] atom-pronama-chan (https://github.com/hurumeki/atom-pronama-chan/issues/1)
  - [x] dev-tools-themes (https://github.com/travs/dev-tools-themes/issues/2)
  - [x] elasticsearch-client (https://github.com/KunihikoKido/atom-elasticsearch-client/issues/9)
  - [x] impress (https://github.com/kaitoy/impress/issues/1)
  - [x] jibo-sdk (https://github.com/jiborobot/jibo-sdk/issues/11)
  - [x] julia-client (https://github.com/JunoLab/atom-julia-client/issues/217)
  - [x] markdown-preview-enhanced (https://github.com/shd101wyy/markdown-preview-enhanced/issues/115)
  - [x] mavensmate-atom (https://github.com/joeferraro/MavensMate-Atom/issues/334)
  - [x] mocha-runner (https://github.com/tswaters/atom-mocha-runner/issues/10)
  - [x] nitrogen (https://github.com/nitrous-io/nitrogen-dist/issues/43)
  - [x] tree-view-bleeding (https://github.com/forivall/atom-tree-view-bleeding/issues/2)

Fixes #12591


50Wliu:
In case any package authors are looking at this, here's a list of packages that need to be updated:
https://gist.github.com/50Wliu/ef22d45854415842d9c59f562eca99bc

mnquintana:
We should set `process.traceDeprecation = true` while we're testing this to make sure we haven't missed changing any deprecated stuff. (cf. http://electron.atom.io/blog/2015/11/17/electron-api-changes)

hultberg:
What is the status on this? Excited to see work on this!

thomasjo:
@hultberg We're currently blocked by https://github.com/less/less.js/pull/2953. Hopefully once the Less.js upgrade is dealt with, we can continue the stepwise upgrade towards the latest release of Electron.

thomasjo:
Updating Electron to v1.1.x (and newer) is currently blocked on path assertions introduced in Node v6. Essentially we are no longer allowed to pass `undefined` to any function in the `path` module. Since this affects not only Atom core, but all packages, we will have to resolve this somehow.

The "easy" option is to _monkey patch_ the `assertPath` function in `path.js`. The hard option is to ensure **all** instances of `undefined` are replaced with e.g. `""` throughout the entire codebase, including community packages.

Deviating from Node proper by patching in support for `undefined` does not feel right, but I believe it's the only viable option. If someone has an alternative approach, I'm all :ear::ear:

joefitzgerald:
@thomasjo This is unavoidable; as node progresses. I don't think we should patch path functions, as this perpetuates the issue. This was done for a reason, and we should fix any area in core that is affected. 

I would advocate for ensuring that we deal with packages that fall afoul of this via:
* Adding some logic that adds context about how to fix issues that are filed as a result of catching this kind of error
* Downloading all published packages and identifying any use of path functions; you could then potentially open an issue warning of this upcoming change, giving the package author time to fix it while a) these changes are in master and b) these changes are in beta, and prior to c) these changes being in stable
* Optionally creating a new channel for electron-latest ("alpha"?) so that we ease the burden of building from source for people who just want to fix their package and don't necessarily want to build from master

thomasjo:
@joefitzgerald As we discussed elsewhere, I agree. We should fix this properly, and as mentioned, I believe this will require us to bump Atom's major version since we will likely Break The World™ 🎱

I've started fixing the issues, and will keep updating the PR description as I identify (and fix) problematic code in core and bundled packages. As of writing this, I will at least have to patch tree-view and autocomplete-plus, but I suspect there are others.

Opening PRs on community packages will have to wait, but if someone wants to start testing this PR against their community packages and resolving issues there, that would be amazing 💖

maxbrunsfeld:
@simurai It seems that on this new version of Chrome (`52.0.2743.82`), the `atom-pane-resize-handle` elements are not being positioned properly in the flow. I think this is related to https://github.com/atom/atom/commit/4980ce30a98615373ca17125a22e168bbb4ef44c and https://github.com/atom/one-dark-ui/commit/82271d5e5749aacb34298a139ad9cd7d1b74f0bd.

Could you build this branch when you get a chance, and investigate if there's another way to achieve the desired look of the resize handles?

simurai:
@maxbrunsfeld Yes, seems Chromium 52 [changed](https://developers.google.com/web/updates/2016/06/absolute-positioned-children?hl=en) how they handle absolutely positioned flexbox items. Should be fixed by using an extra pseudo-element.

The One + Atom Material themes still need a separate fix for the horizontal handles. 🔜

nathansobo:
@maxbrunsfeld: @thomasjo reports that we're getting IndexDB errors when `Atom` and `Atom Beta` are opened side-by-side. I wonder if something about the storage directories isn't being set correctly.

thomasjo:
@maxbrunsfeld, @nathansobo: Yeah I can consistently repro this on my machine; start `atom-beta` from an arbitrary directory, then subsequent launches of `atom` (with beta still running) causes a IndexDB error.

_users_thomasjo_scratch_cuda_cusolver-qr

nathansobo:
@maxbrunsfeld @thomasjo Apparently this happens on stable/beta as well. It's not related to this branch.

s0ber:
@thomasjo Please consider, that `Chrome v49` has a very nasty SVG rendering speed downgrade (when rendering text nodes):
https://bugs.chromium.org/p/chromium/issues/detail?id=589525

I am developing a diagram editor (which is an Atom plugin), and when Atom updated to `1.9.9` it became sooo slow. There are O(n^2) computations in place, and now some interactions take seconds to complete, while previously there were milliseconds.

It would be really nice to update Chrome version to more performant one as soon as possible.

nathansobo:
@thomasjo Looks like we're failing an integration test on this branch.

thomasjo:
@nathansobo The failing intergration test seems to be caused by [atom-application.test.js#L192](https://github.com/atom/atom/blob/d34e09292e8bed3b7aa6eb19affebdec7ec9f5a0/spec/main-process/atom-application.test.js#L192) not actually opening the specified `new-file`, which means no text is inserted into the empty file, and subsequently the re-launched instance does not contain the persisted file and the assertion fails.

Unfortunately I did not find time to track down the root cause. The object constructed by `parseCommandLine` does seem correct, but it might be worth investigating. Hopefully my simple investigation will guide you or someone else in the right direction.

nathansobo:
Sounds good. We'll take a look next week. Looking forward to being on a new electron and fixing international keyboards finally.

thomasjo:
@nathansobo Looks as if the failing integration test is due to some type of race condition. I've yet to figure out exactly what the problem is, but I'm guessing state in IndexedDB isn't persisted by the time the window is closed, or something along those lines.

Will continue trying to figure out what's changed, and how to fix it. Good thing is, this only seems to be a spec problem though, since putting in a slight delay in the spec, "fixes" the problem.

spec/main-process/atom-application.test.js
```diff
@@ -191,7 +191,7 @@ describe('AtomApplication', function () {
       const atomApplication = buildAtomApplication()
       const window1 = atomApplication.launch(parseCommandLine([path.join(tempDirPath, 'new-file')]))
       await focusWindow(window1)
-      await new Promise(resolve => setTimeout(resolve, 1000))
+      await new Promise(resolve => setTimeout(resolve, 2000))
```
thomasjo: @nathansobo This is terrible, and fragile. Hoping you or someone else has a better idea. From what I can tell, this test may always have been brittle, we just never noticed anything because Atom was too slow? 🎱 

/cc @atom/feedback

as-cii:
I took some time today to investigate the failing test and it seems like it might be related to IndexedDB. Creating IndexedDB transactions inside a `beforeunload` handler is considered undefined behavior and could abort the transaction without saving any data (https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#Warning_About_Browser_Shutdown). 

The flaky spec is relying on closing the window to save the project state (i.e. because when we close the window, we call `unloadEditorWindow` which in turn calls `atom.saveState`) but it seems like this new version of Electron is either too fast to tear down the renderer process or simply the behavior has changed with respect to transactions initiated in a `beforeunload` event handler.

I don't think we can simply bypass this failure by adding a timeout, because it seems to be pointing at a potential bug which can cause problems to users with slower machines (or that simply happen to have a busy CPU when closing Atom). Finding a good solution to this problem is also tricky because of the synchronous nature of the `beforeunload` event and the asynchronous API of IndexedDB. An idea would be to:

1. Cancel the close initiated by the user.
2. Wait until `AtomEnvironment.prototype.saveState` is resolved (or rejected).
3. Programmatically close the window from the renderer process.

This would work great for closing a single window, but when shutting down the entire application 3) wouldn't be sufficient because we'd need to call `app.quit` as well. So, alternatively we might:

1. Cancel the close initiated by the user.
2. Wait until `AtomEnvironment.prototype.saveState` is resolved (or rejected).
3. Notify the main process that a window can be closed (`emit 'can-close'`).
4. In the main process:
  1. If the application was quitting, wait until all the windows notify that they can be closed, then call `app.quit`.
  2. If the application  was not quitting, close only the window for which we have a received a `can-close` event.

I worry a little bit that keeping additional state around to handle the above logic could, in certain cases, cause the application to never quit or to have dangling windows (i.e. in the case of exceptions, or other less-predictable scenarios).

Therefore, I can think of three alternatives:

1. If exists, replace IndexedDB with some other database that provides both a synchronous and an asynchronous API (we'd use the former only in the `beforeunload` event handler).
2. Instead of persisting additional state only on exit, save *everything* (including markers, etc.) [while the application is open](https://github.com/atom/atom/tree/4bcdbf4d2f26f7fb91ba5030a4366d978a33421f/src/atom-environment.coffee#L244-L252). The downside here could be that we potentially slow down the UI thread when there's a very large number of markers to serialize.
3. I couldn't reproduce the problem manually, so we might ignore this and worst-case scenario users won't have their markers in the history after reopening Atom.

What do you think?

thomasjo:
I'm leaning towards option #2 or #3. As stated earlier, I've also been unable to reproduce this in a real world scenario.

as-cii:
For posterity, in #12619 we ended up implementing a slight variation of the routine proposed in the previous comment:

> 1. Cancel the close initiated by the user.
> 2. Wait until AtomEnvironment.prototype.saveState is resolved (or rejected).
> 3. Notify the main process that a window can be closed (emit 'can-close').
> 4. In the main process:
>   1. If the application was quitting, wait until all the windows notify that they can be closed, then call app.quit.
>   1. If the application was not quitting, close only the window for which we have a received a can-close event.

thomasjo:
Rebased and removed my "hacky" attempts at resolving the failing integration test. Hopefully everything should be 💚 now with @as-cii's fix in #12619 🎉

as-cii:
Thanks so much @thomasjo, @maxbrunsfeld and everyone else involved in this! ❤️

v3ss0n:
Yes! Yes!!! I am thinking about using nvm and forget worrying about system node veriosn (6.x vs 4.x) is that a good idea?

Issue #12313, opened 2016-08-03
Atom Beta 1.10.beta1 : Indexing Project never completes.


rwatts3:
Since the release this morning 08/03/2016 when trying to search files with `ctrl + p` Atom indexing never finishes and files are not found. Always says 0 and indexing project.
I believe this is related to the `fuzzy-finder` package.

Ben3eeE:
https://github.com/atom/fuzzy-finder/issues/205?

50Wliu:
@rwatts3 If this is not a duplicate of atom/fuzzy-finder#205, please reopen this issue on the fuzzy-finder repo.  Thanks.

rwatts3:
I'm not sure that it is the same because I didn't notice this issue until I updated. 
I downgraded to beta0 and it wen't away. 

Very similar symptoms but something in beta1 is causing it to happen.

Ben3eeE:
Can confirm that this is happening to me in beta1 on windows 10. @rwatts3 please open a new issue in the [fuzzy finder repo](https://github.com/atom/fuzzy-finder/issues/new)

rwatts3:
will do

PR #12317, opened 2016-08-03
Don't update the NODE_PATH when updating the environment on launch

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/12313
Fixes https://github.com/atom/fuzzy-finder/issues/205
Refs https://github.com/atom/atom/pull/12006 (solves same problem, but with a different approach)

Both `NODE_ENV` and `NODE_PATH` affect Atom's own behavior, so when loading environment variables from the main process, Atom needs to control these two variables explicitly.


maxbrunsfeld:
Thanks to @rahatarmanahmed for explaining the problem so clearly in #12006.

src/update-process-env.js
```diff
@@ -0,0 +1,49 @@
...
+
+    for (let key in envToAssign) {
+      if (!ENVIRONMENT_VARIABLES_TO_PRESERVE.has(key)) {
+        process.env[key] = envToAssign[key]
+      }
+    }
+  }
+}
+
+function getEnvFromShell () {
```
steelbrain: @maxbrunsfeld would you guys be interested in replacing this homegrown function with [`consistent-env`](https://www.npmjs.com/package/consistent-env). It supports both sync/async (no startup delay) and a wide variety of shells and is well tested and robust

joefitzgerald: :-1: on the basis that it's not worth taking on additional dependencies to replace this small piece of code (which has since evolved to be more simple and address more shells and more platforms).

I write a lot of go code, so to draw inspiration from one of the creators of go:

> "A little copying is better than a little dependency" 
> - [Rob Pike](http://go-proverbs.github.io)

This is cogent, given parables like `left-pad` (not to suggest consistent-env is in any way like left-pad).

steelbrain: Tell me if I'm wrong but I personally believe that a project should be focused and Atom's mission is to be a Text Editor not an env patcher. The reason I suggest this time and time again is that the current method is not reliable and we still have to patch `PATH` in all of the linter providers, if we can get it done in the core it'll make all those linter providers have one less dependency.

While you can copy a little to **save** a dependency, but if you keep doing this you'll end up with a bunch of bloatware that will be hard to maintain. and we can already see that Atom's maintaining power is being wasted into something that's already available as a module.

https://github.com/atom/atom/pull/12562 https://github.com/atom/atom/pull/12396 https://github.com/atom/atom/pull/11436 https://github.com/atom/atom/pull/12378 https://github.com/atom/atom/pull/12317 https://github.com/atom/atom/pull/12028 https://github.com/atom/atom/pull/11054 and more

joefitzgerald: I think your last statements creating false dichotomies. It is possible to make Atom a great text editor without taking on dependencies in all areas where you could depend on something. And having code in this specific area does not make Atom "bloatware".

If you believe the patching behavior should differ in some way you should file an issue substantiating the ways in which it should differ. The PATH patching you have to do for linters is a good example of a potential issue you could open. I don't currently understand *why* you need to do this, so opening an issue would enhance our shared understanding of your needs.

In general, further discussion of this in this line comment trail seems counter productive as the code has changed materially.

lee-dohm: > Tell me if I'm wrong but I personally believe that a project should be focused and Atom's mission is to be a Text Editor not an env patcher.

By that line of reasoning, Atom shouldn't have any env patching code in it at all and we still shouldn't accept your module as a dependency.

This conversation is done, @steelbrain. I've discussed it with the team, we aren't going to be accepting your module as a dependency, stop suggesting it.

nathansobo: @steelbrain As someone who is ultimately responsible for maintaining this behavior core, I find it easier to be able to browse to a specific chunk of code that performs this logic and immediately see what's going on rather than needing to dig it up in a dependency. It just doesn't seem complex enough to warrant it. That isn't always the case, but for this concern it is. It's a subjective call but that's my preference.

I would be very interested in accepting a specific PR addressing specific issues in the current code, however.

PR #12327, opened 2016-08-04
Handle double-click on custom title bar

Zirro:
Unlike the native version, the custom title bar implemented in #11790 doesn't respond to double-clicks. This pull request implements the double-click behaviour in a manner that is consistent with the native title bar.

This is my first contribution to Atom, and the first time I've written CoffeeScript. I'd be grateful for feedback if there's something I ought to have done differently.

In case this pull request should include a test, I would appreciate some help in putting it together 🙂

### DIFF
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -136,8 +136,8 @@ class ApplicationDelegate
   getPrimaryDisplayWorkAreaSize: ->
     remote.screen.getPrimaryDisplay().workAreaSize
 
-  getAppleActionOnDoubleClick: ->
-    remote.systemPreferences.getUserDefault("AppleActionOnDoubleClick", "string")
+  getUserDefault: (key, type) ->
+    remote.systemPreferences.getUserDefault(key, type)
 
   confirm: ({message, detailedMessage, buttons}) ->
     buttons ?= {}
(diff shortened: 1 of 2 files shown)


Zirro:
This code is the first use the systemPreferences API, but likely not the last. I initially opted for an explicit getAppleActionOnDoubleClick()-method, but this seems excessive for a one-time use. Instead I've changed it to getUserDefault(key, type), reflecting the Electron API and opening it for future use.

nathansobo:
This looks great! It would be nice to get some tests on the `dblclickHandler` in `title-bar-spec.coffee` just to make sure this code doesn't regress in the future. You can give it a shot or I'll try to get to adding them soon so we can merge this.

nathansobo:
Upon further thinking, there's really not much to test here. By the time we mock the application delegate we're basically making empty assertions about the code we wrote in the other file. We could actually interact with these APIs to maximize the test window, but that's pretty obnoxious. I'm going to merge this as is.

nathansobo:
By the way, thanks for this addition. It's great to see an immediate follow on to @brumm's great work and I think people are going to love the ability to style their title bar.

Zirro:
I'm glad that I could contribute. Indeed, I hadn't realized how much I was missing a dark title bar to go with the rest of my theme until I saw it in the application. Now it would be hard to go back. Thanks @brumm!

paddotk:
Does this mean the issue was fixed? I don't really get how this linking stuff works in Github.

Arcanemagus:
The changes from this PR were released in Atom v1.11.0-beta0, if you are using custom title bars and double click isn't working please file a new issue @paddotk, thanks!

PR #12342, opened 2016-08-05
Don't delete package metadata during package slug compilation

nathansobo:
This saves space, but leads to surprises when package code tries to read these files for purposes of introspection. For example, it was causing us not to report exceptions from the exception-reporting package. The .1% of the app bundle size it saves doesn’t seem worth the potential for problems.

/cc @lee-dohm @kattrali


maxbrunsfeld:
The build failure was a flaky TextEditorComponent spec.

PR #12348, opened 2016-08-06
🎨 Remove duplicate "Note:"

Zirro:
Just a small touchup - I noticed that there was a duplicate "Note:" in the custom title bar settings. I also took the opportunity to use HTML5s preferred `<br>` rather than `<br />`, and changed "title-bar" to "title bar" for consistency.


50Wliu:
Thanks!

PR #12350, opened 2016-08-07
Rewrite exports/atom.coffee as JS

mnquintana:
This PR rewrites Atom's exports file as JS – it's a pretty small file, so there shouldn't be anything too weird going on here. 😉 

👀 @atom/feedback

PR #12371, opened 2016-08-09
Transpile exports/**/*.js when running script/build

as-cii:
This fixes a regression introduced in #12350 that was preventing Atom from starting after running `script/build`. Specifically, we were still transpiling coffee script files located in the `exports` directory, ignoring all the javascript assets inside of it, thereby causing the output application to not contain the Atom exports file and to throw an exception on startup. To fix this, we have replaced coffee script transpilation with babel transpilation for `export/**/*.js` files.

/cc: @maxbrunsfeld @mnquintana @CraZySacX

### DIFF
--- a/build/Gruntfile.coffee
+++ b/build/Gruntfile.coffee
@@ -84,7 +84,6 @@ module.exports = (grunt) ->
         'src/**/*.coffee'
         'spec/*.coffee'
         '!spec/*-spec.coffee'
-        'exports/**/*.coffee'
         'static/**/*.coffee'
       ]
       dest: appDir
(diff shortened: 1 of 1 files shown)


as-cii:
The build is ❌ because of some flaky specs, so I'll go ahead and merge this.

maxbrunsfeld:
:+1: Thanks for fixing this @as-cii!

maxbrunsfeld:
Ah that accounts for the persistent integration test failures on master yesterday. I couldn't reproduce them locally, but that was probably because I didn't do a full build.

Issue #12333, opened 2016-08-05
grammar-used activation hook regression


nwolverson:
### Prerequisites

* [ ] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)? **N/A**
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

### Description

My package has a `grammar-used` activation hook, as per #8313 it is triggered on opening a new file but not on opening a previous session.

I have observed this working correctly at some point in an earlier version, but can't say how long ago.

### Steps to Reproduce

(as per #8313)

1. Install language-purescript and ide-purescript
2. Open a file with `.purs` extension - observe package starts
3. Close atom
4. Re-open atom (so file is still open)

**Expected behavior:** [What you expected to happen]
*  Package is started

**Actual behavior:** [What actually happened]

*  Package does not start
*  Package can be observed to neither be started nor activated (via console.log on startup)
*  Opening a new file which was not already open will start the package as normal (but not switching to one of the existing files)

### Versions

apm 1.10.0
atom 1.9.4 (also verified on 1.10.0-beta1)

OSX 10.9.4

yitzchak:
I think this is showing up in thomasjo/atom-latex#200

mrdawn:
Same issue in latex, packages not work with remember tabs. I think it's atom error, not package error

zertosh:
We hit this issue in Nuclide too- in a hastily move, I ended up removing all of the `grammar-used` activation hooks (facebook/nuclide@f413cbfc39341f7728767e8611d81fb960c16cd4) for Atom 1.9.0 compat.

[...]

PR #12375, opened 2016-08-09
Trigger grammar-used hooks no matter how text editors are created

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/12333

### DIFF
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -88,6 +88,7 @@ class Workspace extends Model
   subscribeToEvents: ->
     @subscribeToActiveItem()
     @subscribeToFontSize()
+    @subscribeToAddedItems()
 
   consumeServices: ({serviceHub}) ->
     @directorySearchers = []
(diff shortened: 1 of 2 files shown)


src/workspace.coffee
```diff
@@ -160,6 +161,13 @@ class Workspace extends Model
       @activeItemSubscriptions.add(titleSubscription) if titleSubscription?
       @activeItemSubscriptions.add(modifiedSubscription) if modifiedSubscription?
 
+  subscribeToAddedItems: ->
+    @onDidAddPaneItem ({item, pane, index}) =>
+      if item instanceof TextEditor
+        grammarSubscription = item.observeGrammar(@handleGrammarUsed.bind(this))
+        item.onDidDestroy -> grammarSubscription.dispose()
+        @emitter.emit 'did-add-text-editor', {textEditor: item, pane, index}
+
```
maxbrunsfeld: It seems like we need to look at every editor being added to the workspace in order to make sure we always trigger the `grammar-used` hooks.

src/workspace.coffee
```diff
@@ -383,8 +391,7 @@ class Workspace extends Model
   #
   # Returns a {Disposable} on which `.dispose()` can be called to unsubscribe.
   onDidAddTextEditor: (callback) ->
-    @onDidAddPaneItem ({item, pane, index}) ->
-      callback({textEditor: item, pane, index}) if item instanceof TextEditor
+    @emitter.on 'did-add-text-editor', callback
```
maxbrunsfeld: I figured we should avoid a duplicate `instanceof TextEditor` check, so I moved the emission of the `did-add-text-editor` event into the callback above.

nathansobo:
This looks good to me. Basically there are ways editors can be added to the workspace without going through `openTextFile`. Thanks for fixing this Max.

Issue #12280, opened 2016-08-02
Uncaught (in promise) TypeError: Cannot read property 'bufferId' of null


zertosh:
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

`ctrl+alt+cmd+L` (or `cmd+R` with the dev tools open) with an unsaved editor will reload to an empty window because of:

```
Uncaught (in promise) TypeError: Cannot read property 'bufferId' of null
module.exports.TokenizedBuffer.deserialize  @ /Applications/Atom.app/Contents/Resources/app.asar/src/tokenized-buffer.js:51
module.exports.TextEditor.deserialize @ /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:98
module.exports.DeserializerManager.deserialize  @ /Applications/Atom.app/Contents/Resources/app.asar/src/deserializer-manager.js:45
(anonymous function)  @ /Applications/Atom.app/Contents/Resources/app.asar/src/pane.js:37
module.exports.Pane.deserialize @ /Applications/Atom.app/Contents/Resources/app.asar/src/pane.js:36
module.exports.DeserializerManager.deserialize  @ /Applications/Atom.app/Contents/Resources/app.asar/src/deserializer-manager.js:45
[...]

PR #12377, opened 2016-08-09
Stringify/parse temporary window state to avoid nulling out references

nathansobo:
Fixes https://github.com/atom/atom/issues/12280

If the object passed via IPC contains any keys of the same reference, all but one of these keys will be nulled out. Converting to/from a string avoids this problem.

@atom/electron-maintainers is it expected that objects passed over IPC would have fields nulled out if they point to the same object?

### DIFF
--- a/spec/application-delegate-spec.js
+++ b/spec/application-delegate-spec.js
@@ -1,0 +1,19 @@
+/** @babel */
+
+import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
+import ApplicationDelegate from '../src/application-delegate'
+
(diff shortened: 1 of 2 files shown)


kevinsawicki:
> is it expected that objects passed over IPC would have fields nulled out if they point to the same object?

This behavior changed a bit 1.2.7 via https://github.com/electron/electron/pull/6442

It will not only null out circular references but in pre-1.2.7 it would null out any duplicate references.

nathansobo:
Circle CI is passing with some spurious failures that we really need to 🔨 . Gradually digging out here.

as-cii:
Awesome investigation here! ✨

PR #12393, opened 2016-08-11
Remove config option `editor.backUpBeforeSaving`

as-cii:
After #11828, Atom always creates a backup file in the `~/.atom/recovery` directory, trying to restore it automatically via the main process in case there is a hard crash on the renderer process. This approach should be resilient enough to allow us to delete and stop using the `editor.backUpBeforeSaving` config option.

We can still keep it for a while in text-buffer, in case we need to use it again at some point, but it's probably fine to remove it from there too at some point in the future.

@atom/core: what do you think?

### DIFF
--- a/src/config-schema.coffee
+++ b/src/config-schema.coffee
@@ -237,10 +237,6 @@ module.exports =
         default: true
         title: 'Confirm Checkout HEAD Revision'
         description: 'Show confirmation dialog when checking out the HEAD revision and discarding changes to current file since last commit.'
-      backUpBeforeSaving:
-        type: 'boolean'
-        default: false
-        description: 'Ensure file contents aren\'t lost if there is an I/O error during save by making a temporary backup copy.'
       invisibles:
         type: 'object'
         description: 'A hash of characters Atom will use to render whitespace characters. Keys are whitespace character types, values are rendered characters (use value false to turn off individual whitespace character types).'
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -757,14 +757,14 @@ class TextEditor extends Model
   # Essential: Saves the editor's text buffer.
   #
   # See {TextBuffer::save} for more details.
-  save: -> @buffer.save(backup: @config.get('editor.backUpBeforeSaving'))
+  save: -> @buffer.save()
 
   # Essential: Saves the editor's text buffer as the given path.
   #
   # See {TextBuffer::saveAs} for more details.
   #
   # * `filePath` A {String} path.
-  saveAs: (filePath) -> @buffer.saveAs(filePath, backup: @config.get('editor.backUpBeforeSaving'))
+  saveAs: (filePath) -> @buffer.saveAs(filePath)
 
   # Determine whether the user should be prompted to save before closing
   # this editor.


nathansobo:
If we do crash, can we recover files regardless of the current project? As in, does it work globally?

as-cii:
> If we do crash, can we recover files regardless of the current project? As in, does it work globally?

Yes, the recovery service (running on the main process) gets notified before and after a file is saved. Before saving it, it will store a copy of it in the recovery folder, and (if everything goes smoothly) it will delete the backup after saving the original file.

If a crash happens meanwhile saving (which could corrupt or erase files), the recovery service (which is still up and running) will try to replace the corrupted file with the backup, and delete the backup afterwards. If that fails too, it will tell users where to find the backup (i.e. the specific file name in `~/.atom/recovery`).

All of this happens just via absolute paths and by interacting with the filesystem. The recovery service has no notion of projects or workspaces.

nathansobo:
Cool. If you feel confident the recovery service has us covered, I'm on board with removing the backup option. It never worked in all cases and has been a source of mysterious exceptions when people enable it in certain configurations.

joefitzgerald:
You should remove it from text buffer too if we aren't going to be using it. Otherwise, if you forget to remove it in the future it will be unused code that has to be maintained. We always have git history to get at the code if needed again in the future.

iolsen:
>All of this happens just via absolute paths and by interacting with the filesystem. The recovery service has no notion of projects or workspaces.

How are files named in the recovery directory to ensure no collisions?

BinaryMuse:
@iolsen 

> How are files named in the recovery directory to ensure no collisions?

We use a portion of the filename and then a few random bytes. I guess it's still possible we could collide, but it's relatively unlikely I think.

```javascript
static fileNameForPath (path) {
  const extension = Path.extname(path)
  const basename = Path.basename(path, extension).substring(0, 34)
  const randomSuffix = crypto.randomBytes(3).toString('hex')
  return `${basename}-${randomSuffix}${extension}`
}
```

as-cii:
I think this approach delivers a quite reliable experience, and it feels like users should always be able to retrieve their backups in the `~/.atom/recovery` folder, even when the automatic recovery doesn't work because of problems with permissions or another crash.

@lee-dohm: have we received other reports of data loss after shipping the recovery service? If so, were users able to recover their files manually by finding them in the recovery folder?

maxbrunsfeld:
:ship: ❔

as-cii:
> 🚢 ❔

Yeah, I think I'll go ahead and 🚢 this. In case it's really necessary, we can release a hotfix and put this back. Thanks everyone! 🙇

jonee:
I like this feature in gedit and just recently trying out atom. Not sure why such a basic option / feature would get deleted. 

In gedit, I save it in the same directory though with a ~ suffixed in the file name. 

This feature has saved me a lot- in accidental cases- where I deleted a file or a version control deleted your changes etc

Seems I am not alone
https://discuss.atom.io/t/to-save-a-backup-copy-of-current-file-before-save-it/35391

maxbrunsfeld:
> Not sure why such a basic option / feature would get deleted.

@as-cii Explained the reason pretty well in the comments above. Could you explain why the `~/.atom/recovery` folder doesn't work for you?

jonee:
I am assuming that the previous option editor.backUpBeforeSaving would save a backup file in the recovery directory on each save- but just leave it there and not delete it. So when accidents happen that deleted your file- you know you can find it in the recovery directory.

lee-dohm:
@jonee Thanks for your feedback. Various team members above stated our reasons for removing the previous implementation of the optional feature. We now have a feature that is more resilient, is enabled for all users and doesn't cause weird exceptions in uncommon configurations. If you like gedit's feature, you do have the option of creating a package that implements it for your own use. There is something similar in the [backupper package](https://atom.io/packages/backupper). You may want to try it out.

jonee:
Thanks a lot! Seems backupper would suffice for my needs. I have been using atom a lot these past weeks.

lee-dohm:
I'm glad to hear it! Please feel free to stop by [Discuss, the official Atom message board,](https://discuss.atom.io) if you have any questions :grinning:

jonee:
I will for sure check it out once in a while. Thanks!

micgrivas:
I just lost a lot of code I wrote recently, due to a mistaken save after some "undo" (seeking a change) and an accidental type.
The "backup before save" would have saved me all that code. 
Why did people consider that "a mean to save from an accidental crash" and not "a mean to save us from our stupidity or tiredness" ?   The "backup before save" is kind of auto-versioning, with only one version more. 
How would backupper have saved that case?

massic80:
> I just lost a lot of code I wrote recently, due to a mistaken save after some "undo" (seeking a change) and an accidental type.

@micgrivas check out [Local History](https://github.com/Nicolab/atom-local-history)

mapazarr:
I just recently migrated from emacs to atom and I was looking for a feature to backup files before saving, similar to the `file.ext~` files created by emacs. I don't see any files in the `~/.atom/recovery` files, as the initial post of the thread suggests.

I also tried the autosave package but it's not saving the backups.

Any suggestions? Thanks.

massic80:
@mapazarr check out [Local history](https://github.com/Nicolab/atom-local-history). It doesn't only save a backup, but also keep an history of versions.

PR #12394, opened 2016-08-11
Replace ChromeDriver integration tests main process Mocha tests

nathansobo:
This is a proof of concept of where I think we need to take our main process testing. There's a lot to do to get to ideal, and this PR doesn't do any of that... All we do is the minimal to instantiate an `AtomApplication` in our main process mocha suite and test the results of calling `openWithOptions` repeatedly.

The main reason we did this now is because this test was failing on CircleCI and blocking our build modernization effort and we didn't want to fight with the existing ChromeDriver test.

In the future, I'd like to have a single method on `AtomApplication` called `launch` that takes CLI arguments directly from our main script. We can integration test that method or mock out a method like `openWindow` that takes options to separate unit testing of CLI parsing from window open logic. The point is, the main process code is just JavaScript and we can test it just like we test anything else. If we reduce `main.js` to just calling `launch`, then we only need ChromeDriver coverage of our shell scripts that launch Atom.

/cc @maxbrunsfeld


maxbrunsfeld:
:+1: Nice to see a little cleanup of that pesky `global.atomApplication` in the process.

nathansobo:
@as-cii Almost finished this... there's 4 integration tests remaining in the old ones. I guess we should leave one of them for a 🚬  test.

thomasjo:
Any particular reason for using `function () { ... }` instead of `() => { ... }`?

as-cii:
> Any particular reason for using function () { ... } instead of () => { ... }?

I think a reason for that is that we need to specify the timeout on some mocha tests (via `this.timeout(ms)` and, therefore, we don't want `this` to be reassigned.

nathansobo:
> Any particular reason for using function () { ... } instead of () => { ... }?

I'm conflicted about using `() =>` everywhere just because it looks better. It has an explicit purpose that is different than the `function ()` syntax, which is binding `this`. In places we don't need to bind `this`, then, do we need it? I do see value in cases like `Array.map`, where brevity adds a lot to readability. I'm not convinced we need to use `() =>` everywhere though.

maxbrunsfeld:
> It has an explicit purpose that is different than the `function ()` syntax, which is binding `this`

Arrow functions used to be transpiled using `bind`, but they really have nothing to do with 'binding'. Rather, they're just simpler than normal `functions`, in that they don't automatically create a `this` variable based on the way they are called. Instead, the `this` just uses the normal variable scoping semantics.

In my opinion, arrow functions should be the default way of writing anonymous functions, unless we need to pass a callback to some legacy API that uses a dynamically bound `this`, which is unfortunately the case with Mocha's `it` function.

Here's some V8 maintainer's statement about their perf:

> it took like three months to fix `this` usage in V8 to always just go through the (possibly context-allocated) variable

> Performance-wise, arrow functions are just like functions. They should be just as fast as if you wrote them with function. So use them with joy, use them with abandon, use them judiciously -- however you decide you use them, don't let perf influence your decision one way or the other.

https://wingolog.org/archives/2015/06/18/arrow-functions-coming-to-chrome-45

nathansobo:
> In my opinion, arrow functions should be our default way of writing anonymous functions, unless we need to pass a callback to some legacy API that uses a dynamically bound this, which is unfortunately the case with Mocha's describe/beforeEach/it functions.

I'm sold.

PR #12395, opened 2016-08-11
Give each integration test its own ATOM_HOME

nathansobo:
We symlink the compile cache from the original ATOM_HOME to ensure window startup is still performant, but avoid anything else leaking across tests.

/cc @maxbrunsfeld

### DIFF
--- a/spec/integration/startup-spec.coffee
+++ b/spec/integration/startup-spec.coffee
@@ -13,11 +13,15 @@ runAtom = require './helpers/start-atom'
 CSON = require 'season'
 
 describe "Starting Atom", ->
-  atomHome = temp.mkdirSync('atom-home')
-  [tempDirPath, otherTempDirPath] = []
+  [atomHome, tempDirPath, otherTempDirPath] = []
 
   beforeEach ->
     jasmine.useRealClock()
+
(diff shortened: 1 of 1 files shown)


spec/integration/startup-spec.coffee
```diff
@@ -13,11 +13,15 @@ runAtom = require './helpers/start-atom'
...
 
   beforeEach ->
     jasmine.useRealClock()
+
+    atomHome = temp.mkdirSync('atom-home')
+    # Symlinking the compile cache into the temporary home dir makes the windows load much faster
+    fs.symlinkSync(path.join(process.env.ATOM_HOME, 'compile-cache'), path.join(atomHome, 'compile-cache'))
+
     fs.writeFileSync(path.join(atomHome, 'config.cson'), fs.readFileSync(path.join(__dirname, 'fixtures', 'atom-home', 'config.cson')))
     fs.removeSync(path.join(atomHome, 'storage'))
```
maxbrunsfeld: These two lines for clearing out the ATOM_HOME are probably not needed now, right?

nathansobo:
Closing this because we got rid of all but one of the Chromedriver tests in #12394.

PR #12396, opened 2016-08-11
Fix Regression In #12317 For zsh and Patch Environment On OSX For Users Of fish

joefitzgerald:
In Atom 1.7 and 1.8, the environment would be patched for users of the zsh shell on OS X, who launched Atom from the Dock / Finder / Spotlight. A whitelist of shells was established in #12317, which is extended here to include zsh and fish.

Tested with built-in zsh, homebrew installed zsh, and homebrew installed fish.

### DIFF
--- a/spec/update-process-env-spec.js
+++ b/spec/update-process-env-spec.js
@@ -2,7 +2,7 @@
 /* eslint-env jasmine */
 
 import child_process from 'child_process'
-import {updateProcessEnv, shellShouldBePatched} from '../src/update-process-env'
+import {updateProcessEnv, envShouldBePatched} from '../src/update-process-env'
 import dedent from 'dedent'
 
 describe('updateProcessEnv(launchEnv)', function () {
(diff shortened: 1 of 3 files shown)


src/update-process-env.js
```diff
@@ -33,9 +40,23 @@ export default function updateProcessEnv (launchEnv) {
...
+    return false
+  }
+  for (let s of OSX_SHELLS_TO_PATCH) {
+    if (shell.endsWith(s)) {
+      return shell
+    }
+  }
+
+  return false
+}
```
maxbrunsfeld: I would suggest two fairly minor stylistic changes to this `shellShouldBePatched` function:

1. I think it should *take* a shell as an argument and return a boolean, rather than internally getting the shell from `process.env` and returning either a string or `null`.

2. I think the name is a bit odd; it makes it sound like the shell itself is getting patched. Maybe `canGetLoginEnvFromShell(shellPath)`?

I think this would make the function a little simpler to test, since you wouldn't have to overwrite `process.env.SHELL`. You wouldn't need a loop; you could just do:

```js
expect(canGetLoginEnvFromShell('/bin/bash')).toBe(true)
expect(canGetLoginEnvFromShell('/some/other/bash')).toBe(true)
expect(canGetLoginEnvFromShell('/usr/local/bin/fish')).toBe(true)

expect(canGetLoginEnvFromShell('/bin/give-me-cash')).toBe(false)
expect(canGetLoginEnvFromShell('/usr/bin/birthday-bash')).toBe(false)
```

spec/update-process-env-spec.js
```diff
@@ -84,5 +84,33 @@ describe('updateProcessEnv(launchEnv)', function () {
         expect(process.env).toEqual({FOO: 'bar'})
       })
     })
+
+    describe('shells on osx', function () {
+      it('envShouldBePatched() returns the shell when the shell should be patched', function () {
```
maxbrunsfeld: Could you move the function name up into the `describe`, so that the `it` sentences read a bit better?

```js
describe('envShouldBePatched()', function () {
  it('returns true for shells that ', function () {
    // ...
  })
```

maxbrunsfeld: The `envShouldBePatched` name is a little clearer, but I think it'd be best to make it clear that the return value depends on the shell, not the env. Also, the env is completely replaced (except for the 3 'special' variables), so emphasizing the 'patching' doesn't seem entirely necessary.

How about `shouldGetEnvFromShell`?

maxbrunsfeld:
Great; Thanks for fixing this regression @joefitzgerald!

PR #12397, opened 2016-08-11
Add help, don't build winstaller by default, add skip

damieng:
1. Adds help for the build script to clarify options up-front
2. Does not build Windows installer by default as it is SLOW


script/build
```diff
@@ -52,7 +61,12 @@ dumpSymbols()
       }
       return Promise.resolve(packagedAppPath)
     } else if (process.platform === 'win32') {
-      return createWindowsInstaller(packagedAppPath, argv.codeSign).then(() => packagedAppPath)
+      if (argv.createInstaller) {
+        return createWindowsInstaller(packagedAppPath, argv.codeSign).then(() => packagedAppPath)
+      }
+      else {
+        console.log('Skipping installer. Specify the --create-installer option to create a Squirrel-based Windows installer.'.gray)
```
nathansobo: Skipping installer *creation*?

nathansobo:
These all seem reasonable to me. I don't mind specifying flags to add stuff but would prefer not to specify flags to skip stuff. Which is what you're doing here so 👍 .

as-cii:
Do we need to bundle the electron zip?

PR #12409, opened 2016-08-12
Mock clock and animation frames in TextEditorComponent specs

nathansobo:
A few months ago, I rewrote the editor tests to use real timers and animation frames to address coverage gaps in the existing tests. Unfortunately, it's proven extremely difficult to make these tests run reliably on CI.

We plan to extract the editor completely soon, at which point it may be worth another try to implement more realistic tests, though doing so will probably require a careful look at the many nuanced time-sensitive sequences of operations that occur within the editor. Until then, I'm going back to a less realistic but more reliable approach to testing this component.


nathansobo:
This is passing on Circle and Janky is out of comission right now. Going to move forward.

PR #12426, opened 2016-08-15
Delegate TextEditor.getMarkerCount to default marker layer.

oggy:
Looks like this should follow the other marker methods in delegating to the default marker layer rather than the buffer directly.


nathansobo:
Good catch thanks @oggy!

Issue #10086, opened 2015-12-16
Prompt before opening large files to avoid making the editor hang by mistake


zinkkrysty:
We know that atom hangs when opening big files (#979). Usually I don't intend to open big files, but sometimes I do because of using the CMD+T dialog too fast, so by mistake. Can we prevent this with a small action the user has to take before ? Given that the the user has to wait a while when the file is opened, I think that warning the user and prompting to open is a small request. Then your team can focus on improving performance without getting many users frustrated until then.

![atom_open_big_file](https://cloud.githubusercontent.com/assets/570771/11843199/7df7c7c0-a408-11e5-8a36-477638bbd884.png)

Zireael07:
Good idea.

alexandernst:
Or even better, just make it work with big files... once and for all.

avdg:
If it's possible, it would be nice to include the file size so one could see the difference between a let's say 600kb, a 1.5mb file and for the extreme cases files bigger than let's say 50mb.

samartioli:
This seems like a reasonable intermediate step until Atom supports large buffers. I was a very long time sublime user, and I am loving Atom the past couple weeks, but I'm starting to get very frustrated with clicking on build files and having atom hang and then have to force-quit.. 

As stated above, you could prompt before opening a file larger than 1mb. 

You could also show a unix head preview of the file even, and then have a button to open the rest, while showing the file size to the user. 

Again, 99% of my files are under 100 lines of code... but when I compile or bundle and then accidentally click on the file in the tree view... bad times.

elijahmanor:
@mnquintana yeah, it'd be nice if the threshold could be customized in the main settings with a default value https://github.com/atom/atom/blob/1fac3cb4e2608402831600c5f63dae38329251b1/src/workspace.coffee#L515

Krisa:
Also as temporary workaround, it could be great that when clicking on a large file (size configurable), atom opens the file in a basic editor (also configurable) which supports large file (e.g. notepad ++). We usually don't care of all the atom functionalities when opening very large file and a more primitive editor is just good enough.

remexre:
Anyone working on a PR for @elijahmanor's idea? If not, would one be accepted?

remexre:
Welp, made one and it passes CI.

nathansobo:
@50Wliu can this be closed since we do prompt?

50Wliu:
Yup, since there is a prompt at 20MB I'm closing this.

[...]

PR #12439, opened 2016-08-16
Adds configuration option for large file warning threshold.

remexre:
See #10086.

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -433,30 +433,36 @@ describe "Workspace", ->
           expect(editor.largeFileMode).toBe true
 
     describe "when the file is over user-defined limit", ->
-      it "prompts the user to make sure they want to open a file this big", ->
-        spyOn(fs, 'getSizeSync').andReturn atom.config.get('core.warnOnLargeFileLimit') * 1048577 # 20MB
+      test = (size, shouldPrompt) ->
+        spyOn(fs, 'getSizeSync').andReturn size * 1048577
         atom.applicationDelegate.confirm.andCallFake -> selectedButtonIndex
         atom.applicationDelegate.confirm()
-        selectedButtonIndex = 1 # cancel
-
         editor = null
         waitsForPromise ->
           workspace.open('sample.js').then (e) -> editor = e
+        if shouldPrompt
+          runs ->
+            expect(editor).toBeUndefined()
+            expect(atom.applicationDelegate.confirm).toHaveBeenCalled()
 
-        runs ->
-          expect(editor).toBeUndefined()
-          expect(atom.applicationDelegate.confirm).toHaveBeenCalled()
+            atom.applicationDelegate.confirm.reset()
+            selectedButtonIndex = 0 # open the file
 
-          atom.applicationDelegate.confirm.reset()
-          selectedButtonIndex = 0 # open the file
-
-        waitsForPromise ->
-          workspace.open('sample.js').then (e) -> editor = e
+          waitsForPromise ->
+            workspace.open('sample.js').then (e) -> editor = e
 
         runs ->
           expect(atom.applicationDelegate.confirm).toHaveBeenCalled()
           expect(editor.largeFileMode).toBe true
-
+      it "prompts the user to make sure they want to open a file this big", ->
+        atom.config.set "core.warnOnLargeFileLimit", 20
+        test 20, true
+      it "doesn't prompt on files below the limit", ->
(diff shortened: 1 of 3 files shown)


nathansobo:
@50Wliu Yeah I'm cool with it. Makes sense. If someone can verify this works correctly I'm 👍 to merge.

50Wliu:
I tested this out and it works correctly.  One thing though @remexre...can you please adjust [this spec](https://github.com/atom/atom/blob/f376316d60b7b0e96ceaff8c631ea6458f31907c/spec/workspace-spec.coffee#L435) to respect the new config option and also to test out a different value to make sure that the option doesn't regress?  Thanks!

remexre:
Like so?

nathansobo:
@remexre I don't think that's quite it. What we need is a test that the limit is configurable. So you would assign a limit to the configuration in the test to a number below the mocked file size and confirm we show the dialog, then reassign the limit to a number above the mocked file size and confirm we don't show the dialog.

remexre:
Is this good?

spec/workspace-spec.coffee
```diff
@@ -432,31 +432,37 @@ describe "Workspace", ->
...
-
+      it "prompts the user to make sure they want to open a file this big", ->
+        atom.config.set "core.warnOnLargeFileLimit", 20
+        test 20, true
+      it "doesn't prompt on files below the limit", ->
+        atom.config.set "core.warnOnLargeFileLimit", 30
+        test 20, false
+      it "prompts for smaller files with a lower limit", ->
+        atom.config.set "core.warnOnLargeFileLimit", 5
+        test 10, true
```
nathansobo: Can you space this stuff out?

spec/workspace-spec.coffee
```diff
@@ -432,31 +432,37 @@ describe "Workspace", ->
         runs ->
           expect(editor.largeFileMode).toBe true
 
-    describe "when the file is over 20MB", ->
-      it "prompts the user to make sure they want to open a file this big", ->
-        spyOn(fs, 'getSizeSync').andReturn 20 * 1048577 # 20MB
+    describe "when the file is over user-defined limit", ->
+      test = (size, shouldPrompt) ->
```
nathansobo: Can you give this helper a more descriptive name?

spec/workspace-spec.coffee
```diff
@@ -432,31 +432,37 @@ describe "Workspace", ->
...
+        spyOn(fs, 'getSizeSync').andReturn size * 1048577
         atom.applicationDelegate.confirm.andCallFake -> selectedButtonIndex
         atom.applicationDelegate.confirm()
-        selectedButtonIndex = 1 # cancel
-
         editor = null
         waitsForPromise ->
           workspace.open('sample.js').then (e) -> editor = e
+        if shouldPrompt
```
nathansobo: If `shouldPrompt` is false, we should assert that it didn't prompt.

nathansobo:
Getting close. I left a few more comments.

spec/workspace-spec.coffee
remexre: How would I do that, exactly? I'm not 100% familiar with Atom's internal structure...

remexre:
First two done, not 100% sure how to do the last one.

nathansobo:
You can also do a negative assertion on your mock...

```coffee
expect(atom.applicationDelegate.confirm).not.toHaveBeenCalled()
```

remexre:
I think it's fixed now? For some reason, `script/test` fails, claiming it doesn't have dpkg, so local tests aren't working for me...

50Wliu:
@remexre You can view specs by clicking on the `ci/circleci` check.  Right now there's just an issue where `selectedButtonIndex` is undefined.  In addition, I don't think there will be an available application delegate to `confirm()` if the file size is beneath the limit?

50Wliu:
This looks good now - thanks!

nathansobo:
⚡️ Thanks @remexre!

Ben3eeE:
Closes https://github.com/atom/tree-view/issues/810?

remexre:
I believe so.

adambuczynski:
@remexre does this apply to actually opening files (e.g. double click), or also when previewing files from tree view (single click)? I just had the editor hang on me because I accidentally clicked on a file with 70k+ lines.

remexre:
IIRC previewing

adambuczynski:
Great, thanks

PR #12440, opened 2016-08-16
Serialize active pane item using index instead of URI

matthewwithanm:
Under the current behavior, which item is active won't be restored unless it implements `getURI()`.  This change lifts that restriction by serializing the index of the active item instead.


nathansobo:
@matthewwithanm Heads up... this bra

nathansobo:
Seeing some really weird failures on CI that don't seem related. Stay tuned.

as-cii:
I have rebased this again to make sure the failure wasn't related to some outdated package, but it turns out we were using a deprecated API in one of the new tests. I have updated it to use the right interface in 98fb298 and everything is 💚 now.

Thanks @matthewwithanm  and @nathansobo! 👏

matthewwithanm:
Thanks!

PR #12448, opened 2016-08-17
Introduce autoWidth for TextEditors

### LINKED ISSUE #12441: atom-text-editor should size itself according to contents
### Description

It would be great if the `atom-text-editor` element sized itself according to its contents. We rely on this behavior in Nuclide to provide syntax highlighting inside [type-hints](https://nuclide.io/docs/languages/flow/#type-hinting), which uses an atom-text-editor inside a marker overlay. This used to work fine in Atom ≤1.8.x, although it relied on the `min-width: 1em` trick documented in the init script below. In 1.9.x, `.lines` elements inside the atom-text-editor shadow DOM don't have an explicit pixel width set anymore, which causes this behavior to break.

Ideally there would be an attribute that would make the text editor reliably auto-size itself according to its contents.

### Steps to Reproduce

Use the following init script (`init.js` in ~/.atom):
```es6
'use babel';

import {TextBuffer} from 'atom';

function setupTextEditor(props) {
  const textEditorParams = {
    buffer: props.textBuffer,
    lineNumberGutterVisible: false,
  };
  const textEditor = atom.workspace.buildTextEditor(textEditorParams);
  if (props.grammar != null) {
    textEditor.setGrammar(props.grammar);
  }
  return textEditor;
}

async function setup() {
  const buffer = new TextBuffer('const f = {a: 123};'); // dummy JS code
  const [jsGrammar] = atom.grammars.getGrammars().filter(g => g.name === 'JavaScript'); // hack to get JS grammar
  const textEditor = await atom.workspace.open();
  const range = textEditor.setText('test');
  const marker = textEditor.markBufferRange(range, {invalidate: 'never'});
  const datatipElement = document.createElement('div');
  const floatingEditor = setupTextEditor({
    grammar: jsGrammar,
    textBuffer: buffer,
  }); // floatingEditor is expected to size itself according to ints contents.
  datatipElement.style.border = '1px solid red';
  datatipElement.style.minWidth = '1em'; // hack to auto-size in 1.8.x and prior. Auto-sizing does not work in 1.9.x
  datatipElement.appendChild(floatingEditor.getElement());
  textEditor.decorateMarker(
    marker,
    {
      type: 'overlay',
      position: 'tail',
      item: datatipElement,
    },
  );
  return datatipElement;
}
atom.packages.onDidActivateInitialPackages(setup); // wait for grammar package to activate
```

**Expected behavior:**
1.8.x: The text-editor element (highlighted with red border) expands to fit its contents:
screenshot 2016-08-16 18 28 15

**Actual behavior:**
1.9.x: text-editor element does not set its width; it is `0` by default (1em in the screenshot due to applying the `min-width: 1em` trick):
screenshot 2016-08-16 18 28 34

### Versions
Broken in ≥ 1.9.x.

as-cii:
Fixes #12441 
Refs #12445 

This pull-request is a proposal to introduce an `autoWidth` parameter on `TextEditor` instances. When instantiating an editor with this parameter, the `width` of the `<atom-text-editor>` element will be automatically adjusted to the width of its content (+ the width of the gutter, if present).

autowidth

Since we explicitly assign a `width` to text editors that have enabled `autoWidth`, there's no need to resort to CSS hacks on the parent component (e.g. `style.minWidth = '1em'`) or on the element itself (i.e. `display: inline-block`). This option will be off by default, meaning that editors will inherit the parent element's width, exactly like a `display: block` element. Because this will be an opt-in behavior it shouldn't introduce any regression and, thus, we might hot-fix stable and beta (respectively 1.9 and 1.10, as of this pull-request) with this new parameter.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -4376,7 +4376,6 @@ describe('TextEditorComponent', function () {
       runAnimationFrames()
       expect(wrapperNode.offsetWidth).toBe(contentNode.offsetWidth)
 
-      global.debug = true
       editor.update({autoWidth: true})
       runAnimationFrames()
       const editorWidth1 = wrapperNode.offsetWidth
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -4368,6 +4368,34 @@ describe('TextEditorComponent', function () {
     })
   })
 
+  describe('width', function () {
+    it('sizes the editor element according to the content width when auto width is true, or according to the container width otherwise', function () {
+      contentNode.style.width = '600px'
+      component.measureDimensions()
+      editor.setText("abcdefghi")
+      runAnimationFrames()
+      expect(wrapperNode.offsetWidth).toBe(contentNode.offsetWidth)
+
+      global.debug = true
+      editor.update({autoWidth: true})
+      runAnimationFrames()
+      const editorWidth1 = wrapperNode.offsetWidth
+      expect(editorWidth1).toBeGreaterThan(0)
+      expect(editorWidth1).toBeLessThan(contentNode.offsetWidth)
+
+      editor.setText("abcdefghijkl")
+      editor.update({autoWidth: true})
+      runAnimationFrames()
+      const editorWidth2 = wrapperNode.offsetWidth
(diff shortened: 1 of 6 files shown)


nathansobo:
🎉  Thanks @as-cii!

maxbrunsfeld:
Great work @as-cii!

Issue #11361, opened 2016-04-04
Atom is not maintaining the state


kamlekar:
This is a question actually.

I am opening some project folders through atom. When I restart the atom, the atom is being empty. Again, I am manually adding folders.

I even tried `reopen-last-item`, it doesn't do anything.

Please help.

mnquintana:
When you restart, how are you opening Atom? (e.g. from the command line, from the Dock, from the Start Menu) What OS are you on? What version of Atom are you using? Can you reproduce this after quitting Atom completely and reopening it in safe mode? (`atom --safe`)

kamlekar:
- I am opening from the Dock. 
- I am on Mac 10.11.1. 
- I was having version `1.5.2`. After this issue, I checked for updates and installed the latest update. 
- I also installed `remember-session` plugin, to manage sessions. (_now it seems to be working fine_)
- In future, if I found any issues related to it I will definitely post more info here.

Thanks.

mnquintana:
You should uninstall `remember-session` – that package has been deprecated for awhile now, as its functionality is in Atom core. That probably was the root of your issues here. Glad things seem to be resolved! :smile:

kamlekar:
I installed it because atom wasn't maintaining state. Tomorrow morning i
will uninstall remember-session and check again.

kamlekar:
It happened again this Morning. I closed Atom editor on Friday evening and completely shut down my mac. After starting mac today(monday) morning, I opened atom, where I can't see the files which were kept open on Friday evening. (I am using the latest version of atom)

arcticicestudio:
Have you checked if you disabled the **[ ] Open Empty Editor on Start** core setting?
![atom-ghi-11361-core-setting-empty-editor](https://cloud.githubusercontent.com/assets/7836623/14595845/ac47c2ec-0542-11e6-91c5-de2f330b0d6e.png)

kamlekar:
Yeah it was selected :/ thanks. I think it was selected by default. 


![](http://i.imgur.com/IA6pjed.jpg)

kamlekar:
Hey guys no mean bad just a joke image. I really like your contribution. :). I hope you guys will uncheck that setting by default.

kamlekar:
This is happening again though the "open editor on start" checkbox is unchecked.

Wouter33:
Happening here all the time. Really thinking of switching back to my old editor again, this is so frustrating. 

[...]

PR #12486, opened 2016-08-20
load initialPaths for first new-window event

hongyuanw:
Fix bug https://github.com/atom/atom/issues/11361

1、On macOS,  the sate is not saved when the last window is removed because the progress is running, so "removeWindow" function must be updated to :

> `@saveState(false) unless window.isSpec`

2、We must load paths from '.atom/storage/application.json' for the first window  when the application is running and no window, especially on macOS

### DIFF
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -127,7 +127,7 @@ class AtomApplication
         app.quit()
         return
     @windows.splice(@windows.indexOf(window), 1)
-    @saveState(true) unless window.isSpec
+    @saveState(false) unless window.isSpec
 
   # Public: Adds the {AtomWindow} to the global window list.
   addWindow: (window) ->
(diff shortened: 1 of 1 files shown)


yuanwhy:
@maxbrunsfeld @50Wliu @lee-dohm Could you review this code when you get a chance?

maxbrunsfeld:
@yuanwhy I believe the current behavior is already correct. I may be misunderstanding the problem that you're describing. Could you describe step by step what you're doing, what you expect to see, and what you actually see?

lee-dohm:
@maxbrunsfeld Also see https://github.com/atom/atom/issues/12495

yuanwhy:
@maxbrunsfeld Let's review the original code
```
 # Public: Removes the {AtomWindow} from the global window list.
  removeWindow: (window) ->
    if @windows.length is 1
      @applicationMenu?.enableWindowSpecificItems(false)
      if process.platform in ['win32', 'linux']
        app.quit()
        return
    @windows.splice(@windows.indexOf(window), 1)
    @saveState(true) unless window.isSpec 
```

The atom progress will quit on win32 and linux when the last window is closed,  but is not for macOS.

So, on macOS, when we close the last window, code 
```
@windows.splice(@windows.indexOf(window), 1)
```
will remove the last window and its state from windows container,   then 
```
@saveState(true) unless window.isSpec 
```
will not save the last window's state to application.json becase of "true" argument ...

the saveState function :
```
saveState: (allowEmpty=false) ->
    return if @quitting
    states = []
    for window in @windows  // when you close the last window, @windows is empty !
      unless window.isSpec
        if loadSettings = window.getLoadSettings()
          states.push(initialPaths: loadSettings.initialPaths)
    if states.length > 0 or allowEmpty  // if you use saveState(true), you will write []  to 'application.json'
      @storageFolder.storeSync('application.json', states)
```

abarisain:
It looks like it is saving an empty state, as it should, since no windows are open. That's a valid state on osx. If I manually close all windows and then quit atom, nothing should be reopened on start. It would mean that it is not possible to close the last project if you don't persist an empty state.

If anything, atom has the opposite problem, where it saves its state on quitting, disregarding OSX's state restoration option (in SysPrefs -> general). It's quite handy for an IDE though.

maxbrunsfeld:
> will not save the last window's state to application.json becase of "true" argument

No, it *does* save the state because of the "true" argument. The state is that there are no windows.

Again, it is the expected behavior on Mac that if you quit the app with no windows open, when you reopen the app, you see none of your previous windows.

PR #12501, opened 2016-08-23
Add atom.project.observeBuffers

zertosh:
This makes `onDidAddBuffer` symmetrical with other `onDidAddX` methods and their `observeXs`. It also documents `onDidAddBuffer` and adds tests for it.

Released under CC0.


maxbrunsfeld:
This great. Thanks @zertosh!

nathansobo:
We really need to move buffers out of project someday, but when we do this can come with them.

PR #12502, opened 2016-08-23
Update languages

winstliu:
~~Do not merge; more language updates coming soon!~~
All language updates are done.

Too tired to write up a detailed changelog, but here are the major things to look out for:
- YAML support got seriously improved with this latest release.  But with serious improvements comes the potential for serious regressions, so that might be something to keep an eye out for.
- language-c got a fix ported from the textmate bundle!  Yay!
- Tons of JS improvements, especially regarding JSDoc and object keys.  Again, also has the possibility of regressions.


50Wliu:
Ok, I've narrowed this down to one of archive-view's dependencies receiving a new version sometime after 2bd489e's commit time.  Reason being that I just restarted that CircleCI build (which was passing before) and it now fails.

50Wliu:
archive-view tests are passing now but integration tests are failing 😠.

thomasjo:
@50Wliu Just some kind of hiccup. I restarted the build and there was just an auto-updater test that failed. In other words, this can be considered 💚 as far as the language packages are concerned.

50Wliu:
Haha, you must have better luck than I do then :).  I retried it twice and integration tests failed both times.  Maybe it has something to do with building without the cache.

Issue #9677, opened 2015-11-18
No longer possible for package to have a pane item deserialized along text editors


abe33:
Actually it works on 1.2.1 but is broken on beta.

In [tablr](https://github.com/abe33/atom-tablr) I have made my CSV editor class serializable so that quitting Atom with a csv editor open restores it on the next run. For that purpose I have a method on my package main module that loads all the models and views required to make that possible and registers the deserializers as soon as the package is loaded: 

![tablr_coffee_-__users_c_nehemie_github_tablr_-_atom](https://cloud.githubusercontent.com/assets/247140/11245265/19604b86-8e12-11e5-9f94-06a1fe0ec839.png)

However, as you can see below, when reloading Atom with my table editor open, it first tries to deserialize the object, and then loads the package: 

![tablr_coffee_-__users_c_nehemie_github_tablr_-_atom](https://cloud.githubusercontent.com/assets/247140/11245281/2f636f1c-8e12-11e5-8f71-eafd5d519c2d.png)

Is it a regression? Or was it decided to remove this feature?

benogle:
cc @nathansobo is this fixed with your recent deserializer fix?

nathansobo:
Yes, it should be. Please try again on latest 1.2.2 and let us know if there are any issues.

PR #12513, opened 2016-08-25
Reset the package's local storage setting if loading cannot be deferred

abe33:
Hi there, it’s been a while since I last got some time to work on Atom.

So this PR attempt to fix an annoying issue I struggled with for months now, without being able to figure when it came from, but I finally found an explanation. I initially reported it in #9677, but without any clue as to why it was not working properly the issue was closed expected.

So now, let’s enter the main topic. 

Let’s say we create a new package, we’re running it in dev mode, and for the moment there’s no deserializers or registered views for that package. 

When requiring the main module, we will end up passing in [this branch](https://github.com/atom/atom/blob/master/src/package.coffee#L442-L444) that will set the `getCanDeferMainModuleRequireStorageKey()` flag to `true`.

Now if we start adding deserializers into our package, since the flag is never turned off, the package main module loading will always be deferred on activation as [this condition will always be true](https://github.com/atom/atom/blob/master/src/package.coffee#L102), resulting in error if the there’s serialized pane items that need to be deserialized. 

Of course, as these settings are scoped using the package version, this is generally not an issue with published packages. If serializers or views are added between two versions the local storage keys will be different  and the previous value will be ignored.

However, this issue is making developing a package that have serializers seriously complicated as there’s no longer a way to test them in dev mode once we entered that state (unless knowing that this local storage setting must be removed).

A simple solution here is to simply remove the local storage item if we find that the package now has deserializers or registered views that it didn’t have before.

### DIFF
--- a/src/package.js
+++ b/src/package.js
@@ -917,6 +917,10 @@ module.exports = class Package {
             this.getCanDeferMainModuleRequireStorageKey(),
             'true'
           );
+        } else {
+          localStorage.removeItem(
+            this.getCanDeferMainModuleRequireStorageKey()
+          );
         }
         return this.mainModule;
       }


maxbrunsfeld:
This looks good to me. Restarted the failing circle build.

sadick254:
Hey @abe33. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

PR #12522, opened 2016-08-25
Register all editors added to the workspace

maxbrunsfeld:
This replicates a bugfix that landed on master in via https://github.com/atom/atom/pull/12125, into the `1.10-releases` branch.

Refs https://github.com/atom/autocomplete-plus/issues/752

We need this before we can do another railcar release.

### DIFF
--- a/src/workspace.coffee
+++ b/src/workspace.coffee
@@ -165,7 +165,10 @@ class Workspace extends Model
     @onDidAddPaneItem ({item, pane, index}) =>
       if item instanceof TextEditor
         grammarSubscription = item.observeGrammar(@handleGrammarUsed.bind(this))
-        item.onDidDestroy -> grammarSubscription.dispose()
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Will this not be fixed in 1.9?

maxbrunsfeld:
Oh wow; I did not realize this bug existed in 1.9. Thanks for the heads up @Ben3eeE. I think we are just going to ship `v1.10` (stable) on Monday, so users of 1.9.9 will get upgraded to that.

Ben3eeE:
Alright. Just worried cause i have had this problem myself on 1.9. Excited for many things in 1.10 so happy to hear about this! There are actually few recent issues among top 4 on the autocomplete repo that might be because of this. Also last few comments in the old autocomplete wrong position issue https://github.com/atom/autocomplete-plus/issues/640 seems to be this and not the position glitch. So maybe you can close some of these after 1.10 lands.

PR #12529, opened 2016-08-26
Disable telemetry on AtomApplication tests and smoke tests

as-cii:
Unless a choice has been made by the user, the telemetry consent tab always shows up as the first one when opening Atom, thus breaking some of the assumptions we make in main process tests. This caused our test suite to be ❌, even if behavior-wise everything was okay.

With this pull-request we disable the telemetry tab in main process and smoke tests, so that the build is 💚 again.

/cc: @atom/core


damieng:
Thanks!

PR #12535, opened 2016-08-27
Add support to the xonsh shell

santagada:
support xonsh (http://xon.sh/).


santagada:
if someone can restart the tests on appveyor, the error that it gave is a timeout

Issue #12495, opened 2016-08-22
Atom doesn't remember added projects on start up


tomasx:
This is a known limitation of the window state system; it only saves when changing focus, adding and removing windows. We should save when adding and removing project folders in addition.

## Repro Steps
1. Launch Atom in a local repository
2. Add a second project folder using <kbd>Cmd+Shift+O</kbd>
3. Exit Atom
4. Launch Atom using the Dock icon

**Expected:** Both project folders to be listed in the Tree View
**Actual:** Only the first project folder is listed in the Tree View

Info:
- version 1.9.9 (the bug also existed on 1.9.8)
- Mac OS 10.10.5
- "Open empty editor at start" is unchecked

Nikki1993:
That is not a bug but intended behaviour. Atom opens new instance each time you launch it hence why it "doesn't remember" last opened project. You need to supply filename/project folder to it each time you launch it.

tomasx:
Hey, according to this conversation this is a bug. Please read the last 6 comments https://discuss.atom.io/t/remember-open-projects-when-reopening-atom/7053/24

Also, why on earth you wouldn't provide the functionality to remember the projects that you have added?

Nikki1993:
@tomasx I've been reading comments on github by Atom project members and this is not a bug according to them but the way electron/atom work (not too sure on details here).  You basically launch a new browser instance each time. 

I agree with you, that would be convenient to remember last project opened, it is a feature request afaik but I don't know when/if it will be implemented.

DamnedScholar:
@Nikki1993 Atom is supposed to remember the very last set of root paths you had open. It uses `.atom/storage/application.json` to recall these paths and use the relevant IndexedDB entry to recreate the open tabs. It still does this just fine on Windows, but [according](https://discuss.atom.io/t/remember-open-projects-when-reopening-atom/7053/27?u=damnedscholar) to @tomasx, the functionality broke on update 1.9.8 for Mac.

Nikki1993:
@DamnedScholar that's odd cause it never opened previously opened projects for me on Windows, Mac and Ubuntu/Solus. I remember like couple of month ago someone said that's an intended behaviour for Atom because it's the way it functions. Pardon for the false info then :P

DamnedScholar:
[...]

PR #12545, opened 2016-08-29
send event to save state after 'Add Project Folder' or 'Remove Projec…

hongyuanw:
Fixes https://github.com/atom/atom/issues/12495

When we add or remove project folder,  the window's current project path will be changed, then "initialPaths" will be reloaded, it is the best time to save state

/cc @maxbrunsfeld @lee-dohm

### DIFF
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -126,7 +126,7 @@ class ApplicationDelegate
     loadSettings = getWindowLoadSettings()
     loadSettings['initialPaths'] = paths
     setWindowLoadSettings(loadSettings)
-    ipcRenderer.send("did-save-state")
+    ipcRenderer.send("did-change-paths")
 
 
   setAutoHideWindowMenuBar: (autoHide) ->
--- a/src/application-delegate.coffee
+++ b/src/application-delegate.coffee
@@ -128,7 +128,6 @@ class ApplicationDelegate
     setWindowLoadSettings(loadSettings)
     ipcRenderer.send("did-change-paths")
 
-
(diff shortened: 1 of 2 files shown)


src/application-delegate.coffee
```diff
@@ -126,6 +126,8 @@ class ApplicationDelegate
     loadSettings = getWindowLoadSettings()
     loadSettings['initialPaths'] = paths
     setWindowLoadSettings(loadSettings)
+    ipcRenderer.send("did-save-state")
```
maxbrunsfeld: I think we should rename the event from `did-save-state` to `did-change-paths`. The current name `did-save-state` makes it sound like Atom already *did* save the state.

maxbrunsfeld:
This looks great to me; I left one minor comment. I would like to add a test for this behavior. @yuanwhy are you interested in adding a test for this? I think that it should be added to [this test file](https://github.com/atom/atom/blob/master/spec/main-process/atom-application.test.js).

yuanwhy:
@maxbrunsfeld I renamed the event name. I will appreciate it if you could add a test for this behavior, because I am not good at the test code

src/application-delegate.coffee
```diff
@@ -126,6 +126,8 @@ class ApplicationDelegate
     loadSettings = getWindowLoadSettings()
     loadSettings['initialPaths'] = paths
     setWindowLoadSettings(loadSettings)
+    ipcRenderer.send("did-change-paths")
+
```
thomasjo: 🔥 this empty line

PR #12550, opened 2016-08-29
Run script/bootstrap from script/build

nathansobo:
This is how things worked previously, and it seems both convenient and low-cost. In the [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) standard, there are examples of [scripts calling other scripts](https://github.com/github/scripts-to-rule-them-all/blob/master/script/test#L20).

/cc @as-cii @lee-dohm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
   - nvm install $NODE_VERSION
   - nvm use --delete-prefix $NODE_VERSION
   - npm install -g npm
-  - script/bootstrap
   - script/build --create-debian-package --create-rpm-package --compress-artifacts
 
 script: true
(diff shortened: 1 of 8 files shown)


lee-dohm:
Looks good to me!

as-cii:
I guess not running `script/bootstrap` in `script/build` doesn't really buy us anything, except for a little more explicitness. Originally I was a little concerned about potentially passing CLI arguments to `script/build` that applied only to `script/bootstrap`, but since bootstrap doesn't take any parameter I think it's fair to 🚢 this so that it matches the previous behavior.

I have also updated docs to not mention `script/bootstrap` and just refer to `script/build` (as it was before). Thanks, @nathansobo.

as-cii:
Unfortunately, running `script/build` will require eagerly all the javascript files which might reference dependencies that are not installed yet. `script/bootstrap` was designed so that it could run in an environment without any installed dependencies, but `script/build` was not.

I'll adjust this in a separate pull-request to run `script/bootstrap` as a separate process and require dependencies after it exits successfully.

PR #12551, opened 2016-08-30
Always copy item when splitting from the context menu

matthewwithanm:
This updates the context menu so that splitting behaves consistently, regardless of whether the context menu is used from a text editor or not.

The current behavior comes from #7953 but, as @maxbrunsfeld pointed out when we talked about this, the purpose of that commit was to simplify the command names, so the UX change probably wasn't intentional.

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -237,10 +237,10 @@
   ]
   'atom-pane': [
     {type: 'separator'}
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Thanks for fixing this @matthewwithanm!

PR #12553, opened 2016-08-30
Run script/bootstrap immediately after running script/build

as-cii:
Refs: #12550 

This will ensure all the dependencies used later in `script/build` are installed.

/cc: @nathansobo @lee-dohm

### DIFF
--- a/script/build
+++ b/script/build
@@ -2,6 +2,10 @@
 
 'use strict'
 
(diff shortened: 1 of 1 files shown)


nathansobo:
Oh, yeah, of course it needs to be moved higher.

PR #12562, opened 2016-08-31
Simplify Environment Patching

### LINKED ISSUE #11910: Patch Linux Environment When Atom Is Not Launched From A Shell Environment
Atom can't see my environment variables when starting it via GUI (icon in applications menu)

For example there is problem with go-plus package which can't see GOPATH env variable.

Exec Environment:
Fedora 23 x64
Atom    : 1.7.4
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1
DE: GNOME Shell 3.18.5
shell: ZeroShell

Here is some outputs from Atoms dev console 

* Starting Atom from Gnome:
```
process.env['GOPATH']
undefined
process.env['PATH']
"/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/home/den/.rvm/bin:/home/den/bin"
```

* Starting Atom from Terminal
```
process.env['GOPATH']
"/home/den/projects/gohome"
process.env['PATH']
"/home/den/.autojump/bin:/opt/liteide/bin:/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/home/den/.local/bin:/home/den/bin:/usr/local/go/bin:/home/den/.fzf/bin:/home/den/.rvm/bin"
```
Haven't found here any topic/issue regarding this.
As we can see there might happen some major problems because of Atom can't see user env variables, locate tools, etc..

joefitzgerald:
This PR simplifies the logic used to determine when to patch the environment for users of linux and macOS. Instead of using a whitelist of shells and trying to detect how Atom was launched with environment variable markers, we simply modify `atom.sh` to set a environment variable that is only set when launched from the shell. If this environment variable is not detected, the environment is patched with the environment from a new, interactive, login shell.
- [x] Stop whitelisting shells
- [x] Introduce `ATOM_SUPPRESS_ENV_PATCHING` environment variable, which is only set in `atom.sh`
- [x] Manually test on Ubuntu
- [x] Manually test on macOS

Fixes #11910
Fixes #12535

### DIFF
--- a/spec/update-process-env-spec.js
+++ b/spec/update-process-env-spec.js
@@ -96,16 +96,16 @@ describe('updateProcessEnv(launchEnv)', function () {
         ATOM_HOME: '/the/atom/home'
       }
 
-      updateProcessEnv({ATOM_SUPPRESS_ENV_PATCHING: 'true', PWD: '/the/dir'})
+      updateProcessEnv({ATOM_SUPPRESS_ENV_PATCHING: 'true', PWD: '/the/dir', NODE_ENV: 'the-node-env', NODE_PATH: '/the/node/path', ATOM_HOME: '/the/atom/home'})
       expect(process.env).toEqual({
-        PWD: '/the/dir',
         ATOM_SUPPRESS_ENV_PATCHING: 'true',
+        PWD: '/the/dir',
         NODE_ENV: 'the-node-env',
         NODE_PATH: '/the/node/path',
         ATOM_HOME: '/the/atom/home'
       })
 
-      updateProcessEnv({PWD: '/the/dir'})
+      updateProcessEnv({PWD: '/the/dir', NODE_ENV: 'the-node-env', NODE_PATH: '/the/node/path', ATOM_HOME: '/the/atom/home'})
       expect(process.env).toEqual({
         ATOM_SUPPRESS_ENV_PATCHING: 'true',
         PWD: '/the/dir',
@@ -113,6 +113,30 @@ describe('updateProcessEnv(launchEnv)', function () {
         NODE_PATH: '/the/node/path',
         ATOM_HOME: '/the/atom/home'
       })
+    })
+
+    it('allows an existing env variable to be updated', function () {
+      process.env = {
+        WILL_BE_UPDATED: 'old-value',
+        NODE_ENV: 'the-node-env',
+        NODE_PATH: '/the/node/path',
+        ATOM_HOME: '/the/atom/home'
+      }
+
+      updateProcessEnv(process.env)
+      expect(process.env).toEqual(process.env)
+
+      let updatedEnv = {
+        ATOM_SUPPRESS_ENV_PATCHING: 'true',
+        WILL_BE_UPDATED: 'new-value',
+        NODE_ENV: 'the-node-env',
+        NODE_PATH: '/the/node/path',
+        ATOM_HOME: '/the/atom/home',
+        PWD: '/the/dir'
+      }
+
+      updateProcessEnv(updatedEnv)
+      expect(process.env).toEqual(updatedEnv)
     })
   })
 
@@ -203,7 +227,7 @@ describe('updateProcessEnv(launchEnv)', function () {
         expect(shouldGetEnvFromShell({SHELL: '/usr/local/bin/fish'})).toBe(true)
       })
 
-      it('returns false when the shell should not be patched', function () {
+      it('returns false when the environment indicates that Atom was launched from a shell', function () {
         process.platform = 'darwin'
         expect(shouldGetEnvFromShell({ATOM_SUPPRESS_ENV_PATCHING: 'true', SHELL: '/bin/sh'})).toBe(false)
         process.platform = 'linux'
--- a/spec/update-process-env-spec.js
+++ b/spec/update-process-env-spec.js
@@ -32,9 +32,9 @@ describe('updateProcessEnv(launchEnv)', function () {
 
       const initialProcessEnv = process.env
 
-      updateProcessEnv({ATOM_SUPPRESS_ENV_PATCHING: 'true', PWD: '/the/dir', TERM: 'xterm-something', KEY1: 'value1', KEY2: 'value2'})
+      updateProcessEnv({ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT: 'true', PWD: '/the/dir', TERM: 'xterm-something', KEY1: 'value1', KEY2: 'value2'})
       expect(process.env).toEqual({
-        ATOM_SUPPRESS_ENV_PATCHING: 'true',
+        ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT: 'true',
         PWD: '/the/dir',
         TERM: 'xterm-something',
         KEY1: 'value1',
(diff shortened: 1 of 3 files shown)


thomasjo:
LGTM™ 🚀

thomasjo:
Let's get this in before rolling the 🚃 🚃 ? /cc @atom/feedback @nathansobo @lee-dohm

src/update-process-env.js
```diff
@@ -36,6 +33,10 @@ function updateProcessEnv (launchEnv) {
     for (let key in envToAssign) {
       if (!ENVIRONMENT_VARIABLES_TO_PRESERVE.has(key)) {
         process.env[key] = envToAssign[key]
+      } else {
+        if (!process.env[key] && envToAssign[key]) {
```
nathansobo: So we'll always assign "preserved" environment variables if they're not already defined?

nathansobo:
If I'm reading this correctly, it seems like the name of the variable `ATOM_SUPPRESS_ENV_PATCHING` is a bit misleading. We will update the `process.env` of an existing instance of Atom regardless of this flag. The flag indicates whether we should attempt to invoke the user's shell to retrieve the environment... am I right about that?

nathansobo:
What about an environment variable like `ATOM_LAUNCHED_FROM_SHELL`. That seems like it would more clearly communicate the intent unless I'm missed something.

spec/update-process-env-spec.js
```diff
@@ -124,30 +182,42 @@ describe('updateProcessEnv(launchEnv)', function () {
...
+        expect(shouldGetEnvFromShell({SHELL: '/usr/local/bin/sh'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/bin/bash'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/usr/local/bin/bash'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/bin/zsh'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/usr/local/bin/zsh'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/bin/fish'})).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/usr/local/bin/fish'})).toBe(true)
       })
 
       it('returns false when the shell should not be patched', function () {
```
nathansobo: Maybe this should be phrased `returns false when the environment indicates that Atom was launched from a shell`

spec/update-process-env-spec.js
```diff
@@ -124,30 +182,42 @@ describe('updateProcessEnv(launchEnv)', function () {
...
         process.platform = 'darwin'
-        expect(shouldGetEnvFromShell('/bin/sh')).toBe(true)
-        expect(shouldGetEnvFromShell('/usr/local/bin/sh')).toBe(true)
-        expect(shouldGetEnvFromShell('/bin/bash')).toBe(true)
-        expect(shouldGetEnvFromShell('/usr/local/bin/bash')).toBe(true)
-        expect(shouldGetEnvFromShell('/bin/zsh')).toBe(true)
-        expect(shouldGetEnvFromShell('/usr/local/bin/zsh')).toBe(true)
-        expect(shouldGetEnvFromShell('/bin/fish')).toBe(true)
-        expect(shouldGetEnvFromShell('/usr/local/bin/fish')).toBe(true)
+        expect(shouldGetEnvFromShell({SHELL: '/bin/sh'})).toBe(true)
```
nathansobo: Do we need to test all these now that we have changed the logic? It seems like this test could be combined with the one below it.

spec/update-process-env-spec.js
```diff
@@ -57,31 +60,60 @@ describe('updateProcessEnv(launchEnv)', function () {
...
       expect(process.env).toEqual({
+        ATOM_SUPPRESS_ENV_PATCHING: 'true',
         PWD: '/the/dir',
         NODE_ENV: 'the-node-env',
         NODE_PATH: '/the/node/path',
         ATOM_HOME: newAtomHomePath
       })
     })
+
+    it('allows ATOM_SUPPRESS_ENV_PATCHING to be preserved if set', function () {
```
nathansobo: Why does preserving this variable matter? It seems like we only ever look for it in the incoming environment to determine whether to mutate `process.env`. Does its presence in `process.env` itself ever matter?

nathansobo: I see now, if we *ever* launched from a shell, we'll never shell out for the environment going forward. Makes sense. Disregard above comment.

src/update-process-env.js
nathansobo: Seems like this duplicated assignment could be eliminated by making the conditional:

```
if (!ENVIRONMENT_VARIABLES_TO_PRESERVE.has(key) || !process.env[key]) {
  process.env[key] = envToAssign[key]
}

```

joefitzgerald:
> If I'm reading this correctly, it seems like the name of the variable ATOM_SUPPRESS_ENV_PATCHING is a bit misleading. We will update the process.env of an existing instance of Atom regardless of this flag. The flag indicates whether we should attempt to invoke the user's shell to retrieve the environment... am I right about that?

There are two separate behaviors going on in update-environment:

1. We are potentially patching the environment by launching a new, interactive, login shell and using that environment - if Atom was *not* launched from the shell
1. We are allowing multiple invocations of `atom` from the shell to update the environment with new environment values; we are also allowing for the scenario where you open atom to a directory, and then subsequently open that same directory via `atom ./thedir` and allowing the environment to be updated with the environment from the shell

So, there's patching, and there's updating of the environment. They are different. Suppressing the patching should not suppress the updating. Maybe we need new verbs ;)

src/update-process-env.js
joefitzgerald: This is related to https://github.com/atom/atom/pull/12562#discussion-diff-78041627R91 but I think this new behavior may be too pessimistic. It would prevent the changing of an existing value in process.env which is not the expected behavior. I will write up a spec to validate this assumption and fix if true.

spec/update-process-env-spec.js
joefitzgerald: These were removed in a prior change and that resulted in a regression for zsh. While these may be perceived as redundant, I think they test for the vast majority of shells used by users, and so we should leave the assertions in there to guarantee no regressions occur.

joefitzgerald:
Also worth mentioning: it should be possible for a user to set `ATOM_SUPPRESS_ENV_PATCHING` in their environment used by `launchd` to run apps launched by Dock / Finder / Spotlight to completely suppress environment patching. I believe this will be the minority of users, but I suspect some users will want to do this. It would be nice to provide some other configuration file that could be used to suppress this behavior - possibly residing in `ATOM_HOME` as a future extension / in a subsequent PR.

src/update-process-env.js
```diff
@@ -34,7 +31,7 @@ function updateProcessEnv (launchEnv) {
     }
 
     for (let key in envToAssign) {
-      if (!ENVIRONMENT_VARIABLES_TO_PRESERVE.has(key)) {
+      if (!ENVIRONMENT_VARIABLES_TO_PRESERVE.has(key) || (!process.env[key] && envToAssign[key])) {
```
nathansobo: Why would `envToAssign[key]` be falsy? You're thinking like a blank string environment variable? The key definitely exists because we're looping over keys in the line above.

nathansobo:
> Maybe we need new verbs ;)

We definitely need new verbs. Patching and updating are the same thing. We need a clearer description of what we're suppressing that focuses on how we obtain the thing we're using to update the current environment... `DISABLE_SHELLING_OUT_FOR_ENVIRONMENT`? I'm not happy with the clarity of the existing naming.

src/update-process-env.js
joefitzgerald: Suppose you launch atom from Finder. `ATOM_SUPPRESS_ENV_PATCHING` is not set. It is also a "preserved" environment variable. If you were to subsequently launch Atom via the shell, it would never be set, and would not be preserved in future invocations.

joefitzgerald:
Renamed `ATOM_SUPPRESS_ENV_PATCHING` to `ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT`.

src/update-process-env.js
nathansobo: Right, that explains the `!process.env[key]` part of the expression but not the `&& envToAssign[key]`. It's minor though. I'll just merge anyway.

nathansobo:
Thanks @joefitzgerald. This definitely feels like a more robust approach.

src/update-process-env.js
steakknife: For posterity: That last part is unnecessary because `key` is always unique.

PR #12564, opened 2016-08-31
⬆ language-c@0.53.0

hansonw:
Includes proper syntax highlighting for C++11 string literals.

Apologies if you guys prefer to batch the language bumps all at once - just wanted to make sure this didn't slip under the radar.

cc @50Wliu


50Wliu:
There's already a PR open for it here: #12502.  Just need to resolve some mysterious archive-view failures.

(And yes, I've started to batch the updates together to reduce the workload on CI)

PR #12572, opened 2016-08-31
Update create-debian-package.js

ablakely:
Added support for building debian powerpc packages.


nathansobo:
I have no way of testing this out, but if it works for you I can't see the harm in merging it. Thanks @ablakely!

PR #12573, opened 2016-09-01
Tildify path in title bar

lgeiger:
This will shorten the directory path displayed in the title bar
PR:
bildschirmfoto 2016-09-01 um 02 14 06
Master:
bildschirmfoto 2016-09-01 um 02 23 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "source-map-support": "^0.3.2",
     "temp": "0.8.1",
     "text-buffer": "9.2.12",
+    "tildify": "^1.2.0",
     "typescript-simple": "1.0.0",
     "underscore-plus": "^1.6.6",
     "winreg": "^1.2.1",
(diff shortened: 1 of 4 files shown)


50Wliu:
:-1: on this as it introduces another dependency and because this is unexpected behavior on Windows.

lgeiger:
@50Wliu Got it. I didn't thought about Windows.

iolsen:
I agree with @50Wliu. A version of this that didn't introduce a dependency and didn't modify the displayed path on Windows would get a 👍 from me.

lgeiger:
I opened a PR (https://github.com/atom/fs-plus/pull/32) to include this functionality in `fs-plus`. This won't add another dependency and doesn't have a effect on windows.

lgeiger:
I'm going to close this and reopen when https://github.com/atom/fs-plus/pull/32 is merged.

PR #12580, opened 2016-09-01
Let atom-pane-resize-hanle inherit correct height (fix #7446)

### LINKED ISSUE #7446: pane resize handles overflow over editors
If you have a vertical split, the resize handle overflows to the right, possibly over an adjacent editor. This messes with drag-select. I haven't found out how to solve this. 

The layout to reproduce this:
```
---------------------
|         |         |
+---------|         |
|         |         |
---------------------
```

cc @simurai 

screen shot 2015-06-26 at 10 52 04

frantic1048:
This patch fixes `<atom-pane-resize-handle>`(`the handle`) overflow issue introduced in #7446 .

I was thinking of removing `/static/panes.less#L19` 's `z-index: 3`, but it just hide the handle, its height still overflows.

Then I noticed `the handle` uses `height: 100%` to gain height. The overflowed handle, between two `atom-pane-axis.horizontal.pane-row`, should have inherid it's direct parent, but actual it herited a further parent in higher hierachy `atom-pane-axis.horizontal.pane-column`. `atom-pane-axis` does't have a rule let its direct child uses `height: 100%` to gain expected height.

As a result, I set `position: relative` for `atom-pane-axis` to make atom-pane-resize-handle as it's direct child gain correct height.

screenshot with this patch:

screenshot_20160901_214907

### DIFF
--- a/static/panes.less
+++ b/static/panes.less
@@ -10,6 +10,7 @@ atom-pane-container {
   min-width: 0;
 
   atom-pane-axis {
+    position: relative;
     display: flex;
     flex: 1;
     min-width: 0;


50Wliu:
@simurai

simurai:
@frantic1048 Thanks for looking into this. 🙇 

I'm pretty sure https://github.com/atom/atom/pull/12300 will fix this issue too because the `<atom-pane-resize-handle>` will get `position: relative;` and aren't `absolute` anymore. But not sure how quickly that PR will be merged, so I think it's ok to merge this one already now.

I tried out this PR for a little bit and can't see any side effects. 👍  There might still be other packages trying to position something attached to the `atom-pane-container`. So we can keep an 👀  on that.

PR #12616, opened 2016-09-07
Fix height when splitting panes

simurai:
When splitting a pane **down**, the height isn't evenly split in half, just after resizing the tree-view. Or when some other style is changed in DevTools.. maybe if the change triggers a reflow?

split

> Using `1.11.0-dev-eff5fa6` on macOS.

Adding `min-height: 0;` seems to work. This is probably related to #11866. Or some rendering bug. This PR adds `min-height: 0;` everywhere we already added `min-width: 0;`.

### DIFF
--- a/static/panes.less
+++ b/static/panes.less
@@ -8,12 +8,14 @@ atom-pane-container {
   display: flex;
   flex: 1;
   min-width: 0;
+  min-height: 0;
 
   atom-pane-axis {
     position: relative;
     display: flex;
     flex: 1;
     min-width: 0;
+    min-height: 0;
 
     & > atom-pane-resize-handle {
       position: absolute;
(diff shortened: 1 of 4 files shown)


Ben3eeE:
@simurai I have had all kinds of problems when closing or opening split panes with incorrect sizes and other rendering bugs. I can test this out later today or now if I can do it by just adding something to my stylesheet.

close pane issue

close pane issue

simurai:
@Ben3eeE Thanks. 🙇  You should be able to simulate it with:

```less
atom-pane-container,
atom-pane-container atom-pane-axis,
atom-pane-container atom-pane,
atom-pane-container atom-pane .item-views,
atom-workspace-axis,
atom-text-editor::shadow .scroll-view {
  min-height: 0;
}
```

Ben3eeE:
This does seem to fix size when splitting pane down and sizes when closing in certain configurations, i'll keep it in my stylesheet since it works better and report any issues I might find.

Does not fix the issue with misspell regions being wrong as my 2nd gif seems another issue.

simurai:
ok, let's get this in. Haven't seen any negative side effects.

PR #12619, opened 2016-09-07
Wait for windows' state to be saved before closing the app or any window

as-cii:
This fixes a test failure on #12300 where the new Electron version prevents state from being serialized.

Previously, we used to save the window's state in the renderer process `beforeunload` event handler: because of the synchronous nature of event handlers and the asynchronous design of IndexedDB, this could potentially not save anything if windows close fast enough to prevent IndexedDB from committing the pending transaction containing the state. (Ref.: https://mzl.la/2bXCXDn)

With this commit, we will intercept the `before-quit` events on `electron.app` and the `close` event on `BrowserWindow` (which will fire respectively before quitting the application and before closing a
window), and prevent them from performing the default action. We will then ask each renderer process to save its state and, finally, close the window and/or the app when the IndexedDB transaction completes.

/cc: @atom/core @thomasjo

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -14,9 +14,10 @@ const ATOM_RESOURCE_PATH = path.resolve(__dirname, '..', '..')
 describe('AtomApplication', function () {
   this.timeout(60 * 1000)
 
-  let originalAtomHome, atomApplicationsToDestroy
+  let originalAppQuit, originalAtomHome, atomApplicationsToDestroy
 
   beforeEach(function () {
+    originalAppQuit = electron.app.quit
     originalAtomHome = process.env.ATOM_HOME
     process.env.ATOM_HOME = makeTempDir('atom-home')
     // Symlinking the compile cache into the temporary home dir makes the windows load much faster
@@ -31,6 +32,7 @@ describe('AtomApplication', function () {
   })
 
   afterEach(async function () {
+    electron.app.quit = originalAppQuit
     process.env.ATOM_HOME = originalAtomHome
     for (let atomApplication of atomApplicationsToDestroy) {
       await atomApplication.destroy()
@@ -368,6 +370,23 @@ describe('AtomApplication', function () {
     })
   })
 
+  describe('before quitting', function () {
+    it('waits until all the windows have saved their state and then quits', async function () {
+      mockElectronAppQuit()
+      const dirAPath = makeTempDir("a")
+      const dirBPath = makeTempDir("b")
+      const atomApplication = buildAtomApplication()
+      const window1 = atomApplication.launch(parseCommandLine([path.join(dirAPath, 'file-a')]))
+      await focusWindow(window1)
+      const window2 = atomApplication.launch(parseCommandLine([path.join(dirBPath, 'file-b')]))
+      await focusWindow(window2)
+      electron.app.quit()
+      assert(!electron.app.hasQuitted())
+      await Promise.all([window1.lastSaveStatePromise, window2.lastSaveStatePromise])
+      assert(electron.app.hasQuitted())
+    })
+  })
+
   function buildAtomApplication () {
     const atomApplication = new AtomApplication({
       resourcePath: ATOM_RESOURCE_PATH,
@@ -381,6 +400,20 @@ describe('AtomApplication', function () {
     window.focus()
     await window.loadedPromise
     await conditionPromise(() => window.atomApplication.lastFocusedWindow === window)
+  }
+
+  function mockElectronAppQuit () {
+    let quitted = false
+    electron.app.quit = function () {
+      let shouldQuit = true
+      electron.app.emit('before-quit', {preventDefault: () => { shouldQuit = false }})
+      if (shouldQuit) {
+        quitted = true
+      }
+    }
+    electron.app.hasQuitted = function () {
+      return quitted
+    }
   }
 
   function makeTempDir (name) {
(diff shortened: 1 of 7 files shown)


src/main-process/atom-application.coffee
```diff
@@ -322,6 +324,8 @@ class AtomApplication
 
     @disposable.add ipcHelpers.on ipcMain, 'did-cancel-window-unload', =>
       @quitting = false
+      for window in @windows
+        window.didCancelWindowUnload()
```
thomasjo: Might not be more readable, and certainly not important, but these two lines could be compacted into
```coffee
window.didCancelWindowUnload() for window in @windows
```

thomasjo:
LGTM! 🚀

src/main-process/atom-application.coffee
thomasjo: Oh, and we should probably call it something else to avoid shadowing `window` global? Better safe than sorry 🙈

as-cii: > Oh, and we should probably call it something else to avoid shadowing the window global? Better safe than sorry

I think the main process doesn't have a `window` global, so we should be good. 👍

spec/main-process/atom-application.test.js
```diff
@@ -368,6 +370,23 @@ describe('AtomApplication', function () {
     })
   })
 
+  describe('before quitting', function () {
+    it('waits until all the windows have saved their state and then quits', async function () {
```
nathansobo: Beautiful test!

src/application-delegate.coffee
```diff
@@ -244,6 +244,17 @@ class ApplicationDelegate
     new Disposable ->
       ipcRenderer.removeListener('context-command', outerCallback)
 
+  onSaveWindowState: (callback) ->
```
nathansobo: What about renaming this `onSaveWindowStateRequest` to clarify that it's not called as a result of saving state, but as a result of it being requested by the main process. If you read the code that's clear but it would be good to telegraph that explicitly via the name.

src/main-process/atom-application.coffee
```diff
@@ -233,8 +232,11 @@ class AtomApplication
     @openPathOnEvent('application:open-your-stylesheet', 'atom://.atom/stylesheet')
     @openPathOnEvent('application:open-license', path.join(process.resourcesPath, 'LICENSE.md'))
 
-    @disposable.add ipcHelpers.on app, 'before-quit', =>
-      @quitting = true
+    @disposable.add ipcHelpers.on app, 'before-quit', (event) =>
+      unless @quitting
+        event.preventDefault()
+        @quitting = true
```
nathansobo: What about a more explicit name for this instance variable, like `stateSavedForAllWindows`?

nathansobo: Oh, I see... `quitting` was a pre-existing variable.

src/main-process/atom-window.coffee
```diff
@@ -128,8 +128,12 @@ class AtomWindow
...
 
   handleEvents: ->
-    @browserWindow.on 'close', =>
-      @atomApplication.saveState(false)
+    @browserWindow.on 'close', (event) =>
+      unless @atomApplication.quitting or @unloading
+        event.preventDefault()
+        @unloading = true
+        @atomApplication.saveState(false)
+        @saveState().then(=> @close())
```
nathansobo: Should we wait to assign `@unloading = true` until after the close promise resolves? I could see a weird case where `close` is fired twice before the state finishes saving.

nathansobo:
Quick and awesome work to clear this obstacle @as-cii. ⚡

src/main-process/atom-window.coffee
as-cii: > I could see a weird case where close is fired twice before the state finishes saving.

Yeah, depending on how big the state is (and, thus, how long it takes to serialize it), the user might be able to click the exit button more than once in rapid succession and cause the state to be serialized twice. I don't think it would be extremely, but it seems better to avoid it if we can.

alexandernst:
Why are renderers responsible for writing to disk? A renderer, by definition, is a one-way method. Maybe the entire write-to-disk logic should be moved somewhere else?

as-cii:
> Why are renderers responsible for writing to disk? A renderer, by definition, is a one-way method. Maybe the entire write-to-disk logic should be moved somewhere else?

In this case we're borrowing Electron's terminology to draw a distinction between the main process and the windows (render) processes. Each window is responsible for maintaining its own state and, since renderer processes have also access to the IndexedDB API, they just serialize and save state by themselves.

I guess at some point we might let the main process handle serialization/persistence of windows'  state, but that would come with the cost of transmitting potentially big object graphs via IPC and I am not sure it would be worth it. Moreover, I don't think the IndexedDB API is exposed in the main process, which would probably require us switching to a different backing store.

spec/main-process/atom-application.test.js
steelbrain: Yo @as-cii, is this a new feature in Atom's spec runner? I previously used `it` and stuff from [`jasmine-fix`](https://www.npmjs.com/package/jasmine-fix) to make it accept async callbacks

thomasjo: @steelbrain These are integration tests, and are being executed outside of Atom's test runner harness. They are using a more modern release of Jasmine ✨

nathansobo: These are main process tests and use Mocha. Eventually we want to use Mocha for our core render process tests too but its not a priority right now. You can use Mocha today for packages via [atom-mocha-test-runner](https://github.com/binarymuse/atom-mocha-test-runner].

thomasjo: > These are main process tests and use Mocha.

Oh wow, I totally forgot about that. But I remember it now because of the explicit function vs arrow function discussion. 
Thanks for setting the record straight 🙇

PR #12645, opened 2016-09-09
Fix app not quitting on win32 and linux when closing the last window

as-cii:
This regression was caused by a nuance in the way we maintain state in `AtomApplication` for open windows. Specifically, when closing the last window on Windows and Linux, we were explicitly calling `app.quit` _before_ removing such window from the list of open ones. In turn, this caused the new `before-quit` behavior introduced in #12619 to work improperly because it made the application wait on saving the state of a stale window before exiting.

With this pull-request we are fixing that by making sure the stale window is removed before calling `app.quit` in `removeWindow`.

/cc: @atom/core @damieng


nathansobo:
Nice.

PR #12646, opened 2016-09-09
Run main process tests on CircleCI, Travis and AppVeyor

as-cii:
Previously we used to run tests only on CircleCI which, despite providing a pretty extensive coverage for macOS, is insufficient to understand whether something has regressed on other platforms. The ideal solution to this problem would be to run tests on every operating system: our renderer test suite, however, is particularly flaky on Linux and Windows as, with time, it has evolved without multi-platform compatibility in mind.

This pull-request is a first step towards such solution: instead of using Travis and AppVeyor as build-only providers, we will now run main process tests on them. These kind of tests are more coarse-grained than the renderer process ones, but they provide quick feedback and should act as good smoke tests for when we release new versions. We could have used `chromedriver` as well, but it seems like it's quite flaky on both Windows and Linux, so I have put that on hold for the time being (in a later pull-request, it might be interesting to change our current `smoke-spec` to use [electron/spectron](https://github.com/electron/spectron) instead and see if that changes anything).

/cc: @atom/core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,12 @@ git:
 matrix:
   include:
     - os: linux
-      env: NODE_VERSION=4.4.7 CC=clang CXX=clang++ npm_config_clang=1
+      env: NODE_VERSION=4.4.7 DISPLAY=:99.0 CC=clang CXX=clang++ npm_config_clang=1
 
 sudo: false
+
+before_install:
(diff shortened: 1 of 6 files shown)


nathansobo:
⚡⚡ ⚡

iolsen:
💯❗❗

Issue #9923, opened 2015-12-03
Inconsistent file formatting for exceptions on Windows


50Wliu:
I noticed while running specs that the way that there's absolutely no consistency in how the file is displayed in exceptions (eg the `at file:line:column` display)

Examples:
- `C:\Users\user\Documents\GitHub\atom\spec\text-editor-component-spec.js` - Expected, this seems to be super rare though :frowning: 
- `file:///C:/Users/user/Documents/GitHub/atom/spec/project-spec.coffee` - Acceptable, much more common
- `C:/Users/user/Documents/GitHub/atom/spec/text-editor-component-spec.js` - Technically not correct.  This only seems to appear when the file isn't wrapped in parentheses (so it also doesn't have a function associated with it).  Probably just forgetting to relativize the path.
- `file:\C:\Users\user\Documents\GitHub\atom\spec\config-spec.coffee` - ..._what_?  Seems to be the most common too :worried:.  I'm going to take a quick guess and say that the relativizing of paths is just converting `/` to `\`, leading to `file:\\\C` which is escaped to `file:\C`.
- `/C:/Users/user/Documents/GitHub/atom/src/buffered-process.coffee` - Definitely not correct.  This only occurs in custom logging, for example the config and deserializer validation (`source: /C:/file (line)`).  No idea what's happening here.

In addition, outside of specs, there's also a lot of inconsistency:  most exceptions will thankfully be either of the form `C:\` or `file:///C:/` though.  However, the `/C:/` version is displayed for errors from `init.coffee` (`At /C:/Users/user/.atom/init.coffee:4`).  The actual stacktrace contains the `file:///` version though: `at atom-text-editor.atom.commands.add.custom:trigger-error (file:///C:/Users/user/.atom/init.coffee:17:2)`.

Only the `C:\` and `file:///C:/` versions are properly caught by the Notifications package.

PR #12653, opened 2016-09-09
Fix source maps for typescript and babel on Windows

damieng:
The source maps for babel and typescript never worked correctly on windows as the path was treated by the Chromium debugger as relative because it was not prefixed with file:///.  This led to a weird issue where the source map was thought to be in a different folder to the generated JS.

This corrects it using the same approach we use for Coffeescript and also removes the 'App load time' which is the last thing spewing onto the command line at start-up now Electron has been upgraded.

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -49,6 +49,10 @@ exports.compile = function (sourceCode, filePath) {
     Logger.prototype.verbose = noop
   }
 
+  if (process.platform === 'win32') {
+    filePath = 'file:///' + path.resolve(filePath).replace(/\\/g, '/')
+  }
+
   var options = {filename: filePath}
   for (var key in defaultOptions) {
     options[key] = defaultOptions[key]
--- a/src/main-process/main.js
+++ b/src/main-process/main.js
@@ -64,10 +64,6 @@ function start () {
     app.removeListener('open-url', addUrlToOpen)
     const AtomApplication = require(path.join(args.resourcePath, 'src', 'main-process', 'atom-application'))
     AtomApplication.open(args)
-
-    if (!args.test) {
-      console.log(`App load time: ${Date.now() - global.shellStartTime}ms`)
-    }
   })
 }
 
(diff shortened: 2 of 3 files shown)


50Wliu:
This should fix most of #9923, if not all of it, right?

damieng:
Quite how the different system reports what to me looked like a completely invalid path I'm not quite sure.  

I suspect this will remove or reduce the amount of types 4 (file incorrect root) and 5 (windows incorrect root) in your list but this doesn't do anything to resolve the differences between types 1 (windows correct), 2 (file correct) and 3 (windows inverted slashes).

damieng:
As an aside I'd like to see everything as type 2 for consistency both internally and with other platforms.

Ingramz:
@damieng this one needs to be prefixed as well, right?https://github.com/atom/atom/blob/5fc240580130f9dfb738f355a635eb21eb8b0e17/src/compile-cache.js#L102

damieng:
I'm not sure we even need that line/block of code.  It applies it after calling the compiler and all three of our compilers already add the source map url.

Ingramz:
@damieng currently I am seeing two distinct file entries for transpiled files in my dev tools (1.12.0-dev-7929e26).

shot1
Hovering over, says `application-delegate.coffee (from source map)`
Copy link address gives `file:///C:/projects/atom/out/app/src/application-delegate.coffee`

shot2
Hovering over, says `/C:/projects/atom/out/app/src/application-delegate.coffee`
Copy link address gives `/C:/projects/atom/out/app/src/application-delegate.coffee`

Adding the `file://` prefix to previously mentioned position merged those two entries. I haven't tried without the block. What confuses me however is that the extension for those files is `.coffee` and contents of them is plain javascript.

Ingramz:
Update: the highlighted block is not needed (can/should be removed) and the issue exists because src directory is missing `.coffee` files, making debugging practically impossible.

PR #12661, opened 2016-09-11
Remove window load time console output

thomasjo:
The _"Window load time: ..."_ console message doesn't really provide any tangible benefit, and is generally just annoying as 💩 when running specs.

Additionally, I believe on Windows this is printed in the terminal when Atom is launched via `atom.cmd`.

/cc @atom/feedback @nathansobo @damieng

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -49,7 +49,6 @@
   function setLoadTime (loadTime) {
     if (global.atom) {
       global.atom.loadTime = loadTime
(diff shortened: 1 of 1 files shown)


50Wliu:
:boat:

thomasjo:
Since the builds are 💚 and @50Wliu gave his "blessing", I'm 🚢 this

PR #12674, opened 2016-09-13
Don't include destroyed panes in result of PaneContainer.getPanes() after PaneContainer is destroyed

matthewwithanm:
This is a failing test to illustrate a bug in `getPanes()`. I'm not sure what the right fix for this is 😞

Here's the problem as I understand it now: when you call `PaneContainer::destroy()`, it destroys all of its Panes. Each Pane will notify its PaneAxis, which will then remove the Pane from its list of children. This all works as long as the container's root is a PaneAxis, however, when you go from 2 to 1 Panes, the PaneAxis will reparent the last Pane making it the container's root. From that point on, `PaneContainer::getPanes()` will delegate to [the (destroyed) Pane](https://github.com/atom/atom/blob/v1.11.0-beta5/src/pane.coffee), which returns a list that includes itself.

My gut says that the right way to address this is to never have a Pane be the root (e.g. always contain it in a PaneAxis) but I don't really know.

### DIFF
--- a/spec/pane-container-spec.coffee
+++ b/spec/pane-container-spec.coffee
@@ -255,7 +255,9 @@ describe "PaneContainer", ->
     it "invokes the given callback when panes are destroyed", ->
       container = new PaneContainer(params)
       events = []
-      container.onDidDestroyPane (event) -> events.push(event)
+      container.onDidDestroyPane (event) ->
+        expect(event.pane in container.getPanes()).toBe false
+        events.push(event)
 
       pane1 = container.getActivePane()
       pane2 = pane1.splitRight()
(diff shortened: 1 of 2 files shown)


nathansobo:
@matthewwithanm We've added a fix for this. I can't say it's the most gorgeous thing ever, but this is a pretty weird corner case, because the behavior we want while the PaneContainer is alive is for it to always have a root pane. If we nulled out the pane we'd have to have null-checking anyway, so I just opted to localize the behavior to the specific scenario of the `PaneContainer` being destroyed. Let me know what you think.

By the way, creating a PR with a failing test was a really great approach. Thanks for that.

matthewwithanm:
I like it 😊

nathansobo:
Kicking off the build again since it appears to have timed out.

nathansobo:
Looks like a legit failure. Will investigate.

PR #12689, opened 2016-09-14
Upgrade Electron to v1.3.6

thomasjo:
This might need some manual testing. Judging by the [changelog](https://github.com/electron/electron/releases/tag/v1.3.6) I don't expect this to cause any issues, but it doesn't hurt to be extra careful.

/cc @atom/feedback

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "1.3.5",
+  "electronVersion": "1.3.6",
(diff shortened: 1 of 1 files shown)


thomasjo:
I've tested this for 24+ hours now, noticed zero regressions (as expected from the changelog), so I'm hitting _The Big Green Button™_

nathansobo:
@maxbrunsfeld and I are seeing massive slowdowns in the dev tools with 1.3.5 so I hope they're fixed in this version. Otherwise we need to figure that out before releasing.

PR #12694, opened 2016-09-14
Avoid adding non-existent project directories

### LINKED ISSUE #10297: Opening files that use atom:// URIs create fake folders
Opening Settings or any other file that uses a URI will cause a fake folder to appear in the Tree View (eg config, init-script) and `Settings - atom://config` or similar to show up in the titlebar. Reproduce by using a new window.

Tested using safe mode and also reported in #10194, atom/tree-view#690.

maxbrunsfeld:
Closes https://github.com/atom/atom/issues/10297
Depends on https://github.com/atom/service-hub/pull/9

/cc @peterhal @nmote @mostafaeweda

This implements the plan we described in https://github.com/atom/atom/issues/10297#issuecomment-245997900. This means that Nuclide's `atom.directory-provider` service hook will be called earlier in Atom's startup process, before project deserialization and notably before Nuclide's `activate` method is called. It also means that custom `Directory` instances for which `.existsSync()` returns `false` will _not_ be added to the project, so you may need to handle the state where the user has not yet authenticated specially in that method.

Also note that the `atom.directory-provider` service must be specified in Nuclide's top-level package.json (rather than being registered dynamically) in order to be recognized.

Sorry for requiring you to make these changes, but we really need to fix #10297, which was introduced by changes that we made together in order to accommodate Nuclide in the first place, so it doesn't seem too unreasonable.


nathansobo:
Still having a rough time on Travis. I tried changing `reload` to wait for the entire Application to load and got it to pass once with a focused test. But now it seems to be failing again. We'll have to pick it back up next week.

maxbrunsfeld:
OK, finally all 🍏. Gonna re-run the travis and appveyor builds one time extra, just to be sure.

nathansobo:
✨✨✨

Issue #12652, opened 2016-09-09
Complete loss of subpixel antialiasing


50Wliu:
### Prerequisites
- [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
- [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?
### Description

After the Electron 1.3.5 upgrade, I've noticed a complete loss of subpixel antialiasing in Atom.  Code, tree-view, status-bar, you name it are all grayscale antialiased.  If there's any way I can help debug this please let me know because it's starting to hurt my eyes :(.
### Steps to Reproduce
1. `atom --safe`
2. Zoom in using Magnify

**Expected behavior:** Colorful subpixel antialiasing

**Actual behavior:** Grayscale :(
### Versions

You can get this information from executing `atom --version` and `apm --version` at the command line. Also, please include the OS and what version of the OS you're running.

```
Atom    : 1.12.0-dev-4f1833a
Electron: 1.3.5
Chrome  : 52.0.2743.82
Node    : 6.3.0
```

```
apm  1.12.5
npm  3.10.5
node 4.4.5
python 2.7.12
git 2.9.2.windows.1
visual studio 2015
```

Windows 10 x64.

damieng:
Tested master on my system today with ClearType turned on...  I got ClearType on the editor and the menu drop-down and tooltips. But regular AA on tree-view, menu bar, tabs.

Are you running high DPI?  On my high DPI display (Surface Book) everything was regular AA not ClearType.

50Wliu:
Sorry, forgot to mention that.  Yes, I'm on a high-DPI display (150% scaling).

Issue #7261, opened 2015-06-13
Thin or small fonts rendered on Windows appear colorfully blurry.


xFleury:
![Image of blurry/colorful font rendering](http://i.imgur.com/Crkn7gz.png)

Until looking through the code, I had thought Atom was using Microsoft's WPF graphics rendering interface for the GUI. In which case the solution is simply setting the text-formatting option to "Display" rather than the default "Ideal".

www.hanselman.com/blog/WPFAndTextBlurrinessNowWithCompleteClarity.aspx

I don't know what's causing blurry fonts on Atom, but it's certainly not my system configuration to blame; I'm not using any unusual settings/hardware; default DPI with a 1440x900 display. Other software (Notepad++, Visual Studio, Eclipse, etc) render fonts just fine.

as-cii:
Hey @xFleury, thanks for leaving your feedback! :bow: We are aware that there are some issues with text rendering on Windows and we have already planned to work on them, although they usually arise with high-DPI screens.

Is this something that appeared recently? If yes, would you help us tracking down which version was the culprit? 

Atom doesn't use WPF internally, but it simply delegates to Chromium (and its internals) the hard work of rendering and painting on screen. Thus, I am afraid we're out of luck with the "Text Formatting" option. By having a look at the screenshot you provided, though, I can tell that the text is being render using sub-pixel antialiasing (which normally is a good thing). As a temporary workaround, you may try to force Chrome to ignore this setting by dropping the following snippet into your `styles.less`:

```less
atom-text-editor::shadow .lines {
  -webkit-font-smoothing: antialiased;
}
```

This will make text rendering fall back to grayscale anti-aliasing and should, at least, prevent this unfortunate *colorful blurriness* you're experiencing. Not sure, though, if it will make the experience reasonably acceptable.

My hope is that we will be able to fix this once for all as soon as we complete #6055: that will enable us to make use of the recent `subpixel-font-scaling` feature which, based on the [IE 9 blog](http://blogs.msdn.com/b/ie/archive/2010/11/03/sub-pixel-fonts-in-ie9.aspx), fixes blurriness for *normal* screens as well as high-DPI ones.

/cc: @nathansobo @paulcbetts

paulcbetts:
The text on non-high DPI screens (i.e. 96DPI) actually looks pretty alright, it seems unique to high-DPI screens

nathansobo:
[...]

Issue #3869, opened 2014-10-17
Fonts are a bit blurred compared to other editors


valsamakis:
I am using Windows 8.1 and I always use Consolas fonts for source code.
Antialiasing seems to not working for atom, so if I compare how the fonts are look like with
other editors there is a difference against atom. To see the difference put next to each other a code
opened with Eclipse or Intellj with code open with atom. Even Gvim looks better. This is a major drawback.

mnquintana:
Just out of curiosity, is this still the case with Atom 0.177.0, and do you experience the same thing with Brackets?

valsamakis:
Yes I think this is the case. Brackets is a little better

paulcbetts:
I've found that if you have a High DPI (scaled) display and you have it at the default of 1.5x scaling, you'll see this. Setting it to 2x (i.e. usually one notch toward "Larger" than recommended), it's cleared up. You can check this via opening the console and printing the value of `window.devicePixelRatio`

50Wliu:
Most likely fixed via #15372.

Issue #2833, opened 2014-07-05
High DPI handling (blurry window content due to scaling)


batjko:
Hi guys,

I am working on a laptop with a high DPI screen (Sony Vaio Pro 13, 1920x1080 resolution,166 PPI), and many programs that don't cater for it will be automatically scaled up by Windows (and assume other OS's as well?).

Here's what that looks like on Win8.1 build, with the example of pane titles:
![image](https://cloud.githubusercontent.com/assets/7287076/3486263/8f0c2d54-042b-11e4-9ab2-532ee799f781.png)

Now, of course I can turn the scaling off in the atom.exe compatibility settings which makes the typography much smaller though, and I wanted to log this issue anyway, in case someone knows what to do about it in core.

Also, note that the tree view does not suffer from this, nor the actual code in the editor panes, btw.

Cheers

darklow:
I fixed blurry fonts problem by adding /force-device-scale-factor=1 to app shortcut.
UI is now very small, but at least fonts are not blurry.
"C:\Program Files (x86)\Spark Dev\atom.exe" /force-device-scale-factor=1

DiThi:
@darklow That creates a new file with name `/force-device-scale-factor=1` to me. I have `atom.exe` in 
`AppData\Local\atom\app-0.175.0` (same issue with 0.174).

darklow:
@DiThi yes, it opens new tab for me too, but at least it fixes blurring. I haven't found any other way yet.

paulcbetts:
I'm seeing this too now, and it appears that enabling DirectWrite fixes this issue, change `src/browser/atom-window` and switch `direct-write` to `true` to see the effect (must restart app, not just Reload)

CMCDragonkai:
Is there a way to zoom into Atom on Windows 8? Right now everything looks really small on my surface pro.

jcarmena:
@CMCDragonkai, I have tried this:

body { zoom: 2.3; -moz-transform: scale(2.3); -moz-transform-origin: 0 0}

looks good, but then the mouse events goes crazy (Ubuntu)

alexanderadam:
calling atom with `$ atom --force-device-scale-factor=1` also fixes blurry fonts for me in Ubuntu. Is there any way to set it as a default in the config?

CreamyCookie:
If you have this problem, maybe installing this awesome package could help: https://atom.io/packages/hidpi

barryvan:
[...]

PR #12696, opened 2016-09-15
Upgrade Electron to v1.6.x

### LINKED ISSUE #4084: cmd+ and cmd- should keep you at the same line in the file

### LINKED ISSUE #8006: Emoji image support
Can we get the text editor to show emoji (the Unicode symbols more specifically, not markdown emoji) as images? I didn't see a plugin for it...

### LINKED ISSUE #11692: Uncaught Error: Cannot read property 'object' of undefinedTypeError: Cannot read property 'objec...
1. ...
2. ...

**Atom Version**: 1.7.3
**System**: linux 4.5.1-1-ARCH
**Thrown From**: Atom Core
### Stack Trace

Uncaught Error: Cannot read property 'object' of undefined
TypeError: Cannot read property 'object' of undefined
    at ObjectsRegistry.get (/usr/share/atom/resources/atom.asar/browser/objects-registry.js:44:28)
    at EventEmitter.<anonymous> (/usr/share/atom/resources/atom.asar/browser/rpc-server.js:318:31)
    at emitMany (events.js:108:13)
    at EventEmitter.emit (events.js:182:7)
    at EventEmitter.<anonymous> (/usr/share/atom/resources/atom.asar/browser/api/web-contents.js:144:25)
    at emitTwo (events.js:87:13)
    at EventEmitter.emit (events.js:172:7)

```
At /usr/share/atom/resources/atom.asar/renderer/api/remote.js:170

Error: Cannot read property 'object' of undefined
TypeError: Cannot read property 'object' of undefined
    at ObjectsRegistry.get (/usr/share/atom/resources/atom.asar/browser/objects-registry.js:44:28)
    at EventEmitter.<anonymous> (/usr/share/atom/resources/atom.asar/browser/rpc-server.js:318:31)
    at emitMany (events.js:108:13)
    at EventEmitter.emit (events.js:182:7)
    at EventEmitter.<anonymous> (/usr/share/atom/resources/atom.asar/browser/api/web-contents.js:144:25)
    at emitTwo (events.js:87:13)
    at EventEmitter.emit (events.js:172:7)
    at metaToValue (/usr/share/atom/resources/atom.asar/renderer/api/remote.js:170:13)
    at BrowserWindow.remoteMemberFunction [as isFullScreen] (/usr/share/atom/resources/atom.asar/renderer/api/remote.js:110:18)
    at ApplicationDelegate.module.exports.ApplicationDelegate.isWindowFullScreen (/usr/share/atom/resources/app.asar/src/application-delegate.js:103:40)
    at AtomEnvironment.module.exports.AtomEnvironment.isFullScreen (/usr/share/atom/resources/app.asar/src/atom-environment.js:588:39)
    at AtomEnvironment.module.exports.AtomEnvironment.serialize (/usr/share/atom/resources/app.asar/src/atom-environment.js:775:26)
    at /usr/share/atom/resources/app.asar/src/atom-environment.js:984:27
```
### Commands

```
     -6:33.2.0 settings-view:check-for-package-updates (atom-workspace.workspace.scrollbars-visible-always.theme-one-dark-syntax.theme-one-dark-ui)
```
### Config

``` json
{
  "core": {
    "autoHideMenuBar": true,
    "disabledPackages": [
      "symbols-view",
      "path-open",
      "advanced-open-file",
      "angularjs",
      "atom-ctags",
      "atom-pair",
      "atom-terminal",
      "autocomplete-ctags",
      "autocomplete-php",
      "debugger",
      "external-open",
      "format-sql",
      "go-plus",
      "node-debugger",
      "open-in-external-app"
    ],
    "excludeVcsIgnoredPaths": false
  },
  "editor": {
    "invisibles": {}
  }
}
```
### Installed Packages

``` coffee
# User
atom-ternjs, v0.14.1
autocomplete-html-entities, v0.1.0
autocomplete-modules, v1.4.1
autocomplete-python, v1.7.2
autocomplete-ruby, v0.1.0
autocomplete-sass, v0.1.0
color-picker, v2.1.1
docs-snippets, v0.8.0
dojo-complete, v0.4.0
emmet, v2.4.3
grunt-runner, v0.14.0
highlight-line, v0.11.1
highlight-selected, v0.11.2
jquery-snippets, v11.0.0
js-refactor, v0.6.0
language-css, v0.36.1
language-gfm, v0.86.0
language-git, v0.13.0
language-html, v0.44.1
language-ini, v1.16.0
language-javascript, v0.119.0
language-mustache, v0.13.0
language-todo, v0.27.0
latexer, v0.3.0
linter, v1.11.4
linter-coffeelint, v1.1.2
linter-htmlhint, v1.1.4
linter-jshint, v2.1.0
linter-python, v2.1.7
merge-conflicts, v1.4.2
meteor-api, v2.20.0
minimap, v4.23.3
pigments, v0.26.0
pretty-json, v1.0.3
project-manager, v2.9.7
python-indent, v0.4.3
python-tools, v0.6.7
refactor, v0.6.0
remote-edit, v1.8.24
svg-preview, v0.7.4
term3, v0.22.0

# Dev
No dev packages
```

### LINKED ISSUE #12690: Atom CLI triggers OS X Sierra native tabs
### Prerequisites
- [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
- [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [X] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [X] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

### Description

When opening a second atom window from the CLI it triggers OS X Sierra native tabs in some case(s?). I could reproduce this in safe mode and after a factory-defaults reset.

This may or may not be an Electron issue though I could not produce it in another Electron app.
### Steps to Reproduce
1. have a 2 monitor setup
2. open terminal (iTerm in my case) full screen on monitor 2
3. in terminal `cd` to any project
4. in terminal `atom .`
5. atom opens on monitor 1 (**instance 1**)
6. put **instance 1** in OS X fullscreen mode if it is not and keep it in foreground on monitor 1
7. in terminal `cd` to any other project
8. in terminal `atom .`

**Expected behavior:** Atom opens **instance 2** in a second window (fullscreen optional)

**Actual behavior:** **instance 2** opens in a native tab

atom-native-tabs

**Work-around:** in step 6, make sure atom is nowhere in foreground

**Notes:** AFAIK issue could not be produced with one monitor, considering you need atom in fullscreen foreground to trigger this issue. If you have a terminal that can overlay your fullscreen Atom I'm guessing you can also produce this issue.
### Versions

macOS Sierra 10.12 GM

```
Atom    : 1.10.2
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0
```

### LINKED ISSUE #12832: Constant Crash on Ubuntu, reproducable
_Update: reproduction steps described below. problem appears related to the hover bubble over files  in the file tree._

After some time of running `atom --safe` (1.10.2) it crashes. The nature of the crash is that the window remains open but when I resize the window the canvas does not update. I am no longer able to scroll in the tabs open or do anything with atom other than forcefully quit it. The system I am running on in Ubuntu 14.04.4LTS with Gnome 3.12.2.

I've had this issue consistently through various atom versions. Because atom does an okay job in recovering the previous open state, and because most of my projects involve git which means I'm saving a lot, it doesn't bother me too much. But often I'll get frustrated and update my atom (compiling from latest stable) in hopes it is fixed. 

I would love to figure out how to reproduce this but I haven't found it yet. It seems to happen only when I'm working on a tree of files, and have a few tabs open at least. Then after about 10 or 15 minutes I will be scrolling and it will lock up.

### LINKED ISSUE #13265: Scrollbars move to the top of the file when splitting panes or closing split panes
### Prerequisites

* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?

### Description

- When I open a large file and move the cursor to the bottom, and then click View->Panes->Split Right.
I found the scroll bar position of previous pane rewind to the top but the cursor is still on the last line.

### Steps to Reproduce

1. Start Atom using `atom --safe` from `cmd.exe`
2. Open any file and resize the Atom window so you have a vertical scrollbar visible
3. Scroll to the middle of an opened file
4. Run `pane:split-right-and-copy-active-item`
5. Expected: The scrollbars to stay in the same position for both files. Actual: The scrollbars in both files moves to the top of the file
6. Close the right pane
7. Expected: The scrollbar position of the first file to not change. Actual: The scrollbar in the first file is displayed at the top of the file.

scrollbar

**Expected behavior:** [What you expected to happen]
scroll bar position should no change

**Actual behavior:** [What actually happened]
I found the scroll bar position of previous pane rewind to the top but the cursor is still on the last line.

### Versions
Atom    : 1.12.0
Electron: 1.3.6
Chrome  : 52.0.2743.82
Node    : 6.3.0

atom-2

### LINKED ISSUE #13767: Unable to save a CPU Profile
### Prerequisites

### Description

A security error is produced when attempting to collect a CPU Profile through the developer tools in Atom.

### Steps to Reproduce

1. Launch Atom in your local atom repository
2. Press <kbd>Alt+Cmd+I</kbd> to open the Developer Tools
3. Click the Profiles tab in the Developer Tools
1. Ensure "Record JavaScript CPU Profile" is selected
1. Click Start
1. Do a couple random things in the editor
1. Click Stop
1. Observe that a SecurityError regarding temporary storage is displayed
1. Click the Save link next to the new profile record
1. Save to `~/Downloads`
1. Get info on the new `*.cpuprofile` file

**Expected behavior:** The `cpuprofile` file to be longer than zero bytes

**Actual behavior:** The `cpuprofile` file is always zero bytes

### Versions

```
Atom    : 1.15.0-dev-ae9d949ee
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

I also reproduced this using v1.13.1 and v1.14.0-beta4.

### Additional Information

screen shot 2017-02-07 at 10 28 03 am

thomasjo:
### Reported Issues

* [x] Object undefined errors from Electron https://github.com/atom/atom/pull/12696#issuecomment-261263408 and https://github.com/atom/atom/pull/12696#issuecomment-261527755
* [x] Default CSP changed, breaking file-icons and specs
* [x] `os.tmpDir()` deprecation warnings https://github.com/atom/atom/pull/12696#issuecomment-288871751
* [x] Drag image in `tree-view` incorrect (https://github.com/atom/tree-view/pull/1054)
* [x] Drop marker on `tabs` and `tree-view` missing (https://github.com/atom/tree-view/pull/1055, https://github.com/atom/tabs/pull/426)
* [x] Changes from https://github.com/atom/atom/pull/13795 cause exceptions in minimap (https://github.com/atom-minimap/minimap/pull/582)
* [x] Dragging pane items between docks mistakenly shows drop indicators due to new security constraints for the `dragover` event `dataTransfer` object (https://github.com/atom/tabs/pull/437)
* [x] Developer tools stop updating after reloading a window in dev mode
* [x] Issues folding soft-wrapped lines https://github.com/atom/atom/pull/12696#issuecomment-301161445
* [x] Dev tools reload exception (https://github.com/atom/atom/pull/12696#issuecomment-301207979, filed a bug for this in https://github.com/electron/electron/issues/9493)
* [x] Dev tools reload crash (end of https://github.com/atom/atom/pull/12696#issuecomment-301207979, filed a bug for this in https://github.com/electron/electron/issues/9493)
* [ ] White flash when hiding toolbar https://github.com/atom/atom/pull/12696#issuecomment-301182375
* [ ] Warnings from tree-view https://github.com/atom/atom/pull/12696#issuecomment-301208739
* [ ] Consistently crashing when continuing from a devtools breakpoint (filed a bug for this in https://github.com/electron/electron/issues/9078)

Reported issues resolved by #13880:

* [x] Scrollbars misbehaving on first file that is opened/restored https://github.com/atom/atom/pull/12696#issuecomment-251299649
* [x] Shift-scrolling scrolls vertically instead
* [x] Failing `spec/text-editor-component-spec.js` tests

---

During my initial testing on macOS 10.12, this seems to be working. Haven't noticed any regressions.
Would be great if more folks could give this a spin 🙇

/cc @atom/maintainers 

---

Fixes #11692 
Resolves #12690
Fixes #8006
Fixes #12832 
Fixes #13767 
Fixes #4084 (Will be fixed by the editor rendering rewrite)
Fixes #13265 (Will be fixed by the editor rendering rewrite)


thomasjo:
Since Electron v1.4.1 ships with https://github.com/electron/electron/pull/7209, I hope this PR will resolve #11692.

50Wliu:
Observations from around 10 minutes of light usage:
# NOTHING IS BLURRY ANYMORE!
I have beautiful subpixel antialiasing on everything, even the actual code for the first time ever\* :boom:.  Fixes #12652, maybe #7261, #3869, #2833, #2833.

Scrollbars seem to disappear randomly.  Haven't narrowed this down but I did notice one of my files didn't have a scrollbar.

\* Just noticed that the tabs and mini-editors don't have subpixel AA, but that's super minor compared to before.

50Wliu:
Ok, something else related to subpixel AA:
If I scroll all the way to the end of the file, I lose subpixel AA in the editor until I switch tabs.

Ben3eeE:
> Scrollbars seem to disappear randomly. Haven't narrowed this down but I did notice one of my files didn't have a scrollbar.

Have this happening with the vertical scrollbar on the TextEditor active when restoring session. In the case of split panes (I tried left/right) both TextEditors are missing vertical scrollbars.

1. Open Atom `1.12-dev-fcddf83` open some file that has content enough to have a vertical scrollbar.
2. Close Atom.
3. Open Atom `1.12-dev-fcddf83` and keep an eye on the vertical scrollbar. It appears quickly when opening the previous session and then disappears.

If you resize the pane so that the horizontal scrollbar appears or disappears then you can get the vertical scrollbar to appear, but in all other cases this file seems to be missing the vertical scrollbar. 

Tested with both Atom Light and One Dark themes. This does not reproduce in 1.11-beta5. Don't know about other dev versions.

50Wliu:
The disappearing scrollbar issue appears to happen on the first file that is opened/restored.  From the dev tools, the scrollbar (and scrollbar corner) is being given a width of 10px instead of the usual 15px.  If I change it to 15, then the scrollbar appears.  Anything > 10 works.  /cc @simurai
I've also been seeing a related issue where the scrollbar flickers like crazy when moving the mouse between the tab bar and the editor.  I'll try to find a stable repro case for that, but it also happens on the first file.

Now, some more subpixel antialiasing updates.  As I said before, it works for the most part.  Here's when it fails:
1. In a mini-editor
2. (Almost) anywhere with rendered text (eg Settings View, Tabs, Notifications, About).  The tree view and status bar are fine though?
3. Upon reaching the end of the file
4. Upon scrolling with a touchpad or scrollbar.  Mousewheel scrolling is fine.
5. Upon editing the last three visible lines, causing the editor to scroll and triggering case 4).

damieng:
Chrome tends to switch from sub-pixel to regular AA when it expects to need to redraw it several times such as in animation. This is because AA can be cached and re-used whereas sub-pixel depends on the horizontal offset.

simurai:
The scrollbar issue is super weird. Changing the `width` to anything other than the original (`10px`) works.. but when changing `bottom` it alternates. See the sub-pixel (`0.1px`) steps.

scrollbars

simurai:
Another thing I noticed. The native tabs in macOS Sierra seem to work in this Electron version. Problem with the custom title bar is that the native tabs are just on top and don't push the rest down. Also, hard to read with dark themes.

screen shot 2016-10-06 at 4 06 06 pm

Native title bar works fine.

### Steps to reproduce:

1. macOS Prefereces > Dock > "Prefer tabs when opening documents" to "Always":
  screen shot 2016-10-06 at 4 02 36 pm
2. Open another window with `cmd-o`.

### Possible solution

a. Either disable native tabs when custom title bar is enabled.
b. See if it's possible to detect when native tabs get added and then push the content down.

Maybe not _that_ urgent, since you have to enable two things to run into this. Could also be a separate issue.

thomasjo:
@simurai Electron v1.4.2 disabled the native tabs support, so this shouldn't be an issue anymore 🎱

Ben3eeE:
🙈 Drag image 🙈 

dragimage

v3ss0n:
electron 1.x is coming in atom 1.12?

50Wliu:
Correct.

Palabola:
Blurry Text still exist under Windows 7 even with the newest Electron!

Ben3eeE:
Fixes https://github.com/atom/atom/issues/13104.

Ben3eeE:
Seems the drop marker is missing again:
drop marker
drop marker 2

nathansobo:
It would be helpful if we could assemble a list of issues that need to be resolved before we can merge this in the body of the issue.

Ben3eeE:
Updated the body with the issues I could find. @50Wliu if you could also check that would be great.

@50Wliu I can reproduce the scrollbar flickering issue when hovering over the tab bar. There have been a few reports of various things flickering when moving the mouse over the tab bar since 1.11. Maybe we do something strange?

damieng:
@Palabola can you attach a screenshot showing what you mean by blurry text?

50Wliu:
I've removed the dev tools freezing since I can't reproduce it anymore.

thomasjo:
I've narrowed down the problem with the missing drop indicators to being some problem related to `DataTransfer`.

The values set during the `dragstart` event via `event.dataTransfer.setData(format, value)` are empty strings during in the `dragover` event via `event.dataTransfer.getData(format)`.

Haven't yet figured out what has changed in Chromium and how to fix it. Quick, cursory search did not reveal any obvious Chromium bugs.

/cc @Ben3eeE @atom/maintainers

Ben3eeE:
Got this today when on this branch in dev mode.

image

Ben3eeE:
Another object of undefined error today when reloading the Atom window. I was fooling around with my `init.coffee` writing keystroke api code. Previous was when tabbing out of Atom. This time with Electron 1.4.7.

image

nathansobo:
@Ben3eeE I've added the object undefined errors to the list in the issue body.

kevinsawicki:
@damieng just a heads up, Electron 1.4.14 contains the `normalizeAccessKeys` option from https://github.com/electron/electron/pull/8308

Ben3eeE:
Two errors when testing latest artifact here. I don't see these in 1.15.0-beta3.

```
Failed to activate the status-bar package
contentWidth is not defined
Hide Stack Trace
ReferenceError: contentWidth is not defined
    at AtomClockView.adjustElementSize (atom-clock-view.js:97:5)
    at AtomClockView.initialize (atom-clock-view.js:21:10)
    at AtomClockView.start (atom-clock-view.js:14:10)
    at Object.consumeStatusBar (atom-clock.js:45:24)
    at Provider.module.exports.Provider.provide (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\node_modules\service-hub\lib\provider.js:29:52)
    at ServiceHub.module.exports.ServiceHub.provide (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\node_modules\service-hub\lib\service-hub.js:30:20)
    at Package.module.exports.Package.activateServices (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:396:71)
    at Package.module.exports.Package.activateNow (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:221:16)
    at C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:190:32
    at Package.module.exports.Package.measure (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:96:15)
    at C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:183:26
    at Package.module.exports.Package.activate (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package.js:180:34)
    at PackageManager.module.exports.PackageManager.activatePackage (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package-manager.js:550:34)
    at C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package-manager.js:531:29
    at Config.module.exports.Config.transactAsync (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\config.js:337:18)
    at PackageManager.module.exports.PackageManager.activatePackages (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package-manager.js:526:19)
    at PackageManager.module.exports.PackageManager.activate (C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\package-manager.js:508:46)
    at C:\Users\lineri\AppData\Local\atom\app-1.16.0-dev-49ec99d\resources\app.asar\src\atom-environment.js:812:28
```

```
Uncaught TypeError: atom.views.pollDocument is not a function  C:\Users\lineri\.atom\packages\minimap\lib\minimap-element.js:345 
```

thomasjo:
@Ben3eeE Not seeing those errors on macOS 10.12.3, running Atom 1.16.0-dev-49ec99d3b.

Seems like the error is originating from minimap; I guess it's consuming the status-bar service, and the exception is causing a "catastrophic" error chain. Disable minimap to see if you can repro?

Ben3eeE:
I'm guessing the first one is the `atom-clock` package and the second one is `minimap`. Limited time to test right now because I am at work :slightly_frowning_face:. Just logged it here so I don't forget and in case anyone with more time wants to take a look before I have the chance. Maybe not even be related to this branch since I am comparing master to beta.

nathansobo:
I made some changes to the text editor in #13795 which could be causing the exception in question. I'd like to change the design of the editor before upgrading Electron because the scrollbars are behaving badly on the latest Chrome and I'd rather dispense with the complexity than double down on it.

50Wliu:
One more comment regarding subpixel AA.  Any line that is completely italicized will disable subpixel antialiasing on that line as well as surrounding lines (the exact number appears to be dependent on blank lines and other factors that I can't pinpoint at the moment, maybe tiling?).

Ben3eeE:
@nathansobo I added the minimap exception to the list in the body of the issue.

mybks:
@nathansobo I've compiled the tj-upgrade-electron, and the fonts are still not as sharp as in OS, why is so that? and maybe you know if it's going to change in the future in electron, or chromium? Atom has such amount of features, but everytime I switch to it I feel my eyes just start drying.

lee-dohm:
@mybks Chromium uses its own font rendering engine to provide cross-platform consistency, so the font rendering in Chromium, and by extension Atom, will not always match the OS font rendering on your platform. If the font rendering you see in Atom matches Chromium then that is the best we can do. You would have to take a look at the Chromium team's roadmap to see what their plans are for improving font rendering on your platform long-term.

scherii:
Just wanted to point out that there is a  `#enable-harfbuzz-rendertext` in [chrome://flags](chrome://flags) option that, as far as I understand, turns on the [HarfBuzz engine](https://www.wikiwand.com/en/HarfBuzz) (see https://github.com/behdad/harfbuzz).

thomasjo:
> Just wanted to point out that there is an option called #enable-harfbuzz-rendertext in chrome://flags that, as far as I understand, turns on the HarfBuzz engine (see https://github.com/behdad/harfbuzz).

No such option here using Chrome 56 (Ubuntu 16.04). From a cursory glance in the Chromium source, it looks like this is the default OT engine on (nearly) all supported platforms.

scherii:
Yeah, you're right, this is on Chrome Canary 59 on macOS.
The flag `#enable-harfbuzz-rendertext` is macOS-only.

mybks:
@lee-dohm Thank you very much for your response, will take a look on the roadmap. And thanks also @scherii for suggestion

nathansobo:
@thomasjo @50Wliu Based on some developments concerning v8 snapshots, I think it makes sense to go straight to Electron 1.6 as I have on the [`ns-electron-1.6.x` branch](https://github.com/atom/atom/tree/ns-electron-1.6.x).

Basically, snapshots are broken on all platforms for 1.4.6 and above and broken on Windows for all of 1.4.x. Snapshots work perfectly on 1.6 and the rewards of upgrading to 1.6.x are pretty high anyway. I'd like to save the Electron team the hassle of hotfixing 1.4 to support snapshots since it's an outdated branch.

You could reset this branch to that branch if you agree with me.

thomasjo:
> Based on some developments concerning v8 snapshots, I think it makes sense to go straight to Electron 1.6 as I have on the ns-electron-1.6.x branch.
> ...
> You could reset this branch to that branch if you agree with me.

Agreed — and done! :100:

Note that this is somewhat more risky in some sense, but worth it.

50Wliu:
Unfortunately snapshots aren't working for me on Windows after upgrading to Electron 1.6.  It seems to be the same error that appeared before upgrading to Electron 1.3.14.

kevinsawicki:
> Unfortunately snapshots aren't working for me on Windows after upgrading to Electron 1.6. It seems to be the same error that appeared before upgrading to Electron 1.3.14.

Yeah, we haven't cut an Electron 1.6.x release yet with this fix, should come out early next week.

nathansobo:
> Unfortunately snapshots aren't working for me on Windows after upgrading to Electron 1.6

Oops, sorry I misspoke. Maybe you can disable them if you're living on this branch until next week.

kevinsawicki:
> Yeah, we haven't cut an Electron 1.6.x release yet with this fix, should come out early next week.

This is on electron master now and will be included in the 1.6.4 beta next week.

50Wliu:
Looks like we may need to update the CSP:
```
Refused to load the font 'atom://file-icons/fonts/fontawesome.woff2' because it violates the following Content Security Policy directive: "default-src *". Note that 'font-src' was not explicitly set, so 'default-src' is used as a fallback.
```

file-icons is completely broken as a result.
broken-csp-file-icons

50Wliu:
Shift-scrolling (to scroll horizontally) doesn't work.

50Wliu:
Electron 1.6.x uses Node 7, which deprecated `os.tmpDir()`.  I investigated solving that deprecation warning, but ran into some trouble:
* Atom and its packages currently depend on either `temp` or `tmp` for creating temp files (`temp` is the package we use more)
* Both of those packages rely on `os-tmpdir`, which calls `os.tmpDir()`
* `os-tmpdir` is deprecated
* `temp`'s last release was 2 years ago
* `tmp` is striving for backwards-compatibility, all the way back to Node 0.6, and is reluctant to release a 1.0.0 version that will get rid of the `os-tmpdir` dependency

ZahidDev:
@50Wliu Latest version of `node-tmp` at https://github.com/raszi/node-tmp works for me on node version 7.7.4. The os-tmpdir implementation used in the latest version, while depreciated, doesn't rely on `os.tempDir()` nor `os.tempdir()`. Thus being, backwards compatible. I don't seem to get depreciation warnings on Electron 1.6.x

50Wliu:
Ok, I'll take a look.  Thanks!

as-cii:
The deprecation message should have been solved by e6c74d6. If you are still getting it, please delete your `node_modules` folder and rebuild: npm will resolve the correct dependencies and the deprecation notice will go away. This should happen automatically on CIs after this branch gets merged into master: the dependencies fingerprint will be updated as a result of the electron upgrade, and therefore the generated assets will contain the correctly resolved dependencies.

Ben3eeE:
I see this in the console when activating [minimap-autohider](https://atom.io/packages/minimap-autohider)

```
(node:21264) DeprecationWarning: Calling an asynchronous function without callback is deprecated.
process.on @ internal/process/warning.js:21
emitOne @ events.js:96
emit @ events.js:188
process.nextTick @ internal/process/warning.js:47
_combinedTickCallback @ internal/process/next_tick.js:67
_tickCallback @ internal/process/next_tick.js:98
```

as-cii:
> I see this in the console when activating minimap-autohider

After a quick inspection of `minimap-autohider`, it seems like they are not specifying a callback when calling `fs.writeFile` (https://github.com/ansballard/minimap-autohider/blob/master/lib/minimap-autohider.js#L64). Since it's just a deprecation I believe it's not necessary to provide a shim in Atom,  but it may still be good to open an issue on their repository and inform them of the deprecation.

ansballard:
`minimap-autohider` maintainer here. Not a problem, I'll try to get that patched in tonight. Are there docs on why a callback needs to be specified? I can throw a noop in real quick, but if there's some new standard I should be following I might as well do that.

Arcanemagus:
@ansballard You can find the documentation on `fs.writeFile` [here](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback). Note that the callback is marked as required. You would probably have to dig through the Node.js changelogs to find the reasoning for that though.

ansballard:
Yep, marked as required in `7.0.0`, guess I'll add in some error handling. Thanks for the info!

kevinsawicki:
You may want to disable to new `auxclick` event in this branch.

Chrome 55 shipped with it and it causes the `click` event to not fire anymore for middle clicks which I would guess would break some packages or core behavior that handles middle clicks explicitly. It also causes Electron to open the link in a new window when middle-clicked which is probably rarely the desired behavior.

You can disable it and go back to the pre-55 behavior by specifying the following option in the `BrowserWindow` constructor:

```js
webPreferences: {
  // Disable auxclick event
  // See https://developers.google.com/web/updates/2016/10/auxclick
  disableBlinkFeatures: 'Auxclick',
}
```

https://developers.google.com/web/updates/2016/10/auxclick
https://www.chromestatus.com/feature/5663174342737920

bpasero:
@kevinsawicki thanks for sharing, we just hit this issue too. For `webviews` the relevant thing to do seems to be: `webview.setAttribute('webPreferences', 'disableBlinkFeatures=Auxclick');`

kevinsawicki:
> What I am not sure is: If you set this in webview, are the defaults being overridden? And if so, what are the defaults?

The defaults would still apply and whatever you specify is only additive (similar to the `BrowserWindow` constructor).

You could also use: https://electron.atom.io/docs/api/webview-tag/#disableblinkfeatures

```html
<webview src="https://www.github.com/" disableblinkfeatures="Auxclick"></webview>
```

50Wliu:
I'm seeing the dev tools freeze again after reloading the window.  Steps to reproduce seem to be:
1. `atom --dev`
2. Open dev tools
3. Reload window
4. Notice that the dev tools, while not frozen, doesn't update and is essentially useless

bronson:
Another issue fixed by this PR: https://github.com/electron/electron/issues/7403

Right now, both Atom channels (1.16 and 1.17-beta) produce 0-length files when saving a CPU profile.

50Wliu:
Thanks, I've added the Atom issue to the fixed list.

as-cii:
@50Wliu: I can't reproduce the issue you reported regarding dev tools freezing after reloading in dev mode (I tested it on #13880). I marked it as completed here, but feel free to uncheck it if you can still reproduce.

Stanzilla:
don't forget to bump this to 1.6.8, it has some good bug fixes

50Wliu:
Seeing some odd folding behavior.
When soft-wrapped:
1. Folding indicator does not update state.
2. Incorrect folding indicator at the end of the soft-wrapped line that doesn't do anything.
folding-incorrect-state

Unrelated to soft-wrap:
3. Unable to unfold selection
folding-unable-to-unfold

Arcanemagus:
> Failed to activate the pigments package, cannot find highlights-component in the require cache We'll need to introduce a new API to support the use case that @abe33 is monkey patching private implementation details to support. We'll need to merge this before doing that.

That was marked as checked in the rendering re-write, but I can't find a related issue in the `pigments` repo, and didn't see a related issue filed here. Is that being tracked somewhere?

Arcanemagus:
If the toolbar is hidden and temporarily shown, when it hides again there is a brief white flash as the interface is moved up to take over the newly free space but hasn't yet redone the layout:

Reproduces virtually every time, the GIF only captures a few instances though. Note that Atom v1.17.0-beta5 has the same delay in rendering, but the blank area that shows briefly is the same color as the rest of the app background leading to it not being as noticeable.

nathansobo:
@Arcanemagus I checked the box because I added the raw API to *support* pigments, but perhaps I shouldn't have. Could you maybe create an issue on the pigments repository and cross-link it? I'm pretty sure @as-cii had a conversation with @abe33 about this and he's on board to switch to the new APIs, but may not have had time yet.

As for the other bug, noted. Ill add it to the check list. Since it's not a deal-breaker we may merge this before we fix it.

Arcanemagus:
I think I'm seeing a related issue to what @50Wliu reported in https://github.com/atom/atom/pull/12696#issuecomment-292690910.

1. Open Atom in safe mode and open dev tools
2. Reload Atom (<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F5</kbd>)
3. Get a blank window

Clicking in the blank window shows this error in the console of the open Dev tools:
```Uncaught Error: Pane has been destroyed
    at Pane.module.exports.Pane.activate (C:\Temp\atom\out\Atom x64\resources\app\src\pane.js:872:1)
    at HTMLElement.handleFocus (C:\Temp\atom\out\Atom x64\resources\app\src\workspace-element.js:222:32)
module.exports.Pane.activate @ <embedded>:29041
handleFocus @ C:\Temp\atom\out\Atom x64\resources\app\src\workspace-element.js:222
```
image

Even stranger, attempting to close Dev tools causes it to immediately re-appear with another of those errors in the Console.

Also very likely related, but slightly different: If you try to take a CPU profile and while recording reload the Atom window you get an "Editor has crashed" dialog (and the same blank window).

Arcanemagus:
Also on first opening a window (with a saved session state) I see this in the console:
image
```
IntersectionObserver.observe(target): target element is not a descendant of root.
TreeView @ <embedded>:105039
```

Ben3eeE:
@Arcanemagus Thanks, adding this to the list in the body of the PR. Do you see the dev tools reload misbehaving in safe mode?

Arcanemagus:
> Open Atom in **safe mode** and open dev tools

@Ben3eeE Yep 😛.

Arcanemagus:
Can confirm both of the above on a build from https://github.com/atom/atom/pull/12696/commits/96533998e2da5ee8f3adadc86efb48e9405ae6e5:
```
OS: Windows 10 x64
Atom    : 1.18.0-dev-96533998e
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

50Wliu:
Not too sure what happened or if this is isolated to the Electron upgrade but my Find menu looks like this now:
broken-find-menu

Tentative repro steps seem to be opening the menu through the <kbd>alt-i</kbd> accelerator.

steelbrain:
Just saw a little lag on this branch, can anyone else repo? (filesize was too large so had to upload to Google Drive)

https://drive.google.com/file/d/0B_aQKK-B3pCfQjBncWFGcndfQTA/view

as-cii:
> Just saw a little lag on this branch, can anyone else repo? (filesize was too large so had to upload to Google Drive)

I couldn't reproduce the find-and-replace issue, but it might be worth to keep an eye on it and make sure it didn't regress with these changes. I am pretty sure it's not related to the editor rendering rewrite though.

> Not too sure what happened or if this is isolated to the Electron upgrade but my Find menu looks like this now:
> 
> Tentative repro steps seem to be opening the menu through the alt-i accelerator.

Interesting, unfortunately this is not happening for me on macOS. Can you reproduce consistently? @Ben3eeE: if I recall correctly you use Windows as well. Can you reproduce this on your machine?

> Also very likely related, but slightly different: If you try to take a CPU profile and while recording reload the Atom window you get an "Editor has crashed" dialog (and the same blank window).

I have noticed this one too lately, it might have something to do with the new Electron. Interestingly enough, reloading when DevTools are closed works perfectly. I'll file a bug report on Electron and see if we can fix this somehow. 

> Seeing some odd folding behavior.

This should be fixed as of the latest commit. Nice finding! ✨

Arcanemagus:
> Not too sure what happened or if this is isolated to the Electron upgrade but my Find menu looks like this now:

Just a note I couldn't reproduce this either (on Windows).

Ben3eeE:
> @Ben3eeE: if I recall correctly you use Windows as well. Can you reproduce this on your machine?


I am unable to reproduce on Windows 7.

kaiser101:
How long before this change will be merged into the master?

50Wliu:
I can't reproduce the find menu issue in safe mode, so it looks like that one was a false alarm.

nathansobo:
@kaiser101 We're facing some hard crashes with Electron when the dev tools are open. We may merge anyway and figure it out on master though.

nathansobo:
@steelbrain Unclear from your gif... Were you seeing lag scrolling the editor? If so, the lag could be due to longer pauses garbage collecting DOM nodes as we dropped the recycling strategy. @as-cii I wonder if we should revisit that. At some point we're gonna GC DOM and I worry those pauses could get long. It allocates quite a bit of memory. But first priority is these crashes I suppose.

steelbrain:
@nathansobo The editor was scrolling perfectly, the lag was only visible when scrolling search-and-replace tab (there would be white spaces in screen that would be filled later)

nathansobo:
@steelbrain That's definitely a find and replace bug and is worth reporting on that package if it hasn't been already.

as-cii:
Seems like AppVeyor failed but I believe that's just spurious. I will go ahead and merge it into master so that it can bake for a while before putting it on beta.

Thanks everyone for the help and the hard work on this! 🙏

PR #12711, opened 2016-09-17
Fix indentation for lines where scope changes

### LINKED ISSUE #12708: Indents based on previous lines are incorrectly tested using current/next line scope
https://github.com/atom/atom/blob/b97f317cd1cdc42a95c37274ed2818f665fd921f/src/language-mode.coffee#L272-L273

The method `suggestedIndentForTokenizedLineAtBufferRow` is called with `bufferRow`, `line`, `tokenizedLine` containing information about the line that is to be indented. The indentation is based on the indentation of the previous line (duh), but at the highlighted lines `increaseIndentRegex` may not belong to the scope of that previous line.

For instance in Blade it's possible to embed a block of PHP in a following manner

```blade
@php
echo "hello world";
@endphp
```
For sake of simplicity let's say Blade has defined `increaseIndentPattern` is `@php` and `decreaseIndentPattern` is `@endphp` for its scope.

When `suggestedIndentForTokenizedLineAtBufferRow` is calculating the indentation for the line containing `echo` statement, then at the highlighted section it looks for the previous line: `@php` and tries to match `increaseIndentPattern` of the current line, which is of different scope (`.source.php`) and is different from Blade's scope, meaning the `increaseIndentPattern` of `@php` is never hit.

Ingramz:
Fixes #12708 

Here's my attempt at fixing it, which solves my use case.

### DIFF
--- a/src/language-mode.coffee
+++ b/src/language-mode.coffee
@@ -250,20 +250,18 @@ class LanguageMode
     @suggestedIndentForTokenizedLineAtBufferRow(bufferRow, line, tokenizedLine, options)
 
   suggestedIndentForTokenizedLineAtBufferRow: (bufferRow, line, tokenizedLine, options) ->
-    iterator = tokenizedLine.getTokenIterator()
-    iterator.next()
-    scopeDescriptor = new ScopeDescriptor(scopes: iterator.getScopes())
-
-    increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    decreaseIndentRegex = @decreaseIndentRegexForScopeDescriptor(scopeDescriptor)
-    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(scopeDescriptor)
-
     if options?.skipBlankLines ? true
       precedingRow = @buffer.previousNonBlankRow(bufferRow)
       return 0 unless precedingRow?
     else
       precedingRow = bufferRow - 1
       return 0 if precedingRow < 0
+
+    precedingLineIterator = @editor.tokenizedBuffer.tokenizedLineForRow(precedingRow).getTokenIterator()
+    precedingLineIterator.next()
+    precedingLineScopeDescriptor = new ScopeDescriptor(scopes: precedingLineIterator.getScopes())
+    increaseIndentRegex = @increaseIndentRegexForScopeDescriptor(precedingLineScopeDescriptor)
+    decreaseNextIndentRegex = @decreaseNextIndentRegexForScopeDescriptor(precedingLineScopeDescriptor)
 
     desiredIndentLevel = @editor.indentationForBufferRow(precedingRow)
     return desiredIndentLevel unless increaseIndentRegex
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
@Ingramz Thanks for tracking this down! Could you add a test for this? See [this test](https://github.com/atom/atom/blob/bc2292e7ddb5b47c6dee376d94f9d416c1866706/spec/language-mode-spec.coffee#L99) for an example of how the test could be written.

Are you able to reproduce this with any of Atom's built-in languages (perhaps markdown with an embedded code block, or HTML with JavaScript or CSS embedded?

Ingramz:
@maxbrunsfeld The trick is that the scope has to change on a line where the outer language indents one line before the embedded language, which HTML tags should do in combination with JS/CSS. I'll have to see if this holds and then will add tests.

On a related note, any idea why the algorithm skips indenting lines that consist entirely of a comment? I think there's another bug with that approach.

maxbrunsfeld:
> any idea why the algorithm skips indenting lines that consist entirely of a comment?

Are you referring to [this conditional](https://github.com/atom/atom/blob/1ad6a0f50c09c9d51daa6f1e77eb03b3054fb8cf/src/language-mode.coffee#L271)? I think that commented lines are skipped because otherwise, we might falsely think that they contain tokens that would require indentation.

For example, in this code:

```js
if (a) {
  // }
  b()

  // {
  c()
}
```

If we didn't have that conditional, then `b()` would be out-dented incorrectly, and `c()` would be indented incorrectly.

Ingramz:
@maxbrunsfeld Yes, I was referring to that conditional.

That behavior currently breaks indenting block comments like for instance in HTML:

```html
<p>
<!--
I should be indented
-->
<p>
```

Ingramz:
@maxbrunsfeld I finally managed to look into this.

Embedded JavaScript fails because this oddity https://github.com/atom/language-html/blob/2868437114869fd8e0036ad242fe815ab7bed36d/grammars/html.cson#L217-L227

Because tag inside `end` is detected using positive lookbehind, it will stay inside `contentName` scope, which makes it use the `decreaseIndentPattern` for `source.js`.

Embedded CSS fails because the surrounding scope name for embedded CSS `style` tag contains `source.css`, which makes it use the `decreaseIndentPattern` for `source.css`. https://github.com/atom/language-html/blob/2868437114869fd8e0036ad242fe815ab7bed36d/grammars/html.cson#L123

Yay, more bugs.

50Wliu:
File issues please and I'll try to look into those.

Ingramz:
For script tags, an issue exists already https://github.com/atom/language-html/issues/82

Ingramz:
I rebased this pull request to fix the merge conflicts. Also added a check to ignore scopes from leading whitespace. With recent (yet to be released) language-html fixes, it indents everything as desired no matter how the lines have been indented before re-indenting.

The HTML comment example is unsurprisingly still broken. It can be fixed by stripping comment indent patterns from the language-html settings and enforce a policy that we cannot have indent patterns for comments, although it would be better if there was some thought put into better handling comments.

I am currently looking into a way to filter out all contents of comments except for the punctuation tokens at the beginning / end of the comments and trying to apply the patterns on that string instead. This will allow some basic indentation based on comment punctuation tokens, but will not match any of the contents of the comments, which should make it more flexible compared to the current algorithm.

This approach should fix auto-indent for cases such as this:
```javascript
if (a) {
    // }
    b()
    
    c() // {
}
```

aminya:
@ingramz Could you please fix the conflicts again

You can use Github's UI to fix the conflict issues.
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-on-github

Ingramz:
Due to the 3 years of uninterest from the core team and the nontrivial changes making it more difficult to reapply this pull request verbatim, I will respectfully decline and close the pull request instead.

If someone with merging privileges is interested, I can invest the time to make it work again.

PR #12715, opened 2016-09-18
Improve the "Save As" file path

zacharymarshal:
Currently in Atom when you do a "Save As" (or save a new file) it will have a default name of "Untitled".

Sublime had a feature that would take the first 50 characters of your file and make that the default file name when using "Save As".  I loved this feature because when I was creating new files I would tend to start them with something that could be a useful file name.

This is still a work in progress, but I wanted to get the conversation started.

I am trying to find if there is a good way to do some kind of a hook so that the default "Save As" file name could be configured via a package.

I don't think it is possible right now but I might have missed something.  This is my first stab at implementing a feature in Atom. Any help would be appreciated.

Here is an example:

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -662,13 +662,13 @@ class Pane extends Model
           @handleSaveError(error, item)
 
   getDefaultSaveAsPath: (item) =>
-    itemPath = item.getPath();
+    itemPath = item.getPath()
     if itemPath
       return itemPath
 
     return unless @getActiveEditor()?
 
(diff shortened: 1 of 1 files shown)


damieng:
For most languages this just wouldn't make sense as the first line would include all sorts of symbols, preprocessor instructions, comments etc. and they all have characters that could really mess with the path on the file system.

Given this only makes sense for .txt/.md really where the title might be the name of the document and the negative experience for all other systems we're going to pass on this PR.

Issue #12066, opened 2016-06-28
CONTRIBUTING.md Invalid Links


dakshshah96:
### Description

On line 235 and line 236 of the file CONTRIBUTING.md, the links to the JavaScript style guide as well as CSS style guide are invalid.

**Expected behavior:** Links "JavaScript" and "CSS" should point to their respective style guides.

**Actual behavior:** Links "JavaScript" and "CSS" encounter a 404 not found page.

jerone:
Noticed a while back that those styleguides where removed, don't know what's replacing them:
* https://github.com/styleguide/css
* https://github.com/styleguide/javascript

bastilian:
I would assume that these could be replaced with those mentioned on https://github.com/styleguide, which would be http://primercss.io/ for CSS and https://github.com/styleguide/ruby for Ruby.

bastilian:
Ruby. what am i talking about. :D (me needs more coffee). Ok, no JS styleguide found yet.

mltroutt:
The AirBnB styleguides seem to be held in high regard, what about linking those instead?

- https://github.com/airbnb/javascript
- https://github.com/airbnb/css

PR #12734, opened 2016-09-20
Fix for broken link in the contributing document

felixangell:
I was checking out the contributing document and noticed the JS style guide link is dead. I figured the second best thing would be the JS styleguide in the document itself?


felixangell:
Just noticed the CSS one is broken too, should this be changed to point [here](http://primercss.io)? I'm not sure if a redirect is broken because `github.com/styleguide` works perfectly - no sight of a JS styleguide there though.

Ben3eeE:
This has an open issue: https://github.com/atom/atom/issues/12066

felixangell:
Ah I see, I'll close this :)

PR #12750, opened 2016-09-21
Improve “Editor is not responding” dialogue

jtojnar:
When the “Editor is not responding” dialogue appears and user tries to dismiss it using <kbd>Escape</kbd> key, the Atom window is closed instead. This behaviour is counter-intuitive because closing a dialogue should never result in a destructive action.

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -145,9 +145,10 @@ class AtomWindow
 
       chosen = dialog.showMessageBox @browserWindow,
         type: 'warning'
-        buttons: ['Close', 'Keep Waiting']
+        buttons: ['Keep Waiting', 'Force Close']
         message: 'Editor is not responding'
         detail: 'The editor is not responding. Would you like to force close it or just keep waiting?'
+        cancelId: 0 # Keep Waiting
       @browserWindow.destroy() if chosen is 0
 
     @browserWindow.webContents.on 'crashed', =>


maxbrunsfeld:
> closing a dialogue should never result in a destructive action.

:+1: I agree with this.

I think that we actually probably reorder the buttons, such that we wouldn't have to override the default `cancelId` of zero. From the [macOS Human Interface Guidelines for Dialogs](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/WindowDialogs.html):

> The Cancel button is to the left of the action button.

> Also, make sure that the keyboard shortcut Command-period and the Esc (Escape) key are mapped to the Cancel button.

jtojnar:
For GNOME it is the same. On Windows, however, it is the opposite. We would probably need to change the order based on the platform and that would probably be unnecessarily complicated.

Ben3eeE:
On Windows the dialog looks like this:

keepwait

Actions that close the dialog are the same. I think the look of this dialog is very counter intuitive. I would expect clicking the `Close` button to just close the dialog in this case. Maybe it should have the same look as the `Keep Waiting` button but with the text `Force Close` and the current `Close` button changes text to `Cancel` and just dismisses the dialog?

jtojnar:
Yes, using more specific “Force Close” would be even better. I also changed the order of the buttons, it should look the way you describe on Windows and the order of buttons on other platforms should be correct too.

maxbrunsfeld:
The CI failure looked like a flaky test. I restarted the build.

alexrussell:
While we're improving the dialogue box, I had a few extra comments that I wrote in #13191 which @Ben3eeE suggested I post here.

The main issue is that the wording is potentially misleading. While Atom may call the *window* the editor, most people think of the editor as being the actual code editor inside a tab. As such, the idea of "closing the editor" that the dialogue presents would make a lot of people assume that the current tab (only) will be closed. This is (somewhat) consistent with the way Chrome works, for example.

I'm aware that the implementation of Atom's and Chrome's tabs are far from each other - Chrome using a new process per tab, Atom doing everything in one JS process, so I'm sure even being able to 'close' a tab when 'its' JS becomes unresponsive is impossible, as the JS is not constrained to a tab.

As such, my suggestion (at least, assuming the above Chrome-like suggestion cannot be done, as that would be the best) is simply to change the text to explain to the user that the *window* will be closed not the ambiguous "editor".

Ben3eeE:
The crash dialog has the text "The **editor** has crashed". If the text changes here maybe we should also change that dialog.

/cc: @lee-dohm for :eyes: on the suggestion by @alexrussell

Positive07:
"Atom has crashed" sounds better.

I wanted to ask, maybe it's just my imagination, but this dialog stops Atom from continuing loading and the dialog needs to be closed before Atom resumes...

As an example, Chrome shows the dialog but the tab continues loading in the background, if the tab becomes responsive in this process, the dialog is dismissed automatically. This would be great because sometimes Atom just needed one more second to finish loading but the dialog stops this

50Wliu:
> I wanted to ask, maybe it's just my imagination, but this dialog stops Atom from continuing loading and the dialog needs to be closed before Atom resumes...

You are correct in your observations.  I started a PR to fix that but never finished it - if you want to take a look, the basic idea is that you need to make the message dialog asynchronous (https://electron.atom.io/docs/api/dialog/).

sadick254:
Hey @jtojnar Apologies we didn't get the chance to view this on time. The issue was already fixed by https://github.com/atom/atom/commit/fd4dfefb6aab60762d1f921ca74430cdf18ba47c#diff-23441b1e2ecd41b45a9b24a1f2bbf6af72c96396d34409560a5441813591179d.

I am closing this for now. I am happy to reopen is you feel there is something I have missed.

PR #12760, opened 2016-09-22
Make it easier to define user key bindings

maxbrunsfeld:
1. User key bindings now take precedence over package key bindings regardless of their selector specificity.
2. We now throw errors if key binding files contain duplicate keys. This was an easy error to make because it's easy to think of these CSON files as working like style sheets, rather than JSON objects.

Closes https://github.com/atom/atom-keymap/issues/110

🍐  w/ @nathansobo

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "electronVersion": "1.3.6",
   "dependencies": {
     "async": "0.2.6",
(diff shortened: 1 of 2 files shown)


Ben3eeE:
Sounds like it can apply to https://github.com/atom/settings-view/issues/510 in some way.

PR #12765, opened 2016-09-23
Add instructions to install Atom on Arch Linux

TriforceOfKirby:
The Atom package is available on Arch Linux in the official repositories for both [32-bit](https://www.archlinux.org/packages/community/i686/atom/) and [64-bit](https://www.archlinux.org/packages/community/x86_64/atom/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ Currently only a 64-bit version is available.
 The Linux version does not currently automatically update so you will need to
 repeat these steps to upgrade to future releases.
 
+### Arch Linux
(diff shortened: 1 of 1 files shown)


50Wliu:
The packages you are linking are not official and in fact have caused many problems by changing core Atom components to unsupported versions.  They are substantially less stable than the official versions provided by the Atom team at https://atom.io.

See https://github.com/atom/design-decisions/blob/master/unofficial-atom-distributions.md for more information.

PR #12767, opened 2016-09-24
Hidden title-bar

simurai:
This PR is similar to https://github.com/atom/atom/pull/10208 (Hide title bar on OS X) and contiues from https://github.com/atom/atom/pull/11790. It replaces the current "Use Custom Title Bar" checkbox with a "Title Bar" dropdown. It has the following options:
- `native` (default)
- `custom` adapts to themes
- `hidden` **New**! Only the traffic lights are shown

screen shot 2016-10-15 at 7 40 34 pm
### Native

screen shot 2016-10-15 at 7 34 06 pm
### Custom

screen shot 2016-10-15 at 7 32 23 pm
### Hidden (new)

title-bar-custom
## Benefits:

Be able to hide the title bar leaves more room for the editor and looks less cluttered. Also a package like the tool-bar could use the title bar's place instead.

title-bar-custom 2
## Concerns:

When the tree-view is hidden, a vertical bar is shown to make sure the traffic lights don't overlap anything and that there is an area where the window can be dragged.

screen shot 2016-10-25 at 7 08 22 pm
## Affected packages:
- [ ] tool-bar (nothing major, but could still use some tweaks like centering the icons when it's on the left and tree-view is hidden.

/cc @brumm


brumm:
Maybe we could offer the user a choice between `native`, `hidden` and `hidden-inset`?
Maybe translated as `Native Titlebar`, `Custom Titlebar (Compact)` and `Custom Titlebar (Extended)` or similar.

simurai:
Yeah.. Or if we already add a dropdown, also include frameless? And instead of "Extended" we could try to hide it already in core? Probably not that many people want just the bigger title bar.

1. Native Titlebar
2. Custom Titlebar (Compact)
3. Window controls ("traffic lights") only
4. Frameless

`Window controls only` would need some warning that adding top/left panels might cause overlapping with the window controls and it's a bit "on your own risk".

toioski:
@simurai How can I get this result once I've build the app cloning this branch? With 'this result' I mean the tool-bar on the same line as the traffic lights. 

![Screen]
(https://cloud.githubusercontent.com/assets/378023/18807177/79a71afc-827b-11e6-8bc9-68ec69338e59.png)

simurai:
@toioski In your `styles.less` file, you could try:

```less
.title-bar {
  display: none;
}

// tool-bar
atom-panel-container.header {
  border-bottom: 1px solid @base-border-color;
  -webkit-app-region: drag;

  .tool-bar.tool-bar-top {
    overflow-x: auto;
    flex-wrap: nowrap;
    margin-left: 77px;
  }
}
```

simurai:
Found another issue. If you:

- hide the tree-view
- disable "Always Show Tab Bar" (hide Tab Bar if there is only one tab open)

then the traffic lights overlap the editor:

screen shot 2016-10-25 at 3 28 57 pm

Lots of config options are nice but making it work with all the possible combinations is kinda a pain. :sob:

Anyways.. there is now some minimum width on the left `atom-panel-container`:

container

Some wasted space, but prevents overlapping and adds an area to drag the window.

simurai:
Ok, this is ready for review. Two notes:

1. This PR replaces a config option. To not fall back to the new default and keep people's setting (use custom title bar) the same, I added [this](https://github.com/atom/atom/blob/0efcc88b771a4cf62806916ecac94a79cda8ad24/src/main-process/atom-application.coffee#L87-L91). Is that the right way? It could be removed after 2-3 releases.
2. I tried to add specs but couldn't figure it out. For example I tried to test if the `hidden-title-bar` gets added. But not sure if that's possible because I think the test window doesn't use the same window like Atom does (with the hidden title bar).

joh1234:
Thanks so much!
https://goo.gl/JBljdR

mnquintana:
@simurai With `hidden` enabled, where are the window-draggable regions? Is it just the little area around the stoplights and any empty space after the open tabs?

simurai:
Draggable regions are:

- Around the traffic lights
- tab-bar (but not tabs)
- tree-view (but not files/folders)
- status-bar

Currently the whole status-bar is draggable. I quite like it. But thinking about it, we might have to limit to only the empty space in the status-bar. Because if you wanna click on something and move a little your mouse, it might starts to drag the window?

eboracus:
Is it possible, in how the window is constructed, to not have the editor pane linked to the tabbar/1st tab position (left-right start position)? (Referring to image in Concrens section of the first post). If these two can be set independently then it would seem preferable when hiding the title bar (but not the "hide the tab bar if there is only one tab" option) to have the editor pane start immediately on the left of the page and the tabs start where they do now, after the traffic lights.

simurai:
@eboracus Do you mean if "Always Show Tab Bar" is disabled still add some space above the editor to make room for the traffic lights? Yeah, that would be an option, but some people might think that's a bug and want the space to be removed. It also gets tricky because there are packages like [vertical-tabs](https://atom.io/packages/vertical-tabs) that would conflict with the traffic lights. So having a minimum width for the left panel container seems to be the least damaging. If you close the tree-view and don't like the extra space, you could customize it further overriding it in your `styles.less`.

eboracus:
Actually I am meaning in the case where "Always Show Tab Bar" is still enabled, or where in any case there are tabs visible. That in (only) that case maybe it could look something like this:

hidden_title_bar

That is, of course, if the location/placement of the first tab can be 'de-coupled' from the left edge of the editor pane below it. I don't think I've seen this in practice anywhere so not sure if that would be straight forward or ...not. 😕

This format would look similar to the chrome/firefox browser UI on macOS in this respect which, in my opinion at least, look quite elegant in their implementation of this feature. The small space above the tabs that remains a draggable region of the title bar could also be implemented here (maybe as a show/hide configurable option? Or is that something more for themes anyway? :S).

browser_tabs

simurai:
That would certainly be possible. To add some extra space to the tab-bar. It just gets tricky when considering that any package can add something on the left or top. Then the tab-bar needs to be notified that it should remove the extra space.

In any case, it will always be possible to do it in your `styles.less` file. Just as the default, it's a bit risky and might break.

simurai:
Superseded by #13616

saschaeggi:
@simurai where can I set now the hidden-inset option?

PR #12782, opened 2016-09-26
Remove unused variables (obvious ones)

Ingramz:
This is a split from #12189 to reduce cognitive load. All changes should be obvious enough to convince within 30 seconds for any given change that it is ok to remove/change.

Unlike #12189, this does not require up to date CoffeeScript.


50Wliu:
Build is passing, so I'm going to :ship: this one.

PR #12793, opened 2016-09-27
Fix multiple indentation on paste

### LINKED ISSUE #9045: Pasting with a newline in front adds last lines indentation to pasted one

patstew:
This is a copy of #11267 with a small addition to prevent one of the tests failing and only change the behaviour when there's a leading newline. Fixes #9045 #5473.

### DIFF
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -618,7 +618,10 @@ class Selection extends Model
       newLines
 
     initialNewLines = countInitialNewLines(selectionText)
-    startLevel = @editor.indentLevelForLine(@editor.lineTextForBufferRow(start.row + initialNewLines))
+    if initialNewLines > 0
+      startLevel = @editor.indentLevelForLine(@editor.lineTextForBufferRow(start.row + initialNewLines))
+    else
+      startLevel = @editor.indentLevelForLine(precedingText)
 
     if maintainClipboard
       {text: clipboardText, metadata} = @clipboard.readWithMetadata()
--- a/src/selection.coffee
+++ b/src/selection.coffee
@@ -602,19 +602,23 @@ class Selection extends Model
     {start, end} = @getBufferRange()
     selectionText = @editor.getTextInRange([start, end])
     precedingText = @editor.getTextInRange([[start.row, 0], start])
-    if selectionText.length > 0
-      if selectionText.charAt(0) is '\n'
-        newLineChars = true
-        myCounter = 1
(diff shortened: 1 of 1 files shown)


hgdsraj:
Have you tested this for pasting very large files? It seems as though it may significantly slow down performance as n (str.length) approaches infinity.

Raj

patstew:
Nope, I pretty much just nicked it from the referenced PR, and haven't given it much testing. I think what you're suggesting would only be an issue with a paste consisting solely of millions of newline characters though.

hgdsraj:
Ahh I see, I misread the parameters of the while loop. Looks like a pretty
unlikely corner case. Can we think of any faster ways to do this?

Looks okay otherwise.

Planteur:
Any update on this PR? The indentation problem is quite annoying.

patstew:
I've had another go at fixing this - see this package: https://atom.io/packages/fix-indent-on-paste
It would be useful to get feedback on whether that package works as people expect - if so I'll update this PR with the same code

sadick254:
Hey @patstew. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

Issue #12702, opened 2016-09-16
workspace.getActiveTextEditor() etc don't work if the text editor is not a PaneItem


pfitzseb:
The [docs](https://github.com/atom/atom/blob/v1.10.2/src/workspace.coffee#L646-L658) for `workspace.getTextEditors()` suggests that _all_ `TextEditors` anywhere will be returned, just as `workspace.getActiveTextEditor()` should seemingly return the currently active `TextEditor` no matter what.

This is obviously at odds with the implementation, which only considers `TextEditors` which are registered as `PaneItem`s. In the [ink package](https://github.com/JunoLab/atom-ink/), we have a `Console`-class, which extends `PaneItem` and can contain `TextEditor`s as input cells (those are properly registered, I think), but one can't access them via one of the commands above (also see [#75](https://github.com/JunoLab/atom-ink/issues/75) for why this is problematic).

So either the documentation should clarify that only certain `TextEditor`s will be considered or the implementation should be changed to searching the `TextEditorRegistry` or something like that. 

I can have a look at implementing this if you think that'd be a proper solution.

iolsen:
@nathansobo do you have thoughts on whether this is an api or documentation bug? I lean toward documentation.

nathansobo:
Yeah it's a documentation issue. The `atom.workspace` global really implies that we're talking about text editor *pane items*. For other editors, we could potentially add a method to the `atom.textEditors` global which is intended to deal with all editors. An approach that would work today would be to get the focused element via `document.activeElement`, but when the shadow DOM goes away the result will be the hidden input element embedded within the editor and won't be super convenient to resolve to an actual editor, hence my suggestion that we might add an API.

pfitzseb:
Yeah, I was kinda unsure what `workspace` actually implied. Good to have that cleared up! Still don't quite understand the need for getting editor pane items as opposed to editors, but then again I haven't put much thought into this yet.

I think adding some a `getActiveTextEditor` method to `atom.textEditors` would be great, but I don't think we have any real need for that yet -- it's just surprising that some method of `workspace` (which is supposed to "Represent the state of the user interface for the entire window.") only works for certain parts of it.

Maybe it would make sense to deprecate `atom.workspace.getActiveTextEditor` etc in favor of that new API? I'd guess there are some packages out there which should probably work with text editors everywhere but don't right now (and I don't think to many packages contain non-PaneItem text editors anyways).

[...]

PR #12804, opened 2016-09-28
Add textEditors.getActiveTextEditor() method

pfitzseb:
as per #12702. Also clarify (I think) the docs for `workspace.getActiveTextEditor()` etc.

Two caveats:
- I have no idea if the fallback implementation for when there's no shadow DOM anymore actually works.
- No clue if the spec is correct, since I can't run specs locally right now (something about `Failed to execute 'registerElement' on 'Document': Registration failed for type 'atom-styles'. A type with that name is already registered.`).

### DIFF
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -91,6 +91,17 @@ describe('TextEditorRegistry', function() {
     });
   });
 
+  describe('.getActiveTextEditor', function () {
+    it('gets the currently focused text editor', function () {
+      const disposable = registry.add(editor)
+      var editorElement = editor.getElement()
+      jasmine.attachToDOM(editorElement)
+      editorElement.focus()
+      expect(registry.getActiveTextEditor()).toBe(editor)
+      disposable.dispose()
+    })
+  })
+  
   describe('.maintainConfig(editor)', function() {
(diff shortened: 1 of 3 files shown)


50Wliu:
> No clue if the spec is correct, since I can't run specs locally right now (something about Failed to execute 'registerElement' on 'Document': Registration failed for type 'atom-styles'. A type with that name is already registered.).

Check https://github.com/atom/atom/issues/10719#issuecomment-181890766 for how to fix this.

pfitzseb:
Thanks for the link! The spec from this PR does seem to work now.

As an aside: Atom still shows loads of testing errors for me -- is that to be expected? Also, why aren't the specs run on travis/appveyor (I guess the display specs won't work in headless mode or something?)?

50Wliu:
What OS are you on?  The reason the specs aren't run on Travis and Appveyor is because the specs were initially only intended to be run on macOS.  There's been some work to fix that, but a number them still fail on other platforms.

pfitzseb:
I'm on Windows, so I guess that explains it.

ghost (approved):

aminya:
@pfitzseb Is this still relevant? If so, we need to fix the conflicts.

sadick254 (approved):
@pfitzseb Congratulations on your first contribution. 👍

PR #12831, opened 2016-09-30
Fixes search links for GitHub Issues and Pull Requests

philschatz:
https://github.com/issues and https://github.com/pulls URLs ~~stopped working in the last couple of days~~ do not work for non-logged-in users.

This changes the URLs used in atom to point to https://github.com/search.

It might not be a problem if most users of atom are already logged in, but would they know that logging in would fix the problem?


sadick254:
@philschatz Thank you for your submission.

A lot has changed since this PR was opened and the fix is no longer applicable. I am going to close this. Thank you once again.

Issue #12253, opened 2016-07-28
when opening a file with a line number, atom should scroll down to display that line number at the top of the view


niahmiah:
### Description

When opening atom from the command line, to a specific file and line number, atom does not scroll the content in the most desirable manner...
### Steps to Reproduce
1. atom text/someFile.txt:115

**Expected behavior:** I expect Atom to scroll so that line number 115 is at the top of my editor

**Actual behavior:** Atom scrolls to the line number, but only far enough for it to be presented at the bottom of the screen. Furthermore, plugins like JSHint that are displayed at the bottom will cover the desired line number. Therefore, the user still has to scroll the content to view the line number, defeating the entire purpose of this feature.
### Versions

$ atom --version
Atom    : 1.8.0
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1

mnquintana:
I think this is a regression – this (and most other actions that are supposed to scroll a line into view) is supposed to vertically center that line.

martindale:
Any updates on this?

martindale:
One year later, asking for updates again!

martindale:
Another year, another hopeful request...

jasonrudolph:
> I think this is a regression – this (and most other actions that are supposed to scroll a line into view) is supposed to vertically center that line.

Thanks for this clarification, @mnquintana.

#19272 restores the previous behavior. With the changes in #19272, when opening a file at a specific line, Atom will now:

- Attempt to scroll the editor as needed so that the specified line is centered vertically
- If the file is already open and the requested line is hidden by a fold, unfold so that the line is visible

This change is slated for inclusion in the Atom 1.38.

PR #12834, opened 2016-09-30
Fix: Scroll line vertically to the center if opened with initialRow

vedharish:
Fixes issue #12253 

Added `center: true` for autoscroll to get this behaviour.

Also, regarding the specs, I couldn't find any existing behaviour tests that can distinguish between `center: true` and `center: false` for `autoscroll`. The current `initialLine` tests are [here](https://github.com/atom/atom/blob/706eea2f4298e990a81d117414c766be53146251/spec/workspace-spec.coffee#L385) and the `autoscroll tests` are [here](https://github.com/atom/atom/blob/706eea2f4298e990a81d117414c766be53146251/spec/text-editor-component-spec.js#L4814) (These just check if there is a certain amount of scroll amount from top and not whether the actual location of the cursor is in the center of the displayed screen )

Will add specs if someone can suggest how to proceed.

### DIFF
--- a/src/cursor.coffee
+++ b/src/cursor.coffee
@@ -651,7 +651,8 @@ class Cursor extends Model
   changePosition: (options, fn) ->
     @clearSelection(autoscroll: false)
     fn()
-    @autoscroll() if options.autoscroll ? @isLastCursor()
+    if options.autoscroll ? @isLastCursor()
+      @autoscroll(center: ( options.autoscrollCenter or false ))
 
   getPixelRect: ->
     @editor.pixelRectForScreenRange(@getScreenRange())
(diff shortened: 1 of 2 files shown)


martindale:
FYI, @stale shouldn't mark issues as `stale` that have currently attached fixes (like this one).

lee-dohm:
Thanks very much for your effort on this.

Unfortunately, we can't merge this without specs and since it is so out of date there are significant merge conflicts that would need to be fixed. Given that it is in this state, I'm going to close this. If anyone wants help bringing this across the finish line, come find me in the Atom Slack or ping me here.

Thanks again :+1:

Issue #10839, opened 2016-02-16
Add PackageManager::hasLoadedInitialPackages API


steelbrain:
Currently there is no way to get the current state of package loading, we do have a `onDidLoadInitialPackages` but it's unreliable as it won't be triggered unless your package is activated on boot

steelbrain:
Closing in favor of #9522

PR #12897, opened 2016-10-06
Add methods for checking if initial packages have been loaded/activated

matthewwithanm:
This will allow packages to observe the state without having to worry
about subscribing to the events after they've already fired.

Originally suggested in #10839


nathansobo:
✨  thanks for digging into this @matthewwithanm

PR #12898, opened 2016-10-06
Avoid forcing computation of all screen lines when opening a file

maxbrunsfeld:
When opening a large file, the majority of time before the file appears on screen is spent building the DisplayLayer's index of screen coordinates to buffer coordinates. As of https://github.com/atom/text-buffer/pull/175, the DisplayLayer can now compute this information lazily. This PR adjusts our usage of the display layer so that when possible, we avoid indexing the entire buffer.
#### Before (v1.12-beta0)

On my machine, opening a 10MB file used to take over 5 seconds:

opening-before

Once the file was on screen, there was a long period in which the application was unresponsive because of work being done in the `autocomplete-plus` package. This was partially addressed in https://github.com/atom/autocomplete-plus/pull/782, which disables symbol indexing for large files. The following graph only goes up to 6MB because for files larger than that, the editor crashed 😬 .

before-event-loop
#### After

Opening a file:

Now, when opening a 10MB file, it appears on screen in two seconds second:

after-opening

The application remains responsive afterwards:

after-event-loop


as-cii:
Looks good after merging the pull request on text-buffer! 👍

none-da:
:clap:

PR #12922, opened 2016-10-09
Fix for #10474

stereobooster:
If Atom opened with folder e.g. `atom ./` it will save all buffers to workspace tied to this folder.
But if atom opened without folder e.g. `atom` it will prompt if you wan to save all unsaved buffers.
### Testcase
1. Open Atom (without params)
2. Type some text in untitled buffer
3. Quit Atom (Cmd + Q)

Expected behaviour:
1. Shows Prompt whether you want to save or not
2. Depending on user choice close Atom or not 

Current behaviour
1. Close Atom. All text in unsaved buffer is lost

### DIFF
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -879,7 +879,7 @@ class TextEditor extends Model
   # Determine whether the user should be prompted to save before closing
   # this editor.
   shouldPromptToSave: ({windowCloseRequested, projectHasPaths}={}) ->
-    if windowCloseRequested && projectHasPaths
+    if windowCloseRequested and projectHasPaths
       false
     else
       @isModified() and not @buffer.hasMultipleEditors()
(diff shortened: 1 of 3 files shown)


src/text-editor.coffee
```diff
@@ -875,8 +875,8 @@ class TextEditor extends Model
 
   # Determine whether the user should be prompted to save before closing
   # this editor.
-  shouldPromptToSave: ({windowCloseRequested}={}) ->
-    if windowCloseRequested
+  shouldPromptToSave: ({windowCloseRequested, projectHasPaths}={}) ->
+    if windowCloseRequested && projectHasPaths
```
thomasjo: Change `&&` to `and`.

thomasjo (changes requested):

src/text-editor.coffee
stereobooster: done

stereobooster:
@thomasjo anything else? Squish commits or something?

thomasjo:
This seems OK to me, but I'll let someone else have a look as well. /cc @atom/feedback

iolsen:
💯 Awesome! Thanks for this contribution!

Merging now and this will be in the 1.13 beta.

PR #12925, opened 2016-10-10
Return a `Disposable` when stubbing `observePreferredScrollbarStyle`

as-cii:
This should fix the error we are observing on Circle (i.e. https://circleci.com/gh/atom/atom/1390), as `CompositeDisposable` requires all the added disposables to conform to the `Disposable` interface.

/cc: @atom/core


maxbrunsfeld:
🙇  Thanks for fixing this @as-cii!

PR #12926, opened 2016-10-10
6075 focus tab on save prompt

### LINKED ISSUE #6070: Show the unsaved file while displaying the 'unsaved changes' alert at closing.

stereobooster:
This PR is replacement for https://github.com/atom/atom/pull/6075. But this is not Production Ready yet. Created it to discuss.

Closes https://github.com/atom/atom/issues/6070
Closes https://github.com/atom/atom/pull/6075
Depends on https://github.com/atom/tabs/pull/386
### Testcase
1. Open Atom (without params)
2. Type some text in untitled buffer
3. Create new buffer (Cmd+N)
4. Close first buffer without switching to it

Expected behaviour:
1. Switch to first tab
2. Shows Prompt whether you want to save or not

Current behaviour
1. Shows Prompt whether you want to save or not
### Technical details

graph3

As of now all actions after click on close icon happens synchronously - control not returned to browser, so it has no ability to do paint. Solution make reflow actions in asynchronous way so browser would be able to do paints.

As example we can do something like (`tab-bar-view.coffee`):

``` coffeescript
 onClick: (event) ->
    return unless matches(event.target, ".tab .close-icon")

    tab = closest(event.target, '.tab')
    @setActiveTab(tab) if tab.item.isModified?()
    setImmediate => @pane.destroyItem(tab.item)
    false
```

which will allow browser to do paints

graph4

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -579,6 +579,8 @@ class Pane extends Model
     else
       return true
 
+    # @setActiveItem(item)
+
     saveDialog = (saveButtonText, saveFn, message) =>
       chosen = @applicationDelegate.confirm
         message: message
@@ -587,12 +589,15 @@ class Pane extends Model
       switch chosen
         when 0 then saveFn(item, saveError)
         when 1 then false
-        when 2 then true
+        when 2
+          @activateNextRecentlyUsedItem()
+          true
 
     saveError = (error) =>
       if error
         saveDialog("Save as", @saveItemAs, "'#{item.getTitle?() ? uri}' could not be saved.\nError: #{@getMessageForErrorCode(error.code)}")
       else
+        @activateNextRecentlyUsedItem()
         true
 
     saveDialog("Save", @saveItem, "'#{item.getTitle?() ? uri}' has changes, do you want to save them?")


src/pane.coffee
```diff
@@ -550,14 +550,17 @@ class Pane extends Model
...
-      else
-        false
+      promptToSaveItemAsync = ()=>
+        if @promptToSaveItem(item)
+          @removeItem(item, false)
+          item.destroy?()
+          true
+        else
+          false
+      setTimeout promptToSaveItemAsync, 100
```
stereobooster: `@applicationDelegate.confirm` is sync call and it is freeze execution. But setActiveItem uses emitter (`@emitter.emit 'did-change-active-item', @activeItem`), so it is called asynchronously. As a result prompt get showed before switch to active tab happens and then execution freeze. To resolve this need to make `@applicationDelegate.confirm` async. But `promptToSaveItem` is treated as sync in `confirmClose` and `destroyItem`. So to make `promptToSaveItem` we also will need to make async dependent functions. This would be easy with Promises and `await`. But `await` is not available in coffeescript, see https://github.com/jashkenas/coffeescript/pull/1942 and https://github.com/jashkenas/coffeescript/pull/3757

stereobooster:
reimplemented code with generators and promises, but no luck it swallows error silently. Investigating

stereobooster:
Explored different options. Current code implements feature, but it is ugly. I do not think it should be merged as is. But it is reveal underlying problems with `applicationDelegate.confirm` which is synchronous and blocks consequent renders.

But on the other hand clode like this

```coffeescript
for item in @getItems()
  return false unless @promptToSaveItem(item)
true
```

should work in synchronous manner e.g. show prompts to save unsaved files one by one to user.

I think of some kind of async wrapper for `confirm` which will return promise and call synchronous function in `nextTick`. But the same time it should preserve order of calls.

stereobooster:
Current implementation
 1. Does fix the problem
 2. Some tests still fail
 3. It changes interface of some methods, not sure if this acceptable
 4. It feels hacky

As far as I see it now it is worth to try to explore why prompt freeze electron redraw and if there way to overcome this. I believe this is because of synchronicity, but have no formal proof so far

stereobooster:
@Ben3eeE done see also https://github.com/atom/tabs/pull/386/files. Need to add tests, but will do it later

Ben3eeE:
There are many different ways to close buffers and if I understand this correctly it currently only applies when the user clicks the close icon on the tab bar?

Other ways I can think of on the top of my head are: `pane:close`, `core:close`, middle mouse down on the tab bar and all the variations of the `tabs:close-` commands. How does this ensure that we cover every way possible?

stereobooster:
@Ben3eeE I suppose it doesn't. Will test it against suggested testcases and get back

Ben3eeE (changes requested):
@stereobooster Thanks for contributing :bowing_man: At the current state this PR can not be accepted because relying on `setTimeout` in the tabs package is not a good solution. Depending on the tabs package for displaying the tab is not preferred because pane items live in core. A user that disables the tabs package and still use pane items should also have this new behavior. As it is now I believe this would introduce bugs if you disabled the tabs package because we activate the next recently used item without activating the closed tab before that.

In order for this PR to be accepted we need a solution in core for this behavior that works even if the tabs package is disabled. This PR would also need specs for testing this new behavior.

stereobooster:
@Ben3eeE I'm not working on this anymore.

Yes I do not like `setTimeout` too. But to do this functionality we need to make this functionality asynchronous, so that browser would have ability to do paints. I tried setImmediate instead of setTimeout but it doesn't work that good as setTimeout.

> A user that disables the tabs package and still use pane items should also have this new behavior.

Well than part of functionality of tabs package should be dragged to core.

Ben3eeE:
Electron already supports [asynchronous dialog boxes](https://github.com/electron/electron/blob/master/docs/api/dialog.md#dialogshowmessageboxbrowserwindow-options-callback). Closing this PR since it is abandoned and won't get merged in the current state.

PR #12946, opened 2016-10-11
Fixes #11848 Moving line up without line ending

### LINKED ISSUE #11848: Move Line Up
### Description

If I want to move the last line in a file up I can do it by pressing Ctrl+Cmd+Arrow (Mac)

**Proper behavior:** 

If there is a newline character at the end of that line, it will change from 

```
Previous line
Last line
```

to

```
Last line
Previous line
```

**Wrong behavior:** 

But if there is no newline character at the end of the last line, the hotkey will just concatenate it with the previous line

```
Last linePrevious line
```

---

Version 1.7.4
Problem persists in safe mode.

jamsinclair:
## Bug

Fixes the bug described in issue #11848

move-line-up-bug
## Fix

From what I explored it looks like a previous dev has tried to fix this bug. Although because rows are 0 indexed, to select the penultimate row for the range it requires using `-2`. 

I wonder whether it's worth rewriting the line:

``` coffeescript
lines += '\n' unless lines[lines.length - 1] is '\n'
```

I've attempted to add a valid test. Suggestions to improve or better represent the test welcome 😸 

move-line-up-fix

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -2693,6 +2693,22 @@ describe "TextEditor", ->
             expect(editor.isFoldedAtBufferRow(6)).toBeFalsy()
             expect(editor.lineTextForBufferRow(6)).toBe "    if (items.length <= 1) return items;"
 
+        describe "when the last line of selection does not end with a valid line ending", ->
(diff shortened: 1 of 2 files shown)


borishere:
Just why this is not yet merged??

lee-dohm:
Thanks for this!

PR #12977, opened 2016-10-14
Don't share overlayDimensions across instances

nathansobo:
Relates to #8295 

Thanks to @matthewwithanm for finding this error. @matthewwithanm do you think this definitively closes that issue?

@Ben3eeE would you mind testing this out and see if the behavior is improved in that one flickering case you found dragging panes?

cc @lee-dohm

### DIFF
--- a/src/text-editor-presenter.coffee
+++ b/src/text-editor-presenter.coffee
@@ -9,7 +9,7 @@ class TextEditorPresenter
   startBlinkingCursorsAfterDelay: null
   stoppedScrollingTimeoutId: null
   mouseWheelScreenRow: null
-  overlayDimensions: {}
+  overlayDimensions: null
   minimumReflowInterval: 200
 
   constructor: (params) ->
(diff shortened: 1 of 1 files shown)


matthewwithanm:
@nathansobo!! Thanks for such a quick turnaround on this 😊

I can't say for certain if it's a definitive fix for #8295, but it does fix the situation that I was able to repro!

nathansobo:
> but it does fix the situation that I was able to repro!

I'm going to be optimistic then...

Closes #8295

nathansobo:
Thanks again for tracking this down. Pretty embarrassing but I'm happy to have your team on our side! ❤️ :zap:

Ben3eeE:
Thanks for the summon. This does seem to fix the problem.

Thanks @nathansobo

jeremyfa:
Man, I wake up this morning and see this wonderful merge! This is the most annoying issue I am experiencing with atom, I need to test this right now from git and will tell you if it worked. And let's be optimistic yes 👍

jeremyfa:
The fix seems to work :D ! Thanks a lot for the efforts ❤️

Ben3eeE:
@jeremyfa Awesome to hear! Thanks for testing it out and reporting back :bowing_man:

ypresto:
Wow great work...!!! Thanks a lot!!

Issue #12978, opened 2016-10-14
Slow response of build when --help toggles


liuderchi:
### Prerequisites

* [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [X] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [X] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

```
$ script/build --help
```

lags since we need toggle `--help` to see available options for manually building different packages (_windows-installer_, _*.deb_, _*.rpm_).

Plus, some may try `script/build -h` for help but it is not available.

### Steps to Reproduce

1. clone atom then checkout to release branches
2. `$ script/build --help`

**Expected behavior:**

It prints out available options immediately then return.

**Actual behavior:**

It's lagging on terminal result: `Installing script dependencies` and finally prints out available options.
It looks like processing on this line of code (`script/build`, line 7): 

```javascript
require('./bootstrap')
```

### Versions

```bash
$ atom --version
Atom    : 1.11.1
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0

$ atom-beta --version
Atom    : 1.12.0-beta1
Electron: 1.3.6
Chrome  : 52.0.2743.82
Node    : 6.3.0
```

paulmenzel:
Here is the excerpt from my response to issue #13232.

Looking at the code, the usage section should just be moved to the top before requiring `bootstrap`, shouldn’t it?

```
const yargs = require('yargs')
const argv = yargs
[...]

PR #12979, opened 2016-10-14
make script/build parse args first and supports -h options

liuderchi:
1. Related issue is [#12978](https://github.com/atom/atom/issues/12978)
2. changes:
   - parse args first then checks module dependencies to avoid slow response for `script/build --help`
   - add args alias `-h` for `--help`

Closes https://github.com/atom/atom/issues/12978

### DIFF
--- a/script/build
+++ b/script/build
@@ -2,18 +2,11 @@
 
 'use strict'
 
-// Run bootstrap first to ensure all the dependencies used later in this script
-// are installed.
-require('./bootstrap')
-
-// Needed so we can require src/module-cache.coffee during generateModuleCache
-require('coffee-script/register')
-require('colors')
-
 const yargs = require('yargs')
 const argv = yargs
   .usage('Usage: $0 [options]')
   .help('help')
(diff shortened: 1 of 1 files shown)


script/build
```diff
@@ -2,18 +2,11 @@
...
 
-// Run bootstrap first to ensure all the dependencies used later in this script
-// are installed.
-require('./bootstrap')
-
-// Needed so we can require src/module-cache.coffee during generateModuleCache
-require('coffee-script/register')
-require('colors')
-
 const yargs = require('yargs')
```
50Wliu: This will fail without bootstrap.

winstliu (changes requested):

script/build
liuderchi: These imports (line 5-12) are not removed but __moved__ after `argv` definition (line 14-24).

liuderchi: thanks for your review.
Maybe I should find another way to run `build --help` without external library.
this PR will be closed

PR #12982, opened 2016-10-14
README.md Tranlation to Simplified Chinese

LShun:
Hello, this is a translation of the README.md to simplified Chinese. I translated this using Google Translate as well as using my understanding towards the language to correct confusing sentences and words. However, it still needs some tweaking. I did not translate “Flight Manual" because there isn't a Chinese version for it that I can point to. 

Thank you.

### DIFF
--- a/docs/README-zh.md
+++ b/docs/README-zh.md
@@ -1,0 +1,21 @@
+# Atom 文档
+
+![Atom](https://cloud.githubusercontent.com/assets/72919/2874231/3af1db48-d3dd-11e3-98dc-6066f8bc766f.png)
+
+大多数Atom的用户和开发人员的文档都包含在 [Atom Flight Manual](https://github.com/atom/flight-manual.atom.io) 存储库中。
+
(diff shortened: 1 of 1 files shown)


LShun:
Hello, I think its a bit of weird, I only added a .md file with text and formatting but it has 2 failed tasks. Can someone help me to check if I had mistype something? Both of the two tests reported timed out. Thank you.

wjz27 (commented):
In line 20,  (./ apm-rest-api.md)
Here has an extra space between '/' and 'apm'.
Same with line 21.

LShun:
@wjz27 Thanks for letting me know! I made some changes to fix the links.

damieng:
The brackets on lines 20 and 21 are the wrong type - they need to be normal ascii ( to create hyperlinks.

You can see them not rendering correctly if you click Files Changed and then View on the file.

sadick254:
Thanks for the submission!

The addition of simplified Chinese `README.md` is a good idea, however, I don't see it adding a lot of value to what we already have. This would have been useful if the `README-zh.md` was being picked up by the flight-manual.io site but that's, not the case. I will have to close this PR for now.

Thanks again for your contribution.

PR #12994, opened 2016-10-15
Convert ConfigSchema to JS

mnquintana:
In a similar vein as https://github.com/atom/atom/pull/12350, this PR rewrites `src/config-schema.coffee` as JS – it's also a pretty small file, so hopefully there's nothing weird going on here. 😄 

👀 @atom/feedback


fank (approved):

src/config-schema.js
```diff
@@ -1,68 +1,103 @@
-path = require 'path'
-fs = require 'fs-plus'
-
-# This is loaded by atom.coffee. See https://atom.io/docs/api/latest/Config for
```
50Wliu: This comment seems to have disappeared.

mnquintana: Whoops, damn you `decaf`! I'll add it back in

mnquintana: I think that link's still pointing to the right place – that seems to be the only place we document config schemas (even http://flight-manual.atom.io/behind-atom/sections/configuration-api/ just links there)

PR #13002, opened 2016-10-16
Add Homebrew Cask instructions to README

danielbayley:


thomasjo:
I'm 👎 on this since this is not an official means of installation/distribution. /cc @atom/feedback

danielbayley:
So how is it different from _chocolatey_? The Atom cask uses the official release url and compares the checksum…

thomasjo:
I'm against the usage of Chocolatey as well, but at least that one is official. In other words, it's under our control. The cask recipe can change at any time. I might be too paranoid, but I'm still against officially documenting this as a means of installation.

thomasjo:
I've discussed this with the team, and we're not comfortable with officially documenting an unofficial distribution channel like this. The recipe can essentially change at any time, outside of our control. This has been a real problem with certain Linux-based channels in the past (patching in unsupported Electron version etc).

Regardless, thanks for the helping out! We appreciate the effort 🙇

PR #13020, opened 2016-10-19
Add language-nlf to deprecated packages

idleberg:
All features of `language-nlf` have been merged into `language-nsis`. The former will no longer be updated.


lee-dohm:
Hey @idleberg! Thanks for your patience. We're going to go ahead and merge this for now. We're haven't done any investigation to see if there's some sort of build process that has to happen after this. If you wouldn't mind keeping an eye on this and pinging me if the right thing doesn't happen, I would appreciate it!

Thanks again!

PR #13021, opened 2016-10-19
Use flex display in text editor instead of block

abe33:


as-cii:
Thanks, @abe33! ✨ Sorry for the oversight!

as-cii:
It's worth noting that this fixes the minimap not displaying correctly on master.

/cc: @simurai

simurai:
Nice, @abe33! 👍

PR #13042, opened 2016-10-21
macOS: Let atom.sh find itself when run from an .app

skull-squadron:
**Problem** This patch is needed because `node-gyp` doesn't work with a space in `Atom Beta.app` -> renamed to `Atom-Beta.app`. This error occurs:

```
Cannot locate Atom.app, it is usually located in /Applications. Set the ATOM_PATH environment variable to the directory containing Atom.app.
```

**Use case** Developers/others want to multiple versions of Atom side-by-side.

**Solution** When launched from a Mac .app, a valid, symlinked `atom.sh` -> `atom`/`atom-beta` will always be contained in it's .app. This also works when a different `atom-something` symlink points to another atom.sh in an .app.

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -55,7 +55,12 @@ if [ $EXPECT_OUTPUT ]; then
 fi
 
 if [ $OS == 'Mac' ]; then
-  ATOM_APP="$(dirname "$(dirname "$(dirname "$(dirname "$(readlink "$0")")")")")"
+  if [ -L "$0" ]; then
+    SCRIPT="$(readlink "$0")"
+  else
+    SCRIPT="$0"
+  fi
+  ATOM_APP="$(dirname "$(dirname "$(dirname "$(dirname "$SCRIPT")")")")"
   if [ "$ATOM_APP" == . ]; then
     unset ATOM_APP
   else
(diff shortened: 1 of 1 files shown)


atom.sh
```diff
@@ -55,11 +55,17 @@ if [ $EXPECT_OUTPUT ]; then
 fi
 
 if [ $OS == 'Mac' ]; then
+  ATOM_APP="$(dirname "$(dirname "$(dirname "$(dirname "$(readlink "$0")")")")")"
```
steakknife:
```
/Applications/Atom-Beta.app/Contents/Resources/app/atom.sh
        0            1          2        3      4    5

ungb:
hey @steakknife,

I tried the scenario above and wasn't able to see the error you're seeing before you're PR, could this possibly be fixed already or maybe I'm doing something wrong?  Here's a gif of my repro.


test

ungb:
Ah, I was finally able to repro this issue. it's when you open atom from the terminal.  I'll test your pr later today and see if it fixes the issue. 

Thanks,

Bryant

ungb:
I was able to build and test his changes on beta build.

PR #13046, opened 2016-10-21
Project history api, reopen project menu and command

damieng:
## Implements new
- `atom.history` API containing `.getProjects()`, `.clearProjects()` and `.onDidChangeProjects(callback)` public API surface
- 'Reopen Project' menu that lists the **15** most recently opened projects (configurable via `core.reopenProjectMenuCount`)
- `application:reopen-project` command that provides a search across all projects opened
- `application:clear-project-history` command to clear project history
## Notes
- For now you can type `atom.history.importProjectHistory()` in the debug window to build your initial list from Atom's internal window-state history. Am considering making this happen automatically when you first run Atom with no existing project history state file.
- Multiple folder projects appear in the menu as the last name in the paths separated by a comma e.g. 'atom, settings-view' for width. They appear in full in the pop-up palette.
- Reopening a project always appears in a new window as window-state is preserved between projects and it is quite jarring otherwise.
- 'Clear Project History' appears at the top of the menu instead of ideally at the bottom due to limitations of the MenuManager and how it handles separators. Will be addressed in the future once that is resolved.
- Setting `core.reopenProjectMenuCount` does not disable the menu right now because of the same limitations as above (removing at run-time refuses to remove the separator)
### Reopen Project menu

image
### Reopen Project command list picker

image

### DIFF
--- a/menus/darwin.cson
+++ b/menus/darwin.cson
@@ -36,6 +36,13 @@
       { label: 'New File', command: 'application:new-file' }
       { label: 'Open…', command: 'application:open' }
       { label: 'Add Project Folder…', command: 'application:add-project-folder' }
+      {
+        label: 'Reopen Project',
+        submenu: [
+          { label: 'Clear Project History', command: 'application:clear-project-history' }
+          { type: 'separator' }
+        ]
+      }
       { label: 'Reopen Last Item', command: 'pane:reopen-closed-item' }
       { type: 'separator' }
       { label: 'Save', command: 'core:save' }
--- a/menus/linux.cson
+++ b/menus/linux.cson
@@ -7,6 +7,13 @@
       { label: '&Open File…', command: 'application:open-file' }
       { label: 'Open Folder…', command: 'application:open-folder' }
       { label: 'Add Project Folder…', command: 'application:add-project-folder' }
+      {
+        label: 'Reopen Project',
+        submenu: [
+          { label: 'Clear Project History', command: 'application:clear-project-history' }
+          { type: 'separator' }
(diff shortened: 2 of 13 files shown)


simurai:
Nice. 👍 

Maybe something to consider: Opening a previous project in the **same window** by replacing the current project. It would be pretty fast to switch between projects instead of closing and opening a new window.

I've been using [project-plus](https://atom.io/packages/project-plus) and [project-manager](https://atom.io/packages/project-manager) for that and quite like it.

damieng:
I looked at project-plus and it seems what it actually does when you switch is close the current project and open a new one. I think that is required because of the current atom api and because the new project has its own window position/location.

I could definitely add a config option to enable similar behavior in this quite easily.

jerone:
Would it be useful to register if an directory/project was opened in `dev`-mode...

damieng:
Going to merge this for now and consider those two requests separately.

anothernode:
Thanks a lot for implementing this! I would love it even more if it would be possible to make Atom reopen a project in the same window, like @simurai suggested, instead of in a new one.

Ben3eeE:
@anothernode We have an open issue for adding this that you can subscribe to for updates. https://github.com/atom/atom/issues/13277

danielbayley:
On macOS the convention is to have this kind of functionality under `File` > `Open Recent`… I think the implementation of [Zren/open-recent](https://github.com/Zren/atom-open-recent) is better, as it shows files as well as projects, and lets you set how many to show of each… why not merge that into core?

PR #13086, opened 2016-10-27
Upgrade text-buffer to version that uses superstring

maxbrunsfeld:
Depends on https://github.com/atom/text-buffer/pull/177

This adds a dependency on rust at build time.


as-cii:
I think this is ready to go now. Let's wait until we roll the railcars next week, and then 🚢 it! 🚀

Issue #5076, opened 2015-01-15
Move from LESS to Postcss


vanesyan:
In #5075 I proposed to componentizing CSS, So I also wanna propose to use  [POSTCSS](https://github.com/postcss/postcss) instead of LESS, cause it faster and allow us to use new CSS features, like CSS variables, nested media (by plugins) and desclude features that we really don't need.

Also, we can use autoprefixer to prefixing properties that really need it instead of hard typing prefixes like `display: -webkit-flex` and other.

simurai:
Same reason as in #5075, it's already too close to the API freeze to be considered, at least for the 1.0 release.

Good point about the `-webkit` prefix for Flexbox. That could actually already be un-prefixed. I made a new issue for that: #5078. Otherwise autoprefixer isn't really needed, since Atom only runs in Chromium.

vanesyan:
Autoprefixer is for future-proof CSS.)

simurai:
> Autoprefixer is for future-proof CSS.)

Yes, that's right. I'm just not sure if it warrants having another dependency and the extra compile step. Next to Flexbox, there isn't that many other prefixes needed. Maybe for `keyframes` animation or `user-select`. And it looks like Chrome is moving away from prefixing and using flags instead.

danielbayley:
I just thought of this while [comment]ing on #12903…

I think this would be **awesome** for `2.0`! This would let package authors use _any_ pre/post process they please given that it can already parse [Less], [Sass], and can even be used with [Stylus] (which would have been the more natural choice to pair with CoffeeScript IMO).

so Atom itself wouldn't even _need_ to migrate all the internals for it to work. You could just look at the `main:`/`index.*` file and pass the correct `parser:` flag for PostCSS to compile everything. I think this could be awesome for package authors rather than [kind of] forcing Less on them…

[less]: https://npmjs.com/package/postcss-less
[sass]: https://npmjs.com/package/postcss-scss
[stylus]: https://npmjs.com/package/poststylus

[comment]: https://github.com/atom/atom/pull/12903#issuecomment-281522676

PR #13087, opened 2016-10-27
Update transpilation pipeline

BinaryMuse:
Currently, the Atom transpilation story, via the compile cache, is as follows:
- All TypeScript (`.ts`) files are compiled with the TypeScript compiler.
- All CoffeeScript (`.coffee`) files are compiled with the CoffeeScript compiler.
- JavaScript (`.js`) files are compiled with the Babel compiler _if_ they start with `/** @babel */`, `"use babel"`, `'use babel'`, or `/* @flow */`. Atom currently uses Babel 5 to transpile JS, and sets some explicit options in `static/babelrc.json`.

In addition, the `ThemeManager` uses `LessCompileCache` to compile Less (`.less`) files to CSS.

In an ideal world, individual packages would have more control over how exactly their source is transpiled before being run in Atom. A best case scenario would be that a package author can specify _exactly_ how we should transform their source. At a minimum, packages should be able to pick-and-choose which JavaScript transformations are applied to them by specifying their own options in a local `.babelrc` and including any additional Babel plugins as appropriate.

To start with, I'm updating Babel to version 6 while attempting to keep the resulting JavaScript as similar to the Babel 5 version as possible, to avoid breaking existing packages.
- [x] Update to Babel 6
  - [ ] Include appropriate plugins to mimic Babel 5 behavior
- [ ] Figure out backward-compat opt-out story (e.g. pragma)
- [ ] Figure out local customizations (including ones with package-specific options, like Relay schema location)
- [ ] Other stuff

/cc @atom/core @nathansobo


nathansobo:
I think we should seriously consider an API that allows each package to fully specify transpilation, but continue to manage caching within Atom. This may promote packages consuming more disk space with dependencies, but that will be solved when we globally de-dupe package dependencies which is something I've been wanting to do anyway.

If we allow packages to specify their own handling, at that point we can completely deprecate core support and ignore the pragma stuff until 2.0, at which point we'll rip out baked-in transpilation support entirely.

static/babelrc.json
```diff
@@ -1,7 +1,19 @@
...
+  "plugins": [
+    "transform-class-properties",
+    "transform-async-to-generator",
+    ["transform-runtime", {
+      "polyfill": false,
+      "regenerator": true
+    }],
+    "add-module-exports",
+    ["transform-es2015-modules-commonjs", {
+      "strictMode": false
```
BinaryMuse: Based on the output of Travis, this isn't having any effect.

BinaryMuse:
@nathansobo I quite like that idea, though it does raise the barrier to entry for things like async/await. However, we would be able to leave the current transpilation pipeline alone (let it work as it always has with pragmas present) and then allow opting into some new thing.

nathansobo:
> I quite like that idea, though it does raise the barrier to entry for things like async/await.

`Async/await` will be standard in Chrome 55, which should be available in Electron in not too long.

BinaryMuse:
Okay, so a few questions I guess (sorta thinking out loud here):

* What would such an API look like? How could we make it very flexible but still support caching in core?
* How would we manage deduping given that it's very common to lock-down deps that manage source-level transforms to specific versions?
* Would we only modify the transpilation workflow for JavaScript, or would we also deprecate support for TS, CS, and Less out of the box? (Feels like it'd be okay to leave them, but not certain.)

steelbrain:
This also somewhat fixes atom/apm#498

nathansobo:
Honestly if we fixed our package publishing scenario to allow artifacts rather than just raw git repositories, then people could just transpile stuff up front. That would mean we're only needing to transpile in development mode, which reduces concerns about bloating up the disk or memory quite a bit.

For development mode, it would still be nice to allow people to hook into the require system in a controlled way, however. My thoughts are still evolving as to how, but perhaps something like the following:

A top level `package.json` key called `transpilers`. This would be an array of `{globPattern, transpiler}` objects. The `transpiler` field would contain a require path (a relative path *or* preferably an absolute require that can come from `node_modules`) to a plain JS file that exports a single function.

When we need to transpile in dev mode for a subdirectory of the package matching  the given pattern, we require the file at the given path and call it with the path and source of the file being transpiled. If a string is returned, we use it as the new source. If undefined is returned, we assume no transpilation is desired for that particular file.

I think ultimately we should phase out core support for all transpiled languages. They represent a real liability because they aren't standardized and have the potential to churn. Already we have the question... is a `.ts` file TypeScript 1 or 2? It's not a good long term strategy. Dropping them would mean you have to write your init script in plain JS, but of course you could always require a transpiler and use it to run another init script in your language of choice.

We can't drop anything until Atom 2.0 though.

nathansobo:
I should also add that dropping support for Less is going to be tricky given our current theming situation. Nesting isn't valuable, but the less functions for manipulating colors don't yet have a counterpart in standard CSS. We also bake all sorts of nasty assumptions about Less includes into our theming system. Long term, I want it gone as well though.

arcticicestudio:
> I should also add that dropping support for Less is going to be tricky given our current theming situation. Nesting isn't valuable, but the less functions for manipulating colors don't yet have a counterpart in standard CSS. We also bake all sorts of nasty assumptions about Less includes into our theming system. Long term, I want it gone as well though.
Are there plans to switch to Sass and drop LESSCSS or do you rellay want to use standard CSS?
I think themes should remain in preprocessor language and Sass is the way Atom 2.0 should go :smile:

thomasjo:
> Are there plans to switch to Sass and drop LESSCSS or do you rellay want to use standard CSS?

There is (to my knowledge) no real benefit switching from Less to Sass, so that will likely never happen. Primary problem with Sass is that it doesn't run natively in Node.js. It requires Ruby, which would create a whole new set of problems, dependencies, etc.

**TL;DR** we will _not_ switch to Sass.

BinaryMuse:
Okay, I've fiddled with this a bit the past two days and I think it's just too tricky/risky to update the built-in pipeline to support Babel 6 without breaking people's stuff. Since we want to move away from transpiling in core and allow packages to specify a full transpilation story on their own, I'm going to close this and work on that instead.

Alhadis:
> I think ultimately we should phase out core support for all transpiled languages.

Question: would this target CSON configs too? IMHO, CSON is the most underrated data format in the world: more flexible than YAML, but equally readable ([leaning toothpick syndrome notwithstanding](https://github.com/Alhadis/language-roff/blob/master/grammars/roff.cson#L2147-L2157)).

BinaryMuse:
CSON doesn't operate via the transpilation pipeline, so if we did decide to
deprecate it (I'm not saying we will), it would be a separate effort.

Alhadis:
Good to know! =) Though I'll probably wind up writing a `libcson` library in C one day anyway, just because.

danielbayley:
> Primary problem with Sass is that it doesn't run natively in Node.js. It requires Ruby, which would create a whole new set of problems, dependencies, etc.

Stylus on the other hand _is_ built on Node and is **super** flexible, and seems like a much more natural choice to pair with CoffeeScript…

But anyway forcing a particular processor on users is _less_ than ideal (🙈). It seems like the best solution would either be to…

1. Restructure Atom to run on PostCSS (which [can already parse Less](https://www.npmjs.com/package/postcss-less) and so could be done gradually)… See #5076.

2. > Honestly if we fixed our package publishing scenario to allow artifacts rather than just raw git repositories, then people could just transpile stuff before publishing as they do with npm modules.

Solve this, which is the real problem.

But I think `1.` might be less painful as far as the theme system goes, given that you want to expose standardised `ui` and `syntax` variables to authors, and seem to want avoid addressing `1.` when it comes to transpiling JavaScript…

PR #13095, opened 2016-10-29
Introduce follow through behavior for tooltips

vjeux:
Inside of Nuclide, we have multiple places where we have multiple icons close together that have a tooltip: the left toolbar, the bottom status bar, the debugger icons...

The current behavior is very frustrating as you've got to wait for the delay on every single one of them. But, we have a clear signal that the user wants a tooltip when s/he waits the time to see it and it's likely that moving the mouse over the next item quickly means that s/he wants to see it as well.

This pull request introduces the concept of follow through: if you have seen a tooltip, you're going to instantly see tooltip on the next element you mouse over within a short timer (300ms right now).

Test Plan:
Video before:

Video after:

Released under CC0

### DIFF
--- a/spec/tooltip-manager-spec.coffee
+++ b/spec/tooltip-manager-spec.coffee
@@ -1,4 +1,5 @@
 TooltipManager = require '../src/tooltip-manager'
+Tooltip = require '../src/tooltip'
 _ = require 'underscore-plus'
 
 describe "TooltipManager", ->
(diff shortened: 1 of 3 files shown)


vjeux:
cc @nathansobo

nathansobo:
Awesome idea. Are you cool writing some tests for this?

vjeux:
Absolutely, let me work on tests. I wasn't sure if you'd be interested in it so didn't want to spend the time upfront. Working on it :)

vjeux:
I was checking Sublime to see how it handled this and it seems to implement the same concept:

nathansobo:
Yeah it's going to be great. Sorry for this crazy jQuery derived code. It's pretty bad.

spec/tooltip-manager-spec.coffee
```diff
@@ -149,6 +160,29 @@ describe "TooltipManager", ->
       it "hides the tooltips", ->
         manager.add element, title: "Title"
         hover element, ->
-          expect(document.body.querySelector(".tooltip")).toBeDefined()
+          expect(document.body.querySelector(".tooltip")).not.toBeNull()
```
vjeux: I think it's better to write `.not.toBeNull` because `.toBeDefined` is `.not.toBeUndefined` which is also true if it's `null`.

vjeux:
Added a test that didn't pass before and now passes. Let me know if you want me to write more tests :)

Ran it with
```
atom --test spec/tooltip-manager-spec.coffee
```

nathansobo:
The test looks great. I relocated it to fit a bit better into the existing test structure and rephrased the description a bit to unpack what "follow-through" means to a newcomer to the code. Once this goes green we can :ship:. Nice addition! Thanks so much.

vjeux:
The test failure on appveyor seems spurious

nathansobo:
Thanks again!

PR #13101, opened 2016-10-29
Per-package transpilation

BinaryMuse:
This is a ~~WIP and (thus far) rough~~ implementation of per-package transpilation pipelines mentioned in #13087. It allows packages to opt in to fully managing their own transpilation steps, bypassing the built-in Atom transpilation entirely. Here's roughly how it works:
1. A package declares a property in its `package.json` called `atomTranspilers`. This is an array of objects, with each object specifying `glob`, `transpiler`, and optionally `options`.
2. When a package is loaded, if `atomTranspilers` exists in its metadata, we notify the `CompileCache` that it wants to do its own compilation.
3. A new class called `PackageTranspilationRegistry` wraps all the `CompileCache` compilers with a new object of the same interface that falls back to the original compiler if a package hasn't opted in to its own transpilation, but bypasses it completely otherwise.
4. If a package is doing it's own compilation, then for each file that matches the `glob` specified, the associated `transpiler` is required, and the exported value should be an object with a function called `compile` that takes three arguments: `sourceCode`, `filePath`, and `options` (where `options` comes from the package's `package.json`). It should return a string or Buffer with the transpiled source, or `undefined` to indicate that it doesn't want to transpile that file. Note that returning `undefined` doesn't fall back to the legacy, core behavior — packages which opt in to this mechanism are _fully_ responsible for _all_ their transpilation.

Also worth noting is that the transpiler file cannot be transpiled itself; its source must run in a vanilla Atom environment (which is admittedly pretty capable these days :).

One question I have is how to handle the caching. For example, I currently have `getCachePath` set to

``` javascript
return path.join(
  "package-transpile",
  crypto
    .createHash('sha1')
    .update(transpilerSource, 'utf8')
    .update(sourceCode, 'utf8')
    .digest('hex')
  )
```

but in many cases it's possible that neither the file's source nor the transpiler's source has changed, but the output of transpilation would be different (e.g. when editing a `.babelrc` file). Perhaps we should allow the package author to also specify a cache key for the file?

/cc @nathansobo @atom/core for early 💭

### DIFF
--- a/spec/package-transpilation-registry-spec.js
+++ b/spec/package-transpilation-registry-spec.js
@@ -50,9 +50,11 @@ describe("PackageTranspilationRegistry", () => {
     let hitPathMissSubdir = '/path/to/file4.js'
     let hitPathMissExt = '/path/to/file5.ts'
     let nodeModulesFolder = '/path/to/lib/node_modules/file6.js'
+    let hitNonStandardExt = '/path/to/file7.omgwhatisthis'
 
     let jsSpec = { glob: "lib/**/*.js", transpiler: './transpiler-js', options: { type: 'js' } }
     let coffeeSpec = { glob: "*.coffee", transpiler: './transpiler-coffee', options: { type: 'coffee' } }
+    let omgSpec = { glob: "*.omgwhatisthis", transpiler: './transpiler-omg', options: { type: 'omg' } }
 
     let jsTranspiler = {
       transpile: (sourceCode, filePath, options) => {
@@ -74,18 +76,30 @@ describe("PackageTranspilationRegistry", () => {
       }
     }
 
+    let omgTranspiler = {
+      transpile: (sourceCode, filePath, options) => {
+        return {code: sourceCode + "-transpiler-omg"}
+      },
+
+      getCacheKeyData: (sourceCode, filePath, options) => {
+        return 'omg-transpiler-cache-data'
+      }
+    }
+
     beforeEach(() => {
       jsSpec._transpilerSource = "js-transpiler-source"
       coffeeSpec._transpilerSource = "coffee-transpiler-source"
+      omgTranspiler._transpilerSource = "omg-transpiler-source"
 
       spyOn(registry, "getTranspiler").andCallFake(spec => {
         if (spec.transpiler === './transpiler-js') return jsTranspiler
         if (spec.transpiler === './transpiler-coffee') return coffeeTranspiler
+        if (spec.transpiler === './transpiler-omg') return omgTranspiler
         throw new Error('bad transpiler path ' + spec.transpiler)
       })
 
       registry.addTranspilerConfigForPath('/path/to', [
-        jsSpec, coffeeSpec
+        jsSpec, coffeeSpec, omgSpec
       ])
     })
 
@@ -93,6 +107,7 @@ describe("PackageTranspilationRegistry", () => {
       spyOn(originalCompiler, 'shouldCompile').andReturn(false)
       expect(wrappedCompiler.shouldCompile('source', hitPath)).toBe(true)
       expect(wrappedCompiler.shouldCompile('source', hitPathCoffee)).toBe(true)
+      expect(wrappedCompiler.shouldCompile('source', hitNonStandardExt)).toBe(true)
       expect(wrappedCompiler.shouldCompile('source', hitPathMissExt)).toBe(false)
       expect(wrappedCompiler.shouldCompile('source', hitPathMissSubdir)).toBe(false)
       expect(wrappedCompiler.shouldCompile('source', missPath)).toBe(false)
@@ -115,6 +130,7 @@ describe("PackageTranspilationRegistry", () => {
 
       expect(wrappedCompiler.compile('source', hitPath)).toEqual('source-transpiler-js')
       expect(wrappedCompiler.compile('source', hitPathCoffee)).toEqual('source-transpiler-coffee')
+      expect(wrappedCompiler.compile('source', hitNonStandardExt)).toEqual('source-transpiler-omg')
       expect(wrappedCompiler.compile('source', missPath)).toEqual('source-original-compiler')
       expect(wrappedCompiler.compile('source', hitPathMissExt)).toEqual('source-original-compiler')
       expect(wrappedCompiler.compile('source', hitPathMissSubdir)).toEqual('source-original-compiler')
--- a/spec/package-transpilation-registry-spec.js
+++ b/spec/package-transpilation-registry-spec.js
@@ -77,13 +77,6 @@ describe("PackageTranspilationRegistry", () => {
       jsSpec._transpilerSource = "js-transpiler-source"
       coffeeSpec._transpilerSource = "coffee-transpiler-source"
 
-      const oldFsRealpathSync = fs.realpathSync.bind(fs)
-      spyOn(fs, 'realpathSync').andCallFake(thePath => {
-        if (thePath === '/path/to') return thePath
-        if (thePath === '/path/other') return thePath
-        return oldFsRealpathSync(thePath)
-      })
-
       spyOn(registry, "getTranspiler").andCallFake(spec => {
         if (spec.transpiler === './transpiler-js') return jsTranspiler
         if (spec.transpiler === './transpiler-coffee') return coffeeTranspiler
--- a/spec/package-transpilation-registry-spec.js
+++ b/spec/package-transpilation-registry-spec.js
@@ -55,7 +55,7 @@ describe("PackageTranspilationRegistry", () => {
 
     let jsTranspiler = {
       transpile: (sourceCode, filePath, options) => {
-        return sourceCode + "-transpiler-js"
+        return {code: sourceCode + "-transpiler-js"}
       },
 
       getCacheKeyData: (sourceCode, filePath, options) => {
@@ -65,7 +65,7 @@ describe("PackageTranspilationRegistry", () => {
 
     let coffeeTranspiler = {
       transpile: (sourceCode, filePath, options) => {
-        return sourceCode + "-transpiler-coffee"
+        return {code: sourceCode + "-transpiler-coffee"}
       },
 
       getCacheKeyData: (sourceCode, filePath, options) => {
--- a/spec/package-transpilation-registry-spec.js
+++ b/spec/package-transpilation-registry-spec.js
@@ -98,7 +98,7 @@ describe("PackageTranspilationRegistry", () => {
         throw new Error('bad transpiler path ' + spec.transpiler)
       })
 
-      registry.addTranspilerConfigForPath('/path/to', [
+      registry.addTranspilerConfigForPath('/path/to', 'my-package', [
         jsSpec, coffeeSpec, omgSpec
       ])
     })
--- a/spec/package-transpilation-registry-spec.js
+++ b/spec/package-transpilation-registry-spec.js
@@ -1,0 +1,128 @@
+/** @babel */
+import fs from 'fs'
+import path from 'path'
+
+import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
+
+import PackageTranspilationRegistry from '../src/package-transpilation-registry'
+
+let originalCompiler = {
+  getCachePath: (sourceCode, filePath) => {
+    return "orig-cache-path"
+  },
+
+  compile: (sourceCode, filePath) => {
+    return sourceCode + "-original-compiler"
+  },
+
+  shouldCompile: (sourceCode, filePath) => {
+    return path.extname(filePath) === '.js'
+  }
+}
+
+describe("PackageTranspilationRegistry", () => {
+  let registry
+  let wrappedCompiler
+
+  beforeEach(() => {
+    registry = new PackageTranspilationRegistry()
+    wrappedCompiler = registry.wrapTranspiler(originalCompiler)
+  })
+
+  it('falls through to the original compiler by default', () => {
+    spyOn(originalCompiler, 'getCachePath')
+    spyOn(originalCompiler, 'compile')
+    spyOn(originalCompiler, 'shouldCompile')
+
+    wrappedCompiler.getCachePath('source', '/path/to/file.js')
+    wrappedCompiler.compile('source', '/path/to/filejs')
+    wrappedCompiler.shouldCompile('source', '/path/to/file.js')
+
+    expect(originalCompiler.getCachePath).toHaveBeenCalled()
+    expect(originalCompiler.compile).toHaveBeenCalled()
+    expect(originalCompiler.shouldCompile).toHaveBeenCalled()
+  })
+
+  describe('when a file is contained in a path that has custom transpilation', () => {
+    let hitPath = '/path/to/lib/file.js'
+    let hitPathCoffee = '/path/to/file2.coffee'
+    let missPath = '/path/other/file3.js'
+    let hitPathMissSubdir = '/path/to/file4.js'
+    let hitPathMissExt = '/path/to/file5.ts'
+
+    let jsSpec = { glob: "lib/**/*.js", transpiler: './transpiler-js', options: { type: 'js' } }
+    let coffeeSpec = { glob: "*.coffee", transpiler: './transpiler-coffee', options: { type: 'coffee' } }
+
+    let jsTranspiler = {
+      transpile: (sourceCode, filePath, options) => {
+        return sourceCode + "-transpiler-js"
+      },
+
+      getCacheKeyData: (sourceCode, filePath, options) => {
+        return 'js-transpiler-cache-data'
+      }
+    }
+
+    let coffeeTranspiler = {
+      transpile: (sourceCode, filePath, options) => {
+        return sourceCode + "-transpiler-coffee"
+      },
+
+      getCacheKeyData: (sourceCode, filePath, options) => {
+        return 'coffee-transpiler-cache-data'
+      }
+    }
+
+    beforeEach(() => {
+      jsSpec._transpilerSource = "js-transpiler-source"
+      coffeeSpec._transpilerSource = "coffee-transpiler-source"
+
+      const oldFsRealpathSync = fs.realpathSync.bind(fs)
+      spyOn(fs, 'realpathSync').andCallFake(thePath => {
+        if (thePath === '/path/to') return thePath
+        if (thePath === '/path/other') return thePath
+        return oldFsRealpathSync(thePath)
+      })
+
+      spyOn(registry, "getTranspiler").andCallFake(spec => {
+        if (spec.transpiler === './transpiler-js') return jsTranspiler
+        if (spec.transpiler === './transpiler-coffee') return coffeeTranspiler
+        throw new Error('bad transpiler path ' + spec.transpiler)
+      })
+
+      registry.addTranspilerConfigForPath('/path/to', [
+        jsSpec, coffeeSpec
+      ])
+    })
+
+    it('always returns true from shouldCompile for a file in that dir that match a glob', () => {
+      spyOn(originalCompiler, 'shouldCompile').andReturn(false)
+      expect(wrappedCompiler.shouldCompile('source', hitPath)).toBe(true)
+      expect(wrappedCompiler.shouldCompile('source', hitPathCoffee)).toBe(true)
+      expect(wrappedCompiler.shouldCompile('source', hitPathMissExt)).toBe(false)
+      expect(wrappedCompiler.shouldCompile('source', hitPathMissSubdir)).toBe(false)
+      expect(wrappedCompiler.shouldCompile('source', missPath)).toBe(false)
+    })
+
+    it('calls getCacheKeyData on the transpiler to get additional cache key data', () => {
+      spyOn(registry, "getTranspilerPath").andReturn("./transpiler-js")
+      spyOn(jsTranspiler, 'getCacheKeyData').andCallThrough()
+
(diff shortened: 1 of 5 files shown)


nathansobo:
Excited to see this coming together! I wonder if we can expose something that can be an ingredient to the cache key like a version number that we can hash together with the options. I guess giving them direct control is okay but it would be nice to be able to make our own solid guarantees about the performance of the cached code path.

nathansobo:
A few more thoughts that bubbled into my head...

* Should the transpiler module's method be called `transpile` rather than `compile` for consistency? Or should we go the other way and name the key `compilers`?
* Refining my thoughts on the participation in cache invalidation... I'm not sure a version is enough because transpilers may want to support mechanisms such as `.babelrc`, the contents of which can change. So perhaps it's safest to just offer maximal power with an optional `getCacheKey` method and hope people don't abuse it.

Finally some broader thoughts that may extend beyond the scope of this PR but I wanted to bring them up? Do we need to prefix our manifest fields with `atom`? I'm actually not clear on where else we do this, but considering this `package.json` is actually for an Atom package and not a node module, it seems like it's our namespace to populate. If this were intended to be a vanilla node module I'd feel differently, but since these things are going to run in Atom... If we do decide to prefix, I'd be happier with a single `atom` sub-object containing all the Atom-related stuff.

BinaryMuse:
@nathansobo 

> Should the transpiler module's method be called `transpile` rather than `compile` for consistency? Or should we go the other way and name the key `compilers`?

Good point. "Transpile" is such a weird word, I'd almost just rather use "compile" in user-facing stuff.

> So perhaps it's safest to just offer maximal power with an optional `getCacheKey` method and hope people don't abuse it.

This feels reasonable. People can already do plenty of stuff to fuck up performance, and ideally if anyone abuses it it'll be picked up by timecop and fixed in a PR. 😁 

> Do we need to prefix our manifest fields with `atom`?

I did this partly out of general safety (yes, it's an Atom package, but other Node things that people use stuff shit in `package.json`) and consistency (we already have `atomTestRunner`). I kind of like the idea of a single `atom` sub-object, but it feels like it's too late for that?

BinaryMuse:
Okay, I went with "transpile" because it's the language we use everywhere else. I'm not attached, though.

For the caching, I added an optional method to the transpiler called `getCacheKeyData` that allows the user to return *additional* data that helps determine how a file is cached; for example, in a recent package, I used the contents of all the files that affect transpilation:

```javascript
var cacheKeyData = [
  fs.readFileSync(path.join(__dirname, '.babelrc')),
  fs.readFileSync(path.join(__dirname, 'schema.json')),
  fs.readFileSync(path.join(__dirname, 'relay-babel.js'))
].join('\n')

module.exports = {
  getCacheKeyData: function () {
    return cacheKeyData
  },

  // ...
```

BinaryMuse:
A thought I had as I was working on specs for this: this will only work for `.js`, `.ts`, and `.coffee` files because those are the only three with `require.extensions` hooks. We should probably allow users to specify other extensions they want to load; this would install a "pass-through" compiler on those extensions for files that aren't transpiled by the pipeline. One thought was to add an explicit `extensions` option in the `package.json` bit; it's a little repetitive since many `glob`s will already have an extension listed, but I can't think of a good way to handle it in a generic case.

BinaryMuse:
So, I was wrong — obscure extensions work fine because Node defaults all unknown extensions to the `require.extensions['.js']` compiler, so as long as we keep that one assigned to a `PackageTranspilationRegistry` wrapped compiler, we're good to go.

---

This works perfectly with an internal package I'm working on and doesn't seem to add any significant startup time to the package. I think it's ready for more serious 👀 now. In particular, I'd love feedback on places where lookups and the like are less than optimal, or places where folks experience slowdowns when loading Atom.

Failing tests on appveyor are unrelated.

/cc @atom/feedback

BinaryMuse:
Wondering if there should be some code to specifically ignore `node_modules`; `**/*.js` is likely to be a common glob. Are there any instances where the user would *want* to ensure `node_modules` is included?

nathansobo:
I think we want to always exclude `node_modules`.

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,151 @@
...
+    var lastPath = null
+    // Iterate parents from the file path to the root, checking at each level
+    // to see if a package manages transpilation for that directory.
+    // This means searching for a config for `/path/to/file/here.js` only
+    // only iterates four times, even if there are hundreds of configs registered.
+    while (thisPath !== lastPath) { // until we reach the root
+      if (config = this.configByPackagePath[thisPath]) { // eslint-disable-line no-cond-assign
+        for (var i = 0; i < config.specs.length; i++) {
+          spec = config.specs[i]
+          if (minimatch(filePath, path.join(config.path, spec.glob))) {
```
nathansobo: Maybe you want to compile the minimatch regexp up front? Not sure if they do caching internally.

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,151 @@
...
+      .update(transpilerSource, 'utf8')
+      .update(sourceCode, 'utf8')
+
+    var additionalCacheData
+    if (transpiler && transpiler.getCacheKeyData) {
+      additionalCacheData = transpiler.getCacheKeyData(sourceCode, filePath, spec.options)
+      hash.update(additionalCacheData, 'utf8')
+    }
+
+    return path.join('package-transpile', hash.digest('hex'))
```
nathansobo: I wonder if it's worth grouping transpiled sources in their own subfolder named after the package... that might make it easier to delete them when debugging or something. I don't have a super strong opinion on this one but I thought I'd raise the idea.

nathansobo:
It seems like we only support the extensions `.js`, `.ts` and `.coffee`... am I missing something? If this is true, think you should add probably have some facility for declaring a new extension so this is fully general.

BinaryMuse:
> I think we want to always exclude `node_modules`.

👍 

> Maybe you want to compile the minimatch regexp up front?

Good thought, I'll look into that.

> I wonder if it's worth grouping transpiled sources in their own subfolder named after the package

I considered this but wasn't sure. I feel more confident now that you've mentioned it. I'll try it and see how it feels.

> It seems like we only support the extensions `.js`, `.ts` and `.coffee`... am I missing something?

Yeah, Node [falls back to `.js`](https://github.com/nodejs/node/blob/40366df885ec75c7eeee5e7e7626212ae1a6e770/lib/module.js#L489-L490) if it doesn't recognize the extension.

src/package-transpilation-registry.js
BinaryMuse: Hm, doesn't look like minimatch actually uses a regex under the hood.

BinaryMuse:
Okay, ignoring `node_modules` is in, and I modified the transpilation function to expect the transpiled source code on a `.code` property so we have room to do fancy stuff with source maps and stuff in the future (for now source maps will work if you just add an inline one to the transpiled source code). I also added a test case for file extensions that aren't registered with `require.extensions`. The only problem would be if some package takes over some extension (like `.jsx`) and then another package tries to transpile on that extension.

src/package-transpilation-registry.js
BinaryMuse: This was a good idea, thanks. 👍

BinaryMuse:
I believe this is ready to go.

spec/package-transpilation-registry-spec.js
```diff
@@ -0,0 +1,145 @@
+/** @babel */
+import fs from 'fs'
+import path from 'path'
+
+import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
+
+import PackageTranspilationRegistry from '../src/package-transpilation-registry'
+
+let originalCompiler = {
```
thomasjo: `const`?

spec/package-transpilation-registry-spec.js
```diff
@@ -0,0 +1,145 @@
...
+
+    let omgTranspiler = {
+      transpile: (sourceCode, filePath, options) => {
+        return {code: sourceCode + "-transpiler-omg"}
+      },
+
+      getCacheKeyData: (sourceCode, filePath, options) => {
+        return 'omg-transpiler-cache-data'
+      }
+    }
```
thomasjo: Looks like all declarations above can be `const`-ified.

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
+var crypto = require('crypto')
+var fs = require('fs')
+var path = require('path')
+
+var minimatch = require('minimatch')
+
```
thomasjo: `const`-ify all the above

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+var Resolve = null
```
thomasjo: `let`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+
+  removeTranspilerConfigForPath: function (packagePath) {
+    delete this.configByPackagePath[packagePath]
+  },
+
+  // Wraps the transpiler in an object with the same interface
+  // that falls back to the original transpiler implementation if and
+  // only if a package hasn't registered its desire to transpile its own source.
+  wrapTranspiler: function (transpiler) {
+    var self = this
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    return {
+      getCachePath: function (sourceCode, filePath) {
+        var spec = self.getPackageTranspilerSpecForFilePath(filePath)
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+        if (spec) {
+          return self.getCachePath(sourceCode, filePath, spec)
+        }
+
+        return transpiler.getCachePath(sourceCode, filePath)
+      },
+
+      compile: function (sourceCode, filePath) {
+        var spec = self.getPackageTranspilerSpecForFilePath(filePath)
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        return transpiler.compile(sourceCode, filePath)
+      },
+
+      shouldCompile: function (sourceCode, filePath) {
+        if (self.transpilerPaths[filePath]) {
+          return false
+        }
+        var spec = self.getPackageTranspilerSpecForFilePath(filePath)
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+
+    // ignore node_modules
+    if (filePath.indexOf(path.sep + 'node_modules' + path.sep) > -1) {
+      return false
+    }
+
+    var config = null
+    var spec = null
+    var thisPath = filePath
+    var lastPath = null
```
thomasjo: `let`-ify all these

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    // Iterate parents from the file path to the root, checking at each level
+    // to see if a package manages transpilation for that directory.
+    // This means searching for a config for `/path/to/file/here.js` only
+    // only iterates four times, even if there are hundreds of configs registered.
+    while (thisPath !== lastPath) { // until we reach the root
+      if (config = this.configByPackagePath[thisPath]) { // eslint-disable-line no-cond-assign
+        for (var i = 0; i < config.specs.length; i++) {
```
thomasjo: `let`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    var transpilerSource = spec._transpilerSource || fs.readFileSync(transpilerPath, 'utf8')
+    spec._transpilerSource = transpilerSource
+    var transpiler = this.getTranspiler(spec)
+
+    var hash = crypto
+      .createHash('sha1')
+      .update(JSON.stringify(spec.options || {}))
+      .update(transpilerSource, 'utf8')
+      .update(sourceCode, 'utf8')
+
```
thomasjo: Use `const` and `let` as appropriate above.

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    var additionalCacheData
```
thomasjo: `let`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    return path.join('package-transpile', spec._config.name, hash.digest('hex'))
+  },
+
+  transpileWithPackageTranspiler: function (sourceCode, filePath, spec) {
+    var transpiler = this.getTranspiler(spec)
+
+    if (transpiler) {
+      var result = transpiler.transpile(sourceCode, filePath, spec.options || {})
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    if (transpiler && transpiler.getCacheKeyData) {
+      additionalCacheData = transpiler.getCacheKeyData(sourceCode, filePath, spec.options)
+      hash.update(additionalCacheData, 'utf8')
+    }
+
+    return path.join('package-transpile', spec._config.name, hash.digest('hex'))
+  },
+
+  transpileWithPackageTranspiler: function (sourceCode, filePath, spec) {
+    var transpiler = this.getTranspiler(spec)
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+      var result = transpiler.transpile(sourceCode, filePath, spec.options || {})
+      if (result === undefined || (result && result.code === undefined)) {
+        return sourceCode
+      } else if (result.code) {
+        return result.code.toString()
+      } else {
+        throw new Error('Could not find a property `.code` on the transpilation results of ' + filePath)
+      }
+    } else {
+      var err = new Error("Could not resolve transpiler '" + spec.transpiler + "' from '" + spec._config.path + "'")
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+  getTranspilerPath: function (spec) {
+    Resolve = Resolve || require('resolve')
+    return Resolve.sync(spec.transpiler, {
+      basedir: spec._config.path,
+      extensions: Object.keys(require.extensions)
+    })
+  },
+
+  getTranspiler: function (spec) {
+    var transpilerPath = this.getTranspilerPath(spec)
```
thomasjo: `const`

src/package-transpilation-registry.js
```diff
@@ -0,0 +1,159 @@
...
+    if (transpilerPath) {
+      var transpiler = require(transpilerPath)
```
thomasjo: `const`

thomasjo (dismissed):
Nitpicking related to lack of `const` and `let` usage.

BinaryMuse:
@thomasjo Following the style here: https://github.com/atom/atom/blob/daf790406f42fee77355fed56ece280ac6e60c10/src/compile-cache.js#L3-L6

I'll change the ones in the spec file.

thomasjo:
@BinaryMuse That code is almost a year old though. We have native support for most of ES6/7 now. I really think we should try to ensure new code is as modern as possible. I'll happily modernize `compile-cache.js` if necessary.

nathansobo:
@BinaryMuse Shoot one thing I forgot. Could you document this in the Flight Manual or some other appropriate spot?

BinaryMuse:
@nathansobo Already on it :)

maxbrunsfeld:
👏

zertosh:
I totally spaced on this PR. The thing that worries me about this, is that it encourages package authors to ship (1) untranspiled code, (2) put babel & friends in `dependencies` (and bloating packages)

BinaryMuse:
@zertosh After some discussion, we decided to go with this solution. We don't have a facility to allow package authors to ship transpiled code easily; they'd have to check it into their repository, making diffs and pull requests more difficult. Additionally, we still wanted to have a mechanism for at-runtime transpilaton when in dev mode. The package bloat is a concern, but given the circumstances (and the fact that we want to do global deduping of installed modules), we were willing to eat the cost.

zertosh:
> they'd have to check it into their repository, making diffs and pull requests more difficult.

Yes, but it actually works out if you follow a different publish flow from just running `apm publish` from `master`. We publish Nuclide pre-transpiled. We first transpile everything, commit the code (but we don't merge it into `master`), we tag it (`git tag vX.X.X`), and then publish the tag `apm publish --tag vX.X.X`). So master never has transpiled code - so no conflicts or anything.

> The package bloat is a concern

Bloat is one. Another one is load time. If package authors are not super careful about lazy loading their transpiler, you're going to end up paying a big perf penalty. Babel is not a small dependency, it can take upwards of 50ms to just `require` it.

nathansobo:
I think this should probably only work in development mode and we should fix the inability to publish compiled package data.

BinaryMuse:
> If package authors are not super careful about lazy loading their transpiler, you're going to end up paying a big perf penalty.

Package authors will not be required to require their transpiler as part of their package code path. It's a separate module that the Atom internals require if and only if the given files are not already cached (so on first load of a package and after any package update).

> what if we only enabled this in dev mode and fix the root problem?

I like the idea in theory, though in practice it would require different code/steps for running in dev and compiling for publishing. That said, I'd be open to it if we can make it work smoothly.

nathansobo:
Yeah, I suppose you're right about the caching meaning we'll avoid requiring the transpiler in most cases. After digesting that point again, I'm open to the chance of a one-time slowdown to get stuff transpiled on install. Ultimately, I do think it makes the most sense to publish transpiled assets, but that's obviously uncontroversial it's just a lot of work.

zertosh:
> Package authors will not be required to require their transpiler as part of their package code path. It's a separate module that the Atom internals require if and only if the given files are not already cached 

Hmm, maybe I'm misreading the code, but it looks like you still require `atomTranspilers.transpiler` to check for a `getCacheKeyData` function. So if you put your requires at the top of the `atomTranspilers.transpiler` file (as most people do), you're going to pay the load cost for `babel` or whatever every time - unless the package author inline's the `require` call.

Package authors would have to write something like this to avoid the penalty:

```js
let version;
let babel;

module.exports = {
  transpile(sourceCode, filePath, options) {
    if (babel == null) babel = require('babel-core');
    return babel.transpile(/*....*/);
  },

  getCacheKeyData(sourceCode, filePath, options) {
    // or whatever you want to base the cache key on.
    if (version == null) version = require('babel-core/package.json').version;
    return version;
  }  
}
```

Side note: We use https://github.com/zertosh/babel-plugin-transform-inline-imports-commonjs in Nuclide and in yarn to inline imports to defer `require` calls until the imported identifier is referenced.

nathansobo:
I'm imagining package authors relying on libraries to do the transpiler support. So something like `atom-babel` would implement the transpiler object and you'd just reference that in your `node_modules`. @BinaryMuse that makes me realize i didn't check whether the transpiler path is passed to `require.resolve` like we do with the test runner.

zertosh:
Is `atom-babel` a hypothetical package? I couldn't find it lol.. Maybe it's my trauma talking but the majority of transpile setups I've seen are pretty bad. In any case, I left my 2c.

nathansobo:
Yeah, your 2c is super valuable and we appreciate it. I think at this point we're going to risk moving forward because we need this on a package we're working on and don't have time to setup the publishing flow. Long term you're absolutely right that we should just transpile on publish.

BinaryMuse:
> Hmm, maybe I'm misreading the code, but it looks like you still require `atomTranspilers.transpiler` to check for a `getCacheKeyData` function.

Ah, yes, you're right of course. I forgot about that detail. I like giving user full control over generating (a portion of) the cache key, but I wonder if it'd be better to just allow the user to specify a list of files that go into generation of the key so we can just pull stuff from `package.json` without needing to require the module?

Or, as @nathansobo said, point people toward pre-made modules (I'm working on one for Babel 6 now) that implement transpilation for the user, and highlight the edge cases in the docs for people that want to make their own.

BinaryMuse:
> that makes me realize i didn't check whether the transpiler path is passed to `require.resolve` like we do with the test runner.

Yes, `getTranspilerPath` [runs the path through `Resolve.sync`](https://github.com/atom/atom/pull/13101/files#diff-42678e5bce81ab920184a4a8415043bcR143).

Arcanemagus:
Has this been added to the Flight Manual yet? A search for `atomTranspilers` shows no results at the moment.

BinaryMuse:
@Arcanemagus It hasn't; I've been ironing out the details and wrinkles as I've used the feature, and will be adding it once it's ready.

danielbayley:
> only support the extensions .js, .ts and .coffee

So am I right in thinking…
1. This can't be utilised to take advantage of PostCSS in a say theme? (See also #5076 for a more general idea on that…)
2. This is effectively a temporary patch to address the root problem which is basically that `apm` doesn't work like `npm` when it comes to publishing…

BinaryMuse:
@danielbayley 

> This can't be utilised to take advantage of PostCSS in say, a theme?

This is all about the `CompileCache` in Atom, which is the mechanism which transpiles files via the Node.js `require` mechanism. So it would not be useful for CSS unless you ran it through Node's `require` somehow.

> This is effectively a temporary patch to address the root problem

It's not the solution we think is most ideal, but there certainly isn't any guarantee of or timeline for an alternative solution.

PR #13065, opened 2016-10-25
Convert BufferedProcess and BufferedNodeProcess to JS

mnquintana:
We're gonna get there, bit by bit. 😄 

👀 @atom/feedback

PR #13113, opened 2016-10-31
Convert a bunch of small files to JS

mnquintana:
Continuing on from https://github.com/atom/atom/pull/13065, this PR rewrites a bunch of small files as JS. 🤘 

👀 @atom/feedback


src/model.js
```diff
@@ -0,0 +1,37 @@
...
+    nextInstanceId = 1
+  }
+
+  constructor (params = {}) {
+    this.alive = true
+    this.assignId(params.id)
+  }
+
+  assignId (id) {
+    this.id = this.id || id || nextInstanceId++
```
mnquintana: I don't think it will since we initialize it to `1`, but that's a good point that this could be more robust

mnquintana:
Huh, looks like `atomdoc` is blowing up on some unknown docstring 😞  – I'll take a look into adding more debug info to the build script to make it easier to find which file has invalid Atomdoc going on 🕵️

joh1234:

mnquintana:
Sat on this one a bit too long, and I'm preeeetty sure at least some of these files have seen some edits since I opened this PR. Gonna do a new one with fewer files.

Issue #10079, opened 2015-12-15
Do not handle package activation exceptions in test mode.


thedaniel:
I've run into difficulty writing tests a few times when my package failed to activate but the test environment didn't have anything interesting in the console or test output indicating that's where it failed. In `package.coffee` we should just re-raise if we're in test mode rather than calling `::handleError`. I plan to do this today, just writing up this issue to remind myself when I switch contexts (or in case I never get to it!)

maxbrunsfeld:
:+1: This will probably help a lot of people.

kepta:
Hi, I am a new here. Can I work on this bug?

maxbrunsfeld:
@kepta Absolutely! Thanks for taking an interest. Do you have the information you need?

kepta:
@maxbrunsfeld I went through the docs and the source code specifically package.coffee.
In this file the ```@devMode``` is what seems relevant to me or is it? I am not able to get the complete idea, any pointers?

b-lambert:
Is this still an open issue? If it's beginner-friendly I'd be happy to take it on.

50Wliu:
@b-lambert yes it is!

SahilC:
Hi, I've made a PR #13119 for this bug. Could someone review it?

PR #13119, opened 2016-10-31
Raising exceptions in test mode (Issue: #10079) 

SahilC:


src/atom-environment.coffee
```diff
@@ -171,6 +171,7 @@ class AtomEnvironment extends Model
       devMode, @configDirPath, resourcePath, safeMode, @config, styleManager: @styles,
       commandRegistry: @commands, keymapManager: @keymaps, notificationManager: @notifications,
       grammarRegistry: @grammars, deserializerManager: @deserializers, viewRegistry: @views
+      testMode: @inTestMode()
```
50Wliu: To be consistent with the rest I would change this to be `@testMode` instead.

SahilC:
I've made the requested changes, could someone verify if everything is ok?

thomasjo:
Can you explain the purpose of this PR? How is this different to [spec mode](https://github.com/atom/atom/blob/8229924b96497f0d0d0db47355d3e821c58a77ec/src/atom-environment.coffee#L427-L429)?

thomasjo:
OK, I see from the title that this is meant to fix #10079. Next time, please add all relevant details to the description body of the PR. I'll add some review comments for the remainder of my review.

src/atom-environment.coffee
```diff
@@ -171,6 +171,7 @@ class AtomEnvironment extends Model
       devMode, @configDirPath, resourcePath, safeMode, @config, styleManager: @styles,
       commandRegistry: @commands, keymapManager: @keymaps, notificationManager: @notifications,
       grammarRegistry: @grammars, deserializerManager: @deserializers, viewRegistry: @views
+      testMode: @testMode
```
thomasjo: This needs to be removed. We already have this in the form of `specMode`.

src/atom-environment.coffee
```diff
@@ -420,6 +421,10 @@ class AtomEnvironment extends Model
   inDevMode: ->
     @devMode ?= @getLoadSettings().devMode
 
+  # Public: Returns a {Boolean} that is `true` if the current window is in test mode.
+  inTestMode: ->
+    @testMode ?= @getLoadSettings().testMode
+
```
thomasjo: Same as earlier, needs to be removed.

src/package-manager.coffee
```diff
@@ -30,7 +30,7 @@ module.exports =
 class PackageManager
   constructor: (params) ->
     {
-      configDirPath, @devMode, safeMode, @resourcePath, @config, @styleManager,
+      configDirPath, @devMode, safeMode, @testMode, @resourcePath, @config, @styleManager,
```
thomasjo: `@testMode` needs to be removed.

src/package-manager.coffee
```diff
@@ -397,7 +397,7 @@ class PackageManager
 
       options = {
         path: packagePath, metadata, packageManager: this, @config, @styleManager,
-        @commandRegistry, @keymapManager, @devMode, @notificationManager,
+        @commandRegistry, @keymapManager, @devMode, @testMode, @notificationManager,
```
thomasjo: Bye bye, `@testMode`

src/package.coffee
```diff
@@ -32,7 +32,7 @@ class Package
   constructor: (params) ->
     {
       @path, @metadata, @packageManager, @config, @styleManager, @commandRegistry,
-      @keymapManager, @devMode, @notificationManager, @grammarRegistry, @themeManager,
+      @keymapManager, @devMode, @testMode, @notificationManager, @grammarRegistry, @themeManager,
```
thomasjo: You know the drill by now 😄

src/package.coffee
```diff
@@ -124,7 +124,10 @@ class Package
             else
               @activateNow()
           catch error
-            @handleError("Failed to activate the #{@name} package", error)
+            if @testMode
```
thomasjo: Change `@testMode` to `@specMode`.

thomasjo (changes requested):

50Wliu:
Ah hah, I thought there was a specMode already defined somewhere...but I didn't see anywhere in the diff :P.

SahilC:
Yeah, I was confused about that too! Should I just rename everything to specMode?

thomasjo:
@SahilC It already exists, so remove almost all additions (as per my review comments). Keep the change package.coffee, but rename the `@testMode` there to `@specMode`.

SahilC:
But specMode isn't being passed to package.coffee's constructor. Am I missing something?

thomasjo:
@SahilC Sorry I should have looked at the diffs more closely, but you can use [`atom.inSpecMode()`](https://atom.io/docs/api/v1.11.2/AtomEnvironment#instance-inSpecMode).

src/package.coffee
thomasjo: Sorry I was mistaken earlier, but `@testMode` can be replaced with `atom.inSpecMode()`.

SahilC:
I've removed all occurrences of testMode, and replaced the check with atom.inSpecMode(). Wouldn't it be more consistent to use a variable rather than actually calling that function directly?

thomasjo:
I don't think it's worth adding the necessary changes for that given that we only check if we're in spec mode on a single line. I'd argue we should rather remove `@devMode` when we get around to converting this to JS.

This looks good to me now!

thomasjo (approved):

SahilC:
Cool, thanks!

src/package.coffee
```diff
@@ -124,7 +124,10 @@ class Package
             else
               @activateNow()
           catch error
-            @handleError("Failed to activate the #{@name} package", error)
+            if atom.inSpecMode()
+              raise error
```
maxbrunsfeld: This should be `throw error`.

maxbrunsfeld:
Thanks for looking into this problem @SahilC! Just one comment 👆 .

joh1234:

Issue #8314, opened 2015-08-11
Update to Octicons 3.0


simurai:
The Octicons got [updated to 3.0](https://github.com/github/octicons/pull/58). There are some visual changes, but also some icons got removed and some new ones added.

It's not that straightforward to update them, since some of the community packages might use some of the icons that got removed.. like the `alignment`, `move`, `jump` icons.

A few options:
1. Update to Octicons 3.0. Packages might need to make changes or add their own icons.
2. "Manually" update to Octicons 3.0 and still keep the ones that got removed. There should be no name conflicts.
3. Keep the current Octicons as is.
4. Keep the current Octicons as is (at least until Atom 2.0). And add an additional "Atom icons set". Could be a fork of the Octicons 3.0 and modified to Atom's needs. Class names could be `atom-icon` or so. The find + replace already uses some custom icons. Also having custom icons for autocomplete-plus would be nice.

simurai:
Short term we can just keep as is (3). And then at some point, try 2. And maybe start an additional Atom set for all icons that need to be custom.

thomasjo:
I (or someone else) can grab all community packages and `grep` for usage of the removed octicons. I'm super busy today though due to travel, so I might not have time to do this today.

If there's no usage of the removed icons, I propose we upgrade sooner rather than later.

simurai:
I see. Yeah, if we have some usage insights, that would make it easier. In case, here a list of the class names that got removed:

```
icon-alignment-align
icon-alignment-aligned-to
icon-alignment-unalign
icon-beer
icon-hourglass
icon-jump-down
icon-jump-left
icon-jump-right
icon-jump-up
icon-microscope
icon-move-down
icon-move-left
icon-move-right
icon-move-up
icon-playback-fast-forward
icon-playback-pause
icon-playback-play
icon-playback-rewind
icon-podium
icon-puzzle
icon-split
icon-steps
```

benogle:
> Keep the current Octicons as is (at least until Atom 2.0)

I think we should probably do this. There's no reason to switch them up: no benefit, and opportunity for pain.

mnquintana:
This might be a good time to start an Atom 2.0 milestone or label for breaking changes we won't touch till then.

simurai:
> no benefit, and opportunity for pain.

Well, arguably the new icons look better? :grimacing: But yeah, maybe not enough to cover the potential pain.

thedaniel:
[...]

PR #13138, opened 2016-11-03
Update Octicons

simurai:
This updates the Octicons to `4.4.0` (almost newest). Some of the old icons from `2.1.2` that got removed in the meantime will still be available in Atom, so no packages should have missing icons.

Before | After
--- | ---
screen shot 2016-11-03 at 7 54 01 pm | screen shot 2016-11-03 at 7 54 22 pm

Here the 20 new icons that will be available:

screen shot 2016-11-03 at 7 22 20 pm

## Concerns

All icons should be fine except that the `GitHub` wordmark is much smaller. We can update the font size in places where we use that icon, like on the About page. The "Gist" icon is also quite small, but since it's a new one, it hasn't been used anywhere.

## Remaining tasks:

- [x] Add new icons to the **Style Guide** https://github.com/atom/styleguide/pull/45
- [x] Update **Flight Manual** https://github.com/atom/flight-manual.atom.io/pull/285
- [x] Update the GitHub icon on the **About** page https://github.com/atom/about/pull/43

Closes https://github.com/atom/atom/issues/8314
Will be in Atom `1.13`


simurai:
Btw. this will probably be the last update because the newer Octicons 5.0 [only support SVG](https://github.com/primer/octicons/pull/117).

And we should also move to SVG icons.. @joefitzgerald already got an etch version here: https://github.com/joefitzgerald/etch-octicon

Alhadis:
Needs more padlocks:

kek

Don't think we can ship with only four.

Issue #10517, opened 2016-01-20
Atom launch story is complicated and confusing


lee-dohm:
How to launch Atom in specific states is a bit of magic that is only known to a few. Some features are more obvious like Dev Mode and Safe Mode, but even then the full effects of each of these modes are not fully documented anywhere and therefore they are often misunderstood even by people who use these modes often. The purpose of this Issue initially is to capture the list of initial states that people want to launch Atom into. Eventually we will create a plan and design for making each of them easy, well understood and as obvious as possible to newcomers to Atom. If possible, I'll add links to other Issues, PRs, Discuss topics and other sources to illustrate.
## Atom Launch Modes

These are the modes that Atom more-or-less supports or are implied by the modes that Atom does support.
- Project (whether with single or multiple roots)
- No project &mdash; Open specific files with no tree view
- Restore last state (project or no project)
- Restore specific project state
- Dev Mode &mdash; Since this is designed as an aid for Atom development, it mostly only makes sense in combination with the project mode (but perhaps I'm overlooking something)
- Safe Mode &mdash; Can be combined with any of the above modes (other than Dev Mode)

Each of these should have an obvious and unambiguous way of launching them both from the command line and from the GUI.
### Additional Command Line Thoughts

Sometimes people want to modify the state of Atom from the command line and sometimes they don't. Some consideration should be given to these scenarios:
- Add a file to an open session of Atom
- Add a project root to an open session of Atom
- Launch Atom in any mode when an open session of Atom is running (even when it may seem the user may want one of the two above but doesn't ... for instance "open a file that exists in the project that is open in the current session of Atom, but I want it in a new Atom session with **no** project")
## Mode State

There are things that are serialized and stored when Atom exits. It is completely non-obvious what items belong to this set and what items do not as well as how to restore the saved items when next launching Atom. Decisions should be made about whether some of these items belong to the specific mode that Atom is in or if they are more global state that people expect to be in place the next time Atom is launched (or even an individual file is opened post-launch) no matter what mode Atom is launched in.
- List of open tabs
- Editor state &mdash; Where the cursor is in the file, etc
- Unsaved changes to files
- Untitled buffers
- Per file indentation settings
- Per file line ending settings
- Per file encoding settings
- [Pane configuration](https://github.com/atom/atom/issues/10545)

jeancroy:
Regarding `Unsaved changes to files` & `Untitled buffers` across the different projects state, I'd like to propose a start notification along the line of 

[...]

PR #13152, opened 2016-11-06
Add and improve core.*OnStart setting descriptions

### LINKED ISSUE #10623: "Open Empty Editor On Startup" option not working.

lee-dohm:
### Description of the Change

The current description of the `core.openEmptyEditorOnStart` is misleading which has caused a lot of confusion around what this option is supposed to do or not do. Many people change this setting expecting some behavior difference when they start Atom the next time only to find out that it doesn't do anything from what they can tell. This is because Atom's default behavior in the absence of other instructions is to restore the state from the last time Atom was open, completely ignoring this setting.

There is [a way to configure Atom to not restore the last state](https://github.com/atom/atom/pull/11324). The configuration setting was never exposed in the config schema though so people had to dig through the code or search through Issues and PRs to discover that it was an option. This change adds the `core.restorePreviousWindowsOnStart` setting to the schema so that is no longer necessary.

### Alternate Designs

A more [comprehensive overhaul of Atom's launch story](https://github.com/atom/atom/issues/10517) whether in one large change or multiple small ones is something that is being looked at but won't realistically solve this confusion in the near term.

This isn't new functionality.

### Benefits

A more clear understanding of what these settings do and do not.

### Possible Drawbacks

If not adding the `core.restorePreviousWindowsOnStart` setting to the config schema was intentional, this has the obvious drawback that it is now showing up in the Settings View.

### Applicable Issues

* Applies to #10517
* Fixes #10623

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -122,7 +122,12 @@ const configSchema = {
         ]
       },
       openEmptyEditorOnStart: {
-        description: 'When checked opens an untitled editor on _File > New Window_; otherwise no buffer is opened.',
+        description: 'When checked opens an untitled editor when loading a blank environment (such as with _File > New Window_ or when "Restore Previous Windows On Start" is unchecked); otherwise no buffer is opened when loading a blank environment. This setting has no effect when restoring a previous state.',
+        type: 'boolean',
+        default: true
+      },
+      restorePreviousWindowsOnStart: {
+        description: 'When checked restores the last state of all Atom windows when started from the icon or `atom` by itself from the command line; otherwise a blank environment is loaded.',
         type: 'boolean',
         default: true
       },
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -122,7 +122,7 @@ const configSchema = {
         ]
       },
       openEmptyEditorOnStart: {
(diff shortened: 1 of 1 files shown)


Ben3eeE:
From the [blog post](http://blog.atom.io/2016/06/06/atom-1-8-and-1-9-beta.html)
>This option will be on by default, but you can easily turn it off by changing your config.cson file (please note that this setting won’t appear in the Settings UI).

👍

lee-dohm:
Thanks for finding that @Ben3eeE :grinning:

as-cii:
This looks good to me, @lee-dohm. I recall people were against the overhead of an extra checkbox in the settings-view, but honestly I agree with you it's just easier to make this setting explicit and, as a result, more discoverable.

🚢 at will! 👍

Issue #13084, opened 2016-10-27
Shelling out to collect environment variables is creating runaway 100% CPU processes on Linux


fooishbar:
### Prerequisites
- [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
- [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?
### Description

I'm fairly sure this will end up being an Electron issue, but here goes ...

When I start atom-beta (1.12.0b4, though was present with b3 at least), it spawns a shell ('env') subprocess which burns 100% CPU, attempting to read stdin forever:

```
systemd,1,1 --system --deserialize 126
  ├─atom,15392,10213
  │   ├─atom,15403,10213
  │   │   └─atom,15466,10213
  │   │       ├─zsh,15497,10213 -ilc command env
```

strace on the shell shows this looping forever:

```
rt_sigprocmask(SIG_SETMASK, [], [TSTP TTIN TTOU], 8) = 0
read(0, NULL, 0)                        = 0
rt_sigprocmask(SIG_BLOCK, [TSTP TTIN TTOU], [], 8) = 0
getpgrp()                               = 10213
ioctl(15, TIOCGPGRP, [11445])           = 0
getpgrp()                               = 10213
ioctl(15, TIOCGPGRP, [11445])           = 0
rt_sigprocmask(SIG_SETMASK, [], [TSTP TTIN TTOU], 8) = 0
read(0, NULL, 0)                        = 0
rt_sigprocmask(SIG_BLOCK, [TSTP TTIN TTOU], [], 8) = 0
```

[...]

PR #13186, opened 2016-11-10
Timeout when invoking the shell, log information to the console

joefitzgerald:
- Partially addresses #13084

### DIFF
--- a/src/update-process-env.js
+++ b/src/update-process-env.js
@@ -63,7 +63,18 @@ function getEnvFromShell (env) {
     return
   }
 
-  let {stdout} = spawnSync(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8'})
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for taking care of this @joefitzgerald :grinning:

src/update-process-env.js
```diff
@@ -63,7 +63,18 @@ function getEnvFromShell (env) {
     return
   }
 
-  let {stdout} = spawnSync(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8'})
+  let {stdout, error, status, signal} = spawnSync(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', timeout: 5000})
```
as-cii: I think we should decrease the timeout, as blocking for 5 seconds will still be perceived as unresponsive. Thoughts? 💭

PR #13200, opened 2016-11-11
Make `updateProcessEnv` async and add `core:loaded-shell-environment` activation hook

### LINKED ISSUE #9406: Slow start up time launching from Dock or Mission Control
I noticed that when I launch Atom from my applications that the loading time is significantly longer than if I use the command line.  I'm running a 2014 MacBook Pro.  

I've attached a screen shot of Time Cop showing several packages that have added milliseconds to the load time.  Is there any way to speed up Atom without having to disable packages as I use many of the packages on a regular basis?    

screenshot 2015-11-02 20 06 00

### LINKED ISSUE #13084: Shelling out to collect environment variables is creating runaway 100% CPU processes on Linux

as-cii:
Fixes #9406.
Fixes #13084.

When Atom is launched from the Dock, we invoke the user shell to retrieve environment variables that are assigned by their profile settings. However, we were doing so synchronously during startup, thus causing a [major slowdown](https://github.com/atom/atom/issues/9406) when opening Atom from the Dock and sometimes [hanging the startup process indefinitely](https://github.com/atom/atom/issues/13084).

With this pull request, we are changing `updateProcessEnv` into an asynchronous method. Packages that require shell environment variables to be assigned prior to activation can use the new `core:loaded-shell-environment` activation hook by putting the following snippet in `package.json`:

```json
{"activationHooks": ["core:loaded-shell-environment"]}
```

This will defer packages' activation until `process.env` has been updated with the user environment variables.

/cc: @atom/maintainers

### DIFF
--- a/src/update-process-env.js
+++ b/src/update-process-env.js
@@ -1,7 +1,7 @@
 /** @babel */
 
 import fs from 'fs'
-import {spawnSync} from 'child_process'
+import childProcess from 'child_process'
 
 const ENVIRONMENT_VARIABLES_TO_PRESERVE = new Set([
   'NODE_ENV',
@@ -15,10 +15,10 @@ const PLATFORMS_KNOWN_TO_WORK = new Set([
   'linux'
 ])
 
-function updateProcessEnv (launchEnv) {
+async function updateProcessEnv (launchEnv) {
   let envToAssign
   if (launchEnv && shouldGetEnvFromShell(launchEnv)) {
-    envToAssign = getEnvFromShell(launchEnv)
+    envToAssign = await getEnvFromShell(launchEnv)
   } else if (launchEnv && launchEnv.PWD) {
     envToAssign = launchEnv
   }
@@ -58,21 +58,23 @@ function shouldGetEnvFromShell (env) {
   return true
 }
 
-function getEnvFromShell (env) {
+async function getEnvFromShell (env) {
   if (!shouldGetEnvFromShell(env)) {
-    return
+    return null
   }
 
-  let {stdout, error, status, signal} = spawnSync(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', timeout: 5000})
+  let {stdout, error} = await new Promise((resolve) => {
+    childProcess.execFile(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', timeout: 5000}, (error, stdout) => {
+      resolve({stdout, error})
+    })
+  })
+
   if (error) {
     if (error.handle) {
       error.handle()
     }
+    console.log('warning: ' + env.SHELL + '-ilc "command env" failed with signal (' + error.signal + ')')
     console.log(error)
-  }
-
-  if (status !== 0) {
-    console.log('warning: ' + env.SHELL + '-ilc "command env" failed with status (' + status + ') and signal (' + signal + ')')
   }
 
   if (stdout) {
--- a/spec/update-process-env-spec.js
+++ b/spec/update-process-env-spec.js
@@ -1,6 +1,7 @@
 /** @babel */
 /* eslint-env jasmine */
 
+import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
 import path from 'path'
 import temp from 'temp'
 import child_process from 'child_process'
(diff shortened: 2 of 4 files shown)


src/update-process-env.js
```diff
@@ -58,23 +58,25 @@ function shouldGetEnvFromShell (env) {
...
+    childProcess.execFile(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', timeout: 5000}, (error, stdout) => {
+      resolve({stdout, error})
+    })
+  })
+
   if (error) {
     if (error.handle) {
       error.handle()
     }
+    console.log('warning: ' + env.SHELL + '-ilc "command env" failed with signal (' + error.signal + ')')
```
joefitzgerald: It is possible that the exit code is non-zero but there is no error.

joefitzgerald: Ahh, got it. Makes sense...

ivankravets:
> Packages that require shell environment variables to be assigned prior to activation can use the new

@as-cii, Could you call this hook when package has been installed via package manager? I've tried to use this hook for @PlatformIO IDE and new release broke installation process. Users need to manually restart Atom and only after it package will be activated.

Thanks!

joefitzgerald:
@ivankravets you point out a key bug in the current implementation: if a user disables a package dependent on this hook for activation, or if a user installs a new package dependent on this hook, or if a user updates a package dependent on this hook, it will fail to activate.

ivankravets:
I made hot fix for PlatformIO IDE that adds this hook ONLY when Atom is >=1.12.2 and ony after install/upgrade procedure. Default IDE's manifest doesn't contain activation hooks. Nevertheless, IDE will add them after first installation.

See https://github.com/platformio/platformio-atom-ide/blob/develop/lib/maintenance.js#L78:L94

Issue #13104, opened 2016-10-30
Network mapped paths resolve differently in Atom 1.12


Ben3eeE:
### Prerequisites
- [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
- [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [X] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [X] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.
### Description

Since Atom 1.12 beta the paths of files on network mapped drives are resolved with the full server name instead of the path of the network mapped drive. This caused problems with restoring the session of projects on network mapped drives when upgrading to Atom 1.12.

Example: `\\server\path\to\folder` is mapped to `Y:\`. Atom restores the session with the `Y:\` path but all new files open with the `\\server\path\to\folder` path. So you have to close and reopen all files you previously had opened to not have duplicate files open. This also can cause the UNC path error when running the linter on beta.

I did discover that the `tree-view` still has the old path if you copy the full path of some file and the title bar of Atom also displays the network mapped drive label. So currently we display both of the paths in different parts of Atom.
##### Atom 1.12.0-beta5 paths

![path stuff](https://cloud.githubusercontent.com/assets/1058982/19836364/6d5ec3dc-9e9e-11e6-955e-bc29966a3d65.gif)
##### Atom 1.11.2 paths

![path stuff 1 11](https://cloud.githubusercontent.com/assets/1058982/19836376/21a003a6-9e9f-11e6-8e65-6519716dfa3c.gif)
##### Upgrading from 1.11 to 1.12

[...]

PR #13245, opened 2016-11-16
Upgrade Electron from 1.3.6 to 1.3.7

damieng:
Brings in Node 6.5.0 which solves #13104


damieng:
Closing in favor of https://github.com/atom/atom/pull/13247

Issue #2991, opened 2014-07-16
Rightclick in the treeview doesn't focus the selected item directly


philipgiuliani:
I was not sure if this belongs to treeview, or to the atom context menu, so i decided to open it directly here.

If i press rightclick on a treeview, the item behind my mouse won't be selected. If i click it again, it gets the selected item. So im not sure which one is selected currently and i always have to do left-click and then a right click.

![rightclick in treeview](https://cloud.githubusercontent.com/assets/5446019/3599270/eae7ef68-0cec-11e4-8138-02e78b9d8367.gif)

Atom Version: 0.115.0

alex-martin-xactium:
Interested too. Very useful.

thomaslindstrom:
This is related: #3502

hannes-landeholm:
I have the same problem in Atom 1.0.5 and Mac OS X 10.10.4. It's very annoying to not see what you are right clicking.

sstur:
As @kevinsawicki has eluded to, this is an underlying problem with Electron/Chromium on Mac OS that the context menu blocks the UI thread while shown, so any UI updates that are deferred (requestAnimationFrame?) will not show until after the context menu has been dismissed. This will probably not be an easy fix but I'm guessing we need to either hack together a solution so that context menu does NOT block the UI thread (difficult) or we will need to implement our own context menu (ugly). Should we file a more generic task for fixing the context menu?

zcbenz:
Related: https://github.com/atom/electron/issues/1854.

mnquintana:
FWIW, I imagine that there is some workaround possible for this, if for no other reason than that VS Code managed to do it with their tree view.

ssorallen:
A delay of just 10ms works for me, the same thing @kevinsawicki did above. 10ms still feels instant to me.

```cs
@subscribe $(document), 'contextmenu', (e) ->
  e.preventDefault()
  setTimeout(() ->
    atom.contextMenu.showForEvent(e)
  , 10)
```

Zireael07:
Can @kevinsawicki's solution be incorporated into core?

sstur:
The only issue with picking a timeout value is it seems so arbitrary. I mean it works for one person on one set of hardware but is it robust enough to count on? No matter what it's better than the current, broken behavior but a more elegant solution should be possible. I wonder if requestAnimationFrame in combination with setImmediate would be better.

    @subscribe $(document), 'contextmenu', (e) ->
      e.preventDefault()
[...]

PR #13266, opened 2016-11-18
Flush DOM before displaying context menu

vjeux:
Electron has an issue where displaying the context menu pauses the rendering loop. https://github.com/electron/electron/issues/1854

One unfortunate side effect is that when you right click on an element, the dom mutations to highlight the element have not been flushed yet which makes for a very bad user experience https://github.com/atom/atom/issues/2991

In order to solve this issue, we would like to flush the mutations and only then display the menu. Unfortunately, electron doesn't have APIs for that.

Things I have tried:
1) Electron has `browserWindow.webContents.invalidate()` and `.on('paint' -> ...)` but it only works with offscreen rendering which     atom doesn't use.
2) `.capturePage(-> ...)` takes a screenshot of the page and I hoped it     would do a flush first but it inconsistently displays the highlight     so it won't work.
3) `Using setTimeout((-> ...), 16)` works many times but not always
4) `requestAnimationFrame(-> ...)` should be what we want but unfortunately     it doesn't always show the highlights :(

What I found worked was to nest three `requestAnimationFrame`, it works all the time even when my machine is busy. This is not a 100% solution but I think that it's okay to assume that if the highlight hasn't been updated in three frames then you should probably try and optimize your code. 3 frames is 50ms and that delay isn't noticeable.

Note: the proper way to solve this that was mentioned in the issue is to open the menu from the main process but this is already what atom does and it's still an issue.

Before:

After:

Released under CC0

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -3,6 +3,7 @@ path = require 'path'
 fs = require 'fs'
 url = require 'url'
 {EventEmitter} = require 'events'
+ContextMenu = require './context-menu'
 
 module.exports =
 class AtomWindow
@@ -100,12 +101,7 @@ class AtomWindow
 
   hasProjectPath: -> @getLoadSettings().initialPaths?.length > 0
 
-  setupContextMenu: ->
-    @browserWindow.on 'context-menu', (menuTemplate) =>
-      @openContextMenu(menuTemplate)
-
   openContextMenu: (menuTemplate) ->
-    ContextMenu = require './context-menu'
     new ContextMenu(menuTemplate, this)
 
   containsPaths: (paths) ->
@@ -167,8 +163,6 @@ class AtomWindow
     @browserWindow.webContents.on 'will-navigate', (event, url) =>
       unless url is @browserWindow.webContents.getURL()
         event.preventDefault()
-
-    @setupContextMenu()
 
     if @isSpec
       # Spec window's web view should always have focus
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -101,10 +101,12 @@ class AtomWindow
   hasProjectPath: -> @getLoadSettings().initialPaths?.length > 0
 
   setupContextMenu: ->
-    ContextMenu = require './context-menu'
-
     @browserWindow.on 'context-menu', (menuTemplate) =>
-      new ContextMenu(menuTemplate, this)
+      @openContextMenu(menuTemplate)
+
+  openContextMenu: (menuTemplate) ->
+    ContextMenu = require './context-menu'
+    new ContextMenu(menuTemplate, this)
 
   containsPaths: (paths) ->
     for pathToCheck in paths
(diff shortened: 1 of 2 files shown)


as-cii:
Thanks for the pull request, @vjeux! 💯 

> Note: the proper way to solve this that was mentioned in the issue is to open the menu from the main process but this is already what atom does and it's still an issue.

Can you confirm this is what Atom is doing? In [this line of code](https://github.com/atom/atom/pull/13266/files#diff-e27d4c73fad81aa0e3335d45b857153bR205) it seems like we are calling emit on the `BrowserWindow` instance: this works because `BrowserWindow` is an `EventEmitter` but it looks like that call to `emit` might actually be synchronous. Can we try sending an asynchronous message to the main process instead and see if it fixes it? What am I missing?

vjeux:
Do you know what is the function to send an asynchronous message? (First time I'm discovering that part of the codebase so I have no idea what I am doing :p)

I tried to copy and paste the solution given in the issue to make it multi-process but it didn't seem to work, but it's very likely that I did it wrong.

as-cii:
> Do you know what is the function to send an asynchronous message? (First time I'm discovering that part of the codebase so I have no idea what I am doing :p)

I think you should be able to use the `call` method in https://github.com/atom/atom/tree/master/src/ipc-helpers.js to invoke an instance method defined in https://github.com/atom/atom/tree/master/src/main-process/atom-window.coffee. You can see examples of this in https://github.com/atom/atom/tree/master/src/application-delegate.coffee where, for instance, we use `ipcHelpers.call('window-method', 'minimize')` to asynchronously call `AtomWindow.prototype.minimize`. 👍

vjeux:
O M G, this works!

So good and way less hacky <3

src/main-process/atom-window.coffee
```diff
@@ -101,10 +101,12 @@ class AtomWindow
...
 
   setupContextMenu: ->
-    ContextMenu = require './context-menu'
-
     @browserWindow.on 'context-menu', (menuTemplate) =>
-      new ContextMenu(menuTemplate, this)
+      @openContextMenu(menuTemplate)
+
+  openContextMenu: (menuTemplate) ->
+    ContextMenu = require './context-menu'
```
as-cii: What do you think about declaring a `ContextMenu` variable at the top of the file, and change this line so that we require `./context-menu` only once? Like:

```coffee
# at the top...
ContextMenu = null

# this line…
ContextMenu ?= require('./context-menu')
```

Not a huge deal, but should save some cycles when opening context menus.

as-cii (changes requested):
One minor comment and we can 🚢 this. Thanks again, @vjeux!

as-cii (approved):

as-cii:
I went ahead and performed the changes, thanks again @vjeux!

as-cii:
After testing this out on master, it doesn't seem like it works on macOS. Can you double check it works as expected on your machine?

vjeux:
/facepalm

I checked in a the triple nested request animation frame on Nuclide (which is working) and forgot about it. So the Nuclide fix was applied and I thought this pull request was working, when indeed it isn't. Sorry about that :(

https://github.com/facebook/nuclide/blob/master/pkg/nuclide-file-tree/lib/main.js#L75-L95

PR #13272, opened 2016-11-19
Fix prepending multiple gutters at once

vjeux:
There's a bug when multiple gutters are prepended at once where their order is not correctly preserved in the dom. The issue is that we do not update indexInOldGutters even though we inserted a dom node in the old gutters dom.

Honestly, I'm unconvinced that this entire logic is correct, but this fixes the use case we have in Nuclide so it's more correct than before :)

Released under CC0

### DIFF
--- a/spec/gutter-container-component-spec.coffee
+++ b/spec/gutter-container-component-spec.coffee
@@ -139,3 +139,22 @@ describe "GutterContainerComponent", ->
       expect(expectedCustomGutterNode1).toBe atom.views.getView(customGutter1)
       expectedCustomGutterNode3 = gutterContainerComponent.getDomNode().children.item(2)
       expect(expectedCustomGutterNode3).toBe atom.views.getView(customGutter3)
+
+    it "reorders correctly when prepending multiple gutters at once", ->
+      lineNumberGutter = new Gutter(mockGutterContainer, {name: 'line-number'})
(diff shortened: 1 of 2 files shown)


ungb:
Hey @vjeux,

Is there a way I can repro the issue in Nuclide?  I wanted to test the fix if possible.  

I'm not sure why AppVeyor is failing.  looking at the logs, I didn't find anything.

50Wliu:
> I'm not sure why AppVeyor is failing. looking at the logs, I didn't find anything.

It timed out for whatever reason.  I restarted the build.

vjeux:
If you open a file that has both diagnostics and breakpoint gutter, you can see that the diagnostics is left of the breakpoint even though diagnostic has priority -1000 ( https://github.com/facebook/nuclide/blob/master/pkg/nuclide-diagnostics-ui/lib/gutter.js#L79 ) and breakpoint -1100 ( https://github.com/facebook/nuclide/blob/master/pkg/nuclide-debugger/lib/BreakpointDisplayController.js#L63 )

screen shot 2016-12-06 at 8 13 34 am

ungb:
Verified your change fixes the issue.  

image


Will merge change today unless any one has any feedback on the change @atom/core.

vjeux:
Thanks!

iolsen:
🎳 thanks, @vjeux!

PR #13283, opened 2016-11-21
Workaround for #13084

GabMus:


thomasjo:
This completely prevents Atom launched in a non-shell fashion from shelling out to grab missing environment variables. Since this is not a proper fix, but rather a band aid solution for what seems to be a deeper, underlying issue in Arch, I'm closing this PR.

Thanks for contributing!

/cc @atom/feedback

GabMus:
Thank you for your quick reply, but I think you have to reconsider this being an Arch related problem. As many users reported in #13084 this problem also happens in other Linux distros, including Ubuntu, and it's probably an issue within Atom itself.

thomasjo:
Up until now, we have only been able to reliable reproduce, and verify that this affects users who have installed Atom via unofficial distributions; using AUR packages in Arch, and webupd8 PPAs in the case of Ubuntu.

Long story short, the AUR packages and the webupd8 PPA are not supported and currently apply unofficial patches to the Atom source code. Since we do not support these, we will not dedicate any resources to any directly related issues. In the case of Arch we _only_ support installing from _pristine_, unpatched official source code downloaded from atom.io (or GitHub releases).

PR #13287, opened 2016-11-21
Update documentation for GitRepository::getConfigValue

akonwi:
Adds the `key` argument to the function documentation


mnquintana:
Nice catch – thanks for fixing this! 🙌

Issue #13254, opened 2016-11-17
Remove atom-space-pen-views and custom elements from bundled packages


as-cii:
Refs: #13253 

In order to snapshot bundled packages we need to remove eval time interactions with DOM APIs. The two biggest offenders in this regard are jQuery (which is required by `atom-space-pen-views`) and HTML custom elements. The issues with jQuery *could* be addressed by patching jQuery to not do feature detection, but considering that we want to remove it anyway over the long term and it causes other performance and complexity problems, it's probably cleaner to just remove jQuery completely. Custom elements were a failed experiment and we aren't actually getting value out of registering them with the DOM global. They can be replaced with plain JS objects managing a DOM element.

### Affected packages

* [x] Atom Core: Reopen Project (https://github.com/atom/atom/pull/13375)
* [x] archive-view (https://github.com/atom/archive-view/pull/45)
* [x] autocomplete-plus (https://github.com/atom/autocomplete-plus/pull/826)
* [x] bookmarks (https://github.com/atom/bookmarks/pull/76)
* [x] command-palette (https://github.com/atom/command-palette/pull/75)
* [x] deprecation-cop (https://github.com/atom/deprecation-cop/pull/72)
* [x] encoding-selector (https://github.com/atom/encoding-selector/pull/38)
* [x] find-and-replace @maxbrunsfeld (https://github.com/atom/find-and-replace/pull/838)
* [x] fuzzy-finder (https://github.com/atom/fuzzy-finder/pull/273)
* [x] git-diff (https://github.com/atom/git-diff/pull/119)
* [x] go-to-line (https://github.com/atom/go-to-line/pull/40)
* [x] grammar-selector (https://github.com/atom/grammar-selector/pull/35)
* [x] image-view (https://github.com/atom/image-view/pull/82)
* [x] keybinding-resolver (https://github.com/atom/keybinding-resolver/pull/46)
* [x] line-ending-selector (https://github.com/atom/line-ending-selector/pull/37)
* [x] markdown-preview (https://github.com/atom/markdown-preview/pull/456)
* [x] notifications (https://github.com/atom/notifications/pull/139, https://github.com/atom/notifications/pull/149)
* [x] package-generator (https://github.com/atom/package-generator/pull/48)
* [x] settings-view (https://github.com/atom/settings-view/pull/913)
* [x] snippets (https://github.com/atom/snippets/pull/233)
* [x] spell-check (https://github.com/atom/spell-check/pull/193)
[...]

PR #13300, opened 2016-11-22
Include relatedTarget when dispatching blur events from editor elements

as-cii:
Refs: #13254 

~~In atom-space-pen-views, on require, we call `atom.themes.requireStylesheet` to include a default style sheet for the select list. Now that we are planning to replace all usages of space pen with a more modern alternative (https://github.com/atom/atom-select-list), I believe a better design can be achieved by including a default style sheet in Atom core instead of relying on a require-time call to the `ThemeManager`.~~

This pull request introduces a `relatedTarget` option in the synthesized blur event of `TextEditorElement` to make it easier for users of this API to know the value of `document.activeElement` after the event dispatch. We will use this in atom-select-list to [determine if the user has clicked outside the select list](https://github.com/atom/atom-select-list/blob/fa20fdc7d0763a02aa2f1e2385f8f61e0063eb65/src/select-list-view.js#L35-L41) and, if so, trigger a `didCancelSelection` event.

/cc: @atom/maintainers

### DIFF
--- a/src/text-editor-element.coffee
+++ b/src/text-editor-element.coffee
@@ -141,7 +141,7 @@ class TextEditorElement extends HTMLElement
 
   inputNodeBlurred: (event) ->
     if event.relatedTarget isnt this
-      @dispatchEvent(new FocusEvent('blur', bubbles: false))
+      @dispatchEvent(new FocusEvent('blur', relatedTarget: event.relatedTarget, bubbles: false))
 
   addGrammarScopeAttribute: ->
     @dataset.grammar = @model.getGrammar()?.scopeName?.replace(/\./g, ' ')


simurai:
> In atom-space-pen-views, on require, we call atom.themes.requireStylesheet to include a default style sheet for the select list.

@as-cii Do you know the reason for including the select list styles when using space pen?

The select list styles [are part](https://github.com/atom/atom-ui/blob/master/styles/select-list.less) of atom-ui that gets bundled with Atom. So they should already be available anyways.

as-cii:
> Do you know the reason for including the select list styles when using space pen?

Not really, I imagine it has been there for a long time. It got added in https://github.com/atom/atom-space-pen-views/commit/c9151481219b43a3a680e01330476dd72f55d99b but the commits doesn't specify the reason.

Since we bundle that style sheet anyways, I am going to remove https://github.com/atom/atom/pull/13300/commits/0ba42b26b300cd199b5c1f21450eea066476fc00 from this pull request so that we can just merge https://github.com/atom/atom/pull/13300/commits/14092f1641ea9fe04eecd9279f3687352b0ca5e6 into master.

Thanks @simurai!

Issue #12655, opened 2016-09-10
Opening a recent file causes an "item selected is unavailable" error


pflannery:
### Description

Trying to open a recent file from the taskbar on windows causes an error.
### Steps to Reproduce
1. Right click on atom icon on the taskbar
2. Click on a recent file to open

![image](https://cloud.githubusercontent.com/assets/1727302/18408474/4c10ef28-7763-11e6-8acc-a89fc69074fc.png)

**Expected behavior:**

Atom should open the file in to the editor

**Actual behavior:** 

Error is shown

![image](https://cloud.githubusercontent.com/assets/1727302/18408473/438519ec-7763-11e6-885e-e4ae6e742801.png)
### Versions

Windows 10 64 bit

Atom    : 1.10.2
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0

apm  1.12.5
npm  3.10.5
node 4.4.5
python 2.7.10
git 2.9.0.windows.1
visual studio 2015

damieng:
You can get this error if you have subsequently deleted or otherwise moved/relocated the file.   If you hover over it the tooltip should show you the full path it is looking for.

pflannery:
@damieng Nope, the file exists and the tooltip points to correct location

damieng:
I suspect this is caused by the fact atom.exe changes path between upgrades.  

In the new version (on master) we rewrite the jumplist on launch to address most of this although pinned ones that are would have no longer been on the main list would still get missed.

The new version of Squirrel is going to stop moving the .exe between upgrades that will solve this problem.

PR #13307, opened 2016-11-23
Windows taskbar recent & tasks support

damieng:
The task bar supports a number of cool features in Windows we aren't currently using today.

This PR adds;

- Recently files (opened from Explorer, must be associated, can be pinned)
- Recent projects (same as File > Reopen Project including configurable number of items)
- New Window task (same as File> New Window)

A screenshot is worth n number of words.

image


Ben3eeE:
Does this apply to https://github.com/atom/atom/issues/12655?

damieng:
This was the underlying PR that enabled that functionality.... Will continue discussion on other issue.

PR #13320, opened 2016-11-23
Run shell as detached process

### LINKED ISSUE #13317: In some scenarios, opening Atom on Linux results in 100% CPU usage of a core
* When some Linux users open Atom via .desktop files or via the gnome launcher, they experience 100% CPU usage due to the way we are invoking the shell to patch the user's environment
* This does not happen for all Linux users
* This does seem to happen consistently for users on Arch Linux
* The shell is invoked with `detached: false` and using the `-i` flag for an interactive shell
  * The intent of shelling out to get the environment is to run the same shell initialization scripts that are run when a user opens a terminal window and invokes atom from there
  * Removing the `-i` flag is not the solution, because it doesn't result in the desired behavior; it does result in the shell exiting as expected without consuming 100% of a CPU core
  * Adding `detached: true` to the options used for `child_process.spawn` _does_ fix the issue and retains the desired behavior, but it requires additional code to manage the process lifecycle if Atom exits or an Atom window is reloaded while the shell process is still running

The below is a copy of https://github.com/atom/atom/issues/13084#issuecomment-262507036, included here to open up a new issue:

@phw says:

~~~
I debugged this a bit and can reproduce it with the following test script (extracted from [src/update-process-env.js](https://github.com/atom/atom/blob/1b2d2f029945c58cf4cd474ef1423856b35cf26c/src/update-process-env.js)):

```javascript
const childProcess = require('child_process')
const cmd = 'command env'
const shell = process.env.SHELL  // zsh or bash doesn't matter, behavior is the same
// const shell = '/bin/bash'

const child = childProcess.spawn(shell, ['-ilc', cmd], {
   encoding: 'utf8',
  //  detached: true,
  stdio: ['ignore', 'pipe', process.stderr]
})
child.on('error', (e) => {
  console.log('error', e)
})
child.stdout.on('data', (data) => {
  console.log(data.toString('utf8'))
})
child.on('close', (code, signal) => {
  console.log('close', code, signal)
})
```

Launching this directly from a shell works, even if I set an empty environment like so:

    env -i SHELL=/bin/bash node spawn.js

Launching the same from a desktop file in Gnome 3.22 leads to the shell task endlessly consuming 100% of CPU. I used the following desktop file:

```
[Desktop Entry]
Name=Spawn Node Test
Exec=env SHELL=/bin/bash node /home/phw/devel/spawn.js
Icon=atom
Type=Application
```

Two things help to solve this issue:

a) Enable `detached: true` when spawning the child process
b) Remove the `-i` flag from the shell call

I think b) would actually be correct anyway, since this is in no way an interactive shell.

EDIT: Just to be clear: I still don't know exactly *why*  this is happening, I just know the issue can be circumvented removing the `-i` flag.

joefitzgerald:
- Implement timeout
- Clean up running process if window is reloaded or Atom exits

Fixes #13317

### DIFF
--- a/src/update-process-env.js
+++ b/src/update-process-env.js
@@ -62,17 +62,32 @@ function shouldGetEnvFromShell (env) {
 
 async function getEnvFromShell (env) {
   let {stdout, error} = await new Promise((resolve) => {
+    let child
     let error
     let stdout = ''
-    const child = childProcess.spawn(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', stdio: ['ignore', 'pipe', process.stderr]})
+    let done = false
+    const cleanup = () => {
+      if (!done && child) {
+        child.kill()
+        done = true
+      }
+    }
+    process.once('exit', cleanup)
+    setTimeout(() => {
+      cleanup()
+    }, 5000)
+    child = childProcess.spawn(env.SHELL, ['-ilc', 'command env'], {encoding: 'utf8', detached: true, stdio: ['ignore', 'pipe', process.stderr]})
     const buffers = []
     child.on('error', (e) => {
+      done = true
       error = e
     })
     child.stdout.on('data', (data) => {
       buffers.push(data)
     })
     child.on('close', (code, signal) => {
+      done = true
+      process.removeListener('exit', cleanup)
       if (buffers.length) {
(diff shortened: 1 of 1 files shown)


lee-dohm:
Looks good to me! 👍

PR #13353, opened 2016-11-28
Upgrade to Electron 1.3.13

kevinsawicki:
Upgrade to Electron 1.3.13 which includes some important bug fixes and an upstream Chrome patch that (hopefully) addresses the Linux keybinding issues on certain layouts when pressing `Ctrl`.

/cc @as-cii @Ben3eeE the version of Electron with the keybinding fix (1.3.13) is now on this branch if you want to try it out 🐧

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "1.3.12",
+  "electronVersion": "1.3.13",
   "dependencies": {
     "async": "0.2.6",
     "atom-keymap": "7.1.13",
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Awesome thanks. @as-cii I can check this out tonight or tomorrow. Let me know if you want to go ahead and merge before I have a chance to test this PR.

as-cii:
@Ben3eeE: I'll be optimistic and 🚢 this, but it would be awesome if you can test it out when you get the chance. ❤️ 

A huge shoutout goes to @kevinsawicki for addressing this so quickly! Thank you so much! 💯

Ben3eeE:
Gave this a quick round of testing and it seems that the issue has been fixed. We need to change atom to use the event as it still resolves incorrectly. I will test again after those changes are in to better verify without having to resort to reading the events in the dev tools console.

Thanks so much for this @kevinsawicki awesome work 💯

PR #13358, opened 2016-11-30
Add initialize() hook to packages

matthewwithanm:
This adds a new hook, `initialize()` to packages. This is meant to be called before any interaction with the package. Like `activate()`, it's passed the serialized state from the previous session. Unlike `activate()`, it's called _before_ deserialization, enabling item deserialization that's dependent on serialized package state.

It might be a little weird that a function called "initialize" can be called multiple times…open to bikeshedding the name!

### DIFF
--- a/spec/package-spec.coffee
+++ b/spec/package-spec.coffee
@@ -205,3 +205,26 @@ describe "Package", ->
 
     it "uses the package name defined in package.json", ->
       expect(metadata.name).toBe 'package-with-a-totally-different-name'
+
+  describe "the initialize() hook", ->
+    it "gets called when the package is activated", ->
+      packagePath = atom.project.getDirectories()[0].resolve('packages/package-with-deserializers')
+      pack = buildPackage(packagePath)
+      pack.requireMainModule()
+      mainModule = pack.mainModule
+      spyOn(mainModule, 'initialize')
+      expect(mainModule.initialize).not.toHaveBeenCalled()
+      pack.activate()
+      expect(mainModule.initialize).toHaveBeenCalled()
+      expect(mainModule.initialize.callCount).toBe(1)
+
+    it "gets called when a deserializer is used", ->
+      packagePath = atom.project.getDirectories()[0].resolve('packages/package-with-deserializers')
+      pack = buildPackage(packagePath)
+      pack.requireMainModule()
+      mainModule = pack.mainModule
+      spyOn(mainModule, 'initialize')
+      pack.load()
+      expect(mainModule.initialize).not.toHaveBeenCalled()
+      atom.deserializers.deserialize({deserializer: 'Deserializer1', a: 'b'})
+      expect(mainModule.initialize).toHaveBeenCalled()
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -24,6 +24,7 @@ class Package
   mainModulePath: null
   resolvedMainModulePath: false
   mainModule: null
+  mainInitialized: false
   mainActivated: false
 
   ###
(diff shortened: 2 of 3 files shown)


nathansobo:
I think we should maintain some state so that we call `initialize` once and only once.

matthewwithanm:
Do you mean checking `mainInitialized` at each callsite instead of inside `initialize()`? I was thinking that doing it in that one location would make it less likely to be forgotten.

spec/package-spec.coffee
```diff
@@ -205,3 +205,22 @@ describe "Package", ->
...
+      expect(pack.initialize).toHaveBeenCalled()
+
+    it "gets called when a deserializer is used", ->
+      packagePath = atom.project.getDirectories()[0].resolve('packages/package-with-deserializers')
+      pack = buildPackage(packagePath)
+      spyOn(pack, 'initialize')
+      pack.load()
+      expect(pack.initialize).not.toHaveBeenCalled()
+      atom.deserializers.deserialize({deserializer: 'Deserializer1', a: 'b'})
+      expect(pack.initialize).toHaveBeenCalled()
```
maxbrunsfeld: Could you add an assertion that `initialize` only gets called once?

src/package.coffee
```diff
@@ -114,8 +115,24 @@ class Package
     @menus = []
     @grammars = []
     @settings = []
+    @mainInitialized = false
     @mainActivated = false
 
+  initialize: ->
```
maxbrunsfeld: I think we should rename this method to `initializeIfNeeded`, just to make it clear that we only call `initialize` once.

spec/package-spec.coffee
```diff
@@ -205,3 +205,22 @@ describe "Package", ->
 
     it "uses the package name defined in package.json", ->
       expect(metadata.name).toBe 'package-with-a-totally-different-name'
+
+  describe "the initialize() hook", ->
+
```
maxbrunsfeld: Could you remove this blank line?

maxbrunsfeld:
Great, simple solution. I left a few minor comments. Thanks @matthewwithanm!

spec/package-spec.coffee
```diff
@@ -205,3 +205,28 @@ describe "Package", ->
 
     it "uses the package name defined in package.json", ->
       expect(metadata.name).toBe 'package-with-a-totally-different-name'
+
+  describe "the initialize() hook", ->
+    it "gets called when the package is activated", ->
+      packagePath = atom.project.getDirectories()[0].resolve('packages/package-with-deserializers')
+      pack = buildPackage(packagePath)
+      spyOn(pack, 'initializeIfNeeded').andCallThrough()
```
maxbrunsfeld: Now that you're spying on the main module's `initialize` method, I think you should remove this spy on `initializeIfNeeded`, because it's an implementation detail.

matthewwithanm: I thought maybe it would be good to make sure that it wasn't called when loading the main module…but I could remove it or break it out into another test too.

nathansobo:
@damieng I'm assuming the AppVeyor x64 failure is unrelated?

damieng:
Yeah sorry there are 3 tests that are flakey on x64 right now. I've been trying to repro and get to the bottom of them here but only managed to reproduce 1 of them locally and even that is rare.  Looks like possibly race condition based.  I'll stop them running on Windows for now.

nathansobo:
@damieng Personally I'm cool with you leaving them on if you're in the process of addressing it and turning them off would be a hinderance.

nathansobo:
@matthewwithanm Thanks for the clean work here and documenting it too.

matthewwithanm:
Thanks for the help 😊

abe33:
Hi there, just noticed this change this morning as it started breaking one of my package: https://github.com/abe33/minimap-pigments/commit/f78ed88763d3c4ff13cdad872d5f93709bf58428

The thing is:  if a package already has a `initialize` function whose purpose is different it can lead to errors (which was my case). I'm not sure what would be the best solution in that case, maybe I'm the only one that had a `initialize` function in the main module, but maybe it would be worth making sure it won't break too many packages.

nathansobo:
Yeah this was an oversight. We're doing a search now to look at the potential damage and may need to iterate on this if it's too many packages. I fear it could be a lot.

nathansobo:
Here's a list of packages defining an `initialize` method on their main module, sorted descending by download count:

```
editor-background ├── 7463 downloads
project-ring ├── 6502 downloads
tree-view-finder ├── 3513 downloads
notebook ├── 2337 downloads
tree-view-search-bar ├── 1263 downloads
atom-math ├── 1224 downloads
scroll-searcher ├── 846 downloads
atom-trello ├── 837 downloads
atom-bugs ├── 795 downloads
punchclock ├── 670 downloads
wordpress-suite ├── 566 downloads
atom-gist ├── 460 downloads
timekeeper ├── 303 downloads
autocode ├── 178 downloads
btom-mode ├── 52 downloads
repository-file-actions ├── 30 downloads
```

We're going to open PRs on anything with over 1000 downloads and open issues on the rest.

PR #13365, opened 2016-12-01
Specs: test reliability & auto temp clean-up

damieng:
This started-off as trying to figure out why a few tests fail under Windows and indeed should fix a couple of them.

In doing so I decided to make the tests clean up after themselves so that;

- I could see what was going on
- Any resource leaks/locks would show up (there is at least 1 in libgit2/git-utils)
- I didn't have to keep deleting %TEMP% (not automatic on Windows)
- I could now run the specs against a small RAM disk instead (down from 4 minutes to 2)

### DIFF
--- a/spec/compile-cache-spec.coffee
+++ b/spec/compile-cache-spec.coffee
@@ -21,8 +21,8 @@ describe 'CompileCache', ->
     spyOn(TypeScriptSimple::, 'compile').andReturn 'the-typescript-code'
 
   afterEach ->
-    CSON.setCacheDir(CompileCache.getCacheDirectory())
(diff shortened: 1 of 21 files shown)


spec/compile-cache-spec.coffee
```diff
@@ -81,6 +82,7 @@ describe 'CompileCache', ->
 
       error = new Error("Oops")
       expect(error.stack).toBe 'a-stack-trace'
+      console.log('stack ' + error.getRawStack())
```
kevinsawicki: :fire: ?

damieng: Oops, nice catch. This test was randomly failing without useful info so was logging it. Will remove.

Issue #6082, opened 2015-03-24
Warn before quitting


eddiemonge:
It'd be nice to have an option to warn before quitting when hitting Cmd+Q as its too easy to close all your editors when trying to close a single tab. Chrome has this option and its a time saver.

mnquintana:
:+1:

Personally, I'd love if we could adopt Chrome's approach too – pressing <kbd>Cmd+Q</kbd> pops up a message telling you to hold <kbd>Cmd+Q</kbd> to quit, and holding down <kbd>Cmd+Q</kbd> for some amount of time actually quits. (ie. instead of a dialog box asking you to confirm your quit, which I've always found too obtrusive)

stevus:
+1 all day

ismay:
There is some interesting discussion on this here: https://discuss.atom.io/t/quit-confirmation/5572. One of the suggestions was to just make sure work can't be lost (by saving unsaved changes somehow), and to require the user to press `[command] + [q]` twice in order to quit. I personally agree with both these recommendations, and would love to see this implemented (as I tend to fat-finger closing a tab and accidentally close atom all the time).

chrishough:
+!

ThinkGraphical:
Another option is adding the following to your keymap.cson (Atom -> Open your keymap)
```
'body':
    'cmd-q': 'core:save'
    'cmd-q cmd-q': 'application:quit'
```

stevus:
@ThinkGraphical So does this basically require two `Cmd-Q` keypresses?

ThinkGraphical:
@stevus Yes correct. This overwrites the basic behaviour for 'cmd-q' to save the current file. Because an action is required (empty string doesn't work) I chose this because it seemed like a safe option.
And it now requires 2 'cmd-q' keypresses to quit Atom.

chrishough:
This works perfectly!

mjswensen:
+1ing this while I wait for Atom to start up again

Definitely going to try out @ThinkGraphical 's solution (thank you!!!)

adrianoresende:
My atom quit acidental but not is save >:(
All need this feature.

jesperronn:
The workaround from @ThinkGraphical  is nice. Much better than default behaviour. https://github.com/atom/atom/issues/6082#issuecomment-196381781


[...]

PR #13385, opened 2016-12-04
Modified keymap 'ctrl-q' to save files before quiting

pedroduartecosta:
This discussion (https://discuss.atom.io/t/quit-confirmation/5572) and the issue #6082 discusses how useful would be having a warning message before quitting because, sometimes when multiple buffers and multiple windows are opened, user wants to close current buffer in current window by pressing Ctrl+W. Unfortunately, they inadvertently pressed Ctrl+Q and everything disappears. If it's a completely new file on a new window, all the written code is lost.
Atom's community manager opposed this warning message, he said:
>I philosophically disagree with quit confirmations because I believe that quit confirmations are solving a real problem with red tape. The real problem is "I don't want to lose data accidentally".

So I decide a different approach, changing the keymap 'ctrl-q', to first save all documents and only after it would quit the program.

Hope you find this useful and that I'm using the proper channel.

### DIFF
--- a/keymaps/darwin.cson
+++ b/keymaps/darwin.cson
@@ -1,6 +1,6 @@
 'body':
   # Apple specific
-  'cmd-q': 'application:quit'
+  'cmd-q': 'application:safeQuit'
   'cmd-h': 'application:hide'
   'cmd-alt-h': 'application:hide-other-applications'
   'cmd-m': 'application:minimize'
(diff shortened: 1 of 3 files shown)


sadick254:
Hey @pedroduartecosta. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Atom now has a dialog prompt that notifies the user of unsaved changes on a new file on a new window. Now the user has the choice to keep or discard the changes made. We don't need to make any changes to have safe quit. Thank you for your contribution.

PR #13397, opened 2016-12-06
Revert "Flush DOM before displaying context menu"

### LINKED ISSUE #2991: Rightclick in the treeview doesn't focus the selected item directly

as-cii:
Reverts atom/atom#13266 as it doesn't address #2991.

/cc: @vjeux


vjeux:
👍

PR #13411, opened 2016-12-08
Suppress TextEditorScrollView scrolled when it shouldn't have warning

joefitzgerald:
Per @nathansobo, we’re not sure whether it’s even helpful to know that the text editor has scrolled in this way. It seems to occur often, and I notice it particularly when testing autocomplete-plus providers.

Before:

screen shot 2016-12-09 at 10 09 20 am

After: 

screen shot 2016-12-09 at 10 10 22 am


nathansobo:
👍  Thanks!

Issue #13206, opened 2016-11-12
[v1.13-beta] When two pane exist and destroy one pane, remaining editor does not focused


t9md:
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

### Description

When last paneItem(editor) was closed and pane was destroyed, remaining paneItem(editor) not correctly focused so I have to manually click by mouse.
It's very tedious, this issue is only happens in current beta(v1.13.0-beta2), no issue in stable(v1.12.1).

Maybe shadowDOM removal have some cause.

### Steps to Reproduce

- Confirm `Remove Empty Panes`(`core.destroyEmptyPanes`) is `true`(this is default value).
- Open some text-editor.
- Split editor into **two** by `cmd-k right`(or `cmd-k down` is OK), the point is **there are two panes in workspace**.
- Close pane item by `cmd-w`, then that pane is destroyed.
- The remaining editor on remaining pane don't get focused(expected: focus to remaining editor)

#### v1.12.1: OK no issue

- before: close editor

```
> atom.workspace.getPanes()
  [Pane, Pane]
> document.activeElement
  <atom-text-editor class=​"editor vim-mode-plus normal-mode" tabindex=​"-1" data-grammar=​"source coffee" data-encoding=​"utf8">​…​</atom-text-editor>​
```

- after: close editor and pane destroyed, remaining editor get focused.
```
> atom.workspace.getPanes()
  [Pane]
> document.activeElement
  <atom-text-editor class=​"editor vim-mode-plus normal-mode" tabindex=​"-1" data-grammar=​"source coffee" data-encoding=​"utf8">​…​</atom-text-editor>​
```

[...]

PR #12903, opened 2016-10-07
Remove shadow DOM from `atom-text-editor`

as-cii:
### Background

Back in #3943, we started rendering text editor contents inside a shadow DOM boundary to prevent outer style sheets from altering the editor inner contents. Although this introduced many benefits, it carried several tradeoffs too.

In particular, we started duplicating some of the style sheets (`{context: 'atom-text-editor'}`) inside each shadow root so that package and theme authors could still be able to explicitly target the editor visual elements. An alternative solution to this problem that we suggested (before the shadow DOM spec was finalized) was to use `/deep/` and `::shadow` pseudo selectors, which unfortunately are now [being deprecated](https://bugs.chromium.org/p/chromium/issues/detail?id=489954).

Even more importantly, this architecture was getting in the way of making `atom-text-editor` a reusable component, because it made it even more coupled to Atom and its ecosystem. Other problems involved things like #4590, and a more complex user experience for package and theme authors in general. Ultimately, we started feeling like the extra complexity of the shadow DOM wasn't worth its benefits.
### Removing shadow DOM from `atom-text-editor`

This pull request aims at removing the shadow DOM in a way that prevents elements within `atom-text-editor` that we don't fully control (i.e. the grammar scopes applied as CSS classes to every line's inner `<span>`) from being mistakenly styled from the outside. Specifically, every syntactic class name will now be prepended with `syntax--`; styling a JavaScript operator, for example, should now look like the following:

``` css
.syntax--source.syntax--js .syntax--operator {
  color: green;
}
```

Given the wide scope of this change, we are trying to reduce package breakage to a minimum by [transforming deprecated selectors automatically](https://github.com/atom/atom/blob/as-deprecate-shadow-dom/src/style-manager.js#L251-L306) and emitting a deprecation in [deprecation-cop](https://github.com/atom/deprecation-cop). Applying these transformations can be quite onerous and therefore we have introduced a layer of caching in 07d56b2 to ensure their performance impact stays low after the first time Atom is launched. 

Other notable changes include:
- Block decorations rendering has been rewritten from scratch (it previously worked only with the shadow DOM).
- When the editor is focused, `document.activeElement` will now return the hidden `<input>` inside `atom-text-editor`, instead of `atom-text-editor` itself. I could find only one third party package relying on this behavior, and I am going to submit a pull request shortly to use `document.activeElement.closest('atom-text-editor')` instead.
[...]

PR #13412, opened 2016-12-09
Fix issue where editor fails to regain focus

maxbrunsfeld:
When changing the workspace's pane configuration, the `PaneContainer` needs to sometimes remove intermediate `PaneAxis` objects from the DOM. When it does this, it [tries to restore focus](https://github.com/atom/atom/blob/89e3891d826e1c2eceb2d948203c668a4f65e0f3/src/pane-container-element.coffee#L17) to the previously-focused node.

This stopped working for text editors when we [removed the shadow DOM](https://github.com/atom/atom/pull/12903) because when a text editor is focused, it transfers focus to its hidden input element. This hidden input element was destroyed and re-created every time the editor was detached and re-attached.

This PR solves that problem by making the `TextEditorElement` keep a long-lived reference to a hidden input element, and re-using that same hidden input element every time it is attached to the DOM.

🍐 'd with @damieng

Fixes https://github.com/atom/atom/issues/13206

### DIFF
--- a/spec/text-editor-element-spec.coffee
+++ b/spec/text-editor-element-spec.coffee
@@ -78,6 +78,19 @@ describe "TextEditorElement", ->
       jasmine.attachToDOM(element)
       expect(element.querySelectorAll('.decoration').length).toBe initialDecorationCount
 
+    it "can be re-focused using the previous `document.activeElement`", ->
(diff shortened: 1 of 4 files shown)


nathansobo:
⚡️ Thanks guys.

PR #13415, opened 2016-12-09
Add desktop.ini to ignored names

portal-chan:


lee-dohm:
What do you think @damieng, reasonable?

Issue #10072, opened 2015-12-15
Instructions on how to make Atom portable


Zireael07:
I have Atom 1.3.1. As I recall, we were supposed to get instructions upon updating on how to make our install portable. I have looked everywhere and the program doesn't tell me how. Neither does your homepage, whether by itself or via Help/Documentation.

I need to move Atom off my C drive. Help?

50Wliu:
`atom --portable`
Does that work for you?

Zireael07:
Well, it works. But I still don't understand what that command line switch does. And I'm afraid to experiment with moving the Atom folder wholesale...

More documentation is definitely needed for this switch and what exactly it does. 
Searching for "portable" on atom.io doesn't find anything regarding the switch anywhere, just an old FAQ and forum discussions from Feb and Jul, so long before the switch was implemented. This needs to be rectified ASAP.

raelyard:
Adding to docs so it will be there.  In the meantime, a quick description here.

Portable mode is triggered by having a directory named .atom as a sibling to the directory in which atom.exe lives.  Atom will use this as the home directory instead of the .atom directory in the default location (~\..atom).

Atom's order of preference for finding the home directory is:
1. if an environment variable named ATOM_HOME exists, use the path specified there
2. if the Atom Portable directory exists (.atom as sibling of the parent directory of the atom.exe), use that
3. default location (~\..atom)

The command line switch takes the content of the default home directory and copies it the the portable location.  If, for instance, you have the atom.exe in a folder in your Dropbox called atom and run it with the command line switch, it would make a copy of the Atom Home from your home directory as a sibling of the atom directory in your Dropbox.

eugenesvk:
Pasting here from https://github.com/atom/atom/pull/8442#issuecomment-186853783, with details on why this priority order is incorrect.
Would you kindly make **Atom** prioritize a simple local setting `.alwaysportable` instead of trying to guess the user's setup system?

I have Atom installed locally with a custom path set to a Dropbox folder (so ATOM_HOME is defined in my environment).
Then I've downloaded and unpacked the zipped Atom version to my USB drive. However, when I run it, it still reads the Dropbox settings.
Starting from a command line with atom.exe --portable doesn't change anything either (it doesn't even create any sibling folders (on a side note: a portable app shouldn't have any siblings, everything should be in one folder)).

[...]

PR #13426, opened 2016-12-12
Portable mode enhancements

### LINKED ISSUE #5502: ATOM_HOME does not affect default Chromium storage path
hi

i'm on Win7 64bit and downloaded the newest
atom-windows.zip version 0.178.0 from
https://github.com/atom/atom/releases
to test the new ATOM_HOME env var feature.

i've set the env var ATOM_HOME to the path
D:\portable_programs\Atom\ATOM_HOME
and started atom.exe .
now in the folder
D:\portable_programs\Atom\ATOM_HOME
appeared only one dir "compile-cache" and
only one file "config.cson". it looks like
an incomplete part of the files and folders
that would normally be created on the first run
of atom.exe in the folder
C:\Users\ <user> \.atom
the content of "config.cson" looks different
than usual and contains the line
"exception-reporting"

the folder
C:\Users\ <user> \AppData\Roaming\Atom
was created in the same place, like without
a present ATOM_HOME env var. so for the folder
C:\Users\ <user> \AppData\Roaming\Atom
the redirection via ATOM_HOME does not
work in the actual Atom version 0.178.0 .

for the folder
C:\Users\ <user> \.atom
the redirection via ATOM_HOME env var
works partially and the files
.gitignore , init.coffee , keymap.cson ,
snippets.cson and styles.less and the
folder packages were not created, when
ATOM_HOME was set.

also i would expect that the files and folders of
C:\Users\ <user> \.atom
would be placed in the ATOM_HOME env var mode
in the same subfolder .atom like
<path in ATOM_HOME> \.atom
but they landed directly in
<path in ATOM_HOME> \

### LINKED ISSUE #13164: Portable version for macOS
I would LOVE to use two different portable version of Atom running at the same time, not sharing ANYTHING between them.

It seems this is available for Windows users with something like this:
https://github.com/garethflowers/atom-portable

But no love for macOS 😢

damieng:
This PR addresses a number of niggles with portable mode, namely:

1. Add macOS support - just create a .atom directory alongside Atom.app (fixes #13164)
2. Electron/Chromium userdata can now be portable (create electronUserData dir inside .atom - fixes #5502)
3. Removes the confusing `--portable` switch that didn't enable portable mode (it copied data)
4. Checks for a portable .atom folder in expected place before using ATOM_HOME (Comments in #10072)

Flight Manual needs to be updated once this is on stable.


johann1301h:
Any updates on portable mode for macOS?

damieng:
It shipped in the 1.14 beta yesterday.

johann1301h:
when running beta 1.14.0, the application still creates an .atom folder storing settings, packages etc. Shouldn't this folder exist within the application folder atom.app? (in this case "atom beta.app"). Or perhaps this must be activated after opening the editor?

damieng:
You must manually create a '.atom' folder alongside the atom.app and then Atom will use that folder instead of ~/.atom,

johann1301h:
Yes, that works. Great! But is there any way to change it so that it lies inside Atom.app?

damieng:
No, and you really don't want to do that as the first time you upgrade Atom you'll blow away all your settings, state and installed packages.

martinbean:
Apologies for commenting on a beta feature. But if I install *Atom.app* to *~/Applications*, and I’m supposed to create an *.atom* folder _alongside_ that, how does that make the app portable? Or am I supposed to install *Atom.app* in another, shared location (i.e. my Dropbox folder)?

I guess I’m just trying to understand the process in making Atom portable. I work across multiple Macs in multiple locations, and a portable version that has my settings, packages, snippets, etc on all machines would be a _godsend_.

damieng:
Sorry it's a little confusing - the docs for this aren't merged into the flight manual yet because the flight manual is always at stable not beta.

Basically you could always put Atom.app anywhere you wanted but it would look for a .atom folder in your home directory for its settings/extensions etc.

Atom now looks for a .atom folder first alongside the Atom.app on the mac and will use that instead.  So to get going simply:

Move Atom.app to somewhere 'portable' like a USB drive or dropbox folder then move your .atom folder from your home folder and place it alongside it.  Now you can use them from anywhere.

martinbean:
@damieng Gotcha. Thought that may have been the process. Thanks for clarifying!

jaredreich:
Shouldn't the .atom folder be INISDE Atom.app for this to be truly portable?

damieng:
Then you would lose your settings the first time you upgraded or replaced Atom.  

No macOS application stores individual settings in the .app for this reason.

jaredreich:
Right makes sense, but without that functionality this is not portable at all in the sense of being able to have two Atom app instances on the same machine without them fighting with eachother.

damieng:
'Portable' in this sense is referring to the application being able to live on a location you take between multiple machines - i.e. dropbox/usb.  You can place both the Atom.app and your .atom folder on there and launch the application and take your config/packages/settings with you.

jaredreich:
Thanks for the help!

awsdert:
Isn't the answer to how to make atom portable rather simple? Just add a launcher called atom.exe that defaults to identifying paths like ATOM_BIN, ATOM_LIB & ATOM_USR, all of which should be handed to a file called atom-paths.ini whos path should be passed as a command line argument to either atom-updater.exe or %ATOM_BIN%\atom-bin.exe as ATOM_PATHS_INI.

To get these paths the launcher should first check for ATOM_PATHS_INI in it's own arguments, if it doesn't exist call access( cwd(), R_OK | W_OK ) to check if can default to it's own directory, if not then defaults to %APP_DATA% or whatever variant it was, every update to atom.exe/atom-updater.exe should just download 2 files called atom-tmp.exe and atom.zip/7z etc, once this download is complete atom-tmp.exe should be launched, atom-tmp.exe should then wait for any other executable containing the name atom to finish before deleting atom.exe & atom-updater.exe and unpacking the archive subsequently exiting, when atom.zip/7z etc is detected atom.exe should check if atom-tmp.exe is running, if it is then atom.exe quits, otherwise it deletes atom-tmp.exe and the archive and then continues as normal.

With the above behavior atom would basically be treating local data as a last resort instead of first port of call, it also means it doesn't have to know if it is portable only where to put it's background stuff.

Issue #13389, opened 2016-12-05
Proper usage of script/build for a user-level install?


CodeMan99:
Building from source was not a problem for me up to (and including) **v1.10.2**.

Since then I have been met with multiple problems and errors while trying to build. Today I attempted to build **v1.12.6** and received an error identical to #12549.

Reading through the documentation it seems the workflow is to install at the system level. This doesn't work for me as I always install to "$HOME/.local" so that I replicate my build on servers where I don't have admin (root) access.

Previously I would do the following.

1. `./script/build`
2. `./script/test`
3. `./script/grunt install --install-dir "$HOME/.local"`

Every version since v1.10.2 (v1.11.* and v1.12.*) have failed to build on my system (step 1). I am unsure how to continue since the documentation only includes directions for a system-level install.

CodeMan99:
System info:

* OS: Ubuntu 15.10 (64 bit, Virtual Box VM)
* node: v6.9.1
* npm: v3.10.9
* gcc: v5.2.1

Ben3eeE:
https://github.com/atom/atom/issues/13240?

CodeMan99:
@Ben3eeE Related, but not the whole issue.

Just remembered about the **script/clean** script. Why is this not documented in "docs/build-inscructions/linux.md"?

nathansobo:
Open a PR?

CodeMan99:
Closing this issue because documenting `script/clean` and issue #13240 is what needs to be done.

Issue #13240, opened 2016-11-16
Allow to specify a prefix for installation


paulmenzel:
### Prerequisites

* [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
* [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

Installing Atom, it should be possible to configure the prefix for the installation. By default it’s `/usr/local`, but distribution people normally want to change that. For example, they want to install Atom in `/usr/bin`, and not `/usr/local/bin`.

That was possible in Atom 1.9.0.

```
$ script/grunt install --build-dir /something --install-dir /my/prefix/dir
```

It’s not anymore, as it’s hard coded in `src/command-installer.coffee`.

paulmenzel:
@as-cii, @nathansobo, judging from `git log` you work on the build system. It’d be great if you could clear that up.

nathansobo:
cc @iolsen for reactive prioritization.

damieng:
The hard-coded reference in command-installer.coffee is actually for installing command-line tools from an already-running version of Atom.

The install script is in install-application.js and currently attempts to installs to;

- Linux: `/usr/local/share/applications/atom`
- macOS: `/Applications`
- Windows: `%LOCALAPPDATA%\Atom\app-dev` 

We should probably provide support for all these operating systems with this config switch.

damieng:
[...]

PR #13447, opened 2016-12-14
Allow build script to specify install prefix.

CodeMan99:
This change is a _work in progess_.

Used twice successfully with v1.12.6 and v1.12.7, linux only.

Sparked from #13240 & #13389.

### DIFF
--- a/script/lib/install-application.js
+++ b/script/lib/install-application.js
@@ -7,7 +7,7 @@ const template = require('lodash.template')
 
 const CONFIG = require('../config')
 
-module.exports = function (packagedAppPath) {
+module.exports = function (packagedAppPath, prefix) {
   const packagedAppFileName = path.basename(packagedAppPath)
   if (process.platform === 'darwin') {
     const installationDirPath = path.join(path.sep, 'Applications', packagedAppFileName)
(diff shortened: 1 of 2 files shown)


CodeMan99:
So I am really completely unaware of what work-flows this might be affecting. Any advice is appreciated.

The only thing that seems obvious to me is that the last chunk (desktopEntryContents) shouldn't be "userLocalDirPath" in every case.

script/lib/install-application.js
```diff
@@ -67,7 +67,7 @@ module.exports = function (packagedAppPath) {
     const desktopEntryTemplate = fs.readFileSync(path.join(CONFIG.repositoryRootPath, 'resources', 'linux', 'atom.desktop.in'))
     const desktopEntryContents = template(desktopEntryTemplate)({
       appName, appFileName: atomExecutableName, description: appDescription,
-      installDir: '/usr', iconPath
+      installDir: userLocalDirPath, iconPath
```
damieng: This line actually changes the default here from being /usr to /usr/local - is that intentional?

damieng:
No response regarding issue.

Issue #12391, opened 2016-08-10
Atom 1.9.8 crashes when trying to open 9.2MB XML file on OS X 10.11.6


baz8080:
### Prerequisites
- [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)?
- [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.
### Description

I am going to develop an English / Swedish diction for Android, so I downloaded 4 data files into a directory:
- http://folkets-lexikon.csc.kth.se/folkets/folkets_en_sv_public.xml
- http://folkets-lexikon.csc.kth.se/folkets/folkets_sv_en_public.xml
- http://folkets-lexikon.csc.kth.se/folkets/folkets_en_sv_public.xdxf
- http://folkets-lexikon.csc.kth.se/folkets/folkets_sv_en_public.xdxf

If I attempt to open `folkets_en_sv_public.xdxf`, atom will hang and crash.
### Steps to Reproduce
1. Download all of the files above
2. Start atom in safe mode
3. Open `folkets_en_sv_public.xdxf`

**Expected behavior:** 

Editor renders XML

**Actual behavior:** 

atom hangs, and then crashes
### Versions

You can get this information from executing `atom --version` and `apm --version` at the command line. Also, please include the OS and what version of the OS you're running.

_atom versions_

```
Atom    : 1.9.8
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0
```

[...]

Issue #8864, opened 2015-09-19
Loading large file with many lines sometimes hangs or crashes the editor


fewspider:
File that has 140000 lines can reproduce this issue.
Sublime Text 3 release version has no this problem.

OmgImAlexis:
Duplicate https://github.com/atom/atom/issues/979

50Wliu:
Thanks @OmgImAlexis.

mnquintana:
I'm actually not sure if this is a dupe of #979 – @fewspider mentions that this happens with a file with 140,000 lines, rather than a single really long line.

@fewspider Can you reproduce this by completely quitting Atom and restarting it in in safe mode? (`atom --safe`)

fewspider:
@mnquintana My atom version is `1.0.15`, `atom --safe` is OK👌. My file is a sql file, the syntax highlighting is not working,when `atom` or `atom --safe`. Just a guess, the syntax highlighting problem happens in large file.

centic9:
Sounds like being related to syntax highlighting and thus sounds like a duplicate of #6642

alexchandel:
I can confirm, to this very day, that Atom (1.8.0 safe mode, OSX and Windows) hangs forever or crashes (it did both today) when you try and open a 1MB XML file with only two lines.

mnquintana:
@alexchandel That's https://github.com/atom/atom/issues/979 – this issue is only for large files with many lines causing hangs.

alexchandel:
@mnquintana Ah of course. For the record, neither the UTF-8 nor the UTF-16LE file I tried crashed Atom when I threw a newline after every XML tag (33588 lines total). It seems Atom only hangs or crashes (depending on the file encoding, funnily enough) when the lines are long (1MB in this case).

That's still atrocious behavior though. It completely defeats the point of using Atom to beautify text if Atom crashes trying to open the ugly file.

wogandavid:
Also experiencing crashed when opening large files. I typically need to open GAMS listing files (.lst) which can run from several MBs for a small model run to > 100 MB.

y6nH:
Me too - large XML or JSON files, Atom 1.9.8 on Windows 7, safe mode or not. The current example is a 14 MB json file with 300k lines, pretty-printed with no line more than 70 characters long.

Here's the bit from the console output that looks most relevant:

```
Security context: 2985BF09 <String[7]: file://>
[...]

Issue #11690, opened 2016-05-04
Uncaught RangeError: Invalid string length


szabba:
1. Open Atom.

**Atom Version**: 1.7.3
**System**: "Arch Linux"
**Thrown From**: Atom Core
### Stack Trace

Uncaught RangeError: Invalid string length

```
At /usr/share/atom/resources/app.asar/src/buffered-process.js:83

RangeError: Invalid string length
    at Socket.<anonymous> (/usr/share/atom/resources/app.asar/src/buffered-process.js:83:23)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Socket.Readable.push (_stream_readable.js:110:10)
    at Pipe.onread (net.js:523:20)
```
### Commands

```
  5x -2:00.3.0 core:move-down (atom-text-editor.editor.hidden-line-numbers.is-focused)
  7x -1:58.2.0 core:move-up (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -1:53.1.0 editor:newline (atom-text-editor.editor.hidden-line-numbers.is-focused)
  2x -1:52.3.0 core:backspace (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -1:51.2.0 core:save (atom-text-editor.editor.hidden-line-numbers.is-focused)
 10x -1:49.8.0 core:move-down (atom-text-editor.editor.hidden-line-numbers.is-focused)
  3x -1:48.2.0 core:move-up (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -1:47.1.0 core:move-down (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -1:46.7.0 core:move-up (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -0:08.1.0 editor:consolidate-selections (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -0:08.1.0 core:cancel (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -0:08.0 editor:consolidate-selections (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -0:08.0 core:cancel (atom-text-editor.editor.hidden-line-numbers.is-focused)
     -0:07.4.0 editor:consolidate-selections (atom-text-editor.editor.hidden-line-numbers.is-focused)
  6x -0:07.4.0 core:cancel (atom-text-editor.editor.hidden-line-numbers.is-focused)
[...]

PR #13453, opened 2016-12-14
Use new text-buffer version with a more efficient DisplayLayer

maxbrunsfeld:
Depends on https://github.com/atom/text-buffer/pull/185. See that PR for a more thorough description.

Fixes https://github.com/atom/atom/issues/8864
Fixes https://github.com/atom/atom/issues/12391

This removes some restrictions that we previously enforced when editing large files; you can now use folds and soft wraps. The only feature that remains disabled for > 10MB files is syntax highlighting.

Large files also now open somewhat faster. Here are some benchmark results from Circle CI [before](https://circleci.com/gh/atom/atom/2204?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and [after](https://circleci.com/gh/atom/atom/2205?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) this PR:

benchmarks

There are lot of noise in these results of course, but opening a 10MB file is consistently ~200ms faster.


package.json
```diff
@@ -65,7 +65,7 @@
     "sinon": "1.17.4",
     "source-map-support": "^0.3.2",
     "temp": "0.8.1",
-    "text-buffer": "9.4.3",
+    "text-buffer": "v10.0.0-4",
```
50Wliu: This should be 10.0.0-4 to be consistent, right?

maxbrunsfeld: Oh yeah, thanks. I'll fix this when I bump it to `10.0.0`.

Ben3eeE:
Really loving this :racehorse: :heart_eyes_cat: will test it out more with the new `spell-check` and `whitespace` changes tonight(Late American morning). But don't wait for me if you want to go ahead and :shipit: 

@ungb @maxbrunsfeld Want to drop https://github.com/atom/atom/issues/11690 here since it is an uncaught that reproduces with large files. We don't have a file to repro or any steps other than open a large file but I believe it can be worth some :eyes: given these changes.

maxbrunsfeld:
@Ben3eeE I believe #11690 is due to a separate limitation: we store the entire contents of a file in JavaScript string, and strings have a hard limit on their lengths. This PR does not fix that issue.

Stanzilla:
@maxbrunsfeld This changes nothing for https://github.com/atom/atom/issues/1667, right?

maxbrunsfeld:
> This changes nothing for #1667, right?

Right. It does not address that issue.

alexandernst:
@maxbrunsfeld Is there an issue which we can track for ast parsing/syntax highlight? Where can we expect such a feature to be ready?

PR #13460, opened 2016-12-17
Use package-specific transpilation options during tests

BinaryMuse:
Turns out that packages were not transpiled using the transpiler under `atomTranspilers` during tests due to the way the test window is initialized. This fixes that issue.

/cc @atom/core


nathansobo:
Looks good to me.

PR #13468, opened 2016-12-18
Added keymap to insert date in input buffer

pedroduartecosta:
This feature allows the user to quickly add the current date to the editor buffer by pressing 'ctrl-alt-d' :

### DIFF
--- a/keymaps/base.cson
+++ b/keymaps/base.cson
@@ -8,6 +8,7 @@
 'atom-text-editor:not([mini])':
   # Atom Specific
   'ctrl-C': 'editor:copy-path'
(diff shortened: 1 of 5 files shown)


lee-dohm:
Thanks very much for the submission!

Atom encapsulates most of its features in additional packages. Even basic features like the Tree View, tabs and Find and Replace are in separate packages. Because of this design, including this in atom/atom wouldn't be appropriate. Additionally, this isn't something that we would be including in any of the other built-in Atom packages either. It would be a great start to your own community package though!

Thanks again for taking the time to contribute :+1:

PR #13475, opened 2016-12-20
Fix bad title bar path when viewing files not in a project folder

### LINKED ISSUE #5058: File location is wrong on title
qq20150115-6

qq20150115-7

You can find the correct location on the tree view, but on title the location don't change.

If this location is represent for the current folder on the view, it will confuse the user. It will be a disaster for a user editing 2 files with the same name in different folders.

I think this title is to indicate the current file location, so this must be a bug..

### LINKED ISSUE #5585: Title bar is not appropriate for non-project files
When a project is open (i.e. the file browser is open), if I open the LICENSE or other non-project file, then the title bar will show like:

> LICENSE -- /path/to/project

This is misleading because it implies that the `LICENSE` file is inside the /path/to/project when it is not. In fact... I have no idea where the hell LICENSE is. It would be better to have something else in the title bar. Recommendations are:
- `LICENSE`
- `Atom - LICENSE`
- `Atom - Version 0.179.0 - LICENSE`

Now, for example, when I open "Open your Config", the titlebar shows `config.cson -- /path/to/project`, I am unsure if this is a global config file or if the file is actually saved at /path/to/project/config.cson.

screen shot 2015-02-17 at 10 23 46 am

screen shot 2015-02-17 at 10 22 36 am

### LINKED ISSUE #10710: Directory path on titlebar is not following selected file
The screenshot below shows the titlebar is pointing to ~/.atom/timeline.txt  This is especially troubling because the file path is actually /nok/timeline.txt    So the result is the user (myself..) panics thinking that the original file were lost or displaced.

In fact the only thing happening is that atom is handling the current directory separately from the current file/buffer. It then has a bug in the current directory - using the wrong one. Atom then combines the (incorrect) directory with the current buffer name and we end up with a non-existing path. 

This is annoying because what happens when we have multiple files with the same buffer name but in different directories ?  that is in fact my case here - i have two different timeline.txt files and I do not know to which one atom is referring.

If it helps I am on Yosemite.

The screenshot shows that Atom mixed up the path of the current file "timeline.txt" with the prior one "settings.xml".

screen shot 2016-02-05 at 1 47 38 pm

BinaryMuse:
In cases where the active pane item has a path that is not inside a current project folder, we display the first project folder in the title bar. This can be misleading. For instance, with a project setup like this:

```
project
└ folder1
  └ package.json
└ folder2
  └ package.json
```

If we open `~/random/path/package.json`, the title bar will show `package.json — folder1`, which is incorrect. This made more sense previously, when the project only contained a single folder, but now the behavior seems misleading.

In order to provide the most pertinent information, this PR changes the behavior to the following:

* If the active pane item has a path that is included in a project folder, we show `item title — project folder` (as is the current behavior)
* If the active pane item has a path that is NOT included in a project folder, we show `item title — path to file` (where the path does not include the filename — in the example above, it'd be `~/random/path`)
* If there is no active pane item (i.e., the pane is empty), we simply show `untitled`

Fixes #5058
Fixes #5585
Fixes #10710

/cc @atom/feedback

### DIFF
--- a/spec/workspace-spec.coffee
+++ b/spec/workspace-spec.coffee
@@ -885,12 +885,44 @@ describe "Workspace", ->
       expect(coffeePackage.loadGrammarsSync.callCount).toBe 1
 
   describe "document.title", ->
-    describe "when the project has no path", ->
+    describe "when there is no item open", ->
       it "sets the title to 'untitled'", ->
-        atom.project.setPaths([])
-        expect(document.title).toMatch ///^untitled///
+        expect(document.title).toEqual "untitled"
 
-    describe "when the project has a path", ->
+    describe "when the active pane item's path is not inside a project path", ->
+      beforeEach ->
+        waitsForPromise ->
+          atom.workspace.open('b').then ->
+            atom.project.setPaths([])
+
+      it "sets the title to the pane item's title plus the item's path", ->
+        item = atom.workspace.getActivePaneItem()
+        pathEscaped = fs.tildify(escapeStringRegex(path.dirname(item.getPath())))
+        expect(document.title).toMatch ///^#{item.getTitle()}\ \u2014\ #{pathEscaped}///
+
+      describe "when the title of the active pane item changes", ->
+        it "updates the window title based on the item's new title", ->
+          editor = atom.workspace.getActivePaneItem()
+          editor.buffer.setPath(path.join(temp.dir, 'hi'))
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Awesome. Thanks for fixing this!

lee-dohm:
Looks good to me! 👍

nathansobo:
Looks like this introduced a regression where the window title is `untitled` when there are no pane items open: https://github.com/atom/atom/issues/13647.

Investigating a solution now.

PR #13493, opened 2016-12-22
Add atom.tooltips.findTooltips(target)

vjeux:
Right now, it is not possible to hide a tooltip programatically. This is useful when we know better than the tooltip implementation that we did an action that should hide it.

After discussing with @lee-dohm, he suggested the findTooltips API that mimicks the KeyMapManager API.

Released under CC0

### DIFF
--- a/src/tooltip-manager.coffee
+++ b/src/tooltip-manager.coffee
@@ -130,7 +130,7 @@ class TooltipManager
 
     tooltip = new Tooltip(target, options, @viewRegistry)
 
-    if !@tooltips.has(target)
+    if not @tooltips.has(target)
       @tooltips.set(target, [])
     @tooltips.get(target).push(tooltip)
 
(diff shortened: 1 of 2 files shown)


src/tooltip-manager.coffee
```diff
@@ -129,19 +130,37 @@ class TooltipManager
...
+        if index != -1
+          tooltipsForTarget.splice(index, 1)
+        if tooltipsForTarget.length == 0
+          @tooltips.delete(target)
+
     disposable
 
+  findTooltips: (target) ->
+    if @tooltips.has(target)
+      @tooltips.get(target)
```
lee-dohm: I noticed that other implementations of the `find` pattern use `slice` to create a shallow copy of the array and return the copy. I assume they're doing that to prevent people from corrupting the source array. Can we use `@tooltips.get(target).slice()` here?

vjeux: Sure, good idea!

lee-dohm:
Thanks for the help! 👍

vjeux:
Yaaay!

PR #13523, opened 2016-12-31
Reduce the impact of leaking Editor, Selection, & Cursor objects

maxbrunsfeld:
It is pretty easy for 3rd party Atom packages to accidentally introduce memory leaks, causing destroyed `TextEditor` objects to be retained on the heap indefinitely. This PR updates editors and their sub-objects to do a more aggressive teardown when they are destroyed, so that leaking these objects will have a negligible impact on Atom's memory usage, even for large files.

See also https://github.com/atom/text-buffer/pull/193

🍐'd with @nathansobo


nathansobo:
⚡️

PR #13421, opened 2016-12-10
Multiple fixes for RPM and in docs RPM distros.

philwyett-hemi:
* Correct requires for libXss across Fedora/RedHat (and derivatives) and SUSE.
* Split instructions for Fedora and RHEL etc. i.e. yum and dnf.
* Never use --assumeyes or -y installing packages. atom 1.12.7 shows if
  this had been done by end users it would have caused them great issue.
* Fedora 22 is End Of Life (EOL), so no instructions needed.

Issue #13417, opened 2016-12-09
1.12.7 (64 bit) dep on libXss.so.1 tries to dep install 686 packages


philwyett-hemi:
Terminal below from CentOS 7.3 (7.3 packages as of Continuous Release CR repo)

```
[philwyett@hemi-yoda Downloads]$ sudo yum localinstall atom.x86_64.rpm 
[sudo] password for philwyett: 
Loaded plugins: changelog, fastestmirror, langpacks
Examining atom.x86_64.rpm: atom-1.12.7-0.1.x86_64
Marking atom.x86_64.rpm as an update to atom-1.12.6-0.1.x86_64
Resolving Dependencies
--> Running transaction check
---> Package atom.x86_64 0:1.12.6-0.1 will be updated
---> Package atom.x86_64 0:1.12.7-0.1 will be an update
--> Processing Dependency: libXss.so.1 for package: atom-1.12.7-0.1.x86_64
adobe-linux-x86_64                                       |  951 B     00:00     
base                                                     | 3.6 kB     00:00     
cr                                                       | 3.4 kB     00:00     
epel/x86_64/metalink                                     |  18 kB     00:00     
epel                                                     | 4.3 kB     00:00     
extras                                                   | 3.4 kB     00:00     
hemi                                                     | 3.0 kB     00:00     
hemi-epel                                                | 3.0 kB     00:00     
updates                                                  | 3.4 kB     00:00     
(1/10): base/7/x86_64/group_gz                             | 155 kB   00:00     
(2/10): epel/x86_64/group_gz                               | 170 kB   00:01     
(3/10): extras/7/x86_64/primary_db                         | 166 kB   00:01     
(4/10): hemi/primary_db                                    |  11 kB   00:00     
(5/10): hemi-epel/primary_db                               | 3.7 kB   00:00     
(6/10): epel/x86_64/updateinfo                             | 679 kB   00:04     
(7/10): epel/x86_64/primary_db                             | 4.4 MB   00:10     
(8/10): cr/7/x86_64/primary_db                             | 3.7 MB   00:15     
(9/10): base/7/x86_64/primary_db                           | 5.3 MB   00:15     
(10/10): updates/7/x86_64/primary_db                       | 9.1 MB   00:16     
adobe-linux-x86_64/primary                                 | 1.2 kB   00:00     
Determining fastest mirrors
Including mirror: mirror.bytemark.co.uk
Including mirror: anorien.csc.warwick.ac.uk
[...]

PR #13536, opened 2017-01-04
Fixing Install issue for linux redhat distro

ungb:
### Requirements

Tested on Centos.  the logic before tried to use || for or, but seems to return false for redhat distros, changed statement based on doc here: https://fedoraproject.org/wiki/Packaging:DistTag.  I am new to the build script logic.  @as-cii  @iolsen  can you take a look?  It's working for me on a clean install of centos on VM. 

/cc @philwyett-hemi since I'm modifying the logic that was merged.

### Description of the Change

Change logic in if statement for require for redhat distro.

### Alternate Designs

N/A

This is currently breaking for Centos and Redhat users.

### Benefits

Fixes issue with installing on centos and redhat.

### Possible Drawbacks

N/A, after the last change this is already broken.

see comments in https://github.com/atom/atom/pull/13421

### Applicable Issues

https://github.com/atom/atom/issues/13417

### DIFF
--- a/resources/linux/redhat/atom.spec.in
+++ b/resources/linux/redhat/atom.spec.in
@@ -8,7 +8,7 @@ AutoReqProv:    no # Avoid libchromiumcontent.so missing dependency
 Prefix:         <%= installDir %>
 
 Requires: lsb-core-noarch
-%if 0%{?fedora} || 0%{?rhel}
+%if 0%{?rhel}%{?fedora}
 Requires: libXScrnSaver
 %else
 Requires: libXss1


iolsen:
👍 LGTM

I'll hotfix this tomorrow if nobody beats me to it.

as-cii:
I recall testing #13421 on Fedora but it was probably an old version (or a new one maybe?). Either way, thanks so much for the fix! This looks good to me too. Before merging, do you think we should test it on Fedora and OpenSUSE as well?

ungb:
works on Fedora too!

image

ungb:
I actually don't think my change makes any difference, after further testing, I realize the issue in the end is where we are building.  The conditional statement I'm guessing, is done during build time and once you have the .rpm, it will require the same regardless of which distro you are using.  I tested building on Centos, OpenSuse, and Ubuntu and saw that the build from both OpenSuse and Ubuntu generated the same require for libXss1 where as Centos and Fedora had the libXScrnSaver.

I'm guessing this change won't really fix the issue.

PR #13539, opened 2017-01-04
Call debounced cursor blink resume callback any time the cursor moves

nathansobo:
Previously, we were only calling the debounced resume callback when the cursor was already blinking, but this meant we didn't renew the debounce window. This in turn meant we attempted to start blinking the cursor again every 100ms regardless of cursor interactions, which was consuming an extra frame and causing cursor movement to not feel smooth.


maxbrunsfeld:
Appveyor failure is a known spurious thing.

PR #13545, opened 2017-01-04
Don't convert paths to real paths when building text buffers

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/9879
Refs https://github.com/atom/atom/pull/10856
Refs https://github.com/atom/text-buffer/pull/182
Refs https://github.com/facebook/nuclide/issues/697

/cc @dirk-thomas I have talked to some others on our team, and reconsidered this problem. I feel really bad for the delay in resolving this issue, and all of the work that you've done to propose solutions.

Where we stand right now, we think that the right solution is probably to simply remove the conversion to real paths. There may be some breakage of community packages, but it's hard for me to see how.

Please let me know what you think.

Discussed this with @nathansobo
/cc @noseglid because you maintain `atom-build` and this relates to https://github.com/noseglid/atom-build/issues/245

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "normalize-package-data": "^2.0.0",
     "nslog": "^3",
     "oniguruma": "6.1.0",
-    "pathwatcher": "6.8.0-0",
+    "pathwatcher": "6.8.0",
     "postcss": "5.2.4",
     "postcss-selector-parser": "2.2.1",
     "property-accessors": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "keybinding-resolver": "0.35.0",
     "line-ending-selector": "0.5.1",
     "link": "0.31.2",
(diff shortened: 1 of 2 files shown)


dirk-thomas:
In atom/atom#9879 the concern was brought up that such a change might be too much to be introduced in Atom 1.x. But I agree that just updating the existing `resolvePath` as in this patch is definitely the cleaner more future proof approach. I would be very happy if this get merged into an upcoming Atom release. I am sure I can then address the original problem in the `build` packages to support workspaces which contain symlinks. :+1:

maxbrunsfeld:
Ok, if you're cool with this and it passes our CI (or can be made to pass with minor adjustments to our test suites) it'll go out in the Atom 1.14 beta release in a couple of weeks.

maxbrunsfeld:
It's also just one less synchronous file system call 🐎 .

Issue #13548, opened 2017-01-05
`Config` class needs documentation about keys that aren't in the config schema


lee-dohm:
The [`Config` API class](https://atom.io/docs/api/latest/Config) has a lot of documentation around defining a configuration schema for your package. What it doesn't have good documentation for is that you can `get`, `set`, `observe` and etc values that do not appear in the configuration schema for your package. For example, if "some-package"s config schema is:

```
config:
  someSetting:
    type: 'boolean'
    default: false
```

One can still do the following:

```
let otherSetting  = atom.config.get('some-package.otherSetting')
atom.config.set('some-package.stillAnotherSetting', otherSetting * 5)
```

This needs to be made more clear in the documentation.

shreyasminocha:
I'm new here. I'm interested in helping out with this. Is there something I need to know? Also, where can I find the docs?

lee-dohm:
@shreyasminocha The above :point_up: is most of what you need to know. Mostly the information that needs to be conveyed is that any time a function asks for a `keyPath`, that path doesn't have to be described in the config schema for the package or any package. But it is still good form to only use key paths in your own package's namespace.

You can find the documentation on [the `Config` class itself in the source code](https://github.com/atom/atom/blob/master/src/config.coffee#L17-L344). That's what you'll need to update.

Thanks for helping out :tada:

lee-dohm:
Fixed in atom/atom#13563.

PR #13563, opened 2017-01-07
Updated documentation about keys outside config schema

shreyasminocha:
### Description of the Change

This PR addresses #13548 and adds documentation to config.coffee about accessing and editing keys that do not appear in the configuration schema of the package and also mentions that it is against best practices to do the same.


lee-dohm (approved):

lee-dohm:
Thanks for contributing! Sorry it took so long to get back to you on this, I've been traveling the last couple weeks :blush:

PR #13564, opened 2017-01-07
Save initial paths immediately on change

### LINKED ISSUE #12495: Atom doesn't remember added projects on start up

vovkasm:
This is #12545 (send event to save state after 'Add Project Folder' or 'Remove Projec') + tests. Does it appropriate for merge now?

Copy of the description from #12545:

Fixes #12495

When we add or remove project folder, the window's current project path will be changed, then "initialPaths" will be reloaded, it is the best time to save state

/cc @maxbrunsfeld @lee-dohm

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -47,13 +47,6 @@ describe('AtomApplication', function () {
       const dirB = makeTempDir()
       const atomApplication = buildAtomApplication()
 
-      let cnt = 0
-      const origSaveState = atomApplication.saveState
-      const mockSaveState = function () {
-        cnt++
-        origSaveState.apply(atomApplication, arguments)
-      }
-
       const window = atomApplication.launch(parseCommandLine([]))
       await focusWindow(window)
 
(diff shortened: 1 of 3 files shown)


vovkasm:
Okey, windows fails... most probably because I try to read `ATOM_HOME/storage/application.json` too fast after creation.

There is some variants (and I need help from atom devs to select most appropriate one):

- We can simple test that `AtomApplication#saveState` method called (aka test spy).
- We can add short sleep to ensure application.json saved (something like in [this test](https://github.com/atom/atom/blob/4c4fa2b478b6d1254167cb0446f3896ef2e3b489/spec/main-process/atom-application.test.js#L231) )
- We can test this functional in unit like - split to two tests:
  - `did-change-paths` event generated on right places
  - `did-change-paths` event handled, i.e. call saveState

vovkasm:
Bad, seems like on Windows AtomApplication#saveState does not called at all, will investigate...

vovkasm:
Hmm... I will try to setup dev environment on windows to fix that, but it will need some time...
If anyone with windows can fix it - welcome!

vovkasm:
Done. Now it pass tests.
After fixing test code for windows, I reverted tests to initial behavior to check actual `ATOM_HOME/storage/application.json` content.
I can if needed squash my commits into one and rebase this branch.

maxbrunsfeld:
@vovkasm No need to squash or rebase these commits; they're fine as they are. Really nice job figuring out how to test this! I'll take a look on Monday.

maxbrunsfeld:
👏  Great work @yuanwhy, @vovkasm.

maxbrunsfeld:
This will go out in Atom 1.14 Beta.

PR #13574, opened 2017-01-09
Add Arch Linux and Parabola GNU/Linux-libre

foxmean:
Just add installing of Arch Linux and Parabola GNU/Linux-libre.


damieng:
The version of Atom available via pacman is an unofficial distribution of Atom that is regularly broken as the maintainer uses a different version of Electron. 

We would definitely not be instructing users to install it from there. 

At this time if you want to install an official version of Atom on any Linux distribution that isn't directly supported you will need to build from source.

PR #13583, opened 2017-01-10
Disable periodic serialization for buffers > 2MB

nathansobo:
The overhead of writing large amounts of data to IndexedDB is blocking the main thread for perceptible amounts of time, even though the API is asynchronous. A patch-based representation of the modified state could allows us to pay only for what has changed, but is too complex to justify implementing right now to support full crash recovery for large files. So instead, we'll only serialize the full text of files > 2MB on graceful shutdown of the window. Periodic serialization, which interleaves with fine-grained user interaction, will be limited to smaller files to ensure we don't make the editor unresponsive.

/cc @maxbrunsfeld

### DIFF
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -70,7 +70,15 @@ class Project extends Model
   serialize: (options={}) ->
     deserializer: 'Project'
     paths: @getPaths()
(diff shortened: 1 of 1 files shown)


nathansobo:
We've decided to do more work to enable modifications on larger files to be persisted, but it will take a while to reach the light of day. I still think pause free editing is worth disabling this in the meantime.

PR #13585, opened 2017-01-11
Always reserve two digits on the line number gutter

vjeux:
We are working on a feature that changes the content of the editor when you mouse over some things and it makes the UI jump when going from 1 line to > 10. This makes the UI feel really bad.

I've looked at Sublime and the 1-9 state is the same as 1-99, only when you reach 100 lines then it jumps. I think that it is a better behavior as you want to minimize jumps as much as possible and it is extremely likely that you are going to hit the 9-10 lines threshold.

While this is being reviewed and until the new version shipped, we are going to monkeypatch Atom in order to get this feature.

```js
var presenter = atom.textEditors.editors.entries().next().value[0].presenter.__proto__;
var old_updateLineNumberGutterState = presenter.updateLineNumberGutterState;
presenter.updateLineNumberGutterState = function() {
  var res = old_updateLineNumberGutterState.apply(this, arguments);
  this.lineNumberGutter.maxLineNumberDigits = Math.max(2, this.lineNumberGutter.maxLineNumberDigits);
  return res;
};
```

Released under CC0

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -566,7 +566,7 @@ describe('TextEditorComponent', function () {
           editor.setSoftWrapped(true)
           runAnimationFrames()
 
-          componentNode.style.width = 16 * charWidth + wrapperNode.getVerticalScrollbarWidth() + 'px'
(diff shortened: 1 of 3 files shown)


lee-dohm:
Looks good! Thanks a bunch for your continued contributions @vjeux :bow:

PR #13600, opened 2017-01-12
Setting for settings-view to apply Chromium proxies to apm via env vars

damieng:
The on/off setting for automatic apm proxies, see https://github.com/atom/settings-view/pull/898 for more details.

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -142,6 +142,11 @@ const configSchema = {
         type: 'boolean',
         default: true
       },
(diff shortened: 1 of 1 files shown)


50Wliu:
Nit: A `title` should be added for this that properly capitalizes `apm`.  Much like https://github.com/atom/atom/blob/682532c3070c74a0f15e2fb7aa424201b25437eb/src/config-schema.js#L24.

PR #13047, opened 2016-10-22
Change editor:duplicate-lines to only duplicate once per line

seusher:
Adds a fix to only duplicate a line once, no matter how many selections exist on the same line. The selections are set correctly on the duplicated line.
anim

Fixes https://github.com/atom/atom/issues/12927

PR #13602, opened 2017-01-12
Handle multiple selections intersecting a line in duplicateLines

nathansobo:
duplicate lines 2

Closes #13047 because I opted to take a fairly different approach to also solve for the case where selections spanning multiple lines intersect the same line more than once. That PR also didn't have tests and I didn't have push access to the branch. Thanks to @seusher for getting the process of fixing this started though.

/cc @ungb for more testing. I suggest you try to break this with diabolical selections.


nathansobo:
@ungb Yeah, that's actually intentional. The idea is to duplicate the minimal number of lines intersected by selections. In the case where selections are on adjacent lines but don't span across them, I chose to treat them individually. To me that seems like the most intuitive behavior. Did you find anything else or can we 🚢 ?

Issue #8221, opened 2015-08-04
Path for atom.workspace.open is case sentive


hammeron-art:
Although something like C:\path\Main.ext is the same thing as C:\path\main.ext for the Windows filesystem, Atom opens different windows. One with the title "Main.hx" and other with "main.hx"...
They point to the same file. If I edit and save one the other window is updated.

I'm using a tool that output the path always in lower case and atom.workspace.open to open in a given line and column.

On Windows 8.1, Atom 1.03, Javascript, the follow open two windows for the same file:

``` javascript
atom.workspace.open(path, {
  initialLine: line,
  initialColumn: column,
  searchAllPanes: true
});
```

``` javascript
atom.workspace.open(path.toLowerCase(), {
  initialLine: line,
  initialColumn: column,
  searchAllPanes: true
});
```

damieng:
Given the cross-platform nature of Atom and the case-sensitivity across systems we have decided not to change the behavior on Windows to try and do something different there and will continue to treat paths case-sensitive - this is especially important for other code paths such as opening urls which are still case-sensitive on Windows.

Arcanemagus:
As this is causing issues in some packages it would be nice if this could be fixed here. For now it looks like a workaround like bringing in [true-case-path](https://www.npmjs.com/package/true-case-path) might need to be done in packages where this is a problem.

Arcanemagus:
Note that this is also an issue on OSX (any file system where case isn't sensitive to the file system for resolving a file).

PR #13613, opened 2017-01-13
Normalize disk drive letter in path on Windows

hron:
### Description of the Change

Currently atom creates two buffers for the same file if passed paths use difference case for disk drive letter, e.g. `d:\file.txt` and `D:\file.txt`

This change ensures that the absolute path always has disk drive letter in uppercase. 

### Alternate Designs

The other place where this fix could be applied is [findBufferForPath](https://github.com/atom/atom/blob/master/src/project.coffee#L331) function. However the used function [resolvePath](https://github.com/atom/atom/blob/master/src/project.coffee#L229) already has some path normalization code, so I think this is the best fit.

Actually Windows interpreters `file.txt` and `File.txt` as the same file. Maybe it makes sense to extend this patch and make Atom to do the same on this operation system, but I'm not sure because it can be file system specific. Any suggestions are welcome!

I believe the current behavior of the editor is wrong for Windows users, so it makes no sense to create a plugin to fix that.

### Benefits

The main benefit I would like to get is the ability to click on parsed backtrace in my terminal emulation app (ConEmu). Right now it works, but atom opens second buffer for the same file because it doesn't normalize disk drive letter.

### Possible Drawbacks

The possible drawback is some broken tests, but I've checked that and it seems everything works.

### Applicable Issues

Here is the related issue #8221. It's not exactly about my problem, but it's very near.

### DIFF
--- a/spec/default-directory-provider-spec.coffee
+++ b/spec/default-directory-provider-spec.coffee
@@ -28,6 +28,15 @@ describe "DefaultDirectoryProvider", ->
       directory = provider.directoryForURISync(nonNormalizedPath)
       expect(directory.getPath()).toEqual tmp
 
+    it "normalizes disk drive letter in Windows path", ->
+      provider = new DefaultDirectoryProvider()
+      nonNormalizedPath = tmp[0].toLowerCase()+tmp.slice(1)
+      expect(!tmp.search(/^[a-z]:/)).toBe false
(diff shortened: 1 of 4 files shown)


abishekvashok (approved):

Abhi2424shek:
Squash your commita

hron:
Squashed commits are available for review and merge here: #13725

Issue #13249, opened 2016-11-16
Custom title bar on Windows/Linux


erroid:
Using dark theme and having white title bar is not nice. Please make it more consistent with rest of the app like in Adobe PS or  Brackets (also look #4599).

50Wliu:
Blocked on electron/electron#765.

erroid:
How about Windows?

50Wliu:
Windows does not have a builtin dark theme, and no other program that I can find updates their title bar based on the accent color.

If you mean high contrast, there are a bunch of different Electron issues related to that as well.

erroid:
I talk about title bar (image attached). There is a lot of programs that use their own designed title bars instead of those delivered by windows system, here are some examples:

![atom-title](https://cloud.githubusercontent.com/assets/10036579/20399068/e68ad55c-acef-11e6-8369-6dd8a8d71f52.png)

jerone:
A few more on Windows programs that have a themed title-bar:
* Visual Studio.
* All Microsoft Office programs.

alexlur:
Related: This feature has just [landed](https://github.com/Microsoft/vscode/issues/15098) in Visual Studio Code Insiders.

50Wliu:
Cool, looks like I misunderstood the issue to be talking about Window's accent color vs the Atom theme.  I completely forgot about MS Office.  Going to cc @simurai then to see what can be done.

simurai:
Electron has [an option](http://electron.atom.io/docs/api/frameless-window/#alternatives-on-macos) for macOS to hide the title bar but keep the window controls (traffic lights). Not sure if that will also be available for Windows/Linux at some point.

I guess currently the only way would be to go completely frameless and then re-create the title bar with the window controls that can be themed. Maybe ok for Windows. For Linux with all the different flavors it would probably be a bigger effort.

tj:
It would be awesome to optionally completely remove them, I think most people know how to close a window and minimize etc via shortcuts.

Yajo:
On 🐧 this can help a lot and it seems very simple: https://eischmann.wordpress.com/2017/01/10/dark-title-bars-for-apps-with-dark-ui/

Quoting:

> They just need to set an x window property GTK_THEME_VARIANT=dark

JLCCI:
+1

xtoolkit:
hi :)
i customize atom!
[...]

PR #13616, opened 2017-01-14
More title-bar options (macOS)

simurai:
### Description of the Change

This replaces the current "Use Custom Title Bar" checkbox with a "Title Bar" dropdown. It has the following options:

- `native` (default)
- `custom` adapts to themes
- `custom-inset`  **New**! has more padding
- `hidden` **New**! completely hidden

screen shot 2017-01-14 at 4 30 55 pm

#### Native

title-bar

#### Custom

title-bar 2

#### Custom Inset (new)

title-bar 3

#### Hidden (new)

title-bar 4

### Alternate Designs

I also tried to add a "hide title-bar but keep the traffic lights" option:

title-bar-custom

Unfortunately there are a few [edge cases](https://github.com/atom/atom/pull/10208#issuecomment-248550174) that would make the UI look broken. It's something that a [package](https://atom.io/packages/no-title-bar), theme or `styles.less` could provide when `custom-inset` is used.

Another example with the `tool-bar` package taking the role of a "title bar":

title-bar-custom 2

Needs access to the `BrowserWindow` API.

### Benefits

- Provides more title-bar variations.
- Themes can adapt to the different title-bar modes. Before there was no way to tell.
- Allows packages/themes to customize even further, like hide the title-bar but keep the traffic lights.

### Possible Drawbacks

When hiding the title-bar, managing the window needs to be done with the keyboard or from the menu. Dragging the window needs to be done with the status-bar. Not really a drawback since it's a choice and not the default.

### Applicable Issues

This PR contiues from https://github.com/atom/atom/pull/10208 (Hide title bar on OS X) and https://github.com/atom/atom/pull/11790 (already merged).

/cc @atom/feedback @atom/design


net:
> When hiding the title-bar, managing the window needs to be done with the keyboard or from the menu. Dragging the window needs to be done with the status-bar (or tree-view and tab-bar, but those could be hidden too). Not really a drawback since it's a choice and not the default.

Not entirely true. To move a window on macOS without a title bar, you can hover over the edge of the window so the resize arrow appears, then drag _perpendicular_ to the direction of the resize arrow. For example, when hovering over the left or right edge, drag vertically instead of horizontally. (This also works for windows with a title bar, of course.)

p-e-w:
Since the traffic lights are gone now, it appears that the `hidden` option has nothing left that makes it macOS-specific – so why not enable that option for Linux and Windows as well?

If the only concern is rendering of the menu bar, I'd be more than happy with the menu bar being completely disabled in `hidden` mode (I don't use it anyway – the command palette is enough) if only it meant I could get rid of the eyesore that the light-colored titlebar is in GNOME, and I doubt I'm the only one who feels that way.

simurai:
@net Wow.. :scream_cat: I didn't know that. That's actually really clever. Had to gif it:

dragging

simurai:
> @p-e-w I doubt I'm the only one who feels that way.

Not the only one: https://github.com/atom/atom/issues/13249. And I agree too. But maybe better to create a new PR for Windows/Linux.

tyrollins:
Would it be possible to add "close window" option to context menu when title bar is "hidden"?

simurai:
@tyrollins It should be possible, although that would clutter up the context menu with something that isn't directly related to where your mouse is and more global.

thomasjo:
@tyrollins Possible yes, but as alluded to by @simurai, such an option does not belong there. Adding such an option would then also necessitate adding e.g. *Save*, *Save As…*, *Quit*, *Close Tab*, etc.

tyrollins:
@simurai @thomasjo Thanks for the response, your comments convinced me. I was a bit to prompt with the previous comment. Very keen to see this implemented. Great work!

j-f1:
An updates on this?

saschaeggi:
@simurai how can I get this new feature? :)

simurai:
@saschaeggi this will be in Atom `1.16`. Beta should be out soon-ish.

saschaeggi:
@simurai great news, cheers from Zurich :)

jdsimcoe:
@simurai In my `zero-title-bar` plugin I was able to use the empty space to the right of the tab-bar area to allow dragging:

https://github.com/jdsimcoe/zero-title-bar

I simply set `.tab-bar .tab` to `-webkit-app-region: no-drag;`. This ensured that tabs retained their default behavior. It's worth considering adding support for this when `hidden` mode is activated.

lexicalunit:
@jdsimcoe, it seems like your plugin is a fork with a small modification of my [`no-title-bar`](https://github.com/lexicalunit/no-title-bar) package.

My copyright is even in [your repository's commit history](https://github.com/jdsimcoe/zero-title-bar/commit/3694b9ec8013fe8d1e2a9909cd753728b35a1c9c#diff-37854d19817c792316d481f5beb93cc7).

I am perfectly fine with you or anyone copying, modifying, and distributing my original codebase. However I am not sure that you've satisfied the following provision of the MIT license in the case of `zero-title-bar`:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Considering a substantial portion of your package is my code, if you wouldn't mind please updating your repository's LICENSE and README to reflect as much, I would be grateful!

You might also be interested in the recent PR that author @thfrlw submitted to my package which provides the same functionality of being able to drag the window using the empty space to the right of non-full-width tab-bar. It does so in a way that does not force a tab height as your package does.

**`zero-title-bar`:**

zero

See how the tab hight has been incorrectly adjusted to be wider than intended by the theme I am using?

**`no-title-bar`:**

no

With @thfrlw's solution, I do not have this problem. Perhaps your modification would best be implemented within `no-title-bar`. I would be happy to accept a PR or respond to an issue you are having with my package. Cheers!

jdsimcoe:
@lexicalunit I meant you know direspect, so sorry for lousing that up. I adjusted my README to reflect that my plugin is based on your `no-title-bar` plugin:

https://github.com/jdsimcoe/zero-title-bar

I adjusted the license and put my name in there because I wasn't sure how I should probably document the LICENSE portion but will change it any way that satisfies your request.

lexicalunit:
Thank you @jdsimcoe! I assumed you had no ill-intention. I think diversity in open source is an asset and look forward to seeing where you take `zero-title-bar` if you choose to maintain it and diverge from my package in more and potentially exciting ways.

I have also forked others' source code in the past and in one particular instance I modified, over time, their source to the point where it was nearly an entirely different codebase of my own writing. Yet I maintained some prominent documentation about the history of the source code though in good faith.

Thank you again for the quick turn around on this issue! 🎉

jdsimcoe:
@lexicalunit 👍  Yeah I'm going to dive into the new beta features and explore monkeying with `custom-inset` styles per @simurai's comments above:

screenshot

lexicalunit:
Awesome! Yeah this PR has some big implications for both of our packages. I have been meaning to take some time to explore different options for the direction of my package in the new Atom beta using the new `hidden` custom title bar setting. Thank you @simurai for this feature, I know I'm going to love it!

simurai:
> @jdsimcoe In my zero-title-bar plugin I was able to use the empty space to the right of the tab-bar area to allow dragging
> @lexicalunit being able to drag the window using the empty space to the right of non-full-width tab-bars.

Yeah, was considering that but then decided against it. The reason is that you can't rely on the empty space because it keeps changing. If your muscle memory is trained to move towards the tab bar, it might be frustrating to realized that the tab bar is full and now you have to move your mouse down to the status bar. Probably better to let themes/packages enable it if the tab bar has some padding around the tabs. Or add some padding on the right (maybe with a border, icon on hover) so that you always know there is a place to drag.

Anyways, with this PR there are now [3 CSS classes](https://github.com/atom/atom/blob/master/static/title-bar.less#L43-L87) that you could use to customize the tabs or other things:

1. `.custom-title-bar` 👉  traffic-lights height: `22px`
2. `.custom-inset-title-bar` 👉  traffic-lights height: `38px`
3. `.hidden-title-bar` 👉  no traffic-lights

jdsimcoe:
Fabulous, @simurai! 👏  Thanks for the color!

eboracus:
For the ability to still drag the window with little or no (or unreliable) empty space available, would it be a good suggestion to introduce a hotkey to change dragging a tab from `move tab position` to `move entire window position`.

So for example click and drag on any tab and it behave as it does now and moves the relative position of that particular tab, but hold down say `alt` + click and drag moves the whole window. Similar to alt + click anywhere on window to drag it on some platforms.

Having this off by default and toggle-able would allow the user to know what is needed to move the window, as opposed to introducing a 'hidden' method.

lexicalunit:
@eboracus that actually sounds like a neat idea! I may take a crack at it this weekend.

j-f1:
@lexicalunit You could also toggle the visibility of the tab bar to show a draggable area.

thfrlw:
Managed to get this, using One light theme, hidden title bar, flex-tool-bar package and some custom stylesheets.
screen-shot-2017-03-14-at-15 03 21

simurai:
@thfrlw Hehe.. probably a little against Apple's HIG, but clever hack. :laughing:

Do the custom traffic lights actually work, or would they still need to be wired up?

thfrlw:
It does work, here is a snippet from toolbar.cson, it explains how.
`
	{
		"type": "button",
		"callback": "window:close",
		"icon": "x",
		"className": "window-ctrl, red"
	},
	{
		"type": "button",
		"callback": "application:minimize",
		"icon": "dash",
		"className": "window-ctrl, green"
	},
	{
		"type": "button",
		"callback": "window:toggle-full-screen",
		"icon": "screen-full",
		"className": "window-ctrl, yellow"
	},`

lee-dohm:
Since this PR is merged and this discussion is not directly regarding the PR itself, could we move it to [Discuss](https://discuss.atom.io)? It will also make these suggested enhancements more visible to the community, I think :grinning:

jdsimcoe:
👍

Wolfr:
Haha @thfrlw that is awesome :)

xtoolkit:
hi
my customize for hide titlebar: https://github.com/xtoolkit/arshen-ac/pull/1

danielbayley:
@thfrlw Neat, mind sharing your styles? I kind of got it working, but the behaviour doesn't _quite_ match the native controls…

thfrlw:
@danielbayley Thanks!
I actually stoped using custom buttons & use native buttons now with styling around it.
You should check this gist. It might have something useful.
https://gist.github.com/thfrlw/8a89b1c93ee2bf27ebedcbbdc5ce5ade

PR #13629, opened 2017-01-15
Souda

biesmard:
[git-lfs-darwin-386-1.5.5.tar.gz](https://github.com/atom/atom/files/706798/git-lfs-darwin-386-1.5.5.tar.gz)

### Requirements


biesmard:
Style so that we can: display a summary of the failure of all tests throughout the test.

PR #12916, opened 2016-10-08
optimize png images using zopflipng

PeterDaveHello:
Use Google's zopflipng to re-compress the png images losslessly, 27 images optimized as below:

```
 resources/app-icons/beta/png/1024.png   | Bin 643649 -> 363244 bytes
 resources/app-icons/beta/png/128.png    | Bin 20327 -> 15826 bytes
 resources/app-icons/beta/png/16.png     | Bin 907 -> 838 bytes
 resources/app-icons/beta/png/24.png     | Bin 1582 -> 1432 bytes
 resources/app-icons/beta/png/256.png    | Bin 61483 -> 42817 bytes
 resources/app-icons/beta/png/32.png     | Bin 2388 -> 2148 bytes
 resources/app-icons/beta/png/48.png     | Bin 4467 -> 3843 bytes
 resources/app-icons/beta/png/512.png    | Bin 196695 -> 121414 bytes
 resources/app-icons/beta/png/64.png     | Bin 6981 -> 5950 bytes
 resources/app-icons/dev/png/1024.png    | Bin 615195 -> 332728 bytes
 resources/app-icons/dev/png/128.png     | Bin 19799 -> 15386 bytes
 resources/app-icons/dev/png/16.png      | Bin 905 -> 809 bytes
 resources/app-icons/dev/png/24.png      | Bin 1573 -> 1433 bytes
 resources/app-icons/dev/png/256.png     | Bin 59627 -> 41035 bytes
 resources/app-icons/dev/png/32.png      | Bin 2400 -> 2156 bytes
 resources/app-icons/dev/png/48.png      | Bin 4258 -> 3664 bytes
 resources/app-icons/dev/png/512.png     | Bin 188763 -> 113909 bytes
 resources/app-icons/dev/png/64.png      | Bin 6841 -> 5848 bytes
 resources/app-icons/stable/png/1024.png | Bin 643183 -> 365636 bytes
 resources/app-icons/stable/png/128.png  | Bin 20680 -> 16281 bytes
 resources/app-icons/stable/png/16.png   | Bin 944 -> 851 bytes
 resources/app-icons/stable/png/24.png   | Bin 1647 -> 1485 bytes
 resources/app-icons/stable/png/256.png  | Bin 62800 -> 43741 bytes
 resources/app-icons/stable/png/32.png   | Bin 2486 -> 2230 bytes
 resources/app-icons/stable/png/48.png   | Bin 4610 -> 3966 bytes
[...]

PR #13634, opened 2017-01-16
Losslessly optimize all images with imgult.

ryanpcmcquen:
https://imgult.github.io

Signed-off-by: Ryan P.C. McQuen <ryanpcmcquen@member.fsf.org>

### Description of the Change

All images were losslessly optimized with [imgult](https://github.com/ryanpcmcquen/image-ultimator).

Compare the original size:
`355` KB => https://github.com/atom/atom/blob/master/resources/app-icons/beta/png/1024.png

before_imgult

To the _imgulted_ size:
`131` KB => https://github.com/ryanpcmcquen/atom/blob/master/resources/app-icons/beta/png/1024.png

after_imgult

It's part of the core.

### Benefits

Less disk space used for images that look identical!

### Possible Drawbacks

None are obvious.


damieng:
The contrast/gamma looked off to me - looks a lot flatter - so I ran it through a comparison too at https://huddle.github.io/Resemble.js/

image

MaximSokolov:
These images are already compressed by Google's zopflipng in #12916

ryanpcmcquen:
@MaximSokolov yet they were still able to be reduced to less than half of their current size using other tools (including, but not limited to `optipng` & `pngquant`). Also, the `gif`'s were not optimized at all ... as far as I can tell.

ryanpcmcquen:
@damieng is that subtle difference worth the significant difference in file size?

damieng:
If this were a web image perhaps not but it's part of a 100MB application. 100KB won't make any difference.

ryanpcmcquen:
@damieng I would argue that the difference between the images here is even more significant:

https://github.com/atom/atom/pull/12916/files

Yet that request was merged.

damieng:
Not sure why you think the images are different there - they all look the same to me. I ran the same comparison on the green 1024;

image

ryanpcmcquen:
The greens are vastly different. Just use any color grabbing tool to see the difference between the greens. The lower bottom portion has two very different greens:

Original:
#189660

Optimized:
#14A674

screen shot 2017-01-16 at 9 28 15 am

Also, the lower shadow is more pronounced on the original (to my eyes).

damieng:
Then I think if anything we should consider reverting that original compression PR rather than letting the icons get more and more faded with each attempt at further compression.

ryanpcmcquen:
Atom does have some intention of being a web app at some point? Considering the future, perhaps we should consider optimizing the images as much as possible?

That's my take at least, but it is not my call.

ryanpcmcquen:
For posterity, I `imgulted` the original images, the savings is the same without using `zopflipng`:

`imgult`:
imgulted_1024_green
Original:
original_1024_green
Zopflipng:
zopflipng_1024_green

ryanpcmcquen:
Maybe @simurai could weigh in?

simurai:
Hmm.. it seems imgult uses dithering for compression, similar to a gif?

Original | zopflipng | imgult
--- | --- | ---
original | zopflipng | imgult

It's probably fine on a website where smaller image sizes have high priority. But yeah, for Atom it's more like a bonus and not that crucial.

@ryanpcmcquen What browser and OS did you use to make a [that screenshot](https://github.com/atom/atom/pull/13634#issuecomment-272922681)? I can't reproduce the "washed out" look in Chrome on macOS for the [previous PR](https://github.com/atom/atom/commit/b91768a923958154a486937ad07fa8a4e2ddc660):

screen shot 2017-01-19 at 2 10 18 pm

Maybe it's because zopflipng stripped the color profile?

screen shot 2017-01-19 at 2 47 18 pm

Actually, these icons in PNG format **only** get used on Linux. So would be a good idea to test it there if it has any effect.

ryanpcmcquen:
@simurai, thanks for the response. I'm on GNU/Linux (FreeSlack 14.2) but can also reproduce on Mac using the 'Generic RGB Profile'.

simurai:
If I switch to "Generic RGB Profile" I can see the whole monitor change, but not the right icon.

ok, I would say we don't merge this one and consider reverting #12916 if more people can confirm the ☝️  [above issue](https://github.com/atom/atom/pull/13634#issuecomment-272922681). Otherwise keep the happy medium.

@ryanpcmcquen Thanks for the PR nonetheless. ❤️

ryanpcmcquen:
@simurai I would not call it a happy medium, the `zopflipng` version looks the worst of the 3 on my monitors.

simurai:
Made a new issue https://github.com/atom/atom/issues/13672 to ask for confirmation.

HUN73R13:
I feel like using a flat solid color would look and optimize better, I know beaty is a matter of taste, but minimal icons also scale much better. color #34b17b seems to preserve the visual identity and look great. what do you guys think?

ryanpcmcquen:
For reference, here is what those look like _imgulted_:

2b427096-e15d-11e6-90b2-f786273d2765
c08998a6-e15c-11e6-8966-6ae608e48464

Each of them are about ~`53` KB!

simurai:
Yeah, a more flat icon would compress much better, but as already said, for Atom's use case, it's not the highest priority.

We'll definitely redesign the icon with the next major version of Atom, although there aren't currently any plans when that will happen.

Issue #13318, opened 2016-11-23
Unsaved buffer state isn't silently preserved when the Window has no project folder


iolsen:
Unsaved buffer state is preserved as part of the project system. When there is no project, it doesn't get saved. Prior to [this fix](https://github.com/stereobooster/atom/commit/c6a89a4abe560c22758ea3c755dacc0fc4368039) this behavior could result in data loss. That's now fixed, but seamlessly preserving state as Atom does for projects would be more convenient and is the more elegant fix.

We would readily accept a well-tested PR providing this functionality.

See also #10474.

rreinhardt:
I would like to suggest the addition of the label "data-loss" to this issue. 

Having followed along from #10474, I think it's worth commenting that preserving the unsaved buffer state is important for non-project documents because of situations where Atom is closed without having the time/ability to offer up the save prompt to the user. 

A couple cases that have resulted in data loss for me in the past:

1. When the Computer loses power (laptop runs out of battery).

2. When the Application/System freezes so that the process has to be killed.

---

I also think there is greater usability for the application to constantly be at a saved state than have the user make a decision(s) at the time of closing the application.

hydrandt:
I currently use Atom 1.8.0 on debian unstable, and the session/buffer saving has been working fine for quite a while, even for two atom windows in two different KDE activities.
Then Firefox used Atom to open some file I downloaded and all the sessions (and unsaved buffers) are lost.
Where does Atom save the sessions and unsaved files content? Is there a way to get the sessions back?

50Wliu:
Atom 1.13.0 features a new Reopen Project feature that you can use to return to your old projects.  Otherwise you would have to reopen the folder/folders that you were editing.

Curros:
Wow I just expend 1h looking to all the mess around this "bug".
I'm using W10x64 - Atom 1.12, and i have the same issue.

@50Wliu if the file is not saved, you lost it even using Ctrl+Shift+T

Curros:
I just instal Atom 1.13 beta and works fine! :dancer: 
Using W10x64 - Atom 1.13

I can finally switch from sublime :D

ctwise:
Nope, Atom 1.13 beta still isn't working. It requires you to save the buffers. Sublime doesn't. That makes it _perfect_ for scratch pads, open a new tab and start typing and pasting. Close Sublime, open it back up and the untitled buffers are still there.

50Wliu:
[...]

PR #13644, opened 2017-01-18
Check whether state store is connected when prompting to save files

iolsen:
Because we rely on the state-store's availability to save unsaved changes in files it seems important to be able to check for the case where we never connected to the state store. This would be common for anybody who has multiple versions of Atom installed. Normal users probably don't do this often but as an Atom developer I certainly do. So this PR makes two changes:

1. Introduces an `isConnected` flag to `state-store`.

In the common (for me) case where you open e.g. beta after you've opened stable, and beta fails to connect to the indexedDB, it's set false.

2. Checks `atom.stateStore.isConnected` when determining whether to prompt to save file changes. Without this, changes are silently lost in this scenario.

Related to https://github.com/atom/tabs/issues/306 and https://github.com/atom/atom/issues/13318.

/cc @nathansobo @as-cii for review


src/text-editor.coffee
```diff
@@ -892,7 +892,7 @@ class TextEditor extends Model
   # Determine whether the user should be prompted to save before closing
   # this editor.
   shouldPromptToSave: ({windowCloseRequested, projectHasPaths}={}) ->
-    if windowCloseRequested and projectHasPaths
+    if windowCloseRequested and projectHasPaths and atom.stateStore.isConnected
```
nathansobo: We've been trying to avoid `atom` global references in `TextEditor` to facilitate extraction when the time comes. Could you inject the state store as a dependency or find some way to avoid this global reference? I'm not sure where we reference it downstream from this but if that's also a global reference, then maybe ignore my advice.

src/state-store.js
```diff
@@ -10,10 +10,12 @@ class StateStore {
         db.createObjectStore('states')
       }
       dbOpenRequest.onsuccess = () => {
+        this.isConnected = true
```
nathansobo: Kinda weird but we have a tradition of only ever accessing internal object state via methods, which would imply `isConnected` should be a method. We could consider changing that because ES6 has a convenient facility for defining getters, but originally it was to avoid confusion about which properties were public because of the lack of privacy controls in JS.

nathansobo (changes requested):
Totally approve of the goal but I have a couple comments.

src/text-editor.coffee
iolsen: OK the only other reference I could find is on line 821, which is actually a global reference:

```
for textEditor in atom.workspace.getTextEditors() when textEditor isnt this
```

So I'm ignoring this. 😈

nathansobo (approved):

PR #13646, opened 2017-01-18
Only sign and create-windows-installer on *-releases branches

damieng:


as-cii:
Albeit slower, I think it might actually be nice to keep producing installers and binaries that look like the ones that we release. As far as I understand, an unsigned application hasn't the same privileges of a signed one, so this might make testing a feature (or a change) harder for @ungb.

@damieng: do you have any additional concerns that would be addressed with this pull request, other than speed?

@ungb: thoughts? 💭

damieng:
We can still sign the atom.exe binary - it's just that if you want to download and try it you'd need to unpack a zip instead of running the installer.

I think this is a safer option as the installer is not side-by-side and will 'upgrade' stable and beta builds on your box to the dev build.

damieng:
The other thing to note is that creating the installer and nuget packages is about 30% of the build time on appveyor.

as-cii:
> We can still sign the atom.exe binary - it's just that if you want to download and try it you'd need to unpack a zip instead of running the installer.

> The other thing to note is that creating the installer and nuget packages is about 30% of the build time on appveyor.

I'm 👍 on signing the binary but not creating an installer.

ungb:
Usually when I test things on master I build locally which doesn't signs my build so I don't think this change wouldn't impact testing that much for specific features, but any changes with sign vs unsigned might be a bit difficult to test, but if the time comes for those changes we can figure out a way to test it.

damieng:
I'll add the signing code today and get this merged in.

damieng:
Found other ways to keep AppVeyor queue down. Re-visit if it becomes a problem again.

PR #13657, opened 2017-01-19
Add option for closing deleted file tabs

### LINKED ISSUE #306: Provide option to allow keeping open deleted tabs
I have some testing code that regenerates a directory structure each run. I would like to keep a tab open to observe the changes to the file when it is regenerated.

This make its difficult to check changes from generated files, such as transpilation output, etc.

After #160, the tab is removed, even though it is regenerated again within seconds, before the window gets focus again.

Sublime Text keeps them open, and only updates on focus.

I think there should be an option to allow only refreshing/deleting tabs on window focus.

iolsen:
Adds a core option for keeping open editors whose underlying files are deleted outside Atom.

screen shot 2017-01-18 at 4 40 27 pm

To be merged in tandem with atom/text-buffer#178 (which will include tests for this option) and atom/tree-view#1026 in order to close atom/tabs#306.

Note that the default will change to keep these tabs open, as that's the conservative behavior with respect to potentially losing data.

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -68,6 +68,12 @@ const configSchema = {
         default: true,
         description: 'Trigger the system\'s beep sound when certain actions cannot be executed or there are no results.'
       },
+      closeDeletedFileTabs: {
+        type: 'boolean',
+        default: false,
+        title: 'Close Deleted File Tabs',
+        description: 'Close corresponding editors when a file is deleted outside Atom.'
+      },
       destroyEmptyPanes: {
         type: 'boolean',
         default: true,
--- a/src/project.coffee
+++ b/src/project.coffee
@@ -360,9 +360,13 @@ class Project extends Model
     else
       @buildBuffer(absoluteFilePath)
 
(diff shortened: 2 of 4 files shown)


adamreisnz:
@iolsen with this option disabled, what will be the behaviour when you delete a file from *within Atom*, e.g. using tree view?

The issue was originally that this resulted in orphaned tabs that needed to be manually closed, and which triggered a confirmation dialog because the buffer wasn't saved.

I'm glad we are getting a toggle now, but I hope the original issue is also addressed, and that removing files form within Atom (which comes with it's own confirmation), *will* in fact close the tabs automatically.

iolsen:
As part of this change, deletes from the tree view will close corresponding tabs regardless of the setting: https://github.com/atom/tree-view/pull/1026.

adamreisnz:
Excellent, thanks

deckar01:
It would be nice if the next major release would turn on config options like this by default. I spend a fair amount of time every week debugging and researching atom bugs only to find out it has been fixed by a new option that is disabled by default.

PR #13664, opened 2017-01-20
Add showCursorOnSelection config

zertosh:
### Description of the Change

Adds a new config `editor.showCursorOnSelection`, enabled by default (changing the current default behavior), that makes it so that the cursor is always blinking even when you have a selection(s).

It's pretty tied to the internals of `Cursor` and `TextEditor`.

### Benefits

Brings Atom in line with the behavior of Sublime (and others).

This is super useful for when you do `cmd+d` (`find-and-replace:select-next`) and you can't find where your cursor went.

### Possible Drawbacks

The current behavior is to hide the cursor when there's a selection. Since we're changing the default, that's sure to upset someone. However, since it is configurable, those that don't like it can easily revert to the old behavior.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1347,7 +1347,19 @@ describe('TextEditorComponent', function () {
       expect(cursorsNode.classList.contains('blink-off')).toBe(true)
     })
 
-    it('does not render cursors that are associated with non-empty selections', function () {
+    it('renders cursors that are associated with empty selections', function () {
+      editor.update({showCursorOnSelection: true})
+      editor.setSelectedScreenRange([[0, 4], [4, 6]])
+      editor.addCursorAtScreenPosition([6, 8])
+      runAnimationFrames()
+      let cursorNodes = componentNode.querySelectorAll('.cursor')
+      expect(cursorNodes.length).toBe(2)
+      expect(cursorNodes[0].style['-webkit-transform']).toBe('translate(' + (Math.round(6 * charWidth)) + 'px, ' + (4 * lineHeightInPixels) + 'px)')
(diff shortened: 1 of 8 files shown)


zertosh:
To see what this behavior is like, you can quickly monkeypatch `Cursor.prototype.updateVisibility` with:

```js
require(process.resourcesPath + '/app.asar/src/cursor.js')
  .prototype.updateVisibility = function() {
    this.setVisible(true);
  };
```

zertosh:
Test failures are unrelated. See https://github.com/atom/atom/commit/d50b56b6a1ed07ad407f8cc5072af6c86fe4b61d#commitcomment-20549225

lee-dohm:
Has any testing been done for when people change the appearance of their cursor? Examples:


**Prevent cursor from blinking**

```
atom-text-editor {
  .cursors.blink-off .cursor {
    opacity: 1;
  }
}
```

**Make cursor pulse instead of blink**

```
atom-text-editor {
  .cursor {
    transition: opacity 0.5s;
  }
}
```

zertosh:
@lee-dohm, just tried both and they work just fine.

The change for the new behavior is just 4 lines (see https://github.com/atom/atom/pull/13664/files#diff-24f2fe2bd60cb4194c3cf7d107fa0f99L578). The bulk of this PR is tests and making it a setting that propagates correctly.

lee-dohm:
Awesome, thanks @zertosh :bow:

nathansobo:
Looks super well tested and clean. ✨

iolsen:
blinky-cursor

nicolasrodriguezq:
What do you call the three vertical lines you see in the catch?

50Wliu:
Indent Guide: Settings -> Editor -> Show Indent Guide

PR #13668, opened 2017-01-20
Don't transform deprecated selectors containing malformed CSS

### LINKED ISSUE #13663: 1.13 Regression: Font Family setting causing CssSyntaxErrors
### Prerequisites

Note: I am only able to reproduce some of the issues.

### Description

Starting with the release of Atom 1.13.0, we have received [a](https://discuss.atom.io/t/atom-wont-start-up/38354) [number](https://discuss.atom.io/t/atom-1-13-update-non-functional/38189/10) [of](https://github.com/atom/atom/issues/13586) [bug](https://github.com/atom/settings-view/issues/900) [reports](https://github.com/atom/atom/issues/13631) about the Font Family setting causing a `CssSyntaxError` to be thrown.  These errors did not appear in versions prior to 1.13.0, and while they may be correct, should at the very least be displayed in a more "pleasing" manner.

### Steps to Reproduce

1. Enter a quote or some otherwise malformed CSS into the Font Family box.  Some reports also suggest that a semicolon can cause the error, but I couldn't reproduce it.

**Expected behavior:** Nothing

**Actual behavior:** CssSyntaxError

**Reproduces how often:** 100% for quotes or otherwise invalid CSS, 0% for semicolons

### Versions

Not relevant, this is a 1.13 regression.  Appears to happen on all OSes.

as-cii:
Fixes #13663.

This pull request fixes an error that was thrown by `postcss` when trying to upgrade deprecated selectors of a file containing malformed CSS. With these changes all the CSS files that contain an invalid syntax will be skipped and the automatic transformation won't be applied.

/cc: @atom/core

### DIFF
--- a/src/style-manager.js
+++ b/src/style-manager.js
@@ -250,59 +250,70 @@ module.exports = class StyleManager {
 
 function transformDeprecatedShadowDOMSelectors (css, context) {
   const transformedSelectors = []
-  const transformedSource = postcss.parse(css)
-  transformedSource.walkRules((rule) => {
-    const transformedSelector = selectorParser((selectors) => {
-      selectors.each((selector) => {
-        const firstNode = selector.nodes[0]
-        if (context === 'atom-text-editor' && firstNode.type === 'pseudo' && firstNode.value === ':host') {
-          const atomTextEditorElementNode = selectorParser.tag({value: 'atom-text-editor'})
-          firstNode.replaceWith(atomTextEditorElementNode)
-        }
-
-        let previousNodeIsAtomTextEditor = false
-        let targetsAtomTextEditorShadow = context === 'atom-text-editor'
-        let previousNode
-        selector.each((node) => {
-          if (targetsAtomTextEditorShadow && node.type === 'class') {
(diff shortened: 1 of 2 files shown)


src/style-manager.js
```diff
@@ -250,59 +250,70 @@ module.exports = class StyleManager {
...
-  transformedSource.walkRules((rule) => {
-    const transformedSelector = selectorParser((selectors) => {
-      selectors.each((selector) => {
-        const firstNode = selector.nodes[0]
-        if (context === 'atom-text-editor' && firstNode.type === 'pseudo' && firstNode.value === ':host') {
-          const atomTextEditorElementNode = selectorParser.tag({value: 'atom-text-editor'})
-          firstNode.replaceWith(atomTextEditorElementNode)
-        }
+  let transformedSource
+  try {
```
as-cii: We could wrap the whole function inside a try/catch but that would prevent v8 from optimizing it out. Given that we run this code during startup, I think that narrowing the scope of the unoptimized code that gets executed is a good idea.

PR #13691, opened 2017-01-25
Allow building on Node 7+

winstliu:
### Requirements

### Description of the Change

Removes a restriction that prevented people from trying to build Atom using Node 7+.  Now that apm has been updated this should work again.
In addition, I fixed a potential bug where `stdout` was being declared as a constant but then being reassigned.  I'm not sure this would work; even if it did, it's best that it's rewritten to be a `let` instead.

### Alternate Designs

None.

Already in core.

### Benefits

More supported Node versions when building.

### Possible Drawbacks

None.

### Applicable Issues

None.

If CI is green, I will revert the CI changes and :boat:.


50Wliu:
Turns out that the linting script doesn't lint the `script` folder!  There were a bunch of warnings in there, some of which become full-blown errors when running under Node 7 and improved ES6 support.  So I've added that to the list of files to lint, and will lint all those files as a part of this PR.

I also noticed that specs written in JavaScript aren't being linted, but in the interest of keeping this PR small, I'll be opening a different PR for that.

50Wliu:
Ok, builds are green.  Though this PR turned out larger-than-expected, and I want to point out some stuff that I'm not sure I did right for review.

script/lib/code-sign-on-mac.js
```diff
@@ -10,8 +10,9 @@ module.exports = function (packagedAppPath) {
     return
   }
 
+  // This declaration is outside of the try block because it is also used for the finally block
+  const certPath = path.join(os.tmpdir(), 'mac.p12')
```
50Wliu: I'm reasonably sure this shouldn't ever throw as long as `os.tmpdir()` works.

script/lib/generate-metadata.js
```diff
@@ -27,8 +27,8 @@ function buildBundledPackagesMetadata () {
     const packagePath = path.join(CONFIG.intermediateAppPath, 'node_modules', packageName)
     const packageMetadataPath = path.join(packagePath, 'package.json')
     const packageMetadata = JSON.parse(fs.readFileSync(packageMetadataPath, 'utf8'))
-    normalizePackageData(packageMetadata, () => {
-      throw new Error(`Invalid package metadata. ${metadata.name}: ${msg}`)
+    normalizePackageData(packageMetadata, (msg) => {
+      throw new Error(`Invalid package metadata. ${packageName}: ${msg}`)
```
50Wliu: `metadata.name` doesn't exist, but `packageName` does.  I'm not sure the two are equivalent without knowing what `metadata.name` was supposed to refer to.

script/lib/generate-metadata.js
```diff
@@ -118,7 +118,7 @@ function checkDeprecatedPackagesMetadata () {
   for (let packageName of Object.keys(deprecatedPackagesMetadata)) {
     const packageMetadata = deprecatedPackagesMetadata[packageName]
     if (packageMetadata.version && !semver.validRange(packageMetadata.version)) {
-      throw new Error(`Invalid range: ${version} (${name}).`)
+      throw new Error(`Invalid range: ${packageMetadata.version} (${packageName}).`)
```
50Wliu: Ditto here.

script/lib/transpile-babel-paths.js
```diff
@@ -6,15 +6,6 @@ const glob = require('glob')
...
 const CONFIG = require('../config')
-const BABEL_OPTIONS = require('../../static/babelrc.json')
-const BABEL_PREFIXES = [
-  "'use babel'",
-  '"use babel"',
-  '/** @babel */',
-  '/* @flow */'
-]
-const PREFIX_LENGTH = Math.max.apply(null, BABEL_PREFIXES.map(prefix => prefix.length))
-const BUFFER = Buffer(PREFIX_LENGTH)
```
50Wliu: All these are unused, but seem kinda important.  I hope I'm not missing anything here?

script/test
```diff
@@ -71,7 +71,7 @@ for (let packageName in CONFIG.appMetadata.packageDependencies) {
     console.log(`Executing ${packageName} tests`.bold.green)
     const cp = childProcess.spawn(executablePath, testArguments)
     let stderrOutput = ''
-    cp.stderr.on('data', data => stderrOutput += data)
+    cp.stderr.on('data', data => (stderrOutput += data))
```
50Wliu: Little "hack" to get around standard complaint.

spec/menu-manager-spec.coffee
```diff
@@ -62,7 +62,7 @@ describe "MenuManager", ->
       atom.keymaps.add 'test', 'atom-workspace': 'ctrl-b': 'b'
       menu.update()
 
-      waits 1
+      waits 50
```
50Wliu: Appveyor failed twice on this spec so I bumped the waits to be consistent with the other tests in this file and now it passes.

50Wliu:
Note: While running `script/build` using Node 7 now completes successfully, that _does not_ mean that Atom behaves as expected when built using Node 7, nor is Node 7+ supported.

torn4dom4n:
Why don't you use Node 6 for LTS?

50Wliu:
I'm not planning to use Node 7 :).  This PR just _allows_ for Atom to be built in Nice 7 instead of erroring.  I will be reverting the CI changes before merging.

script/lib/transpile-babel-paths.js
BinaryMuse: Looks like they're here now: https://github.com/atom/atom/blob/e4f2c80eb5f124012524297622f20c540bbe3822/src/babel.js

BinaryMuse:
@50Wliu I plan on taking a closer look at this soon. Are you okay with me resolving the merge conflicts?

50Wliu:
Yeah, of course.

script/test
thomasjo: Am I missing something here? Why not use curly brackets here? Surely the complaint is due to the assignment and return of the original statement?

50Wliu: Yes, you're right.  I've been using curly brackets in other places to fix the warning but not sure why I didn't here.

adamreisnz:
Any progress on this? We will be getting Node 8 soon and at present Atom still can't build on Node 7.9.0.

BinaryMuse:
Atom now builds successfully on Node 7 and Node 8! Thanks @50Wliu for taking care of this.

The Travis test is currently not running due to issues on Travis' infrastructure, but I feel confident in the changes here and it passes with Node 7, so let's get this into master. 🎉

adamreisnz:
Great work, thanks! Just in time for Node 8 ;)

PR #13696, opened 2017-01-25
Allow code signing cert to be specified at a local path

BinaryMuse:
This allows code signing certificates to be specified via `ATOM_MAC_CODE_SIGNING_CERT_PATH` and `ATOM_WIN_CODE_SIGNING_CERT_PATH` on macOS and Windows, respectively. In that case, we don't try to download the certificate, and we also don't delete the certificate after signing is complete.

This also fixes an issue where `certPath` wasn't in scope for the `finally` block in the macOS case.

/cc @as-cii @damieng

### DIFF
--- a/script/lib/create-windows-installer.js
+++ b/script/lib/create-windows-installer.js
@@ -23,7 +23,7 @@ module.exports = function (packagedAppPath, codeSign) {
   }
 
   const signing = codeSign && (process.env.ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL || process.env.ATOM_WIN_CODE_SIGNING_CERT_PATH)
-  let certPath = ATOM_WIN_CODE_SIGNING_CERT_PATH;
+  let certPath = process.env.ATOM_WIN_CODE_SIGNING_CERT_PATH;
 
   if (signing) {
     if (!certPath) {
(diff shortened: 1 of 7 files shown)


script/lib/create-windows-installer.js
```diff
@@ -22,11 +22,14 @@ module.exports = function (packagedAppPath, codeSign) {
     setupIcon: path.join(CONFIG.repositoryRootPath, 'resources', 'app-icons', CONFIG.channel, 'atom.ico')
   }
 
-  const certPath = path.join(os.tmpdir(), 'win.p12')
-  const signing = codeSign && process.env.ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL
+  const signing = codeSign && (process.env.ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL || process.env.ATOM_WIN_CODE_SIGNING_CERT_PATH)
+  let certPath = ATOM_WIN_CODE_SIGNING_CERT_PATH;
+  if (!certPath) {
```
damieng: Should probably be `if (!certPath && codeSign)` otherwise it's going to download the key files on all builds including ones it wasn't going to sign.

damieng (commented):
Looks good!

as-cii (approved):
Looks good to me. Thanks, @BinaryMuse! ✨

PR #13698, opened 2017-01-25
Lint JavaScript specs

winstliu:
### Requirements

### Description of the Change

Lints specs written in JS.

### Alternate Designs

None.

Because everything else gets linted in core?

### Benefits

Better code quality.

### Possible Drawbacks

None.

### Applicable Issues

None.


50Wliu:
Not quite sure if the `eslint-disable-line no-unused-var` statements are the best way to handle the ffit stuff.

ericcornelissen:
Alternatively you could just removed the unused variables :wink: I would suggest one change tho, on [line 13 in `lint-java-script-paths.js`](https://github.com/atom/atom/blob/37f43fb0f5c669a66271029d65f198c83ba882e4/script/lib/lint-java-script-paths.js#L13) I would change `'main-process'` by `'**'`, and remove line 12. This way all tests in the ./spec folder and subfolders will be linted.

Furthermore, your changes seem alright, I don't see why you won't lint the test suites. 😃

50Wliu:
> This way all tests in the ./spec folder and subfolders will be linted.

We don't want to lint the fixtures folder though.

ericcornelissen:
> We don't want to lint the fixtures folder though.

In that case this seems like a good solution 😄

sadick254:
Thanks for opening this PR!

This has already been done on master. Now we also lint the spec files.

PR #13713, opened 2017-01-29
pass through line count options

dirk-thomas:
This is in support of atom/find-and-replace#847.

### Requirements

* atom/find-and-replace#847
* atom/scandal#41
* atom/text-buffer#195

### Description of the Change

Add two additional optional options to `workspace.scan` (namely `lineCountBefore` and `lineCountAfter`) and pass them along to the `scan-handler` and from there to the `PassSearcher`.

### Alternate Designs

I am not aware of an alternative approach.

Since the modified API for searching is part of the core it needs to be extended to support the use case described in atom/find-and-replace#190. If the `find-and-replace` package could handle this without requiring modifications to the core that would be great but with the current API being provided by the core I don't think this is possible.

### Benefits

It implements the long desired feature described in atom/find-and-replace#190.

### Possible Drawbacks

The additional optional argument shouldn't result in any backward compatibility problem as far as I can tell.

### Applicable Issues / Remaining Todos

* ~~In the current state when the search results list a file which is not opened yet and you click on the result the match is being updated and looses the context information. This is due to the fact that the search is being repeated in the buffer (instead of using `scandal` through the workspace API. In order to address this the `TextBuffer.scan` needs to be extended to support the same new options as in the referenced `scandal` ticket. Therefore I have prefixed the title with `[WIP]` for now.~~ This has been address with the PR to the `text-buffer` package.

* Add test coverage / update existing specs as soon as there is a consensus on the selected approach and the feature works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "random-words": "0.0.1",
     "resolve": "^1.1.6",
     "runas": "^3.1",
-    "scandal": "^3.0.0",
+    "scandal": "^3.1.0",
     "scoped-property-store": "^0.17.0",
     "scrollbar-style": "^3.2",
     "season": "^6.0.0",
--- a/spec/workspace-spec.js
+++ b/spec/workspace-spec.js
@@ -1336,7 +1336,9 @@ i = /test/; #FIXME\
       it('calls the callback with all regex results in all files in the project', () => {
         const results = []
         waitsForPromise(() =>
-          atom.workspace.scan(/(a)+/, result => results.push(result))
+          atom.workspace.scan(
+            /(a)+/, {leadingContextLineCount: 1, trailingContextLineCount: 1},
+            result => results.push(result))
         )
 
         runs(() => {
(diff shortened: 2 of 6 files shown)


dirk-thomas:
Since the beta of 1.15 was just released it would great if this extended API could be merged to become available in the next beta (1.16).

dirk-thomas:
I updated the patch to use the same option names as in https://github.com/atom/text-buffer/pull/195#discussion_r104802317.

50Wliu:
You'll need to rebase as workspace.coffee was changed to workspace.js recently.

dirk-thomas:
Rebased.

dirk-thomas:
Rebased again. CI passing.

dirk-thomas:
And rebased yet another time after #13940 was merged...

maxbrunsfeld:
Looks like the one appveyor failure is spurious. We'll have to take a 👀  at that test. Thanks @dirk-thomas!

dirk-thomas:
The failure seems to be unrelated to my changes (I know, every contributor claims that :wink:):

```
atom.themes
  ::requireStylesheet(path)
    it synchronously loads and parses less files at the given path and installs a style tag for it in the head
      Expected '#header {
  color: #4d926f;
}
h2 {
  color: #4d926f;
}
' to be '#header {
  color: #4D926F;
}
h2 {
  color: #4D926F;
}
'.
        at .<anonymous> (file:///C:/projects/atom/spec/theme-manager-spec.coffee:195:35)
```

@maxbrunsfeld Can you retrigger the AppVeyor build to make it pass? (Since I assume it is a flaky test bsaed on your comment.)

dirk-thomas:
Nvm already merged :tada: Thanks.

maxbrunsfeld:
I just meant we (the core Atom team) will have to take a look at that failure sometime soon; definitely not your responsibility.

PR #13731, opened 2017-02-02
:arrow_up: coffee-script

Ingramz:
Just a friendly reminder to occasionally update coffee-script.

Allows new syntax to be used and also provides better source maps.

After merging, make sure to clear compile-cache for updated source maps to take effect.

Changelogs can be found at http://coffeescript.org/#1.12.0 and then by scrolling up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "chai": "3.5.0",
     "chart.js": "^2.3.0",
     "clear-cut": "^2.0.2",
-    "coffee-script": "1.11.1",
+    "coffee-script": "1.12.7",
     "color": "^0.7.3",
     "dedent": "^0.6.0",
     "devtron": "1.3.0",


ericcornelissen:
Works fine on Windows 10 (x64)
- [X] Run tests successfully
- [X] Build
- [X] Use Atom after building

Ingramz:
I don't know what's wrong with CI but I've had the build running for a while now and diff against the parent build only shows differences in sourcemaps, so I would say that it is safe to merge this.

nathansobo:
Hey @Ingramz. Thanks for this.

Unfortunately, after watching the ecosystem evolve we have realized that never should have baked transpilation into Atom. If we upgrade the bundled CoffeeScript transpiler, there is a high probability of breaking a bunch of existing code.

Long term, the solution is for us to fix our package publishing system to allow transpiled assets to be published along with the release in addition to source files that are checked into Git. We haven't gotten there yet 😞.

In the short term, to use a newer CoffeeScript or any other language you want, you can bundle a custom transpiler with your package that will cache the transpiled sources the first time the package is started. This is great for development but it's unfortunate that it forces you to bundle the transpiler as a dependency of the package which increases its storage footprint and install time. If you can accept that trade-off, then the [github package is a good example of how a custom transpiler can be used](https://github.com/atom/github/blob/caaeeba4ab50035f08024e4b9ced72069900514a/package.json#L18-L31).

My apologies for this situation. Baking in transpilation was a short-sighted design decision I made early on in this project and we really need to do the work to allow artifacts to be published with packages, but haven't.

nathansobo:
Okay, after further discussion with the commit author and the rest of the team, I've changed my mind on this.

Bundled transpilation is not something we plan to stick with over the long term, but this minor upgrade seems to be low risk. If it breaks packages, which seems unlikely, we can lint for the offending construct across the ecosystem.

I just want to say right now that merging this upgrade is not a promise to merge future upgrades. We'll have to take it on a case-by-case basis and judge the risk in each scenario. Specifically, it's extremely unlikely that we'll be able to upgrade bundled transpilation to CoffeeScript 2.0.

Ingramz:
Part of my promise to @nathansobo during that discussion was to check for compilation output differences between the old and the new version across all the packages.

Compilation differences between 1.11.1 and 1.12.7 were found in following `.coffee` files using atom-package-downloader and `find . -name "*.coffee" -type f`. This resulted in 13242 files that needed to be checked. I crafted a script that ran `coffeescript.compile(contents)` for both versions and compared their outputs whether they are equal or not. Files that failed to compile with 1.11.1 were excluded.

Differences were found in following files:
```
atom-beautify/src/beautifiers/index.coffee
atom-beautify/src/beautifiers/sqlformat.coffee
atom-gibo/lib/atom-gibo.coffee
atom-latex/spec/main-spec.coffee
cubic-bezier/lib/cubic-bezier-curve.coffee
ever-notedown/lib/markdown-helper.coffee
file-header/lib/file-header.coffee
find-and-replace-fix/lib/project/results-view.coffee
forcedotcom-builder/lib/build.coffee
git-control/lib/dialogs/flow-dialog.coffee
git-control/lib/views/file-view.coffee
git-controlhub/lib/dialogs/flow-dialog.coffee
greppen-grok/lib/greppen-grok-model.coffee
imdone-atom-github/lib/plugin.coffee
quick-editor/lib/quick-editor.coffee
rot13/lib/rot13.coffee
sf-tools/lib/build.coffee
simple-git/lib/editor.coffee
super-duplicate/lib/super-duplicate.coffee
tnote/lib/tnote.coffee
ult-terminal/lib/term-view.coffee
you-complete-me/lib/get-issues.coffee
```

List is small enough that I can go through those by hand.

nathansobo:
Fabulous. Thanks for the quick follow-up.

Ingramz:
Differences are mostly that code which compiled to ternary expressions, the conditional is now surrounded by parentheses, which is harmless.

There are three files that have a different change, which need to be looked at case by case:
[atom-latex/spec/main-spec.coffee](https://github.com/Ingramz/curly-guacamole/commit/9013bd62796dfafb6dfae218417901514c185feb) ([original](https://github.com/Ingramz/curly-guacamole/blob/master/atom-latex/spec/main-spec.coffee))
[quick-editor/lib/quick-editor.coffee](https://github.com/Ingramz/curly-guacamole/commit/5e38c845c1d6aeac7e5bbcc2aa89f4c6fc8e7347) ([original](https://github.com/Ingramz/curly-guacamole/blob/master/quick-editor/lib/quick-editor.coffee))
[you-complete-me/lib/get-issues.coffee](https://github.com/Ingramz/curly-guacamole/commit/0488f71d50fb696c44bdd3e9551684189be3ec49) ([original](https://github.com/Ingramz/curly-guacamole/blob/master/you-complete-me/lib/get-issues.coffee))

GeoffreyBooth:
Please change your dependency to `coffeescript` rather than `coffee-script`. `coffeescript` is the primary module now for CoffeeScript.

Issue #9429, opened 2015-11-04
Can Atom have a custom right-click context menu?


jamen:
When right clicking in Atom, I can't help to notice that the popup is my system's sad attempt at a context menu.

I think it would be interesting if we used a custom right-click context menu in the document, or even a frameless `BrowserWindow`.  This way its more accessible as a piece of the software.  For things like themes and packages to customize and use.

programmatical:
+1

ricardovsilva:
+1 . If someone point me where to start, I can try develop this feature.

50Wliu:
> If someone point me where to start, I can try develop this feature.

First you'll probably need to see whether or not Electron supports custom context menus.  If it does, then it's just a matter of using those APIs within Atom.

Issue #7739, opened 2015-07-06
Add support for disabling context menu items


avida-michaud:
I don't see a way in the contextMenu.add API to dynamically disable/re-enable a menu item: https://atom.io/docs/api/v1.0.0/ContextMenuManager#instance-add. 

Can we get a way to disable (shown but not selectable) and enable context menu items? Something like .shouldEnable(event) - takes a function with event?

PR #13746, opened 2017-02-05
Add ContextMenuManager.remove

UziTech:
### Description of the Change

This will add a `remove` method to `atom.contextMenu` to counter the `add` method

With this method one will be able to remove items from the context menu by specifying a selector(s) and label(s)

Because the only alternative is to override a global object `atom.contextMenu`

### Benefits

This will finally give the ability to remove items that clutter the context menus from packages.
Similar to the benefits of disabling keybindings that packages automatically add

### Possible Drawbacks

None that I can think of.

### Applicable Issues

#9429
#7739

### DIFF
--- a/spec/context-menu-manager-spec.coffee
+++ b/spec/context-menu-manager-spec.coffee
@@ -230,3 +230,138 @@ describe "ContextMenuManager", ->
             }
           ]
         ])
+
+  describe "::remove(itemsBySelector)", ->
+    it "can remove top-level menu items that can be re-added with the returned disposable", ->
+      contextMenu.add
+        '.parent': [{label: 'A', command: 'a'}]
+        '.child': [{label: 'B', command: 'b'}]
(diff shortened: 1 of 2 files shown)


lee-dohm:
Thanks for contributing!

Unfortunately, we're not going to merge this for a few reasons:

1. There is already a way to remove added context menu items: using the `Disposable` returned by [`ContextMenuManager.add`](https://atom.io/docs/api/latest/ContextMenuManager#instance-add)
1. The above is the primary method of removing something that was added using the Atom API, not matching `remove` methods
1. There is no precedent for having a `Disposable` returned to re-add something that was removed, only the reverse

If disabling context menu items from packages is your goal, you should mirror the design of the system used in disabling keybindings from packages: https://github.com/atom/atom/pull/8130

Thanks for the hard work you put into this and the interest you have in Atom!

UziTech:
Ok thank you. I'll take a look at that pull request

PR #13748, opened 2017-02-05
Allow custom directory search providers to also override replace

hansonw:
This change allows providers of `atom.directory-searcher` to provide a custom `replace` method to override the default `scandal`-based replacement logic, and moves the existing replacement logic into the default `DirectorySearch` provider.

For example, in Nuclide, we provide a custom directory searcher to enable searching remote files. However, remote file replacement is naturally still broken.

As a temporary workaround we ended up monkey-patching `atom.workspace.replace` (cc @zertosh): https://github.com/facebook/nuclide/blob/master/pkg/nuclide-remote-projects/lib/patchAtomWorkspaceReplace.js#L41

This will allow us to remove the hack and add the necessary code to our custom remote directory search provider instead.

Released under CC0.

### DIFF
--- a/spec/default-directory-searcher-spec.coffee
+++ b/spec/default-directory-searcher-spec.coffee
@@ -1,6 +1,8 @@
 DefaultDirectorySearcher = require '../src/default-directory-searcher'
 Task = require '../src/task'
+fs = require 'fs-plus'
 path = require 'path'
+temp = require('temp').track()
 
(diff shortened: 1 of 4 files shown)


sadick254:
Hey @hansonw . Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

PR #13754, opened 2017-02-06
Normalize disk drive letter in path on Windows

hron:
### Description of the Change

Currently atom creates two buffers for the same file if passed paths use difference case for disk drive letter, e.g. `d:\file.txt` and `D:\file.txt`

This change ensures that the absolute path always has disk drive letter in uppercase. 

### Alternate Designs

The other place where this fix could be applied is [findBufferForPath](https://github.com/atom/atom/blob/master/src/project.coffee#L331) function. However the used function [resolvePath](https://github.com/atom/atom/blob/master/src/project.coffee#L229) already has some path normalization code, so I think this is the best fit.

Actually Windows interpreters `file.txt` and `File.txt` as the same file. Maybe it makes sense to extend this patch and make Atom to do the same on this operation system, but I'm not sure because it can be file system specific. Any suggestions are welcome!

I believe the current behavior of the editor is wrong for Windows users, so it makes no sense to create a plugin to fix that.

### Benefits

The main benefit I would like to get is the ability to click on parsed backtrace in my terminal emulation app (ConEmu). Right now it works, but atom opens second buffer for the same file because it doesn't normalize disk drive letter.

### Possible Drawbacks

The possible drawback is some broken tests, but I've checked that and it seems everything works.

### Applicable Issues

Here is the related issue #8221. It's not exactly about my problem, but it's very near.

### DIFF
--- a/spec/default-directory-provider-spec.coffee
+++ b/spec/default-directory-provider-spec.coffee
@@ -28,6 +28,15 @@ describe "DefaultDirectoryProvider", ->
       directory = provider.directoryForURISync(nonNormalizedPath)
       expect(directory.getPath()).toEqual tmp
 
+    it "normalizes disk drive letter in path on #win32", ->
+      provider = new DefaultDirectoryProvider()
+      nonNormalizedPath = tmp[0].toLowerCase()+tmp.slice(1)
+      expect(tmp).not.toMatch /^[a-z]:/
+      expect(nonNormalizedPath).toMatch /^[a-z]:/
+
+      directory = provider.directoryForURISync(nonNormalizedPath)
+      expect(directory.getPath()).toEqual tmp
+
     it "creates a Directory for its parent dir when passed a file", ->
       provider = new DefaultDirectoryProvider()
       file = path.join(tmp, "example.txt")
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -15,7 +15,7 @@ class DefaultDirectoryProvider
   # * {Directory} if the given URI is compatible with this provider.
   # * `null` if the given URI is not compatibile with this provider.
   directoryForURISync: (uri) ->
-    normalizedPath = path.normalize(uri)
+    normalizedPath = @normalizePath(uri)
     {host} = url.parse(uri)
     directoryPath = if host
       uri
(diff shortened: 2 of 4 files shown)


hron:
@damieng and @Abhi2424shek please review again. The normalization shouldn't happen on non-Windows  now. I've also fixed linter issues in specs.

abishekvashok (approved):
LGTM

spec/default-directory-provider-spec.coffee
```diff
@@ -1,4 +1,5 @@
 DefaultDirectoryProvider = require '../src/default-directory-provider'
+platform = require './spec-helper-platform'
```
50Wliu: Don't use spec-helper-platform: name your test with a `#win32` in it, like `it "normalizes disk drive letter on #win32"`.  This will cause Jasmine to only run that spec on Windows.

hron: Thank you for the explanation. The specs have been updated.

src/default-directory-provider.coffee
```diff
@@ -42,3 +42,15 @@ class DefaultDirectoryProvider
...
+  # Public: Normalizes path.
+  #
+  # * `uri` {String} The path that should be normalized.
+  #
+  # Returns a {String} with normalized path.
+  normalizePath: (uri) ->
+    # Normalize disk drive letter on Windows to avoid opening two buffers for the same file
+    pathWithNormalizedDiskDriveLetter = uri
+    if !!process.platform.match(/^win/) and matchData = uri.match(/^([A-Za-z]):/)
+      pathWithNormalizedDiskDriveLetter = "#{matchData[1].toUpperCase()}#{uri.slice(1)}"
```
ericcornelissen: Why are you using `!!process.platform.match(/^win/)` and not just `process.platform.match(/^win/)`?

Also, you could potentially use the following construct, it seems cleaner to me:
```coffeescript
pathWithNormalizedDiskDriveLetter =
  if !!process.platform.match(/^win/) and matchData = uri.match(/^([A-Za-z]):/)
    pathWithNormalizedDiskDriveLetter = "#{matchData[1].toUpperCase()}#{uri.slice(1)}"
  else
    uri
```

ericcornelissen (commented):
LGTM, I've got two remarks tho

src/default-directory-provider.coffee
50Wliu: You should in fact be using `if process.platform isnt 'win32' and...`
There is no win64.

hron:
Thank you for the advices! All remarks have been incorporated.

src/default-directory-provider.coffee
```diff
@@ -42,3 +42,17 @@ class DefaultDirectoryProvider
...
+
+  # Public: Normalizes path.
+  #
+  # * `uri` {String} The path that should be normalized.
+  #
+  # Returns a {String} with normalized path.
+  normalizePath: (uri) ->
+    # Normalize disk drive letter on Windows to avoid opening two buffers for the same file
+    pathWithNormalizedDiskDriveLetter =
+      if process.platform is 'win32' and matchData = uri.match(/^([A-Za-z]):/)
```
damieng: Seems redundant to include A-Z - if it's already upper-cased we don't need to process it.

src/default-directory-provider.coffee
```diff
@@ -42,3 +42,17 @@ class DefaultDirectoryProvider
...
+        pathWithNormalizedDiskDriveLetter = "#{matchData[1].toUpperCase()}#{uri.slice(1)}"
```
damieng: Given the assignment is already on line 53 this should just be the quoted result.

damieng (changes requested):

hron:
@damieng The requested changes have been made. Thank you for the detailed review.

damieng (approved):

PR #13760, opened 2017-02-06
Open empty editor as a pending pane item

winstliu:
### Requirements

### Description of the Change

If the `Open Empty Editor on Startup` option is checked, open it as a pending pane item.  This way, if you open a folder and then a file, it doesn't stick around.  However, if you just wanted to open Atom for some note-taking purposes, the existing behavior is not affected as the pending pane item will become non-pending as soon as you write something.  I've always been annoyed at having to manually close the untitled editor after I open a file.

### Alternate Designs

None.

Already in core.

### Benefits

No longer have to manually close the Untitled tab after opening a folder.

### Possible Drawbacks

Maybe some people were relying on this?  Though Ctrl+N shouldn't be that bad.

### Applicable Issues

None that I could find.


50Wliu:
Yeah...so it looks like the crash is reproducible on CI too.  If anyone has any ideas for how to improve this test, I'm all ears, especially since `getPendingItem` is private.

50Wliu:
Ready for review.  I've managed to fix the crashing spec.

rafeca:
Thanks! 😍

As discussed offline, once the conflicts are resolved we can merge this one 😃

50Wliu:
@smashwilson I'm a little confused by some of these specs - for example, shouldn't [this one](https://github.com/atom/atom/blob/41d934fc027913e43aab9af81d4b41dcac394627/spec/main-process/atom-application.test.js#L62) be opening an untitled editor, as per `core.openEmptyEditorOnStart`?

smashwilson:
@50Wliu: The initial empty editor is [opened by `AtomEnvironment`](https://github.com/atom/atom/blob/dffe4556681dcca704746162136dcb12e4cd71b1/src/atom-environment.js#L926-L932). The `AtomApplication` tests stub out the windows that they open, so they don't create a real environment and the empty editor isn't accounted for by the assertions there.

We do have some other cases where `AtomApplication` triggers the creation of an empty editor by passing `{pathsToOpen: [null]}` (although I'm not sure I understand why). These are currently [filtered out in `StubWindow`](https://github.com/atom/atom/blob/41d934fc027913e43aab9af81d4b41dcac394627/spec/main-process/atom-application.test.js#L1071) although arguably we _should_ find a way to account for them 🤔

50Wliu:
@rafeca whoops, this wasn't _quite_ ready yet - my merge conflict resolution simply got rid of all the tests I wrote for it as I waited for Ash's answer :P. I'll open a followup PR adding specs in the correct location, though 🙂.

rafeca:
> @rafeca whoops, this wasn't quite ready yet - my merge conflict resolution simply got rid of all the tests I wrote for it as I waited for Ash's answer :P. I'll open a followup PR adding specs in the correct location, though 🙂.

Ups, sorry about it 😅

50Wliu:
Hah, just kidding, again! Looks like I _did_ have a spec asserting the pending state in atom-environment already. Sorry for the false alarm!

PR #13764, opened 2017-02-07
Disable soft wrap on mini editors

### LINKED ISSUE #473: Unable to see whole path when creating file with very long path/name and dashes
Try creating a new file (or duplicating/renaming one), and type a very long name into the dialog box (e.g. some/path/my-very-very-long-filename-to-test-this-issue).

You will notice that once the text reaches the end of the box, you cannot see the end of the filename anymore. Moving left/right/up/down with arrows doesn't help. Only pressing space seems to trigger the display of the end of the filename.

winstliu:
### Requirements

### Description of the Change

Don't soft wrap mini editors.  It seems like this isn't really useful for mini editors.

### Alternate Designs

There could be a separate programmatic toggle for mini editors that also looks at the global setting.

Because mini editors already live in core!

### Benefits

No more "why can't I scroll right when I type something really long?"

### Possible Drawbacks

Some mini editors may be relying on this, though I think that would be bad practice considering it's a user setting.

### Applicable Issues

Fixes atom/tree-view#473

/cc @simurai

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -5961,7 +5961,7 @@ describe "TextEditor", ->
       expect(editor.getGrammar().name).toBe 'CoffeeScript'
 
   describe "softWrapAtPreferredLineLength", ->
-    it "soft wraps the editor at the preferred line length unless the editor is narrower", ->
+    it "soft wraps the editor at the preferred line length unless the editor is narrower or the editor is mini", ->
       editor.update({
         editorWidthInChars: 30
         softWrapped: true
@@ -5973,6 +5973,9 @@ describe "TextEditor", ->
 
       editor.update({editorWidthInChars: 10})
       expect(editor.lineTextForScreenRow(0)).toBe 'var '
+
+      editor.update({mini: true})
+      expect(editor.lineTextForScreenRow(0)).toBe 'var quicksort = function () {'
 
   describe "softWrapHangingIndentLength", ->
     it "controls how much extra indentation is applied to soft-wrapped lines", ->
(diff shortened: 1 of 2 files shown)


simurai:
> Some mini editors may be relying on this, though I think that would be bad practice considering it's a user setting.

Hmm. yeah, hard to guess how many packages would be affected. I think it's fine since mini editors are meant as "single line" inputs.

Packages could use `<textarea class='input-textarea' placeholder='Text Area'></textarea>` for simple multi-line inputs (if they don't need any grammar support).

A while ago we considered adding a `<atom-text-area>`: https://github.com/atom/atom/issues/5503. I think that's still a good idea. Something in between mini and normal editors.

jeroenzeegers (approved):

andrewstart:
I've got some mini editors that start out as single line, but soft wrap for more lines as they get longer, regardless of user settings for the language (it is one of many fields in the editor pane). Right now I am calling `editor.getModel().setSoftWrapped()` to do this. As such, I'd really like a programatic toggle for soft wrapping mini editors.

50Wliu:
@simurai what do you think about the above? :point_up:

simurai:
This PR https://github.com/atom/atom/pull/10955 adds some options to make an `<atom-text-editor>` more like an `<textarea>` element:

text-area

This might not be documented yet??? Here the options used for the above gif:

```html
<TextEditor
  ref="editor"
  softWrapped={true}
  placeholderText="Type here"
  lineNumberGutterVisible={false}
  showInvisibles={false}
  autoHeight={true}
  scrollPastEnd={false}
/>
```

nkkollaw:
Horray!!!

PR #13795, opened 2017-02-11
Eliminate TextEditorComponent's reliance on MutationObserver for detecting changes in visibility, dimensions, and background color

nathansobo:
With the introduction of Chrome 53 in Electron 1.4, we now have access to a fully functional `IntersectionObserver` API, which can be used to detect when editors change in visibility. This means we can discard the previous `MutationObserver` based polling system, which creates a drag on frame rates in certain circumstances. This impact is hard to quantify because *measuring* the frame rate with mutation observers in place in the current scrolling scheme seems to slow performance dramatically, making this hard to measure. Regardless it's a big simplification to not require full DOM observation.

This PR replaces polling with the following approaches:

* Visibility detection: [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)
* Resize detection of editors and overlay decorations: Scroll event hack via the [`element-resize-detector`](https://github.com/wnr/element-resize-detector) library.
* Transferring background to editor tiles: Plain CSS with `background-color: inherit` (now possible with the removal of the shadow DOM)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "color": "^0.7.3",
     "dedent": "^0.6.0",
     "devtron": "1.3.0",
+    "element-resize-detector": "^1.1.10",
     "event-kit": "^2.1.0",
     "find-parent-dir": "^0.3.0",
     "first-mate": "6.1.0",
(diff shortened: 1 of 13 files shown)


as-cii:
Amazing work, this looks great. ⚡️

nathansobo:
Remaining failures are related to calling `atom.views.performDocumentPoll` in `find-and-replace` and `wrap-guide` tests.

PR #13783, opened 2017-02-10
Correct apm path for Windows

richardtin:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

Correct `apm` path for Windows platform

### Alternate Designs

N/A

### Why Should This Be In Core?

N/A

### Benefits

Make `apm` command can work in Git Bash of Windows version.

### Possible Drawbacks

N/A

### Applicable Issues

N/A

PR #13803, opened 2017-02-13
Wrong location for apm

okonomiyaki3000:
It seems this was set to the wrong location for apm.

### Requirements

### Description of the Change
Change the path to apm in apm.sh

### Alternate Designs
None

Because apm does not work if the path is wrong.

### Benefits
APM will work.

### Possible Drawbacks
None

### Applicable Issues
None


torn4dom4n:
https://github.com/atom/atom/pull/13783

Issue #13377, opened 2016-12-03
Uncaught Error: The element has already been freed!


divlook:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.12.6
**Electron Version**: 1.3.9
**System**: Mac OS X 10.11.6
**Thrown From**: Atom Core

### Stack Trace

Uncaught Error: The element has already been freed!

```
At /Applications/Atom.app/Contents/Resources/app.asar/src/tiled-component.js:76

Error: The element has already been freed!
    at DOMElementPool.module.exports.DOMElementPool.free (/Applications/Atom.app/Contents/Resources/app.asar/src/dom-element-pool.js:83:15)
    at DOMElementPool.module.exports.DOMElementPool.freeElementAndDescendants (/Applications/Atom.app/Contents/Resources/app.asar/src/dom-element-pool.js:63:12)
    at LinesTileComponent.module.exports.LinesTileComponent.destroy (/Applications/Atom.app/Contents/Resources/app.asar/src/lines-tile-component.js:37:34)
    at LinesComponent.module.exports.TiledComponent.removeTileNode (/Applications/Atom.app/Contents/Resources/app.asar/src/tiled-component.js:42:40)
    at LinesComponent.module.exports.TiledComponent.updateTileNodes (/Applications/Atom.app/Contents/Resources/app.asar/src/tiled-component.js:54:16)
    at LinesComponent.module.exports.TiledComponent.updateSync (/Applications/Atom.app/Contents/Resources/app.asar/src/tiled-component.js:30:12)
    at TextEditorComponent.module.exports.TextEditorComponent.updateSyncPreMeasurement (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:263:34)
    at TextEditorComponent.module.exports.TextEditorComponent.updateSync (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:193:12)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor-component.js:323:28
    at ViewRegistry.module.exports.ViewRegistry.performDocumentUpdate (/Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:219:9)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/view-registry.js:3:59
```

### Commands

```
     -0:28.1.0 git-control:toggle (atom-text-editor.editor)
[...]

PR #13806, opened 2017-02-13
Overhaul element pool and debug double-free errors

nathansobo:
In researching the cause of https://github.com/atom/atom/issues/13377, we've decided to make some changes to the element pool.

1) Based on a hypothesis that a package could be inserting foreign nodes into the text editor's DOM structure, we now only free elements that the element pool itself created.
2) If the above hypothesis is incorrect, we now report *assertion failures* rather than throwing exceptions, and include the HTML of the freed element just in case that might be helpful.

In addition, we've converted the class and tests to JS and removed some closure allocations on :fire: code paths.

We think we should probably merge this only to 1.15 (beta) because it's a somewhat risky change, but we'd like to get feedback pretty soon from it.

/cc @maxbrunsfeld


maxbrunsfeld:
👍

PR #13808, opened 2017-02-13
Fix code-signing on CircleCI

nathansobo:
We're seeing the following error message in [this failing build](https://circleci.com/gh/atom/atom/2588) on CircleCI:

```
SecKey API returned: -25293, (null)/Users/distiller/atom/out/Atom.app: unknown error -1=ffffffffffffffff
```

Based on [this Stack Overflow article](http://stackoverflow.com/questions/39868578/security-codesign-in-sierra-keychain-ignores-access-control-settings-and-ui-p) it looks like they *may* have upgraded their images to Sierra and are causing this problem.

This PR runs the ✨  command suggested in that article to hopefully fix the issue.

### DIFF
--- a/script/lib/code-sign-on-mac.js
+++ b/script/lib/code-sign-on-mac.js
@@ -33,6 +33,18 @@ module.exports = function (packagedAppPath) {
       '-T', '/usr/bin/codesign'
     ])
 
+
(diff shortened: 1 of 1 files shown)


nathansobo:
Merging this now because the build is getting past the code signing step on Circle.

PR #13819, opened 2017-02-14
Fix Arch Linux package names

quark-zju:
Arch Linux does not have `devel` packages. The include files and the libraries
are usually built in the main packages. There is neither `libX11-devel`, nor
`libxkbfile-devel`. They should be changed to `libx11` and `libxkbfile`.

### Description of the Change

This is a documentation fix.

### Alternate Designs

No alternatives.

Otherwise people using Arch Linux will be confused.

### Benefits

More friendly to contributors using Arch Linux.

### Possible Drawbacks

None.

### Applicable Issues

None.

### DIFF
--- a/docs/build-instructions/linux.md
+++ b/docs/build-instructions/linux.md
@@ -59,7 +59,7 @@ To also install the newly built application, use `--create-debian-package` or `-
 
 ### Arch
 
-* `sudo pacman -S --needed gconf base-devel git nodejs npm libgnome-keyring python2 libX11-devel libxkbfile-devel`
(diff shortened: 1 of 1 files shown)


gheeshuggah:
This ticket can be closed since the linux.md file has since been moved to https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#arch and the issue has been corrected there as
`sudo pacman -S --needed gconf base-devel git nodejs npm libsecret python2 libx11 libxkbfile`

50Wliu:
Thanks @gheeshuggah.

PR #13820, opened 2017-02-15
Avoid hangs when opening minified files

### LINKED ISSUE #979: Files with large extremely long lines cause hangs

maxbrunsfeld:
Fixes #979

#### Changes:
1. We force soft wraps for lines that exceed 500 chars. This avoids the rendering bottleneck that we used to experience because we don't break lines into horizontal tiles. It's actually a pretty useful behavior, I think, because a single un-wrapped line of code is pretty hard to read.

2. We don't pass more than 500 chars at a time to first-mate. There was already a limit to the number of tokens that we read from first-mate, but because the TextMate grammars can contain regexps with arbitrary lookahead and backtracking requirements, this token limit did not mitigate all of the performance problems of tokenizing large lines. Now we avoid passing more than 500 characters of text at a time to the tokenizer.

#### Benchmark results

We added a benchmark for opening large single-line files. When run agains Atom *prior* to this PR, this benchmark would basically not terminate 😬 . Now, it executes in a reasonable amount of time:

```
Opening a large single-line file:
512	1442.5700000000002
1024	2813.6100000000006
2048	5145.779999999995

Max time event loop was blocked after opening a large single-line file:
512	14.859999999999673
1024	8.095000000001164
2048	5.575000000004366

Clicking the editor after opening a large single-line file:
512	6.609999999999673
1024	3.9499999999970896
2048	4.955000000001746

Scrolling down after opening a large single-line file:
512	0.18499999999949068
1024	0.05000000000291038
2048	0.05000000000291038
```

You can see that the editor is immediately responsive after opening the file, regardless of its size. Unfortunately, the time to open a large single-line file still scales linearly with the size of the file. This is because the `DisplayLayer` currently indexes files on a line-wise basis, so for a single-line file, it has to index the entire file to answer any queries.

Still, a reasonably-sized minified file like [jQuery 3.1.1](https://code.jquery.com/jquery-3.1.1.min.js) (85 kb, 4 lines), opens almost instantly 🎉 .

/cc @ungb Can you give this a test, and link to any relevant issues about hangs when opening files with long lines?


ungb:
I'll take a look!  Thanks for getting to this.  I'll look for all the issues related to performance and add them here.

50Wliu:
Will this fix #979?

maxbrunsfeld:
> Will this fix #979?

Yeah, I think so.

ungb:
I was able to test this on windows x64 bit and saw that jquery.min.js loads up very quickly now where as before it causes Atom to freeze.  I was able to verify other single long line files now respond (after disabling spell-check)  

See https://github.com/atom/spell-check/issues/190 for issue and repro on spell-check package.

benogle:
Oh nice. 🎉

adamreisnz:
This is great, thanks for implementing

timdp:
First of all, I _love_ that there's finally a fix for Atom hanging on long lines. :+1:

So, question: shouldn't the `MAX_LINE_LENGTH_TO_TOKENIZE` const be a setting of some sort? Its value 500 seems pretty arbitrary.

nathansobo:
@timdp We considered making it a setting but it wasn't clear that anybody would want to have control over it since it's just circumventing a deficiency in our tokenizer until we have a chance to replace it with an improved parser. Are you interested in changing this value? If so, what's your rationale?

timdp:
@nathansobo I probably wouldn't change it myself, but as a rule of thumb, I try to avoid magic constants like these in my own code. I can see how this feature is a stopgap though, so the simpler, the better, I suppose.

mbrancato:
I tested 1.15-beta3 with this patch and it does not seem to solve #979 (which has been closed). My example is a file which is either a single long lone of base64 encoded data, of a file which has a line with a large blob of base64 encoded data.

An example here would be an HTML file with embedded images represented as base64 on a single line.

Atom 1.15-beta3 actually crashed while attempting to open the file.

maxbrunsfeld:
Can you provide an example file @mbrancato?

mbrancato:
Here's a file about the size I'm working with. I'm not sure what caused the crash on the file I'm opening. It happened only the first time I tried with the 1.15-beta3. Subsequent attempts open the file, but syntax highlighting is broken.

Opening a large PowerShell file with an embedded binary is what I was attempting. The version 1.15-beta3 appears to be slower at opening files, and handling edits of a file where a large line exists. The stable 1.14.3 appears to be much faster at inserting lines, and even rending typing. I'm running the Mac version, if that matters.

Attached is a 9MB file of random data, then base64 encoded. It takes about 40 seconds to open on 1.15-beta3 for me.

[9MB-random-base64.txt](https://github.com/atom/atom/files/785470/9MB-random-base64.txt)

I'll keep trying to run down the differences on my end to isolate anything useful.

andrewstart:
Possibly related, the mere 149 bytes here causes the JS syntax highlighter to hang - shortening the long line a little more allows it to finish, but it takes long enough to bring up the 'keep waiting/close' dialog. This happens on a Macbook Pro & a Windows 7 box.
[hang.txt](https://github.com/atom/atom/files/805599/hang.txt)

50Wliu:
Ok yeah, this is really bad.  @andrewstart can you please create a new issue on [atom/language-javascript](https://github.com/atom/language-javascript) so that we can prioritize it?  Thanks!

Oblik-Design:
it is getting ridiculus, 500 lines in html and hitting return to get a new line, takes 4 seconds on a strong computer. I love all the rest, but this is insane.

50Wliu:
@Oblik-Design please open a new issue.

Oblik-Design:
there are hundreds of these, performance is a serious issues, a simple website will hit slowdowns atom ranks very last in performance, it is the only reason why it is loosing users by the secondes.

adamreisnz:
I must agree with @Oblik-Design that raising individual issues for each performance related item is not really workable. I have now also reached the point that developing with Atom on my normally quite performant Macbook Air is becoming less and less productive because of intermittent lagging and performance related problems. 

Sometimes it's just typing code that causes intermittent temporary freezes. Needless to say that's not really reproducible in a consistent manner, not will it affect people with a faster computer, so it's hard to debug. On my iMac the issues are less noticeable because there's more memory and processing power, but to be honest, the editor should run smoothly on a decent laptop as well.

Is there anything else we can do to help analyse these performance problems other than opening more new issues?

Oblik-Design:
I'm on a 4 core i7 , coud not imagine how bad it must be on an air... but again, Atom is I guess, single threaded ?

50Wliu:
Opening a new issue allows us to triage it more effectively if you go through the prerequisites.  For example, does it happen in safe mode?  Is there already an open issue describing your performance issue?  Are you on the latest version?  Can you collect a profile of the slow performance? etc.  That gives us _much_ more information to work with than "500 lines in html and hitting return to get a new line, takes 4 seconds on a strong computer".  Creating an issue also ensures that your performance woes aren't lost in the comments of an already-merged PR (that frankly doesn't seem to be that related to what you're seeing @Oblik-Design...unless those 500 lines are also very long).

As some examples, https://github.com/atom/language-javascript/issues/493 and https://github.com/atom/language-javascript/issues/495 were both addressed quickly once the issues were opened.

Oblik-Design:
it is ok in safe mode, but i'm running only a few packages, terns , linter, emmet , highlight selected, and the autocompletes, you know, the ones you need for JS development  , and no typescript or even react stuff. I only do es6 , I do babel through webpack, so not using any packages to do the heavy lifting. I had issues at first , could not use atom, because it did not run on 4k monitors well, this got "fixed", js is fast , scss to, but html is a joke. Simple sites, no long lines.

Oblik-Design:
Still I have to use Brackets if I want to have a look at what webpack is doing.  It craches EVERY time I want to open the js file that webpack creates

50Wliu:
And like I said earlier, it would be better for the both of us if you opened a new issue.  But since you said the performance issues do not happen in safe mode, I would encourage you to go through your packages and disable individual packages until there are no performance issues left, then file issues on the appropriate package repositories.  (You can also use the [package-cop](https://atom.io/packages/package-cop) package to help with that)

Oblik-Design:
well I can saffely say that the problem is with https://atom.io/packages/linter , ouchhh. this is needed

ungb:
@Oblik-Design Thanks for your report on performance issue you are seeing.  As @50Wliu said, it's tough for us to triage issue that is within a closed PR as it's easy to get lost.  When creating another issue and following the template it helps us triage and identify the underlying issue and get a fix out.  Since you were able to pin point the issue to the linter package, please file an issue for them here: https://github.com/steelbrain/linter/issues/new.

Oblik-Design:
well there is another one ... Emmet. I just cant loose both of these.

stopdropandrew:
This caused me some trouble this morning. While I think having soft-wrap @ 500 characters on by default is totally fine, it's pretty strange to View->Toggle Soft Wrap on and off and just switch between a 200 and 500 character soft-wrap.

For reference, I was trying to view a csv roughly 3000 characters per line. Is there a middle ground?

Oblik-Design:
I was completely wrong, it is only EMMET. I got confused.... my bad, linter is ok.
it at least makes sens, linter screwing html made no sense. but it's  Emmet playground.

lee-dohm:
I'm going to lock this conversation. We prefer not to lock conversations but this has gotten out of hand and I feel that valuable information is getting lost. Here's what we need people to do:

* If you have an issue, bug, problem or enhancement you want to report: Open a new issue and fill out the issue template
* If you want to discuss Atom's performance in a more general sense: Please do so on [Discuss, the official Atom and Electron message board](https://discuss.atom.io)

Thanks everyone for your understanding!

Oblik-Design:
I can also now open the webpack build file without issues. It was emmet fault, sorry guys. My bad. But I use EMMET so much ...

lee-dohm:
I've locked this conversation. We prefer not to lock conversations but this has gotten out of hand and I feel that valuable information is getting lost. Here's what we need people to do:

* If you have an issue, bug, problem or enhancement you want to report: Open a new issue and fill out the issue template
* If you want to discuss Atom's performance in a more general sense: Please do so on [Discuss, the official Atom and Electron message board](https://discuss.atom.io)

Thanks everyone for your understanding!

PR #13823, opened 2017-02-15
Upgrade to Babel 6 and apply fewer transformations when transpiling babel files

as-cii:
This pull request upgrades babel to the latest version and runs a minimal set of transformations when transpiling babel files. In particular, only the following cases will be affected by the transpilation step:

1. `import` or `export` expressions.
1. `async`/`await` expressions.
1. Do expressions.
1. Function bind expressions.
1. Decorators.
1. Object rest spread expressions (e.g. `{a: 1, b: 2, …anotherObject}`).
1. Class properties.
1. JSX code.
1. Flow code.

We shouldn't need any other transformation because Electron runs on a pretty recent V8 version. As a result, the generated code should be cleaner and source maps should look less crazy as well. Please, note that I was forced to update babel because the previous `asyncToGenerator` plugin had some bugs when not transpiling ES2015 classes.

/cc: @atom/core


package.json
```diff
@@ -18,7 +18,13 @@
     "atom-keymap": "7.1.20",
     "atom-select-list": "0.0.12",
     "atom-ui": "0.4.1",
-    "babel-core": "5.8.38",
+    "babel-core": "6.22.1",
+    "babel-plugin-add-module-exports": "^0.2.1",
+    "babel-plugin-transform-async-to-generator": "^6.22.0",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.23.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
```
damieng: 👍

nathansobo:
@BinaryMuse I'm curious about the roadblocks you ran into when you tried to upgrade our core Babel recently. Is there anything we're missing here?

BinaryMuse:
@nathansobo I was thinking about that too — I don't recall what issues I ran into with upgrading Babel specifically. One may have been a lack of confidence that the upgraded/reconfigured Babel would perform the same way as the old one and would break packages. That said, one of the main reasons I stopped working on that and started on the per-package transpilation was because the Relay Babel transform requires a project-specific schema and so can't be enabled "globally."

static/babelrc.json
```diff
@@ -1,7 +1,12 @@
...
-  "breakConfig": true,
   "sourceMap": "inline",
-  "blacklist": ["es6.forOf", "useStrict"],
-  "optional": ["asyncToGenerator"],
-  "stage": 0
+  "plugins": [
+    "add-module-exports",
+    "transform-async-to-generator",
+    "transform-class-properties",
+    "transform-es2015-modules-commonjs",
```
BinaryMuse: I seem to remember that there was *some* kind of issue with this transform and strict mode... but I can't for the life of me remember what it was. -_-

static/babelrc.json
```diff
@@ -1,7 +1,12 @@
 {
-  "breakConfig": true,
   "sourceMap": "inline",
-  "blacklist": ["es6.forOf", "useStrict"],
-  "optional": ["asyncToGenerator"],
-  "stage": 0
+  "plugins": [
```
BinaryMuse: We previously supported legacy decorators, right? I think we'd need to include that plugin.

static/babelrc.json
as-cii: Interesting, the only issue I could find was that the following code wasn't being generated:

```js
module.exports = exports['default']
```

This caused interoperability issue with e.g. CoffeeScript, but it was fixed by adding the `add-module-exports` transform (as recommended on https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/).

as-cii:
This morning I compiled all the babel files that package authors have published and have identified some that can't compile with the current babel configuration. Some of them are clearly related to the changes in this pull request, but others contain syntactic errors that might have existed prior to these changes.  For the former group I will submit a pull request and simply open an issue for the latter.

### Packages breaking because of this pull request

* [ ] `ansible-vault/lib/ansible-vault.js: Deleting local variable in strict mode (106:6)`
* [ ] `atom-go-jump-test/lib/atom-go-jump-test.js: package is a reserved word in strict mode (59:6)`
* [ ] `atom-jive/spec/atom-ternjs-spec.js: package is a reserved word in strict mode (16:6)`
* [ ] `atom-ternjs/spec/atom-ternjs-spec.js: package is a reserved word in strict mode (15:6)`
* [ ] `atom-ysp/lib/product/common.js: Invalid number (169:24)`
* [ ] `autocomplete-lodash/lib/autocomplete-lodash-provider.js: arguments is a reserved word in strict mode (65:45)`
* [ ] `autosave-onchange/lib/autosave-delay.js: package is a reserved word in strict mode (22:6)`
* [ ] `autosave-onchange/spec/package-spec.js: package is a reserved word in strict mode (10:8)`
* [ ] `black-bird/lib/RemoteDirectory.js: Invalid number (14:14)`
* [ ] `copy-path/lib/index.js: package is a reserved word in strict mode (33:40)`
* [ ] `language-blob/lib/blob-provider.js: arguments is a reserved word in strict mode (109:12)`
* [ ] `language-markdown/lib/__main.js: Unexpected token, expected , (16:6)`
* [ ] `linter-erlc/lib/init.js: Unexpected token (112:27)`
* [ ] `medium-upload/lib/medium-upload-view.js: package is a reserved word in strict mode (105:46)`
* [ ] `mope/lib/moie.js: interface is a reserved word in strict mode (120:18)`
* [ ] `mope/lib/util.js: interface is a reserved word in strict mode (27:10)`
* [ ] `neovimbed/spec/spec-helper.js: package is a reserved word in strict mode (24:65)`
* [ ] `package-foreman/lib/package-foreman.js: package is a reserved word in strict mode (33:47)`
* [ ] `package-foreman/lib/package.js: package is a reserved word in strict mode (101:39)`
* [ ] `quicknote/src/components/Overlay.js: package is a reserved word in strict mode (104:10)`

### Packages with syntactic errors

* [ ] `atom-calca/lib/parser/CalcaParser.js: Unexpected token, expected , (63:11)`
* [ ] `atom-setup-manager-package/lib/checkout.js: 'super' outside of function or class (17:6)`
* [ ] `autocomplete-jsp/src/tag-provider.js: Unexpected token, expected , (139:5)`
* [ ] `autocomplete-jsp/src/utils.js: Unexpected token, expected , (39:1)`
* [ ] `black-bird/lib/Server.js: Unexpected token, expected , (72:5)`
* [ ] `build-gulp/lib/gulp.js: Unexpected token, expected ; (38:7)`
* [ ] `javascript-importer/lib/import-dialog.js: Unexpected token, expected ; (22:38)`
* [ ] `save-workspace/lib/commands/open.js: Unexpected token, expected , (19:5)`
* [ ] `save-workspace/lib/commands/remove.js: Unexpected token, expected , (19:5)`
* [ ] `save-workspace/lib/commands/save.js: Unexpected token, expected , (19:5)`
* [ ] `xkcd-comics/lib/xkcd.js: Unexpected token, expected , (22:5)`
* [ ] `xkcd-debounce-fix/lib/xkcd.js: Unexpected token, expected , (22:5)`

as-cii:
Once CI is 💚 I think this could be ready to go, so that we can let it bake for a while on master before rolling the railcars next month. 

I am planning to 🚢 it by the end of today, so it'd be great if anyone could give it a last 👀 before merging it.

nathansobo:
Amazing initiative.

spec/main-process/file-recovery-service.test.js
```diff
@@ -112,7 +112,7 @@ describe("FileRecoveryService", () => {
       const mockWindow = {}
       const filePath = temp.path()
       fs.writeFileSync(filePath, "content")
-      fs.chmodSync(filePath, 0444)
+      fs.chmodSync(filePath, 0o444)
```
damieng: Intentional?

50Wliu: Most likely: the new ES6 way of specifying octals is `0o`.  The previous leading-zero way is more confusing as well.

maxbrunsfeld: Yeah, octal literals w/ leading zeros was a non-standard extension that some browsers provided, but I guess the new babel doesn't allow.

quarterto:
Hey @as-cii, I'm running into an issue where `export const {foo} = bar` isn't exporting the variable. I had to change it to `const {foo} = bar; export {foo}`. The former syntax is correct, is this something that could have been broken by this PR?

thomasjo:
@quarterto That syntax isn't valid AFAIK. See e.g. https://developer.mozilla.org/en/docs/web/javascript/reference/statements/export.

quarterto:
MDN is incomplete here. The [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-exports) has `export VariableStatement` as one of the productions. `VariableStatement` encompasses destructuring assignment.

The Babel REPL [correctly transforms this](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=es2015&targets=&browsers=&builtIns=false&code=export%20const%20%7Bfoo%7D%20%3D%20bar%3B), so I wonder what's going on.

thomasjo:
If I'm reading the spec correctly, `VariableStatement` only encompasses `var` statements/declarations.

Briefly scanned through the `Declaration` grammar "chain" to see if that might include [destructuring assignments](http://www.ecma-international.org/ecma-262/6.0/#sec-destructuring-assignment), but it doesn't look like it does. 

I might be wrong, and please let me know if I am, but it looks like Babel has incorrectly implemented the spec.

quarterto:
I think it works via `export Declaration` → `LexicalDeclaration` → `BindingList` → `LexicalBinding` → `BindingIdentifier` → `BindingPattern`.

Either way, this is working in the Babel REPL but for some reason not in Atom's Babel, which I don't understand.

thomasjo:
Since we're not using the `es2015` preset, it doesn't look like we're actually transpiling that export statement. We're using two plugins that affect export statements, but they do not seem to be related to general exports, only providing support for CommonJS, as well as supporting `default` or some such.

So if the syntax you're using is indeed valid, this is a problem with the Node.js that ships with the our version of Electron — it simply doesn't support your syntax. Hopefully something that get's resolved with #12696 (but might also require a bump to Electron v1.6.x).

quarterto:
No version of node supports `export`; you're using `babel-plugin-transform-es2015-modules-commonjs` which should support this. I'm about to try compiling that with Atom's babelrc to see what the output is.

as-cii:
Even though I am a little unclear about whether or not that syntax is supported in the spec, I actually believe this is a bug in Babel. 

Babel seems to recognize that syntax properly with the current configuration, but since we are not using the `transform-es2015-destructuring` plugin it's failing to apply the `transform-es2015-modules-commonjs` transformation correctly. For instance, consider the following snippet:

```js
export const {a} = {a: 1}
```

With the `transform-es2015-destructuring` plugin, it will be converted by Babel into:

```js
Object.defineProperty(exports, "__esModule", {
  value: true
});
/** @babel */

var _a = { a: 1 };
const a = _a.a;
exports.a = a;
```

Viceversa, if we exclude the `transform-es2015-destructuring` plugin, the syntax will still be recognized and parsed properly, but the applied `transform-es2015-modules-commonjs` will look like the following:

```js
Object.defineProperty(exports, "__esModule", {
  value: true
});
/** @babel */

const { a } = { a: 1 };
```

Which is confusing and unexpected. Still, I am a little reluctant to add the `transform-es2015-destructuring` plugin back because I couldn't find other packages using a similar syntax and fixing this issue on the `atom-which-key` package seems also quite straightforward.

@thomasjo @quarterto: what do you think?

quarterto:
Yeah, that seems reasonable. I didn't know about this syntax either until seeing it in `atom-which-key` 😄 

We should probably report this to Babel.

thomasjo:
> No version of node supports export; you're using babel-plugin-transform-es2015-modules-commonjs which should support this.

@quarterto Sorry about the false information; I guess I've been confusing actual support with planned support or some such. Thanks for the clarification!



> Which is confusing and unexpected. Still, I feel a little reluctant to add the transform-es2015-destructuring plugin back because I couldn't find other packages using a similar syntax and fixing this issue on the atom-which-key package seems also quite straightforward.
> 
> @thomasjo @quarterto: what do you think?

@as-cii Makes sense to keep it as is. Might be worth investigating if we have any [complex destructuring assignments](https://github.com/fhinkel/six-speed/blob/master/tests/destructuring/destructuring.es6) in core, since those are seemingly an order of magnitude slower in native ES6, compared to transpiled to ES5 via Babel (ref: https://fhinkel.github.io/six-speed/).

Arcanemagus:
I'm only guessing that this is the PR responsible for this change, but https://github.com/atom/toggle-quotes/pull/48 is now the second package that I've had to make that change to for it to work properly under Atom v1.16.0-beta0. Should this be filed as a new issue?

avindra:
Is it possible to track what version of v8 atom is using easily? I guess it's based on what electron is on?

Because [babel-preset-env](https://github.com/babel/babel-preset-env) is a thing, and it auto-selects the transforms necessary. 

Specifically, it has a `targets.node` option (https://github.com/babel/babel-preset-env#targetsnode) which allows you to transpile only what you need.

50Wliu:
Open Atom and type `process.versions` in the developer tools.  The raw data is available [here](https://github.com/electron/electron.atom.io/blob/gh-pages/_data/versions.json), but it doesn't look like versions other than the latest are displayed on the website.

as-cii:
> I'm only guessing that this is the PR responsible for this change, but atom/toggle-quotes#48 is now the second package that I've had to make that change to for it to work properly under Atom v1.16.0-beta0. Should this be filed as a new issue?

I am a little concerned about this as, again, Babel was transpiling things in a weird way and, even though the code in toggle-quotes is syntactically correct, its behavior will change as a result of this pull request. @Arcanemagus: could you show how you had to change that second package, please? Was it because of the same problem?

Anyways, we have received only a few reports about this pull request breaking existing packages and maybe we can keep it, but I wonder if it's just because these changes were rolled out on the beta channel only. 

Alternatively, I guess we could use different configurations for bundled packages and third-party packages: we'd use the new config for the former, and keep using the old one for the latter.

@atom/maintainers: thoughts? 💭

Arcanemagus:
> @Arcanemagus: could you show how you had to change that second package, please? Was it because of the same problem?

Yep, the exact same issue: It was declaring the `activate` function (and others) as arrow functions, switching them in the same manner as in `toggle-quotes` fixed it. (See https://github.com/AtomLinter/linter-glsl/pull/54/commits/d0c6567d9dd03adbbbeb7799aeb987d64ec4c44f)

as-cii:
After searching for usages of `activate: () =>` I was able to find 60 packages that are mistakenly using an arrow function in a global scope. Using this syntax in the new version of babel:

```js
export default {
  activate: () => {
    this.test = 1
  }
}
```

Will transpile to:

```js
exports.default = {
  activate: () => {
    undefined.test = 1;
  }
}
```

Which will yield the error @Arcanemagus saw. I am seriously considering reverting this, as things have changed quite significantly in the latest babel version and I have some concerns about breaking lots of packages. It is unfortunate because the code was not supposed to work in the first place, but apparently the previous babel version was more forgiving.

thomasjo:
@as-cii Or we divide up the work, and submit PRs to all offending packages. It needs to be fixed, so we might as well do it now before more packages end up using invalid syntax. Thoughts? :thought_balloon:

Arcanemagus:
I can fix all the `linter-*` ones in the AtomLinter org, I'll get started on that.

nathansobo:
We're probably going to revert this change and try again another time. Our current plan is as follows:

* Allow arbitrary assets to be published as part of Atom packages rather than requiring everything to be stored in git.
* Publish pre-transpiled assets for all current Atom packages depending on transpilation.
* Bake a list of packages with assets into Atom, and prompt the user to reinstall these packages on upgrade.
* Remove support for global transpilation from Atom. Support custom package transpilers only in dev mode.

PR #13825, opened 2017-02-15
Relativize spec paths more when reporting

winstliu:
### Requirements

### Description of the Change

First, here's a before/after comparison on Windows:

|Before|After|
|-------|-----|
|specs-before|specs-after|

Much nicer.

Changes include:
* Updating to the new `.<anonymous>` syntax
* Stripping away `file:///` identifiers on Windows paths and converting the rest of the path to use Windows path separators (in order for it to get relativized correctly)
* Relativizing paths in parentheses

### Alternate Designs

Instead of relativizing the path twice, depending on if its in parentheses or not, a more general `line.replace(///#{_.escapeRegExp(spec.specDirectory}#{_.escapeRegExp(path.sep}///, '')` could be used.  I am unsure if that would potentially relativize more than we wanted.

Because it fixes already-implemented code in core.

### Benefits

See before/after picture above.

### Possible Drawbacks

I don't foresee any.

### Applicable Issues

None.

/cc @damieng for the Windows portions.

### DIFF
--- a/spec/atom-reporter.coffee
+++ b/spec/atom-reporter.coffee
@@ -1,4 +1,5 @@
 path = require 'path'
+process = require 'process'
 _ = require 'underscore-plus'
 grim = require 'grim'
 marked = require 'marked'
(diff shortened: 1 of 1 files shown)


50Wliu:
If specs are necessary for this change I can add them.

damieng:
This looks good to me. I'm okay with keeping windows-style paths for local stuff although for the areas where we report remotely - bugsnag, GA, notification to github issue - I prefer to normalize them to Unix style so that they are correctly combined into a single issue where appropriate.

Thanks for this!

PR #13833, opened 2017-02-16
[RFC] Remove DOMElementPool

maxbrunsfeld:
⚠️  Not sure if we should do this ⚠️ 

Since we introduced the `DOMElementPool`, Chrome has [made some improvements](https://chromium.googlesource.com/chromium/src/+/master/third_party/WebKit/Source/platform/heap/BlinkGCAPIReference.md) to its garbage collection of DOM nodes. With those improvements, storing DOM elements in an object pool does not provide the same performance benefits that it once did. This PR removes the `DOMElementPool` and replaces its usages with simple calls to `document.createElement` and `HTMLElement.remove`.

#### Results

I compared the experiences of scrolling for 4 seconds straight in a large file with syntax highlighting (`text-editor-component-spec.js`) before and after removing the element pool. I could discern a difference in the average frame rate or its variance from the timelines.

Looking at the statistics though, it looks like the CPU stays idle for a slightly higher percent of the time **with** the `DOMElementPool`. And as you would expect, the memory usage is slightly more stable with the pool.

**With element pool**

before

**Without element pool**

after

#### Conclusion

Overall, the pool does improve performance very slightly (maybe imperceptibly).

@nathansobo @as-cii @kuychaco In a way, the element pool is still a good idea, but if we keep it, we need to understand that bug that we saw in the line number rendering. What do you think?


as-cii:
Thanks for collecting this data, @maxbrunsfeld! ❤️

Based on your results above, it looks like the difference between the two relative idle times is about 0,04% and we could probably suppose it is noise. Now, assuming the two code paths perform identically over a long time span, I believe the only difference between the two approaches is that with the element pool we pay a small cost for recycling nodes during each frame, while without it we pay a higher cost for garbage collection in between groups of frames, possibly blocking the UI for a long time. To some degree, it's almost like the element pool amortizes the cost of GC across animation frames, making the UI responsiveness more consistent.

So I think the question is: is Chromium intelligently scheduling garbage collection when the user is not performing any task? If not, is there a way we can tune the GC to be less aggressive (https://v8project.blogspot.it/2017/02/one-small-step-for-chrome-one-giant.html)? If either one of these two questions has a positive answer I think we should go ahead and drop it, otherwise it's probably a good idea to keep it and troubleshoot the issue @kuychaco has observed on master.

maxbrunsfeld:
I think in light of the complete rewrite that we're doing of this layer (https://github.com/atom/atom/pull/13880), we can close this.

Issue #13788, opened 2017-02-10
Linux 32bit Source Build Fails Installation


progkix:
I've just built 1.14.0 & 1.15.0-beta0 32bit Linux DEBs, RPMs and GZs on Fedora 25. Builds were all successful.

Trying to install both fails with the foloowing messages:

rpm -Uvh /strg/ftp/ibnd/Atom-LnxX32-1.15.0b0.rpm
error: Failed dependencies:
	libXss.so.1()(64bit) is needed by atom-beta-1.15.0.beta0-0.1.fc25.i686

rpm -Uvh /strg/ftp/ibnd/Atom-LnxX32-1.14.0.rpm
error: Failed dependencies:
	libXss.so.1()(64bit) is needed by atom-1.14.0-0.1.fc25.i686

ungb:
Can you try to run the `script/build --install` should install Atom after building?   
(see https://github.com/atom/atom/blob/master/docs/build-instructions/linux.md#scriptbuild-options).  

We currently only have builds for 64 bit on linux for rpm builds.  We would happily accept a pr that makes the rpm package spec compatible on every architecture.

progkix:
I'm sure it will fail as well.

About a month ago, between 1.14.0-beta3 and 1.14.0-beta4, the file "resources/linux/redhat/atom.spec.in" was changed to require the 64bit version of libXss.so.1.

This effectively prevents any 32bit RPMs from being built.

/cc @as-cii

progkix:
I've looked over some of the code briefly and i think a possible solution is to provide both a 32bit and a 64bit version of the resources/linux/redhat/atom.spec.in" file. Say something like atom32.spec.in and atom64.spec.in.

The atom32.spec.in would have "Requires: lsb-core-noarch, libXss.so.1
The atom64.spec.in would have "Requires: lsb-core-noarch, libXss.so.1()(64bit)"

Then the create-rpm-package.js script could be modified to check process.arch for 32 vs 64 bit and set the value appropriately - see the line below that would need to be replicated and ajusted for the proper spec.in file.

const rpmPackageSpecsTemplate = fs.readFileSync(path.join(CONFIG.repositoryRootPath, 'resources', 'linux', 'redhat', 'atom.spec.in'))

This is right off the top of my head and with no knowledge at all of the inner working of the build scripts.

progkix:
I just build and successfully installed the latest versions (1.14.2 and 1.15.0-beta2) on Fedora 25 32bit from the source archive.

[...]

PR #13835, opened 2017-02-16
Updating requires for linux to support 32bit build

ungb:
### Description of the Change
Added logic from @progkix to support builds for 32bit linux.

### Alternate Designs

N/A

Build scripts are in core.

### Benefits

allows people to build on 32bit linux.

### Possible Drawbacks

None so far. Will test this change before merging
### Applicable Issues

https://github.com/atom/atom/issues/13788#issuecomment-279019713
/cc @progkix

### DIFF
--- a/resources/linux/redhat/atom.spec.in
+++ b/resources/linux/redhat/atom.spec.in
@@ -7,7 +7,11 @@ URL:            https://atom.io/
 AutoReqProv:    no # Avoid libchromiumcontent.so missing dependency
 Prefix:         <%= installDir %>
 
+%ifarch i386 i486 i586 i686
+Requires: lsb-core-noarch, libXss.so.1
+%else
 Requires: lsb-core-noarch, libXss.so.1()(64bit)
+%endif
 
 %description
 <%= description %>


as-cii (approved):
Thanks for working on this, @ungb!

If this works correctly on the architectures listed in the `%ifarch` clause and you have tested it across all distros that we officially support, I'd say we can go ahead and 🚢 it.

An additional test we might perform would be to check whether Travis generates a 64bit package, as it did before this pull request.

nelson6e65:
This fix works for me in `Fedora 24 TwentyFour`, Kernel: `i686 Linux 4.9.10-100.fc24.i686+PAE`:

```
$ script/build --create-rpm-package
$ sudo dnf install ./out/atom.i686.rpm
$ atom --version
Atom    : 1.14.3
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

Thanks, @progkix, @ungb

ungb:
This has been merged in https://github.com/atom/atom/pull/13842

PR #13842, opened 2017-02-17
Updating requires for linux to support 32bit build(For Testing)

ungb:
### Description of the Change
Added logic from @progkix to support builds for 32bit linux.

### Alternate Designs

N/A

Build scripts are in core.

### Benefits

allows people to build on 32bit linux.

### Possible Drawbacks

None so far. Will test this change before merging
### Applicable Issues

Creating PR off branch on master to trigger travis build.  See https://github.com/atom/atom/pull/13835 of original PR of same change.


nelson6e65:
In witch release will be this implemented?

progkix:
I just checked and it's in the 1.16.0-beta0 release that just came out

PR #13877, opened 2017-02-24
Replace language-python with MagicPython

1st1:
This is a follow up to the discussion in https://github.com/atom/language-python/issues/109. Both @50Wliu and @iolsen indicated that they are willing to replace `language-python` with **[MagicPython](https://github.com/magicstack/magicpython)**.

Atom ships with `language-python` syntax highlighter, which was backported from TextMate. It has a lot of unresolved bugs and lacks many features, such as:

* Does not highlight [f-strings](https://github.com/atom/language-python/issues/185), [`yield from`](https://github.com/atom/language-python/issues/119), [fails](https://github.com/atom/language-python/issues/150) to highlight `r` string literals properly, [etc](https://github.com/atom/language-python/issues);

* Does not highlight docstrings as a separate class of strings;

* [Breaks](https://github.com/atom/language-python/issues/68) on complex function parameters and return annotations;

* We've paid special attention to little details, such as highlighting `&&` as an invalid operator (makes switching between Python & JavaScript coding easier), properly highlighting improper syntax in decorators and many other features that make it easier to write Python code.

* MagicPython properly highlights all Python 3 builtins, magic methods, syntax features, string format qualifiers.

* We have a huge number of [tests](https://github.com/MagicStack/MagicPython/tree/master/test). This makes it easy for us to test new features and bug fixes and make releases with confidence.

In addition to the above: 

* GitHub.com [uses](https://github.com/github/linguist/blob/733ef6319378ab8c76dc72681bf2f435083dbcae/grammars.yml#L56) MagicPython to highlight Python source code.

* VSCode now [ships](https://github.com/Microsoft/vscode/tree/master/extensions/python/syntaxes) with MagicPython.

* MagicPython is already used by more than [69k](https://atom.io/packages/magicpython) Atom users, and [360k](https://marketplace.visualstudio.com/items?itemName=magicstack.MagicPython) VSCode users (360k manual installs, bundled MagicPython isn't included in this statistic).

At this point, we believe that integrating MagicPython to Atom is easier, than trying to fix `language-python`.

### Possible Drawbacks

The only possible drawback is that MagicPython does not highlight SQL syntax in Python string literals. `language-python` detects when a string starts with `SELECT`, `INSERT`, `UPDATE`, or `WITH` and delegates highlighting to an SQL syntax.

Due to some limitations of TextMate syntax format, this feature can cause highlighting bugs, such as: https://github.com/atom/language-python/issues/174#issuecomment-266842718 and https://github.com/atom/language-python/issues/143.

Another question is why we need to treat SQL specially. There are many other declarative languages (HTML, templates, etc) that people use in Python strings. 

We think that Python syntax highlighter should only highlight Python language, and highlighting of other languages within Python source code should be delegated to plugins.

### Integration

We've made some changes in MagicPython 1.0.7 specifically for Atom integration:

* backported `python-console` and `python-traceback` syntaxes;

* added generation of Jasmine spec files to make MagicPython testable under Atom, see [test/atom-spec](https://github.com/MagicStack/MagicPython/tree/master/test/atom-spec) directory.

One thing that this PR does not cover is update of `atom/docs/build-instructions/build-status.md` file. If Atom team approves this PR, we probably need to start testing MagicPython on Atom's TravisCI instance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,6 @@
     "language-perl": "0.37.0",
     "language-php": "0.37.4",
     "language-property-list": "0.9.0",
-    "language-python": "0.45.2",
     "language-ruby": "0.70.5",
     "language-ruby-on-rails": "0.25.2",
     "language-sass": "0.57.1",
@@ -172,7 +171,8 @@
     "language-todo": "0.29.1",
     "language-toml": "0.18.1",
     "language-xml": "0.34.16",
-    "language-yaml": "0.28.0"
+    "language-yaml": "0.28.0",
+    "MagicPython": "1.0.7"
   },
   "private": true,
   "scripts": {


50Wliu:
Looks like this is failing the lowercase package name check.

1st1:
> Looks like this is failing the lowercase package name check.

Yeah. Initially, MagicPython was a Sublime Text only package, and for ST camel case is the common naming convention. Would it be possible to whitelist MagicPython, or to lift this check?

Alhadis:
@1st1 Just wanna say how much I appreciate your dedication to this.  =) This is coming from somebody who actually detests Python.

@50Wliu Why is the package-name check being automated? Aren't all dependency additions reviewed by Atom members manually anyway?

In either case, `settings-view`'s [sorting issue](atom/settings-view#816) will probably need fixing:

Figure 1

50Wliu:
Well, yes, but it's not something that I'd immediately think to check for, and as you've demonstrated with Settings View, we do have code portions that rely on the package name being completely lowercase.

@iolsen what do you think is the best way to proceed?  I can think of a couple of options:
1. Remove the lowercase check.
2. Keep language-python, but make it depend on MagicPython in some way.
3. Create a dedicated Atom package, `language-magic-python` or similar.

I don't think 1) is a good idea, since basically the entire npm ecosystem uses lowercased package names (and `normalize-package-data` is maintained by an npm member, so it seems like we're following best practices), and we'd have to go through and update our sorting logic.  2) seems nice but I'm not sure how doable it is.  I like 3) the most because right now the Atom package is including some irrelevant stuff, like Sublime's YAML files and Textmate's tmLanguage files.

mnquintana:
> Create a dedicated Atom package, language-magic-python or similar.

I think this option makes a lot of sense 👍

1st1:
> 1. Remove the lowercase check.
> 2. Keep language-python, but make it depend on MagicPython in some way.
> 3. Create a dedicated Atom package, language-magic-python or similar.

> I don't think 1) is a good idea, since basically the entire npm ecosystem uses lowercased package names (and normalize-package-data is maintained by an npm member, so it seems like we're following best practices), and we'd have to go through and update our sorting logic. 

Well, this is just a package name. apm does use npm, but the end user doesn't know that. Moreover, Atom uses a separate npm repository, because plugins and themes aren't really generic JS packages. 

All end users care about is how useful the package is. As for the sorting logic—there are only a few places where packages are listed. I guess we can make a PR to fix the sorting.

> 2) seems nice but I'm not sure how doable it is. 
> I like 3) the most because right now the Atom package is including some irrelevant stuff, like Sublime's YAML files and Textmate's tmLanguage files.

We can do either, but:

1. It will make it harder for people to report bugs to the correct repo—MagicPython. We really care about support, and always try to react to issues the day they are filed. Essentially, these two options suggest to make a fork of MagicPython which, I'm afraid, can confuse users.

2. It will introduce another build/release indirection for us: implementing a new build process, keeping track of both `MagicPython` and `language-magic-python` CI status, etc.

ambv:
I don't understand the issue. The fully lowercase variant of the repo is perfectly functional as a link:

https://github.com/magicstack/magicpython

Can we simply use the name `magicpython` for APM and move on? People will find the GitHub repo just fine.

50Wliu:
I do believe the issue is specifically the [name](https://github.com/MagicStack/MagicPython/blob/b75b17bf8e3acd3be35b1cecc7eff9da0e734260/package.json#L2), which must be lowercase.

ambv:
@1st1 Can you just lowercase the package name? This is a silly blocker for such an important feature :)

1st1:
> @1st1 Can you just lowercase the package name? This is a silly blocker for such an important feature :)

I worry about consequences, specifically, if users (there are thousands of them) who installed MP manually in Atom and VSCode will continue to receive updates. Most likely, lowercasing the name in `package.json` will break updates or cause some other major inconvenience.

TBH I'm not sure how lowercase package naming policy benefits Atom users. The **ideal** solution would be to fix sorting in Atom (I can make a PR) and keep MP name as is (see also my [previous](https://github.com/atom/atom/pull/13877#issuecomment-282778332) comment).  @iolsen @50Wliu is this an option?

ambv:
Heads up, due to no progress here, Nuclide is now depending on MagicPython and disabling language-python by default. We'd love to get rid of this hack, so we're still keeping our fingers crossed on this getting merged.

iolsen:
Just a quick note that we're not ignoring this on purpose, we just haven't gotten to it yet. Hoping to in the next couple of weeks. 😬

1st1:
Ping.

ilanbiala:
Any progress on this?

iolsen:
My apologies that we haven't gotten to this sooner. We'd still like to see this change we just haven't gotten this reviewed. I'll try to address that soon.

ilanbiala:
What are the final proposed solutions? I know some were listed above, but we should just summarize and get some weigh-in on them in one place so we can decide what to do (if there is a decision to be made).

1st1:
> What are the final proposed solutions?

I believe the options are summarized here: https://github.com/atom/atom/pull/13877#issuecomment-282778332

I think that the less disruptive option is (1) -- **"Remove the lowercase check."** -- i.e. include MagicPython in Atom as a dependancy, as is. This way:

* we keep a single MagicPython repo
* whenever we issue a new version, all Atom users just receive it
* we have single point for reporting issues
* MP is already installed by 70k Atom users

Renaming MagicPython will be disruptive for existing users, which is why it's not an option. Copying-pasting MP grammar or forking the project just to make the package name lowercased, doesn't seem like a good idea either. Atom users will likely be confused about where to report issues to, and it will make harder for us to keep the fork up to date.

jkyeung:
It's disappointing that this seems stalled.

I get that both parties (Atom team and MagicPython team) have legitimate reasons for preferring different ways to handle the naming issue, but it just feels kind of silly and political to outsiders like me, who would simply like to see Atom highlight Python as properly as possible straight out of the box.

Would it help things along at all to indeed fork MagicPython into an *Atom-team-maintained* package, with billboards and blaring sirens everywhere that say "don't report your issues here, post them to MagicPython repo instead"? Seems like someone could probably write a bot that periodically either closes all issues and rejects all PRs, or forwards them all to the MagicPython repo. And also periodically grabs the latest MagicPython and just renames that one field. Maybe I'm oversimplifying, but surely *some* option for handling this (whether fixing the name sorting or whatever) *has* to be feasible.

At the very least, I have to say that it's frustrating that @iolsen has responded twice with "yes, we'd like to Get This Done" yet hasn't ever *clearly* answered @1st1's question of "do you mind if I fix your name sorting for you and keep the MagicPython name as-is?".

Maybe I'm being overdramatic, but it is getting to the point that I *almost* would rather see Atom team just say "sorry, won't fix, anyone who wants proper Python highlighting should install MagicPython separately".

Hey, maybe that's another option. Instead of making `language-python` *depend* on MP in some way (option 2 mentioned above), can we at least change its blurb from "Python language support in Atom" to "Provisional Python support in Atom - we recommend installing MagicPython instead", with more details in the readme?

C'mon, *something*?

damieng:
One option around the naming issue would be to bring MagicPython in as language-python - i.e. make it the defacto Python install.  I am a little concerned that seeing MagicPython alongside language-x is going to get people confused and be a constant source of "why is this package different"

What we did with language-typescript is basically just bring in the syntax grammar files in from upstream with a big note in the README to that affect.

This might let us get unblocked and get this in so people can start enjoying great Python highlighting.

1st1:
>  I am a little concerned that seeing MagicPython alongside language-x is going to get people confused and be a constant source of "why is this package different"

Is there a precedent for this?  Or this is just a guess?

damieng:
A precedent for confusion or a precedent for changing the format/naming of a package?

I'm on the Atom team and seeing MagicPython immediately made me think it was something quite different (additional tooling) to language-python rather than a much better version.

Alhadis:
Could bundled packages not support an Atom-specific `package.json` field which describes an alias that a package gets loaded as? It might require some lengthy refactoring, but not only would it allow `MagicPython` to be transparently loaded as `language-python`, but it would also permit future vendored packages to be used in a similar manner (if there's ever a need to, I mean).

damieng:
Technically anything is possible its just the amount of work required to get this thing to land in an already busy schedule. 

Does MagicPython provide anything beyond snippets and grammar right now?

elprans:
Meanwhile, there's a regression that's preventing users from installing MagicPython using non-lowercase name: https://github.com/atom/atom/issues/16478.

as-cii:
Hello everyone and apologies for letting this pull request sit here for so long.

We definitely see the benefits of including MagicPython into Atom and agree that language-python comes with quite a few shortcomings and bugs. Since this pull request was opened, however, we have been working on tree-sitter, a modern and fast parsing system which is meant to gradually replace first-mate grammars. Considering that Atom already ships with tree-sitter (on master and soon on beta, behind a feature flag) and that merging this pull request would require accepting some unclear tradeoffs, we have decided to not go ahead with it and simply wait until tree-sitter is fully integrated into Atom.

I would like to thank @1st1 for opening this, and to everyone else for the feedback. I apologize for closing this pull request after all the thinking and work that went into it, but hopefully people will be able to enjoy faster and awesome Python highlighting after tree-sitter becomes stable.

PR #13918, opened 2017-03-02
Cause an assertion failure if defaultMarkerLayer is destroyed early

nathansobo:
This is to investigate a case where the default marker layer of the editor is destroyed without the editor itself or its buffer being destroyed, which is causing `Cannot decorate a destroyed marker` exceptions.


nathansobo:
Looks like the settings-view test failure is spurious.

Issue #13298, opened 2016-11-22
Please provide a snap package


vincent-t:
It would be great if you could provide a snap package for atom. Instructions are here: http://snapcraft.io/.

edwinksl:
There is discussion related to snaps at https://github.com/atom/atom/issues/2956.

Ben3eeE:
Thanks @edwinksl. Closing in favor of #2956.

Issue #13118, opened 2016-10-31
Feature request: Provide an AppImage for Atom Stable and Atom Beta releases


fusion809:
Hi,

As @probonopd, the developer of the AppImage format, has requested I do this in [probonopd/AppImages#118](https://github.com/probonopd/AppImages/issues/118#issuecomment-257396828), I have decided to request that this provide provides official AppImage builds for Atom Stable and Atom Beta releases. AppImages are a cross-distribution (but very much Linux-specific, in case you're wondering whether they'd run on FreeBSD) package format that are essentially self-mounting disk images with all the required libraries and executables inside to run the app they provide (in the case of an Atom AppImage this would be Atom, of course). They are similar, in this regard, to `.dmg` (or disk image) files for macOS. I realize this project already provides tarball builds for Atom, but AppImages do not need to be extracted in order to be run, unlike tarball builds. AppImages are extremely easy to build, including from your Travis CI artefacts. 

And yes, I know that AppImages have been previously discussed in this bugtracker, see for example atom/atom#4980. But I would like to make a formal and specific request (as opposed to issue 4980, which merely requested a portable package format for Linux and did not specifically request an AppImage) for them to be provided. 

Thanks for your time,
Brenton

Ben3eeE:
Refs: https://github.com/atom/atom/issues/2956#issuecomment-255622650

probonopd:
See https://github.com/atom/atom/pull/15193

damieng:
Closing - our reasons for not supporting additional distribution formats at this time are documented in #13921

Issue #11837, opened 2016-05-25
Flatpak release support


xavier83:
### Prerequisites
- [x ] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
- [x ] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [ x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [ x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [ x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.
### Description

Need release of linux sandboxed apps [Flatpack](http://flatpak.org/).
### Steps to Reproduce
1. Goto atom release page https://github.com/atom/atom/releases/

**Expected behavior:** [What you expected to happen]
Find flatpak atom package.

**Actual behavior:** [What actually happened]
Can't find flatpak atom package.
### Versions

You can get this information from executing `atom --version` and `apm --version` at the command line. Also, please include the OS and what version of the OS you're running.

jokeyrhyme:
Ubuntu has "click" packages:

- https://developer.ubuntu.com/en/publish/packaging-click-apps/

For everyone else (including Ubuntu):

- http://flatpak.org/

LibreOffice is now available via Flatpak:

- https://whatofhow.wordpress.com/2016/06/01/libreoffice-5-2-beta-flatpak/

Although there are some caveats:

> Flatpak-based apps are not yet able to reach out to other applications installed on the machine (like browsers) to e.g. open URLs through them. That means that e.g. clicking on a hyperlink in a LibreOffice Writer document might not work yet.

[...]

Issue #11000, opened 2016-02-29
Building Atom Beta releases so they can be installed simultaneously, on the same machine, to stable releases


fusion809:
Hi,

I realize you's are probably getting sick of me by now, but I was wondering if there is some option one can pass to `script/build` or `script/grunt install` to build Atom beta with different installed file paths, so that Atom beta and Atom stable releases can be installed simultaneously on the same machine. See running `script/build && sudo script/grunt install` in the beta branch of this repo will install Atom in the standard file locations (that are also used by the stable releases of Atom) on Linux platforms, at least, as those used by stable Atom releases. So you will get `/usr/bin/atom`, `/usr/bin/apm`, _etc._ files installed at those exact locations, while the Debian and RPM packages provided by this repo for Atom beta releases will install the same files to `/usr/bin/atom-beta`, `/usr/bin/apm-beta`, _etc._ instead and let Atom stable releases use the aforementioned installed file locations (i.e., `/usr/bin/atom`, `/usr/bin/apm`, _etc._) for their installed files. 

This issue means that presently the `atom-editor` and `atom-editor-beta` PKGBUILDs in the Arch User Repository cannot be installed simultaneously, even though `atom-editor` and `atom-editor-beta-bin` can be installed simultaneously because `atom-editor-beta-bin` builds its package from the Debian binary provided for Atom beta releases. 

Thanks for your time,
Brenton

PR #13921, opened 2017-03-02
Generate AppImage

probonopd:
### Requirements

* Linux 64-bit machine

### Description of the Change

This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload an [AppImage](http://appimage.org/) to a temporary download URL on transfer.sh (available for 14 days). The download URL is toward the end of each Travis CI build log of each build. __Please note:__ Instead of storing AppImage builds temporarily for 14 days each on transfer.sh, you could use GitHub Releases to store the binaries permanently. In this case, delete [this line](https://github.com/atom/atom/pull/13921/files#diff-354f30a63fb0907d4ad57269548329e3R33) from the `.travis.yml` file.

### Alternate Designs

* Flatpak: Requires the setup of Flatpak and a runtime first, need to use the command line. Does not work out-of-the-box on most distributions but Fedora. Does not run on Live CDs.
* Snap: Canonical centric solution, Requires the setup of Snap and a runtime first, does not work out-of-the-box on most distributions but Ubuntu.

Linux should be a first-class desktop for Atom, and using Atom on Linux should be as easy as on Windows and macOS.

### Benefits

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Works for most Linux distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Just one format for all major distributions
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

### Possible Drawbacks

None. Providing an AppImage could be done in addition to providing other types of Linux packages and/or bundles.

### Applicable Issues

* https://github.com/atom/atom/issues/13118: Feature request: Provide an AppImage for Atom Stable and Atom Beta releases
* https://github.com/atom/atom/issues/4980: Add portable version for Linux
* https://github.com/atom/atom/issues/11000: Building Atom Beta releases so they can be installed simultaneously, on the same machine, to stable releases
* https://github.com/atom/atom/issues/4330: Installing atom without sudo privileges
* https://github.com/atom/atom/issues/2956: Add Linux software repositories for app update support 
* https://github.com/atom/atom/issues/11837: Flatpak release support
* https://github.com/atom/atom/issues/13298: Please provide a snap package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,21 @@ install:
   - nvm use --delete-prefix $NODE_VERSION
   - npm install -g npm
   - script/build --create-debian-package --create-rpm-package --compress-artifacts
+  - find /tmp/atom-*/
+  - cd /tmp/atom-*/
+  - wget -c https://github.com/probonopd/AppImageKit/releases/download/continuous/AppRun-x86_64 -O AppRun
+  - chmod a+x ./AppRun
+  - cp usr/share/applications/atom.desktop .
+  - cp usr/share/pixmaps/atom.png .
+  - cd -
+  - wget -c "https://github.com/probonopd/AppImageKit/releases/download/8/appimagetool-x86_64.AppImage" 
+  - chmod a+x appimagetool*.AppImage
+  - ./appimagetool*.AppImage --appimage-extract
+  - export VERSION=git.$(git rev-parse --short HEAD)
+  - ./squashfs-root/AppRun /tmp/atom-*/
+  - curl --upload-file ./Atom-*.AppImage https://transfer.sh/Atom-$VERSION-x86_64.AppImage # DELETE THIS LINE
+  - mv Atom-*.AppImage out/
+  - ls -lh
 
 script:
   - script/lint
@@ -42,6 +57,7 @@ addons:
       - out/atom-amd64.deb
       - out/atom.x86_64.rpm
       - out/atom-amd64.tar.gz
+      - out/Atom-*.AppImage
     target_paths: travis-artifacts/$TRAVIS_BUILD_ID
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


probonopd:
Ping. Asking for a review.

damieng:
Hi.

At this time we have no resources available to support additional distribution formats beyond the ones for Atom on Linux we currently maintain (deb and rpm).

probonopd:
Hi @damieng I understand that your resources are limited. I hope you have considered that I have essentially already done the work, so this is a "turnkey" solution...

damieng:
The problem is adding .AppImage is just the tip of the iceberg. We'd have to understand it to be able to support it as well as be able to maintain and fix it should it break, answer questions and problems with it's usey, be able to reproduce issues on a variety of distributions that use it...

Serkan-devel:
@damieng Is there any progress on this?

damieng:
@Serkan-devel there is no progress as we aren't pursuing it as per my previous response;

> We are not adding any additional distribution formats at this time due to resource constraints in supporting them.

probonopd:
Come on, I have already done the work...

barni2000:
@damieng You already provide binary tar.gz for linux which is very similar to AppImage and will cause same issues on different distributions.

I am very unhappy about your decisions.

igorlogius:
Did anything change in regard to providing an *.AppImage file for distribution? 
- Just asking since there are now quite a few applications which support this format. (ref. https://appimage.github.io/apps/  or https://www.appimagehub.com/browse)

Issue #13878, opened 2017-02-25
Proposal: User-Controlled View Locations (Workspace Views)


matthewwithanm:
Currently, each package that defines a UI tool is responsible for creating and managing its own panels. This means that (1) there's not consistent UI for toggling panels and (2) users don't have any control over the placement of their tools. In Nuclide, we've tackled this with a feature called “workspace views.” It has two main goals:

* Provide a way for package authors to make location-agnostic views.
* Provide containers for displaying these views.

We'd like to get this functionality into Atom core 😊

![wspv3b](https://cloud.githubusercontent.com/assets/126263/23325856/6786fd74-faad-11e6-926c-3b51838c3949.gif)

Nuclide's workspace views package is based on a lot of the abstractions and concepts already in Atom:

* It exposes an `addOpener()` API for registering model factories.
* It exposes an `open()` function for showing them.
* It creates three panels (left, right, and bottom), each with its own PaneContainer and tab bar.

This allows interop with the workspace pane container (i.e. users can move things into the main workspace when they want it to appear larger) and lends itself to adoption into Atom by extending existing APIs.

# Proposal

We propose extending the existing `atom.workspace.open()` API to support multiple “locations,” including each of the panels. `addOpener()` shouldn't require any changes, but there are some workspace-level which would be helpful to add:

### `destroyItemWhere(predicate: (item: Object) => boolean): void`

Package authors can use this to destroy all of the views they've created when their package is deactivated.

### `toggle(uri: string, options?: Object): void`

Allows toggling visibility of specific items. If a matching item is already visible in a panel, that panel gets hidden.

So far, we haven't had a reason to require inspection of the panels, so we don't have any analogues to `getPanes()`, etc. but it probably makes sense for hackability.

## Model Additions

As far as the view model goes, we followed the precedent set by `getTitle()`, `getIconName()`, et. al. and added a few optional methods:

### `getDefaultLocation?: () => string`

With Nuclide's workspace views, users decide where to place their views (as opposed to package authors specifying a location when they call `open()`). If they move the file tree to the right panel, that location will get serialized and used the next time they open the file tree. This method gives package authors to specify the default location. Allowed locations are “left-panel”, “right-panel”, “bottom-panel”, and “pane” (which corresponds to the workspace's pane container). Nuclide also has an API for registering custom locations, but we probably don't need to support this in Atom at this point.

### `getPreferredInitialHeight?: () => number`

[...]

PR #13922, opened 2017-03-03
Convert workspace to JavaScript

matthewwithanm:
This converts `workspace.coffee` and `workspace-spec.coffee` to JavaScript. I split it into three different commits for review purposes:

- The first is just the automated decaffeinate step and probably doesn't need to be reviewed.
- The second is tweaking to make the linter happy. I tried to avoid any real changes here. (The one exception being `TestItemElement` in the spec bc of the difference bt CS and JS classes)
- The third is the real stuff. It cleans things up, removes implicit returns, and generally makes things a little more idiomatic.


spec/workspace-spec.js
```diff
@@ -0,0 +1,2072 @@
...
+        notificationManager: atom.notifications,
+        applicationDelegate: atom.applicationDelegate,
+        viewRegistry: atom.views,
+        assert: atom.assert.bind(atom),
+        textEditorRegistry: atom.textEditors
+      })
+      return atom.workspace.deserialize(workspaceState, atom.deserializers)
+    }
+
+    describe('when the workspace contains text editors', () =>
```
maxbrunsfeld: While you're here, could you make all the `describe` callbacks use curly braces? It can be kind of confusing when you go to add an `it` to some `describe` if the `describe` uses the single-expression syntax.

spec/workspace-spec.js
```diff
@@ -0,0 +1,2072 @@
...
+        )
+
+        runs(() => {
+          pane3.activeItem.setCursorScreenPosition([2, 4])
+          pane4 = pane2.splitDown()
+        })
+
+        waitsForPromise(() =>
+          atom.workspace.open('../sample.txt').then(editor => pane4.activateItem(editor))
+        )
```
maxbrunsfeld: Are you interested in switching this test to use `async` and `await`? If you feel like moving on, no pressure, but it can be kind of satisfying to :fire: all this `waitsForPromise` and `runs`.

maxbrunsfeld (commented):
Thanks so much @matthewwithanm! 📈

spec/workspace-spec.js
matthewwithanm: oops thought i did! yeah i'll do that now since i'm rebasing past today's commits 😩

spec/workspace-spec.js
matthewwithanm: yeah, i'd like to keep moving on #13878 if that's okay. especially bc this pr is so prone to merge conflicts!

maxbrunsfeld: :+1:

PR #13927, opened 2017-03-03
fix logLongestSuites

UziTech:
### Description of the Change

Fix logLongestSuites in time-reporter

### DIFF
--- a/spec/time-reporter.coffee
+++ b/spec/time-reporter.coffee
@@ -13,7 +13,7 @@ class TimeReporter extends jasmine.Reporter
     window.logLongestSuites = (number) => @logLongestSuites(number)
 
   logLongestSuites: (number=10, log) ->
-    return unless window.timedSuites.length > 0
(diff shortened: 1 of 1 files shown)


UziTech:
still in needs-review

rsese:
Sorry for the delay @UziTech - we talked about the PRs you pinged us on and for this one were hoping you can add more detail about:

* the problem itself and its impact.
* your verification process as described in the current PR templates.

This will be helpful for whoever ends up reviewing the PR.

UziTech:
The issue is `window.timedSuites` is an object not an array so checking it's `.length` doesn't work.

I'm not sure if this actually has any impact since it never actually worked and I don't know of anyone (including atom) that actually uses this file.

A better fix might be deleting the file.

I don't remember how I stumbled upon this issue. I think it was when I was creating [atom-jasmine3-test-runner](https://www.npmjs.com/package/atom-jasmine3-test-runner)

PR #13929, opened 2017-03-04
Fix destroying a `PanelContainer` containing multiple panels

as-cii:
Previously, when calling `destroy` on a `PanelContainer` containing multiple panels, Atom would throw a `Cannot read property 'destroy' of undefined` exception (https://circleci.com/gh/atom/atom/2832). This was due to iterating over the panels while at the same time destroying them, which caused the iterated array to be modified during the loop.

With this commit we slice the array before iterating over it so that destroying a `PanelContainer` doesn't throw exceptions anymore.

### DIFF
--- a/spec/panel-container-spec.coffee
+++ b/spec/panel-container-spec.coffee
@@ -5,7 +5,7 @@ describe "PanelContainer", ->
   [container] = []
 
   class TestPanelItem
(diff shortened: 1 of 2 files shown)


nathansobo:
🙇

Issue #13873, opened 2017-02-23
Data loss scenarios with opening project with atom in empty state  


ungb:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * [x] Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * [x] Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * [x] Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * [x] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom 
        * There's a lot of issues filed for this already :) this is a issue to consolidate the root issues.  
    * [x] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

This is to capture a few data loss scenarios with opening your project.  

There are 3 bugs that we can see so far that will cause data loss

### **Opening a Empty Document and try to `Open` the folder from context menu.**

  Repro:

  1. Open Atom, go to settings, check `Open Empty Editor On Start` and uncheck `Restore Previous Windows on Start` 
  2. Open a folder, add a few untitled documents or make changes to file and don't save anything.
  3. Close Atom
  4. Open Atom, you should get a blank editor
  5. Go to context menu and click `Open` (or cmd+o on mac) and open the folder from step 2.  

Expected: Restores Session
Actual: Unsaved session is lost.

### **Opening a Empty Document and trying to `Add a Project Folder` from context menu**

  Repro:

  1. Open Atom, go to settings, check `Open Empty Editor On Start` and uncheck `Restore Previous Windows on Start` 
  2. Open a folder, add a few untitled documents or make changes to file and don't save anything.
  3. Close Atom
  4. Open Atom, you should get a blank editor
  5. Go to context menu and click `Add a Project Folder...` (or cmd+shift+o on mac) and open the folder from step 2.  

Expected: Since there isn't any folder added yet, Atom should restore the session of that folder.
Actual: Unsaved session is lost. 

    
### **Open `atom .` when there is a empty atom editor open.**
Repro:
  1. Open Atom, go to settings, check `Open Empty Editor On Start` and uncheck `Restore Previous Windows on Start` 
  2. Open a folder, add a few untitled documents or make changes to file and don't save anything.
  3. Close Atom
  4. Open Atom, you should get a blank editor
[...]

PR #13947, opened 2017-03-07
Make an option to always restore the last session

anatoli26:
Make an option to *always* restore the last session, no matter how Atom is invoked (#9643).

### Description of the Change

`src/config-schema.js`: the change adds the option to the Settings window.

`src/main-process/atom-application.coffee`:
- line 114: load previous session if the option is turned on;
- line 563: IMO it's a bug, here the newly created window is activated, but `@lastFocusedWindow` is not updated, so on the second invocation at the same moment in time there'll be no lastFocusedWindow and a second (new) window would be created (at line 542/544 -> 550/552).
- line 604: if any of the options is checked, restore the last session (applicable when starting atom without arguments)
- line 817: a function to duplicate an object, taken from: https://coffeescript-cookbook.github.io/chapters/classes_and_objects/cloning

### Alternate Designs

TL;DR: none, see the issue #9643 for an extended discussion.

This is a standard behavior for most of modern multi-tab apps today, i.e. the ability to restore the last session after crash or normal exit, no matter how the app is invoked for the first time since the last session (i.e. with or without path parameters).

### Benefits

See the previous point.

### Possible Drawbacks

None as this is an option turned off by default, so it won't affect anyone if not explicitly terned on by the user.

### Applicable Issues

So far no issues were discovered.

### Other

script/test: 23 passing (2m), i.e. no errors when the option is turned off (the default behavior)
Jasmine specs in the ./spec folder are not provided

atom_settings

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -258,14 +258,10 @@ const configSchema = {
         default: true
       },
       restorePreviousWindowsOnStart: {
-        description: 'When checked restores the last state of all Atom windows when started from the icon or `atom` by itself from the command line; otherwise a blank environment is loaded.',
-        type: 'boolean',
-        default: true
-      },
-      restorePreviousWindowsOnStartAlways: {
-        description: 'When checked *ALWAYS* restores the last state of all Atom windows.',
-        type: 'boolean',
-        default: false
+        type: 'string',
+        enum: ['no', 'yes', 'always'],
+        default: 'yes',
+        description: "When selected 'no', a blank environment is loaded. When selected 'yes' and Atom is started from the icon or `atom` by itself from the command line, restores the last state of all Atom windows; otherwise a blank environment is loaded. When selected 'always', restores the last state of all Atom windows always, no matter how Atom is started."
       },
       reopenProjectMenuCount: {
         description: 'How many recent projects to show in the Reopen Project menu.',
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -262,6 +262,11 @@ const configSchema = {
         type: 'boolean',
         default: true
       },
+      restorePreviousWindowsOnStartAlways: {
+        description: 'When checked *ALWAYS* restores the last state of all Atom windows.',
+        type: 'boolean',
+        default: false
+      },
       reopenProjectMenuCount: {
         description: 'How many recent projects to show in the Reopen Project menu.',
         type: 'integer',
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -112,7 +112,7 @@ class AtomApplication
   launch: (options) ->
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test or options.benchmark or options.benchmarkTest
       if @config.get('core.restorePreviousWindowsOnStart') is 'always'
-        @loadState(clone(options))
+        @loadState(_.cloneDeep(options))
       @openWithOptions(options)
     else
       @loadState(options) or @openPath(options)
@@ -814,10 +814,3 @@ class AtomApplication
       args.push("--resource-path=#{@resourcePath}")
     app.relaunch({args})
     app.quit()
-
-  clone = (obj) ->
-    return obj if obj is null or typeof (obj) isnt "object"
-    temp = new obj.constructor()
-    for key of obj
-      temp[key] = clone(obj[key])
-    temp
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -111,6 +111,8 @@ class AtomApplication
 
   launch: (options) ->
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test or options.benchmark or options.benchmarkTest
+      if @config.get('core.restorePreviousWindowsOnStartAlways')
+        @loadState(clone(options))
       @openWithOptions(options)
     else
       @loadState(options) or @openPath(options)
@@ -558,6 +560,7 @@ class AtomApplication
       windowDimensions ?= @getDimensionsForNewWindow()
       openedWindow = new AtomWindow(this, @fileRecoveryService, {initialPaths, locationsToOpen, windowInitializationScript, resourcePath, devMode, safeMode, windowDimensions, profileStartup, clearWindowState, env})
       openedWindow.focus()
+      @lastFocusedWindow = openedWindow
 
     if pidToKillWhenClosed?
       @pidsToOpenWindows[pidToKillWhenClosed] = openedWindow
@@ -598,8 +601,7 @@ class AtomApplication
       @storageFolder.storeSync('application.json', states)
 
   loadState: (options) ->
-    restorePreviousState = @config.get('core.restorePreviousWindowsOnStart') ? true
-    if restorePreviousState and (states = @storageFolder.load('application.json'))?.length > 0
+    if (@config.get('core.restorePreviousWindowsOnStartAlways') or @config.get('core.restorePreviousWindowsOnStart')) and (states = @storageFolder.load('application.json'))?.length > 0
       for state in states
         @openWithOptions(Object.assign(options, {
           initialPaths: state.initialPaths
@@ -812,3 +814,10 @@ class AtomApplication
       args.push("--resource-path=#{@resourcePath}")
     app.relaunch({args})
     app.quit()
+
+  clone = (obj) ->
+    return obj if obj is null or typeof (obj) isnt "object"
+    temp = new obj.constructor()
+    for key of obj
+      temp[key] = clone(obj[key])
+    temp
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -15,7 +15,6 @@ net = require 'net'
 url = require 'url'
 {EventEmitter} = require 'events'
 _ = require 'underscore-plus'
-cloneDeep = require 'lodash.clonedeep'
 FindParentDir = null
 Resolve = null
 
@@ -113,7 +112,7 @@ class AtomApplication
   launch: (options) ->
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test or options.benchmark or options.benchmarkTest
       if @config.get('core.restorePreviousWindowsOnStart') is 'always'
-        @loadState(cloneDeep(options))
+        @loadState(_.deepClone(options))
       @openWithOptions(options)
     else
       @loadState(options) or @openPath(options)
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -111,7 +111,7 @@ class AtomApplication
 
   launch: (options) ->
     if options.pathsToOpen?.length > 0 or options.urlsToOpen?.length > 0 or options.test or options.benchmark or options.benchmarkTest
-      if @config.get('core.restorePreviousWindowsOnStartAlways')
+      if @config.get('core.restorePreviousWindowsOnStart') is 'always'
         @loadState(clone(options))
       @openWithOptions(options)
     else
@@ -601,7 +601,7 @@ class AtomApplication
       @storageFolder.storeSync('application.json', states)
 
   loadState: (options) ->
-    if (@config.get('core.restorePreviousWindowsOnStartAlways') or @config.get('core.restorePreviousWindowsOnStart')) and (states = @storageFolder.load('application.json'))?.length > 0
+    if (@config.get('core.restorePreviousWindowsOnStart') in ['yes', 'always']) and (states = @storageFolder.load('application.json'))?.length > 0
       for state in states
         @openWithOptions(Object.assign(options, {
           initialPaths: state.initialPaths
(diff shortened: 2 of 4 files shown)


src/config-schema.js
```diff
@@ -262,6 +262,11 @@ const configSchema = {
         type: 'boolean',
         default: true
       },
+      restorePreviousWindowsOnStartAlways: {
+        description: 'When checked *ALWAYS* restores the last state of all Atom windows.',
```
BrunoBernardino: I'd only add `, even if opening a single file with Atom` right before the full stop.

anatoli26: @BrunoBernardino, actually, it could be more than one file, like: `atom fileX fileY fileZ` and it will restore the last session + there will open the 3 files, so if the short "just always" version is not quite clear, maybe add at the end ", no matter how atom is started".

lee-dohm:
This shouldn't be a separate option from `core.restorePreviousWindowsOnStart`. `core.restorePreviousWindowsOnStart` should become a three-state option: `false`, `true` and `always`, and the description text should be updated to reflect what each means.

src/config-schema.js
BrunoBernardino: I see your point. I think your version is better then.

anatoli26:
Done. Also updated the test to reflect the new set of values.

atom

src/main-process/atom-application.coffee
```diff
@@ -812,3 +814,10 @@ class AtomApplication
       args.push("--resource-path=#{@resourcePath}")
     app.relaunch({args})
     app.quit()
+
+  clone = (obj) ->
```
steelbrain: Maybe use [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or [`_.clone`](https://lodash.com/docs/4.17.4#clone) instead of reinventing clone? :)

anatoli26: Here a recursive clone is needed, so neither of these functions will work, but maybe [`_.cloneDeep`](https://lodash.com/docs/4.17.4#cloneDeep) could be used.

jsolisu:
Please check *always* option, it does not work well, Atom hangs on Windows 10.

anatoli26:
How do you start it? What do you get with 'yes' option? I don't have windows 10 to test, any other OS that manifests the same behavior?

jsolisu:
@anatoli26 *yes* option is ok. You can test using "Open with Atom" (as folder) from Windows Explorer. I only use Windows, sorry.

anatoli26:
If you "Open with Atom" a single file what happens? If you run it from the command line like `atom <path_to_file>`? Do you have access to Win7/8 to test it?

What do you mean by hangs? Complete freeze, atom process has no activity in task manager? Or does it process something (e.g. high CPU usage, disk reads, etc.)?

jsolisu:
- Using "Open with Atom" with a folder, atom process has no activity in task manager.
- Using "Open with Atom" with a single, fails too.
- Using command line atom <path_to_file>, fails too.

In all cases, if you try again (Without killing the frozen process) it works.

Calling Atom directly (using Icon) works well.

Sorry. I don´t have Win7/8 to test.

anatoli26:
Try removing node_modules folder in the root of the repo and rebuild the package again.

anatoli26:
The problem is actually with `cloneDeep` from `lodash` not available as `_` is defined as `underscore-plus` package, not `lodash`, so `_.cloneDeep` is searched there and logically is not found, and when Atom is not running in foreground (with `-f` flag) it doesn't show errors, but silently stops (hangs) in the middle of initialization, that's why it was appearing as hanged to @jsolisu.

In this case I'm not sure if it makes sense to include lodash (looks like it's not included otherwise) or to define own deepclone function as it was before the latest commit. @steelbrain, any suggestions?

steelbrain:
@anatoli26 Whenever I need a lightweight lodash plugin that I don't wanna install the whole thing, I do `npm install lodash.clonedeep` and yes, these are  official packages by the lodash team

anatoli26:
OK, just defined the `cloneDeep` function as `cloneDeep = require 'lodash.clonedeep'` and added a dependency for it in `package.json`.

@jsolisu, please test it again with the latest commit.

50Wliu:
I think we'd prefer not adding the lodash dependency and just going with what you had previously (/cc @atom/maintainers).

@steelbrain my guess would be that lodash didn't exist when Atom was being built.  Underscore was released in 2009; lodash in 2012.

nathansobo:
If we need a *single* function from lodash and it's not trivially implemented in another way or already available in our existing dependencies, I'm comfortable pulling in just that function. In general I want to minimize additional dependencies.

50Wliu:
Also, I just checked and underscore-plus does offer a deep clone, `_.deepClone`. https://github.com/atom/underscore-plus/blob/4a022cf721c2561d20890c5b7faf5e9ba4832a7b/src/underscore-plus.coffee#L98

nathansobo:
Then we definitely shouldn't add a new dependency for something we already have.

anatoli26:
> underscore-plus does offer a deep clone

Haven't seen it due to its reverse-name. Sure now it's much simpler, just had to change the name. No dependency on lodash now.

`src/config.coffee` should probably also use `_.deepClone` from underscore-plus (it implements its own function now (line 999)).

50Wliu:
> src/config.coffee should probably also use _.deepClone from underscore-plus

Can you create a new issue for that so we don't forget about it?

damieng:
I don't see how this is going to solve the problem in the issue linked.

The underlying problem there isn't that Atom doesn't restore your last project (it does) - it's that opening a random file effectively creates a new project that is now your last project.

How does this address that - I can't tell from the diff.

anatoli26:
> Can you create a new issue...

#13980

@damieng, there's a mix of use cases in #9643, most of which are solved with this PR.

The original post said:

> If I have any added project folders they are successfully restored every time I fire up Atom. However, when I choose to open a single file or a directory using the right click menu, all added project folders and open file tabs are removed.

This is the actual behavior with the option set to 'yes', i.e. when Atom is started _without_ arguments, it restores the previous session with all folders and tabs. When it's started _with_ arguments, it doesn't restore the previous session. So the user then said:

> I believe **adding** a folder/file would be much more intuitive

This is the behavior with the option set to 'always', i.e. when Atom is started with arguments, it restores the previous session and _adds_ the files (requested via arguments) to the restored session.

Then other users say:

> Yes this is very annoying. I'm forever re-adding all my project folders.

> This can only be a blatant bug. This makes no sense. Now I have to go add the project folder back---and I'll have to do it every time I use Atom to quickly edit a file on the fly. Which means I won't use Atom to edit files on the fly, and I'll stick with EmEditor or something else.

>Very unhappy about this behaviour as well. I am constantly re-adding my project folders which is very time consuming.

What most users were reporting is that opening Atom with arguments was forgetting the previously opened folders and tabs, so the users had to reopen them manually. What the 'always' option does is that the project folders and open tabs are never lost.

If you just seen the code, I suggest you try to build Atom with this PR and compare the behavior with 'yes' and 'always'. You'll see that with 'always' Atom never forgets the previous folders/tabs.

jsolisu:
@anatoli26 With _.deepClone Atom does not freeze, but now it creates two instances.

anatoli26:
If you mean that it creates 2 windows when you open a _folder_, it's by design. I.e. if you want to open a folder that would be added to the restored session in the same window, you should pass `-a` (`--add`) param (e.g. `atom -a <path_to_a_folder>`). Otherwise, it'll restore the previous session in one window and open a new window for the requested folder. It's the same behavior as if Atom is opened and you open a folder: the folder will be opened in a new window.

If you open _file(s)_, it(they) will be added to the same window where the previous session is restored. It's possible to make some changes to open a new _folder_ always in the same window, but I guess it's not what others would expect.

Please let me know if my explanation is not clear.

jsolisu:
@anatoli26 everything is clear. Thanks.

anatoli26:
Should anything else be done to integrate the PR to the master?

iolsen:
@ungb let's do some testing on this. We can discuss this afternoon.

iolsen:
We have a parallel PR here https://github.com/atom/atom/pull/13963 that should get finished this week. At that time we'll figure out how to handle this one.

/cc @BinaryMuse @kuychaco

BrunoBernardino:
Any updates on this? I just got hit hard by this one (I try to not use Atom for single files, but opened one by mistake and "lost" 15 projects on the tree view).

damieng:
@BrunoBernardino They're not lost - you can get them back just by using Reopen Project from the File menu - it will have all 15 already in there.

anatoli26:
@iolsen, I see that the #13963 was merged, any progress on this PR?

ungb:
@anatoli26,  I'll spend some time today testing this and get back to you! 

I kicked off the builds again since they seem to of failed.

ungb:
Hey @anatoli26,


Sorry took so long to get back to you,  I wanted to understand the issues and try to better understand this fix.  I think that the main issue was a combination of a few issues that have been resolved since.
https://github.com/atom/atom/issues/13873 - Resolved by pr https://github.com/atom/atom/pull/13963
Reopen project Feature: https://github.com/atom/atom/pull/13046


From testing, `always` and `yes` behave the same when opening atom by either `atom` on cmd line/terminal.  

**On Windows**

***Opening file from file explorer***
1. Open Project on Atom and Open a few changes
2. Go to settings and set `Restore Previous Windows` on Start to `Always`
3. Close Atom 
4. Open a file with atom from file explorer.

**Result:** Atom opens the previous open project and opens the file on a new tab.


***Opening folder from file explorer***

1. Open Project on Atom and Open a few changes
2. Go to settings and set `Restore Previous Windows` on Start to `Always`
3. Close Atom 
4. Open Powershell and open a folder with `atom .`

**Result:** Atom opens the previous open project and opens a new atom window with the folder you opened in step 4



**On Mac**

***Opening file from file explorer***
1. Open Project on Atom and Open a few changes
2. Go to settings and set `Restore Previous Windows` on Start to `Always`
3. Close Atom via red x(notice atom application is still running)
4. Open a file with atom from finder

**Result:** Atom opens a window with the project of the file open on step 4, the previous project doesn't open.
Note: The behavior is the same as windows when you completely exit out of Atom, but closing atom via the red x on mac doesn't completely exit out of Atom.


***Opening folder from file explorer***

1. Open Project on Atom and Open a few changes
2. Go to settings and set `Restore Previous Windows` on Start to `Always`
3. Close Atom via red x(notice atom application is still running)
4. Open terminal and open a folder with `atom .`

**Result:** Atom opens a window with the project from step 4, the previous project isn't open.
Note: The behavior is the same as windows when you completely exit out of Atom, but closing atom via the red x on mac doesn't completely exit out of Atom.

On Ubuntu, I saw the same behavior as Windows.



Overall, I think this is a improvement and gives people the flexibility to open single files on their current session.  The only thing is mac behavior is different when you close all windows, but the application is still running.

/cc @iolsen @nathansobo  for code review.

anatoli26:
Bryant, thanks for this thorough review.

The issues and PRs you mentioned, *if I understand them correctly*, are actually for the behavior when the session is NOT restored, i.e. when the current option is set to NO. Aren't they a compliment to this PR?

What this PR is trying to solve is to implement a behavior typical in most multi-tab programs, like browsers, editors, etc.: when you configure the previous session to be restored, it is restored no matter how the program is started, i.e. alone or opening additional files. E.g. if Firefox is instructed to restore previous session, when we start it (say after a reboot) by clicking on a link in a Thunderbird mail, Firefox opens the previous session and creates a new tab for the requested link. The current behavior of Atom is to ignore a previous session if a link is clicked (i.e. a new file is opened), but it restores the previous session if it's started without opening anything in addition.

With respect to Mac, I'm not familiar with Atom on Mac, so I can't say much. From you description it looks like the main process is still running, and the X just makes the windows disappear, so when new windows are created, the initialization logic is not executed. One could say that this particular case doesn't strictly apply to this option as it says "on start", but if Atom is still running, you're actually not starting it.

In any case, if the users will ask to extend this feature to that particular case too, we could investigate the state in which Atom is after closing all windows on Mac and find a solution to restore the previous session for this case too. Does it sound reasonable?

ungb:
yes, the issues that are fixed, make it possible to open old sessions, and not lose data while doing it.  In that sense they are compliments to this PR, and yes the data loss issue occurs when you have a atom window without a project.

The Reopen project Feature: #13046, makes it possible to easily open recent projects, but I do agree this change is a improvement. 

I think the current behavior on mac where Atom application is still open is the expected behavior in this case.  I've compared this behavior with editors like sublime text behave the same way.  If you close all your sessions and click on the dock to open Atom/Sublime, a empty window will open and not your previous session.  So I think on mac, the moment you close all your windows, your current sessions is now empty so the expected behavior is to open the file project, or just the folder project if opened with `atom .`  

I'll follow up on this PR to see if someone can take a 👀

BrunoBernardino:
@ungb totally correct. The big difference between something like Sublime and Atom there is that _after_ you've opened a single file and closed the app, when you open the app again, you have the projects right there. With Atom we currently have to do the "Reopen project and re-collapse all projects on the sidebar one-by-one" dance, which isn't very fun 😢 . While this isn't the most clean solution, I feel like it's an improvement, even if not the same as how other editors behave.

nathansobo:
@ungb If you can confirm this functions as expected, the code looks good to me.

PR #13949, opened 2017-03-07
Minimize synchronous I/O during startup

as-cii:
When Atom boots up it is extremely crucial to perform as little synchronous operations as possible in order to minimize startup time. This pull request does so by:

1. Exposing load settings to `BrowserWindow`s as JSON.

    When accessing objects in the main process via the `remote` module, Electron returns proxy objects that are references to the original ones. This means that trying to access a remote object's property or function results in a synchronous message exchange with the main process. <br>

    In Atom core we frequently access the load settings coming from the main process, especially during startup. This caused a lot of synchronous I/O which was blocking the renderer process for several milliseconds.
  
    With this pull request, instead of exposing load settings as a JavaScript object, we serialize them to JSON in the main process and parse them back to a JavaScript object in the renderer processes. This allows us to get a full copy of the object locally and pay for I/O only once when retrieving load settings from the main process for the first time.

2. Disabling zoom in the main process rather than in the renderer process.

    With this pull request we will register the `display-added` and `display-removed` events only once in the main process in order to disable zoom (see
    #11345) directly instead of unnecessarily paying for I/O in the renderer process during startup.

3. Registering enter/leave fullscreen events in the main process rather than in the renderer process.

    This will still notify render processes when such events are triggered without, however, incurring the additional cost of synchronously retrieving a `BrowserWindow` (and its properties) via `remote` during startup.

4. Replacing `localStorage` with `StateStore` in `HistoryManager`.

    Instead of using `localStorage` to store and retrieve the project history, with this pull request we will use `StateStore` so that we can retrieve state asynchronously without blocking Atom during startup. @damieng: I think this slightly changes the public `clearProjects` API by making it asynchronous. Considering the recent introduction of the API and its low adoption (I couldn't find anyone using it in packages), I believe this might be negligible; if you have any concern about it, though, please don't hesitate to raise them up.

These improvements are particularly evident during the construction of an `AtomEnvironment` instance. This is how a typical chart of that code path looks like today on master:

screen shot 2017-03-07 at 11 10 00

After removing all the synchronous I/O, the same code path looks like the following:

screen shot 2017-03-07 at 11 05 29

On average, with these changes we save `~ 40ms` for the construction of `AtomEnvironment`. Moreover, the experience should be snappier across all the code paths because we won't pay any additional cost to retrieve load settings and to save project history state anymore. 

The next step will involve building an `AtomEnvironment` instance right inside the snapshot (as opposed to doing so at runtime) and thus potentially save 80-90% of the remaining `~ 80ms` observed after the changes introduced by this pull request.

/cc: @atom/maintainers

### DIFF
--- a/spec/history-manager-spec.js
+++ b/spec/history-manager-spec.js
@@ -30,7 +30,7 @@ describe("HistoryManager", () => {
       return projectDisposable
     })
 
-    historyManager = new HistoryManager({stateStore, project, commands: commandRegistry})
+    historyManager = new HistoryManager({stateStore, localStorage: window.localStorage, project, commands: commandRegistry})
     await historyManager.loadState()
   })
 
@@ -75,7 +75,7 @@ describe("HistoryManager", () => {
 
       it("saves the state", async () => {
         await historyManager.clearProjects()
-        const historyManager2 = new HistoryManager({stateStore, project, commands: commandRegistry})
+        const historyManager2 = new HistoryManager({stateStore, localStorage: window.localStorage, project, commands: commandRegistry})
         await historyManager2.loadState()
         expect(historyManager.getProjects().length).toBe(0)
       })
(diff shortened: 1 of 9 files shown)


damieng:
I'm okay with the api change as you say it's quite new and nobody else is using it just yet.

However switching backing store is going to effectively wipe-out current history. This is a real pain for anyone who uses it today - I think we need a fallback where if it finds no project history on load then it looks to the old storage to do a one-off import.   (You can delete the old importProjectHistory method that I left in there)

PR #13965, opened 2017-03-09
Allow specifying installation directory

krnowak:
### Requirements

### Description of the Change

Currently the `--install` flag of the `./script/build` is a boolean flag that either installs atom to some predefined location or skips it. This PR extends the flag so it can accept a value telling the script where to install atom. Passing just `--install` will behave as it used to so far, so it will install atom to `/usr/local` on linux and other default locations on darwin and windows. Same with not passing the flag - atom won't be installed. But passing `--install=/opt/atom` will install atom under the given location.

The documentation is updated accordingly.

### Alternate Designs

I haven't considered other designs, extending the `--install` flag looked natural to me.

It's a part of the builder script, so it can't be a package.

### Benefits

One can install the code in the location of choice. (Like for flatpaks, we want to install atom under `/app`).

### Possible Drawbacks

It's backward compatible, so not sure if there are drawbacks.

### Applicable Issues

Didn't file any.

### DIFF
--- a/docs/build-instructions/linux.md
+++ b/docs/build-instructions/linux.md
@@ -29,7 +29,7 @@ To also install the newly built application, use `--create-debian-package` or `-
 * `--compress-artifacts`: zips the generated application as `out/atom-{arch}.tar.gz`.
 * `--create-debian-package`: creates a .deb package as `out/atom-{arch}.deb`
 * `--create-rpm-package`: creates a .rpm package as `out/atom-{arch}.rpm`
-* `--install`: installs the application in `/usr/local/`.
+* `--install[=dir]`: installs the application in `${dir}`; `${dir}` defaults to `/usr/local`.
 
 ### Ubuntu / Debian
 
--- a/docs/build-instructions/macOS.md
+++ b/docs/build-instructions/macOS.md
@@ -21,7 +21,7 @@ To also install the newly built application, use `script/build --install`.
 
 * `--code-sign`: signs the application with the GitHub certificate specified in `$ATOM_MAC_CODE_SIGNING_CERT_DOWNLOAD_URL`.
 * `--compress-artifacts`: zips the generated application as `out/atom-mac.zip`.
-* `--install`: installs the application at `/Applications/Atom.app` for dev and stable versions or at `/Applications/Atom-Beta.app` for beta versions.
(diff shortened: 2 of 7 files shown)


Ben3eeE:
https://github.com/atom/atom/issues/13240

ungb:
Tested on Mac, Linux, Windows.  Only issue I saw was if you use `~`, it doesn't work and doesn't do anything.  It even shows like it works on mac but doesn't.

see screenshot:
image


I tested both default `--install` and with a path.  Both worked on all.  On CentOs, regular `--install` didn't work due to permission issue, but I am seeing this on master as well
see:
image


Since this isn't related,  The only issue is using ~ as I think it's a fairly common scenario and seems to fail silently which isn't great.

krnowak:
Hey, @ungb, thanks for testing! I didn't thought about handling the tilde, probably because I thought that nodejs will handle this itself, but apparently it is not [nor is planning to](https://github.com/nodejs/node/issues/684).

I updated the PR - rebased it to recent master, replaced `!=` with `!==` and added some code for handling tilde (this adds a dep on the [`passwd-user`](https://www.npmjs.com/package/passwd-user) module). Note that the tilde expansion happens only on linux and mac, and I have tested it only on linux. I checked the following (`kv` is my user name):
- `~`
- `~kv`
- `~/`
- `~kv/`
- `~/usr`
- `~kv/usr`
- `~yaddayaddanosuchuser/blah`
  - `Error: Failed to expand the tilde in ~yaddayaddanosuchuser/blah - user "yaddayaddanosuchuser" does not exist`

thomasjo:
You can greatly simplify this by using the [untildify](https://github.com/sindresorhus/untildify) package.

j-f1:
@thomasjo That only supports `~/...`, not `~user/...` AFAICT.

krnowak:
I considered the untildify package, but it didn't do the `~user` stuff.

Also, the appveyor failure seems unrelated, it even skips the installation, because `--install` flag was not passed.

ungb:
LGTM!  I was able to verify the new changes on linux and mac.

ungb:
Merging as I don't believe this is causing build failures on appveyor as I see other failures for similar things.

PR #13968, opened 2017-03-09
Remove the buildTextEditor binding in workspace

matthewwithanm:
The comment was actually no longer accurate since arrow functions aren't newable. After doing an audit, @maxbrunsfeld determined that it could just be removed.


maxbrunsfeld:
Thanks @matthewwithanm!

PR #13970, opened 2017-03-10
Refactor open; separate creation and placement of items

matthewwithanm:
This is a refactoring that sets up the work for #13878 (docks).

Because items will be able to dictate their preferred locations, we need to detangle the item creation from their placement.

I thought it was a little weird that the error was being thrown synchronously (instead of rejecting the returned promise), but there's a test that checks it so I kept it as-is.


matthewwithanm:
Oops guess tree-view uses it. I'll add an `openItemInPane()` and reimplement `openURIInPane()` in terms of that and `createItemForURI()`

maxbrunsfeld:
:+1:

PR #13977, opened 2017-03-10
Docks

matthewwithanm:
Hey all! I wanted to get this up to get some eyes on it and discuss some stuff. I'm trying to keep all the commits very focused to make things easy to review.

Stuff to Do
-----------

- [x] Add Dock component
- [x] Update atom/tabs to decorate docks
- [x] Serialize docks with workspace
- [x] Update `workspace.open()` to be location-aware
- [x] Remember (and re-use) last known location when opening
- [x] Add `WorkspaceCenter` class and `workspace.getCenter()` method
- [x] Update workspace pane inspection methods (that don't have to do with active panes) to reflect all locations
- [x] Don't create a Panel for docks
- [x] Update event listening strategy for showing toggle buttons
- [x] Swap LocalStorage for StateStore
- [x] Add tab check to drag handler
- [x] Add new `workspace.toggle()` method
- [x] Fix failing tests 😞
- [x] Add tests for new functionality/APIs 😞
- [x] Update Nuclide to new APIs, see if anything breaks
- [x] Fix issue: "core:save" only saves active item in center (#14027)
- [x] [atom/flight-manual] Update docs
- [x] [atom/tabs] Fix issue: Dragging overlay not showing up when over a "peaked" dock. (z-index?) (atom/tabs#425)
- [x] [atom/tabs] Add tests to make sure docks are decorated (atom/tabs#429)

Questions/Stuff to Discuss
---------------------------

I have some FIXMEs in the code for questions that came up along the way, as well as some others:

- ~[Is there any good way to detect whether something's droppable in the dock?][1]~
  - Just going to keep the Nuclide check for now and defer blessing anything without knowing the problem more
- ~[There are some known edge cases with showing the toggle buttons on hover.][2] Should we address them?~
  - Going to listen for leaving the central pane area, then handle it with mousemoves/position comparisons
- There's code that remembers the last place you put an item so that, if you close it and later open another, it'll show in the same place.
  - ~[Is trying to remember the split worth it?][3] It's not enough to just do it when a pane is moved (as is currently happening) since another pane moving would affect it. We could abuse `serialize()` and do it there instead. (It shouldn't be in the serialized value though because we want it to be shared across workspaces.) That wouldn't be perfect either though, because if the same item was in different locations in two windows, we wouldn't know which was preferred. (Either way, we're still kinda making a "best guess" since split location isn't context-independent.)~
    - Not worth it; leave it out for now
  - ~Is there a more efficient storage mechanism than [putting stringified JSON in local storage][4]?~
    - Use StateStore (this means `openItem()` will have to return a promise)
- ~The open API doesn't currently have a way to ensure a new item is created. (It always searches at least the active pane of the center.) Does that matter?~
  - Going to ignore until we have a use case
- ~Is it weird to have some of the workspace API methods (`paneForURI()`, `paneForItem()`)* include the docks while others (`getActivePane()`) don't? Similarly, if `searchAllPanes` includes docks, what's the desired behavior when `searchAllPanes` is false? (Previously, I mistakenly thought it meant to search no panes, so there wasn't any question.)~
  - `paneForURI()` and `paneForItem()` will include docks; active-related stuff won't. therefore, `searchAllPanes = true` will search docks, `false` (which only checks the active pane) won't
  - a `workspace.getCenter()` will be added and return an instance of the new `WorkspaceCenter` class, which trivially wraps `workspace.paneContainer`

* Oops, looks like I dropped the commit that updated those in a rebase. I'll add them back, depending on how we answer this question.

[1]: https://github.com/atom/atom/pull/13977/files#diff-aed5b259cccd55443edaa7b93d0d2bb0R144
[2]: https://github.com/atom/atom/pull/13977/files#diff-aed5b259cccd55443edaa7b93d0d2bb0R166
[3]: https://github.com/atom/atom/pull/13977/files#diff-4f265837c9eeb2d4123aad469237b2a2R295
[4]: https://github.com/atom/atom/pull/13977/commits/29bf4c90200cfec9e348d6518a2968db2bb6eae0


spec/workspace-spec.js
```diff
@@ -116,6 +118,63 @@ describe('Workspace', () => {
...
       })
     })
+
+    describe('where a dock contains an editor', () => {
+      afterEach(() => {
+        atom.workspace.getRightDock().paneContainer.destroy()
+      })
+
+      it('constructs the view with the same panes', () => {
+        const getActivePane = () => atom.workspace.getRightDock().getActivePane()
```
nathansobo: Rename to `getRightDockActivePane` to clarify intent here?

spec/workspace-spec.js
```diff
@@ -116,6 +118,63 @@ describe('Workspace', () => {
         expect(atom.workspace.getTextEditors().length).toBe(0)
       })
     })
+
+    describe('where a dock contains an editor', () => {
+      afterEach(() => {
+        atom.workspace.getRightDock().paneContainer.destroy()
+      })
+
+      it('constructs the view with the same panes', () => {
```
nathansobo: We have this file [`async-spec-helpers.js`](https://github.com/atom/atom/blob/master/spec/async-spec-helpers.js) that might help eliminate some of this `waitsFor` noise by enabling you to use `async`/`await`. Check out some other specs that use it for examples.

src/dock.js
```diff
@@ -0,0 +1,536 @@
...
+const DEFAULT_INITIAL_SIZE = 300
+const HANDLE_SIZE = 4
+const SHOULD_ANIMATE_CLASS = 'atom-dock-should-animate'
+const OPEN_CLASS = 'atom-dock-open'
+const RESIZE_HANDLE_RESIZABLE_CLASS = 'atom-dock-resize-handle-resizable'
+const TOGGLE_BUTTON_VISIBLE_CLASS = 'atom-dock-toggle-button-visible'
+const CURSOR_OVERLAY_VISIBLE_CLASS = 'atom-dock-cursor-overlay-visible'
+
+// Extended: A container at the edges of the editor window capable of holding items.
+// You should not create a `Dock` directly, instead use {Workspace::open}.
```
nathansobo: Maybe we should mention the dock getters; something like...

> You should not create a `Dock` directly. Instead, access one of the three docks of the workspace via {::getLeftDock}, {::getRightDock}, and {::getBottomDock} or add an item to a dock via {Workspace::open}.

src/dock.js
```diff
@@ -0,0 +1,536 @@
...
+      this.paneContainer.observePanes(pane => {
+        pane.onDidAddItem(this.handleDidAddPaneItem.bind(this))
+      }),
+      this.paneContainer.observePanes(pane => {
+        pane.onDidRemoveItem(this.handleDidRemovePaneItem.bind(this))
+      })
+    )
+  }
+
+  // FIXME(matthewwithanm: This is kinda gross. We need to get a view for the pane container so we
```
nathansobo: Could the `PaneContainer` be switched to have a `getElement` method as well that's sort of self-registering or something?

src/panel-container-element.js
```diff
@@ -19,6 +19,12 @@ class PanelContainerElement extends HTMLElement {
     this.subscriptions.add(this.model.onDidAddPanel(this.panelAdded.bind(this)))
     this.subscriptions.add(this.model.onDidDestroy(this.destroyed.bind(this)))
     this.classList.add(this.model.getLocation())
+
+    // Add the dock.
+    if (this.model.dock != null) {
+      this.appendChild(this.views.getView(this.model.dock))
```
nathansobo: What about `this.model.dock.getElement()` to avoid adding another use of the view registry?

nathansobo (commented):
I have some minor feedback and admittedly didn't examine all the mouse handling code in extreme detail, but everything looks pretty solid.

spec/workspace-spec.js
matthewwithanm: Can we do that separately? This file already has > 100 usages of `waitsForPromise()` and I feel like it's better to maintain the consistency and switch them all at once.

spec/workspace-spec.js
matthewwithanm: 👍

src/dock.js
matthewwithanm: 👍

src/dock.js
matthewwithanm: I started switching over to that but I felt like it was a lot of change for this PR so I pulled back.  (Converting PaneContainer means converting Pane.)

But I did it now 😊 98ce09a

spec/workspace-spec.js
nathansobo: Sure thing.

src/dock.js
```diff
@@ -0,0 +1,544 @@
...
+    this.setState({
+      size: serialized.size,
+      // If no items could be deserialized, we don't want to show the dock (even if it was open last time)
+      open: serialized.open && (this.paneContainer.getPaneItems().length > 0)
+    })
+  }
+
+  // PaneContainer-delegating methods
+
+  getPanes () {
```
maxbrunsfeld: Can you add documentation for these `PaneContainer`-delegating methods, making them public? Docks should have a similar set of public methods as `WorkspaceCenter`.

src/project.coffee
```diff
@@ -30,8 +30,8 @@ class Project extends Model
     @consumeServices(packageManager)
 
   destroyed: ->
-    buffer.destroy() for buffer in @buffers
-    repository?.destroy() for repository in @repositories
+    buffer.destroy() for buffer in @buffers.slice()
+    repository?.destroy() for repository in @repositories.slice()
```
maxbrunsfeld: Curious what prompted these. They seem reasonable.

src/workspace.js
```diff
@@ -41,6 +44,9 @@ module.exports = class Workspace extends Model {
     this.assert = params.assert
     this.deserializerManager = params.deserializerManager
     this.textEditorRegistry = params.textEditorRegistry
+    this.hoveredDock = null
+    this.draggingItem = false
+    this.previousLocations = new StateStore('AtomPreviousItemLocations', 1)
```
maxbrunsfeld: Can you rename this to `itemLocationStore`?

src/workspace.js
```diff
@@ -1006,6 +1170,64 @@ module.exports = class Workspace extends Model {
...
+
+  getRightDock () {
+    return this.docks.right
+  }
+
+  getBottomDock () {
+    return this.docks.bottom
+  }
+
+  getPaneLocations () {
```
maxbrunsfeld: The term `location` is used to refer both to strings like `left` and `center` and also to objects like `Docks` and the `WorkspaceCenter`. @nathansobo and I think that when referring generically to these objects, we should call them 'pane containers'. There's some ambiguity with this name, because `PaneContainer` is an internal class which all of these classes *wrap*, but we think this is ok, because these classes have a kind of *is a* relationship with the `PaneContainer` class, even though we use composition rather than inheritance. Also, the `PaneContainer` is a private class, whereas the name `location` shows up in our API in methods like `getDefaultLocation`.

src/workspace.js
```diff
@@ -227,6 +287,27 @@ module.exports = class Workspace extends Model {
     })
   }
 
+  subscribeToMovedItems () {
+    if (this.movedItemSubscription != null) {
+      this.movedItemSubscription.dispose()
+    }
+    const paneLocations = Object.assign({center: this}, this.docks)
```
maxbrunsfeld: We think you should add a `getLocation` method to `WorkspaceCenter` so that this code can be written like this:

```js
    for (const paneContainer of this.getPaneContainers()) {
      paneContainer.onDidAddPaneItem(({item}) => {
        if (typeof item.getURI === 'function') {
          const uri = item.getURI()
          if (uri != null) {
            this.previousLocations.save(item.getURI(), paneContainer.getLocation())
          }
        }
      }
    }
```

We don't need to save off the disposables returned from `onDidAddPaneItem`, because the `Workspace` will always outlive its pane containers.

maxbrunsfeld (changes requested):
I paired with @nathansobo on some further review. This is looking really good!

src/workspace.js
```diff
@@ -657,43 +771,79 @@ module.exports = class Workspace extends Model {
...
+      .then(() => {
+        if (pane != null) return pane
+        pane = paneLocation.getActivePane()
+        switch (split) {
+          case 'left': return pane.findLeftmostSibling()
+          case 'right': return pane.findOrCreateRightmostSibling()
+          case 'up': return pane.findTopmostSibling()
+          case 'down': return pane.findOrCreateBottommostSibling()
+          default: return pane
+        }
```
matthewwithanm: the first one is for finding and the second is for creating and we want to make sure we don't create a new one while we're searching. although looking back now, i think that the split shouldn't even be used in the searching part, should it?

spec/workspace-spec.js
```diff
@@ -116,6 +118,63 @@ describe('Workspace', () => {
...
+        const getRightDockActivePane = () => atom.workspace.getRightDock().getActivePane()
+        const pane1 = atom.workspace.getRightDock().getActivePane()
+        const pane2 = pane1.splitRight({copyActiveItem: true})
+        const pane3 = pane2.splitRight({copyActiveItem: true})
+        let pane4 = null
+
+        waitsForPromise(() =>
+          atom.workspace.open(null, {pane: getRightDockActivePane()}).then(editor => editor.setText('An untitled editor.'))
```
matthewwithanm: the location API is definitely the natural one, but i thought we agreed to leave that out for now so people wouldn't fall into the trap of adding things to specific locations (thereby making them immovable)?

src/project.coffee
matthewwithanm: At some point I got an error (calling destroy on undefined). Something in an event handler was mutating the array being iterated over and, since these CS loops compile to plain old for loops, it would end up accessing bad indexes.

I'm not sure why the existential operator in in the second line but I kinda wondered if it wasn't to work around the same problem and figured, even if it wasn't, it's good not to iterate over a thing you might muck with so I changed that too XD

maxbrunsfeld: The existential operator is because the `@repositories` array can contain `null` entries, since its indices correspond to entries in the `directories` array, which can contain non-repository directories.

:+1: That makes sense about an event listener removing a repository or directory.

src/workspace.js
```diff
@@ -657,43 +771,79 @@ module.exports = class Workspace extends Model {
...
+    let location
+    if (paneLocation == null && pane == null && split == null && uri != null) {
+      location = this.previousLocations.load(uri)
     }
 
-    const index = pane.getActiveItemIndex()
-    const uri = options.uri == null && typeof item.getURI === 'function' ? item.getURI() : options.uri
-    this.emitter.emit('did-open', {uri, pane, item, index})
-    return item
+    return Promise.resolve(location)
```
maxbrunsfeld: Oh I see. Yeah, I think it's fine to change that test.

src/workspace.js
```diff
@@ -625,53 +625,119 @@ module.exports = class Workspace extends Model {
...
+
+    if (options.activateItem !== false) {
+      pane.activateItem(item, {pending: options.pending})
     }
 
-    return Promise.resolve(item)
-      .then(item => this.openItem(item, Object.assign({pane, uri}, options)))
+    if (options.activatePane !== false) {
+      pane.activate()
+      const container = this.getPaneContainers().find(container => container.getPanes().includes(pane))
```
matthewwithanm: a small thing but maybe it would be good to just have one `container` variable and assign it on 688, then only search here if it's undefined?

maxbrunsfeld: Yeah, good point. In the common case, the container will already have been found.

maxbrunsfeld (approved):

nathansobo:
After thinking about putting the default tree view inside the left dock, we've realized that Docks needs a *bit* more polish to ensure a good experience in the default Atom configuration. Basically, we want to make sure that people can't accidentally close their tree view or move it somewhere it doesn't make sense. I know we want to keep this simple, but I think the following changes could really help the UX around docks while they are still new and not many bundled or third party packages are using them.

* Add an optional `getAllowedLocations` method on items and only allow them to be placed in those docks. This can be used to constrain the tree view to the left or right docks.
* If a dock is empty, only show the toggle affordance when dragging an item that can be placed in that dock. This will avoid users opening docks and feeling like "what the hell is this empty space here?"...
* Add an optional `isPermanentDockItem` method on items that will cause no  `X` icon to be rendered for the items tab when appears in a dock. If the item is in a center pane, this method will have no affect (center pane items can always be closed). 

I really think these changes could help us transition more smoothly into the docks experience and prevent them from feeling like they are in the way. If you want to do a synchronous discussion next week I'd be open to it.

matthewwithanm:
> If a dock is empty, only show the toggle affordance when dragging an item that can be placed in that dock. This will avoid users opening docks and feeling like "what the hell is this empty space here?"...

Sorry, this afternoon I thought that this was just applying to the behavior while dragging but now I'm realizing you meant when not dragging too. Are you sure about that? I had felt that dock permanence made the opening behavior clearer and that conditionally hiding those buttons would confuse the "hover the edge to open a dock" story.

src/dock.js
```diff
@@ -0,0 +1,739 @@
...
+      this.cursorOverlayElement.classList.add(CURSOR_OVERLAY_VISIBLE_CLASS)
+    } else {
+      this.cursorOverlayElement.classList.remove(CURSOR_OVERLAY_VISIBLE_CLASS)
+    }
+
+    const shouldBeVisible = state.open || state.showDropTarget
+    const size = Math.max(MINIMUM_SIZE, state.size == null ? this.getInitialSize() : state.size)
+
+    // We need to change the size of the mask...
+    this.maskElement.style[this.widthOrHeight] = `${shouldBeVisible ? size : HANDLE_SIZE}px`
```
simurai: The `4px` gap covers these git markers when a dock is collapsed.

Open | Collapsed
--- | ---
screen shot 2017-03-25 at 8 20 12 pm  | screen shot 2017-03-25 at 8 32 13 pm

Did the gap get added because the One themes had a gap between panes? In https://github.com/atom/one-light-ui/pull/96 I removed the "Spacious" mode, so now there are no more gaps and just `1px` borders. So I had to use `!important` to [override](https://github.com/atom/one-light-ui/pull/96/files#diff-f72770f52e9637dd2337d151b4d73e45R7) the `width/height`.

Was also thinking of adding a `1px` border on all panes and panels. Because currently packages do that but then you get double borders when two packages are next to each other. So if core would take care of borders, packages only have to think about what's "inside". Themes that don't like borders could remove them or just do `@base-border-color: transparent`.

static/docks.less
```diff
@@ -0,0 +1,214 @@
...
+.atom-dock-content-wrapper {
+  position: absolute;
+  display: flex;
+  flex: 1;
+  align-items: stretch;
+  width: 100%;
+  height: 100%;
+  cursor: default;
+  -webkit-user-select: none;
+  white-space: nowrap;
```
matthewwithanm: Good catch! I'll update it.

src/dock.js
matthewwithanm: The 4px is for the resize handle (though I should update it to measure the element instead of hardcoding it). I'm confused about that second screenshot though…if it's collapsed, the entire thing should be hidden.

matthewwithanm: The bottom dock not having a top border is problematic though. If you try to drag something into it without having a file open, the dock becomes totally lost in the empty center…any ideas for that?

matthewwithanm:
The "peaked" dock z-index issue I mentioned in the OP can be fixed by setting the `.tabs-layout-overlay` z-index to 11 in the tabs package (peaked docks have a z-index of 10). @nathansobo @maxbrunsfeld @simurai do you all have any better ideas for dealing with that? If not, I can open a one-line PR for it 😝

simurai:
> can be fixed by setting the .tabs-layout-overlay z-index to 11

👍  I think that's fine. There isn't really a convention around managing z-index and since these overlays are more temporary and semi transparent, shouldn't be too risky.

src/dock.js
simurai: > if it's collapsed, the entire thing should be hidden.

Here a gif:

docks

Notice how line 17 has an orange "changed" marker that gets covered when the dock is collapsed. It's because `.atom-dock-mask` has `width: 4px` inlined. Yeah, I guess it's to reserve that space for the resize handle. Giving `atom-dock`s a min-width/height should also work:

```
atom-dock.left,
atom-dock.right {
  min-width: 4px;
}
atom-dock.bottom {
  min-height: 4px;
}
```

Then it pushes the editor a bit to the side and the "changed" markers shouldn't be covered anymore.

Also, after using it for a bit, I'm not sure about the toggle buttons sliding out all the time. But I'll make a general comment for discussing that.

simurai:
How does everyone feel about the toggle buttons sliding out when you move your mouse over the docks:

buttons

I guess it's nice to give an affordance to toggle the docks, but they also grab so much attention and might feel distracting, no? How about:

- Only show the toggle buttons when a dock is collapsed and you start dragging an item.
- Don't show the resize handle when a dock is collapsed.

In other words, you can only open/close the docks with the keyboard. It wouldn't be any different than the current panels. If a package really wants to support toggling with the mouse, they could add a button to the status-bar. Like linter and others already do. And users could add buttons to the tool-bar. And of course, package specific keybindings would keep working. Like toggling the tree-view or find-and-replace wouldn't change, just that now they share the space with other panels.

I setup keybindings to quickly toggle the docks with: `ctrl-`⬅️  , `ctrl-`⬇️ ,  `ctrl-`➡️ . Feels a bit like playing piano. 🎹 

piano

nathansobo:
@maxbrunsfeld one more failure that I'm having trouble reproducing locally on the deprecation cop status bar view.

matthewwithanm:
@nathansobo @maxbrunsfeld think i got it: atom/deprecation-cop#81

maxbrunsfeld:
👏  Thanks for your hard work on this @matthewwithanm. This is a huge feature.

maxbrunsfeld:
@simurai What do you think of the behavior now: the dock tabs only appear if there is actually something *in* the dock. They also appear when you're dragging a tab, in case you want to drag it to the dock. Does this make it less distracting? We'll likely iterate a bit on the UX, but I feel like this is a good default.

simurai:
> @maxbrunsfeld Does this make it less distracting?

Yeah, a little, but docks will probably not be empty that often? I actually felt bad, suggesting to remove the toggle buttons 😇 , but then I was like: "Well, it's just a suggestion, people don't have to agree." So yeah, I'm ok to keep the toggle buttons as a default and maybe I can try to add a config option to hide them in the One themes. Then it's more like the wrap-guide, have it for easy discoverability, but still let people remove it, if they like a more minimal UI.

@matthewwithanm Thanks so much for this. Really ❤️  how it lets you organize panels exactly how you like them.

nathansobo:
screen shot 2017-03-31 at 3 06 36 pm

@simurai Should we tweak the styling of tabs inside docks to go full width? For some reason the empty space to the right of the tree-view's tab is feeling awkward to me. Also, want to do an icon for the tree-view's tab? BTW I'm calling the "tree view" the "project browser" in the UI because I think it conveys more information.

simurai:
> @nathansobo Should we tweak the styling of tabs inside docks to go full width?

Yeah, for the One themes it will be the new default: https://github.com/atom/one-light-ui/pull/96. It actually goes full width for all tabs, also in the center. I'll merge it so people can try it out and see how it feels. There is still the option to only use the minimum width:

tab-sizing

We could also try to add it to `atom/tabs`, but all themes have their own tabs behaviour, so probably would get overridden anyways.

PR #13982, opened 2017-03-11
Adding 'Run Benchmark' menu item for win32 and linux platform

liuderchi:
### Description of the Change

By latest comment in #12984 from @as-cii 
adding 'Run Benchmark' menu item for win32 and linux platform.
would be an enhancement.

Also use letter `B` as hotkey hint.

### Benefits

we can run benchmark via menu in win32 and linux platform.

### Applicable Issues

enhancement of #12984


as-cii:
Thanks, @liuderchi!

PR #13987, opened 2017-03-12
src/config.coffee should use _.deepClone from underscore-plus

adityaYashwanth:
### Description of the Change

Code cleanup in src/config.coffee file - Remove own implementation of deepClone function and use it from underscore-plus.

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -937,10 +937,10 @@ class Config
       defaultValue = getValueAtKeyPath(@defaultSettings, keyPath)
 
     if value?
-      value = @_.deepClone(value)
+      value = _.deepClone(value)
       @deepDefaults(value, defaultValue) if isPlainObject(value) and isPlainObject(defaultValue)
     else
-      value = @_.deepClone(defaultValue)
+      value = _.deepClone(defaultValue)
 
     value
 
@@ -1006,7 +1006,7 @@ class Config
           result[key] = @deepDefaults(result[key], object[key])
       else
         if not result?
-          result = @_.deepClone(object)
+          result = _.deepClone(object)
     result
 
   # `schema` will look something like this
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -937,10 +937,10 @@ class Config
       defaultValue = getValueAtKeyPath(@defaultSettings, keyPath)
 
     if value?
-      value = @deepClone(value)
+      value = @_.deepClone(value)
       @deepDefaults(value, defaultValue) if isPlainObject(value) and isPlainObject(defaultValue)
     else
-      value = @deepClone(defaultValue)
+      value = @_.deepClone(defaultValue)
 
     value
 
@@ -996,16 +996,6 @@ class Config
         console.warn("'#{keyPath}' could not set the default. Attempted default: #{JSON.stringify(defaults)}; Schema: #{JSON.stringify(@getSchema(keyPath))}")
     return
 
-  deepClone: (object) ->
-    if object instanceof Color
-      object.clone()
-    else if _.isArray(object)
-      object.map (value) => @deepClone(value)
-    else if isPlainObject(object)
-      _.mapObject object, (key, value) => [key, @deepClone(value)]
-    else
-      object
-
   deepDefaults: (target) ->
     result = target
     i = 0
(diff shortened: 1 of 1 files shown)


anatoli26:
Everything looks correct.

ericcornelissen:
If anything, it seems to me that this isn't working because the `deepClone` method from underscore-plus ([see here](https://github.com/atom/underscore-plus/blob/master/src/underscore-plus.coffee#L98)) does not have the `object instanceof Color` branch ([see here](https://github.com/atom/atom/pull/13987/files#diff-1dca5d0fd1e63d76f43245422a3a18efL1000)) that the implementation here does. If you look at the logs from both AppVeyor and CircleCI ([link](https://ci.appveyor.com/project/Atom/atom/build/5601/job/f6ms5bsdkagffwr1#L515)) you see that the failing test suites are related to Color in one way or another...

I expect the `deepClone` "duplicate implementation" in src/config.coffee was put there intentionally because of this issue. I do agree that isn't an elegant solution, but unless you can find a way for the Color class to be cloned properly by underscore-plus, this PR is probably not going to work out...

adityaYashwanth:
Will it be fine, if I add `object instanceof Color` branch to the deepClone implmentation in underscore-plus.([see here](https://github.com/atom/underscore-plus/blob/master/src/underscore-plus.coffee#L98)) If yes, I will create a seperate PR for this in [atom/underscore-plus](https://github.com/atom/underscore-plus) repository.

ericcornelissen:
It would be, were it not that `Color` is a class defined within the Atom source code. It is imported as `Color = require './color'` in src/config.coffee, which refers to [this](https://github.com/adityaYashwanth/atom/blob/17963887b134d0f7a661e152065728348930a625/src/color.js) file...

The problem seems to arise from the fact that the deepClone method in src/config.coffee uses the clone method defined within the `Color` class ([see here](https://github.com/adityaYashwanth/atom/blob/17963887b134d0f7a661e152065728348930a625/src/color.js#L108)). This method corrects for the fact that `Color` internally refers to `red`, `green`, `blue` and `alpha` with a leading underscore in its setters and getters (which is needed because getters and setters have no internal state themselves, so they need to store their data in some other way). However, as a result of this a "normal" clone of a `Color` instance will clone the properties with a leading underscore (as well as the properties without? I'm not sure on the details here tbh), which in turn seem to lead to the failing tests...

Since I do agree that it would be better to use the underscoreJS version of deepClone I think the best solution here is to replace the `get` and `set` methods in the `Color` class with actual methods, this will allow you to store the actual data (`red`, `green`, `blue` and `alpha`) without a leading underscore. I expect that, once that is done, underscoreJS will correctly clone any object, even if it is a `Color`.

I hope that helped 😄

j-f1:
Another option could be to monkey-patch `_.deepClone` in Atom to recognize `Color` instances.

ericcornelissen:
You could, but that kinda defeats the propose of this PR, right? Correct me if I'm wrong but if you would do that your effectively just using the alternative version written in src/config.coffee disguised as being the one provided by underscoreJS

adityaYashwanth:
Thanks, that really helped. I should have seen this before posting the previous comment. I will start making those changes right away. 😃

atomissariyaporn (approved):

mjguru:
Hey, I am new to open source. I would like to contribute. Is this issue still open?

ericcornelissen:
@mjguru I don't think so. As you an read in the discussion using deepClone by underscoreJS (or any library really) breaks functionality with the custom `Color` class Atom implements. That is, unless you can find a good solution different to this one and #15066, of course.

If you're new and want to contribute I recommend you to keep an eye on the [beginners label](https://github.com/atom/atom/issues?q=is%3Aopen+is%3Aissue+label%3Abeginner) under issues. Or maybe check out a smaller project, it is usually easier to contribute there 😄

sadick254 (changes requested):
@adityaYashwanth Thank you for your contribution. Atom has gone through some changes in recent years. We have moved from using `coffee-script` to `javascript`. This move makes this PR outdated. Do you mind updating it by making changes to `config.js` file?

sadick254:
Thank you for your issue!

We haven't gotten a response to our suggestion in our comment above. I'm going to close this but don't hesitate to reach out if you are still willing to work on this, we'll be happy to reopen the PR.

PR #14003, opened 2017-03-14
quickly hide title-bar by config

xtoolkit:
### Description of the Change

hi
first i add minimize function in `atom-environment.coffee` for use this command: `atom.minimize();`
two add `forceHideTitleBar` config. because designer need hide title-bar for beautiful develop atom.

better design develop.

### Benefits

quick

### Possible Drawbacks

nothing

### Applicable Issues

nothing

________________________

### my customize [with this changes] demo:

screenshot 42

[Video demo](http://g1.asset.aparat.com/flv_video_new/2243/536fa42b3fce3ee5d3275c8e09d31dda6726007-720p.apt?direct=1)

sry 4 my bad english

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -551,6 +551,9 @@ class AtomEnvironment extends Model
   maximize: ->
     @applicationDelegate.maximizeWindow()
 
+  minimize: ->
+    @applicationDelegate.minimizeWindow()
+    
   # Extended: Returns a {Boolean} that is `true` if the current window is in full screen mode.
   isFullScreen: ->
     @applicationDelegate.isWindowFullScreen()
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -52,7 +52,7 @@ class AtomWindow
     if @shouldHideTitleBar()
       options.frame = false
 
-    if @froceHideTitleBar()
+    if @forceHideTitleBar()
       options.frame = false
 
     @browserWindow = new BrowserWindow(options)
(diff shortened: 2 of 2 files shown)


xtoolkit:
pre request for https://github.com/xtoolkit/arshen-ac/pull/1 package

Stanzilla:
Well I guess you wanted `forceHideTitleBar` instead of `froceHideTitleBar`?

xtoolkit:
Why can not it be like that?

xtoolkit:
Ok . This is not a good method . Give me a solution instead.
@nathansobo @maxbrunsfeld @damieng

damieng:
The problem is that hiding the title bar currently only works on macOS and a theme really needs to be able to support it by providing similar functionality.

Additionally if we were to go down this path;

1. The correct spelling is 'force' not 'froce'
2. New configuration settings should contain an entry in the schema with the appropriate description

xtoolkit:
@damieng thank you for answering!
can you help me for second path?

simurai:
> first i add minimize function in atom-environment.coffee for use this command: atom.minimize();

👍  on adding a `atom.minimize();` command

> two add forceHideTitleBar config.

Do you mean you would like a command to force the title-bar to be hidden? So you can use that command in your package?

Couldn't you "force" it by directly setting the config to `hidden`:

```
atom.config.set("core.titleBar", "hidden")
```

Note that changing the title bar can't be done "quickly" and needs a reload of Atom. You should see this popup when you try:

screen shot 2017-04-25 at 10 46 21 am

sadick254:
Thanks for your contribution.

I am closing this PR since the ability to hide the title bar has been available for a while now on windows, mac and linux systems.  Please try the latest Atom release. Thank you.

PR #14014, opened 2017-03-16
Test text-buffer@11.4.0

nathansobo:
This PR tests the text-buffer 11.4.0-0 prerelease before publishing it.

Refs https://github.com/atom/text-buffer/pull/223


nathansobo:
@damieng Here's another example of a Windows-only failure... it's the Babel transpilation test which I think you've already seen. Just a heads up.

PR #14016, opened 2017-03-16
More reliable way of stopping Babel stderr output on Windows

damieng:
The debug module used by babel tends to throw stuff onto stderr and stdout when it feels like it.

This is a problem on Windows when Atom is launched from the UI and it doesn't have these attached - writing to them causes errors.

We previously had another fix in but it was quite dependent on the internals of Babel and broke. Hopefully this one lasts longer.

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -44,11 +44,7 @@ exports.getCachePath = function (sourceCode) {
 exports.compile = function (sourceCode, filePath) {
   if (!babel) {
     babel = require('babel-core')
-    var Logger = require('babel-core/lib/transformation/file/logger')
-    var noop = function () {}
-    Logger.prototype.debug = noop
-    Logger.prototype.verbose = noop
-
+    require('debug').log = () => { } // Stop Babel logging to stdout/stderr (would error when no console attached)
     options = {ast: false, babelrc: false}
     for (var key in defaultOptions) {
       if (key === 'plugins') {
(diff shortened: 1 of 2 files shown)


as-cii:
https://github.com/atom/atom/pull/14022 has been merged, so this might not be required anymore.

src/babel.js
```diff
@@ -44,11 +44,7 @@ exports.getCachePath = function (sourceCode) {
 exports.compile = function (sourceCode, filePath) {
   if (!babel) {
     babel = require('babel-core')
-    var Logger = require('babel-core/lib/transformation/file/logger')
-    var noop = function () {}
-    Logger.prototype.debug = noop
-    Logger.prototype.verbose = noop
-
+    require('debug').log = () => { } // Stop Babel logging to stdout/stderr (would error when no console attached)
```
thomasjo: I might be completely wrong about how Node modules work, but won't this affect `debug.log` **everywhere**? 
Pretty sure the module cache is not scoped.

damieng: Yes, it will. But if you use debug.log on Windows right now when launching from the ui with that process variable set it will error.

Short of getting the debug people to not try and write to stderr/stdout on Windows I don't know there is a great solution.

j-f1: @damieng You could make a PR to babel to allow disabling logging.

damieng:
Going to close this down as we're backing out the babel work.

PR #14022, opened 2017-03-17
Revert "Upgrade to Babel 6 and apply fewer transformations when transpiling babel files"

as-cii:
Reverts atom/atom#13823.

For the reasons outlined by @nathansobo in https://github.com/atom/atom/pull/13823#issuecomment-287117624, we are going to revert the aforementioned pull request. Long-term we would like remove support for global transpilation but we need a safer transition path to do so, which the above pull request did not provide.

I will merge this as soon as I have a green build, and cherry pick onto beta as well in order to prevent Babel 6 from going to stable.

/cc: @atom/maintainers


nathansobo:
Thanks for taking care of this.

PR #14024, opened 2017-03-17
Snapshot `AtomEnvironment`

as-cii:
This pull request is a continuation of #13916 and takes advantage of snapshots to create an `AtomEnviroment` instance as part of `script/build`. Then, at runtime, Atom can simply use this instance and complete its initialization by supplying all the information we didn't have at compile-time: setting the `ATOM_HOME`, reading the user configuration, etc.

The following is a profile of how long it takes to initialize an `AtomEnvironment` on master:

screen shot 2017-03-17 at 16 43 10

After these changes, this is how the same profile looks like:

screen shot 2017-03-17 at 16 38 51

You can notice how creating an `AtomEnvironment` is ~2x faster now, as it went down from `~106ms` to `~58ms`.

It's worth noting that just loading the base style sheets took more than `~30ms` (more than half of the `initialize` method). This is due to performing a lot of synchronous I/O but it can be almost eliminated by, again, making use of snapshots and load as much data as possible during `script/build`. This will be addressed in a forthcoming pull request.

/cc: @atom/maintainers


nathansobo:
Making great progress.

Ben3eeE:
Tested this out and did not find any issues :ship:

PR #14027, opened 2017-03-18
Add `saveFocusedPaneItem()` and call it in "core:save" command

matthewwithanm:
### Description of the Change

Add methods that chose what to save based on focus (instead of the active pane). This behavior change means that pane items in docks (#13977) will be savable too.

### Alternate Designs

The alternate we considered was having a single method which attempted to use the focused pane and then fell back to the active one, however this one was chosen because it's more predictable (i.e. doesn't depend on a state the user may not understand intuitively).

### Benefits

Editors in docks can be saved!

### Possible Drawbacks

Some may be relying on the current behavior, and use the command when an editor isn't active.

cc @nathansobo @maxbrunsfeld

### DIFF
--- a/spec/workspace-spec.js
+++ b/spec/workspace-spec.js
@@ -1872,6 +1872,38 @@ i = /test/; #FIXME\
     })
   })
 
+  describe('::saveFocusedPaneItem', () => {
+    let editor, workspaceElement
+
+    beforeEach(() => {
+      workspaceElement = atom.views.getView(atom.workspace)
+      document.body.appendChild(workspaceElement)
+      waitsForPromise(() => atom.workspace.open('a').then(o => { editor = o }))
+    })
+
+    afterEach(() => {
+      workspaceElement.remove()
+    })
+
+    it("calls the focused item's save method", () => {
+      spyOn(editor, 'save')
+      editor.getElement().focus()
+      atom.workspace.saveFocusedPaneItem()
+      expect(editor.save).toHaveBeenCalled()
+    })
+
+    it("doesn't save the active editor if it's not focused", () => {
+      spyOn(editor, 'save')
+      const input = document.createElement('input')
+      document.body.appendChild(input)
+      input.focus()
+      expect(document.activeElement).toBe(input)
+      atom.workspace.saveFocusedPaneItem()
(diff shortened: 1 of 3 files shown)


src/workspace.js
```diff
@@ -867,7 +868,43 @@ module.exports = class Workspace extends Model {
...
+      }
+      el = el.parentElement
+    }
+  }
+
+  // Save the currently focused pane item.
+  //
+  // If the focused pane item currently has a URI according to the item's
+  // `.getURI` method, calls `.save` on the item. Otherwise
+  // {::saveActivePaneItemAs} # will be called instead. This method does nothing
```
maxbrunsfeld: Looks like an extra `#` character in this comment ☝️ . Nice docs.

spec/workspace-spec.js
```diff
@@ -1871,6 +1871,26 @@ i = /test/; #FIXME\
...
+      expect(editor.save).toHaveBeenCalled()
+    })
+    it("doesn't save the active editor if it's not focused", () => {
+      spyOn(editor, 'save')
+      spyOn(atom.workspace, 'getFocusedPane').andReturn(null)
+      atom.workspace.saveFocusedPaneItem()
+      expect(editor.save).not.toHaveBeenCalled()
+    })
+  })
+
```
maxbrunsfeld: I think it'd be good to actually use the real implementation of `getFocusedPane` in these tests, since that's where some of the main complexity lies. You'd attach the workspace element to the DOM, but we often do that in order to test non-trivial focus stuff.

maxbrunsfeld (changes requested):
Thanks @matthewwithanm! 2 small suggestions.

src/workspace.js
matthewwithanm: actually this was copypasta and it's not correct 😰

spec/workspace-spec.js
matthewwithanm: 👍 i was worried about how that might affect other stuff but if you're doing it elsewhere, i guess i'm good (:

matthewwithanm:
@maxbrunsfeld Tests updated! 😊

PR #14038, opened 2017-03-21
Optimize style sheets loading

as-cii:
This pull request takes advantage of snapshots auxiliary data to minimize the I/O associated to reading a core package's style sheets when opening Atom. Specifically, we will now store all the sources, digests and style sheet paths in the snapshot blob, so that we can load them all at once during startup and avoid accessing the file system at runtime.

The following is a CPU profile taken on master after merging #14024:

screen shot 2017-03-21 at 09 52 17

If we zoom in the various sections of this profile, we can see that `LessCache` shows up quite frequently:

screen shot 2017-03-21 at 09 52 50

screen shot 2017-03-21 at 09 53 16

screen shot 2017-03-21 at 09 53 33

After these changes, this is how the same profile looks like:

screen shot 2017-03-21 at 09 56 22

You can notice how Atom now loads `~ 100ms` faster thanks to the minimization of the I/O associated with listing and reading less style sheets; this improvement should also make startup time more consistent across reloads because it will make Atom rely less on the file system.

One thing to note is that we will still need to update `LessCache` and, as a result, write and read from `~/.atom/compile-cache/less` when booting up Atom. We could potentially minimize that cost too by batching reads and writes, but we think we could defer this to later and address other concerns (like packages preloading) first.

/cc: @atom/maintainers


nathansobo:
:zap: Goodness keeps rolling in.

PR #14063, opened 2017-03-25
fixing middle mouse button clipboard paste on linux (atom:atom#8648)

dietmar:
### Description of the Change

The functionality of pasting the clipboard on middle mouse click on Linux has been moved from `onMouseUp` to `onMouseDown`. Now it is consistent with the general behavior on Linux. Also, there was some interplay of this functionality with dragging, which resulted in issue #8648. This is resolved now.

### Alternate Designs

None

Addresses a bug in Core.

### Benefits

Resolution of issue #8648 and consistency of UI behavior with other Linux applications.

### Possible Drawbacks

None

### Applicable Issues

#8648

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -5019,6 +5019,35 @@ describe('TextEditorComponent', function () {
       expect(atom.clipboard.read()).toBe('sort')
       expect(editor.lineTextForBufferRow(10)).toBe('sort')
     })
+
+    it('pastes the previously selected text at the clicked location, left clicks do not interfere', async function () {
+      let clipboardWrittenTo = false
+      spyOn(require('electron').ipcRenderer, 'send').andCallFake(function (eventName, selectedText) {
+        if (eventName === 'write-text-to-selection-clipboard') {
+          require('../src/safe-clipboard').writeText(selectedText, 'selection')
+          clipboardWrittenTo = true
+        }
+      })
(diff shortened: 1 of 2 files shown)


50Wliu:
Can you please add some specs to cover this change?  The current middle-mouse-paste specs are at https://github.com/atom/atom/blob/4acd477559d5cc0defddd840a793fa24d784f646/spec/text-editor-component-spec.js#L4984.

dietmar:
As far as I can tell (which I'm afraid really doesn't mean much, with regard to specs), those current specs are still correct: After something has been copied to the clipboard, a middle mouse button `mousedown` followed by a `mouseup` should result in the clipboard text being pasted. The fact that it is now really the `mousedown` rather than the `mouseup` that triggers the pasting doesn't matter IMO.

50Wliu:
Since this fixes #8648, there should be at the very least a new spec to ensure that a double-paste no longer occurs that fails on master but passes on this branch.

dietmar:
Done: I have added a spec identical to the existing one, but with an additional left click before the middle click. This is a way to reproduce #8648 as described in https://github.com/atom/atom/issues/8648#issuecomment-141464493. This one fails on master (because of double paste) but passes on this branch.

ungb:
Was able to verify this change on ubuntu, Tested before and after and saw it fixed the issue when you don't move the mouse you get double paste.

PR #14080, opened 2017-03-28
Packages preloading

as-cii:
This pull request aims at reducing bundled packages loading and activation time even further by preloading them. During the snapshot phase, in fact, Atom will now scan all the core packages and load their main module, keymaps, menus, and settings. Later, at runtime, it will finish loading those packages by compiling their style sheets, calling `activate` on them, etc.

In addition, this pull request improves the way packages are resolved and, as a result, minimizes the I/O cost associated to it. Previously, in fact, because of the way the code was structured, we would get the list of package directories and constantly issue a `resolvePackagePath` for each of them. This, in turn, would unnecessarily verify again that each path was a directory, thus querying the file system and blocking the main thread. After these changes, we will perform that I/O just once and return a javascript object which contains all the information needed to load the associated package without extra cost.

On master, this is how loading and activating packages looks like:

screen shot 2017-03-28 at 12 10 41

screen shot 2017-03-28 at 12 09 41

This is how the same scenario looks like after the changes contained in this pull request:

screen shot 2017-03-28 at 12 05 44

screen shot 2017-03-28 at 12 07 39

You can notice how startup got `~ 90ms` faster, and packages loading alone became almost instantaneous. I am expecting this to be even more noticeable on machines that have slow hard drives, where the cost of I/O is much more dramatic.

This marks the end of the first round of performance improvements that can be achieved thanks to snapshots; there are still many things we can improve both in the renderer process and in the browser process, especially related to the scheduling of the tasks that currently happen as part of startup. I believe Atom could be a lot smarter about that, but I am going to illustrate what and how we could make it so in a separate issue (that is going to supplant the now outdated #13253).

/cc: @atom/maintainers

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "grammar-selector": "0.49.4",
     "image-view": "0.61.2",
     "incompatible-packages": "0.27.2",
-    "keybinding-resolver": "0.37.0",
+    "keybinding-resolver": "0.38.0",
     "line-ending-selector": "0.6.3",
     "link": "0.31.3",
     "markdown-preview": "0.159.10",
--- a/script/lib/generate-metadata.js
+++ b/script/lib/generate-metadata.js
@@ -48,7 +48,7 @@ function buildBundledPackagesMetadata () {
       }
     }
 
-    const packageNewMetadata = {metadata: packageMetadata, keymaps: {}, menus: {}, grammarPaths: []}
+    const packageNewMetadata = {metadata: packageMetadata, keymaps: {}, menus: {}, grammarPaths: [], settings: {}}
 
     packageNewMetadata.rootDirPath = path.relative(CONFIG.intermediateAppPath, packagePath)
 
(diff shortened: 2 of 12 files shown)


ungb:
✨  @as-cii.  I didn't see any issues with core packages on mac, saw huge performance gains.  

Will continue to test on windows and linux

ungb:
I didn't see any issues on linux or windows.  I tested this with a few different packages installed and tested a few core packages like markdown, autocomplete, etc. LGTM 🚀 it!

Windows:
Before:
image


After:

image

maxbrunsfeld:
👏

whmountains:
Can you also allow third-party packages to be preloaded and shapshoted?  Making a package "snapshot-friendly" would be like one step beyond making it lazy-load everything.

nathansobo:
@whmountains It's something we could explore in the next round of startup improvements. For now we've moved to editor performance for a while but we will cycle back.

PR #14093, opened 2017-03-29
penguin: Respect DISPLAY env variable in Unix like environments

J5:
### Requirements

Please advise on a test I can write. I have not been able to successfully compile HEAD so I'm not sure if I can test or if this code is complete. It is a small patch so feel free to just fork it. Note that I did manually test the concept by using the --socket-path option on startup and writing the socket path in the same way that it would be constructed in this code.

### Description of the Change

Atom connects to a socket to determine if it is already running
and if it should launch a new window in the same process.
This socket is named with the user's name but ignores the user's
display. If a user has multiple displays, such as running a
remote display server, Atom will pop up the new window on the display
that it was first started on.

This patch adds the display name to the socket name if it exists
in the env. Because a separate socket is being opened Atom does
the correct thing and launches on the display it was started on.

### Alternate Designs

No alternatives. On linux it should be using D-Bus to avoid these lifecycle/session issues but this is good enough with the benefit that it will work across different Unix like envs.

Atom windows should open up on the display they are started on.

### Benefits

Users who remote into a different display (such as I have to do for work) will be able to start Atom without first stopping Atom processes running on a different display.

### Possible Drawbacks

None. This is proper behavior. Applications should always respect the environment they are started in.

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -269,6 +269,25 @@ describe('AtomApplication', function () {
       await conditionPromise(async () => (await getTreeViewRootDirectories(reusedWindow)).length > 0)
     })
 
+    it('opens a new window with no project paths to open directories but with different DISPLAYs', async function () {
+      originalGetUnixDisplay = AtomApplication.getUnixDisplay;
+      mockUnixDisplay(':0')
+      const tempDirPath = makeTempDir()
+      const atomApplication = buildAtomApplication()
+      const window1 = atomApplication.launch(parseCommandLine([]))
+      await focusWindow(window1)
+
+      mockUnixDisplay(':1')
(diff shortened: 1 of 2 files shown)


J5:
Tried to make it testable. Let me know if there is a better way to mock. I will see if I can build this on my Fedora box at home.

J5:
Note that I got it to build successfully on Fedora 25 but apm test does not run correctly. Something about not being able to sandbox when launching chrome. Looking into it.

sadick254:
Hey @J5. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

PR #14104, opened 2017-03-31
Fixes #14102 to use ATOM_HOME if present.

### LINKED ISSUE #14102: ATOM_HOME environment variable should be used in all build scripts when available.
This follows on from #6349

There is currently an inconsistency with the the home directory that is used during execution of the build scripts.  Currently in the scripts, `ATOM_HOME` is not respected, however if `ATOM_HOME` is set during `scripts/build` `apm` uses that to place output into the dir specified.  So I can build Atom with an `ATOM_HOME` set but I can't clean that build because `scripts/clean` doesn't respect the var.

The solution is to fix https://github.com/atom/atom/blob/master/script/config.js#L18 to use `ATOM_HOME`, with the default to be `(HOME|USERPROFILE)\.atom`.

https://github.com/atom/atom/blob/master/script/lib/clean-caches.js#L11-L18 also needs to be updated to refer to `CONFIG.atomHomeDirPath`

This is tangentially related to #6345 in that USERPROFILE should not be used. However, in these cases you can't override the current behavior with an environment variable.

kierans:
To test:

`$ ATOM_HOME=<some dir> ./scripts/build`

Check that outputs exist in that dir

`$ATOM_HOME=<some dir> ./scripts/clean`

Check that dir cleaned.

Repeat with no `ATOM_HOME` to check default home dir works

### DIFF
--- a/script/config.js
+++ b/script/config.js
@@ -15,7 +15,7 @@ const intermediateAppPath = path.join(buildOutputPath, 'app')
 const symbolsPath = path.join(buildOutputPath, 'symbols')
 const electronDownloadPath = path.join(repositoryRootPath, 'electron')
 const homeDirPath = process.env.HOME || process.env.USERPROFILE
-const atomHomeDirPath = path.join(homeDirPath, '.atom')
+const atomHomeDirPath = process.env.ATOM_HOME || path.join(homeDirPath, '.atom')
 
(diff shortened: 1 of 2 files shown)


kierans:
Looking the CI output it's not obvious to me if the errors are related to this change.

PR #14116, opened 2017-04-03
Assign `Package.prototype.mainModulePath` eagerly for preloaded packages

as-cii:
Many packages currently use this instance variable instead of calling `Package.prototype.getMainModulePath`. With this pull request we will eagerly compute such variable for preloaded packages too and therefore prevent third party packages that rely on this implementation detail from breaking.

/cc: @Ben3eeE who noticed this. Can you confirm this fixes the issue you experienced? Thanks!

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -102,6 +102,10 @@ class Package
       @path = path.join(@packageManager.resourcePath, @path)
       @loadStylesheets()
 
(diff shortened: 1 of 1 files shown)


Ben3eeE:
@as-cii This fixes the problem I was seeing :tada:

Issue #14109, opened 2017-04-01
atom.showSaveDialogSync broken in version 1.15.0


cs150bf:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

`atom.showSaveDialogSync('some/chosen/path/to/file')` used to work (still worked in version 1.14.0) but doesn't work in version 1.15.0.

### Steps to Reproduce

1. Open Atom in debug mode
2. Open console panel
3. Create a new file and edit something in the editor
4. In the console panel, type `atom.showSaveDialogSync('some/chosen/path/filename')`

**Expected behavior:** A save dialog pops out with the filename and path pre-set according to the param. 

**Actual behavior:** [What actually happens] The save dialog doesn't set to designated path and filename.

**Reproduces how often:** Always in version 1.15.0.

### Versions

Atom    : 1.15.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0

### Additional Information

Another example: 

This function is used in the Markdown Preview package (https://github.com/atom/markdown-preview/blob/master/lib/markdown-preview-view.coffee#L292). When the markdown preview pane is opened, right clicking on the preview and selecting "Save as HTML" used to lead to a popped out save dialog with the "Save As..." field auto-filled with a filename ending with ".html". But in version 1.15.0, the filename is just "Untitled".

cs150bf:
## My suspect: 
I'm looking at https://github.com/atom/atom/commit/c6cae5b8fdfea982084b068a0b38a7c988c52f51 and I have a feeling that all those change related to `showSaveDialog` have to do with this problem.

## I traced this issue as follows (links to blame page):

[...]

Issue #13978, opened 2017-03-10
'Save as' opens a save dialog in a semi-random dir with 'Untitled' as the file name


benogle:
### Prerequisites

* In safe mode
* Was unable to find an existing issue
* Seems like it started happening a couple weeks ago. Electron upgrade?

### Description

Save as used to open a save dialog in the directory of the focused file with the file's name as the filename in the dialog. 

Now it opens in the root of the project (most of the time), with the filename as `Untitled`. Also the dialog is much larger than before.

<img width="1108" alt="screen shot 2017-03-10 at 10 52 42 am" src="https://cloud.githubusercontent.com/assets/69169/23808956/9d337cf8-0580-11e7-86e2-756aaaa4aafe.png">

### Steps to Reproduce

1. Open atom in safe mode
2. open a file
3. cmd-shift-s to save as
4. Get kind of useless dialog

**Expected behavior:** Open a dialog in the current directory of the active file, with the name being that of the active file

**Actual behavior:** The above dialog

**Reproduces how often:** 100%

### Versions

On master HEAD as of today: e35a00

Ben3eeE:
On Windows the dialog opens in the root of the project directory with nothing in the file name dialog.
Thanks for the report @benogle. Adding it to the list of potential regressions ⚡️

jbt:
Just chiming in to say that on Ubuntu the dialog has recently been opening in a background(!) window, with no directory or filename set (so it chooses "Recently Used"), so it looks like this affects all platforms similarly.

50Wliu:
This seems to be a regression in 1.15 that didn't occur in 1.14.

umpirsky:
@50Wliu I can confirm that.

ungb:
I am able to repro as well.

theurge14:
May be related, but this behavior seems to not allow copy and pasting into the filename in this dialog box starting in 1.15.0, had to revert to 1.14.4 to get this back.

markaduffy:
Just noticed this myself. Also noticed dragging a file on to atom, automatically add that files containing folder to project pane.

DLaicH:
I'm also having this issue on Arch Linux, not in safe mode. The file chooser dialog always goes straight to my home directory with a blank filename, and for some reason it always opens on my primary monitor even though the editor is on my second monitor.

mbdmbd:
[...]

PR #14117, opened 2017-04-03
Fix issue #14109 and #13978

cs150bf:
…b.com/atom/atom/issues/13978

First argument to dialog.showSaveDialog() needs to be a BrowserWindow

https://github.com/electron/electron/blob/master/docs/api/dialog.md
https://github.com/electron/electron/blob/master/lib/browser/api/dialog.js

### Requirements

### Description of the Change

This should fix some regression introduced by commit https://github.com/atom/atom/commit/c6cae5b8fdfea982084b068a0b38a7c988c52f51 related to `showSaveDialog` causing things such "Save As", "Save as HTML" to not work properly (default filename "Untitled" and default path missing when it should be autofilled with more "reasonable" ones--e.g. original filename and path). 

This commit fixes the erroneous usage in [`src/main-process/atom-window.coffee#L268`](https://github.com/atom/atom/blob/c6cae5b8fdfea982084b068a0b38a7c988c52f51/src/main-process/atom-window.coffee#L268) introduced in commit https://github.com/atom/atom/commit/c6cae5b8fdfea982084b068a0b38a7c988c52f51
```coffeescript
dialog.showSaveDialog(this, params)
```

Here, `this` refers to an instance of AtomWindow, when the function `dialog.showSaveDialog` requires the first argument to be a BrowserWindow [`electron/lib/browser/api/dialog.js#L24`](https://github.com/electron/electron/blob/master/lib/browser/api/dialog.js#L24). 

So I believe the fix is simply 
```coffeescript
dialog.showSaveDialog(@browserWindow, params)
```

This is consistent with with the function binding usage in [`src/main-process/atom-window.coffee#L86@c6cae5b`](https://github.com/atom/atom/blob/c6cae5b8fdfea982084b068a0b38a7c988c52f51/src/main-process/atom-window.coffee#L86) and with the **old** working function `showSaveDialog` in [`src/application-delegate.coffee#L158`](https://github.com/atom/atom/blob/ae9d949ee4e199d1884bed89083196fa375bbd34/src/application-delegate.coffee#L158)

### Alternate Designs

N/A

Bug fix.

### Benefits

Fixes bug (and see applicable issues)

### Possible Drawbacks

I cannot think of any at this moment.

### Applicable Issues

- https://github.com/atom/atom/issues/14109
- https://github.com/atom/atom/issues/13978

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -296,7 +296,7 @@ class AtomWindow
       title: 'Save File',
       defaultPath: @representedDirectoryPaths[0]
     }, params)
-    dialog.showSaveDialog(this, params)
+    dialog.showSaveDialog(@browserWindow, params)
 
   toggleDevTools: -> @browserWindow.toggleDevTools()
 


Zireael07:
Instead of linking to issues, mention them by the hash and number in PR title.

cs150bf:
Wrong branch.

PR #14131, opened 2017-04-04
Update build to pre-compile packages with per-package transpilation configs

BinaryMuse:
This allows us to bundle packages with per-package transpilation setups as allowed in #13101. It's very similar to the existing built-in Babel pre-transpilation, but looks in `package.json` for the `atomTranspilers` configuration and adds the appropriate `CompileCache` setup *before* we begin to transpile *any* paths.

/cc @kuychaco @as-cii

### DIFF
--- a/script/build
+++ b/script/build
@@ -43,7 +43,7 @@ const transpileBabelPaths = require('./lib/transpile-babel-paths')
 const transpileCoffeeScriptPaths = require('./lib/transpile-coffee-script-paths')
 const transpileCsonPaths = require('./lib/transpile-cson-paths')
 const transpilePegJsPaths = require('./lib/transpile-peg-js-paths')
-const transpilePerPackageTranspilationPaths = require('./lib/transpile-per-package-transpilation-paths')
+const transpilePackagesWithCustomTranspilerPaths = require('./lib/transpile-packages-with-custom-transpiler-paths.js')
 
 process.on('unhandledRejection', function (e) {
   console.error(e.stack || e)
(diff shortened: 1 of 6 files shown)


script/lib/transpile-per-package-transpilation-paths.js
```diff
@@ -0,0 +1,32 @@
...
+
+const CompileCache = require('../../src/compile-cache')
+const fs = require('fs')
+const glob = require('glob')
+const path = require('path')
+
+const CONFIG = require('../config')
+
+module.exports = function () {
+  console.log(`Transpiling per-package transpilation paths in ${CONFIG.intermediateAppPath}`)
```
as-cii: What do you think about rewording this as `Transpiling packages with a custom transpiler`?

script/build
```diff
@@ -52,6 +53,7 @@ process.on('unhandledRejection', function (e) {
 checkChromedriverVersion()
 cleanOutputDirectory()
 copyAssets()
+transpilePerPackageTranspilationPaths()
```
as-cii: Similarly to the comment below, how would you feel about renaming this to `transpilePackagesWithCustomTranspiler`?

src/package.coffee
```diff
@@ -86,7 +86,6 @@ class Package
     @loadMenus()
     @registerDeserializerMethods()
     @activateCoreStartupServices()
-    @registerTranspilerConfig()
```
as-cii: Nice catch! ⚡️

as-cii (approved):
I left two minor comments, but this looks good to me. Nice work here! ✨

PR #14137, opened 2017-04-04
Document pane item interface on Workspace

nathansobo:
We didn't have a single listing of all the workspace item methods. Thought with the new docks feature landing it was a good time.

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -34,28 +34,28 @@ const WorkspaceCenter = require('./workspace-center')
 //
 // ### Required Methods
 //
-// #### `getTitle` Returns the title of the item to display on its associated
+// #### `getTitle()` Returns the title of the item to display on its associated
 // #### tab.
 //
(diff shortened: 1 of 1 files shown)


matthewwithanm (approved):
❤️ awesome!

I think there are a few more you could document, unless you're intentionally leaving them out. From our [type def for PaneItems][1]: `getLongTitle()`, `getIconName()`, `onDidChangeIcon()`, `onDidChangeTitle()`. Also, docks support `getPreferredInitialWidth()` and `getPreferredInitialHeight()`.

[1]: https://github.com/facebook/nuclide/blob/e9c22b3039f1f04765f6aa46da34f066321b2ca1/flow-libs/atom.js.flow#L427-L437
[2]: https://github.com/facebook/nuclide/blob/e9c22b3039f1f04765f6aa46da34f066321b2ca1/pkg/nuclide-workspace-views/lib/types.js#L13-L31

PR #14143, opened 2017-04-05
Refine Dock UX and API

maxbrunsfeld:
* [x] Allow docks to be auto-resized by double clicking the resize handle
* [x] Don't open docks when single-clicking the resize handle
* [x] Docks have a `getElement` method, but other workspace classes still use the legacy view provider system. Add `getElement` methods on those classes instead, and get rid of all the remaining view providers in core.
* [x] Add Workspace APIs for tracking which pane container is focused: `getActivePaneContainer`, `onDidChangeActivePaneContainer`.
* [x] Make workspace methods like `getActivePane` return panes in docks if docks are focused.


Ben3eeE:
Don't know if this is correct place to post or if it should be a separate issue. Noticed today that double clicking the tab bar of a dock opens an untitled tab in the last focused pane instead of the dock. I can open an issue for this if you want ⚡️ 

docks

ungb:
@Ben3eeE I noticed that it always opens a new window in the center.  

I found two issues.

On windows, I can't move the tree view between docks(this works on mac, I haven't tired on linux)

dockmove
NOTE the difference between windows and mac, is mac has the `Projects` on top I believe

Another bug I can repro on mac and window, is if you start atom with the Telemetry report and dock it on the bottom or anywhere, it closes the window that is currently in the center. If there is none, it closes Atom.

telemetryreport

Ben3eeE:
@ungb For the first issue is that because of the `Always Show Tab Bar` setting in the tabs package?

ungb:
this setting doesn't seem to work on windows, which could be the issue.

tabs

nathansobo:
@ungb you can't drag items without a tab. So in the graphic what you're dragging is the root folder of the tree view rather than the tree-view itself. We probably need to get to the bottom of the setting.

nathansobo:
@ungb Are you sure you are using the dock-enabled tree view? Did you run `apm install` in the branch before launching dev mode?

PR #14146, opened 2017-04-06
Improve snapshotting

BinaryMuse:
* Bump electron-link
* Only pre-require core packages during snapshot generation


as-cii:
This looks good to me. Nice work! :shipit:

PR #14161, opened 2017-04-10
Fix a regression in hovered dock tracking

maxbrunsfeld:
This fixes a regression introduced in #14143 and backfills some tests for the dock toggle buttons.

This also fixes a minor bug where, after clicking the dock toggle button to hide the dock, the button would remain visible until the next mouse movement.

/cc @nathansobo @matthewwithanm

### DIFF
--- a/spec/workspace-element-spec.js
+++ b/spec/workspace-element-spec.js
@@ -1,4 +1,4 @@
-'use strict'
+/** @babel */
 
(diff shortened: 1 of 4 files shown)


nathansobo:
Looks great. Thanks for adding the tests.

matthewwithanm:
Thanks @maxbrunsfeld!

PR #14166, opened 2017-04-10
Ensure that atom-panel tags can be created via HTML

maxbrunsfeld:
I've also rewritten the logic for maintaining the `atom-panel` element to avoid the use of custom elements altogether.

🎩  @joefitzgerald


src/panel.js
```diff
@@ -0,0 +1,107 @@
...
+    if (this.element) this.element.remove()
+    this.emitter.emit('did-destroy', this)
+    return this.emitter.dispose()
+  }
+
+  getElement () {
+    if (!this.element) {
+      this.element = document.createElement('atom-panel')
+      if (!this.visible) this.element.style.display = 'none'
+      if (this.className) this.element.classList.add(...this.className.split(' '))
```
Alhadis: Wouldn't this be cleaner?

~~~js
this.element.classList.add(...this.classList);
~~~

maxbrunsfeld: The `Panel` doesn't have a property called `this.classList`; just a `className` string.

Alhadis: Right. My bad.

I'll steer clear of virtual DOMs and stick to working with the "physical" ones....

PR #14178, opened 2017-04-12
Prevent the resize handle from covering the center

matthewwithanm:
Currently, the absolutely positioned element is the only thing with a
size. That means that it'll cover the center. With this change, we give
a size to the (relatively-positioned) atom-dock element so that closed
docks take up canvas space instead.

I think @simurai mentioned this problem before but I wasn't following
and then forgot about it. 😖

cc @nathansobo @maxbrunsfeld

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "autocomplete-atom-api": "0.10.1",
     "autocomplete-css": "0.16.1",
     "autocomplete-html": "0.7.3",
-    "autocomplete-plus": "2.35.2",
+    "autocomplete-plus": "2.35.3",
     "autocomplete-snippets": "1.11.0",
     "autoflow": "0.29.0",
     "autosave": "0.24.2",
--- a/src/dock.js
+++ b/src/dock.js
@@ -205,11 +205,16 @@ module.exports = class Dock {
 
     const shouldBeVisible = state.visible || state.showDropTarget
     const size = Math.max(MINIMUM_SIZE, state.size == null ? this.getInitialSize() : state.size)
+    const resizeHandleSize = this.resizeHandle.getSize()
 
     // We need to change the size of the mask...
-    this.maskElement.style[this.widthOrHeight] = `${shouldBeVisible ? size : this.resizeHandle.getSize()}px`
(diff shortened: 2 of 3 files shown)


simurai (approved):
Oh.. right. I guess [that comment](https://github.com/atom/atom/pull/13977#discussion_r108034302) got collapsed after a code change.

It's [fixed](https://github.com/atom/one-dark-ui/blob/586f4fdbd386c74f3ab5042f056fb82c67bf9438/styles/docks.less#L13-L19) in the One themes, but much better to inline it so the number stays in sync. 👍

nathansobo:
@matthewwithanm These CI failures seem legitimate. Do you think we could release to beta without this today and hotfix it once we get a green build? Seems like we could start getting feedback from beta users.

matthewwithanm:
@nathansobo yeah, that seems fine

matthewwithanm:
Button positioning at window sizes when One shows outer padding and when it doesn't:

screen shot 2017-04-12 at 11 36 45 am
screen shot 2017-04-12 at 11 36 35 am

@simurai I think we'll want to remove the 4px min-width/height from One.

matthewwithanm:
oh, One doesn't add the border anymore XD

Here's an updated screenshot taken with atom/one-light-ui#98:

screen shot 2017-04-12 at 12 12 43 pm

@simurai seriously, great work (:

simurai:
> @matthewwithanm what do you think about making the toggle buttons light (`@tool-panel-background-color`?) when the docks are collapsed?

Sure. Actually, now that there is no resize-handle when collapsed, we could absolute position it when open. And then have just a `1px` border. 

Before | After
--- | ---
screen shot 2017-04-13 at 4 59 54 pm | screen shot 2017-04-13 at 4 59 18 pm

Looks less heavy and more seamless with the docks. I tried that before, but there were issues when dragging that now seem to be resolved. 🎉 

Already pushed to https://github.com/atom/one-light-ui/pull/98 + https://github.com/atom/one-dark-ui/pull/201. 😇

iKlsR:
@matthewwithanm What's that cool sidebar in your editor?

matthewwithanm:
@iKlsR that's @suda's tool-bar, populated with a bunch of Nuclide features (:

PR #14210, opened 2017-04-14
Reduce dock initial size lookups

### LINKED ISSUE #14209: On open, tree-view opens half shown and expands only when mouse action occurs in Atom
### Prerequisites

### Description

tree-view opens half shown and expands only when mouse action occurs in Atom
### Steps to Reproduce

1. Open a project in atom with `atom .`

image

image

### Versions
Atom    : 1.18.0-dev-83efa4b91
Electron: 1.3.14
Chrome  : 52.0.2743.82
Node    : 6.5.0

BETA:
Atom    : 1.17.0-beta1
Electron: 1.3.14
Chrome  : 52.0.2743.82
Node    : 6.5.0

### Additional Information

gif:
cropped

/cc @nathansobo @matthewwithanm @maxbrunsfeld

matthewwithanm:
Previously, we would get the initial size every time we didn't have an
explicit one. With this commit, we only get the initial size when we
deserialize and when we go from 0 -> 1 pane items.

Also, if the dock doesn't already have an explicit size, we'll use the
preferred size of the item being dragged when peaking the dock. That
way, dropping it won't cause it to change size.

This kinda-but-not-really fixes #14209. The tree view will still report sizes inconsistently but, since we only ask it once, we won't see the symptom.

cc @ungb @nathansobo

### DIFF
--- a/src/dock.js
+++ b/src/dock.js
@@ -56,6 +56,7 @@ module.exports = class Dock {
         this.didActivate(this)
       }),
       this.paneContainer.observePanes(pane => {
+        pane.onDidAddItem(this.handleDidAddPaneItem.bind(this))
         pane.onDidRemoveItem(this.handleDidRemovePaneItem.bind(this))
       }),
       this.paneContainer.onDidChangeActivePane((item) => params.didChangeActivePane(this, item)),
@@ -206,7 +207,11 @@ module.exports = class Dock {
     }
 
     const shouldBeVisible = state.visible || state.showDropTarget
-    const size = Math.max(MINIMUM_SIZE, state.size == null ? this.getInitialSize() : state.size)
+    const size = Math.max(MINIMUM_SIZE,
+      state.size ||
+      (state.draggingItem && getPreferredSize(state.draggingItem, this.location)) ||
+      DEFAULT_INITIAL_SIZE
+    )
 
     // We need to change the size of the mask...
     this.maskElement.style[this.widthOrHeight] = `${shouldBeVisible ? size : 0}px`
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
✨  This is a really nice change. Do you have time to add some tests for this?

src/dock.js
```diff
@@ -340,13 +351,13 @@ module.exports = class Dock {
...
-    const activePaneItem = this.paneContainer.getActivePaneItem() || this.paneContainer.getPaneItems()[0]
-    if (activePaneItem != null) {
-      initialSize = getPreferredSize(activePaneItem, this.location)
-    }
-    return initialSize == null ? DEFAULT_INITIAL_SIZE : initialSize
+    const activePaneItem = this.paneContainer.getActivePaneItem() || items[0]
+    let initialSize
+    if (activePaneItem) initialSize = getPreferredSize(activePaneItem, this.location)
+    if (initialSize == null && items.length) initialSize = DEFAULT_INITIAL_SIZE
+    return initialSize ? Math.max(MINIMUM_SIZE, initialSize) : null
```
matthewwithanm: I gotta remove this `Math.max()` call. It's gonna be done in `render()` so this is redundant.

matthewwithanm: Don't need the `items.length` check either. If there are items, there'll be an active one.

matthewwithanm:
@maxbrunsfeld 🎉

maxbrunsfeld:
:zap: Nice. Thanks @matthewwithanm!

PR #14217, opened 2017-04-15
Don't attempt to transpile directories

edjubuh:
Enable option in glob to disable directories from being returned when
finding paths to transpile during compilation

### Description of the Change
Before this PR, building Atom involved transpiling all babel paths, coffeescript files, cson files, etc. by globbing for all paths matching (essentially) `**/*.js`, `**/*.coffee`, etc. Node modules like term.js appear in the directory structure as `node_modules/term.js/<contents>`, which previously caused the folder `node_modules/term.js` to be added to a list of paths to transpile under the assumption that it was a file. This causes an error when [trying to read the folder](https://github.com/atom/atom/blob/master/src/compile-cache.js#L84)) `node_modules/term.js`. This PR fixes that by enabling the `nodir` option, which filters out any directories from being in the results. While I only specifically witnessed this issue in finding javascript files, the bug also exists in the coffeescript, CSON, PEG.js, and custom transpilers if someone names a directory *.coffee, *.cson, or something else expected to be transpiled.

### Alternate Designs
One could check if the path attempting to be transpiled is actually a directory where the error occurs right before transpiling, but glob already checks if the path is a directory.

Proposed update to building atom

### Benefits
Can include node_modules named like module.js.

### Possible Drawbacks
In the future, if the [transpiler ](https://github.com/atom/atom/blob/master/src/compile-cache.js#L83) (or a custom one) may support directories? In this event, a glob pattern that ends with a `/` still grabs directories (even if `nodir` is enabled).

### Applicable Issues
None

### DIFF
--- a/script/lib/transpile-babel-paths.js
+++ b/script/lib/transpile-babel-paths.js
@@ -25,13 +25,13 @@ module.exports = function () {
 
 function getPathsToTranspile () {
   let paths = []
-  paths = paths.concat(glob.sync(path.join(CONFIG.intermediateAppPath, 'benchmarks', '**', '*.js')))
-  paths = paths.concat(glob.sync(path.join(CONFIG.intermediateAppPath, 'exports', '**', '*.js')))
-  paths = paths.concat(glob.sync(path.join(CONFIG.intermediateAppPath, 'src', '**', '*.js')))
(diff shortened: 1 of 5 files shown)


as-cii:
Thanks, @edjubuh!

PR #14221, opened 2017-04-15
Don't close permanent dock items

matthewwithanm:
This prevents permanent dock items from being closed, e.g. with cmd-w.

It also adds a new `force` argument to `Pane::destroyItem()` which devs can use to forcefully destroy permanent items; for example, when their package is deactivated.

cc @nathansobo


maxbrunsfeld:
Just got the tree-view tests passing. I'm wondering if we even need the `force` flag, or we package authors who want to force the item to be removed should just destroy their item and cause it to call its `onDidDestroy` listeners.

@matthewwithanm thoughts?

matthewwithanm:
imo adding an event emitter and implementing those methods is a big enough hurdle that people won't do it. also, having users implement it n times (in every view) instead of once in Atom feels wrong to me.

maxbrunsfeld:
Yeah, that's true; some packages' views won't already have an emitter. Ok, I'm cool w/ the boolean.

PR #14254, opened 2017-04-19
Show correct beta commands in installer dialog

### LINKED ISSUE #9361: Wrong command names shown after installing shell commands in Atom Beta
When you install the shell commands in Atom Beta, the confirmation message says, "The shell commands `atom` and `apm` are installed." It should instead say, "The shell commands `atom-beta` and `apm-beta` are installed."

To reproduce this issue, choose "Atom" in the menu bar, and then choose "Install Shell Commands".
#### Screenshot

screenshot
#### Version info

Atom version: 1.2.0-beta0
OS version: OS X 10.11.1

jamiebuilds:
Resolves #9361

Note that I haven't actually tested this out, I wasn't sure how and it was a small enough change that I figured looking over the code would be enough.

### DIFF
--- a/src/command-installer.coffee
+++ b/src/command-installer.coffee
@@ -35,6 +35,18 @@ class CommandInstaller
 
   getResourcesDirectory: ->
     process.resourcesPath
+    
+  getAtomCommandName: ->
+    if @appVersion.includes("beta")
+      "atom-beta"
+    else
+      "atom"
+
+  getApmCommandName: ->
+    if @appVersion.includes("beta")
+      "apm-beta"
+    else
+      "atom"
 
   installShellCommandsInteractively: ->
     showErrorDialog = (error) =>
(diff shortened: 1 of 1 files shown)


src/command-installer.coffee
```diff
@@ -35,6 +35,18 @@ class CommandInstaller
...
+    if @appVersion.includes("beta")
+      "atom-beta"
+    else
+      "atom"
+
+  getApmCommandName: ->
+    if @appVersion.includes("beta")
+      "apm-beta"
+    else
+      "atom"
```
petetnt: Should this be `apm`?

jasonrudolph:
Thanks for this contribution. The underlying issue (#9361) was resolved at some point in the past several months, so I'm going to close this out.

Here's the current behavior in Atom 1.31.0-beta0:

image

Issue #14023, opened 2017-03-17
Folding keybindings duplicated


jrial:
The keybinding for Fold All is `ctrl-alt-{` and for Fold Selection is `ctrl-alt-shift-[`. On a US keyboard, these are the same. I didn't override these bindings, and they remain duplicated when I start in safe mode.

Secondary problem: Under Settings => Keybindings, Fold All is shown as `ctrl-alt-{`, while in the command palette it's shown as `ctrl-alt-shift-[`. See screenshots below. Result: even on a keyboard layout that shouldn't be susceptible to this key binding duplication (on my Belgian layout, `[` and `{` don't share a physical key), Fold Selection still gets overridden by Fold All.

Keybindings according to Settings:
![2017-03-17 13_49_46-mint - vmware workstation](https://cloud.githubusercontent.com/assets/570247/24043897/212eb746-0b19-11e7-9f9b-b81464623e7d.png)

Keybindings according to Command Palette:
![2017-03-17 13_51_44-mint - vmware workstation](https://cloud.githubusercontent.com/assets/570247/24043898/212f1a88-0b19-11e7-9abe-2c18d07525e5.png)

50Wliu:
Confirmed.  However, please completely fill out the issue template next time.

Arcanemagus:
Just to note (since I was about to hit submit on my own version of this bug before I finally found this in a search): This only affects Windows and Linux as macOS uses a different keybind for this.

@mmoedinhas's #16287 PR solves this by changing `editor:fold-all` / `editor:unfold-all` to <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>f</kbd> / <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>u</kbd> on all platforms.

PR #14262, opened 2017-04-19
Fix folding key bindings

nikselite:
### Description of the Change

The documentation for Linux and Windows says that you need to use `Ctrl+Alt+F` key binding to fold arbitrary sections of your code or text. However, in `linux.cson` and `win32.cson` declared a
`Ctrl-Alt-Shift-[` key binding, which leads to incorrect processing of keystroke and the appearance of duplicates. `Ctrl-Alt-Shift-[` key binding processed as `Ctrl-Alt-{` key binding which is already declared to fold everything.

| Before         | After          |
| :------------: | :------------: |
|before|after|

### Alternate Designs

None.

Bug fixes.

### Benefits

Correct folding key bindings.

### Possible Drawbacks

None.

### Applicable Issues

#14023

### See Also

http://flight-manual.atom.io/using-atom/sections/folding/

### DIFF
--- a/keymaps/linux.cson
+++ b/keymaps/linux.cson
@@ -115,7 +115,7 @@
 'atom-workspace atom-text-editor:not([mini])':
   # Atom specific
   'ctrl-<': 'editor:scroll-to-cursor'
-  'ctrl-alt-f': 'editor:fold-selection'
+  'ctrl-alt-shift-[': 'editor:fold-selection'
 
   # Sublime Parity
   'ctrl-enter': 'editor:newline-below'
--- a/keymaps/linux.cson
+++ b/keymaps/linux.cson
@@ -115,7 +115,7 @@
 'atom-workspace atom-text-editor:not([mini])':
   # Atom specific
   'ctrl-<': 'editor:scroll-to-cursor'
-  'ctrl-alt-shift-[': 'editor:fold-selection'
+  'ctrl-alt-f': 'editor:fold-selection'
 
   # Sublime Parity
   'ctrl-enter': 'editor:newline-below'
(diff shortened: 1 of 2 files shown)


winstliu (approved):
Looks good to me.
/cc @atom/maintainers

damieng:
We can't use CTRL-ALT-alpha bindings on Windows or Linux as they interfere with international keyboard keymaps.

damieng:
If we want a default keybinding for this we'll need to find an alternative.

maxbrunsfeld:
We should probably just remove these bindings by default.

thomasjo:
> We should probably just remove these bindings by default.

Agreed.

nikselite:
Maybe we should use combinations:

| Before             | After               | Command                      |
| :----------------: | :-----------------: | :--------------------------: |
| `ctrl-alt-shift-[` | `ctrl-k ctrl-alt-f` | `editor:fold-selection`      |
| `ctrl-alt-[`       | `ctrl-k ctrl-alt-[` |  `editor:fold-current-row`   |
| `ctrl-alt-]`       | `ctrl-k ctrl-alt-]` |  `editor:unfold-current-row` |
| `ctrl-alt-{`       | `ctrl-k ctrl-alt-{` |  `editor:fold-all`           |
| `ctrl-alt-}`       | `ctrl-k ctrl-alt-}` |  `editor:unfold-all`         |
| `ctrl-alt-=`       | `ctrl-k ctrl-alt-=` |  `pane:increase-size`        |
| `ctrl-alt--`       | `ctrl-k ctrl-alt--` |  `pane:decrease-size`        |

PR #14269, opened 2017-04-20
Update apm to use async keytar with libsecret

BinaryMuse:


BinaryMuse:
@as-cii So it looks like if you have libsecret installed but no providers that implement `org.freedesktop.secrets`, you'll get a message like

> ** Message: Remote error from secret service: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.secrets was not provided by any .service files

It seems that GNOME Keyring provides this; keytar started working in another project when I changed a build to include `gnome-keyring` and also started the `gnome-keyring-daemon` during the tests.

```diff
+before_script:
 +  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then eval $(echo -n "" | /usr/bin/gnome-keyring-daemon --login); fi
 +  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then eval $(/usr/bin/gnome-keyring-daemon --components=secrets --start); fi
 +  - |
 +    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
 +      dbus-launch /usr/bin/python -c \
 +        "import gnomekeyring;gnomekeyring.create_sync('login', '');"
 +    fi
 +
```

I guess the question is, should we continue to include GNOME Keyring as a dependency as well as libsecret? Or should we leave that up to the user?

ungb:
looks good, saw this work on ubuntu for debian build and on centos for rpm build.  

image


I had to use the build from https://travis-ci.org/atom/atom/builds/226594537,  I had a successful rpm build on ubuntu,  but when I tried to use that version on centos it didn't work and gave me an error.

BinaryMuse:
@as-cii Can you think of anything else this needs?

BinaryMuse:
@ungb Thank you so much for doing all that testing! :bowing_woman:

as-cii (approved):
Looks good to me! ✨

PR #14271, opened 2017-04-20
Do not treat packages in unusual locations as bundled despite name

damieng:


maxbrunsfeld:
/cc @as-cii 

We should do another 1.17 beta release with this sometime soon.

src/package-manager.coffee
```diff
@@ -426,7 +426,7 @@ class PackageManager
       pack
     else if packagePath = @resolvePackagePath(nameOrPath)
       name = path.basename(nameOrPath)
-      @loadAvailablePackage({name, path: packagePath, isBundled: @isBundledPackage(name)})
+      @loadAvailablePackage({name, path: packagePath, isBundled: @isBundledPackagePath(name)})
```
as-cii: Shouldn't we use `packagePath` here?

maxbrunsfeld: Yeah.

PR #14276, opened 2017-04-21
Fix bundled package path check to use path, not name

maxbrunsfeld:
Addresses https://github.com/atom/atom/pull/14271#discussion_r112624466

/cc @damieng @as-cii


as-cii (approved):
Looks good! 👍

damieng (approved):

PR #14293, opened 2017-04-25
Update Electron to 1.3.15

BinaryMuse:
This fixes an issue where accessing properties on the `remote` module occasionally threw an exception.


BinaryMuse:
This is pretty low-risk given the patch introduced by Electron 1.3.15. I've also tested this manually and all specs pass, so I'm going to merge this now.

PR #14309, opened 2017-04-26
Recalculate Tooltip positions when their contents change

BinaryMuse:
### Description of the Change

This PR implements `Tooltip#recalculatePosition` which resets the Tooltip's position based on its current size. It also adds a `MutationObserver` that watches for the tooltip to change and calls the aforementioned method on the next animation frame when it does.

### Alternate Designs

`ResizeObserver` is not available until Electron 1.6 (with a flag), so a `MutationObserver` covers the majority of the important use cases until then.

I originally implemented this so the resizing operation took place in an `atom.views.pollDocument` call, but this made the tooltip feel very jerky, so I put it in an rAF instead. Happy to hear other 💭 s.

/cc @nathansobo

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -1,3 +1,5 @@
+/* global MutationObserver */
+
 'use strict'
 
 const EventKit = require('event-kit')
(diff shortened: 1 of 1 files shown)


src/tooltip.js
```diff
@@ -103,6 +106,24 @@ Tooltip.prototype.init = function (element, options) {
...
+  this.mutationObserver.observe(this.getTooltipElement(), {
+    attributes: true, childList: true, characterData: true, subtree: true
+  })
+}
+
+Tooltip.prototype.stopObservingMutations = function () {
+  this.mutationObserver.disconnect()
+}
+
+Tooltip.prototype.handleMutations = function (mutations) {
```
kuychaco: :fire: unused parameter

kuychaco (approved):

BinaryMuse:
I'm going to merge this so we can move forward in the dependant packages, but happy to revisit details before next railcar.

PR #14313, opened 2017-04-27
Add snap install instructions

flexiondotorg:
Atom is now published in the Snap store, this pull request add instructions for installing the snap. Atom is currently in the beta channel but when it's promoted to stable I will file a another pull request to update the documentation.


lee-dohm:
Thank you for the contribution.

We don't list instructions for unofficial distribution points in our documentation since we can't guarantee the quality of the download. Because of this, we won't be accepting this change.

PR #14320, opened 2017-04-28
Unbreak URL handlers in Atom 1.17-beta

hansonw:
This fixes a regression caused by https://github.com/atom/atom/commit/c4d0944c0da51e0a2d88273ff58a47c469d07be8.

The `urlOpen` function in AtomApplication is still constructing the `PackageManager` in the old way, but we need to use `initialize` now. (Let me know if there is a better way to do this now that we have snapshots!)

Released under CC0.


as-cii (approved):
Looks great! Apologies for the oversight. 🙏

PR #14348, opened 2017-05-03
Ensure binary compatibility checks with/without the module cache are consistent

hansonw:
The binary compatibility check normally only looks at `.node` files inside `build/Release`. However, the check with the cached `_atomModuleCache` looks at _all_ `.node` files no matter where they are.

This can cause all sorts of issues depending on where the `.node` files are. Since `_atomModuleCache` blindly indexes all files in the package, this can false-alarm on incompatible binaries even if they're never required.

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -658,10 +658,13 @@ class Package
 
     @activationHooks = _.uniq(@activationHooks)
 
+  getNativeBindingDirectory: (modulePath) ->
+    path.join(modulePath, 'build', 'Release')
+
   # Does the given module path contain native code?
   isNativeModule: (modulePath) ->
     try
-      fs.listSync(path.join(modulePath, 'build', 'Release'), ['.node']).length > 0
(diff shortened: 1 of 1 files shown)


hansonw:
This causes issues for Nuclide since we bundle a prebuilt binaries dependency (https://github.com/facebooknuclide/nuclide-prebuilt-libs) for the Mac and Linux platforms. If you're not on one of those, you can be SOL...

hansonw:
@BinaryMuse: @damieng suggested that you might be a good reviewer :) @damieng was running into this error when installing Nuclide from apm.

sadick254:
Hey @hansonw. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

Issue #14252, opened 2017-04-18
Able to pinch + zoom Atom on touchpad on mac


ungb:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Able to pinch + zoom Atom Editor when user shouldn't be able to do this.

### Steps to Reproduce

1. Open a project in Atom (1.17-beta2 or master)
2. Go to tree-view and pinch zoom like you would on chrome

**Expected behavior:** [What you expect to happen]
It should do nothing.  On stable 1.16 this works

**Actual behavior:** [What actually happens]
It zooms the atom editor 

**Reproduces how often:** [What percentage of the time does it reproduce?]
100% of the time.

### Versions
Atom    : 1.18.0-dev-643ba6e08
Electron: 1.3.14
Chrome  : 52.0.2743.82
Node    : 6.5.0

### Additional Information

on Mac Sierra 10.12.4

Any additional information, configuration or data that might be necessary to reproduce the issue.
![pinchzoom](https://cloud.githubusercontent.com/assets/5624870/25151180/7009a108-2439-11e7-915a-31f7886b6c9e.gif)

/cc @maxbrunsfeld @nathansobo @matthewwithanm Could this be related to docks changes?

whmountains:
I can reproduce this too.  Just pinch-zoom anywhere on the window.  Works whether or not the tree-view is open.

PR #14352, opened 2017-05-04
Disable zoom only after the window has finished loading

as-cii:
Fixes https://github.com/atom/atom/issues/14252.

Previously, we were disabling the zoom functionality right when creating an `AtomWindow` instance, which doesn't work possibly due to `webContents` not being ready yet.

With this pull request we will disable zooming in/out only after receiving the `window:loaded` event from the renderer process.

/cc: @ungb @iolsen


iolsen:
👍 tests successfully for me.

PR #14355, opened 2017-05-04
Fix SyntaxError while building v1.16.0 release on Ubuntu 17.04 i686

NicolasHov:
Fix **SyntaxError** while building v1.16.0 release on Ubuntu 17.04 i686 with `./script/build --create-debian-package`

```
~/atom/script/lib/code-sign-on-mac.js:13
  let certPath = process.env.ATOM_MAC_CODE_SIGNING_CERT_PATH;
    ^^^

    SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

Thank's to @yvan-sraka for it's helpful debug!

### DIFF
--- a/script/lib/code-sign-on-mac.js
+++ b/script/lib/code-sign-on-mac.js
@@ -10,7 +10,7 @@ module.exports = function (packagedAppPath) {
     return
   }
 
-  let certPath = process.env.ATOM_MAC_CODE_SIGNING_CERT_PATH;
+  var certPath = process.env.ATOM_MAC_CODE_SIGNING_CERT_PATH;
   if (!certPath) {
     certPath = path.join(os.tmpdir(), 'mac.p12')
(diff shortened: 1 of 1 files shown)


50Wliu:
Node 6+ is [_required_](https://github.com/atom/atom/blob/master/docs/build-instructions/linux.md#requirements) to build Atom.

yvan-sraka:
Currently, the Node version offered by the official APT sources of the latest version of Ubuntu is `4.7.2`: http://packages.ubuntu.com/zesty/nodejs ; and this change of 3 characters is the only thing we need to support this legacy version of Node, so why not?

lee-dohm:
We want to keep the Node we program against in Atom's build process as close as possible to the Node we program against in the rest of the Atom project. It makes our lives as maintainers simpler.

Issue #14369, opened 2017-05-07
Atom 1.17.0-beta5 - onDidChangeActivePane and onDidStopChangingActivePaneItem on atom.workspace.getCenter() do not work.


hswolff:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

`atom.workspace.getLeftDock().onDidChangeActivePane` and `atom.workspace.getCenter().onDidStopChangingActivePaneItem` do not work.

### Steps to Reproduce

1. Open up Dev tools.
2. Type in either `atom.workspace.getCenter().onDidStopChangingActivePaneItem(console.log)` or `atom.workspace.getCenter().onDidChangeActivePane(console.log)`.
3. Switch tabs in center dock, observe that nothing is logged in the console.

**Expected behavior:** [What you expect to happen]
It should log the new pane that is active.

**Actual behavior:** [What actually happens]
Nothing is logged.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100% of the time.

### Versions

Atom 1.17.0-beta5
MacOS 10.12.4

### Additional Information

These two event handlers work on the workspace itself via `atom.workspace.onDidStopChangingActivePaneItem(console.log)` or `atom.workspace.onDidChangeActivePane(console.log)`.

I'm looking to use these methods for my [tree-view-extended](https://github.com/hswolff/tree-view-extended) package.

BinaryMuse:
I can repo this on `1.18.0-dev-df8dd8b45`.

/cc @nathansobo @matthewwithanm

matthewwithanm:
I can't repro with `onDidChangeActivePane()`. Are you sure this isn't being triggered for you when you change the pane? (Just the item won't do it.)

I can repro with `onDidStopChangingActivePaneItem()` though and I think I see why—that functionality's only implemented in `workspace`. `PaneContainer` lets you subscribe to the event but never emits it. 😩

PR #14375, opened 2017-05-08
Fix `PaneContainer::onDidStopChangingActivePaneItem()`

matthewwithanm:
We we're allowing subscription to this event, but never actually triggering it 😨

This is mostly copypasta from the implementation in workspace. lmk if you'd prefer something different.

I also added a test for `onDidChangeActivePane()` which I can separate into another commit if you want or drop entirely. (I was attempting to repro the other part of #14369 but it seems to work for me.)

cc @BinaryMuse @nathansobo

### DIFF
--- a/spec/pane-container-spec.coffee
+++ b/spec/pane-container-spec.coffee
@@ -158,6 +158,32 @@ describe "PaneContainer", ->
       pane2.activate()
       expect(observed).toEqual [pane1.itemAtIndex(0), pane2.itemAtIndex(0)]
 
+  describe "::onDidStopChangingActivePaneItem()", ->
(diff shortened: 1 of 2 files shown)


matthewwithanm:
Do I need to avoid `setTimeout` for the snapshot to work?

BinaryMuse:
Ah, yeah, if one of these is created during the snapshotting phase, then globals like `setTimeout` aren't available.

nathansobo:
@damieng do you think the timeouts on 32-bit windows are spurious failures?

nathansobo:
@damieng Nevermind. The build failures are my fault.

nathansobo:
Rebased. I think our build should be green no.

PR #14397, opened 2017-05-11
Don't open dev tools on uncaught errors in headless tests

nathansobo:
This is an attempt to debug mysterious failures we're seeing on CI. Our theory is that the dev tools are getting opened due to an uncaught exception and causing subsequent tests to fail.

/cc @as-cii @maxbrunsfeld

### DIFF
--- a/script/test
+++ b/script/test
@@ -41,7 +41,7 @@ function runCoreMainProcessTests (callback) {
   console.log('Executing core main process tests'.bold.green)
   const cp = childProcess.spawn(executablePath, testArguments, {
     stdio: 'inherit',
-    env: Object.assign({}, process.env, {ATOM_GITHUB_INLINE_GIT_EXEC: 'true'})
+    env: Object.assign({ELECTRON_ENABLE_LOGGING: 1}, process.env, {ATOM_GITHUB_INLINE_GIT_EXEC: 'true'})
   })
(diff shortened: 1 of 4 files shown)


as-cii:
The build for 2cb84bf exhibited the same hypothetical uncaught exception, but it seems like dev tools got opened anyways. 

In e58a911 I added the [ELECTRON_ENABLE_LOGGING](https://github.com/electron/electron/blob/master/docs/api/environment-variables.md#electron_enable_logging) environment variable, which should emit every log message from the renderer process (including the one [reporting that the document isn't focused](https://github.com/atom/atom/blob/4d2641962e6b9d70948792f7f1d6f7bf46725e38/spec/spec-helper.coffee#L213-L214)). I restarted the build for that commit several times without ever being able to reproduce the issue again. 

I suspect all those logging statements are skewing timings enough to obscure the race condition.

nathansobo:
I'm tempted to just abandon this PR. @as-cii do you think it's worth changing this?

as-cii:
Seems like on master builds have gotten less flaky, so yeah let's close this and maybe reopen if spurious errors show up again in the future.

Arcanemagus:
Closing this based on the above discussion, please re-open this if it's still something that is needed!

PR #14398, opened 2017-05-12
Fix running packages preloading tests in interactive mode

as-cii:
This change also makes the test more useful, as it verifies that the metadata we read from the package.json's `packagesCache` is used correctly.


as-cii:
The appveyor build failed for some other reason, so I'll go ahead and 🚢 this.

Issue #14173, opened 2017-04-11
Switching tabs does not focus editor


50Wliu:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

After switching tabs, the focus does not get transferred to the editor.

### Steps to Reproduce

1. Open two tabs and activate the first tab.
2. Click on the second tab to switch to it.

**Expected behavior:** Be able to start typing immediately

**Actual behavior:** Nothing happens when you type.  Instead, you have to click on the editor first.

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.17.0-dev-4b77f9050
Electron: 1.6.5
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

```
apm  1.17.0
npm  3.10.5
node 6.9.5 x64
python 2.7.12
git 2.12.0.windows.1
visual studio 2015
```

### Additional Information

This may only be on the Electron 1.6.x branch.  In addition, it also happens when reloading the window, in which case the Tree View gets focus.

maxbrunsfeld:
I reproduced this on master (not just the 1.6.x branch).

PR #14403, opened 2017-05-12
Pane shouldn't steal focus from its children

matthewwithanm:
This was removed in #14175 in order to solve #14173 (editors not being
focused when clicking tabs). However, it means that the pane steals
focus from its children. The solution is to add the guard back and to
solve #14173 in another way: by delaying the activation of the item
(see atom/tabs#439).

### DIFF
--- a/src/pane-element.coffee
+++ b/src/pane-element.coffee
@@ -67,7 +67,7 @@ class PaneElement extends HTMLElement
 
   activated: ->
     @isActivating = true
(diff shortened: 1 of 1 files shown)


nathansobo (approved):
As soon as we get a green build on https://github.com/atom/tabs/pull/439 this looks good. Thanks again!

as-cii:
Thanks!

PR #14405, opened 2017-05-13
Updated instructions for Ubuntu

iamadarshkumar:
Debian Linux (Ubuntu) installation alternative using terminal.

### Requirements

*Ubuntu 14.04 and above


50Wliu:
Sorry, but the installation instructions are only for official installation methods.

PR #14419, opened 2017-05-15
Remove jQuery from dependencies

pablomayobre:
### Description of the Change
jQuery was removed from `package.json > dependencies` section

After a search through this repository I found that jQuery isn't being used anywhere but is still part of `package.json` `dependencies`.

### Benefits

Finally removes jQuery from Atom!

### Possible Drawbacks

There may be something still using jQuery in there, so this needs to pass the CI before actually being done


50Wliu:
✅

PR #14420, opened 2017-05-15
Don't show restore dialog for non-editor dock items

matthewwithanm:
The Nuclide file tree was causing this dialog to be shown every time.
With this commit, we exclude non-editor dock items from the "dirty"
check.

cc @BinaryMuse @maxbrunsfeld


iolsen (approved):

zertosh:
Thanks @matthewwithanm!!! 😄

PR #14421, opened 2017-05-15
Update some of the out-of-date dependencies

pablomayobre:
### Description of the Change
Two outdated packages (dedent and atom-select-list) were upgraded a minor release (from 0.6.0 to 0.7.0 and from 0.1.0 to 0.2.0 respectively)

### Why Should This Be In Core? / Benefits

Tries to decrease the number of outdated dependencies in Atom. 

### Possible Drawbacks

Needs to pass CI to see if there are no regressions

#### ⚠️ NOTE
I will also try to remove the color dependency since it's unused throughout this repository. I want to first see if this commit passes the necessary tests before trying that out

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "async": "0.2.6",
     "atom-keymap": "8.1.2",
-    "atom-select-list": "^0.1.0",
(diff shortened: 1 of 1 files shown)


Positive07:
It fails so I'll be closing this for now...

PR #14428, opened 2017-05-16
Remove 'color' dependency

winstliu:
This dependency appears to be unused.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
     "chart.js": "^2.3.0",
     "clear-cut": "^2.0.2",
     "coffee-script": "1.11.1",
-    "color": "^0.7.3",
     "dedent": "^0.6.0",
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I think it is used [in color.js](https://github.com/atom/atom/blob/8ca558f490b3316e9e110feefa63e0dfe60053b1/src/color.js#L27), for parsing color values. Am I missing something?

50Wliu:
Hmm.  Looks like I missed that when ctrl-f'ing.  Also looks like we don't have spec coverage for that 😬.

maxbrunsfeld:
I would think [this test](https://github.com/atom/atom/blob/c0973a0b601e5044c0e3dcffdb120c01df388146/spec/config-spec.coffee#L1640) would exercise that code path. I wonder what's going on there?

PR #14435, opened 2017-05-17
Use new native text-buffer implementation

### LINKED ISSUE #7210: Uncaught RangeError: Invalid string length
To reproduce:
1. `dd if=/dev/zero of=largefile count=1 bs=1000M`; if this does not reproduce play around with adjusting `bs` and `count`.
2. Open atom and try to open `largefile`.

**Atom Version**: 0.208.0
**System**: Ubuntu 15.04
**Thrown From**: Atom Core
### Stack Trace

Uncaught RangeError: Invalid string length

```
At /usr/share/atom/resources/app.asar/node_modules/pathwatcher/lib/file.js:301

RangeError: Invalid string length
    at Array.join (native)
    at ReadStream.<anonymous> (/usr/share/atom/resources/app.asar/node_modules/pathwatcher/lib/file.js:301:43)
    at emitNone (events.js:72:20)
    at ReadStream.emit (events.js:163:7)
    at _stream_readable.js:891:16
    at process._tickCallback (node.js:357:13)
```
### Commands

```
  2x -3:24.1.0 release-notes:show (atom-text-editor.editor.is-focused)
```
### Config

``` json
{
  "core": {},
  "editor": {
    "invisibles": {},
    "tabLength": 4
  }
}
```
### Installed Packages

``` coffee
# User
No installed packages

# Dev
No dev packages
```

### LINKED ISSUE #11012: Atom hangs when an opened file is written to
### Prerequisites
- [ X ] Can you reproduce the problem in [safe mode](https://atom.io/docs/latest/hacking-atom-debugging#check-if-the-problem-shows-up-in-safe-mode)?
- [ X ] Are you running the [latest version of Atom](https://atom.io/docs/latest/hacking-atom-debugging#update-to-the-latest-version)?
- [ X ] Did you check the [debugging guide](https://atom.io/docs/latest/hacking-atom-debugging)?
- [ X ] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [ X ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [ X ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

### Description

Atom beta hangs when an open file is clobbered and continuously written to
### Steps to Reproduce

```
touch out.txt
atom-beta --safe out.txt
(for i in $(seq 100000); do echo "yyyyyyyyyy"; done) > out.txt
```

**Expected behavior:** [What you expected to happen]

atom changes to new file, and continues to work (atom stable does handle this fine)

**Actual behavior:** [What actually happened]

The two 'atom helper' processes hit > 260% CPU, and stay there. Eventually I am asked if I want to "keep waiting"
### Versions

You can get this information from executing `atom --version` and `apm --version` at the command line.

```
bash-4.3$ atom-beta --version
a1.6.0-beta6
bash-4.3$ apm-beta --version
apm  1.6.0
npm  2.13.3
node 0.10.40
python 2.7.11
git 2.7.2
```

### LINKED ISSUE #11554: Uncaught Error: toString failed
1. ...
2. ...

**Atom Version**: 1.7.2
**System**: Mac OS X 10.11.4
**Thrown From**: Atom Core
### Stack Trace

Uncaught Error: toString failed

```
At buffer.js:378

Error: toString failed
    at Buffer.toString (buffer.js:378:11)
    at Object.fs.readFileSync (fs.js:496:33)
    at Object.fs.readFileSync (ATOM_SHELL_ASAR.js:427:29)
    at File.module.exports.File.readSync (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/node_modules/pathwatcher/lib/file.js:266:36)
    at File.module.exports.File.getDigestSync (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/node_modules/pathwatcher/lib/file.js:167:14)
    at TextBuffer.module.exports.TextBuffer.serialize (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:231:68)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/project.js:137:27
    at Array.map (native)
    at Project.module.exports.Project.serialize (/Applications/Atom.app/Contents/Resources/app.asar/src/project.js:135:41)
    at AtomEnvironment.module.exports.AtomEnvironment.serialize (/Applications/Atom.app/Contents/Resources/app.asar/src/atom-environment.js:769:31)
    at /Applications/Atom.app/Contents/Resources/app.asar/src/atom-environment.js:984:27
```
### Commands

```
```
### Config

``` json
{}
```
### Installed Packages

``` coffee
# User
No installed packages

# Dev
No dev packages
```

maxbrunsfeld:
### Summary

This pull request completely reimplements the `TextBuffer` class - Atom's core data structure representing editable text. The new implementation differs from the old one in several major ways:

* The class is now implemented primarily in C++, as part of our native library [superstring](https://github.com/atom/superstring).
* Instead of storing the text as a mutable array of lines, we store it as flat immutable string representing the file's original content, plus a *Patch* representing the user's unsaved changes. This is somewhat similar to the [piece table](https://en.wikipedia.org/wiki/Piece_table) data structure used in some other editors.
* At the native level, the buffer supports *snapshots* - read-only views of the buffer that can be processed in background threads with no copying or locking, even while the buffer is being mutated on the main thread.
* File I/O and transcoding is now handled by the native library, entirely on background threads and exposed via promises.

### Motivation

For many purposes, the existing array-of-lines implementation was actually very fast; we've never observed slowness associated with the querying or mutating of the text itself. This is a testament to V8's highly efficient String operations. However, there are many operations in Atom that *use* the `TextBuffer` and would benefit from being dropped to C++ and/or being moved to a background thread.

Here are some examples of this that have already been addressed in this PR:

* Previously, saving and loading a `TextBuffer` required its entire contents to be represented as a single JavaScript String. V8 has a hard limit on string length, so this would cause errors (#7210). Now, saving and loading are done in a background thread in C++, with minimal copying of data.
* When the file associated with an unmodified `TextBuffer` changes on disk, we diff the file's new contents with the buffer's contents in order to update the buffer in a way that preserves marker positions. Previously, this diff computation caused unresponsiveness in large files (atom/text-buffer#153). Now the diff is performed in C++ and on a background thread.
* Whenever the application is idle, we serialize the state of the window and persist it, so that we can restore it even if Atom crashes or power is lost. Previously, persisting a large file's text would cause hangs, so we had to disable this safety feature for large files (#13583). Now, because of our new representation of the buffer, we are able to efficiently serialize *just* the outstanding changes, so we can re-enable the feature for arbitrarily large files.
* The [line-ending-selector](https://github.com/atom/line-ending-selector) needs to detect what type of line endings a `TextBuffer` uses. Because we support files with *mixed* line endings, this entails scanning the entire buffer. Previously, this would take upwards of 50ms for a large file. Now, the search is performed asynchronously so that it causes no noticeable lag.

In addition to these improvements, this PR will make it very easy to optimize other features. Here are some follow-up PRs that I have in mind:
* When changing a large amount of text at once, indexing the new text in the `DisplayLayer` takes a long time. This is one of the main reasons that pasting a huge amount of text is slow. Some of this character-by-character computation could probably be done much faster in C++.
* Currently, using the [find-and-replace](https://github.com/atom/find-and-replace) package in a very large file can cause hangs when searches produce more than a hundred thousand results. We will now be able to run searches entirely in a background thread, eliminating any lag.
* Populating the [autocomplete-plus](https://github.com/atom/autocomplete-plus) package's `SymbolStore` can take a very long time for large files, so we had to [disable it](https://github.com/atom/autocomplete-plus/pull/782) for large files. Fuzzy matching should be able to be performed in C++ on a background thread, eliminating this restriction.

### Implementation details
<details>
<summary>Our new representation of the buffer:</summary>
<br/>

Suppose I create a `TextBuffer` with some initial text and then make three edits:

```js
const buffer = new TextBuffer('abc\ndefg\nhijkl')
buffer.setTextInRange([[0, 2], [0, 3]], "C")
buffer.setTextInRange([[2, 0], [2, 1]], "H")
buffer.setTextInRange([[1, 1], [1, 2]], "E")
assert.equal(buffer.getText(), 'abC\ndEfg\nHijkl')
```

This buffer is now represented internally like this:

graph-1

The buffer is composed of two *layers*. The first layer stores the buffer's initial text as a flat, immutable string. This makes freshly-opened files very compact in memory. The second layer stores the unsaved edits in an object that we call a `Patch`. The `Patch` is structured as a [splay tree](https://en.wikipedia.org/wiki/Splay_tree) so that mutations and queries are very efficient, especially in areas that have recently been accessed.

Then, suppose I begin some asynchronous operation like saving the buffer to disk, or searching 
the buffer for a regex. Then, *while* that operation is still happening, I make some *additional* edits:

```js
const savePromise = buffer.save()
buffer.setTextInRange([[0, 0], [0, 1]], 'A')
buffer.setTextInRange([[2, 4], [2, 5]], 'L')
assert.equal(buffer.getText(), 'AbC\ndEfg\nHijkL')
```

At that point, the buffer's internal state would look like this:

graph2

There are now *three* layers. Because the asynchronous `save` is still in progress, further changes that we make to the buffer are saved in a *new* Patch. When the `save` completes, we flatten the text down into one layer again.

</details>

### Effect on packages

I don't believe making `save` async has affected any of our bundled packages' functionality, though it has caused some test failures that I've had to fix:
* [x] https://github.com/atom/autocomplete-plus/pull/852
* [x] https://github.com/atom/whitespace/pull/155
* [x] https://github.com/atom/line-ending-selector/pull/45
* [x] https://github.com/atom/symbols-view/pull/222
* [x] https://github.com/atom/status-bar/pull/193
* [x] https://github.com/atom/tabs/pull/448

Depends on https://github.com/atom/text-buffer/pull/225
Depends on https://github.com/atom/superstring/pull/5
Depends on https://github.com/atom/node-pathwatcher/pull/121
Depends on https://github.com/atom/whitespace/pull/157

Closes https://github.com/atom/atom/issues/7210
Closes https://github.com/atom/atom/issues/11554
Closes https://github.com/atom/atom/issues/11012


maxbrunsfeld:
@atom/maintainers I think this PR is ready for others to test.

@ungb @Ben3eeE Some areas that could use testing:

Loading and saving buffers. This includes actual I/O as well as encoding/decoding. Areas in which I could imagine regressions:
  * Opening a file on a slow drive (e.g. `sshfs` and other network file systems)
  * Opening files in non-UTF8 encodings
  * Changing files on disk while they are open in Atom
  * Handling errors when reading/writing files (e.g. permissions issues)
  * Lots of simultaneous IO operations (e.g. multiple saves at the same time, multiple on-disk changes in rapid succession)

Working with buffers with large numbers of unsaved edits. All the operations that I've tested perform as well or better than before, but because of the new buffer representation, I could imagine some operations getting slower. We need to make sure that doesn't happen. Some types of operations I'm thinking of:
  * Performing a regex replace with lots of search results
  * Indenting large chunks of the buffer
  * Pasting large amounts of text
  * Normal editing *after* doing any of the above

32-bit platforms. I did some testing on a 32-bit linux VM, but possibly not enough.

maxbrunsfeld:
Download links for convenience:

* [atom-mac.zip](https://4010-3228505-gh.circle-artifacts.com/0/Users/distiller/atom/out/atom-mac.zip) from Circle
* [atom-windows.zip (x64)](https://ci.appveyor.com/api/buildjobs/82qxfouodneo0joq/artifacts/out%2Fatom-windows.zip) from Appveyor

ungb:
Few things I noticed while testing this:

-  when you open a decently large file ~70mb and then save it, the heap increases by 10x and never gets freed.  Over time, this will eventually cause atom to crash. 
image
This doesn't repro in beta, talking to @maxbrunsfeld seems to do with caching that was occurring? 
Also, this doesn't seem to occur on smaller files.  Will investigate when this occurs

- Saving also seems to take a bit longer

- File sometimes blanks out.

maxbrunsfeld:
Ok, we've found and addressed a number of bugs in the last two days. Download links for the latest version:

* [mac](https://4070-3228505-gh.circle-artifacts.com/0/Users/distiller/atom/out/atom-mac.zip)
* [windows x64](https://ci.appveyor.com/api/buildjobs/5ifu2mgmwgjp5gdg/artifacts/out%2Fatom-windows.zip)

Glavin001:
I notice you mentioned "making save async". Will there be any new Atom APIs allowing package developers to tell Atom to wait for packages to perform async operations to TextBuffer before saving? See https://discuss.atom.io/t/best-way-to-change-the-texteditor-textbuffer-on-save-when-my-function-is-asynchronous/43531 for more context to my question. Thank you!

lee-dohm:
@Glavin001 The `TextBuffer.onWillSave` event will continue to work the way it always has. I've tested it out using my [tabs-to-spaces package](https://atom.io/packages/tabs-to-spaces) on a [very large file](https://github.com/lee-dohm/tabs-to-spaces/blob/master/sample/jquery-git2.js.txt).

Glavin001:
>  The TextBuffer.onWillSave event will continue to work the way it always has.

Therefore this PR may not solve my issue [as suggested](https://discuss.atom.io/t/best-way-to-change-the-texteditor-textbuffer-on-save-when-my-function-is-asynchronous/43531/3?u=glavin001). Hmm..

I'll elaborate, and hopefully you have some ideas, @lee-dohm. My problem has been `TextBuffer.onWillSave` does not wait for my asynchronous operations to complete before actually saving the buffer.

My situation with [Atom-Beautify](https://github.com/Glavin001/atom-beautify) is similar to the following:

```
const buffer = editor.getBuffer();
buffer.onWillSave(() => {
    // Please wait for async function to finish
    setTimeout(() => {
       // Nope! Buffer has already been saved!
       buffer.setText('Async change completed!');
    }, 1000);
})
```

Looking at the [`TextBuffer.onWillSave` implementation](https://github.com/atom/text-buffer/blob/v11.4.1/src/text-buffer.coffee#L347-L348) I think this only works when the operation is synchronous. *I would like to be able to return a promise (or callback) to let Atom know I have completed my changes and Atom can now save.* If this feature is also desirable to the Atom team, I'd be happy to submit a Pull Request, if I could have a little guidance 😄 .

maxbrunsfeld:
> My problem has been TextBuffer.onWillSave does not wait for my asynchronous operations to complete before actually saving the buffer.

@Glavin001 Thanks for the explanation. This problem comes up a lot. I don't think that I'll solve it in this particular PR, but it will become a lot *easier* to solve after this PR, because I've adjusted all of Atom's codebase to expect `save` be asynchronous instead of synchronous. So now it will be a pretty small change to allow `onWillSave` callbacks to delay the save operation.

joefitzgerald:
@Glavin001 I also have a need for the `onWillSave` event to support asynchronous functions (likely returning a promise).

I want to run `gofmt` or `goimports` for people writing Go code. Today this is synchronous, which leads to a sub par user experience as an external process in invoked.

In general, code formatters need to run at this time; but in a way that protects the user in the event of some failure.

maxbrunsfeld:
@ungb I've tracked down the problem where you found where a large buffer was 'blanked out' after its file was written to externally. I'll have a fix shortly.

Glavin001:
Thank you for your feedback, @maxbrunsfeld ! This is great to hear. I would love to contribute and bring this next change into reality, if someone else is not already planning to. How can I help? 😃

ungb:
found https://github.com/atom/atom/issues/14722 while testing this branch.  Will pull from latest and try again.  Was able to repro on windows and mac.

src/pane.coffee
```diff
@@ -603,55 +605,71 @@ class Pane
...
+            @removeItem(item, false)
+            item.destroy?()
+          result
 
   # Public: Destroy all items.
   destroyItems: ->
-    @destroyItem(item) for item in @getItems()
-    return
+    Promise.all(
+      @getItems().map(@destroyItem.bind(this))
```
t9md: @maxbrunsfeld 
Noticed while I read code, using `bind` here is bad.
Since `Array.prototype.map` pass item index as 2nd arg, which goes to `@destroyItems` which treat 2nd arg as `force` switch. Which is not intentional I believe.

src/pane.coffee
```diff
@@ -603,55 +605,71 @@ class Pane
...
+    )
 
   # Public: Destroy all items except for the active item.
   destroyInactiveItems: ->
-    @destroyItem(item) for item in @getItems() when item isnt @activeItem
-    return
+    Promise.all(
+      @getItems()
+        .filter((item) => item isnt @activeItem)
+        .map(@destroyItem.bind(this))
```
t9md: here too.

src/pane.coffee
maxbrunsfeld: Wow, thanks for catching that @t9md! Are you interested in opening a PR fixing that bug?

t9md: Yes, I will try tommorrow.

maxbrunsfeld:
@Glavin001 @joefitzgerald The `TextBuffer.onWillSave` hook now allows you do perform async work before the buffer is saved. You just need to return a promise from the event handler. For example:

```js
buffer.onWillSave(async () => {
  await somethingSlow()
})
```

See https://github.com/atom/text-buffer/pull/254.

This change has been released in Atom v1.21.0-beta1.

Glavin001:
Awesome! Thank you, @maxbrunsfeld!

PR #14225, opened 2017-04-16
Add `rpm --rebuilddb` to prevent 'Rpmdb checksum' error

TETRA2000:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

I fixed Dockerfile to rebuild RPM DB to prevent checksum error described in #14224 .

### Alternate Designs

* Change the base docker image.
* Force to ignore checksum errors.

### Why Should This Be In Core?

This Dockerfile is used to test core with Copr. (#10945)

### Benefits

CIs and Developers can build atom using Docker.

### Possible Drawbacks

(If present) Current CI settings which depends on this Dockerfile may be affected.

### Applicable Issues

#14224

PR #14444, opened 2017-05-17
Update docker image and add docs

### LINKED ISSUE #14224: Failed to build docker image with Dockerfile.
### Prerequisites

### Description

1. I tried to build  docker image with this command.
```
docker build -t atom .
```

2. Then I got an error like this while executing dependency installation.
```
  Cleanup    : rpm-libs-4.12.0.1-3.fc21.x86_64                            37/40
  Cleanup    : elfutils-libs-0.160-1.fc21.x86_64                          38/40
  Cleanup    : elfutils-libelf-0.160-1.fc21.x86_64                        39/40
  Cleanup    : glib2-2.42.1-1.fc21.x86_64                                 40/40

Rpmdb checksum is invalid: dCDPT(pkg checksums): elfutils-libs.x86_64 0:0.163-4.fc21 - u

The command '/bin/sh -c yum install -y     make     gcc     gcc-c++     glibc-devel     git-core     libgnome-keyring-devel     rpmdevtools' returned a non-zero code: 1
```

### Steps to Reproduce

Same as description.

**Expected behavior:** Build will success.

**Actual behavior:** Build failed with an error.

**Reproduces how often:** Always when I've tried.

### Versions

Current master (d07a26882f11b1a10003ca673359ec73977f7734)

### Additional Information

Build succeeded when I modified Dockerfile as below.

```
# RUN yum install -y \
RUN rpm --rebuilddb && yum install -y \
    make \
    gcc \
```

### LINKED ISSUE #14394: v1.16.0/Dockerfile script/build error
https://github.com/atom/atom/blob/v1.16.0/Dockerfile
```bash
git checkout v1.16.0
docker build -t atom .
docker run -i -t atom bash
script/build --create-debian-package
...
make: Entering directory '/atom/node_modules/keyboard-layout/build'
  CXX(target) Release/obj.target/keyboard-layout-manager/src/keyboard-layout-manager-linux.o
keyboard-layout-manager.target.mk:96: recipe for target 'Release/obj.target/keyboard-layout-manager/src/keyboard-layout-manager-linux.o' failed
make: Leaving directory '/atom/node_modules/keyboard-layout/build'

...

In file included from ../src/keyboard-layout-manager-linux.cc:1:0:
../src/keyboard-layout-manager.h:7:22: fatal error: X11/Xlib.h: No such file or directory
 #include <X11/Xlib.h>
                      ^
compilation terminated.
make: *** [Release/obj.target/keyboard-layout-manager/src/keyboard-layout-manager-linux.o] Error 1
```

hultberg:
### Requirements

### Description of the Change

This PR changes the Dockerfile base image to node:6 which is the current LTS version of node. Following the build instructions in the documentation for linux, the image will be built with alle the requirements for building atom. At the same time, I added a little documentation for how to build and use the image.

Difference from the original Dockerfile image is the ADD and WORKDIR instructions in it, I decided to remove this and instead instruct to mount atom in the docker run command as the filesystem can change. 

### Alternate Designs

1. One might consider adding ADD and WORKDIR instructions back to Dockerfile.
2. #14225 only fixes the current failures with the current docker setup, and can be considered an alternative to this PR.

Because it expands on features in core.

### Benefits

Docker makes it easy to build atom on any system that have Docker installed, without the need to handle dependencies for atom itself.

### Possible Drawbacks

I'm not sure, as I do not know to what extent the Dockerfile is used in core or what purpose it had originally.

### Applicable Issues

Fixes #14394 
Fixes #14224
Closes #14225

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,15 @@
-# VERSION:        0.1
-# DESCRIPTION:    Image to build Atom and create a .rpm file
+# VERSION:        0.2
+# DESCRIPTION:    Image to build Atom
 
 # Base docker image
-FROM nodesource/fedora21:4.2.6
+FROM node:6
 
 # Install dependencies
-RUN yum install -y \
-    make \
-    gcc \
-    gcc-c++ \
-    glibc-devel \
-    git-core \
-    libsecret-devel \
-    rpmdevtools
-
-RUN npm install -g npm --loglevel error
-
-ADD . /atom
-WORKDIR /atom
+RUN apt-get update && \ 
+	apt-get install -y build-essential \
+		git \
+		libsecret-1-dev \
+		fakeroot \
+		rpm \
+		libx11-dev \
+		libxkbfile-dev
--- a/docs/build-instructions/docker.md
+++ b/docs/build-instructions/docker.md
@@ -1,0 +1,25 @@
+# Docker build instructions
+
+You can build atom using the Dockerfile provided in the repository.
+
+## Requirements
+
+* Docker installed
+
+## Building the docker image
+
+Before building, the docker image must be built. The image is based on node:6 and installs the dependencies required for atom.
+
+```
+$ ~/atom: docker build -t atom:latest ./
+```
+
+This will build the image. It will be available later as `atom:latest`
+
+## Building atom
+
+Run the following command to run the atom build container with the image we just built. Note that you can provide arguments to `script/build` at the and as you wish.
+
+```
+$ ~/atom: docker run --rm --name atom-build --volume $(pwd):/atom --workdir /atom atom:latest script/build
+```


andyrichardson:
Nice work! One possible issue though - the dependency for building is Ubuntu (in this case) not Node. 

Using a Node image might cause issues if the maintainer of the Node image decides to change the base OS. I'd suggest it would be better practice to use an Ubuntu/Debian image and install Node and build dependencies from there.

hultberg:
@andyrichardson I see your concern, and I'm not familiar with your claim that Ubuntu is the dependency - in this case. Can you elaborate further?

When I updated the image I followed the official build doc for Linux where it describes ubuntu as being the recommended platform, and that an OS with 64-bit or 32-bit arch is enough on linux.

andyrichardson:
The way I see it, in this case, we're trying to make a container which can build and package Atom. While Node is required for this function, it is required alongside other packages. 

If you think of it as a dependency tree, we could have Ubuntu at the core (base image) and then rpm, dpkg and Node branching off from this. When you use Node as the base image it could imply that it's the core dependency - that isn't the case.

It doesn't make a huge difference and, in fact, using the Node image will be easier and faster when first running the container. I just feel it makes more sense to have the base image reflecting the environment that we intend to use.

sadick254:
Hey @hultberg I know this PR has been open for a long time. Apologies we didn't get time to look into it. We now have support for building Atom using a docker container. https://github.com/atom/atom/blob/a3d4f9cdbd9b8f625df5e639d37e0863a5265126/Dockerfile#L1
Since building with docker is already supported, I have no choice but to close this PR.

Thank you again for your contribution.

PR #14445, opened 2017-05-17
Install dev dependencies when transpiling

kuychaco:
This is to allow us to move babel and other dependencies used only during transpiling to dev dependencies so that we can reduce the size of our build artifacts. 

Ref https://github.com/atom/github/issues/638

### DIFF
--- a/script/lib/transpile-packages-with-custom-transpiler-paths.js
+++ b/script/lib/transpile-packages-with-custom-transpiler-paths.js
@@ -6,6 +6,7 @@ const glob = require('glob')
 const path = require('path')
 
 const CONFIG = require('../config')
+const backupNodeModules = require('./backup-node-modules')
 const runApmInstall = require('./run-apm-install')
 
 require('colors')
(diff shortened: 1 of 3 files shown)


script/lib/transpile-packages-with-custom-transpiler-paths.js
```diff
@@ -1,29 +1,36 @@
...
-const fs = require('fs')
+const fs = require('fs-extra')
 const glob = require('glob')
 const path = require('path')
 
 const CONFIG = require('../config')
+const backupNodeModules = require('./backup-node-modules')
+const runApmInstall = require('./run-apm-install')
+
+require('colors')
```
steelbrain: Fly-by comment, maybe use [`chalk`](https://www.npmjs.com/package/chalk) in the future instead of using a module that modifies prototypes (because modules modifying prototypes are prone to conflicts and generally unrecommended)

Arcanemagus: Although I agree with you in principle, `color` is already used elsewhere in the Atom repo, while `chalk` would have involved bringing in a new dependency.

Arcanemagus:
For anyone reading this PR, since it wasn't clear (to me at least) from the description this _only_ comes into play when bundling the package within Atom itself, not for third party package using the custom transpilation (ref: https://github.com/atom/github/pull/840#issuecomment-355613552).

Issue #14231, opened 2017-04-17
Uncaught TypeError: el.getAttribute is not a function


yacut:
Steps to reproduce:

1. Create a dock pane
2. Select some element on the pane
3. Drag the selected element

**Atom**: 1.17.0-beta2 x64
**Electron**: 1.3.14
**OS**: elementary OS
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: el.getAttribute is not a function

```
At /usr/share/atom-beta/resources/app/src/workspace-element.js:283

TypeError: el.getAttribute is not a function
  at isTab (/usr/share/atom-beta/resources/app/src/workspace-element.js:283:12)
  at atom-workspace.handleDragStart (/usr/share/atom-beta/resources/app/src/workspace-element.js:141:10)

```

### Commands

```
     -1:13.2.0 tree-view:show (atom-workspace.workspace.scrollbars-visible-always)
```

### Non-Core Packages

```
activate-power-mode 1.4.0 
active-editor-info 0.0.0 
atom-beautify 0.29.22 
atom-fuzzy-grep 0.16.0 
atom-jasmine 0.8.1 
atom-ternjs 0.18.2 
atom-typescript 11.0.1 
autocomplete-modules 1.6.9 
autocomplete-paths 1.0.5 
clipboard-plus 0.5.1 
compare-files 0.8.0 
cson 2.0.1 
cucumber 0.6.0 
docblockr 0.9.4 
docker 0.7.0 
docs-snippets 0.8.0 
editorconfig 2.2.2 
escape-utils 0.5.3 
file-icons 2.1.3 
from-and-to-json 1.0.0 
git-blame 1.0.1 
git-plus 7.5.0 
git-split-diff-hyperclick 0.3.12 
git-time-machine 1.5.9 
highlight-selected 0.13.1 
hyperclick 0.0.40 
js-hyperclick 1.10.1 
language-babel 2.57.3 
language-docker 1.1.7 
language-ini 1.19.0 
language-thrift 1.0.2 
linter 2.1.4 
linter-csslint 1.3.4 
linter-docker 0.2.0 
linter-eslint 8.1.6 
linter-flow 5.5.1 
linter-jsonlint 1.3.0 
linter-tslint 1.0.0 
merge-conflicts 1.4.4 
minimap 4.27.1 
minimap-find-and-replace 4.5.2 
minimap-git-diff 4.3.1 
minimap-highlight-selected 4.6.1 
nuclide 0.217.0 
[...]

PR #14485, opened 2017-05-18
Fix exception when starting to drag text

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/14231


maxbrunsfeld:
The CI failure is unrelated to the change.

PR #14495, opened 2017-05-19
Enable MAX_SCREEN_LINE_LENGTH to now be set via a config option

warrenpnz:
### Description of the Change

This change enables the maximum softWrapColumn in the display layer object to be set using a configuration parameter rather than using the MAX_SCREEN_LINE_LENGTH constant. This value is returned by the getSoftWrapColumn function.

The default and minimum is still set at 500 to satisfy the design decision to address performance issues for those opening minified files, however it can be adjusted upwards for those who require a more flexible option.

### Alternate Designs

As the getSoftWrapColumn function used a constant as the default return value, there are no alternate design options.

This needs to be in core as it is affecting the editor display layer parameters directly 

### Benefits

Benefits are that the number of people who work with long line files that are not necessarily minified are able to customise their maximum width by the settings editor instead of using the constant value

### Possible Drawbacks

Possible drawbacks are that a user sets the width high enough to cause performance issues with minified files, however this can be mitigated by reverting the setting to defaults in the editor settings

### Applicable Issues

* https://discuss.atom.io/t/atom-editor-line-length-wraps-at-500-chars-on-1-15/40589/3
* https://discuss.atom.io/t/1-14-1-16-is-forcing-a-line-wrap-even-with-soft-wrap-toggled-off/41758/4
* https://discuss.atom.io/t/how-do-i-turn-off-soft-wrap/40417/11

Related to: https://github.com/atom/atom/pull/13820

### DIFF
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -544,6 +544,21 @@ describe('TextEditorRegistry', function () {
       expect(editor.getSoftWrapColumn()).toBe(80)
     })
 
+    it('allows for custom definition of maximum soft wrap based on config', async function () {
+      editor.update({
+        softWrapped: false,
+        maxScreenLineLength: 1500,
+      })
+
+      expect(editor.getSoftWrapColumn()).toBe(1500)
+
+      atom.config.set('editor.softWrap', false)
+      atom.config.set('editor.maxScreenLineLength', 500)
+      registry.maintainConfig(editor)
+      await initialPackageActivation
+      expect(editor.getSoftWrapColumn()).toBe(500)
+    })
+
     it('sets the preferred line length based on the config', async function () {
       editor.update({preferredLineLength: 80})
       expect(editor.getPreferredLineLength()).toBe(80)
--- a/src/text-editor.coffee
+++ b/src/text-editor.coffee
@@ -17,7 +17,6 @@ TextEditorElement = null
 {isDoubleWidthCharacter, isHalfWidthCharacter, isKoreanCharacter, isWrapBoundary} = require './text-utils'
 
 ZERO_WIDTH_NBSP = '\ufeff'
-MAX_SCREEN_LINE_LENGTH = 500
 
 # Essential: This class represents all essential editing state for a single
 # {TextBuffer}, including cursor and selection positions, folds, and soft wraps.
@@ -158,7 +157,7 @@ class TextEditor extends Model
       @assert, grammar, @showInvisibles, @autoHeight, @autoWidth, @scrollPastEnd, @scrollSensitivity, @editorWidthInChars,
       @tokenizedBuffer, @displayLayer, @invisibles, @showIndentGuide,
       @softWrapped, @softWrapAtPreferredLineLength, @preferredLineLength,
-      @showCursorOnSelection
+      @showCursorOnSelection, @maxScreenLineLength
     } = params
 
     @assert ?= (condition) -> condition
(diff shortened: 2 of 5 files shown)


maxbrunsfeld:
:+1: Thanks for doing this; this is a good feature. We will merge it if you add a test for it.

I think there should be two simple tests:
* one in `text-editor-spec.coffee` that calls `TextEditor.update` with the `maxScreenLineLength` parameter.
* one in `text-editor-registry-spec` that checks that the config setting works correctly.

lee-dohm:
Not something that needs to be addressed in this PR ... but Safe Mode should probably force this configuration value to the default.

warrenpnz:
So I'm having a bit of trouble running local spec tests, but these should work.
How do others do local testing without pushing to github? (OSX or CentOS)

50Wliu:
`atom --test spec`.  You can focus specs by adding an `f` in front of them, eg `fit` or `fdescribe`.  Just don't forget to unfocus them before committing.

warrenpnz:
Finally got local spec testing to work and now tests ok.

src/text-editor.coffee
```diff
@@ -285,6 +285,10 @@ class TextEditor extends Model
             @preferredLineLength = value
             displayLayerParams.softWrapColumn = @getSoftWrapColumn()
 
+        when 'maxScreenLineLength'
+          if value isnt @maxScreenLineLength
+            @maxScreenLineLength = value
```
maxbrunsfeld: Since the `maxScreenLineLength` field affects value returned by `getSoftWrapColumn`, I think we should include the following statement here:

```coffee
displayLayerParams.softWrapColumn = @getSoftWrapColumn()
```

This way, if you *just* update the `maxScreenLineLength` field, we'll still recompute the text layout in the display layer.

spec/text-editor-registry-spec.js
```diff
@@ -544,6 +544,23 @@ describe('TextEditorRegistry', function () {
...
+
+      expect(editor.getSoftWrapColumn()).toBe(1500)
+
+      atom.config.set('editor.softWrap', false)
+      atom.config.set('editor.maxScreenLineLength', 500)
+      registry.maintainConfig(editor)
+      await initialPackageActivation
+      expect(editor.getSoftWrapColumn()).toBe(500)
+
+    })
```
maxbrunsfeld: Could you remove this blank line before the `}`?

maxbrunsfeld (changes requested):
I left two small changes. Other than that, this looks great. Thanks!

src/text-editor.coffee
warrenpnz: Fixed in latest commit. Thanks for the suggestions.

maxbrunsfeld (approved):

battaglr:
Is this still being considered?

iolsen:
@warrenpnz if you can get the builds green and resolve conflicts we'll get this merged.

wpowell-oss:
Not sure what thee errors are as one looks like AppVeyor failed to install a module during the build that was not included in the build, and the CircleCI one is an internal eror for code signing. Any help on figuring these out would be appreiated as I can't see how they relate to the code updates

50Wliu:
A rebase is needed so that Appveyor can rebuild the PR.

wpowell-oss:
Rebase complete.

battaglr:
Awesome! 👏 👏 👏 👏

warrenpnz:
Thank you all for the help in getting this one through.

PR #14640, opened 2017-05-25
Teach directional pane navigation commands to support docks

jasonrudolph:
#### Functionality Description

Prior to this change, the following commands successfully operated on panes in the workspace center, but they could not focus/move/copy between the the panes in the workspace center and panes in the docks:

- `window:focus-pane-above`
- `window:focus-pane-below`
- `window:focus-pane-on-left`
- `window:focus-pane-on-right`
- `window:move-active-item-to-pane-above`
- `window:move-active-item-to-pane-below`
- `window:move-active-item-to-pane-on-left`
- `window:move-active-item-to-pane-on-right`
- `window:copy-active-item-to-pane-above`
- `window:copy-active-item-to-pane-below`
- `window:copy-active-item-to-pane-on-left`
- `window:copy-active-item-to-pane-on-right`

This pull request updates these commands to work across all visible panes, regardless of whether the pane is in the workspace center or in a dock.

#### Implementation Summary

- Prior to this change, the [directional pane focus/move/copy logic resided in `PaneContainerElement`](https://github.com/atom/atom/blob/7cfcdadbfbc2169b99893107f43be0fb3c2bbd30/src/pane-container-element.coffee#L27-L94). By hosting the logic in `PaneContainerElement`, we could only focus/move/copy between panes in a specific pane container (e.g., a single pane container in the workspace center). To make it possible to focus/move/copy between _all_ panes in the workspace, relocate the logic to `WorkspaceElement`.
- Add [tests for `nearestVisiblePaneInDirection`](https://github.com/atom/atom/blob/e3c612b261d658f492d7456dc9c18ccb48f23aa6/spec/workspace-element-spec.js#L37-L232), which provides the core logic for the higher-level functions like `focusPaneViewAbove`, `moveActiveItemToPaneAbove`, `focusPaneViewOnLeft`, etc.
    - Test the generic logic extensively (i.e., the logic that is independent of whether the given pane resides in the workspace center or a dock)
    - *Also* test the navigation between docks and the workspace center
- Since the core directional logic is tested in the new tests above, simplify the [tests for the higher-level functions](https://github.com/atom/atom/blob/e3c612b261d658f492d7456dc9c18ccb48f23aa6/spec/workspace-element-spec.js#L234-L483) (e.g., `focusPaneViewAbove`, `moveActiveItemToPaneAbove`, `focusPaneViewOnLeft`) to avoid unnecessary duplication.

### Alternate Designs

- We could perhaps leave the existing functions in `PaneContainerElement` and update them to focus/move/copy across pane containers, but it would seem odd (to me) to have something in `PaneContainerElement` operate on elements that exist in _other_ pane containers.
- We could leave the existing functionality in place and _add_ new commands that provide the ability to focus/move/copy across pane containers.
    - This would result in more code to maintain.
    - It would require us to come up with command names that distinguish the old behavior from the new behavior (e.g. `window:focus-pane-above` and `window:seriously-focus-pane-above` :stuck_out_tongue_winking_eye:)
    - I can't think of a compelling reason to the limit pane navigation commands to a single container. If you know of real world value that developers get from only being able to focus/move/copy between panes in a single pane container, please let me know. :pray:

The existing commands reside in core, and this change updates them to support the docks that were recently added to core.

### Benefits

Developers gain the ability to focus/move/copy across *all* visible panes. Developers are no longer limited to focusing/moving/copying between panes in the workspace center.

### Possible Drawbacks

`PaneContainerElement` is not part of Atom's public API, but it's possible that some packages are relying on functions in `PaneContainerElement`. This pull request removes several functions from `PaneContainerElement`. If any packages are relying on those non-public functions, those packages will encounter errors when attempting to invoke those functions. Hopefully package authors would discover those issues during the next beta release, so that they could adjust their packages accordingly.

### Applicable Issues

https://github.com/atom/atom/pull/1484 introduced the ability to navigate directionally between panes.

### Demo

demo

### Proposed Test Plan

I propose using the steps below to verify that this PR is providing the desired behavior. For now, I've done a sanity check in dev mode, and things appear to be working as expected. Once folks have had a chance to review this pull request, if folks are :+1: on the overall approach, I'll run through all of these steps to verify the behavior:

- [x] `window:focus-pane-above`
  - [x] You can navigate up from a workspace center pane to the adjacent workspace center pane above
  - [x] You can navigate up from the bottom dock to the bottommost pane in the workspace center
  - [x] When there is no pane above the focused pane, the focused pane remains focused
- [x] `window:focus-pane-below`
  - [x] You can navigate down from a workspace center pane to the adjacent workspace center pane below
  - [x] When the bottom dock is visible, you can navigate down from the bottommost pane in the workspace center to the bottom dock
  - [x] When the bottom dock is hidden and the bottommost pane in the workspace center is focused, the focused pane remains focused
  - [x] When there is no pane below the focused pane, the focused pane remains focused
- [x] `window:focus-pane-on-left`
  - [x] You can navigate left from a workspace center pane to the adjacent workspace center pane on the left
  - [x] When the left dock is visible, you can navigate left from the leftmost pane in the workspace center to the left dock
  - [x] When the left dock is hidden and the leftmost pane in the workspace center is focused, the focused pane remains focused
  - [x] When there is no pane to the left of the focused pane, the focused pane remains focused
- [x] `window:focus-pane-on-right`
  - [x] You can navigate right from a workspace center pane to the adjacent workspace center pane on the right
  - [x] When the right dock is visible, you can navigate right from the rightmost pane in the workspace center to the right dock
  - [x] When the right dock is hidden and the rightmost pane in the workspace center is focused, the focused pane remains focused
  - [x] When there is no pane to the right of the focused pane, the focused pane remains focused
- [x] `window:move-active-item-to-pane-above`
  - [x] You can move the focused pane's active item up from a workspace center pane to the adjacent workspace center pane above
  - [x] You can move the focused pane's active item up from the bottom dock to the bottommost pane in the workspace center
  - [x] When there is no pane above the focused pane, the focused pane's active item remains in the focused pane
- [x] `window:move-active-item-to-pane-below`
  - [x] You can move the focused pane's active item down from a workspace center pane to the adjacent workspace center pane below
  - [x] When the bottom dock is visible, you can move the focused pane's active item down from the bottommost pane in the workspace center to the bottom dock
  - [x] When the bottom dock is hidden and the bottommost pane in the workspace center is focused, the focused pane's active item remains in the focused pane
  - [x] When there is no pane below the focused pane, the focused pane's active item remains in the focused pane
- [x] `window:move-active-item-to-pane-on-left`
  - [x] You can move the focused pane's active item left from a workspace center pane to the adjacent workspace center pane on the left
  - [x] When the left dock is visible, you can move the focused pane's active item left from the leftmost pane in the workspace center to the left dock
  - [x] When the left dock is hidden and the leftmost pane in the workspace center is focused, the focused pane's active item remains in the focused pane
  - [x] When there is no pane to the left of the focused pane, the focused pane's active item remains in the focused pane
- [x] `window:move-active-item-to-pane-on-right`
  - [x] You can move the focused pane's active item right from a workspace center pane to the adjacent workspace center pane on the right
  - [x] When the right dock is visible, you can move the focused pane's active item right from the rightmost pane in the workspace center to the right dock
  - [x] When the right dock is hidden and the rightmost pane in the workspace center is focused, the focused pane's active item remains in the focused pane
  - [x] When there is no pane to the right of the focused pane, the focused pane's active item remains in the focused pane
- [x] `window:copy-active-item-to-pane-above`
  - [x] You can copy the focused pane's active item up from a workspace center pane to the adjacent workspace center pane above
  - [x] You can copy the focused pane's active item up from the bottom dock to the bottommost pane in the workspace center
  - [x] When there is no pane above the focused pane, nothing happens
- [x] `window:copy-active-item-to-pane-below`
  - [x] You can copy the focused pane's active item down from a workspace center pane to the adjacent workspace center pane below
  - [x] When the bottom dock is visible, you can copy the focused pane's active item down from the bottommost pane in the workspace center to the bottom dock
  - [x] When the bottom dock is hidden and the bottommost pane in the workspace center is focused, nothing happens
  - [x] When there is no pane below the focused pane, nothing happens
- [x] `window:copy-active-item-to-pane-on-left`
  - [x] You can copy the focused pane's active item left from a workspace center pane to the adjacent workspace center pane on the left
  - [x] When the left dock is visible, you can copy the focused pane's active item left from the leftmost pane in the workspace center to the left dock
  - [x] When the left dock is hidden and the leftmost pane in the workspace center is focused, nothing happens
  - [x] When there is no pane to the left of the focused pane, nothing happens
- [x] `window:copy-active-item-to-pane-on-right`
  - [x] You can copy the focused pane's active item right from a workspace center pane to the adjacent workspace center pane on the right
  - [x] When the right dock is visible, you can copy the focused pane's active item right from the rightmost pane in the workspace center to the right dock
  - [x] When the right dock is hidden and the rightmost pane in the workspace center is focused, nothing happens
  - [x] When there is no pane to the right of the focused pane, nothing happens


src/workspace-element.js
```diff
@@ -222,21 +222,101 @@ class WorkspaceElement extends HTMLElement {
...
+    const paneView = pane.getElement()
+    const box = this.boundingBoxForPaneView(paneView)
+
+    const visiblePaneContainers = atom.workspace.getPaneContainers()
+      .filter(container => {
+        const isCenter = container === atom.workspace.getCenter()
+        return isCenter || container.isVisible()
+      })
+
+    const visiblePanes = _.flatten(visiblePaneContainers.map(container => container.getPanes()))
```
jasonrudolph: I'm leaning toward extracting this out into `getVisiblePanes` function. Should that live on `WorkspaceElement` or `Workspace`?

jasonrudolph:
:wave: @as-cii @damieng: Would either of you have some cycles to review this PR in the next few days?

src/workspace-element.js
damieng: I like the idea of a getVisiblePanes function and to me Workspace is the place to put it given that other pane functions are there such as getPanes - one option would be to overload the latter with a "onlyVisible" boolean parameter if we want to keep the API surface down - I'm not convinced one way or the other there.

src/workspace-element.js
```diff
@@ -222,21 +222,101 @@ class WorkspaceElement extends HTMLElement {
...
+
+    const paneViews = visiblePanes
+      .map(otherPane => {
```
damieng: You can use the shorter syntax without { } and return for this, .e.g.

``map(otherPane => otherPane.getElement())```

src/workspace-element.js
```diff
@@ -222,21 +222,101 @@ class WorkspaceElement extends HTMLElement {
...
+      return Math.sqrt(Math.pow(x, 2) + Math.pow(y, 2))
+    }
 
-  moveActiveItemToPaneOnRight (params) { this.paneContainer.moveActiveItemToPaneOnRight(params) }
+    const paneView = pane.getElement()
+    const box = this.boundingBoxForPaneView(paneView)
+
+    const visiblePaneContainers = atom.workspace.getPaneContainers()
+      .filter(container => {
+        const isCenter = container === atom.workspace.getCenter()
```
damieng: Again you could filter this down using shorter syntax;

```javascript
.filter(container => container === atom.workspace.getCenter || container.isVisible())
```

Additionally it might be worth putting the atom.workspace.getCenter() before the filter so it only gets evaluated once as it doesn't change per item.  Probably not much of a perf issue here but can be a problem inside larger collections or if the calls are slow.

```javascript
const center = atom.workspace.getCenter()
const visiblePaneContainers = atom.workspace.getPaneContainers)
   .filter(container => container === center || container.isVisible())
```
```

File: src/workspace-element.js [hunk 4]
```diff
@@ -222,21 +222,101 @@ class WorkspaceElement extends HTMLElement {
...
+      activePane.getContainer().moveActiveItemToPane(nearestPaneView.getModel())
+    }
+    nearestPaneView.focus()
+  }
+
+  nearestVisiblePaneInDirection (direction, pane) {
+    const distance = function (pointA, pointB) {
+      const x = pointB.x - pointA.x
+      const y = pointB.y - pointA.y
+      return Math.sqrt(Math.pow(x, 2) + Math.pow(y, 2))
```
damieng: Thanks Pythagoras! :)

src/workspace.js
```diff
@@ -1349,6 +1349,10 @@ module.exports = class Workspace extends Model {
     return _.flatten(this.getPaneContainers().map(container => container.getPanes()))
   }
 
+  getVisiblePanes () {
```
jasonrudolph: I'm intentionally excluding this method from the public API for now. We could consider making this method part of the public API in the future, or we could consider rolling this method's functionality into the existing `::getPanes()` method. xref: https://github.com/atom/atom/pull/14640#discussion_r119168048

src/workspace.js
```diff
@@ -1532,6 +1536,12 @@ module.exports = class Workspace extends Model {
     ]
   }
 
+  getVisiblePaneContainers () {
```
jasonrudolph: I'm intentionally excluding this method from the public API for now. We could consider making this method part of the public API in the future, or we could consider rolling this method's functionality into the existing `::getPaneContainers()` method. xref: https://github.com/atom/atom/pull/14640#discussion_r119168048

jasonrudolph:
We use an item's `getAllowedLocations()` function to determine whether you can drag an item into a dock. Since this PR will allow you to move/copy items across pane containers, it seems like we should use the item's `getAllowedLocations()` function to determine whether you can move/copy an item into a given location. https://github.com/atom/atom/compare/5c5fb28da136fce8535874701179ff495ed30893...de47d3d439f228d4893201617d2c664415dc9d37 implements this logic. The current implementation adds an `isItemAllowed(item)` function on `Pane`. That feels to me like a reasonable home for that logic, but if there's a better place to put it, please let me know.

@maxbrunsfeld: Given your context on docks and allowed locations, would you mind reviewing those latest commits ☝️ and sharing any feedback you might have?

maxbrunsfeld (approved):
Makes perfect sense to me!

PR #14654, opened 2017-05-26
Improve confirm documentation

UziTech:
### Description of the Change

Updated atom.confirm documentation to include what happens if dialog is closed without selecting an option

### Possible Drawbacks

I'm not 100% sure this is correct on all OSes 

I had to do quite a bit of experimentation on Windows to figure out all the edge cases.

Most of the information I got from the [dialog.showMessageBox](https://github.com/electron/electron/blob/master/docs/api/dialog.md#dialogshowmessageboxbrowserwindow-options-callback) documentation but there seems to be quite a few differences between Oses

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -821,6 +821,9 @@ class AtomEnvironment extends Model
     @emitter.emit 'did-beep'
 
   # Essential: A flexible way to open a dialog akin to an alert dialog.
+  # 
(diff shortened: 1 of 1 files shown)


UziTech:
Any reason this hasn't been merged?

rsese:
Sorry for the delay here - we've since tweaked our process a bit to better track issues and PRs so hopefully things don't fall through the cracks like they did here.

Thanks for the update :bow:

Issue #11673, opened 2016-05-02
Shouldn't be able to resize a pane when there is only one


lee-dohm:
### Prerequisites
- [x] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
- [x] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
- [x] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
- [x] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
- [x] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
- [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?
### Description

Atom gets into a perplexing state for some people ([for example](https://discuss.atom.io/t/atom-editor-width-bug/28439)) whether by some combination of keys or package actions where there is only one pane, but it does not take up the full width of the editor window. Atom should not be able to get into this state.
### Steps to Reproduce
1. Launch Atom
2. Ensure there is only one pane in the window
3. Press <kbd>Cmd+Alt+-</kbd> multiple times

**Expected behavior:** Nothing to change

**Actual behavior:** The pane area for the single pane shrinks even though the window does not change size
### Versions

Atom    : 1.9.0-dev-5a9bd28
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1

apm  1.9.3
npm  2.13.3
node 0.10.40
python 2.7.10
git 2.8.1

lee-dohm:
Atom also shouldn't ever restore a resized single pane when starting up, so people shouldn't have to launch with `atom --clear-window-state` to fix this if they get into the weird state.

moehuh:
Experienced exactly this problem. `atom --clear-window-state` solves it, but I agree: Atom should not restore a resized single pane!

randohinn:
Experienced this. `atom --clear-window-state`was a fix. I suggest adding a menu item for it to the View menu.

lee-dohm:
[...]

PR #14665, opened 2017-05-28
Fix pane resizing issue

ghost:
Check if there is more than one pane in the pane-container before resizing the pane

### Requirements

### Description of the Change

When a request to resize a pane is sent, atom never checked how many panes were present in the pane-container. This resulted in a situation where if a package or the user tried to resize the pane, and there was only one pane, the pane would resize itself though there was no other pane to take up the space it freed.

The PR adds a check to the pane model, to check the number of panes in its container, before resizing the pane. If there is only one pane in the pane's container, it does not try to resize the pane, otherwise, resize as before.

### Alternate Designs

NA

Fixes incorrect behaviour when resizing panes.

### Benefits

Fixes issue #11673 

### Possible Drawbacks

None that I can think of

### Applicable Issues

#11673

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -97,9 +97,15 @@ class Pane
 
   getFlexScale: -> @flexScale
 
-  increaseSize: -> @setFlexScale(@getFlexScale() * 1.1)
+  increaseSize: ->
+    if @getContainer().getPanes().length == 1
+      return @getFlexScale()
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for helping out!

This will need tests before it can be reviewed though.

Ben3eeE:
Superseded by https://github.com/atom/atom/pull/16428

PR #14682, opened 2017-05-30
Bring back ASAR archives

anaisbetts:
As part of #13916, ASAR gets removed in favor of using snapshots. While snapshots solves the startup speed problem better than ASAR, using ASAR archives also helps a _lot_ with installation speed as well as the time taken to generate an installer. 

Refs https://github.com/desktop/dugite/pull/103
Refs https://github.com/desktop/dugite/pull/106

## TODO:

- [x] Make sure this actually generates a usable installer
- [x] Check to make sure this doesn't regress startup time
- [x] Look for shell scripts in the github package that need to be moved to `.unpacked`


smashwilson:
This has some implications for the GitHub package: specifically, there are a few shell scripts that we rely on being able to exec through git. Some of them are already being copied to a temporary folder before execution (as a way to deal with being bundled in the .asar to begin with) but not all of them are handled consistently.

Not a reason not to do this, just a heads-up for us to make sure we handle it and test against it :smile:

/cc @BinaryMuse @kuychaco

50Wliu:
Just realized that this _maybe_ also solves the spurious "missing file" issues we've been seeing.

alexandernst:
`Check to make sure this doesn't regress startup time`

Please triple-check this. I'd rather have a slower installer than a slower startup. I install Atom once every 3 months (after a release), while I open Atom every single day, several times per day. Having a faster startup speed and taking it away from the install speed seems like a good trade of.

Stanzilla:
@50Wliu it is, hitting the PATH length limit.

paulcbetts:
@alexandernst

alexandernst:
@paulcbetts I'm just saying that if there are any performance penalties in the start up speed, I'd rather have a slower installer. If there are no start up performance penalty, then sure, I'm fine with bringing back ASAR.

as-cii (approved):
If I recall correctly, introducing .asar archives has a performance penalty of about `50ms`: it's not extremely bad but everything adds up in that code path, so I decided to remove them.

Since we use snapshots, another solution to this problem could have been to not bundle files that are already embedded in the snapshot blob. The only reason why we are keeping those files around is because there is a bug in Chromium that prevents snapshotted code from being debugged, so I thought it would have been nice for users to open Atom in dev mode even if they don't have a local checkout of the atom/atom repository and step through source code.

Overall I think I am cool with adding .asar back, but I feel like we should keep in mind that we may be able to deprecate them again in favor of snapshots as soon as that debugging issue is solved upstream (in either Chromium or Electron) and we can avoid shipping those files entirely.

alexandernst:
Just out of curiosity. What is the performance difference between an installer with and without ASAR?
Are we talking about seconds? Minutes?

My point is, if the difference is ~10 seconds, it doesn't pay off adding 50ms (if @as-cii 's memory is correct) to the startup. Installing Atom is something that does not necessarily have to be fast, but opening the editor does have to be.

Stanzilla:
@as-cii how would that solve the PATH issue on Windows though?

paulcbetts:
> Just out of curiosity. What is the performance difference between an installer with and without ASAR?
Are we talking about seconds? Minutes?

Probably on slow machines, on the order of 10+ minutes, especially if they're applying delta updates. One ASAR file is _much_ faster to delta than 36,000 small files.

as-cii:
> Also, while ASAR might cause a slight startup penalty, in general it's going to make operations faster, especially on Windows where accessing lots of small files is pretty slow

Unless I am missing something, most of the core files are baked into the snapshot so I am not sure using .asar would buy us much in that regard. Still, it might be nice to profile startup time with and without .asar on Windows and get a sense of how performance looks like.

> @as-cii how would that solve the PATH issue on Windows though?

If the paths that exceed the length limit are snapshotted it means that (as soon as that Chromium bug is fixed) we can avoid shipping them as part of the bundle and include them all in the blob that's generated when using `mksnapshot`.

paulcbetts:
Alright, after the typo fix in desktop/dugite#106, it looks like basic Git operations work, though I still need to check pushing via SSH

as-cii:
A quick benchmark of startup performance confirms the extra cost introduced by setting up .asar in Electron is `~50ms`:

screen shot 2017-08-15 at 10 22 50

For now, I think this is a good tradeoff to minimize installation time. Once https://github.com/electron/electron/issues/10234 lands, however, I think we should revisit this for the reasons outlined in https://github.com/atom/atom/pull/14682/#pullrequestreview-41262108.

I'll go ahead and merge this to master as soon as we have a green build. Thanks for your work on this, @paulcbetts. ⚡️

PR #14690, opened 2017-06-01
Add keyboard accelerators to file change & window state dialogs

### LINKED ISSUE #2928: Changed File Close Dialogue not usable with keyboard
(this is possibly windows only, tested with 0.113.0)

**Steps To Reproduce**
- Open Atom
- Type something into a new editor or change an existing file (doesn't matter which)
- Attempt to close the tab (for example `Ctrl+W`)
- Correctly, a dialogue opens, that will give you a choice how to proceed

close-dialog-keyboard-lacking

**current (faulty) behaviour**

The Keyboard is pretty much useless for this dialogue. You can only Hit `[Enter]` to save.

**Desired Behaviour**

These interactions all don't work, but would be very much needed, all of these are very much standard behaviour for any application:
- [X] `[left]` and `[right]` should switch focus between buttos (so we can select a different thing with keyboard only)
- [X] `[esc]` should Cancel
- [ ] `[alt]` should bring up shortcut labels (much like it does for the application menu, i.e. _F_ile, which means press `[Alt+F]`)
- [X] `[tab]` should cycle between the buttons, `[shift-tab]` should cycle backwards.

damieng:
Fixes #2928

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -933,8 +933,8 @@ class AtomEnvironment extends Model
           "Would you like to add the #{nouns} to this window, permanently discarding the saved state, " +
           "or open the #{nouns} in a new window, restoring the saved state?"
         buttons: [
-          'Open in new window and recover state'
-          'Add to this window and discard state'
+          '&Open in new window and recover state'
+          '&Add to this window and discard state'
         ]
       if btn is 0
         @open
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -640,7 +640,7 @@ class Pane
       chosen = @applicationDelegate.confirm
         message: message
         detailedMessage: "Your changes will be lost if you close this item without saving."
-        buttons: [saveButtonText, "Cancel", "Don't Save"]
+        buttons: [saveButtonText, "Cancel", "&Don't Save"]
       switch chosen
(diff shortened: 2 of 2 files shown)


neok-m4700:
@damieng I don't see these changes on master, nor on recent releases, aren't they supposed to be merged ?
Thanks

damieng:
They are definitely on master https://github.com/atom/atom/blob/master/src/atom-environment.coffee#L941and https://github.com/atom/atom/blob/master/src/pane.coffee#L668

I'm also seeing then in use - you do have to press Alt to see the accelerator - that's just the way Windows 10 presents them now.  Are you on Linux or macOS?

image

neok-m4700:
Hi, ok sorry I mean't relases e.g. latest stable https://github.com/atom/atom/archive/v1.19.3.tar.gz. Hope they will be included in 1.20.0 ?

I'm on Linux (Ubuntu 17.04), will try to patch manually then.

Ben3eeE:
It is in the 1.20 beta right now so it will be included in 1.20.0

HelmutPla:
This does not seem to work on macOS. I have tried all possible accelerator keys but no success (I am on macOS Ventura (13.0.1) and Atom 1.60.)

Issue #3466, opened 2014-09-05
"Show Line Numbers" option removes the gutter when turned off


lee-dohm:
Which is very rarely what the user wants. The gutter is a location that shows much other useful information, disabling it necessarily disables all that other functionality. The Show Line Numbers option should just hide the line numbers and _not_ remove the gutter completely.

See Discuss topics:
- https://discuss.atom.io/t/use-the-gutter-without-enabling-linenumbers/2461
- https://discuss.atom.io/t/hide-line-numbers/11987/2

r-owen:
I also would like this feature. The gutter is crucial for linting, but I don't find line numbers useful.

I suggest independent options to show/hide line numbers and show/hide gutter.

benogle:
This will be much easier when @nathansobo is all done with the editor internal changes / performance tuning.

skariel:
+1

spencerlyon2:
:+1:

kaicataldo:
Still looking for help with this? I can take a look at it this week

lee-dohm:
Yes, help would be wonderful :grinning:

kaicataldo:
Cool :) Would having independent options to show/hide line numbers and show/hide the gutter as suggested above be the preferred approach?

kaicataldo:
@r-owen lee-dohm So what's really missing is the git diff in the gutter, correct? I'm using https://github.com/atom-community/linter and am still seeing lines with errors still have a mark and show the correct error. The git diff is currently part of the line number elements themselves, but I don't think it would be difficult to change that.

<img width="533" alt="screen shot 2016-01-19 at 12 54 20 am" src="https://cloud.githubusercontent.com/assets/7041728/12410883/320ba4cc-be47-11e5-91e2-7ed3159d0ccb.png">

wandoujun:
@kaicataldo What is the name of the theme that you using in your screenshots? It is beautiful ~!

kaicataldo:
@wandoujun Hey there, I'm not using Atom much these days and I don't seem to have that theme installed any more, but I think it's a port of [Oceanic Next](https://atom.io/themes/oceanic-next) from Sublime

wandoujun:
@kaicataldo Thanks~man.It is really helpful.

nivv:
Any progress on this? I don't like the clutter line numbers bring. But the git gutter itself is neat

GitTom:
+1

TotomInc:
[...]

PR #13880, opened 2017-02-26
Rewrite editor rendering layout to use new browser features and virtual DOM

nathansobo:
In this PR, I plan to completely rewrite editor rendering to take advantage of CSS layout containment and  intersection observers. This PR will also use Etch to perform declarative updates instead of the existing presenter with manual DOM manipulation, which has not aged well. I'm optimistic that a simpler layout, a cleaner approach to DOM updates, and better use of features such as layout containment can make our rendering faster and much more maintainable.

My goal is to rethink all of the DOM interactions associated with the editor to make sure things are as optimal as possible.

## Tasks

* [x] Content
  * [x] Render initial lines
  * [x] Render initial line numbers
  * [x] Update on scroll
  * [x] Update on buffer changes
  * [x] Adjust soft wrap on resize
  * [x] Scroll past end option
  * [x] Mini editors working
  * [x] Hide line numbers option
  * [x] Placeholder text
  * [x] Auto height and width
  * [x] Synthetic scrolling
  * [x] Dummy scrollbars
* [x] Cursors
  * [x] Basic rendering
  * [x] Auto-scroll
    * [x] Vertical
    * [x] Horizontal
  * [x] Blinking
  * [x] Option to hide with non-empty selections
  * [x] Cursors positioned after a fold marker are not displayed correctly
  * [x] Official API for customizing cursor styling
* [x] Input handling
  * [x] Position hidden input element
  * [x] Focus
  * [x] Editor commands
  * [x] Text input
  * [x] IME input
  * [x] Disabled input
  * [x] Mouse handling
    * [x] Single click to set cursor position
    * [x] Double click to select word
    * [x] Triple click to select line
    * [x] Add cursors with cmd-click
    * [x] Expand selections with shift-click
    * [x] Drag selections
    * [x] Autoscroll when dragging selections
    * [x] Autoscroll when dragging gutter
    * [x] Click/drag gutter to select lines
    * [x] Click gutter to toggle folding
    * [x] Destroy folds when clicking fold markers
    * [x] Middle mouse paste on Linux
    * [x] Scroll horizontally on shift-wheel
* [x] Decorations
  * [x] Make marker layer events synchronous
  * [x] Line
  * [x] Line Number
  * [x] Highlights
  * [x] Highlight flashes
  * [x] Overlays
  * [x] Custom gutters
  * [x] Blocks
  * [x] Never decorate invalidated markers
  * [x] Fix git-diff gutter decorations being shown at the end of the tile
* [x] Update on external changes
[...]

PR #14694, opened 2017-06-01
Render line number gutter without numbers when showLineNumbers is false

nathansobo:
In #13880 we honored the `lineNumberGutterVisible` parameter but neglected to honor the `showLineNumbers` parameter, which is what is used by the text editor registry to interact with the settings. In the previous editor implementation, both of these options simply hid the line number completely. In this version, we've elected to handle `showLineNumbers` differently from `lineNumberGutterVisible`. We continue to render the line number gutter (which contains diff decorations, fold indicators, etc), but we just don't render the line numbers in it. This provides more features and a better visual experience when the line numbers are disabled.

🍐ed with @as-cii

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -596,9 +596,30 @@ describe('TextEditorComponent', () => {
       )
     })
 
-    it('supports the isLineNumberGutterVisible parameter', () => {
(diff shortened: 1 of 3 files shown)


50Wliu:
@nathansobo Does this fix #3466?

50Wliu:
Answering my own question...yes it does.

PR #14695, opened 2017-06-01
Provide API for observing the active text editor

### LINKED ISSUE #194: Status bar is blank until file is "open" (annoying)
### Prerequisites

* [ x] Put an X between the brackets on this line if you have done all of the following:

### Description

The status bar is empty until a file is hard "opened" to a permanent tab.

It's really annoying not being able to see what git branch or file path I'm actually looking at.

### Steps to Reproduce

1. Click a file in the sidebar
2. The status bar is empty (no file, encoding, git info) — GRRRRR
3. Double-click the file
4. Now the status bar works (for "open" tabs only)

**Expected behavior:** No matter what the active tab is, the full status bar should render.

**Actual behavior:** Until a file is "open" the status bar is empty.

### Versions

I think this started in Atom 1.16 (ish). Occurs now in 1.17.1.

### LINKED ISSUE #14648: Opening a pending pane no longer correctly correctly triggers activation
### Prerequisites

### Description

Opening a file as a pending pane no longer triggers activation as expected. The end result is that `atom.workspace.getActiveTextEditor` returns `undefined` until the user interacts with the text editor in some manner (e.g. moving the cursor or making the pane non-pending).

### Steps to Reproduce

1. Launch Atom in any folder that contains some files that can be opened.
2. Click on any file in the tree view (ensure it is opened in a pending pane).
3. Open up the developer tools console.
4. Execute `atom.workspace.getActiveTextEditor()`
5. Notice the method returns `undefined`
6. Click anywhere in the (pending) text editor.
7. Execute `atom.workspace.getActiveTextEditor()`
8. Notice the method returns a `TextEditor` object.

**Expected behavior:** Opening a file in a pending pane should immediately mark the text editor as active.

**Actual behavior:** The text editor in the pending pane is not marked as active until the user directly interacts with the pane/editor.

**Reproduces how often:** 100% of the time.

### Versions

Atom v1.7.2, and v1.19.0-dev-85e65ffd1.
Ubuntu 16.04 LTS

### Additional Information

This regression is the cause of the following issues:
* https://github.com/atom/status-bar/issues/194

jasonrudolph:
### Description of the Change

As the first step toward resolving #14648, this pull request implements the proposal outlined in https://github.com/atom/atom/issues/14648#issuecomment-305017095:

> - Disallow editors in docks. Only allow editors in the workspace center.
> - Change `atom.workspace.getActiveTextEditor()` to return the active text editor (if any) in the workspace center. (This is more consistent with the pre 1.17.0 behavior described in https://github.com/atom/atom/issues/14648#issuecomment-304381899.)
> - Introduce `atom.workspace.observeActiveTextEditor()` as a means of observing the active text editor. (Packages like encoding-selector and grammar-selector _want_ to observe the active text editor, but currently have to rely on observing the active pane item, which is problematic for the reasons described in https://github.com/atom/atom/issues/14648#issuecomment-304391943.)

Once this pull request is merged, I'll then be able to update each of the core packages that currently use `atom.workspace.observeActivePaneItem()` to instead use `atom.workspace.observeActiveTextEditor()` to receive notifications each time a different text editor becomes active. Once that's done, we'll be able to close #14648.

### Alternate Designs

- For packages that need to observe the active text editor (e.g., https://github.com/atom/status-bar/issues/194), we could require those packages to observe all pane containers and watch for a text editor to become active in that pane container. Because observing the active text editor is a common need, I'd rather provide a core API for this functionality and not require each package to implement and maintain this logic.
- Instead of disallowing editors in docks, we could teach `atom.workspace.getActiveTextEditor()` to first check for an active text editor in the active pane container and return it if it exists, and to otherwise return the active text editor (if any) in the workspace center. `atom.workspace.observeActiveTextEditor()` would need to use similar logic to notify observers each time the value of `atom.workspace.getActiveTextEditor()` changes. This approach would likely allow us to resolve https://github.com/atom/status-bar/issues/194, but it feels overly complicated, and it seems like something that is more likely to cause future confusion, surprise, and additional maintenance overhead.

It's the first step toward resolving #14648.

### Benefits

- Developers will have a core API for getting notified when:
  - A text editor becomes the active text editor
  - There is no longer an active text editor
- The behavior of `atom.workspace.getActiveTextEditor()` will [more closely match its pre-1.17.0 behavior](https://github.com/atom/atom/issues/14648#issuecomment-304381899), which various [packages relied on](https://github.com/atom/find-and-replace/pull/904).
- We answer the question, "Does it make sense to have an editor in a dock?" Answer: No. Atom expects to find editors in the workspace center. Docks are intended to provide supporting tools like a tree view, a debugger, a console. xref: http://blog.atom.io/2017/05/23/docks-deep-dive.html
- It reduces (hopefully 🤞) the overall number of scenarios that need to be considered in Atom core and in packages that interact with editors (e.g., https://github.com/atom/find-and-replace/issues/901#issuecomment-304078622).

### Possible Drawbacks

- By disallowing editors in docks, we reduce Atom's overall flexibility. If someone comes up with a compelling reason for using an editor in a dock, they're out of luck.
- Implementing `getAllowedLocations()` on an item affects where the item opens when using `atom.workspace.open()` and when [dragging an item to a dock in the UI](https://github.com/atom/atom/blob/16e1ef917be66637a58bad5e09c7d546782ee6e8/src/dock.js#L228-L229). However, functions like [`Pane::addItem()`](https://atom.io/docs/api/v1.17.2/Pane#instance-addItem) currently allow you to add the item to any location, regardless of the locations returned by the item's `getAllowedLocations()` function. In other words, there will continue to be ways to programmatically add a text editor to a dock, even though we're discouraging it and considering it to be unsupported behavior.

### Applicable Issues

- #14648
- https://github.com/atom/status-bar/issues/194

### TODO

- [x] Add `TextEditor::getAllowedLocations()` to indicate that editors belong in the workspace center
- [x] Add `Workspace::observeActiveTextEditor()`
- [x] Add API docs for new functions in `Workspace`
- [x] Determine and implement reasonable behavior for existing text editor functions in `Dock`:
  - [x] [`observeTextEditors`](https://github.com/atom/atom/blob/16e1ef917be66637a58bad5e09c7d546782ee6e8/src/dock.js#L387-L400) -- 0b314ac1a98ca06c82fc369fffe4f13fc772abe5
  - [x] [`getTextEditors`](https://github.com/atom/atom/blob/16e1ef917be66637a58bad5e09c7d546782ee6e8/src/dock.js#L586-L591) -- 09495dfc35b1294df0f4785e00d4eb1f8c7e27fc
  - [x] [`getActiveTextEditor`](https://github.com/atom/atom/blob/16e1ef917be66637a58bad5e09c7d546782ee6e8/src/dock.js#L593-L600) -- f8ebd71200af4dcc5cf6fc2d85e68247833d9acc

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -345,8 +345,7 @@ module.exports = class Workspace extends Model {
         left: this.paneContainers.left.serialize(),
         right: this.paneContainers.right.serialize(),
         bottom: this.paneContainers.bottom.serialize()
-      },
-      hasActiveTextEditor: this.hasActiveTextEditor
+      }
     }
   }
 
@@ -373,9 +372,7 @@ module.exports = class Workspace extends Model {
       this.paneContainers.center.deserialize(state.paneContainer, deserializerManager)
     }
 
-    if (state.hasActiveTextEditor != null) {
-      this.hasActiveTextEditor = state.hasActiveTextEditor
-    }
+    this.hasActiveTextEditor = this.getActiveTextEditor() != null
 
     this.updateWindowTitle()
   }
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -427,7 +427,7 @@ module.exports = class Workspace extends Model {
     if (paneContainer === this.getCenter()) {
       const itemIsTextEditor = item instanceof TextEditor
       const activeTextEditorChanged =
-        itemIsTextEditor || (this.activeTextEditor instanceof TextEditor)
+        itemIsTextEditor || (this.activeTextEditor != null)
 
       if (activeTextEditorChanged) {
         this.activeTextEditor = itemIsTextEditor ? item : null
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -216,6 +216,7 @@ module.exports = class Workspace extends Model {
       bottom: this.createDock('bottom')
     }
     this.activePaneContainer = this.paneContainers.center
+    this.activeTextEditor = null
 
     this.panelContainers = {
       top: new PanelContainer({viewRegistry: this.viewRegistry, location: 'top'}),
@@ -421,6 +422,17 @@ module.exports = class Workspace extends Model {
     if (paneContainer === this.getActivePaneContainer()) {
       this.didChangeActivePaneItem(item)
       this.emitter.emit('did-change-active-pane-item', item)
+    }
+
+    if (paneContainer === this.getCenter()) {
+      const activeTextEditorChanged =
+        (item instanceof TextEditor) || (this.activeTextEditor instanceof TextEditor)
+
+      if (activeTextEditorChanged) {
+        this.activeTextEditor = (item instanceof TextEditor) ? item : null
+        const itemValue = (item instanceof TextEditor) ? item : undefined
+        this.emitter.emit('did-change-active-text-editor', itemValue)
+      }
     }
   }
 
@@ -648,6 +660,10 @@ module.exports = class Workspace extends Model {
     return this.emitter.on('did-stop-changing-active-pane-item', callback)
   }
 
+  onDidChangeActiveTextEditor (callback) {
+    return this.emitter.on('did-change-active-text-editor', callback)
+  }
+
   // Essential: Invoke the given callback with the current active pane item and
   // with all future active pane items in the workspace.
   //
@@ -658,6 +674,13 @@ module.exports = class Workspace extends Model {
   observeActivePaneItem (callback) {
     callback(this.getActivePaneItem())
     return this.onDidChangeActivePaneItem(callback)
+  }
+
+  observeActiveTextEditor (callback) {
+    const activeTextEditor = this.getActiveTextEditor()
+    if (activeTextEditor != null) { callback(activeTextEditor) }
+
+    return this.onDidChangeActiveTextEditor(callback)
   }
 
   // Essential: Invoke the given callback whenever an item is opened. Unlike
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1282,12 +1282,12 @@ module.exports = class Workspace extends Model {
     return this.getPaneItems().filter(item => item instanceof TextEditor)
   }
 
-  // Essential: Get the active item if it is an {TextEditor}.
+  // Essential: Get the workspace center's active item if it is a {TextEditor}.
   //
-  // Returns an {TextEditor} or `undefined` if the current active item is not an
-  // {TextEditor}.
+  // Returns a {TextEditor} or `undefined` if the workspace center's current
+  // active item is not a {TextEditor}.
   getActiveTextEditor () {
-    const activeItem = this.getActivePaneItem()
+    const activeItem = this.getCenter().getActivePaneItem()
     if (activeItem instanceof TextEditor) { return activeItem }
   }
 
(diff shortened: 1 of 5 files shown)


src/workspace.js
```diff
@@ -422,6 +423,17 @@ module.exports = class Workspace extends Model {
       this.didChangeActivePaneItem(item)
       this.emitter.emit('did-change-active-pane-item', item)
     }
+
+    if (paneContainer === this.getCenter()) {
+      const activeTextEditorChanged =
+        (item instanceof TextEditor) || (this.activeTextEditor instanceof TextEditor)
```
maxbrunsfeld: I think it'd be nice to consolidate these `item instanceof TextEditor` checks into one. Also, I would think we would not need to check `this.activeTextEditor instanceof TextEditor`, because `this.activeTextEditor` is always either a text editor or null.

maxbrunsfeld (commented):
This looks great! I just suggested one change.

src/workspace.js
jasonrudolph: Thanks, @maxbrunsfeld. :bow:

I pushed up a few updates that attempt to incorporate these suggestions. If these updates are different from what you had in mind, just let me know.

jasonrudolph:
5b61c0a949e70d258b5ce2657dc9a1f5507d1878 fixes a bug that I uncovered related to serialization/deserialization.

@maxbrunsfeld: This is my first time touching any serialization/deserialization code. Would you mind taking a look at 5b61c0a949e70d258b5ce2657dc9a1f5507d1878 and letting me know if there's anything I should change?

ungb:
@jasonrudolph I found a few small bugs.  

1.  If you select tree view you can't change the line endings

focustree

2. says file uses undefined line ending (could be unrelated to your pr)
image

3.  If you select the tree view last, close and reopen Atom the file ending is missing on start up, hard to repro.  Also doesn't repro if you completely exit out of Atom.  If you have the file selected last, it seems to work correctly.
missingle

ungb:
issue number 2 I listed, looks like it was a recent pr that was merged. see https://github.com/atom/line-ending-selector/pull/42

jasonrudolph:
> NOTE: 1,2 are issues that already exist on master so they aren't caused by your pr. The only one I think is no.3, but currently you can't see any of the items in the status bar. PR is a HUGE improvement over what's currently out.

@ungb: Thanks for finding these issues. :zap: https://github.com/atom/line-ending-selector/commit/2ff4ee3f833c6dc8489518aa4815e20238011f4d should resolve # 1 and # 3.

lgeiger:
Thanks for adding this API! This is a lot more convenient than doing:
```js
atom.workspace.getCenter().observeActivePaneItem(item => {
  const editor = item instanceof TextEditor ? item : null;
  ...
})
```
May I suggest to backport this API to Atom 1.18 beta if it can be cherry-picked cleanly?

jasonrudolph:
> May I suggest to backport this API to Atom 1.18 beta if it can be cherry-picked cleanly?

@lgeiger: Yes. 😸  Before the 1.18 stable release, we're hoping to publish a new 1.18-beta release including this fix.

src/workspace.js
marlazinger200: 1

marlazinger200:
1

PR #14701, opened 2017-06-02
Remove unneeded files from build result

anaisbetts:
This PR strips a bunch of unnecessary files from our build result, which should make Atom smaller as well as reduce build times.


iolsen:
The tests that fail ☝️ are irrelevant to the change and are passing locally. Merging.

PR #14711, opened 2017-06-03
Native tabs on macOS

### LINKED ISSUE #14197: Window tabs and file tabs overlap with hidden title bar enabled on MacOS
### Prerequisites

### Description

With the new `hidden` title bar enabled, MacOS' native window tabs appear on top of Atom's file tabs rather than above them.

screen shot 2017-04-13 at 12 45 21 pm

### Steps to Reproduce

1. Enable `hidden` title bar.
2. Relaunch Atom.
3. Open multiple windows.

**Expected behavior:** Window tabs appear above file tabs.

**Actual behavior:** Window tabs overlap file tabs.

### Versions

```
Atom    : 1.16.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
apm  1.16.1
npm  3.10.5
node 6.9.5 x64
python 2.7.13
git 2.11.0
MacOS 10.12.4
```

### Additional Information

Problem occurs across all UI and syntax themes.

simurai:
### Description of the Change

This enables native tabs on macOS. Under `macOS > Menu > View` there will be a "Show Tab Bar" option.

screen shot 2017-06-03 at 5 10 32 pm

Which lets you combine multiple Atom windows and tab between them:

native-tabs

> Note: Only available if `Atom > Settings > Core > Title Bar` is set to `native`.

### Alternate Designs

Was wondering if it should be a config, but I think it's not needed. Because:

1. Electron is smart enough to [only show](https://github.com/electron/electron/blob/dfab1043d98067b43b45d8dcbe9d0b84d4820555/atom/browser/native_window_mac.mm#L789-L790) the native tabs if there is also a native title bar. So [issues](https://github.com/atom/atom/issues/14197) with hidden/custom title bar shouldn't be a problem.
2. You can control if Atom should open as a native tab or new window by going to `macOS > System Preferences > Docks` and change this:

    screen shot 2017-06-03 at 4 40 33 pm

It's part of the `browserWindow` options.

### Benefits

Makes it possible to tab between multiple Atom windows.

### Possible Drawbacks

People might not like this change if:

1. They prefer opening as tabs for all other apps (like Finder), but not for Atom.
2. They don't want to switch to a custom title bar.

Something else I noticed: If you have the DevTools open (as a separate window) and close the Atom window, the DevTools window moves as a tab to another existing Atom window. Not sure if that's an edge case to be ok with?

### Applicable Issues

Closes https://github.com/atom/atom/issues/14197


primalmotion:
simple and badly needed fix. Is there a way to hack this in without recompiling Atom?

tsdorsey:
@simurai what kind of testing did you need done? I've running a local build of Atom based on on commit e48b980d41f7a41a6cce0faef45a02b919a07942 with tabbing enabled via this method. Everything is working as expected.

The new window command line argument still works as expected.
The app responds to the dock setting for tabs as expected.

simurai:
@tsdorsey thanks for giving it another test run. I also merged in master and tried it again without issues.

AppVeryor fails, but I think it's unrelated 👇 

```
Build execution time has reached the maximum allowed time for your plan (90 minutes).
```

50Wliu:
Due to https://appveyor.statuspage.io/incidents/2tjbzx1mtnz8

simurai:
@50Wliu thanks.. 🙇  ok merged. 🚢

TerrenceLJones:
I might have missed the discussion around this, but how do you cycle through the Atom native panes? `cmd-shift-{` / `cmd-shift-}` do not cycle through native panes, but navigate through Atom panes (as was the case before this change). Is there currently a way to cycle through the native Atom panes? And if not, is there a plan to add support for this? Thanks, super excited about this feature!

rsese:
As far as I can tell, Atom doesn't currently support switching between native tabs with some command that you can add a keybinding for - can you open an issue for this @TerrenceLJones and fill out the issue template as much as possible so we ca track it?

mpacer:
The native tabs also seem to open in unpredictable way… what is the setting to choose to get back the previous default behaviour?

loadingwyn:
There is no "Show Tab Bar" option under `View`. I'm sure my `Title Bar` is set to `native`.

### Versions

Atom : 1.21.0
OS: MacOS 10.13

rsese:
> what is the setting to choose to get back the previous default behaviour?

There was no config added (see https://github.com/atom/atom/pull/14711 for the rationale) so I believe Atom will follow whatever you have set in `macOS > System Preferences > Docks`.   As mentioned in the PR, if you have `Settings > Core > Title Bar` set to anything besides `native`, Atom also won't use native tabs in this case.

> There is no "Show Tab Bar" option under View. I'm sure my Title Bar is set to native.

I don't have 10.13 to test on but I can confirm that it's available on 10.12.6.  @ungb do you have High Sierra installed to check?

ungb:
I don't have a high sierra just 10.12.6.

@lee-dohm can you check this if you have mac OSX 10.13?

lee-dohm:
With "Title Bar" set to `native` ...

screen shot 2017-10-05 at 7 53 49 am

:point_up: using Atom v1.23.0-dev-c9c495792 on Mac OS X 10.13

screen shot 2017-10-05 at 7 55 27 am

:point_up: using Atom v1.21.0 on Mac OS X 10.13

lee-dohm:
@loadingwyn and others ... If you have what you believe to be a bug, please open a new issue and fill in the issue template with as much detail as you can. All of this discussion is not related to the PR.

Issue #13195, opened 2016-11-11
Feature request: return promise when dispatching commands


UziTech:
It would be nice to have a standard way to tell when a dispatched command finished or failed.

The atom.commands.dispatch command could return a promise or it could return whatever is returned by the command function

``` javascript
atom.commands.add("atom-workspace", "my-package:test-promise",  function() {
  return new Promise(function (resolve, reject) {
    setTimeout(_ => resolve("test"), 1000);
  });
});

atom.commands.dispatch(document.querySelector("atom-workspace"), "my-package:test-promise").then(function(result) {
  // do something
});
```

PR #14720, opened 2017-06-05
Return command value

UziTech:
### Description of the Change

Right now `atom.commands.dispatch` returns true/false to tell whether a dispatched command exists.

This would return an array of information for the executed commands including the `currentTarget`, `dispatchEvent`, and return value of the command.

This will allow the dispatcher to view the information for each listener of the command. 

### Alternate Designs

I was going to return a `Promise` with `Promise.all()` but I thought this would be better for a couple reasons.

 - `Promise.all()` will reject if only 1 `Promise` fails and the rest of the listeners info will be lost.
 - `Promise.all()` will only let us get the return value of each listener.

To keep backwards compatibility I return false if there are no matches. We could change that in the future.

This requires changes to the command registry in core.

### Benefits

- Ability to get a return value from a command.
- Ability to tell when an async command actually finishes.
- Ability to dispatch commands in series.

### Possible Drawbacks

none

### Applicable Issues

#13195


50Wliu:
> To keep backwards compatibility I return false if there are no matches.

Curious: what if a package was doing a strict equality check for `true`?  It seems like this only preserves backwards compatibility for when the command doesn't exist.

UziTech:
That is correct. This will break backwards compatibility if they are strictly checking for `true`

UziTech:
In testing this seems a bit less practical than I originally thought.

The problem is that a command listener can be added by any package and knowing which order listeners will be called in (and therefore be returned in) is not easy.

I'm thinking returning a single `Promise.all` or just an array of Promises would be better. At least the dispatcher would be able to use the promise(s) to tell when the command was finished and run the next command in series.

UziTech:
I'm not sure if the `currentTarget` or `dispatchEvent` are needed in the return value

UziTech:
One option to not break backwards compatibility would be to add an `onDidFinish` event

UziTech:
I moved this to a new pull request since it got a little off topic #14931

PR #14460, opened 2017-05-17
Fix ability to activate next/previous pane in a dock

jasonrudolph:
Fixes #14442.

### Description of the Change

Prior to this change, when a dock had focus, triggering `window:focus-next-pane` or `window:focus-previous-pane` would throw an error:

```
At /Applications/Atom.app/Contents/Resources/app/src/workspace.js:1559

TypeError: this.getActivePaneContainer(...).activateNextPane is not a function
    at Workspace.activateNextPane (/Applications/Atom.app/Contents/Resources/app/src/workspace.js:1559:1)
    at atom-workspace.window:focus-next-pane (/Applications/Atom.app/Contents/Resources/app/src/register-default-commands.js:168:1)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app/src/command-registry.js:265:1)
    at /Applications/Atom.app/Contents/Resources/app/src/command-registry.js:3:1
    at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:610:1)
    at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:401:1)
    at WindowEventHandler.module.exports.WindowEventHandler.handleDocumentKeyEvent (/Applications/Atom.app/Contents/Resources/app/src/window-event-handler.js:100:1)
    at HTMLDocument.<anonymous> (/Applications/Atom.app/Contents/Resources/app/src/window-event-handler.js:3:1)
```

[`Workspace.activateNextPane()` calls `this.getActivePaneContainer().activateNextPane()`](https://github.com/atom/atom/blob/ae9f24602d80cb15fe0edc15c1aae37d44ee786c/src/workspace.js#L1361). `this.getActivePaneContainer()` returns either a `WorkspaceCenter` (if the workspace center is the active pane container) or a `Dock` (if a dock is the active pane container). `WorkspaceCenter` has an [`activateNextPane`](https://github.com/atom/atom/blob/8ed16cea43cfd3f3b04b51ed5851892944ac7f19/src/workspace-center.js#L290-L298) function, but `Dock` does not. Therefore, calling `Workspace.activateNextPane()` results in an error when a `Dock` is the active pane container. (The same goes for `activatePreviousPane`.)

[...]

Issue #14442, opened 2017-05-17
Uncaught TypeError: this.getActivePaneContainer(...).activateNextPane is not a function


AgentAntelope:
[Enter steps to reproduce:]

1. Open a dock (I used the sidebar, but this works in any dock)
2. Trigger `window:focus-next-pane`

expected:
EITHER
1) Next pane/dock is selected
2) Nothing happens, as dock only has one pane

actual:
The below error happens.

**Atom**: 1.17.0 x64
**Electron**: 1.3.15
**OS**: Mac OS X 10.12.4
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: this.getActivePaneContainer(...).activateNextPane is not a function

```
At /Applications/Atom.app/Contents/Resources/app/src/workspace.js:1559

TypeError: this.getActivePaneContainer(...).activateNextPane is not a function
    at Workspace.activateNextPane (/Applications/Atom.app/Contents/Resources/app/src/workspace.js:1559:1)
    at atom-workspace.window:focus-next-pane (/Applications/Atom.app/Contents/Resources/app/src/register-default-commands.js:168:1)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app/src/command-registry.js:265:1)
    at /Applications/Atom.app/Contents/Resources/app/src/command-registry.js:3:1
    at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:610:1)
    at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:401:1)
    at WindowEventHandler.module.exports.WindowEventHandler.handleDocumentKeyEvent (/Applications/Atom.app/Contents/Resources/app/src/window-event-handler.js:100:1)
    at HTMLDocument.<anonymous> (/Applications/Atom.app/Contents/Resources/app/src/window-event-handler.js:3:1)
```

### Commands

```
  2x -0:51.1.0 window:toggle-left-dock (input.hidden-input)
  3x -0:45.6.0 pane:show-previous-item (input.hidden-input)
  6x -0:40.7.0 pane:show-next-item (input.hidden-input)
  5x -0:39.5.0 pane:show-previous-item (input.hidden-input)
[...]

PR #14729, opened 2017-06-07
Fix Docks docs

### LINKED ISSUE #14728: Docks documentation points to non-existent get{Left|Right|Bottom}Dock methods
(Issue template does not apply)

The [Docks](https://atom.io/docs/api/v1.17.2/Dock) documentation links to `getLeftDock`, `getRightDock`, and `getBottomDock`.  However, there are two problems: those methods exist in Workspace and not Dock; and those methods are not publicly documented.

winstliu:
### Requirements

### Description of the Change

Fixes some issues with the new Docks documentation.  In particular, the documentation to `get{Left|Right|Bottom}Dock` was pointing to the wrong class and the methods themselves weren't public.

Please note: While adding documentation to the dock methods, I noticed that while `WorkspaceCenter` had docs for its individual methods, it did not have documentation for the class itself and therefore was not showing up as part of the API.  Therefore I've added a short blurb to the class.

The documentation I added was based off my best judgement and the comments for similar classes/methods; please feel free to improve them.

### Alternate Designs

N/A

Because the documentation is in core.

### Benefits

Better documentation.

### Possible Drawbacks

None.

### Applicable Issues

Fixes #14728

I have no clue how to test this.

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1548,22 +1548,27 @@ module.exports = class Workspace extends Model {
   Section: Pane Locations
   */
 
+  // Essential: Get the {WorkspaceCenter} at the center of the editor window.
   getCenter () {
     return this.paneContainers.center
   }
 
+  // Essential: Get the {Dock} to the left of the editor window.
   getLeftDock () {
     return this.paneContainers.left
   }
 
+  // Essential: Get the {Dock} to the right of the editor window.
   getRightDock () {
     return this.paneContainers.right
   }
 
+  // Essential: Get the {Dock} below the editor window.
   getBottomDock () {
     return this.paneContainers.bottom
   }
 
+  // Essential: Get an {Array} of the current {WorkspaceCenter} and {Dock}s.
   getPaneContainers () {
     return [
       this.paneContainers.center,
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1568,7 +1568,6 @@ module.exports = class Workspace extends Model {
     return this.paneContainers.bottom
   }
 
-  // Essential: Get an {Array} of the current {WorkspaceCenter} and {Dock}s.
   getPaneContainers () {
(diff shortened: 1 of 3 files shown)


src/workspace.js
```diff
@@ -1548,22 +1548,27 @@ module.exports = class Workspace extends Model {
...
   getRightDock () {
     return this.paneContainers.right
   }
 
+  // Essential: Get the {Dock} below the editor window.
   getBottomDock () {
     return this.paneContainers.bottom
   }
 
+  // Essential: Get an {Array} of the current {WorkspaceCenter} and {Dock}s.
```
jasonrudolph: Are we sure we want to make this method part of the public API?

I personally find this method's behavior a bit surprising. Specifically, I find it odd that this method returns a heterogenous collection of objects in which the objects have no explicit shared interface. I first ran into this in https://github.com/atom/atom/pull/14460#issue-229489706:

> It seems like there's an interface that we expect both `Dock` and `WorkspaceCenter` to implement, but I don't see that interface explicitly described or enforced anywhere. I _think_ that lack of explicit interface is a cause of the bug seen in #14442. There may be other bugs that are lurking due to code that implicitly expects a `WorkspaceCenter` and a `Dock` to provide a common interface.

Without an explicit interface that is shared by `WorkspaceCenter` and `Dock`, anyone using `Workspace::getPaneContainers()` will likely have to first check the type of each object in the collection before invoking any methods on that object.

If that sounds as odd to you as it does to me, I recommend excluding this method from the public API for now.

50Wliu: :+1: sounds good to me.  And if any package authors do want an array, they can just populate it themselves using the above four methods.

jasonrudolph (approved):
This looks good. Thanks for taking the time to make these updates! :bow:

PR #14734, opened 2017-06-07
Don't try to measure lines that don't exist

as-cii:
By the time that the animation frame is delivered, the requested autoscroll position could not exist anymore. This could cause the editor component to measure a non-existent line and, as a result, throw an exception.

With this commit we will always ignore measurements for screen lines that do not exist.

/cc: @nathansobo


nathansobo:
👍 Looks good to me.

PR #14736, opened 2017-06-07
Add methods for observing dock visibility

matthewwithanm:
This adds an `onDidChangeVisible()` and `observeVisible()` method to docks so that other objects can be notified when they're toggled.

Since this functionality covers the `didHide()` callback use case, I removed that.

One thing I'd like to discuss is the semantics of this. In the original docks PR, I had avoided calling this "visible" because it'll actually be false while the dock is visible but being animated closed. If you're using this API, for example, to show or hide contents of an item, you could wind up with surprising results.

### DIFF
--- a/spec/dock-spec.js
+++ b/spec/dock-spec.js
@@ -9,12 +9,15 @@ describe('Dock', () => {
     it('opens the dock and activates its active pane', () => {
       jasmine.attachToDOM(atom.workspace.getElement())
       const dock = atom.workspace.getLeftDock()
+      const didChangeVisibleSpy = jasmine.createSpy()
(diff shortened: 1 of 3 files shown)


matthewwithanm:
For posterity: the conclusion after talking to @maxbrunsfeld was that this API should behave as-is. A later PR can toggle the display of the dock item containers before and after animation so that IntersectionObserver can be used. (Currently, [it's not working with "overflow: hidden" masking][1].)

[1]: https://codepen.io/matthewwithanm/pen/LLpypJ

maxbrunsfeld:
:+1:

Issue #11799, opened 2016-05-20
Configure verticalScrollMargin through Settings


hypnoglow:
**tldr**: [verticalScrollMargin](https://github.com/atom/atom/blob/73525a1bcc309ff64c6f930401cedc052ddcf408/src/text-editor.coffee#L71) should be configurable through Settings (`config.cson`).

I had a hard time searching for vim-like [scrolloff](http://vim.wikia.com/wiki/Make_search_results_appear_in_the_middle_of_the_screen#Scrolloff_option) feature in Atom.

Finally, I figured out that for this you have to write something like this in your **init.coffee**:

```
atom.workspace.observeTextEditors (editor) =>
    editor.setVerticalScrollMargin 8
```

There is even a [package](https://atom.io/packages/atom-scrolloff) where they tried to implement this feature because users experience difficulties trying to configure this.

PR #14738, opened 2017-06-08
Add verticalScrollMargin to configurable settings.

daniel-p-gonzalez:
### Description of the Change
Vertical scroll margin was already a feature, but it was hard coded to 2. This PR adds it as a configurable setting.
Primary changes that had to be made:
src/text-editor.coffee: Added verticalScrollMargin to constructor/copy/update.
src/config-schema.js: Added verticalScrollMargin as an editor property with default of 2 and a description of the feature.
src/text-editor-registry.js: Added verticalScrollMargin editor param.
spec/text-editor-spec.coffee: Added tests for the feature.

### Alternate Designs

There is a workaround mentioned in issue [#11799](https://github.com/atom/atom/issues/11799), but this PR simply adds it as a configurable setting.

It's an existing feature in core, but its value is hard coded. In other editors this is usually either a configurable feature or non-existent (i.e. 0 margin).

### Benefits

People who are used to the lack of a vertical scroll margin in other editors, or who wish to customize it, can now easily modify the vertical scroll margin size in the settings.

### Possible Drawbacks

None that I'm aware of.

### Applicable Issues

[#11799](https://github.com/atom/atom/issues/11799)

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -54,6 +54,7 @@ describe "TextEditor", ->
         invisibles: {space: 'S'}
         showInvisibles: true
         editorWidthInChars: 120
+        verticalScrollMargin: 10
       })
 
       # Force buffer and display layer to be deserialized as well, rather than
@@ -73,6 +74,7 @@ describe "TextEditor", ->
       expect(editor2.getSoftWrapHangingIndentLength()).toBe(editor.getSoftWrapHangingIndentLength())
       expect(editor2.getInvisibles()).toEqual(editor.getInvisibles())
       expect(editor2.getEditorWidthInChars()).toBe(editor.getEditorWidthInChars())
+      expect(editor2.getVerticalScrollMargin()).toBe(editor.getVerticalScrollMargin())
       expect(editor2.displayLayer.tabLength).toBe(editor2.getTabLength())
 
   describe "when the editor is constructed with the largeFileMode option set to true", ->
@@ -145,7 +147,7 @@ describe "TextEditor", ->
       returnedPromise = editor.update({
         tabLength: 6, softTabs: false, softWrapped: true, editorWidthInChars: 40,
         showInvisibles: false, mini: false, lineNumberGutterVisible: false, scrollPastEnd: true,
-        autoHeight: false
+        autoHeight: false, verticalScrollMargin: 12
       })
 
       expect(returnedPromise).toBe(element.component.getNextUpdatePromise())
@@ -159,6 +161,7 @@ describe "TextEditor", ->
       expect(editor.isLineNumberGutterVisible()).toBe(false)
       expect(editor.getScrollPastEnd()).toBe(true)
       expect(editor.getAutoHeight()).toBe(false)
+      expect(editor.getVerticalScrollMargin()).toBe(12)
 
   describe "title", ->
     describe ".getTitle()", ->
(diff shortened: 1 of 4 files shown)


spec/text-editor-spec.coffee
```diff
@@ -5924,3 +5927,58 @@ describe "TextEditor", ->
...
+
+      component = element.getComponent()
+      # for a height of 120px and a line height of 17px, this will be 7 rows
+      rowCount = component.getLastVisibleRow()
+
+      for scrollMargin in [0..4]
+        editor.update({verticalScrollMargin: scrollMargin})
+
+        # margin is clamped so that it is no larger than half the screen
+        maxScrollMargin = Math.floor(((component.getScrollContainerClientHeight() / editor.getLineHeightInPixels()) - 1) / 2)
```
lee-dohm: It looks like you're preventing a potential error condition in the tests. I assume this clamping is to prevent the editor getting confused when the scroll margin is larger than the distances to both edges simultaneously. If this can cause problems, this should be prevented by the code and there should be a test to ensure that the feature to prevent the confusion works.

FBMachine: Hi Lee,
This clamping is actually for the right hand side of the expect(...).toBe test below, i.e. it's needed for computing the expected value for the feature. The feature itself implements this behavior, so this test ensure that it matches that behavior.

lee-dohm: I feel like these tests are pretty complicated and would therefore be brittle. It looks like you're trying to test multiple related scenarios in each test. Can you break out each scenario into its own test?

FBMachine: Sure thing. Could you elaborate a little on what conditions you think are best as separate tests? There's only one actual expect().toBe in each test, which tests if the top/bottom of the window is where it is expected given the size of the vertical scroll margin. I'm not sure how I could test the clamping separately, since the test would fail if it weren't present.

Do you maybe mean separately testing the different settings of scrollMargin and row positions in the for loops? So e.g., a hard coded test for scrolling down to line 6, with a margin size of 3, and a window size of 7 (so the bottom would be expected to be line 9). In this situation, it would be possible to test clamping only in scenarios where the clamping is necessary. I thought it would be a more exhaustive test to iterate over a range of values and handle under/overflow for those cases that it does happen, but if you think it would be clearer doing individual tests that works too. The scenario where clamping is required will still require this more complicated clamping test in order to be a correct test of the functionality though.

Or maybe you had something else in mind? Let me know and I will gladly update the tests.

lee-dohm: I would have the following tests:

1. Set the margin to 0, scroll to the top ... verify there is no margin
1. Set the margin to 0, scroll to the bottom ... verify there is no margin
1. Set the margin to [value that is > 0 but < half the screen height], scroll to the top ... verify the appropriate amount of margin
1. Set the margin to [value that is > 0 but < half the screen height], scroll to the bottom ... verify the appropriate amount of margin
1. Set the margin to [value that is > half the screen height], scroll to the top ... verify the appropriately clamped amount of margin
1. Set the margin to [value that is > half the screen height], scroll to the bottom ... verify the appropriately clamped amount of margin

lee-dohm: :point_up: That set of tests makes more clear what the intent of the behavior of the code is also, which makes maintaining the tests in the future much easier.

FBMachine: Thanks, I agree that would be clearer.

sadick254:
Hey @FBMachine. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

PR #14753, opened 2017-06-10
Improved clarity of Editor not responding actions

tjkohli:
### Description of the Change

Changed the label on the positive button of the "Editor is not responding" alert dialog from "Close" to "Force Close" because "Close" might imply closing the Alert/Dialog instead of the editor, whereas "Force Close" better describes the action that button will take.

### Alternate Designs

Other strings may be "Force Quit" or "End Process" or "Close Editor"

This change is a tweak to the core strings.

### Benefits

Makes a (minor) improvement to user experience and reduces decision fatigue for end users. 

### Possible Drawbacks

Slight (neglibile) increase in file size due to longer string.

### Applicable Issues

Translation/localization will need to be updated elsewhere.

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -160,7 +160,7 @@ class AtomWindow
 
       chosen = dialog.showMessageBox @browserWindow,
         type: 'warning'
-        buttons: ['Close', 'Keep Waiting']
+        buttons: ['Force Close', 'Keep Waiting']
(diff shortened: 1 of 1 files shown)


50Wliu:
Gets a :+1: from me.  Anyone from @atom/feedback opposed?

lee-dohm:
Sounds like a great change to me! Thanks for the contribution @tjkohli :tada:

PR #14761, opened 2017-06-12
Update CSP to allow using eval from JavaScript

jasonrudolph:
### Description of the Change

While attempting to interact with an Express.js server within some Atom tests, we ran into the following error related to Atom's current Content Security Policy:

```
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".

  at Function.wrapfunction [as function] (node_modules/depd/index.js:409:43)
  at populateConstructorExports (node_modules/http-errors/index.js:247:45)
  at Object.<anonymous> (node_modules/http-errors/index.js:29:1)
  at Object.<anonymous> (node_modules/http-errors/index.js:262:3)
```

This PR updates the policy to resolve this error ☝️, allowing Atom to run [this code](https://user-images.githubusercontent.com/2988/27043798-58b9ed04-4f69-11e7-80e8-3db5f8336cd5.png).

### Alternate Designs

We could have changed Express.js's dependencies to stop using eval, but we are not aware of any advantages of making such a change upstream.

N/A

### Benefits

Allow package authors to use libraries that rely on eval.

### Possible Drawbacks

We're not aware of any negative consequences that this change would introduce, but it _is_ a change to the CSP, so we definitely want to hear from others. :pray:

@atom/maintainers: Do you know of any negative consequences that would result from this change?

### Applicable Issues

N/A

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <meta http-equiv="Content-Security-Policy" content="default-src * atom://*; img-src blob: data: * atom://*; script-src 'self'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * atom://*;">
+  <meta http-equiv="Content-Security-Policy" content="default-src * atom://*; img-src blob: data: * atom://*; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * atom://*;">
   <script src="index.js"></script>
 </head>
 <body tabindex="-1">


lgeiger:
Thanks for bringing this up 👍 

This would be a very welcome change for [Hydrogen](https://github.com/nteract/hydrogen/) since it would allow us to support a lot more output types, e.g. rich visualisations rendered by [Vega](https://vega.github.io/vega-lite/). Currently this isn't possible, even when using [loophole](https://github.com/atom/loophole).

lgeiger (approved):

UziTech:
I ran into this [issue](https://github.com/UziTech/atom-jasmine2-test-runner/issues/14) with [atom-jasmine2-test-runner](https://github.com/UziTech/atom-jasmine2-test-runner/) when Jasmine 2.6.0 came out [using eval to create it's spies](https://github.com/jasmine/jasmine/pull/1055).

We fixed it by [using hard-coded functions](https://github.com/jasmine/jasmine/pull/1330#issuecomment-297235767) instead of eval, but I agree there seems to be no benefit to not allowing eval.

iolsen (approved):
I'm 👍 on this change.

lgeiger:
Very cool! Thanks for merging 🎉 

Since this is a very small code change, may I suggest to back-port df7f72a3b75ef486e08c8e6fe17da766c5095418 to Atom 1.19-beta?

lee-dohm:
While it is a very small code change, it _could_ have far-reaching implications. Though the final decision is not up to me, I would feel more comfortable if we didn't rush this out the door.

lgeiger:
> While it is a very small code change, it could have far-reaching implications. Though the final decision is not up to me, I would feel more comfortable if we didn't rush this out the door.

@lee-dohm Thanks for the answer, I think it's very unlikely that this introduces bugs since it only relaxes CSP.
But I understand you don't want to rush this out the door, since there may be security concerns that we haven't thought of yet.

lee-dohm:
>  I think it's very unlikely that this introduces bugs since it only relaxes CSP.

As someone who spent eight years as a tester, you have no idea how often I heard this just before :boom: :laughing:

lgeiger:
> As someone who spent eight years as a tester, you have no idea how often I heard this just before 💥 😆

😂  No rush

PR #14769, opened 2017-06-13
Prevent block decorations from mistakenly wrapping during measurements

### LINKED ISSUE #14743: [v1.19] adding blockDecoration makes editor rendering odd
block-decoration-issue-on-v1 9

### Prerequisites

### Description

Block decoration occupy two screen row, and after I render block decoration, cursor, line-gutter shows incorrect place.

Tried last build on #14435, (commit: https://github.com/atom/atom/pull/14435/commits/a5f4472f5d422d6185c6952ae2c3a6cef68eec1d) which is last commit before merged to master.

- [Download link](https://4126-3228505-gh.circle-artifacts.com/0/Users/distiller/atom/out/atom-mac.zip)
- version

```
Atom    : 1.19.0-dev-a5f4472
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Steps to Reproduce

I can reproduce with **safe-mode**

1. open editor, insert three lines of text whatever (`row0\nrow1\nrow3`)
2. open chrome dev-tools and paste following js code.

```javascript
(function reproduceIssue() {
  const cursorRow = editor.getCursorBufferPosition().row
  const item = document.createElement("span")
  item.innerText = "[block decoration]"
  const marker = editor.markBufferPosition([cursorRow, 0])
  editor.decorateMarker(marker, { type: "block", item: item })
})()
```

3. See **line-gutter**, insert text to row0, row1, row2 and see cursor is shown in incorrect screen position.

can edit normal way
block decoration occupy one screen row as like in previous release.

rendering become odd.

always

### Versions

### Additional Information

My [narrow](https://atom.io/packages/narrow) package directly affected this issue.
Here is gif with same-build-Atom.

The first line's buttons are items in single block decoration.
The cursor is row 0 in previous release, but with latest-build, cursor and lines render incorrectly because of this issue.

block-decoration-issue-on-v1 9-2

as-cii:
Fixes #14743.

Before rendering block decorations, we read their heights by putting them into a special div called `blockDecorationMeasurementsArea`.

Previously, this div was not explicitly sized, which was causing decorations to wrap while being measured but not when actually rendering them.

This pull-request fixes this inconsistency by explicitly styling the measurement area so that it has the same width as the component scroll width.

/cc: @atom/maintainers @t9md

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2035,6 +2035,15 @@ describe('TextEditorComponent', () => {
       expect(item6.previousSibling).toBe(lineNodeForScreenRow(component, 12))
     })
 
+    it('bases the width of the block decoration measurement area on the editor scroll width', async () => {
+      const {component, element} = buildComponent({autoHeight: false, width: 150})
+      expect(component.refs.blockDecorationMeasurementArea.offsetWidth).toBe(component.getScrollWidth())
+
+      element.style.width = '800px'
+      await component.getNextUpdatePromise()
+      expect(component.refs.blockDecorationMeasurementArea.offsetWidth).toBe(component.getScrollWidth())
+    })
+
     function createBlockDecorationAtScreenRow(editor, screenRow, {height, margin, position}) {
       const marker = editor.markScreenPosition([screenRow, 0], {invalidate: 'never'})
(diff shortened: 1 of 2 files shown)


nathansobo:
❤️

PR #14782, opened 2017-06-14
Don't recreate Atom shortcut on Desktop on update

### LINKED ISSUE #5323: Atom keeps adding desktop shortcut
Atom keeps recreating its desktop shortcut unexpectedly. I'm guessing it's something involved in the automatic update process, since it doesn't happen all the time. Creating a desktop shortcut on initial installation makes sense, but doing it every time a background update occurs does not.

I'm using Atom 0.176.0 on Windows 8.1 x64.

anaisbetts:
This PR updates the Squirrel hooks to tell Squirrel.Windows directly what kinds of shortcuts to create on update instead of creating them then deleting after-the-fact

Fixes #5323

## TODO:

- [x] Make sure this works


iolsen:
@damieng can you give this a quick look and merge?

PR #14783, opened 2017-06-15
[RFC] Set `display: none` on hidden docks

matthewwithanm:
Sometimes, it's useful to be able to tell if your item is currently visible to the user or not. This can be used to back off of work, for example.

After talking to @nathansobo and @maxbrunsfeld, we decided the best way forward was to be able to use IntersectionObserver for this…but there are some issues.

The first is that—in its current iteration, at least—it doesn't seem to observe when visibility changes as a result of being masked by an `overflow: hidden` parent, which is how docks do it. (See [pen][1].)

To work around this, I put together this commit which toggles the `display` of the wrapper element when the dock is shown and completely masked. This guarantees that the IntersectionObserver will broadcast notifications at the appropriate time. However, this too has an issue: the `intersectionRatio` property (which I'd assume people would use to determine if the element was visible) is `0` at the beginning of the show animation. This kinda makes sense because, well, yeah, it isn't actually shown at all yet at the point when the animation to show it starts. I'm assuming that this is why the [`isIntersecting`][2] property was added (but I'm a n00b wrt the API so maybe not). After some playing around, though, I discovered that there was a way to distinguish between the visible and hidden states as, in the case of the item being shown, the `width + height` was nonzero.

So I wrote up some tests to verify the behavior and they pass…

but then I ran my tests without my dock changes and they still pass. That would seem to imply that the `overflow: hidden` masking _is_ actually triggering an IntersectionObserver notification?? 😖 (It doesn't if you observe the tree view and toggle the dock.)

@nathansobo @as-cii Any guidance you could give about this inconsistent IntersectionObserver behavior would be super appreciated.

[1]: https://codepen.io/matthewwithanm/pen/LLpypJ
[2]: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/isIntersecting

### DIFF
--- a/spec/dock-spec.js
+++ b/spec/dock-spec.js
@@ -2,7 +2,7 @@
 
 const Grim = require('grim')
 
-import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
+import {it, fit, ffit, fffit, beforeEach, afterEach, timeoutPromise} from './async-spec-helpers'
 
 describe('Dock', () => {
   describe('when a dock is activated', () => {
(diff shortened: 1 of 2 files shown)


nathansobo:
Hey, sorry, this has been sitting in my inbox for way too long. I honestly don't have a ton of experience with `IntersectionObserver`. My *only* use of it is in the text editor where I don't rely on any of the data on the event. I simply use it as an opportunity to check that visibility has changed by inspecting `offsetHeight` and `offsetWidth`. It sounds like that might be useful in your case as well, so maybe these tests should simply indicate that an intersection event was fired without making claims about its contents.

Anyway, if you're still interested in getting this in, I am down to keep interacting on it. First step would be a green build.

sadick254:
Hey @matthewwithanm. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

PR #14788, opened 2017-06-15
AppVeyor node_module caches get corrupted, let's just disable them for now

anaisbetts:
This PR removes the node_module caches so builds will complete, even if they're slower

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -58,10 +58,6 @@ artifacts:
     name: atom-full.nupkg
 
 cache:
-  - '%APPVEYOR_BUILD_FOLDER%\script\node_modules'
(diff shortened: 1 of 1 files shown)


Stanzilla:
Did you try installing a 4.x version of npm instead? @5 is super buggy atm and could cause the corruption.

50Wliu:
Good point.  Why _do_ we `npm install -g npm`?

Stanzilla:
@iolsen are you sure this was a good idea?

iolsen:
It's certainly not a long-term fix but we've been manually clearing the cache multiple times a day recently. Hopefully this relieves that pain short-term until we figure out the root cause.

PR #14795, opened 2017-06-16
Always run tests where directed

winstliu:
### Requirements

### Description of the Change

#13947 introduced a regression with `atom --test` where if `core.restorePreviousWindowsOnStart` was set to `always`, `atom --test` would ignore the supplied directory and instead test specs in the first folder you opened.  Had a very fun time trying to figure out why `atom --test spec` was running `language-html` tests instead of `tree-view` ones...

### Alternate Designs

None!

Because I would like to be able to tell Atom where I want the specs to be run!

### Benefits

Being able to test any directory from the command line regardless of where Atom was started!

### Possible Drawbacks

None?!

### Applicable Issues

None!

### DIFF
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -120,7 +120,9 @@ class AtomApplication
     Promise.all(windowsClosePromises).then(=> @disposable.dispose())
 
   launch: (options) ->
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for this @50Wliu :bow:

PR #14837, opened 2017-06-20
:bug: Don't prefix resource path if package path is already absolute

matthewwithanm:
We're running into an issue in Nuclide where sometimes `finishLoading()` is called twice on the `tree-view` package. As a result, the path gets prepended twice.

This doesn't fix the underlying bug, but I think it's still best to use `resolve()` here so that, if `@path` is absolute, the resource path won't be prepended multiple times.

### DIFF
--- a/src/package.coffee
+++ b/src/package.coffee
@@ -96,7 +96,7 @@ class Package
 
   finishLoading: ->
     @measure 'loadTime', =>
-      @path = path.join(@packageManager.resourcePath, @path)
+      @path = path.resolve(@packageManager.resourcePath, @path)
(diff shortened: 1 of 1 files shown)


sadick254 (changes requested):
@matthewmccullough Thank you for your contribution. Atom has gone through some changes in recent years. We have moved from using `coffee-script` to `javascript`. This move makes this PR outdated. Do you mind updating it by making changes to `package.js` file?

matthewmccullough:
> This move makes this PR outdated

No problem. However, I think I'll need to abandon the work for the time being.

PR #14844, opened 2017-06-20
Remeasure block decorations if content changes

### LINKED ISSUE #14836: [1.19-beta] Block decorations don't remeasure if content changes
### Prerequisites

### Description

First of all congratulations on the awesome improvements in the latest Atom releases, that's a great leap forward 💯 

Though it looks like a bug in the new editor rendering rewrite (#13880) slipped through regarding block decoration rendering.

The height of a block decorations doesn't get recalculated anymore if the content changes.

### Steps to Reproduce

1. Place cursor in the middle of a file
2. Execute the following code in the console and observe odd rendering:
```js
const item = document.createElement("span")
const cursorRow = editor.getCursorBufferPosition().row
const marker = editor.markBufferPosition([cursorRow, 0])
editor.decorateMarker(marker, { type: "block", item: item })
setTimeout(() => {
  item.innerText = "[block decoration]\n[block decoration]\n[block decoration]\n[block decoration]"
}, 100)
```
3. Resize window to see layout recalculation and correct rendering

**Expected behavior (Atom 1.18):**

**Actual behavior (Atom 1.19 and `master`):**
block

### Versions

macOS
```
Atom    : 1.19.0-beta0
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```
and
```
Atom    : 1.20.0-dev-ccdae0fc8
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Additional Information

This bug is especially critical for [Hydrogen](https://github.com/nteract/hydrogen/) since we render live results from a running kernel inside a block decoration, that can change dimensions over time.

lgeiger:
### Requirements

### Description of the Change

Since #13880 the height of a block decorations doesn't get recalculated anymore if the content changes. This brings back the behavior of Atom 1.18. For more information on how to reproduce and GIFs see #14836

### Alternate Designs

I considered adding mutation observers to `decoration-manager` but this way we have a finer control over which decoration to schedule for remeasurment.

This fixes a :bug: 😄 

### Benefits

This makes block decoration rendering consistent with the way it was in Atom 1.18

### Possible Drawbacks

I don't know how performance critical multiple `MutationObserver`'s are.

### Applicable Issues

Fixes #14836

/cc: @nathansobo @as-cii

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1941,8 +1941,6 @@ describe('TextEditorComponent', () => {
       item3.style.margin = '10px'
       item2.style.height = '33px'
       item2.style.margin = '0px'
-      component.invalidateBlockDecorationDimensions(decoration2)
-      component.invalidateBlockDecorationDimensions(decoration3)
       await component.getNextUpdatePromise()
       expect(component.getRenderedStartRow()).toBe(0)
       expect(component.getRenderedEndRow()).toBe(9)
@@ -1973,7 +1971,6 @@ describe('TextEditorComponent', () => {
       item3.style.wordWrap = 'break-word'
       const contentWidthInCharacters = Math.floor(component.getScrollContainerClientWidth() / component.getBaseCharacterWidth())
       item3.textContent = 'x'.repeat(contentWidthInCharacters * 2)
-      component.invalidateBlockDecorationDimensions(decoration3)
       await component.getNextUpdatePromise()
 
       // make the editor wider, so that the decoration doesn't wrap anymore.
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1,4 +1,4 @@
-/* global ResizeObserver */
+/* global ResizeObserver, MutationObserver */
 
 const etch = require('etch')
 const {Point, Range} = require('text-buffer')
@@ -2347,6 +2347,17 @@ class TextEditorComponent {
 
     this.blockDecorationsToMeasure.add(decoration)
 
+    let mutationObserver
+    if (decoration.properties.item) {
+      mutationObserver = new MutationObserver(() => {
+        this.invalidateBlockDecorationDimensions(decoration)
+      })
+
+      mutationObserver.observe(decoration.properties.item, {
+        attributes: true, childList: true, characterData: true, subtree: true
+      })
+    }
+
     const didUpdateDisposable = marker.bufferMarker.onDidChange((e) => {
       if (!e.textChanged) {
         this.lineTopIndex.moveBlock(decoration, marker.getHeadScreenPosition().row)
@@ -2358,6 +2369,7 @@ class TextEditorComponent {
       this.lineTopIndex.removeBlock(decoration)
       didUpdateDisposable.dispose()
       didDestroyDisposable.dispose()
+      if (mutationObserver) mutationObserver.disconnect()
       this.scheduleUpdate()
     })
   }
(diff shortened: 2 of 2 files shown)


as-cii:
Thank you for the pull-request, @lgeiger! 🙇 

The reason why we decided not to automatically remeasure items when they change can essentially be attributed to performance, because we were unclear on the cost of observing a potentially large number of DOM nodes (e.g. see https://github.com/atom/atom/pull/13795).

On the other hand, users of this API should know in advance when they are about to change a decoration and, as such, can manually invalidate its dimensions and let Atom re-measure it. I am honestly quite surprised that it used to work in Atom 1.18, as I don't remember adding any code that supported this behavior.

With that in mind, does using `invalidateBlockDecorationDimensions` work for your use case? If not, can you describe a scenario where it wouldn't be applicable?

Thanks!

lgeiger:
Thanks for your answer, I understand the performance considerations.

I didn't know of `invalidateBlockDecorationDimensions` since it isn't [documented anywhere](https://atom.io/docs/api/search/v1.18.0?utf8=%E2%9C%93&q=invalidateBlockDecorationDimensions) (in fact the entire `TextEditorElement` class isn't documented) and it used to work in Atom 1.18, so I guess package Authors aren't aware of that change. It might be good to mention it in the release notes, since this may break other packages as well.

This change requires quite some refactorings in [Hydrogen](https://github.com/nteract/hydrogen/), but you're right we can use Reacts `onDidUpdate` handler to schedule a recalculation and don't need to rely on `MutationObservers`.
I'll try it in the next few days to see if this solution is applicable.

lgeiger:
I tried to manually invalidate the block decorations in Hydrogen (We use React so I simply made us of `onDidUpdate`).

While this works for most of the cases, some outputs may allow users to dynamically change DOM dimensions. Here is a example where this is the case:
modifications

If this PR isn't accepted we'll still need to use `MutationsObservers` inside the package code base, at least I don't know of any better way of doing this. I can imagine this beeing a problem that's also relevant for other packages.

@as-cii What about adding another `decorationParam` for configuring if a block decoration should be automatically invalidated if the dimensions change?

as-cii:
Thanks for providing this scenario @lgeiger: now I am definitely on board with remeasuring block decorations when their size changes. I think we could take a slightly different approach, which should also be more efficient. In particular:

1. Add a `ResizeObserver` to on-screen block decorations. When a block decoration goes out of screen we should stop observing it. Viceversa, when a block decoration appears on screen we should register a new `ResizeObserver`. 
    * One thing to experiment with here is to reuse the same `ResizeObserver` and simply call `observe`/`unobserve`, as opposed to creating a new observer for each decoration.
2. When getting a resize callback, call `invalidateBlockDecorationDimensions`.
    * Another thing that might be worth experimenting with here is to avoid re-measuring the block decoration's item through `invalidateBlockDecorationDimensions` and, instead, simply update `lineTopIndex` with the values provided by the `ResizeObserver` callback.

I can provide more guidance on this if you are interested in changing this pull-request to do what I described above! 👍

lgeiger:
@as-cii Thanks for the great suggestions!

I think point 2 should be fairly easy to implement. Using a global resize observer may require more code change, I'll take a look to see if I can make it work. Thanks for your help 👍

lgeiger:
I tried using a global `ResizeObserver` in ecd24ade6f60239ed23665957e2c69971ecd9891. 

Unfortunately calling `this.scheduleUpdate()` inside the observer callback causes a re-measurement and therefor continuously calls the callback until the block decoration is remove or moved off screen.

I tried using `this.updateSyncAfterMeasuringContent()` as well which is better but the callback still gets called far too many times (e.g. on scroll, editor resize or mouseover events).

@as-cii Any suggestions?

lgeiger:
Closing in favor of #14929

PR #14854, opened 2017-06-20
Update .python-version & Added Webupd8 Team PPA for 32-bit Linux Users

ghost:
### Description of the Change

I changed the [README.md](README.md) file so 32 bit users can be able to know they can use Atom. I know this isn't your repository, but it works :)

And I changed the [.python-version](.python-version) so it has the Python 2 _and_ the Python 3 versions listed, for all the Python developers out there. 

### Alternate Designs

I think the [README.md](README.md) was OK... but the Webupd8 Team repository is more... vast for the 32 bit Ubuntu users out there. And the [.python-version](.python-version) file was only for the Python 2 developers (which Python 2 is old anyways), and I helped by adding the Python 3 version as well.

I think it should be in core because it helps the 32 bit Ubuntu users and it'll also help the Python 3 developers.

### Benefits

32 bit Ubuntu users and Python 3 developers will be able to use Atom!

### Possible Drawbacks

Webupd8 Team's repository is unofficial, and the code hasn't been checked by me. But, you can check it out [HERE](https://launchpad.net/~webupd8team/+archive/ubuntu/atom)

And I have no idea what programming language .python-version is, or if it's a programming language at all. I did my best, but if you see anything bad just tell me.


50Wliu:
Hi @JBoyAdvance, thanks for the PR!  Unfortunately, we aren't able to accept this because of the following reasons:
* As the Atom team has no control over the WebUpd8 repository, we aren't able to verify its integrity and don't feel comfortable listing it as an option on the Atom repo.
* `node-gyp`, one of the tools required to build Atom, only works with Python 2.x.  That's why our `.python-version` is set to 2.7.13.

Thanks again for your interest in Atom :grinning:.

PR #14884, opened 2017-06-24
Return true from Pane#destroyItem if item is destroyed

kuychaco:
This PR ensures that we return the correct boolean value from `Pane.prototype.destroyItem` when the `force` parameter is `true` or the item should not prompt to save. 

Without this, logic such as the following 
```
const destroyed = await pane.destroyItem(item); // returns the value of `item.destroy && item.destroy()` which can be undefined or null
if (!destroyed) {
  pane.activateItem(item); // called on an item that has been destroyed
}
```
... produces an uncaught error error ...
```
Uncaught (in promise) Error: Adding a pane item with URI '/some/path/to/item' that has already been destroyed
```

/cc @maxbrunsfeld for relevant change made in https://github.com/atom/atom/commit/dc320181fc852f3174411e107bfd2a02f2c504ad#diff-1a7d1dda35d8e7ed3ad35b9a9fd2f134L615

NOTE: Fix needs to be cherry-picked to beta on the `1.9-releases` branch in order for [`atom/github` CI](https://ci.appveyor.com/project/Atom/github/branch/master/job/igsd6yot0wjimv48/tests) to pass


maxbrunsfeld:
👍  Thanks for fixing this @kuychaco!

kuychaco:
Somehow this got missed when https://github.com/atom/github/pull/980 got merged, and is now superseded by #15264 so I'll go ahead and close it

Issue #14842, opened 2017-06-20
[1.19.0-beta1] Block decorations mess up line numbering on scrolling


lgeiger:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

If multiple block decorations are shown, line numbers get into an inconsistent state after scrolling

### Steps to Reproduce

1. Open file with a few lines of code
2. Execute the following code in dev tools and observe correct line number callculation:
```js
function createBlock(row) {
  const item = document.createElement("span");
  item.innerText = "[block decoration]\n[block decoration]\n[block decoration]";
  const editor = atom.workspace.getActiveTextEditor();
  const marker = editor.markBufferPosition([row, 0]);
  editor.decorateMarker(marker, { type: "block", item: item });
}

createBlock(1);
createBlock(5);
createBlock(8);
createBlock(10);
```
3. Scroll and observe inconsistent line numbers

**Expected behavior (Atom 1.18):** Line numbers are preserved by block decorations during scrolling

**Actual behavior (Atom 1.19-beta1):**
![line-number](https://user-images.githubusercontent.com/13285808/27329042-6e4bf9cc-55b4-11e7-92e1-5ff61c55029e.gif)

**Reproduces how often:** 100%
### Versions

macOS
```bash
Atom    : 1.19.0-beta1
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Additional Information

This bug is especially critical for [Hydrogen](https://github.com/nteract/hydrogen/) since we render live results from a running kernel inside a block decoration, that can change dimensions over time.

Possibly related: #14817

PR #14893, opened 2017-06-26
Fix resetting styles in NodePool

as-cii:
Fixes https://github.com/atom/atom/issues/14842.

Previously, we were mistakenly not clearing out some styling properties like `marginTop`, thus causing e.g. line numbers to be misaligned. This was caused by manual updates to an element's style object, without a consequent update to the `NodePool`.

With this pull-request we will now rely on `element.styleMap` (a DOM primitive) to detect which styles have been set on a element that is about to be recycled and, if the object being created doesn't use them, simply clear them out.

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -4034,7 +4034,6 @@ class NodePool {
   constructor () {
     this.elementsByType = {}
     this.textNodes = []
-    this.stylesByNode = new WeakMap()
   }
 
   getElement (type, className, style) {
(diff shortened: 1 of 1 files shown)


as-cii:
@atom/core: I think this should go into the next hotfix, as it's causing odd rendering issues like https://github.com/atom/atom/issues/14842.

as-cii:
Build failures seem unrelated, so I'll go ahead and merge this.

avindra:
Hm, this didn't fix 

https://github.com/justmoon/relative-numbers/issues/41

Do you think this is related?

PR #14905, opened 2017-06-27
Only expect a single 'did-destroy' event

matthewwithanm:
These events will only be fired a single time at most, so we should clean up the listeners after that.

This should help minimize accidental memory leaks.

cc @hansonw


maxbrunsfeld (approved):
OMG. Great idea for preventing accidental leaks.

BinaryMuse:
❤️

PR #14929, opened 2017-06-30
Invalidate block decoration dimensions automatically when their size changes

as-cii:
Supersedes https://github.com/atom/atom/pull/14844.
Fixes https://github.com/atom/atom/issues/14836.

In the previous implementation of the editor rendering layer, we used to continuously re-measure block decorations that were on-screen. This was done to allow users to interact with the decorations, without requiring any extra work by package authors if the decorations size changed dynamically as a response to user input. During the rewrite in #13880 we lost this behavior and issues like the one linked above started arising. 

With this pull-request, we will now automatically schedule a new update when detecting that a block decoration got resized and call `invalidateBlockDecorationDimensions` on behalf of package authors. This is powered by a per-component `ResizeObserver` that listens for size changes on every added block decoration.

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -114,6 +114,9 @@ class TextEditorComponent {
     this.horizontalPositionsToMeasure = new Map() // Keys are rows with positions we want to measure, values are arrays of columns to measure
     this.horizontalPixelPositionsByScreenLineId = new Map() // Values are maps from column to horiontal pixel positions
     this.blockDecorationsToMeasure = new Set()
+    this.blockDecorationsByElement = new WeakMap()
+    this.heightsByBlockDecoration = new WeakMap()
+    this.blockDecorationResizeObserver = new ResizeObserver(this.didResizeBlockDecorations.bind(this))
     this.lineNodesByScreenLineId = new Map()
     this.textNodesByScreenLineId = new Map()
     this.overlayComponents = new Set()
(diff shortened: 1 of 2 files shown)


as-cii:
@atom/maintainers: I think this should go into the next beta hot fix.

@lgeiger: I tweaked your pull-request a little bit and tests are now green. I also did some manual testing which confirmed on-screen decorations are invalidated correctly. Could you confirm that this pull-request exhibits the same behavior as the latest stable version of Atom? Thanks so much for your help! ⚡️

src/text-editor-component.js
```diff
@@ -2361,13 +2368,29 @@ class TextEditorComponent {
...
 
+  didResizeBlockDecorations (entries) {
+    if (!this.visible) return
+
+    for (let i = 0; i < entries.length; i++) {
+      const {target, contentRect} = entries[i]
+      const decoration = this.blockDecorationsByElement.get(target)
+      const previousHeight = this.heightsByBlockDecoration.get(decoration)
+      if (this.element.contains(target) && contentRect.height !== previousHeight) {
+        this.invalidateBlockDecorationDimensions(decoration)
```
lgeiger: What about using the following?
```js
this.lineTopIndex.resizeBlock(decoration, contentRect.height)
this.heightsByBlockDecoration.set(decoration, contentRect.height)
this.scheduleUpdate()
```

This would remove an unnecessary re-measurement of the decoration.

lgeiger (approved):
@as-cii Thanks a lot for fixing this, really appreciate it!

It works excellent 👍 🎉 

>  I think this should go into the next beta hot fix.

👍 💯

src/text-editor-component.js
as-cii: Yeah, I considered doing that but the way we measure block decorations is slightly more complex than retrieving their `offsetHeight` (e.g. [we use "sentinel elements"](https://github.com/atom/atom/blob/10a7cb5badc56ac6b5d02476fe330d6e53f1cedd/src/text-editor-component.js#L285-L333)), so it felt more consistent to use the same method across all call sites.

lgeiger: Thanks for the explanation, makes sense!

PR #14931, opened 2017-06-30
Make atom.commands.dispatch return a promise that resolves after listeners complete

UziTech:
### Description of the Change

adds an `atom.commands.onDidFinish` function that is called after commands actually finish.

This will require commands that are asynchronous to return a promise so onDidFinish can know when a command is still working asynchronously.

### Alternate Designs

This comes from trying to return a commands value in #14720 

I decided to go with a `did-finish` event to keep backwards compatibility and use the event emitter already attached to `atom.commands`

This event is emitted once `Promise.all` on the return values of the command listeners finishes. Therefore it will be emitted when all listeners are resolved or one listener is rejected. And only the dispatched event is provided. A couple things could be changed here:

- We could wait until all listeners are resolved or rejected.
- We could emit a different event if rejected.
- We could include the returned values of the commands.

It requires changes to the command registry and package creators to return a promise if the command is async.

### Benefits

- Ability to tell when an async command actually finishes.
- Ability to dispatch commands in series.

### Possible Drawbacks

Package creators will need to come on board for this to be very usefull

### Applicable Issues

#13195

### DIFF
--- a/spec/command-registry-spec.js
+++ b/spec/command-registry-spec.js
@@ -208,10 +208,10 @@ describe("CommandRegistry", () => {
       expect(sequence[1][0]).toBe('listener');
       expect(sequence[2][0]).toBe('onDidDispatch');
 
-      waitsFor(() => sequence.length === 4), "onDidFinish never called");
+      waitsFor(() => sequence.length === 4, "onDidFinish never called");
 
       runs(() => {
-        expect(sequence[3][0]).toBe 'onDidFinish'
+        expect(sequence[3][0]).toBe('onDidFinish');
 
         expect(sequence[0][1] === sequence[1][1] && sequence[1][1] === sequence[2][1] && sequence[2][1] === sequence[3][1]).toBe(true);
         expect(sequence[0][1].constructor).toBe(CustomEvent);
--- a/spec/command-registry-spec.js
+++ b/spec/command-registry-spec.js
@@ -191,10 +191,8 @@ describe("CommandRegistry", () => {
       expect(calls).toEqual([]);
   });
 
-    it("invokes callbacks registered with ::onWillDispatch and ::onDidDispatch and ::onDidFinish", () => {
+    it("invokes callbacks registered with ::onWillDispatch and ::onDidDispatch", () => {
       const sequence = [];
-
-      registry.onDidFinish(event => sequence.push(['onDidFinish', event]));
 
       registry.onDidDispatch(event => sequence.push(['onDidDispatch', event]));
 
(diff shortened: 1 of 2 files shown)


UziTech:
@50Wliu Can i get this reviewed?

UziTech:
I rebased and changed to `command-registry.js` instead of `command-registry.coffee` (good move IMHO) and now the tests are failing but I don't think it has anything to do with my code.

UziTech:
@50Wliu any way we can get this in one of the next versions of Atom?

50Wliu:
Added it to the project board.

maxbrunsfeld:
Can you explain what this allows you to do that `atom.commands.onDidDispatch` doesn't already allow?

UziTech:
This allows for asynchronous commands to be run in series.

For example, maybe you have a large file that takes a couple seconds to format so you don't want to format it every time you save the file (<kbd>ctrl</kbd> + <kbd>s</kbd>) but you want to create a shortcut key that will format the file before  saving (<kbd>ctrl</kbd> + <kbd>shift</kbd> + <kbd>s</kbd>).

You can't call the save command until you know that the format command has completed not just dispatched.

maxbrunsfeld:
Commands are mostly intended to be run by the user via key bindings or menus.

In this case, could you have the package that does the formatting expose a service API, so that in your code you could just call a method to do the formatting instead of having to invoke a command?

UziTech:
Yes, just as you could have every package expose an emitter to let you know when their commands are dispatched, so we wouldn't need onDidDispatch.

onDidFinish would remove the burden from package authors to create a service just to allow other packages to know when an async command finishes. All they would have to do is return a promise, which, in my short research, seems already pretty common.

UziTech:
The code for my example would be:

```javascript
// in init.js
atom.commands.onDidFinish(event => {
  if (event.type === "atom-beautify:beautify-editor") {
    atom.commands.dispatch(event.target, "core:save");
  }
});
```

so much more simple and concise than subscribing to a service API

UziTech:
@maxbrunsfeld any update on this?

nathansobo:
Hey @UziTech sorry for the delay on this. We're dedicating all of January to trying to get a bunch of PRs across the finish line, so hopefully we can get you some resolution on this one.

I'm wondering: Would returning a promise from `atom.commands.dispatch` do the trick for your use case? We'd still need to return `null` if we didn't find any matches during dispatch to maintain backward compatibility, but since a promise is a truthy value we could get away with returning one. This seems like it could be a clearer API so you could tie the resolution of the promise back to a specific command that you dispatched. The rejection of any dispatched would also be handled pretty naturally in that case.

UziTech:
That is what I originally tried to do in #14720  but figured it would be easier to not break backwards compatibility if someone is checking `atom.commands.dispatch(...) === true`. I would prefer that `dispatch` returns a promise if there are any listeners.

The bigger problem that I see is, since a call to `dispatch` can dispatch multiple listeners, should we return a `Promise.all` over all the listeners which will reject if just one of them rejects or should we return an array of promises and let the user check for any rejections?

nathansobo:
>  it would be easier to not break backwards compatibility if someone is checking `atom.commands.dispatch(...) === true`

I'm cool breaking that case, because I think it's pretty unlikely that someone would do that.

> The bigger problem that I see is, since a call to dispatch can dispatch multiple listeners, should we return a Promise.all over all the listeners which will reject if just one of them rejects or should we return an array of promises and let the user check for any rejections?

Since multiple dispatches are uncommon, and one of them rejecting is even less common, I'd be okay returning `Promise.all`. Admittedly it's hard to project myself into the shoes of someone using this API and needing to deal with rejections individually, but it just seems like if there was a rejection in some unknown command handler, what are you possibly going to do to recover from that anyway.

UziTech:
I rebased and changed it so `atom.commands.dispatch` will just return a `Promise.all` around any matched listeners or null if no listeners are found.

nathansobo:
Looking good. Could you add a behavioral test that triggers a mix of commands that return promises and commands that don't to ensure this works as intended and does not regress?

50Wliu:
The easy way to resolve the timeouts is to call `jasmine.useRealClock()` before the setTimeouts.  I know that you can use `advanceClock(3)` to manually advance the timeout which is probably the better solution as the test won't be delayed by 3ms, but I don't have that much experience with it.  I suppose you'd place it before you await the values.

spec/command-registry-spec.js
```diff
@@ -357,12 +358,39 @@ describe("CommandRegistry", () => {
...
-      expect(registry.dispatch(parent, 'command')).toBe(false);
+      expect(registry.dispatch(grandchild, 'command').constructor.name).toBe("Promise");
+      expect(registry.dispatch(grandchild, 'bogus')).toBe(null);
+      expect(registry.dispatch(parent, 'command')).toBe(null);
+    });
+
+    it("returns a promise that resolves when the listeners resolve", async () => {
+      registry.add('.grandchild', 'command', () => 1);
+      registry.add('.grandchild', 'command', () => Promise.resolve(2));
+      registry.add('.grandchild', 'command', () => new Promise((resolve) => {
```
nathansobo: Why not just call `Promise.resolve` here?

UziTech: I'm trying to make sure the call to dispatch won't resolve before all promises are fulfilled in the future.

UziTech:
@50Wliu thanks I knew it was something like that but didn't have time to look for it before

UziTech:
Not sure why appveyor failed but everything else looks good

nathansobo:
Thanks for sticking with this @UziTech and rolling with my change requests. It will be great to support sequenced command dispatch. ⚡️

UziTech:
Thank you, for creating such an amazing editor and being so open to community help. I love being able to improve Atom. 💯

PR #14944, opened 2017-07-03
Show pending IME input instead of just the last typed character

### LINKED ISSUE #14911: 1.19.0-beta2: Cursor stucks at the first letter when using macOS Chinese IME
### Prerequisites

### Description

Cursor stuck at the first letter when using macOS built-in Chinese pinyin IME. 

We expect to input some characters first, and choose a word in IME panel. But from 1.19.0 (maybe beta 1), the cursor always stucks at the first letter. Although I can still type Chinese correctly under the circumstances., it is more helpful to see what you are typing out.

### Steps to Reproduce

**Expected behavior:** 

In 1.18.0, it looks well.

atom

**Actual behavior:** 

However, the cursor moves no more in 1.19.0-beta2.

atom-beta

**Reproduces how often:** every time

### Versions

OS: macOS Sierra 10.12.5
Atom    : 1.19.0-beta2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

### Additional Information

At first, I dived into source code to search for the problem. Soon, I overwhelmed with the large codebase of Atom. I cannot figure out which commit producing this bug. If anyone can point out which part may produce this problem, I can help to solve it. Thanks :)

as-cii:
Fixes #14911.

Back in #13880 we revisited our approach to handling input events, greatly improving support for the [accented character menu](https://support.apple.com/en-us/HT201586) on macOS along the way. In that process, we also rewrote the logic for dealing with IME events, which caused in-progress IME input to be replaced with the last typed character while the IME panel was open:

wrong-behavior

With this pull-request we will now show every character that has been pressed and, only at the end, replace the entire stream of character with the selected IME alternative:

kapture 2017-07-03 at 11 28 24

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1628,7 +1628,7 @@ class TextEditorComponent {
   }
 
   didCompositionUpdate (event) {
-    this.props.model.insertText(event.data, {select: true})
+    this.props.model.insertText(event.data, {select: this.accentedCharacterMenuIsOpen})
   }
 
   didCompositionEnd (event) {


as-cii:
@atom/maintainers: after CI on this pull-request is 💚, I think we should merge it and publish it as part of the next beta hot-fix.

@nathansobo: I couldn't remember the reason why we originally didn't make the change proposed in this pull-request. I think it might just have slipped through without us noticing when changing the accented character menu stuff, but I'd ❤️ to hear your thoughts about this.

@weihanglo @ykpythemind: thanks for reporting #14911 and apologies for that issue! Can you help us confirm this pull-request solves the problem you were observing? For details on how to test it locally, you can check out the [build instructions](https://github.com/atom/atom/tree/master/docs/build-instructions) for your platform. Thanks! ✨

ykpythemind:
@as-cii 
it's not working 
mov
( I typed the key `konnitiwa[enter]sekai[space][enter]`)

weihanglo:
I have followed the build instructions, but the IME still shown only last typed character. I am not sure whether my environment affected the outcome. I built in with 

- Node v7.4.0
- Xcode 8.3.3 (Apple LLVM version 8.1.0, clang-802.0.42)

as-cii:
Upon further investigation, I discovered this problem is actually related to Electron 1.6 (Chrome 56) and should be fixed on Electron 1.7 (Chrome 57). It is being caused by `compositionupdate` reporting just the last inserted character, as opposed to the whole string that's being currently composed (see https://bugs.chromium.org/p/chromium/issues/detail?id=677050).

I will get in touch with the Electron team and see if we can back port a fix to Electron 1.6. I will post further updates on #14911.

Thanks again for the reports, @ykpythemind and @weihanglo!

PR #14948, opened 2017-07-03
Always render 'decoration' class on custom decorations

### LINKED ISSUE #14941: [v1.19-beta2] Custom gutter decorations lack `.decoration` CSS class
Before #13880, all custom gutter decorations had an automatic `.decoration` CSS class (see [here](https://github.com/atom/atom/blob/1f068bde629b8748c7e409e53d56dfadea22108d/src/custom-gutter-component.coffee#L82)).

Now, this is not the case (see [here](https://github.com/atom/atom/blob/245f294cc3bb91ed8701b2f4028af94cac1f4101/src/text-editor-component.js#L1075) and [here](https://github.com/atom/atom/blob/245f294cc3bb91ed8701b2f4028af94cac1f4101/src/text-editor-component.js#L3112)).

This actually creates some compatibility problems.

If this is intentional, I can work around. But it looks more like an oversight tbh.

nathansobo:
Fixes #14941

@lierdakil could you give this a look and see if it addresses the compatibility issues? Thanks very much for the report!

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1727,17 +1727,17 @@ describe('TextEditorComponent', () => {
       let [decorationNode1, decorationNode2] = gutterA.getElement().firstChild.children
       const [decorationNode3] = gutterB.getElement().firstChild.children
 
-      expect(decorationNode1.className).toBe('a')
+      expect(decorationNode1.className).toBe('decoration a')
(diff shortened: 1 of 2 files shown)


lierdakil:
Yep, tested and can confirm it works as intended. Thanks for the quick fix!

PR #14955, opened 2017-07-05
Assign screen-row to each line number as a data field

### LINKED ISSUE #41: Completely broken in 1.19 beta
The line numbers look simply repeat the current line over and over:

screenshot_2017-06-25_23-04-30

screenshot_2017-06-25_23-04-53

as-cii:
Fixes https://github.com/justmoon/relative-numbers/issues/41.
Refs: https://github.com/atom/atom/pull/14893#issuecomment-311430646.

Some packages rely on `data-screen-row` and/or `data-buffer-row` to provide their functionalities. As part of the editor rewrite, we accidentally removed `data-screen-row` from line number nodes, thus causing some packages (e.g. relative-numbers) to stop working.

With this pull-request, we are adding back `data-screen-row` on line numbers, which makes the packages relying on it work correctly again. As part of this bug fix, I noticed that we were not updating `data-buffer-row` and `data-screen-row` after initially constructing the element they referred to, so I have fixed this inconsistency in bcaf655.

I will merge this as soon as the build gets 💚, so that we can cherry pick it onto `1.19-releases` and let this pull-request go out with the next hot-fix release.

/cc: @atom/maintainers @avindra

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -685,6 +685,20 @@ describe('TextEditorComponent', () => {
         ])
       }
 
+      editor.getBuffer().insert([2, 0], '\n')
+      await component.getNextUpdatePromise()
+      {
+        const bufferRows = Array.from(element.querySelectorAll('.line-number:not(.dummy)')).map((e) => e.dataset.bufferRow)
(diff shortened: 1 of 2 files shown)


avindra:
Tested in the beta, works great!

Thanks @as-cii ❤️

PR #14987, opened 2017-07-10
Fix regressions handling mousedown on Linux and macOS

### LINKED ISSUE #8648: Linux middle mouse paste inserts text twice

nathansobo:
Fixes #8648

This PR fixes two regressions related to mouse handling introduced in the latest editor rewrite (#13880) and the upgrade to Electron 1.6.x (#12696).

* On Linux, the middle mouse button pasting text twice (#8648). This regressed because Chrome now seems to synthesize textInput events on mouseup when clicking with the middle button. We had custom logic to emulate this behavior that was rendered redundant by this change.

* On Mac, ctrl-clicking is interpreted by the OS as a right click that brings up the context menu. We weren't bailing out in this case, which caused the cursor to move after the context menu had closed.

We should hot-fix this to 1.19-releases once it is merged.

/cc @ungb

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2412,10 +2412,13 @@ describe('TextEditorComponent', () => {
             ctrlKey: true
           })
         )
-        expect(editor.getCursorScreenPositions()).toEqual([[1, 4]])
+        expect(editor.getSelectedScreenRanges()).toEqual([
+          [[1, 16], [1, 16]]
+        ])
 
         // ctrl-click adds cursors on platforms *other* than macOS
         component.props.platform = 'win32'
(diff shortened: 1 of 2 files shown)


nathansobo:
@ungb Thanks for the lint error fix.

nathansobo:
@ungb As soon as you test this feel free to merge and hotfix if you'd like.

ungb:
@nathansobo sounds good, things are looking good on ubuntu, I'm trying on fedora right now.  I'll merge if everything looks good.

PR #15004, opened 2017-07-12
Pin npm to 5.1 so we can build

damieng:
5.2 causes errors with eslinter-plugin-react


damieng:
Very likely caused by https://github.com/npm/npm/pull/17678

PR #15007, opened 2017-07-13
See what's going on in Travis

damieng:
Want to see what's going on with regards to npm/nvm as the script and the current behavior don't match (former installs latest npm 5.2, latter uses 3.10.5)

**Do not merge this**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,15 @@ before_install:
 install:
   - git clone https://github.com/creationix/nvm.git /tmp/.nvm
   - source /tmp/.nvm/nvm.sh
+  - nvm --version
+  - npm --version
+  - which npm
   - nvm install $NODE_VERSION
   - nvm use --delete-prefix $NODE_VERSION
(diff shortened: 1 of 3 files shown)


50Wliu:
What's the reason for explicitly installing nvm?  Travis comes with nvm pre-installed.

damieng:
I suspect either the intent was we always get the latest - our build script does a `npm install -g npm` or that we expect `npm 3.10.5` as specified in the package.json

Need to loop with the team to find out what is expected. It turns out the build script is also manually running `script\node_modules\bin\npm` to get the version information despite that not being on the path.

50Wliu:
Sorry, I meant n**v**m, not n**p**m.  I appreciate having to install a specific version of npm, but it doesn't seem like installing the bleeding-edge version of nvm will change much.

damieng:
I just wanted to know which version of nvm so I could also check out their source as we install a specific version on circle but the latest on travis.

PR #15011, opened 2017-07-13
Check if incoming element is non-null before attempting to append to it

mehcode:
Fixes #15010

PR #15012, opened 2017-07-13
Support onlyHead, onlyEmpty, and omitEmptyLastRow for gutter decorations

mehcode:
### Description of the Change

Adds support for `onlyHead`, `onlyEmpty` and `omitEmptyLastRow` to custom gutter decorations. 

### Alternate Designs
None

This is changing core TextEditorComponent behavior.

### Benefits
Currently scary hacks [like this](https://github.com/steelbrain/linter-ui-default/blob/3b87b59b1d506f195b9703d1a9d0b23ed412e128/lib/editor/index.js#L94) are needed if a package has a custom gutter and wants to have the cursor-line selection highlight extend over the custom gutter.

This allows a package to opt-in to a cursor-line on their gutter with (example taken from linter-ui-plus package):

```js
    this.textEditor.decorateMarkerLayer(this.textEditor.selectionsMarkerLayer, {
      gutterName: 'linter-ui-plus',
      type: 'gutter',
      class: 'line-number cursor-line-no-selection',
      onlyEmpty: true,
      onlyHead: true
    })

    this.textEditor.decorateMarkerLayer(this.textEditor.selectionsMarkerLayer, {
      gutterName: 'linter-ui-plus',
      type: 'gutter',
      class: 'line-number cursor-line'
    })
```

### Possible Drawbacks
None but I don't understand that much of the text editor code. @nathansobo ?

### Applicable Issues
None that I found

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1112,8 +1112,26 @@ class TextEditorComponent {
       decorations = []
       this.decorationsToRender.customGutter.set(decoration.gutterName, decorations)
     }
-    const top = this.pixelPositionAfterBlocksForRow(screenRange.start.row)
-    const height = this.pixelPositionBeforeBlocksForRow(screenRange.end.row + 1) - top
+
+    let omitLastRow = false
+    if (screenRange.isEmpty()) {
+      if (decoration.onlyNonEmpty) return
+    } else {
+      if (decoration.onlyEmpty) return
+      if (decoration.omitEmptyLastRow !== false) {
+        omitLastRow = screenRange.end.column === 0
+      }
+    }
+
+    let rangeStartRow = screenRange.start.row
+    let rangeEndRow = screenRange.end.row
+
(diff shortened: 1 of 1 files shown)


nathansobo:
Thanks for contributing!

This looks like a good idea. Before we can merge it though we'll need tests. There are already tests for `onlyHead` and related options for the gutter decorations as well as tests for the custom gutter decorations. It would be good to combine them to produce a new test of this behavior so we can be sure it doesn't regress in the future.

The CI failure doesn't seem related. Maybe you could rebase on master to be sure and I can help investigate if it still fails.

mehcode:
I'll see what I can do for a test. I understand very little what's going on in there. This change was more of a gut, "yeah, this looks like it should go here" kind of change.

---

Just want to call your attention to #15011 as a custom gutter decoration in a marker layer hits that bug so it makes utilizing this PR for what I'd like impossible.

PR #15013, opened 2017-07-13
Swap underlying editor correctly when calling setModel on editor element

### LINKED ISSUE #15005: Some text editors in Atom 1.19 don't update to reflect their underlying buffer until the component loses focus
### Description

Nuclide uses text editors for input in some places.  As of Atom 1.19, these text editors don't update to reflect their underlying text buffer until the editor loses focus.

### Steps to Reproduce

1. Install Nuclide package
2. Open Nuclide Http Request Sender via command palette
3. Try to modify the request body.  Note how nothing appears to change
4. Click out of the text editor: all the changes you made now appears

Text editor updates as text buffer is changed

Text editor updates following specific non-input events

Repro demo

100%

### Versions

Atom 1.19.0-beta3

as-cii:
Fixes #15005.

Previously, when `setModel` was called, we forgot to update the pointer to the component in the newly supplied editor. This was causing the element to not update in response to model updates but only as a result of focus or visibility changes. We suspect this regressed during the rewrite of the editor rendering layer.

With this pull request we will now correctly swap the element's underlying editor by updating the component pointer on the newly supplied editor. Also, if the element was already attached to another editor, we will null out the component reference on it, because one instance of `TextEditorElement` can only represent one instance of `TextEditor`.

/cc: @nathansobo @tjfryan

### DIFF
--- a/spec/text-editor-element-spec.js
+++ b/spec/text-editor-element-spec.js
@@ -201,6 +201,33 @@ describe('TextEditorElement', () => {
     })
   })
 
+  describe('::setModel', () => {
+    describe('when the element does not have an editor yet', () => {
+      it('uses the supplied one', () => {
+        const element = buildTextEditorElement({attach: false})
+        const editor = new TextEditor()
+        element.setModel(editor)
+        jasmine.attachToDOM(element)
+        expect(editor.element).toBe(element)
+        expect(element.getModel()).toBe(editor)
+      })
+    })
+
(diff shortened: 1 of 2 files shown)


nathansobo:
This looks reasonable. I'll take a look at the test failure now.

nathansobo:
❤️ @ungb

tjfryan:
Thank you so much!

PR #15027, opened 2017-07-14
Better Icon

jackHedaya:
I believe it is better because it keeps some of the signature Atom green, used the logo, and is more elegant.

### Requirements


lee-dohm:
Thanks for the contribution. As [stated in response to a previous submission](https://github.com/atom/atom/pull/3089#issuecomment-50504224), we won't be accepting brand-related pull requests because of the murky nature of authorship. Changing your own personal icon is totally legit as far as the license goes so have at it if you so choose, but we're going to keep the official brand updates to GitHub staff only.

PR #15045, opened 2017-07-18
Kill all the child processes first on killing a buffered process

### LINKED ISSUE #7252: Problem with killing children processes

karljang:
### Description of the Change

On killing a buffered process, this patch ensures that all of its child processes are killed first.

### Alternate Designs

It might be possible to do the same thing at the client side of BufferedProcess, but BufferedProcess is already doing it on Windows platform as you can see in BufferedProcess.killOnWindows(). That's the reason why I added killOnNix() to BufferedProcess.

BufferedProcess is already doing the same thing on Windows platform, therefore it'll be better to have the same functionality on *nix platforms

### Benefits

We don't have to use Terminal or Task Manager to check if a grand-child process is still running or not.

### Possible Drawbacks

Whenever Atom calls BufferedProcess.kill(), it'll spawn 'pkill' to kill all possible child processes.  But, anyway, Atom already do the same thing by running 'wmic' :)

### Applicable Issues

It fixes #7252 : Problem with killing children processes

### DIFF
--- a/src/buffered-process.js
+++ b/src/buffered-process.js
@@ -205,6 +205,30 @@ export default class BufferedProcess {
     })
   }
 
+  // Kill all child processes of the spawned process on Nix platforms.
+  //
+  killOnNix () {
+    if (!this.process) return
+
+    const parentPid = this.process.pid
+    const cmd = 'pkill'
+    const args = ['-P', `${parentPid}` ]
+
+    let pkillProcess
+
+    try {
+      pkillProcess = ChildProcess.spawn(cmd, args)
+    } catch (spawnError) {
+      this.killProcess()
+      return
+    }
+
+    pkillProcess.on('error', () => { })
+    pkillProcess.stdout.on('close', () => {
+      this.killProcess()
+    })
+  }
+
   killProcess () {
     if (this.process) this.process.kill()
     this.process = null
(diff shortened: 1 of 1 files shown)


karljang:
Fixed a build error caused by coding style inconsistency

PR #15052, opened 2017-07-18
Try an apm that depends on node-gyp explicitly

nathansobo:
For some reason, `npm` isn't flattening all of `apm`'s indirect dependencies in its `node_modules` folder, violating an assumption we're making about the path of `node-gyp`, which is a dependency of the bundled `npm`. To work around this, we're adding an explicit dependency on `node-gyp` so we always know where it will be located.

### DIFF
--- a/apm/package.json
+++ b/apm/package.json
@@ -6,6 +6,6 @@
     "url": "https://github.com/atom/atom.git"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


williamsong1112 (approved):

nathansobo:
Merging without a build since I just bumped from a pre-release to the actual release.

Issue #14162, opened 2017-04-10
Add vibrancy support for macOS


chrisdrackett:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

I wonder if it might be possible to add support for vibrancy (https://github.com/electron/electron/pull/7898) to atom. I think a good first step would be to enable vibrancy for the tree-view. Maybe giving an option in settings to choose between dark and light.

### Steps to Reproduce

None. Feature request

**Expected behavior:** [What you expect to happen]

Ability to set the background of tree view to use vibrancy

**Actual behavior:** [What actually happens]

Not currently possible

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

1.15.0
10.12.4

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

thomasjo:
This will most likely be possible to some extent, once #12696 ships.

Please update your issue body with the issue template (or at very least more details) — we require all issues to be clearly actionable. Non-actionable issues with insufficient details on expected behavior and so on, will be closed.

Thanks!

austin-guevara:
@thomasjo – it seems that this was successfully merged to master. Does that mean that this is now possible in the regular release of Atom? Or am I misreading? Thanks!

50Wliu:
I don't believe it's possible to set vibrancy on _portions_ of a window.  It looks like Atom needs to expose an API or a config to support setting vibrancy, since the Electron APIs for vibrancy are directly tied to BrowserWindows.

chrisdrackett:
still you could set it to the whole window and set things you want to be vibrant as transparent while giving everything else a background color

austin-guevara:
My dream is to have the side panels with vibrancy (tree view, Github panel) and everything else as the standard theme background color. But it sounds like that's not fully possible yet. 

[...]

PR #15077, opened 2017-07-22
Allow to set Vibrancy

the0neyouseek:
### Requirements

### Description of the Change

Allow to set the vibrancy of the browserWindow with the setVibrancy method.

### Alternate Designs

/

Allow to use vibrancy in themes

### Benefits

See above

### Possible Drawbacks

none

### Applicable Issues

See #14162

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -317,3 +317,5 @@ class AtomWindow
 
   disableZoom: ->
     @browserWindow.webContents.setVisualZoomLevelLimits(1, 1)
+
+  setVibrancy: (vibrancy) -> @browserWindow.setVibrancy(vibrancy)


simurai:
You could call `setVibrancy` directly in Electron. E.g.

```js
require('electron').remote.getCurrentWindow().setVibrancy('medium-light');
```

But maybe still a good idea to let people use a more official "Atom" API?

/cc @atom/feedback

the0neyouseek:
@simurai I tried that beforehand but it did not work because `remote.getCurrentWindow()` does not have a `setVibrancy` function (at least not in `1.18.0`)

simurai:
Ohh.. right. It's only available from `1.19`, currently still in [Atom Beta](https://atom.io/beta).

lee-dohm:
Thanks very much for the work on this.

On further review, we have decided that we aren't going to accept this feature at this time. Since, as @simurai points out, people or theme authors can enable this on an individual basis anyway, we don't feel that an official API is necessary or justifies the maintenance cost.

Thanks again for the help and your interest in Atom :+1:

PR #15080, opened 2017-07-23
Replace font stack with Chromium 56's "system-ui" generic font family

### LINKED ISSUE #8656: Atom does not use the system font on Linux (and Mac and Windows)

p-e-w:
### Description of the Change

Atom currently [does not use the system font](https://github.com/atom/atom/issues/8656), but instead hardcodes a bunch of fonts in a "font stack" (priority list), hoping that the actual system font is present in that list and in the right position.

This doesn't always work and gives particularly poor results on Linux, where changing the system font is easy and Atom stands out as the one application that doesn't respect the change. The reason this long-standing issue is still open is that until recently this wasn't possible to fix from CSS (except on macOS with the `BlinkMacSystemFont` placeholder family) and would have required platform-specific code that resolves the correct family.

Enter Chromium 56, shipping with Atom 1.19! It introduces the [`system-ui` generic font family](https://www.chromestatus.com/feature/5640395337760768) that "allows authors to style contents so it fits within the system UI". It works on all platforms! 🎉 🎆 🍾 

Closes #8656. The same change should probably be made also in Atom's default UI themes.

### Alternate Designs

The alternatives are keeping the font stack, or resolving the system font family programmatically. Both are clearly worse.

Because the font family is defined in core.

### Benefits

Atom looks a tiny bit more like a native application.

### Possible Drawbacks

Can't think of any.

### Applicable Issues

#8656

### DIFF
--- a/static/variables/ui-variables.less
+++ b/static/variables/ui-variables.less
@@ -81,5 +81,5 @@
 
 // Other
 
-@font-family: 'BlinkMacSystemFont', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
+@font-family: system-ui;
 @use-custom-controls: true; // false uses native controls


p-e-w:
@atom/maintainers It's been a while... any chance this might get merged soon? It's just a one-line change which should be trivial to test by simply starting Atom on a few different platforms, but it fixes a long-standing UI issue on Linux.

ungb:
Hey @simurai, is this something you would be able to take a look at?

simurai:
Tested `system-ui` in `styles.less` with:

```less
.workspace {
  font-family: system-ui;
}
```

Works on:

- [x] MacOS 👉  `San Francisco`
- [x] Windows 👉  `Segoe UI`

@ungb Do you mind trying it on Linux? No need to build Atom, you can paste that snippet above in your `styles.less` and see if the font looks close to the rest of the OS.

In the meantime, I'll make some PRs for the bundled themes.

ungb:
After further testing with @simurai, everything looks good!

simurai:
Here @ungb screenshot with the change:

image

maxbrunsfeld:
Great fix! Thanks @p-e-w.

Issue #10724, opened 2016-02-07
Atom --safe is talking to other Atom instances


alexandernst:
`atom --safe` spawns an (supposedly) isolated instance of Atom without any plugins/themes, but the reality is that the "safe" instance isn't isolated.
1. Open a normal Atom instance.
2. Open a "safe" Atom instance.
3. Try to open a file from the "safe" Atom instance. 
4. Bug: file is opened in the normal Atom instance window

lee-dohm:
I was able to reproduce this using Atom v1.6.0-dev-ed43eda on Mac OS X 10.11.3. Here are some more detailed repro steps:

1. Change to the directory where the `atom/atom` repo is stored locally (on my system it is in `~/Source/atom`)
1. Launch normal mode Atom with the command: `atom .`
1. Go back to the terminal using <kbd>Cmd+Tab</kbd>
1. Launch Atom in Safe Mode with the command: `atom --safe` (note: **do not** include the period)
1. Open the File Open dialog with <kbd>Cmd+O</kbd>
1. Navigate to the `appveyor.yml` file in the root of the repository and double-click it

**Expected:** `appveyor.yml` to open in the Safe Mode window
**Actual:** `appveyor.yml` opens in the normal mode window

lee-dohm:
With that said, Safe Mode isn't intended as a completely isolated environment. It is designed to launch a session of Atom without:

* Loading community packages
* Executing the `init.coffee`
* Possibly other things I'm forgetting

I just did a simple test in the Safe Mode window I still had open from my repro steps above. I attempted to execute a command from my [tabs-to-spaces package](https://atom.io/packages/tabs-to-spaces) and it wasn't available in the Safe Mode window. It was available from the normal mode window.

alexandernst:
Even so, I'd expect to be able to open files in the safe editor (if I have more editors open). Anyways, the mere fact this is happening proves that there is come kind of communication between atom instances, which could be a place for bugs to appear.

lee-dohm:
I agree that people should be able to open files in the intended instance (which is why I labeled this as a `bug`). If solving that isn't your intention with this Issue and you want Safe Mode to be a complete isolation from the other running instance, that's an architectural thing that should be addressed separately.

Should I open another Issue for the file open bug?

alexandernst:
@lee-dohm Yes, another issue for the file open bug would be fine, thank you!

lee-dohm:
[...]

PR #15081, opened 2017-07-24
Independent Atom instances (per $ATOM_HOME)

mortenpi:
**What problems does this solve?** Let's assume I have two `$ATOM_HOME` directories `A/` and `B/`, with different packages, themes etc. Currently, if I start Atom in the following way

```bash
$ ATOM_HOME=A/ atom
$ ATOM_HOME=B/ atom
```
the second one opens up a new Atom window, but uses packages etc. from `A/`, which feels counter-intuitive. Instead I propose that instances would be per `$ATOM_HOME`. On Linux this can be achieved by having sockets be per-`$ATOM_HOME` it seems.

This would be helpful for IDEs (e.g. [Juno](http://junolab.org/)) which load a lot of IDE-specific features and UI components. With this PR you could simply have a `juno` "executable" that wraps around an existing Atom installation, but does not interfere with normal Atom usage:

```sh
#!/bin/sh
export ATOM_HOME=$HOME/.juno
exec atom $@
```

**How does it solve it?** This PR puts the socket into `$ATOM_HOME`, instead of `os.tmpdir()`.

I've been running `1.20.0-dev` with this change for a week or so now and it has been working fine so far. But I'm not sure whether this particular PR is the best solution.

Also, it is Linux-specifc (and macOS?). I can't test this on macOS, nor provide a solution for Windows. If someone could provide a more complete solution I'd be happy if this PR would be superseded.

**Drawbacks/alternatives.** I don't see this affecting normal Atom usage in any way.

In case sockets have to be under `os.tmpdir()` (due to potential filesystem permission issues or something like that?), then an alternative would be to have a `abspath(process.env.ATOM_HOME)`-dependent piece in the socket name, where `abspath` would resolve any relative paths etc.

### DIFF
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -8,6 +8,7 @@ FileRecoveryService = require './file-recovery-service'
 ipcHelpers = require '../ipc-helpers'
 {BrowserWindow, Menu, app, dialog, ipcMain, shell, screen} = require 'electron'
 {CompositeDisposable, Disposable} = require 'event-kit'
+crypto = require 'crypto'
 fs = require 'fs-plus'
 path = require 'path'
 os = require 'os'
@@ -33,11 +34,16 @@ class AtomApplication
   # Public: The entry point into the Atom application.
   @open: (options) ->
     unless options.socketPath?
+      username = if process.platform is 'win32' then process.env.USERNAME else process.env.USER
+      # Lowercasing the ATOM_HOME to make sure that we don't get multiple sockets
+      # on case-insensitive filesystems due to arbitrary case differences in paths.
+      atomHomeUnique = path.resolve(process.env.ATOM_HOME).toLowerCase()
+      hash = crypto.createHash('sha1').update(username).update('|').update(atomHomeUnique)
+      atomInstanceDigest = hash.digest('hex').substring(0, 32)
       if process.platform is 'win32'
-        userNameSafe = new Buffer(process.env.USERNAME).toString('base64')
-        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{userNameSafe}-#{process.arch}-sock"
+        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{process.arch}-#{atomInstanceDigest}-sock"
       else
-        options.socketPath = path.join(os.tmpdir(), "atom-#{options.version}-#{process.env.USER}.sock")
+        options.socketPath = path.join(os.tmpdir(), "atom-#{options.version}-#{process.arch}-#{atomInstanceDigest}.sock")
 
     # FIXME: Sometimes when socketPath doesn't exist, net.connect would strangely
     # take a few seconds to trigger 'error' event, it could be a bug of node


alexandernst:
This might fix https://github.com/atom/atom/issues/10724

src/main-process/atom-application.coffee
```diff
@@ -37,7 +37,7 @@ class AtomApplication
         userNameSafe = new Buffer(process.env.USERNAME).toString('base64')
         options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{userNameSafe}-#{process.arch}-sock"
```
segevfiner: Supporting this on Windows will probably require adding a *normalized* and *escaped*  `ATOM_HOME` in the pipe name. Normalizing and escaping is necessary to avoid paths that are the same but point to the same directory and to avoid adding backslashes to the pipe path.

mortenpi:
I used the hash instead, so that the socket path wouldn't get too long. But I could also switch to using an escaped version of `atomHomeUnique`. Note that I _have not tested_ whether this code actually works on Windows.

Currently it will not fix #10724 currently, since I think `--safe` has the same `ATOM_HOME` as the normal instance. But I suppose in principle we could have a separate socket for safe instances, e.g. by append `-safe` to the socket path?

cstockton:
@mortenpi I think you're work here may actually fix the behavior I describe in #15605, so depending on the atom teams view of the behavior it may be considered a bug fix rather than RFC?

mortenpi:
@cstockton I think it will, although I didn't test. It would be awesome if you could fetch this branch, compile it and try it out to make sure.

Since no one has argued against the described behaviour, I'll removed the RFC, assuming it means agreement. I fixed the thing that gave a linter error and rebased. After someone tests it on Windows it would be good to go in my opinion. I've actively been running this patched version on Ubuntu since July, generally working in parallel in two different Atom instances with two different `ATOM_HOME`s.

src/main-process/atom-application.coffee
```diff
@@ -35,9 +36,14 @@ class AtomApplication
     unless options.socketPath?
       if process.platform is 'win32'
         userNameSafe = new Buffer(process.env.USERNAME).toString('base64')
```
damieng: Let's join the USERNAME and ATOM_HOME together with a | separator and then MD5 and base64 it.  Will keep things quite unique while keeping the socket name length manageable.

src/main-process/atom-application.coffee
```diff
@@ -35,9 +36,14 @@ class AtomApplication
...
-        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{userNameSafe}-#{process.arch}-sock"
+        # To have a separate session per ATOM_HOME directory, we append the first
+        # few characters of the the hash of ATOM_HOME to the socket file name.
+        atomHomeUnique = path.resolve(process.env.ATOM_HOME)
+        hash = crypto.createHash('sha1').update(atomHomeUnique)
+        atomHomeDigest = hash.digest('hex').substring(0, 12)
+        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{userNameSafe}-#{process.arch}-#{atomHomeDigest}-sock"
       else
-        options.socketPath = path.join(os.tmpdir(), "atom-#{options.version}-#{process.env.USER}.sock")
+        options.socketPath = path.join(process.env.ATOM_HOME, "atom-#{options.version}-#{process.env.USER}.sock")
```
cstockton: @mortenpi Hi- was just taking a look at this, I don't think you want to put sockets in atom home. In some systems `socket` syscall might be denied by selinux / apparmor / etc, so I think it may be better to use the file name to ensure a unique socket while keeping determinism in place of a folder. I.E. Make this small adjustment:

```coffeescript
    # To have a separate session per ATOM_HOME directory, we append the first
    # few characters of the the hash of ATOM_HOME to the socket file name.
    atomHomeUnique = path.resolve(process.env.ATOM_HOME)
    hash = crypto.createHash('sha1').update(atomHomeUnique)
    atomHomeDigest = hash.digest('hex').substring(0, 12)
    unless options.socketPath?
      if process.platform is 'win32'
        userNameSafe = new Buffer(process.env.USERNAME).toString('base64')
        options.socketPath = "\\\\.\\pipe\\atom-#{options.version}-#{userNameSafe}-#{process.arch}-#{atomHomeDigest}-sock"
      else
        options.socketPath = path.join(os.tmpdir(), "atom-#{options.version}-#{process.env.USER}-#{atomHomeDigest}.sock")
```

mortenpi: Thanks for the feedback! I was wondering whether there might be a reason like that to prefer `/tmp` over `/home`. I restored `os.tmpdir()` as you suggested.

mortenpi:
@damieng I stuck with SHA-1 + the first 12 characters of the hex for now, but I can switch to MD5 if that is preferred. But I put both the username and `ATOM_HOME` into the digest as you suggested. I also made the naming of the socket consistent on Windows and Linux.

damieng:
@mortenpi It's not specifically the algorithm just rather if we go with MD5 and base64 we can be a bit more confident about avoiding collisions.

mortenpi:
I made the simplest change of swapping to a longer substring, retaining the same number of bits as a MD5 hash would be. Note that I chose the 12 characters initially based on the fact that it is what they use to truncate the Git hashes in the kernel, without getting collisions.

I do feel that hex is preferred over base64, since wouldn't base64 introduce some semi-weird characters (=, + etc) into the filename? I don't think it would cause any technical issues, but I at least prefer avoiding filenames like that.

src/main-process/atom-application.coffee
```diff
@@ -33,11 +34,15 @@ class AtomApplication
   # Public: The entry point into the Atom application.
   @open: (options) ->
     unless options.socketPath?
+      username = if process.platform is 'win32' then process.env.USERNAME else process.env.USER
+      atomHomeUnique = path.resolve(process.env.ATOM_HOME)
+      hash = crypto.createHash('sha1')
+      hash.update(username).update('|').update(atomHomeUnique)
```
damieng: We can probably merge this line with the line above for clarity.

src/main-process/atom-application.coffee
```diff
@@ -33,11 +34,15 @@ class AtomApplication
   # Public: The entry point into the Atom application.
   @open: (options) ->
     unless options.socketPath?
+      username = if process.platform is 'win32' then process.env.USERNAME else process.env.USER
+      atomHomeUnique = path.resolve(process.env.ATOM_HOME)
```
damieng: One issue here is that case-preserving case-insensitive systems (Windows NTFS, macOS X HFS+ default mode) are going to spawn two instances of Atom for the same folder should the case differ.

mortenpi: I don't believe there's any reliable way of detecting whether a filesystem is case-sensitive or not. We could lowercase the whole string before hashing and risk the user not getting an instance on a case-sensitive filesystem. I do think that the likelihood of a user having two identical paths, except case, on a case-sensitive system is lower than seeing arbitrary case differences on the other type, so it would make sense on those grounds.

mortenpi:
Merged the lines and `.toLowerCase()`d the path.

mortenpi:
Is there anything other than testing that is holding this back?

maxbrunsfeld:
This code doesn't work on macOS. On master, I see this error when launching atom:

```sh
atom $ atom -f .
Application server failed { Error: listen EADDRINUSE /var/folders/fg/g0tf_tmx2nzbyb__zps26kx00000gn/T/atom-1.25.0-dev-a8e457d-x64-f6e618d4342c8babc6b7456a5ec0a564.sock
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at Server._listen2 (net.js:1249:19)
    at listen (net.js:1298:10)
    at Server.listen (net.js:1382:7)
    at AtomApplication.module.exports.AtomApplication.listenForArgumentsFromNewProcess (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/atom-application.js:330:14)
    at AtomApplication.module.exports.AtomApplication.initialize (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/atom-application.js:147:12)
    at Function.module.exports.AtomApplication.open (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/atom-application.js:67:38)
    at App.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/start.js:71:21)
    at emitTwo (events.js:111:20)
    at App.emit (events.js:191:7)
  code: 'EADDRINUSE',
  errno: 'EADDRINUSE',
  syscall: 'listen',
  address: '/var/folders/fg/g0tf_tmx2nzbyb__zps26kx00000gn/T/atom-1.25.0-dev-a8e457d-x64-f6e618d4342c8babc6b7456a5ec0a564.sock',
  port: -1 }
```

I believe the problem is that unlike for regular files, there is a limit to the length of unix domain socket names.

From [the BSD man pages](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_un.h.html#tag_13_67_04)

> The size of sun_path has intentionally been left undefined. This is because different implementations use different sizes. For example, 4.3 BSD uses a size of 108, and 4.4 BSD uses a size of 104. Since most implementations originate from BSD versions, the size is typically in the range 92 to 108.
>
> Applications should not assume a particular length for sun_path or assume that it can hold {_POSIX_PATH_MAX} bytes (256).

On my system, `sys/un.h` contains the following:

```c
struct sockaddr_un {
  unsigned char   sun_len;        /* sockaddr len including null */
  sa_family_t     sun_family;     /* [XSI] AF_UNIX */
  char            sun_path[104];  /* [XSI] path name (gag) */
};
```

It doesn't look like there's an easy way to determine the size limit programmatically, other than using `sizeof` in C.

maxbrunsfeld:
I'm going to revert this in order to fix the single-instance behavior on mac. I'm open to suggestions on how to fix it.

mortenpi:
I guess there no way to make the "/tmp" part shorter since it comes from the OS? `/var/folders/fg/g0tf_tmx2nzbyb__zps26kx00000gn/T/` seems pretty excessive.. Also, that is apparently only a symlink -- the actual path seems to have `/private` prepended to it as well (nodejs/node#11422), meaning the "/tmp" prefix on macOS is 57 characters.

The whole path is 114 characters (121 with `/private`) at the moment. We could get it down to 94 (101) if we keep only 12 characters of the hash, as I had initially (I don't think collisions would be an issue [1]). We could also hide the version number in the hash.

Going with `os.tmpdir()/atom-#{12-char-hash}.sock` would be a total of 79 characters, if we also count the `/private` prefix, which would hopefully fit without problems. And it would also have a completely predictable length.

Note that the current implementation (without the `ATOM_HOME` hash) should also crash if your username is 20-30 characters.

---

[1] [Following this](http://preshing.com/20110504/hash-collision-probabilities/), 12 char hash allows for `N = 16^12` unique hashes, which leads to a [10^-11 probability](https://www.wolframalpha.com/input/?i=1+-+exp(-100*(100-1)+%2F+(2+*+(16%5E12)))) of a collision if you assume that you have 100 different Atom instances running (i.e. different `ATOM_HOME` x Atom version). Simplistically, we would need every person on the planet run 10 computers with 100 separate Atom instances to see a single collision. I think we could make the hash even shorter in principle.

maxbrunsfeld:
> Going with os.tmpdir()/atom-#{12-char-hash}.sock would be a total of 79 characters, if we also count the /private prefix

Yup, that works for me. If we want even fewer collisions with a 12 hash, we could always use `base64` but replace forward slashes with a different path-safe character like a dot.

```js
hash.digest('base64').substring(0, 12).replace(/\//g, '.')
```

But if you prefer a hex hash, I think you're right that the odds of collision are probably sufficiently low. Usually people only have one or maybe two atom home folders.

PR #15082, opened 2017-07-24
Should not use bind for map's callback in pane.coffee.

t9md:
`Array.prototype.map` call callback with `callback(item, index, this)`.
So using `bind` is not appropriate for the function which take 2nd arg.

cc @maxbrunsfeld 

### Requirements

### Description of the Change

I found this via code-reading. Not sure there is already reported issue.
Calling `pane.destroyItems()` don't `@promptToSaveItem` for items with index greater than 0.
Since map pass index, index goes to `force` arg of `destroyItems`. `force` become true unless index `0`.

### Alternate Designs

No

Bug

### Benefits

Bug free.

### Possible Drawbacks

No.

### DIFF
--- a/src/pane.coffee
+++ b/src/pane.coffee
@@ -637,7 +637,7 @@ class Pane
   # Public: Destroy all items.
   destroyItems: ->
     Promise.all(
-      @getItems().map(@destroyItem.bind(this))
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
:zap: Thanks for fixing this!

PR #15096, opened 2017-07-26
Define indent/outdent behavior for single-line selected text

Ingramz:
Coming from atom/snippets#239, I have now realized that there isn't a properly defined behavior for selecting text on single line and indenting/outdenting it:

indent_undefined_behavior

Here is why I don't think this has been defined yet:
1. There is no spec for single-line selected text being indented via `indent()`: https://github.com/atom/atom/blob/a7736b81e3f6628f2a9e685f28f4c1df9a5e4615/spec/text-editor-spec.coffee#L3673-L3789

2. `indentSelectedRows` (which is used in `indent()`) is poorly documented:
https://github.com/atom/atom/blob/a7736b81e3f6628f2a9e685f28f4c1df9a5e4615/src/selection.coffee#L671-L679
    > If the selection spans multiple rows, indent all of them.

    It's possible to interpret this in a way that if the selection does not span across multiple rows, then nothing gets indented. [_Insert your favorite programmer-goes-to-grocery-store-joke here._](https://www.reddit.com/r/Jokes/comments/1nmkfq/a_programmer_is_going_to_the_grocery_store/)

3. The current behavior is not intuitive and unexpected for anyone coming from a different editor. With the exception of Brackets, other text editors tend to agree that selecting text and pressing tab within a single line should not indent the whole line, but insert/overwrite with a tab.

Outdent is done through `outdentSelectedRows`, which [has a spec for single line selected text](https://github.com/atom/atom/blob/2420d12e44f33f0178c8c0ef41e3a2c22cf4fb8e/spec/text-editor-spec.coffee#L4265-L4270). However it seems to be slightly mislabeled, there is actually less than one line selected. That is not a big deal and can be easily fixed, but it's possible that something is missing from here too.

At minimum I am looking to define behavior for following cases:
1. Selecting whole line and issuing `indent()`.
2. Selecting substring within line with length greater than 0 and issuing `indent()`.
3. Selecting whole line and issuing `outdentSelectedRows()`.
4. Selecting substring within line with length greater than 0 and issuing `outdentSelectedRows()`.

There might be additional cases that might need be considered, but I don't want this to be come overly ambitious, so starting with a more managable goal might be better.

The motivation for all of this comes from behavior related to text selections that come from traversing tab stops within snippets, where indenting/outdenting may lead to unexpected results:

indent_snippet1
indent_snippet2

Personally I think that outdent behavior can be kept as-is and it is sufficient to simply improve the spec and cover cases 3 and 4. Outdenting does not break out of snippet's tab stops whereas indenting does, so indicating change with a slightly more destructive (but predictable) operation should be obvious enough that it's impossible to return to snippet's tab stops.

Visual Studio Code's behavior is a suitable candidate, which indents if the whole line is selected, but inserts a tab when only part of line is selected. Outdent is handled in the same way as Atom does it now. If there aren't any objections or better ideas, I might just implement that and propose this as the final solution.

Code at the time of writing in pull request is just to get the ball rolling and does not represent the final proposed solution in any way. But it does replace everything with a tab if at most only one line has been selected.

### DIFF
--- a/spec/text-editor-spec.coffee
+++ b/spec/text-editor-spec.coffee
@@ -3766,13 +3766,27 @@ describe "TextEditor", ->
                 expect(buffer.lineForRow(7)).toMatch /^\t\t\t\t$/
                 expect(editor.getCursorBufferPosition()).toEqual [7, 4]
 
-      describe "when the selection is not empty", ->
+      describe "when selection spans across multiple lines", ->
         it "indents the selected lines", ->
           editor.setSelectedBufferRange([[0, 0], [10, 0]])
           selection = editor.getLastSelection()
           spyOn(selection, "indentSelectedRows")
           editor.indent()
           expect(selection.indentSelectedRows).toHaveBeenCalled()
+
+      describe "when a single line has been selected from start to end", ->
+        it "indents the selected line", ->
+          editor.setSelectedBufferRange([[0, 0], [0, 29]])
+          selection = editor.getLastSelection()
+          spyOn(selection, "indentSelectedRows")
+          editor.indent()
+          expect(selection.indentSelectedRows).toHaveBeenCalled()
+
+      describe "when selection is made within a single line", ->
+        it "inserts a tab in place of the selection", ->
+          editor.setSelectedBufferRange([[0, 4], [0, 13]])
+          editor.indent()
+          expect(buffer.lineForRow(0)).toEqual "var #{editor.getTabText()} = function () {"
 
       describe "if editor.softTabs is false", ->
         it "inserts a tab character into the buffer", ->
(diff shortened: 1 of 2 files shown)


Ingramz:
In the meanwhile I implemented vscode-like behavior, so it can be tested, turned out to be easier than expected. I'll clean up commit list after feedback.

lee-dohm:
Thanks @Ingramz for the work you've put into this.

We agree that the behavior could definitely be better defined. We would love to hear your thoughts on some of the edge-cases and potential drawbacks of whatever approach you are ultimately considering here. It looks like there are a lot of interactions between selections, indenting/outdenting and snippets that could be tricky. As you get closer, please update the main body of the PR with the full details requested in the PR template and ping us again :+1:

xjArea:
@Ingramz Hello, In fact, I also found out that in SublimeText and Notepad++, only when we selected one or more rows, pressing the tab key will indent, otherwise it is just a normal tab operation. To solve this problem, I already spend four hours, but there is no any result. I am very happy to find your Issue, now two years have passed. I found this problem in Atom still unresolved. Is there any way to solve it ?

PR #15108, opened 2017-07-27
Remove duplicate mocking

nathansobo:
Prior to this PR, we were mocking `Date.now` once in the spec helper, then blowing up with a redundant mock error when `jasmine.useMockClock` is called.

Since this was newly introduced, let's only set it up when clock mocking is specifically requested to minimize breakage. We really need to deprecate the global spec helper someday. It's bad news.

This needs to be hot-fixed to 1.19.

/cc @hansonw

### DIFF
--- a/spec/pane-spec.js
+++ b/spec/pane-spec.js
@@ -1350,6 +1350,7 @@ describe('Pane', () => {
     let editor1, pane, eventCount
 
     beforeEach(async () => {
+      jasmine.useMockClock()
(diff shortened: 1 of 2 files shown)


hansonw:
Yay! Thanks for the quick fix.

nathansobo:
@iolsen went ahead and added this to the hotfix tracking column since I think it slipped through the cracks.

PR #15121, opened 2017-07-28
Upgrade text-buffer to fix spurious conflict and reload events

maxbrunsfeld:
Depends on https://github.com/atom/text-buffer/issues/241
Depends on https://github.com/atom/superstring/pull/19
Depends on https://github.com/atom/superstring/pull/21
Refs https://github.com/atom/text-buffer/issues/240

/cc @hansonw Could you give this is a try with some remote files when you get a chance?


hansonw:
Thanks so much for the quick turnaround! Will definitely give it a try.

PR #15124, opened 2017-07-29
Use spec reporters in CI

smashwilson:
Use the JUnit XML reporter from [jasmine-reporters](https://www.npmjs.com/package/jasmine-reporters) to produce test reports on CI builds on CircleCI and AppVeyor. This should expose test metadata that we can use to:

* More quickly diagnose failures
* Produce metadata that we can consume to identify flakes over time

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ platform:
 environment:
   global:
     ATOM_DEV_RESOURCE_PATH: c:\projects\atom
-    TEST_JUNIT_XML_PATH: c:\projects\atom\junit-results.xml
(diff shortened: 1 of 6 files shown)


smashwilson:
One nice thing about Jasmine reporters vs. Mocha ones: Jasmine reporters are additive by default, so we still get console output and don't have to worry about tests timing out because it's only writing XML.

We will have to add support for respecting the `TEST_JUNIT_XML_ROOT` environment variable to all of our core packages that use custom runners.

smashwilson:
screen shot 2017-07-29 at 5 15 52 pm

it's happening

PR #15138, opened 2017-08-01
Update README.md

alanyee:
### Description of the Change

-Added Homebrew install instructions for macOS
-Made explicit HTTPS calls
-Minor url cleanup

### Alternate Designs

N/A

Most macOS developers use Homebrew, so having the instructions on the README would be helpful. HTTPS calls make web browsing safer.

### Benefits

The benefits are the same reason why they should be in core, and that there should be no drawbacks for including them in the README.

### Possible Drawbacks

N/A

### Applicable Issues

N/A


eigan:
```diff
- Run to `brew cask install atom` install the latest version of Atom.
+ Run `brew cask install atom` to install the latest version of Atom.
```

50Wliu:
The URL changes look good.  However, can you please remove the Homebrew installation instructions?  As we don't maintain the Homebrew version, we feel uncomfortable with adding an installation method that we don't have control over.

alanyee:
@50Wliu Wait really? But members of the Atom organization maintain the Chocolatey package?  May I request members of the Atom organization also maintain the cask at https://github.com/caskroom/homebrew-cask? `brew cask install atom` is on their front page after all.

50Wliu:
Yeah, we maintain the Chocolatey version (https://github.com/atom/chocolatey).  It was the main Windows installation method before the switch to Squirrel.Windows.  It's unlikely that we'll be maintaining the cask due to the existing amount of installation methods we have to support, but I'll bring it up to the team.

PR #15147, opened 2017-08-03
Ensure custom gutter decoration elements fill their container

nathansobo:
In the previous implementation of text editor rendering, [we were assigning an inline height on decoration elements](https://github.com/atom/atom/blob/b580ff8642d4401892c730f6bd2e1b38bd00b3e5/src/custom-gutter-component.coffee#L104). The rewrite in #13880 accidentally dropped this behavior, which was causing empty decoration elements to be zero height and not display a background in some cases. We also drop the `width: min-content` style on the `.line-number` class, which is being used by packages to style decorations in some cases. This width setting was needed at one phase of development but doesn't appear to be needed any longer.

Here is an example with `linter-ui-default` (configured with gutter decorations on the right), `nucleus-dark` UI theme, and `dracula` syntax theme.

Before this PR:
screen shot 2017-08-02 at 8 37 33 pm

After this PR:
screen shot 2017-08-02 at 8 38 22 pm

This needs to be hot-fixed to 1.19 before we roll the railcars.

/cc @jasonrudolph

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1737,6 +1737,8 @@ describe('TextEditorComponent', () => {
       const marker3 = editor.markScreenRange([[9, 0], [12, 0]])
       const decorationElement1 = document.createElement('div')
       const decorationElement2 = document.createElement('div')
+      // Packages may adopt this class name for decorations to be styled the same as line numbers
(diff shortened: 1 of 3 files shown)


jasonrudolph (approved):
:zap:

PR #15148, opened 2017-08-03
Don't enable electron logging by default

nathansobo:
@as-cii @jasonrudolph This cleans up the test noise we were seeing on our package.

In #9268, @thomasjo set the `ELECTRON_ENABLE_LOGGING` environment variable to true. He states in that PR that this was necessary to enable `console.log` to go to `stdout`, and I didn't take the time to go back and test that ancient version of Electron to confirm this to be true, but I'll take him at his word.

Now, however, `console.log` seems to work irrespective of this environment variable. Furthermore setting this environment variable has the unfortunate side effect of causing lots of fairly low level warning messages from within Chromium's internals to spam the console when Atom is run headless. In this PR, I remove the lines from `atom.sh` and `atom.cmd` that set this variable. It's still possible to automatically assign the variable via the `--enable-electron-logging` flag.

@damieng would you mind checking my changes in `cmd.sh` and maybe @ungb could you test it out? I just want to make sure I didn't break the script.

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -31,6 +31,9 @@ while getopts ":wtfvh-:" opt; do
         foreground|benchmark|benchmark-test|test)
           EXPECT_OUTPUT=1
           ;;
+        enable-electron-logging)
+          export ELECTRON_ENABLE_LOGGING=1
+          ;;
       esac
       ;;
     w)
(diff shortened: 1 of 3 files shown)


damieng:
We probably want to set ELECTRON_ENABLE_LOGGING to nothing at the start of the file. Otherwise once you specify it once it's on forever until you restart that command prompt (at least on Windows)

nathansobo:
Thanks for the review @damieng. N00b tip on how can I set it to nothing? `set ELECTRON_ENABLE_LOGGING=NO`? `set ELECTRON_ENABLE_LOGGING=''`?

damieng:
Oops, sorry yes it should be `SET ELECTRON_ENABLE_LOGGING=`  (weird syntax)

nathansobo:
Oh duh there's examples already sitting right there.

PR #15154, opened 2017-08-03
Fix flaky test re: blinking cursor

### LINKED ISSUE #15122: Flaky test - it blinks cursors when the editor is focused and the cursors are not moving
https://ci.appveyor.com/project/Atom/atom/build/7387/job/usobg8qq24sqtjdy

```
TextEditorComponent
  rendering
    it blinks cursors when the editor is focused and the cursors are not moving
      Expected '0' to be '1'.
        at it (C:\projects\atom\spec\text-editor-component-spec.js:414:49)
      Expected '0' to be '1'.
        at it (C:\projects\atom\spec\text-editor-component-spec.js:415:49)
```

jasonrudolph:
Fixes #15122

---

As shown in #15122, [these assertions in `spec/text-editor-component-spec.js`](https://github.com/atom/atom/blob/7d6bd2a6b19be004254865d163d74bd5e849be2f/spec/text-editor-component-spec.js#L414-L415) occasionally fail on AppVeyor:

```
TextEditorComponent
  rendering
    it blinks cursors when the editor is focused and the cursors are not moving
      Expected '0' to be '1'.
        at it (C:\projects\atom\spec\text-editor-component-spec.js:414:49)
      Expected '0' to be '1'.
        at it (C:\projects\atom\spec\text-editor-component-spec.js:415:49)
```

I haven't been able to reproduce the failure locally, but I have a hunch regarding the source of the intermittent failures. I think it might be a case of "overspecification" in the test's assertions. The test expects the blinking cursor to *start* in the visible state, and then transition to the invisible state. When we see the failure above, I suspect that the cursor has already transitioned from the visible state to the invisible state by the time the assertions run.

Since the test aims to verify that the cursor blinks [1], it seems like we should focus on the blinking, and not worry about the *initial* state of the cursor. This PR removes the assertions that verify the initial state of the cursor, and instead asserts that the cursor toggles between the visible and the invisible state.

@nathansobo: Given your familiarly with this test, would you mind reviewing this change? I *think* this change has a good chance of resolving the intermittent failures while still maintaining the spirit of the test. What do you think?

---

[1] eb22b58756add00ae8f2a00d123e52e8ae2f6c69 introduced this test as a smoke test for cursor blinking

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -424,15 +424,8 @@ describe('TextEditorComponent', () => {
       editor.moveRight()
       await component.getNextUpdatePromise()
 
-      await conditionPromise(() =>
-        getComputedStyle(cursor1).opacity === '0' && getComputedStyle(cursor2).opacity === '0'
-      )
-      await conditionPromise(() =>
-        getComputedStyle(cursor1).opacity === '1' && getComputedStyle(cursor2).opacity === '1'
-      )
-      await conditionPromise(() =>
-        getComputedStyle(cursor1).opacity === '0' && getComputedStyle(cursor2).opacity === '0'
-      )
+      expect(getComputedStyle(cursor1).opacity).toBe('1')
+      expect(getComputedStyle(cursor2).opacity).toBe('1')
     })
 
     it('gives cursors at the end of lines the width of an "x" character', async () => {
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -411,26 +411,28 @@ describe('TextEditorComponent', () => {
       await component.getNextUpdatePromise()
       const [cursor1, cursor2] = element.querySelectorAll('.cursor')
 
-      expect(getComputedStyle(cursor1).opacity).toBe('1')
-      expect(getComputedStyle(cursor2).opacity).toBe('1')
-
-      await conditionPromise(() =>
-        getComputedStyle(cursor1).opacity === '0' && getComputedStyle(cursor2).opacity === '0'
-      )
-
       await conditionPromise(() =>
         getComputedStyle(cursor1).opacity === '1' && getComputedStyle(cursor2).opacity === '1'
       )
-
       await conditionPromise(() =>
         getComputedStyle(cursor1).opacity === '0' && getComputedStyle(cursor2).opacity === '0'
+      )
+      await conditionPromise(() =>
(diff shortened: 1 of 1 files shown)


spec/text-editor-component-spec.js
```diff
@@ -411,26 +411,28 @@ describe('TextEditorComponent', () => {
...
       )
+      await conditionPromise(() =>
+        getComputedStyle(cursor1).opacity === '1' && getComputedStyle(cursor2).opacity === '1'
+      )
 
       editor.moveRight()
       await component.getNextUpdatePromise()
 
-      expect(getComputedStyle(cursor1).opacity).toBe('1')
-      expect(getComputedStyle(cursor2).opacity).toBe('1')
```
nathansobo: Restructuring the earlier assertions seems good, but I'm concerned about dropping these. We want to assert that the cursor is visible for a short duration after moving before resuming blinking. Did you see a specific failure on these assertions or just the ones above? Losing this wouldn't be the end of the world since a regression in cursor blinking behavior wouldn't go unnoticed for long, but it does drop some coverage.

jasonrudolph: > Did you see a specific failure on these assertions or just the ones above? 

No, but if my [theory](https://github.com/atom/atom/pull/15154#issue-247766691) about the earlier assertions is accurate (i.e., that they've already gone from the visible state to the invisible state by the time the assertion runs), I would expect that these assertions are vulnerable to the same issue.

@nathansobo: Does that ☝️ make sense? Or, should we be able to guarantee that these assertions succeed?

```
editor.moveRight()
await component.getNextUpdatePromise()
expect(getComputedStyle(cursor1).opacity).toBe('1')
expect(getComputedStyle(cursor2).opacity).toBe('1')
```

jasonrudolph: @nathansobo and I discussed this ☝️ today. Summary:

- We think that the assertion changes are appropriate for the [first batch of operations](https://github.com/atom/atom/blob/29810c6cd1924e08126bd18d759b80abfd1c199c/spec/text-editor-component-spec.js#L408-L412) in this test. We don't offer any guarantees about the _initial_ cursor visibility state following that set of operations.
- We think the assertion changes are not needed for the [second batch of operations](https://github.com/atom/atom/blob/29810c6cd1924e08126bd18d759b80abfd1c199c/spec/text-editor-component-spec.js#L424) in this test. After moving the cursor, the cursor should indeed be in the visible state, so the original assertions should allow us to verify that behavior.

PR #15160, opened 2017-08-03
Open the save dialog in the project root for new items

### LINKED ISSUE #14732: "Save As" for new file doesn't default to a sensible directory location
### Prerequisites

### Description

When creating a new file and hitting Save, the save dialog doesn't seem to have any particular directory set and defaults to a platform-dependent location.

### Steps to Reproduce

1. Create a new untitled file
2. Save it

**Expected behavior:**

"Save As" dialog should default to a sensible location (most likely the project root, which definitely _used_ to be the behaviour)

**Actual behavior:**

**On Ubuntu**, Save dialog opens at "Recently Used", sometimes making it a pain to find the logical place to save the file.

**On mac**, seems to default to project root directory, but **only** for the first window opened (I assume it defaults to the process's cwd)

Can't speak for windows right at the moment I'm afraid

Directory is set fine if you're hitting "save as" from a file that already exists.

**Reproduces how often:**

Every time

### Versions

Currently 1.18 beta but basically every version since #13978 was reported

### Additional Information

I guess #13978 got closed because the desired behaviour was fixed on mac for the first-opened window and for already-existing files. Looks like this was also reported in #14304 but that was closed because it didn't follow the template.

segevfiner:
### Description of the Change

When you try to save a new file (Save an item without a uri), the save as dialog will open in a default OS dependent directory. For example, on Windows it opens in the working directory. But if you have a project open, it makes more sense to open it in the project root folder.

### Benefits

When you save a new file, the save as dialog will start in the project folder.

### Possible Drawbacks

I think that if you pass a `defaultPath` to `dialog.showSaveDialog` than it takes precedence over the OS remembering the last directory used in the dialog.

### Applicable Issues

Fixes #14732

### DIFF
--- a/spec/pane-spec.js
+++ b/spec/pane-spec.js
@@ -755,6 +755,17 @@ describe('Pane', () => {
         })
       })
 
+      describe('when the current item has a saveAs method and a project is open', () => {
+        it('opens a save dialog with the project directory', async () => {
+          pane.getActiveItem().saveAs = jasmine.createSpy('saveAs')
+          atom.project.setPaths([__dirname])
+          await pane.saveActiveItem()
+          expect(showSaveDialog).toHaveBeenCalledWith({
+            defaultPath: __dirname
+          })
+        })
+      })
+
       it('does nothing if the user cancels choosing a path', async () => {
         pane.getActiveItem().saveAs = jasmine.createSpy('saveAs')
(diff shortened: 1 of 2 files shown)


segevfiner:
Fixed differently in #15374

PR #15163, opened 2017-08-04
Automatically manage focus for modal panels

wbinnssmith:
This implements automatic focus management for modal panels using the excellent focus-trap module.

Upon being shown, modals will have their first tabbable element automatically focused, and shifting focus with the tab key (or more correctly the core:focus-next command) will be limited to the contents of the modal. If the modal does not have any tabbable elements, focus() will be sent to the panel's root element (if it implements it).

I'm happy to update this to *always* calls focus on the panel's root element, but then modal implementers would need to handle that and focus things on their own. I'd argue the tabbable element behavior is more accessible though :)

This has the effect of not automatically closing most of Atom's own modals whenever the tab key is pressed, which was an odd nonstandard behavior, IMO.

This also automates returning focus to the element that had focus before the modal was shown, something up until now had to be implemented in every modal in Atom.

This likely breaks a few contracts for existing Atom packages that create modals, but I've found this doesn't conflict behaviorally with well-behaved modals like the command palette which implement their own focus management (which can be removed if this lands).

Released under CC0.


wbinnssmith:
Just fixed this up and made it opt-in as an additional option when creating a panel, as `autoFocus: true`. This way we don't break existing packages and tests for core atom modals don't break before I get around to opting them in.

I also documented this so it shows up in the API docs for `atom.workspace.addModalPanel`.

Let me know what you think!

iolsen:
@wbinnssmith this looks like a good change to us. Can you resolve the conflicts, we'll do some testing and get it merged.

wbinnssmith:
Rebased! :)

ungb:
@wbinnssmith I was able to test this out by setting autoFocus: true on command palette in node modules and saw that tabbing doesn't close the command palette, but tabbing also doesn't seem to do anything which is probably the expected behavior in this case.  

Since this is opt-in instead of opt-out the developer has to set the autoFocus flag to even see the behavior I'm going to merge this in.

PR #15165, opened 2017-08-04
Recognize Windows cmd or powershell environment in updateProcessEnv

### LINKED ISSUE #14986: Atom doesn't reflect changes in environment for subsequent invocations of `atom .`
### Prerequisites

### Description

I'm running Atom on Windows (with command prompt), so I'm not sure if the issue is reproducible on other platforms. All invocations of `atom .` seem to reflect the state of the environment from the first time I invoked atom. Moreover, when using WSL (Bash on Windows), atom seems to completely ignore any environment variables I've set in the shell beforehand. This might be an unrelated issue, but I've ruled out using WSL as a workaround.

### Steps to Reproduce

1. In one command prompt window, `set FOO=bar`
2. In the same command prompt window, `atom .`
3. In another command prompt window, `set FOO=foo`
4, In this other command prompt window, `atom .`
5. Print `process.env` in the Javascript console for both atom windows.

**Expected behavior:** `FOO` in the first window should be equal to `bar` and `FOO` in the second window should be equal to `foo`.

**Actual behavior:** Both windows report `FOO` as `bar`.

**Reproduces how often:** Every time.

### Versions

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

Windows 10 Build 15063.332 x64

### Additional Information

I've noticed related closed issues or merged pull requests, but they seem to focus more on OS X. Whatever the fix was in the past, at the very least it doesn't work on Windows.

segevfiner:
### Description of the Change

Windows doesn't have `PWD` in the environment passed to processes from the shell. cmd has `PROMPT` and PowerShell has `PSModulePath`, so also check for those to determine if we were launched from proper shell environment. At least that's what I think this condition is meant to check for.

### Alternate Designs

Maybe a different check is more appropriate.

It's a bug in the core functionality of updating the environment from the shell on Windows.

### Benefits

The environment will be updated when using the `atom` command from the shell like it does on other operating systems.

### Possible Drawbacks

If this starts updating the environment on cases that it shouldn't.

### Applicable Issues

Fixes #14986

### DIFF
--- a/spec/update-process-env-spec.js
+++ b/spec/update-process-env-spec.js
@@ -36,7 +36,7 @@ describe('updateProcessEnv(launchEnv)', function () {
   })
 
   describe('when the launch environment appears to come from a shell', function () {
-    it('updates process.env to match the launch environment', async function () {
+    it('updates process.env to match the launch environment because PWD is set', async function () {
       process.env = {
         WILL_BE_DELETED: 'hi',
         NODE_ENV: 'the-node-env',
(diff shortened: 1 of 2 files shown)


segevfiner:
ping @50Wliu 

This PR has been sitting here for quite a while...

rsese:
Sorry for the delay @segevfiner - the team talked about this one and were wondering even though it's a small change if you think there's any test that could be added?

segevfiner:
Added some specs. They are unit tests, so someone might want to verify that this actually fixes the problem when you are really launching Atom from cmd or powershell.

daviwil (approved):
Looks good to me, thanks a lot @segevfiner!

PR #15191, opened 2017-08-08
Update github to 0.4.2

smashwilson:
Doing this in a PR because it broke snapshotting, likely because 0.4.1 changes the way that dependencies are de-duped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "exception-reporting": "0.41.4",
     "find-and-replace": "0.209.5",
     "fuzzy-finder": "1.5.8",
-    "github": "0.4.1",
+    "github": "0.4.2",
     "git-diff": "1.3.6",
     "go-to-line": "0.32.1",
     "grammar-selector": "0.49.5",
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "exception-reporting": "0.41.4",
     "find-and-replace": "0.209.5",
     "fuzzy-finder": "1.5.8",
-    "github": "0.4.0",
+    "github": "0.4.1",
     "git-diff": "1.3.6",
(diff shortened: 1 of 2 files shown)


smashwilson:
Looks like this depends on atom/electron-link#6 because we aren't transpiling `async/await` away any more. :arrow_up: :arrow_up: :arrow_up:

smashwilson:
Hmm:

```
Verifying if snapshot can be executed via `mksnapshot`
/Users/distiller/atom/out/startup.js:48387
        async activate() {
              ^^^^^^^^
SyntaxError: Unexpected identifier
  at Object.exports.runInNewContext (vm.js:71:16)
  at electronLink.then (/Users/distiller/atom/script/lib/generate-startup-snapshot.js:79:8)
```

@as-cii, any idea what could cause this with snapshotting? Do we not support async/await in snapshot scripts, even though it works natively in Electron... ?

smashwilson:
> @as-cii, any idea what could cause this with snapshotting? Do we not support async/await in snapshot scripts, even though it works natively in Electron... ?

As a follow-up here: my understanding is that `vm.runInNewContext()` is used as a quick syntax check on the electron-link modified `snapshot.js` source, which uses the existing node that's running `script/build`. On CI that's a pre-async/await node. @BinaryMuse suggested using an electron process with`ELECTRON_RUN_AS_NODE` to do the check with the actual electron v8 version. Think that'd be okay, or is there a chance that the `electron-mksnapshot` would fail when that passed... ?

as-cii:
> Think that'd be okay, or is there a chance that the electron-mksnapshot would fail when that passed... ?

Yes, that's a really good idea: I think the version we use to perform the check should match the version of Node Atom ships with. 👍

PR #15208, opened 2017-08-09
Separate tests and installer creation on AppVeyor

smashwilson:
Let's try using a build matrix to split installer creation into a different job than the linting and tests. Hopefully this will help us sneak builds in under the AppVeyor time cutoff.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,8 +36,7 @@ install:
 
 build_script:
   - CD %APPVEYOR_BUILD_FOLDER%
-  - |
-    IF [%TASK%]==[installer] (
+  - IF [%TASK%]==[installer] (
       SET BUILD_ARG=--create-windows-installer
     ) ELSE (
       SET BUILD_ARG=
@@ -45,12 +44,11 @@ build_script:
   - script\build.cmd --code-sign --compress-artifacts %BUILD_ARG%
 
 test_script:
-  - |
-    IF [%TASK%]==[test] (
-      script\lint.cmd
+  - IF [%TASK%]==[test] (
+      script\lint.cmd &&
       script\test.cmd
     ) ELSE (
-      ECHO
+      ECHO Skipping tests
     )
 
 deploy: off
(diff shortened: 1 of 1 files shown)


smashwilson:
@damieng: Here's that build-matrix PR I was alluding to if you want to have a look. On non-release branches the "installer" rows will pass as basically no-ops so we don't eat build queue positions unnecessarily; if you want to see a build where the installer was actually built, [this](https://ci.appveyor.com/project/Atom/atom/build/7560) is the last one before I reinstated the release-branches-only flag.

appveyor.yml
```diff
@@ -34,14 +36,24 @@ install:
...
   - CD %APPVEYOR_BUILD_FOLDER%
-  - script\build.cmd --code-sign --compress-artifacts
+  - IF [%TASK%]==[installer] (
+      IF [%APPVEYOR_REPO_BRANCH:~-9%]==[-releases] (
+        script\build.cmd --code-sign --compress-artifacts --create-windows-installer
+      ) ELSE (
+        ECHO Skipping installer and Atom build on non-release branch
+      )
+    ) ELSE (
+      ECHO Skipping installer build on non-installer build matrix row &&
```
damieng: Not sure why we have a && here.

damieng:
One minor note in a comment but otherwise looks great, thanks!

appveyor.yml
smashwilson: AppVeyor (or YAML?) smashes these [into a single line](https://ci.appveyor.com/project/Atom/atom/build/7566/job/pivaa9st2davrmpa#L428) before passing it off to cmd, so the `&&` is necessary for it to still be valid:

```
IF [%TASK%]==[installer] ( IF [%APPVEYOR_REPO_BRANCH:~-9%]==[-releases] ( script\build.cmd --code-sign --compress-artifacts --create-windows-installer ) ELSE ( ECHO Skipping installer and Atom build on non-release branch ) ) ELSE ( ECHO Skipping installer build on non-installer build matrix row && script\build.cmd --code-sign --compress-artifacts )
```

smashwilson:
Cool, thanks for checking it over! 🙇

appveyor.yml
damieng: Oh, one more thing... you might want to check out and base it on 1.19-releases. I had to split this block into a new .cmd file to avoid weird line endings with appveyor

damieng: Oh, && to avoid the line endings problem, now I see :)

PR #15214, opened 2017-08-09
Update language-javascript

winstliu:
Hopefully specs don't fail~


50Wliu:
Failure seems to be a flaky test from atom/github.

Issue #15194, opened 2017-08-08
1.20.0-beta0 unable to load on Windows


Positive07:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

Note that clearing the state would fix this issue, but the issue should still be fixed in some other way.

### Description

Can't open projects from the previous version of Atom Beta

### Steps to Reproduce

1. Install 1.19 beta 7
2. Open a project, some tabs, and the About page (maybe the Settings View too)
3. Update to 1.20 beta 0 automatically (Squirrel should download the update and you need to Restart and Install once it's downloaded)
4. Reopen Atom once the update completes (the error should happen automatically)
5. If the error doesn't appear, open a project from `File > Reopen Project`

Note that if you manually add a project with `Add Project Folder` or open a project you had already opened in 1.20 the issue doesn't appear

**Expected behavior:** Projects should open in tree view and tabs should be restored

**Actual behavior:** Instead, the error pop-ups and no editor is opened

**Reproduces how often:** Always

### Versions

This happened when updating to 1.20.0 beta0 on Windows 10 x64

```
> apm --version

apm  1.18.3
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.0.windows.1
visual studio 2015
```

### Additional Information
Stack trace:
```
Uncaught (in promise) TypeError: Cannot read property 'node_modules\about\styles\about.less' of undefined
    at ThemeManager.module.exports.ThemeManager.loadLessStylesheet (C:\Users\nano0\AppData\Local\atom\app-1.20.0-beta0\resources\app\src\theme-manager.js:294:65)
    at ThemeManager.module.exports.ThemeManager.loadStylesheet (C:\Users\nano0\AppData\Local\atom\app-1.20.0-beta0\resources\app\src\theme-manager.js:271:27)
[...]

PR #15218, opened 2017-08-09
Debug missing snapshot auxiliary data

smashwilson:
Let's do some in-CI debugging to see where our snapshot auxiliary data (notable, the Less compile cache) is getting dropped.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,6 +57,8 @@ artifacts:
     name: atom-delta.nupkg
   - path: out\atom-*-full.nupkg
     name: atom-full.nupkg
+  - path: out\startup.js
+    name: startup.js
 
 cache:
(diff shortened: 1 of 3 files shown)


jasonrudolph:
Dropping in a cross-reference to https://github.com/atom/atom/issues/15194#issuecomment-321132078. We're hoping that this PR shines some light on the root cause of that issue.

smashwilson:
@jasonrudolph btw, this is where the auxiliary data is injected into the snapshot script:

https://github.com/atom/electron-link/blob/84cf71315806588e79ed11a269d1ca3880e30bf9/src/generate-snapshot-script.js#L90-L97

jasonrudolph:
@smashwilson: Should we close this now that https://github.com/atom/atom/issues/15194 is resolved?

smashwilson:
Yes, yes we should :+1:

Issue #15225, opened 2017-08-10
Buffer rewrite with beautify completely corrupts mixed-EOL files.


garretwilson:
I haven't proven this is the culprit, but all signs point to the [new editor backend in v1.19.0](https://github.com/atom/atom/releases/tag/v1.19.0) (or another "improvement" in v1.19.0) getting confused when there are multiple types of line endings in a document and **completely corrupting the file** when you run https://github.com/Glavin001/atom-beautify on it.

I'm using Atom 1.19.0 ia32 with atom-beautify 0.30.4 on Windows 10 Pro 64-bit.

I'm attaching a zipped file that contains mixed LF and CRLF file. (Yes, it's horrible, but it's out of my control; see the end of this message for the back story.)

Earlier editions of Atom would should the LF line endings using a different symbol, but _this version of Atom shows the LF line endings as if they were CRLF endings_ when I turn on invisibles. (The line ending notification at the bottom also says CRLF when the type is really mixed.) This is what tips me off that there is an underlying problem in the buffers, thinking that all the line endings are the same --- and when I went and saw the release notes that Atom updated it buffer implementation, that made it almost certain.

Install the `atom-beautify` package and open the attached archive. (Don't copy and paste, or that will normalize the line endings.) Make sure the `.jsbeatifyrc` file is in the same directory. Then run `atom-beautify` on it. It will turn this:

```html
    <figure> <figcaption>Examples of local time classes.</figcaption>
      <pre class="line-numbers"><code class="language-java">//someone's date of birth: May 30, 1970
final LocalDate dateOfBirth = LocalDate.of(1980, Month.MAY, 30);
//someone's birthday: May 30
final MonthDay birthday = MonthDay.of(Month.MAY, 30);

//a typical end of a workday: 5:00 p.m.
final LocalTime endOfWorkday = LocalTime.of(17, 0);

//current date and time
final LocalDateTime = LocalDateTime.now();</code></pre>
    </figure>
```

into this:

```html
		<figcaption>Examples of local time classes.</figcaption>
		<pre class="line-numbers"><code class="language-java">//someone's date of birth: May 30, 1970
    <p>Local dates and times are relative or <q>floating</q>; they do not provide sufficient information to determine precisely when an event occurred or will occur. Recording the date and time of a birth, for instance, must include the time zone. A journal entry as well would normally indicate the time zone in order to fix the exact moment in an absolute way worldwide: <q>December 11, 2016 10:53 a.m. America/Los Angeles</q>.</p>
[...]

PR #15228, opened 2017-08-10
Upgrade text-buffer for setTextViaDiff fix

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/15225

@garretwilson This fixes *a* bug, which may be the cause of the problem you were seeing with `atom-beautify`, but I'm not yet positive.

I'm unfortunately going to be away from work until Monday evening, so I wanted to get this fix to you before I leave so you would have a chance to test it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "service-hub": "^0.7.4",
     "sinon": "1.17.4",
     "temp": "^0.8.3",
-    "text-buffer": "13.0.10",
+    "text-buffer": "13.0.11",
     "typescript-simple": "1.0.0",
     "underscore-plus": "^1.6.6",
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I need to go to bed, but once the windows CI build completes, you should be able to download a file called `atom-windows.zip` from [this page](https://ci.appveyor.com/project/Atom/atom/build/7583/job/oaovxbc6fv4sy64t/artifacts).

maxbrunsfeld:
[Here](https://ci.appveyor.com/api/buildjobs/oaovxbc6fv4sy64t/artifacts/out%2Fatom-windows.zip) is a direct download link.

PR #15240, opened 2017-08-10
Create, update and destroy highlights manually

### LINKED ISSUE #15185: [from v1.19] highlight decoration node can be re-rendered which cause keyframe animation restart.
### Prerequisites

### Description

This issue started from v1.19 which use new editor-rendering with etch.

CC @as-cii 

#### Short summary

From v1.19, highlight decoration node is re-rendered in it's lifetime which cause keyframe animation unwillingly re-started.
This was not issue in v1.18 where highlight decoration node was stable in it's lifetime.

See, each removal of brackets-matcher's highlight cause fadeout animation re-started.
Which is because highlight node of `class=t9md-test-flash` is re-created at that timing.

decoration-node-unstable2

### Detail

When highlight decoration is created by `editor.decorateMarker`, corresponding `<div class="highlights">` node is created at top of LinesComponents.
Child nodes of this div represent actual decoration style.

- Till v1.18 these nodes are stable in its lifetime.
- From v1.19 with new editor-rendering, these nodes can be re-rendered during it's lifetime.

This is not problem when highlight class is simply set `background:` color.
But it's become problem when highlight class specify animation.
Lets' say

- I have set decoration which `class=flash-anime` which will be destroyed after 1.0 sec
- `flash-anime` css class have keyframe animation which fadeout flash during 1.0 sec.
- When Atom re-render(re-create) this decoration node within 1.0 sec, it cause re-start fadeout keyframe animation.
- This issue is more likely to happen when brackets-matcher involved, which create/remove decorations and it cause existing highlight node in the same LinesComponents **re-rendered**.

GIF: Animation in 10sec is separated into three because of re-render of highlight node.

untitled_ ___github_atom-vim-mode-plus

GIF: style recalculation highlight decoration re-created which cause animation restart.

untitled_ ___github_atom-vim-mode-plus

### Steps to Reproduce

I can reproduce this issue with safe-mode.

1. Prepare sample.txt and custom-style.

- sample.txt

`{a: 1}`

- style.less

```less
@keyframes t9md-test-animation {
  from { background-color: #5b91c8; }
  to { background-color: transparent; }
}

atom-text-editor .t9md-test-flash .region {
  animation-name: t9md-test-animation;
  animation-fill-mode: forwards;
  animation-duration: 10.0s;
  animation-iteration-count: 1;
}
```

2. Open sample.txt, place cursor at first line.
3. Paste following code in chrome-dev-tools console.

This code does.
- highlight inner range of `a: 1`(inner range of curly brackets).
- marker will destroyed after 10sec.
- highlight class is `t9md-test-flash` which fadeout background color in 10sec.

```javascript
(function flashreproduceIssue() {
  const cursorPosition = editor.getCursorBufferPosition()
  const scanRange = editor.bufferRangeForBufferRow(cursorPosition.row)
  editor.scanInBufferRange(/{.*?}/, scanRange, ({range}) => {
    const innerRange = range.translate([0, +1], [0, -1])
    const marker = editor.markBufferRange(innerRange)
    editor.decorateMarker(marker, {type: "highlight", class: "t9md-test-flash"})
    setTimeout(function () {
      marker.destroy()
      console.log("destroyed");
    }, 10000);
  })
})()
```

4. Within 10sec, move cursor left/right to cause add/remove brackets-macher's highlight.

And see how keyframe animation is re-started because of re-creation of highlight node.

Node used for highlight decoration is stabled(not re-rendered).

Node used for highlight decoration is re-rendered, cause keyframe animation re-started.

100%.

### Versions

```
Atom    : 1.19.0-beta7
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Additional Information

I use fadeout flash effect with keyframe animation a lot in [vim-mode-plus](https://atom.io/packages/vim-mode-plus) to give feedback to user.
E.g. when yank inner range of brackets, flash inner bracket range.

From v.1.19 cursor move cause brackets-matcher's highlights add/removal which cause vmp's flash effect re-started unwillingly.

as-cii:
Fixes https://github.com/atom/atom/issues/15185.

Etch's reconciliation routine causes elements to be sometimes re-ordered. In order to move an element, however, Etch needs to first detach it from the DOM and then re-append it at the right location.

This behavior is unacceptable for highlight decorations because it could re-start CSS animations on a certain highlight decoration when a completely different one is added or removed.

Even though we are still interested in restructuring etch's reconciliation logic to prevent unwanted re-orderings, with this commit we are switching to a custom routine to create/update/remove highlight decorations that prevents unnecessary moves and, as a result, fixes the undesired behavior described above.

/cc: @nathansobo @t9md

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1524,6 +1524,27 @@ describe('TextEditorComponent', () => {
       await setScrollTop(component, component.getLineHeight() * 3)
       expect(element.querySelectorAll('.highlight.a').length).toBe(0)
     })
+
+    it('does not move existing highlights when adding or removing other highlight decorations (regression)', async () => {
+      const {component, element, editor} = buildComponent()
+
+      const marker1 = editor.markScreenRange([[1, 6], [1, 10]])
+      editor.decorateMarker(marker1, {type: 'highlight', class: 'a'})
+      await component.getNextUpdatePromise()
+      const marker1Region = element.querySelector('.highlight.a')
+      expect(Array.from(marker1Region.parentElement.children).indexOf(marker1Region)).toBe(0)
+
+      const marker2 = editor.markScreenRange([[1, 2], [1, 4]])
+      editor.decorateMarker(marker2, {type: 'highlight', class: 'b'})
+      await component.getNextUpdatePromise()
+      const marker2Region = element.querySelector('.highlight.b')
+      expect(Array.from(marker1Region.parentElement.children).indexOf(marker1Region)).toBe(0)
+      expect(Array.from(marker2Region.parentElement.children).indexOf(marker2Region)).toBe(1)
+
+      marker2.destroy()
+      await component.getNextUpdatePromise()
+      expect(Array.from(marker1Region.parentElement.children).indexOf(marker1Region)).toBe(0)
+    })
   })
 
   describe('overlay decorations', () => {
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -3417,8 +3417,10 @@ class CursorsAndInputComponent {
 
 class LinesTileComponent {
   constructor (props) {
+    this.highlightComponentsByKey = new Map()
     this.props = props
     etch.initialize(this)
+    this.updateHighlights()
     this.createLines()
     this.updateBlockDecorations({}, props)
   }
@@ -3432,13 +3434,22 @@ class LinesTileComponent {
         this.updateLines(oldProps, newProps)
         this.updateBlockDecorations(oldProps, newProps)
       }
+      this.updateHighlights()
     }
   }
 
   destroy () {
+    this.highlightComponentsByKey.forEach((highlightComponent) => {
+      highlightComponent.destroy()
+    })
+    this.highlightComponentsByKey.clear()
+
     for (let i = 0; i < this.lineComponents.length; i++) {
       this.lineComponents[i].destroy()
     }
+    this.lineComponents.length = 0
+
+    return etch.destroy(this)
   }
 
   render () {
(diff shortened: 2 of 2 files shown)


nathansobo:
🙇  Thanks for taking this on @as-cii. I think this should be be hot-fixed to 1.20 first and then to 1.19 once we use if for a bit.

t9md:
Thanks you for fixing this very quickly!

nathansobo:
@as-cii I'm not seeing this on 1.19-releases and 1.20-releases but it's marked as cherry-picked.

as-cii:
> I'm not seeing this on 1.19-releases and 1.20-releases but it's marked as cherry-picked.

I have cherry-picked it to `1.20-releases` so that we can release a new beta version, test it for a while and then put it on `1.19-releases`, as you suggested on https://github.com/atom/atom/pull/15240#issuecomment-321593891. Should we put this back to "In Progress" until we cherry-pick it to stable as well?

nathansobo:
Oh, you're right. Sorry for the noise @as-cii.

as-cii:
@iolsen: this should go into `1.19-releases` after we test it for a while on beta, as mentioned in https://github.com/atom/atom/pull/15240#issuecomment-321871228 and https://github.com/atom/atom/pull/15240#issuecomment-321593891. How can we make sure we don't forget to do so? Should we move this back to **In Progress**?

iolsen:
> Should we move this back to In Progress?

Yeah let's do that. (I just did that.) GitHub Projects don't allow a card appearing twice on a board so this is the best we can do. I've added a note, too.

nathansobo:
Worst case scenario beta goes to stable in 2 weeks.

PR #15264, opened 2017-08-11
Ensure Pane.destroyItem always returns a promise

### LINKED ISSUE #15157: Pane.destroyItem() can return a non-boolean value
`Pane.destroyItem()` claims to return a promise of a boolean value. In the force case, or if the user does not need to save, we instead return the result of `item.destroy?()`.

I'll open a pull request to address this.

https://github.com/atom/atom/blob/master/src/pane.coffee#L629

cc @matthewwithanm @hansonw

captbaritone:
Fixes #15157

### DIFF
--- a/spec/pane-spec.js
+++ b/spec/pane-spec.js
@@ -551,10 +551,11 @@ describe('Pane', () => {
             itemURI = 'test'
             confirm.andReturn(0)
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks for the fix, @captbaritone!

PR #15265, opened 2017-08-11
Work around incorrect data on `compositionupdate` events in Chrome 56

### LINKED ISSUE #14911: 1.19.0-beta2: Cursor stucks at the first letter when using macOS Chinese IME

nathansobo:
Fixes #14911

The `compositionupdate` event is associated with a `data` property that is supposed to contain the full in-progress state of the composed string. Due to a [bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=677050) that has been resolved as of Electron 1.7, that field only contains the most recently typed character in Electron 1.6.

This PR attempts to work around this on Chrome 56 by reading the contents of the hidden input field directly rather than relying on the incorrect `data` property. Since `compositionupdate` events seem to be fired prior to the contents of the input field changing, I perform the read on `process.nextTick` after the browser has a chance to update the field.

One extra wrinkle is I now *clear* the input field on `compositionstart` events so that previous input doesn't pollute the previewed composition. The only keystrokes that could potentially pollute the preview are spaces, since we use `preventDefault` in our `textInput` event handler to prevent every other character from actually being inserted. If we `preventDefault` on space however, it causes the browser's default behavior of scrolling on space, so we have to let that one through. It's okay though, because we're clearing it out when the composition starts.

/cc @ungb @Ben3eeE


Ben3eeE:
I tested this using Japanese IME on Windows 10 and following the instructions on https://msdn.microsoft.com/en-us/library/windows/desktop/ee418266(v=vs.85).aspx this seems to fix the issue. I have never used IME before and don't understand japanese but it looks correct 🙂 

/cc: @50Wliu @ungb Can you test this with other languages?

ungb:
I'm testing now.  Mainly following repros in https://github.com/atom/atom/issues/14911.  I was able to see the issue where only one character comes up one by one.  I'll also test japanese as well, but it seems to impact all languages with IME

nathansobo:
Looks like the tests are failing. Will try to get this working tomorrow.

ungb:
Tested this on mac, was able to see it not working on 1.19 and fixed with this PR.  I tested Japanese and Chinese(simplified)

see gif:
before:
beforeime


After:
ime


There's still a few test failures that need to be address if these open up some other regression.

as-cii:
In ca183dd I have addressed a small wrinkle introduced by this that was causing Atom to insert an extra entry in the undo stack. Namely, the insertion of the preview on `process.nextTick` could fire after `textInput` has already been fired or, in other words, when the composition has already ended. That commit adds a guard that ensures composition is still in progress before attempting to display the IME preview.

b0d6783 takes care of making the test suite green again, adding also coverage for the IME behavior observed on Chrome 56 and on other Chrome versions where the bug does not exist. Those specs test only the accented character menu behavior, but since that feature works via IME I think they provide a good safety net to ensure this does not regress in the future. We can also add a specific test for CJK input if we think it would be valuable.

Ben3eeE:
LGTM 🚢

nathansobo:
Thanks for pushing this over the line @as-cii ❤️

Issue #15166, opened 2017-08-04
A text cursor is used when the cursor is on the scroll bars in the editor [1.20.0 from git]


segevfiner:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

A text cursor is used when the cursor is on the scroll bars in the editor. It used to show the default cursor in Atom 1.18.0 at least.

### Steps to Reproduce

1. Mouse over a scroll bar in the editor.

**Expected behavior:** The default cursor appears.

**Actual behavior:** The text cursor remains.

**Reproduces how often:** 100%

### Versions

```
Windows Version 10.0.15063 x64

Atom    : 1.20.0-dev-7d6bd2a6b
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.3
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.3.windows.1
visual studio 2015
```

#### Atom 1.18.0
![atom-cursor-scroll-bar-1_18](https://user-images.githubusercontent.com/24731903/28968681-e91d999c-7929-11e7-93e2-804ffb66eeb0.gif)

#### Atom 1.20.0-dev-7d6bd2a6b
![atom-cursor-scroll-bar](https://user-images.githubusercontent.com/24731903/28968552-2b930a38-7929-11e7-8a72-68ea96034030.gif)

segevfiner:
Also in 1.19.0.

segevfiner:
This was caused by https://github.com/atom/atom/pull/13880.

Just adding a `cursor` style to the dummy scrollbars is not enough since they are now 20px wide (larger than the scrollbars of the default one dark theme) which will result in the cursor changing too early and changing it to 10px makes the scrollbars disappear when the window is at certain sizes. Adding the style will also result in the cursor changing when the scrollbars are hidden.

[...]

Issue #14661, opened 2017-05-27
Scrollbar div still takes up space even if there's no scrollbar


50Wliu:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Even when there is no need for a scrollbar, the scrollbar div is still rendered and has a non-zero size.
![useless-scrollbar-div](https://cloud.githubusercontent.com/assets/2766036/26524304/e751c272-42fc-11e7-94dd-38d406fc6497.png)

### Steps to Reproduce

1. Make sure that scrollbars are enabled (macOS only?)
2. Open a file whose contents can fit horizontally or vertically on the screen without a scrollbar.
3. Through devtools, notice that there are scrollbar divs taking up 20px.

**Expected behavior:** No scrollbar div if there's no scrollbar

**Actual behavior:** Scrollbar div is there

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.19.0-dev-9e64abef3
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

nathansobo:
@50Wliu I always render the dummy divs because we use them to measure the width of scrollbars for use in various calculations. Maybe we can apply a CSS class to the scrollbar divs when the scrollbars are actually visible and only change the cursor then?

nathansobo:
Basically, can you clarify the incorrect behavior that you're seeing as a result of the scrollbars always being rendered? Is it just the cursor styling or might it be other things as well?

50Wliu:
Right now, the only thing I've noticed as a direct effect of this is the incorrect cursor styling.  Though I suppose that any theme that adds extra styles to `.horizontal/vertical-scrollbar` will have those styles appear even when visually, there is no scrollbar.  For example:
![image](https://user-images.githubusercontent.com/2766036/29344479-1600fd02-8206-11e7-9ac5-26d16938655a.png)

[...]

PR #15275, opened 2017-08-12
Use default cursor on dummy scrollbars and make them 15px wide/tall

as-cii:
Fixes https://github.com/atom/atom/issues/15166.

This restores scrollbars to adopt the styling they had prior to #13880, as you can see in the below CSS excerpt from Atom 1.18.0.

https://github.com/atom/atom/blob/783cda8642e42e842a7aac761f267e151f3db324/static/text-editor-light.less#L170-L198

@nathansobo: do you recall the reason why changed the styling to `20px`? I couldn't find an explanation on the commit that introduced them, so I went ahead and created this pull-request.

Also, /cc: @ungb and @Ben3eeE for help on testing this and making sure it restores the original behavior.

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2851,20 +2851,22 @@ class DummyScrollbarComponent {
       outerStyle.bottom = 0
       outerStyle.left = 0
       outerStyle.right = right + 'px'
-      outerStyle.height = '20px'
+      outerStyle.height = '15px'
       outerStyle.overflowY = 'hidden'
       outerStyle.overflowX = this.props.forceScrollbarVisible ? 'scroll' : 'auto'
-      innerStyle.height = '20px'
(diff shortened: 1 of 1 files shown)


Ben3eeE:
LGTM 🚢

nathansobo:
I honestly don't recall. Sorry. Let's just roll with this.

pfitzseb:
Would you consider making this dependant on the scrollbar's visibility? 
For very small editors the (invisible) scrollbars take up most of the editor's area, hence the cursor will be incorrect for most of it:
image

Ben3eeE:
@pfitzseb We have an open issue for scrollbar divs taking up space even when not visible, https://github.com/atom/atom/issues/14661. Can you please comment there? Thank you :bowing_man:

pfitzseb:
Thanks for pointing me there :)

PR #15276, opened 2017-08-12
Remove unused methods to translate legacy menu items

captbaritone:
Searching the code base and the GitHub repository shows no usage or
documentation for these two methods.

From what I can surmise, the call to `convertLegacyItemsBySelector` was removed
in https://github.com/atom/atom/commit/cc4ee926993edc479a1624b269d0c159608f1399
by @thomasjo but these methods just weren't cleaned up.


nathansobo:
I'll take your word for it.

PR #15277, opened 2017-08-12
Add accelerator indicators to context menus

captbaritone:
Electron allows us to pass an "accelerator" property for each menu item, which
is renders to the right of the menu item. We were already adding these for the
application level menus.

This pull request adds the accelerator property to regular context menu items,
which should make it easier for people to discover/recall key mappings for
actions which they usually take via a context menu.

### Before

screen shot 2017-08-12 at 2 56 58 pm

### After

screen shot 2017-08-12 at 2 56 36 pm

### DIFF
--- a/spec/context-menu-manager-spec.coffee
+++ b/spec/context-menu-manager-spec.coffee
@@ -231,3 +231,75 @@ describe "ContextMenuManager", ->
             }
           ]
         ])
+
+  describe "::templateForEvent(target)", ->
+    [keymaps, item] = []
+
+    beforeEach ->
+      keymaps = atom.keymaps.add('source', {
+        '.child': {
+          'ctrl-a': 'test:my-command',
+          'shift-b': 'test:my-other-command'
+        }
+      })
+      item = {
+        label: 'My Command',
+        command: 'test:my-command',
+        submenu: [
+          {
+            label: 'My Other Command',
+            command: 'test:my-other-command',
+          }
+        ]
+      }
+      contextMenu.add('.parent': [item])
+
+    afterEach ->
+      keymaps.dispose()
+
+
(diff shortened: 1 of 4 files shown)


nathansobo:
Thanks for this contribution. This will be a nice improvement.

nathansobo:
@captbaritone I think I've found a bug or unexpected behavior.

If you right click an unstaged file in the Git panel, the "Discard" option has an accelerator of `ctrl-backspace`. This really *should* be bound, but as far as I can tell it is only bound in the diff view itself. Could you investigate this?

captbaritone:
@nathansobo Sure. I'll take a look.

captbaritone:
@nathansobo From what I can tell, this is because the [key bindings](https://github.com/atom/github/blob/master/keymaps/git.cson#L30-L32) are attached to a node that either can't actually get keyboard focus, or only can under very specific circumstances. By adding `tabIndex='-1'` to the `.github-FilePatchListView` node, I was able to focus the node by clicking within it, and then trigger the keyboard shortcut.

The behavior you are seeing in the diff view is actually a separate binding of the same set of keys to a different selector and different (but similar) command: https://github.com/atom/github/blob/master/keymaps/git.cson#L45-L46

captbaritone:
@nathansobo So, I think the real bug here is with the keybinding provided by the github package. That said, I wonder if there is something better we can do inside this accelerator logic to avoid surfacing similarly broken logic. For example, rather than using `event.target` when calling `findKeyBinding`, we could instead use `document.activeElement`. I'm no expert in how focus is handled, but a manual test seems to indicate that:

1. `document.activeElement` is updated on right-click, before this code is called.
2. Making this change correctly filters out this particular broken key binding.
3. This change does not seem to break other accelerators in any obvious way.

PR #15285, opened 2017-08-13
Fix IME composition preview for new editors

nathansobo:
Composition checkpoints are actually *numbers*, and the first checkpoint starts at `0`, which was previously being detected as falsy. The fix is to null check the composition checkpoint explicitly.

/cc @Ben3eeE @as-cii

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1670,7 +1670,7 @@ class TextEditorComponent {
 
     if (this.getChromeVersion() === 56) {
       process.nextTick(() => {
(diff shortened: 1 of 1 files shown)


Ben3eeE:
@nathansobo Thanks for the quick fix. I verified this on Windows 7 and it seems to work :100: :heart: 

It is a bit hard for me to test this because I have never used IME text input before so I don't really have a good understanding of the expected result. I used the gifs in the issues and compared the results to other text editors to verify that our result is the same.

PR #15296, opened 2017-08-14
Prompt user only once when quitting/restarting and canceling save dialog

### LINKED ISSUE #15238: Pressing "Cancel" on an unsaved buffer dialog will not cancel a restart
### Prerequisites

### Description

**Pressing "Cancel" on the unsaved buffer dialog will not cancel a restart**

Atom requests a restart in some situations, such as when a plugin has been updated. If one has an unsaved buffer open in a project-less window, after pressing "restart", the normal unsaved file dialog is presented with, on macOS at least, "Save", "Don't Save", and "Cancel" buttons. In this scenario, "Save" and "Don't Save" work as expected, but pressing "Cancel" will immediately re-present the same dialog. This can be especially problematic when the unsaved buffer is in another tab in the same window as one cannot switch tabs to inspect the file and determine if it should be saved or not.

### Steps to Reproduce

1. Open a new project-less window and buffer
2. Type in some content
3. Trigger a restart by either updating plugins or opening a developer console and entering `atom.restartApplication()`
4. Press "Cancel"

**Expected behavior:**
The restart operation should be canceled, returning me to the state I was in before I pressed "restart".

**Actual behavior:**
I am immediately presented with the same unsaved buffer dialog

**Reproduces how often:**
Every time

### Versions

Atom    : 1.19.0
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
macOS: 10.12.6

as-cii:
Fixes #15238.

I think this slipped through during the refactoring performed in dc32018. With this pull-request we are fixing the issue and adding a new main process test to exercise this behavior and ensure we don't regress in the future.

/cc: @Ben3eeE @ungb for testing this, and @nathansobo @maxbrunsfeld for 👀.

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -18,7 +18,7 @@ describe('AtomApplication', function () {
 
   beforeEach(function () {
     originalAppQuit = electron.app.quit
-    originalShowMessageBox = electron.dialog
+    originalShowMessageBox = electron.dialog.showMessageBox
     mockElectronAppQuit()
     originalAtomHome = process.env.ATOM_HOME
     process.env.ATOM_HOME = makeTempDir('atom-home')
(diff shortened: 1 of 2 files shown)


nathansobo:
The new test looks good. I like the addition of a new promise to make things predictable.

PR #15299, opened 2017-08-14
Ignore clicks on block decorations

### LINKED ISSUE #15282: Block decoration scrolls instead of registering mouse clicks
### Prerequisites

### Description

When interacting with a block decoration near the bottom of the text editor, the decoration scrolls up instead of registering the click event:

scroll

This issue came up with the [Hydrogen](https://github.com/nteract/hydrogen) package which uses block decorations to display interactive outputs with "clickable" elements. But this is a general problem also present in safe mode.

### Steps to Reproduce

1. Open a file with at least 10 to 20 lines of code
2. Run the following code in the dev tools:
```js
function createBlock(row) {
  const div = document.createElement("div");
  const item = document.createElement("span");
  item.innerText = "[block decoration]\n".repeat(10);
  const button = document.createElement("button");
  button.innerText = "Click me!"
  button.addEventListener('click', () => {
    alert("CLICK!")
  });
  div.appendChild(item);
  div.appendChild(button);
  const editor = atom.workspace.getActiveTextEditor();
  const marker = editor.markBufferPosition([row, 0]);
  editor.decorateMarker(marker, { type: "block", item: div });
}

createBlock(8)
```
3. Scroll the block decoration to the bottom of the text editor and try to click the button.

**Expected behavior:** The button click is registered and a alert box pops up.

**Actual behavior:** The editor scrolls up and the click isn't registered. A second click on the button is necessary to show the pop up.

### Versions

macOS 10.12.6:
- Atom 1.20.0-beta0
- Atom 1.19.0

as-cii:
Fixes #15282.

Previously, clicking on a block decoration to interact with it would cause the editor to scroll to the line next to it. This is inconvenient, especially if the decoration was designed to be interactive and contained buttons or links. If the decoration was close to the bottom of the screen, clicking on a button inside of it would make the editor scroll down and abort the click.

This behavior regressed during the editor rendering layer rewrite and with this pull-request we are restoring the original behavior by simply ignoring clicks that land on block decorations.

@ungb @Ben3eeE: can you help me test this out? I think it should work as expected, but it'd be nice to make sure the behavior looks consistent.

/cc: @nathansobo @lgeiger

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2153,6 +2153,39 @@ describe('TextEditorComponent', () => {
       expect(component.refs.blockDecorationMeasurementArea.offsetWidth).toBe(component.getScrollWidth())
     })
 
+    it('does not change the cursor position when clicking on a block decoration', async () => {
+      const {editor, component} = buildComponent()
+
+      const decorationElement = document.createElement('div')
+      decorationElement.textContent = 'Parent'
+      const childElement = document.createElement('div')
+      childElement.textContent = 'Child'
+      decorationElement.appendChild(childElement)
+      const marker = editor.markScreenPosition([4, 0])
+      editor.decorateMarker(marker, {type: 'block', item: decorationElement})
+      await component.getNextUpdatePromise()
+
+      const decorationElementClientRect = decorationElement.getBoundingClientRect()
+      component.didMouseDownOnContent({
+        target: decorationElement,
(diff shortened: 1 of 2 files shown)


nathansobo:
Nice!

lgeiger:
Thanks for the fast fix! I can confirm this fixes #15282.

Keep up the great work 💯

Issue #15227, opened 2017-08-10
Uncaught TypeError: Cannot read property 'refs' of undefined


mh61503891:
### Steps to reproduce

1. Launch Atom

### Abstract

An expected behavior of IME on TexEdit shows "あいうえお" with an bold underbar.

![image](https://user-images.githubusercontent.com/706275/29155923-09f3a206-7dda-11e7-8523-7e21b9ad001d.png)

But, an actual behavior of IME on Atom shows "お" without an bold underbar and does not show "あいうえ".

![image](https://user-images.githubusercontent.com/706275/29155954-2deec118-7dda-11e7-9056-3df59fe7d9fe.png)

The IME' settings on my macOS is Japanese.

**Atom**: 1.19.0 x64
**Electron**: 1.6.9
**OS**: Mac OS X 10.11.6
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Cannot read property 'refs' of undefined

```
At /Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:1455

TypeError: Cannot read property 'refs' of undefined
    at TextEditorComponent.didFocus (/Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:1455:58)
    at HTMLElement.addEventListener (/Applications/Atom.app/Contents/Resources/app/src/text-editor-element.js:36:67)
    at HTMLElement.PaneElement.activeItemChanged (/Applications/Atom.app/Contents/Resources/app/src/pane-element.js:158:25)
    at Function.module.exports.Emitter.simpleDispatch (/Applications/Atom.app/Contents/Resources/app/node_modules/event-kit/lib/emitter.js:25:20)
    at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app/node_modules/event-kit/lib/emitter.js:141:34)
    at Pane.module.exports.Pane.setActiveItem (/Applications/Atom.app/Contents/Resources/app/src/pane.js:337:28)
    at Pane.module.exports.Pane.addItem (/Applications/Atom.app/Contents/Resources/app/src/pane.js:560:20)
    at Pane.module.exports.Pane.activateItem (/Applications/Atom.app/Contents/Resources/app/src/pane.js:489:20)
    at Workspace.<anonymous> (/Applications/Atom.app/Contents/Resources/app/src/workspace.js:1161:20)
    at Generator.next (<anonymous>)
[...]

PR #15302, opened 2017-08-14
Always render hidden input

nathansobo:
Fixes https://github.com/atom/notifications/issues/167
Fixes https://github.com/atom/atom/issues/15227

This avoids mysterious timing issues in which the editor gets a `focus` event in a state where `isVisible` returns false. If we always render the hidden input, we can always focus it. We still don't quite understand why things get into this state, but with this change we don't really need to.

### DIFF
--- a/spec/text-editor-element-spec.js
+++ b/spec/text-editor-element-spec.js
@@ -203,6 +203,22 @@ describe('TextEditorElement', () => {
         expect(document.activeElement).toBe(element.querySelector('input'))
       })
     })
+
+    describe('if focused when invisible due to a zero height and width', () => {
+      it('focuses the hidden input and does not throw an exception', () => {
+        const parentElement = document.createElement('div')
+        parentElement.style.position = 'absolute'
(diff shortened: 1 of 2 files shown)


as-cii:
While looking at something else, I stumbled upon an [error on BugSnag](https://app.bugsnag.com/atom/atom/errors?filters[event.since][0]=30d&filters[error.status][0]=open&filters[event.message][0][value]=updatecursorblinksync&filters[event.message][0][type]=eq&sort=users) that seems to be related to this. The error is slightly different in that case, as the component tries to call `updateCursorBlinkSync` on `hiddenInput` during `updateSync`.

Overall, even if the number of events is pretty high (which I suspect is due to the cursor blinking every 400ms), it seems like ~70-80 users are currently affected by this. Given that the fix didn't make it to stable/beta during yesterday's release, I think we should consider hotfixing again at some point next week.

/cc: @iolsen @nathansobo

nathansobo:
Yeah, I was thinking we'd do one or two more releases this week.

PR #15313, opened 2017-08-15
Ensure extra document updates are not scheduled during `updateSync`

### LINKED ISSUE #15311: control-mousewheel to zoom font may throw uncaught type error
Version: 1.19.1 x64 on mac

Steps to reproduce:
Start atom in safe mode: `atom --safe`

Download and open this markdown file: https://raw.githubusercontent.com/twineworks/tweakflow/master/site/content/reference.md

Scroll down to about line 200.

Press ctrl and use your mouse wheel to zoom the font.

An exception appears:
```
/Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:2171
Hide Stack Trace
TypeError: Cannot read property 'firstChild' of undefined
    at TextEditorComponent.measureLongestLineWidth (/Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:2171:112)
    at TextEditorComponent.measureContentDuringUpdateSync (/Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:383:16)
    at scheduler.readDocument (/Applications/Atom.app/Contents/Resources/app/src/text-editor-component.js:270:20)
    at ViewRegistry.module.exports.ViewRegistry.performDocumentUpdate (/Applications/Atom.app/Contents/Resources/app/src/view-registry.js:191:15)
```

After that atom becomes unstable and many actions generate follow up errors.

Thanks to @Ben3eeE for helping to find a usable repro path!

as-cii:
Fixes https://github.com/atom/atom/issues/15311.

When changing the editor styles, we force the component to remeasure character dimensions. If they change, each line's height could change too, causing the current scroll top position to not match the viewport the user was observing. Thus, when detecting a line height change, we try to show users the area of the screen they were looking prior to tweaking the font size.

In trying to maintain the aforementioned logical position, however, we were mistakenly scheduling a new update before actually finishing the current one. This was problematic because if the first update detected that the longest screen line changed and such line was off-screen, it would try to render it. Before having the chance to measure it, though, the new update would kick in and delete the new longest screen line node, because it assumed it had already been measured. Finally, when
`measureContentDuringUpdateSync` fired, it would notice that the longest screen line node did not exist and throw an exception as a result.

This pull-request changes the `updateSync` method to set the `updateScheduled` flag only before returning control to the caller, as opposed to doing so at the beginning. This prevents calls to `scheduleUpdate` made in `updateSync` from scheduling new unwanted updates.

@Ben3eeE @ungb: can you confirm this fixes the aforementioned issue? Thanks!

/cc: @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -3610,6 +3610,23 @@ describe('TextEditorComponent', () => {
       element.style.display = 'none'
       await component.getNextUpdatePromise()
     })
+
+    it('does not throw an exception when the editor is soft-wrapped and changing the font size changes also the longest screen line', async () => {
+      const {component, element, editor} = buildComponent({rowsPerTile: 3, autoHeight: false})
+      editor.setText(
+        'Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do\n' +
+        'eiusmod tempor incididunt ut labore et dolore magna' +
+        'aliqua. Ut enim ad minim veniam, quis nostrud exercitation'
+      )
+      editor.setSoftWrapped(true)
+      await setEditorHeightInLines(component, 2)
+      await setEditorWidthInCharacters(component, 56)
+      await setScrollTop(component, 3 * component.getLineHeight())
+
+      element.style.fontSize = '20px'
+      TextEditor.didUpdateStyles()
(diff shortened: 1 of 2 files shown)


Ben3eeE:
Verified this with @ungb :shipit:

PR #15314, opened 2017-08-15
Suppress text input for default-prevented keydown events

### LINKED ISSUE #15189: key-mappings with `alt` don't suppress Mac OS composition characters anymore (atom-1.19)
### Prerequisites

### Description

Shortcuts with alt don't suppress Mac OS composition characters.

### Steps to Reproduce

With `ABC - Extended` keyboard layout:

1. `Open Your Keymap`
2. Add the following keymaps:
```
# general
'atom-text-editor':
  # use jkl; vim-like when pressing alt (word by word, paragraph by paragraph)
  # or alt (letter by letter)
  'alt-h': 'core:move-left'
  'alt-j': 'core:move-down'
  'alt-k': 'core:move-up'
  'alt-l': 'core:move-right'
```
3. save
4. try to use them

With `US` keyboard layout /cc https://github.com/atom/atom/issues/15189#issuecomment-321503095.

A GIF from atom-1.18

atom-1 18

A GIF from atom-1.19

atom-1 19

Always.

### Versions

* **Atom:**       1.19.0-beta7
* **Atom-Shell:** 
* **OS:**         Mac OS X 10.12.5
* **Misc**
    * apm  1.18.2
    * npm  3.10.10
    * node 6.9.5 x64
    * python 2.7.10
    * git 2.11.0

### Additional Information

Probably only happens on Mac OS.

as-cii:
Fixes https://github.com/atom/atom/issues/15189.

This is a continuation of #15266. In that pull-request we managed to prevent IME previews from being displayed in the editor when the originating `keydown` event was default-prevented. However, it was still possible for IME input to make it through the previous workarounds, thus triggering the `textInput` event and showing unwanted text.

Pressing another key that would complete the in-progress IME input would, in fact, first replace `this.lastKeydown` and then trigger the `textInput` event. In the handling of that event we would detect `this.lastKeydown` as "non-default-prevented" and therefore mistakenly insert the IME text.

With this pull-request we are adopting a different strategy to mitigate the issue. When receiving the wrong `compositionupdate` event we will first disable the hidden input and then re-enable it on the next tick. Disabling the input causes the in-progress IME input to be aborted and the browser to never fire `textInput` nor `compositionupdate` events anymore after that.

The only downside of this approach is that the hidden input also loses focus, but we transfer it back to it as soon as the next tick of the event loop is served and the input has been re-enabled.

@ungb @Ben3eeE: can you give this a spin and see if it fixes the problem? Thanks.

/cc: @nathansobo

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1569,10 +1569,6 @@ class TextEditorComponent {
   }
 
   didTextInput (event) {
-    // Workaround for Chromium not preventing composition events when
-    // preventDefault is called on the keydown event that precipitated them.
-    if (this.lastKeydown && this.lastKeydown.defaultPrevented) return
-
     if (!this.isInputEnabled()) return
 
     event.stopPropagation()
@@ -1665,7 +1661,16 @@ class TextEditorComponent {
   didCompositionUpdate (event) {
     // Workaround for Chromium not preventing composition events when
     // preventDefault is called on the keydown event that precipitated them.
-    if (this.lastKeydown && this.lastKeydown.defaultPrevented) return
+    if (this.lastKeydown && this.lastKeydown.defaultPrevented) {
+      this.getHiddenInput().disabled = true
+      process.nextTick(() => {
+        // Disabling the hidden input makes it lose focus as well, so we have to
+        // re-enable and re-focus it.
+        this.getHiddenInput().disabled = false
+        this.getHiddenInput().focus()
+      })
+      return
+    }
 
     if (this.getChromeVersion() === 56) {
       process.nextTick(() => {


nathansobo:
One thing makes me nervous here: Is it ever possible to receive two keystrokes in the same tick of the event loop? I'm worried that we might eat the next keystroke, but I'm not sure it could ever come quickly enough after a default-prevented composition event to actually happen, assuming it's even possible.

For now this looks like a pretty good solution until we can upgrade to a Chromium that doesn't have this bug.

Ben3eeE:
The current steps to reproduce are macOS specific so I can't easily verify this fix.
Do we have some steps to reproduce on Windows or is it a good idea to just give this a round of random testing on Windows to check for keystroke problems it might introduce anyway?

nathansobo:
@Ben3eeE If you're on a keyboard like "US International" aren't there a bunch of characters accessed via `alt-` modifiers? When one of those combinations matches a keybinding we want to avoid any input based on that keystroke. On the Mac at least, this regressed in 1.19.

nathansobo:
Going ahead and merging this. Still could use testing though @ungb @Ben3eeE

PR #15316, opened 2017-08-15
Let's see how long our AppVeyor builds are taking

smashwilson:
/cc @iolsen


smashwilson:
It looks like we're at about 30 minutes for `TASK=installer` builds on AppVeyor, and 20 minutes of that is `script/build` prep. That's _much_ better than the 1 hour 15 minute installer build times we were seeing before. I feel like it's still beneficial to have the build matrix, though.

PR #15317, opened 2017-08-15
Move cursors container inside lines container

as-cii:
Fixes #15241.

This will ensure that applying any style that changes the location of the lines container will also correctly position the cursors.

@Ben3eeE @ungb: can you help me test this? I can't reproduce the issue anymore by using the `no-caffeine-syntax` theme or the personal stylesheet provided in https://github.com/atom/atom/issues/15241#issuecomment-322509336, but it'd be nice if you could give this a spin and ensure we didn't break anything else accidentally.

/cc: @nathansobo

PR #15322, opened 2017-08-15
Move placeholder text inside lines div

nathansobo:
Following up on #15317, we need to move the placeholder text inside the lines div so that it aligns with the cursor in the event that themes apply padding or margin to the lines div.

/cc @as-cii @Ben3eeE


nathansobo:
@as-cii can you cherry pick this tomorrow

PR #15327, opened 2017-08-16
Base context menu accelerators on activeElement

captbaritone:
Addresses issue pointed by out @nathansobo in #15277 where keybindings
for unfocusable nodes were being surfaced as accelerator indicators in
context menus.

When you right click in the DOM, your focus goes to the first focusable
ancestor of your click target. This change uses the ancestor that you
are actually focused on when looking for avaliable key bindings rather
than using the event target directly. This ensures that any surfaced key
bindings are actually reachable.


nathansobo:
Thanks very much for the quick followup @captbaritone. I look forward to working with you more!

captbaritone:
Thanks for _your_ quick follow up! Excited to be able to help with the project.

PR #15338, opened 2017-08-16
Hide off-screen lines when we render them for measurement

nathansobo:
Previously, we weren't hiding off-screen lines that we rendered for measurement. That could cause flickering artifacts like the one below when re-triggering measurements, as well as the cursors div being pushed down momentarily.

/cc @as-cii @Ben3eeE


alikkille (approved):

Ben3eeE:
Tested this with @ungb and it looks good  👍 💯

PR #15345, opened 2017-08-16
Round return values of getMaxScrollTop, getScrollHeight

### LINKED ISSUE #15343: onDidChangeScrollTop fired when no scrollTop change / can cause infinite loop
### Prerequisites

### Description

`editor.onDidChangeScrollTop` can be fired with same scrollTop value.
And to make matters worse, it can cause infinite loop when registered callback call `component.getScrollTop` and `component.setScrollTop` indirectly.

Why this happen is here

- `editor.component.getScrollTop` update `component.scrollTop` value without **Math.round**
- guard(so that not fire event with same scrollTop) done in `component.setScrollTop` is compared with **rounded** value.

In my package, atom-narrow hit infinite loop because of this behavior(bug IMO).

What was [happened](https://github.com/t9md/atom-narrow/issues/239) is here.

0. In `editor.setScrollTop`
  scrollTop updated from xxx to `522`.
1. `editor.onDidChangeScrollTop` fired with (scrollTop = `522`)
2. In onDidChangeScrollTop callback, call `markerLayer.clear()` which markerLayer is used for text decoration.
3. `editor.onDidChangeScrollTop` fired again(goto step 1, this is infinite loop)
  - In some place, `editor.component.getScrollTop` is accessed which update `component.scrollTop` to `521.65625`)
  `this.scrollTop = Math.min(this.getMaxScrollTop(), this.scrollTop)` = Math.min(521.65625, 522)
  - During etch's rendering, `component.setScrollTop` is called with `548.65625`.
    - result of `scrollTop = Math.round(Math.max(0, Math.min(this.getMaxScrollTop(), scrollTop)))`
    - result of `scrollTop = Math.round(Math.max(0, Math.min(521.65625, 548.65625)))`
    - result of `scrollTop = Math.round(Math.max(0, 521.65625))`
    - result of `scrollTop = 522`
    - scrollTop = `522`
  - `did-change-scroll-top` fired with `522` (`522(rounded) !== 521.65625(<- this.scrollTop updated via getScrollTop)`)

### Steps to Reproduce

1. Make sure `editor.scrollPastEnd` is `false` (default).
When `editor.scrollPastEnd` is `true` this issue doesn't happen.

2. Prepare `style.less` bellow. paste&call `reproduceIssue()` on chrome-dev-console.

- style.less

```less
@import "octicon-mixins";
.scrolltop-issue-container {
  padding: @component-padding/3 @component-padding;
}

atom-text-editor {
  .scrolltop-issue-text-hl {
    color: @syntax-color-function;
  }
}
```

```javascript
function reproduceIssue() {
  atom.workspace.open().then(editor => {
    // Render 40 lines of text to make editor scrollable.
    const lines = []
    for (let i = 1; i < 40; i++) {
      lines.push(i)
    }
    editor.setText(lines.join("\n"))

    editor.setCursorBufferPosition([0, 0])

    // To make editor.component.maxScrollTop() have fractional portion(decimal point).
    const item = document.createElement("div")
    item.className = "scrolltop-issue-container"
    item.innerText = "[block decoration]"
    editor.decorateMarker(editor.markBufferPosition([0, 0]), {type: "block", item})

    // prepare decoration layer for text highlight
    const markerLayer = editor.addMarkerLayer()
    editor.decorateMarkerLayer(markerLayer, {type: "text", class: "scrolltop-issue-text-hl"})

    let oldScrollTop
    editor.onDidChangeScrollTop(newScrollTop => {
      console.log("fired")

      // This guard is necessary to prevent infinite loop.
      // if (oldScrollTop === newScrollTop) {
      //   return
      // }

      // Refresh highlight
      markerLayer.clear()
      editor.scan(/1|3|5|7|9/g, ({range}) => markerLayer.markBufferRange(range))

      oldScrollTop = newScrollTop
    })
  })
}
reproduceIssue()
```

3. Move cursor down by `core:move-down`
When last row was appeared, editor freeze by infinite loop on `onDidChangeScrollTop` event. Need to force close.

`editor.onDidChangeScrollTop` should not be fired when not changed.

`editor.onDidChangeScrollTop` is emitted with same scrollTop value which possibly cause infinite loop(depends on registered callback).

100%

### Versions

OS
```
ProductName:	Mac OS X
ProductVersion:	10.12.5
```

Atom(All stable/beta/master affects).

- Stable
```
Atom    : 1.19.1
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

- Beta
```
atom-beta  --version
Atom    : 1.20.0-beta2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

And master branch

### Additional Information

Original issue I hit this issue.
https://github.com/t9md/atom-narrow/issues/239

This guard 

https://github.com/atom/atom/blob/15e3fbaa07f90174ab84d35a4281956b54128d12/src/text-editor-component.js#L2713

should be `scrollTop !== Math.round(this.scrollTop)`.

Since calling `getScrollTop` update `this.scrollTop` with non rounded value.

https://github.com/atom/atom/blob/15e3fbaa07f90174ab84d35a4281956b54128d12/src/text-editor-component.js#L2707

nathansobo:
Fixes #15343

If the current `scrollTop`/`scrollLeft` value exceeds the `maxScrollTop` / `maxScrollLeft`, we reassign the values to the max on read. This allows us to behave correctly when the editor has changed in dimensions or the content has changed since the scroll position was previously assigned.

Prior to this PR, we were applying `Math.round` to the scroll positions when assigning them but **not** rounding the values when reassigning them to the max.

In situations where the content height contains fractional pixels, this can lead to a fractional-pixel scroll position when scrolled all the way down/right which causes blurry text. This can also cause unexpected firing of `onDidChangeScrollTop` when the value isn't actually changing due to a decision to notify listeners based on a rounded-up value not matching a fractional clipped value, as was the case in #15343.

/cc @t9md @Ben3eeE

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -205,6 +205,23 @@ describe('TextEditorComponent', () => {
       expect(component.getFirstVisibleRow()).toBe(editor.getScreenLineCount() + 1)
     })
 
+    it('does not fire onDidChangeScrollTop listeners when assigning the same maximal value and the content height has fractional pixels (regression)', async () => {
+      const {component, element, editor} = buildComponent({autoHeight: false, autoWidth: false})
+      await setEditorHeightInLines(component, 3)
+
+      // Force a fractional content height with a block decoration
+      const item = document.createElement("div")
+      item.style.height = '10.6px'
+      editor.decorateMarker(editor.markBufferPosition([0, 0]), {type: "block", item})
+      await component.getNextUpdatePromise()
+
+      component.setScrollTop(Infinity)
+      element.onDidChangeScrollTop((newScrollTop) => {
+        throw new Error('Scroll top should not have changed')
+      })
+      component.setScrollTop(component.getScrollTop())
+    })
+
     it('gives the line number tiles an explicit width and height so their layout can be strictly contained', async () => {
       const {component, element, editor} = buildComponent({rowsPerTile: 3})
 
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2728,7 +2728,7 @@ class TextEditorComponent {
   }
 
   getMaxScrollTop () {
-    return Math.max(0, this.getScrollHeight() - this.getScrollContainerClientHeight())
+    return Math.round(Math.max(0, this.getScrollHeight() - this.getScrollContainerClientHeight()))
   }
 
   getScrollBottom () {
@@ -2756,7 +2756,7 @@ class TextEditorComponent {
   }
 
   getMaxScrollLeft () {
-    return Math.max(0, this.getScrollWidth() - this.getScrollContainerClientWidth())
+    return Math.round(Math.max(0, this.getScrollWidth() - this.getScrollContainerClientWidth()))
   }
 
   getScrollRight () {


nathansobo:
@t9md Could you give this a try to make sure it fixes your infinite loop issue?

t9md:
@nathansobo Confirmed fix with

- reproduction procedure described in #15343
- original issue in my atom-narrow package.

No infinite loop for both now, thanks!

Tested with latest build on cd8233028d584b262f2fd1bb00318de89052caf8 commit.


```
[130] ~/Downloads% ./Atom\ 8.app/Contents/MacOS/Atom  --version
Atom    : 1.21.0-dev-cd82330
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

PR #15346, opened 2017-08-16
Suppress default-prevented composition events on compositionstart instead of compositionupdate

### LINKED ISSUE #15344: Composition characters mess up insertion point
### Prerequisites

### Description

I'm trying out edge (586b818b) since #15314 was merged. I have a command bound to alt-n (a composing diacritic in the macOS US keyboard layout) that moves the cursor. Under Atom 0.19 the keystroke would run the command and insert the diacritic. On this edge version, it only runs the command, which is awesome. 🎉 

BUT, it seems like some internal buffer state then gets messed up. If after the alt-n I move the cursor with the arrows, and then insert a character, the cursor seems to jump back to where the alt-n left me, and the character gets inserted there. If I use a command that isn't a composition character, like `alt-o`, then the problem does not occur.

#### Keymap

```
'atom-text-editor':
  'alt-n': 'core:move-down'
  'alt-o': 'core:move-down'
```

1. Add the keymap above to your ~/.atom/keymap.cson
2. Have two lines of text in your buffer:
```
Foo
Bar
```

3. Have the cursor at the beginning of file(before `F` character) Type `alt-n`, then arrow right, then insert a character. Observe that the character is inserted at the beginning of the line (where the `alt-n` left you).
4. Repeat 2-3 using `alt-o` instead of `alt-n`. Observe the insertion happens where the cursor is, as expected.

**Expected behavior:** Character insertion following `alt-n` inserts the character where the cursor is.

**Actual behavior:** Character insertion following `alt-n` inserts the character at the wrong place (seems to be where the `alt-n` leaves you).

### Versions

Atom    : 1.21.0-dev-586b818ba
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
OS: macOS Sierra 10.12.6

nathansobo:
Fixes #15344

In #15314, @as-cii found a workaround to composition events still firing after a default-prevented keydown event. Unfortunately, applying this workaround in the `compositionupdate` handler instead of the `compositionstart` handler meant we still created a composition checkpoint. Then, on the next input event, we detected the presence of this checkpoint and reverted to it, causing the cursor to move as in #15344.

This PR moves the same logic from #15314 to `compositionstart` and avoids creation of a composition checkpoint following a default-prevented keydown event.

/cc @as-cii @ungb @Ben3eeE

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1671,17 +1671,6 @@ class TextEditorComponent {
   //   4. compositionend fired
   //   5. textInput fired; event.data == the completion string
   didCompositionStart () {
-    if (this.getChromeVersion() === 56) {
-      this.getHiddenInput().value = ''
-    }
-
-    this.compositionCheckpoint = this.props.model.createCheckpoint()
-    if (this.accentedCharacterMenuIsOpen) {
-      this.props.model.selectLeft()
-    }
-  }
-
-  didCompositionUpdate (event) {
     // Workaround for Chromium not preventing composition events when
     // preventDefault is called on the keydown event that precipitated them.
     if (this.lastKeydown && this.lastKeydown.defaultPrevented) {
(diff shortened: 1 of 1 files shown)


ungb:
Was able to verify this fixes issue https://github.com/atom/atom/issues/15344!  LGTM 💯 ✨

oggy:
Wonderful, working for me too. Thanks for the quick fix @nathansobo !

Issue #15247, opened 2017-08-10
Uncaught RangeError: Invalid array length


akaifi:
[Enter steps to reproduce:]

1. Updating atom from 1.10 to 1.19 and .....
2. I can not click update packages !!

**Atom**: 1.19.0 x64
**Electron**: 1.6.9
**OS**: "Fedora release
**Thrown From**: Atom Core

### Stack Trace

Uncaught RangeError: Invalid array length

```
At /usr/share/atom/resources/app/src/text-editor-component.js:2785

RangeError: Invalid array length
    at TextEditorComponent.populateVisibleTiles (/usr/share/atom/resources/app/src/text-editor-component.js:2785:45)
    at TextEditorComponent.updateSyncBeforeMeasuringContent (/usr/share/atom/resources/app/src/text-editor-component.js:363:16)
    at TextEditorComponent.updateSync (/usr/share/atom/resources/app/src/text-editor-component.js:266:16)
    at etch.getScheduler.updateDocument (/usr/share/atom/resources/app/src/text-editor-component.js:220:46)
    at ViewRegistry.module.exports.ViewRegistry.performDocumentUpdate (/usr/share/atom/resources/app/src/view-registry.js:187:15)
    at /usr/share/atom/resources/app/src/view-registry.js:3:65
```

### Commands

```
     -1:37.6.0 tree-view:show (atom-workspace.workspace.scrollbars-visible-always)
```

### Non-Core Packages

```
api-docs 0.0.6 
atom-autocomplete-php 0.25.6 
atom-beautify 0.30.4 
atom-runner 2.7.1 
atom-ternjs 0.18.3 
auto-detect-indentation 1.3.0 
autocomplete-modules 1.6.10 
autocomplete-php 0.3.7 
autocomplete-python 1.10.0 
busy-signal 1.4.3 
caniuse 0.9.2 
color-picker 2.2.5 
docblockr 0.11.0 
file-icons 2.1.10 
flex-tool-bar 0.12.0 
git-plus 7.9.3 
highlight-selected 0.13.1 
intentions 1.1.5 
linter 2.2.0 
linter-jsonlint 1.3.0 
linter-php 1.3.2 
linter-phpcs 1.6.7 
linter-phpmd 1.5.3 
linter-sass-lint 1.7.4 
linter-shellcheck 1.4.4 
linter-ui-default 1.6.4 
minimap 4.29.6 
minimap-highlight-selected 4.6.1 
minimap-pigments 0.2.2 
[...]

PR #15357, opened 2017-08-17
Ignore scroll requests to NaN, null or undefined positions

as-cii:
Fixes https://github.com/atom/atom/issues/15247

This was a guard we used to have in `TextEditorPresenter`, prior to the editor rendering layer rewrite:

https://github.com/atom/atom/blob/b580ff8642d4401892c730f6bd2e1b38bd00b3e5/src/text-editor-presenter.coffee#L723-L732

With this pull-request we are adding those guard clauses back, preventing `NaN`, `null` or `undefined` values that were mistakenly passed to the `setScrollTop` or `setScrollLeft` APIs from putting the editor into an inconsistent state.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1296,6 +1296,38 @@ describe('TextEditorComponent', () => {
     })
   })
 
+  describe('scrolling via the API', () => {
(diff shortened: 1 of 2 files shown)


Ben3eeE:
LGTM :ship:

PR #15359, opened 2017-08-17
Fix folding for lines that contain a comment

### LINKED ISSUE #15308: [Bug] Code folding does not apply consistently to JS files
### Prerequisites

### Description

Code folding does not apply consistently to JS files

### Steps to Reproduce

1. Edit a JS file
2. execute editor:fold-at-indent-level-1 (ctrl-k ctrl-1)

**Expected behavior:** All functions should be folded (as per previous versions)

**Actual behavior:** Some functions are folded, some are not

### Versions
Running Beta to see if this was also an issue in Beta (it is), but first saw this on the current production version.

Atom    : 1.20.0-beta2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

### Additional Information
Here is some sample code to use (custom JS layout used within an app)

The 2nd function doesn't get folded at all

```
pageOnLoadChecks : function() {
	var AuthorisedUserIDs = new Array('ddfdfd','dfdfdfd');

},

onChangeQuestion_SI_89847 : function(obj) { //handset colour 
	if(a == b){
		c = a;
	}

},
//apply current pricing until change date
pricingChanges : function() {

}
```

as-cii:
Fixes #15308

In attempting to optimize the performance of `isFoldableAtBufferRow` in 3c87b74, we mistakenly introduced a bug that caused lines that contained a comment but didn't start with one to not be foldable anymore.

With this commit we are restoring the previous behavior, thus only disabling folding for lines that start with a comment (ignoring leading whitespaces).

### DIFF
--- a/src/tokenized-line.coffee
+++ b/src/tokenized-line.coffee
@@ -1,5 +1,5 @@
 Token = require './token'
-CommentScopeRegex  = /(\b|\.)comment/
+CommentScopeRegex = /(\b|\.)comment/
 
 idCounter = 1
 
@@ -54,7 +54,15 @@ class TokenizedLine
         @isCommentLine = true
         return @isCommentLine
 
+    startIndex = 0
     for tag in @tags
+      # If we haven't encountered any comment scope when reading the first
+      # non-whitespace chunk of text, then we consider this as not being a
+      # comment line.
+      if tag > 0
+        break unless isWhitespaceOnly(@text.substr(startIndex, tag))
+        startIndex += tag
+
       if @isCommentOpenTag(tag)
         @isCommentLine = true
         return @isCommentLine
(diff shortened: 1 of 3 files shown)


src/tokenized-line.coffee
```diff
@@ -55,6 +55,11 @@ class TokenizedLine
         return @isCommentLine
 
     for tag in @tags
+      # If we haven't encountered any comment scope when reading the first
+      # non-whitespace chunk of text, then we consider this as not being a
+      # comment line.
+      break if tag > 0 and not isWhitespaceOnly(@text.substr(startIndex, tag))
```
Ben3eeE: startIndex is not defined

Ben3eeE:
LGTM :heart:

nathansobo:
⚡️

PR #15367, opened 2017-08-17
Set visibility: hidden on dummy scrollbars if native ones are invisible

### LINKED ISSUE #14661: Scrollbar div still takes up space even if there's no scrollbar

as-cii:
Fixes #14661

This prevents the cursor from unexpectedly changing when approaching the bottom/right corner of the editor with the mouse, even when no scrollbar is being shown.

/cc: @nathansobo @ungb @50Wliu @Ben3eeE

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -305,7 +305,9 @@ describe('TextEditorComponent', () => {
       expect(getVerticalScrollbarWidth(component)).toBeGreaterThan(0)
       expect(getHorizontalScrollbarHeight(component)).toBeGreaterThan(0)
       expect(verticalScrollbar.style.bottom).toBe(getVerticalScrollbarWidth(component) + 'px')
+      expect(verticalScrollbar.style.visibility).toBe('')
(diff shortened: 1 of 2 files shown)


Ben3eeE:
I did a very quick test of this and it looks like it fixes the issue. One thing I noticed is that the cursor changes to a pointer before you can use the scroll bar.

nathansobo:
Of course. Visibility hidden. Why am I not thinking of obvious stuff lately?

PR #15374, opened 2017-08-17
Don't assign default path for `showSaveDialog` unless it exists

### LINKED ISSUE #14732: "Save As" for new file doesn't default to a sensible directory location

kuychaco:
Fixes #14732

Problem: "Save As" for a new file would default to the previously used directory, instead of the open project folder in Atom. 

A regression was introduced where we were assigning the default path to be `undefined` when there was no file path for an item, as in the case of a brand new, unsaved file. When no default path is specified the operating system defaults to displaying the previously used directory.

The fix was to simply not set the default path in `Pane.prototype.saveItemAs` if the item has no path, and let `AtomWindow.prototype.showSaveDialog` set it based on the project paths, which happens later on down the code path.


50Wliu:
@kuychaco you might want to take a look at #15160, which uses a slightly different approach.

kuychaco:
Ah yes, thanks for pointing that PR out @50Wliu. I totally missed it somehow. 

The approach in #15160 addresses part of the problem, but won't account for if there are multiple folders open in the project and the desired default directory is anything other than the first folder. In this PR, we're effectively taking the same approach, but letting `AtomWindow` decide which folder to default to based on the project.

kuychaco:
Hmmm, turns out `AtomWindow`'s way of deciding is to simply take the first path in `loadSettings.initialPaths`, which seems to be the most recently added folder. It's not the most intelligent way of determining the default path when there are multiple folders in the project, but it seems to be what we were doing before the regression (reference https://github.com/atom/atom/commit/c6cae5b8fdfea982084b068a0b38a7c988c52f51). 

In any case, this fix as-is gets us to the expected behavior for a majority of the cases and restores the previous functionality. Any other improvements for handling multiple folders in a project can be handled in a separate PR.

nathansobo:
Nice work @kuychaco.

PR #15377, opened 2017-08-18
Async watchPath method

smashwilson:
Rather than synchronously returning a `PathWatcher` from the `watchPath` method and requiring users to await the `getStartPromise()` manually in test cases to avoid flaky tests, return a `Promise` that resolves to the `PathWatcher` after it's started.

I originally did it the other way for efficiency, but it's much harder to reason about, and accessing the start Promise in tests is often challenging. As evidence of this, it's the root cause of the flaky GitHub package test we're seeing in the atom/atom build right now... which _I_ wrote, against my own API 😇 

Along the way, I also fixed a subtle timing bug related to the event timing during the NativeWatcher re-attachment sequence when watchers are split or joined, and introduced a private `printWatchers()` method that's useful as a sanity check on the watchers that are alive at any given moment.

/cc @damieng as the only other current consumer of this API. If you're writing any tests that exercise `atom.project.onDidChangeFiles`, you'll need to await `atom.project.getWatcherPromise(projectPath)` before manipulating the filesystem and expecting the watcher to pick it up. I'd be happy to send PRs to fix any specs up myself if you point me to a repo that needs it.

/cc @nathansobo because this _should_ fix that GitHub package test flake you noticed yesterday.

/cc @maxbrunsfeld because this was his original suggestion and I should have listened to him to begin with :wink:

Remaining work: update the GitHub package to await the result of `watchPath`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "exception-reporting": "0.41.4",
     "find-and-replace": "0.209.5",
     "fuzzy-finder": "1.5.8",
-    "github": "0.4.2",
+    "github": "0.5.0",
     "git-diff": "1.3.6",
     "go-to-line": "0.32.1",
     "grammar-selector": "0.49.5",
--- a/spec/path-watcher-spec.js
+++ b/spec/path-watcher-spec.js
@@ -52,7 +52,7 @@ describe('watchPath', function () {
       const rootDir = await temp.mkdir('atom-fsmanager-test-')
 
       const watcher = await watchPath(rootDir, {}, () => {})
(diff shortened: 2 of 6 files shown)


nathansobo:
⚡️ I like this.

smashwilson:
Hilariously, I was [already awaiting the return value of `watchPath` in the GitHub package](https://github.com/atom/github/blob/7835bc46d9f8dec07b6072293b788866db191bff/lib/models/file-system-change-observer.js#L85). I think that made these a little less likely to flake because they'd at least yield the event loop and give the watchers a chance to start.

PR #15378, opened 2017-08-18
Drop use of GPU tiles for lines and line numbers

### LINKED ISSUE #15171: Fonts are rendered cropped
### Prerequisites

### Description

Some fonts are rendered cropped when line height set to 1.1 or below. It happens with Iosevka font, Source Code Pro and probably others.

### Steps to Reproduce

1. Just set line height setting in editor section to e.g. `1.1` or `1.0`.
2. Display some symbols with tall characters (especially brackets: [ ] ).

**Expected behavior:** 
28961886-2d607cf8-7904-11e7-9c93-cf2771c5334a

**Actual behavior:** 
28961877-283cb200-7904-11e7-84d8-419bbec939c3

### Versions

* 1.18.0
* 1.19.0-beta7

---

Added alternate steps to repro /cc https://github.com/atom/atom/issues/15171#issuecomment-322689734

as-cii:
Fixes #15171

At the cost of a very minimal reduction in layout performance, we gain reliable rendering at all line heights and don't have to worry about characters that extend above/below the footprint of a line.

@ungb @Ben3eeE @50Wliu: this should fix all the reported bugs of cropped text when using certain fonts or line heights. We have tested it on both Linux and macOS, but it would be nice if you could give this a spin too. Please, feel free to add more issues to the "Fixes" list in case you can think of other related bug reports. Thanks! ✨

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -3167,15 +3167,12 @@ class LineNumberGutterComponent {
         children[i] = $.div({
           key: rootComponent.idsByTileStartRow.get(tileStartRow),
           style: {
-            contain: 'strict',
-            overflow: 'hidden',
+            contain: 'layout style',
             position: 'absolute',
             top: 0,
-            height: ceilToPhysicalPixelBoundary(tileHeight) + 'px',
-            width: ceilToPhysicalPixelBoundary(width) + 'px',
-            willChange: 'transform',
-            transform: `translateY(${roundToPhysicalPixelBoundary(tileTop)}px)`,
-            backgroundColor: 'inherit'
+            height: tileHeight + 'px',
+            width: width + 'px',
+            transform: `translateY(${tileTop}px)`
           }
         }, ...tileChildren)
       }
@@ -3546,13 +3543,11 @@ class LinesTileComponent {
     return $.div(
       {
         style: {
-          contain: 'strict',
+          contain: 'layout style',
           position: 'absolute',
-          height: ceilToPhysicalPixelBoundary(height) + 'px',
-          width: ceilToPhysicalPixelBoundary(width) + 'px',
-          willChange: 'transform',
-          transform: `translateY(${roundToPhysicalPixelBoundary(top)}px)`,
-          backgroundColor: 'inherit'
+          height: height + 'px',
+          width: width + 'px',
+          transform: `translateY(${top}px)`
         }
       },
       $.div({
--- a/static/text-editor.less
+++ b/static/text-editor.less
@@ -86,7 +86,6 @@ atom-text-editor {
 
   .line {
     white-space: pre;
-    overflow: hidden;
     contain: layout;
 
     &.cursor-line .fold-marker::after {


src/text-editor-component.js
```diff
@@ -3546,13 +3543,11 @@ class LinesTileComponent {
...
           position: 'absolute',
-          height: ceilToPhysicalPixelBoundary(height) + 'px',
-          width: ceilToPhysicalPixelBoundary(width) + 'px',
-          willChange: 'transform',
-          transform: `translateY(${roundToPhysicalPixelBoundary(top)}px)`,
-          backgroundColor: 'inherit'
+          height: height + 'px',
+          width: width + 'px',
+          transform: `translateY(${top}px)`
         }
```
trusktr: @as-cii What part of this change causes the tile to no longer be on the GPU? I'm just curious, as CSS is so cryptic in this regard.

Also, you said this issue only affects some people. Why not make an option for it so for those people who don't have issues they can use the acceleration?

as-cii: > What part of this change causes the tile to no longer be on the GPU?

Removing the `will-change: transform` property will cause Chromium to not promote the div into its own GPU layer. There are also other properties that trigger this behavior, like `translate3d` or `translateZ`.

> And what do you mean by "tile"?

A tile is a logical container that groups tiles into chunks of an arbitrary length (in our case each tile contains at most 6 lines). They are just `div` elements that we create to reduce layout times, and have nothing to do with the [tiles created by Chromium's compositor](https://www.chromium.org/developers/design-documents/gpu-accelerated-compositing-in-chrome).

> Also, you said this issue only affects some people. Why not make an option for it so for those people who don't have issues they can use the acceleration?

GPU acceleration is still turned on for everyone because the container div that encapsulates the tiles uses `will-change: transform`. Making it an option wasn't really necessary because the added cost for not assigning a different GPU layer to each tile was negligible.

Hope this helps!

trusktr: @as-cii Thanks! How do you know using `will-change` and `translate` (not `translate3d`) will force GPU acceleration? Have you tested with `translate3d`?

The spec says:

> Different browsers can use the information from will-change in different ways, and even a single browser might use it in different ways at different time. For example, a browser that promotes elements to their own “GPU layer” when they have will-change: transform specified might avoid doing that when there are too many elements declaring that, to avoid exhausting GPU memory.

Are we sure that will-change + translate actually uses the GPU? Have you verified that in the Blink source?

---

Scrolling seems to be slow on my MacBook Pro that I just got in June, compared to what can easily be achieved with WebGL, but the graphics in the atom text area aren't _that_ complicated. F.e. I see like 5 tiles containing lines, and one huge tile that the five tiles are re-jiggered inside of while the huge outer one is "scrolled" using `translate` (not `translate3d`).

---

There's other ways to solve the clipping problems: instead of butting the inner tiles together on their edges, make them overlap just a little bit and give them padding, then fonts can stick out past the edges of content-box into the padding areas and still be rendered (and accelerated with `translate3d`).

---

Is there a specific reason to use `translate` and not `translate3d`?

trusktr: Thinking about it, if there's only one big tile with the five re-jiggering tiles inside of it, that's still much too simple to be slowed down when using `translate3d` from experiments I've done. Chrome can animate 100 elements smoothly using `translate3d` on my MacBook pro.

I do notice that the outer layer is _really tall_. If I take a guess, this might mean that the GPU may be doing tons of work at shading a bunch of empty pixels (because the inner re-jiggering tiles don't fill this huge outer layer), which might be a cause for slow-down.

By "inner re-jiggering tiles" I mean the `div` elements directly inside of the `.lines` element.

---

I just verified: a file with 422 lines results in an outer container that is almost `9000px` tall. This is the container whose `translate` property is being animated during scroll. I would prepare for the worst, and assume that the GPU is animating this HUGE area of pixels, and that the re-jiggering of the inner elements isn't helping much (though that concept will help if this outer-layer problem can be eliminated).

---

The "outer layer" is the element that contains the `.lines` element. This layer (and anything inside of it) should only be bigger than the viewport (it should probably be the total size of the inner re-jiggering tiles).

Then, when scrolling up (for example) the outer container would be translated on the Y axis by a negative value, but the Y translation value should not exceed a value that is roughly the size of one of the inner containers. When it gets close to that value, the outer container translation should be reset at the same moment that the inner tiles are re-jiggered, then the vertical scrolling would continue in the same way.

It would be similar to the Super Mario 64 method used in [scrolling the infinite stairs](https://www.youtube.com/watch?v=B-udfiFZcko).  There is only a finite set of stairs, and when the end is reached, the position of the stairs is reset so it appears as if the animation is infinite.

When I make new lines in Atom, the outer container gets bigger and bigger.

In Mario 64, the size of the number of steps would have had to be infinite based on Atom's same concept, and that would have obviously crashed N64.

The same Mario 64 concept can apply here: only make the text area (the steps are the re-jiggering elements, and the whole stair set is the outer container and the `.lines` container) slightly bigger than the viewport and keep resetting the position while scrolling, re-jiggering inside as needed to prevent re-drawing all the lines while scrolling.

So although the stair/step technique does exist in some sense here (because the re-jiggering inner elements) it may not help the actual scrolling because the scrolled surface is just HUGE!

---

My guess is that this might provide a nice performance improvement because the FPS would remain constant, as opposed to now where FPS will decrease as the number of lines gets bigger (unless Chrome has an optimization in place to prevent that, but that would only be an assumption that I'd rather not make).

---

Another approach, though it'd be much further off, would be to make your own text rendering with WebGL (using the same stair/step technique), with the rendering mechanism exposed as custom elements and easily-stylable. I'm working with these sorts of goals in mind over at github.com/trusktr/infamous. It's a ways off still. Similarly to infamous, [A-Frame](https://aframe.io) supplies devs with elements for making 3D scenes, but A-Frame currently has no intent to support anything other than 3D AR/VR (f.e. they don't make it easy to make 2D applications, whereas infamous is aimed to be general purpose for 2D/3D and VR/not-VR). A-Frame has docs too, infamous not yet, but I'm working on getting there!

trusktr: Sara Soueidan's article on `will-change` says

> It is worth mentioning here that only 3D transforms qualify for their own layer; 2D transforms don’t.

It's for Opera, but they use Blink. Maybe she knows the internals.

trusktr: Wow, will-change really can make a difference though. F.e. try with and without will-change while inspecting the bubbles here: http://www.bitlisten.com (it's not mine, those graphics are not my style!) Take profiles with and without will-change. In this case it is using `translate` (no 3d).

Anyways, with will-change, padding the boxes for overflow content, and the sizing of the outer container relative to the viewport size, the performance may be best.

trusktr: Ah, Sara's article does say

> If any non-initial value of a property would create a stacking context on the element, specifying that property in will-change will create a stacking context on the element.

So there _is_ a new stacking context, but based on what she said earlier it's still not on the GPU because it isn't `translate3d`.

PR #15380, opened 2017-08-18
Update README.md

marceloatie:
-Only one installation process for all rpm based distributions
-Changed "yum localinstall" and "dnf install" to just "rpm -i"
-Included distributions Linux Mint and openSUSE


ungb:
I was able to test these changes with the newly added steps and different distros.  Looks good!

PR #15383, opened 2017-08-18
[Commands] Add support for rich listener objects with first-class metadata

### LINKED ISSUE #15143: Support Command Metadata
Atom commands should be able to reference arbitrary metadata, along with a human-written displayName and a description of what to expect when a command is executed.

This way, tools like the command palette could surface something potentially friendlier than the `_.humanizeEventName` version of the name string of the command, which is more of an implementation detail.

Potential arbitrary metadata could be keywords (which could or could not surface themselves in search results), aliases, search weight, etc.

It's possible to retain and simply augment the existing command registration API by supporting a complex object everywhere a callback function might be expected:

```js
atom.commands.add(
  'atom-text-editor', 
  'user:insert-date',
  {
    displayName: 'Insert a really awesome date',
    description: 'Whereever your cursor is, it inserts the current date!',
    keywords: ['timestamp', 'time', 'today', 'now'],
    aliases: ['user:insert-now'],
    callback(e) {
      const editor = this.getModel();
      editor.insertText(new Date().toLocaleString());
    },
  },
  true // throwOnInvalidSelector
);
```

or

```js
atom.commands.add(
  'atom-text-editor', 
  {
    'user:insert-date': {
      displayName: 'Insert a really awesome date',
      description: 'Whereever your cursor is, it inserts the current date!',
      callback(e) {
        const editor = this.getModel();
        editor.insertText(new Date().toLocaleString());
      },
    },
    'user:some-other-command': {
      displayName: 'Does something really cool',
      description: 'Really, it\'s super awesome',
      callback(e) {
        editor = this.getModel();
        editor.insertText('yay!');
      },
    }
  },
  true // throwOnInvalidSelector
);
```

I'm not 100% of whether this metadata should be declared at registration time, and what the consequences are for the same command name registered multiple times with different targets (something I assume from looking at the api is supported?).

Another thought I had was either declaratively defining this metadata in `package.json`, or registering metadata bound to command names separately, in the same way keybindings are decoupled from commands:

```js
atom.commands.register({
  'user:insert-date': {
      displayName: 'Insert a really awesome date',
      description: 'Whereever your cursor is, it inserts the current date!',
      keywords: ['timestamp', 'time', 'today', 'now'],
      aliases: ['user:insert-now'],
    },
  },
)
```

cc @matthewwithanm @hansonw @captbaritone @jinface @karincurkowicz

wbinnssmith:
Fixes #15143 

This adds support for listener objects which, in addition to the existing callback listeners, can optionally provide a `displayName` ahead of time to avoid a potentially awkward humanized displayName.

Every object-based listener must implement `handleEvent()`, which plays the role of the callback fired when the command is dispatched.

We can support other first-class metadata, for example I'm preparing a pull request to the command palette to support displaying a command's `description` field as a second line below the `displayName` in search results. I even have an example of adding "tags" to a command to help capture keywords a user might use when searching for a command, kind of like how macOS shows "System Preferences" as a result for settings:

image

Here's an example of an object command listener:

```js
atom.commands.add('foo:command', document.body, {
  displayName: 'My awesome foo command',
  description: 'Here\'s some detail about my foo command. Try it out!',
  handleEvent() {
    alert('foo!');
  },
```

It should also be technically possible to attach these properties to a function callback. If we'd like to officially support this I'll add some tests to check:

```js
function myListener() {
  alert('foo!');
}

myListener.displayName = 'My Awesome Listener';
myListener.description = 'Does awesome stuff!';
```

I'll cross-link my command palette pull request which demonstrates the use of this API.

Released under CC0.

cc @matthewwithanm @nathansobo  @hansonw @jinface

### DIFF
--- a/spec/command-registry-spec.js
+++ b/spec/command-registry-spec.js
@@ -224,32 +224,35 @@ describe("CommandRegistry", () => {
       expect(addError.message).toContain(badSelector);
     });
 
-    it("throws an error when called with a non-function callback and selector target", () => {
+    it("throws an error when called with a null callback and selector target", () => {
       const badCallback = null;
-      let addError = null;
 
-      try {
+      expect(() => {
         registry.add('.selector', 'foo:bar', badCallback);
-      } catch (error) {
-        addError = error;
-      }
-      expect(addError.message).toContain("Can't register a command with non-function callback.");
+      }).toThrow(new Error('Cannot register a command with a null listener.'));
     });
 
-    it("throws an error when called with an non-function callback and object target", () => {
+    it("throws an error when called with an null callback and object target", () => {
       const badCallback = null;
-      let addError = null;
 
-      try {
+      expect(() => {
         registry.add(document.body, 'foo:bar', badCallback);
-      } catch (error) {
-        addError = error;
-      }
-      expect(addError.message).toContain("Can't register a command with non-function callback.");
+      }).toThrow(new Error('Cannot register a command with a null listener.'));
+    });
+
+    it("throws an error when called with an object listener without a handleEvent method", () => {
+      const badListener = {
+        title: 'a listener without a handleEvent callback',
+        description: 'this should throw an error'
+      };
+
+      expect(() => {
+        registry.add(document.body, 'foo:bar', badListener);
+      }).toThrow(new Error('Listener must be a callback function or an object with a handleEvent method.'));
     });
   });
 
-  describe("::findCommands({target})", () =>
+  describe("::findCommands({target})", () => {
     it("returns commands that can be invoked on the target or its ancestors", () => {
       registry.add('.parent', 'namespace:command-1', () => {});
       registry.add('.child', 'namespace:command-2', () => {});
@@ -268,8 +271,65 @@ describe("CommandRegistry", () => {
         {name: 'namespace:command-2', displayName: 'Namespace: Command 2'},
         {name: 'namespace:command-1', displayName: 'Namespace: Command 1'}
       ]);
-  })
-);
+    });
+
+    it("returns commands with manual displayNames if set in the listener", () => {
+      registry.add('.grandchild', 'namespace:command-1', () => {});
+      registry.add('.grandchild', 'namespace:command-2', {
+        displayName: 'Custom Command 2',
+        metadata: {
+          some: 'other',
+          object: 'data'
+        },
+        handleEvent() {}
+      });
+      registry.add('.grandchild', 'namespace:command-3', () => {});
+
+      const commands = registry.findCommands({target: grandchild});
+      expect(commands).toEqual([
+        {
+          displayName: 'Namespace: Command 1',
+          name: 'namespace:command-1'
+        },
+        {
+          displayName: 'Custom Command 2',
+          metadata: {
+            some : 'other',
+            object : 'data'
+          },
+          name: 'namespace:command-2'
+        },
+        {
+          displayName: 'Namespace: Command 3',
+          name: 'namespace:command-3'
+        }
+      ]);
+    });
+
+    it("ignores a `name` property if passed in registration object", () => {
+      registry.add('.grandchild', 'namespace:command-2', {
+        name: 'some:other:commandname',
+        displayName: 'Custom Command 2',
+        metadata: {
+          some: 'other',
+          object: 'data'
+        },
+        handleEvent() {}
+      });
+
+      const commands = registry.findCommands({target: grandchild});
+      expect(commands).toEqual([
+        {
+          displayName: 'Custom Command 2',
+          metadata: {
+            some : 'other',
+            object : 'data'
+          },
+          name: 'namespace:command-2'
+        }
+      ]);
+    });
+  });
 
   describe("::dispatch(target, commandName)", () => {
     it("simulates invocation of the given command ", () => {
--- a/spec/command-registry-spec.js
+++ b/spec/command-registry-spec.js
@@ -240,15 +240,15 @@ describe("CommandRegistry", () => {
       }).toThrow(new Error('Cannot register a command with a null listener.'));
     });
 
-    it("throws an error when called with an object listener without a handleEvent method", () => {
+    it("throws an error when called with an object listener without a onDidDispatch method", () => {
       const badListener = {
-        title: 'a listener without a handleEvent callback',
+        title: 'a listener without a onDidDispatch callback',
         description: 'this should throw an error'
       };
 
       expect(() => {
         registry.add(document.body, 'foo:bar', badListener);
-      }).toThrow(new Error('Listener must be a callback function or an object with a handleEvent method.'));
+      }).toThrow(new Error('Listener must be a callback function or an object with a onDidDispatch method.'));
     });
   });
 
@@ -281,7 +281,7 @@ describe("CommandRegistry", () => {
           some: 'other',
           object: 'data'
         },
-        handleEvent() {}
+        onDidDispatch() {}
       });
       registry.add('.grandchild', 'namespace:command-3', {
         name: 'some:other:incorrect:commandname',
@@ -290,7 +290,7 @@ describe("CommandRegistry", () => {
           some: 'other',
           object: 'data'
         },
-        handleEvent() {}
+        onDidDispatch() {}
       });
 
       const commands = registry.findCommands({target: grandchild});
--- a/spec/command-registry-spec.js
+++ b/spec/command-registry-spec.js
@@ -273,7 +273,7 @@ describe("CommandRegistry", () => {
       ]);
     });
 
-    it("returns commands with manual displayNames if set in the listener", () => {
+    it("returns commands with arbitrary metadata if set in a listener object", () => {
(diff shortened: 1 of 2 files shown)


src/command-registry.coffee
```diff
@@ -253,7 +254,10 @@ class CommandRegistry
       # registration order.
       for listener in listeners by -1
         break if immediatePropagationStopped
-        listener.callback.call(currentTarget, dispatchedEvent)
+        if typeof listener.listener.handleEvent is 'function'
```
nathansobo: Can we tweak the names here to avoid the `listener.listener` thing? We need a noun that describes the object that contains the listener... `commandMetadata`? Just `command`?

src/command-registry.coffee
```diff
@@ -154,16 +157,14 @@ class CommandRegistry
       for name, listeners of @inlineListenersByCommandName
         if listeners.has(currentTarget) and not commandNames.has(name)
           commandNames.add(name)
-          commands.push({name, displayName: _.humanizeEventName(name)})
+          commands.push(commandFromListener(name, listeners.get(currentTarget)))
```
nathansobo: Could we do this logic at registration time instead? We can make the new object based format the canonical representation of a command and just synthesize the displayName if a function is registered.

spec/command-registry-spec.coffee
```diff
@@ -237,6 +241,61 @@ describe "CommandRegistry", ->
         {name: 'namespace:command-1', displayName: 'Namespace: Command 1'}
       ]
 
+    it "returns commands with manual displayNames if set in the listener", ->
```
nathansobo: I almost think these two tests could be collapsed into one with a description like "returns arbitrary metadata from the command's registration". Testing that the name doesn't get overridden could just ride along with the main test with maybe a comment.

nathansobo (changes requested):
This is going to be cool. My main concern right now is that we need a good name for these new "object listeners". They are really "command descriptors" or something like that. I'm open to batting around some ideas.

Thanks for opening this up. Always feel free to ditch CoffeeScript for JS if it makes you more comfortable coding, though it is more work. ⚡️

src/command-registry.js
```diff
@@ -198,23 +200,20 @@ module.exports = class CommandRegistry {
         listeners = this.inlineListenersByCommandName[name]
         if (listeners.has(currentTarget) && !commandNames.has(name)) {
           commandNames.add(name)
-          commands.push({ name, displayName: _.humanizeEventName(name) })
+          const targetListeners = listeners.get(currentTarget);
+          commands.push(
+            ...targetListeners.map(listener => listener.getPublicCommand())
```
wbinnssmith: @nathansobo not really sure what the best thing to do in the case of multiple matching commands for the same element and commandName. 

I'm matching the behavior below in selector-based listeners, which means returning all matching commands, but am concerned about multiple listings (each with their own metadata) in e.g. command palette. 

Even in that case, choosing only one of the matching commands would have the effect or running every single listener :'(

src/command-registry.js
```diff
@@ -384,9 +381,9 @@ module.exports = class CommandRegistry {
 }
 
 class SelectorBasedListener {
-  constructor (selector, callback) {
+  constructor (selector, commandName, listener) {
     this.selector = selector
-    this.callback = callback
+    this.command = commandFromListener(commandName, listener)
```
wbinnssmith: let me know what you think of the naming here :)

wbinnssmith: everything is normalized here 🎉

src/command-registry.js
```diff
@@ -397,10 +394,41 @@ class SelectorBasedListener {
       this.sequenceNumber - other.sequenceNumber
     )
   }
+
+  getPublicCommand () {
+    return getPublicCommand.call(this)
```
wbinnssmith: `command` is the internal name of the structure inside the listener, and you call `getPublicCommand` to get a version that's safe to hand around externally (basically it doesn't have the `handleEvent` on it)

wbinnssmith:
@nathansobo let me know what you think of this progress. I think we might still be a bit misaligned on the naming of things, but I think this is a lot closer to what you had in mind, and I'm a lot happier with it too 🎉

cc @matthewwithanm

matthewwithanm:
This is so cooooool!

My 2¢ on the naming:

`handleEvent()` seems a little out-of-place to me. The analogue that springs to mind here (registering an object with a callback prop) is buttons in notifications, which uses `onDidClick` iirc. So my impulse would be to call it `onDidDispatch()`, paralleling the name on the command registry.

As for "command descriptors," that sounds fine, but why not just "commands"? `::findCommands()` already exists and returns these things, right?

wbinnssmith:
I love @matthewwithanm's analog with notifications and just adapted everything to `onDidDispatch` handlers, though there's a similarity with `onDidDispatch` on the command registry itself.

Basically there are 3 core objects in this flow, and I frankly think they could all use new/more accurate names:

* THING1 (input), currently called `listener`: The thing you pass to `atom.commands.add(target, commandName, THING1)`, which is either a function or an object describing command metadata *and* how to handle what happens when the command is dispatched:

		type THING1 = ((e: CustomEvent) => void) | {
		  onDidDispatch(e: CustomEvent): void,
		  displayName: string,
		  description: string,
		  ...otherMetadata
		};

* THING2 (internal only) currently also kind of called `Listener`s: The things currently known as `SelectorBasedListener`s and `InlineListener`s which wrap up THING1s, are stored in the registry, and now split out how to *handle* the command from descriptions and metadata *about* the command:

		 type THING2 = {
			 descriptor: CommandDescriptor,
		   onDidDispatch(e: CustomEvent): void,
		 };


* THING3 (output), currently called `commandDescriptor`: The things returned from `atom.commands.findCommands({ target }): Array<THING3>`, which represent the metadata about the command that was split as part of `THING2`. This is basically a normalized form of `THING1` with the handler function removed as a precaution so nobody can manually invoke a command without going through dispatch:

		type THING3 = {
			name: string,
		  displayName: string,
		  description: string,
		  ...otherMetadata
	  };
	  
Basically the only difference between `THING1` and `THING3` is whether they're passed in as input as handlers or returned as a publicly consumable descriptor. That and the `name` property in `THING3` is guaranteed to be the command name, and it will never have `onDidDispatch`. Maybe one is just a `public` version of the other? Or maybe a `ProvidedDescriptor` versus a `ConsumedDescriptor`?

spec/command-registry-spec.js
```diff
@@ -224,32 +224,35 @@ describe("CommandRegistry", () => {
...
         registry.add('.selector', 'foo:bar', badCallback);
-      } catch (error) {
-        addError = error;
-      }
-      expect(addError.message).toContain("Can't register a command with non-function callback.");
+      }).toThrow(new Error('Cannot register a command with a null listener.'));
     });
 
-    it("throws an error when called with an non-function callback and object target", () => {
+    it("throws an error when called with an null callback and object target", () => {
```
nathansobo: minor grammar nitpick I happened to notice: *a* null

spec/command-registry-spec.js
```diff
@@ -268,8 +271,75 @@ describe("CommandRegistry", () => {
         {name: 'namespace:command-2', displayName: 'Namespace: Command 2'},
         {name: 'namespace:command-1', displayName: 'Namespace: Command 1'}
       ]);
-  })
-);
+    });
+
+    it("returns commands with arbitrary metadata if set in a listener object", () => {
```
nathansobo: returns command *descriptors*?

src/command-registry.js
```diff
@@ -91,11 +89,22 @@ module.exports = class CommandRegistry {
...
-  //   element matching the selector. It will be called with `this` referencing
-  //   the matching DOM node.
-  //   * `event` A standard DOM event instance. Call `stopPropagation` or
-  //     `stopImmediatePropagation` to terminate bubbling early.
+  // * `listener` A listener which handles the event.  Either A {Function} to
+  //   call when the given command is invoked on an element matching the
+  //   selector, or an {Object} with a `onDidDispatch` property which is such a
+  //   function.
+  //
+  //   It will be called with `this` referencing the matching DOM node.
```
nathansobo: Maybe here say

```
The listener function will be called with `this` referencing the matching DOM node and the following argument:
```

nathansobo (changes requested):
Hey. Sorry for the delay. It's been hectic. I'm glad you've eliminated `listener.listener` and I think I'm okay referring to it as a "listener object" on the way in and a "descriptor" on the way out. Everything is looking good.

Slight tweak suggested to the naming... `onDidX` is usually used for a method that takes a callback and returns a disposable. In this case, this is more like a hook, which we usually name like `didDispatch`.

This should be the last round of changes. I think we're super close on this. I hope to get it merged before the next release cycle.

wbinnssmith:
@nathansobo no problem at all! just updated with fixes for your suggestions (they're all spot-on -- thanks!)

nathansobo (approved):

nathansobo:
Awesome. This is going to be a really nice improvement. Keep 'em coming! ⚡️

PR #15384, opened 2017-08-19
Fix highlight end pixel position calculation

nathansobo:
Previously, we were calculating the position preceding block decorations for the row following the end of the highlighted range, but that's actually wrong. We just want the position following block decorations of the end of the highlighted range, plus one line height. This prevents us from incorrectly rendering the end of highlight after block decorations that immediately follow the end of the highlighted range.

Before:

screen shot 2017-08-18 at 8 56 51 pm

After:

screen shot 2017-08-18 at 8 56 44 pm

/cc @ungb @as-cii

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1697,6 +1697,22 @@ describe('TextEditorComponent', () => {
       await component.getNextUpdatePromise()
       expect(Array.from(marker1Region.parentElement.children).indexOf(marker1Region)).toBe(0)
     })
+
+    it('correctly positions highlights that end on rows preceding block decorations', async () => {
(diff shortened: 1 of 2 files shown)


ungb:
I was testing this on merge conflict decorators and found the following issues:

The highlighting from the `Our Changes` is off
image


When you first open the merge conflict you can sometimes see the following:
image


When you change the window size the cursor eventually will go to the right spot and the issue is gone and only come back when you close and open the merge window.

Ping me if you can't repro the issue.

as-cii:
Nice work on this one, especially on figuring out the margin collapse issue. 💯 I'll go ahead and merge this once we get a green build.

PR #15389, opened 2017-08-19
Fix line number position when block decorations are at tile boundaries

as-cii:
When there are block decorations, we compute a `marginTop` style for line numbers, so that they can be aligned correctly to the corresponding line nodes.

This pull-request fixes a regression in the logic that calculated the `marginTop` value, which was failing to correctly render line numbers when block decorations were located right before the end of a tile, or exactly at the beginning of it.

| Before | After |
|--------|-------|
| screen shot 2017-08-19 at 15 41 35 | screen shot 2017-08-19 at 15 43 15 |

@Ben3eeE @50Wliu @ungb: I don't think this regression has been reported yet, but please feel free to test it and close any issue that may be related to this.

/cc: @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2233,6 +2233,23 @@ describe('TextEditorComponent', () => {
       expect(item6.previousSibling).toBe(lineNodeForScreenRow(component, 12))
     })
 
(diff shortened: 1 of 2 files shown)


Dubok41:
tyyjghjghjghj

PR #15392, opened 2017-08-21
Use standard (eslint's) 'env' field over globals

wbinnssmith:
Test Plan: ensure every removed global is a part of an added `env` from https://github.com/sindresorhus/globals/blob/master/globals.json

run `./script/lint`

Released under CC0

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,27 +174,14 @@
     "test": "node script/test"
   },
   "standard": {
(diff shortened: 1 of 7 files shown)


nathansobo:
Cool. We run the linter on CI so the fact that it's green is good enough for me.

PR #15393, opened 2017-08-21
[Atom development] Move linter dependencies to root package.json

wbinnssmith:
Atom's linter configurations are stored in the root, but since they were only used in files under `scripts/`, their npm dependencies were stored `scripts/package.json`. This meant that tools like the `coffeelint` and `standard` integrations for Atom couldn't find those binaries from `src/`, `spec/`, etc and either fell back to bundled or global versions (as did Atom's `linter-coffeescript`) which mismatched the lcoal installation, or simply failed silently (as with `linter-js-standard` and `linter-js-standard-engine`).

This change makes the correct versions available to these tools, and they remain accessible from `scripts/` so long as the root as its dependencies installed.

Somewhat unrelated: would you be opposed to me documenting my experience getting Atom dev up and running in a section next to `Building` in the project README? Could also mention these packages and how to build Atom using Atom itself.

Open to alternative suggestions to get these tools to play properly while developing Atom with Atom :)

Released under CC0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,6 +168,11 @@
     "language-xml": "0.35.2",
     "language-yaml": "0.30.1"
   },
+  "devDependencies": {
+    "coffeelint": "1.15.7",
+    "csslint": "1.0.2",
+    "standard": "8.4.0"
+  },
   "private": true,
   "scripts": {
     "preinstall": "node -e 'process.exit(0)'",
--- a/script/package.json
+++ b/script/package.json
@@ -4,9 +4,7 @@
   "dependencies": {
     "async": "2.0.1",
     "babel-core": "5.8.38",
-    "coffeelint": "1.15.7",
     "colors": "1.1.2",
-    "csslint": "1.0.2",
     "donna": "1.0.16",
     "electron-chromedriver": "~1.6",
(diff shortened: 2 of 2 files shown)


nathansobo:
One thing I'm worried about is that for some reason I don't fully recall right now, I'm not sure our build is currently smart enough to screen out dev dependencies when packaging up Atom. @BinaryMuse @as-cii you two were talking about this recently if I recall correctly. What's the deal with `devDependencies` currently? Any thoughts on how we could remedy the situation?

As for enhancing the `README`, that sounds excellent. We have a lot of rough edges in some of these aspects of the system as we've poured more time elsewhere. The kinds of house-keeping improvements you're making to things like the linter are super welcome. Thanks very much for the stewardship.

sadick254:
@wbinnssmith the build process doesn't have a way to screen out devDependencies. 
https://github.com/atom/atom/blob/4975493e80d7f9a211b4941dbc3077e1d3c6e521/script/lib/copy-assets.js#L39-L67
It is better to keep the linters under scripts for now. Sorry to say we cannot accept this PR at the moment untill we figure out how to screen out dev dependencies.

PR #15394, opened 2017-08-21
Convert CommandRegistry to JavaScript

wbinnssmith:
This was bootstrapped with running `decaffeinate` on both command-registry and its spec, but involved a line-by-line eyeing and updating for idioms, style, and in one case correctness:

For some reason, copying the keys to the new dispatchedEvent in coffeescript (command-registry.coffee:235) currently does not result in a thrown exception, as it should when trying to write over the read-only property `isTrusted`, so I added a check to ensure that the key did not already exist in the new event. Moreover, other definitions of `preventDefault`, `stopPropagation` and friends, which are also enumerable afaict, would also overwrite the ones defined just prior (command-registry.coffee:220 and command-registry.js:325), so that check ensures we don't overwrite those either.

Test plan: `./script/lint && ./script/test`, and a lengthy smoke test of various commands through keybindings, command palette, and package interfaces.

Released under CC0.


nathansobo:
☕️ 📉 👍 ❤️

Issue #15405, opened 2017-08-22
Moving lines by control-command up/down goes wrong


ethancfchen:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When I select some lines and want to move it, the editor will also move the line what I don't want to move, or missing some selected lines.

### Steps to Reproduce

1. Insert some random lines.
2. Select the last 2 lines.
3. Moving up 3 times by `control + command + ↑`

**Expected behavior:** [What you expect to happen]

![8 -22-2017 13-41-48](https://user-images.githubusercontent.com/5051791/29550667-7867a320-8741-11e7-9962-da3e4a37e460.gif)

**Actual behavior:** [What actually happens]

![error1](https://user-images.githubusercontent.com/5051791/29550664-73027b08-8741-11e7-84a4-9e18994e17b8.gif)
![error2](https://user-images.githubusercontent.com/5051791/29550665-7460ef02-8741-11e7-812c-3ac5e63b8413.gif)

**Reproduces how often:** [What percentage of the time does it reproduce?]

Sometimes.

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

OS: macOS Sierra 10.12.6

```bash
atom --version

Atom    : 1.19.2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

```bash
apm --version

apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.6.4
```

### Additional Information

I'm using trackpad as mouse and not holding any buttons when moving, but I think it's related to #15217.

Ben3eeE:
[...]

Issue #15217, opened 2017-08-09
Weird when editing text which is partial selected. (with required Prerequisites)


yubaoquan:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * [X] Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * [X] Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * [X] Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * [X] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * [X] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Select a part of a string and type a letter, the selection should disappear but not move to the right part of the string, this is a DEMO

![demo](https://user-images.githubusercontent.com/2340296/29136116-cede213e-7d6e-11e7-89b6-916e08ab2999.gif)

### Steps to Reproduce

1. File -> New File;
2. Input a string `aaaaabbbbbccccc`;
3. Select the middle part of the string `bbbbb`;
4. While still holding down the mouse button. Type a letter `x`;

**Expected behavior:** The text becomes `aaaaaxccccc` and no charcters are selected.

**Actual behavior:** The text becomes `aaaaaxccccc` but the middle 4 `c` are selected.

**Reproduces how often:** 100% reproduce.

### Versions

Atom    : 1.19.0
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.2
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.11.0

MacOS Sierra 10.12.6 

### Additional Information
I did the steps more than 10 times, every time it reproduce. But it will not reproduce when you use undo to get the selection back after type the letter `x`, like typing `cmd-z`.

Another problem is when I select some strings and hit backspace, the selected strings are deleted but strings next to the deleted one are automatically selected.

lee-dohm:
Thanks for reaching out!

We noticed that this is a duplicate of https://github.com/atom/atom/issues/14911. You may want to subscribe there for updates.

[...]

PR #15407, opened 2017-08-22
Don't select when typing while also holding the left mouse button

as-cii:
Fixes https://github.com/atom/atom/issues/15217
Fixes https://github.com/atom/atom/issues/15405

This pull-request addresses a regression that was causing edits performed while users were holding the left mouse button to result in unwanted selections.

| Before                                                                                                        | After                                                                                                        |
|---------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------|
| before | after |

Back in Atom 1.18, we used to handle this subtle behavior via an `onWillChange` event subscription on the buffer:

https://github.com/atom/atom/blob/783cda8642e42e842a7aac761f267e151f3db324/src/text-editor-component.coffee#L727

And this pull-request uses the same strategy to work around the problem.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -4055,6 +4055,79 @@ describe('TextEditorComponent', () => {
       expect(component.refs.horizontalScrollbar.element.scrollLeft).toBe(Math.round(12 * component.getBaseCharacterWidth()))
     })
   })
+
+  describe('handleMouseDragUntilMouseUp', () => {
+    it('repeatedly schedules `didDrag` calls on new animation frames after moving the mouse, and calls `didStopDragging` on mouseup', async () => {
+      const {component} = buildComponent()
+
+      let dragEvents
+      let dragging = false
+      component.handleMouseDragUntilMouseUp({
+        didDrag: (event) => {
+          dragging = true
+          dragEvents.push(event)
+        },
+        didStopDragging: () => { dragging = false }
+      })
+      expect(dragging).toBe(false)
+
+      dragEvents = []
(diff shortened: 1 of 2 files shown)


Ben3eeE (commented):
It seems that there was a test for this in Atom 1.18 that was removed in the editor rendering rewrite PR.

https://github.com/atom/atom/blob/1.18-releases/spec/text-editor-component-spec.js#L2734

The test was added in https://github.com/atom/atom/pull/8627

as-cii:
Yeah, I thought about it but we have changed the way that code path is exercised and testing this specific behavior would be a little trickier now. In particular, we don't synthesize `mousedown` and `mousemove` events anymore, but we spy on `handleMouseDragUntilMouseUp` instead and assume it does the right thing.

Initially I thought about moving `onWillChange` event handler up the call stack so that we can test it, but that would require duplicating some logic for lines and line numbers drag handlers. Another thought was to synthesize mouse events just for this one test, but it wouldn't blend in with the existing ones and seemed suboptimal.

After some more thinking, in e6b84db I ended up writing a unit test that exercises the lower level bits implemented by `handleMouseDragUntilMouseUp`.

Ben3eeE (approved):
LGTM :+1:

Ben3eeE:
It seems that some commands are not working correctly even with this guard. This was an issue before 1.19 was released so it's not a regression or issue with this PR. Just mentioning it here in case someone is testing and discovers this. /cc: https://github.com/atom/find-and-replace/issues/520

nathansobo:
Nice work. Sorry about this regression.

PR #15415, opened 2017-08-23
Adding option to skip main process tests. Cleaning up resources in tests

Xapphire13:
The main process tests take some time to complete, to improve the inner loop when working on features I have added the commandline option to skip these tests.

I have also cleaned up a duplicate import and added some resource cleanup to a few existing tests.

**Note**: These changes are extracted from my now abandoned PR #14456


Ingramz:
I'm liking the idea. 

I've seen other scripts use yargs, please change your code to use it instead of minimist. It would spare us from having another dependency.

Also on macOS there are benchmark tests that will probably always be run, perhaps it would be wise to have arguments for skipping any of the 3 types of tests.

Xapphire13:
Sorry about the delay, updated to use `yargs` 😄

Ingramz:
Looks good to me!

script/test
```diff
@@ -150,17 +151,26 @@ function runBenchmarkTests (callback) {
...
     case 'linux':
-      return [runCoreMainProcessTests]
+      suites = [runCoreMainProcessTests]
+      break
     default:
       console.log(`Unrecognized platform: ${platform}`)
-      return []
   }
+
+  if(argv.skipMainProcessTests) {
```
Ingramz: There should be a space between `if` and `(`

Xapphire13:
Ping

src/atom-environment.coffee
```diff
@@ -42,7 +42,6 @@ PaneContainer = require './pane-container'
 PaneAxis = require './pane-axis'
 Pane = require './pane'
 Dock = require './dock'
-Project = require './project'
```
50Wliu: Was this intentionally removed?

spec/git-repository-provider-spec.js
```diff
@@ -13,6 +13,14 @@ describe('GitRepositoryProvider', () => {
     provider = new GitRepositoryProvider(atom.project, atom.config, atom.confirm)
   })
 
+  afterEach(() => {
```
50Wliu: Why was this added?

src/atom-environment.coffee
Xapphire13: This is a duplicate require, check out [L36](https://github.com/atom/atom/pull/15415/files#diff-91bf726eeaaad75efb1705745a7af175L36)

spec/git-repository-provider-spec.js
Xapphire13: Because repositories are disposable and none of the tests perform cleanup, and each `describe` gets a new provider (losing the handle to the previous and making the provider subject to the GC which may not run before the tests finish). Just trying to make the test environment as clean as possible without memory leaks =]

Romario3000 (approved):
Большое спасибо!! пока в ознакомительном режиме, сильно не ковыряюсь

Xapphire13:
@BinaryMuse =]

BinaryMuse:
Thanks so much for this, @Xapphire13!

PR #15433, opened 2017-08-24
Fix so all directions are tested in the specs for moveActiveDirection

Ben3eeE:
Discovered by me and @jasonrudolph 

The tests for `splitRight`, `splitUp`, and `splitDown` were all testing with `splitLeft` for the `moveActiveItem` test.


jasonrudolph:
@Ben3eeE: Thanks for this. :bow::zap:

Issue #11189, opened 2016-03-17
Uncaught Error: Pane items must be objects. Attempted to add item undefined.


smlombardi:
**Steps to Reproduce:**
1. Open a new empty Atom in safe mode
1. If an untitled tab was created by the `Open Empty Editor on Start` configuration option close it so the pane is empty
1. Run `pane:split-[direction]-and-move-active-item` from the command palette

**Atom Version**: 1.7.0-beta0
**System**: Mac OS X 10.11.4
**Thrown From**: Atom Core
### Stack Trace

Uncaught Error: Pane items must be objects. Attempted to add item undefined.

```
At /Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane.js:450

Error: Pane items must be objects. Attempted to add item undefined.
    at Pane.module.exports.Pane.addItem (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane.js:450:15)
    at Pane.module.exports.Pane.moveItemToPane (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane.js:609:19)
    at PaneContainer.module.exports.PaneContainer.moveActiveItemToPane (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane-container.js:276:23)
    at atom-pane-container.module.exports.PaneContainerElement.moveActiveItemToNearestPaneInDirection (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane-container-element.js:96:20)
    at atom-pane-container.module.exports.PaneContainerElement.moveActiveItemToPaneOnLeft (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/pane-container-element.js:80:19)
    at atom-workspace.module.exports.WorkspaceElement.moveActiveItemToPaneOnLeft (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/workspace-element.js:151:33)
    at atom-workspace.commandRegistry.add.window:move-active-item-to-pane-on-left (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/register-default-commands.js:170:21)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom Beta.app/Contents/Resources/app.asar/src/command-registry.js:260:29)
    at /Applications/Atom Beta.app/Contents/Resources/app.asar/src/command-registry.js:3:61
[...]

PR #15438, opened 2017-08-24
Don't attempt to move undefined items when splitting panes

Ben3eeE:
### Description of the Change

Before this PR running `pane:split-[direction]-and-move-active-item` on an empty pane creates a new empty pane and throws an exception because `undefined` can not be moved. This PR adds a condition to not try and move if `@activeItem` is `undefined`.

With this change the behavior is the same in that we create an empty pane in the direction when doing this but without throwing an exception.

I also added a test for copying empty panes. `pane:split-[direction]-and-copy-active-item` already worked on empty panes before this PR.

### Alternate Designs

We could do nothing if `@activeItem` is `undefined`. But I did not go with this because it changes the behavior and the split functions should return the created pane.

The code is already in core.

### Benefits

* Fixes an exception

### Possible Drawbacks

Did not consider any drawbacks.

### Applicable Issues

Fixes https://github.com/atom/atom/issues/11189

### DIFF
--- a/spec/pane-spec.js
+++ b/spec/pane-spec.js
@@ -1152,6 +1152,32 @@ describe('Pane', () => {
       })
     })
 
+    describe('when the pane is empty', () => {
+      describe('when `moveActiveItem: true` is passed in the params', () => {
+        it('gracefully ignores the moveActiveItem parameter', () => {
+          pane1.destroyItem(item1)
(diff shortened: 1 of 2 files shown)


jasonrudolph (approved):
@Ben3eeE: Nice work! :zap:

PR #15440, opened 2017-08-25
Replace pathwatcher w/ bundled watcher to catch created & rename events

### LINKED ISSUE #14909: Settings lost
### Prerequisites

### Description

### Steps to Reproduce

1. Update packages
2. Restart

### Versions
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.6
git 1.9.1

OS: Linux Mint 17.3 Rosa

### Additional Information

kuychaco:
Fixes #14909 

This PR addresses three issues related to config settings data loss.

### Issue 1
_config file stops being watched_

Repro steps (on MacOS):
1. `mv config.cson config.cson-bak`
2. open Atom and open `config.cson`. Note that it's empty except for some settings set by packages. Add `editor.fontSize` and set it to something unusual. Note that the font updates immediately after setting it.
3. `mv config.cson-bak config.cson`
4. note that the `config.cson` file updates, but settings are not loaded. That is, if you changed the `editor.fontSize` normally you observe it updating on the spot, but making changes to the file after moving it will do nothing
5. reload the editor and note that `config.cson` has reverted back to its bare state 

Turns out that after step 3 we stop watching the `config.cson` file. Pathwatcher gives us a `delete`, and subsequent file changes do not result in more `change` events so we don't reload the user's config and update the in-memory settings. Reloading Atom flushes the stale settings to disk and overwrites the changes that were made resulting in lost settings. 

The fix implemented in this PR involves using @smashwilson's new bundled file watching service. We reload the user's config on `created`, `modified`, and `renamed` events. 

The bug outlined above resulted in the config file reverting to a previous version of the file prior to being overwritten. It seems that most of the reports in #14909 indicate that the config file was reset completely, which points to a root cause other than what was outlined above. 

The rest of this PR aims to make the code around writing to the config file more airtight. There are only two places where we make calls to `writeFileSync`: 

### Issue 2
_race condition in `loadUserConfig` when a config file does not yet exist_

Previously, there was room for a race condition. An external process could have created a config file after we checked for its existence and before we wrote to the file, potentially resulting in overwriting the changes made by an external process. 
```
      unless fs.existsSync(@configFilePath)
        CSON.writeFileSync(@configFilePath, {})
``` 
Solution: Don't write file if it already exists (depends on https://github.com/atom/season/pull/22)
```
      CSON.writeFileSync(@configFilePath, {flag: 'x'}, {}) # fails if file exists
```

### Issue 3

_race condition in `save()` we write the blank in-memory settings to disk before the first `loadUserSettings()` call_

Previously, there was the possibility that this happened before the in-memory settings were updated to reflect the user's config on disk (`@settings` is initialized to `{}` and updated only after `loadUserConfig` is called). In this case, writing to the file would clobber all the user's settings. This is the most likely cause of the issues reported in #14909. 

Solution: Store of `set` and `unset` operations run prior to loading the user's config and run them once the in-memory settings are updated. 

TODO:
- [x] fix tests
- [x] update `season` to take options
- [x] store pending operations in config before loaded
- [ ] for hotfix https://github.com/Axosoft/nsfw/blob/ef23f197cff6641276d90cdd779b4c1824fe6d6c/js/src/index.js#L80-L123

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "scandal": "^3.1.0",
     "scoped-property-store": "^0.17.0",
     "scrollbar-style": "^3.2",
-    "season": "^6.0.0",
+    "season": "^6.0.1",
     "semver": "^4.3.3",
     "service-hub": "^0.7.4",
     "sinon": "1.17.4",
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1750,7 +1750,7 @@ describe "Config", ->
           expect(atom.config.set('foo.bar.str_options', 'One')).toBe false
           expect(atom.config.get('foo.bar.str_options')).toEqual 'two'
 
-  describe "when .set/.unset is called prior to .loadUserConfig", ->
+  fdescribe "when .set/.unset is called prior to .loadUserConfig", ->
     beforeEach ->
       atom.config.settingsLoaded = false
 
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1753,7 +1753,6 @@ describe "Config", ->
   fdescribe "when .set/.unset is called prior to .loadUserConfig", ->
     beforeEach ->
       atom.config.settingsLoaded = false
-
       fs.writeFileSync atom.config.configFilePath, """
         '*':
           foo:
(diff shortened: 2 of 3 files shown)


smashwilson:
> Solution: Store of set and unset operations run prior to loading the user's config and run them once the in-memory settings are updated.

I've made a slight change here. When we store pending operations that happen before the first call to `.loadUserSettings()`, we create the possibility for odd situations like this:

```js
atom.config.set('foo.bar', 10)
atom.config.get('foo.bar')  // => undefined
```

Instead, what I've done is store the pending operation in a closure _and_ allow it to proceed (without triggering a `requestSave()` call). The immediately-applied change allows it to be reflected immediately in `get()` calls, and the pending operation ensures that the change isn't lost after the first load arrives.

nathansobo:
:zap: Issue 3 mentioned in the body definitely sounds suspicious. Nice work you two!

Ben3eeE:
Very nice work here :mag:  :clap:

PR #15446, opened 2017-08-25
Use fs-admin instead of runas

maxbrunsfeld:
Now that we're [using](https://github.com/atom/text-buffer/pull/247) `fs-admin` instead of `runas` in `text-buffer`, we can remove the few usages of `runas` in Atom itself and drop it as a dependency entirely, which will slightly reduce our bundle size and the amount of code we have to maintain.


mjguru:
Shouldn't this be merged by now?

PR #15450, opened 2017-08-26
Updated to New Dependencies.

yoginth:
Updated out of dated and Pinned, out of date to new Stable Dependencies.

### Requirements


alexandernst:
Extremely brave

50Wliu:
Thanks for reaching out!

alexandernst:
@50Wliu Just so that @yoginth doesn't feel pushed away and keeps trying to contribute, it might be a good idea to explain why this PR is actually being rejected. 

@yoginth Filling the template is a requirement, but your PR is being rejected because it utterly breaks everything. You blindly updated all dependencies without realising that most of them have incompatible API changes, which require further changes in Atom itself.

50Wliu:
@alexandernst thanks for the suggestion.  @yoginth I did primarily close this PR because it didn't have the template completed, which is pretty important for a PR that updates eighteen packages that Atom depends on to run.  When updating dependencies, you should give a clear explanation of _why_ you are doing so and _what_ benefits updating will provide.  Any upgrade should also be thoroughly tested to ensure that the possibility of regressions is as small as possible.  Given that Atom didn't even *build* with these changes, this PR is in no shape to be merged.

And yes, the Atom team is aware that many of the dependencies are outdated, and it would definitely be nice to update some of them.  However, they still work, and there are currently higher priorities for the Atom team.  If you would like to take this task on, feel free to do so, but please keep in mind my comments above.  It would probably be safest to update only one dependency at a time.

yoginth:
ok sorry! and thanks for your suggestions!

PR #15486, opened 2017-08-29
Move highlight decorations outside of tiles

### LINKED ISSUE #15449: Recreated highlight node re-trigger keyframe animation on contained line-tile changed
CC @nathansobo, @as-cii

Similar to #15185 except
- This issue happens even in v1.18.0(before editor rewrite).
- Different condition than #15185.

decoration-node-recreated

I understand fixing this is not easy but hope fixed.
Thanks for great work as always.

### Prerequisites

### Description

When highlight node is **moved** to different lines-tile, highlight node on existing lines-tile re-created and re-trigger keyframe animation.

### Steps to Reproduce

Can reproduce in safe-mode.

1. Prepare style to use keyframe animation on your `style.less`

```less
@keyframes t9md-test-animation {
  from { background-color: red; }
  to { background-color: transparent; }
}

atom-text-editor .t9md-test-flash .region {
  animation-name: t9md-test-animation;
  animation-fill-mode: forwards;
  animation-duration: 10.0s;
  animation-iteration-count: 1;
}
```

2. Open editor and prepare following text and select `line 2` to `line5`

```

line 2
line 3
line 4
line 5
```

3. Open chrome dev-console and paste following code.

This code copy selected text to clipboard and flash selected range which flash disappear after 10sec.

```javascript
(function flashreproduceIssue() {
  editor.copySelectedText()
  const marker = editor.markBufferRange(editor.getSelectedBufferRange())
  editor.decorateMarker(marker, {type: "highlight", class: "t9md-test-flash"})
  setTimeout(function () {
    marker.destroy()
    console.log("destroyed");
  }, 10000);
})()
```

4. Within 10 sec, move to point `[0, 0]` and paste from clipboard and see original flash( keyframe animation ) is re-triggered.

What is happening here is
- Pasting 4 lines of text create new lines-tile and inserted **before** old line-tile
- At this timing, highlight node in old lines-tile is re-rendered.
- Which result in re-trigger of keyframe animation.

Don't re-trigger keyframe animation.
Want highlight decoration node being stable once created.

highlight decoration node is re-created when rendered line-tiles are changed.

100%.

### Versions

```
Atom    : 1.20.0-beta4
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

```
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

### Additional Information

Actual issue happened in https://github.com/t9md/atom-vim-mode-plus/issues/843.

as-cii:
As a consequence of https://github.com/atom/atom/pull/15378, we are now able to render highlight decorations in a separate div, as opposed to having an highlight container for each tile.

Code-wise this is much simpler, because highlights spanning multiple tiles can be represented via a single region and don't need to be split across the tiles they span anymore. As a byproduct, performance should improve as well, because the number of nodes that need to be managed should decrease significantly.

This also fixes https://github.com/atom/atom/issues/15449 (and other similar rendering artifacts), because highlight decoration DOM nodes won't need to move between tiles anymore when their position changes.

/cc: @nathansobo @t9md

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -156,7 +156,7 @@ class TextEditorComponent {
     this.decorationsToRender = {
       lineNumbers: null,
       lines: null,
-      highlights: new Map(),
+      highlights: [],
       cursors: [],
       overlays: [],
       customGutter: new Map(),
@@ -164,7 +164,7 @@ class TextEditorComponent {
       text: []
     }
     this.decorationsToMeasure = {
-      highlights: new Map(),
+      highlights: [],
       cursors: new Map()
     }
     this.textDecorationsByMarker = new Map()
@@ -546,7 +546,6 @@ class TextEditorComponent {
   }
 
   renderContent () {
-    let children
     let style = {
       contain: 'strict',
       overflow: 'hidden',
@@ -557,17 +556,6 @@ class TextEditorComponent {
       style.height = ceilToPhysicalPixelBoundary(this.getScrollHeight()) + 'px'
       style.willChange = 'transform'
       style.transform = `translate(${-roundToPhysicalPixelBoundary(this.getScrollLeft())}px, ${-roundToPhysicalPixelBoundary(this.getScrollTop())}px)`
-      children = [
-        this.renderLineTiles(),
-        this.renderBlockDecorationMeasurementArea(),
-        this.renderCharacterMeasurementLine()
-      ]
-    } else {
-      children = [
-        this.renderLineTiles(),
-        this.renderBlockDecorationMeasurementArea(),
-        this.renderCharacterMeasurementLine()
-      ]
     }
 
     return $.div(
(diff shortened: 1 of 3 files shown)


src/text-editor-component.js
```diff
@@ -4009,6 +3923,90 @@ class LineComponent {
   }
 }
 
+class HighlightsComponent {
+  constructor (props) {
+    this.props = {}
+    this.element = document.createElement('div')
+    this.element.className = 'highlights'
+    this.element.style.contain = 'strict'
```
as-cii: @nathansobo: I tentatively set the layout containment to `strict` now that highlights are not inside tiles anymore. Previously, we were using only layout containment (which, as a side note, I don't think was being enforced by Chromium, because we wrote the property name and the value in the opposite order, e.g. `layout: contain` instead of `contain: layout`). 

Do you have any concerns about switching to a stricter form of containment? Rendering-wise everything seems to displayed correctly, but I'd love to have your input on this.

nathansobo: If everything is working well it seems reasonable to apply strict layout containment. We can't apply it to the line tiles because we want to allow text to escape them at awkward line heights and the highlights need to show through from behind, correct?

nathansobo:
👍

t9md:
@as-cii This is not included in v1.20.0-beta6, any reason?

PR #15487, opened 2017-08-29
Always honor scroll intent on mousewheel

### LINKED ISSUE #15467: Editor window scrolling sensitivity has changed since updating to 1.19.3
### Prerequisites

### Description

Since updating to 1.19.3, scrolling in the main text window is slower than it used to be. I'm using a late 2016 MacBook Pro with track pad. The reduction in speed is so noticeable that it's actually possible to scroll your fingers up or down the track pad slowly and Atom will fail to scroll at all.

I've reproduced this both with all my extensions enabled and in Safe Mode.

Unfortunately I don't know which version of Atom I was on prior to updating to 1.19.3.

Other apps continue to scroll at the normal pace.

### Steps to Reproduce

1. Scroll two fingers slowly up or down the track pad.
2. Atom scrolls the editor window more slowly than it should. When done slowly enough, Atom fails to scroll at all.

**Expected behavior:**

Atom's scroll speed matches that of other applications.

**Actual behavior:**

Atom scrolls more slowly than other applications.

**Reproduces how often:**

100% of the time.

### Versions

OS: macOS 10.12.3
Atom    : 1.19.3
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.13.3

as-cii:
Fixes #15467

This pull-request introduces a change in the way we respond to mousewheel events. In particular, if `-1 < delta * scrollSensitivity < 1`, we will scroll up or down by 1 in an attempt to honor the user's scroll intent.

/cc: @iolsen @lee-dohm @ungb @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1209,7 +1209,6 @@ describe('TextEditorComponent', () => {
       }
 
       {
-        global.debug = true
         const expectedScrollTop = component.getScrollTop()
         const expectedScrollLeft = 20 * (scrollSensitivity / 100)
         component.didMouseWheel({deltaX: 20, deltaY: -10})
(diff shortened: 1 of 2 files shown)


spec/text-editor-component-spec.js
```diff
@@ -1209,7 +1209,6 @@ describe('TextEditorComponent', () => {
       }
 
       {
-        global.debug = true
```
nathansobo: 🙈

nathansobo (approved):

PR #15492, opened 2017-08-29
Shrink probot-stale window to one year and close in two weeks

lee-dohm:
[probot-stale](https://github.com/probot/stale) has been helping us clean up old issues. We originally implemented it with a very liberal two year stale window to ensure that it worked reliably. Now that probot-stale has [over five thousand issues across all of GitHub under its belt](https://github.com/search?q=commenter%3Aapp%2Fstale&ref=opensearch), we're confident that we can trust it with a more realistic window.

We're shrinking the window where something will be marked stale to one year and if there is no further activity after that, it will be closed in an additional two weeks. All other settings will remain the same.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -2,9 +2,9 @@
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
 # Starting at two years of no activity
-daysUntilStale: 730
+daysUntilStale: 365
 # Number of days of inactivity before a stale Issue or Pull Request is closed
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -2,9 +2,9 @@
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
 # Starting at two years of no activity
```
Arcanemagus: This line should be removed or updated to say one year.

Issue #11429, opened 2016-04-11
Scroll to center is ignored if already visible


314eter:
### Description

When scrolling to a position that is already visible with the `center` option set to `true` (as described [here](https://atom.io/docs/api/v1.6.2/TextEditor#instance-scrollToBufferPosition)), no scrolling is performed.
### Steps to Reproduce

``` javascript
editor = atom.workspace.getActiveTextEditor() /* open a file with 100 lines */
editor.scrollToBufferPosition([50, 0]) /* scroll to line 50, line 52 is visible now */
editor.scrollToBufferPosition([52, 0], {center: true})
```

**Expected behavior:** According to the API docs, line 52 should be scrolled to the center of the editor.

**Actual behavior:** Nothing happens.
### Further information

Before e3dabd5200afea5074fb63003951ef677539413e, the behavior was as expected. Maybe the change was intentional, but then I don't understand why.

The problem can be fixed by removing

``` coffeescript
unless @getScrollTop() < desiredScrollCenter < @getScrollBottom()
```

in [text-editor-presenter.coffee](https://github.com/atom/atom/blob/master/src/text-editor-presenter.coffee). I can open a PR with this change if necessary.

Ben3eeE:
Activity

314eter:
The issue is still reproducible.

rsese:
:+1:

> What version of Atom you reproduced the issue on

1.32.0-nightly6

> What OS and version you reproduced the issue on

macOS 10.12.6

> What steps you followed to reproduce the issue

The steps in the issue body.

divieira:
Also ran into this. Seems to be related to the (closed) issues #3131 and #3324.

PR #15496, opened 2017-08-29
Scroll to center even if line is already visible

314eter:
### Description of the Change

By removing the check that the `desiredScrollCenter` is not yet between the current `scrollTop` and `scrollBottom`, the scroll is always executed if possible. This is the expected behaviour according to the [docs](https://atom.io/docs/api/v1.19.4/TextEditor#instance-scrollToBufferPosition).

### Alternate Designs

The alternative is not changing it. I don't know why the check was ever added, but I can't think of any possible benefits.

It's already there.

### Benefits

The behaviour will be as documented. This means that it's possible to center a certain position/range, regardless of the current scroll position.

### Possible Drawbacks

Nothing I can think of.

### Applicable Issues

#11429

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1130,8 +1130,15 @@ describe('TextEditorComponent', () => {
       editor.scrollToScreenRange([[4, 0], [6, 0]], {center: true})
       await component.getNextUpdatePromise()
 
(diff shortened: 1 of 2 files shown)


314eter:
The failing test seems unrelated

sadick254:
@314eter This PR has been superseded by https://github.com/atom/atom/pull/21488. I will have to close this PR since the issue has been fixed and available on the latest atom release `v1.57.0`.

PR #15503, opened 2017-08-30
Fix rendering of block decorations for invalid markers

### LINKED ISSUE #15490: It is possible to delete block decorations with core:delete and core:backspace
### Prerequisites

    ~* Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode~

### Description

Selecting text before and after a block decoration and deleting causes the block decoration to be removed but it still takes up space making the lines go out of sync.

I reproduced this using the [preview-inline](https://atom.io/packages/preview-inline) package.

### Steps to Reproduce

block decoration

**Expected behavior:** In Atom 1.18 performing these steps did not delete the block decoration.

**Actual behavior:** Block decoration is removed but still takes up space causing lines to go out of sync.

### Versions

Atom 1.19.4

### Additional Information

/cc: @as-cii @nathansobo

as-cii:
Fixes #15490

Previously, when a marker became invalid we would delete the corresponding block decoration from the DOM, without however removing it from the `lineTopIndex`. Also, we had a similar issue when decorating a marker that was already invalid: we would account for it in the index without adding it to the DOM.

This pull-request addresses the above problems by ensuring that block decorations are never added for invalid markers. At the same time, the editor component will still keep track of changes on the marker that was decorated, so that it can detect when its validity changes and render it appropriately.

/cc: @nathansobo @Ben3eeE

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2460,37 +2460,61 @@ class TextEditorComponent {
     const {model} = this.props
     const decorations = model.getDecorations({type: 'block'})
     for (let i = 0; i < decorations.length; i++) {
-      this.didAddBlockDecoration(decorations[i])
+      this.addBlockDecoration(decorations[i])
     }
   }
 
-  didAddBlockDecoration (decoration) {
+  addBlockDecoration (decoration, subscribeToChanges = true) {
     const marker = decoration.getMarker()
     const {item, position} = decoration.getProperties()
     const element = TextEditor.viewForItem(item)
-    const row = marker.getHeadScreenPosition().row
-    this.lineTopIndex.insertBlock(decoration, row, 0, position === 'after')
 
-    this.blockDecorationsToMeasure.add(decoration)
-    this.blockDecorationsByElement.set(element, decoration)
-    this.blockDecorationResizeObserver.observe(element)
+    if (marker.isValid()) {
+      const row = marker.getHeadScreenPosition().row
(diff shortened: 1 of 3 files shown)


src/text-editor-component.js
```diff
@@ -2460,37 +2460,61 @@ class TextEditorComponent {
...
-      didUpdateDisposable.dispose()
-      didDestroyDisposable.dispose()
       this.scheduleUpdate()
-    })
+    }
+
+    if (subscribeToChanges) {
+      let wasValid = marker.isValid()
+
+      const didUpdateDisposable = marker.bufferMarker.onDidChange(({textChanged}) => {
```
nathansobo: It's really unfortunate to add another consumer of this `onDidChange` API. Maybe it would be worth it to add a method just for subscribing to changes in validity, which would be cheaper to provide.

nathansobo:
Other than my concern about adding more `onDidChange` subscriptions, this looks good.

src/text-editor-component.js
as-cii: Yeah, I agree it's unfortunate to have an `onDidChange` handler here, but we needed it even before this pull-request to update the `LineTopIndex` when markers are moved manually, so I blended in with the existing code.

I wonder if at some point we should change `MarkerIndex` to support arbitrary metadata (or maybe just block decorations). It would make this entire code path simpler, although we would still have to account for invalid markers.

Issue #15442, opened 2017-08-25
Scroll bars not showing in center pane items


lee-dohm:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Scroll bars not showing in pane items (text editors, settings view, etc)

### Steps to Reproduce

1. Go to System Preferences
1. Click General
2. Set "Show Scroll Bars" to "Always"
1. Launch Atom in your local Atom repository
1. Open `README.md`

**Expected behavior:** Scroll bars are visible in the text editor

**Actual behavior:** Scroll bars are only visible in the tree view

### Versions

```
Atom    : 1.19.3
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

```
apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.14.0
```

### Additional Information

Reproduces no matter what the OS preference is for scroll bars, "Always" just makes it more obvious.

lee-dohm:
Possibly related: https://github.com/atom/atom/issues/15401 and https://github.com/atom/atom/pull/15367

iolsen:
@nathansobo @as-cii likely caused by recent changes.

hubertott:
Atom: 1.19.3
macOS: 10.12.5
Chrome: 60.0.3112.101 

For me scroll bars are hidden and unusable unless I set the OS preference to "Always" at which point they show. Note a restart of Atom is required for the OS change to have an effect.

TitanNano:
I made the following discovery after the comment by @hubertott 

1. start with "when scrolling"
    1. tree-view shows native osx scrollbar on demand, editor has no scrollbar
    2. tree-view has static osx scrollbar, editor has no scrollbar
2. start with "always"
    1. tree-view as atoms custom scrollbar, edtior has also atoms custom scrollbar
[...]

PR #15506, opened 2017-08-30
Honor macOS "Show scrollbars only when scrolling" setting

as-cii:
Fixes https://github.com/atom/atom/issues/15442

Previously, we were hiding scrollbars when their height/width was 0. On macOS, however, users can decide to only show scrollbars while scrolling, which causes Atom to detect scrollbars as being invisible during measurements. As a result, we were mistakenly setting the visibility property to `hidden` when this setting was on, thus preventing users from seeing the scrollbar on scroll.

With this pull-request we are:

* Changing the dummy scrollbar components to only become invisible when the content is not scrollable rather than when the scrollbars have zero width or height.
* Moving `DummyScrollbarComponent` back into its own GPU layer. This was a workaround that got added back in bac9922 and it's needed in order for scrollbars to be shown properly when the "Show scrollbars only when scrolling" setting is on.
* Renaming the `is{Horizontal,Vertical}ScrollbarVisible` functions to `canScroll{Horizontally,Vertically}`, to better express their intent. In fact, if the content is scrollable it doesn't necessarily mean that the scrollbar is also visible.

/cc: @nathansobo @Ben3eeE @ungb @lee-dohm

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -688,8 +688,8 @@ describe('TextEditorComponent', () => {
       await component.getNextUpdatePromise()
       element.style.width = component.getGutterContainerWidth() + component.getContentHeight() - 20 + 'px'
       await component.getNextUpdatePromise()
-      expect(component.isHorizontalScrollbarVisible()).toBe(true)
-      expect(component.isVerticalScrollbarVisible()).toBe(false)
+      expect(component.canScrollHorizontally()).toBe(true)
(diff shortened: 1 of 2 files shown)


nathansobo:
:zap:

lee-dohm:
Confirmed working using Atom v1.22.0-dev-50088b16c on Mac OS X 10.12.6.

Alhadis:
# HOLY DAMN, ***THANK YOU!!!***

PR #15510, opened 2017-08-31
Remove delay on window focus (fixes #9544)

### LINKED ISSUE #9544: Atom is slow regaining focus on Windows and Linux with git repo
Atom has a delay of ~1 second to respond to input when re-focusing. Seems to only affect Atom when a project folder is in a git repository.

Steps to reproduce:
Open Atom
Add a directory to your project that is a git repository or a subfolder of a git repository.
Open a file or create a new document in the same Atom session.
Switch to a different application.
Click back on Atom and wait for your input to be accepted.

When you click back on Atom there is a delay of 1~1.5 seconds before your action is performed (tab switch if you clicked a tab, cursor to start blinking if you clicked in the text of a document).

I ran the profiler during the task switching and I got the following data (rename the file to .json if necessary):
[TimelineRawData-20151112T091728.txt](https://github.com/atom/atom/files/32999/TimelineRawData-20151112T091728.txt)

I am able to solve the issue only by removing the git folder from my project.

I'm not versed in reading the profiling data but it looks like a call to `git-repository.js` is taking a lot of time up when the re-focus happens.

HebaruSan:
- Wire up the focus handler unconditionally
- Add new `core.gitRepositoryRefreshOnWindowFocus` setting
- Check `core.gitRepositoryRefreshOnWindowFocus` in focus handler
- Remove `refreshOnWindowFocus` param since it was unused
- Document the `project` and `config` params

### Description of the Change

If the user has one or more projects in the tree view with Git repositories, Atom freezes for a few seconds each time the window gains focus. The editor spends this time re-indexing Git data for all of those projects in the foreground. This makes Atom feel sluggish and unresponsive, and disrupts work flow when the user is likely to be ready to blaze ahead with an active task.

See #9544 for full discussion of the issue.

After this change, the default behavior is to no longer do this re-indexing work on window focus. A new settings checkbox can be used to enable the old behavior:

image

### Alternate Designs

The old code had an unused Boolean parameter controlling whether the focus handler would be wired up. We could have passed the new setting as the value of that parameter. I decided that wasn't desirable because the changes could not take effect immediately when the user changed the setting. To respond to user setting changes on the fly, we need to wire up the event unconditionally and check the setting's current value when it fires. Since the old parameter would conflict with the new setting, it is removed.

Fixes an issue with Core.

### Benefits

- Improved responsiveness when multitasking Atom with other applications
- Multiple users on #9544 reported quitting Atom or almost quitting due to this issue, so possible improvement in user outreach and retention
- Multiple users on #9544 resorted to risky workarounds (deleting or renaming a project's .git folder, deactivating core packages, avoiding the use of project folders), so possible improvement in general stability if those workarounds are no longer needed

### Possible Drawbacks

The old code does not document why this behavior existed, so we do not know the exact rationale for it or what might be lost by removing it. I speculate that it is to account for the possibility that the user might make changes affecting one of Atom's currently open Git repositories in an external application, which would then need to be reflected in Atom. However, this behavior imposes a serious usability penalty for all Git users, and users who do not need it have no recourse to eliminate this penalty. Furthermore, it is not a robust solution to that potential problem, as it is possible to modify a Git repo without switching focus away from Atom (for example, via a scheduled task or a remote login from another computer).

To mitigate the possibility of new issues, the problematic code is not simply removed. Rather, a setting is created so that users who need the old behavior and are willing to tolerate its drawbacks can enable it with a checkbox in the Core settings.

### Applicable Issues

- Fixes #9544

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -19,6 +19,12 @@ const configSchema = {
         title: 'Exclude VCS Ignored Paths',
         description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, projects using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
       },
+      gitRepositoryRefreshOnWindowFocus: {
+          type: 'boolean',
+          default: false,
+          title: 'Re-index Git Repositories on Window Focus',
+          description: 'If true, Atom will pause to re-index all Git repositories that are open in the tree view each time the focus returns to the Atom window. May cause a 3-10 second delay each time you switch to and from another window.'
+      },
       followSymlinks: {
         type: 'boolean',
         default: true,
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -20,10 +20,10 @@ const configSchema = {
         description: 'Files and directories ignored by the current project\'s VCS system will be ignored by some packages, such as the fuzzy finder and find and replace. For example, projects using Git have these paths defined in the .gitignore file. Individual packages might have additional config settings for ignoring VCS ignored files and folders.'
       },
       gitRepositoryRefreshOnWindowFocus: {
-          type: 'boolean',
-          default: false,
-          title: 'Re-index Git Repositories on Window Focus',
-          description: 'If true, Atom will pause to re-index all Git repositories that are open in the tree view each time the focus returns to the Atom window. May cause a 3-10 second delay each time you switch to and from another window.'
+        type: 'boolean',
+        default: false,
+        title: 'Re-index Git Repositories on Window Focus',
+        description: 'If true, Atom will pause to re-index all Git repositories that are open in the tree view each time the focus returns to the Atom window. May cause a 3-10 second delay each time you switch to and from another window.'
       },
       followSymlinks: {
         type: 'boolean',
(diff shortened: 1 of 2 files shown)


HebaruSan:
The latest CI error says:

```
Error: Expected HTMLNode not to be defined.
    at new jasmine.ExpectationResult (/Users/distiller/atom/vendor/jasmine.js:120:32)
    at newMatchersClass.toBeDefined (/Users/distiller/atom/vendor/jasmine.js:1241:29)
    at jasmine.Spec.&lt;anonymous&gt; (/Users/distiller/atom/node_modules/fuzzy-finder/spec/fuzzy-finder-spec.coffee:1268:131)
    at jasmine.Block.execute (/Users/distiller/atom/vendor/jasmine.js:1070:17)
    at jasmine.Queue.next_ (/Users/distiller/atom/vendor/jasmine.js:2102:31)
    at /Users/distiller/atom/vendor/jasmine.js:2092:18
```

If that's actually related to this PR, I'll be happy to make whatever changes are needed, but it looks spurious to me.

50Wliu:
Hi @HebaruSan, thanks for the PR!  However, we already have a different PR fixing this in the short/medium term at #15702, and a long-term fix is in-progress at atom/git-utils#80.  For those two reasons, I'm going to close this PR.  Thanks again for your initiative and interest in Atom though!

HebaruSan:
I look forward to trying out the proper solution!

PR #15528, opened 2017-09-02
Don't remeasure invalid block decorations

as-cii:
This fixes an uncaught exception that was being thrown after invalidating a marker and resizing the editor. I don't think this issue was reported, but it could be reproduced by:

1. Opening a file with merge conflicts in a git repository.
2. Delete the entire region containing conflict markers.
3. Try to resize the editor.
4. 💥

/cc: @nathansobo @Ben3eeE


Ben3eeE:
👍

Issue #15489, opened 2017-08-29
Editor Scrolls Down When Clicking Directly Above Horizontal Scrollbar


nathansomething:
### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

The editor view scrolls down/up if you click near the bottom/top of the page. This is particularly annoying because whenever I try to scroll horizontally, the editor scrolls down. Is there anyway to fix this?

### Steps to Reproduce

1. Open Atom
2. Click right above the horizontal scrollbar
3. The editor window scrolls down

**Expected behavior:** The editor does not scroll

**Actual behavior:** The editor scrolls down

Gif from @rsese https://github.com/atom/atom/issues/15489#issuecomment-325751074

![click-scroll](https://user-images.githubusercontent.com/734194/29837029-a65cbfec-8cbc-11e7-9ce4-085329e95390.gif)

**Reproduces how often:** 100% of the time

### Versions 1.19.4

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

### Additional Information

I'm running macOS Sierra version 10.12.6

From @rsese:

Should set `System Preferences > General > Show scroll bars` to `Always`.

Doesn't reproduce in 1.18.0.

rsese:
Thanks for the report - I can reproduce this as of 1.19.0 and on 1.19.4 on macOS 10.12.6 but not if I click directly on the horizontal scrollbar, only if I click directly above the horizontal scrollbar.

![click-scroll](https://user-images.githubusercontent.com/734194/29837029-a65cbfec-8cbc-11e7-9ce4-085329e95390.gif)

I couldn't reproduce on 1.18.0 though so tagging as a regression.

Ben3eeE:
This is really annoying on Windows because you can't even click on the scrollbar without hitting this bug /cc: https://github.com/atom/atom/issues/15488

[...]

PR #15546, opened 2017-09-05
Don't autoscroll when using the mouse to add, delete or move selections

as-cii:
Fixes https://github.com/atom/atom/issues/15489

Previously, we used to scroll the editor on behalf of the user when clicking near the edges. This behavior was being perceived as jarring, because it would move up/down the line the user was trying to click:

kapture 2017-09-05 at 15 23 22

With this pull-request we are restoring the behavior Atom 1.18 exhibited, which prevents the editor from auto-scrolling when using the mouse to add, delete or move selections:

kapture 2017-09-05 at 15 25 24

/cc: @nathansobo @ungb @Ben3eeE @50Wliu

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2706,6 +2706,8 @@ describe('TextEditorComponent', () => {
           clientY: clientTopForLine(component, 3) + lineHeight / 2
         })
         expect(editor.getCursorScreenPosition()).toEqual([3, 16])
+
+        expect(editor.testAutoscrollRequests).toEqual([])
       })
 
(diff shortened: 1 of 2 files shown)


rsese:
Confirmed the fix on macOS 10.12.6 with this branch :v:

ungb:
I was able to verify as well before and after!  LGTM!

sanada01:
Hi! Where can I find thiese two files?

- text-editor-component-spec.js
- text-editor-component.js

:)

sanada01:
1.19.6 is perfect, thanks. :)

Issue #15441, opened 2017-08-25
Uncaught TypeError: Cannot read property 'id' of undefined


relupselav:
Steps to reproduce by @mangecoeur https://github.com/atom/atom/issues/15441#issuecomment-325651325

[test.zip](https://github.com/atom/atom/files/1260242/test.zip)

Steps:

1. unzip the file and open test.md in atom
1. Make the Atom Window narrow
1. trigger `preview-inline` on BOTH image links
1. Scroll up and down
-> Error seems to be triggered when the top of the block decoration of the first decoration enters or exits the viewable window.
1. Close the first preview
-> Error seems to be triggered when the top of the block decoration of the remainding decoration enters or exits the viewable window. NOTE that if you close the second preview, but keep the first, the error is not triggered `¯\_(ツ)_/¯`
1. Close both preview
-> The error no longer appears.

**Atom**: 1.19.3 x64
**Electron**: 1.6.9
**OS**: Mac OS X 10.12.6
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Cannot read property 'id' of undefined

```
At /Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:1185

TypeError: Cannot read property 'id' of undefined
    at TextEditorComponent.addBlockDecorationToRender (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:1185:67)
    at TextEditorComponent.addDecorationToRender (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:1026:22)
    at decorationsByMarker.forEach (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:995:20)
    at Map.forEach (native)
    at TextEditorComponent.queryDecorationsToRender (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:990:31)
    at TextEditorComponent.updateSyncBeforeMeasuringContent (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:381:16)
    at TextEditorComponent.updateSync (/Users/valentinaspuler/Downloads/Atom.app/Contents/Resources/app/src/text-editor-component.js:276:16)
[...]

PR #15550, opened 2017-09-05
Don't render block decorations located outside the visible range

as-cii:
Fixes https://github.com/atom/atom/issues/15441

Previously, when trying to use block decorations on non-empty markers, Atom could sometimes throw an error if such markers ended or started at a position that was not currently rendered.

In fact, even if we already restricted the decoration query to markers that intersected the visible row range, markers that were only partially visible would still be considered for rendering. If, depending on the `reversed` property, we decided to render the tail or head of the marker in question and this was outside the viewport, Atom would throw the aforementioned exception.

This pull-request addresses the above issue by explicitly ignoring block decorations that are located on rows that are not yet rendered.

/cc: @nathansobo @Ben3eeE @rsese @ungb

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2388,6 +2388,31 @@ describe('TextEditorComponent', () => {
       ])
     })
 
+    it('does not attempt to render block decorations located outside the visible range', async () => {
+      const {editor, component} = buildComponent({autoHeight: false, rowsPerTile: 2})
+      await setEditorHeightInLines(component, 2)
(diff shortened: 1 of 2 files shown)


Ben3eeE:
Can confirm that this fixes the issue with the steps to reproduce that I had 🚢

nathansobo:
Nice find @as-cii 💥

rsese:
Also confirmed fix on macOS 10.12.6 using https://github.com/atom/atom/releases/tag/untagged-e8e64b7f05f18088017b 💻

PR #15554, opened 2017-09-06
Allow Promises to be returned by a packages deactivate method

damieng:


damieng:
All tests and package tests now passing for this, would love to get some extra eyes on it @atom/core

spec/package-manager-spec.coffee
```diff
@@ -1145,44 +1165,60 @@ describe "PackageManager", ->
...
+      thrownError = null
 
       waitsForPromise ->
         atom.packages.activatePackage("package-that-throws-on-deactivate")
 
+      waitsForPromise ->
+        try
+          atom.packages.deactivatePackage("package-that-throws-on-deactivate")
+        catch error
+          thrownError = error
```
smashwilson: I don't think this `catch` clause will ever be triggered. What you probably want is:

```coffee
waitsForPromise ->
  atom.packages.deactivatePackage("package-that-throws-on-deactivate").catch(e -> error = e)
```

maxbrunsfeld (commented):
Looks great so far. I would just like to see a test in `package-manager-spec.coffee` for the new behavior. I'm thinking we'll have a fixture package whose `deactivate` hook returns a promise, and we'll assert that the promise returned by `deactivatePackages` does not resolve until that promise resolves.

src/package-manager.js
```diff
@@ -744,21 +744,21 @@ module.exports = class PackageManager {
...
 
   // Deactivate all packages
-  deactivatePackages () {
-    this.config.transact(() => {
-      this.getLoadedPackages().forEach(pack => this.deactivatePackage(pack.name, true))
-    })
+  async deactivatePackages () {
+    await this.config.transactAsync(() =>
+      Promise.all(this.getLoadedPackages().map(pack => this.deactivatePackage(pack.name, true)))
+    )
```
smashwilson: :+1:

src/package.coffee
```diff
@@ -506,15 +506,18 @@ class Package
...
       try
-        @mainModule?.deactivate?()
-        @mainModule?.deactivateConfig?()
-        @mainActivated = false
-        @mainInitialized = false
+        result = Promise.resolve(@mainModule?.deactivate?()).then =>
+          @mainModule?.deactivateConfig?()
+          @mainActivated = false
+          @mainInitialized = false
       catch e
```
smashwilson: This `catch` block won't catch errors thrown from `deactivate()`. You'd need a `.catch` on the Promise to handle async errors instead:

```coffee
result = Promise.resolve(@mainModule?.deactivate?()).then =>
  @mainModule?.deactivateConfig?()
  @mainActivated = false
  @mainInitialized = false
result = result.catch e =>
  # Async errors from the promise
  console.error "Error deactivating package '#{@name}', e.stack
```

smashwilson (dismissed):
I found a few subtle things related to try/catch and Promise interactions. Overall looking good :+1:

PR #15556, opened 2017-09-06
replace async/await with promises

leroix:
The version of node that the snapshot tool uses doesn't support es7 features, so I've attempted to remove async/await from `dg-package-async-deactivate`. This should work, but I haven't tested.

### DIFF
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -77,15 +77,16 @@ module.exports = class PackageManager {
     this.themeManager = themeManager
   }
 
-  async reset () {
+  reset () {
     this.serviceHub.clear()
-    await this.deactivatePackages()
-    this.loadedPackages = {}
-    this.preloadedPackages = {}
-    this.packageStates = {}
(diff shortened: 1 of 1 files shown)


src/package-manager.js
```diff
@@ -744,24 +745,26 @@ module.exports = class PackageManager {
...
+  deactivatePackage (name, suppressSerialization) {
     const pack = this.getLoadedPackage(name)
     if (!suppressSerialization && this.isPackageActive(pack.name)) {
       this.serializePackage(pack)
     }
-    await pack.deactivate()
-    delete this.activePackages[pack.name]
-    delete this.activatingPackages[pack.name]
-    this.emitter.emit('did-deactivate-package', pack)
+    return pack.deactivate().then(_ => {
```
leroix: Looks like we need to check here if `pack.deactivate()` returns a promise

leroix:
close in favor of https://github.com/atom/atom/pull/15557

PR #15557, opened 2017-09-06
Use the Node version bundled in Electron to verify snapshot script

as-cii:
Previously, we used to verify the snapshot script by running it in a new, empty context (similar to the one that `mksnapshot` creates when generating the startup blob).

However, this context was being created using the Node version that `script/build` was executed with. Such version may not match the Node version shipped with Electron, and could thus cause the build script to report "false negatives" when verifying the snapshot script. For instance, running `script/build` with Node 4 would cause it to throw an error when encountering keywords like `async`/`await`, even if they're 100% supported in Electron 1.6.9.

With this pull-request we are changing the snapshot verification code to use the Node version bundled in Electron in order to avoid the aforementioned mismatches.

@damieng @leroix: I believe this will fix the errors you were encountering on #15554.

/cc: @atom/maintainers

### DIFF
--- a/script/verify-snapshot-script
+++ b/script/verify-snapshot-script
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+const fs = require('fs')
+const vm = require('vm')
+const snapshotScriptPath = process.argv[2]
+const snapshotScript = fs.readFileSync(snapshotScriptPath, 'utf8')
+vm.runInNewContext(snapshotScript, undefined, {filename: snapshotScriptPath, displayErrors: true})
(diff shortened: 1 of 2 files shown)


script/verify-snapshot-script
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
```
hultberg: Isn't naming this file verify-snapshot-*script* a little redundant when it already is in a folder named script?

as-cii: "Script" in this case refers to the snapshot script that gets generated by electron-link, because we want to verify it can be run via `mksnapshot`. Omitting it would be misleading, as `verify-snapshot` seems to indicate we're about the verify the actual snapshot blob, as opposed to the script it is generated from.

PR #15562, opened 2017-09-06
Update README.md

satyamramawat:
Edited Document for mac user, to run ATOM after mac from folder.

### Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Download the latest [Atom release](https://github.com/atom/atom/releases/latest)
 
 Atom will automatically update when a new release is available.
 
+After installing follow this step:
+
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for the contribution.

We don't want the README to get too cluttered, so we only want the basic installation instructions in it. We have more detailed installation instructions for each platform in the [Atom Flight Manual]http://flight-manual.atom.io/getting-started/sections/installing-atom/). These instructions mention the simpler way of doing what you're suggesting, which is to use the `Window: Install Shell Commands` command from the Command Palette.

Because of the reasons mentioned above, we can't accept this change. Thanks for your help!

PR #15564, opened 2017-09-06
Update stale.yml

xtaraim:
### Requirements


lee-dohm:
Thanks for contributing.

I'm not sure what the intent is of these changes since you declined to fill in the PR template. In any case, we like the way we currently have [probot-stale](https://github.com/probot/stale) configured  :grinning: Because of this, we're not going to accept this PR.

Thanks again.

xtaraim:
Turn off is much better than closed..

PR #15580, opened 2017-09-07
Flush scroll position to dummy scrollbar components on re-attach

### LINKED ISSUE #15552: Opening a split causes original pane's scroll position to reset
### Prerequisites

### Description

After splitting a pane, the original pane's scroll position appears to reset to (0, 0).  This is most noticeable when attempting to horizontally scroll the original pane.

file-position-reset

### Steps to Reproduce

1. Open a file
2. `pane:split-<direction>-and-copy-active-item` (note: I have only tested with `right`)
3. If there is no horizontal scrollbar, resize the panes so that the original pane can be horizontally scrolled
4. Look at the scrollbar positions of the original pane and then attempt to horizontally scroll

**Expected behavior:** Scrollbars to correctly represent where the current file is, and for horizontally scrolling to horizontally scroll the file while maintaining the vertical position

**Actual behavior:** Both scrollbars reset to (0, 0) - that is, the vertical scrollbar resets to the top and the horizontal scrollbar resets to the left.  When attempting to horizontally scroll, the file position is reset to the top of the file.

### Versions
```
Atom    : 1.22.0-dev-700fdc592
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

/cc: @as-cii @nathansobo

as-cii:
Fixes #15552

This prevents the dummy scrollbars from resetting their position to `0` when the editor element is moved elsewhere in the DOM (e.g. when splitting a pane item).

/cc: @nathansobo @50Wliu @Ben3eeE @rsese @ungb

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -378,35 +378,50 @@ describe('TextEditorComponent', () => {
       expect(horizontalScrollbar.style.visibility).toBe('')
     })
 
-    it('updates the bottom/right of dummy scrollbars and client height/width measurements without forgetting the previous scroll top/left when scrollbar styles change', async () => {
-      const {component, element, editor} = buildComponent({height: 100, width: 100})
-      expect(getHorizontalScrollbarHeight(component)).toBeGreaterThan(10)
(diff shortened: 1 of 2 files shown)


nathansobo:
:+1:

PR #15596, opened 2017-09-09
Increase scrollbar contrast in dark themes

### LINKED ISSUE #15401: Dark scrollbars over dark syntax themes in macOS (only in editor)
### Prerequisites

### Description

When using a dark syntax theme, the scrollbars in macOS are too dark to be seen correctly. This only happens in the editor (1). For the tree view (2), for example, the color of the scrollbars is as expected (light themed).

complete

### Steps to Reproduce

1. Select **When scrolling** setting in *macOS Preferences > General > Show scroll bars*.
2. Select a **dark syntax theme** in Atom.

**Expected behavior:** Editor uses light scrollbar on syntax themes with dark backgrounds.

**Actual behavior:** Editor uses dark scrollbar on syntax themes with dark backgrounds, rendering them almost invisible.

**Reproduces how often:** I've test it with all default UI and syntax themes combinations and problem persists.

### Versions

Atom    : 1.19.2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.11.0

OS: macOS Sierra 10.12.6

simurai:
### Description of the Change

In #15401 @haidran noticed that the native "only when scrolling" scrollbars of the tree-view have a light handle with dark themes, but not in buffers. Turns out that Chromium/macOS switches between a light and dark handle based on the `background-color` of the element that has `overflow` scrolling enabled. Unfortunately we can't just add a background color to the scrollbars of `atom-text-editor` because it would cover the code underneath.

As a potential solution, this PR adds the syntax background color to scrollbars initially only for a very short time through a keyframe animation and then removes the background color again. This will have the effect that the scrollbars still pick up the right syntax background color and adjust the handle accordingly for dark themes.

Before | After
--- | ---
screen shot 2017-09-09 at 12 47 01 pm | screen shot 2017-09-09 at 12 46 02 pm

### Alternate Designs

Instead of an animation (adding and removing the background color) we could also keep an almost transparent background color, like `background-color: fade(@syntax-background-color, 0.01%);`. It won't be visible by eye. But using an animation feels still slightly better since it's more temporarily. 

Another alternative would be to implement our own custom scrollbars, but that's much bigger in scope.

Since all dark themes are affected.

### Benefits

Better buffer scrollbar handle contrast for dark themes.

### Possible Drawbacks

A keyframe animation might have negative side effects on the buffer, like causing repaint or so. /cc @nathansobo @as-cii 

### Applicable Issues

Fixes #15401

### DIFF
--- a/static/text-editor.less
+++ b/static/text-editor.less
@@ -145,3 +145,17 @@ atom-overlay {
   display: block;
   z-index: 4;
 }
+
+// This adds the background color to scrollbars for only a very short time.
+// So that the native macOS "When scrolling" scrollbars adjust for dark themes.
+.platform-darwin .scrollbars-visible-when-scrolling {
+  atom-text-editor {
+    .vertical-scrollbar,
+    .horizontal-scrollbar {
+      @keyframes scrollbars-visible-when-scrolling-fix {
+        0% { background-color: @syntax-background-color; }
+      }
+      animation: scrollbars-visible-when-scrolling-fix 0.01s;
+    }
+  }
+}


nathansobo:
I do wonder if this could be done more simply in JS, such as briefly applying a class to the scrollbars when they are initially rendered.

the-j0k3r:
This will surely help my problem, Would be extra nice to have a scrollbar that increases width when hover. Atm in a dark theme its somewhat hard to grab the nib without.

haidran:
Any update on this? 😕

simurai:
@haidran Until there is a cleaner JS alternative for this fix, you could paste the following in your [styles.less](http://flight-manual.atom.io/using-atom/sections/basic-customization/#style-tweaks) file:

```less
// This adds the background color to scrollbars for only a very short time.
// So that the native macOS "When scrolling" scrollbars adjust for dark themes.
// See https://github.com/atom/atom/pull/15596
.platform-darwin .scrollbars-visible-when-scrolling {
  atom-text-editor {
    .vertical-scrollbar,
    .horizontal-scrollbar {
      @keyframes scrollbars-visible-when-scrolling-fix {
        0% { background-color: @syntax-background-color; }
      }
      animation: scrollbars-visible-when-scrolling-fix 0.01s;
    }
  }
}
```

simurai:
Closing this as "too risky". See the above snippet if you wanna give it a try.

PR #15598, opened 2017-09-09
[WIP] Do not normalize C: to C:.

winstliu:
### Requirements

### Description of the Change

For unknown reasons, `path.normalize` will add a period to the end of a standalone drive letter when normalizing it.  This results in a path such as `C:.`, which, while accepted in Windows Explorer, is treated as `.` by other programs such as Powershell and `path.basename`.  This PR manually fixes the path normalization when the path is _only_ a drive letter on Windows.

### Alternate Designs

Ideally, this should be fixed upstream in Node.

Because this is a part of the fundamental `atom.project.setPaths` API.

### Benefits

Tree View, with a few additional fixes, should function correctly when being set to a root drive path.

### Possible Drawbacks

None?

### Applicable Issues

Refs #903

### DIFF
--- a/src/default-directory-provider.coffee
+++ b/src/default-directory-provider.coffee
@@ -55,4 +55,10 @@ class DefaultDirectoryProvider
         "#{matchData[1].toUpperCase()}#{uri.slice(1)}"
       else
         uri
-    path.normalize(pathWithNormalizedDiskDriveLetter)
+
+    # On Windows, a bare drive such as 'C:' gets normalized to 'C:.'
+    normalizedPath = path.normalize(pathWithNormalizedDiskDriveLetter)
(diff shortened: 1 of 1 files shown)


damieng:
I wonder if it should be normalized to C:\

50Wliu:
So this gets normalized _again_ in Pathwatcher...

50Wliu:
Unfortunately, this has been made intentional in Node v6.11.3, released one week ago.
https://github.com/nodejs/node/pull/14440/files#diff-b3745c453fcea0dda943b412adff1e7cR429

I'd rather try to work with Node's implementation then adding checks like this everywhere, so I'm closing this PR for now and will see what I can do in tree-view.

PR #15603, opened 2017-09-10
Spelling

jsoref:
### Description of the Change

Spelling fixes

### Alternate Designs

None of the misspellings appeared particularly intentional.

Misspellings hurt project health.

### Benefits

It's easier to use healthier projects, and there are fewer errors when people try to write code with the expected spelling (which doesn't match the used misspelling).

### Possible Drawbacks

It's possible that a misspelling is intentional, or that it's an APIand thus is hard to change (JavaScript and stylesheet are the only likely candidates in this series).


50Wliu:
:100: Thanks @jsoref!

PR #15617, opened 2017-09-11
Added readme warning for #14909

andrew-aladev:
Please do not hide such important issues from users. Give them a chance to make backup of their settings. #14909


andrew-aladev:
#14909

eparejatobes:
I don't care about the exact wording, but the fact that currently user settings can be completely wiped out during normal use should be explicitly and clearly noted.

lee-dohm:
Thanks for the feedback.

We don't currently nor have we ever had warnings for any of our open issues in our README. We don't believe that this is the best place to notify people about important issues and don't believe that this issue warrants preemptive warnings in any case.

Thanks again for the submission.

PR #15637, opened 2017-09-13
Update null-grammar.js

keangkai:
if have comment will easy to read.but i think this code good

### Requirements


50Wliu:
Thanks for the contribution, but as there is no loop where you're commenting, I don't believe this comment is appropriate.

Thanks for your interest in Atom!

PR #15653, opened 2017-09-14
Update panel.js

keangkai:
for easy to read

### Requirements


50Wliu:
Thank you for the pull request.  However, the Atom team prefers using line comments rather than block comments for documentation purposes.

Again, thanks for your interest in Atom!

PR #15669, opened 2017-09-16
:memo: Fix cursor name letter casing

BenRussert:
### Description of the Change

The docs currently say `Cursor` instead of `cursor`, just a quick fix there.

As discussed with @50Wliu on slack


50Wliu:
:100:

PR #15702, opened 2017-09-20
Finish up #14456 (Git refresh backgrounding)

BinaryMuse:
Continuation of https://github.com/atom/atom/pull/14456 by @Xapphire13 including changes from https://github.com/atom/atom/pull/15415

Closes https://github.com/atom/atom/pull/14456
Closes https://github.com/atom/atom/pull/15415


BinaryMuse:
@maxbrunsfeld @nathansobo Might I ask one of you to review this? Prior conversation can be found at https://github.com/atom/atom/pull/14456. :bowing_woman:

src/repository-status-handler.coffee
```diff
@@ -1,36 +1,44 @@
...
+    paths = paths or []
+
+    repo = Git.open(repoPath)
+    statuses = {}
+    upstream = {}
+    submodules = {}
+    branch = null
+
+    try
+      if repo?
```
maxbrunsfeld: We check if `repo` is null, but it looks like even if it is null, we will call `.release` on it in the `finally` block. Maybe the `if` should be moved outside of the `try/finally`, like

```coffee
if repo?
  try
    # ...
  finally
    repo.release()
```

src/repository-status-handler-helper.coffee
```diff
@@ -0,0 +1,30 @@
...
+    responseSub = handlerInstance.on repoPath, (result) ->
+      responseSub.dispose()
+      resolve(result)
+
+    handlerInstance.send {repoPath, paths}
+
+module.exports = {
+  terminateHandler,
+  refreshStatus
+}
```
maxbrunsfeld: It would be nice from a style perspective to export a `RepositoryStatusHandler` *class* here, rather than these top-level functions that use a global variable. Then we would instantiate the class in the constructor of `GitRepositoryProvider` (which only gets called once per Atom window) and pass the `RepositoryStatusHandler` instance to the constructor of `GitRepository`.

That said, there are other global variables in Atom, so it's not that big a deal.

maxbrunsfeld (commented):
Nice job tracking down this performance issue and thanks for the great PR @Xapphire13!

@BinaryMuse Thanks so much for picking this up again.

src/repository-status-handler-helper.coffee
Xapphire13: The intent was to have the handlerInstance variable hidden in this module in such a way where there can be only one per runtime (if my understanding of modules is correct, no matter how many times you require a module, it will only be initialized once per runtime).

The same could be achieved using a class with either static variable, or a class using a module global. Which would you prefer?

src/repository-status-handler.coffee
Xapphire13: Good catch! I didn't even realize that.

@BinaryMuse, since you opened this PR with my old changes, how do I go about making updates?

BinaryMuse:
@maxbrunsfeld Is https://github.com/atom/atom/pull/15702/commits/b01835adb2f69935a0a67c2f703408721842353e more what you had in mind?

maxbrunsfeld:
Yeah, exactly 💎 .

Xapphire13:
@BinaryMuse @maxbrunsfeld 

So now if you open another window of Atom, you will spin up another worker process correct? Do atom windows share the same runtime, or do they have their own?

maxbrunsfeld:
@Xapphire13 Yeah, but that was always the case. Each Atom window is its own process and has its own JavaScript environment.

BinaryMuse:
@maxbrunsfeld 

> Should we actually do something to ensure that only one worker process gets created for the application?

Yeah, this was one of the concerns I had in https://github.com/atom/atom/pull/14456#issuecomment-314197989. Maybe it's not that big of a deal? The GH package already creates a new renderer process per Atom window right now; this is a much more lightweight process that's spawned...

Is there any precedent for using `Task`s in the main process? There'd need to be some additional routing from the main process to the various renderer processes... maybe through the `ApplicationDelegate`? But maybe it wouldn't be that crazy.

Xapphire13 (approved):

maxbrunsfeld:
We solved this via https://github.com/atom/atom/pull/15750 (refreshing git status in a thread via some new native APIs instead of a sub-process).

PR #15713, opened 2017-09-21
Update syntax-related code to use new TextBuffer LanguageMode API (part 1)

maxbrunsfeld:
Refs https://github.com/atom/text-buffer/pull/260

This is the first in a series of PRs to make Atom's syntax related features more flexible.

Currently, code related to auto-indentation and code folding is mostly implemented in a private class called `LanguageMode`, which is constructed with a `TextEditor` instance, and implements most of its functionality by directly manipulating the editor.

This PR removes the `LanguageMode` class and reimplements its functionality in the existing `TokenizedBuffer` class, in a way that does *not* need to manipulate the editor directly. For example, instead of performing folds directly, the `TokenizedBuffer` exposes several pure methods for determining which buffer ranges should be folded.

I converted the `TokenizedBuffer` and its test to JS. For the tests that were previously in `language-mode-spec.coffee`, I have either moved them into `tokenized-buffer-spec.js` or moved them into a new test file - `text-editor-test.js`. I created that new file because I had already converted these tests to JS before realizing that some of them were really about `TextEditor` and not `TokenizedBuffer`, and I didn't want to convert the entirety of `text-editor.coffee` to JS yet. Since the file is so large, we may want to do the conversion gradually.

/cc @nathansobo


nathansobo:
Nice to see some conversions to JS. Liking your overall plan.

PR #15750, opened 2017-09-27
Avoid process spawn overhead for refreshing git status

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/9544
Depends on https://github.com/atom/git-utils/pull/80


nathansobo:
Hot-fixing this to 1.21.1.

PR #15775, opened 2017-09-29
Update the stale markComment message

Ben3eeE:
This is a first draft including what I would like to add to the bot message after reading feedback.

/cc: @lee-dohm @rsese for 💭s about the wording and contents.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -17,7 +17,7 @@ markComment: >
 
   This issue has been automatically marked as stale because it has not had
   recent activity. Because the Atom team treats their issues
-  [as their backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog) stale issues
+  [as their backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), stale issues
   are closed. If you would like this issue to remain open:
 
   1. Verify that you can still reproduce the issue in the latest version of Atom
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -13,9 +13,14 @@ exemptLabels:
 staleLabel: stale
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
+  Thanks for your contribution!
+
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. Because the Atom team treats their issues
+  [as their backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog) stale issues
+  are closed. If you would like this issue to remain open then
+  comment in the issue and verify that you can still reproduce in the latest version.
+  Issues that are labeled as triaged will not be automatically marked as stale.
 # Comment to post when removing the stale label. Set to `false` to disable
 unmarkComment: false
 # Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


lee-dohm (approved):

.github/stale.yml
```diff
@@ -13,9 +13,20 @@ exemptLabels:
 staleLabel: stale
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
+  Thanks for your contribution!
+
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. Because the Atom team treats their issues
+  [as their backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog) stale issues
```
50Wliu: I would recommend a comma after `as their backlog`

.github/stale.yml
```diff
@@ -13,9 +13,20 @@ exemptLabels:
...
+  are closed. If you would like this issue to remain open:
```
50Wliu: Maybe we can be more explicit and say `are closed after one year of inactivity`.

.github/stale.yml
```diff
@@ -13,9 +13,20 @@ exemptLabels:
...
+
+  1. Verify that you can still reproduce the issue in the latest version of Atom
+  1. Comment that the issue is still reproducible and include:
+    * What version of Atom you reproduced the issue on
+    * What OS and version you reproduced the issue on
+    * What steps you followed to reproduce the issue
+
+  Issues that are labeled as triaged will not be automatically marked as stale.
```
50Wliu: Is this necessary?

.github/stale.yml
lee-dohm: I'd prefer to not have to update the text here every time that we decide to change the time windows 😕

.github/stale.yml
lee-dohm: Fixed

.github/stale.yml
lee-dohm: Not necessary, though @Ben3eeE felt that it would be helpful because it is a common question.

Ben3eeE: Yeah there was some confusion around this. Like "This is labeled as bug does that mean...". I don't think it hurts to be explicit about it so that users will know. We do also not automatically close security and regression labeled issues but I intentionally left those out because then it was too complicated and getting your issue labeled as triaged seems like the most common thing to prevent the bot from automatically closing it.

Ben3eeE:
I think the changes made are good 👍 Thank you @lee-dohm and @50Wliu for reviewing 🙇

.github/stale.yml
```diff
@@ -13,9 +13,20 @@ exemptLabels:
...
 markComment: >
+  Thanks for your contribution!
+
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. Because the Atom team treats their issues
+  [as their backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), stale issues
+  are closed. If you would like this issue to remain open:
+
```
rsese: ❤️ I feel like that was a big issue for people.  Even though it was implied that more activity would keep it open, I like us being explicit and saying "here's what you can do if you want this to stay open".

lee-dohm:
Sounds like everyone is generally 👍 on this change, so I'm going to merge it. Thanks for taking this on @Ben3eeE :bow:

PR #15776, opened 2017-09-29
Fix slowness in Cursor.getCurrentWordBufferRange

maxbrunsfeld:
@leroix and I noticed this when doing some profiling https://github.com/atom/autocomplete-plus/pull/886.

Uses https://github.com/atom/superstring/pull/35


src/cursor.js
```diff
@@ -0,0 +1,753 @@
...
+    if (moveToEndOfSelection && !range.isEmpty()) {
+      this.setScreenPosition(range.start)
+    } else {
+      let {row, column} = this.getScreenPosition()
+
+      while (columnCount > column && row > 0) {
+        columnCount -= column
+        column = this.editor.lineLengthForScreenRow(--row)
+        columnCount--
+      } // subtract 1 for the row move
```
jasonrudolph: It looks like `decaffeinate` put this comment on the wrong line. I think it should be on the previous line:

```javascript
  columnCount-- // subtract 1 for the row move
}
```

maxbrunsfeld: Thanks!

PR #15777, opened 2017-09-29
Fix mouseup listener cleanup when dragging in text editor

matthewwithanm:
This listener is added with `{capture: true}`, which means it'll only be removed if `removeEventListener()` is passed `{capture: true}` too. Without that (as it is currently), the listener is never removed and each mouseup event ends up triggering the handler one more time than the last. Eventually, this leads to slowdown. 😞

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -1883,7 +1883,7 @@ class TextEditorComponent {
 
     function didMouseUp () {
       window.removeEventListener('mousemove', didMouseMove)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks for the fix!

nathansobo:
@maxbrunsfeld do you think we should hot-fix this?

maxbrunsfeld:
Yeah. Just cherry-picked it to all the release branches.

PR #15785, opened 2017-10-01
:art: Shortened last three methods in color class

arnavb:
Used the ternary operator to shorten the last three methods in color.js

### Description of the Change

This was a small refactoring to shorten the last three methods in the color class.

### Alternate Designs

N/A

N/A

### Benefits

It doesn't sacrifice clarity, but just reduces unnecessary sizeof code.

### Possible Drawbacks

Possibly readability, although that shouldn't be too big an issue.

### Applicable Issues

None

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -112,27 +112,15 @@ export default class Color {
 
 function parseColor (colorString) {
   const color = parseInt(colorString, 10)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
:+1:

Issue #15263, opened 2017-08-11
Uncaught Error: Requested measurement of a line component that is not currently rendered


robertung:
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.19.0 x64
**Electron**: 1.6.9
**OS**: Microsoft Windows 10 Enterprise
**Thrown From**: Atom Core

### Stack Trace

Uncaught Error: Requested measurement of a line that is not currently rendered

```
At C:\Users\sarr\AppData\Local\atom\app-1.19.0\resources\app\src\text-editor-component.js:2173

Error: Requested measurement of a line that is not currently rendered
    at horizontalPositionsToMeasure.forEach (~/AppData/Local/atom/app-1.19.0/resources/app/src/text-editor-component.js:2171:29)
    at Map.forEach (native)
    at TextEditorComponent.measureHorizontalPositions (~/AppData/Local/atom/app-1.19.0/resources/app/src/text-editor-component.js:2164:45)
    at TextEditorComponent.measureContentDuringUpdateSync (~/AppData/Local/atom/app-1.19.0/resources/app/src/text-editor-component.js:385:16)
    at TextEditorComponent.updateSync (~/AppData/Local/atom/app-1.19.0/resources/app/src/text-editor-component.js:276:18)
    at TextEditorComponent.didMouseWheel (~/AppData/Local/atom/app-1.19.0/resources/app/src/text-editor-component.js:1509:43)
```

### Commands

```
     -1:18.7.0 core:save (input.hidden-input)
     -1:14.6.0 tree-view:duplicate (span.name.icon.icon-file-text)
  2x -1:13.8.0 core:move-right (input.hidden-input)
     -1:11.6.0 core:confirm (input.hidden-input)
     -1:10.6.0 core:close (input.hidden-input)
     -1:10 core:select-all (input.hidden-input)
     -1:09.6.0 core:paste (input.hidden-input)
     -1:09.2.0 core:save (input.hidden-input)
  2x -0:54.4.0 core:backspace (input.hidden-input)
     -0:53.2.0 core:save (input.hidden-input)
     -0:47.7.0 editor:select-to-beginning-of-word (input.hidden-input)
     -0:47 core:move-right (input.hidden-input)
  9x -0:46.3.0 core:move-down (input.hidden-input)
     -0:44.5.0 core:move-up (input.hidden-input)
  2x -0:40.7.0 core:undo (input.hidden-input)
     -0:30.7.0 core:save (input.hidden-input)
[...]

PR #15800, opened 2017-10-03
Fix rendering bug when folds cause the vertical scrollbar to disappear with soft wrapping enabled

### LINKED ISSUE #15549: Expanded code-blocks overlap with previous code
### Prerequisites

### Description

When expanding a collapsed code-block while the current viewport isn't scrollable (i.e. all code that is shown fits on the screen), the expanded code overlaps with the code right among it (See screenshot). Only re-opening the file helps.

**Reproduces how often:** Every time on the code I discovered it, never on some simple, short code.

### Versions

Atom    : 1.19.3
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

### Screenshot

atomcollapsebug

### LINKED ISSUE #15761: Atom editor renders random text over each other.
### Prerequisites

### Description

Atom editor renders random text over each other.

I am currently unable to get repro steps for this but have screenshots of the scenario

Text should show up and other lines shouldn't be rendered on top

Additional random lines are rendered above current lines and don't disappear after scrolling.
image

image

I can't repro this again and unsure what I did to get to this scenario

### Versions

```
Atom    : 1.22.0-dev-1880e1401
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

```

### Additional Information

While scrolling you can see it updates the line numbers but there's additional lines that just stay there.
scrolling

/CC @nathansobo @as-cii

nathansobo:
Closes #15761
Closes #15549
Refs #15263

Due to the fact that the soft-wrapping boundary can change due to a variety of different factors (scrollbars appearing or disappearing, font size changes, editor dimensions changing, etc), we decided to detect changes in the soft-wrap boundary at the start of every frame before proceeding with any other rendering.

This works fine in principle, but prior to this PR it was causing us to write inconsistent values for the `firstRenderedScreenRow` to the per-frame derived dimensions cache in certain edge cases, causing rendering bugs downstream. This could occur when folding caused a document that was scrolled down partially to no longer be tall enough to scroll. This would cause the following unfortunate series of events:

* We would detect the scrollbar becoming invisible due to the shrinking of the content height below the editor height and update the soft-wrap width in [`updateModelSoftWrapColumn`](https://github.com/atom/atom/blob/3a3c58e04e37f3fe653d742134a0067bc8adb8dc/src/text-editor-component.js#L2099).
* Updating the soft wrap boundary would [*reset* the display layer](https://github.com/atom/atom/blob/3a3c58e04e37f3fe653d742134a0067bc8adb8dc/src/text-editor-component.js#L2104), causing it to recompute itself in the background.
* We would then then immediately compute the soft wrap boundary [a second time](https://github.com/atom/atom/blob/3a3c58e04e37f3fe653d742134a0067bc8adb8dc/src/text-editor-component.js#L2106). We always compute the soft wrap boundary twice because the initial change in the soft wrap boundary could change scroll bar visibility, which causes the soft wrap boundary to change again. Unfortunately, we were computing the soft wrap boundary *without* forcing a synchronous update for the visible row range of the display layer, and thus not accounting for the presence of the fold and assuming the scrollbar was visible again.
* In the process of computing the soft wrap boundary a second time, we were caching the incorrect values for the first rendered row in `derivedDimensionsCache`.
* Immediately following the update of the soft wrap boundary, we would [populate the display layer for visible row range](https://github.com/atom/atom/blob/3a3c58e04e37f3fe653d742134a0067bc8adb8dc/src/text-editor-component.js#L365), causing all future derived dimensions to be based on a correctly populated display layer. This would cause a mix of derived values cached against two different versions of the display layer, leading to inconsistent rendering state.

The solution in this PR is to

* Compute the first rendered row prior to resetting the display layer.
* After resetting the display layer, immediately force it to synchronously populate itself for the visible row range before making any additional computations.

These two steps ensure that the `derivedDimensionsCache` is never populated with inconsistent values. I'm hopeful that this issue was also the source of measurement exceptions in #15263, though we'll have to wait to be sure since we never have reproduced them.

Huge props to @ben3eee and @ungb for their help tracking down this edge case. Thanks so much.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -286,6 +286,31 @@ describe('TextEditorComponent', () => {
       expect(lineNumberNodeForScreenRow(component, 0).querySelector('.foldable')).toBeNull()
     })
 
+    it('gracefully handles folds that change the soft-wrap boundary by causing the vertical scrollbar to disappear (regression)', async () => {
+      const text =  ('x'.repeat(100) + '\n') + 'y\n'.repeat(28) + '  z\n'.repeat(50)
+      const {component, element, editor} = buildComponent({text, height: 1000, width: 500})
+
+      element.addEventListener('scroll', (event) => {
+        event.stopPropagation()
+      }, true)
+
+      editor.setSoftWrapped(true)
+      jasmine.attachToDOM(element)
+      await component.getNextUpdatePromise()
+
+      const firstScreenLineLengthWithVerticalScrollbar = element.querySelector('.line').textContent.length
+
+      setScrollTop(component, 620)
+      await component.getNextUpdatePromise()
+
+      editor.foldBufferRow(28)
+      await component.getNextUpdatePromise()
+
+      const firstLineElement = element.querySelector('.line')
+      expect(firstLineElement.dataset.screenRow).toBe('0')
+      expect(firstLineElement.textContent.length).toBeGreaterThan(firstScreenLineLengthWithVerticalScrollbar)
+    })
+
     it('shows the foldable icon on the last screen row of a buffer row that can be folded', async () => {
       const {component, element, editor} = buildComponent({text: 'abc\n  de\nfghijklm\n  no', softWrapped: true})
       await setEditorWidthInCharacters(component, 5)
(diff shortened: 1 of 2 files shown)


Ben3eeE:
I was able to verify that this fixes https://github.com/atom/atom/issues/15549 💯 🚢

PR #15810, opened 2017-10-04
Update clipboard.js

NuttasitBoonwat:
### Requirements


maxbrunsfeld:
I don't think this is a beneficial change. In the future, please don't open a PR unless you can provide some explanation of the benefit.

PR #15812, opened 2017-10-04
Update cursor.js

NuttasitBoonwat:
### Requirements


pingping-n:
I think need "()" for statement.

maxbrunsfeld:
Replacing an `if` statement with a ternary isn't necessarily an improvement.

PR #15821, opened 2017-10-05
Clear dimensions cache when approximate screen line count changes

### LINKED ISSUE #15761: Atom editor renders random text over each other.

as-cii:
Fixes #15761

It is possible for the approximate screen line count to exceed the actual number of lines. This may happen if there are many soft-wraps and the display layer hasn't indexed the entire contents of the buffer. 

In that circumstance, if a user attempts to scroll to a position that does not exist, we are unable to clip it correctly in `setScrollTop` because the line count approximation is also wrong. Therefore, after populating the spatial index in `updateSyncBeforeMeasuringContent`, we will make sure that the approximate screen line count did not change. If it did, we need to clear the derived dimensions cache (which contains, for example, the rendered row range) because it could contain incorrect values.

/cc: @nathansobo @Ben3eeE @ungb @rsese


spec/text-editor-component-spec.js
```diff
@@ -894,6 +896,73 @@ describe('TextEditorComponent', () => {
...
+      beforeEach(() => {
+        originalTimeout = jasmine.getEnv().defaultTimeoutInterval
+        jasmine.getEnv().defaultTimeoutInterval = 60 * 1000
+      })
+
+      afterEach(() => {
+        jasmine.getEnv().defaultTimeoutInterval = originalTimeout
+      })
+
+      it('renders the visible rows correctly after randomly mutating the editor', async () => {
```
nathansobo: Thanks so much for adding this. We probably should have had something like this all along. Lesson learned. We should never again release a system this complex without randomized testing it.

src/text-editor-component.js
```diff
@@ -2884,10 +2884,25 @@ class TextEditorComponent {
...
-    const visibleTileCount = Math.ceil(editorHeightInTiles) + 1
-    const lastRenderedRow = renderedStartRow + (visibleTileCount * this.getRowsPerTile())
-    this.props.model.displayLayer.populateSpatialIndexIfNeeded(Infinity, lastRenderedRow)
+    const {model} = this.props
+    const previousScreenLineCount = model.getApproximateScreenLineCount()
+
+    const renderedEndRow = renderedStartRow + (this.getVisibleTileCount() * this.getRowsPerTile())
+    this.props.model.displayLayer.populateSpatialIndexIfNeeded(Infinity, renderedEndRow)
+
+    // It is possible for the approximate screen line count to exceed the actual
```
nathansobo: The first sentence of this comment is confusing to me. Isn't the problem that the approximate screen line count is actually *less* than the actual screen line count because it's not accounting for all the soft wraps? I wonder if we should just always clear the dimensions cache when going down this code path for simplicity.

maxbrunsfeld:
👏

PR #15828, opened 2017-10-05
Clear the dimensions cache after updating the soft wrap column

nathansobo:
Updating the soft wrap column could cause us to compute different values for derived dimensions, so any dimensions that were cached *in the process* of updating the soft wrap column need to be cleared.

This fixes a randomized test failure found be @Ben3eeE.

/cc @Ben3eeE @as-cii

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2118,6 +2118,7 @@ class TextEditorComponent {
       // rendered start row accurately. 😥
       this.populateVisibleRowRange(renderedStartRow)
       this.props.model.setEditorWidthInChars(this.getScrollContainerClientWidthInBaseCharacters())
+      this.derivedDimensionsCache = {}
 
       this.suppressUpdates = false
     }


Ben3eeE:
This seed is still failing: `const initialSeed = 1507231571985 // Fails with 10 random edits`
The others I found are fixed with this PR.

nathansobo:
@Ben3eeE It would be more helpful to get the value for `seed` instead of `initialSeed` to jump straight to the failure.

nathansobo:
Force pushed a rebase to hopefully resolve an unrelated issue.

I'm having trouble reproducing with @Ben3eeE's failing seed. Maybe it's a platform issue. @Ben3eeE could push another WIP pull request with the failing seed? I think this PR *is* an improvement so I'm inclined to proceed once we build green.

Ben3eeE:
Yeah this is an improvement 🚢 

I'll push a WIP PR with the failing seed to see what the platforms say.

PR #15871, opened 2017-10-11
forceHideTitleBar

sean-codes:
### Requirements

### Description of the Change

Way so we could please hide the title bar on Windows through a config value.

I created a custom titleBar package for atom: https://atom.io/packages/atom-windows-titlebar
To hide the titlebar it requires sort of a hack to change the frame: false in the atom-window.js

Now with asar we can not change the file inside at all because atom is locking the file

### Alternate Designs

A simple config value I think will let us do the trick. Possibly using a different config name would be appropriate.

Other developers have created custom titlebar packages. I think it adds a bit more hackability to the editor

### Benefits

More options for developers of packages

### Possible Drawbacks

Someone accidentally setting this value or the package crashing and then the titlebar gets hidden. Possibly there is another way to hide the electron frame during runtime by opening new BrowserWindow

### DIFF
--- a/src/main-process/atom-window.coffee
+++ b/src/main-process/atom-window.coffee
@@ -49,7 +49,7 @@ class AtomWindow
     if @shouldAddCustomInsetTitleBar()
       options.titleBarStyle = 'hidden-inset'
 
-    if @shouldHideTitleBar()
+    if @shouldHideTitleBar() or @forceHideTitleBar()
       options.frame = false
 
     @browserWindow = new BrowserWindow(options)
@@ -255,7 +255,11 @@ class AtomWindow
     not @isSpec and
     process.platform is 'darwin' and
     @atomApplication.config.get('core.titleBar') is 'hidden'
-
+  
+  forceHideTitleBar: ->
+    not @isSpec and
+    @atomApplication.config.get('core.titleBar') is 'forcehidden'
+   
   close: -> @browserWindow.close()
 
   focus: -> @browserWindow.focus()


juliancruzsanchez (approved):

IngwiePhoenix:
@sean-codes 
I saw your package and wanted to install it - but within your README you never actually elaborated if one should run `apm install` and where the `toggleframe.js` is coming from.

Other than that, I definitively agree. It is already possible on MacOS, and it is also possible on Windows - at least Windows 10, as far as I am aware. Atom claims to be cross-platform, so therefore it should be expected that features are as cross platform as possible - which is not the case for the title bar case.

sean-codes:
@IngwiePhoenix yes, run apm install from the folder toggleframe.js is in. It's also possible to install apm globally `npm install apm -g`!

charlesangus:
Poking on this merge request. Dark themes on Windows are ugly with the default white titlebar.

Any reason why this hasn't been merged in?

aminya:
The code should be updated:
https://github.com/atom/atom/blob/a239092460e9339ee4d56e5d73258ff7d04fd0d8/src/main-process/atom-window.js#L61:
```js
  if (this.shouldHideTitleBar() || this.forceHideTitleBar()) {
   options.frame = false;
 }
```

```js
  forceHideTitleBar() {
    return !this.isSpec &&
    (this.atomApplication.config.get('core.titleBar') === 'forcehidden');
  }
```

sadick254:
Hey @sean-codes, setting `titlebar` to `hidden` is now not restricted by `platform`. 
https://github.com/atom/atom/blob/f4c48f2951a9ed4faa83abddcee075316ba7d4b7/src/main-process/atom-window.js#L406-L411

I am closing this PR since we don't need an additional setting.

PR #15898, opened 2017-10-15
Decaffeinate `src/project.coffee`

jasonrudolph:
This change converts the `Project` class to JavaScript with [standard style](https://standardjs.com).


src/project.js
```diff
@@ -0,0 +1,705 @@
...
+  setPaths (projectPaths, options = {}) {
+    for (let repository of this.repositories) {
+      if (repository != null) repository.destroy()
+    }
+    this.rootDirectories = []
+    this.repositories = []
+
+    for (let path in this.watcherPromisesByPath) {
+      this.watcherPromisesByPath[path].then(watcher => { watcher.dispose() })
+    }
```
jasonrudolph: The CoffeeScript version of this code looked like this:

https://github.com/atom/atom/blob/8952cd315d44b7730ed686c2be14a9dc32d1e529/src/project.coffee#L235

[`decaffeinate`](https://www.npmjs.com/package/decaffeinate) translated that to the following JavaScript:

https://github.com/atom/atom/blob/25b7ddb328ab396e7ed53cad0775fdd0eba5ac74/src/project.js#L294

That seems like a non-equivalent translation. `watcherPromisesByPath` is [initialized as an object literal](https://github.com/atom/atom/blob/8952cd315d44b7730ed686c2be14a9dc32d1e529/src/project.coffee#L236), so `watcherPromisesByPath.length` is `undefined`. 

To address this issue, 5937b95b4969c89689fbbfdec3f77e40a01b9eb9 replaces `decaffeinate`'s translation with the `for` loop shown above.

src/project.js
```diff
@@ -0,0 +1,705 @@
...
+      if (this.rootDirectories.includes(directory)) {
+        this.emitter.emit('did-change-files', events)
+      }
+    })
+
+    for (let path in this.watcherPromisesByPath) {
+      if (!this.rootDirectories.find(dir => dir.getPath() === path)) {
+        this.watcherPromisesByPath[path].then(watcher => { watcher.dispose() })
+      }
+    }
```
jasonrudolph: I think there's a bug lurking in the current CoffeeScript version of this `for` loop:

https://github.com/atom/atom/blob/8952cd315d44b7730ed686c2be14a9dc32d1e529/src/project.coffee#L288-L290

`@rootDirectories` is an Array of Directory objects, and `root` is a String. Therefore, `@rootDirectories.includes(root)` will always evaluate to `false`. We instead need to look for a Directory object in `@rootDirectories` where the Directory object's path is equal `root`.

The JavaScript version above (implemented in 5ec9d0f1347eda1aecea72ab00348a56eb3d0ec6) fixes this issue. (Note that this PR also renames the `root` local to `path` to clarify that we know the item is a _path_, but we don't know whether it is a _root_ path.)

jasonrudolph (commented):
@smashwilson: This decaffeination effort involved some manual tweaking to the watcher-related code in the `Project` class. (See comments below.) Given your recent work on the watcher bits, would you mind offering a sanity check of the changes described in the comments below? I think the changes are reasonable, but I'd feel more confident about those changes if they had another set of :eyes:. :smile:

src/project.js
smashwilson: :+1: Your translation is spot-on to the intention of the original. It looks like I used an [`in` comprehension when I should have used an `of` loop](http://coffeescript.org/#loops) in the original!

src/project.js
smashwilson: Oh yikes. I remember getting a bit confused about which collections contained paths and which contained Directory objects... looks like I missed one!

> Note that this PR also renames the `root` local to `path` to clarify that we know the item is a _path_, but we don't know whether it is a _root_ path.)

I'm :+1: on renaming `root` to clarify this, but I'd advise against calling it `path`. I've hit too many problems with inadvertently shadowing or overwriting the import from node's "path" module :smile: How about `rootPath`?

jasonrudolph: @smashwilson: Thanks for the :eyes:

> I'm :+1: on renaming `root` to clarify this, but I'd advise against calling it `path`. I've hit too many problems with inadvertently shadowing or overwriting the import from node's "path" module :smile: 

Good point! a4ea46c57e078ae691d9aa9781948c39253ed4c9 renames it to `watchedPath`. (I chose `watchedPath` over `rootPath` because we know the path is watched, but we don't know if it's a root path.)

leroix:
☕️

PR #15906, opened 2017-10-16
Link to the flight-manual instead of duplicating documentation

Ben3eeE:
### Description of the Change

The documentation in the flight-manual is the same as this and the flight-manual is the preferred place to have documentation in. This changes the documentation to just be a link to the flight-manual.

/cc: @laurmurclar This answers your question about preferred location for documentation 🙂 

### Benefits

Not duplicating documentation but leave the old document in place in case there are links or bookmarks to it.


lee-dohm:
Thanks for taking care of this @Ben3eeE :bow:

PR #15916, opened 2017-10-17
Stop dragging only when user interacts with keyboard

as-cii:
Refs: https://github.com/atom/atom/pull/15407

Previously, we prevented users from dragging the selection further when the buffer was about to change. This was problematic because any change in the buffer, even one that was performed "automatically" by a package, would cancel the dragging action and result in a confusing experience for the user. On the other hand, we want to prevent users from accidentally selecting text when they perform an edit (see #15217, #15405).

This pull-request addresses both concerns by canceling the dragging as soon as the user interacts with the keyboard, instead of canceling the dragging when the buffer is about to change. One downside of this approach is that it changes the behavior of pressing a key that does not result in a buffer change, e.g. <kbd>Shift</kbd>, <kbd>Right-Arrow</kbd>, etc. In fact, pressing one of those keys will now prevent users from dragging the selection further. This seems like an acceptable tradeoff, but we are open to revisiting it in case it introduces pain in some other area of the system.

🍐'd with @jasonrudolph 

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -4422,7 +4422,7 @@ describe('TextEditorComponent', () => {
       expect(dragEvents).toEqual([])
     })
 
-    it('calls `didStopDragging` if the buffer changes while dragging', async () => {
+    it('calls `didStopDragging` if the user interacts with the keyboard while dragging', async () => {
       const {component, editor} = buildComponent()
 
       let dragging = false
@@ -4435,8 +4435,14 @@ describe('TextEditorComponent', () => {
       await getNextAnimationFramePromise()
       expect(dragging).toBe(true)
 
-      editor.delete()
+      // Buffer changes don't cause dragging to be stopped.
+      editor.insertText('X')
+      expect(dragging).toBe(true)
+
+      // Keyboard interaction prevents users from dragging further.
+      component.didKeydown({code: 'KeyX'})
       expect(dragging).toBe(false)
+
       window.dispatchEvent(new MouseEvent('mousemove'))
       await getNextAnimationFramePromise()
       expect(dragging).toBe(false)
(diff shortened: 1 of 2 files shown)


nathansobo (approved):
Looks great. We could potentially use `textInput` events instead of `keydown`, but I think this is probably fine.

Ben3eeE:
The current implementation also fixes https://github.com/atom/find-and-replace/issues/520 which is a bad experience for touchpad users.

I think that this makes sense and now the interruption is entirely in the users control. Instead of being interrupted by automatic changes that can be outside of their control. I vote that we ship this and revisit if we get issues for it.

as-cii:
> Looks great. We could potentially use textInput events instead of keydown, but I think this is probably fine.

Thanks for the review. Yeah, unfortunately I think using `keydown` is necessary because the user could hit <kbd>Backspace</kbd>, <kbd>Delete</kbd> or paste content too.

> I think that this makes sense and now the interruption is entirely in the users control. Instead of being interrupted by automatic changes that can be outside of their control. I vote that we ship this and revisit if we get issues for it.

Nice! Thanks for testing this.

as-cii:
I will cherry-pick this to beta only, so that it can go out on stable during the next railcar.

alexdevero:
@as-cii Is there a way to disable this? It causes issues when I want to select multiple selections of text - make one selection, hold CTRL and select another. Now, this no longer works.

Ben3eeE:
@alexdevero Can you open an issue and reference this PR? This is something that we want to address but it's easy that it will get lost if we don't have an issue to track.

nathansobo:
@Ben3eeE I can open an issue. This is a pretty serious regression.

alexdevero:
Thank you @nathansobo.

Ben3eeE:
@nathansobo I agree. I would have opened it myself as soon as I could if nothing else :grinning: 

I think we should address this while 1.22 is still in beta.

PR #15918, opened 2017-10-17
textChange(CoreSettings): remove redundancy

FalkF:
### Description of the Change

Just a text change, VCS (VersionControlSystem) already includes "system" :)
"Files and directories ignored by the current project's VCS system will be ignored by some packages, ..."


BinaryMuse:
Thanks!

PR #15930, opened 2017-10-18
Add x-scheme-handler/atom as MimeType in atom.desktop.in

BinaryMuse:
This is an attempt to register Atom as a default protocol client for `atom://` URIs on at least *some* Linux systems. I'm not very familiar with this and am not quite sure if this is even the right path; thoughts from @atom/maintainers (maybe @smashwilson?) welcome!


BinaryMuse:
This documentation seems to indicate that this will not work as written: https://developer.gnome.org/desktop-entry-spec/#extra-actions

BinaryMuse:
Linux support for protocol handlers has landed in Electron; see #16532

PR #15935, opened 2017-10-18
Add core URI handlers

### LINKED ISSUE #2037: open files with atom:// handler

BinaryMuse:
Atom core should include URI handlers for extremely common tasks, like opening files. This PR implements any such handlers. 💭 s for other useful handlers are welcome.

- [x] Open file (with optional line and column numbers) — `atom://core/open/file?filename=urlEncodedFileName&line=n&column=n`

Todo

- [x] Address https://github.com/charliesome/better_errors/pull/356

Closes #2037


BinaryMuse:
For the `/core/open/file` handler, it seems like it might be useful to do some URI parsing from the main process in order to call `AtomApplication#windowForPaths` or something similar to figure out which window we should route the URI event to.

On the other hand, package authors may wish to have access to similar information (that is, meta information about the various open Atom windows), so perhaps we should expose this across the application delegate somehow so it's accessible from renderer process code.

BinaryMuse:
This works, although it's not my *absolute* favorite strategy here. Open to 💭 s from @atom/maintainers if anyone has any :)

UziTech:
I registered the [`open`](https://atom.io/packages/open) package to allow a link with the [TextMate URL scheme](http://blog.macromates.com/2007/the-textmate-url-scheme/) to be open in Atom:

```
atom://open?url=file:///path/to/file.ext&line=10&column=2
```

One thing that would be really nice is if Atom could open an entire project that the file is in if it is not already open.

I'm not sure how we would define project, but that is my two cents.

src/main-process/atom-application.coffee
```diff
@@ -67,7 +67,7 @@ class AtomApplication
     {@resourcePath, @devResourcePath, @version, @devMode, @safeMode, @socketPath, @logFile, @userDataDir} = options
     @socketPath = null if options.test or options.benchmark or options.benchmarkTest
     @pidsToOpenWindows = {}
-    @windows = []
+    @windows = new WindowStack()
```
nathansobo: 🎨  Might be nice to rename `@windows` to `@windowStack` to help emphasize that it's an object, not an array.

Ben3eeE:
:clap: :tada:

nathansobo:
💥

src/main-process/atom-application.coffee
```diff
@@ -648,30 +655,39 @@ class AtomApplication
...
+
+      windowInitializationScript ?= require.resolve('../initialize-application-window')
       windowDimensions = @getDimensionsForNewWindow()
-      @lastFocusedWindow = new AtomWindow(this, @fileRecoveryService, {resourcePath, windowInitializationScript, devMode, safeMode, windowDimensions, env})
-      @lastFocusedWindow.on 'window:loaded', =>
-        @lastFocusedWindow.sendURIMessage url
+      win = new AtomWindow(this, @fileRecoveryService, {resourcePath, windowInitializationScript, devMode, safeMode, windowDimensions, env})
+      @windowStack.addWindow(win)
+      win.on 'window:loaded', ->
+        win.sendURIMessage url
```
UziTech: @BinaryMuse Any way we could use [HistoryManager](https://atom.io/docs/api/v1.21.2/HistoryManager#instance-getProjects) to check if `url` is in a project and load the entire project and open the `url`?

maybe we would want a setting to open individual files or project if found?

RobinDaugherty:
@BinaryMuse it seems that line number in the URL is treated as a zero-indexed number while all other editors (and the Atom UI) treat this number as one-indexed. So a link to `line=56` lands on line 57 in the UI. It seems like it might be best to change this to conform to other editors. Any preference?

BinaryMuse:
@RobinDaugherty Yes, I agree with you. I'll merge a change shortly to fix this. Thanks!

PR #15945, opened 2017-10-19
Use endsWith to match modules to exclude from the snapshot

winstliu:
### Requirements

### Description of the Change

There's a couple of common packages that need to be excluded from the snapshot.  Rather than listing each instance (since they may or may not be deduped), match using `endsWith`.

### Alternate Designs

None.

Already is.

### Benefits

Less code maintenance if a package is un-deduped or a new dependency of that package is added in a different module.  Also fixes an intermittent build issue for me where an additional fs-extra shows up in github/node_modules/nsfw/node_modules/fs-extra.

### Possible Drawbacks

None.

### Applicable Issues

None.

/cc @as-cii just wanted to make sure this is safe to do.


as-cii:
Looks good, thanks!

PR #15954, opened 2017-10-20
Only update overlay instead of text editor when resize occurs

leroix:
We took care of an unnecessary loop that was occurring when an overlay was present in https://github.com/atom/atom/pull/15894. However, resizes of the overlay were still taking longer than needed because they were invoking `.updateSync` on the text editor component. In this PR, we take the approach of only updating the dimensions of the overlay element and invoking a render on just the overlay component when a resize occurs. The amount of time spent in the overlay observer callback is greatly reduced which means less time spent scripting on keystrokes. This is important because any scripting that occurs as a result of a new keystroke needs to finish well ahead of the next keystroke to avoid holding up its handling.

overlay_resizing

In the above image, the time spent in the resize observer callback is reduced from 2.36ms to 0.32ms, and the time to handle this frame goes from 11.75ms down to 9.10ms.

/cc @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1896,8 +1896,7 @@ describe('TextEditorComponent', () => {
       const decoration = editor.decorateMarker(marker, {type: 'overlay', item: overlayElement, class: 'a'})
       await component.getNextUpdatePromise()
 
-      let overlayComponent
-      component.overlayComponents.forEach(c => overlayComponent = c)
+      const overlayComponent = component.overlayComponents.values().next().value
 
       const overlayWrapper = overlayElement.parentElement
       expect(overlayWrapper.classList.contains('a')).toBe(true)
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1896,6 +1896,9 @@ describe('TextEditorComponent', () => {
       const decoration = editor.decorateMarker(marker, {type: 'overlay', item: overlayElement, class: 'a'})
       await component.getNextUpdatePromise()
 
+      let overlayComponent
+      component.overlayComponents.forEach(c => overlayComponent = c)
+
       const overlayWrapper = overlayElement.parentElement
       expect(overlayWrapper.classList.contains('a')).toBe(true)
       expect(overlayWrapper.getBoundingClientRect().top).toBe(clientTopForLine(component, 5))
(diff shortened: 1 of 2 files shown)


spec/text-editor-component-spec.js
```diff
@@ -1896,6 +1896,9 @@ describe('TextEditorComponent', () => {
       const decoration = editor.decorateMarker(marker, {type: 'overlay', item: overlayElement, class: 'a'})
       await component.getNextUpdatePromise()
 
+      let overlayComponent
+      component.overlayComponents.forEach(c => overlayComponent = c)
```
Ingramz: It looks like you are trying to assign the last element of array to the `overlayComponent ` variable. Perhaps it could be grabbed from the array directly?

src/text-editor-component.js
```diff
@@ -1339,42 +1347,47 @@ class TextEditorComponent {
...
+  }
+
   updateOverlaysToRender () {
     const overlayCount = this.decorationsToRender.overlays.length
     if (overlayCount === 0) return null
 
-    const windowInnerHeight = this.getWindowInnerHeight()
-    const windowInnerWidth = this.getWindowInnerWidth()
-    const contentClientRect = this.refs.content.getBoundingClientRect()
     for (let i = 0; i < overlayCount; i++) {
```
Ingramz: Possible to replace with `this.decorationsToRender.overlays.forEach(this.updateOverlayToRender)` ?

spec/text-editor-component-spec.js
leroix: Yea, sorry for the odd syntax. `overlayComponents` is a `Set`. In `Set`, items don't have positions, but you can iterate over all the items in the `Set`. In the case of this test, I know there's only going to be one item.

src/text-editor-component.js
leroix: That's a good point and switching to `.forEach` may have a minimal impact on performance, but I haven't measured it.

spec/text-editor-component-spec.js
maxbrunsfeld: Very minor, but this might be slightly clearer:

```js
const overlayComponent = component.overlayComponents.values().next().value
```

maxbrunsfeld:
:zap: Awesome improvement!

spec/text-editor-component-spec.js
leroix: Ah, yea. That's better. I hadn't looked at the `Iterator` api. Will Change.

nathansobo: Why do you know there is only one item?

src/text-editor-component.js
```diff
@@ -804,7 +804,15 @@ class TextEditorComponent {
           key: overlayProps.element,
           overlayComponents: this.overlayComponents,
           measuredDimensions: this.overlayDimensionsByElement.get(overlayProps.element),
-          didResize: () => { this.updateSync() }
+          didResize: (overlayComponent) => {
```
nathansobo: Good idea passing in the component reference in this callback. This is such a twisty data relationship but I think this is a reasonable way to navigate it.

nathansobo (commented):
Looking forward to this improvement.

nathansobo (approved):

PR #15975, opened 2017-10-23
Add custom start tiles

richiksc:
### Requirements

### Description of the Change
Add Windows 10-style Atom start tiles, which look like this:
Large Windows 10 start tile
Small Windows 10 start tile

### Alternate Designs

The monogram icon on a solid color background follows other Windows 10 apps the best.

To bring a pleasant experience to all Windows 10 users using Atom pinned to start.

### Benefits

Atom users on Windows 10 will get proper start tiles.

### Possible Drawbacks

People will lose their old default Atom start tile.

### Applicable Issues

May not go into directory where `atom.exe` is stored, because I do not understand Atom's build system.


richiksc:
I need to include Hi-DPI assets as well.

praveeno (changes requested):
should add HDPI image for big screen

50Wliu:
atom-start-tile
For comparison, here's the current pinned Atom tile.

simurai:
Looks great! ✨ 

Would it be possible to have a different background color for stable, beta, dev version? So you could still tell which is which? Probably needs 3 `visualElementsManifest.xml` files and then include the right one?

richiksc:
@simurai Right. What colors are you thinking?

richiksc:
I've added the HiDPI assets and a list view tile.

simurai:
> What colors are you thinking?

There probably will be a 4th "nightly" channel that is purple: https://github.com/atom/atom/pull/17538#issuecomment-399025690. So how about this:

- `#5FB57D` Stable
- `#E0B447` Beta
- `#9142C5` Nightly
- `#74B7E3` Dev

screen shot 2018-06-28 at 1 57 56 pm

Note: I somewhat "randomly" color-picked it from the gradient. :grimacing:

richiksc:
@simurai Do you know how I would have separate `atom.visualElementsManifest.xml` files for each channel? I wouldn't be able to modify the background color without it.

simurai:
> Do you know how I would have separate `atom.visualElementsManifest.xml` files for each channel? I wouldn't be able to modify the background color without it.

No idea. 🤔 If not possible, maybe just have green for all channels?

Yvonne1952-1959 (approved):
resources/win/atom.visualElementsManifest.xml

icecream17:
That spam approval was helpful because I got notified for it. Interesting.

This reminds me of https://github.com/atom/atom/pull/22123 - although I also don't know enough about atom filestructure to know where the put those files.

kutluhanazafli (approved):

kutluhanazafli (approved):

Issue #5046, opened 2015-01-14
Paste without formatting/indentation


dmnd:
In #1407 @nathansobo wrote:

> I wonder if we should offer a non-normalized variant of paste, like alt-cmd-v.

Opening this issue to ask for the above.

Rationale: I leave auto indentation on, and it (mostly) does the right thing, but sometimes I am moving code around that has indentation rules more complicated than Atom can understand. I'd love to have a quick way to paste the "raw" text without having Atom try to indent it. Kind of like "paste without formatting" in a spreadsheet.

An example of complicated indentation might be a docstring, or PEP8-style indenting to opening parens in Python. Ideally Atom would understand all these rules and do the right thing, but right now I often find myself re-indenting things after Atom strips them during paste.

50Wliu:
:+1:.  On Windows and Linux the keybinding would be <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>.

mnquintana:
:+1: This is pretty much what I had configured in Sublime (`Cmd+Shift+V` for Macs of course :smiley:)

bronson:
Totally agree.  Cmd-V should paste-and-reindent, Cmd-Shift-V paste verbatim.

I would like it if the editor pasted, and THEN indented, in two separate steps.  If I didn't like the reindenting, I could hit Undo once to get back the original text.  And if I didn't want to paste at all, I could hit Undo twice.

I think.

This seems discoverable and doesn't require settings/modes...  Any downsides?

imbenclifford:
+1 I'd love this feature too. Would be a game changer for atom.

jenius:
Would also really enjoy this feature!

felixkempf:
This!

jonathanphz:
Yeah I've been looking for this functionality too.

Misterdjack:
Did we just become best friends? Yup! Wanna practice karate in the basement? Yup!  Wanna paste with formatting? Yup!

klarkc:
Nice feature, having to switch off auto indent everytime I need plain paste is a pain

jackson-sandland:
Can we get this feature? This is a time saver, and improves our efficiency while keeping us focused on more important lines of thought!

Ben3eeE:
Closed via https://github.com/atom/atom/pull/15976

PR #15976, opened 2017-10-23
Add command to paste without reformatting

nathansobo:
This PR adds a command to paste without performing any reformatting, regardless of the source of the copied text or the settings for the current file. The following transformations are suppressed:

* **Line ending normalization**. All instances of `\r\n` and `\n` in the pasted text will be preserved regardless of the line endings in the file.
* **Auto-indent** There is a global "auto-indent on paste" setting that can also be assigned per-language, but this command is a way to momentarily bypass that setting.
* **Relative indent level adjustment** By default, we attempt to keep the relative indent level of all lines consistent with the copied text. Sometimes people don't want that.

screen shot 2017-10-23 at 5 16 51 pm

Relevant to recent discussion in #1407

### DIFF
--- a/keymaps/darwin.cson
+++ b/keymaps/darwin.cson
@@ -132,6 +132,7 @@
   'ctrl-shift-w': 'editor:select-word'
   'cmd-ctrl-left': 'editor:move-selection-left'
   'cmd-ctrl-right': 'editor:move-selection-right'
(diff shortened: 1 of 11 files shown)


50Wliu:
Fixes #5046

maxbrunsfeld (approved):

Issue #15565, opened 2017-09-06
Select all automatically unfolds any folded content


MikeWillis:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When performing a select-all, any folded content is automatically unfolded.

### Steps to Reproduce

1. Fold some content
2. Select all

**Expected behavior:**

Folded content remains folded (but is available as part of any cut/copy/paste operations)

**Actual behavior:**

Folded content unfolds

**Reproduces how often:**

100%

### Versions

Atom    : 1.19.6
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
OS: Windows 10 64 bit

### Additional Information

Note that when you highlight a folded section by dragging the mouse over it, it remains folded, and its contents are available for cut/copy/paste operations.

rsese:
Thanks for the report!  I can reproduce on 1.19.6 on macOS 10.12.6 but just noting that this isn't a regression from 1.18.0 (h/t to @Ben3eeE for checking) and likely even some amount of time before that.  This behavior was referenced back in https://github.com/atom/atom/issues/2953#issuecomment-280672731 but it doesn't look like an issue was opened for this behavior.

I'll tag it as a bug for now and ask the team to take a look.

PR #15993, opened 2017-10-25
Preserve folds that are fully contained by the selection when changing selection ranges

nathansobo:
Fixes https://github.com/atom/atom/issues/15565
Depends on https://github.com/atom/text-buffer/pull/272

Previously, whenever we changed the range of a selection, we destroyed all *intersecting* folds. This caused us to destroy even folds that the selection completely contained, which was a counter-intuitive behavior.

This PR upgrades to a new version of `text-buffer` containing a new `destroyFoldsContainingBufferPositions` method that takes an array of positions and destroys any folds containing one of these positions (exclusive of its endpoints). I also use the new method on some other code paths that only need to destroy folds intersecting individual points rather than ranges.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "service-hub": "^0.7.4",
     "sinon": "1.17.4",
     "temp": "^0.8.3",
-    "text-buffer": "13.6.0-0",
+    "text-buffer": "13.6.0-2",
     "typescript-simple": "1.0.0",
     "underscore-plus": "^1.6.6",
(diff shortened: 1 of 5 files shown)


maxbrunsfeld (approved):

nathansobo:
I'm going to go ahead and merge this because I just bumped from the pre-release of atom/text-buffer#272 to an identical official release.

ungb:
Looks good!  🚢 ... oh wait the ship as sail already!  ✨

PR #15996, opened 2017-10-25
Config file concurrency and async

### LINKED ISSUE #14909: Settings lost

damieng:
A more thorough attempt to fix #14909.

Decisions made:

1. Split file/io out from Config class in order to theoretically make testing simpler and allow in the future one of the processes to have changes marshaled across without going to file system. Unfortunately existing tests are more integration tests than unit tests so may need to merge classes back together.

2. Lockfile dependency. Linux doesn't have exclusive locking so we have to use a lock file. This seemed the most popular library.

3. Capture changes and replay them. Borrowed from existing Config class pendingoperations but changed it from function call capture to text so that they could be replayed over process boundaries in the future.

4. Load before save. If the file was locked it's because another process was saving. If we don't reload there's a chance changes will be lost depending on whether we got to handling the file change notification or not yet.

5. Manual debounce & retry. Removed existing debounce code because I need to re-fire the event if an error occurs during save (such as not getting the lock file or some other IO error).

6. Lock file recovery. After 8 attempts if the lock file is still there it might be left over from a crash or something. Rather than silently fail we present an error message with both a 'Retry' button and a 'Force' button - the latter will forcefully remove the lock file so we can continue.

7. Async io. We're trying to move away from sync blocking operations in general and now with extra logic in here it made sense to go async with callbacks.

Issues remaining:

1. What to do when file is locked and we're trying to save on exit?
2. What to do when load fails for any reason and it was requested by file change notification?
3. Current specs for config.cson are not unit tests but integration tests - e.g. setting a value and making sure it saves to disk.  Looks like a big effort to split them out. Alternative is to merge configstorage back into config class.


iolsen:
👍 I'll look again when the automated tests are sorted but, big picture, this looks sane to me. This hole went way deeper than expected. Nice job exploring its depths!

> What to do when file is locked and we're trying to save on exit?

Under what circumstances do we save settings on exit? It seems like our approach here should change depending on how common we expect this to be. If it's common we probably need to do the hard work of deferring exit (like with language server shutdown). If it's strictly a sad path, edge case scenario I think we can get away with doing the same thing you're doing for lock file recovery.

>What to do when load fails for any reason and it was requested by file change notification?

So this should be rare and we just need to not grossly misbehave. Something like:
1. Keep the current in-memory settings
2. Disable saving settings until we're back to a known-good state
3. Show a notification briefly explaining that we've entered this state

For testing with @ungb, I wonder if we can whip up a simple package that does nothing but read/write a bunch of config settings on startup. Get into a state with a bunch of open windows and simply exit/restart it repeatedly. That might enable us to repro the original issue but even if not it'd be a good way to exercise the new locking. Just an idea.

damieng:
Sorry the "trying to save on exit" is more about "we have unsaved config changes and now it's time to exit" (they haven't been written to disk yet because they were either debounced or are retrying waiting on the lock)

src/atom-environment.js
```diff
@@ -182,7 +180,18 @@ class AtomEnvironment {
...
+        config: this.config,
+        configDirPath: this.configDirPath,
+        resourcePath,
+        notificationManager: this.notifications
+      })
+    }
+    this.config.initialize({
+      configFilePath: ConfigStorage.createConfigFilePath(this.configDirPath),
+      projectHomeSchema: ConfigSchema.projectHome
+    })
```
maxbrunsfeld: @damieng I merged `master` into this branch so you wouldn't have to deal with the conflicts I created in converting `atom-environment.coffee` to JS. Luckily your changes to `atom-environment.coffee` were pretty small so it was easy to bring them over to `atom-environment.js`. You might want to give these a 👀  though, just in case I missed something.

damieng:
Unfortunately this unraveled really quickly.   A smaller more opportunistic fix might be to only have one of the two processes write to the config and have the other one marshal the changes across.  That should solve the problem in the scenario where only one copy of atom was open at least.

PR #16016, opened 2017-10-27
Remove hyphen from coffee-script

Ingramz:
CoffeeScript is switching from `coffee-script` to `coffeescript` as the primary module for various reasons.

Refs https://github.com/atom/atom/pull/13731#issuecomment-340002601

cc @nathansobo


50Wliu:
I suggest we stick with `coffee-script` to avoid having to whitelist coffeescript twice in the snapshot module code.

Ingramz:
Yeah... this is dumb.

PR #16025, opened 2017-10-30
☠☕ Decaffeinate `src/theme-manager.coffee`

jasonrudolph:
This change converts `src/theme-manager.coffee` to JavaScript with [standard style](https://standardjs.com).


src/theme-manager.js
```diff
@@ -0,0 +1,401 @@
...
+      this.onDidChangeActiveThemes(() => this.packageManager.reloadActivePackageStyleSheets())
+    })
+  }
+
+  initialize ({resourcePath, configDirPath, safeMode, devMode}) {
+    this.resourcePath = resourcePath
+    this.configDirPath = configDirPath
+    this.safeMode = safeMode
+    this.lessSourcesByRelativeFilePath = null
+    if (devMode || (typeof snapshotAuxiliaryData === 'undefined')) {
```
jasonrudolph: @as-cii: The linter was complaining about `snapshotAuxiliaryData` not being defined. c06745f098a2c462c3bacc569b4aa53c868ed62b tells the linter that `snapshotAuxiliaryData` is a global variable, and that suppresses the linter error. Given your work in this area of the code in 4c6805bf7647efdf2d5d607e0b6dea1f98e9ed73, does this seem like a reasonable approach to dealing with the linter error?

as-cii: Yep, that looks good! Another approach would be to move that global into `package.json`, where I believe we define other globals not recognized by the linter.

jasonrudolph: > Another approach would be to move that global into package.json, where I believe we define other globals not recognized by the linter.

Thanks, @as-cii. Based on the search below, it looks like `snapshotAuxiliaryData` is only referenced as a global in this one file, so I'll whitelist it just in this file. Thanks for the help! :bow:

```
$ ag snapshotAuxiliaryData
script/config.js
40:  snapshotAuxiliaryData: {}

script/lib/prebuild-less-cache.js
32:  CONFIG.snapshotAuxiliaryData.lessSourcesByRelativeFilePath = {}
35:    if (!CONFIG.snapshotAuxiliaryData.lessSourcesByRelativeFilePath.hasOwnProperty(relativeFilePath)) {
36:      CONFIG.snapshotAuxiliaryData.lessSourcesByRelativeFilePath[relativeFilePath] = {
43:  CONFIG.snapshotAuxiliaryData.importedFilePathsByRelativeImportPath = {}
65:        if (!CONFIG.snapshotAuxiliaryData.importedFilePathsByRelativeImportPath.hasOwnProperty(relativeImportPath)) {
66:          CONFIG.snapshotAuxiliaryData.importedFilePathsByRelativeImportPath[relativeImportPath] = []
68:            CONFIG.snapshotAuxiliaryData.importedFilePathsByRelativeImportPath[relativeImportPath].push(

script/lib/generate-startup-snapshot.js
17:    auxiliaryData: CONFIG.snapshotAuxiliaryData,

src/theme-manager.js
1:/* global snapshotAuxiliaryData */
36:    if (devMode || (typeof snapshotAuxiliaryData === 'undefined')) {
40:      this.lessSourcesByRelativeFilePath = snapshotAuxiliaryData.lessSourcesByRelativeFilePath
41:      this.importedFilePathsByRelativeImportPath = snapshotAuxiliaryData.importedFilePathsByRelativeImportPath
```

PR #16041, opened 2017-10-31
Update Dockerfile

bs34:
### Requirements


50Wliu:
Thanks for your submission!

As the name of the file is named `Dockerfile` and Dockerfiles are used only for building Docker images, we feel that it is not necessary to specify the same thing in the comment.

PR #16064, opened 2017-11-01
Exclude Shift from keydown events that terminate selection drags

### LINKED ISSUE #16063: Selection terminates when shift-scrolling
Related to #16059.  I can no longer continue my selection after starting to shift-scroll (on Windows, that means horizontally scrolling).

nathansobo:
Fixes #16063

Follows up on #16060 by including `Shift` in the modifiers that are exempt from terminating a drag operation.

/cc @50Wliu @as-cii

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -4459,6 +4459,7 @@ describe('TextEditorComponent', () => {
       expect(dragging).toBe(true)
       component.didKeydown({key: 'Control'})
       component.didKeydown({key: 'Alt'})
+      component.didKeydown({key: 'Shift'})
       component.didKeydown({key: 'Meta'})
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks!

Ben3eeE:
I compared Atom to ST3 selection cancel keys:
Atom doesn't cancel on: ctrl, shift (soon), alt
ST3 doesn't cancel on: ctrl, shift, alt, caps lock, num lock

I can't figure out any specific reason to cancel or not on the lock keys.
I can't test meta or scroll lock because those keys don't work on my keyboard.

nathansobo:
> I can't figure out any specific reason to cancel or not on the lock keys.

I would guess that they're doing some kind of check for non-printing keys.

PR #16065, opened 2017-11-01
Actually require @atom/nsfw dependency in path-watcher.js

### LINKED ISSUE #15966: Massive memory leak when opening projects with symlink loops
### Prerequisites

### Description

Hi!
First of I know this sounds silly, but I experience this problem only when opening one of my projects, so bear with me.
When opening a ~checkout of my project (https://github.com/thetechnick/nginx-ingress)~ project folder in the size range of >150MB atoms memory consumption is rising constantly at about ~30 MiB per second until it reaches around 14GB and freezes my system.

First I suspected the golang plugin `go-plus` and `ide-go`, but now I have disabled all community packages and the `atom-fuzzyfinder` and the issue still occurs. (same for `--safe` and `--clear-window-state`

I have done a short CPU Profile in Chrome, but it shows nothing I can identify as a Problem:
[CPU-20171022T113255.zip](https://github.com/atom/atom/files/1404767/CPU-20171022T113255.zip)

### Steps to Reproduce

1. Open the project
2. Look at the `top`/`the gnome taks thingy` or whatever shows the amount of memory atom uses

**Actual behavior:** Atom allocates ~30 MiB of memory per second until it crashes the system

**Reproduces how often:** 15~20 times before opening this issue

### Versions

OS: Fedora 26

`atom --version`
Atom    : 1.21.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0

`apm --version`
apm  1.18.5
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.6

### Additional Information
- all community packages are disabled
- can reproduce this issue when opening the content of this ZIP in atom:
https://drive.google.com/file/d/0ByW_Oz6LLJGWVVAwbUdoS3pwS00/view?usp=sharing

nathansobo:
Fixes #15966

In #16017, I upgraded the `nsfw` dependency to `@atom/nsfw` (a temporary fork), but unfortunately I forgot to change the `require` call to `@atom/nsfw` in `path-watcher.js` 🤦‍♂️ . This means the fix didn't actually go out and we must have been requiring some other version of `nsfw` somehow. Not sure which one 😬 .

### DIFF
--- a/src/path-watcher.js
+++ b/src/path-watcher.js
@@ -4,7 +4,7 @@ const fs = require('fs')
 const path = require('path')
 
 const {Emitter, Disposable, CompositeDisposable} = require('event-kit')
-const nsfw = require('nsfw')
+const nsfw = require('@atom/nsfw')
 const {NativeWatcherRegistry} = require('./native-watcher-registry')
 
 // Private: Associate native watcher action flags with descriptive String equivalents.


Ben3eeE:
I can no longer reproduce the issue using the same steps I did before from https://github.com/atom/atom/issues/15966#issuecomment-341226555

nathansobo:
Thanks for the confirmation @Ben3eeE.

alxnorden:
I have encountered memory hogging and CPU overload in the following scenario when trying out atom. 
Fresh atom install on ubuntu, edit some text (lisp), menu->save , save to home directory (as it comes up as default directory for saving). On atom restart, system and editor become slow to respond and top shows 30 to 100 percent atom cpu load. Could this be caused by all the symlinks in the home dir? Is this an issue worth reporting? If reproducible, forcing not saving to home dir might be good...

safe mode did not resolve. reproducible with both ubuntu snap atom install and atom-beta via atom repository . 
Ubuntu 19.10
atom-beta installed via apt as per instructions on site.

`Atom    : 1.42.0-beta0
Electron: 4.2.7
Chrome  : 69.0.3497.128
Node    : 10.11.0`

PR #16074, opened 2017-11-02
Make notifications.clear public and emit event

UziTech:
### Description of the Change

make `atom.notifications.clear()` public and emit a `did-clear-notifications` event

NotificationsManager would need to emit the event

### Benefits

Could let notifications packages know when to clear the notifications log

### Possible Drawbacks

none

### Applicable Issues

https://github.com/atom/notifications/pull/164#issuecomment-341341995

### DIFF
--- a/spec/notification-manager-spec.js
+++ b/spec/notification-manager-spec.js
@@ -68,7 +68,7 @@ describe('NotificationManager', () => {
   })
 
   describe('clearing notifications', function () {
-    it('clears the notifications when ::clear has been called', function(){
+    it('clears the notifications when ::clear has been called', () => {
       manager.addSuccess('success')
       expect(manager.getNotifications().length).toBe(1)
       manager.clear()
@@ -76,16 +76,16 @@ describe('NotificationManager', () => {
     })
 
     describe('adding events', () => {
-      let addSpy
+      let clearSpy
 
       beforeEach(() => {
-        addSpy = jasmine.createSpy()
-        manager.onDidClearNotifications(addSpy)
+        clearSpy = jasmine.createSpy()
+        manager.onDidClearNotifications(clearSpy)
       })
 
       it('emits an event when the notifications have been cleared', () => {
         manager.clear()
-        expect(addSpy).toHaveBeenCalled()
+        expect(clearSpy).toHaveBeenCalled()
       })
     })
   })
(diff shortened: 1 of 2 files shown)


50Wliu:
An event should be added as well.

UziTech:
good call

@50Wliu What do you think about a `will-clear-notifications` event as well. I'm on the fence about it.

50Wliu:
Can you think of any situation where you would need to perform an action that depends on knowing that the notifications are _about_ to be cleared?  If not, I say we skip it.  It can always be added later if a use case does appear.

UziTech:
True, I'm not coming up with any really good reason to have `will-clear-notifications` right now.

UziTech:
I think appveyor is failing because of https://github.com/bruce/node-temp/pull/69

This should be ready for review

maxbrunsfeld:
Triggered a rebuild on Appveyor. Sorry for the delay.

UziTech:
Once this is merged I will be able to create a pull request for a button to clear the new notifications log https://github.com/atom/notifications/pull/164#issuecomment-341341995

 👍

spec/notification-manager-spec.js
```diff
@@ -66,4 +66,27 @@ describe('NotificationManager', () => {
       expect(notification.getType()).toBe('success')
     })
   })
+
+  describe('clearing notifications', function () {
```
maxbrunsfeld: Could you use an arrow function here? That's the style we're using in the rest of the file.

spec/notification-manager-spec.js
```diff
@@ -66,4 +66,27 @@ describe('NotificationManager', () => {
...
+    it('clears the notifications when ::clear has been called', function(){
+      manager.addSuccess('success')
+      expect(manager.getNotifications().length).toBe(1)
+      manager.clear()
+      expect(manager.getNotifications().length).toBe(0)
+    })
+
+    describe('adding events', () => {
+      let addSpy
```
maxbrunsfeld: The name `addSpy` doesn't make sense in this context. It looks like that name is used in the previous `describe` because it is an event listener that is registered using `onDidAddNotification`. I think `clearSpy` would make more sense here.

maxbrunsfeld (changes requested):

maxbrunsfeld (approved):

PR #16075, opened 2017-11-02
Fix select-word command between word and non-word chararacters

nathansobo:
In #15776, we accidentally stopped passing an option to the `Cursor.wordRegExp` method that caused us to prefer word characters when selecting words at a boundary between word and non-word characters.

This is only on master so glad we caught it before the next release.

/cc @maxbrunsfeld

### DIFF
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -2007,13 +2007,17 @@ describe('TextEditor', () => {
 
       describe('when the cursor is between two words', () => {
         it('selects the word the cursor is on', () => {
-          editor.setCursorScreenPosition([0, 4])
+          editor.setCursorBufferPosition([0, 4])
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:bow: Thanks for fixing this!

maxbrunsfeld:
Looks like the appveyor failure is a flake:

```
  1) AtomApplication "after each" hook for "does not reopen any previously opened windows when launched with no path and `core.restorePreviousWindowsOnStart` is no":
     Error: timeout of 60000ms exceeded. Ensure the done() callback is being called in this test.
      at Timeout.<anonymous> (C:\projects\atom\node_modules\mocha\lib\runnable.js:226:19)
```

Filed as https://github.com/atom/atom/issues/16079

PR #16080, opened 2017-11-02
Allow directory providers to implement a custom onDidChangeFiles

hansonw:
### Description of the Change

Atom has supported custom directory providers for quite a long time - the main use case being remote directories, e.g. in Nuclide (https://github.com/facebook/nuclide). The recent addition of file watchers causes several uncaught rejections when a remote directory is added, as it attempts to resolve remote URIs to create a file watcher.

To piggyback on the existing custom directory provider setup, we'll additionally allow custom directories to provide a `onDidChangeFiles` listener (with the same signature as `Project::onDidChangeFiles`) to override the default file watcher.

### Alternate Designs

Another possibility is to suppress file watching for non-local URIs, but this approach allows for greater extensibility as well as real remote file watching if we wanted it. This also opens the interesting option of having the default `Directory` class in `pathwatcher` implement `onDidChangeFiles`, allowing the file watching code to move out of this repo.

There's no other way for a user-land package to override the file watching APIs.

### Benefits

Improved compatibility with custom directory providers (in particular remote directories).

### Possible Drawbacks

It's unclear how to better document this functionality. It seems that custom directory providers are generally underdocumented though?

### DIFF
--- a/spec/project-spec.js
+++ b/spec/project-spec.js
@@ -335,9 +335,14 @@ describe('Project', () => {
       isRoot () { return true }
       existsSync () { return this.path.endsWith('does-exist') }
       contains (filePath) { return filePath.startsWith(this.path) }
+      onDidChangeFiles (callback) {
+        onDidChangeFilesCallback = callback
+        return {dispose: () => {}}
+      }
     }
 
     let serviceDisposable = null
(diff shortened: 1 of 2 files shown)


smashwilson (approved):
:+1: LGTM

PR #16083, opened 2017-11-02
Judge resize of overlay by contentRect changing

### LINKED ISSUE #16077: Repeated overlay measurement in some circumstances
After #15894, @DavidLGoldberg reports that the [Jumpy](https://atom.io/packages/jumpy) package has become unusably slow.

Here's an image he posted where timing violations are being reported measuring overlay decorations containing labels:
image

@leroix I'm going to assign you to this one since you have the most context on the change that caused it, but am happy to help out with it. Since this is on stable we should investigate it 🔜 .

leroix:
fix https://github.com/atom/atom/issues/16077

In the `OverlayComponent` resize observer, we're judging whether the element resized by comparing the `contentRect` with the result of a call to `getBoundingClientRect()` [here](https://github.com/atom/atom/blob/master/src/text-editor-component.js#L4235). The dimensions reported by these two methods aren't always the same[[1]](https://wicg.github.io/ResizeObserver/)[[2]](https://medium.com/dev-channel/after-mutationobserver-performanceobserver-and-intersectionobserver-we-have-another-observer-for-2c541bcb531b). This results in a resize observer loop while the overlay decorations are present:

screen shot 2017-11-02 at 3 17 47 pm

screen shot 2017-11-02 at 3 18 18 pm

In this fix, we store the previous `contentRect` as reported in the `ResizeObserver` callback as state on the `OverlayComponent`. That way in subsequent calls of the `ResizeObserver` callback, we're comparing `contentRect`'s to `contentRect`'s.

@DavidLGoldberg thanks for reporting this bug and the work to help reproduce it on various builds!

/cc @nathansobo

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -126,7 +126,6 @@ class TextEditorComponent {
     this.blockDecorationResizeObserver = new ResizeObserver(this.didResizeBlockDecorations.bind(this))
     this.lineComponentsByScreenLineId = new Map()
     this.overlayComponents = new Set()
-    this.overlayDimensionsByElement = new WeakMap()
     this.shouldRenderDummyScrollbars = true
     this.remeasureScrollbars = false
     this.pendingAutoscroll = null
@@ -803,15 +802,9 @@ class TextEditorComponent {
         {
           key: overlayProps.element,
           overlayComponents: this.overlayComponents,
-          measuredDimensions: this.overlayDimensionsByElement.get(overlayProps.element),
           didResize: (overlayComponent) => {
             this.updateOverlayToRender(overlayProps)
-            overlayComponent.update(Object.assign(
-              {
-                measuredDimensions: this.overlayDimensionsByElement.get(overlayProps.element)
-              },
-              overlayProps
-            ))
+            overlayComponent.update(overlayProps)
           }
         },
         overlayProps
@@ -1357,7 +1350,6 @@ class TextEditorComponent {
     let wrapperTop = contentClientRect.top + this.pixelPositionAfterBlocksForRow(row) + this.getLineHeight()
     let wrapperLeft = contentClientRect.left + this.pixelLeftForRowAndColumn(row, column)
     const clientRect = element.getBoundingClientRect()
-    this.overlayDimensionsByElement.set(element, clientRect)
 
     if (avoidOverflow !== false) {
       const computedStyle = window.getComputedStyle(element)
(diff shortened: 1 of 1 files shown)


nathansobo:
⚡️ Thanks @leroix @DavidLGoldberg.

DavidLGoldberg:
Unfortunately guys, My app is still in a state of disarray.

Oddly it still works with the baby blue dev build OR any other build with `atom -d`.

If it were only working with just the `-d` I'd say maybe I *don't* know how to use git or the dev mode but...I've been doing it for a while now ;)

Here's a bit more description:
https://github.com/DavidLGoldberg/jumpy/issues/108

50Wliu:
@DavidLGoldberg the commit indicates that this fix is only in Atom 1.23.0-beta0 and later...have you tried that version?

DavidLGoldberg:
Oh very sorry I'll check as soon as I get home!  When will that go live?

DavidLGoldberg:
@50Wliu @leroix @nathansobo ,

I tested on the Atom 1.23.0-beta0 and it works smoothly.  Thanks for catching that @50Wliu .

I launched a 4.1.0 with a small little feature / fix and a yellow warning notification explaining the situation.

BTW, I put the notification object on the package's constructor / activation.  I guess this will trigger for any new windows / reloads, but it's pretty easy to dismiss.

Is there a built in way to have this show once and only once to users?  I didn't want to start fumbling with localStorage or any of that, but could it be done?

Would this be a useful feature to add into Atom's core notifications if I can get around to it?

Actually I kind of like the idea of being able to push these things to some managed bulletin but I know there are security concerns / limitations.  It'd be *really* cool if it just read from any GH issue tagged with a certain tag :-p

nathansobo:
> Would this be a useful feature to add into Atom's core notifications if I can get around to it?

Yeah, that's a great idea. Some kind of way of identifying a notification and requesting that it only runs once would be a good feature. I'll be interested to hear your thoughts on the design.

nathansobo:
I thought that #16086 was a fix for 1.22 but it seems like we must have missed something. I think we should follow up rather than leaving this broken for a whole release cycle.

DavidLGoldberg:
Ok, so this might sound unbelievable with my limited Atom expertise (I have a hard time believing it myself) but I believe there is a *much* bigger issue at play.

This may have started since I did a local build of atom (with baby blue icon, I love that by the way).

But I did an install using the downloadable executable from atom.io and now have the good ole stable green icon for both

`atom`
and
`atom -d`

The weird thing is though that everytime I do `atom -d`  Jumpy performance has no issue.  My jumpy repo has a clean git status and it's equivalent to my master branch now (4.1.0).

Atom beta NON dev mode also of course works because it has the fix.

Is this all just a problem with my machine?

I only really have one dev env at the moment unfortunately.

If one of you could do
`apm develop jumpy` and verify that `atom -d` doesn't perform well I'd feel better.  If it *does* well, then there's something scary wrong with dev mode as of late.

Right?  What do I have confused?

(FYI: You have to perform a full jump now to see the 2nd iteration of the performance issue.)

-David

PR #16092, opened 2017-11-03
Autoscroll to cursor position after folding or unfolding

nathansobo:
When folding or unfolding large blocks, it is possible to move the cursor off screen. Whenever this happens it is annoying and confusing, so this PR changes all code paths that create or destroy folds to autoscroll to the cursor position afterwards.

There were several different fold/unfold code paths, so I decided to weave simple autoscroll assertions into the existing tests.


maxbrunsfeld (approved):

nathansobo:
I have no idea why Travis keeps failing it seems unrelated to me.

PR #16100, opened 2017-11-04
Activate package when deserializing

### LINKED ISSUE #16099: Deserializing a package should activate it
### Prerequisites

### Description

I feel like deserializing a package should activate it as well.  For example, I'm experimenting with delaying Timecop activation until the `timecop:show` command is called.  However, I also serialize TimecopView so that if it is open when Atom closes it remains open when Atom restarts.  The problem occurs when TimecopView is deserialized and re-added to the workspace, but Timecop itself is not activated (leading to an odd half-activated state where Timecop is running but has no styling).  This could be avoided by (optionally?) activating the package when the deserializer function is called.

winstliu:
### Requirements

### Description of the Change

This PR unconditionally activates packages when a deserializer for that package is called.  It also adds a new way of deferring package activation through the `workspaceOpeners` key.  Supply it an array of workspace opener filepaths/URIs that your package watches for and it will remain deactivated until one of those openers is called.  This is useful when used in conjunction with deserializers, which often are created using workspace openers.

To demonstrate how a package would benefit from this PR, consider a package that listens to the opener `atom://example` to create a new `ExampleView` that has a serializer.  `atom://example` is listed as a workspace opener in `package.json`.  The `example:open-example-view` command calls `atom.workspace.open('atom://example')` and is listed as an activation command in `package.json`.  This package structure is in use by packages such as Timecop and Markdown Preview.
Previously, there was no way to safely defer activation for a package that implemented serialization because it would not be activated when deserialization occurred.
* Defining the activation command would break deserialization, as the view would get added but the package was technically not activated and as such had no registered styles, commands, menus, etc.
  * With this PR, the deserializer forces the package to be activated.
* Opening the view, closing it, reloading Atom, then reopening the view through methods such as `pane:reopen-closed-item` would throw an `EINVAL: open atom:\example` error on Windows due to the workspace opener not being registered.
  * With this PR, `workspaceOpeners` registers a dummy opener for the opener that then activates the package and calls the real opener.

### Alternate Designs

There could be another key added to deserializers stating whether or not that deserializer should activate the package.

Deserializing happens in core.

### Benefits

Deserializing, openers, and deferred activation techniques can be used together.

### Possible Drawbacks

Perhaps there are some deserializers that don't need the package to be activated?  In which case this will slow down Atom's load time unnecessarily.

### Applicable Issues

Closes #16099


50Wliu:
@BinaryMuse I think this is ready for a review.  Interested in hearing if you have any alternative approaches in mind :).

maxbrunsfeld:
I think this is a great change. The `workspaceOpeners` key is a great idea as well.

I'm wondering if this will break any assumptions that packages would previously have made regarding the timing of the call to `activate`.

Previously, packages' `activate` methods were never called until `atom.workspace.getElement()` was attached to the DOM, so packages might assume that any items they add to the workspace will immediately be on the DOM as well.

/cc @atom/maintainers Thoughts? Is it ok to change the timing in this way? If not, should we just attach the workspace element to the DOM before we do any deserialization?

lee-dohm:
@maxbrunsfeld If it isn't too much trouble, I would prefer we maintain the assumption since many packages immediately start working with the DOM on activation. It could be quite disconcerting to try and figure out why 1 out of 15 times it doesn't work.

50Wliu:
@maxbrunsfeld I've delayed package activation until initial package activation in order to preserve the activation timing.  However if initial packages have already been activated by the time the deserializer is called, activate the package immediately.

rafeca:
Hey! 👋

This is good news: now the Atom code is more consistent and it's much easier to re-format the code to follow the codestyle (now you only need to run `script/lint --fix`).

With ❤️, the Atom team.

nathansobo:
:heart: Thanks @50Wliu. This is an important improvement, especially fixing the holes in the serialization scheme.

Issue #15567, opened 2017-09-06
Non-standard scrolling rate in 1.19.6


billyjanitsch:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Since upgrading from to 1.19.6 on macOS, I've noticed that the scrolling "rate" has slowed down considerably. By "rate", I mean the y-position delta per scroll event. It's hard to measure precisely, but if I compare 1.19.6 to the previous version, or any other native scrolling application (e.g. a browser), the same finger movement on my trackpad results in about 1/3 the movement. Scrolling *isn't* lagging or stuttering; it's just not shifting the viewport by the standard amount (i.e. it takes much longer to scroll through a large text file).

### Steps to Reproduce

1. open a new editor
2. fill it with dummy text
3. scroll around using the trackpad
4. compare to scrolling around in a browser or previous Atom version

**Expected behavior:** the scrolling rate should be standard across all viewports

**Actual behavior:** the scrolling rate is much lower in 1.19.6

**Reproduces how often:** 100%

### Versions

atom 1.19.6
apm 1.18.4
macOS 10.12.6

### Additional Information

I'm using a high-DPI retina display; I don't have a standard display to test on so I'm not sure if this is related to the issue.

billyjanitsch:
A better repro: the tree view (and the settings pane, and basically anything that isn't an editor) retains the correct scrolling behavior in 1.19.6, so you can flip between scrolling in an editor and scrolling the tree view to experience the difference between the two.

rsese:
Thanks for the report - was the fix for https://github.com/atom/atom/issues/15467 mainly about when you scroll so slowly that Atom didn't scroll at all @as-cii?   I feel like I also remember you referencing https://github.com/atom/atom/pull/15310 when we were talking about scrolling issues.  

[...]

Issue #11976, opened 2016-06-14
MouseWheel Scrolling Ignores Operating System Settings


vrgoggles:
### Prerequisites

* [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
* [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [ ] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [ ] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

The number of lines the mouse wheel scrolls does not correspond to the setting in the operating system.

### Steps to Reproduce

1. Open file in Atom
2. Scroll up and down with the mouse wheel

**Expected behavior:** [What you expected to happen]

The number of lines scrolled should be equal to the setting in the operating system.

**Actual behavior:** [What actually happened]

The number of lines scrolled gnores the operating system's setting.

### Versions

1.8

Ben3eeE:
On which OS do you observe this behavior?

50Wliu:
I can confirm this on Windows 10 x64, Atom 1.9.0-dev-3b9b28d.  The default is to scroll three lines but in Atom it's more like 2.5ish lines.

Ben3eeE:
Atom has a Scroll Sensitivity core setting that increases the number of lines scrolled.
I tested this on Windows 7 and can confirm that OS settings in the control panel are ignored.

AntonioRedondo:
[...]

PR #16114, opened 2017-11-06
Use wheelDelta instead of delta in the scroll handler

Ben3eeE:
In Atom 1.19 we changed the scroll handler to use deltaX and deltaY instead of wheelDeltaX/wheelDeltaY. wheelDelta is larger so this caused the scrolling speed to slow down. This change in speed was especially noticeable on Linux.

Fixes https://github.com/atom/atom/issues/15567

# Test plan

* Scrolling on macOS with a mouse
  * [x]  Scroll in every direction
  * [x]  Scroll in every direction while holding shift
* Scrolling on macOS with a touchpad
  * [x]  Scroll in every direction
  * [x]  Scroll in every direction while holding shift
  * [x]  Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * [x]  Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.
* Scrolling on Windows with a mouse
  * [x]  Scroll in every direction
  * [x]  Scroll in every direction while holding shift
* Scrolling on Windows with a touchpad
  * [ ]  Scroll in every direction
  * [ ]  Scroll in every direction while holding shift
  * [ ]  Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * [ ]  Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.
* Scrolling on Linux with a mouse
  * [X] Scroll in every direction
  * [X]  Scroll in every direction while holding shift
* Scrolling on Linux with a touchpad
  * [x] Scroll in every direction
  * [x]  Scroll in every direction while holding shift
  * [x]  Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * [x]  Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.

/cc: @nathansobo @as-cii

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -25,7 +25,7 @@ document.registerElement('text-editor-component-test-element', {
   })
 })
 
-fdescribe('TextEditorComponent', () => {
+describe('TextEditorComponent', () => {
   beforeEach(() => {
     jasmine.useRealClock()
 
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -25,7 +25,7 @@ document.registerElement('text-editor-component-test-element', {
   })
 })
 
-describe('TextEditorComponent', () => {
+fdescribe('TextEditorComponent', () => {
   beforeEach(() => {
     jasmine.useRealClock()
 
@@ -1325,7 +1325,7 @@ describe('TextEditorComponent', () => {
       {
         const expectedScrollTop = 20 * (scrollSensitivity / 100)
         const expectedScrollLeft = component.getScrollLeft()
-        component.didMouseWheel({deltaX: 5, deltaY: 20})
+        component.didMouseWheel({wheelDeltaX: -5, wheelDeltaY: -20})
         expect(component.getScrollTop()).toBe(expectedScrollTop)
         expect(component.getScrollLeft()).toBe(expectedScrollLeft)
         expect(component.refs.content.style.transform).toBe(`translate(${-expectedScrollLeft}px, ${-expectedScrollTop}px)`)
@@ -1334,7 +1334,7 @@ describe('TextEditorComponent', () => {
       {
         const expectedScrollTop = component.getScrollTop() - (10 * (scrollSensitivity / 100))
         const expectedScrollLeft = component.getScrollLeft()
-        component.didMouseWheel({deltaX: 5, deltaY: -10})
+        component.didMouseWheel({wheelDeltaX: -5, wheelDeltaY: 10})
         expect(component.getScrollTop()).toBe(expectedScrollTop)
         expect(component.getScrollLeft()).toBe(expectedScrollLeft)
         expect(component.refs.content.style.transform).toBe(`translate(${-expectedScrollLeft}px, ${-expectedScrollTop}px)`)
@@ -1343,7 +1343,7 @@ describe('TextEditorComponent', () => {
       {
         const expectedScrollTop = component.getScrollTop()
         const expectedScrollLeft = 20 * (scrollSensitivity / 100)
-        component.didMouseWheel({deltaX: 20, deltaY: -10})
+        component.didMouseWheel({wheelDeltaX: -20, wheelDeltaY: 10})
         expect(component.getScrollTop()).toBe(expectedScrollTop)
         expect(component.getScrollLeft()).toBe(expectedScrollLeft)
         expect(component.refs.content.style.transform).toBe(`translate(${-expectedScrollLeft}px, ${-expectedScrollTop}px)`)
@@ -1352,7 +1352,7 @@ describe('TextEditorComponent', () => {
       {
         const expectedScrollTop = component.getScrollTop()
         const expectedScrollLeft = component.getScrollLeft() - (10 * (scrollSensitivity / 100))
-        component.didMouseWheel({deltaX: -10, deltaY: 8})
+        component.didMouseWheel({wheelDeltaX: 10, wheelDeltaY: -8})
         expect(component.getScrollTop()).toBe(expectedScrollTop)
         expect(component.getScrollLeft()).toBe(expectedScrollLeft)
         expect(component.refs.content.style.transform).toBe(`translate(${-expectedScrollLeft}px, ${-expectedScrollTop}px)`)
@@ -1364,14 +1364,14 @@ describe('TextEditorComponent', () => {
       const {component, editor} = buildComponent({height: 50, width: 50, scrollSensitivity})
 
       {
-        component.didMouseWheel({deltaX: 0, deltaY: 3})
+        component.didMouseWheel({wheelDeltaX: 0, wheelDeltaY: -3})
         expect(component.getScrollTop()).toBe(1)
         expect(component.getScrollLeft()).toBe(0)
         expect(component.refs.content.style.transform).toBe(`translate(0px, -1px)`)
       }
 
       {
-        component.didMouseWheel({deltaX: 4, deltaY: 0})
+        component.didMouseWheel({wheelDeltaX: -4, wheelDeltaY: 0})
         expect(component.getScrollTop()).toBe(1)
         expect(component.getScrollLeft()).toBe(1)
         expect(component.refs.content.style.transform).toBe(`translate(-1px, -1px)`)
@@ -1379,14 +1379,14 @@ describe('TextEditorComponent', () => {
 
       editor.update({scrollSensitivity: 100})
       {
-        component.didMouseWheel({deltaX: 0, deltaY: -0.3})
+        component.didMouseWheel({wheelDeltaX: 0, wheelDeltaY: 0.3})
         expect(component.getScrollTop()).toBe(0)
         expect(component.getScrollLeft()).toBe(1)
         expect(component.refs.content.style.transform).toBe(`translate(-1px, 0px)`)
       }
 
       {
-        component.didMouseWheel({deltaX: -0.1, deltaY: 0})
+        component.didMouseWheel({wheelDeltaX: 0.1, wheelDeltaY: 0})
         expect(component.getScrollTop()).toBe(0)
         expect(component.getScrollLeft()).toBe(0)
         expect(component.refs.content.style.transform).toBe(`translate(0px, 0px)`)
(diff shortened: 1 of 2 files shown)


50Wliu:
Furthermore, why are we using the [deprecated, non-standard `mousewheel` event](https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel)?
 https://github.com/atom/atom/blob/42657780d51c357d693328393efebf922ccd7207/src/text-editor-component.js#L481

The standard event is [`wheel`](https://developer.mozilla.org/en-US/docs/Web/Events/wheel), but that event does not have `wheelDeltaX` or `wheelDeltaY`.  It might, however, report more accurate values for `deltaX` and `deltaY`.

50Wliu:
See also MDN's documentation on what the delta values actually mean.  https://developer.mozilla.org/en-US/docs/Web/Events/wheel#The_delta_values

Ben3eeE:
> Furthermore, why are we using the deprecated, non-standard mousewheel event?

I have also noticed this and I am not sure why we use `mousewheel`. But the two events `wheel` and `mousewheel` currently seem to be identical. I couldn't find any difference between the two.

> See also MDN's documentation on what the delta values actually mean. https://developer.mozilla.org/en-US/docs/Web/Events/wheel#The_delta_values

This is what firefox does but chrome doesn't follow this at all.

50Wliu:
> This is what firefox does but chrome doesn't follow this at all.

Well, that sucks :/.  Is that for `mousewheel` or `wheel` as well?

Ben3eeE:
>  Is that for mousewheel or wheel as well?

Since mousewheel and wheel are identical it's for both. We could probably use  deltaX/deltaY and the standard event but currently scrolling is not really usable on Linux so I propose we change it back to the way it was in Atom 1.18 that worked and then figure it out later.

Searching around I found a bunch of issues on the chromium project about delta values being too small on Linux compared to Windows and all were closed as wontfix. We didn't get issues about scrolling on Linux prior to Atom 1.19 as we do now so I think it's fair to go back to the previous behavior for now.

I know that the tests are currently failing because I didn't change them at all.

as-cii:
Looks good to me if it fixes the scrolling issues on Linux. Have you tested all scenarios revolving around scrolling? e.g.

* Scrolling on macOS
  * Scroll in every direction
  * Scroll in every direction while holding shift
  * Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.
* Scrolling on Windows
  * Scroll in every direction
  * Scroll in every direction while holding shift
  * Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.
* Scrolling on Linux
  * Scroll in every direction
  * Scroll in every direction while holding shift
  * Slightly scroll down/up, ensuring that for small values of `wheelDeltaY * scrollSensitivity` we still honor the scrolling.
  * Slightly scroll left/right, ensuring that for small values of `wheelDeltaX * scrollSensitivity` we still honor the scrolling.

I am also a little unclear still on the difference between `mousewheel` vs. `wheel`, and between `wheelDelta` vs `delta`. @50Wliu @Ben3eeE: can you help me understand how and why they differ?

Nice work as usual, @Ben3eeE. ⚡️

Ben3eeE:
Thanks for the feedback and test plan @as-cii :100: :sparkles: 

I would like to change the test plan so we test scrolling with a touchpad **and** mouse. I've updated your test plan and added it to the body of the PR. I fear that while this improves the experience with scrolling with a mouse on Linux it might make touchpads worse. I hope the current test plan will ensure we don't regress either touchpads or mice and will tell us if we should implement a different solution.

---

> I am also a little unclear still on the difference between mousewheel vs. wheel, and between wheelDelta vs delta. @50Wliu @Ben3eeE: can you help me understand how and why they differ?

In Chrome there is no difference between `mousewheel` and `wheel`. `mousewheel` is a non standard deprecated event and `wheel` is the standard event. But all properties of the events are the same. I think `mousewheel` is just kept around for compatibility and that `wheelDelta` was added to the `wheel` event for the same reason. But this is just me guessing.

The difference between `wheelDelta` and `delta` is more complicated. `wheelDelta` is a non standard property that was around before the DOM3 standard event.

**According to MDN:**

**`delta`:**
Reports delta in one of the available [`deltaModes`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode). To my knowledge Chromium never reports delta using `DOM_DELTA_LINE` but it can report using `DOM_DELTA_PAGE`. If we don't merge this PR and instead implement scrolling using `delta` we should make sure we respect all available modes in the scroll handler.
The problem with this delta is that it is really small on Linux so it doesn't seem like it works well on all platforms. On Windows the difference using the default OS settings is 20 less down from 120 to 100. On Linux the delta was 53 when I tested using a standard Ubuntu 16.04 with gnome. This difference makes scrolling very slow on Linux and the sudden change in speed resulted in issues from users.

**`wheelDelta`:**
* On Windows this delta is based on the delta value of `WM_MOUSEWHEEL`
* On Linux it is always `+/-120` making it "The same as IE and Chrome on Windows"
* On macOS it's complicated and depends on the device used and what scrolling modes it supports

Because of the difference between `delta` and `wheelDelta` I feel that testing with both a mouse and a touchpad is required. I fear that this might make scrolling with a touchpad worse on Linux because the change is always 120?

---

I am not sure how I feel about this solution and I think we could maybe further improve the experience and better respect the OS settings by using the standard `delta` property. Regardless of the solution we pick we should go through the same test plan to ensure that scrolling is fast and feels good on all platforms with both a mouse and a touchpad.

nathansobo:
It could be that using `delta` and interpreting the `deltaMode` would lead to a more correct solution, but if people were happy with things when we were using the old deprecated property I think we should return to using it. There may be more for us to learn here. So I say we do what we need to do to relieve the immediate pain and optionally think harder about using the most standards-compliant solution possible in the future.

Ben3eeE:
@nathansobo Thank you for the feedback ✨ 

I am going to need some help going through this test plan because I don't have the required hardware. I can test Windows and Linux using a mouse but I don't have access to a touchpad or macOS.

We can post in https://github.com/atom/atom/issues/11976 about switching to `delta` to improve the native experience and share our current knowledge about these properties. Maybe someone from the community wants to pick it up.

nathansobo:
Once you confirm on Linux and Windows I can test on macOS.

Ben3eeE:
@nathansobo I have tested Windows and Linux with a mouse and can confirm that scrolling works.

I can also confirm that this goes back to the old behavior of ignoring the scroll sensitivity settings that you can configure in Windows. And that we don't slow down scrolling to one pixel per notch when configured to page scrolling because we don't honor `deltaMode` in the old handler.

nathansobo:
> I can also confirm that this goes back to the old behavior of ignoring the scroll sensitivity settings that you can configure in Windows. And that we don't slow down scrolling to one pixel per notch when configured to page scrolling because we don't honor deltaMode in the old handler.

Not sure if I'm interpreting your meaning clearly, but this sounds like a bad thing?

Ben3eeE:
> Not sure if I'm interpreting your meaning clearly, but this sounds like a bad thing?

I don't know. It is an expected thing because `delta` changes with the OS settings and `wheelDelta` does not. And as I have said if we use `delta` we can do a better job at respecting the OS settings (At least on Windows with a mouse, I have no idea about other platforms/hardware). But our current implementation doesn't respect the OS setting on Windows because `delta` based scrolling was added by mistake.

So while the speed would increase if you increased it in the Windows settings it would increase by an amount that does not correspond to the new setting.

nathansobo:
Okay. I'm a bit confused, but I think we should improve the experience on Linux if it doesn't hurt the experience on Windows and take a deeper look at scrolling. Sounds like we need to interpret the `deltaMode` parameter.

nathansobo:
@rsese Do you think you could run through the test plan stated in the body of this PR some time soon? I'd love to get this merged soon and relieve some pain for Linux users in #15567.

rsese:
Yep I can give this a try today @nathansobo :+1:   I don't have a solid working build of Atom on Windows though so probably won't get to that last Windows + touchpad section (and there's no artifact to download from AppVeyor it looks like?).

One question - can you clarify the behavior I should see for these items?

>   - [ ] Slightly scroll down/up, ensuring that for small values of wheelDeltaY * scrollSensitivity we still honor the scrolling
>   - [ ] Slightly scroll left/right, ensuring that for small values of wheelDeltaX * scrollSensitivity we still honor the scrolling.

Is this just verifying that very slight scrolling results in scrolling at all?

nathansobo:
> Is this just verifying that very slight scrolling results in scrolling at all?

I think so. And that it doesn't scroll by too much maybe? Linus would have the best answer but I think he's out for a bit.

nathansobo:
AppVeyor failure looks like a flake so I kicked it.

rsese:
> I think so. And that it doesn't scroll by too much maybe? Linus would have the best answer but I think he's out for a bit.

:+1:

From this branch, I ran through the macOS section (touchpad + mouse) and the Linux + touchpad section (Ubuntu VM).  To compare, I ran through things from Atom `1.24.0-dev-4f6e8ed5f` first.

On macOS, scrolling in every direction with a mouse works fine and it does feel a tiny bit faster and additionally, you scroll ~2 times as much now.  With the touchpad I saw the same thing - feels a bit faster and you scroll ~2 times as much.  If anyone changed their scroll sensitivity to 80 because of https://github.com/atom/atom/issues/15567, they'll end up scrolling quite a bit and may switch back to the default 40.  Unless I mis-understood, touchpad scrolling while holding shift doesn't change the scrolling behavior (like it does with the mouse to scroll horizontally).  Scrolling extremely slowly did scroll with the touchpad and mouse.

On Linux + touchpad, scrolling in every direction worked fine (had to hold shift to scroll horizontally with the touchpad on Ubuntu).  Scrolling did feel faster and like with macOS, you scroll about ~2 as much.  You don't scroll as much as you do on macOS, but definitely a better experience when running off of master which is pretty sluggish.  Scrolling extremely does scroll eventually though Atom doesn't scroll continuously while I scroll on the touchpad like it does with macOS.

rsese:
Checked off the macOS and the Linux + touchpad section - looks like AppVeyor failed again, I'll see if I can get my Windows VM building later tonight.

rsese:
Just saw AppVeyor passed but I ended up getting Atom building on my Windows vm - went through the Windows + touchpad section and didn't notice a difference between master and this branch.  Not sure if that was expected with the changes in this PR?

nathansobo:
I think it's good that the behavior is the same. It means we just make Linux better.

Issue #16011, opened 2017-10-27
Atom 1.21.1 high CPU usage


stefanb2:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

After upgrading from atom 1.21.0 to 1.21.1 atom starts to use ~300% CPU at start and then falls back to 100% CPU. It is busy looping, i.e. at some point the CPU fan of the laptop starts to spin up.

The problem disappears after downgrading to 1.21.0.

### Steps to Reproduce

1. Upgrade to 1.21.0
2. `atom --safe --clear--window-state`
3. check CPU usage with `top`

**Expected behavior:** [What you expect to happen]

No CPU hogging

**Actual behavior:** [What actually happens]

Initial 300% CPU usage, drops to 100% after a short while.

**Reproduces how often:** [What percentage of the time does it reproduce?]

Always

### Versions

```
$ atom --version
Atom    : 1.21.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
$ apm --version
apm  1.18.5
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.14.2
$ inxi -SMCG
System:    Host: XXXXX Kernel: 4.13.6-300.fc27.x86_64 x86_64
           bits: 64
           Desktop: KDE Plasma 5.10.5
           Distro: Fedora release 27 (Twenty Seven)
Machine:   Device: laptop System: LENOVO product: 20F90058MS v: ThinkPad T460s serial: N/A
           Mobo: LENOVO model: 20F90058MS v: SDK0J40697 WIN serial: N/A
           UEFI: LENOVO v: N1CET60W (1.28 ) date: 09/26/2017
CPU:       Dual core Intel Core i7-6600U (-HT-MCP-) cache: 4096 KB
           clock speeds: max: 3400 MHz 1: 2800 MHz 2: 2800 MHz 3: 2800 MHz
[...]

PR #16124, opened 2017-11-07
Back filesystem watchers with @atom/watcher

smashwilson:
Add a new "experimental" setting for `core.fileSystemWatcher` that backs watchPath with [@atom/watcher](https://github.com/atom/watcher). There should be no changes for API consumers.

> ⚠️ There's a reason this is marked "experimental!" Our intention is to not merge this PR until I'm back from pending paternity leave, then let it percolate its way through master, beta, and stable before we consider flipping the feature flag on for everyone. In the meantime, if you want to see if this fixes a filesystem watching issue that you're having, you'll need to build Atom from this branch. :warning:

#### Remaining Work

These are known issues in @atom/watcher that I'd like to address before this merges. Mostly, these avoid regressing on some edge cases that have been reported with the current watcher.

- [x] Translate Windows short names to long names (atom/watcher#85)
- [x] Handle symlinks (atom/watcher#19, atom/atom#15966)
- [x] Test for filesystem case sensitivity issues (atom/watcher#16)
- [x] Test for unicode paths (atom/watcher#12)
- [x] Manual "just poll everything" setting
- [x] @smashwilson back from paternity leave :baby:
- [x] Complete the testing matrix below :point_down:
- [x] Tag and publish @atom/watcher 1.0.0

#### Features and Limitations

In its current state, the experimental watcher:

* Uses one or two threads for all watcher threads, rather than two threads per watcher.
* Fails gracefully to polling when (a) more than 451 watch roots are created on MacOS; (b) `ReadDirectoryChangesW()` reports that a directory does not support watching on Windows; or (c) the maximum number of inotify watch descriptors is exceeded on Linux.
* Can log a bunch of diagnostic information to track down edge cases.
* Scales to thousands of watch roots containing tens of thousands of files.
* Watch directories recursively or non-recursively.
* Watch individual files without polling.

It cannot currently:

* Follow symlinks.
* Explicitly include or exclude subpaths of a watched directory.
* Receive notifications on certain network shares.
* Report drive mount and unmount events.

#### Test Plan

Before we merge, I'd like to verify that we handle or fail gracefully in a number of trickier situations. In each scenario, if we:

* :red_circle: **crash in native code** or **lock up**, I would consider that a blocker for merging this.
* :warning: throw a **JavaScript stack** or **silently fail to deliver any events**, I'd consider merging anyway, with an issue filed against @atom/watcher.

These are some of the cases I've thought to try:

- [x] Open a project directory on a network drive.
  - [x] MacOS: sshfs
  - [x] MacOS: Samba
  - [x] Windows: native Windows-to-Windows network share
  - [x] Windows: Samba
  - [x] Linux: sshfs
  - [x] Linux: Samba
- [x] Open a project directory from a flash drive.
  - [x] MacOS
  - [x] Windows
  - [x] Linux
- [x] Unmount a filesystem that contains an open project directory.
  - [x] MacOS
  - [x] Windows
  - [x] Linux
- [x] Mount a new filesystem beneath a watched directory.
  - [x] MacOS
  - [x] Linux
- [x] Cloning a large git repository beneath a watched directory.
  - [x] MacOS
  - [x] Windows
  - [x] Linux

#### To Opt In

If you want to give this a try before it merges, clone and [build Atom](https://github.com/atom/atom/tree/master/docs/build-instructions) from this branch:

```sh
git clone https://github.com/atom/atom.git --branch aw-watcher
cd atom
script/build --install
```

Launch Atom and set `core.fileSystemWatcher` to "experimental" in your core settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "electronVersion": "1.7.11",
   "dependencies": {
     "@atom/nsfw": "^1.0.18",
+    "@atom/watcher": "0.0.8",
     "@atom/source-map-support": "^0.3.4",
     "async": "0.2.6",
     "atom-keymap": "8.2.9",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "electronVersion": "1.7.11",
   "dependencies": {
     "@atom/nsfw": "^1.0.18",
-    "@atom/watcher": "0.0.8",
+    "@atom/watcher": "1.0.1",
     "@atom/source-map-support": "^0.3.4",
     "async": "0.2.6",
     "atom-keymap": "8.2.9",
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -338,6 +338,14 @@ const configSchema = {
             description: 'Native operating system APIs'
           },
           {
+            value: 'experimental',
+            description: 'Experimental filesystem watching library'
+          },
+          {
+            value: 'poll',
+            description: 'Polling'
+          },
+          {
             value: 'atom',
             description: 'Emulated with Atom events'
           }
--- a/src/path-watcher.js
+++ b/src/path-watcher.js
@@ -3,6 +3,7 @@ const path = require('path')
 
 const {Emitter, Disposable, CompositeDisposable} = require('event-kit')
 const nsfw = require('@atom/nsfw')
+const watcher = require('@atom/watcher')
 const {NativeWatcherRegistry} = require('./native-watcher-registry')
 
 // Private: Associate native watcher action flags with descriptive String equivalents.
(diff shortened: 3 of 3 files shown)


lee-dohm:
@smashwilson I feel like for network support we should test:

1. Native Windows-to-Windows network shares
1. sshfs
1. Samba

in that priority order. Also, we should test them through a connection with restricted network throughput and latency generator to see what the limits are for our implementation. (In other words, will sshfs over a high-latency 3G cellular connection work as expected?)

50Wliu:
I'll try to run through flash drive and cloning a large git repository on Windows soon.

50Wliu:
Got around to checking the scenarios I mentioned above and everything seems all right 👍.  Decent performance, no crashes or anything.

smashwilson:
> Got around to checking the scenarios I mentioned above and everything seems all right 👍. Decent performance, no crashes or anything.

✨ Excellent, thank you! You're on Windows, right?

smashwilson:
On reflection I'm likely going to punt on atom/watcher#70 until after this is merged. It isn't entirely clear to me what the right thing to do is there, and I don't want to hold this up when a better solution might not involve @atom/watcher at all. What's more, nsfw also doesn't handle this well, so we're no worse off than what's currently shipped.

I would like to tackle atom/watcher#19 though.

@ungb: Do we have a samba share or an sshfs mount already available somewhere that we could use for network drive tests? It might be nice to have those available for general testing in any case 🤔

50Wliu:
> You're on Windows, right?

Yep.

smashwilson:
> on Windows

... I can read, really

ungb:
> @ungb: Do we have a samba share or an sshfs mount already available somewhere that we could use for network drive tests? It might be nice to have those available for general testing in any case 🤔

@smashwilson we do have something for sshfs, let me look into samba share.

smashwilson:
I saw a crash when you disconnect a Samba share on Windows while it contains a watched directory. Tracking it in atom/watcher#118.

image

gothicserpent:
#16011 #15966

I just wanted to chime in by saying I recently added github functionality to my atom version, and the default filesystem watcher was constantly at around 5% when i was working on small tasks.

looking forward to this experimental file system watcher and i hope the cpu util will go down! thanks for the effort.

Issue #16004, opened 2017-10-26
Navigating through a form with atom-text-editor as inputs with tab key


akonwi:
I opened an issue in the GitHub package repo (https://github.com/atom/github/issues/1234) with a gif demonstrating the problem. 

I'm seeing the same behavior when I use `<atom-text-editor>`s as input elements. The focus moves from the input to the next button.

akonwi:
Giving the first input a tabIndex of -1 and working up from there solved the issue for me.

rsese:
Thanks for the report @akonwi!  So it seems like you're reporting a higher level issue where https://github.com/atom/github/issues/1234 is an example of it?

Could you please edit the issue body and fill out the report using the issue template?  That will make it a lot easier to ask the team to take a look.  If you can share as much detail as possible, that would be helpful too (e.g. did this used to work in an older version of Atom, can you share a minimal code reproduction, etc.).

itsmichaelwang:
I think I found the cause. If you trigger the modal with Cmd+Shift+P > GitHub: Clone and you look at the modal that pops up, the text boxes look something like this:

    <atom-text-editor tabindex="-1" mini="" class="editor mini" data-encoding="utf8" data-grammar="text plain null-grammar">
    ...
    </atom-text-editor>

When tabindex = -1, then it makes it so that you can't tab to those input boxes. I made the tabindex = 1 for the first box, and tabindex = 2 for the second, and it looks like that fixed the issue.

I don't know enough about atom to actually make the right PR. I'll look around, but that's my take.

50Wliu:
Right here. https://github.com/atom/atom/blob/4fd0345e8488d2fa09151316b4e99656e7861644/src/text-editor-component.js#L480

itsmichaelwang:
My thought was to take the correct tabIndex (which is present in the props object in the constructor of `text-editor-component.js`) and to write it to a variable in the constructor. And save it for later.

Is there a better way to debug this then making changes and rebuilding? It takes about 5-10 minutes to rebuild for every little change I make, and it makes debugging... less than ideal. If not, that's fine too. I was just wondering if there was a trick I'm missing to improve my workflow.

I saw this: http://flight-manual.atom.io/hacking-atom/sections/contributing-to-official-atom-packages/
but it refers to packages and not the main project itself.

50Wliu:
1. Set `ATOM_DEV_RESOURCE_PATH` to where you cloned Atom.
2. Make changes.
3. `atom --dev`

[...]

PR #16127, opened 2017-11-08
Allow you to tab through modal text box

itsmichaelwang:
### Description of the Change
We have a modal with elements (1) text box, (2) text box, (3) cancel button, and (4) clone button, respectively, from the top-left to the bottom-right.
screen shot 2017-11-08 at 12 43 16 am

For this modal element, the tabIndexes are -1, -1, 3, 4, respectively. This causes the tabbing order to be (1) > (3) > (4) or (2) > (3) > (4), and then looping forever between (3) and (4), because tabIndex=-1 signifies to skip tabbing for that element. It would appear the root cause is the tabIndex for those elements is hardcoded to be -1. We should probably change that.

Instead, allow a tabIndex to be inferred from the constructor. Now the modal has tabIndexes of 1, 2, 3, 4, as it was intended to be, from the atom github repo.

But now there is a new issue. We can only tab from (1) > (2), but never to (3) or (4). I have tried to understand why this is happening, but I can't come to a conclusion after playing around with the values, nor do I understand how to debug tabbing better than the trial and error approach I have now. Seems like something that's probably better for another ticket, because the tabIndexes are correct now.

### Alternate Designs
Maybe instead of inferring tabIndex from `this.props.element.tabIndex` in line 173, we should instead have some kind of explicit field to specify it, like `this.props.tabIndex`. I would not be opposed to that. Of course, then we need to do some more refactoring work to make tabIndex work for all objects that create this modal. It's a tradeoff between time investment and tech debt.

Also, I'm not sure if this would break other things. Seems like a pretty broad change.

The core issue here is that the tabIndex is being overwritten in `text-editor-component.js` from a hardcoded value. Even if another package (in this case, the Github package) specifies a tabIndex, it will be squashed by what's hardcoded, and that seems wrong.

### Benefits
You can tab through your text boxes now.

### Possible Drawbacks
You can't tab through your buttons now. I think you also can't use the escape button to exit out of a modal, but you couldn't do that anyways unless the focus was on the "Cancel" button, so the problems are linked.

I think the tabbing thing is actually a separate issue, and this PR just solves the tabIndex being squashed by a hardcoded value.

### Applicable Issues
#16004

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -170,7 +170,7 @@ class TextEditorComponent {
     this.textDecorationBoundaries = []
     this.pendingScrollTopRow = this.props.initialScrollTopRow
     this.pendingScrollLeftColumn = this.props.initialScrollLeftColumn
-    this.tabIndex = this.props.element && this.props.element.tabIndex ? this.props.element.tabIndex : -1;
+    this.tabIndex = this.props.element && this.props.element.tabIndex ? this.props.element.tabIndex : -1
 
     this.measuredContent = false
     this.queryGuttersToRender()
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -170,6 +170,7 @@ class TextEditorComponent {
     this.textDecorationBoundaries = []
     this.pendingScrollTopRow = this.props.initialScrollTopRow
     this.pendingScrollLeftColumn = this.props.initialScrollLeftColumn
+    this.tabIndex = this.props.element && this.props.element.tabIndex ? this.props.element.tabIndex : -1;
 
     this.measuredContent = false
     this.queryGuttersToRender()
@@ -477,7 +478,7 @@ class TextEditorComponent {
         style,
         attributes,
         dataset,
-        tabIndex: -1,
+        tabIndex: this.tabIndex,
         on: {mousewheel: this.didMouseWheel}
       },
       $.div(
@@ -3586,7 +3587,7 @@ class CursorsAndInputComponent {
         compositionupdate: didCompositionUpdate,
         compositionend: didCompositionEnd
       },
-      tabIndex: -1,
+      tabIndex: this.tabIndex,
       style: {
         position: 'absolute',
         width: '1px',


itsmichaelwang:
I will address the failing tests first.

itsmichaelwang:
@50Wliu unit tests passed

akonwi:
bump

src/text-editor-component.js
```diff
@@ -3586,7 +3587,7 @@ class CursorsAndInputComponent {
         compositionupdate: didCompositionUpdate,
         compositionend: didCompositionEnd
       },
-      tabIndex: -1,
+      tabIndex: this.tabIndex,
```
jasonrudolph: I think this is an attempt to access the [`tabIndex` property on the `TextEditorComponent` object](https://github.com/atom/atom/blob/df2f3c3124a6e7ea7f9e23c912cff9abfefda0b9/src/text-editor-component.js#L173). However, we're inside the [`CursorsAndInputComponent` class](https://github.com/atom/atom/blob/df2f3c3124a6e7ea7f9e23c912cff9abfefda0b9/src/text-editor-component.js#L3494-L3604) here, and the `CursorsAndInputComponent` objects don't currently have a `tabIndex` property. As a result, `this.tabIndex` is always `undefined`, and that results in the `input` element getting its tabIndex set to `0`. I _think_ this is the cause of the issues you noted in the [possible drawbacks section](https://github.com/atom/atom/pull/16127#issue-151324878).

I've attempted to resolve those issues in #16492. If you have a chance to look at that pull request, I'd love to have your feedback. :bow:

jasonrudolph:
I cherry-picked the commits from this pull request into #16492, and I added some additional commits to 1) address the drawbacks noted in the pull request body above and 2) add tests for this new behavior.

Since #16492 is merged now, and since it includes the commits from this pull request, I'm going to close this pull request. These improvements improvements will likely be available in Atom 1.25.

@itsmichaelwang: Thanks again for this contribution!

PR #16156, opened 2017-11-11
Correct grammatical error

fjwldap:
### Requirements

It is just a correcting grammar...

### Description of the Change
In 'atom/dot-atom/keymap.cson', there are more than one example.
There are 'atom-text-editor' and 'atom-workspace'.
But in line 11, it said "Here's _an example_ taken from Atom's built in keymap:".
So I changed it to "_Examples_ taken from Atom's built in keymap:"

To correct grammar.

### Benefits

Just correct grammar.

### DIFF
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -8,7 +8,7 @@
 # You can create a new keybinding in this file by typing "key" and then hitting
 # tab.
 #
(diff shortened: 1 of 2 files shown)


sadick254:
I see no significant value that this adds to the atom project.

PR #16161, opened 2017-11-12
Add .toString() method for Color objects

winstliu:
### Requirements

### Description of the Change

Color had a lot of helpful `to*` methods, but it was missing arguably the most important one: `toString()`.  This fixes the Settings Panel from displaying `[object Object]` as the default value for a color:
color-object-object-default-value

### Alternate Designs

I chose to use `this.toRGBAString()` because the Windows color picker does not have a hex picker.  Alternate designs include creating a more human-readable presentation such as `Red: 255, Green: 255, Blue: 255, Alpha: 1` or using `this.toJSON()`, but then the default will be fairly useless for Windows users.  However, it would appear that Linux's default color picker only has hex input: https://github.com/atom/settings-view/issues/520.  Perhaps both should be used, such as `rgba(255, 255, 255, 1) (#FFFFFF)`?

Color is an essential class and it makes sense for it to support `.toString()`.

### Benefits

More helpful string representation for Settings View.

### Possible Drawbacks

None.

### Applicable Issues

None.

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -88,7 +88,7 @@ export default class Color {
   toJSON () {
     return this.alpha === 1 ? this.toHexString() : this.toRGBAString()
   }
-  
+
   toString () {
     return this.toRGBAString()
   }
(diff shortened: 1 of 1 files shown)


NGsolution:
Update file

iolsen:
👍

PR #16186, opened 2017-11-14
Autoscroll after folding from keyboard, take 2

nathansobo:
This PR reverts a previous attempt in #16092 which was a bit too aggressive. It auto-scrolled at the model level after various folds. This PR tries a more conservative approach that simply auto-scrolls after fold *commands*. Auto-scrolling seems most intuitive when executed from the keyboard via a command.

Thanks @maxbrunsfeld for the early feedback. 💯


maxbrunsfeld (approved):
:zap:

Issue #16167, opened 2017-11-13
Non-SSL/TLS Links in README Building Section Should Use SSL/TLS


U8NWXD:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

The links under the `Building` header in the README point to the `http://` versions of sites even though the sites support SSL/TLS. Many other links in the README point to `https://` sites, and I think these should too.

### Steps to Reproduce

1. Open the README file in the [core](https://github.com/atom/atom) repository
2. Navigate to the [Building](https://github.com/atom/atom/blob/master/README.md#building) section
3. Follow the link my clicking on it or copying it into a browser

**Expected behavior:** The link leads to an SSL/TLS-secured site

**Actual behavior:** The link leads to a site without SSL/TLS

**Reproduces how often:** 100% of the time

### Versions

The current version of the README on GitHub

### Additional Information

Specifically, the links for `Linux`, `macOS`, and `Windows` are affected. I have not really contributed to open source projects before, but I am interested in learning by fixing this issue.

The specific code I am referencing is copied below (copied from `README.md` in `atom/atom` on GitHub; copyrighted by GitHub Inc.):
```markdown
## Building

* [FreeBSD](./docs/build-instructions/freebsd.md)
* [Linux](http://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#platform-linux)
* [macOS](http://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#platform-mac)
* [Windows](http://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#platform-windows)
```

U8NWXD:
Does this change make sense, or is there a reason to keep these links as `http://` ones? If this is a change that makes sense, I'd be happy to submit a PR for it.

50Wliu:
Definitely makes sense.  PRs would be appreciated.

PR #16190, opened 2017-11-15
Change HTTP Links to HTTPS in CONTRIBUTING.md

U8NWXD:
### Description of the Change

Change HTTP links to HTTPS in CONTRIBUTING.md since the linked-to sites support SSL/TLS

### Alternate Designs

The alternative would be leaving the links as HTTP ones. This leaves users less secure while having only marginal user experience benefits.

The documentation changed is in Core.

### Benefits

Makes user interaction more secure by default.

### Possible Drawbacks

Marginal potential for devices not supporting SSL/TLS and marginal performance hits when accessing the links.

### Applicable Issues

Similar to issue #16167 and PR #16173

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ This project and everyone participating in it is governed by the [Atom Code of C
 
 ## I don't want to read this whole thing I just have a question!!!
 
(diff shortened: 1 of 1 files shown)


50Wliu:
Thank you!! :100:

Issue #13617, opened 2017-01-14
Rename, reveal in tree view, and show in finder with split files can affect the wrong split 


feakuru:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When I open 2 files in horizontal split and try to rename (RMB on file tab -> "Rename") the right one, it shows the name of the left one and renames it, not the right one.  Reveal in tree view and show in finder are similarly affected.

### Steps to Reproduce

1. Open two files
2. Drag one of them to the right until editor highlights a right split for it
3. Focus the file on the left.
3. Right-click the file on the right and try to rename it.

**Expected behavior:** a pop-up appears with an input box for renaming the "right" file

**Actual behavior:** a pop-up appears with an input box for renaming the "left" file

**Reproduces how often:** was able to catch it once

### Versions

Xubuntu 16.04
atom v6.5.0
apm  1.12.9
npm  3.10.5
node 4.4.5
python 2.7.12
git 2.7.4

### Additional Information

*none*

feakuru:
update: i seem to have incorrectly understood the nature of the issue. the renaming of the wrong file happens if the wrong tab is active. however, i still believe that the behaviour discovered is faulty and am not closing the issue.

50Wliu:
Thanks @feakuru, confirmed.  I've updated your reproduction steps with the extra information in your update.  Looks somewhat related to #13398.

jeffreyscottfrench:
I'm also experiencing this (Atom 1.18.0 on MacOS 10.12.5) - contextual menu from right-click appears over the click point, not the active tab. Right click should activate the tab underneath the click, or not display options that can't be performed on an inactive tab (as it does when right-clicking elsewhere in the top of an open pane with tabs). Thanks!

rsese:
Updating the issue title and body since the bug includes Reveal in Tree View and Show in Finder 
[...]

PR #16192, opened 2017-11-15
Make ContextMenu async

### LINKED ISSUE #2991: Rightclick in the treeview doesn't focus the selected item directly

winstliu:
### Requirements

### Description of the Change

Makes the context menu async, which is important for perceived performance and UX.  No longer will _everything_ freeze when opening a context menu.  Here's a gif of snowfall!
async-context-menu

### Alternate Designs

None.

Context menu handling occurs in the main process.

### Benefits

Context menus no longer block the renderer process until they are closed.

### Possible Drawbacks

Per https://github.com/electron/electron/issues/9441 it is not possible to detect when the context menu is closed.  Atom currently does not track context menu state so I think this compromise is fine.

### Applicable Issues

Supersedes and closes #13398
Fixes #2991
Refs #13617, though it looks like right-click isn't properly focusing the tab so more investigation will be needed.


50Wliu:
Follow-up issues:
* Right-clicking to open the context menu does not move the cursor
* Right-clicking on a tab does not focus the tab

BinaryMuse:
The concern I had about the right-click hijacking hack in the GitHub package appears to be fine with this change. 👍  to merge once it's green!

PR #16197, opened 2017-11-15
Change HTTP Links to HTTPS in /docs/build-instructions/

U8NWXD:
### Description of the Change
Change HTTP links to HTTPS in /docs/build-instructions/ since the linked-to sites support SSL/TLS
### Alternate Designs
The alternative would be leaving the links as HTTP ones. This leaves users less secure while having only marginal userexperience benefits.
The documentation changed is in Core.
### Benefits
Makes user interaction more secure by default.
### Possible Drawbacks
Marginal potential for devices not supporting SSL/TLS and marginal performance hits when accessing the links.
### Applicable Issues
Similar to issue #16167 and PR #16173


50Wliu:
Great work as usual @U8NWXD.

PR #16199, opened 2017-11-15
Convert Package to JS

maxbrunsfeld:


50Wliu:
@maxbrunsfeld any chance this can wait until #16100 is merged?

src/package.js
```diff
@@ -0,0 +1,1106 @@
...
+  getActivationCommands () {
+    if (this.activationCommands) return this.activationCommands
+
+    this.activationCommands = {}
+
+    if (this.metadata.activationCommands) {
+      for (let selector in this.metadata.activationCommands) {
+        const commands = this.metadata.activationCommands[selector]
+        if (!this.activationCommands[selector]) this.activationCommands[selector] = []
+        if (_.isString(commands)) {
```
YurySolovyov: why not `typeof`?

src/package.js
```diff
@@ -0,0 +1,1106 @@
...
+          this.activationCommands[selector].push(commands)
+        } else if (_.isArray(commands)) {
```
YurySolovyov: Why not `Array.isArray` ?

src/package.js
```diff
@@ -0,0 +1,1106 @@
...
+    }
+
+    const incompatibleNativeModules = []
+    for (let nativeModulePath of this.getNativeModuleDependencyPaths()) {
+      try {
+        require(nativeModulePath)
+      } catch (error) {
+        let version
+        try {
+          ({version} = require(`${nativeModulePath}/package.json`))
```
YurySolovyov: are parens doing anything here?

src/package.js
```diff
@@ -0,0 +1,1106 @@
...
+  disable () {
+    return this.config.pushAtKeyPath('core.disabledPackages', this.name)
+  }
+
+  isTheme () {
+    return this.metadata && this.metadata.theme
+  }
+
+  measure (key, fn) {
+    const startTime = Date.now()
```
YurySolovyov: `performance.now()` could be less expensive and more precise here

src/package.js
50Wliu: Yes, you cannot have a bracket start a line because it interprets it as a block.

src/package.js
```diff
@@ -0,0 +1,1106 @@
...
+      CompileCache.removeTranspilerConfigForPath(this.path)
+    }
+  }
+
+  loadKeymaps () {
+    if (this.bundledPackage && this.packageManager.packagesCache[this.name]) {
+      this.keymaps = []
+      for (const keymapPath in this.packageManager.packagesCache[this.name].keymaps) {
+        const keymapObject = this.packageManager.packagesCache[this.name].keymaps[keymapPath]
+        this.keymaps.push([`core:${keymapPath}`, keymapObject])
```
YurySolovyov: empty array + unconditional push seems like `.map` to me

maxbrunsfeld:
> any chance this can wait until #16100 is merged?

Absolutely. This is low priority. Just something easy to do when I need a break from working on another problem.

src/package.js
maxbrunsfeld: This is a loop over an object, so `for in` is convenient. An alternative would be `Object.keys(...).map`, though that would allocate an intermediate array of keys. Readability-wise, they seem pretty much equivalent.

PR #16208, opened 2017-11-16
Ensure style updates take effect after the editor is re-attached

as-cii:
Refs: https://github.com/atom/teletype/issues/174
Refs: https://github.com/atom/teletype/pull/185

Previously, performing a style change while an editor was detached would cause the editor component to not observe the change. Thus, if the editor was later re-attached, it would use potentially incorrect measurements to position lines, cursors, etc.

With this pull-request we will now keep track of a global counter of how many times the styles have changed. Also, each editor component will have its own local counter tracking how many style changes it has observed. If on re-attach the global counter is different from the local counter, it means that the editor component is using stale measurements and, as such, needs to clear them out and re-measure.

/cc: @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -442,6 +442,18 @@ describe('TextEditorComponent', () => {
         expect(component.getScrollContainerClientHeight()).toBe(100 - 10)
         expect(component.getScrollContainerClientWidth()).toBe(100 - component.getGutterContainerWidth() - 10)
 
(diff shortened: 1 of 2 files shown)


sadick254:
Hey @as-cii . Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254:
Thank you for your PR!

Issue #15909, opened 2017-10-16
Typescript transpilation is "essentially broken" due to change in location of core library definitions


mattdiamond:
From [this comment](https://github.com/jakesankey/line-diff-details/issues/45#issuecomment-336309778):

> ...this has to do with the built-in TypeScript support that Atom has, which utilizes typescript-simple =1.0.0, which in turn depends on typescript ~1.4.1. The location of the core library definitions was moved from bin/ to lib/ somewhere between TypeScript 1.4 and 1.5, which means that typescript-simple is unable to find the core library defs (lib.d.ts) when installed with the most recent version of TypeScript matching ~1.4.1. In other words, Atom's built-in TypeScript support is essentially broken, and when it sees lib/line-diff-details.ts, it tries to transpile it, but the definitions it needs aren't where it expects them, and boom, error.

This issue is breaking plugins that depend on TypeScript transpilation, such as [line-diff-details](https://atom.io/packages/line-diff-details). According to the same user, the fix [should be simple](https://github.com/jakesankey/line-diff-details/issues/45#issuecomment-337011435):

> I think it should just require bumping Atom's dependency on typescript-simple from 1.0.0 (which is now broken) to any version that uses TypeScript 1.5 or later. It should probably not be 2.x, though, since that has the potential to leave some packages already using TS just as broken as they were before.

The exception:

```
At ENOENT, node_modules/typescript/bin/lib.d.ts not found in /Applications/Atom.app/Contents/Resources/app.asar

Error: ENOENT, node_modules/typescript/bin/lib.d.ts not found in /Applications/Atom.app/Contents/Resources/app.asar
    at notFoundError (ELECTRON_ASAR.js:115:19)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:514:9)
    at TypeScriptSimple.createService (/Applications/Atom.app/Contents/Resources/app/node_modules/typescript-simple/index.js:71:73)
    at TypeScriptSimple.compile (/Applications/Atom.app/Contents/Resources/app/node_modules/typescript-simple/index.js:60:43)
    at Object.exports.compile (/Applications/Atom.app/Contents/Resources/app/src/typescript.js:54:53)
    at Object.compile (/Applications/Atom.app/Contents/Resources/app/src/package-transpilation-registry.js:79:33)
    at compileFileAtPath (/Applications/Atom.app/Contents/Resources/app/src/compile-cache.js:97:37)
[...]

PR #16209, opened 2017-11-16
Modernize built-in TypeScript transpiler to support JSX (React) files.

thomasjo:
### Description of the Change
This PR upgrades `typescript-simple` to the latest version to make it possible for package developers to use TypeScript + JSX (React) transpilation.

JSX support was added in TypeScript 1.6, but our version of `typescript-simple` bundles TypeScript 1.4. The proposed change will give us TypeScript 2.6, which is the latest and greatest.

### Alternate Designs
As I see it right now, there are two alternatives approaches;
1. Update `typescript-simple` to a version that bundles TypeScript 1.6, to minimize the risk of breaking changes affecting packages, etc.
2. Find another, more scalable solution for allowing packages with custom transpilers, or Atom Core, to add additional extensions to `require.extensions`.

Note that we could technically speaking also just add the `.tsx` mapping using the current version of TypeScript, but it will never work unless the package actually uses a custom transpiler with a more modern TypeScript. I think this is a bad idea since it would mean shipping something that doesn't work.

Even though we now have custom transpiler support on package-level, this functionality is still heavily reliant upon us monkey patching `require.extensions` in [`compile-cache.js`](https://github.com/atom/atom/blob/6cbc9988ecf935ab3905700a68bf3ce4f854cb3b/src/compile-cache.js#L226-L237). Therefore, if a package developer wants to use JSX in TypeScript, it will not currently work since we are not mapping the `.tsx` extension to any transpiler. It is technically possible to work around this limitation by explicitly specifying the `.tsx` when requiring JSX components from a TypeScript module, but this causes TypeScript to spit out warnings/errors, since it is terrible idea that will not work in general scenarios because `.tsx` files will be transpiled to `.js` or `.jsx` files depending upon the compiler options. This just happens to magically work in Atom due to the compile cache, and what not.

### Benefits

Allows package developers to use TypeScript's JSX support. Another, almost accidental, benefit is that we'll provide modern TypeScript transpilation support out of the box.

### Possible Drawbacks

This might break packages that are currently using the antiquated, built-in TypeScript support due to the remarkable number of [breaking changes](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes) introduced in TypeScript since v1.4.

### Applicable Issues

Quick search did not reveal any obviously applicable issues.

### Additional tasks
- [ ] Find all community packages that use our built-in TypeScript support.
  - [ ] Check if any of these packages are affected by breaking changes.
  - [ ] Submit PRs to affected packages.
- [ ] Celebrate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "sinon": "1.17.4",
     "temp": "^0.8.3",
     "text-buffer": "13.8.6",
-    "typescript-simple": "1.0.0",
+    "typescript-simple": "^8.0.5",
     "underscore-plus": "^1.6.6",
     "winreg": "^1.2.1",
     "yargs": "^3.23.0"
--- a/spec/fixtures/typescript/foo.tsx
+++ b/spec/fixtures/typescript/foo.tsx
@@ -1,0 +1,3 @@
+export default function Foo (props) {
+  return <div>foo</div>
+}
--- a/spec/typescript-spec.coffee
+++ b/spec/typescript-spec.coffee
@@ -1,4 +1,4 @@
-fdescribe "TypeScript transpiler support", ->
+describe "TypeScript transpiler support", ->
   describe "when there is a .ts file", ->
     it "transpiles it using typescript", ->
       transpiled = require('./fixtures/typescript/valid.ts')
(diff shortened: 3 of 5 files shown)


smhxx:
Related: #15909

lee-dohm:
Closing in favor of #17001. Thanks @thomasjo for your help with this :bow:

PR #16229, opened 2017-11-18
Make atom.confirm async

winstliu:
### Requirements

### Description of the Change

This change is a bit more involved than making ContextMenu async.  `dialog.showMessageBox` accepts an optional callback as a third argument that, when supplied, makes the resulting message box non-blocking.  Instead of giving the result as a return value, the response is passed to the callback.  As such, always passing the callback (even when it didn't exist) would have been a backwards-compatible change.  To make Atom support both sync and async versions, I have updated `atom.confirm` to also accept an optional callback that determines whether or not to use the async version.  I also noticed while doing this that Atom's implementation was not fully in line with Electron's implementation and was limited in its functionality, and so the async version corrects this by passing all options directly to Electron.  Hopefully in Atom 2.0 we can make async the only supported version.  In the meantime, I've updated all of atom/atom's uses of `atom.confirm` to use the async API and updated the documentation to prefer the async version.

### Alternate Designs

The callbackiness is quite messy, especially when we used to return something based on the response value.  Anywhere where that occurred I had to pseudo-promisify the dialog and then await the resulting promise.  Would it therefore make more sense to directly promisify the async version of `dialog.showMessageBox`?

`atom.applicationDelegate.confirm` is implemented in core.

### Benefits

Improved UX when a message box appears.  See, for example, snow continuing to fall even when a message box appears:
async-message-box

### Possible Drawbacks

Unlike making the context menu async, this is an opt-in change.  Package authors will have to update their code to support the async message dialog.  As such, there is a possibility that people will create issues on atom/atom asking why certain message boxes are async while others aren't.  Furthermore, as discussed above, using the callback is much less clean than simply blocking and waiting for a return value.

### Applicable Issues

None.


UziTech:
👍 +1 on promisifying the async version

src/application-delegate.coffee
```diff
@@ -130,26 +130,34 @@ class ApplicationDelegate
...
-    })
-
-    if Array.isArray(buttons)
-      chosen
-    else
-      callback = buttons[buttonLabels[chosen]]
-      callback?()
+      # Legacy sync version: options can only have `message`,
+      # `detailedMessage` (optional), and buttons array or object (optional)
+      {message, detailedMessage, buttons} = options
```
UziTech: It seems like it would be very unintuitive, and therefore cause difficult to debug issues, if the sync and async versions don't have the same options

UziTech: I would vote for deprecating `detailedMessage` for `detail` but keep `buttons` as an array or object

50Wliu: Since the new async version is opt-in, I think that should immediately warn package authors that they need to look at their code and update it accordingly.  I explicitly didn't update the sync code path because the goal I had in mind when creating this PR was to completely deprecate and feature-freeze it to give package authors more "incentive" to moving to async.

iolsen:
Like the other similar PRs, we're 👍. Merge when ready.

PR #16230, opened 2017-11-18
Make issues links work even when not logged in

U8NWXD:
### Description of the Change

Update links in CONTRIBUTING.md to issue searches which work when not logged in to GitHub.

### Alternate Designs

The alternative would be to leave the links as they are, which could hinder user interaction.

The affected documentation is in Core.

### Benefits

The links will now work for users who are not logged in to GitHub.

### Possible Drawbacks

None.

### Applicable Issues

None.


50Wliu:
Hmm, both `/search` and `/issues` work for me: is that not what you're seeing?

U8NWXD:
Maybe it is dependent on whether you are signed into GitHub? Normally when I try to open https://github.com/issues?q=+is%3Aissue+user%3Aatom I get a 404 error page from GitHub, but when I try while logged in, it works. When I am logged in, both `/search` and `/issues` work, though the pages are designed differently.

50Wliu:
Ah. @lee-dohm I believe you're more familiar with this.

lee-dohm:
I updated the title and description to make it more clear what the problem and goals were.

PR #16254, opened 2017-11-21
Add comment to atom-environment.coffee

fjwldap:
### Description of the Change

I add comment to atom/src/atom-environment.coffee in line 566.
Most of them has comment of the function. For example, like in line 562, "# Extended: Returns a {Boolean} that is `true` if the current window is maximized." 
But function 'maximizeWindow()' doesn't have a comment about it.
So I add the comment.

It will be little bit more helpful to understand with comment. 

### Benefits

It will be little bit more helpful to understand with comment. 

### Possible Drawbacks

it is just comment...

### DIFF
--- a/src/atom-environment.coffee
+++ b/src/atom-environment.coffee
@@ -563,6 +563,7 @@ class AtomEnvironment extends Model
   isMaximized: ->
     @applicationDelegate.isWindowMaximized()
 
+  # Extended: Maximize the current window.  
   maximize: ->
     @applicationDelegate.maximizeWindow()
 


50Wliu:
@atom/maintainers is this a method that we want to expose to the public API?

lee-dohm:
> is this a method that we want to expose to the public API?

In my opinion, not really? I don't think that it will be any great cost but I'm not certain that packages should be maximizing the Atom window?

maxbrunsfeld:
@fjwldap These comments affect the documentation that gets generated for Atom. In general, we don't necessarily want to document every method on every class, because we don't want Atom packages to start depending on every method.

In addition, this method is so trivial that the comment doesn't add substantially to the code's readability.

Thanks anyway for taking the time to contribute, we are definitely open to documentation improvements in general.

pyaephyoswe:
pyaephyoswe569@gmail.com

PR #16257, opened 2017-11-22
Change dot-atom keymap.cson

isoeugenol:
- ctrl-shift-p and ctrl-p exist only macOS keymap
- i think example is same all platforms

### Requirements

dot-atom keymap.cson example is only exist macOS 
i think example should work all platforms.
### Benefits

if someone use example keymaps they work all platfroms

### DIFF
--- a/dot-atom/keymap.cson
+++ b/dot-atom/keymap.cson
@@ -14,8 +14,8 @@
 #   'enter': 'editor:newline'
 #
 # 'atom-workspace':
-#   'ctrl-shift-p': 'core:move-up'
-#   'ctrl-p': 'core:move-down'
+#   'ctrl-pageup': 'pane:show-previous-item'
+#   'ctrl-pagedown': 'pane:show-next-item'
 #
 # You can find more information about keymaps in these guides:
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for the submission!

The point of the examples is to simply give things that could be mapped if one chose to and to give people a basic understanding of the syntax. They aren't intended to be actual keybindings that one would want to keep, only to experiment with. Also, since not all keyboards have a <kbd>PgUp</kbd> or <kbd>PgDn</kbd> key (most notably many Mac keyboards don't) but every modern keyboard has a <kbd>Ctrl</kbd>, <kbd>Shift</kbd>, and <kbd>P</kbd> key, I feel the current examples work well for their intended purpose.

Thanks again for helping out! I'm going to close this but if you have any questions, please let us know :+1:

isoeugenol:
I have never used a Mac, so I can not imagine that reason.
when i was using window and linux i tried the example. but it is don`t work.
How do you feel about changing to "shift-up, shift-down"?

lee-dohm:
They're just examples to show the format, not something that people would seriously want in their keymap. For that goal, the current examples are fine. Thanks for your help.

Issue #12848, opened 2016-10-02
implement smooth scrolling speed setting


alvarolm:
a general scroll speed setting in the same manner sublime "scroll_speed" (http://docs.sublimetext.info/en/latest/reference/settings.html)  

why ?
the current atom editor has all the proper looks like any other major one, but severely lacks the feeling. Sublime, textmate and others have implemented smooth scrolling which gives an easy to fallow sensation when scrolling thru the code, this is a feature that makes an incredibly difference in terms of comfort and new users adoption.

whats wrong with the current alternatives ?
the css wont work with the new tile rendering system, see https://github.com/atom/atom/issues/5359
"scroll sensitivity" works only when using a mouse or trackpad, besides the "steeps" between every scroll update aren't smooth at all.

I propose an artificial delay in the same way that a linear transition of -webkit-transform would work over the tiles.

rel https://github.com/atom/atom/issues/1666

slayerjain:
I'm new to electron, but have experience in nodejs, wanna try!

lee-dohm:
@alvarolm Please update this Issue with a more detailed explanation of what it is you're requesting. External documentation links can be moved or change. We need to have a more permanent record of what it is you're asking for so that we can a) make a determination as to whether this is something we want to do and b) ensure that if it is something we want to include that we have actually achieved the goals when we make a change.

Ben3eeE:
What is wrong with the current setting that changes scroll speed?

![image](https://cloud.githubusercontent.com/assets/1058982/19068830/7f38bcb8-8a24-11e6-9a09-24957b6fc673.png)

alvarolm:
@lee-dohm  @Ben3eeE updated

alvarolm:
https://www.facebook.com/notes/daniel-colascione/buttery-smooth-emacs/10155313440066102/

PR #16286, opened 2017-11-28
Implemented smooth scrolling for mouse in editor.

mpcmarques:
### Description of the Change

Since the program relies on web technologies to build its user interface, we came up with a simple solution of introducing a smooth transition when the scrolling action occurs. 
The end goal was to make the scrolling transition last 0.2 seconds instead of being instant, leading to a smother user experience.
To achieve this goal, we wrote a css transition code in the style.less file. 

### Alternate Designs

The proposed version was selected because the issue can be solved with a simple css script.

Smooth scrolling makes scrolling feel more natural.

### Benefits

Scrolling is more natural for mouse users.

### Possible Drawbacks

No possible drawbacks were found.

### Applicable Issues

https://github.com/atom/atom/issues/12848


eltian:
I tested it and it does look good, but it doesn't sit well with other types of scrolling (dragging the minimap, using page up/ page down, etc). Is there a way to only apply the effect when scrolling with the mouse wheel?

oomek:
What is the status of that pr? I know it's old. Trying to find a way to enable smooth scrolling in Atom and this is the only thing that popped up.

PR #16294, opened 2017-11-29
Add `readonly` attribute to text editor element

kuychaco:
This PR allows for the use of a `readonly` attribute which disables the input for text editor elements: 

```<atom-text-editor readonly />```

The read-only state is stored on the `TextEditor` model. As of now this state merely prevents text from being input into the editor. Future considerations include using the state to determine modified status, preventing saving, etc.

/cc @maxbrunsfeld @damieng for 💭s

### CODE REVIEW

**Comment by maxbrunsfeld** [comment]
I might have missed some discussion here - why does saving the buffer clear the readonly status?

**Review by maxbrunsfeld — APPROVED** [review]

**Comment by t9md** [comment]
> I might have missed some discussion here - why does saving the buffer clear the readonly status?
>
> Other than that question, this looks great to me.

I want to ask this too.
I'm maintainor of [vim-mode-plus](https://atom.io/packages/vim-mode-plus).
In vim-mode-plus,

- normal-mode is achieved by calling `this.editor.element.component.setInputEnabled(false)`.
  - So `editor.component.isInputEnalbed()` return `false`
  - But once I save buffer by `cmd-s`, `editor.component.isInputEnalbed()` return `true`

**This fundamental editor.component's capability( input enabled or not ) should be not affected by irrelevant action like `cmd-s`**.

File: src/text-editor-component.js [hunk 1]
```diff
@@ -819,7 +823,7 @@ class TextEditorComponent {
 
     const oldClassList = this.classList
     const newClassList = ['editor']
-    if (this.focused) newClassList.push('is-focused')
+    if (this.focused && this.isInputEnabled()) newClassList.push('is-focused')
     if (model.isMini()) newClassList.push('mini')
```
t9md: Why this code assume `is-focused` only need to set only when `isInputEnabled()` was `true`?

src/text-editor.js
```diff
@@ -1106,7 +1129,7 @@ class TextEditor {
   setEncoding (encoding) { this.buffer.setEncoding(encoding) }
 
   // Essential: Returns {Boolean} `true` if this editor has been modified.
-  isModified () { return this.buffer.isModified() }
+  isModified () { return this.isReadOnly() ? false : this.buffer.isModified() }
```
t9md: I don't want this unnecessary guard. If readonly really disallow buffer modification, just returning `buffer.isModified()` should be enough.
This method should be just proxy of `this.buffer.isModified()` for clarity, and remove possibility of inconsistency   when it become `this.isReadonly !== this.buffer.isModifed()`.
Also this breaks https://github.com/t9md/atom-narrow pkg's important feature.

damieng: This is necessary otherwise it is not possible to create a read-only text editor from some in-memory data without it showing as modified on the tab.

damieng:
One of our scenarios for use of a read-only text editor is when we want to show you and editor but you can't edit the content because of where the text comes from - e.g. source that lives on a https:// url or inside a .jar file.  If however they chose to save that content to disk then by the very nature it is now editable as it lives somewhere the content can be changed.

src/text-editor.js
t9md: then, I want to say hack `this.buffer.isModified` by overriding `buffer.isModified = () => true`.
I feel some anxiety when seeing this fundamental return different answer based on `isReadOnly` status, and that status is unstable(changed by just saving buffer).
I can't believe such method, so just trying to relay on `buffer.isModified` directly.

t9md:
Sorry, I was really nervous for now, I know I should not write any comment with such mental state.
I won't respond for a while.

damieng:
No code is perfect :)   We did consider whether saving should clear the flag or not after @maxbrunsfeld also brought it up.  If it causes you difficulty we can remove that.

maxbrunsfeld:
We should remove it. We need to not break vim-mode-plus, and they are orthogonal concepts anyway.

Same thing with isModified. We can address the issue of opening files from .jars in another way that doesn’t bake assumptions into core.

damieng:
Removed the isModified/clear-on-save code ca53cf9ec1aed43edc05681029a23d379791586c

t9md:
I need `is-focused` breaking recerted  too.
This is direct breaking where style based in `is-focused` scope.
Readonly state and focus state is completely irrelevant.
Should not be coupled and should not break existing style just becausr of one single new feature.

t9md:
Basically want revert whole this PR and rethink to achieve it without breaking existing setInputEnabled capability.
I’m not insisting to not break existing feature ib every case of new enhancements, but I say should not break existing feature this time, since readonly feature should not be coupled/blended to these feature.
This is very fundamental feature vim-mode-plus depends on.
And original vim-mode was created by GitHub.
Please communicate each other if your team conside vmp as part of Atom’s goodness(I want say so).

src/text-editor-component.js
t9md: I need this unchanged, don't change meaning of `is-focused`.
Should allow `readonly` and `is-focused` buffer.
If some feature need exclude inputDisabled buffer from focused editor, do this by client side by checking state manually, don' bend core-meaning just because it's convenient.

src/text-editor.js
```diff
@@ -400,6 +401,16 @@ class TextEditor {
           }
           break
 
+        case 'readOnly':
+          if (value !== this.readOnly) {
+            this.readOnly = value
+            if (this.component != null) {
+              this.component.scheduleUpdate()
+            }
+            this.buffer.emitModifiedStatusChanged(this.isModified())
```
t9md: Should not emit "modified state changed" event when it's really not changed.
If some feature need hook for this, fire distinct event like `did-set-read-only` and do manual work in this.

t9md: I noticed this part is reverted in ca53cf9. sorry.

damieng:
Yes we're investigating and seeing what we can do. We've been able to
reverse out party of this but some of what remains it's needed elsewhere.
One of our team who worked on this is traveling and we needed some more
time to consider our options.

t9md:
@damieng 

> it's needed elsewhere.

I want to know where.

src/text-editor.js
t9md: Oh, I wanted to say use `buffer.isModified = () => false` to hide modified icon on tab.

t9md:
@damieng @maxbrunsfeld 

Confirmed with current master that I can maintain vmp if `is-focused` meaning is reverted to original.

```diff
diff --git a/src/text-editor-component.js b/src/text-editor-component.js
index 08af5ada1..263556ee0 100644
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -823,7 +823,7 @@ class TextEditorComponent {

     const oldClassList = this.classList
     const newClassList = ['editor']
-    if (this.focused && this.isInputEnabled()) newClassList.push('is-focused')
+    if (this.focused) newClassList.push('is-focused')
     if (model.isMini()) newClassList.push('mini')
     for (var i = 0; i < model.selections.length; i++) {
       if (!model.selections[i].isEmpty()) {
```

I think this request is reasonable, in vmp's `normal-mode`, it's `readonly` as long as it's `focused`.
The `is-focused` and `readonly` states should not be mutually exclusive.

kuychaco:
@t9md the change has been reverted in 853b9a4e24502af82efaba289f7e0e3fcdf0bc19. We'll get beta hotfixed later today.

t9md:
@kuychaco Thank you for hearing.

fabianfiorotto:
Readonly editors can be modified using middle click paste on Linux. I tested it on Linux Mint 17.3 with atom 1.24.0-beta1

damieng:
@fabianfiorotto Nice find, that should be raised as a separate issue - comments on PR's can't be tracked as work items.

Issue #13909, opened 2017-03-01
Window close does not trigger save window state


Krajsnick:
### Description

Closing the Atom window does not save the window UI state, _read_ window size and position. To save the window state I have to **blur** (click outside) the window, then close it. Reopening the window after blurring it works.

According to the code in `window-event-handler.coffee` the window state should be stored both when a window closes and when blurred:
```coffee
handleWindowBlur: =>
    @atomEnvironment.storeWindowDimensions()

handleWindowBeforeunload: (event) =>
  @atomEnvironment.storeWindowDimensions()
```

### Steps to Reproduce

1. `atom .` to open project of your choice
2. Resize window, maintain window focus
3. Close window
4. Back to terminal, `atom .` again

Redo step **1-2** and before closing the window in step **3**, **blur** the window, click your desktop. Then close the window and continue with step **4**, the new window should open with the same dimensions as the window you just closed.

**Expected behavior:** The window you open in step **4** should be restored to the resized state, when you closed the window.

**Actual behavior:** The window you open in step **4** is restored back to the state before you resized, in step **2**.

### Versions

Atom    : 1.14.4
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
macOS 10.12.3

### Additional Information

I did try to start Atom in safe mode, and also with the `--clear-window-state` flag. Even deleting my `~/.atom/storage`.

You can also reload the window and see the same behaviour, instead of closing and re-opening a window. Resize Atom window then reload, watch it reset back to old dimensions, before reload. Watch the gif for a demo:

![window-state](https://cloud.githubusercontent.com/assets/979555/23480337/e966c98c-fec7-11e6-9ecd-91e23df89891.gif)

rsese:
On 1.14.4 and macOS 10.12.3, I was able to reproduce this case:

> You can also reload the window and see the same behaviour, instead of closing and re-opening a window. Resize Atom window then reload, watch it reset back to old dimensions, before reload.

But not the clicking outside of Atom and then closing/opening a window case.

@Krajsnick when you say:

> before closing the window in step 3, blur the window, click your desktop. Then close the window

Do you mean that after clicking on the desktop, you just go back to the Atom window and click the x to close the Window?

Krajsnick:
@rsese 
> Do you mean that after clicking on the desktop, you just go back to the Atom window and click the x to close the Window?

Yes, after having the window lose focus I just click the Atom editor pane again, then go for the red traffic light.

[...]

PR #15514, opened 2017-08-31
Fix window state not saving on close

richardsimko:
### Description of the Change

Fix #13909. It seems there have been fixes for saving window state previously however these only save on `blur`, and `beforeUnload`. None of these are triggered if you immediately close the window.

### Alternate Designs

None

### Why Should This Be In Core?

It's a bug fix.

### Benefits

The bug is fixed.

### Possible Drawbacks

None

### Applicable Issues

#13909

PR #16295, opened 2017-11-29
Add event handler for window resizing.

xfontes42:
### The Issue
Atom window state is not being saved. This means that after closing and reopening Atom, the window size is not the same it was before closing. However, blurring the window saves its dimensions.

### Source code files
The window dimensions are being saved (_.atomEnvironment.storeWindowDimensions()_) on _handleWindowBlur()_ and _handleWindowBeforeunload()_ functions in [atom/src/window-event-handler.js](https://github.com/atom/atom/blob/master/src/window-event-handler.js). However, there should be an event handler function responsible for handling window resizing. If that event handler existed, then it would save the window dimensions every time they were modified.

### Design of the fix

AtomEnvironment is a class used to deal with packages, themes, menus and the window. This class contains the functions to save and get the current window dimensions.
 
WindowEventHandler is the class containing all the handlers for the window. The handlers worth checking for this issue are _handleWindowBlur()_ and _handleWindowBeforeUnload()_. These are the handlers where the window dimensions are saved. 

In the diagram above we colored the function that we added in green. This function is a handler that activates whenever the user changes the window dimensions. 

### Fix source code
We first created the new event handler by adding this line of code:
```
this.addEventListener(this.window, 'resize', this.handleWindowResize)
```

Then, we had to bind the handler:
```
this.handleWindowResize = this.handleWindowResize.bind(this)
```

Finally inside the handler, we placed the function that saves the window dimensions, like this:
```
handleWindowResize () {
    this.atomEnvironment.storeWindowDimensions()
  }
```

With this fix, we make sure that the window dimensions are always saved whenever the window is resized.

## Alternate Designs

We considered using setInterval to routinely call the function that stores the window dimensions but this method was discarded as it wasn't event driven and slightly lowered atom's performance.

## Why Should This Be In Core?

People expect atom to save the window size everytime they resize it so that when they open atom the previously saved size is restored.

## Benefits

Explained above.

## Possible Drawbacks

No relevant drawbacks.

## Applicable Issues
[Window close does not trigger save window state #13909](https://github.com/atom/atom/issues/13909)

### DIFF
--- a/spec/window-event-handler-spec.js
+++ b/spec/window-event-handler-spec.js
@@ -59,8 +59,6 @@ describe('WindowEventHandler', () => {
     })
   )
 
-
-
   describe('when a link is clicked', () =>
     it('opens the http/https links in an external application', () => {
       const {shell} = require('electron')
--- a/spec/window-event-handler-spec.js
+++ b/spec/window-event-handler-spec.js
@@ -58,7 +58,7 @@ describe('WindowEventHandler', () => {
       spyOn(atom, 'close')
       window.dispatchEvent(new CustomEvent('window:close'))
       expect(atom.close).toHaveBeenCalled()
-    })     
+    })
   )
 
   describe('when a link is clicked', () =>
--- a/spec/window-event-handler-spec.js
+++ b/spec/window-event-handler-spec.js
@@ -51,7 +51,15 @@ describe('WindowEventHandler', () => {
       window.dispatchEvent(new CustomEvent('window:close'))
       expect(atom.close).toHaveBeenCalled()
     })
+
+    it ('saves the window state', () => {
+      spyOn(atom, 'storeWindowDimensions')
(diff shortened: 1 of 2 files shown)


praveeno (approved):
looks good to me

praveeno:
I like the way you describe you PR, +1

ungb:
I was able to test these changes and it looks good to me.  @xfontes42 any chance you can add test to this?  There's another PR (https://github.com/atom/atom/pull/15514) that tries to do the solve the same issue but I think this solution actually solves it for all the cases as I tried closing with the red-x, cmd-q and windows:close and the size is saved.

The other PR has a test, could you possibly move that over to this PR and add any others you feel would be useful to add?  

/cc @leroix  for code review

leroix:
Assuming the tests pass, it looks good to me 👍

ungb:
Tested this and also added test for this!

thanks @xfontes42 for making this change!

Issue #11838, opened 2016-05-25
Code folding based on grammar instead of indentation


fruggiero:
The actual folding mechanism is simply wrong and unreliable.
I often work with big files and use folding to speed up my work.

But, if the file has a wrong indentation (something not rare if the files are human edited) atom also fold the code in a wrong manner. 
This in some case can be extremely dangerous , because if you don't remember to check every time what is folded you take that as given and work on the other parts of the file in a wrong manner.

The actual code folding is based firstly on indentation then on grammar.

I think that the code folding should be based ONLY on grammar.
Because, actually, even when you have a correct grammar (with correct folding rules) you can't be sure that the editor fold the code in the right manner.

Here is some examples to show , editing an XML file (**i want to fold at line number 10**):

**Unfolded text (with wrong indentation at line 18 and 19)**:
![image](https://cloud.githubusercontent.com/assets/4760358/15536250/ce8dff60-2270-11e6-9307-8247d72f057e.png)

**Folded text (WRONG because of wrong indentation)**:

![image](https://cloud.githubusercontent.com/assets/4760358/15536276/ef14bfd0-2270-11e6-99be-cec427338a9d.png)

What the correct folding should be:

![image](https://cloud.githubusercontent.com/assets/4760358/15536312/19aa2244-2271-11e6-8548-e1a5aa3345bc.png)

But for achieve this we must fix the indentation (on line 18 and 19):

![image](https://cloud.githubusercontent.com/assets/4760358/15536453/8e0a602c-2271-11e6-91a7-0076bec79f33.png)

I think that correct indentation CAN'T be a requirement to correct code folding.

timrosenblatt:
Bumping this. I love Atom, but haven't been able to find a solution for this. I notice it with HEREDOC syntax as well. 1000 internet points to whoever fixes this :)

twainj:
Agreed. I'm a new Atom user, I am trying it because I picked up work on a Classic ASP project, and have to navigate poorly organized code in overlarge files. Atom actually having a contributed grammar for ASP has been a godsend.

However, being new to Atom, I never would have thought that the folding was based on indentation.
I discovered it today purely by accident, and I'm glad I did, as I was starting to wonder if the uncollapsible code I was finding was because of something strange in the code.  I was prepared to trust Atom instead of the devs that came before me on this project.

Folding based on indentation makes sense in a whitespace-sensitive language (python, etc), and I can see how this might have been a shortcut to get some functionality fast. But it really should be based on the grammar.

[...]

PR #16299, opened 2017-11-30
Allow Tree-sitter parsers to be used for syntax highlighting and code folding

maxbrunsfeld:
### Overview

This pull request adds a new config setting to Atom: `core.useTreeSitterParsers`, which defaults to `false`. If you set that setting to `true`, when editing files written in certain languages, Atom will use a new parsing system called [Tree-sitter](https://github.com/tree-sitter/tree-sitter) to provide improved syntax highlighting and code folding.

### Syntax Highlighting Improvements

Using the syntax trees provided by Tree-sitter, we can produce syntax highlighting that is more consistent and specific than before. Different kinds of *names* in your code like types, functions, properties/fields,  and local variables should be grouped by color in a more consistent way.

The new syntax highlighting uses the same CSS classes as the old system, so you should be able to continue using your current themes without disruption.

### Code Folding Improvements

Currently, Atom's code folding is based on indentation: in a fold, lines with greater indentation are hidden. This usually gives good results, but it is unintuitive in some cases. Now, by basing the code folding on the *syntax tree* provided by Tree-sitter, we can fold code in an intuitive way regardless of its indentation.

<details>

<summary>Example</summary>

Code like this:

```c
int add(int left,
        int right) {
  return left + right;
}
``` 

would previously fold like this:

```c
int add(int a…
```

Now, we are able to fold it like this:

```c
int add(int left,
        int right) {…}
```

or, if you choose to fold the first *two* lines:

```c
int add(…) {…}
```

</details>

### New Editing Commands

This PR adds two new text editing commands: `editor:select-larger-syntax-node` and `editor:select-smaller-syntax-node`. These commands can be used to quickly select syntactically meaningful units of code: 

<details>

<summary>Example</summary>

extend-selection-2

</details>

<details>

<summary>Multi-Cursor Example</summary>

extend-selection-3

</details>

### Supported Languages

Initially, the following languages will be supported:

* [x] JavaScript (https://github.com/atom/language-javascript/pull/546)
* [x] TypeScript (https://github.com/atom/language-typescript/pull/14)
* [x] Python (https://github.com/atom/language-python/pull/220)
* [x] C (https://github.com/atom/language-c/pull/257)
* [x] C++ (https://github.com/atom/language-c/pull/257)
* [x] Go (https://github.com/atom/language-go/pull/146)
* [x] Bash (https://github.com/atom/language-shellscript/pull/93)

### Next Steps

The new parsing system is disabled by default because there are some known bugs and it needs more testing before we roll it out to all Atom users. After this PR lands, I'll be focused on fixing bugs in the new parsing system and the existing grammars so that soon, we can enable the new parsing system by default.


alexandernst:
\o/ bravo  \o/

Alex3937 (approved):

calebmeyer:
@maxbrunsfeld I noticed you have some commits on https://github.com/tree-sitter/tree-sitter-ruby, but it's not included in the list of initial languages. Do you have a sense of the scope of the work to get it integrated? Is this something there is interest in?

maxbrunsfeld:
@calebmeyer Yeah, `tree-sitter-ruby` is actually quite complete and well-tested. I have [a pull request open](https://github.com/atom/language-ruby/pull/225) to incorporate it into atom's `language-ruby` package.

Unfortunately, there is an issue with `tree-sitter-ruby` where the generated C code is quite large (much larger than any of the other parsers), so that it takes around 30 seconds to compile and the binary adds upwards of 8MB to the size of the app. I believe this is an issue with `tree-sitter` itself, and I'd like to address it before shipping the ruby parser with Atom.

In the meantime though, if you are interested, you can try cloning `language-ruby` and checking out that branch. I think there are currently some issues with the folding that need to be sorted out as well.

Arcanemagus:
Looks like this will fix #11838, at least for the supported languages.

src/tree-sitter-language-mode.js
```diff
@@ -0,0 +1,506 @@
...
+  constructor ({buffer, grammar, config}) {
+    this.id = nextId++
+    this.buffer = buffer
+    this.grammar = grammar
+    this.config = config
+    this.document = new Document()
+    this.document.setInput(new TreeSitterTextBufferInput(buffer))
+    this.document.setLanguage(grammar.languageModule)
+    this.document.parse()
+    this.rootScopeDescriptor = new ScopeDescriptor({scopes: [this.grammar.id]})
```
Xapphire13: This will break certain atom packages that depend on this being `this.grammar.scopeName` like the current [text-mate-language-mode](https://github.com/atom/atom/blob/8c92ed96525e29397b706510556b2caf54f572e4/src/text-mate-language-mode.js#L34).

Examples of packages that will break:
* [atom-typescript](https://github.com/TypeStrong/atom-typescript/blob/34072c896ebf8b54620b447139c2df0492f1faf0/lib/main/atom/utils/atom.ts#L24-L27)

stadelmanma:
@maxbrunsfeld  With the right changes do you think this parser logic could be applied to provide smarter autocomplete suggestions? i.e. not showing variables that are locally scoped in a different function, class, etc. and giving priority to those present in the current scope?

src/tree-sitter-language-mode.js
maxbrunsfeld: Yeah, we're intentionally using different-looking scope names because the whole idea of the dot-separated names like `source.ts` originates from TextMate and isn't our ideal API. In the long-term, we'd like for scoped settings to use simple language names like `typescript` instead.

We have added the concept of 'scope synonyms' to the Atom config system, so that in the meantime, you'll be able to apply scoped settings to `source.js` in your config system and it will apply to the TreeSitter grammar whose id is `javascript`.

I agree that this is going to break packages. Before we enable the `core.useTreeSitterParsers` setting by default, we'll need to figure out more of a solution for this. Most likely, it will involve a combination of API additions and also opening large numbers of pull requests on community packages to help keep everyone's code working.

maxbrunsfeld:
Alright, I've been using this for my daily work for a few weeks. There are things that need to be fixed before we turn on the feature flag by default, but I think it's usable enough to merge so that more people can test it out.

calebmeyer:
If I want to try this out, will I need to build from master? I assume it will go out in the next beta, but github doesn't publish releases early.

If I do encounter an issue, where is the right place to report it?

50Wliu:
https://ci.appveyor.com/api/buildjobs/rr249tsf58qam1ar/artifacts/out%2Fatom-windows.zip, but it doesn't look like any of the language PRs have been merged so you'll need to check them out and `apm link` them manually.

> If I do encounter an issue, where is the right place to report it?

Create a new issue here unless you're pretty sure it's an issue with one of the parsers.

maxbrunsfeld:
Oh thanks for reminding me to merge the language PRs, @50Wliu. You wouldn't need to `apm link` them though; this PR upgrades the language packages (just to a prerelease version).

maxbrunsfeld:
Merging those language PRs now so that we can point Atom back at the stable versions instead of prerelease versions.

maxbrunsfeld:
Ok, the language package PRs are now merged and the latest commit of Atom is now pointing at stable versions of the language packages, rather than prereleases.

maxbrunsfeld:
For convenience: [Mac download](https://6518-3228505-gh.circle-artifacts.com/0/Users/distiller/atom/out/atom-mac.zip), [Windows download](https://ci.appveyor.com/api/buildjobs/f54vvonq2y2ink4a/artifacts/out%2Fatom-windows.zip).

maxbrunsfeld:
@stadelmanma Yeah, I do think we can build a smarter autocomplete using Tree-sitter. I'm not exactly sure how it will work yet.

Aerijo:
@maxbrunsfeld Any tips on developing new community packages with tree-sitter? I haven't been able to get any working in the Atom build you linked to here. The built in ones are applied, but no community tree-sitter packages are being recognised.

I asked about this [in the forums](https://discuss.atom.io/t/help-setting-up-tree-sitter-grammar-development/52113) yesterday, but I think I might get better luck here.

maxbrunsfeld:
@Aerijo So repositories like [tree-sitter-javascript](http://github.com/tree-sitter/tree-sitter-javascript) are not themselves Atom packages; they are just node modules which can be *used* by Atom packages to provide syntax highlighting. For an example of an Atom package that uses this module, see [language-javascript](https://github.com/atom/language-javascript). Specifically, the [tree-sitter-javascript.cson](https://github.com/atom/language-javascript/blob/master/grammars/tree-sitter-javascript.cson) file.

Basically, to use a Tree-sitter grammar for syntax highlighting and code folding (and other features in the future), you create an Atom package with a `grammars` folder. This is similar to how the existing TextMate-style grammars work. There should be a file in the `grammars` folder with a similar format to the `tree-sitter-javascript.cson` file I linked above. The file's name does not matter. The important keys in the file are:

* `parser` - the name of the tree sitter parser node module that will be used for parsing
* `id` - a unique, stable identifier for the language
* `scopes` - an object where the keys are CSS selectors that select nodes in the syntax tree, and the values are strings that imply HTML classes to apply to the given nodes for styling.

Sorry for the lack of documentation. This feature is still a bit unstable currently, but I will be finishing it up and documenting it in the coming months.

Aerijo:
@maxbrunsfeld So I need to publish the actual tree sitter grammar to npm, and add a file to my existing Atom grammar package that has the name of this npm module? Is there a way for it to be self contained within an Atom package?

maxbrunsfeld:
The parser field can also be a relative path, so the actual parser module could just be embedded in your package repository.

mkutny:
@maxbrunsfeld , really decent feature, thanks for developing it!

Have noticed that the following snipped gets folded non-intuitively (as to my eyes). Want to know if its work as designed or I should open an issue.

Code likes this:
```javascript
function add({
  a, b
}) {
  return a + b;
}
```

Folds into:
```javascript
function add({…}) {
  return a + b;
}
```
with no possibility to fold further (no angle bracket to the left).

maxbrunsfeld:
@mkutny Ah yeah, that's a bug: Atom doesn't know what kind of 'fold arrow' to show in the gutter if a line is foldable *and* already has one fold. You should still be able to fold the body of the function with the keyboard though: `Editor: Fold Current Row` is the command; `cmd-alt-[` on mac.

mkutny:
Yeah, that's understandable.

Still, one would expect the whole function to collapse, not only its args.

From grammar perspective
```javascript
function add({
  a, b
}) {
  return a + b;
}
```
is equivalent to
```javascript
function add({ a, b }) {
  return a + b;
}
```
but folding `function` line gives different results in these two cases.

maxbrunsfeld:
@calebmeyer Just an update on the Ruby parser. It took a long time because of some internal issues, but it's now working well and will be included in Atom 1.30.

PR #16328, opened 2017-12-05
Update auxiliary click logic for moving cursors

### LINKED ISSUE #16324: Right clicking on selected text deselects the text
### Prerequisites

### Description

selection-broken

Previously:
selection-ok

### Steps to Reproduce

1. Open a document
2. Select some text
3. Use right click to for example to copy the text

**Expected behavior:** Selections remain intact

**Actual behavior:** Everything is deselected and cursor is created.

**Reproduces how often:** Always

### Versions

Reproducible using the latest build 4f73e81080017303a66dcb020c8b451e0c4ca4dd

### Additional Information

Tried only on Windows 10.

Seems to be broken by #16192 

cc @50Wliu

winstliu:
### Requirements

### Description of the Change

* When right-clicking when more than one cursor is present or if there is a non-empty selection, the cursor should not be moved to retain proper context
* When middle-clicking, the cursor should always be moved (same behavior as on master)

### Alternate Designs

None.

This is a regression fix for #16324.

### Benefits

More native right-click handling.

### Possible Drawbacks

Potentially more cases that I didn't think of.

### Applicable Issues

Fixes #16324

/cc @Ingramz

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2840,83 +2840,149 @@ describe('TextEditorComponent', () => {
 
   describe('mouse input', () => {
     describe('on the lines', () => {
-      it('positions the cursor on single-click or when middle/right-clicking', async () => {
-        for (const button of [0, 1, 2]) {
+      describe('when there is only one cursor and no selection', () => {
+        it('positions the cursor on single-click or when middle/right-clicking', async () => {
+          for (const button of [0, 1, 2]) {
+            const {component, element, editor} = buildComponent()
+            const {lineHeight} = component.measurements
+
+            editor.setCursorScreenPosition([Infinity, Infinity], {autoscroll: false})
+            component.didMouseDownOnContent({
+              detail: 1,
+              button,
+              clientX: clientLeftForCharacter(component, 0, 0) - 1,
+              clientY: clientTopForLine(component, 0) - 1
+            })
+            expect(editor.getCursorScreenPosition()).toEqual([0, 0])
+
+            const maxRow = editor.getLastScreenRow()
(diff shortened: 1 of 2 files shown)


Ingramz:
Compared to textmate/sublime text (at least on mac) and the way how it worked in Atom before, right click moving the cursor doesn't seem to make much sense (when pasting for example). Previously it just worked as if the user simply used the context menu key on the keyboard (on windows, maybe linux too), so that the right click did not mutate the cursors in any way.

50Wliu:
Hmm, I would consider it a lot more intuitive for the context menu action to occur where I right-clicked rather than where my cursor previously was.  This way I don't have to move the cursor and then right-click in order to paste something.  I agree however that the dedicated context menu button on a keyboard should be rooted at the cursor position.  This is consistent with how Chrome operates.

Ingramz:
I am not going to argue in regards to what the correct behavior of right click should be, because I don't know the answer. What I am trying to say is that I think that there was nothing wrong with the way it worked before.

It makes sense if everything worked the same regardless of the number of cursors/selections that have been made. Multiple selections is also something that you can't do in Chrome, which is why I find that the UX of a web browser doesn't always translate very well to a text editor. Also compare text field behavior to address bar in Chrome - inconsistent.

I do not mind adapting to the new behavior, if it is collectively decided that we are going this route. However, I can see this change upsetting some if not handled properly, so one should be prepared for that. Maybe I am alone in that regard, who knows.

Let's just see what others think before hastily making any decisions.

50Wliu:
@Ingramz I'll be merging this for now but will revisit after holidays.

Issue #16024, opened 2017-10-30
'Decrease Text Size' for Atom in Mac shows an underscore in the menu


TristanPeers:
### Prerequisites

    [x] Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    [x] Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    [x] Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    [x] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    [x] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Decrease Text Size for Atom in Mac shows an underscore in the menu, which while it is the same key showing a minus symbol would both look better and be more intuitive.

Screenshot:

![image](https://user-images.githubusercontent.com/9379533/32160048-f16abfbe-bd47-11e7-856e-a28df770637c.png)

### Steps to Reproduce

1. Open Atom on Mac
2. Click 'View' menu.

**Expected behavior:** A minus symbol is used

**Actual behavior:** An underscore symbol is used

**Reproduces how often:** 100%

### Versions
```
~ $ atom --version
Atom    : 1.21.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
~ $ apm --version
apm  1.18.5
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.13.5
```
### Additional Information

None

50Wliu:
[Both keybindings](https://github.com/atom/atom/blob/2f0e472e629fe8fef8534a668ab0d552fb95f3a9/keymaps/darwin.cson#L82-L83) are correct, though it looks like the accelerators are preferring the one that is defined later, which may be counterintuitive.

PR #16342, opened 2017-12-07
'cmd--': As default for 'window:decrease-font-size' on MacOs

davidrsfalcao:
### Description of the Change

The keybinding 'cmd--' appears  instead of  'cmd-_' on  'View' drop down 
<br>
captura de ecra 2017-12-07 as 19 06 29

### Alternate Designs

none.

Fix issue https://github.com/atom/atom/issues/16024

### Benefits

'cmd--' is more used and more pratical too.

### Possible Drawbacks

none

### Applicable Issues

https://github.com/atom/atom/issues/16024

### DIFF
--- a/keymaps/darwin.cson
+++ b/keymaps/darwin.cson
@@ -79,8 +79,8 @@
   'ctrl-shift-tab ^ctrl': 'pane:move-active-item-to-top-of-stack'
   'cmd-=': 'window:increase-font-size'
   'cmd-+': 'window:increase-font-size'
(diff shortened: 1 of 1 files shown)


Aerijo:
Nice. This technically makes it inconsistent (regarding pressing `shift` or not) with the displayed `=/+` shortcut, but seeing as both `_` and `-` do the same thing it shouldn't be a problem.

More to the point, who designed the keyboard to have `-=` and `_+` be pairs, but not `-+`? 😄

PR #16375, opened 2017-12-13
Try adding python 2.x back - newer Travis images dropped it

damieng:
Newer travis Linux build images have dropped Python 2.x 

node-gyp requires this to build Atom so builds are now failing.


nathansobo:
Nice work!

Issue #15992, opened 2017-10-25
Uncaught TypeError: Cannot read property 'length' of undefined


peterbe:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

```
<embedded>:25444 Uncaught (in promise) TypeError: Cannot read property 'length' of undefined
    at Project.module.exports.Project.setPaths (/Applications/Atom.app/Contents/Resources/app/src/project.js:271:44)
    at /Applications/Atom.app/Contents/Resources/app/src/project.js:143:30
```
![screen shot 2017-10-25 at 1 34 06 pm](https://user-images.githubusercontent.com/26739/32013452-30cb0f8c-b989-11e7-80ff-c763783d62ba.png)

Happens with or without `atom --safe .` 

### Steps to Reproduce

1. `git clone git@github.com:mozilla/balrog.git`
2. `cd balrog`
3. `docker-compose up`
3. `atom .`

Note. The `docker-compose up` creates a bunch of files in `.cache` and a bunch of files in `ui/node_modules`

**Expected behavior:** [What you expect to happen]
That it opens without errors and that I can open files.

**Actual behavior:** [What actually happens]
It never opens properly. No side-pane. See screenshot. 

**Reproduces how often:** [What percentage of the time does it reproduce?]
Every time. 

### Versions

```
▶ atom --version
Atom    : 1.21.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
▶ apm --version
apm  1.18.5
npm  3.10.10
node 6.9.5 x64
python 2.7.11
git 2.9.0
```

I'm on macOS Sierra, 10.12.6

### Additional Information

I don't know how to debug (or even share) what extensions I have installed that might affect this. 

Note that the error only happens *after* you've run `docker-compose up` which creates a bunch of files. But it's nevertheless reproducible. The files that are created should all be in `.gitignore`.

[...]

PR #16392, opened 2017-12-15
Resolve Uncaught TypeError

dabao12321:
A TypeError was raised when a certain file, or lack thereof, causes Atom to open improperly, without the side pane appearing. An issue has already been filed (#15992) recreating the steps to get this error message. The solution seems to be that users must clear their Atom window state, then reopen the program. This pull request alerts users to do so manually-- in the future, perhaps a more streamlined solution would be to automatically clear the window state had this specific TypeError been thrown.


sadick254:
@dabao12321 Thank you for your contribution. Belated congratulations on your first contribution. There are conflicts that need resolving. Do you mind resolving them.

src/project.js
```diff
@@ -350,7 +350,13 @@ module.exports = class Project extends Model {
   // Public: Get an {Array} of {String}s containing the paths of the project's
   // directories.
   getPaths() {
-    return this.rootDirectories.map(rootDirectory => rootDirectory.getPath());
+    try {
+      return this.rootDirectories.map(rootDirectory => rootDirectory.getPath());
+    } catch (e) {
+      console.log(
+        "Please clear Atom's window state with: atom --clear-window-state"
+      );
```
aminya, line 358:
```suggestion
      atom.notifications.addError(
        "Please clear Atom's window state with: atom --clear-window-state"
      );
```

aminya (changes requested):

sadick254 (approved):
Good addition @aminya 👍

src/project.js
```diff
@@ -350,7 +350,13 @@ module.exports = class Project extends Model {
...
   // directories.
   getPaths() {
-    return this.rootDirectories.map(rootDirectory => rootDirectory.getPath());
+    try {
+      return this.rootDirectories.map(rootDirectory => rootDirectory.getPath());
+    } catch (e) {
+      atom.notifications.addError(
+        "Please clear Atom's window state with: atom --clear-window-state"
+      );
+    }
```
DeeDeeG, line 359: I wonder if the `try{}catch{}` could be moved to inside the `this.rootDirectories.map()` function? So that individual valid file paths are still returned, but just the specific non-valid paths are skipped?

aminya: That has a negative performance issue for those who don't have this issue. Try-catch is not free.

DeeDeeG:
@sadick254 I wonder if this PR is the reason for failure to launch in some cases in the Electron 9 PR (https://github.com/atom/atom/pull/21777).

I see the function doesn't return anything on one of the code paths now.

I think it should always return an array, even if the `try{} catch{}` fails.

PR #16397, opened 2017-12-16
Initialize ProtocolHandlerInstaller after initializing Config

### LINKED ISSUE #16201: Registration of protocol handler prompts when set to 'never'
I get this with every launch even when I set the config to never;

image

I put some debugging in and at run-time when it does the check it is 'prompt' although when I examine it with atom.config.get after launch it is 'never'.

I suspect the check happens too early in the start-up to get the actual value.

Additionally we want to make sure this dialog doesn't come up when running specs?  It looks like it interferes with some of the UI tests.  @BinaryMuse

segevfiner:
### Requirements

### Description of the Change

Initialize `ProtocolHandlerInstaller` after initializing Config. Config is initialized synchronously, so it should be OK to simply reorder them. Without this `ProtocolHandlerInstaller` fails to read the config value on `initialize` which causes the notification to be shown each time on Atom startup even if you select 'never'.

This allows it to correctly read `core.uriHandlerRegistration` and avoids popping the notification even if set to 'never'.

### Alternate Designs

N/A

Fixes a bug.

### Benefits

Fixes a bug.

### Possible Drawbacks

Code formatting?

### Applicable Issues

Fixes #16201

cc @damieng @BinaryMuse

### DIFF
--- a/src/atom-environment.js
+++ b/src/atom-environment.js
@@ -206,11 +206,12 @@ class AtomEnvironment {
     this.themes.initialize({configDirPath: this.configDirPath, resourcePath, safeMode, devMode})
 
     this.commandInstaller.initialize(this.getVersion())
-    this.protocolHandlerInstaller.initialize(this.config, this.notifications)
     this.uriHandlerRegistry.registerHostHandler('core', CoreURIHandlers.create(this))
     this.autoUpdater.initialize()
 
     this.config.load()
+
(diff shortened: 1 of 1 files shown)


BinaryMuse:
Thanks for opening a PR for this. Looks good 👍 

I'm re-running the failed build (I believe it was just a flake) and once it goes green will merge.

PR #16441, opened 2017-12-24
Remove trailing whitespace from text editor docs

malept:
### Description of the Change

Remove trailing whitespace from text editor API docs so the linter is happy (and thus the tests pass 😄)


lee-dohm:
Thanks for this :+1:

PR #16455, opened 2017-12-27
Ensure that new editors get unique ids

### LINKED ISSUE #16454: TextEditor IDs are not always unique
### Prerequisites

### Description

In short, the `.id` attribute of each text editor in the workspace is not always unique among the set of all text editors. This breaks packages that look up a reference to an editor by ID, including [markdown-preview](https://github.com/atom/markdown-preview) and [glsl-preview](https://github.com/fordhurley/atom-glsl-preview).

### Steps to Reproduce

1. Open a project directory containing several files, clearing window state: 

        atom --clear-window-state --safe .

2. Open text editors for a few files.

3. Open dev tools and print the `.id` for each editor: 

       atom.workspace.getTextEditors().map(e => e.id)
       -> [0, 1, 2]

4. Quit and relaunch atom, this time without clearing the window state:

        atom --safe .

5. The previous three editors should be restored by atom. Open a fourth for a different file.

6. Open dev tools and print the `.id` for each editor: 

       atom.workspace.getTextEditors().map(e => e.id)
       -> [0, 1, 2, 0]

**Expected behavior:** I expected the fourth editor to get the id `3`, unique from the others.

**Actual behavior:** The fourth editor was assigned the id `0`, the same as the first editor.

### Versions

    $ atom --version
    Atom    : 1.23.1
    Electron: 1.6.15
    Chrome  : 56.0.2924.87
    Node    : 7.4.0

    $ apm --version
    apm  1.18.11
    npm  3.10.10
    node 6.9.5 x64
    atom 1.23.1
    python 2.7.14
    git 2.13.2

macOS 10.13.1

### Additional Information

I believe this bug was introduced by e9e23a2d09832f578874020ed10deada2c954eb6. This commit also caused editor ids to start from 0 instead of 1, which caused a minor issue in [glsl-preview](https://github.com/fordhurley/atom-glsl-preview) that led me down this road of investigation.

fordhurley:
This restores the behavior from when `TextEditor` was written in Coffeescript and extended the `Model` class.

### Description of the Change

With this change, editors opened after restoring the window state will get assigned IDs starting from `max(editor IDs) + 1`.

### Alternate Designs

An alternative could be to select the lowest available unique ID. This would require a more complicated ID generator, but is certainly possible if a more dense ID space is desirable. 

Assuming that editor IDs are supposed to be unique, this is a bug in core.

### Benefits

Packages that look up editor instances by ID will be able to resolve to the correct editor.

### Possible Drawbacks

I'm not aware of any drawbacks.

### Applicable Issues

Fixes #16454.

### DIFF
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -20,14 +20,15 @@ describe('TextEditor', () => {
     await atom.packages.activatePackage('language-javascript')
   })
 
-  it('generates unique ids for each editor', () => {
-    // Deserialized editors are initialized with an id:
-    new TextEditor({id: 0})
-    new TextEditor({id: 1})
-    new TextEditor({id: 2})
-    // Initializing an editor without an id causes a new id to be generated:
-    const generatedId = new TextEditor().id
-    expect(generatedId).toBe(3)
+  it('generates unique ids for each editor', async () => {
+    // Deserialized editors are initialized with the serialized id. We can
+    // initialize an editor with what we expect to be the next id:
+    const deserialized = new TextEditor({id: editor.id+1})
+    expect(deserialized.id).toEqual(editor.id+1)
+
+    // The id generator should skip the id used up by the deserialized one:
+    const fresh = new TextEditor()
+    expect(fresh.id).toNotEqual(deserialized.id)
   })
 
   describe('when the editor is deserialized', () => {
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -20,6 +20,16 @@ describe('TextEditor', () => {
     await atom.packages.activatePackage('language-javascript')
   })
 
+  it('generates unique ids for each editor', () => {
+    // Deserialized editors are initialized with an id:
+    new TextEditor({id: 0})
+    new TextEditor({id: 1})
+    new TextEditor({id: 2})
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks for fixing this. Could you add a test?

fordhurley:
I'm not very familiar with writing jasmine tests, but I gave it a shot. If someone more experienced could provide guidance, I'd be happy to expand this a bit. I would personally prefer a test that actually checks for uniqueness of all IDs after initializing the workspace with a mixture of deserialized and "fresh" editors. The test I wrote is more a test of the exact bug this PR fixes, than protection against future regressions. I confirmed that it fails before this fix and passes after, at least...

maxbrunsfeld:
The test looks fine to me. I'll merge this if CI passes.

maxbrunsfeld:
@fordhurley It looks like your new test if failing on [our CI](https://circleci.com/gh/atom/atom/6408?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

fordhurley:
Yup, it passed only when run in isolation (`atom --test spec/text-editor-spec.js`). I just took another stab at it.

maxbrunsfeld:
Thanks @fordhurley!

Issue #13953, opened 2017-03-07
Make control + arrow keys (move to beginning/end of word shortcut) not move to end of periods or commas


sn0wyfall:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

So, I couldn't find a package which did this.

Essentially, the `control + right/left arrow` (or `alt + right/left arrow` if you're on a Mac) allows you to go to the end or beginning of the word. I think that's well-known. However, for some reason, Atom works a bit differently.

This is what happens in Atom as you press the `alt + right arrow` shortcut in a line (the `|` is the cursor)

    |if 20,000

    if| 20,000

    if 20|,000

    if 20,|000

    if 20,000|

So, as you can see, the cursor stops at the comma. However, this is much different to how it works in other apps. I tested it in Pages, TextEdit, Safari, as well as a bunch of other third-party apps, and they all work like this:

    |if 20,000

    if| 20,000

    if 20,000|

Simple and easy to use.

So, why do I want Atom to change its behavior to the second way instead of using its current method? Because it makes text editing and going to other places within a line *extremely annoying.* You *always* have to keep pressing `alt + right` to go somewhere, and sometimes this problem makes navigating as slow as having to press the right arrow key on its own, which is a real bummer.

Is the comma the only character that causes this to happen? No, there is also `.` and `'`. One this about the single quote `'`. Additionally, there's another problem: `alt + backspace` (or `control + backspace`) also stops at these characters. That means that if you want to backspace the word `don't` in Atom, you need to press `alt + backspace` three times, whereas virtually in app other than Atom this shortcut only needs to be pressed once. You have the same problem with text selection via `alt + shift + arrow key` as well, making text editing in Atom a burden in general.

This is why I honestly think that this behavior should be changed.

### Versions

macOS 10.12.3

Atom    : 1.13.1
Electron: 1.3.13
Chrome  : 52.0.2743.82
[...]

Issue #13651, opened 2017-01-18
Adopt Sublime Text's handling of punctuation at the start of alt-left/right


daviddoran:
### Description

Atom alt-left/right and alt-shift-left/right (skip over or select word) should ignore certain punctuation characters in front of the cursor. This makes it easier to select parts of a JavaScript property chain, for example.

### Steps to Reproduce

Tab through this text in Atom vs. Sublime Text:

    this.props.horse.cart.battery.staple

Tab through this text in Atom vs. Sublime Text:

    this.props.horse.cart['battery'][staple]

Screencast showing Sublime Text vs. Atom behaviour:
https://www.youtube.com/watch?v=qovtU8MIygo&feature=youtu.be

**Expected behavior:**

In the first example when you start a selection before a period Sublime Text includes the period and continues to the next punctuation character. In the second example Sublime Text also skips over `['` at the beginning of the selection.

**Actual behavior:**

Atom stops at every punctuation character.

### Versions

Atom 1.13.0 x64 Mac

lee-dohm:
I'll bring this to the dev team. It is unlikely that we will replace the current behavior, though we may add new commands that behave in the way this specifies.

lee-dohm:
I reviewed this with the development team and they agree that it would be a nice option to have. On the other hand, this is not something that we will get the chance to work on any time soon. Because we treat the open issues list as our backlog of what we will be working on, I'm going to close this.

We would be open to accepting a well-crafted pull request that implements new commands for these movements, so I'll mark this with `help-wanted`.

a00958821:
Hello, im new here but i would like to help with this feature

PR #16457, opened 2017-12-28
Change SubWord selection to avoid selecting both sides of a dot "."

RmenaLopez:
### Requirements

Atom    : 1.23.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0

### Description of the Change

Atom's Subword (ctrl-left/right arrow) right now move to both sides of a dot. Here are examples of the current behavior. 

Current Atom's subWord selection moving forward (the "|" represents the cursor):
`|this.aMethod`
`this|.aMethod`
`this.|aMethod`
`this.a|Method`
`this.aMethod|`
Current Atom's SubWord selection moving backwards (the "|" represents the cursor):
`this.aMethod|`
`this.a|Method`
`this.|aMethod`
`this|.aMethod`
`|this.aMethod`

Now the same example with the proposed behavior:

Proposed Atom's subWord selection moving forward (the "|" represents the cursor):
`|this.aMethod`
`this|.aMethod`
`this.a|Method`
`this.aMethod|`
Porpoised Atom's SubWord selection moving backwards (the "|" represents the cursor):
`this.aMethod|`
`this.a|Method`
`this.|aMethod`
`|this.aMethod`

As it can be seen, the proposed behavior only changes the dot "." selection, everything else remains the same.

The current behavior creates an unnecessary need for an extra input to move across  subwords.
The current behavior also repeats functionality since the same behavior can be obtained by using just left/right arrow. Example (the "|" represents the cursor):
If the cursor is at the left of a dot and we are moving forward, like shown:
`this|.method`
Then the input alt-right arrow would produce the same output as simply right arrow:
`this.|method`

This also happens when selecting text by using shit-alt-arrow.

### Benefits

This changes makes moving trough subwords faster, but it also is a more natural way of doing it.

### Possible Drawbacks

Some user may struggle to accommodate to this change at  first.

### Applicable Issues

This issues asked for a change in behavior, however, this change does not supply every change requested in them.

https://github.com/atom/atom/issues/13651
https://github.com/atom/atom/issues/13953

### DIFF
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -1048,6 +1048,24 @@ describe('TextEditor', () => {
         expect(editor.getCursorBufferPosition()).toEqual([0, 1])
       })
 
+      it('stops at word and dot boundaries', () => {
+        editor.setText('sub.word \n')
+        editor.setCursorBufferPosition([0, 9])
+        editor.moveToPreviousSubwordBoundary()
+        expect(editor.getCursorBufferPosition()).toEqual([0, 8])
+
+        editor.moveToPreviousSubwordBoundary()
+        expect(editor.getCursorBufferPosition()).toEqual([0, 4])
+
+        editor.moveToPreviousSubwordBoundary()
+        expect(editor.getCursorBufferPosition()).toEqual([0, 0])
+
+        editor.setText(' word\n')
+        editor.setCursorBufferPosition([0, 3])
+        editor.moveToPreviousSubwordBoundary()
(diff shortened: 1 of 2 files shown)


lee-dohm:
Apologies for taking so long to look into this and thank you very much for the contribution.

The `help-wanted` issue specifically requests that new commands be implemented for this option, not to change the default behavior. Because of this, we won't be able to accept this PR. Thanks again for your submission.

PR #16458, opened 2017-12-28
Independent Atom instances (per $ATOM_HOME) v2

mortenpi:
An updated version of #15081 which was reverted in #16456.

The changes relative to the original PR, based on the discussion following https://github.com/atom/atom/pull/15081#issuecomment-354180722, are:

* The name of the socket file is now just `atom-#{atomInstanceDigest}.sock` with explicit version info removed. This way the socket file name will have a fixed length and should always fit within the limit.
* Atom version and platform architecture information goes into the digest.
* The digest is now the first 12 characters of the `base64` encoded hash (using the [URL and Filename Safe Alphabet](https://tools.ietf.org/html/rfc4648#section-5), so besides numbers and letters the digest may also contain `-_=`).

I cherry picked the original commit and added the updates as a separate commit. The PR should probably be squashed when merging.

cc @maxbrunsfeld

### DIFF
--- a/src/main-process/atom-application.coffee
+++ b/src/main-process/atom-application.coffee
@@ -8,6 +8,7 @@ FileRecoveryService = require './file-recovery-service'
 ipcHelpers = require '../ipc-helpers'
 {BrowserWindow, Menu, app, dialog, ipcMain, shell, screen} = require 'electron'
 {CompositeDisposable, Disposable} = require 'event-kit'
+crypto = require 'crypto'
 fs = require 'fs-plus'
 path = require 'path'
 os = require 'os'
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
> The digest is now the first 12 characters of the base64 encoded hash (using the URL and Filename Safe Alphabet, so besides numbers and letters the digest may also contain -_=).

Perfect. Thanks for following up with this. I saw the appveyor build failed it looks unrelated so I am re-running it.

tyleri:
Was this pull request ever reverted? I am using version 1.24.1 on Mac OS 10.13.3 (High Sierra), and it seems like any additional instances created after the first will use the ATOM_HOME environment of the first instance, rather than using the ATOM_HOME variable set in the context that it is launched.

mortenpi:
This was merged after 1.24 was released, so it will be in 1.25 (as far as I can tell). You could give the 1.25.0 beta a whirl -- it should have the expected behaviour.

PR #15183, opened 2017-08-07
Fix regression re: closing tabs via middle mouse click on Linux

jasonrudolph:
Fixes atom/tabs#461
Fixes atom/tabs#52

---

On Linux, when the user performs a middle-button mouse click, Chromium fires both a `mousedown` event *and* a `paste` event. This PR teaches the `TextEditorComponent` to ignore the `paste` event.

When the user performs a middle-mouse click on a tab, we first get the `mousedown` event. In response to that event, we [close the tab and attempt to prevent Chromium's default processing for the event](https://github.com/atom/tabs/blob/ce1d92e0abb669155caa178bb71166b9f16f329a/lib/tab-bar-view.coffee#L416-L418). This prevents Chromium's default processing for the *`mousedown`* event, but then Chromium also fires a *`paste`* event, and that event pastes the clipboard's current content into the newly-focused text editor. 🤦

#14987 removed some of the middle-button mouse click handling related to pasting on Linux. This PR restores that logic and the related tests. (See `TextEditorComponent::didMouseDownOnContent(event)`.) With that change in place, since Atom already has its own logic for handling pasting, we shouldn't need to handle browser `paste` events. By ignoring the browser `paste` events on Linux, this PR fixes atom/tabs#461.

### TODO

- [x] Fix atom/tabs#461
- [x] Add test for `TextEditorComponent::didPaste(event)`
- [x] Address TODO in `TextEditorComponent::didPaste(event)` (https://github.com/atom/atom/pull/15183#discussion_r131672090)

### Test plan

- [x] Test on Ubuntu:
  - [x] Verify fix for https://github.com/atom/tabs/issues/461
  - [x] Verify the previous fix for https://github.com/atom/atom/issues/8648 is still intact
- [x] Test on macOS: Verify that the ctrl-click behavior fixed in https://github.com/atom/atom/pull/14987 is still intact

PR #16484, opened 2018-01-03
Add "Verification Process" section to pull request template

jasonrudolph:
### Motivation

**As a contributor**, I often find it useful to explain the steps that I've taken to attempt to verify the results of my changes. For example, see the "Test Plan" section in https://github.com/atom/atom/pull/15183 and https://github.com/atom/teletype-server/pull/36#issuecomment-345392976. This serves two purposes:

1. It forces me to consider the scenarios that deserve to be tested. In writing out a test plan, I often think of things that I hadn't previously considered, and this sometimes reveals other cases to address before shipping.
2. It gives maintainers an opportunity to spot additional manual verification that is warranted. (As a contributor, my test plan is often a way of asking, "What else should I be testing that I don't know about? Please tell me. :pray:")

**As a maintainer**, I often want to know what steps a contributor has taken to manually verify the results of their changes. This helps me understand the level of rigor that they've already applied, and it can serve as a guide for how much scrutiny is warranted prior to merging their contribution.

### Description of the Change

With the motivation above in mind, this pull request adds a "Verification Process" section to `PULL_REQUEST_TEMPLATE.md`, prompting contributors to describe the approach they used to verify that the change has the desired effects.

---

/cc @lee-dohm @50Wliu @ungb @iolsen @as-cii @nathansobo since we've previously talked about adding something like this to the pull request template


nathansobo:
NICE

lee-dohm (approved):
Looks awesome 💖

PR #16491, opened 2018-01-04
Fix typo in protocol handler installer popup

tobiasvl:
### Requirements

### Description of the Change

Correct the typo "defaut" to the correct "default" in the atom:// URI protocol handler popup

### Alternate Designs

None

A prominent typo should not be one of the first things a new Atom user sees

### Benefits

Not having a typo in a prominent popup

### Possible Drawbacks

None

### Applicable Issues

None


50Wliu:
:100:

PR #16493, opened 2018-01-04
Remove unused package-lock.json files before building

lee-dohm:
### Description of the Change

`package-lock.json` files are generated by the build on certain configurations. When they are, they can sometimes create strange, hard-to-diagnose errors that prevent successive builds. This change deletes the `package-lock.json` files (that we don't use for Atom) before building so that people who aren't consistently building Atom don't get confused.

### Alternate Designs

Considered actually using the `package-lock.json` files but we're not ready to do that yet.

Because that's where builds happen.

### Benefits

Easier build experience for everyone :tada:

### Possible Drawbacks

We'll have to rip this out if/when we decide to use `package-lock.json` files.

### Applicable Issues

None known?

### DIFF
--- a/script/build
+++ b/script/build
@@ -58,6 +58,7 @@ const CONFIG = require('./config')
 let binariesPromise = Promise.resolve()
 
 if (!argv.existingBinaries) {
+  cleanPackageLock()
   checkChromedriverVersion()
   cleanOutputDirectory()
   copyAssets()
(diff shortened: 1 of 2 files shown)


hypersad:
You also could call it `remove-lockfiles`, I think that looks better and, as I know, it is their right name.

PR #16521, opened 2018-01-09
If protocol handler set to never, unregister it on Windows

damieng:
We should remove the protocol handler if never is set however the electron unregister does not appear to work so blow away the registry key instead.  This is only possible on Windows.

### DIFF
--- a/src/protocol-handler-installer.js
+++ b/src/protocol-handler-installer.js
@@ -11,10 +11,6 @@ class ProtocolHandlerInstaller {
     return ['win32', 'darwin'].includes(process.platform)
   }
 
-  isOldDefaultProtocolClient () {
-    return remote.app.isDefaultProtocolClient('atom', process.execPath, ['--uri-handler'])
-  }
-
   isDefaultProtocolClient () {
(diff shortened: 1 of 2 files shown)


src/protocol-handler-installer.js
```diff
@@ -25,32 +21,32 @@ class ProtocolHandlerInstaller {
...
-        default:
-          // Do nothing
-      }
+        }
+        break
+      case NEVER:
+        if (process.platform === 'win32') {
+          // Only win32 supports deregistration
+          const Registry = require('winreg')
+          const commandKey = new Registry({hive: 'HKCR', key: `\\atom`})
```
BinaryMuse: Will the protocol client definitely be the only thing in this location?

damieng: Yeah, all our other registry things live under HKCU (current user).  You can see them in WinShell.js

BinaryMuse:
@damieng Looks like `winreg` isn't compatible with snapshotting. It'll need to go into the ignore script.

BinaryMuse:
Merging with 1 failure, a known flake on appveyor.

PR #16526, opened 2018-01-10
Directly require clipboard from Electron

winstliu:
### Requirements

### Description of the Change

As far as I can tell the "safe-clipboard" workaround was for an Electron bug that was fixed somewhere around atom-shell 0.24.0.  Searching for "clipboard linux" on the electron issues list doesn't yield any crash results and the docs don't have anything Linux-specific, so I think that we should be good now.

### Alternate Designs

None.

This cleans up some old code present in core.

### Benefits

Requiring the clipboard is more straightforward.

### Possible Drawbacks

It _is_ possible that the crash still exists and I'm searching for the wrong thing.

### Verification Process

I currently don't have a Linux distribution set up, so I'm unable to verify that using the renderer clipboard does not crash Atom.

### Applicable Issues

Refs https://github.com/electron/electron/issues/1372


maxbrunsfeld:
I just tested this on my Ubuntu VM. Requiring clipboard directly from electron and calling `writeText` on it worked just fine from the renderer process.

maxbrunsfeld:
Thanks @50Wliu!

PR #16533, opened 2018-01-10
Teach AppVeyor to run installer task for master branch

jasonrudolph:
### Motivation

We'd like to add support for a "dev" release channel (in addition to the beta and stable release channels). The dev channel would represent the latest successful build on the master branch.

The primary goal is to make it easy to run a package's CI suite against the current Atom master branch to detect potential issues early, as opposed to only discovering issues after master branch is promoted to the next beta release. For example, if teletype's CI suite periodically ran teletype's tests against an Atom dev channel, it would have automatically exposed the issue in https://github.com/atom/teletype/issues/300, instead of us having to manually discover that issue. 😅 

### Making it happen

For macOS on Circle CI, and rpm and debian on Travis CI, we're already producing the necessary build artifacts to support a dev channel.

For Windows on AppVeyor, the installer task produces the build artifacts that we need ([x64 example](https://ci.appveyor.com/project/Atom/atom/build/9355/job/g0vwhhtj59nv283q/artifacts), [x86 example](https://ci.appveyor.com/project/Atom/atom/build/9355/job/wypnj96kr308iioo/artifacts)). As of https://github.com/atom/atom/pull/15208, we only run the installer task on release branches (e.g., `1.23-releases`). We skip the installer task on all other branches, including master. As a result, our builds on the master branch don't provide the necessary artifacts for a dev channel (e.g., [x64 example](https://ci.appveyor.com/project/Atom/atom/build/9371/job/egys9532ba9c59x3/artifacts), [x86 example](https://ci.appveyor.com/project/Atom/atom/build/9371/job/egys9532ba9c59x3/artifacts)).

In addition to running the installer task on release branches, **this pull request aims to update the AppVeyor build to also run the installer task on the master branch.**

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -46,7 +46,7 @@ build_script:
         ECHO Building on release branch - Creating production artifacts &&
         script\build.cmd --code-sign --compress-artifacts --create-windows-installer
       ) ELSE (
-        IF [%APPVEYOR_REPO_BRANCH%]==[master] (
+        IF [%APPVEYOR_REPO_BRANCH%]==[master] IF NOT DEFINED [%APPVEYOR_PULL_REQUEST_NUMBER%] (
           ECHO Building on master branch - Creating signed zips &&
           script\build.cmd --code-sign --compress-artifacts
         ) ELSE (
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,6 +41,8 @@ build_script:
   - CD %APPVEYOR_BUILD_FOLDER%
   - IF NOT EXIST C:\tmp MKDIR C:\tmp
   - SET SQUIRREL_TEMP=C:\tmp
+  - ECHO APPVEYOR_REPO_BRANCH is '%APPVEYOR_REPO_BRANCH'
+  - ECHO APPVEYOR_PULL_REQUEST_NUMBER is '%APPVEYOR_PULL_REQUEST_NUMBER%'
   - IF [%TASK%]==[installer] (
       IF [%APPVEYOR_REPO_BRANCH:~-9%]==[-releases] (
         ECHO Building on release branch - Creating production artifacts &&
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -50,8 +50,7 @@ build_script:
           ECHO Building on master branch - Creating signed zips &&
           script\build.cmd --code-sign --compress-artifacts
         ) ELSE (
-          ECHO Building on non-master branch - Creating unsigned zips &&
-          script\build.cmd --compress-artifacts
+          ECHO Skipping installer build for non-release/non-master branch
         )
       )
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -45,7 +45,11 @@ build_script:
       IF [%APPVEYOR_REPO_BRANCH:~-9%]==[-releases] (
         script\build.cmd --code-sign --compress-artifacts --create-windows-installer
       ) ELSE (
-        ECHO Skipping installer and Atom build on non-release branch
+        IF [%APPVEYOR_REPO_BRANCH%]==[master] (
+          script\build.cmd --code-sign --compress-artifacts --create-windows-installer
+        ) ELSE (
+          ECHO Skipping installer and Atom build on non-release branch
+        )
```
jasonrudolph: I suspect there's a way to streamline this `if/else` statement. If anyone has suggestions, please let me know. :bow:

The one constraint we have to keep in mind is that [this entire conditional](https://github.com/atom/atom/blob/058c13e25258c977f6a56ccdd35532441d855d7a/appveyor.yml#L44-L57) gets collapsed into a single line when it runs on AppVeyor.

daviwil: Using an intermediate variable could help remove the duplication between the two cases: https://stackoverflow.com/a/2143396

damieng:
I do not recommend enabling the installer build task on master. It at least doubles the time it takes to run on AppVeyor and causes long delays for devs wanting a green light on their PR's etc.

One option would be to build on a 'dev' or 'nightly' branch and have something fork at a predefined time each night.  Would keep the dev/nightly builds to 1 per day and prevent build backpressure.

damieng (approved):

jasonrudolph:
@damieng: Thanks for all your help. :zap::bow:

In [the build](https://ci.appveyor.com/project/Atom/atom/build/9385/job/x1mevvj9xvak2um7#L435) for 63d3a47ea85f5a126a6d8c21d370d468da482645, I noticed that the script echoed:

> Building on master branch - Creating signed zips 

I looked into why the script is was concluding that this PR is the master branch, and I committed 9188b98b8e9a4a6c8e258ced620e11c7e52da2cd which seems to resolve the issue. With that change in place, [the script now echoes](https://ci.appveyor.com/project/Atom/atom/build/9392/job/ncindkfjn9p5mbxw#L439):

> Building on non-master branch - Creating unsigned zips 

I think we're good to go now. I'll merge this today and verify that it creates a signed zip when it builds on the master branch.

jasonrudolph:
@damieng: I was already seeing the the build queue get backed up a bit, so I committed https://github.com/atom/atom/commit/992ebe352d01d68354a318f8ef64964f71464807 to restore the previous behavior that skips the `installer` build for non-master/non-release branches.

PR #16542, opened 2018-01-11
Preserve some TextEditor settings when language mode changes

### LINKED ISSUE #13829: Soft wrap settings reset when saving new file
### Prerequisites

### Description

If soft wrap is enabled on a new, unsaved file and you save that file, then soft wrap is disabled on that file.

### Steps to Reproduce

1. Open a new, untitled window.
2. Put in enough text so as to require horizontal scrolling, and enable the soft wrapping under "View" -> "Toggle Soft Wrap"
3. Save the file

**Expected behavior:**

I expect the file to save and retain the soft wrap settings.

**Actual behavior:** 

The file saves but soft wrap is reset, and you get a single long line of text.

**Reproduces how often:**

100% reproducible. Only happens when saving a *new* file - if you enable soft wrap again, make some changes in the file and save it again, then the soft wrap settings are retained.

### Versions

Atom:
```
Atom    : 1.14.2
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```
Apm:
```
apm  1.15.3
npm  3.10.5
node 4.4.5 x64
python 2.7.13
git 2.11.1
```
OS: OSX Sierra 10.12.3
```
$ uname -mprsv
Darwin 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64 i386
```

daviwil:
### Description of the Change

This change fixes #13829 which reports that the `softWrapped` setting of an untitled TextEditor is lost when the buffer is saved to a file.  This is caused by logic that updates TextEditor settings when the buffer's language mode changes.

The fix is to mark some editor settings as "preserved" so that they do not get change when switching between two language modes that do not have a different value for those preserved settings.

### Alternate Designs

@maxbrunsfeld and I looked into this together and tried to make TextBuffer not use `NullLanguageMode` as the default but instead a `text.plain` mode to avoid the apparent language mode change that caused the user's original issue.  It turned out that this solution would affect setting changes when going from an untitled buffer to a saved text buffer, so it wasn't sufficiently general.  This seemingly small change also had a pretty big impact on existing tests so it seemed too risky as a fix for such a narrow use case.

### Benefits

This solution allows us to easily mark a subsection of `TextEditor` settings to be preserved when the buffer's language mode changes.  This will cause less user confusion while still allowing buffers to have other relevant settings to be updated to reflect the new mode.

### Possible Drawbacks

The current algorithm for checking value changes is not as efficient as it could be.  I was about to try to change `TextEditorRegistry.textEditorParamsForScope` to use chained underscore collection methods but I didn't want to disrupt the current design any more before getting some feedback.

### Verification Process

Verified using @lee-dohm's excellent verification steps in his comment below.

### Applicable Issues

#13829

### DIFF
--- a/src/text-editor-registry.js
+++ b/src/text-editor-registry.js
@@ -1,3 +1,4 @@
+const _ = require('underscore-plus')
 const {Emitter, Disposable, CompositeDisposable} = require('event-kit')
 const TextEditor = require('./text-editor')
 const ScopeDescriptor = require('./scope-descriptor')
@@ -147,11 +148,11 @@ class TextEditorRegistry {
     }
     this.editorsWithMaintainedConfig.add(editor)
 
-    this.subscribeToSettingsForEditorScope(editor)
-    const grammarChangeSubscription = editor.onDidChangeGrammar(() => {
-      this.subscribeToSettingsForEditorScope(editor)
+    this.updateAndMonitorEditorSettings(editor)
+    const languageChangeSubscription = editor.buffer.onDidChangeLanguageMode((newLanguageMode, oldLanguageMode) => {
+      this.updateAndMonitorEditorSettings(editor, oldLanguageMode)
     })
-    this.subscriptions.add(grammarChangeSubscription)
+    this.subscriptions.add(languageChangeSubscription)
 
     const updateTabTypes = () => {
       const configOptions = {scope: editor.getRootScopeDescriptor()}
@@ -169,8 +170,8 @@ class TextEditorRegistry {
     return new Disposable(() => {
       this.editorsWithMaintainedConfig.delete(editor)
       tokenizeSubscription.dispose()
-      grammarChangeSubscription.dispose()
-      this.subscriptions.remove(grammarChangeSubscription)
+      languageChangeSubscription.dispose()
+      this.subscriptions.remove(languageChangeSubscription)
       this.subscriptions.remove(tokenizeSubscription)
     })
   }
@@ -214,13 +215,40 @@ class TextEditorRegistry {
     atom.grammars.autoAssignLanguageMode(editor.getBuffer())
   }
 
-  async subscribeToSettingsForEditorScope (editor) {
+  async updateAndMonitorEditorSettings (editor, oldLanguageMode) {
     await this.initialPackageActivationPromise
+    this.updateEditorSettingsForLanguageMode(editor, oldLanguageMode)
+    await this.subscribeToSettingsForEditorScope(editor)
+  }
 
+  updateEditorSettingsForLanguageMode (editor, oldLanguageMode) {
+    const newLanguageMode = editor.buffer.getLanguageMode()
+
+    if (oldLanguageMode) {
+      const newSettings = this.textEditorParamsForScope(newLanguageMode.rootScopeDescriptor)
+      const oldSettings = this.textEditorParamsForScope(oldLanguageMode.rootScopeDescriptor)
+
+      const updatedSettings = {}
+      for (const [, paramName] of EDITOR_PARAMS_BY_SETTING_KEY) {
+        // Update the setting only if it has changed between the two language
+        // modes.  This prevents user-modified settings in an editor (like
+        // 'softWrapped') from being reset when the language mode changes.
+        if (!_.isEqual(newSettings[paramName], oldSettings[paramName])) {
+          updatedSettings[paramName] = newSettings[paramName]
+        }
+      }
+
+      if (_.size(updatedSettings) > 0) {
+        editor.update(updatedSettings)
+      }
+    } else {
+      editor.update(this.textEditorParamsForScope(newLanguageMode.rootScopeDescriptor))
+    }
+  }
+
+  async subscribeToSettingsForEditorScope (editor) {
     const scopeDescriptor = editor.getRootScopeDescriptor()
     const scopeChain = scopeDescriptor.getScopeChain()
-
-    editor.update(this.textEditorParamsForScope(scopeDescriptor))
 
     if (!this.scopesWithConfigSubscriptions.has(scopeChain)) {
       this.scopesWithConfigSubscriptions.add(scopeChain)
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -154,6 +154,45 @@ describe('TextEditorRegistry', function () {
       expect(editor.getEncoding()).toBe('utf8')
     })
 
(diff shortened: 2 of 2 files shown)


daviwil:
Still need to add new tests for this, will do that after lunch.  I'd definitely appreciate some feedback on any performance concerns about the setting checking code!

src/text-editor-registry.js
```diff
@@ -1,19 +1,21 @@
...
 const TextEditor = require('./text-editor')
 const ScopeDescriptor = require('./scope-descriptor')
 
 const EDITOR_PARAMS_BY_SETTING_KEY = [
+  // Columns: settingsKey, paramName, preserveSetting
   ['core.fileEncoding', 'encoding'],
-  ['editor.atomicSoftTabs', 'atomicSoftTabs'],
-  ['editor.showInvisibles', 'showInvisibles'],
-  ['editor.tabLength', 'tabLength'],
+  ['editor.atomicSoftTabs', 'atomicSoftTabs', true],
```
daviwil: Happy to add/remove any "preserved" settings in this list

lee-dohm:
Given the following scenario:

1. Set the defaults to "show invisibles on" and "soft wrap off"
1. Set the language-specific settings for GitHub-Flavored Markdown to "show invisibles off" and "soft wrap on"
1. Open a new file using <kbd>Cmd+N</kbd>
1. Enter some text
1. Change the language to "GitHub-Flavored Markdown" using the language selector

The expected behavior (after this change is merged) is for invisibles to disappear and soft wrap to be enabled?

And given the following scenario:

1. Set the defaults to "show invisibles on" and "soft wrap off"
1. Set the language-specific settings for GitHub-Flavored Markdown to "show invisibles off" and "soft wrap on"
1. Open a new file using <kbd>Cmd+N</kbd>
1. Enter some text
1. **Set soft wrap to on**
1. Enter some more text
1. Change the language to "JavaScript" using the language selector

The expected behavior (after this change is merged) is for invisibles to stay invisible and soft wrap to continue to be enabled?

src/text-editor-registry.js
```diff
@@ -214,14 +218,40 @@ class TextEditorRegistry {
...
+      const newSettings = this.textEditorParamsForScope(newLanguageMode.scopeDescriptor);
+      const oldSettings = this.textEditorParamsForScope(oldLanguageMode.scopeDescriptor);
+
+      let updatedSettings = {}
+      for (const [settingsKey, paramName, preserveSetting] of EDITOR_PARAMS_BY_SETTING_KEY) {
+        // Log the updated setting if it doesn't need to be preserved or if
+        // it has changed between the two language modes.  This prevents user-
+        // modified settings in an editor (like 'softWrapped') from being reset
+        // when the language mode changes.
+        if (!preserveSetting || !_.isEqual(newSettings[paramName], oldSettings[paramName])) {
```
maxbrunsfeld: Can you explain why the `preserveSetting` flag is needed? I would think that the comparison between `oldSettings` and `newSettings` would be enough to fix the problem.

src/text-editor-registry.js
```diff
@@ -214,14 +218,40 @@ class TextEditorRegistry {
     atom.grammars.autoAssignLanguageMode(editor.getBuffer())
   }
 
-  async subscribeToSettingsForEditorScope (editor) {
+  updateEditorSettingsForLanguageMode(editor, oldLanguageMode) {
+    const newLanguageMode = editor.buffer.getLanguageMode()
+
+    if (oldLanguageMode) {
```
maxbrunsfeld: I'm surprised that `oldLanguageMode` would ever be null. When does that happen?

src/text-editor-registry.js
```diff
@@ -214,14 +218,40 @@ class TextEditorRegistry {
...
+      const newSettings = this.textEditorParamsForScope(newLanguageMode.scopeDescriptor);
+      const oldSettings = this.textEditorParamsForScope(oldLanguageMode.scopeDescriptor);
+
+      let updatedSettings = {}
```
maxbrunsfeld: Minor nitpick - you could use `const` here since `updatedSettings` is never reassigned.

daviwil: :+1: will fix!

src/text-editor-registry.js
daviwil: In [this call](https://github.com/atom/atom/pull/16542/files#diff-c07ba0f74b082a7f8c36ffec1a1ef1b8R152) that establishes the settings of the editor in `maintainConfig`.  Maybe there's a cleaner way to accomplish the same thing?

src/text-editor-registry.js
daviwil: Basically just trying to make it so that some settings always got overwritten if there's not a chance the user could tweak them.  However, if you think it's not necessary I'm more than happy to take it out!

daviwil:
Tests have been added!  I'll probably squash the changes (minus the text-buffer bump) once I've taken all future feedback into account.

@lee-dohm thanks a lot for the verification steps!  Your assumptions for the results are correct.  I've verified that these steps give the expected outcome after my changes.

daviwil:
CircleCI tests failed, anyone know what causes the error `Focused spec is running on CI`?

```
TextEditorRegistry
  .maintainConfig(editor)
    it updates editor settings that have changed between previous and current language modes
      Error: Focused spec is running on CI
        at window.(anonymous function) (/Users/distiller/atom/out/app/spec/jasmine-test-runner.coffee:56:15)
        at module.exports.(anonymous function) (/Users/distiller/atom/spec/async-spec-helpers.js:28:17)
        at jasmine.Suite.<anonymous> (/Users/distiller/atom/spec/text-editor-registry-spec.js:177:5)
        at jasmine.Suite.<anonymous> (/Users/distiller/atom/spec/text-editor-registry-spec.js:77:3)
        at Object.<anonymous> (/Users/distiller/atom/spec/text-editor-registry-spec.js:7:1)
```

50Wliu:
https://github.com/atom/atom/pull/16542/files#diff-9e847bd17f963014b1ee47e59e7b32a6R177
un-fit it :)

daviwil:
Aaaargh, this is what I get for working on it for too many hours ;) Thanks for catching that!

daviwil:
CI tests are passing reliably so it's ready for final review/feedback!

lee-dohm (approved):
Looks good to me 👍

maxbrunsfeld (approved):
⚡️ Tests look perfect. 🚢

daviwil:
Thanks a lot for all the help!

Issue #14803, opened 2017-06-16
Can't scroll the editor when the mouse is at the right most edge of the screen


segevfiner:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

I can't scroll the editor when the mouse is at the right most edge of the screen. Also happens when I close the right docks.

### Steps to Reproduce

1. Open Atom in full screen with a long enough file so that there will be a vertical scroll bar.
2. Move your mouse to the very right most edge of the screen and try to scroll by clicking the scroll bar. It doesn't work.
3. Move it slightly to the left and you can click and scroll.

**Expected behavior:** I can click the scroll bar even when the mouse is at the right most edge of the screen.

**Actual behavior:** I can't scroll. I have to move the mouse slightly to the left.

**Reproduces how often:** Always

### Versions

```
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

```
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.1.windows.2
visual studio 2015
```

Windows 10.0.15063 x64

### Additional Information

![atom-1_18-scroll-bug](https://user-images.githubusercontent.com/24731903/27223712-5ce3f5d0-529a-11e7-8696-a891021d10fd.gif)

rsese:
Thanks for the report!  I was able to reproduce in Windows 10 and macOS 10.12.5 with Atom 1.18.0.  h/t @lee-dohm for finding https://github.com/atom/one-dark-ui/issues/181 which sounds like the same issue.  

It's in the One Dark repo though and I was able to reproduce with multiple UI themes (for One Dark I disabled the dock toggle buttons too) - can you confirm which UI theme you're using @segevfiner and if you can reproduce with multiple UI themes?

segevfiner:
[...]

PR #16546, opened 2018-01-12
Programmatically detect when mouse approaches the edge of a dock

as-cii:
### Description of the Change

Previously, we would assign dock elements a minimum width/height of 2 pixels so that we could detect when the mouse approached the edge of a hidden dock in order to show the chevron buttons. This, however, was causing confusion for users, who expected that extra space to be clickable in order to scroll editors located in the center dock.

With this pull request we will instead register a global `mousemove` event on the window right when attaching the workspace element to the DOM (the event handler is debounced, so this shouldn't have any performance consequence). Then, when the mouse moves, we will programmatically detect when it approaches to the edge of a dock and show the chevron button accordingly. This allows us to remove the `min-width` property from the dock container element, which eliminates the confusing behavior described above.

### Alternate Designs

None.

### Possible Drawbacks

This pull request registers a global mouse move event handler and it could be problematic if we did a lot of work inside of it. However, this handler is debounced and runs only after 100ms since the mouse has stopped moving.

### Verification Process

See https://github.com/atom/atom/issues/14803.

### Applicable Issues

Fixes https://github.com/atom/atom/issues/14803

/cc: @simurai @maxbrunsfeld @matthewwithanm for your previous work on docks.

### DIFF
--- a/spec/dock-spec.js
+++ b/spec/dock-spec.js
@@ -201,11 +201,7 @@ describe('Dock', () => {
         const dockElement = atom.workspace.getBottomDock().getElement()
         dockElement.querySelector('.atom-dock-resize-handle').dispatchEvent(new MouseEvent('mousedown', {detail: 2}))
         expect(dockElement.offsetHeight).toBe(0)
-
-        // There should still be a hoverable, absolutely-positioned element so users can reveal the
-        // toggle affordance even when fullscreened.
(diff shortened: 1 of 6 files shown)


as-cii:
Thanks for testing this, @50Wliu! 💯

PR #15498, opened 2017-08-30
Refactor install script for linux (install dev channel parallel, use icon themes)

MoritzKn:
### Description of the Change

* Some refactoring on the install script (mainly on the linux part) (1.-3. commit).
* Change the install script to create a separate installation for the dev channel (`atom-dev`) like it is already done for the beta channel (`atom-beta`) (4. commit).
* Change the way icons are handled in the xdg desktop file. This replaces the absolute path used previously with a icon name. The name will then be interpreted by the desktop environment, which in turn selects the icon from a icon theme. This is also how the desktop files in the dep and rpm packages already work. (5. commit).
* Change the install script to copy the application icons (in all resolutions) to the default icon theme. This is required to make to point above work when no alternative icon theme is used or the alternative icon theme does not provide an icon. (5. commit).

### Alternate Designs

None

### Why Should This Be In Core?

* Refactors core code.
* The beta channel is already installed in parallel so the dev channel should be too.
* The new way desktop files are handled is more "linuxy" and preferable as default.

### Benefits

* Cleaner code
* Allows all channels to be installed in parallel.
* Users are able to use alternative icon themes with atom; makes atom more hackable.
* The desktop environment gets to choose between different resolutions.
* Aligns the way the manually installed version behaves with the what is installed from a package.

### Possible Drawbacks

None known

### Applicable Issues

* Perhaps fix #12993 (icon scaling issue).
* Fixes icon behavior mentioned in #14480 (wrong icon path). This is fixed with the 2. commit and with the alternative design from the 5. commit.

PR #16548, opened 2018-01-13
Link spec for finding icon dir in install script

MoritzKn:
This is a follow up of #15498

### Description of the Change

Adds a comment that links to the relevant specification for a function.

### Alternate Designs

None

The documented function is in core

### Benefits

Future changes are more likely to be in line with the spec too

### Possible Drawbacks

None

### Verification Process

None

### Applicable Issues

None


daviwil (approved):
Looks good, thanks a lot for adding that!

PR #16593, opened 2018-01-18
decaffeinate config.coffee and config-spec.coffee

philipfweiss:
### Description of the Change

Used the decaffeinate utility to update the file from coffeescript to js. Manually went through the file to update many (but not all) of the bad idioms used by decaffeinate (needed to resolve coffeescript ambiguity). For example, removed the usage of __ guard __ functions, IIFEs, initClass() methods, and some unnecessary returns. 

This PR is part of a larger effort of migrating atom core away from Coffeescript. The changes introduce no new functionality, only update content already within core. 

### Benefits

This PR is part of a larger effort of migrating atom core away from Coffeescript.

### Possible Drawbacks

Coffeescript has implicit returns, and therefore the decaffeinate utility does not know when a return is necessary or not. Because of this, the decaffeinated code had returns in every function. I removed some of those returns (such as in private functions where nowhere else was using the returned result), but for the most part I kept these returns in. 

The reason I did this is because I didn't want to break the contract between atom.config and packages that rely on it. I did not want to break the functionality of packages that for some reason use the return value of a private method on atom.config, even if that value was not intentionally returned. 

### Verification Process

- How did you verify that all new functionality works as expected?
There is no new functionality. 

- How did you verify that all changed functionality works as expected?
There is no new changed functionality, except for the removal of return values on some private methods. 

- How did you verify that the change has not introduced any regressions?
I ran the test suite, played around with a dev version of atom using this config file. Overall the only regressions this PR should be able to cause are from functions that relied on the implicit return value of the changed private Coffeescript functions.


src/config.js
```diff
@@ -0,0 +1,1603 @@
+/** @babel */
+
+/*
+ * decaffeinate suggestions:
+ * DS101: Remove unnecessary use of Array.from
```
philipfweiss: Removed some instances of Array.from but mostly kept them in to emulate the behaviour the Coffeescript code.

src/config.js
```diff
@@ -0,0 +1,1603 @@
...
+ * DS207: Consider shorter variations of null checks
```
philipfweiss: Feedback on null checks appreciated.

src/config.js
```diff
@@ -0,0 +1,1603 @@
+/** @babel */
```
50Wliu: We try not to use Babel anymore.  Looks like the only thing you have to change is how you're exporting the class.

src/config.js
```diff
@@ -0,0 +1,1603 @@
...
+      }
+    }
+
+    if (error != null) { throw error }
+    return value
+  }
+
+    // Created during initialization, available as `atom.config`
+  constructor (param) {
+    if (param == null) { param = {} }
```
50Wliu: Looks like this could be set as a default value instead.

src/config.js
```diff
@@ -0,0 +1,1603 @@
...
+  set () {
+    let [keyPath, value, options] = Array.from(arguments)
+
+    if (!this.settingsLoaded) {
+      this.pendingOperations.push(() => this.set(keyPath, value, options))
+    }
+
+    const scopeSelector = options != null ? options.scopeSelector : undefined
+    let source = options != null ? options.source : undefined
+    const shouldSave = (options != null ? options.save : undefined) != null ? (options != null ? options.save : undefined) : true
```
50Wliu: It looks like `options != null ? options.save : undefined` is repeated twice here?

winstliu (commented):
Looks like you have some indentation issues with most of the comments and some of the multiline method calls using 4-space indents while the rest of the code uses two spaces.

src/config.js
philipfweiss: Fixed!

src/config.js
philipfweiss: Fixed!

src/config.js
philipfweiss: Fixed!

philipfweiss:
@50Wliu, could you give me an example of a place where my multiline method calls are indented improperly?

src/config.js
```diff
@@ -0,0 +1,1602 @@
+
+/*
+ * decaffeinate suggestions:
+ * DS101: Remove unnecessary use of Array.from
+ * DS207: Consider shorter variations of null checks
+ * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
+ */
```
matthewwithanm: don't forget to remove this block (:

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+  static executeSchemaEnforcers (keyPath, value, schema) {
+    let error = null
+    let types = schema.type
+    if (!Array.isArray(types)) { types = [types] }
+    for (let type of types) {
+      try {
+        const enforcerFunctions = schemaEnforcers[type].concat(schemaEnforcers['*'])
+        for (let enforcer of Array.from(enforcerFunctions)) {
+            // At some point in one's life, one must call upon an enforcer.
```
matthewwithanm: dedent

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    this.savePending = false
+    this.configFileHasErrors = false
+    this.transactDepth = 0
+    this.pendingOperations = []
+    this.legacyScopeAliases = {}
+
+    this.requestLoad = _.debounce(() => {
+      this.loadUserConfig()
+    }
+      , 100)
```
matthewwithanm: the indentation here is pretty weird. maybe this?

```js
_.debounce(
  () => { this.loadUserConfig() },
  100
)
```

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+    const debouncedSave = _.debounce(() => {
+      this.savePending = false
+      this.save()
+    }
+      , 100)
```
matthewwithanm: here too

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  }
+
+  endTransaction () {
+    this.transactDepth--
+    this.emitChangeEvent()
+  }
+
+  pushAtKeyPath (keyPath, value) {
+    let left
+    const arrayValue = (left = this.get(keyPath)) != null ? left : []
```
matthewwithanm: i think this would be better like this:

```js
const left = this.get(keyPath)
const arrayValue = left == null ? [] : left
```

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    const result = arrayValue.push(value)
+    this.set(keyPath, arrayValue)
+    return result
+  }
+
+  unshiftAtKeyPath (keyPath, value) {
+    let left
+    const arrayValue = (left = this.get(keyPath)) != null ? left : []
```
matthewwithanm: same here

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    const result = arrayValue.unshift(value)
+    this.set(keyPath, arrayValue)
+    return result
+  }
+
+  removeAtKeyPath (keyPath, value) {
+    let left
+    const arrayValue = (left = this.get(keyPath)) != null ? left : []
```
matthewwithanm: and here (:

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+      throw new Error(`Error loading schema for ${keyPath}: schemas can only be objects!`)
+    }
+
+    if (schema.type == null) {
+      throw new Error(`Error loading schema for ${keyPath}: schema objects must have a type attribute`)
+    }
+
+    let rootSchema = this.schema
+    if (keyPath) {
+      for (let key of Array.from(splitKeyPath(keyPath))) {
```
matthewwithanm: is this `Array.from()` needed?

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    const onConfigDirFile = sourcePath => {
+      const relativePath = sourcePath.substring(templateConfigDirPath.length + 1)
+      const destinationPath = path.join(this.configDirPath, relativePath)
+      return queue.push({sourcePath, destinationPath})
+    }
+    return fs.traverseTree(templateConfigDirPath, onConfigDirFile, path => true, () => {})
+  }
+
+  loadUserConfig () {
+    let error
```
matthewwithanm: are the errors here actually being used outside their catch blocks? i think this can be removed and you can just use `error` instead of (`error1` and `error2`) in the catch blocks below.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+  observeUserConfig () {
+    if (this.shouldNotAccessFileSystem()) { return }
+
+    try {
+      if (this.watchSubscriptionPromise == null) {
+        this.watchSubscriptionPromise = watchPath(this.configFilePath, {}, events => {
+          return (() => {
+            const result = []
+            for (let {action} of Array.from(events)) {
```
matthewwithanm: another `Array.from()` that seems unnecessary.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+  unobserveUserConfig () {
+    if (this.watchSubscriptionPromise != null) {
+      this.watchSubscriptionPromise.then(watcher => watcher != null ? watcher.dispose() : undefined)
+    }
+    this.watchSubscriptionPromise = null
+  }
+
+  notifyFailure (errorMessage, detail) {
+    return (this.notificationManager != null ? this.notificationManager.addError(errorMessage, {detail, dismissable: true}) : undefined)
```
matthewwithanm: i think this would be clearer like this:

```js
if (this.notificationManager == null) return
return this.notificationManager.addError(errorMessage, {detail, dismissable: true})
```

and honestly i doubt that the value even has to be returned.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+      this.transact(() => {
+        return (() => {
+          const result = []
+          for (let key in defaults) {
+            const childValue = defaults[key]
+            if (!defaults.hasOwnProperty(key)) { continue }
+            result.push(this.setDefaults(keys.concat([key]).join('.'), childValue))
+          }
+          return result
+        })()
```
matthewwithanm: get rid of this IIFE?

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  /*
+  Section: Private Scoped Settings
+  */
+
+  priorityForSource (source) {
+    if (source === this.getUserConfigPath()) {
+      return 1000
+    } else {
+      return 0
+    }
```
matthewwithanm: seems like a good case for ternary

```js
return source === this.getUserConfigPath() ? 1000 : 0
```

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  }
+
+  emitChangeEvent () {
+    if (!(this.transactDepth > 0)) { return this.emitter.emit('did-change') }
```
matthewwithanm: looks like a decaffeinated `unless`? assuming `transactDepth` is always a number, you can remove the negation and do `<=`.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  getRawScopedValue (scopeDescriptor, keyPath, options) {
+    let legacyScopeDescriptor
+    scopeDescriptor = ScopeDescriptor.fromObject(scopeDescriptor)
+    const result = this.scopedSettingsStore.getPropertyValue(
+        scopeDescriptor.getScopeChain(),
+        keyPath,
+        options
+      )
+
+    legacyScopeDescriptor = this.getLegacyScopeDescriptor(scopeDescriptor)
```
matthewwithanm: can this be a `const` instead?

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  }
+
+  onDidChangeScopedKeyPath (scope, keyPath, callback) {
+    let oldValue = this.get(keyPath, {scope})
+    return this.emitter.on('did-change', () => {
+      const newValue = this.get(keyPath, {scope})
+      if (!_.isEqual(oldValue, newValue)) {
+        const event = {oldValue, newValue}
+        oldValue = newValue
+        return callback(event)
```
matthewwithanm: i don't think that event-kit's emitter actually cares about the callback return value.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+  getLegacyScopeDescriptor (scopeDescriptor) {
+    const legacyAlias = this.legacyScopeAliases[scopeDescriptor.scopes[0]]
+    if (legacyAlias) {
+      const scopes = scopeDescriptor.scopes.slice()
+      scopes[0] = legacyAlias
+      return new ScopeDescriptor({scopes})
+    }
+  }
+  };
```
matthewwithanm: did indenting get off somewhere?

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    }
+  },
+
+  'array': {
+    coerce (keyPath, value, schema) {
+      if (!Array.isArray(value)) { throw new Error(`Validation failed at ${keyPath}, ${JSON.stringify(value)} must be an array`) }
+      const itemSchema = schema.items
+      if (itemSchema != null) {
+        const newValue = []
+        for (let item of Array.from(value)) {
```
matthewwithanm: the check on 1513 verifies that this is an array so we don't need the `Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+
+      if (Array.isArray(possibleValues)) {
+        possibleValues = possibleValues.map(value => {
+          if (value.hasOwnProperty('value')) { return value.value } else { return value }
+        })
+      }
+
+      if ((possibleValues == null) || !Array.isArray(possibleValues) || !possibleValues.length) { return value }
+
+      for (let possibleValue of Array.from(possibleValues)) {
```
matthewwithanm: we already know this is an array (ln 1562) so we don't need the `Array.from()` here

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    }
+  }
+})
+
+let isPlainObject = value => _.isObject(value) && !_.isArray(value) && !_.isFunction(value) && !_.isString(value) && !(value instanceof Color)
+
+let sortObject = value => {
+  if (!isPlainObject(value)) { return value }
+  const result = {}
+  for (let key of Array.from(Object.keys(value).sort())) {
```
matthewwithanm: `Object.keys().sort()` returns an array so we don't need the `Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  }
+
+  static executeSchemaEnforcers (keyPath, value, schema) {
+    let error = null
+    let types = schema.type
+    if (!Array.isArray(types)) { types = [types] }
+    for (let type of types) {
+      try {
+        const enforcerFunctions = schemaEnforcers[type].concat(schemaEnforcers['*'])
+        for (let enforcer of Array.from(enforcerFunctions)) {
```
matthewwithanm: `enforcerFunctions` is the result of `.concat()` so we don't need `Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  //     for more information.
+  // * `callback` {Function} to call when the value of the key changes.
+  //   * `value` the new value of the key
+  //
+  // Returns a {Disposable} with the following keys on which you can call
+  // `.dispose()` to unsubscribe.
+  observe () {
+    let callback, keyPath, options, scopeDescriptor
+    if (arguments.length === 2) {
+      [keyPath, callback] = Array.from(arguments)
```
matthewwithanm: you don't need `Array.from()` here

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    } else if ((arguments.length === 3) && (_.isString(arguments[0]) && _.isObject(arguments[1]))) {
+      [keyPath, options, callback] = Array.from(arguments)
```
matthewwithanm: …or here

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  //   * `event` {Object}
+  //     * `newValue` the new value of the key
+  //     * `oldValue` the prior value of the key.
+  //
+  // Returns a {Disposable} with the following keys on which you can call
+  // `.dispose()` to unsubscribe.
+  onDidChange () {
+    let callback, keyPath, scopeDescriptor
+    if (arguments.length === 1) {
+      [callback] = Array.from(arguments)
```
matthewwithanm: you don't need `Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    } else if (arguments.length === 2) {
+      [keyPath, callback] = Array.from(arguments)
```
matthewwithanm: -`Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    } else {
+      let options;
+      [keyPath, options, callback] = Array.from(arguments)
```
matthewwithanm: -`Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  //     See [the scopes docs](http://flight-manual.atom.io/behind-atom/sections/scoped-settings-scopes-and-scope-descriptors/)
+  //     for more information.
+  //
+  // Returns the value from Atom's default settings, the user's configuration
+  // file in the type specified by the configuration schema.
+  get () {
+    let keyPath, options, scope
+    if (arguments.length > 1) {
+      if ((typeof arguments[0] === 'string') || (arguments[0] == null)) {
+        [keyPath, options] = Array.from(arguments);
```
matthewwithanm: -`Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+        ({scope} = options)
+      }
+    } else {
+      [keyPath] = Array.from(arguments)
```
matthewwithanm: -`Array.from()`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  //     See [the scopes docs](http://flight-manual.atom.io/behind-atom/sections/scoped-settings-scopes-and-scope-descriptors/)
+  //     for more information.
+  //   * `source` (optional) {String} The name of a file with which the setting
+  //     is associated. Defaults to the user's config file.
+  //
+  // Returns a {Boolean}
+  // * `true` if the value was set.
+  // * `false` if the value was not able to be coerced to the type specified in the setting's schema.
+  set () {
+    let [keyPath, value, options = {}] = Array.from(arguments)
```
matthewwithanm: -`Array.from()`

also, can this be `const`?

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  //
+  // * `keyPath` The {String} name of the key.
+  //
+  // Returns an {Object} eg. `{type: 'integer', default: 23, minimum: 1}`.
+  // Returns `null` when the keyPath has no schema specified, but is accessible
+  // from the root schema.
+  getSchema (keyPath) {
+    const keys = splitKeyPath(keyPath)
+    let { schema } = this
+    for (let key of Array.from(keys)) {
```
matthewwithanm: i think `splitKeyPath()` returns an array right? probably don't need this`Array.from()` either.

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+  // Returns a {Promise} that is either resolved or rejected according to the
+  // `{Promise}` returned by `callback`. If `callback` throws an error, a
+  // rejected {Promise} will be returned instead.
+  transactAsync (callback) {
+    let endTransaction
+    this.beginTransaction()
+    try {
+      endTransaction = fn => (...args) => {
+        this.endTransaction()
+        return fn(...Array.from(args || []))
```
matthewwithanm: `args` is always going to be an array here (and therefore non-falsey). this can just be `fn(...args)`

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+      delete scopedSettings['*']
+      this.resetUserScopedSettings(scopedSettings)
+    }
+
+    return this.transact(() => {
+      this.settings = {}
+      this.settingsLoaded = true
+      for (let key in newSettings) { const value = newSettings[key]; this.set(key, value, {save: false}) }
+      if (this.pendingOperations.length) {
+        for (let op of Array.from(this.pendingOperations)) { op() }
```
matthewwithanm: i think this `Array.from()` isn't needed either

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    }
+  }
+
+  deepDefaults (target) {
+    let result = target
+    let i = 0
+    while (++i < arguments.length) {
+      const object = arguments[i]
+      if (isPlainObject(result) && isPlainObject(object)) {
+        for (let key of Array.from(Object.keys(object))) {
```
matthewwithanm: -`Array.from()`

matthewwithanm (dismissed):

src/config.js
philipfweiss: I saw in other decaffeinated files that people kept these in until they were addressed! Since there are still a lot of array.froms and null checks I left it in. 

I will fix those issues though and then delete this block!

src/config.js
YurySolovyov: I'd even say rest params would be even better here

src/config.js
```diff
@@ -0,0 +1,1602 @@
...
+    if (source == null) { source = this.getUserConfigPath() }
+
+    if (scopeSelector != null) {
+      if (keyPath != null) {
+        let settings = this.scopedSettingsStore.propertiesForSourceAndSelector(source, scopeSelector)
+        if (getValueAtKeyPath(settings, keyPath) != null) {
+          this.scopedSettingsStore.removePropertiesForSourceAndSelector(source, scopeSelector)
+          setValueAtKeyPath(settings, keyPath, undefined)
+          settings = withoutEmptyObjects(settings)
+          if (settings != null) { this.set(null, settings, {scopeSelector, source, priority: this.priorityForSource(source)}) }
```
YurySolovyov: Seems like a long line, this `if` is begging for breaking it down to separate lines

src/config.js
philipfweiss: Hmm it actually looks like these are being set and so I think I have to use `let`.

src/config.js
philipfweiss: Sadly this conflicts with Standard.js :( I'll do something sensible, though.

src/config.js
philipfweiss: ^

src/config.js
philipfweiss: See above ^

src/config.js
```diff
@@ -0,0 +1,1588 @@
...
+    const settingsBySelector = {}
+    settingsBySelector[selector] = value
+    this.scopedSettingsStore.addProperties(source, settingsBySelector, {priority: this.priorityForSource(source)})
+    return this.emitChangeEvent()
+  }
+
+  getRawScopedValue (scopeDescriptor, keyPath, options) {
+    scopeDescriptor = ScopeDescriptor.fromObject(scopeDescriptor)
+    const result = this.scopedSettingsStore.getPropertyValue(
+        scopeDescriptor.getScopeChain(),
```
50Wliu: I think here and the one below this one were the indentation issues I was talking about.

maxbrunsfeld:
Alright, great work @philipfweiss, and thanks so much for the careful reviews @matthewwithanm, @50Wliu, @YurySolovyov!

maxbrunsfeld:
By the way @philipfweiss, congratulations on getting us our JavaScript percentage up to :nine: :two: ❗️

PR #16594, opened 2018-01-19
Stop using Babel in Atom Core (benchmarks, exports, src)

as-cii:
### Description of the Change

This pull request stops transpiling (via Babel) JavaScript files located in the following folders:

* `benchmarks`
* `exports`
* `src`

Please, note that we still transpile core packages. Eventually, it might be worth transitioning away from Babel for those as well.

### Alternate Designs

None.

Core is where transpilation and packaging takes place.

### Benefits

* Reduced complexity of the build, which now only transpiles babel files found in core packages.
* Smaller footprint for JS files, thanks to removing source maps.
* Easier debugging, thanks to removing Babel async generators.

### Possible Drawbacks

Unclear, but theoretically none.

### Verification Process

* Run tests and see that they pass.
* Run Atom and see that no syntax error occurs.
* Snapshots are generated correctly.

/cc: @atom/maintainers

### DIFF
--- a/exports/atom.js
+++ b/exports/atom.js
@@ -1,13 +1,12 @@
-/** @babel */
-
-import TextBuffer, {Point, Range} from 'text-buffer'
-import {File, Directory} from 'pathwatcher'
-import {Emitter, Disposable, CompositeDisposable} from 'event-kit'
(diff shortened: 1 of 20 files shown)


nathansobo:
🎉

PR #16595, opened 2018-01-19
Don't break subpixel AA when cursor is at the end of longest line

as-cii:
### Description of the Change

With the Electron upgrade, something changed in the way characters are rendered/measured, and that was causing subpixel anti-aliasing to stop working when cursors were at the end of the longest line.

before

Every cursor has a width that is calculated in the following way:

* If there's a character after the cursor, the width corresponds to the width of such character.
* Otherwise, the width equals to the "base character width" measured on a dummy line.

In the latter case, even if we were setting the width of the content container to account for the width of such cursor, some rounding problem was causing the cursor to be able to escape the container and thus break subpixel anti-aliasing.

With this pull request, instead of rounding the value we assign to the container width, we will always ceil it. This ensures that cursors are always strictly contained within the parent element and resolves the subpixel anti-aliasing issue.

after

### Alternate Designs

None.

### Benefits

Fixes a regression in the subpixel anti-aliasing rendering code path that was causing text to look blurry when cursor was at the end of the longest line.

### Possible Drawbacks

None.

### Verification Process

See previously attached GIFs.

### Applicable Issues

It doesn't seem like this problem has been reported anywhere. /cc: @Ben3eeE @ungb @rsese in case you are aware of any recent report about this. If so, feel free to edit the body of this pull request to automatically close any relevant issue once this pull request gets merged.

/cc: @nathansobo

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -104,7 +104,7 @@ describe('TextEditorComponent', () => {
 
       {
         expect(editor.getApproximateLongestScreenRow()).toBe(3)
-        const expectedWidth = Math.round(
+        const expectedWidth = Math.ceil(
           component.pixelPositionForScreenPosition(Point(3, Infinity)).left +
           component.getBaseCharacterWidth()
         )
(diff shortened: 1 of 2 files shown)


as-cii:
Yeah that's strange. I wonder if we should round up to the nearest physical sub pixel instead of simply using `Math.ceil`. 

Can you test replacing the call to `Math.ceil` in this diff  with `ceilToPhysicalPixelBoundary` and report what you see? Thanks!

50Wliu:
@as-cii My intent was to file an actual issue about this a few days ago but when I went to re-reproduce just to be safe I found I couldn't.  If it happens again I'll test the change you suggested.

PR #16619, opened 2018-01-23
Update marked to guard against regex DoS

jaysonvirissimo:
### Description of the Change
This bumps the version of the `marked` package, since the current version contains a security flaw.

This should be in core, because having security vulnerabilities in core is bad?

### Benefits
It fixes [this](https://nodesecurity.io/advisories/531).

### Possible Drawbacks
None, that I'm aware of.


iolsen:
Thanks!

PR #16628, opened 2018-01-25
Perform all config file IO in the main process

### LINKED ISSUE #14909: Settings lost

maxbrunsfeld:
### Motivation

When your config settings are updated from within Atom (e.g. when you change settings in the [settings view](https://github.com/atom/settings-view)), the new settings are written to your `config.cson`. Each window process does this independently, and also sets up its own independent file watcher to detect when `config.cson` changes.

When settings are updated simultaneously in *multiple* Atom windows, the `config.cson` file can be written to simultaneously by multiple window processes and become corrupted. This rarely happens since it is hard for a user to simultaneously interact with multiple Atom windows. The most common way that it occurs is when quitting (or restarting) the app. Certain packages update your settings during deactivation, which is problematic if multiple windows are being closed at once.

### Solution

This PR reorganizes our interaction with `config.cson` so that it is centralized in the main process. The main process sets up one *single* file watcher and notifies all of the window processes when `config.cson` changes. Similarly, when settings are updated in some window, that window sends the new settings to the main process, and *it* takes care of writing them to disk.

In the process, I've made all the IO relating to `config.cson` asynchronous.

### Drawbacks

If you run multiple instances of Atom (e.g. Atom Beta and Atom Stable) simultaneously, it is theoretically possible that the two *main* processes could still end up writing to `config.cson` simultaneously. It would be more difficult to cause this to happen though; you'd have to perform some action that affected not only multiple windows, but *multiple applications* simultaneously.

We could still later introduce some kind of lock file, but I'm not sure it's worth it anymore, since contention will now be so uncommon.

Fixes https://github.com/atom/atom/issues/14909

### DIFF
--- a/spec/config-file-spec.js
+++ b/spec/config-file-spec.js
@@ -1,0 +1,109 @@
+const {it, fit, ffit, beforeEach, afterEach, conditionPromise} = require('./async-spec-helpers')
+const fs = require('fs-plus')
+const path = require('path')
+const temp = require('temp').track()
+const dedent = require('dedent')
+const ConfigFile = require('../src/config-file')
+
+describe('ConfigFile', () => {
+  let filePath, configFile, subscription
+
+  beforeEach(async () => {
+    jasmine.useRealClock()
+    const tempDir = fs.realpathSync(temp.mkdirSync())
+    filePath = path.join(tempDir, 'the-config.cson')
+  })
+
+  afterEach(() => {
+    subscription.dispose()
+  })
+
+  describe('when the file does not exist', () => {
+    it('returns an empty object from .get()', async () => {
+      configFile = new ConfigFile(filePath)
+      subscription = await configFile.watch()
+      expect(configFile.get()).toEqual({})
+    })
+  })
+
+  describe('when the file is empty', () => {
+    it('returns an empty object from .get()', async () => {
+      writeFileSync(filePath, '')
+      configFile = new ConfigFile(filePath)
+      subscription = await configFile.watch()
+      expect(configFile.get()).toEqual({})
+    })
+  })
+
+  describe('when the file is updated with valid CSON', () => {
+    it('notifies onDidChange observers with the data', async () => {
+      configFile = new ConfigFile(filePath)
+      subscription = await configFile.watch()
+
+      const event = new Promise(resolve => configFile.onDidChange(resolve))
+
+      writeFileSync(filePath, dedent `
+        '*':
+          foo: 'bar'
+
+        'javascript':
+          foo: 'baz'
+      `)
+
+      expect(await event).toEqual({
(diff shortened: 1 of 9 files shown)


maxbrunsfeld:
This is almost good to go, except that I noticed an annoying behavior when you change the font size config setting rapidly by holding down `Command +`. Something about the way I'm debouncing the file change events causes the config setting to briefly be set back to an older previous value, so the font size flickers.

maxbrunsfeld:
/cc @ungb @Ben3eeE 

This was a tough bug to reproduce. Here's how I did it.

1. Pick a non-default theme (like `one-light-ui`).
2. Put code in `init.js` (or `init.coffee`) that writes to the `atom.config` constantly:

```js
setInterval(() => {
  atom.config.set('editor.tabLength', 2)
}, 250)
```

3. Open 4 Atom windows.
4. Eventually, see that out of nowhere, the config settings get briefly reset back to the defaults, causing the theme to switch to `one-dark-ui`.

cdmackay:
hmm. I just got a config.cson wipeout whilst running 1.25beta0, before updating to beta1. did this fix make beta0?

Arcanemagus:
Yes, this was in Atom v1.25.0-beta0 and you aren't the only one (#16786, @rsese).

cdmackay:
thanks much.

cdmackay:
Should I log a new bug, or are these comments here sufficient to note that there's still a problem here (albeit possibly different cause)?

rsese:
The comments are ok for now @cdmackay, thanks for reporting - we'll be talking about this one soon and if we open another issue we'll ping you both in it.

> I had just updated a pkg, and restarted after that pkg update, and Atom came back with default theme, and all settings gone.

Do you happen to remember what package you were updating?

rsese:
Ugh sorry, I missed that Max opened https://github.com/atom/atom/pull/16810 as a follow-on to this work :zap:

maxbrunsfeld:
An additional fix for this will go out in Atom 1.25.0-beta2.

cdmackay:
@rsese It may not matter now, but yes.

I was running 1.25.0-beta0; I saw a notification that beta1 was available, but first (whilst running beta0) I updated pkg [todo-show](https://atom.io/packages/todo-show). Restarted after that pkg update, and dark theme, settings gone, still on beta0.

Btu I see #16810 is in now too, thanks.

PR #16633, opened 2018-01-25
Fix typo in TreeSitterLanguageMode.onDidChangeHighlighting

maxbrunsfeld:
😬  A typo in the event name was causing the `onDidChangeHighlighting` method to *never* call its callback. The `TreeSitterLanguageMode` tests weren't catching this because they would always invalidate the display layer's screen line cache in order to test another aspect of highlighting. This fixes the tests so that they properly test the invalidation.

/cc @nathansobo @Ben3eeE

### DIFF
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -138,6 +138,38 @@ describe('TreeSitterLanguageMode', () => {
         ]
       ])
     })
+
+    it('updates lines\' highlighting when they are affected by distant changes', () => {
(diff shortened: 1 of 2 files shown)


Ben3eeE:
Amazing :100: I'll pull and recheck the syntax-highlight issues that I saw.

Ben3eeE:
I was still able to reproduce some of the issues I have been seeing and opened issues for them /cc: https://github.com/atom/atom/issues/16643 and https://github.com/atom/atom/issues/16642

PR #16638, opened 2018-01-27
:arrow_up: electron 1.6.16 to Fix protocol handler bug. CVE-2018-1000006

the-j0k3r:
### Requirements

### Description of the Change

Electron fix -> [SECURITY] Fixed protocol handler bug. CVE-2018-1000006

Because it fixes a known exploit

### Benefits

Close a known exploit in Atom dues to electron

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "1.6.15",
+  "electronVersion": "1.6.16",
   "dependencies": {
     "@atom/nsfw": "^1.0.18",
     "@atom/source-map-support": "^0.3.4",


50Wliu:
> Close a known exploit in Atom dues to electron

I don't believe Atom is vulnerable to this exploit on 1.23.3 or 1.24.0-beta3.  That said, I'm also obviously not against merging this.  However, if you create a PR for 1.23-releases then I think you'd also need to create a PR with similar changes for 1.24-releases.

damieng:
We actually have mitigation already in Atom 1.23 via the `--` option at https://github.com/atom/atom/blob/1.23-releases/src/protocol-handler-installer.js#L15

I did upgrade 1.24 to 1.6.16 and master to 1.7.11 as an extra precaution.  While there is no harm in upgrading 1.13 to 1.6.16 as well 1.24 will shortly become the stable branch anyway.

the-j0k3r:
My due diligence as a concerned user is done. 😸 🍰 
This goes without saying, but I'll say it anyway, merging this asap only matters, if a release of the 1.23 branch is made shortly after.

If there is full confidence by the Atom team in the mentioned mitigation as is, then feel free to close this.

Have a nice weekend people ;)

PR #16654, opened 2018-01-30
WIP: Add support for project and path specific configs. 

philipfweiss:
Work in progress PR to give atom support for project specific configuration files. 

### Description of the Change

The purpose of this change is to add the ability for atom to act upon path and project specific configuration files (instead of just a global `~/.atom/config.{json, cson}`. After these changes, key/value pairs set within a `.atom/config.{json, cson}` on any  mounted root will be honored by the editor. Additionally, packages will be able to set project specific settings on `atom.config`. 

This pull request includes mostly changes to the internals of atom.config. While all existing APIs (even private class methods of atom.config) still work, their implementation has in many cases been moved out to more generic functions that can be called on arbitrary setting contexts. 

Previously, the only settings context in atom was the global context, which was acted upon by the various methods of the atom.config class. There are now multiple settings contexts, each with different precedences. While `atom.config.set` still acts upon only the global context, `atom.config.get` now looks through all of the settingsContext objects, grabbing values in order of precedence. The settings contexts are:

1. Dirty Settings. Any setting changed by atom.config.set within the current atom session will be added to the dirty settings, which have the highest precedence. This is to prevent issues where a value is set by a package, but then cannot be retrieved because a settings context of higher precedence has already defined it. 

2. Project Settings. Atom config now has the concept of a "project", which packages can write settings to by calling `atom.config.resetProjectSettings`. 

3. Path Settings. Atom config now stores settings contexts for each of the roots mounted in the current atom session. 

4. Global Settings. This is the original settings context found in `~/.atom/config.{json, cson}`

### Benefits

To allow users of atom to have project and path specific configuration, instead of just a personal .atom directory. It also lays the groundwork for other project-specific code within atom. Future directions include the ability to include suggested packages in a project(!!), and other features currently in `~/.atom`, such as project specific `init.js` files. 

### Possible Drawbacks

Although no APIs were changed, the behavior of `atom.config.get` has been changed. `get` now searches through all the settings contexts, while it previously only looked through the global settings. These changes include new tests and pass previous tests, but because of the centrality of atom.config to the operation of many packages, and the many changes made to the file, we should be on the lookout for possible regressions introduced by these changes. 

### Verification Process
New tests were added to ensure proper behavior and to protect against regressions. 

### Applicable Issues

WIP


src/config.js
```diff
@@ -606,36 +646,56 @@ class Config {
   // Returns an {Array} of {Object}s with the following keys:
   //  * `scopeDescriptor` The {ScopeDescriptor} with which the value is associated
   //  * `value` The value for the key-path
-  getAll (keyPath, options) {
+  getAll (keyPath, options = {}) {
     let globalValue, result, scope
```
wbinnssmith: these declarations at the top are probably a relic of decaffeination. I'd just inline them as they're initialized below.

src/config.js
```diff
@@ -606,36 +646,56 @@ class Config {
...
+        const scopeDescriptor = ScopeDescriptor.fromObject(scope)
+        result = result.concat(setting.scopedSettings.getAll(
+            scopeDescriptor.getScopeChain(),
             keyPath,
             options
-          ) || []))
+          ))
+        legacyScopeDescriptor = this.getLegacyScopeDescriptor(scopeDescriptor)
+        if (legacyScopeDescriptor) {
+          result.push(...Array.from(setting.scopedSettings.getAll(
```
wbinnssmith: why a concat above and a push with spread here?

given the complexity of this and its similarity to the concat call above, I would probably extract this as a function and give it a meaningful name

src/config.js
```diff
@@ -589,12 +608,33 @@ class Config {
...
-      return this.getRawValue(keyPath, options)
+    // Get should return settings in the following priority:
+    // 1. dirtySettings (settings that have been changed in the recent session.)
+    // 1. projectSettings (active project should have highest precedence)
+    // 2. rootSettings (mounted roots should take precedence over personal settings)
+    // 3. globalSettings
+    let getVal = null
+    const settings = _.flatten([this.dirtySettings, this.projectSettings, Array.from(this.pathSettingsMap.values()), this.globalSettings])
+
+    for (let setting of settings) {
```
wbinnssmith: `const` should do here

src/config.js
```diff
@@ -606,36 +646,56 @@ class Config {
...
-    } else {
-      result = []
+      allResults.push(result)
     }
 
-    globalValue = this.getRawValue(keyPath, options)
+    const reducedResults = this.reduceAllResults(allResults)
+
+    delete options.scope
+    globalValue = this.get(keyPath, options)
```
wbinnssmith: `globalValue = this.get(keyPath, _.without(options, 'scope'))` here should take care of the issue above 😄

src/config.js
```diff
@@ -764,7 +847,13 @@ class Config {
   // Extended: Get an {Array} of all of the `source` {String}s with which
   // settings have been added via {::set}.
   getSources () {
-    return _.uniq(_.pluck(this.scopedSettingsStore.propertySets, 'source')).sort()
+    const arrs = []
+    arrs.push(_.uniq(_.pluck(this.globalSettings.scopedSettings.propertySets, 'source')))
+    arrs.push(_.uniq(_.pluck(this.projectSettings.scopedSettings.propertySets, 'source')))
```
wbinnssmith: is there a better name than `arrs`? maybe something like `settingsSources`?

src/config.js
```diff
@@ -764,7 +847,13 @@ class Config {
   // Extended: Get an {Array} of all of the `source` {String}s with which
   // settings have been added via {::set}.
   getSources () {
-    return _.uniq(_.pluck(this.scopedSettingsStore.propertySets, 'source')).sort()
+    const arrs = []
+    arrs.push(_.uniq(_.pluck(this.globalSettings.scopedSettings.propertySets, 'source')))
```
wbinnssmith: are these `uniq`s needed if there's one that takes place after a flattening below?

src/config.js
```diff
@@ -764,7 +847,13 @@ class Config {
...
+    arrs.push(_.uniq(_.pluck(this.projectSettings.scopedSettings.propertySets, 'source')))
+    for (let settingsContext of this.pathSettingsMap.values()) {
+      arrs.push(_.uniq(_.pluck(settingsContext.scopedSettings.propertySets, 'source')))
+    }
+    return (_.uniq(_.flatten(arrs))).sort()
```
wbinnssmith: might be a bit too much, but this can be expressed as a large expression. Personally (with the addition of some comments re: what's going on with pathSettingsMap) this is more readable than building up a bunch of lists over time:

```
return _.unique(_.flatten([
  ...this.globalSettings.scopedSettings.propertySets.map(s => s.source),
  ...this.projectSettings.scopedSettings.propertySets.map(s => s.source),
  [...(_.flatten(this.pathSettingsMap.values().map(context =>
    context.scopedSettings.propertySets.map(s => s.source)
  )))
])).sort()
```

cc @matthewwithanm

src/config.js
```diff
@@ -959,43 +1085,57 @@ class Config {
...
   getRawValue (keyPath, options = {}) {
-    let value
-    if (!options.excludeSources || !options.excludeSources.includes(this.mainSource)) {
-      value = getValueAtKeyPath(this.settings, keyPath)
+    const value = this.getRawValueFrom(this.globalSettings, keyPath, options)
+    return this.replaceWithDefaultValue(value, keyPath, options)
+  }
+
+  getRawValueFrom (settings, keyPath, options = {}) {
+    const configIndex = (options.excludeSources || []).indexOf(this.getUserConfigPath())
```
wbinnssmith: `.includes()`?

src/config.js
```diff
@@ -1162,26 +1308,42 @@ class Config {
...
+  resetUserScopedSettingsOn (newScopedSettings, settings, options = {}) {
+    const emitChange = options.emitChange == null ? true : options.emitChange
 
+    const source = this.getUserConfigPath()
+    const priority = this.priorityForSource(source)
+    settings.scopedSettings.removePropertiesForSource(source)
     for (let scopeSelector in newScopedSettings) {
-      let settings = newScopedSettings[scopeSelector]
-      settings = this.makeValueConformToSchema(null, settings, {suppressException: true})
+      let curSettings = newScopedSettings[scopeSelector]
```
wbinnssmith: I would just inline the line below:

```js
const curSettings = this.makeValueConformToSchema(
  null,
  newScopedSettings[scopeSelector], 
  {suppressException: true}
);
```

src/config.js
```diff
@@ -1239,6 +1408,87 @@ class Config {
...
     }
   }
+
+  collectFilePromises(configPaths) {
+    return configPaths.map((curPath) => {
+      const jsonPath = path.join(curPath, '.atom', 'config.json')
+      const csonPath = path.join(curPath, '.atom', 'config.cson')
+
+      return new Promise((resolve, reject) => {
+        fs.access(jsonPath, fs.constants.R_OK, (err) => {
```
wbinnssmith: is there already an abstraction in atom over promisifying `fs` and other errback modules?

wbinnssmith (changes requested):
Apologies for the nits. A few of them are just personal preference (hopefully is clear from the inline comments).

Overall this looks great. Can't wait 😁

src/config.js
philipfweiss: I'm not sure - although I know this exists in Nuclide.

skylize:
Is it _really_ the best API choice to overload `atom.config.get` like this? What if a package _very specifically_ wants to know what the user's global setting is, and has some particular reason to ignore project or path settings? It's now inaccessible, unless there are no "higher priority" settings.

**Example alternative api**
```
atom.config.set('myGlobalSetting')
atom.config.get('myGobalSetting')

atom.config.project.set('myProjectSetting', 'value')
atom.config.project.get('myProjectWideSetting')

atom.config.rootPath.set('myRootPathSetting', 'value')
atom.config.rootPath.get('myRootPathSetting')

// Explicitly request automatic resolution
atom.config.autoResolve.get('mySettingThatMayBeInAnyScope')
```

src/config.js
philipfweiss: This is actually from the original Atom code - I can refactor this though!

philipfweiss:
@skylize, The main reason we decided to go with this API design is that we want project specific settings to be applied to packages that are already using the `atom.config.set` and `atom.config.get` APIs. For example, lets say a project wants to set the default tabwidth to 15. To do this, they put inside their project's `config.json` something along the lines of `tabwidth: 15`. 

In this case, existing packages are already calling `atom.config.get('tabwidth')` to get the tab width. If we were to implement the separate API `atom.config.project.get...`, then the existing packages would not be getting these new project specific settings, and the changes would not be applied. 

Essentially, the reason we decided to overload is that moving the API to `atom.config.project...` or `atom.config.paths...` would make these new configs useless for any package that isn't written to respect the new API.

If a package really wants to know what a user's global setting is, than they could use the (non-public) `atom.config.getRaw{Scoped}ValueFrom`API. 


@matthewwithanm

skylize:
Thx for the reply @philipfweiss. 

Your argument definitely seems sensible at first glance. I certainly see the benefit of making the _default_ API automatically add the new features for existing packages.

You say that "project/path configs are read-only by design". This _sounds_ nice from the perspective of simplicity, but that also means _all_ project/path settings _must_ be set _explicitly_ **by the user** in the Settings pane. 

From a UI perspective, this can easily become a PITA for the user depending on the specific setting. If a package has a good method of adjusting certain configs based on user interaction _and_ the user wants those changes stored at a project or path level, then developers should have the ability to store these interactive configs on the project.

With that in mind, it would be _much_ better from the user's perspective if the API gave developers a reasonable way to opt in to explicitly handling project/path configs separately from global ones, thus allowing users to opt in to allowing the package to make project and path specific adjustments on their behalf.

philipfweiss:
Hi @skylize, I think there might be a confusion about what these changes are doing. It is not the case that "all project/path settings must be explicitly set by the user in the settings pane". 

The typical user will interact with these changes in the following way:

1. Project owner writes a .atom/config.js in the project's home directory. The can put things in here such as suggested packages, default tabwidth, editor color, or whatever.
2. A user loading this project will get these settings by default whenever they have mounted the project root, since .get will resolve to the project specific settings context. One can imagine a new team member joining a project getting their team's settings just because their directory already has the config file within it (and doing no work otherwise). No other work is involved from the user's perspective. This functionality is currently in VSCode. 

There is no mechanism for the user to programmatically write to the project's config.js through atom. They can change their *global settings* by going into the settings pane, or they can make a change to the config file and commit it. Changing the settings in the settings pane will write to the global settings file, and change the setting for that instance of atom (since it will be written to dirty settings), but it will not change the project config and the next time the user loads the project it will read from the same project config and reset the settings.

skylize:
> but it will not change the project config and the next time the user loads the project it will read from the same project config and reset the settings

This sounds _terrible_. User interacts with a package that used to save some settings permanently to a global config, and now those settings get disposed of on every restart in favor of project config, with no opportunity for packages to "just do the right thing". 👎 👎

wbinnssmith:
>  and now those settings get disposed of on every restart in favor of project config

that's not what's happening. Any package that uses `config.set` still ends up writing to the user's `~/.atom/config.json` as it always has.

I'd give it a spin or read through the code if you have further questions.

skylize:
>> and now those settings get disposed of on every restart in favor of project config
>
> that's not what's happening. Any package that uses config.set still ends up writing to the user's ~/.atom/config.json as it always has.

Ummmm.... If that's not what's happening, then why did @philipfweiss say this?

> Changing the settings in the settings pane will write to the global settings file, and change the setting for that instance of atom (since it will be written to dirty settings), but it will not change the project config and the next time the user loads the project it will read from the same project config and reset the settings.

Okay. I guess I _slightly_ mispoke when I said "disposed of". But what good is it to write to `.atom/config.json` if those settings will _always_ be shadowed by some another file somewhere else in the system? There is effectively zero difference to the user between shadowed and disposed of. The point is that every time they restart, they lose all the custom settings they worked out for themselves to make their own workflow a little easier.

***User should be able to customize her own editor***, no matter what the project leader says the _default_ config should be. 

Maybe you will make the Project Leader happy by always resetting to his chosen config. Maybe you will even make the New Junior Dev _think_ that he is happy on his first day at work. 

But after a few weeks of having to rebuild his own custom config _every single time_ he opens up his editor, he's probably going to lobby his boss to switch to VS Code.

skylize:
> I'd give it a spin or read through the code if you have further questions.

This conversation began as a discussion of API. Any attempt to "give it a spin" in this context would be a _very significant_ time investment because I would need to build a package that puts a new and unfamiliar API through its paces.

Since the concerns are very high-level, and not to do with any specific details of the code-quality (which I assume is _amazing_), it seems preferable to keep discussions at that same high level, rather than me spending hours digging through code to discover something that somebody here could be explain to me in a paragraph.

___
I sincerely apologize if I am coming across aggressively to anybody. I shared what I thought was a possible flaw in the API, only to discover that the API was chosen because it supports an interface that (**as so far described**) sounds _to me_ like it will be very unfriendly to users. If I am wrong, I am _more than happy_ for someone to correct me. I am not here to argue, but rather to try and assist with making Atom the best it can be, and to get the jump on what I see as a potential problem before it gets pushed onto the master branch.

damieng:
What does 'project' mean in this context?  Today it means the set of root folders with associated window state but such a concept doesn't have a project-specific location to store a settings config file.

matthewwithanm:
Hey all! Good points here! Hopefully I can clear some things up and enumerate the options and their tradeoffs 😊

We have broad agreement that there should be different "layers" of configuration, each with different precedence, and that `config.get()` should get the value from the layer of the highest precedence that contains the setting.

The real question is **"what does `config.set()` update?"** (Of course, we can add arguments or other API to make it target a specific layer if needed, but what is its default behavior?) It seems reasonable that it should update the global config (`~/.atom/config.json`) since that's the current behavior. But we also need to maintain the invariant that **getting a value after setting it returns the value you set it to.**

In other words, we always need to write to the layer of highest precedence.

Possible solutions are (1) to make sure that the global config always has the highest precedence (as suggested by @Arcanemagus), (2) write to a different (or multiple) files, or (3) write to both the global config and an in-memory "dirty settings" which has the highest precedence (as currently done in the PR).

The tricky part here is that precedence involves two concepts: ownership (user vs the project owner's) and breadth (global vs per-folder).

| | Per-folder | Global |
| ----------------- |:--------------------:|---------------------:|
| **User** | | `~/.atom/config.json` |
| **Project Owner** | `./.atom/config.json` | |

Just considering the "ownership axis", solution 1 seems good: the user's choice is favored over the owner's. However, this is ignoring the "breadth axis," which is a pretty severe limitation. Note that it would be impossible for a user to even override *their own* settings on a per-folder basis. (Incidentally, it looks like this is the case [with Sublime][1] while [VSCode gives precedence to per-project settings][2].)

On the other hand, an approach that didn't require the global user config to be given the highest precedence would allow us to fill in the user/per-folder slot of the matrix, say with a `./.atom/config-user.json`. This file would specify per-project overrides, have the highest priority, and (unlike `./.atom/config.json`) wouldn't be checked in to version control. (This is actually something we've already discussed but figured could be added later.) Again, solution 1 precludes this because it requires the global user config to have the highest precedence.

For these reasons, I think solution 1 isn't worth the tradeoffs. Solution 2 is a possibility but has many unanswered questions. (Do we write to more than one config? Do we need to implement `config-user.json` immediately to avoid changing version controlled configs?). Solution 3 ("dirty settings") has the issues you've outlined but I'm not sure I agree about the severity; it seems pretty sensible that owner-contributed settings wouldn't be updated automatically, and that it's the responsibility of the project owner not to put crazy stuff (themes) in there.

> What does 'project' mean in this context? Today it means the set of root folders with associated window state but such a concept doesn't have a project-specific location to store a settings config file.

This is actually not meant to be public API and is a way for us (the Nuclide team) to experiment with the concept of a project that isn't directly analogous to the set of root folders. At Facebook, people tend to mount the root of a monorepo even when they're working on a conceptual "project" deeply nested within it. This would give us a way to tell Atom to use its config even though it wasn't a root.

[1]: https://www.sublimetext.com/docs/3/settings.html
[2]: https://code.visualstudio.com/docs/getstarted/settings#_language-specific-editor-settings

damieng:
How will atom.config.get know which folder provides the right context-specific setting?  

It seems to me the safest option is to make atom.config.set and atom.config.get stay with just setting global and each should have an optional parameter that can specific a folder to get a more specific configuration from.

The negative is that it does mean existing atom apps/plugins need to be updated to take advantage of this but I don't see how atom.config.get is going to know the context otherwise.

Additionally this approach means we don't have to do the dirty settings hack.

damieng:
Meeting outcome from discussion with Facebook on project config:  

- FB are redefining [their PR](https://github.com/atom/atom/pull/16654) to just be an explicit project config proposal.  i.e.   `myProject.atomProject.json` will be a file that will contain an array of root paths and associated config.  If you open from new File menu option or `-p` on command line it will open it as a project not as a text file.  atom.config.get will read from this config then global.  There will be no folder-level stuff here.
 - (RFC-001](https://github.com/atom/atom/pull/16698) will be tweaked to be for the root-folder level config that LSP needs which IDE team will develop.  It will only deal with root folder configs (not project level) and any atom.config.get/set call will specify the folder name (because we can't magically guess which folder the call was running in as we don't have separate contexts for each root folder that they could honor for it)
 - We'll rework the File/Atom menu a bit to make the project stuff clearer
 - Initial config file editing/viewing will be to open the file in Atom as a file and edit as text - I have some ideas on how we can make settings-view provide something down the line that is understandable that I bounced off Simurai at mini-summit (edited)

Issue #6270, opened 2015-04-06
Allow a location option when adding menu items


jessegrosjean:
Right now when you add menu items they get inserted at the bottom of the specified menu. I would like to have some optional way to the specify a location when inserting menu items. From what I can tell this isn't possible now. Is anyone working on this already? If not and I create a solution what are the chances that I can get it merged into Atom?

My basic design would follow how I think Eclipse solves this problem. Haven't actually used Eclipse for years, but I think the basic idea is: Give each item an optional "location" attribute that's used to determine where the item is inserted relative to existing item IDs in the menu.

The ID of existing menu items would be an "ID" attribute if found, otherwise it would fallback to using "label", though that's problematic if menus will ever be localized.

The set of locations specifiers would be:

`location="top"` - The item will be inserted at the top of the menu.

`location="bottom"` - The item will be inserted at the bottom of the menu.

`location="before:id"` - The item will be inserted before the existing item with matching ID.

`location="after:id"` - The item will be inserted after the existing item with matching ID.

`location="endof:id"` - The item will be inserted at the end of the separator group with matching ID. (this means it will be inserted right before the next separator group starts.

jessegrosjean:
I've created a pull request in Atom Shell to get this started:

https://github.com/atom/atom-shell/pull/1373

Once that's in place then I think the next changes would be to define ids for existing Atom menus and to update menu-helpers.coffee to include ids and positions attributes when merging templates. Feedback still welcome of course.

mnquintana:
For anyone else following this, see https://discuss.atom.io/t/is-there-a-way-to-specify-menu-item-order-when-inserting-menu-items/14937 for more design discussion.

mnquintana:
@jessegrosjean Looking at the API you implemented in atom/electron#1373, I think rather than a string, I'd rather see it implemented as object properties, so something like:

```cson
  'atom-pane[data-active-item-path] .item-views': [
    {
      'label': 'Reveal in Tree View', 
      'command': 'tree-view:reveal-active-file', 
      'position': {
        'endof': 'tree-view:show-in-file-manager'
      }
    }
  ]
```

Instead of:

```cson
  'atom-pane[data-active-item-path] .item-views': [
    {
      'label': 'Reveal in Tree View', 
      'command': 'tree-view:reveal-active-file', 
      'position': 'endof=tree-view:show-in-file-manager'
    }
  ]
```

What do you think?

jessegrosjean:
[...]

PR #16661, opened 2018-02-01
Enable packages to control order of context menu items

captbaritone:
### Goals:

* Allow packages to share context menu “groups” (sets of context menu items delineated by separators). For example, a package should be able to provide a “Copy Full Path” context menu item which lives in the same group as the core “Copy” item.
* Allow packages to control the ordering of context menu items.
* Allow packages to control the order of context menu groups.

### Constraints:

* Packages should be able to position their items alongside other package's items without requiring the other package to make any changes.
* The current sorting behavior should be maintained for context menus which do not contain any items which use this new option.
* Packages may make ordering requests which are in contradictory. In this case, some ordering requests must be ignored in a deterministic way.

### Prior work:

* There has been at least [one prior proposal](https://github.com/atom/atom/issues/6270) to (discussion [here](https://discuss.atom.io/t/is-there-a-way-to-specify-menu-item-order-when-inserting-menu-items/14937)) support ordering in Atom context menus. It got as far as [merging support into Electron](https://github.com/electron/electron/pull/1373). Unfortunately, the Electron implementation has [some problems](https://github.com/electron/electron/issues/11668), and the Atom support was never implemented.
* The prior proposal seems to be based upon the the [API that Eclipse has](https://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_ui_menus.html).

> screen shot 2018-01-25 at 2 28 25 pm

-- Extract from the Eclipse documentation.

### Proposed Solution:

* Add four new optional properties to the context menu items accepted by atom.contextMenu.add() each accepting as its value either an array of atom commands. The four properties are:
    * `before`
    * `after`
    * `beforeGroupContaining`
    * `afterGroupContaining`
* `before`/`after` Provide a means for a single context menu item to declare their placement relative to another context menu item. These also imply that menu item in question should be placed in the same “group” as the item 
* `beforeGroupContaining`/`afterGroupContaining` Provide a means for a single context menu to declare the placement of their containing group, relative to the containing group of the specified item. Ideally these could be declared on some entity representing the notion of a “group”, but no such entity exists at this time, and I don't believe it's worth introducing a new construct within this API.

Edge case handling:

* If no menu item with the given command exists, the relationship is ignored. This allows package authors to declare positioning relative to packages which the user may or may not have installed.
* When a menu item attempts to position itself relative to another item, the two groups are joined. Note that this can happen recursively.
* If two sets of menu items are provided by two consecutive plugins, and they don't delineate their items with separators, the items at the boundaries will be merged together into a group. If any item in that group specifies its position, it would cause all items in that new group to be relocated.

Possible challenges:

* This approach assumes that context menus do not contain multiple entries that dispatch the same command. I can't think of any rational case in which this would happen, but it could be that there is some case I have not thought of. Another option would be for menu items to declare their position relative to other items via the other item's label. This would have the advantage that Atom is already enforcing label uniqueness, but it does not seem as stable for me. Localization would be one reason to avoid using labels as an item's unique identifier.

### DIFF
--- a/spec/menu-sort-helpers-spec.js
+++ b/spec/menu-sort-helpers-spec.js
@@ -7,12 +7,14 @@ describe('contextMenu', () => {
       expect(sortMenuItems(items)).toEqual(items)
     })
   })
+
   describe('dedupes separators', () => {
     it('trims leading separators', () => {
       const items = [{ type: 'separator' }, { command: 'core:one' }]
       const expected = [{ command: 'core:one' }]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it('preserves separators at the begining of set two', () => {
       const items = [
         { command: 'core:one' },
@@ -25,11 +27,13 @@ describe('contextMenu', () => {
       ]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it('trims trailing separators', () => {
       const items = [{ command: 'core:one' }, { type: 'separator' }]
       const expected = [{ command: 'core:one' }]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it('removes duplicate separators across sets', () => {
       const items = [
         { command: 'core:one' }, { type: 'separator' },
@@ -43,6 +47,7 @@ describe('contextMenu', () => {
       expect(sortMenuItems(items)).toEqual(expected)
     })
   })
+
   describe('can move an item to a different group by merging groups', () => {
     it('can move a group of one item', () => {
       const items = [
@@ -61,6 +66,7 @@ describe('contextMenu', () => {
       ]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it("moves all items in the moving item's group", () => {
       const items = [
         { command: 'core:one' },
@@ -80,6 +86,7 @@ describe('contextMenu', () => {
       ]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it("ignores positions relative to commands that don't exist", () => {
       const items = [
         { command: 'core:one' },
@@ -99,6 +106,7 @@ describe('contextMenu', () => {
       ]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it('can handle recursive group merging', () => {
       const items = [
         { command: 'core:one', after: ['core:three'] },
@@ -112,6 +120,7 @@ describe('contextMenu', () => {
       ]
       expect(sortMenuItems(items)).toEqual(expected)
     })
+
     it('can merge multiple groups when given a list of before/after commands', () => {
       const items = [
         { command: 'core:one' },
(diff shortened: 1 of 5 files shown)


spec/menu-sort-helpers-spec.js
```diff
@@ -0,0 +1,227 @@
+const {sortMenuItems} = require('../src/menu-sort-helpers')
+
+describe('contextMenu', () => {
+  describe('dedupes separators', () => {
+    it('preserves existing submenus', () => {
+      const items = [{ submenu: [] }]
+      expect(sortMenuItems(items)).toEqual(items)
+    })
+  })
+  describe('dedupes separators', () => {
```
maxbrunsfeld: Could you add blank lines between these `it` and `describe` calls?

maxbrunsfeld (approved):
This seems very well designed. Nice work!

Does sorting the menu items ever have a noticeable performance impact when right clicking? It seems like the number of context menu items is usually small, so it's probably not an issue, but it'd be good to look at a profile just in case.

captbaritone:
@maxbrunsfeld Thanks for taking the time to review. I'll try to construct some reasonable worst-case scenarios and profile them.

captbaritone:
For a rather complex context menu of >30 elements containing 20 relationships of various types it takes under 10 milliseconds to sort them.

maxbrunsfeld:
In a lot of code paths in Atom, ten milliseconds would be pretty significant, but since this only occurs on a fairly course-grained basis (once per right click) it doesn't seem like a big deal here.

szeck87:
Do packages that use a `.cson` file to build the context menu need to make any changes to use?  Or does this not have anything to do with them?

captbaritone:
This should work from either CSON or via the `atom.contextMenu.add()`.

lierdakil:
There's no documentation on this feature. Is this intended to be left undocumented, or is this an oversight?

captbaritone:
@lierdakil Yeah, this should probably get documented now that it's been accepted. I'll add it to my list.

lierdakil:
> now that it's been accepted

Uh... to my understanding, it's not only "accepted", but also released with [v1.26.0](https://github.com/atom/atom/releases/tag/v1.26.0)... so yeah, the feature is released, but documentation isn't there. I'm aware that's how it usually is, but it doesn't mean that's a good thing. I would argue the docstring should've been updated as a part of *this* PR.

This is especially not fun considering Atom had a policy "undocumented means unsupported" since the early days. So this feature is in a kind of an awkward spot right now.

To be clear, I'm not trying to start anything, just sharing my point of view. Please don't take it personally.

captbaritone:
@lierdakil Not sure about the official stance, but I'm okay with this being an "unsupported" API until it gets documented. In fact, it might be good to get some people trying it out and proving that it doesn't have any unforeseen problems before it's fully blessed. That said, I don't know exactly how these things generally get rolled out.

The Nuclide team will be doing some work to adopt this in the near future and we will probably get some additional single there to see how this works at a larger scale than my initial tests.

PR #16668, opened 2018-02-02
Optimize process kill by pid

Connormiha:
### Description of the Change
Minor refactoring code for killing process.
Used **1** `for loop` instead of **4** `for loops`

### Benefits

Improve performance

### DIFF
--- a/src/buffered-process.js
+++ b/src/buffered-process.js
@@ -189,12 +189,12 @@ class BufferedProcess {
       output += data
     })
     wmicProcess.stdout.on('close', () => {
-      const pidsToKill = output.split(/\s+/)
-        .filter((pid) => /^\d+$/.test(pid))
-        .map((pid) => parseInt(pid))
-        .filter((pid) => pid !== parentPid && pid > 0 && pid < Infinity)
+      for (let pid of output.split(/\s+/)) {
+        if (!/^\d{1,10}$/.test(pid)) continue
+        pid = parseInt(pid, 10)
 
-      for (let pid of pidsToKill) {
+        if (!pid || pid === parentPid) continue
+
         try {
           process.kill(pid)
         } catch (error) {}


src/buffered-process.js
```diff
@@ -189,12 +189,13 @@ class BufferedProcess {
       output += data
     })
     wmicProcess.stdout.on('close', () => {
-      const pidsToKill = output.split(/\s+/)
-        .filter((pid) => /^\d+$/.test(pid))
-        .map((pid) => parseInt(pid))
-        .filter((pid) => pid !== parentPid && pid > 0 && pid < Infinity)
+      for (let pid of output.split(/\s+/)) {
+        if (!isFinite(pid)) continue
```
Ingramz: Should be redundant because `parseInt` only returns an integer value or `NaN`, both cases are covered at the check below.

src/buffered-process.js
```diff
@@ -189,12 +189,13 @@ class BufferedProcess {
...
+
+        pid = parseInt(pid, 10)
+
+        if (!pid || pid === parentPid || pid < 0 || pid === Infinity) continue
```
Ingramz: `pid === Infinity` will be never hit because `parseInt` won't return `Infinity`, otherwise should sufficiently cover rest of the possible outputs of `parseInt`.

src/buffered-process.js
Connormiha: This check allow to aviod parse digits + text. For example.
```js
let num = '12345aaaaa';
parseInt(num); // 12345
isFinite(num); // false
```

In this case we should parse only clear number, without text tail.
Before was regexp `/^\d+$/`

src/buffered-process.js
Connormiha: Yes, check for `Infinity` is needless in this case.

src/buffered-process.js
Ingramz: Fair enough. What about floats?
```js
let num = '1.2345';
parseInt(num); // 1
isFinite(num); // true
```

Connormiha: Ok. What do you think about use `Number.isSafeInteger` instead of `ifFinite`?
```js
// if (!isFinite(pid)) continue
pid = +pid
if (!Number.isSafeInteger(pid)) continue

Number.isSafeInteger(+'1234'); // false
Number.isSafeInteger(+'1.1'); // false
Number.isSafeInteger(+'123a'); // false
Number.isSafeInteger(+'123a'); // false
```

As bonus we can get rid of `parseInt` and check for `Infinity`

Only a very incredible case pass
```js
Number.isSafeInteger(+'1.'); // true
Number.isSafeInteger(+'1.0'); // true
```
I think it is not neccessery, but we can use regExp to be sure 100%

So the final code is:
```js
for (let pid of output.split(/\s+/)) {
        if (!/^\d{1,8}$/) continue
        pid = parseInt(pid, 10);

        if (!pid || pid === parentPid) continue

        try {
          process.kill(pid)
        } catch (error) {}
}
```

Ingramz: The final code looks a lot more reasonable. Just increase the 8 digits in regex to something that can contain all unsigned 32bit integers, maybe 10.

Ingramz (approved):

src/buffered-process.js
```diff
@@ -189,12 +189,12 @@ class BufferedProcess {
       output += data
     })
     wmicProcess.stdout.on('close', () => {
-      const pidsToKill = output.split(/\s+/)
-        .filter((pid) => /^\d+$/.test(pid))
-        .map((pid) => parseInt(pid))
-        .filter((pid) => pid !== parentPid && pid > 0 && pid < Infinity)
+      for (let pid of output.split(/\s+/)) {
+        if (!/^\d{1,10}$/.test(pid)) continue
+        pid = parseInt(pid, 10)
```
YurySolovyov: I'd use `Number.parseInt` for clarity

Connormiha: In whole project used global `parseInt`. I think migration to `Number.parseInt` should be in special PR.

YurySolovyov (approved):

maxbrunsfeld:
Thanks!

PR #16675, opened 2018-02-03
 Document TextEditor::bufferRangeForScopeAtBufferPosition

chrisjshull:
### Description of the Change

It seemed odd for `bufferRangeForScopeAtCursor` to exist without `bufferRangeForScopeAtBufferPosition` existing, so that one could (e.g.) handle multiple cursors.
`bufferRangeForScopeAtPosition` already existed (used by `bufferRangeForScopeAtCursor`). When I went to add docs I noticed that the other public methods of `TextEditor` used a particular naming scheme, so I also changed the name to "bufferRangeForScopeAtBufferPosition" to match.

### Alternate Designs

N/A

Super handy for figuring out scopes from a package. And has more possible uses than the existing `bufferRangeForScopeAtCursor`.

### Benefits

Ditto.

### Possible Drawbacks

If a package were using this undocumented API then it would break.
I could leave "bufferRangeForScopeAtPosition" in as an undocumented alias if that's a concern.

### Verification Process

It's really just docs. Let me know what you'd like.

### Applicable Issues

N/A

### DIFF
--- a/src/text-editor.js
+++ b/src/text-editor.js
@@ -3668,20 +3668,20 @@ class TextEditor {
   //
   // Returns a {Range}.
   bufferRangeForScopeAtCursor (scopeSelector) {
-    return this.bufferRangeForScopeAtBufferPosition(scopeSelector, this.getCursorBufferPosition())
+    return this.bufferRangeForScopeAtPosition(scopeSelector, this.getCursorBufferPosition())
   }
 
   // Extended: Get the range in buffer coordinates of all tokens surrounding the
   // given position in buffer coordinates that match the given scope selector.
   //
   // For example, if you wanted to find the string surrounding the cursor, you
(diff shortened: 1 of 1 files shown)


cat199018 (approved):

maxbrunsfeld:
Good PR description and good docs, but unfortunately, even though the method is not documented, it appears that a large number of community packages have ended up using it, so it's probably not worth it for us to change the name at the moment. I'd merge this if you undid the name change though.

sadick254:
I made another [PR](https://github.com/atom/atom/pull/22475) since I could not make edits on this PR. I have maintained the authorship of the code and commits so that @chrisjshull gets the deserved credits. I will close this PR as it has now been superseded by https://github.com/atom/atom/pull/22475

PR #16698, opened 2018-02-07
RFC-001: Project-Specific Configuration

lee-dohm:
Add a proposed design for project-specific configuration.

[**Rendered view**](https://github.com/atom/atom/blob/project-settings/docs/rfcs/001-project-specific-configuration.md)

/cc @daviwil @damieng


docs/rfcs/001-project-specific-configuration.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Explanation
+
+There are two configuration files: the global `~/.atom/config.cson` and the project-specific `atom-config.cson`. (Hereafter these two files will be referred to as the "global settings" and "project settings" respectively.) The project settings file is stored in the root directory of the project root.
+
+When reading from the configuration using `atom.config.get()`, the contents of the two configuration files is merged using [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or some similar mechanism, with the values in the project settings taking precedence over the global settings. For example:
```
Arcanemagus: Note that the example retrieval is from within the project, and maybe add an example for global retrieval?

docs/rfcs/001-project-specific-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+
+```javascript
+// Sets the editor.tabLength setting to 4 for GitHub-flavored Markdown files globally
+atom.config.set('editor.tabLength', 4, {scopeSelector: '.source.gfm'})
+
+// Sets the editor.tabLength setting to 2 for GitHub-flavored Markdown files in this project
+atom.config.set('editor.tabLength', 2, {project: true, scopeSelector: '.source.gfm'})
+```
+
+**Note:** [Behavior is undefined](https://blogs.msdn.microsoft.com/oldnewthing/20140627-00/?p=633) when there are multiple project roots that have `atom-config.cson` files whose individual values are different. For example, given a project that has two project roots "foo" and "bar" with project settings like the following:
```
Arcanemagus: Should this fall back to the `global` value instead?

docs/rfcs/001-project-specific-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+editor:
+  fontSize: 16
+```
+
+```javascript
+let fontSize = atom.config.get('editor.fontSize')
+console.log(`fontSize = ${fontSize}`)
+// "fontSize = 16" is output
+
+atom.config.observe('editor.fontSize', (value) => { console.log(`fontSize = ${value}`) })
```
Arcanemagus: This needs a `// "fontSize = 16" is output` line after it. (Since `observe` callbacks are called when initially observing).

docs/rfcs/001-project-specific-configuration.md
Arcanemagus: Just to be clear, leaving it `undefined` is valid at this point as far as I'm concerned.

docs/rfcs/001-project-specific-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+
+atom.config.set('editor.fontSize', 16)
+// "fontSize = 16" is output even though the value that would be returned at this point is still `16`
+```
+
+## Drawbacks
+
+It introduces complexity and some non-determinism into the Atom configuration system.
```
Arcanemagus: Maybe explicitly note the potential for user confusion and increased support tickets?

Arcanemagus (commented):
Does it need to be explicitly stated what happens if the user removes a project folder and adds a new one?

This is implied, but maybe it should be stated somewhere that a user _must_ change the project config file in order to change settings manually. I'm still not sure I like the idea of the user being unable to see in the interface anywhere _where_ a setting is coming from, but as this will initially be in limited use a UI showing that and allowing editing of project settings can come later.

docs/rfcs/001-project-specific-configuration.md
lee-dohm: Ah, good point ... thanks!

docs/rfcs/001-project-specific-configuration.md
lee-dohm: Added.

docs/rfcs/001-project-specific-configuration.md
lee-dohm: Added.

docs/rfcs/001-project-specific-configuration.md
lee-dohm: Leaving this alone for now.

damieng:
I'd really like us to be able to support root-folder level configuration setting - we need this even more than project level in atom-ide.

I think it could exist with the api proposed here but with a different get/set;

`atom.config.set('ide-java.classPath', '/blah/blah;, {folder: '/src/myjavaproject1})`

Get would also need to specify the path - we can't automagically know which folder to get it in as by the time atom.config.get happens it's probably lost all app-level context of where that call was made and what folder it could possibly relate to.

Arcanemagus:
How do you propose `atom.config.observe` would work when requiring a path to be specified @damieng?

Note that I definitely prefer specifying a root-folder when setting a value as that removes the `undefined` behavior when there are multiple roots added to the "project".

damieng:
Yeah the project level stuff seems fraught with confusion until we have an explicit project system with it's own configuration of folders and settings.  e.g. a  damien.atomproject file with;

`json
{
  folders: [ '/src/first', '/src/second' ],
  configuration: {
    core: {
      editor: {
        tab-length: 2
      }
    }
  }
}
```

I'm +1 on just extending set/get/reset/observe to allow them to specify a root folder. 

There will also have to be some expectation set-up that not every setting will work per-folder, e.g. editor.fontSize probably won't be happy switching back and forth as you jump between files?

damieng:
Meeting outcome from discussion with Facebook on project config:  

- FB are redefining [their PR](https://github.com/atom/atom/pull/16654) to just be an explicit project config proposal.  i.e.   `myProject.atomProject.json` will be a file that will contain an array of root paths and associated config.  If you open from new File menu option or `-p` on command line it will open it as a project not as a text file.  atom.config.get will read from this config then global.  There will be no folder-level stuff here.
 - This RFC will be tweaked to be for the root-folder level config that LSP needs which IDE team will develop.  It will only deal with root folder configs (not project level) and any atom.config.get/set call will specify the folder name (because we can't magically guess which folder the call was running in as we don't have separate contexts for each root folder that they could honor for it)
 - We'll rework the File/Atom menu a bit to make the project stuff clearer
 - Initial config file editing/viewing will be to open the file in Atom as a file and edit as text - I have some ideas on how we can make settings-view provide something down the line that is understandable that I bounced off Simurai at mini-summit (edited)

Arcanemagus:
> This RFC will be tweaked to be for the root-folder level config that LSP needs which IDE team will develop.

Awesome, that should cover the essentially same need that I see issues for on the `linter-*` providers!

----

> How do you propose atom.config.observe would work when requiring a path to be specified?

> @Arcanemagus not sure I understand the question.

Currently most of the `linter-*` providers use `atom.config.observe` on their settings, using the callback to cache the setting on activation and whenever it changes. I was wondering how you were thinking that would work with requiring a root path to be specified to enable grabbing the specific values for a root folder. The RFC as it is currently written "makes it work" by essentially just transparently filling in the current root automatically, but that won't be able to be done if it needs to be specified manually.

* Will `observe` be limited to `global` settings? 😕
* Will packages need to `observe` _all_ root paths they see, managing these subscriptions and deciding on their own which value to use?

From the last time I looked I think most of the LSP clients are currently just using `atom.config.get`, so it likely makes no difference there.

shoppingandmore2017:
docs/rfcs/000-template.md

sadick254:
@lee-dohm looking at the outcome of the meeting about this RFC. It looks like a lot of work needed to be done and back then Atom had the capacity to do it. However, right now we don't have an IDE team and also we don't see ourselves doing a rework on the File/Atom menu a bit to make the project stuff clearer. I am closing this for now, but happy to reopen when there is capacity.

PR #16701, opened 2018-02-08
add `configurable: true` to browserwindow.loadSettingsJSON

philipfweiss:
### Description of the Change

Some packages, including Facebook's Nuclide, rely on monkey-patching browserWindow.loadSettingsJSON. Recent changes to the master branch of atom (the addition of the ConfigFile class / related code) changed the behavior of this property, adding a getter using Object.defineProperty. 

With Object.defineProperty, `configurable` defaults to false, and it is impossible for packages to monkeypatch loadSettingsJSON. At the moment, this is a breaking change in Nuclide. Adding the `configurable: true` flag fixes this regression and allows us to redefine the getter.

### Alternate Designs

Ideally we would like to fix the regression with only changes to Nuclide, but we could not find another way to monkeypatch browserWindow.loadSettingsJSON, without adding `configurable: true` to Object.defineProperty. If someone knows of an approach we would love to hear! 

To fix a regression caused by a recent update to core.

### DIFF
--- a/src/main-process/atom-window.js
+++ b/src/main-process/atom-window.js
@@ -57,8 +57,9 @@ class AtomWindow extends EventEmitter {
 
     Object.defineProperty(this.browserWindow, 'loadSettingsJSON', {
       get: () => JSON.stringify(Object.assign({
-        userSettings: this.atomApplication.configFile.get()
-      }, this.loadSettings))
+        userSettings: this.atomApplication.configFile.get(),
+      }, this.loadSettings)),
+      configurable: true
     })
 
     this.handleEvents()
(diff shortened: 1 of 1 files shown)


margrettt (approved):

margrettt (approved):

maxbrunsfeld:
> Some packages, including Facebook's Nuclide, rely on monkey-patching browserWindow.loadSettingsJSON.

At some point, it'd be good to talk about what the monkey patch is for and add some kind of public API to take its place, but I'm cool with this if it fixes an immediate problem.

Ben3eeE:
Should this be cherry picked to the 1.25-releases branch so it's included in the upcoming beta?

https://github.com/atom/atom/blob/714dbf170badba3a840b5dbab4a067d2c672cd7e/src/main-process/atom-window.js#L58-L62

PR #16731, opened 2018-02-14
Add custom properties to atom-workspace to track editor properties

wbinnssmith:
### Description of the Change

Rather than directly setting rules on `atom-text-editor`, declare several CSS custom properties on the `atom-workspace` and then reference them in equivalent styling for `atom-text-editor`. These track the editor's current font size, font family, and line height.

### Alternate Designs

Declaring a set of static rules at the same time as the dynamically injected ones feels weird. On the other hand, I attempted to define these statically in `text-editor.less` and this resulted in inconsistent application when reloading the editor.

Many packages can rely on this without bundling similar functionality for monitoring these changes. Additionally, in core we can use generic names like `--editor-font-size` without the need for something like a prefix.

### Benefits

Many packages are manually monitoring these changes and updating styles directly. This avoids situations like:

* [atom/github tracking styles](https://github.com/atom/github/blob/cab6f0d37c0deda6bb218bb64a37ed97254cd530/lib/github-package.js#L158-L172)
* [Atom IDE Outline tracking styles](https://github.com/facebook-atom/atom-ide-ui/blob/master/modules/atom-ide-ui/pkg/atom-ide-outline-view/lib/OutlineView.js#L75-L85)

### Possible Drawbacks

We may move to custom properties in the long run for theming, and this might commit us to an API we may want to change later. Other than that, none I can think of. 

### Verification Process

Reload Atom and verify the three text editor properties. Change each at runtime through settings and verify each is reflected.

### Applicable Issues

### DIFF
--- a/src/workspace-element.js
+++ b/src/workspace-element.js
@@ -58,7 +58,7 @@ class WorkspaceElement extends HTMLElement {
   updateGlobalTextEditorStyleSheet () {
     const styleSheetSource = `atom-workspace {
   --editor-font-size: ${this.config.get('editor.fontSize')}px;
-  --editor-font-family: ${this.config.get('editor.fontFamily')}, monospace;
+  --editor-font-family: ${this.config.get('editor.fontFamily')};
   --editor-line-height: ${this.config.get('editor.lineHeight')};
 }`
     this.styleManager.addStyleSheet(styleSheetSource, {sourcePath: 'global-text-editor-styles', priority: -1})
--- a/src/workspace-element.js
+++ b/src/workspace-element.js
@@ -56,10 +56,16 @@ class WorkspaceElement extends HTMLElement {
   }
 
   updateGlobalTextEditorStyleSheet () {
-    const styleSheetSource = `atom-text-editor {
-  font-size: ${this.config.get('editor.fontSize')}px;
-  font-family: ${this.config.get('editor.fontFamily')};
-  line-height: ${this.config.get('editor.lineHeight')};
+    const styleSheetSource = `atom-workspace {
+  --editor-font-size: ${this.config.get('editor.fontSize')}px;
+  --editor-font-family: ${this.config.get('editor.fontFamily')};
+  --editor-line-height: ${this.config.get('editor.lineHeight')};
+}
+
+atom-text-editor {
+  font-family: var(--editor-font-family);
+  font-size: var(--editor-font-size);
+  line-height: var(--editor-line-height);
 }`
     this.styleManager.addStyleSheet(styleSheetSource, {sourcePath: 'global-text-editor-styles', priority: -1})
   }
--- a/src/workspace-element.js
+++ b/src/workspace-element.js
@@ -58,14 +58,8 @@ class WorkspaceElement extends HTMLElement {
   updateGlobalTextEditorStyleSheet () {
     const styleSheetSource = `atom-workspace {
   --editor-font-size: ${this.config.get('editor.fontSize')}px;
-  --editor-font-family: ${this.config.get('editor.fontFamily')};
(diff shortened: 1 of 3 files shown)


simurai:
> Declaring a set of static rules at the same time as the dynamically injected ones feels weird. On the other hand, I attempted to define these statically in `text-editor.less` and this resulted in inconsistent application when reloading the editor.

I tried ☝️ moving [these](https://github.com/atom/atom/blob/ebac04f729119ff55388d55058717849d061caf8/src/workspace-element.js#L65-L69) to `text-editor.less` and haven't noticed any issues. It might be that `text-editor.less` gets loaded before the `global-text-editor-styles`, resulting in showing the default font for a short time?

---

Should we move the current [default](https://github.com/atom/atom/blob/ebac04f729119ff55388d55058717849d061caf8/static/text-editor.less#L7) to the [settings](https://github.com/atom/atom/blob/ebac04f729119ff55388d55058717849d061caf8/src/config-schema.js#L375)? Otherwise using `var(--editor-font-family)` will have an empty value.

src/workspace-element.js
```diff
@@ -56,10 +56,16 @@ class WorkspaceElement extends HTMLElement {
...
-  font-family: ${this.config.get('editor.fontFamily')};
-  line-height: ${this.config.get('editor.lineHeight')};
+    const styleSheetSource = `atom-workspace {
+  --editor-font-size: ${this.config.get('editor.fontSize')}px;
+  --editor-font-family: ${this.config.get('editor.fontFamily')};
+  --editor-line-height: ${this.config.get('editor.lineHeight')};
+}
+
+atom-text-editor {
+  font-family: var(--editor-font-family);
```
simurai: We could add a fallback, so that while you type in the settings or mispell, it doesn't show `Times New Roman` and make the cursor jump. 

font

Maybe just `font-family: var(--editor-font-family), monospace;`

font2

wbinnssmith:
Awesome ideas. Just incorporated all of them :)

src/workspace-element.js
wbinnssmith: :)

static/text-editor.less
```diff
@@ -2,6 +2,11 @@
 @import "octicon-utf-codes";
 @import "octicon-mixins";
 
+:root {
+  // Fixes specs
+  --editor-font-family: Menlo, Consolas, 'DejaVu Sans Mono', monospace;
```
simurai: Looks like this change causes a [lint error](https://travis-ci.org/atom/atom/builds/342625366#L1137):

```
static/text-editor.css:1 Rule is empty. (empty-rules)
```

Could it be because the linter is old and doesn't know about custom properties?

simurai: We might could add some dummy property. Just so the linter doesn't think it's empty, but this all feels like a big hack. :see_no_evil:

Arcanemagus: Oh _wow_, `csslint` is the JSCS of CSS linters: It hasn't been updated in years.

I would _highly_ recommend looking into https://stylelint.io/ to replace that, it's based on PostCSS and is highly extensible. Think of it as the ESLint of style languages.

wbinnssmith:
cc @BinaryMuse this should allow you to clean up tracking this state in the GH package 🤩

simurai (approved):

idleberg:
Is there a reason why only `--editor-font-family` has been added? Wouldn't it make sense to add CSS properties for all of the [LESS variables](https://github.com/atom-community/ui-theme-template/blob/d6ab40b831ec3a1efe17d8664fbfb5c813209d62/styles/ui-variables.less) currently in use? I think the two could co-exist for a considerable time, eventually the LESS variables could be phased out.

PR #16748, opened 2018-02-15
return value of sync confirm function

UziTech:
### Description of the Change

return the value of a callback function for the synchronous version of `atom.confirm` 

this is returned for Atom v1.24

### Benefits

restores backwards compatibility

### Possible Drawbacks

none


50Wliu:
Yep, there was an implicit Coffeescript return. /sigh

Issue #16743, opened 2018-02-14
Fixing scroll sensitivity on macOS trackpad


billyjanitsch:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Following up on #15567 and #16114 now that they've landed in a stable release.

The default scroll sensitivity on a macOS trackpad now now feels much closer to native scrolling! (yay) But I notice two issues:
1. Editor scrolling now feels _more_ sensitive than native scrolling, with the default sensitivity.
2. There's some visual jank while scrolling -- much more than in a browser.

I think (1) is partially caused by [this bit of code](https://github.com/atom/atom/blob/7e0635d2c0d4b50cb76e4eccd619ab1041e5c86d/src/text-editor-component.js#L1524), which resolves each scroll delta to a minimum of 1px. Trackpads fire scroll events very often, so unless you're scrolling very quickly, the deltas can easily be less than a pixel, especially after being multiplied by the sensitivity (default 0.4).

Scaling up these small deltas makes scrolling feel weird, because it prevents the trackpad from distinguishing between velocities greater than 0 but less than a certain value. What's the rationale for this scaling? I thought it might be to align everything to a pixel grid, but that doesn't seem for happen for anything larger than 1px.

Another factor in (1) is the use of `wheelDelta` over `delta`, introduced in #16114. It seems like on a macOS trackpad, the former is exactly 3x the latter, or 1.2x when multiplied by the default scroll sensitivity (0.4). This makes me think that `delta` is actually the correct value -- the default sensitivity is what's throwing things off. Sure enough, using `delta` with a sensitivity of 1 (or emulating this by setting the sensitivity to 0.33 with the current `wheelDelta` implementation) feels _much_ more natural.

[...]

PR #16753, opened 2018-02-15
Stop scaling up small scroll wheel events

billyjanitsch:
### Description of the Change

As discussed in #16743, this reverts #15487.

#15487 essentially compensated for the fact that `wheelDelta` was switched to `delta` in the 1.19 editor rewrite, reducing the scroll sensitivity by 1/3. Now that it's been switched back in 1.24, #15487 is no longer necessary, and in fact it stops the trackpad from distinguishing between different small velocities.

Per #16743, it would actually be best to move back to `delta` eventually, but only when the interaction with scroll sensitivity is properly figured out, such that #15487 isn't necessary either way.

cc @Ben3eeE @50Wliu @nathansobo

### Benefits

This makes scrolling at low velocities feel slightly more natural on a macOS trackpad.

### Possible Drawbacks

This should be safe because it only affects small scroll deltas which are only generated by trackpads.

### Verification Process

Opened an editor window, pasted a large block of text, then scrolled around slowly using a macOS trackpad at different scroll sensitivities.

### Applicable Issues

See #16743 and #15567.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -1368,40 +1368,6 @@ describe('TextEditorComponent', () => {
       }
     })
 
-    it('always scrolls by a minimum of 1, even when the delta is small or the scroll sensitivity is low', () => {
-      const scrollSensitivity = 10
-      const {component, editor} = buildComponent({height: 50, width: 50, scrollSensitivity})
-
-      {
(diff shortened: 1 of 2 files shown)


nathansobo:
I'm really hoping this doesn't have unintended consequences, but I think your thought process is reasonable.

50Wliu:
@nathansobo do you think this should be hotfixed?

Issue #16747, opened 2018-02-15
atom.textEditors.getGrammarOverride incorrectly reports override for non-overridden files


Alhadis:
### Description

See file-icons/atom#704. This regression was introduced in Atom 1.24.0.

### Steps to Reproduce

1. Create a new file
2. Save as `foo.js`
3. Open developer tools, and run the following snippet:
~~~js
atom.textEditors.getGrammarOverride(atom.workspace.getActiveTextEditor());
~~~

**Expected behaviour:** A blank string should be returned.

**Actual behaviour:** The scope of the current grammar is returned, even though the file's grammar wasn't manually overridden by the user.

**Reproduces how often:** 100%

### Versions

~~~console
λ uname -a
Linux alhadis-P55-UD4P 4.13.0-32-generic #35-Ubuntu SMP Thu Jan 25 09:13:46 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
~~~

~~~console
λ atom --version | xclip
Atom    : 1.24.0
Electron: 1.6.16
Chrome  : 56.0.2924.87
Node    : 7.4.0
~~~

~~~console
λ apm --version | xclip
[31mapm[39m  [31m1.18.12[39m
[32mnpm[39m  [32m3.10.10[39m
[34mnode[39m [34m6.9.5[39m [34mx64[39m
[36matom[39m [36m1.24.0[39m
[33mpython[39m [33m2.7.14[39m
[35mgit[39m [35m2.14.1[39m
~~~

~~~console
λ apm --version; # Manually selecting with mouse + copying
apm  1.18.12
npm  3.10.10
node 6.9.5 x64
atom 1.24.0
python 2.7.14
git 2.14.1
~~~

PR #16754, opened 2018-02-15
Only return overridden language ids in .getGrammarOverride

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/16747


Alhadis:
Thanks for getting a fix out so quickly, @maxbrunsfeld! :tada:

PR #16755, opened 2018-02-15
Upgrade electron to 1.8.6

ungb:
Initial PR for Electron upgrade to 1.8.6.  

Fixes https://github.com/atom/settings-view/issues/924


damieng:
Definitely appear to be some issues here that the new Electron is causing - rather than CI failures.  This will require more investigation from the team - I've not been involved in any of the previous Electron upgrades so probably not the best person to dig into this solo.

ungb:
Yeah, same here. I wanted to get this PR up as I was talking to electron team it's good to identify these issues early on.   Wonder if @ckerr can help with this.  


I think @as-cii and @nathansobo helped with the last electron upgrade and might have some insight on these failures?

swrobel:
Interesting to know that the electron & atom teams are now two separate entities. I always figured they were one in the same. Regardless, looking forward to this upgrade!

yzgyyang:
Looking forward to this upgrade. How's it going?

jasonrudolph:
Now that #17273 has merged, we're running on Electron 2.0.0, so I'm gonna close this out.

PR #16791, opened 2018-02-20
Add `StartupWMClass` to .desktop for easier window-app association

Evpok:
### Description of the Change

Just adds a `StartupWMClass` entry to `atom.desktop`. This helps window managers associating Atom's window with the .desktop file that started Atom. It should help with some issues relative atom in docks, such as missing or duplicated icons.

### References

  - [Startup notification spec](https://specifications.freedesktop.org/startup-notification-spec/startup-notification-0.1.txt)

### DIFF
--- a/resources/linux/atom.desktop.in
+++ b/resources/linux/atom.desktop.in
@@ -8,3 +8,4 @@ Type=Application
 StartupNotify=true
 Categories=GNOME;GTK;Utility;TextEditor;Development;
 MimeType=text/plain;
+StartupWMClass=Atom


iolsen:
🎉 thanks for the contribution!

shadow-light:
Did this work for anyone? I had to make "Atom" lowercase for it to work for me. Prior to that my atom windows would have no icon if launched from a terminal etc.

I'm happy to submit a new pull request to change this, as long as it won't stop it working for other users. I'm on Ubuntu 16.04

Evpok:
@shadow-light Both cases work for me, which they should, since xprop gives `WM_CLASS(STRING) = "atom", "Atom"` for atom windows. Can you check what it gives for you ?

shadow-light:
Hi @Evpok, yes that's the same for me, but one of those is the class name and one is the class. I recently discovered this via xdotool. For example, with Firefox:

```shell
~ xprop | grep WM_CLASS
WM_CLASS(STRING) = "Navigator", "Firefox"
~ xdotool search --onlyvisible --class "Navigator"
~ xdotool search --onlyvisible --classname "Navigator"
62914577
~ xdotool search --onlyvisible --class "Firefox"
62914577
~ xdotool search --onlyvisible --classname "Firefox"
~ 
```

So it would seem that application icons are matched via the classname rather than the class. Note, you can't actually test this out with xdotool for Atom as it does a case-insensitive search.

I double checked and atom definitely needs to be lowercase for my system to match it. If both work on yours, then perhaps it would be safe to switch it to lowercase then?

Evpok:
Sure, if it fixes the issue on your side, open a PR, by all means :-)

PR #16803, opened 2018-02-21
Replace csslint with stylelint

wbinnssmith:
## Description of the Change
Includes the ["standard"](https://github.com/stylelint/stylelint-config-standard) config (no relation to the js formatting style) which is a bunch of accepted best-practices. Explicitly disables failing rules. Those that I think would be useful in the future I've marked with a `TODO`, at which point we can refactor rule-by-rule and remove the blacklist.

## Alternate Designs
Leaving csslint in place, but it is aging and the project wants to merge with stylelint anyway.

## Why Should This Be In Core?
To unblock #16731, which can't land because it won't pass lint due to csslint being unable to parse css custom properties.

## Benefits
A healthy ecosystem of rules, a modern parser and infrastructure for linting stylesheets.

## Verification Process
Remove some of the rule blacklisting in `stylelint.config.js` and verify linter errors show up in `./script/lint`


wbinnssmith:
cc @simurai @matthewwithanm @Arcanemagus

script/lib/lint-less-paths.js
```diff
@@ -1,64 +1,33 @@
...
+  return stylelint
+    .lint({
+      files: path.join(CONFIG.repositoryRootPath, 'static/**/*.less'),
+      configBasedir: __dirname,
+      configFile: path.resolve(__dirname, '..', '..', 'stylelint.config.js')
+    })
+    .then(({results}) => {
+      return flatten(
+        results.filter(_ => _.errored).map(result => {
+          const errors = result.warnings.filter(_ => _.severity === 'error')
```
Arcanemagus: This is explicitly ignoring all non-`error`s `warnings`.

In addition the other results, including `invalidOptionWarnings`, `deprecations`, and `ignored`, are also being ignored.

Is that what is intended here?

Arcanemagus (commented):
Other than the several data types missing from the report noted below, this looks like it should work.

script/lib/lint-less-paths.js
wbinnssmith: Yeah, this is intentional. I didn't think any of those other types of results warranted failing the build.

I'll push an update that logs these to `stdout` without going through the lint infrastructure (which doesn't handle severity levels). What do you think?

Arcanemagus: > Yeah, this is intentional. I didn't think any of those other types of results warranted failing the build.

Excellent point, I forgot that what this plugs into has no severity levels 😆.

> I'll push an update that logs these to stdout without going through the lint infrastructure

Makes sense to me, that way they would at least be logged somewhere, and if somebody happens to catch them or spot them when looking at a failing build they can get fixed.

I just didn't want things like `deprecations` to be silently completely ignored until they turned into errors later on 😉.

Arcanemagus (approved):
The `stylelint` part at least looks good.

PR #16809, opened 2018-02-21
Serialize settings as JSON when sending them to the main process

### LINKED ISSUE #16789: Color settings can't pass to the main process
### Prerequisites

### Description

It looks like `color` type values aren't being synced properly to the main process after https://github.com/atom/atom/pull/16628 was merged.

### Steps to Reproduce

1. Open two windows of Atom v1.25.0-beta0+
2. Have a package that has a `type: color` setting (eg. `atom-material-ui`)
3. Open Developer Tools on the main window
4. Open theme settings in the second window, and open the `atom-material-ui` config
5. Change one of the colors in the second window
6. Watch the first window fail to validate the color

**Expected behavior:**

The `color` setting properly transfers to the main window.

**Actual behavior:**

The following `Error` is thrown:
```
Error: Validation failed at atom-material-ui.colors.abaseColor, {"_alpha":1,"_blue":131,"_green":20,"_red":117} cannot be coerced into a color
    at Function.coerce (C:\temp\atom\src\config.js:1379)
    at Function.executeSchemaEnforcers (C:\temp\atom\src\config.js:389)
    at Function.coerce (C:\temp\atom\src\config.js:1339)
    at Function.executeSchemaEnforcers (C:\temp\atom\src\config.js:389)
    at Function.coerce (C:\temp\atom\src\config.js:1339)
    at Function.executeSchemaEnforcers (C:\temp\atom\src\config.js:389)
    at Config.makeValueConformToSchema (C:\temp\atom\src\config.js:1136)
    at Config.set (C:\temp\atom\src\config.js:702)
    at transact (C:\temp\atom\src\config.js:953)
    at Config.transact (C:\temp\atom\src\config.js:815)
    at Config.resetUserSettings (C:\temp\atom\src\config.js:950)
    at disposables.add.applicationDelegate.onDidChangeUserSettings.settings (C:\temp\atom\src\atom-environment.js:786)
    at EventEmitter.outerCallback (C:\temp\atom\src\application-delegate.js:194)
    at emitThree (events.js:121)
    at EventEmitter.emit (events.js:197)
```

From what I can tell so far the bug is that the [`Color`](https://github.com/atom/atom/blob/f866978d119e48734c82cad4a685ed7c9261d04f/src/color.js) Object from the second window is being `emit`ed to the main window... but the problem is that it only has `_name` properties defined _directly_ on it, the main `name` properties are all `get`ters. This means that when it gets to the main window all that is left is an Object like `{"_alpha":1,"_blue":192,"_green":107,"_red":92}` instead of a real `Color` object. Since that Object isn't a `Color` object, it fails the schema enforcer for a `color` setting.

**Reproduces how often:**

100% of the time.

### Versions

This started happening in this version, based on https://github.com/atom/atom/commit/e74ea4797521c845f040638d01d41a1677c9ddb8:
```
Atom    : 1.25.0-dev-e74ea4797
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
```

A build based on https://github.com/atom/atom/commit/a48624dc0255458a4bd73ffca02a545080a9d34a (right before #16628 was merged) **doesn't** show this issue:
```
Atom    : 1.25.0-dev-a48624dc0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
```

OS: Windows 10 x64

### Additional Information

This was originally filed at https://github.com/atom-material/atom-material-ui/issues/462, but further investigation showed this to be a bug here.

maxbrunsfeld:
Fixes #16789

Config settings of type `Color` should be written to the config file as a string like `#af01cd`. To achieve this, the class [overrides `toJSON` ](https://github.com/atom/atom/blob/1d9d17bc3935246789fdb2d8fed40286431a2bb7/src/color.js#L87-L89).

When I refactored the config system to IPC settings to the main process for writing, this behavior was inadvertently broken because sending a value over electron's IPC does not serialize it as JSON; it uses its own serialization mechanism that does not preserve objects' `toJSON` methods.

### DIFF
--- a/src/application-delegate.js
+++ b/src/application-delegate.js
@@ -182,7 +182,7 @@ class ApplicationDelegate {
   async setUserSettings (config) {
     this.pendingSettingsUpdateCount++
     try {
-      await ipcHelpers.call('set-user-settings', config)
+      await ipcHelpers.call('set-user-settings', JSON.stringify(config))
     } finally {
       this.pendingSettingsUpdateCount--
     }
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -559,7 +559,7 @@ class AtomApplication extends EventEmitter {
     }))
 
     this.disposable.add(ipcHelpers.respondTo('set-user-settings', (window, settings) =>
-      this.configFile.update(settings)
+      this.configFile.update(JSON.parse(settings))
     ))
 
     this.disposable.add(ipcHelpers.respondTo('center-window', window => window.center()))


Arcanemagus:
Thanks! I dug far enough to figure out _why_ it was breaking, but not _where_ 😆.

maxbrunsfeld:
@Arcanemagus Yeah, thanks so much for the info in your report. That made it very easy to fix!

PR #16814, opened 2018-02-21
Provide a default scope descriptor if language mode doesn't provide them

### LINKED ISSUE #290: Missing stub method 'scopeDescriptorForPosition' in NullLanguageMode
Atom's autocomplete-plus provides a watch editor feature that lets us attach autocompletion to any text editor, which previously worked even when the editor had null grammar and its buffer did not set a language mode.

In Atom 1.24 if we attach autocomplete to an editor whose buffer uses the NullLanguageMode, autocomplete-plus fails because it cannot find a scope for the cursor position, as it [calls](https://github.com/atom/autocomplete-plus/blob/8f192dc3d8d5c19cf111db122d1697c7a0774aa0/lib/autocomplete-manager.js#L245) the method [scopeDescriptorForPosition](https://git.io/vNRUi) on the language mode.

One fix would be to add `scopeDescriptorForPosition(_) { return { scopes: [] } }` to the null-language-mode.js methods.

maxbrunsfeld:
Fixes https://github.com/atom/text-buffer/issues/290

/cc @pelmers I decided to fix the problem in the `TextEditor` class rather than adding the method in the `NullLanguageMode` because currently, the *LanguageMode* interface already has a lot of optional methods that are only used in the `TextEditor` layer (and non in `text-buffer` itself).

### DIFF
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -6747,6 +6747,14 @@ describe('TextEditor', () => {
     editor.destroy()
   })
 
+  describe('.scopeDescriptorForBufferPosition(position)', () => {
+    it('returns a default scope descriptor when no language mode is assigned', () => {
+      editor = new TextEditor({buffer: new TextBuffer()})
+      const scopeDescriptor = editor.scopeDescriptorForBufferPosition([0, 0])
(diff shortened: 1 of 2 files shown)


pelmers:
thanks, this change looks good to me

PR #16832, opened 2018-02-25
Add a new keybinding for window:reload Fixes #1794

cvx:
### Description of the Change

This commit adds a new (and default) keybinding for "View/Developer/Reload Window" that isn't in conflict with 1Password's keybindings.

### Alternate Designs

N/A?

It's not unreasonable to presume that there's a large percentage of 1Password users among Atom developer community, so it would be beneficial fix this binding clash.

### Benefits

See above.

### Possible Drawbacks

I've left previous binding in, so users unaffected by this problem won't have to adjust.
It could conflict with other applications or Atom packages, but I've not experienced any issues.

### Verification Process

I've been using both keybindings in my own keymap for a while now.

### Applicable Issues

Issue #1794 (while closed, the it is still relevant).

### DIFF
--- a/keymaps/darwin.cson
+++ b/keymaps/darwin.cson
@@ -27,6 +27,7 @@
   'left': 'core:move-left'
   'right': 'core:move-right'
   'ctrl-alt-cmd-l': 'window:reload'
+  'ctrl-alt-cmd-k': 'window:reload'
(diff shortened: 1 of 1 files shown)


alissonpelizaro (commented):
Approved

PR #16845, opened 2018-02-27
Add the concept of an `atomproject` file to atom. 

philipfweiss:
### Description of the Change

Adds explicit project configs to Atom. Ie:  One can now create a `*.atomproject.{json, cson}` containing an array of root paths, as well as an associated config. Opening this file from the command line with the -p flag (ie: `atom -p ./foo.atomproject.json` will, in a new window, open up all the paths specified in the file, as well as layer the configurations specified in the file in memory on top of the global config, at a level of precedence higher than settings from ~/atom/config.json. Atom windows not opened through the -p flag will not have any of the project specific settings. 

This PR does not add the ability to launch atom from a project file through the UI, although it would be fairly simple to implement using the APIs (ie: `atom.project.replace`) added in this PR. 

Please note that this PR is separate from (RFC-001](#16698). As @damieng has stated elsewhere, this RFC will be tweaked to be for the root-folder level config that LSP needs which IDE team will develop. That RFC will only deal with root folder configs (not project level). 

Example of launching an atomproject on Giphy:
https://giphy.com/gifs/1wqpNX1QdxAflmELbh/fullscreen

EDIT: Project configs must now match personal configs - ie: the config in the above gif would need to be `{'*': {'core': {'themes': ['one-light-ui', 'one-light-syntax']}}}`

### Alternate Designs

When a user invokes` atom -p. *.atomproject.json`, `command-line-parser.js` checks for a -p flag, and if it finds it, passes the data from the file over the process boundary to the newly created atom window. There, it calls `atom.project.replace` with the contents of this file. This API, `atom.project.replace`, uses `project.setPaths` to replace the paths on the atom window. An alternative way to do this could have been to use the `pathsToOpen` array in `parse-command-line.js`. Doing this would allow other command line options  (ie: --test) to use the paths field of the atomproject file as well. I chose not to use this approach, and instead to just use `project.setPaths`. This is so that `replace` does all of the work of replacing an atomproject, instead of just doing most of it and delegating the paths bit to the Atom application process launcher. Additionally, this design simplifies the task of implementing UI to swap atomprojects. 

In terms of layering project settings on top of the global config, this PR takes a very different approach to an earlier PR which was meant to add root-folder level config. First, there is no "dirty state" (settings that have been set within the current atom session), or "settingsContexts" (seperate in-memory settings objects) - instead, project configs are just implemented using the existing source properties on the config settings. No settings are saved to disk, and only layer on top of the atom window spawned for the project. 

This change allows users to set config settings for a set of paths, which is useful for teams or individuals who want settings specific to a particular project. It also lays the groundwork for features like suggested packages (which could live inside a .atomproject file). 

### Verification Process

Unit testing in config-spec, project-spec, and extensive manual testing of atomproject files. 

### Applicable Issues

(RFC-001)(#16698)


src/main-process/atom-application.js
```diff
@@ -93,14 +93,14 @@ class AtomApplication extends EventEmitter {
...
     this.getLastFocusedWindow = this.getLastFocusedWindow.bind(this)
-
     this.resourcePath = options.resourcePath
     this.devResourcePath = options.devResourcePath
     this.version = options.version
     this.devMode = options.devMode
     this.safeMode = options.safeMode
     this.socketPath = options.socketPath
     this.logFile = options.logFile
+    this.projectSettings = options.projectSettings
```
maxbrunsfeld: I wouldn't think that we'd need to store the project settings on the entire `AtomApplication`; I would think that they would only apply to a particular window, kind of like the `pathsToOpen` options. Do we need this property?

src/main-process/atom-window.js
```diff
@@ -152,6 +151,16 @@ class AtomWindow extends EventEmitter {
...
 
+  loadDataOverProcessBoundary () {
+    Object.defineProperty(this.browserWindow, 'loadSettingsJSON', {
+      get: () => JSON.stringify(Object.assign({
+        userSettings: this.atomApplication.configFile.get(),
+        projectSettings: this.projectSettings
+      }, this.loadSettings)),
+      configurable: true
+    })
+  }
```
maxbrunsfeld: The name `loadDataOver...` makes it sound like this method itself is going to load or retrieve some data, when really it just defines an accessor method that allows the window process to retrieve data later. It's hard to give this weird little code path a name; I might just leave the code un-abstracted in the constructor.

Another though - should we just define this getter using `get` syntax as a method? I can't remember why it needs to be configurable.

src/main-process/parse-command-line.js
```diff
@@ -52,6 +53,7 @@ module.exports = function parseCommandLine (processArgs) {
     'When in test mode, waits until the specified time (in minutes) and kills the process (exit code: 130).'
   )
   options.alias('v', 'version').boolean('v').describe('v', 'Print the version information.')
+  options.alias('p', 'atomproject').describe('p', 'Start atom with an atomproject file.')
```
maxbrunsfeld: I'd change the long-form name of this argument to just `project`, rather than `atomproject`.

src/main-process/parse-command-line.js
```diff
@@ -177,7 +193,34 @@ module.exports = function parseCommandLine (processArgs) {
   }
 }
 
-function normalizeDriveLetterName (filePath) {
+const readProjectSettingsSync = (filepath, executedFrom) => {
+  if (!hasAtomProjectFormat(path.basename(filepath))) {
+    throw new Error('File must match format: *.atomproject.{json, cson}')
```
maxbrunsfeld: Since we have an explicit flag for specifying a project, I don't think we need to actually enforce that the filename matches any particular pattern. Thoughts?

src/project.js
```diff
@@ -77,6 +78,29 @@ class Project extends Model {
...
+    this.emitter.emit('did-replace', newSettings)
+  }
+
+  onDidReplace (callback) {
+    return this.emitter.on('did-replace', callback)
+  }
+
+  getProjectSettings () {
+    return this.projectSettings
+  }
```
maxbrunsfeld: Why does the `Project` need to store the settings, as opposed to them just being stored in the config?

src/config.js
```diff
@@ -589,8 +594,18 @@ class Config {
...
 
+    const noSources = !options.sources || (Array.isArray(options.sources) && options.sources.length === 0)
+    if (this.hasCurrentProject && noSources) {
+      const projectScope = Array.isArray(scope) ? scope.push(PROJECT) : [PROJECT]
+      const projectOptions = Object.assign({sources: [PROJECT]}, options)
+      value = this.getRawScopedValue(projectScope, keyPath, projectOptions)
+      if (value != null) {
+        return value
+      }
+    }
```
maxbrunsfeld: Can you explain why this logic :point_up: (and also the additions to `getAll` below) is needed?

I would think that `get` and `getAll` would not need to change much if at all; we'd just need to give the `PROJECT` source a higher priority, kind of like what [we do for the user settings](https://github.com/atom/atom/blob/7b490c66ec6a9b277c44db082383ea1c5f50a2f3/src/config.js#L1170-L1172).

maxbrunsfeld (commented):
Nice work @philipfweiss! I left some questions.

src/main-process/atom-window.js
philipfweiss: You're definitely right, I can just move it back. 

I think we changed it to configurable because Nuclide previously had to monkey-patch the property (https://github.com/atom/atom/pull/16701). We've since (within the last few days) fixed this so removing it sounds good to me!

src/main-process/parse-command-line.js
philipfweiss: I had the same thought while writing this - I'm fine with not enforcing a filename format!

src/project.js
philipfweiss: This is just so that third party packages can see which project is active (as well as which settings, paths, etc are on the project). I can change the way this is handled though if you think this shouldn't live on project.

src/config.js
```diff
@@ -360,6 +360,7 @@ const ScopeDescriptor = require('./scope-descriptor')
 // * Don't depend on (or write to) configuration keys outside of your keypath.
 //
 const schemaEnforcers = {}
+const PROJECT = '__atomproject__'
```
maxbrunsfeld: Could we just use the actual project file path as the `source`?

philipfweiss: That was my original implementation, but what if in the future configs sourced to individual files becomes common? Could that lead to collisions? 

Maybe I could do something like `project__${source}`?

src/project.js
matthewwithanm: i think the only thing we need here is the path to the file so that we can have UI to show the active project?

philipfweiss: Alright - then we can just grab it from config! (Or we can keep it on project, I'm not opinionated either way).

spec/config-spec.js
```diff
@@ -1840,4 +1835,86 @@ describe('Config', () => {
       expect(atom.config.get('do.ray')).toBe('me')
     })
   })
+
+  describe('project specific configs', () => {
```
maxbrunsfeld: Let the minor nitpicks begin! I have a few wording suggestions:

* I think `project specific configs` should be `project-specific settings` (here as well as in later `it/describe` strings)
* Can you drop the 'should' from the test descriptions (e.g. `should get` becomes `gets`)

spec/project-spec.js
```diff
@@ -274,6 +274,51 @@ describe('Project', () => {
     })
   })
 
+  describe('atomProject', () => {
```
maxbrunsfeld: I think the description here should just be `describe('.replace')`.

spec/project-spec.js
```diff
@@ -274,6 +274,51 @@ describe('Project', () => {
...
+      atom.project.replace(null)
+      projectPath1 = temp.mkdirSync('project-path1')
+      projectPath2 = temp.mkdirSync('project-path2')
+      newSettings = {
+        paths: [projectPath1, projectPath2],
+        originPath: "originPath",
+        config: {
+          "baz": "buzz"
+        }
+      }
```
maxbrunsfeld: I'd like to avoid using the word 'settings' for this object, since there already is a very closely-related concept of 'project-specific settings' meaning just the config settings. Since this object also contains the paths, maybe we should call it something more general. How about 'project specification', 'project state', or just 'project info'?

@matthewwithanm any naming suggestions?

src/atom-environment.js
```diff
@@ -210,7 +209,7 @@ class AtomEnvironment {
     this.blobStore = params.blobStore
     this.configDirPath = params.configDirPath
 
-    const {devMode, safeMode, resourcePath, userSettings} = this.getLoadSettings()
+    const {devMode, safeMode, resourcePath, userSettings, projectSettings} = this.getLoadSettings()
```
maxbrunsfeld: Same comment here about the word 'settings'.

src/config.js
```diff
@@ -954,24 +958,60 @@ class Config {
...
       this.settingsLoaded = true
-      for (let key in newSettings) { const value = newSettings[key]; this.set(key, value, {save: false}) }
+      for (let key in newSettings) {
+        const value = newSettings[key]
+        if (options.resetProject) {
+          this.set(key, value, {save: false, source: this.projectFile})
+        } else {
+          this.set(key, value, {save: false})
+        }
+      }
```
maxbrunsfeld: Just naming-wise, it seems a bit confusing to generalize `resetUserSettings` in this way and have `resetProjectSettings` call it. Could you instead extract a third method that they would both call?

Something like:

```js
resetUserSettings (newSettings) {
  this.settings = {}
  this._resetSettings(this.mainSource, newSettings)
}

resetProjectSettings (projectFile, newSettings) {
  this.projectSettings = {}
  this._resetSettings(projectFile, newSettings)
}

_resetSettings (source, newSettings) {
  newSettings = Object.assign({}, newSettings)
  if (newSettings.global != null) {
    newSettings['*'] = newSettings.global
    delete newSettings.global
  }
 
  if (newSettings['*'] != null) {
    const scopedSettings = newSettings
    newSettings = newSettings['*']
    delete scopedSettings['*']
    this.resetUserScopedSettings(scopedSettings, {save: false, source})
  }

  return this.transact(() => {
    for (let key in newSettings) { const value = newSettings[key]; this.set(key, value, {save: false, source}) }
    if (this.pendingOperations.length) {
      for (let op of this.pendingOperations) { op() }
      this.pendingOperations = []
    }
  })
}
```

I'm not sure exactly how the extraction needs to work, but it seems like you don't need all these conditionals *within* the method, since `set` should already do the right thing based on the `source`.

src/config.js
```diff
@@ -1168,15 +1211,22 @@ class Config {
...
   }
 
   emitChangeEvent () {
     if (this.transactDepth <= 0) { return this.emitter.emit('did-change') }
   }
 
-  resetUserScopedSettings (newScopedSettings) {
-    const source = this.mainSource
+  resetUserScopedSettings (newScopedSettings, options = {}) {
+    const source = options.source == null ? this.mainSource : options.source
```
maxbrunsfeld: Maybe we should rename this to `resetScopedSettings` since it's not necessarily the user's settings any more.

src/main-process/atom-window.js
```diff
@@ -57,9 +58,9 @@ class AtomWindow extends EventEmitter {
 
     Object.defineProperty(this.browserWindow, 'loadSettingsJSON', {
       get: () => JSON.stringify(Object.assign({
-        userSettings: this.atomApplication.configFile.get()
-      }, this.loadSettings)),
-      configurable: true
+        userSettings: this.atomApplication.configFile.get(),
+        projectSettings: this.projectSettings
+      }, this.loadSettings))
```
maxbrunsfeld: Now that this isn't configurable anymore, could we just remove it and define it as a regular getter method like

```js
constructor () {
  // ...
}

get loadSettingsJSON () {
  return JSON.stringify(/* ... */)
}
```

src/project.js
```diff
@@ -77,6 +77,23 @@ class Project extends Model {
     }
   }
 
+  // Layers the contents of an atomProject file's config
+  // on top of the current global config.
+  replace (newSettings) {
```
maxbrunsfeld: Same comment as above about the word 'settings'.

maxbrunsfeld (commented):
This is starting to look good. @matthewwithanm @nathansobo want to review?

src/main-process/parse-command-line.js
```diff
@@ -125,6 +128,7 @@ module.exports = function parseCommandLine (processArgs) {
     }
   }
 
+  // Check to see if atomproject flag is set, then add all paths from the .atom-project.
```
matthewwithanm: "atomproject" -> "project", ".atom-project" -> ".atomproject"

hm. now that i think about it, i guess the code doesn't care what the extension is…we should probably pick something (i vote for ".atomproject.json"), be consistent, and make sure that that's advertised as "the" extension in the docstrings and stuff.

src/main-process/parse-command-line.js
```diff
@@ -134,6 +138,18 @@ module.exports = function parseCommandLine (processArgs) {
     devMode = true
   }
 
+  let projectSettings = {}
+  if (atomProject) {
+    const contents = Object.assign({}, readProjectSettingsSync(atomProject, executedFrom))
+    const config = contents.config
+    const originPath = atomProject
+    const paths = contents.paths.map((curPath) =>
+      relativizeToAtomProject(curPath, path.dirname(path.join(executedFrom, atomProject))
```
matthewwithanm: nothing about this relativize method is Atom project-specific so the naming seems a little weird to me. also, is `executedFrom` absolute? if so, we could just do this:

```js
const pathToProjectFile = path.join(executedFrom, atomProject)
const base = path.dirname(pathToProjectFile)
const paths = contents.paths.map(curPath => path.resolve(base, curPath))
```

src/main-process/parse-command-line.js
```diff
@@ -134,6 +138,18 @@ module.exports = function parseCommandLine (processArgs) {
...
+    ))
+    pathsToOpen.push(path.dirname(atomProject))
```
matthewwithanm: what's this for?

src/main-process/parse-command-line.js
```diff
@@ -177,7 +193,25 @@ module.exports = function parseCommandLine (processArgs) {
   }
 }
 
-function normalizeDriveLetterName (filePath) {
+const readProjectSettingsSync = (filepath, executedFrom) => {
+  try {
+    const readPath = path.isAbsolute(filepath) ? filepath : path.join(executedFrom, filepath)
```
matthewwithanm: any reason to have this function do the path resolution instead of passing in the path already resolved? it seems like you're doing that work at the callsite anyway.

also, i could be wrong but i think the convention in the codebase is to favor function declaration (i.e. use the "function" keyword) for top-level functions like this, unless it's like a single line implicit return deal.

src/main-process/parse-command-line.js
```diff
@@ -177,7 +193,25 @@ module.exports = function parseCommandLine (processArgs) {
...
+    const contents = CSON.readFileSync(readPath)
+    if (contents.paths && contents.config) {
```
matthewwithanm: @maxbrunsfeld what do you think about making `paths` optional and just making it default to the directory of the file?

src/main-process/parse-command-line.js
```diff
@@ -177,7 +193,25 @@ module.exports = function parseCommandLine (processArgs) {
...
+      return contents
+    }
+    throw new Error()
```
matthewwithanm: throwing an empty error and then immediately catching it is kinda weird. you can just move the later throw out of the catch block:

```js
try {
    const contents = CSON.readFileSync(readPath)
    if (contents.paths && contents.config) return contents
} catch (err) {}
throw new Error(...)
```

or maybe even have two separate errors

matthewwithanm (commented):
this is looking pretty good! a few comments (:

src/main-process/parse-command-line.js
maxbrunsfeld: I think it would be ok to be extension-agnostic now. Why not open it up to any JSON or CSON file?

src/main-process/parse-command-line.js
maxbrunsfeld: :+1:

src/main-process/parse-command-line.js
matthewwithanm: i don't think the code has to care, but imo it's better to have a standard so that we can build tooling on top of it (find all atom project files, etc)

src/main-process/atom-window.js
philipfweiss: Hmm - aren't you defining this property on `this.browserWindow` rather than on the AtomWindow class? 

I think that Object.define was being used before `configurable: true` was being added.

src/main-process/parse-command-line.js
philipfweiss: Atom's logic expects a path to be sent in the pathsToOpen array. I pass this in so it knows it needs to open paths, and then call replace before it loads.

src/main-process/atom-window.js
maxbrunsfeld: Oh yeah you're right. If we wanted to make it a getter, we'd have to change other code to access it from the atom window instead of the browserWindow. Doable, but no need to do it as part of this PR.

src/main-process/atom-window.js
```diff
@@ -441,4 +442,7 @@ class AtomWindow extends EventEmitter {
   disableZoom () {
     return this.browserWindow.webContents.setVisualZoomLevelLimits(1, 1)
   }
+
+  get loadSettingsJSON () {}
```
matthewwithanm: remove?

src/main-process/parse-command-line.js
```diff
@@ -52,6 +53,7 @@ module.exports = function parseCommandLine (processArgs) {
     'When in test mode, waits until the specified time (in minutes) and kills the process (exit code: 130).'
   )
   options.alias('v', 'version').boolean('v').describe('v', 'Print the version information.')
+  options.alias('p', 'project').describe('p', 'Start atom with an project specification file.')
```
matthewwithanm: capitalize Atom?

src/main-process/parse-command-line.js
```diff
@@ -177,6 +193,22 @@ module.exports = function parseCommandLine (processArgs) {
...
+    const contents = CSON.readFileSync(readPath)
+
+    if (contents.paths == null) {
+      contents.paths = [path.dirname(readPath)]
+    }
+    if (contents.config) {
+      return contents
+    }
+  } catch (e) {}
+  const errorMessage = `Unable to read supplied project specification file. This file must have a valid array of paths, as well as a valid config object.`
```
matthewwithanm: backticks -> single quotes

matthewwithanm (commented):
minus these few tiny things, this is looking done to me! 😍

matthewwithanm (approved):
lgtm! let's get signoff from the Atom team 🙂

maxbrunsfeld (approved):
This looks great to me. Feel free to 🚢 at will (after any FB reviewers have weighed in).

src/main-process/parse-command-line.js
```diff
@@ -134,6 +138,18 @@ module.exports = function parseCommandLine (processArgs) {
     devMode = true
   }
 
+  let projectSpecification = {}
+  if (projectSpecificationFile) {
+    const contents = Object.assign({}, readProjectSpecificationSync(projectSpecificationFile, executedFrom))
+    const config = contents.config
```
wbinnssmith: nit: I would just inline this into the object literal on 150

src/main-process/parse-command-line.js
```diff
@@ -134,6 +138,18 @@ module.exports = function parseCommandLine (processArgs) {
...
+    const originPath = projectSpecificationFile
```
wbinnssmith: this one too, unless it's clearer to call this `originPath` in the next line as well

src/main-process/parse-command-line.js
```diff
@@ -177,6 +193,22 @@ module.exports = function parseCommandLine (processArgs) {
   }
 }
 
+function readProjectSpecificationSync (filepath, executedFrom) {
+  try {
+    const readPath = path.isAbsolute(filepath) ? filepath : path.join(executedFrom, filepath)
+    const contents = CSON.readFileSync(readPath)
```
wbinnssmith: can the `try`/`catch` be moved to just wrap the `readFileSync` call?

src/config.js
```diff
@@ -954,24 +963,58 @@ class Config {
...
+      }
       if (this.pendingOperations.length) {
         for (let op of this.pendingOperations) { op() }
         this.pendingOperations = []
       }
     })
   }
 
+  _clearUnscopedSettingsForSource (source) {
+    switch (source) {
```
wbinnssmith: nit: does this really need to be a switch? feels like an if/else would do.

wbinnssmith (approved):
Looks good to me! Left some nits, up to you :)

src/main-process/parse-command-line.js
```diff
@@ -177,6 +193,22 @@ module.exports = function parseCommandLine (processArgs) {
   }
 }
 
+function readProjectSpecificationSync (filepath, executedFrom) {
+  try {
+    const readPath = path.isAbsolute(filepath) ? filepath : path.join(executedFrom, filepath)
```
captbaritone: Should this conversion be moved to the caller?

src/main-process/parse-command-line.js
```diff
@@ -177,6 +193,22 @@ module.exports = function parseCommandLine (processArgs) {
...
+    const contents = CSON.readFileSync(readPath)
+
+    if (contents.paths == null) {
+      contents.paths = [path.dirname(readPath)]
+    }
+    if (contents.config) {
```
captbaritone: So, if `contents` does not have a `config` property, we return undefined? Then we are relying on the fact that `Object.assign({}, undefined)` makes an empty object?

Seems like it would be better to either explicitly return an empty object, or throw.

src/config.js
```diff
@@ -954,24 +963,58 @@ class Config {
...
+      case (this.projectFile):
+        this.projectSettings = {}
+        return
+      default:
+        this.settings = {}
+    }
```
captbaritone: Is there a reason this is a switch statement and not just an if/else?

src/config.js
```diff
@@ -954,24 +963,58 @@ class Config {
...
+    this.resetProjectSettings({}, null)
+  }
+
   getRawValue (keyPath, options = {}) {
     let value
     if (!options.excludeSources || !options.excludeSources.includes(this.mainSource)) {
       value = getValueAtKeyPath(this.settings, keyPath)
+      if (this.projectFile != null) {
+        const projectValue = getValueAtKeyPath(this.projectSettings, keyPath)
+        value = projectValue || value
```
captbaritone: Are there not cases where we would want a "falsy" projectValue to trump the settings value?

captbaritone (commented):
Looks good overall. A few small questions.

src/config.js
philipfweiss: I left it as a switch just in case more source settings are added in the future (which might need to happen for path specific configs)- but since there are only two an if-else would definitely do!

src/main-process/parse-command-line.js
philipfweiss: If contents.config doesn't exist, it falls through to the error and does not return undefined.

src/main-process/parse-command-line.js
matthewwithanm: hey I asked for this too @philipfweiss!

src/config.js
philipfweiss: Great catch!

src/main-process/parse-command-line.js
matthewwithanm: ooh I don't think we should require that to be defined. good catch!

src/main-process/parse-command-line.js
philipfweiss: We want to immediately fall through and error as soon as the file is not found. If the other code is moved outside of the try/catch block, then if the read fails, it is executed before the error is thrown (and may result in another error itself: ie: `cannot read property of undefined`, rather than the error message we want). I could move it outside and add more null checks, but I think  this approach is cleaner.

src/main-process/parse-command-line.js
```diff
@@ -177,6 +193,22 @@ module.exports = function parseCommandLine (processArgs) {
...
+      return contents
+    }
+  } catch (e) {}
+  const errorMessage = 'Unable to read supplied project specification file. This file must have a valid array of paths, as well as a valid config object.'
```
matthewwithanm: gotta update this message too. both of these should be optional.

Alhadis:
Sorry I'm late. ❤️Took me a full six days to notice this and add the new format to everybody's favourite file-icon package.

Oh yeah, I also [updated Atom's manpage](https://rawgit.com/Alhadis/0211a1cfc77e1ff4bd5a4f8f5e32b0d4/raw/5a8f525f84d8d56efcdb4231187cf4a06a1dc1ac/atom.1.html). You're welcome.

jianhuafan:
should this line "pathsToOpen.push(path.dirname(projectSpecificationFile))" be changed to "pathsToOpen.push.apply(pathsToOpen, contents.paths)", cause when I open a new window it didn't open the path specified by the "paths" variable in "*.atomproject.cson", instead opened the path that the "*.atomproject.cson is at.

manuelbieh:
I honestly can't think of a use case where I would want to define files to open in a (project specific) config file. Why would anyone do that? 

I have a very specific and current other use case though: I've just joined a new team a week ago and they're - for whatever reason - using the .md file suffix for their handlebars templates. I'd love to have a way to simply add a `.atomconfig.cson` to that project folder with:
```
'*':
  'file-types':
    'md': 'text.html.handlebars'
```
Which always gets automatically loaded whenever I open that project. No matter if via Dock, via Project Manager or Recent Files 😕

When talking about predictability: I'd throw consistency into the ring. I want to have always the same config when opening a project folder, no matter which way I open that project 😉

Alhadis:
> I honestly can't think of a use case where I would want to define files to open in a (project specific) config file.

That was just an example. My point is that it goes against user expectations. There could be any number of reasons why an unfamiliar config file resides in a project directory - it could be a folder of test fixtures, files downloaded *en masse* from GitHub searches (it [*does* happen...](https://github.com/Alhadis/Harvester/)), or even a confused user who accidentally checked their config into a package's version control and forgot about it (and never noticed).

> I want to have always the same config when opening a project folder, no matter which way I open that project

Back to that ol' *"programs should not be smart-asses"* thing again. =) If a user *didn't* want to load a project's config when opening a directory, they'd have no obvious way of saying *"No Atom, **don't** load that damn file!"*

AlexWayfer:
@Alhadis What about some notification like "I found project file and loaded it" with button "No, don't!"?

Alhadis:
That's going way too far...

If automatic loading of project-configs are desired, I'd recommend setting something like this in your environment:

~~~shell
export ATOM_AUTOLOAD_PROJECTS=1
~~~

Alternatively, `$ATOM_AUTOLOAD_PROJECTS` could be set to a list of paths to restrict "autoloading" to (using the same `:` delimited format as `$PATH`). So to always load `.atomproject.*` configs from any directories located inside `~/Projects` or `~/.atom`, you could do:

~~~shell
export ATOM_AUTOLOAD_PROJECTS="$HOME/Projects:$HOME/.atom"
~~~

Or just set it to `1` to unconditionally enable the behaviour if the user desires.

djh329:
These settings will override global settings. Is this a problem for functionality such as zoom. From my quick playing around with this, it seems that the zoom function no longer works if I override `fontSize` in the editor configuration. This is because the project level configuration overwrites it. Should project configuration have listeners for these types of things

YurySolovyov:
Is there a way to save current project layout as atomproject config?

YurySolovyov:
@lee-dom is there a ticket I can follow?

lee-dohm:
No, and I would be against a change like that since the concept of the `atomproject` file is to store project-wide settings that are shared between all people working on the project.

lierdakil:
`Project.onDidReplace` is left undocumented, and it's doesn't seem to be used anywhere, so it looks to be intended as public. I assume it's an oversight?

Issue #16858, opened 2018-02-28
'await' may be missing at async function call


jerry88kim:
### Description
Hi, I am a developer of JavaScript code quality service called DeepScan(https://deepscan.io).

We analyzed atom project's code with DeepScan for benchmarking, and found some issues which would be better to modify it.

I will report on the issues we found below and create pull request for fixing them.

And here is DeepScan's detailed result link for atom project.
https://deepscan.io/dashboard/#view=project&pid=2005&bid=9394&subview=issues

### Additional Information

Alarm 1: **Condition 'isOpenInEditor(realPath)' is always true because it evaluates to a 'Promise' object. The object is originated from the return value of the function defined at line 226.**
* Src: src/path-watcher.js(242:24-242:48)
    `this.subs.add(treeView.onEntryDeleted(async event => {`
    `  const realPath = await getRealPath(event.path)`
    `  if (!realPath || isOpenInEditor(realPath) /*alarm*/) return`
    `...})`
* Cause
  - 'isOpenInEditor()' is async function, and returns Promise object when calling without 'await', so the condition is always true.
  - 'isOpenInEditor()' function's body returns boolean, so maybe 'await' is unintendedly missing.
  - 'isOpenInEditor()' function code
    `const isOpenInEditor = async eventPath => {`
    `  const openPaths = await Promise.all(`
    `    atom.workspace.getTextEditors().map(editor => getRealPath(editor.getPath()))`
    `  )`
    `  return openPaths.includes(eventPath) // return boolean value`
    `}`

* Guide
  - Call 'isOpenInEditor()' with 'await'.

Alarm 2: **Number of arguments passed to 'RegExp.prototype.test()' should be 1. But 2 arguments are passed.**
* Src: src/main-process/application-menu.js (204:14-204:68)
  `if (item.command) {`
  `  item.accelerator = this.acceleratorForCommand(item.command, keystrokesByCommand)`
  `  item.click = () => global.atomApplication.sendCommand(item.command, item.commandDetail)`
  `  if (!/^application:/.test(item.command, item.commandDetail) /* alarm */) {`
  `    item.metadata.windowSpecific = true`
* Cause
  - 'RegExp.prototype.test()' only has one argument, so the second argument 'item.commandDetail' is ignored.
* Guide
  - Remove the second argument 'item.commandDetail'.

50Wliu:
Pretty cool project, thanks for creating the issue and PR!

PR #16859, opened 2018-02-28
Add 'await' keyword at async function call and remove unnecessary parameter

jerry88kim:
### Description of the Change

* Add 'await' keyword at async function 'isOpenInEditor()' call
  - 'isOpenInEditor()' returns a Promise object, so 'await' keyword should be added to use the boolean value for condition check
* Remove unnecessary parameter of 'RegExp.prototype.test()'
  - Remove second parameter of 'RegExp.prototype.test()' because it only handles the first parameter. (See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test)

See [DeepScan results page](https://deepscan.io/dashboard/#view=project&pid=2005&bid=9394&subview=issues) for more details.

### Benefits

Fix bug and optimize code

### Applicable Issues
https://github.com/atom/atom/issues/16858

### DIFF
--- a/src/main-process/application-menu.js
+++ b/src/main-process/application-menu.js
@@ -201,7 +201,7 @@ class ApplicationMenu {
       if (item.command) {
         item.accelerator = this.acceleratorForCommand(item.command, keystrokesByCommand)
         item.click = () => global.atomApplication.sendCommand(item.command, item.commandDetail)
-        if (!/^application:/.test(item.command, item.commandDetail)) {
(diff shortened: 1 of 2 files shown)


50Wliu:
Regex change looks good, @smashwilson I think you're probably the best person to review the path-watcher changes.

smashwilson (approved):
Oh hey, good catch 👍

PR #16863, opened 2018-03-01
Fix dock dragging and dropping

### LINKED ISSUE #16769: Dragging to closed Dock impossible
### Prerequisites

### Description

If you attempt to drag an item to a closed Dock panel, the panel goes into a fit of opening/closing and doesn't allow you to place the item in it.

### Steps to Reproduce

1. Open Atom v1.25.0-beta0 in safe mode
2. Have a closed Dock item
2. Attempt to drag something into the closed Dock item

**Expected behavior:**

The Dock panel opens and allows dropping the item in it.

Here is how Atom v1.24.0 behaves:

_Note: You can see that the Dock panel has a little "hiccup" that looks similar to what is now happening, but resolves itself._

**Actual behavior:**

The Dock panel rapidly opens and closes, preventing the item from being dropped there until you can open it (with another item).

**Reproduces how often:**

100% of the time.

### Versions

OS: Windows 10 x64 (Enterprise and Pro tested)

```
Atom    : 1.25.0-beta0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.25.0-beta0
python 2.7.14
git 2.16.1.windows.4
visual studio 2015
```

Compared against:
```
Atom    : 1.24.0
Electron: 1.6.16
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Additional Information

matthewwithanm:
This fixes #16769, which described an issue whereby trying to drop into
a closed dock would cause the dock to repeatedly open and close.

I tried to describe in comments why this was happening and how we can
make sure to avoid it.

The solution adds another DOM measurement which I'm not too thrilled
about but I profiled and it didn't seem to be an issue.

### DIFF
--- a/src/dock.js
+++ b/src/dock.js
@@ -296,7 +296,7 @@ module.exports = class Dock {
   }
 
   handleDrag (event) {
-    if (!this.pointWithinHoverArea({x: event.pageX, y: event.pageY}, false)) {
+    if (!this.pointWithinHoverArea({x: event.pageX, y: event.pageY}, true)) {
       this.draggedOut()
     }
   }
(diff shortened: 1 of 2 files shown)


as-cii (approved):
Looks good @matthewwithanm! Measuring one more time shouldn't be a huge concern because we were already calling `getBoundingClientRect` inside `pointWithinHoverArea`, so unless I am missing something this pull request won't add extra reflows.

Thanks for jumping on this, please merge at will once tests are green again. ⚡️

daviwil (approved):
Thanks a lot Matthew!

maxbrunsfeld (approved):

PR #16868, opened 2018-03-01
GitHub package Q1 roadmap

smashwilson:
Alright, fine, I'll go first 😅 These are the GitHub package team's quarterly goals for Feb 2018 to Apr 2018.

/cc @iolsen for feedback on the format and placement in the repo. In particular, I wasn't sure if we wanted to have a single, living document at `docs/quarterly/github-package.md` that we continuously update, or if we wanted to do new directories for each quarter like I have it here.

/cc @atom/github-package

### DIFF
--- a/docs/quarterly/fy19q1/github-package.md
+++ b/docs/quarterly/fy19q1/github-package.md
@@ -1,0 +1,71 @@
+# Git and GitHub integration
+
+## Docs
+
+## Repositories
+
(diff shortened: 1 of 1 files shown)


iolsen (approved):
Thanks, @smashwilson! We definitely want these to be updated as the quarter progresses, but I like preserving past quarters, which this allows for. Looks good to me.

PR #16878, opened 2018-03-03
Optimize get decorations

Connormiha:
### Description of the Change
Refactoring `getDecorations`. The main goal is to avoid extra loops.
In previos solutions it collected all `decorations` and twice use `Array.prototype.filter`.
In new solution it checks for filters and type before adding to array.

- [ ] Add benchmark

### DIFF
--- a/src/decoration-manager.js
+++ b/src/decoration-manager.js
@@ -2,6 +2,17 @@ const {Emitter} = require('event-kit')
 const Decoration = require('./decoration')
 const LayerDecoration = require('./layer-decoration')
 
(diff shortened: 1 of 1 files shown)


lee-dohm:
Can you supply some benchmarks and extra context here? The PR template has things that we want to see or have explained with regard to changes.

maxbrunsfeld:
I'm going to close this out since we haven't heard back. I'm not against this change, but I do wonder if this code path is worth optimizing; were you seeing it show up as a significant time sink in profiles? If you are still interested in pursuing this and get some time to finish it up, we can reopen.

PR #15831, opened 2017-10-05
Handle edits that scroll up due to hiding the horizontal scrollbar

Ben3eeE:
## TODO

- [ ] Style the horizontal scrollbar in tests so it's the same height on all platforms
- [x] Check if this fixes https://github.com/atom/atom/issues/16289 

---

**Before:**
before safe mode

**After:**
after dev mode

Fixes https://github.com/atom/atom/issues/16289
Fixes https://github.com/atom/atom/issues/16722
Fixes https://github.com/atom/atom/issues/16710
Fixes https://github.com/atom/atom/issues/15263

PR #16884, opened 2018-03-05
Randomize fontSize and lineHeight in the random test

Ben3eeE:
This causes the failure fixed by https://github.com/atom/atom/pull/15831 to be caught by the randomized test more easily.

Also removes the test merged with https://github.com/atom/atom/pull/15831 since it is not failing.

### Verification Process

- [x] Revert the changes in https://github.com/atom/atom/pull/15831 and observe test failure with the seed `1520247533732`
- [x] Restore the changes in https://github.com/atom/atom/pull/15831 and observe that this seed is passing.

/cc: @nathansobo


as-cii:
⚡️✨🔥

PR #16885, opened 2018-03-05
[wip] Improve ability to detect Python 2 during build

smashwilson:
### Description of the Change

Use [@atom/dowsing-rod](https://github.com/atom/dowsing-rod) to aggressively detect python 2 installations and pass it to node-gyp subprocesses. We won't be able to get away from requiring python 2 entirely, but we can at least accept a python 2 path from `npm config get python` or `${PYTHON}`.

Without this in place, even with `${PYTHON}` set to a python2 install, the build for git-utils fails with a python 2/3 compatibility problem. It turns out that the Makefile generated by node-gyp sometimes executes a file with a `#!/usr/bin/env python` shebang line, so unless the python2 binary is actually discoverable on the `${PATH}` as "python", the build fails.

### Alternate Designs

The clearest alternative is to actually fix node-gyp at the source, by porting gyp to Python 3 or JavaScript. That would require getting a patch accepted into gyp, then waiting for it to be included in a node-gyp release, then a node release, then upgrade our build to use that node version.

It's where the build is.

I'm also going to create a parallel pull request for apm to improve our Python detection during apm builds (atom/apm#775).

### Benefits

Developers who:

* Use `homebrew` on MacOS and have the `python` formula installed.
* Use Arch Linux or other distributions that use Python 3 as their default `python` package.

...will be able to build Atom without having to break things on their system that rely on Python 3 being the default. We'll also have a nice message if Python 2 cannot be found.

### Verification Process

The ultimate test is to run `script/build` with Python 3 on the PATH.

### Applicable Issues

* atom/git-utils#48
* atom/git-utils#72
* atom/apm#698
* atom/apm#706
* atom/apm#703

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -3,6 +3,8 @@
 'use strict'
 
 const childProcess = require('child_process')
+const path = require('path')
+const dowsingRod = require('@atom/dowsing-rod')
 const CONFIG = require('./config')
 const cleanDependencies = require('./lib/clean-dependencies')
 const deleteMsbuildFromPath = require('./lib/delete-msbuild-from-path')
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Re: Alternate designs: https://github.com/nodejs/node-gyp/issues/1337

smashwilson:
There's also this: nodejs/node-gyp#960

smashwilson:
I believe (hope?) that this has been obsoleted by new npm/node-gyp releases.

PR #16886, opened 2018-03-05
Fix shaking editor when typing

### LINKED ISSUE #15786: Atom Editor is shaking when writing
### Prerequisites

### Description

[Description of the issue]

### Steps to Reproduce

1. Fill editor with text so that the cursor is at the bottom of the screen when writing
2. Press enter for a new line and start writing
3. The whole editor window will start shaking up and down
4. Does not do it consistently (sometimes it will shake some other times it wont just press enter again and keep writing until it occurs)

No shaking whatsoever as it is annoying

Shaking (Can sent a video if you want)

When in safe mode at random.
When not all the time the cursor reaches the bottom of the screen. Even if there is text beneath the cursor in the document it still happens

### Versions
atom-beta 1.21

Ben3eeE:
### Requirements

### Description of the Change

When using DPI scaling the scroll position could be set to a position that was not possible causing a scroll event to scroll back. This was causing the editor to shake when editing. This PR fixes this by rounding scrollTop and scrollLeft to a physical pixel instead of rounding using Math.round

### Alternate Designs

N/A

It is already in Core

### Benefits

* Fixes a bug causing the editor to shake when typing

### Verification Process

**Before:**
before

**After:**
after

### Applicable Issues

Fixes https://github.com/atom/atom/issues/15786

/cc: @nathansobo

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2811,7 +2811,7 @@ class TextEditorComponent {
   setScrollTop (scrollTop) {
     if (Number.isNaN(scrollTop) || scrollTop == null) return false
 
-    scrollTop = Math.round(Math.max(0, Math.min(this.getMaxScrollTop(), scrollTop)))
+    scrollTop = roundToPhysicalPixelBoundary(Math.max(0, Math.min(this.getMaxScrollTop(), scrollTop)))
     if (scrollTop !== this.scrollTop) {
       this.derivedDimensionsCache = {}
       this.scrollTopPending = true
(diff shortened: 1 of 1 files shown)


daviwil (approved):

daviwil:
Definitely worth including this in a potential Atom Beta hotfix this week.  Is it also worth shipping a stable hotfix if we have some other fixes to include?  /cc @iolsen

as-cii (approved):
Looks good, nice work reproducing this @Ben3eeE! 

To recap, we were setting the scroll position to a position that was not physically addressable due to `window.devicePixelRatio` being a number between 1 and 2. As such, the browser clipped the requested scroll position, which in turn triggered a new scroll top change to a position that was actually addressable. This new event was intercepted by `TextEditorComponent` which attempted to scroll again, but rounding differently this time around due to the clipping the browser was applying.

By always rounding the scroll top position to a physically addressable pixel we solve this problem and prevent the browser from doing the clipping itself, thereby avoiding the flickering as well.

Ben3eeE:
@daviwil I agree that this would be good to include in an upcoming beta hotfix.

PR #16926, opened 2018-03-10
Remove extra hyphen in option description

lee-dohm:
### Description of the Change

Extra hyphen is in documentation but not on the actual parameter.

### Alternate Designs

N/A

Because it already is.

### Benefits

People will stop being confused by incorrect documentation.

### Possible Drawbacks

None known.

### Verification Process

Checked to make sure the `--help` documentation was correct.


as-cii:
Thanks!

Issue #16933, opened 2018-03-11
Dock Bug: Crossing dock toggle with drag and drop breaks dock positioning


Mootly:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: Attempted. None of the directions to enter safe mode or attempts to work around them worked. Might because this computer is running beta and there is a space in the file name (but everything from atom to 'Atom\ Beta' errors when executing from within the Contents/MacOS/ folder).
    * Followed all applicable steps in the debugging guide: 
    * Checked the FAQs on the message board for common solutions: N/A
    * Checked that your issue isn't already filed: None found.
    * Checked that there is not already an Atom package that provides the described functionality: N/A

### Description
Dragging a file from the finder so the mouse path crosses the dock toggle of a closed left dock (tree-view pane) triggers the dock toggle, but does not redraw the rest of the window to accommodate the pop-out.

### Steps to Reproduce
(Note: I have vertical tabs and tree-view enabled.)
1. With Atom open and the tree-view pane closed, open file manager/finder to the left of the window.
2. Drag a file from the file manager/finder to the Atom window so it crosses the toggle control on the closed dock.

**Expected behavior:** [What you expect to happen]
File appears in new tab with no other events triggered.

**Actual behavior:** [What actually happens]
The tree-view pane opens, but the rest of the window does not redraw to accommodate it. Toggling the dock after this moves the rest of the window back and forth as if the dock were toggling but not the dock, which remain open over overlaps the other content in the window.

Only resolved by restarting Atom.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%
Only occurs when dragging and dropping. Just dragging over the window with the left button pressed and nothing selected has no effect.

### Versions
Various. Tested both on Mac (beta) and multiple Windows boxes (current non-beta and older standalone).

### Additional Information
Disabled vertical tabs and toolbar, both to no effect.
Reproducible on bottom dock as well. Nothing docked to right.
Maybe the toggle is looking for a mouse event without checking whether the event occurred over the toggle or entered the toggle after occurring elsewhere? The dev console doesn't show any listeners for the toggle or its immediate parent divs.
<img width="630" alt="dock-open" src="https://user-images.githubusercontent.com/1514228/37257058-3346aeac-253a-11e8-8c39-4a33766c80d8.png">
[...]

PR #16864, opened 2018-03-01
Convert Dock class to Etch

matthewwithanm:
Since #16116 is going to require big changes to how docks are layed out (for which I'll be using etch), I wanted to convert Dock itself. Then I'll update `Dock::render()` to replace the PaneContainer with a `DockGroupContainer` (which itself will render `DockGroups`).

Dock was already written in a super Reacty way so there isn't much being changed here. Basically, `render()` just returns vdom instead of being a side-effect cesspool.

While working on this, I (re)discovered #16769. In order to be able to verify that I didn't make a mess of things with this PR, I fixed that issue in #16863 and rebased on that. I opened this as a PR against that branch for review purposes, but after that's merged I'll rebase it onto master and switch the base. Please let me know if there was a better way to do it!

PR #16936, opened 2018-03-12
Fix bug caused by dragging file from desktop through dock handle

matthewwithanm:
This is a fix for #16933. To avoid rebase issues I based it on #16864.

The problem is that, while the workspace element only cares about drags originating from the current document, the same isn't true of the "dragenter" event that that dock uses. Rather than trying to handle this class of events specially in the dock, we simply limit our listening to when something is being dragged. Since we have a single source of truth for this information, this should be resilient to future changes in logic.

### DIFF
--- a/src/dock.js
+++ b/src/dock.js
@@ -233,7 +233,7 @@ module.exports = class Dock {
         ),
         $(DockToggleButton, {
           ref: 'toggleButton',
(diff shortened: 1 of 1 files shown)


daviwil (approved):
Awesome, thanks a lot for fixing this!

PR #16942, opened 2018-03-13
Fix snapshot creation after spell-check upgrade

as-cii:
### Description of the Change

This pull request pulls in the latest changes from electron-link which allow to exclude every module required from a particular file. With this new capability, we can now workaround the issue introduced with the upgrade of the spell-check package, which started requiring the `xregexp` module.

Specifically, we will now exclude all the modules required from `node_modules/xregexp/xregexp-all.js`; this file is already "browserified", meaning that it should be fine to embed it verbatim inside the snapshot. The errors observed in e.g. https://circleci.com/gh/atom/atom/7139 were in fact caused by trying to process `xregexp`'s already "browserified" modules again using electron-link.

### Alternate Designs

Unclear.

### Benefits

The build will start passing again and we can use the latest version of spell-check as well.

### Possible Drawbacks

Unclear.

### Verification Process

* CI is green again.
* Verify that spell-check works correctly by typing non-English words in an untitled buffer.

/cc: @lee-dohm @daviwil


daviwil:
:+1: Thanks @as-cii!

PR #16950, opened 2018-03-14
Uninstall blacklisted packages automatically

as-cii:
In the scenario where a bad actor has published a malicious package (e.g., harvesting computer resources for use in cryptocurrency mining without user consent, transmitting sensitive data without user consent, etc.) and users have installed it, we would like a way to helpfully remove that package from affected user systems. 

As such, with this pull request we are introducing a new mechanism that fetches a list of blacklisted packages from atom.io before any 3rd-party package gets loaded. If any of the blacklisted packages happen to be installed on the user system, Atom will attempt to uninstall automatically:

successful

If that fails for some reason (e.g., path permission issues), the package will still **not** be loaded but users will be informed that they need to uninstall the malicious package(s) manually:

permission denied

A few notes about how this was implemented:

* Checking whether a package is malicious or not happens **before** loading any third party code. This is to prevent an already `require`d package from monkey-patching and bypassing our safety measures. Please, note that there is no way (at least with plain Node APIs) of stopping a package's code from running once it has been loaded into the V8 isolate.
* The atom.io API URL is hardcoded and is not read from `process.env`. This adds some friction to local development, but it prevents malicious packages from permanently modifying a user's environment (which would always bypass the security feature herein proposed).
* `PackageManager::loadPackages` has now become asynchronous due to `window.fetch`ing the list of malicious packages from atom.io. What this means is that startup time _may_ become slightly worse, but we think we can mitigate this by using `Cache-Control` headers in the server response.
* The atom.io API returns a list of all blacklisted packages. While this list is potentially unbounded, we didn't want to send the installed user packages over the wire to respect user privacy concerns.
* There is a chance that a user had `apm link`ed a locally developed package whose name gets blacklisted at some point. Uninstalling the package is the wrong thing to do in that case, but we still think it is okay because it is a non-destructive action. Atom will not try to delete the actual folder but will just remove the symlink instead.

### TODO

* [x] Deploy blacklist endpoint to staging
* [x] Test that this code works correctly using staging
* [ ] Deploy blacklist endpoint to atom.io

@lee-dohm @jasonrudolph @haacked: I would :heart: your input on the messaging I have used in the notifications above. Please, feel free to edit this branch to make any modification you think might make that UX more user-friendly.

/cc: @atom/core

### DIFF
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -931,6 +931,10 @@ module.exports = class PackageManager {
           It is strongly recommended that you delete the above folders manually.
         `
       })
+    } else {
+      this.notificationManager.addSuccess('Successfully uninstalled malicious packages', {
+        dismissable: true
+      })
     }
   }
 }
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -18,7 +18,7 @@ const dedent = require('dedent')
 // blacklisted yet) from permanently modifying the env when they are installed.
 // This adds some friction to the development process but guarantees that Atom
 // will always hit the right endpoint.
-const BLACKLIST_API_URL = 'https://atom.io/api/blacklist'
+const BLACKLIST_API_URL = 'https://atom.io/api/packages/blacklisted'
 
 // Extended: Package manager for coordinating the lifecycle of Atom packages.
 //
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -11,6 +11,14 @@ const Package = require('./package')
 const ThemePackage = require('./theme-package')
 const {isDeprecatedPackage, getDeprecatedPackageMetadata} = require('./deprecated-packages')
 const packageJSON = require('../package.json')
+const dedent = require('dedent')
+
+// Note that we are hardcoding the URL into the source code as opposed to using
+// an environment variable here to prevent malicious packages (that haven't been
+// blacklisted yet) from permanently modifying the env when they are installed.
+// This adds some friction to the development process but guarantees that Atom
+// will always hit the right endpoint.
+const BLACKLIST_API_URL = 'https://atom.io/api/blacklist'
 
 // Extended: Package manager for coordinating the lifecycle of Atom packages.
 //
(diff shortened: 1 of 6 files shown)


src/package-manager.js
```diff
@@ -869,4 +887,54 @@ module.exports = class PackageManager {
...
     }
   }
+
+  async fetchBlacklist () {
+    try {
+      const response = await window.fetch(BLACKLIST_API_URL)
+      const blacklist = await response.json()
+      return new Set(blacklist)
+    } catch (e) {
+      return new Set()
```
steelbrain: This should log the error somewhere so we at least know when it's failing. Maybe `console.warn`?

alexandernst:
IMHO this is not a good idea, no matter how good the intention is. The user is solely responsible for whatever packages get installed into the editor. The editor should not act like an antivirus-like software.

damieng:
@alexandernst totally disagree. If a bitcoin miner comes in via a package then people will blame Atom for taking 100% cpu.

alexandernst:
Ok, here is how I see this.

This patch will actively protect users only from the very basic forms of "a bad plugin", that is, the type of malicious plugins that do `<whatever>` in the code of the plugin itself. Think about that as if you were a person that is willing to use Atom's plugins ecosystem to spread, let's say, a bitcoin miner.

Would you mine bitcoins in the plugin itself? Or would you just use the plugin to install some sort of software that would

1) hide itself better than a plugin (rootkit-like)
2) have better performance than a plugin (native code vs js)
3) can't be stopped by disabled plugins in Atom

You'd use the second option, right?

Which leads us back to my main point: this patch will protect only agains very very very basic forms of malicious code. And as a con, it will hurt the already not-so-great startup performance of Atom.

So, unless you're willing to actively review all the submitted plugins at atom.io and publish only the ones that really pass a certain level of "code quality" (a-la Apple with their App Store), this patch won't really make any difference.

Ingramz:
I have to agree with @alexandernst here. It will be an endless cat-and-mouse game: every time a new kind of _protection_ is added, malicious actors will find a way around it. So for example if you start blacklisting packages by name, the first workaround would be to locally rename the package to something random before the malicious package's original name gets blacklisted. It does cut the package from being updated in the future, but most likely things will keep working for a long time before the user notices. But if you already know how to do that, you might as well just do any of the things mentioned in the previous reply.

Also some privacy-conscious people might not like their text editor hitting an api endpoint every time they launch the editor (leaking date/time, approximate location via ip address when editor was launched).

My conclusion is essentially the same, either start reviewing the packages before they are submitted to atom.io/packages (control all of the code entering packages repository) or educate users before installing their first package that they should do this only at their own discretion (accept that packages are too powerful for you to control and advertise it as a feature).

cengizhan:
i like the idea. but disabling the package may be a better idea. and when the user tries to enable it back, he should be informed that the package had been disabled for the reason written in the blacklist (malicious, incompatible, depreceted, etc)

PR #16957, opened 2018-03-16
Fix documentation for the Config class

### LINKED ISSUE #16955: Config missing from Atom 1.25 docs
The entire Config class is missing from the [Atom 1.25 documentation](https://atom.io/docs/api/v1.25.0/Config), whereas it was present in the [Atom 1.24 documenation](https://atom.io/docs/api/v1.24.0/Config).

A quick search of the `atom-api.json` file for the beta indicates that it's missing there too.

Arcanemagus:
### Requirements

### Description of the Change

The parser that generates the documentation requires the comment for the documentation to be immediately preceding the class definition. As part of the decaffeination process in #16593 a variable declaration was placed between the class definition and it's documentation, causing the parser to think there was none... and excluding it from the final result.

### Alternate Designs

Possibly make the parser to be more flexible about nodes it ignores when finding the documentation, however this will greatly increase the complexity for little benefit.

This fixes a bug in Core.

### Benefits

The documentation for the `Config` class will now show up.

### Possible Drawbacks

None.

### Verification Process

I manually ran the documentation generation part of the build process from a "primed" state and verified that the information necessary for the documentation of the `Config` class was present.

### Applicable Issues

Fixes #16955.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -8,6 +8,8 @@ const Color = require('./color')
 const ScopedPropertyStore = require('scoped-property-store')
 const ScopeDescriptor = require('./scope-descriptor')
 
+const schemaEnforcers = {}
+
 // Essential: Used to access all of Atom's configuration details.
 //
 // An instance of this class is always available as the `atom.config` global.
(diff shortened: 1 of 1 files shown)


daviwil (approved):
Thanks a lot for fixing this @Arcanemagus!

PR #16999, opened 2018-03-20
Prevent default commands from modifying readonly TextEditors

### LINKED ISSUE #16736: Many ways to write to readonly editors
### Prerequisites

### Description

The new `setReadOnly` command (introduced in `v1.24.0`) only prevents typing "normal" characters.
Commands and other packages can still change the editor, including:

- `core:backspace`
- `core:delete`
- `editor:newline`
- `editor:indent`
- `core:paste`
- `core:cut`

Same thing for packages and any other editor command, really.

**Expected behavior:** A read-only editor does not change at all.

**Actual behavior:** A read-only editor can be edited.

### Versions

```
Debian 9

Atom    : 1.24.0
Electron: 1.6.16
Chrome  : 56.0.2924.87
Node    : 7.4.0

Atom    : 1.25.0-beta0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
```

smashwilson:
The `readonly` property prevents users from interacting with a TextEditor by typing characters, but other default commands (such as `core:backspace`, `core:paste`, `core:transpose`...) still modify the buffer.

If external packages attempt to modify a readonly TextEditor, let's allow this, but require them to pass an option to indicate that they explicitly intend to, and throw an Error otherwise:

```js
editor.setReadOnly(true)
editor.insertText('okay', {bypassReadOnly: true}) // Works
editor.insertText('oops') // Throws an Error
```

Fixes #16736.

### DIFF
--- a/spec/selection-spec.js
+++ b/spec/selection-spec.js
@@ -154,4 +154,118 @@ describe('Selection', () => {
       expect(editor.isFoldedAtBufferRow(0)).toBe(false)
     })
   })
+
+  describe('within a read-only editor', () => {
+    beforeEach(() => {
+      editor.setReadOnly(true)
+      selection.setBufferRange([[0, 0], [0, 13]])
+    })
+
+    const modifications = [
+      {
+        name: 'insertText',
+        op: opts => selection.insertText('yes', opts)
+      },
+      {
+        name: 'backspace',
+        op: opts => selection.backspace(opts)
+      },
+      {
+        name: 'deleteToPreviousWordBoundary',
+        op: opts => selection.deleteToPreviousWordBoundary(opts)
+      },
+      {
+        name: 'deleteToNextWordBoundary',
+        op: opts => selection.deleteToNextWordBoundary(opts)
+      },
+      {
+        name: 'deleteToBeginningOfWord',
+        op: opts => selection.deleteToBeginningOfWord(opts)
+      },
+      {
+        name: 'deleteToBeginningOfLine',
+        op: opts => selection.deleteToBeginningOfLine(opts)
+      },
+      {
+        name: 'delete',
+        op: opts => selection.delete(opts)
+      },
+      {
+        name: 'deleteToEndOfLine',
+        op: opts => selection.deleteToEndOfLine(opts)
+      },
+      {
+        name: 'deleteToEndOfWord',
+        op: opts => selection.deleteToEndOfWord(opts)
+      },
+      {
+        name: 'deleteToBeginningOfSubword',
+        op: opts => selection.deleteToBeginningOfSubword(opts)
+      },
+      {
+        name: 'deleteToEndOfSubword',
+        op: opts => selection.deleteToEndOfSubword(opts)
+      },
+      {
+        name: 'deleteSelectedText',
+        op: opts => selection.deleteSelectedText(opts)
+      },
+      {
+        name: 'deleteLine',
+        op: opts => selection.deleteLine(opts)
+      },
+      {
+        name: 'joinLines',
+        op: opts => selection.joinLines(opts)
+      },
+      {
+        name: 'outdentSelectedRows',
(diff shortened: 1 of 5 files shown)


smashwilson:
> Why throw errors? Why don't just ignore the input?

That's the other option, yeah. I waffled a bit but I wrote this to throw an error because it seemed like the more informative thing to do rather than fail by silently doing nothing. Also, read-only editors are a fairly recent addition.

> The packages that use read-only editors don't have control over the packages that alter editors globally.

That's a really good point.

Maybe I should change it to silently ignore by default, but throw in spec or dev mode? That may be a nice alternative to give package developers a chance to notice the failure case, but not give package users a bad experience.

t9md:
Noticed, this change impact is big for vim-mode-plus.
I noticed this issue at [most vim-mode-plus's operator got error](https://github.com/t9md/atom-vim-mode-plus/issues/1058).

Before #16294, vim-mode-plus have been using `editor.component.setInputEnabled(false)` to achieve `normal-mode` which allow `h, j, k, l` to move cursor instead of inserting `h`, `j`, `k`, `l` chars.
But after #16294, setting `editor.component.setInputEnabled(false)` is now equivalent to calling `editor.setReadOnly(true)`.
Now, this PR, we forced to set `{bypassReadOnly: true}` for **every mutation command** if we want to mutate readOnly editor.

vim-mode-plus does lots of mutation in `normal-mode` through operators such as `p`, `P`, `surround`, `upper-case`(`g U`) and `lower-case`(`g u`)...
After this PR, all operators now need to add `{bypassReadOnly: true}` options when calling text mutation function.
But doing so is still not perfect.

Since
- Before this PR, normal mutation commands or package(which is completely not related to vim-mode-plus) was still being able to change text **in normal-mode(readOnly editor)** which was useful for some workflow(user can change text without explicitly entering `insert-mode`(readOnly=false).
- But after this PR all these commands no longer work unless each package author add `bypassReadOnly: true` option.

I understand this is what you wanted to do for some feature like GitHub package's readOnly editor, but for vim-mode-plus's point of view, I want disable all this `bypassRedOnly: true` enforcement.
I just want to allow any mutations without specifying `bypassRedOnly: true`.
Since I want normal text transformation package is usable in vim-mode-plus's `normal-mode` without asking each package author to set `bypassReadOnly: true` options.

I want editor's textInput being ignored on `normal-mode` but still want to allow mutation without explicitly specifying `bypassRedOnly: true` for vim-mode-plus enabled editor.

I want something like `editor.setDefaultBypassReadOnly(true)` and want to call it for vim-mode-plus enable editor to revive previous behavior.
What do you think?

t9md:
@smashwilson , @maxbrunsfeld give me some comment, what do you think?
This is very critical breaking change for vim-mode-plus and other package was used with vim-mode-plus.

lee-dohm:
Hey @t9md, it was my fault that there's been a delay in getting back to you about this. We discussed it in a team meeting yesterday and I was supposed to summarize the thoughts for @smashwilson and got side-tracked. I've done so now and we're talking over what the best approach is and we'll let you know what we figure out :+1:

t9md:
Thanks for your help @lee-dohm and am glad to hear your team discussed about it!

t9md:
Btw one idea for this is to introduce soft-read-only mode which basically prevent insertion of text through key typing but can change text without special `bypass` instruction.
Which is enabled by `editor.setReadOnly(true, {soft: true})` something like that.

smashwilson:
> Btw one idea for this is to introduce soft-read-only mode which basically prevent insertion of text through key typing but can change text without special bypass instruction.

👍 This matches what I was thinking. I feel like the name "read-only" doesn't describe what you actually want to accomplish - you want to take responsibility for keyboard input, but you still want the editor to be modifiable.

I'll file a new issue to track this.

PR #17008, opened 2018-03-21
allow the use of double slash flow comments

saadq:
### Description of the Change
Currently, a Flow comment needs to be used with the following syntax: `/* @flow */`. This PR allows the usage of `// @flow` as well.

To allow flexibility the same way that both `'use babel'` and `"use babel"` are supported.

### Benefits
It's a convenience.

### Possible Drawbacks
It is kind of a trivial addition.

### Verification Process
I wrote a test case for this, but I was having trouble getting `npm test` or `script/test` to work for me. I got an error like so:

```
AssertionError [ERR_ASSERTION]: More than one application to run tests against was found. 
```

I'm not sure how I'm supposed to develop on Atom if I already have Atom installed on my computer which I imagine is causing this error.

### Applicable Issues

N/A

### DIFF
--- a/spec/babel-spec.coffee
+++ b/spec/babel-spec.coffee
@@ -42,6 +42,11 @@ describe "Babel transpiler support", ->
       transpiled = require('./fixtures/babel/flow-comment.js')
       expect(transpiled(3)).toBe 4
 
+  describe 'when a .js file starts with // @flow', ->
+    it "transpiles it using babel", ->
+      transpiled = require('./fixtures/babel/flow-slash-comment.js')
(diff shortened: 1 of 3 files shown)


lee-dohm:
The instructions for [how to test changes to Atom Core are in the Flight Manual.](https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/)

maxbrunsfeld:
Since tests (including the new one) pass and the change is very straightforward, I'm ok 🚢 this at this point. You cool with that @lee-dohm ?

PR #17021, opened 2018-03-23
Synchronously track URIs in the process of opening

### LINKED ISSUE #13914: workspace.open() opens two editors when called in quick succession (with searchAllPanes = true)
### Steps to Reproduce

`atom.workspace.open('/path/to/file', {searchAllPanes: true}); atom.workspace.open('/path/to/file', {searchAllPanes: true});`

**Expected behavior:** Only one editor

**Actual behavior:** Two editors

### Versions

1.15.0-beta4

### Additional Information

@ebluestein found this while working on a feature for Nuclide. It's occurring because the item's being searched for synchronously but then added asynchronously. To solve it, we probably have to introduce a queue so that the second call doesn't happen until the first completes. (This can be as simple as storing a promise per-URI while the `open()` is in progress and chaining to that.)

cc @maxbrunsfeld @ebluestein

smashwilson:
Synchronously track a set of `incoming` URIs within a Workspace. If another item is being opened at the same URI, `await` its Promise before continuing, to ensure that ::open() has a chance to find the Item added by the first call if it succeeds.

Fixes #13914.


smashwilson:
Hmm, that CircleCI failure might be legit:

```
ResultsView
  opening results
    it opens the file containing the result in a non-pending state when the search result is double-clicked
      Expected <TextEditor 1552> to be null.
        Error: Expected <TextEditor 1552> to be null.
        at /Users/distiller/atom/node_modules/find-and-replace/spec/results-view-spec.js:417:75
        at Generator.next (<anonymous>)
        at step (/Users/distiller/atom/node_modules/find-and-replace/spec/results-view-spec.js:1:428)
        at <anonymous>
```

smashwilson:
atom/find-and-replace#1006 should get this green again. I'll crunch on a few issues over there to... conserve release numbers I guess?

matthewwithanm:
thanks @smashwilson!

PR #17041, opened 2018-03-27
Add Probot lock configuration

lee-dohm:
### Description of the Change

Add lock configuration in preparation for adding [Lock Threads](https://probot.github.io/apps/lock/) to the Atom organization.

### Alternate Designs

N/A

It needs to be everywhere.

### Benefits

The triage team will have to deal less with people saying "me too" on closed issues that, while they may appear related, often aren't.

### Possible Drawbacks

People may get upset that we're preventing them from commenting on old issues. Of course, they'll have to open new issues to let us know about it, which is the whole point.

### Applicable Issues

Too many to list.

/cc @Ben3eeE @50Wliu because you mentioned that you have a canned response for these kinds of things and maybe that would be better than what I've tossed together for the lock comment here.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -14,4 +14,4 @@ lockComment: >
 exemptLabels:
   - help-wanted
 # Limit to only `issues` or `pulls`
-# only: issues
+only: issues
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,13 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 180
+# Comment to post before locking. Set to `false` to disable
+lockComment: >
+  Locking this issue since it is closed and has had enough time for people
+  to give further feedback on. If you believe that you are encountering
+  the same problem described here, please open a new issue and fill out
+  [the entire template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md)
+  to ensure that we have enough information to get this fixed. Thanks!
+# Limit to only `issues` or `pulls`
+# only: issues
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -6,7 +6,7 @@ daysUntilLock: 180
 lockComment: >
   This issue has been automatically locked since there has not been
   any recent activity after it was closed. If you can still reproduce this issue in
-  [safe mode](https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)
+  [Safe Mode](https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode)
   then please open a new issue and fill out
   [the entire issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md)
   to ensure that we have enough information to address your issue. Thanks!
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -10,5 +10,8 @@ lockComment: >
   then please open a new issue and fill out
   [the entire issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md)
   to ensure that we have enough information to address your issue. Thanks!
+# Issues or pull requests with these labels will not be locked
+exemptLabels:
+  - help-wanted
 # Limit to only `issues` or `pulls`
 # only: issues
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -4,10 +4,10 @@
 daysUntilLock: 180
 # Comment to post before locking. Set to `false` to disable
 lockComment: >
-  Locking this issue since it is closed and has had enough time for people
-  to give further feedback on. If you believe that you are encountering
-  the same problem described here, please open a new issue and fill out
-  [the entire template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md)
-  to ensure that we have enough information to get this fixed. Thanks!
+  This issue is now locked since it had enough time for everyone to give
+  further feedback on after it was closed. If you can still reproduce this
+  issue in safe mode `atom --safe` then please open a new issue and fill out
+  [the entire issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md)
+  to ensure that we have enough information to address your issue. Thanks!
 # Limit to only `issues` or `pulls`
 # only: issues
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -4,10 +4,11 @@
 daysUntilLock: 180
 # Comment to post before locking. Set to `false` to disable
 lockComment: >
-  This issue is now locked since it had enough time for everyone to give further
-  feedback on after it was closed. If you can still reproduce this issue in
+  This issue has been automatically locked since there has not been
+  any recent activity after it was closed. If you can still reproduce this issue in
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Our canned response is like

> if you can reproduce the issue after starting Atom in safe mode `atom --safe` then please open a new issue and completely fill out the provided issue template.

I like mentioning safe mode in the canned response in addition to it being a prerequisite in the issue template since it double confirms that safe mode has been checked.

Something like this is how I would write it if I combined the two and changed it up some ✨ 

> This issue is now locked since it had enough time for everyone to give further feedback on after it was closed. If you can still reproduce this issue in safe mode `atom --safe` then please open a new issue and fill out [the entire issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md) to ensure that we have enough information to address your issue. Thanks!

lee-dohm:
Would you mind updating the text for me? That looks ✨

Ben3eeE:
@lee-dohm Sure thing. I updated it in https://github.com/atom/atom/pull/17041/commits/5ad7a0a0b31b0aed77b17fb6aa97cc9870e815f3

lee-dohm:
Sorry I didn't think of it. Can we link the Safe Mode suggestion to the
Flight Manual?

.github/lock.yml
```diff
@@ -0,0 +1,13 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 180
+# Comment to post before locking. Set to `false` to disable
+lockComment: >
+  This issue is now locked since it had enough time for everyone to give further
```
Arcanemagus: > This issue is now locked since it had enough time for everyone to give further feedback on after it was closed.

I think at the least this should have the "after it was closed" part removed, since the timeout is based on inactivity, not the close date.

> This issue is now locked as it has had enough time for everyone to give further feedback on it.

Or alternatively, with more finality to it:

> This issue is now locked as the time for any further feedback has passed.

rsese: I like the the conciseness of the last form 👍

rsese:
I'm assuming the bot will lock all closed issues currently?  Not sure if it can be configured to not lock some types of issues?   `help-wanted` issues for example, I'm guessing we wouldn't want to lock those?

Arcanemagus:
> Not sure if it can be configured to not lock some types of issues?

It [looks like ](https://github.com/dessant/lock-threads/blob/38084ff77462d9d09711313a01deb74a261076d6/src/lock.js#L41-L62) currently this isn't possible.

.github/lock.yml
Ben3eeE: The bot only locks closed issues. I wanted to have the keyword `closed` in there to communicate that only closed issues are locked. But I agree that the current wording might be misleading. I don't get any immediate ideas on how to change this to include both things :thought_balloon:

Clearly communicating how and why the bot acts is something I learned is important when rewording the stale message after feedback. If we remove closed there is someone that will believe we started locking all issues and PRs.

Ben3eeE: Maybe

> This issue has been automatically locked since there has not been any recent activity after it was closed.

Ben3eeE:
@rsese Great question about `help-wanted` issues. There are also issues that are closed without being resolved. For example due to inactivity by stale.

Options for these issues that I can see are:

1. Contribute to the bot to add an exempt labels configuration option
1. Lock them and reword the message to include actions that we want users to take if they are tackling a closed+locked issue that is not resolved.

So something like this for the 2nd option (Quick draft):

> Thanks for your contribution!
>
> This issue has been automatically locked since there has not been any recent activity after it was closed. 
> * If you are still experiencing this issue when running [atom in safe mode](https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode) then please open a new issue and fill out [the entire issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md) to ensure that we have enough information to address your issue. 
> * If this issue was closed without being resolved and you want to contribute with a PR to fix this issue then please [reach out to us on Slack](https://discuss.atom.io/t/join-us-on-slack/16638?source_topic_id=25406) to discuss the issue and changes required to Atom.
>

.github/lock.yml
Arcanemagus: > The bot only locks closed issues. I wanted to have the keyword closed in there to communicate that only closed issues are locked.

This message would only be shown on closed issues, but making it explicit doesn't hurt, especially since this would likely be the last comment on these threads.

Your new message works for me!

Arcanemagus:
Atom should be capitalized in the safe mode link 😉.

Would it make sense to just submit a PR there adding the ability to filter out specific labels?

> This doesn't really make sense as a message for something like merged PRs?

This is currently only applying [to issues](https://github.com/atom/atom/pull/17041/files#diff-7761990043d6a34d155f65206eb35214R13).

lee-dohm:
> Would it make sense to just submit a PR there adding the ability to filter out specific labels?

Yes, I think it would :+1:

Ben3eeE:
With https://github.com/dessant/lock-threads/commit/8a29d89aa81da815a0a471f9d85f24400bd03305 being released in version 0.2.0 of the lock threads bot we can now use `exemptLabels`. I added this in https://github.com/atom/atom/pull/17041/commits/090ff41405de46a0e62400c289ba8565abee217f

@lee-dohm This should be ready for review now. Do you need to update the bot version we use as well?

dessant:
> This is currently only applying to issues.

`only: issues` needs to be uncommented to limit locking to issues.

> This doesn't really make sense as a message for something like merged PRs?

Yeah, it would be better to have a different comment for PRs, I'll copy the way Stale handles configuration in the near future and update you here, for now please limit the bot to issues only.

dessant:
@Ben3eeE, regarding the bot instance, the latest version has been deployed to [Glitch](https://probot-lock-threads.glitch.me), only the config needs to be added after the [app](https://probot.github.io/apps/lock/) has been installed.

lee-dohm:
> Do you need to update the bot version we use as well?

I haven't installed the bot yet. I didn't want to install it until the configuration was in place.

Ben3eeE:
@dessant thank you so much for helping out!!

dessant:
@Ben3eeE, no problem, thank *you* for Atom. :heart_eyes:

Arcanemagus (approved):

dessant:
Just noticed the number of inactive issues you have: https://github.com/atom/atom/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+updated%3A%3C2017-09-30

This will mean 30 emails every hour in the next two weeks for repo watchers, 10k in total. Then there are the issue subscribers. :fearful:

Are you guys prepared for The Great Spammy Atom Easter? :rabbit: 

Alternatively you can opt to lock issues which have been inactive for more than a year without posting a comment, then switch back to the current config two weeks from now. That way repo watchers would get 1.3k emails instead of 10k.

I agree that locking without mentioning any reason could be confusing for those who stumble upon these threads, even if you decide to do so for very old issues only. I'm just not sure how will people react to their inboxes.

dessant:
Setting custom config options for pull requests is now possible: https://github.com/dessant/lock-threads#configuration.

Though pull requests stop being processed after a while regardless of the config values due to a search API [issue](https://github.com/dessant/lock-threads/issues/5).

PR #17050, opened 2018-03-29
Remove duplicate command bindings for undo and redo

maxbrunsfeld:
This fixes a regression introduced in #16999. The `core:undo` and `core:redo` commands were registered twice: once for non-read-only editors and once for *all* editors. This caused the `undo` and `redo` methods to get called twice when typing `command-z` and `command-shift-z` (or when dispatching `core:undo`/`core:redo` from the command palette).


smashwilson (approved):
D'oh! Nice catch.

PR #17057, opened 2018-03-30
Add --generate-api-docs to build script

Arcanemagus:
### Requirements

### Description of the Change

Add a new option `--generate-api-docs` that changes the build script to only do what is required to build the API documentation, skipping the process of building the full Atom binaries.

### Alternate Designs

Continuing to hack the build scripts manually to follow this behavior whenever a documentation only change is needed.

It is an enhancement for the build script already in core.

### Benefits

Adds the ability to _only_ build the documentation when testing changes that only affect that and don't need the full binary to be built.

### Possible Drawbacks

Slightly higher complexity to the build script.

### Verification Process

Tested that this only does the necessary steps to build the documentation.

### Applicable Issues

None.

### DIFF
--- a/script/build
+++ b/script/build
@@ -6,7 +6,11 @@
 // are installed.
 require('./bootstrap')
 
+// Needed so we can require src/module-cache.coffee during generateModuleCache
+require('coffee-script/register')
 require('colors')
+
+const path = require('path')
 const yargs = require('yargs')
 const argv = yargs
   .usage('Usage: $0 [options]')
@@ -22,10 +26,6 @@ const argv = yargs
   .string('install')
   .wrap(yargs.terminalWidth())
   .argv
-
-// Needed so we can require src/module-cache.coffee during generateModuleCache
-require('coffee-script/register')
-const path = require('path')
 
 const checkChromedriverVersion = require('./lib/check-chromedriver-version')
 const cleanOutputDirectory = require('./lib/clean-output-directory')
--- a/script/build
+++ b/script/build
@@ -78,69 +78,67 @@ if (!argv.existingBinaries) {
   }
 }
 
-if (argv.generateApiDocs) {
-  process.exit()
-}
-
-binariesPromise
-  .then(packageApplication)
-  .then(packagedAppPath => generateStartupSnapshot(packagedAppPath).then(() => packagedAppPath))
-  .then(packagedAppPath => {
-    switch (process.platform) {
-      case 'darwin': {
-        if (argv.codeSign) {
-          codeSignOnMac(packagedAppPath)
-        } else {
-          console.log('Skipping code-signing. Specify the --code-sign option to perform code-signing'.gray)
-        }
-        break
-      }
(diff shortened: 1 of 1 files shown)


script/build
```diff
@@ -21,11 +17,16 @@ const argv = yargs
   .describe('create-debian-package', 'Create .deb package (Linux only)')
   .describe('create-rpm-package', 'Create .rpm package (Linux only)')
   .describe('compress-artifacts', 'Compress Atom binaries (and symbols on macOS)')
+  .describe('generate-api-docs', 'Only build the API documentation')
   .describe('install', 'Install Atom')
   .string('install')
   .wrap(yargs.terminalWidth())
   .argv
 
+// Needed so we can require src/module-cache.coffee during generateModuleCache
```
daviwil: Moving this just because it isn't used in the previous section or was there another benefit?

script/build
```diff
@@ -72,7 +73,13 @@ if (!argv.existingBinaries) {
...
   generateMetadata()
   generateAPIDocs()
-  binariesPromise = dumpSymbols()
+  if (!argv.generateApiDocs) {
+    binariesPromise = dumpSymbols()
+  }
+}
+
+if (argv.generateApiDocs) {
+  process.exit()
```
daviwil: I think you'll want to wait for the promise to complete before exiting, just in case.  Maybe it's better to add another conditional around the block of code below to check for the presence of that flag before adding the `.then()`s to `binariesPromise`?  That way the process will exit on its own after the promise finishes if we're only generating API docs.

daviwil (commented):
Thanks for adding the new argument, this will be handy! Just a couple comments here, nothing major

script/build
Arcanemagus: I think this is a leftover from a separate change where I was trying to get rid of _all_ the dependency `require`s (to exit early when `--help` is called), however as the arguments are parsed _in_ a dependency, it doesn't actually make that much of a difference. I can remove this change if you would prefer.

script/build
Arcanemagus: As at this stage `binariesPromise` is just a reference to `Promise.resolve()` that hasn't been called yet, there are no "in-flight" promises to wait for a resolution of as all of the above code is synchronous.

That being said, wrapping the entire block below in a check of that flag _is_ the more correct way of doing this if you want me to do that.

script/build
daviwil: Removing it would be ideal, thanks a bunch!

script/build
daviwil: Yep, using the if statement might make it a bit more clear that the promise isn't just being ignored.  `binariesPromise` gets assigned from `dumpSymbols` if `--existing-binaries` isn't passed, so there will actually be a pending promise at that point.

Arcanemagus:
@daviwil Updated, thanks for looking over this!

daviwil (approved):
Sweet, thanks a lot for making the changes!

PR #17107, opened 2018-04-10
Convert arguments to absolute paths

okonomiyaki3000:
This script launches cmd and has it launch atom with the passed arguments. The problem is that cmd will not (necessarily) start in the pwd. So, relative paths like `.` and `..` have a different meaning. So we should first convert them to absolute paths with cygpath. 

Try it yourself. Go to some directory in bash and run `atom .` Rather than opening your current bash pwd, you'll open $HOME because when cmd opens, that's where it will be. Now apply this change and try the same thing. 

By the way, I only see this problem affecting MSysGit and MSys2. I can't say whether it affects other environments.

### Requirements

### Description of the Change

In environments where `cygpath` is available, use it to convert any paths passed to `atom.sh` to absolute paths. 

### Alternate Designs

None

Because it makes atom work properly on the command line.

### Benefits

MSys users will be able to run `atom .` to open the current directory in atom.

### Possible Drawbacks

Probably none.

### Verification Process

Tested as:
`atom .`
`atom ..` 
`atom . ..`

As well as other combinations of relative and absolute paths. On both MSys2 and MSysGit

### Applicable Issues

None

### DIFF
--- a/resources/win/atom.sh
+++ b/resources/win/atom.sh
@@ -3,7 +3,9 @@
 if command -v "cygpath" > /dev/null; then
   # We have cygpath to do the conversion
   ATOMCMD=$(cygpath "$(dirname "$0")/atom.cmd" -a -w)
+  ARGS=( $(cygpath -a -w "$@" | tr '\n' ' ') ) 
 else
+  ARGS=$@
   pushd "$(dirname "$0")" > /dev/null
   if [[ $(uname -r) == *-Microsoft ]]; then
     # We are in Windows Subsystem for Linux, map /mnt/drive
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
This looks like a good change, but I'm not familiar with Cygwin myself. Could anyone (atom maintainer or otherwise) take this for a spin, just to confirm that it works as expected?

@Ben3eeE, @50Wliu, maybe? (Not sure if either of you use cygwin; if not, sorry for the ping).

sadick254 (approved):
This works as expected. I have been able to test it out on windows with Cygwin. 👍

PR #17123, opened 2018-04-12
allow independently ignore text-input

### LINKED ISSUE #17121: Disable native TextEditor input in non-readonly editors
To complement readonly mode, provide a way to tell a TextEditor that it should not accept keyboard input, but should still be modifiable. This would be useful for any package that wants to take control of the way the editor input is processed (i.e. vim-mode-plus).

@lee-dohm and I were brainstorming names for this mode the other day; `setKeyboardInput(false)`? `setNativeInput(false)`?

/cc @t9md

t9md:
Since current strict read-only-ness does not fit to vim-mode-plus

Explain what `vim-mode-plus` want as code.
Demonstration how to fix #17121

This is simple, but this is exactly what vim-mode-plus want.

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -111,6 +111,7 @@ class TextEditorComponent {
     }
     this.derivedDimensionsCache = {}
     this.visible = false
+    this.ignoreTextInput = false
     this.cursorsBlinking = false
     this.cursorsBlinkedOff = false
     this.nextUpdateOnlyBlinksCursors = null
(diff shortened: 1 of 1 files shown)


t9md:
I confirmed broken vim-mode-plus work again with this PR and https://github.com/t9md/atom-vim-mode-plus/pull/1060.

smashwilson:
I was literally just writing up #17124 to do this 😄 Looks pretty similar, and you shouldn't even need to probe for the new method - I'm making `TextEditorComponent.setInputEnabled()` only disable keyboard input rather than making the editor readonly.

t9md:
OK, will close this then and will update t9md/atom-vim-mode-plus#1060 to use your PR then. 👍

t9md:
> update t9md/atom-vim-mode-plus#1060 to use your PR then. 👍

Oh, I don't need my PR above updated, since original `setInputEnabled` is now reverted to the behavior I wanted 🎉 💘

PR #17124, opened 2018-04-12
Separate keyboard enablement from read-only editor state

### LINKED ISSUE #17121: Disable native TextEditor input in non-readonly editors

smashwilson:
Add an `enableKeyboardInput()` configuration to `TextEditor` that allows a package to explicitly take control of the editor's input without making the editor read-only. I'm also changing `TextEditorComponent.setInputEnabled()` to modify the editor's keyboard enablement rather than read-onliness to restore the behavior of calling it directly.

Fixes #17121.

### DIFF
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2965,7 +2965,7 @@ class TextEditorComponent {
   }
 
   setInputEnabled (inputEnabled) {
-    this.props.model.update({enableKeyboardInput: inputEnabled})
+    this.props.model.update({keyboardInputEnabled: inputEnabled})
   }
 
   isInputEnabled () {
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -2965,11 +2965,11 @@ class TextEditorComponent {
   }
 
   setInputEnabled (inputEnabled) {
-    this.props.model.update({readOnly: !inputEnabled})
+    this.props.model.update({enableKeyboardInput: inputEnabled})
   }
 
-  isInputEnabled (inputEnabled) {
-    return !this.props.model.isReadOnly()
+  isInputEnabled () {
+    return !this.props.model.isReadOnly() && this.props.model.isKeyboardInputEnabled()
   }
 
   getHiddenInput () {
--- a/src/text-editor-component.js
+++ b/src/text-editor-component.js
@@ -477,7 +477,7 @@ class TextEditorComponent {
       attributes.mini = ''
     }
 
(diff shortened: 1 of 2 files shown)


daviwil (approved):
Looks great!

t9md:
Doesn't work, I'm now investigating this.
At least this part should be removed.
https://github.com/atom/atom/blob/master/src/text-editor-component.js#L480-L482

But after that, I still cant' make vim-mode-plus work in master.
Why don't you simply manage state at component level like I did in closed #1060 which seem to much simpler and explicit for me.

Will update after I found the cause.

t9md:
I need to go to bed now.
Still couldn't make it work.
And also noticed at least user need to invoke atom by `--clear-wind-sate` since old `setInputEnabled` have make editor readonly and that prop is **restored** from editorElement's state(which is not we want this time).

Again I want make ignoreTextInput explicitly separate, mean no shared-logic with read-only-ness.
It make us confuse when following code and changing code.
I bet `IgnoreTextInput` or `setInputEnabled` is used only by `vim-mode-plus`, why this has to be so general?

I want this stupidly simple so that no confusion, no breaking in future read-only behavior modification.

smashwilson:
No, it doesn't work - I hadn't actually had a chance to try it out myself yet. I'm working on it presently.

smashwilson:
Okay, I have vim-mode-plus' tests on master passing now, and editors behaving correctly as I toggle `.enableInputEnablement()` and `.setReadOnly()` in the dev console. I think all I need to do now is address the TextEditor serialization issue.

daviwil:
:+1:

smashwilson:
Okay, I've bumped the serialization key for `readOnly` to `readOnly2`. Any editors previously serialized with `TextEditorComponent.setInputEnabled(false)` will now be deserialized as read-write, not read-only.

There is a risk of us restoring an editor that used to be read-only as read-write now, but I feel like that's better than breaking everyone's serialized vim-mode editors, especially because `readOnly` is a relatively recent addition.

t9md:
Thanks @smashwilson , me too confirmed that vim-mode-plus worked properly with this PR.
Thanks!!

Issue #3043, opened 2014-07-22
Tab Length Detection


muchweb:
Default setting for tab length is 2 spaces.
- When I open a file that uses 3 spaces for indentation, and press `Tab`, I expect editor to insert 3 spaces (currently 2)
- When I open a file that uses 4 spaces for indentation, and press `Tab`, I expect editor to insert 4 spaces (currently 2)

There already were discussions to allow custom settings [per language](https://github.com/atom/atom/issues/1723) or [per project](http://rosettacode.org/wiki/Knapsack_problem/Bounded#JavaScript), although **this issue is do detect indentation automatically on file open**, because there can be different guidelines for same language files on same project. It is also handy when editing unrelated files.

Default 'tab width' setting is to be still used when creating new files and for `\t` display width.

---

Here is a screenshot from proprietary Sublime Text (some of these features are for [atom/whitespace](https://github.com/atom/whitespace), but I think initial detection should be handled in core.

![tab_size](https://cloud.githubusercontent.com/assets/7157049/3666025/3ab2ba06-11ec-11e4-81c3-d125259a0eef.png)

driskell:
@muchweb I've been tampering with Atom to make the way indentation is displayed separate from the way it is saved to a file.

So if you have a indentation width set to "4 characters" - and you load a file with tab width of 2, the loaded file, you see this:

![Image](https://github.com/driskell/tabulation-control/blob/master/SNAPSHOT1.png)

Planning to make it so it can automatically set the indentation width to match in the event you don't really have a width preference.

Just thought I'd gather some thoughts :-)

driskell:
@muchweb I've been tampering with Atom to make the way indentation is displayed separate from the way it is saved to a file.

So if you have a indentation width set to "4 characters" - and you load a file with tab width of 2, the loaded file, you see this:

![Image](https://github.com/driskell/tabulation-control/raw/master/SNAPSHOT1.png)

Planning to make it so it can automatically set the indentation width to match in the event you don't really have a width preference.

Just thought I'd gather some thoughts :-)

JordiOrlando:
@driskell What you've done looks absolutely fantastic. Is it anywhere near stable? If so, I'd love to try it out and give you my thoughts.

driskell:
It is stable. I'll do some rebasing and update to latest atom though. I've been testing other patches lately.

JordiOrlando:
Awesome, I look forward to testing it out.

driskell:
[...]

PR #17134, opened 2018-04-13
Tab Length Detection

dolsenj:
### Requirements

### Description of the Change
Currently, tab length for the text editor is set depending on what programming language is being used. This PR sets the text editor's tab length automatically when a file is opened to be edited. The tab length is calculated by an algorithm which takes an array of the lines of code in the file being edited as input and outputs the calculated tab length, accounting for variability in tab length. This change was made in workspace.js in the didChangeActivePaneItemOnPaneContainer method because tab length detection is done whenever the active pane item is changed. 

### Alternate Designs
There were four alternative designs for the tab detection algorithm. They were determining tab length by finding the Greatest Common Divisor, Minimum Width, Comparing lines, and using a Neural network. Comparing lines was chosen because it performed the best amongst the four algorithms for correctly determining tab length when testing on different files of code. For more information, refer to this [article](https://medium.com/firefox-developer-tools/detecting-code-indentation-eff3ed0fb56b) on tab length detection. 

This PR should be in the Core because it was requested to be in the core by the issue report. Tab length is also a part of the text editor itself so changing the core makes sense. 

### Benefits
Users with unique tab length preferences do not need to manually change the default tab length every time they use Atom. For example, users who use four space tabs for JavaScript, two space tabs for Python, or three space tabs for another language do not need to change the default tab length.

### Possible Drawbacks
A possible drawback is that tab length detection is performed when it is not necessary to do so.

### Verification Process
We verified that new and changed functionality works as expected by adding Jasmine specs to the workspace-spec.js file. We performed four tests on tab length detection. We ran a test on a normal JavaScript file with tab length of 2, a file with tab length of 4, a file with tab length of 3, and a file with a tab length of two but also has variability in tab length. In each case, tab length detection returned the expected tab length correctly and the changes did not introduce any regressions. We also used Travis CL to perform continuous integration testing and had a successful build.  Currently, testing on CircleCI and AppVeyor.

### Applicable Issues
https://github.com/atom/atom/issues/3043

### DIFF
--- a/docs/focus/2018-04-23.md
+++ b/docs/focus/2018-04-23.md
@@ -1,7 +1,7 @@
 ## Highlights from the past week
 
 - Atom core
-- Atom IDE
+  - Shipped Atom 1.26.0 and 1.27.0-beta0 :shipit:
 - GitHub Package
   - Regression fix: "view (un)?staged changes" on untitled buffers [atom/github#1402](https://github.com/atom/github/pull/1402)
   - Regression fix: DOMPurify and snapshotting [atom/github#1401](https://github.com/atom/github/pull/1401)
@@ -21,13 +21,9 @@
 
 ## Focus for week ahead
 
-- Atom core
-  - Shipped Atom 1.26.0 and 1.27.0-beta0 :shipit:
-- Atom IDE
 - GitHub Package
   - :notebook: Planning, roadmapping, prioritizing, scheming
 - Teletype
   - :bug: Improve handling of scenario where guest puts their computer to sleep while participating in a portal ([atom/teletype#354](https://github.com/atom/teletype/issues/354))
 - Tree-sitter
   - Continue working on https://github.com/tree-sitter/tree-sitter/pull/162, adjusting the syntax tree representation to allow the trees to be updated and queried simultaneously on multiple threads.
-- Xray
--- a/docs/focus/2018-04-30.md
+++ b/docs/focus/2018-04-30.md
@@ -1,0 +1,28 @@
+## Highlights from the past week
+
+- GitHub Package
+  - Lots of planning, still ongoing
+  - Rework the way that PaneItems are managed [atom/github#1414](https://github.com/atom/github/pull/1414)
+  - Make ListSelection models immutable in preparation for the final React port [atom/github#1421](https://github.com/atom/github/pull/1421)
+- Teletype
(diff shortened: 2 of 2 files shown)


maxbrunsfeld (changes requested):
Hey @dolsenj, this is a great feature; thanks for taking it on! Overall, your logic for inferring tab length looks good. There's a few changes I'd like to see regarding *where* the new logic should be implemented and *how* we compute the most frequent indent length.

I don't think that the code should live in `Workspace`. There is already some very closely-related functionality in `TextEditor` in a method called [`usesSoftTabs`](https://github.com/atom/atom/blob/d50349f0042f347a193229f7af5fa042631c2fbc/src/text-editor.js#L3611-L3620). This method examines the first 1000 lines of the buffer in order to decide whether to use tabs or spaces for indentation.

I think that what we should do is create a new method called `TextEditor.getCurrentSoftTabLength()`, which checks if the buffer uses spaces or tabs. If we encounter a hard tab character first, we should simply return `null`, to indicate that the editor doesn't use soft tabs. Otherwise, we should perform your logic for detecting indent length.

This new method would basically *replace* the current `usesSoftTabs` method. We would have to keep the old method for compatibility, but we could implement it in terms of the new method, like this:

```js
usesSoftTabs() {
  return this.getCurrentSoftTabLength() != null
}
```

This new method can then be used in the `TextEditorRegistry` to configure each editor's tabLength setting, instead of just using `usesSoftTabs` [like we do today](https://github.com/atom/atom/blob/d50349f0042f347a193229f7af5fa042631c2fbc/src/text-editor-registry.js#L157-L164).

It's important that we limit the number of lines that we scan, because otherwise this would severely slow down the opening of large files. Even 1000 lines may be too many, because the old `usesSoftTabs` code used to break out as soon is it found an indented line, whereas we'll need to keep going. We'll have to measure the performance impact. If it turns out to be expensive at all, one option would be to break the loop once we've seen some number of indentation changes, like 5.

maxbrunsfeld:
Also, it looks like you accidentally checked in some files that were not checked in previously, and also changed the CircleCI config file. Could you undo that stuff? Finally, you might want to rebase your branch onto the latest master; CI may be failing because of some issues that have already been fixed on master.

dolsenj:
Hey @maxbrunsfeld, I attempted to undo my changes and do a rebase of my branch onto the latest master. In trying to do this, I accidentally did a force push on my version of master which closed the pull-request.

Is it possible to re-open the pull-request or should I make a new pull-request to continue working on task?

daviwil:
@dolsenj Sorry for the delay!  Looks like GitHub won't allow me to reopen this PR because it thinks there aren't any new commits on your `master` branch.  Could you try opening a new PR?  Thanks a lot!

Issue #16900, opened 2018-03-08
Different Panes can have the same id


matthewwithanm:
Different instances of the Pane class can have the same `id` value. This is because they can come from [either a serialized object from a previous session or from an incremented integer](https://github.com/atom/atom/blob/2e19cf4a7c0a4d624b1f9c1072cd292de112cdb0/src/pane.js#L61-L66). If you deserialize a Pane with id `2`, it has no bearing on whether the autoassigned ids will create a Pane with the id `2`.

I'm not sure if this is causing any user-facing bugs at the moment; I noticed it while developing in the codebase. If it is, they're probably really weird ones that are hard to repro!

One possible fix would be to change the ids to strings and have them be a combination of a session id—e.g. a module-level (calculated once) `Date.now().toString()`—and the auto incremented integer. This would guarantee that a Pane deserialized from a previous session would never have the same id as one created this session. I'm not sure if anything requires the ids to be numbers, but it seems like a pretty low-risk change to me.

lee-dohm:
Maybe this occurs with pane items as well?

@rsese or @matthewwithanm can you see about repro steps or writing a test?

/cc @Ben3eeE in case @rsese needs help

Ben3eeE:
Hi @matthewwithanm :wave: 

I was looking into this and I am unable to reproduce the problem you describe. When I start an empty Atom window I have 4 panes. These are the center of the workspace, and the three docks. Each autoassigned an id from an integer that starts at 1. So the pane ids will be 1,2,3 and 4. If I create a new pane it will get an id of 5.

Here is what I tired:

1. Open Atom in some project folder
1. Open two new split panes and close the first pane. I now have two panes with id 6 and id 7
1. Close Atom to serialize this state
1. Reopen Atom and split one of the panes down. I now have three panes with id 6, 7 and 8

This seems like the expected behavior when reading the code that you linked:
https://github.com/atom/atom/blob/2e19cf4a7c0a4d624b1f9c1072cd292de112cdb0/src/pane.js#L61-L66

`nextInstanceId` is modified here to be one above the max id from the deserialized state.

Is there anything I am missing that could help reproduce the issue of different panes having the same id?

matthewwithanm:
Hm, well one thing you can do that demonstrates it (though in a limited way), is to add some code to the Pane constructor that stores a global list of Panes. If you load Atom, you'll see that there are actually multiple instances created with those initial ids (1-4).

sosaadat:
[...]

PR #17140, opened 2018-04-14
Pane id conflicts

Rezayovan:
### Requirements
Resolved potential pane id conflicts by making pane id a function of the previously designated pane id and time
### Description of the Change
In the pane class, to create unique pane id's, we concatenated the current timestamp to the nextInstanceId value. 

### Alternate Designs
Future designs may include finding the source of the bug mentioned in issue #16900 as to actually why identical pane id's are being created for different documents. For now, this guarantees that identical id's for different panes will not be created. 

This should be done in the core because this bug was reported in the core. 

### Benefits
The benefit is that identical pane id's will not be generated. 

### Possible Drawbacks
Any dependence on the order of pane id's for additional functionality is now broken since the pain id's are not incrementing by 1 each time. 

### Verification Process

In the Pane constructore, console.log(this.id) printed out the unique ids of the opened panes. 

Opened new files, quit atom, reopened atom, opened project folders with multiple files, restarted atom, right-click inspect element to check the logs (which had the pane id's printed)


sadick254 (commented):
Hey, @Rezayovan Thank you for your submission. It is not clear how conflicts might occur with the current implementation. We also have not seen any issues as a result of pane id conflicts. Due to that, I will have to decline to accept this PR.

Issue #16904, opened 2018-03-08
Improve soft-wrapping by breaking words like web browsers


ariasuni:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Current soft-wrapping behavior is not aggressive enough and sometimes leads to bad results.

### Steps to Reproduce

1. Enable «Soft Wrap» and «Soft Wrap at Preferred Line Length» in «Editor» tab of Settings
2. Copy an URL in a text editor tab or a sequence of non-spaces characters that have more characters than the Preferred Line Length setting

**Expected behavior:**

* Break character sequences on spaces, dashes and slashes, like in a web browser
* Break character sequences at most at the «Preferred Line Length» setting
* When the Preferred Line Length + 1 character is one space, don’t show it; if there are other spaces after, show them on the next line
* Show a grayed out character, like those for EOL or spaces, when Atom break a word that had no spaces in it. Suggestion for default character: broken bar `¦`, which is present in the same Unicode block than  `¤¬»`.

For example, if Preferred Line Length is set to 20, then it should behave like this (`|` character represent the wrap-guide):

```
https://github.com¦ |
/atom/atom/issues¦  |
/new                |
                    |
Averylongword-¦     |
thatshouldbreak     |
                    |
Averylongwordwithnos|¦  # no space, dash or slash in the word
eparator            |
                    |
separated by spacess|  # 21th character is space
only                |
```

**Actual behavior:**

* Character sequences breaks only on spaces
* If the character sequence doesn’t contain space and fill the whole line, the word is cut at the window edge instead of the expected Preferred Line Length (see screenshot).
![screenshot of current work-wrapping behavior](https://user-images.githubusercontent.com/2446451/37145899-785416b2-22c2-11e8-8229-d8eefa909d57.png)

```
https://github.com/a|tom/atom/issues/new
                    |
Averylongword-thatsh|ouldbreak
                    |
[...]

PR #17141, opened 2018-04-15
More aggressive wrapping

Rezayovan:
### Description of the Change
Wrapping now includes wrap on dash and slash. This is a response to issue #16904, in which the user wanted more "aggressive" wrapping. To implement this change, the isWrapBoundary variable was given character checks for spaces, slashes, and dashes as wrap boundaries.

### Alternate Designs
The proposed version was selected because it produced the simplest fix to this problem. 

This should be in the core because it has to do with editor functionality. 

### Benefits
A much more intuitive soft-wrap functionality that doesn't breakup words in the middle if it doesn't have to. 

### Possible Drawbacks
None

### Verification Process
The new soft-wrap was tested with tests such as these: 
screen shot 2018-04-14 at 11 34 11 pm

### DIFF
--- a/src/text-utils.js
+++ b/src/text-utils.js
@@ -103,7 +103,8 @@ const isWordStart = (previousCharacter, character) =>
   ((character !== ' ') && (character !== '\t'))
 
 const isWrapBoundary = (previousCharacter, character) =>
-  isWordStart(previousCharacter, character) || isCJKCharacter(character)
+  isWordStart(previousCharacter, character) || isCJKCharacter(character) ||
+    character === '-' || character === '/'  || character === ' '
 
 // Does the given string contain at least surrogate pair, variation sequence,
 // or combined character?
--- a/src/text-utils.js
+++ b/src/text-utils.js
@@ -104,7 +104,7 @@ const isWordStart = (previousCharacter, character) =>
 
 const isWrapBoundary = (previousCharacter, character) =>
   isWordStart(previousCharacter, character) || isCJKCharacter(character) ||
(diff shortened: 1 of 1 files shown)


Rezayovan:
Any idea on why these are failing given the addition of only one line of code?

maxbrunsfeld:
The failings tests deal explicitly with soft wrapping. I'm guessing you'll want to change the tests since you're changing the soft-wrapping behavior?

Also, in prose, the convention is that dashes should stay at the end of the line. Can you adjust this code so that the dash is not moved to the following line?

lee-dohm:
@Rezayovan Do you have any questions around this?

maxbrunsfeld:
Closing this out since we haven't heard back. Let us know if you're interested in picking this back up and we can reopen it.

ariasuni:
https://github.com/atom/atom/blob/3c9620dfa54d1cb8e1a5f356d9a5c095b95788e9/src/text-utils.js#L105-L106

should be replaced by

```javascript
const isWrapBoundary = (previousCharacter, character) =>
  isWordStart(previousCharacter, character) || isCJKCharacter(character) ||
  previousCharacter === '-' || character === '/' || character === ' '
```

I would have preferred to create a PR myself, I can’t seem to get a dev environment for Atom.

ariasuni:
I finally managed to set up a dev environment, and I’m seeing an issue (which was already here): the cursor can’t be after the last character, because it’s the same position than the beginning of the following line, and so “Move to end of line” puts the cursor before the last character of the line visually.

Issue #17060, opened 2018-03-31
Writing to config in deactivate can wipe settings


Arcanemagus:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When a package writes to the configuration during it's `deactivate` step there is a chance it can cause Atom to wipe the user's settings when Atom is restarted (such as during updating packages).

![2018-03-31_00-35-59](https://user-images.githubusercontent.com/427137/38161341-2b3940a6-3482-11e8-988d-5da8625af9da.gif)

### Steps to Reproduce

1. Create an empty folder and set the `ATOM_HOME` environment variable to this folder's path
2. Download [settings-wiper.zip](https://github.com/atom/atom/files/1865268/settings-wiper.zip), extract it, and run `apm link` from that folder in the environment from step 1
3. Launch Atom, you should get an information notification saying the current value is `0, 0, 0, 0, 0`, as well as Atom's regular first run screens
4. Set the Welcome screen to disabled in order to make it obvious when this bug hits you.
5. Launch "enough" Atom windows to trigger this bug on restarting, I found 10 windows to be a good number
6. Close one of the windows and launch a replacement
7. Note that the number increments
8. Open the developer tools and run `atom.restartApplication()` to restart Atom
9. Either the configuration will save and you will simply see an incremented number, or Atom's configuration will reset and you will see all `0`s again, and the Welcome screen will show in all windows
10. If the settings _didn't_ reset, simply run the command again until it does.

**Expected behavior:** Atom's configuration doesn't get wiped.

**Actual behavior:** Atom's configuration occasionally gets completely reset.

**Reproduces how often:** This is somewhat hard to tell, initially it was around 80%, but it can take many calls to `atom.restartApplication()` to reproduce this. Atom v1.26.0-beta0 reproduces this more often than Atom v1.25.0 does, but both versions _do_ reproduce this.

### Versions

```
[...]

PR #17166, opened 2018-04-18
Avoid writing to config file while quitting

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/17060

/cc @Arcanemagus - I was never able to reproduce the config loss using your steps. It seems like it might be more unlikely on macOs, or depending on the speed of your disk. I'm going to go ahead speculatively merge this PR that I *think* might fix the problem. When you get a chance, would you mind trying to reproduce the bug with a new build?

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -439,7 +439,11 @@ class AtomApplication extends EventEmitter {
         event.preventDefault()
         const windowUnloadPromises = this.getAllWindows().map(window => window.prepareToUnload())
         const windowUnloadedResults = await Promise.all(windowUnloadPromises)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
> In any case I did manage to reproduce it on a build based on the commit right before this merged

Nice work!

Arcanemagus:
It seems I spoke too soon, opened up that profile/build again to check something else and the settings were wiped. I'm wondering if a very rare bug I was seeing during testing might actually be related. Trying to find better reproduction steps.

SystemLogoff:
This has happened to me as well, using only the love2d ide packages.

It's a huge pain, but I've become great at setting up atom again, and again, and again.

PR #17168, opened 2018-04-18
Replace obsolete FreeBSD build instructions with current status

yzgyyang:
### Requirements

### Description of the Change

This removes the obsolete FreeBSD build instructions and add current status and tracking issues to the page.

The obsolete FreeBSD build instructions is misleading nowadays, making people continue to try to build on FreeBSD without success.

### DIFF
--- a/docs/build-instructions/freebsd.md
+++ b/docs/build-instructions/freebsd.md
@@ -1,19 +1,8 @@
 # FreeBSD
 
-FreeBSD -RELEASE 64-bit is the recommended platform.
+FreeBSD is currently not supported due to the lack of official support for FreeBSD from Electron.
 
-## Requirements
+The community has been actively attempting to bring Atom back on FreeBSD. Ongoing work and discussions can be found on the following tracking issues:
 
-* FreeBSD
-* `pkg install node`
-* `pkg install npm`
-* `pkg install libgnome-keyring`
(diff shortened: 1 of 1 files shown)


lee-dohm:
If Atom simply can't be built on any version of FreeBSD, then we should remove the document and the reference to it in the README. But Atom did work on some version of FreeBSD in the past, so perhaps it would be better to just include that the build instructions have only been validated on a specific version?

yzgyyang:
@lee-dohm I believe it is an older version of Atom rather than FreeBSD, but I have no ways to easily verify that. The doc was last updated in 2016 so I suppose it's a really old Atom version, if any. 

The reason that I think we should keep this page is that there are unofficial help and build instructions in those tracking issues and people can start from there. I personally would also like more traffic to those tracking issues, since I am actively working on bringing Electron & Atom back to FreeBSD.

shkhln:
> I believe it is an older version of Atom rather than FreeBSD, but I have no ways to easily verify that. The doc was last updated in 2016 so I suppose it's a really old Atom version, if any.

Build instructions were already completely wrong when initially submitted. Specifically, Atom install script downloads prebuilt Electron binaries, those were never officially provided for FreeBSD by GitHub and there is no mention of overriding download location in the instructions.

PR #17169, opened 2018-04-18
Remove FreeBSD build instructions since they don't work

lee-dohm:
### Description of the Change

As mentioned in #17168, the current FreeBSD build instructions don't work. This removes them to prevent people from wasting time on instructions that we know are broken.

### Alternate Designs

* https://github.com/atom/atom/pull/17168#issuecomment-382516855

It already is.

### Benefits

People will stop wasting time on broken instructions and be directed to or search for #6035, directing more attention to the effort of getting things working again.

### Possible Drawbacks

People will think that Atom doesn't work on FreeBSD (which at this point, it doesn't) and give up.

### Verification Process

N/A

### Applicable Issues

* #6035
* #17168
* electron/electron#3797


Arcanemagus (approved):

Ben3eeE:
@lee-dohm https://github.com/atom/atom/blob/master/docs/README.md also mentions FreeBSD.

lee-dohm:
Thanks @Ben3eeE! Done.

Ben3eeE (approved):

PR #17202, opened 2018-04-23
Don't mention authors in issues created by the move bot

Ben3eeE:
### Requirements

### Description of the Change

With https://github.com/dessant/move-issues/issues/1 being fixed in https://github.com/dessant/move-issues/commit/d561ee2614896b8c8f404d3696c98e8679a77380 we can now configure the move-issues bot to not mention authors.

There is also a `keepContentMentions` configuration option that will preserve mentions to teams/users in the issue content. This defaults to false.

### Alternate Designs

N/A

It should be everywhere

### Benefits

* Less mentions to the maintainers that move issues
* Less mentions to every user involved in the issue

### Possible Drawbacks

This applies to all comments and the command comment. So the user that opened the issue will have to subscribe to the new issue if they want updates.

### Verification Process

N/A

### Applicable Issues

https://github.com/dessant/move-issues/issues/1

### DIFF
--- a/.github/move.yml
+++ b/.github/move.yml
@@ -1,0 +1,4 @@
+# Configuration for move-issues - https://github.com/dessant/move-issues
+
+# Mention issue, comment and command authors
+mentionAuthors: false


lee-dohm:
Oh, I would think that we _want_ the original author subscribed to the new issue just like if they opened it themselves. I'd rather unsubscribe from issues I move than worry about original authors not getting notified of the issues they opened.

dessant:
After further thinking, the options for controlling mentions are useful, but it's not clear anymore why would we ever want to @mention the command author during the move, they are notified anyway about the successful move with a comment.

dessant:
Command author mentions have been disabled in the latest release.

Ben3eeE:
Closing this since we are going to use the default settings for the bot 😀

PR #17212, opened 2018-04-25
Remove 'project' command line flag

### LINKED ISSUE #17178: Intermittent "Unable to read supplied project specification file" error when opening files
*Edit by @rsese to add repro steps and error*

1. Install Atom 1.25.1.
2. Open some project file e.g. from atom/atom `atom --safe src/text-editor-component.js`
3. After the update downloads, `Atom > Restart and Install Update` (after clicking that I waited for a minute or but I had to click it again for Atom to restart).

Expected is an Atom window opens with `src/text-editor-component.js`, actual is 
no Atom window opens, the application menu bar only has `Atom` (like in https://github.com/atom/atom/issues/17178#issuecomment-383180280), and I see this error in `/var/log/system.log`:

```
Apr 22 20:57:03 milhouse Atom[33796]: Error: Unable to read supplied project specification file.
	    at readProjectSpecificationSync (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/parse-command-line.js:212:11)
	    at parseCommandLine (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/parse-command-line.js:147:40)
	    at start (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/start.js:37:16)
	    at Object.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/main.js:40:1)
	    at Object.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/src/main-process/main.js:42:3)
	    at Module._compile (module.js:571:32)
	    at Object.Module._extensions..js (module.js:580:10)
	    at Module.load (module.js:488:32)
	    at tryModuleLoad (module.js:447:12)
	    at Function.Module._load (module.js:439:3)
```

That error message looks specific to https://github.com/atom/atom/pull/16845?

---

### Description

Atom no longer opening on my Mac, when I try to open a file. If I just open the APP itself, it opens, and then i can load said file, but If I just try to open the file from my desktop, atom "opens" but not fully.... nothing appears on the screen but atom pops up in my dock.

### Steps to Reproduce

1. Click a document to open.
2. Nothing opens

**Expected behavior:**
Up until last night, If I click a document, it would open it. 

**Actual behavior:**
As of last night, If I click a document, atom will not open.

**Reproduces how often:** 
100%

### Versions
Atom    : 1.26.0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0

apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.26.0
python 2.7.10
git 2.16.2
### Additional Information

I have not changed or touched any of the files in Atom. I up until last night it worked perfectly fine, no issues. I thought it was a problem with my computer. I have restarted the computer and nothing.
I tried the safe mode thing, but Like described, the problem is before atom actually opens. If I open atom, and then open a document its totally fine.
Not sure if this info helps but I am using a MacBook Pro OS X 10.13.4

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/17178

This (temporarily) removes the command line interface to the feature added in https://github.com/atom/atom/pull/16845. As far as I can tell, that interface never quite worked correctly. In my testing, the `paths` specified in the given project file were not opened. I think there are actually multiple causes of this:

1. Because the project specification was not integrated into the state serialization system, once the state was loaded from indexedDB, the paths would immediately [be replaced](https://github.com/atom/atom/blob/db36ca059080e6a4651b11176b1c4d69146bb48d/src/atom-environment.js#L1227) by different paths.
2. Project paths are also updated after the call to `project.replace` when the asynchronous [`open-paths` message](https://github.com/atom/atom/blob/db36ca059080e6a4651b11176b1c4d69146bb48d/src/atom-environment.js#L1351-L1406) is received.
2. As [someone reported](https://github.com/atom/atom/pull/16845#issuecomment-379592937) on the initial PR, some logic in the command line parsing caused the directory containing the project path itself to be opened.

The actual reason that I discovered this is while investigating #17178. macOS users are getting [this error](https://github.com/atom/atom/blob/db36ca059080e6a4651b11176b1c4d69146bb48d/src/main-process/parse-command-line.js#L212) when trying to open files in Atom by double clicking them. I added some code to log out the path when that error occurs, and I'm seeing that sometimes when double clicking a file, Atom is getting invoked with a `project` flag set to some weird non-existent path like `/sn_1_01235345`. I have no idea what causes that flag to get set.

/cc @matthewwithanm @philipfweiss It seemed like the command line `project` flag was not an important feature on your end; you mainly needed the `Project.replace` API. Is that right?

I think we can put back the `project` flag at some point, but for now I just need to get a hotfix out for #17178 and don't have time to fully backfill tests for this feature, think through how to integrate the `project` flag with the rest of the system, and avoid the bug described above.

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -204,7 +204,6 @@ class AtomApplication extends EventEmitter {
 
   openWithOptions (options) {
     const {
-      projectSpecification,
       initialPaths,
       pathsToOpen,
       executedFrom,
@@ -259,7 +258,6 @@ class AtomApplication extends EventEmitter {
         profileStartup,
         clearWindowState,
         addToLastWindow,
-        projectSpecification,
         env
       })
     } else if (urlsToOpen.length > 0) {
@@ -829,7 +827,6 @@ class AtomApplication extends EventEmitter {
     window,
     clearWindowState,
     addToLastWindow,
-    projectSpecification,
     env
   } = {}) {
     if (!pathsToOpen || pathsToOpen.length === 0) return
@@ -863,7 +860,7 @@ class AtomApplication extends EventEmitter {
     }
 
     let openedWindow
-    if (existingWindow && (projectSpecification == null || projectSpecification.config == null)) {
+    if (existingWindow) {
       openedWindow = existingWindow
       openedWindow.openLocations(locationsToOpen)
       if (openedWindow.isMinimized()) {
@@ -899,7 +896,6 @@ class AtomApplication extends EventEmitter {
         windowDimensions,
         profileStartup,
         clearWindowState,
-        projectSpecification,
         env
       })
       this.addWindow(openedWindow)
--- a/src/main-process/atom-window.js
+++ b/src/main-process/atom-window.js
@@ -22,7 +22,6 @@ class AtomWindow extends EventEmitter {
     this.safeMode = settings.safeMode
     this.devMode = settings.devMode
     this.resourcePath = settings.resourcePath
-    this.projectSpecification = settings.projectSpecification
 
     let {pathToOpen, locationsToOpen} = settings
     if (!locationsToOpen && pathToOpen) locationsToOpen = [{pathToOpen}]
@@ -60,8 +59,7 @@ class AtomWindow extends EventEmitter {
       get: () => JSON.stringify(Object.assign({
         userSettings: !this.isSpec
           ? this.atomApplication.configFile.get()
-          : null,
-        projectSpecification: this.projectSpecification
+          : null
       }, this.loadSettings))
     })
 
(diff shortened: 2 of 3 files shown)


matthewwithanm:
Wow, this is a bummer. But yeah, `atom.project.replace()` is the most important thing for us.

kopischke:
Forgive me for jumping in, but I think I can shed some light on what happens here, as I was bitten by the underlying macOS quick in an unrelated project recently. As succinctly stated in [this SO answer](https://stackoverflow.com/a/10242200):

>Mac OS X assigns a unique process serial number ("PSN") to all apps launched via GUI. It's used for identifying various processes and instances of executables.

This is passed in the form of a `-psn_*` argument. Of course, Atom interprets the `-p` as the short form of the `--project` switch, with the results linked to above. Note that, unlike stated in the comments to the answer linked above, this behaviour persists up to at least macOS 10.12 _Sierra_, which is where I experienced it.

PR #17256, opened 2018-05-01
Fix out of date doc block method name

captbaritone:
Looks like this method name changed in https://github.com/atom/atom/pull/3456/files#diff-88c69a7ba7e0c0d0090f4501ca44472c but the comment did not get updated.

### DIFF
--- a/src/decoration.js
+++ b/src/decoration.js
@@ -157,7 +157,7 @@ class Decoration {
   // ## Examples
   //
   // ```coffee
-  // decoration.update({type: 'line-number', class: 'my-new-class'})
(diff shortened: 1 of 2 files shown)


50Wliu:
Was cda95c9 intentional?

captbaritone:
No. Strange. I did this via the GitHub web UI. No idea how that got picked up. I'll get rid of it. Hold on...

captbaritone:
Looks like it's not so easy to update a PR created from the Web UI. I've opened a PR to replace: https://github.com/atom/atom/pull/17263

PR #17277, opened 2018-05-05
Use the configuration from AtomApplication in the main process

smashwilson:
When initializing a PathWatcher to watch for config file changes in the main process, read the config from `AtomApplication` instead of `AtomEnvironment`.

Something small that I noticed trying to track down a root cause for https://github.com/atom/watcher/issues/124.

### DIFF
--- a/src/path-watcher.js
+++ b/src/path-watcher.js
@@ -544,10 +544,26 @@ class PathWatcherManager {
   // Private: Access the currently active manager instance, creating one if necessary.
   static active () {
     if (!this.activeManager) {
-      this.activeManager = new PathWatcherManager(atom.config.get('core.fileSystemWatcher'))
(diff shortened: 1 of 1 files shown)


src/path-watcher.js
```diff
@@ -544,12 +544,28 @@ class PathWatcherManager {
...
+
+      this.activeManager = new PathWatcherManager(config.get('core.fileSystemWatcher'))
+      this.sub = config.onDidChange('core.fileSystemWatcher', ({newValue}) => {
+        this.transitionTo(newValue)
+      })
     }
     return this.activeManager
   }
 
+  static findConfig() {
```
sstt011: There should be a space before the parentheses.
Before:
`static findConfig() {`
After:
`static findConfig () {`

fosslinux (changes requested):
These changes should fix the linter.

PR #17293, opened 2018-05-09
Add new listener for config change

jianhuafan:
### Description of the Change
Add a new listener {onDidUpdate} in atom.config, it's different from {:onDidChange} in that it will always be called, regardless whether the new value is different than the old. This new listener is used to solve the problem that when a layered project specific settings override the global settings, the {:onDidChange} won't work cause the global settings was override and the callback never occur even if the user changes the setting in menu or global config file. The new listener, however, will observe the global config change as desired.

This should be in core because recently atom merged a pull request (#16845 ) that allow per-project settings through {:resetProjectSettings}, the project settings layers the global settings by owning higher priority, the corresponding config change listener should be changed cause when user change the layered global settings the listener should be fired correctly.

### Benefits

This change add a config change listener that can behave correctly.

### Applicable Issues

This situation origins from our packages which is currently developed for per-project settings based on {:resetProjectSettings}, user can have a local config file in current project folder, our package will parse it and apply the config on the current window, our plan is when user changes the settings in global, our package should be disabled temporarily so as to apply the global config setting changes that the user made, so we used {:atom.config.onDidChange} to listen to the behavior, but it didn't work as described above.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -414,6 +414,7 @@ class Config {
       this.schema.properties.core.properties.projectHome = projectHomeSchema
       this.defaultSettings.core.projectHome = projectHomeSchema.default
     }
+    this.indicator = []
   }
 
   clear () {
@@ -435,6 +436,7 @@ class Config {
     this.pendingOperations = []
     this.legacyScopeAliases = new Map()
     this.requestSave = _.debounce(() => this.save(), 1)
+
   }
 
   /*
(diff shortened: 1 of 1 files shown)


Arcanemagus:
`onDidChange` is _almost_ never the right event to subscribe to, in virtually every case `atom.config.observe` is the right choice as it is fired when initially called _and_ when the value changes.

Does that not work with the project settings?

> it's different from {:onDidChange} in that it will always be called, regardless whether the new value is different than the old.

I'm not entirely sure what you mean by this, if the values haven't changed, why should _any_ listener be fired if the value isn't different?

----

What are the benefits for creating a new listener? What are the drawbacks of changing the existing listeners?

src/config.js
```diff
@@ -414,6 +414,7 @@ class Config {
       this.schema.properties.core.properties.projectHome = projectHomeSchema
       this.defaultSettings.core.projectHome = projectHomeSchema.default
     }
+    this.indicator = []
```
thomasjo: What is up with this property? It only seems to store "magic numbers" (1, 2, and 3), but the value is never read, and the array is never cleared, so this will keep accumulating values. Is this debugging code you have simply forgotten to remove?

src/config.js
```diff
@@ -1049,16 +1095,20 @@ class Config {
...
-    return this.emitChangeEvent()
+    return this.emitChangeEvent() && this.emitUpdateEvent(value)
   }
 
   observeKeyPath (keyPath, options, callback) {
     callback(this.get(keyPath))
     return this.onDidChangeKeyPath(keyPath, event => callback(event.newValue))
   }
 
+  listenKeyPath (keyPath, options, callback) {
```
thomasjo: This function name doesn't really make sense as far as readability is concerned. Essentially all other functions in this class read like English, but this function does not.

A better name might be something along the lines of `listenForKeyPath`.

thomasjo (changes requested):
Besides my specific comments, I have concerns about the need for this functionality. Like @Arcanemagus has already mentioned, why not use the much more appropriate `observe` functionality.

Please flesh out the template with more details such as drawbacks, alternative designs/implementations, etc.

src/config.js
```diff
@@ -435,6 +436,7 @@ class Config {
     this.pendingOperations = []
     this.legacyScopeAliases = new Map()
     this.requestSave = _.debounce(() => this.save(), 1)
+
```
sstt011: This blank line should be removed.

src/config.js
```diff
@@ -701,6 +747,7 @@ class Config {
 
     if (!source) source = this.mainSource
 
+
```
sstt011: This blank line should be removed.

src/config.js
```diff
@@ -1049,16 +1095,20 @@ class Config {
...
     return this.onDidChangeKeyPath(keyPath, event => callback(event.newValue))
   }
 
+  listenKeyPath (keyPath, options, callback) {
+    return this.onDidUpdateKeyPath(keyPath, event => callback(event.newValue))
+  }
+
   onDidChangeKeyPath (keyPath, callback) {
-    let oldValue = this.get(keyPath)
+    let oldValue = this.get(keyPath);
```
sstt011: Atom discourages the use of semicolons.
Before:
`let oldValue = this.get(keyPath);`
After:
`let oldValue = this.get(keyPath)`

src/config.js
```diff
@@ -1069,6 +1119,17 @@ class Config {
     })
   }
 
+  onDidUpdateKeyPath (keyPath, callback) {
+    const oldValue = this.get(keyPath)
```
sstt011: On lines 1128 and 1372 you try to assign `oldValue`. To do this, you need to remove `const`.
Before:
`const oldValue = this.get(keyPath)`
After:
`oldValue = this.get(keyPath)`

src/config.js
```diff
@@ -1223,7 +1284,15 @@ class Config {
   }
 
   emitChangeEvent () {
-    if (this.transactDepth <= 0) { return this.emitter.emit('did-change') }
+    if (this.transactDepth <= 0) {
+      return this.emitter.emit('did-change') && this.emitter.emit('did-update')
+    }
+  }
+
+  emitUpdateEvent (keypath='', newValue={}) {
```
sstt011: The `=` signs should have spaces before and after them.
Before:
`emitUpdateEvent (keypath='', newValue={}) {`
After:
`emitUpdateEvent (keypath = '', newValue = {}) {`

fosslinux (changes requested):
These changes should fix the linter.

sadick254:
Thank you for your PR!

We haven't gotten a response to the reviews in the comments above. I'm going to close this due to inactivity, but don't hesitate to reach out if you have the time to resolve the conflicts and review comments, we'll be happy to reopen the PR.

Issue #16590, opened 2018-01-18
Editing is slow when parsing the wrong language


Ben3eeE:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

![cut crash](https://user-images.githubusercontent.com/1058982/35115023-0753ff38-fc87-11e7-910a-8cbc5d8c13b6.gif)

### Steps to Reproduce

1. Enable the `Use Tree Sitter Parsers` setting
1. Copy https://raw.githubusercontent.com/atom/language-php/3e45569d8fb3af6747f6e4208b16d17deb577b97/spec/php-spec.coffee to an untitled buffer
1. Change language mode to JavaScript
1. Select all text and `core:cut`
1. Atom crashes

**Expected behavior:** Atom to not crash when cutting text

**Actual behavior:** Atom crashes

**Reproduces how often:** 100%

### Versions

**Atom**: 1.25.0-dev-3d8902d x64
**Electron**: 1.7.10
**OS**: Microsoft Windows 7 Professional

### Additional Information

/cc: @maxbrunsfeld

Ben3eeE:
I can not reproduce the crash on latest master but I see a significant slowdown when editing this file.

The slowdown is not reproducible when disabling tree-sitter.

Arcanemagus:
I'm also unable to reproduce on v1.30.0-nightly15, marking as closed. We can always re-open if this is found to still be an issue.

PR #17339, opened 2018-05-16
Perform parsing off the main thread when Tree-sitter is enabled

maxbrunsfeld:
Refs https://github.com/atom/atom/issues/16590

### Overview

Currently, all parsing in Atom takes place on the main thread, whether you're using the new Tree-sitter system or the default TextMate system. In some circumstances, parsing can take long enough that it reduces the app's responsiveness. This PR makes it so that when you've enabled Tree-sitter, parsing will take place on a background thread so that it can never delay UI responses.

This will also improve Atom's overall parsing speed, because the native parser is now reading directly from the native [`superstring`](https://github.com/atom/superstring) `TextBuffer` instance without any intermediate C++ -> JS -> C++ calls which add overhead and generate garbage.

### Tasks

* [x] Update Atom to use the new Tree-sitter API that allows for multi-threaded parsing
* [x] When the buffer changes, parse asynchronously
* [x] When the buffer changes *during* a parse, enqueue another parse and record the changes
* [x] Parse buffers asynchronously when initially opening them

### Related PRs

https://github.com/tree-sitter/tree-sitter/pull/165
https://github.com/tree-sitter/node-tree-sitter/pull/11
https://github.com/atom/superstring/pull/65
https://github.com/tree-sitter/tree-sitter/pull/170


maxbrunsfeld:
When editing, parsing no longer impacts the app's frame rate at all:

high-frame-rate

maxbrunsfeld:
Another interesting twist: parsing asynchronously is great for Atom's frame rate and responsiveness, but can cause Atom to do *more* work overall (bad for battery life) because for many edits, it has to render twice: once immediately after the edit, and then again once the parsing completes in order to update the syntax highlighting.

Here's a flame graph of this behavior:

sync-operations-before

I've just added a refinement to the algorithm to address this. Now, we perform a limited amount of parsing work immediately on the main thread. If this completes, we can re-render with the correct highlighting from the start. Only when parsing takes a long time do we need to do in the background. The underlying Tree-sitter API is described in [this PR](https://github.com/tree-sitter/tree-sitter/pull/170).

Here's the new flame graph for the same editing action as above (the scale is different):

sync-operations-after

Parsing and resolving the parse promise only took half of a millisecond, so it's better for the user's energy usage to finish the parse up front and render only once.

maxbrunsfeld:
The amount of synchronous work that we allow the parser to do is specified as a number of abstract 'parsing operations'. These don't correspond directly to any unit of time, but they just serve as a rough way to limit the the work we do that has very low overhead to keep track of.

I've currently hard-coded it to 1000 parsing operations. If we want, we could do something more sophisticated like measuring how long parsing operations take on average on the user's machine and tuning the limit appropriately. I don't think the exact number is too important though. We just want *some* limit on the amount of synchronous work we'll do.

PR #17362, opened 2018-05-18
⬆️ electron@2.0.1

jasonrudolph:
Release notes: https://electronjs.org/releases#2.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "2.0.0",
+  "electronVersion": "2.0.1",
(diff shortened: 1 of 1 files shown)


daviwil (approved):
:+1:

the-j0k3r:
So this is the electron in 1.29.0 is v2.0.1 but the electron in 1.28.0 was version 2.0 then 1.28.1 was electron 2.0.4 then 1.28.2 was electron 2.0.5

Can someone explain to me how this makes any sense and actually fixes issues?

the-j0k3r:
AH never mind, the release notes under All Changes are a bunch of arbitrary entries that are actually mostly irrelevant and thus confusing.

PR #17373, opened 2018-05-20
Fix deprecated hidden-inset title bar style

npezza93:
As of https://github.com/electron/electron/pull/11975 the hidden-inset
titleBarStyle has been removed so the custom-inset title bar option
renders a normal title bar which is not desired.


Arcanemagus:
This looks like it might be the cause of https://github.com/atom/atom/issues/17377?

npezza93:
@Arcanemagus Yea that's correct

daviwil (approved):
Looks right to me: https://electronjs.org/docs/api/frameless-window#hiddeninset

PR #17379, opened 2018-05-21
Add proposed RFC 'Updateable Bundled Packages'

daviwil:
This feature will enable an opt-in subset of bundled Atom packages to be updated with `apm` outside of the Atom release cycle.  This will enable users to receive new functionality and bug fixes for some bundled packages as regularly as needed without waiting for them to be included in a new Atom release. 

[**Rendered view**](https://github.com/atom/atom/blob/dw-updateable-packages-rfc/docs/rfcs/001-updateable-bundled-packages.md)


docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+2. The user downloads and installs Atom 1.28.0 from atom.io which includes GitHub package version 0.15.0.  Two weeks later, GitHub package 0.16.0 is released with a few new features.  The user is prompted to update to the new version and gets the new features even though Atom 1.29.0 hasn't been released yet.
+
+3. In the future, a user has an old install of Atom 1.28.0 and waits a long time between installing Atom updates.  The GitHub package releases version 0.25.0 but the user is not prompted to install it because the GitHub package has set `engines` in `package.json` to restrict to Atom 1.32.0 and above.
+
+### Rules for Updateable Bundled Packages
+
+Any package that opts into this behavior must follow one rule: **its `engines` field must be regularly updated to reflect the necessary Atom version for the Atom, Electron, and Node.js APIs used in the package**.  This field defines the range of Atom versions in which the package is expected to work.  The field should always be set to the lowest possible Atom version that the package supports.
```
smashwilson: 👍 I like this idea.

As a safety guard against regressions, we could include a CI build against the oldest supported major Atom version in our build matrix. When the oldest build fails, we'd know that we're depending on new functionality and we need to bump the `engines` key.

There may need to be additional work to our CI scripts to support this. atom/github uses its own build scripts to deal with its transpilation pipeline (I think) so we would implement the change there.

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+If a package wants to use API features of a newer version of Atom while still supporting older Atom versions, it must do so in a way that is aware of the user's version and adjust itself accordingly.
+
+## Drawbacks
+
+> Why should we *not* do this?
+
+The primary drawback of this approach is that updateable bundled packages might exhibit problems on older Atom versions due to missing or changed APIs in Atom, Electron, or Node.js.  The solution for these packages is to keep their `engines` field updated appropriately, but there's still a chance that some updates will slip through without the necessary engine version changes.
+
+One other possible drawback is that an updated version of a bundled package might not be compatible across two different Atom channels.  For example, if the user installs a new update to a bundled package that only supports the current Atom Beta release or higher, the user will no longer have access to that package if they open Atom Stable.
```
smashwilson: Until we have support for multiple release channels in apm (for example), this'll be a problem. We'll just need to always support the latest stable. Notice that [we already do this](https://travis-ci.org/atom/github) with a `CHANNEL=stable` build.

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Rationale and alternatives
+
+> - Why is this approach the best in the space of possible approaches?
+> - What other approaches have been considered and what is the rationale for not choosing them?
+> - What is the impact of not doing this?
+
+This is the best approach for updating bundled packages because it allows those packages to take control of their own release cycle so long as they manage their Atom engine version correctly.  It also does so in a way that allows us to decide which packages can be updated independently, reducing the likelihood of problems for users.
+
+The primary alternative to this approach is to speed up the Atom release cycle so that bundled Atom package updates will reach users more frequently.  This approach will be investigated independently of this RFC as it may still be valuable even with updateable bundled packages.
```
smashwilson: 👍

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Unresolved questions
+
+> - What unresolved questions do you expect to resolve through the RFC process before this gets merged?
+
+Is it enough to just depend on the `engines` field of `package.json` to protect users from installing a package update that doesn't work with their version of Atom?
+
+Is `updateable` the right name for the field in `package.json`?  Is there a clearer name?
```
smashwilson: Naming things is hard 😆 

I think `updateable` is fine, personally.

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+> - What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of Atom?
+
+Can package authors ship updates to stable-only and beta-only versions of their packages simultaneously?  For example, can the GitHub package keep shipping hotfixes to 0.14.x which targets Atom >=1.27.0 while also shipping updates to 0.15.x which targets >=1.28.0?
```
smashwilson: @annthurium brought up introducing some infrastructure to support feature flagging at minisummit. When we tackle that, we could introduce the ability to mark features as "enabled by default on stable" and "enabled by default on beta" as separate feature flag states to handle this. 🤔 

I think "shipping new features to stable as well by default" would also not be a bad place to be in. But I think I lean to a little more risk tolerance 😆

smashwilson (approved):
:100:

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+The primary use case for this improvement is enabling the GitHub package to ship improvements more frequently than Atom's release cycle since many of its improvements can be done without changes to Atom itself.  If this approach is proven to work well for the GitHub package, we might also consider using it to ship Teletype as a bundled Atom package.
+
+## Explanation
+
+> Explain the proposal as if it was already implemented in Atom and you were describing it to an Atom user. That generally means:
+> - Introducing new named concepts.
+> - Explaining the feature largely in terms of examples.
+> - Explaining any changes to existing workflows.
+
+Bundled packages are treated differently than community packages that you can install using `apm`:
```
Arcanemagus: This section describes how it works _currently_, but it's supposed to describe how it _will work_ in the future if this is accepted/implemented.

docs/rfcs/001-updateable-bundled-packages.md
Arcanemagus: Maybe `coreUpdateable`? We don't want community package authors looking at a core package as an example to see that and think "Oh, I need to add this or my package won't get updates!".

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+Despite all this, if the user *does* install a bundled package using `apm`, it will be loaded into the editor and updated dutifully as releases occur.
+
+### Implementation Details
+
+Because the necessary infrastructure is already in place to enable updates to bundled packages using `apm`, the only work required is to provide a way for packages to opt in to this behavior and for `apm` to include those packages in its update checks if they haven't already been installed in the user's packages folder.
+
+Any bundled Atom package will be able to opt in to updates by adding `"updateable": true` to its `package.json` file.  This will cause `apm` to consider it as part of the list of packages it checks for updates.  If a community (non-bundled) package sets this field to `true` or `false` it will be ignored as it's only relevant to bundled packages.
+
+`apm` will be updated to include the list of bundled packages with `"updateable": true` set in their `package.json` so that the user will be notified of new package versions that support the engine version of their current Atom build.
```
Arcanemagus: `dalek` will also need to be updated to stop alerting on these packages.

docs/rfcs/001-updateable-bundled-packages.md
annthurium: We can update our documentation and maybe even write a blog post to communicate that `updateable` is only for atom core packages.

jasonrudolph: screen shot 2018-05-22 at 10 04 03 am

😆 :trollface:

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+> One paragraph explanation of the feature.
+
+This feature will enable an opt-in subset of bundled Atom packages to be updated with `apm` outside of the Atom release cycle.  This will enable users to receive new functionality and bug fixes for some bundled packages as regularly as needed without waiting for them to be included in a new Atom release.  This is especially important for packages like [GitHub](https://github.com/atom/github/) and  [Teletype](https://github.com/atom/teletype/) which provide essential Atom functionality and could be improved independently of Atom.
+
+## Motivation
+
+> Why are we doing this? What use cases does it support? What is the expected outcome?
+
+Atom currently uses a monthly release cycle with staged Stable and Beta releases so that major issues get caught early in Beta before reaching the Stable release.  Because Atom releases updates monthly, this means that a new feature merged into `master` right after a new Atom release could take one month to reach the next Beta and then another month to reach Stable.
```
jasonrudolph: > Because Atom releases updates monthly, this means that a new feature merged into `master` right after a new Atom release could take one month to reach the next Beta and then another month to reach Stable.

This is the primary reason that Teletype is not bundled with Atom. Since its initial launch, we've shipped a Teletype release almost every week, delivering new features and bug fixes as soon as they're ready, as opposed to waiting up to 2 months to get each release into users' hands.

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+Since a large part of Atom's built-in functionality is provided by bundled packages, it makes sense to allow some of those packages to be updated independently of Atom's monthly release cycle so that users can receive new features and fixes whenever they become available.
+
+The primary use case for this improvement is enabling the GitHub package to ship improvements more frequently than Atom's release cycle since many of its improvements can be done without changes to Atom itself.  If this approach is proven to work well for the GitHub package, we might also consider using it to ship Teletype as a bundled Atom package.
+
+## Explanation
+
+> Explain the proposal as if it was already implemented in Atom and you were describing it to an Atom user. That generally means:
```
jasonrudolph: The text below explains the proposal as if it _will_ be implemented. Can I talk you into updating it to explain the proposal as if it was _already_ implemented?

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+- You are not prompted to update them when new versions are released on `apm`
+- `apm` will warn you at the command line when you try to install or update a bundled package
+- If a user intentionally installs a bundled package from `apm` the [Dalek package](https://github.com/atom/dalek/) will show a warning in the "deprecations" view asking the user to remove the offending package
+
+Despite all this, if the user *does* install a bundled package using `apm`, it will be loaded into the editor and updated dutifully as releases occur.
+
+### Implementation Details
+
+Because the necessary infrastructure is already in place to enable updates to bundled packages using `apm`, the only work required is to provide a way for packages to opt in to this behavior and for `apm` to include those packages in its update checks if they haven't already been installed in the user's packages folder.
```
jasonrudolph: > ... if they haven't already been installed in the user's packages folder.

I don't understand the relevance of this clause. (Sorry. 😬) Shouldn't `apm` always include these packages when checking for updates?

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+Any bundled Atom package will be able to opt in to updates by adding `"updateable": true` to its `package.json` file.  This will cause `apm` to consider it as part of the list of packages it checks for updates.  If a community (non-bundled) package sets this field to `true` or `false` it will be ignored as it's only relevant to bundled packages.
+
+`apm` will be updated to include the list of bundled packages with `"updateable": true` set in their `package.json` so that the user will be notified of new package versions that support the engine version of their current Atom build.
+
+### User Experience Examples
+
+1. The user downloads Atom 1.28.0 from atom.io which includes GitHub package version 0.15.0.  After Atom 1.28.0 was released, a hotfix release was shipped for the GitHub package as 0.15.1.  When the user installs and starts Atom, they are prompted to install the update to the GitHub package.
```
jasonrudolph: Does this example imply that GitHub package 0.15.1 is released _between_ A) the time that the user downloads Atom 1.28.0 and B) the time they install Atom 1.28.0? If so, I think this could be written more generally to describe any scenario where GitHub package 0.15.1 is shipped after Atom 1.28.0 is released:

> The user downloads and installs Atom 1.28.0 from atom.io which includes GitHub package version 0.15.0. After Atom 1.28.0 was released, a hotfix release was shipped for the GitHub package as 0.15.1. When the user starts Atom, they are prompted to install the update to the GitHub package.

What do you think?

docs/rfcs/001-updateable-bundled-packages.md
jasonrudolph: > its `engines` field must be regularly updated to reflect the necessary Atom version for the Atom, Electron, and Node.js APIs used in the package

If I'm understanding correctly, it's not that the `engines` field needs to be regularly updated; it's that it needs to be correct each time we publish a new version of the package. Is that right? If so, perhaps we could rephrase this like so:

> each release must ensure that the `engines` field reflects the necessary Atom version for the Atom, Electron, and Node.js APIs used in the package

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+Any package that opts into this behavior must follow one rule: **its `engines` field must be regularly updated to reflect the necessary Atom version for the Atom, Electron, and Node.js APIs used in the package**.  This field defines the range of Atom versions in which the package is expected to work.  The field should always be set to the lowest possible Atom version that the package supports.
+
+If a package wants to use API features of a newer version of Atom while still supporting older Atom versions, it must do so in a way that is aware of the user's version and adjust itself accordingly.
+
+## Drawbacks
+
+> Why should we *not* do this?
+
+The primary drawback of this approach is that updateable bundled packages might exhibit problems on older Atom versions due to missing or changed APIs in Atom, Electron, or Node.js.  The solution for these packages is to keep their `engines` field updated appropriately, but there's still a chance that some updates will slip through without the necessary engine version changes.
```
jasonrudolph: > there's still a chance that some updates will slip through without the necessary engine version changes.

If this happens, how would we resolve the problem?

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+One other possible drawback is that an updated version of a bundled package might not be compatible across two different Atom channels.  For example, if the user installs a new update to a bundled package that only supports the current Atom Beta release or higher, the user will no longer have access to that package if they open Atom Stable.
+
+However, this drawback is no different than what the user would face today installing a community package under the same circumstances, so this could be considered a general problem in the Atom package ecosystem.
```
jasonrudolph: This sentence seems to apply to the preceding paragraph. Would it make sense to move this sentence to end of the preceding paragraph?

docs/rfcs/001-updateable-bundled-packages.md
jasonrudolph: > Can package authors ship updates to stable-only and beta-only versions of their packages simultaneously?

I don't think anything prevents us from _shipping_ versions in this way. The issue is more about how it affects people installing and using these versions, right? For example, if I update the github package, that update is presumably stored at `~/.atom/packages/github`. If the version at `~/.atom/packages/github` is only compatible with Atom >=1.28, and I open Atom 1.27.0, what happens?

jasonrudolph (commented):
Thanks for putting this together, @daviwil. :bow:

I've included inline questions and suggestions in the review below. Additionally, I'm wondering about a couple other things:

1. As I understand it, the bundled packages are snapshotted, which makes Atom's startup time significantly faster. If I update to a newer version of a bundled package, what impact will that have on Atom's performance and the package's performance?

2. Assume Atom 1.28.0 ships with GitHub package 0.16.0. Assume GitHub package 0.18.0 is later released, and I update to use GitHub package 0.18.0 with Atom 1.28.0. Assume Atom 1.29.0 ships with GitHub package 0.17.0. When I install and launch Atom 1.29.0, what version of the GitHub package will I be using, and why is that the appropriate choice?

docs/rfcs/001-updateable-bundled-packages.md
jasonrudolph: > We'll just need to always support the latest stable.

@smashwilson: Good point. :zap:

@daviwil: I think it's worth including this ☝️ in the RFC. In other words, when we ship a new version of a bundled package, it's fine for the package to support things that are only present in the beta or dev channels, but the package should _function_ when used with the stable channel.

docs/rfcs/001-updateable-bundled-packages.md
maxbrunsfeld: The single biggest concern for me is how out-of-band updates would impact v8 snapshotting of the github package in particular. Are we going to try to share *some* snapshotted code but not all, switch to a completely different strategy of computing startup snapshots where we ship `mksnapshot` and actually compute them on the user's machine, or just lose the benefit of startup snapshots if the user upgrades their bundled package.

None of those sound very appealing to me. Are there alternatives I haven't thought of?

Arcanemagus: It would take a lot of work in APM and Atom to handle it (and probably an even newer version of Electron? This shipped in Chrome 65), but [Lazy deserialization](https://v8project.blogspot.com/2018/02/lazy-deserialization.html) at least _sounds_ like it might be a good fit for that issue if we can somehow ship snapshots for each package.

maxbrunsfeld: Lazy deserialization seems like it will benefit us, but I think it's orthogonal to the concern about how to update Atom's startup snapshot when a bundled packages is updated, right?

Arcanemagus: Mostly, there's no way to update the startup snapshot without bundling `mksnapshot` and regenerating it every time a package installs/updates/is removed. Assuming I'm not completely wrong with how that works though, once support for that is in the Electron version we are using I don't think we would _need_ to regenerate the startup snapshot, as each package would provide its own snapshot that would be loaded on the fly.

That's all way in the future enhancements though, I think your concern here is very valid and there isn't currently anyway around losing the benefit of the startup snapshot for packages if they are updated outside of core.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, will include this as a rule!

docs/rfcs/001-updateable-bundled-packages.md
daviwil: I like `coreUpdateable`, makes it a bit more clear.

On the topic of `updateable` vs `updatable`, somehow without the 'e' it looks weird to me but I can be convinced of `updatable` if more folks prefer that.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Feature flags would be super nice for adding preview features that could be accessible in Stable, would love to see that happen!

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, will mention that.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: I'll tighten up the language here so that it's clearer!

Basically what I was trying to convey is that `apm` will check bundled packages for updates *right now* if you've installed one manually with `apm` and it's in your `~/.atom/packages` folder, but it needs to be taught how to find bundled packages to update when they haven't been installed using `apm` yet.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Thanks for bringing this up @maxbrunsfeld, I will capture this concern in the RFC because it's likely to be a problem.  Before I finish up the document I'll do some research on how much start up time is impacted by using a non-snapshotted GitHub package then I'll add it here.

If we decide that the cost of not being able to update the snapshot is too high to consider making this change right now, I'm totally on board with postponing or rejecting this RFC and looking for a better overall approach.

Definitely curious to look into @Arcanemagus's suggestion as a future optimization when we're running on more recent Chrome relases.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, I phrased that from the wrong perspective, will update it accordingly.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, I'll reword it and move some stuff around to make it more focused.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: :+1:

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Actually, based on someting @smashwilson [pointed out before](https://github.com/atom/atom/pull/17379#discussion_r189599200) I think the scenario I'm positing here isn't even possible so I'll just remove it :scissors:

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Thanks!  I'll update the wording here.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Looks like I worded this one poorly, here's the scenario:

1. Atom 1.28.0 is released
2. One week later GitHub package 0.15.1 is released
3. Two weeks later a user installs Atom 1.28.0
4. Upon starting Atom, user is prompted to install GitHub 0.15.1.

Updated wording to better reflect the scenario:

> The user downloads and installs Atom 1.28.0 which was released two weeks prior.  Since that release the GitHub package has been updated to version 0.15.1 on `apm`.  When the user starts Atom for the first time, they are prompted to update the GitHub package.

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, originally it was part of that paragraph but it felt like the paragraph was too long.  I'll join them back together.  :heavy_check_mark:

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Added a solution for this situation in the updated document :heavy_check_mark:

daviwil:
Thanks for the feedback so far!  I've updated the RFC in response to many of the comments so please take another look when you have a chance.

In response to @jasonrudolph's questions:

1. I'll do some analysis on this tomorrow and update the document to capture the findings.

2. Atom 1.29.0 will load GitHub package version 0.18.0 because it will already be in the user's packages folder.  This is considered the correct behavior because Atom would offer an update to 0.18.0 soon after the user started up the new version.

After this change, the latest compatible version of a bundled package should be considered the "right" one.  Ideally we'd make sure that updatable bundled packages are always set to the latest release before shipping an update so that the user isn't immediately prompted to update a package that should be up to date in the box (especially when you consider the perf cost of losing the snapshotted code).

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,92 @@
...
+## Explanation
+
+> Explain the proposal as if it was already implemented in Atom and you were describing it to an Atom user. That generally means:
+> - Introducing new named concepts.
+> - Explaining the feature largely in terms of examples.
+> - Explaining any changes to existing workflows.
+
+Any bundled Atom package can opt in to new updates released via `apm` by adding `"coreUpdateable": true` to its `package.json` file.  This causes `apm` to consider it as part of the list of packages it checks for updates.  If a community (non-bundled) package sets this field to `true` or `false` it will be ignored as it's only relevant to bundled packages.
+
+`apm` includes each updatable bundled package in its list of.  The user is notified of new updates to the package so long as the update supports the engine version of their current Atom build.
```
jasonrudolph: > `apm` includes each updatable bundled package in its list of.  

Looks like this sentence got cut off.

jasonrudolph:
> I've updated the RFC in response to many of the comments so please take another look when you have a chance.

I've reviewed all the updates, and they look good to me.

>> Assume Atom 1.28.0 ships with GitHub package 0.16.0. Assume GitHub package 0.18.0 is later released, and I update to use GitHub package 0.18.0 with Atom 1.28.0. Assume Atom 1.29.0 ships with GitHub package 0.17.0. When I install and launch Atom 1.29.0, what version of the GitHub package will I be using, and why is that the appropriate choice?
>
> Atom 1.29.0 will load GitHub package version 0.18.0 because it will already be in the user's packages folder. This is considered the correct behavior because Atom would offer an update to 0.18.0 soon after the user started up the new version.
>
> After this change, the latest compatible version of a bundled package should be considered the "right" one. 

How would you feel about incorporating this rule into the RFC?

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Thanks for catching that!

daviwil:
Thanks @jasonrudolph!  Added a third rule that states the package should always be kept up to date in Atom's package.json, let me know if it's clear!

I also added some startup timing details when using the snapshotted vs non-snapshotted code of the GitHub package, definitely take a look at that too.

docs/rfcs/001-updateable-bundled-packages.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Rules for Updateable Bundled Packages
+
+Any package that opts into this behavior must adhere to these rules:
+
+1. **Each release must ensure that its `engines` field in `package.json` reflects the necessary Atom version for the Atom, Electron, and Node.js APIs used in the package**.  This field defines the range of Atom versions in which the package is expected to work.  The field should always be set to the lowest possible Atom version that the package supports.
+
+2. **Any new update to a bundled package *must* support current Stable *and* Beta releases**.  This enables the user to upgrade the package and continue to use it in side-by-side Stable and Beta installs on their machine.  If a package wants to use API features of a newer version of Atom while still supporting older Atom versions, it must do so in a way that is aware of the user's version and adjust itself accordingly.
+
+3. **Atom's `package.json` *must* stay up to date with the latest supported version of the package** in the `master` and Beta release branches.  This ensures that the user always gets the latest version of the package in a new release and also benefits from its inclusion in Atom's snapshot.
```
jasonrudolph: @daviwil: I think I understand the motivation for this rule, but if a human needs to manually bump the package version on atom's master and beta branches each time a new version of the package is released, I'm skeptical that this rule will be reliably followed. I think we'd need some automation to either perform this step for us, or alert us when this rule is violated so that we can remedy the situation.

What do you think?

Hendis:
[npm-debug.log](https://github.com/atom/atom/files/2087123/npm-debug.log)

Hendis:
octocat-spinner-128

docs/rfcs/001-updateable-bundled-packages.md
daviwil: Yep, it'd be nice if we had something like [Greenkeeper](https://greenkeeper.io/) for Atom packages.  Actually, I think it wouldn't be hard to have a special build automation for that which only runs a check against known updatable bundled packages (i.e. GitHub and possibly Teletype) and bumps Atom's package.json dependencies when a newer compatible version is available.

I'll add a blurb about that to the RFC.

daviwil:
OK folks, I think this RFC is ready for final review/signoff!  Please let me know if there's any more feedback or concerns you'd like me to address.

I think we should be able to [mitigate the startup time increase](https://github.com/atom/atom/blob/dw-updateable-packages-rfc/docs/rfcs/001-updateable-bundled-packages.md#increased-atom-startup-time) for an updated GitHub package in Atom without snapshotting it, so I feel confident that we can move forward with implementing this RFC without too much downside.

If folks are on board with proceeding with this RFC, I'll change its status to Accepted, merge it, and then plan to start implementing it in an upcoming sprint.

annthurium (approved):

winstliu (approved):

daviwil:
I've set the RFC back to Proposed on `master` for the time being.

Hendis:
add proposed

PR #17402, opened 2018-05-24
Prevent creation of files with reserved names on Windows

### LINKED ISSUE #17175: Atom can create files with reserved/invalid filenames in Windows 10
### Description

Atom allows the user to create files with reserved filenames in Windows 10, which results in an unusable file that cannot be deleted, renamed, or moved. Reserved filenames in Windows 10 are as follows:

- CON
- PRN
- AUX
- NUL
- COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM0
- LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9, LPT0

### Steps to Reproduce

1. In Windows 10, open any directory in Atom
2. In the file tree pane, right-click and select "New File"
3. In the field that appears, enter "Aux.txt", or any of the other filenames listed above with any file extension, and press Enter
4. Observe that the newly created file cannot be opened, deleted, renamed, moved, or otherwise meaningfully interacted with

**Expected behavior:** Atom should throw an error when the user attempts to create a file with a reserved filename, and prohibit the creation of the file

**Actual behavior:** A file is created which cannot be deleted without first booting into a different operating system, such as Linux

### Versions

1.25.0 x64 - Windows 10
1.26.0 x64 - Windows 10

FabianPetersen:
### Description of the Change
The change prevents users from creating files with restricted filenames in windows.

### Alternate Designs
The usage of the slash library could be omitted for a solution based on regular expressions. However, this seemed like the better solution as covers all the edge cases.

It's problematic that users can accidentally create files which can't be edited, moved or deleted. As this may cause much confusion for the average user.

### Benefits
Atom will no longer be able to create files with restricted names on windows 10 machines.

### Possible Drawbacks
1. The drawbacks are slightly more code on each saveAs operation. This makes the operation a touch slower on windows machines, however the greater scheme of things. The performance loss should be negligible.
2. Two new dependencies (os, slash)

### Verification Process
Steps to Reproduce the issue
1. In Windows 10, open any directory in Atom
2. In the file tree pane, right-click and select "New File"
3. In the field that appears, enter "Aux.txt", or any of the other filenames listed above with any file extension, and press Enter
4. Observe that the newly created file cannot be opened, deleted, renamed, moved, or otherwise meaningfully interacted with

After the fix, an error is thrown instead of the file being created.

### Applicable Issues
Atom can create files with reserved/invalid filenames in Windows 10 #17175

### DIFF
--- a/src/text-editor.js
+++ b/src/text-editor.js
@@ -12,6 +12,8 @@ const Selection = require('./selection')
 const NullGrammar = require('./null-grammar')
 const TextMateLanguageMode = require('./text-mate-language-mode')
 const ScopeDescriptor = require('./scope-descriptor')
+const os = require('os')
+const slash = require('slash')
 
 const TextMateScopeSelector = require('first-mate').ScopeSelector
 const GutterContainer = require('./gutter-container')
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for your submission :+1:

While it would be useful to have something like this to prevent this problem, this PR won't be accepted until:

1. All of the current CI errors are fixed
1. It contains tests that verify the functionality works as expected

Additionally, the use of the slash library is unnecessary. See Node's [`path.basename` function](https://nodejs.org/dist/latest-v8.x/docs/api/path.html#path_path_basename_path_ext) for an easy replacement that should work no matter which directory separator is used (as noted in [the Atom Flight Manual section on cross-platform considerations](https://flight-manual.atom.io/hacking-atom/sections/cross-platform-compatibility/#file-paths)).

Also, is case significant? Or is `con` an invalid filename as well as `CON`? Since case isn't significant in Windows filenames to my recollection, this fix should take that into account as well.

fosslinux (changes requested):
These changes should fix the linter errors.

sadick254:
Thank you for your PR!

We haven't gotten a response to our comment above. With only the information that is currently in the PR, we don't have enough information to take action. I'm going to close this but don't hesitate to reach out if you have the time to address the comment and resolve the conflicts on this PR, we'll be happy to reopen the PR.

PR #17403, opened 2018-05-24
Change atom.sh not to spawn a sleep process every second on --wait

lllusion3469:
### Description of the Change
As sleep(1) is not a bash builtin, every second a new process is spawed.
To  prevent this, the POSIX read can be used instead.
Since it is (required to be) a bash builtin, it is immediately killed along with bash unlike a longer running sleep would be.
In case stdin is e.g. /dev/null for whatever reason (this would break `EDITOR=nano`), sleep is still kept to prevent a tight loop.

### Alternate Designs
* a longer sleep interval: this would require job management as sleep would keep running even if bash is killed and thus cause huge delays

not possible otherwise

### Benefits
Those sleep processes can be annoying in e.g. `execsnoop` or even `htop`

### Possible Drawbacks
e: does consume stdin, though I don't see how that is a problem

### Verification Process
* edited my distributions `/usr/bin/atom` accordingly (Archlinux, Atom 1.27) (this file hasn't changed since)
* opened a file using `/usr/bin/atom --wait file` and verified that it exited upon closing the tab

### Applicable Issues

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -149,6 +149,7 @@ trap 'on_die' SIGQUIT SIGTERM
 # If the wait flag is set, don't exit this process until Atom tells it to.
 if [ $WAIT ]; then
   while true; do
+    read
     sleep 1
   done
 fi
--- a/atom.sh
+++ b/atom.sh
@@ -148,8 +148,16 @@ trap 'on_die' SIGQUIT SIGTERM
 
 # If the wait flag is set, don't exit this process until Atom tells it to.
 if [ $WAIT ]; then
+  WAIT_FIFO="$ATOM_HOME/.wait_fifo"
   while true; do
-    read
+    [ -f "$WAIT_FIFO" ] && rm "$WAIT_FIFO"
+    [ ! -p "$WAIT_FIFO" ] && mkfifo "$WAIT_FIFO"
+    read < "$WAIT_FIFO" || break
+    sleep 1 # prevent a tight loop
+  done
+  
+  # fall back to sleep
+  while true; do
     sleep 1
   done
 fi


maxbrunsfeld:
> As sleep(1) is not a bash builtin, every second a new process is spawned.

Great point; thanks for bringing this up.

Rather than reading from stdin, what do you think about actually creating a pipe with `mkfifo` and reading from that?

atom.sh
```diff
@@ -148,6 +148,15 @@ trap 'on_die' SIGQUIT SIGTERM
...
 if [ $WAIT ]; then
+  WAIT_FIFO="$ATOM_HOME/.wait_fifo"
+  while true; do
+    [ -f "$WAIT_FIFO" ] && rm "$WAIT_FIFO"
+    [ ! -p "$WAIT_FIFO" ] && mkfifo "$WAIT_FIFO"
+    read < "$WAIT_FIFO" || break
+    sleep 1 # prevent a tight loop
+  done
+  
+  # fall back to sleep
```
maxbrunsfeld: Can you explain why this second "fall back" while loop is still needed?

If the `read` completes (presumably due to an error), won't we `sleep` and then recreate the pipe? I would think that we would already "fall back" to sleep, even without the second `while` loop.

lllusion3469: If we (for some reason) can't create the fifo or remove the file that's there in its place, I guess.

If you want, I can remove it, though.

maxbrunsfeld: My knowledge of bash is somewhat limited, so I might be missing something.

I would think that it'd be impossible to reach that code though, because the `while` loop above would never complete. Is that true?

lllusion3469: there's a break after the `read`

maxbrunsfeld: Oh, thanks for explaining; I didn't notice that.

So why is the `read` wrapped in a `while` loop? In what circumstance would `read` exit successfully?

lllusion3469: For example if you write to the FIFO.
I don't know why you would, again, but it's possible.

maxbrunsfeld: I see, so in that case it should be fine to `read` again rather than falling back to a `sleep`, but we don't want to retry if there appears to be a problem with the fifo. That actually makes perfect sense.

maxbrunsfeld (approved):

maxbrunsfeld:
Another benefit of this PR is that the `atom.sh` process exits instantly after you close the waited buffers, instead of waiting for up to one second for the `sleep` to complete.

MartinBonner:
Version 1.28.2 of atom (which doesn't have this change) works fine as my git commit editor.  Version 1.29.0 does not - git commit hangs indefinitely waiting for the editor to finish.  I am not convinced this change is what causes the problem - but it does seem the most likely point to start looking.  

What diagnostics would be useful to identify what is causing the problem?

lllusion3469:
* What's your operating system?
* What are your `EDITOR`/`VISUAL`/`GIT_EDITOR` environment variables / `core.editor` setting?
* How did you install Atom?

Also, belated  thanks to @maxbrunsfeld for fixing my mistake completely breaking platforms other than linux.

MartinBonner:
Sorry for the belated reply.  The problem is on my work computer, and the weekend got in the way.

Operating system is Ubuntu 14.04.  uname -a says:

    Linux [redacted] 4.4.0-139-generic #165~14.04.1-Ubuntu SMP Wed Oct 31 10:55:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

None of the environment variables are set.  `core.editor` is `atom --wait`

I installed atom with `sudo apt-get atom=1.28.2`.  `dpkg -s` shows:

    Package: atom
    Status: install ok installed
    Priority: optional
    Section: devel
    Installed-Size: 562156
    Maintainer: GitHub <atom@github.com>
    Architecture: amd64
    Version: 1.28.2
    Depends: git, gconf2, gconf-service, libgtk2.0-0, libudev0 | libudev1, libgcrypt11 | libgcrypt20, libnotify4, libxtst6, libnss3 (>= 2:3.22), python, gvfs-bin, xdg-utils, libcap2, libx11-xcb1, libxss1, libasound2 (>= 1.0.16), libxkbfile1
    Recommends: lsb-release
    Suggests: libsecret-1-0, gir1.2-gnomekeyring-1.0
    Description: A hackable text editor for the 21st Century.
     Atom is a free and open source text editor that is modern, approachable, and hackable to the core.

`apt-cache policy atom` suggests I am getting the package from `https://packagecloud.io/AtomEditor/atom/any/`

I can get you the corresponding output for atom 1.29.0 if that would be helpful.

lllusion3418:
I finally came around to installing a VM on Ubuntu 14.04 and I can indeed replicate this with Atom 1.29.0...
Fiddling around with `/usr/bin/atom` seems to indicate that is this PR that caused it:

Removing `read < "$WAIT_FIFO"` fixes it
and so does `read <> "$WAIT_FIFO"`.

It seems this is due to a change in behavior in bash.

MartinBonner:
Ah HA!  Brilliant thanks.  I can patch that in no problem.
Martin

lllusion3469:
Edited version of my deleted comment:
> I finally came around to installing Ubuntu 14.04 in a VM and I can indeed replicate this with Atom 1.29.0...
> Fiddling around with `/usr/bin/atom` seems to indicate that this PR caused it:
> 
> Removing `read < "$WAIT_FIFO"` fixes it

though it kind of goes against the point of this PR

> and so does `read <> "$WAIT_FIFO"`.

the version I would prefer, though I don't know why it works.


It does not, in fact, seem to have anything to do with bash:
Even with a manually compiled (just `make install`) bash on master branch it doesn't work on Ubuntu 14.04.

I don't really want to create a PR to fix this as long as I don't know what causes this.

lllusion3469:
I still don't get why this is happening, but:

1. `read < "$WAIT_FIFO"` blocks, not because of `read(2)`, but at the  `open(2)` syscall before calling `read`, so any builtin could be used, even `true` and the result would be the same
2. with `read <> "$WAIT_FIFO"` it is important that `read` is used because here it is `read(2)` that blocks since now the FIFO can actually be opened (requires a reader and a writer)
3. on Ubuntu 14.04, for some reason, (only) with `read < "$WAIT_FIFO"` the signal handler doesn't get called and the program resumes where it stopped - at `open(2)`

lllusion3469:
It's [this bash commit](http://git.savannah.gnu.org/cgit/bash.git/commit/?id=591b09c48ffef7e5aeaa28d1fb5181f8dbe66905) that fixed it

PR #17424, opened 2018-05-29
Avoid masking exceptions as cancellations in Workspace.open

maxbrunsfeld:
Previously, when opening a file via `Workspace.open`, if an exception was thrown by `Project.bufferPath` or `TextEditorRegistry.build`, the exception would be dropped, and we would instead throw a special error with the code `CANCELLED`, which instructs [another code path in `Workspace`](https://github.com/atom/atom/blob/f5ee8b5efd2a39efb1ca5ca994e71530485831ad/src/workspace.js#L1204-L1205) to silently halt.

This PR simplifies that `Workspace` method to avoid using a `try` block entirely, so that uncaught exceptions can be surfaced.

I believe this problem was introduced in https://github.com/atom/atom/pull/16229. It's generally not a big deal, because in most cases, the methods I mentioned above won't throw exceptions. @nathansobo noticed this bug when attempting to use an incompatible version of [tree-sitter-rust](https://github.com/tree-sitter/tree-sitter-rust). An exception is thrown when trying to use a parser whose ABI is incompatible with the current version of Tree-sitter.

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1244,37 +1244,26 @@ module.exports = class Workspace extends Model {
 
     const fileSize = fs.getSizeSync(filePath)
 
-    let [resolveConfirmFileOpenPromise, rejectConfirmFileOpenPromise] = []
-    const confirmFileOpenPromise = new Promise((resolve, reject) => {
-      resolveConfirmFileOpenPromise = resolve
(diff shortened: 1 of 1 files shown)


daviwil (approved):
Looks good to me!

PR #17464, opened 2018-06-04
Extend the color profile PR (#17380)

thomasjo:
### Description of the Change

Extends #17380 by wiring up the dialog prompt for restarting Atom whenever the `core.colorProfile` setting is changed. Since this removes the need for the explicit description remark about restarting Atom, that sentence is also removed.

The third (and optional) change alters the bound functions used when wiring up the change event listeners to arrow functions. We can drop that commit if it's deemed risky, not necessary, or some such.

### Alternate Designs

A more elaborate opt-in mechanism for "prompt for relaunch" was considered, but rejected after taking into consideration the concept of [YAGNI™](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

For obvious reasons.

### Benefits

Users do not have to manually restart Atom. Some users will likely not read the description and realize they need to manually restart Atom, which would most likely lead to issues being submitted, etc.

### Possible Drawbacks

More code, and complexity?

### Verification Process

Manually tested on Ubuntu 16.04. Works as expected.

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -174,6 +174,7 @@ class AtomApplication extends EventEmitter {
       this.configFilePromise = this.configFile.watch()
       this.disposable.add(await this.configFilePromise)
       this.config.onDidChange('core.titleBar', this.promptForRestart.bind(this))
+      this.config.onDidChange('core.colorProfile', this.promptForRestart.bind(this))
     }
 
     const optionsForWindowsToOpen = []
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -173,8 +173,8 @@ class AtomApplication extends EventEmitter {
     if (!this.configFilePromise) {
       this.configFilePromise = this.configFile.watch()
       this.disposable.add(await this.configFilePromise)
-      this.config.onDidChange('core.titleBar', this.promptForRestart.bind(this))
-      this.config.onDidChange('core.colorProfile', this.promptForRestart.bind(this))
+      this.config.onDidChange('core.titleBar', () => this.promptForRestart())
+      this.config.onDidChange('core.colorProfile', () => this.promptForRestart())
     }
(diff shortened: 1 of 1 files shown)


src/config-schema.js
```diff
@@ -357,7 +357,7 @@ const configSchema = {
         description: 'Experimental: Use the new Tree-sitter parsing system for supported languages.'
       },
       colorProfile: {
-        description: "Specify whether Atom should use the operating system's color profile (recommended) or an alternative color profile.<br>Changing this setting will require a relaunch of Atom to take effect.",
+        description: "Specify whether Atom should use the operating system's color profile (recommended) or an alternative color profile.",
```
50Wliu: The relaunch line is included in https://github.com/atom/atom/pull/17464/files#diff-8c99efd21425e5773e52ba4fbb820bf0L574, the other setting that requires a restart.  If this one is removed, then the other should be as well.

thomasjo: Oh, I thought I checked that one, but guess I didn't scroll far enough. Good catch, @50Wliu :1st_place_medal: 
Personally I vote for removing it in both places. What do other folks think? /cc @atom/feedback

daviwil: I think it's good to inform the user that a restart will be required even if we're going to prompt them.  This will make sure they understand why they're being prompted for a restart.  I'd say leave the message in for the sake of clarity

daviwil:
This change is simple enough that I don't see any problem merging this PR right after I merge the color profile PR.  I think I'd prefer leaving the "restart" language in the setting description though, just for the sake of clarity.

Thanks again for putting this together!

daviwil:
That was ninja-fast ;)

daviwil (approved):
:bowing_man:

daviwil:
Just waiting on CI to finish before merging this.  Super slow today!

PR #17474, opened 2018-06-05
Don't allow vertical scrollbar to affect soft wrap column

### LINKED ISSUE #13769: Atom keeps trying to wrap long lines on markdown file causing text to blink
### Prerequisites

### Description

When I open an specific file (described below) with an specific windows size (see screencap) atom will try to accommodate the text by wrapping the long lines, then revert it and try again,resulting in the text alternating between this too states in a loop. The bottom half of the text will show and disappear while the top is being rearranged.

### Steps to Reproduce

1. Open this file on atom https://github.com/docker-library/wordpress/blob/d8e5577dcc7096d2b694dda433e2b5c7c4561a82/README.md
2. If the text is still, resize the window until it reproduces the bug.

**Expected behavior:** 
The text should be rendered and stay still until the window is resized.

**Actual behavior:** 
The text will "blink" between these three states.
bug

**Reproduces how often:** 
100%

### Versions

Atom    : 1.13.1
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0

### Additional Information

### LINKED ISSUE #17268: Wrapping line with no spaces causes text "vibration"
### Prerequisites

### Description

If you have a single line of content that has _no_ spaces in it, have wrapping enabled, the line is long enough to be wrapped, and the file is _exactly_ the height of the window, the text will "vibrate" as the cursor is moved around it.

### Steps to Reproduce

1. Open Atom in Safe Mode.
2. Enable soft wrap of text
3. Write a line long enough to wrap
4. Fill the rest of the file with content (blank lines work) till the last line number is _right_ at the bottom of the window. There **must not** be a scroll bar, and there **must be** _less than_ ~50% of an empty line's worth of space between the last line number and the bottom. Shrinking the window till _just_ before it shows a scroll bar seems to be the easiest place.
    * Reproduces: image
    * Fails to reproduce: image
5. Move the text cursor around, or try to select text with the mouse.

**Expected behavior:**

Text rendering to stay the same.

**Actual behavior:**

Text "vibrates" as the wrapping rapidly changes what should be wrapped or not.

**Reproduces how often:**

100% of the time.

### Versions

OS: Windows 10 x64
```
Atom    : 1.27.0-beta1
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.27.0-beta1
python 2.7.14
git 2.17.0.windows.1
visual studio 2015
```

### Additional Information

Possibly related to https://github.com/atom/atom/issues/13769.

maxbrunsfeld:
### Background

Previously, the decision of whether to show or hide scrollbars in a text editor was made by estimating whether or not a scrollbar was *needed* based on the editor's content. It involved a circular data flow:

```
vertical scrollbar visibility
      ↓
editor width
      ↓
horizontal scrollbar visibility
      ↓
editor height
      ↓
vertical scrollbar visibility
```

From what I can tell, there are some subtle issues with the timing of our measurements that currently cause this calculation to produce unstable results, resulting in the issues linked below.

Aside from these rendering bugs, this overall approach for showing and hiding scrollbars seems different from other applications that I'm familiar with. In other applications, there does not seem to be this circular feedback loop.

For example, in `TextEdit.app`, note how the soft wrap behavior is never affected by the vertical scrollbar. Rather, the scrollbars' visibility is controlled by the macOS `Show Scrollbars` setting. They behave as follows:

* When `Show Scrollbars > When Scrolling` (aka 'overlay' mode) is selected, the scrollbars appear and disappear based on the editor's content, but they do not affect the editor's soft wrapping.
* When `Show Scrollbars > Always` (aka 'legacy' mode) is selected, the scrollbars are always visible, so again the editor's soft wrapping does not change based on the height of its content.

text-edit-scrollbar-2

I did a similar test with Sublime Text 3, and I observed a similar behavior. I think that these apps' behavior seems simpler than Atom's from an implementation perspective, and probably better from a UX perspective as well. 

### Description of the Change

In this PR, I've changed Atom's logic for showing and hiding scrollbars in two ways:

1. The editor's soft wrap column always leaves room for the vertical scrollbar, even if it is hidden.
2. In auto-size mode, the editor always grows large enough to account for both scrollbars, even if they are hidden.

### Alternate Designs

We could also fix the bugs referenced below within the current paradigm. @Ben3eeE had [one idea](https://github.com/atom/atom/issues/17268#issuecomment-386596339) for fixing this bug at the cost of performing extra DOM measurements. From my testing, this fix does indeed work.

### Benefits

The reasons that I'm proposing this solution instead is:
1. From what I can tell, it makes the `TextEditor` behave more like other native widgets.
2. It does not require any extra DOM measurements; in fact it results in us doing less work overall in `TextEditorComponent`.

### Possible Drawbacks

This is a bigger change, and it definitely requires careful testing on all three platforms.

### Verification Process

On all three platforms (and with all possible scrollbar settings on mac):

1. Open a file with soft wrapping (e.g. a markdown or text file).
2. Add and remove content at the end of the file so that the vertical scrollbar enables and disables.
3. Check that the soft wrapping does not change.
4. Check that the scrollbars look and behave correctly overall.

Also,

1. Open a file with no soft wrapping and some long lines.
2. Check that the scrollbars look and behave correctly overall.

### Applicable Issues

Fixes #13769
Fixes #17268

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -26,6 +26,7 @@ document.registerElement('text-editor-component-test-element', {
 })
 
 const editors = []
+let verticalScrollbarWidth, horizontalScrollbarHeight
 
 describe('TextEditorComponent', () => {
   beforeEach(() => {
@@ -33,8 +34,15 @@ describe('TextEditorComponent', () => {
 
     // Force scrollbars to be visible regardless of local system configuration
     const scrollbarStyle = document.createElement('style')
-    scrollbarStyle.textContent = '::-webkit-scrollbar { -webkit-appearance: none }'
+    scrollbarStyle.textContent = 'atom-text-editor ::-webkit-scrollbar { -webkit-appearance: none }'
     jasmine.attachToDOM(scrollbarStyle)
+
+    if (verticalScrollbarWidth == null) {
+      const {component, element} = buildComponent({text: 'abcdefgh\n'.repeat(10), width: 30, height: 30})
+      verticalScrollbarWidth = getVerticalScrollbarWidth(component)
+      horizontalScrollbarHeight = getHorizontalScrollbarHeight(component)
+      element.remove()
+    }
   })
 
   afterEach(() => {
@@ -184,8 +192,8 @@ describe('TextEditorComponent', () => {
     })
 
     it('makes the content at least as tall as the scroll container client height', async () => {
-      const {component, element, editor} = buildComponent({text: 'a', height: 100})
-      expect(component.refs.content.offsetHeight).toBe(100)
+      const {component, element, editor} = buildComponent({text: 'a'.repeat(100), width: 50, height: 100})
+      expect(component.refs.content.offsetHeight).toBe(100 - getHorizontalScrollbarHeight(component))
 
       editor.setText('a\n'.repeat(30))
       await component.getNextUpdatePromise()
@@ -201,7 +209,7 @@ describe('TextEditorComponent', () => {
       await setEditorHeightInLines(component, 6)
 
       // scroll to end
-      await setScrollTop(component, scrollContainer.scrollHeight - scrollContainer.clientHeight)
+      await setScrollTop(component, Infinity)
       expect(component.getFirstVisibleRow()).toBe(editor.getScreenLineCount() - 3)
 
       editor.update({scrollPastEnd: false})
@@ -211,7 +219,7 @@ describe('TextEditorComponent', () => {
       // Always allows at least 3 lines worth of overscroll if the editor is short
       await setEditorHeightInLines(component, 2)
       await editor.update({scrollPastEnd: true})
-      await setScrollTop(component, scrollContainer.scrollHeight - scrollContainer.clientHeight)
+      await setScrollTop(component, Infinity)
       expect(component.getFirstVisibleRow()).toBe(editor.getScreenLineCount() + 1)
     })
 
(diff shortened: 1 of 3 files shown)


Arcanemagus:
I'm on Windows, so from the description this would make the scrollbar always visible...

This sounds like an improvement for macOS users and a major regression for everyone else, unless I'm not understanding this correctly?

maxbrunsfeld:
> I'm on Windows, so from the description this would make the scrollbar always visible...

I think it would just leave *room* for the scrollbar, similarly to what happens on mac os, when `Show Scrollbars > Always` is selected. I would think that Chrome would decide whether or not to actually *show* a scrollbar there based on the content height, just like it does for things like the Tree view. I might be misunderstanding how scrollbars behave on Windows though...

Ben3eeE:
I can do a test on Windows tomorrow and a code review if I've got any comments.

thomasjo:
I can test this on Ubuntu 16.04 tomorrow morning (UTC+1), when I get to work.

daviwil (approved):
Code changes look great!  I prefer the reduced complexity of the code here.  I'm running a build to try it out on Fedora Linux right now, will report back

nathansobo:
From your description, this seems like a big improvement. I never consulted other applications when I implemented it, and it never crossed my mind to do it more simply. Thanks!

Arcanemagus:
Behavior looks good to me, I'll leave the code review to others that actually know what it does 😉

as-cii:
This looks reasonable, @maxbrunsfeld. Thanks for jumping on it!

thomasjo:
Quick and dirty test on Ubuntu 16.04

peek 2018-06-06 12-37

Been running this branch for a few hours now, and not noticed any regressions. Seems to work great. Awesome job, @maxbrunsfeld :1st_place_medal:

Ben3eeE:
I've checked it on Windows 10 and it works. I think this change makes sense as well.

wowmediaofficial (approved):

maxbrunsfeld:
> Aside from these rendering bugs, this overall approach for showing and hiding scrollbars seems different from other applications that I'm familiar with. In other applications, there does not seem to be this circular feedback loop.

Actually, I take this back. Regular DOM elements *do* seem to have this circular feedback loop where the state of the vertical scrollbar can affect soft wrap. So it's not as cut and dry as I originally made it sound. Still, in TextEdit, Sublime Text, and VS Code, the vertical scrollbar does *not* affect soft wrap, so I think it's ok for us to adopt this approach too.

nathansobo:
+💯 worse is better.

maxbrunsfeld:
After testing this on windows, I decided to make this even simpler. I removed the conditional logic based on the system-level scrollbar style detection. I've updated the 'Description of the Change' section.

PR #17519, opened 2018-06-16
Toggle comment cursor placement

### LINKED ISSUE #4784: Cursor is not between comment tags when toggling line comment with start/end components
## Problem

When you use the shortcut for comments in a CSS file, your cursor should be placed between the two comment tags. Currently it is placed _after_ the comment tags, which forces the user to go back two characters with the left arrow.

capture d ecran 2014-12-28 a 10 33 20
## Steps to reproduce

Just hit Cmd-Shift-: in a CSS file.
## Config

Atom version 0.165.0
Mac OS X version 10.10.2.

Aerijo:
### Description of the Change

When `toggleLineComments` is triggered, it will now correct the selection range to be within the comment delimiters.

It will activate when 
  - `correctSelection` is true (it is hardcoded `true` right now when toggling comments), 
  - The end comment delim is truthy,
  - The line is being commented

If these conditions pass, it will then get the selection range and see if the head or tail needs to be moved. The range is adjusted, with `autoscroll` disabled to be consistent with exisiting behaviour (and it would be _really_ annoying otherwise).

  - Note: because the delim is added by modifying the entire line, there is no need to move the cursor if it is not at an end. The bug only happens because the cursor being at an end counts as outside the modification range, and so is pushed along when the change is applied.

### Alternate Designs

I originally had it only work for empty selections, but decided this way is better.

It does seem quite long compared to what it's supposed to do though. I don't know how much of a problem that is.

This is applied to every selection in the editor when line comments are toggled. It's possible people only want it when there is just one selection / cursor. An interesting bug(?) is that when there's an even number of cursors on a line, it gets toggled twice, resulting in no change (this bug isn't related to this PR though).

It's a core mechanic

### Benefits

Consistent with other text editors, and makes sense.

### Possible Drawbacks

There's an interesting `...(this.getBufferRowRange() || [])` statement in the original version. I'm not sure when `getBufferRowRange` will return a falsey value, but that would shift the arguments of the function, moving the `options` parameter to `start`. 

Originally, if it does evaluate to `[]`, no error is thrown but the top line would be comment toggled. That seems like a bug, but isn't related to this PR. I've changed the empty array to `[null, null]` though, to ensure the argument positions are not changed.

### Verification Process

I tested manually, with the CSS grammar. I'll try to write proper tests after some feedback for changes, etc.

For now, this is what I'm expecting (where `|` is selection head/tail)

```
## Single cursor in line
ab|c
->
/* ab|c */

## Selection in line
a|b|c
->
/* a|b|c */

## Single cursor at end
abc|
->
/* abc| */

## Single cursor at beginning
|abc
->
/* |abc */

## Selection on whole line
|abc|
->
/* |abc| */

## Multi-whole line selection
|a
b|
 ->
/* |a
b| */

## Multi-partial line selection
a|bc
abc
abc|
->
/* a|bc
abc
abc| */

a|bc
abc
ab|c
->
/* a|bc
abc
ab|c */
```

### Applicable Issues

Fix #4784

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -770,7 +770,10 @@ class Selection {
   //   * `bypassReadOnly` (optional) {Boolean} Must be `true` to modify text within a read-only editor. (default: false)
   toggleLineComments (options = {}) {
     if (!this.ensureWritable('toggleLineComments', options)) return
-    this.editor.toggleLineCommentsForBufferRows(...(this.getBufferRowRange() || []))
+    let bufferRowRange = this.getBufferRowRange() || []
+    let cursor = this.cursor
+    let setCursor = this.isEmpty() && this.cursor.isAtEndOfLine()
+    this.editor.toggleLineCommentsForBufferRows(...bufferRowRange, { setCursor, cursor })
   }
 
   // Public: Cuts the selection until the end of the screen line.
--- a/src/selection.js
+++ b/src/selection.js
@@ -771,9 +771,7 @@ class Selection {
   toggleLineComments (options = {}) {
     if (!this.ensureWritable('toggleLineComments', options)) return
     let bufferRowRange = this.getBufferRowRange() || [null, null]
-    let cursor = this.cursor
-    let setCursor = this.isEmpty() && this.cursor.isAtEndOfLine()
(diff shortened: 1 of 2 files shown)


Aerijo:
Moved to https://github.com/atom/atom/pull/18471

PR #17521, opened 2018-06-16
Support activation hooks by grammar scope

### LINKED ISSUE #14148: Support for grammar scopeName based activation hooks
At the moment Atom supports `activationHooks` like `${packageName}:grammar-used`, but these only work on package level and not on the level of the actual grammar. This has two downsides:
1. If a package defines multiple grammars, the `activationHooks` will be triggered for any grammar. An example for this is the `language-java` package.
2. if multiple packages define the same grammar, a package that should be activated if this grammar is used has to add `activationHooks` for every package and update the `activationHooks` if a new package implements that grammar. An example for this is the `language-javascript` and `language-babel` package.

I think it wound be best to have another activation hook that works on the `scopeName` of a grammar. Something like: `${scopeName}:grammar-scope-used`.

Aerijo:
### Description of the Change

Adds activation hook for grammar scope name.

### Alternate Designs

None

Is core mechanic

### Benefits

Makes activation package agnostic, much like the services API.

### Possible Drawbacks

None

### Verification Process

Tested manually. ~Will add tests.~

Test added.

### Applicable Issues

Fixes #14148

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1268,6 +1268,7 @@ module.exports = class Workspace extends Model {
 
   handleGrammarUsed (grammar) {
     if (grammar == null) { return }
+    this.packageManager.triggerActivationHook(`${grammar.scopeName}:root-scope-used`)
     return this.packageManager.triggerActivationHook(`${grammar.packageName}:grammar-used`)
   }
 
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1269,7 +1269,7 @@ module.exports = class Workspace extends Model {
   handleGrammarUsed (grammar) {
     if (grammar == null) { return }
     this.packageManager.triggerActivationHook(`${grammar.scopeName}:root-scope-used`)
-    return this.packageManager.triggerActivationHook(`${grammar.packageName}:grammar-used`)
+    this.packageManager.triggerActivationHook(`${grammar.packageName}:grammar-used`)
   }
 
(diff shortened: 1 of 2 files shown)


src/workspace.js
```diff
@@ -1268,6 +1268,7 @@ module.exports = class Workspace extends Model {
 
   handleGrammarUsed (grammar) {
     if (grammar == null) { return }
+    this.packageManager.triggerActivationHook(`${grammar.scopeName}:root-scope-used`)
     return this.packageManager.triggerActivationHook(`${grammar.packageName}:grammar-used`)
```
thomasjo: Outside the scope of this PR, but remarking on it anyway; should we :fire the `return` here? There is no value returned by `triggerActivationHook`, so returning here doesn't make any sense and is simply confusing.

Aerijo: I'd remove it. Should be fine to do it in this PR right?

thomasjo: Yeah, definitely fine in my opinion. Just didn't want to expand the scope of your PR :sweat_smile:

Aerijo:
The test is just a copy of the one above, but the hook changed to `source.js:root-scope-used` etc.

I think it passes, but running it on my computer seemed to always cause a timeout on the promises (for this, the one above, and several others that changed on each run).

thomasjo (approved):
LGTM :ship:

daviwil (approved):
Thanks a lot @Aerijo, looks great!

PR #17538, opened 2018-06-19
Implement RFC 002: Atom Nightly Releases

daviwil:
### Description of the Change

This change adds automation for producing nightly Atom releases using VSTS CI.  Most of the changes are just slight modifications to Atom's existing build scripts to produce another build channel and publish those artifacts in a way that can be installed and updated when new
releases are available.

See **[Atom RFC 002](https://github.com/atom/atom/blob/master/docs/rfcs/002-atom-nightly-releases.md)** for more details.

### Alternate Designs

See [this section](https://github.com/atom/atom/blob/master/docs/rfcs/002-atom-nightly-releases.md#rationale-and-alternatives) of RFC 002.

This is a new release channel for Atom Core!

### Benefits

See [this section](https://github.com/atom/atom/blob/master/docs/rfcs/002-atom-nightly-releases.md#motivation) of RFC 002.

### Possible Drawbacks

There isn't a major downside to this effort since it would run in parallel to the existing Atom release process without affecting it.

### Verification Process

Produced signed Nightly builds using CI and verified that they are updatable using Electron's update service on Windows and macOS (the only platforms that support Squirrel).

At this moment nightly builds are not auto-updating because I decided to switch back to using Atom's update service on atom.io.  Updating will work again once I push a change to atom.io today or tomorrow to enable the Nightly channel.

### Applicable Issues

None.

### Remaining Tasks

There are a few remaining tasks that need to be finished before this PR can be merged:

- [x] Upload release assets to S3 bucket
- [x] Update title bar (and other relevant strings) to read "Atom Nightly"
- [x] Add Atom and apm launcher scripts (`atom-nightly.sh` and `apm-nightly.sh`) for Linux builds
- [x] Enlist @simurai's help to produce a special icon color for the Nightly release channel (purple shade?)
- [x] Update the metrics package to [report the Nightly channel](https://github.com/atom/metrics/blob/master/lib/reporter.js#L20) (https://github.com/atom/metrics/pull/92)
- [x] Find a way to set the version of the build in nightly-releases.yml so the major/minor parts aren't hardcoded
- [x] Update build environment variables to use ones defined by us and not VSTS
- [x] Update atom.io to support the new Nightly channel
- [x] Uncomment the assignment of `remoteReleases` in `script/lib/create-windows-installer.js` so that Windows delta updates can be generated
- [x] Fix or disable tests that fail on VSTS (more info below)
- [x] Document the VSTS setup in a short Markdown file for future reference
- [x] Update the About package to link to the right repo for Nightly release notes (atom/about#82)

#### Failing Tests

There are 4 tests that fail consistently on VSTS for no obvious reason:

**macOS**

keybinding-resolver package:

```
 KeyBindingResolverView
   when a keydown event occurs
     it displays all commands for the keydown event but does not clear for the keyup when there is no keyup binding
       TypeError: Cannot read property 'textContent' of null
         at it (/Users/vsts/agent/2.134.2/work/1/s/node_modules/keybinding-resolver/spec/keybinding-resolver-view-spec.js:91:80)
         at <anonymous>
```

GitHub package:
```
 GithubLoginModel default strategy manages passwords:
    AssertionError: assert.equal((await loginModel.getToken('test-account')), TOKEN): expected Symbol(UNAUTHENTICATED) to equal 'TOKEN'
       at Context.<anonymous> (/Users/vsts/agent/2.134.2/work/1/s/node_modules/github/test/models/github-login-model.test.js:22:18)
       at <anonymous>
```

GitHub package:
```
 GithubLoginModel SecurityBinaryStrategy manages passwords:
     Error: Command failed: security add-generic-password -s atom-github -a test-account -w TOKEN -U
 security: SecKeychainItemCreateFromContent (<default>): The specified item already exists in the keychain.
       at ChildProcess.exithandler (child_process.js:287:12)
       at emitTwo (events.js:126:13)
       at ChildProcess.emit (events.js:214:7)
       at maybeClose (internal/child_process.js:925:16)
       at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
```

I'd certainly appreciate any ideas for how these issues might be diagnosed!

### Future Tasks

There are other tasks that will be finished in future PRs:

- Enable x86 Windows releases (VSTS currently doesn't support x86 builds) (#17668)
- Automatically generate release notes for nightly releases (#17669)
- Switch back to using Electron's update.electronjs.org service once it gains features that Atom needs (#17670)
- Fix an issue in our Windows releases where different release channels can't be used side-by-side (#9247)

I'll file issues for these when this PR is approved and merged.


script/config.js
```diff
@@ -41,7 +44,9 @@ module.exports = {
 }
 
 function getChannel () {
-  if (appMetadata.version.match(/dev/)) {
+  if (process.env.BUILD_DEFINITIONNAME === 'Atom Nightly') {
```
daviwil: This is a VSTS environment variable but I'm planning to change it to something specific to Atom

script/config.js
```diff
@@ -50,6 +55,17 @@ function getChannel () {
   }
 }
 
+function computeAppVersion (version, channel) {
+  if (channel === 'dev') {
+    const result = spawnSync('git', ['rev-parse', '--short', 'HEAD'], {cwd: repositoryRootPath})
+    const commitHash = result.stdout.toString().trim()
+    version += '-' + commitHash
+  } else if (channel === 'nightly') {
+    version = process.env.BUILD_BUILDNUMBER
```
daviwil: I'll also change this env variable to something that isn't specific to VSTS

script/lib/create-windows-installer.js
```diff
@@ -1,32 +1,45 @@
...
-const fs = require('fs-extra')
+const fs = require('fs')
 const glob = require('glob')
 const path = require('path')
 
 const CONFIG = require('../config')
 
 module.exports = (packagedAppPath) => {
-  const archSuffix = process.arch === 'ia32' ? '' : '-' + process.arch
+  // const archSuffix = process.arch === 'ia32' ? '' : '-' + process.arch
```
daviwil: This and the other commented line below will be uncommented once atom.io changes are merged/deployed.

script/lib/create-github-release.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
```
daviwil: This file wasn't supposed to make it in, will take it back out

script/vsts/nightly-release.yml
```diff
@@ -0,0 +1,39 @@
+name: 1.29.0-nightly$(Rev:r)
```
daviwil: Need to find a way to pull this version number from package.json on master so that it doesn't need to be updated manually each time we roll the railcars

script/config.js
jasonrudolph: It looks like this uses the build number as Atom's version number. If that's the current behavior, how would you feel about changing it to append the build number to the version number (e.g., "1.30.0-nightly-42")?

script/lib/create-windows-installer.js
```diff
@@ -1,32 +1,45 @@
...
+    }
+
+    for (let nupkgPath of glob.sync(`${CONFIG.buildOutputPath}/atom-*.nupkg`)) {
       if (!nupkgPath.includes(CONFIG.appMetadata.version)) {
         console.log(`Deleting downloaded nupkg for previous version at ${nupkgPath} to prevent it from being stored as an artifact`)
-        fs.removeSync(nupkgPath)
+        fs.unlinkSync(nupkgPath)
+      } else {
+        if (process.arch === 'x64') {
+          const newNupkgPath = `${CONFIG.buildOutputPath}/atom-x64${path.basename(nupkgPath).slice(4)}`
```
jasonrudolph: What's the significance of the number `4` on this line? Could that number be extracted to a constant whose name explains the significant of the number?

script/publish-release
```diff
@@ -0,0 +1,40 @@
...
+  .wrap(yargs.terminalWidth())
+  .argv
+
+let assetsPath = argv.assetsPath || path.join(CONFIG.repositoryRootPath, 'out')
+let assets = glob.sync(path.join(assetsPath, '*(*.exe|*.zip|*.nupkg|*.tar.gz|*.rpm|*.deb|RELEASES*)'))
+
+publishRelease({
+  token: process.env.GITHUB_TOKEN,
+  owner: 'atom',
+  repo: CONFIG.channel !== 'nightly' ? 'atom' : 'atom-nightly-releases',
```
jasonrudolph: Can you remind me why it makes sense to have a separate repository for the nightly releases?

script/vsts/linux.yml
```diff
@@ -0,0 +1,52 @@
+phases:
+- phase: Linux
+  queue:
+    name: Hosted Linux Preview
+    timeoutInMinutes: 180
+
+  steps:
+  - task: NodeTool@0
+    inputs:
+      versionSpec: 8.11.3
```
jasonrudolph: I think the current builds all use Node 6.9.4 [[example](https://github.com/atom/atom/blob/26cd4a07d1a33c47fb71709a674d749b9e84e828/.travis.yml#L18)]. Can you share the motivation for using Node 8.11.3 here?

I'm thinking that we'd want all builds (including the new nightly builds) using the same Node version.

jasonrudolph (commented):
@daviwil: Thanks for getting this going. :zap:

I've included a few questions inline in the review below.

In addition to those inline questions, I have a few higher-level questions/notes:

1. What happens if I'm using the nightly release channel, and Atom's master branch hasn't had any new commits in more than 24 hours? Does a new nightly release still get created?
2. What happens if the nightly build fails for one platform and passes for the other platforms? Does the nightly release get created for the platforms where the build succeeded?
3. The task list says:

    > - [ ] Fix or disable tests that fail on VSTS (more info below)

    I have a fairly strong preference that we fix those tests instead of disabling them. :pray:

simurai:
> Enlist @simurai's help to produce a special icon color for the Nightly release channel (purple shade?)

How about:

atom-nightly

Are the stars too cheesy? :smile: It's a hint for "nightly". It would also differ from the rest that only have different hues, maybe ok.

atom-icons

I was also thinking to replace all icons and redesign them with the "portal style" from atom.io. But that could also be done in a later PR.

thomasjo:
@simurai I absolutely love that! Sure, it might be a little bit "cheesy", but :cheese: is delicious :heart:

maxbrunsfeld:
OMG @simurai that is awesome.

simurai:
>  I think doing an icon overhaul sometime this year in the new style would also be awesome, but this icon works for now!

Ok, let's keep it in mind, but no rush. For now I added the "purple with stars" version.

daviwil (commented):
@jasonrudolph Thanks!  To answer your questions:

1. Nightly builds only get kicked off if there have been changes to `master` within the last 24 hours.

2. If any part of the build process fails on any platform, nothing gets published.  The "release" phase of the build only proceeds if all 3 platform builds and their tests pass.

3. I've had to disable tests in the github and keybinding-resolver packages (only on VSTS) since they will take more time to investigate and fix.  I've filed issues atom/github#1568 and atom/keybinding-resolver#60, respectively, to track the need to fix these.  It turns out that they're both issues that appear on other CI platforms but VSTS happens to reproduce them more consistently.

daviwil:
This PR is ready for final review!  If you'd like some overall context on the process, check out the documentation I [wrote up here](https://github.com/atom/atom/blob/dw-nightly-releases/script/vsts/README.md).  I'm happy to continue making improvements to the code and documentation until we feel it's ready to be merged.

Thanks!

maxbrunsfeld (approved):
This looks great!

script/vsts/linux.yml
daviwil: Mainly because we now use Node 8 inside of Electron 2.0, so it's helpful to have the same JavaScript language features (async/await, improved perf) inside our build scripts.  I started off using Node 6 in the builds and saw a noticable performance increase (~10 minutes saved) by switching to 8.  More recently in this PR, I switched to the exact version of node in Electron 2.0, 8.9.3.

script/publish-release
daviwil: Two reasons:

1. atom/atom's Releases feed will get swamped with nightly releases and it'll be harder to scan through the list to find older stable/beta releases
2. My goal was to use Electron's new update service for this and they use the Releases feed as the source of truth for updates.  Seemed to make more sense to have it be a separate repo for now.  Can always change it later!

script/config.js
daviwil: This was actually the case but it wasn't super clear since I was using VSTS' environment variable :)  At this point in the code, I had configured the build to use the full `1.30.0-nightly42` version as the "build number".  I've since changed it to use a more Atom-specific environment variable that gets set in the VSTS YAML config.

daviwil (commented):
Missed some other comments from Jason, sorry about that!

script/vsts/linux.yml
jasonrudolph: @daviwil Thanks for the insight. Can I talk you into opening a follow-up PR to replace the existing uses of Node 6 with Node 8?

https://github.com/atom/atom/blob/3109958dc1804da4e22830d2d8750a2ea9fe5f3a/.travis.yml#L18
https://github.com/atom/atom/blob/3109958dc1804da4e22830d2d8750a2ea9fe5f3a/appveyor.yml#L23
https://github.com/atom/atom/blob/3109958dc1804da4e22830d2d8750a2ea9fe5f3a/circle.yml#L20-L21

daviwil: Definitely, I'll send that out later today!

daviwil:
Tests are green, issues are filed for future work.  I think this is ready to be merged!  Once it's in `master` I'll set up the scheduled build so that we start pumping out these releases on the regular :)

jasonrudolph:
🚚📦:atom:

maxbrunsfeld:
That was a ton of work. Nice job powering through 💥 ⚡️

daviwil:
image

PR #17551, opened 2018-06-22
Add Tree-sitter grammar injections

### LINKED ISSUE #17392: Tree transforms for tree sitter
## Summary

We'd like to be able to nest languages with tree sitter. That is, we want to somehow be able to specify that certain AST nodes (all terminals, currently, and probably in most cases) should have their content parsed with another grammar, generating a nested tree.

In the general case, we'd like to transform an antecedent (source) tree into a consequent (result) tree, and keep them both up to date with the buffer.

Basically, add a map method to a tree sitter tree like `Tree<T>::map<X>(transform: T -> X) -> Tree<X>`, where the returned tree handles traversal of nested trees and incremental parsing transparently in the case where `X` (somehow) contains a child tree. (This isn't intended as a final implementation, but as a general way of articulating the problem). We can compose multiple transforms together.

## Motivation

1. We'd like to offer support for styling escape sequences within string literals and regex operators and productions within regex literals. The TextMate grammars are able to do this, so it's technically a regression.

2. We'd like to be able to style certain variable names as keywords—for example, styling `module` or `exports` differently from just variables, even though they are really just variables.

3. In general, we'd like to be able to support nested languages—Javascript inside an HTML `<script>` tag, for example.

4. There are more exotic use cases for nested languages that would have great wow factor. Say, GraphQL / SQL expressions within appropriately tagged template literals in JS code. This is a little harder, because we either have to figure out what nested language to use (ML language detector anyone?) or wire in some config for users to do so.

For use cases like (1), we could probably shim in an additional parse pass as we walk the tree. We could look for escape sequences before emitting styling tags specifically within string literal nodes, say. This doesn't scale that well, and opens up other questions, like whose regex grammar do we use, if we're highlighting regexes?

We could also just throw the necessary productions into the host language. This works well for escape sequences, regexes, Linq expressions, and other node types that really are conceptually part of the language. It's really not ideal for situations like nesting JS inside HTML; we'd prefer if the HTML parser didn't have to import the entire JS parser.

In the case of nesting languages where there's any possibility of dynamism in how the nesting occurs, simply having the parent language eat the child seems to be a non-starter—we'd have to re-compile the JS parser to include support for GraphQL tagged expressions *or* we'd have to include every child language we want to support in the host language, creating significant overhead (in terms of parser size if not parse time).

# Additional context & open questions

Ideally, we wouldn't have to change tree sitter proper, and I don't think we do. In all the cases above, we want to map the content of terminal nodes onto a new tree. @maxbrunsfeld and I discussed building a facade over the current tree cursor abstraction which understand when it needs to descend into a child tree and can do so, creating a nested cursor and iterating over it to completion, and then continuing on.

We'll also need some way to specify how languages are chained. That is, we'll want to be able to write a function that detects `<script>` tags or `graphql`-tagged template literals, and then knows how / is somehow handed a language dictionary that lets it parse the content of those nodes with a parser from another language.

maxbrunsfeld:
## Motivation

The TextMate parsing system allows grammars to be composed in order to syntax-highlight things like:
* JavaScript inside of script tags in an HTML file
* HTML inside of template strings in JavaScript
* SQL inside of a string literal in Python

Currently, these types of things don't work when Tree-sitter is enabled.

## Solution

This PR adds the concept of 'grammar injection' to Tree-sitter grammars. Specifically, it adds two new APIs associated with Tree-sitter grammars. These APIs might be revised before this PR is merged.

#### 1. Adding Injection Points

```js
atom.grammars.addInjectionPoint(
  languageId,
  {
    type: syntaxNodeType,
    language: languageCallback,
    content: contentCallback
  }
)
```

This method allows you to express ideas like: "In JavaScript, tagged template strings are injection points. For each tagged template string, try to identify its language by looking at the name of its tag function. Parse the content between the backticks, omitting any template substitutions."

```js
atom.grammars.addInjectionPoint(
 'javascript',
  {
    // tagged template literals are simply parsed as call expressions 
    // with a template string instead of an argument list
    type: 'call_expression', 
  
    // The language name can be found in the template string's "tag"
    language (callNode) {
      if (callNode.lastChild.type === 'template_string') {
        return callNode.firstChild.text
      }
    },

    // Parse the content inside of the template string
    content (callNode) {
      return callNode.lastChild
    }
  }
)

```
</details>

Note that this API does *not* indicate which grammar to use when parsing the content. That information can be provided by other packages, using the second API: 

#### 2. Specifying Injection Patterns

Grammars that use Tree-sitter have a new field called `injectionRegExp`. This field allows you to express ideas like: "The HTML language can be injected. Whenever there is an injection point where the language-name includes the string 'html', parse that injection point's content using the HTML parser."

```coffee
id: 'html'

injectionRegExp: 'html|HTML|Html$'
```

This two-part API allows languages to be embedded within one another without every grammar having to know about every other grammar.

## Example

[EJS](http://ejs.co) is a popular JavaScript templating system where JavaScript code is interspersed with HTML markup using the delimiters `<%` and `%>`. The HTML can of course contain more JavaScript code inside of `script` tags. And that JavaScript code can contain HTML inside of string literals.

html-in-js-in-html-in-ejs

## Tasks

* [x] Add tests and documentation for the new `GrammarRegistry` APIs
* [x] When the buffer changes, only update the affected injections, not all of them
* [x] Update highlighting when grammars with `injectionRegExps` are added
* [x] Allow injections that are spread across many child nodes (needed for PHP, ERB, EJS, mustache, etc)
* [x] Use injections for the expand selection command
* [x] Use injections for folding specific lines
* [x] Fix syntax highlighting bugs
* [x] Use injections for getting scope descriptors
* [ ] Use injections for folding at a given nesting level

## Related Issues / PRs

Closes https://github.com/atom/atom/issues/17392
Depends on https://github.com/tree-sitter/tree-sitter/pull/181
Depends on https://github.com/tree-sitter/node-tree-sitter/pull/14
Depends on https://github.com/tree-sitter/node-tree-sitter/pull/18


thomasjo (commented):
~Stupid style comment; any specific reason for prefixing some of these functions with an underscore? I'm assuming they are "private", but I don't think we do this anywhere else in the core code?~ This wasn't meant to be a review, just a line comment...

src/tree-sitter-language-mode.js
```diff
@@ -384,11 +364,273 @@ class TreeSitterLanguageMode {
...
+        this.tree.edit(this._treeEditForBufferChange(
+          oldStart, oldEnd, newEnd, oldText, newText
+        ))
+      }
+      this.patchSinceCurrentParseStarted = null
+      this.update(containingNode)
+    }
+  }
+
+  async _performUpdate (containingNode) {
```
thomasjo: Stupid style comment; any specific reason for prefixing some of these functions with an underscore? I'm assuming they are "private", but I don't think we do this anywhere else in the core code?

50Wliu:
Looks like package-lock files were committed by accident?

fcomrqz:
Hi @maxbrunsfeld, 

I'm trying to use the `tree-sitter-vue`.

Syntax highlight is working well. 🎉 

Folding at given level and autocomplete only works for HTML.

In this pull request you leave a task incomplete:  
- [ ] Use injections for folding at a given nesting level.

Can you give me a hint about how can I fix this?
I don't know if is something related to the Vue parser, the Atom grammar, the `autocomplete-plus` or anything else.

In between I'm learning how to create a parser from .

Thanks. ✌️

Issue #17515, opened 2018-06-14
open file in line:column from command line


fedelibre:
## Summary

Many editors allow to open a file in a precise position from the command line. The usual syntax is something like `editor +line:column file`. Here are a few examples (gedit and Visual Studio Code):

    gedit +line:column file
    code -g file:line:column

## Motivation

This is useful especially when you are debugging/grepping the code in a terminal and you want to jump to the exact location of a file you want to fix.

I have also a very specific use case. I write sheet music using an open source software called [lilypond](http://lilypond.org). lilypond converts music written as plain text to PDF. It can create PDF files containing textedit URIs to the precise point in the source file which generated the object in the PDF file that you clicked. This feature is called, in lilypond terms, point-and-click.

lilypond supports already a few editors (emacs, gvim, gedit and jedit), but I'd like to add more editors, including Atom. Without this feature I cannot include it.

## Describe alternatives you've considered

I've tried the [open package](https://atom.io/packages/open), but could not make it work. It opens the file but not at the desired line and column.

Also,  I believe this is a feature that should be part of Atom core and not be available only as plugin.

lee-dohm:
Thanks  for reaching out.

Atom already supports this with the second syntax. For example, in this repo you could do `atom README.md:5:13` and end up with the cursor just before the "A" in "Atom" on this line:

```
[![Join the Atom Community on Slack](https://atom-slack.herokuapp.com/badge.svg)](https://atom-slack.herokuapp.com)
```

-----

Additionally, we require the [issue template](https://github.com/atom/atom/blob/master/ISSUE_TEMPLATE.md) to be filled out on all new issues and pull requests. We do this so that we can be certain we have all the information we need to address your submission efficiently. This allows the maintainers to spend more time fixing bugs, implementing enhancements, and reviewing and merging pull requests. Please open a new issue and fill out the template for us.

Thanks for understanding and meeting us half way :grinning:

lee-dohm:
I apologize about the issue template remark. I thought the new feature request template still included a requirement to test out things in Safe Mode. I was a little quick on the trigger there. My apologies :pray:

fedelibre:
@lee-dohm This is great news! I've tested it and it works great alone and in combination with lilypond point-and-click.

However, this feature and its syntax is hidden, as it's not documented in `atom --help`. Can I change the title of this issue to reflect this request? Or better open a new issue?

Ben3eeE:
[...]

PR #17569, opened 2018-06-23
Doc: explain how to open a file at a certain line and column

fedelibre:
This is a follow-up to the discussion in issue #17515.

The first place where I'd have expected to find this information is in the `atom --help` output, so it's where I've put it.
I can't find any other place, as Atom doesn't have a man page and The Flight Manual seems dedicated entirely to the GUI of Atom.


lee-dohm:
Thanks for the help! 👍 

It could also be added to the paragraph about opening files on the command line [roughly here](https://flight-manual.atom.io/getting-started/sections/atom-basics/#opening-a-file) in the Flight Manual, if you're interested 💖

Issue #17565, opened 2018-06-23
Unable to start Atom: "dyld: Symbol not found: _CIDetectorTypeText"


textchimp:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Atom 1.28.0 fails to start on OSX 10.10.1

### Steps to Reproduce

1. Start Atom as usual from the Applications folder, Spotlight, or the command line

**Expected behavior:** [What you expect to happen]

App to launch as usual

**Actual behavior:** [What actually happens]

- When started from Finder/Spotlight, a popup window appears which states `Atom cannot be opened because of a problem`.

- When started from the command line with `/Applications/Atom.app/Contents/MacOS/Atom`, the error text is:
```
dyld: Symbol not found: _CIDetectorTypeText
  Referenced from: /Applications/Atom.app/Contents/MacOS/../Frameworks/Electron Framework.framework/Electron Framework
  Expected in: /System/Library/Frameworks/QuartzCore.framework/Versions/A/QuartzCore
 in /Applications/Atom.app/Contents/MacOS/../Frameworks/Electron Framework.framework/Electron Framework
Trace/BPT trap: 5
```

- When started with `atom --safe`, the error text is 
```
LSOpenURLsWithRole() failed for the application /Applications/Atom.app with error -10810.
```

**Reproduces how often:** [What percentage of the time does it reproduce?]

100% of the time; unable to start Atom normally

### Versions

The Atom version commands are not working. They cause the "Atom cannot be opened because of a problem" popup to appear.

I downloaded the latest version of Atom (1.28.0) and replaced my previous installation with it, but the result was the same.

OSX Yosemite 10.10.1

### Additional Information

- I have tried re-downloading Atom and reinstalling it, which made no difference to the error
- I have tried removing `~/.atom`, which made no difference
- I have tried restarting my system several times

[...]

PR #17620, opened 2018-07-03
⬆️ electron@2.0.4

jasonrudolph:
Release notes: https://electronjs.org/releases#2.0.4

---

We anticipate that the upgrade from Electron 2.0.3 to 2.0.4 will resolve the following issues:

- https://github.com/atom/atom/issues/17557 - We suspect this issue was caused by https://github.com/electron/electron/issues/12676.
- https://github.com/atom/atom/issues/17565 - We suspect this issue was caused by https://github.com/electron/electron/issues/13404.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "2.0.3",
+  "electronVersion": "2.0.4",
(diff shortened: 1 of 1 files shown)


daviwil:
:shipit:

daviwil (approved):

keteipor:
veamos la utilidad

PR #17628, opened 2018-07-05
Fix getEnvFromShell() to correctly handle newlines in env vars

### LINKED ISSUE #17627: getEnvFromShell() does not handle newlines in environment variables correctly
### Prerequisites

### Description

If the user has an environment variable that contains newlines when atom is launched (other than from the command-line), then that environment variable will not be correctly passed to subprocesses. Environment variables containing newlines commonly occur when bash functions are exported.

### Steps to Reproduce

1. Add this to the end of ~/.bashrc: ```export atom_test=$'foo\nbar'```
2. Launch atom from the desktop (not from a shell)
3. Inspect the value of ```process.env['atom_test']```

**Expected behavior:**
The value should be:
```
foo
bar
```

The value is:
```
foo
```

**Reproduces how often:**
100% of the time

### Versions
```
[brettle@localhost tmp]$ atom --version
Atom    : 1.28.0
Electron: 2.0.3
Chrome  : 61.0.3163.100
Node    : 8.9.3
[brettle@localhost tmp]$ apm --version
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.28.0
python 2.7.15
git 2.17.1
[brettle@localhost tmp]$ uname -a
Linux localhost.localdomain 4.17.3-200.fc28.x86_64 #1 SMP Tue Jun 26 14:17:07 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
[brettle@localhost tmp]$ rpm -q fedora-release
fedora-release-28-2.noarch
```
### Additional Information

The essence of the problem is at [this line](https://github.com/atom/atom/blob/5b485dcb4b48386d9480ab45dc389cbcaad3d454/src/update-process-env.js#L110) in ```update-process-env.js```. It is using newlines to split the result of executing
```
command env
```
within the user's shell at [this line](https://github.com/atom/atom/blob/5b485dcb4b48386d9480ab45dc389cbcaad3d454/src/update-process-env.js#L77)

Instead, I recommend using \0 to split the result of executing
```
command awk 'BEGIN{for(v in ENVIRON) printf("%s=%s\0", v, ENVIRON[v])}'
```
within the user's shell. This would address newlines in the variable values and the extremely rare case of newlines in the variable names.

brettle:
### Description of the Change

Fix #17627 by using awk to portably separate env vars with \0 instead of \n.

getEnvFromShell() was supposed to copy the user's environment as it exists in a new shell. However, it did that by running `env` in that shell and using \n to split the output into var=value pairs. However, values can contain \n (e.g. exported bash functions) and such values will break the parsing. The fix is to have the var=value pairs separated by \0s. GNU's ```env -0``` does that but won't work on Mac or BSD. Instead, we get the same result using a one line POSIX-compliant  ```awk``` script.

### Alternate Designs

I was going to just use ```env -0``` until I realized that it was GNU-specific.

It fixes a bug in core.

### Benefits

It fixes and prevents issues where a subprocess is started with an incorrect environment due to multiline environment variables.

### Possible Drawbacks

If ```awk``` corresponds to GNU's ```awk```, it will add AWKLIBPATH and AWKPATH to the environment (with default values) if they don't already exist.

### Verification Process

I modified the existing spec to include multiline values and confirmed it worked.
I also confirmed that multiline values set in .bashrc are actually showing up correctly in process.env after the change was made.

### Applicable Issues

Issue #17627


50Wliu:
Looks like you have linting errors @brettle.

thomasjo:
@brettle Looks like the tests are failing in a consistent manner, and the failures definitely seem to be related to the changes in this PR. Let us know if you need any assistance with resolving the problems.

brettle:
@thomasjo Thanks for the offer, but it looks like I've got it fixed.

sadick254:
Hey @brettle. Apologies for not looking into this on time. I know this PR has been here for a very long time, during that time Atom has seen a number of changes. Due to the changes, this PR has now become outdated. Is it possible for you to resolve the conflicts?

sadick254 (approved):
Looks good 👍

PR #17636, opened 2018-07-06
Reset goal column on all cursor changes

matthewwithanm:
Previously, pressing the home key (move-to-first-character-of-line)
while on an empty line wouldn't clear the goal column. This is because
it was only cleared on cursor *change* and that didn't result in a
change. With this commit, it's *always* cleared. Operations that want
to preserve the goal column can reset it afterwards.

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -326,7 +326,9 @@ class Cursor extends Model {
 
   // Public: Moves the cursor to the bottom of the buffer.
   moveToBottom () {
+    const column = this.goalColumn
     this.setBufferPosition(this.editor.getEofBufferPosition())
+    this.goalColumn = column
   }
 
   // Public: Moves the cursor to the beginning of the line.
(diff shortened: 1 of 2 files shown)


src/cursor.js
```diff
@@ -326,7 +326,9 @@ class Cursor extends Model {
 
   // Public: Moves the cursor to the bottom of the buffer.
   moveToBottom () {
+    const column = this.goalColumn
     this.setBufferPosition(this.editor.getEofBufferPosition())
+    this.goalColumn = column
```
maxbrunsfeld: Why is the `goalColumn` only preserved this way in this one method?

maxbrunsfeld (commented):
Thanks for jumping on this!

This change seems like it'll need some manual testing to make sure that everything related to goal columns still works as expected. Does everything feel right after coding for a while with this change?

src/cursor.js
matthewwithanm: honestly, there may be. this is the only one that caused a test to fail though.

matthewwithanm:
yeah, it feels right. granted i haven't coded much with it, but just doing the normal things (selecting lines, home, end, etc.) all seem good.

maxbrunsfeld:
Awesome!

PR #17651, opened 2018-07-09
Tooltip should fade on keybaord event 17431

### LINKED ISSUE #17431: Tooltips should fade away on keyboard event
### Prerequisites

    * [X] Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * [X] Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * [X] Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * [X] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * [X] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Tooltips like for example file path when hovering over tab name at the top or tooltip "Click to fetch" when over "Fetch" at the bottom should GO AWAY the moment keyboard event happens (typing anything or simple arrow keys clicks, etc) even if mouse cursor is still over the "tooltip triggering area".

Otherwise, tooltip becomes annoyance instead of helping hand, obstructing my workspace.

### Steps to Reproduce

---
*Edit by @rsese to add alternate repro steps*

Using atom/atom as an example:

1. `atom --safe .`
2. Open `SUPPORT.md`
3. Hover over the tab for the file so the path tooltip appears
4. Start typing

tooltip-hiding

---

1. Type in workspace document.
2. Hover mouse over "tooltip triggering area", wait for tooltip to appear.
3. Continue typing while not moving away the mouse.

**Expected behavior:** Tooltip(s) should be gone either on mouse hover away OR when any input event happens like keyboard click, mouse click (?), etc.

**Actual behavior:** Tooltip keeps in the place until I move mouse away only.

### Versions

macOS High Sierra 10.13.4

```
$ atom --version
Atom    : 1.27.1
Electron: 1.7.15
Chrome  : 58.0.3029.110
Node    : 7.9.0

$ apm --version
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.27.1
python 2.7.10
git 2.15.1
```

### Additional Information

atom_tooltip_bug

cacheflow:
### Description of the Change

Fix #17431 by adding an event listener that listens for keydown events and hides a tooltip when a user types.

### Alternate Designs

1. Initially I had attached the event listener in the tooltip file instead of the tooltip manager, but I placed in the latter because that's where the hiding tooltips on resized lived as well. 

Because it fixes a problem in core and most default editors have this feature. 

### Benefits

It fixes an issue where users may want to see the workspace they're typing in, but a tooltip is hiding it. 

### Possible Drawbacks

It adds an additional event listener for each tooltip.

### Verification Process

I modified the existing tooltip-manager spec to include a new test for this feature, ran the test suite, and also manually tested it by reproducing the steps listed in #17431.

### Applicable Issues

#17431

### DIFF
--- a/spec/tooltip-manager-spec.js
+++ b/spec/tooltip-manager-spec.js
@@ -213,6 +213,18 @@ describe('TooltipManager', () => {
       })
     )
 
+    describe('when a user types', () =>
+      it('hides the tooltips', () => {
+        const disposable = manager.add(element, { title: 'Title' })
+        hover(element, function () {
+          expect(document.body.querySelector('.tooltip')).not.toBeNull()
+          window.dispatchEvent(new CustomEvent('keydown'))
+          expect(document.body.querySelector('.tooltip')).toBeNull()
+          disposable.dispose()
+        })
+      })
+    )
+
     describe('findTooltips', () => {
       it('adds and remove tooltips correctly', () => {
         expect(manager.findTooltips(element).length).toBe(0)
--- a/spec/tooltip-manager-spec.js
+++ b/spec/tooltip-manager-spec.js
@@ -212,7 +212,7 @@ describe('TooltipManager', () => {
         })
       })
     )
-
+    
     describe('when a user types', () =>
       it('hides the tooltips', () => {
         const disposable = manager.add(element, { title: 'Title' })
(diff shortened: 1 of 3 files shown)


winstliu (approved):
Makes sense to me!

thomasjo (approved):
LGTM

lee-dohm:
Thanks very much for the awesome contribution! :tada:

PR #17674, opened 2018-07-12
RFC 003: Consolidate Core Atom Packages

daviwil:
This RFC proposes that by consolidating most core Atom packages back into the `atom/atom` repo, we will see the following benefits:

- Less confusion for new contributors
- Simpler core package contribution experience
- Greatly reduced burden for maintainers

The RFC contains details about which packages could be consolidated and the process by which we can accomplish this.

([**Rendered**](https://github.com/atom/atom/blob/rfc-003-consolidate-core-packages/docs/rfcs/003-consolidate-core-packages.md))

### DIFF
--- a/docs/rfcs/003-consolidate-core-packages.md
+++ b/docs/rfcs/003-consolidate-core-packages.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-Proposed
+Accepted
 
 ## Summary
 
--- a/docs/rfcs/003-consolidate-core-packages.md
+++ b/docs/rfcs/003-consolidate-core-packages.md
@@ -241,11 +241,13 @@ One alternative approach would be to break this core Atom functionality out of p
 
   One concern here is that there exist projects which depend directly on these repositories for the TextMate syntax grammars they contain.  Moving the code into `atom/atom` would require that we notify the consumers of the grammars so that they can redirect their requests to the `atom/atom` repo.
 
+- Should we use `git subtree` to migrate the entire commit history of these packages over or just depend on the history from a package's original repository?
+
+  For now, we won't use `git subtree` due to the possibility that bringing over thousands of commits could cause unknown problems in the Atom repo.  We may try this for newly consolidated packages in the future if we decide that not having the package commit history is a sufficient impediment to problem investigations.
+
 - What are the criteria we might use to eventually decide to move larger packages like `tree-view`, `settings-view`, and `find-and-replace` back into `atom/atom`?
 
 - Will we be losing any useful data about these packages if we don't have standalone repositories anymore?
-
-- Should we use `git subtree` to migrate the entire commit history of these packages over or just depend on the history from a package's original repository?
 
 - Should we use this as an opportunity to remove any unnecessary packages from the core Atom distribution?
 
--- a/docs/rfcs/003-consolidate-core-packages.md
+++ b/docs/rfcs/003-consolidate-core-packages.md
@@ -72,7 +72,7 @@ The more core packages live in `atom/atom`, the less work Atom maintainers have 
 
 ## Explanation
 
-Many of Atom's core packages now live in the core `atom/atom` repository.  To the Atom user, this change will be imperceptible as these packages still show up in the list of Core Packages in the Settings View.  Users can still disable these packages if they want to replace their behavior with other packages.
+Many of Atom's core packages now live in the core `atom/atom` repository.  To the Atom user, this change will be imperceptible as these packages still show up in the list of Core Packages in the Settings View.  Users can still optionally disable these packages.
 
 For maintainers and contributors, there will be less juggling of repositories and no more publishing of updates to these packages with `apm`:
 
@@ -86,7 +86,7 @@ Atom maintainers no longer have to publish new versions to consolidated core pac
 
 One possible drawback of this approach is that there might be some initial confusion where core Atom packages live, especially if some are consolidated into `atom/atom` and others still live in their own repositories.  We will manage this confusion by doing the following:
 
-- Include folders for _all_ core packages in the `packages` folder of the Atom repo and add README.md files to folders of those packages that still live in separate repos.  This will allow us to direct users to the proper home for packages that are not yet consolidated.
+- Include a `README.md` file in the `packages` folder which lists core packages that are not consolidated in the Atom repo.  This will enable users to find the home repositories of those packages.
 
 - Archive the repositories for consolidated core packages, but only after migrating existing issues, merging or closing existing PRs, and updating the README.md to point to the new home of the package code.
 
@@ -188,6 +188,7 @@ The following packages will not be consolidated until the stated reasons can be 
 |---------|-------------|----------|---------------------|--------------|-------|
 | **[find-and-replace]** | 219 | 17 | 0 | 6/4/18 | Too many open PRs |
 | **[fuzzy-finder]** | 89 | 22 | 0 | 5/17/18 | Too many open PRs |
+| **[github]** |  |  |  |  | Independent project |
 | **[language-c]** | 53 | 15 | 0 | 7/10/18 | Too many open PRs |
 | **[language-go]** | 12 | 2 | **1** | 6/18/18 | Package maintainer, possibly inactive? |
 | **[language-java]** | 8 | 2 | **1** | 6/11/18 | Package maintainer |
@@ -204,16 +205,13 @@ The following packages will not be consolidated until the stated reasons can be 
 
 #### Packages to Never Consolidate
 
-These packages will not be consolidated for the following reasons:
+These packages will not be consolidated because they would inhibit contributions from our friends in the Nuclide team at Facebook:
 
-| Package | Open Issues | Open PRs | Outside Maintainers | Last Updated | Reason |
-|---------|-------------|----------|---------------------|--------------|-------|
-| **[autocomplete-atom-api]** |  |  |  |  | Blocks contribution from Facebook |
-| **[autocomplete-css]** |  |  |  |  | Same as above |
-| **[autocomplete-html]** |  |  |  |  | Same as above |
-| **[autocomplete-plus]** |  |  |  |  | Same as above |
-| **[autocomplete-snippets]** |  |  |  |  | Same as above |
-| **[github]** |  |  |  |  | Independent project |
+- **[autocomplete-atom-api]**
+- **[autocomplete-css]**
+- **[autocomplete-html]**
+- **[autocomplete-plus]**
+- **[autocomplete-snippets]**
 
 ### Consolidation Process
 
@@ -222,7 +220,8 @@ To consolidate a single core package repository back into `atom/atom`, the follo
 1. All open pull requests on the package's repository must either be closed or merged before consolidation can proceed
 1. The package repository's code in `master` will be copied over to Atom's `packages` folder in a subfolder bearing that package's name.
 1. Atom's `package.json` file will be updated to change the package's `packageDependencies` entry to reference its local path with the following syntax: `"tree-view": "file:./packages/tree-view"`
-1. A test build will be created to manually verify the package loads and works correctly at first glance and also that package specs pass
+1. A test build will be created locally to manually verify that the package loads and works correctly at first glance
+1. The package specs for the newly-consolidated package will be run against the local Atom build
 1. A PR will be sent to `atom/atom` to verify that CI passes with the introduction of the consolidated package
 1. Once CI is clean and the PR is approved, the PR will be merged
 1. The package's original repository will have all of its existing issues moved over to `atom/atom` using a bulk issue mover tool, assigning a label to those issues relative to the package name, like `packages/tree-view`
@@ -238,11 +237,17 @@ One alternative approach would be to break this core Atom functionality out of p
 
 ## Unresolved questions
 
-- What are the criteria we might use to eventually decide to move larger packages like `tree-view`, `settings-view`, and `find-and-replace` back into `atom/atom`?
-
 - Is there a good reason to not move the `language-*` packages into `atom/atom`?
 
+  One concern here is that there exist projects which depend directly on these repositories for the TextMate syntax grammars they contain.  Moving the code into `atom/atom` would require that we notify the consumers of the grammars so that they can redirect their requests to the `atom/atom` repo.
+
+- What are the criteria we might use to eventually decide to move larger packages like `tree-view`, `settings-view`, and `find-and-replace` back into `atom/atom`?
+
 - Will we be losing any useful data about these packages if we don't have standalone repositories anymore?
+
+- Should we use `git subtree` to migrate the entire commit history of these packages over or just depend on the history from a package's original repository?
+
+- Should we use this as an opportunity to remove any unnecessary packages from the core Atom distribution?
 
 [about]: https://github.com/atom/about
 [archive-view]: https://github.com/atom/archive-view
--- a/docs/rfcs/003-consolidate-core-packages.md
+++ b/docs/rfcs/003-consolidate-core-packages.md
@@ -1,0 +1,332 @@
+# Consolidate Core Atom Packages
+
+## Status
+
+Proposed
+
+## Summary
+
+Atom's official distribution is comprised of 91 core packages which provide its built-in functionality.  These packages currently live in their own independent repositories in the Atom organization, all with their own separate issues, PRs, releases, and CI configurations.  This RFC proposes that by consolidating most, if not all, of these core packages back into the `atom/atom` repo, we will see the following benefits:
+
+- Less confusion for new contributors
+- Simpler core package contribution experience
+- Greatly reduced burden for maintainers
+
+## Motivation
+
+Let's cover each of the bullet points mentioned above:
+
+### Less confusion for contributors
+
+Imagine that a new contributor wants to add a small new feature to the `tree-view` package.  The first place they are likely to look is the `atom/atom` repository.  Scanning through the folders will lead to a dead end, nothing that looks like `tree-view` code can be found.  They might take one of the following steps next:
+
+- By reading README.md, maybe they will decide to click the link to the Atom Flight Manual and maybe_ find the [Contributing to Official Atom Packages](https://flight-manual.atom.io/hacking-atom/sections/contributing-to-official-atom-packages/) page there.
+- They could read the CONTRIBUTING.md file which [has a section](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages) that explains where to find the repos for core packages and how to contribute, but we don't really have a clear pointer to that in our README.md
+- If they don't happen to find that page, they might use Google to search for "atom tree view" and find the atom/tree-view repo and _maybe_ read the CONTRIBUTING.md file which sends them to Atom's overall contribution documentation
+- They might go to the Atom Forum or Slack community to ask how to contribute to
+
+Having all of the core Atom packages represented in a top-level `packages` folder, even if they don't actually live in the repo, will go a long way to making the core package code more discoverable.
+
+### Simpler core package contribution experience
+
+Separating core Atom features out into separate repositories and delivered via `apm` is a great idea in theory because it validates the Atom package ecosystem and gives developers many examples of how to develop an Atom package.  It also gives Atom developers real-world experience working with Atom's APIs so that we ensure community package authors have the same hackability that the Atom developers enjoy.
+
+On the other hand, having these packages live in separate repositories and released "independently" introduces a great deal of overhead when adding new features.  Here is a comparison of the current package development workflow contrasted to what we could achieve with consolidated packages:
+
+#### Current Package Development Workflow
+
+For example, to add a single feature to the `tree-view` package, one must:
+
+1. Fork and clone the `tree-view` repository to their computer (making sure to pull the commit relevant to the version of Atom they are working with)
+1. Run `apm install` and `apm link` inside of the repo folder
+1. Make their desired changes to the code
+1. Open a PR to the `tree-view` repo and wait for CI to pass and a maintainer to review it
+1. Work with maintainers to get the PR approved and merged
+
+After this is finished, an Atom maintainer must take the following steps
+
+1. Clone the `tree-view` repo
+2. Run `apm publish` to publish a new release of the package
+3. Edit `package.json` in the Atom repo to reflect the new version of `tree-view`
+4. Commit and push the changes to the relevant branch where the change belongs (`master` or `1.nn-releases`)
+
+If `tree-view` was moved into the `atom/atom` repository
+
+#### Simplified Package Development
+
+If we were to move `tree-view` (or any other core Atom package) back into `atom/atom`, the development workflow would look more like this:
+
+1. Fork and clone `atom/atom` and switch to a release branch if necessary
+1. Build Atom and launch it in dev mode
+1. Make desired changes to the code in `packages/tree-view`
+1. Open a PR on `atom/atom` and wait for CI to pass and a maintainer to review it
+1. Work with maintainers to get the PR approved and merged
+
+At this point, the change is merged into Atom and ready for inclusion in the next release.
+
+### Greatly reduced burden for maintainers
+
+Since packages all have their own repositories, this means that we have to watch 91 different repos for issues and pull requests.  This also means that we have to redirect issues filed on `atom/atom` to the appropriate repository when a user doesn't know where it belongs.  Even more importantly, there's not an easy way to prioritize and track issues across the Atom organization without using GitHub projects.
+
+Also, as mentioned above, there's the added duty of doing the package "version dance" when we merge any new PRs to a package repository: publish the package update, update `package.json` in Atom.  It's very easy to forget to do this and not have community contributions included in the next Atom release!
+
+The more core packages live in `atom/atom`, the less work Atom maintainers have to do overall.
+
+## Explanation
+
+Many of Atom's core packages now live in the core `atom/atom` repository.  To the Atom user, this change will be imperceptible as these packages still show up in the list of Core Packages in the Settings View.  For maintainers and contributors, there will be less juggling of repositories and no more publishing of updates to these packages with `apm`.
+
+Contributors now clone and build `atom/atom` to work on improvements to core packages.  They will no longer have to use `apm link` in dev mode to test changes they make to packages in the repo's `packages` folder.
+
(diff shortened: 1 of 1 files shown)


daviwil:
Hoping to run an experiment for this with one or two small packages in the next week or so to gather some more technical details on what problems we'll have to solve.

jasonrudolph (commented):
Nice work on this draft, @daviwil. I found so many of my questions getting answered as I continued reading through the RFC. :zap:

I've included a handful of questions and comments inline in the diff. Two additional topics I'm curious about:

1. **Disabling a package**: Will it still be possible to disable a core package that has been moved into atom/atom?
2. **Package versions**: The biggest thing I'm unsure of is how package versioning will be affected:
    - Based on the RFC, it seems to me that packages in atom/atom will no longer have version numbers and will no longer get published to atom.io. Is that true?
    - If packages in atom/atom don't have version numbers anymore:
        - Do they still have a `package.json` file? If so, what happens to the `version` property in a package's `package.json` file?
        - Will the packages still be declared in the top-level `package.json` file in atom/atom? If so, how will we determine what version number to specify?
    - If the affected packages will no longer be published to atom.io, what should happen to the existing versions of those packages on atom.io? (For example, fast forward to a year from now: It would seem strange to me to have atom.io continue to show an outdated mid-2018 README for a package that got moved into atom/atom and then received substantial updates.)

I hope this review helps. I'm happy to discuss any of these ideas further if that would be useful. :v:

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,332 @@
...
+1. Work with maintainers to get the PR approved and merged
+
+After this is finished, an Atom maintainer must take the following steps
+
+1. Clone the `tree-view` repo
+2. Run `apm publish` to publish a new release of the package
+3. Edit `package.json` in the Atom repo to reflect the new version of `tree-view`
+4. Commit and push the changes to the relevant branch where the change belongs (`master` or `1.nn-releases`)
+
+If `tree-view` was moved into the `atom/atom` repository
```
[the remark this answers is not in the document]
daviwil: Argh, apparently I was in a rush to get the PR sent out and forgot to finish some thoughts :)  Will finish these up!

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,332 @@
...
+- Number of package-only maintainers on the repo (exclude any with package maintainers for now)
+
+Using this criteria, all 91 packages have been evaluated and categorized to determine whether they are good candidates for consolidation:
+
+#### Initial Consolidation Candidates
+
+| Package | Open Issues | Open PRs | Outside Maintainers | Last Updated |
+|---------|-------------|----------|---------------------| -------------|
+| **[about]** | 2 | 0 | 0 | 7/11/18 |
+| **[archive-view]** | 10 | 0 | 0 | 6/3/18 |
```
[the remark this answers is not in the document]
daviwil: Totally on board with not bringing that into the core codebase, though not sure how easy it'd be to remove it as a bundled package at this point.  Worth discussing at some point!

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,332 @@
...
+1. All open pull requests on the package's repository must either be closed or merged before consolidation can proceed
+1. The package repository's code in `master` will be copied over to a subfolder in Atom's `packages` folder with a subfolder bearing that package's name.
+1. A test CI build will be run to ensure that the package loads and works correctly at first glance
+1. The package's original repository will have all of its existing issues moved over to `atom/atom` using a bulk issue mover tool
+1. The package's original repository will have its README.md to point contributors to the code's new home in `atom/atom`
+1. The package's original repository will now be archived
+
+### Alternative Approaches
+
+We haven't yet identified another approach which allows us to achieve the goals set forth in this RFC without consolidating these packages into `atom/atom`.
```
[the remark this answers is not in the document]
daviwil: Yeah, definitely.  I'll add it as an alternative just as a straw man.

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,332 @@
...
+
+Many of Atom's core packages now live in the core `atom/atom` repository.  To the Atom user, this change will be imperceptible as these packages still show up in the list of Core Packages in the Settings View.  For maintainers and contributors, there will be less juggling of repositories and no more publishing of updates to these packages with `apm`.
+
+Contributors now clone and build `atom/atom` to work on improvements to core packages.  They will no longer have to use `apm link` in dev mode to test changes they make to packages in the repo's `packages` folder.
+
+When a contributor sends a PR to `atom/atom` that only affects files in a folder under `packages`, only the specs for the relevant package folders will be executed using Atom's CI scripts.  This means that a full Atom build will not be required when no Atom Core code is changed in a PR.  Package specs are also now run against all 3 OSes on Atom `master` and release builds.
+
+Core packages that aren't consolidated still have folders under `packages` with README.md files that point to the home repository for that package.
+
+## Drawbacks
```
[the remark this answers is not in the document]
daviwil: I'm pretty sure that `apm link`'ed package forks should still work, but I'll add that to the list of things to verify in a possible prototype.

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,332 @@
...
+
+- Archive the repositories for consolidated core packages, but only after migrating existing issues, merging or closing existing PRs, and updating the README.md to point to the new home of the package code.
+
+Also, contributors will now have to fork, clone, and build `atom/atom` to contribute to core packages where they would previously just need to clone the package repository.  This might put added burden on them such as installing necessary build dependencies on their machine that they wouldn't otherwise need.  It is very likely we could simplify this process for them, though.
+
+One final drawback is that it will now be harder to have single-package maintainers. We currently have 7 core packages where there is a maintainer who isn't a part of the core Atom maintainers team.  These maintainers generally are able to merge community PRs and make commits to those packages with their own judgement.  If we get rid of individual package repositories, do we now make those maintainers full Atom maintainers?
```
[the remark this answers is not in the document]
daviwil: Great idea!  I wonder if it allows non-admins on the repo to be specified as code owners?

daviwil:
Thanks for the great feedback @jasonrudolph!

> 1. **Disabling a package**: Will it still be possible to disable a core package that has been moved into atom/atom?

Yep, that's the plan.  These packages will still be treated as packages even though they'll be part of the core Atom repo.  I'll call this out explicitly in the RFC.

> 2. **Package versions**: The biggest thing I'm unsure of is how package versioning will be affected:
>    - Based on the RFC, it seems to me that packages in atom/atom will no longer have version numbers and will no longer get published to atom.io. Is that true?
>    - If packages in atom/atom don't have version numbers anymore:
        - Do they still have a `package.json` file? If so, what happens to the `version` property in a package's `package.json` file?

The idea is that we'd just stop updating the version in `package.json`, though we'll still need that file to specify its dependencies.  From the point at which we integrate the package code into Atom's codebase, we'll consider its version to be the version of Atom it ships with, we just won't update its `package.json` to reflect that.

>        - Will the packages still be declared in the top-level `package.json` file in atom/atom? If so, how will we determine what version number to specify?

Nah, we'll take them out of that list and use a separate build step to walk the `packages` folder, running `npm install` in each of the folders which contain a `package.json` file.  One open question is where the final package folder will live in the app package,  `node_modules` or `packages`.  That's something I want to figure out with some quick prototyping so that I can answer it in the RFC.

>    - If the affected packages will no longer be published to atom.io, what should happen to the existing versions of those packages on atom.io? (For example, fast forward to a year from now: It would seem strange to me to have atom.io continue to show an outdated mid-2018 README for a package that got moved into atom/atom and then received substantial updates.)

Good question, we might need to come up with a strategy to hide or unpublish those packages on atom.io to remove confusion.

I'll capture the answers to these questions in the RFC!

daviwil:
Updated thinking on whether the packages will be listed in Atom's `package.json`: looks like you can use local, relative file paths to refer to dependencies with npm:

https://docs.npmjs.com/files/package.json#local-paths

This would allow us to keep those dependencies and benefit from node_modules dependency sharing.  Here's what it'd look like in `package.json`:

```
{
  "packageDependencies": {
    "tree-view": "file:./packages/tree-view"
  }
}
```

daviwil:
Just tried that and apparently `apm` expects the version field to be a semver-compatible string.  If we want to take that approach we'll have to tweak `apm` to expect file paths in the version field...  I'll look for another way for now.

daviwil:
I've got a WIP proof-of-concept of this working in PR #17686.  Pretty minimal changes required to achieve much of what's in the RFC.  I'll update the RFC on Monday based on what I've learned!

daviwil:
Hey folks, I've updated the RFC based on @jasonrudolph's feedback and updated some other details using what I learned while implementing this change in PR #17686.  I'm feeling good about the state of this RFC so please take a look at let me know if you think anything is missing or if there are other problems that should be addressed.

Thanks a lot!

Arcanemagus:
> Is there a good reason to not move the `language-*` packages into `atom/atom`?

Some other projects pull these languages directly from the `master` branch of these repos when those projects are built, so pulling these into `atom/atom` at minimum will require changes to the build scripts there.

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+Since packages all have their own repositories, this means that we have to watch 91 different repos for issues and pull requests.  This also means that we have to redirect issues filed on `atom/atom` to the appropriate repository when a user doesn't know where it belongs.  Even more importantly, there's not an easy way to prioritize and track issues across the Atom organization without using GitHub Projects.
+
+Also, as mentioned above, there's the added duty of doing the package "version dance" when we merge any new PRs to a package repository: publish the package update, update `package.json` in Atom.  It's very easy to forget to do this and not have community contributions included in the next Atom release!
+
+The more core packages live in `atom/atom`, the less work Atom maintainers have to do overall.
+
+## Explanation
+
+Many of Atom's core packages now live in the core `atom/atom` repository.  To the Atom user, this change will be imperceptible as these packages still show up in the list of Core Packages in the Settings View.  Users can still disable these packages if they want to replace their behavior with other packages.
```
jasonrudolph: > Users can still disable these packages if they want to replace their behavior with other packages.

You could disable a package with no intention of replacing its functionality. Right?

Suggested edit:

> Users can still optionally disable these packages.

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+| **[github]** |  |  |  |  | Independent project |
+
+### Consolidation Process
+
+To consolidate a single core package repository back into `atom/atom`, the following steps will be taken:
+
+1. All open pull requests on the package's repository must either be closed or merged before consolidation can proceed
+1. The package repository's code in `master` will be copied over to Atom's `packages` folder in a subfolder bearing that package's name.
+1. Atom's `package.json` file will be updated to change the package's `packageDependencies` entry to reference its local path with the following syntax: `"tree-view": "file:./packages/tree-view"`
+1. A test build will be created to manually verify the package loads and works correctly at first glance and also that package specs pass
```
jasonrudolph: Can you help me understand what this means? Does this mean locally running `script/build --install`, locally running the package's specs, and locally performing some manual sanity testing to verify the package's functionality?

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+Contributors now clone and build `atom/atom` to work on improvements to core packages.  They will no longer have to use `apm link` in dev mode to test changes they make to packages in the repo's `packages` folder.  Core packages that aren't consolidated still have folders under `packages` with README.md files that point to the home repository for that package.
+
+When a contributor sends a PR to `atom/atom` that only affects files in a folder under `packages`, only the specs for the relevant package folders will be executed using Atom's CI scripts.  This means that a full Atom build will not be required when no Atom Core code is changed in a PR.  Package specs are also now run against all 3 OSes on Atom `master` and release builds.
+
+Atom maintainers no longer have to publish new versions to consolidated core packages and then edit `package.json` to bump the package version in a particular Atom release branch (Stable, Beta, or `master`).  When a PR against a consolidated core package in `atom/atom` is merged, no version number change is required and the changes will immediately be a part of the next release from that branch.
+
+## Drawbacks
```
jasonrudolph: Following up from the discussion in https://github.com/atom/atom/pull/17674#discussion_r202082961:

> I'll add that to the list of things to verify in a possible prototype.

With the prototype in #17686, were you able to try out the forking/linking workflow discussed in https://github.com/atom/atom/pull/17674#discussion_r202082961?

jasonrudolph (commented):
Thanks for the updates, @daviwil. This is coming along nicely!

I noted just a few questions inline below.

docs/rfcs/003-consolidate-core-packages.md
daviwil: Yep, it works as expected in dev mode if you've used `apm link --dev` for the forked package.  In that PR I'm adding the `atom/packages` path to the list of possible package folders in dev mode *after* the user's `~/.atom/dev/packages` folder, meaning that their dev-linked packages get prioritized.

In the case that they do a plain `apm link` of a forked package and run Atom in normal, non-dev mode, their non-dev linked package will take precedence over the bundled package (just as it always did).

docs/rfcs/003-consolidate-core-packages.md
daviwil: That's correct, I'll update the language to be more clear!

docs/rfcs/003-consolidate-core-packages.md
daviwil: Yep!  I'll make that change.

daviwil:
@Arcanemagus you're absolutely right, we'll have to assess possible impact before moving any of the `language-*` packages since we don't know who might be depending on those repos.

maxbrunsfeld (approved):
Big 👍 for this proposal, I think it will be a big simplification. Thanks for writing it up @daviwil!

One additional drawback to doing this is that when making a change to a bundled package, there will be a longer CI feedback cycle, because the entire Atom CI process will need to run.

In some cases, (e.g. `language-javascript`) this will actually makes a lot of sense, because some test in Atom core (and maybe the test suites of other packages) actually activate that `language-javascript` and use it during tests.

For other packages like `go-to-line`, it won't make as much sense, because the package code can't possibly influence other parts of Atom's build. In the future, it might make sense to somehow automatically detect that only `go-to-line` changed and perform a simpler, fast build on CI.

daviwil:
Thanks @maxbrunsfeld!  I've actually got CI testing called out here:

> When a contributor sends a PR to `atom/atom` that only affects files in a folder under `packages`, only the specs for the relevant package folders will be executed using Atom's CI scripts.  This means that a full Atom build will not be required when no Atom Core code is changed in a PR.  Package specs are also now run against all 3 OSes on Atom `master` and release builds.

My plan is to update our CI test script to detect when there are only changes to packages in a PR and only run the specs for those packages.  However, I didn't consider the possibility of packages that might want to have a full CI run, so I might need to keep a whitelist of packages for which the full spec run should be performed anyway.

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+3. Edit `package.json` in the Atom repo to reflect the new version of `tree-view`
+4. Commit and push the changes to the relevant branch where the change belongs (`master` or `1.nn-releases`)
+
+#### Simplified Package Development
+
+If we were to move `tree-view` (or any other core Atom package) back into `atom/atom`, the development workflow would look more like this:
+
+1. Fork and clone `atom/atom` and switch to a release branch if necessary
+1. Build Atom and launch it in dev mode
+1. Make desired changes to the code in `packages/tree-view`
```
smashwilson: While you're developing, how easy is it to play with your changes in a live environment? Do you need to do a full `atom/atom` rebuild, or just launch in dev mode? What about if you change dependencies?

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+At this point, the change is merged into Atom and ready for inclusion in the next release.
+
+### Greatly reduced burden for maintainers
+
+Since packages all have their own repositories, this means that we have to watch 91 different repos for issues and pull requests.  This also means that we have to redirect issues filed on `atom/atom` to the appropriate repository when a user doesn't know where it belongs.  Even more importantly, there's not an easy way to prioritize and track issues across the Atom organization without using GitHub Projects.
+
+Also, as mentioned above, there's the added duty of doing the package "version dance" when we merge any new PRs to a package repository: publish the package update, update `package.json` in Atom.  It's very easy to forget to do this and not have community contributions included in the next Atom release!
+
+The more core packages live in `atom/atom`, the less work Atom maintainers have to do overall.
```
smashwilson: Yessss

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+For maintainers and contributors, there will be less juggling of repositories and no more publishing of updates to these packages with `apm`:
+
+Contributors now clone and build `atom/atom` to work on improvements to core packages.  They will no longer have to use `apm link` in dev mode to test changes they make to packages in the repo's `packages` folder.  Core packages that aren't consolidated still have folders under `packages` with README.md files that point to the home repository for that package.
+
+When a contributor sends a PR to `atom/atom` that only affects files in a folder under `packages`, only the specs for the relevant package folders will be executed using Atom's CI scripts.  This means that a full Atom build will not be required when no Atom Core code is changed in a PR.  Package specs are also now run against all 3 OSes on Atom `master` and release builds.
```
smashwilson: 👍 I like that we get faster feedback on PR builds, but still have a full build cycle to validate the changes in context. (And our full build times could likely be dropped further with another round of optimizations at some point as well...)

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+Atom maintainers no longer have to publish new versions to consolidated core packages and then edit `package.json` to bump the package version in a particular Atom release branch (Stable, Beta, or `master`).  When a PR against a consolidated core package in `atom/atom` is merged, no version number change is required and the changes will immediately be a part of the next release from that branch.
+
+## Drawbacks
+
+One possible drawback of this approach is that there might be some initial confusion where core Atom packages live, especially if some are consolidated into `atom/atom` and others still live in their own repositories.  We will manage this confusion by doing the following:
+
+- Include folders for _all_ core packages in the `packages` folder of the Atom repo and add README.md files to folders of those packages that still live in separate repos.  This will allow us to direct users to the proper home for packages that are not yet consolidated.
```
smashwilson: Personally I'd find this pretty confusing - I'd want to be able to do an `ls ./packages` to see what _is_ in the monorepo. Think we could do a single README at `./packages/README.md` that lists all core packages and breaks out which lives where, rather than having a bunch of README-only directories?

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+- Archive the repositories for consolidated core packages, but only after migrating existing issues, merging or closing existing PRs, and updating the README.md to point to the new home of the package code.
```
smashwilson: By the way, I'm pretty sure we could bulk-migrate open pull requests with the GraphQL API and some `git subtree` magic.

docs/rfcs/003-consolidate-core-packages.md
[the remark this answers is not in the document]
smashwilson: ... What does archive-view even do? 👀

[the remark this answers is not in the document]
smashwilson: <even-further-off-topic> It'd be nice to be able to re-evaluate what's bundled periodically, especially in terms of the language packages...

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+These packages will not be consolidated for the following reasons:
+
+| Package | Open Issues | Open PRs | Outside Maintainers | Last Updated | Reason |
+|---------|-------------|----------|---------------------|--------------|-------|
+| **[autocomplete-atom-api]** |  |  |  |  | Blocks contribution from Facebook |
+| **[autocomplete-css]** |  |  |  |  | Same as above |
+| **[autocomplete-html]** |  |  |  |  | Same as above |
+| **[autocomplete-plus]** |  |  |  |  | Same as above |
+| **[autocomplete-snippets]** |  |  |  |  | Same as above |
+| **[github]** |  |  |  |  | Independent project |
```
smashwilson: I actually wouldn't mind seeing atom/github consolidated into the main repo at some point. We certainly have our fair share of issues opened in atom/atom that our support folks need to migrate. Are there any technical obstacles beyond "it's big"?

docs/rfcs/003-consolidate-core-packages.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Consolidation Process
+
+To consolidate a single core package repository back into `atom/atom`, the following steps will be taken:
+
+1. All open pull requests on the package's repository must either be closed or merged before consolidation can proceed
+1. The package repository's code in `master` will be copied over to Atom's `packages` folder in a subfolder bearing that package's name.
```
smashwilson: Do we care about preserving blame and history? We could do it with [git subtree](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt) :sparkles:

smashwilson (approved):
A few random comments, no blockers 👍

docs/rfcs/003-consolidate-core-packages.md
daviwil: I like that, that's a better idea.  Will change this!

docs/rfcs/003-consolidate-core-packages.md
daviwil: Just launch in dev mode!  However, I think you found the first major issue with how I implemented PR #17686: the folders under `packages` never have `apm install` run directly on them so their `node_modules` folders are not populated by the time they're loaded in dev mode.

Asking the user to go run `apm install` inside the package folders they want to work on sounds like a lot of extra friction so I'll have to think about this a bit.  Worst case scenario we can use the original strategy I started with, running `apm install` on each package folder individually and then copying those folders into Atom's`node_modules` during the full build, possibly running `npm dedupe` afterward.

docs/rfcs/003-consolidate-core-packages.md
daviwil: Certainly could work, and might be easier than going through and trying to merge or close them all.  We'd probably lose existing review comments though, not sure if it's possible to recreate them all in a faithful way

docs/rfcs/003-consolidate-core-packages.md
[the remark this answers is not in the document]
daviwil: Maybe this would be a good time to start that process?  If certain packages aren't that useful anymore, I'd love to trim some fat.

docs/rfcs/003-consolidate-core-packages.md
daviwil: It'd probably be no different than any other package, I think it's feasible.  My thinking was that consolidating the github package would slow you down, but maybe it will actually make things easier for you?  I'll move it to the "maybe" list :)

docs/rfcs/003-consolidate-core-packages.md
daviwil: I suppose that would be ideal, but I wonder if pulling over commits from so many repos would have an adverse effect on the Atom repo (here, have 15,000 new commits, you're welcome!).  I'll pose that question to everyone before this RFC is complete.

daviwil:
Adding @lee-dohm to get his take on possible impact of moving packages, issues, and PRs around.

lee-dohm:
I think this is an interesting idea and definitely can have some benefits. Thoughts:

* Archiving the package repos is :+1: because there are four years of links from various locations into the package repos for examples, tutorials, suggestions on how to do things in other people's packages, etc
* I'm all for moving the issues _en masse_ from the package repos into atom/atom
* I'm not so sure about moving PRs, we might want to spend a little time curating before moving them? We've let a lot of them lay fallow for too long. I suppose we can curate them after moving them too.

Like I say, I think this can have a lot of workflow benefits and simplifying triage is always a good thing :grinning:

docs/rfcs/003-consolidate-core-packages.md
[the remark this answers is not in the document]
jasonrudolph: > ... What does archive-view even do? 👀

As far as I can tell, its most robust feature is its ability to consume the time of Atom's maintainers. :trollface::sob:

https://github.com/atom/archive-view/pull/62
https://github.com/atom/archive-view/pull/65

docs/rfcs/003-consolidate-core-packages.md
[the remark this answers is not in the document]
the-j0k3r: A small non related comment, I see thoughout this doc you're typing out two spaces after a full stop, but its not causing the `<br>` You are probably trying to trigger. I think only happens if you add two spaces and the new text is on another line, i.e. pressing enter after the trailing spaces.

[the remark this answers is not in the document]
daviwil: @the-j0k3r the double-space after the period is a long-held typing habit of mine and isn't meant in this case to trigger newlines :)

daviwil:
OK folks, updated once again based on feedback.  One actionable remaining question is whether we should use `git subtree` to migrate over a package repo's full commit history or leave the commit history in the original package repository?

Here are the pros and cons I see for migrating the full commit history:

**Pros**

- Easy to dive into a package's commit history inside of `atom/atom`, run `git blame`, etc, as if the code was always in this repo
- Community contributors to those packages get credit for their commits in `atom/atom`

**Cons**

- Possibly drags 10,000+ commits into `atom/atom`, greatly enlarging the commit history and causing unknown issues
- Slightly more complex package consolidation steps
- Consolidation PRs pages/diffs might possibly have issues with the number of commits involved

So Atom maintainers, should we migrate the package commit history? :+1: or :-1: on this comment!

daviwil:
Since I got a pretty evenly-split response on the `git subtree` question, I think I'll err on the side of not carrying over the history on any repos we migrate for now since there's more of a likelihood of problems from bringing them over than there are benefits.

To move forward with this RFC, I'm planning to set it to **Accepted** and merge it at **3PM PST** today unless I hear any objections before then.  Thanks again for all the feedback!

lee-dohm (approved):

PR #17680, opened 2018-07-12
Use a consistent app naming strategy across non-stable channels

daviwil:
### Description of the Change

This change makes the naming of the Atom application consistent across all non-stable channels (Beta, Nightly, Dev) to simplify the logic of scripts which consume these channels (like `atom/ci`).  This builds on some changes I made in PR #17538 which generalize some of the app naming code we were using to take other non-stable channels into account (like Nightly).

The only real impact this change has is that the .app file that's created from a `master` build on macOS will now be called **Atom Dev.app**.  I'm bringing this to general awareness in case anyone has any personal scripts for regularly building and using Atom from `master`.

This issue originally came up when @smashwilson found that the `atom/ci` script didn't properly find the Atom Dev build it unpacks on Linux because the folder names have changed.  While trying to fix that script, I realized that the Windows ZIP contained a folder called "Atom null" because of a bug in the code.  This change fixes it for all cases.

### Alternate Designs

Special case the dev channel to be called "Atom" and not "Atom Dev".  I'm fine with doing this if folks think it's a better approach.

It's part of the Atom build process.

### Benefits

Consistent naming for non-stable channel builds, easier to predict what the app name will be without special casing.

### Possible Drawbacks

Breaking some existing workflows, but not terribly likely and easy to fix if so.

### Verification Process

- [x] Let CI complete and verify artifacts that are generated to make sure "Atom Dev" is being used as the app name on all 3 platforms:
  - [x] Windows
  - [x] macOS
  - [x] Linux

### DIFF
--- a/script/config.js
+++ b/script/config.js
@@ -56,7 +56,7 @@ function getChannel (version) {
   return 'stable'
 }
 
-function getAppName(channel) {
+function getAppName (channel) {
   return channel === 'stable'
     ? 'Atom'
     : `Atom ${process.env.ATOM_CHANNEL_DISPLAY_NAME || channel.charAt(0).toUpperCase() + channel.slice(1)}`
--- a/script/config.js
+++ b/script/config.js
@@ -22,13 +22,13 @@ const appMetadata = require(path.join(repositoryRootPath, 'package.json'))
 const apmMetadata = require(path.join(apmRootPath, 'package.json'))
 const computedAppVersion = computeAppVersion(process.env.ATOM_RELEASE_VERSION || appMetadata.version)
 const channel = getChannel(computedAppVersion)
-const channelDisplayName = getChannelDisplayName(channel)
+const appName = getAppName(channel)
 
 module.exports = {
   appMetadata,
   apmMetadata,
   channel,
-  channelDisplayName,
+  appName,
   computedAppVersion,
   repositoryRootPath,
   apmRootPath,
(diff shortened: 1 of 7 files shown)


daviwil:
image

script/config.js
```diff
@@ -56,9 +56,10 @@ function getChannel (version) {
   return 'stable'
 }
 
-function getChannelDisplayName (channel) {
-  if (channel === 'stable' || channel === 'dev') return null
-  return process.env.ATOM_CHANNEL_DISPLAY_NAME || channel.charAt(0).toUpperCase() + channel.slice(1)
+function getAppName (channel) {
+  return channel === 'stable'
+    ? 'Atom'
+    : `Atom ${process.env.ATOM_CHANNEL_DISPLAY_NAME || channel.charAt(0).toUpperCase() + channel.slice(1)}`
```
smashwilson: Custom display name support, eh? Time to build "Atom 3000" :sparkles:

smashwilson (approved):
I'm 👍 on this because I'm a sucker for consistency across platforms and across channels. Being able to install stable, beta, and dev side-by-side on macOS would be really nice too. Although it might warrant an `@here` in the maintainer's chat if this is merged 😄 

...what I might do is use a bash alias to map `atom` to `atom-dev` on my Mac, btw, just to keep my muscle memory working :wink:

script/config.js
daviwil: Haha, yeah, just in case we want some one-off release branches :)

daviwil:
Which reminds me, I need to fix atom.sh for this too!

daviwil:
Aaaand now I need to update the GitHub package to use `Atom Dev Helper.app` instead of `Atom Helper.app` :man_facepalming:

PR #17686, opened 2018-07-13
Enable repo-local core packages in the 'packages' folder

daviwil:
### Description of the Change

This PR is the first step to implementing [RFC 003](https://github.com/atom/atom/pull/17674), enabling core Atom packages to live inside of the `packages` folder of the Atom repo while still being treated as packages.

I originally created this PR with the `about` and `welcome` packages moved into the `packages` folder but I decided to split those out into separate PRs so that we can more closely follow the [process described in RFC 003](https://github.com/atom/atom/blob/master/docs/rfcs/003-consolidate-core-packages.md#consolidation-process).

### Alternate Designs

There is [another possible design](https://github.com/atom/atom/commit/834320361be4a5b0b0294ed95cab9b535ae9ce1f) that does not require a change to `apm` and instead includes the `packages` folder into Atom's app package, but this approach has the following drawbacks:

- More code needed to make these packages show up as bundled packages in Atom
- Code changes are needed in Atom to include each individual package in the startup snapshot
- We no longer have a list of core packages in `packageDependencies` that live inside of `atom/atom`

Ultimately, installing these local package dependencies into `node_modules` is a cleaner and more consistent experience to what we do today.

This enables some core packages to be moved back into `atom/atom` to remove the need for separate repositories.

### Benefits

Some technical benefits of this approach:

- Minimal code changes are necessary to move a core package under `packages`.  Just copy the package's code into a folder under `packages` and then change Atom's `package.json` to use `file:./packages/package-name` in the version field of the package's `packageDependencies` entry.
- A user can easily test changes to package code under `packages` just by running Atom in dev mode.
- It's easy for an Atom developer to see which packages are repo-local just by looking at `packageDependencies` in `package.json`

See [RFC 003](https://github.com/atom/atom/pull/17674) for more benefits.

### Possible Drawbacks

One possible drawback is that the change to `apm` to support local package dependency paths doesn't yet have a way to determine that the package under `packages` and its "installed" copy under `node_modules` are the same and don't require another install.  This is mainly due to the fact that we no longer change the version in the package's `package.json` file, so it's hard to know for certain whether the package code has changed between builds.  See the comments in atom/apm#797 for more details.

On the other hand, no tarball has to be downloaded any longer, so npm's own internal caching of dependencies should render the time required to "install" the package relatively moot.

See [RFC 003](https://github.com/atom/atom/pull/17674) for other possible drawbacks.

### Verification Process

- [x] Building Atom causes packages under `packages` to be included in Atom's `node_modules` folder
- [x] Packages from `packages` show up as Core Packages in Atom's settings view
- [x] Running Atom in dev mode enables changes to packages under `packages` to be reflected in the next reload without rebuilding atom, no `apm link` required.
- [x] Running `atom --dev` with `ATOM_DEV_RESOURCE_PATH` set causes that path to be used as the dev resource path
- [x] Running `atom --dev` with **no** `ATOM_DEV_RESOURCE_PATH` set causes current directory to be used as dev resource path only if the current directory appears to contain an Atom repo
- [x] Running `atom --dev` with **no** `ATOM_DEV_RESOURCE_PATH` set in a directory that isn't an Atom repo causes the default Atom repo path `~/github/atom` to be used

### Applicable Issues

- RFC 003: https://github.com/atom/atom/pull/17674
- Relevant change to `apm` to enable this: https://github.com/atom/apm/pull/797

### Tasks

- [x] Add tests to apm changes: https://github.com/atom/apm/blob/dw-repo-local-core-packages/spec/install-spec.coffee#L306
- [x] Update to latest apm release with repo-local packageDependency support
- [x] Move `about` and `welcome` package migrations out of this PR to be performed separately, following our established consolidation process
- [x] Delete [other prototype branch](https://github.com/atom/atom/commit/834320361be4a5b0b0294ed95cab9b535ae9ce1f) if this gets approved and merged


daviwil:
OK, this is ready for review!

My priority in this PR was to make the core package dev process from cloning `atom/atom` to running `atom --dev` as smooth as possible.  This change enables the following workflow:

1. Install Atom (at least a channel using Electron compatible with `master`)
2. Clone `atom/atom`
3. Run `script/dev` to install all dependencies and then launch Atom in dev mode

This removes the need for a contributor to need to do a full build of Atom or to understand how to set up `ATOM_DEV_RESOURCE_PATH` to run Atom in dev mode.

Would love to hear feedback on any part of this!

script/dev
```diff
@@ -0,0 +1,55 @@
...
+
+// Install the local core packages in-place so they can be used in dev mode
+var files = glob.sync(path.join(CONFIG.repositoryRootPath, 'packages/*/package.json'))
+if (files.length > 0) {
+  console.log('\nInstalling core packages for use in dev mode...')
+
+  files.forEach(file => {
+    const packageDir = path.dirname(file)
+    process.stdout.write(`Installing packages/${path.basename(packageDir)} `)
+    runApmInstall(path.dirname(file), false, true)
```
daviwil: I'm suppressing apm's output here so that I can write my own similar string for each package so that we don't see `Installing modules...` 50 times in a row :)

daviwil: Also realized I should have used packageDir in the first argument here, will fix it

script/dev
```diff
@@ -0,0 +1,55 @@
...
+      return process.stdout.write('done\n'.green);
+    } else {
+      return process.stdout.write('\u2713\n'.green);
+    }
+  })
+}
+
+// Launch Atom in dev mode
+let atomToLaunch =
+  (process.argv.length > 2 && process.argv[2]) ||
```
jasonrudolph: What additional arguments are supported when running `script/dev`? Is there a way that we help people discover this information (e.g., maybe `script/dev --help` could describe this script's usage)?

script/dev
```diff
@@ -0,0 +1,55 @@
...
+  windowsHide: true,
+  env: Object.assign(process.env, { ATOM_DEV_RESOURCE_PATH: CONFIG.repositoryRootPath })
+})
+
+atomProcess.on('error', e => {
+  if (e.code === 'ENOENT') {
+    console.error(`The executable '${atomToLaunch}' could not be found!\n`.red)
+  } else {
+    console.error(`An error occurred when attempting to launch '${atomToLaunch}'\n`.red, e)
+  }
```
jasonrudolph: Suggestion: Add steps to the [verification process in the pull request body](https://github.com/atom/atom/pull/17686#issue-201416121) to verify that these two codepaths are working as expected.

script/lib/run-apm-install.js
```diff
@@ -4,7 +4,7 @@ const childProcess = require('child_process')
 
 const CONFIG = require('../config')
 
-module.exports = function (packagePath, ci) {
+module.exports = function (packagePath, ci, stderrOnly) {
```
jasonrudolph: Food for thought: Without additional context, the name `stderrOnly` doesn't feel very expressive to me. How would you feel about changing it to something like `stdioOptions` with a default value of `'inherit'`?  And in the places that want to deviate from the default (e.g., the new code in `script/dev`), we can pass a custom value (e.g., `['inherit', 'pipe', 'inherit']`.

What do you think?

jasonrudolph:
> Running Atom in dev mode enables changes to packages under packages to be reflected in the next reload without rebuilding atom, no apm link required.

jasonrudolph (commented):
I've noted a few questions and suggestions inline below. I hope this helps!

maxbrunsfeld (commented):
> Run `script/dev` to install all dependencies and then launch Atom in dev mode

> This removes the need for a contributor to need to do a full build of Atom or to understand how to set up `ATOM_DEV_RESOURCE_PATH` to run Atom in dev mode.

In the future, when we have many packages in the `packages` folder, I would think installing dependencies would take significant time. Even now, redundant runs of `script/bootstrap` take 17 seconds on my laptop.

This seems like something I'd want to keep as an explicit separate step from the default way that I start Atom in development. Could we address the papercut about `ATOM_DEV_RESOURCE_PATH` by defaulting the `devResourcePath` to `process.cwd()` in the case where the environment variable isn't set and `~/github/atom` (the current default) does not exist? Maybe we could check if you're currently in the atom repo by looking for `$PWD/atom.sh` or some other distinctive file.

smashwilson (changes requested):
This is pretty cool 👍 I'm on board with this approach.

* I'd like to see `script/dev` pass its arguments to the launched Atom instance. Otherwise we can't use this to iterate on Atom bugs that only manifest in non-Atom projects.
* To address @maxbrunsfeld's point about bootstrapping taking so long: in addition to improving dev resource path autodetection, maybe we could only invoke it if `scripts/node_modules/` or `apm/node_modules/` are missing? That would preserve the goal of minimizing manual steps from fresh clone to development iteration, but also let us avoid redundant bootstrapping.

smashwilson:
> Even now, redundant runs of script/bootstrap take 17 seconds on my laptop.

Yikes. 😱 

See, this seems like a problem to me regardless. script/bootstrap is invoked even if you try to run `script/build --help`. atom/apm#808 should help there but it sounds like we have more work to do 🔧 .

script/dev
```diff
@@ -0,0 +1,55 @@
...
+
+  files.forEach(file => {
+    const packageDir = path.dirname(file)
+    process.stdout.write(`Installing packages/${path.basename(packageDir)} `)
+    runApmInstall(path.dirname(file), false, true)
+
+    if (process.platform === 'win32') {
+      return process.stdout.write('done\n'.green);
+    } else {
+      return process.stdout.write('\u2713\n'.green);
```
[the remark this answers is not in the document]
50Wliu: I remember modifying `script/lint` to lint JS files in the `script` folder - turns out I set the glob pattern to files ending in `.js` only.  I'll take a look in a different PR :)

maxbrunsfeld:
> `script/bootstrap` is invoked even if you try to run `script/build --help`

Ha, yeah that's pretty bad. We could avoid that just by explicitly special-casing `--help` though.

> To address maxbrunsfeld's point about bootstrapping taking so long: in addition to improving dev resource path autodetection, maybe we could only invoke it if scripts/node_modules/ or apm/node_modules/ are missing? That would preserve the goal of minimizing manual steps from fresh clone to development iteration, but also let us avoid redundant bootstrapping.

I also think it's ok if people have to run `script/bootstrap` manually though. Overall, I think it's valuable to avoid unnecessary differences between production and development, and I question the value of bundling build and launch into one step.

script/dev
```diff
@@ -0,0 +1,55 @@
...
+
+'use strict'
+
+// Run bootstrap first to ensure all the dependencies used later in this script
+// are installed.
+require('./bootstrap')
+
+const path = require('path')
+const glob = require('glob')
+const colors = require('colors')
```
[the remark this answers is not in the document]
daviwil: No assign needed, changing it

script/dev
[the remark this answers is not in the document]
daviwil: Fixed the linter errors :paintbrush:

script/dev
daviwil: Added and verified

script/lib/run-apm-install.js
daviwil: Definitely better, I'll go with that!

daviwil:
@maxbrunsfeld and @smashwilson: Thanks for the feedback!  Would it be better if I just got rid of `script/dev` completely and moved the package installation logic back into `script/bootstrap`?  The first-time workflow for this would be:

1. Install Atom Nightly (for best results)
1. Clone the Atom repo
1. Run `script/bootstrap` to install Atom dependencies and those of the packages under `packages`
1. Run `atom --dev .`

The last step would be dependent on me implementing @maxbrunsfeld's suggestion to make Atom smarter about finding the `ATOM_DEV_RESOURCE_PATH`, which shouldn't be too hard.  Might be able to look for `package.json` in `process.cwd()` and see if it contains `"name": "atom"` in the first 5 lines.

If the developer ever updated package dependencies in `packages/somepackage` they could just manually run `apm install` there or run `script/bootstrap` all over again.

What do you think?

maxbrunsfeld:
> What do you think?

That seems like a good setup to me. But if anybody has strong feelings in favor of `script/dev`, I don't want to crush the idea.

daviwil:
OK folks, I've responded to all the feedback here so I think this is ready for another pass.  Summary of the changes:

- Moved package bootstrapping back into `script/bootstrap`
- Taught Atom how to look for an atom repo at `process.cwd()` and use it as the dev resource path when `--dev` is passed if `ATOM_DEV_RESOURCE_PATH` is not set
- Deleted `script/dev`
- Changed the `stderrOnly` param of `runApmInstall` to `stdioOptions` for more flexibility

Let me know if there's anything else you'd like to see!

smashwilson (approved):

smashwilson:
Approved ✅ 

make it so

daviwil:
Thanks all!  Merging it :ship:

Issue #17694, opened 2018-07-16
Option to disable discrete GPU


jcayzac:
## Summary

Atom will by default use any discrete GPU available. It used to be possible to disable this by launching it with `--disable-gpu` but Electron [recently removed this feature](https://github.com/electron/electron/commit/63c34f3dcf5f1a87a40e7cbf9061f49c1f3a89fa) and apps now have to explicitly call `app.disableHardwareAcceleration()` instead.

## Motivation

Using the discrete GPU typically eats twice the battery as the integrated one, and turn laptops into furnaces.

lee-dohm:
Thanks for reaching out.

Electron will still pass along the `--disable-gpu` flag to Chromium when passed at the command line and Chromium still supports the `--disable-gpu` flag. The change you reference is a documentation change because it won't work when passed via `appendSwitch()`, a different mechanism than passing on the command-line.

I hope that helps!

jcayzac:
@lee-dohm thanks for the quick answer.

However, passing `--disable-gpu --disable-gpu-compositing` to the latest version of Atom doesn’t prevent it from switching to the discrete GPU (whereas the “Disable Hardware Acceleration” setting in Chromium achieves exactly that).

It’s not clear to me how Atom passes those options to Electron so I don’t know where the problem lies —just that this is currently broken.

Cheers,
Julien.

lee-dohm:
Interesting, thanks!

@daviwil Would you mind looking into this?

jcayzac:
If it help, I’m running the latest version of macOS Sierra on a MBP from 2010, and use https://github.com/codykrieger/gfxCardStatus to show switches from integrated to discrete.

simurai:
Somewhat related: There are [two reports](https://discuss.atom.io/t/invisible-text-in-ui-text-boxes/56120) about "text not showing in mini editors" and the only solution so far seems to open Atom with the `--disable-gpu` flag.

jcayzac:
More information: on my newer (early 2013) MBP Atom keeps using the integrated GPU even without using the `--disable-*` flags.

How which GPU to use is chosen looks mysterious 😆

jcayzac:
> However, passing `--disable-gpu --disable-gpu-compositing` to the latest version of Atom doesn’t prevent it from switching to the discrete GPU (whereas the “Disable Hardware Acceleration” setting in Chromium achieves exactly that).

Actually, it has an effect!

```sh
[...]

PR #17699, opened 2018-07-16
Added option to disable hardware acceleration

brennenberkley:
Implements #17694


lee-dohm:
Thanks for contributing :grinning:

As stated in https://github.com/atom/atom/issues/17694#issuecomment-405289704, Electron (and Atom) still passes along the `--disable-gpu` flag to Chromium when specified on the command line. We also don't want this to be a configuration option. Because of these reasons, I'm going to close this.

Thanks again for the assistance :bow:

PR #17702, opened 2018-07-17
Modify TextEditorComponent updateClassList to always add managed classes to element

rholinshead:
### Description of the Change

Currently, the `updateClassList` method on the `TextEditorComponent` does not re-add the managed classes (editor, mini, is-focused) whenever the `<atom-text-editor>` element is re-rendered with a changed class. This PR fixes the issue by always adding the `newClassList` classes to the element, relying on `element.classList.add` to determine if the classes already exist (and should be ignored).

### Alternate Designs

N/A

This address an issue that exists in one of the main core classes, which is used within other atom/atom classes. Implementing the change in a separate package would add an unnecessary additional dependency

### Benefits

This change will allow external packages to dynamically set <atom-text-editor> element classes without having to worry about losing the managed classes (editor, mini, is-focused). This is especially useful in React components that reuse an <atom-text-editor> element when re-rendering

### Possible Drawbacks

This change will affect packages' styling of atom-text-editor elements which rely on the managed classes not being re-applied. In such cases, the packages can regain specificity by adding the appropriate managed class names to the selectors.

### Verification Process
 - Added new spec test and verified passing tests
 - Verified by modifying the `find-and-replace` package's `project-find-view` to set the findEditor's class based on the text (see video), and ensuring the managed classes remain on the element
 - Manually tested <atom-text-editor> elements with changes to verify correct functionality

#### Before:
updateclasslist_before

#### After:
updateclasslist_after

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -818,6 +818,18 @@ describe('TextEditorComponent', () => {
       expect(element.className).toBe('editor a b')
     })
 
+    it('does not blow away class names managed by the component when packages change the element class name', async () => {
+      assertDocumentFocused()
+      const {component, element, editor} = buildComponent({mini: true})
+      element.classList.add('a', 'b')
+      element.focus()
+      await component.getNextUpdatePromise()
+      expect(element.className).toBe('editor mini a b is-focused')
+      element.className = 'a c d';
+      await component.getNextUpdatePromise()
+      expect(element.className).toBe('a c d editor is-focused mini')
+    })
+
     it('ignores resize events when the editor is hidden', async () => {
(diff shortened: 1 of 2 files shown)


lee-dohm:
@maxbrunsfeld Can you take a look at this?

maxbrunsfeld:
This seems fine. [DOMTokenList.add](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add) is already a noop if the provided class is already present.

maxbrunsfeld:
Thanks @rholinshead!

rholinshead:
My pleasure. Thanks for the quick review and merge @maxbrunsfeld @lee-dohm!

PR #17704, opened 2018-07-17
Update GitHub package on Beta

smashwilson:
Bring the improvements from dev over to the beta hotfix branch to ensure they're included in the next stable release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "exception-reporting": "0.43.1",
     "find-and-replace": "0.215.11",
     "fuzzy-finder": "1.8.2",
-    "github": "0.17.0",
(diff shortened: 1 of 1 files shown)


daviwil:
Good idea!  Should we wait to merge this until just after we roll the railcars, probably late next week?  Never know if we'll end up needing to do another hotfix for Electron 2.0 stuff

smashwilson:
I can double-check what the changelog is for `v0.17.0..v0.17.2`, but generally we only merge bugfixes on older release branches, so this should be fairly safe.

daviwil:
Ahhh, I was assuming this was bumping to 0.18.0, should have read the friggin diff ;)

PR #17721, opened 2018-07-20
Add TreeSitterLanguageMode::getSyntaxNodeContainingRange and TreeSitterLanguageMode::getSyntaxNodeAtPosition

queerviolet:
Adds:
  * `TreeSitterLanguageMode::getSyntaxNodeContainingRange(range: Range, where?: SyntaxNode -> Boolean)`
  * `TreeSitterLanguageMode::getSyntaxNodeAtPosition(range: Range, where?: SyntaxNode -> Boolean)`

These methods let you find a `SyntaxNode` at a location, optionally specifying a predicate function.

Changes:
  * `bufferRangeForScopeAtPosition(position)` to `bufferRangeForScopeAtPosition(selector, position)`.
  * `getRageForSyntaxNodeAtPosition(position)` to `getRangeForSyntaxNodeAtPosition(selector, where)`.

This brings it in line with the `TextMateLanguageMode` interface. The selector may be a string selector, or a predicate function of type `SyntaxNode -> Boolean`.

Needed for https://github.com/atom/toggle-quotes/issues/57.

### DIFF
--- a/src/selectors.js
+++ b/src/selectors.js
@@ -4,7 +4,7 @@ const {isSubset} = require('underscore-plus')
 
 // Private: Parse a selector into parts.
 //          If already parsed, returns the selector unmodified.
-// 
+//
 // * `selector` a {String|Array<String>} specifying what to match
 // Returns selector parts, an {Array<String>}.
 function parse (selector) {
--- a/src/selectors.js
+++ b/src/selectors.js
@@ -1,0 +1,42 @@
+module.exports = {selectorMatchesAnyScope, matcherForSelector}
+
+const _ = require('underscore-plus')
+
+/**
+ * Parse a selector into parts. If already parsed, returns the selector
+ * unmodified.
+ * 
+ * @param {String|Array<String>} selector
+ * @returns {Array<String>} selector parts
+ */
+function parse (selector) {
+  return typeof selector === 'string'
+    ? selector.replace(/^\./, '').split('.')
+    : selector
+}
+
+const always = scope => true
+
+/**
+ * Return a matcher function for a selector.
+ * 
+ * @param {String} selector
+ * @returns {(scope: String) -> Boolean} a matcher function
+ */
+function matcherForSelector (selector) {
+  const parts = parse(selector)
+  return selector
+    ? scope => _.isSubset(parts, parse(scope))
+    : always
+}
+
+/**
+ * Return true iff the selector matches any provided scope.
+ * 
+ * @param {String} selector
+ * @param {Array<String>} scopes
+ * @returns {Boolean} true if any scope matches the selector
+ */
+function selectorMatchesAnyScope (selector, scopes) {
+  return !selector || scopes.some(matcherForSelector(selector))
+}
--- a/src/selectors.js
+++ b/src/selectors.js
@@ -1,14 +1,12 @@
 module.exports = {selectorMatchesAnyScope, matcherForSelector}
 
-const _ = require('underscore-plus')
+const {isSubset} = require('underscore-plus')
 
-/**
- * Parse a selector into parts. If already parsed, returns the selector
- * unmodified.
- * 
- * @param {String|Array<String>} selector
- * @returns {Array<String>} selector parts
- */
+// Private: Parse a selector into parts.
+//          If already parsed, returns the selector unmodified.
+// 
+// * `selector` a {String|Array<String>} specifying what to match
+// Returns selector parts, an {Array<String>}.
 function parse (selector) {
   return typeof selector === 'string'
     ? selector.replace(/^\./, '').split('.')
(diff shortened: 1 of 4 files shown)


src/selectors.js
```diff
@@ -0,0 +1,42 @@
...
+
+const _ = require('underscore-plus')
+
+/**
+ * Parse a selector into parts. If already parsed, returns the selector
+ * unmodified.
+ * 
+ * @param {String|Array<String>} selector
+ * @returns {Array<String>} selector parts
+ */
```
maxbrunsfeld: There's a slightly different (and more unusual 😬) style of doc comments called [AtomDoc](https://github.com/atom/atomdoc) that we usually use in this codebase. There's good examples of it in [this file](https://github.com/atom/atom/blob/master/src/text-editor.js#L2565-L2576). I wouldn't necessarily be opposed to switching to JSDoc at some point. In the meantime, could adjust these to use atomdoc?

src/tree-sitter-language-mode.js
```diff
@@ -348,25 +349,28 @@ class TreeSitterLanguageMode {
   Section - Syntax Tree APIs
   */
 
-  getRangeForSyntaxNodeContainingRange (range) {
+  getRangeForSyntaxNodeContainingRange (range, selector) {
```
maxbrunsfeld: How do you imagine the compatibility working for this? The legacy TextMate selectors often target multiple scopes like `string.quoted`, but in a Tree-sitter syntax tree, a node never has more than one `type`; strings for example are just `string`.

I'd imagine that we would want the selector `.string` to match a `string` node, but what do we do about `.string.quoted`? It seems like there's no way around changing callers to provide new selectors. Did you envision changing `toggle-quotes` to just pass `.string` instead of `.string.quoted`?

maxbrunsfeld (commented):
Thanks for jumping on this bug!

It looks like we need some tests for `getRangeForSyntaxNodeContainingRange ` in `tree-sitter-language-mode-spec.js`. Maybe in those tests, you can show the ways (if any) that we deal with selectors that were written with a TextMate parser in mind.

src/tree-sitter-language-mode.js
```diff
@@ -348,25 +349,30 @@ class TreeSitterLanguageMode {
...
     const startIndex = this.buffer.characterIndexForPosition(range.start)
     const endIndex = this.buffer.characterIndexForPosition(range.end)
     const searchEndIndex = Math.max(0, endIndex - 1)
 
+    const matches = matcherForSelector(selector)
+
```
Arcanemagus: Trailing whitespace here 😛

src/tree-sitter-language-mode.js
queerviolet: Yeah, the backcompat story here is a bummer.

I did actually change `toggle-quotes` to do that, and it... well, it works for that use case with the JS grammar ;) Perhaps even for most grammars—I imagine the C family and Ruby grammars call their quoted strings `string`s. HTML, meanwhile, does not, calling its text `text` IIRC. (OTOH, HTML has `attribute_value`s, and that's a bummer, because it means that we can't toggle quotes on those (though we couldn't before, either)).

One alternative I considered is looking up the scope map and querying on those. But it would be nice if we could start divesting ourselves of the TextMate CSS classes, particularly for this query interface. We *should* provide a more expressive query interface than just matching on node type, though—ideally, what `toggle-quotes` could do is say, "find me the closest node starting with `"|'` and ending with `"|'`", which would be quite doable.

I guess the question is, how bad is it to semi-break the behavior of `bufferRangeForNodeInRange`? I mean, it was completely broken before, and this is the first time someone's noticed.

src/tree-sitter-language-mode.js
queerviolet: 👍

src/selectors.js
queerviolet: 👍

src/tree-sitter-language-mode.js
maxbrunsfeld: > We should provide a more expressive query interface than just matching on node type, though—ideally, what toggle-quotes could do is say, "find me the closest node starting with "|' and ending with "|'", which would be quite doable.

Yeah, you're right. That's the logic that `toggle-quotes` should be using; it probably shouldn't rely on the names of any syntax nodes.

But I don't necessarily think we need to build any special query language for finding that syntax node; we should just allow the package to directly inspect the syntax nodes and walk up the tree using `SyntaxNode.parent`. What do you think of adding a new experimental API called something like `syntaxNodeForPosition()` or `syntaxNodeForRange()` that `toggle-quotes` can use to inspect the syntax tree manually?

queerviolet: Thing is, there are potentially multiple trees at that point, and the deepest one may not be right.  Consider toggling quotes inside the rhs of `document.body.innerHTML = "<span>hello</span>"`.

But this change makes it quite easy for us to accept a node predicate function as a selector, so the language mode can still handle walking the trees and finding the smallest matching node, and extensions can pass in a function describing the node they want.

Updated to support and test that behavior.

queerviolet: On second though, it would be nice to be able to query for support of the new behavior, so I did add `getSyntaxNodeForRange` and `getSyntaxNodeForPosition`. (`bufferRangeForScopeAtPosition` also supports match functions)

It works for toggle-quotes.

queerviolet: (See https://github.com/atom/toggle-quotes/pull/58)

PR #17736, opened 2018-07-24
Multiple, custom line number gutters

smashwilson:
To complete atom/github#1512, I need to be able to add multiple Gutters to an editor, each of which has a custom labelling function. I'm emulating this effect:

screen shot 2018-07-24 at 9 41 31 am

I did make a bit of an effort to do this with the existing Gutter API. I was able to get the root DOM element of an added Gutter with `getElement()` and render a React portal into it, but I ran into problems duplicating the logic from `LineNumberGutterComponent` because I couldn't find a way to trigger re-renders of the React component when the parent Etch component re-rendered, so the two would quickly get out of sync.

---

This adds a `type` property to the options object accepted by `TextEditor::addGutter()`. If set to `'line-number'`, a `LineNumberGutterComponent` is created instead of a `CustomGutterComponent`, and a `labelFn` option may be supplied to customize the created line number labels.

---

##### Remaining

- [x] Decorations added with `TextEditor::decorateMarker(marker, {type: 'line-number', ...})` should only decorate the One True line-number gutter.
- [x] Support a `className` option in `TextEditor::addGutter()`. Add the provided CSS class to the gutter's root element. This will make it easier to style line-number gutters.
- [x] Support `onMouseDown` and `onMouseMove` callbacks in `TextEditor::addGutter()`.
- [x] Tests green.


smashwilson:
Progress:

```js
atom.commands.add('atom-workspace', 'me:add-line-number-gutters', () => {
  if (!editor) {
    return;
  }

  editor.addGutter({
    name: 'custom-1',
    type: 'line-number',
    priority: 2,
    labelFn: ({bufferRow}) => `1.${bufferRow}`
  })

  editor.addGutter({
    name: 'custom-2',
    type: 'line-number',
    priority: 3,
    labelFn: ({screenRow}) => `2.${screenRow}`
  })
})
```

screen shot 2018-07-24 at 10 11 57 am

The custom gutters scroll correctly and everything, too 🚀 

One interesting side-effect is that the `line-number` decorations from the git-diff package show up on _every_ line-number gutter. That's a bit odd, so I'll fix those to target the One True line-number gutter by default; you should still be able to go through `Gutter::decorateMarker()` to add `line-number` decorations to other line-number gutters.

smashwilson:
I'm a bit leery about touching code in TextEditorElement and friends on my own without an extra set of eyes for safety. Reviewers, when you get a moment, could you look for any obvious red flags for performance or correctness? API change feedback is welcome as well if you have any. 🙇

maxbrunsfeld (approved):
The code all looks good to me!

Would it be worth adding a test for the rendering side of things in `text-editor-component-spec.js`, where you add a gutter with a custom `labelFn`, and assert that the right labels show up, and that the line numbers and line tiles have the right widths?

daviwil (approved):
:+1: :zap:

smashwilson:
> Would it be worth adding a test for the rendering side of things in text-editor-component-spec.js, where you add a gutter with a custom labelFn, and assert that the right labels show up, and that the line numbers and line tiles have the right widths?

👍 I'll get this in place, then merge.

Thanks for the eyes 🙇

PR #17750, opened 2018-07-26
Fix errors causing tests to fail when run from the UI

maxbrunsfeld:
This fixes some errors and noise in the console that has been occurring for me when running the tests in the UI.

1. There was a test in `atom-environment-spec.js` that emitted an auto-update-related event in the main process. This caused exceptions to be thrown in all of your other Atom windows. I found that this code path is now covered by other tests, so I simply removed the offending test.
2. Our strategy for subscribing to multiple different ipc messages in `application-delegate.js` entailed adding many individual ipc listeners, resulting in frequent `EventEmitter leak detected` warnings. I've refactored this code to use a single event listener.

### DIFF
--- a/spec/atom-environment-spec.js
+++ b/spec/atom-environment-spec.js
@@ -746,29 +746,6 @@ describe('AtomEnvironment', () => {
     })
   })
 
-  describe('::updateAvailable(info) (called via IPC from browser process)', () => {
-    let subscription
-
-    afterEach(() => {
-      if (subscription) subscription.dispose()
-    })
-
-    it('invokes onUpdateAvailable listeners', async () => {
-      if (process.platform !== 'darwin') return // Test tied to electron autoUpdater, we use something else on Linux and Win32
-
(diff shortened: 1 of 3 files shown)


daviwil (approved):

daviwil:
Quick question about quick-resetting project paths: is there any project state that might not be cleared out if calling `reset()` with the existing paths results in basically a no-op?

maxbrunsfeld:
I decided to remove my changes regarding the project paths. This was an attempt to fix a problem that had a deeper cause: [a performance regression in libgit2](https://github.com/libgit2/libgit2/issues/4740) caused instating `GitRepository` objects to become much more expensive. I have now [downgraded `git-utils`](https://github.com/atom/atom/commit/12875c19395224da980485e7dc2581a60b57ef76) to avoid that performance regression, so it's no longer necessary to avoid this code path so carefully.

maxbrunsfeld:
The Circle CI failure is due to us no longer using Circle CI, and the VSTS windows failure is an unrelated flaky test.

PR #17752, opened 2018-07-27
Fix repository tests that fail when run in fork of atom/atom

jasonrudolph:
Prior to this change, two of the repository-related tests assumed that they were running in a local checkout of https://github.com/atom/atom, and the tests unintentionally failed if the local repository was a fork of atom/atom.

We first noticed this failure in https://circleci.com/gh/marcomorain/atom-1/29:

```
Project
  .onDidAddRepository()
    it invokes callback when a path is added and the path is the root of a repository
      Expected 'git@github.com:marcomorain/atom-1.git' to contain 'atom/atom'.
        at jasmine.Spec.it (/Users/distiller/project/spec/project-spec.js:1006:42)
    it invokes callback when a path is added and the path is subdirectory of a repository
      Expected 'git@github.com:marcomorain/atom-1.git' to contain 'atom/atom'.
        at jasmine.Spec.it (/Users/distiller/project/spec/project-spec.js:1017:42)
```

Instead of depending on the local checkout of the atom repository for these tests, this pull request updates the tests to use one of the fixture repositories. 😅

### DIFF
--- a/spec/fixtures/git/master.git/config
+++ b/spec/fixtures/git/master.git/config
@@ -4,3 +4,6 @@
 	bare = false
 	logallrefupdates = true
 	ignorecase = true
+[remote "origin"]
+	url = https://github.com/example-user/example-repo.git
+	fetch = +refs/heads/*:refs/remotes/origin/*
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (approved):
⚡️  Great solution!

PR #17761, opened 2018-07-30
Add missing x64 build artifacts to appveyor.yml

daviwil:
This change fixes an issue introduced by PR #17538 which changed how we name build artifacts on Windows.  Previously, our build scripts named AtomSetup.exe, atom-windows.zip, and RELEASES the same for x64 and x86 builds.  The aforementioned PR changed the build scripts to [add `x64` into these filenames](https://github.com/atom/atom/pull/17538/files#diff-131cff56e21060c243d2fd3d0339d39f) on x64 builds to match what we deliver as release assets.

This change updates `appveyor.yml` to ensure that the x64 artifacts are uploaded on successful AppVeyor builds.

/cc @jasonrudolph

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -77,6 +77,12 @@ artifacts:
     name: atom-windows.zip
   - path: out\RELEASES
     name: RELEASES
+  - path: out\AtomSetup-x64.exe
+    name: AtomSetup-x64.exe
+  - path: out\atom-x64-windows.zip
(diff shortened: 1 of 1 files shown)


daviwil:
Since artifacts aren't generated on PR builds, we've verified that this fix works on a test release branch: https://ci.appveyor.com/project/Atom/atom/build/10760

Issue #16786, opened 2018-02-19
Atom config reset


Arcanemagus:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

My Atom configuration was reset.

### Steps to Reproduce

1. Open Atom
2. Note that the configuration was reset

**Expected behavior:**

Config not to be reset.

**Actual behavior:**

Config was reset.

**Reproduces how often:**

At least once, but so far no more than that.

### Versions

OS: Windows 10 x64
```
Atom    : 1.25.0-beta0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.25.0-beta0
python 2.7.14
git 2.16.1.windows.4
visual studio 2015
```

### Additional Information

This _could_ be related to https://github.com/atom/atom/pull/16628, however I was also testing _many_ older versions of Atom for something unrelated and I'm not sure when this would have actually happened. Feel free to close it, I only filed this as a heads up that there _might_ be an issue!

raphinesse:
My settings have been reset to default probably 5 times over the last few months. The last time just now, using Atom 1.26.1 on Ubuntu 18.04.

rsese:
@raphinesse there have been additional potential fixes that landed as of 1.27 (e.g. https://github.com/atom/atom/pull/17166).  Can you update and see if that helps?

raphinesse:
Thanks for the info. I have updated and will report back if the issue re-occurs.

raphinesse:
Shortly after writing above comment, my config was reset again. This tie using 1.27. I wouldn't rule out a package wreaking havoc, but it's hard to confirm since the resets occur sporadically and I can't run in safe mode all the time :unamused:

joelhellman:
[...]

PR #17774, opened 2018-08-03
Write config file atomically

matthewwithanm:
### Description of the Change

We (the Nuclide team) continue to have reports of #17060 from Windows users, even after #17166. The cause isn't certain, however, we do know that [the `before-quit` event isn't reliable on Windows][1]. In any case, atomic writes (write + move) are less error-prone.

I rolled back the fix from #17060 to get back to a state where the issue could be reproduced somewhat reliably, then made the change an was unable to repro. This isn't a guaranteed fix (since the same could be said about #17060), but it seems more robust.

I was initially worried about the watcher not picking up the replace, but I tested that file updates were reflected in UI and that UI updates were mirrored to the file on both MacOS and Windows and we're good.

### Alternate Designs

None.

### Possible Drawbacks

* We're doing two operations (write and move) where we used to do one so this is going to be slower.
* There could be something subtle with the watcher I didn't pick up on when testing.
* I haven't verified on Linux.

### Verification Process

* For both MacOS and Windows:
    * Update `editor.fontSize` in my `config.cson` and see change in Atom.
    * Use `⌘+` to increase the font size and see change in `config.cson`.
* Install @Arcanemagus's settings-wiper from #17060 and repeat his instructions in that issue for wiping settings many (~50) times without settings loss.

### Applicable Issues

* #17166
* #17060
* #16786 
* #14909

cc @ebluestein

[1]: https://github.com/electron/electron/blob/master/docs/api/app.md#event-before-quit

### DIFF
--- a/src/config-file.js
+++ b/src/config-file.js
@@ -6,6 +6,7 @@ const {watchPath} = require('./path-watcher')
 const CSON = require('season')
 const Path = require('path')
 const async = require('async')
+const temp = require('temp')
 
 const EVENT_TYPES = new Set([
   'created',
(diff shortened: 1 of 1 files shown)


maxbrunsfeld (approved):
⚡️

Arcanemagus (approved):

maxbrunsfeld:
Thanks @matthewwithanm!

jasonrudolph:
FYI: We unfortunately ran into an issue (#18118) with this change after rolling it out in Atom 1.31.0 and 1.32.0-beta0. As a result, we had to revert this change in https://github.com/atom/atom/pull/18134.

PR #17779, opened 2018-08-03
Don't take destructive action on esc

matthewwithanm:
### Description of the Change

People instinctively hit escape to ignore alerts, so we shouldn't have it do something destructive.

### Alternate Designs

We could also switch the button order but I kept it the same in most regards.

### Possible Drawbacks

Maybe somebody memorized that hitting escape did the destructive thing? Seems unlikely.

### Verification Process

Did the following in the console:

```js
require('electron').remote.getCurrentWindow().emit('unresponsive')
```

```js
require('electron').remote.getCurrentWindow().emit('crash')
```

Hit escape after each and the window wasn't destroyed. Repeated, clicking the destructive buttons and it was.

### DIFF
--- a/src/main-process/atom-window.js
+++ b/src/main-process/atom-window.js
@@ -185,6 +185,7 @@ class AtomWindow extends EventEmitter {
       dialog.showMessageBox(this.browserWindow, {
         type: 'warning',
         buttons: ['Force Close', 'Keep Waiting'],
+        cancelId: 1, // Canceling should be the least destructive action
         message: 'Editor is not responding',
         detail:
           'The editor is not responding. Would you like to force close it or just keep waiting?'
@@ -202,6 +203,7 @@ class AtomWindow extends EventEmitter {
       dialog.showMessageBox(this.browserWindow, {
         type: 'warning',
         buttons: ['Close Window', 'Reload', 'Keep It Open'],
+        cancelId: 2, // Canceling should be the least destructive action
         message: 'The editor has crashed',
         detail: 'Please report this issue to https://github.com/atom/atom'
       }, response => {


maxbrunsfeld (approved):
👍 👍 👍

Ben3eeE:
@50Wliu Yeah I've mentioned this before. I don't remember if there is an issue for this.

This change will not work on Windows according to the api documentation. We need to change the label to be `Cancel` or `No` for it to work on Windows instead of using `cancelId`.

Adding a `Cancel` or `No` label will also make the dialog render buttons with the correct style on Windows. See https://github.com/atom/github/pull/1452 for pictures.

From the [electron api documentation](https://electronjs.org/docs/api/dialog):

> cancelId Integer (optional) - The index of the button to be used to cancel the dialog, via the Esc key. By default this is assigned to the first button with "cancel" or "no" as the label. If no such labeled buttons exist and this option is not set, 0 will be used as the return value or callback response. **This option is ignored on Windows**.

50Wliu:
It might actually work, according to https://github.com/electron/electron/pull/13882.

daviwil:
Latest VSTS build for this PR is green despite what the check says (known VSTS bug where PR rebuilds don't update PR status).  AppVeyor is clean too so I'm merging this.  Thanks a bunch @matthewwithanm!

daviwil:
Actually, I just read @Ben3eeE's comment while doing one last pass over everything and it looks like this change might not resolve the issue on Windows?  Is it worth trying to change these button labels so that the behavior is also correct on Windows?

Ben3eeE:
@daviwil It seems that `cancelId` does work on Windows since the PR to update the electron docs was merged https://github.com/electron/electron/pull/13882.

daviwil:
Great, thanks for verifying @Ben3eeE!  Merging this then :)

PR #17781, opened 2018-08-05
Allow comma-separated selectors as keys in Tree-sitter scope mappings

maxbrunsfeld:
Now that scope map values have a bit more structure (see https://github.com/atom/atom/pull/17738), it's nice to be able to avoid duplicating these values when they are used with multiple selectors. This PR allows the keys of a scope map to contain comma-separated selectors, similar to CSS.


sakdaBoodprom:
ok

sakda0533@gmail.com

PR #17445, opened 2018-06-01
Add proposed RFC "Atom Nightly Releases"

daviwil:
This RFC proposes that Atom add a third official release channel which delivers new builds of Atom nightly from the `master` branch.  Nightly releases will allow new improvements to reach users long before a new Stable or Beta release is shipped.  This effort will also give us the opportunity to experiment with new release automation strategies that could eventually be used to speed up the Stable and Beta release cadence.

[**Rendered view**](https://github.com/atom/atom/blob/dw-nightly-releases-rfc/docs/rfcs/002-atom-nightly-releases.md)

PR #17788, opened 2018-08-06
Remove CircleCI configuration

jasonrudolph:
**tl;dr** - Now that we're using VSTS for macOS builds, this pull request removes the CircleCI 1.0 build configuration before it is [EOL at the end of the month](https://circleci.com/blog/sunsetting-1-0/).

---

Atom has historically used CircleCI for macOS builds. CircleCI [announced](https://circleci.com/blog/sunsetting-1-0/) that CircleCI 1.0 builds will no longer be available after August 31, 2018. That means that we either need to migrate to CircleCI 2.0 or find another solution for Atom's macOS builds.

Given that we recently implemented cross-platform builds (macOS, Linux, and Windows) on VSTS for Atom Nightly Releases (https://github.com/atom/atom/pull/17445, https://github.com/atom/atom/pull/17538), and given that we don't yet have a working CircleCI 2.0 configuration (https://github.com/atom/atom/pull/17445), it feels like the pragmatic choice is to use the already-working VSTS builds as our definitive macOS CI solution, and remove our soon-to-be-EOL CircleCI configuration.


daviwil (approved):
:tada:

PR #17795, opened 2018-08-07
Atom from WSL respects windows drive mount location defined in /etc/wsl.conf

### LINKED ISSUE #17794: Atom in WSL does not respect changing the windows drive root mount point
### Description

If the mount point for windows drives in WSL is changed via editing /etc/wsl.conf, detailed [here](https://blogs.msdn.microsoft.com/commandline/2018/02/07/automatically-configuring-wsl/), atom no longer works to open windows directories/files that it once opened without issue.

### Steps to Reproduce

1. Change WSL default windows drive mount points to anything other than /mnt/%drive_letter%. 
This is done in the /etc/wsl.conf file, specified [here](https://blogs.msdn.microsoft.com/commandline/2018/02/07/automatically-configuring-wsl/)
2. Navigate to a directory on the windows filesystem inside WSL. Eg /c/dev/
3. Run "atom ."

**Expected behavior:** Atom to open using the current directory

**Actual behavior:** Command line outputs "The system cannot find the path specified."

### Versions

Windows 10 - 18.03 + WSL Ubuntu 18.04

### Additional Information

Contents of /etc/wsl.conf are as follows:

> ```
> [network]
> generateResolvConf = false
> 
> [automount]
> root = /
> options = "metadata"
> ```

Jinivus:
### Description of the Change

Allows the default atom shell script to use the WSL windows drive mount point defined in /etc/wsl.conf

The current core supports WSL except for cases when the windows drive mount point is not default

### Benefits

Any wsl.conf configured windows drive mount point will work instead of just the default

### Possible Drawbacks

Potentially doesn't account for every edge case mount point?

### Verification Process

1. Change WSL default windows drive mount points to anything other than /mnt/%drive_letter%.
This is done in the /etc/wsl.conf file, specified here
3. Navigate to a directory on the windows filesystem inside WSL. Eg /c/dev/
4. Run "atom ."

### Applicable Issues

#17794

### DIFF
--- a/resources/win/atom.sh
+++ b/resources/win/atom.sh
@@ -7,6 +7,10 @@ else
   pushd "$(dirname "$0")" > /dev/null
   if [[ $(uname -r) == *-Microsoft ]]; then
     # We are in Windows Subsystem for Linux, map /mnt/drive
+    root="/mnt/"
+    # If different root mount point defined in /etc/wsl.conf, use that instead
+    eval $(grep "^root" /etc/wsl.conf | sed -e "s/ //g")
+    root="$(echo $root | sed 's|/|\\/|g')"
     ATOMCMD="$(echo $PWD | sed 's/\/mnt\/\([a-z]*\)\(.*\)/\1:\2/')/atom.cmd"
     ATOMCMD="${ATOMCMD////\\}"
   else


daviwil:
I restarted the VSTS build and it succeeded despite what the PR build status says:

https://github.visualstudio.com/Atom/_build/results?buildId=4686&_a=summary&view=logs

daviwil:
Thanks @Jinivus!

PR #17803, opened 2018-08-08
Use npm and apm ci in CI builds

smashwilson:
I've added a `--ci` argument to `script/build`. When specified, invocations of npm and apm use the "ci" command rather than "install".

* [`npm ci` documentation](https://docs.npmjs.com/cli/ci)
* ["Introducing npm ci" on the npm blog](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable)

I doubt this will substantially improve our build times; we're shaving time off of a step that isn't taking the bulk of our build time. The most notably benefit will be _reliability_: by installing the package versions pinned in the lockfile exactly, patch releases of transitive dependencies should no longer implicitly creep in and break our build.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -17,6 +17,9 @@ process.on('unhandledRejection', function (e) {
   process.exit(1)
 })
 
+// We can't use yargs until installScriptDependencies() is executed, so...
+const ci = process.argv.indexOf('--ci') !== -1
+
 verifyMachineRequirements()
 
 if (dependenciesFingerprint.isOutdated()) {
@@ -25,13 +28,13 @@ if (dependenciesFingerprint.isOutdated()) {
 
 if (process.platform === 'win32') deleteMsbuildFromPath()
 
-installScriptDependencies()
-installApm()
+installScriptDependencies(ci)
+installApm(ci)
 childProcess.execFileSync(
   CONFIG.getApmBinPath(),
   ['--version'],
   {stdio: 'inherit'}
 )
-runApmInstall(CONFIG.repositoryRootPath)
+runApmInstall(CONFIG.repositoryRootPath, ci)
 
 dependenciesFingerprint.write()
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -18,7 +18,12 @@ process.on('unhandledRejection', function (e) {
 })
 
 // We can't use yargs until installScriptDependencies() is executed, so...
-const ci = process.argv.indexOf('--ci') !== -1
+let ci = process.argv.indexOf('--ci') !== -1
+
+if (!ci && process.env.CI === 'true' && process.argv.indexOf('--no-ci') === -1) {
+  console.log('Automatically enabling --ci because CI is set in the environment')
+  ci = true
+}
 
 verifyMachineRequirements()
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
   - nvm install $NODE_VERSION
   - nvm use --delete-prefix $NODE_VERSION
   - npm install --global npm@6.2.0
-  - script/build --ci --create-debian-package --create-rpm-package --compress-artifacts
+  - script/build --create-debian-package --create-rpm-package --compress-artifacts
 
 script:
   - script/lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - source /tmp/.nvm/nvm.sh
   - nvm install $NODE_VERSION
   - nvm use --delete-prefix $NODE_VERSION
-  - npm install -g npm@6.1.0
+  - npm install --global npm@6.2.0
   - script/build --ci --create-debian-package --create-rpm-package --compress-artifacts
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
   - nvm install $NODE_VERSION
   - nvm use --delete-prefix $NODE_VERSION
   - npm install -g npm@6.1.0
-  - script/build --create-debian-package --create-rpm-package --compress-artifacts
+  - script/build --ci --create-debian-package --create-rpm-package --compress-artifacts
 
 script:
   - script/lint
--- a/apm/package-lock.json
+++ b/apm/package-lock.json
@@ -21,7 +21,6 @@
         "node-gyp": "3.4.0",
         "npm": "6.2.0",
         "open": "0.0.5",
-        "plist": "git+https://github.com/nathansobo/node-plist.git#bd3a93387f1d4b2cff819b200870d35465796e77",
         "q": "~0.9.7",
         "read": "~1.0.5",
         "request": "^2.87.0",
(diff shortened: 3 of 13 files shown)


script/bootstrap
```diff
@@ -17,6 +17,9 @@ process.on('unhandledRejection', function (e) {
   process.exit(1)
 })
 
+// We can't use yargs until installScriptDependencies() is executed, so...
+const ci = process.argv.indexOf('--ci') !== -1
```
daviwil: One alternative here is to look at `process.env.CI` and see if it's set to `true`, that'd save you from dealing with the command line parameter (and subsequenct .yml changes).  It's set by default on [AppVeyor](https://www.appveyor.com/docs/environment-variables/) and [Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) and I've added it to the builds on VSTS too.

daviwil (approved):
Aside from the possible use of `process.env.CI`, it looks great!

daviwil:
Lockfiles are definitely a pain, I've been scratching my head every day recently and worrying about whether I'm going to break everything by committing changes.  No harm in trying it for a while though until we get used to it and understand what to expect.

I had also thought about the auto-PR of new `package-lock.json` files but I think it'll be harder for us to actually keep up with those since they'll likely be coming every day.  I think the only viable solution is to just start making lockfile change commits an obvious and expected part of our workflow so that we're more aware of when it's happening.  But that also means that **we can't just bump versions in `package.json` anymore**, we'll have to run `npm/apm install` before committing them so that `package-lock.json` gets updated.

My biggest problem with the lockfiles changing is that I'm not sure I trust whether it's actually due to dependencies changing or `npm` doing funny business.  More than once I've seen `npm` switch the `package-lock.json` dependencies back and forth between explicit versions and version ranges for seemingly no good reason.

Regardless of any solution, I think we'll have to be vigilant at first and make sure to really scrutinize the changes so that we don't end up with unexpected problems in release builds.

smashwilson:
> But that also means that we can't just bump versions in package.json anymore, we'll have to run npm/apm install before committing them so that package-lock.json gets updated.

Yeah. Would it be valuable to have an alternative script that runs `npm install --package-lock-only` to update the lockfiles without doing a full build? Or would we want to make sure we do a full `script/build` before committing dependency changes?

> My biggest problem with the lockfiles changing is that I'm not sure I trust whether it's actually due to dependencies changing or npm doing funny business. More than once I've seen npm switch the package-lock.json dependencies back and forth between explicit versions and version ranges for seemingly no good reason.

True... I've seen that too. For a while there were problems with platform-specific optional dependencies flapping in and out of the lockfile and some ordering inconsistencies. In general I think npm is trying to make lockfile generation deterministic but there are definitely some cases where it isn't quite.

I think the ideal path forward here would be to increase our faith in our CI catching the kinds of breakage that we want to avoid, so that a green check gives us a strong assurance that an update is acceptable (and a red X gives us a signal that we need to tighten a dependency range or send a PR upstream).

Specifically, the greatest source of uncaught breakage that I've seen is when something breaks electron-link or snapshotting. Our transitive dependencies don't test or consider these because neither are common, so it's easy for a patch-version change upstream to break our build. React 16.4.2 did this: a change in the way their source was minimized breaks electron-link! A while ago we had a patch version of Relay that broke, too (but that would have been caught by our regular test suite, I believe - this was before we had any test coverage for our Relay code).

> Aside from the possible use of `process.env.CI`, it looks great!

Ah, interesting. Can do 👍

smashwilson:
Okay, this is weird. After running `npm ci --global-style` in apm, there are a bunch of unmet dependencies:

```
smashwilson @ hubtop ~/src/atom/atom/apm (aw/build-ci>) [1]
$ npm install --global-style
npm WARN atom-bundled-apm@ No license field.

added 339 packages from 160 contributors, removed 238 packages, updated 1 package and audited 6044 packages in 8.27s
found 21 vulnerabilities (20 moderate, 1 critical)
  run `npm audit fix` to fix them, or `npm audit` for details
smashwilson @ hubtop ~/src/atom/atom/apm (aw/build-ci>)
$ npm ls pacote
atom-bundled-apm@ /Users/smashwilson/src/atom/atom/apm
└─┬ atom-package-manager@2.1.0
  └─┬ npm@6.2.0
    ├─┬ libcipm@2.0.0
    │ └── pacote@8.1.6  deduped
    └── pacote@8.1.6
$ npm ci --global-style
# ...
$ npm ls pacote
smashwilson @ hubtop ~/src/atom/atom/apm (aw/build-ci>)
$ npm ls pacote
atom-bundled-apm@ /Users/smashwilson/src/atom/atom/apm
└─┬ atom-package-manager@2.1.0
  └─┬ npm@6.2.0
    ├─┬ UNMET DEPENDENCY libcipm@2.0.0
    │ └── UNMET DEPENDENCY pacote@8.1.6
    └── UNMET DEPENDENCY pacote@8.1.6

npm ERR! missing: libcipm@2.0.0, required by npm@6.2.0
npm ERR! missing: pacote@8.1.6, required by npm@6.2.0
npm ERR! missing: pacote@8.1.6, required by libcipm@2.0.0
```

My current guess is that apm's [postinstall script](https://github.com/atom/apm/blob/master/script/postinstall.sh) is borking up npm:

```sh
echo ">> Downloading bundled Node"
node script/download-node.js


echo
echo ">> Rebuilding apm dependencies with bundled Node $(./bin/node -p "process.version + ' ' + process.arch")"
./bin/npm rebuild


echo
echo ">> Deduping apm dependencies"
./bin/npm dedupe
```

script/bootstrap
smashwilson: Done 👍  I have it logging a message and allowing an override with `--no-ci` just in case 👌

daviwil:
> Would it be valuable to have an alternative script that runs npm install --package-lock-only to update the lockfiles without doing a full build? Or would we want to make sure we do a full script/build before committing dependency changes?

One of those could be useful, and it might be even better if we can include a pre-commit hook check to see if `package.json` is newer than `package-lock.json` (or if `package-lock.json` is modified) and prompt the user to run `npm install --package-lock-only` before committing (or do it for them?)

> I think the ideal path forward here would be to increase our faith in our CI catching the kinds of breakage that we want to avoid, so that a green check gives us a strong assurance that an update is acceptable (and a red X gives us a signal that we need to tighten a dependency range or send a PR upstream).

:heavy_check_mark: 

Regarding the `apm dedupe` issue, is it possible that the `package-lock.json` files are once again out of sync and causing the dependencies to not be installed?

smashwilson:
> One of those could be useful, and it might be even better if we can include a pre-commit hook check to see if `package.json` is newer than `package-lock.json` (or if `package-lock.json` is modified) and prompt the user to run npm install --package-lock-only before committing (or do it for them?)

Oh, neat, I like those ideas. git doesn't make it easy to distribute hooks, though. 🤔 

> Regarding the `apm dedupe` issue, is it possible that the `package-lock.json` files are once again out of sync and causing the dependencies to not be installed?

No, it shouldn't be... `npm ci` refuses to run if they aren't in sync.

Basically, the apm package includes a postinstall script that calls `npm dedupe`, presumably to deal with our path-too-long issues. atom/atom's build scripts install apm with the `--global-style` flag set so that apm's dependencies will be laid out beneath `apm/node_modules/atom-package-manager/` and other scripts will work. Somehow, the combination of these two causes `npm ci` to omit certain apm dependencies, or install them to unexpected places, so that require calls would break at runtime. I couldn't figure out exactly what caused it, though.

daviwil:
CI is :green_apple:, looks good to go!

PR #17829, opened 2018-08-13
Add todo-list as a deprecated package

tobiasvl:
### Description of the Change

This marks the `todo-list` package as deprecated.

In addition to using deprecated API functions, the package is in fact no longer maintained, [according to its README](https://github.com/AnthonyCalandra/todo-list/blob/master/README.md).

Note that this notice was added to the README without a new package version release; the notice is not visible from the "Install Packages" view in Atom.

### Alternate Designs

* [This package does use deprecated API functions.](https://github.com/AnthonyCalandra/todo-list/issues/10) It's not clear to me why it has not been marked as deprecated already.

* I looked for an alternate way to deprecate packages, and didn't find any – not even a way that the package maintainer himself could have used. In this case, he has chosen the alternate design of writing it in the repository's README file, but he has not published a new version of the package, so this deprecation notice does not actually display in Atom.)

* [There is precedence for using this file to manually deprecate packages.](https://discuss.atom.io/t/how-to-mark-a-package-as-deprecated/34745/5)

Therefore, I propose this way to deprecate this package.

I chose what seems to be the most popular similar package as the alternative. Feel free to remove this suggestion or suggest another alternative package.

There is no alternate, official way to mark a package as deprecated. There should be. Also, this package uses deprecated API functions.

### Benefits

The 5000 people who have installed the non-functioning package, as well as all the new people (judging by the issues) who try to install it without realizing it's deprecated in the newest unpublished version, will not get an error message after installing.

### Possible Drawbacks

None that I can see.

### Verification Process

None

### Applicable Issues

* AnthonyCalandra/todo-list#9
* AnthonyCalandra/todo-list#10
* AnthonyCalandra/todo-list#14
* AnthonyCalandra/todo-list#15
* AnthonyCalandra/todo-list#18

### DIFF
--- a/script/deprecated-packages.json
+++ b/script/deprecated-packages.json
@@ -1561,6 +1561,11 @@
     "hasDeprecations": true,
     "latestHasDeprecations": true
   },
+  "todo-list": {
+    "hasAlternative": true,
+    "message": "This package is no longer maintained.",
+    "alternative": "todo-show"
+  },
   "toggle-tabs": {
     "version": "<=0.1.8",
     "hasDeprecations": true,


lee-dohm:
Thanks for the suggestion.

The "deprecated packages" list was a workaround for packages that were released prior to version 1.0 of Atom, when there was a large amount of churn in the API. You are correct that we don't have a mechanism for marking packages as deprecated. Community package maintainers are free to use whatever system works for them. We don't intend to maintain this list going forward. Because of this, we're going to close this PR.

Thanks again for the suggestion.

tobiasvl:
Understandable, but I hope there will be an official mechanism for visibly deprecating packages in the future.

PR #17862, opened 2018-08-14
Enable stable and beta draft release creation on VSTS

daviwil:
### Description of the Change

This PR adds logic to our VSTS release build definition to enable the creation of draft GitHub releases when successful builds are produced on release branches (those named `1.xx-releases`).

**NOTE:** This change will not be turned on by default until 32-bit Windows builds are being produced on VSTS.  This PR just adds the necessary release automation so that it's ready once those builds are being produced.

This is a build script for Atom.

### Benefits

Draft releases are now created automatically for every successful build on a release branch.  This removes the need for an Atom developer to go manually create a draft release with a separate internal tool.

This change also sets the stage for performing the final publishing steps with VSTS in an automated way.

### Possible Drawbacks

None that I can think of.

### Verification Process

- [x] Verify that a draft release successfully gets created from a simulated release build

This is a screenshot of a draft release created as a result of [this build](https://github.visualstudio.com/Atom/_build/results?buildId=4687&_a=summary&view=logs):

image

### Applicable Issues

None.

### TODO

- [x] Add changelog generation and save changelog from previously published release with the same version


daviwil:
The work is almost done here, just need to test with a few more simulated release builds before merging!

daviwil:
Merging this despite unrelated test flakes, planning to test-drive this for the upcoming release.

PR #17866, opened 2018-08-15
Better auto-indent for TextMate grammars

### LINKED ISSUE #17364: Wrong auto-indent behavior
Reposting #10973 and #8570 because they've been locked.

This is still a bug. Copy-paste the java example below and try typing something on the `else if` line. The indentation will change on its own:

```java
class MyClass {
    public void myMethod() {
        if (true) {

        // A comment
        } else if (/* type something here */) {

        }
    }
}
```

Aerijo:
### Description of the Change

Makes auto-indent skip over lines that start with a comment, so they do not interfere.

### Alternate Designs

I considered going through the tags of each line manually, which would allow better handling of the following:
```
stuff
           /* comment */ if (true) {
}          }
|          ^ we technically want it here
^ it will actually go here
```
This happens because the line we wanted to align with starts with a comment, so we skip over it with the current implementation. However, checking tokens would be a lot more work, and we don't get the caching for free like we do with `isCommented()`. Plus, Tree-sitter can give us the corresponding line directly, so this is only needed while TextMate grammars are still widely used.

Is core mechanic

### Benefits

Fixes bad auto-indenting with things like
```
class MyClass {
    public void myMethod() {
        if (true) {

        // A comment <- originally used as line to align with; now ignored
        } else if (/* type something here */) {

        }
    }
}
```

### Possible Drawbacks

The check is _slightly_ slower when many comments are present, but only for the very first run. Subsequent runs cache whether lines are commented, so performance is nearly identical to the original. (benchmarks to be added)

### Verification Process

Manually (tests to be added?) with above example code, and file with 10 000 consecutive comment lines above an auto-indenting line.

### Applicable Issues

Fixes #17364

### DIFF
--- a/benchmarks/text-editor-autoindent.bench.js
+++ b/benchmarks/text-editor-autoindent.bench.js
@@ -1,0 +1,48 @@
+const {TextEditor, TextBuffer} = require('atom')
+module.exports = async ({test}) => {
+  const data = []
+  const workspaceElement = atom.workspace.getElement()
+  document.body.appendChild(workspaceElement)
+  atom.packages.loadPackages()
+  await atom.packages.activate()
+  console.log(atom.getLoadSettings().resourcePath);
+  for (let pane of atom.workspace.getPanes()) {
+    pane.destroy()
+  }
+  const stepSize = 1000
+  const minLines = 10
+  const maxLines = 10 * stepSize + minLines
+  const commentBodyLine = "Lorem ipsum dolor sit amet\n"
+  const commentBody = commentBodyLine.repeat(maxLines)
+  const commentStart = "/*\n"
+  const commentEnd = "*/\n"
+  const startContext = "switch (x) { case 0: ".repeat(12) + "\n"
+  const endContext = "}\n" + "}".repeat(11)
+  for (let numLines = maxLines; numLines >= minLines; numLines -= stepSize) {
+    let testCommentBody = commentBody.slice(0, numLines * commentBodyLine.length)
+    let comment = commentStart + testCommentBody + commentEnd
+    let text = startContext + comment + endContext
+    const buffer = new TextBuffer({text})
+    const editor = new TextEditor({buffer, autoHeight: false, largeFileMode: true})
+    atom.grammars.assignLanguageMode(buffer, "source.js")
+    atom.workspace.getActivePane().activateItem(editor)
+    let indentRow = editor.getLineCount() - 2
+    editor.setSelectedBufferRange([[indentRow, 0], [indentRow, Infinity]])
+    let t0 = window.performance.now()
+    editor.autoIndentSelectedRows()
+    let t1 = window.performance.now()
+    data.push({
+      name: 'Auto indenting with a lot of commented lines above',
+      x: numLines,
+      duration: t1 - t0
+    })
+    editor.destroy()
+    buffer.destroy()
+    await timeout(2000)
+  }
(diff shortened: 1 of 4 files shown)


Aerijo:
The last commit puts the changes in a single method, which is then used in the original location & in the other autoindent place. This means the fix also applies to `Editor: Auto Indent`.

On a different note, how does one write benchmarks 😢?

Ben3eeE:
> On a different note, how does one write benchmarks?

The PR that added benchmarks has some details that might help get you started https://github.com/atom/atom/pull/12984

Aerijo:
Well, that was my attempt at a benchmark. It seems that earlier rounds are interferring with later ones, so I iterate backwards (most comments are tested first). I only test block comments here, inside of 12 nested switch statements, because checking for specific open scopes when there are many is the most likely performance cost (switch statements were chosen because their contents are scoped). 

Checking if it's a line comment is comparatively much faster, because we only look at the first one or two elements in the tokenised line tags.

maxbrunsfeld:
> Turns out this breaks Tree-Sitter grammars. They currently just copy indent rules from TextMate for now, so I don't think this can be merged until that's sorted.

It looks like you'd need to add your new method `getPrecedingIndentRow` to this list of methods to copy. You'd also need to avoid using `tokenizedLineForRow`, since that's a textmate-specific API. Fortunately, I think you could just replace this:

```js
tokenizedLine = this.tokenizedLineForRow(activeRow)
if (!tokenizedLine.isComment() && /\S/.test(tokenizedLine.text)) {
  // ...
}
```

with this:

```js
if (!this.isRowCommented(bufferRow) && !this.buffer.isRowBlank(bufferRow)) {
  // ...
}
```

It might be worth running your benchmark with Tree-sitter enabled to make sure everything's ok performance-wise.

noname2534:
Good job

Aerijo:
Update: Tree-sitter leads to very poor performance here. I revised it to use the more 'sophisticated' behaviour for the first 1000 lines, but revert to the original if that is exceeded. This will cover 99% of reasonable use cases, including the original issue. 1000 is arbitrary, but felt smooth enough when testing with > 1000 lines of 50 spaces above the search start.

Tangentially; maybe we should consider restricting indentation checks to special characters? It currently uses a regex, but that regex is for the whole line. This would be in addition / an alternative to that, up to where the cursor is. E.g. `\s*}$` would match if the cursor just typed a `}` at the start of the line. This would eliminate any lag entirely while typing on the rest of the line AND prevent the jumping while typing that I'm not convinced isn't a bug itself.

~~I also deleted my original fork with these changes, so I'll be submitting to a new PR.~~ Or, renaming the branch works too.

sadick254:
Thanks very much for reaching out!

As you may have heard, we are migrating from our old [first-mate grammar engine](https://github.com/atom/first-mate) to the [new Tree-sitter engine](https://blog.github.com/2018-10-31-atoms-new-parsing-system/). This will enable a number of new features, more consistent syntax highlighting, and better performance, among other benefits. In order to free up our limited resources, we have decided to stop maintaining the first-mate grammar when there is a built-in Tree-sitter grammar available. Since this issue applies only to the first-mate grammar, we are going to close it. Please let us know if we have misunderstood the description of the problem.

We are also planning to leverage tree-sitter for indentation. See https://github.com/atom/atom/pull/18321

Thanks again for your interest in Atom.

PR #17873, opened 2018-08-16
During quit, close unloaded windows

stepanhruda:
### Description of the Change

When quitting the app (e.g. by Cmd-Q on macOS), in `before-quit` Atom asks each open window to unload. Any one of them can resolve the promise with `false` (e.g. because user realized they didn't want to quit and hit Cancel on unsaved changes dialog), which stops the app from quitting.

When some of the windows end up unloading successfully, but the quit is stopped this way, these windows stay open, left in a weird unloaded state (Atom expects them to ultimately get closed by a quit, which doesn't happen). This adds logic to immediately close windows after they are successfully unloaded during a quit.

### Alternate Designs

We could close the window in `prepareToUnload()`? We'd need to check that this makes sense for all callsites.

It's a bugfix.

### Benefits

1. No unloaded windows after an interrupted quit.
2. Windows get closed as soon as they are unloaded - even during a successful quit, you could see the unloaded UI if one window's unload is significantly slower than others'.

### Possible Drawbacks

State restoration on other platforms? Everything worked fine in my testing on macOS.

### Verification Process

1. Added a test.
2. Had 2 windows, hit Cmd-Q. Window A returned `false` from `prepareToUnload()`. Window B still closed.

### Applicable Issues

Don't know.

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -606,6 +606,8 @@ describe('AtomApplication', function () {
     assert(!electron.app.didQuit())
 
     await Promise.all([window1.lastPrepareToUnloadPromise, window2.lastPrepareToUnloadPromise])
+    assert(!electron.app.didQuit())
+    await atomApplication.lastBeforeQuitPromise
     await new Promise(process.nextTick)
     assert(electron.app.didQuit())
   })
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (approved):
This looks great. Thanks for fixing this!

daviwil:
I noticed this test failure on VSTS:

```
2018-08-16T20:19:10.6236474Z   1) AtomApplication prevents quitting if user cancels when prompted to save an item:
2018-08-16T20:19:10.6236742Z      Error: timeout of 60000ms exceeded. Ensure the done() callback is being called in this test.
2018-08-16T20:19:10.6237013Z       at Timeout.<anonymous> (D:\a\1\s\node_modules\mocha\lib\runnable.js:226:19)
```

Not sure if this is caused by the change or a test flake so I've restarted the build to see if it goes green then next time.

stepanhruda:
I have only been testing on macOS, it's very possible that Windows behaves slightly differently 😡 I might need to debug this in a Windows environment

stepanhruda:
Can I ask why this was reverted?

rsese:
David is OoO right now but it looks like it was reverted because it was causing failures on AppVeyor, not sure if there was more to it @daviwil?

Issue #17859, opened 2018-08-14
CSS classes do not match reported cursor scopes with Tree-Sitter parser


robertrossmann:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

With Tree-Sitter parser enabled, the CSS classes attached to a DOM node representing a piece of JavaScript code do not match the scopes listed via the `Log Cursor Scope` command. As per the [Flight Manual](https://flight-manual.atom.io/behind-atom/sections/scoped-settings-scopes-and-scope-descriptors/#scope-names-in-syntax-tokens), they should be identical:

> Scope names work just like CSS classes. In fact, in the editor, scope names are attached to a token's DOM node as CSS classes.

### Original intent

I wanted to customise the look of a particular syntax structure using CSS. I expected that if I write the following CSS override it would match the intended parts of the code with the new look:

```css
.object .pair .property-identifier {
  // custom CSS here
}
```

However, the snippet did nothing, which lead me to realise that the DOM node does not have this class.

### Steps to Reproduce

1. Enable Tree-Sitter parsers
1. Open JavaScript file
1. Place the cursor on an object literal's property identifier
1. Open DevTools and inspect that element
1. Log the cursor scope (Command Pallette -> Log Cursor Scope)
1. Observe difference in the scopes listed in the notification and the scopes listed on the DOM node

**Expected behavior:**

The DOM node would have the same classes attached as the scopes shown in the notification

**Actual behavior:**

The DOM node has CSS classes which do not correspond to any of the listed scopes

**Reproduces how often:**

All the time in JavaScript files with Tree-Sitter enabled

### Versions

```
Atom    : 1.31.0-nightly7
Electron: 2.0.7
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

```
apm  2.1.0
npm  6.2.0
node 8.9.3 x64
atom 1.31.0-nightly7
```

### Additional Information

[...]

Issue #17820, opened 2018-08-10
data-grammar change causes some keybindings to not work


Ben3eeE:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

With tree-sitter the data-grammar attribute changed so some keybindings don't work.

This change should either be documented in the flight manual or be backwards compatible!

### Steps to Reproduce

1. Open your keymap file
1. Add the example from the flight manual https://flight-manual.atom.io/behind-atom/sections/keymaps-in-depth/#selectors-and-custom-packages
1. I modified it slightly to:
```cson
"atom-text-editor[data-grammar='source js']":
  'ctrl-alt-g': 'editor:log-cursor-scope'
```

**Expected behavior:** The keybinding to work regardless of tree-sitter being enabled or not.

**Actual behavior:** The keybinding works with text mate grammars but with tree-sitter you need to change it to:
```cson
"atom-text-editor[data-grammar='javascript']":
  'ctrl-alt-g': 'editor:log-cursor-scope'
```

**Reproduces how often:** 100%

### Versions

1.31.0-nightly6

### Additional Information

/cc: @queerviolet @maxbrunsfeld

PR #17881, opened 2018-08-20
Support a single scope name field on Tree-sitter grammars

maxbrunsfeld:
### Background

When we originally [introduced](https://github.com/atom/atom/pull/16299) Tree-sitter parsers into Atom, we gave them entirely new root scope names, which we called *ids* instead of scope names. The plan was to make scope descriptors work differently with the new parsers, containing entries for each node in the syntax tree.

Since then, we've received a lot of issues about the new scope names causing problems. A lot of packages use scope names to control activation. We've also realized that we should be able to simply replace TextMate parsers with Tree-sitter parsers with the *same* scope name without introducing any breaking changes.

### Change

This PR remove the mechanisms for Tree-sitter grammars to track their correspondence to TextMate grammars while keeping separate scope names. Instead, we're just going to use the same scope names. I believe this will result in less overall confusion for users when we turn on Tree-sitter by default.

Fixes https://github.com/atom/atom/issues/17820
Fixes https://github.com/atom/atom/issues/17859


t9md:
I found another scope incompatibility.

- `editor.bufferRangeForScopeAtPosition()` takes scope specifier as string, but behavior is not consistent.

Can I expect this to be fixed too?

When cursor is at `|` char in following text.

```javascript
let a, b /* 2nd |var */, c
```

In both TextMateParser and TreeSitterParser, command `Editor: Log Cursor Scope`(`alt-cmd-p`) shows `source.js` and `comment.block.js`.
But when I tried to use `.comment.block` as scope for `bufferRangeForScopeAtPosition` to retrieve block comment range, TreeSitterParser return nothing whereas TextMateParser return block-comment range.
Seemingly TreeSitter expect different scope string specifier.

```javascript
const range = editor.bufferRangeForScopeAtPosition(".comment.block", editor.getCursorBufferPosition())
if (range) {
  console.log(editor.getTextInBufferRange(range))
}
// -> In TextMateParser: output "/* 2nd |var */"
// -> In TreeSitterParser: no ouptut
```

Some of the vim-mode-plus's function depends on this behavior.

maxbrunsfeld:
Thanks for pointing that out, @t9md. I will fix this.

t9md:
Thanks for the quick response as always!
I still have several issues on vim-mode-plus when switching from TextMate parser to TreeSitter parser.
Mostly it is related to scopes compatibility and tokenized buffer. Am I OK to ask about it here?

t9md:
@maxbrunsfeld Could you check and give comment on the following issue?
I can't find a good solution for now.
https://github.com/t9md/atom-vim-mode-plus/issues/1095

PR #17892, opened 2018-08-21
Improve repo-local packages implementation

daviwil:
### Description of the Change

This PR is a continuation of PR #17686 to address some feedback about how we detect the dev resource path in main.js.  It streamlines the dev resource path detection process so that it's only performed once, cutting the number of synchronous file system checks in half.

Also included is a removal of the separate repo-local package installation step in `script/bootstrap`.  It turns out that the more recent `npm` version used in `apm` causes a symbolic link to be created to the local package path, resulting in the package dependencies being installed simultaneously in the repo-local package path and Atom's `node_modules`.

### Benefits

- No unnecessary synchronous filesystem checks on startup
- No unnecessary consolidated core package installations

### Verification Process

- [x] `atom --dev .` uses `ATOM_DEV_RESOURCE_PATH` if one is set
- [x] `atom --dev .` in Atom repo path uses current directory as dev resource path when there is no `ATOM_DEV_RESOURCE_PATH` set and the current directory appears to be a clone of `atom/atom`
- [x] `atom --dev .` uses `~/github/atom` as the dev resource path if it exists and none of the path sources above are present
- [x] `atom --dev .` uses its bundled resource path when none of the path sources above are present
- [x] `atom -r path/to/atom` overrides all other possible resource path sources

### Applicable Issues

https://github.com/atom/atom/pull/17686


maxbrunsfeld (approved):
Awesome!

PR #17899, opened 2018-08-22
Default to @atom/watcher

smashwilson:
"What's the worst that could happen?"


smashwilson:
Hmmm:

```
Project
2018-08-22T14:39:44.6743900Z   .onDidChangeFiles()
2018-08-22T14:39:44.6758610Z     it reports filesystem changes within project paths
2018-08-22T14:39:44.6774750Z       timeout: timed out after 60000 msec waiting for promise to be resolved or rejected
2018-08-22T14:39:44.6896950Z 
```

Possibly a legit failure?

nathansobo:
Failed due to a render process crash on Windows. I was hoping this PR might be the *solution* to such crashes, but I'd also like to see if we can get a green build.

@smashwilson It looks like you've added some aggressive logging that fails with permissions issues on my local machine. Do you think that can just be removed? Or what are your thoughts on how we should handle logging for `@atom/watcher` in tests over the long term? Happy to do the work, but wanted to understand where your head was at when you were working on this.

smashwilson:
> Failed due to a render process crash on Windows. I was hoping this PR might be the solution to such crashes, but I'd also like to see if we can get a green build.

Awww. Although I'm a little surprised by the actual test that failed:

```
DefaultDirectoryProvider > .directoryForURISync(uri) > returns a Directory with a path that matches the uri [pass]
DefaultDirectoryProvider > .directoryForURISync(uri) > normalizes its input before creating a Directory for it [pass]
DefaultDirectoryProvider > .directoryForURISync(uri) > normalizes disk drive letter in path on #win32 [pass]
DefaultDirectoryProvider > .directoryForURISync(uri) > creates a Directory for its parent dir when passed a file Renderer process crashed, exiting
EBUSY: resource busy or locked, unlink 'C:\Users\VSSADM~1\AppData\Local\Temp\atom-test-data2019317-3484-1ajej98.008bj\databases\Databases.db'
```

That test shouldn't be creating any nsfw or @atom/watcher watchers. I think it does create a node-pathwatcher one though, by way of `Directory`. Some interaction between @atom/watcher and node-pathwatcher maybe?

>  It looks like you've added some aggressive logging that fails with permissions issues on my local machine. Do you think that can just be removed?

100% remove it. I added that stuff to try to get _some_ kind of actionable information out of CI failures, since I couldn't repro anything on my machine at all. It was a quick and dirty thing so I hardcoded paths and everything. Yank it all out if it's not helpful :smile:

nathansobo:
Going to merge in master now that we have crash reporting on Windows PR builds and see if I can reproduce the previous render process crashse.

nathansobo:
So the above CI failure is a similar crash. I downloaded the crash report and confirmed that the crash originates from `watcher.node`. I've spent 2 hours trying to get symbols for the stack trace to no avail. I'm starting to wonder if we should take a step back and consider a different approach to path watching.

rafeca:
Hey! 👋

With ❤️, the Atom team.

PR #17901, opened 2018-08-22
Fix error when launching graphical spec runner

daviwil:
### Description of the Change

This change fixes an issue introduced by PR #17892 which caused Atom's graphical spec runner to stop working.  This was caused by a simplification of dev resource path detection logic which left a couple of code paths lacking a proper resource path.  The solution in this PR is to roll back that change and introduce a simpler change which achieves the same benefits without the aforementioned issue.


daviwil:
Supplanted by https://github.com/atom/atom/pull/17903

PR #17903, opened 2018-08-22
Always provide a devResourcePath in load settings

daviwil:
### Description of the Change

This change fixes an issue introduced by PR #17892 which caused Atom's graphical spec runner to stop working. This was caused by a simplification of dev resource path detection logic which left a couple of code paths lacking a proper dev resource path. The solution in this PR is to always add a `devResourcePath` property to the load settings object that gets created when Atom starts up.

### Verification Process

Same verification steps as #17892:

- [x] `atom --dev .` uses `ATOM_DEV_RESOURCE_PATH` if one is set
- [x] `atom --dev .` in Atom repo path uses current directory as dev resource path when there is no `ATOM_DEV_RESOURCE_PATH` set and the current directory appears to be a clone of `atom/atom`
- [x] `atom --dev .` uses `~/github/atom` as the dev resource path if it exists and none of the path sources above are present
- [x] `atom --dev .` uses its bundled resource path when none of the path sources above are present
- [x] `atom -r path/to/atom` overrides all other possible resource path sources

Also validating the following two cases for each of the above items:

- [x] Package specs for settings-view run successfully
- [x] Opening Settings view and navigating to Packages tab does not throw an error about missing dependency module

### DIFF
--- a/src/main-process/parse-command-line.js
+++ b/src/main-process/parse-command-line.js
@@ -6,7 +6,7 @@ const {app} = require('electron')
 const path = require('path')
 const fs = require('fs-plus')
 
-module.exports = function parseCommandLine (processArgs, initialResourcePath) {
+module.exports = function parseCommandLine (processArgs, resourcePath, devResourcePath) {
   const options = yargs(processArgs).wrap(yargs.terminalWidth())
   const version = app.getVersion()
   options.usage(
@@ -119,8 +119,6 @@ module.exports = function parseCommandLine (processArgs, initialResourcePath) {
   let pathsToOpen = []
   let urlsToOpen = []
   let devMode = args['dev']
-  let devResourcePath = initialResourcePath
-  let resourcePath = null
 
   for (const path of args._) {
     if (path.startsWith('atom://')) {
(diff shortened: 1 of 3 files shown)


src/main-process/parse-command-line.js
```diff
@@ -130,21 +128,8 @@ module.exports = function parseCommandLine (processArgs, initialResourcePath) {
...
-    devResourcePath = args['resource-path']
-  }
-
-  if (test) {
-    devMode = true
-  }
-
-  if (devMode) {
-    resourcePath = devResourcePath
-  }
```
maxbrunsfeld: Is this conditional (assigning `resourcePath = devResourcePath` in dev mode) not needed anymore?

daviwil: It shouldn't be needed since it's being done here: https://github.com/atom/atom/blob/bb1f7b0a2fd8eb37e94843f6ef7337a61f5f3d83/src/main-process/main.js#L50

maxbrunsfeld: Oh yeah, sweet.

maxbrunsfeld:
⚡️

PR #17907, opened 2018-08-23
:arrow_up: metrics 1.6.2

annthurium:
Fold the changes from https://github.com/atom/metrics/pull/100 into atom/atom.


annthurium:
@daviwil mentioned that vsts linux builds have been busted all day, and gave me the thumbs up to merge even though this is red.

PR #17913, opened 2018-08-23
Avoid serialization race condition with slow package deactivation

captbaritone:
If any package takes longer than one second (the `saveStateDebounceInterval`) to deactivate, and the unload was triggered by a key or mouse down, mouse event, you can end up in a situation where
sate is serialized _after_ the packages are deactivated.

The result in a bug where panes, such as the File Tree, will randomly be closed when you reload or reopen Atom. We've seen multiple reports of this type of behavior on the Nuclide team, and I suspect this is the source.

This can be reproduced by creating a package that has an artificially slow `deactivate` method. With such a package enabled, every reload ends up serializing a state where all panes are closed.

I'm a bit nervous about this exact fix, since we have to track every place where it's possible for `prepare-to-unload` to be fired, without the window actually closing.

I handled the only instance I saw, but the logic is complex enough, that I'm not 100% confident there are not other instances.

If it did happen that `prepare-to-unload` was fired and some other logic caused the window to not actually close, we could end up in a state where mousedown/keydown events were no longer causing state to get serialized.

@matthewwithanm @maxbrunsfeld @wbinnssmith

### DIFF
--- a/spec/atom-environment-spec.js
+++ b/spec/atom-environment-spec.js
@@ -258,7 +258,7 @@ describe('AtomEnvironment', () => {
       atomEnv.destroy()
     })
 
-    it('ignores mousedown/keydown events happening after calling unloadEditorWindow', () => {
+    it('ignores mousedown/keydown events happening after calling prepareToUnloadEditorWindow', async () => {
       const atomEnv = new AtomEnvironment({
         applicationDelegate: global.atom.applicationDelegate
       })
@@ -276,18 +276,19 @@ describe('AtomEnvironment', () => {
 
       let mousedown = new MouseEvent('mousedown')
       atomEnv.document.dispatchEvent(mousedown)
-      atomEnv.unloadEditorWindow()
       expect(atomEnv.saveState).not.toHaveBeenCalled()
+      await atomEnv.prepareToUnloadEditorWindow()
+      expect(atomEnv.saveState).toHaveBeenCalledWith({isUnloading: true})
 
       advanceClock(atomEnv.saveStateDebounceInterval)
       idleCallbacks.shift()()
-      expect(atomEnv.saveState).not.toHaveBeenCalled()
+      expect(atomEnv.saveState.calls.length).toBe(1)
 
       mousedown = new MouseEvent('mousedown')
       atomEnv.document.dispatchEvent(mousedown)
       advanceClock(atomEnv.saveStateDebounceInterval)
       idleCallbacks.shift()()
-      expect(atomEnv.saveState).not.toHaveBeenCalled()
+      expect(atomEnv.saveState.calls.length).toBe(1)
 
       atomEnv.destroy()
     })
--- a/src/application-delegate.js
+++ b/src/application-delegate.js
@@ -325,13 +325,6 @@ class ApplicationDelegate {
     return new Disposable(() => ipcRenderer.removeListener('prepare-to-unload', outerCallback))
   }
 
-  onUnloadAborted (callback) {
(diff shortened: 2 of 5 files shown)


captbaritone:
Just talked with @matthewwithanm and he pointed me at #17873 which I think means I can get rid of the `unload-aborted` event. I'll update the PR.

Further, its seems that since  the callback triggered by `onRequestUnload` is deactivating packages, that we should just assume that its return value will always be respected. So, if we return `true` from our `onRequestUnload` callback, we can trust that the window is going to be closed and we don't have to worry about a case where the window continues to exist.

maxbrunsfeld:
This looks good to me.

> This can be reproduced by creating a package that has an artificially slow deactivate method. With such a package enabled, every reload ends up serializing a state where all panes are closed.

Have you already done some manual testing that this change fixes that scenario? If so, I'm cool with merging this. It looks like you just need to appease the linter; the actual main process tests passed on CI.

captbaritone:
@maxbrunsfeld My manual testing consisted of creating package which deactivates slowly (see below) and testing with it loaded.

_Before_ this PR change, reloading Atom via a hotkey would cause the file tree to be closed after Atom reloads.

_After_ this PR change, reloading Atom via a hotkey keeps the file tree open after Atom reloads.


```JavaScript
'use babel';

export default {
  activate(state) {
    console.log('activated slow-deactivator')
  },

  deactivate() {
    console.log('starting to deactivate slow-deactivator', new Date())
    return new Promise(resolve => {
      setTimeout(() => {
        console.log('done deactivating slow-deactivator', new Date())
        resolve();
      }, 2000);
    });
  }
};
```

captbaritone:
From what I can tell the build failure is unrelated to this change. I think this is ready to review/merge.

maxbrunsfeld:
@captbaritone On both mac and windows, a test related to state serialization failed. Can you repro this test failure locally if you run test enough times?

```
2018-08-27T16:01:49.4262160Z AtomEnvironment
2018-08-27T16:01:49.4275530Z   saving and loading
2018-08-27T16:01:49.4289160Z     it ignores mousedown/keydown events happening after calling unloadEditorWindow
2018-08-27T16:01:49.4303550Z       Expected spy saveState not to have been called.
2018-08-27T16:01:49.4322510Z         at it (/Users/vsts/agent/2.138.4/work/1/s/spec/atom-environment-spec.js:284:37)
2018-08-27T16:01:49.4352810Z         at jasmine.Spec.global.(anonymous function) (/Users/vsts/agent/2.138.4/work/1/s/spec/async-spec-helpers.js:27:22)
2018-08-27T16:01:49.4369860Z       Expected spy saveState not to have been called.
2018-08-27T16:01:49.4386890Z         at it (/Users/vsts/agent/2.138.4/work/1/s/spec/atom-environment-spec.js:290:37)
2018-08-27T16:01:49.4403310Z         at jasmine.Spec.global.(anonymous function) (/Users/vsts/agent/2.138.4/work/1/s/spec/async-spec-helpers.js:27:22)
```

captbaritone:
@maxbrunsfeld Ah yes. I can reproduce. That test needs to be updated. Working on that now.

captbaritone:
Okay. AppVeyor is now passing.

maxbrunsfeld:
It looks like the VSTS errors all occurred *after* the tests passed; there were problems uploading artifacts.

maxbrunsfeld:
Nice job tracking this down and fixing it, @captbaritone. This will ship in Atom 1.32.

captbaritone:
Awesome. Thanks for the timely review! 👏

PR #17926, opened 2018-08-25
Minify the startup snapshot script with terser

maxbrunsfeld:
### Background

I was just looking at a heap snapshot in Atom and I noticed something striking - over a third of the memory allocated on V8's heap is due to a single `script` object: Atom's concatenated startup script.

In particular, the script retains a reference to its source code as a string, which consumes around 45MB of RAM:

screen shot 2018-08-24 at 5 08 19 pm

In addition, the script maintains an array of `line ends`, which I assume V8 uses for computing stack traces. This consumes almost 3MB.

screen shot 2018-08-24 at 5 08 37 pm

### Proposed Change

In this PR, I minify the startup script that `electron-link` generates using [`terser`](https://github.com/fabiosantoscode/terser#mangle-properties-options) (a fork of uglify that works with es6).

### Results

When I compute a heap snapshot after minification, the startup script's `source` now consumes only 16MB on the heap, and its `line ends` no longer shows up in the snapshot (there are no line endings any more).

To get a sense of the overall impact, I rebuilt `Atom Dev` with this change and ran it simultaneously with the most recent `Atom Beta`. In each case, I opened one file in the `atom/atom` repo. When all is said and done, it looks like this reduces our memory consumption by **45 MB per window**, and the same amount in the main process!

screen shot 2018-08-24 at 5 20 53 pm

### Drawbacks

The minification step does take some time. It adds 10 - 20 seconds to the build.

### DIFF
--- a/script/lib/generate-startup-snapshot.js
+++ b/script/lib/generate-startup-snapshot.js
@@ -2,6 +2,7 @@ const childProcess = require('child_process')
 const fs = require('fs')
 const path = require('path')
 const electronLink = require('electron-link')
+const terser = require('terser')
 const CONFIG = require('../config')
 
 module.exports = function (packagedAppPath) {
@@ -67,8 +68,16 @@ module.exports = function (packagedAppPath) {
       )
     }
   }).then(({snapshotScript}) => {
-    fs.writeFileSync(snapshotScriptPath, snapshotScript)
     process.stdout.write('\n')
+
+    process.stdout.write('Minifying startup script')
+    const minification = terser.minify(snapshotScript, {
+      keep_classnames: true,
+      compress: {keep_fargs: true, keep_infinity: true}
+    })
+    if (minification.error) throw minification.error
+    process.stdout.write('\n')
+    fs.writeFileSync(snapshotScriptPath, minification.code)
 
     console.log('Verifying if snapshot can be executed via `mksnapshot`')
     const verifySnapshotScriptPath = path.join(CONFIG.repositoryRootPath, 'script', 'verify-snapshot-script')
--- a/script/package-lock.json
+++ b/script/package-lock.json
@@ -9926,6 +9926,32 @@
         }
       }
     },
+    "terser": {
+      "version": "3.8.1",
+      "resolved": "https://registry.npmjs.org/terser/-/terser-3.8.1.tgz",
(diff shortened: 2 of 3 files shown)


daviwil (approved):
That's pretty amazing! Does it have a similar reduction in the app size?

maxbrunsfeld:
Looks like the macOS failure was a flake. Restarting...

smashwilson (approved):
😍 Neat! Awesome find.

One question:  I see that we aren't generated sourcemaps for the minified code. Does this result in mangled stacktraces? An inline sourcemap would defeat the purpose of the minification I'm sure... but maybe one saved in a separate file wouldn't be loaded unless you actually needed it... ?

I've actually wondered if we could get away with not shipping source at all by default - relying on the snapshot for distribution of core and bundled packages. Maybe bundling it separately and downloading it on demand so that people could debug. It would really slim down our app download. It's _probably_ not worth it though 🤔

maxbrunsfeld:
Yeah, great points about stack traces and source maps. We’d need to do that before merging this. Hopefully it won’t negate the memory savings.

I think the only reason we include the source in the bundle is so you can run Atom in dev mode for debugging without having the source checked out. Not sure how many people do that though.

maxbrunsfeld:
One other drawback: the dev tools becomes unresponsive for about 10 seconds if you try to open the `embedded` script in the `Sources` tab. I guess the dev tools text editor (like several others) has trouble with long lines. This can be annoying, but I don't think it's a big enough problem to outweigh the benefits of minification; you can still use the dev tools for most normal purposes.

maxbrunsfeld:
Alright, CI passed after a rebuild: https://github.visualstudio.com/Atom/_build/results?buildId=5607&view=logs

maxbrunsfeld:
> Does it have a similar reduction in the app size?

It looks like it does reduce the size of `app.asar` from about 260MB to 220MB.

> I've actually wondered if we could get away with not shipping source at all by default - relying on the snapshot for distribution of core and bundled packages.

I agree. The source code is not very useful anyway in the form that we ship it. If people really want to debug Atom's bundled code, they pretty much need to clone and build Atom anyway. I'm inclined to attack this in a separate PR, but I do think it's worth doing.

PR #17938, opened 2018-08-28
Fix classes whose constructor name was broken by minification

maxbrunsfeld:
In Atom, several classes are registered as deserializers in a way that relies on their `constructor.name` property. When I [introduced minification](https://github.com/atom/atom/pull/17926), I disabled constructor name mangling because I anticipated that code might rely on `constructor.name`. But classes that are written in Coffee-script are translated to old-school functions as opposed to ES6 classes, so their names were still mangled, breaking deserialization.

* [x] `PaneAxis`
* [x] `TextBuffer`

/cc @Arcanemagus


Arcanemagus:
Confirming that fixes the deserialization errors I was seeing on v1.32.0-nightly2. 🎉

maxbrunsfeld:
That linux failure on VSTS looks unrelated to this change; the test script exited before it even started. The Travis-CI build passed.

PR #17958, opened 2018-08-30
Fix: Added capturing to keydown events when hiding tooltips

### LINKED ISSUE #17957: Tooltips still not closing on keydown events
### Prerequisites

* [X] Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-
atom/sections/debugging/#using-safe-mode
* [X] Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
* [X] Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
* [X] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
* [X] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Tooltips should fade away on keydown events. 

### Steps to Reproduce

*Edit by @rsese to format repro steps*

Using atom/atom as an example:

1. `atom --safe .`
2. Open `SUPPORT.md`
3. Hover over the tab for the file so the path tooltip appears
4. Start typing

Expected behavior: Tooltip(s) should be gone either on mouse hover away OR when any input event happens like keyboard click, mouse click, etc..

Actual behavior: Tooltip keeps in the place until I move mouse away only.

Reproduces how often: 100%

### Versions

Atom 1.3.0^

### Additional Information

tooltip-without-fix

cacheflow:
Description of the Change
Fix #17957 by ensuring capturing occurs and the event is bubbled up to the window. 

Alternate Designs
None.

Benefits
It fixes an issue where users may want to see the workspace they're typing in, but a tooltip is hiding it.

Possible Drawbacks
It adds an additional event listener for each tooltip.

Verification Process
I modified the existing tooltip-manager spec to include a new test for this feature, ran the test suite, and also manually tested it by reproducing the steps listed in #17957. 

Before fix
tooltip-without-fix

After fix
tooltip-with-fix

Applicable Issues
#17957

### DIFF
--- a/spec/tooltip-manager-spec.js
+++ b/spec/tooltip-manager-spec.js
@@ -218,7 +218,9 @@ describe('TooltipManager', () => {
         const disposable = manager.add(element, { title: 'Title' })
         hover(element, function () {
           expect(document.body.querySelector('.tooltip')).not.toBeNull()
-          window.dispatchEvent(new CustomEvent('keydown'))
+          window.dispatchEvent(new CustomEvent('keydown', {
+            bubbles: true
+          }))
           expect(document.body.querySelector('.tooltip')).toBeNull()
           disposable.dispose()
         })
--- a/src/tooltip-manager.js
+++ b/src/tooltip-manager.js
@@ -153,11 +153,11 @@ class TooltipManager {
     }
 
     window.addEventListener('resize', hideTooltip)
-    window.addEventListener('keydown', hideTooltip)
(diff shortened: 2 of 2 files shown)


daviwil (approved):
Restarted VSTS build failed for an unrelated reason, I think this PR is good.  Thanks a lot @cacheflow!

PR #17989, opened 2018-09-06
Improve RPM package dependencies

### LINKED ISSUE #16977: RPM: Stop requiring LSB altogether
### Prerequisites

### Description
Atom requires LSB to be installed on my system. LSB is very obsolete see for example https://bugzilla.opensuse.org/show_bug.cgi?id=1071688 openSUSE Tumbleweed currently has LSBv4 but can't provide LSBv5 because that would require adding obsolete unmaintained insecure packages IIUC.
Introduced in https://github.com/atom/atom/commit/f919bc40dc033c914dcec014597c7b59da1617c5 DEB packages only got a recommends while RPM packages got a requires. Please stop relying on LSB or at least downgrade to a Recommends only.

### Steps to Reproduce
1. `docker run --rm opensuse:tumbleweed sh -c "zypper in -y curl && curl -L -o atom.rpm https://atom.io/download/rpm && zypper in -y -f atom.rpm"`

**Expected behavior:** does not pull in unnecessary dependencies.

**Actual behavior:** atom pulls in for example Qt3 o.O

**Reproduces how often:** Always

### Versions
every atom version since https://github.com/atom/atom/commit/f919bc40dc033c914dcec014597c7b59da1617c5 (0.202?)
OS: openSUSE Tumbleweed 20180308 (yeah I need to update... but qt3 dep is broken...)

### Additional Information
Why I (tried to) killed the LSB, Didier Raboud, DebConf 17: https://debconf17.debconf.org/talks/181/ Slides: https://annex.debconf.org/debconf-share/debconf17/slides/181-why_i_tried_to_kill_the_lsb.pdf

Debian dropping the Linux Standard Base: https://lwn.net/Articles/658809/

Tumbleweed is missing the lsb5 package: https://bugzilla.opensuse.org/show_bug.cgi?id=1071688

### LINKED ISSUE #17241: Missing GConf2 dependency in rpm package
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:

### Description
The Atom rpm package is missing GConf2 dependency. 

### Steps to Reproduce

1. Install Atom (using repository or directly with rpm package)
2. Try to launch atom

**Expected behavior:**
Atom gets launched correctly

**Actual behavior:**
Atom it's not launched because of the missing dependency. The error is:
$ /usr/share/atom/atom: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory

**Reproduces how often:** 
Always (until GConf2 is installed)

### Versions
All versions. Running on Fedora 27

### Additional Information
The missing dependency  doesn't get noticed when installing Atom on default Fedora Workstation
because some other package requires GConf2 which is already installed. However installing on a
customized environment highlights the problem. Altering the rpm with rpmrebuild and adding GConf2 as dependency solves the issue.

daviwil:
### Description of the Change

This PR makes a couple small improvements to Atom's RPM package build:

- Makes Linux Standard Base "recommended" (optional) so that users of some Linux distros don't get a bunch of unwanted packages installed
- Adds GConf2 as a required dependency of the package (as we do in the .deb package)

### Alternate Designs

None.

This improves our RPM package build for official Atom releases.

### Benefits

See above.

### Possible Drawbacks

Could be possible that removing LSB from required dependencies could cause an important dependency to be missed, but this is something we will be able to fix if and when it shows up.

### Verification Process

- [ ] Green CI build
- [ ] Build an RPM package and check its dependencies
- [ ] Build an RPM package and try to install it

### Applicable Issues

Fixes #16977.
Fixes #17241.

### DIFF
--- a/resources/linux/redhat/atom.spec.in
+++ b/resources/linux/redhat/atom.spec.in
@@ -8,9 +8,11 @@ AutoReqProv:    no # Avoid libchromiumcontent.so missing dependency
 Prefix:         <%= installDir %>
 
 %ifarch i386 i486 i586 i686
-Requires: lsb-core-noarch, libXss.so.1 libsecret-1.so.0
+Requires: libXss.so.1 libsecret-1.so.0 lib
+Recommends: lsb-core-noarch
 %else
-Requires: lsb-core-noarch, libXss.so.1()(64bit) libsecret-1.so.0()(64bit)
+Requires: libXss.so.1()(64bit) libsecret-1.so.0()(64bit)
+Recommends: lsb-core-noarch
 %endif
 
 %description
--- a/resources/linux/redhat/atom.spec.in
+++ b/resources/linux/redhat/atom.spec.in
@@ -8,10 +8,10 @@ AutoReqProv:    no # Avoid libchromiumcontent.so missing dependency
 Prefix:         <%= installDir %>
 
 %ifarch i386 i486 i586 i686
-Requires: libXss.so.1 libsecret-1.so.0 lib
+Requires: libgconf-2.so.4 libXss.so.1 libsecret-1.so.0 lib
 Recommends: lsb-core-noarch
 %else
-Requires: libXss.so.1()(64bit) libsecret-1.so.0()(64bit)
+Requires: libgconf-2.so.4()(64bit) libXss.so.1()(64bit) libsecret-1.so.0()(64bit)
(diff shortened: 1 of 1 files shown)


daviwil:
Apparently the `Recommended:` syntax isn't standardized across distros enough yet for this to work for all of them.  Closing this for now until we find a better solution.

PR #18012, opened 2018-09-10
Update to Electron 2.0.9

mfonville:
Updated Electron fixes segfault on glibc 2.28 (see https://github.com/electron/electron/issues/13972 )

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/atom/atom/issues"
   },
   "license": "MIT",
-  "electronVersion": "2.0.7",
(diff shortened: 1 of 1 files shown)


daviwil:
Thanks for sending this!  This must be the fix for the issue that prevented us from getting a clean Linux build for 2.0.8.  Looks like the VSTS Linux build failed for an unrelated issue so I'm going to run it again just in case.

daviwil (approved):

daviwil:
Build looks good, thanks again!

kappapiana:
That solves my issues too. Thanks

PR #18014, opened 2018-09-10
➡️ Migrate core package 'autoflow' into ./packages

daviwil:
### Description of the Change

This PR migrates the core [`autoflow`](https://github.com/atom/autoflow) package into `atom/atom`.  See issue #17833 for more information.


packages/autoflow/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "autoflow",
+  "version": "0.29.4",
+  "main": "./lib/autoflow",
+  "description": "Format the current selection to have lines no longer than 80 characters.\n\nThis packages uses the config value of `editor.preferredLineLength` when set.",
+  "activationCommands": {
+    "atom-text-editor": [
+      "autoflow:reflow-selection"
+    ]
+  },
+  "repository": "https://github.com/atom/autoflow",
```
jasonrudolph: Should we change this to `https://github.com/atom/atom`?

(Note: I should have caught this during the other package migrations. Sorry! 😳)

jasonrudolph (approved):

packages/autoflow/package.json
daviwil: Hmmm, yeah, I didn't think about that either!  I'll go update this package and the others and then add another step to the migration issues to mention this.

daviwil:
Same ol' Linux flake on VSTS, I'll look into that after merging this.

PR #18015, opened 2018-09-11
Accept a TextMate scope selector in bufferRangeForScopeAtPosition

maxbrunsfeld:
Refs https://github.com/atom/atom/pull/17881#issuecomment-419965914

/cc @t9md

### DIFF
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -1504,20 +1504,21 @@ describe('TreeSitterLanguageMode', () => {
       it('returns the range of the smallest matching node at position', async () => {
         const grammar = new TreeSitterGrammar(atom.grammars, jsGrammarPath, {
           scopeName: 'javascript',
-          parser: 'tree-sitter-javascript'
+          parser: 'tree-sitter-javascript',
+          scopes: {
+            'property_identifier': 'variable.other.object.property',
+            'template_string': 'string.quoted.template'
+          }
         })
 
-        buffer.setText('foo({bar: baz});')
+        buffer.setText('a(`${b({ccc: ddd})} eee`);')
 
         buffer.setLanguageMode(new TreeSitterLanguageMode({buffer, grammar}))
-        expect(editor.bufferRangeForScopeAtPosition('.property_identifier', [0, 6])).toEqual(
-          buffer.findSync('bar')
+        expect(editor.bufferRangeForScopeAtPosition('.variable.property', [0, 9])).toEqual(
+          [[0, 8], [0, 11]]
         )
-        expect(editor.bufferRangeForScopeAtPosition('.call_expression', [0, 6])).toEqual(
-          [[0, 0], [0, buffer.getText().length - 1]]
-        )
-        expect(editor.bufferRangeForScopeAtPosition('.object', [0, 9])).toEqual(
-          buffer.findSync('{bar: baz}')
+        expect(editor.bufferRangeForScopeAtPosition('.string.quoted', [0, 6])).toEqual(
+          [[0, 2], [0, 24]]
         )
       })
 
@@ -1525,7 +1526,9 @@ describe('TreeSitterLanguageMode', () => {
         const jsGrammar = new TreeSitterGrammar(atom.grammars, jsGrammarPath, {
           scopeName: 'javascript',
           parser: 'tree-sitter-javascript',
-          scopes: {},
+          scopes: {
+            'property_identifier': 'variable.other.object.property',
+          },
           injectionRegExp: 'javascript',
           injectionPoints: [HTML_TEMPLATE_LITERAL_INJECTION_POINT]
         })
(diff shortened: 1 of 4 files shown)


t9md:
state

Thanks, I tried your branch and I could select inline block comment by `comment.block` scope.
But still have some minor diff.
Above is this branch's behavior. Double slash line comment start with `//` can also be selected by `comment.block`.
Which was NOT in TextMate parser(it doesn't return range of line-comment).
TreeSitter does not distinguish line-comment and inline-block-comment.

I think this is as designed.
Can't I expect 100% of scope compatibility when switching to TreeSitter?
For this one, previous behavior is better since it allow granular level of control to the package author.

t9md:
I'm getting to understand it.

TreeSitter parses code more correctly and consistently without depending on error-prone regex matching.

By that, tree-sitter doesn't distinguish function name used in declaration and function call.
Also doesn't distinguish inline-comment or line-comment, it's all just a "comment".
So we see same syntax color whenever function name appears.

Although it was error-prone, regex-based parsing also provided more class names(since a variety of regex matchings are directly converted to the class names used in CSS), it can be used when matching to a very specific context using a scope name.

As my understanding, this approach is no longer feasible in tree-sitter. SInce tree-sitter doesn't make unnecessary distinguishment which was the natural consequence where regex-based parsing is exercised in old parsing mechanism.

Aerijo:
> tree-sitter doesn't distinguish function name used in declaration and function call

It can. I believe your talking about the translation to scopes... maybe.

> Also doesn't distinguish inline-comment or line-comment, it's all just a "comment".

Again, it can. It just appears that the grammar writer [did not decide to do this](https://github.com/tree-sitter/tree-sitter-javascript/blob/d16ed0303e6926351a3ed54fb2de34c11c1d1419/grammar.js#L745). If you make a case for why it would help, I'm sure changes will be considered.

> regex-based parsing also provided more class names(since a variety of regex matchings are directly converted to the class names used in CSS), it can be used when matching to a very specific context using a scope name

Well, if you judge Tree-sitter by the scopes it provides, then yeah, probably. My understanding is you're supposed to interact with the nodes of a Tree-sitter grammar, and not so much the scopes. Of course you _can_ work with scopes, but as you say they seem to be lacking compared to TextMate grammars. This will be improved as people raise issues and make PRs though.

>  SInce tree-sitter doesn't make unnecessary distinguishment which was the natural consequence where regex-based parsing is exercised in old parsing mechanism.

I'm not sure what you mean here.

t9md:
@Aerijo I understand what you explained, thanks.
Yeah, I agree with your suggestion, maybe I'd better work on the nodes than the scopes.
But that means I need to do lots of work to migrate to tree-sitter in my vim-mode-plus package.
So I tried to find the way to creating the shim which allows old code works without making big changes.

Anyway, I'll continue to dig further to get me to understand the changes.

Thanks.

maxbrunsfeld:
@t9md Yeah, I'm still figuring out what we need to do for packages to transition. Basically, Tree-sitter can provide you much *more* information about the code than TextMate could, but the data is structured differently - in a syntax tree instead of in a list of scopes.

The problem with scopes is that if there are a lot of nested scopes, they must all be rendered to the DOM as a series of nested spans, which has a steep performance cost during rendering. So initially, when introducing Tree-sitter and designing the [scope mapping](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/#syntax-highlighting) system for theme compatibility, I wanted to produce the most *minimal* set of scopes that would be needed for themes to work.

For example, as you mentioned, we don't produce shims for scopes like `meta.method-call.js`. I believed that this was ok because those scopes were already quite unreliable with TextMate. For example, adding or removing a `\n` character can change `meta.method-call.js` into `meta.function-call.js`, or completely remove a scope like `meta.function.definition.js`.

I think that if you do want to make more sophisticated queries of syntactic structure, Tree-sitter should be more useful in the long term, but you'll need to use different APIs. Currently, the new APIs are not documented, because I haven't fully decided which ones we should commit to. Maybe if you get some time, you could let me know which APIs would be useful to you. I would think that Vim-mode-plus's text object system might be a good use case for these new syntax tree APIs. You could try something like this:

```js
const languageMode = editor.getBuffer().getLanguageMode()

let range
if (languageMode.getSyntaxNodeAtPosition) {
  // new code

  const node = languageMode.getSyntaxNodeAtPosition(position)

  // The names of these nodes will depend on the grammar
  const functionNode = node.closest(['function', 'arrow_function', 'method_definition'])

  range = functionNode.range

} else {
  // old code
}
```

Sorry that this isn't all figured out yet; I know you might not have time to dive into this stuff. But in case you do, here are some things that might be useful to look at:

* [The JS Grammar](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js) (to see the possible types of the syntax nodes available)
* [The type declarations for the Tree-sitter API](https://github.com/tree-sitter/node-tree-sitter/blob/master/tree-sitter.d.ts)

t9md:
Thanks for the elaborate explanation, I didn't know that the flight-manual have such a detailed explanation of new tree-sitter.
I've already checked some example grammars and also your talk on YouTube.

Anyway, I completely agree that the tree-sitter way is more reliable and give more structured information into code.
Nice work and thank you for creating such an ambitious project(I knew tree-sitter from a few years ago when I have no idea what it goes.)

Your example code for text-object is very helpful, I've already been trying similar thing as well [here](https://github.com/t9md/atom-vim-mode-plus/pull/1096).

If I have reached to clear understanding for API I need, I'll tell you.

PR #18016, opened 2018-09-11
Fix autocomplete icon

simurai:
### Description of the Change

This fixes the icons in the autocomplete popover.

Before | After
--- | ---
image | image

### Alternate Designs

The selector used to win in specificity is pretty ugly, but we can't really change the class names because other packages and themes use it too.

It's a bundled package

### Benefits

Icon's background doesn't look broken anymore.

### Possible Drawbacks

Even harder to override

### Verification Process

1. Open a file
2. Start typing so that the autocomplete popover shows up
3. Verify that the icon's background looks ok

### Applicable Issues

None, pointed out in Slack by @maxbrunsfeld

### DIFF
--- a/packages/one-dark-ui/styles/lists.less
+++ b/packages/one-dark-ui/styles/lists.less
@@ -123,6 +123,12 @@
   .list-group li {
     padding-left: @popover-list-padding;
   }
+
+  // Un-reset icon in popover lists
+  .icon.icon {
+    display: inline-block;
+    height: inherit;
+  }
 }
 
 .ui-sortable {
--- a/packages/one-light-ui/styles/lists.less
+++ b/packages/one-light-ui/styles/lists.less
@@ -123,6 +123,12 @@
   .list-group li {
     padding-left: @popover-list-padding;
   }
+
+  // Un-reset icon in popover lists
+  .icon.icon {
+    display: inline-block;
+    height: inherit;
+  }
 }
 
 .ui-sortable {


thomasjo:
How did this regression sneak in? Maybe we should experiment with some (simple) means of preventing this kind of breakage in the future — ideally some type of automated test. If not, we should at least have some manual smoke testing process that should pass anytime we touch GUI bits.

simurai:
> How did this regression sneak in?

By trying to fix this [issue](https://github.com/atom/one-light-ui/issues/137) with this [PR](https://github.com/atom/one-light-ui/pull/141). Reminds me of this [joke](https://twitter.com/thomasfuchs/status/493790680397803521?lang=en). :joy:

> ideally some type of automated test

I think there are some testing frameworks that do "visual diffing". Take before and after screenshots and test if pixels changed. No idea how easy it would be to add something like that.

> manual smoke testing process

Yeah, have a list of things to manually click through in the UI. It wouldn't catch everything and there are probably community packages that we constantly break, but at least for the core UI.

Another solution would be to make components more robust and independant. Like in this PR's case, the CSS class `.list-group` gets used by the autocomplete popover, but also by the tree-view, command palette, search results and probably other places and also community packages. They are all "lists" but still different enough that it would make more sense to style them independently so they wouldn't need overriding for each case. But it's kinda too late because themes, packages and user's styles.less already use the `.list-group` class and we have to keep it till a next major version of Atom.

So yeah, maybe back to a smoke test?

maxbrunsfeld:
The package migration is paying off; it's nice to see both of these changes in one PR.

Thanks for fixing this!

simurai:
> The package migration is paying off; it's nice to see both of these changes in one PR.

Yes, totally. 💯 It's super nice to not have to make two "identical" PRs.

We might could go even further and have a "master" theme from where other color variations could import the styles from. I'll have to test that one day.

PR #18049, opened 2018-09-13
Use a custom Ubuntu 18.04 container for VSTS Linux builds

daviwil:
### Description of the Change

This change seeks to fix a recurring failure in our VSTS CI tests on Linux possibly caused by a build dependency on Ubuntu 16.04 that clashes with Electron 2.0.8+.  The fix is to use a [custom container image]() based on Ubuntu 18.04 so that we can build and test successfully with Electron 2.0.8+.

### Alternate Designs

Find the issue in Electron causing our tests to fail and get it fixed, but that will be much more time consuming.  I'll pursue this also but only after getting our builds green again.

It fixes an issue preventing our Linux CI runs on VSTS from succeeding reliably.

### Benefits

It fixes an issue preventing our Linux CI runs on VSTS from succeeding reliably.

### Possible Drawbacks

Possibly slower builds due to the upgrade step.

### Verification Process

- [x] Ensure CI passes after using new container image
- [x] Try a build generated on Ubuntu 18.04 on an Ubuntu 16.04 VM and make sure it runs correctly and passes a basic sanity check

### Applicable Issues

None.

### DIFF
--- a/script/vsts/nightly-release.yml
+++ b/script/vsts/nightly-release.yml
@@ -1,3 +1,8 @@
+resources:
+  containers:
+  - container: atom-linux-ci
+    image: daviwil/atom-linux-ci:latest
+
 jobs:
 
 - job: GetReleaseVersion
@@ -17,7 +22,7 @@ jobs:
 - template: platforms/linux.yml
 
 - job: Release
-  pool: 
+  pool:
     vmImage: vs2015-win2012r2 # needed for Python 2.7 and gyp
 
   dependsOn:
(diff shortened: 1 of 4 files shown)


daviwil:
Build of Atom from the Ubuntu 18.04 container works successfully on Ubuntu 16.04.  :shipit:

daviwil:
Also huge thanks to @Arcanemagus who provided some valuable insight and a [great starting point](https://github.com/Arcanemagus/atom-docker-ci) for the Linux container image :sparkles:

PR #18053, opened 2018-09-14
Refactoring, remove reduanded else

janczer:
### Description of the Change

We can delete last `else` because this no necessary.

For example, if we have code like this:

```javascript
if (typeof child === 'number') {
  return child
} else {
  return this.nodes.indexOf(child)
}
```

And variable `child` has type `number` then the function always return `child` and never goes on the lines after if-statment. So we can delete `else`-branch:

```javascript
if (typeof child === 'number') {
  return child
}
return this.nodes.indexOf(child)
``` 

I think this part of code more readable then first one.

It can do code more clear.

### Benefits

Another developer can easily understand the code. 

### Possible Drawbacks

Non

### Verification Process

Run the tests

### Applicable Issues

Non

### DIFF
--- a/src/atom-environment.js
+++ b/src/atom-environment.js
@@ -443,9 +443,8 @@ class AtomEnvironment {
     if (this.shellEnvironmentLoaded) {
       callback()
       return new Disposable()
-    } else {
-      return this.emitter.once('loaded-shell-environment', callback)
     }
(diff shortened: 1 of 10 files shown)


lee-dohm:
Thanks for the suggestion.

We prefer the explicitness of including the else case.

Issue #18058, opened 2018-09-15
Crash in directories with invalid package.json


Arcanemagus:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

If you attempt to open Atom in a directory containing a `package.json` that has invalid syntax it will crash with an error about a JavaScript error in the main process when calling this line:
https://github.com/atom/atom/blob/e6507c59cb846cd02b84f4a3b10dd43b2572a939/src/main-process/main.js#L23

### Steps to Reproduce

1. Save a `package.json` with invalid syntax (for example, a trailing `,`)
3. Open an Atom window in the directory containing the invalid `package.json`

#### Expected behavior:
No error is thrown.

#### Actual behavior:
The following error is thrown:
![image](https://user-images.githubusercontent.com/427137/45584137-a13a9e00-b883-11e8-9323-cbd11cab2912.png)

```
[Window Title]
Error

[Main Instruction]
A JavaScript error occurred in the main process

[Content]
Uncaught Exception:
SyntaxError: Unexpected token } in JSON at position 2838
    at JSON.parse (<anonymous>)
    at parseObject (C:\Users\Lando\AppData\Local\atom\app-1.32.0-nightly9\resources\app.asar\node_modules\season\lib\cson.js:54:19)
    at parseContentsSync (C:\Users\Lando\AppData\Local\atom\app-1.32.0-nightly9\resources\app.asar\node_modules\season\lib\cson.js:68:16)
    at Object.readFileSync (C:\Users\Lando\AppData\Local\atom\app-1.32.0-nightly9\resources\app.asar\node_modules\season\lib\cson.js:180:14)
    at isAtomRepoPath (C:\Users\Lando\AppData\Local\atom\app-1.32.0-nightly9\resources\app.asar\src\main-process\main.js:23:28)
[...]

PR #18060, opened 2018-09-15
Catch exceptions for bad package.json when looking for Atom repo in pwd

daviwil:
### Description of the Change

Fixes an issue encountered by @Arcanemagus where opening Atom in a folder with a malformed `package.json` will cause Atom's main process to crash.

### Alternate Designs

None.

### Benefits

No more :boom:

### Possible Drawbacks

None.

### Verification Process

- [x] Opened Atom in a folder with the following `package.json` and verified that it does not crash

```
{ name: "atom": , }
```

### Applicable Issues

Fixes https://github.com/atom/atom/issues/18058.


daviwil:
Looks like AppVeyor is failing pretty regularly now, I'll look at that on Monday.

PR #18067, opened 2018-09-17
Try out the text-buffer prerelease

smashwilson:
atom/text-buffer#303 fixes what looks an awful lot like a bug in TextBuffer, but it could have non-obvious ramifications beyond TextBuffer code. Let's see if the test suite catches anything 🤞 

cc @maxbrunsfeld

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5420,9 +5420,9 @@
       }
     },
     "text-buffer": {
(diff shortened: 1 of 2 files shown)


smashwilson:
Green on Windows on VSTS 👍

PR #18087, opened 2018-09-21
Update soft wrap margins on gutter re-measurement

### LINKED ISSUE #18086: Editors with soft wrap and custom gutters render with an incorrect soft wrap column first
### Prerequisites

### Description

This is a very narrow edge case in the implementation of atom/github#1512. The symptom I observed is described in @Arcanemagus' [comment](https://github.com/atom/github/pull/1512#issuecomment-422573966):

> If lines are wrapped in the diff, on first click in the diff window after initial rendering the line wrapping is recalculated. Probably an issue with the initial width calculation?
>
> soft-wrap-changing

Essentially, what's happening is that a soft-wrapped TextEditor with custom gutters is being rendered on the first frame with an incorrect soft wrap margin, because a change in gutter width measurement isn't triggering a re-update.

### Steps to Reproduce

This one's tricky to reproduce because (a) we don't use custom gutters that heavily, (b) when we do, they're often decorated, which triggers a component update and conceals the bug, and (c) it only manifests with soft-wrapping enabled.

1. Disable core packages that create decorations on TextEditor creation (`apm disable spell-check && apm disable git-diff`).
2. Create a text file that will be soft-wrapped at your typical screen width.
3. Add the following to your init script:
   ```js
   atom.workspace.observeTextEditors(editor => {
      editor.addGutter({name: 'repro', priority: 10})
   });
   ```
4. Add the following to your stylesheet:
    ```less
    .gutter[gutter-name="repro"] {
      width: 100px;
      background-color: grey;
    }
    ```
5. Enable soft wrapping (_not_ at preferred line length) in TextEditors in the Settings View or in your dev console.
    ```js
    atom.config.set('editor.softWrap', true)
    atom.config.set('editor.softWrapAtPreferredLineLength', false)
    ```
4. (Re-)open Atom in safe mode.
5. Open the file you prepared in step 2.
6. Interact with the opened editor pane in any way. A mouse click anywhere will trigger this.

**Expected behavior:**

The soft-wrap column should be calculated on the first render accounting for the width of the custom gutter. Editor interactions such as clicking to position a cursor or scrolling vertical content should not alter the soft-wrap column.

**Actual behavior:**

The soft-wrap column is calculated initially _without_ accounting for the custom gutter width, but then accounted for on the next re-render for any cause, resulting in some jank:

gutter-measurement-bug

### Versions

```
$ atom --version
Atom    : 1.32.0-nightly15
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3
$ apm --version
apm  2.1.2
npm  6.2.0
node 8.9.3 x64
atom 1.32.0-nightly15
python 2.7.15
git 2.19.0
```

### Additional Information

_N/A_

smashwilson:

### Identify the Bug

This fixes #18086. I've verified that this corrects the editor jank I observed both in my super-contrived isolated repro case _and_ in the case where @Arcanemagus originally spotted it in atom/github#1512.

### Description of the Change

Return `true` from `TextEditorComponent::measureContentDuringUpdateSync()` if the horizontal scrollbar has toggled its visibility _or_ if gutter measurements have changed. This will cause `updateSync()` to restart the initial frame with the correct gutter measurements in place and calculate the correct soft-wrap column in time for the first render to complete.

### Alternate Designs

I _could_ have worked around this by manually triggering a TextEditorComponent update after the initial render has completed. That seems pretty hacky, though, and that PR already depends on features from 1.32, so I figured I might as well fix The Real Problem.

### Possible Drawbacks

This could cause a more expensive initial render. We're already paying this price when a horizontal scrollbar is needed, though.

### Verification Process

I've followed the repro steps from #18086 to make sure that the editor renders correctly on the first try. I've also used dev mode to verify that it fixes the way that this manifests in the GitHub package.

### DIFF
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2085,6 +2085,22 @@ describe('TextEditorComponent', () => {
       const bLabels = Array.from(bNumbers, e => e.textContent)
       expect(bLabels).toEqual(['b - 0', 'b - 1', 'b - 2', 'b - 3', 'b - 4', 'b - 5'])
     })
+
+    it("updates the editor's soft wrap width when a custom gutter's measurement is available", () => {
+      const {component, element, editor} = buildComponent({
+        lineNumberGutterVisible: false,
+        width: 400,
+        softWrapped: true,
+        attach: false,
+      })
+      const gutter = editor.addGutter({name: 'a', priority: 10})
+      gutter.getElement().style.width = '100px'
+
+      jasmine.attachToDOM(element)
+
+      expect(component.getGutterContainerWidth()).toBe(100)
+      expect(editor.getSoftWrapColumn()).toBe(39);
+    })
   })
 
   describe('block decorations', () => {
--- a/spec/text-editor-component-spec.js
+++ b/spec/text-editor-component-spec.js
@@ -2099,7 +2099,11 @@ describe('TextEditorComponent', () => {
       jasmine.attachToDOM(element)
 
       expect(component.getGutterContainerWidth()).toBe(100)
-      expect(editor.getSoftWrapColumn()).toBe(39);
+
+      // Component client width - gutter container width - vertical scrollbar width
+      const softWrapColumn = Math.floor(
+        (400 - 100 - component.getVerticalScrollbarWidth()) / component.getBaseCharacterWidth())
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
👏  Way to move fast and fix things.

PR #18093, opened 2018-09-24
➡️ Migrate core package 'atom-dark-syntax' into ./packages

daviwil:
### Description of the Change

This PR migrates the core [`atom-dark-syntax`](https://github.com/atom/atom-dark-syntax) package into `atom/atom`.  See issue #17849 for more information.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -299,8 +299,7 @@
       }
     },
     "atom-dark-syntax": {
-      "version": "https://www.atom.io/api/packages/atom-dark-syntax/versions/0.29.1/tarball",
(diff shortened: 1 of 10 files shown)


jasonrudolph:
I see that [atom/atom-dark-syntax repo has a LICENSE.md file](https://github.com/atom/atom-dark-syntax/blob/698c718f6c7048f5c611cb237ae1a5dcfb052b1f/LICENSE.md), and that it's not included as part of this migration. Is that intentional? 

I ask because [I've brought over the LICENSE.md file in previous migrations](https://github.com/atom/atom/blob/64fc36f664f7ed7ae5333e6655ba008dc20ce4e4/packages/one-light-syntax/LICENSE.md). If we'd prefer to exclude the LICENSE.md files from these package subdirectories, I'm totally 👌 with that, but I'd like us to do the same thing for all of them. 😇 

Whaddya think?

jasonrudolph (approved):
Just [one minor question](https://github.com/atom/atom/pull/18093#issuecomment-424051883). Overall, this looks great. 🚀

PR #18095, opened 2018-09-24
Keep recovery files around for an hour

queerviolet:
### Identify the Bug

#11406 

### Description of the Change

Keep recovery files around longer. Instead of removing recovery files as soon as they've been saved, keep them around for about an hour. This provides an inconvenient but better than nothing safeguard against data loss.

Remediates, but doesn't fix, #11406.

When they would previously have been unlinked, recovery files are now renamed to `$filename~` (yes, ala Emacs). Once an hour, we garbage collect any recovery files that are more than one hour old.

### Alternate Designs

I mean, we should actually fix the bug, I just haven't been able to reliably repro it, and I don't really know what's going on. (More common in my experience was corrupting the state file, but that just resulted in a lost session).

I thought maybe this would be temporary, but I've talked myself into believing that recovery files should work this way generally, precisely to protect against data loss in unforeseen circumstances. It may also help our forensics on the root bug, since we'll basically have a log of save actions

#### Disk space impact
|Save rate / min|Size of `~/.atom/recovery`|
|---|---|
|1k|3.6mb|
|10k|30.6mb|
|100k|300.6mb|

It gets a bit big if you're saving 100k file once per minute, but that's a... really big Atom buffer to be editing with such wild abandon. I think it's okay, but welcome thoughts.

Technically, this introduces a bug if a file save operation takes more than a hour. I think we'll live.

### Verification Process

Running some specs. Making sure files still save. The usual.

### DIFF
--- a/spec/main-process/file-recovery-service.test.js
+++ b/spec/main-process/file-recovery-service.test.js
@@ -26,23 +26,28 @@ describe("FileRecoveryService", () => {
   })
 
   describe("when no crash happens during a save", () => {
-    it("creates a recovery file and deletes it after saving", async () => {
+    it("creates a recovery file and renames it after saving", async () => {
       const mockWindow = {}
       const filePath = temp.path()
 
       fs.writeFileSync(filePath, "some content")
       await recoveryService.willSavePath(mockWindow, filePath)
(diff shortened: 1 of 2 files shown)


lee-dohm:
Closing this for now until we get a better handle on the scope of the problem and can fix the tests.

PR #18116, opened 2018-09-26
 Add a simple server for testing Atom updates, enable Atom to use it via environment variable

daviwil:
### Description of the Change

This change adds a simple script that replicates Atom's update API on atom.io for local testing of Atom's update process on Windows and macOS.  It looks for a full Atom build in the `./out` and then serves that to Atom when Atom is launched with a special `ATOM_UPDATE_URL_PREFIX` environment variable that causes update checks to be sent to the local server.

### Alternate Designs

One can already test upgrades using a local build of atom.io, but that's a pretty heavy requirement if one only needs to test Atom's update checks.

### Possible Drawbacks

One drawback in the short term is that current Atom releases do not look for the `ATOM_UPDATE_URL_PREFIX` environment variable to override the update URL so one will either need to test with Nightly builds (after this merges) or make a build of Atom stable or beta that includes this change.

### Verification Process

- [x] Built an update on Windows, used the update server to successfully upgrade Atom
- [x] Built an update on macOS, used the update server to successfully upgrade Atom

### DIFF
--- a/script/build
+++ b/script/build
@@ -102,6 +102,10 @@ if (!argv.generateApiDocs) {
           break
         }
         case 'win32': {
+          if (argv.testSign) {
+            console.log('Test signing is not supported on Windows, skipping.'.gray)
+          }
+
           if (argv.codeSign) {
             const executablesToSign = [ path.join(packagedAppPath, 'Atom.exe') ]
             if (argv.createWindowsInstaller) {
@@ -117,8 +121,6 @@ if (!argv.generateApiDocs) {
                 argv.codeSign && codeSignOnWindows([installerPath])
                 return packagedAppPath
               })
-          } else if (argv.testSign) {
-            console.log('Test signing is not supported on Windows, skipping.'.gray)
           } else {
             console.log('Skipping creating installer. Specify the --create-windows-installer option to create a Squirrel-based Windows installer.'.gray)
           }
--- a/script/build
+++ b/script/build
@@ -21,6 +21,7 @@ const argv = yargs
   .help('help')
   .describe('existing-binaries', 'Use existing Atom binaries (skip clean/transpile/cache)')
   .describe('code-sign', 'Code-sign executables (macOS and Windows only)')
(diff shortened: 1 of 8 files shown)


daviwil:
Still need to add some more docs about test signing on macOS, but it's working well for updates there so far!

daviwil:
Mac Development certificate steps have been added.  This PR is ready to be reviewed!

jasonrudolph:
> This PR is ready to be reviewed!

Thanks, @daviwil. :zap: I'll review this PR today.

jasonrudolph (commented):
Thanks for putting this together, @daviwil. ✨ 

I verified that I can run the server locally, and I hit a few of the endpoints to verify that they're working locally for me (e.g., `/api/updates`).

I also reviewed the diff and noted a few minor questions/suggestions inline below.

Based on the following excerpt from the PR body, I don't _think_ it's possible yet to fully test Atom's ability to interact with the local update server: 

> One drawback in the short term is that current Atom releases do not look for the ATOM_UPDATE_URL_PREFIX environment variable to override the update URL so one will either need to test with Nightly builds (after this merges) or make a build of Atom stable or beta that includes this change.

With that in mind, it sounds like we'll have a wait until after this PR is merged before we can verify that Atom is able to successfully interact with the local update server. If that's not the case, please let me know.

daviwil (commented):
There is a way to test that Atom interacts with the test update server correctly before this PR is merged, it's just a bit of extra work:

1. Follow the README instructions to run a local, test-signed build of Atom from this branch with `ATOM_RELEASE_VERSION="1.38.0-beta0`
2. Copy the resulting `Atom Beta.app` into `Applications`
3. Run the local test-signed build again with `ATOM_RELEASE_VERSION="1.38.0-beta1`
4. Start the test update server using the README instructions
5. Launch the `Atom Beta.app` that you moved to `Applications` with the `ATOM_UPDATE_URL_PREFIX="http://localhost:3456"` environment variable applied
6. Verify that the Atom Beta build finds and installs the update

PR #18134, opened 2018-09-27
Fix #18118 by reverting PR #17774: "Write config file atomically"

### LINKED ISSUE #18118: Atom doesn't remember changed settings
### Prerequisites

### Description

Atom doesn't remeber or store changed settings

### Steps to Reproduce

1. Change any editor setting
2. Close the application
3. Open application and check the setting

**Expected behavior:** The setting should be remembered

**Actual behavior:** Atom did what asked untill session ended, on the new session old settings are loaded

### Versions

Atom    : 1.32.0-beta0
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3

### Additional Information

Noticed since 1.31.0-beta1. As a long-term beta user, had to switch to a stable version 1.30. Recently 1.31 became a new stable, and after upgrade I noticed the the error persist. Also in brand new 1.32.0-beta0. A searched through issues, can't beleive I'm the only one affected, so it could be a duplication, but I was unable to find any similar issue.
It's Debian.
I tried to remove .atom and .config/Atom

daviwil:
### Identify the Bug

Fixes https://github.com/atom/atom/issues/18118.

### Description of the Change

This change reverts the commit from PR #17774 which caused a new issue in Atom 1.31.0+ where settings changed in the Settings View UI never get persisted to the user's `config.cson` file.  Reverting this change resolves this issue.

### Alternate Designs

A better approach would be to fix the underlying issue introduced by this PR, but we're reverting for now so that we can ship hotfixes to unblock users affected by this issue.

### Possible Drawbacks

Reverting this change could cause the [original issue](https://github.com/atom/atom/issues/17060) that was meant to be fixed by this PR  to reappear, but since that issue only appears when configuration save is invoked in a package's `deactivate` method, we're considering that less impactful than the issue reported in #18118.

### Verification Process

- [x] Reproduced the issue reported in #18118 on Linux with the official 1.31.0 release
- [x] Reproduced the issue in the latest 1.33.0-dev build
- [x] Verified that the issue is resolved after reverting PR #17774 and trying a new dev build

### DIFF
--- a/src/config-file.js
+++ b/src/config-file.js
@@ -6,7 +6,6 @@ const {watchPath} = require('./path-watcher')
 const CSON = require('season')
 const Path = require('path')
 const async = require('async')
-const temp = require('temp')
 
 const EVENT_TYPES = new Set([
   'created',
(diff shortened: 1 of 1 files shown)


jasonrudolph (approved):
:bow:

PR #18159, opened 2018-09-29
Fix atom.desktop for some Linux systems

shadow-light:
For some linux systems the window manager does not associate atom.desktop with Atom processes unless the class name is in lowercase. See #16791


daviwil:
Thanks @shadow-light!  This change looks fine to me.  Out of curiosity, which distro and desktop environment are you using?

shadow-light:
:+1: I'm using Ubuntu 16.04 which uses Unity.

PR #18189, opened 2018-10-03
Add the text "It also let us to teletype so that we can work together…

silentstorm29:
… online easily." to the article.

⚛👋 Hello there! Welcome. Please follow the steps below to tell us about your contribution.

1. Copy the correct template for your contribution
  - 🐛 Are you fixing a bug? Copy the template from https://bit.ly/atom-bugfix
  - 📈 Are you improving performance? Copy the template from https://bit.ly/atom-perf
  - 📝 Are you updating documentation? Copy the template from https://bit.ly/atom-docs
  - 💻 Are you changing functionality? Copy the template from https://bit.ly/atom-behavior
2. Replace this text with the contents of the template
3. Fill in all sections of the template
4. Click "Create pull request"

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,4 @@ repeat these steps to upgrade to future releases.
 [MIT](https://github.com/atom/atom/blob/master/LICENSE.md)
 
 When using the Atom or other GitHub logos, be sure to follow the [GitHub logo guidelines](https://github.com/logos).
+It also let us to teletype so that we can work together online easily.


lee-dohm:
Thanks for the contribution.

While adding a bit with more detail about Teletype to the [Atom Flight Manual](https://github.com/atom/flight-manual.atom.io) would be welcome, this change doesn't really fit in the Atom README. Additionally, we do require the instructions in the template to be followed.

PR #18208, opened 2018-10-07
:pencil: logo and buttons aligned to centre

philson-philip:
### Description of the Change

The Logo and buttons below the logo in README.md was left aligned and was looking ugly. I have aligned both logo and buttons to the centre and now it is good looking like a professional repo. If you like the change merge my PR


lee-dohm:
Thanks for the suggestion.

For maintenance reasons, we'd like to keep the Markdown-style links for the image and badges. I agree your change does look nice though.

On the other hand, in the future, you may want to avoid implications that people aren't being professional before you come along. These kinds of comments are against the [Atom Code of Conduct](https://github.com/atom/atom/blob/master/CODE_OF_CONDUCT.md).

Thanks for helping out!

PR #18254, opened 2018-10-17
Add a command to log the syntax tree

Ben3eeE:

### Description of the Change

This adds back the functionality that the `editor:log-cursor-scope` command used to provide when tree-sitter was used. The command will run default back to `editor:log-cursor-scope` if text mate is used for the current editor.

This is because the command is very useful when developing grammars 🙂 

@maxbrunsfeld Opening this as a PR to
* See if you are ok with adding back this functionality with a new command name
* Ask what we should do when it's run on editors with textmate language mode
* Ask if this implementation looks ok
* Ask if I'm naming things in a way that makes sense 😅 

If all is 👍 I can write some tests for this functionality and fill out the rest of the template to get this PR merged.

### Verification Process

- [x] Run the command `editor:log-cursor-syntax-tree-scope` in a JavaScript file with tree-sitter enabled
  - [x] Expected: Show a notification with the syntax tree scope descriptor
- [x] Run the command `editor:log-cursor-syntax-tree-scope` in a JavaScript file with tree-sitter disabled
  - [x] Expected: Show a notification with the scope descriptor
  - [x] Expected: Same ScopeDescriptor is in a notification when also running the `editor:log-cursor-scope` command
- [x] Run the command `editor:log-cursor-syntax-tree-scope` in a CoffeScript file with tree-sitter enabled
  - [x] Expected: Show a notification with the scope descriptor
  - [x] Expected: Same ScopeDescriptor is in a notification when also running the `editor:log-cursor-scope` command
- [x] Run the command `editor:log-cursor-syntax-tree-scope` in a CoffeScript file with tree-sitter disabled
  - [x] Expected: Show a notification with the scope descriptor
  - [x] Expected: Same ScopeDescriptor is in a notification when also running the `editor:log-cursor-scope` command
- [x] Bind the command in your keymap and check that it works when running it from a keybinding
- [x] Passing tests

CoffeScript is a language that currently does not have tree-sitter syntax highlighting

Keymap used for testing the command from a keymap:
```
'atom-text-editor:not([mini])':
  'ctrl-shift-g': 'editor:log-cursor-syntax-tree-scope'
```

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -198,11 +198,11 @@ class Cursor extends Model {
     return this.editor.scopeDescriptorForBufferPosition(this.getBufferPosition())
   }
 
-  // Public: Retrieves the syntax tree descriptor for the cursor's current position.
+  // Public: Retrieves the syntax tree scope descriptor for the cursor's current position.
   //
   // Returns a {ScopeDescriptor}
-  getSyntaxTree () {
-    return this.editor.syntaxTreeForBufferPosition(this.getBufferPosition())
+  getSyntaxTreeScopeDescriptor () {
+    return this.editor.syntaxTreeScopeDescriptorForBufferPosition(this.getBufferPosition())
   }
 
   // Public: Returns true if this cursor has no non-whitespace characters before
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -198,6 +198,13 @@ class Cursor extends Model {
     return this.editor.scopeDescriptorForBufferPosition(this.getBufferPosition())
   }
 
+  // Public: Retrieves the syntax tree descriptor for the cursor's current position.
+  //
+  // Returns a {ScopeDescriptor}
+  getSyntaxTree () {
+    return this.editor.syntaxTreeForBufferPosition(this.getBufferPosition())
+  }
+
   // Public: Returns true if this cursor has no non-whitespace characters before
   // its current position.
   hasPrecedingCharactersOnLine () {
--- a/src/register-default-commands.coffee
+++ b/src/register-default-commands.coffee
@@ -273,7 +273,7 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
         @foldAllAtIndentLevel(8)
         @scrollToCursorPosition()
       'editor:log-cursor-scope': -> showCursorScope(@getCursorScope(), notificationManager)
-      'editor:log-cursor-syntax-tree': -> showSyntaxTree(@getSyntaxTree(), notificationManager)
+      'editor:log-cursor-syntax-tree-scope': -> showSyntaxTree(@getCursorSyntaxTreeScope(), notificationManager)
       'editor:copy-path': -> copyPathToClipboard(this, project, clipboard, false)
       'editor:copy-project-path': -> copyPathToClipboard(this, project, clipboard, true)
       'editor:toggle-indent-guide': -> config.set('editor.showIndentGuide', not config.get('editor.showIndentGuide'))
--- a/src/register-default-commands.coffee
+++ b/src/register-default-commands.coffee
@@ -273,6 +273,7 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
         @foldAllAtIndentLevel(8)
         @scrollToCursorPosition()
       'editor:log-cursor-scope': -> showCursorScope(@getCursorScope(), notificationManager)
+      'editor:log-cursor-syntax-tree': -> showSyntaxTree(@getSyntaxTree(), notificationManager)
       'editor:copy-path': -> copyPathToClipboard(this, project, clipboard, false)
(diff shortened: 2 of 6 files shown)


src/cursor.js
```diff
@@ -198,6 +198,13 @@ class Cursor extends Model {
     return this.editor.scopeDescriptorForBufferPosition(this.getBufferPosition())
   }
 
+  // Public: Retrieves the syntax tree descriptor for the cursor's current position.
+  //
+  // Returns a {ScopeDescriptor}
+  getSyntaxTree () {
```
maxbrunsfeld: Regarding naming - I think `getSyntaxTree` makes it sound like you are retrieving the entire tree. Since we're returning a scope descriptor, which describes a *path* from the root of the tree to the current node, how about something like:
* `getSyntaxTreePath`
* `getSyntaxTreePathDescriptor`
* `getSyntaxTreeScopeDescriptor`

Thoughts?

src/register-default-commands.coffee
```diff
@@ -273,6 +273,7 @@ module.exports = ({commandRegistry, commandInstaller, config, notificationManage
         @foldAllAtIndentLevel(8)
         @scrollToCursorPosition()
       'editor:log-cursor-scope': -> showCursorScope(@getCursorScope(), notificationManager)
+      'editor:log-cursor-syntax-tree': -> showSyntaxTree(@getSyntaxTree(), notificationManager)
```
maxbrunsfeld: If we rename `getSyntaxTree`, we should rename the command as well.

src/text-editor.js
```diff
@@ -3850,6 +3850,28 @@ class TextEditor {
...
+  // `["source.js", "program", "expression_statement", "assignment_expression", "class", "class_body", "method_definition", "formal_parameters", "identifier"]`
+  // if tree-sitter is used
+  // and the following scopes array:
+  // `["source.js"]`
+  // if textmate is used
+  //
+  // * `bufferPosition` A {Point} or {Array} of `[row, column]`.
+  //
+  // Returns a {ScopeDescriptor}.
+  syntaxTreeForBufferPosition (bufferPosition) {
```
maxbrunsfeld: We should adjust the name of this one too.

src/text-editor.js
```diff
@@ -3881,6 +3903,11 @@ class TextEditor {
     return this.getLastCursor().getScopeDescriptor()
   }
 
+  // Get the syntax nodes at the cursor.
+  getSyntaxTree () {
```
maxbrunsfeld: This method name should probably start with `getCursor`, to make it clear that the positioning is derived from the cursor. Maybe `getCursorSyntaxTreePath`.

src/tree-sitter-language-mode.js
```diff
@@ -429,6 +429,45 @@ class TreeSitterLanguageMode {
...
+        while (i > 0 && compareScopeDescriptorIterators(iterator, iterators[i - 1]) < 0) i--
+        if (i < length - 1) iterators.splice(i, 0, iterators.pop())
+      } else {
+        iterators.pop()
+      }
+    }
+
+    scopes.push(this.grammar.scopeName)
+    return new ScopeDescriptor({scopes: scopes.reverse()})
+  }
```
maxbrunsfeld: I think this could be simplified a bit now. We don't necessarily need to exclude anonymous nodes like `(` because we're not using the output in `getScopeDescriptor`.

I think something like this should work:

```js
const nodes = []
this._forEachTreeWithRange(new Range(point, point), tree => {
  let node = tree.rootNode.descendantForPosition(point)
  while (node) {
    nodes.push(node)
    node = node.parent
  }
})

// The nodes are mostly already sorted from smallest to largest,
// but for files with multiple syntax trees (e.g. ERB), each tree's
// nodes are separate. Sort the nodes from largest to smallest.
nodes.reverse()
nodes.sort((a, b) =>
  a.startIndex - b.startIndex || b.endIndex - a.endIndex
)

const nodeTypes = nodes.map(node => node.type)
return new ScopeDescriptor({scopes: nodeTypes})
```

maxbrunsfeld (changes requested):
I left a few comments. I'm not 100% sure on the naming so I proposed a few options.

src/cursor.js
Ben3eeE: `getSyntaxTreeScopeDescriptor` sounds the best to me considering the other function is `getScopeDescriptor` and we return a `ScopeDescriptor`. I will use this one!

src/tree-sitter-language-mode.js
Ben3eeE: Used exactly this but I added in the `scopeName` in the beginning of the `ScopeDescriptor` and I also added back the `point = Point.fromObject(point)` line.

maxbrunsfeld: Awesome, thanks for taking my partially-working sketch and figuring out how to make it fully correct.

maxbrunsfeld (approved):
Looks great!

Issue #18174, opened 2018-10-02
Remove `!important` from styles


daviwil:
_From @dsandstrom on June 12, 2017 17:28_

Makes it more of a pain to override the background color.

https://github.com/atom/deprecation-cop/blob/master/styles/deprecation-cop.less#L16

_Copied from original issue: atom/deprecation-cop#82_

RyanPisuena:
'This repository has been archived by the owner. It is now read-only.' When attempting to submit a pull request.

Arcanemagus:
@RyanPisuena As explained in the archived repository's README, the code for that now lives here: https://github.com/atom/atom/tree/master/packages/deprecation-cop

RyanPisuena:
Issue can now be closed per #18266

thomasjo:
@RyanPisuena Thanks! :tada:

PR #18266, opened 2018-10-19
Update deprecation-cop.less

RyanPisuena:
### Description of the Change

Removing '!important argument from background-color of deprecation-cop'

According to #18174 , removing !important makes changing background color changes easier

### Possible Drawbacks

May mess with current background color settings

### Verification Process

Did not verify

### Applicable Issues

N/A

### DIFF
--- a/packages/deprecation-cop/styles/deprecation-cop.less
+++ b/packages/deprecation-cop/styles/deprecation-cop.less
@@ -13,7 +13,7 @@
 .deprecation-cop {
   overflow: auto;
   -webkit-flex: 1;
-  background-color: @app-background-color !important;
+  background-color: @app-background-color;
 
   .deprecation-overview {    
     &:after {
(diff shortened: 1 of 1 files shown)


simurai (changes requested):
👍  on removing `!important`.

Maybe `!important` could also be removed when hovering the `.deprecation-info`?

https://github.com/atom/atom/blob/af4f144237ab3044d979d5faad83668e622d579d/packages/deprecation-cop/styles/deprecation-cop.less#L31

simurai (changes requested):
👍  on removing `!important`.

Maybe `!important` could also be removed when hovering the `.deprecation-info`?

https://github.com/atom/atom/blob/af4f144237ab3044d979d5faad83668e622d579d/packages/deprecation-cop/styles/deprecation-cop.less#L31

RyanPisuena:
@simurai changes made

simurai (approved):

simurai:
CI failure seems unrelated (timeout), merging.

Thanks @RyanPisuena 🙇

PR #18340, opened 2018-10-26
maintainGrammar -> maintainLanguageMode

Ben3eeE:
Not sure if this will work. But it seems that `maintainGrammar` doesn't exist and was replaced with `maintainLanguageMode` and that this was never changed.

/cc: @maxbrunsfeld


maxbrunsfeld (approved):
✨

PR #18354, opened 2018-10-29
Add "Release Notes" section to PR templates

daviwil:
### Description of the Change

This change adds a "Release Notes" section to our PR templates so that contributors can help our release process by adding appropriate release notes for the change they are contributing.

This section of the PR template may be targeted by automation in the future, both to prevent PRs without Release Notes from being merged and also to consume these notes for changelog generation.


jasonrudolph (approved):
I'm a bit hesitant to add anything new to the pull request templates unless it replaces something else, because we already struggle with getting people to thoughtfully complete the existing template. That said, let's see how useful this ends up being in practice. If it can earn its keep in the templates, great! If not, we should revisit. 😅

daviwil:
Totally fine with removing this later if it doesn't work out, but I think it'll be a boon to our release automation if we can make it a standard part of the workflow.  Thanks for the review! :bowing_man:

PR #18375, opened 2018-10-31
Fix error causing tree-sitter highlighting to get out-of-sync when editing rapidly

maxbrunsfeld:
### Identify the Bug

When using Tree-sitter, it was previously possible to get the buffer and the syntax tree into an inconsistent state by making many rapid changes to the buffer.

See also:
* https://twitter.com/jetzajac/status/1054895895760773120
* https://github.com/atom/language-python/issues/281#issuecomment-434021610

### Description of the Change

Whenever you make a change to a buffer that's using Tree-sitter, we start incrementally re-parsing the buffer. If that parse does *not* complete quickly, we allow it to finish parsing in the background. Then, if you make additional edits while that parse is still going, we need to store those edits so that we can apply them after-the-fact to the new syntax tree.

We store the edits in a `Patch` in order to group the changes efficiently. Unfortunately, the code that was mutating that patch was passing the wrong values to `Patch.splice`. I'm not sure how this huge logic error didn't cause more problems than it did!

### Verification Process

The easiest way I found to reproduce the original bug was to open a large JavaScript file, place the cursor somewhere near the top of the file, and hold down `command-x` (`Editor: Delete Line`). Previously, you would periodically get broken syntax highlighting.

Now, with this change, I'm not seeing this problem.

### Release Notes

* Fixed an issue where making many rapid edits could cause incorrect syntax highlighting to be rendered.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5529,9 +5529,9 @@
       "integrity": "sha512-DlX6dR0lOIRDFxI0mjL9IYg6OTncLm/Zt+JiBhE5OlFcAR8yc9S7FFXU9so0oda47frdM/JFsk7UjNt9vscKcg=="
     },
     "tree-sitter": {
-      "version": "0.13.15",
-      "resolved": "https://registry.npmjs.org/tree-sitter/-/tree-sitter-0.13.15.tgz",
-      "integrity": "sha512-FuvU+csO7t/rQqLdL3+w4Jg+4Zl22Y4uCi4L9X/qJG57Zn71ZzP3oHtDSRgpiIms6g3Y7cEJvF7K/rCw11q92Q==",
+      "version": "0.13.16",
+      "resolved": "https://registry.npmjs.org/tree-sitter/-/tree-sitter-0.13.16.tgz",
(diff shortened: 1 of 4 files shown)


Ben3eeE:
I was still able to reproduce the issue with syntax highlighting breaking and Atom freezing that I saw after building Atom on this commit.

I saw syntax highlighting break when pasting a 42 lines long erb file holding `ctrl-v` until it pasted to a couple of thousand lines.
I got atom to freeze by pasting a 7000 line erb file and undoing rapidly.

I send the files I used on Slack.

maxbrunsfeld:
> I was still able to reproduce the issue with syntax highlighting breaking and Atom freezing that I saw after building Atom on this commit.

Ok, thanks for testing that. I think that there's a separate issue relating to ERB specifically. I'll do a little more testing to make sure that it is actually ERB-specific.

as-cii:
> I think this is the bug that you mentioned to me the other day.

Yes, it seems quite possible.

> Gonna take your suggestion of using a randomized test to cover this.

Nice! ✨

Ben3eeE:
@maxbrunsfeld Just wanna say I was able to reproduce the issue that this PR fixes to confirm the fix.

I could reproduce it by just typing fast at the top of `text-editor-component`. On this branch it did not reproduce again.

ERB files seem like a different issue 👍

PR #18382, opened 2018-11-01
Distinguish between anonymous * token and wildcard selector

Ben3eeE:
Paired with @maxbrunsfeld 

### Identify the Bug

An anonymous token `"*"` is not distinguished with the wildcard selector `*`. So if a tree-sitter grammar has a scope mapping like this:

```
'"*"': 'keyword.operator'
'relational_expression > ">"': 'keyword.operator'
```

It will scope any `>` as `keyword.operator`

### Description of the Change

This changes `setTableDefaults` to add a parameter for allowing wildcard selectors so we can disallow it when mapping the anonymous nodes.

### Verification Process

* This PR https://github.com/atom/language-c/pull/303 adds scope mappings for `relational_expression > ">"` and `"*"` to be `keyword.operator`. Without this change tree-sitter will still scope `>` and `<` in templates as `keyword.operator`
* Passing tests

### Release Notes

### DIFF
--- a/spec/syntax-scope-map-spec.js
+++ b/spec/syntax-scope-map-spec.js
@@ -74,4 +74,14 @@ describe('SyntaxScopeMap', () => {
     expect(map.get(['a', 'c', 'b'], [0, 1, 1])).toBe('z')
     expect(map.get(['a', 'c', 'b'], [0, 2, 1])).toBe('w')
   })
+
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I'd say we don't need a release note for this one, because its only user-facing impact is via https://github.com/atom/language-c/pull/303 - basically fixing a bug with how `<` and `>` were highlighted.

PR #18394, opened 2018-11-04
fixed doc comment js syntax in project::onDidChangeFiles

jneuendorf:
### Requirements for Contributing Documentation

* The pull request must only contribute documentation (for example, markdown files or API docs). To contribute other changes, you must use a different template. You can see all templates at https://github.com/atom/atom/tree/master/.github/PULL_REQUEST_TEMPLATE.

### Description of the Change

In https://github.com/atom/atom/blob/44b7c1dd97f4b51de149c92993251af28084eb73/src/project.js#L199-L211 there currently is a closing parenthesis missing.
This PR fixed the syntax error.

Since (I assume) the [API docs page](https://atom.io/docs/api/v1.32.1/Project#instance-onDidChangeFiles) is generated from the docstring the syntax should be correct, since people likely copy the source code.

### Release Notes

- Fixed the JavaScript syntax in the example of `Project::onDidChangeFiles`

### DIFF
--- a/src/project.js
+++ b/src/project.js
@@ -208,7 +208,7 @@ class Project extends Model {
   //       console.log(`.. renamed from: ${event.oldPath}`)
   //     }
   //   }
-  // }
+  // })
   //
   // disposable.dispose()
   // ```


jneuendorf:
I just noticed that the example uses `event.type` at the end but `event.action` at the top. According to the docs `action` seems more correct.
If you want I would add a change to `event.action` to this pull request.

rsese:
Thanks! Someone from the team will take a look as soon as they can.

smashwilson:
Oh! Excellent catch. And yes, `event.action` is correct - ❤️ if you have time to fix that before I re-run the tests and look at it again :smile:

smashwilson:
... Oh, the build aged off, so I can't re-run it.

I'll touch up and merge by hand to override it 👌

jneuendorf:
Awesome, looks good - thank you 👍

Issue #18342, opened 2018-10-26
Atom freezing when editing ERB files with tree-sitter


Ben3eeE:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

https://github.com/atom/atom/issues/18327#issuecomment-433267915 wrote:

> I've experienced issues with syntax highlighting in ERB files since updating to 1.32.0. I don't think I can reproduce intentionally, but it did occur to me several times today where syntax highlighting will break during edits and eventually Atom will freeze with high CPU usage for about 10 seconds at a time. It doesn't seem to resolve itself unless I close the file and reopen it.

### Steps to Reproduce

* Open [this ERB file](https://gist.github.com/Ben3eeE/d5c285849a7b86d51cbb8b6ddd266f43)
* Select all
* Cut
* Repeat:
  * Paste
  * Undo

**Expected behavior:** Not freezing/incorrect highlighting 

**Actual behavior:** Atom freezing/incorrect highlighting

**Reproduces how often:** Unknown

### Versions

1.32.0

### Additional Information

/cc: @pjrebsch

Ben3eeE:
@maxbrunsfeld Can you provide some details about how you reproduced this issue to see if we can get some reliable repro steps? 🙇

pjrebsch:
Since adding `useTreeSitterParsers: false`, it hasn't happened to me again. I'll re-enable the feature to see how it might be reproducible.

turgs:
I have the same problem. I'm using Ubuntu 18.10. Atom freezes then crashes. As it freezes, I always have an ERB file open and I can see the syntax highlighting is messed up similar to #18327. I'm going to try set `useTreeSitterParsers: false` in the interim.

pjrebsch:
[...]

PR #18411, opened 2018-11-08
Fix failure to emit did-change-highlighting event in TreeSitterLanguageMode

maxbrunsfeld:
Fixes https://github.com/atom/atom/issues/18342


Ben3eeE:
I can confirm that this fixes https://github.com/atom/atom/issues/18342#issuecomment-436295896

But I can still reproduce the `Overlapping ranges` exception though even on this branch.

PR #18469, opened 2018-11-20
Adjust cursor scope when at end of line

### LINKED ISSUE #1006: Disable autocomplete in comments and strings
It is incredibly irritating to have the autocomplete pop up for every key stroke while I'm typing a simple string inside quotation. It is free human text I want to type, I don't need it autocompleted.

It should be a simple setting, <Turn off for text & comments>, I don't need to be an Atom expert to write my own providers and search for scope names in order to get this turned off.

### LINKED ISSUE #18463: Scope of cursor at end of line
### Description

The scope of the cursor is taken to be of the character to the right of it. This works most of the time, but can have unexpected behaviour when the cursor is at the end of a line and scopes do not extend to the newline character.

E.g., in [this issue](https://github.com/atom/autocomplete-plus/issues/1006) the TS parser does not count the newline as part of the comment, so autocomplete suggestions were not suppressed when typing at the end of the line. I raised [this issue](https://github.com/tree-sitter/tree-sitter-javascript/issues/87) to suggest a fix, but decided this is more general than comments.

I think Atom should in general use the left character if the next is a newline. This behaviour matches the TextMate grammars, but I'm not sure if that's because they apply the scope to the newline character or because they already do this.

It potentially makes the following confusing, but it also makes sense in a way (and matches TextMate behaviour, so not really confusing at all). Here, `|` is the cursor (not a character), and the scopes are how TextMate handles it
```js
// End of line: 'source string' 
"foo"|

// Not end of line: 'source'
"foo"| bar 
```

/cc @maxbrunsfeld This falls under Tree-sitter, and I've made my argument.

Aerijo:

### Identify the Bug

Closes https://github.com/atom/atom/issues/18463
Fixes https://github.com/atom/autocomplete-plus/issues/1006

### Description of the Change

When the cursor is at the end of a non-empty line, it's effective position will be shifted back one column for Tree-sitter grammars (for both scope and node queries).

This is to match the existing TextMate behaviour, which extended the scopes of the last character to the newline character (whether this was intentional IDK).

### Alternate Designs

I considered adding an options object that lets you specify if the newline should be used or not. I decided not to make it for this though, because passing it around started getting more complicated than I wanted.

I also considered implementing it in the text editor API, so the "raw" behaviour could still be accessed using the language mode API, but the position is not necessarily a Point yet at that stage. 

### Possible Drawbacks

If a user really wants to inspect the scope of a newline for a TS grammar, they will be unable to. This has always been true for TextMate grammars, but may be desirable for adoptors of Tree-sitter grammars? I don't know any convincing usecase, and we can add in that options object if someone raises an issue about it.

Behaviour when the row length is 0 and the column number is > 0 will make the column -1, which is sort of bad, but I figured behaviour there is undefined anyway.

### Verification Process

Manually tested, and automatic tests.

### Release Notes

- Adjust cursor scope to left character when at end of line

### DIFF
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -479,6 +479,10 @@ class TreeSitterLanguageMode {
 
   scopeDescriptorForPosition (point) {
     point = Point.fromObject(point)
+
+    if (point.column > 0)
+      point.column = Math.min(point.column, this.buffer.lineLengthForRow(point.row) - 1)
+
     const iterator = this.buildHighlightIterator()
     const scopes = []
     for (const scope of iterator.seek(point, point.row + 1)) {
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -457,7 +457,7 @@ class TreeSitterLanguageMode {
     const nodes = []
     point = Point.fromObject(point)
 
-    if (point.column > 0 && point.column == this.buffer.lineLengthForRow(point.row)) {
+    if (point.column > 0 && point.column === this.buffer.lineLengthForRow(point.row)) {
       point.column--
     }
 
@@ -485,7 +485,7 @@ class TreeSitterLanguageMode {
   scopeDescriptorForPosition (point) {
     point = Point.fromObject(point)
 
-    if (point.column > 0 && point.column == this.buffer.lineLengthForRow(point.row)) {
+    if (point.column > 0 && point.column === this.buffer.lineLengthForRow(point.row)) {
       point.column--
     }
 
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -453,9 +453,13 @@ class TreeSitterLanguageMode {
     })
   }
 
-  syntaxTreeScopeDescriptorForPosition (point) {
+  syntaxTreeScopeDescriptorForPosition (point, {includeNewline=false}) {
     const nodes = []
     point = Point.fromObject(point)
+
+    if (!includeNewline && point.column > 0)
+      point.column = Math.min(point.column, this.buffer.lineLengthForRow(point.row) - 1)
+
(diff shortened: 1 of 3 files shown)


maxbrunsfeld (approved):

src/tree-sitter-language-mode.js
```diff
@@ -456,6 +456,11 @@ class TreeSitterLanguageMode {
   syntaxTreeScopeDescriptorForPosition (point) {
     const nodes = []
     point = Point.fromObject(point)
+
+    if (point.column > 0 && point.column === this.buffer.lineLengthForRow(point.row)) {
+      point.column--
+    }
```
maxbrunsfeld: Since we're mutating this point, we should pass `true` for [the `copy` parameter to `Point.fromObject`](https://github.com/atom/text-buffer/blob/d10f7595e600c775bd8bf505d11d5ba7a02a1290/src/point.coffee#L36-L45) above, to make sure we're not mutating a point that might get reused after the call to `scopeDescriptorForPosition`.

src/tree-sitter-language-mode.js
```diff
@@ -479,6 +484,11 @@ class TreeSitterLanguageMode {
 
   scopeDescriptorForPosition (point) {
     point = Point.fromObject(point)
+
+    if (point.column > 0 && point.column === this.buffer.lineLengthForRow(point.row)) {
+      point.column--
+    }
```
maxbrunsfeld: Same suggestion here as above, regarding mutating the point.

maxbrunsfeld:
Oops, thought of a change suggestion after reviewing.

Aerijo:
I added the comments explaining what it does and why. Instead of always making a copy, I only make one if we are going to change the column. 

I also made it pass through the `clipPosition`, which is what the TextMate version does first. This should mean any undefined behaviour is handled consistently.

maxbrunsfeld:
Looks great. 🚢

PR #18471, opened 2018-11-21
Toggle comment cursor placement

Aerijo:

https://github.com/atom/atom/pull/17519 (made this new one because I'd deleted the old PR branch)

### Description of the Change

As in the above PR, this changes selection and cursor behaviour when toggling block comments.

### Alternate Designs

An option to do this at the same place that changes the range in the first place would be better, but I don't know how to go about that.

### Possible Drawbacks

It makes the behaviour more complicated, and could have a performance decrease. The behaviour is toggleable though, at the TextEditor level (I did not pass the options received by the Selection method though).

### Verification Process

Manual testing, with all the scenarios laid out in the linked PR.

Also tested with indentation.

### Release Notes

- More intuitive cursor positioning when toggling block comments


src/text-editor.js
```diff
@@ -4794,6 +4794,24 @@ class TextEditor {
...
+                let endCol = endLineLength - commentEndString.length - 1
+                options.selection.setBufferRange([[end, endCol], [end, endCol]], {autoscroll: false})
+                return
+              }
+            } else {
+              let startDelta = oldRange.start.column === indentLength ? [0, commentStartString.length + 1] : [0, 0]
+              let endDelta = oldRange.end.column === endLineLength ? [0, -commentEndString.length - 1] : [0, 0]
+              options.selection.setBufferRange(oldRange.translate(startDelta, endDelta), {autoscroll: false})
+            }
+          }
```
thomasjo: All of the `let` statements above can/should be `const`.

src/text-editor.js
```diff
@@ -4794,6 +4794,24 @@ class TextEditor {
...
+          // Prevent the cursor from selecting / passing the delimiters
+          // See https://github.com/atom/atom/pull/17519
+          if (options.correctSelection && options.selection) {
+            let endLineLength = this.buffer.lineLengthForRow(end)
+            let oldRange = options.selection.getBufferRange()
+            if (oldRange.isEmpty()) {
+              if (oldRange.start.column === endLineLength) {
+                let endCol = endLineLength - commentEndString.length - 1
+                options.selection.setBufferRange([[end, endCol], [end, endCol]], {autoscroll: false})
+                return
```
thomasjo: Maybe I'm missing something, but as far as I can tell this `return` statement is unnecessary.

Aerijo:
This does change functionality for pressing `tab` when text is selected (the failing test). I'll have to remove the `exclusive` flag, which is a little disappointing. Unless that can be special cased? I still think `exclusive` fits better, but I don't know what else it might break.

Aerijo:
There is one undesired behaviour now (that I know of), for three selections where the first goes right to left on the leftmost character (line 13 in first GIF).

It currently will select the start delim, but only when there are three or more of them. It's weird.

Aerijo:
FYI @maxbrunsfeld [This test](https://ci.appveyor.com/project/Atom/atom/builds/20463609/job/riruaptl9gl6taym#L107) seems to have randomly failed.

Aerijo:
@lee-dohm (I moved to a different PR)

I'm pretty happy with the behaviour now. There are technically still more cases that can be tested, but at the same time I don't think we can exhaustively cover them all. The behaviour, except for as note about line 13, is as in the GIFs above.

spec/text-editor-spec.js
```diff
@@ -7314,6 +7314,41 @@ describe('TextEditor', () => {
         editor.toggleLineCommentsForBufferRows(0, 0)
         expect(editor.lineTextForBufferRow(0)).toBe('test')
       })
+
+      it('does not select the new delimiters'), () => {
```
nathansobo: Hi. Unfortunately, the closing parentheses after the description string in this test means that the block following the test isn't running at all. So the test is green only because it isn't really running. I'll try and get the test passing, but I may kick this back to you if it ends up being to much work.

nathansobo:
Okay, I merged in master and fixed a couple problems with the test once I got it actually running. I'd like to stare at this a bit longer and see if we can simplify it at all. Stay tuned.

nathansobo:
Alright. Your earlier comments about using `exclusive` prompted some investigation. As you also seemed to discover, the `exclusive` option on the selection's marker works great unless the selection is empty. In that case, the cursor is still pushed to the right when commenting at the end of the line. We really need a mechanism like a `bias` that causes the endpoints of a marker to prefer to stay left or move right when there is an insertion at their location. It's something we had in Xray. But for the purposes of fixing this issue, I think your approach is reasonable. Just gonna wait on a green build.

nathansobo:
Oh jeez I left a test focused. Rebuilding 🙄.

PR #18484, opened 2018-11-23
Migrate 'atom-ui' into core

simurai:
> ⚠️ Wait with merging until the next release is out, so we can test this a bit longer.

### Description of the Change

This PR moves [atom-ui](https://github.com/atom/atom-ui) back into `static/`. In addition, core styles and icons got their own folders.

- [x] Move `atom-ui` styles
- [x] Remove `atom-ui` dependency
- [x] Move core styles to `static/core-ui/`
- [x] Move icons to `static/icons/`

After this PR is merged.

- [x] Archive the [atom/atom-ui](https://github.com/atom/atom-ui) repo.

### Benefits

It's part of the package migration initiative.

Originally `atom-ui` got moved to its own repo in hopes of increased community contribution. Unfortunately that didn't happen. Also, once the `styleguide` package gets migrated, it will be easier to make updates since they are both part of the same repo.

### Possible Drawbacks

There is a chance for regressions.

### Verification Process

1. Open Atom in dev mode: `atom . -d`
2. Verify that the UI still looks the same.
3. Verify that there are no errors in the console.
4. Verify that the icons get rendered.

### Applicable Issues

N/A

### Release Notes

N/A


simurai:
The build keeps failing with:

```
Generating pre-built less cache in /Users/simurai/github/atom/out/app/less-compile-cache

/Users/simurai/github/atom/node_modules/less-cache/lib/less-cache.js:296
            throw error;
            ^
Error: variable @use-custom-controls is undefined
```

I think it's because all files found in `static/` get compiled "individually" and not based on the import paths. Like the `@use-custom-controls` gets imported first and would be available, but `less-cache` doesn't know about it.

I guess we have two options:

1. Move the `atom-ui` styles somewhere outside of `static/`. Then `less-cache` doesn't check them automatically.
2. Make sure that every `.less` file imports all the variables, even though they get already imported a before.

I might try option 2, since caching more styles sounds like it adds extra benefits to performance.

simurai:
Ok, this seems good. I'll merge now, so we still have till January to catch any regressions.

PR #18570, opened 2018-12-11
Only deref top-level symlinks in node_modules when copying assets

### LINKED ISSUE #18490: Atom installation contains 120+ identical executables in dugite-native package, wasting 270 MB disk space
*This issue is a continuation of #18452 and https://github.com/desktop/dugite-native/issues/138.*

*Edit by @rsese - see @shiftkey's comment at https://github.com/desktop/dugite-native/issues/138#issuecomment-438070892* - an excerpt:

> We tried ASAR ages ago and ended up just avoiding it (see https://github.com/desktop/desktop/pull/476 and the linked PR for more info), and I see someone added support for being able to control the symlink behaviour (to opt-out of the default behaviour) in https://github.com/electron-userland/electron-packager/pull/410.

> Ultimately, this is up to the packaging apps rather than something I have control over. I don't see a `derefSymlinks` flag being set in the [Atom packager config](https://github.com/atom/atom/blob/master/script/lib/package-application.js#L17-L38) so maybe it's worth asking over in [`atom/atom`](https://github.com/atom/atom) if they've tried `derefSymlinks` when packaging.

---

### Prerequisites

Doesn't apply. This is not a runtime issue.

### Description

The Git distribution bundled with dugite-native contains hundreds of symlinks used to create aliases such as `git-imap-send`. The way Atom is currently packaged emulates those symlinks by creating a full copy of the link target for each of them. This wastes 270 MB of disk space.

### Steps to Reproduce

1. Install Atom
2. Look inside `/node_modules/dugite/git/libexec/git-core` in the installation folder
3. Notice there are 136 executables totaling 300 MB, 121 of which are identical (with different names)

**Expected behavior:** The Atom installer/installation doesn't waste disk space and bandwidth like that.

**Actual behavior:** Disk space and download bandwidth is wasted.

### Versions

*Edit by @rsese to note specific releases where there was a large increase in the size of the .rpm size and `atom-mac.zip`*

For the .rpm, saw the 2x size increase from 1.30 where the .rpm was 134 MB and atom-mac.zip was 139 MB:

https://github.com/atom/atom/releases/tag/v1.30.0

And 1.31.0 where the .rpm size was 270 MB and atom-mac.zip was 272 MB:

https://github.com/atom/atom/releases/tag/v1.31.0

---

1.32.2, at least macOS and Linux are affected, but likely all platforms.

### Additional Information

None.

daviwil:
### Identify the Bug

Fixes #18490

### Description of the Change

This change fixes an issue that caused some of Atom's installation packages (like the .rpm package) to gain around 2x in size (135 MB to 270 MB) between Atom 1.30 and 1.31.

The core issue came from a [small change](https://github.com/atom/atom/commit/09b45911bc9a2f1195450c1760e00b704ab9623c#diff-b56263f1f1a47bfce6e83ca748aae979) in our packaging logic that caused all symlinks under `node_modules` to be dereferenced.  This was a big problem for `dugite` which includes a full Git binary distribution where many Git commands are symlinked back to the core `git` binary.  Since we dereferenced symlinks, that caused all of these command symlinks to turn into full-sized `git` binaries, bloating our distribution size greatly.

### Alternate Designs

Instead of dereferencing all top-level symlinks under Atom's `node_modules` folder , we could scan the folders under `/packages` and copy those directly into the output folder.  I ultimately thought this was less flexible and forward-looking than scanning symlinks in `node_modules`, but I'm happy to go with the other approach if people think it's necessary.

### Possible Drawbacks

With this logic, other non-package symlinked folders under `node_modules` will also be copied into the output folder if they exist, but this should only be a problem (or benefit?) when producing local Atom builds.

### Verification Process

- [x] Produce full build of Atom and ensure that `git-*` commands are still symlinked inside of:
  - [x] `out/app/node_modules/dugite/git/libexec/gitcore`
  - [x] `out/atom-*/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core`
- [x] Produce Atom .rpm package and ensure that it's once again in the neighborhood of 150MB
- [x] Produce release packages on all OSes to ensure that Git functionality works in the GitHub package
  - [x] Windows
  - [x] Linux
  - [x] macOS

### Release Notes

Fixed an issue causing some Atom release packages to double in size

### DIFF
--- a/script/lib/copy-assets.js
+++ b/script/lib/copy-assets.js
@@ -28,7 +28,7 @@ module.exports = function () {
 
   // Run a second copy pass for symlinked directories under node_modules.
   // We do this to ensure that symlinked repo-local bundled packages get
-  // copied to the output folder correctly.  We also copy only the top-level
+  // copied to the output folder correctly.  We dereference only the top-level
   // symlinks and not nested symlinks to avoid issues where symlinked binaries
   // are duplicated in Atom's installation packages (see atom/atom#18490).
   const nodeModulesPath = path.join(CONFIG.repositoryRootPath, 'node_modules')
@@ -36,9 +36,8 @@ module.exports = function () {
     .map(p => path.join(nodeModulesPath, p))
     .filter(p => fs.lstatSync(p).isSymbolicLink())
     .forEach(modulePath => {
+      // Replace the symlink that was copied already
       const destPath = path.join(CONFIG.intermediateAppPath, 'node_modules', path.basename(modulePath))
-
-      // Remove the symlink that was copied already
       fs.unlinkSync(destPath)
       fs.copySync(modulePath, destPath, { filter: includePathInPackagedApp, clobber: true })
     })
--- a/script/lib/copy-assets.js
+++ b/script/lib/copy-assets.js
@@ -15,7 +15,6 @@ module.exports = function () {
     path.join(CONFIG.repositoryRootPath, 'benchmarks', 'benchmark-runner.js'),
     path.join(CONFIG.repositoryRootPath, 'dot-atom'),
     path.join(CONFIG.repositoryRootPath, 'exports'),
-    path.join(CONFIG.repositoryRootPath, 'node_modules'),
     path.join(CONFIG.repositoryRootPath, 'package.json'),
     path.join(CONFIG.repositoryRootPath, 'static'),
     path.join(CONFIG.repositoryRootPath, 'src'),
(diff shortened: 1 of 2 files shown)


jasonrudolph (commented):
> * [x]  Produce Atom .rpm package and ensure that it's once again in the neighborhood of 150MB

Nice! 

Are there user-facing operations we should perform in Atom to verify that this change hasn't introduced any regressions?

daviwil:
Yep!  I'm gonna produce packaged builds on all 3 platforms and verify that Git functionality works in the GitHub package, will add that to the verification list.

PR #18651, opened 2019-01-04
Update PULL_REQUEST_TEMPLATE.md

mahi97:
### Description of the Change

add needed articles.


rsese:
Thanks for the contribution - we feel the current wording is sufficient so won't be accepting this change, but thank you again for your interest!

Additionally, please use the templates when opening issues and pull requests as detailed below.

---

Issue #5168, opened 2015-01-20
Per-project config settings


benogle:

wyand:
I'd love the ability to specify a configuration directory on the command line... the use case I have in mind is using a separate configuration when invoked via git's `core.editor` or other similar tools. Would also be nice to have the same option on `npm` so that packages for customized configuration can be managed with `npm`.

Is that a part of this feature? Should it be? Should this be a separate enhancement request?

idris:
:+1:

joefitzgerald:
:+1: in the form of core support for EditorConfig.

Willyham:
:+1: Lack of per-project settings when moving between projects with different tab spacing is a real pain.

dark-panda:
:+1 on the different tab spacing pain.

entr:
A must-have.

gpupo:
:+1: must-have feature

fru:
+1

sikhote:
Yes, please

umpirsky:
Would be useful in deed. :+1:

neuling:
+1

glen-84:
You can use [this extension](https://github.com/danielbrodin/atom-project-manager).

tschijnmo:
+1

[Atom Project Manager](https://github.com/danielbrodin/atom-project-manager) is definitely a good thing for this purpose.  However, it still requires the *projects* to be added to the user's configurations.  Maybe what could be better is some kind of project file that can be committed into version control in the source tree.  Then developers of the same project could do something like `atom an_awesome_project.cson` then configurations required by the project could be immediately ready.

dmoutal:
@tschijnmo Exactly!

Sublime Text allows you to place a project.json file within the project root folder with all the project specific settings. This file can become part of your source control so everyone has the same settings.

abadcafe:
+1

ZakariaRidouh:
:+1: this would actually be a very good idea!

rudineirk:
+1

pwnall:
:+1:

mehcode:
A project file isn't the core-level solution here. 

Atom understands a single layer of configuration right now. Atom needs to be able to understand layered configuration (2-layer but still multi-layer). 

The primary problem with existing solutions (eg. project manager) is they just modify the global configuration on project load. If you were to touch the global configuration at all (even just hiding the menu bar) it would actually commit the current configuration as the new global configuration. 

[...]

PR #18700, opened 2019-01-15
[wip] RFC: Session Management

smashwilson:
This is an RFC on a potential direction we could take in managing Atom's serialized state. My goals are to make it more explicit and intentional and to make it more difficult for users to "lose" the setup that they like or their unsaved buffer changes.

I'm thinking this out as I write it, so don't be surprised if there are contradictions or things that are still unclear.

Related to #16698 - I'd love to roll in a way to manage project-scoped config.

---

[Rendered](https://github.com/atom/atom/blob/aw/rfc/session-management/docs/rfcs/004-session-management.md)

### DIFF
--- a/docs/rfcs/004-session-management.md
+++ b/docs/rfcs/004-session-management.md
@@ -1,0 +1,112 @@
+# Improved Session Management
+
+## Status
+
+Proposed
+
+## Summary
+
+Atom's management of serialized project state is complicated and unintuitive. From [atom/atom#10517](https://github.com/atom/atom/issues/10517):
+
+> There are things that are serialized and stored when Atom exits. It is completely non-obvious what items belong to this set and what items do not as well as how to restore the saved items when next launching Atom. Decisions should be made about whether some of these items belong to the specific mode that Atom is in or if they are more global state that people expect to be in place the next time Atom is launched (or even an individual file is opened post-launch) no matter what mode Atom is launched in.
+
+Let's revisit the way we manage session state so that it becomes an Atom feature that users can understand and trust.
+
+## Motivation
+
+_pending_
+
+Problems to address:
+
+* The active project (which is deserialized on open, serialized on close) is chosen implicitly based on the set of project folders.
+* The active project isn't explicitly named anywhere in the UI. Two open Atom windows with two different projects do not obviously differ; two open Atom windows on the same project are not obviously the same.
+* Adding or removing a root folder silently changes the active project.
+* When the same project is open in two Atom windows at once, the last-closed window clobbers any project state saved by the first-closed.
+* Unsaved content in a TextBuffer is only accessible through the project that was open when it was created.
+
+## Explanation
+
+A _Project_ is the unit of persistent window state in Atom related to a coherent development effort. It includes:
+
+* The number of windows open;
+* The set of root folders;
+* Pane and dock configuration;
+* Open pane and dock items;
+* Cursor position, fold state, and other TextEditor characteristics;
+* Per-package [serialized state](https://flight-manual.atom.io/behind-atom/sections/serialization-in-atom/).
+
+Each Project has a unique, user-customizable name.
+
+Other persistent state is tracked _per-file_, based on its inode or fileID to work through filesystem softlinks or hardlinks. This includes:
+
+* Unsaved buffer changes;
+* Per-file manually changed indentation, line ending, and encoding settings.
+
+### Add an explicit indication of the current Project to the tree-view
+
+At the top of the tree-view, display the current Project's name. Clicking it opens the project management UI described below.
+
+> TODO: graphic
+
+### Add a command-line argument to explicitly choose the Project
+
+The `--project` or `-p` flags to the `atom` command-line script explicitly designate the name of a Project to which the newly opened paths should belong. If no Project with that name exists, a new one is created. If a window is already open on the named Project, the provided paths are added to the existing window.
+
+```sh
+# Open a new Atom window containing:
+# * the root folder "src/project-one"
+# * TextEditor on "${HOME}/writing/tps-report.txt"
+$ atom --project work src/project-one ~/writing/tps-report.txt
+
+# Open a new Atom window containing:
+# * the root folder "src/game"
+$ atom -p fun src/game
+
+# Add the root folder "src/project-three" to the existing "work" window
+$ atom -p work src/project-three
+
+# (close the "work" window)
+
+# Re-open a new Atom window with:
+# * root folders "src/project-one" and "src/project-three"
+# * TextEditor on "${HOME}/writing/tps-report.txt"
+$ atom -p work
+```
+
+It's mutually exclusive with the `--add` option (it is an error to provide both).
+
+### Create a new bundled package to list and manage serialized state
+
+Clicking the "project" tile in the tree-view opens a UI for listing, creating, loading, modifying, and deleting known projects. This is where projects can be renamed and managed.
+
+> TODO: graphic
+
+Loading a different Project immediately serializes the window's state into the previously active project and deserializes the window state from the chosen one into this window.
+
+### Create a new bundled package to list and manage unsaved buffers
+
+> TODO: graphic
+
+### When opening a buffer, restore its per-file state
+
+## Drawbacks
+
+<!--
+Why should we *not* do this?
+-->
+
+## Rationale and alternatives
+
+<!--
+- Why is this approach the best in the space of possible approaches?
+- What other approaches have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+-->
+
+## Unresolved questions
+
+<!--
+- What unresolved questions do you expect to resolve through the RFC process before this gets merged?
+- What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of Atom?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+-->
--- a/docs/rfcs/004-session-management.md
+++ b/docs/rfcs/004-session-management.md
@@ -28,12 +28,13 @@ Problems to address:
 
 A _Project_ is the unit of persistent window state in Atom related to a coherent development effort. It includes:
 
-* The number of windows open;
 * The set of root folders;
 * Pane and dock configuration;
 * Open pane and dock items;
 * Cursor position, fold state, and other TextEditor characteristics;
 * Per-package [serialized state](https://flight-manual.atom.io/behind-atom/sections/serialization-in-atom/).
+
+Each Atom window is associated with one Project, and each Project is associated with at most one Atom window.
 
 Each Project has a unique, user-customizable name.
 
--- a/docs/rfcs/004-session-management.md
+++ b/docs/rfcs/004-session-management.md
@@ -26,7 +26,7 @@ Problems to address:
 
 ## Explanation
 
-A _Project_ is the unit of persistent window state in Atom related to a coherent development effort. It includes:
+A _Session_ is the unit of persistent window state in Atom related to a coherent development effort. It includes:
 
 * The set of root folders;
 * Pane and dock configuration;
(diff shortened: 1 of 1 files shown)


smashwilson:
We had an RFC for project files floating around too, didn't we? I can't seem to find it.

docs/rfcs/004-session-management.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Add a command-line argument to explicitly choose the Session
+
+When opening Atom from the command line with one or more locations specified:
+
+1. If exactly one non-open Session matches those paths, that Session is loaded into the newly opened Atom window.
+2. If no non-open Sessions match, a new Session is created for that Atom window.
+3. If multiple Sessions match, a new Session is created for that Atom window, and a notification is displayed allowing the user to switch to any of the matching Sessions and destroy the just-created one.
+
+In this context, "matches" means _every path in the set, normalized through realpath and case sensitivity, begins with the path of at least one project folder in the Session under consideration_, and, as a special case, _a Session containing zero project folders matches everything_.
```
smashwilson: I'm curious if this is still too much magic. I'm trying to thread the needle between preventing the "oh shit, where did my work go?" reaction on every Atom load and not having too much implicit, poorly understood behavior.

docs/rfcs/004-session-management.md
```diff
@@ -0,0 +1,125 @@
...
+
+The `--session` or `-s` flags to the `atom` command-line script explicitly designate the name of a Session which the newly opened window should open. If no Session with that name exists, a new one is created. If a window is already open on the named Session, the provided paths are added to the existing window and it is brought to the foreground.
```
smashwilson: `-s` collides with `--safe` but we don't currently allow `--safe` to be shortened. I think `--session` would be the more common expansion anyway.

docs/rfcs/004-session-management.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Create a new bundled package to list and manage serialized state
+
+Clicking the "Session" tile in the tree-view opens a workspace item for listing, creating, loading, modifying, and deleting known Sessions. This is where Sessions can be renamed and managed.
+
+> TODO: graphic
+
+Loading a different Session immediately serializes the window's state into the previously active project and deserializes the window state from the chosen one into this window.
+
+A separate tab within the Session management shows known unsaved buffers. This acts as a kind of "dead letter office" to track down and recover lost changes.
```
smashwilson: ☝️ Unsure about this being a "tab". Maybe just a button at the top that opens a different "lost and found" item? I'm kind of waiting to make some UI mocks to figure out the best way to do this.

docs/rfcs/004-session-management.md
```diff
@@ -0,0 +1,125 @@
...
+
+> TODO: graphic
+
+### When opening a buffer, restore its per-file state
+
+Each time a new buffer is opened, restore its per-file state if any has been persisted, regardless of the window's active session. If the file on disk has been modified externally since the unsaved changes were created, display a banner in the TextEditor to give the user a notice.
```
smashwilson: Unfortunately, due to the way we serialize unsaved buffers, I don't think we have a good way to every recover these 😢 . I'd love to offer some kind of git-powered merge for the unsaved state.

Oh, unless there's something we can do with the files in `~/.atom/recovery/`?

smashwilson (commented):
Some ongoing thoughts for discussion.

smashwilson:
@daviwil @lee-dohm @rsese: I'm curious what your thoughts are about the direction I'm going with this. What edge cases am I not seeing? Will this be too intrusive? Will new users be less surprised by the way project state is handled, and will existing users by able to understand it well?

docs/rfcs/004-session-management.md
TriMoon: No idea how atom does it's serialization you talk about in respect to unsaved buffers, but why not use an actual git repo for that?
Then it would be easier to spot differences?

TriMoon:
> We had an RFC for project files floating around too, didn't we? I can't seem to find it.

Do you mean #5168 and https://github.com/harmsk/atomic-management/issues/40 ?

rsese:
> My goals are to make it more explicit and intentional and to make it more difficult for users to "lose" the setup that they like or their unsaved buffer changes.

:+1: :+1:

Some early thoughts, this particular question was in my head:

> Will new users be less surprised by the way project state is handled, and will existing users by able to understand it well?

I've been trying to think about the issues I've seen with project state and how that might relate to the proposal so far.    It feels like new users might not dig into something like this right away but I can definitely see existing users benefiting from this since they've likely run into issues with saved state somewhere along the line.  Giving existing users the ability to look under the hood and recover unsaved work seems like a big plus (or at least that's my gut reaction after reading through what you have).  

> What edge cases am I not seeing?

Not sure how pertinent it is here,  but one state related thing that comes up relatively often is clearing saved state.  That's a common debugging suggestion we mention and document and when it works it's fine but it also blows away all saved state even if it's just a single project causing the problem.  Again though, not sure if that's relevant here and if/how that could tie into this RFC.

> Will this be too intrusive?

I feel like since it's behind a click on the project name in the tree-view, it doesn't seem super intrusive in that sense (not sure if you meant something else though about this feature being intrusive)?

One general question that I wasn't 100% clear on - does the RFC currently define a change to how state is saved or is it more defining a way to interact with saved state through a UI and the command line?

smashwilson:
> Not sure how pertinent it is here, but one state related thing that comes up relatively often is clearing saved state. That's a common debugging suggestion we mention and document and when it works it's fine but it also blows away all saved state even if it's just a single project causing the problem. Again though, not sure if that's relevant here and if/how that could tie into this RFC.

Ooh yeah. Maybe give `--clear-window-state` an optional argument to name the session to clear?

I'd also hope that we could add a control to delete sessions from the session UI, too.

> (not sure if you meant something else though about this feature being intrusive)

I think I'm worried about turning session management into something you _have_ to understand to effectively use Atom, rather than something you can find and learn about when you need it?

docs/rfcs/004-session-management.md
```diff
@@ -0,0 +1,125 @@
...
+
+> TODO: graphic
+
+### Add a command-line argument to explicitly choose the Session
+
+When opening Atom from the command line with one or more locations specified:
+
+1. If exactly one non-open Session matches those paths, that Session is loaded into the newly opened Atom window.
+2. If no non-open Sessions match, a new Session is created for that Atom window.
+3. If multiple Sessions match, a new Session is created for that Atom window, and a notification is displayed allowing the user to switch to any of the matching Sessions and destroy the just-created one.
```
akonwi: Why allow multiple sessions of the same name? I'd expect the names to be unique.

docs/rfcs/004-session-management.md
akonwi: This makes sense to me

docs/rfcs/004-session-management.md
akonwi: I'm a big fan of having a separate tree-view for sessions and the ability to manage them.

akonwi:
> I'd also hope that we could add a control to delete sessions from the session UI, too.

There should definitely be a UI control for deleting sessions. Ideally all session management can happen from the UI.

> I think I'm worried about turning session management into something you _have_ to understand to effectively use Atom, rather than something you can find and learn about when you need it?

One way to make the sessions more clear is to have opening atom, either default to opening the last used session or if the "Restore Previous Window on Start" option is disabled, open a blank window and replace the "Getting Started" tab from the initial install with the session controls. Sort of like how Intellijj gives the user the chance to pick the project they want.

docs/rfcs/004-session-management.md
smashwilson: Session names are unique, but multiple sessions may contain the same project folders. (For an example of why this may be useful, I often have two Atom windows open on the project I'm working, one for "reference" and one for "input".)

For example, given:

```sh
atom --session one project-a/ project-b/
atom --session two project-b/
atom --session three project-b/ project-c/
```

If you run:

```sh
atom project-b/
```

Atom would:

1. Open a new window containing `project-b/`
2. Create a new session (say "session-123")
3. Display a notification offering to switch to sessions "one", "two", or "three".
   * If the user chooses an existing session -- say "three" -- the current window loads the chosen session (so project folders `project-b/` and `project-c/`, any open editors, pane arrangement and so on) and "session-123" is destroyed
  * If the user dismisses the notification, "session-123" remains in use.

Does that make sense?

smashwilson:
> There should definitely be a UI control for deleting sessions. Ideally all session management can happen from the UI.

👍 

> One way to make the sessions more clear is to have opening atom [..]

Just to be sure: do you mean opening Atom with no paths (or from the Dock/Start Menu)?

> [..] either default to opening the last used session or if the "Restore Previous Window on Start" option is disabled, open a blank window and replace the "Getting Started" tab from the initial install with the session controls.

Small change: for consistency with what we do now, I'd like to open _all_ last open sessions. If you quit Atom with multiple windows open and then re-open it, the same windows (containing the same sessions) all re-open.

Other than that, yeah, I like the idea of doing something with the welcome package to expose this 🤔

smashwilson:
@rsese I just realized that I hadn't actually answered this:

> One general question that I wasn't 100% clear on - does the RFC currently define a change to how state is saved or is it more defining a way to interact with saved state through a UI and the command line?

It does change how we manage state, yes. Currently, saved sessions are selected implicitly by the set of root project folders.

I'm proposing that we decouple this and make the active session a first-class, named thing that a user chooses deliberately. My hope is that this will (a) make saved state more predictable to users and (b) this will sidestep long-running issues that we have because the root project folder set is _mutable_ -- so you can add or remove project folders and trample different state, for example.

I'm also proposing that we separate file-specific state from session state. This bit is to avoid problems with unsaved buffers getting "lost" if you don't know what project folders you had open when you were working on it.

Aerijo:
Would this also work across Atom versions? I always find the build from source Atom picks up on some parts of the session, but gets sort of corrupted (no idea why this is; there also tends to be a failed GET error in the console). There has also been an update that broke backwards compatibility entirely (that was not a fun week).

I would really like a robust solution that works seamlessly across all Atoms.

smashwilson:
We'd need a way to upgrade existing session data (which I should add to the RFC, yeah). Including a version number in the state data might help us migrate in the future as well. I'm not sure what else we can do to ensure this, though; compatibility is tricky to guarantee given that so much of it is arbitrary JSON from third-party packages. Do you have any suggestions?

Aerijo:
+1 for version number. It should help identify how to deserialize the payload, at least for core stuff like the session. I think this should be passed to third party packages, but expect them to handle the rest.

But no suggestions on how to implement it. I don’t even know how it currently works (some sort of database?)

docs/rfcs/004-session-management.md
akonwi: Yea that's more clear to me. 👍

akonwi:
> > One way to make the sessions more clear is to have opening atom [..]
> 
> Just to be sure: do you mean opening Atom with no paths (or from the Dock/Start Menu)?

Correct to both. If no path/session is specified, prompt the user to choose one.

ulidtko:
Please, *please* guys, while you're at it, perhaps take some care of **bookmarks** as well?

Bookmarks look nice, but are utterly dysfunctional (as of 1.34, and long before): merely closing and reopening a tab loses all bookmarks in that file. That's useless! Like, literally kills any utility of bookmarks.

I know there've been reports on this, and I'm not asking for immediate fixes -- just take into consideration that Atom bookmarks look like joke (of the non-funny kind) currently. It'd be awesome if you make it easy enough for someone else to fix in an evening.

matt9698:
This is just an idea I had while reading through this thread today. To be honest, I'm not sure if this fits neatly within the context of this idea - but is rather an auxiliary feature/package that could be added later.

The ability to save editor window layouts (separate from individual sessions/projects) would be quite nice e.g. some sort of menu View > Layout

Users could choose to create named layouts e.g. "Default",  "Zen" (Remove all panes except the active editor window) and "Split" (of course its rather easy to split panes already, but for the sake of completeness).

Maybe even "Clear" (close all panes, to quickly get a blank window to build a new layout).

If you don't quite understand what I am trying to describe see [this document](https://docs.blender.org/manual/en/latest/interface/window_system/screens.html) from the Blender manual (which is basically the exact thing I'm describing here).

Please let me know if this idea is too off topic for this RFC and the Atom core repo in general, I'd be happy to suggest the idea somewhere more appropriate instead.

agent255:
@matt9698 the link you have given leads to a 404

systemespectr (approved):

vishnuaq (approved):

vishnuaq:
- **f[@]()**

sadick254:
@smashwilson Thank your efforts on the RFC on session management. I am sorry to say we don't have the capacity to work on or maintain such a feature should it be present on Atom. I am closing this for now but will be happy to reopen when we can actually provide meaningful input on the RFC.

Issue #18702, opened 2019-01-15
Running "atom ." in bash on Ubuntu WSL fails on Windows 10 1809


Jinivus:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Something changed between windows builds 1803 and 1809 that caused the atom.sh script to fail to launch atom.exe on the windows side FROM Ubuntu bash correctly.

### Steps to Reproduce

1. Open WSL Bash
2. Move to windows accessible directory ie `cd /c/dev/`
3. Run `atom .`

**Expected behavior:** Atom should launch opening the current directory (/c/dev/)

**Actual behavior:** Prints "The system cannot find the path specified." and does not launch

**Reproduces how often:** 100%

### Versions

Windows 1809 + Ubuntu 18.04 + Atom 1.35.0-beta0 x64

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

rsese:
Thanks for the report - I'm unable to reproduce the error that you see with 1.34.0 on Windows 10.  If I try to run `atom` in WSL Ubuntu 18.04 I get the error reported in https://github.com/atom/atom/issues/18126.  

Do you think your PR will fix https://github.com/atom/atom/issues/18126?

mslinn:
When I type `atom .` on WSL I get this error message:

        shared memfd open() failed: Function not implemented

PR #18703, opened 2019-01-15
Update windows atom.sh launch script to support WSL changes in 1809

Jinivus:
### Identify the Bug

#18702

### Description of the Change

Change windows atom.sh launcher to:

- Not replace "//" with "\" when building the ATOMCMD when detected in WSL
- Change to launch ATOMCMD in a similar fashion to Msys

### Alternate Designs

Could potentially format the path in a way that was friendly for the original Cygwin launch method

### Possible Drawbacks

Potentially breaks on previous builds of Windows, although I don't have the facilities to test

### Verification Process

Repeated the steps outlined in the linked issue and atom launched as expected.

### Release Notes

- Fixed an issue where the windows version of atom would not launch from inside WSL


resources/win/atom.sh
```diff
@@ -12,15 +12,14 @@ else
     eval $(grep "^root" /etc/wsl.conf | sed -e "s/ //g")
     root="$(echo $root | sed 's|/|\\/|g')"
     ATOMCMD="$(echo $PWD | sed 's/\/mnt\/\([a-z]*\)\(.*\)/\1:\2/')/atom.cmd"
-    ATOMCMD="${ATOMCMD////\\}"
```
solvaholic: > Not replace "//" with "" when building the ATOMCMD when detected in WSL

In [bash](https://linux.die.net/man/1/bash), this 👇 changes all the `/` in ATOMCMD into `\`:

https://github.com/atom/atom/blob/c5ebfbddd096a24bb99b832432a5dfcd67da4714/resources/win/atom.sh#L15

PR #18705, opened 2019-01-16
Enable autoFocus option to accept an element to focus on pane creation

### LINKED ISSUE #18511: allow panel option autoFocus to be used as initialFocus
## Summary

The [autoFocus](https://atom.io/docs/api/v1.32.2/Workspace#instance-addModalPanel) option for atom.workspace.addModalPanel should be able to be the object that you want focused initially instead of only focusing on the first tabbable element.

## Motivation

```js
atom.workspace.addModalPanel({item: panelElement, autoFocus: inputElement});
```

or

```js
atom.workspace.addModalPanel({item: panelElement, autoFocus: '.input-element'});
```

The current way to get this to work is to focus the element after adding the panel

```js
atom.workspace.addModalPanel({item: panelElement, autoFocus: true});
inputElement.focus();
```

I think it makes it a little bit easier and a lot more intuitive to allow the autoFocus options to be the element that you want automatically focused.

## Additional context

[focus-trap](https://github.com/davidtheclark/focus-trap#usage) (which is used when autoFocus is set) can be passed an option for initialFocus that can be an element, selector string, or function that returns an element.

The change would be a few lines of code to check if the autoFocus option is one of those and set the initalFocus option.

https://github.com/atom/atom/compare/master...UziTech:autoFocus-element?expand=1#diff-287ffd9dadb8379157bf7474679cbbd0
```diff
-      if (panel.autoFocus) {
-        const modalFocusTrap = focusTrap(panelElement, {
+      if (panel.autoFocus !== false) {
+        const focusOptions = {
           // focus-trap will attempt to give focus to the first tabbable element
           // on activation. If there aren't any tabbable elements,
           // give focus to the panel element itself
           fallbackFocus: panelElement,
           // closing is handled by core Atom commands and this already deactivates
           // on visibility changes
           escapeDeactivates: false
-        })
+        }
+         if (panel.autoFocus !== true) {
+          focusOptions.initialFocus = panel.autoFocus
+        }
+        const modalFocusTrap = focusTrap(panelElement, focusOptions)
```

UziTech:
### Identify the Bug

fixes #18511 

### Description of the Change

Use `panel.autoFocus` as the element for `focusTrap.initialFocus` if it is not a boolean value. 

### Possible Drawbacks

None. It is backwards compaitible.

### Verification Process

added a test

### Release Notes

The autoFocus option for panels can now be an element that will receive initial focus.


src/panel-container-element.js
```diff
@@ -54,16 +54,21 @@ class PanelContainerElement extends HTMLElement {
...
           // on activation. If there aren't any tabbable elements,
           // give focus to the panel element itself
           fallbackFocus: panelElement,
           // closing is handled by core Atom commands and this already deactivates
           // on visibility changes
           escapeDeactivates: false
-        })
+        }
+
+        if (panel.autoFocus !== true) {
```
daviwil: What if someone deliberately sends `undefined` or `null`?  It seems like those values will get through these two `panel.autoFocus` guards and possibly cause an error in `focusTrap`.

daviwil (commented):
Thanks @UziTech, this looks good!  One possible change needed, but should be easy to merge this afterward.

src/panel-container-element.js
UziTech: Nice catch! I updated it so any falsy value will be treated the same as false

rafeca:
@UziTech: Thanks for contributing!

Since @daviwil already gave the 👍 , I've just resolved the conflicts and I'll merge this once the CI tests pass 🎉

PR #18738, opened 2019-01-24
Show Tree-sitter grammars in grammar selector

### LINKED ISSUE #226: Different Parsing between "HTML" and "Auto Detect"
### Prerequisites

### Description

I get different parsing behavior between when I open a file and select the "Auto Detect" language option and when I specifically select "HTML".

### Steps to Reproduce

1. Open an HTML file
2. open the language selector and select "Auto Detect"
3. View the parsed classes in the Inspector
4. open the language selector and select "HTML"
5. see that the classes for tokens has changed

The parsing is the same, whether auto-detected or manually selected.

The parsing is different.

Always

### Versions

OS: Ubuntu 18.04.02

I suspect I am having the same issue on Mac OS X on my work computer. I can check tomorrow if it is helpful.

```
$ atom --version
Atom    : 1.36.1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
```
```
$ apm --version
apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.36.1
python 2.7.15rc1
git 2.17.1
```

### Additional Information

When I disable Tree Sitter parsers, this problem goes away.  It seems like when I pick "HTML" it is using Tree-Sitter parsing  but when I auto-detect I get the old parsing.

### LINKED ISSUE #17029: getGrammars() ignores tree-sitter grammars
### Description

Tree-sitter grammars will not be returned by the `getGrammars()` function of a `GrammarRegistry`. Instead, only TextMate grammars are returned.

```js
  // Extended: Get all the grammars in this registry.
  //
  // Returns a non-empty {Array} of {Grammar} instances.
  getGrammars () {
    return this.textmateRegistry.getGrammars()
  }
```

This causes issues with the `grammar-selector` package, which uses that function to build the list of available grammars. All tree-sitter grammars will be hidden from it.

Interestingly, if a regex grammar has the same `name` value as a tree-sitter one, then the regex grammar will appear in the list, but selecting it will apply the tree-sitter grammar

This can be seen by selecting `Python`, and noticing the green tick mark is not applied to it when you reopen the menu (in contrast to a solely regex grammar, with no tree-sitter alternative).

### Steps to Reproduce

1. Open the grammar menu and select `Python`
2. Open the grammar menu and look at the `Python` entry; there is no tick mark
3. Select `Python Console` (or any other regex-only grammar)
4. Open the grammar menu and see the tick mark next to `Python Console`

Alternatively, a grammar that is solely tree-sitter based will not appear at all in the selection menu, but will be automatically applied when the right file type is opened.

Tick next to `Python` because it was selected

No tick, because we were actually selecting the regex gramar and the tree-sitter one "intercepted" it as it was being applied. So the result is that the regex grammar was _not_ applied, and so no tick mark.

**Reproduces how often:**
100%

### Versions

```
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.26.0-beta0
python 2.7.10
git 2.14.3

Atom    : 1.26.0-beta0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0

macOS 10.13.3
```

Aerijo:

Fixes #17029 (I endorse it...)
Closes https://github.com/atom/language-html/issues/226

### Description of the Change

Adds an optional argument to `atom.grammars.getGrammars` to allow Tree-sitter grammars as well. It defaults to false to avoid any backwards compatibility issues, though I'd be fine with making it true or removing it entirely, given that Tree-sitter is now on by default for widely used languages like JS, C, etc.

Tests have not been added yet. Still need to style the grammar selector and fix my Dev mode ideally.

### Alternate Designs

Aside from the above, the grammar selector element constructor also needs additional styling to mark the type of grammar. Leaving TextMate grammars as they are is fine and least confusing, but Tree-sitter grammars often share a name with these, so they look like duplicate entries. I was thinking of adding `Ts` before the scope, but I would ask @simurai or someone to show me how to do it (my attempt  failed miserably...).

### Possible Drawbacks

Current design - making the argument a direct parameter could be awkward if we decide to add more options in future. Making it an object would future proof it.

If we enable it by default, it may break some community packages that ... handle grammars? That aren't already somewhat broken if they are ignoring Tree-sitter anyway? I'm not convinced this is an issue, so would like some thoughts on it.

### Verification Process

Manually overrided it in my init script. Dev mode currently does not open for me, and I have not worked out why.

### Release Notes

Show Tree-sitter grammars in grammar selector


simurai:
> Leaving TextMate grammars as they are is fine and least confusing, but Tree-sitter grammars often share a name with these, so they look like duplicate entries. I was thinking of adding `Ts` before the scope

Or write `Ts` out to `Tree-sitter`?

screen shot 2019-02-12 at 4 36 53 pm

Although I wonder... For users that don't follow Atom development or read the Atom blog, do they know what "Tree-sitter" means? We could also be more descriptive and say that there is a normal and an "advanced" version:

screen shot 2019-02-12 at 4 37 23 pm

Aerijo:
@simurai Writing our the full Tree sitter is fine, but I don’t like calling it advanced; that would also be just as confusing IMO. They are already referred to as Tree-sitter in the core setting, and “advanced” is a generic word that could refer to anything, including potential future supported grammar engines.

E.g., if the Sublime grammar engine gets ported successfully (even a community package), there could end up be three types of grammar. 

It would be best if there was a reliable way to tell users about Tree-sitter vs TextMate grammars, and why they’d pick one over the other. Unfortunately, they’re unlikely to go to the grammar selector package README for this. Maybe a small description somewhere on the prompt could link to an explanation?

Aerijo:
Did not mean to close

50Wliu:
Is it possible for the Grammar Selector to hide grammars that also have a tree-sitter variant?

Aerijo:
@50Wliu Yes, there should be a setting for that (normal users don’t need TextMate, but there are still cases where I would want to specify the TextMate one without disabling things)

simurai:
Ohh.. It just occurred to me that `JavaScript (Tree-siter)` might also be confusing because it looks like it's some sort of "flavor" of JavaScript and not the parsing system used. How about having green badges that just say `Tree-sitter` in front of the scopes. I'll push that.

screen shot 2019-02-13 at 2 05 13 pm

> Maybe a small description somewhere on the prompt could link to an explanation?

That would be helpful. Not sure how to describe it in the very small space left. Without, users that are curious would have to google it and hopefully find a blog post like https://github.blog/2018-10-31-atoms-new-parsing-system/.

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -19,9 +19,19 @@ class GrammarListView {
 
         const div = document.createElement('div')
         div.classList.add('pull-right')
+
+        if (grammar.constructor.name === "TreeSitterGrammar") {
+          console.log("TS", grammar)
```
simurai: TODO: Remove before merging.

Arcanemagus:
> That would be helpful. Not sure how to describe it in the very small space left.

Can those `Tree-sitter` badges have alt-text added on hover?

simurai:
> Can those `Tree-sitter` badges have alt-text added on hover?

Yes, here with a `title` attribute:

title

Does the following description make sense?

```
A fast parsing system with improved syntax highlighting, code folding, syntax-aware selection.
```

Aerijo:
My suggestion:
```
(Recommended) A faster parser with improved syntax highlighting & code navigation support.
```
Shorter and, in case the user can't tell from the listed improvements, explicitly says it's recommended.

Aerijo:
OK, couple of outstanding issues I think can be fixed in this PR:
1. Order of TS and TM grammars is random when both have same name. E.g., C has TS first, but C++ has TM first.
2. Choosing TM grammar does not actually choose the TM grammar.

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -20,7 +20,7 @@ class GrammarListView {
         const div = document.createElement('div')
         div.classList.add('pull-right')
 
-        if (grammar.constructor.name === "TreeSitterGrammar") {
+        if (isTreeSitter(grammar)) {
```
Aerijo: This is used a few times, and as a function adds better separation of the idea from the implementation

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -82,32 +82,64 @@ class GrammarListView {
   async toggle () {
     if (this.panel != null) {
       this.cancel()
-    } else if (atom.workspace.getActiveTextEditor()) {
-      this.editor = atom.workspace.getActiveTextEditor()
+      return
+    }
+
+    const editor = atom.workspace.getActiveTextEditor()
```
Aerijo: It's trivial, but the original felt wasteful compared to just keeping the editor the first time. It's also technically more correct perhaps? Can the editor change between two consecutive steps?

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -82,32 +82,64 @@ class GrammarListView {
...
         this.currentGrammar = this.autoDetect
       }
 
-      const grammars = atom.grammars.getGrammars(true).filter((grammar) => {
+      let grammars = atom.grammars.getGrammars({includeTreeSitter: true}).filter(grammar => {
         return grammar !== atom.grammars.nullGrammar && grammar.name
       })
+
+      if (atom.config.get("grammar-selector.hideDuplicateTextMateGrammars")) {
+        const oldGrammars = grammars
```
Aerijo: This is a mess, but seems to work. Open for improvements.

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -82,32 +82,64 @@ class GrammarListView {
...
           return -1
         } else if (b.scopeName === 'text.plain') {
           return 1
-        } else if (a.name) {
-          return a.name.localeCompare(b.name)
-        } else if (a.scopeName) {
-          return a.scopeName.localeCompare(b.scopeName)
-        } else {
-          return 1
+        } else if (a.name === b.name) {
```
Aerijo: Equivalent logic (besides TS check), as we filtered out grammars with no name already.

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -82,32 +82,64 @@ class GrammarListView {
...
       this.attach()
     }
   }
 }
+
+function isTreeSitter (grammar) {
+  return grammar.constructor.name === "TreeSitterGrammar"
+}
+
+function compareGrammarType (a, b) {
```
Aerijo: Separate function to make extending to multiple grammar types easier and not bloat the main sort function

src/grammar-registry.js
```diff
@@ -549,10 +549,12 @@ class GrammarRegistry {
   // Extended: Get all the grammars in this registry.
   //
   // Returns a non-empty {Array} of {Grammar} instances.
-  getGrammars (includeTreesitter = false) {
-    let grammars = this.textmateRegistry.getGrammars()
-    if (includeTreesitter) grammars = grammars.concat(Object.values(this.treeSitterGrammarsById))
-    return grammars
+  getGrammars (params = {includeTreeSitter: false}) {
```
Aerijo: Turned into options object so we can add / change settings later without breaking changes

Maybe add TM-TS duplicate removal here? It would be slightly easier, and allow it to be used by all packages instead of making them do it themselves

Aerijo:
Added specs for most behaviour (aren't checking the order of Tree-sitter / TextMate when both show).

The specs are quite rigid, expecting specific ordering and nesting of the DOM nodes, but they should not be flakey.

I also refactored getting the list node into it's own function for some quality-of-life improvement.

Aerijo:
@simurai The button text is smaller than the label text
image

I'm using a different monitor now, and the high pixel density means app components look really tiny, especially text. In the image, the list items are 15px, but the badge text is 12px. Would it be possible to make the badge text inherit from the list by default?

rafeca:
Hey! 👋

This is good news: now the Atom code is more consistent and it's much easier to re-format the code to follow the codestyle (now you only need to run `script/lint --fix`).

With ❤️, the Atom team.

Aerijo:
I've been considering returning Tree-sitter grammars by default, or explicit opt in. I've gone with by default, because they are supposed to be the preferred grammar now. There were a couple of hitches with other parts of GrammarRegistry that expose grammars, but I think everything works (the specs pass).

The failing specs seem completely unrelated; the mac build failed to load, throwing
```
[command]/bin/bash --noprofile --norc /Users/vsts/agent/2.153.1/work/_temp/d77d45f0-4dd1-4d72-a48e-8ec869406f78.sh
internal/modules/cjs/loader.js:596
    throw err;
    ^

Error: Cannot find module 'colors'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/vsts/agent/2.153.1/work/1/s/script/test:5:1)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
```

I think this is ready to be reviewed and tried in beta.

packages/grammar-selector/lib/grammar-list-view.js
```diff
@@ -72,32 +88,68 @@ module.exports = class GrammarListView {
...
       if (this.currentGrammar === atom.grammars.nullGrammar) {
         this.currentGrammar = this.autoDetect;
       }
 
-      const grammars = atom.grammars.getGrammars().filter(grammar => {
+      let grammars = atom.grammars.getGrammars().filter(grammar => {
         return grammar !== atom.grammars.nullGrammar && grammar.name;
       });
+
+      if (atom.config.get('grammar-selector.hideDuplicateTextMateGrammars')) {
```
nathansobo: In this case, do you really need the second call to `atom.grammars.getGrammars({ textMateOnly: true }`? It seems like you could simply iterate through the grammars array twice, once to populate the blacklist of Tree Sitter grammars and then a second time to filter. It's not a huge deal though.

src/grammar-registry.js
```diff
@@ -601,9 +617,19 @@ module.exports = class GrammarRegistry {
...
   // Extended: Get all the grammars in this registry.
   //
+  // * `options` (optional) {Object}
+  //   * `textMateOnly` (optional) {Boolean} Set to ignore
+  //     [Tree-sitter](https://github.blog/2018-10-31-atoms-new-parsing-system/) grammars
+  //
   // Returns a non-empty {Array} of {Grammar} instances.
-  getGrammars() {
-    return this.textmateRegistry.getGrammars();
+  getGrammars(params) {
```
nathansobo: I think that if we want to return Tree Sitter grammars by default, which I agree would be the cleaner choice long term, then we need `TreeSitterGrammar` to conform to [the documented public API](https://atom.io/docs/api/v1.38.2/Grammar) of the `Grammar` objects we claim we are returning from this method. I think the `onDidUpdate` method could be a no-op, but we'd want a compatible implementation for `tokenizeLine` and `tokenizeLines`. If we return objects from this method that don't conform to that interface, we'll basically be invalidating our documented API. So I see two choices, both acceptable:

* Change this method to *not* return Tree Sitter grammars by default
* Backfill support for the public `Grammar` API on `TreeSitterGrammar` (this would actually be my preference)

nathansobo (changes requested):
I really love the UX you came up with that explains what a Tree Sitter grammar is to the user. Nice work. From a code perspective, this all seems legit. I'm concerned about the API implications however and have a comment with more details about it inline.

src/grammar-registry.js
Aerijo: I've backfilled the public methods. `onDidUpdate` does nothing, and the `tokenizeLine`(s) functions return a generic result, applying only the root scope to each line.

packages/grammar-selector/lib/grammar-list-view.js
Aerijo: Yeah, cleaned it up

nathansobo:
The CI failure is really odd.

nathansobo:
Okay, I merged master into this branch after merging #19580, so hopefully we get better behavior out of the tests.

nathansobo:
Before we merge, I'm going to spend some time seeing if we can actually return tokens from the methods on grammar. The current shims won't *break* the current API, but I also think they don't really perform as we advertise in the documentation, which could be surprising. Let's see what I can come up with.

nathansobo:
Sorry. After digging into this I really don't think we can sanely shim `tokenizeLine`. It's intended to carry the state of the first-mate parser across calls, and my scan of all the package code out there shows that people are indeed using it. I think we should avoid returning objects that violate the documented interface of `Grammar` without an explicit option. And maybe we leave that option undocumented for now. I can take care of making the relevant changes.

Aerijo:
OK, it's now set back up to require opt in for Tree-sitter. The change took a few tries, because there are a couple of places grammars are exposed, and it wasn't consistent to begin with (see `forEachGrammar`, which exposed TS grammars for the auto assignment methods).

nathansobo:
Thanks so much! It will be really nice to be able to select these grammars via the UI. Really, the functionality was incomplete without this.

nathansobo:
A nice next step would be to start thinking about how we can expose the unique capabilities of TreeSitter grammars and language modes in our official API.

maxbrunsfeld:
Nice work @Aerijo and @simurai. Looks great.

PR #18759, opened 2019-01-28
RFC: Block decoration ordering

smashwilson:
An idea for an API to explicitly order block decorations that occur at the same screen row in a TextEditor. This came out of our work in https://github.com/atom/github/pull/1913.

-----
[View rendered docs/rfcs/004-decoration-ordering.md](https://github.com/atom/atom/blob/aw/rfc/marker-ordering/docs/rfcs/004-decoration-ordering.md)

### DIFF
--- a/docs/rfcs/004-decoration-ordering.md
+++ b/docs/rfcs/004-decoration-ordering.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-Proposed
+Accepted
 
 ## Summary
 
--- a/docs/rfcs/004-decoration-ordering.md
+++ b/docs/rfcs/004-decoration-ordering.md
@@ -1,0 +1,41 @@
+# Decoration ordering
+
+## Status
+
+Proposed
+
+## Summary
+
+Order block decoration items in the DOM in a deterministic and controllable way.
+
+## Motivation
+
+When multiple block decorations are created at the same screen line, they are inserted into the DOM in an order determined by the sequence of their creation; from oldest to newest when `position` is set to `"before"`, from newest to oldest when `position` is set to `"after"`. While this is deterministic, it is limited: it isn't possible to insert decorations within a sequence of existing ones, and it's difficult to control the order of decorations when creating and destroying and moving markers around an editor.
+
+We hit the need for this in [atom/github#1913](https://github.com/atom/github/pull/1913) when we have a block decoration for multiple consecutive collapsed file patches.
+
+## Explanation
+
+[TextEditor::decorateMarker()](https://atom.io/docs/api/v1.34.0/TextEditor#instance-decorateMarker) accepts an additional `order` parameter in its `decorationParams` argument when `type` is "block". When multiple block or overlay decorations occur at the same screen line, they are ordered within the DOM in increasing "order" value. Block decorations with no `order` property are rendered after those with one, in the order in which they were created, oldest to newest.
+
+## Drawbacks
+
+This is a breaking change for co-located block decorations created with an "after" position - they'll now appear in the reverse order.
+
+When multiple packages create block decorations at the same screen line, they'll need to coordinate their `order` values to have expected behavior. There may not even be a clear, universal answer about how block decorations from distinct packages _should_ be ordered.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/004-decoration-ordering.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Explanation
+
+[TextEditor::decorateMarker()](https://atom.io/docs/api/v1.34.0/TextEditor#instance-decorateMarker) accepts an additional `order` parameter in its `decorationParams` argument when `type` is "block". When multiple block or overlay decorations occur at the same screen line, they are ordered within the DOM in increasing "order" value. Block decorations with no `order` property are rendered after those with one, in the order in which they were created, oldest to newest.
+
+## Drawbacks
+
+This is a breaking change for co-located block decorations created with an "after" position - they'll now appear in the reverse order.
+
+When multiple packages create block decorations at the same screen line, they'll need to coordinate their `order` values to have expected behavior. There may not even be a clear, universal answer about how block decorations from distinct packages _should_ be ordered.
```
Arcanemagus: What happens if packages _don't_ coordinate and set things to the same `order`? Is it treated like a priority then, and the ordering within the same `order` falls back to the current first/last added?

smashwilson: Oh, yeah, I should clarify that. Yes, I was thinking of falling back to creation order for `order` collisions. (Note that I'm proposing changing this to _always_ be oldest-to-newest as well - I suspect the current behavior is there mostly by accident. If this would break any of your stuff, let me know, that's why this is a request for comments :wink:)

smashwilson:
FYI: My current plan is to merge this at COB today (Eastern time) unless there are any objections.

kuychaco (approved):
Looks good to me! Thanks @smashwilson for the thoughtful RFC and for tackling this ♥️

docs/rfcs/004-decoration-ordering.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Explanation
+
+[TextEditor::decorateMarker()](https://atom.io/docs/api/v1.34.0/TextEditor#instance-decorateMarker) accepts an additional `order` parameter in its `decorationParams` argument when `type` is "block". When multiple block or overlay decorations occur at the same screen line, they are ordered within the DOM in increasing "order" value.
+
+Block decorations with the same `order` property are rendered in the order they were created, oldest to newest. Block decorations with no `order` property are rendered after those with one, in the order in which they were created, oldest to newest.
+
+## Drawbacks
+
+This is a breaking change for co-located block decorations created with an "after" position - they'll now appear in the reverse order.
```
daviwil: Would 'after' still be respected for decorations without an order value?  Basically, once a decoration opts into using an order value, any other decoration that needs to come after should now use an order, otherwise it would still work as expected.

daviwil (approved):
All sounds good to me!  Just one non-blocking question

docs/rfcs/004-decoration-ordering.md
smashwilson: The "after" and "before" position determines where the decoration's item is placed relative to the screen line element - that won't chage; the order value determines the decoration's position relative to the other decorations with the same position at the same screen row. So it'll work something like this:

```
screen-line-0
screen-line-1
{id: 1, order: 1, position: 'before'}
{id: 4, order: 2, position: 'before'}
{id: 5, order: 2, position: 'before'} // <-- fall back to age when order ties
{id: 3, order: 3, position: 'before'}
{id: 0, position: 'before'} . // <-- no order occurs after those with orders; then falls back to age
{id: 2, position: 'before'}
screen-line-2
{id: 7, order: 1, position: 'after'}
{id: 8, order: 2, position: 'after'}
{id: 9, order: 3, position: 'after'}
{id: 6, position: 'after'} // <-- fall back to age again here
{id: 10, position: 'after'}
screen-line-3
screen-line-4
```

The breaking change I pointed out is that, before this, decorations 6 and 10 would render in the opposite order. But I'm pretty sure that's actually just an accident.

PR #18788, opened 2019-02-04
Use a custom Jasmine reporter to actually show spec names

smashwilson:
Our CI builds have been really flaky again recently, and it's difficult to see _why_ because a renderer crash prevents anything but a bunch of dots from showing up in test output. I'm replacing our Jasmine reporter with one that displays the full name of each spec before it begins so that we can diagnose these issues more rapidly.

#### Remaining work

- [x] Do the same for the main process Mocha tests
- [x] Make the new reporters opt-in

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
   include:
     - os: linux
       dist: trusty
-      env: NODE_VERSION=8.9.3 DISPLAY=:99.0 CC=clang CXX=clang++ npm_config_clang=1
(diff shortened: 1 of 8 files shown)


smashwilson:
Oh, right. We already are doing this for the Mocha tests:

```
   . AtomApplication waits until all the windows have saved their state before quitting: 28716ms
```

smashwilson:
Merging despite failures because... the failures are what this is supposed to help us narrow down 🙈

PR #18808, opened 2019-02-06
Build performance improvements

smashwilson:
These are some miscellaneous build performance improvements that I'm tinkering with. May or may not turn out to be valuable.

First: `CI=true` was not actually being passed to the build process, so modules are being installed with `npm install` instead of `npm ci`. This should get us a small performance boost and a more solid guarantee of reproducibility.
 
- [ ] `npm` upgrades aren't "sticking" on Azure DevOps nodes?

Second: today, the build begins by (synchronously) copying source files from the repository root into `out/app/`, including almost all of `src/` and `node_modules/` (pruning things like `.DS_Store` files and specs). Then, we transpile a variety of source files in place within `out/`, and then pass them through electron-link to generate the concatenated, minified script used to generate the v8 snapshot. Ultimately, we include all of these files within the generated `.asar`.

Here are some quick counts from my machine:

| Source | File Count |
| -- | -- |
| Original repository | 36202 |
| Copied to `out/` | 30535 |
| Present in `.asar` | 30534 |

This is a lot of build time disk I/O and `.asar` storage space to shuffle around files that _are never used_. At runtime, the code from the majority of these files is loaded from the snapshot, not from disk. We don't even use it to generate stacktraces.

What I'm trying to do instead is:

1. Modify electron-link to transpile source on demand (see atom/electron-link#18). This will prevent us from needing to stage source to `out/` before transpiling and linking it.
2. Selectively copy only modules that are excluded from the snapshot (and their unique dependencies) into `out/`.

My hope is that this will result in fairly dramatic decreases in build time and distributed app size, especially on Windows where I/O involving many small files is especially costly.

Here are some recent build times from CI, for reference:

| Platform | "Atom Build" duration |
| -- | -- |
| Windows x64 | 29m24s |
| Windows x86 | 27m38s |
| macOS | 14m47s |
| Linux | 18m20s |


script/vsts/platforms/linux.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   steps:
   - script: script/build --create-debian-package --create-rpm-package --compress-artifacts
     env:
+      CI: true
```
daviwil: I could have sworn I read somewhere in Azure Pipelines docs that this is already set, but now I can't find it :/

smashwilson: Haha also adding this broke everything 🙃

smashwilson:
Huh. It looks like the npm upgrades aren't "sticking." No idea why right now.

rafeca:
Hey! 👋

But this change has caused conflicts on your PR that we unfortunately haven't been able to solve due to existing conflicts on this PR.
If you want to solve them, the easiest way to do it is to follow the steps [on this gist](https://gist.github.com/rafeca/eaa21f7ff1cf78374b2006711f3e003d).

Thanks and apologies for the inconvenience! 🤗

With ❤️, the Atom team.

sadick254:
Hey @smashwilson . This branch now has lots of conflicts. Is this something you intend to continue working on?

smashwilson:
Ah, probably not in the foreseeable future, I'm afraid. (And if this is something we did want to ship it would likely be easier at this point to start from scratch anyway.)

aminya:
#21401 addresses the issue in a different way. By running the tasks in parallel, the time is reduced.

PR #18832, opened 2019-02-11
Report macOS renderer crashes in CI

smashwilson:
When we have a renderer process crash in CI, we aren't left with a lot of information to work with. #18788 lets us see at least the name of the crashing spec in the build output, but if the crash location is inconsistent, or if we can't repro locally, we're still out of luck.

If possible, I'd like to upload crash reports from each OS in our Azure DevOps builds so that we can investigate further.

Note that Windows will be the tricky one to handle here. Windows crash dumps include all of the crashed process' environment variables, which can include build secrets in release builds! To avoid leaking those, we'll need to find somewhere private to upload them instead.

##### Remaining work

- [x] macOS
  - [x] Upload crash reports from `~/Library/Logs/DiagnosticReports` as a build artifact

### DIFF
--- a/script/vsts/platforms/macos.yml
+++ b/script/vsts/platforms/macos.yml
@@ -52,6 +52,12 @@ jobs:
       cp $(Build.SourcesDirectory)/out/*.zip $(Build.ArtifactStagingDirectory)
     displayName: Stage Artifacts
 
+  - script: |
+      mkdir -p $(Build.ArtifactStagingDirectory)/crash-reports
+      cp ${HOME}/Library/Logs/DiagnosticReports/*.crash $(Build.ArtifactStagingDirectory)/crash-reports
+    displayName: Stage Crash Reports
+    condition: failed()
+
   - task: PublishBuildArtifacts@1
     inputs:
       PathtoPublish: $(Build.ArtifactStagingDirectory)/atom-mac.zip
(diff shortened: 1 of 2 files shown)


smashwilson:
Works on macOS:

screen shot 2019-02-11 at 11 40 13 am

daviwil:
Fantastic!

smashwilson:
Honestly it might even be worth narrowing this PR to only implement macOS and tackle the other two platforms later. What do you think?

daviwil:
Sounds fine to me!

Side note: one little trick you can use for uploading artifacts without a separate `PublishBuildArtifacts` is to write out a line to stdout [like this](https://github.com/atom/atom/commit/ffd039e172a4e89bd1cf223fbfc83e0364481d26#diff-cfe9645df492474d5071114ebfd5803dR74):

```
console.log(`##vso[artifact.upload containerfolder=OldReleaseNotes;artifactname=OldReleaseNotes;]${oldReleaseNotesPath}`)
```

smashwilson:
> one little trick you can use for uploading artifacts without a separate `PublishBuildArtifacts` is to write out a line to stdout

Whoa magic

smashwilson:
Merged with macOS to track down the crashes we've been seeing recently 👌

Issue #10374, opened 2016-01-11
Double-click drag selections have unusual and bad behavior on non-word characters


russellmcc:
Note: This may not be a bug in the sense that the behavior may be intended, but many text editors (apple text input boxes, emacs) have better behavior.

To reproduce, enter the following text:

```
multiple words;)
```

Double-click and hold on the word `multiple` - this will select the word `multiple` and as expected enter a "per-word" selection drag.  Move the mouse to the right and the word `words` will be correctly selected all at once.  However, it is impossible to add the `;` to the selection without the `)`.  This is different behavior than you would see in some other text editors, including emacs and Mac OS default input fields.

The reason this is bad is it is harder than other editors to select `multiple words;`, without the close parenthesis.  This seems to come up a lot when programming and is my biggest pet peeve with atom.

mherodev:
The bug fix for this, conceptually: 

Atom should exclusively group **alphanumeric** characters as "words" when performing word-selection, considering non-alphanumeric characters as individual words. This is Sublime Text's behavior. (Note: This will have its subtleties, e.g. white space of varying types is another type of word.)

There is another bug related to this that I uncovered, here: https://discuss.atom.io/t/drag-selection-bug/28599/7

When drag-selecting words, Atom groups the remaining newline at the end of every line with other end-of-line non-alphanumeric characters, resulting in weird selection behavior.

To reproduce:

1) Enter the following text into Atom:
```
function select() {
  console.log('Test.');
};
```
2) Double click and drag from **to-the-right-of the last line** up to **to-the-right-of the line with the log statement**.
3) Notice that everything after the word "Test" is highlighted, despite not being hovered.

Behavior in other editors is to only highlight the newline character in this case.

stevenwaller:
I recently switched to Atom and this issue has been bugging me too. I come across it a lot when using jQuery and I'm trying to highlight just the selector. 

For example if I double click on "element" then drag to select "#my-element" it inevitably include "$('" in my selection.

![highlight](https://cloud.githubusercontent.com/assets/1389963/21858646/ca888560-d7de-11e6-96d0-b300770d380b.gif)

mherodev:
Oh hey, I've forgotten about this bug. This shouldn't still be labeled as needs-reproduction, as far as I can tell.

hhking:
in my mac (atom 1.19.0) to reproduce:
1. double click
2. use three finger to drag select
3. delete the selections
4. Notice that the other words are highlighted

[...]

PR #18854, opened 2019-02-14
Fixed issue #10374

v57:
### Identify the Bug
#10374

### Description of the Change
Just removed "+" character from regex so when you double click on things like `)));` it will only select single character

### Alternate Designs
Because this is how word selection works in any OS. Only in atom you can select `)));` by double clicking on it

### Possible Drawbacks
People who leave atom because of this can go back now

### Verification Process
It's just a word selection regex.

### Release Notes
"Not applicable"

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -675,7 +675,7 @@ class Cursor extends Model {
     const nonWordCharacters = _.escapeRegExp(this.getNonWordCharacters())
     let source = `^[\t ]*$|[^\\s${nonWordCharacters}]+`
     if (!options || options.includeNonWordCharacters !== false) {
-      source += `|${`[${nonWordCharacters}]+`}`
+      source += `|[${nonWordCharacters}]`
     }
     return new RegExp(source, 'g')
   }


Arcanemagus:
Thanks for your submission!

We require automated tests for all pull requests that include new or changed code. We do this so that we can ensure that we don't accidentally break your shiny new code the next time we or some other contributor submits a change. If you need help writing automated tests for Atom, check out [Discuss, the official Atom and Electron message board](https://discuss.atom.io) or [join the Atom Slack team](http://atom-slack.herokuapp.com/). There are a bunch of helpful community members that should be willing to point you in the right direction.

Once this includes tests we can re-open it and look at it further.

v57:
I don't have time for this. I just fixed annoying issue and want to give it to others. I'm using is for two weeks and i can't see any problems with this. If you want to make people happy, just approve it please.

50Wliu:
I'm glad that this patch has improved Atom for you. However, we are unable to accept it without tests as 1) we need to ensure that this behavior does not accidentally regress in the future, and 2) we need to ensure that this fix behaves as intended. We will happily take another look at a PR that contains tests.

PR #18885, opened 2019-02-21
Add linter to packages/ folder

rafeca:
### Requirements for Contributing a Bug Fix

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only fix an existing bug. To contribute other changes, you must use a different template. You can see all templates at https://github.com/atom/atom/tree/master/.github/PULL_REQUEST_TEMPLATE.
* The pull request must update the test suite to demonstrate the changed functionality. For guidance, please see https://flight-manual.atom.io/hacking-atom/sections/writing-specs/.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see https://github.com/atom/atom/tree/master/CONTRIBUTING.md#pull-requests.

### Identify the Bug

Linter was not checking the `packages/` folder.

### Description of the Change

The JS code style of the `packages/` folder is not matching the style of the rest of the repository, this is caused by not having the linter running on this folder.

This PR enables the linter and fixes all the linting issues that appeared on that folder. Most of the issues have been fixed automatically by running [`prettier-standard`](https://www.npmjs.com/package/prettier-standard) on that folder, the things that needed to be fixed manually are in separate commits.

### Alternate Designs

N/A

### Possible Drawbacks

This PR changes a bunch of files, and while `prettier` should only do style changes, there's always the possibility of breaking something when making big changes. To mitigate this risk I'm gonna review some of the changes manually and make sure that they're ok.

### Verification Process

I'm going to manually verify the automatic fixes and rely on the existing automated tests.

### Release Notes

N/A (it's only an internal change)

PR #18896, opened 2019-02-22
Add linter to spec/ folder

rafeca:
### Identify the Bug

Linter was not checking the `spec/` folder.

### Description of the Change

**Note**: This PR depends on https://github.com/atom/atom/pull/18885

The JS code style of the `spec/` folder is not matching fully the style of the rest of the repository, this is caused by not having the linter running on this folder.

This PR enables the linter and fixes all the linting issues that appeared on that folder. Most of the issues have been fixed automatically by running [`prettier-standard`](https://www.npmjs.com/package/prettier-standard) on that folder, the things that needed to be fixed manually are in separate commits.

### Alternate Designs

N/A

### Possible Drawbacks

N/A (since this PR only changes test files).

### Verification Process

Check that tests keep passing on CI.

### Release Notes

N/A (it's only an internal change).

### DIFF
--- a/script/lib/lint-java-script-paths.js
+++ b/script/lib/lint-java-script-paths.js
@@ -25,7 +25,7 @@ module.exports = async function () {
   ])
 
   const paths = includePaths.filter(
-    myPath => excludePaths.indexOf(myPath) === -1
+    myPath => !excludePaths.includes(myPath)
   )
 
   return new Promise((resolve, reject) => {
--- a/script/lib/lint-java-script-paths.js
+++ b/script/lib/lint-java-script-paths.js
@@ -6,29 +6,46 @@ const path = require('path')
 
 const CONFIG = require('../config')
 
-module.exports = function () {
+module.exports = async function () {
   const globPathsToLint = [
     path.join(CONFIG.repositoryRootPath, 'exports', '**', '*.js'),
     path.join(CONFIG.repositoryRootPath, 'packages', '**', '*.js'),
     path.join(CONFIG.repositoryRootPath, 'script', '**', '*.js'),
+    path.join(CONFIG.repositoryRootPath, 'spec', '**', '*.js'),
     path.join(CONFIG.repositoryRootPath, 'src', '**', '*.js'),
     path.join(CONFIG.repositoryRootPath, 'static', '*.js')
   ]
-  return expandGlobPaths(globPathsToLint).then((paths) => {
-    return new Promise((resolve, reject) => {
-      standard.lintFiles(paths, (error, lintOutput) => {
-        if (error) {
-          reject(error)
-        } else {
-          const errors = []
-          for (let result of lintOutput.results) {
-            for (let message of result.messages) {
-              errors.push({path: result.filePath, lineNumber: message.line, message: message.message, rule: message.ruleId})
-            }
+  const globPathsToIgnore = [
+    path.join(CONFIG.repositoryRootPath, 'spec', 'fixtures', '**', '*.js')
+  ]
+
+  const [includePaths, excludePaths] = await Promise.all([
+    expandGlobPaths(globPathsToLint),
+    expandGlobPaths(globPathsToIgnore)
+  ])
+
+  const paths = includePaths.filter(
+    myPath => excludePaths.indexOf(myPath) === -1
+  )
+
+  return new Promise((resolve, reject) => {
+    standard.lintFiles(paths, (error, lintOutput) => {
+      if (error) {
+        reject(error)
+      } else {
+        const errors = []
+        for (let result of lintOutput.results) {
+          for (let message of result.messages) {
+            errors.push({
+              path: result.filePath,
+              lineNumber: message.line,
+              message: message.message,
+              rule: message.ruleId
+            })
           }
-          resolve(errors)
         }
-      })
+        resolve(errors)
+      }
     })
   })
 }
--- a/spec/application-delegate-spec.js
+++ b/spec/application-delegate-spec.js
@@ -1,13 +1,6 @@
 /** @babel */
 
-import {
-  it,
-  fit,
-  ffit,
-  fffit,
-  beforeEach,
-  afterEach
-} from './async-spec-helpers'
+import { it } from './async-spec-helpers'
 import ApplicationDelegate from '../src/application-delegate'
 
 describe('ApplicationDelegate', function () {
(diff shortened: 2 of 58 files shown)


rafeca:
Oh, since GitHub does not support stacked PRs, this PR is showing also the commits from #18885 . Please ignore the commit with message `Enable linter on packages/ folder` and all commits older than that one.

jasonrudolph:
> Oh, since GitHub does not support stacked PRs, this PR is showing also the commits from #18885 . Please ignore the commit with message `Enable linter on packages/ folder` and all commits older than that one.

@rafeca: Once #18885 is merged, would you mind re-requesting review on this PR so that people can easily review the changes that are specific to this PR?

spec/text-editor-component-spec.js
```diff
@@ -173,21 +173,19 @@ describe('TextEditorComponent', () => {
...
 
-        await component.getNextUpdatePromise()
-      }
+      TextEditorComponent.getScheduler().readDocument(() => {
+        // This will happen before the measurement phase of the update
+        // triggered above.
+        component.pixelPositionForScreenPosition(Point(11, Infinity))
+      })
+
+      await component.getNextUpdatePromise()
```
jasonrudolph: Does the linter mandate that we remove this block? 

This use of blocks is a pattern we employ to exercise multiple scenarios in a single test, without having to come up with unique variable names for each scenario.

jasonrudolph (approved):
Nice work! :zap:

> Most of the issues have been fixed automatically by running [`prettier-standard`](https://www.npmjs.com/package/prettier-standard) on that folder, the things that needed to be fixed manually are in separate commits.

I did not review the automatic fixes, but I did review each of the commits that performed manual fixes. I noted one question inline below, but overall, these changes look good. 👍

script/lib/lint-java-script-paths.js
```diff
@@ -6,29 +6,46 @@ const path = require('path')
...
+    path.join(CONFIG.repositoryRootPath, 'spec', 'fixtures', '**', '*.js')
+  ]
+
+  const [includePaths, excludePaths] = await Promise.all([
+    expandGlobPaths(globPathsToLint),
+    expandGlobPaths(globPathsToIgnore)
+  ])
+
+  const paths = includePaths.filter(
+    myPath => excludePaths.indexOf(myPath) === -1
```
Arcanemagus:
```suggestion
    myPath => !excludePaths.includes(myPath)
```
[Array.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) is almost always clearer for showing the intent of the code.

spec/application-delegate-spec.js
```diff
@@ -1,14 +1,14 @@
 /** @babel */
 
-import {it, fit, ffit, fffit, beforeEach, afterEach} from './async-spec-helpers'
+import { it } from './async-spec-helpers'
```
Arcanemagus: It's generally safer to _always_ import those. If you don't import them and you use one while trying to test something but forget to add it to the list again the test will silently "pass" and any errors will be hidden.

(The same comment applies to all the spec files where the "unused" imports are getting removed.)

spec/text-editor-component-spec.js
rafeca: Yes, this is mandated by the [`no-lone-blocks`](https://eslint.org/docs/rules/no-lone-blocks) rule, it only complains when there are no variables defined inside the block.

If this is a common pattern across the codebase I can disable this rule for the `spec/` folders (I have to figure out how to do this since we're using the `standard` rules for linting).

spec/application-delegate-spec.js
rafeca: ohh I was not aware of this, thanks for rising it! I've just tried it and that behaviour is very counter intuitive 😔

@Arcanemagus do you have any context about why does the test suite behaves like this? Can we just make the test fail in these scenarios? Having to import many unused modules that could potentially be used in the test just to overcome this issue in the test runner feels wrong 😕

I'm gonna hold on merging this PR until I get some clarity about it.

The only good thing is that now that we have the linter enabled in the specs, the editor should warn about using non-declared variables 🤷‍♂️

spec/text-editor-component-spec.js
jasonrudolph: > If this is a common pattern across the codebase I can disable this rule for the `spec/` folders (I have to figure out how to do this since we're using the `standard` rules for linting).

That works for me.

Alternatively, could we use an `eslint-disable` comment to tell the linter to ignore that particular rule in this particular spec?

spec/application-delegate-spec.js
Arcanemagus: If you attempt to use `async` code in a version of those functions that doesn't know it needs to wait for them it will fire off the `async` code... but not wait for it. So it will race through firing all that off and reach the end seeing nothing wrong and return. Then when the `async` code returns the failures are ignored since the spec has already "finished".

It's was _very_ confusing before I discovered that. We actually wrote [jasmine-fix](https://www.npmjs.com/package/jasmine-fix) and [use it](https://github.com/AtomLinter/linter-stylelint/blob/master/spec/linter-stylelint-spec.js#L7-L10) in all of the AtomLinter packages that have `async` specs in order to avoid this issue.

rafeca:
Once the CI tests pass I'm gonna merge https://github.com/atom/atom/pull/18917, which should address the issues that @Arcanemagus pointed out with `it`, `fit`, etc.

Would you folks be alright with merging this PR afterwards?

Arcanemagus:
Actually with  #18917 merged almost all the imports of `async-spec-helpers` should be able to be removed entirely since the base versions now support `async`/`await`. It looks like there are a few cases where it would still be needed though ([example](https://github.com/atom/atom/pull/18896/files/966fbc2e5819db4bfce887e2aebc0cf1e46fbafe..3c8005420cb98d18ba5d10b8ec7cb459160f56d5#diff-f0f549cb501e7daf4daa0fb536c14c2fR5)).

rafeca:
> Actually with #18917 merged almost all the imports of `async-spec-helpers` should be able to be removed entirely since the base versions now support `async`/`await`. It looks like there are a few cases where it would still be needed though ([example](https://github.com/atom/atom/pull/18896/files/966fbc2e5819db4bfce887e2aebc0cf1e46fbafe..3c8005420cb98d18ba5d10b8ec7cb459160f56d5#diff-f0f549cb501e7daf4daa0fb536c14c2fR5)).

Yup, I'm planning to remove all these imports in a follow-up PR 😄

Issue #18837, opened 2019-02-13
Uncaught TypeError: Cannot read property 'length' of undefined


Arcanemagus:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Sometimes when restoring a session (rendering content?) an uncaught error is thrown.

### Steps to Reproduce

1. Unknown, saw while restoring a session on restarting Atom.

**Expected behavior:** [What you expect to happen]
No crash

**Actual behavior:** [What actually happens]
#### Stack Trace

Uncaught TypeError: Cannot read property 'length' of undefined

```
At C:\Users\Lando\AppData\Local\atom\app-1.37.0-nightly0\resources\app\static\<embedded>:14

TypeError: Cannot read property 'length' of undefined
    at DisplayLayer.updateSpatialIndex (~/AppData/Local/atom/app-1.37.0-nightly0/resources/app/static/<embedded>:14:93829)
    at DisplayLayer.populateSpatialIndexIfNeeded (~/AppData/Local/atom/app-1.37.0-nightly0/resources/app/static/<embedded>:14:94974)
    at DisplayLayer.doBackgroundWork (~/AppData/Local/atom/app-1.37.0-nightly0/resources/app/static/<embedded>:14:81248)
    at TextEditor.doBackgroundWork (~/AppData/Local/atom/app-1.37.0-nightly0/resources/app/static/<embedded>:11:641)
```

**Reproduces how often:** [What percentage of the time does it reproduce?]
Unknown, however it looks like this has been happening since at least v1.32.2.

### Versions

**OS**: Microsoft Windows 10 Pro x64
```
Atom    : 1.37.0-nightly0
Electron: 3.1.3
Chrome  : 66.0.3359.181
Node    : 10.2.0
apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.37.0-nightly0
python 2.7.15
git 2.20.1.windows.1
[...]

Issue #18790, opened 2019-02-05
Renderer process crash in tests: libdyld.dylib


smashwilson:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

I'm seeing a consistent renderer process crash when running the core renderer process tests.

### Steps to Reproduce

1. `script/test --core-renderer`

**Expected behavior:**

The renderer process tests to complete.

**Actual behavior:**

The renderer process crashes midway:

```
[..]
Project > .addPath(path, options) > calls callbacks registered with ::onDidChangePaths [FAIL]
Project > .addPath(path, options) > doesn't add redundant paths [FAIL]
Project > .addPath(path, options) > doesn't add non-existent directories [FAIL]
Project > .addPath(path, options) > optionally throws on non-existent directories [FAIL]
Project > .removePath(path) > removes the directory and repository for the path [FAIL]
Project > .removePath(path) > does nothing if the path is not one of the project's root paths [FAIL]
Project > .removePath(path) > doesn't destroy the repository if it is shared by another root directory [FAIL]
Project > .removePath(path) > removes a path that is represented as a URI 2019-02-05 10:29:26.033 Atom Dev[35931:306634] Renderer process crashed, exiting
```

**Reproduces how often:**

100% on my machine, although the final started spec doesn't seem to be consistent.

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

### Additional Information

Here's the native stack from Console.app:

<details>

```
Thread 0 Crashed:: CrRendererMain  Dispatch queue: com.apple.main-thread
0   ???                           	0x0000002297c3359f 0 + 148575040927
1   ???                           	0x00000022975a57e3 0 + 148568168419
2   ???                           	0x00000022976ebdd6 0 + 148569505238
3   ???                           	0x00000022975ad1f7 0 + 148568199671
[...]

PR #18603, opened 2018-12-20
Upgrade to Electron 3.0

daviwil:
This change upgrades Atom to Electron 3 :electron:.  This gives us (and package authors) access to Chromium 66 and Node 10.

#### Related Issues

Fixes https://github.com/atom/atom/issues/17819.
Fixes https://github.com/atom/atom/issues/6666.
Fixes #18535.

#### TODO

- [x] Fix test failures
- [x] Check to see whether this PR needs to be merged: https://github.com/atom/tabs/pull/552

---

### General testing

- [x] Review issues tagged with [`electron`](https://github.com/search?utf8=%E2%9C%93&q=org%3Aatom+label%3Aelectron+is%3Aopen+is%3Aissue) to see if any of them are fixed
- [x] Verify that you can collaborate with Teletype
	- [x] Host using Atom's current stable release can collaborate with guest using Atom with the new version of Electron
	- [x] Guest using Atom's current stable release can collaborate with host using Atom with the new version of Electron
	- [x] Host and guest can collaborate when both are using Atom with the new version of Electron
  <details>
  <summary>Steps</summary>
	1. Open a file
	1. Share a portal
	1. Have a guest join the portal
	1. When guest edits the text in the file, verify that the host sees those changes
	</details>

#### Text Input/Keybindings

See how to [setup keyboard layouts](https://github.com/atom/maintainers-log/blob/master/how-tos/setup-keyboard-layouts.md).

- [x] IME not working (https://github.com/atom/atom/issues/14911)
	- [x] macOS
	- [x] Windows 10
	- [x] Ubuntu
  <details>
  <summary>Repro steps</summary>

  1. Install an IME layout, for example Pinyin simplified  
  1. Open Atom and use the layout you installed as the keyboard layout
  1. Example, if you picked Pinyin simplified type in `zhongwenshuru`

  **Expected:** in Pinyin simplified `中文输入` is expected

  gif

  **Issues we've seen:** Only shows latin first character and not every character typed.

  gif
  </details>

- [x] macOS: Key binding with composition characters are ignored (https://github.com/atom/atom/issues/15189)
  <details>
  <summary>Repro steps</summary>

  1. Open your keymap.cson file and add the following:
      ```cson
      'atom-text-editor':
        # use hjkl; vim-like when pressing alt (word by word, paragraph by paragraph)
        # or alt (letter by letter)
        'alt-h': 'core:move-left'
        'alt-j': 'core:move-down'
        'alt-k': 'core:move-up'
[...]

PR #18908, opened 2019-02-25
Revert to Electron 2 until we can overcome crashes on Electron 3

jasonrudolph:
### Description of the Change

#18603 and #18815 upgraded Atom from Electron 2 to Electron 3. (Specifically, we went from 2.0.16 to 3.1.13.) Those PRs satisfied the initial "smoke tests" enumerated in https://github.com/atom/atom/pull/18603#issue-240025844, but we've unfortunately been experiencing frequent crashes when using Atom day-to-day on Electron 3 (e.g., https://github.com/atom/atom/issues/18790, https://github.com/atom/atom/issues/18837, https://github.com/atom/atom/issues/18841) and when running on CI.

Ideally, we want to resolve those issues and move forward with Electron 3, but it will take some time to resolve those issues. In the meantime, we need Atom's master branch to provide a more usable version of Atom. :innocent: With that in mind, this pull request temporarily reverts #18603 and #18815, thus restoring Atom to its previous Electron version (2.0.16). By giving us a more stable master branch, this pull request will unblock our ability to ship Atom 1.36-beta0.

### Alternate Designs

Instead of reverting all of the changes in #18603 and #18815, I first attempted to only change the Electron versions in `package.json`. (See https://github.com/atom/atom/compare/042fb45...f5349eb.) To get Atom to build locally after that change, I also had to revert to the previous version of `script/lib/generate-startup-snapshot.js`, as explained in the commit message in 2fd935652ef7891906b69804514b53682f5df56e. Those changes were sufficient to _build_ Atom, but a [large number of tests fail related to `text-editor-component.js`](https://github.visualstudio.com/Atom/_build/results?buildId=33992).

### Possible Drawbacks

We'll have to undo these reversions when we're ready to move forward with the Electron 3 upgrade.

### Verification Process

- [x] macOS: Verify that Atom launches, and that you can create, save, and re-open files.
- [x] Windows: Verify that Atom launches, and that you can create, save, and re-open files.
- [x] Linux: Verify that Atom launches, and that you can create, save, and re-open files.
- [x] Ensure all tests are passing -- https://github.com/atom/atom/pull/18908#issuecomment-467554406


script/vsts/release-branch-build.yml
```diff
@@ -1,6 +1,7 @@
 trigger:
 - master
 - 1.* # VSTS only supports wildcards at the end
+- electron-*
```
smashwilson: 👍  This will be useful in the future. We'll have to make sure we don't revert _it_ away when we revert these reversions.

spec/decoration-manager-spec.coffee
```diff
@@ -1,7 +1,6 @@
 DecorationManager = require '../src/decoration-manager'
 TextEditor = require '../src/text-editor'
 
-# Tests crash the renderer process on Electron 3.0, disabling for now.
```
smashwilson: Hrm, suspicious. Something we forgot?

spec/workspace-spec.js
```diff
@@ -2086,7 +2086,7 @@ describe('Workspace', () => {
           })
 
           runs(() => {
-            fs.renameSync(path.join(projectPath, 'git.git'), path.join(projectPath, '.git'))
+            fs.rename(path.join(projectPath, 'git.git'), path.join(projectPath, '.git'))
```
smashwilson: This change looks odd to me. Was `fs.rename` always synchronous in the old version of fs-extra?

smashwilson (approved):
👍 Nothing in here that should break the changes we've made since then, at a glance anyway. I guess we'll find out pretty quickly what we missed 😅

spec/workspace-spec.js
daviwil: Prior to Node 10 `fs.rename` accepted a callback parameter and didn't complain if you didn't provide a callback. In Node 10 it seems that the callback is now mandatory if you use the non-Sync variant. I suppose the function worked as synchronous before if no callback was specified?

jasonrudolph:
> * [ ]  Ensure all tests are passing

screen shot 2019-02-26 at 1 18 55 pm

The statuses above show that the `Atom Pull Requests` build for this branch is failing, but it's actually passing: https://github.visualstudio.com/Atom/_build/results?buildId=34110

2_26_19__1_20_pm

I'm not sure why visualstudio.com isn't reporting the passing build to GitHub, but solving that is out of scope for this pull request. 😇

script/vsts/release-branch-build.yml
jasonrudolph: Good point. I've opened https://github.com/atom/atom/pull/18916 to revert these reversions 🥴💫😵, but this change is still intact in that pull request.

PR #18917, opened 2019-02-26
Override global jasmine spec functions

rafeca:
## Issue or RFC Endorsed by Atom's Maintainers

[Comment on previous PR](https://github.com/atom/atom/pull/18896#discussion_r260396102).

## Description of the Change

Currently, if a spec uses the standard `it` function on an async test (the one that's stored in the global object), that test will always pass. This happens because the jasmine version checked into Atom does not natively support tests that return promises, so the failure happens asynchronously and the test runner cannot detect it.

To overcome this limitation, and because of the many repos that Atom use, a bunch of `async-spec-helpers.js` files have emerged (right now there are [31 of them](https://github.com/search?p=1&q=org%3Aatom+filename%3Aasync-spec-helpers.js&type=Code)) which implement custom `it`, `fit`, `beforeEach`, `afterEach`,... methods with support for async tests.

This can be confusing since some tests are currently using the builtin jasmine methods (mostly old tests that don't use async functions yet) and some other tests use the overriden versions located across the different `async-spec-helpers.js` files.

This PR modifies the builtin `jasmine` functions on the global test runner, so any Atom test will be able to use async functions, making the behaviour uniform across all tests and addressing the [valid concern](https://github.com/atom/atom/pull/18896#discussion_r260396102) stated by @Arcanemagus in #18896 .

As a nice side-effect, this PR will allow us to remove many of the `async-spec-helpers.js` files that we have hanging around 😃 

## Alternate Designs

Ideally we should upgrade the version of Jasmine that Atom tests are using: since v2.7 Jasmine [supports async tests](https://jasmine.github.io/tutorials/async#async-keyword).

Doing the upgrade may be very tricky, since Atom uses its own checked in version of Jasmine which has local modifications and [hasn't been updated since 6 years ago](https://github.com/atom/atom/commits/master/vendor/jasmine.js).

In the case that we want to do an upgrade of the test runner, it would be worth to explore how feasible would be to switch to a more advanced test runner like [Jest](https://jestjs.io/) which will probably provide many benefits like reduced test times or better developer experience (though it would be tricky to integrate since Atom tests are run in the Atom runtime, while Jest currently only runs on Node.js).

## Possible Drawbacks

Overriding globals is ugly and I should feel bad about it 🤡

## Verification Process

Check that all tests in CI keep passing.

## Release Notes

N/A.

### DIFF
--- a/spec/jasmine-test-runner.coffee
+++ b/spec/jasmine-test-runner.coffee
@@ -10,6 +10,16 @@ module.exports = ({logFile, headless, testPaths, buildAtomEnvironment}) ->
   window[key] = value for key, value of require '../vendor/jasmine'
   require 'jasmine-tagged'
 
+  # Rewrite global jasmine functions to have support for async tests.
+  # This way packages can create async specs without having to import these from the
+  # async-spec-helpers file.
+  global.it = asyncifyJasmineFn global.it, 1
+  global.fit = asyncifyJasmineFn global.fit, 1
+  global.ffit = asyncifyJasmineFn global.ffit, 1
+  global.fffit = asyncifyJasmineFn global.fffit, 1
+  global.beforeEach = asyncifyJasmineFn global.beforeEach, 0
+  global.afterEach = asyncifyJasmineFn global.afterEach, 0
+
   # Allow document.title to be assigned in specs without screwing up spec window title
   documentTitle = null
   Object.defineProperty document, 'title',
@@ -58,6 +68,28 @@ module.exports = ({logFile, headless, testPaths, buildAtomEnvironment}) ->
 
   jasmineEnv.execute()
   promise
+
+asyncifyJasmineFn = (fn, callbackPosition) ->
+  (args...) ->
+    if typeof args[callbackPosition] is 'function'
+      callback = args[callbackPosition]
+
+      args[callbackPosition] = (args...) ->
+        result = callback.apply this, args
+        if result instanceof Promise
+          waitsForPromise(-> result)
+
+    fn.apply this, args
+
+waitsForPromise = (fn) ->
+  promise = fn()
+
+  global.waitsFor('spec promise to resolve', (done) ->
+    promise.then(done, (error) ->
+      jasmine.getEnv().currentSpec.fail error
+      done()
+    )
+  )
 
 disableFocusMethods = ->
   ['fdescribe', 'ffdescribe', 'fffdescribe', 'fit', 'ffit', 'fffit'].forEach (methodName) ->


alawi247:
need help

spec/jasmine-test-runner.coffee
```diff
@@ -10,6 +10,16 @@ module.exports = ({logFile, headless, testPaths, buildAtomEnvironment}) ->
   window[key] = value for key, value of require '../vendor/jasmine'
   require 'jasmine-tagged'
 
+  # Rewrite global jasmine functions to have support for async tests.
+  # This way packages can create async specs without having to import these from the
+  # async-spec-helpers file.
+  global.it = asyncifyJasmineFn global.it
+  global.fit = asyncifyJasmineFn global.fit, 1
```
jasonrudolph: Can you help me understand why we need to specify the `posCallback` arg for `fit`, but not for `ffit` and `fffit`?

jasonrudolph (commented):
Thanks for taking this on!

I have one question about the implementation, but this is looking good overall. If it would be easier to to talk through this over video chat, just let me know, and we'll make it happen. 😄

spec/jasmine-test-runner.coffee
rafeca: `fit` is the only of these methods that does not have the callback as the last argument ([original source code](https://github.com/atom/jasmine-focused/blob/master/src/jasmine-focused.coffee#L30)), so I need to override the position of the callback.

Now that I look at it, this may be confusing for the future reader, so I'm gonna update the PR to make the position of the callback required and always pass it.

jasonrudolph:
@Arcanemagus: Given your [familiarity with this space](https://github.com/atom/atom/pull/18896#discussion_r260528195), would you mind reviewing this pull request. If these changes look good to you, I'm 👍 on merging.

Arcanemagus (approved):
Tested Atom using the code from this branch and confirmed that this does indeed translate over to allowing any package spec to use `async`/`await`.

I had to add this block ([source](https://github.com/steelbrain/jasmine-fix/blob/c7bc55cc9d6de1f4493f3f7a1930c60b80c00557/index.js#L15-L30)) to a few of the [test package](https://github.com/AtomLinter/linter-eslint/blob/122d048ecf36d094870aba7f0be4a2cc1cb9c444/spec/linter-eslint-spec.js)'s `beforeEach` blocks, but I think this is to be expected and not something we want to mess with here because the timing stuff is actually utilized for the specs here:
```js
Object.keys(jasmine.Clock.real).forEach((key) => {
  window[key] = jasmine.Clock.real[key];
})
jasmine.unspy(Date, 'now');
```

rafeca:
Note that this PR can potentially cause failures on tests that were passing previously (an example is the one I fixed in https://github.com/atom/snippets/pull/290 ).

This can happen whenever an error was swallowed by a `Promise` that was not awaited or whenever a `Promise` that never gets resolved is returned on a spec method. IMHO failing in these situations is a good thing, and so far the only test that I've seen failing because of this is the one that I fixed on https://github.com/atom/snippets/pull/290, so I'm gonna go ahead and merge this PR.

rafeca:
I'm gonna land this PR with the failing test (it's a flaky check and after 3 retries I keep getting different failures)

cpojer:
@rafeca If you are considering using Jest for Atom, I recommend checking out the electron runner: https://github.com/facebook-atom/jest-electron-runner

We use this at FB for running Nuclide's tests using Jest. It works!

rafeca:
Hii @cpojer! 👋

Wow that looks promising! I'll take a look and check with @aaronabramov 😃

Thanks for the pointer

PR #18919, opened 2019-02-27
⬆️ fuzzy-finder@1.9.2

rafeca:
Includes https://github.com/atom/fuzzy-finder/pull/366, which reduces the startup time of the fuzzy finder on large repositories.

---------

[*(list of changes between `fuzzy-finder@1.9.1` and `fuzzy-finder@1.9.2`)*](https://github.com/atom/fuzzy-finder/compare/v1.9.1...v1.9.2)

Issue #10037, opened 2015-12-12
Add ability to save as default filetype


grenzor:
Add ability to save as default filetype. So I can set to ".html" and all files onwards will have the automatic extension .html appended to the filename when trying to save the file.

grenzor:
Anyone able to add this feature?

PR #18924, opened 2019-02-28
Update versions of fuzzy-finder and snippets

rafeca:
On https://github.com/atom/atom/pull/18923 and https://github.com/atom/atom/pull/18919 I forgot to change the second place where the package versions are specified. This PR fixes that


ghost (commented):
Duplicate of #10037

Issue #18978, opened 2019-03-13
show the first line of the file contents in the file tab for unsaved file


modernNeo:
## Summary
I am trying to transition from Sublime which has a feature I like where if a file is not yet saved, instead of showing the string "Untitled" as Atom does in the file tab or something similar, it instead displays as much of the first line as is possible on the file tab. I tried to look through all of atom's customisation for this feature and couldn't find it. Did I miss it or is it something that Atom does not have?
### Sublime Text
![Screenshot from 2019-03-12 20-37-29](https://user-images.githubusercontent.com/14242088/54252183-18857d80-4507-11e9-816c-2cb83cef2c9a.png)
### Atom
![Screenshot from 2019-03-12 20-37-58](https://user-images.githubusercontent.com/14242088/54252348-ab261c80-4507-11e9-9259-279b684433cb.png)

## Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

It provides a nice way to distinguish each unsaved files from each other rather than having to go through each unsaved file specifically to see if it has the contents we want.

## Describe alternatives you've considered

A clear and concise description of the alternative solutions you've considered. Be sure to explain why Atom's existing customizability isn't suitable for this feature.

I am unsure if any alternatives exist, if they do, please feel free to close this Issue and point me in the right direction.

rsese:
Thanks for the suggestion @modernNeo!  While this does seem like it could be useful for some people, as mentioned in the issue template the team is unlikely to work on feature requests right now (in favor of things like bug fixes).  Additionally:

> I am unsure if any alternatives exist, if they do, please feel free to close this Issue and point me in the right direction.

I searched a bit and found this package that seems to do what you want:

https://atom.io/packages/tab-title

I installed it and tried it quickly and it looks like it works.

So since this isn't something the team would work on and since the functionality can be implemented outside of core Atom (and is available now with https://atom.io/packages/tab-title) , we'll go ahead and close this.

modernNeo:
@rsese thanks for the link, thats exactly what I needed! 👍

PR #18980, opened 2019-03-13
Added feature: show the first line of the file contents in the file tab for unsaved file

budavarimark:
It's not (yet) endorsed by atom's maintainers but I thinks it's good idea to make it possible. (#18978)
### Description of the Change
If the user is doesn't want to see "Untitled" tab everywhere if the user opening unsaved files, then the user is now able to enable this option in the "Settings" under the "Editor" tab.

### Alternate Designs
\-

### Possible Drawbacks
Nothing.

### Verification Process
- How did you verify that all new functionality works as expected?

0, **Go to settings and check or uncheck "Show File Content Instead Of Untitled"**
1, **Opening new file.** 
The tab title should be 'untitled', because it's empty.

**_If the "Show File Content Instead Of Untitled" option is checked._**
2, **Write something in the first row.**
The tab title should change to the first row content.
3, **Delete everything from first row.**
The tab title should be 'untitled' again.
4, **Write something in the first row.**
The tab title should change to the first row content again.
5, **Save the file**
The tab title should be the file name

**_If the "Show File Content Instead Of Untitled" option is unchecked._**
2, **Write something in the first row.**
Nothing happens
3, **Delete everything from first row.**
Nothing happens
4, **Write something in the first row.**
Nothing happens
5, **Save the file**
The tab title should be the file name

### Release Notes
Added option: Show the first line of the file contents in the file tab for unsaved file

### DIFF
--- a/src/config-schema.js
+++ b/src/config-schema.js
@@ -507,6 +507,11 @@ const configSchema = {
         default: false,
         description: 'Allow the editor to be scrolled past the end of the last line.'
       },
+      showFileContentInsteadOfUntitled: {
+        type: 'boolean',
+        default: false,
+        description: 'Show the file first row as a file name instead of untitled if the file is not saved yet'
+      },
       undoGroupingInterval: {
         type: 'integer',
         default: 300,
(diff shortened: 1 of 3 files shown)


rsese:
Thanks for the contribution @budavarimark!  We just closed https://github.com/atom/atom/issues/18978, see https://github.com/atom/atom/issues/18978#issuecomment-472523478 for more details, but it looks like this functionality is already available through an existing community package.  Additionally, the team is trying to concentrate as much as possible on things like bug fixes rather than taking on new functionality so we'll go ahead and close this out but thank you again for the contribution.

Issue #18981, opened 2019-03-13
Reassigned shortcut is not removed from the menu


malnvenshorn:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Reassigned shortcut is not removed from the menu. Even after restarting Atom it shows the shortcut next to the new and old entry.

### Steps to Reproduce

1. Reassign shortcut, e.g. ctrl+shift+s
```
'body':
    'ctrl-shift-S': 'window:save-all'
```
2. Open menu

**Expected behavior:**

![expected](https://user-images.githubusercontent.com/13801805/54278660-3e586400-4593-11e9-8103-1ba53a2dc4ba.png)

**Actual behavior:**

![actual](https://user-images.githubusercontent.com/13801805/54278668-43b5ae80-4593-11e9-8dcf-cbfbb116d381.png)

**Reproduces how often:** 100%

### Versions

```
$ atom --version
Atom    : 1.35.0
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

```
apm --version
apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.35.0
python 2.7.15
git 2.21.0
```

```
$ uname -a
Linux archdesk 5.0.0-arch1-1-ARCH #1 SMP PREEMPT Mon Mar 4 14:11:43 UTC 2019 x86_64 GNU/Linux
```

### Additional Information

None

rsese:
Thanks for the report!  I'm pretty positive this has been reported before but maybe I'm wrong,  I can't find an issue with this report in my searching.  Otherwise, reproduced with 1.34.1 on Ubuntu 18.04.

PR #18987, opened 2019-03-14
fixed "Reassigned shortcut is not removed from the menu"

budavarimark:
### Identify the Bug

If the user is writing a keystroke in the keymap.cson, then it's priority is 100. The default keystroke priority is 0. 
So if the keystroke is already exists in default config, then it's will be overwritten by the user.
So in theory, if the user is overwrite one default keystroke then the menu should change, but actually there is no priority checking so there will be a duplicate.
Just like in #18981

### Description of the Change
I wrote a function which is checking the priorities and delete the lower priority one from the _"keystrokesByCommand_" variable if there is any keystroke duplication.

### Verification Process

Write this to keymap.cson
```
'body':
      'ctrl-shift-S': 'window:save-all'
```
The actual behavior is the expected behavior. 
expected

<hr>
Delete the code from keymap.cson and the default keymapping is also work.

expected2

### Release Notes
Fixed an issue where reassigned shortcut is not removed from the menu

### DIFF
--- a/src/menu-manager.coffee
+++ b/src/menu-manager.coffee
@@ -145,6 +145,14 @@ class MenuManager
 
     false
 
+  checkKeyStrokePriority = (binding, keystrokesByCommand, currentPriorities) ->
+    for command, keystrokes of keystrokesByCommand
+      for id, keystroke of keystrokes
+        if binding.keystrokes is keystroke and currentPriorities[command] < binding.priority
+          delete keystrokesByCommand[command][id]
+
+    return keystrokesByCommand
+
   # Public: Refreshes the currently visible menu.
   update: ->
     return unless @initialized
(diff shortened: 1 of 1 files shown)


50Wliu:
Thanks for your submission!

We require automated tests for all pull requests that include new or changed code. We do this so that we can ensure that we don't accidentally break your shiny new code the next time we or some other contributor submits a change. If you need help writing automated tests for Atom, check out [Discuss, the official Atom and Electron message board](https://discuss.atom.io) or [join the Atom Slack team](http://atom-slack.herokuapp.com/). There are a bunch of helpful community members that should be willing to point you in the right direction.

Also, are there any potential drawbacks with this approach? For example, if I have two mutually exclusive selectors (`.tree-view` and `atom-text-editor`), and use the same keybinding for commands under those selectors, would both keybindings remain active?

sadick254:
Thank you for your issue!

We haven't gotten a response to our test addition request in our comment above. Without tests it would be difficult for us to prevent regression should this make it to Atom stable. I'm going to close this but don't hesitate to reach out if you find time to add the requested tests, we'll be happy to reopen the issue.

Issue #18488, opened 2018-11-24
Error when running `atom -version` instead of `atom --version`


kane-sei:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When I made a typo atom `-version` for atom `--version` , the following error dialog was displayed.

### Steps to Reproduce

1. [First Step]
```
atom -version
```

**Expected behavior:** [What you expect to happen]

I think this should be return
```
invalid option -version
```
in the terminal.
(And should not display the error dialog)

**Actual behavior:** [What actually happens]

The following program error dialog is displayed.
![2018-11-24 14 28 13](https://user-images.githubusercontent.com/45293757/48965801-697c6f00-f007-11e8-8563-28dc9fe0d1a6.png)

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions

Atom : 1.32.2
Electron: 2.0.9
Chrome : 61.0.3163.100
Node : 8.9.3

OS MacOS 10.13.6

PR #18988, opened 2019-03-14
fixed "Error when running `atom -version` instead of `atom --version`"

budavarimark:
### Identify the Bug
#18488

When the user misspell `atom --version` and write `atom -version` instead, then the yargs understand this command as -v -e -r -s -i -o -n (and also as 'version', but the error happens before args parse) where -r is resource-path.
Because there is no resource-path the JavaScript error pop up.

### Description of the Change
Added one simple condition. Check if the resource-path is string if it's exists.

### Verification Process
Now if the user write -version instead of --version this happens:
image

### Release Notes
Fixed and issue where error happened if the user ran `atom -version` instead of `atom --version`


rsese:
Thanks @budavarimark :+1:  Can you add tests for this fix?  As mentioned in the `Requirements` section above, the team will need tests to be added before they can review and tests will help guard against future changes breaking your work 😄

rafeca:
Hey! 👋

With ❤️, the Atom team.

ElianVanCutsem:
I think this request should be closed since it's already implemented

PR #19013, opened 2019-03-19
Add GITHUB_TOKEN env var to the npm install step (take 2)

rafeca:
Trying to fix the Azure build failures: https://dev.azure.com/github/Atom/_build/results?buildId=35928

### DIFF
--- a/script/vsts/platforms/linux.yml
+++ b/script/vsts/platforms/linux.yml
@@ -14,6 +14,7 @@ jobs:
   steps:
   - script: script/build --create-debian-package --create-rpm-package --compress-artifacts
     env:
+      GITHUB_TOKEN: $(GITHUB_TOKEN)
       ATOM_RELEASE_VERSION: $(ReleaseVersion)
     displayName: Build Atom
 
--- a/script/vsts/platforms/macos.yml
+++ b/script/vsts/platforms/macos.yml
@@ -27,6 +27,7 @@ jobs:
       fi
     displayName: Build Atom
     env:
+      GITHUB_TOKEN: $(GITHUB_TOKEN)
       IS_RELEASE_BRANCH: $(IsReleaseBranch)
       IS_SIGNED_ZIP_BRANCH: $(IsSignedZipBranch)
       ATOM_RELEASE_VERSION: $(ReleaseVersion)
--- a/script/vsts/platforms/windows.yml
+++ b/script/vsts/platforms/windows.yml
@@ -52,6 +52,7 @@ jobs:
         )
       )
     env:
+      GITHUB_TOKEN: $(GITHUB_TOKEN)
       BUILD_ARCH: $(buildArch)
       ATOM_RELEASE_VERSION: $(ReleaseVersion)
       ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL: $(ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL)


rafeca:
Ok, got the Azure pipelines correctly building... now it's time for Travis 😅

rafeca:
Ok, it's going to be challenging to fix the issue with Travis, at least for external PRs: the main issue is that Travis [does not expose env variables to external PRs for security reasons](https://docs.travis-ci.com/user/environment-variables/).

I think that the behaviour of Travis is correct in most of the cases: by exposing env vars to PRs somebody could create a malicious PR and get the value of any variable without any code review from a repo owner... In this case, though, the vector attack is almost non-existant: the `GITHUB_TOKEN` that we're using as an env var is a completely unprivileged token with only read-access to the Atom public repos which allows us to get a higher rate limit from the GitHub API than unauthenticated requests... so there's not much value on stealing that token.

In the case of Azure pipelines and AppVeyor I could expose the single `GITHUB_TOKEN` env var to PRs, so the only problem remains with Travis.

I see 2 possible solutions here:

### Find out if we can disable Travis builds from PRs (or from everywhere)

From a previous conversation with @jasonrudolph :

> We rely on Travis for Linux builds of PRs and release branches

Since we also do Linux builds on Azure right now, we may be able to disable the Travis ones, but somebody with more familiarity on our build system will know better.

### Stop using `vscode-ripgrep` to download the `ripgrep` binaries

This means that we'll have to find another way to either download the binaries for each platform or build them ourselves during our CI builds.

If follow this path, we'll have to take care of a few things that `vscode-ripgrep` currently does for us:

1. Build a binary for each platform ([more info](https://github.com/roblourens/ripgrep/wiki).
2. Update the binaries every time there's a new version of `ripgrep`.
3. Be sure that we don't include multiple `ripgrep` binaries for different platforms on each installation, since each binary is quite big.

I'm not sure which solution it's going to be easier to implement, but my gut tells me that's going to be the first one...

rafeca:
Ups, it seems like atom.io uses the linux artifacts [created by Travis builds](https://github.com/github/atom.io/blob/6d2ad057ba189d1806c57b3403181df618a8deda/app/models/release_index.rb#L200) 😞.

jasonrudolph:
> Find out if we can disable Travis builds from PRs

I think this might be the way to go. A few thoughts:

- For *all* PRs, we'd like to be able to verify that the changes can still build successfully on Linux. The Azure DevOps Linux builds already meet this need. ([example](https://github.visualstudio.com/Atom/_build/results?buildId=36046&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=73c4289c-110b-5085-4a98-4009fa5cc9c8&lineStart=3926&lineEnd=3927&colStart=1&colEnd=1))
- For release branches inside the atom/atom repo (e.g., the `1.35-releases` branch), we need to publish build artifacts. The Travis CI build currently performs this duty. The Azure DevOps build does not yet perform this duty. For the near term, it's probably easiest to continue having the Travis CI build perform this duty. Since this is only needed for branches inside the atom/atom repo (i.e., not from forks), these builds will have access to the protected environment variables. With that in mind, if there's a way to tell Travis CI to *not* build PRs from forks (and instead just let Azure DevOps run those builds), then I think that could meet our needs.

rafeca:
Thanks for the response @jasonrudolph!

> The Travis CI build currently performs this duty. The Azure DevOps build does not yet perform this duty.

Makes sense... Just to understand better, what we would need to do to generate linux releases from Azure DevOps is to make both `github/atom.io` and `atom/atom-release-publisher` use the artifacts generated from Azure right?

> For the near term, it's probably easiest to continue having the Travis CI build perform this duty.

I agree, from what I saw the changes needed to `atom.io` and the release publisher website are not trivial...

> With that in mind, if there's a way to tell Travis CI to not build PRs from forks (and instead just let Azure DevOps run those builds), then I think that could meet our needs.

I think that this should be possible, I'm gonna take a look at it 😄

jasonrudolph:
> Just to understand better, what we would need to do to generate linux releases from Azure DevOps is to make both `github/atom.io` and `atom/atom-release-publisher` use the artifacts generated from Azure right?

@rafeca: That sounds right. As prior art, https://github.com/atom/maintainers-log/issues/148 tracked the process of migrating the macOS builds from CircleCI to Azure DevOps, and most of the same concerns likely apply to migration the Linux builds from Travis CI to Azure DevOps.

https://github.com/github/atom.io/pull/1267 and https://github.com/atom/atom-release-publisher/pull/14 handled the atom.io and atom-release-publisher tasks respectively.

rafeca:
Thanks for the background! That's gonna be useful if we ever want to do that migration 😄

Something I realized also is that the `atom-release-publisher` logic assumes that there's a build status for each platform ([code](https://github.com/atom/atom-release-publisher/blob/89c2f6d1b58f8d8ddf86821305b7de5bb837f781/src/github-repository.js#L122-L129)), which is currently the case:

Screen Shot 2019-03-20 at 14 36 26

If we ever want to build releases for multiple platforms from Azure we'd have to change either that logic or make Azure report a separate status for each platform.

rafeca:
Ok, I've been able to disable Travis builds on Pull Requests and after force-pushing a commit on this one it seems that Travis doesn't pick it up anymore 😄 

Once the other builds finish I'll merge this PR and check that everything is fine on `master`

jasonrudolph:
> Ok, I've been able to disable Travis builds on Pull Requests and after force-pushing a commit on this one it seems that Travis doesn't pick it up anymore 😄

Oh, cool. Did that require a configuration change on https://travis-ci.com?

rafeca:
Yup exactly, I disabled the following checkbox on the project settings for Atom:

Screen Shot 2019-03-20 at 15 02 15

PR #19016, opened 2019-03-20
Convert line endings when copy and pasting

### LINKED ISSUE #8365: Copy-paste doesn't work properly with Unix-style line endings under Windows
Copying (Ctrl+C-Ctrl+V) from a file with Unix-style line ending (LF) doesn't work as expected, since line endings are _preserved_ on the clipboard.

This causes a wide range of issues:
- you can't paste text in an application that doesn't handle LF (notepad for example, but if you're using vim over PuTTY that also doesn't work)
- even if the other app handles LF, you have to be consistent, so you can't paste text from an CRLF file into a LF one or vice versa

This issue is briefly discussed in #2822, and it is the reason of #5808.

**All other editors convert line endings in text placed on the clipboard** (Sublime Text, jEdit, etc.), so in this regard Atom behaves rather strangely compared to other editors.

The proposed solution in #2822 (configurable line endings) is insufficient since it's based on the assumption that _all_ files edited by the user are using the same line ending.
For example, if I've got a file with LF line endings and I'd like to use copy-paste, then I **have to convert it to CRLF** and then **copy the text** I want, and **then convert it back to LF** to preserve the original format. I wouldn't call editing LF files on Windows a rare use case, especially among Atom's users.

Atom should convert any text's line ending to the platform's native line ending _before_ it is placed on the clipboard. This conversion should only be done on the content that is placed on the clipboard—the text in the buffer should not be changed.

winstliu:

#8365, endorsed by @maxbrunsfeld [here](https://github.com/atom/atom/issues/8365#issuecomment-433170963).

### Description of the Change

Whenever text is copied to the clipboard, all line endings are converted to the underlying OS's default line endings (CRLF on Windows, LF otherwise). Fixes #8365.

### Alternate Designs

I could have done `const os = require('os')` and then used `os.EOL` instead of hardcoding the line endings in `clipboard.js`, but that didn't seem worth it.

This could also be implemented as a Windows-only feature, which I believe is how VSCode does it.

### Possible Drawbacks

Sometimes you may want to explicitly copy the text as-is without conversions, and this will prevent that.

### Verification Process

On Windows, I copied text with LF line endings and pasted them into http://www.unit-conversion.info/texttools/hexadecimal/, which confirmed that the pasted text had `0d 0a` (CRLF).

### Release Notes

Line endings are now converted when copy and pasting.

### DIFF
--- a/spec/clipboard-spec.js
+++ b/spec/clipboard-spec.js
@@ -1,32 +1,41 @@
 describe('Clipboard', () => {
   describe('write(text, metadata) and read()', () => {
     it('writes and reads text to/from the native clipboard', () => {
-      expect(atom.clipboard.read()).toBe('initial clipboard content')
-      atom.clipboard.write('next')
-      expect(atom.clipboard.read()).toBe('next')
-    })
+      expect(atom.clipboard.read()).toBe('initial clipboard content');
+      atom.clipboard.write('next');
+      expect(atom.clipboard.read()).toBe('next');
+    });
 
     it('returns metadata if the item on the native clipboard matches the last written item', () => {
-      atom.clipboard.write('next', {meta: 'data'})
-      expect(atom.clipboard.read()).toBe('next')
-      expect(atom.clipboard.readWithMetadata().text).toBe('next')
-      expect(atom.clipboard.readWithMetadata().metadata).toEqual({meta: 'data'})
-    })
-  })
+      atom.clipboard.write('next', { meta: 'data' });
+      expect(atom.clipboard.read()).toBe('next');
+      expect(atom.clipboard.readWithMetadata().text).toBe('next');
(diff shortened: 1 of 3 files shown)


50Wliu:
Note: this won't affect pasting in Atom that much: copying LF and pasting it into another LF file will still result in LF line endings, not CRLF.

50Wliu:
Hmm, this breaks some tests. I can add a boolean option whether to do the conversion or not.

rafeca:
Hey! 👋

With ❤️, the Atom team.

sadick254 (approved):
Hey @50Wliu, I have just resolved some conflicts on this and I think it's good to go. Thank you for your contribution.

50Wliu:
@sadick254 feel free to merge, I really shouldn't have merge permissions anymore :).

Issue #19007, opened 2019-03-18
Try upgrading libgit2 to resolve renderer crashes


jasonrudolph:
As a potential resolution for the libgit2 crashes in https://github.com/atom/atom/issues/18990, let's upgrade to a newer version of libgit2.

### Background

As described in https://github.com/atom/atom/issues/18990#issuecomment-474078139, many of the build failures reported in https://github.com/atom/atom/issues/18990 are due to a crash happening inside libgit2.

For example, in [this build](https://github.visualstudio.com/Atom/_build/results?buildId=34137), the tree-view tests fail due to a renderer crash [here](https://github.visualstudio.com/Atom/_build/results?buildId=34137&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&taskId=bde16fc7-7659-5b1c-12aa-0110866d993c&lineStart=2789&lineEnd=2790&colStart=1&colEnd=1), and the build artifacts include a crash report in the `crash-reports.zip` file [here](https://github.visualstudio.com/_apis/resources/Containers/1160191?itemPath=crash-reports.zip%2FAtom%20Dev%20Helper_2019-02-26-204525_Mac-345.crash). The crash report points to code inside libgit2:

```
...
Crashed Thread:        17
...
Thread 17 Crashed:
0   libsystem_c.dylib             	0x00007fff68f39232 strlen + 18
1   libsystem_c.dylib             	0x00007fff68f7da86 __vfprintf + 4716
2   libsystem_c.dylib             	0x00007fff68fa5059 __v2printf + 473
3   libsystem_c.dylib             	0x00007fff68f8a34b _vsnprintf + 415
4   libsystem_c.dylib             	0x00007fff68f8a3fe vsnprintf + 80
5   git.node                      	0x0000000112a22697 git_buf_vprintf + 183 (buffer.c:387)
6   git.node                      	0x0000000112a3c8b9 giterr_set + 203 (errors.c:56)
7   git.node                      	0x0000000112a63477 git_path_diriter_next + 164 (path.c:1316)
8   git.node                      	0x0000000112a4fc5c fs_iterator__expand_dir + 577 (iterator.c:1256)
9   git.node                      	0x0000000112a4df8c fs_iterator__advance_into + 75 (iterator.c:1434)
10  git.node                      	0x0000000112a32f73 git_diff__from_iterators + 3078 (iterator.h:181)
[...]

PR #19021, opened 2019-03-21
⬆️ git-utils@5.5.0

rafeca:
This new version contains libgit2/libgit2@9084712 and we're hoping to get some of the crashes and flaky tests solved ([more info](https://github.com/atom/atom/issues/19007)).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3040,9 +3040,9 @@
       }
     },
     "git-utils": {
-      "version": "5.2.1",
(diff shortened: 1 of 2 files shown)


rafeca:
After the investigation done in https://github.com/atom/atom/issues/19007, we're cautiously optimistic that this will fix some of the crashes that we had in our CI and we don't think it's going to introduce new crashes

PR #19054, opened 2019-03-26
Do not delete global scope setting when resetting user settings

### LINKED ISSUE #18372: atom.config.set will erase settings when scopeSelector is "*"
### Prerequisites

### Description

```js
atom.config.set("test.test", true, {scopeSelector: "*"}); 
```

The above command will erase all settings in the "*" selector and add `test.test: true`

### Steps to Reproduce

1. Back up your config.cson file.
2. Open debug console. (<kbd>Ctrl/Cmd</kbd>+<kbd>Shift</kbd>+<kbd>i</kbd>)
3. Paste the above command into the console.
4. Press <kbd>Enter</kbd>.

**Expected behavior:**

`test.test: true` is added to your config.cson file under the "*" scope

**Actual behavior:**

All settings under "\*" scope are erased, then `test.test: true` is added to your config.cson file under the "*" scope

### Versions

```
Windows 10 Pro Insider Preview

Atom    : 1.33.0-beta0
Electron: 2.0.11
Chrome  : 61.0.3163.100
Node    : 8.9.3

apm  2.1.2
npm  6.2.0
node 8.9.3 x64
atom 1.33.0-beta0
python 2.7.14
git 2.19.1.windows.1
visual studio 2015
```

### Additional Information

related https://github.com/atom/atom/issues/14909#issuecomment-434554388

rafeca:
### Identify the Bug

This PR fixes https://github.com/atom/atom/issues/18372

### Description of the Change

**Edit**: This PR changes the `atom.config.set()` method to never use the scoped store when setting global config params (even when using the `scopeSelector`). This is to prevent incorrect values to be stored in the user config file, which happens when the scoped store receives a call to set a global property (since it does not hold any global properties).

~~This PR changes the logic of the `atom.config.resetUserSettings()` method to not delete the globally scoped settings (the ones under the scope `*`) before storing them on the [`scoped-settings-store`](https://github.com/atom/scoped-property-store).~~

~~Before, since the loaded globally scoped settings were not stored in the `scoped-settings-store`, the store did erase them if a user executed `atom.store.set(something, anything, {scopeSelector: '*'})` because it did not have the previous fields.~~

~~This seems to have been introduced by https://github.com/atom/atom/pull/4606 (4 years ago, lol), and my hypothesis is that it may had been just a small mistake when modifying the old logic, but I may be missing something and maybe this was intentional (cc/ @nathansobo if you remember it 😅).~~

### Alternate Designs

🤷‍♂️

### Possible Drawbacks

I'm not aware of any potential drawback apart from the risk of breaking some other logic that was depending on the previous behaviour

### Verification Process

* Manual verification
  1. Open Atom
  2. Execute `atom.config.set("foo.bar", "baz", {scopeSelector: '*'});`
  3. Open `~/.atom/config.cson` and check that my previous config is not gone 😄 
* Added an automated test that fails without this PR change.

### Release Notes

`atom.config.set` does not  erase settings when scopeSelector is `"*"` anymore.

### DIFF
--- a/spec/config-spec.js
+++ b/spec/config-spec.js
@@ -1239,6 +1239,29 @@ describe('Config', () => {
         expect(events[0]).toEqual({ oldValue: 2, newValue: 12 })
       })
     })
+
+    it('keeps all the global scope settings after overriding one', () => {
+      atom.config.resetUserSettings({
+        '*': {
+          foo: {
+            bar: 'baz',
+            int: 99
+          }
+        }
+      })
+
+      atom.config.set('foo.int', 50, { scopeSelector: '*' })
+
+      advanceClock(100)
+
+      expect(savedSettings[0]['*'].foo).toEqual({
+        bar: 'baz',
+        int: 50
+      })
+      expect(atom.config.get('foo.int', { scope: ['*'] })).toEqual(50)
+      expect(atom.config.get('foo.bar', { scope: ['*'] })).toEqual('baz')
+      expect(atom.config.get('foo.int')).toEqual(50)
+    })
   })
 
   describe('.pushAtKeyPath(keyPath, value)', () => {
--- a/src/config.js
+++ b/src/config.js
@@ -691,7 +691,9 @@ class Config {
       this.pendingOperations.push(() => this.set(keyPath, value, options))
     }
 
-    const scopeSelector = options.scopeSelector
+    // We should never use the scoped store to set global settings, since they are kept directly
+    // in the config object.
+    const scopeSelector = options.scopeSelector !== '*' ? options.scopeSelector : undefined
     let source = options.source
     const shouldSave = options.save != null ? options.save : true
 


maxbrunsfeld:
I don't remember how this works super well... but I feel like there was some reason that we didn't put the global settings in the scoped settings store, even though I always wished they were unified whenever I worked on this code. I think the reason may have had to do with mutation. The `ScopedPropertyStore` doesn't have the best support for mutating properties in place, and a lot of the settings that are frequently mutated (like `editor.fontSize`) are global.

Does the bug only happen when `*` is explicitly passed for the `scopeSelector` option? If so, would another solution be to special-case `*` in `Config.set` to behave the same as if no `scopeSelector` was provided?

maxbrunsfeld:
I also might be misunderstanding the new code. With this change, would global settings always be stored both in the `config.settings` object and in the `ScopedPropertyStore`? Or would the scoped property store just contain the global settings at the time the config was loaded, but not reflect any updates to the global settings (done via `atom.config.set('foo', 'bar')`)?

rafeca:
> I don't remember how this works super well... but I feel like there was some reason that we didn't put the global settings in the scoped settings store, even though I always wished they were unified whenever I worked on this code. I think the reason may have had to do with mutation. The ScopedPropertyStore doesn't have the best support for mutating properties in place, and a lot of the settings that are frequently mutated (like editor.fontSize) are global.

Oh thanks for the context! I think the mutability issues may be related to https://github.com/atom/scoped-property-store/pull/8 🤦‍♂️

> Does the bug only happen when * is explicitly passed for the scopeSelector option? If so, would another solution be to special-case * in Config.set to behave the same as if no scopeSelector was provided?

Yeah, I guess I'll have to do this... I tried to avoid adding yet another a special case here but that's going to be the easier way to get this fixed...

> I also might be misunderstanding the new code. With this change, would global settings always be stored both in the config.settings object and in the ScopedPropertyStore? Or would the scoped property store just contain the global settings at the time the config was loaded, but not reflect any updates to the global settings (done via atom.config.set('foo', 'bar'))?

Ups, I did not even realize that global settings are stored in a separate data structure 😐. In this case the global settings may get desynchronised between the two data structures with my change which is bad...

Thanks for the insights! I'm gonna change the solution 😄

maxbrunsfeld:
> I think the mutability issues may be related to atom/scoped-property-store#8 🤦‍♂️

Oh wow, I knew I had taken a look at this problem before; I forgot that it was right when I joined the team, and I ended up giving up on it 😬 .

maxbrunsfeld (approved):

PR #19058, opened 2019-03-26
Fix reopen project when there are no open windows

### LINKED ISSUE #17530: [macOS] File->Reopen Project does not work when there are no open windows
Steps to reproduce:
1. Have Atom on macOS.
2. Open Atom app.
3. Open some project.
4. Close all the windows.
5. In File menu, click "Reopen Project".
6. Click some project in the list.
Expected behaviour: A new window with the selected project should open.
What happens: no visible response.

rafeca:
### Identify the Bug

Fixes https://github.com/atom/atom/issues/17530

### Description of the Change

This PR adds some logic in the atom application process to handle the reopen commands, so when one of these commands gets triggered without an atom environment it still works.

### Alternate Designs

I cannot think about an alternative way of implementing this...

### Possible Drawbacks

Cannot identify any.

### Verification Process

1. Close all Atom windows from OSX
2. Go to `File -> Recent Project` and open one of the projects listed there.
3. Open two folders in a single Atom window and close that window.
4. Go to `File -> Recent Project` and open the recent project containing the two folders.

### Release Notes

Fixed an issue in OSX that prevented File->Reopen Project  to work when there were no open windows

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -900,6 +900,43 @@ describe('AtomApplication', function () {
       )
       atomApplication.promptForPathToOpen.reset()
     })
+
+    it('allows reopening an existing project after all windows are closed', async () => {
+      const tempDirPath = makeTempDir('reopen')
+
+      const atomApplication = buildAtomApplication()
+      sinon.stub(atomApplication, 'promptForPathToOpen')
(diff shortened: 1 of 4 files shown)


jasonrudolph (approved):
:zap:

PR #19060, opened 2019-03-27
Update tree-sitter to v0.14.0

maxbrunsfeld:
This PR bumps the `tree-sitter` node module to the latest version. The only Tree-sitter changes that affect this module are:

* a fix for [one rare bug](https://github.com/tree-sitter/tree-sitter/commit/213ccfd3a47b4e0ce0f52d2db3b1875b4ef37998) caught by the randomized tests
* PRs https://github.com/tree-sitter/tree-sitter/pull/301 and https://github.com/tree-sitter/tree-sitter/pull/308, which make the parsing timeout API more straightforward to use. These changes affected the module's public API slightly, and I've adjusted the Tree-sitter calls to use the new API.

For context, the initial reason behind this ☝️ timeout API is described [here](https://github.com/atom/atom/pull/17339#issuecomment-391527608); I wanted a way to parse synchronously (within a certain "cutoff") when possible, to prevent unnecessary DOM updates. Previously, the cutoff was specified in fairly arbitrary units, whereas now it's specified as a real time duration, in microseconds. I've set the synchronous parsing cutoff at one millisecond for now, which is the amount of time I was trying to approximate before.

### Testing

There haven't been any large code changes, so this should be low risk. But in order to test exhaustively that Tree-sitter is working well in Atom, I plan to do this:

* [x] Make sure the `TreeSitterLanguageMode` unit tests pass
* [x] Get a list of all the grammars that Atom currently bundles

```js
Object.keys(atom.grammars.treeSitterGrammarsById)
```

* [x] View and edit some code in each of these languages:

  * [x] `source.c`
  * [x] `source.cpp`
  * [x] `text.html.basic`
  * [x] `text.html.ejs`
  * [x] `text.html.erb`
  * [x] `source.js`
  * [x] `source.flow`
  * [x] `source.ts`
  * [x] `source.ruby`
  * [x] `source.css`
  * [x] `source.go`
  * [x] `source.jsdoc`
  * [x] `source.js.regexp`
  * [x] `source.python`
  * [x] `source.rust`
  * [x] `source.shell`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6194,42 +6194,12 @@
       "integrity": "sha512-4hjqbObwlh2dLyW4tcz0Ymw0ggoaVDMveUB9w8kFSQScdRLo0gxO9J7WFcUBo+W3C1TLdFIEwNOWebgZZ0RH9Q=="
     },
     "tree-sitter": {
-      "version": "0.13.23",
-      "resolved": "https://registry.npmjs.org/tree-sitter/-/tree-sitter-0.13.23.tgz",
-      "integrity": "sha512-75AiPbMEstv+YK8h4FkAHnmAJ6nNIUj/NFzRvKCHovmwSEKMi8Wc/E/crB4lJnHBOfV/f/DMQjN+e1Y36kagug==",
+      "version": "0.14.0",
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
Ok, this is ready for review.

jasonrudolph (approved):

PR #19075, opened 2019-03-29
 ⬆️ fuzzy-finder@1.12.0

rafeca:
This version includes an option for using the super-fast `fuzzy-native` filtering system ([more info](https://github.com/atom/fuzzy-finder/pull/374)).

----

*List of changes between `fuzzy-finder@1.10.2` and `fuzzy-finder@1.12.0`: https://github.com/atom/fuzzy-finder/compare/v1.10.2...v1.12.0*


rafeca:
All builds are failing 😢 

My current hypothesis is that `electron-link` is not able to manage the cyclic dependency in `fuzzy-native`: https://github.com/atom/fuzzy-native/blob/master/lib/main.js#L4, I'm gonna investigate more

/cc @smashwilson

rafeca:
Oh, the error comes from [here](https://github.com/atom/electron-link/blob/0dc9eb2c4c1527a0cec8a12a439e813ca052015c/src/file-require-transform.js#L131) which throws when processing [this line](https://github.com/atom/fuzzy-native/blob/master/lib/main.js#L16).

I'm gonna try to remove that file from the v8 snapshot...

PR #19106, opened 2019-04-05
:memo: Update README [ci skip]

thomasstep:
### Description of the Change

This change adds a clarification of key mapping for Mac users as well as a warning for a system shortcut that may be unintentionally used.

### Release Notes

Updated documentation to include a warning to Mac users.

-----
[View rendered packages/autoflow/README.md](https://github.com/thomasstep/atom/blob/master/packages/autoflow/README.md)

### DIFF
--- a/packages/autoflow/README.md
+++ b/packages/autoflow/README.md
@@ -1,5 +1,7 @@
 # Autoflow package
 
-Format the current selection to have lines no longer than 80 characters using `cmd-alt-q` on macOS and `ctrl-shift-q` on Windows and Linux. If nothing is selected, the current paragraph will be reflowed.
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks very much for the kind contribution.

On review with the team, we've decided that we're not going to accept this change. We don't believe that duplicating macOS key bindings containing `alt` with ones that contain `opt` is going to make our README's or other documentation easier to follow on the whole. Additionally, there are many key bindings that, when mistyped, could have unforeseen consequences. There are hundreds of key bindings in Atom and placing and maintaining caveats on all the possible ways they could be mistyped is not something we want to take on.

Thanks again for the work you put in on this and your other submissions.

Issue #19089, opened 2019-04-02
Flaky build: Windows builds failing even when all tests are passing


jasonrudolph:
Sometimes Windows build end up failing with the following error even after all the tests succeeded:

```
...
##[error]Cmd.exe exited with code '1'.
##[section]Finishing: Run tests
```

Rebuilding can sometimes resolve the failure.

---

**Affected CI provider(s)**: Azure DevOps

**Affected platform(s)**: Windows x86, Windows x64

**Example commit experiencing flakiness**: https://github.com/atom/atom/commit/aca476c48ac241092e2c5970e52eadd37328f30d

**Example failing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=36073&view=logs&jobId=97a617bf-bcbd-5dfa-bba2-cfba2747b693&taskId=1e1369e9-1c56-59b0-830c-f1606a680c8a&lineStart=14&lineEnd=30&colStart=1&colEnd=1

**Example passing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=36136

**Example log from failing build**: https://github.com/atom/atom/files/3034360/build.log

jasonrudolph:
A few more occurrences:

- https://github.visualstudio.com/Atom/_build/results?buildId=35799&view=logs
- https://github.visualstudio.com/Atom/_build/results?buildId=36208&view=logs
- https://github.visualstudio.com/Atom/_build/results?buildId=36846&view=logs

jasonrudolph:
Another one:

- https://github.visualstudio.com/Atom/_build/results?buildId=33073

jasonrudolph:
Another one:

- https://ci.appveyor.com/project/Atom/atom/builds/23463872/job/7pg68dkgeo6ehfb2?fullLog=true#L70

    ```
    Executing core main process tests
      . AtomApplication waits until all the windows have saved their state before quitting: 14168ms
        AtomApplication prevents quitting if user cancels when prompted to save an item: Object has been destroyed
    Error: Object has been destroyed
        at WebContents.send (C:\projects\atom\out\Atom Dev x64\resources\electron.asar\browser\api\web-contents.js:101:15)
        at EventEmitter.exports.on (C:\projects\atom\src\ipc-helpers.js:41:18)
        at <anonymous>:null:null
      . AtomApplication prevents quitting if user cancels when prompted to save an item: 7995ms
    ...
    ```

jasonrudolph:
Another one:

[...]

PR #19124, opened 2019-04-09
Fix atom-application tests

### LINKED ISSUE #18993: Flaky test: AtomApplication launch when closing the last window quits the application
In our Windows x86 builds, we sometimes see the build fail at [this point in the test suite](https://github.visualstudio.com/Atom/_build/results?buildId=33073&view=logs&jobId=97a617bf-bcbd-5dfa-bba2-cfba2747b693&taskId=1e1369e9-1c56-59b0-830c-f1606a680c8a&lineStart=64&lineEnd=67&colStart=1&colEnd=32)...

```
  ...
  . AtomApplication launch when the `--wait` flag is passed kills the specified pid after a newly-opened directory in an existing window is closed: 8005ms
    AtomApplication launch when closing the last window quits the application:
  . AtomApplication launch when closing the last window quits the application: 7243ms
##[error]Cmd.exe exited with code '1'.
##[section]Finishing: Run tests
```

---

**Affected platform(s)**: Windows x86

**Example commit experiencing flakiness**: https://github.com/atom/atom/commit/24ed380e786fa23af761a51e5afded58fe3ca1b5

**Example failing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=33073

**Example passing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=33116

**Example log from failing build**: https://github.com/atom/atom/files/2967685/build.log

rafeca:
When closing a window with a file that does not exist, Atom opens a dialog asking the users if they want to save the changes. This dialog prevents the tests from finishing correctly (as seen in the [CI builds](https://github.visualstudio.com/Atom/_build/results?buildId=37183&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&taskId=8b9bff53-2841-58c7-68fc-e30c30330fc3&lineStart=197&lineEnd=204&colStart=1&colEnd=3)).

I still don't know why this test can ever pass (currently it always fails on my OSX machine).

This should fix https://github.com/atom/atom/issues/18993

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -663,6 +663,10 @@ describe('AtomApplication', function () {
             parseCommandLine([path.join(makeTempDir('a'), 'file-a')])
           )
           await focusWindow(window)
+
+          // Choosing "Don't save"
+          mockElectronShowMessageBox({ response: 2 })
+
           window.close()
           await window.closedPromise
           await atomApplication.lastBeforeQuitPromise
@@ -675,6 +679,10 @@ describe('AtomApplication', function () {
             parseCommandLine([path.join(makeTempDir('a'), 'file-a')])
           )
           await focusWindow(window)
+
+          // Choosing "Don't save"
+          mockElectronShowMessageBox({ response: 2 })
+
           window.close()
           await window.closedPromise
           await timeoutPromise(1000)


rafeca:
It seems that this test passes whenever the following trace is shown on the logs:

```
Object has been destroyed
Error: Object has been destroyed
    at WebContents.send (/Users/rafeca/github/atom/atom/out/Atom Dev.app/Contents/Resources/electron.asar/browser/api/web-contents.js:101:15)
    at EventEmitter.exports.on (/Users/rafeca/github/atom/atom/src/ipc-helpers.js:41:18)
    at <anonymous>:null:null
```

So probably a different error is causing Atom to crash and that makes this test pass 🤦‍♂️

This is related to https://github.com/atom/atom/issues/19089

rafeca:
After further investigation it looks like the error mentioned above is unrelated with this flakiness: the problem is that the test does not wait until Atom has been fully loaded so if Atom loads faster than expected the confirm dialog will be shown and the test will fail.

rafeca:
Gonna land this one and continue investigating the root issue, which is maybe causing other failures

PR #19135, opened 2019-04-10
Add ability to retry flaky Jasmine tests

nathansobo:
Our builds are currently plagued by a handful of timing-dependent tests that fail intermittently. The ideal solution would be to find and fix the source of indeterminacy in these tests, as I did in https://github.com/atom/settings-view/pull/1123, but that can be time consuming. An even *better* solution would be to restructure these code paths and tests to not depend on timing in a critical way, but that would be even *more* difficult. Another alternative is just to delete the offending tests, but I'm reluctant to lose the coverage on functionality that they are trying to document.

So in the interests of reducing spurious build failures *quickly* without resorting to deleting tests, I'm introducing the ability to retry flaky tests as an experiment. Teammates have expressed concerns about the long-term impact of doing this and I myself am not totally sure it's a good idea, but I think it's worth trying. This feature essentially transforms flaky tests that cause massive delays in our build process into technical debt. We still to prioritize cleaning them up, but we can mitigate their impact on our build process in the short term. If we're still seeing a test be unreliable after retrying it, at that point I think we 
should either delete it or make fixing it a top priority in that sprint.

So if you encounter a flaky test and you've spent some time trying to figure it out and just can't, you now have a new option. You can add the following to the top of the test body:

```js
 it('fails 5% of the time on Windows x86 for unknown reasons and ruins our lives', async function () {
    this.RETRY_FLAKY_TEST_AND_SLOW_DOWN_THE_BUILD()

    // ... insert remainder of poorly-designed test case here ;-)
})
```

I've deliberately made using this option ugly and screamingly obvious to discourage its use. Sorry in advance if this ends up being a bad idea.

### DIFF
--- a/spec/native-watcher-registry-spec.js
+++ b/spec/native-watcher-registry-spec.js
@@ -100,6 +100,8 @@ describe('NativeWatcherRegistry', function () {
   })
 
   it('reuses an existing NativeWatcher on the same directory', async function () {
+    this.RETRY_FLAKY_TEST_AND_SLOW_DOWN_THE_BUILD()
+
     const EXISTING = new MockNative('existing')
     const existingPath = absolute('existing', 'path')
     let firstTime = true
--- a/vendor/jasmine.js
+++ b/vendor/jasmine.js
@@ -2021,6 +2021,17 @@ jasmine.Queue = function(env) {
   this.abort = false;
 };
 
+jasmine.Queue.prototype.clone = function() {
+  var queue = new jasmine.Queue(this.env);
+  queue.ensured = this.ensured.slice(0)
+  queue.blocks = this.blocks.slice(0)
+  queue.running = this.running;
+  queue.index = this.index;
+  queue.offset = this.offset;
+  queue.abort = this.abort;
+  return queue;
+}
+
 jasmine.Queue.prototype.addBefore = function(block, ensure) {
   if (ensure === jasmine.undefined) {
     ensure = false;
@@ -2269,7 +2280,11 @@ jasmine.Spec.prototype.addToQueue = function (block) {
  * @param {jasmine.ExpectationResult} result
  */
 jasmine.Spec.prototype.addMatcherResult = function(result) {
-  this.results_.addResult(result);
+  if (this.retries > 0 && !result.passed_) {
+    this.retry();
+  } else {
+    this.results_.addResult(result);
+  }
 };
 
 jasmine.Spec.prototype.expect = function(actual) {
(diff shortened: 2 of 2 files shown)


50Wliu:
Will this work for tests that are using arrow functions? (e.g. in the test you changed, if it would still work with `async () =>`)

nathansobo:
@50Wliu I know it works for `async` test functions. I didn't test with `=>` syntax, but I don't see any reason why it shouldn't work. It essentially just restores the spec's queue and re-enqueues it, so anything that works already in Jasmine *should* work unless I'm missing something.

nathansobo:
One thing I forgot to note. Don't use this in package specs until 1.38.0 hits stable, because it won't be available in CI. Or at the very least feature detect it.

rafeca:
> Will this work for tests that are using arrow functions? (e.g. in the test you changed, if it would still work with async () =>)

I don't think that this is going to work for tests defined with arrow functions, since the arrow function will bind the context of the spec to be the context of the `describe` callback, which based on [Jasmine's code](https://github.com/atom/atom/blob/master/vendor/jasmine.js#L825) is going to be a `jasmine.Suite` instance.

Additionally, any spec that uses the `async-spec-helpers.js` helper methods that we have around the many Atom repos is not going to be able to use the retry logic, since the overriden `it` functions defined there pass an undefined context to the spec function ([source](https://github.com/atom/fuzzy-finder/blob/f096a3e1faea35573d1619e410496031e9cccce4/spec/async-spec-helpers.js#L27)).

Good thing is that after https://github.com/atom/atom/pull/18917 there's no need to use the `it` implementations from `async-spec-helpers.js` anymore (in fact [I removed all of them](https://github.com/atom/atom/pull/18932) from Atom core), and the new global `it` functions introduced in https://github.com/atom/atom/pull/18917 handle the context correctly.

nathansobo:
Of course. Feeling pretty dumb. Arrows bond the context and I depend on `this`. I did that so you could easily combine retires with `fit` etc. The async helpers situation is more concerning. Perhaps I should update them.

Arcanemagus:
> The async helpers situation is more concerning. But they are gone now?

After #18917 they are no longer needed as the functionality they included is now the default behavior of the built in Jasmine runner, however as mentioned _only_ the bundled packages that have already been migrated into `atom/atom` have been updated, the rest would still potentially be using an existing separate implementation.

nathansobo:
@Arcanemagus It seems like if we wanted to use this feature in such a package, we could migrate the tests to use the new global versions of the Jasmine functions at that time.

I just encountered another instance of a [flaky pathwatcher test](https://github.visualstudio.com/Atom/_build/results?buildId=38196) that this PR would have avoided, so I'm going to merge it. We should use this tool sparingly and reevaluate going forward.

PR #19136, opened 2019-04-10
"Open file" in existing window

### LINKED ISSUE #19133: "Open file..." opens a new window
### Prerequisites

### Description

On Windows, when opening a file using the "Open file..." item from the app menu, the file gets opened in a new Atom window instead of reusing the existing window.

### Steps to Reproduce

1. Open Atom.
2. Go to "File" -> "Open file..." and open a random file.

**Expected behavior:** The file should get opened in the same Atom window.

**Actual behavior:** a new Atom window gets opened.

### Versions

This started failing on the 1.36 branch.

### Additional Information

N/A

smashwilson:

### Identify the Bug

Fixes #19133.

### Description of the Change

In AtomApplication:

* Made the `getLoadSettings()` helper optionally include the last focused window in its output.
* Modified `openPaths()` to more strongly respect an explicitly provided `{window}` argument. (Previously, `{window}` would be used only if `{addToLastWindow}` was set to true.)
* Changed `promptForPathsToOpen()` to pass a `{window}` argument on to `openPaths()` if and only if all of the chosen paths are non-directories. (This is to be consistent with the behavior in 1.35.1; "open folder" opens a new window, "open file" opens in the existing window.)
* Used `getLoadSettings(true)` in all of the codepaths that an "open" or "open file" dialog can hit.

### Alternate Designs

We could refactor the opening logic completely I guess to be clearer and less error-prone. Now isn't the time to do that, though, and it went poorly the last time I tried it.

### Possible Drawbacks

`application:open-file` and `application:open-folder` have inconsistent and confusing behavior.

### Verification Process

I did a manual Atom build and verified that the File -> Open... menu item and the `cmd-O` keybinding opened files in the current window and folders in a new window.

### Release Notes

This is fixing a regression, so no release notes.

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -968,11 +968,21 @@ class AtomApplication extends EventEmitter {
     const normalizedPathsToOpen = locationsToOpen.map(location => location.pathToOpen).filter(Boolean)
 
     let existingWindow
-    if (!newWindow && normalizedPathsToOpen.length > 0) {
+
+    // Explicitly provided AtomWindow has precedence unless a new window is forced.
+    if (!newWindow) {
+      existingWindow = window
+    }
+
+    // If no window is specified, a new window is not forced, and at least one path is provided, locate
+    // an existing window that contains all paths.
+    if (!existingWindow && !newWindow && normalizedPathsToOpen.length > 0) {
       existingWindow = this.windowForPaths(normalizedPathsToOpen, devMode)
     }
 
-    if (addToLastWindow && !existingWindow) {
+    // No window specified, new window not forced, no existing window found, and addition to the last window
+    // requested. Find the last focused window.
+    if (!existingWindow && !newWindow && addToLastWindow) {
       let lastWindow = window || this.getLastFocusedWindow()
       if (lastWindow && lastWindow.devMode === devMode) {
         existingWindow = lastWindow
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -452,13 +452,17 @@ class AtomApplication extends EventEmitter {
 
   // Registers basic application commands, non-idempotent.
   handleEvents () {
-    const getLoadSettings = () => {
+    const getLoadSettings = includeWindow => {
       const window = this.focusedWindow()
-      return {devMode: window && window.devMode, safeMode: window && window.safeMode}
+      return {
+        devMode: window && window.devMode,
+        safeMode: window && window.safeMode,
+        window: includeWindow && window
+      }
     }
 
     this.on('application:quit', () => app.quit())
-    this.on('application:new-window', () => this.openPath(getLoadSettings()))
+    this.on('application:new-window', () => this.openPath(getLoadSettings(false)))
     this.on('application:new-file', () => (this.focusedWindow() || this).openPath())
     this.on('application:open-dev', () => this.promptForPathToOpen('all', {devMode: true}))
     this.on('application:open-safe', () => this.promptForPathToOpen('all', {safeMode: true}))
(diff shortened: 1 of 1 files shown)


jasonrudolph:
> ### Possible Drawbacks
> `application:open-file` and `application:open-folder` have inconsistent and confusing behavior.

@smashwilson: Am I correct in thinking that this inconsistent/confusing behavior predates this pull request (i.e., that this has been Atom's behavior for a while now)?

smashwilson:
@jasonrudolph: Yup - basically, I downloaded Atom 1.35.1 and did my best to match its behavior.

Oh, also, our test coverage for AtomApplication is not great. I'd take the time to add tests for this, but ugh.

src/main-process/atom-application.js
```diff
@@ -1412,14 +1426,32 @@ class AtomApplication extends EventEmitter {
...
+        // new window instead.
+        if (type === 'folder') {
+          targetWindow = null
+        } else if (type === 'file') {
+          targetWindow = window
+        } else if (type === 'all') {
+          const areDirectories = await Promise.all(
+            pathsToOpen.map(pathToOpen => new Promise(resolve => fs.isDirectory(pathToOpen, resolve)))
+          )
+          if (!areDirectories.some(Boolean)) {
```
jasonrudolph: Current status: Wishing JavaScript Array provided a [`none` method](https://ruby-doc.org/core-2.6.2/Enumerable.html#method-i-none-3F).

src/main-process/atom-application.js
```diff
@@ -452,13 +452,17 @@ class AtomApplication extends EventEmitter {
...
+      return {
+        devMode: window && window.devMode,
+        safeMode: window && window.safeMode,
+        window: includeWindow && window
+      }
     }
 
     this.on('application:quit', () => app.quit())
-    this.on('application:new-window', () => this.openPath(getLoadSettings()))
+    this.on('application:new-window', () => this.openPath(getLoadSettings(false)))
```
jasonrudolph: This isn't something that should hold up this pull request, but I think calls to `getLoadSettings` could be more intention-revealing/self-documenting we used a named argument. 

```js
getLoadSettings({includeWindow: false})
```

What do you think?

jasonrudolph (commented):
@smashwilson: Thanks for jumping on this. :bow:💟

Do you anticipate this pull request resulting in any changes on other platforms (i.e., Linux and macOS)? Would it make sense try out a build on those other platforms to ensure that it's behaving as expected?

src/main-process/atom-application.js
smashwilson: Yeah... I spent a little time trying to juggle this in a way that was more readable, but I didn't have much luck. Notably I couldn't avoid the negative by looking for files instead because paths that don't exist yet are also not "files". And negating it to produce `notDirectories` just made it worse 🙃

src/main-process/atom-application.js
smashwilson: Yah, that's reasonable. Honestly this whole class could use an overhaul to its readability. I tried to at least leave comments where I was changing things.

smashwilson (commented):
Thanks @jasonrudolph 🙇

smashwilson:
> Do you anticipate this pull request resulting in any changes on other platforms (i.e., Linux and macOS)? Would it make sense try out a build on those other platforms to ensure that it's behaving as expected?

This will actually have the same behavior on all three, although the menu items present in the "file" menu differ from platform to platform. I tested builds on macOS by running commands from the command palette. But I'll be sure to give it a final check on the others.

trisa88:
Help mi

Issue #18318, opened 2018-10-24
[Atom 1.32.0] Does not inherit environment variables when launched not from the command line


jzazo:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When I launch Atom in Ubuntu 18.04 it does not inherit environmental variables. Previous versions did inherit them. My suspicion is that since the startup script has been reduced, a regression might have been introduced.

If I launch Atom from command line, it does inherit variables correctly.

*Edit by @rsese to add explicit repro steps based on https://github.com/atom/atom/issues/18318#issuecomment-432821319, expected/actual behavior, and how often it reproduces*

### Steps to Reproduce 

1. Edit and save `~/.bashrc` to change `$PATH`, added this at the end:

```bash
export PATH="$PATH:my/new/path"
```

2. Start Atom from Activities or your Favorites
3. Open devtools <kbd>ctrl-shift-i</kbd>
4. In Console, run `process.env["PATH"]`

**Expected behavior:** [What you expect to happen]

Change in step 1 to be available.

**Actual behavior:** [What actually happens]

Change in step 1 is not available.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100% with repro steps and if you start Atom from Activities or Favorites, doesn't reproduce if you start Atom from the command line.  

Doesn't reproduce on Atom 1.31.1.

### Versions

```
Atom    : 1.32.0
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

*Edit by @rsese to mention the OS confirmed on*

Ubuntu 18.04.1 (Also noticed that this doesn't reproduce on macOS if I start Atom from spotlight or the dock.)

rsese:
Thanks for the report!

[...]

PR #17508, opened 2018-06-13
.Desktop SCL compat by setting exec path

FossPrime:
### Description of the Change

Based on this zombie PR:
https://github.com/atom/atom/pull/15887

Meant to address #13451 which produces about 4 SCL warnings on plugins that use terminal utilities on RHEL based distributions. Including the official github package.

### Why Should This Be In Core?

It's a frustrating bug to anyone who uses any terminal, git and other Atom plugins in Fedora.

### Benefits

Clean running Atom plugins on RHEL

### Possible Drawbacks

None.

### Verification Process

Started Atom with the change and the SCL warnings were gone.

Testing procedure:
1. open a project with git in atom
2. set the origin url to a bad origin
3. `git remote set-url origin git@gitlab.com:tdps/non-existant-repo`
4. Behold the strange output

Pre patch:
screen shot 2018-06-13 at 9 24 32 am

Post patch:
screen shot 2018-06-13 at 9 28 11 am

### Applicable Issues

#13451
https://github.com/platformio/platformio-atom-ide-terminal/issues/120#issuecomment-274756065

PR #19138, opened 2019-04-11
Calculate env vars when launching Atom from the desktop

rafeca:
### Identify the Bug

Atom [has some logic](https://github.com/atom/atom/blob/master/src/update-process-env.js#L61-L95) to get the correct env variables when calling directly the Atom binary (the one located in `/usr/share/atom/atom`).

This logic is bypassed when calling Atom via the bash CLI (the script located in `/usr/bin/atom`). This is done [by checking](https://github.com/atom/atom/blob/master/src/update-process-env.js#L54) the `ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT` env var which [is set](https://github.com/atom/atom/blob/master/atom.sh#L27) by the CLI (there's more info in [the PR](https://github.com/atom/atom/pull/12562) that introduced this).

This logic assumes that entry points that don't contain the correct env vars (like a desktop launcher) will call the binary directly, while a shell script or terminal (which contain the correct env vars) will call the CLI. After #17508, this is no longer true, and this is the root cause of the issue.

Applicable issue: https://github.com/atom/atom/issues/18318

### Description of the Change

This PR makes Atom keep using the previous logic to calculate env vars instead. To do so, it modifies the linux launcher to define the `ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT` env var and changes the bash script to not define it if it's already set.

### Alternate Designs

* Create a different bash script which is only used by desktop launchers. This will require more changes and it does not provide any major benefit.
* Change the bash script to use bash in interactive mode. This can cause unexpected behaviours in some configurations (depending on the users' `.bashrc` files).

### Possible Drawbacks

None that I can think of.

### Verification Process

- [x] Patched a working copy of Atom in a linux machine and checked that it works correctly.
- [x] Download an Atom build from this PR and verify it.

### Release Notes

Fixed issue inheriting environmental variables when opening Atom from the desktop on Linux.

### DIFF
--- a/atom.sh
+++ b/atom.sh
@@ -24,7 +24,11 @@ case $(basename $0) in
     ;;
 esac
 
-export ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT=true
+# Only set the ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT env var if it hasn't been set.
+if [ -z "$ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT" ]
+then
+  export ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT=true
+fi
 
 ATOM_ADD=false
 ATOM_NEW_WINDOW=false
--- a/resources/linux/atom.desktop.in
+++ b/resources/linux/atom.desktop.in
@@ -2,7 +2,7 @@
 Name=<%= appName %>
 Comment=<%= description %>
 GenericName=Text Editor
-Exec=<%= installDir %>/bin/<%= appFileName %> %F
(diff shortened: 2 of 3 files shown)


rafeca:
>  Download an Atom build from this PR and verify it.

Uhm, it seems that we don't create artifcats on builds for PRs, so if nobody opposes, I'll do the last bit of the verification process after this PR gets merged into `master`.

rafeca:
CI build passed, I'm gonna merge this one for now and check that the next Linux build on `master` works fine.

/cc @atom/core

rafeca:
I've verified that the functionality works correctly on Ubuntu by downloading the `.dev` binary generated by the azure pipelines on commit https://github.com/atom/atom/commit/3f2a702baaa353d24464e6b61fbc751d049b1913

Issue #19127, opened 2019-04-09
How can I change tree-view default width?


bl00mber:
## How can I change default width of the tree-view?

I've tried to modify contents of the package tree-view and styles.less to modify the default width of tree-view `atom-dock-mask` container, but it doesn't appear to have any effect and I can't find the place in the code of module that defines this value. For example, package `tree-view` not even contains `300` (default value of the width) whatsoever.

rsese:
Thanks for reaching out!

Because we treat our issues list as [the Atom team's backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), we close issues that are questions since [they don't represent a task needing to be completed](http://blog.atom.io/2016/04/19/managing-the-deluge-of-atom-issues.html). For most questions about Atom there are a lot of options, check out:

* User Documentation &mdash; [The Atom Flight Manual](http://flight-manual.atom.io)
* Developer Documentation &mdash; [Atom API Documentation](https://atom.io/docs/api/latest)
* FAQ &mdash; [The Atom FAQ on Discuss](https://discuss.atom.io/c/faq)
* Message Board &mdash; [Discuss, the official Atom and Electron message board](https://discuss.atom.io)
* Chat &mdash; [Join the Atom Slack team](https://discuss.atom.io/t/join-us-on-slack/16638)

On Discuss and in the Atom Slack team, there are a bunch of helpful community members that should be willing to point you in the right direction.

---

I poked at this a tiny bit and I'm not positive but I think the default width of 300 comes from the default width of a dock?  

https://github.com/atom/atom/blob/2f827bfc23260eece024b8f179421fa833b1834d/src/dock.js#L10

In any case, as mentioned above the best place to ask questions like this is the Atom Slack or the Message Board.

PR #19159, opened 2019-04-13
Add option to customize dock `DEFAULT_INITIAL_SIZE`

bl00mber:
https://github.com/atom/atom/issues/19127
https://github.com/atom/tree-view/issues/250
https://github.com/atom/tree-view/issues/520

### Description of the Change

Add option to customize dock `DEFAULT_INITIAL_SIZE` using Editor Settings to keep dock size when opening new window

### Verification Process
* Enable tree-view
* Execute atom.config.set('editor.dockSize', 400);
* Open new Atom window
* Check changed tree-view dock initial size

### Release Notes

Added option allowing to customize initial dock size.

### DIFF
--- a/spec/dock-spec.js
+++ b/spec/dock-spec.js
@@ -26,6 +26,46 @@ describe('Dock', () => {
       expect(document.activeElement).toBe(dock.getActivePane().getElement())
       expect(didChangeVisibleSpy).toHaveBeenCalledWith(true)
     })
+
+    it('opens the vertically-oriented dock with the size of editor.dockSize config value', async () => {
+      atom.config.set('editor.dockSize', 400)
+
+      jasmine.attachToDOM(atom.workspace.getElement())
+
+      const item = {
+        element: document.createElement('div'),
+        getDefaultLocation () {
+          return 'left'
+        }
+      }
+
+      await atom.workspace.open(item)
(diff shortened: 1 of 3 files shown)


lee-dohm:
Thanks for the suggestion!

Unfortunately, it appears that the API here may be a little lacking. This appears to configure all three docks (left, right, and bottom) with the same value. People may want different values for different docks or even have different defaults based on what appears in the dock.

Rather than create some huge configuration scheme that would probably still not cover every eventuality, we would prefer that people use packages to overcome this. This would give them the control to specify exactly which dock and under which circumstances they would like the dock size to be different. If, however, you find that packages are unable to change the dock size, then we would definitely consider a well-written pull request to add a way for packages to control the dock size.

Thanks for your hard work and interest in making Atom better!

bl00mber:
@lee-dohm is there a way to configure `DEFAULT_INITIAL_SIZE` on a package level for package `tree-view`?
Just asking for the possibility, not implementation.

ThatXliner:
> @lee-dohm is there a way to configure `DEFAULT_INITIAL_SIZE` on a package level for `tree-view`?
> 
> Just asking for the possibility, not implementation.

I would definitely want that. I want tree view to shrink as much as possible but not so much.

I think every frame should have a size API

PR #19167, opened 2019-04-15
go-to-line: Support for jumping to relative line

cabageman2:
Issue or RFC Endorsed by Atom's Maintainers
Issue #18246 [feature request] Support for jumping to relative line
Description of the Change
Modifies Go to Line feature to allow input of “+20” or “-20,” which moves the cursor up or down the respective number of lines. Does not affect the existing functionality of Go To (input such as “5”, “4:3”), although input is now limited to only valid strings (does not allow “2::3::::5:4” as it used to). 
Alternate Designs
Considered continuing to check whether each individual character was valid, as opposed to checking the string as a whole, however this allowed nonsensical input (Go to line/column “2::3::::5:4”). This would also make it more complicated when “+” and “-” were involved as well (didn’t want to allow input such as “:5+23--::”).  
Possible Drawbacks
If a user tries to input a string that isn’t valid, the editor simply won’t let the user type invalid characters. This may be a bit confusing because there is no warning message, but should be fairly easy to figure out the cause.
Verification Process
Added test cases to the spec file to demonstrate that the existing go-to-line functionality still works (5:3), as well as the added functionality (+5, -4). Also test cases that ensure that letters and invalid syntax are not able to be typed to avoid bad input.
Release Notes
Allows you to jump to a line in your project relative to your current line, rather than specifying the exact line number.


50Wliu:
Thanks for the submission! Please fix the tests and we'll take another look.

rsese:
Thanks again @cabageman2 :bow:

In addition to @50Wliu's comment, we were also wondering if you can talk more about any edge cases you've considered?  For example, how does this work if you have multiple cursors?

Also, the other maintainers were hoping for stronger tests around the functionality you're adding?  Is that something you can expand on?

script/lib/install-application.js
```diff
@@ -1,157 +0,0 @@
-'use strict'
```
50Wliu: What happened here?

cabageman2:
Sorry, install-application.js was deleted, so we added that back in. Good catch on the functionality tests, we added some tests in spec.js. Also we never knew that the multiple cursor package existed, currently if you use go-to-line when your cursor size is greater than 1, if you go to a higher row it'll change position relative to the cursor's upper limit, if you go to a lower row it'll change position relative to the cursor's lower limit. Also in both cases it'll change the cursor size back to 1.

rafeca:
Hey! 👋

With ❤️, the Atom team.

mortenkjaergaard:
Hi @cabageman2 - just writing to ask if you had any success following up on this? Seems your changes were not approved (I see 'all checks have failed' on the merge request 😥 . I'm super interested in this feature, should you have a chance to address any lingering issues!

sadick254:
Thank you for your issue!

We haven't gotten a response to our questions in our comment above. I'm going to close this but don't hesitate to reach out if you have resolved the failing tests, we'll be happy to reopen the PR.

Issue #18991, opened 2019-03-14
Flaky build: Windows build sometimes fails with "resource busy or locked"


jasonrudolph:
In our Windows 64-bit builds, we frequently see the build fail at this point in the test suite...

```
...
TextEditor > buffer manipulation > .outdentSelectedRows() > when one line is selected > outdents line and retains editor Renderer process crashed, exiting
EBUSY: resource busy or locked, unlink 'C:\Users\VSSADM~1\AppData\Local\Temp\atom-test-data119118-3988-dplic9.gszd4\databases\Databases.db'
Error: EBUSY: resource busy or locked, unlink 'C:\Users\VSSADM~1\AppData\Local\Temp\atom-test-data119118-3988-dplic9.gszd4\databases\Databases.db'
    at Object.fs.unlinkSync (fs.js:1061:3)
    at rimrafSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:211:13)
    at D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:245:5
    at Array.forEach (<anonymous>:null:null)
    at rmkidsSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:244:26)
    at rmdirSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:237:7)
    at fixWinEPERMSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:150:5)
    at rimrafSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:216:26)
    at D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:245:5
    at Array.forEach (<anonymous>:null:null)
    at rmkidsSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:244:26)
    at rmdirSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:237:7)
    at fixWinEPERMSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:150:5)
    at rimrafSync (D:\a\1\s\node_modules\temp\node_modules\rimraf\rimraf.js:216:26)
    at cleanupDirsSync (D:\a\1\s\node_modules\temp\lib\temp.js:143:5)
    at process.cleanupSync (D:\a\1\s\node_modules\temp\lib\temp.js:196:19)
[...]

PR #19171, opened 2019-04-16
Upload Windows crash reports on Azure build pipelines for pull requests only

nathansobo:
To combat hard crashes like https://github.com/atom/atom/issues/18991, we need access to crash the dumps produced by Electron for a given build on Windows. Crash reports leak environment variables, so we can't upload them as public artifacts on release branches. This PR will enable uploading on non-release branches, and I will follow up with a second PR to upload them to a private location for release branches.

Tasks:
* [x] Initial implementation for Azure build pipelines
* [x] Ensure passing PR builds still succeed when there are no crash reports to publish
* [x] Ensure failing PR builds with no crashes don't produce errors for build steps associated with crash report upload.
* [x] Ensure PR builds can publish crash dumps

### DIFF
--- a/script/vsts/platforms/windows.yml
+++ b/script/vsts/platforms/windows.yml
@@ -99,6 +99,23 @@ jobs:
       testRunTitle: Windows $(buildArch)
     condition: ne(variables['Atom.SkipTests'], 'true')
 
+  - script: |
+      IF NOT EXIST "%ARTIFACT_STAGING_DIR%\crash-reports" MKDIR "%ARTIFACT_STAGING_DIR%\crash-reports"
+      IF EXIST "%Temp%\Atom Crashes" (
+        FOR %%a in ("%Temp%\Atom Crashes\*.dmp") DO XCOPY "%%a" "%ARTIFACT_STAGING_DIR%\crash-reports" /I
+      )
+    displayName: Stage crash reports
+    condition: failed()
+    env:
+      ARTIFACT_STAGING_DIR: $(Build.ArtifactStagingDirectory)
+
+  - task: PublishBuildArtifacts@1
+    inputs:
+      PathtoPublish: $(Build.ArtifactStagingDirectory)/crash-reports
+      ArtifactName: crash-reports
+    displayName: Publish crash reports on non-release branch
+    condition: and(failed(), eq(variables['IsReleaseBranch'], 'false'))
+
   - task: PublishBuildArtifacts@1
     inputs:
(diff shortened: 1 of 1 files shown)


nathansobo:
I'm going to focus just on PR builds for now and focus on release branch builds in a subsequent PR. Updated the body accordingly.

nathansobo:
I'm going to merge this so we can start getting crash reports on PRs ASAP. A review would still be appreciated though in case I should follow up with any adjustments.

script/vsts/platforms/windows.yml
```diff
@@ -99,6 +99,23 @@ jobs:
       testRunTitle: Windows $(buildArch)
     condition: ne(variables['Atom.SkipTests'], 'true')
 
+  - script: |
+      IF NOT EXIST "%ARTIFACT_STAGING_DIR%\crash-reports" MKDIR "%ARTIFACT_STAGING_DIR%\crash-reports"
+      IF EXIST "%Temp%\Atom Crashes" (
+        FOR %%a in ("%Temp%\Atom Crashes\*.dmp") DO XCOPY "%%a" "%ARTIFACT_STAGING_DIR%\crash-reports" /I
+      )
```
smashwilson: For future reference, you're not tied to batch here; you can use [`powershell`](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops) to write in PowerShell instead:

```yaml
- powershell: |
    if (Path-Exists "${env:ARTIFACT_STAGING_DIR}\crash-reports") {
      Create-Item -Type Directory "${env:ARTIFACT_STAGING_DIR}\crash-reports"
    }
    # etc
```

script/vsts/platforms/windows.yml
```diff
@@ -99,6 +99,23 @@ jobs:
...
+    condition: failed()
+    env:
+      ARTIFACT_STAGING_DIR: $(Build.ArtifactStagingDirectory)
+
+  - task: PublishBuildArtifacts@1
+    inputs:
+      PathtoPublish: $(Build.ArtifactStagingDirectory)/crash-reports
+      ArtifactName: crash-reports
+    displayName: Publish crash reports on non-release branch
+    condition: and(failed(), eq(variables['IsReleaseBranch'], 'false'))
```
smashwilson: 🤔 What's left to do for release branch support other than loosening this condition? Something to do with wrangling the secrets appropriately?

nathansobo: Yeah, as you probably saw on the other PR, I uploaded to a private S3 object in the release branch case.

PR #19172, opened 2019-04-16
Try upgrading Tree-sitter for a third time

maxbrunsfeld:
This time, I've configured Travis so that it will run on this branch, so we will be able to tell if the branch breaks the Travis build before merging it.

I still have no idea why this change would affect the main process tests. AFAICT, Tree-sitter is not loaded in any main process test.

See #19060

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7185,42 +7185,12 @@
       "integrity": "sha512-4hjqbObwlh2dLyW4tcz0Ymw0ggoaVDMveUB9w8kFSQScdRLo0gxO9J7WFcUBo+W3C1TLdFIEwNOWebgZZ0RH9Q=="
     },
     "tree-sitter": {
-      "version": "0.13.23",
-      "resolved": "https://registry.npmjs.org/tree-sitter/-/tree-sitter-0.13.23.tgz",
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
The windows failure on Azure looks spurious:

```
Project
  .onDidChangeFiles()
    it Project .onDidChangeFiles() reports filesystem changes within project paths
      timeout: timed out after 60000 msec waiting for promise to be resolved or rejected
```

maxbrunsfeld:
Ok, finally 💚 .

I still don't get why the Travis build was broken by https://github.com/atom/atom/pull/19060, but I now know that it had something to do with the fact that we [started providing pre-built binaries](https://github.com/tree-sitter/node-tree-sitter/pull/19) of the Tree-sitter module.

As far as I can tell, something about those prebuilt binaries causes the Travis build to time out on those two integration tests, even though the Azure build has always been fine.

It may have to do with the fact that the Travis build [is configured to use clang](https://github.com/atom/atom/blob/1de17649eb3914f2610773796ecc26654da14026/.travis.yml#L19) for C and C++ code, whereas the pre-built tree-sitter binaries were built with gcc. Maybe that caused them to link against different versions of `libc` or `libc++`/`libstdc++`.

For now, I've deleted the Linux Electron artifacts from [the `tree-sitter release`](https://github.com/tree-sitter/node-tree-sitter/releases/tag/v0.15.0), and it has fixed the issue. I might just stop creating linux binaries for now.

Issue #9873, opened 2015-11-30
Default font size reset / deleted after command window:reset-font-size


DerToti:
Hello everyone im not really sure if this is a bug or not.

In my config.cson i have fontSize setup to 11 and on the settings section as well on 11.

The problem now is when im zooming in in the editor, and press cmd-0 to call the reset-font-size command i would suggest to have my font size reseted to 11 again (my default font size) but instead it is reseted to 14 i think and on the config.cson my fontSize 11 is gone as well on the settings section.

T think this was working before (don't know when it broke).

jirsbek:
Exactly same problem here. Sometimes I accidentally hit wrong key combination on num keyboard and font get bigger. I'm used to hitting reset font size when that happens but since 1.2.3 or 1.2.4 it just deletes font size setup.

mnquintana:
Thanks for the report! I can reproduce this on 1.2.4 - this is definitely a regression.

brentshermana:
The default font size for atom is size 14, which is defined in src/config-schema.coffee. I would expect a command called 'Reset font size'  to switch the font size to a default value, so I don't consider this a bug.

Perhaps what you have in mind is a command which undoes the last font change? That sounds a bit nichè, but it could be implemented pretty easily.

thomasjo:
@brentshermana It's supposed to reset to the user's default font size, i.e. what was originally specified in their `config.cson` file. This was fixed at some point, but then later regressed. This regression is still alive, but should be squashed.

@atom/feedback Anyone have any immediate thoughts on what might have caused this regression? Perhaps the config schema in package.json changes caused this somehow? We should definitely look into this, starting by implementing a failing spec to prevent further regressions in the future.

iolsen:
cc @damieng

lee-dohm:
https://github.com/atom/atom/pull/6539 is the PR where @mnquintana made the change to `reset-font-size` so that it restored the font size at window startup.

damieng:
I would imagine reset font size should reset the font size back to what it was before it was increased/decreased using the appropriate zoom functions.  

I'm not quite sure what that PR was aiming to achieve but the net effect seems to be it just clears the setting in your config for you.

thomasjo:
If you look at the commits, the PR watched for changes to font size in user config, and stored the original value prior. Then upon invoking reset, it sets font size to the original value instead of clearing the value. 

This used to work, but broke at some point. It's unfortunate that the spec from the PR did not catch the regression.

[...]

PR #19192, opened 2019-04-21
Fix reset-font-size

vinkla:
### Identify the Bug

#9873

### Description of the Change

If the user config is yet to be loaded `oldValue` equals the default font size `14px`. This breaks the reset font size functionality. Fetching the font-size from the config object instead of using `oldValue` seems to fix the issue. I'm running:

**macOS:** 10.14.3
**Node:** v11.14.0
**npm:** 6.9.0

### Alternate Designs

Not sure if this fix is needed at all TBH. [It has been pointed out that it works fine on Windows](https://github.com/atom/atom/issues/9873#issuecomment-457648226) and I know this has worked in previous versions of Atom on macOS as well. It could be an issue where the configuration hasn't yet been loaded for some reason.

### Possible Drawbacks

N/A

### Verification Process

- Set the configured font size to e.g. 10
- Increase or decrease the editor font size multiple times
- Reset the font size `window:reset-font-size`
- The font size should equal 10

### Release Notes

- Fix reset font size behaviour

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -1543,9 +1543,9 @@ module.exports = class Workspace extends Model {
   }
 
   subscribeToFontSize () {
-    return this.config.onDidChange('editor.fontSize', ({oldValue}) => {
+    return this.config.onDidChange('editor.fontSize', () => {
       if (this.originalFontSize == null) {
-        this.originalFontSize = oldValue
+        this.originalFontSize = this.config.get('editor.fontSize')
       }
     })
   }


vinkla:
[One of the AppVeyor tests seem to fail](https://ci.appveyor.com/project/Atom/atom/builds/24004593/job/9n08q5h9tyefxtib#L2201). It seems unrelated to this pull request.

50Wliu:
> [The AppVeyor tests failed.](https://ci.appveyor.com/project/Atom/atom/builds/24004593/job/9n08q5h9tyefxtib#L2201) It seems unrelated to this pull request. Right?

Yes; this is another instance of #18991.

vinkla:
@50Wliu is there something I can on my end to rerun the tests without having to close and open the pull request? Just want to get rid of the red cross.

50Wliu:
I restarted it :+1:

rafeca:
Thanks a lot for the contribution! Codewise it looks good to me, can you add a test to verify that the behaviour does not regress in the future? 🤗

vinkla:
Thanks for taking a look at the PR @rafeca! 

I would like to add tests but I'm not really sure how I can replicate the behaviour since it now works as expected. It should be covered with this test:

https://github.com/atom/atom/blob/cdb5ac98caa511d7f157c7a91297d40eea40044f/spec/workspace-spec.js#L1729-L1730

What is expected is that we should only save the `originalFontSize` when the user updates the font size through the editor settings page. Right now the `oldValue` always returns 14 and ignores the user settings.

rafeca:
Ok nevermind, let's just merge this one then :shipit:

PR #19206, opened 2019-04-24
Set the panel flag as focused when activate() is called

### LINKED ISSUE #18913: Programmatically calling pane.activate() does not update its isFocused() state
### Prerequisites

### Description

Programmatically activating a pane does not update the pane's `isFocused()` status (the pane is still reporting as not being focused), despite the docs stating that upon calling this function the pane should gain focus. The pane actually _does gain focus_ as I can type in the pane's view and perform all the expected operations inside the view; this is only an internal inconsistency with how the pane's focus is reported when the pane is focused programmatically.

### Steps to Reproduce

1. Programmatically activate a pane:
  ```js
  const pane = atom.workspace.paneForURI('atom://nuclide-terminal-view')
  // Check that the pane is not focused
  assert.ok(!pane.isFocused())
  pane.activate()
  ```
2. Programmatically observe the pane's focus state
  ```js
  assert.ok(pane.isFocused()) // fails
  ```

*Edit by @rsese: [alternate repro code snippet](https://github.com/atom/atom/issues/18913#issuecomment-479285351)*

```javascript
const assert = require('assert')
const pane = atom.workspace.paneForURI('atom://config')
assert.ok(!pane.isFocused())
pane.activate()
assert.ok(pane.isFocused()) 
```

**Expected behavior:** Programmatically focusing a pane should update it's `.isFocused()` status

**Actual behavior:** Programmatically focusing a pane does not update the pane's `.isFocused()` status, causing plugins to not recognise that pane as being focused

### Versions

```sh
Atom    : 1.35.0-beta0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

### Additional Information

I believe the `Pane.activate()` method should call `this.focus()` to toggle the focus programmatically. Alternatively, this is some Electron issue where the expectation would be that the "view layer" toggles the focus once the pane becomes activated, though I am uncertain how these things work in Atom. 🤔 

-----

#### Possibly related

The `Pane` class also has an `.activateItemForURI(URI)` method. This method's name suggests that the pane would become _activated_, much like what would happen when you call `.activate()` on the pane, thus gaining focus etc. However, this method does not make the pane _activated_ at all. In other words, to actually activate the pane, I have to manually activate it after activating the item for URI:

```js
pane.activateItemForURI('atom://nuclide-terminal-view')
// Dock does not appear if it is currently hidden, but the pane switches to the correct view
// To actually open the dock and focus the item, I have to activate the pane explicitly:
pane.activate()
```

Not sure if this is intended behaviour or a separate bug. 🤷‍♂️

rafeca:
### Identify the Bug

Programmatically calling `pane.activate()` does not currently update its `isFocused()` state.

This PR fixes https://github.com/atom/atom/issues/18913

### Description of the Change

This PR changes the logic to set the `focused` flag on the `activate()` method instead of the `focus()` method. Since `focus()` always calls `activate()` this should not affect that method.

### Alternate Designs

N/A

### Possible Drawbacks

🤷‍♂️

### Verification Process

Open Atom and execute the following from the developer console:

```js
const assert = require('assert')
const pane = atom.workspace.paneForURI('atom://config')
assert.ok(!pane.isFocused())
pane.activate()
assert.ok(pane.isFocused()) 
```

### Release Notes

Fixed `pane.activate()` to update the panel `isFocused()` state.

### DIFF
--- a/src/pane.js
+++ b/src/pane.js
@@ -393,7 +393,6 @@ class Pane {
 
   // Called by the view layer to indicate that the pane has gained focus.
   focus () {
-    this.focused = true
     return this.activate()
   }
 
@@ -1011,6 +1010,8 @@ class Pane {
   // Public: Makes this pane the *active* pane, causing it to gain focus.
   activate () {
     if (this.isDestroyed()) throw new Error('Pane has been destroyed')
+    this.focused = true
+
     if (this.container) this.container.didActivatePane(this)
     this.emitter.emit('did-activate')
   }


nathansobo:
Nice, simple change.

PR #19208, opened 2019-04-25
Remove sudo settings from .travis.yml

abetomo:
It does not need to set sudo.
 Remove sudo settings from .travis.yml
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

I was not able to determine which template is best, so please let me know if there is a better template to use.


rafeca:
Thanks! ✨

PR #19242, opened 2019-04-30
Use async randomBytes API to speedup Windows startup

### LINKED ISSUE #19239: Performance regression in v1.36.1
By looking at the [Atom usage metrics](https://github.com/atom/metrics), we've discovered that the [Shell load time](https://github.com/atom/metrics/blob/master/docs/events.md#shell-load-time) (which is a proxy for the Atom startup time) has regressed on the v1.36.1 release on Windows machines:

Screen Shot 2019-04-30 at 15 14 53

After doing some investigation, this regression was introduced in https://github.com/atom/atom/pull/19109, and more specifically it's caused [by this line](https://github.com/atom/atom/blob/master/src/main-process/atom-application.js#L79), which takes up to a second to execute on Windows (blocking the main process and therefore the startup of Atom).

Looks like this issue may be caused by https://github.com/electron/electron/issues/2073

rafeca:
This PR changes a bit the logic of `atom-application` to make the authentication through sockets async, which fixes https://github.com/atom/atom/issues/19239.

In order to make the code async and keep it as robust as possible a few changes needed to be done (basically to delay the execution of some logic while avoiding as much as possible potential race conditions).

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -15,6 +15,7 @@ const path = require('path')
 const os = require('os')
 const net = require('net')
 const url = require('url')
+const {promisify} = require('util')
 const {EventEmitter} = require('events')
 const _ = require('underscore-plus')
 let FindParentDir = null
@@ -75,18 +76,21 @@ const getExistingSocketSecret = (atomVersion) => {
   return fs.readFileSync(socketSecretPath, 'utf8')
 }
 
-const createSocketSecret = (atomVersion) => {
-  const socketSecret = crypto.randomBytes(16).toString('hex')
+const getRandomBytes = promisify(crypto.randomBytes)
+const writeFile = promisify(fs.writeFile)
 
-  fs.writeFileSync(getSocketSecretPath(atomVersion), socketSecret, {encoding: 'utf8', mode: 0o600})
+const createSocketSecret = async (atomVersion) => {
+  const socketSecret = (await getRandomBytes(16)).toString('hex')
+
+  await writeFile(getSocketSecretPath(atomVersion), socketSecret, {encoding: 'utf8', mode: 0o600})
 
   return socketSecret
 }
 
-const encryptOptions = (options, secret) => {
+const encryptOptions = async (options, secret) => {
   const message = JSON.stringify(options)
 
-  const initVector = crypto.randomBytes(16)
+  const initVector = await getRandomBytes(16)
 
   const cipher = crypto.createCipheriv('aes-256-gcm', secret, initVector)
 
@@ -144,10 +148,12 @@ class AtomApplication extends EventEmitter {
 
     return new Promise(resolve => {
       const client = net.connect({path: socketPath}, () => {
-        client.write(encryptOptions(options, socketSecret), () => {
-          client.end()
-          app.quit()
-          resolve(null)
+        encryptOptions(options, socketSecret).then((encryptedOptions) => {
+          client.write(encryptedOptions, () => {
+            client.end()
+            app.quit()
+            resolve(null)
+          })
         })
       })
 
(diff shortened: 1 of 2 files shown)


src/main-process/atom-application.js
```diff
@@ -15,6 +15,7 @@ const path = require('path')
 const os = require('os')
 const net = require('net')
 const url = require('url')
+const {promisify} = require('util')
```
smashwilson: Oh hey! I didn't realize we had access to `promisify` now. This will come in handy.

src/main-process/atom-application.js
```diff
@@ -227,11 +230,16 @@ class AtomApplication extends EventEmitter {
     this.applicationMenu = new ApplicationMenu(this.version, this.autoUpdateManager)
     this.atomProtocolHandler = new AtomProtocolHandler(this.resourcePath, this.safeMode)
 
-    this.listenForArgumentsFromNewProcess()
+    // Don't await for the following method to avoid delaying the opening of a new window.
+    // (we await it just after opening it).
```
smashwilson: 👍 Excellent.

src/main-process/atom-application.js
```diff
@@ -421,10 +429,15 @@ class AtomApplication extends EventEmitter {
   // You can run the atom command multiple times, but after the first launch
   // the other launches will just pass their information to this server and then
   // close immediately.
-  listenForArgumentsFromNewProcess () {
-    if (!this.socketPath) return
+  async listenForArgumentsFromNewProcess (options) {
+    if (!options.test && !options.benchmark && !options.benchmarkTest) {
```
smashwilson: I still wish we had a more succinct way to express "a regular editor window" in these options. (Out of scope for this PR though.)

smashwilson (approved):
👍 LGTM

src/main-process/atom-application.js
```diff
@@ -75,18 +76,26 @@ const getExistingSocketSecret = (atomVersion) => {
...
 
   return socketSecret
 }
 
 const encryptOptions = (options, secret) => {
   const message = JSON.stringify(options)
 
-  const initVector = crypto.randomBytes(16)
+  // Even if the following IV is not cryptographically secure, there's a really good chance
+  // it's going to be unique between executions which is the requirement for GCM.
```
nathansobo: I think it might be helpful to future readers if you explain the motivation for avoiding cryptographic randomness on this code path (that it's too slow).

src/main-process/atom-application.js
nathansobo: Again, worth mentioning explicitly that this method is slow on Windows?

nathansobo (commented):
I think it might be helpful to clarify the motivations behind not awaiting in comments. So just checking my understanding... this increases the window of time in which a race condition might exist, correct? I suppose there was always a small window anyway if you ran `atom` twice quickly before the first Electron could start listening on the socket. This just makes it more likely.

Another question: Is there some way we can remember to revisit this when we upgrade Electron and see if `crypto` is still so slow?

rafeca:
> I think it might be helpful to clarify the motivations behind not awaiting in comments. So just checking my understanding... this increases the window of time in which a race condition might exist, correct? I suppose there was always a small window anyway if you ran atom twice quickly before the first Electron could start listening on the socket. This just makes it more likely.

Exactly, it'll make this window (which is already a few hundred ms) unnoticeable longer in OSX/Linux (a few ticks) and quite much longer (up to a second) in Windows (at least until we upgrade to electron v3).

I think that's a fair tradeoff, since the only side effect that this will cause is that two Atom windows will get opened under that situation.

> Another question: Is there some way we can remember to revisit this when we upgrade Electron and see if crypto is still so slow?

Maybe we can use a specially-crafted comment that's easy to search on GitHub for the whole Atom org, and then we add a point to the electron upgrade manual doc (if it exists) to revisit all these messages whenever we upgrade the electron version?

Something like:

```js
// TodoElectronIssue: `randomBytes` is slow on electron v2. This issue is fixed in v3.
```

(I'm not using spaces/special characters since GitHub search is particularly bad when searching for non-words).

50Wliu:
> Another question: Is there some way we can remember to revisit this when we upgrade Electron and see if crypto is still so slow?

In addition to what @rafeca suggested, we could also create a tracking issue for it and apply the `electron` label, as our Electron upgrade checklist specifies to check each one and see if they have been fixed/can be resolved with the new Electron version.

src/main-process/atom-application.js
rafeca: I've added more explicit comments for both things in https://github.com/atom/atom/pull/19246

PR #19244, opened 2019-04-30
Default to @atom/notify for file system notifications

### LINKED ISSUE #18991: Flaky build: Windows build sometimes fails with "resource busy or locked"

nathansobo:
### Description of the Change

The library that we are currently using for file system notifications ([`@atom/nsfw`](https://github.com/atom/nsfw) is [causing render process crashes on Windows](https://github.com/atom/atom/issues/18991), which is introducing unreliability in our continuous integration pipeline and presumably causing occasional crashes for users. After obtaining crash reports on a local VM, I was unable to find the cause of the crashes after trying for about a half a day.

We've already forked `nsfw` for other issues in the past, and @smashwilson actually built an alternative library called [`@atom/watcher`](https://github.com/atom/watcher) in hopes of addressing `nsfw`'s shortcomings for our use case. It's been possible to use `@atom/watcher` as the back end for file system notifications in Atom for about 2 years now via the "Experimental file system watching library" and "Poll" options in the core settings.

Screen Shot 2019-04-30 at 3 14 51 PM

There's an [open pull request](https://github.com/atom/atom/pull/17899/files) that enables this option by default, and my first reaction upon encountering the inexplicable crashes from `nsfw` was to get that PR over the finish line. Unfortunately, `@atom/watcher` is *also* the source of crashes on Windows in CI builds for that PR. After about a day trying, I was unsuccessful in reproducing these crashes in a local VM and unable to associate symbols with the crash report downloaded from CI.

This leaves us in a difficult position. We need a *stable*, *cross-platform* provider of file system notifications to drive several critical aspects of Atom's UI and to support the `watchPath` API that is exposed to third party packages. Both of our current solutions are crashing on Windows in C++ code paths that are hard to debug.

This PR aims to provide an alternative. I built a new module for file system notifications called [`@atom/notify`](https://github.com/atom/notify). It is different from `nsfw` and `watcher` in that it is not a native add-on that uses C++ to access cross platform file system notification APIs. Instead, it leans on a mature, cross-platform implementation of file system notifications provided in the Rust [`notify` crate](https://docs.rs/notify/4.0.10/notify/). The node module embeds a small Rust executable that wraps the `notify` API. It spawns this executable as a subprocess and communicates with it over stdio.

The `notify` crate takes care of the heavy lifting such as interfacing with the appropriate APIs on all three platforms and debouncing the event stream to coalesce related events. Since it is written in Rust and used widely in that ecosystem, I am confident in its stability. The subprocess eats about 680KB of memory on my machine, but I think the simplicity of talking to Rust over stdio versus building a native add-on is worth it.

### Alternate Designs

* We could fix the crash in `nsfw`. I tried to figure it out but got nowhere after several hours.
* We could fix the crash in `@atom/watcher`, but I had similar challenges finding its source. Also, `@atom/watcher` is a fairly complex C++ codebase that nobody else is really using but Atom, and I'm not certain we wouldn't find additional issues even if we fixed this one. @smashwilson is the only one with significant context on it, and we can't guarantee that he'll always be available to the Atom team in the future.
* We could go with VS Code's solution. They use [`chokidar`](https://github.com/paulmillr/chokidar) on macOS and Linux and spawn [a custom .Net subprocess](https://github.com/Microsoft/vscode-filewatcher-windows) on Windows. This solution still involves a subprocess, and reusing their code wouldn't have been a drop in replacement since their implementation exists across several files in the VS Code repository. I feel like using the same JS code and subprocess across all three platforms will lead to more consistent results and be cleaner from a code perspective.

### Possible Drawbacks

* Turning `@atom/notify` on by default introduces risk, but if we're hitting lots of issues on nightly and beta we can turn it off again before we hit stable, or *only* turn it on by default on non-stable release channels.
* Spawning a subprocess introduces some memory overhead. It's unclear how much smaller a native Node add-on might be, but I think 600-700KB for a stable implementation of a critical facility is worth it. There are other places where we could more easily save memory, such as the GitHub package's sidecar process.
* Rust is a new language that we haven't used yet anywhere else in Atom, and it has a fairly steep learning curve. However, we aren't introducing very *much* Rust (about 250 lines), and it's all pretty straightforward with a well-defined, narrow interface. I'd argue that for the use case in question, Rust is a more approachable and much safer alternative to C++, and that team member investment in learning it could pay off when attacking other problems.
* I'm probably going to require a window reload when you change file system notification back-ends in the settings view, because the code for automatically swapping things is complex and not an easy drop-in for the new implementation.

### Verification Process

* I have some test coverage on the `@atom/notify` module which I am running on all three platforms. I plan to make it more extensive before merging this PR.

### Release Notes

* Atom now defaults to [a new back-end for file system notifications](https://github.com/atom/notify) based on the Rust [`notify`](https://docs.rs/notify/4.0.10/notify/) crate. If you're experiencing any issues that seem related, you can switch back to the previous implementation by choosing `Native operating system APIs` as your `File Watcher` in core settings.

### Remaining tasks

* [x] Fix issues building the V8 snapshot
* [x] Get `@atom/notify` tests passing again on Linux
* [x] Prompt to reload when switching notification back-ends in settings
* [x] Green builds on every platform
  * [x] macOS
  * [x] Linux
  * [x] Windows
* [x] Decide how to implement the `onDidError` method on the returned watcher.
* [x] Surface `@atom/notify` errors to the user via a notification so we can get feedback
* [x] Test more edge cases on all platforms – They exist, but we're not going to solve them. [See comment below](https://github.com/atom/atom/pull/19244#issuecomment-490277764)
* [x] Enable "Poll" option
* [x] ~~Preserve pre-existing default on the stable channel but use `@atom/notify` otherwise~~
* [x] Document `@atom/notify` (both the README and the source code)
* [x] Address windows failures observed in GitHub package 😞 

/cc @atom/maintainers 
Closes https://github.com/atom/atom/issues/18991


jasonrudolph:
@nathansobo: A fantastic write-up, as always. :zap: :bow:  Not only does this write-up help the reader understand all the considerations that you're taking into account *today*, but it will be a great resource for "Future Us" to easily refer back to these considerations and tradeoffs in the months and years to come. :heart_decoration:

nathansobo:
Ugh. My Windows and Linux builds just seem to be hung on Azure Pipelines. Not getting any output from the tests.

nathansobo:
💚

nathansobo:
Today I investigated what happens when a watched directory itself is removed or renamed. At least on macOS, `@atom/nsfw` throws an error and stops watching.

The Rust `notify` crate doesn't error, but its behavior in this scenario is platform-dependent and inconsistent. I spent a few hours today trying to iron out the inconsistencies with additional code in the subprocess that wraps the `notify` crate, but it's starting to feel like a quagmire.

In attempting to test rename logic, I discovered that the `notify` crate doesn't do a good job coalescing raw events from macOS's `FSEvents` API to report renames when files are created and renamed in rapid succession. There's an [open issue](https://github.com/passcod/notify/issues/181) about the problem on the repository, and they actually mention the crate's integration tests sleeping for *35 seconds* for the `FSEvents` buffer to clear in certain scenarios 😬. When I fire up a Node process and watch an existing directory, everything works fine, but the pressure of rapid-fire file system operations in automated tests seems to trigger weirdness.

At this point, I see the following paths forward:

1. Dive into the `notify` code and try to fix https://github.com/passcod/notify/issues/181 so I can write reliable automated tests that rename files, then proceed to try to paper over platform inconsistencies around renaming watched directories. Is a fix even possible? Not sure. I suspect that it probably is, but it's unclear how long it would take.

2. Ignore this issue. If you rename watched directories, you get the weird platform specific behavior. Also, if you create a file and then quickly rename it, you might not always get perfect rename events.

I'm leaning toward option 2. The primary use case for `watchPath` is watching project roots, and Atom doesn't deal well with those being renamed anyway for other reasons. The rapid create-rename scenario happens in automated tests, but I don't see it being a huge issue in practice.

Alas, the `notify` crate isn't as perfect as I'd hoped, but it's still the case that it's used in plenty of production Rust projects including Cargo and *should* work well in practice (🤞). Despite these imperfections, this path still solves the render process crashes that are slowing down our build cycle and potentially makes Electron upgrades easier.

50Wliu:
> I'm leaning toward option 2.

I agree - since it doesn't work at the moment, there's no real pressure for it to work with notify. It's a nice-to-have, for sure, but something that can be investigated later.

nathansobo:
I just saw what looks to be a timing-related failure in the GitHub package related to file system events 😢.

```
  1) Repository‌
       cache invalidation
         from filesystem events
           when committing:
     Error: Timeout of 20000ms exceeded.
```

I'm thinking about what to do about this. I almost think that polling on a fairly tight interval in tests might be our best bet. Unfortunately, I think a large percentage of our tests start watchers indirectly that don't ever end up being used.

nathansobo:
Just realized it's not possible to set a schema default dynamically without interfering with the startup snapshot. So, new plan. If Atom 1.38 is approaching stable and we're not feeling confident about `@atom/notify`, we can swap out the default before release.

nathansobo:
Marking this ready for review. I need a plan for the `watchPath`-related failures in the GitHub package but otherwise we should be good to go.

src/path-watcher.js
```diff
@@ -608,83 +577,91 @@ class PathWatcherManager {
...
+      await Promise.all(
+        Array.from(this.live, ([, w]) => w.stop())
+      )
     }
-
-    return Promise.all(
-      Array.from(this.live, ([, w]) => w.stop())
-    )
   }
+
```
nathansobo: @smashwilson I left these methods in here to get your thoughts on whether I can go ahead and delete them. They seem to be unused.

src/path-watcher.js
```diff
@@ -736,24 +713,24 @@ function stopAllWatchers () {
...
-// Private: Sample usage statistics for the active watcher.
-watchPath.status = function () {
-  return PathWatcherManager.active().status()
-}
-
-// Private: Configure @atom/watcher ("experimental") directly.
-watchPath.configure = function (...args) {
-  return watcher.configure(...args)
-}
+// // Private: Show the currently active native watchers in a formatted {String}.
```
nathansobo: Ditto. We can drop these, right?

src/path-watcher.js
smashwilson: 👍 Yeah, these are diagnostic methods used to investigate stuff like failing tests in CI, or as a diagnostic tool to investigate odd behaviors. If they don't make sense with @atom/notify, :fire:

src/path-watcher.js
smashwilson: 👍

smashwilson (approved):
👍 No blockers that I see.

src/config-schema.js
```diff
@@ -351,6 +351,11 @@ const configSchema = {
           }
         ]
       },
+      fileSystemWatcherPollInterval: {
+        description: "If the 'Polling' option is selected for the file system watcher, this will be the interval between polls.",
+        type: 'number',
+        default: 1000
```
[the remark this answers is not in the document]
nathansobo: Yeah, I played with it and 3 seconds just felt really awful. 1 second felt sane though. Definitely want your computer to be plugged in with this option on a big folder I'd imagine. But I want the option to feel usable with this default.

nathansobo:
Okay, I'm not sure how I've gotten *any* green builds before (you can see from the record that they did exist!), but I'm now getting failures due to the GitHub package making assumptions about `watchPath` reporting paths in the shortened 8.3 format. So we're comparing paths containing `VSSADM~1` with paths containing `VssAdministrator` on Azure and barfing.

I'm going to explore using Windows APIs to normalize to the 8.3-compatible paths in the Rust subprocess and see how we do.

nathansobo:
Trying a new approach where I normalize `%TEMP%` to a long path in the build configuration before running the tests.

nathansobo:
The last build was green on Linux and Windows but failed in the fuzzy finder specs on macOS for a seemingly unrelated reason. Merging in master and trying again. 😓

nathansobo:
Okay. Green build. Going to wait to merge this until after the next release cycle which should occur early next week. That will maximize our time to test the new notification backend on nightly and beta before it goes to stable. Worst case scenario, we change the default while dealing with anything unforeseen.

smashwilson:
Yessss ✨ 🚀 

When I get the chance, I'll take a look at flushing any flakiness out of the github package cache invalidation tests.

rafeca:
Leaving this comment for the future reader: in case we want to explore making `@atom/notify` the default watcher, we'll need to tackle https://github.com/atom/atom/issues/19333 first.

PR #19246, opened 2019-05-01
Do not wait until config watcher is ready in order to launch the Atom window

rafeca:
Based on some profiling done locally, this `await` takes ~300-500ms on Windows machines (I couldn't find the root cause of it but may be related to https://github.com/atom/atom/pull/19242).


smashwilson (approved):
👍 Good catch.

50Wliu:
Do you think a comment should be added saying why it's not awaited? Because I could potentially see someone changing it back in the future if they weren't aware.

nathansobo (commented):
I'm concerned about this introducing a race condition where we don't catch config file changes that occur just after the window starts up. Does it make sense to skip the `await` only on Windows so that we only introduce the race condition when it is the lesser of two evils? The downside of *that* is that we're introducing inconsistent behavior across platforms.

How can we remember to revisit this change once we upgrade Electron and see if it's still an issue?

rafeca:
> I'm concerned about this introducing a race condition where we don't catch config file changes that occur just after the window starts up. Does it make sense to skip the await only on Windows so that we only introduce the race condition when it is the lesser of two evils? The downside of that is that we're introducing inconsistent behavior across platforms.

I can do that

> How can we remember to revisit this change once we upgrade Electron and see if it's still an issue?

What do you think about what I suggested in https://github.com/atom/atom/pull/19242#issuecomment-488337557 ?

src/main-process/atom-application.js
```diff
@@ -255,7 +261,14 @@ class AtomApplication extends EventEmitter {
...
     if (!this.configFilePromise) {
       this.configFilePromise = this.configFile.watch()
-      this.disposable.add(await this.configFilePromise)
+
+      // TodoElectronIssue: In electron v2 awaiting the watcher causes some delay
+      // in Windows machines, which affects directly the startup time.
+      if (process.platform === 'win32') {
+        this.configFilePromise.then(disposable => this.disposable.add(disposable))
+      } else {
+        this.disposable.add(await this.configFilePromise)
```
nathansobo: The `this.config.onDidChange` calls below this `await` really do need to happen *after* the config file is loaded. That is what is causing #19323. I have a fix in the works and will get it out later today.

nathansobo: Fixed in #19354.

PR #19255, opened 2019-05-02
Add async into package.json and bump async in package-lock.json to 2.6.2.

szb512:
### Description of the Change

Async is a utility module which provides straight-forward, powerful functions for working with asynchronous JavaScript. Although originally designed for use with Node.js and installable via npm install --save async, it can also be used directly in the browser.

This version of the package is optimized for the Node.js environment. If you use Async with webpack, install async-es instead.

Add powerful functions working with nodejs, with huge support backing it.

### Possible Drawbacks

Possible side-effects include more commands to deal with. Negative impacts include more possible failures.

### Verification Process
I ran all supported commands for async, with many passes and 1 failure.

### Applicable Issues

No applicable issues.

### Release Notes

You can now use lots of packages bundled in async + have more support for different real-life scenarios that involve async.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,10 @@
           "version": "0.2.0",
           "bundled": true
         },
+          "async": {
+          "version": "2.6.2",
+          "bundled": true
+        },
         "asynckit": {
           "version": "0.4.0",
           "bundled": true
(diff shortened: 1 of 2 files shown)


lee-dohm:
Thanks for the suggestion.

Packages can load whatever Node modules they need without them having to be included in the base Atom installation. Because of this, we don't see a need to make this change.

PR #19257, opened 2019-05-03
Increase the timeout for the FileRecoveryService tests

### LINKED ISSUE #19235: Flaky build: atom-application test sometimes fails on Linux
In our Linux builds on the Azure pipelines, sometimes the `atom-application` test step fails with a timeout:

```
  1) FileRecoveryService when no crash happens during a save creates a recovery file and deletes it after saving:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at <anonymous>:null:null
  

  2) FileRecoveryService when a crash happens during a save restores the created recovery file and deletes it:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at <anonymous>:null:null
```

---

**Affected platform(s)**: Linux

**Example commit experiencing flakiness**: https://github.com/atom/atom/commit/0219306c873dc3b8cfda640007b1f7f15a7df170

**Example failing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=39012

**Example passing build at that commit**: https://github.visualstudio.com/Atom/_build/results?buildId=39028

rafeca:
This PR increases the timeout of the `FileRecovery` tests to 10s, so we don't get the weird timeouts that have started occurring on Linux recently.

This fixes https://github.com/atom/atom/issues/19235

### DIFF
--- a/spec/main-process/file-recovery-service.test.js
+++ b/spec/main-process/file-recovery-service.test.js
@@ -8,8 +8,10 @@ const sinon = require('sinon')
 const { escapeRegExp } = require('underscore-plus')
 const temp = require('temp').track()
 
-describe('FileRecoveryService', () => {
+describe('FileRecoveryService', function () {
   let recoveryService, recoveryDirectory, spies
+
+  this.timeout(10 * 1000)
 
   beforeEach(() => {
     recoveryDirectory = temp.mkdirSync('atom-spec-file-recovery')


rafeca:
I'm gonna run the CI tests for this PR a few times to make sure that the test does not timeout anymore.

In order to verify that this PR reduces the flakiness, I'm hoping to see a few tests that take longer than 2s to execute but less than 10s ([example of test result](https://github.visualstudio.com/Atom/_build/results?buildId=39738&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=6c939d89-0d1a-51f2-8b30-091a7a82e98c&lineStart=240&lineEnd=248&colStart=1&colEnd=108)).

rafeca:
I've run half a dozen of CI builds with this PR and haven't seen any test failure, so it's good to merge

PR #19279, opened 2019-05-07
Improve handling of comment characters in ERB and EJS

jasonrudolph:
### Identify the Bug

https://github.com/atom/language-ruby/issues/263 identifies a bug where Ruby line comments can break ERB syntax highlighting. This pull request is the first step in resolving that bug. Once this pull request is merged, we'll need to make a [small change in language-html](https://github.com/atom/language-html/pull/227) to use the new abilities introduced in this pull request. Once this pull request and https://github.com/atom/language-html/pull/227 are both merged, it will resolve the bug.

### Description of the Change

In languages like ERB and EJS, there is an implicit newline at the end of each directive. For example, in the ERB below, even though it only involves one line, there is an implicit newline between the two ERB directives, thus resulting in two separate Ruby statements.

```
<% def some_method %><% end %>
```

In Ruby, this becomes:

```rb
def some_method
end
```

The implicit newline is essential, because `def some_method end` isn't valid Ruby.

Prior to this pull request, these implicit newlines worked correctly in most cases, but didn't work correctly for comments. When an ERB directive contained a Ruby-style comment (or a JavaScript style comment in EJS), all subsequent ERB directives were treated as a continuation of that comment. (See the before and after screenshots below.)

In this pull request, we add the ability for injection points to specify whether an implicit newline should be applied between each directive. For example, 11cdac739271eb027c9b6b8c236d4c4e0cfef25f makes use of this new option to fix the bug identified in #263:

```javascript
atom.grammars.addInjectionPoint('text.html.erb', {
  type: 'template',
  language (node) { return 'ruby' },
  content (node) { return node.descendantsOfType('code') },
  newlinesBetween: true
})
```

#### Before and After

The screenshots below use following code sample:

```
<%# some erb-style comment %>

<% puts 'some ruby code' %>

<% puts 'some ruby code followed by a comment' # some ruby-style-comment %>

<% puts 'now everything is treated as a continuation of the previous comment' %>
```

##### Before

Screen Shot 2019-05-08 at 11 07 55 AM

##### After

Screen Shot 2019-05-08 at 11 08 37 AM

### Alternate Designs

See the "Possible Solutions" section in https://github.com/tree-sitter/tree-sitter/issues/327.

### Possible Drawbacks

The approach implemented in this pull request only works if a newline character exists between the two directives; however, we expect this to cover a majority of cases. Future enhancements in Tree-sitter may provide a way to remove this limitation (https://github.com/tree-sitter/tree-sitter/issues/327).

### Verification Process

- [x] Open JavaScript file and verify proper syntax highlighting of interspersed functions and comments
- [x] Open Ruby file and verify proper syntax highlighting of interspersed functions and comments
- [x] Open HTML+ERB file and verify proper syntax highlighting of interspersed ERB comment directives, ERB directives with Ruby comments, and ERB directives without Ruby comments
- [x] Open HTML+EJS file and verify proper syntax highlighting of interspersed EJS comment directives, EJS directives with JavaScript comments, and EJS directives without JavaScript comments

### Release Notes

N/A: See https://github.com/atom/atom/pull/19291.


spec/tree-sitter-language-mode-spec.js
```diff
@@ -766,6 +766,61 @@ describe('TreeSitterLanguageMode', () => {
...
+          ],
+          [
+            { text: '<%', scopes: ['directive'] },
+            { text: ' ', scopes: [] },
+            { text: 'b', scopes: ['function'] },
+            { text: '() ', scopes: [] },
+            { text: '%>', scopes: ['directive'] }
+          ]
+        ])
+      })
```
jasonrudolph: @maxbrunsfeld: This test borrows heavily from [the preceding test](https://github.com/atom/atom/blob/c163564a9e97820d82b66a4934fbb52999a9ecd0/spec/tree-sitter-language-mode-spec.js#L690-L767), and attempts to cover the updates introduced in ef94657438f4b3a70f0e0a7cc7895939626f1c72.

I would expect to end up with the tokens seen above, but I'm actually ending up with a slightly different set of tokens:

```javascript
[
  { text: '<%', scopes: ['directive'] },
  { text: ' // js comment', scopes: [] },
  { text: '%>', scopes: ['directive'] },
  { text: '', scopes: ['html'] }
],
[
  { text: '<%', scopes: ['directive'] },
  { text: ' ', scopes: [] },
  { text: 'b', scopes: ['function'] },
  { text: '() ', scopes: [] },
  { text: '%>', scopes: ['directive'] }
]
```

A couple questions for you:

- In general, how does this approach compare to what you had in mind for testing the updates introduced in ef94657438f4b3a70f0e0a7cc7895939626f1c72?
- If this approach is consistent with what you had in mind, do you happen to see anything that I might be doing wrong, so that I can understand why I'm getting different results than I'm expecting? I'm wondering if there's some change I need to make to the way the EJS grammar gets set up around [here](https://github.com/atom/atom/blob/c163564a9e97820d82b66a4934fbb52999a9ecd0/spec/tree-sitter-language-mode-spec.js#L776-L779).

(By the way, if it's easier to push changes to this branch than to suggest them via prose, feel free to push directly to this pull request. I'll study any changes you push up to make sure I understand what needed to change and why.)

jasonrudolph: I chatted with @maxbrunsfeld in Slack, and he confirmed that this approach is what he had in mind for testing this behavior. He also helped me figure out what needed to change in the test, and e00ea44310da1d9727ce4c66c232f2f01e4a77db makes that change.

spec/tree-sitter-language-mode-spec.js
```diff
@@ -766,6 +766,61 @@ describe('TreeSitterLanguageMode', () => {
...
+            { text: '', scopes: ['html'] }
+          ],
+          [
+            { text: '<%', scopes: ['directive'] },
+            { text: ' ', scopes: [] },
+            { text: 'b', scopes: ['function'] },
+            { text: '() ', scopes: [] },
+            { text: '%>', scopes: ['directive'] }
+          ]
+        ])
```
jasonrudolph: Prior to the changes to `src/tree-sitter-language-mode.js` introduced in this pull request, Atom would produce the following tokens when executing this test:

```javascript
[
  [
    { text: '<%', scopes: ['directive'] },
    { text: ' // js comment ', scopes: [] },
    { text: '%>', scopes: ['directive'] },
    { text: '', scopes: ['html'] }
  ],
  [
    { text: '<%', scopes: ['directive'] },
    { text: ' b() ', scopes: [] },
    { text: '%>', scopes: ['directive'] }
  ]
]
```

Note that in this tokenization ☝️, `b()` is not tokenized as a function. Instead, `b()` is treated as a continuation of the comment on the preceding line:

```
<% // js comment %>
<% b() %>
```

The enhancements in this pull request (specifically ef94657438f4b3a70f0e0a7cc7895939626f1c72) allow Atom to correctly identify that the JavaScript comment is implicitly terminated at the end of the first line, and the second line begins a new JavaScript statement.

maxbrunsfeld: I think it might help tell the story if you added a scope mapping for comment tokens in the `jsGrammar` defined above in the `beforeEach`:

```js
jsGrammar = new TreeSitterGrammar(atom.grammars, jsGrammarPath, {

  // ...

  scopes: {

    // ...

    'comment': 'comment'
  }
})
```

That way, you'd see a `comment` token in the expected output, something like:

```js
{ text: '<%', scopes: ['directive'] },
{ text: ' ', scopes: [] },
{ text: '// js comment ', scopes: ['comment'] },
{ text: '%>', scopes: ['directive'] },

```

jasonrudolph: Sweet! Thanks, @maxbrunsfeld. I explored something similar this morning, but couldn't get it to cooperate for some reason. Thanks for pointing me in the right direction.

f18ad1eea191821bc26ee4ba24cb5e7eb5f4b92d implements this suggestion.

maxbrunsfeld (approved):

PR #19297, opened 2019-05-09
Add markers for more granular logging of startup time

rafeca:
## Context

Currently we're only logging two measures to give us information about the startup performance of Atom: the shell load time and the core load time.

They have been useful to identify performance regressions so far, but they are not enough to debug potential bottlenecks in Atom startup in real-life conditions.

## Summary of changes

This PR aims to give us much more granular data about the startup performance of Atom by doing two things:
* Allowing us to create arbitrary markers at any point during the startup of Atom which will get logged via the `metrics` package.
* Unifying the startup time taken by the main process and the renderer process by allowing to create markers anywhere.

As an example, with this PR these are the markers created during the startup of Atom (the ones happening in the main process are prepended by `main-process:` and the ones happening in the renderer process start with `window:`):

```json
{
  "startTime":1557502272939,
  "markers":[
    {
      "label":"main-process:start",
      "time":0
    },
    {
      "label":"main-process:electron-onready:start",
      "time":51
    },
    {
      "label":"main-process:electron-onready:end",
      "time":252
    },
    {
      "label":"main-process:atom-application:open",
      "time":324
    },
    {
      "label":"main-process:atom-application:constructor:start",
      "time":324
    },
    {
      "label":"main-process:atom-application:constructor:end",
      "time":329
    },
    {
      "label":"main-process:atom-application:initialize:start",
      "time":329
    },
    {
      "label":"main-process:atom-application:create-window",
      "time":446
    },
    {
      "label":"main-process:atom-window:start",
      "time":447
    },
    {
      "label":"main-process:atom-window:end",
      "time":463
    },
    {
      "label":"main-process:atom-application:initialize:end",
      "time":480
    },
    {
      "label":"window:start",
      "time":687
    },
    {
      "label":"window:onload:start",
      "time":693
    },
    {
      "label":"window:setup-window:start",
      "time":729
    },
    {
      "label":"window:initialize:start",
      "time":1156
    },
    {
      "label":"window:environment:start-editor-window:start",
      "time":1197
    },
    {
      "label":"window:onload:end",
      "time":1198
    },
    {
      "label":"window:environment:start-editor-window:display-window",
      "time":1206
    },
    {
      "label":"window:environment:start-editor-window:load-packages",
      "time":1445
    },
    {
      "label":"window:environment:start-editor-window:deserialize-state",
      "time":1598
    },
    {
      "label":"window:environment:start-editor-window:activate-packages",
      "time":3222
    },
    {
      "label":"window:environment:start-editor-window:open-editor",
      "time":3862
    },
    {
      "label":"window:environment:start-editor-window:end",
      "time":3893
    },
    {
      "label":"window:initialize:end",
      "time":3894
    },
    {
      "label":"window:setup-window:end",
      "time":3894
    }
  ]
}
```

### Verification steps

- [ ] Check that the markers get correctly created when opening a new Atom window from scratch.
- [ ] Check that markers do not get created under the following conditions:
  - [ ] When refreshing the Atom window
  - [ ] When opening a new window via Open new WIndow"
  - [ ] When opening a new window via Re-open project"
  - [ ] When running `atom` CLI on a different path (which opens a new window).

### Next steps

- [ ] Create some automated tests.

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -3,16 +3,29 @@
   // when using `NativeCompileCache` for the first time.
   require('cached-run-in-this-context')
 
+  // Define the window start time before the requires so we get a more accurate
+  // window:start marker.
+  const startWindowTime = Date.now()
+
   const electron = require('electron')
   const path = require('path')
   const Module = require('module')
   const getWindowLoadSettings = require('../src/get-window-load-settings')
+  const StartupTime = require('../src/startup-time')
   const entryPointDirPath = __dirname
   let blobStore = null
   let useSnapshot = false
 
+  const startupMarkers = electron.remote.getCurrentWindow().startupMarkers
+
+  if (startupMarkers) {
+    StartupTime.importData(startupMarkers)
+  }
+  StartupTime.addMarker('window:start', startWindowTime)
+
   window.onload = function () {
     try {
+      StartupTime.addMarker('window:onload:start')
       const startTime = Date.now()
 
       process.on('unhandledRejection', function (error, promise) {
@@ -64,12 +77,16 @@
       if (getWindowLoadSettings().profileStartup) {
         profileStartup(Date.now() - startTime)
       } else {
-        setupWindow()
+        StartupTime.addMarker('window:setup-window:start')
+        setupWindow().then(() => {
+          StartupTime.addMarker('window:setup-window:end')
+        })
         setLoadTime(Date.now() - startTime)
       }
     } catch (error) {
       handleSetupError(error)
     }
+    StartupTime.addMarker('window:onload:end')
   }
 
   function setLoadTime (loadTime) {
@@ -103,7 +120,11 @@
 
     const initScriptPath = path.relative(entryPointDirPath, getWindowLoadSettings().windowInitializationScript)
     const initialize = useSnapshot ? snapshotResult.customRequire(initScriptPath) : require(initScriptPath)
+
+    StartupTime.addMarker('window:initialize:start')
+
     return initialize({blobStore: blobStore}).then(function () {
+      StartupTime.addMarker('window:initialize:end')
       electron.ipcRenderer.send('window-command', 'window:loaded')
     })
   }
--- a/script/lib/generate-startup-snapshot.js
+++ b/script/lib/generate-startup-snapshot.js
@@ -69,7 +69,10 @@ module.exports = function (packagedAppPath) {
         requiredModuleRelativePath === path.join('..', 'node_modules', 'tree-sitter', 'index.js') ||
         requiredModuleRelativePath === path.join('..', 'node_modules', 'yauzl', 'index.js') ||
         requiredModuleRelativePath === path.join('..', 'node_modules', 'winreg', 'lib', 'registry.js') ||
-        requiredModuleRelativePath === path.join('..', 'node_modules', '@atom', 'fuzzy-native', 'lib', 'main.js')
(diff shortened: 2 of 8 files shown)


rafeca:
For now, I'm keeping the code to log the `shell` load time and `core` load time, so we can keep detecting potential regressions until the new metrics get some history.

rafeca:
This has made me realize that the timings that we're currently logging don't give us a very accurate image of the real times that it takes to load Atom.

In the startup example posted on the issue description, I had to wait for ~3-4s from when I launched Atom until some content was display in the Atom window.

This aligns with the new marker measures, but the old logged measures were ~500ms for the shell and 500ms for the window, which is way smaller than the reality.

static/index.js
```diff
@@ -3,16 +3,29 @@
...
   const electron = require('electron')
   const path = require('path')
   const Module = require('module')
   const getWindowLoadSettings = require('../src/get-window-load-settings')
+  const StartupTime = require('../src/startup-time')
   const entryPointDirPath = __dirname
   let blobStore = null
   let useSnapshot = false
 
+  const startupMarkers = electron.remote.getCurrentWindow().startupMarkers
```
nathansobo: Have you profiled the overhead of this? Under the hood it does a synchronous RPC and might cost us a few milliseconds. On the other hand I don't have a great idea for an alternative approach other than finding some async approach to perform the RPC.

nathansobo (commented):
This seems really good. I'm a bit worried about the synchronous RPC to the main process to import the startup times into the window. Might be good to just check on its cost.

Also, I wonder how we could include the time to start a secondary Electron when launching an Atom window from the command line when Atom is already running. Could we send timing data from *that* Electron as well? (This could be in another PR, just bringing it up while we're discussing this stuff)

static/index.js
rafeca: Good point! Just measured it with `performance.now()` and that call takes 3.2-3.5ms (the time is quite consistent on my machine).

If we consider that this is too much, we can always add some sampling to e.g only log the markers data on 1/10 of the starts.

rafeca:
> Also, I wonder how we could include the time to start a secondary Electron when launching an Atom window from the command line when Atom is already running. Could we send timing data from that Electron as well? (This could be in another PR, just bringing it up while we're discussing this stuff)

Yup, we should definitely do that! (in a different PR though 😄 )

static/index.js
nathansobo: That feels like an unfortunate amount of time. Is there any way we can fetch this asynchronously?

rafeca: I'm not sure, we're following the same pattern in a few other places (like getting [all the loadSettings](https://github.com/atom/atom/blob/master/src/get-window-load-settings.js#L7)) so I thought it was ok.

I'm gonna check if I can speed this up somehow... if not we can just sample it

rafeca: Ok I took more measures and compared the time this call takes to [the call to load the settings](all the loadSettings](https://github.com/atom/atom/blob/master/src/get-window-load-settings.js#L7), which happens a few lines below this, and turns out that these 3ms may include some bootstrap time for the IPC or something... let me show you the timings:

**With this call**:

- `getTimingSettings`: 3.5ms
- `getWindowLoadSettings`: 1.5ms

**Removing this call**:

- `getWindowLoadSettings`: 4ms

So basically, the net time that's added by this call is ~1ms

nathansobo: Okay, I think it's reasonable to just proceed for now. Measuring this will allow us to attack other aspects and we'll probably get the time back and then some.

nathansobo (approved):

PR #19315, opened 2019-05-12
Add protip about global flag when scanning for multiple matches

oggy:
### Description of the Change

I spent a while scratching my head wondering why `scanInBufferRange` was stopping after the first match. I don't think the global flag requirement is obvious from the existing documentation.

### Release Notes

N/A

### DIFF
--- a/src/text-editor.js
+++ b/src/text-editor.js
@@ -3523,7 +3523,8 @@ class TextEditor {
   }
 
   // Essential: Scan regular expression matches in a given range, calling the given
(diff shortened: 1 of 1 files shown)


rsese:
Thanks for the contribution @oggy!  We reviewed your suggestion and while we agree it could be helpful, we also think that it’s more a general reminder about regular expressions rather than something specific to these functions so we’ve decided to leave it as is.

But thank you again for your interest and contribution :bow:

PR #15431, opened 2017-08-24
Upgrade to Electron Packager 11

malept:
### Description of the Change

Hi there, I'm the maintainer of Electron Packager. I just released [version 9.0.0](https://github.com/electron-userland/electron-packager/releases/tag/v9.0.0), and thought that I could help upgrade a bunch of the projects that depend on it.

### Alternate Designs

N/A

### Why Should This Be In Core?

This already packages Atom :smile:

### Benefits

There's so many more features in 11 than 7.3.0, but the most notable ones are that version 8 added armv7l support, 9 added arm64 support, and 11 added mips64el support, so whenever you're ready to start officially supporting those platforms, Packager's ready to help!

### Possible Drawbacks

As with any upgrade of a module, there might be some bugs that crept in, but hopefully this whole exercise will help find them.

### Applicable Issues

None that I'm aware of.

PR #19318, opened 2019-05-13
⬆️ electron-packager@12.2.0

rafeca:
This PR updates the version of `electron-packager` used by Atom to almost the latest one (latest one is 13.0.1 but I thought it would be sensible to use the previous major, which has many more minor and patch releases under its belt).

The main reason for the update is to prepare for a future Electron v4 upgrade which will require https://github.com/electron-userland/electron-packager/pull/894 to be able to work (I found that [while exploring to upgrade](https://github.com/atom/atom/tree/electron-4) to Electron v4 directly).

Since the new version of `electron-packager` should be compatible with both electron v2 and v3, better to upgrade right now (and find potential issues now) than when doing the electron upgrade.

The list of changes between v7.3.0 and v12.2.0 is huge (v7.3.0 is a 3 year old version), so the easier way to get a glimpse of the big changes is to look [at its releases page](https://github.com/electron-userland/electron-packager/releases).

----

*List of changes between `electron-packager@7.3.0` and `electron-packager@12.2.0`: https://github.com/electron-userland/electron-packager/compare/v7.3.0...v12.2.0*

### DIFF
--- a/script/lib/package-application.js
+++ b/script/lib/package-application.js
@@ -15,24 +15,24 @@ module.exports = function () {
   const appName = getAppName()
   console.log(`Running electron-packager on ${CONFIG.intermediateAppPath} with app name "${appName}"`)
   return runPackager({
-    'app-bundle-id': 'com.github.atom',
-    'app-copyright': `Copyright © 2014-${(new Date()).getFullYear()} GitHub, Inc. All rights reserved.`,
-    'app-version': CONFIG.appMetadata.version,
+    'appBundleId': 'com.github.atom',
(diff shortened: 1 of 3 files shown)


50Wliu:
fyi there were a few more changes made in the PR to upgrade to 11.x by the author of electron-packager: #15431
Not sure if they're necessary or not.

malept:
They're not *necessary*, but they do clean up the packaging code a lot.

rafeca:
Oh thanks for the heads up @50Wliu ! Didn't know that that PR existed! I'm gonna cherry pick the relevant changes from there

kadirselcuk:
Selam atom kullanımı için electron ortamına ihtiyacımız var

Issue #19320, opened 2019-05-13
Atom 1.39.0-nightly1 fails to open


jasonrudolph:
### Description

With the latest Atom Nightly build (1.39.0-nightly1), Atom fails to open

### Steps to Reproduce

1. Attempt to launch Atom from the command line:
    `atom-nightly --safe --foreground`

**Expected behavior:** Atom opens without any errors

**Actual behavior:** Command line shows errors and no Atom window is shown:

```
2019-05-13 09:42:45.823 Atom Nightly[67481:723155] Arguments to CompositeDisposable.add must have a .dispose() method
2019-05-13 09:42:45.824 Atom Nightly[67481:723155] TypeError: Arguments to CompositeDisposable.add must have a .dispose() method
    at assertDisposable (/Applications/Atom Nightly.app/Contents/Resources/app.asar/node_modules/event-kit/dist/composite-disposable.js:162:11)
    at CompositeDisposable.add (/Applications/Atom Nightly.app/Contents/Resources/app.asar/node_modules/event-kit/dist/composite-disposable.js:106:13)
    at AtomApplication.launch (/Applications/Atom Nightly.app/Contents/Resources/app.asar/src/main-process/atom-application.js:281:25)
    at <anonymous>:null:null
```

**Reproduces how often:** 100%

### Versions

macOS 10.14.4

```
$ atom-nightly --version
Atom    : 1.39.0-nightly1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

nathansobo:
Closed via https://github.com/atom/atom/pull/19321.

PR #19321, opened 2019-05-13
Always return a Disposable from the watch() method

rafeca:
When there's an error watching the config file (as it happened [here](https://github.com/atom/atom/issues/19320)), we're currently swallowing the actual error and returning `undefined` from the `watch` method. This basically produces a different error in `atom-application` since that return value is not a disposable.

In order to fix that while maintaining what I think is the original goal of the code, this PR adds a `return` statement to return an empty disposable. This way the error gets propagated correctly to `atom-application` and it actually does not prevent Atom to get opened.

Note that this does not fix the original error reported in https://github.com/atom/atom/issues/19320: it just makes it easier to debug.

## Alternate designs

Instead of returning an empty `Disposable`, we can just remove the `catch` block and completely fail if there's an error watching the config file. I haven't decided to do so to maintain as much as possible the intention of the original code (which was to not fail under that circumstance), but I can change that if somebody has strong opinions on it.

### DIFF
--- a/src/config-file.js
+++ b/src/config-file.js
@@ -1,7 +1,7 @@
 const _ = require('underscore-plus')
 const fs = require('fs-plus')
 const dedent = require('dedent')
-const {Emitter} = require('event-kit')
+const {Disposable, Emitter} = require('event-kit')
 const {watchPath} = require('./path-watcher')
 const CSON = require('season')
 const Path = require('path')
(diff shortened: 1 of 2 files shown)


nathansobo (approved):
Argh. First lurking dragon of the notify changes has been uncovered. Thanks for addressing this. I'll jump on the other issue.

PR #19324, opened 2019-05-13
Integrate packager improvements from https://github.com/atom/atom/pull/15431

rafeca:
This PR is a follow-up of https://github.com/atom/atom/pull/19318 to integrate the changes made by @malept on https://github.com/atom/atom/pull/15431


script/lib/package-application.js
```diff
@@ -4,38 +4,42 @@ const assert = require('assert')
...
+    dir: CONFIG.intermediateAppPath,
+    electronVersion: CONFIG.appMetadata.electronVersion,
+    extendInfo: path.join(CONFIG.repositoryRootPath, 'resources', 'mac', 'atom-Info.plist'),
+    helperBundleId: 'com.github.atom.helper',
+    icon: path.join(CONFIG.repositoryRootPath, 'resources', 'app-icons', CONFIG.channel, 'atom'),
+    name: appName,
+    out: CONFIG.buildOutputPath,
+    overwrite: true,
+    platform: process.platform,
+    // Atom doesn't have devDependencies, but if prune is true, it will delete the non-standard packageDependencies.
```
rafeca: That's actually really cool! @malept does this mean that if we enable this option we can start using `devDependencies` on the top-level `package.json` without adding them to the app package? 😍 

/cc @nathansobo , @smashwilson , @jasonrudolph

malept: As I mentioned in the comment, `packageDependencies` is still problematic (i.e., gets pruned) if `prune: true` is set. If Packager needs to deal with this use case, we'd have to do some refactoring of how pruning works (that is, add support for custom `*Dependencies` in `package.json` to be whitelisted).

rafeca: Oh, I understood the comment incorrectly, thanks for clarifying 👍

PR #19336, opened 2019-05-14
Allow escaped quote literals in the SyntaxScopeMap

nathansobo:
As raised by @Ben3eeE in https://github.com/atom/language-json/pull/68#issuecomment-492440942, we are currently unable to apply decorative syntax scopes to `"` literals for TreeSitter grammars, which is forcing us to break user expectations around how syntax themes apply to strings.

In this PR, I add some special handling for escaped `"` literals to the `SyntaxScopeMap`. Internally, the `SyntaxScopeMap` uses the `postcss-selector-parser` library to parse any scope it is provided. This library allows `"` literals inside of selectors, but they need to be *escaped* in the selector. So a valid post CSS selector looks like this:

```
string > "\""
```

Unfortunately, these selectors are themselves expressed as strings within CSON or JSON files in our grammars. That means the escaping gets a bit intense, and looks like this:

```CSON
'string > "\\""': 'puncuation.quote.double.whatever'
```

Notice that I include **2** backslashes when escaping the quote so that they end up getting passed through to the parser as a single `\`. If I only include one backslash it ends up looking like an escaped quote at the interpretation level of CSON rather than the level of the selector language. 😓

Anyway, once properly escaped, the quote literal is handled by the parser, but the value it spits out continues to include the escaping backslash. In order to match the text of the anonymous nodes returned by TreeSitter, we need to remove the escaping after trimming off the outer quotes.

@Ben3eeE hopefully this gives you what you need to style quotes properly in various grammars.

cc @50Wliu @maxbrunsfeld

### DIFF
--- a/spec/syntax-scope-map-spec.js
+++ b/spec/syntax-scope-map-spec.js
@@ -49,13 +49,15 @@ describe('SyntaxScopeMap', () => {
     const map = new SyntaxScopeMap({
       '"b"': 'w',
       'a > "b"': 'x',
-      'a > "b":nth-child(1)': 'y'
+      'a > "b":nth-child(1)': 'y',
+      '"\\""': 'z'
     })
 
     expect(map.get(['b'], [0], true)).toBe(undefined)
     expect(map.get(['b'], [0], false)).toBe('w')
     expect(map.get(['a', 'b'], [0, 0], false)).toBe('x')
     expect(map.get(['a', 'b'], [0, 1], false)).toBe('y')
+    expect(map.get(['a', '"'], [0, 1], false)).toBe('z')
   })
 
   it('supports the wildcard selector', () => {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (approved):
👍 Makes sense!

It's not surprising that there are some gotchas with the CSS-within-CSON approach I went with here  😬. Glad to see you were able to solve this issue though.

As an aside - We've since developed a system where you write this stuff in [regular css files](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/properties/highlights.css), and the CSS files get compiled ahead-of-time into [JSON state machines](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/src/highlights.json) that let you match the properties even more efficiently than the `SyntaxScopeMap`. I don't know if it would be worth porting this to work in Atom though. It'd be a fun project, but it'd mean another API change.

nathansobo:
That CSS syntax is pretty sweet. Curious this and the state machine, but it's probably low on the list of priorities right now for integration into Atom.

nathansobo:
Flaky test in GitHub package that's definitely unrelated. Merging.

PR #19337, opened 2019-05-15
Enable installing apm using the ci command

winstliu:
There's been a few dependency changes (notably: the dropping of node-gyp), so let's give this another try. Seems to work on my machine™.


50Wliu:
That didn't work! All right then. I'll try again after atom/apm#845.

50Wliu:
Actually...let me try ci _with_ global-style.

PR #19338, opened 2019-05-15
pane.js: Additional destroy arguments of `force`

robobenklein:

Such a small change I haven't created an issue yet.

### Description of the Change

Adds the option `force` to the other Pane Item destruction functions.

### Alternate Designs

Originally I had to either extend `Pane` and override `destroyItems`/`destroyInactiveItems`/etc or create a custom function to force-destroy items in Panes.

### Possible Drawbacks

Hopefully none, this should purely add options for other plugin developers.

### Verification Process

Ran Atom locally and tested each function with and without `force` option supplied, all seemed to work.

### Release Notes

Change is not user-facing: `N/A`.

### DIFF
--- a/src/pane.js
+++ b/src/pane.js
@@ -736,9 +736,11 @@ class Pane {
 
   // Public: Destroy the active item and activate the next item.
   //
+  // * `force` (optional) {Boolean} Same behavior as destroyItem(item, force)
+  //
   // Returns a {Promise} that resolves when the item is destroyed.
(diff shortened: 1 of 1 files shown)


lee-dohm:
After reviewing this, we're not sure that we want an API that allows for destroying all `PaneItems` without prompting to save them. Please open an issue to describe why this needs to be an official, supported API.

Additionally, any pull request requires tests to be reviewed and accepted.

PR #19354, opened 2019-05-17
Wait for config file to load before prompting to restart in config.onDidChange callback

### LINKED ISSUE #19323: Prompts for restart when certain core settings aren't default
### Prerequisites

### Description

If one of the core setting that are configured to prompt for a restart are changed from their default value then Atom will prompt for a restart every time it starts up.

### Steps to Reproduce

1. Change a core setting from [the following list](https://github.com/atom/atom/blob/b79d9088e21970f0dd1d05ed0e2a27765d4b6dd4/src/main-process/atom-application.js#L283-L285) to a non-default value:
    * `core.titleBar`
    * `core.colorProfile`
    * `core.fileSystemWatcher`
2. Restart Atom

No prompt to restart Atom on launch.

Prompts to restart Atom on _every_ launch, regardless of how many times you restart it.

100% of the time.

### Versions

OS: Windows 10 x64
```
Atom    : 1.39.0-nightly2
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
apm  2.1.7
npm  6.2.0
node 8.9.3 x64
atom 1.39.0-nightly2
python 2.7.15
git 2.19.1.windows.1
visual studio 2015
```

### Additional Information

N/A

nathansobo:
Fixes #19323.

In #19246, we stopped awaiting the watching of the config file on Windows for performance reasons. However, directly after the `await` of the promise returned by `watch`, we call `this.config.onDidChange` with the assumption that it has already been loaded. This PR moves those observations into a `then` callback chained onto the original promise so that we only observe changes once the file is loaded, regardless of whether we call `await`. It also adds the returned disposable in the `then` callback on all platforms just to unify the logic as much as possible.

cc @Arcanemagus

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -271,17 +271,17 @@ class AtomApplication extends EventEmitter {
 
   async launch (options) {
     if (!this.configFilePromise) {
-      this.configFilePromise = this.configFile.watch()
+      this.configFilePromise = this.configFile.watch().then(disposable => {
+        this.disposable.add(disposable)
+        this.config.onDidChange('core.titleBar', () => this.promptForRestart())
+        this.config.onDidChange('core.colorProfile', () => this.promptForRestart())
+      })
 
       // TodoElectronIssue: In electron v2 awaiting the watcher causes some delay
       // in Windows machines, which affects directly the startup time.
-      if (process.platform === 'win32') {
-        this.configFilePromise.then(disposable => this.disposable.add(disposable))
-      } else {
-        this.disposable.add(await this.configFilePromise)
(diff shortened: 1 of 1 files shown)


rafeca:
Thanks for fixing this @nathansobo and apologies for introducing this issue 🤡

rafeca (approved):

nathansobo:
> apologies for introducing this issue

No worries. I pinged you just to promote awareness but definitely not to lay blame. Easy mistake to make and a small regression.

nathansobo:
Cherry-picked the merge commit to `1.38-releases`, so we'll need to issue a release at some point.

PR #19359, opened 2019-05-17
🚱 Fix leak of LanguageMode objects in tests

### LINKED ISSUE #18991: Flaky build: Windows build sometimes fails with "resource busy or locked"

nathansobo:
Fixes #18991

In #18991, we were experiencing render process crashes caused by the render process running out of memory. Upon investigating, we discovered that the memory used by the `Atom Helper` process rapidly increased during `text-editor-spec.js`, from a couple hundred megabytes to 3+ gigabytes by the end of those tests.

We were confused, because a heap snapshot of the process indicated a normal heap size of just a couple hundred megabytes. Upon further investigation, I discovered that we were leaking `TextMateLanguageMode` objects. These language modes reference TextMate grammars, which themselves reference a large number of Oniguruma regular expression objects, which are implemented in a native C++ and therefore not tracked as part of the JS heap. That explains the discrepancy between the process memory consumed and what was being reported in the Chrome dev tools.

The cause of the leak was subtle. In `TextEditorRegistry`, we were assigning a promise to an instance field named `initialPackageActivationPromise`. This would resolve when the `PackageManager` emitted an `onDidActivateInitialPackages` event. We then `await`ed this promise in `updateAndMonitorEditorSettings`, and this ended up implicitly capturing the `oldLanguageMode` variable that we interacted with after resuming from the `await`. Unfortunately, in many tests, we never emit the `onDidActivateInitialPackages` event at all, so this promise was hanging around unresolved. This was causing the `oldLanguageMode` to be kept in scope by the `initialPackageActivationPromise` we were `await`ing. Not sure how clear that was, but hopefully it helps explain the diff to a careful reader.

My solution is to expose a `getActivatePromise` method on the `PackageManager` itself. We store off the promise during activation and then null it out when we're done. Most importantly, we *also* null out this promise when we call `PackageManager.prototype.reset`, so that any objects that are implicitly captured by code `await`ing this promise get garbage collected.

As a result, we end `text-editor-spec.js` with 6 `TextMateLanguageMode` objects on the heap instead of 954, and the memory consumed by the process is kept more reasonable (though Jasmine still seems to be leakier than I'd prefer).

### Before:

Screen Shot 2019-05-17 at 4 32 37 PM

### After:

Screen Shot 2019-05-17 at 4 39 15 PM

### DIFF
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -90,6 +90,7 @@ module.exports = class PackageManager {
     this.packagesCache = packageJSON._atomPackages != null ? packageJSON._atomPackages : {}
     this.packageDependencies = packageJSON.packageDependencies != null ? packageJSON.packageDependencies : {}
     this.triggeredActivationHooks.clear()
+    this.activatePromise = null
   }
 
   /*
@@ -112,6 +113,14 @@ module.exports = class PackageManager {
   // Returns a {Disposable} on which `.dispose()` can be called to unsubscribe.
   onDidActivateInitialPackages (callback) {
     return this.emitter.on('did-activate-initial-packages', callback)
+  }
+
+  getActivatePromise() {
+    if (this.activatePromise) {
+      return this.activatePromise
+    } else {
+      return Promise.resolve()
+    }
   }
 
   // Public: Invoke the given callback when a package is activated.
@@ -659,10 +668,11 @@ module.exports = class PackageManager {
       const packages = this.getLoadedPackagesForTypes(types)
       promises = promises.concat(activator.activatePackages(packages))
     }
-    return Promise.all(promises).then(() => {
+    this.activatePromise = Promise.all(promises).then(() => {
       this.triggerDeferredActivationHooks()
       this.initialPackagesActivated = true
       this.emitter.emit('did-activate-initial-packages')
+      this.activatePromise = null
     })
   }
 
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -674,6 +674,7 @@ module.exports = class PackageManager {
       this.emitter.emit('did-activate-initial-packages')
       this.activatePromise = null
     })
+    return this.activatePromise
   }
 
   registerURIHandlerForPackage (packageName, handler) {
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -14,7 +14,11 @@ describe('TextEditorRegistry', function () {
       assert: atom.assert,
       config: atom.config,
       grammarRegistry: atom.grammars,
-      packageManager: { deferredActivationHooks: null }
+      packageManager: {
+        getActivatePromise() {
+          return initialPackageActivation
+        }
+      }
     })
 
     editor = new TextEditor({ autoHeight: false })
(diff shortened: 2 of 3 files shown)


src/package-manager.js
```diff
@@ -674,6 +674,7 @@ module.exports = class PackageManager {
       this.emitter.emit('did-activate-initial-packages')
       this.activatePromise = null
     })
+    return this.activatePromise
```
nathansobo: Oops, thanks.

PR #19403, opened 2019-05-24
Fix handling of binary files when using ripgrep scanner

rafeca:
This PR is a follow-up of https://github.com/atom/atom/pull/19348

By pure chance I've found an issue when handling files that have lines with binary content while testing the `ripgrep` scanner on the [`gecko-dev`](https://github.com/mozilla/gecko-dev) repository.

The main problem is found when a file has lines with non-UTF characters it returns a base64 representation of that line in the json output (to not break the JSON formatting). There's more information about it [here](https://docs.rs/grep-printer/0.1.1/grep_printer/struct.JSON.html#text-encoding).

Before this PR, we were not handling this scenario and the parser miserably failed 😅

### DIFF
--- a/src/ripgrep-directory-searcher.js
+++ b/src/ripgrep-directory-searcher.js
@@ -155,11 +155,7 @@ function processSubmatch (submatch, lineText, offsetRow) {
 }
 
 function getText (input) {
-  if (input.text) {
-    return input.text
-  }
-
-  return Buffer.from(input.bytes, 'base64').toString()
+  return input.text ? input.text : Buffer.from(input.bytes, 'base64').toString()
 }
 
 module.exports = class RipgrepDirectorySearcher {
--- a/src/ripgrep-directory-searcher.js
+++ b/src/ripgrep-directory-searcher.js
@@ -50,7 +50,7 @@ function updateLeadingContext (message, pendingLeadingContext, options) {
   }
 
   if (options.leadingContextLineCount) {
-    pendingLeadingContext.push(cleanResultLine(message.data.lines.text))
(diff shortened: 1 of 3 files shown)


src/ripgrep-directory-searcher.js
```diff
@@ -146,8 +150,16 @@ function processSubmatch (submatch, lineText, offsetRow) {
...
+    lineText: cleanResultLine({ text: lineParts.join('\n') })
+  }
+}
+
+function getText (input) {
+  if (input.text) {
+    return input.text
   }
+
+  return Buffer.from(input.bytes, 'base64').toString()
```
nathansobo: Minor: Putting this in an `else` block would be more readable to me.

src/ripgrep-directory-searcher.js
```diff
@@ -280,7 +292,7 @@ module.exports = class RipgrepDirectorySearcher {
 
           if (message.type === 'begin') {
             pendingEvent = {
-              filePath: message.data.path.text,
+              filePath: getText(message.data.path),
```
nathansobo: Wow, base64 encoded paths? Awesome.

nathansobo (approved):
Nice find! ⚡️

nathansobo:
In case anyone is wondering whether there's anything possible in these JSON objects other than `text` and `bytes`, [there isn't](https://github.com/BurntSushi/ripgrep/blob/master/grep-printer/src/jsont.rs#L82).

src/ripgrep-directory-searcher.js
rafeca: I've just used a ternary at the end 😁

PR #19405, opened 2019-05-24
Upgrade atom-keymap to incorporate changes to keyboard-layout

nathansobo:
In https://github.com/atom/keyboard-layout/pull/51, @MarshallOfSound made some behavior-preserving changes to make the `keyboard-layout` module context-aware. I wanted to get that PR merged for him, and I also want to keep Atom up-to-date. This PR is to ensure we build green against the updated dependency. Once we get a green build I will publish non prerelease versions of `atom-keymap` and `keyboard-layout`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1709,16 +1709,16 @@
       }
     },
     "atom-keymap": {
-      "version": "8.2.13-0",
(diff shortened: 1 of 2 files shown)


nathansobo:
The latest commit is identical to the previous green build, just not depending on a pre-release. I wanted to see dependencies install cleanly, but there's no reason this should fail. Going to merge.

PR #19408, opened 2019-05-27
Enable eslint rules that got disabled

rafeca:
On https://github.com/atom/atom/pull/19302, a few `eslint` rules were disabled to keep that PR small and manageable, now it's the time to enable a few of them.

There are two rules that I haven't enabled:

* `standard/no-callback-literal`: We have quite a few Atom public APIs that expect a callback with the result on the first argument, so to enable this rule we would have to either add a bunch of `ignore` comments or to create a quite big breaking change.
* `node/no-deprecated-api`: This is a quite useful rule, but I haven't been able to enable it because it complains about using `url.parse()`. Changing this to the new `URL()` constructor would be a breaking change, since we're passing `url` objects to packages.

We can follow-up what to do for these two rules in separate PRs, to keep the discussion focused.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,7 +25,6 @@
   "rules": {
     "standard/no-callback-literal": ["off"],
     "node/no-deprecated-api": ["off"],
-    "no-use-before-define": ["off"],
     "prettier/prettier": ["off"] // disable prettier rules for now.
   },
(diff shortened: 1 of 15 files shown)


as-cii (approved):
Looks reasonable! ✨

PR #19373, opened 2019-05-20
Upgrade to Electron 4

nathansobo:
This PR will track our progress upgrading to Electron 4.

We just upgraded to Electron 3, but we're experiencing a crash deep in the bowels of V8 (https://github.com/atom/atom/issues/19372). I'm interested in whether Electron 4 might fix this crash.

## Issues to resolve

- [x] https://github.com/atom/atom/issues/19374 Crash in mksnapshot when building with Electron 4
- [x] Update any native modules Atom is using to make sure `win_delay_load_hook` isn't false: https://electronjs.org/docs/tutorial/using-native-node-modules#a-note-about-win_delay_load_hook
  - [x] @atom/nsfw: atom/nsfw#6
  - [x] git-utils: atom/git-utils#94
  - [x] scandal (git-utils version bump): atom/scandal#47
- [x] Failing integration test
- [x] Native tabs super slow (https://github.com/atom/atom/issues/19611).

PR #19418, opened 2019-05-29
Don't create a promise when preloading package settings in snapshot

### LINKED ISSUE #19374: Crash in mksnapshot when building with Electron 4
When attempting to upgrade to Electron 4 in https://github.com/atom/atom/pull/19373, we fail the build with the following error when attempting to generate the startup snapshot. (I modified the build script slightly to pass `stdio: "inherit"` to `mksnapshot` for better error output)

```
Minifying startup script
Verifying if snapshot can be executed via `mksnapshot`
Generating startup blob with mksnapshot
Loading script for embedding: /Users/nathan/src/atom/out/startup.js
Error running mksnapshot.
Moving generated startup blob into "/Users/nathan/src/atom/out/Atom Dev.app/Contents/Frameworks/Electron Framework.framework/Resources/v8_context_snapshot.bin"
Error: ENOENT: no such file or directory, rename '/Users/nathan/src/atom/out/v8_context_snapshot.bin' -> '/Users/nathan/src/atom/out/Atom Dev.app/Contents/Frameworks/Electron Framework.framework/Resources/v8_context_snapshot.bin'
  at Object.fs.renameSync (fs.js:766:18)
  at electronLink.then (/Users/nathan/src/atom/script/lib/generate-startup-snapshot.js:137:10)
  at <anonymous>:null:null
```

After hacking some logging into `electron-mksnapshot`, it appears that the `mksnapshot` is crashing with a segmentation fault. In the Node wrapper script, the exit code on the child process is `null` and the `signal` is `SEGV`.

Running `mksnapshot` *directly* as follows also produces a segfault, although `electron-mksnaphsot` seems to do some special magic to move files to a single working directory, so it's unclear whether this is even a good way to run `mksnapshot`.

```
Nathans-MacBook-Pro:atom nathan$ script/node_modules/electron-mksnapshot/bin/mksnapshot --startup_blob snapshot_blob.bin --turbo_instruction_scheduling /Users/nathan/src/atom/out/startup.js
Loading script for embedding: /Users/nathan/src/atom/out/startup.js
Segmentation fault: 11
Nathans-MacBook-Pro:atom nathan$ echo $?
139
```

## Trying the *latest* electron-mksnapshot

I tried to generate the startup snapshot with `electron-mksnapshot@6.0.0-beta.1` and got a bit further. I *was* able to generate `snapshot_blob.bin`, but the script failed when invoking `v8_context_snapshot_generator`. The output in this failure mode is pretty bad... the script just prints out the entire child process object, so I'm eliding some of its details here for readability.

```
Nathans-MacBook-Pro:atom nathan$ cat /Users/nathan/.nvm/versions/node/v8.15.1/lib/node_modules/electron-mksnapshot/package.json | grep "\"version\":"
  "version": "6.0.0-beta.1"
Nathans-MacBook-Pro:atom nathan$ /Users/nathan/.nvm/versions/node/v8.15.1/lib/node_modules/electron-mksnapshot/mksnapshot.js /Users/nathan/src/atom/out/startup.js
Loading script for embedding: /Users/nathan/src/atom/out/startup.js
Error running the v8 context snapshot generator. { status: null,
  signal: 'SIGTRAP',
  output: [ null, null, null ],
  pid: 94008,
  stdout: null,
  stderr: null,
  envPairs: ELIDED
  options:        { cwd: '/Users/nathan/.nvm/versions/node/v8.15.1/lib/node_modules/electron-mksnapshot/bin',
     env: ELIDED
     args: [ '/var/folders/wy/zs1k37fn0d95lsfcv2d3z6m80000gn/T/mksnapshot-workdir119420-94006-vcutmj.h9pos/v8_context_snapshot_generator',
     '--output_file=/Users/nathan/src/atom/v8_context_snapshot.bin' ],
  file: '/var/folders/wy/zs1k37fn0d95lsfcv2d3z6m80000gn/T/mksnapshot-workdir119420-94006-vcutmj.h9pos/v8_context_snapshot_generator' }
```

I'm not sure where to go from here. To some degree, we're abusing V8 startup snapshots in Atom. They weren't designed for public consumption, and the snapshot that ships with Chrome is much smaller and probably way simpler than what we are attempting in Atom's `startup.js` script.

cc @MarshallOfSound

as-cii:
Fixes #19374 

Promise creation is forbidden within `mksnapshot` starting from Electron 3 (see https://github.com/electron/libchromiumcontent/pull/363, https://github.com/nodejs/node/pull/13242 and https://github.com/electron/electron/issues/18420). Since preloading a package's settings is a synchronous action anyway, we just avoid instantiating a new Promise when calling `loadSettings`.

This fixes the crashes we were observing in #19373 when generating a new snapshot, and it is a prerequisite for shipping the Electron 3 upgrade after a hotfix with https://github.com/electron/electron/pull/18426 is released.

### DIFF
--- a/src/package.js
+++ b/src/package.js
@@ -648,16 +648,15 @@ class Package {
       })
     }
 
-    return new Promise(resolve => {
-      if (this.preloadedPackage && this.packageManager.packagesCache[this.name]) {
-        for (let settingsPath in this.packageManager.packagesCache[this.name].settings) {
-          const properties = this.packageManager.packagesCache[this.name].settings[settingsPath]
-          const settingsFile = new SettingsFile(`core:${settingsPath}`, properties || {})
-          this.settings.push(settingsFile)
-          if (this.settingsActivated) settingsFile.activate(this.config)
-        }
-        return resolve()
-      } else {
+    if (this.preloadedPackage && this.packageManager.packagesCache[this.name]) {
+      for (let settingsPath in this.packageManager.packagesCache[this.name].settings) {
+        const properties = this.packageManager.packagesCache[this.name].settings[settingsPath]
+        const settingsFile = new SettingsFile(`core:${settingsPath}`, properties || {})
+        this.settings.push(settingsFile)
+        if (this.settingsActivated) settingsFile.activate(this.config)
+      }
+    } else {
+      return new Promise(resolve => {
         const settingsDirPath = path.join(this.path, 'settings')
         fs.exists(settingsDirPath, (settingsDirExists) => {
           if (!settingsDirExists) return resolve()
@@ -666,8 +665,8 @@ class Package {
             async.each(settingsPaths, loadSettingsFile, () => resolve())
           })
         })
-      }
-    })
+      })
+    }
   }
 
   serialize () {


rafeca (approved):
😍

as-cii:
I think the failures are unrelated to this pull request, so I'll go ahead and merge.

PR #19423, opened 2019-05-31
Display multi-keystroke key bindings in menu item's label

### LINKED ISSUE #5343: View > Panes doesn't show keybindings in menu
_(Issue version of https://github.com/atom/atom/issues/2594#issuecomment-56977044)_

**Expected:** When I click View > Panes and look at the available Panes operations, I should see their keybindings.

**Actual:** When I click View > Panes, I see no keybindings even though the Panes operations have keybindings by default.

**Steps to reproduce:**
1. Click on View in the main menu
2. Move to Panes
3. None of the Panes operations show their keybindings

image

However, you can confirm that Panes operations do in fact have keybindings by checking the Keybindings list or by opening the Command Palette and searching for `pane`.

image

image

**Atom:** 0.176.0 (safe mode)
**OS:** OS X 10.10.2 Yosemite

---

**EDIT:** Reproduced in Windows 8 too:
image

as-cii:
Closes #5343 

This pull request introduces a workaround to a limitation in Electron's [accelerator system](https://electronjs.org/docs/api/accelerator), which refutes to display shortcuts for a menu item when it is a multi-keystroke key binding. 

I performed a brief investigation of what it would take to fix this upstream in Electron, but unfortunately the shortcut code lives within `libchromiumcontent` and it would be a major undertaking to replace Chromium's system with a custom one.

Instead, with these changes, multi-keystroke keybindings will be displayed next to the item's label:

<p align=center>
Screen Shot 2019-05-31 at 15 10 57
</p>

This is clearly a suboptimal choice in terms of how these shortcuts will be rendered, but there is at least one precedent for it in Visual Studio Code (see https://github.com/Microsoft/vscode/issues/30617). Considering that it does not impact the UX in any way (shortcuts will continue to work as they did before), I think this is an acceptable workaround that will help users find which keystrokes a menu item maps to.


src/context-menu-manager.coffee
```diff
@@ -158,8 +159,16 @@ class ContextMenuManager
...
-        item.accelerator = accelerator if accelerator
+        keystrokes = keymaps?[0]?.keystrokes
+        if keystrokes
+          # Electron does not support multi-keystroke accelerators. Therefore,
+          # when the command maps to a multi-stroke key binding, show the
+          # keystrokes next to the item's label.
+          if keystrokes.includes(' ')
+            item.label += " [#{_.humanizeKeystroke(keystrokes)}]"
+          else
+            accelerator = MenuHelpers.acceleratorForKeystroke(keystrokes)
```
rafeca: nit: you don't need this temporary variable

rafeca (approved):
LGTM!

The solution is not ideal, but it's a nice workaround.

I could not find any related issue on Electron about this, should we create one and link to it on the comment?

Issue #19366, opened 2019-05-20
Investigate faster CI builds


as-cii:
As of today, testing out a change on CI approximately takes 50-60 minutes. Not only this introduces major friction in the feedback loop of the development process, but it is also source of slowdowns when a new version of Atom needs to be published. This is made worse by some flaky tests, which e.g. may cause us to wait multiple hours before publishing an important hotfix.

## Caching

Related: #19326 

After transitioning to Azure Pipelines as our CI provider, we lost an essential functionality for a project like Atom: caching. Our dependencies don't change that often, and when they do, only a small part of them gets updated/removed. [On any given build](https://dev.azure.com/github/Atom/_build/results?buildId=40844), we have to wait approximately 15-20 minutes (~ 30-50% of our build times) to setup dependencies and generate a snapshot script. 

<img width="1054" alt="Screen Shot 2019-05-20 at 10 02 21" src="https://user-images.githubusercontent.com/482957/58005755-af158280-7ae6-11e9-9e93-e87e4bcf9a7e.png">

Even though it seems like official support for caching hasn't yet been added to Azure Pipelines (apparently, it is [slated for 2019 Q2](https://dev.azure.com/mseng/AzureDevOpsRoadmap/_workitems/edit/1458319)), I discovered [an extension](https://github.com/microsoft/azure-pipelines-artifact-caching-tasks) (developed by Microsoft) that, if installed, claims to enable caching for arbitrary artifacts.

While it may not be fully baked yet, I would be inclined to try it out and see whether it makes a difference on this repository.

## Testing

Atom's test suite is quite thorough, and we even run tests against the latest bundled version of core packages to catch unwanted API changes. Having an extensive set of test is valuable, but every test contributes to a slower build.

<img width="1054" alt="Screen Shot 2019-05-20 at 10 02 21" src="https://user-images.githubusercontent.com/482957/58006339-239cf100-7ae8-11e9-9d47-ede770983501.png">

The above screenshot captures how long it takes to run main-process, renderer-process and package tests on macOS. On platforms like Windows x64, this time drops to 15 minutes because we simply don't run package tests there.

A simple solution to this could be to take advantage of the pipeline infrastructure offered by Azure. We could build Atom once on macOS and, in parallel, run two sub-jobs, each taking care of running main/renderer-process tests and package tests, respectively. This would dramatically decrease variance within an individual build, as macOS and Windows would roughly take the same amount of time to bootstrap and run tests.

[...]

PR #19437, opened 2019-06-03
Cache node_modules on Azure Pipelines

as-cii:
Refs: #19366 

This pull request's aim is to speed up our CI build times by taking advantage of https://github.com/microsoft/azure-pipelines-artifact-caching-tasks on this repository. 

Caches are looked up by a key based on the contents of all `package-lock.json` files in the repository. When (at least) one of them changes, the cache won't be used and dependencies will be re-bootstrapped from scratch. We could probably find ways of starting with a base set of dependencies (e.g., the ones found on `master`) and only perform a partial bootstrap, but @rafeca pointed out that this could lead to some non-determinism in the package installed by `npm`. Maybe this is one argument in favor of switching to `yarn` at some point.

For now, this pull request picks the low-hanging fruit of skipping `script/bootstrap` entirely if the lockfile perfectly matches the set of dependencies that we have in the cache. When that happens, it saves ~8-9 minutes of build time.

## Next Steps
In the future, we may want to investigate a couple of different options:

* Switch to `yarn`. Doing so should speed `script/bootstrap` up and, most importantly, it could remove the aforementioned non-determinism, which would allow for more aggressive caching.
* Fork https://github.com/microsoft/azure-pipelines-artifact-caching-tasks. This would allow us to introduce the notion of a _fallback cache_ to use when a perfect match for the lockfile could not be found. Obviously, this would come at the cost of maintainability and it would be interesting to see if the [planned Azure Pipelines caching](https://dev.azure.com/mseng/AzureDevOpsRoadmap/_workitems/edit/1458319) will offer such fallback mechanism out-of-the-box.
* Use caching for `script/build` too. That may still require using a fork of https://github.com/microsoft/azure-pipelines-artifact-caching-tasks, but we're spending a lot of time doing redundant work, such as generating the snapshot or transpiling files.


rafeca (approved):
🎉🎉🎉🎉🎉🎉🎉

PR #19444, opened 2019-06-04
Improvements to ripgrep scanner

rafeca:
This PR includes support for a few options that were ignored on the initial PR that added support for using `ripgrep` as the `scan()` backend.

List of options:

* `excludeVcsIgnores`: Whether or not to ignore files/folders that are ignored by `git`.
* `follow`: Whether or not to follow symlinks.
* `includeHidden`: Whether or not to search on hidden files (`workspace.scan()` hardcodes this to `true` though).

We did not catch this because there was no test checking them, so I've added tests for each of them to ensure that we don't add regressions in the future.

## Missing thing

I've noticed that the `scandal` scanner is able to search on a subfolder of a vcs ignored folder even when `excludeVcsIgnores` is enabled, so if we have (with `node_modules` being ignored by `git`):

```
node_modules/
  left-pad/
    index.js
```

1. If the user performs a normal search, neither `scandal` or `ripgrep` will search on the `node_modules/left-pad/index.js` file.
2. If the user performs a search on the `node_modules` folder, `scandal` and `ripgrep` will search on the `node_modules/left-pad/index.js` file.
3. If the user performs a search on the `node_modules/left-pad` folder, `scandal` will search on the `node_modules/left-pad/index.js` file but `ripgrep` won't.

I'll check how can we fix the scenario 3 (it may require bigger changes) and send a follow-up PR.

### DIFF
--- a/spec/workspace-spec.js
+++ b/spec/workspace-spec.js
@@ -2755,6 +2755,61 @@ describe('Workspace', () => {
           });
         });
 
+        describe('when the core.followSymlinks config is used', () => {
+          let projectPath;
+
+          beforeEach(async () => {
+            const sourceProjectPath = path.join(
+              __dirname,
+              'fixtures',
+              'dir',
+              'a-dir'
+            );
+            projectPath = path.join(temp.mkdirSync('atom'));
+
+            const writerStream = fstream.Writer(projectPath);
+            fstream.Reader(sourceProjectPath).pipe(writerStream);
+
+            await new Promise(resolve => {
+              writerStream.on('close', resolve);
+              writerStream.on('error', resolve);
+            });
+
+            fs.symlinkSync(
+              path.join(__dirname, 'fixtures', 'dir', 'b'),
+              path.join(projectPath, 'symlink')
+            );
+          });
+
+          afterEach(() => {
+            if (fs.existsSync(projectPath)) {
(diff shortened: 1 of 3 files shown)


nathansobo:
> If the user performs a search on the node_modules/left-pad folder, scandal will search on the node_modules/left-pad/index.js file but ripgrep won't.

This is probably related to the fact that we search within gitignored folders *any* time an explicit path is specified. That behavior is actually really annoying. The perfect fix would be to detect when the user was explicitly specifying a gitignored folder, but is probably more work than whatever Scandal is doing.

nathansobo (approved):
Nice work adding test coverage.

rafeca:
> This is probably related to the fact that we search within gitignored folders any time an explicit path is specified. That behavior is actually really annoying. The perfect fix would be to detect when the user was explicitly specifying a gitignored folder, but is probably more work than whatever Scandal is doing.

There were CI failures on this PR for Windows, caused by the test that I added for this specific behaviour.

After testing it locally, it seems that even the behaviour from `scandal` is inconsistent across platforms: while the search works in OSX, it does not seem to work on Windows, so I've removed the test and I'm not going to "fix" this at least for now in `ripgrep`.

rafeca:
Seems like current failures are due to flaky tests, so I'm gonna merge this one

PR #19446, opened 2019-06-04
Parallelize macOS tests

as-cii:
Refs: #19366 

This pull request changes our CI configuration to run tests in parallel on macOS. Specifically, it splits the macOS step into a **build phase** and a **test phase**. 

## Build Phase

During this phase, Azure Pipelines takes care of bootstrapping and building Atom. It then uploads the produced artifacts so that they can be used by the test phase described below.

Note that, perhaps counterintuitively, we will still execute main process tests as part of this phase. This is unfortunate, but there seems to be an issue when trying to execute main process tests on a container different from the one Atom was built on. These tests usually take less than 10 seconds anyways, so they should have a negligible cost on parallelism.

## Test Phase

During this phase, Azure Pipelines takes care of downloading the artifacts published by the build phase and use them to run tests. In particular, this phase is further split into 3 jobs that are executed in parallel:

- `macOS test core`, which runs core renderer tests.
- `macOS test packages-1`, which runs tests for the first 22 packages
- `macOS test packages-2`, which runs tests for the remaining packages

Splitting tests evenly across containers is non-trivial, especially since we don't know how long each test suite takes to execute. We could probably download that information from previous builds and schedule tests accordingly, but in this pull request I decided to go for a more low-tech approach. Empirically, I observed that the splitting strategy illustrated above achieves a good level of parallelism without introducing a ton of complexity.

## Conclusion

With these changes, build times when the `node_modules` cache is warm drop from ~1hr to ~40 minutes. In other words, after this pull request lands, we should be able to observe a consistent ~33% improvement in our build times. 🐎 

Please, note that there is still room for improvement, but it seems like additional enhancements to CI would come at a higher cost and we may want to work on them in a separate moment.

### DIFF
--- a/script/test
+++ b/script/test
@@ -203,8 +203,10 @@ function testSuitesForPlatform (platform) {
     case 'darwin':
       if (process.env.ATOM_RUN_CORE_TESTS === 'true') {
         suites = [runCoreMainProcessTests, runCoreRenderProcessTests]
-      } else if (process.env.ATOM_RUN_PACKAGE_TESTS === 'true') {
-        suites = packageTestSuites
+      } else if (process.env.ATOM_RUN_PACKAGE_TESTS === '1') {
+        suites = packageTestSuites.slice(0, Math.floor(packageTestSuites.length / 2))
+      } else if (process.env.ATOM_RUN_PACKAGE_TESTS === '2') {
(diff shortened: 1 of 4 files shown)


nathansobo (approved):
:zap: Put that in your pipeline and smoke it!

Issue #14952, opened 2017-07-04
Editor hangs/freeze with open a big (~4MB) file that only have a single long line


Daijobou:
First, thanks for your great work! I love this editor. :)

### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Editor hangs with open a 5MB json-file, but it open successful a other 6MB json-file.

### Steps to Reproduce

Start atom and open (my) ~5MB json-file

**Expected behavior:** [What you expect to happen]

Loading the json-file. Other editors load it all in <1s

**Actual behavior:** [What actually happens]

Atom hangs with no response, so windows ask to wait or to close it. I have waited multiple times, never come back. CPU process is ~20% and RAM ~500MB.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

### Additional Information

I think here is something in the file that atom not like. Other bigger files working fine. The difference here is that the file with issue is a single line json. Maybe that here the problem?

rsese:
Thanks for the report!  Can you share what operating system you're running?  Also, would you be able to share the file that's causing the hang so we can try it out?

Daijobou:
It's Windows 10 64bit. You only need a single line file like this code:

[{"FieldA":1,"FieldB":"Text","FieldC":"Text"},{"FieldA":1,"FieldB":"Text","FieldC":"Text"},,.....]

I have add a example as "TXT", you must rename it to "JSON" as file-type. I can't upload json here.

[atom_test.txt](https://github.com/atom/atom/files/1168290/atom_test.txt)

Zireael07:
@Daijobou: You said the file is a single line. Does the problem happen if it's plain text (i.e. no syntax highlighting)?

Daijobou:
[...]

PR #19453, opened 2019-06-05
Upgrade text buffer

nathansobo:
I'm opening this PR to test the optimization from https://github.com/atom/text-buffer/pull/312 against Atom. I think it should be fine, but I want to see a green Atom build first before publishing an official `text-buffer` release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6551,9 +6551,9 @@
       }
     },
     "text-buffer": {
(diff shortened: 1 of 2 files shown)


nathansobo:
Failure is a flake. Merging.

nathansobo:
Fixes #14952

PR #19455, opened 2019-06-06
Don't listen for arguments from new processes in test or benchmarks

as-cii:
This pull request prevents the following macOS firewall prompt from showing up when running tests:
<p align=center>
Screen Shot 2019-06-06 at 10 24 59</p>

The prompt was being shown because we had some code within `listenForArgumentsFromNewProcess` that would conditionally create a domain socket file when `AtomApplication` wasn't being instantiated in test or benchmark environments. For test/benchmark environments, this meant that `this.socketPath` would be undefined.

Unfortunately, the conditional didn't extend all the way down until listening for incoming connections, which had the unfortunate side effect of calling `server.listen` with an undefined argument (`this.socketPath`). In turn, this would instruct Node to listen for any TCP connection on a random port, which would then kick off the firewall given that the machine could potentially accept connections coming from the Internet or LAN.

With this pull request, we are changing that behavior by moving the conditional up, and only call `listenForArgumentsFromNewProcess` when instantiating a real `AtomApplication`.

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -283,7 +283,12 @@ module.exports = class AtomApplication extends EventEmitter {
     // We need to do this because `listenForArgumentsFromNewProcess()` calls `crypto.randomBytes`,
     // which is really slow on Windows machines.
     // (TodoElectronIssue: This got fixed in electron v3: https://github.com/electron/electron/issues/2073).
-    const socketServerPromise = this.listenForArgumentsFromNewProcess(options);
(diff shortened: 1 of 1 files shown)


rafeca (approved):
Ups! 😅Thanks for fixing this!

as-cii:
I am still not sure why autocomplete-plus tests are failing, I am investigating that right now. They seem to fail consistently after #19446 got merged, but I am not really sure how that pull request could cause those tests to fail.

PR #19457, opened 2019-06-06
Run each test suite with a pristine ATOM_HOME directory

as-cii:
I had originally opened this pull request to address flakiness in autocomplete-plus tests, but in investigating that issue I discovered a couple of problems:

1. `compile-cache` was being symlinked in `AtomWindow` tests. This was causing some issues when the `ATOM_HOME` folder did not exist, but it has been fixed as part of 6a88fa4. Furthermore, it allowed us to move main process tests out of the `macOS Build` step, which resulted in removing quite a bit of duplication between `macOS_build` and `macOS_test`.
2. When running `script/test`, there was the potential for previous tests to affect the behavior of subsequent ones, because `ATOM_HOME` would be reused across test suites. I addressed this in ee0ddaa, which will now create a new temporary directory for each suite.

I've run tests multiple times with the two improvements illustrated above, and those seem to mitigate some of the flakiness we were observing for `autocomplete-plus`. I am still not entirely sure of what the root cause of the flakiness was (I couldn't reproduce the problem locally, and debugging tests on CI has an extremely bad feedback loop), but these seem to be good enhancements to our test suite nonetheless.

### DIFF
--- a/script/test
+++ b/script/test
@@ -39,6 +39,7 @@ const childProcess = require('child_process')
 const fs = require('fs-extra')
 const glob = require('glob')
 const path = require('path')
+const temp = require('temp').track()
 
 const CONFIG = require('./config')
 const backupNodeModules = require('./lib/backup-node-modules')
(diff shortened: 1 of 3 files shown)


rafeca:
It seems like we're still getting build failures on `autocomplete-plus` ([link](https://dev.azure.com/github/Atom/_build/results?buildId=42640&view=logs&jobId=6de74252-975b-522d-7a3d-227ca79ca6c1&taskId=b7ded238-9370-50c9-8500-e02f948495a4&lineStart=273&lineEnd=295&colStart=2&colEnd=46)). I'm gonna investigate them

PR #19459, opened 2019-06-06
Add delay to watch-path tests to make them less flaky

rafeca:
I hate this PR, but that's the only way I've found to be able to fix the flakiness of the `path-watcher` tests (more info about the root cause on https://github.com/atom/atom/issues/19442#issuecomment-499433407).

This does not fix the root problem, which seems to be that on Electron v3, `nsfw` sometimes needs several milliseconds to "warm up" on Windows to start receiving change events, but I've spent ~1 day debugging this and my current thinking is that fixing the root cause will take significantly longer.

In order to remove the flakiness (which is quite brutal in `master`), I've sent this PR.

### DIFF
--- a/spec/path-watcher-spec.js
+++ b/spec/path-watcher-spec.js
@@ -7,6 +7,7 @@ import { promisify } from 'util';
 
 import { CompositeDisposable } from 'event-kit';
 import { watchPath, stopAllWatchers } from '../src/path-watcher';
+import { timeoutPromise } from './async-spec-helpers';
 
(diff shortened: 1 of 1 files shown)


nathansobo (approved):
Is it worth only `await`ing on Windows or is it better to be consistent?

rafeca:
> Is it worth only awaiting on Windows or is it better to be consistent?

I don't think it hurts to `await` on other platforms, and this way there's less platform-specific logic...

PR #19460, opened 2019-06-06
⬆️ nslog@3.2.0

rafeca:
This updates `nslog` to add support for Node v12 (which will be useful when upgrading to Electron v5).

----

*List of changes between `nslog@3.0.0` and `nslog@3.2.0`: https://github.com/atom/node-nslog/compare/v3.0.0...v3.2.0*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5011,11 +5011,18 @@
       }
     },
     "nslog": {
-      "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


rafeca:
I'm not going to merge this one until the railcars run tomorrow to minimize the risk of breakages on the beta version.

PR #19465, opened 2019-06-07
Better parallelization on CI

as-cii:
🚧 🚧 🚧


as-cii:
This may be good to do in general, but in order to make it work we would need to change the github package to account for the fact that tests may be run using the default electron app, so I am going to close it for now. We can resume it at some later point if we want to.

PR #19482, opened 2019-06-11
Update apm to remove node-gyp logic

winstliu:
Template does not apply - this is a package version bump.

See atom/apm#845 for the majority of the details. We now delegate to npm to handle Electron header installation rather than trying to do it ourselves through node-gyp.

2.3.0 also contains a Node version bump to 10.2.1, in order to match Electron 3.

### DIFF
--- a/apm/package-lock.json
+++ b/apm/package-lock.json
@@ -4,9 +4,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "atom-package-manager": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/atom-package-manager/-/atom-package-manager-2.3.0.tgz",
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Looks like all the CI builds failed to download `https://iojs.org/download/release/v3.1.10/iojs-v3.1.10-headers.tar.gz`, however I'm not sure why it was trying to grab that in the first place.

50Wliu:
Yep! I have a fix locally and will push it tonight.

Background as to why this is happening:
- node-gyp uses really simple logic to determine the dist URL. Any target version `>=1 <4` is assumed to be `iojs`. That's where the iojs URL is coming from.
- But wait! apm passes the `--dist-url` to npm! Yep...except it doesn't seem like npm passes this flag to node-gyp, rendering it effectively useless.
  - (npm's install logic is also extremely convoluted, and I can't for the life for me figure out _where_ npm shells out to node-gyp or how it does the command-line argument -> environment variable fallback, so this is just an educated guess).

I'm confused as to how node-gyp is picking up on the correct Electron version but not the dist URL (is it only picking up a subset of the flags we pass to npm?), but specifying all of these options using environment variables seems to work, so that's my fix.

PR #19523, opened 2019-06-13
Increase timeout for atom-application.test.js on CI

### LINKED ISSUE #19512: Flaky test: AtomApplication command-line interface behavior: atom:// URLs with a "core" host sends a URI message to the most recently focused non-spec window that owns the open locations
[This test](https://github.com/atom/atom/blob/898c12fe21ad35015793f6b1479cf18713f349ba/spec/main-process/atom-application.test.js#L674-L693) sometimes fails:

```
  1) AtomApplication command-line interface behavior atom:// URLs with a "core" host sends a URI message to the most recently focused non-spec window that owns the open locations:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

Rebuilding usually resolves the failure.

---

**Affected CI provider(s)**: Travis CI, Azure Pipelines

**Affected platform(s)**: Linux

**Example failing build**: https://github.visualstudio.com/Atom/_build/results?buildId=42903&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=6c939d89-0d1a-51f2-8b30-091a7a82e98c&lineStart=270&lineEnd=277&colStart=4&colEnd=1

as-cii:
Fixes #19512 

On Azure and Travis, we recently started observing a few test failures in atom-application.test.js.

After staring at those tests for a bit, it doesn't seem like they are failing due to a race condition. Instead, it is possible that these tests are simply timing out due to CI containers sometimes being overloaded and, thus, slower.  I tested this hypothesis locally by running tests on a VM while stress-testing the host machine. Tests would eventually have passed, but they  timed out before having the chance to do so.

This commit increases the timeout on CI to 10 seconds for `AtomApplication` tests in  an attempt to fix the spurious failures we were observing. This is similar to what we've historically done for renderer process tests (see [spec/spec-helper.coffee](https://github.com/atom/atom/blob/7faea50190a8c6738b6ecee105645ee274a32b6e/spec/spec-helper.coffee#L50)).

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -47,6 +47,10 @@ const {
 describe('AtomApplication', function() {
   let scenario, sinon;
 
+  if (process.env.CI) {
+    this.timeout(10 * 1000);
+  }
+
   beforeEach(async function() {
     sinon = sandbox.create();
     scenario = await LaunchScenario.create(sinon);


jasonrudolph (approved):

gifhuppp:
atom-application.test.js บน CI #19523

PR #19528, opened 2019-06-13
Assign originalFontSize when the Workspace is initialized

### LINKED ISSUE #19521: Setting font size to 14 doesn't persist and has unexpected behaviour on `window:reset-font-size`
### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:

### Description

I got the following behaviour starting with Atom 1.38.x.
When I set the font size as 14 (which is the default), the setting doesn't persist beyond restart and when executing `window:reset-font-size` it sets font size to whatever was previously set by first executed command of either `window:increase-font-size` or `window:decrease-font-size`.

### Steps to Reproduce

If I set the font size as 14 and restart Atom, font size appears as empty (renders placeholder: *Default: 14*), I then get the following two different behaviour depending on the steps:

* Increase font-size with `window:increase-font-size`
* Reset font-size with `window:reset-font-size`
* Result: Font size is now set to `15`

Another result:

* Decrease font-size with `window:decrease-font-size`
* Reset font-size with `window:reset-font-size`
* Result: Font size is now set to `13`

Additional if it Atom already knows a previous font size (like the 13 or 15), `window:reset-font-size` will keep using this font size, even if the user enters another font size manually, fe. 16.

**Expected behavior:** Should be able to manually enter font size 14 and `window:reset-font-size` should reset to whatever was the last manually entered font size

**Actual behavior:** Manually setting font size 14 doesn't persist and `window:reset-font-size` sets font size to whatever was previously known or last used by/in Atom

### Versions
Atom    : 1.38.1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3

apm  2.1.7
npm  6.2.0
node 8.9.3 x64
atom 1.38.1
python 2.7.12
git 2.21.0

OS: Linux Mint 17.3 (Ubuntu 16.04 based)

### Additional Information

Perhaps related to this pull request?
https://github.com/atom/atom/pull/19192

Cheers!

nathansobo:
Fixes https://github.com/atom/atom/issues/19521

When the `Workspace` is created, we capture the `originalFontSize` so that we can restore to it via the `window:reset-font-size` command.

If we tried to capture this font size in the `Workspace`'s constructor, it would always be assigned to the default because the workspace is constructed before the user's settings are loaded. Previously, we attempted to address this by storing the value when the setting `editor.fontSize` *changed*, which would happen when the user settings were loaded. Unfortunately, if the user had their font size assigned to the *default* size, the setting *wouldn't* change when their settings were loaded, and we'd never end up capturing the `originalFontSize`.

In this PR, I go with a different approach. Rather than waiting for the font size to change, I break out an `initialize` method on `Workspace` that only gets called after the settings have been loaded. This is a pattern we use with other major objects in the application, so I think it blends in reasonably well.

### DIFF
--- a/src/atom-environment.js
+++ b/src/atom-environment.js
@@ -313,6 +313,8 @@ class AtomEnvironment {
     this.attachSaveStateListeners();
     this.windowEventHandler.initialize(this.window, this.document);
 
+    this.workspace.initialize();
+
     const didChangeStyles = this.didChangeStyles.bind(this);
     this.disposables.add(this.styles.onDidAddStyleElement(didChangeStyles));
     this.disposables.add(this.styles.onDidUpdateStyleElement(didChangeStyles));
@@ -429,7 +431,7 @@ class AtomEnvironment {
     this.workspace.reset(this.packages);
     this.registerDefaultOpeners();
     this.project.reset(this.packages);
-    this.workspace.subscribeToEvents();
+    this.workspace.initialize();
     this.grammars.clear();
     this.textEditors.clear();
     this.views.clear();
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -256,8 +256,6 @@ module.exports = class Workspace extends Model {
     };
 
     this.incoming = new Map();
-
-    this.subscribeToEvents();
   }
 
   get paneContainer() {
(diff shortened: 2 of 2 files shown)


vinkla:
This probably also fix #9873. Previous attempt to fix this: #19192

REVOLUTIONS001:
warning effeted body

REVOLUTIONS001:
bug all

PR #19530, opened 2019-06-13
Update to Tree-sitter 0.15.5

maxbrunsfeld:
This PR bumps the `tree-sitter` to 0.15.5, which contains the very latest version of the core Tree-sitter library.

We were already using *almost* the latest version, after yesterday's upgrade due to https://github.com/atom/atom/issues/19497, but this brings us up to master.

### Small Behavior Change

There was a small fix to the behavior of the `ts_node_descendant_for_position` method in Tree-sitter, which slightly changes the behavior of some methods in Atom:

In particular, it fixes what I consider to be a bug in the behavior of the `editor:select-larger-syntax-node` where if you had your cursor between two adjacent tokens it would select the token *before* the cursor instead of the token immediately afterward.

It also fixes what I think was a bug in the `bufferRangeForScopeAtPosition` method. I've updated a test to reflect that change.

### Prebuilt Binaries

Today I merged https://github.com/tree-sitter/node-tree-sitter/pull/50, which I think *may* make it so that the prebuilt linux binaries of `tree-sitter` will work for Atom, so that we don't have to manually delete those binaries to avoid weird hangs on Travis. 🤞

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6532,9 +6532,9 @@
       "integrity": "sha512-4hjqbObwlh2dLyW4tcz0Ymw0ggoaVDMveUB9w8kFSQScdRLo0gxO9J7WFcUBo+W3C1TLdFIEwNOWebgZZ0RH9Q=="
     },
     "tree-sitter": {
-      "version": "0.15.4",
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Closing in favor of https://github.com/atom/atom/pull/19530, because I forgot to name this branch in a way that would cause Travis to run.

Issue #19269, opened 2019-05-06
Reduce the number of modules loaded by the main process on startup


rafeca:
## Summary

We want to reduce the number of JS modules that get loaded during the startup time by the Atom main process in order to improve the startup time.

In some situations, we can just deduplicate dependencies (or remove unneeded dependencies) to get a net reduction of loaded modules, in some other situations we may need to delay the loading of modules (which would just delay the cost to whenever that module is used).

## Hypothesis

By delaying the loading (and parsing/evaluation) of non-essential JS modules, we can speed up the startup of Atom

### Atom timing data

Currently, the Shell load time (which is one part of the startup time that corresponds to the time it takes from the [first JS file gets executed](https://github.com/atom/atom/blob/133d9153ede248529204b5bfd6dff5bbc02eba69/src/main-process/start.js#L13) in the main process [until the Atom window gets opened](https://github.com/atom/atom/blob/286a110ef4dc1f50e146e7e08ec320728bae41f4/src/main-process/atom-window.js#L85)) is the main contributor to the Atom startup time.

For the last 5 Atom versions, the shell load time on real users have been around ~1.1s (median time), while the window open time has been around ~160ms.

### Current list of required modules

Currently, Atom's main process loads [396 JS modules](https://gist.github.com/rafeca/6d00e5300fd4ea3741a15bf71980a20e) before starting a new window (~200 of them are unique).

There's some duplication of modules (`fs-extra`) and some modules that shouldn't be needed since they are provided by nodejs core library (`nodegit-promise`, `promisify-node`).

### How to debug the required modules on startup

I've used some hacky code to locally plug into the require system of the main process and store each single loaded module.

To do so, I've added the following logic at the very top of the [main process start file](https://github.com/atom/atom/blob/master/src/main-process/start.js):

```js
const Module = require('module')
const origRequire = Module._load

const modules = new Set()
const startTime = Date.now()

const rootPath = require('path').join(__dirname, '..', '..')
function pathRelative(input) {
  return input.replace(rootPath, '.')
}

Module._load = function (moduleId, from) {
  const modulePath = pathRelative(Module._resolveFilename(moduleId, from))

  if (modulePath.startsWith('./')) {
    // Only log non-core JS files.
    modules.add(
      modulePath + ' (required by ' + pathRelative(from.id) + ')'
    )
  }

  return origRequire.apply(this, arguments)
}

global.__debugSaveStartupModules = () => {
  modules.add('-------')
[...]

PR #19534, opened 2019-06-14
Tweak package-lock to avoid loading `fs-extra` twice on main-process

rafeca:
Following the instructions on https://github.com/atom/atom/issues/19269 I've identified that during the main process startup flow we're loading the `fs-extra` (and all its dependant modules) twice: once for `@atom/watcher` and another time for `@atom/nsfw` ([details](https://gist.github.com/rafeca/b4b44b5dcc8c764e4984221ede9d2ade#file-startup_modules-txt-L163-L338)).

Both of them have the same version (v7.0.1), but for some reason the npm lock file at some point in the past decided to put an older version (v4.0.3) which is used by the `github` package on the root folder, and this prevented these two duplicates of the same version to be hoisted up on the dependency tree.

In this PR I've modified manually the `package-lock` file to make `npm` recalculate the best hoisting for modules, so libraries don't get duplicated (the manual modifications consisted on deleting all the `fs-extra` instances on the `package-lock`).

With this, the number of modules loaded during the main process start gets reduced from [373](https://gist.github.com/rafeca/b4b44b5dcc8c764e4984221ede9d2ade) to [286](https://gist.github.com/rafeca/850576596cc544f2ab6eafb95ebd7309).

I've tried to check if this impacted startup time but I couldn't see noticeable changes (something around ~10-20ms only, and still hard to tell if it was just attributed to noise), but this could have more impact on other slower platforms.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,18 +35,6 @@
       "requires": {
         "fs-extra": "^7.0.0",
         "nan": "^2.10.0"
-      },
-      "dependencies": {
-        "fs-extra": {
-          "version": "7.0.1",
-          "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-7.0.1.tgz",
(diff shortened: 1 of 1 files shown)


rafeca:
I'm gonna merge this one since it only changes the `package-lock.json` file, if there's any concern afterwards I can revert

PR #19538, opened 2019-06-14
Shim tokens in TokenizedLines returned from TreeSitterLanguageMode

### LINKED ISSUE #18973: TreeSitterLanguageMode.tokenizedLineForRow appears to have a bug
### Description

Just for context, I came across this while digging in to figure out an issue in the `atom-aligner` package (https://github.com/adrianlee44/atom-aligner/issues/78) but that is not directly relevant to the issue.

The issue happens when Tree Sitter is enabled.

`TreeSitterLanguageMode.tokenizedLineForRow` explicitly passes in `null` for the `tokenIterator` parameter of the `TokenizedLine` constructor here:
https://github.com/atom/atom/blob/97f7ab3438498b7eb3c3d22396decd81ff7de6a6/src/tree-sitter-language-mode.js#L442-L452

This causes an error if the `getTokenIterator` function of `TokenizedLine` is called:
https://github.com/atom/atom/blob/c1d270dd2a4ebb42100d25cf65537110b5a4d985/src/tokenized-line.coffee#L15

Which, of course, happens by just using `TokenizedLine` at all.

So either `TreeSitterLanguageMode` needs a more complete backwards compatibility shim, or `TokenizedLine` needs to handle the null token iterator and create something to fill the void, but it's clearly broken as is.

### Versions

Atom    : 1.35.0
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3

nathansobo:
Fixes #18973

🍐'd with @as-cii

Package authors are using the private `tokenizedLineForRow` method on language modes to get syntactic information about lines. Even though this method is private, @maxbrunsfeld provided a shim implementation in `TreeSitterLanguageMode` that enabled usage of some basic properties on returned `TokenizedLine` object, but this did not include the ability to read the `tokens` field. This is causing exceptions for at least one popular package, `atom-aligner`.

In this PR, @as-cii and I have enhanced the `TokenizedLine`s returned by `TreeSitterLanguageMode.prototype.tokenizedLineForRow` to include a valid `tokens` array. We construct these tokens eagerly via the highlight iterator, which could impose a cost for those that call this method and don't use `tokens`, but making the method lazy would have been extremely complex due to the potential for the state of the buffer and the syntax tree to change after the `TokenizedLine` is returned.

### DIFF
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -2210,6 +2210,47 @@ describe('TreeSitterLanguageMode', () => {
       expect(editor.getSelectedText()).toBe('html ` <b>c${def()}e${f}g</b> `');
     });
   });
+
+  describe('.tokenizedLineForRow(row)', () => {
+    it('returns a shimmed TokenizedLine with tokens', () => {
+      const grammar = new TreeSitterGrammar(atom.grammars, jsGrammarPath, {
+        parser: 'tree-sitter-javascript',
+        scopes: {
+          program: 'source',
+          'call_expression > identifier': 'function',
+          property_identifier: 'property',
+          'call_expression > member_expression > property_identifier': 'method',
+          identifier: 'variable'
+        }
+      });
+
+      buffer.setText('aa.bbb = cc(d.eee());\n\n    \n  b');
+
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
👏

PR #19551, opened 2019-06-17
Remove delays that we no longer need in watch-path tests

jasonrudolph:
This pull request reverts the changes introduced in https://github.com/atom/atom/pull/19459. With the upgrade to atom/nsfw v1.0.25 in #19525, we should no longer need the delays introduced in #19459. 😅

To gain confidence that we no longer need these delays, I pushed up a [temporary commit](https://github.com/atom/atom/commit/2e6cbd9c19781e1e11697ba289d257eb8e3820db#diff-a5307c09bb725b0581c516daa542522eR93) to [run these two tests 10,000 times on CI](https://github.visualstudio.com/Atom/_build/results?buildId=43309):
- The [first test](https://github.com/atom/atom/blob/2e6cbd9c19781e1e11697ba289d257eb8e3820db/spec/path-watcher-spec.js#L93-L120) ran successfully 10,000 times.
- The [second test](https://github.com/atom/atom/blob/2e6cbd9c19781e1e11697ba289d257eb8e3820db/spec/path-watcher-spec.js#L122-L177) ran successfully 7,406 times before the CI job timed out. (It turns out that Azure Pipelines didn't want to keep running a build that's already been running for 3 hours. 😇😅) Despite "only" running 7,406 times, I think we can safely assume that this test no longer needs the delays introduced in #19459.

---

Refs: https://github.com/atom/atom/issues/19442

### DIFF
--- a/spec/path-watcher-spec.js
+++ b/spec/path-watcher-spec.js
@@ -7,7 +7,6 @@ import { promisify } from 'util';
 
 import { CompositeDisposable } from 'event-kit';
 import { watchPath, stopAllWatchers } from '../src/path-watcher';
-import { timeoutPromise } from './async-spec-helpers';
 
 temp.track();
 
@@ -22,7 +21,6 @@ describe('watchPath', function() {
   let subs;
 
   beforeEach(function() {
-    jasmine.useRealClock();
     subs = new CompositeDisposable();
   });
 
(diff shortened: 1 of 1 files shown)


as-cii (approved):
Nice! ⚡️⚡️⚡️

PR #19552, opened 2019-06-17
⬆️ git-utils@5.6.0, scandal@3.2.0

rafeca:
These two new versions contain fixes for Electron v4.

----

- *List of changes between `git-utils@5.5.0` and `git-utils@5.6.0`: https://github.com/atom/git-utils/compare/v5.5.0...v5.6.0*
- *List of changes between `scandal@3.1.0` and `scandal@3.2.0`: https://github.com/atom/scandal/compare/v3.1.0...v3.2.0*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3192,9 +3192,9 @@
       }
     },
     "git-utils": {
-      "version": "5.5.0",
-      "resolved": "https://registry.npmjs.org/git-utils/-/git-utils-5.5.0.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -5667,18 +5667,27 @@
...
         "argparse": "^1.0.2",
-        "git-utils": "^5.0.0",
+        "git-utils": "^5.6.0",
         "isbinaryfile": "^2.0.4",
         "minimatch": "^2.0.9",
         "split": "^1.0.0",
         "temp": "^0.8.3"
       },
       "dependencies": {
+        "git-utils": {
```
50Wliu: This is odd...seems like it should be deduped with the top-level version as the two version numbers are the same.

rafeca: good point, I'm gonna try to dedupe then in a followup PR

PR #19556, opened 2019-06-17
Add a new injection point setting to support parsing Rust macros via language injection

### LINKED ISSUE #19435: Cannot fold `match` in Rust
### Prerequisites

### Description

The `match` block does not offer a fold arrow, like the `if`, `loop`, etc. blocks do.

### Steps to Reproduce

1. Open Rust file with contents
```
// foo.rs

if foo {
    println!("folds")
}

match foo {
    _ => println!("no fold")
}

```

Can fold `match` block

Can't

100%

### Versions

```
Atom    : 1.37.0-beta1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
```
(last commit Dec 2018 though).

maxbrunsfeld:
Fixes #19435

### Summary

This PR upgrades [`tree-sitter-rust`](https://github.com/tree-sitter/tree-sitter-rust) to the latest version. Because of some changes that I've made to that parser, we need to change the way that we highlight *macro definitions* and *macro invocations* in Rust code. This change required adding a new option to the existing injection point API. It should improve the experience of editing Rust code that contains macros.

### Background - Rust Macros

Macros come up a lot in Rust, especially macro *invocations*, like this:

```rust
assert_eq!(a.b(), C::D { e: f });
```

The Rust parser does not actually *parse* the arguments to macros (everything after the `!` in the example above) as written using normal Rust syntax. Instead, it parses them into simple [*token trees*](https://rust-lang.github.io/rustc-guide/macro-expansion.html) - loose sequences of tokens with bracket-matching as the only structure. The contents of the Token trees get parsed later, after macro expansion.

### Background - Syntax Highlighting Macros

Even though macro arguments are technically just token trees, in practice, they usually contain valid Rust syntax. So it's desirable to highlight macro arguments as if they were Rust expressions.

In the old version of `tree-sitter-rust` (0.13.7 and below), the parser itself would [*try* to parse token trees as Rust code](https://github.com/tree-sitter/tree-sitter-rust/blob/v0.13.7/grammar.js#L224-L249), and fall back to treating them as token trees if this parsing failed. It did this using Tree-sitter's GLR system for handling ambiguity. So in most cases, macro calls would appear in the syntax tree with their arguments fully parsed.

### Changing How Macros are Highlighted

There were a couple of problems with this :point_up: approach:

First of all, if there was *anything* invalid inside of a macro call, the entire macro call would switch back to being parsed as a simple token tree. During the course of typing, this sometimes caused unexpectedly large changes to the highlighting. Also, the ambiguity of parsing token trees as Rust code caused the parser to have a larger binary size than it would otherwise need to.

I've since realized that Rust macros should be highlighted using [**language injection**](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/#language-injection), similarly to what we do for [Ruby code within ERB](https://github.com/atom/language-html/blob/153f176a006e1276888c9ff3309ab76ffebf0701/lib/main.js#L34-L39), and [C code within C macro definitions](https://github.com/atom/language-c/blob/3cd4db63bc916447b08b0201e55e79a464a26deb/lib/main.js#L5-L9). This way, we won't totally give up on parsing the macros due to small parse errors, and we don't need to complicate the Rust parser itself to handle the behavior.

### The `includeChildren` API

Unfortunately, the [current language injection API](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/#language-injection) isn't quite suitable for parsing Rust token trees as Rust code.

Currently, when you add an *injection point* for a given node `N`, we assume that if `N` has internal structure, we should preserve that internal structure. Specifically:

* We continue to render highlighting for the tokens *within* `N`
* When computing which ranges of text we should pass to the injected language parser, we *subtract* those tokens' ranges from `N's` range, so that we can pass only `N's` *own* text, not its children's text.

This is not the behavior we want for handling Rust macros. All of the *tokens* within a token tree are already parsed out as separate tokens by the outer Rust parser. We *do want* to reinterpret them via a second pass of the Rust parser.

I've added a new boolean option to the `GrammarRegistry.addInjectionPoint` method called `includeChildren` that implements this new behavior. If you call `addInjectionPoint` with `includeChildren` set to `true`, then the matching nodes' entire text will be passed to the injected parser, including the nodes' children's text.

### QA Process

In addition to writing one unit test for this new behavior, I've been testing this manually by opening large Rust source files containing lots of macro invocations and macro definitions, like [this one from `serde`](https://github.com/serde-rs/serde/blob/master/serde/src/de/impls.rs), editing, and verifying that syntax highlighting looks good.

### DIFF
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -1315,6 +1315,9 @@ class NullHighlightIterator {
   seek() {
     return [];
   }
+  compare() {
+    return 1;
+  }
   moveToSuccessor() {}
   getOffset() {
     return Infinity;
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -32,7 +32,7 @@ class TreeSitterLanguageMode {
     this.config = config;
     this.grammarRegistry = grammars;
     this.parser = new Parser();
-    this.rootLanguageLayer = new LanguageLayer(this, grammar, 0, true);
+    this.rootLanguageLayer = new LanguageLayer(this, grammar, 0);
     this.injectionsMarkerLayer = buffer.addMarkerLayer();
 
     if (syncTimeoutMicros != null) {
@@ -637,14 +637,13 @@ class TreeSitterLanguageMode {
 }
 
 class LanguageLayer {
-  constructor(languageMode, grammar, depth, isOpaque) {
+  constructor(languageMode, grammar, depth) {
     this.languageMode = languageMode;
     this.grammar = grammar;
     this.tree = null;
     this.currentParsePromise = null;
     this.patchSinceCurrentParseStarted = null;
     this.depth = depth;
-    this.isOpaque = isOpaque;
   }
 
   buildHighlightIterator() {
@@ -886,8 +885,7 @@ class LanguageLayer {
           marker.languageLayer = new LanguageLayer(
             this.languageMode,
             grammar,
-            this.depth + 1,
-            injectionPoint.includeChildren
+            this.depth + 1
           );
           marker.parentLanguageLayer = this;
         }
@@ -940,7 +938,6 @@ class HighlightIterator {
   constructor(languageMode) {
     this.languageMode = languageMode;
     this.iterators = null;
-    this.opaqueLayerDepth = 0;
   }
 
   seek(targetPosition, endRow) {
@@ -961,6 +958,9 @@ class HighlightIterator {
     if (iterator.seek(targetIndex, containingTags, containingTagStartIndices)) {
       this.iterators.push(iterator);
     }
+
+    // Populate the iterators array with all of the iterators whose syntax
+    // trees span the given position.
     for (const marker of injectionMarkers) {
       const iterator = marker.languageLayer.buildHighlightIterator();
       if (
@@ -969,46 +969,54 @@ class HighlightIterator {
         this.iterators.push(iterator);
       }
     }
+
+    // Sort the iterators so that the last one in the array is the earliest
+    // in the document, and represents the current position.
     this.iterators.sort((a, b) => b.compare(a));
+    this.detectCoveredScope();
+
     return containingTags;
   }
 
   moveToSuccessor() {
+    // Advance the earliest layer iterator to its next scope boundary.
     let leader = last(this.iterators);
+
+    // Maintain the sorting of the iterators by their position in the document.
     if (leader.moveToSuccessor()) {
       const leaderIndex = this.iterators.length - 1;
       let i = leaderIndex;
       while (i > 0 && this.iterators[i - 1].compare(leader) < 0) i--;
       if (i < leaderIndex) {
         this.iterators.splice(i, 0, this.iterators.pop());
-        this.trackLayerDepth();
       }
     } else {
+      // If the layer iterator was at the end of its syntax tree, then remove
+      // it from the array.
       this.iterators.pop();
-      this.trackLayerDepth();
     }
+
+    this.detectCoveredScope();
   }
 
-  trackLayerDepth() {
-    let i = this.iterators.length - 1;
-    let iterator = this.iterators[i];
-    if (!iterator) return;
-    const offset = iterator.getOffset();
-    if (iterator.languageLayer.isOpaque) {
-      this.opaqueLayerDepth = iterator.languageLayer.depth;
-    }
-
-    while (i > 0) {
-      i--;
-      iterator = this.iterators[i];
-      if (iterator.startOffset > offset) break;
-      if (iterator.languageLayer.isOpaque) {
-        const { depth } = iterator.languageLayer;
-        if (depth > this.opaqueLayerDepth) {
-          this.opaqueLayerDepth = depth;
-        }
+  // Detect whether or not another more deeply-nested language layer has a
+  // scope boundary at this same position. If so, the current language layer's
+  // scope boundary should not be reported.
+  detectCoveredScope() {
+    const layerCount = this.iterators.length;
+    if (layerCount > 1) {
+      const first = this.iterators[layerCount - 1];
+      const next = this.iterators[layerCount - 2];
+      if (
+        next.offset === first.offset &&
+        next.atEnd === first.atEnd &&
+        next.languageLayer.depth > first.languageLayer.depth
+      ) {
+        this.currentScopeIsCovered = true;
+        return;
       }
     }
+    this.currentScopeIsCovered = false;
   }
 
   getPosition() {
@@ -1022,20 +1030,18 @@ class HighlightIterator {
 
   getCloseScopeIds() {
     const iterator = last(this.iterators);
-    if (iterator && iterator.languageLayer.depth >= this.opaqueLayerDepth) {
+    if (iterator && !this.currentScopeIsCovered) {
       return iterator.getCloseScopeIds();
-    } else {
-      return [];
     }
+    return [];
   }
 
   getOpenScopeIds() {
     const iterator = last(this.iterators);
-    if (iterator && iterator.languageLayer.depth >= this.opaqueLayerDepth) {
+    if (iterator && !this.currentScopeIsCovered) {
       return iterator.getOpenScopeIds();
-    } else {
-      return [];
     }
+    return [];
   }
 
   logState() {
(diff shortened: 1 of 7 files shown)


src/tree-sitter-language-mode.js
```diff
@@ -637,13 +637,14 @@ class TreeSitterLanguageMode {
...
-  constructor(languageMode, grammar, contentChildTypes) {
+  constructor(languageMode, grammar, depth, isOpaque) {
     this.languageMode = languageMode;
     this.grammar = grammar;
     this.tree = null;
     this.currentParsePromise = null;
     this.patchSinceCurrentParseStarted = null;
-    this.contentChildTypes = contentChildTypes;
+    this.depth = depth;
+    this.isOpaque = isOpaque;
```
maxbrunsfeld: I added a boolean flag on `LanguageLayer` that I'm calling `isOpaque`, which indicates whether or not the parent layer's token should *show through* this layer or not. It basically matches the value of `includeChildren`. So by default, when `includeChildren` is false, the parent layer's highlighting does show. But for injections with `includeChildren: true`, the parent layer is covered up.

src/tree-sitter-language-mode.js
```diff
@@ -947,55 +950,97 @@ class HighlightIterator {
...
-    this.iterators.sort((a, b) => b.getIndex() - a.getIndex());
+    for (const marker of injectionMarkers) {
+      const iterator = marker.languageLayer.buildHighlightIterator();
+      if (
+        iterator.seek(targetIndex, containingTags, containingTagStartIndices)
+      ) {
+        this.iterators.push(iterator);
+      }
+    }
+    this.iterators.sort((a, b) => b.compare(a));
```
maxbrunsfeld: I've also done a refactor which made it easier to implement the new functionality. I've changed the invariant of the `HighlightIterator` a bit. Previously, the the `iterators` array always contained one `HighlightLayerIterator` for *every* `HighlightLayer`. Layer iterators that were no longer active had a `done` flag set to `true`. Now, once a `HighlightLayerIterator` advances to the end of its syntax tree, I remove it from the `iterators` array. This removes the need for the `done` flag.

The reason for this diff is that `LayerIterator.seek` can now return `false`, which means that the iterator has already fallen off the end of its tree. If `seek` returns false, we avoid adding the iterator to the array at all.

src/tree-sitter-language-mode.js
```diff
@@ -947,55 +950,97 @@ class HighlightIterator {
...
+      while (i > 0 && this.iterators[i - 1].compare(leader) < 0) i--;
+      if (i < leaderIndex) {
+        this.iterators.splice(i, 0, this.iterators.pop());
+        this.trackLayerDepth();
+      }
+    } else {
+      this.iterators.pop();
+      this.trackLayerDepth();
+    }
+  }
```
maxbrunsfeld: The main difference in `moveToSuccessor` is that like `seek`, I've updated it so that it can return `false`, meaning that the iterator has fallen off the end and should be removed.

src/tree-sitter-language-mode.js
```diff
@@ -947,55 +950,97 @@ class HighlightIterator {
...
+      i--;
+      iterator = this.iterators[i];
+      if (iterator.startOffset > offset) break;
+      if (iterator.languageLayer.isOpaque) {
+        const { depth } = iterator.languageLayer;
+        if (depth > this.opaqueLayerDepth) {
+          this.opaqueLayerDepth = depth;
+        }
+      }
+    }
```
maxbrunsfeld: Whenever the ordering of the iterators changes, we call this method which records the depth of any *opaque* layers that span the current position. By consulting this depth value, we prevent highlights from being emitted from parent layers while we're inside of child layers.

In particular, when we've injected a Rust layer into a token tree node in the parent syntax tree, it prevents us from rendering highlights for the token tree node. That would be bad, because we'd get duplicated scopes, and conflicting scopes for some tokens. For example, identifier in the code would get styles as *both* `type_identifier` (via the injected layer) and regular `identifier` (via the original token tree).

src/tree-sitter-language-mode.js
```diff
@@ -1293,18 +1346,20 @@ class NodeRangeSet {
...
+          if (nextIndex > index) {
+            this._pushRange(buffer, previousRanges, result, {
+              startIndex: index,
+              endIndex: nextIndex,
+              startPosition: position,
+              endPosition: child.startPosition
+            });
+          }
+          position = child.endPosition;
+          index = child.endIndex;
```
maxbrunsfeld: This is where we would subtract out the ranges of child nodes when computing injection ranges. The `includeChildren` flag suppresses that behavior.

maxbrunsfeld:
Here's an example of the improvement

**Previous behavior** - when typing inside of a macro call, temporarily invalid code would cause the macro to lose its Rust syntax highlighting. Notice it happening at different levels of nesting:

rust-macro-bad

**New behavior** - we get normal error recovery in the injected parse tree, so the highlighting doesn't change due to minor errors:

rust-macro-good

maxbrunsfeld:
Hey @as-cii or @nathansobo any chance you could give this a 👀 at some point? It contains @as-cii's `tree-sitter-rust` fix and some other improvements. I'm trying to stay more on top of the task of keeping all the Tree-sitter stuff up-to-date in Atom.

Unfortunately, the code changes here are kinda complex. Some of the naming might need some work too. I could talk about it synchronously at some point if it'd be helpful.

maxbrunsfeld:
Looks like a spurious CI [failure](https://github.visualstudio.com/Atom/_build/results?buildId=43605) in `find-and-replace`:

```
ResultsView
  copying items with core:copy
    it ResultsView copying items with core:copy copies the selected line onto the clipboard
  copying path with find-and-replace:copy-path
    it ResultsView copying path with find-and-replace:copy-path copies the selected file path to clipboard
    it ResultsView copying path with find-and-replace:copy-path copies the selected file path to the clipboard when there are multiple project folders
```

as-cii (commented):
I left a couple of comments and started reviewing this alongside @nathansobo as well, who may have additional feedback.

Overall, I think this approach works well and implementation-wise this seems pretty good already. As for `isOpaque`, I think it conveys the idea and fits well in the layer analogy, so I'd be okay with keeping it! 👍

src/tree-sitter-language-mode.js
```diff
@@ -910,7 +913,6 @@ class LanguageLayer {
     }
 
     if (markersToUpdate.size > 0) {
-      this.lastUpdateWasAsync = true;
```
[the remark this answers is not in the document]
maxbrunsfeld: Yeah, that was a spurious line that I accidentally introduced a long time ago in an earlier PR. I don't think that property is ever read, or written anywhere else.

src/tree-sitter-language-mode.js
```diff
@@ -947,75 +947,132 @@ class HighlightIterator {
...
+      if (
+        next.offset === first.offset &&
+        next.atEnd === first.atEnd &&
+        next.languageLayer.depth > first.languageLayer.depth
+      ) {
+        this.currentScopeIsCovered = true;
+        return;
+      }
+    }
+    this.currentScopeIsCovered = false;
```
maxbrunsfeld: Ok, I changed the way that we suppress the scope boundaries from the outer layers. I got rid of the concept of "opaque" language layers, and instead I just detect duplicate scope boundaries individually. After each call to `seek` or `moveToSuccessor`, we check if the leading layer iterator is on top of a more deeply-nested iterator. If so, we hide its scope boundaries.

src/tree-sitter-language-mode.js
maxbrunsfeld: Yeah, it's ok for `moveToSuccessor` to not actually change position, but to just update the opening and closing scopes. In the `TreeSitterLanguageMode`, this actually happens move frequently -  it happens whenever one scope starts at the same position where another scope ends. For simplicity of implementation, we don't report the closing and opening scopes in the same iteration.

src/tree-sitter-language-mode.js
nathansobo: Are you concerned about boundaries from a shallower language mode that don't line up to the same offsets as boundaries in the deeper mode? I guess in the Rust macro scenario, the boundaries in the token tree produced by the outer grammar align with the boundaries in the inner grammar, because the tree produced by the inner grammar is going to align with those tokens. Will there ever be cases where scope boundaries in the outer grammar don't align?

If so, I'm wondering if it makes sense to somehow assign a `depth` on the `HighlightIterator` as a whole that can be incremented when we enter an opaque zone and decremented when it ends. Then you could just ignore boundaries from a language mode that's shallower than the current depth regardless of the offset at which they occur. You could sort the deeper layers first and assign the `depth` whenever you start emitting token boundaries on a deeper layer. As soon as that layer gets popped from the array, you could decrement the depth to the next layer.

I suppose if we're not worried about scope boundaries from shallower layers that don't align with boundaries at the deeper layer, this doesn't matter. Just thought I'd share the thought process in case it's valuable.

maxbrunsfeld: > Are you concerned about boundaries from a shallower language mode that don't line up to the same offsets as boundaries in the deeper mode?

I'm not totally sure, because I don't have a use case in mind, but I actually think it's better that we allow these types non-aligned scopes from the parent layer to still be rendered.

Even though my initial implementation had this concept of "opaque" layers, which would cover up other layers, I don't really think that concept is needed. And if we ever do need it, we might not want to conflate it with the `includeChildren` option. I had conflated the two at first, but then I realized that all I was really trying to solve was the problem of duplicate scope boundaries from inner and outer layers.

It's also kind of tricky to implement the opaque language layer thing, because technically, even an opaque language layer could have "holes" in it, due to holes in its parent layer 😬. My previous code that didn't really handle that situation.

PR #19557, opened 2019-06-18
Store/retrieve state from windows that have no opened project

rafeca:
This PR attempts to solve the root issue that causes the confusion around the "Restore previous window on start" config setting (reported in https://github.com/atom/settings-view/issues/1129).

## Problem

The root problem is that in order to store (or retrieve) an editor window state, Atom uses the list of folders that are currently opened as the key. This is a simple approach but has the limitation that cannot be used to store the state when opening a single file or a set of files (as feature which was introduced in https://github.com/atom/atom/pull/18608).

## Solution

In order to fix this, I've introduced the `windowId` param, which is a unique identifier for each `window` which is used for storing and retrieving the state of windows that do not have any opened project.

This allows to keep the full state of Atom windows (even if there are many windows without any opened project) when opening Atom without passing any arguments.

## Major drawback

There's a big drawback on this solution, which IMHO we need to fix before considering merging it: Since there's no cleaning mechanism on the [`StateStore`](https://github.com/atom/atom/blob/master/src/state-store.js) (the `IndexedDB` database which keeps the state of the Atom windows), adding this logic will make its size increase indefinitely, since window identifiers will get generated every time a new Atom window without a project gets created (so e.g every time a user runs `atom Readme.md`).

This problem is already happening on `master`, since every time that a user opens a folder on Atom, this folder state will permanently be stored on the `IndexedDB` database, but with this PR the issue is amplified considerably (specially for people that use Atom to edit single files).

## Next steps

As I mention, I don't think this PR is shippable in the current state, but I've sent it anyways to:

- Gather feedback about the approach and whether it makes sense to invest time on fixing the drawback.
- If we want to go forward with this solution, I'll let anyone who's willing to fix the drawback to do so. The easiest way to fix it is to make the `StateStore` a LRU cache.
- The last step would be to add some tests to this PR 😃

### DIFF
--- a/src/atom-environment.js
+++ b/src/atom-environment.js
@@ -1499,7 +1499,18 @@ or use Pane::saveItemAs for programmatic saving.`);
   }
 
   getStateKey(paths) {
-    if (paths && paths.length > 0) {
+    if (!paths) {
+      return null;
+    }
+
+    if (paths.length === 0) {
+      // If the current window does not have any project opened, use the
+      // windowId to retrieve its state. Maybe we could do that for all
+      // the windows.
+      paths = [this.getLoadSettings().windowId];
+    }
+
+    if (paths.length > 0) {
       const sha1 = crypto
         .createHash('sha1')
         .update(
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -384,6 +384,7 @@ module.exports = class AtomApplication extends EventEmitter {
       clearWindowState,
       addToLastWindow,
       preserveFocus,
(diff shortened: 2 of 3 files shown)


sadick254:
Hey @rafeca. I see we have conflicts on this PR. I have also noticed that this PR has been open for long. Are you still planning to work on this?

sadick254:
We haven't gotten a response to our questions in our comment above.  I'm going to close this but don't hesitate to reach out if you get time to resolve the conflicts, we'll be happy to reopen the PR.

PR #19575, opened 2019-06-19
Fix exception when multiple language layers are not yet highlighted

maxbrunsfeld:
This fixes a regression introduced in #19556.

The code introduced in that PR ☝️ relied on the existence of a property `LayerHighlightIterator.languageLayer` which was not present on the `NullHighlightIterator`. This caused intermittent exceptions in the case where there were multiple injected language layers whose asynchronous parsing had not yet completed.


as-cii (approved):
Nice catch!

PR #19580, opened 2019-06-20
Upgrade npm on Azure Pipelines when running macOS tests

as-cii:
This fixes an error in macOS tests that would occur when creating a pull requests from a fork.

/cc: @Aerijo


nathansobo:
The test failure is a flaky test that I just fixed. Merging.

PR #19591, opened 2019-06-24
Upload Electron crash reports to Bugsnag via Atom.io if user has consented to Telemetry

nathansobo:
🍐'd with @jasonrudolph

Currently, we are not capturing crashes of the Electron render process or main process that occur in production. With this PR, we tell the Electron crash reporter that we want to upload crash reports to `https://atom.io/crash_reports` if the user has consented to telemetry. We include the value of `require('os').release()` and `require('os').arch()` as well as the `releaseChannel` as `extra` fields to be sent by the reporter.

I will open another PR on Atom.io that uses the reported information to report the crash to Bugsnag via Bugsnag's REST API.


50Wliu:
This should be documented on the metrics repo.

PR #19597, opened 2019-06-25
Rely on Azure Pipelines instead of Travis for Linux builds

nathansobo:
Since we currently build Atom on all three platforms in Azure Pipelines, it makes sense to stop relying on Travis for our Linux builds and artifacts.

-----
[View rendered README.md](https://github.com/atom/atom/blob/ns/remove-travis/README.md)
[View rendered docs/build-instructions/build-status.md](https://github.com/atom/atom/blob/ns/remove-travis/docs/build-instructions/build-status.md)


jasonrudolph (approved):
Looks good! 👌

PR #19602, opened 2019-06-26
Update getInvisibles documentation

### LINKED ISSUE #16042: Inline documentation refers to setInvisibles, which doesn't exist
https://github.com/atom/atom/blob/bc061bb608773244b32edf18bc48db7b92efc609/src/text-editor.js#L3595-L3603

A project-wide search reveals only a single result for `setInvisibles`, which is the comment itself.

UziTech:
### Description of the Change

fixes #16042

Updates documentation for getInvisibles since setInvisibles is no longer a method.

### Release Notes
N/A

### DIFF
--- a/src/text-editor.js
+++ b/src/text-editor.js
@@ -4084,7 +4084,9 @@ module.exports = class TextEditor {
   }
 
   // Returns an {Object} representing the current invisible character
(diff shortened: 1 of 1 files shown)


as-cii:
Thanks, @UziTech! 💯

PR #19612, opened 2019-06-28
Debounce the resize event handler

rafeca:
## Context

Atom is currently listening to window `resize` events to store the window width/height and position and be able to restore the window dimensions when it gets reopened.

In order to do so, the `resize` handler is doing the following:
* Get the current `BrowserWindow` object from the main process (synchronous IPC call).
* Get the `BrowserWindow` size from the main process (synchronous IPC call).
* Get the `BrowserWindow` position from the main process (synchronous IPC call).
* Validate that data is correct.
* Serialize the data into JSON and store it on `localStorage` (all synchronous).

All this processing usually takes 2-5ms on my high end machine, but under some circumstances (don't know what triggers it) it can take 40-50ms (or even up to 200-400ms, for example on Electron v4 because of https://github.com/electron/electron/issues/19027). These times are way above the frame budget and cause very noticeable lagging on the UI when resizing.

## Solution

The solution is a one-line change: just using a simple debouncer to only store the windows dimensions when the resizing has finished makes the UI lag-free.

[Here there's a comparison video](https://imgur.com/a/ySTlc17) between current master and this PR, to see the noticeable difference (I didn't upload a `gif` since it cannot be appreciated well with that format due to the framerate).

### DIFF
--- a/spec/window-event-handler-spec.js
+++ b/spec/window-event-handler-spec.js
@@ -1,5 +1,6 @@
 const KeymapManager = require('atom-keymap');
 const WindowEventHandler = require('../src/window-event-handler');
+const { conditionPromise } = require('./async-spec-helpers');
 
 describe('WindowEventHandler', () => {
   let windowEventHandler;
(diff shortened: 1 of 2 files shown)


rafeca:
Probably there are some failing tests... I'll leave CI to discover them for me 🙃

robertrossmann:
Wow, this simple change suddenly makes Atom feel like a native app. Nice and thanks! ❤️ 🚀

PR #19615, opened 2019-06-30
Add ripgrep pcre2 search support

### LINKED ISSUE #571: No support for lookbehind
Implementing JavaScript's native RegExp engine for a find/replace tool in a text editor seems like a dramatic oversight.

As find-and-replace leans on JS's RegExp, of course the following find expressions is invalid

``` js
(?<=hello)world
(?<!hello)world
```

Errors

``` plain
Invalid regular expression: /(?<=hello)world/: Invalid group
Invalid regular expression: /(?<!hello)world/: Invalid group
```

This module is unusable without such features

dwelle:

Suggestion discussed [in this PR](https://github.com/atom/find-and-replace/pull/1086#issuecomment-502605862).

Fixes https://github.com/atom/find-and-replace/issues/571

### Description of the Change

Added option to enable `pcre2` ripgrep regex engine for search.

### Alternate Designs

None applicable.

### Possible Drawbacks

Enabling PCRE2 potentially incurs performance penalty, as explained on [ripgrep's wiki](https://github.com/BurntSushi/ripgrep/blob/7b3fe6b3251a18d2b8d3efe7ee6a85c9e9e4e565/FAQ.md#why-does-ripgrep-get-slower-when-i-enable-pcre2-regexes), thus this PR depends on a new `atom/find-and-replace` PCRE2 option, being added [in this PR](https://github.com/atom/find-and-replace/pull/1095).

### Verification Process

I haven't verified anything because I haven't managed to build Atom ATM. Considering the nature of the change, I'm relying on the CI. If this is not acceptable, feel free to close or keep the PR open until I manage to build Atom locally, but I give no promises as to when I'll get the time to do so.

### Release Notes

- Added support to enable `pcre2` for ripgrep search engine in find-and-replace package to enable advanced regex features such as lookbehind.

### DIFF
--- a/spec/workspace-spec.js
+++ b/spec/workspace-spec.js
@@ -2652,6 +2652,30 @@ describe('Workspace', () => {
               });
             });
           });
+          describe('pcre2 enabled', async () => {
+            it('supports lookbehind searches', async () => {
+              const results = [];
+
+              await scan(/(?<!a)aa\b/, { PCRE2: true }, result =>
+                results.push(result)
+              );
+
+              expect(results.length).toBe(1);
+              const { filePath, matches } = results[0];
+              expect(filePath).toBe(
+                atom.project.getDirectories()[0].resolve('a')
+              );
+              expect(matches).toHaveLength(1);
+              expect(matches[0]).toEqual({
+                matchText: 'aa',
+                lineText: 'cc aa cc',
+                lineTextOffset: 0,
+                range: [[1, 3], [1, 5]],
+                leadingContextLines: [],
+                trailingContextLines: []
+              });
+            });
+          });
         }
 
         it('returns results on lines with unicode strings', async () => {
--- a/src/ripgrep-directory-searcher.js
+++ b/src/ripgrep-directory-searcher.js
@@ -269,6 +269,10 @@ module.exports = class RipgrepDirectorySearcher {
       args.push('--no-ignore-vcs');
     }
 
+    if (options.PCRE2) {
+      args.push('--pcre2');
+    }
+
     args.push('.');
 
     const child = spawn(this.rgPath, args, {
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -2104,6 +2104,7 @@ module.exports = class Workspace extends Model {
         follow: this.config.get('core.followSymlinks'),
         leadingContextLineCount: options.leadingContextLineCount || 0,
         trailingContextLineCount: options.trailingContextLineCount || 0,
+        PCRE2: options.PCRE2,
         didMatch: result => {
           if (!this.project.isPathModified(result.filePath)) {
             return iterator(result);


spec/workspace-spec.js
```diff
@@ -2652,6 +2652,26 @@ describe('Workspace', () => {
...
           });
+          describe('pcre2 enabled', async () => {
+            it('supports lookbehind searches', async () => {
+              const results = [];
+
+              await scan(/(?<!a)aa\b/, {}, result => results.push(result));
+
+              expect(results.length).toBe(1);
+              const { filePath, matches } = results[0];
+              expect(filePath).toBe(atom.project.getDirectories()[0].resolve('a'));
```
rafeca: I think that this should fix the linter error:

```suggestion
              expect(filePath).toBe(
                atom.project.getDirectories()[0].resolve('a')
              );
```

rafeca:
Thanks for the contribution!

> I haven't verified anything because I haven't managed to build Atom ATM. Considering the nature of the change, I'm relying on the CI. If this is not acceptable, feel free to close or keep the PR open until I manage to build Atom locally, but I give no promises as to when I'll get the time to do so.

That's fine 😃 As long as CI passes it should be ok 😄 

Currently there's a linter error, these errors are fixed it by running `script/lint --fix` (or by integrating `prettier` on Atom), but if you haven't been able to bootstrap the environment correctly, that script may not work.

I've added a suggestion that I think should fix the error [based on the CI error message](https://github.visualstudio.com/Atom/_build/results?buildId=44712&view=logs&jobId=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&taskId=b808d44d-d1c2-5399-c2e1-8c4b890e97af&lineStart=14&lineEnd=15&colStart=1&colEnd=1).

Codewise looks good to me, so once the CI build passes we can merge this one 😄

spec/workspace-spec.js
dwelle: My bad - I'm used to local eslint installations. That being said, can't install `babel-eslint` due to that gyp/msbuild.exe error again (on windows it needs to be installed locally), so I'll trust you and commit your lint fix directly.

dwelle:
Seems I forgot to pass `PCRE2: true` in specs --- fixed. Oh the perils of coding blind :).

dwelle:
lint again.. should I squash those commits?

rafeca:
> lint again.. should I squash those commits?

👍

dwelle:
Removed expectations for `leadingContextLines`/`trailingContextLines` since I guess it's not necessary to assert context for this spec.

rafeca (approved):
🤗

BurntSushi:
Y'all might also be also be interested in ripgrep's `--auto-hybrid-regex` flag, which was added in ripgrep 11:

```
       --auto-hybrid-regex
           When this flag is used, ripgrep will dynamically choose
           between supported regex engines depending on the features
           used in a pattern. When ripgrep chooses a regex engine, it
           applies that choice for every regex provided to ripgrep
           (e.g., via multiple -e/--regexp or -f/--file flags).

           As an example of how this flag might behave, ripgrep will
           attempt to use its default finite automata based regex
           engine whenever the pattern can be successfully compiled
           with that regex engine. If PCRE2 is enabled and if the
           pattern given could not be compiled with the default regex
           engine, then PCRE2 will be automatically used for searching.
           If PCRE2 isn’t available, then this flag has no effect
           because there is only one regex engine to choose from.

           In the future, ripgrep may adjust its heuristics for how it
           decides which regex engine to use. In general, the
           heuristics will be limited to a static analysis of the
           patterns, and not to any specific runtime behavior observed
           while searching files.

           The primary downside of using this flag is that it may not
           always be obvious which regex engine ripgrep uses, and thus,
           the match semantics or performance profile of ripgrep may
           subtly and unexpectedly change. However, in many cases, all
           regex engines will agree on what constitutes a match and it
           can be nice to transparently support more advanced regex
           features like look-around and backreferences without
           explicitly needing to enable them.

           This flag can be disabled with --no-auto-hybrid-regex.
```

dwelle:
@BurntSushi thanks for heads-up. I see vscode recently implemented it [as well](https://github.com/Microsoft/vscode/issues/72459), with a plan to deprecate `--pcre2` flag altogether.

PR #19623, opened 2019-07-03
RFC: Evaluating scope name additions to built-in grammars

savetheclocktower:
This RFC is about how to evolve grammars and syntax themes so that their design goals don't get in each others' way.

I am utterly certain that a maximum of four people on earth will care about this, but I'd love to find out I'm wrong. Hopefully some discussion can help refine exactly what is being proposed here.

[Rendered version](https://github.com/savetheclocktower/atom/blob/apd-rfc-scope-names/docs/rfcs/005-scope-naming.md).

-----
[View rendered docs/rfcs/005-scope-naming.md](https://github.com/savetheclocktower/atom/blob/apd-rfc-scope-names/docs/rfcs/005-scope-naming.md)

### DIFF
--- a/docs/rfcs/005-scope-naming.md
+++ b/docs/rfcs/005-scope-naming.md
@@ -30,17 +30,17 @@ This is a fine design goal; I don’t think @maxbrunsfeld is wrong to argue for 
 
 This isn’t just a beard-stroking nitpick on my part. I can think of a handful of reasons why someone might want to be able to spot import names at a glance. It’s reasonable for someone to want `foo` in the example above to remain the same color throughout the file. It’s also reasonable, I think, to want `foo` to have a special color on the line where it’s introduced. Or to communicate that this token has special behavior — as it would if you have the [js-hyperclick](https://atom.io/packages/js-hyperclick) package installed and are in the habit of cmd-clicking package names to jump to the files where they’re defined.
 
-I don’t mind that the One Dark syntax theme doesn’t want to give `foo` a special color; I mind that its decision is also binding on _all possible_ syntax themes. If that scope name is present, and it’s undesirable to the a syntax theme, that theme can apply the overrides necessary to ignore it. But if that scope name is missing altogether, that constrains _all_ syntax themes, and I’m unable to write a syntax theme that behaves differently.
+I don’t mind that the One Dark syntax theme doesn’t want to give `foo` a special color; I mind that its decision is also binding on _all possible_ syntax themes. If that scope name is present, and it’s undesirable to a syntax theme, that theme can apply the overrides necessary to ignore it. But if that scope name is missing altogether, that constrains _all_ syntax themes, and I’m unable to write a syntax theme that behaves differently.
 
 Thus, here’s what I propose:
 
-If an issue or PR proposes adding a scope and can justify its presence somehow — including the goal of parity with its non-tree-sitter predecessor — An answer of “no, because the built-in syntax themes don’t want to highlight it that way” should become “OK, but only if someone does the associated work to ensure no visual regressions in the built-in syntax themes.” That someone could be the PR’s author or anyone else who has an interest in getting it landed.
+If an issue or PR proposes adding a scope and can justify its presence somehow — including the goal of parity with its non-tree-sitter predecessor — an answer of “no, because the built-in syntax themes don’t want to highlight it that way” should become “OK, but only if someone does the associated work to ensure no visual regressions in the built-in syntax themes.” That someone could be the PR’s author or anyone else who has an interest in getting it landed.
 
 This is tricky, of course — not only the coordination of PRs across packages, but also the need to apply overrides to all six (is it six? I think it’s six) of the built-in syntax themes. If all built-in themes are going to share an austere philosophy ([and it seems like that’s the plan](https://github.com/atom/atom/pull/18383#issuecomment-435460854)), then perhaps it makes sense for them to start sharing a core set of contextual rules. The only difference between them would be the specific color choices that they make.
 
 ## Drawbacks
 
-The drawback is that what I’m suggesting is a lot of work. I don’t propose it lightly; I propose it because it strikes me as the least bad of available choices.
+The drawback is that what I’m suggesting is a lot of work. I don’t propose it lightly; I propose it because it strikes me as the least bad of all available choices.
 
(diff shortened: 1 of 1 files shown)


lee-dohm:
It looks like this is a recommendation for the triage workflow mainly, if not totally. If that is correct and this is accepted, this document or excerpts from it will probably end up in atom/design-decisions.

maxbrunsfeld:
I don't have strong opinions on this, and am not working on Atom full-time any more, but I'll give my 2 cents here, in case anyone finds anything useful in it.

>  For reference, the `language-babel` grammar scopes `foo` as `variable.other.readwrite.js`. I’d probably opt for something like `variable.import`; others may want to put it into the `support` namespace. There’s actually little cross-language consensus here.

I'm a bit skeptical that there will ever be cross-language consensus with that level of detail. I'd actually love for all of the scopes to become much, much **simpler** - ideally one word like (`type`, `function`, `tag`, `variable`, `property`, `string`), and *occasionally* two words (e.g. `type.builtin`), but only when necessary. IMO, the more complex the scopes become the less compatible themes will be across different languages, the more tightly coupled themes will become to specific grammars, and the more bike shedding will go on endlessly.

When introducing the Tree-sitter grammars, I put a lot of work into trying to make themes look consistent across languages, and I found that I could do it to some degree, by simplifying the scopes. But people ended up needing to add back some of the specificity, mostly for compatibility with community themes. Backward compatibility is a huge impediment in this area.

> I’ve got lots of commands that behave in different ways based on the surrounding scope. The richer the scope descriptor, the better.

In my ideal long-term vision, the scopes we use for syntax highlighting would be decoupled from APIs like `atom.commands` and `atom.config`. The syntax tree itself is a much more precise and performant way to customize behavior syntactically, as we have done in https://github.com/atom/bracket-matcher/pull/367, and with the new folding system.

Unfortunately, I don't have detailed designs for how to use the syntax tree to serve your use cases. And once the API is designed, it's a lot of work to document it and try to migrate existing code to use it.

savetheclocktower:
First off, @maxbrunsfeld: my main fear when writing this is that it’d come off as unreasonably critical or dismissive of your efforts and design choices, because that’s honestly not how I feel.

> I'm a bit skeptical that there will ever be cross-language consensus with that level of detail. I'd actually love for all of the scopes to become much, much simpler - ideally one word like (type, function, tag, variable, property, string), and occasionally two words (e.g. type.builtin), but only when necessary. IMO, the more complex the scopes become, the less compatible themes will be across different languages, the more tightly coupled themes will become to specific grammars, and the more bike shedding will go on endlessly.

I see what you’re saying here, but I think this is still conflating the design goals of syntax themes with the design goals of grammars. Any syntax theme can _choose_ to behave in this way — to color `variable` the same whether it’s `variable.foo` or `variable.bar.baz.thud` — and I’d even agree that the built-in themes should shoot for that kind of simplicity out of the gate.

If I were giving advice to someone writing their first syntax theme, I’d tell them to start out by paying attention to only the initial part of a scope name. Pick your colors for variables, comments, strings, and such, and then you’re 80% done, and left with a theme that will look decent in any conforming grammar. But the last 20% of writing a syntax theme is about distinguishing the exceptions: going through the most popular grammars and applying any necessary tweaks based on the semantics of the particular language or, hell, just personal preference.

I agree that it’s not feasible to get cross-language consensus on how that last 20% should be scoped. Should the `foo` in `import foo from "thing"` be scoped as `variable` or `constant`? I don’t know. Either argument could be made. But if I want it to _look_ like a variable in my syntax theme, I’m out of luck if it’s simply scoped `constant`. At least if it’s `constant.imported-package` I’ve got something to work with.

I understand the view that the existing hierarchy is a bit too left-brained, and even needlessly complex, but I think that the examples you’re proposing are too simple to do the job.

> When introducing the Tree-sitter grammars, I put a lot of work into trying to make themes look consistent across languages, and I found that I could do it to some degree, by simplifying the scopes. But people ended up needing to add back some of the specificity, mostly for compatibility with community themes. Backward compatibility is a huge impediment in this area.

Before I edited this RFC for brevity, I had written several paragraphs on how challenging this task must have been, and how different developers would’ve approached it in various ways, all equally valid. There’s never a _good_ time to try to harmonize the scoping of built-in grammars — it’ll change things, and people will complain — but, since tree-sitters were going to change syntax highlighting anyway, this was a natural time to try.

> In my ideal long-term vision, the scopes we use for syntax highlighting would be decoupled from APIs like atom.commands and atom.config. The syntax tree itself is a much more precise and performant way to customize behavior syntactically, as we have done in atom/bracket-matcher#367, and with the new folding system.

I agree that bracket-matching and folding are better done outside of the scope system. I still think that it’s better to have scope names be the “public interface” around the syntax tree because (a) we still live in a world with TM-style grammars, for which there’s no syntax tree to use; (b) scopes allow someone to customize behavior in an abstract way, without having to know the details of how a certain grammar’s tree-sitter nodes are named.

Let me clarify what I’m talking about in the latter point:

* The built-in `link` package allows you to open a URL if your cursor is within it. To figure out when the cursor is within a URL, it checks for the presence of the `markup.underline.link` scope. This enables it to work in any grammar that scopes URLs in that manner. (The only specific knowledge it has is of Markdown, so that it can follow named hyperlinks like `[link text][footnote]`.)
* The built-in `toggle-quotes` package allows you to toggle a string between single quotes and double quotes (and some other quote delimiters on a user-configurable, per-language basis). To figure out when the cursor is inside of a quoted string, it checks for the presence of the `string.quoted` scope.

These are packages that use scopes for their semantic value apart from syntax highlighting. Of course, these packages could inspect syntax trees instead. But for that to happen, we’d have to go beyond the tree-sitter grammars and come up with some naming conventions for tree-sitter _parsers_. At that point, after a lot of work, we’d end up with a standard for semantic, language-independent naming of common constructs in programming/markup languges. In other words, we’d have reinvented scopes.

> Unfortunately, I don't have detailed designs for how to use the syntax tree to serve your use cases. And once the API is designed, it's a lot of work to document it and try to migrate existing code to use it.

Understood. I’m definitely up for as much of that work as I’m able to do, provided a consensus emerges on how to proceed.

sadick254 (approved):
This looks like something I would consider doing some time in the future. @savetheclocktower Thank you for the detailed RFC.

PR #19673, opened 2019-07-15
Upgrade apm

nathansobo:
This PR upgrades apm to include https://github.com/atom/apm/pull/854.

This is a sensitive component, so I want a green build before disrupting master.

### DIFF
--- a/apm/package-lock.json
+++ b/apm/package-lock.json
@@ -4,9 +4,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "atom-package-manager": {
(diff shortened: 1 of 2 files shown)


Aerijo:
@nathansobo The problem is more than this actually; e.g., `apm remove ..` will delete the whole `.atom` folder. Perhaps it should also validate the folder is a package?

nathansobo:
@Aerijo great point. I opened https://github.com/atom/apm/pull/858 to try to defend against weird edge cases that resolve to a non-package folder.

PR #19726, opened 2019-07-30
Only cover scope boundaries in the parent layer if there is a scope boundary in the injected layer

### LINKED ISSUE #19582: [nightly] Comment highlighting is leaking
### Prerequisites

### Description

Comment highlighting is leaking to non comments in latest nightly (and build from master).

image

### Steps to Reproduce

1. Use 1.40.0-nightly13, and `atom-one-dark` syntax theme
2. Open the Atom repo
3. Open `/packages/grammar-selector/lib/grammar-list-view.js`
4. Look at something following a comment that should have plain highlighting
5. Make an edit inside a wrongly highlighted word. It should be corrected.
6. Close and reopen the file. The bit that was fixed should be broken again.

**Expected behavior:**

Plain white identifiers

**Actual behavior:**

Comment like identifiers

**Reproduces how often:**

100%

I am 99% sure it's not something I've done to my local Atom build.

Edit: good thing I left that 1% :sweat_smile: 

### Versions

```
15:26 $ atom-nightly -v
Atom    : 1.40.0-nightly13
Electron: 3.1.10
Chrome  : 66.0.3359.181
Node    : 10.2.0
```

nathansobo:
Fixes #19582

#19556 introduced new behavior for scopes from injected grammars, causing them to "cover" or "shadow" scope boundaries in the parent grammar. This was added to prevent duplicated and conflicted scope boundaries from being rendered in cases where the syntax tree of a parent grammar contains nodes that overlap with the syntax tree of the child grammar. See the body of #19556 for more details.

In #19582, we discovered an edge case in the injection of the JSDoc grammar into JavaScript block comments. In some cases, the `HighlightIterator` associated with the nested JSDoc grammar ends up getting parked at the end of the comment but *not* getting associated with any `openTags` or `closeTags`. In that scenario, the `HighlightIterator` continues to shadow the scopes from the parent grammar even though it is not expressing a scope boundary itself. This causes the close tag for the comment to be hidden, incorrectly extending the `comment` scope beyond the bounds of the actual comment block.

The solution in this PR is to *only* hide scopes from the parent grammar in situations where the child grammar's `HighlightIterator` has at least one close tag or open tag. This prevents the close of the comment scope from being obscured and solves #19582.

In the process of exploring this issue, it occurred to @maxbrunsfeld and I that there might be more diabolical cases where a node in the child grammar shadows only a close tag in the parent grammar. However, this actually seems unlikely. If we shadow a scope *close* at a given position, there's a high probability that we *also* shadow the scope *open* due to the structure of the syntax tree, thus preventing an unbalanced scope open from occurring. In the event we encounter an actual case of an unbalanced tag caused by shadowing, we will have more information on which to base a solution. In the meantime, the hypothetical possibility doesn't justify the complexity of a full solution.

### DIFF
--- a/src/tree-sitter-language-mode.js
+++ b/src/tree-sitter-language-mode.js
@@ -1010,7 +1010,8 @@ class HighlightIterator {
       if (
         next.offset === first.offset &&
         next.atEnd === first.atEnd &&
-        next.depth > first.depth
+        next.depth > first.depth &&
+        next.openTags.length + next.closeTags.length > 0
       ) {
         this.currentScopeIsCovered = true;
         return;
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -521,7 +521,10 @@ describe('TreeSitterLanguageMode', () => {
             'template_substitution > "}"': 'interpolation'
           },
           injectionRegExp: 'javascript',
-          injectionPoints: [HTML_TEMPLATE_LITERAL_INJECTION_POINT, JSDOC_INJECTION_POINT]
+          injectionPoints: [
+            HTML_TEMPLATE_LITERAL_INJECTION_POINT,
+            JSDOC_INJECTION_POINT
+          ]
         });
 
         htmlGrammar = new TreeSitterGrammar(atom.grammars, htmlGrammarPath, {
@@ -848,12 +851,10 @@ describe('TreeSitterLanguageMode', () => {
             injectionRegExp: 'jsdoc',
             injectionPoints: []
           }
-        )
-
+        );
         atom.grammars.addGrammar(jsGrammar);
         atom.grammars.addGrammar(jsdocGrammar);
 
-        // await atom.packages.activatePackage('language-javascript');
         editor.setGrammar(jsGrammar);
         editor.setText('/**\n*/\n{\n}');
 
@@ -2451,10 +2452,10 @@ const SCRIPT_TAG_INJECTION_POINT = {
 
 const JSDOC_INJECTION_POINT = {
   type: 'comment',
-  language (comment) {
-    if (comment.text.startsWith('/**')) return 'jsdoc'
+  language(comment) {
+    if (comment.text.startsWith('/**')) return 'jsdoc';
   },
-  content (comment) {
-    return comment
+  content(comment) {
+    return comment;
   }
-}
+};
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -401,6 +401,35 @@ describe('TreeSitterLanguageMode', () => {
       ]);
     });
 
+    it('correctly closes the scopes of nodes that contain injected grammars', async () => {
+      await atom.packages.activatePackage('language-javascript');
+      editor.setGrammar(atom.grammars.grammarForScopeName('source.js'));
+      editor.setText('/**\n*/\n{\n}');
+
+      expectTokensToEqual(editor, [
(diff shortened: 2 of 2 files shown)


as-cii (approved):
Thanks for picking this up and pushing it over the finish line, @nathansobo! ✨ I fixed a few linter errors in a467402 to make CI green.

It seems like in the current situation we sometimes allow for a shallower grammar to emit scopes even when a deeper grammar is parked at the same position. If I understand correctly, this may lead to one of those diabolical scenarios you and @maxbrunsfeld illustrated in the body of this pull request in which we could close a tag that was never opened (or vice versa).

If that's correct, I wanted to mention a couple of observations (which we may want to address later, if ever at all) about the invariants of the injection system. I am not super familiar with this mechanism in general, so pardon my ignorance on the subject. 🙏

- I am a bit unclear on why the injected grammar is reporting a boundary at _the end_ of the injection point (`[1, 2]` in the test introduced with this pull request) even if no scopes are being reported there. Could we make the underlying iterator not report such position at all when scopes needn't be yielded? If we do so, does it mean we could avoid the extra conditional in the `HighlightIterator`'s orchestration logic?
- Alternatively, could we change the injection mechanism to _always_ report the parent's scopes at the _beginning_ and _end_ of an injection, but skip everything in between? This would seem to work well for constructs such as Markdown code-blocks or the JsDoc example in this pull request, but I am unclear on whether it would be something that works without loss of generality.

Either way, this seems like a reasonable workaround to me and I am totally cool with shipping it as it is until we gather more real-world examples to drive a more general solution. 🚢

src/tree-sitter-language-mode.js
```diff
@@ -1010,7 +1010,8 @@ class HighlightIterator {
       if (
         next.offset === first.offset &&
         next.atEnd === first.atEnd &&
-        next.depth > first.depth
+        next.depth > first.depth &&
+        next.openTags.length + next.closeTags.length > 0
```
maxbrunsfeld: Oh wow you ended up with a really tiny diff! I think this fix makes sense.

Down the road, if this type of problem occurs in other cases, we'll probably just have to make this conditional even more specific.

maxbrunsfeld (approved):
Sorry for introducing this gnarly issue. Thanks for diving in and fixing it.

nathansobo:
> Sorry for introducing this gnarly issue.

You introduced a lot of good stuff in the process so no need to apologize. Thanks for your time yesterday talking it through.

maxbrunsfeld:
Just some additional thoughts on this:

**Recap** - Right now, we "cover up" scope start events from shallower layers if a deeper layer has a scope start event at the same position, and similarly for scope end events. This PR fixes a problem where we were incorrectly covering up a scope end event that should not have been covered.

**Remaining Error Case** - It's still theoretically possible that we might incorrectly "cover up" a closing scope, leading to a similar kind of scope "spillage" that happened in #19582. It could happen in a situation like this diagram, in which curly braces represent scope boundaries:

layer-1: `a{b c}`
layer-2: `a  b{c}`

In the outer layer, layer-1, a scope contains `b c`, but in the inner layer, a scope contains just `c`. We would not cover the scope start event before `b` in layer 1, but we would cover the scope end event after `c`, because layer-2 also has a scope ending there.

**Possible Solution** - I think that to solve this (theoretical) problem, we would need to *further* restrict the criteria for when to "cover up" outer layers' scope boundaries. We should only cover up scope events if the two layers are parked on syntax nodes with the same range (i.e. `first.treeCursor.startOffset === next.treeCursor.startOffset && first.treeCursor.endOffset === next.treeCursor.endOffset`.

I think that this would guarantee that we would only cover closing scopes if we *also* covered up the opening scopes.

PR #19738, opened 2019-08-01
Avoid covering up scopes at boundaries of injection layers

maxbrunsfeld:
This is a possible fix to https://github.com/atom/atom/issues/19582#issuecomment-517091238, mostly solved by @as-cii.

@as-cii I tweaked the implementation to use the syntax Tree instead of the buffer range to detect whether an iterator is at the start of end of its injection.

My concern about using the range is that the layers' ranges can change over time as the document is edited, so if we store a `range` property, it can become stale. I think that we can achieve a similar result by checking if we're parked at the start or end of the root node of the syntax tree.

I think there are still there are theoretically cases where we might erroneously cover scope boundaries with this logic, and we should probably be even more restrictive about when we cover scope boundaries, but I can't think of a concrete case where they would occur, so I think this idea is a good way to fix the problem we're currently seeing.

### DIFF
--- a/spec/tree-sitter-language-mode-spec.js
+++ b/spec/tree-sitter-language-mode-spec.js
@@ -866,6 +866,23 @@ describe('TreeSitterLanguageMode', () => {
         ]);
       });
 
+      it('reports scopes from shallower layers when they are at the start or end of an injection', async () => {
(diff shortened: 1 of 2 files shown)


as-cii:
Thanks for polishing this up, @maxbrunsfeld! ⚡️

PR #19832, opened 2019-08-23
Recognize git worktree directories as valid git repositories

lexi-lambda:
### Identify the Bug

This PR fixes issue #8168.

### Description of the Change

The cause of the bug is an outdated implementation of `isValidGitDirectory`. The existing implementation is based on a version of `valid_repository_path` from `libgit2` that predates worktrees. This updates the implementation of `isValidGitDirectory` (and `isValidGitDirectorySync`) to reflect [the current version of `valid_repository_path`](https://github.com/libgit2/libgit2/blob/60319788928def21d2164e472a0347bcd8cdab70/src/repository.c#L190-L228), which recognizes the repository layout used by worktrees.

### Alternate Designs

Since this bug arose in the first place due to code needing to be kept in sync with `libgit2`, an alternate approach would be to find a way to use the functionality in `libgit2` directly, avoiding this issue from coming up in the future. However, that would be a relatively invasive change, and this functionality seems unlikely to change frequently, so this approach seems alright for now.

### Possible Drawbacks

As far as I am aware, the only drawback of this change is that it makes the implementation of `isValidGitDirectory` somewhat more complicated.

### Verification Process

I have both added additional tests for `isValidGitDirectory` and `isValidGitDirectorySync` and tested that this change works on a worktree directory for a real repository.

### Release Notes

- Fixed an issue that prevented directories created by `git worktree` from being recognized as git repositories.


rsese:
Thanks @lexi-lambda!  For completeness and to help whoever would review this pull request, can you update the PR body to discuss the `Alternate Designs` and `Possible Drawbacks` sections of the [template](https://raw.githubusercontent.com/atom/.github/master/.github/PULL_REQUEST_TEMPLATE/bug_fix.md)?

alissonpelizaro (approved):

PR #19921, opened 2019-09-16
fix deb package generation on armhf platform (raspbian)

aguegu:
Sorry for not following the bug template. 

It is a small tweak. There are many issues about `arm` port of Atom. Since now we got raspberry pi 4 with 4GB ram. Now we get a platform popular enough to give the porting another try.

I follow the `build from source` guideline, it almost works, if you got the right chromedriver. But when building the deb package, the deb architecture has to be set to 'armhf' to make the deb installable on raspbian.

### DIFF
--- a/script/lib/create-debian-package.js
+++ b/script/lib/create-debian-package.js
@@ -23,6 +23,8 @@ module.exports = function(packagedAppPath) {
     arch = 'amd64';
   } else if (process.arch === 'ppc') {
     arch = 'powerpc';
+  } else if (process.arch === 'arm') {
+    arch = 'armhf';
   } else {
     arch = process.arch;
   }


script/lib/create-debian-package.js
```diff
@@ -23,6 +23,8 @@ module.exports = function(packagedAppPath) {
     arch = 'amd64';
   } else if (process.arch === 'ppc') {
     arch = 'powerpc';
+  } else if (process.arch === 'arm') {
+    arch = 'armhf';
   } else {
     arch = process.arch;
   }
```
thor1704: 3344

rsese:
Thanks for the contribution!  As you've probably seen in the related issues, we don't officially support Atom on ARM - with the team's current resources and priorities, supporting Atom on more platforms isn't something we can take on.  Since we're not going to maintain changes related to running Atom on ARM, I'm going to close this out but thank you again for your interest and contribution.

aguegu:
That is okay. I will try to smooth atom porting process on arm/raspbian on my fork. Anyway, porting to arm is easier than before due to some main dependiences upgrades.

RaspberryPiFan:
@aguegu look at my fork. This is not finished, but I will tell you when it's finished

PR #19924, opened 2019-09-17
reinstall electron-mksnapshot and fix dhcp arch generation

aguegu:
Hello, Atom Team. It is me again, from #19921 . Sorry for not following the contribution guide again. 

I have work out the armv7l porting of Atom at my repo: https://github.com/aguegu/atom/tree/master/arm32v7

It is executed within Docker, So I believe many Atom lovers could repeat and verify my process and get their Atom on ARM, especially raspbian. 

Even though you officially refuse to support Arm platforms, I have to say that our current state is just one step away from successful arm porting. As shown in this PR, all I ask are:

1. reinstall `electron-mksnapshot` in `script/`, yes, the version is still `4.2.0` as the `package.json` said, but its dependences would got upgraded. `electron-download` upgrading from `4.1.0` to `4.1.1` makes a big deal, because with the later version, it can download the right chromedriver with the right arm surfix, like armv7l. 

2. correct the `arch` generation in `script/lib/create-debian-package.js`, as #19921 . But this time, the function is more accurate and universal. I believe for the main stream platforms, this function should work too, as long as `dpkg` is involved. But for now, I only changed the `else` part.

With these 2 small patches, ARM users would just be much `luckier` when trying to build Atom for their target platforms. And I believe it would not affect a thing for your officially supported platforms.

I have written my own build instruction at https://github.com/aguegu/atom/tree/master/arm32v7 . But I will not introduce the folder to the main stream. with this patch, it process mainly just like the [doc](https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#platform-linux). The tricky things are how to setup the minimum building environment and the process always take a while. But I got them all figured out. If you do not follow the `docker` way, the native way should just work too. As I said, you would get `luckier`. :)

Again, even though official support for ARM is sadly refused. I have to say that modern editor users long for their favourite editors on ARM. Because basically all your peers look like fail to achieve it too.  VSCode bin does not work, Sublime text is even more far away. Java based editors are just unbearable slow. 

Screenshot from 2019-09-18 05-45-57

For the community and be the community.

### DIFF
--- a/script/lib/create-debian-package.js
+++ b/script/lib/create-debian-package.js
@@ -24,7 +24,7 @@ module.exports = function(packagedAppPath) {
   } else if (process.arch === 'ppc') {
     arch = 'powerpc';
   } else {
-    arch = process.arch;
+    arch = spawnSync('dpkg', ['--print-architecture']).stdout.toString().trim();
   }
 
   const outputDebianPackageFilePath = path.join(
--- a/script/package-lock.json
+++ b/script/package-lock.json
@@ -2076,34 +2076,61 @@
       }
     },
     "electron-download": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/electron-download/-/electron-download-4.1.0.tgz",
-      "integrity": "sha1-v5MsdG8vh//MCdHdRy8v9rkYeEU=",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/electron-download/-/electron-download-4.1.1.tgz",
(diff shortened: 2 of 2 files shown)


lee-dohm:
We appreciate and admire your persistence. Unfortunately, for the reasons [stated before](https://github.com/atom/atom/pull/19921#issuecomment-531990071), having a build process for ARM isn't something that we are currently willing to support or maintain. Because of this, I'm going to close this PR.

Thanks for your passion for Atom!

aguegu:
It is quite sad that a small patch like this one got refused. 

If you look into this PR, I believe the mainstream would benefit from it too with broke anything. This PR does not mean I would like to see official releases on ARM as binary afterwards. It just offers some help to someone who wants to give it a try.

Meanwhile, it has been a week since I issued the PR, which means I have been using my own build ATOM on my Raspbian for a week. I have to say it works great and all the plugins that I choose work too.

PR #20013, opened 2019-10-13
Add remember-folds to deprecated packages

forivall:
### Description of the Change

This adds my "remember-folds" package to the deprecated list, as it appears that this functionality works in core as of v1.32.0 (as I could best guess from my analysis).

Following the process shown in ["how to mark a package as deprecated"](https://discuss.atom.io/t/how-to-mark-a-package-as-deprecated/34745), as indicated on the [issue on the `remember-folds` repo](https://github.com/forivall/atom-remember-folds/issues/31), this PR adds it to the deprecated list.

### Release Notes

- The `remember-folds` package is now marked deprecated.

### DIFF
--- a/script/deprecated-packages.json
+++ b/script/deprecated-packages.json
@@ -1275,6 +1275,11 @@
     "hasDeprecations": true,
     "latestHasDeprecations": true
   },
+  "remember-folds": {
+    "version": "<=1.32.0",
+    "hasAlternative": true,
+    "alternative": "core"
+  },
   "remember-session": {
     "hasAlternative": true,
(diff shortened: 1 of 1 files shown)


lee-dohm:
This isn't what we intended for the deprecated packages list. It was specifically to deprecate packages that hadn't been updated for the v1.0 release four years ago. Since your package supports v1.0 APIs, it isn't a candidate for this treatment.

My suggestion is to add a notice to the README, publish a new version so that the atom.io page displays that notice, and [archive the repo](https://help.github.com/en/articles/archiving-repositories).

PR #20061, opened 2019-10-24
Update tree-sitter to 0.15.13

maxbrunsfeld:
This PR updates the `tree-sitter` module to include some recent changes:

* https://github.com/tree-sitter/node-tree-sitter/pull/45, which makes the library compatible with Node 12, and removes uses of deprecated V8 APIs. This way, whenever we decide to upgrade Electron, Tree-sitter won't be a blocker.
* https://github.com/tree-sitter/tree-sitter/pull/334, which adds forward-compatible support for the next Tree-sitter parser ABI, which further reduces the binary sizes and static memory footprints of Tree-sitter parsers. Once a version of Atom is released with this support, I can start regenerating Tree-sitter parsers with the new ABI.


package.json
```diff
@@ -157,8 +157,7 @@
     "temp": "^0.9.0",
     "text-buffer": "13.17.0",
     "timecop": "https://www.atom.io/api/packages/timecop/versions/0.36.2/tarball",
-    "tree-sitter": "0.15.6",
-    "tree-sitter-css": "^0.13.7",
```
maxbrunsfeld: The `tree-sitter-css` module shouldn't be listed as a direct dependency here. I added that by mistake; it's only used by `language-css` module, which [depends](https://github.com/atom/language-css/blob/09c09b5b9d933b6b5a1591a3c954ec1829b2149e/package.json#L25) on it explicitly.

PR #20115, opened 2019-11-04
Bump Tree-sitter to 0.15.14

maxbrunsfeld:
This updates Tree-sitter to include this change: https://github.com/tree-sitter/tree-sitter/pull/475.


teemoney7239 (commented):
**good**

PR #20137, opened 2019-11-10
Define indent/outdent behavior for single-line selected text

Ingramz:
### Description of the Change

Text borrowed from #15096, which I had to close due to a force-push.

While working on atom/snippets#239, I realized that there isn't a properly defined behavior for selecting a segment of text on single line and indenting/outdenting it:

indent_undefined_behavior

Here is why I don't think this has been defined yet:
1. There is no spec for single-line selected text being indented via `indent()`: https://github.com/atom/atom/blob/a7736b81e3f6628f2a9e685f28f4c1df9a5e4615/spec/text-editor-spec.coffee#L3673-L3789

2. `indentSelectedRows` (which is used in `indent()`) is poorly documented:
https://github.com/atom/atom/blob/a7736b81e3f6628f2a9e685f28f4c1df9a5e4615/src/selection.coffee#L671-L679
    > If the selection spans multiple rows, indent all of them.

    It's possible to interpret this in a way that if the selection does not span across multiple rows, then nothing gets indented. [_Insert your favorite programmer-goes-to-grocery-store-joke here._](https://www.reddit.com/r/Jokes/comments/1nmkfq/a_programmer_is_going_to_the_grocery_store/)

3. The current behavior is not intuitive and unexpected for anyone coming from a different editor. With the exception of Brackets, other text editors tend to agree that selecting text and pressing tab within a single line should not indent the whole line, but insert/overwrite with a tab.

Outdent is done through `outdentSelectedRows`, which [has a spec for single line selected text](https://github.com/atom/atom/blob/2420d12e44f33f0178c8c0ef41e3a2c22cf4fb8e/spec/text-editor-spec.coffee#L4265-L4270). However it seems to be slightly mislabeled, there is actually less than one line selected. That is not a big deal and can be easily fixed, but it's possible that something is missing from here too.

At minimum I am looking to define behavior for following cases:
1. Selecting whole line and issuing `indent()`.
2. Selecting substring within line with length greater than 0 and issuing `indent()`.
3. Selecting whole line and issuing `outdentSelectedRows()`.
4. Selecting substring within line with length greater than 0 and issuing `outdentSelectedRows()`.

There might be additional cases that might need be considered, but I don't want this to be come overly ambitious, so starting with a more managable goal might be better.

The motivation for all of this comes from behavior related to text selections that come from traversing tab stops within snippets, where indenting/outdenting may lead to unexpected results:

indent_snippet1
indent_snippet2

Personally I think that outdent behavior can be kept as-is and it is sufficient to simply improve the spec and cover cases 3 and 4. Outdenting does not break out of snippet's tab stops whereas indenting does, so indicating change with a slightly more destructive (but predictable) operation should be obvious enough that it's impossible to return to snippet's tab stops.

Visual Studio Code's behavior is a suitable candidate, which indents if the whole line is selected, but inserts a tab when only part of line is selected. Outdent is handled in the same way as Atom does it now. If there aren't any objections or better ideas, I might just implement that and propose this as the final solution.

Code at the time of writing in pull request is just to get the ball rolling and does not represent the final proposed solution in any way. But it does replace everything with a tab if at most only one line has been selected.

Currently the code implements Visual Studio Code behavior where fully selected single line indents and otherwise inserts a tab. Outdent remains the same as it is now.

### Alternate Designs

Sublime text's behavior was considered, which replaces the line with a tab if the start and end cursor is on the same line. This however can be at times more inconvenient than the Visual Studio Code behavior of indenting the line instead.

### Possible Drawbacks

Change in behavior that might upset users who are already used to the less flexible behavior in place right now.

### Verification Process

Both hand-testing and automated testing of use cases that were not previously covered (selecting right amount of text and pressing <kbd>Tab</kbd>.

### Release Notes

Not finalized.

### DIFF
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -5032,7 +5032,9 @@ describe('TextEditor', () => {
         it('inserts a tab in place of the selection', () => {
           editor.setSelectedBufferRange([[0, 4], [0, 13]]);
           editor.indent();
-          expect(buffer.lineForRow(0)).toEqual(`var ${editor.getTabText()} = function () {`);
+          expect(buffer.lineForRow(0)).toEqual(
+            `var ${editor.getTabText()} = function () {`
+          );
         });
       });
 
@@ -5755,8 +5757,8 @@ describe('TextEditor', () => {
           expect(buffer.lineForRow(1)).toBe('var sort = function(items) {');
           expect(editor.getSelectedBufferRange()).toEqual([
             [1, 0],
-            [1, 30 - editor.getTabLength()]]
-          );
+            [1, 30 - editor.getTabLength()]
+          ]);
         });
       });
 
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -5018,7 +5018,7 @@ describe('TextEditor', () => {
         });
       });
 
-      describe('when a single line has been selected from start to end'), () => {
+      describe('when a single line has been selected from start to end', () => {
         it('indents the selected line', () => {
           editor.setSelectedBufferRange([[0, 0], [0, 29]]);
           const selection = editor.getLastSelection();
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -5019,7 +5019,7 @@ describe('TextEditor', () => {
       });
 
       describe('when a single line has been selected from start to end'), () => {
-        it('indents the selected line'), () => {
+        it('indents the selected line', () => {
           editor.setSelectedBufferRange([[0, 0], [0, 29]]);
           const selection = editor.getLastSelection();
           spyOn(selection, 'indentSelectedRows');
--- a/spec/text-editor-spec.js
+++ b/spec/text-editor-spec.js
@@ -5008,13 +5008,31 @@ describe('TextEditor', () => {
         });
       });
 
-      describe('when the selection is not empty', () => {
+      describe('when selection spans across multiple lines', () => {
         it('indents the selected lines', () => {
           editor.setSelectedBufferRange([[0, 0], [10, 0]]);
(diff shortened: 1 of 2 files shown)


Ingramz:
@xjArea I recreated the pull request and brought the code up-to-date.

To answer your question, as far as I know this is the best way of solving it - we propose a change and see if maintainers are fine with it.

xjArea:
@Ingramz Thank you very much for your reply. I'm sorry that it didn't pass the Atom test ? But I found that the only error is the MAC environment, and I use the Window system, so I don't care about this error😊 . 
If I want to use your improved code, how can I put them into my Atom ? I searched related files and found that they didn't exist in my folder(version by atom-x64-windows.zip
). I don't know the structure of the software. I guess it needs to be built before it can be used, but I don't know these. So, what can I do ? Just sit and waiting Atom Group solve it 😂?

Ingramz:
The build failure comes from the fact that the test in snippets package expects current atom behavior. That test is simply too tightly coupled to the core, which is fine for now. I'll see what I can do about it.

Unfortunately windows build artifacts did not get uploaded, which is why I cannot tell you to use those. Building Atom on your own system is not impossible, but can be difficult if you haven't done it before.

xjArea:
OK, although I didn't solve the problem, I am still very grateful for everything you have done. I will continue to use Atom and look forward to the day when this problem can be solved 🙏.

darangi:
@Ingramz do you mind taking a look at the failing tests? I believe these changes caused the failure, The CI is not just being flaky.

sadick254:
Thank you for your PR!

We haven't gotten a response to our failing tests request. I'm going to close this but don't hesitate to reach out if you have or find time to fix the tests, we'll be happy to reopen the PR.

Issue #20246, opened 2019-12-16
Track-pad scroll velocity effects zoom even after hit the top of the document


Strikeeaglechase:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

After scolling to top of page on atom if you press ctrl even though it apears that there is no scrolling occuing the page will suddenly zoom in or out. It apears that the zoom is being effected by the "scoll velocity" and that the scroll velocity is not being zeroed out once you hit the top or bottem of the document

### Steps to Reproduce
1. Using two fingures on a trackpad sharply guester up or down so that the page begins scrolling
2. Once the page hits either the top or the bottem simply press the "ctrl" key

**Expected behavior:**
Nothing, page does not drasticly change zoom

**Actual behavior:**
Page unexpectedly zooms way in or out

**Reproduces how often:**
100% of the time

### Versions

Atom    : 1.41.0
Electron: 4.2.7
Chrome  : 69.0.3497.128
Node    : 10.11.0

marios772:
C:\Users\zaxar\.atom\init.coffee

Issue #9434, opened 2015-11-04
Pressing Ctrl while the page is being inertially scrolled decreases the font size


Valve:
I think it maybe only reproducible on a Mac. I'm using two-finger scroll on MBP, which has this 'inertia' effect, causing the page to slowly decelerate scrolling after two fingers were already lifted from the touch pad surface. So if I press Ctrl during the inertia scrolling (after I physically stopped dragging fingers onto the touch pad surface), - the font becomes tiny as if I pressed Ctrl + <minus> multiple times.

I'm using Atom 1.1.0 on OSX 10.10.5.

50Wliu:
Also happens on Windows 10.

sylistine:
Isn't this sort of a 'if it hurts when you hit yourself, then don't hit yourself' problem? This happens on mice that have analog scroll wheels with momentum scrolling, too. If the inertia is OS level, there's probably no way for atom to deal with it.
If so, suggest closing due to intended-behaviour.

sylistine:
I take it back, as per this stackoverflow post, it might be able to detect simulated inertial scroll on mac: https://stackoverflow.com/questions/4339196/how-to-detect-disable-inertial-scrolling-in-mac-safari

YemSalat:
I am not sure how scrolling is implemented in Atom, but the way you can deal with it in a browser - is continuously check the window Y offset (e.g. on a timer) and in the onScroll event make sure that it has not changed before increasing the font size.

Issue #7884, opened 2015-07-14
zoomFontWhenCtrlScrolling causes surprising behavior when pressing Ctrl with inertial scrolling


exogen:
Due to inertial scrolling, `mousewheel` events may continue to be fired even after the user has stopped "actively" scrolling. This is expected. What may be surprising is that while this inertial scrolling is happening, pressing/holding Ctrl will behave the same as if the user is "actively" scrolling.

Try this on a Mac to see how surprising it feels: enable zoomFontWhenCtrlScrolling, then quickly swipe up/down on your trackpad in a file. Shortly after, hold Ctrl. Due to the file still being scrolled, the font size starts changing.

It seems like zoomFontWhenCtrlScrolling handling could be "smarter" about detecting whether the user intended this by detecting whether Ctrl was pressed first or if scrolling was already happening before Ctrl was pressed.

PR #20557, opened 2020-03-26
Fix Ctrl+Mousewheel triggering zoom during inertial scrolling

rdnt:

### Identify the Bug

- https://github.com/atom/atom/issues/9434 : Pressing Ctrl while the page is being inertially scrolled decreases the font size
 - https://github.com/atom/atom/issues/7884 : zoomFontWhenCtrlScrolling causes surprising behavior when pressing Ctrl with inertial scrolling
- https://github.com/atom/atom/issues/20246 : Track-pad scroll velocity effects zoom even after hit the top of the document
### Description of the Change

When a user scrolls with mousewheel using a trackpad, most OSs implement some sort of inertial scrolling, which causes additional scroll events to fire. If a user scrolls rapidly and triggers that behavior, they press Ctrl while the inertial scrolling is ongoing, *and* the option 'Zoom Font When Ctrl Scrolling' is enabled, Atom changes font size as if the user is actively holding Ctrl and scrolling with trackpad. 
There is a mousewheel event that is fired right after the user stops manually scrolling, which has a delta of 0 for both Y and X, something that should never happen under any other circumstance (if actively scrolling, increments/decrements are always at least 1px * OS zoom level).
This commit takes advantage of this additional event and sets a timer, resetting that timer after each mousewheel event unless an event takes more than 250ms to occur. Chromium seems to not fire inertial scroll events at a constant rate, but 250ms is a safe value. This value is the default on this library: https://github.com/d4nyll/lethargy

### Alternate Designs

If Chromium was giving any additional details regarding mousewheel events (if they were generated by the user or triggered as a result of inertial scroll, for example), we could use those, but providing such details is currently at the very end of their backlog.
Additionally, there is an experimental flag called 'Experimental Web Platform features' that adds two additional scroll-related events to the window, will investigate if we can use those to be more certain inertial scrolling is occuring.

### Possible Drawbacks

Nothing important that I can think of. There is a possibility the inertial scrolling events stop prematurely (before the 250ms window ends), but I don't think a user can actually start holding Ctrl *after* having let the Ctrl key *and* start scrolling again in that ~125ms timeframe. Also, sometimes (very rarely, only happened once in dozens of tests I manually performed), an inertial-scroll-caused mousewheel event is fired after that 250ms window, triggering a zoom in/out depending on the delta. I don't see this as an issue as, if this change is not implemented, that zoom would actually be always occuring.

### Verification Process

I held Ctrl and scrolled with touchpad without letting go of the touchpad, and Atom zoomed in/out as intended.
I held Ctrl and rapidly swiped up on the touchpad, triggering an initial zoom of 1-2 levels and then stopped (because the event with the 0 deltas was fired and it cancelled the rest of the zoom changes).
I also held Ctrl and pinched to zoom in/out, zooming worked as expected.

### Release Notes

Fixed an issue where font zoom level would change when pressing Ctrl key during inertial scrolling.

### DIFF
--- a/src/workspace-element.js
+++ b/src/workspace-element.js
@@ -285,9 +285,9 @@ class WorkspaceElement extends HTMLElement {
     this.timer = setTimeout(() => {
       this.inertialScroll = false;
     }, 250);
-
     if (
-      event.ctrlKey && !this.inertialScroll &&
+      event.ctrlKey &&
+      !this.inertialScroll &&
       this.config.get('editor.zoomFontWhenCtrlScrolling') &&
       event.target.closest('atom-text-editor') != null
     ) {
--- a/src/workspace-element.js
+++ b/src/workspace-element.js
@@ -278,13 +278,14 @@ class WorkspaceElement extends HTMLElement {
   }
 
   handleMousewheel(event) {
-    if (event.deltaY == 0 && event.deltaX == 0) {
+    if (event.deltaY === 0 && event.deltaX === 0) {
       this.inertialScroll = true;
     }
     clearTimeout(this.timer);
     this.timer = setTimeout(() => {
       this.inertialScroll = false;
     }, 250);
+
     if (
       event.ctrlKey && !this.inertialScroll &&
       this.config.get('editor.zoomFontWhenCtrlScrolling') &&
@@ -299,7 +300,6 @@ class WorkspaceElement extends HTMLElement {
       event.stopPropagation();
     }
   }
-
 
   handleFocus(event) {
     this.model.getActivePane().activate();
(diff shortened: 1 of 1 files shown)


SHT:
Currently investigating whether the `overscroll` or `scrollend` events could be used to detect the end of inertial scrolling.
See: https://github.com/WICG/overscroll-scrollend-events

blueedgetechno:
facing the same problem

sadick254:
@sht Do you have any plans for continuing working on this fix?

sht:
@sadick254 I couldn't find a better way to implement this, and wasn't quite satisfied with the result (lots of false positives). Feel free to close if needed.

byronigoe:
This tiny package seems like the best way to detect inertial scrolling: https://github.com/d4nyll/lethargy

PR #20599, opened 2020-04-06
:arrow_up: github

smashwilson:
GitHub package bump.


smashwilson:
Closing so that we don't introduce https://github.com/atom/github/issues/2435 into Atom nightly (oops) 😅

Issue #20668, opened 2020-04-25
Packeges failing to install on Ubuntu 20.04


MarcusAlmert:
### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Atom packages can't be installed on Ubuntu 20.04 

### Steps to Reproduce

1. Installing apm package trough atom directly or via manual install 

**Expected behavior:**

Package installs.

**Actual behavior:**

For example when installing `terminus` via `apm install terminus`
Log:
```
marcus@~$ apm install terminus
Installing terminus to /home/marcus/.atom/packages ✗
node-pty-prebuilt-multiarch@0.9.0 install /tmp/apm-install-dir-120325-178860-l8s1ro.s8njo/node_modules/terminus/node_modules/node-pty-prebuilt-multiarch
prebuild-install || node scripts/install.js
Aborted (core dumped)
Aborted (core dumped)
gyp: Call to 'node -e "require('nan')"' returned exit status 134 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/share/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/lib/configure.js:345:16)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:237:12)
gyp ERR! System Linux 5.4.0-26-generic
gyp ERR! command "/usr/share/atom/resources/app/apm/bin/node" "/usr/share/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
[...]

Issue #20585, opened 2020-04-01
Migrate Python dependency From Python 2 to Python 3


DeeDeeG:
## Summary

Python 2 is being deprecated by Debian and Ubuntu, in favor of Python 3, in a move toward not shipping Python 2 at all. ([Python 2 is already End-Of-Life'd by the Python maintainers](https://www.python.org/psf/press-release/pr20191220/) [\*](http://pyfound.blogspot.com/2019/12/python-2-sunset.html?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+PythonSoftwareFoundationNews+%28Python+Software+Foundation+News%29)). In the long term, it will help Atom stay installable/working on these distros to specify a Python 3 dependency (in addition to or instead of Python 2). Even longer-term, Python 2 will no-longer be available, so it will make sense to drop it entirely. If Python 2 gives Atom any trouble in the medium-term, it can simply be dropped.

## Motivation

Doing this keeps Atom installable and working on newer distros/operating systems.

Furthermore, making sure the build process doesn't require Python 2 makes it easy to continue building Atom once Python 2 is End-of-Life'd.

## Describe alternatives you've considered

- Un-bundle `node-gyp` and remove any other usage of Python from within this package.
- OR: Allow users to install Atom on all platforms without requiring Python on their system.

## Additional context

- According to [this comment](https://github.com/atom/atom/issues/20356#issuecomment-578379533), Python is only needed for node-gyp.
  - [Node-gyp currently works with both Python 2.7.x and Python 3.5.x or newer.](https://github.com/nodejs/node-gyp#on-unix)
- See #20406 for some prior discussion, see #20356 for lots more discussion. These are examples of the kinds of packaging troubles that might arise relating to the Python 2 EOL and transition to Python 3.
- I am personally more of a Debian/Ubuntu person myself, but this applies to all OSes and platforms.

DeeDeeG:
Info for Fedora (RPM package): I don't think we have to worry about this for Fedora/RPM-based distros.

The current RPM package doesn't even list python in its dependencies:

```
# Examining the atom 1.45.0 RPM package...
 
$ rpm -qpR atom.x86_64.rpm 
lsb-core-noarch
libXss.so.1()(64bit)
libsecret-1.so.0()(64bit)
polkit
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(CompressedFileNames) <= 3.0.4-1
```

[...]

PR #19358, opened 2019-05-17
Update .deb package dependencies

Arcanemagus:
### Requirements for Contributing a Bug Fix

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only fix an existing bug. To contribute other changes, you must use a different template. You can see all templates at https://github.com/atom/atom/tree/master/.github/PULL_REQUEST_TEMPLATE.
* The pull request must update the test suite to demonstrate the changed functionality. For guidance, please see https://flight-manual.atom.io/hacking-atom/sections/writing-specs/.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see https://github.com/atom/atom/tree/master/CONTRIBUTING.md#pull-requests.

### Identify the Bug

First reported in https://github.com/atom/ci/issues/90, it seems that we aren't reporting the correct dependencies list in our `.deb` package files.

### Description of the Change

This PR:
* Replaces the `gconf2` and `gconf-service` dependencies with `libgconf-2-4 (>= 3.2.5) | libgconf2-4`
    * Originally added in https://github.com/atom/atom/commit/f431bb6396f9f5c5c5f98fa08040dc3410896d69 as a blanket list from the `webupd8` package of Atom at the time, these were likely mistaken attempts at the correct dependency.
    * `electron-installer-debian` lists [`libgconf-2-4 | libgconf2-4`](https://github.com/electron-userland/electron-installer-debian/blob/4ad7774f4e172bf11992efe81d9d12704df693d4/src/dependencies.js#L6), we are using the slightly more restrictive version range from Chrome 64.0.3282.167-1 documented [here](https://github.com/atom/atom/pull/16812#issuecomment-367502699).
* Removes the `libudev0 | libudev1` dependency
    * This was also added as part of https://github.com/atom/atom/commit/f431bb6396f9f5c5c5f98fa08040dc3410896d69. Our current binary doesn't link against this so if it was used in the past it is no longer required.
*  Removes the `libcap2` dependency
    * This was added in response to #7066, however no verification appears to have been done that this was actually necessary, and our current binary doesn't link against this library.

### Alternate Designs

The first point there is a mistake that definitely should be requested. The latter two removals of `libudev0 | libudev1` and `libcap2` could simply be left in though as it's possible some other part requires them.

### Possible Drawbacks

[...]

PR #20670, opened 2020-04-26
Deb packaging: Support Python 3, move "python" to "Recommends:"

DeeDeeG:
### Brief Summary

Makes python an optional dependency (but still installed by default). Edit to add: Now depends on any of `python3 (>= 3.5) | python2 (>= 2.7) | python` to support both Python 2 and Python 3.

One can do `apt install atom --no-install-recommends` to skip installing `python`, if they want. If the `python`/`python2` packages aren't available (as is expected for the next upcoming Debian/Ubuntu releases), this PR will let the user install Atom.

### Identify the Bug

Follow-up to #20356 and #20406 after doing some more research. 

### Description of the Change

Move Python dependency from the `Depends:` field (meaning a hard dependency) to the `Recommends:` field (making it an optional dependency).

<details><summary>More about Recommends: in .deb packages (click to expand)</summary>

The default [behavior of `Recommends:` in a `.deb` package](https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends) is to install the Recommended packages along with any Required packages. However, when a Recommended package is **not available** in the package repositories, or the user specifies [**`--no-install-recommends`**](http://manpages.ubuntu.com/manpages/focal/en/man8/apt-get.8.html#options), `atom` will still be installed, and a little message shows that another package had been Recommended but isn't going to be installed. Example apt output:

```
$ sudo apt install atom --no-install-recommends
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Suggested packages:
  gir1.2-gnomekeyring-1.0
Recommended packages:
  python
The following NEW packages will be installed:
  atom
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
[...]
```
_[End of expanded info.]_

</details>

#### Thoughts on why this is a good approach

Making Python an optional dependency for Atom seems okay to me, because Atom actually works without any Python installed on the end-user's computer. We only need `python` in limited circumstances. (Specifically: The current release of [`apm`](https://github.com/atom/apm) needs Python 2 to compile/install certain Atom packages containing native C or C++ code. See: https://github.com/atom/atom/pull/20406#issuecomment-616162951. This is a small minority of Atom packages.)

Furthermore, the Fedora, Windows and macOS packaging of Atom do not include a Python requirement to install or run Atom. Loosening the python dependency would be in line with the packaging for those operating systems.

We won't be able to install Python 2 in upcoming Debian/Ubuntu releases. The tentative plan for Debian Bullseye and Ubuntu 20.10 is to not offer Python 2 in the package repos. (So having less than a hard dependency (if any) on Python 2 is desirable for continued installability in those upcoming distros.)

~~_This is intended to be a "best solution available" to keep Atom installable on newer Debian/Ubuntu distributions, at least until the better solution (compatibility with both Python 2 and 3) can be merged at `apm`: https://github.com/atom/apm/pull/875_~~

Edit: Python 3 support has been added! See #20703 and https://github.com/atom/apm/pull/887.

### Alternate Designs

Various degrees along the spectrum of requiring Python... from hard-requiring `python`, to dropping `python` entirely:

- Use `Requires: python`
  - (makes Python a hard dependency. Overkill IMO, based on how rarely Python is actually used by Atom. Also, brittle with respect to future Debian/Ubuntu Releases that don't plan to ship any form of Python 2.)
- _Use `Recommends: python`_
  - _(This is the proposal in the PR right now. Installs `python` by default, if available, but `python` is still entirely optional. Can be deliberately skipped with `apt install --no-install-recommends` if the user wants.)_
- Use `Suggests: python`
  - (This doesn't install `python` by default; It simply shows a message "Suggested packages: python". Users can install with `apt install atom --install-suggests`, if they want. I just found out about that flag today, so I reckon most don't know how to install the suggested packages like that. Unclear what value this would have for end-users IMO. Effectively almost the same as dropping Python altogether. Pro: Hints to the user they might want Python. Con: Most people probably don't read the message?)
- Don't specify a Python dependency at all.
  - (Seems like a missed opportunity to let the user know they might want Python. But would be in line with packaging for other OSes.)

### Possible Drawbacks

This is a fairly low-stakes topic. Atom rarely uses Python. Only when installing an Atom package with native code to be compiled with `node-gyp`. See: https://github.com/atom/atom/pull/20406#issuecomment-616162951. Any of the four approaches above would be acceptable, but I think this is the better choice among them.

_(The implementation in #20406 was somewhat flawed, because `apm` can't detect Python from the `python2` package without the symlink `/usr/bin/python` --> `/usr/bin/python2`, which can be had in the `python-is-python2` package in Ubuntu 20.04... which is the package that "Provides" the `python` package in Ubuntu 20.04. As it turned out, #20406 was unnecessary, and this PR is designed to be a better approach than was taken in that PR, having had the benefit of more research.)_

### Verification Process

I built Atom on this PR's branch, following the instructions here: https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/ ... and tested installing with/without the `--no-install-recommends` flag.

#### More Verification Process

For initial investigation, I needed quicker iteration, so I repackaged the stable Atom 1.45.0 `.deb` package using the method described here: https://github.com/atom/atom/issues/20356#issuecomment-603762624

I added `python` and a fake package I knew wasn't in the repos ("`marcopolo`") to the `Recommends:` field of `debian/control.in` to observe `apt install`'s behavior. I can confirm that `--no-install-recommends` skips installing `python` and "`marcopolo`". And I can confirm that if I try to install Atom, the fake `marcopolo` package shows in a message like so:

```
Recommended packages:
  marcopolo
```

This is only an advisory message, and I can still install Atom despite the fact that the Recommended package `marcopolo` doesn't exist. This simulates the `python` package not being available in the package repos (`python` "not existing") in upcoming Debian and Ubuntu releases.

As such, this PR should be forward-compatible with the upcoming Debian/Ubuntu.

### Release Notes

Not applicable

(This whole saga turned out to not be a user-facing issue, yet, in stable releases of Atom on stable releases of Ubuntu/Debian.)

### DIFF
--- a/resources/linux/debian/control.in
+++ b/resources/linux/debian/control.in
@@ -1,7 +1,7 @@
 Package: <%= appFileName %>
 Version: <%= version %>
-Depends: git, libgconf-2-4 (>= 3.2.5) | libgconf2-4, libgtk-3-0 (>= 3.9.10), libgcrypt11 | libgcrypt20, libnotify4, libxtst6, libnss3 (>= 2:3.22), python | python2, gvfs-bin, xdg-utils, libx11-xcb1, libxss1, libasound2 (>= 1.0.16), libxkbfile1, libcurl3 | libcurl4, policykit-1
-Recommends: lsb-release
+Depends: git, libgconf-2-4 (>= 3.2.5) | libgconf2-4, libgtk-3-0 (>= 3.9.10), libgcrypt11 | libgcrypt20, libnotify4, libxtst6, libnss3 (>= 2:3.22), gvfs-bin, xdg-utils, libx11-xcb1, libxss1, libasound2 (>= 1.0.16), libxkbfile1, libcurl3 | libcurl4, policykit-1
+Recommends: lsb-release, python
 Suggests: libsecret-1-0, gir1.2-gnomekeyring-1.0
 Section: devel
 Priority: optional
--- a/resources/linux/debian/control.in
+++ b/resources/linux/debian/control.in
@@ -1,7 +1,7 @@
 Package: <%= appFileName %>
 Version: <%= version %>
 Depends: git, libgconf-2-4 (>= 3.2.5) | libgconf2-4, libgtk-3-0 (>= 3.9.10), libgcrypt11 | libgcrypt20, libnotify4, libxtst6, libnss3 (>= 2:3.22), gvfs-bin, xdg-utils, libx11-xcb1, libxss1, libasound2 (>= 1.0.16), libxkbfile1, libcurl3 | libcurl4, policykit-1
-Recommends: lsb-release, python
+Recommends: lsb-release, python3 (>= 3.5) | python2 (>= 2.7) | python
 Suggests: libsecret-1-0, gir1.2-gnomekeyring-1.0
 Section: devel
 Priority: optional


MarcusAlmert:
This could be very likely the root of the #20668 issue.

DeeDeeG:
cc @lkashef this is basically my take on what to do with #20356 and #20406 (Python dependency situation in Debian/Ubuntu).

I think the test failure is unrelated to this PR. This PR only touches a Linux file, but the test failure is [a timeout in a Winows x64 test](https://github.visualstudio.com/Atom/_build/results?buildId=66630&view=logs&j=4fbced83-508e-5fe0-c978-5c71ec0fc506&t=0cce7676-f298-5a2d-7f62-c11c27855377&l=2429).

```
TextEditorRegistry
  .build
    it TextEditorRegistry .build constructs a TextEditor with the right parameters based on its path and text


Finished in 1177.37 seconds
2145 tests, 14175 assertions, 1 failure, 8 skipped

Error! The 'core-render-process' test step finished with a non-zero exit code
##[error]Cmd.exe exited with code '1'.
Finishing: Run tests
```

lkashef:
@DeeDeeG Thanks again for your contribution 🙇‍♂️

The CI can be flaky sometimes, you are right, it seems unrelated.

I re-ran the windows job and just to be on the safe side.

DeeDeeG:
Thank you @lkashef for looking at this so quickly!!

lkashef:
You are most welcome @DeeDeeG 🙇‍♂️ I really appreciate all the insightful research you have been doing

DeeDeeG:
For anyone finding this PR and wanting more context, I am posting this bit of the main comment on its own, for visibility:

_This is intended to be a "best solution available" to keep Atom installable on newer Debian/Ubuntu distributions, at least until the better solution (compatibility with both Python 2 and 3) can be merged at apm: atom/apm#875_

And anything I find out about making sure Atom works with Python 3 I am still posting here: #20585 (though at the moment I'm writing this, that apm PR is the blocker for continuing migrating Atom to work fully on Python 3.)

DeeDeeG:
By the way, some people do 'apt install --no-install-recommends` to save disk space, especially in Docker. So those people might skip installing Python. If they are also one of the few people installing an Atom package with native c++/c code, they might be surprised that they need to go back and do `apt install python`

The Docker hing seems unlikely, because I can't picture folks running Atom/apm in Docker? Unless as a CI test for their Atom package?? I hope those users are able to understand and fix the situation.

To build native C++/C code with `node-gyp`, users also need to have `g++` and `gcc`. And we don't specify a dependency on either of those. `node-gyp` requires a lot more things like that than Atom does over all. As mentioned above, compiling native code in Atom packages seems like not a common thing for users to have need to do.

DeeDeeG:
@lkashef tests are passing now! :tada:

DeeDeeG:
I would understand waiting to merge this for now, since it could inconvenience a few folks (make them have to update a line in a script/Dockerfile, etc.) and isn't providing a benefit yet. Will only be needed once Debian or Ubuntu make a new release.

I would still recommend reverting #20406 if this PR is not merged for a while.

DeeDeeG:
With #20703 merged, `apm` is updated to support Python 2 & 3!! :tada:

---

A bit more research:

The `python`, `python2` and `python3` packages are avilable in the following Debian / Ubuntu releases:

```
debian:
python  [  jessie  ] [  stretch  ] [  buster  ] [  bullseye  ] [  sid  ]

python3 [  jessie  ] [  stretch  ] [  buster  ] [  bullseye  ] [  sid  ]
              ^ too old in jessie (3.4.x)

python2                            [  buster  ] [  bullseye  ] [  sid  ]

---

ubuntu:
python  [  xenial  ] [  xenial-updates  ] [  bionic  ] [  eoan  ]

python3 [  xenial  ] [  bionic  ] [  bionic-updates  ] [  eoan  ] [  focal  ] [  groovy  ]

python2                                                [  eoan  ] [  focal  ] [  groovy  ]
```

(Source: https://packages.ubuntu.com and https://packages.debian.org)

Also, this is what versions of python are installed by default on jessie / xenial desktop (the oldest still-supported releases) (both come with Python 2 and 3, but the packaging is different compared to recent distros):

```
xenial:
"python3" and "python" packages installed ("python2" doesn't exist as a package, but python2 binary does, and is installed.)

/usr/bin/python and /usr/bin/python2 are symlinks to /usr/bin/python2.7, /usr/bin/python3 is a symlink to python3.5.

---

jessie:
"python3" and "python" packages installed ("python2" doesn't exist as a package, but python2 binary does, and is installed.)

/usr/bin/python and /usr/bin/python2 are symlinks to /usr/bin/python2.7, /usr/bin/python3 is a symlink to /usr/bin/python3.4.
```

In contrast to these older distros, in recent distros, `python2` exists as a package, and the plan is for `python` not to exist as a package.

I note that even these old releases of debian/ubuntu desktop depend more heavily on Python 3 than on Python 2. Doing `apt remove python3` shows it will remove more packages than doing `apt remove python` (aka Python 2).

---

Given all this, and for best backward compatibility, I suggest we depend on this:
`python3 (>= 3.5) | python2 (>= 2.7) | python`

DeeDeeG:
Options are more flexible now. Given that we can depend on Python 3, the "optional dependency" approach isn't strictly needed.

We could just do `Depends: [...] python3 | python2 | python [...]` to have Python as a hard dependency, and that would work.

---

Either way, the current `master` branch of this repo would make `.deb` files that aren't installable on Ubuntu 20.10 "Groovy Gorilla" or Debian 11 "Bullseye".

- I suggest adding Python 3 as one way to satisfy the Python dependency for Atom.
  - Done in this PR :heavy_check_mark:
- Changing Python to be an optional dependency still might be a good idea, but isn't strictly needed for installability on the next Ubunu/Debian.
  - Done in this PR :heavy_check_mark: (but happy to undo that if requested)
- Dropping the Python dependency from the `.deb` package entirely is another option. Would match the Fedora package, and the Windows/macOS packaging as well.
  - Not done in this PR at the moment. :x: 

cc @lkashef I think this is ready for another review and/or merge? Thank you!

lkashef:
Thanks @DeeDeeG for the great input.

I have a couple of questions before bringing this to the team and try to make a decision:

1. Do you think leaving [this PR](https://github.com/atom/atom/pull/20406) for the time being would have any drawbacks?
2. Even though I like the idea of making python as an optional dependency (`Recommends:`) or removing it from `.deb` completely, to avoid the hassle and let users manage their python (since it's the majority would have it by default) but now that we can support all 3 variants of python dependencies, I don't see a drawback in making a required dependency (`Depends:`), like you suggested below?
> Depends: [...] python3 | python2 | python [...]
3. I want to confirm my understanding, if having the 3 variants of python as optional or required dependencies, won't make our decision biased anymore, like you mentioned here
> So merging this is more forward-looking. There's not really a neutral choice; It's either slightly biased toward backward-compatibility or slightly biased toward forward-compatibility.]

DeeDeeG:
Sorry for the long answers, there is a tl;dr (meaning _"too long; didn't read"_, which is a Reddit-ism) summary for each answer, though. So hopefully that's more skimmable.

> 1. Do you think leaving [this PR](https://github.com/atom/atom/pull/20406) for the time being would have any drawbacks?

Two answers based on which Atom release we are talking about. On the Atom 1.47.x branch, which includes older `apm`. Older `apm` looked for versionless `python` in a rather hard-coded way. Whereas the `python2` Debian/Ubuntu package only installs a `python2` (note the "2") binary. Older `apm` can't find this `python2`, only exactly `python`.

On any release of Atom with older `apm`, depending strictly on `python` is more correct, vs `python | python2`. For atom 1.47.x, I would suggest adding newer `apm`, or reverting #20406.

(Overly-detailed side-note: To be fair, dependencies are tried left to right, so with `python | python2`, `apt` will try `python` first if neither is installed. The isue is when a user has `python2` installed, but not `python`; in which case `apt` sees `python2` is installed and considers the dependency met. This creates rare corner cases where `apm` doesn't work, which to me seems like a bad idea.)

_tl;dr given older `apm`, #20406 was a small step in the wrong direction, provided no benefit, and opened a small corner case where `apm` can fail to install a package with native code (needing `node-gyp`). It was a result of some confusion, and with hindsight should ideally have not been merged. :frowning_face: This corner case is only applicable to Atom 1.47.x at the moment. But leaving it as-is on 1.47.x seems like not the best, IMO._

---

For the upcoming Atom release with newer `apm` (queued up for Atom 1.48.x at the moment?) leaving #20406 as-is, i.e. not merging the present PR #20670, will have the downside of not being installable on Ubuntu 20.10 "Groovy Gorilla", because Ubuntu are "for real for real" removing Python 2 entirely in that release of Ubuntu. In Ubuntu 20.10, if you don't support Python 3, you don't support Python at all. (The same is planned for Debian 11 "Bullseye", but they haven't gotten around to deleting Python 2 from their package repos, so maybe they will kick that can down the road to Debian 12??)

_tl;dr: Just leaving #20406 as-is would be leaving Atom installability broken for Ubuntu 20.10, and probably for Debian 11._

Subjective note about getting out ahead of the Ubuntu 20.10/Debian 11 releases: They're not final releases yet, but there are installable ISOs of each, so some users are out there already, and I personally feel it's good to be proactive with these things. Deadlines creep up, release cycles stretch out, etc. Ubuntu comes out on a pre-announced calendar date. Good to be ahead of the curve.

> 2. Even though I like the idea of making python as an optional dependency (Recommends:) or removing it from .deb completely, to avoid the hassle and let users manage their python (since it's the majority would have it by default) but now that we can support all 3 variants of python dependencies, I don't see a drawback in making a required dependency (Depends:), like you suggested below?

Agreed. In terms of making conservative changes, the thing to do is have it be a hard dependency.

Especially now that there is no chance for the hypothetical scenario, where "there is no usable Python to install from the Ubuntu/Debian repos" -- Python 3 support solved that problem.

_It remains a little odd/arbitrary that the `.deb` (Debian/Ubuntu) package is the `only one` that requires Python. On any of the other platforms (Fedora, macOS, Windows), you can be without `git`, `python[2,3]`, `cc`/`gcc`, `c++`/`g++` and Atom will install. Only the `.deb` checks that you have what one needs to build native code packages with `apm`._

_But that's a conversation that's maybe worth opening another issue about, or just letting the status quo continue/prefer continuity rather than shake things up. The motivation here was originally bug-fixing, so I'd understand sticking to what's needed for bug fixing and potentially follow up later._

_tl;dr: Sure, just doing hard `Depends:` seems reasonable, and I am comfortable following up in another GitHub issue if need be._

> 3. I want to confirm my understanding, if having the 3 variants of python as optional or required dependencies, won't make our decision biased anymore, like you mentioned here

> So merging this is more forward-looking. There's not really a neutral choice; It's either slightly biased toward backward-compatibility or slightly biased toward forward-compatibility.]

I'm not sure what I was getting at, to be honest.

I think I'm alluding to a couple of things. One is the "rock and a hard place" situation that existed when Python 3 wasn't an option.

Rock: Having Atom being installable on the upcoming Ubuntu 20.10/Debian 11, but with no native module building support in `apm`... And making a slightly wonky/off-the-beaten-path change to get there (Using optional dependencies)... Hard place: Atom wouldn't be installable on Ubuntu 20.10/Debian 11 without these changes.

I think I was also getting into the fact that depending on the versionless "`python`" package is discouraged/deprecated as of Ubuntu Focal. So having to use that as our sole way to depend on Python was kind of "eh," and the workaround in Ubuntu's transitional `python-is-python2` package in Focal could annoy some users or cause problems.

Now that Python 3 is an option, none of these trade-offs are needed.

---

Technically there is some "bias" remaining, but  I think it's not an issue. It only happens when we need Python, but the user doesn't have ANY Python version installed. We need to pick SOME version of Python to prefer (there is no non-deterministic "just pick one" feature in `apt` dependency resolution.) So we prefer Python 3 in that narrow case.

(Most users have BOTH installed, and if someone is missing one of them, it is likely to be Python 2 as that begins to be phased out. As pointed out in my Research comment just above, even the oldest supported Debian/ Ubuntu, jessie and Xenial, respectively, include both in their desktop releases. That's why, when making this arbitrary choice, I think it's good to lean toward Python 3. On that note, the versionless `python` package is deprecated as of Ubuntu 20.04, so that should definitely be last.)

If any of `python3`, `python2` or `python` are installed, `apt` considers the dependency satisfied and doesn't need to install any new Python packages. User is already covered, no problem.

If _none_ of `python3`, `python2` or `python` is installed, `apt` will check for those from left to right. It will install the first of the three it can find in the Debian/Ubuntu repos. (See research comment just above for which packages are available in which releases of Debian/Ubuntu).

_So technically, having `python3` first is biased toward installing `python3`. But we need some version of `python` at that point, and any version will do, so I don't think it's an issue._

_If a user prefers Python 2 and doresn't want Python 3, they can install Python 2 first, and Python 3 WILL NOT be installed. Vice-versa, if they prefer Python 3 and have it installed, Python 2 WILL NOT be installed._

Having the `python3 | python2 | python` dependency will install _any_ form of Python that Debian/Ubuntu have ever shipped, to my knowledge, and they all work with newer `apm` (with newer `npm` + `node-gyp`). so the one it picks doesn't really matter.

There are some Debian/Ubuntu installs in the wild with no Python installed, such as a headless, Server oriented Ubuntu/Debian install. Or the `debian`/`ubuntu` docker images. Anything extremely minimal or without any emphasis on GUI programs. In which case... We prefer PYthon 3, but if it's not available we install Python 2. Modern defaults, works with the old stuff.

_tl;dr: It's as non-biased as we can possibly be, short of dropping our Python dependency and letting users sort out the error messages :grimacing:. We have to pick something, so slightly prefer Python 3. But works with Python 2 as well. IMO this whole situation is MUCH better now that Python 3 is an option._

DeeDeeG:
I want to note that `python2` is in the still-in-development package repos for Ubuntu 20.10 and Debian 11 right now, but the maintainers for each distro have repeatedly announced a desire to not ship Python 2 in those releases.

Doing #20406 was a result of acting based on the current state of the development branches, rather than going by announced policy, which I think was a mistake. Lesson learned, follow the announced policy. So I think it's safe (or at least safer) to say there will be no Python 2 in Ubuntu 20.10 or Debian 11.

_Which means supporting Python 3/specifying `python3` as a dependency would be required for installability on those distros._

This stuff is a lot to learn and keep straight in my head sometimes, so I apologize if I missed or mis-stated something about that in the past.

lkashef:
@DeeDeeG thanks for the prompt response, I didn't get to be through with the previous comments, but I was just discussing this with the team and @Arcanemagus just pointed me to [his PR](https://github.com/atom/atom/pull/19358#issue-280050249) and the interesting part was this:
> python This was Atom's first ever listed dependency... but with no reasoning attached. It's likely this is listed for APM. 

It made more biased towards dropping `python` as a dependency to begin with, because even we resolved this issue today, the issue will continue popping up with new releases of `python`, but since you said this:

> Which means supporting Python 3/specifying python3 as a dependency (not just python | python2) would be required for installability on those upcoming releases.

I wanted to confirm if you were referring to the case where we add `python` as a dependency and you didn't mean that we will be required to explicitly add it as a dependency.

DeeDeeG:
> I wanted to confirm if you were referring to the case where we add python as a dependency and you didn't mean that we will be required to explicitly add it as a dependency.

Correct. :+1:

DeeDeeG:
I saw the history of the `control.in` file and noticed an Ubuntu tinkerer/fan site "webupd8" had their own versoin of Atom with a bunch of loosely-needed package hard-required.

IIRC this was copy-pasted to the official Atom `control.in` file.

lkashef:
@DeeDeeG, that's a great catch and explains how it go it in there in the first place.

Based on the recent collaborations, I would consider reverting https://github.com/atom/atom/pull/20406 and dropping python as a dependency.

I know it's the only solution out of the 3 recent options, that we didn't explore as much, so I would be more comfortable, before we move forward, to open a PR summarizing the 3~4 PRs opened addressing this subject, into Alternatives and Drawbacks, to make sure we are taking the best decision from a bit of a high-level point of view.

If you would like to do that, I would be more than happy to push it through, if not, I will get to it once I have a chance for sure.

DeeDeeG:
Edited my comment, apparently come liner wanted a python dependency, presumably because there was an actual python script in the `atom/atom` repo, or else it somehow detected `node-gyp` in `apm`.

DeeDeeG:
FWIW, Atom has several lintian warnings. This is the lintian output for the current Atom 1.46.0 .deb file:

<details><summary>lintian output (click to expand.)</summary>

```console
$ lintian atom-amd64.deb 
E: atom: binary-or-shlib-defines-rpath usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git-http-fetch /tmp/build/curl/lib
E: atom: binary-or-shlib-defines-rpath usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git-http-push /tmp/build/curl/lib
E: atom: binary-or-shlib-defines-rpath usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git-imap-send /tmp/build/curl/lib
E: atom: binary-or-shlib-defines-rpath ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: atom: missing-dep-for-interpreter node => nodejs (usr/share/atom/resources/app/apm/node_modules/ajv/scripts/info) #!node
E: atom: missing-dep-for-interpreter node => nodejs (usr/share/atom/resources/app/apm/node_modules/asar/bin/asar) #!node
E: atom: missing-dep-for-interpreter node => nodejs (usr/share/atom/resources/app/apm/node_modules/coffee-script/bin/cake) #!node
E: atom: missing-dep-for-interpreter ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: atom: missing-dependency-on-libc needed by usr/share/atom/atom and 56 others
E: atom: package-installs-python-bytecode usr/share/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/MSVSUtil.pyc
E: atom: package-installs-python-bytecode usr/share/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/MSVSVersion.pyc
E: atom: package-installs-python-bytecode usr/share/atom/resources/app/apm/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.pyc
E: atom: package-installs-python-bytecode ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: atom: shlib-with-executable-bit usr/share/atom/libEGL.so 0755
E: atom: shlib-with-executable-bit usr/share/atom/libGLESv2.so 0755
E: atom: shlib-with-executable-bit usr/share/atom/libffmpeg.so 0755
E: atom: shlib-with-executable-bit ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: atom: statically-linked-binary usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git-lfs
E: atom: statically-linked-binary usr/share/atom/resources/app.asar.unpacked/node_modules/vscode-ripgrep/bin/rg
W: atom: backup-file-in-package usr/share/atom/resources/app/apm/node_modules/console-control-strings/README.md~
W: atom: binary-without-manpage usr/bin/apm
W: atom: binary-without-manpage usr/bin/atom
W: atom: description-synopsis-starts-with-article
W: atom: desktop-command-not-in-package usr/share/applications/atom.desktop env
W: atom: embedded-javascript-library usr/share/atom/resources/app/apm/node_modules/ajv/lib/compile/async.js please use libjs-async
W: atom: embedded-javascript-library usr/share/atom/resources/app/apm/node_modules/async/lib/async.js please use libjs-async
W: atom: embedded-javascript-library usr/share/atom/resources/app/apm/node_modules/asynckit/lib/async.js please use libjs-async
W: atom: embedded-javascript-library ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: atom: executable-not-elf-or-script usr/share/atom/resources/app.asar.unpacked/node_modules/github/bin/git-askpass-atom.js
W: atom: executable-not-elf-or-script usr/share/atom/resources/app.asar.unpacked/node_modules/github/bin/git-credential-atom.js
W: atom: executable-not-elf-or-script usr/share/atom/resources/app.asar.unpacked/node_modules/symbols-view/vendor/ctags-darwin
W: atom: executable-not-elf-or-script ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: atom: extended-description-line-too-long
W: atom: hardening-no-pie usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/bin/git
W: atom: hardening-no-pie usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git
W: atom: hardening-no-pie usr/share/atom/resources/app.asar.unpacked/node_modules/dugite/git/libexec/git-core/git-credential-cache
W: atom: hardening-no-pie ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: atom: hardening-no-relro usr/share/atom/atom
W: atom: hardening-no-relro usr/share/atom/libEGL.so
W: atom: hardening-no-relro usr/share/atom/libGLESv2.so
W: atom: hardening-no-relro ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: atom: package-contains-vcs-control-file usr/share/atom/resources/app/apm/node_modules/extsprintf/.gitmodules
W: atom: package-contains-vcs-control-file usr/share/atom/resources/app/apm/node_modules/npm/node_modules/extsprintf/.gitmodules
W: atom: script-not-executable usr/share/atom/resources/app/apm/node_modules/npm/bin/npx
W: atom: script-not-executable usr/share/atom/resources/app/apm/node_modules/npm/node_modules/lockfile/gen-changelog.sh
W: atom: script-not-executable usr/share/atom/resources/app/apm/node_modules/npm/scripts/release.sh
W: atom: script-not-executable ... use --no-tag-display-limit to see all (or pipe to a file/program)
N: 107 tags overridden (107 errors)
```

</details>

So, passing the python-related lint would not make or break passing `lintian` with no warns/errors.

I don't personally feel that passing one python-related lint, with all of those failing, is a compelling reason to keep the Python dependency. (Many lintian errors/warns seem to not be actual issues, e.g. not having `Depends: nodejs` to provide a `node` executable, which is referenced by the `apm` lib files, but we bundle a node binary with `apm` anyhow so it's moot.).

DeeDeeG:
> open a PR summarizing the 3~4 PRs opened addressing this subject

To make sure I understand this correctly: All PRs related to Python 3 support/Python deb dependencies?

Some are already merged e.g. (https://github.com/atom/atom/pull/20703), and one (#20406) is already shipped in Atom Beta 1.47.x, but I can summarize all of them, as a recap.

The PR diff is likely to be fairly minimal. Only this present PR and #20711 are not merged yet.

Also, I'm picturing the "summary" PR can go against master, but another PR or some cherrry-picking would be required to get an appropriate approach into the 1.47.x branch.

Would be glad to do that a bit later today.

lkashef:
@DeeDeeG I agree, maintaining `python` dependency, just to support `Lintian` doesn't seem reasonable, on top of all the warnings and things that could be solved through hacks like the `nodejs` error.

Also, while it's full of warnings, there has not been a single mention of `Lintian`, in years.

## Summary of what I have on my mind:

1. We are not sure if Atom needs `python` for something else.
2. Also Lintian needs python, so for sure we will need another PR to remove Lintian support and any dependencies that only it needs.
3. I meant all PRs related to it, including the merged ones, the goal is to have a centralized ticket:
  3.1. to contain the timeline of the discussions we had
  3.2 the reason behind the decision we want to take (drop python support)
  3.3 a test run for atom and check the .deb artifacts, to make sure Atom can live without python
  3.4 close any related PR that will not get merged (probably has long discussions like this one) and link it to the new PR, so anyone interested to learn about python support in Atom can refer to a single PR and have this new PR focused on the high-level and TL;DRs rather than thoroughly explain everything in details.
4. Probably if we this PR proved our concept/decision, we could do a separate PR to drop `Lintian` support and any remaining dependencies or files it required
5. Regarding cherry-picking and other release branches, we could create a simple TODO to make sure we covered our bases and didn't affect or will affect other releases, sort of due-diligence/clean-up, but again the new PR is just to help make a final decision, so I would say let's worry about propagating the PR after we make a final decision.

How do you feel about that?

DeeDeeG:
@lkashef Sounds good to me! I opened #20716, please let me know if anything is missing, or any feedback. Excited to see this likely coming to a (relatively) tidy conclusion.

DeeDeeG:
Addressing the things you mentioned:


> 1. We are not sure if Atom needs `python` for something else.

I would like to point out the Windows installer for Atom does not include nor require Python, and Windows does not ship with Python. Which means, I believe there are many users of Atom who don't have Python installed, and who are doing fine. I expect if Atom needed Python very often, Windows users would have been vocal about it and we would know. (I still believe it is only required for `node-gyp`, which stands out as being the one and only prominent tool in the NodeJS ecosystem written in Python.)

Also, doing a search of the code doesn't turn up any actual uses of Python directly in this repo, only indirectly via `apm` that I could find. The Python results in this repo are related to supporting code highlighting for when editing Python scripts, but that doesn't run in a Python interpreter, it's JS code that reads text files and highlights them properly as Python syntax.

I also see a metadata file that seems to inform macOS that Atom can open Python source files (for editing, not executing!!) such as `.py` `.rpy` `.cpy` and `.python` files. (`resources/mac/atom-Info.plist`)

Last but not least, the python dependency in the `.deb` package... Which is proposed to be removed in #20711 

(Also, there is a dictionary file of hundreds/thousands of common English language words, one of which is "python", at `spec/helpers/words.js`, but this isn't really a reference to the programming language at all! Just a funny coincidence.)

Specifically, I ran `git grep -i python` and `git grep --name-only -i python` in this repo, and searched "python" using the GitHub.com repo search feature. As I said, it only turned up verious things relating to but not executing Python.

> 2. Also Lintian needs python, so for sure we will need another PR to remove Lintian support and any dependencies that only it needs.

I believe (from reading #3599) Mr. Sawicki was running `lintian` manually. The repo is set up to support running `lintian` manually against the built`.deb` package, but the repo does not do so automatically. It's not part of CI, for example.

The lintian "support" just consists of having a metadata file saying "ignore the following warnings/don't print them".

Deleting these files could tidy up the repo a bit, but it's not a crucial thing. It has no practical impact, IMO. Other than being unused/a bit crufty.

> 3. I meant all PRs related to it, including the merged ones, the goal is to have a centralized ticket:
>   3.1. to contain the timeline of the discussions we had
>   3.2 the reason behind the decision we want to take (drop python support)
>   3.3 a test run for atom and check the .deb artifacts, to make sure Atom can live without python
>   3.4 close any related PR that will not get merged (probably has long discussions like this one) and link it to the new PR, so anyone interested to learn about python support in Atom can refer to a single PR and have this new PR focused on the high-level and TL;DRs rather than thoroughly explain everything in details.

I posted #20716. I just now updated the main comment so it will auto-close related issues/PRs if/when it gets merged.
> 4. Probably if we this PR proved our concept/decision, we could do a separate PR to drop Lintian support and any remaining dependencies or files it required

Sounds reasonable, and see above in response to point 2.
> 5. Regarding cherry-picking and other release branches, we could create a simple TODO to make sure we covered our bases and didn't affect or will affect other releases, sort of due-diligence/clean-up, but again the new PR is just to help make a final decision, so I would say let's worry about propagating the PR after we make a final decision.

Again, sounds reasonable to me. #20716 briefly mentions this as a TODO item.

DeeDeeG:
Actually, I did find some parts of this repo that could be updated to reflect Python 3. Here are my findings looking at all files that mention "python" (case-insensitive search):

#### Not related to actual use of the `python` interpreter/executable by Atom:

- CONTRIBUTING.md (A sentence from this documentation file refers to hypothetically editing a Python source file, as an example of context you might want to include in a bug report if that's when a bug happens.)
- docs/build-instructions/build-status.md (A line shows the build status of `language-python`, a package that "Adds syntax highlighting and snippets to Python files in Atom.")
- docs/rfcs/003-consolidate-core-packages.md (`language-python` again.)
- package.json (`language-python`)
- package-lock.json (`language-python` and `tree-sitter-python`, which are both related to Python syntax highlighting.)
- packages/README.md (`language-python` again)
- packages/one-dark-syntax/index.less
- packages/one-dark-syntax/styles/syntax/python.less
- packages/one-light-syntax/index.less
- packages/one-light-syntax/styles/syntax/python.less
- packages/solarized-dark-syntax/index.less
- packages/solarized-dark-syntax/styles/syntax/python.less
- packages/solarized-light-syntax/index.less
- packages/solarized-light-syntax/styles/syntax/python.less
  - (all of the above are fancy CSS stylesheets having to do with syntax highlighting for Python.)
- resources/mac/atom-Info.plist (Appears to inform macOS that Atom is a good program to open Python source files with... I understand this would be for editing, not running said Python files.)
- script/deprecated-packages.json (A json array listing some old deprecated packages, some of which have "python" in the name.)
- script/license-overrides.js (Some package named JSChardet, which is a "Port of python's chardet")
- spec/helpers/words.js (A long list of English-language words. One of which is the word "python", as in literally the type of snake. Not related to the programming language whatsoever. Test-related, not used in the final Atom product.)
- spec/tree-sitter-language-mode-spec.js (`language-python` and `tree-sitter-python` again. Test-related anyhow, not built/used in the final Atom product for end-users.)
- src/deprecated-syntax-selectors.js (Another list of old, deprecated packages, two of which are called `python` and `python-function`, respectively.)

#### Actually related to `python` usage:

- README.md (A line of this documentation file suggests installing the versionless `python` package on Ubuntu. Maybe this should be updated to suggest `python3` or `python2` if `python` isn't available.)
- resources/linux/debian/control.in (versionless `python` dependency in the `.deb` package for Debian/Ubuntu of Atom. Debating whether to drop this/make it optional, update it, etc. Part of #20716.)
- resources/linux/debian/lintian-overrides (This is just metadata for a linter (`lintian`) that checks for common issues with `.deb` packages. But it's very sensitive and IMO has lots of false positives/noise. No-one has been running this linter apparently, so maybe remove the metadata for it?)
- script/lib/verify-machine-requirements.js (build-time script that appears to check for Python on the system in a way that isn't `python3`-aware, or ready for the split of `python` into `python2` and `python3` on Linux. Should probably be updated.)
- script/vsts/platforms/windows.yml (vsts = "Visual Studio Team System"? aka Azure DevOps Server? This seems to be the CI configuration file for the project, specifically the configuration for the  "Windows" tests. This hard-codes that the build should use Python 2.7... maybe consider also testing on Python 3? Or updating to just use Python 3?)

darkdragon-001:
`node-gyp` should be compatible with Python 3 in the meantime: https://github.com/nodejs/node-gyp/issues/1687#issuecomment-541471076

DeeDeeG:
@darkdragon-001, yep. #20703 has bumped `apm` --> `npm` -- `node-gyp` (Now using `node-gyp` 5.1.0). All that should be available in the latest `atom-nightly`!

The main stuff is done now, All that's left is just adjusting packaging/scripts/documentation and other loose ends to be 100% updated for Python 3.

DeeDeeG:
This PR's approach is still one of two approaches I like for the Debian packaging issue. The other is in #20716. Closing this issue so discussion can happen in one place, at #20716. (Cleans up the PR queue for this project a bit also.)

PR #20703, opened 2020-05-06
Upgrade apm to 2.5.0

smashwilson:


DeeDeeG:
Hi again!

I noticed the CI failure is a simple timeout (when the CI environment tried to do `apt install` during its setup on Linux). I expect it would pass if re-started.

This change nearly makes Atom fully build-able with Python 3. Wanted to mention that (alongside updating `apm` in this repo), updating `npm` in `scripts/package[-lock].json` to 6.14.2 or later would make Atom fully build-able with Python 3! (Might be good to also adjust dependencies in the `.deb` and `.rpm` packaging. Since Debian/Ubuntu are planning to not ship Python 2 in their next releases.)

I already tried this (see [my comment](https://github.com/atom/atom/issues/20585#issuecomment-623260753) in #20585) and it worked.

smashwilson:
> I expect it would pass if re-started.

Me too, but I can't seem to log in to Azure DevOps to actually restart it 🙈 I'll give it another shot later I suppose, if nobody beats me to it!

PR #20707, opened 2020-05-07
📝Update Contributing.md

dhairya2019:
Grammatical Error 
Verb have does not agree with the subject
Changed have to has in line number 357

-----
[View rendered CONTRIBUTING.md](https://github.com/dhairya2019/atom/blob/master/CONTRIBUTING.md)


package.json
```diff
@@ -43,7 +43,7 @@
     "chart.js": "^2.3.0",
     "clear-cut": "^2.0.2",
     "coffee-script": "1.12.7",
-    "color": "^0.7.3",
+    "color": "^3.1.2",
```
aminya, line 46: You should revert this

sadick254: @dhairya2019 Update on documentation should not contain dependancy changes. The change to `package.json` should be reverted.

PR #20741, opened 2020-05-15
Enable WebView on main-process in electron options

### LINKED ISSUE #20604: Atom nightly breaks webviews (Electron 5.x API changes)
### Description
Atom nightly now uses electron 5.x, which [disables `webview` and `nodeintegration`](https://github.com/electron/electron/releases/tag/v5.0.0) by default. 

1bc06505d5cccb452724c5628f6288787f16aa44 enables the `nodeIntegration` but `webview` is still disabled by default. 
Could we have it enabled? 

cc: https://github.com/ashthespy/Atom-LaTeX/issues/244

lkashef:
Fixes https://github.com/atom/atom/issues/20604

### DIFF
--- a/src/main-process/atom-window.js
+++ b/src/main-process/atom-window.js
@@ -48,7 +48,8 @@ module.exports = class AtomWindow extends EventEmitter {
         // response to a middle-click.
         // (Ref: https://github.com/atom/atom/pull/12696#issuecomment-290496960)
         disableBlinkFeatures: 'Auxclick',
-        nodeIntegration: true
+        nodeIntegration: true,
+        webview: true
       }
     };
 


src/main-process/atom-window.js
```diff
@@ -48,7 +48,8 @@ module.exports = class AtomWindow extends EventEmitter {
         // response to a middle-click.
         // (Ref: https://github.com/atom/atom/pull/12696#issuecomment-290496960)
         disableBlinkFeatures: 'Auxclick',
-        nodeIntegration: true
+        nodeIntegration: true,
+        webview: true
```
sean-codes, line 52: Should this be `webviewTag : true`? 

I seen in the electron webview documentation:
```
webviewTag Boolean (optional) - Whether to enable the <webview> tag. Defaults to false. Note: The preload script configured for the <webview> will have node integration enabled when it is executed so you should ensure remote/untrusted content is not able to create a <webview> tag with a possibly malicious preload script. You can use the will-attach-webview event on webContents to strip away the preload script and to validate or alter the <webview>'s initial settings.
```

vivi90:
Deployment failed, see: https://github.com/atom/atom/issues/20783

lkashef:
Thanks @vivi90 for the heads up.

src/main-process/atom-window.js
lkashef: Yes, it was an oversight, corrected by this PR https://github.com/atom/atom/pull/20855

PR #20750, opened 2020-05-18
⬆️ spell-check@0.76.1

lkashef:


ashthespy:
Any chance this gets pushed as a hotfix for the current Atom?

lkashef:
@ashthespy we just released 1.47 and 1.48-beta and we've patched both releases

PR #20899, opened 2020-06-07
Faster atom.packages.getAvailablePackages

aminya:
### Description of the Change

- Removes the unnecessary directory existence check. This path is read from `fs.readdir`, and so we are sure that it exists!

- Increasing the performance of getting the list of all available atom packages

For my setup (having 209 packages), it speeds up the command by ~10ms.

### Possible Drawbacks
Nothing I can think of

### Verification Process

 I measured it using this method:
```js
const t1 = window.performance.now()
const packages = atom.packages.getAvailablePackages()
console.log(window.performance.now() - t1)

// to check the result
console.log(packages)
```

Also, by using this change, it decreased the loading time of my package [package-switch](https://github.com/aminya/package-switch) by using this custom command which implements the change: 
https://github.com/aminya/package-switch/blob/003894af9eb31bb279721bb102788194078da086/src/bundles.ts#L90

### Release Notes

- Increasing the performance of getting the list of all available atom packages

### DIFF
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -424,8 +424,7 @@ module.exports = class PackageManager {
           const packageName = path.basename(packagePath);
           if (
             !packageName.startsWith('.') &&
-            !packagesByName.has(packageName) &&
-            fs.isDirectorySync(packagePath)
+            !packagesByName.has(packageName)
           ) {
             packages.push({
               name: packageName,
--- a/src/package-manager.js
+++ b/src/package-manager.js
@@ -419,12 +419,12 @@ module.exports = class PackageManager {
 
     for (const packageDirPath of this.packageDirPaths) {
       if (fs.isDirectorySync(packageDirPath)) {
-        const packagePaths = fs
+        const packageNames = fs
           .readdirSync(packageDirPath, { withFileTypes: true })
           .filter(dirent => dirent.isDirectory())
           .map(dirent => dirent.name);
 
-        for (const packageName of packagePaths) {
+        for (const packageName of packageNames) {
           if (
             !packageName.startsWith('.') &&
             !packagesByName.has(packageName)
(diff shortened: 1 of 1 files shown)


aminya:
The macOS single test failure is totally unrelated to this PR.

UziTech:
I believe that check is for whether it is a directory or file not whether the directory exists.

I'm guessing it will fail if there is a file in the packages directory without that check.

aminya:
@UziTech I used the native `withFileTypes `in `readdir` to filter out non-folders. This is both fast and reliable.

lkashef:
Hey @aminya, thanks for the contribution 🙇‍♂️

Please check the CI failure or even better run `script/lint` locally.

src/package-manager.js
```diff
@@ -419,14 +419,17 @@ module.exports = class PackageManager {
 
     for (const packageDirPath of this.packageDirPaths) {
       if (fs.isDirectorySync(packageDirPath)) {
-        for (let packagePath of fs.readdirSync(packageDirPath)) {
-          packagePath = path.join(packageDirPath, packagePath);
-          const packageName = path.basename(packagePath);
+        const packagePaths = fs
```
sadick254, line 422:
```suggestion
        const packageNames = fs
```
I thinks the most appropriate name for this would be `packageNames`.

src/package-manager.js
```diff
@@ -419,14 +419,17 @@ module.exports = class PackageManager {
...
+          .readdirSync(packageDirPath, { withFileTypes: true })
+          .filter(dirent => dirent.isDirectory())
+          .map(dirent => dirent.name);
+
+        for (const packageName of packagePaths) {
```
aminya, line 427:
```suggestion
        for (const packageName of packageNames) {
```

sadick254 (approved):

tada369:
im new in the club , how to code is their any simple chain or blocks to learn plz help me.

DeeDeeG:
@tada369 here are some "learn programming" beginner experiences, I hope they can help you!

https://www.javascript.com/try

https://try.ruby-lang.org/

edimaliusman (commented):
Give mehh

PR #20965, opened 2020-06-21
Upgrade babel to 7

aminya:
# Edit: I am experimenting with this pull request here: https://github.com/atom-ide-community/atom/pull/67
# Edit 2: That PR is working. I need to do a few other things.

### Description of the Change

This upgrades babel to babel 7. This allows using the new babel plugins (e.g. optional chaining) just by including `"use babel"`. This also enhances the performance of babel compiling. 

By using Electron 5 preset, many things are handled automatically, and the configuration will be just right for that Electron version, which can enhance the performance. 

Also, because the previous config was using the deprecated stage-0, the proposal plugins are added based on the [deprecation](https://www.npmjs.com/package/@babel/preset-stage-0) instructions 

### Release Notes
- Updates babel to babel 7
- Uses Electron 5 preset in the default babelrc
- Replaces the deprecated stage-0 with the proposal plugins


aminya:
There is a missing module called "acorn", which does not appear in package.json. I think we need to upgrade Eslint as well to fix this issue.

aminya:
This is blocked by https://github.com/babel/babel/issues/11757.

aminya:
Now this is blocked by another electron-link issue: https://github.com/atom/electron-link/issues/2

aminya:
Closed in favor of https://github.com/atom/atom/pull/20965

Issue #4885, opened 2015-01-07
File replacement does not work against directories


elifoster:
When saving a new file, if you decide to save it as the same name of a directory (no extension), it will ask if you would like to replace it. If you choose yes, it will error saying that that file is already a directory 

I am running Atom 0.165.0 for OSX.

There is no log in the debug console to provide.

There are basically three solutions to this:
1. Have it error after the "replace" button is clicked but before the dialog box is closed
2. Have it replace the directory
3. Have it handle directories and files separately (I'm not sure about other operating systems, but I'm pretty sure on OSX it considers them <name>.folder)

elifoster:
This is still kind of an issue. The error it outputs is different now: `Unable to save file: EISDIR: illegal operation on a directory, open 'path/to/folder/being/replaced'`

Tested on Atom 1.17.2.

50Wliu:
I am unable to reproduce this on Windows. @elifoster can you describe the exact steps you're taking to save the file?

elifoster:
1. Create new folder, name it something
2. Create new blank untitled file
3. Save as whatever you named the folder
4. Select "Replace"

50Wliu:
> Create new folder, name it something

How are you creating this folder?  Outside of Atom?  Using the `tree-view:add-folder` command?  Or some other way?

> Save as whatever you named the folder

Similarly, how are you saving?  `cmd/ctrl-s`?

elifoster:
I'm creating a folder by right clicking the project section(?) and selecting new folder.

I'm saving using cmd s, yes.

50Wliu:
Thanks for the info.  I'm still unable to reproduce this on Windows.  Here are the steps I followed, please tell me if I did something wrong.
1. `tree-view:add-folder`, enter `test` as the folder name
2. `application:new-file`
3. `core:save`
4. Navigate to the directory that has the `test` folder in it
5. Type `test` as the file name
6. Click save

What happens here is that Windows' file dialog intelligently realizes that the file is the same name as the directory and automatically cd's into the folder for you.  Meaning that once I hit save, I am now inside the `test` directory and have to choose a different filename.

Perhaps things are different on macOS and/or Linux: /cc @ungb for reproduction on those platforms.

rsese:
I tried on macOS 10.12.5 following @50Wliu's steps and I can reproduce there with Atom 1.17.2- at step 6 I get:

[...]

PR #20981, opened 2020-06-25
open file with conflict name test using chai

duochen13:
- Description:
While opening a file with the name same as the existing directory's name, it will throw an EISDIR error, this PR is about a test suite that check the file is opened properly.

- Additional Notes: 
 **EISDIR** error, an error from '**fs**' package in node.js, specifically means 'try to open a file but the path give is a directory''
This PR is related to issue[#4885](https://github.com/atom/atom/issues/4885)

### DIFF
--- a/spec/test-file-spec.js
+++ b/spec/test-file-spec.js
@@ -1,0 +1,24 @@
+const chai = require('chai')
+const expect = chai.expect
+const fs = require('fs');
+const assert = require('assert');
+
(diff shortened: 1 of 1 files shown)


sadick254:
Thank you for opening PR. 

This has been resolve in version `1.50.0`. See https://github.com/atom/atom/issues/4885#issuecomment-680794430

PR #21023, opened 2020-07-04
Makes code-sign optional

### LINKED ISSUE #8: Sign built Atom packages, or delete CI steps requiring code signing
CI is failing on a step that tries to sign the built packages:

- Windows: https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=21&view=logs&j=4fbced83-508e-5fe0-c978-5c71ec0fc506&t=8ab1786f-98fc-5bc4-a8d0-c08e0675c13a&l=4873
  - `Error: The protocol "null" is not supported, cannot load "$(ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL)"`
- macOS: https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=15&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=daf62785-c7df-523f-33d8-78a3cf89d33a&l=4863
  - `Error: The protocol "null" is not supported, cannot load "$(ATOM_MAC_CODE_SIGNING_CERT_DOWNLOAD_URL)"`

We should set a secret "code signing cert" download URL in the env variables of our CI so that these steps can pass... Or edit the code so we don't even try to sign the code.

<details><summary>Thoughts on security and keeping the signing cert secret, if we get one (click to expand):</summary>

(We need to keep this secret only amongst maintainers, or else anyone can impersonate our digital "signature". Not sure if this is a big deal after all, because what are the stakes to impersonating us? But yeah, we should keep it secret in case we find out the stakes were higher than we thoughts, e.g. security problems for our users).

(Not signing might break things on macOS? I know they started expecting code to be signed by a known developer over the past few years. Then again, playing ball with Apple usually costs money/involves having an Apple Developer account, so I'm not sure whether or not this is doable for free.)

</details>

I'd like to try to figure out (or just ask) what format or standard this signing certificate takes at upstream, so we can make one for Atom Community and properly sign ours.

aminya:
### Identify the Bug
Fixes https://github.com/atom-ide-community/atom/issues/8

### Description of the Change

Adds a `SHOULD_SIGN` env variable which allows skipping code signing when you are on the release branch. This allows building Atom without purchasing a code signing certificate. It is meant to facilitate the process of testing atom in the forks (particularly atom-community/atom)

## Possible Drawbacks
Requires setting `SHOULD_SIGN=true` as an environment variable inside Azure.

## Release Notes
N/A; this is a developer-facing change, not an end-user facing change.

### DIFF
--- a/script/vsts/get-release-version.js
+++ b/script/vsts/get-release-version.js
@@ -78,6 +78,7 @@ async function getReleaseVersion() {
       buildBranch.startsWith('electron-') ||
       (buildBranch === 'master' &&
         !process.env.SYSTEM_PULLREQUEST_PULLREQUESTNUMBER));
+  const SHOULD_SIGN = process.env.SHOULD_SIGN;
 
   console.log(
     `##vso[task.setvariable variable=AppName;isOutput=true]${getAppName(
@@ -89,6 +90,9 @@ async function getReleaseVersion() {
   );
   console.log(
     `##vso[task.setvariable variable=IsSignedZipBranch;isOutput=true]${isSignedZipBranch}`
+  );
+  console.log(
+    `##vso[task.setvariable variable=SHOULD_SIGN;isOutput=true]${SHOULD_SIGN}`
   );
 }
 
--- a/script/vsts/platforms/macos.yml
+++ b/script/vsts/platforms/macos.yml
@@ -47,7 +47,7 @@ jobs:
         displayName: Run linter
 
       - script: |
-          if [ $IS_RELEASE_BRANCH == "true" ] || [ $IS_SIGNED_ZIP_BRANCH == "true" ]; then
+          if [ $SHOULD_SIGN == "true" ] && ([ $IS_RELEASE_BRANCH == "true" ] || [ $IS_SIGNED_ZIP_BRANCH == "true" ]); then
             script/build --no-bootstrap --code-sign --compress-artifacts
           else
             script/build --no-bootstrap --compress-artifacts
--- a/script/vsts/platforms/windows.yml
+++ b/script/vsts/platforms/windows.yml
@@ -101,11 +101,19 @@ jobs:
           SET SQUIRREL_TEMP=C:\tmp
           IF [%IS_RELEASE_BRANCH%]==[true] (
             ECHO Creating production artifacts for release branch %BUILD_SOURCEBRANCHNAME%
-            node script\vsts\windows-run.js script\build.cmd --no-bootstrap --code-sign --compress-artifacts --create-windows-installer
+            IF [%SHOULD_SIGN%]==[true] (
+             node script\vsts\windows-run.js script\build.cmd --no-bootstrap --code-sign --compress-artifacts --create-windows-installer
+            ) ELSE (
+             node script\vsts\windows-run.js script\build.cmd --no-bootstrap --compress-artifacts --create-windows-installer
+            )
           ) ELSE (
             IF [%IS_SIGNED_ZIP_BRANCH%]==[true] (
               ECHO Creating signed CI artifacts for branch %BUILD_SOURCEBRANCHNAME%
+             IF [%SHOULD_SIGN%]==[true] (
               node script\vsts\windows-run.js script\build.cmd --no-bootstrap --code-sign --compress-artifacts
+             ) ELSE (
+              node script\vsts\windows-run.js script\build.cmd --no-bootstrap --compress-artifacts
+             )
             ) ELSE (
               ECHO Pull request build, no code signing will be performed
               node script\vsts\windows-run.js script\build.cmd --no-bootstrap --compress-artifacts


DeeDeeG (approved):
This change makes it possible for forks of Atom to set up and run the CI, even if they do not have/cannot afford to buy code signing certificates.

DeeDeeG:
A smaller change with similar effect would be to skip notarization when the required env vars are blank (just as code signing is already skipped when the env vars are blank).

 Just add this snippet at the beginning of the main function of `script/lib/notarize-on-mac.js`:

```js
  if (
    !process.env.ATOM_MAC_CODE_SIGNING_CERT_DOWNLOAD_URL &&
    !process.env.ATOM_MAC_CODE_SIGNING_CERT_PATH
  ) {
    console.log(
      'Skipping notarization because the ATOM_MAC_CODE_SIGNING_CERT_DOWNLOAD_URL environment variable is not defined'
        .gray
    );
    return;
  }
```

The snippet is copied from `script/lib/code-sign-on-mac.js`: https://github.com/atom/atom/blob/master/script/lib/code-sign-on-mac.js#L16-L25 (with a tweaked string of what is skipped, notarization instead of code signing.)

This makes natural sense, because per [Apple documentation](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution), notarization is for after you successfully sign your code.

If code signing is skipped, notarization should necessarily be skipped or it won't work anyhow.

---

One downside: in my testing, it requires setting all four of these vars as blank in Azure DevOps UI if you want to skip code signing/notarization on both Windows and macOS:

- `ATOM_MAC_CODE_SIGNING_CERT_DOWNLOAD_URL`
- `ATOM_MAC_CODE_SIGNING_CERT_PATH`
- `ATOM_WIN_CODE_SIGNING_CERT_DOWNLOAD_URL`
- `ATOM_WIN_CODE_SIGNING_CERT_PATH`

sadick254 (commented):
Looks great.
I think we could set the environment variable on the CI yaml files. Like we do here
https://github.com/atom/atom/blob/d9d4e3756307dca5999b3099cf797b3859d0cb8d/script/vsts/platforms/macos.yml#L53-L63

DeeDeeG:
@aminya I don't know your thoughts on this, but last I remember we just found a way to do this kind of skipping automatically, no new env vars needed.

https://github.com/atom-ide-community/atom/pull/99/files

I think we should be able to do that for this PR (skipping code signing) as well.

@aminya @sadick254 if you don't mind I'm going to code up an implementation with that design.

aminya:
> Looks great.
> I think we could set the environment variable on the CI yaml files. Like we do here
> https://github.com/atom/atom/blob/d9d4e3756307dca5999b3099cf797b3859d0cb8d/script/vsts/platforms/macos.yml#L53-L63

@sadick254 You should add the variable in the Azure UI.

>  I don't know your thoughts on this, but last I remember we just found a way to do this kind of skipping automatically, no new env vars needed.

No, this variable acts as a kill switch. It is supposed to be independent of the provided code sign variables.

DeeDeeG:
To clarify: I mean to propose an alternate solution. (The one we found for the API keys.)

Apologize for bringing it up last minute, but we only discovered/mastered this technique a few days ago.

sadick254:
@aminya Any reason you prefer having it added in the Azure UI?

aminya:
> @aminya Any reason you prefer having it added using the Azure UI?

The point of this variable is to be like an environment variable you pass to the program, not hardcoding it in the code. The reason we made this was to allow others to skip this by adding that variable to Azure UI.

If you hardcode this here, other people need to change it, which brings conflicts.

DeeDeeG:
It would be inconvenient for forks to have to make a commit changing this variable.

That means they have to test on one branch (with changed variable) so that their CI works. Then rebase to another branch without the changed variable, to avoid sending the changed variable as part of their PR.

DeeDeeG:
@aminya @sadick254 by the way the alternative I mentioned is working:
- successfully skips code signing when vars not set: https://dev.azure.com/DeeDeeG/b/_build/results?buildId=608&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=b808d44d-d1c2-5399-c2e1-8c4b890e97af&l=4955
- Doesn't skip when credentials/URL vars set: https://dev.azure.com/DeeDeeG/b/_build/results?buildId=609&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=b808d44d-d1c2-5399-c2e1-8c4b890e97af&l=4955 (CI failure is expected here because I used fake/placeholder credentials.)

sadick254:
@DeeDeeG Thank you for the alternative implementation. Your solution is great, the only downside is it uses a workaround for the unset env vars.
I prefer us being more explicit by setting the `SHOULD_SIGN` env var.

sadick254 (approved):

DeeDeeG:
Fair enough, works for me. 👍 Thanks for the mini-review, haha.

Issue #21063, opened 2020-07-10
Cleaner build process


Aerijo:
## Summary

The build process has no structure right now, it would be nice to add some structure to it.

## Motivation

I'm trying to build Atom from source on Arch Linux. It's broken somewhere, but the build is opaque (no progress indication when installing modules) and hard to follow in general. Working out what step it's up to is not easy.

I would like to rewrite it to be built from "Task" components. These Tasks would provide a uniform interface to a chunk of "work" (mainly the existing exported functions in `/lib`). This interface would include
- A task name to be printed when it starts
- A set of log methods (subtask for when it's doing different things in the task, plus info / warn / error)

Making everything conform to the Task interface would provide a more uniform output, as well as make tweaking the output easier (e.g., formatting and log levels).

## Describe alternatives you've considered

There are npm packages that do something like this, but the idea is to be simple enough to just include in the `scripts` folder. It's important that the solution works before bootstrapping. 

## Additional context

I've started working on a concept for it [here](https://github.com/Aerijo/atom/tree/clean-build). It defines a `Task` class that all other tasks inherit from.

I experimented with support for hierarchical & concurrent execution of tasks, as well as an event based life cycle (see the`finished` type methods), but these things increase the complexity and would make it a burden to implement and maintain. I think the current design is non intrusive, while still providing a good base for structuring the build.

That branch also bumps the dependencies in `script` and tries to reorganise the build script to only load modules when they are used. Hopefully I'll be able to start testing and making PR's for this once I get the build working.

therobgirl:
Omg stop it 🤪

therobgirl:
##I love ATOM @

PR #12410, opened 2016-08-12
Replace Grunt-based build system with simple scripts

nathansobo:
Our existing Grunt-based build system is really difficult to follow, so this PR replaces it with a simpler approach involving plain old JavaScript.

The `build` directory is deleted, and instead the `script` directory takes on an expanded role, getting its own `package.json` and `lib` directories to support an overhauled version of `script/bootstrap`, `script/build`, and `script/test`. Instead of relying on Grunt, we do everything via functions defined in the `script/lib` directory.
### Build script

The _basic_ structure of the new build script is as follows:
- Copy all sources to `atom/out/app`
- Transpile Babel, CoffeeScript, and Less files in place
- Generate pre-cached metadata
- At this point, `out/app` is a valid Electron app that doesn't require transpilation support to run.
- Run [`electron-packager`](https://github.com/electron-userland/electron-packager) on `out/app` to package it into a distributable application on the current platform.
- Build installer on Windows, code sign on macOS
- Zip stuff up

This is all designed to run only on third-party build providers (CircleCI for macOS, Travis for Linux, and AppVeyor on Windows). We've put zero effort into making these scripts run on internal Janky boxes.
### Release publishing plan

We deliberately omit the artifact uploading step, because our plan is to orchestrate release publishing outside of the build. We'll rely on the build services to store artifacts associated with all builds, and when we're ready to release we'll have a script download these artifacts from Circle, Travis, and AppVeyor and upload them to the releases API and S3. This will help to ensure that all artifacts stay in sync based on a specific commit SHA.
### Tasks
- [x] Build scripts working on all platforms
  - [x] macOS
    - [x] Build a complete electron app
    - [x] Code sign
    - [x] Green build on Circle
    - [x] Implement `--install` flag
  - [x] Windows
    - [x] Build a complete electron app
    - [x] Build an installer
    - [x] Delete unnecessary .nupkg delta file uploaded as artifact
    - [x] Code sign
    - [x] Green build on AppVeyor
    - [x] Installing won't be supported, because users can launch the generated installer.
  - [x] Linux
    - [x] Build a complete electron app 
    - [x] Generate `.deb` package
    - [x] Generate `.rpm` package
    - [x] Green build on Travis
    - [x] Make Travis artifacts downloadable
    - [x] Installing won't be supported, because users can launch the package manager on the generated packages.
- [x] Add `script/lint`
- [x] Release publishing
  - [x] Visualize build status for release branch across all platforms
  - [x] Download and publish artifacts for a specific SHA
[...]

PR #21067, opened 2020-07-12
WIP: Structure build scripts

Aerijo:
**NOTE: This PR includes some unrelated changes right now. I'm looking for feedback on if this direction makes sense / any changes to make.**

#21063

### Description of the Change

This introduces the concept of a "Task" to the build scripts. The point is to be simple enough for a single file implementation, but leave room to tweak output without having to edit every lib file (e.g., add log level filters, change the style based on message kind, etc.).

It is trivial to use; a function `taskify` is provided, which takes a Task name and arbitrary function. This allows existing methods to be turned into tasks with no effort, and makes defining "anonymous" tasks simple. Task bodies may be arbitrary, and subtasks may be used wherever. A helper method `subtask` is provided to set the indentation of the subtask properly. This lets you easily compose tasks and keep a hierarchical output.

A Task always returns a promise resolving to whatever the Task wants to return (it uses `async` function semantics when returning non-promises vs promises, i.e., non-promises are automatically wrapped in one). 

The Task implementation itself is heavily commented.

Below is an example of the `bootstrap` script as a Task
```
const {taskify} = require("../lib/task");

const bootstrapTask = taskify("Bootstrap", async function () {
  await this.subtask(require("./verify-machine-requirements"));
  if (process.argv.includes("--clean")) {
    await this.subtask(require("./clean"));
  }
  await this.subtask(require("./install-script-dependencies"));
});

module.exports = bootstrapTask;
```

Below is an example output using the current style. `>` indicates running a task, `>>` is an update notification on what the task is working on, `-` is information (see it's use on printing component versions), and child tasks automatically get indented appropriately.
```
> Running Build
  > Running Generate binaries
    > Running Check chromedriver version
      - chromedriverVer: target=^9.0.0, actual=9.0.0
      - mksnapshotVer: target=^9.0.2, actual=9.0.2
    > Running Clean ouptut directory
      >> Cleaning /data/github/atom/out
      >> Cleaning /data/github/atom/docs/output
    > Running Copy assets
      >> Copying 14 assets to /data/github/atom/out/app
      >> Running copy pass
    > Running Transpile packages with custom transpiler
      >> Transpiling packages with custom transpiler configurations in /data/github/atom/out/app
      >> Transpiling package github
    > Running Transpile Babel paths
      >> Transpiling 780 Babel paths in /data/github/atom/out/app
    > Running Transpile CoffeeScript paths
      >> Transpiling 80 CoffeeScript paths in /data/github/atom/out/app
    > Running Transpile CSON paths
      >> Transpiling 194 CSON paths in /data/github/atom/out/app
    > Running Transpile PEG.js paths
      >> Transpiling 1 PEG.js paths in /data/github/atom/out/app
    > Running Generate module cache
      >> Generating module cache for /data/github/atom/out/app
    > Running Prebuild LESS cache
      >> Generating pre-built less cache in /data/github/atom/out/app/less-compile-cache
    > Running Generate metadata
      >> Generating metadata for /data/github/atom/out/app/package.json
    > Running Generate API docs
      >> Generating API docs at /data/github/atom/docs/output/atom-api.json
    > Running Dump symbols
      >> Dumping symbols in /data/github/atom/out/symbols
  > Running Package application
    >> Running electron-packager on /data/github/atom/out/app with app name "atom"
Packaging app for platform linux x64 using electron v5.0.13
    >> Changing permissions for node files in /data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64
    >> Copying non-ASAR resources to /data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64/resources
    >> Writing LICENSE.md to /data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64/resources
    - Application bundle created at /data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64
  > Running Generate startup snapshot

Minifying startup script
    >> Verifying if snapshot can be executed via `mksnapshot`
    >> Generating startup blob with mksnapshot
    >> Moving generated startup blob into "/data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64/v8_context_snapshot.bin"
    >> Moving generated startup blob into "/data/github/atom/out/atom-dev-1.50.0-dev-e5ebf3ad9-amd64/snapshot_blob.bin"
  - Skipping creating debian package. Specify the --create-debian-package option to create it.
  - Skipping creating rpm package. Specify the --create-rpm-package option to create it.
  - Skipping artifacts compression. Specify the --compress-artifacts option to compress Atom binaries (and symbols on macOS)
  - Skipping installation. Specify the --install option to install Atom
``` 

### Alternate Designs

- Alternative task runners
  - I imagine it's like Gulp / Grunt but less powerful. But it is intended to work before bootstrapping, so using Gulp / Grunt would be a problem. As a single file with no dependencies, it is also more robust than introducing another dependency.

- Always return a promise:
  - Synchronous is difficult to do when the task body is async. I considered allowing the return value to be totally determined by the Task return value, but I don't know if that could cause confusion about which tasks execute sync / async.

- Currently, there is only a single task instance. This could lead to weird things if a task has itself as an ancestor.
  - This came about because I didn't want `(new MyTask()).start()` everywhere, but it should be possible to automatically build a new Task object every time `start` is invoked.

- Error handling
  - Haven't looked into this yet. I've defined an `error` message, but the current implementation normally just throws any errors.

### Possible Drawbacks

Adds a framework to the build scripts, and I don't like having to learn a whole new technology just to do / understand something. But I tried to keep it minimal for this reason, so I don't think this is a big concern compared to the tools it provides for extending and composing functionality. You could think of a task as an async function if you like (and write it like one too, using `taskify`).

Also doesn't handle child process output (see how some lines aren't indented in the above example), which is unfortunate. But I don't know if it's even possible to properly do this without an entire terminal emulator because some processes will use fancy ANSI escape sequences to move the cursor around. I consider this acceptable though, as child process output is valuable.

### Verification Process

Some commands to use for the current state. Note that the functionailty has been adjusted, where `bootstrap` is now doing the minimal necessary to bootstrap the script files, and installing apm + atom dependencies has been moved to `build` (with activation configurable with `--no-deps` flag). 
```
script/bootstrap
script/bootstrap --clean
script/build
script/clean
script/build --bootstrap --no-deps

```

### Release Notes
NA

-----
[View rendered apm/README.md](https://github.com/Aerijo/atom/blob/build-at-all/apm/README.md)


aminya:
@Aerijo This looks pretty awesome, but it should be broken into smaller pull requests. Could you separate the parts that are not dependant on each other?

aminya (commented):
@Aerijo This looks pretty good, but it should be broken into smaller pull requests. Could you separate the parts that are not dependant on each other? 

How does this affect building performance?

script/bootstrap-old
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
```
aminya, line 1: Instead of renaming the files to `bootstrap-old`, please remove the old files instead. This way the GitHub diff shows the difference between the files.

script/lib/task.js
```diff
@@ -0,0 +1,138 @@
+/**
```
aminya, line 1: Can't we make this an npm package? This allows more flexibility for updating it.

script/lib/task.js
```diff
@@ -0,0 +1,138 @@
...
+
+  /** Indicate that an unrecoverable problem occurred. See `canFail` to indicate if this should propagate to the parent Task (as an exception), or silently resolve this Task. */
+  error(msg) {
+    this.print(msg, "  !!!");
+  }
+
+  /** Log the current work being done by this Task. Typically for work that is not a Task itself. */
+  update(msg) {
+    this.print(msg, "  >>");
+  }
```
aminya, line 109: It would be good to use Azure syntax for some of these
https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands

Aerijo:
@aminya This is a draft, more of a proof of concept than an actual PR. I'm looking for feedback on if it's something that would even be considered, and what changes might be needed. The original issue got the triaged label (and now enhancement I see), which is is nice first step, but it will take a lot more back and forth on the design before it's good to be properly implemented. I'm not really concerned about splitting into components / git diffs / old files right now.

For starters, the points in Alternate Designs need to be resolved (besides the 1st I guess). 
- I'm thinking the "always async" requirement may be unnecessary. There's no real benefit to enforcing a consistent return type.
- Is it possible for some task (now or in future) to be run recursively? How would the design accommodate this (after all, there's no point if conforming to the task interface is too restrictive compared to the current "free style").
- Exception handling: do tasks handle errors by default, or leave it to whoever is writing the task? What does error handling look like?

And more:
- The `name` and `indent` properties should be placed behind a single `task` or similar name member. This minimises polluting the object properties, and allows more to be added in future without risking collisions. This is less important for the logging methods, as they are meant to be easily accessed.

And the solutions to these problems need to be simple enough for anyone to understand and to _never_ cause unexpected problems. I've had enough bad experiences building Atom (some of the unrelated changes in this PR are there just so I could build at all and test the thing), I would not have the build script tool itself be the cause of more.

If a design is finalised, I would make new PRs with the changes I intend to get merged. The first could just introduce the task building tool. Other PRs could then gradually / all at once convert to using it.

script/lib/task.js
Aerijo: Should be fine. The current values are more like placeholders, just to get the pretty output. Looks like sub tasks could use the `##[group]` syntax too.

aminya:
You should also take a look at this https://github.com/atom/atom/pull/12410 which replaced the grunt based build system with the scripts we have now.

sadick254:
Really looking forward to this @Aerijo. Am happy to help if needed.

aminya (changes requested):
In my opinion, we should go towards this path, only if it has a performance benefit. 

For example, I was looking into using Node worker threads to run independent tasks in parallel. I think the [threads](https://threads.js.org/) is a very good package.

In https://github.com/atom/atom/pull/12410, Gulp based system was removed because of its complexity and dependency issues. It did not have a real performance benefit.

However, if we can use Node workers to run things in parallel (like transpiling Babel code in parallel to Less file transpilation), we can speed up the entire process.

Aerijo:
I find structure in the build inherently valuable, so long as it doesn't noticeably _decrease_ build times. Otherwise you either do an ad hoc version of it by console logging or worse. With structure, specifically logging methods, you can control log levels (info, error, verbose, etc.) in  a centralised way, and enforce things like every task having a name. And you can add prefixes, so you can easily identify the screen lines of a given message (when some lines get wrapped, like the dependencies fingerprint, it is difficult to tell which lines are new entries and which are continuations). 

In fact, if this (or similar) gets merged, my next plan is to add _a lot_ more verbose logging methods to the scripts. Think things like printing version numbers, bin paths, etc. This would be hidden by default, but ready to go if you ever need a more in depth view of the build.

Regarding #12410, I wasn't around at the time so don't know the thought behind it, but it seems backwards to me. I would think plain JS be the most difficult to understand, as it doesn't have any consistency enforced on it.

Parallel tasks is definitely a good thing for build performance, and I don't know if the task design in this PR would be enough. This is where it might be good to explore a proper build tool. It might be possible to detect and just prepend the task ancestors to the log methods, but child process output is more difficult.

An alternative design is to pass a task object to the function (instead of extending a task object). This would remove any chance of conflict. I'm giving that a trial in #21288.

aminya:
The solution you are offering for solving this logging problem adds complexity while you could easily insert `console.log` at the top of the files and functions. #12410 happened because writing/reading Vanilla JavaScript does not require any prior knowledge of any framework, and working with them easier.

For example, see this PR that I removed Grunt for simple npm scripts. If you check you see how many dependencies and boilerplate code Grunt with itself (without improving the performance)
https://github.com/atom/apm/pull/903

My big problem with bootstrapping and building Atom is the time it takes. I never sit and stare at my terminal for 10 min. If we use threads to reduce this time, then the structured log messages will make more sense.

If you can integrate [parallelcode](https://github.com/manuel-di-iorio/parallelcode) into this, then this pull request will be more beneficial.

Aerijo:
> The solution you are offering for solving this logging problem adds complexity while you could easily insert console.log at the top of the files and functions.

No, `console.log` is insufficient. It does not offer ways to filter output based on log level, there is no way to attribute the output to a specific task (more important for parallel builds), it doesn't support customising the prefix of messages to best suit the logging environment, timing the length of each stage, etc.. As soon as you add these things, you roughly get #21288 (this PR's successor).

> #12410 happened because writing/reading Vanilla JavaScript does not require any prior knowledge of any framework, and working with them is easier.

I was saying I disagreed about the benefits (with respect to build transparency) because everything becomes ad hoc. Writing them becomes easier, but you are reading the output of them most of the time. I personally never learned Grunt though, so don't know if it supported the kind of logging I'm looking for. #21288 shows a clearer version of what I want, that is more focused on logging and is less intrusive to the build scripts. I don't think I'll continue with the approach in this PR which also tries to handle the lifecycle of the Task, like catching errors, skipping steps, etc. The #21288 version is strictly for logging, and leaves the control flow alone.

> For example, see this PR that I removed Grunt for simple npm scripts.

That's not really relevant here. These build scripts weren't simple before then, and they aren't simple now. Maybe they could be rewritten to be simpler, but as is the most an npm script would do is call the existing `script/foo` script anyway. That's perhaps nice for familiarity, but not really in the scope of this PR.

> My big problem with bootstrapping and building Atom is the time it takes. I never sit and stare at my terminal for 10 min. If we use threads to reduce this time, then the structured log messages will make more sense.

> If you can integrate parallelcode into this, then this pull request will be more beneficial, and I will agree with the added complexity.

I think you misunderstand the purpose of this PR. It is not to change the build performance, it is to make the build more transparent. It might make it easier to support parallel work, but the changes are useful in and of themselves.

I happen to think the bootstrap script is doing a lot more than just bootstrapping though (which in turn makes it take longer to run the linter, which is annoying), so will probably work on that after this. Maybe the linter specifically could be fixed by adding an extra CI job, as it's not like the platform matters.

aminya:
I parallelized the bootstrap script here using "threads": https://github.com/atom-ide-community/atom/pull/155/files

Aerijo:
The progress is continuing over in #21288. I've changed the goal from the original focus on making a framework for running tasks, to providing a logging interface that is controlled by the tasks. The main implementation is in place, I just need to tweak the formatting and clean up the stuff like duplicated task/lib files. Feedback on the output formatting would be appreciated at this stage.

/cc @sadick254

PR #21103, opened 2020-07-24
use `performance` instead of `Date` for measuring package load time

aminya:
### Description of the change

This uses Node's `performance` instead of `Date` for measuring package load time, which is more accurate than `Date.now`.

Reference: 
https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_now
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

### Verification
The CI passes

### Drawbacks
N/A

### Release Notes
- More accurate measure time in the TimeCope

### DIFF
--- a/src/package.js
+++ b/src/package.js
@@ -4,6 +4,7 @@ const CSON = require('season');
 const fs = require('fs-plus');
 const { Emitter, CompositeDisposable } = require('event-kit');
 const dedent = require('dedent');
+const { performance } = require('perf_hooks');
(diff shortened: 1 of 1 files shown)


sadick254 (approved):

sadick254:
@aminya Thanks for your contribution. Could you push an empty commit to trigger the CI?

aminya:
Azure removes old CI after some time.

PR #18270, opened 2018-10-19
WIP: Turn on render process tests on Linux in script/test

daviwil:
Experimenting with turning on render process tests on Linux in CI.

PR #21081, opened 2020-07-16
Cross-platform CI scripts

aminya:
### Description of the change
- This uses Powershell 7 as a cross-platform shell for CI scripts:
	- PowerShell 7 allows using the same syntax and code across all the operating systems.
	- This simplifies CI scripts. For example, previously CMD was used in Windows, which is an old shell with limited functionality.
	- The old linux_trusty container did not detect pwsh, so we use the Linux image provided by Microsoft, and we use Clang 9 that is provided in it.

- This refactors CI scripts into separate files (templates): 
	- It also allows using parallelization of the tests without copy-pasting the same code. For example: 
		- In macOS, the same templates are reused in the test phase. 
		- In Windows x86, the templates are used for parallelization of tests in [this PR](https://github.com/atom-ide-community/atom/pull/31)

	- This makes it easier to manage CI code. For example, we can quickly compare the publish step for all operating systems without going through big files and trying to find the relevant steps.

Previously, the CI scripts were written back in ~2017. At that time, Azure Templates had not been introduced yet, and because of that using PowerShell was not suitable yet. Now in 2020, we can benefit from these and improve the CI. This is also a requirement for the future Test Parallelization pull request.

### Verification
The CI passes. Tested in atom-ide-community

### Release Notes
N/A

~### Includes~
~This includes #21023.~
~This supersedes #21057.~

~### Thing you should do~
~Because this includes #21023, it requires you to set `SHOULD_SIGN=true` as an environment variable inside Azure.~

### Fork Pull request
https://github.com/atom-ide-community/atom/pull/46

Issue #20511, opened 2020-03-13
Running atom --spec test in local project = Error: Cannot find module `nslog`


the-j0k3r:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

:warning:  Atom does not have to be running at all.

Running `atom --spec test` within local clone of https://github.com/bus-stop/x-terminal results in Core error. 

![Capture](https://user-images.githubusercontent.com/31389848/76630006-94157280-653f-11ea-9bb2-2a18b0dc870e.PNG)

### Steps to Reproduce

Do not open Atom or install any 3rd party addons into Atom.

1. git clone https://github.com/bus-stop/x-terminal.git
2. cd x-terminal 
3. run `npm install` and `npm install --only=dev` 
4. run ` npm run test` aka `atom --spec test`
5. Error pops up.

5a. Remove Atom 1.44.0 remove all Atom directories left behind after reboot
5b. Install Atom 1.45.0

6. Install nslog globally
7. Delete project `node_modules`
8. cd x-terminal 
9. run `npm install --only=dev` 
10. run ` npm run test` aka `atom --spec test`
11. Watch error

**Expected behavior:**

spec test to execute without Atom errors.

**Actual behavior:**

**Reproduces how often:**

100% or always

### Versions

Issue was first noticed with Atom production 1.44.0 but its also duplicated in system with a fresh install of Atom 1.45.0

OS name and version: Windows 10 Pro 1809 build 17763.1039

```
> atom --version

Atom    : 1.44.0
Electron: 4.2.7
Chrome  : 69.0.3497.128
Node    : 10.11.0
```

```
> apm --version

apm  2.4.3
npm  6.2.0
node 10.2.1 x64
atom 1.44.0
python 
git 2.25.0.windows.1
visual studio 
```

### Uninstalled Atom and reset all atom dirs and reinstalled Atom production 1.45.0

```
[...]

Issue #19441, opened 2019-06-04
TextMate Grammar scope specifications with multiple scopes is incorrectly handled


msftrncs:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: https://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

If a TextMate grammar scope specifier (`name` or `contentName`) specifies multiple scopes, they are not correctly handled, and thus highlighting fails.

![image](https://user-images.githubusercontent.com/26179051/58852662-09712000-865d-11e9-9920-6fba209631ad.png)

The above image shows a region which should have 6 scopes, but three of them are specified on a single specification, and so they appear as a single scope.

In addition to highlighting failing, automated tests using `atom-grammar-test` also fail.

The scope specifier in the grammar file is just the list of scopes separated with spaces.

I know VSCode-TextMate explicitly accepts this, as I have seen the code that makes sure to break up the specification list at the spaces, and so my grammar file works correctly in VS Code.  I am certain I saw this in another textmate grammar file, and given enough time I will find it.

(VS Code, same grammar file)
![image](https://user-images.githubusercontent.com/26179051/58853078-9c5e8a00-865e-11e9-986d-f7bfd8d050e5.png)

I'll get a link posted to the grammar file in question, if its needed.  These specific changes are not yet posted online.

(snippet of one repository item)
```CSON
	interpolation:
		begin: "(\\$)(\\()"
		beginCaptures:
			"1":
				name: "keyword.other.substatement.powershell punctuation.definition.substatement.powershell punctuation.section.embedded.substatement.begin.powershell"
			"2":
				name: "punctuation.section.group.begin.powershell punctuation.section.embedded.substatement.begin.powershell"
		contentName: "interpolated.complex.source.powershell"
		end: "\\)"
		endCaptures:
			"0":
				name: "punctuation.section.group.end.powershell punctuation.section.embedded.substatement.end.powershell"
		name: "meta.embedded.substatement.powershell"
[...]

PR #21109, opened 2020-07-25
Parallelize Tests in CI (+ Refactor and Improvements)

aminya:
### Description of the Change
This:
- On Windows, runs core renderer tests using two Azure agents. This in addition to double the overall test speed, will allow rerunning the failed ones in case some fail due to timeouts.
- allows requesting any tests on any operating system
- parallelizes core renderer tests by using separate child processes for each file. 
- allows rerunning only the failed tests rather than restarting the whole CI.
- reduces Windows timeouts by running the tests on windows-2019.

### Verification
The CI passes

### Release Notes
Same as the description

### Closes
This supersedes these old pull requests: 
#18270

### DIFF
--- a/script/test
+++ b/script/test
@@ -87,9 +87,11 @@ function spawnTest(executablePath, testArguments, options, callback, testName, f
   const cp = childProcess.spawn(executablePath, testArguments, options)
 
   // collect outputs and errors
-  let stderrOutput = ''
-  cp.stderr.on('data', data => { stderrOutput += data })
-  cp.stdout.on('data', data => { stderrOutput += data })
+	let stderrOutput = ''
+	if (cp.stdout) {
+	  cp.stderr.on('data', data => { stderrOutput += data })
+	  cp.stdout.on('data', data => { stderrOutput += data })
+	}
 
   // on error
   cp.on('error', error => {
--- a/script/test
+++ b/script/test
@@ -194,34 +194,19 @@ function runBenchmarkTests (callback) {
   cp.on('close', exitCode => { callback(null, {exitCode, step: 'core-benchmarks'}) })
 }
 
-let testSuitesToRun = requestedTestSuites() || testSuitesForPlatform(process.platform)
+let testSuitesToRun = requestedTestSuites(process.platform)
 
-function requestedTestSuites () {
-  const suites = []
-  if (argv.coreMain) {
-    suites.push(runCoreMainProcessTests)
-  }
-  if (argv.coreRenderer) {
-    suites.push(...coreRenderProcessTestSuites)
-  }
-  if (argv.coreBenchmark) {
-    suites.push(runBenchmarkTests)
-  }
-  if (argv.package) {
-    suites.push(...packageTestSuites)
-  }
-  return suites.length > 0 ? suites : null
-}
-
-function testSuitesForPlatform (platform) {
-  let suites = []
-  // Env variable options
-  let coreMain = process.env.ATOM_RUN_CORE_MAIN_TESTS === 'true'
+function requestedTestSuites (platform) {
+  // env variable or argv options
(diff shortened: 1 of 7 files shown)


DeeDeeG:
We could re-write this PR to not include the heavy use of Azure DevOps templates. Which would make for smaller and more targeted changes, and be easier to review.

Let us know; We'd be very glad to see this improvement land at this repo, so CI can run faster here.

aminya:
> We could re-write this PR to not include the heavy use of Azure DevOps templates.

I don't think rewriting this is a good idea. They should merge #21081 before this. Not using templates just makes a soup of code that it is hard to maintain.

script/vsts/platforms/linux.yml
```diff
@@ -30,12 +30,26 @@ jobs:
...
         displayName: Update npm
 
-      - task: 1ESLighthouseEng.PipelineArtifactCaching.RestoreCacheV1.RestoreCache@1
-        displayName: Restore node_modules cache
+      - task: Cache@2
+        displayName: Cache node_modules
         inputs:
-          keyfile: 'package.json, script/vsts/platforms/linux.yml, **/package-lock.json, !**/node_modules/**/package-lock.json, !**/.*/**/package-lock.json'
-          targetfolder: '**/node_modules, !**/node_modules/**/node_modules'
-          vstsFeed: 'bae1bc26-220d-43c7-a955-4de039370de2'
```
Duv4lC4ruso: #19441 
#10326 
#20511

aminya: Please remove the spam comment which references incorrect issues. If I report your account that is much worse @Duv4lC4ruso.

DeeDeeG: I've already reported it as spam. I don't see any particularly legitimate activity on the account. Sorry if it's a real person, but the account activity looks like a bot.

sadick254:
@aminya I think this needs rebasing.

aminya:
@sadick254 This is ready to go! 😉

sadick254 (commented):
@aminya Thank you for your contribution. Looking at this PR 2 questions come to mind.
- Does this affect the overall speed of running `script/test`
- What is the impact of spawning a `child_process` for every spec file?

Another thing i have noted is, this PR changes the behaviour of running `script/test`. This PR makes it mandatory to specify the suite one would like to run. Initially `script/test` would run all the core tests without extra `args`.

aminya:
> @aminya Thank you for your contribution. Looking at this PR 2 questions come to mind.
> 
> * Does this affect the overall speed of running `script/test`
> * What is the impact of spawning a `child_process` for every spec file?
> 

Thank you for asking. I haven't noticed any negative impact on performance. The speed bottleneck of running the tests is not in spawning the child_process (which might take less than 1 second), but in running the tests themselves. 

Using separate child_process allows us to selectively choose what we want to run, and rerun them if they fail (#21295). Using this I have reduced the overall performance and test pass rate.

With that being said there is still room for improvement. If you have noticed we use `async.series` which is not ideal. I have been mostly concerned with increasing the overall test experience. To maximize the speed of running these, we should apply some other changes and change this to `async.parallel`. I can do that after we merged #21295.

> Another thing i have noted is, this PR changes the behaviour of running `script/test`. This PR makes it mandatory to specify the suite one would like to run. Initially `script/test` would run all the core tests without extra `args`.

Yes, and I think this is a better approach. However, if you don't like this, it has an easy fix. Here instead of throwing an error, we can run all the tests:
https://github.com/atom/atom/pull/21109/files#diff-0e2bc0f03fe54e5e0fff129205e6ee92R294

sadick254:
@aminya Thanks for the clarification. 
> Yes, and I think this is a better approach. However, if you don't like this, it has an easy fix. Here instead of throwing an error, we can run all the tests

No need. I think this is ok. 👍

PR #21112, opened 2020-07-25
support ES6 default require for packages

aminya:
#### In short: 
This allows loading the packages that are Es module and export their function wrapped in a default object. This is the case for any package that uses Babel 6, Babel 7, or other modern compilers for transpiling their package.

This will also allow us to use babel 7 in Atom's babel transpiler as in https://github.com/atom/atom/pull/20965

#### Longer:
The behavior of ES6 is to use `exports.default = ....` for exporting something as default. 
This is equal to using `module.exports = ....` or `module.exports = {...}`.

In ES6 requiring the modules that exported like need explicitly calling `.default`
```js
const x = require(module).default // require(module) is wrong
```

`_require` function in this pull request checks if `default` property exists and returns what it is inside it instead of returning the useless module itself (which will not have the activate function for the package). 

`_require` function checks this phenomenon and adds support for it. Many compilers or bundlers use `_interopDefault` which serves the same purpose. 

### Drawbacks
This function is written pretty conservatively and has no drawbacks.

In Atom, all the packages can only export their functions either named or wrapped in an object. There is no real useful default export that can be called directly. So the mixed situation does not happen, and we can safely merge this PR. 

### Verification
The CI passes

### Release Notes
- Adding support for the packages that export their main functions under an object and are transpiled using modern transpilers (e.g babel 6 or later, etc).

### DIFF
--- a/src/module-utils.js
+++ b/src/module-utils.js
@@ -1,5 +1,5 @@
 // a require function with both ES5 and ES6 default export support
-function requireX(path) {
+function requireModule(path) {
   const modul = require(path);
   if (modul === null || modul === undefined) {
     // if null do not bother
@@ -18,4 +18,4 @@ function requireX(path) {
   }
 }
 
-exports.requireX = requireX;
+exports.requireModule = requireModule;
--- a/src/module-utils.js
+++ b/src/module-utils.js
@@ -5,7 +5,10 @@ function requireX(path) {
     // if null do not bother
     return modul;
   } else {
-    if (modul.__esModule === true && typeof modul.default === 'object') {
+    if (
+      modul.__esModule === true &&
+      (modul.default !== undefined || modul.default !== null)
+    ) {
       // __esModule flag is true and default is exported, which means that
       // an object containing the main functions (e.g. activate, etc) is default exported
       return modul.default;
--- a/src/module-utils.js
+++ b/src/module-utils.js
@@ -1,0 +1,18 @@
+// a require function with both ES5 and ES6 default export support
+function requireX(path) {
+  const modul = require(path);
+  if (modul === null || modul === undefined) {
+    // if null do not bother
+    return modul;
+  } else {
+    if (modul.__esModule === true && typeof modul.default === 'object') {
+      // __esModule flag is true and default is exported, which means that
+      // an object containing the main functions (e.g. activate, etc) is default exported
+      return modul.default;
+    } else {
+      return modul;
+    }
+  }
+}
+
+exports.requireX = requireX;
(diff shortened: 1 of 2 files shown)


src/module-utils.js
```diff
@@ -0,0 +1,21 @@
+// a require function with both ES5 and ES6 default export support
+function requireX(path) {
+  const modul = require(path);
+  if (modul === null || modul === undefined) {
+    // if null do not bother
+    return modul;
+  } else {
+    if (
+      modul.__esModule === true &&
+      (modul.default !== undefined || modul.default !== null)
```
UziTech, line 10: This will always be true since `modul.default` can not be `undefined` and `null`.

Also this check should not be necessary. if a module wants to export `null` or `undefined` than I would rather the return be `undefined` or `null` than `{default: undefined}` or `{default: null}`.

If a module exports `__esModule = true` but doesn't provide a `default` export than this should return `undefined`.

src/module-utils.js
```diff
@@ -0,0 +1,21 @@
+// a require function with both ES5 and ES6 default export support
+function requireX(path) {
```
UziTech, line 2: I think a more descriptive name should be used. When someone sees this function called in the code they should have some idea of what it does and why it is used.

Something like: `requirePossibleEsModule` or `requireDefault`

src/module-utils.js
aminya: This check is necessary if the user has named their exports. 
```
export activate function {}
export deactivate function {}
```

We only want to use `default` if it is something other than undefined/null.

src/module-utils.js
aminya: I wanted the name to be concise. I could not think of any descriptive name 😄. I am open to all suggestions.

UziTech: I like `requireDefault`. It is not too long and fairly descriptive of what it does.

src/module-utils.js
UziTech: true for packages but not in general. Maybe we do need two separate functions for our different use cases.

That line is still always `true`.

```suggestion
      (modul.default !== undefined && modul.default !== null)
```

src/module-utils.js
aminya: What about `requireModule`? My problem with `requireDefault` is that this may not use "default", and so the name becomes a little misleading.

src/module-utils.js
aminya: The code now works in those cases too. Unless literally, someone wants to export null! 😄

src/module-utils.js
UziTech: sounds good to me

aminya: Fixed in https://github.com/atom/atom/pull/21112/commits/2ea54ac3b284cb4b9b7c667de350c5b79fb51b9e

sadick254 (approved):
Looks great. 👍

PR #21124, opened 2020-07-29
Avoid replacing : in filename

### LINKED ISSUE #20077: file ends with colon open but content isnt read
### Prerequisites

### Steps to Reproduce

1. 
2. 
3. 

**Expected behavior:**

**Actual behavior:**

**Reproduces how often:**

### Additional Information

im using atom 1.37 x64 version on linux.
i tried to open a file that ends with ':' and it opened but content of file didn't load. it loaded it empty.

sadick254:

### Identify the Bug

Resolves: #20077

### Description of the Change

Avoid replacing `:` in a file name.

### Verification Process

1. Run `script/test`
2. Open atom editor and chose a file with trailing colons
3. Confirmed that the text editor has the correct content

### Release Notes
- Fix an issue where opening a file with a colon at the end didn't work.

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -751,7 +751,7 @@ describe('AtomApplication', function() {
 
     it('truncates trailing whitespace and colons', async function() {
       await scenario.open(parseCommandLine('b/2.md::  '));
-      await scenario.assert('[_ 2.md]');
+      await scenario.assert('[_ 2.md::]');
 
       const w = scenario.getWindow(0);
       assert.lengthOf(w._locations, 1);
@@ -1406,7 +1406,7 @@ class LaunchScenario {
     );
 
     await Promise.all(
-      ['a/1.md', 'b/2.md'].map(
+      ['a/1.md', 'b/2.md', 'b/2.md::'].map(
         filePath =>
           new Promise((resolve, reject) => {
             const fullFilePath = path.join(this.root, filePath);
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -749,8 +749,18 @@ describe('AtomApplication', function() {
       assert.isFalse(w._locations[0].isFile);
     });
 
-    it('truncates trailing whitespace and colons', async function() {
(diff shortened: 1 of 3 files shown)


spec/main-process/atom-application.test.js
```diff
@@ -749,8 +749,8 @@ describe('AtomApplication', function() {
       assert.isFalse(w._locations[0].isFile);
     });
 
-    it('truncates trailing whitespace and colons', async function() {
```
aminya, line 752: Doesn't this affect Windows? File names with a colon in their name can't be written to the disk.

sadick254: It does affect windows. And you are right that filenames with colons cannot be written to disk on windows. We could do two things
- Add a condition check on `platform` and strip the `:` based on the platform.
- Not fix this bug

Adding a condition check on the code means we also have to do that in our tests, otherwise we will end up with failing tests on windows.

aminya: Yes you should add a condition to do this only on Unix based systems.

src/main-process/atom-application.js
```diff
@@ -1895,7 +1895,7 @@ module.exports = class AtomApplication extends EventEmitter {
       return result;
     }
 
-    result.pathToOpen = result.pathToOpen.replace(/[:\s]+$/, '');
+    result.pathToOpen = result.pathToOpen.replace(/\s+$/, '');
```
aminya, line 1898:
```suggestion
	if (process.platform === 'win32') {
		result.pathToOpen = result.pathToOpen.replace(/[:\s]+$/, '');
    } else {
        result.pathToOpen = result.pathToOpen.replace(/\s+$/, '');
    }
```

aminya (changes requested):
You should add a condition to do this only on Unix based systems (everything other than Windows).

jeff-hykin:
Shouldn't all invalid file names end up as a user error? (Or at least a warning) 

I don't mean to be the bearer of bad news, but also macos actually allows having a colon in a file name.

aminya:
I think https://github.com/atom/atom/pull/21124#discussion_r475553821 solves this

PR #21133, opened 2020-07-30
Show appropriate error when running test without building

sadick254:
When you run `script/test` without first building, you get the following error which is not very helpful
Screenshot 2020-07-30 at 13 11 32

This adds an assertion to provide a helpful error message in the event someone runs `script/test` without building

### DIFF
--- a/script/test
+++ b/script/test
@@ -45,19 +45,24 @@ const CONFIG = require('./config')
 const backupNodeModules = require('./lib/backup-node-modules')
 const runApmInstall = require('./lib/run-apm-install')
 
+function assertExecutablePaths(executablePaths) {
+	assert(executablePaths.length !== 0, `No atom build found. Please run "script/build" and try again.`)
+	assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
+}
+
 const resourcePath = CONFIG.repositoryRootPath
 let executablePath
 if (process.platform === 'darwin') {
   const executablePaths = glob.sync(path.join(CONFIG.buildOutputPath, '*.app'))
-  assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
+  assertExecutablePaths(executablePaths)
   executablePath = path.join(executablePaths[0], 'Contents', 'MacOS', path.basename(executablePaths[0], '.app'))
 } else if (process.platform === 'linux') {
   const executablePaths = glob.sync(path.join(CONFIG.buildOutputPath, 'atom-*', 'atom'))
-  assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
+  assertExecutablePaths(executablePaths)
   executablePath = executablePaths[0]
 } else if (process.platform === 'win32') {
   const executablePaths = glob.sync(path.join(CONFIG.buildOutputPath, '**', 'atom*.exe'))
-  assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
+  assertExecutablePaths(executablePaths)
   executablePath = executablePaths[0]
 } else {
   throw new Error('Running tests on this platform is not supported.')


aminya (changes requested):
#21109 includes this change. I don't think we should merge it here.

DeeDeeG:
Not for me to be unnecessarily contrary @aminya, but personally this is my preferred approach for this error message. :+1:

sadick254:
@aminya Thank you for working on https://github.com/atom/atom/pull/21109. 
I think we could change `A single application to run tests against was not found` to something that provides the user with not only clear information but also next steps to take to fix the error.

My suggestion is we merge this in because https://github.com/atom/atom/pull/21109 might take a while before getting merged.

aminya (changes requested):
If you want to update the messages, you should do it for all the operating systems not just Mac that you are on. You should also revert the package-lock changes. Then we can discuss if the message is helpful.

sadick254:
@aminya 
- Added the message to the other platforms
- Reverted package.json change.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "atom",
-  "version": "1.50.0-dev",
+  "version": "1.51.0-dev",
```
aminya, line 3:
```suggestion
```

script/test
```diff
@@ -49,14 +49,17 @@ const resourcePath = CONFIG.repositoryRootPath
...
   executablePath = path.join(executablePaths[0], 'Contents', 'MacOS', path.basename(executablePaths[0], '.app'))
 } else if (process.platform === 'linux') {
   const executablePaths = glob.sync(path.join(CONFIG.buildOutputPath, 'atom-*', 'atom'))
+  assert(executablePaths.length !== 0, `No atom build found. Please run "script/build" and try again.`)
   assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
   executablePath = executablePaths[0]
 } else if (process.platform === 'win32') {
   const executablePaths = glob.sync(path.join(CONFIG.buildOutputPath, '**', 'atom*.exe'))
+  assert(executablePaths.length !== 0, `No atom build found. Please run "script/build" and try again.`)
   assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
```
aminya, line 63: I think we can refactor this to prevent copy-pasting the same thing three times.
```js
function assertexecutablePaths(executablePaths) {
	assert(executablePaths.length !== 0, `No atom build found. Please run "script/build" and try again.`)
  	assert(executablePaths.length === 1, `More than one application to run tests against was found. ${executablePaths.join(',')}`)
}
```

aminya (approved):
Now, it looks good to me. Thanks!

aminya:
@sadick254 Could you rebase and drop the package-lock changes altogether?

sadick254:
@aminya doing that now. Just noticed this.

aminya:
Maybe drop this commit https://github.com/atom/atom/pull/21133/commits/c889b065c62525b98f7851e09ee951805f31f0e0

aminya (approved):

DeeDeeG:
Thanks for improving this error message.

---

Sorry if this is unwanted "help", but I think `git remote update` and then `git rebase -i origin/master`, then delete or comment out any lines having to do with `package-lock.json`, would do it pretty tidily.

(assuming your `origin` remote is pointing to `github.com/atom/atom`)

(Or more properly, replace the `pick` command with `drop` for those commits to do with `package-lock.json`, in the interactive rebase "to-do" screen. Same result though.)

But yeah, if having a couple of extra commits are okay then this looks good to me :+1:

DeeDeeG:
I was wrong, sorry. If you still want to do this the interactive rebase way, change the command from `pick` to `edit` for the first commit. Then you can change the version in `package-lock.json` to match `master` branch (1.50.0 rather than 1.51.0)... then `git add package-lock.json` and `git rebase --continue`.

sadick254:
@DeeDeeG Fixed now.

DeeDeeG (approved):

PR #21139, opened 2020-08-02
enable multi-threading on Atom

aminya:
### Description of the Change

This enables the possibility of multithreading in Atom as stated in Electron documentation. It allows using Node features in WebWorkers.
https://www.electronjs.org/docs/tutorial/multithreading

This allows offloading the parts of the code that are blocking the main thread to another thread, which can result in a huge performance gain.

#### Alternate Designs

N/A

#### Possible Drawbacks

N/A

### Verification Process
- CI passes
- Using the built Atom

### Release Notes

- Now Atom allows using the multi-threading features of Nodejs

### DIFF
--- a/src/main-process/atom-window.js
+++ b/src/main-process/atom-window.js
@@ -49,7 +49,10 @@ module.exports = class AtomWindow extends EventEmitter {
         // (Ref: https://github.com/atom/atom/pull/12696#issuecomment-290496960)
         disableBlinkFeatures: 'Auxclick',
         nodeIntegration: true,
-        webviewTag: true
+        webviewTag: true,
+
+        // node support in threads
+        nodeIntegrationInWorker: true
       },
       simpleFullscreen: this.getSimpleFullscreen()
     };


jeff-hykin:
This seems like a pretty big (awesome!) change. 


Are there any discussions around standardizing some workers? Both for atom core, but also something like a low-priority worker that extensions can share instead of all of them spinning up with own web-worker at startup

johntrouble69:
I'm new to this and I frankly have no idea what I'm doing all I know is
from the actions I've seen is there has been a lot of data taken from me me
for what ever reason unknown

aminya:
> This seems like a pretty big (awesome!) change.
> 

> Both for atom core

We need to find ways to run things in separate workers.

> Are there any discussions around standardizing some workers? Both for atom core, but also something like a low-priority worker that extensions can share instead of all of them spinning up with own web-worker at startup

We already have some API for tasks: https://flight-manual.atom.io/api/v1.50.0/Task/. We can create something similar for workers. 

I like the API of [threads.js](https://threads.js.org/) and I think that is a particularly good library. We can integrate it to Atom, so people can use workers without adding this to their dependencies.


After the basic integrations, to truly make everything parallel, we need to provide some IPC functions that provide/change the editor properties (starting with the methods that are required in language server protocol). This allows running a package fully on a separate worker.

aminya:
@darangi @sadick254 Any plan for merging this? This does not change anything by itself. We should not keep this unmerged.

sadick254:
@aminya This will be merged. Thanks for reminding us about it.

adamhayrettin11:
7 Ara 2020 Pzt 22:33 tarihinde Sadick <notifications@github.com> şunu yazdı:

sadick254 (commented):
@aminya Could you please resolve the conflicts on this PR? Let me know once that is done. I would be happy to merge it 🚀 .

aminya:
> @aminya Could you please resolve the conflicts on this PR? Let me know once that is done. I would be happy to merge it 🚀 .

Rebased. Thanks!

sadick254 (approved):

PR #21174, opened 2020-08-13
[Don't Close] Comment out activating packages

sadick254:
Work in progress. Investigating flaky CI.

### DIFF
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -75,10 +75,17 @@ describe('TextEditorRegistry', function() {
 
       atom.config.set('editor.tabLength', 8, { scope: '.source.js' });
 
+      const buffer = new TextBuffer({ filePath: 'test.js'})
+      atom.grammars.assignLanguageMode(buffer, 'source.js')
+
       const editor = registry.build({
-        buffer: new TextBuffer({ filePath: 'test.js' })
+        buffer
       });
+
       expect(editor.getTabLength()).toBe(8);
+      expect(editor.getRootScopeDescriptor().getScopesArray()).toEqual([
+        'source.js'
+      ]);
     });
   });
 


spec/text-editor-registry-spec.js
```diff
@@ -69,9 +69,9 @@ describe('TextEditorRegistry', function() {
   });
 
   describe('.build', function() {
-    it('constructs a TextEditor with the right parameters based on its path and text', async function() {
-      await atom.packages.activatePackage('language-javascript');
-      await atom.packages.activatePackage('language-c');
+    it('constructs a TextEditor with the right parameters based on its path and text', function() {
+      // await atom.packages.activatePackage('language-javascript');
+      // await atom.packages.activatePackage('language-c');
```
aminya, line 74: What is the goal of this? This does not seem right to me

sadick254: Recently atom has had numerous CI pipeline fails. This is an attempt to review the failing tests and understand why the tests are failing due to timeouts.

This test case doesn't depend on the packages being activated to perform the assertion correctly.

aminya: To fix the CI failures, you should also review and merge #21081 and #21109. Using these two, we are having a much better CI in our fork. Atom-ide-community CI runs in half the time. 

https://github.com/atom-ide-community/atom

Arcanemagus: It looks like those package activations were in there when the test was first added: https://github.com/atom/atom/commit/09d232ba4108c13d350af49abcd325ae113d7289

At least the JavaScript package looks necessary so it will recognize the language and load the grammar, perhaps the C one is there to test it isn't loading the wrong language?

sadick254: @Arcanemagus unfortunately activating the packages alone don't load the grammar. `setLanguageMode` has to be explicitly called for the grammar to be loaded. If you `log` grammar
https://github.com/atom/atom/blob/613f76d5945d2d58435aa6ec839930c4c5b1f021/src/text-editor-registry.js#L96
it will be `null` when running the test.

aminya: So shouldn't we do that instead of removing this?

PR #21221, opened 2020-08-26
Ci preparation template

aminya:
The preparation template cherry-picked from https://github.com/atom/atom/pull/21081.

See that pull request for more information.


script/vsts/platforms/templates/preparation.yml
```diff
@@ -0,0 +1,66 @@
...
+    displayName: Install npm 6.9.0
+    condition: eq(variables['Agent.OS'], 'Windows_NT')
+
+  - script: |
+      npm install --global --production windows-build-tools@4.0
+    displayName: Install windows build tools
+    condition: eq(variables['Agent.OS'], 'Windows_NT')
+
+  - script: |
+      cd script\vsts
```
sadick254, line 63: This is just a nitpick
```suggestion
      cd script/vsts
```

sadick254 (commented):
This looks great @aminya

script/vsts/platforms/templates/preparation.yml
aminya: This only runs in CMD and only on Windows. So this is not needed.

aminya: Fixed in https://github.com/atom/atom/pull/21221/commits/ec373db827b2ac2811602c69004c24d9089ebf87

DeeDeeG:
Just a head's up this has some meaningful changes in it for Linux:

- Runs Linux CI outside of Docker (should be much faster)
- Runs Linux CI on a newer Ubuntu (16.04 "Xenial", vs the older 14.04 "Trusty")
- Newer Ubuntu comes with newer Clang C/C++ compiler (Clang 9 vs Clang 5)

We have been building with these for I'd like to say about a month at our fork, so it should all work fine. Just mentioning in case we ever need to track down when these changes were introduced.

PR #21057, opened 2020-07-09
CI: Use Cache@2 task in place of the old SaveCache@1

DeeDeeG:
<details><summary>Requirements for Adding, Changing, or Removing a Feature (from template)</summary>

### Requirements for Adding, Changing, or Removing a Feature

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must contribute a change that has been endorsed by the maintainer team. See details in the template below.
* The pull request must update the test suite to exercise the updated functionality. For guidance, please see <https://flight-manual.atom.io/hacking-atom/sections/writing-specs/>.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see <https://github.com/atom/.github/tree/master/CONTRIBUTING.md#pull-requests>.

</details>

### Issue or RFC Endorsed by Atom's Maintainers

https://github.com/atom/atom/pull/19437

^ The initial CI caching PR for Atom mentioned looking into using official Azure Pipelines Caching when it was implemented. Caching has been officially implemented in Azure Pipelines for some time now, via the `Cache@2` task.

### Description of the Change

This (`Cache@2`) is the officially supported successor to the old,
unofficial, Microsoft-DevLabs-hosted `SaveCache@1` task.

It is simpler/easier to configure, and thus much easier for forks of Atom to set up.

(It "just works" without further configuration, actually.)

---

Cache the `[repo_root]/node_modules`, `script/node_modules` and `apm/node_modules` folders with the official [`Cache@2`](https://docs.microsoft.com/azure/devops/pipelines/release/caching?view=azure-devops) task, rather than the Microsoft DevLabs [Lighthouse-branded caching task](https://marketplace.visualstudio.com/items?itemName=1ESLighthouseEng.PipelineArtifactCaching).

[...]

PR #21228, opened 2020-08-28
CI: Restore caches in sync

DeeDeeG:

Discussed here: https://github.com/atom/atom/pull/21057#issuecomment-680170267

### Description of the Change

To simplify how CI handles caches, restore all three `node_modules` folders in sync with one-another.

- Track the same files for restoring all three caches. Either they all get restored, or none of them do.
- Stop tracking `apm/package-lock.json`, as it hasn't been stable lately.
- Start tracking `script/vsts/platforms/templates/preparation.yml`, because that's where Node and npm bumps and so on will live if/when the CI config gets fully "template-ized". (See: https://github.com/atom/atom/pull/21081) (This file will be automatically ignored until such a time as it is added to the repository).

### Alternate Designs

Independent cache restores (which we're doing as of before this PR), or maybe more-optimistically restoring `apm` since it gets native code rebuilt during install anyhow.

- Independent cache restores make it hard to track where restored files came from. Also, all the `node_modules` folders other than `apm/node_modules` are deleted by `npm ci`, so restoring them is just wasted time.
  - Verdict: not as good as this PR.
- Optimistically restoring `apm/node_modules` might be interesting, since it is installed with `npm install`, not `npm ci`; It's not deleted during bootstrap. But that's one more thing to test if it works well, so I'm considering it out of scope for this PR.
  - Verdict: maybe later?

### Possible Drawbacks

Reduces the scenarios in which the `apm/node_modules` cache will be restored. Could add a few minutes to bootstrapping.

(See alternate designs above; More-optimistically restoring `apm/node_modules` could be explored in a follow-up PR.)

### Verification Process

Tested in my personal fork's CI and the `atom-ide-community` fork's CI. Should work here as well.

(To verify: watch that the macOS and Windows test jobs restore all three caches, skip bootstrapping, and successfully proceed on to run their tests.)

### Release Notes

N/A

### DIFF
--- a/script/vsts/platforms/templates/cache.yml
+++ b/script/vsts/platforms/templates/cache.yml
@@ -11,20 +11,20 @@ steps:
   - task: Cache@2
     displayName: Cache node_modules
     inputs:
-      key: 'npm | "$(Agent.OS)" | "$(buildArch)" | package.json, package-lock.json, script/vsts/platforms/${{ parameters.OS }}.yml'
+      key: 'npm_main | "$(Agent.OS)" | "$(BUILD_ARCH)" | package.json, package-lock.json, apm/package.json, script/package.json, script/package-lock.json, script/vsts/platforms/${{ parameters.OS }}.yml, script/vsts/platforms/templates/preparation.yml'
       path: 'node_modules'
       cacheHitVar: MainNodeModulesRestored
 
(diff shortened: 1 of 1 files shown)


sadick254:
Thank you @DeeDeeG for this PR. It looks great.

PR #21232, opened 2020-08-29
Merge upstream master

DeeDeeG:
Just merging in a few changes from upstream.

We have gone back to `npm install` vs `npm ci` for installing `script/vsts/node_modules` dependencies. There are so few dependencies there that it only saved about 5 seconds or so to use `npm ci`, and that's on average. It varies a fair amount. I can live with this, I suppose.

I might still like to revert that change, but that would be in a follow-up commit or PR. Keeping the "merge from upstream" PR clean, faithful to what changed upstream.

There were a couple of improved comments in the templates. And on macOS the system `npm` path is used, rather than the `npm` shipped with Node. (Still gets updated to the pinned version we set it to, though? I'll take a close look and double-check once CI starts for this PR.) I went with this, because it's what upstream is doing. If we want to change that here, it's probably worth changing it at upstream as well, for consistency's sake.


DeeDeeG:
I posted this to the wrong repo, sorry!

PR #21256, opened 2020-09-05
⬆️ focus-trap@6.0.1

atom-build:
Bumps focus-trap from 2.4.5 to 6.0.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3383,11 +3383,11 @@
       "integrity": "sha512-dVsPA/UwQ8+2uoFe5GHtiBMu48dWLTdsuEd7CKGlZlD78r1TTWBvDuFaFGKCo/ZfEr95Uk56vZoX86OsHkUeIg=="
     },
     "focus-trap": {
-      "version": "6.0.1",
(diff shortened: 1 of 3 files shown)


sadick254:
`focus-trap` doesn't work as intended because of a change introduced to add a delay before autofocus. Here is the PR that introduces that change https://github.com/focus-trap/focus-trap/pull/51.

sadick254:
I have opened a PR on the `focus-trap` repo to fix the problem highlighted on the comment above. https://github.com/focus-trap/focus-trap/pull/154. Once this is merged we should be good to bump and merge this PR.

stefcameron:
Please bump to `focus-trap@6.1.0` and configure your traps with the new `delayInitialFocus: false` option. That should resolve your issue, and thanks for @sadick254's contribution to add the new option.

sadick254:
@stefcameron Thank you for facilitating the change. Really appreciate the work you are doing with `focus-trap`

stefcameron:
> @stefcameron Thank you for facilitating the change. Really appreciate the work you are doing with `focus-trap`

My pleasure!

PR #21262, opened 2020-09-05
Ci publish template

aminya:
The publish template cherry-picked from https://github.com/atom/atom/pull/21081.

See that pull request for more information.


script/vsts/platforms/templates/publish.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
+  # artifacts:
+  # - fileName
+  # - fileDir
+  # - condition
+  # - continueOnError
```
sadick254, line 6: Is there a reason why these are commented out?

aminya: This is the API documentation. It is not code.

aminya: I changed the explanation in https://github.com/atom/atom/pull/21262/commits/366c3b56f6e8da86664baa6c5ef373e142e269a7

sadick254 (approved):
This looks great. 👍

PR #21230, opened 2020-08-29
CI build template

aminya:
The build template cherry-picked from https://github.com/atom/atom/pull/21081.

See that pull request for more information.

PR #21264, opened 2020-09-05
CI build template patch

aminya:
This is a patch for #21230 

It includes the code signing keys in the code map and uses Azure macros for that instead. Azure does not pass secret variables to the program without using macros.

PR #21269, opened 2020-09-06
CI build template including patches

aminya:
#21230 including patches

This moves the secret variable to the env map. The problems of the old PR was because I did not know which variables are saved as secret in the Azure UI, and so I assumed they are available as environment variables in the shells.

Didn't notice that AC user and password are secrets variables too.

### DIFF
--- a/script/vsts/platforms/linux.yml
+++ b/script/vsts/platforms/linux.yml
@@ -20,19 +20,7 @@ jobs:
       - script: script/lint
         displayName: Run linter
 
-      - script: script/build --no-bootstrap --create-debian-package --create-rpm-package --compress-artifacts
-        env:
-          GITHUB_TOKEN: $(GITHUB_TOKEN)
-          ATOM_RELEASE_VERSION: $(ReleaseVersion)
-          CC: clang
-          CXX: clang++
-          npm_config_clang: 1
-        displayName: Build Atom
-
-      - script: |
-          sudo chown root ./out/atom*-amd64/chrome-sandbox
-          sudo chmod 4755 ./out/atom*-amd64/chrome-sandbox
-        displayName: Tweaking chrome-sandbox binary
+      - template: templates/build.yml
 
       - script: script/test
         env:
--- a/script/vsts/platforms/macos.yml
+++ b/script/vsts/platforms/macos.yml
@@ -23,24 +23,7 @@ jobs:
       - script: script/lint
         displayName: Run linter
 
-      - script: |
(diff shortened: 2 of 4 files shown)


aminya:
@sadick254 This moves the secret variable to the env map. The problems that you see is because I did not know which variables are saved as secret in the Azure UI, and so I assumed they are available as environment variables in the shells.

This should fix the issue.

DeeDeeG:
Yeah, it would be great to try this on a production type of branch. (It would have to be `master` or `electron-[something]`, or else CI will skip code signing.)

---

@aminya Here's what I looked at to troubleshoot this issue. The CI run after #21264 was merged.

- Windows successfully code-signing: https://github.visualstudio.com/Atom/_build/results?buildId=80599&view=logs&j=4fbced83-508e-5fe0-c978-5c71ec0fc506&t=a63e206b-6c6b-5e96-21cf-c8a36853d4af&l=4964
- macOS successfully code-signing, but not notarizing due to "missing" appleID and appleApiKey: https://github.visualstudio.com/Atom/_build/results?buildId=80599&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=9fd275d3-5d0a-5af8-58ae-a01a7d61dbd5&l=4966.
  - "missing" must mean they were set as secret, as you mentioned at the top post of this PR. So this patch should finally pass all the secret env vars we need to code sign _and_ notarize.

PK-Patomsiri (approved):

sadick254:
@DeeDeeG @aminya I have created a branch `electron-code-signing` to test this out. I have kicked off the CI https://github.visualstudio.com/Atom/_build/results?buildId=80937&view=results. If this code signs without any issues i would consider this PR ready to go.

DeeDeeG:
@sadick254 thanks! Looks to me like it's working.

Here's it notarizing: https://github.visualstudio.com/Atom/_build/results?buildId=80937&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=9fd275d3-5d0a-5af8-58ae-a01a7d61dbd5&l=4965

> Notarizing application at /Users/runner/work/1/s/out/Atom Dev.app

No errors there. (Some test failures after that, but I would assume they are unrelated and would pass if re-run?)

(The Linux "fail after successfully building" thing happened as well, but on closer look this [`electron-code-signing` branch still uses Node 12.4.0 in CI](https://github.com/atom/atom/blob/electron-code-signing/script/vsts/platforms/templates/preparation.yml#L23-L26), so I still think the Node 12.16.3 fix is probably correct.)

PR #21288, opened 2020-09-10
WIP: Add logging support to build script

Aerijo:
This is a continuation of #21067. This approach passes a "task" object to the scripts, instead of requiring they extend one. This makes it easier to integrate, as there are no more magic members to worry about.

Right now I just want to get a sense of how it looks on CI logs. The change is not complete. I'd probably also make it use a different log format depending on if it's CI or not, as it looks quite weird without highlighting and group collapsing.


Aerijo:
Still needs
- Work on concurrent tasks / tracking 'active' task.
- Smarter Azure CI aware output (formatting looks bad on non-CI display, and Azure doesn't support rewriting lines like the snapshot progress indicator tries to do).

Example output:

Screen Shot 2020-09-10 at 10 45 36 pm

Screen Shot 2020-09-10 at 10 19 03 pm

I like this approach a lot more than the extend one though. It's a lot closer in spirit to  #12410 while still providing the logging tools I want.

Aerijo:
The implementation is subject to change, this PR was opened to experiment with the results in CI. For example, it is no longer necessary to separate the tasks into a separate `tasks` folder because they work fine as regular functions too now (because of the `DefaultTask` default param). The `TaskManager` may also be redundant, but I'd need to get working parallel task and custom prefix support before deciding whether to drop it. Identifying parallel tasks should be possible by detecting if an unexpected task is logging output, or a `parallel: true` flag as a parameter to `task.start`. The result would be to prefix the task stack to each message

aminya (changes requested):
If ever we decided to merge this, we should not have two identical folders (lib and task). They should be consolidated. People cannot maintain two scripts!

I gave my opinion about the approach here: 
https://github.com/atom/atom/pull/21067#pullrequestreview-485505204. 

I would be more interested in this if it helps to parallelize the tasks using worker threads (If the changes make parallelization harder, I strongly disagree). Probably one simple case of parallel tasks (in a simple example) would justify the whole approach.

I parallelized the bootstrap script here using "threads". Bootstrap speeds up 30% on Mac and 20% on Windows and Linux:
https://github.com/atom-ide-community/atom/pull/155

Using simple functions allows doing this. I wonder if your tasks accept threads as well!

Aerijo:
Updated sample output (now with more Azure compatible grouping and timing information for each task):

image


image

Mellbabe (approved):

sadick254:
Thanks for the PR!

PR #21292, opened 2020-09-10
Refactor text-editor-registry .build test from async to sync

sadick254:
### Reason for change
This test has been flaky on our CI due to its async nature. It is the most frequent failing test on our CI. 
The failures are due to timeouts on the CI. The timeouts could be as a result of slow infrastructure (slow vm's) which we have no control over.

This aims to resolve the timeout failures by making the async test sync.

### Expectation
The expectation is for `.build` to construct `TextEditor` with the right params.
We don't have to load the correct language package for this test. We can load a stub `languageMode` and assert that the correct grammar is set on `TextEditor`

### DIFF
--- a/spec/text-editor-registry-spec.js
+++ b/spec/text-editor-registry-spec.js
@@ -3,6 +3,7 @@ const TextEditor = require('../src/text-editor');
 const TextBuffer = require('text-buffer');
 const { Point, Range } = TextBuffer;
 const dedent = require('dedent');
+const NullGrammar = require('../src/null-grammar');
 
 describe('TextEditorRegistry', function() {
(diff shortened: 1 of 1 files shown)


aminya:
The change itself seems a good idea. but does this actually make a difference? It would be interesting to know.

sadick254:
> The change itself seems a good idea. but does this actually make a difference? It would be interesting to know.

Yes it does. The time out experienced on the test is as a result of activating 2 packages to test `.build`. This PR removes the async nature of the test since its not necessary for the test.

With this change in place, we are 100% sure that even if we put in place the retry mechanism in our test, we would never have this test fail due to timeout.

PR #21295, opened 2020-09-10
Tests: retry timed out steps automatically

aminya:
### Description of the change

This automatically detects the timed out tests and reruns them. 

This is much faster than rerunning the tests manually in Azure. This is because the rerunning happens at a low JavaScript level compared to CI reruns that need to set up the whole environment and rerun all the tests just because one was timed out.

### Verification
The CI passes. The functions are commented with details.

### Release Notes
N/A

### DIFF
--- a/script/test
+++ b/script/test
@@ -148,7 +148,10 @@ function createTestKey(executablePath, testArguments, testName) {
 // check if a test is timed out
 function isTimedOut(stderrOutput) {
   if (stderrOutput) {
-    return ( stderrOutput.includes("timeout: timed out after") )
+    return (
+			stderrOutput.includes("timeout: timed out after") ||
+			stderrOutput.includes("Error Downloading Update: Could not get code signature for running application")
+		)
   } else {
     return false
   }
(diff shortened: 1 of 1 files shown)


aminya:
After using this PR in @atom-ide-community, I did not have any timed-out tests that are not solved by retrying. Only real genuine test failures show themselves. #21292

sadick254:
@aminya This looks great. I like the retrying bits of this PR. Could you clean your commits to make it ready for review?

aminya:
It is possible to make this PR independent of #21109, however, if we do this #21109 needs to be rebased a little different.

sadick254:
@aminya needs rebasing

aminya:
@sadick254 This is rebased! Ready to go.

If you want we can decrease the timeout interval instead of waiting for 12 seconds:

https://github.com/atom/atom/blob/7a2cc453a24030ae2eb2747f56b91c96134f4797/spec/spec-helper.coffee#L55-L60

aminya:
Also, another note: I haven't run prettier on this script (to keep the diff small). If you are done reviewing, we should do that.

sadick254:
> If you want we can decrease the timeout interval instead of waiting for 12 seconds, and let our retry mechanism deal with it instead.

Lets leave it as it is for now.

sadick254 (approved):
@aminya Thank you for your continuous contributions.

PR #21303, opened 2020-09-13
Async bootstrap

Aerijo:

None yet. This PR should show if it's worth looking into.

### Description of the Change

Attempts to make key parts of the build script async. Many stages of the build do not depend on others, but a lot of synchronous file operations are used. This PR converts many of them to use promises, to allow multiple steps to run concurrently. This is expected to be an improvement because the build seems to be IO heavy.

This change is relatively conservative, as there are still parts that use synchronous calls. I also haven't worked out all the dependencies yet, so it's possible some steps may still be run concurrently.

### Alternate Designs

I noticed @aminya is trying out worker threads. It would be interesting to see if that approach offers any significant gains over Promises.

### Possible Drawbacks

It makes logging more difficult :(. It's not a problem though, the work in #21288 theoretically supports logging for async jobs (even across worker threads).

### Verification Process

CI passes and time for bootstrap ~~+ build~~ is decreased.

Edit: The initial approach had a lot of changes, and something broke. I've restricted it to just the bootstrap right now.

### Release Notes
NA

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -35,7 +35,9 @@ async function bootstrap() {
 
   if (process.platform === 'win32') deleteMsbuildFromPath()
 
-  process.env.JOBS = "max";
+  if (proces.env.JOBS === undefined) {
+    process.env.JOBS = "max";
+  }
 
   await Promise.all([
     installScriptDependencies(ci),
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -2,45 +2,61 @@
 
 'use strict'
 
-const path = require('path')
-const CONFIG = require('./config')
-const childProcess = require('child_process')
-const cleanDependencies = require('./lib/clean-dependencies')
-const deleteMsbuildFromPath = require('./lib/delete-msbuild-from-path')
-const dependenciesFingerprint = require('./lib/dependencies-fingerprint')
-const installApm = require('./lib/install-apm')
-const runApmInstall = require('./lib/run-apm-install')
-const installScriptDependencies = require('./lib/install-script-dependencies')
-const verifyMachineRequirements = require('./lib/verify-machine-requirements')
+async function bootstrap() {
+  const path = require('path')
+  const CONFIG = require('./config')
+  const childProcess = require('child_process')
+  const cleanDependencies = require('./lib/clean-dependencies')
+  const deleteMsbuildFromPath = require('./lib/delete-msbuild-from-path')
+  const dependenciesFingerprint = require('./lib/dependencies-fingerprint')
+  const installApm = require('./lib/install-apm')
+  const runApmInstall = require('./lib/run-apm-install')
+  const installScriptDependencies = require('./lib/install-script-dependencies')
+  const verifyMachineRequirements = require('./lib/verify-machine-requirements')
 
-process.on('unhandledRejection', function (e) {
-  console.error(e.stack || e)
-  process.exit(1)
-})
+  process.on('unhandledRejection', function (e) {
+    console.error(e.stack || e)
+    process.exit(1)
+  })
 
-// We can't use yargs until installScriptDependencies() is executed, so...
-let ci = process.argv.indexOf('--ci') !== -1
+  // We can't use yargs until installScriptDependencies() is executed, so...
+  let ci = process.argv.indexOf('--ci') !== -1
 
-if (!ci && process.env.CI === 'true' && process.argv.indexOf('--no-ci') === -1) {
-  console.log('Automatically enabling --ci because CI is set in the environment')
-  ci = true
+  if (!ci && process.env.CI === 'true' && process.argv.indexOf('--no-ci') === -1) {
+    console.log('Automatically enabling --ci because CI is set in the environment')
+    ci = true
+  }
+
+  verifyMachineRequirements(ci)
+
+  if (dependenciesFingerprint.isOutdated()) {
+    cleanDependencies()
+  }
+
+  if (process.platform === 'win32') deleteMsbuildFromPath()
+
+  process.env.JOBS = "max";
+
+  await Promise.all([
+    installScriptDependencies(ci),
+    installApm(ci, true)
+      .then(() => {
+        childProcess.execFileSync(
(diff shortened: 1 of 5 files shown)


aminya (changes requested):
> I noticed @aminya is trying out worker threads. It would be interesting to see if that approach offers any significant gains over Promises.

Nodejs is single-threaded by default which means "any Nodejs code" will eventually run on the main thread. A promise only helps if the operation happens outside of Nodejs. That includes I/O intensive work (such as reading/writing/downloading/etc), etc.

Multi-threading is completely different since you spread the "Nodejs code" across different threads. For example, you can compile apm in parallel to doing something else. 

```js
    await Promise.all([
      transpilePackagesWithCustomTranspilerPaths(),
      transpileBabelPaths(),
      transpileCoffeeScriptPaths(),
      transpileCsonPaths(),
      transpilePegJsPaths(),
    ]);
```

Here you just changing the code that runs on the main thread "async" and then awaiting it later, and you expect it to become faster. This will not make any difference, and it will be probably worse since Nodejs needs to switch between all of these operations while they all "run on the main thread"

script/lib/install-apm.js
```diff
@@ -4,12 +4,27 @@ const childProcess = require('child_process');
 
 const CONFIG = require('../config');
 
-module.exports = function(ci) {
+module.exports = function(ci, promise = false) {
```
Aerijo, line 7: I can't just change the function, because this (and `runApmInstall` below) is used elsewhere in places I don't want to touch in this PR (installing test dependencies).

script/bootstrap
```diff
@@ -2,45 +2,61 @@
...
+    installApm(ci, true)
+      .then(() => {
+        childProcess.execFileSync(
+          CONFIG.getApmBinPath(),
+          ['--version'],
+          { stdio: 'inherit' }
+        );
+      })
+      .then(() => runApmInstall(CONFIG.repositoryRootPath, ci, undefined, true)),
+  ]);
```
Aerijo, line 51: Main change (git does not like wrapping sections of code into a function it seems).

Adds the JOBS=max environment config for node-gyp.

Makes the script dependencies install concurrently with apm + apm install. The other parts are too quick to be worth making concurrent (except perhaps the `cleanDependencies` function; the effect can only be seen locally though, so will need to run more tests with that and maybe make a followup PR).

Aerijo:
Making the processes async helps a little, but they compete for bandwidth and disk usage so running them in parallel (with async or worker threads) isn't as useful as it first appears. The main performance boost here comes from the `JOBS=max` setting (added to match the extra changes in #21308).

Aerijo:
Initial results indicate this out performs #21308 on mac and Linux builds, but is slower on Windows. But this is only after a single try for both of them, and the CI stage times are highly variable (at least +- a few minutes, which is all they are different by). If they are to be compared, both should be run a few more times to get a better sense of the performance.

Having said that, the difference should not be significant at all. In the end, both are limited by the performance of the two child processes.

aminya:
> Initial results indicate this out performs #21308 on mac and Linux builds, but is slower on Windows.

Just adding a note about #21308. It does not have the parallelization of the build yet, and it is only for bootstrap.

 To see the effect of threads on the build you should check this:

MacOS builds in [7:15](https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=806&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af&t=e716fffa-162f-53e5-ffce-5e575d04822c).
Linux builds in [11:45](https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=806&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=09bcc455-e770-5f06-0452-f817068d38b7).

Aerijo:
> Just adding a note about #21308. It does not have the parallelization of the build yet, and it is only for bootstrap.

What a coincidence, the same applies for this PR.

Aerijo:
This currently has a race condition; the `npmBinPath` used to install apm depends on the npm installed by installing script dependencies, so making the two processes run in parallel changes behaviour. It is probably better to work on removing the dependency on `npm` first before continuing with this.

I do not think this race is the cause of the test failures though; they seem to be flaky (if it was a problem with this PR I would have expected an import error or build time issue).

aminya:
> This currently has a race condition; 

No that is not correct. The scripts use the externally installed npm for installing apm.
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/config.js#L110

Aerijo:
...only if that variable is set though. This runs on more than just CI.

aminya:
Outside of CI the system npm is used:
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/config.js#L122

Aerijo:
No, the local one is preferred if it exists
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/config.js#L113-L122

aminya:
`fs.existsSync(localNpmBinPath)` will be false.

Aerijo:
It might not be though, that's what makes it a race condition. 

image

The first run is as per normal: first `scripts` dependencies, then apm install. It indicates the apm install is using the local npm. The second (after deleting `scripts/node_modules`) skips the dependencies and does apm install straight away. It indicates the apm install is using the environment npm.

The second scenario may occur if both are run in parallel, as it depends on if `npm` is installed before the apm install does the file exists check. In fact, it's highly likely to occur.

Now I've made it clear I don't care for having 3 different npm installs in the first place, but this PR would still change existing behaviour into a theoretically unpredictable way.

aminya:
We can't control the local condition currently. You can try to bootstrap using Node 14, and it will fail. In a local condition, you are expected to use the recommended versions. 

We have a whole issue on this. In the near future, I will write some PowerShell scripts to fix this issue.
https://github.com/atom-ide-community/atom/issues/111

sadick254:
Thanks for the PR!

PR #21317, opened 2020-09-16
Don't reinstall npm in CI

Aerijo:

### Description of the Change

Removes the `Update npm` CI build step.

This step was added to the Azure CI builds 2 years ago in #17803 to get the new `--ci` flag. There have been a couple of bumps over the years, and I believe it was introduced in 17ec08067bcc189c83ba1ab3f961abc2e8525c39 initially (it at least didn't exist in `.travis.yml` before then).

From what I can tell, this upgrading / setting of the npm version was done to take advantage of newer features or bug fixes. I do not think it should be necessary any more.

On Linux and macOS the step takes < 10 seconds to complete. On Windows it takes 30-60 seconds (4% of build time up to boostrap finish). This isn't a lot with respect to the other steps, but it happens before linting and is a cost that doesn't seem necessary. If it builds without setting it now, then we get the same effect from the stability of CI environments themselves. Issues would be isolated to upgrading the CI environment.

Regarding the use of npm, it _should not_ affect anything to do with the built Atom. All Atom packages should be installed with `apm install`, which controls it's own npm version as well as several environment variables to ensure native packages build for the correct Electron version. If npm is being used for this purpose that would be a problem. So it should only be relevant to installing the dependencies in `scripts` (interestingly, `npm` is listed as a dependency in `scripts/package.json`. It should be possible to remove it from there too).

And finally, for some reason Linux already doesn't use the fixed npm version anyway. This can be seen in any of the Bootstrap job logs: e.g., https://github.visualstudio.com/Atom/_build/results?buildId=81413&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=09bcc455-e770-5f06-0452-f817068d38b7
```
Node:	v12.16.3
Npm:	v6.14.8
Python:	v2.7.12
```
(expected is npm@6.9.0)

### Alternate Designs

Leave as is. It's not a significant gain by itself. But it may be possible to find several more changes, and the cumulative effect could be significant.

### Possible Drawbacks

Maybe no one thinks to check npm version if something breaks after it is updated in the CI environment? But then the Linux job would break anyway because it doesn't seem to use the installed npm.

### Verification Process

CI passes

### Release Notes

NA

### DIFF
--- a/script/vsts/platforms/templates/bootstrap.yml
+++ b/script/vsts/platforms/templates/bootstrap.yml
@@ -2,15 +2,12 @@ steps:
   - pwsh: |
       # OS specific env variables
       if ($env:AGENT_OS -eq "Windows_NT") {
-        $env:NPM_BIN_PATH="C:/npm/prefix/npm.cmd"
         $env:npm_config_build_from_source=true
       }
       if ($env:AGENT_OS -eq "Darwin") {
-        $env:NPM_BIN_PATH="/usr/local/bin/npm"
         $env:npm_config_build_from_source=true
       }
       if ($env:AGENT_OS -eq "Linux") {
-        $env:NPM_BIN_PATH="/usr/local/bin/npm"
         $env:CC=clang
         $env:CXX=clang++
         $env:npm_config_clang=1
--- a/script/vsts/platforms/templates/preparation.yml
+++ b/script/vsts/platforms/templates/preparation.yml
@@ -25,9 +25,6 @@ steps:
       versionSpec: 12.16.3
     displayName: Install Node.js 12.16.3
 
-  - script: npm install --global npm@6.14.8
-    displayName: Update npm
-
   # Windows Specific
   - task: UsePythonVersion@0
     inputs:


Aerijo:
The [macOS 10.14 image](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.14-Readme.md) image uses npm 3. This would be fixed by the [macOS 10.15 image](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.15-Readme.md)

aminya (changes requested):
This npm is used to bootstrap the Atom. Why do you want to get rid of this? I don't think it is a good idea to use the npm that ships with Microsoft agents. An exact version of npm is used both as a dependency and also in the CI. See https://github.com/atom/atom/pull/21212 for example. 


The 30s does not worth reducing the reliability of the CI. We can easily bring the bootstrapping time to almost 0, if we always restore the cache and use `npm install`: 
https://github.com/atom-ide-community/atom/pull/33

But this means we reduce the reliability of CI. The CI time is not a benchmark competition, and you should not sacrifice the testing for less time. The CI should only be used for testing not as a platform for development.

If you want to fix bootstrapping errors here, you should change this path to where the Microsoft agents store the npm.
https://github.com/Aerijo/atom/blob/8f063b9e9edfdaff8803be7480194dbc9bda5ab5/script/vsts/platforms/templates/bootstrap.yml#L5
https://github.com/Aerijo/atom/blob/8f063b9e9edfdaff8803be7480194dbc9bda5ab5/script/vsts/platforms/templates/bootstrap.yml#L9
https://github.com/Aerijo/atom/blob/8f063b9e9edfdaff8803be7480194dbc9bda5ab5/script/vsts/platforms/templates/bootstrap.yml#L13

But I highly discourage this, as I stated above. [Azure documentation also does not recommend hard coding the paths](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#avoid-hard-coded-references))

Aerijo:
@aminya I am not competing, I am (1) trying to minimise the time between submitting a PR and getting an indication of failure, and (2) trying to remove any unnecessary things.

> This npm is used to bootstrap the Atom.

Except it is not, as I point out in the description. Linux already uses the built in npm for some reason, and upgrading the macOS image appears to make it do the same thing.

I am quite familiar with how apm works too. The npm in this PR is used to install script dependencies. After this, the npm installed as a dependency is used. This dependency npm is used to install apm, which in turn installs its own npm for installing packages. So we are installing npm three times here, when the CI image itself already has it installed. I can understand apm wanting to lock the version down, as it is used to build Atom, but the npm version in the CI and build scripts need not be so rigid; it is only used to install `script` dependencies.

As I outline in the description, the history of requiring specific npm versions appears to be to work around discovered bugs in npm / get new features earlier (specifically the `--ci` flag).

> But this means we reduce the reliability of CI.

@DeeDeeG makes some great points [here](https://github.com/atom-ide-community/atom/pull/33#issuecomment-678581292). But I don't consider installing npm separately three times a reasonable sacrifice for stability. And the failure mode would be that the initial `script` dependencies install fails, which should be quite obvious. Not to mention, there are many other dependencies we use without installing specific versions. E.g., the C/C++ compiler on macOS, the many Linux packages that are installed.

So in the end I disagree because we will always be using components we didn't explicitly version, and the role of this particular npm is very minor.

I do see the appeal in a fixed environment, though. But the decision to merge isn't mine. It would be ideal if we could configure it to install one or the other of a fixed image npm or the `npm` dependency. Having both is just redundant. 

> An exact version of npm is used both as a dependency and also in the CI.

Yes, I hope to remove that dependency as well. This might be more realistic too, as CI could still be using fixed npm version and local builds are already a free-for-all anyway (e.g., the Node 12.18+ issue with the cryptic build error).

> We can easily bring the bootstrapping time to almost 0, if we always restore the cache and use npm install:

The bootstrap time on CI is already minimal; it is cached. But the npm reinstallation step runs regardless.

> If you want to fix bootstrapping errors here, you should change this path to where the Microsoft agents store the npm.

I'm not sure what you are suggesting here. The point is to use the built in npm, the one invoked by the `npm` command, not try to manage our own version.

aminya:
> I'm not sure what you are suggesting here. The point is to use the built in npm, the one invoked by the npm command, not try to manage our own version.

I am showing you how to do what you want to do. Although I disagree with the approach. 😄

This line specifies where npm located:
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/vsts/platforms/templates/bootstrap.yml#L5
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/vsts/platforms/templates/bootstrap.yml#L9
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/vsts/platforms/templates/bootstrap.yml#L13

and here this environment variable is used:
https://github.com/atom/atom/blob/0628ba58039824165f9dc8df04cea4c6f3872746/script/config.js#L110

So this is npm is used to run all the operations done by `npm`.


Again, I am giving the solution to your goal, but I want to stress that I don't agree with it.

Aerijo:
Yes, I would just be removing those lines. The point is not wanting to specify where npm is.

DeeDeeG:
I am deliberately not commenting about the debate above, I have only skimmed reading it anyhow.

On another note: I want to point out that we install specific Node version in CI already. Node comes with a version of npm. We can use that copy of npm if we want it to stay pinned longer (its stability would match however frequently we change Node in CI). Some platforms are already doing that in our CI).

aminya (changes requested):
There are so many factors to rely on here, while we don't have/need to. We can just spend 30 seconds and have a repeatable and reliable CI every time. This helps us to test the changes without any issues. 

Again, we can make the bootstrap time near to 0 by using `npm install` instead of `npm ci`, but that is not the point of CI.
https://github.com/atom-ide-community/atom/pull/33

Aerijo:
This didn't occur to me originally, but I'd also like to point out that the fixed npm version is installed via .... the image npm.

So even the current set up relies upon the image npm and has the same reliability concerns.

Aerijo:
While I still think this change is reasonable, it's not worth justifying now that linting is a separate job, meaning time until bootstrapped is less/not important.

PR #21318, opened 2020-09-16
Update the macOS Azure CI image to macos-10.15

Aerijo:

### Description of the Change

Updates the macOS Azure CI image from [macos-10.14](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.14-Readme.md) to [macos-10.15](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md). This
- Makes sure we are testing on the latest possible macOS image
- Updates npm from 3.10.10 to 6.14.6, most likely removing the need to upgrade it separately (see #21317)

### Alternate Designs

NA

### Possible Drawbacks

Could fail to detect changes that are incompatible with 10.14. But then, by the same logic, the current setup doesn't detect incompatibilities with 10.15, which is more recent and therefore more important IMO.

### Verification Process

CI passes

### Release Notes

NA

### DIFF
--- a/script/vsts/nightly-release.yml
+++ b/script/vsts/nightly-release.yml
@@ -56,7 +56,7 @@ jobs:
     timeoutInMinutes: 180
 
     pool:
-      vmImage: macos-10.14
+      vmImage: macos-10.15
 
     steps:
       - template: platforms/templates/preparation.yml
--- a/script/vsts/platforms/macos.yml
+++ b/script/vsts/platforms/macos.yml
@@ -10,7 +10,7 @@ jobs:
       IsSignedZipBranch: $[ dependencies.GetReleaseVersion.outputs['Version.IsSignedZipBranch'] ]
       RunCoreMainTests: true
     pool:
-      vmImage: macos-10.14
+      vmImage: macos-10.15
 
     steps:
       - template: templates/preparation.yml
@@ -48,7 +48,7 @@ jobs:
     dependsOn: macOS_build
     timeoutInMinutes: 180
     pool:
-      vmImage: macos-10.14
+      vmImage: macos-10.15
     strategy:
       maxParallel: 3
       matrix:


Aerijo:
Located all macOS builds by searching for `macos-10.14` in `scripts`.

Aerijo:
This appears to cancel the effects of the `Update npm` job ([source](https://github.visualstudio.com/Atom/_build/results?buildId=81455&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af))
```
Node:	v12.16.3
Npm:	v6.14.6
Python:	v2.7.17
```

It should not be a concern if the tests pass though, and the Linux job already doesn't use the intended npm version anyway.

aminya (approved):
This seems a good idea.

aminya:
> This appears to cancel the effects of the `Update npm` job ([source](https://github.visualstudio.com/Atom/_build/results?buildId=81455&view=logs&j=e2cf4b02-5697-54ad-cf7c-fc2a840d53af))
> 
> ```
> Node:	v12.16.3
> Npm:	v6.14.6
> Python:	v2.7.17
> ```
> 
> It should not be a concern if the tests pass though, and the Linux job already doesn't use the intended npm version anyway.

That's probably because npm is installed somewhere else in the new Mac

https://github.com/atom/atom/blob/b17a5ab6f61ff9d512a4d3c69db60d3e0203926f/script/vsts/platforms/templates/bootstrap.yml#L9

script/vsts/platforms/linux.yml
```diff
@@ -8,6 +8,9 @@ jobs:
       vmImage: ubuntu-16.04
 
     steps:
+      - script: which npm
+        displayName: Check npm location
```
sadick254, line 12: Do we need this here?

sadick254 (changes requested):

script/vsts/platforms/linux.yml
Aerijo: I think I was just testing which npm versions are being used because of how Linux isn't using the expected one, and this PR makes mac not use the expected one. It can be removed. I'm not interested in 'fixing' the wrong npm usage though.

PR #21326, opened 2020-09-16
Separate lint into own job

Aerijo:

### Description of the Change

Separates the lint step into its own job.

Linting works on the source code, is not platform dependent, and does not require a lot of installations (i.e., not apm). This makes it ideal to run on a single job separately from the actual builds to get much faster lint feedback.

### Alternate Designs

Have linting as part of the steps in a single build (e.g., the Linux one). This should not affect the lint result, but does not offer the quicker feedback a separate job might. Currently, the bootstrap task responsible for installing `script` dependencies (necessary for linter) also installs apm and all of Atom's dependencies, which aren't necessary to start linting.

   - It might be possible to add linting as part of the bootstrap task then. But that design would be weird.

### Possible Drawbacks

- ~~An extra job might slow down the start of the other jobs.~~
     - On the contrary, it appears to run fine and a significant amount is run when the GetReleaseVersion task is running

- ~~I think the current design will only lint PRs now.~~
     - Should now be fixed to lint and be added as a dependency everywhere the platform templates are used.

### Verification Process

See how CI looks.

### Release Notes

NA


sadick254:
I have been thinking of doing this. Thanks @Aerijo for this PR.

Aerijo:
Looks like Lint can run alongside the GetReleaseVersion job, which is nice. I quite like the 2 minute turnaround time. Definitely beats 20.

- Does anyone know if this can benefit from caching? Unfortunately it installs more than just the lint dependencies, including some unrelated native modules that need to be compiled.

- Could the original linting be left as part of the build jobs, and just disabled if the Lint job will be run / has started? This prevents oversights in forgetting to lint.

script/vsts/lint.yml
```diff
@@ -0,0 +1,13 @@
+jobs:
+  - job: Lint
+    timeoutInMinutes: 10
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - script: |
+          cd script
+          npm install --ci
```
aminya, line 9:
```suggestion
          npm ci
```

aminya (approved):
This is a very good improvement.

aminya:
Should we stop running the actual CI if the lint fails?

Aerijo:
> Should we stop running the actual CI if the lint fails?

I don't think it's necessary. Any tests failures may still be relevant, and if the author pushes changes before the tests finish then the old run will be cancelled anyway.

script/vsts/lint.yml
```diff
@@ -0,0 +1,16 @@
+jobs:
+  - job: Lint
+    timeoutInMinutes: 10
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - template: templates/cache.yml
+        parameters:
+          OS: linux
```
aminya: The cache template restores more than needed (top level deps and apm folder)! Installing script dependencies is faster.

Aerijo: Yeah, that's what I'm trying to measure now. It looks like it's 30 seconds per cache hit, and 1:30 for straight install. Now the CI invocation is fixed the straight install may be faster, but is it possible to only restore the cache we care about?

Aerijo: Ah, it's too much hassle. I've removed caching.

Aerijo:
I've been trying to reproduce the `find-and-replace` flaky test failures locally and on the package repo CI, but so far nothing. https://github.com/atom/find-and-replace/issues/1133

script/vsts/lint.yml
```diff
@@ -0,0 +1,12 @@
+jobs:
+  - job: Lint
+    timeoutInMinutes: 10
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - script: |
+          cd script
+          npm ci
```
aminya, line 9:
```suggestion
          npm install -g @aminya/group-dependencies
          cd script
          deps install lint
```

aminya (changes requested):
I made a tool to install only lint dependencies. Please merge that PR and apply the changes I suggested.
https://github.com/Aerijo/atom/pull/1/

script/vsts/lint.yml
Aerijo: I think I'd rather leave that to a followup PR. The separation of the lint job works without needing any extra dependencies (I also just dislike adding dependencies, so don't want to make that PR myself).

aminya: OK, I will create a separate PR.

PR #21329, opened 2020-09-16
Support sync condition in spec

aminya:
### Description of the change

1) Some conditions may be sync functions that we don't need to await! This removes the needs for awaiting these functions.

One of such conditions: https://github.com/atom/atom/blob/55abad9861aa96101423aa4751c0243d37dd9096/spec/pane-spec.js#L673

2) It also removes then on an awaited function!


aminya:
@sadick254 Could you merge this into your branch?

PR #21332, opened 2020-09-17
Group lint dependencies

aminya:
### Description of the change

I made a tool called [group-dependencies](https://github.com/aminya/group-dependencies) that allows installing only a certain group of dependencies.

For example, using this I can install lint deps only:
```ps1
          npm install -g @aminya/group-dependencies
          cd script
          deps install lint
		  ./lint
```

Behind the scenes, it makes a new package.json, and then uses that for installation. Once the installation is done, it restores the original package.json.

It is also capable of using group-specific lock files, which can speed up the installation process.

### Verification

This can be tested on https://github.com/atom/atom/pull/21326 once it is merged.

### DIFF
--- a/script/package-lock.json.lint
+++ b/script/package-lock.json.lint
@@ -1,0 +1,4134 @@
+{
+  "name": "atom-build-scripts",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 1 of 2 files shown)


sadick254:
@aminya Thank you for your efforts. The lint job at https://github.com/atom/atom/pull/21326 is already fast enough even though it installs extra script deps not necessary for linting.
> Behind the scenes, it makes a new package.json, and then uses that for installation. Once the installation is done, it restores the original package.json.

While replacing `package.json` is not really an issue. We view it as a hack to get the desired effect. Grouping dependancies is not a problem we would like to solve at this point therefore I will close this for now. 

Thank you for understanding.

PR #21335, opened 2020-09-17
🚧 Do not merge 🚧 Isolate flaky tests

Aerijo:
The flakes for the `find-and-replace` package have been bugging me recently. They aren't timeouts, so it should be an issue with the test or package logic. I couldn't reproduce them locally or on the package repo (https://github.com/atom/find-and-replace/pull/1134), so I'm testing the CI directly with this PR to try and reproduce it and then work out why it fails.


Aerijo:
It seems the project state is incorrect even in the closing of the `core:move-to-top and core:move-to-bottom` `beforeEach` function. There seems to be an assumed state with the `fixtures/project` path open as the `atom.project` paths; perhaps this project path is incorrect from another test. The `results-view-spec` specs only set the project paths to `fixtures/project`, but I've also only been able to reproduce the problem so far with all spec files enabled.

Aerijo:
Got a local failure, yay. Only seems to fail on macOS; it's never failed on Linux for me. The mac is much older though, so it could just be because it's slower. Linux doesn't seem to open anything, whereas the mac tests open an Atom instance (one appears in the dock).

Noticed that when it fails the tests sometimes log `Specs will fail because the Dev Tools have focus. To fix this close the Dev Tools or click the spec runner.`, which is interesting because I didn't think the dev tools would open in CI (or in the local tests).

Aerijo:
The dev tools are definitely being opened as well, and was the cause of some of the failures. This is meant to happen on an uncaught exception, but I've had it happen apparently randomly before, so that will need looking into as well.

Aerijo:
Preventing dev tools from opening appears to fix most of the errors. I haven't tried it without the render patch though, so I don't know if the two combined are important.

Applying this as a patch may lower flakes in the short term, but working out what is causing them to open would be better long term.

Of 100 runs, 5 errored. I did not observe any failures in `results-view-spec`. All were in the project view ripgrep specs. These errors were all similar to
```
2020-09-20T06:56:25.9056460Z ProjectFindView (ripgrep=true)
2020-09-20T06:56:25.9056800Z   finding
2020-09-20T06:56:25.9057160Z     when the find string contains an escaped char
2020-09-20T06:56:25.9057600Z       when regex search is disabled
2020-09-20T06:56:25.9058910Z         it ProjectFindView (ripgrep=true) finding when the find string contains an escaped char when regex search is disabled doesn't insert a escaped char if there are multiple backslashs in front of the char
2020-09-20T06:56:25.9059740Z           Expected object with length 0 to have length 1
2020-09-20T06:56:25.9060780Z             at jasmine.Spec.<anonymous> (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:345:84)
2020-09-20T06:56:25.9061450Z             at Generator.next (<anonymous>)
2020-09-20T06:56:25.9062420Z             at step (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:3:273)
```
and
```
2020-09-20T06:37:36.6222490Z ProjectFindView (ripgrep=true)
2020-09-20T06:37:36.6222830Z   finding
2020-09-20T06:37:36.6223200Z     when core:confirm is triggered
2020-09-20T06:37:36.6223580Z       when results exist
2020-09-20T06:37:36.6224750Z         it ProjectFindView (ripgrep=true) finding when core:confirm is triggered when results exist doesn't update the results list when a buffer outside the project changes
2020-09-20T06:37:36.6225530Z           Expected object with length 0 to have length 13
2020-09-20T06:37:36.6226580Z             at jasmine.Spec.<anonymous> (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:854:84)
2020-09-20T06:37:36.6227250Z             at Generator.next (<anonymous>)
2020-09-20T06:37:36.6228220Z             at step (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:3:273)
2020-09-20T06:37:36.6229420Z           Expected undefined to have class 'selected'.
2020-09-20T06:37:36.6231220Z             at jasmine.Spec.<anonymous> (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:859:94)
2020-09-20T06:37:36.6231930Z             at Generator.next (<anonymous>)
2020-09-20T06:37:36.6232940Z             at step (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:3:273)
2020-09-20T06:37:36.6234070Z           Expected undefined to have class 'selected'.
2020-09-20T06:37:36.6235130Z             at jasmine.Spec.<anonymous> (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:865:94)
2020-09-20T06:37:36.6235770Z             at Generator.next (<anonymous>)
2020-09-20T06:37:36.6236710Z             at step (/Users/runner/work/1/s/packages/find-and-replace/spec/project-find-view-spec.js:3:273)
```

sadick254:
@Aerijo We could try the [dev tools patch here ](https://github.com/atom/atom/pull/21335/commits/05bf222cbe43e06dac25a9d691bc9a683120af4e)on find and replace and then observe for a couple of runs. If the failure still occurs we can include the [list view patch](https://github.com/atom/atom/pull/21335/commits/58da99231d28993d0647f20ddc9a4afd8b82b9e6)

Aerijo:
I've put in atom/find-and-replace#1137 with the devtools patch.

sadick254:
Closing in favour of https://github.com/atom/find-and-replace/pull/1137

PR #21342, opened 2020-09-18
linux dpkg-deb: speed up build

aminya:
### Description of the change

This PR speeds up the build process of dpkg-deb on the Linux machines (by about 3 min):
- It uses `sudo` when available instead of `fakeroot` (in CI or when user has started in root) -> 2 min speed up
- It does not compress the artifacts internally unless we are on a release/nightly branch. -> 1 min speed up

### Verification

Before changes (16 min and 38 seconds): https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=878&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=cccdf131-0893-5d7b-ccea-9ccfc8953f95

After changes(13min and 18seconds): https://dev.azure.com/atomcommunity/atomcommunity/_build/results?buildId=879&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=cccdf131-0893-5d7b-ccea-9ccfc8953f95

### Release Notes
- Faster builds by dpkg-deb on Linux

### DIFF
--- a/script/lib/create-debian-package.js
+++ b/script/lib/create-debian-package.js
@@ -208,8 +208,13 @@ module.exports = function(packagedAppPath) {
     compressionLevel = 6;
     compressionType = 'xz';
   }
+  // use sudo if available to speed up build
+  let sudoCommand = 'fakeroot';
+  if (process.env.CI === true || (process.getuid && process.getuid() === 0)) {
+    sudoCommand = 'sudo';
+  }
   spawnSync(
-    'fakeroot',
+    sudoCommand,
     [
       'dpkg-deb',
       `-Z${compressionType}`,
--- a/script/lib/create-debian-package.js
+++ b/script/lib/create-debian-package.js
@@ -210,7 +210,7 @@ module.exports = function(packagedAppPath) {
   }
   // use sudo if available to speed up build
   let sudoCommand = 'fakeroot';
-  if (process.env.CI === true || (process.getuid && process.getuid() === 0)) {
+  if (process.env.CI || (process.getuid && process.getuid() === 0)) {
     sudoCommand = 'sudo';
   }
   spawnSync(
--- a/script/lib/create-debian-package.js
+++ b/script/lib/create-debian-package.js
@@ -200,9 +200,27 @@ module.exports = function(packagedAppPath) {
   );
 
   console.log(`Generating .deb file from ${debianPackageDirPath}`);
-  spawnSync('fakeroot', ['dpkg-deb', '-b', debianPackageDirPath], {
-    stdio: 'inherit'
-  });
+
+  // don't compress by default to speed up build
+  let compressionLevel = 0;
+  let compressionType = 'none';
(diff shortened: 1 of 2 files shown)


sadick254 (commented):
@aminya The speed gain stated here would need more than 1 run to confirm. It is also important for us to keep things on atom simple.
```
spawnSync('fakeroot', ['dpkg-deb', '-b', debianPackageDirPath], {	
    stdio: 'inherit'
});
```
The above code is simple enough, not buggy and gets the job done. We can keep it as it is for now. I am still open to reopening this provided with more data on the speed gain. Thank you for looking to improve the build time.

aminya:
The current code is more complex because it uses `fakeroot`. I changed fakeroot to sudo in a backward-compatible manner. I am happy to drop fakeroot if that is the "complexity" you are talking about. 

I have included a benchmark in the OP, but I can add more runs.

aminya:
@DeeDeeG Since you are good at doing benchmarks, it would be nice if you can run a local benchmark for me. The online benchmarks show 3 minutes improvement.

script/lib/create-debian-package.js
```diff
@@ -200,9 +200,32 @@ module.exports = function(packagedAppPath) {
...
+  // don't compress by default to speed up build
+  let compressionLevel = 0;
+  let compressionType = 'none';
+  if (process.env.IS_RELEASE_BRANCH || process.env.IS_SIGNED_ZIP_BRANCH) {
+    compressionLevel = 6;
+    compressionType = 'xz';
+  }
+  // use sudo if available to speed up build
+  let sudoCommand = 'fakeroot';
+  if (process.env.CI || (process.getuid && process.getuid() === 0)) {
```
DeeDeeG, line 213: > `process.getuid() === 0`

This seems to check if the build script is running as root. I don't think folks should run the build script as root/with `sudo`... So this condition won't be hit by most users who are being responsible with security on their systems.

Perhaps don't check for that?

aminya: To replace `fakeroot` with `sudo` which is way faster.

DeeDeeG:
@aminya this PR might be more important if we were going to run the tests on Linux. Otherwise I think we're fine, as Linux is already either the fastest CI job, or close behind the macOS build job.

aminya:
I don't really care about CI. I want to speed up the build for the users who want to build Atom locally. 

https://github.com/atom/atom/issues/21760

DeeDeeG:
If such a feature gets merged, I would like to see compression toggle-able with the `--compress-artifacts` flag.

That way it's off by default, but can be easily enabled for CI, or for anyone else who wants to host releases of Atom. (Or for anyone who just wants to save space on their hard-drive).

We should also add a "no compression" option to the `.rpm` package build if doing it for the `.deb`s.

DeeDeeG:
Actually, about my above comment... `--compress-artifacts` is already used on Linux for another purpose. I suppose a flag for this feature would need a new name, something like `--compress-linux-packages`.

PR #21345, opened 2020-09-18
Support sync condition in spec - timeout fixer

aminya:
### Description of the change

1) Some conditions may be sync functions that we don't need to await! This removes the needs for awaiting these functions.

One of such conditions: https://github.com/atom/atom/blob/55abad9861aa96101423aa4751c0243d37dd9096/spec/pane-spec.js#L673

2) It also removes then on an awaited function!

### Verification 
The CI passes. Since the timeouts happen randomly, it is hard to check if this fixes some of them, but it is better to apply these patches.

### Drawback
none

### DIFF
--- a/spec/async-spec-helpers.js
+++ b/spec/async-spec-helpers.js
@@ -7,7 +7,12 @@ async function conditionPromise(
   while (true) {
     await timeoutPromise(100);
 
-    if (await condition()) {
+    // if condition is sync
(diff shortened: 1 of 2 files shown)


sadick254 (approved):
Hey, @aminya nice catch on the sync condition being awaited in the spec helper. We can't reliably say that this will fix the timeouts on the pane specs but It is a good start.

PR #20879, opened 2020-06-03
Use the "ELECTRON_CUSTOM_VERSION" env var in build scripts

DeeDeeG:
### Requirements for Adding, Changing, or Removing a Feature

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must contribute a change that has been endorsed by the maintainer team. See details in the template below.
* The pull request must update the test suite to exercise the updated functionality. For guidance, please see <https://flight-manual.atom.io/hacking-atom/sections/writing-specs/>.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see <https://github.com/atom/.github/tree/master/CONTRIBUTING.md#pull-requests>.

### Issue or RFC Endorsed by Atom's Maintainers

Useful for #20764, but applicable to `master` branch as well.

### Description of the Change

Use an environment variable, `ELECTRON_CUSTOM_VERSION`, to decouple the version of Electron-vendored binaries we want to download (`chromedriver`, `mksnapshot`), from the versions of the download helper modules we use to do the downloading (`electron-chromedriver` and `electron-mksnapshot`).

<details><summary>Implementation details (click to expand).</summary>

This is handled during the bootstrap process here:

https://github.com/DeeDeeG/atom/blob/3e7b532/script/lib/install-script-dependencies.js#L7

Also, make sure this is handled if an Atom developer manually runs `npm install` in the `script/` directory. Handled here:

https://github.com/DeeDeeG/atom/blob/3e7b532/script/redownload-electron-bins.js

and

https://github.com/DeeDeeG/atom/blob/3e7b532/script/package.json#L53

(This is important, because `electron-chromedriver` and `electron-mksnapshot` download chromedriver and mksnapshot as npm postinstall scripts, which are only run the first time they are downloaded. Installing the scripts npm packages the first time, if `ELECTRON_CUSTOM_VERSION` isn't set, the wrong Electron-vendored would be downloaded and would persist for further builds.)

</details>

### Alternate Designs

Maybe `script/redownload-electron-bins.js` should be moved under `script/libs/` instead?

### Possible Drawbacks

Adds complexity to the build scripts.

Users manually running `npm install` in the `script` dir will download two versions each of `chromedriver` and `mksnapshot`, unless they have manually set `ELECTRON_CUSTOM_VERSION` to match `electronVersion` from the main Atom package.json beforehand.
[...]

PR #21354, opened 2020-09-20
Remove electron configuration

Aerijo:
### Description of the Change

Removes the environment configuration concerning the electron version. Managing the environment is apm's job, and it [already sets this variable anyway](https://github.com/atom/apm/blob/ff9170e2efaf11949dd9e194b0f6f3e09d37569c/src/command.coffee#L110).

### Alternate Designs

None. Only apm should be managing electron compatibility logic.

### Possible Drawbacks

None. It has no effect now anyway because apm overwrites it.

### Verification Process

CI passes

### Release Notes

NA

### DIFF
--- a/script/lib/install-script-dependencies.js
+++ b/script/lib/install-script-dependencies.js
@@ -4,7 +4,7 @@ const childProcess = require('child_process');
 
 const CONFIG = require('../config');
 
-// For the 'electron-mksnapshot' dependency
+// For the 'electron-mksnapshot' and 'electron-chromedriver' dependency
 process.env.ELECTRON_CUSTOM_VERSION = CONFIG.appMetadata.electronVersion;
 
 module.exports = function(ci) {
--- a/script/lib/install-script-dependencies.js
+++ b/script/lib/install-script-dependencies.js
@@ -4,7 +4,7 @@ const childProcess = require('child_process');
 
 const CONFIG = require('../config');
 
-// For the 'electron-mksnapshot' and 'electron-chromedriver' dependency
+// Recognised by '@electron/get', used by the 'electron-mksnapshot' and 'electron-chromedriver' dependencies
 process.env.ELECTRON_CUSTOM_VERSION = CONFIG.appMetadata.electronVersion;
 
 module.exports = function(ci) {
--- a/script/lib/install-script-dependencies.js
+++ b/script/lib/install-script-dependencies.js
@@ -4,6 +4,7 @@ const childProcess = require('child_process');
 
 const CONFIG = require('../config');
 
+// For the 'electron-mksnapshot' dependency
 process.env.ELECTRON_CUSTOM_VERSION = CONFIG.appMetadata.electronVersion;
 
 module.exports = function(ci) {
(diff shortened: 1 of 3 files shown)


script/lib/install-script-dependencies.js
```diff
@@ -4,6 +4,7 @@ const childProcess = require('child_process');
 
 const CONFIG = require('../config');
 
+// For the 'electron-mksnapshot' dependency
```
DeeDeeG, line 7: > // For the `electron-mksnapshot` dependency

This is also for `electron-chromedriver`.

script/lib/run-apm-install.js
```diff
@@ -8,9 +8,7 @@ module.exports = function(packagePath, ci, stdioOptions) {
   const installEnv = Object.assign({}, process.env);
   // Set resource path so that apm can load metadata related to Atom.
   installEnv.ATOM_RESOURCE_PATH = CONFIG.repositoryRootPath;
-  // Set our target (Electron) version so that node-pre-gyp can download the
-  // proper binaries.
-  installEnv.npm_config_target = CONFIG.appMetadata.electronVersion;
+
```
DeeDeeG, line 11: 👍 

apm should indeed be setting this very same env var here:

https://github.com/atom/apm/blob/1384050001726904ee00b30b596df57e0169e131/src/command.coffee#L110

DeeDeeG: Oops, I see you already linked this same line of source code. My bad.

script/lib/install-script-dependencies.js
sadick254: @Aerijo could you add `electron-chromedriver` to the comment.

sadick254 (commented):
Looks great. 👍

script/lib/install-script-dependencies.js
Aerijo: I've added it, but I couldn't find any direct reference to Chromedriver using it (then again, I couldn't find their complete changelogs / source either). The closest I've found is https://github.com/electron/chromedriver/pull/52#issuecomment-630845509.

DeeDeeG: It's briefly mentioned in the `README.md`, and I've read the source code and found that it is indeed used.

https://github.com/electron/chromedriver#overriding-the-version-downloaded

`electron-mksnapshot` and `electron-chromedriver` use `@electron/get`.
https://github.com/electron/chromedriver/blob/3d0e75f88efe5e7026ef47e0c6997e25196ea14d/download-chromedriver.js#L3-L8

`@electron/get` respects `ELECTRON_CUSTOM_VERSION`.

- https://github.com/electron/chromedriver/blob/3d0e75f88efe5e7026ef47e0c6997e25196ea14d/download-chromedriver.js#L3-L8
- https://github.com/electron/get/blob/7b295f33d20cfafe556cdb81c56b3e6d006a9ea6/README.md#overriding-the-version-downloaded

This was all tested/confirmed during the electron 6 PR here at the main Atom repo. (Was merged separately from Electron 6). https://github.com/atom/atom/pull/20879

Aerijo: Ah thanks, I thought they were talking about the Chromedriver project itself in that comment. I didn't think to check `@electron/get`. I've updated the comment to reflect what is actually recognising that variable, which should help guard against it becoming out of date.

PR #21367, opened 2020-09-23
Fix version information

### LINKED ISSUE #18488: Error when running `atom -version` instead of `atom --version`

sadick254:
### Identify the Bug
Resolves #18488

### Description of the Change

Use `yargs` version api and add it to `main.js`. The error is as a result to `parseCommandLine` being called after `resourcePath` checks.

### Alternate Designs
We could parse command line arguments when the main process starts. I am not sure how that would impact the start time of atom.

### Possible Drawbacks
N/A

### Verification Process
Build and tested with the following variations 
```
atom --version
atom -v
atom -version
```
All those return the correct version.

### Release Notes

- Fixed `resourcePath` error when running `atom -version`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1379,6 +1379,11 @@
         "defer-to-connect": "^1.0.1"
       }
     },
+    "@types/color-name": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/@types/color-name/-/color-name-1.1.1.tgz",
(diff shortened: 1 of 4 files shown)


aminya (approved):
Thanks for this! Upgrading yargs is a very good improvement too!

tomas1971 (commented):
N

PR #21393, opened 2020-09-27
Appendable scopes in Tree-sitter grammars

### LINKED ISSUE #20147: Append scopes instead of overriding them in tree-sitter grammars
## Summary

When creating a grammar, there needs to be an option to append scopes instead of overriding them.

## Motivation

When targeting tree-sitter nodes in a grammar file, scopes inevitably get repeated. This makes grammars cumbersome and hard to maintain.

None.

## Additional context

Consider the following example:

```cson
# Identifier
'identifier': [
  {
    # Match a long list of exceptions
    match: '^(Exception|TypeError|VeryLongListOfExceptions)$',
    scopes: 'entity.class.exception'
  },
  'entity'
]

# Inherited class
'class_definition > argument_list > identifier': [
  {
    # Match a long list of exceptions (again)
    match: '^(Exception|TypeError|VeryLongListOfExceptions)$',
    scopes: 'entity.class.exception.inherited'
  },
  'entity.class.inherited'
]
```

Would be much easier to maintain if scopes were appendable:

```cson
# Identifier
'identifier': [
  {
     # Match a long list of exceptions
    match: '^(Exception|TypeError|VeryLongListOfExceptions)$',
    scopes: 'entity.class.exception'
  },
  'entity'
]

# Inherited class
'class_definition > argument_list > identifier': + 'class.inherited' # Append
```

chbk:
## Issue

Fixes #20147

## Current state

In a Tree-sitter grammar file, the scopes of the most specific selector override other scopes. That leads to many repetitions of identical rules for different selectors. An example:

```cson
'identifier': [
  # Some rules
  {
    exact: 'String',
    scopes: 'entity.type.support' # String
  },
  {
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type' # Type
  },
  'entity.variable' # variable
]

'call > identifier': [
  # The same rules
  {
    exact: 'String',
    scopes: 'entity.type.support.constructor.call' # String()
  },
  {
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type.constructor.call' # Type()
  }
  'entity.function.call' # function()
]

'attribute > identifier:nth-child(2)': [
  # The same rules again
  {
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type.member' # _.Type
  },
  'entity.variable.member' # _.member
]
```

Grammars quickly become convoluted. This PR introduces two features to alleviate this issue.

## New feature: appendable scopes

Appendable scopes start with a period and are simply appended to less specific scopes:

```cson
'identifier': [
  {
    exact: 'String',
    scopes: 'entity.type.support'
  },
  {
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type'
  },
  'entity.variable'
]

'call > identifier': '.call' # Append '.call'

'attribute > identifier:nth-child(2)': '.member' # Append '.member'
```

## New feature: "with" rules

The "with" rule matches already applied scopes:

```cson
'identifier': [
  {
    exact: 'String',
    scopes: 'entity.type.support'
  },
  {
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type'
  },
  'entity.variable'
]

'call > identifier': [
  {
    with: 'entity.type', # If the token has the scopes 'entity' and 'type'
    scopes: '.constructor.call' # Append '.constructor.call'
  },
  'entity.function.call' # Else, override all previously applied scopes with 'entity.function.call'
]

'attribute > identifier:nth-child(2)': [
  {
    with: 'entity.type.support', # If the token has the scopes 'entity' and 'type' and 'support'
    scopes: 'entity.type.member' # Override all previously applied scopes with 'entity.type.member'
  },
  '.member' # Else, append '.member'
]
```

## Scopes selection logic

Scopes for a token are applied with the following algorithm:
1. Go to the least specific selector.
2. Set `overridable = true`.
3. Go to the first ruleset of the selector.
4. Select the first matching rule, or go to step 7 if there are no matches.
5. If the rule's scopes are appendable, append them to previously applied scopes.
6. Else, if `overridable == true`, override all previously applied scopes, and set `overridable = false`.
7. If there is another ruleset for the selector, continue to that ruleset and go to step 4.
8. If there is a more specific selector, continue to that selector and go to step 2.

When there are no appendable scopes in the grammar, this algorithm outputs the same scopes as before.

### Examples

```cson
# selector: ruleset
'identifier': [
  {
    # rule
    exact: 'String',
    scopes: 'entity.type.support' # String
  },
  {
    # rule
    match: '^[0-9_]*[A-Z]',
    scopes: 'entity.type' # Type
  },
  # rule (default)
  'entity.variable' # variable
]

# selector: ruleset
'attribute > identifier:nth-child(2)': [
  {
    # rule
    with: 'entity.type.support',
    scopes: 'entity.type.member' # _.String
  },
  # rule (default)
  '.member' # _.member
]

# selector, selector: ruleset
'''
call > identifier,
call > attribute > identifier:nth-child(2)
''': [
  {
    # rule
    with: 'entity.type',
    scopes: '.constructor.call' # Type()
  },
  # rule (default)
  'entity.function.call' # function()
]

# selector: ruleset
'call > attribute > identifier:nth-child(2)': {
  # rule
  with: 'entity.function',
  scopes: '.method' # _.method()
}
```

`_.String()` will be scoped:  
`entity.type.support` -> `entity.type.member` -> `.constructor.call` = `entity.type.member.constructor.call`

`_.method()` will be scoped:  
`entity.variable` -> `.member` -> `entity.function.call` -> `.method` = `entity.function.call.method`

#### `src/tree-sitter-grammar.js`

- `scopeSelectors[selector]` is now always an array of rulesets.
- `preprocessScopes()` now translates null/undefined scopes to `'.'` (= append nothing).
- `preprocessScopes()` parses "with" rules.  
- `preprocessScopes()` requires strings for "match", "exact", and "with" rules.  

#### `src/syntax-scope-map.js`

- Changed the switch statement in `addSelector()` to a series of if conditions to avoid code repetition since the linter doesn't allow fallthroughs (cases without breaks).
- Tweaked the structure of `SyntaxScopeMap.namedScopeTable` and `SyntaxScopeMap.anonymousScopeTable`. They are now:
    ```
    table = {
        node: {
            indices: {
                (nth-child number or *): {
                    results: ...,
                    parents: {
                        node: {
                            ...
    ```
    So for example, the selector `a > *:nth-child(3) > b:nth-child(1)` will give this table:
    ```
    namedScopeTable = {
        b: {
            indices: {
                1: {
                    parents: {
                        *: {
                            indices: {
                                3: {
                                    parents: {
                                        a: {
                                            indices: {
                                                *: {
                                                    results: ...
    ```
- Changed `setTableDefaults()` so that results of wildcards are prepended to results of other selectors, and created `setTableResults()` so that results of less specific selectors are prepended to results of more specific selectors. Results are now deep arrays of selectors sorted by specificity and rulesets sorted by order of appearance:
    ```
                       selector                             selector
       |----------------------------------------|   |---------------------|
       |           ruleset           ruleset    |   |       ruleset       |
       |  |----------------------|  |--------|  |   |  |---------------|  |
    [  [  [  rule,  rule,  rule  ], [  rule  ]  ],  [  [  rule,  rule  ]  ]  ]
    ```
- `get()` now returns that deep array.

#### `src/tree-sitter-language-mode.js`

- `applyLeafRules()` now parses a deep array to return a string of scopes.

## Possible Drawbacks

No regressions. Scopes in existing Tree-sitter grammar files are applied like before.

## Verification Process

- Added extensive specs with complex test cases.
- Built atom from this branch and tested the changes in many languages.

## Release Notes

- Allow appendable scopes in Tree-sitter grammars.

### DIFF
--- a/spec/syntax-scope-map-spec.js
+++ b/spec/syntax-scope-map-spec.js
@@ -8,11 +8,11 @@ describe('SyntaxScopeMap', () => {
       c: 'z'
     });
 
-    expect(map.get(['a', 'b', 'c'], [0, 0, 0])).toBe('x');
-    expect(map.get(['d', 'b', 'c'], [0, 0, 0])).toBe('y');
-    expect(map.get(['d', 'e', 'c'], [0, 0, 0])).toBe('z');
-    expect(map.get(['e', 'c'], [0, 0, 0])).toBe('z');
-    expect(map.get(['c'], [0, 0, 0])).toBe('z');
+    expect(map.get(['a', 'b', 'c'], [0, 0, 0])).toEqual(['z', 'y', 'x']);
+    expect(map.get(['d', 'b', 'c'], [0, 0, 0])).toEqual(['z', 'y']);
+    expect(map.get(['d', 'e', 'c'], [0, 0, 0])).toEqual(['z']);
+    expect(map.get(['e', 'c'], [0, 0, 0])).toEqual(['z']);
+    expect(map.get(['c'], [0, 0, 0])).toEqual(['z']);
     expect(map.get(['d'], [0, 0, 0])).toBe(undefined);
   });
 
@@ -24,12 +24,12 @@ describe('SyntaxScopeMap', () => {
       'b:nth-child(2)': 'z'
     });
 
-    expect(map.get(['a', 'b'], [0, 0])).toBe('w');
-    expect(map.get(['a', 'b'], [0, 1])).toBe('x');
-    expect(map.get(['a', 'b'], [0, 2])).toBe('w');
-    expect(map.get(['b'], [0])).toBe('y');
-    expect(map.get(['b'], [1])).toBe('y');
-    expect(map.get(['b'], [2])).toBe('z');
+    expect(map.get(['a', 'b'], [0, 0])).toEqual(['y', 'w']);
+    expect(map.get(['a', 'b'], [0, 1])).toEqual(['y', 'w', 'x']);
+    expect(map.get(['a', 'b'], [0, 2])).toEqual(['y', 'z', 'w']);
+    expect(map.get(['b'], [0])).toEqual(['y']);
+    expect(map.get(['b'], [1])).toEqual(['y']);
+    expect(map.get(['b'], [2])).toEqual(['y', 'z']);
   });
 
   it('can match :nth-child pseudo-selectors on interior nodes', () => {
@@ -40,9 +40,10 @@ describe('SyntaxScopeMap', () => {
     });
 
     expect(map.get(['b', 'c'], [0, 0])).toBe(undefined);
-    expect(map.get(['b', 'c'], [1, 0])).toBe('w');
-    expect(map.get(['a', 'b', 'c'], [1, 0, 0])).toBe('x');
-    expect(map.get(['a', 'b', 'c'], [1, 2, 0])).toBe('y');
+    expect(map.get(['b', 'c'], [1, 0])).toEqual(['w']);
+    expect(map.get(['a', 'b', 'c'], [1, 0, 0])).toEqual(['x']);
+    expect(map.get(['a', 'b', 'c'], [1, 1, 0])).toEqual(['w', 'x']);
+    expect(map.get(['a', 'b', 'c'], [1, 2, 0])).toEqual(['x', 'y']);
   });
 
   it('allows anonymous tokens to be referred to by their string value', () => {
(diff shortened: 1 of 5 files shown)


therobgirl (approved):

aminya:
@jeff-hykin Could you review this?

chbk:
@sadick254 @darangi @maxbrunsfeld Any chance of getting this merged?

icecream17 (approved):
LGTM

<details>
<summary> Here's a dump of notes I took while staring at the code </summary>

#### `src/syntax-scope-map.js`
(See the original description of changes above)

That's a silly linter error...

I assume the second last change was done in the (switch to if statement) rewrite, but I didn't look too closely there.

The last change seems to be done by renaming `table.result` to `table.results` and implementing it in `mergeTable`.

<details>
<summary>mergeTable is basically equivalent</summary>

The new function isn't literally equivalent because the `||` operator short circuits...
In the new function, if defaultTable.indices, it skips defaultTable.parents.
...but in practice the case where both `indices` and `parents` exist never happens
(see old switch statement code)

The if statement now at the top (necessary precedence) now says

`if (defaultTable.results) table.results = [].concat(defaultTable.results, table.results || []);`

which implements the result --> results change

</details>

`setTableResults` is called all the way back at the start

#### `src/tree-sitter-language-mode.js`

<details>
<summary> applyLeafRules() </summary>

Here's some manual typing
```ts
array path (= string[])
 depth > 1: string path
 depth = 1: [head, tail] | undefined
    head: [...(string path)[0]] | undefined
    tail: [...(string path)[1]] | undefined
 depth = 0: string | undefined
string path: [] | [undefined, string[]] | [string[] | []]
  aka
  [undefined, undefined] | [undefined, string[]] | [string[], undefined]
object path: array path
```

which I think is quite helpful for understanding this function.
</details>

For the rest, the original post covers everything

</details>

chbk:
Thanks for checking out the code @icecream17

sadick254:
@chbk Thank you for all the work you have done on this PR. 🙇 
Am sorry to say, this is not a feature we are interested in at the moment. We are currently interested in features or fixes that have a direct impact on Atom users. I am happy to revisit this feature sometime in the future when the verbosity of the grammar become a real user issue.

PR #21398, opened 2020-09-28
exclude test folders from the packaged app

aminya:
### Description of the change

This excludes `node_modules/*/tests?` from being included in the final packaged Atom. This does not affect any testing as it only excludes these in the final Atom executable.

### Benefits
- Results in a smaller packaged app which means that loading it will be faster.
- The build time will be faster since the build scripts don't need to copy/transpile/compile the tests 

### Verification 
The tests are not needed in the packaged Atom. There was already a script for jest tests (__tests__), but that does not capture the tests by other libraries.


teemoney7239 (commented):
#ok

PR #21401, opened 2020-09-28
Parallel transpiling (build)

aminya:
### Description of the change

This PR parallelizes the transpiling part of the build, so they run on different threads. This is built on top of #21308 and uses threads as its script-runner. The changes to the scripts/lib are very minimal. The build script is changed to support running the threads. 

Note: This PR is part of the series that speed up the build/bootstrap of Atom #21308 #21342 #21398 and more PRs that are coming (such as parallelizing copy assets, clean, etc). See https://github.com/atom-ide-community/atom/pull/155.

### Verification
The local builds/CI pass all the tests,

### Benefits
Results in a faster building of Atom locally or in the CI.

### Release Notes
Faster building by using threads for transpiling.

# Requires #21308

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -8,9 +8,7 @@ const childProcess = require('child_process')
 const cleanDependencies = require('./lib/clean-dependencies')
 const deleteMsbuildFromPath = require('./lib/delete-msbuild-from-path')
 const dependenciesFingerprint = require('./lib/dependencies-fingerprint')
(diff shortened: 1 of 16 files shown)


sadick254:
@aminya This PR now has conflicts. Do you mind resolving them before I give it a second look?

sadick254:
Thanks for the PR!

PR #21426, opened 2020-10-02
script executables should be LF

aminya:
### Description of the change + Benefits

This makes the script executables LF so they can be edited on Windows. 

### Issue
I had ended up editing these files on Windows and the line ending changes made them non-working inside WSL or my Linux CI.

I fixed the issue by running the following to make sure everything is back to its state:
```
 git update-index --chmod=+x script/bootstrap
 git update-index --chmod=+x script/build
 git update-index --chmod=+x script/cibuild
 git update-index --chmod=+x script/clean
 git update-index --chmod=+x script/lint
 git update-index --chmod=+x script/postprocess-junit-results
 git update-index --chmod=+x script/test
 git update-index --chmod=+x script/verify-snapshot-script
```

### Verification
The CI passes

### Drawbacks
N/A
### Release Notes
N/A

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,3 +15,13 @@ spec/fixtures/sample.txt text eol=lf
 
 # Windows bash scripts are also Unix LF endings
 *.sh		eol=lf
+
+# The script executables should be LF so they can be edited on Windows
+script/bootstrap text eol=lf
+script/build text eol=lf
+script/cibuild text eol=lf
+script/clean text eol=lf
+script/lint text eol=lf
+script/postprocess-junit-results text eol=lf
+script/test text eol=lf
+script/verify-snapshot-script text eol=lf


saizai:
FWIW, is there a good reason to have _any_ CRLF based files?

CRLF makes anything Unix-based barf, but AFAIK "LF makes Windows barf" is not true. Windows just writes that way but, I think, will accept LF files without problem. 

However, I'm not certain whether my belief is correct.

If it is, I think there entire repo should be converted to LF, to make it more compatible and _very_ slightly smaller.

If it isn't, I'd be interested to know how I'm wrong.

saizai:
Also, just not familiar with `.gitattributes` `eol=lf` — doesn't this require deleting the CRs from all impacted files?

aminya:
> FWIW, is there a good reason to have _any_ CRLF based files?

cmd files require CRLF. All the other files are fine to be LF. However, I did not want to force LF for every file in this PR. Since Atom is a text editor, there may be some files in the tests that are set to CRLF by will.

> Also, just not familiar with .gitattributes eol=lf — doesn't this require deleting the CRs from all impacted files?

After setting this in .gitattributes, you should also change the actual line ending of the file.

sadick254 (approved):
Looks good to me. Thank you for your contribution.

PR #21325, opened 2020-09-16
CI: Release preparation

aminya:
### Description of change

This refactors the scripts/vsts installation for release or nightly branch into preparation. The is already written in preparation.yml, so we just need to update the condition so it runs on these two branches.

### Verification
The CI passes. Pretty obvious change.

PR #21490, opened 2020-10-12
CI: Get 32-bit Node via the official NodeTool task

DeeDeeG:

None, but was mentioned as something that would be nice to have in a PR way back in 2018: https://github.com/atom/atom/pull/18048#issue-215400204

### Description of the Change

The official NodeTool@0 task can now download 32-bit Node even on Windows x64 hosts. (As of this PR: https://github.com/microsoft/azure-pipelines-tasks/pull/13399)

This PR makes use of that, and drops `script/vsts/windows-run.js` (which was a custom script to download and run a 32-bit copy of Node on Windows based on whether an environment variable `BUILD_ARCH` is set to `x86` or `x64`). We can get 32-bit Node the official way now instead.

### Alternate Designs

Two alternative options:

- Hard-code `SHOULD_FORCE_32_BIT_NODE` everywhere we hard-code `BUILD_ARCH` to `x86` ([demo implementation](https://github.com/DeeDeeG/atom/commit/8c046786938dd226722a8870944ddf4d552d5d4d))
- Consolidate `BUILD_ARCH` and `SHOULD_FORCE_32_BIT_NODE` into one boolean. ([demo implementation](https://github.com/DeeDeeG/atom/commit/219726b6d68e6929267d49928f0ca80892f526c5))

In either case, there would need to be a hard-coded boolean that we can give to `NodeTool`'s `force32bit` input. That eliminates the need to add a (run-time-evaluated) script to interpret the arch and set an additional `FORCE_32_BIT` variable like I did in this PR -- the short bash script I added in `script/vsts/platforms/templates/preparation.yml` could be deleted.

<details><summary>Notes about the requirements of setting `SHOULD_FORCE_32_BIT_NODE` properly (click to expand)</summary>

---

#### Technical Background -- Azure Pipelines templates and variables, when do they evaluate?

We need a boolean value to give to `NodeTool`'s `force32bit` input, so as to tell it whether to force downloading 32-bit Node. (That is, "force" downloading 32-bit Node even on 64-bit hosts -- We can only get 64-bit hosts on Azure DevOps Services, the Microsoft-hosted version of Azure DevOps, which Atom's CI currently uses.) (The un-appealing alternative is breaking the two Windows arches into separate template files, which would be much less [DRY](https://wikipedia.org/wiki/Don%27t_repeat_yourself).)

Doing this in a multi-config situation (`matrix` strategy for different arches of Node on Windows) was surprisingly tricky. See this documentation link for details of when variables/templates are evaluated: https://docs.microsoft.com/azure/devops/pipelines/process/runs?view=azure-devops#process-the-pipeline. `matrix` variables are evaluated very late, much later than templates like `preparation.yml` where the `NodeTool` task lives, so in some way or another `SHOULD_FORCE_32_BIT_NODE` needs to be evaluatable at (or even set at) runtime, ideally right before downloading Node.

#### Solution

Macro syntax `$(VAR)` expressions evaluate at runtime, so that is a key part of making this work. However, reading `BUILD_ARCH` would yield `x86` or `x64`, which are strings, not the boolean we need to feed into `NodeTool`'s `force32bit` input.

Either `SHOULD_FORCE_32_BIT_NODE` needs to be hard-coded to `true` wherever we want to use x86 Node on Windows (currently just for the "Windows build x86" job), or it can be set dynamically in a script just before installing Node. I chose to do it in a script, so that modifying the rest of the `yml` and CI configs in general is easier without having to remember to hard-code `SHOULD_FORCE_32_BIT_NODE` as `true` in the relevant places, just basing everything off of the existing `BUILD_ARCH` variable.

---

</details>

### Possible Drawbacks

None.

### Verification Process

Three ways to verify:

- In the "Install Node.js 12.16.3" (`NodeTool@0`) task, check that it prints out `x86` filenames/paths as it extracts Node for the "Windows x86" job, and prints `x64` filenames/paths when extracting Node on the "Windows x64" job.
- In the "bootstrapping build environment" (`script/bootstrap`) task, observe which arch of Node `apm --version` prints out. Should be `ia32` on the "Windows x86" job, and `x64` on the "Windows x64" job.
- In the "Build Atom" (`script/build`) task, check the line "Packaging app for platform win32 [arch here] using electron v6.1.12" -- it should be `ia32` on the "Windows x86" job, and `x64` on the "Windows x64" job.

### Release Notes

N/A


script/vsts/platforms/templates/preparation.yml
```diff
@@ -20,9 +20,17 @@ steps:
     condition: eq(variables['Agent.OS'], 'Linux')
 
   # Common
+  - bash: |
+      if [ "${BUILD_ARCH}" = "x86" ]; then \
+        echo "##vso[task.setvariable variable=SHOULD_FORCE_32_BIT_NODE]true"
+      fi
```
aminya, line 26: This should use pwsh to be consistent with all the other scripts.

```suggestion
  - pwsh: |
      if ($env:BUILD_ARCH -eq "x86") {
        echo "##vso[task.setvariable variable=SHOULD_FORCE_32_BIT_NODE]true"
      }
```

script/vsts/platforms/templates/preparation.yml
```diff
@@ -20,9 +20,17 @@ steps:
...
+    displayName: Set SHOULD_FORCE_32_BIT_NODE based on BUILD_ARCH
+    # Templates evaluate too early for the matrix variable "BUILD_ARCH" to be available. Scripts evaluate at runtime, so "BUILD_ARCH" will be available.
```
aminya, line 28: This sentence is not needed, and it a little inaccurate. Matrix variables (BUILD_ARCH) are calculated before running anything and so are available everywhere. It is not related to templates, scripts, etc.

https://github.com/atom/atom/blob/9e172191c08bd5b7e330b984abee774814a534df/script/vsts/platforms/windows.yml#L10

```suggestion
```

aminya (approved):
Thanks!

I am happy that Azure has finally added support for Node 32.

script/vsts/platforms/templates/preparation.yml
```diff
@@ -20,9 +20,17 @@ steps:
...
+
   - task: NodeTool@0
     inputs:
       versionSpec: 12.16.3
+      force32bit: $(SHOULD_FORCE_32_BIT_NODE)
```
aminya, line 33: Why not just set when it is needed? When I wrote these scripts, the reason I used the PowerShell syntax to set the environment variables was because `Agent.OS` was not available in the following simpler syntax. 

But for BUILD_ARCH that should not be the case.
```suggestion
      ${{ if eq(variables['BUILD_ARCH'], 'x86') }}:
      		force32bit: true
```

script/vsts/platforms/templates/preparation.yml
DeeDeeG: I suppose this is the right thing to do.

If I could go back in time and standardize on `bash`, I would. I'm much more familiar with it, and most Atom maintainers don't seem to use Windows very much. So the chance is they are better with `bash` than Powershell. And I personally am familiar with `bash`, not Powershell. Somehow when we first discussed this I didn't think `bash` was an option, or that it wouldn't work on Windows, but I see now that it does. 🤷 So anyhow, some time in the future I would personally like it if we moved everything to `bash`... but it's better for now to be consistent.

script/vsts/platforms/templates/preparation.yml
DeeDeeG: If I commit this change, you will see that both  of the Windows builds will install 64-bit Node.

The code you suggest here was my first instinct, and it does not work.

I applied this suggested diff on a branch and ran CI. See that the "Windows x86" build installs 64-bit Node:
https://dev.azure.com/DeeDeeG/b/_build/results?buildId=857&view=logs&j=e4115def-01ab-568c-a31a-bc15166ffabd&t=12c9d5a1-842e-5bcb-4a43-64a451213953

script/vsts/platforms/templates/preparation.yml
DeeDeeG: You are incorrect about this. This is the documentation:

> To turn a pipeline into a run, Azure Pipelines goes through several steps in this order:
> 
> - 1. First, expand [templates](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops) and evaluate [template expressions](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops).
> [ ... ]
> - 4. For each job selected to run, expand [multi-configs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops#parallelexec) (`strategy: matrix` or `strategy: parallel` in YAML) into multiple runtime jobs.
> [ ... ]
> 
> This ordering helps answer a common question: why can't I use certain variables in my template parameters? Step 1, template expansion, operates solely on the text of the YAML document. Runtime variables don't exist during that step. After step 1, template parameters have been completely resolved and no longer exist.

From here: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/runs?view=azure-devops#process-the-pipeline

---

I have tested many variations of this. The way I put in this pull request works, as do the "Alternate Designs" I mention in the body text of my pull request.

It is difficult to explain in a brief comment in the yml.

DeeDeeG: I could change it to this, which is more precise:

```diff
-    # Templates evaluate too early for the matrix variable "BUILD_ARCH" to be available. Scripts evaluate at runtime, so "BUILD_ARCH" will be available.
+    # Templates evaluate too early for the matrix variable "BUILD_ARCH" to be available in ${{ template expressions }}.
+    # Scripts are run at runtime, so this script will make "SHOULD_FORCE_32_BIT_NODE" available to the following NodeTool task.
```

script/vsts/platforms/templates/preparation.yml
aminya: I like your [first alternate design](https://github.com/DeeDeeG/atom/commit/8c046786938dd226722a8870944ddf4d552d5d4d). It is much simpler. Could you use that and instead rename `SHOULD_FORCE_32_BIT_NODE` to `IsWinX86`?

This new variable can be handy in other places as well.

DeeDeeG: Note to Atom's maintainers: If you have an opinion on which approach is best, I'd appreciate hearing from you on this.

- This pull request as it is now
- Alternate Design 1 (as mentioned in the body text of the pull request)
- Alternate Design 2

In any of these cases, the variable named `SHOULD_FORCE_32_BIT_NODE` could be renamed `IsWinX86` instead, if that's nicer. (As suggested by @aminya).

sadick254: I prefer the approach in this PR. I don't see anything problematic with it. As for the variable name `SHOULD_FORCE_32_BIT_NODE` It makes sense to rename it to `IsWinX86`.

aminya: There is nothing wrong with it. The alternate design 1 is more succinct. 

Similarly, in [this PR](https://github.com/atom/atom/pull/21325/files#diff-f0efa4d06560f5cba2917290fef9cad255405df2775d4a29cc79f8f0bdf82e3cR25), I introduced `IsNightlyBranch`, and I used it every where.


I'll let the final decision up to you. To me that seems simpler.

script/vsts/platforms/templates/preparation.yml
aminya: The comment is correct now. 

 I still do not see why you are writing this so much verbose in the code. Better approach is to just put a link to the [Variable Definition section of Azure docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#what-syntax-should-i-use) on top of the YAML code, so everyone can read an learn about all the aspects of variable definition.

I have shortened the link for you: 
https://bit.ly/3dJk31m

script/vsts/platforms/templates/preparation.yml
DeeDeeG: @sadick254 thanks, I'll update the PR.

---

@aminya For the record "all the unnecessary code" is one powershell script spanning three lines. (If not for nice formatting, the script could easily be one line -- if not for spacing out the `{}` brackets of the `if` statement.) With a line of metadata to have a human-readable name in the CI. And one or two lines of comments.

This pull request actually removes more lines than it adds.

When a pull request is this close to optimal, and the different options are so trivially similar, it causes a lot of communication and technical review to have this many comments back and forth. For very little technical difference. It's not worth this much discussion or debate, in my opinion, in order to shuffle one or two variables around.

When a PR is "good enough," I think it's polite to just leave it be. As I have said elsewhere, we all have the chance to contribute in our own style when we contribute. If there are disagreements, official maintainers can decide.

DeeDeeG:
I could see dropping the technical info from the comment.

I added this to the comment: `# Convert "BUILD_ARCH" to a boolean ("IsWinX86") for the following NodeTool task.` Maybe that is enough on its own.

I might put that followed simply by: `# TODO: Consider refactoring "BUILD_ARCH" into a boolean and merging it with "IsWinX86".`

- Drop technical info about "matrix variables" and `${{ template expressions }}` and scripts from the comment?
- Add a `TODO` comment about potentially refactoring `BUILD_ARCH` into a boolean?


But that's all to do with comments, which are about as low-stakes as pull request reviews get... I'm comfortable with the PR as-is, it's just the subjective matter of whether the comment needs changing.

Thanks folks for the feedback and review.

script/vsts/platforms/templates/preparation.yml
aminya: Ok. That's fine. It's always possible to update the code in follow up PRs.

aminya:
> I could see dropping the technical info from the comment.
> 
> I added this to the comment: `# Convert "BUILD_ARCH" to a boolean ("IsWinX86") for the following NodeTool task.` Maybe that is enough on its own.

Yes. That is better in my opinion. It describes your goal of this script better. The technical information about Azure syntax would be better left out.

sadick254 (approved):

PR #21495, opened 2020-10-13
Add defaultFontSize setting

### LINKED ISSUE #9873: Default font size reset / deleted after command window:reset-font-size

sadick254:
Resolves https://github.com/atom/atom/issues/9873

## Description
https://github.com/atom/atom/issues/9873#issuecomment-617433194

### DIFF
--- a/src/workspace.js
+++ b/src/workspace.js
@@ -375,10 +375,12 @@ module.exports = class Workspace extends Model {
 
   initialize() {
     this.originalFontSize = this.config.get('editor.fontSize');
+    this.defaultFontSize = this.config.get('editor.defaultFontSize');
     this.project.onDidChangePaths(this.updateWindowTitle);
     this.subscribeToAddedItems();
     this.subscribeToMovedItems();
     this.subscribeToDockToggling();
+    this.subscribeToFontSize();
   }
 
   consumeServices({ serviceHub }) {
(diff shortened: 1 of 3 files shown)


src/workspace.js
```diff
@@ -1744,13 +1746,18 @@ module.exports = class Workspace extends Model {
     }
   }
 
-  // Restore to the window's original editor font size.
+  // Restore to the window's default editor font size.
   resetFontSize() {
-    if (this.originalFontSize) {
-      this.config.set('editor.fontSize', this.originalFontSize);
+    if (this.defaultFontSize) {
+      this.config.set('editor.fontSize', this.defaultFontSize);
```
UziTech, line 1752: `this.originalFontSize` was only needed to reset the font size to what it was when the window was opened. I don't think `this.defaultFontSize` is needed since it will always be the same as `atom.config.get('editor.defaultFontSize')`.
```suggestion
    this.config.set('editor.fontSize', this.config.get('editor.defaultFontSize'));
```

src/workspace.js
```diff
@@ -375,10 +375,12 @@ module.exports = class Workspace extends Model {
 
   initialize() {
     this.originalFontSize = this.config.get('editor.fontSize');
```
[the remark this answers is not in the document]
sadick254: It is not used anywhere. I have removed the references.

src/workspace.js
sadick254: Removed the subscription as well. 👍

src/workspace.js
```diff
@@ -363,7 +363,6 @@ module.exports = class Workspace extends Model {
       })
     };
 
-    this.originalFontSize = null;
```
aminya, line 366: Isn't this change breaking? It would be nice to have 
```js
this.originalFontSize = this.config.get('editor.defaultFontSize')
```

sadick254: Good point @aminya

l0v3us (commented):
>

rmwalk85:
Fuck u

Erraach-walid:
****-])))

PR #21561, opened 2020-10-21
performance: add more CSS containment to text-editor

aminya:
### Description of the Change

This adds more [CSS containment](https://drafts.csswg.org/css-contain/#contain-property) to tree-view CSS classes. This results in performance improvements by giving the hint to the browser that the size/paint/layout of these elements does not affect other elements around them.

I have added the CSS containment for each class selectively by testing all the functionalities. As you see, these do not change the functionality. 

I have explicitly specified the excluded/incompatible values, so no confusion happens in the future that why some of the `contain` values are missing.

### Benefits
Improves the performance and responsiveness of the text editor. It has improved scrolling performance. The datatips and decorations are rendered faster than before. 

The reason is that the browser will not need to recalculate the styles on clicks, scrolling, drag-drop, during the loading, etc.

Quoting [CSS specifications](https://drafts.csswg.org/css-contain/#contain-property)
> The contain property allows an author to indicate that an element and its contents are, as much as possible, independent of the rest of the document tree. This allows user agents to utilize much stronger optimizations when rendering a page using contain properly and allows authors to be confident that their page won’t accidentally fall into a slow code path due to an innocuous change.

<details>
<summary>Some benchmarks</summary>

The benefits are more apparent during interactions.

Here I opened the src/text-editor.js and I waited for 15s (linter/linter-ui-default and atom-ide-javascript were enabled)
After this PR
image
image

Before this PR
image
image

</details>

### Future improvements
- I have some PRs on hold which adds CSS containment to other classes. such as this one: https://github.com/atom-ide-community/atom/pull/238

- We can provide an optional API for using a more contained class. For example, we can have an `overlay-contained` which allows creating a fully contained datatip. As I tested this, this class works with atom-ide-datatips.

### Verifications 

I have tested all the functionalities including:
- Writing text
- Removing text
- Selecting text
- Highlighting text
- Folding and unfolding 
- Hovering on the gutter
- Line numbers preserved
- Text entries in settings view, fuzzy finder, etc work without issues
- Datatips and Linter overlays work as expected
- Linter decorations work as expected
- Autocomplete works as expected
- Pigments colors work as expected
- Scrolling works as expected
- Themes can be changed
- Text editors can be resized or moved

This is in addition to all of the tests passing.

### Release Notes
- Improving the performance of text editor by adding more CSS containment

### DIFF
--- a/static/core-ui/utils.less
+++ b/static/core-ui/utils.less
@@ -7,4 +7,4 @@
 @contain_but_paint: layout size style;
 @contain_but_layout_size: paint style;
 @contain_but_layout_paint: size style;
-@containt_but_size_paint: layout style;
+@contain_but_size_paint: layout style;
--- a/static/core-ui/utils.less
+++ b/static/core-ui/utils.less
@@ -1,0 +1,10 @@
+
+// CSS containment variables.
+// The words after `but` explicitly show the excluded/incompatible values.
+@contain_all: layout size paint style;
+@contain_but_layout: size paint style;
+@contain_but_size: layout paint style;
+@contain_but_paint: layout size style;
+@contain_but_layout_size: paint style;
+@contain_but_layout_paint: size style;
+@containt_but_size_paint: layout style;
--- a/static/core-ui/utils.less
+++ b/static/core-ui/utils.less
@@ -2,9 +2,9 @@
 // CSS containment variables.
 // The words after `but` explicitly show the excluded/incompatible values.
 @contain_all: layout size paint style;
-@contain_but_layout: size paint style;
-@contain_but_size: layout paint style;
-@contain_but_paint: layout size style;
-@contain_but_layout_size: paint style;
-@contain_but_layout_paint: size style;
-@contain_but_size_paint: layout style;
+@contain_except_layout: size paint style;
+@contain_except_size: layout paint style;
+@contain_except_paint: layout size style;
+@contain_except_layout_size: paint style;
+@contain_except_layout_paint: size style;
+@contain_except_size_paint: layout style;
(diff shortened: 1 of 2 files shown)


static/core-ui/utils.less
```diff
@@ -0,0 +1,10 @@
+
+// CSS containment variables.
+// The words after `but` explicitly show the excluded/incompatible values.
+@contain_all: layout size paint style;
+@contain_but_layout: size paint style;
+@contain_but_size: layout paint style;
+@contain_but_paint: layout size style;
+@contain_but_layout_size: paint style;
+@contain_but_layout_paint: size style;
+@containt_but_size_paint: layout style;
```
hSaria, line 10: Nice work, mate. Small typo on this one (gotta fix the reference too).
```suggestion
@contain_but_size_paint: layout style;
```

Also, might want to use `except` instead of `but`, like `contain_except_layout_paint`, but not a big deal.

aminya: Thanks for the catch. I fixed this.

> Also, might want to use except instead of but, like contain_except_layout_paint, but not a big deal.

`but` seems easier to type 😃

sadick254: `except` conveys exclusion better than `but`. Could we change this. @aminya

aminya: Renamed it in
https://github.com/atom/atom/pull/21561/commits/f37d7f8577724932e8e5b403cfc43ce355992da7

sadick254 (approved):

sadick254:
@aminya Can we rebase this branch?

PR #21620, opened 2020-10-31
Refactor TextEditor updating

aminya:
### Description of the change

This PR refactors the codes that were responsible for updating different parts of TextEditor out of `update` function, and, then, it uses those refactored functions directly, which bypasses the loop and switch case inside `update`.

`finishUpdate` is also refactored that allows finishing updating of TextEditor from inside themselves without the need for creating objects outside and passing them in.

### Benefits
- This reduces the complexity of the update algorithm for those that are updated directly (`O(n^2)`->`O(1)`). This removes the overhead of checking for the given value against the keys in switch-case, and it also removes the overhead of the for-loop. Directly calling these functions is similar to the compile-time method dispatch used in other languages like C++. The previous method was a runtime dispatch algorithm that does not allow the V8 engine to optimize these calls and indirections.

- Having these functions separate makes it easier to maintain, find, and reason about them. 

### Verification

I have moved the code inside `update` to a separate function as it is (exact copy-paste), and I have not changed "any part of the actual code". If you follow my commits one by one, you will see that the moved code is kept exactly identical.

All the tests must pass because this does not change the functionality. 

### How to review
Follow my commits one by one.

### Release Notes
- Refactor the codes that were responsible for updating different parts of TextEditor

### DIFF
--- a/src/text-editor.js
+++ b/src/text-editor.js
@@ -4207,7 +4207,7 @@ module.exports = class TextEditor {
   // * `softTabs` A {Boolean}
   setSoftTabs(softTabs) {
     this.softTabs = softTabs;
-    this.update({ softTabs: this.softTabs });
+    this.updateSoftTabs(this.softTabs, true);
   }
 
   // Returns a {Boolean} indicating whether atomic soft tabs are enabled for this editor.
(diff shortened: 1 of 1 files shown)


sadick254 (approved):
This looks pretty neat. 👍

aminya:
> This looks pretty neat. 👍

You're welcome!

Issue #21091, opened 2020-07-20
Unable to build with nodejs > 12.16 due to ERR_PACKAGE_PATH_NOT_EXPORTED


michaellass:
### Prerequisites

* [ x ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

atom fails to build with nodejs versions newer than 10 (e.g., 12 or 14).

### Steps to Reproduce

1. install nodejs 12 or higher
2. run `script/build --ci`

**Expected behavior:**

build script should run without errors.

**Actual behavior:**

```
internal/modules/cjs/loader.js:498
  throw new ERR_PACKAGE_PATH_NOT_EXPORTED(basePath, mappingKey);
  ^

[Error [ERR_PACKAGE_PATH_NOT_EXPORTED] [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main resolved in /opt/local/var/macports/build/_Users_lass_ports_editors_atom/atom/work/atom-atom-1d0abfa/out/app/node_modules/github/node_modules/@babel/helper-compilation-targets/package.json
  at applyExports (internal/modules/cjs/loader.js:498:9)
  at resolveExports (internal/modules/cjs/loader.js:514:23)
  at Function.Module._findPath (internal/modules/cjs/loader.js:642:31)
  at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1017:27)
  at Function.Module._load (internal/modules/cjs/loader.js:899:27)
  at Module.require (internal/modules/cjs/loader.js:1090:19)
  at require (internal/modules/cjs/helpers.js:75:18)
  at Object.<anonymous> (/opt/local/var/macports/build/_Users_lass_ports_editors_atom/atom/work/atom-atom-1d0abfa/out/app/node_modules/github/node_modules/@babel/preset-env/lib/debug.js:8:33)
  at Module._compile (internal/modules/cjs/loader.js:1201:30)
[...]

PR #21623, opened 2020-10-31
Babel 7 - Take 2

aminya:
## Things remaining
Any help or idea regarding these is appreciated. I decided to open this PR to get some fresh ideas to solve these issues after months of trying different things:
- [ ] Spying @babel/core in the compile-cache test does not seem to work. We might need to use `Sinon` or something similar for that. (the test related to `CompileCache-> addPathToCache(filePath, atomHome)`)

- [ ] In the CI, the packages (like language-*) are not transformed (macOS package tests). This does not happen offline.

### Description of the Change

This upgrades babel to babel 7. 

This PR is the result of several months of effort and trying. 
I had to fix electron-link so:
- it can support parsing new JavaScript: https://github.com/atom/electron-link/pull/20
- it can bundle lazy require https://github.com/atom/electron-link/pull/27 -> This is merged but not registered. That's why I am using @aminya/electron-link
- [and some](https://github.com/atom/electron-link/pull/26), [other changes](https://github.com/atom/electron-link/pull/31)

I developed two plugins specifically for Atom to support its special situation:
- [babel-preset-atomic](https://github.com/atom-ide-community/babel-preset-atomic) this is the plugin that includes all the other babel plugins. It has babel-preset (configured for Electron 6), babel-flow, babel-react, and many of the proposal plugins that are equivalents of the similar plugins included in the deprecated stage-0 package (e.g., class properties, private properties, decorators, private methods, etc.). It adds some of the new proposal plugins, such as optional chaining. It also allows us to do compile-time calculations using codegen and prevel.

See the full list of plugins that babel-preset-atomic has [in this package.json](https://github.com/atom-ide-community/babel-preset-atomic/blob/dd2843291989ea99e184e7c9279b57c05c7f6412/package.json#L19).

- [babel-plugin-transform-not-strict](https://github.com/atom-ide-community/babel-plugin-transform-not-strict): this removes 'use strict' whenever 'use babel' (or similar triggers) are used. This is because Babel 6/7 now assumes all the ES modules (those who use `import/export`) are strict by default. However, this breaks some of the Atom packages because they do non-strict things in ES modules. So, I developed this plugin to remove 'use strict' when the triggers are used. I have baked a feature in this plugin that allows removing 'use strict' selectively by adding 'not strict' directive. Ultimately, we should add this 'not strict' to any file that uses JS loosely in ES modules. An example of this is `spec/update-process-env-spec.js`

There are two move plugins included for backward compatibility:
- [babel-add-module-exports](https://github.com/lijunle/babel-plugin-add-module-exports): this plugin circumvents the fact that ES modules export default by "default" keyword. This means we need to call `require().default` to consume them. This plugin works in combination with transform-not-strict to add `module.exports=` to the packages that export default things that should have been named exports. For the packages' root, I already made [another PR that adds support for both cases](https://github.com/atom/atom/pull/21112), but we cannot patch require for the packages' internal requires as we only have access to the root.

- [@babel/plugin-transform-reserved-words](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-reserved-words) that allows the uses of the old banned keywords such as `package` or `debugger`.

### Benefits
This PR has several benefits:
- Allows building Atom with Node 12.16 or above (see [this issue](https://github.com/atom/atom/issues/21091)). This will be crucial for upgrading Electron.
- This allows using the new babel plugins (e.g., optional chaining) just by including `"use babel"`. 
- This also enhances the performance of babel compiling. 

### Previous attempts
https://github.com/atom-ide-community/atom/pull/67
https://github.com/atom/atom/pull/20965

### Release Notes
- Updates babel to babel 7 🎉


aminya:
The build passes on our fork: https://github.com/atom-ide-community/atom/pull/67. Not sure why it has failed here.

UziTech:
We need to do extensive testing to make sure this doesn't break any packages or this should be a major version bump for Atom.

UziTech:
We should use the [atom-package-downloader](https://github.com/atom/atom-package-downloader) to check for ones that use babel and compile them against this configuration for babel to check for how many will fail.

aminya:
> We need to do extensive testing to make sure this doesn't break any packages or this should be a major version bump for Atom.

This is not ready yet, but I have done my best to prevent any breakage. There may be rare cases that I am not still aware of. I have tested the core packages and all of those that I have installed. I encourage everyone to do the same. 

For future babel changes (not this PR), we should make the directive version-specific. For example, `'use babel 7'`.

UziTech:
I feel like using something like [@atom/babel7-transpiler](https://www.npmjs.com/package/@atom/babel7-transpiler) is a better idea so packages can use their own babel config.

aminya:
> I feel like using something like [@atom/babel7-transpiler](https://www.npmjs.com/package/@atom/babel7-transpiler) is a better idea so packages can use their own babel config.

This is used rarely by the current packages, and so that is not an option for those that are already using `'use babel'`.
https://github.com/atom/atom-babel7-transpiler/network/dependents?dependents_before=MTM2NjY2NDY4MDg

UziTech:
It would be nice to make `'use babel'` look at a configuration file so we don't need to update Atom once babel releases v8 and v9, etc.

UziTech:
> This is used rarely by the current packages

probably because `atomTranspilers` is not mentioned anywhere in the docs. I would bet that most package developers don't even know it is an option.

aminya:
> It would be nice to make `'use babel'` look at a configuration file so we don't need to update Atom once babel releases v8 and v9, etc.

In addition to this, we should provide a version based default configuration. For example, having a babel 8 config for those who `'use babel 8'` and a default configuration for those who use `'babel 7', etc. This removes the need for having babel.config.js in each package.

 After we successfully merge this PR, we can do these types of changes. These should be gradual (by giving deprecation warnings).

UziTech:
> This removes the need for having babel.config.js in each package.

but we should at least allow it. Some packages might want something different than the default configuration.

That is the beauty of `atomTranspilers` it gives the package creator the flexibility to use their own language and configuration.

aminya:
> > This removes the need for having babel.config.js in each package.
> 
> but we should at least allow it. Some packages might want something different than the default configuration.

Yeah sure. We can make `'use babel config'` to use the provided babel config.

UziTech:
I would say default to babel v5 so there are no breaking packages that don't change and add `'use babel 7/8'` for any packages that want to use newer versions.

Win win!

aminya:
> I would say default to babel v5 so there are no breaking packages that don't change and add `'use babel 7/8'` for any packages that want to use newer versions.
> 
> Win win!

Well, the story is not that easy. The problem is that we "must" upgrade the babel if we want to bump Node and Electron. The last Babel 5 version was developed in 2016 (developed using Node 6)! Node 6 to Node 12 has had many changes. We can't run Babel 5 forever. That's why I made two special plugins that circumvent these issues. 

Check the changelog:
https://nodejs.org/en/download/releases/

The issue of this current PR is not the plugins now (I have fixed those to a great extent). The remaining issue is that the transformation is not triggered at all in the CI.

aminya:
The build error is strange. It is not something I have seen before.

```js

Generating snapshot script at "D:\a\1\s\out\startup.js" (2064)
Error: ENOENT: no such file or directory, open 'supports-color'
  at Object.openSync (fs.js:458:3)
  at Object.readFileSync (fs.js:360:35)
  at module.exports (D:\a\1\s\script\node_modules\electron-link\lib\generate-snapshot-script.js:25:25)
  at module.exports (D:\a\1\s\script\node_modules\electron-link\lib\index.js:10:18)

```

sadick254:
@aminya there are conflicts on this PR. Do you mind resolving them? I would like to review this PR but it's still on draft state. Do you plan on adding more commits?

sadick254:
Thank you for your issue!

We haven't gotten a response to our comment above.  I'm going to close this but don't hesitate to reach out if you plan to add more commits, we'll be happy to reopen the issue.

aminya:
I sent several emails to @atom and @github and opened several pull requests and issues, and many of them did not get any response. I even messaged the CEO directly (@nat) but still got no response. Considering all of this, do you expect me to continue supporting this organization?

https://twitter.com/AminYa74/status/1425025879294500864

PR #21637, opened 2020-11-03
Revert "⬆️ postcss@8.1.4"

therobgirl:
Reverts atom/atom#21578


therobgirl (commented):
✌💛🚀
[atom-master.zip](https://github.com/atom/atom/files/5485254/atom-master.zip)
octocat-spinner-128

PR #21661, opened 2020-11-06
check if testRunner is es module

UziTech:
### Requirements for Contributing a Bug Fix

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only fix an existing bug. To contribute other changes, you must use a different template. You can see all templates at <https://github.com/atom/.github/tree/master/.github/PULL_REQUEST_TEMPLATE>.
* The pull request must update the test suite to demonstrate the changed functionality. For guidance, please see <https://flight-manual.atom.io/hacking-atom/sections/writing-specs/>.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see <https://github.com/atom/.github/tree/master/CONTRIBUTING.md#pull-requests>.

### Identify the Bug

If the `atomTestRunner` is transpiled by typescript or anything else using `atomTranspilers` into an ES module the test runner will not run.

https://github.com/UziTech/terminal/blob/a116a214b6690390dfa656f61813c0596ba7523e/package.json#L30-L36

### Description of the Change

set `testRunner` to `testRunner.default` if `testRunner.__esModule` is set to `true`;

### Possible Drawbacks

none

### Verification Process

compiled Atom and with this fix and tested the code at https://github.com/UziTech/terminal/tree/custom-runner-ts

### Release Notes

Allow `atomTestRunner` to point to an ES Module

PR #21715, opened 2020-11-18
check if testRunner is es module

UziTech:
copy of #21661 but using #21112 

### Identify the Bug

If the `atomTestRunner` is transpiled by typescript or anything else using `atomTranspilers` into an ES module the test runner will not run.

https://github.com/UziTech/terminal/blob/a116a214b6690390dfa656f61813c0596ba7523e/package.json#L30-L36

### Description of the Change

set `testRunner` to `testRunner.default` if `testRunner.__esModule` is set to `true`;

### Possible Drawbacks

none

### Verification Process

compiled Atom and with this fix and tested the code at https://github.com/UziTech/terminal/tree/custom-runner-ts

### Release Notes

Allow `atomTestRunner` to point to an ES Module


UziTech:
Should we decaffeinate `initialize-test-window.coffee` in this PR as well?

src/initialize-test-window.js
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+
+    document.title = 'Spec Suite';
+
+    const clipboard = new Clipboard();
+    TextEditor.setClipboard(clipboard);
+    TextEditor.viewForItem = (item) => atom.views.getView(item);
+
+    const testRunner = requireModule(testRunnerPath);
```
UziTech, line 112: using `requireModule` here is the only change other than decaffeinating the file.

darangi:
Thanks @UziTech

PR #21777, opened 2020-12-01
Upgrade to electron 9.3.5

sadick254:
This upgrades atom from `6.1.12` to `9.3.5`

## Issues to resolve
 - [x] Failure due to node-abi not detecting abi for version 9.3.5 and runtime electron

## Electron Upgrade Test Manifest

- [x] Review issues tagged with [`electron`](https://github.com/search?utf8=%E2%9C%93&q=org%3Aatom+label%3Aelectron+is%3Aopen+is%3Aissue) to see if any of them are fixed
- [x] Verify that you can collaborate with Teletype
    - [x] Host using Atom's current stable release can collaborate with guest using Atom with the new version of Electron
    - [x] Guest using Atom's current stable release can collaborate with host using Atom with the new version of Electron
    - [x] Host and guest can collaborate when both are using Atom with the new version of Electron
    <details>
    <summary>Steps</summary>

    1. Open a file
    1. Share a portal
    1. Have a guest join the portal
    1. When guest edits the text in the file, verify that the host sees those changes
    </details>
### Text Input/Keybindings

See how to [setup keyboard layouts](https://github.com/atom/maintainers-log/blob/master/how-tos/setup-keyboard-layouts.md).

- [x] IME not working (https://github.com/atom/atom/issues/14911)
	- [x] macOS
	- [x] Windows 10
	- [x] Ubuntu
  <details>
  <summary>Repro steps</summary>

  1. Install an IME layout, for example Pinyin simplified  
  1. Open Atom and use the layout you installed as the keyboard layout
  1. Example, if you picked Pinyin simplified type in `zhongwenshuru`

  **Expected:** in Pinyin simplified `中文输入` is expected

  gif

  **Issues we've seen:** Only shows latin first character and not every character typed.

  gif
  </details>

- [x] macOS: Key binding with composition characters are ignored (https://github.com/atom/atom/issues/15189)
  <details>
  <summary>Repro steps</summary>

  1. Open your keymap.cson file and add the following:
      ```cson
      'atom-text-editor':
        # use hjkl; vim-like when pressing alt (word by word, paragraph by paragraph)
        # or alt (letter by letter)
        'alt-h': 'core:move-left'
        'alt-j': 'core:move-down'
        'alt-k': 'core:move-up'
        'alt-l': 'core:move-right'
      ```
  1. Save keymap.cson and reload atom.
  1. Try the new keymappings under `ABC - Extended` keyboard layout

  **Expected**: the keybindings work

[...]

PR #21805, opened 2020-12-16
set app.allowRendererProcessReuse to false + fix package.json + fix remote/shell deprecation

aminya:
For Electron 9: #21777

Addresses https://github.com/atom/atom/pull/21777#issuecomment-746936053

Addresses https://github.com/atom/atom/pull/21777#issuecomment-747066738

### DIFF
--- a/src/main-process/start.js
+++ b/src/main-process/start.js
@@ -40,7 +40,7 @@ module.exports = function start(resourcePath, devResourcePath, startTime) {
   const previousConsoleLog = console.log;
   console.log = nslog;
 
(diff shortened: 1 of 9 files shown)


src/main-process/start.js
```diff
@@ -40,6 +40,9 @@ module.exports = function start(resourcePath, devResourcePath, startTime) {
   const previousConsoleLog = console.log;
   console.log = nslog;
 
+  // TODO this should be set to true before Electron 12 - https://github.com/electron/electron/issues/18397
```
aminya, line 43:
```suggestion
  // TodoElectronIssue this should be set to true before Electron 12 - https://github.com/electron/electron/issues/18397
```

Issue #21422, opened 2020-10-01
Atom is not install-able on Ubuntu Groovy 20.10


mfonville:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Atom cannot be installed on Ubuntu Groovy 20.10 because `gvfs-bin` is no longer part of `gvfs` source package, but Atom has it specified as dependency.

### Steps to Reproduce

1. Upgrade to Groovy
2. `atom` gets uninstalled
3. `atom` cannot be re-installed because of the missing dependency

mfonville:
@DeeDeeG probably also of interest to you!

DeeDeeG:
HMMMM yes I'll take a look at this if I get a chance. Thanks for the heads up.

DeeDeeG:
gvfs-bin was added in this commit: https://github.com/atom/atom/commit/f431bb6396f9f5c5c5f98fa08040dc3410896d69#diff-25608b78e299bdba245c59c930e97a00

From the commit message, direct quote:

>  Add more complete list of Debian dependencies
>
> This list was taken from the webupd8 .deb file
>
> Closes #4114

As far as I can tell, there was never a place where [WebUpd8](http://www.webupd8.org/) documented why they added these packages. So we're left to sort of reverse-engineer why (or if) we need these.

So, I'm thinking: Delete the dependency and see if that breaks something (which would be useful to reverse-engineer why we need this). Or just add [`gvfs-bin`'s dependencies](https://packages.ubuntu.com/focal/gvfs-bin) and see if those are good enough. Namely: `gvfs-common` and `libglib2.0-bin`.

[...]

PR #21806, opened 2020-12-17
Drop libgconf dependency for electron7+

mfonville:
Drop the libgconf debian packaging dependency. Not necessary for Electron 7 and higher. See for more information https://github.com/atom/atom/issues/21422#issuecomment-702402584


aminya:
You should also remove it from here: https://github.com/atom/atom/blob/f1b0e3b47b30ed9379821b341b3337696b8a2a12/script/vsts/platforms/templates/preparation.yml#L7

mfonville:
Thanks, will do!

mfonville:
@sadick254 I think you can merge this in your branch :-)

sadick254 (approved):
Looks good.

PR #21808, opened 2020-12-17
Revert "set app.allowRendererProcessReuse to false + fix package.json + fix remote/shell deprecation"

Daniel199604:
Reverts atom/atom#21805

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,9 +21,9 @@
       }
     },
     "@atom/fuzzy-native": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/@atom/fuzzy-native/-/fuzzy-native-1.1.2.tgz",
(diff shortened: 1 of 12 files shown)


aminya (changes requested):
What is the point of this? This a spam PR and should be closed.

ThatXliner:
I'm no JavaScript expert but how is this spam? I just see some deprecation removal. Although I also see some version downgrades.... 😕

aminya:
I have already reported this to GitHub. They always pay attention to the reports and they take down the accounts that make spam. 

Even if you don't check the code: this pull request is just reverting my PR that was reviewed and merged by the core team for no logical reason. In this repository, the pull requests should follow the template, but this one just uses "reverts #21805". This means it is an automated PR.


This is the pull request template: https://github.com/atom/atom/blob/master/PULL_REQUEST_TEMPLATE.md

PR #21843, opened 2021-01-03
Fix Atom not quitting on MacOS

### LINKED ISSUE #17672: Atom does not quit on OS X
### Description

Atom does not quit on OS X (both shortcut and menu)

### Steps to Reproduce

1. Open Atom
2. Try to quit Atom by pressing CMD-Q or by selecting the "Quit Atom" in the menu.

**Expected behavior:** Atom closes

**Actual behavior:** doesn't react to the command

### Versions

Atom: 1.28.1
OS X: High Sierra 10.13.4

### LINKED ISSUE #20831: I cannot quit Atom
I've tried to quit Atom but it hasn't been working.
I'm pretty sure that I've tried everything:
Command + Q
and selecting quit from the dropdown menu.
I'm using a Mac.
If anyone has any solutions, or this problem has been fixed, please tell me.

aminya:
### Description of the change

This makes Atom quitting on MacOS like any other operating system.

### Applicable issues
Fixes the critical issues that have been open for a long time:
Fixes #17672
Fixes #20831

### Verification

Atom will be closed like macOS and Linux. So, there will not be any behavior difference causing errors like #17672

The Official CI's download links for the PR's build
https://github.visualstudio.com/Atom/_build/results?buildId=94291&view=artifacts&type=publishedArtifacts

### Release Notes
- Fix Atom not quitting on MacOS in some cases

### DIFF
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -1107,19 +1107,11 @@ describe('AtomApplication', function() {
   });
 
   describe('when closing the last window', function() {
-    if (process.platform === 'linux' || process.platform === 'win32') {
-      it('quits the application', async function() {
-        const [w] = await scenario.launch(parseCommandLine(['a']));
-        scenario.getApplication(0).removeWindow(w);
-        assert.isTrue(electron.app.quit.called);
-      });
-    } else if (process.platform === 'darwin') {
-      it('leaves the application open', async function() {
-        const [w] = await scenario.launch(parseCommandLine(['a']));
-        scenario.getApplication(0).removeWindow(w);
-        assert.isFalse(electron.app.quit.called);
-      });
-    }
+    it('quits the application', async function() {
+      const [w] = await scenario.launch(parseCommandLine(['a']));
+      scenario.getApplication(0).removeWindow(w);
+      assert.isTrue(electron.app.quit.called);
+    });
   });
 
   describe('quitting', function() {
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -458,10 +458,8 @@ module.exports = class AtomApplication extends EventEmitter {
       if (this.applicationMenu != null) {
         this.applicationMenu.enableWindowSpecificItems(false);
       }
-      if (['win32', 'linux'].includes(process.platform)) {
-        app.quit();
-        return;
-      }
+      app.quit();
+      return;
     }
     if (!window.isSpec) this.saveCurrentWindowOptions(true);
   }


DeeDeeG:
@aminya there is a test that would need updating as well, see this commit for the updated test: DeeDeeG@d82cc45

DeeDeeG:
IMO this should not close the linked issues, as it is a workaround that removes a desirable (if currently broken) feature. Those issues are not solved, we just avoid running into them by removing the feature that's broken.

We can certainly update the folks in those threads that they won't be hitting this issue. But, ideally, I hope the "run in the background" feature will be restored to working condition in the long term.

aminya:
It should close them as it fixes them. You may open other separate issues for other features.

DeeDeeG:
I hope this video clarifies the difference between closing a window and quitting a program on macOS:
https://youtu.be/i6IOhuzl68k?t=24

Once the bugs are really fixed, users should be able to close the last Atom window and still have Atom running in the background. It's very common on macOS. It's how Safari, Firefox and Chrome work, as well as Visual Studio Code. Pretty much any GUI program that can open multiple documents/pages works this way in macOS. Closing and Quitting are often totally separate actions on macOS.

So it's sad to see that feature go away, and those original issues are not "fixed," we just took a feature away. Taking the feature away isn't the same as fixing it. Fixing the bugs and restoring the functionality should be done as soon as possible.

mtoensing:
Hey, please solve the issue with Atom not quitting on macOS. Even a work-around would be great. 

"Closing and Quitting are often totally separate actions on macOS."

Yes, I know. But Atom can only be closed with force quit. And it can be reproduced everytime.

aminya:
> Hey, please solve the issue with Atom not quitting on macOS. Even a work-around would be great.
> 
> "Closing and Quitting are often totally separate actions on macOS."
>
> Yes, I know. But Atom can only be closed with force quit. And it can be reproduced everytime.

Thanks for the comment. I know it is a horrible situation, but this PR fixes it.

 You don't longer need to force quit Atom just to update a package, Atom will not block your computer from shutting down or logging off, and it will not eat your memory, CPU resources, and battery.

btangmu:
One of the greatest of all mac apps -- namely Calculator -- quits when you close its window. So, Atom will be in good company if it does the same. I've stopped using Atom since I can't quit; it would be great to be able to use it again.

rseichter:
The status quo described in #17672 means that a bug in Atom prevents both logging out the current macOS user and cleanly shutting down a machine. This is far, far worse than the process of launching Atom taking a little more time. I therefore support merging this pull request if it reliably solves the Atom-fails-to-quit bugs. Personally, I consider keeping Atom running in the background after the last window has been closed a mere afterthought, given that users can simpliy minimise Atom.

DeeDeeG:
Thanks to some strong leads/initial investigation by @aminya and @Aerijo, and people's helpful description of the problem in the relevant issues, I think I've found an actual fix for the underlying problem? Pull request: #21847

Feel free to test this and let me know if it works for you. It's working for me locally, anyhow.

Download link: https://dev.azure.com/DeeDeeG/_apis/resources/Containers/11602065/atom-mac.zip?itemPath=atom-mac.zip%2Fatom-mac.zip

mtoensing:
@DeeDeeG  I can not download the file. I get a "not authorized" error.

DeeDeeG:
@mtoensing I see. _(I can download that link, but only as I am logged in to my own account! In a Private Browsing window, I also see a "login" prompt)._

Here is the build from my pull request: https://github.visualstudio.com/9e67709c-1fe5-47f2-8cf6-4902878a11f5/_apis/build/builds/94308/artifacts?artifactName=atom-mac.zip&api-version=6.0&%24format=zip

All artifacts: https://github.visualstudio.com/Atom/_build/results?buildId=94308&view=artifacts

The link in my previous comment is updated and should work now, as well. _(These builds are based on the same code. The PR version has an amended commit message, and a different code comment, but otherwise is identical)_.

ThatXliner:
Is this superseded by https://github.com/atom/atom/pull/21847?

DeeDeeG:
They are each technically viable solutions to the problem. But they are mutually incompatible.

<details><summaryTechnical background (click to expand):</summary>

Running the `app.quit()` funciton once in Atom makes everything buggy.

This current PR increases the number of `app.quit()` calls to at least two, to ensure Atom quits cleanly on macOS when the last window closes.

However, that would introduce a behavior change; Up until now, Atom has stayed open when the last window closes on macOS.

The other PR https://github.com/atom/atom/pull/21847 decreases the number of `app.quit()` calls to zero, ensuring Atom stays open in the background when the last window closes on macOS.

</details>

As far as I can tell most macOS users expect Atom to stay open in the background when the last window closes. _(Quitting when the last window closes would be a surprise, then.)_ In which case the other PR https://github.com/atom/atom/pull/21847 is the desirable solution. If Atom users on macOS offer strong feedback that they'd like Atom to quit when the last window closes, then this current PR is the right solution.

If folks want to take some sort of poll of Atom's macOS users, I suppose we could try to find out what people really prefer.

There has not really been a discussion on that, so I think restoring the current behavior to working order is correct for now. That means going with https://github.com/atom/atom/pull/21847

ThatXliner:
IMO, #21847 is the way to go. Safari is still open when I close the last window.

But shouldn't it be that when you press "cmd-q" Atom quits cleanly? I have problems when clicking "restart" when updating packages. Or is this already in the current code?

sadick254:
Closing in favour of https://github.com/atom/atom/pull/21847

DeeDeeG:
> But shouldn't it be that when you press "cmd-q" Atom quits cleanly?

Yes, and either PR should be adequate to fix that.

> I have problems when clicking "restart" when updating packages. Or is this already in the current code?

I am looking into that... Hopefully either of these PRs would also fix that. But restarting for updates is its own special process. And startup probably shouldn't be affected by this bug, no-matter what. So it's likely a separate issue.

PR #10967, opened 2016-02-25
Fix a inconsistent getLineCount() use

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
This PR refer to [vim-mode-issue 961](https://github.com/atom/vim-mode/issues/961) and [vim-mode-PR 970](https://github.com/atom/vim-mode/pull/970).

To explain the problem more details, I uploaded a video file.
[video reproduce the problem](https://youtu.be/cu8EDDZt_aU)

Thank you.

PR #21847, opened 2021-01-04
🍎 Don't quit upon 'window-all-closed' event on macOS

### LINKED ISSUE #17672: Atom does not quit on OS X

### LINKED ISSUE #20831: I cannot quit Atom

DeeDeeG:
This prevents a lot of buggy partial-quitting stuff from happening.

<details><summary>Requirements for Contributing a Bug Fix (from template, click to expand)</summary>

</details>

### Identify the Bug

Fixes: https://github.com/atom/atom/issues/17672
Fixes: https://github.com/atom/atom/issues/20831

### Description of the Change

In order to avoid a buggy, "half-quit" state, don't attempt to quit on macOS when the last window is closed.

More technically: listen for and handle the `'window-all-closed'` event in a custom way, overriding the default behavior Electron provides. ([Electron's default behavior for this event simply calls `app.quit()`, regardless of OS.](https://github.com/electron/electron/blob/v11.1.1/lib/browser/init.ts#L193-L198) We can do essentially the same, but conditionally not quit if the OS is `'darwin'` meaning macOS.)

_[`'window-all-closed'` event documentation at electronjs.org](https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed)_

In this way, we can allow Atom to stay running in the background ready to open new windows quickly, as it is intended to.

_(For reasons that aren't entirely clear to me, Atom's code has been partly preventing this quitting, but still setting an `atomApplication.quitting` variable to `true`, which later on causes some buggy behavior around quitting and closing windows. This pull request addresses and prevents the bugginess.)_

### Alternate Designs

- Remove the "Atom process in the background" feature altogether, to avoid these bugs: #21843
  - Having Atom run in the background, ready to open new windows quickly, is a desirable feature. So we should fix the underlying issue rather than disable the feature.
- Introduce some way of force quitting from within the app, that is call [`app.exit()`](https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#appexitexitcode)?
  - Not necessary now that Atom behaves itself and quits properly when asked to. Most of the time. 😛

### Possible Drawbacks

None.

### Verification Process

I built Atom locally with this change. I confirmed that the steps to reproduce identified in https://github.com/atom/atom/issues/17672#issuecomment-753424712 are no-longer reproducible.

CI passes.

### Release Notes

Fix buggy quitting behavior on macOS after all windows are closed

### DIFF
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -455,8 +455,7 @@ module.exports = class AtomApplication extends EventEmitter {
   removeWindow(window) {
     this.windowStack.removeWindow(window);
     if (this.getAllWindows().length === 0 && process.platform !== 'darwin') {
-      // We should be quitting now in the 'window-all-closed' event handler.
-      // We can skip anything past this point in removeWindow().
+      app.quit();
       return;
     }
     if (!window.isSpec) this.saveCurrentWindowOptions(true);
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -768,13 +768,13 @@ module.exports = class AtomApplication extends EventEmitter {
       })
     );
 
-    // Don't quit when the last window is closed on macOS. Overrides the default Electron behavior.
-    // See: https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed
+    // See: https://www.electronjs.org/docs/api/app#event-window-all-closed
     this.disposable.add(
       ipcHelpers.on(app, 'window-all-closed', () => {
         if (this.applicationMenu != null) {
           this.applicationMenu.enableWindowSpecificItems(false);
         }
+        // Don't quit when the last window is closed on macOS.
         if (process.platform !== 'darwin') {
           app.quit();
         }
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -773,6 +773,16 @@ module.exports = class AtomApplication extends EventEmitter {
       })
     );
 
+    // Don't quit when the last window is closed on macOS. Overrides the default Electron behavior.
+    // See: https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed
+    this.disposable.add(
+      ipcHelpers.on(app, 'window-all-closed', () => {
+        if (process.platform !== 'darwin') {
+          app.quit();
+        }
+      })
+    );
+
     // Triggered by the 'open-file' event from Electron:
     // https://electronjs.org/docs/api/app#event-open-file-macos
     // For example, this is fired when a file is dragged and dropped onto the Atom application icon in the dock.
--- a/src/main-process/atom-application.js
+++ b/src/main-process/atom-application.js
@@ -454,14 +454,10 @@ module.exports = class AtomApplication extends EventEmitter {
   // Public: Removes the {AtomWindow} from the global window list.
   removeWindow(window) {
     this.windowStack.removeWindow(window);
-    if (this.getAllWindows().length === 0) {
-      if (this.applicationMenu != null) {
-        this.applicationMenu.enableWindowSpecificItems(false);
-      }
-      if (['win32', 'linux'].includes(process.platform)) {
-        app.quit();
-        return;
-      }
+    if (this.getAllWindows().length === 0 && process.platform !== 'darwin') {
+      // We should be quitting now in the 'window-all-closed' event handler.
+      // We can skip anything past this point in removeWindow().
+      return;
     }
     if (!window.isSpec) this.saveCurrentWindowOptions(true);
   }
@@ -777,6 +773,9 @@ module.exports = class AtomApplication extends EventEmitter {
     // See: https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed
     this.disposable.add(
       ipcHelpers.on(app, 'window-all-closed', () => {
+        if (this.applicationMenu != null) {
+          this.applicationMenu.enableWindowSpecificItems(false);
+        }
         if (process.platform !== 'darwin') {
           app.quit();
         }
--- a/src/main-process/start.js
+++ b/src/main-process/start.js
@@ -104,14 +104,6 @@ module.exports = function start(resourcePath, devResourcePath, startTime) {
     })
   );
 
-  // Don't quit when the last window is closed on macOS. Overrides the default Electron behavior.
-  // See: https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed
-  app.on('window-all-closed', () => {
-    if (process.platform !== 'darwin') {
-      app.quit();
-    }
-  });
-
   if (args.userDataDir != null) {
     app.setPath('userData', args.userDataDir);
   } else if (args.test || args.benchmark || args.benchmarkTest) {
--- a/src/main-process/start.js
+++ b/src/main-process/start.js
@@ -104,6 +104,14 @@ module.exports = function start(resourcePath, devResourcePath, startTime) {
     })
   );
 
+  // Don't quit when the last window is closed on macOS. Overrides the default Electron behavior.
+  // See: https://github.com/electron/electron/blob/v11.1.1/docs/api/app.md#event-window-all-closed
+  app.on('window-all-closed', () => {
+    if (process.platform !== 'darwin') {
+      app.quit();
+    }
+  });
+
   if (args.userDataDir != null) {
     app.setPath('userData', args.userDataDir);
   } else if (args.test || args.benchmark || args.benchmarkTest) {


Aerijo:
I would put this in `atom-application` with the other lifetime event handlers. But otherwise seems fine. I was worried it would lead to a double quit on non-macOS, but then realised that the behaviour would be no different to the original default handler, so this PR wouldn't be introducing any issues relating to that.

It would be better to be consistent with the existing check, so picking one of `['win32', 'linux'].includes(process.platform)` or `process.platform !== 'darwin'` and make both do it. This is mostly a style change, as I don't imagine Atom will ever support another OS. The latter is technically better if it were to though, as it's safer to close the app when all windows are gone if the behaviour gets overlooked.

DeeDeeG:
Thanks for the review.

> I would put this in atom-application with the other lifetime event handlers.

I can do that. This appears to work just as well to fix the bug: https://github.com/DeeDeeG/atom/commit/af1daa9917a42a13664b117bc8bd8f3bb76e6b25

Can you give some thoughts as to why that's better? I'm not intimately familiar with these files, and would like to know what the benefit is.

> It would be better to be consistent with the existing check, so picking one of `['win32', 'linux'].includes(process.platform)` or `process.platform !== 'darwin'` and make both do it.

Hmm. I don't know which is faster, but since these are done once in a blue moon each, and they aren't in loops, it likely doesn't matter, they're both blazingly fast, I'd imagine. I think that `!== 'darwin'` seems more honest than `['win32', 'linux'].includes()`. If Atom got ported to BSD, for example, the code that runs there tends to be more Linux-ey than macOS-ey. And other things are probably more Windows-ey than macOS-ey. (macOS is rather proprietary and secretive, so it's less likely people will attempt a compatible or closely-related piece of software that doesn't get to run on actual macOS...) If there is an actual fourth OS supported by Atom, they surely will look at these sorts of checks as a first order of business when porting... So whoever tries to port this to a fourth OS, I think this is ultimately their problem to sort out. Not that we can't try to leave things in good order.

I think I'd like a second opinion on the need for updating all of those, as smaller diffs are usually my preference. I'd be a tiny bit sad to go from `!== 'darwin'` to `['win32', 'linux'].includes()` for this little listener, but I don't have any technical reason, just aesthetic preference mostly.

> The latter [`!== 'darwin'`] is technically better if [Atom] were to [support another OS] though, as it's safer to close the app when all windows are gone if the behaviour gets overlooked.

Hmm, makes sense. 👍

Overall: Going to think this over and/or wait for some more feedback before I go ahead and add either of these changes to the PR.

DeeDeeG:
> https://github.com/atom/atom/issues/17672#issuecomment-753774155
> @DeeDeeG You can try moving the other 0 windows logic to that handler too, [like this](https://github.com/Aerijo/atom/commit/88780a28d23f6a02f752ea59ce240c700fd1b30e).

It would be nice to take a look at this suggestion again before this PR got merged, if possible, but I can't really properly test this right at the moment. Leaving this as a note-to-self to try it out, and for others to comment on if they have any feedback on that suggestion.

Aerijo:
> This appears to work just as well to fix the bug

It does, I just think it's better to have similar code together. The `atom-application` file already has the other 0 window handling logic, as well as the quit and most of the other app event handlers.

I've only looked at these files for the first time myself mind.


> If there is an actual fourth OS supported by Atom, they surely will look at these sorts of checks as a first order of business when porting

Maybe, or they might just try to build it and fix errors as they come, like with the electron upgrades. This wouldn't error at build time, and isn't tested, so it would be easily missed. Speed has nothing to do with it (as you point out).


> I think I'd like a second opinion on the need for updating all of those

I only meant [this](https://github.com/atom/atom/blob/f1b0e3b47b30ed9379821b341b3337696b8a2a12/src/main-process/atom-application.js#L461) other instance. It just seems weird to me to be using two different ways of checking the same condition in the same file (assuming it's moved). 

These were just small things though, I'm happy with the change itself and believe it would work as is without adding regressions.

DeeDeeG:
I think I'll just test your commit some more (when I get a good moment to!) -- It's a pretty clean-looking change and tidies up the now-redundant code a bit. Seems like nothing wrong with that.

I'm mildly nervous at changing the exact timing of things, but as I'm looking at this code, it doesn't feel like every bit of it has been cautiously audited for race conditions every step of the way, to be quite honest. So the status quo may or may not be something to look to as an example. (So... testing is the way!)

DeeDeeG:
TODO:
- Test on Windows and/or Linux to confirm quitting works as intended there.
- Decide exactly which tweaks/cleanups to do or not do. Testing a few variations on this and inspecting/debugging.

src/main-process/atom-application.js
```diff
@@ -454,14 +454,10 @@ module.exports = class AtomApplication extends EventEmitter {
...
   removeWindow(window) {
     this.windowStack.removeWindow(window);
-    if (this.getAllWindows().length === 0) {
-      if (this.applicationMenu != null) {
-        this.applicationMenu.enableWindowSpecificItems(false);
-      }
-      if (['win32', 'linux'].includes(process.platform)) {
-        app.quit();
-        return;
-      }
```
DeeDeeG, line 464: If the `app.quit()` call is removed from `removeWindow()`, then there are background process left not-cleaned-up after the last window is closed on Windows.

An observation: Most of these event handlers seem to expect to be called more than once. The `'close'` and `'closed'` handlers car run many times with no issue. The `'before-quit'` almost doesn't seem to work fully if not called multiple time. And I've had it called multiple times in a lot (all?) of my non-buggy quits that I've inspected and stepped through.

Not to get into too broad a discussion, but: I wish these handlers were re-written to be simpler, less interdependent on one-another, and most especially I would like to see them reliably work when run only once each.

DeeDeeG: Here's what a successful quit looks like on Windows, according to my debugging:

```js
'close' event handler:
  event.preventDefault()
  this.unloading = true
  saveCurrentWindowOptions()

 // [async/unrelated]
 blurHandler() // 'blur' event emitted because Atom is unfocused while I'm looking at the debugger
   saveCurrentWindowOptions() // This function runs a lot!
 // [end async/unrelated]

// still in the 'close' event handler:
  browserWindow.close() // Why do this?? Is this how we communicate with Chromium?
    'close' event handler // Nested/self-call.. is this necessary?? No-op in our code because this.unloading === true

'window-all-closed' event handler:
  app.quit()
    'before-quit' event handler:
      event.preventDefault() // I think Atom forgets to actually explicitly quit the first time around, which the default behavior is meant to do?

'closed' event handler:
  removeWindow()
    app.quit()
      'before-quit' event handler // event.preventDefault() is skipped because this.quitting === true, so I guess we actually quit now... But that's just a guess?
      'will-quit' event handler:
        this.killAllProcesses() // Or is this when we actually quit? Perhaps this is only triggered when event.preventdefault() is skipped in the 'before-quit' handler?
```

This is what happens when pressing the red "x" button on the last open window on Windows.

DeeDeeG:
Yeah, it seems to me that the `'before-quit'` event handler doesn't actually make Atom quit unless it runs twice.

DeeDeeG:
@sadick254 Please consider reviewing/merging this PR. More details below:

---

There has been an issue for a long time on macOS where closing the last window puts Atom in a "half-quit" state. _(In this state, opening and closing another window either instantly quits Atom, or causes Atom to become un-quitable unless resorting to some sort of "Force Quit" action. Either of these outcomes is unexpected and undesirable for Atom users on macOS.)_

This Pull Request solves the underlying problem: Atom did not handle the `'window-all-closed'` event, so the default behavior from Electron was used. The default behavior from Electron is to quit, regardless of OS. _(This is not intended behavior for Atom on macOS; Atom should stay open in the background, even with no windows up, unless and until it is Quit.)_

_(Technical detail: Atom must run its `'before-quit'`event handler at least twice in order to properly quit. This bug was caused by activating the `'before-quit'` event and the associated handler only once.)_

I have tested this Pull Request on macOS, Windows and Linux. On all three operating systems, I was able to open and close windows as expected, without running into the bug on macOS. I can quit with multiple windows open and restore them. And so on. Generally everything is working. CI also passes.

So if you or another Atom maintainer can take a look, it would be much appreciated.


Thank you.

\- DeeDeeG

src/main-process/atom-application.js
Aerijo: The original close is cancelled and cleanup is done. The second call to close is done to resume the close operation, which shouldn't be prevented the second time, hence the `close` handler being a no-op. This seems reasonable to me; the idea is to reschedule the close to after some potentially asynchronous work is finished.

Atom doesn't exactly forget to quit the first time, it's just waiting for a response from a closed window (if I recall my debugging correctly). But while waiting for the window promises to resolve Atom is left in a state that really shouldn't be allowed (with `quitting` true, so subsequent quits will actually do a quit, but with quitting work still being performed).

The above problem is exactly what happens during the removeWindow quit. As the before-quit handler does not cancel the event this time, the quit progresses and the `will-quit` handler is called after all windows have been closed.

I've been looking at the Electron source code that handles all this because I suspect the original issue is caused by Electron itself getting into an invalid quitting state. That's my only hypothesis for why app.quit() can run, but the before-quit handler doesn't.

DeeDeeG: Thank you, this is definitely helpful to have laid out in clearer terms. And the more I've looked at this I've tended closer to seeing it how you described.

The pattern I'm seeing could be summarized shortly like this: "disable default handling" --> "custom work" --> "Re-enable default handling, or finish action in Atom's code manually if need be"

> I've been looking at the Electron source code that handles all this because I suspect the original issue is caused by Electron itself getting into an invalid quitting state.

Yes, past a certain point you can attempt to quit from the app any number of times and any number of ways and nothing happens.  It's hard for me to fathom how Atom's code alone could cause this. And the JS debugger becomes nearly useless to track what's going on, if anything, so I suspect native code is involved at or near the point of failure... (Though I can't say for sure with the experience and information I have with this).

Aerijo (approved):
LGTM @sadick254

sadick254:
@DeeDeeG This is on my radar. It is a small change and I will review it ASAP.

rseichter:
I'm very much looking forward to this fix becoming available as part of an official Atom release. My thanks to @DeeDeeG and all others involved.

src/main-process/atom-application.js
DeeDeeG: @Aerijo I would like it if Atom's `'before-quit'` handler ran only once, and simply emitted the `'will-quit'` event when it thought it was done.

That's the kind of cleanup and simplification I would love to see. No need to re-enable the default handling for `'before-quit'`. No need to run through any `'before-quit'` handler more than once.

It should all be flattened out so most event handlers don't need to run more than once. (My opinion).

Aerijo: I don't think that's possible; Electron events are synchronous, but Atom needs to support asynchronous package deactivation (among other things). The only option is to cancel the first quit and run it again when the asynchronous work is done.

I do agree that the current design doesn't do that correctly though. I'll probably look into fixing it, but it would be a high impact change and might take a long time to develop and document.

DeeDeeG: Using just the native handlers is probably impossible.

I do wonder if the handlers could just use more async/await or promises to do what they need to do in one go, though.

Aerijo: The handlers just don't support asynchronous code in any sense. If you don't cancel the event by the time it returns, then it will continue. The fact some of the handlers return promises is misleading (the return value is not used).

And the handlers are the entry point to the quitting / closing, so we can't start any of the work before a handler runs, so I can't see any way around cancelling the first and triggering it again when ready.

DeeDeeG: What I mean is, if there is anything we can do to adequately perform quits and closes in Atom's codebase, we could avoid any interaction with the native handlers (beyond preventing their default behaviors). And gain the flexibility to neaten up the code a lot.

Aerijo: The only way to quit / close is through the Electron API, so the corresponding events will always run. Do you mean something like unregistering the event listener when it triggers the first time?

I'm just not really sure how what you're suggesting is different to what it currently does.

DeeDeeG: I'm just a person with a dream who barely understands Electron at all! 😆

_(In theory, anything is possible!)_ It's just a loose goal while I get a taste for Atom's main process code and how electron works.

My strong suit is testing and reading/understanding/writing documentation. But a person can dream...

DeeDeeG: Example of communicating with BrowserWindow directly: https://stackoverflow.com/a/31174463

DeeDeeG: I don't want to take up any more of your time and energy. It was a deliberately provocative proposal, but I thought was clear I didn't know how to implement it. I'm okay with that. Especially while I slowly learn more. It's too big a problem for me to tackle yet, I just want to have a goal in mind to use to help me get better with Electron and Atom. If I don't achieve what I set out to do, hopefully something adjacent to it will be improved by the time I'm done.

Aerijo: It's fine, I'm only looking at this stuff in depth for the first time myself. 

The BrowserWindow is still part of Electron, the only difference in the linked examples being it used on the renderer process (I assume the `remote` module provides a mock of the BrowserWindow that will send all requests to the main process). But using it that way [has issues](https://www.electronjs.org/docs/api/remote#passing-callbacks-to-the-main-process).

Aerijo: @DeeDeeG You might like [this](https://github.com/atom/atom/files/5784509/ElectronFlow.pdf) for a flattened overview of the relevant steps Electron is taking under the hood. The regular boxes mostly correspond to functions, which can be searched for in the Electron code base. Each emit is synchronous. Could be useful for hypothesising and testing the cause of the original issue too.

DeeDeeG: That's interesting. (I assume based on the notes in the yellow boxes this is your work and made for debugging this issue?)

I believe we (Atom) will be in a newer version of Electron soon, so before hypothesizing too much about the Electron side of it, I'd be interested to see if Atom on Electron 9 (with this fix reverted) still exhibits the issue.

Random question if you get a moment: Do you happen to have an SSD? I had reproduced the issue with an HDD.

Aerijo: It is an SSD.

It's possible it's been fixed in Electron. I haven't seen many changes in the codebase for this area (though they've since renamed a lot of files making blames harder to follow), but it's also useful to me for just getting a sense of Electron itself so I'm not too concerned even if it has been patched.

The `>` is just how I notated "function is called". Those boxes aren't required to understand the event order, but help relating it back to the Electron source code.

- The ovals are code entry points, 
- the deformed boxes are user actions / JS entry points, 
- the barred boxes are references to other entry points in the diagram, 
- diamonds are choices (outgoing arrows answer question / boolean in diamond), 
- the rounded edged boxes represent emitting a JS event, and 
- the wavy boxes are things I don't fully understand yet.

jen801 (commented):
#10967

PR #21861, opened 2021-01-12
Fix buggy partial startup on macOS due to flag added by Gatekeeper/XProtect

### LINKED ISSUE #21013: Application not showing in front, unable to view any file
Install ATOM in Mac (Catalina - 10.15.4), use it for some time, use in full screen, after several times, no file is showing in ATOM. Double click on file, or select open with ATOM, it's showing Atom window open on top but no file is showing in screen.

If I re-install ATOM it works fine for some time, but again getting same problem.

Check this image for sample: https://drive.google.com/file/d/1nllSAPuSUyZztTl3WycVy5whzYOM2gs8/view?usp=sharing

### LINKED ISSUE #21751: On MacOS Catalina, Atom doesn't display a window at launch when launched via icon (click on icon) or via launcher (e.g. Alfred or Spotlight) but does produce one when launched via the 'atom' command in a terminal
### Prerequisites

### Description

On MacOS Catalina 10.15.7, Atom 1.53 (installed via homebrew) doesn't display a window at launch when launched via icon (click on icon) or via launcher (e.g. Alfred or Spotlight) but does produce one when launched via the 'atom' command in a terminal

### Steps to Reproduce

1. Click on Atom icon or launch via launcher (Spotlight or Alfred)

**Expected behavior:**

Main Atom window created

**Actual behavior:**

No window is created. No window can be opened via keyboard shortcut or by trying to open recent documents.

**Reproduces how often:**

Reproducible 100% of the time

### Versions

atom --version
Atom    : 1.53.0
Electron: 6.1.12
Chrome  : 76.0.3809.146
Node    : 12.4.0

apm --version
apm  2.5.2
npm  6.14.8
node 12.4.0 x64
atom 1.53.0
python 3.7.9
git 2.29.2

### Additional Information

Behaviour is as expected if atom is launched from the command line via the command 'atom'.

This seems related to issue #21013

### LINKED ISSUE #21820: MacOS new install Menu has no options
### Description

Non-technical colleague downloaded Atom 1.53 from the website, atom-mac.zip for MacOS Mojave 14.6 and his Atom app appeared with a menu bar but almost no options... No way to create a new file or open a file. 

In finder, he was able to find the option to Open With Atom for a text file and this did nothing.

### Steps to Reproduce

1. Download Atom-mac.zip
2. Open Atom

**Expected behavior:**

menu has file options

**Actual behavior:**

menu does not have file options

one user on Mac Mojave 14.6 always

### Versions
1.53 atom

DeeDeeG:
<details><summary>Requirements for Contributing a Bug Fix (from template, click to expand)</summary>

</details>

### Identify the Bug

<details><summary>Deeper technical explanation (click to expand)</summary>

_(Paraphrased from: https://github.com/atom/atom/issues/21751#issuecomment-757617110)_

[This bug appears] related to Gatekeeper.

[This] bug happens [when you] are launching Atom and see Gatekeeper pop up with a loading bar and a message 'Verifying "Atom"', or something about "This was downloaded from the internet, are you sure you want to open this?". This bug happens when Gatekeeper [intervenes in] launching the app [but the user clicks though/the app is verified and launching continues].

I see a command-line flag `-psn_0_[some six or seven digits here]` (something like `-psn_0_1234567`) applied to the buggy Atom process, when inspecting using [`htop`](https://htop.dev) or `ps -ax`. I have seen reports elsewhere that this flag is added by Gatekeeper. For example: https://github.com/godotengine/godot/pull/37719

Our argument parser library `yargs` thinks this `-psn_0_1234567` flag is a bunch of short flags, equivalent to `-p`, `-s`, `-n`, `-_`, `-0`, `-_`, `-1`, `-2`, `-3`, `-4`, `-5`, `-6`, `-7`. The `-_` are converted to boolean `true`, and then we try to do JS string-only operations on the flags. Doing [`true.startsWith()`](https://github.com/atom/atom/blob/v1.53.0/src/main-process/parse-command-line.js#L190) results in a TypeError ("`path.startsWith` is not a function") very early in our initialization process, leaving the app barely started up at all, and unable to respond to most kinds of input other than quitting.

Quitting and restarting means Gatekeeper is already familiar with the binary, and it won't intervene with launching the app. So the flag doesn't get added, and Atom works fine after a restart.

</details>

Fixes: https://github.com/atom/atom/issues/21013 (probably)
Fixes: https://github.com/atom/atom/issues/21751
Fixes: https://github.com/atom/atom/issues/21820

### Description of the Change

- Filter out a flag (`-psn_0_[1234567]`) that macOS Gatekeeper/XProtect adds to Atom. Filter it out before `yargs` tries to parse it.
  - More-precisely: Filter out any argument passed to Atom that starts with `-psn_`.
  - (This flag is bogus input for our purposes. `yargs` thinks this is many shortflags (`-p` `-s` `-n` `-_` `-0` `-_`...); Parsing and handling all of them without it causing trouble is somewhat difficult. Filtering it out early avoids all the complications and bugs (TypeErrors, partial startups).)
- Also, as a further step to avoid problems from potentially similar flags in the future, completely ignore non-string items in the `args._` array, in one of the functions that decides whether an argument is a path or a URI.

### Alternate Designs

We could use just one of these two fixes instead of both. I included both to be thorough and handle more potential situations more correctly.

<details><summary>Miscellaneous other proposals (click to expand):</summary>

 I also considered converting number arguments to strings, so users can easily open a file like `55` or `100` in the current directory (no file extension or relative/absolute path needed). But I left that out because it felt like it was veering off-topic for a bug-fix. (`yargs` has [an option](http://yargs.js.org/docs/#api-reference-stringkey) that supposedly would do that for us, but it doesn't help very much in my testing; we would still need to filter out the zero-length strings it converts boolean `true` to with that option, and we would still need to manually convert numbers to strings, because that part doesn't work. I think that's a bug in `yargs`.)

</details>

### Possible Drawbacks

None.

_(I expected tiny slowdowns, but the relevant tests run at the same ~16-18ms speed on this branch as `master` branch on my machine.)_

### Verification Process

Added tests to describe new expected behavior. Tests pass.

macOS Gatekeeper flag no-longer prevents Atom from completely starting up.

To reproduce the problem, run this in the Terminal app: `atom "-psn_0_1234567"`. You should see the partial startup bug as described in the issues (linked at the top of this PR).

Try again on this branch to confirm the bug goes away; For example, using Atom in dev mode:

```
cd atom
git remote add deedeeg https://github.com/DeeDeeG/atom
git remote update deedeeg
git switch bugfix-deal-with-macos-gatekeeper-flag
./out/Atom\ Dev.app/Contents/MacOS/Atom\ Dev --dev --resource-path=/Users/[you]/path/where/you/cloned/atom -psn_0_1234567
```

### Release Notes

Fixed buggy partial startup on macOS due to flag added by Gatekeeper/XProtect

### DIFF
--- a/spec/main-process/parse-command-line.test.js
+++ b/spec/main-process/parse-command-line.test.js
@@ -41,4 +41,27 @@ describe('parseCommandLine', () => {
       assert.deepEqual(args.pathsToOpen, []);
     });
   });
+
+  describe('when evil macOS Gatekeeper flag "-psn_0_[six or seven digits here]" is passed', () => {
+    it('ignores any arguments starting with "-psn_"', () => {
+      const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
+      const args = parseCommandLine([
+        getPsnFlag(),
+        '/some/path',
+        getPsnFlag(),
+        getPsnFlag(),
+        'some/other/path',
+        'atom://test/url',
+        getPsnFlag(),
+        'atom://other/url',
+        '-psn_ Any argument starting with "-psn_" should be ignored, even this one.',
+        './another-path.file'
+      ]);
+      assert.deepEqual(args.urlsToOpen, [
+        'atom://test/url',
+        'atom://other/url'
+      ]);
+      assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+    });
+  });
 });
--- a/spec/main-process/parse-command-line.test.js
+++ b/spec/main-process/parse-command-line.test.js
@@ -24,7 +24,6 @@ describe('parseCommandLine', () => {
 
     describe('and --_ or -_ are passed', () => {
       it('does not attempt to parse booleans as paths or URIs', () => {
-        const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
         const args = parseCommandLine([
           '--_',
           '/some/path',
@@ -44,13 +43,16 @@ describe('parseCommandLine', () => {
           'atom://test/url',
           'atom://other/url'
         ]);
-        assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+        assert.deepEqual(args.pathsToOpen, [
+          '/some/path',
+          'some/other/path',
+          './another-path.file'
+        ]);
       });
     });
 
     describe('and a non-flag number is passed as an argument', () => {
       it('does not attempt to parse numbers as paths or URIs', () => {
-        const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
         const args = parseCommandLine([
           '43',
           '/some/path',
@@ -67,7 +69,11 @@ describe('parseCommandLine', () => {
           'atom://test/url',
           'atom://other/url'
         ]);
-        assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+        assert.deepEqual(args.pathsToOpen, [
+          '/some/path',
+          'some/other/path',
+          './another-path.file'
+        ]);
       });
     });
   });
@@ -93,7 +99,9 @@ describe('parseCommandLine', () => {
 
   describe('when evil macOS Gatekeeper flag "-psn_0_[six or seven digits here]" is passed', () => {
     it('ignores any arguments starting with "-psn_"', () => {
-      const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
+      const getPsnFlag = () => {
+        return `-psn_0_${Math.floor(Math.random() * 10_000_000)}`;
+      };
       const args = parseCommandLine([
         getPsnFlag(),
         '/some/path',
@@ -110,7 +118,11 @@ describe('parseCommandLine', () => {
         'atom://test/url',
         'atom://other/url'
       ]);
-      assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+      assert.deepEqual(args.pathsToOpen, [
+        '/some/path',
+        'some/other/path',
+        './another-path.file'
+      ]);
     });
   });
 });
--- a/spec/main-process/parse-command-line.test.js
+++ b/spec/main-process/parse-command-line.test.js
@@ -22,6 +22,10 @@ describe('parseCommandLine', () => {
       assert.deepEqual(args.pathsToOpen, ['/some/path']);
     });
 
+    // The "underscore flag" with no "non-flag argument" after it
+    // is the minimal reproducer for the macOS Gatekeeper startup bug.
+    // By default, it causes the addition of boolean "true"s into yargs' "non-flag argument" array: `argv._`
+    // Whereas we do string-only operations on these arguments, expecting them to be paths or URIs.
     describe('and --_ or -_ are passed', () => {
       it('does not attempt to parse booleans as paths or URIs', () => {
         const args = parseCommandLine([
--- a/spec/main-process/parse-command-line.test.js
+++ b/spec/main-process/parse-command-line.test.js
@@ -21,6 +21,55 @@ describe('parseCommandLine', () => {
       ]);
       assert.deepEqual(args.pathsToOpen, ['/some/path']);
     });
+
+    describe('and --_ or -_ are passed', () => {
+      it('does not attempt to parse booleans as paths or URIs', () => {
+        const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
+        const args = parseCommandLine([
+          '--_',
+          '/some/path',
+          '-_',
+          '-_',
+          'some/other/path',
+          'atom://test/url',
+          '--_',
+          'atom://other/url',
+          '-_',
+          './another-path.file',
+          '-_',
+          '-_',
+          '-_'
+        ]);
+        assert.deepEqual(args.urlsToOpen, [
+          'atom://test/url',
+          'atom://other/url'
+        ]);
+        assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+      });
+    });
+
+    describe('and a non-flag number is passed as an argument', () => {
+      it('does not attempt to parse numbers as paths or URIs', () => {
+        const getPsnFlag = () => { return `-psn_0_${Math.floor(Math.random() * 10_000_000)}` };
+        const args = parseCommandLine([
+          '43',
+          '/some/path',
+          '22',
+          '97',
+          'some/other/path',
+          'atom://test/url',
+          '885',
+          'atom://other/url',
+          '42',
+          './another-path.file'
+        ]);
+        assert.deepEqual(args.urlsToOpen, [
+          'atom://test/url',
+          'atom://other/url'
+        ]);
+        assert.deepEqual(args.pathsToOpen, ['/some/path', 'some/other/path', './another-path.file']);
+      });
+    });
   });
 
   describe('when --uri-handler is passed', () => {
--- a/spec/main-process/atom-application.test.js
+++ b/spec/main-process/atom-application.test.js
@@ -724,7 +724,7 @@ describe('AtomApplication', function() {
     });
 
     it('opens a file to a specific line number and column', async function() {
-      await scenario.open(parseCommandLine('b/2.md:12:5'));
+      await scenario.open(parseCommandLine(['b/2.md:12:5']));
       await scenario.assert('[_ 2.md]');
 
       const w = scenario.getWindow(0);
@@ -750,7 +750,7 @@ describe('AtomApplication', function() {
     });
 
     it('truncates trailing whitespace and colons', async function() {
-      await scenario.open(parseCommandLine('b/2.md::  '));
+      await scenario.open(parseCommandLine(['b/2.md::  ']));
       await scenario.assert('[_ 2.md]');
 
       const w = scenario.getWindow(0);
--- a/src/main-process/parse-command-line.js
+++ b/src/main-process/parse-command-line.js
@@ -5,7 +5,13 @@ const yargs = require('yargs');
 const { app } = require('electron');
 
 module.exports = function parseCommandLine(processArgs) {
-  const options = yargs(processArgs).wrap(yargs.terminalWidth());
+  // macOS Gatekeeper adds a flag ("-psn_0_[six or seven digits here]") when it intercepts Atom launches.
+  // (This happens for fresh downloads, new installs, or first launches after upgrading).
+  // We don't need this flag, and yargs interprets it as many short flags. So, we filter it out.
+  const filteredArgs = processArgs
+    .filter(arg => !arg.startsWith('-psn_'));
+
+  const options = yargs(filteredArgs).wrap(yargs.terminalWidth());
   const version = app.getVersion();
   options.usage(
     dedent`Atom Editor v${version}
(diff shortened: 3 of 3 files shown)


DeeDeeG:
@sadick254 and @Aerijo this is another bugfix PR for macOS. Please take a look if you have time.

I am confident this works well, but as for the particular approach, I would be interested in any feedback for this implementation and feedback on the tests I added.

---

For example: There is different function that handles these inputs if the `--uri-handler` flag is passed to Atom. Since the checks and filtering added in this PR ended up not being noticeably slower than `master`, I could move the filtering out of the one function to earlier before either of the two functions.

Also, converting number arguments (for example: `55` `100`) to strings would let you open files with those names. No extension or path hints (`./55`, `/absolute/path/to/55`, `55.txt`) needed. I think that could be neat, but fairly niche in usage and arguably off-topic for this PR. But I could add that in, or do it as another PR.

As mentioned in the body of the PR, this PR includes two similar and largely redundant fixes, but as I also point out this hasn't resulted in slowdowns, and I think safer, more-cautious code during startup may be worth it. That said, I would understand a "Keep It Simple" approach and so I wonder if just one of these solutions is adequate.

DeeDeeG:
Also I understand if looking deep into this bugfix PR isn't terribly interesting. I tested this implementation a lot, and it works, and it's fast. So, no worries if not interested in giving in-depth feedback for this admittedly simple bugfix.

spec/main-process/parse-command-line.test.js
```diff
@@ -21,6 +21,61 @@ describe('parseCommandLine', () => {
       ]);
       assert.deepEqual(args.pathsToOpen, ['/some/path']);
     });
+
+    describe('and --_ or -_ are passed', () => {
```
Aerijo, line 25: Why single out and ignore underscore flags? I guess it's related to `-psn_...` containing an underscore? Is there a drawback to just allowing (and ignoring) it?

Aerijo:
> I expected tiny slowdowns, but the relevant tests run at the same ~16-18ms speed on this branch as master branch on my machine

1. Working at all is better than sometimes failing for the sake of 'performance'
2. If filtering an array of (like 3 / 4 ?) values once, in the entire app lifetime, caused a measurable slowdown, there are bigger problems here.

I was interested in how VS Code handles `-psn_...`, but it seems [they don't](https://github.com/microsoft/vscode/blob/b6435bc42407d53470e1bc0a0261183148bfde5a/src/main.js#L420-L432). I guess they just type check before using values.

And I would like to see the positional arguments always be treated as strings, but that can go in a different PR.

spec/main-process/parse-command-line.test.js
DeeDeeG: _Passing `-_` or `--_`, as either the last argument, or "without a non-flag argument immediately after", is the minimal reproducer for the Gatekeeper partial startup bug._

Maybe I should add this as a code comment in the test file? Or incorporate that into the test's `describe`/`it` text?

---

`yargs` stores non-flag arguments in an array called `argv._`, but this can easily be polluted with the content of actual underscore `_` flags. (See Gatekeeper's `-psn_0_[1234567]`). And `yargs` treats any flag with "no content" associated with it (no no-flag argument passed right after the flag) as boolean `true`s.

Passing the "underscore flag" is the main way to get stray boolean `true`s into the non-flag argument pool. It's an easy and convenient way to test the Gatekeeper flag bug more generally than a flag starting with `-psn_`.

Restated in terms relative to the normal non-bug behavior: This test handles boolean pollution in the non-flag argument array, which is the one where we look for paths and URIs that Atom should open.

<details><summary>Huge long reply, since I'm not totally sure what else you may need clarifying, and I was kind of on a roll. And it's late where I am. (click to expand if there's anything else you were wondering, I may have answered it here):</summary>

_(Before I get into it: Admittedly none of the following complicated information would make sense from just looking at the test as I wrote it. So maybe I should add comments in the test file. Or make the purpose of the test more obvious from the `describe` and `it` text. TODO: make the test self-explanatory.)_

Also to address just this question directly:

> Why single out and ignore underscore flags? I guess it's related to -psn_... containing an underscore? Is there a drawback to just allowing (and ignoring) it?

I felt that it was just a matter of time before we saw a similarly-formed flag. If it didn't happen to start with exactly `-psn_`, then we wouldn't be filtering it out, and then we're back to experiencing essentially this same issue.

The general solution of handling/ignoring shortflag sequences containing effectively `-_`... is very simple. It happens to also filter out `--_`. I may have gone overboard in explicitly testing for both, but `yargs` treats them the same under the hood right now, the solution to handle all of this comprehensively is a few lines of code, and I just wanted to be thorough. I figured being mildly paranoid about it now would make our code and test coverage robust to whatever strange inputs may come, and alert us to whatever undocumented minor behavior changes may happen in `yargs`, for instance.

---

So, the reason I went there at all is a bit of a rabbit-hole of how yargs works. Well, it seems like a rabbit-hole if you don't know where it ends up. I will try to make this concise but complete.

### Background info: what's the normal, desired behavior?

Starting with the normal, desired behavior: `yargs` stores all non-flag arguments in an array called `argv._`.

So, Atom (rather reasonably) expects non-flag arguments to be filenames or URLs we want to open. For example:

Running `atom file1 ./file2 /path/to/file3` will launch Atom with tabs open for these files.

### Quirks in yargs

A few things make it complicated:
- Users and the OS can specify a literal `_` flag of either the short form `-_` or `--_`, and `yargs` will dump it right into the same `argv._` array with the "non-flag" arguments.
- `yargs` by default treats any flag with no (non-flag) argument after it as boolean `true`
- Bonus complication that's not strictly a part of the bug I'm trying to fix: Any bare number passed as a non-flag argument, say `33`, will be turned into a JS `number`. (There is a feature to treat everything in `yargs.argv._` as strings, but it didn't work for numbers in my testing. It didn't coerce numbers to strings in my testing, it left them as JS `number` types. I think this is a bug in `yargs`).

Side note: _(`yargs` seems to follow a Linux-style convention about shortflags. It's a conventional behavior on Linux to treat a string of characters after a single dash as being a number of shortflags. For example, a typical `tar` invocation: `tar -xvf my_tarball.tgz`; in English that's "tar, extract verbosely a file named `my_tarball.tgz`.")_

### The purpose of that test (filtering out booleans)

The purpose of that test is to test an array of inputs that has been polluted with booleans. This is the scenario that the `-psn_0_[1234567]` flag presents.

**_Passing `-_` or `--_`, without a non-flag argument immediately after, is the minimal reproducer for the Gatekeeper partial startup bug._**

### Why I didn't just pass booleans into the parser function

Passing non-strings to this function doesn't resemble its real-world use. The function is only used in one place. [Here in `src/main-process/start.js`](https://github.com/atom/atom/blob/v1.53.0/src/main-process/start.js#L45). Atom always, 100% of the time, passes the `parseCommandLine()` function either an empty array, or an array of strings. These are read from Node's `process.argv` immediately before passing to the `parseCommandLine()` function. Node's `process.argv` will only ever contain strings, apparently. The diverse typing come in only _after_ `yargs` has taken a look at the inputs and decided whether they are strings, numbers, booleans...

</details>

Aerijo: Makes sense. I think a short comment on that test to explain why would be useful.

(turns out yargs has an open bug related to this too https://github.com/yargs/yargs/issues/878)

DeeDeeG:
> I was interested in how VS Code handles `-psn_...`, but it seems they don't. I guess they just type check before using values.

I suppose total and strict coercion to strings (and ignoring/filtering out the things that shouldn't become strings) would be adequate. I found it surprisingly involved to do with based on how `yargs` was acting.

Particularly, [the option to treat everything as strings](http://yargs.js.org/docs/#api-reference-stringkey) seemed to miss bare numbers (they stayed `number` type), and the booleans turned into zero-length strings that would trigger a blank tab/new document to be opened in Atom. So we are back to explicitly filtering out numbers and deleting what would have been the boolean `true`s, deleting them as zero-length strings instead. (That admittedly sounds safer, in a way; never even touching a boolean is one small additional degree away from TypeErrors).

I didn't use this feature as it added a line of code and some processing time in the `yargs` library to practically _almost_ zero effect.

DeeDeeG:
_(Note: It is very late where I am and I must go to sleep. But I am planning on adding a comment to the test tomorrow so it's explained to some extent within the test file. It admittedly does not make intuitive sense if just looking at what's in that file at the moment. What underscore flags have to do with booleans is an implementation detail from yargs. I'll think of how to phrase it.)_

DeeDeeG:
Thinking of how VSCode doesn't specifically handle this flag:

The supposed disadvantage of that for us would be letting all the psuedo-random number shortflags through. (`-1`, `-2`... whatever the Process Serial Number happens to be.)

We don't use _any of these_ for anything right now. But I'm thinking ahead that we might want to use them for something in the future? That's a hypothetical concern as of the moment though. In actuality there's not much real-world-relevant justification I can give for specifically and specially handling that flag.

(It's fast/practically free performance-wise. So there's not really any harm either.)

---

I'm running out of things to discuss on this one, so I think it might be ready.

ThatXliner:
If your comparing it to VS CODE, you might as well just ~~plagiarise~~ copy the code because that’s what coders do.

DeeDeeG:
It's more food for thought/sanity check, two heads are better than one, etc. Considering the reasonable alternatives. But I sense you may not be entirely literally serious in your comment.

DeeDeeG:
Test failure is unrelated. It's the pane-spec test on the Windows renderer-1 tests job again.

DeeDeeG:
Here's the other function that processes the array meant to contain paths and URIs. Just like the function I _did_ add protections to, this one also absolutely expects its inputs to be strings (or it gives `TypeError`s). It is the `--uri-handler` function:

https://github.com/atom/atom/blob/v1.54.0/src/main-process/parse-command-line.js#L137-L144

I left this function alone and didn't add any protections to it, because I wanted a small diff and to keep the parser simple and readable, and because in my testing, this flag is _only_ organically used on Windows, so the offending macOS-only bug is not applicable.

<details><summary>Miscellaneous notes on this (click to expand):</summary>

- When a web browser uses Atom as an `atom://` URI handler (for example, from a package's page at https://atom.io/packages, it only passes two flags: `--uri-handler` and `atom://settings-view/show-package?package=[insert package name here]`
  - This constrained scenario has no obvious way to give bad inputs to the `--uri-handler` function. It should be safe enough for real-world use.
- In theory, it would be good to handle non-string inputs to this function, in case someone is making some bash/batch scripts to take advantage of this flag. Or has a habit of using `atom://` URIs from the command-line. But that is both a niche use-case, and a scenario in which the person writing the script or manually composing the flags can do some due diligence in order to avoid passing non-strings.
- Surprisingly, the Linux app doesn't even try to register itself to the OS as a handler of `atom://` URIs? Should probably add this functionality if possible.
- On macOS, paths and URIs are not handed as CLI arguments to Atom. There are apparently macOS APIs that trigger Electron events [`open-file`](https://www.electronjs.org/docs/api/app#event-open-file-macos) or [`open-url`](https://www.electronjs.org/docs/api/app#event-open-url-macos) that give the URL or file to Electron/Atom. So this function is never "organically" used on macOS. Only via typing it manually, or potentially via scripting, I suppose.
- This flag is not publicly documented, so you'd pretty much have to be hacking on Atom to be aware of the possibility of using it. And this flag wouldn't bring most users any advantage over not using this flag, so the chance of this being a popular thing to do are slim.
- This flag is also not part of a public API, so by implication we've not made any guarantees about supporting its use in unexpected ways.

</details>

(That's the last topic _I_ can think of to discuss here. Recapping some previous comments I posted: I'm on the fence about whether to remove the specific filter for `-psn_` flags, but I'm of the opinion the general fix for ignoring non-string inputs in the main path/URI parser function should definitely stay. Open to any feedback or alternate opinions though.)

sadick254 (approved):

DeeDeeG:
@sadick254 Thank you for merging!

PR #21877, opened 2021-01-19
fixed the catch statement causing an error

SamuelKiragu:


sadick254 (commented):
Hey @SamuelKiragu. Thank you for taking the time to contribute to Atom.

In this case, we are not handling the error. It is, therefore, ok to ignore the error and omit the exception value.
Here is some documentation you can read through about [the exception identifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_exception_identifier)

That being said, you can still add a console message to make the error visible. `console.log(error)`

SamuelKiragu:
Thanks

PR #21880, opened 2021-01-20
Revert  tree-sitter-rust@0.17.0

sadick254:
This reverts commit ea5dbdfb571595bb327318a1a4748fde82b1c9b7 due to failing tests in our production branch. We need to further investigate why the failures are happening. 

I am currently unable to reproduce the failures locally.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4861,17 +4861,7 @@
     "language-rust-bundled": {
       "version": "file:packages/language-rust-bundled",
       "requires": {
-        "tree-sitter-rust": "^0.17.0"
-      },
-      "dependencies": {
-        "tree-sitter-rust": {
-          "version": "0.17.0",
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Looking at the `package-lock.json` diff, it appears that this commit didn't just change the *version* of `tree-sitter-rust` that was used; it also changed *where* the package was placed in the package tree. It might be worth redoing the upgrade, but  making sure to keep `tree-sitter-rust` where it was before: as a dependency of the `language-rust-bundled` package.

DeeDeeG:
The brittleness of CI for moved dependencies was fixed in https://github.com/atom/atom/pull/21884.

Now that that's fixed, it would be nice to see the `language-rust-bundled` update merged again. I could do a PR if desired.

